PR #15815, opened 2020-01-08
Record and publish ghproxy github request timeout

### LINKED ISSUE #15600: Transient ghproxy "context deadline exceeded" logged as error
Seeing errors like:

jsonPayload: {
  cache-key: "https://api.github.com/repos/kubernetes/kubernetes/pulls/86222"   
  component: "ghproxy"   
  error: "context deadline exceeded"   
  file: "ghproxy/ghcache/coalesce.go:125"   
  func: "k8s.io/test-infra/ghproxy/ghcache.(*requestCoalescer).RoundTrip.func1"   
  level: "error"   
  msg: "Error from cache transport layer."   
 }

jsonPayload: {
  cache-key: "https://api.github.com/repos/kubernetes/kubernetes/pulls/86222"   
  component: "ghproxy"   
  error: "context deadline exceeded"   
  file: "ghproxy/ghcache/ghcache.go:167"   
  func: "k8s.io/test-infra/ghproxy/ghcache.upstreamTransport.RoundTrip"   
  level: "error"   
  msg: "Error from upstream (GitHub)."   
 }

Sounds like this is a transient error and not a real concern (found it logged several times at 10:37 PST), so think we want to downgrade the error to a warning.

/kind oncall-hotlist
/area prow

adshmh:
Fixes #15600

### DIFF
--- a/ghproxy/ghcache/coalesce.go
+++ b/ghproxy/ghcache/coalesce.go
@@ -122,7 +122,7 @@ func (r *requestCoalescer) RoundTrip(req *http.Request) (*http.Response, error) 
 		waiter.L.Unlock()
 
 		if err != nil {
-			logrus.WithField("cache-key", key).WithError(err).Error("Error from cache transport layer.")
+			logrus.WithField("cache-key", key).WithError(err).Warn("Error from cache transport layer.")
 			return nil, err
 		}
 		cacheMode = cacheResponseMode(resp.Header)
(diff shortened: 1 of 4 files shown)


adshmh:
Note: I have used the name "github_request_timeouts", but currently this covers all GitHub request errors. I think either the metric's name needs to be changed or a checking of the error needs to be done to only count actual timeout errors, but not sure which is the intended behaviour.

PR #15817, opened 2020-01-08
enable race checks for structured-merge-diff

kwiesmueller:
We would like to enable go race checks for https://github.com/kubernetes-sigs/structured-merge-diff

I tried running the job locally to see if my change works, but didn't get it to work using phaino :-/


apelisse:
Not an expert here but that looks good to me, any downside to using `-race`?

kwiesmueller:
> The cost of race detection varies by program, but for a typical program, memory usage may increase by 5-10x and execution time by 2-20x.
https://golang.org/doc/articles/race_detector.html?h=race

kwiesmueller:
From my local test I didn't feel a difference.
Only when running the tests multiple times in parallel, but that's slow without race as well.

PR #15822, opened 2020-01-09
pull-cip-e2e: use new e2e entrypoint path

listx:
This should fix the job for changes to the promoter, taking into account the refactoring in https://github.com/kubernetes-sigs/k8s-container-image-promoter/pull/166


listx:
Pinging testinfra oncall

/cc @michelle192837

PR #15825, opened 2020-01-09
Kops - Add periodic e2e for supported distros

hakman:
With this, we should have a more complete picture of how supported distros behave and what needs fixing.

I remember this being discussed during Office Hours.


rifelpet:
nice work! I wonder if we should make a `ubuntu.com` alias for account ID `099720109477`. I would suggest the same for debian.org except apparently they have [two account IDs](https://wiki.debian.org/Cloud/AmazonEC2Image/Marketplace) :( 
I also wonder if we could find a better way to specify the centos 7 AMI rather than just its id `ami-4bf3d731`. we can address this stuff later though.

/approve
/lgtm

hakman:
Thanks @rifelpet. I already have a PR for new aliases in works. I wanted to add also simpler aliases like "centos", "debian9", "debian10" and so on.
For the CentOS AMI, should be easy, but wanted to get first the direction on https://github.com/kubernetes/kops/issues/8291.

PR #15826, opened 2020-01-09
stop skipping some more e2e tests in kind presubmit

BenTheElder:
/cc @spiffxp @aojea 

I'm reasonably confident these should pass fine now, did some testing in good ol PR 46662


BenTheElder:
xref: https://github.com/kubernetes/kubernetes/pull/86924, should also enable TCP CLOSE_WAIT test

aojea:
/lgtm
should we review again the list of skipped tests to check if we can enable more?
or is it something you are tracking already?

BenTheElder:
I'm on it :-)

PR #15699, opened 2019-12-29
Add verification scripts for  service-apis

aledbf:
Configuration for https://github.com/kubernetes-sigs/service-apis

xref: kubernetes-sigs/service-apis#2
xref: kubernetes-sigs/service-apis#7
xref: kubernetes-sigs/service-apis#24

PR #15828, opened 2020-01-09
Add service-apis image pushing job

aojea:
xref https://github.com/kubernetes-sigs/service-apis/issues/3

### DIFF
--- a/config/jobs/image-pushing/k8s-staging-service-apis.yaml
+++ b/config/jobs/image-pushing/k8s-staging-service-apis.yaml
@@ -1,0 +1,32 @@
+postsubmits:
+  # this is the github repo we'll build from; this block needs to be repeated for each repo.
+  kubernetes-sigs/k8s-container-image-promoter:
+    - name: service-apis-postsubmit-push-to-staging
+      cluster: test-infra-trusted
+      annotations:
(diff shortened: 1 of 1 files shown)


BenTheElder:
cc @aledbf - IIRC you were also working with this right?

aledbf:
@BenTheElder no, I was working in this https://github.com/kubernetes/test-infra/pull/15699

config/jobs/image-pushing/k8s-staging-service-apis.yaml
```diff
@@ -0,0 +1,32 @@
+postsubmits:
+  # this is the github repo we'll build from; this block needs to be repeated for each repo.
+  kubernetes-sigs/k8s-container-image-promoter:
+    - name: service-apis-postsubmit-push-to-staging
+      cluster: test-infra-trusted
```
fejta: Jobs that run in the trusted cluster need to be defined in the trusted file

aojea: I've followed the instructions from https://github.com/kubernetes/test-infra/blob/master/config/jobs/image-pushing/README.md

How do I add the job to the "trusted" file :sweat_smile: ?

PR #15829, opened 2020-01-09
Revert "stop skipping dynamic PV tests"

BenTheElder:
This reverts commit b282c40fb004aed7ebd27aec8c08428462d58f0a. from https://github.com/kubernetes/test-infra/pull/15826

You can see that kind started failing reliably, plagued by "exit 137" issues with kubectl exec (SIGKILL <> OOM?)

https://testgrid.k8s.io/presubmits-kubernetes-blocking#pull-kubernetes-e2e-kind&width=5&sort-by-flakiness=

I suspect the addition of the Dynamic PV tests must be causing this somehow (?)
If this doesn't clear things up I will also revert running the TCP CLOSE_WAIT test.

There were no changes to kind in this time period and nothing interesting in k/k.
You can however see when we started running these additional tests that we went solid red :/

/cc @spiffxp @aojea

### DIFF
--- a/config/jobs/kubernetes-sigs/kind/kind-presubmits.yaml
+++ b/config/jobs/kubernetes-sigs/kind/kind-presubmits.yaml
@@ -79,7 +79,7 @@ presubmits:
         # NOTE: changes should filter down to pull-kubernetes-e2e-kind-canary
         # and then pull-kubernetes-e2e-kind
         - name: SKIP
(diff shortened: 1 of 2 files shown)


BenTheElder:
> Success rate over time: 3h: 32%, 12h: 64%, 48h: 73%

Screen Shot 2020-01-09 at 1 37 54 AM

Left side is now. the first red from the right in the grouping now is from  `Jan 08 23:19:34Jan 08 2020, 23:19:34 UTC-0800`

We merged at ~ `Jan 08 23:19:34Jan 08 2020, 23:10:00 UTC-800` https://github.com/kubernetes/test-infra/pull/15826#event-2933840949, after which we've only passed once.

BenTheElder:
Will try to root-cause more in the morning.

BenTheElder:
aside: due to conflicts this is not strictly a revert, but 🤷‍♂

PR #15834, opened 2020-01-09
ExternalDNS: trigger build when tags are pushed

linki:
Trigger the pipeline when a git tag is pushed to the repository.

This should allow us to build images like `external-dns:v0.5.18` etc.

(copied from azure provider)

### DIFF
--- a/config/jobs/image-pushing/k8s-staging-external-dns.yaml
+++ b/config/jobs/image-pushing/k8s-staging-external-dns.yaml
@@ -9,8 +9,7 @@ postsubmits:
       decorate: true
       branches:
         - ^master$
-        # this is a regex for semver, from https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string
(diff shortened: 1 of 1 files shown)


config/jobs/image-pushing/k8s-staging-external-dns.yaml
```diff
@@ -9,6 +9,8 @@ postsubmits:
       decorate: true
       branches:
         - ^master$
+        # this is a regex for semver, from https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string
+        - ^v(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(?:-((?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$
```
fejta: Do we need to be so correct? Seems like `v[0-9].*` will work just as well and is far more readable

linki: Not really. I simplified it to your suggestion.

PR #15849, opened 2020-01-10
prow/plugins: removed require-sig

idealhack:
as it's now been replaced by require-matching-label

after https://github.com/kubernetes/test-infra/pull/15827, the second step to address https://github.com/kubernetes/test-infra/pull/9158#issuecomment-415892529


cjwagner:
/hold
This is a change that would break any other Prow instance that still uses the`require-sig` plugin. We can't remove a plugin without a deprecation period during which hook emits warnings if any org or repo enables it.

idealhack:
Sorry, I had that in mind but forgot about it later 🤦‍♂️

I’ll draft a PR next week to add the deprecation announcement and warnings (in logs?).

cjwagner:
> (in logs?)

Yep, consider using this function: https://github.com/kubernetes/test-infra/blob/c58a0eb3660b82caab746e57454b9af8dbd9ad22/prow/plugins/config.go#L635-L654

idealhack:
@cjwagner if we use `warnDeprecated()`, how do we tell if users are using `require-sig` in `plugins/config.go`? I see only `RequireSIG.GroupListURL` but you don't need to set it to enable this plugin: https://github.com/kubernetes/test-infra/blob/a9b921ef5933c6cdef13b3e28054e9695bae26a4/prow/plugins/config.go#L191-L195

or we could use `logrus.Warning()` in the plugin (`plugins/requiresig/requiresig.go`).

cjwagner:
> how do we tell if users are using `require-sig` in `plugins/config.go`? I see only `RequireSIG.GroupListURL` but you don't need to set it to enable this plugin

We can check if any orgs or repos enable the deprecated plugin with this function: https://github.com/kubernetes/test-infra/blob/a9b921ef5933c6cdef13b3e28054e9695bae26a4/prow/plugins/config.go#L795-L814
We probably want to do this check (with `warnDeprecated`) every time we load the plugin config.

PR #15852, opened 2020-01-10
Fix pull-release-test job for k/release

saschagrunert:
I removed the path alias in https://github.com/kubernetes/test-infra/pull/15613/files#r365258628 which may broke our CI completely. I think it's still necessary to have it.

/cc @justaugustus

### DIFF
--- a/config/jobs/kubernetes/release/release-config.yaml
+++ b/config/jobs/kubernetes/release/release-config.yaml
@@ -41,16 +41,17 @@ presubmits:
   - name: pull-release-test
     always_run: true
     decorate: true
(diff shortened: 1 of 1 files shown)


saschagrunert:
> I think reverting everything but the image change from #15613 should fix this.
> 
> k/test-infra/hack/bazel.sh
> https://github.com/kubernetes/test-infra/blob/348fdc86a277d763556543dab2cfc957ff81bd6d/hack/bazel.sh#L22-L27
> 
> I think that coalesce.py script is what's presenting test logs from the bazel run.

Okay thanks, PTAL again.

PR #15854, opened 2020-01-10
Temporarily disable all GKE periodics.

Katharine:
They're all broken anyway.

/cc @michelle192837


Katharine:
/hold

The ones for version branches are missing.

Katharine:
/hold cancel

Apparently we don't fork any periodic GKE jobs? Or I'm missing something.

PR #15861, opened 2020-01-10
Azure: Fix failing master branch jobs

chewong:
/assign @adelina-t 

We should start using versioned aks-engine for release branch jobs and `https://aka.ms/aks-engine/aks-engine-k8s-e2e.tar.gz` for master branch jobs because there might be breaking changes in k8s master that we need to incorporate in aks-engine immediately (example: https://github.com/Azure/aks-engine/pull/2554).

We should also update `https://aka.ms/aks-engine/aks-engine-k8s-e2e.tar.gz` when https://github.com/Azure/aks-engine/pull/2554 is merged since it is causing jobs running on master branch to fail.


chewong:
/hold
In the processing of updating aka.ms/aks-engine/aks-engine-k8s-e2e.tar.gz.
/cc @marosset @PatrickLang

marosset:
I'm working on updating the ms.aka link now

chewong:
@marosset has updated https://aka.ms/aks-engine/aks-engine-k8s-e2e.tar.gz to point to the latest aks-engine tarball.

This PR also fixes test failure in https://github.com/kubernetes/kubernetes/pull/87035.
/assign @justaugustus

PatrickLang:
I do agree with the idea of matching shipped releases of K8s with shipped releases of AKS-Engine. That will better match what someone would use in production vs deploying from aks-engine master. To match production, we should also be sure that the apimodel passed to aks-engine accepts its defaults instead of overriding CNI or Docker versions.

For master branch - I think keeping the link is the right thing to do. As changes to feature flags come in master, we often need to update defaults in aks-engine in order to make sure things work. We also test updated versions of things like metrics-server and coredns in aks-engine master which may need to be updated based on changes for k8s master.

@adelina-t - does this approach make sense to you?

PR #15863, opened 2020-01-10
Set janitor log level to trace

Katharine:
/cc @michelle192837 @ixdy @fejta


Katharine:
/hold

actually probably don't want this right now

Katharine:
Not doing this.

Issue #15806, opened 2020-01-07
Prow components should support using workload identity


fejta:
We should stop sending prow secret.json files when running on GKE. Instead we should use 

https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity#enable_workload_identity_on_an_existing_cluster

fejta:
Enabling workload-identity on the prow service cluster

fejta:
/remove-lifecycle stale

Core prow components all using workload identity... aside from pod utils which is a WIP

Some jobs have been migrated

PR #15870, opened 2020-01-11
Prepare utilities for workload identity

fejta:
Stop requiring `GOOGLE_APPLICATION_CREDENTIALS` to be set and/or passing in explicit credentials.

ref #15806

### DIFF
--- a/boskos/update_prow_config.sh
+++ b/boskos/update_prow_config.sh
@@ -24,20 +24,19 @@ cd "$(git rev-parse --show-toplevel)"
 # Script triggered by prow postsubmit job
 # Update boskos configmap in prow
 
-if [[ -a "${PROW_SERVICE_ACCOUNT:-}" ]] ; then
-	gcloud auth activate-service-account --key-file="${PROW_SERVICE_ACCOUNT}"
+if [[ -a "${PROW_SERVICE_ACCOUNT:-}" ]] ; then # TODO(fejta): delete this in a subsequent PR
+  echo "Use GOOGLE_APPLICATION_CREDENTIALS='$PROW_SERVICE_ACCOUNT' " >&2
+  echo "Migrate away from PROW_SERVICE_ACCOUNT" >&2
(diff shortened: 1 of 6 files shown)


fejta:
Review please!

boskos/update_prow_config.sh
```diff
@@ -24,20 +24,19 @@ cd "$(git rev-parse --show-toplevel)"
 # Script triggered by prow postsubmit job
 # Update boskos configmap in prow
 
-if [[ -a "${PROW_SERVICE_ACCOUNT:-}" ]] ; then
-	gcloud auth activate-service-account --key-file="${PROW_SERVICE_ACCOUNT}"
+if [[ -a "${PROW_SERVICE_ACCOUNT:-}" ]] ; then # TODO(fejta): delete this in a subsequent PR
```
chases2: The docs I'm looking at say that `-a` is "and". Should this be `-n` instead?

I know this is the same as it was before, but I don't know why it was that way, either.

testgrid/cmd/configurator/main.go
```diff
@@ -273,7 +273,11 @@ func main() {
 	var client *storage.Client
 	if strings.HasPrefix(opt.output, "gs://") {
 		var err error
-		client, err = gcs.ClientWithCreds(ctx, opt.creds)
+		var creds []string
+		if opt.creds != "" {
+			creds = append(creds, opt.creds)
```
chases2: It looks like you're appending opt.creds to `""`. What is this snippet doing that's different from how it was before?

boskos/update_prow_config.sh
fejta: `-a` means file exists: http://tldp.org/LDP/Bash-Beginners-Guide/html/sect_07_01.html

testgrid/cmd/configurator/main.go
fejta: Your reading it backwards. If creds are non empty then add send it to ClientWithCreds, otherwise use do not send this argument (which causes us to use default credentials, what we want).

PR #15873, opened 2020-01-12
Remove usage of removed feature gate

odinuge:
The feature gate was removed in: https://github.com/kubernetes/kubernetes/pull/86210

Running with it results in:

```
unrecognized feature gate: PodPriority
```

/sig node


odinuge:
Ref :arrow_up: 
/assign @yujuhong

PR #15878, opened 2020-01-13
Unify presubmit and periodic test for gce alpha features.

oxddr:
Affected jobs: pull-kubernetes-e2e-gce-alpha-features and ci-kubernetes-e2e-gci-gce-alpha-features. I suspect they should be identical.

### DIFF
--- a/config/jobs/kubernetes/sig-cloud-provider/gcp/gcp-gce.yaml
+++ b/config/jobs/kubernetes/sig-cloud-provider/gcp/gcp-gce.yaml
@@ -127,6 +127,8 @@ presubmits:
         - --build=bazel
         - --cluster=
         - --env=KUBE_FEATURE_GATES=AllAlpha=true,CSIMigration=false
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes/sig-cloud-provider/gcp/gcp-gce.yaml
```diff
@@ -127,6 +127,8 @@ presubmits:
         - --build=bazel
         - --cluster=
         - --env=KUBE_FEATURE_GATES=AllAlpha=true,CSIMigration=false
+        - --env=KUBE_PROXY_DAEMONSET=true
```
cheftako: Seems like this should probably be pushed back to the 1.16 branch as well.

oxddr: Do you mean this particular line or whole PR?

PR #15884, opened 2020-01-13
update rules_k8s container

nlopezgi:
New container has Bazel 2.0.0 which is required for several dependency updates


nlopezgi:
fyi @fejta

PR #15761, opened 2020-01-03
Move hack job to python3 and enable PY3 linting

matthyx:
This is basically a retry for https://github.com/kubernetes/test-infra/pull/14015 by @clarketm 
Using the updated rules_python.
Fixes #14654

PR #15885, opened 2020-01-13
Revert "Move hack job to python3 and enable PY3 linting"

michelle192837:
Reverts kubernetes/test-infra#15761

I believe this is causing some test-infra jobs to fail (see https://k8s-testgrid.appspot.com/sig-testing-prow#push-prow and https://k8s-testgrid.appspot.com/sig-testing-maintenance#testgrid-config-upload). Reverting and retrying those jobs.


matthyx:
Not sure why, please let me know once you find out...

michelle192837:
Not 100% sure, but https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/post-test-infra-push-prow/1216775390270328837 has an example of the actual errors.

matthyx:
Missing sysconfig in requirements3.txt

matthyx:
I can have a look after dinner... But how do I run the failing test from my computer?

michelle192837:
Not 100% sure? The upload-testgrid job failed, which mostly runs Configurator. Can test if that fails with the CL and works otherwise.

Issue #15886, opened 2020-01-13
k8s-testimages/bazelbuild is missing pip3


matthyx:
https://github.com/kubernetes/test-infra/pull/15885

In order to get this PR working on the current test infrastructure, we need `pip3` or at least `python3-distutils` installed in that image.

/assign
/cc @fejta 
/cc @michelle192837

matthyx:
Testgrid looks fine...
/close

PR #15889, opened 2020-01-13
Add pip3 to bazelbuild image

matthyx:
Related to #15886 
/assign @fejta 
Then I need to build that image and bump the version used by prow... how?


matthyx:
@fejta I don't see the PR to update `gcr.io/k8s-testimages/bazelbuild:v20190916-ec59af8-0.29.1` in https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/test-infra/test-infra-trusted.yaml
I do see https://github.com/kubernetes/test-infra/pull/15888 but it's unrelated to my change...

PR #15895, opened 2020-01-14
logexporter: add option to specify extra files to dump via flag

oxddr:
/hold
/assign @mm4tt 
ref https://github.com/kubernetes/kubernetes/issues/87143

### DIFF
--- a/logexporter/cmd/main.go
+++ b/logexporter/cmd/main.go
@@ -21,7 +21,6 @@ package main
 
 import (
 	"bytes"
-	"flag"
 	"fmt"
 	"io"
 	"io/ioutil"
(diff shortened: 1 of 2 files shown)


logexporter/cmd/main.go
```diff
@@ -51,7 +51,7 @@ var (
 
 	// Node-type specific logfiles.
 	// Currently we only handle nodes, and neglect master.
-	nodeLogs = []string{"kube-proxy", "node-problem-detector", "fluentd"}
+	nodeLogs = []string{"kube-proxy", "node-problem-detector", "fluentd", "dns-lookup", "ping-client", "ping-server"}
```
mm4tt: I wonder whether we could do it more generically. E.g. collecting everything that starts with cl2-probe? This way we wouldn't have to update this file every time we add a new probe in CL2. WDYT?

logexporter/cmd/main.go
```diff
@@ -51,7 +51,8 @@ var (
 
 	// Node-type specific logfiles.
 	// Currently we only handle nodes, and neglect master.
-	nodeLogs = []string{"kube-proxy", "node-problem-detector", "fluentd"}
+	// NOTE(oxddr): cl2-* is prefix we expect from files created by clusterloader2
+	nodeLogs = []string{"kube-proxy", "node-problem-detector", "fluentd", "cl2-*"}
```
mm4tt: Sorry for being so picking, I definitely don't want to block this, so I'm giving LGTM

But I think it would be better if we introduced a new flag for "extra-logfiles". By doing that we wouldn't have to build a  new version of logexporter every time we want to add new logfiles. 

/lgtm

oxddr: Done.

oxddr:
/hold cancel

/assing @wojtek-t 
Can you have a look as Matt is absent.

logexporter/cmd/main.go
```diff
@@ -44,6 +45,9 @@ var (
 	enableHollowNodeLogs = flag.Bool("enable-hollow-node-logs", false, "Enable uploading hollow node logs too. Relevant only for kubemark nodes")
 	sleepDuration        = flag.Duration("sleep-duration", 60*time.Second, "Duration to sleep before exiting with success. Useful for making pods schedule with hard anti-affinity when run as a job on a k8s cluster")
 	dumpSystemdJournal   = flag.Bool("dump-systemd-journal", false, "Whether to dump the full systemd journal")
+
+	// NOTE(oxddr): cl2-* is prefix we expect from files created by clusterloader2
+	extraLogFiles = pflag.StringSlice("extra-log-files", []string{"cl2-*"}, "Extra log files to dump")
```
wojtek-t: I'm fine with adding a flag, but I don't think we should default it to cl2 - if we really want that, this should be hidden somewhere below when we call logexporter for scalability jobs...

oxddr: Removed. This needs to be set in log-dump script in k/k - I'll send a separate PR.

oxddr:
@mm4tt or @wojtek-t PTAL

logexporter/cmd/main.go
```diff
@@ -32,18 +31,22 @@ import (
...
+	gcsPath              = pflag.String("gcs-path", "", "Path to the GCS directory under which to upload logs, for eg: gs://my-logs-bucket/logs")
+	journalPath          = pflag.String("journal-path", "/var/log/journal", "Path where the systemd journal dir is mounted")
+	cloudProvider        = pflag.String("cloud-provider", "", "Cloud provider for this node (gce/gke/aws/kubemark/..)")
+	gcloudAuthFilePath   = pflag.String("gcloud-auth-file-path", "/etc/service-account/service-account.json", "Path to gcloud service account file, for authenticating gsutil to write to GCS bucket")
+	enableHollowNodeLogs = pflag.Bool("enable-hollow-node-logs", false, "Enable uploading hollow node logs too. Relevant only for kubemark nodes")
+	sleepDuration        = pflag.Duration("sleep-duration", 60*time.Second, "Duration to sleep before exiting with success. Useful for making pods schedule with hard anti-affinity when run as a job on a k8s cluster")
+	dumpSystemdJournal   = pflag.Bool("dump-systemd-journal", false, "Whether to dump the full systemd journal")
+
+	// NOTE(oxddr): cl2-* is prefix we expect from files created by clusterloader2
```
mm4tt: We don't need this here anymore, right? We can add such note in the place where logexporter is invoked

.gitignore
```diff
@@ -24,3 +24,6 @@ pod.yaml
 # Things running grandmatriarch locally might create
 cookies
 secret.yaml
+
+# logexported binary
```
mm4tt: Do we need this? I won't be able to approve changes in this file

wojtek-t:
LGTM once @mm4tt `s comments are addressed

.gitignore
oxddr: No, I'll move this to a separate PR.

oxddr:
/retitle logexporter: add option to specify extra files to dump via flag

PR #15898, opened 2020-01-14
Fix setting up prometheus on kubemark cluster by using newer perf-tests

jkaniuk:
Use 1.16 perf-tests branch in older kubemark tests.

Older perf-tests branches (1.14, 1.15) have some problems with setting up Prometheus stack.
This solution takes less time that debugging all issues on those branches but in the same time it does not invalidate the promise/reason for creation of perf-tests branches - so that changes on perf-tests master would not break presubmits/release breaking tests on old releases.

In general, perf-tests branches should match kubernetes branches, however value of keeping purity of such assignments on old branches is probably not worth time spent on debugging all issues on them.
This assignment should be eventually consistent ;), as new releases would have 1 to 1 mapping to perf-tests, and old branches will be gradually phased out.

Ref: https://github.com/kubernetes/kubernetes/issues/87200

/cc mm4tt

### DIFF
--- a/config/jobs/kubernetes/sig-release/release-branch-jobs/1.14.yaml
+++ b/config/jobs/kubernetes/sig-release/release-branch-jobs/1.14.yaml
@@ -348,7 +348,9 @@ periodics:
     containers:
     - args:
       - --repo=k8s.io/kubernetes=release-1.14
-      - --repo=k8s.io/perf-tests=release-1.14
+      # Use newer perf-tests branch with additional fixes for kubemark
+      # This mismatch is intended only for release branches older than 1.16
+      - --repo=k8s.io/perf-tests=release-1.16
       - --root=/go/src
       - --timeout=140
       - --scenario=kubernetes_e2e
--- a/config/jobs/kubernetes/sig-release/release-branch-jobs/1.15.yaml
+++ b/config/jobs/kubernetes/sig-release/release-branch-jobs/1.15.yaml
@@ -221,7 +221,9 @@ periodics:
     containers:
     - args:
       - --repo=k8s.io/kubernetes=release-1.15
-      - --repo=k8s.io/perf-tests=release-1.15
+      # Use newer perf-tests branch with additional fixes for kubemark
+      # This mismatch is intended only for release branches older than 1.16
+      - --repo=k8s.io/perf-tests=release-1.16
       - --root=/go/src
       - --timeout=140
       - --scenario=kubernetes_e2e


mm4tt:
/assign @justaugustus

Stephen,  do you see a way to add more people from sig-scalability as approvers for these files? Looking at the commit history we update them quite often. 
Could we add sig-scalability-leads as approvers in this directory?

justaugustus:
Hi @mm4tt -- the intent of this directory was so that SIG Release/Release Engineering would have approval over jobs that were purported to be release-blocking or informing. Config changes here should be reviewed/approved by SIG Release.

The reason @wojtek-t is a `config/` OWNER is here: https://github.com/kubernetes/test-infra/pull/11277

I would expect approvers of content for this directory to active in the Release Engineering subproject and would request that in future, PRs be assigned to [OWNERS listed in the directory](https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/sig-release/OWNERS) instead of seeking a merge from a top-level approver.

justaugustus:
FYI @tpepper @calebamiles

mm4tt:
> I'm curious about the reason for using branches that don't match the kubernetes/kubernetes branch for the release they're targeting.
> These are the kind of changes that could/would be undone when we rotate jobs during the release process.
> 
> Is it possible to instead fix the configs of the perf-tests `release-1.16`, `release-1.15` branches, so they work against the correct branches as intended?
> 
> /hold

It is possible, but would require non-negligible amount of work. 
We're in the process of introducing release branches in perf-tests, some of the reasons are to make sure a changes in perf-test/master don't break release branch tests or to make development in perf-test repo easier as we don't have to worry about backward compatibility across releases.

I don't consider it a big issue to have this temporary mismatch here, meaning to have a kubernetes from branch X and perf-test from branch Y where Y is newer than X. It was always like this as before the release branches in perf-test we used to have kubernetes at release X and perf-test was always taken from master. Of course ideally we would have the same perf-test and k8s branches used, but we'll have that from 1.16 going forward. Please not that this is only for 1.14 and 1.15 branches, so the issue you mentioned with rotate job configs during release process wouldn't apply here.

That being said, I think we should add a comment in the config explaining why there is a mismatch. 
@jkaniuk, could you add a comment?

justaugustus:
> I don't consider it a big issue to have this temporary mismatch here

Just a note that nowhere in this PR description or the associated issue does it suggest that the mismatch is temporary, so anyone reading would not have the appropriate context to make that judgment.

> It was always like this as before

But there's a reason we're changing it. I would expect creating a branch `release-x.y` would mean it works with k/k `release-x.y`.

Given that this is only for 1.14 and 1.15, that's reasonable to get this in.

Agreed that we should get a note on each of the changes pointing out that the mismatch is intended and won't continue from 1.16 onwards.

Thanks!

mm4tt:
> > I don't consider it a big issue to have this temporary mismatch here
> 
> Just a note that nowhere in this PR description or the associated issue does it suggest that the mismatch is temporary, so anyone reading would not have the appropriate context to make that judgment.
> 

I agree sorry about that, I knew the context and didn't really look at the PR description. @jkaniuk, can we update the PR description as well?

> > It was always like this as before
> 
> But there's a reason we're changing it. I would expect creating a branch `release-x.y` would mean it works with k/k `release-x.y`.
> 

Yeah, we expected the same but it turned out the be more complicated than we initially thought. We discussed it a lot internally and we don't think there is much gain in fixing that for older branches (other than "code purity", i.e. having the same perf-test version and kubernetes versions) as long as tests are behaving as expected and we, as owners of these tests, can guarantee that they do. Please note that even with the mismatch the change is still serving its purposes, as changes to perf-test master don't affect tests for older releases anymore.


> Given that this is only for 1.14 and 1.15, that's reasonable to get this in.
> 
> Agreed that we should get a note on each of the changes pointing out that the mismatch is intended and won't continue from 1.16 onwards.
> 
Yep, release versions will match going forward. 

> Thanks!

Thanks!

jkaniuk:
Sure, I have extended PR description.

justaugustus:
@jkaniuk -- still need to add descriptions in the code itself

jkaniuk:
Done

justaugustus:
Thanks!
/lgtm
/approve
/hold cancel

PR #15899, opened 2020-01-14
Azure: Add run_if_changed field in job configuration

chewong:
This will automatically run Azure-related presubmit jobs without commenting `/test pull-kubernetes-e2e-...` if any file with a prefix `azure` is modified in the PR.

/assign @feiskyer


chewong:
/assign @justaugustus
due to timezone difference

PR #15901, opened 2020-01-14
Fix generate_tests.py

Katharine:
Fix the mismatch between the generated tests and the output of generate_tests.py:

- Fix a bug that caused duplicate arguments (e.g. multiple `--env`s) in job-specific overrides to only appear once
  - If an argument appears in both a general class and an override, the override still replaces the general class
- Update the location of generate_tests.py in the autobumper
- Regenerate the generated yaml files with the appropriate changes.

For clarity, the first commit contains code changes, the second contains meaningful generated config changes, and the third contains meaningless generated comment noise.

/cc @michelle192837 @justaugustus


justaugustus:
@Katharine -- YOU ARE THE BEST!
/lgtm
/approve
/retest

justaugustus:

PR #15903, opened 2020-01-14
Add test result properties from the junit xml report

tony-yang:
Add test result properties from the junit xml report as target properties under each resultstore's invocation.

### DIFF
--- a/experiment/resultstore/convert.go
+++ b/experiment/resultstore/convert.go
@@ -37,7 +37,7 @@ func convertSuiteMeta(suiteMeta gcs.SuitesMeta) resultstore.Suite {
 		Files: []resultstore.File{
 			{
 				ContentType: "text/xml",
-				ID:          resultstore.UUID(),
+				ID:          path.Base(suiteMeta.Path),
 				URL:         suiteMeta.Path, // ensure the junit.xml file appears in artifacts list
 			},
 		},
@@ -46,6 +46,18 @@ func convertSuiteMeta(suiteMeta gcs.SuitesMeta) resultstore.Suite {
 		child := resultstore.Suite{
 			Name:     suite.Name,
 			Duration: dur(suite.Time),
+		}
+
+		for _, test := range suite.Results {
+			if test.Properties != nil {
+				for _, p := range test.Properties.PropertyList {
+					resultProperty := resultstore.Property{
+						Key:   fmt.Sprintf("%s:%s", test.Name, p.Name),
+						Value: p.Value,
+					}
+					child.Properties = append(child.Properties, resultProperty)
+				}
+			}
 		}
 		switch {
 		case suite.Failures > 0 && suite.Tests >= suite.Failures:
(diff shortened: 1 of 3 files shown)


tony-yang:
> Are there unit tests for this functionality?

For the underlying change on the OSS TestGrid's resultstore Target struct (in a different PR) there is unit test to ensure the property is properly structured. Here, no, since it is just a simple append of whatever key-value pair to the properties field.

experiment/resultstore/convert.go
```diff
@@ -47,6 +47,18 @@ func convertSuiteMeta(suiteMeta gcs.SuitesMeta) resultstore.Suite {
 			Name:     suite.Name,
 			Duration: dur(suite.Time),
 		}
+
+		for _, test := range suite.Results {
+			if test.Properties != nil {
+				for _, p := range test.Properties.PropertyList {
+					resultProperty := resultstore.Property{
+						Key:   fmt.Sprintf("%s %s", test.Name, p.Name),
```
fejta: Why should the test name be part of the key? That seems over specified

Also forced spaces in the key name makes them annoying to search for

tony-yang: This is based on the existing resultstore structure, where all tests fall under the same default target. Property keys might not be unique across tests, so we need to add the test name. I'm happy to discuss and maybe refactor the existing experimental code to make resultstore more structural.

I can definitely change the space to something else. I used space because I was thinking of searching only by property key that would return a list of all tests with the same key when we are not looking for a specific key within a specific test, and a space seems like a natural delimiter here.

stevekuznetsov:
> Here, no, since it is just a simple append of whatever key-value pair to the properties field.

Why don't we have a test that ensures we're sending the correct  client call out to ResultStore?

experiment/resultstore/convert.go
fejta: a) You can add properties on targets, test suites and test cases -- https://github.com/googleapis/googleapis/blob/master/google/devtools/resultstore/v2/test_suite.proto#L135

b) We are not shipping software on a CD and can release immediately. Therefore we should start with the simplest, most obvious design.

We should not preemptively address hypothetical problems (which have undesirable side effects). We should avoid the side effect and wait to see if it is actually a problem, at which point we'll be in a better position to decide how to respond.

tony-yang: Great, I like statement b), exactly what I was trying to do. for a), I'll update the properties to testcase, not sure if it will show up on the resultstore UI, we will find out. Thanks.

tony-yang: After going through the entire resultstore_upload.proto, it appears that we cannot upload a TestCase or a TestSuite with property. This link shows the resources that we can upload: https://github.com/googleapis/googleapis/blob/master/google/devtools/resultstore/v2/resultstore_upload.proto#L1166

At the ResultStore UI level, there is also no way to see properties associated with a test suite or a test case. We can see invocation property, target property, or action property. It seems that if we want to properly divide the properties then dividing an invocation into multiple targets that maps to multiple TestGrid targets is the right abstraction level. Otherwise, all properties (as we defined in TestGrid property) will have to live under the default target property. Then, we will need to use test name to differentiate between duplicated keys.

tony-yang:
> > Here, no, since it is just a simple append of whatever key-value pair to the properties field.
> 
> Why don't we have a test that ensures we're sending the correct client call out to ResultStore?

Thanks Steve. Unit tests are added as suggested.

tony-yang:
Also, this PR is pending on https://github.com/kubernetes/test-infra/pull/15978. PR 15978 fixed dependencies and should be merged first.

tony-yang:
Merged PR 15978 and resolved any conflict. PR ready for review.

cjwagner:
/hold
In case @michelle192837 wants to take a look too.

PR #15904, opened 2020-01-14
Again - move hack job to python3 and enable PY3 linting

matthyx:
This reverts commit ed64df76c2b78ff00229d8520c7b488f315087ca.
We try again to move almost all jobs to python3 now that prow images are updated with pip3.


matthyx:
Thanks @fejta !

PR #15911, opened 2020-01-15
Fix a few bugs in the Boskos GCP janitor

ixdy:
This PR addresses a few issues I discovered in the Boskos GCP janitor.

The main issue is that subcommands fall into a few different categories w.r.t. location flags:
- some operate only at the global level, and thus expect no flags at all
- some operate only at zonal
- some operate only at regional
- some operate at only zonal + regional
- some operate at zonal + global
- some operate at regional + global

The existing logic conflates these categories. I've changed things to make it explicit which flags/options are expected, as well as updating the subcommands to reflect the correct flag values.
I've additionally added support for regional GKE clusters, and fixed the ordering of deletion (networks should be last).

I switched the GCP janitor to use the official cloud SDK image rather than `gcloud-in-go`, since the current version of `gcloud-in-go` used in test-infra is about a year old. I don't feel strongly about this, and would be willing to revert if we update the dependency.

### DIFF
--- a/boskos/janitor/BUILD.bazel
+++ b/boskos/janitor/BUILD.bazel
@@ -42,7 +42,7 @@ pkg_tar(
 
 container_image(
     name = "janitor_image",
-    base = "@gcloud-go//image",
+    base = "@cloud-sdk-slim//image",
     tars = [":janitor_tar"],
 )
 
--- a/boskos/janitor/gcp_janitor.py
+++ b/boskos/janitor/gcp_janitor.py
@@ -34,38 +34,38 @@ DEMOLISH_ORDER = [
 
     # compute resources
     Resource('', 'compute', 'instances', None, 'zone', None, False, True),
+    Resource('', 'compute', 'addresses', None, 'global', None, False, True),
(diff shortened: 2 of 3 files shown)


boskos/janitor/BUILD.bazel
```diff
@@ -42,7 +42,7 @@ pkg_tar(
 
 container_image(
     name = "janitor_image",
-    base = "@gcloud-go//image",
+    base = "@cloud-sdk-slim//image",
```
fejta: How do you know this will work?

ixdy: I've done some manual testing and it appears to work correctly.

ixdy:
Updated to latest gcloud (as I believe there were minor bugfixes included). Please take another look. Thanks!

ixdy:
/reopen

Wrong button. Why do we even have that button?

ixdy:
I've done some manual dry-run testing locally and everything seems to be working as far as I can tell.

/hold cancel

Assuming this merges today, I'll watch the rollout tomorrow.

PR #15919, opened 2020-01-15
handle invalid assign/review comment

### LINKED ISSUE #15811: Bad `/assign` causes "Error handling GenericCommentEvent" instead of user feedback
We saw someones `/assign <someone> please take a look` command to log an error instead of replying to the user that "please, take, a, look" could not be assigned. It used to do this in the past.

Log:
```
{
  component: "hook"   
  error: "status code 404 not one of [201], body: {"message":"Not Found","documentation_url":"https://developer.github.com/v3/issues/assignees/#add-assignees-to-an-issue"}"   
  event-type: "issue_comment"   
  level": "error"   
  msg: "Error handling GenericCommentEvent."   
  plugin: "assign"   
 }
```

See: https://github.com/openshift/console/pull/3765#issuecomment-571508312

xref: https://github.com/openshift/release/issues/6581

adshmh:
Fixes #15811

### DIFF
--- a/prow/github/client.go
+++ b/prow/github/client.go
@@ -2129,7 +2129,11 @@ func (c *client) AssignIssue(org, repo string, number int, logins []string) erro
 		method:      http.MethodPost,
 		path:        fmt.Sprintf("/repos/%s/%s/issues/%d/assignees", org, repo, number),
 		requestBody: map[string][]string{"assignees": logins},
-		exitCodes:   []int{201},
+		// GitHub API ignores users who can not be assigned due to permission,
+		// etc., but returns 404 if even one of the specified users does not exist.
+		// The response body from 404 does not list the invalid users,
+		// so we treat 404 as all specified users being invalid.
+		exitCodes: []int{201, 404},
 	}, &i)
 	if err != nil {
 		return err
--- a/prow/github/client_test.go
+++ b/prow/github/client_test.go
@@ -680,7 +680,6 @@ func TestAssignIssue(t *testing.T) {
 			if ps["assignees"][0] != "george" || ps["assignees"][1] != "jungle" {
 				t.Errorf("Wrong assignees: %v", ps)
 			}
-
 		} else {
 			t.Errorf("Wrong assignees length: %v", ps)
 		}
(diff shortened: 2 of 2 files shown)


prow/plugins/assign/assign.go
```diff
@@ -208,15 +240,122 @@ func newReviewHandler(e github.GenericCommentEvent, gc githubClient, log *logrus
...
+		return sets.String{}, err
+	}
+	comments, err := gc.ListIssueComments(org, repo, number)
+	if err != nil {
+		return sets.String{}, err
+	}
+
+	allAssignees := sets.NewString(logins(assignees)...)
+	allAssignees.Insert(github.NormLogin(author))
+	allAssignees.Insert(commenters(comments)...)
```
petr-muller: Huh, this is strange - I would have thought that the [API under `ListAssignees`](https://developer.github.com/v3/issues/assignees/#list-assignees) already includes the author and commenters in the set. It doesn't? Maybe worth a comment...

adshmh:
/retest

Failure does not seem related (perhaps a flake?).

prow/plugins/assign/assign.go
```diff
@@ -145,6 +150,22 @@ func handle(h *handler) error {
 	}
 	if len(toAdd) > 0 {
 		h.log.Printf("Adding %s to %s/%s#%d: %v", h.userType, org, repo, e.Number, toAdd)
+		invalidLogins, err := h.invalidLogins(org, repo, e.Number, e.IssueAuthor.Login, toAdd)
+		if err != nil {
+			return err
+		}
+		if len(invalidLogins) > 0 {
+			h.log.Printf("Invalid users %v requested to be added to %s/%s#%d", invalidLogins, org, repo, e.Number)
```
stevekuznetsov: nit: why do we use `Printf` in this handler? How does that even get routed? If this is application logging, `logrus.Debug`?

prow/plugins/assign/assign.go
adshmh: Thank you for the review. Based on a discussion with @stevekuznetsov I updated the code to handle 404 from GitHub API, rather than using tokens to build a list of valid assignees.

prow/plugins/assign/assign.go
adshmh: Thank you for the review. As discussed on slack, I have updated the github client to handle 404 (which is how GitHub API responds to one or more non-existent logins in the assign issue request).

PR #15924, opened 2020-01-15
add pull-cip-auditor-e2e test

listx:
This test will invoke the e2e tests for the auditing mechanism built
into the promoter.

Currently the e2e tests for the auditing mechanism do nothing (see https://github.com/kubernetes-sigs/k8s-container-image-promoter/pull/169), but soon they will be populated with real tests.

@thockin The secret key for `k8s-infra-gcr-promoter@k8s-gcr-audit-test-prod.iam.gserviceaccount.com.json` will be used by these tests and must be given to the Prow admins to add as a kubernetes secret.

### DIFF
--- a/config/jobs/kubernetes/sig-release/cip/container-image-promoter.yaml
+++ b/config/jobs/kubernetes/sig-release/cip/container-image-promoter.yaml
@@ -87,21 +87,11 @@ presubmits:
     branches:
     - ^master$
     spec:
+      serviceAccountName: k8s-gcr-audit-test-prod
       containers:
       - image: gcr.io/k8s-testimages/kubekins-e2e:v20200114-6d2c483-master
         command:
         - "./test-e2e/cip-auditor/entrypoint-from-container.sh"
-        env:
-        - name: CIP_E2E_KEY_FILE
-          value: "/etc/k8s-gcr-audit-test-prod-service-account/service-account.json"
-        volumeMounts:
-        - name: k8s-gcr-audit-test-prod-service-account-creds
-          mountPath: /etc/k8s-gcr-audit-test-prod-service-account
-          readOnly: true
-      volumes:
-      - name: k8s-gcr-audit-test-prod-service-account-creds
-        secret:
-          secretName: k8s-gcr-audit-test-prod-service-account
     annotations:
       testgrid-dashboards: sig-release-misc
   # Build all binaries and also run unit tests.
--- a/config/jobs/kubernetes/sig-release/cip/container-image-promoter.yaml
+++ b/config/jobs/kubernetes/sig-release/cip/container-image-promoter.yaml
@@ -76,6 +76,34 @@ presubmits:
           secretName: k8s-cip-test-prod-service-account
     annotations:
       testgrid-dashboards: sig-release-misc
+  - name: pull-cip-auditor-e2e
+    decorate: true
+    path_alias: "sigs.k8s.io/k8s-container-image-promoter"
+    skip_report: false
+    always_run: true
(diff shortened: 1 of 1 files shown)


listx:
Pinging test-infra oncall @fejta

config/jobs/kubernetes/sig-release/cip/container-image-promoter.yaml
```diff
@@ -76,6 +76,34 @@ presubmits:
...
+        - name: CIP_E2E_KEY_FILE
+          value: "/etc/k8s-gcr-audit-test-prod-service-account/service-account.json"
+        volumeMounts:
+        - name: k8s-gcr-audit-test-prod-service-account-creds
+          mountPath: /etc/k8s-gcr-audit-test-prod-service-account
+          readOnly: true
+      volumes:
+      - name: k8s-gcr-audit-test-prod-service-account-creds
+        secret:
+          secretName: k8s-gcr-audit-test-prod-service-account
```
fejta: Can we use workload identity, rather than secrets that can be stolen?

https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity

listx: I'm sure we can, but (1) it doesn't really matter a ton if the secret is stolen because it only has powers to mess around in the test project (filled with meaningless test-only data) and (2) can we still not block on this?

fejta: 1) Merge https://github.com/kubernetes/test-infra/pull/15984
2) Delete the volumes, volume mounts, env 
3) Add `serviceAccountName: k8s-gcr-audit-test-prod` to the pod spec (aka between L89 and L90)

Everything at this point is declarative and self-managed. AKA the owner of the service account will have to run 

```
gcloud iam service-accounts add-iam-policy-binding \
  --role roles/iam.workloadIdentityUser \
  --member "serviceAccount:[PROJECT_ID].svc.id.goog[[K8S_NAMESPACE]/[KSA_NAME]]" \
  [GSA_NAME]@[PROJECT_ID].iam.gserviceaccount.com
```

From the above docs, and #15984 will have to merge, but once that happens `gcloud auth list` inside the pod will show that things authenticate to GCP as `[GSA_NAME]@[PROJECT_ID].iam.gserviceaccount.com`

Alternatively, yes, we can definitely do the service account thing. Can you hand that over to @Katharine tomorrow if you want to go that route?

listx: #15984 looks very promising and since we got this far, I would like to try it out (the config example looks almost too easy/good to be true...!). I was going to ask how do I figure out the PROJECT_ID, K8S_NAMESPACE/KSA_NAME but it looks like you already answered those questions by adding the initial config for the GSA I wanted to add in this PR (https://github.com/kubernetes/test-infra/pull/15984/files#diff-2ee02ef3182b8522bfa83a6c7a2d6978R6-R8) --- bravo!

Anyway I'll edit this PR as you instructed. Brb

listx:
@fejta To confirm, from my side (to empower the KSA to act as GSA), what would the correct invocation be? I can't remember what GCP project Prow runs on.

Following your docs, this is my guess:

```
gcloud iam service-accounts add-iam-policy-binding \
  --role roles/iam.workloadIdentityUser \
    --member "serviceAccount:[PROJECT_ID].svc.id.googtest-pods/k8s-gcr-audit-test-prod" \
      k8s-infra-gcr-promoter@k8s-gcr-audit-test-prod.iam.gserviceaccount.com
```

wdyt?

PR #15947, opened 2020-01-17
Switched aws-encryption-provider test image to golang:1.13

micahhausler:
We need `make` to build kubernetes-sigs/aws-encryption-provider


rajatjindal:
thanks for taking care of it @micahhausler

PR #15948, opened 2020-01-17
Added Gyuho Lee as owner on aws-encryption-provider

micahhausler:
cc @gyuho


gyuho:
/lgtm
/approve

Thanks all!

PR #15951, opened 2020-01-17
Temporary janitor zone hack.

Katharine:
Try to get the janitors running again.

/cc @fejta @cjwagner @ixdy


boskos/janitor/gcp_janitor.py
```diff
@@ -14,6 +14,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# I am not dealing with this today.
+# pylint: disable=line-too-long
```
cjwagner: 😆 👍

PR #15955, opened 2020-01-18
Use bazel 2.0.0

fejta:


alvaroaleman:
This seems to have broken the boskos push job: https://github.com/kubernetes/test-infra/issues/16245

It uses the `gcr.io/k8s-testimages/bazelbuild:v20200205-602500d-0.29.1` image, which presumably is the reason https://github.com/kubernetes/test-infra/pull/15956 didn't change it. Not sure what the v2 equivalent is, there seems to be no `gcr.io/k8s-testimages/bazelbuild:v20200205-602500d-2.0.0`

fejta:
Once post-test-infra-push-bazelbuild runs then these images will exist

fejta:
`gcr.io/k8s-testimages/bazelbuild:v20200212-b304d89-2.0.0` and `gcr.io/k8s-testimages/bazelbuild:v20200212-b304d89-2.1.0` now exist

PR #15956, opened 2020-01-18
Update test-infra jobs to bazel 2.0.0

fejta:


chases2:
I think this should wait until #15955 merges. Otherwise, I don't have any faith that `post-test-infra-bazel` will pass regularly.
/hold

fejta:
/hold cancel
Its the other way around. This needs to merge before that one will pass.

oxddr:
Version 2.1.0 has been release [few days ago](https://github.com/bazelbuild/bazel/releases). Shall we switch to this one instead of 2.0.0?

fejta:
I would rather first do this as its tested and will work with other things

PR #15958, opened 2020-01-20
Grant release-managers rerun privs on kubernetes-build and releng jobs

justaugustus:
It was suggested on Slack that it might be a good idea for @kubernetes/release-managers to have access to rerun jobs, without having to poke test-infra on-call.

ref: https://kubernetes.slack.com/archives/CJH2GBF7Y/p1579303883016000

/assign @Katharine @fejta 
cc: @tpepper @kubernetes/release-engineering

### DIFF
--- a/config/jobs/image-pushing/k8s-staging-release-test.yaml
+++ b/config/jobs/image-pushing/k8s-staging-release-test.yaml
@@ -33,6 +33,9 @@ periodics:
       - name: creds
         secret:
           secretName: deployer-service-account
+  rerun_auth_config:
+    github_team_ids:
+      - 2241179 # release-managers
   annotations:
     # TODO(justaugustus): Enable forking for prototype GCB-based build jobs.
     testgrid-dashboards: sig-release-prototype-master-blocking
@@ -73,6 +76,9 @@ periodics:
       - name: creds
         secret:
           secretName: deployer-service-account
+  rerun_auth_config:
+    github_team_ids:
+      - 2241179 # release-managers
   annotations:
     testgrid-dashboards: sig-release-prototype-master-blocking
     testgrid-tab-name: build-master-fast
@@ -112,6 +118,9 @@ postsubmits:
           - name: creds
             secret:
               secretName: deployer-service-account
+      rerun_auth_config:
+        github_team_ids:
+          - 2241179 # release-managers
     - name: post-release-push-image-releng-ci-bazel
       cluster: test-infra-trusted
       annotations:
@@ -143,6 +152,9 @@ postsubmits:
           - name: creds
             secret:
               secretName: deployer-service-account
+      rerun_auth_config:
+        github_team_ids:
+          - 2241179 # release-managers
     - name: post-release-push-image-kubepkg
       cluster: test-infra-trusted
       annotations:
@@ -174,3 +186,6 @@ postsubmits:
           - name: creds
             secret:
               secretName: deployer-service-account
+        rerun_auth_config:
+          github_team_ids:
+            - 2241179 # release-managers
--- a/config/jobs/kubernetes/release/release-config.yaml
+++ b/config/jobs/kubernetes/release/release-config.yaml
@@ -81,6 +81,9 @@ periodics:
       command:
       - make
       - verify-published-debs
+  rerun_auth_config:
+    github_team_ids:
+      - 2241179 # release-managers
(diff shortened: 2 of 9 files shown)


cblecker:
/hold

I know there are concerns that rerunning certain jobs can cause even bigger CI problems. We should ensure we have clear sign off and guidelines before we open this outside of on-call.

justaugustus:
SGTM!

cblecker:
@fejta @Katharine: Is it possible to speak to the specific jobs that shouldn't be touched, or we have concerns with? I think it was the scale jobs, correct?

justaugustus:
FWIW, the jobs that releng will care most about are `ci-kubernetes-(bazel-)?build*` jobs + anything else under our OWNERS purview.

justaugustus:
> I know there are concerns that rerunning certain jobs can cause even bigger CI problems. We should ensure we have clear sign off and guidelines before we open this outside of on-call.

@fejta -- Thoughts on @cblecker's comment? I want to make sure we're in the clear, before moving on anything here.

justaugustus:
Where did we land on this one?
Can we unhold?

cblecker:
I know retesting certain jobs has has caused problems before (in particular scalability). I'd like to have a policy written down somewhere so that release managers know which jobs are okay and which ones aren't.

We can either do that by saying "these jobs are dangerous and don't touch them" which would have to come from the current on-callers (@fejta, @Katharine, etc). Or the release managers could have a written policy/SOP saying "We have access to all jobs, but these are the only ones we will touch".

But I'd like something written so we don't have folks stepping on each others toes.

stevekuznetsov:
Instead of writing policy, write code! Have scalability jobs request leases from `boskos` before they execute to programatically limit their concurrency and render an errant re-test benign.

justaugustus:
> Instead of writing policy, write code!

Heh. I think we'll need both.

From the Release Engineering side, the jobs I care about most are:
- ones that produce version markers (`ci-kubernetes[-bazel]-build*`), so we can rerun in emergencies
- `test-infra-push-kubekins-e2e` (during branch creation)

I can write a quick releng SOP for that, but I'd like to verify with test-infra that those jobs are reasonable before "putting pen to paper".

Katharine:
> `test-infra-push-kubekins-e2e` (during branch creation)

Why do you need to manually run this? It should run whenever the directory changes, which would include when adding a new image variant.

> ones that produce version markers (ci-kubernetes[-bazel]-build*), so we can rerun in emergencies

Seems fine to me — but if that's all you want, we can add the ability to rerun to those specific jobs instead of all possible jobs.

fejta:
Scalability jobs are the ones that should cause problems.  If the number of jobs to rerun are small -- most likely just the build jobs -- then I agree with just configuring those jobs to have the extra team.

justaugustus:
> > `test-infra-push-kubekins-e2e` (during branch creation)
> 
> Why do you need to manually run this? It should run whenever the directory changes, which would include when adding a new image variant.

@Katharine -- I guess we don't need to? I recall some time a few cycles ago needing to have it run by t-i-oncall.

> Seems fine to me — but if that's all you want, we can add the ability to rerun to those specific jobs instead of all possible jobs.

> Scalability jobs are the ones that should cause problems. If the number of jobs to rerun are small -- most likely just the build jobs -- then I agree with just configuring those jobs to have the extra team.

Is that something I can fix up in this PR? If so, how?
Or is it a feature request? :sweat_smile:

fejta:
Configure the jobs you want to have the rerun config you want, here's a job that does this:
https://github.com/kubernetes/test-infra/blob/a0a273a08a4c35e150c09fd8a7f558d4cc12c793/config/jobs/kubernetes-sigs/vsphere-csi-driver/vsphere-csi-driver.yaml#L201-L203

justaugustus:
@fejta -- Is this usage documented outside of code?

@Katharine -- Do we know if this rerun config will be clobbered during forking/rotating?

fejta:
Looks like no? I was surprised this feature already existed when I asked about it :) https://github.com/kubernetes/test-infra/search?l=Markdown&q=rerun_auth_config

justaugustus:
/hold cancel
(per @fejta: https://kubernetes.slack.com/archives/C2C40FMNF/p1581721182215500?thread_ts=1581721024.214900&cid=C2C40FMNF)

PR #15961, opened 2020-01-20
checkconfig: Use strict Unmarshal for unknown fields

### LINKED ISSUE #15316: checkconfig returns error for valid config
**What happened**:
`checkconfig` returns an error for a valid config. More specifically, `checkUnknownFields(..)` returns *unknown* fields that are, in fact, *known*.
 
https://github.com/kubernetes/test-infra/blob/c730fab62edb4442547269b17e314b6dd861e925/prow/cmd/checkconfig/main.go#L402-L431

A valid, functional configuration should be considered valid by `checkconfig`.

1. Add a legacy [`slack_reporter`](https://github.com/istio/test-infra/blob/104369d76e8affc99f793cb6315f046ac22a2e7f/prow/config.yaml#L61-L71) field to your Prow configuration.
2. Run `checkconfig` with `--strict` and/or `--warnings=unknown-fields` warning enabled.

[example](https://storage.googleapis.com/istio-prow/pr-logs/pull/istio_test-infra/2102/pull-test-infra-prow-checkconfig/3124/build-log.txt)

afirth:
- Remove custom reflection-based checks for unknown fields and use strict loading from the upstream library instead.
- Fix panic with error message checking in tests by diffing `foo.Error()` instead of `foo` @ main_test.go:540
- Replace legacyDiff with cmp
- Split a test case which used to check multiple failures to ensure they are all covered

/fixes #15316
/label "tide/merge-squash"

### DIFF
--- a/prow/cmd/checkconfig/main_test.go
+++ b/prow/cmd/checkconfig/main_test.go
@@ -25,6 +25,7 @@ import (
 	"regexp"
 	"testing"
 
+	"github.com/google/go-cmp/cmp"
 	"k8s.io/apimachinery/pkg/util/diff"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"sigs.k8s.io/yaml"
@@ -527,7 +528,8 @@ size:
 			got := validateUnknownFields(tc.cfg, tc.configBytes, tc.filename)
 			if !reflect.DeepEqual(got, tc.expectedErr) {
 				t.Errorf("%s: did not get expected validation error:\n%v", tc.name,
-					diff.ObjectGoPrintDiff(tc.expectedErr, got))
+					cmp.Diff(tc.expectedErr.Error(), got.Error()))
+				// cmp.Diff(tc.expectedErr.Error(), got.Error(), cmp.AllowUnexported()))
 			}
 		})
 	}
--- a/prow/cmd/checkconfig/main_test.go
+++ b/prow/cmd/checkconfig/main_test.go
@@ -23,6 +23,7 @@ import (
 	"os"
 	"reflect"
 	"regexp"
+	"strings"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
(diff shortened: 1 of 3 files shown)


prow/cmd/checkconfig/main_test.go
```diff
@@ -527,12 +537,18 @@ size:
 			got := validateUnknownFields(tc.cfg, tc.configBytes, tc.filename)
 			if !reflect.DeepEqual(got, tc.expectedErr) {
```
cjwagner: This is fragile to changes in the YAML library's error formatting.
Instead of trying to match the errors with `DeepEqual` (which requires identical types and error message content), consider checking that the error message text contains some key words that are more likely to remain if the YAML library changes it's error formatting. e.g. match if the error message contains "unknown field" and the unknown field name itself. It may also be appropriate to simply check that an error is returned and use happy path test cases to increase confidence that we aren't getting errors for some other reason.

afirth: Thanks for the feedback Cole. I'll implement that.
/hold

afirth: @cjwagner I took a stab at this but only succeeded in making the test uglier and more fragile when matching the three expected substrings and also catering for the `nil` case. I guess the yaml library error is pretty stable, but if not the change would be contained entirely in https://github.com/kubernetes/test-infra/pull/15961/files#diff-1bccbd8d60a651295f7c8bb39ac87d4dR548 and 549.
/unhold

cjwagner: Since this prints the error diff if they aren't equal it shouldn't be too burdensome to fix this test when upgrading the YAML library dependency in the rare case that this changes. That being said, I don't really see why using an equality check function that checks for substrings and handles the nil case would be uglier or more fragile than using the equality operator. If you think you can do that cleanly, that would be preferable since it makes fewer assumptions about dependency implementation, but if you just want to merge this use `/hold cancel`.

afirth: you're totally right, this was a good exercise for me. Pushed in 7fc20d3, would appreciate another set of eyes/lgtm and then you could unhold

afirth:
@cjwagner you're totally right, this was a good exercise for me. Pushed in 7fc20d3, would appreciate another set of eyes/lgtm
/unhold

cjwagner:
@afirth Thanks for updating, looks great!

PR #15962, opened 2020-01-20
Allow connections to GKE nodes on custom ports

jkaniuk:
Several integration tests require master to connect to pods on custom ports.

For GKE clusters with private nodes, GKE master needs additional firewall rules in order to connect to pods. Hard-coded list of predefined ports helps with setting up e.g. Prometheus:
`e2eAllow      = "tcp:22,tcp:80,tcp:8080,tcp:9090,tcp:30000-32767,udp:30000-32767"`

This change specifies GKE flag opening additional, user defined ports on nodes.
That way, without changes in test-infra repository, new integration tests could run on private GKE clusters even when custom ports are needed.

/cc mm4tt krzyzacy

### DIFF
--- a/kubetest/gke.go
+++ b/kubetest/gke.go
@@ -54,6 +54,7 @@ var (
 	gkeCustomSubnet                = flag.String("gke-custom-subnet", "", "(gke only) if specified, we create a custom subnet with the specified options and use it for the gke cluster. The format should be '<subnet-name> --region=<subnet-gcp-region> --range=<subnet-cidr> <any other optional params>'.")
 	gkeReleaseChannel              = flag.String("gke-release-channel", "", "(gke only) if specified, bring up GKE clusters from that release channel.")
 	gkeSingleZoneNodeInstanceGroup = flag.Bool("gke-single-zone-node-instance-group", true, "(gke only) Add instance groups from a single zone to the NODE_INSTANCE_GROUP env variable.")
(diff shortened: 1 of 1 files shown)


kubetest/gke.go
```diff
@@ -54,6 +54,7 @@ var (
 	gkeCustomSubnet                = flag.String("gke-custom-subnet", "", "(gke only) if specified, we create a custom subnet with the specified options and use it for the gke cluster. The format should be '<subnet-name> --region=<subnet-gcp-region> --range=<subnet-cidr> <any other optional params>'.")
 	gkeReleaseChannel              = flag.String("gke-release-channel", "", "(gke only) if specified, bring up GKE clusters from that release channel.")
 	gkeSingleZoneNodeInstanceGroup = flag.Bool("gke-single-zone-node-instance-group", true, "(gke only) Add instance groups from a single zone to the NODE_INSTANCE_GROUP env variable.")
+	gkeNodePorts                   = flag.String("gke-node-ports", "", "(gke only) List of ports for which custom firewall rule should be created allowing master to connect to nodes. The format should be 'protocol[:port[-port]],[...]' as in gcloud compute firewall-rules create --allow.")
```
mm4tt: Correct me if I'm wrong, but this is not fully true. In fact it will create a firewall rule that will open these ports on all nodes to all incoming traffic, not only from masters, right? Please fix

jkaniuk: Right, comment is wrong.
For the rule that allows only master, we would need separate way of passing master CIDR block, so that gets complicated.
I would go with current approach (same as with Prometheus), fixing the comment. WDYT?

PR #15978, opened 2020-01-21
Update the repo dependency on OSS testgrid

tony-yang:
Update the repo dependency on OSS testgrid from alpha.3 to alpha.4, resolve any deprecated field.


michelle192837:
This looks good to me, but I'm not super-confident on the Prow side of things, so just approving.

/approve

tony-yang:
> This looks good to me, but I'm not super-confident on the Prow side of things, so just approving.
> 
> /approve

Thank you Michelle, adding Cole to help with the Prow side.
/assign @cjwagner

prow/initupload/run.go
```diff
@@ -61,6 +54,13 @@ func specToStarted(spec *downwardapi.JobSpec, cloneRecords []clone.Record) gcs.S
 		started.Repos[ref.Org+"/"+ref.Repo] = ref.String()
 	}
 
+	if mainRefs := spec.MainRefs(); mainRefs != nil {
+		started.Repos["RepoVersion"] = shaForRefs(*mainRefs, cloneRecords)
+		if started.Repos["RepoVersion"] == "" {
```
cjwagner: This should be outside the `mainRefs` conditional to match the previous behavior. Currently this will always be false.

prow/initupload/run.go
```diff
@@ -61,6 +54,13 @@ func specToStarted(spec *downwardapi.JobSpec, cloneRecords []clone.Record) gcs.S
 		started.Repos[ref.Org+"/"+ref.Repo] = ref.String()
 	}
 
+	if mainRefs := spec.MainRefs(); mainRefs != nil {
+		started.Repos["RepoVersion"] = shaForRefs(*mainRefs, cloneRecords)
```
cjwagner: Does setting this field in the `repos` map have the same behavior as setting the top level `repo-version` field?

prow/sidecar/run.go
```diff
@@ -190,9 +190,6 @@ func (o Options) doUpload(spec *downwardapi.JobSpec, passed, aborted bool, metad
 		// TODO(fejta): JobVersion,
 	}
 
-	// TODO(fejta): move to initupload and Started.Repos, RepoVersion
-	finished.Revision = downwardapi.GetRevisionFromSpec(spec)
```
cjwagner: Use `finished.DeprecatedRevision` like you did for started.json.

prow/initupload/run.go
tony-yang: Based on the comment from the TestGrid repo, it should be: https://github.com/GoogleCloudPlatform/testgrid/blob/master/metadata/job.go#L34

Also, all references to the original RepoVersion is updated to use the Repos map, which is only in the run.go according to grep.

prow/initupload/run.go
tony-yang: Thanks Cole.

prow/sidecar/run.go
tony-yang: Added, thanks Cole

prow/initupload/run.go
cjwagner: That comment does not seem to suggest that, did you mean to link to something else? The comment says `Repos holds the RepoVersion of all commits checked out.` and the next line says `{repo: branch_or_pull} map`. The map keys are expected to be in `org/repo` format and AFAICT there is no special handling of a `RepoVersion` key.

Is simply replacing `started.RepoVersion` with `started.DeprecatedRepoVersion` insufficient for some reason?

> all references to the original RepoVersion is updated to use the Repos map

I see that you've made that change here, but that does not appear to be supported by TestGrid.

tony-yang: Switching to DeprecatedRepoVersion might be sufficient for now (based on grep of the DeprecatedRepoVersion, which seems to only live in the metadata/job.go). I think I'll update it to set DeprecatedRepoVersion than switch to the map for now.

Based on the comments, my understanding is that fejta@ is going to remove DeprecatedRepoVersion in the future while the original RepoVersion should go into the Repos{} map (based on the first comment above the Repos field). I don't know why there's two comments for the same Repos field. But we can revisit this when that happens. Thanks.

cjwagner: IIUC `RepoVersion` belonging in the `Repos` map doesn't mean literally adding a `RepoVersion` key to the map. The map is `org/repo` -> `<value of RepoVersion for that repo>`. This change was made because multiple repos can be checked out so we need a value per repo. I agree that the documentation is unclear, especially because `RepoVersion` is unexplained.

tony-yang: Thanks Cole

PR #15979, opened 2020-01-21
Improve crier started.json behaviour.

Katharine:
Currently, crier is reporting jobs to GCS as soon as they are triggered. The jobs do not actually have a BuildID at this point, so they're being reported to the wrong place. Change the GCS reporter to only consider jobs that actually have a build ID.

Additionally, change started.json reporting to occur in all states, in case we miss the good triggered state for any reason (e.g. crier being temporarily unavailable). This helps to ensure that there are always started.json files.

The actual change here is tiny; it's mostly unit test changes.

### DIFF
--- a/prow/gcs/reporter/reporter_test.go
+++ b/prow/gcs/reporter/reporter_test.go
@@ -102,8 +102,9 @@ func TestGetJobDestination(t *testing.T) {
 		prowjobGcsConfig  *prowv1.GCSConfiguration
 		prowjobType       prowv1.ProwJobType
 		prowjobRefs       *prowv1.Refs
+		buildID           string
 		expectBucket      string
-		expectDir         string // tip: this will always end in "my-little-job/123"
+		expectDir         string // tip: this will always end in "my-little-job/[buildID]"
 		expectErr         bool
 	}{
 		{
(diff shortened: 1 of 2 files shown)


prow/gcs/reporter/reporter_test.go
```diff
@@ -255,40 +255,28 @@ func (ta *testAuthor) NewWriter(ctx context.Context, bucket, path string, overwr
...
-		expectedFile   string
 		completionTime *metav1.Time
 		passed         bool
 	}{
-		{
-			jobState:     prowv1.TriggeredState,
-			expectedFile: "started.json",
-		},
-		{
-			jobState: prowv1.PendingState,
```
cjwagner: We should keep these test cases to ensure we don't report `finished.json` before the job completes.

stevekuznetsov:
Changed LGTM modulo Cole's nits

prow/gcs/reporter/reporter_test.go
Katharine: The function being tested here doesn't actually check check that (it used to, but it was removed for being redundant). It's handled by `reportJobState`, which is now annoying to unit test because I have to rejig my harness to support varying numbers of files being written. Ah well...

Katharine: As it happens I have added the check back anyway because apparently if you try reporting an incomplete job it panics.

cjwagner: Yeah we should have the function that reads the completion time return a bool or error to indicate if the job is complete instead of having a separate `Complete()` function. That would make it harder to forget to check `Complete()` before reading the completion time.
/shrug

PR #15980, opened 2020-01-21
Bump pull-kubernetes-e2e-gce mem req to 8Gi

spiffxp:
This is an experiment to address https://github.com/kubernetes/kubernetes/issues/87441

My working theory is that bazel's jvm is getting oomkilled. If this reduces Stage flakes on this job, there are others that could benefit from a bump.

OTOH maybe this requires an adjustment to our build cluster and needs discussion first

### DIFF
--- a/config/jobs/kubernetes/sig-cloud-provider/gcp/gcp-gce.yaml
+++ b/config/jobs/kubernetes/sig-cloud-provider/gcp/gcp-gce.yaml
@@ -58,7 +58,7 @@ presubmits:
         image: gcr.io/k8s-testimages/kubekins-e2e:v20200118-eaed6c6-master
         resources:
           requests:
-            memory: "6Gi"
+            memory: "8Gi"
 
   - name: pull-kubernetes-e2e-gce-rbe
     always_run: true


spiffxp:
/uncc @chases2 @andrewsykim 
/cc @Katharine @fejta @BenTheElder 
if you have any better ideas I'm all ears

BenTheElder:
> My working theory is that bazel's jvm is getting oomkilled. If this reduces Stage flakes on this job, there are others that could benefit from a bump.

memory requests are just requests.

my theory is that making this non-privileged broke our ability to free memory and that prow is broadly wasting RAM (see kubetest internals, search for `sync`), which we do because these are long lived jobs that only need lots of memory during the initial build. docker won't release memory on it's own so the e2e presubmit guzzle memory.

the periodics don't build so are fine.

BenTheElder:
```
W0121 22:25:24.710] 2020/01/21 22:25:24 util.go:265: Flushing memory.
I0121 22:25:24.810] make: Leaving directory '/go/src/k8s.io/kubernetes'
W0121 22:25:34.262] 2020/01/21 22:25:34 util.go:275: flushMem error (page cache): exit status 1
```
https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/80724/pull-kubernetes-e2e-gce/1219745430556905472/build-log.txt

(aside: these log files have gotten huge and are painful to view / search now :/)

BenTheElder:
another option is probably just to switch over to the RBE jobs which we are running in duplicate, that hopefully offloads some ram to the RBE cluster

spiffxp:
>another option is probably just to switch over to the RBE jobs which we are running in duplicate, that hopefully offloads some ram to the RBE cluster

Sadly the RBE jobs are behaving basically the same from what I can tell. That said it's not clear to me what they're getting in terms of memory.

BenTheElder:
/lgtm
I think this is a good idea anyhow (increase the memory requests => avoid scheduling as many of these to the same box) but I don't think this hits the root of the problem where we consume tons of memory in a container and then run something far less intensive for an hour or more.

the kubetest hack sidesteps this, possibly better would be something with pipelines

Issue #15557, opened 2019-12-10
boskos: ranch: /acquire /update /reset and /release median response time is ~90 seconds


stevekuznetsov:
![boskos-server-dashboard](https://user-images.githubusercontent.com/7328370/70570028-f5e4cb80-1b4f-11ea-9e5a-45fe3082a026.png)

/area boskos
/cc @krzyzacy @sebastienvas

stevekuznetsov:
```json
{"component":"boskos","file":"boskos/boskos.go:195","func":"main.handleAcquire.func1","handler":"handleStart","level":"info","msg":"From 172.16.74.74:33078","time":"2019-12-09T23:29:07Z"}
{"component":"boskos","file":"boskos/boskos.go:217","func":"main.handleAcquire.func1","level":"info","msg":"Request for a free gcp-quota-slice from -e8c8c, dest leased","time":"2019-12-09T23:29:07Z"}
{"component":"boskos","file":"boskos/boskos.go:238","func":"main.handleAcquire.func1","level":"info","msg":"Resource leased: {\"type\":\"gcp-quota-slice\",\"name\":\"01229229-ce89-4b67-b8d6-54537e57de01\",\"state\":\"leased\",\"owner\":\"-e8c8c\",\"lastupdate\":\"2019-12-09T23:29:08.296273508Z\",\"userdata\":{}}","time":"2019-12-09T23:29:08Z"}
```

I am having trouble finding logs that correspond with this behavior, makes me think it's something more pathological.

stevekuznetsov:
No, looks like a bona fide slowness:

```json
{
  "msg": "Request for a free aws-quota-slice from ci-op-mcgczqp0-76f5e, dest leased",
  "time": "2019-12-10T14:43:41Z"
}
{
  "msg": "Resource leased: {\"type\":\"aws-quota-slice\",\"name\":\"8ccc9a07-ed5d-4584-b845-3773bc5da3ff\",\"state\":\"leased\",\"owner\":\"ci-op-mcgczqp0-76f5e\",\"lastupdate\":\"2019-12-10T14:45:05.631972956Z\",\"userdata\":{}}",
  "time": "2019-12-10T14:45:05Z"
}
```

stevekuznetsov:
That means this has to be a slowdown in `ranch.Acquire()` or `json.Marshal()` as those are the only calls between those logs. I'd imagine it's the former.

sebastienvas:
[...]

PR #15990, opened 2020-01-23
Expose and collect Boskos HTTP metrics

ixdy:
The main Boskos service exposes useful Prometheus metrics about its HTTP API:
https://github.com/kubernetes/test-infra/blob/cd538657b7aaacccb650e49ad48de1785d7d3230/boskos/boskos.go#L60-L62

We should collect these.

We'll need to create a new static IP to bind to the load balancer. Assuming we think this is a good idea, I'll do that and then update this PR.

I'm still figuring how to update the various grafana configurations to graph this data, but we may as well start by collecting it.

/area boskos
/area monitoring
/hold

/assign @cjwagner @stevekuznetsov 
cc @Katharine as FYI

### DIFF
--- a/prow/cluster/boskos-metrics.yaml
+++ b/prow/cluster/boskos-metrics.yaml
@@ -21,11 +21,11 @@ spec:
   selector:
     app: boskos-metrics
   ports:
-  - name: default
+  - name: metrics
     protocol: TCP
-    port: 80
+    port: 9090
     targetPort: 9090
-  loadBalancerIP: 104.197.27.114
+  loadBalancerIP: 104.197.27.114  # boskos-metrics
   type: LoadBalancer
 ---
 apiVersion: apps/v1
--- a/prow/cluster/boskos.yaml
+++ b/prow/cluster/boskos.yaml
@@ -91,3 +91,19 @@ spec:
     protocol: TCP
     port: 80
     targetPort: 8080
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: boskos-http-metrics
+  namespace: test-pods
+spec:
+  selector:
+    app: boskos
+  ports:
+  - name: metrics
+    port: 9090
+    protocol: TCP
+    targetPort: 9090
+  loadBalancerIP: 34.70.169.97  # boskos-http-metrics
+  type: LoadBalancer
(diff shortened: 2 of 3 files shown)


prow/cluster/boskos-metrics.yaml
```diff
@@ -22,8 +22,12 @@ spec:
     app: boskos-metrics
   ports:
   - name: default
-    protocol: TCP
     port: 80
+    protocol: TCP
+    targetPort: 8080
```
ixdy: This is intentional: the boskos-metrics service proxies the main `/metric` API on port 8080, and this was masked in https://github.com/kubernetes/test-infra/pull/15356. Configuring the scraper to look at port 9090 (also included here) is a better fix.

ixdy:
(related to #15557 - it'd be good to know how the k8s boskos is doing in comparison.)

prow/cluster/boskos.yaml
```diff
@@ -91,3 +91,19 @@ spec:
...
+  namespace: test-pods
+spec:
+  selector:
+    app: boskos
+  ports:
+  - name: metrics
+    port: 9090
+    protocol: TCP
+    targetPort: 9090
+  loadBalancerIP: [fill this in]
```
cjwagner: Will this dummy value cause problems? (i.e. do we need to emit the field entirely until we have a valid value?)

prow/cluster/monitoring/additional-scrape-configs_secret.yaml
```diff
@@ -9,7 +9,8 @@ stringData:
       metrics_path: /metrics
       static_configs:
         - targets:
-            - "104.197.27.114"
+            - "104.197.27.114:9090"  # boskos resource metrics
+            - "fillmein-boskos-http-metrics:9090"  # boskos http metrics
```
cjwagner: We should probably wait to add this until we have a valid value to avoid failing scrapes.

prow/cluster/boskos.yaml
ixdy: right, I don't intend to submit this until we've reserved an IP address. I can do that, assuming there's consensus that this is a good idea.

prow/cluster/boskos-metrics.yaml
ixdy: It used to be exposed, prior to #15356. it basically just proxies the `/metric` API:

https://github.com/kubernetes/test-infra/blob/0a8437ab4b6cf302d8e170b945424d532fdb09e8/boskos/metrics/metrics.go#L131-L170

Nobody has complained that this is missing, so maybe we don't need it.

ixdy: (removed)

ixdy:
/hold cancel
:eyes:

PR #15998, opened 2020-01-23
Add build cluster secret to crier

Katharine:
It needs this to be able to look up pods.

/cc @cjwagner @BenTheElder


prow/cluster/crier_deployment.yaml
```diff
@@ -77,3 +81,7 @@ spec:
       - name: service-account
         secret:
           secretName: service-account
+      - name: cluster
+        secret:
+          defaultMode: 420
```
BenTheElder: seems legit

PR #16003, opened 2020-01-23
Ramp podinfo.json to 50%.

Katharine:
No perceptible impact thus far.

/cc @BenTheElder


BenTheElder:
I wouldn't expect be one as long as it's an ongoing load increase, GKE auto scales the control plane and we're just doing reads.
/lgtm
/approve

PR #16004, opened 2020-01-23
Try capping bazel's memory in some pull-kubernetes-e2e jobs

spiffxp:
My working theory is bazel's jvm is asking for more memory than is 
available, and getting oomkilled. Instead, let's cap its memory so that it
tries to GC instead of grow memory.

The 6Gi number was there for -gce before, capping bazel's jvm to 4Gi is a
total guess to leave some overhead for non-bazel stuff.

Add a 6Gi resource request for -rbe so we can compare apples and oranges to
see how this affects RBE and non-RBE.

This also reverts https://github.com/kubernetes/test-infra/pull/15980

ref: https://github.com/kubernetes/kubernetes/issues/87441

### DIFF
--- a/config/jobs/kubernetes/sig-cloud-provider/gcp/gcp-gce.yaml
+++ b/config/jobs/kubernetes/sig-cloud-provider/gcp/gcp-gce.yaml
@@ -42,6 +42,8 @@ presubmits:
         - --timeout=105
         - --scenario=kubernetes_e2e
         - --
+        # cap bazel's JVM memory usage to resources.requests.memory - 2Gi
+        - --inject-bazelrc=startup --host_jvm_args=-Xmx4g
         - --build=bazel
         - --cluster=
         - --extract=local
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes/sig-cloud-provider/gcp/gcp-gce.yaml
```diff
@@ -42,6 +42,8 @@ presubmits:
         - --timeout=105
         - --scenario=kubernetes_e2e
         - --
+        # cap bazel's JVM memory usage to resources.requests.memory - 2Gi
+        - --inject-bazelrc=startup --host_jvm_args=-Xmx4g
```
BenTheElder: has this been tested?

spiffxp: I tested this by modifying my .bazelrc locally.  Capping to 2g caused excessive garbage collection

BenTheElder:
/lgtm
/approve
here goes nothing

PR #16005, opened 2020-01-23
[capi] add the capi e2es to the presubmits

chuckha:
Signed-off-by: Chuck Ha <chuckh@vmware.com>

/assign @detiber @wfernandes

Fixes kubernetes-sigs/cluster-api#1884

### DIFF
--- a/config/jobs/kubernetes-sigs/cluster-api/cluster-api-presubmits.yaml
+++ b/config/jobs/kubernetes-sigs/cluster-api/cluster-api-presubmits.yaml
@@ -104,6 +104,29 @@ presubmits:
     annotations:
       testgrid-dashboards: sig-cluster-lifecycle-cluster-api
       testgrid-tab-name: pr-integration
+  - name: pull-cluster-api-e2e
+    labels:
+      preset-dind-enabled: "true"
+      preset-kind-volume-mounts: "true"
+    decorate: true
+    path_alias: sigs.k8s.io/cluster-api
+    always_run: true
+    spec:
+      containers:
+        - image: gcr.io/k8s-testimages/kubekins-e2e:v20200123-f79dd58-master
+          command:
+            - runner.sh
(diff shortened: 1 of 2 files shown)


config/jobs/kubernetes-sigs/cluster-api/cluster-api-presubmits.yaml
```diff
@@ -104,6 +104,29 @@ presubmits:
     annotations:
       testgrid-dashboards: sig-cluster-lifecycle-cluster-api
       testgrid-tab-name: pr-integration
+  - name: pull-cluster-api-e2e
```
detiber: Should we also add a periodic (and possibly also a postsubmit) job for this as well?

chuckha: this e2e tests that machine CRD and machine webhook is validating correctly -- I don't think we'd need a post submit or periodic for this. why do you think we need those?

wfernandes:
/lgtm
I'm fine with having them run at the same cadence as the capd tests for now.

chuckha:
Ah, the capd tests aren't running as a pull job, they're running as a periodic and as an optional PR job that you can trigger if you wish

config/jobs/kubernetes-sigs/cluster-api/cluster-api-presubmits.yaml
detiber: I'd like to have it as a signal that the current state of HEAD (or a particular release branch) is in a good state for release signal.

chuckha: sure, no problem, one sec

chuckha:
@detiber Updated! Ptal!

chuckha:
my fault, really thought i had checked all the possible duplicate values. Fixed now.

PR #16007, opened 2020-01-23
install rsync in KRTE

BenTheElder:
/cc @amwat @Katharine


images/krte/Dockerfile
```diff
@@ -67,6 +67,7 @@ RUN echo "Installing Packages ..." \
         python \
         python-dev \
         python-pip \
+        rsync \
```
BenTheElder: (needed for kubernetes make build)

PR #16008, opened 2020-01-23
Move stuck pod deletion from plank to sinker

Katharine:
Currently, when plank sees a stuck job, it marks the prowjob as errored and then deletes the pod. This prevents anything from subsequently inspecting the pod to determine why it errored, which is unfortunate.

Instead, have plank _only_ mark the job as errored. Sinker will then remove the pod after the completed job TTL expires.

Since the pod will still be running, this PR changes sinker's definition of termination to be when the job terminated, instead of when the pod terminated, ensuring that running jobs can be cleaned up if the prowjob state indicates they should be.

This is part of [improving prowjob visibility](https://github.com/kubernetes/test-infra/pull/16008).

/cc @cjwagner

### DIFF
--- a/prow/cmd/sinker/main_test.go
+++ b/prow/cmd/sinker/main_test.go
@@ -373,6 +373,28 @@ func TestClean(t *testing.T) {
 				},
 			},
 		},
+		&corev1api.Pod{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "completed-prowjob-ttl-expired-while-pod-still-pending",
+				Namespace: "ns",
+				Labels: map[string]string{
+					kube.CreatedByProw: "true",
+				},
+			},
+			Status: corev1api.PodStatus{
+				Phase:     corev1api.PodPending,
+				StartTime: startTime(time.Now().Add(-terminatedPodTTL * 2)),
+				ContainerStatuses: []corev1api.ContainerStatus{
+					{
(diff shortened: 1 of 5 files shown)


stevekuznetsov:
Do we have an overarching decision of what goes where? Why do we want two controllers operating on Pods and their lifecycle? Will `sinker` grow to encompass every type of job execution style?

Katharine:
We already have two controllers operating on pods and their lifecycle. Arguably sinker should not touch pods at all, but currently it does.

The point of the change here is not actually about moving the handling of this around, but to delay the deletion by the specified `terminated_pod_ttl`, enabling other tools to have a chance to inspect the pod. If pod lifecycle is ever centralised in plank, this logic would presumably move with it.

prow/cmd/sinker/main_test.go
```diff
@@ -373,6 +373,28 @@ func TestClean(t *testing.T) {
 				},
 			},
 		},
+		&corev1api.Pod{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "ttl-expired-while-still-pending",
```
cjwagner: nit: It would be helpful to clarify that this is referring to a pending pod and that the ProwJob is completed. I was initially mislead by assuming that this meant the ProwJob was still pending.

Katharine: Done.

PR #16011, opened 2020-01-24
Enable allow-same-origin for Spyglass sandbox

### LINKED ISSUE #16002: Spyglass returns 403 for `static` files on private instance
**What happened**:
Spyglass is returning a **403** when fetching `static` files on a private Deck instance that is behind an oauth proxy for some ~non-Google~ users.

I expect Spyglass to return a **200** when fetching `static` files, similar to the retrieval of other resources.

1. Configure an oauth proxy for a private Deck instance ([example](https://github.com/istio/test-infra/blob/830ea2859027294013b5340950a0fba8ca0b7a95/prow/cluster/deck_private_deployment.yaml)).
2. Ensure you are not signed in with a Github or Gmail account that may already have access to the logs/resources and are associated with Google employment. 
2. Navigate to a Spyglass view of a job (i.e. `/view/gcs/...`) .
3. Static files will not load; the error will be a **403**.

Screen Shot 2020-01-23 at 2 16 45 PM
Screen Shot 2020-01-23 at 2 17 20 PM

/assign

clarketm:
See https://github.com/kubernetes/test-infra/issues/16002#issue-554421476 and https://github.com/kubernetes/test-infra/issues/16002#issuecomment-578030986 for a description of this issue and rationale for this change.

fix #16002

### DIFF
--- a/prow/cmd/deck/template/spyglass.html
+++ b/prow/cmd/deck/template/spyglass.html
@@ -39,7 +39,7 @@
     <div class="mdl-card__title lens-title"><h3 class="mdl-card__title-text">{{$config.Title}}</h3></div>
     <div id="{{$config.Name}}-view-container" class="lens-view-content mdl-card__supporting-text">
       <img src="/static/kubernetes-wheel.svg" alt="loading spinner" class="loading-spinner is-active lens-card-loading" id="{{$config.Name}}-loading">
-      <iframe class="lens-container" style="visibility: hidden;" id="iframe-{{$index}}" sandbox="allow-scripts allow-top-navigation allow-popups" data-lens-index="{{$index}}" data-lens-name="{{$config.Name}}"{{if $config.HideTitle}} data-hide-title="true"{{end}}></iframe>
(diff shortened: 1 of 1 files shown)


Katharine:
Doing this effectively completely disables the sandbox, so I am generally opposed. It might be acceptable as a flag, but even then I'd rather not.

clarketm:
> Doing this effectively completely disables the sandbox, so I am generally opposed. It might be acceptable as a flag, but even then I'd rather not.

Ad there is a workaround by whitelisting the host, this change is not strictly necessary. So, if you are opposed to the change I will close. What exactly is the concern here?

Katharine:
The intent of sandboxing the iframes is primarily to prevent them from doing things that are not permitted by their APIs - for instance, accessing other deck endpoints or mutating the page they are inside. This is currently to encourage developers to behave. In the future, when we support [external lenses](https://docs.google.com/document/d/1YBqwFvlerplpb9kMm7rO0eRLYheemmkbUL6GUmKlopg/edit), this will be doubly true as the lenses are not necessarily under our control.

By making the iframe have the same origin, it is free to make requests to arbitrary parts of deck and modify its containing page freely, which are both undesirable in the long term.

PR #16014, opened 2020-01-24
ghproxy: do not record zero-value rate-limits when empty

stevekuznetsov:
When a GitHub response does not have any rate limit remaining header, we
cannot update our metric.

```json
{"component":"ghproxy","file":"ghproxy/ghmetrics/ghmetrics.go:107","func":"k8s.io/test-infra/ghproxy/ghmetrics.CollectGitHubTokenMetrics","header":"","level":"debug","msg":"Parsed GitHub header as indicating no remaining rate-limit.","time":"2020-01-24T17:49:49Z","user-agent":""}
```

/assign @cjwagner

### DIFF
--- a/ghproxy/ghmetrics/ghmetrics.go
+++ b/ghproxy/ghmetrics/ghmetrics.go
@@ -93,6 +93,9 @@ func init() {
 // `github_token_usage` as well as `github_token_reset` on prometheus.
 func CollectGitHubTokenMetrics(tokenHash, apiVersion string, headers http.Header, reqStartTime, responseTime time.Time) {
 	remaining := headers.Get("X-RateLimit-Remaining")
+	if remaining == "" {
+		return
+	}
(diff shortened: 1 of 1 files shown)


ghproxy/ghmetrics/ghmetrics.go
```diff
@@ -93,6 +93,9 @@ func init() {
 // `github_token_usage` as well as `github_token_reset` on prometheus.
 func CollectGitHubTokenMetrics(tokenHash, apiVersion string, headers http.Header, reqStartTime, responseTime time.Time) {
 	remaining := headers.Get("X-RateLimit-Remaining")
+	if remaining == "" {
+		return
```
cjwagner: Should we log a debug message so that we can generate some data on how often this happens?

stevekuznetsov: We can look at the data from the last N weeks this has been live and/or all the zero-blips ... unclear if it's useful.

PR #16016, opened 2020-01-24
Add newlines to --inject-bazelrc lines

spiffxp:
So that if we pass --inject-bazelrc=foo --inject-bazelrc=bar, it
gets written to bazelrc as
```
foo
bar
```
instead of
```
foobar
```

This is what caused me to open the revert PR #16015


BenTheElder:
what could go wrong
/lgtm
/approve

PR #16017, opened 2020-01-24
Make sinker robust to unreachable build clusters.

cjwagner:
Previously sinker would abort cleaning immediately if it failed to list pods in a build cluster. Since build clusters are processed sequentially this means that an unreachable build cluster can cause working build clusters to be skipped and pods won't be GCed.

This PR makes sinker skip problematic build clusters instead of aborting completely and also improves the logging.

/assign @Katharine


cjwagner:
Going to update the unit tests.

cjwagner:
/hold cancel
Updated the unit test to prevent a regression. PTAL.

matthyx:
<3
/lgtm
/hold

cjwagner:
`/b/f/w/prow/cmd/sinker/main.go:165:34: undefined: corev1api`
Hmmm, it is both imported and listed as a dependency in the bazel target.
/retest

cjwagner:
Ugh, bazel is the worst. Both `bazel build //...` and `bazel test //...` pass locally. Given that I didn't add any dependencies to the repo I don't know how I could have made this pass locally and not in CI.
Does anyone have any idea what is going on here or know how I can debug the RBE stuff to figure out why this build isn't deterministic?

cjwagner:
Fixed the problem (thanks @BenTheElder) PTAL
/hold cancel

matthyx:
Interesting... I would like to hear the reason if you don't mind.
/lgtm

BenTheElder:
Conflicting commit with the imports in main.go

matthyx:
Thanks!

PR #16024, opened 2020-01-25
fix dashboard typo on kind job

aojea:
it seems that the testgrid dashboad config has a typo
> testgrid-dashboards: sig-testing-kkind


BenTheElder:
thanks!

PR #16027, opened 2020-01-26
Add kubernetes signal on flakes

aojea:
debugging flakes is challenging, mainly because you can't reproduce the error and you don't know in advance what can be causing the flake behavior so you have to experiment and try different hypothesis.
KIND provides you a local environment that allows you to do a more advanced troubleshooting in your local workstation.
Adding a job that gives signal on flakes using KIND can be very useful to try to reproduce it locally.

### DIFF
--- a/config/jobs/kubernetes/sig-testing/kubernetes-kind.yaml
+++ b/config/jobs/kubernetes/sig-testing/kubernetes-kind.yaml
@@ -426,3 +426,106 @@ presubmits:
       testgrid-num-failures-to-alert: '10'
       testgrid-alert-stale-results-hours: '24'
       testgrid-create-test-group: 'true'
+
+periodics:
+- name: periodic-kubernetes-e2e-kind-flakes
+  decorate: true
+  interval: 2h
+  labels:
+    preset-bazel-scratch-dir: "true"
+    preset-bazel-remote-cache-enabled: "true"
+    preset-dind-enabled: "true"
+    preset-kind-volume-mounts: "true"
+  decoration_config:
+    timeout: 60m
+  path_alias: k8s.io/kubernetes
+  spec:
+    containers:
+    - image: gcr.io/k8s-testimages/krte:v20200123-dd36597-master
+      command:
+      - wrapper.sh
+      - bash
+      - -c
+      - gsutil cp -P gs://bentheelder-kind-ci-builds/latest/kind-linux-amd64 "${PATH%%:*}/kind" && gsutil cat gs://bentheelder-kind-ci-builds/latest/e2e-k8s.sh | sh
+      env:
+      - name: FOCUS
+        value: \[Flaky\]
+      # TODO(bentheelder): reduce the skip list further
+      - name: SKIP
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes/sig-testing/kubernetes-kind.yaml
```diff
@@ -426,3 +426,97 @@ presubmits:
...
+    containers:
+    - image: gcr.io/k8s-testimages/krte:v20200123-dd36597-master
+      command:
+      - wrapper.sh
+      - bash
+      - -c
+      - gsutil cp -P gs://bentheelder-kind-ci-builds/latest/kind-linux-amd64 "${PATH%%:*}/kind" && gsutil cat gs://bentheelder-kind-ci-builds/latest/e2e-k8s.sh | sh
+      env:
+      - name: FOCUS
+        value: \[Flaky\]
```
aojea: Please verify the regex are correct?

config/jobs/kubernetes/sig-testing/kubernetes-kind.yaml
```diff
@@ -426,3 +426,100 @@ presubmits:
...
+      # TODO(bentheelder): reduce the skip list further
+      - name: SKIP
+        value: \[Slow\]|\[Disruptive\]|\[Feature:.+\]|PodSecurityPolicy|LoadBalancer|load.balancer|In-tree.Volumes.\[Driver:.nfs\]|PersistentVolumes.NFS|Simple.pod.should.support.exec.through.an.HTTP.proxy|subPath.should.support.existing|ReplicationController.light.Should.scale.from.1.pod.to.2.pods|should.provide.basic.identity|\[NodeFeature:PodReadinessGate\]
```
aojea: and these regexes are necessary if with Focus on Flaky only?

aojea:
@BenTheElder I've renamed the file name to include periodic jobs inside, instead of creating a new file. Check if that is ok with the way you want to have organized the jobs under k8s

config/jobs/kubernetes/sig-testing/kubernetes-kind.yaml
BenTheElder: the skip are things we don't want to run for various reasons... including no support.

this is going to be a pain to cleanup later.

config/jobs/kubernetes/sig-testing/kubernetes-kind.yaml
BenTheElder: sure, but I'm not clear that the idea is even useful

BenTheElder:
I'm not sure that running all the flaky tests together is super helpful...?

I think if anyone wants to be running the flaky tests they should open a PR to debug them and get them to not be flaky, by just tackling _a_ test and getting the tag removed.

Flaky is a tag we use for bad tests. People should actively fix those. Passively running them seems :man_shrugging:

liggitt:
> I think if anyone wants to be running the flaky tests they should open a PR to debug them and get them to not be flaky, by just tackling _a_ test and getting the tag removed.
> 
> Flaky is a tag we use for bad tests. People should actively fix those. Passively running them seems 🤷‍♂

Agree. I'd go so far as to say we shouldn't release with tests tagged as flaky. If they are important tests, we need their signal to know if we should release. If they're not important, we should delete them.

aojea:
makes sense

PR #16029, opened 2020-01-27
kops: Ensure kops installs the same version of k8s we're using

justinsb:
Some tests are tightly bound to a particular version of k8s, for
example https://github.com/kubernetes/kubernetes/pull/87440 both
changed the behaviour and the tests in one PR.  We need therefore to
run the exact same k8s version as the tests.  This is a follow-on to
k8s version we download.

### DIFF
--- a/config/jobs/kubernetes/kops/kops-config.yaml
+++ b/config/jobs/kubernetes/kops/kops-config.yaml
@@ -81,6 +81,8 @@ presubmits:
         - --aws-cluster-domain=test-cncf-aws.k8s.io
         - --check-leaked-resources=false
         - --cluster=
+        - --env=KOPS_DEPLOY_LATEST_URL=https://storage.googleapis.com/kubernetes-release/release/latest.txt
+        - --env=KOPS_KUBE_RELEASE_URL=https://storage.googleapis.com/kubernetes-release/release/
         - --extract=release/latest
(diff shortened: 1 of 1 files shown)


justinsb:
cc @rifelpet / @hakman - does this look right?  I think this is what is causing the kops tests to fail after https://github.com/kubernetes/kubernetes/pull/87440 .

Looking at e.g. https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kops/8408/pull-kops-e2e-kubernetes-aws/1221612662731837444/build-log.txt, I can see that we are running kops with `--kubernetes-version https://storage.googleapis.com/kubernetes-release-dev/ci/v1.18.0-alpha.2.167+9162c5d7f4a912` but we downloaded the 1.18.0-alpha.2:

```
Downloading kubernetes release v1.18.0-alpha.2
  from https://storage.googleapis.com/kubernetes-release/release/v1.18.0-alpha.2/kubernetes.tar.gz
```

rifelpet:
Yes this looks reasonable to me. FWIW these variables are referenced here:

https://github.com/kubernetes/test-infra/blob/263b0e66f75741b01ceeb4f67506e0a8dfcd2563/images/kubekins-e2e/kops-e2e-runner.sh#L51-L59

/lgtm

PR #16031, opened 2020-01-27
Kops - Minor tweaks for periodics tests

hakman:
Make all periodics for versions look the same.

### DIFF
--- a/config/jobs/kubernetes/sig-cloud-provider/aws/kops/kops-periodics-versions.yaml
+++ b/config/jobs/kubernetes/sig-cloud-provider/aws/kops/kops-periodics-versions.yaml
@@ -7,7 +7,7 @@ periodics:
     preset-aws-credential: "true"
   decorate: true
   decoration_config:
-    timeout: 140m
+    timeout: 90m
   spec:
     containers:
     - command:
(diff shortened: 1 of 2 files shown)


config/jobs/kubernetes/sig-cloud-provider/aws/kops/kops-periodics-versions.yaml
```diff
@@ -19,24 +19,24 @@ periodics:
       - --env=KUBE_SSH_USER=admin
       - --extract=ci/latest-1.18
       - --ginkgo-parallel
-      - --kops-version=https://storage.googleapis.com/kops-ci/bin/latest-ci-updown-green.txt
```
rifelpet: this kops version is what is published by this `updown` job here:

https://github.com/kubernetes/test-infra/blob/0d84f4912879c865dae644733c62379c77dcdd76/config/jobs/kubernetes/sig-cloud-provider/aws/kops/kops-periodics.yaml#L219-L247

https://testgrid.k8s.io/sig-cluster-lifecycle-kops#kops-aws-updown

I believe that job just creates and destroys a simple cluster to make sure that kops' master branch isn't totally broken. if it succeeds it publishes that commit to this URL. This way if something goes wrong on master, we don't end up failing all of our periodics because of it. Many of the other jobs in kops-periodics.yaml use updown-green as well, so maybe we should try to be consistent?

hakman: Updated :)

rifelpet:
Awesome, great work. I incorrectly thought that periods weren't allowed in testgrid tab names...

/approve
/lgtm

PR #15609, opened 2019-12-14
Switch to ubuntu 18.04 image for cpu-manager tests

odinuge:
Use the same image as the other serial tests.

The old one fails during preflight check becuse docker cli
is missing some args.

eg. https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-node-kubelet-serial-cpu-manager/1205820052486492165

PR #16037, opened 2020-01-27
Sync e2e_node approvers with sig-node approvers

derekwaynecarr:
Synchronizes with approvers in main `kubernetes/kubernetes` repository for `/test/e2e_node`.

### DIFF
--- a/jobs/e2e_node/OWNERS
+++ b/jobs/e2e_node/OWNERS
@@ -4,3 +4,8 @@ approvers:
 - Random-Liu
 - yguo0905
 - yujuhong
+- tallclair
+- derekwaynecarr
+- ConnorDoyle
+- klueska
+- dchen1107


derekwaynecarr:
Synchronizing the set of approvers based on experience from https://github.com/kubernetes/test-infra/pull/15609

jobs/e2e_node/OWNERS
```diff
@@ -4,3 +4,8 @@ approvers:
 - Random-Liu
 - yguo0905
 - yujuhong
+- tallclair
+- derekwaynecarr
+- ConnorDoyle
+- klueska
+- dchen1107
```
yujuhong: Is this the list of OWNERS for the kubelet package or all sub-packages under it? It doesn't seem like either...

derekwaynecarr: @yujuhong I didnt remove any existing right, but just added the names from the /test/e2e_node/OWNERS listing here:
https://github.com/kubernetes/kubernetes/blob/master/test/e2e_node/OWNERS

PR #16038, opened 2020-01-27
Add ci flaky repro job

spiffxp:
Currently when a job is tagged as Flaky it is only run in CI in a job
that focuses exclusively on Flaky tests. However, many times flakes do
not end up reproducing in this flakes-only job because there is not
enough test volume to cause them. This means efforts to prove that a
flake is fixed involve opening a PR that removes the [Flake] tag from
a test and repeatedly manually triggering tests.

Let's introduce a gce job named flaky-repro that runs flaky tests in
the conditions that reproduce flakes. Take the "default" gce job, and
subtract its skip of [Flaky]. Since this job is most likely going to
be red or flaky all of the time, keep it off of sig-release boards and
drop alerting.

This should give us more evidence (or at least less manually-generated)
that a given flake has been fixed and no longer reproduces alongside
other tests at high volume

### DIFF
--- a/config/jobs/kubernetes/sig-cloud-provider/gcp/gcp-gce.yaml
+++ b/config/jobs/kubernetes/sig-cloud-provider/gcp/gcp-gce.yaml
@@ -239,6 +239,36 @@ periodics:
     description: Uses kubetest to run e2e tests (+Feature:many, -many) against a cluster created with cluster/kube-up.sh
 
 - interval: 30m
+  name: ci-kubernetes-e2e-gci-gce-flaky-repro
+  labels:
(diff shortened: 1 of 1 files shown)


spiffxp:
/cc @liggitt @aojea 
ref: https://kubernetes.slack.com/archives/C09QZ4DQB/p1579990652033300

aojea:
heh, another bookmark for my browser ;) `gce-cos-master-flaky-repro`
LGTM, I think the config is correct but prefer others to check

aojea:
/lgtm
what can be wrong :smile:

PR #16039, opened 2020-01-27
Tide: support query by "author"

clarketm:
Add support in Tide for query by "author". I am weighting an incorrect author with very high diff (similar to branch) due to the relative importance of the field being accurate (when specified).

This would be a useful, extensible way to expose functionality like requested in https://github.com/istio/test-infra/issues/2304

### DIFF
--- a/prow/cmd/deck/static/pr/pr.ts
+++ b/prow/cmd/deck/static/pr/pr.ts
@@ -32,6 +32,7 @@ interface ProcessedQuery {
   missingLabels: ProcessedLabel[];
   excludedBranches?: string[];
   includedBranches?: string[];
+  author?: string;
   milestone?: string;
 }
 
@@ -46,6 +47,7 @@ class TideQuery {
     public missingLabels?: string[];
     public excludedBranches?: string[];
     public includedBranches?: string[];
+    public author?: string;
     public milestone?: string;
 
     constructor(query: ITideQuery) {
@@ -56,6 +58,7 @@ class TideQuery {
         this.missingLabels = query.missingLabels;
         this.excludedBranches = query.excludedBranches;
         this.includedBranches = query.includedBranches;
+        this.author = query.author;
         this.milestone = query.milestone;
     }
 
@@ -476,7 +479,7 @@ function createTidePoolLabel(pr: PullRequest, tidePool?: TidePool): HTMLElement 
  */
 function createTitleLabel(isMerge: boolean, jobStatus: VagueState, noQuery: boolean,
                           labelConflict: boolean, mergeConflict: boolean,
-                          branchConflict: boolean, milestoneConflict: boolean): HTMLElement {
+                          branchConflict: boolean, authorConflict: boolean, milestoneConflict: boolean): HTMLElement {
     const label = document.createElement("span");
     label.classList.add("title-label");
 
@@ -486,6 +489,9 @@ function createTitleLabel(isMerge: boolean, jobStatus: VagueState, noQuery: bool
     } else if (isMerge) {
         label.textContent = "Merged";
         label.classList.add("merging");
+    } else if (authorConflict) {
+        label.textContent = "Blocked from merging by current author";
+        label.classList.add("pending");
     } else if (branchConflict) {
         label.textContent = "Blocked from merging into target branch";
         label.classList.add("pending");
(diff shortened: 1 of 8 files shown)


prow/config/tide_test.go
```diff
@@ -35,6 +35,7 @@ var testQuery = TideQuery{
 	Repos:                  []string{"k/k", "k/t-i"},
 	Labels:                 []string{labels.LGTM, labels.Approved},
 	MissingLabels:          []string{"foo"},
+	Author:                 "batman",
```
cjwagner: :bat:

prow/tide/status.go
```diff
@@ -148,6 +148,15 @@ func requirementDiff(pr *PullRequest, q *config.TideQuery, cc contextChecker) (s
 		}
 	}
 
+	// Weight incorrect author with very high diff so that we select the query
+	// for the correct author.
+	if q.Author != "" && string(pr.Author.Login) != q.Author {
```
cjwagner: Use `github.NormLogin` to normalize these before comparing (just strings.ToLower under the hood).

prow/tide/status_test.go
```diff
@@ -420,6 +446,7 @@ func TestExpectedStatus(t *testing.T) {
 					IncludedBranches: tc.branchWhitelist,
 					Labels:           neededLabels,
 					MissingLabels:    forbiddenLabels,
+					Author:           "batman",
```
cjwagner: Please test the behavior when only one of the two queries specifies an author and the PR author does not match.

cjwagner:
Changes look good, but please also update the front end to display the new requirement when appropriate: https://github.com/kubernetes/test-infra/blob/master/prow/cmd/deck/static/pr/pr.ts

clarketm:
> Changes look good, but please also update the front end to display the new requirement when appropriate: https://github.com/kubernetes/test-infra/blob/master/prow/cmd/deck/static/pr/pr.ts

Thanks. Will update that file soon and update.

clarketm:
@cjwagner - updated the FE code. PTAL.

prow/cmd/deck/static/pr/pr.ts
```diff
@@ -489,6 +492,9 @@ function createTitleLabel(isMerge: boolean, jobStatus: VagueState, noQuery: bool
     } else if (branchConflict) {
         label.textContent = "Blocked from merging into target branch";
         label.classList.add("pending");
+    } else if (authorConflict) {
```
cjwagner: nit: This should probably be prioritized over branch conflicts since a PR can switch branches, but not authors.

prow/cmd/deck/static/pr/pr.ts
```diff
@@ -803,6 +809,7 @@ function createQueryDetailsBtn(query: ProcessedQuery): HTMLTableDataCellElement
     const allRequired = document.getElementById("query-all-required")!;
     const allForbidden = document.getElementById("query-all-forbidden")!;
     iconButton.addEventListener("click", () => {
+        fillDetail("#query-detail-author", query.author);
```
cjwagner: Doesn't this need a corresponding template change?

prow/cmd/deck/static/pr/pr.ts
```diff
@@ -1132,6 +1136,13 @@ function closestMatchingQueries(pr: PullRequest, queries: TideQuery[]): Processe
     return processedQueries;
 }
 
+/**
+ * Normalizes GitHub login strings
+ */
+function normLogin(login: string = ""): string {
+    return login.toLowerCase().replace(/^@/, "");
```
cjwagner: Is this replacement necessary? I think it would make more sense for this to match the behavior of the go version of `NormLogin`.

clarketm: The `replace` here is the JS equivalent of the Go `TrimPrefix`.
https://github.com/kubernetes/test-infra/blob/5c17574a9a87e3fce30d20925bf8ed50ed533f64/prow/github/types.go#L175-L178

prow/cmd/deck/static/pr/pr.ts
clarketm: You are correct; updated.

prow/cmd/deck/static/pr/pr.ts
clarketm: I moved it above the `branchConflict`.

prow/cmd/deck/static/pr/pr.ts
cjwagner: Oh I didn't realize we handled the `@` prefix.

PR #16050, opened 2020-01-29
Make GCE Windows versions explicit, add Windows 1909 job.

pjh:
This PR creates a new test job that will run the GCE Windows e2e tests against nodes running Windows Server version 1909, and explicitly annotates existing test jobs as using Windows Server 2019.

The new 1909 job has the same annotations as the current 2019 job, which is intended. Going forward we'll want to test K8s on GCE against both an LTSC version of Windows (2019) and an SAC version (1909 for now, 20-04 for 1.19 or 1.20, ...).


pjh:
/hold

https://github.com/kubernetes/kubernetes/pull/87629 needs to be merged first.

pjh:
Rebased.

/unhold
https://github.com/kubernetes/kubernetes/pull/87629 has approval, should get merged soon.

PR #16052, opened 2020-01-29
Kops - Switch Amazon Linux 2 e2e test to containerd

hakman:
Amazon Linux 2 cannot install latest Docker because of dependencies issues with `container-selinux`. Best solution for now is to install containerd from static package instead.

See https://github.com/kubernetes/kops/pull/8425 for reference.


rifelpet:
do you think we should recommend containerd for anyone wanting to use AL2 ?

/lgtm
/approve

hakman:
@rifelpet let's see this works as we want it and we could update the instructions later.

I am thinking of improving a bit the support for static binaries for Docker also, so it may not be a requirement to use containerd with AL2 when 1.18 is released.

PR #16054, opened 2020-01-29
Kops - Switch distros marker to release/latest

hakman:


rifelpet:
Whats the motivation for this exactly? That we already use "prerelease" markers for the k8s version jobs, so we have sufficient coverage and would rather opt to use stable versions for the other batch of e2e jobs? That makes sense to me, just want to confirm.

hakman:
@rifelpet would like to use something stable in general. Based on our chat in last office hours, I remember that we said we can use stable everywhere except 1-2 jobs on head.

rifelpet:
makes sense, and agreed!

/lgtm
/approve

PR #16055, opened 2020-01-29
Bump kubemark-500 timeout to 80m

mm4tt:
The current timeout (70m) seems to be causing spurious failures and
making the job flaky.

Example failure: https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-kubemark-500-gce/1222309408109760514
Looking at the recent graphs 70min seems to be to close to the actual time the job takes - 
ujkEC8mhfxp


jkaniuk:
What about older tests?

https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-kubemark-500-gce-stable2/1221810341223075840

mm4tt:
> What about older tests?
> 
> https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-kubemark-500-gce-stable2/1221810341223075840

Good point, I missed those.

PR #16058, opened 2020-01-29
Support for NAT in GKE clusters

jkaniuk:
Support for NAT configuration in GKE clusters.

Clusters with private nodes needs NAT if public access is needed, e.g. for downloading public docker images.

NAT min-ports-per-vm flag defines NAT ports to nodes mapping and could be important for 5k tests to limit number of external IPs used.
https://cloud.google.com/nat/docs/overview#number_of_nat_ports_and_connections

/cc mm4tt wojtek-t

### DIFF
--- a/kubetest/gke.go
+++ b/kubetest/gke.go
@@ -55,6 +55,8 @@ var (
 	gkeReleaseChannel              = flag.String("gke-release-channel", "", "(gke only) if specified, bring up GKE clusters from that release channel.")
 	gkeSingleZoneNodeInstanceGroup = flag.Bool("gke-single-zone-node-instance-group", true, "(gke only) Add instance groups from a single zone to the NODE_INSTANCE_GROUP env variable.")
 	gkeNodePorts                   = flag.String("gke-node-ports", "", "(gke only) List of ports on nodes to open, allowing e.g. master to connect to pods on private nodes. The format should be 'protocol[:port[-port]],[...]' as in gcloud compute firewall-rules create --allow.")
+	gkeCreateNat                   = flag.Bool("gke-create-nat", false, "(gke only) Configure Cloud NAT allowing outbound connections in cluster with private nodes.")
+	gkeNatMinPortsPerVm            = flag.Int("gke-nat-min-ports-per-vm", 64, "(gke only) Specify number of ports per cluster VM for NAT router. Number of ports * number of nodes / 64k = number of auto-allocated IP addresses (there is a hard limit of 100 IPs).")
 
 	// poolRe matches instance group URLs of the form `https://www.googleapis.com/compute/v1/projects/some-project/zones/a-zone/instanceGroupManagers/gke-some-cluster-some-pool-90fcb815-grp`. Match meaning:
 	// m[0]: path starting with zones/
@@ -85,6 +87,8 @@ type gkeDeployer struct {
 	network                     string
 	subnetwork                  string
 	subnetworkRegion            string
+	createNat                   bool
+	natMinPortsPerVm            int
 	image                       string
 	imageFamily                 string
 	imageProject                string
@@ -156,6 +160,8 @@ func newGKE(provider, project, zone, region, network, image, imageFamily, imageP
 	g.additionalZones = *gkeAdditionalZones
 	g.nodeLocations = *gkeNodeLocations
 	g.nodePorts = *gkeNodePorts
+	g.createNat = *gkeCreateNat
+	g.natMinPortsPerVm = *gkeNatMinPortsPerVm
 
 	err := json.Unmarshal([]byte(*gkeShape), &g.shape)
 	if err != nil {
@@ -461,6 +467,9 @@ func (g *gkeDeployer) TestSetup() error {
 	if err := g.ensureFirewall(); err != nil {
 		return err
 	}
+	if err := g.ensureNat(); err != nil {
+		return err
+	}
 	if err := g.setupEnv(); err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


kubetest/gke.go
```diff
@@ -55,6 +55,8 @@ var (
 	gkeReleaseChannel              = flag.String("gke-release-channel", "", "(gke only) if specified, bring up GKE clusters from that release channel.")
 	gkeSingleZoneNodeInstanceGroup = flag.Bool("gke-single-zone-node-instance-group", true, "(gke only) Add instance groups from a single zone to the NODE_INSTANCE_GROUP env variable.")
 	gkeNodePorts                   = flag.String("gke-node-ports", "", "(gke only) List of ports on nodes to open, allowing e.g. master to connect to pods on private nodes. The format should be 'protocol[:port[-port]],[...]' as in gcloud compute firewall-rules create --allow.")
+	gkeNat                         = flag.Bool("gke-nat", false, "(gke only) Configure Cloud NAT allowing outbound connections in cluster with private nodes.")
```
mm4tt: Given that this is a bool flag, can we rename it to be something like `gke-set-up-nat` or `gke-create-nat`?

kubetest/gke.go
```diff
@@ -55,6 +55,8 @@ var (
...
+	gkeNatMinPortsPerVm            = flag.Int("gke-nat-min-ports-per-vm", 64, "(gke only) Specify number of ports per cluster VM for NAT router. Number of ports * number of nodes / 64k = number of used static addresses.")
```
mm4tt: What do you mean by static? Wouldn't be better to say "auto-allocated" as this is the way you're configuring cloud nat below?

mm4tt: Should we mention here that there are limits on max number of nat addresses? See https://cloud.google.com/nat/quota

kubetest/gke.go
```diff
@@ -85,6 +87,8 @@ type gkeDeployer struct {
 	network                     string
 	subnetwork                  string
 	subnetworkRegion            string
+	nat                         bool
```
mm4tt: Same here - enableNat, createNat or something

kubetest/gke.go
```diff
@@ -608,12 +617,136 @@ func (g *gkeDeployer) cleanupNetworkFirewalls() (int, error) {
...
+	return strings.TrimSuffix(string(result), "\n"), nil
+}
+
+func (g *gkeDeployer) getNatName() (string, error) {
+	if err := g.getInstanceGroups(); err != nil {
+		return "", err
+	}
+	// Create unique NAT router name based on hash
+	// from the lexically first node pool instead.
+	return "nat-router-" + g.instanceGroups[0].uniq, nil
```
mm4tt: Cannot we use g.cluster for the nat name?

kubetest/gke.go
```diff
@@ -608,12 +617,136 @@ func (g *gkeDeployer) cleanupNetworkFirewalls() (int, error) {
...
+
 func (g *gkeDeployer) Down() error {
 	firewall, err := g.getClusterFirewall()
 	if err != nil {
 		// This is expected if the cluster doesn't exist.
 		return nil
 	}
+	nat, err := g.getNatName()
+	if err != nil {
+		// This is expected if the cluster doesn't exist.
```
mm4tt: I don't like this, getNatName returns error whenever getInstanceGroups returns error and the latter doesn't return error only when cluster doesn't exist.

kubetest/gke.go
```diff
@@ -608,12 +617,136 @@ func (g *gkeDeployer) cleanupNetworkFirewalls() (int, error) {
...
+		"--router="+nat,
+		"--router-region="+region,
+		"--format=value(name)")) != nil {
+		log.Printf("Couldn't describe nat '%s', assuming it doesn't exist and creating it", nat)
+		if err := control.FinishRunning(exec.Command("gcloud", "compute", "routers", "nats", "create", nat,
+			"--project="+g.project,
+			"--router="+nat,
+			"--router-region="+region,
+			"--auto-allocate-nat-external-ips",
+			"--nat-all-subnet-ip-ranges",
```
mm4tt: If you use this setting you'll effectively allocate extra 1024 ports per node (for secondary ip ranges). This may introduce scalability limitations for large clusters

kubetest/gke.go
```diff
@@ -608,12 +617,136 @@ func (g *gkeDeployer) cleanupNetworkFirewalls() (int, error) {
...
+			"--min-ports-per-vm="+strconv.Itoa(g.natMinPortsPerVm),
+			"--enable-logging")); err != nil {
```
mm4tt: Is there any perf impact of nat logging? I think I've never used it

kubetest/gke.go
jkaniuk: OK.
create-nat sounds better

kubetest/gke.go
jkaniuk: Rewritten that part

kubetest/gke.go

kubetest/gke.go

kubetest/gke.go
jkaniuk: Changed

kubetest/gke.go
jkaniuk: Changed to use only primary range

kubetest/gke.go
jkaniuk: Removed

kubetest/gke.go
```diff
@@ -610,12 +619,127 @@ func (g *gkeDeployer) cleanupNetworkFirewalls() (int, error) {
...
+	}
+	nat := g.getNatName()
+
+	// Create this unique NAT router only if it does not exist yet.
+	if control.NoOutput(exec.Command("gcloud", "compute", "routers", "describe", nat,
+		"--project="+g.project,
+		"--region="+region,
+		"--format=value(name)")) != nil {
+		log.Printf("Couldn't describe router '%s', assuming it doesn't exist and creating it", nat)
+
```
mm4tt: NIT: Remove this blank line or add one at line 656 for consistency

kubetest/gke.go
mm4tt: Now this if clause is duplicated?

jkaniuk: Fixed

kubetest/gke.go

kubetest/gke.go
```diff
@@ -610,6 +619,116 @@ func (g *gkeDeployer) cleanupNetworkFirewalls() (int, error) {
...
+		"--router="+nat,
+		"--router-region="+region,
+		"--format=value(name)")) == nil {
+		log.Printf("Found NAT configuration '%s', deleting", nat)
+		err = control.FinishRunning(exec.Command("gcloud", "compute", "routers", "nats", "delete", "-q", nat,
+			"--project="+g.project,
+			"--router="+nat,
+			"--router-region="+region))
+		if err != nil {
+			return fmt.Errorf("error deleting NAT config: %v", err)
```
wojtek-t: This means that problems with deleting NAT config will also lead to leaking router. Which isn't perfect

jkaniuk: I thought that removing router requires removal of config first, but that is not the case. Changed to only remove router (with config being deleted automatically).

mm4tt: I also thought that you need to first delete config before deleting router.
Nice finding!

kubetest/gke.go
```diff
@@ -619,6 +628,99 @@ func (g *gkeDeployer) cleanupNetworkFirewalls() (int, error) {
...
+	if !g.createNat {
+		return nil
+	}
+	region, err := g.getRegion(g.region, g.zone)
+	if err != nil {
+		return fmt.Errorf("error finding region for NAT router: %v", err)
+	}
+	nat := g.getNatName()
+
+	if control.NoOutput(exec.Command("gcloud", "compute", "routers", "describe", nat,
```
wojtek-t: Please add a comment that this will automatically delete the config will be deleted automatically here (to ensure that future readers will not wonder about that nor they will have to dig it themselves).

wojtek-t: ideally with a pointer to documentation?

jkaniuk: I have not found anything relevant in GCP docs.

PR #16061, opened 2020-01-29
Kops - Disable NodeFeature:RuntimeHandler test for containerd

hakman:
This test fails in all containerd setups. The only reference I could find is this:
https://github.com/kubernetes/kubernetes/issues/78249


rifelpet:
This test is skipped in sig-node’s containerd tests so it seems appropriate to follow suit. It’d be nice if we could get the test to pass though, if applicable.

/lgtm
/approve

BenTheElder:
IIRC the issue is this https://github.com/kubernetes-sigs/kind/blob/0d3780371091b2dc9ff6eea1b6054f14ff5d970a/images/base/files/etc/containerd/config.toml#L21-L24

you have to have special test-config in your cluster settings, it can pass on containerd, but it's perhaps not desirable for a tool like kops to be configuring this magic test handler

Issue #8422, opened 2018-06-21
required_reviewers on PRs


cblecker:
Proposal:
blunderbuss currently selects a certain number of reviewers from the reviewer pool. By design it doesn't select the same reviewers for every PR. However, there may be use cases (high sensitivity parts of the code base, such as API changes) where a person should be cc'd on every PR that touches that part of the code.

I'd like to suggest that the OWNERS file be extended with a field for required_reviewers. Github users/teams that are listed in that section should always be added to the review list, in addition to the normal rotating reviewers (not instead of).

/area prow

cblecker:
cc @BenTheElder @cjwagner @matthyx @fejta

BenTheElder:
Interesting. I've had one previous request for a more advanced version of this to detect changes to flags and CC a particular reviewer.

This sounds useful and pretty straightforward to implement. Should it also support regex &/or globs so we can require reviews for specific files?

cblecker:
Yes, I was thinking so. I believe blunderbuss supports this already, but I think the approvers plugin doesn't.

cjwagner:
The [`no_parent_owners` field](https://github.com/kubernetes/test-infra/blob/1d1789f127517ea6778ff6e8a0eed2f0a848f509/prow/repoowners/repoowners.go#L48). May be of interest. It doesn't exactly match the behavior or what you are suggesting, but it does facilitate 'locking down' a sub directory.

fejta:
This duplicate the behavior of CODE_OWNERS, does that matter?

cblecker:
@cjwagner Yeah that's not quite what I'm looking for. I don't want to short circuit either the reviewers or approvers that would normally be assigned -- this would be additive.

@fejta It *is* similar to the CODEOWNERS behaviour with two key differences:
- CODEOWNERS [must have write access](https://help.github.com/articles/about-codeowners/)
- CODEOWNERS aren't additive. It runs through the file top to bottom, and the last line that matches the file is chosen.

fejta:
> CODEOWNERS must have write access

👎 🙄 oh github

BenTheElder:
Wow, did not know about that write access requirement, that is a major point to mention next time someone asks about the difference...

matthyx:
Do we have a place to document these features? I saw @cjwagner speaking about doc improvement efforts during last SIG.

cjwagner:
[...]

PR #16065, opened 2020-01-30
Point docs to community version of same content.

evankanderson:
It turns out that there are two copies of this content (as highlighted in https://github.com/kubernetes/test-infra/issues/8422#issuecomment-404667828). Preferring the copy in community that documents the expected format of the OWNERS files a bit more explicitly.


Katharine:
I'm not sure this is the right thing to do - the Kubernetes documentation is very Kubernetes-specific, and references a bunch of concepts that aren't real (e.g. `emeritus` is not a real role that means anything, as far as the tooling is concerned).

Prow is not a Kubernetes-specific tool, so it should have more generic documentation.

spiffxp:
/close
These docs are specific to how `/approve` uses owners files, whereas the https://go.k8s.io/owners docs are k/k specific, and describe more than just approve's use of them.  I agree there is overlap, but removing these docs outright is not the way to go.

PR #16066, opened 2020-01-30
Kops - Add network plugins e2e periodics

hakman:


rifelpet:
we can ask in office hours on friday if we should add e2e jobs for kube-router and romana.
/lgtm
/approve

hakman:
> we can ask in office hours on friday if we should add e2e jobs for kube-router and romana.

Probably we should check if there are some more ways to stress the ones we already added with defaults also. For example, looking forward to add the new Debian Buster images.

rifelpet:
yes, and CoreDNS as well

PR #16067, opened 2020-01-30
Kops - Make periodic e2e more stable #1

hakman:


rifelpet:
I confirmed all of these docker images and release markers exist 👍 
/lgtm
/approve

Issue #15913, opened 2020-01-15
Allow Branches and SkipBranches for jobs in repo


eddycharly:
Currently, jobs defined in repo are not allowed to declare branches and skip branches patterns.

https://github.com/kubernetes/test-infra/blob/master/prow/config/inrepoconfig.go#L123-L140

This makes it difficult to version jobs that run only on specific branches.

I don't see any technical reason to forbid branches and skip branches for jobs in repo, allowing them would be useful to let the jobs evolve naturally in git, it could be modified, merged, reviewed, etc...

Maybe we could add a config element to enable/disable this behavior if needed.

stevekuznetsov:
Not sure I follow -- the branch the job is configured in is the branch on which it runs. The file in `master` applies to `master` and `master` only.

/assign @alvaroaleman

eddycharly:
Let’s put it simply, imagine you have a job that must run only on master.

When you create a branch from master, you’re not going to delete the job before pushing to the new branch right ?

It makes sense that the job exists in multiple branches even if it applies to only one.

eddycharly:
@stevekuznetsov is it more clear ? Shall I go ahead and open a pr ?

This makes the feature pretty hard to use, especially with postsubmit jobs when not using forks.

alvaroaleman:
Yeah, so while we initially assumed this doesn't make sense, Loodse also run into a need for this pretty quickly, as there the usecase of having a rolling release branch that gets forwarded from a different branch exists and the two do not have an identical job config.

IMHO we should loosen the validation and allow this, but make the `branches` property default to the current branch if unset. @stevekuznetsov do you have any objections to that?

CC @xrstf @kdomanski

eddycharly:
@alvaroaleman i took a little time to work on this.

> make the branches property default to the current branch if unset

Can you elaborate on this ? Defaulting to the current branch or leaving it unset should not change anything as the job is coming from the current branch anyway, am i right ?

alvaroaleman:
> Can you elaborate on this ? Defaulting to the current branch or leaving it unset should not change anything as the job is coming from the current branch anyway, am i right ?

The reason is that the validation verifies that there is only one job with a given name per org/repo/branch. But thinking about this, we do not have this info available anyways, so I guess making inrepoconfig jobs on any branch conflict with a job of the same name in the static config in any other branch is okay,

eddycharly:
Thanks for your explanations.
[...]

PR #16068, opened 2020-01-30
Relax inrepoconfig branches and skipbranches validation

eddycharly:
This pull requests allows specifying branches and skip branches in prow jobs stored in a repository.

It fixes the issue discussed here https://github.com/kubernetes/test-infra/issues/15913

### DIFF
--- a/prow/config/inrepoconfig_test.go
+++ b/prow/config/inrepoconfig_test.go
@@ -122,33 +122,19 @@ func TestDefaultProwYAMLGetter(t *testing.T) {
 			},
 		},
 		{
-			name: "Branchconfig on presubmit is not allowed",
+			name: "Branchconfig on presubmit is allowed",
 			baseContent: map[string][]byte{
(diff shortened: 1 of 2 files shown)


prow/config/inrepoconfig_test.go
```diff
@@ -122,33 +122,19 @@ func TestDefaultProwYAMLGetter(t *testing.T) {
...
-				if err.Error() != expectedErrMsg {
-					return fmt.Errorf("expected error message to be %q, was %q", expectedErrMsg, err.Error())
+			validate: func(p *ProwYAML, err error) error {
+				if err != nil {
+					return fmt.Errorf("unexpected error: %v", err)
 				}
-				return nil
-			},
-		},
-		{
```
alvaroaleman: In the future, please do not just remove testcases because they happen to be implemented in a way that doesn't work anymore.

The test in question is not super important though, so lets get this in.
/lgtm
/approve

eddycharly: For the `Branchconfig on presubmit is not allowed` test it is not pertinent anymore as i removed the constraint.
For the `Multiple errors are aggregated` test i can put it back if you want.

Thanks

PR #16069, opened 2020-01-30
Kops - Make tests more stable #2

hakman:


config/jobs/kubernetes/kops/kops-config.yaml
```diff
@@ -15,7 +15,7 @@ presubmits:
     path_alias: k8s.io/kops
     spec:
       containers:
-      - image: gcr.io/k8s-testimages/kubekins-e2e:v20200128-8dae294-experimental
+      - image: gcr.io/k8s-testimages/kubekins-e2e:v20200128-8dae294-master
```
rifelpet: so this will downgrade our bazel version:

https://github.com/kubernetes/test-infra/blob/master/images/kubekins-e2e/variants.yaml#L6-L12

and kops is already "trying" to use an even newer version than both of these images:

https://github.com/kubernetes/kops/blob/master/.bazelversion

so i wonder if the jobs will still succeed with this older version of bazel?

config/jobs/kubernetes/kops/kops-config.yaml
```diff
@@ -116,7 +116,7 @@ presubmits:
     path_alias: k8s.io/kops
     spec:
       containers:
-      - image: gcr.io/k8s-testimages/kubekins-e2e:v20200128-8dae294-experimental
+      - image: gcr.io/k8s-testimages/kubekins-e2e:v20200128-8dae294-1.15
```
rifelpet: we should definitely use these versioned tags for all the jobs that target release branches 👍

config/jobs/kubernetes/kops/kops-config.yaml
hakman: Let's try. We can quickly revert this change if needed.
Or, we can keep the `experimental` for everything non e2e.

rifelpet:
ok, im fine with trying it out. once this merges we'll need to `/test all` some open kops PRs to test it out.

/approve
/lgtm

PR #16071, opened 2020-01-30
Add a --shutdown-bazel-pre-stage flag to kubetest

spiffxp:
kubetest currently calls out to kubernetes/release/push-build.sh for its
stage step, which ends up calling kubernetes/kubernetes'
"bazel run //:push-build" target

Unfortunately, in CI, bazel is occasionally getting abruptly killed
during this step. My working theory is that bazel is getting
oomkilled during this step because of memory it's already consumed during
build beforehand. This is happening even in RBE jobs that should be
offloading this to RBE.

Let's add the option to kill bazel before attempting stage, so that it
has as much room to grow its memory as possible

ref: https://github.com/kubernetes/kubernetes/issues/87441


spiffxp:
/close
Closing as this has been obviated by https://github.com/kubernetes/kubernetes/pull/87866

PR #16078, opened 2020-01-31
Kops - Use a newer image with RHEL 8 for network plugins

hakman:
We already have a test that passes when using an image with a newer kernel.
Maybe this is all we need for fixing at least some of the network plugins tests.


rifelpet:
Attempting to get our remaining cni jobs to pass as well as the amazon Linux 2 job.

/approve
/lgtm

PR #16084, opened 2020-01-31
Kops - Fix ssh user for RHEL8 image

hakman:
We also reduce the interval to 1 hour and will change it to 4 or 8 hours later today.


rifelpet:
/lgtm
/approve

We’ll set the interval back to a reasonable amount after today’s kops office hours

PR #16097, opened 2020-02-01
Add failure hints to the summary/metadata lens

Katharine:
Use some heuristics on pod state and events to attempt to provide useful information about job misconfigurations.

Hints include:
- Trying to pull an inaccessible or nonexistent image
- Misconfigured secret/configmap/volume references
- Unsatisfiable taints/tolerances/node selectors or resource requests
- Various types of node failure we've seen before (can't create sandbox, etc.)

If you have more ideas, please suggest them :D

An example:
image

/cc @BenTheElder @cjwagner

### DIFF
--- a/prow/spyglass/lenses/metadata/lens.go
+++ b/prow/spyglass/lenses/metadata/lens.go
@@ -20,6 +20,9 @@ package metadata
 import (
 	"bytes"
 	"encoding/json"
+	"regexp"
+	"sort"
+	"strings"
 	"time"
 
 	"fmt"
@@ -28,6 +31,8 @@ import (
 
 	"github.com/GoogleCloudPlatform/testgrid/metadata"
 	"github.com/sirupsen/logrus"
+	v1 "k8s.io/api/core/v1"
+	k8sreporter "k8s.io/test-infra/prow/crier/reporters/gcs/kubernetes"
 	"k8s.io/test-infra/prow/pod-utils/gcs"
 	"k8s.io/test-infra/prow/spyglass/lenses"
 )
@@ -81,6 +86,7 @@ func (lens Lens) Body(artifacts []lenses.Artifact, resourceDir string, data stri
 		StartTime    time.Time
 		FinishedTime time.Time
 		Elapsed      time.Duration
+		Hint         string
 		Metadata     map[string]interface{}
 	}
 	metadataViewData := MetadataViewData{Status: "Pending"}
@@ -91,12 +97,13 @@ func (lens Lens) Body(artifacts []lenses.Artifact, resourceDir string, data stri
 		if err != nil {
 			logrus.WithError(err).Error("Failed reading from artifact.")
 		}
-		if a.JobPath() == "started.json" {
+		switch a.JobPath() {
+		case "started.json":
 			if err = json.Unmarshal(read, &started); err != nil {
 				logrus.WithError(err).Error("Error unmarshaling started.json")
 			}
 			metadataViewData.StartTime = time.Unix(started.Timestamp, 0)
-		} else if a.JobPath() == "finished.json" {
+		case "finished.json":
 			if err = json.Unmarshal(read, &finished); err != nil {
 				logrus.WithError(err).Error("Error unmarshaling finished.json")
 			}
@@ -104,6 +111,8 @@ func (lens Lens) Body(artifacts []lenses.Artifact, resourceDir string, data stri
 				metadataViewData.FinishedTime = time.Unix(*finished.Timestamp, 0)
 			}
 			metadataViewData.Status = finished.Result
+		case "podinfo.json":
+			metadataViewData.Hint = hintFromPodInfo(read)
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


prow/spyglass/lenses/metadata/lens.go
```diff
@@ -137,6 +146,79 @@ func (lens Lens) Body(artifacts []lenses.Artifact, resourceDir string, data stri
 	return buf.String()
 }
 
+var failedMountRegex = regexp.MustCompile(`MountVolume.SetUp failed for volume "(.+?)" : (.+?)`)
+
+func hintFromPodInfo(buf []byte) string {
```
cjwagner: Please unit test this function for the types of errors that are exposed and for happy cases like running and successful pods.

prow/spyglass/lenses/metadata/lens.go
```diff
@@ -137,6 +146,79 @@ func (lens Lens) Body(artifacts []lenses.Artifact, resourceDir string, data stri
...
+					return fmt.Sprintf("The pod could not start because it could not mount the volume %q: %s", results[1], results[2])
+				}
+			}
+		}
+		if failedMount {
+			return fmt.Sprintf("The job could not start because one or more of the volumes could not be mounted.")
+		}
+	}
+	// Check if we cannot be scheduled
+	// This is unlikely - we only outright fail if a pod is actually scheduled to a pod it does not support
```
cjwagner:
```suggestion
	// This is unlikely - we only outright fail if a pod is actually scheduled to a node that can't support it.
```

prow/spyglass/lenses/metadata/lens.go
```diff
@@ -137,6 +146,79 @@ func (lens Lens) Body(artifacts []lenses.Artifact, resourceDir string, data stri
 	return buf.String()
 }
 
+var failedMountRegex = regexp.MustCompile(`MountVolume.SetUp failed for volume "(.+?)" : (.+?)`)
+
+func hintFromPodInfo(buf []byte) string {
+	var report k8sreporter.PodReport
+	if err := json.Unmarshal(buf, &report); err != nil {
+		// This error isn't worth highlighting here, and will come up again
+		// elsewhere.
```
cjwagner: Can you explain this more, please? If the `podinfo.json` exists, but is empty or otherwise unparseable that seems like something we should expose. I don't see where this "will come up again elsewhere" and we should probably at least log the error.

prow/spyglass/lenses/metadata/lens.go
```diff
@@ -137,6 +146,79 @@ func (lens Lens) Body(artifacts []lenses.Artifact, resourceDir string, data stri
...
+	// Usually we would fail to schedule it at all, so it will be pending forever.
+	if report.Pod.Status.Phase == v1.PodPending {
+		for _, e := range report.Events {
+			if e.Reason == "FailedScheduling" {
+				return fmt.Sprintf("There are no nodes that your pod can schedule to - check your requests, tolerations, and node selectors (%s)", e.Message)
+			}
+		}
+	}
+
+	// There are a bunch of fun ways for the node to fail that we've seen before
```
cjwagner: We probably should also gate this with a check that the pod is `pending` or `failed`. Some of these might be transient errors that pods can recover from. The same applies to the ImagePullBackoff check above.
Alternatively, above where we abort if the pod phase is `success`, we could also abort if the pod phase is `running` so long as none of the errors we want to alert on present themselves on `running` pods.

prow/spyglass/lenses/metadata/lens.go
Katharine: It is not a critical test failure if the file is unreadable (which is what we're using it for here) — this will be surfaced when the podinfo lens attempts to render.

We can log the error at some log level that we will subsequently never see, but I'm not sure how valuable that actually is.

cjwagner: >  this will be surfaced when the podinfo lens attempts to render.

👍 
> We can log the error at some log level that we will subsequently never see, but I'm not sure how valuable that actually is.

Its valuable if we end up hitting the error case unexpectedly, but I agree that seems unlikely and the error will be logged by the podinfo lens if it is enabled. 
/shrug

prow/spyglass/lenses/metadata/lens.go
Katharine: We only report podinfo.json for completed jobs. If your job has completed but your pod is still running, something has probably gone wrong, at which point it may be worth reporting on these failures even if the pod did start (but not end) successfully.

I don't feel terribly strongly about this one way or the other, though.

re ImagePullBackoff — in that case we're checking the _current_ state of the containers (at time of job termination), not the historical events, so it is definitionally impossible for the pod to be running while the containers are still in ImagePullBackoff.

prow/spyglass/lenses/metadata/lens.go
```diff
@@ -137,6 +146,79 @@ func (lens Lens) Body(artifacts []lenses.Artifact, resourceDir string, data stri
...
+
+func hintFromPodInfo(buf []byte) string {
+	var report k8sreporter.PodReport
+	if err := json.Unmarshal(buf, &report); err != nil {
+		// This error isn't worth highlighting here, and will come up again
+		// elsewhere.
+		return ""
+	}
+
+	// We're more likely to pick a relevant event if we use the last ones first.
```
stevekuznetsov: Aren't they already sorted in the other direction?

prow/spyglass/lenses/metadata/lens.go
```diff
@@ -137,6 +146,79 @@ func (lens Lens) Body(artifacts []lenses.Artifact, resourceDir string, data stri
...
+			if e.Reason == "FailedScheduling" {
+				return fmt.Sprintf("There are no nodes that your pod can schedule to - check your requests, tolerations, and node selectors (%s)", e.Message)
+			}
+		}
+	}
+
+	// There are a bunch of fun ways for the node to fail that we've seen before
+	for _, e := range report.Events {
+		if e.Reason == "FailedCreatePodSandbox" || e.Reason == "FailedSync" || e.Reason == "FailedKillPod" {
+			return "The job may have executed on an unhealthy node. Contact your prow maintainers or check the detailed pod information."
```
stevekuznetsov: If it's knowable here, adding a link to this message to the pod info (or directing users to send the link) would be useful to help the on-call get the information they need and not some vague "a node was bad" message

prow/spyglass/lenses/metadata/lens.go
```diff
@@ -137,6 +146,79 @@ func (lens Lens) Body(artifacts []lenses.Artifact, resourceDir string, data stri
...
+		}
+	}
+
+	// We've got nothing.
+	return ""
```
stevekuznetsov: We could perhaps change this to be something like

```
No known issues were found by inspecting the job resources. <some blurb about how to constructively add new cases or make an issue to track new cases>
```

prow/spyglass/lenses/metadata/lens.go
Katharine: In practice yes, but I am not sure whether any guarantee of that exists, so I chose to sort them rather than just flip the list.

prow/spyglass/lenses/metadata/lens.go
Katharine: In almost all cases when a job fails, it's because the test failed. I don't think it's useful for us to suggest a job issue or infrastructure failure in the UI in the case where we don't see anything obviously wrong - it'll mostly just be because their code was bad.

prow/spyglass/lenses/metadata/lens.go
Katharine: Done.

prow/spyglass/lenses/metadata/lens.go
Katharine: Done.

prow/spyglass/lenses/metadata/lens.go
Katharine: Added a log.

Katharine:
There is now an example in the PR description.

prow/spyglass/lenses/metadata/lens_test.go
```diff
@@ -95,3 +101,242 @@ func TestFlattenMetadata(t *testing.T) {
...
+					},
+					Status: v1.PodStatus{
+						Phase:  v1.PodFailed,
+						Reason: "MatchNodeSelector",
+					},
+				},
+			},
+		},
+		{
+			name:     "pod scheduled to an illegal node is reported",
```
cjwagner: nit: Duplicated test name.

prow/spyglass/lenses/metadata/lens.go
stevekuznetsov: Sorry, I thought we'd hit this case when the obvious case (test failed) was not what happened. I understand your point

PR #16103, opened 2020-02-03
Use newest gcloud build in gce-cos-1.16-scalability-100 job

mm4tt:
Ref. https://github.com/kubernetes/perf-tests/issues/1005


wojtek-t:
Let's try

/lgtm
/approve

PR #16107, opened 2020-02-03
Extend retryCmd with configurable retry and bo factor

clarketm:
Modify `retryCmd` so that it accepts a number of `tries` and a `backoffFactor`. 

> **Note:** Using `retries=3` and `backoffFactor=1` will retain the original behavior of `retryCmd` (i.e.  base 2 exponential). In this PR, I am also increasing the *tries* to **5** in an attempt to mitigate the issue described: #16130

### DIFF
--- a/prow/git/git.go
+++ b/prow/git/git.go
@@ -23,6 +23,7 @@ import (
 	"errors"
 	"fmt"
 	"io/ioutil"
+	"math"
 	"os"
 	"os/exec"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


clarketm:
/hold

I received a response on the GitHub support ticket I filed for the `"not our ref"` issue that drove this change. I will open an issue to summarize the problem and track the alternatives given for mitigation (including increasing retries), so the ultimate solution can be informed and objective.

clarketm:
Can we approve/merge this as mitigation until a [more permanent solution](https://github.com/kubernetes/test-infra/issues/16130#issuecomment-586503546) is implemented?

PR #16108, opened 2020-02-03
Add rbac rules so crier can read pods and events

Katharine:
I have only moderate confidence in the correctness of this.

/cc @BenTheElder

### DIFF
--- a/prow/cluster/crier_rbac.yaml
+++ b/prow/cluster/crier_rbac.yaml
@@ -33,6 +33,14 @@ rules:
     - "watch"
     - "list"
     - "patch"
+- apiGroups:
+    - ""
+  resources:
+    - "pods"
+    - "events"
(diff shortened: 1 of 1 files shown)


prow/cluster/crier_rbac.yaml
```diff
@@ -33,6 +33,14 @@ rules:
...
     - "list"
     - "patch"
+- apiGroups:
+    - ""
+  resources:
+    - "pods"
+    - "events"
+  verbs:
+    - "get"
+    - "list"
```
BenTheElder: do we even have RBAC in prow.k8s.io in either cluster?

Katharine: No (which is why it works), but these files are the basis for the rbac setup used in other clusters.

PR #16114, opened 2020-02-04
Kops - Validate cluster before starting e2e

hakman:
Kops should validate the cluster before starting e2e tests. There are cases when nodes are in `Ready` state, but some core components are failing:
https://github.com/kubernetes/kops/pull/8454#issuecomment-581462566


rifelpet:
/lgtm
we can add "consecutive successes" logic in a followup PR once we confirm this works on its own.

PR #16128, opened 2020-02-04
Configurator: sort periodics, presubmits, and postsubmits before write

### LINKED ISSUE #13624: configurator output order is non deterministic
**What happened**:
From this PR: https://github.com/istio/test-infra/pull/1542/files

I ran `make testgrid`, running `configurator --prow-config prow/config.yaml --prow-job-config prow/cluster/jobs --output-yaml --yaml testgrid/default.yaml --oneshot --output testgrid/istio-generated.yaml`. Each time I run it the order is different

Output is the same every time I run it

Checkout the PR and run `make testgrid` a couple times

cc @chases2

clarketm:
fix #13624


testgrid/cmd/configurator/prow.go
```diff
@@ -19,13 +19,13 @@ package main
...
 	"fmt"
 	"path"
+	"sort"
 	"strconv"
 	"strings"
 
 	"github.com/GoogleCloudPlatform/testgrid/config"
 	"github.com/GoogleCloudPlatform/testgrid/config/yamlcfg"
 	configpb "github.com/GoogleCloudPlatform/testgrid/pb/config"
-
```
chases2: nit: I think this spacing between the GCP imports and the k8s imports is desired

testgrid/cmd/configurator/prow.go
```diff
@@ -173,32 +173,85 @@ func applySingleProwjobAnnotations(c *configpb.Configuration, pc *prowConfig.Con
...
 		if err := applySingleProwjobAnnotations(c, pc, j.JobBase, prowapi.PeriodicJob, "", reconcile); err != nil {
 			return err
 		}
 	}
 
-	for repo, js := range jobs.PostsubmitsStatic {
-		for _, j := range js {
-			if err := applySingleProwjobAnnotations(c, pc, j.JobBase, prowapi.PostsubmitJob, repo, reconcile); err != nil {
+	post := jobs.PostsubmitsStatic
+	postK := sortPostsubmits(post)
```
chases2: naming nit: what does `K` stand for in this context?

alternately, `post := sortPostsubmits(jobs.PostsubmitsStatic)`

clarketm: It signifies `keys` (for the map keys). I will rename it to `postKeys` to be more explicit. Also, `post` ->` map[string][]prowConfig.Presubmit` and `postK` -> `[]string` so I cannot collapse the assignment into one; both references are used later in the loop (`preK` for key ordering and `post` for retrieving the value for that key).

testgrid/cmd/configurator/prow_test.go
```diff
@@ -571,6 +570,231 @@ func Test_applySingleProwjobAnnotation_WithDefaults(t *testing.T) {
...
+			a := test.initialJobConfig
+			e := test.expectedJobConfig
+
+			aPer := a.AllPeriodics()
+			ePer := e.AllPeriodics()
+			sortPeriodics(aPer)
+			if len(aPer) != len(ePer) {
+				t.Fatal("Periodic lengths do not match")
+			}
+			for i, actualPer := range aPer {
```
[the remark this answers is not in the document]
clarketm: > Would `reflect.DeepEqual(expected, actual)` work?

It will for the individual arrays. Updated to use `DeepEqual`.

> Also consider a few test cases to check 1) if jobs sort right, 2) if repositories sort right, and 3) the function doesn't crash if given no jobs.
>
> That may give you a clearer signal for what went wrong without having to maintain complicated "if-then" equality checking.

I subdivided the test cases by `periodics`, `presubmits`, and `postsubmits` as suggested covering cases where **1** jobs sort, **2** repos sort, **3** no crash on empty.

PR #16131, opened 2020-02-05
Add resource requests for pd driver CI to speed up Kubernetes builds

davidz627:
/assign @msau42 @sebastienvas 

Fixes: https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/issues/464


msau42:
/lgtm

This is fine to solve the immediate problem, but longer term I think we don't need to actually build K8s in our jobs.

PR #16136, opened 2020-02-05
Publishing arm64 conformance test results in conformance-arm page

lubinsz:
I think it is more appropriate to move the arm64-conformance
logs to conformance-arm webpage, so as not to be misunderstood.


lubinsz:
cc @BenTheElder

fejta:
/retest
ref https://github.com/kubernetes/test-infra/pull/15955

PR #16142, opened 2020-02-05
Kops - Skip test "Services should be rejected when no endpoints exist"

hakman:
`[sig-network] Services should be rejected when no endpoints exist` is failing for Calico, Cilium and Weave. Seems to be because these are not service aware. In any case, nothing we can do about it for now.

https://github.com/cilium/cilium/issues/10002
https://github.com/projectcalico/calico/issues/1055


rifelpet:
/lgtm
/approve

We'll keep an eye on those open issues and revert this when we are able.

PR #16143, opened 2020-02-05
Verify GitHub SSH key fingerprint in clonerefs

### LINKED ISSUE #9638: prow should ship github's fingerprint not skip key checking
follow up to https://github.com/kubernetes/test-infra/issues/9440

github has a well known SSH fingerprint. we should ship it in prow as a default, and we should at least optionally enable host key checking

/area prow
/kind bug

/priority important-soon

/cc @fejta

matthyx:
Fixes #9638


stevekuznetsov:
@matthyx looks like you need to `hack/update-bazel.sh` !

BenTheElder:
/lgtm
Thank you!

matthyx:
Will fix after fitness!

alvaroaleman:
/hold
To prevent bot spam

matthyx:
@alvaroaleman could you?

PR #16149, opened 2020-02-05
Azure: add MSI role assignment logic after cluster creation

chewong:
Follow-up PR for #16129. This PR assigns 'Owner' role to the user identity so that it can pull images from private registry within the subscription.

/assign @feiskyer 
/cc @andyzhangx

### DIFF
--- a/kubetest/azure_helpers.go
+++ b/kubetest/azure_helpers.go
@@ -23,10 +23,22 @@ import (
 	"os"
 	"time"
 
-	resources "github.com/Azure/azure-sdk-for-go/services/resources/mgmt/2018-05-01/resources"
+	"github.com/Azure/azure-sdk-for-go/services/authorization/mgmt/2015-07-01/authorization"
+	"github.com/Azure/azure-sdk-for-go/services/preview/msi/mgmt/2015-08-31-preview/msi"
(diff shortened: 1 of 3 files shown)


kubetest/azure_helpers.go
```diff
@@ -282,6 +296,31 @@ func (az *AzureClient) DeleteResourceGroup(ctx context.Context, groupName string
...
+	roleDefinitionID := fmt.Sprintf(aadRoleReferenceTemplate, az.subscriptionID, aadOwnerRoleID)
+	scope := fmt.Sprintf(aadRoleResourceGroupScopeTemplate, az.subscriptionID)
+	roleAssignmentParameters := authorization.RoleAssignmentCreateParameters{
+		Properties: &authorization.RoleAssignmentProperties{
+			RoleDefinitionID: stringPointer(roleDefinitionID),
+			PrincipalID:      stringPointer(identityPrincipalID),
+		},
+	}
+
+	if _, err := az.authorizationClient.Create(ctx, scope, uuid.NewV1().String(), roleAssignmentParameters); err != nil {
```
feiskyer: have we checked our e2e credentials have the permission to do the role assignment?

chewong: I assigned 'Owner' role to the service principal in prow.

Issue #16001, opened 2020-01-23
boskos: merge Prometheus resource metrics into boskos core


ixdy:
**What should be cleaned up or changed**:
On k8s prow, the Boskos service runs inside the Prow builds cluster. 
Back in 2017, we wanted to collect metrics on Boskos resources to show in Velodrome, so it was necessary to expose these metrics on an external IP, since the Velodrome collectors did not run inside the builds cluster.

To accomplish this, https://github.com/kubernetes/test-infra/pull/3971 added the boskos-metrics component, which both proxied the `/metric` API on `/`, as well as exposing Prometheus-style metrics on `/prometheus`.

Since then, we've switched to serving Prometheus-style metrics on port 9090, and since #15356 have not even exposed the `/metric` proxy, but nobody has seemed to complain (so maybe it's not used?). We're now collecting metrics using the Prow monitoring stack, but since that runs in a separate cluster, we still need to expose the metrics via a public IP.

At the same time, the main Boskos deployment is also serving its own Prometheus metrics on port 9090. In #15990, I'm adding a new Service that exposes these metrics. Rather than running a separate `boskos-metrics` component, we could just move these resource metrics into the Prometheus metrics in boskos core, simplifying deployments slightly.

(I also intend to serve metrics on janitors. We might want to configure some sort of cluster-local aggregator for these, since they'll also be private to the cluster.)

/area boskos

cc @stevekuznetsov @krzyzacy is there something obvious I'm missing here?

ixdy:
regarding metric "aggregation" - maybe [federation](https://prometheus.io/docs/prometheus/latest/federation/) is relevant?

stevekuznetsov:
You could do that, yes, or configure for Prometheus in the build cluster to scrape everything there and `remote_write` to push out to something public./

ixdy:
I think this is actually done, though I haven't added metrics to the janitors yet.

/close

Issue #14715, opened 2019-10-10
monitor boskos cleanup timing


BenTheElder:
**What would you like to be added**: export and graph metrics for boskos cleanup timing

**Why is this needed**: so we can determine if this is increasing and we need to increase the janitor or fix boskos xref #14697 

Possibly this should also move to the new monitoring stack? cc @cjwagner @detiber 

/area boskos
/assign @krzyzacy 
cc @fejta @mm4tt

cjwagner:
/cc @chaodaiG 
Is also working on boskos metric using the new monitoring stack.

stevekuznetsov:
We have boskos metrics using the new stack... what do you mean? We should sync this week.

ixdy:
Boskos is now using the Prow monitoring stack (#15344). I guess next thing is to expose cleanup timing metrics?

ixdy:
@BenTheElder if you remember, is the intent here to track how long resources are in a given state (e.g. how long something remains as dirty before transitioning to free)? Or how long the actual cleanup itself takes? (Maybe both?)

BenTheElder:
IIRC how long the cleanup takes, which we sometimes have issues with, but
both would be useful

ixdy:
Moving this issue to https://github.com/kubernetes-sigs/boskos/issues/13.
/close

PR #16152, opened 2020-02-05
Refactor Boskos directory structure

ixdy:
The Boskos directory had main packages mixed in with library packages. I've refactored most of the main methods into a `cmd/` subdirectory, similar to what is done with Prow. (I have not touched the aws-janitor, at least not yet, because it has its own internal structure.)

Besides making everything a bit easier to maintain and understand, this is a precursor to some additional things I'm working on:
- I want to add metrics to the janitors (#14715), including the AWS janitor, but the existing structure didn't offer a good place to put common code.
- I want to simplify how core resource metrics are exposed (#16001), but again, there's no good place to refactor the metrics currently.

/assign @stevekuznetsov @sebastienvas @krzyzacy


ixdy:
@krzyzacy thanks for the fast review!

I'm not sure what else there is to be done with mason right now.

I moved the `fake-mason` command into `cmd`, and what remains is just the interface that implementations implement.

The actual GCP implementation still lives in istio/test-infra (https://github.com/istio/test-infra/tree/master/boskos). I'm not sure why it wasn't moved along with the mason interface - maybe @sebastienvas or @chases2 knows.

ixdy:
/hold cancel

Rebased. Can I get a re-LGTM before things change again?

stevekuznetsov:
cc @alvaroaleman

PR #16155, opened 2020-02-05
Update concurrency for CAPA jobs that use Boskos

detiber:
- Try to avoid multiple quick PRs getting merged from exhausting our Boskos pool of resources

/assign @ncdc


ncdc:
/lgtm
/approve

thanks!

PR #16158, opened 2020-02-05
hook: expose metrics for plugin handle times and error rates

stevekuznetsov:

### DIFF
--- a/prow/hook/events.go
+++ b/prow/hook/events.go
@@ -17,6 +17,9 @@ limitations under the License.
 package hook
 
 import (
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
 	"github.com/sirupsen/logrus"
 
 	"k8s.io/test-infra/prow/github"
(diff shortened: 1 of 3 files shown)


prow/hook/events.go
```diff
@@ -77,9 +80,13 @@ func (s *Server) handleReviewEvent(l *logrus.Entry, re github.ReviewEvent) {
 				re.Repo.Name,
 				re.PullRequest.Number,
 			)
+			start := time.Now()
+			labels := prometheus.Labels{"event_type": l.Data["event-type"].(string), "action": string(re.Action), "plugin": p}
```
stevekuznetsov: Forgive me for the sin ... but threading through the event type data would be challenging otherwise

alvaroaleman: Can we then at least make the key a const?

stevekuznetsov:
@alvaroaleman updated

PR #16162, opened 2020-02-06
Make high-density job use forked density config

mm4tt:
Ref. https://github.com/kubernetes/perf-tests/issues/1007

/hold

Wait for https://github.com/kubernetes/perf-tests/pull/1026

### DIFF
--- a/config/jobs/kubernetes/sig-scalability/sig-scalability-periodic-jobs.yaml
+++ b/config/jobs/kubernetes/sig-scalability/sig-scalability-periodic-jobs.yaml
@@ -315,15 +315,11 @@ periodics:
       - --test-cmd-args=--nodes=600
       - --test-cmd-args=--provider=kubemark
       - --test-cmd-args=--report-dir=/workspace/_artifacts
-      - --test-cmd-args=--testconfig=testing/density/config.yaml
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes/sig-scalability/sig-scalability-periodic-jobs.yaml
```diff
@@ -315,15 +315,11 @@ periodics:
       - --test-cmd-args=--nodes=600
       - --test-cmd-args=--provider=kubemark
       - --test-cmd-args=--report-dir=/workspace/_artifacts
-      - --test-cmd-args=--testconfig=testing/density/config.yaml
+      # TODO(https://github.com/kubernetes/perf-tests/issues/1007): load test should be used to test high-density.
```
wojtek-t: Since you're touching it - can you also open an issue to setup regular GCE high-density test (so that we not only have kubemark)?
[I guess we may need bigger nodes or sth, that's why we shouldn't do that in this PR]

mm4tt: Done - https://github.com/kubernetes/perf-tests/issues/1028

PR #16163, opened 2020-02-06
GitHub client: Log request durations

alvaroaleman:
This PR is a somewhat naive attempt at making the github client log the durations of all its methods.

### DIFF
--- a/prow/github/client.go
+++ b/prow/github/client.go
@@ -515,15 +515,19 @@ func NewFakeClient() Client {
 	}
 }
 
-func (c *client) log(methodName string, args ...interface{}) {
+func (c *client) log(methodName string, args ...interface{}) (logDuration func()) {
 	if c.logger == nil {
-		return
+		return func() {}
 	}
 	var as []string
 	for _, arg := range args {
 		as = append(as, fmt.Sprintf("%v", arg))
 	}
+	start := time.Now()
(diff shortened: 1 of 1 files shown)


prow/github/client.go
```diff
@@ -515,15 +515,19 @@ func NewFakeClient() Client {
...
+		return func() {}
 	}
 	var as []string
 	for _, arg := range args {
 		as = append(as, fmt.Sprintf("%v", arg))
 	}
+	start := time.Now()
 	c.logger.Infof("%s(%s)", methodName, strings.Join(as, ", "))
+	return func() {
+		c.logger.Debugf("%s(%s) duration: %s", methodName, strings.Join(as, ", "), time.Since(start).String())
```
alvaroaleman: Theoretically it would make sense to record metrics here, but with the current variadic arg approach this will lead to a huge cardinality

stevekuznetsov: nit: can we `c.logger.WithField("duration", time.Since(start).String()).Debugf(`

stevekuznetsov: And maybe then edit the `Debugf(` to have a `"finished"` message to make it visually distinct?

alvaroaleman: Thats what I did initially but I was afraid having two identical log messages would confuse aggregation. WDYT?

Issue #15985, opened 2020-01-22
boskos leased the same resource to two owners


ixdy:
**What happened**: Openshift reported that Boskos leased the same resource to two different jobs: https://github.com/openshift/release/issues/6795

We have a gratuitous amount of locking, so in theory this shouldn't happen, but I think I may have figured out what may have happened:

1. `/aquire` is called by two different jobs at the same time. We have a [global lock](https://github.com/kubernetes/test-infra/blob/5e7071bdbd0408d480d7cab02a5ff0757e945dae/boskos/ranch/ranch.go#L125), so at least one of these waits while the other proceeds.
1. The first job [gets all resources](https://github.com/kubernetes/test-infra/blob/5e7071bdbd0408d480d7cab02a5ff0757e945dae/boskos/ranch/ranch.go#L133), which involves [listing](https://github.com/kubernetes/test-infra/blob/5e7071bdbd0408d480d7cab02a5ff0757e945dae/boskos/crds/crd_storage.go#L74) all resources on the Kubernetes API server, then [converting](https://github.com/kubernetes/test-infra/blob/5e7071bdbd0408d480d7cab02a5ff0757e945dae/boskos/crds/crd_storage.go#L80) to `common.Item`.
1. It then selects a free resource, fills in fields, and [calls](https://github.com/kubernetes/test-infra/blob/5e7071bdbd0408d480d7cab02a5ff0757e945dae/boskos/ranch/ranch.go#L154) `r.Storage.UpdateResource`.
1. `UpdateResource` in turn [calls](https://github.com/kubernetes/test-infra/blob/5e7071bdbd0408d480d7cab02a5ff0757e945dae/boskos/ranch/storage.go#L101) `s.resources.Update`.
1. The CRD-based `Update` method first [gets](https://github.com/kubernetes/test-infra/blob/5e7071bdbd0408d480d7cab02a5ff0757e945dae/boskos/crds/crd_storage.go#L53) the current object from the Kubernetes API server, then [updates fields of this object locally using the item](https://github.com/kubernetes/test-infra/blob/5e7071bdbd0408d480d7cab02a5ff0757e945dae/boskos/crds/crd_storage.go#L57) before [updating the object](https://github.com/kubernetes/test-infra/blob/5e7071bdbd0408d480d7cab02a5ff0757e945dae/boskos/crds/crd_storage.go#L58) on the Kubernetes API server.
1. Assuming the update was successful, we release the lock and return the updated resource.

[...]

PR #16165, opened 2020-02-06
Boskos: remove global lock

alvaroaleman:
Boskos always uses a CRD storage, which means that we get the
built-in optimistic concurrency from Kubernetes. This allows us
to get rid of the lock, as the Update requests we do will fail in
case another request was quicker and updated the object in the meantime.

From a users POV this means that there is a higher chance of
Acquire/AcquireByState requests failing, but a much faster response
time.

As a follow-up, we could make the Acquire/AcquireByState handlers retry
the whole allocation to reduce the likelyhood of transient failures
there.

/assign sebastienvas @stevekuznetsov

### DIFF
--- a/boskos/boskos.go
+++ b/boskos/boskos.go
@@ -139,6 +139,8 @@ func main() {
 	v.WatchConfig()
 	v.OnConfigChange(func(in fsnotify.Event) {
 		logrus.Infof("Updating Boskos Config")
+		// TODO: We need retrying here, as we do not delete
+		// resources that currently have an owner
 		if err := r.SyncConfig(*configPath); err != nil {
 			logrus.WithError(err).Errorf("Failed to update config")
 		} else {
--- a/boskos/ranch/ranch.go
+++ b/boskos/ranch/ranch.go
@@ -19,7 +19,6 @@ package ranch
 import (
 	"fmt"
 	"strings"
-	"sync"
 	"time"
 
 	"github.com/sirupsen/logrus"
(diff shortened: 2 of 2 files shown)


alvaroaleman:
/hold
We actually short-circuit kubes optimistic concurrency by having a downstream `Update` implementation that does Get->Modify->Update. lol.

ixdy:
Yup, I think that was the root cause of https://github.com/kubernetes/test-infra/issues/15985. We have a global lock, and it doesn't really even help.

ixdy:
One thought I had (but haven't really fully fleshed out yet) is to keep track of the `resourceVersion` when converting from the Kubernetes object to the internal items. We might also need to add some sort of resource version for the internal memory-based storage, or deprecate/remove it.

stevekuznetsov:
I think we need to get rid of the internal version entirely and operate against an indexer/lister. Let the server do retries up to N so we do not need to change client behavior.

alvaroaleman:
> One thought I had (but haven't really fully fleshed out yet) is to keep track of the resourceVersion when converting from the Kubernetes object to the internal items. We might also need to add some sort of resource version for the internal memory-based storage, or deprecate/remove 

I would really like to get rid of the internal versions as well, this would allow us to just use a normal client and it does IMHO make the code much easier to understand.

alvaroaleman:
Closing this PR, we can not use it today as our Storage interface does not provide the same guarantees as the Kube api. We may reconsider it once https://github.com/kubernetes/test-infra/issues/16207 got fixed, but at that point re-creating it will be easier than rebasing.

PR #16168, opened 2020-02-06
build all tests and package in bazel-build presubmit

mikedanese:
See https://github.com/kubernetes/kubernetes/pull/85861#pullrequestreview-354529531 for something we missed.

### DIFF
--- a/config/jobs/kubernetes/sig-testing/bazel-build-test.yaml
+++ b/config/jobs/kubernetes/sig-testing/bazel-build-test.yaml
@@ -25,7 +25,7 @@ presubmits:
         args:
         - -c
         - |
-          ../test-infra/hack/bazel.sh build --config=remote --remote_instance_name=projects/k8s-prow-builds/instances/default_instance $(bazel query --keep_going --noshow_progress "kind(.*_binary, rdeps(//... -//vendor/..., //...)) except attr('tags', 'manual', //...)") //build/release-tars
+          ../test-infra/hack/bazel.sh build --config=remote --remote_instance_name=projects/k8s-prow-builds/instances/default_instance -- //... -//vendor/... -//build/... //build/release-tars
   - name: pull-kubernetes-bazel-test
     decorate: true
     always_run: true


mikedanese:
Master is broken in k/k.

/hold

BenTheElder:
FYI @dims @liggitt

config/jobs/kubernetes/sig-testing/bazel-build-test.yaml
```diff
@@ -25,7 +25,7 @@ presubmits:
         args:
         - -c
         - |
-          ../test-infra/hack/bazel.sh build --config=remote --remote_instance_name=projects/k8s-prow-builds/instances/default_instance $(bazel query --keep_going --noshow_progress "kind(.*_binary, rdeps(//... -//vendor/..., //...)) except attr('tags', 'manual', //...)") //build/release-tars
+          ../test-infra/hack/bazel.sh build --config=remote --remote_instance_name=projects/k8s-prow-builds/instances/default_instance -- //... -//vendor/... -//build/... //build/release-tars
```
BenTheElder: this target pattern is bad imo, it should just be `//...`
we will build any vendor targets that are pulled in via a useful target.
build and release-tars are under //...

mikedanese: vendor doesn't build for multiple reasons. the symlinks mess things up and some vendored deps use windows syscalls but don't have proper build tags.

BenTheElder: vendor is tagged manually intentionally so we don't build / run vendored tests.

we don't build windows with bazel? and we do run typecheck on all platforms.

mikedanese: e.g. https://github.com/kubernetes/kubernetes/blob/6858c25ee474ac2c90b49097c0ad04a61c6c54e1/vendor/github.com/Microsoft/hcsshim/internal/hcserror/hcserror.go#L33-L35

I'm fine with building all of //build/... though.

BenTheElder: er that is, if `hack/update-bazel.sh` is working properly we shouldn't need to exclude vendor, and I'm not sure we should exclude build either...?

mikedanese: hack/update-bazel.sh depends on build files being properly tagged for platforms. That file has neither a *_windows.go or // build -windows comment.

BenTheElder: :/

cblecker: Do we need to except the manual tag as in the previous query?

I wish we had a way to test this.

mikedanese: I'm not sure why we care if something has a manual tag. It should still build. I don't have context on why this was picked initially.

BenTheElder: the manual tag needed to be excluded before because we did a query and then passed the query results back I think. whereas if you tell `bazel build` the target is `//...` then I know it will ignore manual tags unless you expliclty do like `bazel build //... //my-manual-target`

BenTheElder: (or if the manual target is a dependency of a non manual target)

mikedanese: Ok, so what should I do to move this forward?

BenTheElder: let's start with this, but I think we probably need to fix the upstream build such that it doesn't need `-//vendor/...`

ideally we should switch to invoking a make alias here so we can correct the target list in the same Kubernetes PR

BenTheElder:
I agree in principle but have been a bit overloaded today (:fire: prow :fire:) so lending a
/approve
with the intention that anyone CCed can LG and merge this when ready.

mikedanese:
Tests look happy still:

https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/directory/pull-kubernetes-bazel-build/1225825638795972610

PR #16171, opened 2020-02-06
prow/interrupts: Do not handle the ABRT signal

smarterclayton:
ABRT is typically a debugging signal and is not used by Kube
during termination. By registering for it we hide the very useful
in-process "dump all goroutine state on a wedged go process"
default go runtime behavior. Since this package is included by init,
any of our tools that use it won't be able to leverage that
function.

Remove ABRT from being automatically handled, and allow prow
to crash on ABRT. A developer debugging one of these processes
would have to manually clean up anyway (or they could fire TERM
and then ABRT).

/assign @stevekuznetsov


matthyx:
I understand the reason behind.
/lgtm

PR #16173, opened 2020-02-06
fix busybox timeout arguments

BenTheElder:
/cc @cjwagner


BenTheElder:
https://github.com/mirror/busybox/commit/c9720a761e88e83265b4d75808533cdfbc66075b

busybox is now compatible with non-busy-box timeout, this broke triage

PR #16175, opened 2020-02-06
support running build with bazel version requested in .bazelversion

mikedanese:
This allows a commit in a seperate repo to upgrade bazel in a commit
before changing the default version. If the requested version is not
installed in the build image, we fall back to the default.

This also install 2.2.0 in the CI image but does not switch the default.

### DIFF
--- a/config/jobs/kubernetes/sig-testing/bazel-build-test.yaml
+++ b/config/jobs/kubernetes/sig-testing/bazel-build-test.yaml
@@ -21,11 +21,16 @@ presubmits:
       containers:
       - image: launcher.gcr.io/google/bazel:0.25.2
         command:
-        - bash
+        - ../test-infra/hack/bazel.sh
         args:
-        - -c
-        - |
-          ../test-infra/hack/bazel.sh build --config=remote --remote_instance_name=projects/k8s-prow-builds/instances/default_instance -- //... -//vendor/... -//build/... //build/release-tars
+        - build
+        - --config=remote
+        - --remote_instance_name=projects/k8s-prow-builds/instances/default_instance
+        - --
+        - //...
+        - -//vendor/...
+        - -//build/...
+        - //build/release-tars
   - name: pull-kubernetes-bazel-test
     decorate: true
     always_run: true
--- a/images/kubekins-e2e/Dockerfile
+++ b/images/kubekins-e2e/Dockerfile
@@ -69,9 +69,12 @@ RUN wget -q "https://storage.googleapis.com/golang/${GO_TARBALL}" && \
     rm "${GO_TARBALL}"
 
 # install bazel
+COPY ./install-bazel.sh /
+ENV BAZEL_VERSION 0.25.2
+RUN bash /install-bazel.sh
+
 ARG BAZEL_VERSION_ARG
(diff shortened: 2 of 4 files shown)


scenarios/kubernetes_bazel.py
```diff
@@ -37,13 +37,31 @@ def check_output(*cmd):
     print >>sys.stderr, 'Run:', cmd
     return subprocess.check_output(cmd)
 
+def find_bazel():
```
BenTheElder: imo we should stop using this script to invoke bazel directly and instead use some well defined target in k/k
can we manipulate the path instead? (maybe symlink .. or see if we can get bazelisk to pre-download these versions and use bazelisk then..?)

images/bazelbuild/install-bazel.sh
```diff
@@ -0,0 +1,47 @@
...
+  cat <<EOF >&2
+Usage:
+
+  ${0} [default bazel version]
+
+EOF
+  exit 1
+}
+
+function download_version() {
```
fejta: Why don't you just use bazelisk here?

mikedanese: Done.

scenarios/kubernetes_bazel.py
mikedanese: I added pre-downloading with bazelisk. Looking at the code, it's going to be hard to use bazelisk in a way that falls back to a version we like instead of trying to download and use latest.

> we should stop using this script to invoke bazel directly and instead use some well defined target in k/k

Would this require extensive changes? Is there anything minimal we can do here to bump the bazel version?

fejta: Are you familiar with `BAZEL_VERSION=real`? See https://github.com/kubernetes/test-infra/pull/16255

mikedanese: Awesome, did not know about the wrapper. Fixed.

mikedanese:
Am I updating the right image? What is this:

https://github.com/kubernetes/test-infra/blob/1ebe1f3846ca9b04ef23b7bb63068f46e48bad76/config/jobs/kubernetes/sig-testing/bazel-build-test.yaml#L22

fejta:
>  What is this:

This is an official image from bazel, you can just update to any released version (aka `:2.2.0` etc)

https://github.com/kubernetes/test-infra/blob/1ebe1f3846ca9b04ef23b7bb63068f46e48bad76/config/jobs/kubernetes/sig-testing/bazel-build-test.yaml#L22

images/bazelbuild/variants.yaml
```diff
@@ -1,4 +1,6 @@
 variants:
+  2.2.0:
+    BAZEL_VERSION: 2.2.0
```
fejta: This is for the `gcr.io/k8s-testimages/bazelbuild` images

fejta:
Pushed:
* kubekins-e2e: https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/post-test-infra-push-kubekins-e2e/1242228930178453504

* bazel: https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/post-test-infra-push-bazel/1242228930174259200

PR #16176, opened 2020-02-06
don't use -t in traige job

BenTheElder:
follow up to https://github.com/kubernetes/test-infra/pull/16173


BenTheElder:
you can run this locally and all docker images have an entrypoint. when you use an image without pod-utils the entrypoint is typically used instead of overridding it.

cjwagner:
No longer immediately failing :+1:  https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-test-infra-triage/1225558852359950344

Issue #16177, opened 2020-02-07
Branchprotector exits when unable to protect a branch


tehcyx:
**What happened**:
branchprotector exits, when a branch has a special character due to the query returning a 404, the code will actually just do a log.Fatalf and exit the process:
https://github.com/kubernetes/test-infra/blob/master/prow/cmd/branchprotector/protect.go#L135

Furthermore in our case, we also had a leftover config skeleton of a repository that got recently deleted, that also ended up with an error. Failing on the same Fatalf line.

Due to recreation of the branchprotector pod, branches that should have postsubmit jobs running might be subject to not having them run at all (which we encountered on our instance).

**What you expected to happen**:
branchprotector logs a warning but does not shut down.

**How to reproduce it (as minimally and precisely as possible)**:
1. Create a branch in your repo, name it e.g. test_#123 and have org or repo level protection enabled.
```
branch-protection:
  protect: true
  orgs:
    cfgdef:
```
2. Add a non-existent repository to protection list. Might only work if the repo existed at one point.

**Please provide links to example occurrences, if any**:
```
{"component":"branchprotector","error":"update kyma-incubator: update kyma-dns-webhook: list branches: return code not 2XX: 404 Not Found","level":"error","msg":"0","time":"2020-02-06T22:00:13Z"}
{"component":"branchprotector","error":"update kyma-project/test-infra=prow_installer_#1556 protection to {\"required_status_checks\":{\"strict\":false,\"contexts\":[\"license/cla\",\"pre-test-infra-validate-configs\",\"pre-test-infra-validate-scripts\"]},\"enforce_admins\":null,\"required_pull_request_reviews\":{\"dismissal_restrictions\":{},\"dismiss_stale_reviews\":false,\"require_code_owner_reviews\":true,\"required_approving_review_count\":1},\"restrictions\":null} failed: status code 404 not one of [200], body: {\"message\":\"Not Found\",\"documentation_url\":\"https://developer.github.com/v3\"}","level":"error","msg":"1","time":"2020-02-06T22:00:13Z"}
{"component":"branchprotector","level":"fatal","msg":"Encountered 2 errors protecting branches","time":"2020-02-06T22:00:13Z"}
```
https://github.com/kyma-project/test-infra/issues/2010

**Anything else we need to know?**:
This might be related: #14858

petr-muller:
Maybe I'm missing something, but can you elaborate how failing to set up branch protection causes postsubmits to not run?

[...]

PR #16178, opened 2020-02-07
branchprotector should not exit on 404

tehcyx:
For #16177, trying to provide a start for URL encode on branchprotector, as well as change the log.Fatalf to log.Errorf to not exit the process and force pod recreation.

edit: Change of logging is not necessary.

### DIFF
--- a/prow/cmd/branchprotector/protect.go
+++ b/prow/cmd/branchprotector/protect.go
@@ -133,7 +133,7 @@ func main() {
 		for i, err := range errors {
 			logrus.WithError(err).Error(i)
 		}
-		logrus.Fatalf("Encountered %d errors protecting branches", n)
+		logrus.Errorf("Encountered %d errors protecting branches", n)
 	}
 }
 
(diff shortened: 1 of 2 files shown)


tehcyx:
/cc @fejta @stevekuznetsov 
Would this be how we should deal with this? #16177

prow/cmd/branchprotector/protect.go
```diff
@@ -132,7 +133,7 @@ func main() {
 		for i, err := range errors {
 			logrus.WithError(err).Error(i)
 		}
-		logrus.Fatalf("Encountered %d errors protecting branches", n)
+		logrus.Errorf("Encountered %d errors protecting branches", n)
```
petr-muller: I don't think this is a correct thing to do. Branchprotector is not a daemon, it executes and terminates. Exiting with non-zero is the right thing to do when something fails.

tehcyx: I see and guess I didn't read it too well. In that case we can leave the Fatalf for sure.

tehcyx: Reverted to Fatalf

PR #16181, opened 2020-02-07
Update ingress-nginx prow job

aledbf:
Update `kindest/node` version to `v1.17.2`

xref: https://github.com/kubernetes-sigs/kind/pull/1318

### DIFF
--- a/config/jobs/kubernetes/sig-network/ingress-nginx.yaml
+++ b/config/jobs/kubernetes/sig-network/ingress-nginx.yaml
@@ -254,7 +254,7 @@ presubmits:
         - name: REPO_INFO
           value: https://github.com/kubernetes/ingress-nginx
         - name: K8S_VERSION
-          value: v1.17.0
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes/sig-network/ingress-nginx.yaml
```diff
@@ -254,7 +254,7 @@ presubmits:
         - name: REPO_INFO
           value: https://github.com/kubernetes/ingress-nginx
         - name: K8S_VERSION
-          value: v1.17.0
+          value: v1.17.2
```
BenTheElder: if you're only passing this value to KIND, consider using `kindest/node:v1.17.2@sha256:59df31fc61d1da5f46e8a61ef612fa53d3f9140f82419d1ef1a6b9656c6b737c"`

aledbf: done

PR #12440, opened 2019-05-01
Implement ConfigTree in approve plugin

matthyx:
Fixes #9757 
@cjwagner I couldn't make this work without encapsulating the config inside a config section... if you have any idea on how to make that work with `json.RawMessage` pls advise.

PR #16184, opened 2020-02-07
enabledRepos should be a list of org/repo

matthyx:
https://github.com/kubernetes/test-infra/pull/12440#discussion_r337277447
> enabledRepos won't contain orgs so this case will never be reached. enabledRepos is a list of org/repo strings (no org strings).


matthyx:
Ping @stevekuznetsov @alvaroaleman

matthyx:
> /lgtm
> /approve
> 
> What Cole wrote in the other PR sounds right and hence this PR is an improvement over the status quo, but IMHO we should stop passing repo strings around and use a proper `Repo` struct so that we do not string split or join everywhere and that we do not have to continously ask ourselves the quesiton "Is this `repo` an `org/repo` or just the repo or maybe depending on the situation?"

Thanks @alvaroaleman I take the point and will work on it :-)

PR #16185, opened 2020-02-07
Delete pipeline resources when pipeline run is deleted

eddycharly:
This pull request changes the way pipeline resources are created.
We use `ResourceSpec` instead of `ResourceRef` so that pipeline resources are passed inline with pipeline runs.

Also the creation of a pipeline run and the associated pipeline resources becomes a single operation. This either succeeds or fails but cannot partially fail or succeed anymore.

This way we don't need to garbage collect pipeline resources and delete them when they are no longer referenced by the pipeline run.

### DIFF
--- a/prow/cmd/pipeline/controller.go
+++ b/prow/cmd/pipeline/controller.go
@@ -272,7 +272,6 @@ type reconciler interface {
 	getPipelineRun(context, namespace, name string) (*pipelinev1alpha1.PipelineRun, error)
 	deletePipelineRun(context, namespace, name string) error
 	createPipelineRun(context, namespace string, b *pipelinev1alpha1.PipelineRun) (*pipelinev1alpha1.PipelineRun, error)
-	createPipelineResource(context, namespace string, b *pipelinev1alpha1.PipelineResource) (*pipelinev1alpha1.PipelineResource, error)
 	pipelineID(prowjobv1.ProwJob) (string, string, error)
 	now() metav1.Time
 }
@@ -315,6 +314,7 @@ func (c *controller) deletePipelineRun(context, namespace, name string) error {
 	}
 	return p.client.TektonV1alpha1().PipelineRuns(namespace).Delete(name, &metav1.DeleteOptions{})
 }
+
 func (c *controller) createPipelineRun(context, namespace string, p *pipelinev1alpha1.PipelineRun) (*pipelinev1alpha1.PipelineRun, error) {
 	logrus.Debugf("createPipelineRun(%s,%s,%s)", context, namespace, p.Name)
 	pc, err := c.getPipelineConfig(context)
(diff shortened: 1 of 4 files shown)


eddycharly:
/assign @alvaroaleman 
we discussed this in an earlier PR

prow/cmd/pipeline/controller.go
```diff
@@ -435,11 +445,15 @@ func reconcile(c reconciler, key string) error {
...
-		// Create the any git resources that are needed.
+		var pipelineResources []*pipelinev1alpha1.PipelineResource
+
+		// Create the git resources that are needed.
 		for _, res := range resources {
 			resKey := toKey(ctx, namespace, res.Name)
 			logrus.Infof("Create PipelineResource/%s", resKey)
-			if _, err = c.createPipelineResource(ctx, namespace, &res); err != nil {
+			if pr, err := c.createPipelineResource(ctx, namespace, &res); err != nil {
+				pipelineResources = append(pipelineResources, pr)
```
alvaroaleman: Why do we do the append only when we had an error? Also, its not safe to use other return args when err was not nil

prow/cmd/pipeline/controller.go
```diff
@@ -455,6 +469,21 @@ func reconcile(c reconciler, key string) error {
 			// the pipeline cannot be executed (e.g. referenced pipeline does not exist)
 			return updateProwJobState(c, key, newPipelineRun, pj, newpj, prowjobv1.ErrorState, jerr.Error())
 		}
+
+		// Set pipeline resources owner so that they are deleted when the pipeline run goes away
+		for _, pr := range pipelineResources {
```
alvaroaleman: What about instead just iterating over the `resources`, doing a `Get`, checking if the `OwnerRef` is already there and adding it & `Update` if not?

prow/cmd/pipeline/controller.go
eddycharly: Fixed

prow/cmd/pipeline/controller.go
eddycharly: The pipeline resources were created by us a few lines above without an owner.
Why should we make an additional `Get` for something we already have, and check `OwnerRef` when we created the resource without one ?
Do you think about something that could break in this case ?

eddycharly:
Ok for writing a unit test

prow/cmd/pipeline/controller.go
alvaroaleman: Because we tolerate `IsConflict` errors in which case the pipeline resource might be nil

eddycharly: Thinking about it again, i'm going to do it completely differently.

I am not sure when it was added in tekton but we now have the ability to define resources _inline_ in the pipeline run, that would eliminate the need to create pipeline resources and pipeline runs separately.

eddycharly:
@alvaroaleman I quickly changed the implementation to use `ResourceSpec` instead of `ResourceRef`. I still have a few tests to adapt but if you can give a look and tell me what you think i would appreciate.

The code is a lot simpler imho.

The more i think about it, the more i feel like we don't need the `PROW_IMPLICIT_GIT_REF` magic strings anymore. We could do everything using params instead (we already have all we need, sha, refs, etc to build the resource in the pipeline run directly).

It's still good to have for backward compatibility though.

alvaroaleman:
> @alvaroaleman I quickly changed the implementation to use ResourceSpec instead of ResourceRef. I still have a few tests to adapt but if you can give a look and tell me what you think i would appreciate

Embedding it instead of creating a different CRD is indeed a lot nicer here.
/lgtm
/approve

From my POV this pull looks good or is there anything you would like to add? If not, feel free to remove the hold.
/hold

eddycharly:
All is good for me, i remove the hold.
/hold cancel

PR #16186, opened 2020-02-07
Experimental jobs for merged density and load

mm4tt:
Job comes in two flavors with containerd and without (with docker).

Ref. https://github.com/kubernetes/perf-tests/issues/1007

/hold
Wait for https://github.com/kubernetes/perf-tests/pull/1008

### DIFF
--- a/config/jobs/kubernetes/sig-scalability/sig-scalability-experimental-periodic-jobs.yaml
+++ b/config/jobs/kubernetes/sig-scalability/sig-scalability-experimental-periodic-jobs.yaml
@@ -215,3 +215,120 @@ periodics:
       - --test-cmd-name=ClusterLoaderV2
       - --timeout=120m
       - --use-logexporter
+
+- interval: 6h
+  name: ci-kubernetes-e2e-scalability-experimental-load
+  tags:
+    - "perfDashPrefix: gce-100Nodes-master-experimental"
+    - "perfDashJobType: performance"
+    - "perfDashBuildsCount: 500"
+  labels:
+    preset-service-account: "true"
+    preset-k8s-ssh: "true"
+    preset-e2e-scalability-common: "true"
+  annotations:
+    testgrid-dashboards: sig-scalability-experiments
+    testgrid-tab-name: experimental-load
+  spec:
+    containers:
+      - image: gcr.io/k8s-testimages/kubekins-e2e:v20200206-f88edef-master
+        args:
+          - --timeout=140
+          - --repo=k8s.io/kubernetes=master
+          - --repo=k8s.io/perf-tests=master
+          - --root=/go/src
+          - --scenario=kubernetes_e2e
+          - --
+          - --check-leaked-resources
+          # TODO(oxddr): remove once debugging is finished
+          - --env=KUBEPROXY_TEST_LOG_LEVEL=--v=4
+          - --cluster=e2e-big-load
+          - --extract=ci/latest
+          - --gcp-node-image=gci
+          - --gcp-nodes=100
+          - --gcp-project-type=scalability-project
+          - --gcp-zone=us-east1-b
+          - --provider=gce
+          - --test=false
+          - --test-cmd=$GOPATH/src/k8s.io/perf-tests/run-e2e.sh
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes/sig-scalability/sig-scalability-experimental-periodic-jobs.yaml
```diff
@@ -215,3 +215,120 @@ periodics:
...
+      - image: gcr.io/k8s-testimages/kubekins-e2e:v20200206-f88edef-master
+        args:
+          - --timeout=140
+          - --repo=k8s.io/kubernetes=master
+          - --repo=k8s.io/perf-tests=master
+          - --root=/go/src
+          - --scenario=kubernetes_e2e
+          - --
+          - --check-leaked-resources
+          # TODO(oxddr): remove once debugging is finished
```
wojtek-t: Do we need it in our experimental one?

config/jobs/kubernetes/sig-scalability/sig-scalability-experimental-periodic-jobs.yaml
```diff
@@ -215,3 +215,120 @@ periodics:
...
+          - --test-cmd-args=cluster-loader2
+          - --test-cmd-args=--experimental-gcp-snapshot-prometheus-disk=true
+          - --test-cmd-args=--experimental-prometheus-disk-snapshot-name=${JOB_NAME}-${BUILD_ID}
+          - --test-cmd-args=--nodes=100
+          - --test-cmd-args=--prometheus-scrape-node-exporter
+          - --test-cmd-args=--provider=gce
+          - --test-cmd-args=--report-dir=/workspace/_artifacts
+          - --test-cmd-args=--testconfig=testing/load/experimental-config.yaml
+          - --test-cmd-args=--testoverrides=./testing/chaosmonkey/override.yaml
+          - --test-cmd-args=--testoverrides=./testing/load/experimental/overrides/enable_configmaps.yaml
```
wojtek-t: Do we need those? They are all enabled by default now (i.e. we don't use use those env vars IIRC).

config/jobs/kubernetes/sig-scalability/sig-scalability-experimental-periodic-jobs.yaml
```diff
@@ -215,3 +215,120 @@ periodics:
...
+      - image: gcr.io/k8s-testimages/kubekins-e2e:v20200206-f88edef-master
+        args:
+          - --timeout=140
+          - --repo=k8s.io/kubernetes=master
+          - --repo=k8s.io/perf-tests=master
+          - --root=/go/src
+          - --scenario=kubernetes_e2e
+          - --
+          - --check-leaked-resources
+          # TODO(oxddr): remove once debugging is finished
```
wojtek-t: Same comments as above for this job.

config/jobs/kubernetes/sig-scalability/sig-scalability-experimental-periodic-jobs.yaml
mm4tt: They aren't enabled by default (even though they should be) and actually we have this litany of testoverrides in ~10+  places already (3 x kubemark, 2x gce, branch releases, etc.). I will file an issue to clean it up.

config/jobs/kubernetes/sig-scalability/sig-scalability-experimental-periodic-jobs.yaml
mm4tt: We don't, but I wanted to keep it to original test as close to the perf-100 as possible.
That means this job config is almost the same as ci-kubernetes-e2e-gci-gce-scalability config.

config/jobs/kubernetes/sig-scalability/sig-scalability-experimental-periodic-jobs.yaml
mm4tt: https://github.com/kubernetes/perf-tests/issues/1036

PR #16196, opened 2020-02-08
Boskos: add option to serve Prometheus resource metrics from core server

ixdy:
This PR adds a flag `--serve-resource-metrics` to the Boskos core, which if enabled, will cause Boskos to serve metrics on resources via Prometheus.

These metrics previously were served via the `boskos-metrics` component, but since we are already scraping the Boskos core server for HTTP metrics, we may as well scrape resource metrics from there, too. This also means that adding a new type of resource won't require updating the `boskos-metrics` deployment - we'll just automatically start serving metrics for the new type.

I've put this behind a flag (disabled by default) since I'm not sure whether Prometheus would do the right thing scraping the same metric from two different sources. Ideally we'd push out the config change to stop scraping the `boskos-metrics` service at the same time we flip this flag.

Addresses #16001. I'll wait to close that issue until we actually deploy this.

cc @stevekuznetsov @alvaroaleman

### DIFF
--- a/boskos/metrics/resources.go
+++ b/boskos/metrics/resources.go
@@ -1,0 +1,90 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package metrics
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/sirupsen/logrus"
+
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/test-infra/boskos/common"
+	"k8s.io/test-infra/boskos/ranch"
+)
+
+const (
+	// ResourcesMetricName is the name of the Prometheus metric used to monitor Boskos resources.
+	ResourcesMetricName = "boskos_resourcess"
+	// ResourcesMetricDescription is the description for the Prometheus metric used to monitor Boskos resources.
+	ResourcesMetricDescription = "Number of resources recorded in Boskos by resource type and state."
+)
+
+var (
+	// ResourcesMetricLabels is the list of labels used for the Prometheus metric used to monitor Boskos resources.
+	ResourcesMetricLabels = []string{"type", "state"}
+)
+
+type resourcesCollector struct {
+	boskosResources *prometheus.Desc
+	ranch           *ranch.Ranch
+}
+
+// NewResourcesCollector returns a collector which exports the current counts of
+// Boskos resources, segmented by resource type and state.
+func NewResourcesCollector(ranch *ranch.Ranch) prometheus.Collector {
+	return resourcesCollector{
+		boskosResources: prometheus.NewDesc(ResourcesMetricName, ResourcesMetricDescription, ResourcesMetricLabels, nil),
+		ranch:           ranch,
+	}
+}
+
+func (rc resourcesCollector) Describe(ch chan<- *prometheus.Desc) {
+	ch <- rc.boskosResources
+}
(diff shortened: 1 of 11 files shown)


boskos/metrics/resources.go
```diff
@@ -0,0 +1,90 @@
...
+	NormalizeResourceMetrics(metrics, common.KnownStates, func(rtype, state string, count float64) {
+		ch <- prometheus.MustNewConstMetric(rc.boskosResources, prometheus.GaugeValue, count, rtype, state)
+	})
+}
+
+// NormalizeResourceMetrics "normalizes" the list of provided Metrics by
+// bucketing any state not in states into the "Other" state, and by ensuring
+// every state in states has some count (even if zero).
+// It then applies the function for each combination of resource type and state.
+func NormalizeResourceMetrics(metrics []common.Metric, states []string, updateFunc func(rtype, state string, count float64)) {
```
ixdy: I'm open to a better name than "Normalize".

boskos/metrics/resources.go
```diff
@@ -0,0 +1,90 @@
...
+)
+
+type resourcesCollector struct {
+	boskosResources *prometheus.Desc
+	ranch           *ranch.Ranch
+}
+
+// NewResourcesCollector returns a collector which exports the current counts of
+// Boskos resources, segmented by resource type and state.
+func NewResourcesCollector(ranch *ranch.Ranch) prometheus.Collector {
```
alvaroaleman: What keeps us from using a `prometheus.GaugeVec` rather than building a collector ourselves?

boskos/metrics/resources.go
sebastienvas: You don't need to keep that functionality, that can probably be done in the dashboard. This was done for mason resources, since they have a name of a different resource as a state. Up to you.

ixdy: The [Prometheus docs do warn against having too many different label values](https://prometheus.io/docs/practices/naming/#labels), so I think it makes sense to have a catch-all "other" category.
> CAUTION: Remember that every unique combination of key-value label pairs represents a new time series, which can dramatically increase the amount of data stored. Do not use labels to store dimensions with high cardinality (many different label values), such as user IDs, email addresses, or other unbounded sets of values.

boskos/metrics/resources.go
ixdy: I started by using a `GaugeVec`, using a similar approach where we'd export metrics for any resource types present in Boskos. The problem is that if a type were deleted (by completely removing its resources), we'd continue to serve stale metrics for that type until Boskos was restarted (and the gauge was recreated).

The first way I attempted to fix this issue was by deleting any label values that didn't match resource types currently in Boskos. But the Prometheus Go API doesn't have any way to list all known values for a label, which suggested this was probably not the right approach. I considered keeping a list of previously-reported types, but that didn't seem great, either.

I stepped back and considered that having a periodic loop to update metrics in the server was also probably the wrong design. I started work to update the metrics in Ranch, when resources were added, deleted, or updated, but realized there were possible consistency issues there (e.g. if something changed the values in the API server without going through the Boskos API directly).

Ultimately, I noticed that latencies for `/metric` are pretty low, the current metrics server calls this API multiple times for each type, and the implementation of `/metric` already lists all resources and then filters out the ones matching the type. I figured that it shouldn't be that much more expensive to just generate this list when Prometheus metrics are scraped. (Your caching work should make this even less of an issue.)

Also, I think this code ended up being fairly easy to understand and easy to test.

stevekuznetsov:
> I've put this behind a flag (disabled by default) since I'm not sure whether Prometheus would do the right thing scraping the same metric from two different sources. 

You don't need to worry here, it will scrape and annotate them separately with different `job` identifiers.

ixdy:
@stevekuznetsov I think I was most concerned about prow.k8s.io, where we scrape both in the same job:
https://github.com/kubernetes/test-infra/blob/c8dc24a7f0351e687352719646aec49af2f260f8/prow/cluster/monitoring/additional-scrape-configs_secret.yaml#L8-L13

I guess I could refactor those before merging this. Or we could just live with duplicated metrics for a short time.

ixdy:
Doing a bit more research, I see that Prometheus adds both `job` and `instance` labels automatically, so that might be enough for filtering. We'd probably still get some duplication in grafana for a short time, but it looks like the alerts already do the right thing.

ixdy:
Rebased and removed the flag.

PR #16201, opened 2020-02-09
[kops-e2e] --ginkgo.flakeAttempts=1 to 2 to be a bit more forgiving

geojaz:
I've been getting failures in e2e that don't seem to make sense. I thought bumping this setting by 1 would potentially help us not have to re-run from scratch... of course, this doesn't actually solve the problem of flaking tests :/

### DIFF
--- a/config/jobs/kubernetes/kops/kops-config.yaml
+++ b/config/jobs/kubernetes/kops/kops-config.yaml
@@ -88,7 +88,7 @@ presubmits:
         - --kops-build
         - --kops-priority-path=/home/prow/go/src/k8s.io/kops/kubernetes/platforms/linux/amd64
         - --provider=aws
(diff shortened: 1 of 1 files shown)


dims:
Please don't!!! it's better to hunt down and fix the flakes :(

geojaz:
> Please don't!!! it's better to hunt down and fix the flakes :(

/shrug 
yeah..... i know.... i felt like giving a second chance was not the worst thing. trying 5 times in a row... ;)

geojaz:
That said, I added a PR for statefulset tests: https://github.com/kubernetes/kubernetes/pull/87963
and I think https://github.com/kubernetes/kubernetes/pull/87819 will make my life (and others) much better.

PR #16204, opened 2020-02-10
Update dns provider flag for e2e test

hakman:
Updating the CoreDNS test after https://github.com/kubernetes/test-infra/pull/16192.


rifelpet:
the kubetest changes this depends on are only in the latest-master image so we're using that until images are rebuilt and bumped.

/lgtm

PR #16206, opened 2020-02-10
Boskos: Remove downstream client

alvaroaleman:
This PR is a first step towards making Boskos use a cache underneath, which we need to speed the whole thing up, as it lists all Resources for each [`Acquire`](https://github.com/kubernetes/test-infra/blob/f51bb12132dbc3b35a9223d1d4caf8d6361c1555/boskos/ranch/ranch.go#-L117-L134)/[`AcquireByState`](https://github.com/kubernetes/test-infra/blob/f51bb12132dbc3b35a9223d1d4caf8d6361c1555/boskos/ranch/ranch.go#L207-L217) call, which is pretty slow because we have about 4k objects.

This PR removes the downstream kube client we have in boskos and replaces it with [the one from controller-runtime](https://godoc.org/sigs.k8s.io/controller-runtime/pkg/client#Client), of which an api-compatible cache-backed version exists.

Note that the PR does not change the client to be cache-backed yet, I will do that in a followup as this PR is big enough already.

/assign @ixdy @sebastienvas @stevekuznetsov


boskos/cmd/boskos/boskos.go
```diff
@@ -68,6 +69,8 @@ func init() {
 func main() {
 	logrusutil.ComponentInit("boskos")
 	kubeClientOptions.AddFlags(flag.CommandLine)
+	var namespace string
+	flag.StringVar(&namespace, "namespace", corev1.NamespaceDefault, "namespace to install on")
```
ixdy: is there a reason you put this in main() instead of with the rest of the flags above?

boskos/crds/client.go
```diff
@@ -111,233 +105,50 @@ type Collection interface {
...
-func registerResource(config *rest.Config, t Type) error {
+// registerResources sends a request to create CRDs
+func registerResources(config *rest.Config) error {
 	c, err := apiextensionsclient.NewForConfig(config)
 	if err != nil {
 		return err
 	}
 
-	crd := &apiextensionsv1beta1.CustomResourceDefinition{
+	resourceCRD := &apiextensionsv1beta1.CustomResourceDefinition{
```
ixdy: idle comment: we should probably update to apiextensions/v1 at some point.

boskos/crds/client.go
```diff
@@ -111,233 +105,50 @@ type Collection interface {
...
-// registerResource sends a request to create CRDs and waits for them to initialize
-func registerResource(config *rest.Config, t Type) error {
+// registerResources sends a request to create CRDs
+func registerResources(config *rest.Config) error {
 	c, err := apiextensionsclient.NewForConfig(config)
 	if err != nil {
 		return err
 	}
 
-	crd := &apiextensionsv1beta1.CustomResourceDefinition{
```
ixdy: we could've kept this generic and looped over the two types, right?

I guess the new version is a little easier to read.

boskos/ranch/storage.go
```diff
@@ -27,37 +29,43 @@ import (
...
 	"k8s.io/test-infra/boskos/common"
-	"k8s.io/test-infra/boskos/storage"
+	"k8s.io/test-infra/boskos/crds"
 )
 
 // Storage is used to decouple ranch functionality with the resource persistence layer
 type Storage struct {
-	resources, dynamicResourceLifeCycles storage.PersistenceLayer
-	resourcesLock                        sync.RWMutex
+	ctx           context.Context
```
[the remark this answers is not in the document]
alvaroaleman: Never, added a commit to properly plug it through

boskos/ranch/storage.go
```diff
@@ -27,37 +29,43 @@ import (
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+
 	"k8s.io/test-infra/boskos/common"
-	"k8s.io/test-infra/boskos/storage"
+	"k8s.io/test-infra/boskos/crds"
 )
 
 // Storage is used to decouple ranch functionality with the resource persistence layer
```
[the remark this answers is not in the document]
alvaroaleman: Do you think so? IMHO it is accurate for as long as we have our internal representation, because the `Storage` hides the external (CRD) one.

boskos/ranch/storage.go
```diff
@@ -87,115 +95,134 @@ func NewStorage(res, lf storage.PersistenceLayer, storage string) (*Storage, err
...
 // UpdateResource updates a resource if it exists, errors otherwise
 func (s *Storage) UpdateResource(resource common.Resource) (common.Resource, error) {
 	resource.LastUpdate = s.now()
-	i, err := s.resources.Update(resource)
-	if err != nil {
-		return common.Resource{}, err
+
+	o := &crds.ResourceObject{}
+	name := types.NamespacedName{Namespace: s.namespace, Name: resource.GetName()}
+	if err := s.client.Get(s.ctx, name, o); err != nil {
```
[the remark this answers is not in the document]
alvaroaleman: Yes, it is an orthogonal change, the overarching goal of this PR is to allow using a cache

boskos/cmd/boskos/boskos.go
alvaroaleman: Oversight, fixed

boskos/crds/client.go
alvaroaleman: Well yeah, looping would require to have a slice somewhere everyone who could think about adding a CRD will definitely find and I was too lazy to think about the best place for that :)

IMHO we should drop the whole CRD creation entirely, as that requires us to give CRD Create permissions, which shouldn't be needed. I didn't do that in this PR to keep the potential basis for debate as minimal as possible :)

boskos/ranch/ranch_test.go
```diff
@@ -24,10 +24,17 @@ import (
...
 	"time"
 
+	"github.com/sirupsen/logrus"
+	fakectrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client/fake"
+
 	"k8s.io/test-infra/boskos/common"
-	"k8s.io/test-infra/boskos/crds"
 )
 
+// Make debugging a bit easier
```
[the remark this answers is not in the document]
alvaroaleman: The go tests suppress all logrus logs unless the test fails, in which case the person debugging them most likely wants debug-level logs.

Not a strong opinion here though, I can also remove it again.

boskos/cmd/fake-mason/main.go
```diff
@@ -47,6 +47,7 @@ var (
 	username          = flag.String("username", "", "Username used to access the Boskos server")
 	passwordFile      = flag.String("password-file", "", "The path to password file used to access the Boskos server")
 	cleanerCount      = flag.Int("cleaner-count", defaultCleanerCount, "Number of threads running cleanup")
+	namespace         = flag.String("namespace", corev1.NamespaceDefault, "namespace to install on")
```
stevekuznetsov: This precludes a rolling upgrade, right, since we can't add the flag first, then upgrade the version, or the other way around

alvaroaleman: The flag just got moved, it was part of the kubeOpts before

boskos/ranch/storage.go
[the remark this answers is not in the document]
ixdy: well, I mostly meant that we're no longer depending on the storage.PersistenceLayer interface, and some things (like the runtime client and namespace) are now leaking through.

boskos/crds/client.go
ixdy: > IMHO we should drop the whole CRD creation entirely, as that requires us to give CRD Create permissions, which shouldn't be needed. I didn't do that in this PR to keep the potential basis for debate as minimal as possible :)

I agree. https://github.com/kubernetes/test-infra/pull/16266

PR #16212, opened 2020-02-10
prow cluster: create loop devices in advance

pohly:
This is a workaround for dynamic device creation not working inside
KinD nodes.

Fixes: https://github.com/kubernetes-sigs/kind/issues/1248 https://github.com/kubernetes/kubernetes/issues/87953

### DIFF
--- a/prow/cluster/loop_dev_daemonset.yaml
+++ b/prow/cluster/loop_dev_daemonset.yaml
@@ -1,0 +1,54 @@
+# A daemonset to create /dev/loopX device nodes before they are
+# needed.
+#
+# Normally, new nodes are created dynamically by the kernel. But nodes
+# in a KinD cluster are started with a copy of /dev from the host and
+# loop devices created later on do not show up in that static /dev
+# (https://github.com/kubernetes-sigs/kind/issues/1248). Creating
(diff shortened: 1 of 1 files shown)


pohly:
I have verified that:
- Deploying this daemonset on a normal cluster populates /dev as intended.
- That the nodes in a KinD  cluster with fully populated /dev also then have the same /dev/loopX entries.

I have not verified that this really works in the Kubernetes prow. I'm not even sure whether the daemonset will be deployed automatically?

I modeled this after https://github.com/kubernetes/test-infra/blob/b2d632960440e47f671dd3b997709aa299911456/prow/cluster/tune-sysctls_daemonset.yaml.

/cc @BenTheElder

BenTheElder:
it can be deployed if added to the bazel rules for deploying prow, see https://github.com/kubernetes/test-infra/commit/75edf5a28b3f77dbd596775cbf1bfbdaede17c28#diff-f677234fb7028152fac5bac409d77567

BenTheElder:
/lgtm
/approve
will do deployment in a follow up
this hack is sorta gross, but also only something we should need, the average KIND user is probably not using block devices ... I hope!

PR #16215, opened 2020-02-10
Introduce log collection for aks-engine Windows Nodes

adelina-t:


adelina-t:
related: https://github.com/kubernetes-sigs/windows-testing/pull/127

kubetest/azure.go
```diff
@@ -1056,20 +1064,54 @@ func (c *Cluster) DumpClusterLogs(localPath, gcsPath string) error {
...
+		return nil
 	}
 
-	return control.FinishRunning(exec.Command("bash", "-c", logDumpScript))
+	var errors []string
+	if err := logDumper(); err != nil {
+		errors = append(errors, err.Error())
+	}
+	if err := logDumperWindows(); err != nil {
+		errors = append(errors, err.Error())
```
feiskyer: would this report errors if there're no Windows nodes in the cluster?

adelina-t: No. The collection script will just exit and log that there are no windows nodes it can access.

feiskyer: thanks for confirm

PR #16219, opened 2020-02-11
GCE Windows: specify version-specific manifest of test containers to prepull.

pjh:
This PR takes advantage of the changes in https://github.com/kubernetes-sigs/windows-testing/pull/126 to prepull the right set of test containers onto Windows test nodes. This may reduce test flakiness due to tests failing to pull their test containers in time (Windows containers can take 5+ minutes to download and extract).


pjh:
@calebamiles @justaugustus @tpepper looking for approval for this change which should reduce test flakiness, please take a look. Thanks!

pjh:
Thanks @justaugustus, the daily kubekins update makes this PR need a rebase daily :( If you have a chance to approve before the next one happens that would be great!

PR #16220, opened 2020-02-11
Fix network proxy e2e presubmit to point to PR

Jefftree:
Fix network proxy e2e presubmit to point to PR instead of the master branch.

`decorate: true` was also accidentally added and we are not using pod utilities so it should be removed.

/cc @BenTheElder 
/assign @lavalamp

### DIFF
--- a/config/jobs/kubernetes/sig-api-machinery/sig-api-machinery-config.yaml
+++ b/config/jobs/kubernetes/sig-api-machinery/sig-api-machinery-config.yaml
@@ -27,14 +27,13 @@ periodics:
 
 - interval: 24h
   name: ci-kubernetes-e2e-gci-gce-network-proxy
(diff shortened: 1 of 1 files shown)


BenTheElder:
>     jobs_test.go:887: Invalid Scenario Args : presubmit jobs pull-kubernetes-e2e-gce-network-proxy - need to stage to gs://kubernetes-release-pull/ci/pull-kubernetes-e2e-gce-network-proxy

I guess some check in the repo wants a --stage based on the job name?

I don't remember this stuff super well anymore KIND > GCE :upside_down_face:

Jefftree:
@BenTheElder Thanks for the review! Is there anything that needs to be done with the stage as part of the setup, or would it already be configured based on the job name?

PR #16222, opened 2020-02-11
kops staging builds: build release branches also

justinsb:
We typically release from a release-1.nn branch, so build those
branches also.


rifelpet:
/lgtm
We'll also want to cherry-pick our recent cloudbuild.yaml changes in the kops repo back to the release branches

PR #16074, opened 2020-01-30
releng: Add ci-release-anago-stage job to test anago stage functionality

justaugustus:
(Needs https://github.com/kubernetes/release/pull/1052)

It's overdue that we be able to test anago functionality in CI, instead of relying on Release Managers to test changes against their personal forks/branches.
This PR adds a CI job for anago staging (`./gcbmgr stage master`) and a new dashboard for tracking Release Engineering jobs (`sig-release-releng-informing`).

I'll shuffle some of the existing jobs on `sig-release-misc` to this new dashboard in a follow-up PR.

Signed-off-by: Stephen Augustus <saugustus@vmware.com>

/assign @tpepper @hoegaarden @saschagrunert @cpanato 
cc: @kubernetes/release-engineering

PR #16224, opened 2020-02-11
releng: Remove ci-release-anago-stage job

justaugustus:
~The gcbmgr/anago staging test is currently failing because it's
attempting to mount in a secret ('deployer-service-account') that
doesn't exist on the standard Prow clusters.~

~The job will continue to fail after this merges, but likely because Prow
doesn't have access to run GCB jobs within the kubernetes-release-test
GCP project. Once we see the next set of errors, we can go about
adjusting IAM on the releng staging account to support this.~

---

Deciding instead to remove this job, as we're not quite ready to make this work.
Closes: https://github.com/kubernetes/kubernetes/issues/88804

/assign @Katharine @saschagrunert @hasheddan @tpepper @hoegaarden
ref: https://testgrid.k8s.io/sig-release-releng-informing#ci-release-anago-stage, https://github.com/kubernetes/release/pull/1052, https://github.com/kubernetes/test-infra/pull/16074, https://github.com/kubernetes/release/pull/1081, https://github.com/kubernetes/release/pull/1082

### DIFF
--- a/config/jobs/kubernetes/release/release-config.yaml
+++ b/config/jobs/kubernetes/release/release-config.yaml
@@ -61,28 +61,6 @@ presubmits:
       testgrid-alert-email: release-managers@kubernetes.io
       testgrid-num-columns-recent: '30'
 periodics:
-- interval: 2h
-  name: ci-release-anago-stage
-  decorate: true
-  extra_refs:
-  - org: kubernetes
-    repo: release
-    base_ref: master
-    path_alias: k8s.io/release
-  spec:
-    #erviceAccountName: deployer -- TODO- this is not valid for the untrusted cluster, fix
-    containers:
-    - image: gcr.io/k8s-staging-release-test/k8s-cloud-builder:latest
-      command:
-        - ./gcbmgr
-      args:
-        - stage
-        - master
-        - --attended
-  annotations:
-    testgrid-dashboards: sig-release-releng-informing
-    testgrid-alert-email: release-managers@kubernetes.io
-
 # package tests
 - name: periodic-release-verify-packages-debs
   interval: 6h


config/jobs/image-pushing/k8s-staging-release-test.yaml
```diff
@@ -1,4 +1,34 @@
...
+  cluster: test-infra-trusted
+  decorate: true
+  extra_refs:
+  - org: kubernetes
+    repo: release
+    base_ref: master
+    path_alias: k8s.io/release
+  spec:
+    containers:
+    - image: gcr.io/k8s-staging-release-test/k8s-cloud-builder:latest
```
Katharine: This cannot be permitted to run in the trusted cluster without further discussion with the test-infra team: we do not permit code we do not maintain to run there for security reasons.

/hold

BenTheElder: Also even if we did, we probably would not be super excited about a `:latest` ...
/lgtm cancel

justaugustus:
Chatted a bit w/ @Katharine about this today and the best options are:
- Tweak the job so it's only running a GCB submission (the Makefile being an unknown factor here)
- Secure a releng GKE cluster, so we've got some flexibility in what we can do w/o needing the test-infra-trusted cluster

Option #1 is faster; option #2 is a bit more future-proof.
I'll need to noodle on this a little, but it's low-priority w.r.t. some other things on my plate at the moment.

/hold explicit (until we think of a solution)

justaugustus:
Outright deleting this job instead, since I need to think a little bit more about how we want to do test staging.

PR #16236, opened 2020-02-11
Enforce that image pushing jobs use our GCB builder.

Katharine:
Ensure that things like #16224 don't accidentally slip through the net.


Katharine:
/cc @BenTheElder 

Ensure that things like #16224 don't accidentally slip through the net.

PR #16237, opened 2020-02-11
Remove some debt

BenTheElder:
we haven't persisted images in a long time. this is a much simpler way to do cleanup without a custom binary

this cleanup snippet comes from KRTE where we've been using it just fine for a couple KIND jobs prototyping a simpler squashed down kubekins, so backporting this to the existing jobs now ...


BenTheElder:
I hoped this would let us drop some dependencies but hack/update-deps.sh has no diff. we must pull in docker somewhere else.

PR #16238, opened 2020-02-11
Use image to choose bazel version in CI

fejta:
/assign @BenTheElder


BenTheElder:
whoo green
/lgtm
/approve

PR #16240, opened 2020-02-12
Run post-deploy job on changes under "examples" folder

figo:
Because those example yaml files goes into cluster-api-provider-vsphere/ci/manifests container image.

otherwise, when PR changing files under examples, it will not trigger a new manifest image.

Signed-off-by: Hui Luo <luoh@vmware.com>

cc @akutz @codenrhoden


codenrhoden:
/lgtm

Good catch, @figo. Like you said, this would have made PRs that *should* update the manifests CI image not do so.

PR #16241, opened 2020-02-12
add mkpj utility for prow.k8s.io

BenTheElder:
This is a much faster way to create a prowjob inspired by pj-on-kind.sh but for creating jobs on the production prow (e.g. for an admin to trigger a postsubmit again with updated job config).

It is simpler by assuming everything must be at certain paths under config/.

use like: `config/mkpj.sh --job=post-test-infra-push-bootstrap | kubectl create -f -`

### DIFF
--- a/config/jobs/README.md
+++ b/config/jobs/README.md
@@ -171,6 +171,18 @@ recognize this is going to depend on review latency.
 
 For more details, please refer to [How to Test a ProwJob](/prow/build_test_update.md#how-to-test-a-prowjob)
 
+## Running a Production Job
+
+Normally prow will automatically schedule your job, however if for some reason you
+need to trigger it again you have a few options:
+
+- you can use the rerun feature in prow.k8s.io to run the job again *with the same config*
+- you can use [`config/mkpj.sh`](/config/mkpj.sh) to create a prowjob CR from your local config
+- you can use `bazel run //prow/cmd/mkpj -- --job=foo ...` to create a prowjob CR from your local config
+
+For the latter two options you'll need to submit the resulting CR via `kubectl` configured against
+the prow services cluster.
+
 ## Generated Jobs
 
 There are some sets of jobs that are generated and should not be edited by hand.
--- a/config/jobs/README.md
+++ b/config/jobs/README.md
@@ -174,7 +174,7 @@ For more details, please refer to [How to Test a ProwJob](/prow/build_test_updat
 ## Running a Production Job
 
 Normally prow will automatically schedule your job, however if for some reason you
-need to trigger it again you have a few options:
+need to trigger it again and are a Prow administrator you have a few options:
 
 - you can use the rerun feature in prow.k8s.io to run the job again *with the same config*
 - you can use [`config/mkpj.sh`](/config/mkpj.sh) to create a prowjob CR from your local config
--- a/config/mkpj.sh
+++ b/config/mkpj.sh
@@ -1,0 +1,35 @@
+#!/usr/bin/env bash
+# Copyright 2020 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Usage: mkpj.sh --job=foo ...
+#
+# Arguments to this script will be passed to a dockerized mkpj
+#
+# Example Usage:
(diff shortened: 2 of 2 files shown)


config/mkpj.sh
```diff
@@ -0,0 +1,28 @@
...
+#
+# Example Usage:
+# config/mkpj.sh --job=post-test-infra-push-bootstrap | kubectl create -f -
+# (type "master" at the Base ref prompt)
+
+root="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd -P)"
+config="${root}/config/prow/config.yaml"
+job_config_path="${root}/config/jobs"
+
+docker run -i --rm -v "${PWD}:${PWD}" -v "${root}:${root}" -w "${PWD}" gcr.io/k8s-prow/mkpj "--config-path=${config}" "--job-config-path=${job_config_path}" "$@"
```
cjwagner: I don't think you need to mount `${PWD}` or set it as the working dir.

config/mkpj.sh
```diff
@@ -0,0 +1,28 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Usage: mkpj.sh --job=foo ...
+#
+# Arguments to this script will be passed to a dockerized mkpj
+#
+# Example Usage:
+# config/mkpj.sh --job=post-test-infra-push-bootstrap | kubectl create -f -
```
cjwagner: nit: Mention that kubectl must be pointed to the k8s-prow service cluster.

config/mkpj.sh
```diff
@@ -0,0 +1,28 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Usage: mkpj.sh --job=foo ...
+#
+# Arguments to this script will be passed to a dockerized mkpj
```
cjwagner: nit: Mention the existence of the bazel version of the script.

Also consider adding this script to the docs: https://github.com/kubernetes/test-infra/tree/master/config/jobs#kubernetes-prow-job-configs

config/mkpj.sh
stevekuznetsov: Depending on the flavor of Linux and your settings, the volumes *might* need `-v a:b:z`

config/mkpj.sh
chases2: Is a containerized version of mkpj preferable to `bazel run //prow/cmd/mkpj`? If one method is preferable, why?

stevekuznetsov: It is, because the version of `mkpj` you use ideally should match the version you have deployed, or you risk nuking your deployment by pushing a ProwJob CR that has fields that the current controller levels don't know about.

cjwagner: FWIW the bazel target I'm talking about does point to the k8s Prow instance's config: https://github.com/kubernetes/test-infra/blob/a0a273a08a4c35e150c09fd8a7f558d4cc12c793/config/BUILD.bazel#L7

I agree that the containerized version is preferable though.

config/jobs/README.md
```diff
@@ -171,6 +171,18 @@ recognize this is going to depend on review latency.
...
+
+Normally prow will automatically schedule your job, however if for some reason you
+need to trigger it again you have a few options:
+
+- you can use the rerun feature in prow.k8s.io to run the job again *with the same config*
+- you can use [`config/mkpj.sh`](/config/mkpj.sh) to create a prowjob CR from your local config
+- you can use `bazel run //prow/cmd/mkpj -- --job=foo ...` to create a prowjob CR from your local config
+
+For the latter two options you'll need to submit the resulting CR via `kubectl` configured against
+the prow services cluster.
```
cjwagner: I think we need to highlight that this is only available to oncall members. This document is geared toward job authors, most of whom won't have permission to do this.

chases2: nit: Oncall members or other administrators. One target of these docs are people who run their own Prow instances.

BenTheElder: not config/, config/ is entirely full of things specifically for prow.k8s.io. this mkpj.sh is for that.

prow/ is supposed to be general but isn't fully yet (mainly prow/cluster/).

config/jobs/README.md
```diff
@@ -171,6 +171,18 @@ recognize this is going to depend on review latency.
 
 For more details, please refer to [How to Test a ProwJob](/prow/build_test_update.md#how-to-test-a-prowjob)
 
+## Running a Production Job
+
+Normally prow will automatically schedule your job, however if for some reason you
```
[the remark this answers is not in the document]
cjwagner: Do you mean in response to https://github.com/kubernetes/test-infra/pull/16241/files#r378583462?

I don't think this really makes it clear that this is not an option for the majority of readers.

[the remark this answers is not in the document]
BenTheElder: I wrote this comment right after pushing but it was pending submitting my review :/

I agree, will update later today.

PR #16243, opened 2020-02-12
Improve docs about build clusters and auto-deployment.

cjwagner:
Updating these docs to elaborate on some things after a discussion with @chizhg


cjwagner:
> This seems to rely entirely on people reading long paragraphs, when generally what people want to do is cut and paste code snippets. Paragraphs are rarely read or understood

The docs I touched discuss best practices so there aren't really any code snippets to offer, this is a higher level suggestion about how to arrange things and an explanation why. I do link to specifics about how to implement the suggested practices which include example configs/code that can be copied.
/hold cancel

PR #16246, opened 2020-02-12
Make alvaroaleman reviewer in boskos/

alvaroaleman:
Mostly so I get notified when changes happen and can take a stab at reviewing them.

/assign @ixdy @stevekuznetsov

### DIFF
--- a/boskos/OWNERS
+++ b/boskos/OWNERS
@@ -7,6 +7,7 @@ approvers:
 
 reviewers:
 - ixdy
+- alvaroaleman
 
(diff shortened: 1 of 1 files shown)


boskos/OWNERS
```diff
@@ -7,6 +7,7 @@ approvers:
 
 reviewers:
 - ixdy
+- alvaroaleman
```
ixdy: I think usually we try to keep these alphabetical, but otherwise 👍

PR #16253, opened 2020-02-12
Turn down the boskos-metrics deployment

ixdy:
The boskos-metrics server is redundant after #16196, so we can turn it down.
I've moved the `boskos-metrics` IP from this service to point to boskos core, replacing the `boskos-http-metrics` IP (which we can probably delete).

I'm guessing we'll need to manually delete the following resources from the k8s-prow-builds cluster, since I don't think kubectl apply will remove them:
- `deployment/boskos-metrics`
- `service/boskos-http-metrics`

/hold
~since we shouldn't merge this until the new code is live.~ (now live and working)

/area boskos


ixdy:
This should be good to go now - we're already pulling metrics from the boskos core service: https://monitoring.prow.k8s.io/d/wSrfvNxWz/boskos-resource-usage?orgId=1

fejta:
Deleted the service and deployment

PR #16255, opened 2020-02-12
Use real bazel version in bazelbuild images

fejta:
Otherwise we will get errors like https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/post-test-infra-upload-testgrid-config/1227675097968414720 when the image's bazel version does not match `.bazelversion`


alvaroaleman:
@fejta Rerun pretty please? :)

fejta:
oh duh the boskos thing will have the same problem, one sec

PR #16259, opened 2020-02-12
Revert boskos client changes

alvaroaleman:
/assign @ixdy 
/cc @stevekuznetsov


ixdy:
/hold

I'd like to see if there's a quick and easy way to roll-forward with the CRD config changes first.

ixdy:
The issue is when we upgrade an existing deployment; the existing CRDs use `ResourceCollection` and `DRLCCollection` for the `listKind`s, while the new code uses `ResourceObjectList` and `DRLCObjectList`, and so the caching client fails to initialize properly:
```
E0212 22:31:57.049418       1 reflector.go:125] external/io_k8s_client_go/tools/cache/reflector.go:98: Failed to list *crds.DRLCObject: no kind "DRLCCollection" is registered for version "boskos.k8s.io/v1" in scheme "pkg/runtime/scheme.go:101"
E0212 22:31:57.803985       1 reflector.go:125] external/io_k8s_client_go/tools/cache/reflector.go:98: Failed to list *crds.ResourceObject: no kind "ResourceCollection" is registered for version "boskos.k8s.io/v1" in scheme "pkg/runtime/scheme.go:101"
{"component":"boskos","error":"timeout waiting for cache sync","file":"boskos/cmd/boskos/boskos.go:96","func":"main.main","level":"fatal","msg":"unable to get client","time":"2020-02-12T22:31:57Z"}
```

From some simple local experiments, I think it's possible to update the `listKind`s in the CRD specs without affecting the existing CRs. Maybe it would be worth extracting the CRD specs from the code (into separate YAML specs that can be applied) so that we can easily update the specs at the same time we update the code?

ixdy:
We definitely can't roll forward without #16270, but I think this change works otherwise.

ixdy:
/close

We've fixed the issue by pulling the CRDs out of boskos code and then updating and applying them to the cluster.

PR #16261, opened 2020-02-12
Use workload identity to push prow and boskos

fejta:
Switch from using a secret file to a service account


fejta:
Dup of https://github.com/kubernetes/test-infra/pull/16260
/close

fejta:
This appears to be passing tests now?

Katharine:
/lgtm
/hold
feel free to unhold

fejta:
Will it push? https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/post-test-infra-push-prow/1228041459496652801

fejta:
Yes

PR #16264, opened 2020-02-13
Use workload-identity for testgrid-config-updater

fejta:
/assign @chases2 @michelle192837 

Use a service account rather than a secret file


fejta:
Deleted this secret

PR #16265, opened 2020-02-13
Use workload identity to deploy and push

fejta:
/assign @Katharine @cjwagner 

Need to fix image builder to not require a secret file, but can migrate the rest.
Need to also switch image pushing jobs to use the pusher account.


Katharine:
/lgtm
/hold

Feel free to unhold

PR #16266, opened 2020-02-13
boskos core: don't attempt to create CRDs

ixdy:
We recently had a brief outage because we changed some of the names in the Boskos CRDs, and updating these right now is tricky since Boskos attempts to create CRDs itself.

A simpler solution would be to just put the CRDs in YAMLs that can be applied with the rest of the Kubernetes resources, and not have Boskos attempt to create them itself. I'm working on that in a separate PR, to try to isolate concerns. (The existing code already does nothing if the CRDs already exist.)

Credit: @alvaroaleman already suggested doing this in https://github.com/kubernetes/test-infra/pull/16206#discussion_r377559436.

/assign @alvaroaleman @stevekuznetsov


alvaroaleman:
It has conflicts with your other PR :grimacing:

prow/cluster/boskos.yaml
```diff
@@ -18,7 +18,7 @@ metadata:
 rules:
 - apiGroups: ["apiextensions.k8s.io"]
   verbs: ["*"]
-  resources: ["customresourcedefinitions"]
+  resources: ["customresourcedefinitions"]  # TODO(ixdy): remove once Boskos no longer attempts to create CRDs itself
```
fejta: Prefer an issue over a user

ixdy: Done.

PR #16267, opened 2020-02-13
Add Boskos CRDs to prow/cluster, and bump Boskos back to latest

ixdy:
This is a revert of #16260, along with the required config changes so that Boskos hopefully won't crash.

This goes along with #16266, but it shouldn't be necessary to wait for it, since Boskos is not supposed to do anything if the CRDs already exist.

/hold
because I'm still validating these changes (and also we shouldn't merge this late in the day)

/assign @fejta 
cc @alvaroaleman 

/area boskos

### DIFF
--- a/prow/cluster/boskos.yaml
+++ b/prow/cluster/boskos.yaml
@@ -11,6 +11,43 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: dynamicresourcelifecycles.boskos.k8s.io
+spec:
+  group: boskos.k8s.io
+  names:
+    kind: DRLCObject
+    listKind: DRLCCollection
+    plural: dynamicresourcelifecycles
+    singular: dynamicresourcelifecycle
+  scope: Namespaced
+  version: v1
+  versions:
+  - name: v1
+    served: true
+    storage: true
+---
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: resources.boskos.k8s.io
+spec:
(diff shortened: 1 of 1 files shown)


ixdy:
Testing this locally in kind, this seems to do the right thing for the CRDs, but it seems like boskos is now failing to start for a different reason:
```
{"component":"boskos","error":"listen tcp :8080: bind: address already in use","file":"prow/interrupts/interrupts.go:156","func":"k8s.io/test-infra/prow/interrupts.ListenAndServe.func1","level":"info","msg":"Server exited.","time":"2020-02-13T01:31:20Z"}
```

ixdy:
Something is causing prometheus to serve metrics for [`rest_client_request_duration_seconds`](https://github.com/kubernetes/component-base/blob/e34e74d72164decb088310cc02f2b189e76c9178/metrics/prometheus/restclient/metrics.go#L30-L39) on port 8080. Still investigating...

ixdy:
found it: https://github.com/kubernetes-sigs/controller-runtime/blob/a7c8a93c1cf395974911ef0ece977a7d27b2bf4b/pkg/metrics/listener.go#L24-L26

ixdy:
We definitely can't roll forward without #16270, but I think this change would work otherwise.

ixdy:
Updated boskos to v20200213-7790e5cec. This includes #16270 and #16271. I think this works.

alvaroaleman:
Yeah saw it, I want us (openshift) to do it first, because there I can see logs if something goes south again. Will lgtm once that was successful

alvaroaleman:
Yup, works fine with the CRD update. Does bazel do that automatically if you just add the file?
/lgtm

alvaroaleman:
And thanks a lot for taking care here @ixdy

fejta:
Looks like it isn't crash looping

ixdy:
[Monitoring](https://monitoring.prow.k8s.io/d/wSrfvNxWz/boskos-resource-usage?orgId=1) appears to be working again too.

Issue #16223, opened 2020-02-11
Use `Repo` struct instead of `org/repo` string


matthyx:
**What should be cleaned up or changed**:
Use `Repo` struct instead of `org/repo` string between test-infra components.

**Provide any links for context**:
https://github.com/kubernetes/test-infra/pull/16184#pullrequestreview-356477033

matthyx:
@alvaroaleman I see `tide` relies a lot on "org/repo" strings... not sure we want to touch all that?
Also I will check if some parts of `hook` can be simplified with that struct.

alvaroaleman:
> @alvaroaleman I see tide relies a lot on "org/repo" strings... not sure we want to touch all that?
Also I will check if some parts of hook can be simplified with that struct.

So without having looked at the tide code, the intuitive answer would be "Depends on if we ever split these". If we just always use `org/repo` strings everywhere and never split or join them, then its probably not worth the effort.

matthyx:
I think we're done for now. Thanks for your help Alvaro!

PR #16277, opened 2020-02-13
Replace "org/repo" string with struct

matthyx:
Relates to #16223

This is the first commit of the work:
- add new `Repo` struct
- use `Repo` in helpProvider

Subsequently I will replace "org/repo" in other places, but I wanted to keep this PR manageable.

/cc @alvaroaleman

### DIFF
--- a/prow/pluginhelp/hook/hook.go
+++ b/prow/pluginhelp/hook/hook.go
@@ -83,7 +83,12 @@ func (ha *HelpAgent) generateNormalPluginHelp(config *plugins.Configuration, rev
 			ha.log.Warnf("No help is provided for plugin %q.", name)
 			continue
 		}
-		help, err := provider(config, revMap[name])
+		var enabledRepos []plugins.Repo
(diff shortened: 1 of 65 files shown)


prow/pluginhelp/hook/hook.go
```diff
@@ -83,7 +83,12 @@ func (ha *HelpAgent) generateNormalPluginHelp(config *plugins.Configuration, rev
 			ha.log.Warnf("No help is provided for plugin %q.", name)
 			continue
 		}
-		help, err := provider(config, revMap[name])
+		var enabledRepos []plugins.Repo
+		for _, repo := range revMap[name] {
```
alvaroaleman: Whats the reason for not using the `Repo` struct as values in the `revMap`?

matthyx: I just wanted to focus on the help provider in this PR. Then I will refine in specific areas like here and some parts of Config

PR #16283, opened 2020-02-13
[Test] Add docker in docker preset in KFP post submit test

rui5i:
This PR adds docker in docker preset in KFP post submit test. 

Part of https://github.com/kubeflow/pipelines/issues/2950


jlewi:
@rui5i did you talk with the prow team to see whether running docker in docker is supported on prow clusters? I think that requires mounting the docker socket into the container which is a security risk.

rui5i:
Hi @jlewi , I followed this example: https://github.com/kubernetes/test-infra/blob/a0a273a08a4c35e150c09fd8a7f558d4cc12c793/config/prow/config.yaml#L644 provided by prow team. I think this is supported by prow.

fejta:
Can you do this in your own cluster instead?

PR #16284, opened 2020-02-13
approve: add debug logging

stevekuznetsov:
We are seeing events where the approval plugin never takes action, or
takes quite a long time to do so. Often there are no logs whatsoever in
these cases. This change adds some debug logging to the cases where the
plugin would have exited without logging anything at all in order to
help us identify why such actions occur.

/assign @cjwagner @alvaroaleman @petr-muller

### DIFF
--- a/prow/plugins/approve/approve.go
+++ b/prow/plugins/approve/approve.go
@@ -165,6 +165,7 @@ func handleGenericCommentEvent(pc plugins.Agent, ce github.GenericCommentEvent) 
 
 func handleGenericComment(log *logrus.Entry, ghc githubClient, oc ownersClient, githubConfig config.GitHubOptions, config *plugins.Configuration, ce *github.GenericCommentEvent) error {
 	if ce.Action != github.GenericCommentActionCreated || !ce.IsPR || ce.IssueState == "closed" {
+		log.Debug("Event is not a creation of a comment on an open PR, skipping.")
 		return nil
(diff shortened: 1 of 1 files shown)


prow/plugins/approve/approve.go
```diff
@@ -236,15 +241,18 @@ func handleReview(log *logrus.Entry, ghc githubClient, oc ownersClient, githubCo
 	// genericCommentEventHandler handle this event. Approval commands override
 	// review state.
 	if isApprovalCommand(botName, opts.LgtmActsAsApprove, &comment{Body: re.Review.Body, Author: re.Review.User.Login}) {
+		log.Debug("Review constitutes approval, skipping event.")
```
cjwagner: nit: This is a little confusing without the context of the comment.
/shrug

stevekuznetsov: Eh, we get the line number in the message, figured it would be fine

stevekuznetsov:
470 has a log?

```go
		events, err := ghc.ListIssueEvents(org, repo, number)
		if err != nil {
			log.WithError(err).Errorf("Failed to list issue events for %s/%s#%d.", org, repo, number)
			return false
		}
```

/hold cancel

cjwagner:
> 470 has a log?

I meant the other 470 in the diff (the `return nil` at the end of the handle function).

PR #16290, opened 2020-02-13
Add seans3 and monopole to sig-cli testgrid OWNERS

monopole:
We're expanding test coverage in sig-cli
as we add more code to cli-utils and move kustomize libs.


monopole:
@soltysh is this OK with you?

@seans3

PR #16294, opened 2020-02-14
Move hook to Repo struct for enabledRepos

matthyx:
Relates to #16223

We continue the migration...

/assign @alvaroaleman

### DIFF
--- a/prow/external-plugins/cherrypicker/server.go
+++ b/prow/external-plugins/cherrypicker/server.go
@@ -57,7 +57,7 @@ type githubClient interface {
 }
 
 // HelpProvider construct the pluginhelp.PluginHelp for this plugin.
-func HelpProvider(enabledRepos []plugins.Repo) (*pluginhelp.PluginHelp, error) {
+func HelpProvider(_ []plugins.Repo) (*pluginhelp.PluginHelp, error) {
 	pluginHelp := &pluginhelp.PluginHelp{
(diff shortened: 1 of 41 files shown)


prow/external-plugins/cherrypicker/server.go
```diff
@@ -57,7 +57,7 @@ type githubClient interface {
 }
 
 // HelpProvider construct the pluginhelp.PluginHelp for this plugin.
-func HelpProvider(enabledRepos []string) (*pluginhelp.PluginHelp, error) {
+func HelpProvider(enabledRepos []plugins.Repo) (*pluginhelp.PluginHelp, error) {
```
alvaroaleman: Nit: Since the signature is the only change, I presume the var is unused, what about using `(_ []plugins.Repo)` in the signature? (I really wish go would enforce that)

matthyx: What about other places where it's not used like:
https://github.com/matthyx/test-infra/blob/c0631fc87e7fb879d28aaa3e8eccd53a22e593dd/prow/plugins/yuks/yuks.go#L48
or:
https://github.com/matthyx/test-infra/blob/b2359d3c38374d7e050538b2c78cc14b2f74d501/prow/pluginhelp/externalplugins/externalplugins.go#L37

alvaroaleman: I would make the same comment there :)

matthyx:
@alvaroaleman PTAL

alvaroaleman:
/lgtm
/hold cancel

Thanks again!

PR #16307, opened 2020-02-14
Increase Cluster API verify presumbmit resources

vincepri:


detiber:
Not a blocking comment, but do we actually think we need 6 cores? I suspect the issue is probably related to noisy neighbors (that aren't setting requests or requests too low).

PR #16308, opened 2020-02-14
Simplify gencred instructions.

fejta:
Only advertise the code paths we want poeple to use.
/assign @clarketm @cjwagner


clarketm:
tbh, not 100% enthused about deleting useful, general info about `gencred`'s capabilities. Personally, I like to be comprehensive (or at least link to a comprehensive source in another file for completeness). 

But if the docs are confusing to you and others as a first-time reader, then I cannot argue with that logic. I will defer to @cjwagner on the lgtm since he is much better than me with writing intuitive, user-friendly documentation.

gencred/README.md
```diff
@@ -1,241 +1,60 @@
...
-
-### Library
-
-#### Generate a service account token for a cluster. 
-✅ **PREFERRED** method.
-
-```go
-// Import serviceaccount
-import "k8s.io/test-infra/gencred/pkg/serviceaccount"
+The most relevant parts here are `token:` and `certificate-authority-data:`.
```
cjwagner: The `context` section seems equally relevant. It is unclear to me why this is included at all.

PR #16310, opened 2020-02-14
[WIP] Use kaniko for pushing images and caching

julianvmodesto:
https://cloud.google.com/cloud-build/docs/kaniko-cache

/wg k8s-infra


BenTheElder:
our image builds haven't exactly had speed issues, how much do we trust this cache correctness ..?

PR #16311, opened 2020-02-15
Gencred: require --context; do not default to current-context

clarketm:


gencred/README.md
```diff
@@ -26,7 +26,7 @@ The following is a list of supported options for the `gencred` CLI. All options
 
 ```console
   -c, --certificate      Authorize with a client certificate and key.
-      --context string   The name of the kubeconfig context to use.
+      --context string   The name of the kubeconfig context to use. (default current-context)
```
fejta: Using the current context is a footgun, can we just remove it?

clarketm: At the expense of additional verbosity in some cases, yes, we can definitely make the context a *required* options.

fejta: I'm more concerned about potentially creating admin credentials for the wrong cluster than avoiding the hassle and typing of determining the correct context.

clarketm: Done and added unit test(s) to verify.

PR #16315, opened 2020-02-15
Metrics: Also expose controller-runtime metrics

alvaroaleman:
This PR changes the metrics package to use a prometheus.Gatherers
to combine the metrics from the prometheus default registry and
and from the controller-runtime registry.

This gives us some nice additional metrics, some of which are even
reported when c-r is not used, like the ones for workqueues.

### DIFF
--- a/prow/metrics/metrics.go
+++ b/prow/metrics/metrics.go
@@ -25,6 +25,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/sirupsen/logrus"
+	ctrlruntimemetrics "sigs.k8s.io/controller-runtime/pkg/metrics"
 
 	"k8s.io/test-infra/prow/config"
 	"k8s.io/test-infra/prow/interrupts"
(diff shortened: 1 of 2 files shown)


prow/metrics/metrics.go
```diff
@@ -53,9 +54,15 @@ func ExposeMetrics(component string, pushGateway config.PushGateway) {
...
+		handler = promhttp.HandlerFor(
+			prometheus.Gatherers{prometheus.DefaultGatherer, ctrlruntimemetrics.Registry},
+			promhttp.HandlerOpts{},
+		)
 	} else {
-		handler = promhttp.HandlerFor(reg, promhttp.HandlerOpts{})
+		handler = promhttp.HandlerFor(
+			prometheus.Gatherers{reg, ctrlruntimemetrics.Registry},
+			promhttp.HandlerOpts{},
+		)
```
petr-muller: nit: I think this would be less verbose and therefore better readable:

```
var gatherer ..
if reg == nil {
  gatherer = p.DefaultGatherer
} else {
  gatherer = reg
}
handler := ...Gatherers{gatherer, ...
```

Maybe even change `serveMetrics` to accept a `prometheus.Gatherer` right away?

/hold
Fix or unhold, your choice.

alvaroaleman: Good point, fixed

PR #16316, opened 2020-02-15
Crier github reporter: Allow using more than one worker

alvaroaleman:
This is achieved by doing pr-level locking for presubmit jobs.

Fixes https://github.com/kubernetes/test-infra/issues/13306

### DIFF
--- a/prow/crier/reporters/github/reporter.go
+++ b/prow/crier/reporters/github/reporter.go
@@ -19,6 +19,13 @@ limitations under the License.
 package github
 
 import (
+	"errors"
+	"fmt"
+	"sync"
+	"time"
+
+	"github.com/sirupsen/logrus"
+
 	"k8s.io/test-infra/prow/apis/prowjobs/v1"
 	"k8s.io/test-infra/prow/config"
 	"k8s.io/test-infra/prow/gerrit/client"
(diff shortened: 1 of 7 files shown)


prow/crier/reporters/github/reporter.go
```diff
@@ -35,6 +39,21 @@ type Client struct {
...
 	reportAgent v1.ProwJobAgent
+	prLocks     *shardedLock
+}
+
+type shardedLock struct {
+	mapLock *sync.Mutex
+	locks   map[string]*sync.Mutex
+}
+
+func (s *shardedLock) getLock(key string) *sync.Mutex {
```
alvaroaleman: I should probably mention that this is a memory leak, as we allocate a lock for each unique PR for which we observe a job and can never clean it up again. I would think though that this is not going to allocate so much memory that this will be a problem. WDYT?

prow/crier/reporters/github/reporter.go
```diff
@@ -68,6 +91,33 @@ func (c *Client) ShouldReport(pj *v1.ProwJob) bool {
...
+	if pj.Spec.Type != v1.PresubmitJob {
+		return "", fmt.Errorf("can only get lock key for presubmit jobs, was %q", pj.Spec.Type)
+	}
+	if pj.Spec.Refs == nil {
+		return "", errors.New("pj.Spec.Refs is nil")
+	}
+	if n := len(pj.Spec.Refs.Pulls); n != 1 {
+		return "", fmt.Errorf("prowjob doesn't have one but %d pulls", n)
+	}
+	return fmt.Sprintf("%s/%s#%d", pj.Spec.Refs.Org, pj.Spec.Refs.Repo, pj.Spec.Refs.Pulls[0].Number), nil
```
petr-muller: Is there a good reason for turning this into a string instead of having a simple triplet struct to serve as a map key?

alvaroaleman: Yeah, I don't need to define a struct. What would be the advantage of doing that over the string, one `printf` computation less?

prow/crier/reporters/github/reporter.go
alvaroaleman: Under the shower I realized that we can actually free them, by just acquiring the map lock, then acquiring the locks themselves, then deleting them. This essentially prevents all presubmit reporting from happening so shouldn't be done too often, but no reason to not do that.

I'll fix up the PR to do that once I find some time.

petr-muller: SGTM!

prow/crier/reporters/github/reporter.go
petr-muller: I don't care about the performance of `Sprintf` (or even the key hashing in the map). I just like to use type system. Having an explicit simple struct to serve as a key IMO makes all these function signatures carry more information about their purpose. 

I don't insist.

prow/crier/reporters/github/reporter.go
alvaroaleman: Added a cleanup for the locks

prow/crier/reporters/github/reporter.go
alvaroaleman: Added it.

alvaroaleman:
This has been sitting uncommented except for your remarks for 10 days, I think it should be good.

cc @michelle192837 as test-infra oncall, since this PR increases the worker number of crier without the version supporting it being deployed yet, it will result in a warning that goes away once you bump Prow.

/hold cancel

PR #16317, opened 2020-02-15
refactor io package and migrate tide & status-reconciler

sbueringer:
First steps to [S3 support](https://github.com/kubernetes/test-infra/issues/11260). This PR only refactors the pkg/io package and enables tide, gerrit and status-reconciler to use S3 storage. 

To use it you have to use another secret for tide & status-reconciler with the following content:
```
{
  "region": "us-east-2",
  "access_key": "keyhere",
  "secret_key": "secrethere",
  "endpoint": "s3.amazonaws.com",
  "insecure": false,
  "s3_force_path_style": true,
  "bucket": "s3://s3bucketname"
}
```

Now you should be able to use paths like the following in your YAMLs:
* Tide
   ```
   - --status-path=s3://tide/tide-status
   - --history-uri=s3://tide/tide-history.json
   ```
* status-reconciler
    ````
    - --status-path=s3://status-reconciler/status-reconciler-status
    ````
* gerrit (I'm not using this in my environment so I couldn't test it)
    ````
   - --last-sync-fallback=s3://path
    ````

Notes:
* I kept the whole implementation of the io package in this PR. A few of the methods are still unused with this PR. I'm not sure if it's better to introduce them in a later PR or right now. Adding them right now provides a better overview over the final package.


pkg/io/providers/providers.go
```diff
@@ -0,0 +1,71 @@
...
+var (
+	storageProviderCreators        = map[string]StorageProviderCreator{}
+	storageProviderPathIdentifiers = map[string]StorageProviderPathIdentifiers{}
+)
+
+type StorageProvider interface {
+	ParseStoragePath(storagePath string) (bucket, relativePath string, err error)
+	GetBucket(ctx context.Context, bucketName string) (*blob.Bucket, error)
+	SignedURL(ctx context.Context, bucketName, relativePath string, opts *blob.SignedURLOptions) (string, error)
+}
```
alvaroaleman: Can you explain why this interface has to exist? `ParseStoragePath` is a wrapper around `util.ParseStoragePath` for all three providers, `SignedURL` is a wrapper around `GetBucket().SignedURL()`.

Generally I would love it if we wouldn't have to write code for every provider we want to support. We could for example achieve that by extending prow/flagutil and switch based on auth flag, e.G. none -> directly call `gocloud.dev/blob.OpenBucket`, `--gcs-credentials-file` is set -> read it, constuct a gcp client, call `gcsblob.OpenBucket` etc.

That way we would support all providers that are supported by `gocloud.dev/blob` via their default auth mechanism without having to write any code for it.

pkg/io/opener.go
```diff
@@ -17,106 +17,392 @@ limitations under the License.
 package io
```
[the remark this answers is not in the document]
sbueringer: Sounds good to me

pkg/io/providers/providers.go
sbueringer: Hm, I think it comes down to if we can move the code which is different here to other places or if it is so generic that it just works the same for every provider.

* ParseStoragePath
  * it was the same in all 3 providers at the time of your review. But in the meantime (through unit tests) I found out that file must be handled differently. In the s3/gs case we can just strip the prefix. The first segment is the bucket everything else the relative Path. In the file case we have to split in directory and file. Of course this can also be handled by one util and doesn't have to be implemented in the provider

  * GetBucket: works like you described. But this more or less just means putting the code in the flagutil instead of the io package.
  
  * SignedURL: Works the same for s3 and file. Prow currently also uses anonymous access to public GCS buckets, so that differs. Of course the differences can also be handled in a util.

Overall I think we can move each of these 3 methods to 3 functions of the io package (or somewhere else e.g. flagutil). But I think we still have to write code for each provider. I think right now I would prefer keeping the code in the `io` (or `io2`). Because it's similar to the current implementation and it's probably easier to refactor. Of course this doesn't mean we necessarily have to keep the provider interface.

pkg/io/providers/file/file.go
```diff
@@ -0,0 +1,59 @@
+package file
+
+import (
+	"context"
+	"fmt"
+	"gocloud.dev/blob"
+	"gocloud.dev/blob/fileblob"
+	"k8s.io/test-infra/pkg/io/providers"
```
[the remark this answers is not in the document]
sbueringer: Done

pkg/io/providers/providers.go
alvaroaleman: Both the special case of `file` not having a bucket portion and `SignedURL` used anonymously can be covered by just using `gocloud.dev/blob.OpenBucket` or not?

I guess my main point is that I do not want all providers to implement some method just because one provider happens to have a special case and that I find the current approach of registering providers way too complicated.

We could for example make the providers export an implementation of
```
type StorageProviderFactory interface {
    AddFlags(fs *flag.FlagSet)
    ValidateFlags() error
    // True if any auth flag for this provider was set
    IsEnabled() bool
    Bucket(path string)(*gocloud.dev/blob.Bucket, error)
}
```

Then iterate over all of them, error if more than one `IsConfigured` and use `gocloud.dev/blob.OpenBucket` if none is configured (e.G. in case of file or unauthenticated gcs).

This allows to keep provider-specifc details like `What auth flags in what combination can be used` and `how do I parse paths` private to themselves and have a fallthrough to the generic `OpenBucket`, allowing to use any supported provider by their default auth method/anonymously.

The other special cases of no prefix => gcs, gcs => gs should IMHO be handled by defaulting (url has no `://` component, prefix it with `gs://`) and replacing (url starts with `gcs://`, replace that with `gs://`)

WDYT?

prow/cmd/tide/main.go
```diff
@@ -38,6 +38,9 @@ import (
 	"k8s.io/test-infra/prow/metrics"
 	"k8s.io/test-infra/prow/pjutil"
 	"k8s.io/test-infra/prow/tide"
+
+	// Import storage providers
+	_ "k8s.io/test-infra/pkg/io/provider-imports"
```
alvaroaleman: Would be nicer to have `pkg/io`/`io2` import `provider-imports` or the providers directly instead of forcing each cmd to do that

pkg/io/opener.go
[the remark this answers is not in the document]
sbueringer: I moved the package and restored the old io package. I just realized again that the old package is completely unused now, because the only components which did use it where tide, gerrit and status-reconciler. The other components which will be migrated in follow-up PRs aren't using the `io` or `io/v2` package yet.

prow/cmd/tide/main.go
sbueringer: You're absolutely right. I guess I had some circular dependencies in some version of this PR and then later on I didn't see that I could simplify it again.

sbueringer: In the newest commits the import won't be necessary at all anymore.

pkg/io/providers/providers.go
sbueringer: I think I just realized that we have to decide on the following point and then a few things follow from that:

Currently I implemented the package so that we don't have to know which storage provider is used when creating the Opener. So we don't need different flags in the future for GCS/S3/Azure.. and we don't have to add (and propagate) flags for every new provider. We could just use `blob-store-credentials-file` (after a migration from `gcs-credentials-file`). When we then later on access the Opener with a path like `gs://` or `s3://` the correct provider is used based on the path. (see also: https://docs.google.com/document/d/13O1zRX5rpmPCqsPzukavRVqyh4pLp95IeIBVWAZx6Oo/edit?disco=AAAADsYsZxI)

So I think it would be a good idea to decide if we want to have different flags for different providers (or a flag signaling which provider should be used) or just one flag `blob-store-credentials-file`. This would affect the following components:
* deck, gerrit, status-reconciler, tide, gcupload, artifact-uploader, initupload, sidecar
I think this means we have to add the flag(s) to `prowjobs/v1.DecorationConfig` and propagate the flag(s) from the ProwJob to our test Pods. I guess that's why @fejta proposed to [differentiate based on path](https://docs.google.com/document/d/13O1zRX5rpmPCqsPzukavRVqyh4pLp95IeIBVWAZx6Oo/edit?disco=AAAADsYsZxI). I'm not sure if we really should add different flags for different storage providers.


> Both the special case of `file` not having a bucket portion and `SignedURL` used anonymously can be covered by just using `gocloud.dev/blob.OpenBucket` or not?

The difference in parsing the bucket name in `file` could be handled initially by parsing it when using OpenBucket (just fyi `file`
has a bucket portion it's just not the first segment it's the whole directory). I tried OpenBucket via `bucket, err := blob.OpenBucket(context.Background(), "gs://k8s-conform-capi-openstack")`
but it returns the following error: `open bucket gs://k8s-conform-capi-openstack: google: could not find default credentials. See https://developers.google.com/accounts/docs/application-default-credentials for more information.`. I guess that's why this was implemented like this (I only copied the code): https://github.com/kubernetes/test-infra/blob/b46003a27ce81f83c1754bbf1e0cb6765b97287b/pkg/io/v2/providers/gcs/gcs.go#L95-L102


> I guess my main point is that I do not want all providers to implement some method just because one provider happens to have a special case and that I find the current approach of registering providers way too complicated.

> We could for example make the providers export an implementation of
>
> ```
> type StorageProviderFactory interface {
>     AddFlags(fs *flag.FlagSet)
>     ValidateFlags() error
>     // True if any auth flag for this provider was set
>     IsEnabled() bool
>     Bucket(path string)(*gocloud.dev/blob.Bucket, error)
> }
> ```

I refactored it (currently mostly still the same interface) in the export-style and it looks far better imho. Thx! If you want
commit > https://github.com/kubernetes/test-infra/pull/16317/commits/47b2e6b76031013f871b9acb7fda511d99e4aeea
I can also remove the providers now and just implement GetBucket & SignedURL in the `providers.go` once. I implement this in another commit
so I can rollback more easily
commit > upcoming

> Then iterate over all of them, error if more than one `IsConfigured` and use `gocloud.dev/blob.OpenBucket` if none is configured (e.G. in case of file or unauthenticated gcs).

As mentioned above:
 * OpenBucket won't work with unauthenticated GCS
 * Let's decide if we want to use separate flags for separate providers. Then it's a whole other discussion :)

> This allows to keep provider-specifc details like `What auth flags in what combination can be used` and `how do I parse paths` private to themselves and have a fallthrough to the generic `OpenBucket`, allowing to use any supported provider by their default auth method/anonymously.
>
> The other special cases of no prefix => gcs, gcs => gs should IMHO be handled by defaulting (url has no `://` component, prefix it with `gs://`) and replacing (url starts with `gcs://`, replace that with `gs://`)

> WDYT?

sbueringer:
@alvaroaleman If we aggree on that the last commit is the best version yet. Let's "restart" the review :)

I think that would leave us with the following issues:
* can/should we open the bucket in `flagutils` already and hand it over to the opener
* this highly depends on if we want to discover the storage providers based on command line flags or configured paths

I think using a common credential file, no additional flags and handling the differences in the `io/v2` package is the easiest / most straightforward way.

Minor issue:
* not sure if I should also name the new package v2 if the old one is not used anymore (but both variants are okay for me)

pkg/io/v2/providers/providers.go
```diff
@@ -0,0 +1,235 @@
...
+	bkt, err := gcsblob.OpenBucket(ctx, client, bucketName, nil)
+	if err != nil {
+		return nil, fmt.Errorf("error opening GCS bucket: %v", err)
+	}
+	return bkt, nil
+}
+
+type s3Credentials struct {
+	Region           string `json:"region"`
+	Bucket           string `json:"bucket"`
```
alvaroaleman: The bucket is IMHO not part of the credential

pkg/io/v2/providers/providers.go
```diff
@@ -0,0 +1,235 @@
...
+	Endpoint         string `json:"endpoint"`
```
alvaroaleman: Shouldn't this get discovered when AWS S3 service is used (as opposed to S3-compatible self-hasted service)?

pkg/io/v2/providers/providers.go
```diff
@@ -0,0 +1,235 @@
...
+
+	switch storageProvider {
+	case providerFile, providerS3:
+		bucket, err := GetBucket(ctx, credentials, storagePath)
+		if err != nil {
+			return "", err
+		}
+		return bucket.SignedURL(ctx, relativePath, opts)
+	case providerGS:
+		if len(credentials) == 0 {
```
alvaroaleman: Why do we implement handling for unauthenticated usage here rather than in `GetBucket`?

pkg/io/v2/providers/providers.go
```diff
@@ -0,0 +1,235 @@
...
+
+	bkt, err := s3blob.OpenBucket(ctx, sess, bucketName, nil)
+	if err != nil {
+		return nil, fmt.Errorf("error opening S3 bucket: %v", err)
+	}
+	return bkt, nil
+}
+
+func SignedURL(ctx context.Context, credentials []byte, storagePath string, opts *blob.SignedURLOptions) (string, error) {
+	storageProvider, bucketName, relativePath, err := ParseStoragePath(storagePath)
```
alvaroaleman: This func (if it needs to exist in the first place, see other comment), could be drastically simplified by only having provider-specific handling for google:

```
if !storageProvider == providerGS {
// Do GCS anonymous specific stuff
}
// GetBucket
// HandleErr
// Get Signed URL

File: pkg/io/v2/providers/providers.go [hunk 9]
```diff
@@ -0,0 +1,235 @@
...
+		if err != nil {
+			return "", err
+		}
+		return bucket.SignedURL(ctx, relativePath, opts)
+	default:
+		return "", fmt.Errorf("unknown storageProvider: %s", storageProvider)
+	}
+}
+
+func ParseStoragePath(storagePath string) (storageProvider, bucket, relativePath string, err error) {
```
alvaroaleman: This function is unused

pkg/io/v2/providers/providers.go
```diff
@@ -0,0 +1,235 @@
...
+	}
+
+	pathSplit := strings.Split(storagePath, "/")
+	if len(pathSplit) < 2 {
+		return "", "", "", fmt.Errorf("path %q is not a valid %s path", storagePath, storageProvider)
+	}
+	return storageProvider, pathSplit[0], path.Join(pathSplit[1:]...), nil
+}
+
+func PathHasStorageProviderPrefix(storagePath string) bool {
```
alvaroaleman: This is unused

pkg/io/v2/providers/providers.go
```diff
@@ -0,0 +1,235 @@
...
+	for _, spName := range storageProviders {
+		if strings.HasPrefix(storagePath, spName+storageSeparator) {
+			return true
+		}
+	}
+	return false
+}
+
+func URLHasStorageProviderPrefix(url string) bool {
```
alvaroaleman: This is unused

pkg/io/v2/providers/providers.go
```diff
@@ -0,0 +1,235 @@
...
+		if strings.HasPrefix(url, spName+urlSeparator) {
+			return true
+		}
+	}
+	return false
+}
+
+// EncodeStorageURL encodes storage path to URL,
+// e.g.: s3://prow-artifacts => s3/prow-artifacts
+func EncodeStorageURL(storagePath string) string {
```
alvaroaleman: This is unused

pkg/io/v2/providers/providers.go
```diff
@@ -0,0 +1,235 @@
...
+		if strings.HasPrefix(storagePath, spName+storageSeparator) {
+			return strings.Replace(storagePath, spName+storageSeparator, spName+urlSeparator, 1)
+		}
+	}
+	return storagePath
+}
+
+// DecodeStorageURL decodes storage URL to path,
+// e.g.: s3/prow-artifacts => s3://prow-artifacts
+func DecodeStorageURL(url string) string {
```
alvaroaleman: This is unsed

pkg/io/v2/opener.go
```diff
@@ -0,0 +1,363 @@
...
+)
+
+// Aliases to types in the standard library
+type (
+	ReadCloser  = io.ReadCloser
+	WriteCloser = io.WriteCloser
+)
+
+// Opener has methods to read, write paths, create signed URLs,...
+type Opener interface {
```
alvaroaleman: All the consumers of this interface currently use `Reader` and `Writer` only, right? Can we keep it to just those two until we have a consumer that needs more (Presumably never going to happen if status-reconciler, gerrit and tide are the only users of this)?

pkg/io/v2/opener.go
```diff
@@ -0,0 +1,363 @@
...
+func (o *opener) Reader(ctx context.Context, path string, opts *blob.ReaderOptions) (io.ReadCloser, error) {
+	bucket, relativePath, err := o.getBucket(ctx, path)
+	if err != nil {
+		return nil, err
+	}
+	reader, err := bucket.NewReader(ctx, relativePath, opts)
+	if err != nil {
+		return nil, err
+	}
+	return readerCloser{reader, bucket}, nil
```
alvaroaleman: Why not just return the `reader`? It implements `io.ReadCloser`

pkg/io/v2/opener.go
```diff
@@ -0,0 +1,363 @@
...
+func (o *opener) RangeReader(ctx context.Context, path string, offset, length int64, opts *blob.ReaderOptions) (io.ReadCloser, error) {
+	bucket, relativePath, err := o.getBucket(ctx, path)
+	if err != nil {
+		return nil, err
+	}
+	reader, err := bucket.NewRangeReader(ctx, relativePath, offset, length, opts)
+	if err != nil {
+		return nil, err
+	}
+	return readerCloser{reader, bucket}, nil
```
alvaroaleman: Why not just return the `reader`? It implements `io.ReadCloser`

pkg/io/v2/opener.go
```diff
@@ -0,0 +1,363 @@
...
+	}
+	reader, err := bucket.NewReader(ctx, relativePath, opts)
+	if err != nil {
+		return nil, err
+	}
+	return readerCloser{reader, bucket}, nil
+}
+
+type readerCloser struct {
+	r *blob.Reader
```
alvaroaleman: I dislike the embedded Reader because its unclear for which path it is. If the reader is needed, it should be requested on-demand from the bucket

pkg/io/v2/opener.go
```diff
@@ -0,0 +1,363 @@
...
+	}
+	writer, err := bucket.NewWriter(ctx, relativePath, opts)
+	if err != nil {
+		return nil, err
+	}
+	return &writerCloser{writer, bucket}, nil
+}
+
+type writerCloser struct {
+	w *blob.Writer
```
alvaroaleman: Similarly, lets embedd just the bucket and get the Writer when we need it to avoid ambiguity

alvaroaleman:
> @alvaroaleman If we agree on that the last commit is the best version yet. Let's "restart" the review :)

Yes :) This looks a lot better and simpler now. I guess my only left point would then be that we still never fall through to `gocloud.dev/blob.OpenBucket`. The reason I think this is important is that I think ppl will expect things like `export GOOGLE_APPLICATION_CREDENTIALS=<<credentials_file>>` and `AWS_ACCESS_KEY_ID` + `AWS_SECRET_ACCESS_KEY` to just work and thats IMHO a reasonable expectation.

> this highly depends on if we want to discover the storage providers based on command line flags or configured paths

Yeah so this is an interesting question. The reason I always assumed that we have provider-specific flags is that in both the status-reconciler and tide the flag is called `gcs-credentials-file`. On top of that though, the bucket obviously includes its type and that should at the end of the day be the canonical source of truth for "what provider do we use".

pkg/io/v2/providers/providers.go
sbueringer: Yup unused => I'll remove it

pkg/io/v2/providers/providers.go
sbueringer:
```
// An optional endpoint URL (hostname only or fully qualified URI)
// that overrides the default generated endpoint for a client. Set this
// to `""` to use the default generated endpoint.
```	
If it's not set, the default value of AWS is used. My personal use case is Minio so I need the property. So to me it looks like this works as expected. If the endpoint is set in the config, the specified endpoint is used, if not the one from AWS.

pkg/io/v2/providers/providers.go
sbueringer: Looks like gocloud (and I think also the Google GCS lib) is not able to do this:
```go
	newClient := &gcp.HTTPClient{
		Client: http.Client{
			Transport: &oauth2.Transport{
				Base: gcp.DefaultTransport(),
			},
		},
	}
	bkt, _ := gcsblob.OpenBucket(context.Background(), newClient, "gs://k8s-conformance-openstack", nil)
	test, err := bkt.SignedURL(context.Background(),"README.md", nil)
	fmt.Println(err)
```
Leads to this error:
```
blob (key "README.md") (code=Unknown): to use SignedURL, you must call OpenBucket with a valid Options.GoogleAccessID and exactly one of Options.PrivateKey or Options.SignBytes
```
Which is produced here: https://github.com/google/go-cloud/blob/master/blob/gcsblob/gcsblob.go#L598-L600

Looking at the documentation I think signed URLs are not really meanted to be used on public files with anonymous access: https://cloud.google.com/storage/docs/access-control/signing-urls-manually

pkg/io/v2/providers/providers.go
sbueringer: Sounds good. I'll adjust this and then move the method to the next PR

pkg/io/v2/opener.go
sbueringer: I though we also have to close the bucket

pkg/io/v2/providers/providers.go
alvaroaleman: Grepping to the code I see a single usage of `SignedURL` in prow/spyglass/gcsartifact_fetcher.goL77 and that very explicitly uses auth, so why is this needed?

sbueringer: I integrated the stuff from here:
https://github.com/kubernetes/test-infra/blob/master/prow/spyglass/gcsartifact_fetcher.go#L150-L157

But we can discuss/remove it on the next PR if you want

pkg/io/v2/opener.go
sbueringer: Yup. I'll remove the other (now) unused methods and reintroduce them in the next PRs

pkg/io/v2/opener.go
sbueringer: As discussed. I'll cache the bucket(s) and return the reader here

pkg/io/v2/opener.go
sbueringer: As discussed. I'll cache the bucket(s) and return the reader here

pkg/io/v2/providers/providers.go
sbueringer: Nope, it's still used by providers.GetBucket  and opener.GetBucket

sbueringer:
> > @alvaroaleman If we agree on that the last commit is the best version yet. Let's "restart" the review :)
> 
> Yes :) This looks a lot better and simpler now. I guess my only left point would then be that we still never fall through to `gocloud.dev/blob.OpenBucket`. The reason I think this is important is that I think ppl will expect things like `export GOOGLE_APPLICATION_CREDENTIALS=<<credentials_file>>` and `AWS_ACCESS_KEY_ID` + `AWS_SECRET_ACCESS_KEY` to just work and thats IMHO a reasonable expectation.
> 

I think I can implement it like this

```go
func GetBucket(ctx context.Context, credentials []byte, path string) (*blob.Bucket, error) {
	storageProvider, bucket, _, err := ParseStoragePath(path)
	if err != nil {
		return nil, err
	}
	if storageProvider == providerFile || len(credentials) == 0 {
		bkt, err := blob.OpenBucket(context.Background(), fmt.Sprintf("%s://%s", storageProvider, bucket))
		if err != nil {
			return nil, fmt.Errorf("error opening file bucket: %v", err)
		}
		return bkt, nil
	}
	
	switch storageProvider {
	case providerGS:
		return getGCSBucket(ctx, credentials, bucket)
	case providerS3:
		return getS3Bucket(ctx, credentials, bucket)
	default:
		return nil, fmt.Errorf("unknown storageProvider: %s", storageProvider)
	}
}
```
If no credentials are given, gocloud can try to auto-discover based on e.g.` gs://` google credentials, etc.. Every provider supported by gocloud would be supported as long as we blank-import it and as long as the credentials are given via one of the default ways of the used provider.

We then only have to write code for providers which should be supported via our `blog-store-credentials-file` flag.

prow/cmd/status-reconciler/main.go
```diff
@@ -73,6 +73,7 @@ func gatherOptions() options {
 	fs.StringVar(&o.configPath, "config-path", "/etc/config/config.yaml", "Path to config.yaml.")
 	fs.StringVar(&o.jobConfigPath, "job-config-path", "", "Path to prow job configs.")
 	fs.StringVar(&o.pluginConfig, "plugin-config", "/etc/plugins/plugins.yaml", "Path to plugin config file.")
+	fs.StringVar(&o.gcsCredentialsFile, "gcs-credentials-file", "", "File where Google Cloud authentication credentials are stored. Required for GCS writes.")
```
alvaroaleman: Lol so this never worked? I think we should start naming the flag provider-agnostic, something like `blob-storage-credential-file` and deprecate the old `gcs-credentials-file`  flag

prow/tide/history/history.go
```diff
@@ -45,19 +46,24 @@ type History struct {
 	sync.Mutex
 	logSizeLimit int
 
-	opener io.Opener
+	opener iov2.Opener
```
alvaroaleman: Nit: If we define the interface here, we should use that rather than `iov2.Opener`

pkg/io/v2/providers/providers.go
alvaroaleman: Okay, that makes sense. Can we please add this as godoc then?

pkg/io/v2/opener.go
```diff
@@ -0,0 +1,122 @@
...
+		cachedBuckets: map[string]*blob.Bucket{},
+	}, nil
+}
+
+func (o *opener) getBucket(ctx context.Context, path string) (*blob.Bucket, string, error) {
+	_, bucketName, relativePath, err := providers.ParseStoragePath(path)
+	if err != nil {
+		return nil, "", fmt.Errorf("could not get bucket: %w", err)
+	}
+	if bucket, ok := o.cachedBuckets[bucketName]; ok {
```
alvaroaleman: I think we should make this threadsafe by protecting all accesses to this map with a mutex

pkg/io/v2/providers/providers.go
```diff
@@ -0,0 +1,152 @@
...
+	storageSeparator = "://"
+
+	providerFile = "file"
+	providerGS   = "gs"
+	providerS3   = "s3"
+)
+
+var storageProviders = []string{providerFile, providerGS, providerS3}
+
+func GetBucket(ctx context.Context, credentials []byte, path string) (*blob.Bucket, error) {
```
alvaroaleman: Re
```
func GetBucket(ctx context.Context, credentials []byte, path string) (*blob.Bucket, error) {
	storageProvider, bucket, _, err := ParseStoragePath(path)
	if err != nil {
		return nil, err
	}
	if storageProvider == providerFile || len(credentials) == 0 {
		bkt, err := blob.OpenBucket(context.Background(), fmt.Sprintf("%s://%s", storageProvider, bucket))
		if err != nil {
			return nil, fmt.Errorf("error opening file bucket: %v", err)
		}
		return bkt, nil
	}
	
	switch storageProvider {
	case providerGS:
		return getGCSBucket(ctx, credentials, bucket)
	case providerS3:
		return getS3Bucket(ctx, credentials, bucket)
	default:
		return nil, fmt.Errorf("unknown storageProvider: %s", storageProvider)
	}
}
```
> If no credentials are given, gocloud can try to auto-discover based on e.g. gs:// google credentials, etc.. Every provider supported by gocloud would be supported as long as we blank-import it and as long as the credentials are given via one of the default ways of the used provider.

> We then only have to write code for providers which should be supported via our blog-store-credentials-file flag.

Yup, that sounds perfect. Please also add godocs to `GetBucket` explaining this and its args

prow/cmd/status-reconciler/main.go
sbueringer: I guess yes. But I also didn't find any example with the status-path flag in the repo :)

Then I'll also add the blob-storage-credentials-file flag in all affected binaries here

pkg/io/v2/providers/providers.go
sbueringer: Done

prow/cmd/status-reconciler/main.go
sbueringer: Done

prow/tide/history/history.go
sbueringer: Oh that was an oversight. Changed it in a few places but missed that one

pkg/io/v2/opener.go
sbueringer: Done

pkg/io/v2/providers/providers.go
sbueringer: Done

fejta:
Please do not refactor every binary at the same time.

Can we switch one binary to use iov2 and back it by a flag?

sbueringer:
> Will this continue to authenticate to GCP without passing explicit credentials?

Currently not. 
I think we have to decide if want to fall back to gocloud credential discovery or access the bucket unauthenticated in the gcs case. But sounds to me like we have to access it unauthenticated.

sbueringer:
> Please do not refactor every binary at the same time.
> 
> Can we switch one binary to use iov2 and back it by a flag? AKA if `--foo=true` then use iov2 else use the current io package

Is it good enough if I add the feature flag to all 3 components or start with one even with the flag?

sbueringer:
> > Will this continue to authenticate to GCP without passing explicit credentials?
> 
> Currently not.
> I think we have to decide if want to fall back to gocloud credential discovery or access the bucket unauthenticated in the gcs case. But sounds to me like we have to access it unauthenticated.

I have to correct myself. What I can do is the following:
```
func GetBucket(ctx context.Context, credentials []byte, path string) (*blob.Bucket, error) {
	storageProvider, bucket, _, err := ParseStoragePath(path)
	if err != nil {
		return nil, err
	}
	if storageProvider == providerGS {
		return getGCSBucket(ctx, credentials, bucket)
	}
...
}


// getGCSBucket opens a gocloud blob.Bucket based on given credentials in the Google
// credential JSON format (see documentation of GetBucket for an example)
func getGCSBucket(ctx context.Context, credentials []byte, bucketName string) (*blob.Bucket, error) {
	client, err := getGCPHTTPClient(ctx, credentials)
	if err != nil {
		return nil, fmt.Errorf("error creating GCP Http Client: %v", err)
	}

	bkt, err := gcsblob.OpenBucket(ctx, client, bucketName, nil)
	if err != nil {
		return nil, fmt.Errorf("error opening GCS bucket: %v", err)
	}
	return bkt, nil
}

func getGCPHTTPClient(ctx context.Context, credentials []byte) (*gcp.HTTPClient, error) {
	if len(credentials) > 0 {
		googleCredentials, err := google.CredentialsFromJSON(ctx, credentials, storage.ScopeFullControl)
		if err != nil {
			return nil, fmt.Errorf("error getting Google credentials from JSON: %v", err)
		}
		return gcp.NewHTTPClient(gcp.DefaultTransport(),gcp.TokenSource(googleCredentials.TokenSource))
	}
	
	var creds *google.Credentials
	if os.Getenv("STORAGE_EMULATOR_HOST") != "" {
		creds, _ = google.CredentialsFromJSON(ctx, []byte(`{"type": "service_account", "project_id": "my-project-id"}`))
	} else {
		// We ignore the error because that just means we couldn't find default credentials
		creds, _ = gcp.DefaultCredentials(ctx)
	}
	if creds != nil {
		return gcp.NewHTTPClient(gcp.DefaultTransport(), creds.TokenSource)
	}
	return &gcp.HTTPClient{
		Client: http.Client{
			Transport: &oauth2.Transport{
				Base: gcp.DefaultTransport(),
			},
		},
	}, nil
}
```

This would implement all required cases imho. This means, in case of GCS we would:
* if we have credentials handed over (i.e. a credentials file): use the given credentials
* if we don't have credentials handed over:
  * Handle STORAGE_EMULATOR_HOST: https://github.com/google/go-cloud/issues/2718. Not sure if we want to support that
  * Then try to discover GCP default credentials
  * If all that fails/ doesn't match we would use an unauthenticated client

(The second part is equivalent to the implementation in gcsblob)

WDYT? @fejta @alvaroaleman

alvaroaleman:
> Will this continue to authenticate to GCP without passing explicit credentials?

Yes:
* If explicit credential file is set, we use that
* If not, we use `blob.OpenBucke` which
    * Uses this to open the bucket for gcs: https://github.com/google/go-cloud/blob/4338bbbd464f92223c6f2fe6168a433824906bc3/blob/gcsblob/gcsblob.go#L82
    * calls into `gcp.DefaultCredentials` to find the credentials: https://github.com/google/go-cloud/blob/4338bbbd464f92223c6f2fe6168a433824906bc3/blob/gcsblob/gcsblob.go#L105
    * which in turn calls `google.FindDefaultCredentials` that checks all the usual locations: https://godoc.org/golang.org/x/oauth2/google#FindDefaultCredentials

> I think we have to decide if want to fall back to gocloud credential discovery or access the bucket unauthenticated in the gcs case. But sounds to me like we have to access it unauthenticated.

Yeah as you mentioned, credential discovery and unauthenticated access are not mutually exclusive, we just have to build sth ourselves as gocloud never uses an unauthenticated client. But this is not a problem for this PR, the io package never supported unauthenticated access, so we shouldn't add until we actually need it.

sbueringer:
> > Will this continue to authenticate to GCP without passing explicit credentials?
> 
> Yes:
> 
> * If explicit credential file is set, we use that
> * If not, we use `blob.OpenBucke` which
>   
>   * Uses this to open the bucket for gcs: https://github.com/google/go-cloud/blob/4338bbbd464f92223c6f2fe6168a433824906bc3/blob/gcsblob/gcsblob.go#L82
>   * calls into `gcp.DefaultCredentials` to find the credentials: https://github.com/google/go-cloud/blob/4338bbbd464f92223c6f2fe6168a433824906bc3/blob/gcsblob/gcsblob.go#L105
>   * which in turn calls `google.FindDefaultCredentials` that checks all the usual locations: https://godoc.org/golang.org/x/oauth2/google#FindDefaultCredentials
> 
> > I think we have to decide if want to fall back to gocloud credential discovery or access the bucket unauthenticated in the gcs case. But sounds to me like we have to access it unauthenticated.
> 
> Yeah as you mentioned, credential discovery and unauthenticated access are not mutually exclusive, we just have to build sth ourselves as gocloud never uses an unauthenticated client. But this is not a problem for this PR, the io package never supported unauthenticated access, so we shouldn't add until we actually need it.

I think it did support unauthenticated access: https://github.com/kubernetes/test-infra/blob/master/pkg/io/opener.go#L57-L68

alvaroaleman:
> I think it did support unauthenticated access: https://github.com/kubernetes/test-infra/blob/master/pkg/io/opener.go#L57-L68

It does not, its just confusing code. Client gets set to `nil`, then `openGCS` bails out: https://github.com/kubernetes/test-infra/blob/41283ce58705c0980baa711b8f472af70840b8ec/pkg/io/opener.go#L88
and `openGCS` is used by both `Reader`: https://github.com/kubernetes/test-infra/blob/41283ce58705c0980baa711b8f472af70840b8ec/pkg/io/opener.go#L101 and `Writer`: https://github.com/kubernetes/test-infra/blob/41283ce58705c0980baa711b8f472af70840b8ec/pkg/io/opener.go#L114

pkg/io/v2/providers/providers.go
```diff
@@ -0,0 +1,215 @@
...
+// compatibility reasons.
+// File paths are split into a directory and a file. Directory is returned as bucket, file is returned
+// as relativePath.
+// For all other paths the first part is treated as storageProvider prefix, the second segment as bucket
+// and everything after the bucket as relativePath
+func ParseStoragePath(storagePath string) (storageProvider, bucket, relativePath string, err error) {
+	if strings.HasPrefix(storagePath, "gcs://") {
+		storagePath = strings.Replace(storagePath, "gcs://", "gs://", 1)
+	}
+	for _, spName := range storageProviders {
```
alvaroaleman: There is no reason the whole known providers dance, it doesn't matter if we know the provider or not and we should not error if its unknown, because that puts us buck to square one of only supporting the provider we wrote code for. Instead we should:

* Replace the `gcs` prefix with `gcs` as we already do
* If `storagePath` contains no `://` string, prefix `storagePath` with `gs://`
* Parse the storagePath using a default scheme of `<<provider-name, delimited by ://>>://<< bucket name, delimited by string end or slash>>[/<< optional relative path >>]`
* Handle the special case of `file` if `storageProvider == file`
* Error if `bucket` is an empty string
* Return `storageProvider` (which may be one we don't know), `bucket` and `relativePath`

pkg/io/v2/providers/providers.go
```diff
@@ -0,0 +1,215 @@
...
+//      "s3_force_path_style": true,
+//      "access_key": "access_key",
+//      "secret_key": "secret_key"
+//    }
+func GetBucket(ctx context.Context, credentials []byte, path string) (*blob.Bucket, error) {
+	storageProvider, bucket, _, err := ParseStoragePath(path)
+	if err != nil {
+		return nil, err
+	}
+	if storageProvider == providerFile || len(credentials) == 0 {
```
alvaroaleman: I think we should remove the first part of the condition, passing in credentials and using the file provider is invalid. Also, nitpicky comment, I find it easier to read if the `len(credential) == 0` case is at the bottom of the func, because its the default we fall through to.

pkg/io/v2/providers/providers.go
```diff
@@ -0,0 +1,215 @@
...
+// GetBucket opens and returns a gocloud blob.Bucket based on credentials and a path.
+// The path is used to discover which storageProvider should be used.
+//
+// If the storageProvider file is detected, we don't need any credentials and just open a file bucket
+// If no credentials are given, we just fall back to blob.OpenBucket which tries to auto discover credentials
+// e.g. via environment variables. For more details, see: https://gocloud.dev/howto/blob/
+//
+// If we specify credentials and an gs:// or s3:// path is used, credentials must be given in one of the
+// following formats:
+// * Google Cloud Storage (gs://):
```
alvaroaleman: I think its enough if we say that is has to be a service account key, that format is standardized in contrast to the S3 stuff

pkg/io/v2/providers/providers.go
sbueringer: Okay, how should we handle file paths. Currently file paths are specified like this `/tmp/path`. If I cannot handle it here. Every consumer has to detect if the path is a local path and has to add `file://`. Seem like that conflicts with: every path without prefix is a `gs://` path

I can only default to either file or gs. If i default to gs we have a breaking change for file paths

sbueringer:
> > I think it did support unauthenticated access: https://github.com/kubernetes/test-infra/blob/master/pkg/io/opener.go#L57-L68
> 
> It does not, its just confusing code. Client gets set to `nil`, then `openGCS` bails out:
> 
> https://github.com/kubernetes/test-infra/blob/41283ce58705c0980baa711b8f472af70840b8ec/pkg/io/opener.go#L88
> 
> 
> and `openGCS` is used by both `Reader`:
> https://github.com/kubernetes/test-infra/blob/41283ce58705c0980baa711b8f472af70840b8ec/pkg/io/opener.go#L101
> 
> and `Writer`:
> https://github.com/kubernetes/test-infra/blob/41283ce58705c0980baa711b8f472af70840b8ec/pkg/io/opener.go#L114

Ah yeah. Overlooked the `client = nil` part. So we don't have to handle the unauthenticated case and keep the code as is.

pkg/io/v2/providers/providers.go
alvaroaleman: Yeah it indeed does conflict. What consumers assume that unprefixed path is gs and what consumers assume its a file? If we have both, the consumer may need to do the defaulting.

sbueringer: Right now (in this PR) we only have consumers assuming unprefixed paths are local files. I'm not sure right now if we have future consumers which assume the same for gcs. That's why I changed it

pkg/io/v2/providers/providers.go
sbueringer: Done

pkg/io/v2/providers/providers.go
sbueringer: Done

pkg/io/v2/providers/providers.go
sbueringer: Done

pkg/io/v2/opener.go
```diff
@@ -0,0 +1,135 @@
...
+// ErrNotFoundTest can be used for unit tests to simulate NotFound errors.
+// This is required because gocloud doesn't expose its errors.
+var ErrNotFoundTest = fmt.Errorf("not found error which should only be used in tests")
+
+// IsNotExist will return true if the error shows that the object does not exist.
+func IsNotExist(err error) bool {
+	if err == ErrNotFoundTest {
+		return true
+	}
+	return gcerrors.Code(err) == gcerrors.NotFound
```
alvaroaleman: This also needs to return true if err is or contains(errors.As) an os.IsNotExist

prow/cmd/status-reconciler/main.go
sbueringer: I'll remove the old flag. As far as I can see there was no way to set it

pkg/io/v2/providers/providers.go
```diff
@@ -0,0 +1,198 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("error opening file bucket: %v", err)
+	}
+	return bkt, nil
+}
+
+// getGCSBucket opens a gocloud blob.Bucket based on given credentials in the Google
+// credential JSON format (see documentation of GetBucket for an example)
+func getGCSBucket(ctx context.Context, credentials []byte, bucketName string) (*blob.Bucket, error) {
+	googleCredentials, err := google.CredentialsFromJSON(ctx, credentials, storage.ScopeFullControl)
```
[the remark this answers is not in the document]
sbueringer: I'm not sure, but it's hardcoded into cloud.google.com/go/storage.NewClient which was used by the old opener: https://github.com/googleapis/google-cloud-go/blob/master/storage/storage.go#L104

That's why I did use the same as before. I assume it wouldn't really be a problem if I downgrade to `ScopeReadWrite`

pkg/io/v2/providers/providers.go
```diff
@@ -0,0 +1,198 @@
...
+// and everything after the bucket as relativePath.
+func ParseStoragePath(storagePath string) (storageProvider, bucket, relativePath string, err error) {
+	if strings.HasPrefix(storagePath, "/") {
+		storagePath = fmt.Sprintf("file://%s", storagePath)
+	}
+	if strings.HasPrefix(storagePath, "gcs://") {
+		storagePath = strings.Replace(storagePath, "gcs://", "gs://", 1)
+	}
+
+	parsedPath, err := url.Parse(storagePath)
```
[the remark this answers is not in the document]
sbueringer: Stole it from gocloud :)

pkg/io/v2/providers/providers.go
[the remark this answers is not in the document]
alvaroaleman: Should be fine if we used that before

prow/cmd/gerrit/main.go
```diff
@@ -237,7 +263,16 @@ func main() {
...
-	op, err := io.NewOpener(ctx, o.gcsCredentialsFile)
+	credentialsFile := o.blobStorageCredentialsFile
+	if credentialsFile == "" {
+		credentialsFile = o.gcsCredentialsFile
+	}
+	var op iov2.Opener
+	if o.useGenericBlobStorage {
+		op, err = iov2.NewOpener(credentialsFile)
+	} else {
+		op, err = io.NewOpener(ctx, credentialsFile)
```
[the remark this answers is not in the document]
sbueringer: I trim the prefix in the opener

pkg/io/v2/opener.go
sbueringer: I added the following. Guess errors.Is is also enough and you meant os.ErrNotExist:
```
	if errors.Is(err, os.ErrNotExist) {
		return true
	}
```

pkg/io/v2/providers/providers.go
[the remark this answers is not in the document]
sbueringer: Okay then I leave it as is

prow/cmd/tide/main.go
```diff
@@ -78,6 +87,12 @@ func (o *options) Validate() error {
 			return err
 		}
 	}
+	if o.gcsCredentialsFile != "" {
+		logrus.Error("-gcs-credentials-file is deprecated.  Use -blob-storage-credentials-file instead.")
```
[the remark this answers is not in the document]
sbueringer: Done

prow/cmd/gerrit/main.go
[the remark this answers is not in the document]
sbueringer: Done

prow/cmd/gerrit/main.go
```diff
@@ -68,8 +80,15 @@ func (o *options) Validate() error {
 		return errors.New("--last-sync-fallback must be set")
 	}
 
-	if strings.HasPrefix(o.lastSyncFallback, "gs://") && o.gcsCredentialsFile == "" {
-		logrus.WithField("last-sync-fallback", o.lastSyncFallback).Warn("--gcs-credentials-file unset, will try and access with a default service account")
+	if o.gcsCredentialsFile != "" {
+		logrus.Error("-gcs-credentials-file is deprecated.  Use -blob-storage-credentials-file instead.")
```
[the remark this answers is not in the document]
sbueringer: Done

pkg/io/v2/providers/providers_test.go
```diff
@@ -0,0 +1,132 @@
...
+			wantStorageProvider: "s4",
+			wantBucket:          "prow-artifacts",
+			wantRelativePath:    "pr-logs/bazel-build/test.log",
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			gotStorageProvider, gotBucket, gotRelativePath, err := providers.ParseStoragePath(tt.args.storagePath)
+			if (err != nil) != tt.wantErr {
+				t.Errorf("ParseStoragePath() error = %v, wantErr %v", err, tt.wantErr)
```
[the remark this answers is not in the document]
sbueringer: Done

pkg/io/v2/providers/providers.go
```diff
@@ -0,0 +1,198 @@
...
+		case providerGS:
+			return getGCSBucket(ctx, credentials, bucket)
+		case providerS3:
+			return getS3Bucket(ctx, credentials, bucket)
+		default:
+			return nil, fmt.Errorf("passing credentials is only supported for S3 and GCS not %s", storageProvider)
+		}
+	}
+
+	bkt, err := blob.OpenBucket(context.Background(), fmt.Sprintf("%s://%s", storageProvider, bucket))
```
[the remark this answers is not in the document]
sbueringer: Done

pkg/io/v2/opener.go
```diff
@@ -0,0 +1,135 @@
...
+}
+
+type opener struct {
+	credentials        []byte
+	cachedBuckets      map[string]*blob.Bucket
+	cachedBucketsMutex sync.Mutex
+}
+
+// NewOpener returns an opener that can read GCS, S3 and local paths.
+func NewOpener(credentialsFile string) (Opener, error) {
```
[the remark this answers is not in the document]
sbueringer: Done

pkg/io/v2/providers/providers_test.go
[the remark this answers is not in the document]
alvaroaleman: (The reason for the fatalf is that you do not need the return anymore, it will end the testcase right there)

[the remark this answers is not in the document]
sbueringer: Ah got it

alvaroaleman:
@fejta any chance you can have a look?

pkg/io/v2/providers/providers.go
```diff
@@ -0,0 +1,192 @@
...
+	}
+
+	bkt, err := s3blob.OpenBucket(ctx, sess, bucketName, nil)
+	if err != nil {
+		return nil, fmt.Errorf("error opening S3 bucket: %v", err)
+	}
+	return bkt, nil
+}
+
+// ParseStoragePath parses storagePath and returns the storageProvider, bucket and relativePath
```
fejta: Why do we need bucket to be a first class abstraction?

prow/cmd/gerrit/main.go
```diff
@@ -68,8 +80,15 @@ func (o *options) Validate() error {
 		return errors.New("--last-sync-fallback must be set")
 	}
 
-	if strings.HasPrefix(o.lastSyncFallback, "gs://") && o.gcsCredentialsFile == "" {
-		logrus.WithField("last-sync-fallback", o.lastSyncFallback).Warn("--gcs-credentials-file unset, will try and access with a default service account")
+	if o.gcsCredentialsFile != "" {
+		logrus.Warning("-gcs-credentials-file is deprecated and will be removed in September 2020. Use -blob-storage-credentials-file instead.")
+	}
+	if o.gcsCredentialsFile != "" && o.blobStorageCredentialsFile != "" {
+		return fmt.Errorf("please set either -blob-storage-credentials-file or -gcs-credentials-file but not both")
```
fejta: This is not true. We should not require a credentials file to be present on GCP (and prow.k8s.io no longer uses them). Best practice is to use workload identity

pkg/io/v2/opener.go
```diff
@@ -0,0 +1,144 @@
...
+	WriteCloser = io.WriteCloser
+)
+
+// Opener has methods to retrieve a Reader or Writer for the given path
+type Opener interface {
+	Reader(ctx context.Context, path string, opts *blob.ReaderOptions) (io.ReadCloser, error)
+	Writer(ctx context.Context, path string, opts *blob.WriterOptions) (io.WriteCloser, error)
+}
+
+type opener struct {
```
fejta: I would expect to be able to pass both gcs and aws credentials and be able to read from/write to both locations.

prow/cmd/gerrit/main.go
sbueringer: So I guess in case no credentialsFile is specified `storage.NewClient` detects the workload identity and no error is returned from `storage.NewClient` here: https://github.com/kubernetes/test-infra/blob/master/pkg/io/opener.go#L60-L68

pkg/io/v2/opener.go
sbueringer: I could implement is the only question is in which way:

* preserve the original file formats and change the signature to `func NewOpener(gcsCredentialsFile, awsCredentialsFile string) (Opener, error) {` I think this also means we need additional flags for each credentialsfile (even more if we would start supporting Azure,...)
* Only use one file but change the JSON format, e.g.:
```
{
  "s3_credentials": {
    "region": "us-east-1",
    ...
  },
  "gs_credentials": {
    "type": "service_account",
    ...
  }
}
```

pkg/io/v2/providers/providers.go
sbueringer: I guess you mean why I'm not directly opening the file instead of the bucket and therefore if I have to parse the path here this way to find out the bucket name

* I wasn't aware of that but looks like I don't need the bucket name for OpenBucket. I can also just e.g: `blob.OpenBucket(context.TODO(), "gs://k8s-conform-capi-openstack/pr-logs")`
* Currently we only open one blob.Bucket per bucket and cache it. Imho this isn't possible if we open a Bucket for every single file

I tested the following:
```
bkt, err := blob.OpenBucket(context.TODO(), "gs://k8s-conform-capi-openstack/README.md")
attr, err := bkt.Attributes(context.TODO(), "")
content, err := bkt.ReadAll(context.TODO(), "")
reader, err := bkt.NewReader(context.TODO(), "", nil)
``
Only the first line worked, everything after that returned an `object name is empty` error.

The following works:
```
bkt, err := blob.OpenBucket(context.TODO(), "gs://k8s-conform-capi-openstack")
attr, err := bkt.Attributes(context.TODO(), "README.md")
content, err := bkt.ReadAll(context.TODO(), "README.md")
reader, err := bkt.NewReader(context.TODO(), "README.md", nil)
```

So I assume we can exchange bucket with something like path which then is the "directory". But I would also have to remove the caching per-bucket.

File: pkg/io/v2/opener.go (hunk 37 above)
Comment (alvaroaleman) [reply pkg/io/v2/opener.go]: @fejta why? Do we have any code that does this? And wouldn't it be more logical to then construct two openers?

If possible I would like to keep this PR scoped to replicating the functionality in `pkg/io` and not add new features. If we want that, we can still do that in a follow-up.

Comment (fejta) [reply pkg/io/v2/opener.go]: I should be able to have `this` repo write to S3 and `that` one write to GCS. Right now that is not possible because we do not support S3. It will still not be possible after this change because we still only support a single provider.

The point of this abstraction is that we can pass crier, tide, sidecar, etc an interface that allows it to remain ignorant of whether it is talking to the local filesystem, a PVC, or some remote storage system.

I would also like to keep the credentials flags specific to the provider, aka keep `--gcs-credentials` and `--aws-credentials`. GCP credentials are specific to GCP. We should not risk leaking them to AWS (and visa versa).

File: pkg/io/opener.go [hunk 38]
```diff
@@ -81,6 +82,7 @@ func LogClose(c io.Closer) {
 }
 
 func (o opener) openGCS(path string) (*storage.ObjectHandle, error) {
+	path = strings.TrimPrefix(path, "file://")
```
fejta: Seems like we should only do this when its prefix isn't `gs://`, right?

sbueringer: I did it like this because TrimPrefix has the following behaviour:
```
// TrimPrefix returns s without the provided leading prefix string.
// If s doesn't start with prefix, s is returned unchanged.
```
But I guess it would be more obvious if I write it like this:

```
if strings.HasPrefix(path, "file://") {
   path = strings.TrimPrefix(path, "file://")
}
```

fejta: The edge case here is that is `file://gs://foo` should be treated as a local path, but will currently go to GCS

prow/cmd/gerrit/main.go
sbueringer: Removed the flag validation

pkg/io/opener.go
sbueringer: Okay I moved it. So the prefix is only stripped if the gs:// prefix is not detected. But I doubt the `os.Create` function can handle paths like `gs://foo` (It couldn't even handle `f, err := os.Create("file:///tmp/test")`

pkg/io/v2/opener.go
sbueringer: Okay so this variant would mean we have to add a flag for every provider we want to support via explicit configuration via a config file (auto-discovery could still be supported for all providers via gocloud). Then I also should remove the blob-storage-credentials-file flag and use the other flag only to decide if the old or new package is used. Obviously I would remove the deprecation of the gcs-credentials-file flag.

sbueringer: Okay. So how do we get consensus on this issue? :) For me it's okay either way

mikesplain:
@sbueringer looks like you have conflicts and may need to rebase.

pkg/io/v2/opener.go
alvaroaleman: Yeah, Erick has a point, someone might want to multiple providers

sbueringer:
> @sbueringer looks like you have conflicts and may need to rebase.

Thx. Should be fixed now. I also squashed all review fix commits from the past. Rebasing would have been a nightmare otherwise

pkg/io/v2/opener.go
sbueringer: Should be fixed now

sbueringer:
@fejta @alvaroaleman All findings should be addressed now. Please take another look. Thx :)

alvaroaleman:
@sbueringer Another merge conflict, sorry for that :( Other than that, lgtm and thanks again for your efforts of pushing things forward here

sbueringer:
> @sbueringer Another merge conflict, sorry for that :( Other than that, lgtm and thanks again for your efforts of pushing things forward here

Rebased. No problem :)

pkg/io/opener.go
```diff
@@ -98,24 +99,30 @@ func (o opener) openGCS(path string) (*storage.ObjectHandle, error) {
...
+		}
 		return os.Open(path)
 	}
 	return g.NewReader(ctx)
 }
 
 // Writer returns a writer that overwrites the path.
-func (o opener) Writer(ctx context.Context, path string) (io.WriteCloser, error) {
+func (o opener) Writer(ctx context.Context, path string, _ *blob.WriterOptions) (io.WriteCloser, error) {
 	g, err := o.openGCS(path)
```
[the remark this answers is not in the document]
alvaroaleman: > I'm not convinced we need/want io2

How do you want to implement the featuregate without adding a new package?

> Instead of replacing this whole package with io2, why don't we just continue extending this package to support S3? AKA:

People also asked about Azure and may ask about more providers in the future. What issue do you see with the abstraction used here?

> We can move the GCS related stuff into a gcs.go file and the s3 stuff into an s3.go or aws.go... Eventually these two files should go into their own package, and importing them can have an inti() which registers a particular schema like gs://.

Why should they go into their own package? That would mean that everything that uses storage needs to be aware of the underlying technology and that is IMHO exactly what we want to avoid and why we started to use the gocloud abstraction in the first place.

prow/statusreconciler/status.go
```diff
@@ -103,12 +109,12 @@ func (s *statusController) loadState() (storedState, error) {
 	entry := s.logger.WithField("path", s.statusURI)
 	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
 	defer cancel()
-	reader, err := s.opener.Reader(ctx, s.statusURI)
```
fejta: The primary thing I want is to keep the interface like this, aka we shoudn't be passing bucket options to an abstract Reader interface.

pkg/io/opener.go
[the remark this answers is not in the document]
fejta: > How do you want to implement the featuregate without adding a new package?

If we continue using the existing code path for `gs://` then we do not need a new package or featuregate (as the behavior isn't changing)

This PR should not need to make changes to tide aside from adding the flag for S3 and using it when constructing the opener. Instead we're changing all the calls to `Reader()` and `Writer()`

> What issue do you see with the abstraction used here?

Consumers of this interface should not have to import gcloud.dev/blob in order to use the Opener.

> Why should they go into their own package?

Not totally necessary, but it would provide an easy way for people who build their own images to reduce dependencies by changing what they import.

AKA if I wanted my prow distribution to not support honking, I can provide my own plugin-import package that drops this line: https://github.com/kubernetes/test-infra/blob/master/prow/hook/plugin-imports/plugin-imports.go#L36

I do not need/want to do that now, but we should keep it in mind.

pkg/io/v2/opener.go
```diff
@@ -0,0 +1,153 @@
...
+	return &opener{
+		gcsCredentials: gcsCredentials,
+		s3Credentials:  s3Credentials,
+		cachedBuckets:  map[string]*blob.Bucket{},
+	}, nil
+}
+
+// getBucket opens a bucket
+// The storageProvider is discovered based on the given path.
+// The buckets are cached per bucket name. So we don't open a bucket multiple times in the same process
```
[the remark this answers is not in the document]
sbueringer: I had a previous implementation which returned a readerCloser which contained the bucket and the reader/writer. The close method both closed the reader and the writer. To simplify this we decided to just return the reader and writer and cache the bucket so it doesn't have to be closed.

I see multiple options:
* return a readerCloser / writerCloser again which encapsulates the bucket and the reader/writer and closes both: See: https://github.com/kubernetes/test-infra/commit/f75838e0ee47004e4faedc0099e16f93871aafea#diff-12f70e7ec70b1a60275ef5d6dab027beR101
* keep it as is
* only return reader/ writer and ignore the bucket.Close(). (That might be an option because all implemented providers in gophercloud currently have a Close() implementation like this:
   ```
      func (b *bucket) Close() error {
	return nil
      }
   ```

pkg/io/opener.go
```diff
@@ -43,8 +44,8 @@ type (
 
 // Opener has methods to read and write paths
 type Opener interface {
-	Reader(ctx context.Context, path string) (ReadCloser, error)
-	Writer(ctx context.Context, path string) (WriteCloser, error)
+	Reader(ctx context.Context, path string, opts *blob.ReaderOptions) (ReadCloser, error)
```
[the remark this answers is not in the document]
sbueringer: I'm not sure how we are leaking buckets here. The path can be any valid storage path you want.

Regarding the *blob.ReadOptions etc. I can remove it in this PR and we can look for a solution in later PRs. In this PR it's not strictly required. We just sooner or later have the issue with at least *blob.Attributes. Maybe we can avoid it on ReaderOptions and WriterOptions.

And then I only see the options to either expose the blob.Attributes or wrap them.

prow/statusreconciler/status.go
sbueringer: Done > https://github.com/kubernetes/test-infra/pull/16317/commits/b767f4531f54a8e65753d4da3b0a38f1219cd4f9

pkg/io/opener.go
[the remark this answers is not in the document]
sbueringer: Removed the Reader/WriterOptions > https://github.com/kubernetes/test-infra/pull/16317/commits/b767f4531f54a8e65753d4da3b0a38f1219cd4f9

prow/tide/history/history.go
```diff
@@ -45,19 +50,19 @@ type History struct {
...
 
-	opener io.Opener
+	opener Opener
 	path   string
 }
 
-func readHistory(maxRecordsPerKey int, opener io.Opener, path string) (map[string]*recordLog, error) {
+func readHistory(maxRecordsPerKey int, opener Opener, path string) (map[string]*recordLog, error) {
 	reader, err := opener.Reader(context.Background(), path)
-	if io.IsNotExist(err) { // No history exists yet. This is not an error.
```
fejta: Can we update `io` handle this scenario and continue using this package?

pkg/io/opener.go
```diff
@@ -94,29 +131,74 @@ func (o opener) openGCS(path string) (*storage.ObjectHandle, error) {
...
+		}
+	}
+
+	bucket, relativePath, err := o.getBucket(ctx, path)
 	if err != nil {
-		return nil, fmt.Errorf("bad gcs path: %v", err)
+		return nil, err
 	}
-	if g == nil {
-		return os.Open(path)
```
fejta: We should either continue using this for paths that do not specify a scheme or else put it behind a flag. I'm not a fan of forcing `/local/path` to become `file:///local/path` so ideally we just preserve this behavior.

prow/cmd/tide/main.go
```diff
@@ -55,17 +55,27 @@ type options struct {
...
 	gcsCredentialsFile string
+	// 	s3CredentialsFile string is used for reading/writing to s3 block storage.
+	// If you want to write to "file://" paths, this parameter is optional.
+	// If set, this file is used to read/write to s3:// paths
+	// If not, go cloud credential auto-discovery is used
+	// For more details see the pkg/io/v2/providers pkg.
+	s3CredentialsFile string
 	// historyURI where Tide should store its action history.
-	// Can be a /local/path or gs://path/to/object.
+	// Can be file:///local/path, gs://path/to/object or s3://path/to/object.
```
fejta: I would like to continue supporting `/local/path`

pkg/io/opener.go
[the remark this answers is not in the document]
fejta: Makes sense, but can we start with the existing interface for now?

prow/tide/history/history.go
sbueringer: Sorry there was another commit. I think this comment is outdated. After the latest commit there are no changes to history.go anymore

pkg/io/opener.go
sbueringer: Actually they still are supported. On of the first things I'm doing is the following:
```
        if strings.HasPrefix(storagePath, "/") {
		storagePath = fmt.Sprintf("file://%s", storagePath)
	}
```
Good enough? (I would update the doc that both is valid)

prow/cmd/status-reconciler/main.go
```diff
@@ -56,10 +57,22 @@ type options struct {
...
+	// If you want to write to "file://" paths, this parameter is optional.
+	// If set, this file is used to read/write to s3:// paths
+	// If not, go cloud credential auto-discovery is used
+	// For more details see the pkg/io/v2/providers pkg.
+	s3CredentialsFile string
+	// useGenericBlobStorage enables the usage of the new generic blob storage package
+	useGenericBlobStorage bool
 	// statusURI where Status-reconciler stores last known state, i.e. configuration.
-	// Can be a /local/path or gs://path/to/object.
+	// Can be file:///local/path, gs://path/to/object or s3://path/to/object.
```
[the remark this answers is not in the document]
sbueringer: Actually they still are supported. On of the first things I'm doing is the following:
```
        if strings.HasPrefix(storagePath, "/") {
		storagePath = fmt.Sprintf("file://%s", storagePath)
	}
```
Good enough? (I would update the doc that both is valid)

prow/cmd/tide/main.go
sbueringer: see the other conversations

pkg/io/opener.go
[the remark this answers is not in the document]
sbueringer: Sorry should have pushed both at once

pkg/io/opener.go
fejta: Ah I missed that, seems good enough for now (although doesn't work well for relative paths, right?) We can patch it if it becomes an issue.

pkg/io/opener.go
```diff
@@ -94,29 +131,74 @@ func (o opener) openGCS(path string) (*storage.ObjectHandle, error) {
...
+	}
+	o.cachedBuckets[bucketName] = bucket
+	return bucket, relativePath, nil
 }
 
 // Reader will open the path for reading, returning an IsNotExist() error when missing
-func (o opener) Reader(ctx context.Context, path string) (io.ReadCloser, error) {
-	g, err := o.openGCS(path)
+func (o *opener) Reader(ctx context.Context, path string) (io.ReadCloser, error) {
+	if strings.HasPrefix(path, "gcs://") || strings.HasPrefix(path, "gs://") {
```
fejta: Please drop support for `gcs://`, the only commonly recognized scheme is `gs://`

```console
$ gsutil ls gcs://
InvalidUrlError: Unrecognized scheme "gcs".
```

prow/cmd/gerrit/main.go
```diff
@@ -41,14 +41,27 @@ import (
...
+	// If set, this file is used to read/write to s3:// paths
+	// If not, go cloud credential auto-discovery is used
+	// For more details see the pkg/io/v2/providers pkg.
+	s3CredentialsFile string
+	cookiefilePath    string
+	configPath        string
+	jobConfigPath     string
+	projects          client.ProjectsFlag
+	// lastSyncFallback is the path to sync the latest timestamp
+	// Can be file:///local/path, gs://path/to/object or s3://path/to/object.
```
fejta: Can we change the docs back to `/local/path`?

pkg/io/providers/providers.go
```diff
@@ -0,0 +1,150 @@
+/*
+Copyright 2018 The Kubernetes Authors.
```
fejta: 2020?

prow/cmd/gerrit/main.go
sbueringer: Done

pkg/io/opener.go
sbueringer: Done

sbueringer:
@fejta Squashed :)

PR #16318, opened 2020-02-15
Prow/project plugin supports project name with spaces

chaodaiG:
Part of: https://github.com/knative/test-infra/issues/1265


mattmoor:
Thanks for fixing this!

prow/plugins/project/project.go
```diff
@@ -114,22 +114,42 @@ func updateProjectNameToIDMap(projects []github.Project) {
 // proposed column name, whether the command is to remove issue/PR from project,
 // and the error message
 func processRegexMatches(matches []string) (string, string, bool, string) {
```
taragu: We can send in just `matches[1]` since the project and column names will be grouped together now

prow/plugins/project/project.go
```diff
@@ -114,22 +114,42 @@ func updateProjectNameToIDMap(projects []github.Project) {
...
+	proposedProject := ""
 	proposedColumnName := ""
-	if len(matches) > 1 && proposedProject != clearKeyword {
-		proposedColumnName = matches[2]
+
+	var shouldClear = false
+	content := strings.TrimSpace(matches[1])
+
+	// Take care of clear
+	if strings.HasPrefix(content, clearKeyword) {
```
taragu: Can we have the prefix as `clear ` instead (also same for replace)? The `parts = strings.SplitN(content, " ", 2)` down below might be problematic because of this. For a command like `/project clear project name`, the `parts` will be 
`part[0]`: `(empty string)`
`part[1]`: `project name`

prow/plugins/project/project_test.go
```diff
@@ -394,6 +394,55 @@ func TestParseCommand(t *testing.T) {
...
+		},
+		{
+			hasMatches:      true,
+			command:         "/project 'something 0.0.0' \"To do\"",
+			proposedProject: "something 0.0.0",
+			proposedColumn:  "To do",
+			shouldClear:     false,
+		},
+		{
+			hasMatches:      true,
```
taragu: Can we add a test case for `/project clear project name`?

chaodaiG: good point, added

prow/plugins/project/project.go
chaodaiG: I have tried that in the first place `clear ` failed a few test cases, I think it's supporting `/project clear` as a valid command. The problem you mentioned did happen, trimming spaces at both ends solved the problem, it's tested by the test case you suggested below

prow/plugins/project/project.go
chaodaiG: done

prow/plugins/project/project.go
taragu: Ah I see. Sounds good!

prow/plugins/project/project.go
```diff
@@ -113,23 +113,43 @@ func updateProjectNameToIDMap(projects []github.Project) {
 // processRegexMatches processes the user command regex matches and returns the proposed project name,
 // proposed column name, whether the command is to remove issue/PR from project,
 // and the error message
-func processRegexMatches(matches []string) (string, string, bool, string) {
-	var shouldClear = false
-	proposedProject := matches[1]
+func processRegexMatches(match string) (string, string, bool, string) {
```
taragu: We can now rename this function to something like `processCommand(...)`

chaodaiG: done

taragu:
Thanks for the fix! 😄
/lgtm
/assign @fejta

PR #16321, opened 2020-02-17
kubetest2/kind: able to specify kind configuration, etc

cofyc:
add --config, --kubeconfig and --verbosity support for kubetest2 kind deployer

### DIFF
--- a/kubetest2/kubetest2-kind/deployer/deployer.go
+++ b/kubetest2/kubetest2-kind/deployer/deployer.go
@@ -20,6 +20,7 @@ package deployer
 import (
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


cofyc:
/assign @amwat

what do you think?

kubetest2/kubetest2-kind/deployer/deployer.go
```diff
@@ -74,6 +78,15 @@ func bindFlags(d *deployer) *pflag.FlagSet {
 	flags.StringVar(
 		&d.nodeImage, "build-type", "", "--type for kind build node-image",
```
BenTheElder: note that this may go away upstream depending on how K/K handles changes to their build we will focus on whatever is going to be supported properly.

PR #16322, opened 2020-02-17
kubetest2/gke: able to specify gke endpoint via --environment flag

cofyc:


kubetest2/kubetest2-gke/deployer/deployer.go
```diff
@@ -135,7 +135,7 @@ func (d *deployer) location() (string, error) {
 	if d.zone != "" {
 		return "--zone=" + d.zone, nil
 	}
-	return "--region" + d.region, nil
+	return "--region=" + d.region, nil
```
cofyc: a minor bug

kubetest2/kubetest2-gke/deployer/commandutils.go
```diff
@@ -29,6 +29,22 @@ func (d *deployer) prepareGcpIfNeeded() error {
...
+		endpoint = "https://staging-container.sandbox.googleapis.com/"
+	case env == "staging2":
+		endpoint = "https://staging2-container.sandbox.googleapis.com/"
+	case env == "prod":
+		endpoint = "https://container.googleapis.com/"
+	case urlRe.MatchString(env):
+		endpoint = env
+	default:
+		return fmt.Errorf("--gke-environment must be one of {test,staging,prod} or match %v, found %q", urlRe, env)
+	}
```
cofyc: copied from kubetest/gke.go

https://github.com/kubernetes/test-infra/blob/f5dd3ee0e9db3b4c67073815b399aa66a1620d2e/kubetest/gke.go#L192-L206

cofyc: my personal GCP account can only access production endpoint, this is required outside of the prow platform

cofyc:
/assign @amwat

what do you think?

kubetest2/kubetest2-gke/deployer/commandutils.go
```diff
@@ -29,6 +29,22 @@ func (d *deployer) prepareGcpIfNeeded() error {
...
+	case env == "staging":
+		endpoint = "https://staging-container.sandbox.googleapis.com/"
+	case env == "staging2":
+		endpoint = "https://staging2-container.sandbox.googleapis.com/"
+	case env == "prod":
+		endpoint = "https://container.googleapis.com/"
+	case urlRe.MatchString(env):
+		endpoint = env
+	default:
+		return fmt.Errorf("--gke-environment must be one of {test,staging,prod} or match %v, found %q", urlRe, env)
```
amwat: I think our flag here is just `--environment` ?

kubetest2/kubetest2-gke/deployer/commandutils.go
```diff
@@ -29,6 +29,22 @@ func (d *deployer) prepareGcpIfNeeded() error {
 	// TODO(RonWeber): This is an almost direct copy/paste from kubetest's prepareGcp()
 	// It badly needs refactored.
 
+	var endpoint string
+	switch env := d.environment; {
```
amwat: nit: use env as the switch condition so that all cases are simplified
```
switch env := d.environment; env {
case "test": ...
}
```

kubetest2/kubetest2-gke/deployer/commandutils.go
cofyc: fixed

kubetest2/kubetest2-gke/deployer/commandutils.go
cofyc: can't switch on `env` string because the last case is matching on a regex `urlRe.MatchString(env)` and the result is a bool.

amwat: ahh right

Issue #16120, opened 2020-02-04
Pipeline (Tekton) extension for Prow is not working - pipeline pod is dying


ruzickap:
**What happened**:

Image specified in the k8s manifest `gcr.io/k8s-testimages/pipeline:latest` for pipeline doesn't exists: https://github.com/kubernetes/test-infra/blob/master/prow/cmd/pipeline/dev.yaml#L18

If I change it the image to `gcr.io/k8s-prow/pipeline` like:

```yaml
...
    spec:
      serviceAccount: prow-pipeline
      containers:
      - name: pipeline
        image: gcr.io/k8s-prow/pipeline:latest
---
```

then the pod is dying with this error:

```text
$ kubectl logs prow-pipeline-6cf5d55b4d-74rhs -n prow             
{"component":"pipeline","file":"prow/kube/config.go:147","func":"k8s.io/test-infra/prow/kube.LoadClusterConfigs","level":"info","msg":"Loading cluster contexts...","time":"2020-02-04T13:03:44Z"}
{"component":"pipeline","file":"prow/cmd/pipeline/main.go:134","func":"main.main","level":"warning","msg":"Truncating to default context","time":"2020-02-04T13:03:44Z"}
{"component":"pipeline","file":"prow/cmd/pipeline/controller.go:148","func":"main.newController","level":"info","msg":"Setting up event handlers","time":"2020-02-04T13:03:44Z"}
{"component":"pipeline","file":"prow/cmd/pipeline/controller.go:202","func":"main.(*controller).Run","level":"info","msg":"Starting Pipeline controller","time":"2020-02-04T13:03:44Z"}
{"component":"pipeline","file":"prow/cmd/pipeline/controller.go:203","func":"main.(*controller).Run","level":"info","msg":"Waiting for informer caches to sync","time":"2020-02-04T13:03:44Z"}
{"component":"pipeline","file":"prow/cmd/pipeline/controller.go:106","func":"main.(*controller).hasSynced","level":"info","msg":"Synced prow jobs","time":"2020-02-04T13:03:44Z"}
{"component":"pipeline","file":"prow/cmd/pipeline/controller.go:120","func":"main.(*controller).hasSynced","level":"info","msg":"Synced default pipelines","time":"2020-02-04T13:03:44Z"}
{"component":"pipeline","file":"prow/cmd/pipeline/controller.go:208","func":"main.(*controller).Run","level":"info","msg":"Starting workers","time":"2020-02-04T13:03:44Z"}
[...]

PR #16323, opened 2020-02-17
[Pipeline] Fix nil pointer when accessing config

eddycharly:
This PR fixes https://github.com/kubernetes/test-infra/issues/16120

When the prow config is not set on the command line, the config returned by the config agent is `nil`.
This must be verified before accessing properties and use fallback/default values when it is `nil`.

### DIFF
--- a/prow/cmd/pipeline/main.go
+++ b/prow/cmd/pipeline/main.go
@@ -47,7 +47,7 @@ import (
 type options struct {
 	allContexts  bool
 	buildCluster string
-	config       string
+	configPath   string
 	kubeconfig   string
 	totURL       string
 }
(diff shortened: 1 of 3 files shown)


matthyx:
I am not sure I like the current implementation... isn't there a better way to deal with default configs?

eddycharly:
One option would be to make the command line argument mandatory.
Having the agent manage a default config would be great but it would require more changes.

What do you prefer ?

matthyx:
I would make it mandatory.
/cc @cjwagner

stevekuznetsov:
Is it valid to run Prow without a config?

eddycharly:
I just found the config.ConfigPath function, I think it should do the trick.

eddycharly:
In the end i modified `pipeline/dev.yaml` to mount the prow config and added a call to `config.ConfigPath` to support default config path.

I reverted the changes where i check for `nil` when getting the config.

matthyx:
Looks 100x better, thanks!
/lgtm

cblecker:
This looks okay to me, but I'd like to get some more eyes on it.

/cc @stevekuznetsov @cjwagner @fejta 
/hold

prow/cmd/pipeline/main.go
```diff
@@ -64,18 +64,19 @@ func (o *options) parse(flags *flag.FlagSet, args []string) error {
 	flags.BoolVar(&o.allContexts, "all-contexts", false, "Monitor all cluster contexts, not just default")
 	flags.StringVar(&o.totURL, "tot-url", "", "Tot URL")
 	flags.StringVar(&o.kubeconfig, "kubeconfig", "", "Path to kubeconfig. Only required if out of cluster")
-	flags.StringVar(&o.config, "config", "", "Path to prow config.yaml")
+	flags.StringVar(&o.configPath, "config", "", "Path to prow config.yaml")
```
stevekuznetsov: We should add a validation that this is set

eddycharly: I didn’t add validation here because I call config.ConfigPath later that sets the default config path if it was not set on the command line.

It is what is done in other components (hook, deck, etc...) and I thought it would be more consistent to do the same thing here. 

I can add validation if you prefer though.

cjwagner: It is bad that other components default this field. We only continue to do that for backwards compatibility and it looks like that should have been deprecated by now so please don't add more components that use it: https://github.com/kubernetes/test-infra/blob/7ac63bdbab0f3666497bede973d7703bb30e9cd4/prow/config/config.go#L1324
Validation is preferable.

matthyx: Sorry, didn't know that... I should probably follow more closely sig meetings where we communicate this -_-"

eddycharly: Thank you for your explanation, i removed the call to `config.ConfigPath` and added validation.

prow/cmd/pipeline/main_test.go
```diff
@@ -29,18 +29,25 @@ func TestOptions(t *testing.T) {
 		expected *options
 		err      bool
 	}{{
-		name:     "defaults work",
+		name:     "defaults doesn't work (--config argument is mandatory)",
```
fejta: Please fix this so it provides the minimal amount necessary to work (aka add --config)

eddycharly: I enhanced the validation message and the test name with a more explicit message about what should be done to make it work.

PR #16335, opened 2020-02-18
Add experimental gce-100-perf job with containerd enabled

mm4tt:


config/jobs/kubernetes/sig-scalability/sig-scalability-experimental-periodic-jobs.yaml
```diff
@@ -298,3 +298,65 @@ periodics:
...
+          - --repo=k8s.io/kubernetes=master
+          - --repo=k8s.io/perf-tests=master
+          - --root=/go/src
+          - --scenario=kubernetes_e2e
+          - --
+          - --check-leaked-resources
+          # TODO(oxddr): remove once debugging is finished
+          - --env=KUBEPROXY_TEST_LOG_LEVEL=--v=4
+          - --cluster=e2e-big-containerd
+          - --env=KUBE_CONTAINER_RUNTIME=containerd
```
wojtek-t: Also add: 
`--env=LOG_DUMP_SYSTEMD_SERVICES=container`

mm4tt: Done

PR #16337, opened 2020-02-18
Boskos handler tests: Verify the wire format doesn't get changed

alvaroaleman:
This PR extends the existing boskos handler tests to check their payloads against fixtures, to make sure we don't actidentally change the wire format down the line.

/assign @petr-muller @stevekuznetsov 

This should be reviewed with whitespace changes disabled


boskos/handlers/server_client_test.go
```diff
@@ -33,11 +36,42 @@ import (
...
 )
 
-func makeTestBoskos(r *ranch.Ranch) *httptest.Server {
-	handler := NewBoskosHandler(r)
+func makeTestBoskos(t *testing.T, r *ranch.Ranch) *httptest.Server {
+	handler := &testMuxWrapper{t: t, ServeMux: NewBoskosHandler(r)}
 	return httptest.NewServer(handler)
 }
 
+type testMuxWrapper struct {
```
alvaroaleman: All of this feels pretty clunky, but the only thing I could find in the stdlib is the httptest.Recorder and that needs to be plugged into each handler. I'd appreciate any pointers if this can be simplified.

boskos/handlers/handlers_test.go
```diff
@@ -651,55 +672,40 @@ func TestUpdate(t *testing.T) {
 			code:   http.StatusOK,
 			method: http.MethodPost,
 		},
-		{
```
stevekuznetsov: What happened to this test?

alvaroaleman: It's a duplicate of the test above

PR #16357, opened 2020-02-19
only create non-existent device nodes

BenTheElder:
we already fixed the bad entries, we don't need to keep using rm


BenTheElder:
cc @pohly

PR #16362, opened 2020-02-19
Prow/metrics: Fix registry defaulting

alvaroaleman:
This caused a NPD, because the nil `*prometheus.Registry` passed into `serveMetrics` that accepts a `prometheus.Gatherer` caused its check if the gatherer is nil to always return false, as an interface is only nil if both its type and value are unset.

### DIFF
--- a/prow/metrics/BUILD.bazel
+++ b/prow/metrics/BUILD.bazel
@@ -41,7 +41,13 @@ filegroup(
 
 go_test(
     name = "go_default_test",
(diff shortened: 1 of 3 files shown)


fejta:
Can we revert this and/or make it more robust (stop requiring a specific port)

alvaroaleman:
The codechange needs to stay, otherwise everything that serves metrics will crash with a npd

PR #16365, opened 2020-02-19
Use util.Repo in prow/tide and prow/config

matthyx:
Relates to #16223
I had to move the struct to its own package to avoid circular dependency between prow/plugins and prow/config.
/assign @alvaroaleman


matthyx:
> Can we maybe move it into the `prow/config` package?

Done.

PR #14192, opened 2019-09-04
bugzilla: generalize statuses to bug states

petr-muller:
The Bugzilla plugin has four fields to configure "statuses": either
they are statuses that the BZ expects referenced bugs to have, or they
specify the status to which the bugs would be moved after some action:

```
  ...
  statuses:
  - VALIDATED
  dependent_bug_statuses:
  - VALIDATED
  status_after_validation: "PRE"
  status_after_merge: MODIFIED`
```

This ability is insuficient as soon as "resolutions" come to play:
resolutions are sub-states of BZ closed statuses. Therefore, this
commit generalizes a status into a "state", which is used for both
purposes (bugs being checked against and updated to).

```
  ...
  valid_states:
  - status: VALIDATED
  - status: CLOSED
    resolution: ERRATA
  dependent_bug_states:
  - status: VALIDATED
  state_after_validation:
    status: "PRE"
  state_after_merge:
    status: MODIFIED`
```

The old "status" config fields are kept to avoid breaking existing
instances and converted to states at runtime.

/cc @stevekuznetsov

PR #16373, opened 2020-02-19
prow/plugins/bugzilla: Document DependentBug* in plugin help

wking:
To make the current restrictions more discoverable, since plugin response comments on particular pull requests can go stale after config changes.

Also flesh out the comments on the config settings to clarify the "all dependents must satisfy" behavior and hint at mergeStatusesIntoStates. More on status vs. state in e5ddec2974 (#14192).

CC @petr-muller, @stevekuznetsov


wking:
CC @bbguimaraes to, to fill out the Bugzilla plugin approvers ;)

wking:
7da244a161 -> ece1eca401 fixes the unit tests, and we're now green here.

/hold cancel

But needs a fresh `lgtm`.

PR #16383, opened 2020-02-20
Remove fetcher, transform, and sql from velodrome

apelisse:
I've removed fetcher (that used to fetch metrics from github), sql (which was managing the database where we stored the github data) and transform (which was transforming  the data from mysql to influxdb).

I've also removed istio, beam and tensorflow data gathering and dashboard since I don't think they've really ever been used.

I think the prober as well as the influx database are still used (mostly for bigquery metrics?).

I'm trying to clean-up what's definitely not used and also generate some discussion about how to proceed from here.


apelisse:
I've also removed the token-counter since I couldn't find any graph with the data anywhere, does anyone know if it's still used?

apelisse:
Based on @Katharine's comment, I'm assuming we can close that.

PR #16384, opened 2020-02-20
fix kind download permissions

BenTheElder:

something changed with `gsutil -P` kind. I suspect something with UID / GID of the file or the image we're running in. not sure why, but we don't want it preserving those bits anyhow, the goal was just to preserve the executable bit.

will write something cleaner in a future patch

/cc @aojea

### DIFF
--- a/config/jobs/kubernetes/sig-testing/kubernetes-kind-presubmits.yaml
+++ b/config/jobs/kubernetes/sig-testing/kubernetes-kind-presubmits.yaml
@@ -240,7 +240,7 @@ presubmits:
         - wrapper.sh
         - bash
         - -c
-        - gsutil cp -P gs://bentheelder-kind-ci-builds/latest/kind-linux-amd64 "${PATH%%:*}/kind" && gsutil cat gs://bentheelder-kind-ci-builds/latest/e2e-k8s.sh | sh
+        - gsutil cp gs://bentheelder-kind-ci-builds/latest/kind-linux-amd64 "${PATH%%:*}/kind" && chmod +x $(which kind) && gsutil cat gs://bentheelder-kind-ci-builds/latest/e2e-k8s.sh | sh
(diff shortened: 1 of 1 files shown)


BenTheElder:
ref https://github.com/kubernetes-sigs/kind/issues/1342

config/jobs/kubernetes/sig-testing/kubernetes-kind-presubmits.yaml
```diff
@@ -240,7 +240,7 @@ presubmits:
         - wrapper.sh
         - bash
         - -c
-        - gsutil cp -P gs://bentheelder-kind-ci-builds/latest/kind-linux-amd64 "${PATH%%:*}/kind" && gsutil cat gs://bentheelder-kind-ci-builds/latest/e2e-k8s.sh | sh
+        - gsutil cp gs://bentheelder-kind-ci-builds/latest/kind-linux-amd64 "${PATH%%:*}/kind" && chmod +x $(which kind) && gsutil cat gs://bentheelder-kind-ci-builds/latest/e2e-k8s.sh | sh
```
aojea: is `which` the right thing or should we use the full path from the previous command?

BenTheElder:
this may not be the actual fix, but -P appears to not do what we want regardless. looking into another possible issue

config/jobs/kubernetes/sig-testing/kubernetes-kind-presubmits.yaml
BenTheElder: the previous command ensures it is. also this is in a fixed CI image environment. the path is not variable.

BenTheElder: the previous command specifically places it into the first PATH element

PR #16385, opened 2020-02-20
use curl to avoid gsutil failing to auth

BenTheElder:
i don't know why it's even authenticating when the bucket is public read, but the timing of the error appearing lines up to workload identity.

also the error is about the bucket already existing rather than auth, but this is my hunch.

using curl instead of gsutil should avoid the problem

fyi @amwat 
/cc @aojea 
xref: https://github.com/kubernetes-sigs/kind/issues/1342

### DIFF
--- a/config/jobs/kubernetes/sig-testing/kubernetes-kind-presubmits.yaml
+++ b/config/jobs/kubernetes/sig-testing/kubernetes-kind-presubmits.yaml
@@ -240,7 +240,7 @@ presubmits:
         - wrapper.sh
         - bash
         - -c
-        - gsutil cp gs://bentheelder-kind-ci-builds/latest/kind-linux-amd64 "${PATH%%:*}/kind" && chmod +x $(which kind) && gsutil cat gs://bentheelder-kind-ci-builds/latest/e2e-k8s.sh | sh
+        - curl -sSLo "${PATH%%:*}/kind" https://storage.googleapis.com/bentheelder-kind-ci-builds/latest/kind-linux-amd64 && chmod +x $(which kind) && curl -sSL https://storage.googleapis.com/bentheelder-kind-ci-builds/latest/e2e-k8s.sh | sh
         env:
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes/sig-testing/kubernetes-kind-presubmits.yaml
```diff
@@ -240,7 +240,7 @@ presubmits:
         - wrapper.sh
         - bash
         - -c
-        - gsutil cp gs://bentheelder-kind-ci-builds/latest/kind-linux-amd64 "${PATH%%:*}/kind" && chmod +x $(which kind) && gsutil cat gs://bentheelder-kind-ci-builds/latest/e2e-k8s.sh | sh
+        - curl -sSLo "${PATH%%:*}/kind" https://storage.googleapis.com/bentheelder-kind-ci-builds/latest/kind-linux-amd64 && chmod +x $(which kind) && gsutil cat gs://bentheelder-kind-ci-builds/latest/e2e-k8s.sh | sh
```
aojea: and the `gsutil cat`?

BenTheElder: good point

BenTheElder: done

BenTheElder:
/hold
the other option is an explicit service account mount

BenTheElder:
https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kind-build/1230395125159432192 works fine, with `    preset-service-account: "true"`

BenTheElder:
/hold cancel
this is simpler anyhow / more portable to local usage. we don't need auth.

I'm sketching out a more elegant pattern

BenTheElder:
TBR, this is blocking all k/k PRs

PR #16392, opened 2020-02-20
[Deck] Add material theme config

eddycharly:
This pull request adds a branding config to customize material design theme.

If not defined the default indigo-pink theme is used, but the user can change it in the deck branding config.

It's definitely useless but i thought it would be fun to be able to change it ;-)


eddycharly:
@Katharine if you have a few minutes to review this PR it would be nice. It shouldn’t be too long ;-)

eddycharly:
@Katharine Did you give a look to this PR ?

Katharine:
The implementation of this feature is fine.

However, I'm not sure we should add it. I think we make assumptions about the colours elsewhere that would be violated by these changes. Additionally, it may become difficult to support this setting in the future (if e.g. we change how we reference this library or stop using it at all).

Given that, I'm inclined not to add this feature right now.

eddycharly:
Thanks for your feedback.

I see your point, and given that it doesn't bring anything useful i will close the PR.

PR #16393, opened 2020-02-20
Boskos cleaner: Allow enabling debug logging

alvaroaleman:


boskos/cmd/cleaner/main.go
```diff
@@ -55,6 +56,7 @@ func init() {
 	flag.StringVar(&passwordFile, "password-file", "", "The path to password file used to access the Boskos server")
 	flag.IntVar(&cleanerCount, "cleaner-count", defaultCleanerCount, "Number of threads running cleanup")
 	flag.StringVar(&namespace, "namespace", corev1.NamespaceDefault, "namespace to install on")
+	flag.BoolVar(&enableDebugLogging, "enable-debug-logging", false, "Whether to enable debug logging")
```
ixdy: maybe we should expose the log level as a flag instead? it seems to be what we've done with boskos and the janitor:
https://github.com/kubernetes/test-infra/blob/4a4c4aa0579418acdc10f9186c25d9fb71de0704/boskos/cmd/boskos/boskos.go#L56

(We could even add it to a list of common flags for all boskos components? but I'm not sure whether that's a good practice.)

alvaroaleman: Other components get that from the config file. I initially wanted to do the string level flag, but the log level in logrus is an iota and I was too lazy to build a mapping for that

ixdy: I think logrus has a helper for that?

https://github.com/kubernetes/test-infra/blob/4a4c4aa0579418acdc10f9186c25d9fb71de0704/boskos/cmd/boskos/boskos.go#L75-L79

alvaroaleman: Good point, missed that. Updated the PR now.

ixdy:
looks like you need to update-bazel, though.

PR #16397, opened 2020-02-20
Kops - Use latest-experimental tag for presubmit jobs

hakman:
Since workload identity was enabled on the e2e clusters, presubmit jobs have been failing because the computeMetadata endpoints are no longer available.
https://github.com/kubernetes/test-infra/pull/16388 tries to address this and moving to `latest-experimental` tag should unblock things.


rifelpet:
once the compute metadata issues have stabilized and new kubekins-e2e images have been built and pushed we will revert back.

PR #16399, opened 2020-02-20
Add storage for alertmanager to allow persisting silences.

cjwagner:
Adding storage for the alert manager will allow us to silence alerts and have the silences persist across pod restarts (which happen whenever changes are merged to `/prow/cluster/`).

@hongkailiu noted that this section of the Alertmanager resource is immutable so our deployment automation will fail to apply the changes. We'll need to delete the alertmanager resource manually and recreate it (or let the deployment automation recreate it). We'll also need to do this again if the storage section ever changes. I don't know of a way to avoid this, but if anyone else does please share.

/hold
/assign @michelle192837 @fejta 
/cc @hongkailiu

### DIFF
--- a/prow/cluster/monitoring/prow_alertmanager.yaml
+++ b/prow/cluster/monitoring/prow_alertmanager.yaml
@@ -14,3 +14,14 @@ spec:
     runAsUser: 1000
   serviceAccountName: alertmanager
   version: v0.15.2
+  storage: # Note that this section is immutable so changes require deleting and recreating the resource.
+    volumeClaimTemplate:
+      metadata:
+        name: prometheus
+      spec:
+        accessModes:
+        - "ReadWriteOnce"
+        storageClassName: "standard"
+        resources:
+          requests:
+            storage: 10Gi


prow/cluster/monitoring/prow_alertmanager.yaml
```diff
@@ -14,3 +14,14 @@ spec:
     runAsUser: 1000
   serviceAccountName: alertmanager
   version: v0.15.2
+  storage: # Note that this section is immutable so changes require deleting and recreating the resource.
```
fejta: There's no better alternative? This sounds like a maintenance hassle.

cjwagner: Yeah it is. We had to do something similar when changing some Prow services from `NodePort` to `ClusterIP` since that is also an immutable field.

Can the bazel rules be configured to do a replace instead of apply for just one component? If so that could work, but I'm also not sure that we want to use replace on this resource (or all the other partially immutable resource this applies to) unless we need to change an immutable section. The delete and create operations might cause some downtime while the prometheus operator reacts whereas an apply wouldn't. So conditionally replacing instead of applying or if applying fails would be ideal.

FWIW I don't think this particular section will need to be changed once we have it set up.

fejta: I don't think there's that level of granularity

cjwagner:
I deleted the alertmanager resource before the deploy job ran on this PR's merge commit. Everything looks right.

PR #16404, opened 2020-02-20
Canary RBAC on sinker and horologium

fejta:
/assign @michelle192837 

Should work, and if there are problems we should be able to recover without causing a user-facing outage (just degraded service)


fejta:
ref https://github.com/kubernetes/test-infra/issues/9867

PR #16417, opened 2020-02-21
testgrid config for arm64 integration tests

lubinsz:


chases2:
Looks good to me, as far as the config goes.
/lgtm
These conformance tests have their own GCS write location. Where are the tests themselves defined?

PR #16419, opened 2020-02-21
Kops - Add back subnet mask for external IP

hakman:
The `externalIP` variable is still incorrectly set. At the moment it's missing the `/32` at the end, as it was previously in `kops-e2e-runner.sh`:
https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/kops/8603/pull-kops-e2e-kubernetes-aws/1230779675597344769#1:build-log.txt%3A552

Followup to #16388 and #16401.


rifelpet:
Oops, thanks for catching this
/lgtm

PR #16420, opened 2020-02-21
migratestatus tool: add url check for `-description` parameter

dhiller:
We encountered a 422 response when we set `-description` to a non url value:

```
{"component":"migratestatus","error":"status code 422 not one of [201], body: {\"message\":\"Validation Failed\",\"errors\":[{\"resource\":\"Status\",\"code\":\"custom\",\"field\":\"target_url\",\"message\":\"target_url must use http(s) scheme\"}],\"docu
mentation_url\":\"https://developer.github.com/v3/repos/statuses/#create-a-status\"}","file":"maintenance/migratestatus/migratestatus.go:144","func":"main.main","level":"fatal","msg":"Error during status migration","time":"2020-02-21T12:33:34+01:00"}
``` 

This commit adds a url check to the content of the description parameter to avoid failing on the server and fail early instead.


dhiller:
Can do!

dhiller:
@stevekuznetsov I've added the tests you requested.

dhiller:
Thanks everyone, fixed the issues. PTAL.

maintenance/migratestatus/migratestatus.go
```diff
@@ -75,6 +76,13 @@ func (o *options) Validate() error {
 		return errors.New("'--dest' is required unless using '--retire' mode.\n")
 	}
 
+	if o.descriptionURL != "" {
+		_, err := url.ParseRequestURI(o.descriptionURL)
+		if err != nil {
+			return errors.New("'--description' URL is not valid\n")
```
petr-muller: nit: we should propagate the `err` out to better inform what went wrong

dhiller: Good point.

dhiller:
@michelle192837 can you please take another look?

PR #16422, opened 2020-02-21
Boskos: Add retrying to client

alvaroaleman:
This PR adds retrying to the boskos client. I explicitly choose to not do anything time-based and build it number-of-tries-based, because of the huge latency we have sometimes due to the lock contention issues.

The retrying is implemented within each individual func rather than further down, because some of them have specific http errors where we know that retrying wont help.

I recommend looking at the diff with whitespace changes hidden.

/cc @bbguimaraes

### DIFF
--- a/boskos/client/client.go
+++ b/boskos/client/client.go
@@ -33,10 +33,11 @@ import (
 	"syscall"
 	"time"
 
-	"github.com/hashicorp/go-multierror"
-
 	"github.com/google/uuid"
+	"github.com/hashicorp/go-multierror"
 	"github.com/sirupsen/logrus"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+
 	"k8s.io/test-infra/boskos/common"
 	"k8s.io/test-infra/boskos/storage"
 	"k8s.io/test-infra/prow/config/secret"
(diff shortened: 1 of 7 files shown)


alvaroaleman:
/hold

Tests need fixing

alvaroaleman:
/hold cancel
tests pass

boskos/client/client.go
```diff
@@ -450,43 +471,62 @@ func (c *Client) Release(name, dest string) error {
...
 
-	if resp.StatusCode != http.StatusOK {
-		return fmt.Errorf("status %s, status code %v updating %s", resp.Status, resp.StatusCode, name)
+	work := func(retriedErrs *[]error) (bool, error) {
+		// As the body is an io.Reader and hence its content
+		// can only be read once, we have to copy it for every request we make
+		var body io.Reader
+		if bodyData != nil {
+			body = bytes.NewBuffer(bodyData.Bytes())
+		}
```
bbguimaraes: A `bytes.Reader` can be used instead.

boskos/client/client.go
```diff
@@ -640,3 +693,34 @@ func isDialErrorRetriable(err error) bool {
...
+type workFunc func(retriedErrs *[]error) (bool, error)
+
+// SleepFunc is called when requests are retried. This may be replaced in tests.
+var SleepFunc = time.Sleep
+
+func retry(work workFunc) error {
+	var retriedErrs []error
+
+	maxAttempts := 4
+	for i := 0; i < maxAttempts; i++ {
```
bbguimaraes: The `+1` and `-1` below can be removed by making the loop index 1-based.

boskos/client/client.go
```diff
@@ -640,3 +693,34 @@ func isDialErrorRetriable(err error) bool {
...
+		success, err := work(&retriedErrs)
+		if err != nil {
+			return err
```
bbguimaraes: May be useful to return `append(retriedErrs, err)` to differentiate immediate from eventual failures.

boskos/client/client.go
```diff
@@ -640,3 +693,34 @@ func isDialErrorRetriable(err error) bool {
...
+		}
+		if success {
+			return nil
+		}
+		if i == maxAttempts-1 {
+			break
+		}
+
+		SleepFunc(time.Duration(math.Pow(float64(i+1), 2)) * time.Second)
```
bbguimaraes: This can be just `(i+1) * (i+1) * time.Second` (or `i * i * …`, see the first comment), `pow` is a heavy operation.

boskos/client/client.go
alvaroaleman: The problem is that some of the errors returned by the work funcs return specific errors other places in the code check on. So we would have to first adjust all dependent code to use errors.Is, rather than equal checks

boskos/client/client.go
alvaroaleman: Hm yeah, I didn't do it initially because a loop that starts at index one breaks eveyones assumptions, but then again it does make the code much more readable

boskos/client/client.go
alvaroaleman: Not on line 511, otherwise we will do a NPD down the line because of "interface is only nil if type and value are unset". Updated to a Reader in line 513

alvaroaleman:
@bbguimaraes PTAL

boskos/client/client.go
alvaroaleman: (On top of that, utilerrors.Aggregate does not support matching via errors.Is today. I'll make an upstream PR for that so it gets in asap due to their long release cycles)

bbguimaraes: Sounds reasonable.

bbguimaraes:
Implementation LGTM.

PR #16431, opened 2020-02-22
Kops - Set interval to 1h for Cilium network plugins test

hakman:
There were quite a few changes in Cilium lately and would help to get more results before the next release.


rifelpet:
we'll revert it after the next release 👍 
/lgtm
/approve

PR #16432, opened 2020-02-22
Kops - Reorder e2e jobs in TestGrid

hakman:
There are quite a lot of jobs in the Kops TestGrid tab and would help to see them grouped by purpose.


rifelpet:
looks good. we might lose some job history due to the renames but i think thats acceptable.

/lgtm
/approve

PR #16438, opened 2020-02-23
[Doc] Added artifact-uploader entry to prow/cmd/README.md

eddycharly:
This pr adds `artifact-uploader` to the list of auxiliary components in the `prow/cmd/README.md` doc file.


fejta:
/assign @Katharine 

How does this and crier relate?

Katharine:
Technically their functionality does not overlap at all - IIRC this uploads pod logs when jobs terminate; crier uploads everything else at appropriate times (including when jobs terminate).

Philosophically, crier could perhaps subsume this (though it is not necessary that it would). In practice, we have thus far declined to use something like artifact-uploader (though it would be helpful when jobs fail catastrophically without uploading logs) due to concerns about apiserver load.

eddycharly:
Even if it is not used by the main prow instance it could be useful to someone, adding the doc doesn’t hurt.

It could be useful for uploading pod logs when using Tekton as there is no decoration support for Tekton yet.

alvaroaleman:
We actually removed the artifact uploader in #16633, because it doesn't work at all in a setup where the cluster that holds the prowjob CRs is not the same as the cluster that runs the jobs. Also, anything it did that is not covered by crier already and is still wanted should probably just get moved into crier.
/close

PR #16452, opened 2020-02-24
kubetest: add AKS deployer

alexeldeib:
This PR breaks the single existing deployer for AKS into two Azure deployers, one using AKS-Engine and one for AKS. Currently they share a lot of flags, and I haven't gone into a deep effort to dedupe or shuffle some of these around since they're use somewhat deeply throughout kubetest. I wanted to open this as-is and start getting feedback/validation before I attempted any major cleanup.

I'll open a separate PR in k/k shortly with some of the even-less-polished bash to integrate with kubemark and clusterloader2 for scalability tests.

### DIFF
--- a/kubetest/aks.go
+++ b/kubetest/aks.go
@@ -1,0 +1,303 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"context"
+	"crypto/rand"
+	"crypto/rsa"
+	"crypto/x509"
+	"encoding/json"
+	"encoding/pem"
+	"fmt"
+	"io/ioutil"
+	"log"
+	"net/http"
+	"os"
+	"os/exec"
+	"path"
+	"strings"
+	"time"
+
+	"github.com/Azure/azure-sdk-for-go/services/containerservice/mgmt/2019-10-01/containerservice"
+	"github.com/Azure/go-autorest/autorest/azure"
+	uuid "github.com/satori/go.uuid"
+	"golang.org/x/crypto/ssh"
+)
+
+type aksDeployer struct {
+	azureCreds    *Creds
+	azureClient   *AzureClient
+	templateUrl   string
+	outputDir     string
+	resourceGroup string
+	resourceName  string
+	location      string
+	k8sVersion    string
+}
+
+func newAksDeployer() (*aksDeployer, error) {
+	if err := validateAksFlags(); err != nil {
+		return nil, err
+	}
+
+	creds, err := getAzCredentials()
+	if err != nil {
+		return nil, fmt.Errorf("failed to get azure credentials: %v", err)
+	}
+	env, err := azure.EnvironmentFromName(*aksAzureEnv)
+	if err != nil {
+		return nil, fmt.Errorf("failed to determine azure environment: %v", err)
+	}
+
+	var client *AzureClient
+	if client, err = getAzureClient(env,
+		creds.SubscriptionID,
+		creds.ClientID,
+		creds.TenantID,
+		creds.ClientSecret,
+	); err != nil {
+		return nil, fmt.Errorf("error trying to get Azure Client: %v", err)
+	}
+
+	tempdir, err := ioutil.TempDir(os.Getenv("HOME"), "aks")
+	if err != nil {
+		return nil, fmt.Errorf("error creating tempdir: %v", err)
+	}
+
+	return &aksDeployer{
+		azureCreds:    creds,
+		azureClient:   client,
+		templateUrl:   *aksTemplateURL,
+		outputDir:     tempdir,
+		resourceGroup: *aksResourceGroupName,
+		resourceName:  *aksResourceName,
+		location:      *aksLocation,
+		k8sVersion:    *aksOrchestratorRelease,
+	}, nil
+}
+
+func validateAksFlags() error {
+	if *aksCredentialsFile == "" {
+		return fmt.Errorf("no credentials file path specified")
+	}
+	if *aksResourceName == "" {
+		*aksResourceName = "kubetest-" + uuid.NewV1().String()
+	}
+	if *aksResourceGroupName == "" {
+		*aksResourceGroupName = *aksResourceName
+	}
+	if *aksDNSPrefix == "" {
+		*aksDNSPrefix = *aksResourceName
+	}
+	return nil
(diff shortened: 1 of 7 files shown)


feiskyer:
@alexeldeib is this ready for review?

alexeldeib:
@feiskyer it's a bit rough still, but enough of the plumbing works to run conformance. 

If you're open to reviewing as-is, we can work towards an initial merge, and then follow up with a periodic prowjob to run conformance and use that as signal for where to continue iterating.

kubetest/aks.go
```diff
@@ -0,0 +1,297 @@
...
+	if err != nil {
+		return fmt.Errorf("error downloading AKS cluster template: %v with error %v", a.templateUrl, err)
+	}
+
+	template, err := ioutil.ReadFile(templateFile)
+	if err != nil {
+		return fmt.Errorf("failed to read downloaded cluster template file: %v", err)
+	}
+
+	var model containerservice.ManagedCluster
```
alexeldeib: i'd like to convert this to raw JSON and use a REST client so we can use any version, but for now I will do it with a specific type.

alexeldeib:
@feiskyer please do take a pass on review on this. I think the main things we will want to change will be `prepareAks()` + flags to plumb for kubemark/scalability. For this cut, I'm focusing only on the kubetest deployer and targeting conformance. @marwanad is going to pick up this side: https://github.com/kubernetes/kubernetes/pull/88477 and drive the kubemark integrations separately.

kubetest/aks.go
```diff
@@ -0,0 +1,297 @@
+/*
+Copyright 2018 The Kubernetes Authors.
```
marwanad: is the boilerplate header correct here?

kubetest/aks.go
```diff
@@ -0,0 +1,297 @@
...
+func newAksDeployer() (*aksDeployer, error) {
+	if err := validateAksFlags(); err != nil {
+		return nil, err
+	}
+
+	creds, err := getAzCredentials()
+	if err != nil {
+		return nil, fmt.Errorf("failed to get azure credentials: %v", err)
+	}
+	env, err := azure.EnvironmentFromName(*aksAzureEnv)
```
marwanad: just double-checking, this is still in the aks-engine deployer and i guess this is fine for the PR purposes and the goal is to split into a common Azure or even just define them locally in `aks.go` based on what we need later?

alexeldeib: I think i'd like to split the flags and share a common azure set. Since you'll never invoke with two deployers, I didn't think it was a high priority fix. Something frustrating is that the flags are mostly used directly from different parts of the code, not passed down, which makes it a slightly more involved change, but nothing crazy.

kubetest/aks.go
alexeldeib: this actually came up in cluster-api -- I don't know if we're supposed to update these or not?

feiskyer: s/2018/2020/

kubetest/aks.go
```diff
@@ -0,0 +1,297 @@
...
+		return fmt.Errorf("no credentials file path specified")
+	}
+	if *aksResourceName == "" {
+		*aksResourceName = "kubetest-" + uuid.NewV1().String()
+	}
+	if *aksResourceGroupName == "" {
+		*aksResourceGroupName = *aksResourceName
+	}
+	if *aksDNSPrefix == "" {
+		*aksDNSPrefix = *aksResourceName
```
feiskyer: would dns prefix too long since uuid is used in the name?

kubetest/aks.go
```diff
@@ -0,0 +1,297 @@
...
+
+	_, sshPublicKey, err := newSSHKeypair(4096)
+	if err != nil {
+		return fmt.Errorf("failed to generate ssh key for cluster creation: %v", err)
+	}
+
+	*(*model.LinuxProfile.SSH.PublicKeys)[0].KeyData = string(sshPublicKey)
+	model.ManagedClusterProperties.DNSPrefix = aksDNSPrefix
+	model.ManagedClusterProperties.ServicePrincipalProfile.ClientID = &a.azureCreds.ClientID
+	model.ManagedClusterProperties.ServicePrincipalProfile.Secret = &a.azureCreds.ClientSecret
```
feiskyer: I think it's better to also set region and kubernetes version here

kubetest/e2e.go
```diff
@@ -615,33 +615,53 @@ func kubemarkUp(dump string, o options, deploy deployer) error {
...
+		masterInternalIP, err = control.Output(exec.Command(
+			"gcloud", "compute", "instances", "describe",
+			os.Getenv("MASTER_NAME"),
+			"--project="+o.gcpProject,
+			"--zone="+o.gcpZone,
+			"--format=value(networkInterfaces[0].networkIP)"))
+		if err != nil {
+			return fmt.Errorf("failed to get masterInternalIP: %v", err)
+		}
+	case "aks":
```
feiskyer: @nilo19 do we need those env for aks-engine deployer?

kubetest/aks.go
alexeldeib: The generated names with form `kubetest-$UUID` work okay in Azure. If the user provides one, I'd expect them to make sure it validates properly.

kubetest/aks.go
alexeldeib: fixed, parsing from the shared flags with aks-e currently

kubetest/aks.go
alexeldeib: fixed

feiskyer:
could you fix the bazel errors and squash the commits?

kubetest/e2e.go
```diff
@@ -615,33 +615,53 @@ func kubemarkUp(dump string, o options, deploy deployer) error {
...
-		os.Getenv("MASTER_NAME")+"-ip",
-		"--project="+o.gcpProject,
-		"--region="+o.gcpZone[:len(o.gcpZone)-2],
-		"--format=value(address)"))
-	if err != nil {
-		return fmt.Errorf("failed to get masterIP: %v", err)
+	var masterIP, masterInternalIP []byte
+
+	switch o.deployment {
+	case "gce":
```
mborsz: existing kubemark tests don't set --deployment=gce (it's not even correct value of deployment flag). We use --deployment=bash with --provider=gce

This breaks our kubermark tests in gce.

alexeldeib: Apologies for the inconvenience @mborsz. I started a revert, maybe you could take a look: https://github.com/kubernetes/test-infra/pull/16733.

Looking again at the default deployment path, is it safe to assume --deployment=bash only supports the gce provider currently? (i.e., this behavior is correct barring that i've put "gce" and not "bash" in the switch)

mborsz: My understanding is that right now gce is an only supported provider in deplyment=bash mode, but I'm not sure if we want to use this assumption here (what if some day someone adds support for some other provider)?

Personally, I would go with
```
if deployment == "aks" {
   use logic specific for aks
} else if deployment == "bash" and provider == "gce" {
   use logic for gce
} else {
  fatal error
}
```
What do you think about this solution?

alexeldeib: Makes sense to me. I’m happy to make this change on top of the revert, assuming that’s the sequence you’d prefer.

I wonder if there is some CI automation we can improve to avoid this situation in the future?

PR #16456, opened 2020-02-24
bugzilla plugin: add /bugzilla qa command

AlexNPavel:
Adds a `/bugzilla qa` command that assigns the qa contact for a bugzilla bug to the PR. To identify the github username, a graphql query is done based on the email for the QA contact in bugzilla. If the QA contact's GitHub public email is the email listed in bugzilla, the query should be able to identify the correct user.

This command also runs a refresh to make sure that the bug is valid, and only assigns a QA contact if validation succeeds.

/cc @stevekuznetsov

### DIFF
--- a/prow/plugins/bugzilla/bugzilla.go
+++ b/prow/plugins/bugzilla/bugzilla.go
@@ -18,6 +18,7 @@ limitations under the License.
 package bugzilla
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"regexp"
@@ -25,6 +26,7 @@ import (
 	"strconv"
 	"strings"
 
+	githubql "github.com/shurcooL/githubv4"
 	"github.com/sirupsen/logrus"
 
 	"k8s.io/test-infra/prow/bugzilla"
@@ -36,8 +38,9 @@ import (
 )
 
 var (
-	titleMatch   = regexp.MustCompile(`(?i)^.*?Bug ([0-9]+):`)
-	commandMatch = regexp.MustCompile(`(?mi)^/bugzilla refresh\s*$`)
+	titleMatch          = regexp.MustCompile(`(?i)^.*?Bug ([0-9]+):`)
+	refreshCommandMatch = regexp.MustCompile(`(?mi)^/bugzilla refresh\s*$`)
+	qaCommandMatch      = regexp.MustCompile(`(?mi)^/bugzilla assign-qa\s*$`)
 )
 
 const (
@@ -150,6 +153,13 @@ func helpProvider(config *plugins.Configuration, enabledRepos []config.OrgRepo) 
 		WhoCanUse:   "Anyone",
 		Examples:    []string{"/bugzilla refresh"},
 	})
+	pluginHelp.AddCommand(pluginhelp.Command{
+		Usage:       "/bugzilla assign-qa",
+		Description: "Assign PR to QA contact specified in Bugzilla",
+		Featured:    false,
+		WhoCanUse:   "Anyone",
+		Examples:    []string{"/bugzilla assign-qa"},
+	})
 	return pluginHelp, nil
 }
 
@@ -159,6 +169,7 @@ type githubClient interface {
 	GetIssueLabels(org, repo string, number int) ([]github.Label, error)
 	AddLabel(owner, repo string, number int, label string) error
 	RemoveLabel(owner, repo string, number int, label string) error
+	Query(ctx context.Context, q interface{}, vars map[string]interface{}) error
 }
 
 func handleGenericComment(pc plugins.Agent, e github.GenericCommentEvent) error {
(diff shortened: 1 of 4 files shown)


prow/plugins/bugzilla/bugzilla.go
```diff
@@ -320,6 +340,55 @@ func (e *event) comment(gc githubClient) func(body string) error {
...
+*/
+type emailToLoginQuery struct {
+	Search querySearch `graphql:"search(type:USER query:$email first:5)"`
+}
+
+// processQueryResult generates a response based on a populated emailToLoginQuery
+func processQuery(query *emailToLoginQuery, email string, log *logrus.Entry) string {
+	switch len(query.Search.Edges) {
+	case 0:
+		return fmt.Sprintf("ERROR: No GitHub users found with public email == %s", email)
```
stevekuznetsov: This looks more like a log message than something we should post in a comment. Maybe:

```
No GitHub users were found matching the public email listed for the QA contact in Bugzilla (%s), skipping assignment.
```

prow/plugins/bugzilla/bugzilla.go
```diff
@@ -320,6 +340,55 @@ func (e *event) comment(gc githubClient) func(body string) error {
...
+	case 1:
+		return fmt.Sprintf("/assign @%s", query.Search.Edges[0].Node.User.Login)
+	default:
+		response := fmt.Sprintf("ERROR: Multiple GitHub users found with public email == %s:", email)
```
stevekuznetsov: Similarly, formatting this with more prose would be better. Also let's make sure we're not at-mentioning people in this comment

prow/plugins/bugzilla/bugzilla.go
```diff
@@ -377,6 +446,20 @@ To reference a bug, add 'Bug XXX:' to the title of this pull request and request
...
+			// if bug is valid and qa command was used, identify qa contact via email
+			if e.command == qaCommand {
+				query := &emailToLoginQuery{}
+				email := bug.QAContactDetail.Email
+				queryVars := map[string]interface{}{
+					"email": githubql.String(email),
+				}
+				err := gc.Query(context.Background(), query, queryVars)
+				if err != nil {
+					response += fmt.Sprintf("\nERROR: Failed to run github query to identify user with email == %s", email)
```
stevekuznetsov: The user does not need to know that we tried to run a query -- it's immaterial to them. Can we follow the error formatting for other types of errors that tells users to contact an admin?

prow/plugins/bugzilla/bugzilla.go
```diff
@@ -377,6 +453,20 @@ To reference a bug, add 'Bug XXX:' to the title of this pull request and request
 					response += " The bug has been updated to refer to the pull request using the external bug tracker."
 				}
 			}
+			// if bug is valid and qa command was used, identify qa contact via email
+			if e.command == qaCommand {
+				query := &emailToLoginQuery{}
+				email := bug.QAContactDetail.Email
```
stevekuznetsov: What if there is no QA contact or there is no email listed?

AlexNPavel: No QA contact would result in a panic (nil pointer dereference) and no email for the contact would result in no results from the GH query. I'll add some checks for these cases.

prow/plugins/bugzilla/bugzilla_test.go
```diff
@@ -119,13 +119,21 @@ orgs:
...
+				Description: "Check Bugzilla for a valid bug referenced in the PR title",
+				Featured:    false,
+				WhoCanUse:   "Anyone",
+				Examples:    []string{"/bugzilla refresh"},
+			}, {
+				Usage:       "/bugzilla qa",
+				Description: "Assign PR to QA contact specified in Bugzilla",
+				Featured:    false,
+				WhoCanUse:   "Anyone",
+				Examples:    []string{"/bugzilla qa"},
```
petr-muller: This needs to reflect the change of command to `assign-qa` too (both usage and examples).

prow/plugins/bugzilla/bugzilla.go
```diff
@@ -150,6 +153,13 @@ func helpProvider(config *plugins.Configuration, enabledRepos []config.OrgRepo)
 		WhoCanUse:   "Anyone",
 		Examples:    []string{"/bugzilla refresh"},
 	})
+	pluginHelp.AddCommand(pluginhelp.Command{
+		Usage:       "/bugzilla qa",
+		Description: "Assign PR to QA contact specified in Bugzilla",
+		Featured:    false,
+		WhoCanUse:   "Anyone",
+		Examples:    []string{"/bugzilla qa"},
```
petr-muller: `/bugzilla assign-qa` in both usage and examples

AlexNPavel:
/retitle "bugzilla plugin: add /bugzilla assign-qa command"

AlexNPavel:
Fixed and rebased

PR #16462, opened 2020-02-25
Stop requiring triage to have secret credentials

fejta:
/assign @spiffxp @Katharine 

ref https://github.com/kubernetes/test-infra/issues/15806 #16464


Katharine:
/hold

Actually holding; we have ongoing data collection issues I want to investigate before making additional changes, even though they should be unrelated.

fejta:
What is the issue tracking these ongoing data collection issues? Do you have a resolution ETA?

spiffxp:
I believe https://github.com/kubernetes/test-infra/pull/16508 addressed the issues, go.k8s.io/triage is showing current data

PR #16471, opened 2020-02-25
Boskos cleaner as controller

alvaroaleman:
This PR adds a featuregated, off-by-default, v2 implementation of the boskos cleaner.  Advantages:

* It doesn't need to do any api calls to the boskos server when there is nothing to do, thus not making the lock contention issue we have worse (current cleaner does `for sleep 15s; do acquire && clean; done`) 
* When there is work to do, it is much quicker, as it doesn't sleep each time it did something (cleaner v1 took about 28h to clean up 2k resources that didn't contain any leased resources in their userdata, or in other words: To change their state from `toBeDeleted` to `tombstone`)
* No acquiring at all needed for resources that do not contain leased resources in their userdata
* Still save to run concurrently with the boskos server even before https://github.com/kubernetes/test-infra/pull/16367 as the final `Update` call that changes the status will fail if something else updated the object in the meantime (e.G. boskos set the owner)

/assign @sebastienvas @stevekuznetsov

### DIFF
--- a/boskos/cleaner/v2/cleaner.go
+++ b/boskos/cleaner/v2/cleaner.go
@@ -29,6 +29,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 
+	"k8s.io/test-infra/boskos/cleaner"
 	"k8s.io/test-infra/boskos/common"
 	"k8s.io/test-infra/boskos/crds"
 	"k8s.io/test-infra/boskos/mason"
@@ -37,10 +38,11 @@ import (
 const controllerName = "boskos-cleaner"
 
 // Add creates a new cleaner controller
-func Add(mgr manager.Manager) error {
+func Add(mgr manager.Manager, boskosClient cleaner.RecycleBoskosClient) error {
 	reconciler := &reconciler{
-		ctx:    context.Background(),
-		client: mgr.GetClient(),
+		ctx:          context.Background(),
+		client:       mgr.GetClient(),
+		boskosClient: boskosClient,
 	}
 
 	c, err := controller.New(controllerName, mgr, controller.Options{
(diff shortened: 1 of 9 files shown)


boskos/cleaner/v2/cleaner.go
```diff
@@ -0,0 +1,125 @@
...
+	isDynamic, err := r.isResourceDynamic(resourceObject)
+	if err != nil {
+		return fmt.Errorf("failed to check if resource is dynamic: %v", err)
+	}
+	if !isDynamic {
+		return nil
+	}
+
+	resourceObject.Status.Owner = controllerName
+	if err := r.client.Update(r.ctx, resourceObject); err != nil {
```
stevekuznetsov: Is this a common pattern? Why not let the controller race with itself if it gets multiple events to handle and let one win?

alvaroaleman: The controller doesn't race with itself because it has leader election and the underlying machinery makes sure that only one routine works on a given key.

When using multiple controller, just trying optimistically and being okay with sometimes failing the update would be the correct approach, this here is done only to prevent boskos from somehow leasing out the resource.

alvaroaleman:
Hmm, bazel compile throws errors about things a `go test` is fine with :thinking:

boskos/cleaner/v2/cleaner.go
stevekuznetsov: That would only occur if someone actually tried to lease a toBeDeleted resource, right? I just wonder if we need to double the API calls with this approach or if we can just do the state update and let other clients retry if their updates get muddled.

alvaroaleman: Its more like:

* We find the empty resource and start doing stuff
* Someone else acquires the resource, boskos updates the owner field
* Our state transition fails, because the object got changed

Hypothetically this may be okay because it will eventually get back to unowned, we clean again and transition the state. But it muddles with the boskos concept of "handing out crds to arbitrary actors that do arbitrary things with it".

boskos/cleaner/v2/cleaner.go
```diff
@@ -0,0 +1,125 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/source"
+
+	"k8s.io/test-infra/boskos/cleaner"
+	"k8s.io/test-infra/boskos/common"
+	"k8s.io/test-infra/boskos/crds"
+)
+
+const controllerName = "boskos-cleaner"
+
+// Add creates a new claner controller
```
petr-muller: nit: claner -> cleaner

alvaroaleman: Fixed

PR #16476, opened 2020-02-25
Try to warn about not using secret json files

fejta:
/assign @BenTheElder @Katharine @amwat 

Add some log visibility about migrating to workload identity

ref #15806


fejta:
Can I get a review please?

Issue #16303, opened 2020-02-14
Test-Infra-Push-Prow job broken


alvaroaleman:
**What happened**:

https://prow.k8s.io/?job=post-test-infra-push-prow is pretty red

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

**Please provide links to example occurrences, if any**:

**Anything else we need to know?**:

/kind oncall-hotlist

stevekuznetsov:
```
 2020/02/14 14:11:10 Successfully pushed Docker image to gcr.io/k8s-prow/grandmatriarch:latest-root
ERROR: (gcloud.auth.docker-helper) timed out
This may be due to network connectivity issues. Please check your network settings, and the status of the service you are trying to reach. 
```

stevekuznetsov:
```
 2020/02/14 14:13:53 Error pushing image to gcr.io/k8s-prow-edge/issue-creator:latest: unable to push image to gcr.io/k8s-prow-edge/issue-creator:latest: error getting credentials - err: exit status 1, out: ``
ERROR: (gcloud.auth.docker-helper) You do not currently have an active account selected.
Please run:
  $ gcloud auth login
to obtain new credentials, or if you have already logged in with a
different account:
  $ gcloud config set account ACCOUNT
to select an already authenticated account to use. 
```

stevekuznetsov:
https://github.com/kubernetes/test-infra/pull/16265/files ?

fejta:
Suspect this job is DOSing the metadata server

fejta:
We already have retries (and we are retrying): 

https://github.com/kubernetes/test-infra/blob/c71503f9a7646488b0bef5ac2d5ad485fb25e6ab/prow/push.sh#L69-L78

I'll file an internal bug around workload identity, but we should probably just make that job start using the secret file again until its resolved.

alvaroaleman:
Seems to be fixed
/close

PR #16481, opened 2020-02-25
Revert "Revert "Use workload identity for grandmatriach, boskos-janitor

fejta:
This reverts commit 9da9fee58b92f5f55fd29cb641bdf2820d1ca329.

/hold

/assign @michelle192837 @ixdy @Katharine 

Looks similar to https://github.com/kubernetes/test-infra/issues/16303 and applying the same fix on GKE side of things.

### DIFF
--- a/prow/cluster/boskos-janitor.yaml
+++ b/prow/cluster/boskos-janitor.yaml
@@ -24,16 +24,7 @@ spec:
         - --resource-type=gke-project
         - --pool-size=20
         - --
-        - --service_account=/etc/service-account/service-account.json
         - --hours=0
-        volumeMounts:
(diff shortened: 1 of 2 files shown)


ixdy:
can we document what this "GKE fix" is? it looks like this has broken things at least twice?

fejta:
Previously fixed the service cluster, this is to fix the build cluster. 

Fix involves whitelisting the cluster for a version of workload identity that has performance improvements for heavy workloads.

fejta:
/hold cancel

At Google we converted boskos over to use workload identity without issue (applying the same cluster fix) so we should be good to go here

PR #16484, opened 2020-02-25
bugzilla plugin: list validations run on refresh

AlexNPavel:
This PR makes the bugzilla plugin more verbose on successful validation, listing which validations it ran to determine that the bug was valid.

/cc @stevekuznetsov

### DIFF
--- a/prow/plugins/bugzilla/bugzilla.go
+++ b/prow/plugins/bugzilla/bugzilla.go
@@ -445,14 +445,16 @@ To reference a bug, add 'Bug XXX:' to the title of this pull request and request
 				}
 			}
 
+			response += "\n\n<details>"
 			if len(validationsRun) == 0 {
-				response += "\n\nNo validations were run on this bug."
+				response += "<summary>No validations were run on this bug</summary>"
 			} else {
-				response += "\n\nValidations run on this bug:"
+				response += fmt.Sprintf("<summary>%d validation(s) run on this bug</summary>", len(validationsRun))
 			}
 			for _, validation := range validationsRun {
 				response += fmt.Sprint("\n\t- ", validation)
 			}
+			response += "</details>"
 
 			// if bug is valid and qa command was used, identify qa contact via email
 			if e.assign {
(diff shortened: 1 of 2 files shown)


prow/plugins/bugzilla/bugzilla.go
```diff
@@ -546,6 +557,12 @@ func validateBug(bug bugzilla.Bug, dependents []bugzilla.Bug, options plugins.Bu
 			was = "is"
 		}
 		errors = append(errors, fmt.Sprintf("expected the bug to %sbe open, but it %s", not, was))
+	} else if options.IsOpen != nil {
+		was := "isn't"
+		if *options.IsOpen {
+			was = "is"
+		}
+		validations = append(validations, fmt.Sprintf("bug %s open", was))
```
stevekuznetsov: This is missing the expected state

prow/plugins/bugzilla/bugzilla_test.go
```diff
@@ -802,6 +814,10 @@ Instructions for interacting with me using PR comments are available [here](http
 			expectedLabels: []string{"bugzilla/valid-bug"},
 			expectedComment: `org/repo#1:@user: This pull request references [Bugzilla bug 123](www.bugzilla/show_bug.cgi?id=123), which is valid.
 
+Validations run on this bug:
+	- dependent bug [Bugzilla bug 124](www.bugzilla/show_bug.cgi?id=124) is in the state VERIFIED, which is one of the valid states (VERIFIED)
+	- bug has dependents
```
stevekuznetsov: This is only an implicit requirement, right?

prow/plugins/bugzilla/bugzilla_test.go
```diff
@@ -626,6 +626,8 @@ Instructions for interacting with me using PR comments are available [here](http
 			expectedLabels: []string{"bugzilla/valid-bug"},
 			expectedComment: `org/repo#1:@user: This pull request references [Bugzilla bug 123](www.bugzilla/show_bug.cgi?id=123), which is valid.
 
+No validations were run on this bug.
```
stevekuznetsov: I would expect at least one complicated test case in this which has a long list.

prow/plugins/bugzilla/bugzilla_test.go
AlexNPavel: Yes, but it's a validation we run, so I added it to the list. We can remove it if you think it's unnecessary.

petr-muller:
LOL what, I'm an approver?

PR #16507, opened 2020-02-26
Boskos: Keep ObjectMeta when updating DRLC from configfile

### LINKED ISSUE #16504: boskos fails to update dynamicresourcelifecycles
**What happened**: after updating to the latest version of boskos, we're now seeing the following error message:

```
{"component":"boskos","error":"failed to update dlrc [resourcetype]: dynamicresourcelifecycles.boskos.k8s.io \"[resourcetype]\" is invalid: metadata.resourceVersion: Invalid value: 0x0: must be specified for an update","file":"boskos/ranch/storage.go:528","func":"k8s.io/test-infra/boskos/ranch.(*Storage).persistDynamicResourceLifeCycles","level":"error","msg":"unable to update resource type life cycle [resourcetype]","time":"2020-02-26T18:57:21Z"}
```

I suspect this is related to https://github.com/kubernetes/test-infra/pull/16367.
In particular, I think this questionable bit of code no longer works:
https://github.com/kubernetes/test-infra/blob/de5dfb3349d7688b933e87cf0ac4aa56973224f7/boskos/ranch/storage.go#L416-L420

We used to implicitly assume we'd get the old version of the DRLC object first and then update (which would include the `resourceVersion`), but we no longer do that, so passing the "new" DRLC (with no resource version) to `Update` is incorrect behavior that needs to be fixed.

/area boskos
cc @chemikadze @alvaroaleman

alvaroaleman:
Fixes #16504

/assign @ixdy

### DIFF
--- a/boskos/ranch/storage.go
+++ b/boskos/ranch/storage.go
@@ -415,6 +415,9 @@ func (s *Storage) syncDynamicResourceLifeCycles(newDRLCByType, existingDRLCByTyp
 	for _, existingDRLC := range existingDRLCByType {
 		newDRLC, existsInNew := newDRLCByType[existingDRLC.Name]
 		if existsInNew {
+			// Copy the ObjectMeta so we can compare, the update doesn't fail due to unset ResourceVersion
+			// and to keep any additional metadata that was added.
+			newDRLC.ObjectMeta = *existingDRLC.ObjectMeta.DeepCopy()
 			if !reflect.DeepEqual(existingDRLC, newDRLC) {
 				dRLCToUpdate = append(dRLCToUpdate, newDRLC)
 			}


boskos/ranch/storage.go
```diff
@@ -415,7 +415,9 @@ func (s *Storage) syncDynamicResourceLifeCycles(newDRLCByType, existingDRLCByTyp
 	for _, existingDRLC := range existingDRLCByType {
 		newDRLC, existsInNew := newDRLCByType[existingDRLC.Name]
 		if existsInNew {
-			if !reflect.DeepEqual(existingDRLC, newDRLC) {
+			if !reflect.DeepEqual(existingDRLC.Spec, newDRLC.Spec) {
```
alvaroaleman: I am not 100% happy with this as this may surprise us down the line should we ever add a Status field. Alternate would be to copy `ObjectMeta` from `existingDLRC` onto `newDLRC`, but that seems just as bad, just different.

ixdy: The alternate-alternate would be to copy the spec from new into existing, then add the existing object to the list to update. I'm not sure which of these is best (or least worst).

ixdy: I don't especially like how the current version loses any other metadata that may have been added to the object.

alvaroaleman: Updated to just copy ObjectMeta

boskos/ranch/storage.go
```diff
@@ -415,7 +415,11 @@ func (s *Storage) syncDynamicResourceLifeCycles(newDRLCByType, existingDRLCByTyp
 	for _, existingDRLC := range existingDRLCByType {
 		newDRLC, existsInNew := newDRLCByType[existingDRLC.Name]
 		if existsInNew {
+			// Copy the ObjectMeta so we can compare, the update doesn't faile due to unset ResourceVersion
+			// and to keep any additional metadata that was added.
+			newDRLC.ObjectMeta = existingDRLC.ObjectMeta
 			if !reflect.DeepEqual(existingDRLC, newDRLC) {
+				newDRLC.ResourceVersion = existingDRLC.ResourceVersion
```
ixdy: is it necessary to do this now if we're copying `ObjectMeta` already?

boskos/ranch/storage.go
```diff
@@ -415,7 +415,11 @@ func (s *Storage) syncDynamicResourceLifeCycles(newDRLCByType, existingDRLCByTyp
 	for _, existingDRLC := range existingDRLCByType {
 		newDRLC, existsInNew := newDRLCByType[existingDRLC.Name]
 		if existsInNew {
+			// Copy the ObjectMeta so we can compare, the update doesn't faile due to unset ResourceVersion
+			// and to keep any additional metadata that was added.
+			newDRLC.ObjectMeta = existingDRLC.ObjectMeta
```
ixdy: should this be a deep copy?

boskos/ranch/storage.go
```diff
@@ -415,7 +415,11 @@ func (s *Storage) syncDynamicResourceLifeCycles(newDRLCByType, existingDRLCByTyp
 	for _, existingDRLC := range existingDRLCByType {
 		newDRLC, existsInNew := newDRLCByType[existingDRLC.Name]
 		if existsInNew {
+			// Copy the ObjectMeta so we can compare, the update doesn't faile due to unset ResourceVersion
```
ixdy: nit: typo on fail

boskos/ranch/storage.go
```diff
@@ -415,6 +415,9 @@ func (s *Storage) syncDynamicResourceLifeCycles(newDRLCByType, existingDRLCByTyp
 	for _, existingDRLC := range existingDRLCByType {
 		newDRLC, existsInNew := newDRLCByType[existingDRLC.Name]
 		if existsInNew {
+			// Copy the ObjectMeta so we can compare, the update doesn't fail due to unset ResourceVersion
+			// and to keep any additional metadata that was added.
+			newDRLC.ObjectMeta = *existingDRLC.ObjectMeta.DeepCopy()
```
ixdy: nit: could also be `existingDRLC.ObjectMeta.DeepCopyInto(&newDRLC.ObjectMeta)` I think, though this should work.

PR #16508, opened 2020-02-26
Enlarge kettle PVC again

Katharine:
Resize the kettle PVC (and, therefore the PV).

Also add the `ssd` storage class, which is referenced by the pvc but does not actually exist.

These changes have already been applied to the cluster in question.

/cc @BenTheElder


fejta:
> These changes have already been applied to the cluster in question.

I guess we don't yet have a postsubmit job to apply these on merge?

PR #16510, opened 2020-02-26
    Adding support for having unique HMAC token per github repository or organization.

agarg2008:
Adding support for having unique HMAC token per github repository or organization.

New secret file would look like this : 
```
'*':
  - value: key1
    expiry: 9999-10-02T15:00:00Z
  - value: key2
    expiry: 2018-10-02T15:00:00Z
foo:
  - value: key3
    expiry: 9999-10-02T15:00:00Z
  - value: key4
    expiry: 2018-10-02T15:00:00Z
foo/bar:
  - value: 123abc
    expiry: 9999-10-02T15:00:00Z
  - value: key6
    expiry: 2018-10-02T15:00:00Z
```

https://docs.google.com/document/d/1m3nFfPt95kbe6BQdWp5BrZMSPkBo5mFb2nuMq1OeE24/edit?ts=5e1f65f9#heading=h.6dyyhz5krl9v

### DIFF
--- a/prow/github/hmac.go
+++ b/prow/github/hmac.go
@@ -20,11 +20,31 @@ import (
 	"crypto/hmac"
 	"crypto/sha1"
 	"encoding/hex"
+	"encoding/json"
+	"errors"
+	"github.com/sirupsen/logrus"
+	"sigs.k8s.io/yaml"
 	"strings"
+	"time"
 )
 
+// hmacSecret contains a hmac token and its expiration time.
+type hmacSecret struct {
+	Value  string    `json:"value"`
+	Expiry time.Time `json:"expiry"`
+}
+
+// hmacsForRepo contains all hmac tokens configured for a repo, org or globally.
+type hmacsForRepo []hmacSecret
+
 // ValidatePayload ensures that the request payload signature matches the key.
-func ValidatePayload(payload []byte, sig string, key []byte) bool {
+func ValidatePayload(payload []byte, sig string, tokenGenerator func() []byte) bool {
+	var event GenericEvent
+	if err := json.Unmarshal(payload, &event); err != nil {
+		logrus.WithError(err).Info("validatePayload couldn't unmarshal the github event payload")
+		return false
+	}
+
 	if !strings.HasPrefix(sig, "sha1=") {
 		return false
 	}
@@ -33,10 +53,23 @@ func ValidatePayload(payload []byte, sig string, key []byte) bool {
 	if err != nil {
 		return false
 	}
-	mac := hmac.New(sha1.New, key)
-	mac.Write(payload)
-	expected := mac.Sum(nil)
-	return hmac.Equal(sb, expected)
+
+	hmacs, err := extractHmacs(event.Repo.FullName, tokenGenerator)
+	if err != nil {
+		logrus.WithError(err).Error("couldn't unmarshal the hmac secret")
+		return false
+	}
+
+	// If we have a match with any valid hmac, we can validate successfully.
+	for _, key := range hmacs {
+		mac := hmac.New(sha1.New, key)
+		mac.Write(payload)
+		expected := mac.Sum(nil)
+		if hmac.Equal(sb, expected) {
+			return true
+		}
+	}
+	return false
 }
 
 // PayloadSignature returns the signature that matches the payload.
@@ -45,4 +78,47 @@ func PayloadSignature(payload []byte, key []byte) string {
 	mac.Write(payload)
 	sum := mac.Sum(nil)
 	return "sha1=" + hex.EncodeToString(sum)
+}
+
+// extractHmacs returns all *valid* HMAC tokens for given repository/organization.
+// It considers only the tokens at the most specific level configured for the given repo.
+// For example : if a token for repo is present and it doesn't match the repo, we will
+// not try to find a match with org level token. However if no token is present for repo,
+// we will try to match with org level.
+func extractHmacs(repo string, tokenGenerator func() []byte) ([][]byte, error) {
+	t := tokenGenerator()
+	repoToTokenMap := map[string]hmacsForRepo{}
+
+	if err := yaml.Unmarshal(t, &repoToTokenMap); err != nil {
+		// To keep backward compatibility, we are going to assume that in case of error,
+		// whole file is a single line hmac token.
+		//TODO : Once this code has been released and file has been moved to new format,
+		// we should delete this code and return error.
+		logrus.WithError(err).Info("couldn't unmarshal the hmac secret as hierarchical file. Parsing as single token format")
+		return [][]byte{t}, nil
+	}
+
+	orgName := strings.Split(repo, "/")[0]
+
+	if val, ok := repoToTokenMap[repo]; ok {
+		return extractValidTokens(val), nil
+	}
+	if val, ok := repoToTokenMap[orgName]; ok {
+		return extractValidTokens(val), nil
+	}
+	if val, ok := repoToTokenMap["*"]; ok {
+		return extractValidTokens(val), nil
+	}
+	return nil, errors.New("Invalid content in secret file. Global token doesn't exist.")
+}
+
+// extractValidTokens return valid tokens for any given level of tree. Validity is determined based on time till they are valid.
+func extractValidTokens(allTokens hmacsForRepo) [][]byte {
+	var validTokens [][]byte
+	for _, token := range allTokens {
+		if token.Expiry.After(time.Now()) {
+			validTokens = append(validTokens, []byte(token.Value))
+		}
+	}
+	return validTokens
 }
--- a/prow/github/hmac_test.go
+++ b/prow/github/hmac_test.go
@@ -20,53 +20,53 @@ import (
 	"testing"
 )
 
+var globalSecret = `
+'*':
+  - value: abc
+    expiry: 9999-10-02T15:00:00Z
+  - value: key2
+    expiry: 2018-10-02T15:00:00Z
+`
+
+var defaultTokenGenerator = func() []byte {
+	return []byte(globalSecret)
+}
+
 // echo -n 'BODY' | openssl dgst -sha1 -hmac KEY
 func TestValidatePayload(t *testing.T) {
 	var testcases = []struct {
-		payload string
-		sig     string
-		key     string
-		valid   bool
+		payload        string
+		sig            string
+		tokenGenerator func() []byte
+		valid          bool
 	}{
 		{
 			"{}",
 			"sha1=db5c76f4264d0ad96cf21baec394964b4b8ce580",
-			"abc",
+			defaultTokenGenerator,
 			true,
 		},
 		{
 			"{}",
 			"db5c76f4264d0ad96cf21baec394964b4b8ce580",
-			"abc",
+			defaultTokenGenerator,
 			false,
 		},
 		{
 			"{}",
 			"",
-			"abc",
+			defaultTokenGenerator,
 			false,
-		},
-		{
-			"",
-			"sha1=cc47e3c0aa0c2984454476d061108c0b110177ae",
-			"abc",
-			true,
-		},
-		{
-			"",
-			"sha1=fbdb1d1b18aa6c08324b7d64b71fb76370690e1d",
-			"",
-			true,
 		},
 		{
 			"{}",
 			"",
-			"abc",
+			defaultTokenGenerator,
 			false,
 		},
 	}
 	for _, tc := range testcases {
-		if ValidatePayload([]byte(tc.payload), tc.sig, []byte(tc.key)) != tc.valid {
+		if ValidatePayload([]byte(tc.payload), tc.sig, tc.tokenGenerator) != tc.valid {
 			t.Errorf("Wrong validation for %+v", tc)
 		}
 	}
--- a/prow/github/types.go
+++ b/prow/github/types.go
@@ -209,6 +209,13 @@ const (
 	PullRequestActionReadyForReview PullRequestEventAction = "ready_for_review"
 )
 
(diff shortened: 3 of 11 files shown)


prow/github/hmac.go
```diff
@@ -46,3 +64,37 @@ func PayloadSignature(payload []byte, key []byte) string {
...
+
+	if val, ok := repoToTokenMap[repo]; ok {
+		return extractValidToken(val)
+	}
+	if val, ok := repoToTokenMap[orgName]; ok {
+		return extractValidToken(val)
+	}
+	if val, ok := repoToTokenMap["global_hmac"]; ok {
+		return extractValidToken(val)
+	}
```
Katharine: I wonder whether we should instead permit all the possible for any parent - for instance, if we're migrating from repo-level webhooks to org-level webhooks, we would need to add the new key for every existing repo as part of the migration. We could avoid that with something like this:

```suggestion
  var validTokens []byte
	if val, ok := repoToTokenMap[repo]; ok {
		validTokens = append(validTokens, extractValidToken(val)...)
	}
	if val, ok := repoToTokenMap[orgName]; ok {
		validTokens = append(validTokens, extractValidToken(val)...)
	}
	if val, ok := repoToTokenMap["global_hmac"]; ok {
		validTokens = append(validTokens, extractValidToken(val)...)
	}
	return validTokens
```

(I'm not certain this suggestion is a good idea, but no reason it wouldn't be comes to mind. To the best of my knowledge, anyone who can see org-level hooks also has full control over repos - if I'm wrong, that would be a good reason to not do this. I think I may have argued against something like this before, too...)

prow/github/hmac.go
```diff
@@ -46,3 +64,37 @@ func PayloadSignature(payload []byte, key []byte) string {
...
+	repoToTokenMap := map[string]map[string]time.Time{}
+
+	err := yaml.Unmarshal(t, &repoToTokenMap)
+	if err != nil {
+		fmt.Printf("error: %v", err)
+		return nil
+	}
+
+	orgName := strings.Split(repo, "/")[0]
+	var extractValidToken = func(allTokens map[string]time.Time) [][]byte {
```
Katharine: This returns multiple tokens, so `extractValidTokens`.

Also, why is this an inner function instead of just a private top-level one?

prow/github/hmac.go
```diff
@@ -46,3 +64,37 @@ func PayloadSignature(payload []byte, key []byte) string {
 	sum := mac.Sum(nil)
 	return "sha1=" + hex.EncodeToString(sum)
 }
+
+// For given repository/organization, returns all *valid* HMAC tokens. Validity is determined based on time till they are valid.
```
Katharine: /lint

prow/github/hmac.go
```diff
@@ -46,3 +64,37 @@ func PayloadSignature(payload []byte, key []byte) string {
...
+func ExtractHmacs(repo string, tokenGenerator func() []byte) [][]byte {
```
Katharine: Is this public for a specific reason?

prow/github/types.go
```diff
@@ -209,6 +209,13 @@ const (
 	PullRequestActionReadyForReview PullRequestEventAction = "ready_for_review"
 )
 
+// A light weight struct containing just Sender and Repo as all events are expected to have this information.
```
Katharine: lightweight

prow/github/hmac.go
```diff
@@ -20,11 +20,22 @@ import (
...
 	"strings"
+	"time"
 )
 
 // ValidatePayload ensures that the request payload signature matches the key.
-func ValidatePayload(payload []byte, sig string, key []byte) bool {
+func ValidatePayload(payload []byte, sig string, tokenGenerator func() []byte) bool {
+	var event GenericEvent
+	if err := json.Unmarshal(payload, &event); err != nil {
+		fmt.Printf("error: %v", err)
```
Katharine: Use the logging library (`github.com/sirupsen/logrus`), and be specific about what errored.

```suggestion
		logrus.WithError(err).Info("ValidatePayload couldn't unmarshal the github event payload")
```

prow/github/hmac.go
```diff
@@ -46,3 +64,37 @@ func PayloadSignature(payload []byte, key []byte) string {
...
 }
+
+// For given repository/organization, returns all *valid* HMAC tokens. Validity is determined based on time till they are valid.
+func ExtractHmacs(repo string, tokenGenerator func() []byte) [][]byte {
+	t := tokenGenerator()
+	repoToTokenMap := map[string]map[string]time.Time{}
+
+	err := yaml.Unmarshal(t, &repoToTokenMap)
+	if err != nil {
+		fmt.Printf("error: %v", err)
```
Katharine: Logging library, and also be specific about what errored.

```suggestion
		logrus.WithError(err).Error("Couldn't unmarshal the hmac secret")
```

prow/github/hmac.go
```diff
@@ -46,3 +64,37 @@ func PayloadSignature(payload []byte, key []byte) string {
 	sum := mac.Sum(nil)
 	return "sha1=" + hex.EncodeToString(sum)
 }
+
+// For given repository/organization, returns all *valid* HMAC tokens. Validity is determined based on time till they are valid.
+func ExtractHmacs(repo string, tokenGenerator func() []byte) [][]byte {
+	t := tokenGenerator()
+	repoToTokenMap := map[string]map[string]time.Time{}
```
Katharine: I'd rather see the structure as originally specced, for both reader clarity and future extensibility.

prow/github/hmac.go
clarketm: Personally, I am in favor but I think you and/or @cjwagner had concerns? 
https://docs.google.com/document/d/1m3nFfPt95kbe6BQdWp5BrZMSPkBo5mFb2nuMq1OeE24/edit?disco=AAAAI5tP1FU

prow/github/hmac.go
```diff
@@ -46,3 +64,37 @@ func PayloadSignature(payload []byte, key []byte) string {
...
+		return validTokens
+	}
+
+	if val, ok := repoToTokenMap[repo]; ok {
+		return extractValidToken(val)
+	}
+	if val, ok := repoToTokenMap[orgName]; ok {
+		return extractValidToken(val)
+	}
+	if val, ok := repoToTokenMap["global_hmac"]; ok {
```
clarketm: While I admit it is *more* descriptive, I am curious why we chose `global_hmac` over `*` here ([ref](https://docs.google.com/document/d/1m3nFfPt95kbe6BQdWp5BrZMSPkBo5mFb2nuMq1OeE24/edit?disco=AAAAI7hQyF8)); the latter is what we *consistently* use elsewhere in Prow configuration to denote globals/wildcards (e.g. `job_url_prefix_config`, `default_decoration_configs`, `slack_reporter_configs`, etc)?

Katharine: `*` would be a more consistent choice, +1

prow/github/hmac.go
agarg2008: It is not clear to me how migration from repo lvl to org lvl hooks would be simpler in proposed scenario than in current scenario. We can discuss this offline and update here.

prow/github/hmac.go
agarg2008: Nope, it didn't need to be. Making private.

prow/github/hmac.go
agarg2008: Done.

prow/github/hmac.go
agarg2008: Renamed. I wanted to keep this an inner function as it has limited scope and its a short one. Don't have a strong inclination one way or the other.

prow/github/hmac.go
agarg2008: We discussed offline and decided that this change is not necessary and we will continue with what was discussed in design doc.

prow/github/types.go
```diff
@@ -1099,3 +1106,12 @@ type ProjectCard struct {
 	ContentType string `json:"content_type"`
 	ContentURL  string `json:"content_url"`
 }
+
+// HmacSecret contains a hmac token and its expiration time.
+type HmacSecret struct {
```
Katharine: I don't think this is really a public github type, as such - it probably belongs in `hmac.go`, and probably as non-public.

prow/github/hmac.go
```diff
@@ -46,3 +64,39 @@ func PayloadSignature(payload []byte, key []byte) string {
...
 }
+
+// extractHmacs returns all *valid* HMAC tokens for given repository/organization. It iterates over full tree as necessary (repo->org->global)
+func extractHmacs(repo string, tokenGenerator func() []byte) [][]byte {
+	t := tokenGenerator()
+	repoToTokenMap := map[string]HmacsForRepo{}
+
+	err := yaml.Unmarshal(t, &repoToTokenMap)
+	if err != nil {
+		logrus.WithError(err).Error("Couldn't unmarshal the hmac secret")
```
Katharine: This should probably actually return its error to be handled by its caller, instead of logging it.

(Its caller will probably then just log it, I guess)

prow/github/hmac.go
```diff
@@ -46,3 +64,39 @@ func PayloadSignature(payload []byte, key []byte) string {
 	sum := mac.Sum(nil)
 	return "sha1=" + hex.EncodeToString(sum)
 }
+
+// extractHmacs returns all *valid* HMAC tokens for given repository/organization. It iterates over full tree as necessary (repo->org->global)
+func extractHmacs(repo string, tokenGenerator func() []byte) [][]byte {
+	t := tokenGenerator()
+	repoToTokenMap := map[string]HmacsForRepo{}
+
+	err := yaml.Unmarshal(t, &repoToTokenMap)
```
Katharine: This will fail for all existing prows, which isn't something we can ship. You need to support the existing secret format (i.e. "just a string"), too.

Unfortunately, that probably means that you won't be able to do much meaningful error handling, because the old format was an arbitrary string...

prow/github/hmac_test.go
```diff
@@ -20,53 +20,53 @@ import (
...
+  - value: key2
+    expiry: 2018-10-02T15:00:00Z
+`
+
+var defaultTokenGenerator = func() []byte {
+	return []byte(globalSecret)
+}
+
 // echo -n 'BODY' | openssl dgst -sha1 -hmac KEY
 func TestValidatePayload(t *testing.T) {
```
Katharine: add a test that still uses the old format, too.

prow/github/hmac.go
```diff
@@ -46,3 +79,40 @@ func PayloadSignature(payload []byte, key []byte) string {
...
+
+// extractHmacs returns all *valid* HMAC tokens for given repository/organization. It iterates over full tree as necessary (repo->org->global)
+func extractHmacs(repo string, tokenGenerator func() []byte) ([][]byte, error) {
+	t := tokenGenerator()
+	repoToTokenMap := map[string]hmacsForRepo{}
+
+	if err := yaml.Unmarshal(t, &repoToTokenMap); err != nil {
+		//NOTE : To keep backward compatibility, we are going to assume that in case of error, whole while is a single line hmac token.
+		//TODO : Once this code has been released and file has been moved to new format, we should delete this code and return error.
+		logrus.WithError(err).Error("Couldn't unmarshal the hmac secret as hierarchical file. Parsing as single token format.")
```
Katharine: This should be either reduced to info level, or provide immediately actionable information.

prow/github/hmac.go
```diff
@@ -46,3 +79,40 @@ func PayloadSignature(payload []byte, key []byte) string {
 	sum := mac.Sum(nil)
 	return "sha1=" + hex.EncodeToString(sum)
 }
+
+// extractHmacs returns all *valid* HMAC tokens for given repository/organization. It iterates over full tree as necessary (repo->org->global)
```
Katharine: The second part of this comment implies that it will match at any level, which isn't true. Also, considered hard wrapping to maintain reasonable line lengths.

```suggestion
// extractHmacs returns all *valid* HMAC tokens for given repository/organization.
// It considers only the tokens at the most specific level configured for the given repo.
```

prow/github/hmac.go
```diff
@@ -46,3 +79,40 @@ func PayloadSignature(payload []byte, key []byte) string {
...
+func extractHmacs(repo string, tokenGenerator func() []byte) ([][]byte, error) {
+	t := tokenGenerator()
+	repoToTokenMap := map[string]hmacsForRepo{}
+
+	if err := yaml.Unmarshal(t, &repoToTokenMap); err != nil {
+		//NOTE : To keep backward compatibility, we are going to assume that in case of error, whole while is a single line hmac token.
```
Katharine: Nit: I don't think "NOTE" is useful here. Also, s/while/file

```suggestion
		// To keep backward compatibility, we are going to assume that in case of error
		// the whole file is a single line hmac token.
```

prow/github/hmac.go
agarg2008: I guess 'warning' should be an appropriate level. Also we don't expect this to be status quo for long because this would be removed as part of T+1 deployment.

prow/github/hmac.go
agarg2008: It says "as necessary" .. I know its not super clear. If you have a verbiage in mind do share.

Katharine: I suggested verbiage above.

prow/github/hmac.go
Katharine: There are many prow deployments, most of which are not controlled by us. When changing things, we must provide for a substantial deprecation period with clear direction as to what must change.

(I would also argue we should deprecate things less than we do, since the benefit is usually minimal. In this case, some benefit does exist.)

Katharine: Given the context, a warning would still need actionable advice on resolving.

agarg2008: fair enough.

PR #16517, opened 2020-02-27
First pass migrating off of kubernetes_e2e for kops_gce

geojaz:
Big picture: I want to get GCE tests enabled for PRs. In order to do this, there's a need to shuffle some things around and probably move the kops tests out of the cloud-provider section (perhaps into cluster-lifecycle). 
When I started to dig in, i discovered that folks want us to stop using kube_e2e.py. This is an attempt to migrate to kubetest on this job, just to see how it works and get our story straight.

### DIFF
--- a/config/jobs/kubernetes/sig-cloud-provider/aws/kops/kops-periodics-gce.yaml
+++ b/config/jobs/kubernetes/sig-cloud-provider/aws/kops/kops-periodics-gce.yaml
@@ -5,15 +5,15 @@ periodics:
   labels:
     preset-service-account: "true"
     preset-k8s-ssh: "true"
-  decorate: true
-  decoration_config:
-    timeout: 140m
   spec:
     containers:
-    - command:
-      - runner.sh
-      - /workspace/scenarios/kubernetes_e2e.py
+    - image: gcr.io/k8s-testimages/kubekins-e2e:v20200226-84852b3-master
       args:
+      - --bare
+      - --timeout=120
+      - --scenario=kubernetes_e2e
+      - --
+      - --check-leaked-resources
       - --cluster=e2e-kops-gce-latest.k8s.local
       - --deployment=kops
       - --env=KOPS_PUBLISH_GREEN_PATH=gs://kops-ci/bin/latest-ci-gce-green.txt
@@ -24,8 +24,6 @@ periodics:
       - --kops-zones=us-central1-c
       - --provider=gce
       - --test_args=--ginkgo.skip=\[Slow\]|\[Serial\]|\[Disruptive\]|\[Flaky\]|\[Feature:.+\]|\[HPA\]|Dashboard|Services.*functioning.*NodePort
-      - --timeout=120m
-      image: gcr.io/k8s-testimages/kubekins-e2e:v20200226-84852b3-master
   annotations:
     testgrid-dashboards: google-kops-gce, sig-cluster-lifecycle-kops
     testgrid-tab-name: kops-gce-latest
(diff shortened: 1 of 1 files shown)


hakman:
Very nice!
/cc @rifelpet

hakman:
Doesn't seem to work. Maybe the decorate section break it or it's missing something like this:
https://github.com/kubernetes/test-infra/blob/5112784dd8468af850a47b1aee6c84a24734b0f5/config/jobs/kubernetes/sig-api-machinery/sig-api-machinery-config.yaml#L43-L44

geojaz:
@hakman Cool, thanks! i'll iterate on this locally and push something back. For some reason, i was hoping that by pushing code prow would test the job for me :) Do you know if that's possible?

hakman:
I don't think there is such a feature :). It will take a bit of trial and error to get things running in general. I think @rifelpet has worked the most on these jobs and may have some ideas.

If you need a second pair of eyes or a quick LGTM, just CC me in the PR.

rifelpet:
Get ready for a slew of links 😄 

I _think_ we still want to use [runner.sh](https://github.com/kubernetes/test-infra/blob/master/images/bootstrap/runner.sh). It sets up DinD and Bazel remote cache. It is the entrypoint for the image (kubekins-e2e [is FROM](https://github.com/kubernetes/test-infra/blob/da16e1b1780da64f9802c3538ccec06332e259e2/images/kubekins-e2e/Dockerfile#L18) the bootstrap image which sets [it as the entrypoint](https://github.com/kubernetes/test-infra/blob/da16e1b1780da64f9802c3538ccec06332e259e2/images/bootstrap/Dockerfile#L133)). A while back I migrated our jobs to use Pod Utilities ([PR](https://github.com/kubernetes/test-infra/pull/15762/files#diff-4b35fa989c42aa2acd6ef9b9a18c6048L497)) which according to their [docs](https://github.com/kubernetes/test-infra/blob/master/prow/pod-utilities.md#how-to-configure) say you must specify a command instead of using the image's entrypoint. The migration to pod utilities was to get off the deprecated [bootstrap.py](https://github.com/kubernetes/test-infra/blob/da16e1b1780da64f9802c3538ccec06332e259e2/jenkins/bootstrap.py#L1059) (more info [here](https://github.com/kubernetes/test-infra/blob/master/prow/pod-utilities.md#why-use-pod-utilities)). I think your `--scenario` argument that you've added [uses bootstrap.py](https://github.com/kubernetes/test-infra/blob/da16e1b1780da64f9802c3538ccec06332e259e2/jenkins/bootstrap.py#L1181) and still uses `kubernetes_e2e.py`: it [shells out](https://github.com/kubernetes/test-infra/blob/da16e1b1780da64f9802c3538ccec06332e259e2/jenkins/bootstrap.py#L908) to [this file](https://github.com/kubernetes/test-infra/blob/master/scenarios/kubernetes_e2e.py).

I think a nice goal would be towards moving all logic to kubetest (you can see that effort documented [here](https://github.com/kubernetes/test-infra/blob/master/images/kubekins-e2e/kops-e2e-runner.sh#L24). I migrated all of our periodic jobs off of that script but our presubmits still require some of its logic). All the args in the job after `--` are arguments for kubetest, so if we can add any missing logic from `kubernetes_e2e.py` and `kops-e2e-runner.sh` into kubetest and then have the job's command just be `kubetest` I think that would be ideal and help get rid of all the layers. I've tried to find a "model" prow job that we could strive towards but its not clear to me what the current recommended way of doing this stuff is. Perhaps we should ask someone in #sig-testing for assistance here. All i know is a bunch of stuff is deprecated 😄 and I've made progress towards getting off of it.


My brain is fried just from writing that but if you have any questions let me know.

rifelpet:
Ah i missed your followup PR #16520, ill add comments there.

PR #16520, opened 2020-02-27
Trying to fix the kops gce e2e

geojaz:
@hakman I'm going to bed. will continue to iterate in the morning....

### DIFF
--- a/config/jobs/kubernetes/sig-cloud-provider/aws/kops/kops-periodics-gce.yaml
+++ b/config/jobs/kubernetes/sig-cloud-provider/aws/kops/kops-periodics-gce.yaml
@@ -8,7 +8,7 @@ periodics:
   spec:
     containers:
     - image: gcr.io/k8s-testimages/kubekins-e2e:v20200226-84852b3-master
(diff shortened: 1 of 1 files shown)


hakman:
Good night :)
/lgtm

rifelpet:
We should be able to fix the google service account issue with the work I'm doing to migrate our jobs to use workload identity. See #16514 and the chain of PRs and issues it links to. Once that is merged we could add `serviceAccountName` to this job, remove the `preset-service-account` label, and it should get the credentials as expected. That should address [this warning](https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/e2e-kops-gce-latest/1233026126952009731#1:build-log.txt%3A2) but I'm not familiar enough with google cloud to know if it will fix the bad request error.

geojaz:
@rifelpet was this the confirmation that we were looking for? https://github.com/kubernetes/test-infra/pull/16514#issuecomment-592189742

rifelpet:
@geojaz yes that PR should be good to merge now

PR #16524, opened 2020-02-27
Update controller-runtime to v0.5.0

alvaroaleman:

PR #16529, opened 2020-02-27
Boskos: Retry on conflict

alvaroaleman:
This PR adds retrying on conflicts to the boskos handlers, in order to resolve the occasional `the object has been modified; please apply your changes to the latest version and try again` errors we are seeing and that will presumably vastly increase once we remove the global lock.

Fixes  https://github.com/openshift/release/issues/7183

/assign @ixdy @bbguimaraes

### DIFF
--- a/boskos/ranch/ranch.go
+++ b/boskos/ranch/ranch.go
@@ -24,7 +24,11 @@ import (
 	"time"
 
 	"github.com/sirupsen/logrus"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/util/retry"
 
 	"k8s.io/test-infra/boskos/common"
 	"k8s.io/test-infra/boskos/crds"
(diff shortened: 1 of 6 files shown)


boskos/handlers/handlers.go
```diff
@@ -390,3 +410,35 @@ func handleMetric(r *ranch.Ranch) http.HandlerFunc {
 		res.Write(js)
 	}
 }
+
+// retryOnConflict is a copy of https://godoc.org/k8s.io/client-go/util/retry#RetryOnConflict
```
ixdy: it looks like this is really a copy of `retry.OnError` except that you curried the `retriable` parameter.

Couldn't we instead do this?
```golang
func retryOnConflict(backoff wait.Backoff, fn func() error) error {
	return retry.OnError(backoff, isConflict, fn)
}
```

Or am I missing something?

alvaroaleman: yeah, `retry.RetryOnError` is not available in the client-go version we currently use. I am working on another PR that will also bump the kube dependencies, once that is in I could change this to use `retry.OnError`: https://github.com/kubernetes/test-infra/pull/16524

alvaroaleman:
> this somewhat feels weird to me, like the abstraction is leaking - why have the handlers do retries rather than putting that into the ranch implementation? (this is arguably easier, probably - the ranch functions are very long.)

That is a very good point, updated the PR accordingly. Good you caught this, because we do the locking within ranch, so the retying on handler level would have meant that we might have to acquire the lock multiple times for a single request :grimacing: 

I also added retrying to `SyncResources`, which required me to adjust the error propagation a bit so we can still properly recognize conflicts.

boskos/ranch/ranch.go
```diff
@@ -463,3 +479,43 @@ func (r *Ranch) AllMetrics() ([]common.Metric, error) {
 func newResourceFromNewDynamicResourceLifeCycle(name string, dlrc *crds.DRLCObject, now time.Time) *crds.ResourceObject {
 	return crds.NewResource(name, dlrc.Name, dlrc.Spec.InitialState, "", now)
 }
+
+// retryOnConflict is a copy of https://godoc.org/k8s.io/client-go/util/retry#RetryOnConflict
+// that only differs in the isConflict check, we use a variant that supports wrapped errors.
+// TODO: Simplify this by using retry.OnError once we have a sufficiently new client-go dependency
```
stevekuznetsov: Do we want a tracking issue on this

boskos/handlers/handlers.go
ixdy: are you working on a follow-up PR now that https://github.com/kubernetes/test-infra/pull/16524 is in?

ixdy: to answer my own question: https://github.com/kubernetes/test-infra/pull/16698

PR #16536, opened 2020-02-27
this is fine

BenTheElder:
/cc @krzyzacy @dims @ixdy


ixdy:
should we open an issue to properly fix this (i.e. by recognizing which projects are presubmit based on whether the job is presubmit)?

BenTheElder:
yep, off to a meeting, if you haven't filed one I will after

PR #16301, opened 2020-02-14
Set up benchmarkjunit job for structured-merge-diff

julianvmodesto:
It would be great to use the benchmarkjunit output to track the scalability work for structured-merge-diff.

- [pkg/benchmarkjunit](https://github.com/kubernetes/test-infra/tree/master/pkg/benchmarkjunit)
- [TestGrid demo](https://testgrid.k8s.io/sig-testing-canaries#benchmark-demo&width=5&graph-metrics=avg%20op%20duration%20(ns%2Fop)&graph-metrics=test-duration-minutes&graph-metrics=op%20count&graph-metrics=MB%2Fs&graph-metrics=alloced%20B%2Fop&graph-metrics=allocs%2Fop)

PR #16539, opened 2020-02-28
Create a gencred image and a ProwJob to publish it automatically.

cjwagner:
Running gencred currently requires cloning the entire repo and building from source with the correct version of bazel. A container image will be faster and require fewer dependencies.
Based on https://github.com/kubernetes/test-infra/pull/16301
/assign @clarketm 
/cc @Katharine

### DIFF
--- a/gencred/cloudbuild.yaml
+++ b/gencred/cloudbuild.yaml
@@ -1,0 +1,40 @@
+# Copyright 2020 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+steps:
+  - name: golang:$_GO_VERSION
+    args:
+    - go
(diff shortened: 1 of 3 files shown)


gencred/cloudbuild.yaml
```diff
@@ -0,0 +1,40 @@
...
+    env:
+    - CGO_ENABLED=0
+    - GOOS=linux
+    - GOARCH=amd64
+  - name: gcr.io/cloud-builders/docker
+    args:
+    - build
+    - --tag=gcr.io/$PROJECT_ID/gencred:$_GIT_TAG
+    - --tag=gcr.io/$PROJECT_ID/gencred:latest
+    - ./pkg/gencred
```
clarketm: This should be the relative path to the directory with the `gencred` `Dockerfile`?
```suggestion
    - ./gencred
```

gencred/cloudbuild.yaml
```diff
@@ -0,0 +1,40 @@
...
+steps:
+  - name: golang:$_GO_VERSION
+    args:
+    - go
+    - build
+    - -mod=readonly
+    - -a
+    - -installsuffix=cgo
+    - -o=./gencred/gencred
+    - ./gencred/cmd/gencred
```
clarketm: This should be the relative path to the top-level `gencred` `main.go`?
```suggestion
    - ./gencred
```

gencred/cloudbuild.yaml
cjwagner: Copy paste oversight.
Fixed.

gencred/cloudbuild.yaml
cjwagner: Yes, I didn't see that.
Fixed.

PR #16543, opened 2020-02-28
Consolidates kops prowjob specs from cloud-provider=>kubernetes/kops

geojaz:
kops e2es were split into two locations: `sig-cloud-provider/aws` and `kubernetes/kops`. Since we are starting to focus on other clouds, I propose to put all the tests in `kubernetes/kops`, which reflects the github org. 

NB: This is a first step, I'd like to break out tests by cloud, but one step at a time.


geojaz:
This is a first step in further consolidation/migration of kops e2es

rifelpet:
+1 from me!

neolit123:
this seems like a valid move. +1

one Q: should the file:
> sig-cluster-lifecycle/kops/sig-aws-credentials.yaml

be renamed to something like `credentials.yaml`?

geojaz:
It's likely that we will rename or deprecate that file as we move into the next era in kops e2e, but I'm fine with leaving it as is for today.

geojaz:
/assign @rifelpet @robinpercy 
I think now that I changed the target of the mv, we can approve our own PR :)

rifelpet:
/lgtm
👍

PR #16547, opened 2020-02-28
E2E-Topology Manager - Update image file

vpickard:
Use the same image file as used for CPU Manager
E2E tests.

Signed-off-by: vpickard <vpickard@redhat.com>


derekwaynecarr:
i lack rights at this level.

@dchen1107 can you approve?

lets track fixing rights here at this level in our retro.

/assign @dchen1107

PR #16552, opened 2020-02-28
Don't need all the kubetest lifecycle args with kubernetes_e2e

geojaz:
This is continued iteration. Should be almost there.


geojaz:
/assign @rifelpet  
:)

PR #16560, opened 2020-03-01
GA-only kind conformance test job

liggitt:
Enables optional, manually-triggered presubmit to run kind conformance with almost all beta APIs and features turned off.

Depends on https://github.com/kubernetes-sigs/kind/pull/1378

/hold
/cc @BenTheElder

### DIFF
--- a/config/jobs/kubernetes-sigs/kind/kind-presubmits.yaml
+++ b/config/jobs/kubernetes-sigs/kind/kind-presubmits.yaml
@@ -235,6 +235,47 @@ presubmits:
             memory: "9000Mi"
             # during the tests more like 3-20m is used
             cpu: 2000m
+  # conformance test against kubernetes release-1.16 branch with `kind`, skipping
+  # serial tests so it runs in ~20m
+  - name: pull-kind-conformance-parallel-1-16
+    labels:
+      preset-service-account: "true"
+      preset-bazel-scratch-dir: "true"
+      preset-bazel-remote-cache-enabled: "true"
+      preset-dind-enabled: "true"
+      preset-kind-volume-mounts: "true"
+    always_run: true
+    spec:
+      containers:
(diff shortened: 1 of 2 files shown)


config/jobs/kubernetes/sig-testing/conformance-e2e.yaml
```diff
@@ -47,6 +47,55 @@ presubmits:
       testgrid-dashboards: sig-testing-misc
       description: Runs conformance tests using kind and the conformance image
 
+  # conformance test using image against kubernetes master branch with `kind`
+  - name: pull-kubernetes-conformance-kind-ga-only
```
liggitt: copied `pull-kubernetes-conformance-image-test` as a starting point, not sure if there's a better one to start from

aojea:
Possible conflict with https://github.com/kubernetes-sigs/kind/pull/1269

config/jobs/kubernetes/sig-testing/conformance-e2e.yaml
aojea: I think that the recommendations is to move to podutils 
https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/sig-testing/kubernetes-kind-presubmits.yaml#L222-L268

liggitt: ah, ok, I was trying to set up a conformance run so I based it on the existing conformance one.

liggitt:
added `testgrid-dashboards: sig-testing-kind`

liggitt:
will test this out manually, and once I verify it works as expected, will add a periodic and a run_if_changed on the conformance folder to the presubmit

BenTheElder:
https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/46662/pull-kubernetes-conformance-kind-ga-only/1234288354271432704 🎉

PR #16565, opened 2020-03-02
test: change some test regions due to quota issue

andyzhangx:
hit quota issue again on `westus2` region:

```
2020/03/02 07:35:27 main.go:316: Something went wrong: starting e2e cluster: error creating cluster: cannot deploy: cannot create deployment: resources.DeploymentsClient#CreateOrUpdate: Failure sending request: StatusCode=400 -- Original Error: Code="InvalidTemplateDeployment" Message="The template deployment 'kubetest-c5826cf8-5c52-11ea-a109-16b0cfbd1c11' is not valid according to the validation procedure. The tracking id is 'd7dae513-51ac-4ba0-aead-f17588db0e1e'. See inner errors for details." Details=[{"code":"ResourceCountExceedsLimitDueToTemplate","details":[],"message":"Subscription 0e46bd28-a80f-4d3a-8200-d9eb8d80cb2e has a quota of 20 for resources of type PublicIpAddress with sku Basic. Subscription currently has 20 resources and the template contains 1 new resources of the this type which exceeds the quota. Please contact support to increase the quota for resource type PublicIpAddress"}]
```

/assign @feiskyer

### DIFF
--- a/config/jobs/kubernetes-sigs/azuredisk-csi-driver/azuredisk-csi-driver-config.yaml
+++ b/config/jobs/kubernetes-sigs/azuredisk-csi-driver/azuredisk-csi-driver-config.yaml
@@ -233,7 +233,7 @@ presubmits:
         - --aksengine-creds=$(AZURE_CREDENTIALS)
         - --aksengine-orchestratorRelease=1.18
         - --aksengine-deploy-custom-k8s
-        - --aksengine-location=westus2
+        - --aksengine-location=eastus
         - --aksengine-public-key=$(AZURE_SSH_PUBLIC_KEY_FILE)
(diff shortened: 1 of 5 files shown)


chewong:
Let's revert this PR since https://testgrid.k8s.io/provider-azure-presubmit#pr-k8s-azure-disk-e2e-vmss-master and https://testgrid.k8s.io/provider-azure-presubmit#pr-k8s-azure-file-e2e-master are experiencing quota issue. As shown in the log, there are only 10 cores available in eastus, which is definitely not enough (westus2 & eastus2 have 100 cores respectively). I noticed that there were a lot of unused public IPs in the subscription, which could the cause of the issue. I've manually deleted them so you don't have to worry about having the same issue again.

PR #16577, opened 2020-03-03
Improve bazel rbe cache hit rate

fejta:
/assign @BenTheElder 

Not entirely sure what is up with these two targets, so just run them locally for now.

### DIFF
--- a/prow/interrupts/BUILD.bazel
+++ b/prow/interrupts/BUILD.bazel
@@ -26,4 +26,5 @@ go_test(
     name = "go_default_test",
     srcs = ["interrupts_test.go"],
     embed = [":go_default_library"],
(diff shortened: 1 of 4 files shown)


prow/interrupts/BUILD.bazel
```diff
@@ -26,4 +26,5 @@ go_test(
     name = "go_default_test",
     srcs = ["interrupts_test.go"],
     embed = [":go_default_library"],
+    tags = ["local"],  # TODO(fejta): non-hermetic tests fails on RBE
```
BenTheElder: is there an issue for this?

BenTheElder:
/lgtm
seems like the tests only being run locally should have a tracking issue to ensure they don't stay that way?

fejta:
Maybe they are just flaky? I'm going to see if I can turn it off

PR #16585, opened 2020-03-03
Ignore delete action for label handler

### LINKED ISSUE #16579: Delete label comments may produce extra bot comments
**What happened**:

If I add a comment like `/sig abc` which does not exist. The bot will reply something like `The label(s) sig/abc cannot be applied, because the repository doesn't have them`. Then if I delete the wrong comment, the bot will reply the same message again.

When I delete the wrong comment, nothing should happen.

1. comment in an issue/pr like `/sig abc`
1. delete the comment

gavinfish:
Fix #16579.

### DIFF
--- a/prow/plugins/label/label_test.go
+++ b/prow/plugins/label/label_test.go
@@ -59,6 +59,7 @@ func TestLabel(t *testing.T) {
 		repoLabels            []string
 		issueLabels           []string
 		expectedCommentText   string
+		action                github.GenericCommentEventAction
 	}
 	testcases := []testCase{
 		{
@@ -69,6 +70,7 @@ func TestLabel(t *testing.T) {
 			repoLabels:            []string{},
 			issueLabels:           []string{},
 			commenter:             orgMember,
+			action:                github.GenericCommentActionCreated,
 		},
 		{
 			name:                  "Empty Area",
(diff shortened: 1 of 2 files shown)


prow/plugins/label/label_test.go
```diff
@@ -505,6 +551,29 @@ func TestLabel(t *testing.T) {
 			commenter:             orgMember,
 			expectedBotComment:    true,
 			expectedCommentText:   "The label(s) `/remove-label orchestrator/jar` cannot be applied. These labels are supported: `orchestrator/foo, orchestrator/bar`",
+			action:                github.GenericCommentActionCreated,
+		},
+		{
+			name:                  "Delete adding Labels is Case Insensitive",
```
cblecker:
```suggestion
			name:                  "Don't comment when deleting label addition",
```

prow/plugins/label/label_test.go
```diff
@@ -505,6 +551,29 @@ func TestLabel(t *testing.T) {
...
+			repoLabels:            []string{"area/infra", "priority/critical", labels.Bug},
+			issueLabels:           []string{},
+			expectedNewLabels:     []string{},
+			expectedRemovedLabels: []string{},
+			commenter:             orgMember,
+			expectedBotComment:    false,
+			action:                github.GenericCommentActionDeleted,
+		},
+		{
+			name:                  "Delete remove Committee Label",
```
cblecker:
```suggestion
			name:                  "Don't comment when deleting label removal",
```

PR #16587, opened 2020-03-03
Add Baiducloud conformance test into testgrid

tizhou86:
Hi all, we'd like to add baiducloud conformance test into testgrid, I've just pushed our test result into the gcs bucket, if there's something wrong, please let me know, thanks!

### DIFF
--- a/config/testgrids/conformance/conformance-all.yaml
+++ b/config/testgrids/conformance/conformance-all.yaml
@@ -7,6 +7,7 @@ dashboard_groups:
     - conformance-kind
     - conformance-cloud-provider-openstack
     - conformance-cloud-provider-huaweicloud
+    - conformance-cloud-provider-baiducloud
     - conformance-alibaba-cloud-provider
     - conformance-cloud-provider-vsphere
     - conformance-vsphere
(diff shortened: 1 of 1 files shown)


tizhou86:
#16587

tizhou86:
Hi @BenTheElder, any comments on this PR? Thanks!

config/testgrids/conformance/conformance-all.yaml
```diff
@@ -219,6 +220,12 @@ dashboards:
       description: Runs conformance tests using kubetest against kubernetes v1.17 with cloud-provider-huaweicloud
       test_group_name: cloud-provider-huaweicloud-e2e-conformance-release-v1.17
 
+- name: conformance-cloud-provider-baiducloud
+  dashboard_tab:
+    - name: Baidu Cloud Provider, v1.13
```
BenTheElder: I'm not going to block this right now as I don't think we have policy, but 1.13 is way out of support upstream and cannot be certified for conformance currently

BenTheElder: we were really mostly hoping for people to publish tests against the latest kubernetes code when we set up the program for people to publish conformance results to testgrid :disappointed:

BenTheElder:
going to let @spiffxp decide this

spiffxp:
/approve
/lgtm
I won't hold this up for now, because there are already many other tabs for kubernetes <=1.13.  But I agree we may consider a policy of pruning off older releases, so it would be preferable to upload results for something more current

PR #16594, opened 2020-03-03
prow: kubernetes/k8s.io: add blockade regexps

listx:
These changes add some blockades for sensitive files containing metadata for the promoter.

Related discussion: https://github.com/kubernetes/k8s.io/pull/623#discussion_r387265898

/cc @justaugustus @thockin

### DIFF
--- a/config/prow/plugins.yaml
+++ b/config/prow/plugins.yaml
@@ -170,6 +170,11 @@ blockades:
   blockregexps:
   - 'manifests/.*/promoter-manifest\.yaml$'
   explanation: "Changes to the metadata pose a security risk. This PR must be explicitly approved by k8s.io repo admins."
+- repos:
+  - kubernetes/k8s.io
+  blockregexps:
+  - 'images/legacy-backfill/images\.yaml$'
+  explanation: "Changes to the backfill manifests are forbidden, as they are effectively frozen. This PR must be explicitly approved by k8s.io repo admins."
 
 blunderbuss:
   max_request_count: 2
--- a/config/prow/plugins.yaml
+++ b/config/prow/plugins.yaml
@@ -165,6 +165,11 @@ blockades:
   blockregexps:
   - ^anago|branchff|build\/|changelog-update|compile-release-tools|debian\/|find_green_build|gcb\/|gcbmgr|lib\/|prin|push-build.sh|release-notify|relnotes|rpm\/
   explanation: "Changes to certain release tools can affect our ability to test, build, and release Kubernetes. This PR must be explicitly approved by SIG Release repo admins."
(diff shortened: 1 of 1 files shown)


listx:
I'm seeing an HTTP 401 error for the `pull-test-infra-bazel` test. @cjwagner help?

cjwagner:
Looks like a network flake. Unfortunately those are pretty common thanks to all the stuff bazel needlessly DLs.
/retest

listx:
> Looks like a network flake. Unfortunately those are pretty common thanks to all the stuff bazel needlessly DLs.
> /retest

Ack, thanks Cole!

listx:
> @listx -- I _think_ by including the `manifests/` directory, we're going to end up forcing admins to manually merge PRs.

I thought we **do** explicitly want manual merges of PRs for anything that touches any */promoter-manifest.yaml ?

cblecker:
The restriction is on a directory boundary, so lets use `no_parent_owners` if we need to be explicit.

Manually removing a `do-not-merge` label is an anti-pattern I wish Stephen hadn't twisted my arm to make happen on k8s.io/release.

/close

listx:
> Manually removing a `do-not-merge` label is an anti-pattern I wish Stephen hadn't twisted my arm to make happen on k8s.io/release.

We want to mark all PRs as "do-not-merge" if they touch promoter manifest files as a default stance, *even if* the OWNERS file is restricted to a small subset. The goal is to force maintainers to think twice about changing promoter manifest metadata once they are merged in (for the first time).

I can see how it would be very very annoying for maintainers to have to remove this label every time someone creates a *brand new* promoter manifest file for the first time, but that is a first-time cost that makes sense to me.

Thoughts?

cblecker:
It's still an anti-pattern. Blockade is for "this folder should never ever change".

riking:
The legacy backfill folder and manifest is a legitimate blockade target, though.

listx:
> Blockade is for "this folder should never ever change".

This is what has to be enforced, for all folders under `^k8s.gcr.io/manifests`. They should never ever change after they are first merged in.

Maybe "blockade" is not a great plugin, but for the use case here we want to be EXTRA careful when anyone modifies anything under the `manifests` folder, because changing that carries with it security risks.

cblecker:
@listx And an owners file like https://github.com/kubernetes/k8s.io/blob/master/groups/OWNERS ensures that only dims, thockin, and myself can change that folder. If we have high security folders, let's just do that.

cblecker:
Oh, and for a legacy backfill folder, yes, I'd agree that subdirectory is a legitimate blockade target

PR #16595, opened 2020-03-03
Add periodic job for basic SCTP e2e test cases

janosi:
We would like to add a new periodic job to kubernetes/sig-network to test the basic SCTP related e2e test cases.
The enhancement doc: https://github.com/kubernetes/enhancements/blob/master/keps/sig-network/0015-20180614-SCTP-support.md#basic-tests

The PR for the e2e test cases: 
https://github.com/kubernetes/kubernetes/pull/88196

### DIFF
--- a/config/jobs/kubernetes/sig-network/sig-network-misc.yaml
+++ b/config/jobs/kubernetes/sig-network/sig-network-misc.yaml
@@ -536,7 +536,7 @@ periodics:
       - --timeout=300m
       image: gcr.io/k8s-testimages/kubekins-e2e:v20200401-d2349a1-master
 - interval: 12h
-  name: ci-basic-sctp-e2e
+  name: ci-kubernetes-e2e-gce-sctp
   labels:
     preset-service-account: "true"
     preset-k8s-ssh: "true"
@@ -551,14 +551,10 @@ periodics:
       - --env=KUBE_FEATURE_GATES=SCTPSupport=true
       - --env=ALLOW_PRIVILEGED=true
       - --env=NET_PLUGIN=kubenet
-      - --cluster=
-      - --deployment=gke
-      - --extract=ci/latest
-      - --gcp-cloud-sdk=gs://cloud-sdk-testing/ci/staging
+      - --gcp-master-image=gci
       - --gcp-node-image=gci
-      - --gcp-zone=us-west1-c
-      - --gke-environment=test
-      - --provider=gke
+      - --gcp-zone=us-west1-b
+      - --provider=gce
       - --test_args=--ginkgo.focus=\[Feature:SCTP\]
       - --timeout=300m
       image: gcr.io/k8s-testimages/kubekins-e2e:v20200401-d2349a1-master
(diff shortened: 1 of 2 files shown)


aojea:
I'm not an expert on these kind of jobs but the diff against this other jobs https://github.com/kubernetes/test-infra/blob/325d4e4808847c0f1b01c4d31be9b10f51f7d39b/config/jobs/kubernetes/sig-network/sig-network-misc.yaml#L416
is significant.
Also, is it worth adding a new file or just adding the job to the linked file previously?

janosi:
@aojea Yeah, this is my first test job here, so I hope I get some guidance on what I missed, or on what is not needed.

aojea:
> @aojea Yeah, this is my first test job here, so I hope I get some guidance on what I missed, or on what is not needed.

I would duplicate the test that I've added in my previous comment 
https://github.com/kubernetes/test-infra/pull/16595#issuecomment-594201314 in the same file, seems that you only have to change that filter

Adding one file only for these tests seems excessive, mainly because reading the KEP it seems there is no many of them and are not expected to grow

janosi:
@aojea OK, let's have it like that.

aojea:
I think that is better that your squash the commits to avoid the conflict error
I defer to previous approvers in the file the final review, I don't have experience on those gce jobs :sweat_smile: 
/assign @spiffxp @MrHohn

config/jobs/kubernetes/sig-network/sig-network-misc.yaml
```diff
@@ -495,3 +495,36 @@ periodics:
       - --test_args=--ginkgo.focus=\[sig-network\]\sServices --ginkgo.skip=\[Serial\]|\[Disruptive\]|\[Flaky\]|\[Feature:GCEAlphaFeature\] --minStartupPods=8
       - --timeout=300m
       image: gcr.io/k8s-testimages/kubekins-e2e:v20200306-3e08456-master
+- interval: 12h
+  name: ci-basic-sctp-e2e
```
spiffxp: The job names live in a global namespace so they need to be more informative
```suggestion
  name: ci-kubernetes-e2e-gci-gke-sctp
```

MrHohn:
> One nit, although I think you may want to reconsider using gke. Looking at the other jobs in https://testgrid.k8s.io/sig-network-gke none of them seem to be coming up. It may be worth trying to copy a GCE based job instead, such as the gce-lb-finalizer job above this

+1, running as a GCE job also gives you more visibility in case of debugging failure.

config/jobs/kubernetes/sig-network/sig-network-misc.yaml
janosi: Fixed. Now it has a GCE "namespaced" name, just like to lb-finalizer above this one.

config/jobs/kubernetes/sig-network/sig-network-misc.yaml
```diff
@@ -495,3 +495,32 @@ periodics:
...
+      - --env=NET_PLUGIN=kubenet
+      - --gcp-master-image=gci
+      - --gcp-node-image=gci
+      - --gcp-zone=us-west1-b
+      - --provider=gce
+      - --test_args=--ginkgo.focus=\[Feature:SCTP\]
+      - --timeout=300m
+      image: gcr.io/k8s-testimages/kubekins-e2e:v20200306-3e08456-master
+  annotations:
+    testgrid-dashboards: sig-network-sctp
```
spiffxp: You need to add this to https://github.com/kubernetes/test-infra/blob/master/config/testgrids/kubernetes/sig-network/config.yaml

janosi: Thank you for the guidance! Fixed now.

spiffxp:
/approve
Looks like this still has a merge conflict with config/jobs/kubernetes/sig-network/sig-network-misc.yaml

rebase and I'll /lgtm

janosi:
@spiffxp Rebased. Thank you!

PR #16600, opened 2020-03-04
Minor upgrade of mattn and jinzhu modules

fejta:
ref https://github.com/kubernetes/test-infra/pull/16599

* This includes a patch to the index names for velodrome's model (must now be unique).
* Also a patch to our nogo exclusion list as the path appears to have changed for this repo to something weird.


fejta:
The other PR merged all the other changes aside from the sqlite ones

fejta:
/retest

@stevekuznetsov FYI these prow/metrics tests are flaky

velodrome/sql/model.go
```diff
@@ -80,12 +80,12 @@ type Assignee struct {
 
 // Comment is either a pull-request comment or an issue comment.
 type Comment struct {
-	Repository       string `gorm:"primary_key;index:repo_created"`
+	Repository       string `gorm:"primary_key;index:repo_comment_created"`
```
Katharine: Is this safe without a schema migration of some form?

It seems like it probably won't matter as long as nothing pays too much attention…

fejta: Don't indexes just make querying the table faster? AKA you cannot access them directly.

fejta: And the code has some attractive sounding `db.AutoMigrate()` after the `Open()` call

https://github.com/kubernetes/test-infra/blob/3dbf7d4c04b94e9ef8951139fa9581ebfdd2c0ef/velodrome/sql/mysql.go#L60

Katharine: Correct — they shouldn't be accessed directly, but if some meta-code tries looking at the indexes it might get upset about the inconsistent state between declaration and reality.

That said it's probably fine and we can revert if not.

fejta: I wouldn't be surprised if previous versions created the index for Comment and then later replaced the index with one for IssueEvent (or visa versa).

fejta: (and we never realized because it just resulted in lower performance, which we didn't recognize)

Katharine: It certainly seems unlikely it's actually doing what it should be

PR #16612, opened 2020-03-04
updates to cip jobs

listx:
Instead of running once a day, this makes the job run every 4 hours.

/cc @thockin


listx:
Related context: https://github.com/kubernetes/k8s.io/pull/623#issuecomment-594832438

listx:
I'm also adding a commit here to update the promoter images for the jobs. This should allow us to re-merge the backfill PR https://github.com/kubernetes/k8s.io/pull/623.

PR #16613, opened 2020-03-04
pod-utils: allow setting resource requests and limits

stevekuznetsov:
In a well-formed k8s deployment is is important to be setting requests
and limits for resources used by workloads. While users can already set
these for the test container in a decorated ProwJob, it is not currently
possible to tune them for the decorated containers. For jobs with large
amounts of source to clone or artifacts to upload, the usage of these
containers can become large and it is important to allow tuning these
settings. As with all other settings, the normal rules of cascading
defaulting apply.

/assign @petr-muller @alvaroaleman


stevekuznetsov:
Looks like a flake?

```
 INFO: From Testing //prow/metrics:go_default_test:
==================== Test output for //prow/metrics:go_default_test:
--- FAIL: TestExposeMetrics (0.01s)
    metrics_test.go:30: failed getting metrics: Get http://127.0.0.1:9090/metrics: dial tcp 127.0.0.1:9090: connect: connection refused
FAIL
```

prow/apis/prowjobs/v1/types.go
```diff
@@ -294,6 +294,9 @@ type DecorationConfig struct {
 	// UtilityImages holds pull specs for utility container
 	// images used to decorate a PodSpec.
 	UtilityImages *UtilityImages `json:"utility_images,omitempty"`
+	// Resources holds resource requests and limits for utility
+	// containers used to decorate a PodSpec.
+	Resources *Resources `json:"resources,omitempty"`
```
alvaroaleman: Please mention that by default, we set none

stevekuznetsov:
/hold cancel

Will follow up with some test cases, no pre-existing ones for this.

PR #16622, opened 2020-03-05
Add resource requests for smd benchmark jobs

julianvmodesto:
As mentioned by @BenTheElder, benchmark jobs should set resource requests.

This change sets resource requests for the structured-merge-diff jobs.

For https://github.com/kubernetes/kubernetes/issues/86935.

cc @apelisse

### DIFF
--- a/config/jobs/kubernetes-sigs/structured-merge-diff/structured-merge-diff-ci.yaml
+++ b/config/jobs/kubernetes-sigs/structured-merge-diff/structured-merge-diff-ci.yaml
@@ -58,6 +58,13 @@ periodics:
       - --log-file=$(ARTIFACTS)/benchmark-log.txt
       - --output=$(ARTIFACTS)/junit_benchmarks.xml
       - ./...
+      resources:
+        requests:
+          cpu: "6000m"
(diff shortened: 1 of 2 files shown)


apelisse:
Does it make sense to use the same value for request and limit, so that we know exactly what we'll run with?

config/jobs/kubernetes-sigs/structured-merge-diff/structured-merge-diff-ci.yaml
```diff
@@ -58,6 +58,10 @@ periodics:
       - --log-file=$(ARTIFACTS)/benchmark-log.txt
       - --output=$(ARTIFACTS)/junit_benchmarks.xml
       - ./...
+      resources:
+        requests:
```
BenTheElder: i think it also needs to set limits?

julianvmodesto: Done!

BenTheElder:
ugh this really fell off the PR backlog
/lgtm
/approve
sorry

PR #16629, opened 2020-03-05
Migrate logexporter to use klog.

oxddr:
glog doesn't work well with pflag library (or any other flag replacemnt FWIW), see https://github.com/golang/glog/pull/13.

This will prevent annoying prefixes in every log line from logexporter, see: https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/88653/pull-kubernetes-e2e-gce-100-performance/1235224572588986368/artifacts/e2e-88653-95a39-minion-group-7s2j/logexporter-qxvpg.log

/assign @mm4tt


mm4tt:
/lgtm
/approve

Thanks!

PR #16632, opened 2020-03-05
Add an optional parallel ga-only conformance presubmit

liggitt:
Part of https://github.com/kubernetes/enhancements/blob/master/keps/sig-architecture/20191023-conformance-without-beta.md

Would have flagged https://github.com/kubernetes/kubernetes/pull/87759#issuecomment-595279933

Making this optional for now, we can consider making it required once we gain confidence with it. It doesn't include serial tests, but it does ensure clusters can start up successfully.

### DIFF
--- a/config/jobs/kubernetes/sig-testing/kubernetes-kind-presubmits.yaml
+++ b/config/jobs/kubernetes/sig-testing/kubernetes-kind-presubmits.yaml
@@ -470,3 +470,49 @@ presubmits:
       testgrid-alert-stale-results-hours: '24'
       testgrid-create-test-group: 'true'
       testgrid-dashboards: sig-testing-kind
+
+  - name: pull-kubernetes-conformance-kind-ga-only-parallel
+    optional: true
+    always_run: true
+    decorate: true
+    skip_branches:
+    - release-\d+\.\d+ # per-release settings
+    labels:
+      preset-bazel-scratch-dir: "true"
+      preset-bazel-remote-cache-enabled: "true"
+      preset-dind-enabled: "true"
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes/sig-testing/kubernetes-kind-presubmits.yaml
```diff
@@ -470,3 +470,49 @@ presubmits:
       testgrid-alert-stale-results-hours: '24'
       testgrid-create-test-group: 'true'
       testgrid-dashboards: sig-testing-kind
+
+  - name: pull-kubernetes-conformance-kind-ga-only-parallel
+    optional: true
+    always_run: true
+    decorate: true
+    skip_branches:
```
liggitt: Trying to decide whether to scope this with a run_if_changed, at least initially. There's not a super-clear subset of things we could limit this to

BenTheElder: run_if_changed is hard to get right, I generally don't recommend it.

liggitt:
will wait for https://github.com/kubernetes/kubernetes/issues/88855 to be resolved before letting this job loose on PRs on code freeze day (even though it's optional)

BenTheElder:
you can change it to skip_report: true so that it doesn't even show up on
github but start getting data

liggitt:
well... we know the data we would get... all clusters would fail to start

liggitt:
https://github.com/kubernetes/kubernetes/pull/88857 merged

/hold cancel

PR #16652, opened 2020-03-06
Push resultstore images

fejta:
Also push prow images when its dependencies change.


fejta:
Now with passing tests!

Issue #14009, opened 2019-08-21
resultstore-upload job is continually failing


spiffxp:
**What happened**:
https://testgrid.k8s.io/sig-testing-prow#resultstore-upload is continually failing

**What you expected to happen**:
for it not to fail

**Please provide links to example occurrences, if any**:
https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-test-infra-resultstore-upload/1164243772695908355

**Anything else we need to know?**:
It's not clear to me whether we rely on this and so I suggest at a minimum we stop running this 280 times a day

/assign @fejta

spiffxp:
Opened https://github.com/kubernetes/test-infra/pull/14010 to slow the job down

fejta:
From the logs:

```
time="2019-08-21T18:34:55Z" level=fatal msg="Failed transfer" error="transfer: 3 errors transferring 957 groups: transfer ci-kubernetes-e2e-gci-gce-coredns-nodecache in gs://kubernetes-jenkins/logs/ci-kubernetes-e2e-gci-gce-coredns-nodecache: gs://kubernetes-jenkins/logs/ci-kubernetes-e2e-gci-gce-coredns-nodecache/1164160204770643969/: download: suites: parse: not valid testsuites: expected element type <testsuites> but have <testsuite> nor testsuite: XML syntax error on line 346: unescaped < inside quoted string, transfer ci-kubernetes-e2e-gce-scale-correctness in gs://kubernetes-jenkins/logs/ci-kubernetes-e2e-gce-scale-correctness: gs://kubernetes-jenkins/logs/ci-kubernetes-e2e-gce-scale-correctness/1164145604574580738/: upload https://source.cloud.google.com/results/invocations/3930d827-cbd5-4d0b-ac37-569c31c10ab6/targets/test: create action: rpc error: code = InvalidArgument desc = Request contains an invalid argument., transfer ci-kubernetes-e2e-gce-scale-performance in gs://kubernetes-jenkins/logs/ci-kubernetes-e2e-gce-scale-performance: gs://kubernetes-jenkins/logs/ci-kubernetes-e2e-gce-scale-performance/1163858705188392960/: upload https://source.cloud.google.com/results/invocations/4cecdc0b-8121-4cc7-ba6d-309370d20965/targets/test: create action: rpc error: code = InvalidArgument desc = Request contains an invalid argument."
```

[...]

PR #16653, opened 2020-03-06
Set a ceiling on the number of extra files to attach.

fejta:
Otherwise the server will reject invocations describing large jobs (which have tens of thousands of files).

Also:
* rename the "deadline" timeout to "timeout" (https://golang.org/pkg/context/#WithDeadline)
* Stop checking for bucket writability when we explicitly pass this list.

ref https://github.com/kubernetes/test-infra/issues/14009

### DIFF
--- a/experiment/resultstore/convert.go
+++ b/experiment/resultstore/convert.go
@@ -118,7 +118,7 @@ func convertSuiteMeta(suiteMeta gcs.SuitesMeta) resultstore.Suite {
 }
 
 // Convert converts build metadata stored in gcp into the corresponding ResultStore Invocation, Target and Test.
-func convert(project, details string, url gcs.Path, result downloadResult) (resultstore.Invocation, resultstore.Target, resultstore.Test) {
+func convert(project, details string, url gcs.Path, result downloadResult, maxFiles int) (resultstore.Invocation, resultstore.Target, resultstore.Test) {
 	started := result.started
 	finished := result.finished
 	artifacts := result.artifactURLs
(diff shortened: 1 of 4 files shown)


experiment/resultstore/convert.go
```diff
@@ -175,13 +175,18 @@ func convert(project, details string, url gcs.Path, result downloadResult) (resu
 		artifacts[i] = "gs://" + bucket + "/" + a
 	}
 
+	var total int
 	for _, a := range artifacts { // add started.json, etc to the invocation artifact list.
+		if total >= maxFiles {
+			continue
```
michelle192837: Should we log somewhere when we hit this case?

experiment/resultstore/main.go
```diff
@@ -213,25 +249,34 @@ func run(opt options) error {
...
+type bucketChecker func(context.Context, string) bool
+
+func filterBuckets(parent context.Context, checkBucket bucketChecker, groups ...configpb.TestGroup) ([]configpb.TestGroup, error) {
 	buckets := map[string]bool{}
-	var path gcs.Path
-	for _, p := range bucketPaths {
-		if err := path.Set(p); err != nil {
-			return nil, fmt.Errorf("bad bucket: %w", err)
+	valid := func(path gcs.Path) bool {
+		buck := path.Bucket()
```
michelle192837: nit: 'b' or 'bucket'? 'buck' reads weird

experiment/resultstore/convert.go
fejta: Done

experiment/resultstore/main.go
fejta: Done

fejta:
/hold cancel
PTAL?

PR #16654, opened 2020-03-06
Enable NodeKiller on ClusterLoader presubmit

jprzychodzen:
/sig scalability
/assign @mm4tt

### DIFF
--- a/config/jobs/kubernetes/sig-scalability/sig-scalability-presubmit-jobs.yaml
+++ b/config/jobs/kubernetes/sig-scalability/sig-scalability-presubmit-jobs.yaml
@@ -342,6 +342,8 @@ presubmits:
         - --test-cmd-args=--testconfig=testing/density/config.yaml
         - --test-cmd-args=--testconfig=testing/load/config.yaml
         - --test-cmd-args=--testconfig=testing/access-tokens/config.yaml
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes/sig-scalability/sig-scalability-presubmit-jobs.yaml
```diff
@@ -288,6 +288,8 @@ presubmits:
         - --test-cmd-args=--testconfig=testing/load/config.yaml
         - --test-cmd-args=--testconfig=testing/access-tokens/config.yaml
         - --test-cmd-args=--testoverrides=./testing/experiments/enable_prometheus_api_responsiveness.yaml
+        - --test-cmd-args=--testoverrides=./testing/chaosmonkey/override.yaml
+        - --test-cmd-args=--testoverrides=./testing/chaosmonkey/ignore_node_killer_container_restarts_100.yaml
```
mm4tt: There is also perf-test presubmit, we should try to keep them in sync

jprzychodzen: Well, probably we event should start from `pull-perf-tests-clusterloader2` before adding nodekiller to other presubmits. Moved flags to that part of config.

PR #16658, opened 2020-03-06
gcsupload: limit maximum upload concurrency

stevekuznetsov:

/cc @fejta @cjwagner @smarterclayton

### DIFF
--- a/prow/gcsupload/options.go
+++ b/prow/gcsupload/options.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"flag"
 	"fmt"
+	"runtime"
 	"strings"
 
 	"github.com/GoogleCloudPlatform/testgrid/util/gcs"
@@ -114,6 +115,10 @@ func (o *Options) Complete(args []string) {
 		o.GCSConfiguration.MediaTypes[extension] = mediaType
 	}
 	o.mediaTypes = flagutil.NewStrings()
+
+	if o.GCSConfiguration.MaxConcurrency == 0 {
+		o.GCSConfiguration.MaxConcurrency = int64(4 * runtime.NumCPU())
+	}
 }
 
 // AddFlags adds flags to the FlagSet that populate
--- a/prow/pod-utils/gcs/upload.go
+++ b/prow/pod-utils/gcs/upload.go
@@ -26,6 +26,7 @@ import (
 
 	"cloud.google.com/go/storage"
 	"github.com/sirupsen/logrus"
+	"golang.org/x/sync/semaphore"
 	"google.golang.org/api/googleapi"
 
 	"k8s.io/test-infra/prow/errorutil"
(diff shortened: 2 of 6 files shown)


prow/pod-utils/gcs/upload.go
```diff
@@ -52,18 +53,23 @@ func LocalExport(exportDir string, uploadTargets map[string]UploadFunc) error {
...
 
-func upload(dtw destToWriter, uploadTargets map[string]UploadFunc) error {
+func upload(dtw destToWriter, uploadTargets map[string]UploadFunc, maxConcurrency int64) error {
 	errCh := make(chan error, len(uploadTargets))
 	group := &sync.WaitGroup{}
 	group.Add(len(uploadTargets))
+	sem := semaphore.NewWeighted(maxConcurrency)
 	for dest, upload := range uploadTargets {
 		log := logrus.WithField("dest", dest)
+		if err := sem.Acquire(context.Background(), 1); err != nil {
```
stevekuznetsov: Note: it's actually the `dtw(dest)` call we make in creating the closure that allocates, so we need to guard _that_

prow/gcsupload/options.go
```diff
@@ -63,6 +63,8 @@ type Options struct {
 	gcsPath gcs.Path
 }
 
+const defaultMaxConcurrency int64 = 50
```
fejta: https://github.com/GoogleCloudPlatform/testgrid/blob/e150868219d53c2e3cbc56310e5bb9dc8a3a4336/cmd/updater/main.go#L62 is what I've been using (cpus * 4)

stevekuznetsov: Ooh, I like that. Do you have any clue what it returns in a container in k8s under a CPU-limited cgroup?

fejta: No clue, but seems to have always provided me sufficiently fast results

prow/pod-utils/gcs/upload.go
```diff
@@ -52,18 +53,23 @@ func LocalExport(exportDir string, uploadTargets map[string]UploadFunc) error {
...
+			log.WithError(err).Error("Could not begin upload.")
```
smarterclayton: Will having an unpaired release break/crash?

stevekuznetsov: How could it be unpaired?

smarterclayton: If acquire fails here, you still call the go func, which will call release?

stevekuznetsov: Ah, eek

stevekuznetsov: Should only fail if the context is cancelled, which it cannot be, but yes

prow/pod-utils/gcs/upload.go
```diff
@@ -38,11 +39,11 @@ type destToWriter func(dest string) dataWriter
 // Upload uploads all of the data in the
 // uploadTargets map to GCS in parallel. The map is
 // keyed on GCS path under the bucket
-func Upload(bucket *storage.BucketHandle, uploadTargets map[string]UploadFunc) error {
+func Upload(bucket *storage.BucketHandle, uploadTargets map[string]UploadFunc, maxConcurrency int64) error {
 	dtw := func(dest string) dataWriter {
 		return gcsObjectWriter{bucket.Object(dest).NewWriter(context.Background())}
```
cjwagner: This is kind of orthogonal to this PR, but we should be using a context with a timeout here so that we don't try to upload a file forever, especially now that that could prevent other files from being uploaded as well.

PR #16659, opened 2020-03-06
add: fixes to aks kubetest deployer

alexeldeib:


feiskyer:
/hold
@alexeldeib could you confirm whether the kubekins-e2e image has included your last patch?

alexeldeib:
@feiskyer yes, that was actually the main blocker for validation. I've figured out a local workaround inside the kubekins image to validate against my branch which is what i've been able to do here to get validation using the actual prowjob config + kind cluster locally, azure credentials as secrets, etc. 

That said, Marwan and I were still struggling with one final piece of bash on the conformance jobs, so please leave the hold. 

re: building kubekins, I found out it's built immediately as a postsubmit https://github.com/kubernetes/test-infra/blob/c94d13d23adec691302314857301ebcf524d1084/config/jobs/kubernetes/test-infra/test-infra-trusted.yaml#L450-L475

feiskyer:
Because kubetest is also changed in the PR, you need spliting it into two:

1) change kubetest
2) add the test jobs when kubetest image has included the changes from 1)

alexeldeib:
@feiskyer removed the prowjobs, squashed, and both conformance + scalability are looking good. Would you mind reviewing + taking a look at https://github.com/kubernetes/kubernetes/pull/88477 which is done as well, barring test failures?

kubetest/aksengine.go
```diff
@@ -340,14 +339,19 @@ func newAKSEngine() (*aksEngineDeployer, error) {
 		}
 	}
 
+	outputDir, err := ioutil.TempDir(os.Getenv("HOME"), "tmp")
```
feiskyer: could we use os.TempDir() here?

alexeldeib: sure. changed

alexeldeib: actually, i'd prefer to keep this under a predictably named path so we can find the kubeconfig easily from other steps without access to this variable (I already have a need for this).

feiskyer:
please fix the conflicts.
/lgtm cancel

alexeldeib:
/hold

To account for https://github.com/kubernetes/test-infra/pull/16452#discussion_r390879570

alexeldeib:
going to close this, revert the revert, and merge those into one PR with the changes to the GCE condition.

PR #16660, opened 2020-03-06
Update resultstore exporter to v20200306-2c8702645

fejta:
/assign @michelle192837 

@Katharine  what will make this auto-bump?

ref https://github.com/kubernetes/test-infra/issues/14009


fejta:
Can I get a review?

Katharine:
> what will make this auto-bump?

It should be already, and has been [before](https://github.com/kubernetes/test-infra/pull/16609) — maybe you had bad timing with the bumps?

fejta:
Awesome thanks! Probably I'm just impatient.

PR #15803, opened 2020-01-07
Move gce-cos-master-default back to release-blocking

spiffxp:
I'd like to see this back in release-blocking. It's the CI equivalent
to the merge-blocking pull-kubernetes-e2e-gce job that all code has to
go through.

http://velodrome.k8s.io/dashboard/db/job-health-release-informing?orgId=1&var-rjob=ci-kubernetes-e2e-gci-gce&from=now-6M&to=now

Its metrics are back in line with the criteria for release-blocking jobs

/sig release
/sig testing
FYI @kubernetes/ci-signal

Issue #9363, opened 2018-09-12
Different set of blocking jobs on stable1, 2 and 3 releases


MaciekPytel:
In particular soak-gci-gce-1.X job is blocking for stable2, but is not blocking on stable1 and stable3. As a result soak job will be blocking for just a part of each release lifecycle. We should have a consistent set of blocking jobs for all patch releases for a single branch. I think it would be more correct if we just have a specific set of blocking jobs defined for each release, rather than have rolling stable1-3.
If that's not possible we should at least make sure the set of jobs is the same for all three (the downside of that is that it will be hard to change the set of blocking jobs).

cc: @krzyzacy @BenTheElder @kubernetes/sig-release-bugs

krzyzacy:
cc @spiffxp 

they are *mostly* same except for, say, 1.12 doesn't have kubeadm upgrade jobs set up

we can probably bring in a presubmit to enforce this

I'll remove the soak-stable2 job from release blocking dashboard.

spiffxp:
/area jobs
We absolutely should have the same jobs.  The only time I can see that differing is if we add new blocking jobs in the current release that couldn't work against older releases.  

A presubmit sounds good.

I haven't done a thorough audit, but last I checked we were also missing a scalability-related stable3 job.

krzyzacy:
/assign

I'll create a doc under sig-release to define a list of release-blocking jobs, and we can use that as a source of truth in the future.

spiffxp:
/milestone v1.13
FYI @cjwagner @jberkus

spiffxp:
The plan is to accomplish this before 2018-10-23 aka Enhancements Freeze of v1.13 release cycle:
- get release-1.y-blocking reconciled using release-1.12-blocking as the "authoritative" source
- get sig-release-master-blocking reconciled agains release-1.12-blocking
- reconcile means: same set of jobs, same naming scheme for testgrid tabs

jberkus:
I've done some comparison of master-blocking and 1.12-blocking.  Here's where they don't match.  Note that I'm using the actual job names below instead of the label you see in testgrid, because it's hard to figure out which job it is from the label:

**Tests that are in 1.12-blocking with no equivalent in master-blocking**:

* ci-kubernetes-e2e-kubeadm-gce-1-11-on-1-12
* periodic-kubernetes-bazel-build-1-12
* periodic-kubernetes-bazel-test-1-12

**Tests that are in master-blocking with no equivalent in 1.12-blocking**

[...]

PR #16662, opened 2020-03-06
Make gce-cos-k8sbeta-default release-blocking

spiffxp:
gce-cos-master-default is release-blocking, this should be too

trimmed testgridNumFailuresToAlert to match other release-blocking jobs

/sig release
FYI @kubernetes/ci-signal

Followup to https://github.com/kubernetes/test-infra/pull/15803


justaugustus:
Thanks Aaron!
/lgtm
/approve

spiffxp:
ref: https://github.com/kubernetes/test-infra/issues/9363

PR #16676, opened 2020-03-08
Add aojea to azure ipv6 alerts

aojea:
Just to keep an eye on the overall ipv6 health of kubernetes, adding myself to the job alert


justaugustus:
/hold
@aojea -- is there an alert group we can use instead of your personal address?

aojea:
> /hold
> @aojea -- is there an alert group we can use instead of your personal address?

don't know, how can I join one?
per example `kubernetes-sig-network-test-failures@googlegroups.com`

aojea:
/close
seems the goal is to have these jobs as release informers, so I'm sure failures will not be unnoticed
:smile:

PR #16678, opened 2020-03-09
Allow kops tests to run with not yet supported versions of kubernetes

johngmyers:
Add the KOPS_RUN_TOO_NEW_VERSION env var of kubernetes/kops#8700 in order to allow that PR to land.

/cc @rifelpet


rifelpet:
Sorry for the delay. I'm wondering if it would be easier to have this env var set for all kops e2e jobs automatically by setting it in kubetest like this:

https://github.com/kubernetes/test-infra/blob/02a6dffa9047369e2c1a7b18d9f53f70841d4325/kubetest/kops.go#L243-L256

johngmyers:
@rifelpet I was thinking we didn't want it set for the release-branch e2e tests.

rifelpet:
ah ok. yea that seems reasonable to me.

/lgtm
/approve

PR #16680, opened 2020-03-09
Turn on KFP backend prow test

jingzhang36:
Turn on the prow test.

Related issue:
https://github.com/kubeflow/pipelines/issues/2885

Tested this backend prow test at https://github.com/kubeflow/pipelines/pull/3240#issuecomment-596371819

/assign @jlewi


Bobgy:
@kunmingg @richardsliu Can one of you approve?
(Also shall we add us into OWNERS here?)

PR #16687, opened 2020-03-09
schedulerf_perf: constrain compute resources

ingvagabund:
We need to constraint compute resources so all the tests
finish approximatelly at the same time. More compute power
can increase scheduling throughput and make consequent results
incomparable.

All test containers run by scheduler_perf jenkins job set memory resource request to "1Gi". Not sure what's the right request for cpu so setting it to 4 cores for now to see how it will behave.

CC @ahg-g @liu-cong

### DIFF
--- a/config/jobs/kubernetes/sig-scalability/sig-scalability-periodic-jobs.yaml
+++ b/config/jobs/kubernetes/sig-scalability/sig-scalability-periodic-jobs.yaml
@@ -410,6 +410,15 @@ periodics:
       # even on very powerful machines
       - name: BENCHTIME
         value: 1ns
+      # We need to constraint compute resources so all the tests
+      # finish approximately at the same time. More compute power
+      # can increase scheduling throughput and make consequent results
+      # incomparable.
+      resources:
+        requests:
+          cpu: 4
+        limits:
+          cpu: 4
 
 - name: ci-benchmark-kube-dns-master
   interval: 2h


config/jobs/kubernetes/sig-scalability/sig-scalability-periodic-jobs.yaml
```diff
@@ -410,6 +410,14 @@ periodics:
...
       - name: BENCHTIME
         value: 1ns
+      # We need to constraint compute resources so all the tests
+      # finish approximately at the same time. More compute power
+      # can increase scheduling throughput and make consequent results
+      # incomparable.
+      resources:
+        requests:
+          memory: "1Gi"
+          cpu: 4
```
ahg-g: Lets not set the memory limit so that it doesn't get killed if the limit is exceeded. I don't think we set memory limits in the default deployment.

ahg-g: also, we need to set the limit for the cpu so that it gets throttled.

ingvagabund: It's resource requests, not limits. Though, I can remove it and have it set automatically.

ingvagabund: Done

ahg-g: Right, we should set the limit for the cpu as I argued in the second comment.

PR #16691, opened 2020-03-09
Fix call to 'git diff' in transfigure

munnerz:
Fixes the boolean logic in transfigure.sh that checks whether any changes have been made 😅 

ref https://kubernetes.slack.com/archives/C09QZ4DQB/p1583754596266200
/cc @clarketm @chases2

### DIFF
--- a/testgrid/cmd/transfigure/transfigure.sh
+++ b/testgrid/cmd/transfigure/transfigure.sh
@@ -69,7 +69,7 @@ main() {
     --output "${testgrid_dir}/${testgrid_subdir}/gen-config.yaml"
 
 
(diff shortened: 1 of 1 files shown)


chases2:
I don't think it's as simple as doing this. When I run this simple script:
```
if git diff --quiet ; then
  echo "No changes"
fi
```
... it returns "No changes" when there are uncommitted changes AND when there are committed changes. It just seems to be hammered down.

chases2:
The issue isn't the command, it's the _single space_ before the semicolon.

Removing the space makes the script work as intended.

munnerz:
Closing in favour of #16692

PR #16692, opened 2020-03-09
Transfigure: Change to detect changes with git

chases2:
In the current implementation, the script is unable to detect changes at all.
See https://kubernetes.slack.com/archives/C09QZ4DQB/p1583754596266200 and #16691 

This version does the right boolean thing, while being harder to mistype (single spaces are 👎 )
It also prints to logging the differences that transfigure found.

/assign @munnerz @clarketm @michelle192837

### DIFF
--- a/testgrid/cmd/transfigure/transfigure.sh
+++ b/testgrid/cmd/transfigure/transfigure.sh
@@ -69,7 +69,7 @@ main() {
     --output "${testgrid_dir}/${testgrid_subdir}/gen-config.yaml"
 
 
-  if ! git diff --quiet ; then
+  if [[ $(git diff --exit-code) -eq 0 ]]; then
     echo "Transfigure did not change anything. Aborting no-op bump"
     exit 0
   fi


testgrid/cmd/transfigure/transfigure.sh
```diff
@@ -69,7 +69,7 @@ main() {
     --output "${testgrid_dir}/${testgrid_subdir}/gen-config.yaml"
 
 
-  if ! git diff --quiet ; then
+  if [[ $(git diff --exit-code) -eq 0 ]]; then
```
munnerz: This works fine, except for in the case where the config directory does not already exist. If there are untracked files in the repository, `git diff --exit-code` still returns 0.

This is probably sufficient to fix the current issues, but thought it worth mentioning, especially as there is handling for creating the subdirectory above too:
```
  if [[ ! -d "${testgrid_dir}/${testgrid_subdir}" ]]; then
    echo "Subdirectory ${testgrid_subdir} doesn't exist; creating it" >&2
    mkdir -p "${testgrid_dir}/${testgrid_subdir}"
  fi
```

munnerz: I think the previous (working) iteration of the script also had the same issue, so I'll add my lgtm + a hold - feel free to unhold if you want to fix it later 😄 

/lgtm
/hold

clarketm:
@chases2 

https://github.com/istio/test-infra/pull/2472#issuecomment-596802353

https://storage.googleapis.com/istio-prow/pr-logs/pull/istio_test-infra/2472/pull-test-infra-check-testgrid-config/674/build-log.txt

```console
cat: '': No such file or directory
Using user from GitHub: 
cat: '': No such file or directory
Using email from GitHub: 
Ensuring kubernetes/test-infra repo
Cloning into 'test-infra'...
Created temporary repository
Checking out transfigure-branch
Switched to a new branch 'transfigure-branch'
Generating testgrid yaml
/transfigure.sh: line 72: diff: unbound variable
Cleaning temporary repository at /home/prow/go/src/github.com/istio/test-infra/test-infra
```

The `pull-test-infra-check-testgrid-config` job in **istio** in unhappy about this change. 😭

Issue #16281, opened 2020-02-13
boskos: remove apiextensions/customresourcedefinitions rule from boskos ClusterRole


ixdy:
**What should be cleaned up or changed**:
Once Boskos no longer attempts to create the Boskos CRDs itself (i.e. #16266 has rolled out), we should remove that permission from its ClusterRole:
https://github.com/kubernetes/test-infra/blob/49ecce6586df439db2f4bdce4c4d6a3956e87d79/prow/cluster/boskos.yaml#L56-L58

The CRDs are now saved as YAMLs in the cluster deployment directory (#16267), which makes deployment and upgrades easier, and it also reduces the permissions we need to give to Boskos.

riking:
Dependent PR is merged and seems rolled out.

PR #16699, opened 2020-03-09
remove customresourcedefinitions from boskos ClusterRole

ixdy:
Fixes https://github.com/kubernetes/test-infra/issues/16281.

cc @alvaroaleman @stevekuznetsov

### DIFF
--- a/prow/cluster/boskos.yaml
+++ b/prow/cluster/boskos.yaml
@@ -82,9 +82,6 @@ apiVersion: rbac.authorization.k8s.io/v1
 metadata:
   name: boskos
 rules:
-- apiGroups: ["apiextensions.k8s.io"]
-  verbs: ["*"]
-  resources: ["customresourcedefinitions"]  # TODO(#16281): remove once Boskos no longer attempts to create CRDs itself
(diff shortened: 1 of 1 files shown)


ixdy:
cc @BenTheElder

prow/cluster/boskos.yaml
```diff
@@ -98,7 +95,7 @@ metadata:
 kind: ClusterRoleBinding
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
-  name: crd-creator
+  name: boskos
```
alvaroaleman: Someone needs to manually delete the old `crd-creator` role after this gets deployed

ixdy: @BenTheElder is the current on-call

BenTheElder: presumably after the next bump? I've had a full schedule today and haven't bumped yet.

ixdy: This has been deployed now, so the old role binding (named `crd-creator`) can be deleted.

BenTheElder: this is done

PR #16712, opened 2020-03-10
Migrate scalability tests to use e2- machine family for nodes

mborsz:
This PR migrates nodes in scalability tests to use e2- instance family.

This excludes:
* masters
* some experimental jobs
* presubmit jobs

/assign @wojtek-t


config/jobs/kubernetes/sig-scalability/sig-scalability-release-blocking-jobs.yaml
```diff
@@ -72,6 +72,7 @@ periodics:
       - --env=KUBE_DNS_MEMORY_LIMIT=300Mi
       - --extract=ci/latest
       - --gcp-nodes=5000
+      - --gcp-node-size=e2-standard-2
```
wojtek-t: As discussed offline - we need to ensure that we have enough quota for these jobs.

Issue #16711, opened 2020-03-10
crier v20200309-111b23480 panicing with concurrent map write


alvaroaleman:
**What happened**:

Crier crashing with concurrent map write

https://gist.github.com/alvaroaleman/8d0e506d49b1ac2587b94b95309a695f

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

**Please provide links to example occurrences, if any**:

**Anything else we need to know?**:

/area prow/crier

PR #16714, opened 2020-03-10
Fix a data race in decoration config defaulting

alvaroaleman:
Fixes https://github.com/kubernetes/test-infra/issues/16711
/assign fejta

### DIFF
--- a/experiment/resultstore/convert_test.go
+++ b/experiment/resultstore/convert_test.go
@@ -18,6 +18,7 @@ package main
 
 import (
 	"encoding/xml"
+	"sort"
 	"testing"
 	"time"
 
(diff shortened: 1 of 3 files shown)


prow/apis/prowjobs/v1/types.go
```diff
@@ -474,7 +474,7 @@ func (u *UtilityImages) ApplyDefault(def *UtilityImages) *UtilityImages {
 		return u
 	}
 
-	merged := *u
+	merged := *u.DeepCopy()
```
alvaroaleman: This isn't really necessary today as the `*UtilityImages` does not contain any pointers, but I added it to make sure it stays threadsafe should that ever change

prow/apis/prowjobs/v1/types.go
```diff
@@ -533,9 +533,9 @@ func (g *GCSConfiguration) ApplyDefault(def *GCSConfiguration) *GCSConfiguration
 	}
 	var merged GCSConfiguration
 	if g != nil {
-		merged = *g
+		merged = *g.DeepCopy()
```
alvaroaleman: For the concrete issue at hand, doing the `DeepCopy` in the `ApplyDefaults` of `*DecorationConfig` is enough, but since this is public, there may be other callers that expect it to be threadsafe

experiment/resultstore/convert_test.go
```diff
@@ -94,23 +95,26 @@ func TestStartedReposToProperties(t *testing.T) {
...
+				{Key: "Repo", Value: "repo1"},
+				{Key: "Repo", Value: "repo2"},
 			},
 		},
 	}
 
 	for _, tc := range cases {
 		t.Run(tc.name, func(t *testing.T) {
 			properties := startedReposToProperties(tc.repos)
+			sort.Slice(properties, func(i, j int) bool {
```
fejta: Thank you for fixing this!!

PR #16722, opened 2020-03-10
s3: refactor gcs-upload and enable s3 support for gcsupload, sidecar & initupload

sbueringer:
Goal of this PR is to leverage the new s3-enabled opener package for gcsupload, sidecar & initupload which is the next step to implement https://github.com/kubernetes/test-infra/issues/11260.
The initial commit on this PR is just one of the variants how this can be achieved. Imho we should first get consensus on some of the following issues:

* Refactoring of `gcs.Upload` (and also FileUpload, DataUpload, ..)
    * Implementation options:
        * Move implementation from gcs to opener package
        * Keep implementation in gcs package (replace gcsObjectWriter through something generic which uses the opener)
        * Create a new package which replaces gcs (e.g. blobstorage) and leverages opener
    * Considerations:
        * How invasive is the change
        * In which variants would we hide it behind a feature gate
    * Notes:
        * gcs currently exposes *storage.ObjectAttrs. I guess we also don't want to expose *blob.ObjectAttrs
            * So we need our own struct which hides both implementations?

* Changes to `DecorationConfig`
    * In all variants: add `s3_credentials_secret` field similar to `gcs_credentials_secret`
    * Options regarding GcsConfiguration:s3_credentials_secrets3_credentials_secret
        * extending GcsConfiguration
        * create new BlobStorageConfiguration and deprecate GcsConfiguration 
    * Options regarding GcsConfiguration fields:
        * keeping the bucket field and changing the behaviour:
            * use the protocol the bucket starts with, e.g. s3://prow-artifacts => s3
            * defaulting to file if it starts with /
            * defaulting to gs:// if no prefix is discovered 
        * deprecating the bucket and similar fields and adding a path field 
            * path field can now be something like:
                * gs://prow-artifacts
                * /local/path/to/files
                * s3://prow-artifacts/some/deep/folder/structure
    * Notes:
        * this has impact on how the decoration config is propagated to / parsed by the test pods

* What about the gcsupload binary:
  * extend (and possibly rename it) to support other storage provider
  * create a new blobstorageupload binary   
 
* Notes about the current implementation on this PR:
    * Moved implementation from gcs to opener package
    * Extended GcsConfiguration to avoid introducing a new struct and the corresponding migration efforts for setting / parsing both

Please ignore some details in the current implementation for now like the path handling in gcsupload/run.go. I would like to get 
consensus on the major issues so I can then refactor it further on the basis of the current implementation.

### DIFF
--- a/pkg/io/opener.go
+++ b/pkg/io/opener.go
@@ -23,6 +23,7 @@ import (
 	"io"
 	"io/ioutil"
 	"os"
+	"path"
 	"strings"
 	"sync"
 
@@ -50,7 +51,7 @@ type (
 // Opener has methods to read and write paths
 type Opener interface {
 	Reader(ctx context.Context, path string) (ReadCloser, error)
-	Writer(ctx context.Context, path string) (WriteCloser, error)
+	Writer(ctx context.Context, path string, opts ...WriterOptions) (WriteCloser, error)
 }
 
 type opener struct {
@@ -164,9 +165,7 @@ func (o *opener) Reader(ctx context.Context, path string) (io.ReadCloser, error)
 		if err != nil {
 			return nil, fmt.Errorf("bad gcs path: %v", err)
 		}
-		if g != nil {
-			return g.NewReader(ctx)
-		}
+		return g.NewReader(ctx)
 	}
 	if strings.HasPrefix(path, "/") {
 		return os.Open(path)
@@ -184,25 +183,36 @@ func (o *opener) Reader(ctx context.Context, path string) (io.ReadCloser, error)
 }
 
 // Writer returns a writer that overwrites the path.
-func (o *opener) Writer(ctx context.Context, path string) (io.WriteCloser, error) {
-	if strings.HasPrefix(path, "gs://") {
-		g, err := o.openGCS(path)
+func (o *opener) Writer(ctx context.Context, p string, opts ...WriterOptions) (io.WriteCloser, error) {
+	if strings.HasPrefix(p, "gs://") {
+		g, err := o.openGCS(p)
 		if err != nil {
 			return nil, fmt.Errorf("bad gcs path: %v", err)
 		}
-		if g != nil {
-			return g.NewWriter(ctx), nil
+		writer := g.NewWriter(ctx)
+		for _, opt := range opts {
+			opt.Apply(writer, nil)
 		}
+		return writer, nil
 	}
-	if strings.HasPrefix(path, "/") {
-		return os.Create(path)
+	if strings.HasPrefix(p, "/") {
+		// create parent dir if doesn't exist
+		dir := path.Dir(p)
+		if err := os.MkdirAll(dir, 0755); err != nil {
+			return nil, fmt.Errorf("create directory %q: %w", dir, err)
+		}
+		return os.OpenFile(p, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0666)
 	}
 
-	bucket, relativePath, err := o.getBucket(ctx, path)
+	bucket, relativePath, err := o.getBucket(ctx, p)
 	if err != nil {
 		return nil, err
 	}
-	writer, err := bucket.NewWriter(ctx, relativePath, nil)
+	var wOpts blob.WriterOptions
+	for _, opt := range opts {
+		opt.Apply(nil, &wOpts)
+	}
+	writer, err := bucket.NewWriter(ctx, relativePath, &wOpts)
 	if err != nil {
 		return nil, err
 	}
--- a/pkg/io/option.go
+++ b/pkg/io/option.go
@@ -1,0 +1,69 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package io
+
+import (
+	"cloud.google.com/go/storage"
+	"gocloud.dev/blob"
+	"google.golang.org/api/googleapi"
+)
+
+// WriterOptions are options for the opener Writer method
+type WriterOptions struct {
+	BufferSize      *int64
+	ContentEncoding *string
+	ContentType     *string
+	Metadata        map[string]string
+}
+
+// Apply applies the WriterOptions to storage.Writer and blob.WriterOptions
+// Both arguments are allowed to be nil.
+func (wo WriterOptions) Apply(writer *storage.Writer, o *blob.WriterOptions) {
+	if writer != nil {
+		if wo.BufferSize != nil {
+			if *wo.BufferSize < googleapi.DefaultUploadChunkSize {
+				writer.ChunkSize = int(*wo.BufferSize)
+			}
+		}
+		if wo.ContentEncoding != nil {
+			writer.ObjectAttrs.ContentEncoding = *wo.ContentEncoding
+		}
+		if wo.ContentType != nil {
+			writer.ObjectAttrs.ContentType = *wo.ContentType
+		}
+		if wo.Metadata != nil {
+			writer.ObjectAttrs.Metadata = wo.Metadata
+		}
+	}
+
+	if o == nil {
+		return
+	}
+
+	if wo.BufferSize != nil {
+		o.BufferSize = int(*wo.BufferSize)
+	}
+	if wo.ContentEncoding != nil {
+		o.ContentEncoding = *wo.ContentEncoding
+	}
+	if wo.ContentType != nil {
+		o.ContentType = *wo.ContentType
+	}
+	if wo.Metadata != nil {
+		o.Metadata = wo.Metadata
+	}
+}
--- a/prow/apis/prowjobs/v1/types.go
+++ b/prow/apis/prowjobs/v1/types.go
@@ -332,6 +332,9 @@ type DecorationConfig struct {
 	// GCSCredentialsSecret is the name of the Kubernetes secret
 	// that holds GCS push credentials.
 	GCSCredentialsSecret string `json:"gcs_credentials_secret,omitempty"`
+	// S3CredentialsSecret is the name of the Kubernetes secret
+	// that holds blob storage push credentials.
+	S3CredentialsSecret string `json:"s3_credentials_secret,omitempty"`
 	// SSHKeySecrets are the names of Kubernetes secrets that contain
 	// SSK keys which should be used during the cloning process.
 	SSHKeySecrets []string `json:"ssh_key_secrets,omitempty"`
@@ -462,8 +465,8 @@ func (d *DecorationConfig) Validate() error {
 	if d.GCSConfiguration == nil {
 		return errors.New("GCS upload configuration is not specified")
 	}
-	if d.GCSCredentialsSecret == "" {
-		return errors.New("GCS upload credential secret is not specified")
+	if d.GCSCredentialsSecret == "" && d.S3CredentialsSecret == "" {
+		return errors.New("neither GCS nor S3 credential secret are specified")
 	}
 	if err := d.GCSConfiguration.Validate(); err != nil {
 		return fmt.Errorf("GCS configuration is invalid: %v", err)
@@ -530,7 +533,10 @@ const (
 // GCSConfiguration holds options for pushing logs and
 // artifacts to GCS from a job.
 type GCSConfiguration struct {
-	// Bucket is the GCS bucket to upload to
+	// Bucket is the bucket to upload to, it can be:
+	// * a GCS bucket: with gs:// prefix
+	// * a S3 bucket: with s3:// prefix
+	// * a GCS bucket: without a prefix (deprecated, it's discouraged to use Bucket without prefix please add the gs:// prefix)
 	Bucket string `json:"bucket,omitempty"`
 	// PathPrefix is an optional path that follows the
 	// bucket name and comes before any structure
@@ -549,7 +555,7 @@ type GCSConfiguration struct {
 	// to media types, for example: MediaTypes["log"] = "text/plain"
 	MediaTypes map[string]string `json:"mediaTypes,omitempty"`
 
-	// LocalOutputDir specifies a directory where files should be copied INSTEAD of uploading to GCS.
+	// LocalOutputDir specifies a directory where files should be copied INSTEAD of uploading to blob storage.
 	// This option is useful for testing jobs that use the pod-utilities without actually uploading.
 	LocalOutputDir string `json:"local_output_dir,omitempty"`
 }
--- a/prow/gcsupload/options.go
+++ b/prow/gcsupload/options.go
@@ -36,7 +36,7 @@ func NewOptions() *Options {
 }
 
 // Options exposes the configuration necessary
-// for defining where in GCS an upload will land.
+// for defining where in storage an upload will land.
 type Options struct {
 	// Items are files or directories to upload.
 	Items []string `json:"items,omitempty"`
@@ -46,10 +46,19 @@ type Options struct {
 
 	*prowapi.GCSConfiguration
 
-	// GcsCredentialsFile is the path to the JSON
-	// credentials for pushing to GCS.
+	// GcsCredentialsFile is used for reading/writing to GCS block storage.
+	// It's optional, if you want to write to local paths or GCS credentials auto-discovery is used.
+	// If set, this file is used to read/write to gs:// paths
+	// If not, credential auto-discovery is used
 	GcsCredentialsFile string `json:"gcs_credentials_file,omitempty"`
-	DryRun             bool   `json:"dry_run"`
+	// S3CredentialsFile is used for reading/writing to s3 block storage.
+	// It's optional, if you want to write to local paths or S3 credentials auto-discovery is used.
+	// If set, this file is used to read/write to s3:// paths
+	// If not, go cloud credential auto-discovery is used
+	// For more details see the pkg/io/providers pkg.
+	S3CredentialsFile string `json:"s3_credentials_file,omitempty"`
+
+	DryRun bool `json:"dry_run"`
 
 	// mediaTypes holds additional extension media types to add to Go's
 	// builtin's and the local system's defaults.  Values are
@@ -79,8 +88,8 @@ func (o *Options) Validate() error {
 			return errors.New("GCS upload was requested no GCS bucket was provided")
 		}
 
-		if o.GcsCredentialsFile == "" {
-			return errors.New("GCS upload was requested but no GCS credentials file was provided")
+		if o.GcsCredentialsFile == "" && o.S3CredentialsFile == "" {
+			return errors.New("blob storage upload was requested but neither GCS nor S3 credentials file was provided")
 		}
 	}
 
@@ -127,6 +136,7 @@ func (o *Options) AddFlags(fs *flag.FlagSet) {
 
 	fs.Var(&o.gcsPath, "gcs-path", "GCS path to upload into")
 	fs.StringVar(&o.GcsCredentialsFile, "gcs-credentials-file", "", "file where Google Cloud authentication credentials are stored")
+	fs.StringVar(&o.S3CredentialsFile, "s3-credentials-file", "", "file where the S3 credentials are stored")
 	fs.BoolVar(&o.DryRun, "dry-run", true, "do not interact with GCS")
 
 	fs.Var(&o.mediaTypes, "media-type", "Optional comma-delimited set of extension media types.  Each entry is colon-delimited {extension}:{media-type}, for example, log:text/plain.")
--- a/prow/gcsupload/run.go
+++ b/prow/gcsupload/run.go
@@ -17,17 +17,15 @@ limitations under the License.
 package gcsupload
 
 import (
-	"context"
 	"fmt"
 	"mime"
+	"net/url"
 	"os"
 	"path"
 	"path/filepath"
 	"strings"
 
-	"cloud.google.com/go/storage"
 	"github.com/sirupsen/logrus"
-	"google.golang.org/api/option"
 
 	prowapi "k8s.io/test-infra/prow/apis/prowjobs/v1"
 	"k8s.io/test-infra/prow/pod-utils/downwardapi"
@@ -40,13 +38,16 @@ import (
 // to their destination in GCS, so the caller can
 // operate relative to the base of the GCS dir.
 func (o Options) Run(spec *downwardapi.JobSpec, extra map[string]gcs.UploadFunc) error {
-	logrus.WithField("options", o).Debug("Uploading to GCS")
+	logrus.WithField("options", o).Debug("Uploading to blob storage")
 
 	for extension, mediaType := range o.GCSConfiguration.MediaTypes {
 		mime.AddExtensionType("."+extension, mediaType)
 	}
 
-	uploadTargets := o.assembleTargets(spec, extra)
+	uploadTargets, err := o.assembleTargets(spec, extra)
+	if err != nil {
+		return fmt.Errorf("assembleTargets: %w", err)
+	}
 
 	if o.DryRun {
 		for destination := range uploadTargets {
@@ -56,26 +57,21 @@ func (o Options) Run(spec *downwardapi.JobSpec, extra map[string]gcs.UploadFunc)
 	}
 
 	if o.LocalOutputDir == "" {
-		gcsClient, err := storage.NewClient(context.Background(), option.WithCredentialsFile(o.GcsCredentialsFile))
-		if err != nil {
-			return fmt.Errorf("could not connect to GCS: %v", err)
+		if err := gcs.Upload(o.Bucket, o.GcsCredentialsFile, o.S3CredentialsFile, uploadTargets); err != nil {
+			return fmt.Errorf("failed to upload to blob storage: %w", err)
 		}
-
-		if err := gcs.Upload(gcsClient.Bucket(o.Bucket), uploadTargets); err != nil {
-			return fmt.Errorf("failed to upload to GCS: %v", err)
-		}
-		logrus.Info("Finished upload to GCS")
+		logrus.Info("Finished upload to blob storage")
 	} else {
 		if err := gcs.LocalExport(o.LocalOutputDir, uploadTargets); err != nil {
-			return fmt.Errorf("failed to copy files to %q: %v", o.LocalOutputDir, err)
+			return fmt.Errorf("failed to copy files to %q: %w", o.LocalOutputDir, err)
 		}
 		logrus.Infof("Finished copying files to %q.", o.LocalOutputDir)
 	}
 	return nil
 }
 
-func (o Options) assembleTargets(spec *downwardapi.JobSpec, extra map[string]gcs.UploadFunc) map[string]gcs.UploadFunc {
-	jobBasePath, gcsPath, builder := PathsForJob(o.GCSConfiguration, spec, o.SubDir)
+func (o Options) assembleTargets(spec *downwardapi.JobSpec, extra map[string]gcs.UploadFunc) (map[string]gcs.UploadFunc, error) {
+	jobBasePath, blobStoragePath, builder := PathsForJob(o.GCSConfiguration, spec, o.SubDir)
 
 	uploadTargets := map[string]gcs.UploadFunc{}
 
@@ -84,7 +80,17 @@ func (o Options) assembleTargets(spec *downwardapi.JobSpec, extra map[string]gcs
 		// ensure that an alias exists for any
 		// job we're uploading artifacts for
 		if alias := gcs.AliasForSpec(spec); alias != "" {
-			fullBasePath := "gs://" + path.Join(o.Bucket, jobBasePath)
+			parsedBucket, err := url.Parse(o.Bucket)
+			if err != nil {
+				return nil, fmt.Errorf("parse bucket %q: %w", o.Bucket, err)
+			}
+			// only add gs:// prefix if o.Bucket itself doesn't already have a scheme prefix
+			var fullBasePath string
+			if parsedBucket.Scheme == "" {
+				fullBasePath = "gs://" + path.Join(o.Bucket, jobBasePath)
+			} else {
+				fullBasePath = fmt.Sprintf("%s/%s", o.Bucket, jobBasePath)
+			}
 			uploadTargets[alias] = gcs.DataUploadWithMetadata(strings.NewReader(fullBasePath), map[string]string{
 				"x-goog-meta-link": fullBasePath,
 			})
@@ -93,14 +99,14 @@ func (o Options) assembleTargets(spec *downwardapi.JobSpec, extra map[string]gcs
 		if latestBuilds := gcs.LatestBuildForSpec(spec, builder); len(latestBuilds) > 0 {
 			for _, latestBuild := range latestBuilds {
 				dir, filename := path.Split(latestBuild)
-				metadataFromFileName, attrs := gcs.AttributesFromFileName(filename)
-				uploadTargets[path.Join(dir, metadataFromFileName)] = gcs.DataUploadWithAttributes(strings.NewReader(spec.BuildID), attrs)
+				metadataFromFileName, writerOptions := gcs.WriterOptionsFromFileName(filename)
+				uploadTargets[path.Join(dir, metadataFromFileName)] = gcs.DataUploadWithOptions(strings.NewReader(spec.BuildID), writerOptions)
 			}
 		}
 	} else {
 		// Remove the gcs path prefix in local mode so that items are rooted in the output dir without
 		// excessive directory nesting.
-		gcsPath = ""
+		blobStoragePath = ""
 	}
 
 	for _, item := range o.Items {
@@ -110,27 +116,27 @@ func (o Options) assembleTargets(spec *downwardapi.JobSpec, extra map[string]gcs
 			continue
 		}
 		if info.IsDir() {
-			gatherArtifacts(item, gcsPath, info.Name(), uploadTargets)
+			gatherArtifacts(item, blobStoragePath, info.Name(), uploadTargets)
 		} else {
-			metadataFromFileName, attrs := gcs.AttributesFromFileName(info.Name())
-			destination := path.Join(gcsPath, metadataFromFileName)
+			metadataFromFileName, writerOptions := gcs.WriterOptionsFromFileName(info.Name())
+			destination := path.Join(blobStoragePath, metadataFromFileName)
 			if _, exists := uploadTargets[destination]; exists {
 				logrus.Warnf("Encountered duplicate upload of %s, skipping...", destination)
 				continue
 			}
-			uploadTargets[destination] = gcs.FileUploadWithAttributes(item, attrs)
+			uploadTargets[destination] = gcs.FileUploadWithOptions(item, writerOptions)
 		}
 	}
 
 	for destination, upload := range extra {
-		uploadTargets[path.Join(gcsPath, destination)] = upload
+		uploadTargets[path.Join(blobStoragePath, destination)] = upload
 	}
 
-	return uploadTargets
+	return uploadTargets, nil
 }
 
 // PathsForJob determines the following for a job:
-//  - path in GCS under the bucket where job artifacts will be uploaded for:
+//  - path in blob storage under the bucket where job artifacts will be uploaded for:
 //     - the job
 //     - this specific run of the job (if any subdir is present)
 // The builder for the job is also returned for use in other path resolution.
@@ -140,14 +146,14 @@ func PathsForJob(options *prowapi.GCSConfiguration, spec *downwardapi.JobSpec, s
 	if options.PathPrefix != "" {
 		jobBasePath = path.Join(options.PathPrefix, jobBasePath)
 	}
-	var gcsPath string
+	var blobStoragePath string
 	if subdir == "" {
-		gcsPath = jobBasePath
+		blobStoragePath = jobBasePath
 	} else {
-		gcsPath = path.Join(jobBasePath, subdir)
+		blobStoragePath = path.Join(jobBasePath, subdir)
 	}
 
-	return jobBasePath, gcsPath, builder
+	return jobBasePath, blobStoragePath, builder
 }
 
 func builderForStrategy(strategy, defaultOrg, defaultRepo string) gcs.RepoPathBuilder {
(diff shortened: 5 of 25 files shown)


sbueringer:
@alvaroaleman @fejta What do you think about the issues I posted above?

prow/apis/prowjobs/v1/types.go
```diff
@@ -501,6 +504,11 @@ const (
 // GCSConfiguration holds options for pushing logs and
 // artifacts to GCS from a job.
 type GCSConfiguration struct {
+	// Path is the path to upload to
+	// can be:
+	// * a local path: /path/to/storage
+	// * a remote path: e.g. gs://prow-artifacts or s3://prow-artifacts
+	Path string `json:"path,omitempty"`
```
alvaroaleman: Why do we need this, I think this should be configured in the `Bucket` field or not?

pkg/io/opener.go
```diff
@@ -51,6 +52,7 @@ type (
 type Opener interface {
 	Reader(ctx context.Context, path string) (ReadCloser, error)
 	Writer(ctx context.Context, path string) (WriteCloser, error)
+	Upload(ctx context.Context, uploads map[string]UploadFunc) error
```
alvaroaleman: Not sure if this belongs in here, it essentially just wraps the `Writer`, right? Feels to me like it should be a level above the `Opener`

pkg/io/opener.go
```diff
@@ -208,3 +210,86 @@ func (o *opener) Writer(ctx context.Context, path string) (io.WriteCloser, error
...
+	}
+	return nil
+}
+
+func convertAttributesToWriterOptions(a *blob.Attributes) *blob.WriterOptions {
+	if a == nil {
+		return nil
+	}
+	return &blob.WriterOptions{
+		CacheControl:       a.CacheControl,
```
alvaroaleman: Which of these do we actually use?

pkg/io/opener.go
sbueringer: Yup that's the question I'm asking above:
* Move implementation from gcs to opener package
* Keep implementation in gcs package (replace gcsObjectWriter through something generic which uses the opener)
* Create a new package which replaces gcs (e.g. blobstorage) and leverages opener

prow/apis/prowjobs/v1/types.go
sbueringer: Depends on what our goal is that's this point, options:
* keeping the bucket field and changing the behaviour:
   * use the protocol the bucket starts with, e.g. s3://prow-artifacts => s3
   * defaulting to file if it starts with /
   * defaulting to gs:// if no prefix is discovered
* deprecating the bucket and similar fields and adding a path field

pkg/io/opener.go
sbueringer: Not used in the current version. That's because of the rebase. Depending on how/where we want to implement the upload functionality I would either keep it or move it away. (It was actually used in l.248. but that's not possible anymore because `Writer(...)` doesn't have a WriterOptions parameter anymore. I would clean this up of course as soon as we decide where the Upload functionality should be implemented.

prow/pod-utils/gcs/metadata.go
```diff
@@ -42,8 +41,8 @@ type Finished = metadata.Finished
 //
 // and the simplified filename would be build-log.txt (excluding the
 // content encoding extension).
-func AttributesFromFileName(filename string) (string, *storage.ObjectAttrs) {
```
fejta: Why are we confident that there are no unintended systemic side effects switching `storage.ObjectAttrs` to `blob.Attributes`?

prow/sidecar/run.go
```diff
@@ -31,6 +31,7 @@ import (
 
 	"github.com/sirupsen/logrus"
 
+	pkgio "k8s.io/test-infra/pkg/io"
```
fejta: Intent is that doing this should involve deleting the "io" import not having both existing side by side where one is renamed (but I suspect this should really just be the gcs import)

prow/apis/prowjobs/v1/types.go
alvaroaleman: I do not have objections against the idea of a new `Path` or `PathPrefix` field, but it is IMHO a change that is orthogonal to S3 support. It also has a huge blast radius as it basically requires adjustment of everything in Prow that somehow relies on knowing paths, so we should keep it out of the S3 support story and instead just start allowing a provider name prefix that gets defaulted to `gs://`.

sbueringer:
Thx for the feedback. So I proceed for now with:
* keeping the gcs package and changing the internal implementation to the io package
* not adding a path property for now and changing the semantics of the bucket property so it can be prefixed with the protocol. When no prefix is found it will default to gs://

Let's see how the result of the refactoring looks :)

fejta:
POR SGTM

> keeping the gcs package and changing the internal implementation to the io package

Anticipating my future review comment, "this change looks scary, can we continue the existing behavior if the path doesn't match `s3://`"

There may be good reasons why want to both enable s3 and switch gs:// over to use the io package in the same release (as opposed to s3 in one and gs in the next one), so if you think this is best may try to point out why once you've considered both options.

sbueringer:
@fejta @alvaroaleman 
I think I found a way to implement it all how we want it to be :)

The only thing I don't like is that we have no way to set WriterOptions (attributes) when we're upload to any non-GCS provider. But I'm not sure what the consequences are of not setting the attributes

prow/pod-utils/gcs/upload.go
```diff
@@ -171,12 +197,51 @@ func (w gcsObjectWriter) ApplyAttributes(attrs *storage.ObjectAttrs) {
...
+type openerObjectWriter struct {
+	pkgio.Opener
+	Bucket      string
+	Dest        string
+	writeCloser pkgio.WriteCloser
+}
+
+func (w *openerObjectWriter) Write(p []byte) (n int, err error) {
+	if w.writeCloser == nil {
+		w.writeCloser, err = w.Opener.Writer(context.Background(), fmt.Sprintf("%s/%s", w.Bucket, w.Dest))
```
alvaroaleman: The context should be part of the `openerObjectWriter`

prow/pod-utils/gcs/upload.go
```diff
@@ -171,12 +197,51 @@ func (w gcsObjectWriter) ApplyAttributes(attrs *storage.ObjectAttrs) {
 	w.Writer.ObjectAttrs = *attrs
 }
 
+type openerObjectWriter struct {
+	pkgio.Opener
+	Bucket      string
+	Dest        string
+	writeCloser pkgio.WriteCloser
```
alvaroaleman: if we ever want to use this for GCS as well or pass any kind of opts, we need to change the `Writer` method in the `io` package to accept some kind of opts. Could probably be a simplistic struct that only contains GCS options and gets extended in the future if the need arises. We can do this backwards-compatible by leveraging the functional opts pattern.

prow/pod-utils/gcs/upload.go
alvaroaleman: if we ever want to use this for GCS as well or pass any kind of opts, we need to change the `Writer` method in the `io` package to accept some kind of opts. Could probably be a simplistic struct that only contains GCS options and gets extended in the future if the need arises. We can do this backwards-compatible by leveraging the functional opts pattern.

prow/pod-utils/gcs/upload.go
```diff
@@ -171,12 +197,51 @@ func (w gcsObjectWriter) ApplyAttributes(attrs *storage.ObjectAttrs) {
...
+}
+
+func (w *openerObjectWriter) SetSize(_ int64) {
+	// ChunkSize is in gocloud only implemented by GCS
+	// but we're not using the openerObjectWriter for GCS
+	// so we don't have to set this
+	return
+}
+
+func (w *openerObjectWriter) ApplyAttributes(_ *storage.ObjectAttrs) {
```
alvaroaleman: if we ever want to use this for GCS as well or pass any kind of opts, we need to change the `Writer` method in the `io` package to accept some kind of opts. Could probably be a simplistic struct that only contains GCS options and gets extended in the future if the need arises. We can do this backwards-compatible by leveraging the functional opts pattern.

prow/pod-utils/gcs/upload.go
```diff
@@ -171,12 +197,51 @@ func (w gcsObjectWriter) ApplyAttributes(attrs *storage.ObjectAttrs) {
...
+	return
+}
+
 type localFileWriter struct {
 	filePath string
 	file     *os.File
 }
 
-func (*localFileWriter) SetSize(size int64) {}
+func (*localFileWriter) SetSize(_ int64) {}
```
alvaroaleman: Probably more of a follow-up but this can be removed and replaces with the `openerObjectWriter` I guess?

prow/gcsupload/run.go
```diff
@@ -84,7 +76,17 @@ func (o Options) assembleTargets(spec *downwardapi.JobSpec, extra map[string]gcs
 		// ensure that an alias exists for any
 		// job we're uploading artifacts for
 		if alias := gcs.AliasForSpec(spec); alias != "" {
-			fullBasePath := "gs://" + path.Join(o.Bucket, jobBasePath)
+			parsedBucket, err := url.Parse(o.Bucket)
+			if err != nil {
+				logrus.Fatalf("Encountered error when parsing bucket name %s: %v", o.Bucket, err)
```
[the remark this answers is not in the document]
sbueringer: Got it. Not sure what I though when writing this. I'll just return an error and therefore change the signature of `assembleTargets`

prow/pod-utils/gcs/upload.go
sbueringer: I'm not entirely sure. Currently we use only `os.Create` in the opener and `MkdirAll, OpenFile, Write` in the `localFileWriter`.  I'm not sure if `os.Create` also creates directories if necessary. But this might be something we could add to the implementation in opener

prow/pod-utils/decorate/podspec.go
```diff
@@ -615,7 +641,7 @@ func decorate(spec *coreapi.PodSpec, pj *prowapi.ProwJob, rawEnv map[string]stri
 		}
 	}
 
-	gcsVol, gcsMount, gcsOptions := GCSOptions(*pj.Spec.DecorationConfig, localMode)
+	gcsVol, s3Vol, gcsMount, s3Mount, gcsOptions := BlobStorageOptions(*pj.Spec.DecorationConfig, localMode)
```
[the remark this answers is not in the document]
sbueringer: Done

prow/pod-utils/decorate/podspec.go
```diff
@@ -500,38 +502,62 @@ func GCSOptions(dc prowapi.DecorationConfig, localMode bool) (*coreapi.Volume, *
...
 
-	vol := &coreapi.Volume{
-		Name: gcsCredentialsMountName,
-		VolumeSource: coreapi.VolumeSource{
-			Secret: &coreapi.SecretVolumeSource{
-				SecretName: dc.GCSCredentialsSecret,
+	var gcsVol *coreapi.Volume
+	var s3Vol *coreapi.Volume
+	var gcsMount *coreapi.VolumeMount
+	var s3Mount *coreapi.VolumeMount
```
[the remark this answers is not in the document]
sbueringer: You're absolutely right. Done

prow/gcsupload/run.go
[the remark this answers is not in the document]
sbueringer: Done

prow/pod-utils/gcs/upload.go
```diff
@@ -36,11 +40,33 @@ type UploadFunc func(writer dataWriter) error
...
+func Upload(bucket, gcsCredentialsFile, s3CredentialsFile string, uploadTargets map[string]UploadFunc) error {
+	parsedBucket, err := url.Parse(bucket)
+	if err != nil {
+		return fmt.Errorf("cannot parse bucket name %s: %v", bucket, err)
+	}
+	if parsedBucket.Scheme == "" || parsedBucket.Scheme == "gs" {
+		if parsedBucket.Scheme == "gs" {
+			bucket = strings.Trim(bucket, "gs://")
+		}
+		gcsClient, err := storage.NewClient(context.Background(), option.WithCredentialsFile(gcsCredentialsFile))
```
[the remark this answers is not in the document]
sbueringer: This would mean SetSize and ApplyAttributes are not possible anymore because the opener interface does not provide that right now. But I gues that's your point in the comments below

prow/pod-utils/gcs/upload.go
sbueringer: Let's keep this comment about the context and the other ones about the options. context is fixed

prow/pod-utils/gcs/upload.go
sbueringer: That was the reason why I added the blob.WriterOptions in some variant of the previous. Although I admit that the functional opts pattern is better for that kind of requirement.

Anyway. I think the main problem is that we cannot fullfill both use cases at the same time:

* The intent of this package is to:
   *replace https://golang.org/pkg/io/ imports with k8s.io/test-infra/pkg/io and have the code work the same as before (i.e. not changing the opener interface)

* Make the writer customizable so we can set WriterOptions or Attributes when writing

alvaroaleman: > Anyway. I think the main problem is that we cannot fullfill both use cases at the same time:

I think we can, because the opts are set when constructing the `io.WriteCloser` in the `Writer` method, not when calling `Write`, only the latter is part of the `io.Writer`/`io.WriteCloser` interface.

I _think_ @fejta wasn't generally opposed to allowing opts, just to using `*blob.WriterOptions` for it...?

sbueringer: > I think we can, because the opts are set when constructing the io.WriteCloser in the Writer method, not when calling Write, only the latter is part of the io.Writer/io.WriteCloser interface.

This would mean during the construction of the opener right? If that's the case the problem is that we don't know which files we want to write when the opener is constructed and the properties are different per file.

I would prefer adding options for Write(), if that's an option

prow/pod-utils/gcs/upload.go
sbueringer: If we want, I can extend the file writing of the opener like this and use the opener: 
:+1: :-1: ?

prow/pod-utils/gcs/upload.go
alvaroaleman: In theory this sounds good, in practise its not possible to change opts on a `Writer`: https://godoc.org/gocloud.dev/blob#Writer

I am actually surprised the GCS implementation allows this, because that means its not threadsafe. For us, setting the `Opts` when constructing the `Writer` is completely fine as we construct a `Writer` per file anyways (which makes sense due to thread safety and to have a per-file timeout)

sbueringer: Yup. The only question is if we want to change the signature of the Writer() method to include this options in any way. The only alternative would be to create an opener every time when we need an writer and set the options in Opener. But I think that would be really bad and breaks a lot of stuff :)

alvaroaleman: Yes, we do, this doesn't belong into the `Opener`

sbueringer: *push* fejta (if possible :))

alvaroaleman: /assign fejta
WDYT?

sbueringer: *push* @fejta

fejta: In addition to size, it looks like these are the field we set:

https://github.com/kubernetes/test-infra/blob/d2e2e04d1b0335335ae006973ba606e95f585af9/prow/pod-utils/gcs/metadata.go#L54-L76

https://github.com/kubernetes/test-infra/blob/d2e2e04d1b0335335ae006973ba606e95f585af9/prow/pod-utils/gcs/upload.go#L129

It does seem like adding options like this to the opener.Writer interface is the most logical place.

I like the []Option pattern, which should allow us to distinguish between parameters that are common between providers (size, content-type) from possibly provider specific functionality (if-generation-match, chunksize)

prow/pod-utils/gcs/upload.go
fejta: Replacing localFileWriter with openerWriter sgtm

We may want to make auto-creation of subdirs an option on either the opener struct or Writer() call

prow/pod-utils/gcs/upload.go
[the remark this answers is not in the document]
sbueringer: sgtm

prow/gcsupload/run_test.go
```diff
@@ -214,7 +214,11 @@ func TestOptions_AssembleTargets(t *testing.T) {
 			}
 
 			var uploadPaths []string
-			for uploadPath := range testCase.options.assembleTargets(spec, testCase.extra) {
+			targets, err := testCase.options.assembleTargets(spec, testCase.extra)
+			if err != nil {
+				t.Errorf("%s: did not assemble targets correctly: got unexpected error: %v", testCase.name, err)
```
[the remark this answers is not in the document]
sbueringer: My understanding was that %w should only be used when you want to expose the underlying error and therefore make it part of your public API

prow/gcsupload/options.go
```diff
@@ -46,10 +46,19 @@ type Options struct {
 
 	*prowapi.GCSConfiguration
 
-	// GcsCredentialsFile is the path to the JSON
-	// credentials for pushing to GCS.
+	// gcsCredentialsFile string is used for reading/writing to GCS block storage.
+	// If you want to write to local paths, this parameter is optional.
```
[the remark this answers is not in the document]
sbueringer: Updated the comment with. Hope that summarizes it correctly: `// It's optional, if you want to write to local paths or GCS credentials auto-discovery is used.`. Didn't want to list the cases as they might change

prow/apis/prowjobs/v1/types.go
```diff
@@ -501,7 +504,10 @@ const (
 // GCSConfiguration holds options for pushing logs and
 // artifacts to GCS from a job.
 type GCSConfiguration struct {
-	// Bucket is the GCS bucket to upload to
+	// Bucket is the bucket to upload to, it can be:
+	// * a GCS bucket: without a prefix
```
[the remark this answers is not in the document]
sbueringer: Added a comment about it

prow/gcsupload/run.go
```diff
@@ -40,13 +37,16 @@ import (
...
+	logrus.WithField("options", o).Debug("Uploading to blob storage")
 
 	for extension, mediaType := range o.GCSConfiguration.MediaTypes {
 		mime.AddExtensionType("."+extension, mediaType)
 	}
 
-	uploadTargets := o.assembleTargets(spec, extra)
+	uploadTargets, err := o.assembleTargets(spec, extra)
+	if err != nil {
+		return err
```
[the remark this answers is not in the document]
sbueringer: Good point, done

prow/gcsupload/run.go
```diff
@@ -84,7 +79,17 @@ func (o Options) assembleTargets(spec *downwardapi.JobSpec, extra map[string]gcs
 		// ensure that an alias exists for any
 		// job we're uploading artifacts for
 		if alias := gcs.AliasForSpec(spec); alias != "" {
-			fullBasePath := "gs://" + path.Join(o.Bucket, jobBasePath)
+			parsedBucket, err := url.Parse(o.Bucket)
+			if err != nil {
+				return nil, fmt.Errorf("encountered error when parsing bucket name %s: %v", o.Bucket, err)
```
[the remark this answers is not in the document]
sbueringer: Yup makes sense

prow/gcsupload/run_test.go
```diff
@@ -214,7 +214,11 @@ func TestOptions_AssembleTargets(t *testing.T) {
...
+			}
```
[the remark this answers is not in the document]
sbueringer: Added a test case where the bucket name is invalid

prow/pod-utils/gcs/upload.go
```diff
@@ -36,11 +40,34 @@ type UploadFunc func(writer dataWriter) error
...
+		if err != nil {
+			return fmt.Errorf("could not connect to GCS: %v", err)
```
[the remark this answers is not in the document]
sbueringer: Done

prow/pod-utils/gcs/upload.go
```diff
@@ -36,11 +40,34 @@ type UploadFunc func(writer dataWriter) error
...
+func Upload(bucket, gcsCredentialsFile, s3CredentialsFile string, uploadTargets map[string]UploadFunc) error {
+	parsedBucket, err := url.Parse(bucket)
+	if err != nil {
+		return fmt.Errorf("cannot parse bucket name %s: %v", bucket, err)
+	}
+	if parsedBucket.Scheme == "" || parsedBucket.Scheme == "gs" {
+		if parsedBucket.Scheme == "gs" {
+			bucket = strings.Trim(bucket, "gs://")
+		}
+		gcsClient, err := storage.NewClient(context.Background(), option.WithCredentialsFile(gcsCredentialsFile))
```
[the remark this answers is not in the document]
sbueringer: Fine for me. I just preserved the previous behavior of `prow/gcsupload`. I think it will be less redundant once we replaced everything with the opener (as you said after we proved that s3 support works everywhere)

prow/pod-utils/gcs/upload.go
sbueringer: I replaced the localFileWriter with the openerObjectWriter. I would suggest always executing MkdirAll for now as sane default. When we / if we have cases where we don't want to create the parent dir I would add a option for it. There are slight differences in the file permissions:
* opener: `return OpenFile(name, O_RDWR|O_CREATE|O_TRUNC, 0666)`
* localFileWriter: `os.OpenFile(w.filePath, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0755)`

Currently I'm just using the one from opener. Is that okay, should I use something else, or should I add an option for it (not sure if it's important in our current use cases as localFileWriter is just a local debug mode and I don't think we have to write executable files with it)?

prow/pod-utils/gcs/upload.go
sbueringer: I implemented the WriterOptions

Some notes:
* gocloud uses BufferSize as ChunkSize when the gcs implemention of gocloud is used
* I only implemented ContentEncoding, ContentType and MetaData as these are the only attributes we're currently using. The disadvantage is that if we set other attributes e.g. `FileUploadWithAttributes` they are not used. I didn't want to expose `storage.ObjectAttrs` in the opener package. We could:
  * keep it as is
  * add a WithGcsObjectAttrs option (won't solve the problem for other providers)
  * try to implement all options (which are a lot)
  * refactor `FileUploadWithAttributes` to accept only `contentType`, `contentEncoding` and `metadata` 
  The last one is my favorite because it has the least potential for errors and isn't to verbose. In this case, we should probably also refactor `gcs.AttributesFromFileName` because it effectively only calculates `contentType` and `contentEncoding`.
*

prow/pod-utils/gcs/upload.go
sbueringer: I resolve this as it's handled in: https://github.com/kubernetes/test-infra/pull/16722#discussion_r400704730

sbueringer:
@fejta @alvaroaleman all findings should be addressed, please take another look

pkg/io/opener.go
```diff
@@ -184,25 +185,38 @@ func (o *opener) Reader(ctx context.Context, path string) (io.ReadCloser, error)
...
-func (o *opener) Writer(ctx context.Context, path string) (io.WriteCloser, error) {
-	if strings.HasPrefix(path, "gs://") {
-		g, err := o.openGCS(path)
+func (o *opener) Writer(ctx context.Context, p string, opts ...WriterOption) (io.WriteCloser, error) {
+	if strings.HasPrefix(p, "gs://") {
+		g, err := o.openGCS(p)
 		if err != nil {
 			return nil, fmt.Errorf("bad gcs path: %v", err)
 		}
 		if g != nil {
```
alvaroaleman, line 194: This check shouldn't be needed, `openGCS` should either return an error or an `*ObjectHandle`

prow/pod-utils/gcs/upload.go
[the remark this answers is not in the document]
alvaroaleman: Not sure I understand your comment @fejta , how do we "prove it works" and how does that give us a guarantee that migration GCS over will not result in failures?

I am in favor of migrating it over now, as this is a pretty binary thing: It will work or not, but its unlikely that it will work sometimes. So one way or another, we get very fast feedback.

prow/pod-utils/gcs/upload.go
alvaroaleman: :+1: for only allowing to set the options we actually use

pkg/io/option.go
```diff
@@ -0,0 +1,94 @@
...
+
+import (
+	"cloud.google.com/go/storage"
+	"gocloud.dev/blob"
+	"google.golang.org/api/googleapi"
+)
+
+// A WriterOption is an option for the opener Writer method
+type WriterOption interface {
+	Apply(*storage.Writer, *blob.WriterOptions)
```
alvaroaleman, line 27: This must mention that implementations have to nilcheck both args

pkg/io/option.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+// WithBufferSize returns a WriterOption that specifies the chunk / buffer size
+func WithBufferSize(size int64) WriterOption {
```
alvaroaleman, line 31: This is a question of taste, but I personally prefer the simple option of having a public struct with pointer fields for all options we allow to set that implements `WriterOptions` and then make the `WithSomething` funcs use that struct, because 

* it reduces the amount of types and `WriterOption` implementations that are needed
* It allows to see all possible options quickly by jumping into the definition of that struct, rather than having to jump into all `With` funcs

```
type BlobOption struct {
  Size *int64
  ContentType *string
  ...
}

func (bo *BlobOption) Apply(writer *storage.Writer, o *blob.WriterOptions) {
  if writer != nil {
    if bo.Size != nil {
      writer.ChunkSize = bo.Size
   }
   if bo.ContentType != nil {
      writer.ContentType = bo.ContentType
   }
  }
  if o != nil {
      ...
   }
}

func WithBufferSize (size int64) WriterOption {
   return &BlobOption{Size: utilpointer.Int64(size)}
}
```

prow/gcsupload/run.go
```diff
@@ -17,18 +17,15 @@ limitations under the License.
...
 	"fmt"
 	"mime"
+	"net/url"
 	"os"
 	"path"
 	"path/filepath"
 	"strings"
 
-	"cloud.google.com/go/storage"
 	"github.com/sirupsen/logrus"
```
alvaroaleman, line 28: Please keep our imports in their own group

pkg/io/option.go
alvaroaleman: (Having one public type rather than many `WithSomething` funcs based on distinct private types would also allow us to use that type in the places where we currently use `storage.ObjectAttrs`, e.G. in `prow/pod-utils/gcs/upload.go`)

prow/pod-utils/gcs/upload.go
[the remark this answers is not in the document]
alvaroaleman: ("Now" not necessarily being this PR as its huge and there is a good chance we miss things, but as a follow-up pretty directly after this PR is in and which we can easily revert should we see issues)

prow/gcsupload/run.go
sbueringer: Ups sorry. Of course I'll check all files to be sure. Maybe should fork goimports for that feature :P

pkg/io/opener.go
sbueringer: Yup

pkg/io/option.go
sbueringer: Fixed, but commented on the new Apply method

pkg/io/option.go
sbueringer: Thx, you're right. Your variant is better

prow/pod-utils/gcs/upload.go
sbueringer: I migrated from storage.ObjectAttrs to io.WriterOptions. Looks cleaner imho

prow/gcsupload/run.go
sbueringer: Okay wasn't aware of the -local option of goimports. Hope this fixes it for good..

prow/pod-utils/gcs/upload.go
```diff
@@ -149,57 +183,65 @@ func DataUploadWithAttributes(src io.Reader, attrs *storage.ObjectAttrs) UploadF
...
-
-func (w gcsObjectWriter) ApplyAttributes(attrs *storage.ObjectAttrs) {
-	if attrs == nil {
+func (w gcsObjectWriter) ApplyWriterOptions(opts *pkgio.WriterOptions) {
+	if opts == nil {
 		return
 	}
-	attrs.Name = w.Writer.ObjectAttrs.Name
-	w.Writer.ObjectAttrs = *attrs
+	if opts.BufferSize != nil {
```
alvaroaleman, line 197: Why not simply `opts.Apply(w.Writer)`?

sbueringer: Nice. Just didn't recognize it. Fixed

pkg/io/option.go
```diff
@@ -0,0 +1,87 @@
...
+package io
+
+import (
+	"cloud.google.com/go/storage"
+	"gocloud.dev/blob"
+	"google.golang.org/api/googleapi"
+	utilpointer "k8s.io/utils/pointer"
+)
+
+// WriterOptions are options for the opener Writer method
```
fejta, line 26: Why are these all pointers? This just sounds like a nil pointer headache. The code base prefers non-reference values unless strictly necessary.

pkg/io/option.go
```diff
@@ -0,0 +1,87 @@
...
+type WriterOptions struct {
+	BufferSize      *int64
+	ContentEncoding *string
+	ContentType     *string
+	Metadata        *map[string]string
```
fejta, line 31: Don't use pointers to maps
https://github.com/golang/go/wiki/CodeReviewComments#receiver-type

pkg/io/option.go
```diff
@@ -0,0 +1,87 @@
...
+		if wo.ContentEncoding != nil {
+			writer.ObjectAttrs.ContentEncoding = *wo.ContentEncoding
+		}
+		if wo.ContentType != nil {
+			writer.ObjectAttrs.ContentType = *wo.ContentType
+		}
+		if wo.Metadata != nil {
+			writer.ObjectAttrs.Metadata = *wo.Metadata
+		}
+	}
```
fejta, line 52: return? Then you don't need the indent below

pkg/io/option.go
```diff
@@ -0,0 +1,87 @@
...
+type WriterOptions struct {
+	BufferSize      *int64
+	ContentEncoding *string
+	ContentType     *string
+	Metadata        *map[string]string
+}
+
+// Apply applies the WriterOptions to storage.Writer and blob.WriterOptions
+// Both arguments are allowed to be nil.
+func (wo WriterOptions) Apply(writer *storage.Writer, o *blob.WriterOptions) {
```
fejta, line 36: Oh, probably because you want to do this?

```golang

type WriterOptions struct {
  BufferSize int64
}

// NOTE: receiver should be a pointer here
func (wo *WriterOptions) Apply(writer *storage.Writer, o *blob.WriterOptions) {
  writer.ChunkSize = wo.BufferSize
}
```

pkg/io/option.go
sbueringer: I'm assuming you mean it like this?
```go
if o == nil {
   return 
}
... (everything else without indent)
```

pkg/io/option.go
sbueringer: Now I'm confused :)
It was mostly to differentiate between set/unset values similar to a lot of Kubernetes config structs (e.g. kubeadm) are doing it. I guess it's equivalent

sbueringer: Would just mean it's not possible to set the values to the zero values on purpose but we probably don't need it as setting it to the zero value is equivalent to not setting the option at all.

pkg/io/option.go
sbueringer: My bad, fixed

pkg/io/option.go
sbueringer: Should I just change the pointer fields to non-pointers and check for non-zero values instead of non-nil values?
If yes, should I change the receiver of Apply to a pointer receiver? (if yes, why)


I can also just overwrite the values in writer and blobl.WriterOptions independent of if they are zero values or not and change the parameter of the Writer method to a single WriterOptions struct and not a vararg of WriterOptions. There are currently no cases anymore where we hand over multiple WriterOptions (the With functions are unused). But this would in turn mean we have to handover nil everywhere to the Writer method because it's not a vararg anymore.

fejta: This is different from the pattern used by storage:

https://github.com/googleapis/google-api-go-client/blob/v0.21.0/option/option.go#L17-L45

which would be something more along the lines of:

```golang

type WriterOption interface {
  Apply(writer *storage.Writer, o *blob.WriterOptions) error // I'd recommend allowing returning an error -- aka applying an AWS options to GCS
}

type contentEncoding struct {
  encoding string
}

func (ce contentEncoding) Apply(w *storage.Writer, o *blob.WriterOptions) error {
  if writer != nil { w.ContentEncoding = ce.encoding }
  return nil
}

func WithContentEncoding(encoding string) WriterOption {
  return contentEncoding{encoding}
}

// etc
```

Usage would then look like:

```golang
writer, err := opener.Writer(ctx, path, WithContentEncoding("text/plain"), WithBufferSize(1024*1024))
```

Thoughts on following this pattern?


IMO if you make the Writer signature:
```golang
func (o *opener) Writer(ctx context.Context, p string, opts ...WriterOptions) (io.WriteCloser, error) {
```

Then this decorator pattern makes more sense to me. If you want to do the `if opt.BufferSize != nil`  check, then the `...WriterOptions` seems unnecessary (since you can express everything in a single writeroptions)

alvaroaleman: @fejta he did that because of my comment here: https://github.com/kubernetes/test-infra/pull/16722#discussion_r403507366

alvaroaleman: This boils down to the fact that we have two opts structs which we need to manipulate, so we dont have a single place that contains all possible options anymore. Thats why a struct that has all possible options that are supported on both implementations makes IMHO sense.

The variadic arg makes sense even if you at most pass in one option, because it makes that particular argument optional.

sbueringer: I would like to keep the single struct. @alvaroaleman pointed it out here: https://github.com/kubernetes/test-infra/pull/16722#discussion_r403507366 
I think it's better to have all options in one place. 

Okay so next try :). I replaced the vararg and the pointers in the options struct. Should make more sense now ;)

pkg/io/option.go
```diff
@@ -0,0 +1,53 @@
...
+	if writer != nil {
+		if wo.BufferSize < googleapi.DefaultUploadChunkSize {
+			writer.ChunkSize = int(wo.BufferSize)
+		}
+		writer.ObjectAttrs.ContentEncoding = wo.ContentEncoding
+		writer.ObjectAttrs.ContentType = wo.ContentType
+		writer.ObjectAttrs.Metadata = wo.Metadata
+	}
+
+	if o == nil {
```
fejta, line 45: I'd put the return in the `writer != nil` block. We never pass both sets of options. Is sending both invalid?

pkg/io/option.go
fejta: Alvaro is right that to `Writer(ctx, p string, opts *WriterOptions)` breaks other users of the opener in a way that `...WriterOptions` does not, an makes it non-optional. The `...WriterOptions` seems superior to me. It makes the options actually optional

Can we put the `...` back?

sbueringer: Ah @alvaroaleman didn't saw you're comment. Okay let's find consensus first. I can revert it back again or change it to some compromise in between

sbueringer: btw. the variant with a single struct without pointers would be equivalent of the corresponding signature of gocloud (https://gocloud.dev/howto/blob/#writing)

sbueringer: I think we should decide between:
* single struct with pointers for nil checking which is handed over via varargs
* single struct without pointers which is handed over as a pointer (we would just overwrite all values which are given without checking for nil or zero values)

fejta: I'm fine with using a single struct, but I don't get why it is superior.

```golang
options := []io.WriterOption{
  io.WithContentType("text/plain"), 
  io.WithBufferSize(1024*1024),
}

// func WriterOptionsFromFileName(path string) []io.WriterOption
options = append(options, WriterOptionsFromFileName(path)...)

return opener.Writer(ctx, path, options...)
```

seems at least as clean to me as:

```golang

options := io.WriterOptions{
  ContentType: "text/plain",
  BufferSize: 1024*1024,
}

moreOptions := WriterOptionsFromFileName(path)

return opener.Writer(ctx, path, options, moreOptions) // how do we merge these two?
```

Hypothetically (not relevant now) it makes it more obvious which options are common `io.WithFoo()` and which are provider specific `s3.WithBar()`, `google.WithFrobber()`

pkg/io/option.go
sbueringer: We currently are passing either one or the other. Just not sure if we may handover both in the future. That's why I wanted still be able to handle that case
(nitpick: we're not handing over both options but a writer and/or options where the io.WriterOptions should be applied to)

pkg/io/option.go
sbueringer: Probably a minor difference in that regard yup. Single struct is in exactly the same place in the options.go. The With* functions are distributed across the file. The advantage with s3.WithBar() only works if we create new packages for options of different providers, right?

sbueringer: @fejta Fine for me change back to vararg. This also means reverting back to the pointers, right?

sbueringer:
```go
// how do we merge these two?
```
If we use vararg in both cases, the problem is the same right (if we're only applying values which are set (i.e. != nil)? If we pass in the same option multiple times the naive solution would be last one wins.

fejta: Yes, I would consider `io.WithS3Bar(), io.WithGCSFoo()` and `WriterOptions{S3Bar: bar, GCSFoo: foo}` equivalent.

And yes, vararg and pointers or vararg and WithFoo() then how to merge makes sense in both examples.

sbueringer: I reverted to a single struct with pointers handed over via varargs.

pkg/io/opener.go
```diff
@@ -184,25 +183,36 @@ func (o *opener) Reader(ctx context.Context, path string) (io.ReadCloser, error)
...
 
-	bucket, relativePath, err := o.getBucket(ctx, path)
+	bucket, relativePath, err := o.getBucket(ctx, p)
 	if err != nil {
 		return nil, err
 	}
-	writer, err := bucket.NewWriter(ctx, relativePath, nil)
+	var wOpts *blob.WriterOptions
+	for _, opt := range opts {
+		opt.Apply(nil, wOpts)
```
alvaroaleman, line 213: This is a no-op, because `wOpts` is nil

sbueringer: Time to test it in production in my env :/. Thx I fix it later today

fejta: Exhibit 57 in why pointers are the worst :)

```golang
var wOpts blob.WriterOptions
for _, opt := range opts {
  opt.Apply(nil, &wOps)
}
```

sbueringer: :)

sbueringer:
Okay I hope everything is fixed. I guess time for rebase? :) @fejta @alvaroaleman 

I would test it tomorrow in our env anyway before merging it here.

sbueringer:
> squash and LGTM

done

Issue #16159, opened 2020-02-06
Make Tide check available merge methods on a repo before attempting to merge.


cjwagner:
**What would you like to be added**:
Tide should use [this API](https://developer.github.com/v3/repos/#get) before attempting a merge to check if the requested merge method (merge, squash, or rebase) is allowed by the repo. If it is not, the Tide status should report failure and Tide should ignore the PR until the requested merge method is changed to something valid.

**Why is this needed**:
Currently Tide thinks the PR is mergeable, assumes this is a transient error, and will try to merge this PR every sync loop, potentially preventing the merge of other PRs.

/help
/area prow/tide

PR #16160, opened 2020-02-06
Allow tide merge method labels to be applied with the /label plugin.

cjwagner:
ref: https://github.com/kubernetes/test-infra/issues/16124#issuecomment-582073696
Note that requesting a merge method that is forbidden by the repo's settings will prevent Tide from merging the PR and possibly others in the repo so we may want to address this issue first: https://github.com/kubernetes/test-infra/issues/16159
/hold
/cc @thockin

Issue #10001, opened 2018-11-02
tide: do not stop merging when one PR is unmergable


BenTheElder:
test-infra stopped merging for three hours while trying to squash merge when squash merging was blocked, #9890 #9976 #9997 #9999 #10000 were all passing, but #9890 was first and unmergable

/kind bug
/area prow/tide

cc @cjwagner @stevekuznetsov

BenTheElder:
We can always attempt to merge the older PR first but when a serial merge fails _because the merge failed_ we can attempt to merge another ready serial PR before testing again.

We definitely need to raise this to admins, but there's not a super great mechanism for that right now.

BenTheElder:
I don't think we should *promise* that older PRs are merged first, it should be best effort. This same situation can occur with github API flakes or.... We should not get into a situation where this sits for 3 hours doing nothing.

Even _if_ we had alerted the prow admins, what about a EG a weekend? The prow admins don't necessarily have the same productive hours as the users, we should avoid tide stalling out for any reason.

stevekuznetsov:
Sounds good, let's think through the implications of this best-effort in terms of the current workflow to make sure it doesn't lead to weird states we had not considered before.

stevekuznetsov:
One option -- have `tide` comment on the PR with an explanation of the error and `/hold` to pull the PR out of the queue

cjwagner:
> One option -- have `tide` comment on the PR with an explanation of the error and `/hold` to pull the PR out of the queue

That seems reasonable and is a lot simpler than adding state to `tide` to track failed merges, but there is the down side of not retrying the merge before giving up meaning that transient failures could kick PRs out of the merge pool.
In particular if a repo requires the `tide` status context (which isn't recommended but does have use case) we typically fail to merge the PR on the first attempt because the status controller hasn't set the `tide` status to passing yet. We would need to synchronize Tide's status and sync controllers to avoid the race that causes that issue: https://github.com/kubernetes/test-infra/issues/6289

This also assumes that the `hold` plugin is enabled on the repo.

stevekuznetsov:
`checkconfig` will error if you require the `tide` context and it's considered harmful to do so, so why should we support users that do it?

Good point that this requires the `hold` plugin, I guess we could add a `tide`-specific label that it won't merge into or choose one at random from the list of invalid labels in the query ;)

PR #16726, opened 2020-03-10
Make Tide validate the requested merge method against the allowed methods.

cjwagner:
fixes: https://github.com/kubernetes/test-infra/issues/16159
unblocks: https://github.com/kubernetes/test-infra/pull/16160
PRs that request or configure an invalid merge method for the repo:
- Are filtered from pool so we don’t try to batch or retest them
- Are given a status context that indicates the conflict

The set of valid method per repo is cached and shared between status and sync controllers. The cache is reset once per token reset period to pick up changes to GH settings. (If we want this to be faster for instances that have ghproxy we can make the reset period configurable, but I don't think it matters much since these settings should rarely change for a repo, if ever.)
/cc @alvaroaleman @stevekuznetsov @fejta

### DIFF
--- a/prow/tide/tide.go
+++ b/prow/tide/tide.go
@@ -56,6 +56,7 @@ type githubClient interface {
 	GetCombinedStatus(org, repo, ref string) (*github.CombinedStatus, error)
 	GetPullRequestChanges(org, repo string, number int) ([]github.PullRequestChange, error)
 	GetRef(string, string, string) (string, error)
+	GetRepo(owner, name string) (github.FullRepo, error)
 	Merge(string, string, int, github.MergeDetails) error
 	Query(context.Context, interface{}, map[string]interface{}) error
 }
@@ -84,6 +85,8 @@ type Controller struct {
 	// changedFiles caches the names of files changed by PRs.
 	// Cache entries expire if they are not used during a sync loop.
 	changedFiles *changedFilesAgent
+
+	mergeChecker *mergeChecker
 
 	History *history.History
 }
@@ -232,17 +235,18 @@ func NewController(ghcSync, ghcStatus github.Client, mgr manager, cfg config.Get
 	if err != nil {
 		return nil, fmt.Errorf("error initializing history client from %q: %v", historyURI, err)
 	}
+	mergeChecker := newMergeChecker(cfg, ghcSync)
 
-	sc, err := newStatusController(logger, ghcStatus, mgr, gc, cfg, opener, statusURI)
+	sc, err := newStatusController(logger, ghcStatus, mgr, gc, cfg, opener, statusURI, mergeChecker)
 	if err != nil {
 		return nil, err
 	}
 	go sc.run()
 
-	return newSyncController(logger, ghcSync, mgr, cfg, gc, sc, hist)
+	return newSyncController(logger, ghcSync, mgr, cfg, gc, sc, hist, mergeChecker)
 }
 
-func newStatusController(logger *logrus.Entry, ghc githubClient, mgr manager, gc git.ClientFactory, cfg config.Getter, opener io.Opener, statusURI string) (*statusController, error) {
+func newStatusController(logger *logrus.Entry, ghc githubClient, mgr manager, gc git.ClientFactory, cfg config.Getter, opener io.Opener, statusURI string, mergeChecker *mergeChecker) (*statusController, error) {
 	if err := mgr.GetFieldIndexer().IndexField(&prowapi.ProwJob{}, indexNamePassingJobs, indexFuncPassingJobs); err != nil {
 		return nil, fmt.Errorf("failed to add index for passing jobs to cache: %v", err)
 	}
@@ -252,6 +256,7 @@ func newStatusController(logger *logrus.Entry, ghc githubClient, mgr manager, gc
 		ghc:            ghc,
 		gc:             gc,
 		config:         cfg,
+		mergeChecker:   mergeChecker,
 		newPoolPending: make(chan bool, 1),
 		shutDown:       make(chan bool),
 		opener:         opener,
@@ -267,6 +272,7 @@ func newSyncController(
 	gc git.ClientFactory,
 	sc *statusController,
 	hist *history.History,
+	mergeChecker *mergeChecker,
 ) (*Controller, error) {
 	if err := mgr.GetFieldIndexer().IndexField(
 		&prowapi.ProwJob{},
@@ -287,7 +293,8 @@ func newSyncController(
 			ghc:             ghcSync,
 			nextChangeCache: make(map[changeCacheKey][]string),
 		},
-		History: hist,
+		mergeChecker: mergeChecker,
+		History:      hist,
 	}, nil
 }
 
@@ -384,7 +391,7 @@ func (c *Controller) Sync() error {
 	if err != nil {
 		return err
 	}
-	filteredPools := c.filterSubpools(c.config().Tide.MaxGoroutines, rawPools)
+	filteredPools := c.filterSubpools(c.mergeChecker.isAllowed, rawPools)
 
 	// Notify statusController about the new pool.
 	c.sc.Lock()
@@ -467,12 +474,12 @@ func subpoolsInParallel(goroutines int, sps map[string]*subpool, process func(*s
 // filterSubpools filters non-pool PRs out of the initially identified subpools,
 // deleting any pools that become empty.
 // See filterSubpool for filtering details.
-func (c *Controller) filterSubpools(goroutines int, raw map[string]*subpool) map[string]*subpool {
+func (c *Controller) filterSubpools(mergeAllowed func(*PullRequest) (string, error), raw map[string]*subpool) map[string]*subpool {
 	filtered := make(map[string]*subpool)
 	var lock sync.Mutex
 
 	subpoolsInParallel(
-		goroutines,
+		c.config().Tide.MaxGoroutines,
 		raw,
 		func(sp *subpool) {
 			if err := c.initSubpoolData(sp); err != nil {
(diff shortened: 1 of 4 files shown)


prow/tide/tide.go
```diff
@@ -573,6 +581,90 @@ func filterPR(ghc githubClient, sp *subpool, pr *PullRequest) bool {
...
+}
+
+func (m *mergeMethodChecker) clearCache() {
+	// Only do this once per token reset since it could be a bit expensive for
+	// Tide instances that handle hundreds of repos.
+	ticker := time.NewTicker(time.Hour)
+	for {
+		<-ticker.C
+		m.Lock()
+		m.cache = make(map[config.OrgRepo]map[github.PullRequestMergeType]bool)
```
alvaroaleman: ghproxy already knows how to cache and can also check if its cache is still valid. Why do we need a second caching layer here?

prow/tide/tide.go
```diff
@@ -573,6 +581,90 @@ func filterPR(ghc githubClient, sp *subpool, pr *PullRequest) bool {
...
+			github.MergeMerge:  fullRepo.AllowMergeCommit,
+			github.MergeSquash: fullRepo.AllowSquashMerge,
+			github.MergeRebase: fullRepo.AllowRebaseMerge,
+		}
+		m.cache[orgRepo] = repoMethods
+	}
+	return repoMethods, nil
+}
+
+func (m *mergeMethodChecker) isAllowed(pr *PullRequest) error {
```
alvaroaleman: Lets please change the return of this to properly distinguish "is merge method allowed" from "an error occurred when checking". The caller can then still decide to drop the PR if either of the two is not as expected.

prow/tide/tide.go
```diff
@@ -573,6 +581,90 @@ func filterPR(ghc githubClient, sp *subpool, pr *PullRequest) bool {
...
+	log := m.log.WithFields(pr.logFields())
+	if pr.Mergeable == githubql.MergeableStateConflicting {
+		return fmt.Errorf("PR has a merge conflict")
```
alvaroaleman: If this also checks for conflicts, maybe rename the whole thing to `mergeChecker`? Having this check in the `mergeMethodChecker` is for me pretty unexpected.

prow/tide/tide.go
cjwagner: If we don't do any in-process caching this would require 1-2 extra API requests every sync period for every PR that matches a Tide query and an additional request whenever an open PR outside the tide pool is updated. Thats a pretty significant token increase that would make ghproxy required for anything, but small Tide instances if other Prow components are running.
We've been trying to keep ghproxy as an optional component that only large Prow instances need to use, but perhaps it is time to reconsider that. It seems relatively reliable and would make token management easier. We should consider making it HA before we do that though.

prow/tide/tide.go
cjwagner: SGTM, I couldn't think of a good name. `mergeChecker` seems a little vague since it isn't clear what all it checks about mergeability, but it is certainly more sane and correct than what it is now. I'll make sure the go docs are clear.

prow/tide/tide.go
alvaroaleman: So the reason I asked about that is that I dislike both the added complexity in the code itself and when trying to debug this. Also, I think this is pretty subpar for users as well, because if someone does change their allowed mergeMethods, they have to wait for up to one hour without getting any feedback on why Tide tells them their newly-configured mergeMethod wasn't allowed.

Maybe a middleground would be to just cache during the sync?

> We've been trying to keep ghproxy as an optional component that only large Prow instances need to use, but perhaps it is time to reconsider that. It seems relatively reliable and would make token management easier. 

Yeah, IMHO this would simplify things a lot because we do not have to think anymore about how many requests we do, but just about how many requests we do that may not be served from cache. Also, I don't think we even have data on the token usage of a small prow without ghproxy, so we can not even say if we support that usecase well. On top of that even when keeping the token issue aside, ghproxy also speeds things up.

> We should consider making it HA before we do that though.

Definitely +1 for making it HA (we had a brief outage yesterday due to it going down) but I don't think that should be a requirement for making it mandatory.

cjwagner: > Also, I think this is pretty subpar for users as well, because if someone does change their allowed mergeMethods, they have to wait for up to one hour without getting any feedback on why Tide tells them their newly-configured mergeMethod wasn't allowed.

Thats fair, but as noted in the PR description this rarely, if ever, changes so I figured a delay was worth the benefits (especially if we made the delay configurable). It is also unclear to me why this would be difficult to debug. I think the logs and errors are informative and debugging a single process would be easier than debugging the interaction with ghproxy.

> I don't think we even have data on the token usage of a small prow without ghproxy, so we can not even say if we support that usecase well.

Well I don't have hard data from a small prow, but given that k8s Prow (very large) worked without ghproxy and that we have been careful to avoid relying on its existence, I don't know what would have changed to make it impossible to run Tide without ghproxy at reasonable scale.

I agree that assuming ghproxy is in use simplifies the code. But it does so in exchange for extra network traffic (not a big issue, but more errors), a hard requirement to use ghproxy (maybe not an issue, but a breaking change), and a single point of failure (we fall back to using the GH API directly, but that would cause us to quickly exhaust tokens without caching in process). The last point is the main one that concerns me, especially if it caused an outage for you as recently as yesterday.

I'm happy to change this if we want to start requiring ghproxy, but I'd like to get this functionality in and not block it on that policy change (since it is a breaking change we need a warning period where we warn if it looks like components are using the GH API directly). There are a number of places where we do similar in-process caching because we don't want to assume ghproxy exists so this isn't a problem or pattern unique to this PR.

alvaroaleman: > Thats fair, but as noted in the PR description this rarely, if ever, changes so I figured a delay was worth the benefits (especially if we made the delay configurable). It is also unclear to me why this would be difficult to debug. I think the logs and errors are informative and debugging a single process would be easier than debugging the interaction with ghproxy.

Well, because if an allowed merge method is refused due to cache staleness, ppl essentially have to look at the code and find this particular cache in order to understand why that happens.

> There are a number of places where we do similar in-process caching because we don't want to assume ghproxy exists so this isn't a problem or pattern unique to this PR.

Do you have an example at hand for this? I am not aware of other places where there is up to an hour waiting time for a cache to not be stale anymore.

cjwagner: I don't have an example where we only update the cache once a sync period, but that is because every other cache is for relatively dynamic data, not something as static as the allowed merge methods which don't ever change for most repos. It is certainly possible that someone could change the allowed merge methods and be confused why there is a delay before Tide reacts, but I don't think that will be a significant problem in practice. Currently we see entire repos get blocked by this and typically noone notices until we see the errors in the Tide logs.

More importantly, even if there is a one time delay if settings are changed this is still a big improvement because it unblocks the repos and I'd rather not wait months for a warning period to enable ghproxy when we could merge this today.
I agree that this isn't perfect and requiring ghproxy is the end goal, but wouldn't we rather have this in the meantime? In particular I'm trying to make this PR safe: https://github.com/kubernetes/test-infra/pull/16160

fejta: Do we know anyone running prow without ghproxy?

prow/tide/tide.go
```diff
@@ -573,6 +584,86 @@ func filterPR(ghc githubClient, sp *subpool, pr *PullRequest) bool {
...
+		}
+		m.cache[orgRepo] = repoMethods
+	}
+	return repoMethods, nil
+}
+
+// isAllowed checks if a PR does not have merge conflicts and requests an
+// allowed merge method. If there is no error it returns a string explanation if
+// not allowed or "" if allowed.
+func (m *mergeChecker) isAllowed(pr *PullRequest) (string, error) {
```
fejta: Why can't we simply react to the failed merge? AKA if a merge fails, then paste a comment on the PR about why (and blacklist the PR until it is updated).

This strategy has transparency (and therefore educational) advantages. I suspect the design will be simpler and consume fewer tokens.

cjwagner: Tide uses status contexts to convey information about the state of the PR. We could use a comment instead for this case, but I think it makes more sense to use the status so we don't add a comment that contradicts the status context. We also don't want to waste time retesting a PR that won't be mergeable.
We don't have a mechanism for blacklisting problematic PRs, there is a separate issue for that: https://github.com/kubernetes/test-infra/issues/10001

It is unclear to me how that strategy has more transparency than this. I think displaying a status that immediately indicates the PR has an invalid merge method is more transparent than waiting until Tide tries to merge the PR. We would definitely consume fewer tokens listing allowed merge methods, but we would waste time and potentially other tokens by considering and retesting PRs that won't be able to merge.

prow/tide/tide.go
cjwagner: Yes. Looks like OSS Prow doesn't have it running and Istio didn't have ghproxy until last July. Looking around at some other instances we know of I see that Knative, Azure aks-engine, and GoogleCloudPlatform/compute-image-tools instances also do not have it deployed.

prow/tide/tide.go
alvaroaleman: What about https://github.com/kubernetes/test-infra/issues/10001#issuecomment-455638750 if we hit that error?

Just posting `/hold` means we do not really need to add a blacklisting mechanism, we just re-use what we already have.  The comment could then also include a the info that the `/hold` should be removed if the PRs or repos merge method was updated.

cjwagner: Thats a good idea, but I think there are still too many downsides. We would still waste time retesting an unmergeable PR and most importantly we could get into an untested state if that PR was part of a batch, but we aren't able to merge it and accidentally merged a partial batch.

This also assumes that hook is deployed and the hold plugin is enabled. We could give Tide its own version of the hold label as Steve suggested, but then Tide would need to scrape comments or handle webhooks in order to expose a way for users to remove the label.

It seems cleanest to me to make Tide aware of this merge requirement just like the other merge requirements so that it can display it consistently with the others on the status context and properly handle the PR instead of retesting hopelessly and adding logic to recover from preventable failures.

prow/tide/tide.go
alvaroaleman: Okay, so i continue to not really like this since understanding why a PR gets refused while the cache is stale basically requires to be aware of this cache which none of our users will be, but I do realize that having tide blocked by an unmergeable PR is the bigger evil. Can we create an issue to have this cache invalidated after each sync loop after may first (xref https://github.com/kubernetes/test-infra/pull/16763)?

/lgtm
/approve
/hold

cjwagner: Opened https://github.com/kubernetes/test-infra/issues/16800.
I definitely agree that this is the lesser of two evils, but still an evil and I am eager to exorcise when we have all Prow's using ghProxy.
Thanks for the review Alvaro!

PR #16732, opened 2020-03-11
Remove references to prowjobs/status subresource

alvaroaleman:
We do not activate the status subresource in our CRD definitions, nor do
we have any code that interacts with it.


alvaroaleman:
Yeah this applies immediately, but we do not have the status subresource enabled nor any code that tries to interact with it (it would fail even if rbac allows it).
/hold cancel

PR #16737, opened 2020-03-11
Update sig-windows-config.yaml

adelina-t:
Add K8s 1.18 parallel & serial conformance jobs for Windows.


adelina-t:
I'm actually wondering if maybe we should add the job more often than the 8 hour interval.
Don't know what implications that has for resource capacity.

marosset:
actually - should we add these to https://testgrid.k8s.io/sig-release-1.18-informing?

marosset:
It looks like the fast one is but the serial-slow one is not - I guess that is ok.

marosset:
I think @chewong was saying we were running out of public IP addresses in the Azure sub we use for these test passes so.

adelina-t:
Yeah, I've slow down the containerd + flannel runs & have a fix in the works to reduce the number of Public IPs used there

PR #15582, opened 2019-12-12
it's clap clap now 👏👏

BenTheElder:
https://twitter.com/ipedrazas/status/1205077331695603713

/hold
/approve cancel
/cc @munnerz @justaugustus

PR #16738, opened 2020-03-11
Revert/remove clap clap functionality in assign plugin

cblecker:
This functionality is problematic on a number of levels and shouldn't have been approved.

First, it's weird, undocumented behaviour. If someone is not in the joke, it will be confusing as to why emoji are triggering a review request. There is nothing in the plugin's documentation or help page saying that this is an option.

Second, it's an easter egg in a core command. It's not an optional plugin that can be turned on or off.. most prow users use the assign plugin, and there is no flag or option to disable this undocumented behaviour.

Third, this messes with [devstats](https://github.com/cncf/devstats/blob/dc4f67ac9db47600ce444da4d024c294a4e85e7e/metrics/kubernetes/bot_commands.sql#L9) where we collect data on bot commands that are issued.

And lastly, it's non-inclusive. Beyond the issues raised in the first point that it's in undocumented inside joke, it also uses emojis in a command which messes with screen readers and other accessibility bits. It's just (unfortunately) not really a great idea from a contributor experience standpoint. I'm happy to see that are tests for skin colours, but unfortunately that's not enough.

To be clear: I like memes and jokes, and feel there is room for fun. Great examples of this are the dog/cat/pony/goose plugins (they have full help documentation, are optional, and don't mess with the actual PR workflow). Unfortunately, these inclusivity considerations are important, and any changes to prow that impact that contributor experience need to be run through sig-contribex.

FYI @kubernetes/sig-contributor-experience @kubernetes/sig-testing @kubernetes/test-infra-maintainers

This reverts commit 46626dea0bee533ddec96ee1f6e39a93bd9308ec, reversing
changes made to fd45432efbe1eecb848d9a87127a61db4f2eeee3.

### DIFF
--- a/prow/plugins/assign/assign.go
+++ b/prow/plugins/assign/assign.go
@@ -34,7 +34,7 @@ const pluginName = "assign"
 var (
 	assignRe = regexp.MustCompile(`(?mi)^/(un)?assign(( @?[-\w]+?)*)\s*$`)
 	// CCRegexp parses and validates /cc commands, also used by blunderbuss
-	CCRegexp = regexp.MustCompile(`(?mi)^/(un)?(?:cc|👏[\x{1f3fb}-\x{1f3ff}]?👏[\x{1f3fb}-\x{1f3ff}]?)(( +@?[-/\w]+?)*)\s*$`)
+	CCRegexp = regexp.MustCompile(`(?mi)^/(un)?cc(( +@?[-/\w]+?)*)\s*$`)
 )
 
 func init() {
--- a/prow/plugins/assign/assign_test.go
+++ b/prow/plugins/assign/assign_test.go
@@ -295,18 +295,6 @@ func TestAssignAndReview(t *testing.T) {
 			requested: []string{"cjwagner"},
 		},
 		{
-			name:      "clap clap works too",
-			body:      "/👏👏 @cjwagner ",
-			commenter: "rando",
-			requested: []string{"cjwagner"},
-		},
-		{
-			name:      "clap clap with skin tone works too",
-			body:      "/👏🏽👏🏽 @cjwagner ",
-			commenter: "rando",
-			requested: []string{"cjwagner"},
-		},
-		{
 			name:        "multi commands",
 			body:        "/cc @cjwagner\n/uncc @spxtr",
 			commenter:   "rando",


BenTheElder:
> First, it's weird, undocumented behaviour. If someone is not in the joke, it will be confusing as to why emoji are triggering a review request. There is nothing in the plugin's documentation or help page saying that this is an option.

documenting this does not require reverting the plugin

> Second, it's an easter egg in a core command. It's not an optional plugin that can be turned on or off.. most prow users use the assign plugin, and there is no flag or option to disable this undocumented behaviour.

I'd be happy to make it optional, and I'd note that a maintainer of another prow instance outside Kubernetes lgtmed...

> Third, this messes with devstats where we collect data on bot commands that are issued.

devstats has existing data holes, we can fix this

> And lastly, it's non-inclusive. Beyond the issues raised in the first point that it's in undocumented inside joke, it also uses emojis in a command which messes with screen readers and other accessibility bits. It's just (unfortunately) not really a great idea from a contributor experience standpoint. I'm happy to see that are tests for alternative skin colours, but unfortunately that's not enough.

I'm not familiar with emoji and screen readers but this is hardly the only place we use emoji, and there is still the github native assignment event.

So far this is used sparingly, not as the defacto standard for CCing.

Again, documentation is a simple PR.

> To be clear: I like memes and jokes, and feel there is room for fun. Great examples of this are the dog/cat/pony/goose plugins (they have full help documentation, are optional, and don't mess with the actual PR workflow). Unfortunately, these inclusivity considerations are important, and any changes to prow that impact that contributor experience need to be run through sig-contribex.

I'd like to hear more about these beyond just "it's not documented". Lack of documentation has never stopped our tools, and documenting this is a much smaller change.

/hold

BenTheElder:
I also imagine that "/cc [@]user" => assign for review is no more intuitive without reading https://prow.k8s.io/command-help which probably has it's own issues. 

Googlers may be used to CC on critique (though the meaning is not the same!) and some may guess from email carbon copy but it's hardly an actual standard, both are something we've bolted onto github.

WRT screenreaders (AIUI reading now) they will read the descriptive name so we get something along the lines of "slash clapping hands clapping hands at-sign munnerz" versus "slash cee cee clapping hands at-sign munnerz"

From [the first post I found on this](https://blog.easterseals.com/emojis-and-accessibility-the-dos-and-donts-of-including-emojis-in-texts-and-emails/):
- Do use one or two emojis if you like, most blind people get a kick out of the descriptions;

If emoji are a concern, should we banish emoji usage elsewhere? What about slack? Email? Release updates / CI signal?

BenTheElder:
Upon further reflection this is not worth wasting anyone's time. I'm sure we could be doing something useful.
/hold cancel
/un👏👏@bentheelder

nikhita:
While this would have been fun to have on, I agree that having easter eggs in core workflow commands could end up being confusing for new/not-so-active contributors. Documentation could help us bridge the gap half way... though given how overwhelming our automation is to new contributors already, IMHO adding more complexity in core commands could end up making it even more mystifying. 

Sorry for the revert-dance on this, Ben. I know that there was enthusiastic support on this in https://github.com/kubernetes/test-infra/pull/15582 (it's fun for sure!) and getting boggled down by the process blows. :heart: :pensive: 

I'll take it as a follow-up to see how we can document information about adding such fun commands/easter eggs and make it discoverable so we won't need to go through this again.

/lgtm

Katharine:
I do not particularly like this alias, but the way in which it was removed is _incredibly_ frustrating to me in several dimensions.

PR #16742, opened 2020-03-11
jobs: move gce-cos-master-reboot to release-master-blocking

spiffxp:
The release branch equivalent is present in all other release-blocking
dashboards, eg:
- https://testgrid.k8s.io/sig-release-1.18-blocking#gce-cos-k8sbeta-reboot
- https://testgrid.k8s.io/sig-release-1.17-blocking#gce-cos-k8sstable1-reboot
- https://testgrid.k8s.io/sig-release-1.16-blocking#gce-cos-k8sstable2-reboot
- https://testgrid.k8s.io/sig-release-1.15-blocking#gce-cos-k8sstable3-reboot

ref: https://github.com/kubernetes/test-infra/issues/9363

### DIFF
--- a/config/jobs/kubernetes/sig-cloud-provider/gcp/gcp-gce.yaml
+++ b/config/jobs/kubernetes/sig-cloud-provider/gcp/gcp-gce.yaml
@@ -571,9 +571,11 @@ periodics:
       - --timeout=180m
       image: gcr.io/k8s-testimages/kubekins-e2e:v20200311-1e25827-master
   annotations:
-    testgrid-dashboards: sig-release-master-informing, google-gce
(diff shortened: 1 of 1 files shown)


spiffxp:
So it looks like this job has no owner in the form of an e-mail address to send alerts to.  The test that checks this is only checking against master, hence why the release-branch jobs aren't failing.

I have two options:
- demote the job to release-informing everywhere
- find contact info for sig-cloud-provider/cloud-provider-gcp

spiffxp:
I checked with the author of the last PR that fixed this job (https://github.com/kubernetes/kubernetes/pull/88099) and have updated with an alert e-mail address

justaugustus:
Thanks @spiffxp!
/lgtm
/approve

PR #16753, opened 2020-03-12
Add oxddr and jkaniuk as owners for scalability tests.

oxddr:
Add remaining scalability oncallers, so they are not blocked by lack of approvals if urgent changes are required. 

/assign @wojtek-t


wojtek-t:
As mentioned by @wasylkowski-a during our internal meeting, we not necessary want to merge changes to tests (especially for those release blocking ones) without deep enough review (which I fully agree with). Given there are already 3 people from our team (and Shyam from outside Google, and worst case higher-level owners), I don't think we need everyone to be there.

/close

PR #16756, opened 2020-03-12
Disable node-killer in for release-1.{16,17,18} branches.

oxddr:
This should be reverted once we understand why it caused tests to fail. 

/assign @wojtek-t 
ref https://github.com/kubernetes/kubernetes/issues/89051


wojtek-t:
Seems that it helped in master - 5 greens in a row.

/lgtm
/approve

PR #16760, opened 2020-03-12
Add managed webhook/HMAC config and associated validation.

agarg2008:
Design doc https://docs.google.com/document/d/1we2DgZd42U0PRhB-XsvP2FJjLHX5dzJPEbXC6gqWkn0/edit#  

This is part 1 of the change which would allow us to have a declarative file which tracks which org/repo are using private hmac tokens for github integrations.

### DIFF
--- a/prow/config/config.go
+++ b/prow/config/config.go
@@ -44,6 +44,7 @@ import (
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/validation"
+	"k8s.io/test-infra/prow/errorutil"
 	"sigs.k8s.io/yaml"
 
 	prowapi "k8s.io/test-infra/prow/apis/prowjobs/v1"
@@ -153,6 +154,10 @@ type ProwConfig struct {
 	// DefaultJobTimeout this is default deadline for prow jobs. This value is used when
 	// no timeout is configured at the job level. This value is set to 24 hours.
 	DefaultJobTimeout *metav1.Duration `json:"default_job_timeout,omitempty"`
+
+	// ManagedWebhooks contains information about all github repositories and organizations which are using
+	// non-global Hmac token.
+	ManagedWebhooks ManagedWebhooks `json:"managed_webhooks,omitempty"`
 }
 
 type InRepoConfig struct {
@@ -666,6 +671,14 @@ type GitHubOptions struct {
 	LinkURL *url.URL `json:"-"`
 }
 
+// ManagedWebhookInfo contains metadata about the repo/org which is onboarded.
+type ManagedWebhookInfo struct {
+	TokenCreatedAfter time.Time `json:"tokenCreatedAfter"`
+}
+
+// ManagedWebhooks contains information about all the repos/orgs which are onboraded with private tokens.
+type ManagedWebhooks map[string]ManagedWebhookInfo
+
 // SlackReporter represents the config for the Slack reporter. The channel can be overridden
 // on the job via the .reporter_config.slack.channel property
 type SlackReporter struct {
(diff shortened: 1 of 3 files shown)


prow/cmd/checkconfig/main.go
```diff
@@ -324,6 +333,20 @@ func main() {
 
 	logrus.Info("checkconfig passes without any error!")
 }
+
+func validateManagedHmacEntityConfig(prowConfig config.ProwConfig) error {
+	var validationErrs []error
+
+	if prowConfig.ManagedHmacEntity != nil {
```
cjwagner: This should be included in [core Prow config validation](https://github.com/kubernetes/test-infra/blob/b47e83d8b2752d38d18bc35bbc2ec532d247df59/prow/config/config.go#L1117). `checkconfig` does the core validation when loading the config then optionally performs additional lint checks. Since this is a validity check, not a lint check, it belongs in the core config validation.

prow/config/config.go
```diff
@@ -153,6 +153,10 @@ type ProwConfig struct {
 	// DefaultJobTimeout this is default deadline for prow jobs. This value is used when
 	// no timeout is configured at the job level. This value is set to 24 hours.
 	DefaultJobTimeout *metav1.Duration `json:"default_job_timeout,omitempty"`
+
+	// ManagedHmacEntity contains information about all github repositories and organizations which are using
+	// non-global Hmac token.
+	ManagedHmacEntity ManagedHmacEntity `json:"managed_hmac_entity,omitempty"`
```
cjwagner: nit: Something like `managed_webhooks` might be a little clearer. GH webhooks are what is really being managed, the HMAC is just a part of that.
Currently there isn't really an indication that this is related to GH webhooks.

prow/config/config.go
```diff
@@ -666,6 +670,14 @@ type GitHubOptions struct {
 	LinkURL *url.URL `json:"-"`
 }
 
+// ManagedHmacEntityInfo contains information about the repo/org which is onboarded and associated metadata.
+type ManagedHmacEntityInfo struct {
+	TokenCreatedAfter time.Time `json:"tokenCreatedAfter"`
+}
+
+// ManagedHmacEntity contains information about all the repos/orgs which are onboraded with private tokens.
+type ManagedHmacEntity map[string]ManagedHmacEntityInfo
```
cjwagner: nit: If you are going to create a type for this please pluralize it.

prow/config/config.go
```diff
@@ -666,6 +670,14 @@ type GitHubOptions struct {
 	LinkURL *url.URL `json:"-"`
 }
 
+// ManagedHmacEntityInfo contains information about the repo/org which is onboarded and associated metadata.
+type ManagedHmacEntityInfo struct {
+	TokenCreatedAfter time.Time `json:"tokenCreatedAfter"`
```
cjwagner: ~~FYI, if we use `time.Time` here instead of a string that we deserialize ourselves this will be the time format:
`2009-11-10T23:00:00Z` https://play.golang.org/p/lpWRnIuWZ-X~~

Just saw that matches your DD so 👍

prow/config/config.go
agarg2008: Done.

prow/config/config.go
agarg2008: Done.

prow/config/config.go
```diff
@@ -153,6 +154,10 @@ type ProwConfig struct {
 	// DefaultJobTimeout this is default deadline for prow jobs. This value is used when
 	// no timeout is configured at the job level. This value is set to 24 hours.
 	DefaultJobTimeout *metav1.Duration `json:"default_job_timeout,omitempty"`
+
+	// ManagedWebhooks contains information about all github repositories and organizations which are using
+	// non-global Hmac token.
+	ManagedWebhooks ManagedWebhooks `json:"managed_hmac_entity,omitempty"`
```
cjwagner:
```suggestion
	ManagedWebhooks ManagedWebhooks `json:"managed_webhooks,omitempty"`
```

cjwagner: Otherwise LGTM.

agarg2008: I am sure I made this change! not sure how it got reverted.

PR #16761, opened 2020-03-12
Azure: change aks-engine download URL for windows 1.18 jobs

chewong:
Since `release-1.18` included the kubectl breaking change (https://github.com/Azure/aks-engine/issues/2845), we will need to point the aks-engine download URL of windows 1.18 jobs to a custom-built aks-engine tarball.

/assign @marosset


marosset:
We should point this back to an aks-engine release once a new release comes out with these fixes. (i think v0.48.0 should be coming out any day now)

PR #16763, opened 2020-03-13
Improve ghProxy docs and make Prow warn if ghproxy isn't being used.

cjwagner:
The first commit adds ghProxy docs about how to use with Prow.

The second commit makes the GH client warn if it doesn't look like ghProxy is being used and sets an ultimatum of May 1, 2020 when Prow components may start assuming ghProxy is used. This assumption will let us avoid a lot of in-process caching which will simplify code.
Holding for approval from some other [Prow OWNERS](https://github.com/kubernetes/test-infra/blob/master/prow/OWNERS).
/hold 

/assign @fejta @stevekuznetsov @alvaroaleman @cblecker

### DIFF
--- a/ghproxy/README.md
+++ b/ghproxy/README.md
@@ -1,6 +1,34 @@
 # ghProxy
 
 ghProxy is a reverse proxy HTTP cache optimized for use with the GitHub API (https://api.github.com).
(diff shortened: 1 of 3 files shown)


matthyx:
nice thing!
/lgtm
/hold

PR #16772, opened 2020-03-13
GHE support for clonerefs

sbueringer:
clonerefs currently always hardcodes github.com. This PR fixes that by parsing the RepoLink 
(which is e.g.: `https://github.com/kubernetes/test-infra`) and uses the Host of the URL instead.


sbueringer:
Apparently repoLink is not set in every case. So my current implementation won't work. Does anybody have an idea how to fix this? :)

sbueringer:
Let's close this for now. Maybe I'll open another PR if I have a clean solution for this problem

PR #16783, opened 2020-03-14
ghproxy: record the user agent on cache response metrics

stevekuznetsov:
Often it is not enough to see which user agent is making requests to
GitHub through the cache, as those requests may be cached and not
responsible for token exhaustion. It is therefore necessary to be able
to track cache responses by user agent.

### DIFF
--- a/ghproxy/ghcache/coalesce.go
+++ b/ghproxy/ghcache/coalesce.go
@@ -129,7 +129,7 @@ func (r *requestCoalescer) RoundTrip(req *http.Request) (*http.Response, error) 
 		return resp, nil
 	}()
 
-	ghmetrics.CollectCacheRequestMetrics(string(cacheMode), req.URL.Path)
+	ghmetrics.CollectCacheRequestMetrics(string(cacheMode), req.URL.Path, req.Header.Get("User-Agent"))
 	if resp != nil {
 		resp.Header.Set(CacheModeHeader, string(cacheMode))
 	}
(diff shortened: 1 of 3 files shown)


ghproxy/ghcache/coalesce.go
```diff
@@ -129,7 +129,7 @@ func (r *requestCoalescer) RoundTrip(req *http.Request) (*http.Response, error)
 		return resp, nil
 	}()
 
-	ghmetrics.CollectCacheRequestMetrics(string(cacheMode), req.URL.Path)
+	ghmetrics.CollectCacheRequestMetrics(string(cacheMode), req.URL.Path, req.Header.Get("User-Agent"))
```
cjwagner: Should we strip out the version and just use the component name to avoid creating too many prometheus label values? I don't know what the scale limits of prometheus are in terms of number of label values, but I know that was a concern with the request path metrics.

stevekuznetsov: Hm, good question. Since we generally don't expect to have separate versions of e.g. `hook` running at once it may be fine to do. Over a four week span I expect to see ~80 (2x/day * 5day/week * 4 weeks) distinct versions of a component, so the cardinality is in the hundreds for all components -- generally I believe that to be fine, it's when we go into thousands and up that we see performance impacts.

stevekuznetsov:
Since we do not really expect to have one `ghproxy` deployment fielding requests from e.g. two versions of one plugin at once, we do not gain much by having the versions in there anyway. I've removed them, this should be ready to go.

PR #16787, opened 2020-03-14
Kops/gce: promote gce-latest to optional presubmit

geojaz:
Promotes kops-gce-latest to presubmit so we can start to use the signal.
Also turns other kops/gce e2e on and adjusts various args.


geojaz:
/hold
it's not useful to have this promoted until we get the bucket ACL issues figured out

PR #16788, opened 2020-03-14
update open-bug template to go to bugzilla for openshift 4.5 jobs

bparees:


bparees:
/assign @smarterclayton @stevekuznetsov 

i've got no way of testing this that i know of, so let's just give it a shot?  i doubt anyone's using the existing functionality.

if we like the results i'll do PRs for our other dashboards.

PR #16824, opened 2020-03-17
verify-owners: use a cache-able call

stevekuznetsov:

### DIFF
--- a/prow/github/client.go
+++ b/prow/github/client.go
@@ -2966,6 +2966,23 @@ func (c *client) Merge(org, repo string, pr int, details MergeDetails) error {
 //
 // See https://developer.github.com/v3/repos/collaborators/
 func (c *client) IsCollaborator(org, repo, user string) (bool, error) {
+	// This call does not support etags and is therefore not cacheable today
+	// by ghproxy. If we can detect that we're using ghproxy, however, we can
+	// make a more expensive but cache-able call instead. Detecting that we
+	// are pointed at a ghproxy instance is not high fidelity, but a best-effort
+	// approach here is guaranteed to make a positive impact and no negative one.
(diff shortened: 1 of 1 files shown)


prow/plugins/trigger/trigger.go
```diff
@@ -155,10 +155,14 @@ func handlePush(pc plugins.Agent, pe github.PushEvent) error {
 func TrustedUser(ghc trustedUserClient, onlyOrgMembers bool, trustedOrg, user, org, repo string) (bool, error) {
 	// First check if user is a collaborator, assuming this is allowed
 	if !onlyOrgMembers {
-		if ok, err := ghc.IsCollaborator(org, repo, user); err != nil {
+		if users, err := ghc.ListCollaborators(org, repo); err != nil {
```
alvaroaleman: Can we not use this terrible notation of using `else` after error checks? We return in the `if` branch, we do not need to make the `else` branch

prow/plugins/trigger/trigger.go
```diff
@@ -155,10 +155,14 @@ func handlePush(pc plugins.Agent, pe github.PushEvent) error {
 func TrustedUser(ghc trustedUserClient, onlyOrgMembers bool, trustedOrg, user, org, repo string) (bool, error) {
```
cjwagner: This can be called multiple times per handled event right? This seems like it could be prohibitively expensive unless ghProxy is enabled which we can't yet assume. Instead of working around this can we first see if GH will make their API properly generate Etags for the isCollaborator call?

stevekuznetsov: So this would increase token use for repos where this call will scale with the number of pages it needs to read, but I don't see an alternative right now.

PR #16826, opened 2020-03-17
Add sample description to e2e-aws-4.4 job

bparees:
also includes fixing up the BZ template for openshift jobs

generated from https://github.com/openshift/release/pull/7728


bparees:
@smarterclayton ready for merge i think, if the goal is just to see how this shows up in testgrid

PR #16834, opened 2020-03-18
Kops - Retry 5 times to get the external IP

hakman:
Some e2e tests are failing randomly because the external IP cannot be retrieved. Retrying 5 times should fix this.

For reference, a failed e2e test:
https://storage.googleapis.com/kubernetes-jenkins/logs/e2e-kops-aws-k8s-latest/1240180113656516609/build-log.txt


kubetest/kops.go
```diff
@@ -414,12 +414,20 @@ func (k kops) Up() error {
...
+		count := 0
+		for count < 5 && err != nil && net.ParseIP(strings.TrimSpace(b.String())) == nil {
+			time.Sleep(2 * time.Second)
+			count++
 			b.Reset()
-			if err := httpRead("https://v4.ifconfig.co", &b); err != nil {
-				return err
-			}
+			err = httpRead("https://v4.ifconfig.co", &b)
+			log.Printf("external IP cannot be retrieved: %v\n", err)
```
rifelpet: this would be printed even if the call succeeded. maybe put it in an `if err != nil` although we learned that this doesnt cover all errors (hence the `net.ParseIP` checking) so maybe we just get rid of this log line.

kubetest/kops.go
```diff
@@ -414,12 +414,20 @@ func (k kops) Up() error {
...
 		}
+
+		if err != nil || net.ParseIP(strings.TrimSpace(b.String())) == nil {
+			return fmt.Errorf("external IP cannot be retrieved: %v", err)
```
rifelpet: likewise, we could be failing here even if `err` is nil so maybe we can elaborate on the Errorf to include those cases as well. perhaps including `b.String()` in the error message.

kubetest/kops.go
```diff
@@ -414,12 +414,19 @@ func (k kops) Up() error {
...
+
+		count := 0
+		for count < 5 && err != nil && net.ParseIP(strings.TrimSpace(b.String())) == nil {
+			time.Sleep(2 * time.Second)
+			count++
 			b.Reset()
-			if err := httpRead("https://v4.ifconfig.co", &b); err != nil {
-				return err
-			}
+			err = httpRead("https://v4.ifconfig.co", &b)
```
tanjunchen: how about show fail log count in here?

hakman: I don't think it's so important. If it fails the 5th time, the error will probably be the same and will be displayed later.
If it doesn't fail, no one will look for the counter.

tanjunchen: Sure . get it 
/LGTM

hakman: Thanks @tanjunchen

hakman:
@krzyzacy @feiskyer Any chance you can take a look at this? Thanks!

PR #16839, opened 2020-03-18
Jenkins controller: Use patch

alvaroaleman:
When using crier, another actor is present in the system that may change
jobs, which results in update failures. We assume that the fields we
care about are different from the ones crier cares about.

Fixes https://github.com/openshift/release/issues/7679

/assign @petr-muller


petr-muller:
Thanks!

PR #16847, opened 2020-03-19
Revert "Enable NodeKiller on ClusterLoader presubmits"

mm4tt:
This reverts commit 53f19ac1b60b0a056afd47dae713aa49c6c31160.

Here is an example of PR when presubmit failed twice due to NK - https://github.com/kubernetes/perf-tests/pull/1126
An example presubmit run https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/perf-tests/1126/pull-perf-tests-clusterloader2/1240479871608557568/
In the above run, NK killed 6 nodes during density which in the end resulted in test timing out after 1h40min.

We should change the logic to not kill the nodes so many times.


oxddr:
/hold 
Could comment out those lines and add a comment with a todo? This way we won't forget to turn it on, once it's fixed.

mm4tt:
Done, ptal
/hold cancel

oxddr:
> Done, ptal
Thanks!

PR #15437, opened 2019-11-27
Deck: configurable rerun

clarketm:
fix #15415 (proposal)
motive: https://github.com/istio/test-infra/pull/2122

**Usage:**

There is no enigmatic inheritance. The `rerun_auth_config` key with the **highest** specificity (`job` -> `org/repo` -> `org` -> `wildcard`) is honored.

```yaml
rerun_auth_configs:

  # wildcard
  '*':
    allow_anyone: true

  # org 1
  istio:
    github_team_slugs:
    - org: istio
      slug: maintainers

  # org 2
  helm:
    github_team_slugs:
    - org: helm
      slug: developers

  # org 3
  kubernetes:
    github_orgs:
    - kubernetes
    - kubernetes-sigs

  # org/repo
  clarketm/tools:
    github_users:
      - clarketm

```

> **Note:** this is_backwards compatible with **all** *existing* Deck deployments and rerun configurations. Support for the deprecated `rerun_auth_config` will be removed in **May 2020**.

PR #16848, opened 2020-03-19
Deck configurable rerun: Part 2

clarketm:
On job rerun, look up the `RerunAuthConfig` from the `JobConfig` for the job, then patch the new ProwJob with the _latest_ configuration.

Part 2 for: https://github.com/kubernetes/test-infra/pull/15437

### DIFF
--- a/prow/pjutil/pjutil.go
+++ b/prow/pjutil/pjutil.go
@@ -239,6 +239,44 @@ func GetLatestProwJobs(pjs []prowapi.ProwJob, jobType prowapi.ProwJobType) map[s
 	return latestJobs
 }
 
+// ClonePJ clones a ProwJob using the definition from the JobConfig.
+func ClonePJ(pj *prowapi.ProwJob, cfg config.Getter) prowapi.ProwJob {
+	var repos []string
+
+	if pj.Spec.Refs != nil {
+		repos = []string{pj.Spec.Refs.Org + "/" + pj.Spec.Refs.Repo}
+	}
+
+	// Attempt to lookup ProwJob from JobConfig by type, repo and name.
+	switch pj.Spec.Type {
+	case prowapi.PresubmitJob:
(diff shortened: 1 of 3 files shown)


clarketm:
@alvaroaleman 

Before I proceed with refining the _approach_, can you please confirm that this effectively aligns with your proposal from Part 1: https://github.com/kubernetes/test-infra/pull/15437#discussion_r352287153.

/cc @alvaroaleman

prow/pjutil/pjutil.go
```diff
@@ -239,6 +239,44 @@ func GetLatestProwJobs(pjs []prowapi.ProwJob, jobType prowapi.ProwJobType) map[s
...
+
+	if pj.Spec.Refs != nil {
+		repos = []string{pj.Spec.Refs.Org + "/" + pj.Spec.Refs.Repo}
+	}
+
+	// Attempt to lookup ProwJob from JobConfig by type, repo and name.
+	switch pj.Spec.Type {
+	case prowapi.PresubmitJob:
+		for _, pre := range cfg().AllStaticPresubmits(repos) {
+			if pre.Name == pj.Spec.Job {
```
alvaroaleman: This also needs to take the branch into consideration (i.E. `pj.Spec.Refs.BaseRef`) because we unfortunately allow two jobs with the same name as long as they do not run on the same branch (Which we can not easily change, because it might cause a lot of breakages).

(And while you are at it, could you add to the godoc of `Refs.BaseRef` that its just a plain branch name? I look everytime at live prowjobs to find out if its  `refs/$branch` or just `$branch`)

clarketm: Sure thing, I will additionally include a branchname match and update the godoc.

clarketm: Oh, and add tests too 😉 .

PR #16849, opened 2020-03-19
Disable PVs for kubemark jobs

maniSbindra:
Currently PVs do not work on kube mark clusters due to reasons mentioned in the issue https://github.com/kubernetes/perf-tests/issues/803 .

This change disables PVs for kubemark jobs. Please see conversation around this in the PR https://github.com/kubernetes/perf-tests/pull/1125#discussion_r393675913.


mm4tt:
/ok-to-test

FTR, this change is no-op as currently CL2_ENABLE_PVS defaults to false. We're making this change because we want to change the default to true

mm4tt:
Thanks, @maniSbindra !

/lgtm
/approve

PR #16860, opened 2020-03-19
Fix broken CI job - ci-kubernetes-e2e-gci-gce-ipvs

dims:
The CI job fails as the 3 worker nodes fail to come up. A quick look at
the kubelet logs show:
```
F0318 21:59:03.419573    3167 server.go:274] failed to run Kubelet: docker API version is older than 1.26.0
```
So essentially the image has a stale runtime.

We need to resurrect the broken CI job. the closest working CI job to
this one is "ci-containerd-e2e-ubuntu-gce". With the updates the
configure.sh from the `containerd/cri` updates the runtime before
installing kubernetes.

### DIFF
--- a/config/jobs/kubernetes/sig-network/sig-network-misc.yaml
+++ b/config/jobs/kubernetes/sig-network/sig-network-misc.yaml
@@ -294,17 +294,20 @@ periodics:
   labels:
     preset-service-account: "true"
     preset-k8s-ssh: "true"
+    preset-e2e-containerd: "true"
+    preset-e2e-containerd-image-load: "true"
   spec:
     containers:
     - args:
+      - --repo=github.com/containerd/cri=master
       - --timeout=170
-      - --bare
       - --scenario=kubernetes_e2e
       - --
       - --check-leaked-resources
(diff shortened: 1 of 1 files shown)


dims:
@MrHohn @bowei this will help with testing the changes in https://github.com/kubernetes/kubernetes/pull/89116

aojea:
@dims seems that's the only job that fails in that file and I was checking executions of other jobs in that file and they install docker 19.03.1

I couldn't look much into this but a quick diff makes me think that these lines are the one causing this behavior

https://github.com/kubernetes/test-infra/blob/c05cefcf009ea833552835cc400a8d4cabf558f4/config/jobs/kubernetes/sig-network/sig-network-misc.yaml#L311-L312

Shouldn't be better to use the same configuration that the others jobs in that file for consistency?
Like this one per example 
https://github.com/kubernetes/test-infra/blob/c05cefcf009ea833552835cc400a8d4cabf558f4/config/jobs/kubernetes/sig-network/sig-network-misc.yaml#L320-L341

dims:
@aojea ok let us go with that. if that does not work we can come back to this.

PR #16862, opened 2020-03-19
Pass commit sha to metrics-server GCB

serathius:
/cc @Katharine


Katharine:
This LGTM but I’d like @cjwagner to weigh in on whether this is the correct variable for the SHA in a postsubmit.

cjwagner:
> This LGTM but I’d like @cjwagner to weigh in on whether this is the correct variable for the SHA in a postsubmit.

That is a valid variable for presubmits, but not for postsubmits. The available envvars for each job type are described here: https://github.com/kubernetes/test-infra/blob/master/prow/jobs.md#job-environment-variables

serathius:
Changed to `PULL_BASE_SHA` which should be available in post-submits. PTAL

PR #16863, opened 2020-03-19
Make Plank use a caching PJ client.

cjwagner:
We're seeing some issues with the API server taking too long to list PJs this is a mitigation for the most critical components.
/assign @chases2 @Katharine @stevekuznetsov 

I couldn't figure out what the default value of the timeout is from the docs and gave up trying to find it in the code.

### DIFF
--- a/prow/plank/controller.go
+++ b/prow/plank/controller.go
@@ -195,6 +195,23 @@ func (c *Controller) setPreviousReportState(pj prowapi.ProwJob) error {
 	return err
 }
 
+func (c *Controller) Start(stopChan <-chan struct{}) error {
+	ticker := time.NewTicker(30 * time.Second)
+	for {
+		select {
+		case <-stopChan:
+			c.log.Info("Stop signal received, quitting.")
+			return nil
+		case <-ticker.C:
+			start := time.Now()
+			if err := c.Sync(); err != nil {
+				logrus.WithError(err).Error("Error syncing.")
+			}
+			logrus.WithField("duration", fmt.Sprintf("%v", time.Since(start))).Info("Synced")
+		}
+	}
+}
+
 // Sync does one sync iteration.
 func (c *Controller) Sync() error {
 	var syncErrs []error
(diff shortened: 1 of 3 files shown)


prow/cmd/sinker/main.go
```diff
@@ -297,7 +297,18 @@ func (c *controller) clean() {
 
 	// Clean up old prow jobs first.
 	prowJobs := &prowapi.ProwJobList{}
-	if err := c.prowJobClient.List(c.ctx, prowJobs, ctrlruntimeclient.InNamespace(c.config().ProwJobNamespace)); err != nil {
+	timeoutSecs := (*int64)(nil)
+	if dur := c.config().Sinker.ProwJobListTimeout; dur != nil {
+		val := int64(dur.Duration.Seconds())
+		timeoutSecs = &val
+	}
+	listOpts := &ctrlruntimeclient.ListOptions{
```
alvaroaleman: We use a cache-backed client for sinker, this is effectively a timeout for copying the objects in memory

prow/plank/controller.go
```diff
@@ -200,9 +200,17 @@ func (c *Controller) Sync() error {
 	var syncErrs []error
 
 	pjs := &prowapi.ProwJobList{}
+	timeoutSecs := (*int64)(nil)
+	if dur := c.config().Plank.ProwJobListTimeout; dur != nil {
```
alvaroaleman: Maybe just change to also use a cache-backed client so we reduce the total  number of lists calls needed?

cjwagner:
> > We're seeing some issues with the API server taking too long to list PJs this is a mitigation for the most critical components.
> 
> Can you elaborate? Where do you see that, what impact does it have?

We've been seeing errors like this: `error listing prow jobs: the server was unable to return a response in the time allotted, but may still be processing the request (get prowjobs.prow.k8s.io)` on an internal Prow cluster with fewer ProwJob CRDs than k8s-prow. This prevents plank from syncing any ProwJobs. We're looking into the API server slowness as well, but would like to mitigate in the meantime.

prow/cmd/sinker/main.go
cjwagner: ACK, updated to only change plank.

prow/plank/controller.go
cjwagner: Done.

stevekuznetsov:
/hold cancel

Would love to see APIServer request latency and size metrics for these calls.

PR #12565, opened 2019-05-09
[WIP] Add Kubespray yamllint CI job

Miouge1:
Adding a simple CI job for Kubespray on Prow.

PR #16869, opened 2020-03-20
Add Kubespray yamllint CI job

Miouge1:
Adding a simple CI job for Kubespray on Prow.

Update OWNERS based on latest https://github.com/kubernetes-sigs/kubespray/blob/master/OWNERS_ALIASES in kubespray, minus myself and @lucckysb whoa re not members of the kubernetes org (see https://github.com/kubernetes/test-infra/pull/16869#issuecomment-601672819)

After I got stuck with #12565 I'm taking another shot at this, following the documentation in [config/jobs/README.md](https://github.com/kubernetes/test-infra/blob/master/config/jobs/README.md#adding-or-updating-jobs)

### DIFF
--- a/config/jobs/kubernetes-sigs/kubespray/kubespray-presubmits.yaml
+++ b/config/jobs/kubernetes-sigs/kubespray/kubespray-presubmits.yaml
@@ -1,0 +1,15 @@
+presubmits:
+  kubernetes-sigs/kubespray:
+  - name: pull-kubespray-yamllint
+    annotations:
+      testgrid-dashboards: sig-cluster-lifecycle-kubespray
+      testgrid-tab-name: yamllint
+    always_run: false
+    skip_report: true
+    spec:
+      containers:
+      - image: quay.io/kubermatic/yamllint:0.1
+        args:
+        - yamllint
+        - "--strict"
+        - .
--- a/config/jobs/kubernetes-sigs/kubespray/OWNERS
+++ b/config/jobs/kubernetes-sigs/kubespray/OWNERS
@@ -5,11 +5,7 @@ approvers:
 - mattymo
 - atoms
 - chadswen
-- rsmitty
-- bogdando
-- bradbeam
-- woopstar
+- mirwan
 - riverzhang
-- holser
-- smana
 - verwilst
+- woopstar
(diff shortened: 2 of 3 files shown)


config/jobs/kubernetes-sigs/kubespray/kubespray-presubmits.yaml
```diff
@@ -0,0 +1,20 @@
+presubmits:
+  kubernetes-sigs/kubespray:
+  - name: pull-kubespray-yamllint
+    annotations:
+      testgrid-dashboards: sig-cluster-lifecycle-kubespray
+      testgrid-tab-name: yamllint
+    branches:
+    - master
+    always_run: false
+    skip_report: true
```
neolit123: just to double check, do you want to only run this job on demand?
`always_run: false` means you have to trigger the job manually with `/test jobname`
`skip_report: true` means that you can only see failures in test-grid.

https://github.com/kubernetes/test-infra/blob/master/prow/jobs.md#how-to-configure-new-jobs

config/jobs/kubernetes-sigs/kubespray/kubespray-presubmits.yaml
```diff
@@ -0,0 +1,20 @@
...
+    labels:
+      preset-service-account: "true"
+      preset-dind-enabled: "true"
```
neolit123: the job does not appear to need these presets (docker-in-docker and google service accounts), so the labels section can be removed.

config/jobs/kubernetes-sigs/kubespray/kubespray-presubmits.yaml
Miouge1: We want to try it out with a manual trigger first. Inspired by the pattern described in https://github.com/kubernetes/test-infra/blob/ff6c355751962efd5d6b44f0f07567a9be1ea4f9/prow/jobs.md

config/jobs/kubernetes-sigs/kubespray/kubespray-presubmits.yaml
Miouge1: Good catch, probably comes from a copy paste. I removed that in the last commit.

config/jobs/kubernetes-sigs/kubespray/kubespray-presubmits.yaml
neolit123: ok, sounds good.

Miouge1:
@neolit123 sorry I updated the PR to remove the `branches: [master]` since I think that yamllint should be applicable to any branch.

neolit123:
/lgtm

you must find someone from this file to approve too:
https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes-sigs/kubespray/OWNERS
not sure who is active.

PR #16870, opened 2020-03-20
Change merge-method of cluster-api-provider-openstack to default (=me…

sbueringer:
…rge)

Changed to the default which is merge so we can you the CAPI release-notes tool which 
depends on merge commits


sbueringer:
/cc @jichenjc @hidekazuna @prankul88 
fyi. This leads to merge instead of squash when a PR is "merged". After this change we can 
use the ClusterAPI release notes tool to generate release notes (because it only greps for merge commits)

cblecker:
/approve

LGTM to @kubernetes-sigs/cluster-api-provider-openstack-maintainers

PR #16871, opened 2020-03-20
Fix for ci-kubernetes-e2e-gci-gce-ipvs

dims:
the ipvs job should use the same images as the other jobs. Currently
this job fails because the runtime version required by kubelet is very
old.


andrewsykim:
thanks @dims 

/lgtm

PR #16875, opened 2020-03-20
github: log a stacktrace when making a call to /repositories

stevekuznetsov:

/assign @alvaroaleman


stevekuznetsov:
Thirty minutes later, Bazel is still doing .... something ...

PR #16883, opened 2020-03-20
ci-k8sio-cip: do not use secret

listx:
Ever since https://github.com/kubernetes/k8s.io/pull/655, this job
should already run with an authenticated GCP SA (before the job's code
even runs). This means there is no need to use the keyfile to
authenticate (again), as it is redundant.

We still keep the `-use-service-account` flag for now because the
promoter currently still tries to extract OAUTH tokens as the GCP
account at the very beginning of its run. It's a nice sanity check.

Once we confirm that the OAUTH tokens do still get populated in the
business logic of the job, we don't need to keep this flag any more
because the GCRs that the promoter deals with today are all public (the
promoter was designed to deal with private repos), and so the very
first "read GCR" API calls do not need the OAUTH tokens for just reading
the repos. That takes care of the GCR read API calls.

As for the GCR write calls, the gcrane.Copy() method that perform the
GCR writes do their own authentication dance, but that is separate and
automatic as long as the GCP account is authenticated before that
function call.

This is why the `-use-service-account` flag is marked for future
deprecation in this commit.

### DIFF
--- a/config/jobs/kubernetes/test-infra/test-infra-trusted.yaml
+++ b/config/jobs/kubernetes/test-infra/test-infra-trusted.yaml
@@ -1143,17 +1143,8 @@ periodics:
       - cip
       args:
       - -thin-manifest-dir=/home/prow/go/src/github.com/kubernetes/k8s.io/k8s.gcr.io
-      - -key-files=/etc/k8s-artifacts-prod-service-account/service-account.json
       - -dry-run=false
-      - -use-service-account
-      volumeMounts:
-      - name: k8s-artifacts-prod-service-account-creds
-        mountPath: /etc/k8s-artifacts-prod-service-account
-        readOnly: true
-    volumes:
-    - name: k8s-artifacts-prod-service-account-creds
-      secret:
-        secretName: k8s-artifacts-prod-service-account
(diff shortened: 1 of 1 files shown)


listx:
Thanks!

listx:
The job is now failing with:

```
*looking at "/home/prow/go/src/github.com/kubernetes/k8s.io/k8s.gcr.io"
E0323 23:16:47.862361      12 token.go:57] could not execute cmd /usr/bin/gcloud auth --account=k8s-infra-gcr-promoter@k8s-artifacts-prod.iam.gserviceaccount.com print-access-token
E0323 23:16:47.862477      12 inventory.go:1141] could not get service account token for k8s-infra-gcr-promoter@k8s-artifacts-prod.iam.gserviceaccount.com
F0323 23:16:47.862495      12 cip.go:238] exit status 1
```

I've checked the k8s-artifacts-prod side of IAM configs and it appears that it was correctly actuated. So the settings look fine.

I thought about this a liitle more and I think this error is happnening because I forgot to define a `serviceAccountName` here for this job. Oops!

In my case it should be `k8s-artifacts-prod` because this job is already in the `test-pods` K8s namespace, as defined in https://github.com/kubernetes/k8s.io/pull/655.

PR incoming...!

PR #16889, opened 2020-03-21
Merge PR with base branch before checking OWNERS

### LINKED ISSUE #16330: verify-owners: checks out PR branch, instead of merging PR branch with base branch
**What happened**:
verify-owners checks out the PR branch when verifying the owners files and tests that, rather than merging the PR branch with the base branch, and then testing.

verify-owners would merge the PR with the base branch, and check on the result. This is prevalent in the case that OWNERS_ALIASES are being used, and are modified between when the PR branch is created, but before the PR is merged.

- Create a PR referencing an alias that doesn't exist
- Create the matching entry in OWNERS_ALIASES, and have that PR merge
- Attempt to /verify-owners on the first PR. It will not succeed, even though the alias exists in master

https://github.com/kubernetes/perf-tests/pull/1030

Section of logic: 
https://github.com/kubernetes/test-infra/blob/92e4176fa4877ba1823eabfc576fe1665f8f3762/prow/plugins/verify-owners/verify-owners.go#L223-L241

Instead of using `CheckoutPullRequest` (in `prow/git/git.go`) we should be able to use `MergeAndCheckout` with the base/head of the PR to merge it and then test against that.

@cjwagner @stevekuznetsov @fejta: Is it worth possibly a new helper function like `CheckoutMergedPullRequest` that has just the PR input like `CheckoutPullRequest`, but uses `MergeAndCheckout` to merge the PR before returning the checked out branch?

adshmh:
The verify-owners plugin now merges a PR with its base branch before
validating the OWNERS file(s). This is to cover the scenario in which the
error (e.g. a missing owners alias) has been fixed through a separate PR,
but the plugin would still report an invalid OWNERS file, because it only
performs a checkout on the PR rather than merging it with the base branch.

Closes #16330

### DIFF
--- a/prow/plugins/verify-owners/verify-owners.go
+++ b/prow/plugins/verify-owners/verify-owners.go
@@ -231,16 +231,18 @@ func handle(ghc githubClient, gc git.ClientFactory, roc repoownersClient, log *l
 			log.WithError(err).Error("Error cleaning up repo.")
 		}
 	}()
-	if err := r.CheckoutPullRequest(number); err != nil {
+	if err := r.Config("user.name", "prow"); err != nil {
 		return err
 	}
-	// If we have a specific SHA, use it.
-	if pr.Head.SHA != "" {
-		if err := r.Checkout(pr.Head.SHA); err != nil {
-			return err
-		}
+	if err := r.Config("user.email", "prow@localhost"); err != nil {
+		return err
 	}
-
+	if err := r.Config("commit.gpgsign", "false"); err != nil {
+		log.WithError(err).Errorf("Cannot set gpgsign=false in gitconfig: %v", err)
+	}
+	if err := r.MergeAndCheckout(pr.Base.Ref, string(github.MergeMerge), pr.Head.SHA); err != nil {
+		return err
+	}
 	// If OWNERS_ALIASES file exists, get all aliases.
 	// If the file was modified, check for non trusted users in the newly added owners.
 	nonTrustedUsers, repoAliases, err := nonTrustedUsersInOwnersAliases(ghc, log, triggerConfig, org, repo, r.Directory(), modifiedOwnerAliasesFile.Patch, ownerAliasesModified, skipTrustedUserCheck)
--- a/prow/plugins/verify-owners/verify-owners_test.go
+++ b/prow/plugins/verify-owners/verify-owners_test.go
@@ -123,6 +123,22 @@ labels:
     labels:
     - label1
 `),
+	"referencesToBeAddedAlias": []byte(`approvers:
+- not-yet-existing-alias
+`),
+}
+
+var patches = map[string]string{
+	"referencesToBeAddedAlias": `
++ - not-yet-existing-alias
+`,
+}
+
+var ownerAliasesFiles = map[string][]byte{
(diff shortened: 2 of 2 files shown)


adshmh:
side note: while working on the unit tests, I changed the test function to use `Run` for better consistency with the rest of test functions in the package.

prow/plugins/verify-owners/verify-owners.go
```diff
@@ -231,16 +231,9 @@ func handle(ghc githubClient, gc git.ClientFactory, roc repoownersClient, log *l
 			log.WithError(err).Error("Error cleaning up repo.")
 		}
 	}()
-	if err := r.CheckoutPullRequest(number); err != nil {
+	if err := r.MergeAndCheckout(pr.Base.Ref, string(github.MergeMerge), pr.Head.SHA); err != nil {
```
alvaroaleman: I think the first argument needs to be `pr.Base.SHA` instead of `pr.Base.Ref`

prow/plugins/verify-owners/verify-owners_test.go
```diff
@@ -136,10 +152,31 @@ func IssueLabelsAddedContain(arr []string, str string) bool {
 	return false
 }
 
-func newFakeGitHubClient(changed []string, removed []string, pr int) *fakegithub.FakeClient {
+func ownersFilePatch(files []string, ownersFile string, nonEmptyPatch bool) map[string]string {
+	changes := emptyPatch(files)
+	if !nonEmptyPatch {
```
alvaroaleman: Nit: Not inversing too much makes reading the code easier, so here instead of `!nonEmptyPatch` just `emptyPatch`

adshmh: Thank you for the review. Fixed.

prow/plugins/verify-owners/verify-owners.go
adshmh: Thank you for the review. I have used `pr.Base.Ref` because `pr.Base.SHA` points to the version of the base branch off which the PR was originally created, thus missing updates to the base branch (e.g. addition of the missing alias through a separate PR). I tested with the specified scenario using `pr.Base.SHA` and it missed the update to base branch.

prow/plugins/verify-owners/verify-owners.go
```diff
@@ -231,16 +231,18 @@ func handle(ghc githubClient, gc git.ClientFactory, roc repoownersClient, log *l
...
-	// If we have a specific SHA, use it.
-	if pr.Head.SHA != "" {
-		if err := r.Checkout(pr.Head.SHA); err != nil {
-			return err
-		}
+	if err := r.Config("user.email", "prow@localhost"); err != nil {
+		return err
+	}
+	if err := r.Config("commit.gpgsign", "false"); err != nil {
+		log.WithError(err).Errorf("Cannot set gpgsign=false in gitconfig: %v", err)
```
alvaroaleman: Nit: The error in the `Errorf` statement is redundant, it is already logged in the `WithError` statement

alvaroaleman:
Merging before this starts rotting.
/hold cancel

PR #16892, opened 2020-03-22
Switch SIG Docs merge method from squash to merge (default)

jimangel:
By removing the custom prow config to squash commits, SIG Docs will begin using the default prow merge method ([merge](https://github.com/kubernetes/test-infra/blob/master/prow/cmd/tide/config.md#general-configuration)).

Early on, `squash` was used for SIG Docs as it was easier to revert / fix commits. However, today we find it to be problematic in the release cycle as the release lead needs to change the label to `merge` to resolve conflicts.

As a result, @kubernetes/sig-docs-en-reviews and @kubernetes/sig-docs-en-owners will need to be more strict around squashing commits before merging.

This is the first step of a larger effort to simplify the docs release process.

/cc @kubernetes/sig-docs-zh-reviews @kubernetes/sig-docs-ko-reviews @kubernetes/sig-docs-de-reviews @kubernetes/sig-docs-es-reviews @kubernetes/sig-docs-pt-reviews @kubernetes/sig-docs-id-reviews @kubernetes/sig-docs-ru-reviews @kubernetes/sig-docs-pl-reviews @kubernetes/sig-docs-vi-reviews
/cc @zacharysarah @kbarnard10 @sftim @kbhawkey

/hold

(for discussion if needed)

### DIFF
--- a/config/prow/config.yaml
+++ b/config/prow/config.yaml
@@ -490,7 +490,6 @@ tide:
     kubernetes/cloud-provider-openstack: squash
     kubernetes/dashboard: squash
     kubernetes/kube-deploy: squash
-    kubernetes/website: squash
   pr_status_base_urls:
     '*': https://prow.k8s.io/pr
   blocker_label: tide/merge-blocker


zacharysarah:
💯 agreed. 

/lgtm

justaugustus:
I was going to suggest this next week as an immediate improvement you could make, but I see everything was already wonderfully written up in https://github.com/kubernetes/release/issues/1204.

Strong +1 on this, but I would suggest an explicit hold until 1.18.0 is released, just so we're not changing anyone's workflow underneath them during crunch time.
/hold
/lgtm

zacharysarah:
@justaugustus 👋 1.18.0 is out the door, so I think we can move forward with this.

@jimangel Are you ready for this to merge?

gochist:
I think `[scripts/diff_l10n_branches.py](https://github.com/kubernetes/website/blob/master/scripts/diff_l10n_branches.py)` will work well in this scenario as well.

/lgtm

PR #16073, opened 2020-01-30
Add a pull-kubernetes-e2e-gce-canary job

spiffxp:
Try running pull-kubernetes-e2e-gce with --build=quick instead of
--build=bazel

ref: https://github.com/kubernetes/kubernetes/issues/87441

PR #16895, opened 2020-03-23
Rename 'pull-kubernetes-e2e-gce-canary' and enable as non-blocking

justaugustus:
- s/`pull-kubernetes-e2e-gce-canary`/`pull-kubernetes-e2e-gce-canary-no-bazel`

  So we actually know in the job name that it's running e2e-gce without
  bazel, specifically with the '--build=quick' flag.

- Set to always run
- Make non-blocking
- Add to `sig-release-master-informing` and `sig-testing-canaries`
- Add alert emails

---

Follow-up to @BenTheElder's review [here](https://github.com/kubernetes/test-infra/pull/16073#discussion_r374298253):

>  always_run: false + skip_report will mean when you /test pull-kubernetes-e2e-gce-canary you need to go track down the results yourself /shrug

ref: https://github.com/kubernetes/kubernetes/issues/88553, https://github.com/kubernetes/kubernetes/pull/89275, https://github.com/kubernetes/test-infra/pull/16073, https://github.com/kubernetes/kubernetes/issues/87441

### DIFF
--- a/config/jobs/kubernetes/sig-cloud-provider/gcp/gcp-gce.yaml
+++ b/config/jobs/kubernetes/sig-cloud-provider/gcp/gcp-gce.yaml
@@ -60,15 +60,17 @@ presubmits:
           requests:
             memory: "6Gi"
 
-  - name: pull-kubernetes-e2e-gce-canary
-    always_run: false
-    skip_report: true
+  - name: pull-kubernetes-e2e-gce-canary-no-bazel
+    always_run: true
+    optional: true
     skip_branches:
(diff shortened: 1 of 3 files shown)


spiffxp:
/hold
I would recommend closing this.  I added pull-kubernetes-e2e-gce-canary for this purpose via https://github.com/kubernetes/test-infra/pull/16073

justaugustus:
Aha! I was wondering what that job was for.
@spiffxp -- I've updated the PR to add a few things to the existing job. See updated description.

/assign @spiffxp @BenTheElder

PR #16896, opened 2020-03-23
Make pull-kubernetes-bazel-{build,test} jobs non-blocking

justaugustus:
ref: https://github.com/kubernetes/kubernetes/issues/88553, https://github.com/kubernetes/kubernetes/pull/89275


justaugustus:
@Katharine -- do we have non-bazel analogs for those jobs?

Katharine:
Not currently. Without significant work, the drop-in replacements would be several times slower than the bazel versions. This is also not tolerable in presubmit.

justaugustus:
Gotcha. Closing this.
/close

Issue #14810, opened 2019-10-16
Migrate away from deprecated `--build-cluster` flag and format.


clarketm:
**What should be cleaned up or changed**:

Migrate away from and deprecate `--build-cluster` flag and format.
- [x] Migrate K8s Prow deployments to use `--kubeconfig` (#16906)
- [x] Deprecation and removal of `mkbuild-cluster`
  - [x] Create an [announcement](https://github.com/kubernetes/test-infra/blob/master/prow/ANNOUNCEMENTS.md) and set removal date, #15285.
  - [x] Replace `mkbuild-cluster` references in [`getting_started_deploy.md`](https://github.com/kubernetes/test-infra/blob/master/prow/getting_started_deploy.md) with the equivalent `gencred` command(s), #15285.  
  - [x] Remove `mkbuild-cluster` source code ([ref](https://github.com/kubernetes/test-infra/blob/master/label_sync/labels.yaml)), **to be completed May 2020**, #15287.
  - [x] Remove `mkbuild-cluster` label ([ref](https://github.com/kubernetes/test-infra/blob/master/prow/cmd/mkbuild-cluster)), **to be completed May 2020**, #15287.

/assign

BenTheElder:
> Migrate K8s Prow deployments to use --kubeconfig

Doesn't seem like contributors can handle this? Also you've already assigned yourself...?

clarketm:
/kind oncall-hotlist

The only thing left is on-call **process**: create a `kubeconfig` file, save it in a secret, and define it (using `--kubeconfig`) on the k8s resources that require it (e.g. `plank`, `sinker`, `deck`).

fejta:
Would be great to make some progress on here, happy to help out during my oncall shift.

fejta:
/assign @fejta @clarketm 

Do we have a tool to create the kubeconfig file? (convert the existing one?)

clarketm:
> Do we have a tool to create the kubeconfig file? (convert the existing one?)

Yes, that is [`gencred`](https://github.com/kubernetes/test-infra/tree/master/gencred). Its usage has been added to the Prow getting started guide https://github.com/kubernetes/test-infra/pull/15285.

fejta:
/unassign @clarketm 
I'll give it a go then, thanks!

fejta:
Still pending

clarketm:
>  Migrate K8s Prow deployments to use `--kubeconfig`

[...]

PR #16906, opened 2020-03-23
Migrate k8s Prow to use --kubeconfig

clarketm:
Last item for: https://github.com/kubernetes/test-infra/issues/14810


alvaroaleman:
/uncc
Change in itself looks correct, but I can neither deploy nor verify it.

PR #16907, opened 2020-03-23
allow squash merging for new sig-service-catalog repos

jberkhahn:
We got these new repos a while back but never added them here. We're doing work on them now, so it would be nice to be able to merge stuff. Hopefully I did this right?

Thanks!

### DIFF
--- a/config/prow/config.yaml
+++ b/config/prow/config.yaml
@@ -487,6 +487,8 @@ tide:
     kubernetes-sigs/krew: squash
     kubernetes-sigs/kubespray: squash
     kubernetes-sigs/service-catalog: squash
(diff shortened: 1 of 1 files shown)


config/prow/config.yaml
```diff
@@ -487,6 +487,8 @@ tide:
     kubernetes-sigs/krew: squash
     kubernetes-sigs/kubespray: squash
     kubernetes-sigs/service-catalog: squash
+    kubernetes-sigs/go-open-service-broker-client: squash
+    /kubernetes-sigs/minibroker: squash
```
cblecker:
```suggestion
    kubernetes-sigs/minibroker: squash
```

jberkhahn:
yes, that's fine

PR #16908, opened 2020-03-23
Create launcher.gcr.io/google/bazel:2.2.0-from-0.25.2

fejta:
/assign @mikedanese @Katharine 

This will create a `gcr.io/k8s-testimages/launcher.gcr.io/google/bazel:2.2.0-from-0.25.2` that has both `2.2.0` and `0.25.2` installed, selectable in a PR via the `.bazelversion` file.


mikedanese:

PR #16910, opened 2020-03-23
Add support for a second bazel version in kubekins-e2e

fejta:
/assign @mikedanese @Katharine


images/kubekins-e2e/Dockerfile
```diff
@@ -74,6 +76,9 @@ ENV BAZEL_VERSION=${BAZEL_VERSION_ARG}
 COPY ./install-bazel.sh /
 RUN bash /install-bazel.sh
 
+COPY --from=old \
+    /usr/local/lib/bazel/bin/bazel-real /usr/local/lib/bazel/bin/bazel-${OLD_VERSION}
```
mikedanese:
```suggestion
    /usr/local/lib/bazel/bin/bazel-real /usr/local/lib/bazel/bin/bazel-${OLD_BAZEL_VERSION?}
```

fejta: Done

images/kubekins-e2e/Dockerfile
```diff
@@ -74,6 +76,9 @@ ENV BAZEL_VERSION=${BAZEL_VERSION_ARG}
 COPY ./install-bazel.sh /
 RUN bash /install-bazel.sh
 
+COPY --from=old \
```
mikedanese: Add
```
ARG OLD_BAZEL_VERSION
```
Again before this line. Previous declaration was not in this stage.

fejta: Done

justaugustus:
@fejta @mikedanese -- so... how do we use it?

fejta:
$REPOROOT/.bazelversion determines the actual version of bazel used.

Set it to `0.23.2` and it will run with `0.23.2`
Set it to `2.2.0` and it will run with `2.2.0`
Set it to `2.1.0` and it will error with a message of how to install that version (I only setup the image to come with two versions preinstalled).

This is the default behavior of bazel 1.0+ (and the default behavior of [bazelisk](https://github.com/bazelbuild/bazelisk))

BenTheElder:
trying to figure out which change caused breakage, not long after this the 1.15 bazel e2e build started failing, one of the only changes is the kubekins image bump https://testgrid.k8s.io/conformance-all#kind,%20v1.15%20%28dev%29

BenTheElder:
we don't have a .bazelversion on branches prior to 1.17

fejta:
Nice find

BenTheElder:
https://github.com/kubernetes/kubernetes/pull/89446 & https://github.com/kubernetes/kubernetes/pull/89448

justaugustus:
Both CPs have been approved. Thanks y'all!!

mikedanese:

PR #16912, opened 2020-03-23
Create a 2.2.0-from-0.23.2

fejta:
/assign @mikedanese

PR #16913, opened 2020-03-23
sync canary bazel jobs with HEAD

mikedanese:
And use 2.2.0-from-0.23.2 image created in https://github.com/kubernetes/test-infra/pull/16912

Jobs here:

https://github.com/kubernetes/test-infra/blob/022b1408435254aea1523c0921a94783de56c0ec/config/jobs/kubernetes/sig-testing/bazel-build-test.yaml#L22

/assign @fejta


mikedanese:

PR #16916, opened 2020-03-24
fix canary images

mikedanese:
I can't use git. This was in my stage but not commited.

/assign @fejta


mikedanese:

PR #16917, opened 2020-03-24
ci-k8sio-cip: use k8s service account

listx:
This is a follow-up to https://github.com/kubernetes/k8s.io/pull/655 and
https://github.com/kubernetes/test-infra/pull/16883.

/cc @fejta @thockin 

Aside: I think the KSA name "k8s-artifacts-prod" is bad (too generic) and should be changed in the future, but let's do that later when we need to disambiguate.

### DIFF
--- a/prow/cluster/trusted_serviceaccounts.yaml
+++ b/prow/cluster/trusted_serviceaccounts.yaml
@@ -31,12 +31,18 @@ metadata:
   name: deployer
   namespace: test-pods
 ---
+kind: ServiceAccount
+apiVersion: v1
+metadata:
+  annotations:
+    iam.gke.io/gcp-service-account: k8s-infra-gcr-promoter@k8s-artifacts-prod.iam.gserviceaccount.com
+  name: k8s-artifacts-prod
+  namespace: test-pods
 # TODO(fejta): https://github.com/kubernetes/test-infra/issues/15806
 # * Run experiment/workload-identity/bind-service-accounts.sh on the above
 # * Config service account on job
 # Do the same for the following:
 # k8s-artifacts-graveyard-service-account
 # k8s-artifacts-prod-bak-service-account
-# k8s-artifacts-prod-service-account
 # k8s-gcr-prod-service-account
 # service-account
--- a/config/jobs/kubernetes/test-infra/test-infra-trusted.yaml
+++ b/config/jobs/kubernetes/test-infra/test-infra-trusted.yaml
@@ -1152,6 +1152,9 @@ periodics:
     repo: k8s.io
     base_ref: master
   spec:
+    # The k8s-artifacts-prod name was chosen in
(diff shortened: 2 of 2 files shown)


listx:
@fejta: @spiffxp added a hold for now because it's not clear if the `k8s-artifacts-prod` KSA exists yet in the `test-pods` namespace, and he does not have access to create one.

Can you remove hold if you can confirm that one has been created (or is this step not necessary)? Thanks.

fejta:
These are self service, we apply what is declared on merge:

https://github.com/kubernetes/test-infra/blob/0132e79daa5376236ba23c099bb3331c6776d04a/prow/cluster/BUILD.bazel#L63-L73

As Aaron says, if you want a cluster with that name to exist, you need to declare it (in the correct cluster where the job runs):

https://github.com/kubernetes/test-infra/blob/master/prow/cluster/trusted_serviceaccounts.yaml
https://github.com/kubernetes/test-infra/blob/master/prow/cluster/build_serviceaccounts.yaml

prow/cluster/trusted_serviceaccounts.yaml
```diff
@@ -31,12 +31,18 @@ metadata:
   name: deployer
   namespace: test-pods
 ---
+kind: ServiceAccount
+apiVersion: v1
+metadata:
+  annotations:
+    iam.gke.io/gcp-service-account: k8s-infra-gcr-promoter@k8s-artifacts-prod.iam.gserviceaccount.com
```
fejta: Note that you'll still need to run the gcloud comment to authorize `k8s-artifacts-prod`'s ability to authenticate as `k8s-infra-gcr-promoter@k8s-artifacts-prod.iam.gserviceaccount.com` to GCP.

listx: This has already been done by @thockin; I've indepently verified that there is a "Workload Identity" role added to the `k8s-infra-gcr-promoter@k8s-artifacts-prod.iam.gserviceaccount.com` account from the IAM web ui for the `k8s-artifacts-prod` project which owns that GCP SA, so I think we're ready for merge.

listx:
> These are self service, we apply what is declared on merge:
> 
> https://github.com/kubernetes/test-infra/blob/0132e79daa5376236ba23c099bb3331c6776d04a/prow/cluster/BUILD.bazel#L63-L73
> 
> As Aaron says, if you want a service account with that name to exist, you need to declare it (in the correct cluster where the job runs):
> 
> https://github.com/kubernetes/test-infra/blob/master/prow/cluster/trusted_serviceaccounts.yaml
> https://github.com/kubernetes/test-infra/blob/master/prow/cluster/build_serviceaccounts.yaml

ACK. ISTR these files before when you were first doing this stuff earlier this year Erick, thanks for the reminder.

Issue #13059, opened 2019-06-17
prow: hook: poor ux with invalid /test and /retest commands


stevekuznetsov:
As a developer interacting with Prow,
I want to get feedback when I've issued a command that could not be parsed
So that I can fix my command and get the service to do what I intended

Examples of comment commands that should generate an error message in a comment response:

 - `/test` with no target
 - `/retest` with trailing words
 - `/test` with a target that is not recognized

/area prow/hook
/good-first-issue
/help
/cc @spiffxp @cblecker @nikhita

cblecker:
Is this something the generic command framework would do? Is it worth waiting, or is this simple to do in trigger as it is?

stevekuznetsov:
I think it would be better implemented with per-plugin logic anyway, I want to see something like:

> @user: to request testing, provide a test name like `/test test-name`

Or 

> @user: no test `fake-test` was found for the `oops` branch on this repo, please select a test from the following options
> - `unit`
> - `integration`

etc. 

The framework would have to expose hooks for intelligent responses, so the amount of work taken should be similar in both cases. Changing `trigger` to do this today should not be too hard.

nikhita:
For anyone interested to take this up, the trigger plugin is here: https://github.com/kubernetes/test-infra/tree/master/prow/plugins/trigger

alvaroaleman:
Fixed in https://github.com/kubernetes/test-infra/pull/16926
/close

PR #16926, opened 2020-03-24
prow command to list possible tests on PR

### LINKED ISSUE #16765: Prow command to list all possible tests to run on a PR
Crossposting this issue from https://github.com/knative/test-infra/issues/1805

> It's not obvious what tests I can run against a PR. I tend to go to the prow config in this repo but that's not ideal.
> 
> I'd be nice if prow have a command that would list all the tests.
- @dprotaso 
cc @chizhg 

Sounds like a really useful feature to have. 💯

adshmh:
The command `/test ?` or `/test -ls` results in a response comment from the `trigger` plugin listing all available tests for a PR. Closes #16765

### DIFF
--- a/prow/plugins/trigger/generic-comment.go
+++ b/prow/plugins/trigger/generic-comment.go
@@ -18,6 +18,8 @@ package trigger
 
 import (
 	"fmt"
+	"regexp"
+	"strings"
 
 	"github.com/sirupsen/logrus"
 
@@ -28,6 +30,8 @@ import (
 	"k8s.io/test-infra/prow/pjutil"
 	"k8s.io/test-infra/prow/plugins"
 )
+
+var testHelpRe = regexp.MustCompile(`(?m)^/test\s*\?\s*$`)
 
 func handleGenericComment(c Client, trigger plugins.Trigger, gc github.GenericCommentEvent) error {
 	org := gc.Repo.Owner.Login
@@ -53,8 +57,12 @@ func handleGenericComment(c Client, trigger plugins.Trigger, gc github.GenericCo
 
 	refGetter := config.NewRefGetterForGitHubPullRequest(c.GitHubClient, org, repo, number)
 	presubmits := getPresubmits(c.Logger, c.GitClient, c.Config, org+"/"+repo, refGetter.BaseSHA, refGetter.HeadSHA)
+
 	// Skip comments not germane to this plugin
-	if !pjutil.RetestRe.MatchString(gc.Body) && !pjutil.OkToTestRe.MatchString(gc.Body) && !pjutil.TestAllRe.MatchString(gc.Body) {
+	if !pjutil.RetestRe.MatchString(gc.Body) &&
+		!pjutil.OkToTestRe.MatchString(gc.Body) &&
+		!pjutil.TestAllRe.MatchString(gc.Body) &&
+		!testHelpRe.MatchString(gc.Body) {
 		matched := false
 		for _, presubmit := range presubmits {
 			matched = matched || presubmit.TriggerMatches(gc.Body)
@@ -87,6 +95,31 @@ func handleGenericComment(c Client, trigger plugins.Trigger, gc github.GenericCo
 		}
 	}
 
+	pr, err := refGetter.PullRequest()
+	if err != nil {
+		return err
+	}
+
+	// Process "help" comments
+	if testHelpRe.MatchString(gc.Body) {
+		branch := pr.Base.Ref
+		available, err := availablePresubmits(c.GitHubClient, gc.Body, org, repo, branch, pr.Number, presubmits, c.Logger)
+		if err != nil {
+			return err
+		}
+		var resp string
+		if len(available) > 0 {
+			var listBuilder strings.Builder
+			for _, name := range available {
+				listBuilder.WriteString(fmt.Sprintf("\n* `%s`", name))
+			}
+			resp = fmt.Sprintf("The following commands are available to trigger jobs:%s\n\nUse `/test all` to run all jobs.", listBuilder.String())
+		} else {
+			resp = fmt.Sprintf("No presubmit jobs available for %s/%s@%s", org, repo, branch)
+		}
+		return c.GitHubClient.CreateComment(org, repo, number, plugins.FormatResponseRaw(gc.Body, gc.HTMLURL, gc.User.Login, resp))
+	}
+
 	// At this point we can trust the PR, so we eventually update labels.
 	// Ensure we have labels before test, because TrustedPullRequest() won't be called
 	// when commentAuthor is trusted.
@@ -108,10 +141,6 @@ func handleGenericComment(c Client, trigger plugins.Trigger, gc github.GenericCo
 		}
 	}
 
-	pr, err := refGetter.PullRequest()
-	if err != nil {
-		return err
-	}
 	baseSHA, err := refGetter.BaseSHA()
 	if err != nil {
 		return err
@@ -168,6 +197,22 @@ func FilterPresubmits(honorOkToTest bool, gitHubClient GitHubClient, body string
 	number, branch := pr.Number, pr.Base.Ref
 	changes := config.NewGitHubDeferredChangedFilesProvider(gitHubClient, org, repo, number)
 	return pjutil.FilterPresubmits(filter, changes, branch, presubmits, logger)
+}
+
+func availablePresubmits(githubClient GitHubClient, body, org, repo, branch string, number int, presubmits []config.Presubmit, logger *logrus.Entry) ([]string, error) {
+	changes := config.NewGitHubDeferredChangedFilesProvider(githubClient, org, repo, number)
+	all := func(p config.Presubmit) (bool, bool, bool) {
+		return true, true, true
+	}
+	toTest, _, err := pjutil.FilterPresubmits(all, changes, branch, presubmits, logger)
+	if err != nil {
+		return nil, err
+	}
+	var available []string
+	for _, pre := range toTest {
+		available = append(available, pre.RerunCommand)
+	}
+	return available, nil
 }
 
 func getContexts(combinedStatus *github.CombinedStatus) (sets.String, sets.String) {
--- a/prow/plugins/trigger/generic-comment.go
+++ b/prow/plugins/trigger/generic-comment.go
@@ -31,7 +31,23 @@ import (
 	"k8s.io/test-infra/prow/plugins"
 )
 
-var testHelpRe = regexp.MustCompile(`(?m)^/test\s*\?\s*$`)
+var (
+	testHelpRe          = regexp.MustCompile(`(?m)^/test\s*\?\s*$`)
+	emptyTestRe         = regexp.MustCompile(`(?m)^/test\s*$`)
+	retestWithTargetRe  = regexp.MustCompile(`(?m)^/retest\s+\S+$`)
+	testWithAnyTargetRe = regexp.MustCompile(`(?m)^/test\s+\S+$`)
+
+	testWithoutTargetNote = "The `/test` command needs one or more targets.\n"
+	retestWithTargetNote  = "The `/retest` command does not accept any targets.\n"
+	targetNotFoundNote    = "The specified target(s) for `/test` were not found.\n"
+)
+
+func mayNeedHelpComment(body string) bool {
+	return emptyTestRe.MatchString(body) ||
+		retestWithTargetRe.MatchString(body) ||
+		testWithAnyTargetRe.MatchString(body) ||
+		testHelpRe.MatchString(body)
+}
 
 func handleGenericComment(c Client, trigger plugins.Trigger, gc github.GenericCommentEvent) error {
 	org := gc.Repo.Owner.Login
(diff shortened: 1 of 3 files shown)


prow/plugins/trigger/generic-comment.go
```diff
@@ -29,6 +31,8 @@ import (
 	"k8s.io/test-infra/prow/plugins"
 )
 
+var testHelpRe = regexp.MustCompile(`(?m)^/test\s*(\?|-ls)$`)
```
petr-muller: I'm not sure `/test ?` or `/test -ls` is the best command for this. The first is fine-ish, the second is IMO unnecessarily technical and unix-y.  I'm thinking if something more human oriented, like `/list-tests` would be better?

prow/plugins/trigger/generic-comment.go
```diff
@@ -87,6 +95,24 @@ func handleGenericComment(c Client, trigger plugins.Trigger, gc github.GenericCo
...
+		return err
+	}
+
+	// Process "help" comments
+	if testHelpRe.MatchString(gc.Body) {
+		available, err := availablePresubmits(c.GitHubClient, gc.Body, org, repo, pr.Base.Ref, pr.Number, presubmits, c.Logger)
+		if err != nil {
+			return err
+		}
+		resp := fmt.Sprintf("No presubmit jobs available for %s/%s", org, repo)
```
petr-muller: Should this message include also the branch? We may get here even if jobs are actually available for org/repo, but not the branch for which the PR was submitted.

Also, I think this should be in an `else` rather than conditionally overwritten.

prow/plugins/trigger/generic-comment.go
```diff
@@ -87,6 +95,24 @@ func handleGenericComment(c Client, trigger plugins.Trigger, gc github.GenericCo
...
+		if len(available) > 0 {
+			resp = fmt.Sprintf("The following jobs can be triggered by using their name after a `/test` message: %s", strings.Join(markdownNames(available), ", "))
```
petr-muller: comma separated list can get messy, I think we should do a markdown list instead.

prow/plugins/trigger/generic-comment.go
```diff
@@ -87,6 +95,24 @@ func handleGenericComment(c Client, trigger plugins.Trigger, gc github.GenericCo
...
 	}
 
+	pr, err := refGetter.PullRequest()
+	if err != nil {
+		return err
+	}
+
+	// Process "help" comments
+	if testHelpRe.MatchString(gc.Body) {
+		available, err := availablePresubmits(c.GitHubClient, gc.Body, org, repo, pr.Base.Ref, pr.Number, presubmits, c.Logger)
```
alvaroaleman: I think we should also fix https://github.com/kubernetes/test-infra/issues/13059 in this PR by responding with the same comment if someone requested an invalid test

prow/plugins/trigger/generic-comment.go
adshmh: Thank you for the review. I have updated the PR to use `/test ?` and `/list-tests` as help commands.

prow/plugins/trigger/generic-comment.go
adshmh: Thank you for the review. Updated the message to include the branch, and fixed the condition.

prow/plugins/trigger/generic-comment.go
adshmh: Thank you for the review. Fixed.

prow/plugins/trigger/generic-comment.go
adshmh: Thank you for the review. Fixed. I included `/test` with unrecognised target, `/test` without any targets, and `/retest` with trailing words in the fix to address all cases listed on #13059

prow/plugins/trigger/generic-comment.go
```diff
@@ -121,6 +147,12 @@ func handleGenericComment(c Client, trigger plugins.Trigger, gc github.GenericCo
 	if err != nil {
 		return err
 	}
+
+	addedComment, err := addHelpComment(c.GitHubClient, len(toTest), len(toSkip), gc.Body, org, repo, branch, pr.Number, presubmits, gc.HTMLURL, commentAuthor, c.Logger)
+	if addedComment {
```
petr-muller: This does not feel right.

1. Do we ever want to continue to the `RunAndSkipJobs...` if `err != nil`?
2. The usual convention is that if `err != nil` then the other return values should not be depended on - this may surprise anyone who would fix something in `addHelpComment` without knowing the callsite depends on the first return value being correctly set even on an error.

adshmh: Thank you for the review. Fixed.

prow/plugins/trigger/generic-comment.go
```diff
@@ -29,6 +31,24 @@ import (
 	"k8s.io/test-infra/prow/plugins"
 )
 
+var (
+	testHelpRe          = regexp.MustCompile(`(?m)^(/test\s*\?|/list-tests)$`)
+	emptyTestRe         = regexp.MustCompile(`(?m)^/test\s*$`)
+	retestWithTargetRe  = regexp.MustCompile(`(?m)^/retest\s+\S+$`)
+	testWithAnyTargetRe = regexp.MustCompile(`(?m)^/test\s+\S+$`)
+
+	testWithoutTargetNote = "The `/test` command needs one or more targets. "
```
alvaroaleman: Wouldn't it make sense to add a trailing `\n` to all of them so they get their own line?

prow/plugins/trigger/generic-comment.go
```diff
@@ -29,6 +31,24 @@ import (
 	"k8s.io/test-infra/prow/plugins"
 )
 
+var (
+	testHelpRe          = regexp.MustCompile(`(?m)^(/test\s*\?|/list-tests)$`)
```
alvaroaleman: @cblecker WDYT about the `/test ?`/`/list-tests` commands?

prow/plugins/trigger/generic-comment.go
```diff
@@ -53,8 +73,18 @@ func handleGenericComment(c Client, trigger plugins.Trigger, gc github.GenericCo
 
 	refGetter := config.NewRefGetterForGitHubPullRequest(c.GitHubClient, org, repo, number)
 	presubmits := getPresubmits(c.Logger, c.GitClient, c.Config, org+"/"+repo, refGetter.BaseSHA, refGetter.HeadSHA)
+
+	pr, err := refGetter.PullRequest()
```
stevekuznetsov: Why do we need to do this so early? Calling it here will run it for every comment, If this call uses tokens, that scales very poorly.

prow/plugins/trigger/generic-comment.go
```diff
@@ -121,6 +147,12 @@ func handleGenericComment(c Client, trigger plugins.Trigger, gc github.GenericCo
 	if err != nil {
 		return err
 	}
+
+	addedComment, err := addHelpComment(c.GitHubClient, len(toTest), len(toSkip), gc.Body, org, repo, branch, pr.Number, presubmits, gc.HTMLURL, commentAuthor, c.Logger)
```
stevekuznetsov: We should not call a method that decides if it should or should do something -- decide here and conditionally add the help comment.

prow/plugins/trigger/generic-comment.go
```diff
@@ -183,3 +231,47 @@ func getContexts(combinedStatus *github.CombinedStatus) (sets.String, sets.Strin
 	}
 	return failedContexts, allContexts
 }
+
+func markdownList(names []string) string {
```
stevekuznetsov: nit: inline this if you have but one caller, unless a standalone function is useful for testing

prow/plugins/trigger/generic-comment.go
```diff
@@ -183,3 +231,47 @@ func getContexts(combinedStatus *github.CombinedStatus) (sets.String, sets.Strin
...
+}
+
+func buildHelpComment(githubClient githubClient, body, org, repo, branch string, number int, presubmits []config.Presubmit, HTMLURL, user, note string, logger *logrus.Entry) error {
+	available, err := availablePresubmits(githubClient, body, org, repo, branch, number, presubmits, logger)
+	if err != nil {
+		return err
+	}
+	var resp string
+	if len(available) > 0 {
+		resp = fmt.Sprintf("%sThe following jobs can be triggered by using their name after a `/test` message: %s", note, markdownList(available))
```
stevekuznetsov: This is not necessarily true. There's `Presubmit.RerunCommand` you should use instead.

prow/plugins/trigger/generic-comment.go
```diff
@@ -183,3 +231,47 @@ func getContexts(combinedStatus *github.CombinedStatus) (sets.String, sets.Strin
...
+	} else {
+		resp = fmt.Sprintf("%sNo presubmit jobs available for %s/%s=%s", note, org, repo, branch)
```
stevekuznetsov: nit: don't we idiomatically write `org/repo@branch` instead of `org/repo=branch`?

prow/plugins/trigger/generic-comment.go
cblecker: I'm not a fan of `/list-tests` as we don't use the list keyword anywhere. I don't mind `/test` with no keyword, or `/test ?` as they follow the current format.

prow/plugins/trigger/generic-comment.go
adshmh: Thank you for the review. Fixed.

prow/plugins/trigger/generic-comment.go
adshmh: Thank you for the review. Fixed.

prow/plugins/trigger/generic-comment.go
adshmh: Thank you for the review. Fixed.

prow/plugins/trigger/generic-comment.go
adshmh: Thank you for the review. Fixed.

prow/plugins/trigger/generic-comment.go
adshmh: Thanks. Inlined.

prow/plugins/trigger/generic-comment.go
adshmh: Thank you for the review. Fixed. Added a unit test to cover it.

prow/plugins/trigger/generic-comment.go
adshmh: Thank you for the review. Fixed.

adshmh:
> I expect this also closes #13059 as well?

Yes. Scenarios mentioned in #13059 are now covered here. If there are any others, please let  me know and I'll update the PR.

prow/plugins/trigger/trigger.go
```diff
@@ -81,6 +81,13 @@ func helpProvider(config *plugins.Configuration, enabledRepos []config.OrgRepo)
 		WhoCanUse:   "Anyone can trigger this command on a trusted PR.",
 		Examples:    []string{"/retest"},
 	})
+	pluginHelp.AddCommand(pluginhelp.Command{
```
alvaroaleman, line 84: I think we should also extend the description of `test` and mention that it will list all possible jobs when no jobs/an invalid job was specificed. Also change the `Usage` to `"/test [<job name>|all]"` to clarify that the target argument became optional.

prow/plugins/trigger/generic-comment.go
```diff
@@ -183,3 +226,40 @@ func getContexts(combinedStatus *github.CombinedStatus) (sets.String, sets.Strin
...
+func addHelpComment(githubClient githubClient, body, org, repo, branch string, number int, presubmits []config.Presubmit, HTMLURL, user, note string, logger *logrus.Entry) error {
+	available, err := availablePresubmits(githubClient, body, org, repo, branch, number, presubmits, logger)
+	if err != nil {
+		return err
+	}
+	var resp string
+	if len(available) > 0 {
+		resp = fmt.Sprintf("%sThe following commands are available to trigger jobs:%s\n\nUse `/test all` to run all jobs.",
+			note,
+			func(names []string) string {
```
alvaroaleman, line 239: The inline func to be able to do that in the arg is a bit too fancy for my taste, can we just build the string after the `len(available) > 0 ` check?

adshmh: Thank you for the review. Fixed.

prow/plugins/trigger/trigger.go
adshmh: Thank you for the review. Fixed.

cblecker: Does the PR need to be trusted for this command to work?

adshmh: Thank you for the review. Currently yes. It seemed to me that it could be confusing if available presubmits were listed in response to a help comment, when the user (who is not trusted) cannot run any of them.

alvaroaleman: @cblecker WDYT?

stevekuznetsov: Seems reasonable to me

stevekuznetsov:
/hold cancel
@adshmh :tada:

PR #16928, opened 2020-03-24
Azure: minor fix to windows jobs

chewong:
- Moved `K8S_SSH_PUBLIC_KEY_PATH` and `K8S_SSH_PRIVATE_KEY_PATH` to `preset-azure-windows` so that we don't have to define those environment variables in every job
- Disabled `ci-kubernetes-e2e-aks-engine-azure-master-windows`
- Added parentheses to `K8S_SSH_PUBLIC_KEY_PATH ` and `K8S_SSH_PRIVATE_KEY_PATH ` for `ci-kubernetes-e2e-aks-engine-azure-master-staging-windows` to mitigate job failure
- Increased `ci-kubernetes-e2e-aks-engine-azure-master-staging-windows` frequency to 8h
- Added `--aksengine-orchestratorRelease=1.18` to `ci-kubernetes-e2e-aks-engine-azure-master-windows-1909` to fix master-agent version mismatch

/hold
Discussing whether we should delete release-1.14-windows.yaml.


chewong:
Removed `--num-nodes` flag in `--test_args` since the test framework will help us determine the number of schedulable nodes before the test run. This will prevent any failure when we change the number of nodes in the API model.

adelina-t:
/lgtm

Also, from thread discussing with aks-engine folks on slack about removal of 1.14 jobs, it seems that we can go ahead with removing it. Coverage is sufficient in aks-engine.

PR #16930, opened 2020-03-24
Handle Slack API not using HTTP status codes

xrstf:
Slack's API does not seem to make use of HTTP status codes. I had a bad token configured and never saw an error being logged, because I got

```
HTTP/1.0 200 OK
{"ok":false,"error":"invalid_auth"}
```

I also updated the README slightly, because for me currently the form looks like this:

screenshot-2020-03-24-181430

and I guess the old `:bot` permission is what is now `.public`.

### DIFF
--- a/prow/cmd/crier/README.md
+++ b/prow/cmd/crier/README.md
@@ -61,16 +61,16 @@ The **OAuth Access Token** can be obtained as follows:
 1. Click **Create New App**.
 1. Provide an **App Name** (e.g. Prow Slack Reporter) and **Development Slack Workspace** (e.g. Kubernetes).
 1. Click **Permissions**.
-1. Add the `chat:write:bot` scope using the **Scopes** dropdown and **Save Changes**.
+1. Add the `chat:write.public` scope using the **Scopes / Bot Token Scopes** dropdown and **Save Changes**.
 1. Click **Install App to Workspace**
 1. Click **Allow** to authorize the Oauth scopes.
 1. Copy the **OAuth Access Token**.
 
-Once the *access token* is obtained, you can create a `secret` in the cluster using that value: 
+Once the *access token* is obtained, you can create a `secret` in the cluster using that value:
 
 ```shell
 kubectl create secret generic slack-token --from-literal=token=< access token >
-``` 
+```
 
 Furthermore, to make this token available to **Crier**, mount the *slack-token* `secret` using a `volume` and set the `--slack-token-file` flag in the deployment spec.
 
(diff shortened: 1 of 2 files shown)


prow/cmd/crier/README.md
```diff
@@ -61,16 +61,16 @@ The **OAuth Access Token** can be obtained as follows:
 1. Click **Create New App**.
 1. Provide an **App Name** (e.g. Prow Slack Reporter) and **Development Slack Workspace** (e.g. Kubernetes).
 1. Click **Permissions**.
-1. Add the `chat:write:bot` scope using the **Scopes** dropdown and **Save Changes**.
+1. Add the `chat:write.public` scope using the **Scopes / Bot Token Scopes** dropdown and **Save Changes**.
```
alvaroaleman: Are you sure this is needed?

prow/slack/client.go
```diff
@@ -83,19 +83,28 @@ func (sl *Client) urlValues() *url.Values {
...
+	apiResponse := struct {
+		Ok    bool   `json:"ok"`
+		Error string `json:"error"`
+	}{}
+
+	if err := json.Unmarshal(body, &apiResponse); err != nil {
+		return fmt.Errorf("API returned invalid JSON (%q): %v", string(body), err)
+	}
+
+	if resp.StatusCode != 200 || !apiResponse.Ok {
```
alvaroaleman: Any api docs in this?

prow/cmd/crier/README.md
xrstf: Without the .public permission, crier got `not_in_channel` as a response. And the Slack dropdown does not contain a `:bot` permission anymore.

prow/slack/client.go
xrstf: https://api.slack.com/messaging/webhooks#handling_errors

xrstf: Though the documentation says

> in most cases you'll receive a "HTTP 200" response with a **plain text** `ok`

a successful POST results in

```
{"ok":true,"channel":"CEK044BGA","ts":"1585070782.024400","message":{"type":"message","subtype":"bot_message","text":"my test text here","ts":"1585070782.024400","username":"prow","icons":{"emoji":":prow:"},"bot_id":"B010B1B429H"}}
```

alvaroaleman: wow, that is sad.
/lgtm
/approve

I guess you will be among the first to notice if this doesn't work anymore anyways :)

PR #16931, opened 2020-03-24
look for version in bazel-bin

mikedanese:
bazel-bin and bazel-genfiles have pointed to the same location since
0.25. -genfiles is deprecated.

This was test running:

../test-infra/scenarios/kubernetes_bazel.py --push

With bazel version 2.2.0 and 0.23.2.

Issue: https://github.com/bazelbuild/bazel/issues/8651

/assign @fejta


mikedanese:

PR #16933, opened 2020-03-24
add multi-bazel support to krte

mikedanese:
/assign @fejta


BenTheElder:
> variants.yaml is a symlink? I can't decide if that's cute or gross.

it's both. blame @justaugustus :+)

mikedanese:

PR #16936, opened 2020-03-24
Remove tracer

alvaroaleman:
Openshift stopped using it in openshift/release#7607 and it has no docs, so its unlikely anyone else was using it.

/assign @stevekuznetsov


stevekuznetsov:
RIP cc @kargakis 

/lgtm
/approve

PR #16943, opened 2020-03-25
Allow report templates to be configured as per org or repo.

droslean:
Adds a new field `report_templates` that allows having different report templates for each organization or repository.

Since the same logic has already been used for `pr_status_base_urls` and `default_decoration_configs`, so it makes sense to have different values in the report template configuration.

### DIFF
--- a/prow/ANNOUNCEMENTS.md
+++ b/prow/ANNOUNCEMENTS.md
@@ -1,9 +1,13 @@
 # Announcements
 
 New features added to each component:
+ - *March 25, 2020* Added a `report_templates` option to the Plank config that allows
+    to specify different report templates for each organization or a specific repository.
+    The `report_template` option is deprecated and it will be removed on *September 2020*
+    which is going to be replaced with the `*` value in `report_templates`.
  - *January 03, 2020* Added a `pr_status_base_urls` option to the Tide config
    that allows to specify different tide's URL for each organization or a specific repository.
-   The `pr_status_base_url` will be deprecated on *June 2020* and it will be replaces with the
+   The `pr_status_base_url` will be deprecated on *June 2020* and it will be replaced with the
    `*` value in `pr_status_base_urls`.
  - *November 05, 2019* The `config-updater` plugin supports update configs on build clusters
     by using [`clusters`](https://github.com/kubernetes/test-infra/tree/master/prow/plugins/updateconfig#usage).
--- a/prow/config/config.go
+++ b/prow/config/config.go
@@ -419,10 +419,14 @@ type Controller struct {
 
 	// ReportTemplateString compiles into ReportTemplate at load time.
 	ReportTemplateString string `json:"report_template,omitempty"`
-	// ReportTemplate is compiled at load time from ReportTemplateString. It
-	// will be passed a prowapi.ProwJob and can provide an optional blurb below
-	// the test failures comment.
-	ReportTemplate *template.Template `json:"-"`
+
+	// ReportTemplateStrings is a mapping of template comments.
+	// Use `org/repo`, `org` or `*` as a key.
+	ReportTemplateStrings map[string]string `json:"report_templates,omitempty"`
+
+	// ReportTemplates is a mapping of templates that is compliled at load
+	// time from ReportTemplateStrings.
+	ReportTemplates map[string]*template.Template `json:"-"`
 
 	// MaxConcurrency is the maximum number of tests running concurrently that
 	// will be allowed by the controller. 0 implies no limit.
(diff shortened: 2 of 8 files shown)


prow/ANNOUNCEMENTS.md
```diff
@@ -1,9 +1,13 @@
 # Announcements
 
 New features added to each component:
+ - *March 25, 2020* Added a `report_templates` option to the Plank config that allows
+    to specify different report templates for each organization or a specifig reposiroty.
+    The `report_template` option will be deprecated on *September 2020* and it will be replaced
```
alvaroaleman: No, it will get deprecated when this PR merges and removed in September 2020

prow/config/config.go
```diff
@@ -441,6 +449,21 @@ type Controller struct {
 	AllowCancellations *bool `json:"allow_cancellations,omitempty"`
 }
 
+// GetDefaultReportTemplate returns the template that belong to a specific repository.
+// If the repository doesn't exist in the report_templates configuration it will
+// inherit the values from its organization, otherwise the default values will be used.
+func (c *Controller) GetDefaultReportTemplate(org, repo string) *template.Template {
```
alvaroaleman: Can we rename this to `ReportTempateForRepo`? This isn't necessarily a default

prow/config/config.go
```diff
@@ -441,6 +449,21 @@ type Controller struct {
...
+	def := c.ReportTemplates["*"]
```
alvaroaleman: There is no need for this variable to exist, you can just `return c.Reporttemplates["*"]` at the end.

Also, I am wondering if templates are threadsafe? Not a problem introduced by this PR though.

prow/config/config.go
```diff
@@ -1919,6 +1940,29 @@ func ValidateController(c *Controller) error {
 	return nil
 }
 
+func ValidateReportTemplates(c *Controller) error {
+	if len(c.ReportTemplateString) > 0 {
```
alvaroaleman: Can we please do empty string checks via `if stringVar == ""`? It makes reading the code easier

prow/config/config.go
```diff
@@ -1919,6 +1940,29 @@ func ValidateController(c *Controller) error {
 	return nil
 }
 
+func ValidateReportTemplates(c *Controller) error {
```
alvaroaleman: I don't think this needs to be exporters. Also, can we rename to `DefaultAndValidateReportTemplate`? I wouldn't expect a fund called `ValidateReportTemplates` to mutate, but that's what we do

prow/config/config.go
```diff
@@ -1919,6 +1940,29 @@ func ValidateController(c *Controller) error {
...
+		logrus.Warning("report_template will be deprecated on September 2020, and it will be replaced with report_templates['*']")
+		c.ReportTemplateStrings = make(map[string]string)
+		c.ReportTemplateStrings["*"] = c.ReportTemplateString
+	}
+
+	c.ReportTemplates = make(map[string]*template.Template)
+	for orgRepo, value := range c.ReportTemplateStrings {
+		reportTmpl, err := template.New("Report").Parse(value)
+		if err != nil {
+			return fmt.Errorf("parsing template: %v", err)
```
alvaroaleman: Please make the life of ppl operating Prow better by adding the key to the error

prow/config/config.go
```diff
@@ -1919,6 +1940,29 @@ func ValidateController(c *Controller) error {
 	return nil
 }
 
+func ValidateReportTemplates(c *Controller) error {
+	if len(c.ReportTemplateString) > 0 {
+		if len(c.ReportTemplateStrings) > 0 {
+			return errors.New("both report_template and report_templates are specified")
+		}
+
+		logrus.Warning("report_template will be deprecated on September 2020, and it will be replaced with report_templates['*']")
```
alvaroaleman: Same as the announcement, it is deprecated and will be removed

alvaroaleman:
And some tests would be great :)

droslean:
> And some tests would be great :)

I didn't add any tests because this should be really straighforward. 
Perhaps unittest for the whole `Controller` would be nice.

prow/config/config.go
droslean: it is the default since we pass the org, repo vars. Also if there isn't any it will return the '*' default values which they are not related to any repo.

alvaroaleman: By your definition, we only have defaults. As mentioned, this isn't necessarily a default, because we may get something that is more specific than `'*'`

alvaroaleman:
> I didn't add any tests because this should be really straightforward.

The purpose of the tests is mostly to make sure that it doesn't break in the future, not to demonstrate in this PR that it currently works.

droslean:
WIP to add some tests.

prow/ANNOUNCEMENTS.md
```diff
@@ -1,9 +1,13 @@
 # Announcements
 
 New features added to each component:
+ - *March 25, 2020* Added a `report_templates` option to the Plank config that allows
+    to specify different report templates for each organization or a specifig reposiroty.
```
stevekuznetsov: sp: reposiroty

prow/config/config.go
```diff
@@ -441,6 +449,26 @@ type Controller struct {
 	AllowCancellations *bool `json:"allow_cancellations,omitempty"`
 }
 
+// GetDefaultReportTemplate returns the template that belong to a specific repository.
```
stevekuznetsov: bad godoc

prow/jenkins/controller.go
```diff
@@ -222,9 +222,10 @@ func (c *Controller) Sync() error {
 
 	var reportErrs []error
 	if !c.skipReport {
-		reportTemplate := c.config().ReportTemplate
 		reportTypes := c.cfg().GitHubReporter.JobTypesToReport
+		jConfig := c.config()
 		for report := range reportCh {
+			reportTemplate := jConfig.ReportTemplateForRepo(report.Spec.Refs)
```
stevekuznetsov: Why does this look different between `plank` and `jenkins-operator`?

droslean: Because `plank` is using `config config.Getter` but `jenkins-operator` has its own [method](https://github.com/kubernetes/test-infra/blob/master/prow/jenkins/controller.go#L113-L131) which it doesn't return a pointer. So if we just use `c.config().ReportTemplateForRepo(report.Spec.Refs)` we will get 
```
/controller.go:228:32: cannot call pointer method on c.config()
./controller.go:228:32: cannot take the address of c.config()
``` 
/shrug

PR #16945, opened 2020-03-25
prow/pod-utils: Default TerminationMessagePolicy fallback to logs

smarterclayton:
In failure scenarios this leads to a faster path to recover the
failure message, and helps in exceptional eviction cases (when disk
pressure exists). Callers already using TerminationMessagePolicy
explicitly are unaffected.


smarterclayton:
Tweaked to cover all containers

PR #16898, opened 2020-03-23
Migrate Kubeflow: part 1

clarketm:
- [x] Upload `kubeflow` kube config context.
- [x] Migrate Prow to use `--kubeconfig` (https://github.com/kubernetes/test-infra/pull/16906)

See: https://docs.google.com/document/d/17sA-rRBe30bM034nL353CgrXETfy2vMe2m_g_bB_ILY/edit#

PR #16961, opened 2020-03-26
Migrate Kubeflow: part 2

clarketm:
Successor to #16898


jlewi:
/lgtm
/approve

I think this is safe to merge as we can always revert it.

clarketm:
> I think this is safe to merge as we can always revert it.

Great, thanks. Removing the wip label now.

PR #16963, opened 2020-03-26
Use release/stable for Kops master pre-submit e2e

hakman:
Kops pre-submit e2e is failing because of https://github.com/kubernetes/kops/pull/8783. Until it is merged we need to use the release/stable marker.

/cc @rifelpet


rifelpet:
release/latest was pointing to a 1.18 tag but after the k8s 1.18 release it now points to a 1.19 tag, exposing the basic auth issue mentioned above. we'll revert this once we get that sorted out.

/lgtm
/approve

Issue #16793, opened 2020-03-15
prow should recommend relevant approvers


BenTheElder:
**What happened**: Prow suggested a root approver for review when the PR had a single file covered by leaf approvers that did not include the root approver.

The PR only touches this file `config/testgrids/conformance/conformance-all.yaml `

With `config/testgrids/conformance/OWNERS` existing with the following contents:
```
# See the OWNERS docs at https://go.k8s.io/owners

approvers:
- spiffxp
- timothysc
```

The raw comment posted by Prow is:
```
[APPROVALNOTIFIER] This PR is **NOT APPROVED**

This pull-request has been approved by: *<a href="https://github.com/kubernetes/test-infra/pull/16292#" title="Author self-approved">lubinsz</a>*
To complete the [pull request process](https://git.k8s.io/community/contributors/guide/owners.md#the-code-review-process), please assign **bentheelder**
You can assign the PR to them by writing `/assign @bentheelder` in a comment when ready.

The full list of commands accepted by this bot can be found [here](https://go.k8s.io/bot-commands?repo=kubernetes%2Ftest-infra).

<details open>
Needs approval from an approver in each of these files:

- **[config/testgrids/conformance/OWNERS](https://github.com/kubernetes/test-infra/blob/master/config/testgrids/conformance/OWNERS)**

Approvers can indicate their approval by writing `/approve` in a comment
Approvers can cancel approval by writing `/approve cancel` in a comment
</details>
<!-- META={"approvers":["bentheelder"]} -->
```

https://github.com/kubernetes/test-infra/pull/16292#issuecomment-593761410

There's no reason I should be suggested here.

**What you expected to happen**:

Since `config/testgrids/conformance/OWNERS` has existing approvers, is the only directory in need of approval, and the comment is about getting **approval**, the approvers from this file should be suggested. I am not one of them, so I should not be suggested.

**How to reproduce it (as minimally and precisely as possible)**: 

Wait for prow to make suggestions for approval on a PR.

**Please provide links to example occurrences, if any**:

https://github.com/kubernetes/test-infra/pull/16292#issuecomment-593761410

**Anything else we need to know?**:

This behavior leads to excessive assignment of root approvers, we should be spreading workload to leaf owners.

cc @cjwagner 
/area prow

BenTheElder:
[...]

PR #16972, opened 2020-03-26
backfill missing reviewers from approvers

BenTheElder:
add owners to reviewers in any OWNERS file without a reviewers entry

mitigates https://github.com/kubernetes/test-infra/issues/16793

**NOTE**: This PR does NOT attempt to correct the OWNERS files in any other way, this change was plenty large enough. Any approvers / reviewers that should be changed should be done in a different PR, this PR is purely copying approves into reviewers where no reviewers exist so we will assign leave reviewers with context on changes instead of always assigning top level OWNERS.

This has been especially bad for `config/` where prowjob directories have intended leaf owners, but they nearly all have `approvers` only so those people do not in fact wind up CCed on those PRs, instead some of the root OWNERS are reviewing all job PRs which is not sustainable.

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -1,5 +1,10 @@
 # See the OWNERS docs at https://go.k8s.io/owners
 
+reviewers:
+- fejta # lead
+- spiffxp # lead
+- stevekuznetsov # lead
+- test-infra-oncall # oncall
 approvers:
 - fejta # lead
 - spiffxp # lead
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -205,3 +205,13 @@ aliases:
     - frapposelli
 ## END CUSTOM CONTENT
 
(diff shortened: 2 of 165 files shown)


BenTheElder:
these names were already in the OWNERS files. I added no new names. :man_shrugging:

neolit123:
i recall, @fejta once recommended to not have the list of reviewers if they match the approvers.

BenTheElder:
that doesn't work, see https://github.com/kubernetes/test-infra/issues/16793 for multiple examples of `approvers` not being used for assigning reviews.

having a reviewers entry is more granular anyhow. you can explicitly tweak both lists.

spiffxp:
/lgtm
I opened https://github.com/kubernetes/test-infra/issues/16988 to address the verify-owners failures

Issue #14046, opened 2019-08-23
stop active ci runs when PR is closed


kikisdeliveryservice:
**What would you like to be added**:
I would like CI runs to be automatically aborted when a PR with active runs is closed.

**Why is this needed**:
This will free up testing resources by stopping unneeded tests immediately

cc: @stevekuznetsov

PR #16973, opened 2020-03-26
Abort jobs on PR close

alvaroaleman:
This PR:
* Changes `trigger` to set ProwJobs to aborted when it receives a PR close event
* Adds handling for aborted jobs to Plank and the Jenkins controller

The motivation for the change is to free resources occupied by tests we do not care about and to not create useless notifications.

Fixes https://github.com/kubernetes/test-infra/issues/14046


prow/plugins/trigger/pull-request.go
```diff
@@ -141,10 +149,63 @@ func handlePR(c Client, trigger plugins.Trigger, pr github.PullRequestEvent) err
...
+	}
+
+	var errs []error
+	for _, job := range jobs.Items {
+		// Do not abort jobs that already completed
+		if job.Complete() {
+			continue
+		}
+		job.Status.State = prowapi.AbortedState
+		if _, err := c.ProwJobClient.Update(&job); err != nil {
```
alvaroaleman: I've explicitly chosen to `Update` and not to `Patch` so this will fail if another actor changed the job in the meantime. Consequently, this is a best-effort solution that may fail incorrectly in some cases like `crier` updating the job wrong. IMHO that is better allowing races with the responsible `agent` for the jobs status.

prow/apis/prowjobs/v1/types.go
```diff
@@ -53,7 +53,7 @@ type ProwJobState string
 const (
 	// TriggeredState means the job has been created but not yet scheduled.
 	TriggeredState ProwJobState = "triggered"
-	// PendingState means the job is scheduled but not yet running.
+	// PendingState means the job is currently running and we are waiting for it to finish.
```
stevekuznetsov: I've dreamt about fixing this at some point. We should have a `Running` state, right?

prow/jenkins/controller_test.go
```diff
@@ -1050,3 +1052,65 @@ func TestOperatorConfig(t *testing.T) {
 		}
 	}
 }
+
+func TestSyncAbortedJob(t *testing.T) {
```
stevekuznetsov: Is it possible to test that a failure to abort the jenkins job on one run will lead to a re-try in the future?

prow/pjutil/pjutil_test.go
```diff
@@ -262,6 +262,7 @@ func TestPartitionActive(t *testing.T) {
 
 		pending   map[string]struct{}
 		triggered map[string]struct{}
+		aborted   map[string]struct{}
```
stevekuznetsov: nit: could we change these to be `sets.String` (maybe as an unrelated prereq?)

prow/plank/controller.go
```diff
@@ -511,6 +513,27 @@ func (c *Controller) syncPendingJob(pj prowapi.ProwJob, pm map[string]corev1.Pod
 	return c.prowJobClient.Patch(c.ctx, pj.DeepCopy(), ctrlruntimeclient.MergeFrom(&prevPJ))
 }
 
+func (c *Controller) syncAbortedJob(pj prowapi.ProwJob, pm map[string]corev1.Pod, _ chan<- prowapi.ProwJob) error {
+	if pj.Status.State != prowapi.AbortedState || pj.Complete() {
+		return nil
+	}
+
+	pod, podExists := pm[pj.ObjectMeta.Name]
```
stevekuznetsov: nit: no need to leak these out of the `if` scope

```go
if pod, podExists :=...
```

prow/plugins/trigger/pull-request.go
```diff
@@ -141,10 +149,63 @@ func handlePR(c Client, trigger plugins.Trigger, pr github.PullRequestEvent) err
...
 	return nil
 }
 
+func abortAllJobs(c Client, pr *github.PullRequest) error {
+	selector, err := labelSelectorForPR(pr)
+	if err != nil {
+		return fmt.Errorf("failed to construct label selector: %w", err)
+	}
+
+	listOpts := metav1.ListOptions{LabelSelector: selector.String()}
```
stevekuznetsov: nit: we can inline this to the `List()` call?

prow/plugins/trigger/pull-request.go
stevekuznetsov: Can we add this comment to the code?

prow/plugins/trigger/pull-request.go
```diff
@@ -141,10 +149,63 @@ func handlePR(c Client, trigger plugins.Trigger, pr github.PullRequestEvent) err
...
+		job.Status.State = prowapi.AbortedState
+		if _, err := c.ProwJobClient.Update(&job); err != nil {
+			errs = append(errs, fmt.Errorf("failed to abort job %s: %w", job.Name, err))
+		}
+	}
+
+	return utilerrors.NewAggregate(errs)
+}
+
+func labelSelectorForPR(pr *github.PullRequest) (klabels.Selector, error) {
```
stevekuznetsov: Wasn't there a utility to make a selector from a map, and in that case couldn't we make one from

```go
requirements := map[string]string{
    kube.OrgLabel: pr.Base.Repo.Owner.Login,
// ...
}
```
?

stevekuznetsov: or if not, 

```go
selector := klabels.NewSelector()
for key, value := range map[string]string{
    kube.OrgLabel: pr.Base.Repo.Owner.Login,
// ...
} {
    req, err := klabels.NewRequirement(key, selection.Equals, []string{value})
// ...
    selector.Add(*req)
}
return selector
```

prow/apis/prowjobs/v1/types.go
alvaroaleman: Yeah so very technically it may not be running but e.G. `pending` if the cluster has no capacity or sth. But I find this super confusing, because `triggered` to me sounds like it was what `pending` is (We `triggered` the test workload and are waiting for it).

prow/jenkins/controller_test.go
alvaroaleman: Not trivially in a unit test, because the logic for that is spread around:
* We call `syncAbortedJob` for aborted job that are not `completed` (the latter indicated by `.Status.CompletionTime` being non-nil
* in `syncAbortedJobs` we return errors and do not set the `.Status.CompletionTime` if we had one - For that I will add a test

alvaroaleman:
@stevekuznetsov all feedback addressed, PTAL

prow/jenkins/controller_test.go
stevekuznetsov: That's what I thought

prow/plank/controller.go
```diff
@@ -511,6 +513,26 @@ func (c *Controller) syncPendingJob(pj prowapi.ProwJob, pm map[string]corev1.Pod
...
+	if pj.Status.State != prowapi.AbortedState || pj.Complete() {
+		return nil
+	}
+
+	if pod, podExists := pm[pj.Name]; podExists {
+		client, ok := c.buildClients[pj.ClusterAlias()]
+		if !ok {
+			return fmt.Errorf("unknown cluster alias %q", pj.ClusterAlias())
+		}
+		if err := client.Delete(c.ctx, pod.DeepCopy()); err != nil && !kerrors.IsNotFound(err) {
```
cjwagner: We need to respect the value of `AllowCancellations` in both plank and jenkins-operator. I don't think this is acceptable if `AllowCancellations` is false. This could share code with the `terminateDupes` function.

alvaroaleman: `AllowCancellations` is deprecated and has a ` // TODO(fejta): delete this Mar 2020` comment on it.

Should I create a follow-up to remove it?

cjwagner: 😆 Yes that sounds good too.

PR #16975, opened 2020-03-26
Enabling aks-engine Windows conatinerd periodic test jobs

marosset:


andyzhangx:
so many e2e tests are enabled. Do we have subs quota for those tests? 
cc @ritazh

chewong:
> so many e2e tests are enabled. Do we have subs quota for those tests?
> cc @ritazh

I think we do. We have 200 cores per region and we run a cleanup service periodically to delete resource groups that are older than 6 hours to free up the cores. We encountered some quota issues awhile ago because of https://github.com/kubernetes/test-infra/pull/16854 (there are only 10 spot cores per region) but I have reverted the PR.

marosset:
@andyzhangx @ritazh @chewong I plan on reducing frquency of the new Windows containerd jobs in a few days. Right now we are trying to establish a baseline of what the gaps are between containerd and dockershim on Windows.

We've been reducing frequency of other Windows jobs and I'm going to reduce the frequency further with another PR I'm authoring now.

PR #16982, opened 2020-03-27
Use release/stable marker for most Kops jobs

hakman:
Kops periodic tests are still failing in TestGrid. This should move all our e2e to release/stable and keep one that uses ci/latest to see if we bump into any major issues.
Kops is usually 1 release behind Kubernetes so, release branch is not really a big deal.

### DIFF
--- a/config/jobs/kubernetes/kops/kops-periodics-versions.yaml
+++ b/config/jobs/kubernetes/kops/kops-periodics-versions.yaml
@@ -33,38 +33,6 @@ periodics:
     testgrid-tab-name: kops-aws-k8s-latest
 
 - interval: 2h
-  name: e2e-kops-aws-k8s-1-19
-  labels:
-    preset-service-account: "true"
-    preset-aws-ssh: "true"
-    preset-aws-credential: "true"
-  decorate: true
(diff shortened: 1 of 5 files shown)


config/jobs/kubernetes/kops/kops-periodics-versions.yaml
```diff
@@ -32,38 +32,6 @@ periodics:
     testgrid-dashboards: google-aws, sig-cluster-lifecycle-kops
     testgrid-tab-name: kops-aws-k8s-latest
 
-- interval: 2h
```
rifelpet: any reason to remove this? are you thinking that its a duplicate of the k8s-latest job above? that probably makes sense.

hakman: Yes, pretty much duplicate. I did not see a good reason for it to exist.

rifelpet: ok that sounds good to me. as long as we keep at least one job testing with the latest k8s commit then we should be sufficient here 👍

PR #16989, opened 2020-03-27
Add IPVS presubmit job based on pull-kubernetes-e2e-gce

aojea:
Current job wasn't working, failing[ with following error](https://prow.k8s.io/log?job=pull-kubernetes-e2e-gci-gce-ipvs&id=1243588299969269760)

```
suffix=/pull-kubernetes-e2e-gci-gce-ipvs --allow-dup' finished in 636.095µs
W0327 17:19:15.675] 2020/03/27 17:19:15 process.go:96: Saved XML output to /workspace/_artifacts/junit_runner.xml.
W0327 17:19:15.680] 2020/03/27 17:19:15 main.go:316: Something went wrong: failed to acquire k8s binaries: error starting /go/src/k8s.io/release/push-build.sh --nomock --verbose --noupdatelatest --bucket=kubernetes-release-pull --ci --gcs-suffix=/pull-kubernetes-e2e-gci-gce-ipvs --allow-dup: fork/exec /go/src/k8s.io/release/push-build.sh: no such file or directory
``` 

This adds the jobs based on current pull-kubernetes-e2e-gce job

https://github.com/kubernetes/test-infra/blob/fc9a8d3c7a4ee5160a3a398500c9b7a6abbe2a3d/config/jobs/kubernetes/sig-cloud-provider/gcp/gcp-gce.yaml#L17-L27

### DIFF
--- a/config/jobs/kubernetes/sig-network/sig-network-misc.yaml
+++ b/config/jobs/kubernetes/sig-network/sig-network-misc.yaml
@@ -3,32 +3,44 @@ presubmits:
   - name: pull-kubernetes-e2e-gci-gce-ipvs
     always_run: false
     optional: true
-    skip_report: true
+    skip_report: false
     branches:
     - master
+    annotations:
+      testgrid-create-test-group: "true"
     labels:
       preset-service-account: "true"
       preset-k8s-ssh: "true"
-    annotations:
-      testgrid-create-test-group: 'true'
+      preset-bazel-scratch-dir: "true"
+      preset-bazel-remote-cache-enabled: "true"
+      preset-pull-kubernetes-e2e: "true"
+      preset-pull-kubernetes-e2e-gce: "true"
     spec:
       containers:
       - args:
-        - --timeout=170
-        - --bare
+        - --root=/go/src
+        - --repo=k8s.io/kubernetes=$(PULL_REFS)
+        - --repo=k8s.io/release
+        - --upload=gs://kubernetes-jenkins/pr-logs
+        - --timeout=105
         - --scenario=kubernetes_e2e
         - --
+        - --build=bazel
+        - --cluster=
         - --env=KUBE_PROXY_MODE=ipvs
-        - --extract=ci/latest
-        - --gcp-master-image=gci
+        - --extract=local
         - --gcp-node-image=ubuntu
         - --gcp-zone=us-west1-b
         - --ginkgo-parallel=30
         - --provider=gce
         - --stage=gs://kubernetes-release-pull/ci/pull-kubernetes-e2e-gci-gce-ipvs
-        - --test_args=--ginkgo.skip=\[Serial\]|\[Disruptive\]|\[Flaky\]|\[Feature:.+\] --minStartupPods=8
-        - --timeout=150m
-        image: gcr.io/k8s-testimages/kubekins-e2e:v20200327-9ba073a-master
+        - --test_args=--ginkgo.skip=\[Slow\]|\[Serial\]|\[Disruptive\]|\[Flaky\]|\[Feature:.+\] --minStartupPods=8
+        - --timeout=80m # thinking about making this longer? don't! 80m is a hard cap, and should get down to no more than 60m.
+        image: gcr.io/k8s-testimages/kubekins-e2e:v20200326-e946722-master
+        resources:
+          requests:
+            memory: "6Gi"
+
   kubernetes/dns:
   - name: pull-kubernetes-dns-test
     branches:


config/jobs/kubernetes/sig-network/sig-network-misc.yaml
```diff
@@ -3,32 +3,44 @@ presubmits:
   - name: pull-kubernetes-e2e-gci-gce-ipvs
     always_run: false
     optional: true
-    skip_report: true
+    skip_report: false
```
aojea: I initially suggested to skip_report, but I found that since it's optional it won't run until it is called, 
making it difficult to debug because you can't see the output from the PR you are calling it 
my fault

config/jobs/kubernetes/sig-network/sig-network-misc.yaml
```diff
@@ -3,32 +3,44 @@ presubmits:
...
+    annotations:
+      testgrid-create-test-group: "true"
     labels:
       preset-service-account: "true"
       preset-k8s-ssh: "true"
-    annotations:
-      testgrid-create-test-group: 'true'
+      preset-bazel-scratch-dir: "true"
+      preset-bazel-remote-cache-enabled: "true"
+      preset-pull-kubernetes-e2e: "true"
```
aojea: copy paste, is this ok?

config/jobs/kubernetes/sig-network/sig-network-misc.yaml
```diff
@@ -3,32 +3,44 @@ presubmits:
...
+        - --extract=local
         - --gcp-node-image=ubuntu
         - --gcp-zone=us-west1-b
         - --ginkgo-parallel=30
         - --provider=gce
         - --stage=gs://kubernetes-release-pull/ci/pull-kubernetes-e2e-gci-gce-ipvs
-        - --test_args=--ginkgo.skip=\[Serial\]|\[Disruptive\]|\[Flaky\]|\[Feature:.+\] --minStartupPods=8
-        - --timeout=150m
-        image: gcr.io/k8s-testimages/kubekins-e2e:v20200324-0897dba-master
+        - --test_args=--ginkgo.skip=\[Slow\]|\[Serial\]|\[Disruptive\]|\[Flaky\]|\[Feature:.+\] --minStartupPods=8
```
aojea: @saschagrunert @justaugustus 
do we need something for this `--stage=gs://kubernetes-release-pull/ci/pull-kubernetes-e2e-gci-gce-ipvs` ? 
cc: @dims

config/jobs/kubernetes/sig-network/sig-network-misc.yaml
```diff
@@ -3,32 +3,44 @@ presubmits:
...
+        - --upload=gs://kubernetes-jenkins/pr-logs
+        - --timeout=105
         - --scenario=kubernetes_e2e
         - --
+        - --build=bazel
+        - --cluster=
         - --env=KUBE_PROXY_MODE=ipvs
-        - --extract=ci/latest
-        - --gcp-master-image=gci
+        - --extract=local
```
aojea: extract local or ci/latest?

config/jobs/kubernetes/sig-network/sig-network-misc.yaml
BenTheElder: trying to parse this ... " since the job optional, it won't run until it is called,"

`always_run` and `run_if_changed` control if it runs.
`optional` is a way to make something that runs not required for merge
`skip_report` is a way to make it not post a comment or status on the PR. you can still see the results in prow.k8s.io / testgrid

aojea: hehe, I misunderstood the `optional` thing :man_shrugging: 

Let me rephrase since this job can only run manually and is optional, there is no risk that somebody runs it accidentally and/or block any PR, thus reporting the status if it runs should not be a problem

andrewsykim: > Let me rephrase: since this job can only run manually and is optional, there is no risk that somebody runs it accidentally and/or block any PR, thus reporting the status when runs should not be a problem, because it means that , who runs the job, wants to see the result sweat_smile

Make sense to me and this was my original intent, thanks for fixing @aojea!

config/jobs/kubernetes/sig-network/sig-network-misc.yaml
dims: nope, it's a new directory in an existing bucket.

andrewsykim:
/hold 

In case there are other comments

dims:
we can always rev .. let's give this a shot

/hold cancel

Issue #16296, opened 2020-02-14
hack/verify-deps.sh fails on master


matthyx:
**What happened**:
```
hack/verify-deps.sh
+ bazel test --test_output=streamed //hack:verify-deps
$TEST_TMPDIR defined: output root default is '/mnt/bazel/' and max_idle_secs default is '15'.
$TEST_TMPDIR defined: output root default is '/mnt/bazel/' and max_idle_secs default is '15'.
WARNING: Streamed test output requested. All tests will be run locally, without sharding, one at a time
ERROR: Skipping '//hack:verify-deps': no such target '//hack:verify-deps': target 'verify-deps' not declared in package 'hack' (did you mean 'verify-deps.sh'?) defined by /home/mbertsch/workspace/go/src/k8s.io/test-infra/hack/BUILD.bazel
ERROR: no such target '//hack:verify-deps': target 'verify-deps' not declared in package 'hack' (did you mean 'verify-deps.sh'?) defined by /home/mbertsch/workspace/go/src/k8s.io/test-infra/hack/BUILD.bazel
INFO: Elapsed time: 1.712s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (0 packages loaded)
FAILED: Build did NOT complete successfully (0 packages loaded)
```

**What you expected to happen**:
PASS

**How to reproduce it (as minimally and precisely as possible)**:
- checkout master
- pull latest
- run `hack/verify-deps.sh`

/assign @fejta

PR #16993, opened 2020-03-27
Use correct target for verify-deps

fejta:
Fixes https://github.com/kubernetes/test-infra/issues/16296

This is passing https://testgrid.k8s.io/sig-testing-misc#post-bazel&include-filter-by-regex=verify-deps just the wrong target in the bash script

/assign @matthyx

### DIFF
--- a/hack/verify-deps.sh
+++ b/hack/verify-deps.sh
@@ -23,4 +23,4 @@ if ! command -v bazel &>/dev/null; then
 fi
 
 set -o xtrace
-bazel test --test_output=streamed //hack:verify-deps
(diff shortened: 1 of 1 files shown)


matthyx:
/lgtm
/hold
How come I didn't see it?

fejta:
/hold cancel

See what?

matthyx:
That it was failing on local machine.

fejta:
It wasn't failing, the target doesn't exist: `ERROR: no such target '//hack:verify-deps':`

fejta:
/uncc @michelle192837 
hmm is prow and/or github sad?

matthyx:
Ok thanks :-)

PR #16994, opened 2020-03-27
Remove the deprecated allow_cancellations option

alvaroaleman:
/assign @cjwagner


alvaroaleman:
@cjwagner seems that did not get handled, can you re-lgtm?

PR #16997, opened 2020-03-27
Make verify-owners limit trusted user checks.

cjwagner:
This is an imperfect and dirty quick fix to limit verify-owners token usage.

/assign @stevekuznetsov

### DIFF
--- a/prow/plugins/verify-owners/verify-owners.go
+++ b/prow/plugins/verify-owners/verify-owners.go
@@ -243,7 +243,7 @@ func handle(ghc githubClient, gc git.ClientFactory, roc repoownersClient, log *l
 
 	// If OWNERS_ALIASES file exists, get all aliases.
 	// If the file was modified, check for non trusted users in the newly added owners.
-	nonTrustedUsers, repoAliases, err := nonTrustedUsersInOwnersAliases(ghc, log, triggerConfig, org, repo, r.Directory(), modifiedOwnerAliasesFile.Patch, ownerAliasesModified, skipTrustedUserCheck)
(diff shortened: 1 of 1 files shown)


prow/plugins/verify-owners/verify-owners.go
```diff
@@ -482,7 +474,11 @@ func nonTrustedUsersInOwners(ghc githubClient, log *logrus.Entry, triggerConfig
 
 // checkIfTrustedUser looks for newly addded owners by checking if they are in the patch
 // and then checks if the owner is a trusted user.
-func checkIfTrustedUser(ghc githubClient, log *logrus.Entry, triggerConfig plugins.Trigger, owner, patch, fileName, org, repo string, nonTrustedUsers map[string][]string, repoAliases repoowners.RepoAliases) (map[string][]string, error) {
+func checkIfTrustedUser(ghc githubClient, log *logrus.Entry, triggerConfig plugins.Trigger, owner, patch, fileName, org, repo string, nonTrustedUsers map[string][]string, trustedUsers sets.String, repoAliases repoowners.RepoAliases) (map[string][]string, error) {
+	// cap the number of checks to avoid exhausting tokens in case of large OWNERS refactors.
+	if len(nonTrustedUsers)+trustedUsers.Len() > 50 {
```
alvaroaleman: So this effectively means that on larger changes, we might end up merging invalid owners since `nonTrustedUsers` could be empty, right?

alvaroaleman: Also not sure if this correct, because the `nonTrusteduser` s are a `map[string][]string`, you re checking the key lenght here (repo I guess?) rather than the unique values

cjwagner: Correct. We'll still validate that the files parse and do other checks included in the plugin, but after seeing 51 unique users we'll stop checking if they are trusted to avoid using too many tokens.

cjwagner: The keys are the untrusted owners and the values are the files containing the untrusted owner

PR #17003, opened 2020-03-28
Run cluster-api e2e and integration tests only for go, sh and makefile changes

tahsinrahman:
Fixes: https://github.com/kubernetes-sigs/cluster-api/issues/2803

/cc @vincepri

### DIFF
--- a/config/jobs/kubernetes-sigs/cluster-api/cluster-api-presubmits.yaml
+++ b/config/jobs/kubernetes-sigs/cluster-api/cluster-api-presubmits.yaml
@@ -91,7 +91,7 @@ presubmits:
       preset-kind-volume-mounts: "true"
     decorate: true
     path_alias: sigs.k8s.io/cluster-api
-    always_run: true
+    run_if_changed: '\.go$|Makefile|\.sh$|go\.mod'
     spec:
       containers:
       - image: gcr.io/k8s-testimages/kubekins-e2e:v20200327-9ba073a-master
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes-sigs/cluster-api/cluster-api-presubmits.yaml
```diff
@@ -118,7 +118,7 @@ presubmits:
       - release-0.2
     path_alias: sigs.k8s.io/cluster-api
     optional: true
-    always_run: true
+    run_if_changed: '\.go$|Makefile|\.sh$'
```
vincepri: Might be worth also to add `go.mod` files ?

tahsinrahman: added!

ncdc:
Do we need to react to go.sum changes too (absent any go.mod changes)?

vincepri:
We could add it to the list, but afaik go.sum is dependent on go.mod, and we have a different check if those haven't been updated

ncdc:
The e2e has a couple of .conf files that it uses. I assume we should add that to the regex. And is there any local yaml that the e2e test loads & uses?

ncdc:
I really want to err on the side of caution here. I'd much rather run tests when we don't need them, than skip running tests when we should be.

vincepri:
hmm, that's fair, maybe we can just leave as is then

vincepri:
Closing this for now given https://github.com/kubernetes-sigs/cluster-api/issues/2803#issuecomment-606711604

/close

PR #13328, opened 2019-07-04
sig-release: Blockade changes to critical k/release tooling

justaugustus:
ref: https://github.com/kubernetes/release/issues/816
Signed-off-by: Stephen Augustus <saugustus@vmware.com>

PR #17013, opened 2020-03-30
releng: Loosen blockade regex on kubernetes/release repo

justaugustus:
Since the [blockade was initiated](https://github.com/kubernetes/test-infra/pull/13328), we've [migrated several of the
blockaded tools over to go and added tests for them](https://github.com/kubernetes/release/issues/918).

This also fixes a bug where the blockade was being triggered
unintentionally in subdirectories (like `images/build/cross`) because
the regex was not properly grouped (`^foo|bar|baz` as opposed to
`^(foo|bar|baz)`.

/assign @cblecker @BenTheElder 
cc: @liggitt (ref: https://github.com/kubernetes/release/pull/1208#issuecomment-602918850)
ref: https://github.com/kubernetes/release/issues/816

### DIFF
--- a/config/prow/plugins.yaml
+++ b/config/prow/plugins.yaml
@@ -166,7 +166,7 @@ blockades:
 - repos:
   - kubernetes/release
   blockregexps:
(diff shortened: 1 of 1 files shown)


cblecker:
/approve
LGTM to releng

PR #17016, opened 2020-03-30
Add DecorateAllJobs option to job config

### LINKED ISSUE #16650: Jobs should default to decorated
/area prow

We should either always decorate jobs or at least provide an option to control whether jobs default to decorated or undecorated.

adshmh:
Add a configuration option to allow all jobs to be decorated by-default.
Closes #16650

### DIFF
--- a/prow/config/config.go
+++ b/prow/config/config.go
@@ -93,6 +93,9 @@ type JobConfig struct {
 	// ProwYAMLGetter is the function to get a ProwYAML. Tests should
 	// provide their own implementation.
 	ProwYAMLGetter ProwYAMLGetter `json:"-"`
+
+	// DecorateAllJobs determines whether all jobs are decorated by default
+	DecorateAllJobs bool `json:"decorate_all_jobs,omitempty"`
 }
 
 // ProwConfig is config for all prow controllers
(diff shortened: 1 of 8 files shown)


alvaroaleman:
I think this would be a nice addition, maybe even make it default to `true`. WDYT @fejta @stevekuznetsov ?

prow/config/config.go
```diff
@@ -1045,21 +1048,21 @@ func mergeJobConfigs(a, b JobConfig) (JobConfig, error) {
 }
 
 func setPresubmitDecorationDefaults(c *Config, ps *Presubmit, repo string) {
-	if ps.Decorate {
+	if c.DecorateAllJobs || ps.Decorate {
```
fejta, line 1051: This prevents someone from defining an undecorated job. We need to be able to define all three things:
* Jobs which inherit the default behavior
* Jobs which must be decorated
* Jobs which must be undecorated.

Long term the default behavior should always be decorated. However I would like to see a migration period:
1) AKA start out with the ability to opt into a decorate default
2) Later the ability to opt out of decorated default
3) Finally requiring decorated defaults (but people should still be able to declare a job that does not use decoration)

fejta:
> maybe even make it default to true

a) We need to preserve the ability for jobs to opt out of this behavior
b) Before making this an opt-out behavior we need to ensure all prow.k8s.io jobs specify `decorate: false` that currently specify nothing.
  - Might be more user friendly to ship a release that notices jobs which do not specify decorate and warn them that this will soon change
c) This will need a breaking change in prow/ANNOUNCE.md

adshmh:
> You also need to update `decorationRequested` in `prow/config/config.go`

Thank you for the review. Fixed.

prow/config/config.go
adshmh: Thank you for the detailed review. I have added an `Undecorated` option to allow opting out of default behavior and define undecorated jobs.

prow/config/jobs.go
```diff
@@ -415,6 +415,9 @@ func NewGitHubDeferredChangedFilesProvider(client githubClient, org, repo string
 type UtilityConfig struct {
 	// Decorate determines if we decorate the PodSpec or not
 	Decorate bool `json:"decorate,omitempty"`
+	// Undecorated determines if the job should not be decorated. Overrides
+	// other decorate options.
+	Undecorated bool `json:"undecorated,omitempty"`
```
fejta, line 420: `Decorate *bool` `json:"decorate,omitempty"` is generally how we do this, which makes the behavior clear:

```yaml
# will decorate
decorate: true
---
# will not decorate
decorate: false
---
# will use default behavior
decorate: null # equivalent to it being missing
```

This seems unnecessarily messy -- aka what happens for `decorate: true, undecorated: true` jobs?

fejta: A couple examples:

https://github.com/kubernetes/test-infra/blob/0110928adec52dd109255b9ebec3b7a8a1c89197/prow/config/jobs.go#L99-L102


https://github.com/kubernetes/test-infra/blob/0110928adec52dd109255b9ebec3b7a8a1c89197/prow/config/branch_protection.go#L31-L32

prow/config/config.go
stevekuznetsov: I don't think it's safe to change a current config with no `decorate` set to one that has `decorate: true` so if we follow that path we need a migration period where you have users change their config to set `undecorated` to true

prow/config/jobs.go
adshmh: Thank you for the review. Fixed.

adshmh:
> /hold
> 
> Want to add something to https://github.com/kubernetes/test-infra/blob/master/prow/ANNOUNCEMENTS.md#new-features noting that people can control whether jobs decorate by default with the `decorate_all_jobs` setting?

Thank you for the reminder. Done.

PR #17031, opened 2020-03-31
Enable Chaos Monkey for ClusterLoader2 presubmit

jprzychodzen:
Chaos Monkey already runs for an experimental jobs and tests are green. Enabling for presubmit job to prevent breaking Node Killer with ClusterLoader2 changes.

* [Testgrid](https://testgrid.k8s.io/sig-scalability-experiments#gce-cos-master-scalability-100-nodekiller)

/sig scalability

Ref. kubernetes/kubernetes#89051
/assign @mm4tt


mm4tt:
@oxddr, FYI

PR #12783, opened 2019-05-28
Enable mergecommitblocker plugin for k/k

nikhita:
It's necessary for k/k because the publishing-bot breaks whenever it sees merge commits, and for each such case (which is rare), we need to explicitly black list each merge commit.

This should help us in avoiding that. More context and info on this slack thread - https://kubernetes.slack.com/archives/C1TU9EB9S/p1559061476023400

/hold
sending an email to k-dev with lazy consenus

/assign @cblecker 
fyi @sttts @dims

PR #17037, opened 2020-03-31
Respect do-not-merge/contains-merge-commits label

spiffxp:
This label is set by the mergecommitblocker plugin, currently enabled
for:

- kubernetes/kubernetes
- kubernetes/release
- kubernetes/sig-release

But tide wasn't setup to block merges on the presence of this label.

Have tide respect it in the query for all kubernetes orgs, and the
query for k/k

/sig testing
/sig release
/sig contributor-experience

/cc @nikhita @cblecker @justaugustus

Notice for this change was given back in May 2019, do we need another
lazy consensus period, or just a followup notice?
- ref: https://groups.google.com/forum/#!msg/kubernetes-dev/x88DyA-CfOI/puGgMGMpBgAJ
- ref: https://github.com/kubernetes/test-infra/pull/12783

### DIFF
--- a/config/prow/config.yaml
+++ b/config/prow/config.yaml
@@ -440,6 +440,7 @@ tide:
     missingLabels:
     - do-not-merge
     - do-not-merge/blocked-paths
+    - do-not-merge/contains-merge-commits
     - do-not-merge/hold
     - do-not-merge/invalid-commit-message
     - do-not-merge/invalid-owners-file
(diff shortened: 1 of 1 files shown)


spiffxp:
/hold
for comment

https://github.com/pulls?q=is%3Apr+is%3Amerged+label%3Ado-not-merge%2Fcontains-merge-commits+org%3Akubernetes+

https://github.com/kubernetes/kubernetes/pull/89364 made it through and may break publishing-bot?

The others are feature branch updates on k/k, all manually merged.

cblecker:
Also: I'd suggest this is a bug in the existing implementation, and we should fix it without an additional notice period

justaugustus:
> Also: I'd suggest this is a bug in the existing implementation, and we should fix it without an additional notice period

Agreed. Let's get this in.
/lgtm
/approve

FYI @kubernetes/release-engineering @kubernetes/release-team

PR #17040, opened 2020-03-31
Update cluster-api-provider-gcp jobs for v1alpha3

detiber:
Updates cluster-api-provider-gcp jobs for v1alpha3.


detiber:
Similar changes to the CAPA changes previously made, but for CAPG
/assign @ncdc

PR #17044, opened 2020-03-31
Update aks-engine for containerd

ksubrmnn:
Update config to use latest aks-engine build with azure cni + containerd changes


ksubrmnn:
@marosset

PR #17048, opened 2020-04-01
pull-k8sio-backup: use workload identity

listx:
The sister PR from the GCP SA side is
https://github.com/kubernetes/k8s.io/pull/710.

### DIFF
--- a/config/jobs/kubernetes/sig-release/cip/container-image-promoter.yaml
+++ b/config/jobs/kubernetes/sig-release/cip/container-image-promoter.yaml
@@ -26,6 +26,7 @@ presubmits:
     branches:
     - ^master$
     spec:
+      serviceAccountName: k8s-infra-gcr-promoter-test
       containers:
       - image: gcr.io/k8s-testimages/kubekins-e2e:v20200329-0d2a077-1.17
         command:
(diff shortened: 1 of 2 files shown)


listx:
/cc @fejta 

Also @jonjohnsonjr, FYI I'm going to try to run gcrane without GOOGLE_APPLICATION_CREDENTIALS set to a JSON key. Any reason to think grane won't work with Workload Identity?

spiffxp:
/approve
/lgtm
/hold
I leave it to @listx to `/hold cancel` when ready

listx:
> Also @jonjohnsonjr, FYI I'm going to try to run gcrane without GOOGLE_APPLICATION_CREDENTIALS set to a JSON key. Any reason to think grane won't work with Workload Identity?

I've done a quick test and it appears that gcrane can work without that env var as long as:
(1) gcloud is set as credHelper:

```
{
  "credHelpers": {
    "gcr.io": "gcloud", 
    "marketplace.gcr.io": "gcloud", 
    "eu.gcr.io": "gcloud", 
    "us.gcr.io": "gcloud", 
    "staging-k8s.gcr.io": "gcloud", 
    "asia.gcr.io": "gcloud"
  }
}
```

and (2) `docker-credential-gcloud` is [installed](https://cloud.google.com/container-registry/docs/advanced-authentication#gcloud-helper) (this seems to be an older name for the `docker-credential-gcr` command).

/hold cancel

PR #16852, opened 2020-03-19
Remove --hyperkube and KUBE_BUILD_HYPERKUBE references

dims:
follow up from change in k/k where we removed hyperkube, we should fix
up the jobs and scripts to remove any references.

Signed-off-by: Davanum Srinivas <davanum@gmail.com>

PR #17060, opened 2020-04-01
kubetest: remove kubeadmdind deploy

spiffxp:
It's not used anywhere in this repo, nor anywhere in github that I can
tell. The original author of this is no longer on github, and the last
meaningful commit to this was 1.5 years ago.

Discovered while reviewing https://github.com/kubernetes/test-infra/pull/16852


neolit123:
thanks, this removal SGTM as kubeadm-dind is archived.
https://github.com/kubernetes-retired/kubeadm-dind-cluster

/lgtm
/hold

Issue #17043, opened 2020-03-31
GCSweb is not automatically built


alvaroaleman:
**What happened**:

GCS web does not seem to be automatically built and pushed. There are also no manifests for it in this repo.

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

**Please provide links to example occurrences, if any**:

**Anything else we need to know?**:

/kind bug

PR #17072, opened 2020-04-02
Build gcsweb image

alvaroaleman:
Fixes https://github.com/kubernetes/test-infra/issues/17043

/assign @fejta @ixdy 

I have no idea what I am doing here but I validated that the result looks correct in a browser
/shrug


fejta:
> in the k8s-prow bucket and not testimages

k8s-prow seems more appropriate to me

alvaroaleman:
Should be good as two ppl had a look.
/hold cancel

PR #17078, opened 2020-04-03
Update pipeline to v.011.0 (v1beta1)

eddycharly:
This PR updates `tektoncd/pipeline` package and the `pipeline` controller to use the latest tekton pipeline version.

This PR covers code and tests.


alvaroaleman:
@eddycharly is this backwards-compatible, i.E. can an existing v1alpha1 pipeline resource be deserialized into a v1beta1 resource? Otherwise this will break all prow setups that have any such prowjob

eddycharly:
Good point @alvaroaleman.
I will need to check that. If it’s not the case how should we deal with this ?
/wip

alvaroaleman:
> I will need to check that. If it’s not the case how should we deal with this ?

Not sure. How many ppl are using this? Generally the options we have include:
* Tell ppl to manually change/delete prowjobs that ference the old version
* During startup of the pipeline controller, list all pipeline prowjobs, check if they are using the old type (how?) if yes, copy the data into the new type (assumes that all info that was on the old type is representable on the new type), put it onto the prowjob, update the prowjob

eddycharly:
At first sight, this will line will break at startup:
https://github.com/kubernetes/test-infra/blob/91337b319b0e79d8318f217b57582e2dd2baabdb/prow/cmd/pipeline/main.go#L98

Shall we assume that tekton has been updated to v0.11.0 in the target cluster or do we need to support both v1beta1 and v1alpha1 ?

eddycharly:
I think I mixed two things, updating to v0.11.0 and switching to v1beta1. 

I will revert the v1beta1 related changes so we keep using v1alpha1.

Switching to v1beta1 will need more thinking.
Maybe add a pipeline version in prow jobs (defaulting to v1alpha1 when not specified) ?
Maybe support both v1alpha1 and v1beta1 in the controller ?

This would allow for a smooth transition.

PR #17079, opened 2020-04-03
Add --gke-subnet-mode flag to GKE deployer in kubetest

tosi3k:


kubetest/gke.go
```diff
@@ -171,6 +173,15 @@ func newGKE(provider, project, zone, region, network, image, imageFamily, imageP
 		return nil, fmt.Errorf("--gke-shape must include a node pool named 'default', found %q", *gkeShape)
 	}
 
+	switch subnetMode := *gkeSubnetMode; subnetMode {
+	case "auto", "custom":
+		g.subnetMode = subnetMode
+	case "legacy":
```
mm4tt, line 179: NIT: Let's delete this, saying that we support either 'custom' or 'auto' is enough.

kubetest/gke.go
```diff
@@ -171,6 +173,15 @@ func newGKE(provider, project, zone, region, network, image, imageFamily, imageP
...
+		return nil, fmt.Errorf("--gke-subnet-mode=legacy is illegal since legacy subnet creation mode is deprecated in gcloud, please use 'auto' or 'custom'")
+	default:
+		return nil, fmt.Errorf("--gke-subnet-mode must be set either to 'auto' or 'custom'")
```
mm4tt, line 182: NIT: Let's add what was provided, e.g. `"..., got: %s", subnetMode)`

kubetest/gke.go
wojtek-t: +1

tosi3k: Fixed.

kubetest/gke.go
tosi3k: Fixed as well.

PR #17088, opened 2020-04-03
azure: add option to test secrets-store-csi-driver

aramase:
- Adds flag to test `secrets-store-csi-driver` using azure kubetest

/cc @chewong 
cc @ritazh

### DIFF
--- a/kubetest/aksengine.go
+++ b/kubetest/aksengine.go
@@ -52,38 +52,39 @@ const (
 )
 
 var (
-	aksResourceName        = flag.String("aksengine-resource-name", "", "Azure Resource Name")
-	aksResourceGroupName   = flag.String("aksengine-resourcegroup-name", "", "Azure Resource Group Name")
-	aksLocation            = flag.String("aksengine-location", "", "Azure AKS location")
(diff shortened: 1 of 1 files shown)


kubetest/aksengine.go
```diff
@@ -1303,6 +1304,8 @@ func (c *aksEngineDeployer) BuildTester(o *e2e.BuildTesterOptions) (e2e.Tester,
 		csiDriverName = "azurefile-csi-driver"
 	} else if *testBlobfuseCSIDriver {
 		csiDriverName = "blobfuse-csi-driver"
+	} else if *testSecretStoreCSIDriver {
+		csiDriverName = "secrets-store-csi-driver"
```
chewong, line 1308: Note that all CSI drivers call `make e2e-test` in `$(GOPATH)/src/sigs.k8s.io/${csiDriverName}` to kick off the e2e tests.
LGTM otherwise.

aramase: That's right. I'm going to be add a rule `e2e-test` in the makefile which will install the dependencies and kick off the tests.

PR #17086, opened 2020-04-03
Fix sig-scalability release-branch jobs

spiffxp:
The referenced files were removed from kubernetes/perf-tests, and
most sig-scalability jobs were updated to stop using them. And now
also these jobs

- https://github.com/kubernetes/perf-tests/pull/1151 removed the files
- https://github.com/kubernetes/test-infra/pull/17034 missed these jobs

Example affected PR's:
- https://github.com/kubernetes/kubernetes/pull/89789
- https://github.com/kubernetes/kubernetes/pull/89783
- https://github.com/kubernetes/kubernetes/pull/89693
- https://github.com/kubernetes/kubernetes/pull/89827

PR #17089, opened 2020-04-03
Fix scalability release branch jobs (take 2)

spiffxp:
This is take 2 of https://github.com/kubernetes/test-infra/pull/17086

In my haste to grep at the problem I missed the fact that k8s.io/perf-tests itself has release branches (ref: https://github.com/kubernetes/test-infra/pull/17086#issuecomment-608600996)

It turns out there is a presubmit in 1.15 using the `release-1.16` branch, and both presubmits in 1.18 are using the `master` branch.  I reverted the previous change and corrected the branches.

/assign @wojtek-t @mm4tt


wojtek-t:
Thanks a lot for the fix Aaron! I also filed https://github.com/kubernetes/test-infra/issues/17097 for double-checking this.

/lgtm
/approve

PR #17098, opened 2020-04-04
Skip session affinity timeouts with IPVS

aojea:
It seems that current IPVS jobs fail in the CI because of some limitation with the minimum timeout.

```
// The session affinity timeout must be bigger than 120, because IPVS module has a hardcoded TIME_WAIT timeout of 120s,
// and that value can't be sysctl'ed now.
// TODO(SataQiu): change to a small number to speed up testing when IPVS does really respect session affinity timeout.
// Ref: https://github.com/torvalds/linux/blob/master/net/netfilter/ipvs/ip_vs_proto_tcp.c
```

maybe we should skip those tests temporary?

### DIFF
--- a/config/jobs/kubernetes/sig-network/sig-network-misc.yaml
+++ b/config/jobs/kubernetes/sig-network/sig-network-misc.yaml
@@ -34,7 +34,8 @@ presubmits:
         - --ginkgo-parallel=30
         - --provider=gce
         - --stage=gs://kubernetes-release-pull/ci/pull-kubernetes-e2e-gci-gce-ipvs
-        - --test_args=--ginkgo.skip=\[Slow\]|\[Serial\]|\[Disruptive\]|\[Flaky\]|\[Feature:.+\] --minStartupPods=8
+        # Skip "Services.should.have.session.affinity.timeout.work" #89358
+        - --test_args=--ginkgo.skip=\[Slow\]|\[Serial\]|\[Disruptive\]|\[Flaky\]|\[Feature:.+\]|Services.should.have.session.affinity.timeout.work --minStartupPods=8
         - --timeout=80m # thinking about making this longer? don't! 80m is a hard cap, and should get down to no more than 60m.
         image: gcr.io/k8s-testimages/kubekins-e2e:v20200402-36cead4-master
         resources:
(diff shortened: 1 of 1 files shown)


SataQiu:
This is another way to get the CI through. But the skip way can not check session affinity timeout feature at all. Session affinity timeout feature is OK, just with some limit. If we set the timeout to a reasonable value, CI can verify that the function is working. The downside is that it takes longer to complete the test. I'm not sure which is better. Defer to @andrewsykim

aojea:
> Session affinity timeout feature is OK, just with some limit

I think that you are really spot on about the debate here, my reasoning to categorize this limitation as a bug is because kube-proxy with iptables works fine with any timeout, hence ipvs should PASS the same tests that iptables mode passes not the other way around.

> But the skip way can not check session affinity timeout feature at all

Right, but the question here is who should be responsible to update the test once the limitation is gone?
I think that's on IPVS maintainers/developers/owners to add a TODO to the code and enable the tests once the limitation is removed, not to the test maintainers to remove the timeout on the IPVS removes the limitation.

Anyway, this is my personal interpretation, I'll be fine with any final decision 😄

andrewsykim:
> I think that you are really spot on about the debate here, my reasoning to categorize this limitation as a bug is because kube-proxy with iptables works fine with any timeout, hence ipvs should PASS the same tests that iptables mode passes not the other way around.

I agree, this is a bug which is unfortunately hard to fix at the moment because of kernel limitations. I don't think we should skip the tests _yet_, because the presubmit is optional anyways. I put this as a topic of discussion for the next SIG meeting so we can discuss possible next steps.

SataQiu:
> Right, but the question here is who should be responsible to update the test once the limitation is gone?
I think that's on IPVS maintainers/developers/owners to add a TODO to the code and enable the tests once the limitation is removed, not to the test maintainers to remove the timeout on the IPVS removes the limitation.

Yeah, totally agree with you @aojea . I'm also not sure who should be responsible to update the test once the limitation is gone. If you agree, maybe I can fill in your name 😄 . He should be IPVS maintainers/developers/owners.

aojea:
/hold
until we have better consensus, keeping the test failing is a good option to me that I didn't consider :)

aojea:
/close
sig-network decision explained in comment 
https://github.com/kubernetes/kubernetes/issues/89358#issuecomment-615438734

Thanks for the nice discussion 😄

PR #17103, opened 2020-04-05
Kops - Switch Amazon Linux 2 e2e test to Docker runtime

hakman:
Installing Docker from .tgz should permit running it on Amazon Linux 2 without issues.


rifelpet:
nice :)

/lgtm
/approve

PR #17104, opened 2020-04-06
Use KOPS_RUN_TOO_NEW_VERSION only where needed

hakman:
The `kops-aws-k8s-latest` test is still failing. Seems the previous commit missed the place where it was actually needed. The rest will work ok without this env var for now.
https://testgrid.k8s.io/sig-cluster-lifecycle-kops#kops-aws-k8s-latest


johngmyers:
I suspect that after we create a release-1.18 branch we would switch several of these jobs to latest.

hakman:
@johngmyers I would prefer not to switch to 1.19. It will be buggy and flaky and doesn't offer much value. 1.18 will be released in 3 months and then 1.19 in another 3 months. I thing we will have a very long time to address any issues would appear in 1.19.

We shouldn't keep the env vars unless we use them, same way we do for obsolete versions. If we want to put them everywhere, they should be added to kubetest, right?

johngmyers:
The question would be how we want to handle the k/k 1.19 release branching. Kops won't be able to keep master on stable once it creates the 1.18 release branch.

hakman:
@johngmyers let's chat in next office hours. For now this is not really that important :).
/close

PR #17111, opened 2020-04-06
Spyglass: Add remote lenses support

alvaroaleman:
This PR aims to be a minimal implementation of the remote lense proposal in https://docs.google.com/document/d/1YBqwFvlerplpb9kMm7rO0eRLYheemmkbUL6GUmKlopg/edit?ts=5d115af0#heading=h.x9snb54sjlu9 (which some changes where needed) that does not require any configuration changes.

Whats included:
* All built-in lenses use the external lens machinery
* To not require config changes, spyglass creates an additional listener for the lenses on Deck

What differs from the proposal:
* The response is not send as `multipart/form-data`, instead Spyglass just proxies through the response from the lens server. This allows us to have a very stupid proxy endpoint rather than requiring a very smart one that inspects the response as a stream and injects the spyglass template
* There is no new `remote_lens` config option, as that would require ppl to configure the builtin-lenses as `remote_lenses` which is confusing. Instead, the remote settings are a part of the lens config

TODOs/Followups:
* Refactor the interaction in a way that the lensserver never needs to find out the gcs key for a prowjob, as that requires it to have a prowjob client
* Serve static resources from remote lens (this pr is only the dynamic resources)
* Docs (I want the code to be finalized first)

/assign @Katharine @stevekuznetsov


alvaroaleman:
A running version of this is at https://deck-canary-ci.svc.ci.openshift.org/, you can e.G. check spyglass on https://deck-canary-ci.svc.ci.openshift.org/view/gcs/origin-ci-test/logs/periodic-ci-image-import-to-build01/20996

prow/cmd/deck/main.go
```diff
@@ -277,7 +279,7 @@ func main() {
 
 	// setup config agent, pod log clients etc.
 	configAgent := &config.Agent{}
-	if err := configAgent.Start(o.configPath, o.jobConfigPath); err != nil {
+	if err := configAgent.Start(o.configPath, o.jobConfigPath, spglassConfigDefaulting); err != nil {
```
alvaroaleman, line 282: The reason for adding passing this in rather than defining it in the `config` package is that:
* This needs the `lens` package because it accesses its registr
* The lens package needs the `config` package, so importing the former into the later causes an import cycle

Even if the import cycle issue was solved, defining this in the `config` package would mean that we have to import all lenses everywhere where we use the `config` package (so more or less everywhere) and I think we should try to avoid that.

stevekuznetsov: Why should we avoid it? We do it for all other config for every part of the system. If we want to break from that let's do it in a focused effort?

prow/cmd/deck/main.go
```diff
@@ -1131,68 +1161,75 @@ func handleArtifactView(o options, sg *spyglass.Spyglass, cfg config.Getter) htt
...
+		ArtifactSource: request.Source,
+		LensIndex:      request.Index,
+	}
+	serializedRequest, err := json.Marshal(lensRequest)
+	if err != nil {
+		http.Error(w, fmt.Sprintf("failed to marshal request to lens backend: %v", err), http.StatusInternalServerError)
+		return
 	}
+
+	(&httputil.ReverseProxy{
```
stevekuznetsov, line 1226: This is weird enough to merit some eyeballs -- so this shares a lifetime with the goroutine, right? Does it leak memory?

prow/config/config.go
```diff
@@ -585,6 +585,24 @@ type LensFileConfig struct {
 	OptionalFiles []string `json:"optional_files,omitempty"`
 	// Lens is the lens to use, alongside any lens-specific configuration.
 	Lens LensConfig `json:"lens"`
+	// RemoteConfig specifies how to access remote lenses
+	RemoteConfig *LensRemoteConfig `json:"remote_config,omitempty"`
+}
+
+// LensRemoteConfig is the configuration for a remote lens.
```
stevekuznetsov, line 592: Are any of these optional (priority, title), and, if so, could we use `omitempty` ?

alvaroaleman: No they are not optional

prow/cmd/deck/main.go
alvaroaleman: Yes it shares a lifetime with the goroutine, I don't think this leaks memory

prow/cmd/deck/main.go
alvaroaleman: But in this particular case it implies that we do not only have to import some api types, we have to import everything because we need a populated registry. This is like adding a `_ "k8s.io/test-infra/prow/tide` import to plank.

alvaroaleman: (Except that its not just for plank but for everything that uses the `config` package)

prow/cmd/deck/main.go
alvaroaleman: (I've burned my fingers already in the past by assuming this can be implemented by doing a simple io.Copy on the response, but turns out, HTTP is more complicated than that so I've refrained from trying to be smart)

stevekuznetsov:
It's been a couple weeks, I think we're reaching lazy consensus

/lgtm
/approve
/hold

PR #17113, opened 2020-04-06
Add metrics-server presubmits

serathius:
/cc @s-urbaniak 
Ref https://github.com/kubernetes-sigs/metrics-server/issues/483


serathius:
Adding dashboards needs to be done in same pr as jobs :(

PR #17119, opened 2020-04-06
prow/plugins/config: Add DependentBugTargetReleases

wking:
[This comment][1] shows the Bugzilla plugin complaining on a 4.2.z backport where the parent 4.3 bug targets 4.3.0 instead of the expected 4.3.z.  By adding a new slice config, admins can explicitly say that 4.2 bugs should depend on 4.3 bugs which target either 4.3.0 or 4.3.z.

For backwards compatability with existing configurations, the singular `dependent_bug_target_release` is appended to the plural slice if the string is not already included.

The:

```go
_ = append(...)
```

avoids both:

```
prow/plugins/config.go:1545:40: cannot take the address of append(*output.DependentBugTargetReleases, *child.DeprecatedDependentBugTargetRelease)
```

from:

```go
output.DependentBugTargetReleases = &append(...)
```

and:

```
prow/plugins/config.go:1545:10: append(*output.DependentBugTargetReleases, *child.DeprecatedDependentBugTargetRelease) evaluated but not used
```

from:

```go
append(...)
```

[1]: https://github.com/openshift/kubecsr/pull/17#issuecomment-609981202


prow/plugins/config.go
```diff
@@ -1464,8 +1473,27 @@ func ResolveBugzillaOptions(parent, child BugzillaBranchOptions) BugzillaBranchO
 			output.DependentBugStatuses = parent.DependentBugStatuses
 			output.DependentBugStates = mergeStatusesIntoStates(output.DependentBugStates, parent.DependentBugStatuses)
 		}
-		if parent.DependentBugTargetRelease != nil {
-			output.DependentBugTargetRelease = parent.DependentBugTargetRelease
+		if parent.DependentBugTargetReleases != nil {
+			output.DependentBugTargetReleases = parent.DependentBugTargetReleases
+		}
+		if parent.DeprecatedDependentBugTargetRelease != nil {
```
stevekuznetsov, line 1479: This does not seem DRY

prow/plugins/config.go
```diff
@@ -1518,8 +1546,27 @@ func ResolveBugzillaOptions(parent, child BugzillaBranchOptions) BugzillaBranchO
...
+					contains = true
+					break
+				}
+			}
+		}
+		if !contains {
+			if output.DependentBugTargetReleases == nil {
+				output.DependentBugTargetReleases = &[]string{*child.DeprecatedDependentBugTargetRelease}
+			} else {
+				_ = append(*output.DependentBugTargetReleases, *child.DeprecatedDependentBugTargetRelease)
```
stevekuznetsov, line 1567: Unless I'm crazy, this dereferences `output.DependentBugTargetRelease` and allocates for a new slice (by copy) that appends the child's release, so it does not change the output. Do we test this?

prow/plugins/config.go
wking: > This does not seem DRY

Well, I'd really like `ResolveBugzillaOptions` to be private for the config repo.  Then I'd rework the callers to ensure that we were always chaining these down to a dummy root.  And then we'd only need to perform unification logic like this in the child (and parents would be unified when recursion had them as the child).  But that seemed like more than I should bite off in this PR ;)

prow/plugins/bugzilla/bugzilla.go
```diff
@@ -92,15 +92,15 @@ func helpProvider(config *plugins.Configuration, enabledRepos []config.OrgRepo)
...
 				conditions = append(conditions, "depend on at least one other bug")
 			}
 			if opts[branch].DependentBugStates != nil {
 				pretty := strings.Join(prettyStates(*opts[branch].DependentBugStates), ", ")
 				conditions = append(conditions, fmt.Sprintf("have all dependent bugs in one of the following states: %s", pretty))
 			}
-			if opts[branch].DependentBugTargetRelease != nil {
-				conditions = append(conditions, fmt.Sprintf("have all dependent bugs target the %q release", *opts[branch].DependentBugTargetRelease))
+			if opts[branch].DependentBugTargetReleases != nil {
+				conditions = append(conditions, fmt.Sprintf("have all dependent bugs target a release in %s", strings.Join(*opts[branch].DependentBugTargetReleases, ", ")))
```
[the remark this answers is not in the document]
wking: > to be parallel with how we display the other conditions...

Done with 919e07511c -> 2f5d2fb0cb.

prow/plugins/bugzilla/bugzilla.go
```diff
@@ -627,36 +627,45 @@ func validateBug(bug bugzilla.Bug, dependents []bugzilla.Bug, options plugins.Bu
...
-				// the BugZilla web UI shows one option for target release, but returns the
-				// field as a list in the REST API. We only care for the first item and it's
-				// not even clear if the list can have more than one item in the response
-				valid = false
-				errors = append(errors, fmt.Sprintf("expected dependent "+bugLink+" to target the %q release, but it targets %q instead", bug.ID, endpoint, bug.ID, *options.DependentBugTargetRelease, bug.TargetRelease[0]))
+				errors = append(errors, fmt.Sprintf("expected dependent "+bugLink+" to target a release in %s, but no target release was set", bug.ID, endpoint, bug.ID, strings.Join(*options.DependentBugTargetReleases, ", ")))
 			} else {
-				validations = append(validations, fmt.Sprintf("dependent "+bugLink+" targets the %q release, matching the expected (%s) release", bug.ID, endpoint, bug.ID, bug.TargetRelease[0], *options.DependentBugTargetRelease))
+				found := false
+				for _, release := range *options.DependentBugTargetReleases {
```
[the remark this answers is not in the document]
wking: Ported to `sets` with 2f5d2fb0cb -> 5502a372bd.

prow/plugins/bugzilla/bugzilla.go
```diff
@@ -627,36 +627,45 @@ func validateBug(bug bugzilla.Bug, dependents []bugzilla.Bug, options plugins.Bu
...
+					// the BugZilla web UI shows one option for target release, but returns the
+					// field as a list in the REST API. We only care for the first item and it's
+					// not even clear if the list can have more than one item in the response
+					if release == bug.TargetRelease[0] {
+						found = true
+						break
+					}
+				}
+				if found {
+					validations = append(validations, fmt.Sprintf("dependent "+bugLink+" targets the %q release (%s is allowed)", bug.ID, endpoint, bug.ID, bug.TargetRelease[0], strings.Join(*options.DependentBugTargetReleases, ", ")))
```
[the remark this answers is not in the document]
wking: Could be singular.  Went with `..., which is one of the valid target releases: ...` in 92e9d951dd.

prow/plugins/config.go
wking: Yeah, looks like this was cutting too many corners.  Stopped assuming `append` mutated in place with 92e9d951dd.

wking:
Pushed 92e9d951dd -> d13cc8bce8 to fix [the Bazel issue][1].  Seems like it's stuck in `Expected`, but we'll see if the new round runs at some point.

[1]: https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/test-infra/17119/pull-test-infra-bazel/1247313221530947589

wking:
Pushed a commit-timestamp bump, in case that unsticks it.  But I'd guess the main issue is the `needs-ok-to-test` label.

wking:
All green now, just needs a fresh `/lgtm`.

PR #17121, opened 2020-04-06
deck: use prow/github instead of go-github in oauth handlers

stevekuznetsov:
There was no real reason to be using the older, separate client in
deck's handlers and it meant that new tooling to track client usage in
logs and metrics was not supported for these calls.

### DIFF
--- a/prow/flagutil/github.go
+++ b/prow/flagutil/github.go
@@ -24,18 +24,14 @@ import (
 	"strings"
 
 	"github.com/sirupsen/logrus"
-	"golang.org/x/oauth2"
-
-	"k8s.io/test-infra/pkg/ghclient"
 	"k8s.io/test-infra/prow/config/secret"
 	"k8s.io/test-infra/prow/git"
 	"k8s.io/test-infra/prow/github"
-	"k8s.io/test-infra/prow/githuboauth"
 )
 
 // GitHubOptions holds options for interacting with GitHub.
 type GitHubOptions struct {
-	host                string
+	Host                string
 	endpoint            Strings
 	graphqlEndpoint     string
 	TokenPath           string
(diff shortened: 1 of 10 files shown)


prow/flagutil/github.go
```diff
@@ -175,17 +171,3 @@ func (o *GitHubOptions) GitClient(secretAgent *secret.Agent, dryRun bool) (clien
...
-	oauthConfig.Endpoint = oauth2.Endpoint{
-		AuthURL:  fmt.Sprintf("https://%s/login/oauth/authorize", o.host),
-		TokenURL: fmt.Sprintf("https://%s/login/oauth/access_token", o.host),
-	}
-	return githuboauth.NewClient(oauthConfig)
-}
-
-// GetGitHubClient returns a github client wrapper.
-func (o *GitHubOptions) GetGitHubClient(accessToken string, dryRun bool) githuboauth.GitHubClientWrapper {
-	return ghclient.NewClientWithEndpoint(o.endpoint.Strings()[0], accessToken, dryRun)
```
stevekuznetsov, line 190: You might ask -- is it even valid to use `prow/github` if we're routing through `ghproxy` for these calls to get the ID of the authenticated user? But we did in the past, as well!

alvaroaleman:
So the purpose of this change is to allow us tracking api requests done with a user-provided token in the re-run handlers by routing them through ghproxy?

stevekuznetsov:
The purpose of this change is two-fold:

1. stop using multiple libraries to do the same thing
2. as a side-effect, track token usage better

We were *already* pushing this through `ghproxy`, I believe, as noted in my comment in the code above -- we also see `go-github` user agents in `ghproxy` today.

stevekuznetsov:
@cjwagner as noted my change here does not change how we use `ghproxy` at all -- I can follow up to use the last item in the endpoints list

Issue #12216, opened 2019-04-16
tide: failed batches should re-trigger only failed presubmits


stevekuznetsov:
When `tide` determines than an in-progress batch has failed due to one or more presubmits in the batch failing, it re-triggers all of them. When a batch contains a perma-failed test, this causes `tide` to re-trigger batches on every sync. When other jobs in the batch require third-party resources like cloud computing quota, the issue quickly balloons. `tide` should re-trigger only the jobs it needs.

/cc @cjwagner @fejta 
/area prow/tide

stevekuznetsov:
This is fixed by proxy by correctly aborting older batches but is still an issue when aborting them and the associated cleanup/grace period takes a while.

alvaroaleman:
I've also noticed that tide seems to trigger retest for _all_ presubmits if one or more of them needs retesting, even for the ones that have up-to-date results. Is that related or a different issue? I'll probably look into this, this is really, really annoying.

cjwagner:
> I've also noticed that tide seems to trigger retest for _all_ presubmits if one or more of them needs retesting, even for the ones that have up-to-date results. Is that related or a different issue? I'll probably look into this, this is really, really annoying.

Are you talking about when PRs have tests run against different base branch commits? We could optimize for that edge case more by trying to make Tide's triggering aware of existing jobs. This would complicate things, but is certainly doable.
Can you link to some occurrences? Roughly how frequently does this happen?

alvaroaleman:
I don't have a public example because most of our code is not public, but the example I have basically boils down to:

* Some did a PR
* One test failed
* A couple of days later it succeeded after a `/retest`
* All except this one test need retesting
* Tide restarts _all_ tests
* We now have to successfull Prowjobs with the same base and head sha that were both successful:
```
$ k get prowjob -l prow.k8s.io/job=pull-dashboard-e2e-test,prow.k8s.io/refs.pull=1415,prow.k8s.io/job=pull-dashboard-e2e-test|grep 'success'|awk '{ print $1 }'|xargs kubectl get prowjob -o json|jq '.items[]|.spec.refs'
{
  "base_ref": "master",
  "base_sha": "6d38e4cbb5ccfed8e18cb2535378627f76edea95",
  "clone_uri": "git@github.com:kubermatic/dashboard-v2.git",
  "org": "kubermatic",
  "pulls": [
    {
      "author": "kdomanski",
      "number": 1415,
[...]

PR #17122, opened 2020-04-07
Tide: Don't create retests for unfinished or successful batch jobs

alvaroaleman:
Fixes https://github.com/kubernetes/test-infra/issues/12216

Makes Tide not create a new Batch job if a batch job of the same name for the same(org, repo, baseBranch, baseSHA, pullNumbers+PullHeadSHAs) already exists that:
* Is not finished
* Finished successfully

Diff is a lot smaller when disabling whitespace changes.

/assign @cjwagner @stevekuznetsov

### DIFF
--- a/prow/tide/tide_test.go
+++ b/prow/tide/tide_test.go
@@ -1129,13 +1129,14 @@ func testTakeAction(clients localgit.Clients, t *testing.T) {
 	testcases := []struct {
 		name string
 
-		batchPending bool
-		successes    []int
-		pendings     []int
-		nones        []int
-		batchMerges  []int
-		presubmits   map[int][]config.Presubmit
-		mergeErrs    map[int]error
+		batchPending    bool
+		successes       []int
+		pendings        []int
+		nones           []int
+		batchMerges     []int
+		presubmits      map[int][]config.Presubmit
+		preExistingJobs []runtime.Object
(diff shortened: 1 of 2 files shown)


prow/tide/tide_test.go
```diff
@@ -1421,153 +1541,182 @@ func testTakeAction(clients localgit.Clients, t *testing.T) {
 		},
 	}
 
+	//logrus.SetLevel(logrus.DebugLevel)
```
matthyx, line 1544: nit here?

alvaroaleman:
> Could get @cjwagner eyes on this. Only comment is -- if we care about DRY or potentially future-proofing this if we change Refs in the future -- could we use Refs.String() as the key?

Not sure. `Refs.String()` does not e.G. contain the org and repo. Ppl changing it might not be aware of transient requirements incurred on it by using it that way. We know very precisely what we want here (reponame, basehash, headhashes) and thats IMHO very unlikely to change.

alvaroaleman:
@stevekuznetsov are you okay with merging this by the end of the week if @cjwagner  doesn't find time to look into it till then? The worst case scenario for this pull is that we can not merge batches anymore

alvaroaleman:
I'll roll this out and watch batches.
/hold cancel

PR #17123, opened 2020-04-07
Allow image push for k8s dns images

prameshj:
This includes kube-dns and node-cache images.
The cloudbuild file is being added in https://github.com/kubernetes/dns/pull/364

@MrHohn @rramkumar1 @listx

### DIFF
--- a/config/jobs/image-pushing/k8s-staging-dns.yaml
+++ b/config/jobs/image-pushing/k8s-staging-dns.yaml
@@ -1,0 +1,32 @@
+postsubmits:
+  # this is the github repo we'll build from; this block needs to be repeated for each repo.
+  kubernetes/dns:
+    - name: dns-push-images
+      cluster: test-infra-trusted
+      annotations:
+        # this is the name of some testgrid dashboard to report to.
+        testgrid-dashboards: sig-network-dns
+      decorate: true
+      branches:
+        - ^master$
+      spec:
+        containers:
+          - image: gcr.io/k8s-testimages/image-builder:v20200213-0032cdb
+            command:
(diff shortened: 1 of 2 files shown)


config/jobs/image-pushing/k8s-staging-dns.yaml
```diff
@@ -0,0 +1,32 @@
...
+    - name: dns-push-images
+      cluster: test-infra-trusted
+      annotations:
+        # this is the name of some testgrid dashboard to report to.
+        testgrid-dashboards: sig-network-dns
+      decorate: true
+      branches:
+        - ^master$
+      spec:
+        containers:
```
listx, line 13: I think you can just use a `serviceAccountName` like in https://github.com/kubernetes/test-infra/blob/master/config/jobs/image-pushing/k8s-staging-artifact-promoter.yaml#L13. Not sure if the TODO comment there is applicable @fejta

prameshj: Done

fejta: Yes, the image pusher should use the pusher, but I believe everything is erroneously granted to the deployer right now (which is supposed to be used to change k8s resources)

So keep the TODO

prameshj: Thanks Erick. I have made the change.

listx:
/lgtm

@prameshj I cannot approve as I am not a maintainer of test-infra, maybe ping the oncall https://storage.googleapis.com/test-infra-oncall/oncall.html ?

PR #17125, opened 2020-04-07
Skip SCTP test on alpha features job

aojea:
The tests tagged as `[Feature:SCTPConnectivity]"` were not supposed to run on Kubernetes CI, however, the regex in the job `ci-kubernetes-e2e-gci-gce-alpha-features` wasn't excluding it and started to run them, with the problem that they started to fail :)

https://github.com/kubernetes/kubernetes/issues/89782#issuecomment-610369830

> These will be tagged "[Feature:SCTPConnectivity]". They can be run
> by network plugin developers but will not run as part of any of the
> standard Kubernetes test runs.

We should filter these tests to not run them, it also refactors the regex.

Fixes: https://github.com/kubernetes/kubernetes/issues/89782

### DIFF
--- a/config/jobs/kubernetes/sig-cloud-provider/gcp/gcp-gce.yaml
+++ b/config/jobs/kubernetes/sig-cloud-provider/gcp/gcp-gce.yaml
@@ -289,7 +289,7 @@ presubmits:
         - --provider=gce
         - --runtime-config=api/all=true
         - --stage=gs://kubernetes-release-pull/ci/pull-kubernetes-e2e-gce-alpha-features
-        - --test_args=--ginkgo.focus=\[Feature:(ServiceAccountIssuerDiscovery|StorageVersionHash|Audit|PodPreset|ExpandCSIVolumes|ExpandInUseVolumes|RunAsGroup|TTLAfterFinished|NodeLease|VolumeSnapshotDataSource|CSIInlineVolume)\]|Networking --ginkgo.skip=Networking-Performance|IPv6|Feature:Volumes|csi-hostpath-v0 --minStartupPods=8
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes/sig-cloud-provider/gcp/gcp-gce.yaml
```diff
@@ -460,7 +460,7 @@ periodics:
       - --gcp-zone=us-west1-b
       - --provider=gce
       - --runtime-config=api/all=true
-      - --test_args=--ginkgo.focus=\[Feature:(StorageVersionHash|Audit|PodPreset|ExpandCSIVolumes|ExpandInUseVolumes|RunAsGroup|TTLAfterFinished|NodeLease|VolumeSnapshotDataSource|CSIInlineVolume)\]|Networking --ginkgo.skip=Networking-Performance|IPv6|Feature:Volumes|csi-hostpath-v0 --minStartupPods=8
+      - --test_args=--ginkgo.focus=\[Feature:(StorageVersionHash|Audit|PodPreset|ExpandCSIVolumes|ExpandInUseVolumes|RunAsGroup|TTLAfterFinished|NodeLease|VolumeSnapshotDataSource|CSIInlineVolume)\]|Networking --ginkgo.skip=\[Feature:(SCTPConnectivity|Volumes|Networking-Performance|Networking-IPv6)\]|csi-hostpath-v0 --minStartupPods=8
```
danwinship, line 463: previously it would skip both `[Feature:Networking-IPv6]` and `[Feature:IPv6DualStackAlphaFeature]` but with this change it would only skip the former

aojea: great catch
fixed

justaugustus:
@aojea -- lgtm, but needs a rebase :)

PR #17130, opened 2020-04-07
prow/spyglass: preserve formatting in failures output

dmage:


alvaroaleman:
@dmage Do you have a link for an example job where this makes a difference?

dmage:
@alvaroaleman you can pick any failed test, for example this one: https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/89937/pull-kubernetes-integration/1247643271635668994

Click on TestServiceAlloc. The test output looks like a single line, at least on Chrome/80.0.3987.163.

PR #17134, opened 2020-04-07
jobs: skip_report for pull-kubernetes-e2e-node-containerd

spiffxp:
It's been failing continuously for days. It's not merge-blocking,
but contributors aren't paying attention to that fact and spamming
`/retest`. Let's skip reporting until the job is fixed.


spiffxp:
/hold
ref: https://github.com/kubernetes/kubernetes/issues/89847#issuecomment-610493661

cheftako:
Working on the fix now. I agree that we should silence this until the fix is ready.

PR #17136, opened 2020-04-07
jobs: add sig/ labels to OWNERS for kubernetes/ jobs

spiffxp:
I pushed a change to a sig-node job and was surprised it didn't get auto-labeled as such


justaugustus:
This is great. Thanks @spiffxp!

PR #17138, opened 2020-04-07
Make status-reconciler use ghproxy.

cjwagner:
Resolves this warning:
```yaml
jsonPayload: {
  component: "status-reconciler"   
  file: "prow/flagutil/github.go:90"   
  func: "k8s.io/test-infra/prow/flagutil.(*GitHubOptions).Validate"   
  level: "error"   
  msg: "It doesn't look like you are using ghproxy to cache API calls to GitHub! This has become a required component of Prow and other components will soon be allowed to add features that may rapidly consume API ratelimit without caching. Starting May 1, 2020 use Prow components without ghproxy at your own risk! https://github.com/kubernetes/test-infra/tree/master/ghproxy#ghproxy"   
}
```
/assign @BenTheElder


BenTheElder:
/lgtm
/approve
Thanks!

PR #17142, opened 2020-04-07
Reply to pr for failed postsubmit jobs

### LINKED ISSUE #10402: prow: reply to PR with pass/fail status of merged postsubmit
Post-submit jobs that are tied to a PR should be able to post the results of the job back to the PR.

We do something similar for the `updateconfig` plugin, but job status (pass/fail) and a link to the job log should be something all postsubmits can do.

/kind feature

adshmh:
Part of work on #10402


prow/plugins/trigger/push.go
```diff
@@ -55,11 +57,33 @@ func createRefs(pe github.PushEvent) prowapi.Refs {
...
+
 func handlePE(c Client, pe github.PushEvent) error {
 	if pe.Deleted || pe.After == "0000000000000000000000000000000000000000" {
 		// we should not trigger jobs for a branch deletion
 		return nil
 	}
+	// TODO: it seems there is a bug in GH API, which returns an empty list of
+	// PRs for all commits listed by the push event. Waiting for 1 second seems
+	// to successfully work around this.
+	time.Sleep(1 * time.Second)
```
petr-muller, line 86: Huh, this is really surprising. This looks like a possible source of surprising flakes (how do we know a second is always enough?) Can we get at least a confirmation that this is really a bug and this is a preferred way to work around it?

prow/cmd/mkpj/main.go
```diff
@@ -85,12 +75,7 @@ func (o *options) genJobSpec(conf *config.Config, name string) (config.JobBase,
 		}
 		for _, p := range ps {
 			if p.Name == o.jobName {
-				return p.JobBase, pjutil.PostsubmitSpec(p, prowapi.Refs{
-					Org:     org,
-					Repo:    repo,
-					BaseRef: o.baseRef,
-					BaseSHA: o.baseSha,
-				})
+				return p.JobBase, pjutil.PostsubmitSpec(p, o.createRefs(org, repo))
```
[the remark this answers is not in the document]
adshmh: Thank you for the review. Fixed.

prow/github/fakegithub/fakegithub.go
```diff
@@ -103,6 +103,9 @@ type FakeClient struct {
 	Column             string
 	OrgRepoIssueLabels map[string][]github.Label
 	OrgProjects        map[string][]github.Project
+
+	// PullRequests associated with a commit
+	CommitPullRequests map[string][]github.PullRequest
```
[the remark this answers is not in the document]
adshmh: Thank you for the review. Fixed.

prow/plugins/trigger/push.go
adshmh: Thanks for the review. I reported this to GitHub support and asked for confirmation and a work-around about the same time I submitted the PR (about 2 weeks ago). They have not confirmed yet, but based on their responses it does seem to be a bug.

adshmh: I have logged a bug with GitHub support team. Based on their feedback, they confirm this can happen but have not promised a fix. They agree the workaround I have used could work, but cannot guarantee that. The GH support ticket is titled "bug in endpoint for listing commit PRs", if further follow up on it is needed.

petr-muller (approved):
/lgtm

prow/plugins/trigger/push.go
```diff
@@ -59,11 +59,33 @@ func createRefs(pe github.PushEvent) prowapi.Refs {
...
 }
 
+func getPullRequests(c githubClient, pe github.PushEvent) ([]github.PullRequest, error) {
+	org := pe.Repo.Owner.Login
+	repo := pe.Repo.Name
+	for _, commit := range pe.Commits {
+		if commit.ID == "" {
+			continue
+		}
+		pulls, err := c.ListCommitPullRequests(org, repo, commit.ID)
```
spiffxp, line 69: I think we need to be aggregating pulls for all commits vs. returning pulls for the first commit that has any

prow/plugins/trigger/push.go
```diff
@@ -73,20 +95,29 @@ func handlePE(c Client, pe github.PushEvent) error {
...
 		} else if !shouldRun {
 			continue
 		}
-		refs := createRefs(pe)
-		labels := make(map[string]string)
-		for k, v := range j.Labels {
-			labels[k] = v
+		var pj prowapi.ProwJob
+		if len(prs) > 0 {
+			pj = pjutil.NewPostsubmit(prs[0], prs[0].Base.SHA, j, pe.GUID)
```
spiffxp, line 111: I wouldn't trust GitHub's api to provide consistent ordering of PR's associated with a commit

I would at least recommend filtering down to whichever PR has most recently been merged

prow/plugins/trigger/push_test.go
```diff
@@ -151,61 +153,131 @@ func TestHandlePE(t *testing.T) {
...
-			Config:        &config.Config{ProwConfig: config.ProwConfig{ProwJobNamespace: "prowjobs"}},
-			Logger:        logrus.WithField("plugin", PluginName),
-		}
-		postsubmits := map[string][]config.Postsubmit{
-			"org/repo": {
-				{
-					JobBase: config.JobBase{
-						Name: "pass-butter",
+		{
+			name: "postsubmit job has PR reference",
```
spiffxp, line 157: This needs test cases for push events that have multiple commits and multiple PRs in various states (merged and not merged)

spiffxp (changes requested):
Thank you for working on this!

One specific concern: this needs more thought put into how to handle multiple commits and pull requests per push event

A more general concern: I don't think this fully fixes #10402, I would remove that from the description.  The reason is that there's no way to report _to the PR_ on success.  Crier's github reporter will post status contexts to a merge commit for postsubmits, and with this change it will create a comment on the PR if there are any failures.

But there is no way to comment to the PR on success.  This will require some changes to crier's reporting logic.  I want this so that:
- I can see if a PR caused something to successfully deploy
- I can see if a PR caused an image to be built and pushed

spiffxp:
need rebase

prow/plugins/trigger/push.go
adshmh: Thank you for the review. Fixed.
My tests seem to show the GH API bug is still there, and the 1 second sleep helps work around it.

prow/plugins/trigger/push.go
adshmh: Thank you for the review. Fixed.

prow/plugins/trigger/push_test.go
adshmh: Thank you for the review. Fixed.

adshmh:
> A more general concern: I don't think this fully fixes #10402, I would remove that from the description. The reason is that there's no way to report _to the PR_ on success. Crier's github reporter will post status contexts to a merge commit for postsubmits, and with this change it will create a comment on the PR if there are any failures.
> 
> But there is no way to comment to the PR on success. This will require some changes to crier's reporting logic. I want this so that:
> 
>     * I can see if a PR caused something to successfully deploy
> 
>     * I can see if a PR caused an image to be built and pushed

Thank you for the review. I have edited the PR title accordingly. 

Regarding comments on the PR on success, would it be acceptable to keep the scope of this PR to failed postsubmits? 
Once this PR is merged, I will follow up with another one, to make changes to crier's logic  necessary to support successful postsubmits.

spiffxp:
>Regarding comments on the PR on success, would it be acceptable to keep the scope of this PR to failed postsubmits?
Once this PR is merged, I will follow up with another one, to make changes to crier's logic necessary to support successful postsubmits.

That sounds good to me.  I'll review during working hours tomorrow/

/uncc @Katharine @michelle192837 
/cc @cjwagner @alvaroaleman @fejta 
in case you have any input

spiffxp:
Will try to take a look today

prow/plugins/trigger/push.go
spiffxp: [Postsubmits default to always-run][1], do you know if this means they're always going to report to PRs as well?  That would probably be an unwelcome default for something like kubernetes/kubernetes

[1]: https://github.com/kubernetes/test-infra/blob/0214d73dd35b87730d5fa05d713c81c9404b51fa/prow/config/jobs.go#L330-L331

spiffxp (changes requested):
We're much closer, but right now I'm concerned over whether this would spam or confuse our contributors.  

What would get me to merge-and-iterate:
- gate the `handlePE` logic behind a config, flag, or job annotation that defaults to "false"
- put some TODO comments within `handlePE` about what needs to be resolved as far as reporting to correct PR, not spamming, etc.
- address the `pjutil.NewPostsubmit` review comment

prow/plugins/trigger/push.go
```diff
@@ -72,20 +98,35 @@ func handlePE(c Client, pe github.PushEvent) error {
 	}
 
 	postsubmits := getPostsubmits(c.Logger, c.GitClient, c.Config, org+"/"+repo, shaGetter)
+	// TODO: It seems there is a bug in GH API, which returns an empty list of
+	// PRs for all commits listed by a push event, if it is queried immediately after
+	// receiving a push event. Waiting for 1 second seems to successfully work around this. Remove this once the bug is fixed.
+	if len(postsubmits) > 0 {
+		time.Sleep(1 * time.Second)
+	}
```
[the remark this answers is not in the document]
adshmh: From my tests, it seems to happen quite consistently, e.g. calling GitHub API to list a commit's PRs in the push event handler returned an empty list every time I tested it. When adding a 1 second sleep, it did not return an empty list even once.

I reported this to GitHub support in April 2020, and the responses from GitHub team seems to confirm the bug. I just asked for an update, and they seem to acknowledge the issue, but there is currently no promise of or an ETA for a fix.

Thank you for the detailed review. I am currently working on addressing the requested changes. Perhaps part of https://github.com/kubernetes/test-infra/issues/10402#issuecomment-785387313 can also be addressed here.

fejta:
Can we:
1) Do most reporting on the commit, rather than the PR?
2) Optionally drop a note on the PR that postsubmits were scheduled to commit X and to look there fore results? 

Mentioned this in #10402

adshmh:
> Can we:
> 
>     1. Do most reporting on the commit, rather than the PR?
> 
>     2. Optionally drop a note on the PR that postsubmits were scheduled to commit X and to look there fore results?
> 
> 
> Mentioned this in #10402

Thank you for the review. 
Should the reporting be kept as is, i.e. to the most recent commit, identified by the `after` field of the push event?  Or should it be changed to report each postsubmit job to the specific commit that triggered it?

Considering the scenario in [this review comment](https://github.com/kubernetes/test-infra/pull/17142#discussion_r579710077), the latter seems more consistent with dropping a single note on the specific PR that triggered a postsubmit.

fejta:
Is this looking for review again? Not sure.

adshmh:
> Is this looking for review again? Not sure.

Yes, would appreciate another review. Previous reviews have been addressed, in particular https://github.com/kubernetes/test-infra/pull/17142#issuecomment-788316093  and https://github.com/kubernetes/test-infra/issues/10402#issuecomment-785387313 (the latter is partially addressed, as I think a separate PR is needed to listen to /retest on commits)

PR #17153, opened 2020-04-08
Add serathius to sig-instrumentation approvers

serathius:
/cc @brancz


brancz:
While we’re at it could you also add @dashpole?

serathius:
done

brancz:
Sorry for not catching earlier, as Piotr is emeritus, we should probably remove him from here.

serathius:
done

PR #17154, opened 2020-04-08
don't use kind + bazel on 1.17

BenTheElder:
/cc @spiffxp @justaugustus 

FYI @neolit123, the "stable" version doesn't work now, and we don't want to upgrade a release branch to HEAD.

This is the only place I found us using kind + bazel in the kubernetes 1.17 branch, so we can just disable that for now until we upgrade or find the change.


BenTheElder:
ref https://github.com/kubernetes/kubernetes/pull/89278

PR #17156, opened 2020-04-08
Remove ci-kubernetes-e2e-gce-min-node-permissions

spiffxp:
It has never passed in 923 days since it first ran on 2017-09-28


spiffxp:
Took a while to track down, it was added via https://github.com/kubernetes/test-infra/pull/4685

PR #17157, opened 2020-04-08
Ghcache: partition by Authorization header

alvaroaleman:
Currently, ghcache uses the request url as cache key which means that if one instance is used with multiple tokens, requests to the same path will invalidate each other because they return different etags:

```
$ curl -Ss -i -H "Authorization: Bearer << some-token>>" https://api.github.com/users/alvaroaleman|grep -i etag:
ETag: "a14c08e67675f0ed57c680ea06a8ca1f"
```
vs
```
$ curl -Ss -i https://api.github.com/users/alvaroaleman|grep -i etag:
ETag: W/"0b9ccb547400349ca1f622f2cb02601a"
```

This PR fixes that by adding wrapper that creates a cache per unique `Authorization` header. 

### DIFF
--- a/ghproxy/ghcache/ghcache.go
+++ b/ghproxy/ghcache/ghcache.go
@@ -206,10 +206,29 @@ func (u upstreamTransport) RoundTrip(req *http.Request) (*http.Response, error) 
 	return resp, nil
 }
 
+const LogMessageWithDiskPartitionFields = "Not using a partitioned cache because legacyDisablePartitioningByAuthHeader is true"
+
 // NewDiskCache creates a GitHub cache RoundTripper that is backed by a disk
 // cache.
 // It supports a partitioned cache.
-func NewDiskCache(delegate http.RoundTripper, cacheDir string, cacheSizeGB, maxConcurrency int) http.RoundTripper {
+func NewDiskCache(delegate http.RoundTripper, cacheDir string, cacheSizeGB, maxConcurrency int, legacyDisablePartitioningByAuthHeader bool) http.RoundTripper {
+	if legacyDisablePartitioningByAuthHeader {
+		diskCache := diskcache.NewWithDiskv(
+			diskv.New(diskv.Options{
+				BasePath:     path.Join(cacheDir, "data"),
+				TempDir:      path.Join(cacheDir, "temp"),
+				CacheSizeMax: uint64(cacheSizeGB) * uint64(1000000000), // convert G to B
+			}))
+		return NewFromCache(delegate,
+			func(partitionKey string) httpcache.Cache {
+				logrus.WithField("cache-base-path", path.Join(cacheDir, "data", partitionKey)).
+					WithField("cache-temp-path", path.Join(cacheDir, "temp", partitionKey)).
+					Warning(LogMessageWithDiskPartitionFields)
+				return diskCache
+			},
+			maxConcurrency,
+		)
+	}
 	return NewFromCache(delegate,
 		func(partitionKey string) httpcache.Cache {
 			return diskcache.NewWithDiskv(
--- a/ghproxy/ghcache/ghcache.go
+++ b/ghproxy/ghcache/ghcache.go
@@ -91,10 +91,19 @@ var pendingOutboundConnectionsGauge = prometheus.NewGauge(prometheus.GaugeOpts{
 	Help: "How many pending requests are waiting to be sent to GitHub servers.",
 })
 
+var cachePartitionsCounter = prometheus.NewCounterVec(
+	prometheus.CounterOpts{
+		Name: "ghcache_cache_parititions",
+		Help: "Which cache partitions exist",
+	},
+	[]string{"token_hash"},
+)
+
 func init() {
 
 	prometheus.MustRegister(outboundConcurrencyGauge)
 	prometheus.MustRegister(pendingOutboundConnectionsGauge)
+	prometheus.MustRegister(cachePartitionsCounter)
 }
 
 func cacheResponseMode(headers http.Header) CacheResponseMode {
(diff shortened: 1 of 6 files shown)


BenTheElder:
this owners file needs more reviewers / approvers https://github.com/kubernetes/test-infra/blob/master/ghproxy/OWNERS

ghproxy/ghcache/ghcache.go
```diff
@@ -202,39 +211,55 @@ func (u upstreamTransport) RoundTrip(req *http.Request) (*http.Response, error)
...
-			TempDir:      path.Join(cacheDir, "temp"),
-			CacheSizeMax: uint64(cacheSizeGB) * uint64(1000000000), // convert G to B
-		})),
+	return NewFromCache(delegate,
+		func(partitionKey string) httpcache.Cache {
+			return diskcache.NewWithDiskv(
+				diskv.New(diskv.Options{
+					BasePath:     path.Join(cacheDir, "data", partitionKey),
+					TempDir:      path.Join(cacheDir, "temp", partitionKey),
+					CacheSizeMax: uint64(cacheSizeGB) * uint64(1000000000), // convert G to B
```
stevekuznetsov, line 220: This makes the `cacheSizeGB` misleading as it's the size of one cache partition and not the whole thing, right?

ghproxy/ghcache/ghcache.go
```diff
@@ -202,39 +211,55 @@ func (u upstreamTransport) RoundTrip(req *http.Request) (*http.Response, error)
...
+				}))
+		},
 		maxConcurrency,
 	)
 }
 
 // NewMemCache creates a GitHub cache RoundTripper that is backed by a memory
 // cache.
```
stevekuznetsov, line 228: This does or does not support partitioning?

ghproxy/ghcache/partitioner.go
```diff
@@ -0,0 +1,60 @@
...
+	return r.Header.Get("Authorization")
+}
+
+func (prt *partitioningRoundTripper) RoundTrip(r *http.Request) (*http.Response, error) {
+	cachePartition := getCachePartition(r)
+
+	prt.lock.Lock()
+	roundTripper, found := prt.roundTrippers[cachePartition]
+	if !found {
+		cachePartitionsCounter.WithLabelValues(cachePartition).Add(1)
```
stevekuznetsov, line 53: Doesn't this put the paintext auth header into the prom label?

ghproxy/ghcache/ghcache.go
```diff
@@ -91,10 +91,19 @@ var pendingOutboundConnectionsGauge = prometheus.NewGauge(prometheus.GaugeOpts{
 	Help: "How many pending requests are waiting to be sent to GitHub servers.",
 })
 
+var cachePartitionsCounter = prometheus.NewCounterVec(
+	prometheus.CounterOpts{
+		Name: "ghcache_cache_parititions",
+		Help: "How many cache partitions exist",
```
stevekuznetsov, line 97: Not really, since the value is always 1 and it's just a vector of information.

ghproxy/ghproxy.go
```diff
@@ -112,7 +112,7 @@ func (o *options) validate() error {
 func flagOptions() *options {
 	o := &options{}
 	flag.StringVar(&o.dir, "cache-dir", "", "Directory to cache to if using a disk cache.")
-	flag.IntVar(&o.sizeGB, "cache-sizeGB", 0, "Cache size in GB if using a disk cache.")
+	flag.IntVar(&o.sizeGB, "cache-sizeGB", 0, "Cache size in GB per unique token if using a disk cache.")
```
stevekuznetsov, line 115: Ah I see

ghproxy/ghcache/ghcache.go
alvaroaleman: yes, thats why I updated the flag description

ghproxy/ghcache/ghcache.go
alvaroaleman: It does, it just doesn't need to have a parition key for that

ghproxy/ghcache/partitioner.go
alvaroaleman: Oups

ghproxy/ghcache/ghcache.go
stevekuznetsov: Could you leave a comment to that effect? It's not obvious from a read of the code

ghproxy/ghcache/ghcache.go
```diff
@@ -91,10 +91,19 @@ var pendingOutboundConnectionsGauge = prometheus.NewGauge(prometheus.GaugeOpts{
 	Help: "How many pending requests are waiting to be sent to GitHub servers.",
 })
 
+var cachePartitionsCounter = prometheus.NewCounterVec(
+	prometheus.CounterOpts{
+		Name: "ghcache_cache_parititions",
+		Help: "Which cache partitions exist",
+	},
+	[]string{"name"},
```
stevekuznetsov, line 99: Might as well call it `token_hash`

alvaroaleman:
@stevekuznetsov @cjwagner one thing that should be noted is that this PR will invalidate any pre-existing cache entries

ghproxy/ghcache/ghcache.go
alvaroaleman: done

ghproxy/ghcache/ghcache.go
alvaroaleman: done

stevekuznetsov:
Oh, true. Would it be possible to write a simple utility to migrate?

alvaroaleman:
> Oh, true. Would it be possible to write a simple utility to migrate?

I have written some naive mechanism for that in the second commit, WDYT? It has the drawback though that the first request with a given token may take quite some time because of the copying

alvaroaleman:
> I have written some naive mechanism for that in the second commit, WDYT? It has the drawback though that the first request with a given token may take quite some time because of the copying

I benchmarked this and for us, copying this over takes about five minutes. Doing this blocks the entire ghproxy. On top of that, we currently pull all tokens, including user-provided ones in Deck, through ghcache, so the total number of paritions may get higher.

Given this, we decided that a feature flag that turns this off by default and advises ppl on how to manually warm up a partitioned cache is the better option.

cjwagner:
> Given this, we decided that a feature flag that turns this off by default and advises ppl on how to manually warm up a partitioned cache is the better option.

Where are the instructions for how to manually warm up the cache? I'm seeing `Not using a partitioned cache because legacyDisablePartitioningByAuthHeader is true` in the logs which is not actionable. When I search for `legacyDisablePartitioningByAuthHeader` and dig around the code I can see the flag it is associated with in ghproxy, but the flag help text just says `Bigger setups are advise to manually warm this up from an existing cache.` without explaining how that is achieved.

alvaroaleman:
> Where are the instructions for how to manually warm up the cache? I'm seeing Not using a partitioned cache because legacyDisablePartitioningByAuthHeader is true in the logs which is not actionable. When I search for legacyDisablePartitioningByAuthHeader and dig around the code I can see the flag it is associated with in ghproxy, but the flag help text just says Bigger setups are advise to manually warm this up from an existing cache. without explaining how that is achieved.

@cjwagner well the message says ` If you are a bigger Prow setup, you should copy your existing cache directory to the directory mentioned in the `%s` messages to warm up the partitioned-by-auth-header cache, then set the flag to false` ([ref](https://github.com/kubernetes/test-infra/blob/1e6a379fba6ad25d299b2ac68b7e6e35d49d124b/ghproxy/ghproxy.go#L139)). Is that not enough? We ended up just using `kubectl cp`. Presumably k8s and openshift are the only setups big enough to need this.

cjwagner:
@alvaroaleman That message is sufficiently actionable, but I didn't notice that error message. I was looking at the frequent `Not using a partitioned cache because legacyDisablePartitioningByAuthHeader is true` messages and then went straight to the flag description.
Thanks for the clarification.

PR #17158, opened 2020-04-08
use kind from newer commit

BenTheElder:
https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/89278/pull-kubernetes-e2e-kind/1248012078044680192 ???

how does anything ever work. I'm also not positive that this works properly with bootstrap but i'm trying not to completely rewrite the job >.>

/cc @amwat @spiffxp 
we really need to replace the kubetest-kind thing.

### DIFF
--- a/config/jobs/kubernetes/sig-release/release-branch-jobs/1.17.yaml
+++ b/config/jobs/kubernetes/sig-release/release-branch-jobs/1.17.yaml
@@ -1249,9 +1249,11 @@ presubmits:
     spec:
       containers:
       - args:
+        - --repo=k8s.io/$(REPO_NAME)=$(PULL_REFS)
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes/sig-release/release-branch-jobs/1.17.yaml
```diff
@@ -1249,9 +1249,11 @@ presubmits:
     spec:
       containers:
       - args:
+        - --repo=k8s.io/$(REPO_NAME)=$(PULL_REFS)
+        - --repo=sigs.k8s.io/kind=7b28ee53357ee77f111d2e7b1b602876dc6d0d4c
```
BenTheElder, line 1253: ughhh this job is actually podutils + kubetest, not bootstrap + kubetest. too bad we can't auto validate this

PR #17161, opened 2020-04-09
prow: fixup nil pointer introduced by s3 refactoring of gcs-upload

sbueringer:
The refactoring of gcs-upload yesterday introduced a critical bug (and probably breaks a lot of builds, #16722). This PR fixes the nil pointer introduced #16722


alvaroaleman:
I should be able to approve this, pkg/io is only used by Prow, xref: https://github.com/kubernetes/test-infra/pull/17162

sbueringer:
@fejta got rid of a few more pointers :)

fejta:
> pkg/io is only used by Prow

You sure this is true?

PR #17162, opened 2020-04-09
Move pkg/io to prow/io

alvaroaleman:
The package is only used by Prow, where I already am an approver. Alternatively, we could also move the package.

/assign @fejta @stevekuznetsov


pkg/io/OWNERS
```diff
@@ -0,0 +1,4 @@
+# See the OWNERS docs at https://go.k8s.io/owners
+
+approvers:
```
fejta, line 3: Please add the other relevant approvers. This will make you the only person who ever gets assigned io changes.

alvaroaleman: Its definitely only used by prow:
```
$ ack -Q pkg/io -l|egrep -v '^prow'
BUILD.bazel
pkg/io/providers/providers_test.go
pkg/io/providers/BUILD.bazel
pkg/io/opener.go
pkg/io/BUILD.bazel
```

Do you mind if I instead just move the package under `prow/`?

fejta: SGTM

PR #17164, opened 2020-04-09
add getRepo function in the fake github client

droslean:

The `getRepo()` function was missing from the fake github client. 

The func will return a basic `github.FullRepo` object with few dynamically changed fields given from the arguments


alvaroaleman:
We should get the fakegithub client to implement the full interface. I at some point had a local branch for that but never PRd it.
/lgtm
/approve

PR #16782, opened 2020-03-14
hook: allow for plugin-specific user agents

stevekuznetsov:
Signed-off-by: Steve Kuznetsov <skuznets@redhat.com>

/assign @cjwagner

PR #17166, opened 2020-04-09
Replace the removed plugin log label.

cjwagner:
This was accidentally removed in https://github.com/kubernetes/test-infra/pull/16782

/assign @stevekuznetsov


cjwagner:
/hold cancel
This shouldn't be contentious.

PR #17173, opened 2020-04-09
GCSUpload: Remove special case for GCP

alvaroaleman:
This PR removes the special case for GCS in the podutils/gcs package. We already have a special case for it within pkg/io.

Doing this in a dedicated PR so its trivial to revert if need be.

/assign @fejta @sbueringer

### DIFF
--- a/prow/pod-utils/gcs/upload.go
+++ b/prow/pod-utils/gcs/upload.go
@@ -22,12 +22,9 @@ import (
 	"io"
 	"net/url"
 	"os"
-	"strings"
 	"sync"
 
-	"cloud.google.com/go/storage"
 	"github.com/sirupsen/logrus"
-	"google.golang.org/api/option"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	utilpointer "k8s.io/utils/pointer"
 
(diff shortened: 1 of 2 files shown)


prow/pod-utils/gcs/upload.go
```diff
@@ -44,24 +42,10 @@ type destToWriter func(dest string) dataWriter
...
-		gcsClient, err := storage.NewClient(context.Background(), options...)
-		if err != nil {
-			return fmt.Errorf("create client: %w", err)
-		}
-		dtw := func(dest string) dataWriter {
-			return gcsObjectWriter{gcsClient.Bucket(bucket).Object(dest).NewWriter(context.Background())}
-		}
-		return upload(dtw, uploadTargets)
+	if parsedBucket.Scheme == "" {
+		parsedBucket.Scheme = "gs"
```
sbueringer, line 48: gcsObjectWriter struct should be unused now

alvaroaleman: removed it

sbueringer:
/lgtm
/hold
I guess @fejta also wants to take a look

PR #17175, opened 2020-04-09
github: publish user info for token hashes

stevekuznetsov:
Currently, GitHub API interactions are sharded by the hash of the
authorization token used to make them, in order to track between
accounts that may be using ghproxy concurrently. This new metric allows
for a join on labels to identify those users by name.

/assign @alvaroaleman @cjw

### DIFF
--- a/prow/github/client.go
+++ b/prow/github/client.go
@@ -778,8 +778,8 @@ func (c *client) doRequest(method, path, accept string, body interface{}) (*http
 	if err != nil {
 		return nil, err
 	}
-	if token := c.getToken(); len(token) > 0 {
-		req.Header.Set("Authorization", "Token "+string(token))
+	if header := c.authHeader(); len(header) > 0 {
+		req.Header.Set("Authorization", header)
 	}
 	if accept == acceptNone {
 		req.Header.Add("Accept", "application/vnd.github.v3+json")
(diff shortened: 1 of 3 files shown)


stevekuznetsov:
/assign @cjwagner 

Dove-tails with https://github.com/kubernetes/test-infra/pull/17157

prow/github/client.go
```diff
@@ -811,6 +827,14 @@ func (c *client) getUserData() error {
 	// email needs to be publicly accessible via the profile
 	// of the current account. Read below for more info
 	// https://developer.github.com/v3/users/#get-a-single-user
+
+	// record information for the user
+	authHeader := bytes.NewBufferString("Token ")
+	authHeader.Write(c.getToken())
+	hasher := sha256.New()
+	hasher.Write(authHeader.Bytes())
```
alvaroaleman, line 835: Why are we not simply using `crypto/sha256.Sum256([]byte(fmt.Sprintf("Token: %s@, c.getToken())))`?

stevekuznetsov: Is this python?!

stevekuznetsov: I copy-pasted this from how I did it in `ghmetrics.go` -- if the `Sum256` is identical, I can refactor these two to do that. WDYT?

alvaroaleman: yeah please do

stevekuznetsov: updated

PR #17176, opened 2020-04-09
we have kind v0.7 as the stable version in kubetest / kubekins now, r…

BenTheElder:
…evert workarounds for 1.17

/cc @spiffxp @amwat


BenTheElder:
currently the only issue is timeout. wasting time building kind + quick building is expensive

PR #17179, opened 2020-04-10
jobs: move pull-kubernetes-{verify,dependencies} release-branch jobs

### LINKED ISSUE #13886: pull-kubernetes-{dependencies,verify} shouldn't be manually forked
Ref: https://github.com/kubernetes/test-infra/pull/13723#issuecomment-517158167

See https://github.com/kubernetes/test-infra/pull/13672 and https://github.com/kubernetes/test-infra/pull/13885 where the jobs are being added manually per-release. Instead, this should be automatically handled by config-forker and shouldn't need manual update.

/cc @Katharine

spiffxp:
pretend as if they had been forked with config-forker, this way they'll get
pruned when the rest of the release-branch jobs do

Fixes #13886


spiffxp:
/assign @justaugustus @nikhita 
for the /lgtm

PR #17183, opened 2020-04-10
prow: deck: add s3 support

sbueringer:
This PR adds S3 Support to deck (see #11260)

Follow-up PRs:
* [ ] rename gcsBucket & gcsArtifactHandle to something more neutral like storageBucket & storageArtifactHandle
* [ ] remove noGcsAuth flag in a few weeks
* [ ] create constants for filenames like `finished.json`
* [ ] because of the job_url_prefix format (e.g. prow.k8s.io/view/gcs) only one storageProvider can be used at the same time. We need another PR to fix that


prow/cmd/deck/main.go
```diff
@@ -697,16 +700,26 @@ func initSpyglass(cfg config.Getter, o options, mux *http.ServeMux, ja *jobs.Job
 	}
 	c, err := storage.NewClient(ctx, options...)
 	if err != nil {
-		logrus.WithError(err).Fatal("Error getting GCS client")
+		// We're assuming that storage client creation must be successful when gcsCredentialsFile is set
+		// In other cases it might be okay that it fails, because the s3 client might be used instead.
+		if o.gcsCredentialsFile != "" {
+			logrus.WithError(err).Fatal("Error getting GCS client")
+		} else {
+			logrus.WithError(err).Info("Error getting GCS client, this is not a problem if only the s3 client is used")
```
fejta, line 708: I don't agree with this. Either we want credentials and the error here is a problem, or we do not care and should be using the anonymous client.

Ignoring errors doesn't seem like a great way to handle this.

`gcsCredentialsFile == ""` + using GCS  is not an edge case, should be the best practice on GKE

prow/cmd/deck/main.go
```diff
@@ -697,16 +700,26 @@ func initSpyglass(cfg config.Getter, o options, mux *http.ServeMux, ja *jobs.Job
 	}
 	c, err := storage.NewClient(ctx, options...)
```
fejta, line 701: Perhaps we should flip around the flags to where we use anonymous authentication unless a flag is specified

prow/cmd/deck/pr_history.go
```diff
@@ -268,8 +270,19 @@ func getPRHistory(url *url.URL, config *config.Config, gcsClient *storage.Client
...
 	jobCommitBuilds := make(map[string]map[string][]buildData)
 
-	for bucketName, gcsPaths := range toSearch {
-		bucket := gcsBucket{bucketName, gcsClient.Bucket(bucketName)}
+	for bucket, gcsPaths := range toSearch {
+		// bucket is the bucket field of the GCSConfiguration, which means it could be missing the
+		// storageType prefix (but it's deprecated to use a bucket name without <storage-type>:// prefix)
+		parsedBucket, err := url.Parse(bucket)
+		if err != nil {
+			return template, fmt.Errorf("parse bucket %s: %v", bucket, err)
```
fejta, line 278: Really would prefer we use %w to %v here. It is useful to include the inner error in case the caller has a need to consider it.

%w should be our reflexive behavior for `return fmt.Errorf()` and the only reason it isn't is because this feature is very new.

prow/io/iterator.go
```diff
@@ -0,0 +1,78 @@
...
+type ObjectIterator interface {
+	Next(ctx context.Context) (done bool, attr ObjectAttributes, err error)
+}
+
+// GCSObjectIterator implements ObjectIterator for GCS
+type GCSObjectIterator struct {
+	Iterator *storage.ObjectIterator
+}
+
+func (g GCSObjectIterator) Next(_ context.Context) (done bool, attr ObjectAttributes, err error) {
```
fejta, line 34: Please remove the named return values, as they are not being used correctly (nor are they necessary here).

* this isn't a `(float, float)` situation like `(lat, lon float)` or `(temp, humidity float)` where the names clarify things
* You aren't even using them, aka
  - inner declaration of `err` on L35
  - never set done
  - do not use attr on L37, L40


This should just be `(bool, ObjectAttributes, error)`

https://github.com/golang/go/wiki/CodeReviewComments#named-result-parameters

prow/io/iterator.go
```diff
@@ -0,0 +1,78 @@
...
+	} else {
+		// directory
+		attr.Name = oAttrs.Prefix
+		attr.IsDir = true
+	}
+	return false, attr, nil
+}
+
+// GCSObjectIterator implements ObjectIterator via the gocloud/blob package
+type OpenerObjectIterator struct {
```
fejta, line 56: No need to export this type

prow/cmd/deck/main.go
sbueringer: I agree with you I just have no idea how to detect this :). For example with the current code in the io package it's possible to handover neither s3-credentials nor gcs-credentials. The GCS credential discovery fails and the one from s3 works.

Ah maybe I have an idea. I guess we can make it mandatory to set `gcs-no-auth` when s3 should be used, right? So then I can throw an error whenever the gcs client creation fails?

sbueringer: My case which failed was:
* s3-credentials-file was set
* gcs-credentials-file was not set

opener would have worked but process failed because the gcs client couldn't be created. So this case should be solvable by setting `gcs-no-auth`

prow/cmd/deck/job_history.go
```diff
@@ -165,20 +187,33 @@ func (bucket gcsBucket) listSubDirs(prefix string) ([]string, error) {
 		prefix += "/"
 	}
 	dirs := []string{}
-	it := bucket.Objects(context.Background(), &storage.Query{
-		Prefix:    prefix,
-		Delimiter: "/",
-	})
+	ctx := context.TODO()
```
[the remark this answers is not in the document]
sbueringer: Of course. How high should I set the timeout?

prow/cmd/deck/job_history.go
```diff
@@ -90,18 +93,31 @@ type jobHistoryTemplate struct {
...
+	if bucket.storageType == "gs" {
+		obj := bucket.Object(key)
+		rc, err := obj.NewReader(context.Background())
+		if err != nil {
+			return []byte{}, fmt.Errorf("failed to get reader for GCS object: %v", err)
+		}
+		return ioutil.ReadAll(rc)
+	}
+
+	rc, err := bucket.Opener.Reader(context.Background(), fmt.Sprintf("%s://%s/%s", bucket.storageType, bucket.name, key))
```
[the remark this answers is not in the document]
sbueringer: I would also switch to context.WithTimeout + TODO() here. Just not sure how high to set the timeout

prow/tide/history/history.go
```diff
@@ -45,11 +45,17 @@ type History struct {
...
 	logSizeLimit int
 
-	opener io.Opener
+	opener Opener
 	path   string
 }
 
-func readHistory(maxRecordsPerKey int, opener io.Opener, path string) (map[string]*recordLog, error) {
+// Opener has methods to read and write paths
+type Opener interface {
```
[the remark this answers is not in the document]
sbueringer: Problem is that the Opener interface in the io package now has a few more methods. So I either specify a smaller interface here with just the necessary methods or extend the fakeOpener in the corresponding tests (same problem in gerrit & statusreconciler).

I guess extending the fakeopener would have been better? :)

prow/cmd/deck/job_history.go
```diff
@@ -90,18 +93,31 @@ type jobHistoryTemplate struct {
 }
 
 func (bucket gcsBucket) readObject(key string) ([]byte, error) {
-	obj := bucket.Object(key)
-	rc, err := obj.NewReader(context.Background())
+	if bucket.storageType == "gs" {
+		obj := bucket.Object(key)
+		rc, err := obj.NewReader(context.Background())
```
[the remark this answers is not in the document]
sbueringer: Ah just saw the comment. I refactor it to include a context

prow/cmd/deck/job_history.go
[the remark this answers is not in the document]
sbueringer: Let's also pass in a context here

prow/cmd/deck/pr_history.go
sbueringer: Yup got it. Some of the code was written before the last review and probably I also forgot it. Will try to default to it in the future

prow/cmd/deck/job_history.go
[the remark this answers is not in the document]
sbueringer: Let's also pass in a context here

prow/tide/history/history.go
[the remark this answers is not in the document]
sbueringer: Ah maybe I understood you wrong. So I see 2 options. Extending the fakeopener with all methods defined in the io package or changing the opener interface to unexported (I guess the second one was your point?).

prow/cmd/deck/main.go
fejta: I'm fine with that, although it might be suboptimal for s3 users.

Another approach would be to replace `gcs-no-auth` with `gcs-require-default-creds`:
* Require creating a client with explicit credentials to succeed (aka gcsCredFile != "")
* Create a client with default credentials (aka gcsCredFile == "")
  - If this succeeds, use it
  - If it fails and `gcs-require-default-creds` then fatal
  - otherwise create an anonymous client, which should never fail

This will provide GCP users with the expected default behavior (use default creds by default) and also provide an option to blow up if default creds aren't being used.

prow/tide/history/history.go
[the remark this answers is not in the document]
fejta: I assuming that history.Opener is only used by the history package, thus it isn't obvious to me why the interface needs to be exported?

prow/io/iterator.go
sbueringer: Yup. Done

prow/io/iterator.go
sbueringer: Yup you're right. Removed them.

I thought it would be useful, but you're right and I guess the threshold for using named return values should be a lot higher considering how easy it is to introduce bugs when using them.

prow/cmd/deck/pr_history.go
sbueringer: Done also changed it in a few other cases where I touched the code.

prow/tide/history/history.go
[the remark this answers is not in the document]
sbueringer: Doesn't have to be.. Just a mistake on my side

prow/cmd/deck/job_history.go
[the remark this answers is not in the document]
sbueringer: Okay so I refactored deck to hand over context for the 3 endpoints pr-history / job-history and the main spyglass page. Now the only question is if and where I should add a timeout. Rather top-level (where I get the context from the request) or on the leaves right before s3/gcs is called?

prow/tide/history/history.go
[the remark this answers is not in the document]
sbueringer: So I made the Opener interface private, also added the same to gerrit and statusreconciler

prow/cmd/deck/main.go
sbueringer: This would mean adding the new flag and deprecating the old one, right? I should probably also enforce that both flags (`gcs-no-auth` and `gcs-require-default-creds`) are not set at the same time.

prow/cmd/deck/main.go
sbueringer: Let's continue the conversation here: https://github.com/kubernetes/test-infra/pull/17183#discussion_r406898027

sbueringer:
@fejta @alvaroaleman Can you please take another look? Thank you :)

prow/cmd/deck/job_history.go
```diff
@@ -160,39 +182,63 @@ func readJSON(bucket storageBucket, key string, data interface{}) error {
...
+	if bucket.storageType == "gs" {
+		it = io.GCSObjectIterator{
+			Iterator: bucket.Objects(ctx, &storage.Query{
+				Prefix:    prefix,
+				Delimiter: "/",
+			}),
+		}
+	} else {
+		var err error
+		it, err = bucket.Opener.Iterator(ctx, fmt.Sprintf("%s://%s/%s", bucket.storageType, bucket.name, prefix), "/")
```
alvaroaleman, line 200: Why can we not move the extra-sausage-for-gcs into the `Opener.Iterator` implementation?

prow/cmd/deck/job_history.go
```diff
@@ -160,39 +182,63 @@ func readJSON(bucket storageBucket, key string, data interface{}) error {
...
+		if err != nil {
+			return []string{}, err
```
alvaroaleman, line 202: similarly, why construct an empty string slice rather than just returning `nil, err` here?

prow/cmd/deck/job_history.go
```diff
@@ -240,16 +286,45 @@ func (bucket gcsBucket) listBuildIDs(root string) ([]int64, error) {
...
 	buildID = emptyID
 	p := strings.TrimPrefix(url.Path, "/job-history/")
-	s := strings.SplitN(p, "/", 2)
-	if len(s) < 2 {
-		err = fmt.Errorf("invalid path (expected /job-history/<gcs-path>): %v", url.Path)
-		return
+	// examples for p:
+	// * new format: s3/prow-artifacts/pr-logs/directory/bazel-build-test
+	// * old format: kubernetes-jenkins/pr-logs/directory/pull-cluster-api-provider-openstack-test
+	if strings.HasPrefix(p, "gs") || strings.HasPrefix(p, "s3") {
```
alvaroaleman, line 306: please change that to `gs/`/`s3/` so it doesn't match a bucketName that starts with `gs` or `s3`

prow/cmd/deck/job_history.go
```diff
@@ -240,16 +286,45 @@ func (bucket gcsBucket) listBuildIDs(root string) ([]int64, error) {
 	return ids, nil
 }
 
-func parseJobHistURL(url *url.URL) (bucketName, root string, buildID int64, err error) {
+// parseJobHistURL parses the job History URL
+// example urls:
+// * new format: https://prow.k8s.io/job-history/s3/kubernetes-jenkins/pr-logs/directory/pull-capi?buildId=1245584383100850177
+// * old format: https://prow.k8s.io/job-history/kubernetes-jenkins/pr-logs/directory/pull-capi?buildId=1245584383100850177
```
alvaroaleman, line 292: Please update the docs on `handleJobHistory` to describe the new format.

prow/cmd/deck/job_history.go
alvaroaleman: Also, lets simplify this and not have two completely different handlings:

* If there is no provider in the url path, inject `gs`
* Only handle three element urls and remove the old handling (The error needs to mention both options though as we wont know which one was used by the time we emit it)

prow/cmd/deck/job_history.go
```diff
@@ -240,16 +286,45 @@ func (bucket gcsBucket) listBuildIDs(root string) ([]int64, error) {
...
+		bucketName = s[1]
+		root = s[2] // `root` is the root "directory" prefix for this job's results
+	} else {
+		// handle old format
+		s := strings.SplitN(p, "/", 2)
+		if len(s) < 2 {
+			err = fmt.Errorf("invalid path (expected /job-history/<gcs-path>): %v", url.Path)
+			return
+		}
+		storageType = "gs"
```
alvaroaleman, line 323: Another nit: All of the `gs` or `s3` should be a const that is define in one central place (pkg/io/providers?)

prow/cmd/deck/job_history.go
```diff
@@ -353,18 +428,18 @@ func (a int64slice) Swap(i, j int)      { a[i], a[j] = a[j], a[i] }
...
 	tmpl := jobHistoryTemplate{}
 
-	bucketName, root, top, err := parseJobHistURL(url)
+	storageType, bucketName, root, top, err := parseJobHistURL(url)
 	if err != nil {
 		return tmpl, fmt.Errorf("invalid url %s: %v", url.String(), err)
 	}
 	tmpl.Name = root
-	bucket := gcsBucket{bucketName, gcsClient.Bucket(bucketName)}
+	bucket := gcsBucket{bucketName, storageType, gcsClient.Bucket(bucketName), opener}
```
alvaroaleman, line 440: Could you make a commit at the end where you rename `gcsBucket` to something more neutral like `objStorageBucket`?

prow/cmd/deck/job_history_test.go
```diff
@@ -23,30 +23,34 @@ import (
...
+			storageType: "gs",
+			root:        "logs/bar-e2e",
+			id:          emptyID,
 		},
 		{
-			address: "http://www.example.com/job-history/foo-bucket/logs/bar-e2e?buildId=123456789123456789",
-			bktName: "foo-bucket",
-			root:    "logs/bar-e2e",
-			id:      123456789123456789,
+			address:     "http://www.example.com/job-history/foo-bucket/logs/bar-e2e?buildId=123456789123456789",
```
alvaroaleman, line 49: Please add tests for the new format

prow/cmd/deck/pr_history_test.go
```diff
@@ -291,7 +299,7 @@ func TestGetPRBuildData(t *testing.T) {
 				jobName:      "build-snowman",
 				index:        0,
 				ID:           "456",
-				SpyglassLink: "/view/gcs/chum-bucket/pr-logs/pull/123/build-snowman/456",
+				SpyglassLink: "/view/gs/chum-bucket/pr-logs/pull/123/build-snowman/456",
```
alvaroaleman, line 302: We have to keep supporting the old urls because of compatibility reasons

prow/io/iterator.go
```diff
@@ -0,0 +1,95 @@
...
+	"strings"
+
+	"cloud.google.com/go/storage"
+	"gocloud.dev/blob"
+	"google.golang.org/api/iterator"
+)
+
+type ObjectAttributes struct {
+	// Name is the name of the object or directory
+	Name string
```
alvaroaleman, line 31: This is confusing. So its the full path of the obj?

prow/pjutil/pjutil.go
```diff
@@ -278,7 +279,18 @@ func JobURL(plank config.Plank, pj prowapi.ProwJob, log *logrus.Entry) string {
 		_, gcsPath, _ := gcsupload.PathsForJob(gcsConfig, &spec, "")
 
 		prefix, _ := url.Parse(plank.GetJobURLPrefix(pj.Spec.Refs))
-		prefix.Path = path.Join(prefix.Path, gcsConfig.Bucket, gcsPath)
+
+		// extract bucket name from gcsConfig
+		var bucketName string
+		if strings.Contains(gcsConfig.Bucket, "://") {
```
alvaroaleman, line 285: This needs tests. Also it might be simpler to always split and do the if/else branching based on the length of the split result

prow/spyglass/gcsartifact_fetcher_test.go
```diff
@@ -37,7 +38,7 @@ func TestNewGCSJobSource(t *testing.T) {
 	}{
 		{
 			name:        "Test standard GCS link",
-			src:         "test-bucket/logs/example-ci-run/403",
+			src:         "gs://test-bucket/logs/example-ci-run/403",
```
alvaroaleman, line 41: Please leave the existing tests, they have to continue passing due to compatibility concerns and add new ones

prow/spyglass/gcsartifact_fetcher_test.go
```diff
@@ -94,7 +95,7 @@ func TestArtifacts_ListGCS(t *testing.T) {
 	}{
 		{
 			name:   "Test ArtifactFetcher simple list artifacts",
-			source: "test-bucket/logs/example-ci-run/403",
+			source: "gs://test-bucket/logs/example-ci-run/403",
```
alvaroaleman, line 98: Same, please leave the existing tests and add new ones

prow/spyglass/gcsartifact_fetcher_test.go
```diff
@@ -149,19 +150,19 @@ func TestFetchArtifacts_GCS(t *testing.T) {
 		{
 			name:         "Fetch build-log.txt from valid source",
 			artifactName: "build-log.txt",
-			source:       "test-bucket/logs/example-ci-run/403",
+			source:       "gs://test-bucket/logs/example-ci-run/403",
```
alvaroaleman, line 153: leave+add

prow/tide/history/history.go
```diff
@@ -45,11 +45,17 @@ type History struct {
...
 	logSizeLimit int
 
-	opener io.Opener
+	opener opener
 	path   string
 }
 
-func readHistory(maxRecordsPerKey int, opener io.Opener, path string) (map[string]*recordLog, error) {
+// opener has methods to read and write paths
+type opener interface {
```
alvaroaleman, line 53: Could you move all the consumer definitions of the `opener` interface from this PR into a dedicated one that we can merge already to keep the size of this a bit more manageable?

prow/cmd/deck/job_history.go
```diff
@@ -89,22 +92,35 @@ type jobHistoryTemplate struct {
...
 
-func (bucket gcsBucket) readObject(key string) ([]byte, error) {
-	obj := bucket.Object(key)
-	rc, err := obj.NewReader(context.Background())
+func (bucket gcsBucket) readObject(ctx context.Context, key string) ([]byte, error) {
+	if bucket.storageType == "gs" {
+		obj := bucket.Object(key)
+		rc, err := obj.NewReader(ctx)
+		if err != nil {
+			return []byte{}, fmt.Errorf("failed to get reader for GCS object: %w", err)
```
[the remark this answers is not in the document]
sbueringer: Nope, just copy&paste from the previous implementation. I change it to nil

prow/tide/history/history.go
sbueringer: Done > https://github.com/kubernetes/test-infra/pull/17301

prow/cmd/deck/job_history.go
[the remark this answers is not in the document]
sbueringer: Fixed

prow/cmd/deck/job_history.go
```diff
@@ -89,22 +92,35 @@ type jobHistoryTemplate struct {
...
+	defer rc.Close()
 	return ioutil.ReadAll(rc)
 }
 
 func (bucket gcsBucket) getName() string {
 	return bucket.name
 }
 
-func readLatestBuild(bucket storageBucket, root string) (int64, error) {
+func (bucket gcsBucket) getStorageType() string {
```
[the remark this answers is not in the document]
sbueringer: Fixed

prow/cmd/deck/job_history.go
sbueringer: Fixed

prow/cmd/deck/job_history.go
sbueringer: Fixed

prow/cmd/deck/job_history.go
sbueringer: Fixed

prow/cmd/deck/job_history.go
sbueringer: Fixed

prow/cmd/deck/job_history.go
sbueringer: I added it to the PR description.

prow/pjutil/pjutil.go
sbueringer: Fixed. I'll add tests in another commit

prow/cmd/deck/job_history.go
sbueringer: Fixed

prow/cmd/deck/main.go
sbueringer: Because of another comment I changed the implementation to use the gcsClient inside of the opener. So now the problem doesn't exist anymore. The NewOpener implementation:
* if gcsCredentialsFile is set: throws an error if the gcp client cannot be created
* if gcsCredentialsFile is not set: sets gcsClient to nil and logs a warning

The error is then thrown the first time the gcsClient is acutally used (which won't happen for s3 paths)

fejta:
If `--gcs-credentials-file` is unset we should:
* First try to use the default credentials (which may/probably exist)
* Otherwise try to create an anonymous client (which should always work)
* Otherwise set it to nil

We must not require the usage of a secret credentials file in order to authenticate.

alvaroaleman:
> First try to use the default credentials (which may/probably exist)

What would be really cool is if we could automatically fall back to anonymous auth rather than hard fail. I had to add a flag to Deck to explicitly tell it to do that, it would be great if it would just do that by default if all other mechanisms fail.

prow/cmd/deck/pr_history_test.go
sbueringer: Yup agree. What changes in this test is not really what urls we're supporting. In this cases we're using the same `testBucket` as before and we're also getting the PR build data of the same jobs. 

What changes is the generated spyglass links. So from now on we would generate spyglass links with `gs` instead of `gcs`. Doesn't mean if somebody still has an old gcs link that it won't work. 

Of course this assumes that it's okay that after this PR we would generate links with `gs` instead of `gcs`

prow/io/opener.go
```diff
@@ -93,6 +98,23 @@ func NewOpener(ctx context.Context, gcsCredentialsFile, s3CredentialsFile string
 	}, nil
 }
 
+func createGCSClient(ctx context.Context, gcsCredentialsFile string) (storageClient, error) {
```
sbueringer, line 101: @fejta @alvaroaleman So I think that should implement what you suggested. If that's alright, the only question is what to do with the `gcsNoAuth` flag. Should I just ignore it, deprecate it and add a deprecation warning if its set?

sbueringer:
@fejta @alvaroaleman please see: https://github.com/kubernetes/test-infra/pull/17183/files#r412400144

P.S. Most of the findings are addressed now. But I still have to extend a lot of the tests and also want to review my changes by myself because I rebased it onto master today after the remote lense PR was merged.

prow/io/opener.go
alvaroaleman: implementation looks good. For the flag, can you just make it a no-op and emit a deprecation warning when set? We can then remove it in a couple of weeks

prow/pjutil/pjutil.go
```diff
@@ -271,23 +272,38 @@ func ProwJobFields(pj *prowapi.ProwJob) logrus.Fields {
...
+		// extract bucket name from gcsConfig
+		var bucketName string
+		bucketSplit := strings.Split(gcsConfig.Bucket, "://")
+		switch len(bucketSplit) {
+		case 1:
+			bucketName = bucketSplit[0]
+		case 2:
+			bucketName = bucketSplit[1]
+		default:
+			return "", fmt.Errorf("calculating joburl: bucket field has invalid format %q, expected either <bucket-name> or <storage-provider>://<bucket-name>", gcsConfig.Bucket)
```
alvaroaleman, line 292: Nit: `bucket field %q has invalid format` rather than `bucket field has invalid format %q`

prow/pjutil/pjutil_test.go
```diff
@@ -701,6 +701,7 @@ func TestJobURL(t *testing.T) {
 		plank    config.Plank
 		pj       prowapi.ProwJob
 		expected string
+		wantErr  bool
```
alvaroaleman, line 704: At least for places where you add it, please do not use a boolean for expected errors, check if the error you get is the error you expect. I have found a decent set of tests in this codebase that used the boolean approach and happened to pass because something else errored :)

(I do realize there is currently only one possible error, but that might be different in say six or 12 months)

prow/plank/controller.go
```diff
@@ -433,8 +434,11 @@ func (c *Controller) syncPendingJob(pj prowapi.ProwJob, pm map[string]corev1.Pod
 			return nil
 		}
 	}
-
-	pj.Status.URL = pjutil.JobURL(c.config().Plank, pj, c.log)
+	var err error
+	pj.Status.URL, err = pjutil.JobURL(c.config().Plank, pj, c.log)
+	if err != nil {
```
alvaroaleman, line 439: The error we get here and below is terminal (as in: We know it will never succeed if it failed once. because we never update the DecorationConfig on a ProwJob), hence we should IMHO just log the error and proceed, otherwise jobs may get stuck in a given state

prow/pjutil/pjutil.go
alvaroaleman: Also, I think we should extract the part here that can error into a function that is run by checkconfig for all possible `DecorationConfigs`, to avoid getting to the point of deploying an invalid config

prow/spyglass/gcsartifact_fetcher.go
```diff
@@ -152,81 +154,88 @@ const (
...
+			PrivateKey:     []byte(auth.PrivateKey),
+		})
 	}
-	return storage.SignedURL(bucket, obj, &storage.SignedURLOptions{
-		Method:         "GET",
-		Expires:        time.Now().Add(10 * time.Minute),
-		GoogleAccessID: auth.ClientEmail,
-		PrivateKey:     []byte(auth.PrivateKey),
-	})
+	return af.opener.SignedURL(ctx, fmt.Sprintf("%s%s/%s", linkPrefix, bucket, obj))
```
alvaroaleman, line 206: I think the GCS implementation of this should also be moved into pkio or not? Because of this we have to pass around both an `io.Opener` and a `gcsCredsFile`

prow/spyglass/gcsartifact_fetcher.go
```diff
@@ -152,81 +154,88 @@ const (
...
 	if err != nil {
 		return nil, fmt.Errorf("Failed to get GCS job source from %s: %v", key, err)
 	}
 
 	bucketName, prefix := extractBucketPrefixPair(src.jobPath())
-	bkt := af.client.Bucket(bucketName)
 	objName := path.Join(prefix, artifactName)
-	obj := &gcsArtifactHandle{bkt.Object(objName)}
-	signedURL, err := af.signURL(bucketName, objName)
+	obj := &gcsArtifactHandle{Opener: af.opener, Name: fmt.Sprintf("%s%s/%s", src.linkPrefix, src.bucket, objName)}
```
alvaroaleman, line 237: In a follow-up, we should probably rename the `gcsArtifactHandle` to something neutral as well to make reading the code less confusing

prow/spyglass/gcsartifact_fetcher.go
alvaroaleman: Since that will require adding more options to `pkio.NewOpener` it would probably be nice to use the functional Opts pattern there

prow/cmd/deck/job_history.go
```diff
@@ -160,39 +172,42 @@ func readJSON(bucket storageBucket, key string, data interface{}) error {
...
 	dirs := []string{}
-	it := bucket.Objects(context.Background(), &storage.Query{
-		Prefix:    prefix,
-		Delimiter: "/",
-	})
 	for {
-		attrs, err := it.Next()
-		if err == iterator.Done {
+		done, attrs, err := it.Next(ctx)
+		if done {
```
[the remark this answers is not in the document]
sbueringer: Yup, you're of course right about the error checking. I would be fine with returning an error instead of the additional bool var. The only question is which one. The internally used `*storage.ObjectIterator` returns `iterator.Done`; `*blob.ListIterator` returns `io.EOF`. I would suggest defining a new error in `iterator.go` in the same style as `iterator.Done`:
```
var IteratorDone = errors.New("no more items in iterator")
```

prow/cmd/deck/job_history.go
```diff
@@ -68,15 +68,17 @@ type buildData struct {
...
-	listAll(prefix string) ([]string, error)
-	readObject(key string) ([]byte, error)
+	getStorageProvider() string
+	listSubDirs(ctx context.Context, prefix string) ([]string, error)
+	listAll(ctx context.Context, prefix string) ([]string, error)
+	readObject(ctx context.Context, key string) ([]byte, error)
 }
 
 // gcsBucket is our real implementation of storageBucket
 type gcsBucket struct {
```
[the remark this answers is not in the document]
sbueringer: I started a list here: https://github.com/kubernetes/test-infra/pull/17183#issue-402002610:)

prow/pjutil/pjutil.go
sbueringer: Sounds reasonable. So implement the validation  in `checkConfig` and assume here that the error cannot happen (ignore it)?

prow/pjutil/pjutil_test.go
sbueringer: Would you rather compare the errors by string or define a separate error type and compare via `==` ?

prow/spyglass/gcsartifact_fetcher.go
sbueringer: Also added above: https://github.com/kubernetes/test-infra/pull/17183#issue-402002610
Just didn't rename it already because the diff would be huge :)

prow/spyglass/gcsartifact_fetcher.go
sbueringer: I can move it there. We just had a discussion a few PRs ago where the idea was to not put special GCS handling inside the opener package. But of course we can reconsider now that we see the problem which wasn't obivous at that point :)

prow/plank/controller.go
sbueringer: Depends on outcome of: https://github.com/kubernetes/test-infra/pull/17183#discussion_r413253056

prow/cmd/deck/job_history.go
```diff
@@ -160,39 +172,42 @@ func readJSON(bucket storageBucket, key string, data interface{}) error {
...
-		Prefix: prefix,
-	})
 	for {
-		attrs, err := it.Next()
-		if err == iterator.Done {
+		done, attrs, err := it.Next(ctx)
+		if done {
 			break
 		}
 		if err != nil {
```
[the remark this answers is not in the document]
sbueringer: Done

prow/pjutil/pjutil_test.go
sbueringer: Done

prow/io/opener.go
sbueringer: Done

prow/pjutil/pjutil_test.go
alvaroaleman: I don't really care much but if you go the string route it might be simpler to do define it on the test as string, because then you do not have to do the
```
var expectedErrStr string
if actualErr != nil {
  actualErrStr = actualErr.Error()
}
```

prow/pjutil/pjutil.go
alvaroaleman: Implement it in pjutil, export it, call it in `checkconfig` and still return the error here. Ppl might not use checkconfig, we might implement the validation wrong or incomplete, there is not a guarantee this is correct just because we add a validation

prow/cmd/deck/job_history.go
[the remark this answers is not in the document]
alvaroaleman: I am fine with both `io.EOF` or a custom error type

prow/spyglass/gcsartifact_fetcher.go
alvaroaleman: yeah, basically we have to decide between making the interface a bit leaky (passing in provider-specific options) and making it incomplete. Since the latter means that ppl in the future might think they can use it when they actually can not I think the former is the lesser evil. Additionally, it means we properly separate concerns.

prow/cmd/deck/job_history.go
```diff
@@ -89,22 +91,26 @@ type jobHistoryTemplate struct {
...
+	defer rc.Close()
 	return ioutil.ReadAll(rc)
 }
 
 func (bucket gcsBucket) getName() string {
 	return bucket.name
 }
 
-func readLatestBuild(bucket storageBucket, root string) (int64, error) {
+func (bucket gcsBucket) getStorageProvider() string {
```
alvaroaleman, line 107: Nit: This func seems superfluous? A private getter that just returns a private property?

prow/cmd/deck/job_history.go
```diff
@@ -301,7 +339,7 @@ func getBuildData(bucket storageBucket, dir string) (buildData, error) {
 		b.commitHash = commitHash
 	}
 	finished := gcs.Finished{}
-	err = readJSON(bucket, path.Join(dir, "finished.json"), &finished)
+	err = readJSON(ctx, bucket, path.Join(dir, "finished.json"), &finished)
```
alvaroaleman, line 342: Obviously not your PR but this filename really belongs into a `const` in an api package

sbueringer: Yup. Added it to the list :)
https://github.com/kubernetes/test-infra/pull/17183#issue-402002610

prow/pjutil/pjutil.go
sbueringer: Couldn't add it in pjutil or the config package because of circular dependencies. I'll add it in `prow/apis/prowjobs/v1/types.go`. I don't really have an overview, but if there's a more appropriate place I can move it there.

prow/cmd/deck/job_history.go
sbueringer: I did it this way because otherwise I would have to hand over the struct instead of the interface in `getPRBuildData`

prow/cmd/deck/job_history.go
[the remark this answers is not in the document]
sbueringer: Fixed

prow/pjutil/pjutil_test.go
sbueringer: Fixed

prow/plank/controller.go
sbueringer: Done

prow/spyglass/gcsartifact_fetcher.go
sbueringer: I moved it into the opener package. I think it's better to add the option to `SignedUrl` than to `NewOpener` as it's only used for `SignedUrl`. I'm not sure which options pattern to use. For now I just added `SignedURLOptions` wit a non-pointer flag for `UseGSCookieAuth`.

Just not sure if I should change it because it's inconsistent:
* we used a vararg in the Writer method instead (because we wanted to stay compatible)
* we used pointer vars instead of regular types

But I think when ignoring inconsistencies it's better to implement it this way, because we can avoid using pointers and it would be more complicated (and I think it doesn't make sense) to handle a vararg of `SignedURLOptions`. WDYT?

sbueringer:
@alvaroaleman all findings except tests should be fixed. I start implementing/ extending the tests now

prow/spyglass/gcsartifact_fetcher_test.go
sbueringer: `newGCSJobSource` did only handle the new format with `<storage-provider>://<path>` and was only used like this. But I extended `newGCSJobSource` and the tests to handle old and new format

prow/spyglass/gcsartifact_fetcher_test.go
sbueringer: Same as above. Was only used with scheme, but can handle both now (also extended the tests)

prow/spyglass/gcsartifact_fetcher_test.go
sbueringer: Same as above. Was only used with scheme, but can handle both now (also extended the tests)

prow/spyglass/lenses/common/common.go
```diff
@@ -228,24 +235,30 @@ type ProwJobFetcher interface {
...
 	prefix := config().Plank.GetJobURLPrefix(job.Spec.Refs)
 	if !strings.HasPrefix(url, prefix) {
-		return "", fmt.Errorf("unexpected job URL %q when finding GCS path: expected something starting with %q", url, prefix)
+		return "", "", fmt.Errorf("unexpected job URL %q when finding GCS path: expected something starting with %q", url, prefix)
 	}
-	return url[len(prefix):], nil
-
+	// we parse the storageProvider from the last jobURLPrefix segment, e.g.:
+	// https://prow.k8s.io/view/gcs/ => gcs
+	urlPrefixSegments := strings.Split(strings.TrimSuffix(prefix, "/"), "/")
```
sbueringer, line 256: @alvaroaleman I think we need something better then this as it only works if the jobURLPrefix ends with gcs/gs/s3. Do you have an idea how we can/should detect the storage provider? `ProwToGCS` is used when we handle `prowjob` urls

sbueringer: It it possible to do something like this:
* `prowjobs/<storage-provider>/<path>`
* `prowjobs/<path>`  (fall back to gs)

Do you know where these prowjobs urls come from?

prow/cmd/deck/job_history_test.go
sbueringer: Done

prow/pjutil/pjutil.go
sbueringer: Added test

prow/spyglass/lenses/common/common.go
sbueringer: What do you think about using the job retrieved in l.244 and using the storageProvider discovered from prowjob.Spec.DecorationConfig.GCSConfiguration.Bucket 
(if prefix found => use that one, if not gs)

sbueringer:
@alvaroaleman @fejta can you please take another look?  Thx :)

prow/cmd/deck/job_history_test.go
```diff
@@ -17,36 +17,99 @@ limitations under the License.
...
 import (
+	"context"
 	"net/url"
+	"reflect"
 	"testing"
+	"time"
+
+	"github.com/fsouza/fake-gcs-server/fakestorage"
+
+	"k8s.io/test-infra/prow/io"
```
alvaroaleman, line 28: Nit: group  our own imports together

prow/cmd/deck/job_history_test.go
```diff
@@ -176,3 +242,162 @@ func TestLinkID(t *testing.T) {
...
+	}
+	tests := []struct {
+		name    string
+		args    args
+		want    jobHistoryTemplate
+		wantErr string
+	}{
+		{
+			name: "get job history pr-logs (old format)",
+			args: args{
```
alvaroaleman, line 358: Nit: Having a one property struct does't really add any value?

prow/cmd/pipeline/controller.go
```diff
@@ -343,7 +343,10 @@ func (c *controller) pipelineID(pj prowjobv1.ProwJob) (string, string, error) {
 		return "", "", err
 	}
 	pj.Status.BuildID = id
-	url := pjutil.JobURL(c.config().Plank, pj, logrus.NewEntry(logrus.StandardLogger()))
+	url, err := pjutil.JobURL(c.config().Plank, pj, logrus.NewEntry(logrus.StandardLogger()))
```
alvaroaleman, line 346: Same as in plank I think we should just log the error as its non-recoverable and shouldn't result in blocking processing the job forever

prow/spyglass/gcsartifact_fetcher.go
```diff
@@ -147,86 +148,48 @@ func (af *GCSArtifactFetcher) artifacts(key string) ([]string, error) {
...
-		return "", fmt.Errorf("only service_account GCS auth is supported, got %q", auth.Type)
-	}
-	return storage.SignedURL(bucket, obj, &storage.SignedURLOptions{
-		Method:         "GET",
-		Expires:        time.Now().Add(10 * time.Minute),
-		GoogleAccessID: auth.ClientEmail,
-		PrivateKey:     []byte(auth.PrivateKey),
+func (af *GCSArtifactFetcher) signURL(ctx context.Context, key string) (string, error) {
+	return af.opener.SignedURL(ctx, key, pkgio.SignedURLOptions{
+		UseGSCookieAuth: af.useCookieAuth,
```
alvaroaleman, line 153: Wouldn't it make more sense to pass this in when opening the opener? As far as I can see we either do this always or never or did I miss something?

prow/spyglass/gcsartifact_fetcher_test.go
```diff
@@ -223,21 +326,21 @@ RU5EIFBSSVZBVEUgS0VZLS0tLS1cbgo=`)
 		useCookie bool
 		expected  string
 		contains  []string
-		err       bool
+		err       string
```
alvaroaleman, line 329: Thanks for improving that :)

prow/spyglass/lenses/common/common.go
alvaroaleman: yeah, that sgtm.

prow/apis/prowjobs/v1/types.go
```diff
@@ -607,6 +607,9 @@ func (g *GCSConfiguration) ApplyDefault(def *GCSConfiguration) *GCSConfiguration
 
 // Validate ensures all the values set in the GCSConfiguration are valid.
 func (g *GCSConfiguration) Validate() error {
+	if _, err := ExtractBucketName(g.Bucket); err != nil {
```
alvaroaleman, line 610: `Checkconfig` needs to be extended to excercise this

prow/cmd/deck/job_history_test.go
sbueringer: Done

prow/cmd/deck/job_history_test.go
sbueringer: Done

prow/cmd/pipeline/controller.go
sbueringer: Done

prow/apis/prowjobs/v1/types.go
sbueringer: Done

prow/spyglass/lenses/common/common.go
sbueringer: Done

prow/spyglass/gcsartifact_fetcher.go
sbueringer: Let's continue here: https://github.com/kubernetes/test-infra/pull/17183#discussion_r415277033

prow/spyglass/gcsartifact_fetcher.go
sbueringer: > Wouldn't it make more sense to pass this in when creating the opener? As far as I can see we either do this always or never or did I miss something?

Yup, I can move it the NewOpener. My only concern was that adding UseGSCookieAuth to NewOpener would suggest CookieAuth is used for the GCS client (for reader, writer,..). But I guess this could be fixed by naming it something like UseGSCookieAuthForSignedURLs

WDYT?

sbueringer:
> Some nits, but looks very good overall. Would also love to hear @fejta opinion about the current state

*push* @fejta

prow/apis/prowjobs/v1/types.go
```diff
@@ -621,6 +624,23 @@ func (g *GCSConfiguration) Validate() error {
 	return nil
 }
 
+// ExtractBucketName tries to extract the bucket name from, e.g.:
+// * <bucket-name>
+// * <storage-provider>://<bucket-name>
+func ExtractBucketName(bucket string) (string, error) {
+	var bucketName string
+	bucketSplit := strings.Split(bucket, "://")
```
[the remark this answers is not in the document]
sbueringer: I'll replace it with SplitN for now

prow/apis/prowjobs/v1/types.go
```diff
@@ -621,6 +624,26 @@ func (g *GCSConfiguration) Validate() error {
 	return nil
 }
 
+// ParseBucketName tries to extract the storageProvider and bucket name from, e.g.:
+// * <bucket-name> (storageProvider gs)
+// * <storage-provider>://<bucket-name>
+func ParseBucketName(bucket string) (string, string, error) {
```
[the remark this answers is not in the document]
sbueringer: The problem is that this only works as long as we have the prefix:
```go
	u, _ := url.Parse("gs://prow-artifacts/path/to/object")
	fmt.Printf("scheme: %s, host: %s, path: %s\n", u.Scheme, u.Host, u.Path)
	u, _ = url.Parse("prow-artifacts/path/to/object")
	fmt.Printf("scheme: %s, host: %s, path: %s\n", u.Scheme, u.Host, u.Path)
```
returns:
```
scheme: gs, host: prow-artifacts, path: /path/to/object
scheme: , host: , path: prow-artifacts/path/to/object
```

I can first do something like:
```go
if !strings.Contains(bucket, "://"){
    bucket = "gs://" + bucket
}
```
And then use url.Parse.
WDYT?

prow/cmd/deck/job_history.go
```diff
@@ -115,40 +122,46 @@ func readLatestBuild(bucket storageBucket, root string) (int64, error) {
 	return n, nil
 }
 
-// resolve sym links into the actual log directory for a particular test run
-func (bucket gcsBucket) resolveSymLink(symLink string) (string, error) {
-	data, err := bucket.readObject(symLink)
+// resolve symlinks into the actual log directory for a particular test run, e.g.:
+// * input:  s3://prow-artifacts/pr-logs/pull/cluster-api-provider-openstack/1687/bazel-build/1248207834168954881
```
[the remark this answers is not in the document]
sbueringer: Yup. Just didn't thought about it. I'll change it to GCS examples

prow/spyglass/lenses/common/common.go
```diff
@@ -228,24 +235,42 @@ type ProwJobFetcher interface {
...
 	prefix := config().Plank.GetJobURLPrefix(job.Spec.Refs)
 	if !strings.HasPrefix(url, prefix) {
-		return "", fmt.Errorf("unexpected job URL %q when finding GCS path: expected something starting with %q", url, prefix)
+		return "", "", fmt.Errorf("unexpected job URL %q when finding GCS path: expected something starting with %q", url, prefix)
+	}
+
+	// try to parse storageProvider from DecorationConfig.GCSConfiguration.Bucket
+	// if it doesn't work fallback to URL parsing
+	if job.Spec.DecorationConfig != nil && job.Spec.DecorationConfig.GCSConfiguration != nil {
+		storageProvider, _, err := prowv1.ParseBucketName(job.Spec.DecorationConfig.GCSConfiguration.Bucket)
```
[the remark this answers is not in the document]
sbueringer: In this case it's not about the bucket name itself or behaving differently depending on the bucket name. My only requirement is that I have to be able to build an URL with a storage provider prefix based on a prowjob URL. As far as I know the "prowjob" URLs look like this: `https://prow.k8s.io/view/prowjob/<prowjob-name>/1248318192296660992` (I couldn't find a place where this URLs are built nor did I find any examples of it, but the code looks like it handles these kind of URLs. Are these URLs still supported or documented anywhere?)

So If we still want to be able to handle these URLs (and also not only for GCS), I have to be able to find out the used StorageProvider somehow to decide If I should access GCS or S3.
common/common.go ~l.190:
```go
	case api.ProwKeyType:
		storageProvider, key, err := ProwToGCS(pjFetcher, cfg, key)
		if err != nil {
			logrus.Warningln(err)
		}
		gcsKey = fmt.Sprintf("%s://%s", storageProvider, strings.TrimSuffix(key, "/"))
...
	art, err := gcsArtifactFetcher.Artifact(ctx, gcsKey, name, sizeLimit)
```
`gcsArtifactFetcher` requires a path like `gs://..` or `s3://..` because otherwise we don't have a clue which storage provider to access.

I'm absolutely happy to change this handling if there's an alternative, I just don't know one.

prow/spyglass/spyglass.go
```diff
@@ -148,18 +150,16 @@ func (sg *Spyglass) ResolveSymlink(src string) (string, error) {
 	switch keyType {
 	case prowKeyType:
 		return src, nil // prowjob keys cannot be symlinks.
-	case gcsKeyType:
-		parts := strings.SplitN(key, "/", 2)
-		if len(parts) != 2 {
-			return "", fmt.Errorf("gcs path should look like '{bucket}/{path}', missing at least one.")
+	case gcsKeyType, providers.GS, providers.S3:
```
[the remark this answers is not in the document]
sbueringer: I don't need to handle the individual providers. I just have to differentiate between `/prowjob` and `/<valid-storage-provider`. Imho there's two ways of doing it:
* like currently implemented matching for `prowjob`  and a list of valid storage provider `gcs`/`gs`/`s3` and throwing an error if nothing matched
* match against `prowjob`, if prowjob doesn't match assuming that it's a valid storage provider and handling it in the default case, e.g.:
```go
switch keyType {
	case prowKeyType:
		return src, nil // prowjob keys cannot be symlinks.
	default: //path for all storage providers
        ...
```
So it comes down to validating that it is in fact a valid storage provider and throwing an error if it isn't vs. assuming it's a valid storage provider if it isn't `prowjob`
Both are fine for me. WDYT?

prow/pjutil/pjutil.go
```diff
@@ -271,23 +271,32 @@ func ProwJobFields(pj *prowapi.ProwJob) logrus.Fields {
...
-		return prefix.String()
+
+		_, bucketName, err := prowapi.ParseBucketName(gcsConfig.Bucket)
+		if err != nil {
+			return "", fmt.Errorf("calculating joburl: %w", err)
+		}
+
+		// Final path will be, e.g.:
+		// prefix.Path                   + bucketName         + gcsPath
+		// https://prow.k8s.io/view/gcs/ + kubernetes-jenkins + pr-logs/pull/kubernetes-sigs_cluster-api-provider-openstack/541/pull-cluster-api-provider-openstack-test/1247344427123347459
```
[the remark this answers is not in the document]
sbueringer: I'll adjust the example. It depends on the configuration of `job_url_prefix_config` but both keeping `https://prow.k8s.io/view/gcs` or replacing it with `https://prow.k8s.io/view/gs` will work.

prow/io/opener.go
```diff
@@ -48,13 +55,26 @@ type (
...
+	Size int64
+}
+
 // Opener has methods to read and write paths
 type Opener interface {
 	Reader(ctx context.Context, path string) (ReadCloser, error)
+	RangeReader(ctx context.Context, path string, offset, length int64) (io.ReadCloser, error)
 	Writer(ctx context.Context, path string, opts ...WriterOptions) (WriteCloser, error)
+	Attributes(ctx context.Context, path string) (Attributes, error)
+	SignedURL(ctx context.Context, path string, opts SignedURLOptions) (string, error)
```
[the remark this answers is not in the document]
alvaroaleman: It is an interaction with the underlying storage provider that needs a similar set of information (what provider is it, credentials). If we don't do this, we will have to pass around both an `io.Opener` and cloud provider credentials in Deck.

You might now argue that this leaks through implementation details since it does not apply to block storage, but my opinion is that in the context of Prow, object storage is the first class citizen and block storage is more of a fallback.

prow/spyglass/lenses/common/common.go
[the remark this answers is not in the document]
fejta: So the format is:
`HOST/view/<provider>/<job>/<build>`

where provider is either:
* `prowjob` -- aka access the kubernetes api directly, no storage configured
* `gcs` -- deprecated name for gs
* `other` -- uses the new opener to access it

Since we have GCS turned on for prow.k8s.io we try to never provide the prowjob URLs because they are ephemeral -- aka they stop working as soon as sinker deletes the pod and/or prowjob.

But you can create them:
* Go to deck and find a failing job - https://prow.k8s.io/?state=failure
  - Or any job, but sinker deletes these more slowly
* Click on the magnifying glass to get the job yaml - https://prow.k8s.io/prowjob?prowjob=fd653895-8ffc-11ea-990f-52c6b219fb85
  - note metadata.labels["prow.k8s.io/job"] = ci-benchmark-microbenchmarks
  - note status.build_id = 1258197944759226371
* Now construct that into https://prow.k8s.io/view/prowjob/ci-benchmark-microbenchmarks/1258197944759226371
  - Note this example will probably be broken by the time you read this..

Interestingly also look at status.url:
```yaml
  url: https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-benchmark-microbenchmarks/1258197944759226371
```

So I suspect you can use this value... but maybe it is better to have deck detect that there is a url status field and autoredirect https://prow.k8s.io/view/prowjob/ci-benchmark-microbenchmarks/1258197944759226371 to https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-benchmark-microbenchmarks/1258197944759226371? And otherwise have view/prowjob only support reading information exclusively from what is found in the apiserver (nothing on storage).

<details>
<summary> yaml snippet from example </summary>

```yaml
metadata:
  annotations:
    prow.k8s.io/job: ci-benchmark-microbenchmarks
    testgrid-dashboards: sig-scalability-benchmarks
    testgrid-tab-name: microbenchmarks
  creationTimestamp: "2020-05-07T00:52:13Z"
  generation: 7
  labels:
    created-by-prow: "true"
    prow.k8s.io/job: ci-benchmark-microbenchmarks
    prow.k8s.io/type: periodic
  name: fd653895-8ffc-11ea-990f-52c6b219fb85
  namespace: default
  resourceVersion: "419795230"
  selfLink: /apis/prow.k8s.io/v1/namespaces/default/prowjobs/fd653895-8ffc-11ea-990f-52c6b219fb85
  uid: fd833ffe-8ffc-11ea-94be-42010a8000c3
spec:
  agent: kubernetes
  cluster: default
  decoration_config:
    gcs_configuration:
      bucket: kubernetes-jenkins
      default_org: kubernetes
      default_repo: kubernetes
      path_strategy: legacy
    gcs_credentials_secret: service-account
    grace_period: 15s
    timeout: 2h0m0s
    utility_images:
      clonerefs: gcr.io/k8s-prow/clonerefs:v20200501-e6124e633
      entrypoint: gcr.io/k8s-prow/entrypoint:v20200501-e6124e633
      initupload: gcr.io/k8s-prow/initupload:v20200501-e6124e633
      sidecar: gcr.io/k8s-prow/sidecar:v20200501-e6124e633
  extra_refs:
  - base_ref: master
    org: kubernetes
    path_alias: k8s.io/test-infra
    repo: test-infra
  - base_ref: master
    org: kubernetes
    path_alias: k8s.io/kubernetes
    repo: kubernetes
  job: ci-benchmark-microbenchmarks
  namespace: test-pods
  pod_spec:
    containers:
    - args:
      - --log-file=$(ARTIFACTS)/benchmark-log.txt
      - --output=$(ARTIFACTS)/junit_benchmarks.xml
      - --pass-on-error
      - ../kubernetes/pkg/...
      - ../kubernetes/plugin/...
      - ../kubernetes/vendor/k8s.io/apimachinery/...
      - ../kubernetes/vendor/k8s.io/apiserver/...
      - ../kubernetes/vendor/k8s.io/client-go/...
      command:
      - /benchmarkjunit
      env:
      - name: GOPROXY
        value: https://proxy.golang.org
      image: gcr.io/k8s-testimages/benchmarkjunit:latest
      name: ""
      resources: {}
  report: true
  type: periodic
status:
  build_id: "1258197944759226371"
  completionTime: "2020-05-07T00:54:50Z"
  description: Job failed.
  pendingTime: "2020-05-07T00:52:20Z"
  pod_name: fd653895-8ffc-11ea-990f-52c6b219fb85
  prev_report_states:
    gcsk8sreporter: failure
    gcsreporter: failure
    slackreporter: failure
  startTime: "2020-05-07T00:52:13Z"
  state: failure
  url: https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-benchmark-microbenchmarks/1258197944759226371
```

</summary>

prow/spyglass/spyglass.go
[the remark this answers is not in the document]
fejta: The latter SGTM (special case prowjob, else assume a valid provider) 

Won't we also have to special-case translating gcs to gs (or does opener handle this?)

IMO all new links should be view/gs/foo but there are going to be a lot of view/gcs/foo links around for awhile so we'll need to continue loading those (a redirect to view/gs might be a cool way to handle this)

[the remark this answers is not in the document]
sbueringer: Okay so I'll implement the latter.
Yup we will have to handle GCS. But I think it's okay (and already implemented). Opener currently does not handle gcs:// paths. I think we should keep it that way, especially as `gcs` is only really used in deck/spyglass

Starting with this PR all generated links will be `view/gs/foo`. We could implement a redirect, but I would like to keep it as is and just handle both `view/gcs` and `view/gs`. Not sure if it leads to problems elsewhere and it's probably less risk to starting handling the new paths and still supporting the old ones excatly the same way

prow/spyglass/lenses/common/common.go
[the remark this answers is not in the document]
sbueringer: Thx for these infos :)
I'll start with writing a unit test for this based on your example (and I'll run deck locally with the config for prow.k8s.io to double check this). Let's see what the exact input/outputs of these func are and then what's the best implementation variant is

prow/apis/prowjobs/v1/types.go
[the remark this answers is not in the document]
sbueringer: Fixed

prow/spyglass/spyglass.go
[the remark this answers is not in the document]
sbueringer: Fixed

prow/spyglass/lenses/common/common.go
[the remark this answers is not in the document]
sbueringer: Hmm okay so independent if I add a redirect or read directly from storage on the request to e.g. https://prow.k8s.io/view/prowjob/ci-benchmark-microbenchmarks/1258197944759226371 the problem remains the same that I have to either:
* redirect to https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-benchmark-microbenchmarks/1258197944759226371? or
* read from gs://kubernetes-jenkins/logs/ci-benchmark-microbenchmarks/1258197944759226371?

In both cases I have to find out which storageProvider should be:
* added to the URL or 
* used to read from storage

We can get this info from the ProwJob or the configured `job_url_prefix_config`, thus we have the options
1. get it from the ProwJob `status.url` or
2. get it from the ProwJob `DecorationConfig`
3. parse it from the last segment of the `job_url_prefix_config`

Currently I first try 2. and if this doesn't work 3. The problem with 1. is how to detect the correct URL segment for the storageProvider (`https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-benchmark-microbenchmarks/1258197944759226371`). I'm not sure if there is a way to detect the storageProvider correctly from only the `status.url`

P.S. I pushed the unit test for `ProwToGCS`
P.S.2 I just saw that as of now as long as `job_url_prefix_config` has to end with the storageProvider (`gcs`, `gs` or `s3`). Thus, one Prow instance won't be able to handle s3 and gs at the same time. I just fear if I also start to introduce changes to the `job_url_prefix_config` in this PR, it will make it a lot more likely to break something. But I think we need a follow-up PR with a refactoring which makes it possible to configure `job_url_prefix_config` with e.g. `https://prow.k8s.io/view/` and deprecate the format with storageProvider (`https://prow.k8s.io/view/gcs/`)

prow/cmd/deck/job_history.go
```diff
@@ -240,16 +256,39 @@ func (bucket gcsBucket) listBuildIDs(root string) ([]int64, error) {
...
 
-func parseJobHistURL(url *url.URL) (bucketName, root string, buildID int64, err error) {
+// parseJobHistURL parses the job History URL
+// example urls:
+// * new format: https://prow.k8s.io/job-history/gs/kubernetes-jenkins/pr-logs/directory/pull-capi?buildId=1245584383100850177
+// * old format: https://prow.k8s.io/job-history/kubernetes-jenkins/pr-logs/directory/pull-capi?buildId=1245584383100850177
+// Newly generated URLs will include the storageProvider. We still support old URLs so they don't break.
+// For old URLs we assume that the storageProvider is `gs`.
+// examples return values:
+// * storageProvider: s3, gs
```
fejta, line 266: Wouldn't it make sense to sort these alphabetically?

prow/cmd/deck/job_history.go
```diff
@@ -240,16 +256,39 @@ func (bucket gcsBucket) listBuildIDs(root string) ([]int64, error) {
...
+// * root: pr-logs/directory/pull-capi
+// * buildID: 1245584383100850177
+func parseJobHistURL(url *url.URL) (storageProvider, bucketName, root string, buildID int64, err error) {
 	buildID = emptyID
 	p := strings.TrimPrefix(url.Path, "/job-history/")
-	s := strings.SplitN(p, "/", 2)
-	if len(s) < 2 {
-		err = fmt.Errorf("invalid path (expected /job-history/<gcs-path>): %v", url.Path)
+	// examples for p:
+	// * new format: gs/prow-artifacts/pr-logs/directory/bazel-build-test
```
fejta, line 274: shouldn't this be `kubernetes-jenkins` not `prow-artifacts`? I'd be fine with changing both to `prow-artifacts`

prow/cmd/deck/job_history.go
```diff
@@ -240,16 +256,39 @@ func (bucket gcsBucket) listBuildIDs(root string) ([]int64, error) {
...
+	// * old format: kubernetes-jenkins/pr-logs/directory/pull-cluster-api-provider-openstack-test
+
+	// inject gs/ if old format is used
+	if !strings.HasPrefix(p, providers.GS+"/") && !strings.HasPrefix(p, providers.S3+"/") {
```
fejta, line 278: Instead of doing this can we make the URL:

`gs://prow-aritfacts/pr-logs/directory`

rather than just `gs/prow-artifacts/pr-logs/directory`

and if there is no scheme in the path then add gs://`?

https://golang.org/pkg/net/url/#URL
```golang
package main

import (
	"fmt"
	"log"
	"net/url"
)

func main() {
	u, err := url.Parse("http://bing.com/search?q=dotnet")
	if err != nil {
		log.Fatal(err)
	}
	u.Scheme = "https"
	u.Host = "google.com"
	q := u.Query()
	q.Set("q", "golang")
	u.RawQuery = q.Encode()
	u.Path = "gs://hello/world"
	fmt.Println(u) // https://google.com/gs://hello/world?q=golang
}
```

prow/apis/prowjobs/v1/types.go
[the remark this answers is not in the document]
fejta: `ParsePath`? This isn't just the bucket.

Also, how about just returning the URL? And you could even

```golang
type ProwPath url.URL
func (pp ProwPath) Provider() string {
  return pp.Scheme
}

func (pp ProwPath) Bucket() string {
  return pp.Host
}

func (pp ProwPath) FullPath() string {
  return pp.Host + "/" + pp.Path
}
```

prow/apis/prowjobs/v1/types.go
```diff
@@ -621,6 +625,21 @@ func (g *GCSConfiguration) Validate() error {
...
+	// default to GCS if no storage-provider is specified
+	if !strings.Contains(bucket, "://") {
+		bucket = "gs://" + bucket
+	}
+	parsedBucket, err := url.Parse(bucket)
+	if err != nil {
+		return "", "", fmt.Errorf("bucket field %q has invalid format, expected either <bucket-name> or <storage-provider>://<bucket-name>", bucket)
+	}
+	return parsedBucket.Scheme, parsedBucket.Host + parsedBucket.Path, nil
```
fejta, line 640: Shouldn't this be `Host + "/" + Path`? Are we unit testing this?

prow/cmd/deck/job_history.go
sbueringer: Done

prow/cmd/deck/job_history.go
sbueringer: I changed it to `kubernetes-jenkins` A lot of other examples are already using `kubernetes-jenkins`

prow/cmd/deck/job_history.go
sbueringer: I'm not sure if I got it right. Do you mean changing the external URL (the one used by the browser) to something like: `https://prow.k8s.io/job-history/gs://kubernetes-jenkins/pr-logs/directory/pull-capi?buildId=1245584383100850177`

If we want to do that we have to be really careful as every path.Join we're doing anywhere will break it by changing "gs://" to "gs:/"

prow/apis/prowjobs/v1/types.go
sbueringer: Yup that's unit tested in `TestParseBucket`. Looks like Path contains the leading slash, e.g.:
```
gs://my-floppy-backup/a://doom2.wad.006
Scheme: gs
Host: my-floppy-backup
Path: /a://doom2.wad.006
```

prow/apis/prowjobs/v1/types.go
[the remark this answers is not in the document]
sbueringer: Yup, done. The func was mainly named like this because it was used to parse the bucket field. But yes it can be also used in other contexts

prow/cmd/deck/job_history.go
fejta: I mostly just don't like that we're special casing providers here and trying to think of an alternative.

How about if we whitelist the set of buckets that we will redirect? AKA
`--default-logs-provider=gs --migrate-bucket=kubernetes-jenkins --migrate-bucket=other-bucket` (`flagutils.Strings` will support flag repetition)

And so then if `p` starts with `kubernetes-jenkins/` or `other-bucket/` we'll replace it with `gs/kubernetes-jenkins` as appropriate.

Having a temporary `--require-provider=gs` would also work (if p doesn't start with gs then insert this at the front, and log that this support will go away by X and they need to set --default-logs-provider=gs and --migrate-bucket)

sbueringer: If possible, it would be nice to avoid additional flags for this.

We have to differentiate between:
*  `<storage-provider>/<bucket>/<path>`
*  `<bucket>/<path>`

There are two ways of doing this:
* detect that the string starts with a valid storage provider
* detect that the string starts with a valid bucket (which could be a lot of different ones)

I guess we still have to handle the old format for a while as we don't want to risk broken links.
The main problem of the current solution is (I guess) that we have to enumerate all valid `storageProviders` and if we add a new one it also has to be added here.

We are already exposing the constants `providers.S3` and `providers.GS` in the `providers` package. What do you think about adding a function like `providers.HasStorageProviderPrefix` to the `providers` package?

I think that's considering the tradeoffs the best solution.

prow/apis/prowjobs/v1/types.go
```diff
@@ -625,19 +625,33 @@ func (g *GCSConfiguration) Validate() error {
...
+
+func (pp ProwPath) FullPath() string {
+	return pp.Host + pp.Path
+}
+
+// ParsePath tries to extract the ProwPath from, e.g.:
 // * <bucket-name> (storageProvider gs)
 // * <storage-provider>://<bucket-name>
-func ParseBucketName(bucket string) (string, string, error) {
+func ParsePath(bucket string) (ProwPath, error) {
```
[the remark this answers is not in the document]
sbueringer: Fixed

prow/cmd/deck/job_history.go
fejta: That sounds perfectly reasonable, yes.

Can you either make that change or else squash your commits (or both)? Then LGTM from me.

sbueringer: Done. I pushed the commit invidually. I'll squash it all now

sbueringer:
@alvaroaleman @fejta Hope I got it all :). Also squashed it to one commit.

alvaroaleman:
looks great, thanks for all your work on this!
/approve
/hold cancel

alvaroaleman:
Needs a rebase:
```
/b/f/w/prow/cmd/checkconfig/main.go:819:52: invalid operation: presubmit.JobBase.Agent == string("k8s.io/test-infra/prow/apis/prowjobs/v1".KubernetesAgent) && presubmit.JobBase.UtilityConfig.Decorate (mismatched types bool and *bool)
/b/f/w/prow/cmd/checkconfig/main.go:827:53: invalid operation: postsubmit.JobBase.Agent == string("k8s.io/test-infra/prow/apis/prowjobs/v1".KubernetesAgent) && postsubmit.JobBase.UtilityConfig.Decorate (mismatched types bool and *bool)
/b/f/w/prow/cmd/checkconfig/main.go:835:51: invalid operation: periodic.JobBase.Agent == string("k8s.io/test-infra/prow/apis/prowjobs/v1".KubernetesAgent) && periodic.JobBase.UtilityConfig.Decorate (mismatched types bool and *bool)
```

/lgtm cancel

sbueringer:
@alvaroaleman fixed, ptal :)

sbueringer:
@alvaroaleman  @fejta  

Thx for the reviews :) I know it's also a lot of work especially in components like Deck where it's easy to break something.

Issue #12127, opened 2019-04-09
Add testgrid for Kubernetes deployable build on K8s/cloud-provider-gcp


cheftako:
**What would you like to be added**:
Testgrid integration for  K8s/cloud-provider-gcp.
We added support for a Kubernetes deployable to GCE build with https://github.com/kubernetes/cloud-provider-gcp/pull/75.

**Why is this needed**:
We need testgrid integration as part of the cloud-provider extraction effort.
The cloud-provider-gcp build is intended to replace the in-tree GCE build.
We need an equivalent signal as to everything being working.

amwat:
https://testgrid.k8s.io/provider-gcp-presubmits#cloud-provider-gcp-tests

PR #17184, opened 2020-04-10
Add testgrid for k/cp-gcp presubmits.

amwat:
fixes https://github.com/kubernetes/test-infra/issues/12127

/cc @cheftako @BenTheElder @spiffxp

### DIFF
--- a/config/jobs/kubernetes/cloud-provider-gcp/cloud-provider-gcp-presubmits.yaml
+++ b/config/jobs/kubernetes/cloud-provider-gcp/cloud-provider-gcp-presubmits.yaml
@@ -4,6 +4,10 @@ presubmits:
     max_concurrency: 5
     always_run: true
     decorate: true
+    annotations:
+      testgrid-dashboards: provider-gcp-presubmits
+      description: Build and unit test for kubernetes/cloud-provider-gcp.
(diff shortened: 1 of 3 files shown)


BenTheElder:
failing tests

amwat:
```
--- FAIL: TestKubernetesProwInstanceJobsMustHaveMatchingTestgridEntries (4.48s)
    config_test.go:444: Testgrid group cloud-provider-gcp-tests does not have a matching jenkins or prow job
FAIL
```

hmm seems like some chicken and egg problem here? considering i'm adding the testgrid annotations **to the prowjob** there definitely exists a matching prowjob. 

Do we need to separately merge the changes to the testgrid config first? 
/cc @chases2

chases2:
I've run into this issue once before, and it was because `kubernetes/cloud-provider-gcp` is in the [`noPresubmitsInTestgridPrefixes`](https://github.com/kubernetes/test-infra/blob/9beef71a5dacf41b15fa1957b545922876ceb6a5/config/tests/testgrids/config_test.go#L349-L372) exemption group.

Try removing that group from the exemption list and adding similar annotations as needed. I think that will resolve this issue.

As a resulting issue, I'd like to fix that error message. The error is right, but the message isn't.

PR #17191, opened 2020-04-13
Fix cluster-api-provider-aws pull-cluster-api-provider-aws-make-conformance timeout

detiber:
Fixes a misspelling of `timeout` that was preventing the default 2h job timeout from being overriden properly.

/assign @ncdc


ncdc:
😢 
/lgtm

PR #17201, opened 2020-04-14
Support for S3 default credential provider chain

ramirezag:
This PR enables default credential provider chain using the fallback gocloud default discovery. Thanks to @sbueringer for the direction.

This has been tested with AWS S3 and the changes done are:
- Removed checking of GCS or S3 credentials in DecorationConfig validation 
- Fixes wrong bucket path passed to [opener.Writer](https://github.com/kubernetes/test-infra/blob/master/prow/pod-utils/gcs/upload.go#L182) when `gcs_configuration.bucket` contains query parameter
- Fixes query parameter not passed in provider [blob.OpenBucket](https://github.com/kubernetes/test-infra/blob/master/prow/io/providers/providers.go#L71)

/cc @sbueringer

### DIFF
--- a/prow/io/providers/providers.go
+++ b/prow/io/providers/providers.go
@@ -67,8 +67,12 @@ func GetBucket(ctx context.Context, s3Credentials []byte, path string) (*blob.Bu
 	if storageProvider == providerS3 && len(s3Credentials) > 0 {
 		return getS3Bucket(ctx, s3Credentials, bucket)
 	}
-
-	bkt, err := blob.OpenBucket(ctx, fmt.Sprintf("%s://%s", storageProvider, bucket))
+	urlstr := fmt.Sprintf("%s://%s", storageProvider, bucket)
+	qmarkIdx := strings.Index(path, "?")
+	if qmarkIdx >= 0 {
+		urlstr += path[qmarkIdx:]
+	}
+	bkt, err := blob.OpenBucket(ctx, urlstr)
 	if err != nil {
 		return nil, fmt.Errorf("error opening file bucket: %v", err)
 	}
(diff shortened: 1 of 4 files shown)


ramirezag:
Added hold label. I just found out that the `pr-logs/directory/build-and-test/*.txt` contains something like `s3://mybucket?region=eu-central-1/pr-logs/pull/org_repo/51/build-and-test/1250059586593886208` instead of `s3://mybucket/pr-logs/pull/org_repo/51/build-and-test/1250059586593886208`

prow/io/providers/providers.go
```diff
@@ -67,8 +67,12 @@ func GetBucket(ctx context.Context, s3Credentials []byte, path string) (*blob.Bu
 	if storageProvider == providerS3 && len(s3Credentials) > 0 {
```
sbueringer, line 67: Might be the better solution to implement your original idea, configuring s3 via config file and only configure access key and secret key when set in the json file.

ramirezag: Agree! Simple and doesn't involve multiple files :). Should I proceed?

sbueringer: I think so. @alvaroaleman wdyt?

alvaroaleman: yeah, sounds much more elegant

sbueringer:
@ramirezag @alvaroaleman I think this PR  can be closed as afaik it's been implemented here: https://github.com/kubernetes/test-infra/pull/17642

alvaroaleman:
Yeah this isn't needed anymore thanks to #17642
/close

Issue #10065, opened 2018-11-07
[prow] Add support for github teams in OWNERS files


jtnord:
as per title,  if a team has a large amount of repositories it is  cumbersome to update all the `OWNERS` files when the membership of the team changes.

https://github.com/kubernetes/test-infra/blame/master/prow/plugins/approve/approvers/README.md#L36 points out erroneously that teams are not audited - but that is not the case as [documented](https://help.github.com/articles/reviewing-the-audit-log-for-your-organization/)

So it would be great if prow added support for teams in the owners file.

cjwagner:
How would we would distinguish OWNERS_ALIASES from GitHub teams?
Using GitHub teams instead of OWNERS_ALIASES would dramatically increase API token usage for all plugins that use OWNERS files unless the github cache is deployed (which is something we cannot assume) so I'd be hesitant to accept this change unless we can demonstrate that that is not a concern.
Even if this is feasible I won't have time to work on this Q4 so I'm unassigning myself.
/unassign

jtnord:
> how would we would distinguish OWNERS_ALIASES from GitHub teams

I may be wrong but today aliased and users have the same form in the `OWNERS` file ( `- string`)

```
approvers:
  - alice
  - bob     # this is a comment
reviewers:
  - alice
  - carol   # this is another comment
  - sig-foo # this is an alias
```

There is no difference here between `sig-foo` and `alice` for example.

A team in github is referenced in the form `org/team` so it can be distinguished by virtue of having a slash.  Now I do not find a definitive spec on what is or is not allowed in an alias so it could possibly contain arbitrary characters including the `slash` but then I would turn the table and ask how you distinguish today between users and aliases?  Isn't that the exact same problem that is already solved?
either the string is an alias or it is not.  As users can not contain a `/` it logically follows that if the entry is not an alias and contains a slash it should be treated as a team.

The API token usage is a valid concern but given this would be opt in (no one is making anyone use teams, and you can tell up front if the thing smells like a team before even calling the API) it should be possible to implement with no impact on anyone not using teams in an `OWNERS`.

BenTheElder:
>  it should be possible to implement with no impact on anyone not using teams in an OWNERS.

Only if the feature were disable-able. If one repo uses this, the token usage affects everyone on that Prow deployment.

I also dislike this a little bit, because other than `hook`, prow nominally supports other git hosts (mainly gerrit atm) and teams is a github thing, whereas owners and aliases are quite portable.

[...]

PR #17204, opened 2020-04-14
releng: Update OWNERS

justaugustus:
- Remove @aleksandra-malinowska from Patch Release Team (ref: https://github.com/kubernetes/sig-release/issues/987)
- Promote @hasheddan to Branch Manager (ref: https://github.com/kubernetes/sig-release/issues/1041)
- Use `OWNER_ALIASES` for Release Engineering reviewers/approvers

/assign @spiffxp @Katharine @fejta 

### DIFF
--- a/config/jobs/kubernetes/release/OWNERS
+++ b/config/jobs/kubernetes/release/OWNERS
@@ -6,7 +6,6 @@ approvers:
 - tpepper
 
 reviewers:
-- aleksandra-malinowska # Patch Release Team
 - cpanato # Branch Manager
 - dougm # Patch Release Team
 - feiskyer # Patch Release Team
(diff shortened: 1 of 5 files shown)


config/jobs/kubernetes/release/OWNERS
```diff
@@ -1,18 +1,10 @@
 # See the OWNERS docs at https://go.k8s.io/owners
 
 approvers:
-- calebamiles
-- justaugustus
-- tpepper
+- release-engineering-approvers
```
fejta, line 4: I wish we had a better way of displaying who these members were to people unfamiliar with the system.

spiffxp: Agree, I feel like natively supporting GitHub Teams (ref: https://github.com/kubernetes/test-infra/issues/10065) would be ideal, but this is the system we have in place today.

PR #17205, opened 2020-04-14
ghproxy: add token hash label to cache response metric

stevekuznetsov:
Without this label on the metric it's not possible to filter cache
responses by the user making the requests.

### DIFF
--- a/ghproxy/ghcache/ghcache.go
+++ b/ghproxy/ghcache/ghcache.go
@@ -158,20 +158,7 @@ type upstreamTransport struct {
 
 func (u upstreamTransport) RoundTrip(req *http.Request) (*http.Response, error) {
 	etag := req.Header.Get("if-none-match")
-
-	// get authorization header to convert to sha256
-	authHeader := req.Header.Get("Authorization")
-	if authHeader == "" {
-		logrus.Warn("Couldn't retrieve 'Authorization' header, adding to unknown bucket")
-		authHeader = "unknown"
-	}
-	logrus.WithFields(logrus.Fields{
-		"words":      len(strings.Split(authHeader, " ")),
-		"length":     len(authHeader),
-		"user-agent": req.Header.Get("User-Agent"),
-		"source-ip":  req.RemoteAddr,
-	}).Debug("Hashing auth header.")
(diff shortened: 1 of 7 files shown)


ghproxy/ghcache/ghcache.go
```diff
@@ -206,6 +193,16 @@ func (u upstreamTransport) RoundTrip(req *http.Request) (*http.Response, error)
...
 }
 
+func authHeaderHash(req *http.Request) string {
+	// get authorization header to convert to sha256
+	authHeader := req.Header.Get("Authorization")
+	if authHeader == "" {
+		logrus.Warn("Couldn't retrieve 'Authorization' header, adding to unknown bucket")
+		authHeader = "unknown"
+	}
+	return fmt.Sprintf("%x", sha256.Sum256([]byte(authHeader))) // use %x to make this a utf-8 string for use as a label
```
fejta, line 203: Can we choose a computationally less expensive hash?

cjwagner: This does need to be a cryptographic hash. Is SHA256 considered slow?
We could also keep a simple cache of hash values.

stevekuznetsov: Happy to follow up with a simple in-memory cache

fejta: I do not agree that this needs to be a cryptographic hash. It needs to be a one-way function that has a low chance of collisions. If we intend to perform it on every request, it must be computationally cheap.

fejta: Although it looks like we are already using sha256? In which case this isn't a regression so is less important to fix (but should be fixed, sha256 is almost certainly a poorer choice for this than say md5 given the need)

stevekuznetsov: I think the performance hit of this is miniscule but we should see when this lands ... I have added a simple in-memory cache. Isn't md5 cryptographic (but broken?)

stevekuznetsov:
Finally done wrestling with `bazel` ... @cjwagner @fejta @alvaroaleman

ghproxy/ghcache/ghcache.go
fejta: There are scenarios where md5 is unsuitable, but it isn't obvious to me how this is one of them

PR #17211, opened 2020-04-14
Deprecate plank reporting.

fejta:
Crier now owns this behavior.
* Delete plank's reporting logic
* Make plank spawn a crier github reporter (if necessary).
  - Warn if this is happening.
* DRY the logic for creating the github reporter controller
  - So plank and crier can both use it.
* Remove plank reporting unit tests.

ref https://github.com/GoogleCloudPlatform/oss-test-infra/issues/181

### DIFF
--- a/prow/cmd/plank/main.go
+++ b/prow/cmd/plank/main.go
@@ -17,6 +17,7 @@ limitations under the License.
 package main
 
 import (
+	"context"
 	"flag"
 	"fmt"
 	"os"
@@ -28,7 +29,6 @@ import (
 
 	"k8s.io/test-infra/pkg/flagutil"
 	"k8s.io/test-infra/prow/config"
-	"k8s.io/test-infra/prow/config/secret"
 	prowflagutil "k8s.io/test-infra/prow/flagutil"
 	"k8s.io/test-infra/prow/interrupts"
 	"k8s.io/test-infra/prow/logrusutil"
@@ -48,7 +48,7 @@ type options struct {
 
 	dryRun     bool
 	kubernetes prowflagutil.KubernetesOptions
-	github     prowflagutil.GitHubOptions
+	github     prowflagutil.GitHubOptions // TODO(fejta): remove
 }
 
 func gatherOptions(fs *flag.FlagSet, args ...string) options {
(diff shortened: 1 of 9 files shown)


prow/cmd/plank/main.go
```diff
@@ -96,22 +98,28 @@ func main() {
...
-	githubClient, err := o.github.GitHubClient(secretAgent, o.dryRun)
-	if err != nil {
-		logrus.WithError(err).Fatal("Error getting GitHub client.")
+		const resync = 0 * time.Minute
+		prowjobInformerFactory := prowjobinformer.NewSharedInformerFactoryWithOptions(prowjobClientset, resync, prowjobinformer.WithNamespace(cfg().ProwJobNamespace))
+		controller, err := githubreporter.Controller(o.github, o.dryRun, cfg, "", prowjobClientset, prowjobInformerFactory.Prow().V1().ProwJobs(), 1)
+		if err != nil {
+			logrus.WithError(err).Fatal("Error creating github reporter")
+		}
+		reporter = controller.Run
```
alvaroaleman, line 121: Why? We can just leave it as-is and remove all of it in July.

This change means plank ends up having two caches for Prowjobs if it has reporting enabled.

fejta: Can't it share a cache? And it isn't super relevant as people shouldn't be doing this. 

This allows us to simplify the code now.

alvaroaleman: It can not, unless you change crier to be a controller-runtime based controller (which would be a good idea anyways, as it would get rid of all the boilerplate needed for manually written controllers)

fejta: Ack

fejta:
Refactored this such that it makes no changes in crier. Now just removes the plank reporting and ensures crier's github reporter is started if requested (and warns to migrate)

prow/cmd/plank/reporter.go
```diff
@@ -0,0 +1,62 @@
...
+	prowjobClientset, err := kubernetes.ProwJobClientset(cfg().ProwJobNamespace, dryRun)
+	if err != nil {
+		return nil, fmt.Errorf("prow client: %w", err)
+	}
+
+	const resync = 0
+	prowjobInformerFactory := prowjobinformer.NewSharedInformerFactoryWithOptions(prowjobClientset, resync, prowjobinformer.WithNamespace(cfg().ProwJobNamespace))
+	informer := prowjobInformerFactory.Prow().V1().ProwJobs()
+	githubReporter := githubreporter.NewReporter(githubClient, cfg, prowapi.ProwJobAgent(""))
+	controller := crier.NewController(
```
alvaroaleman, line 55: Didn't you at some point heavily oppose linebreaking for func arguments?

fejta: lol, not sure. Want me to switch it around?

alvaroaleman: No, I like it this way, that's why I am surprised. The discussion we had was for the case where the function calls return is captured: https://github.com/kubernetes/test-infra/pull/15078#discussion_r344463949

fejta: Ah, my beef there was `if err := foo()` and `; err != nil {` being on separate lines, not putting the arguments on separate lines


```golang
// fine
if bar, err := foo(); err != nil {
  return fmt.Errorf("food: %w", err)
}

// fine
something_interesting(
  with,
  lots,
  of,
  arguments)

// less excited about this
if this, that, other, err := something_interesting(
  with,
  lots,
  of,
  arguments); err != nil {
  return fmt.Errorf("less exciting: %w", err)
}
```

PR #17213, opened 2020-04-15
jobs: remove jobs that use --gke-environment=test

spiffxp:
these shouldn't be running on prow.k8s.io


spiffxp:
/hold
for requested reviewers

liggitt:
lgtm but will defer to @sebastienvas

PR #17214, opened 2020-04-15
[Scalability][Containerd] Add containerd preset

jprzychodzen:
Also:
* Migrate from LOG_DUMP_SYSTEMD_SERVICES to LOG_DUMP_SAVE_SERVICES as it covers both ssh and logexporter logdump

/sig scalability
/assign @wojtek-t

### DIFF
--- a/config/jobs/kubernetes/sig-scalability/sig-scalability-experimental-periodic-jobs.yaml
+++ b/config/jobs/kubernetes/sig-scalability/sig-scalability-experimental-periodic-jobs.yaml
@@ -236,6 +236,7 @@ periodics:
     preset-service-account: "true"
     preset-k8s-ssh: "true"
     preset-e2e-scalability-common: "true"
+    preset-e2e-scalability-containerd: "true"
   annotations:
     testgrid-dashboards: sig-scalability-experiments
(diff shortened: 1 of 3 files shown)


config/jobs/kubernetes/sig-scalability/sig-scalability-experimental-periodic-jobs.yaml
```diff
@@ -250,10 +251,8 @@ periodics:
           - --scenario=kubernetes_e2e
           - --
           - --check-leaked-resources
-          # TODO(oxddr): remove once debugging is finished
-          - --env=KUBEPROXY_TEST_LOG_LEVEL=--v=4
           - --cluster=e2e-big-load-containerd
-          - --env=KUBE_CONTAINER_RUNTIME=containerd
+          - --env=KUBEPROXY_TEST_LOG_LEVEL=--v=4
```
wojtek-t, line 255: nit - revert the TODO

jprzychodzen: Done

PR #3401, opened 2017-07-08
Add support for logexporter in our setup & config scripts

shyamjvs:
Ref issue https://github.com/kubernetes/kubernetes/issues/48513

This complements changes on the k8s side.
Need to wait for https://github.com/kubernetes/kubernetes/pull/48651 to get merged before merging this one.

/cc @kubernetes/sig-scalability-misc @kubernetes/test-infra-maintainers @fejta @wojtek-t @gmarek

PR #17215, opened 2020-04-15
Add GCS_ARTIFACTS_DIR env var handling in pod-utils

tosi3k:
Handling of `GCS_ARTIFACTS_DIR` environment variable is required prior porting SIG Scalability tests to pod-utils. It was previously handled by _test-infra/jenkins/bootstrap.py_ and it's required to properly export logs using logexporter from cluster nodes after tests execution.

This issue explains the reason behind using logexporter for SIG Scalability tests: [kubernetes/issues/48513](https://github.com/kubernetes/kubernetes/issues/48513) and #3401 was the way it was solved.

/sig scalability

### DIFF
--- a/prow/pod-utils/decorate/podspec.go
+++ b/prow/pod-utils/decorate/podspec.go
@@ -18,6 +18,7 @@ package decorate
 
 import (
 	"fmt"
+	"net/url"
 	"path"
 	"path/filepath"
 	"sort"
@@ -52,6 +53,7 @@ const (
 	gopathEnv               = "GOPATH"
 	toolsMountName          = "tools"
 	toolsMountPath          = "/tools"
+	gcsArtifactsDirEnv      = "GCS_ARTIFACTS_DIR"
 	gcsCredentialsMountName = "gcs-credentials"
 	gcsCredentialsMountPath = "/secrets/gcs"
 	s3CredentialsMountName  = "s3-credentials"
@@ -593,6 +595,22 @@ func decorate(spec *coreapi.PodSpec, pj *prowapi.ProwJob, rawEnv map[string]stri
 	// TODO(fejta): we should pass around volume names rather than forcing particular mount paths.
 
 	rawEnv[artifactsEnv] = artifactsPath
+
+	if gcsConfig := pj.Spec.DecorationConfig.GCSConfiguration; gcsConfig != nil {
+		jobSpec := downwardapi.NewJobSpec(pj.Spec, rawEnv[downwardapi.BuildIDEnv], pj.Name)
+		_, gcsPath, _ := gcsupload.PathsForJob(gcsConfig, &jobSpec, "")
+
+		u, err := url.Parse(gcsConfig.Bucket)
+		if err != nil {
+			return fmt.Errorf("could not parse the GCS bucket name %s: %v", gcsConfig.Bucket, err)
+		}
+		if u.Scheme == "" {
+			u.Scheme = "gs"
+		}
+		u.Path = path.Join(u.Path, gcsPath, "artifacts")
+		rawEnv[gcsArtifactsDirEnv] = u.String()
+	}
+
 	rawEnv[gopathEnv] = codeMountPath // TODO(fejta): remove this once we can assume go modules
 	logMount := coreapi.VolumeMount{
 		Name:      logMountName,
--- a/prow/pod-utils/decorate/podspec_test.go
+++ b/prow/pod-utils/decorate/podspec_test.go
@@ -665,6 +665,7 @@ func TestProwJobToPod(t *testing.T) {
 								{Name: "BUILD_ID", Value: "blabla"},
 								{Name: "BUILD_NUMBER", Value: "blabla"},
 								{Name: "CI", Value: "true"},
+								{Name: "GCS_ARTIFACTS_DIR", Value: "gs://my-bucket/pr-logs/pull/org-name_repo-name/1/job-name/blabla/artifacts"},
 								{Name: "GOPATH", Value: "/home/prow/go"},
 								{Name: "JOB_NAME", Value: "job-name"},
 								{Name: "JOB_SPEC", Value: `{"type":"presubmit","job":"job-name","buildid":"blabla","prowjobid":"pod","refs":{"org":"org-name","repo":"repo-name","base_ref":"base-ref","base_sha":"base-sha","pulls":[{"number":1,"author":"author-name","sha":"pull-sha"}],"path_alias":"somewhere/else"}}`},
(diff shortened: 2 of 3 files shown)


stevekuznetsov:
Why can't the job place the artifacts it needs to into the appropriate place for the Prow infrastructure to upload them? I don't think exposing this to jobs is a good separation of concerns.

jkaniuk:
logexporter upload logs directly from each node, to vastly improve time of such operation.
Please check an original issue https://github.com/kubernetes/kubernetes/issues/48513

prow/pod-utils/decorate/podspec.go
```diff
@@ -593,6 +594,15 @@ func decorate(spec *coreapi.PodSpec, pj *prowapi.ProwJob, rawEnv map[string]stri
 	// TODO(fejta): we should pass around volume names rather than forcing particular mount paths.
 
 	rawEnv[artifactsEnv] = artifactsPath
+	gcsConfig := pj.Spec.DecorationConfig.GCSConfiguration
+	rawEnv[gcsArtifactsDirEnv] = path.Join(
+		gcsConfig.Prefix,
```
jkaniuk, line 599: I do not see Prefix here:
https://github.com/kubernetes/test-infra/blob/master/prow/apis/prowjobs/v1/types.go#L535

prow/pod-utils/decorate/podspec.go
```diff
@@ -593,6 +594,15 @@ func decorate(spec *coreapi.PodSpec, pj *prowapi.ProwJob, rawEnv map[string]stri
...
+		gcsConfig.Bucket,
+		"logs",
```
jkaniuk, line 601: For e.g. presubmits path is different than "logs".

Could you use:
https://github.com/kubernetes/test-infra/blob/24c8505d7bd2a59849f6a51fcd03c1239de190ed/prow/gcsupload/run.go#L143
like here:
https://github.com/kubernetes/test-infra/blob/master/prow/pjutil/pjutil.go#L278

tosi3k: Done.

prow/pod-utils/decorate/podspec.go
tosi3k: Outdated - `PathPrefix` is handled inside `PathsForJob` call. Good catch, though.

prow/pod-utils/decorate/podspec.go
```diff
@@ -593,6 +594,17 @@ func decorate(spec *coreapi.PodSpec, pj *prowapi.ProwJob, rawEnv map[string]stri
 	// TODO(fejta): we should pass around volume names rather than forcing particular mount paths.
 
 	rawEnv[artifactsEnv] = artifactsPath
+
+	if gcsConfig := pj.Spec.DecorationConfig.GCSConfiguration; gcsConfig != nil {
+		jobSpec := downwardapi.NewJobSpec(pj.Spec, rawEnv[downwardapi.BuildIDEnv], pj.Name)
+		_, gcsPath, _ := gcsupload.PathsForJob(gcsConfig, &jobSpec, "")
+		rawEnv[gcsArtifactsDirEnv] = "gs://" + path.Join(
+			gcsConfig.Bucket,
```
jkaniuk, line 602: According to: https://github.com/kubernetes/test-infra/blob/master/prow/apis/prowjobs/v1/types.go#L536
that could be:
	// * a GCS bucket: with gs:// prefix
	// * a S3 bucket: with s3:// prefix
	// * a GCS bucket: without a prefix (deprecated)
So prefixing that with gs:// will be wrong in two first cases.

jkaniuk: Could you add a test case for Bucket="gs://..." case?

tosi3k: Changed the logic of the env var creation accordingly.
I added the test case with the bucket name starting with "gs://" as well.

jkaniuk: Could you handle s3:// case as well (in general, any link with the scheme specified?

jkaniuk: OK, I found s3 condition in url above. I am still worried about generality of that.
I would suggest to extract scheme in one place and handle cases (lack of, gs, other)

tosi3k: Applied an appropriate change. Now, if no scheme is provided, it's set to `gs`, otherwise it's being left as it is.

tosi3k:
@stevekuznetsov could you please give a second look at the current version of the PR?

prow/pod-utils/decorate/podspec.go
```diff
@@ -593,6 +595,22 @@ func decorate(spec *coreapi.PodSpec, pj *prowapi.ProwJob, rawEnv map[string]stri
...
+	if gcsConfig := pj.Spec.DecorationConfig.GCSConfiguration; gcsConfig != nil {
+		jobSpec := downwardapi.NewJobSpec(pj.Spec, rawEnv[downwardapi.BuildIDEnv], pj.Name)
+		_, gcsPath, _ := gcsupload.PathsForJob(gcsConfig, &jobSpec, "")
+
+		u, err := url.Parse(gcsConfig.Bucket)
+		if err != nil {
+			return fmt.Errorf("could not parse the GCS bucket name %s: %v", gcsConfig.Bucket, err)
+		}
+		if u.Scheme == "" {
+			u.Scheme = "gs"
```
cjwagner, line 608: Please include unit tests for other schemes. In particular, the cases listed here: https://github.com/kubernetes/test-infra/pull/17215#discussion_r410331907

jkaniuk:
log-exporter is a drop-in replacement for log dump utility, as the latter do not scale.
They save logs to the same location, using different methods.
Requiring to separate buckets for those two methods breaks consistency.

We also do not require that credentials are shared, but only logs location.
However that would prevent log-exporter from being universal tool (which it was), instead of scalability-only one for specific Prow setups.

wojtek-t:
> If the logs are so large can they even be displayed properly by Prow's FE? Can the job upload to a bucket owned by scalability instead of using Prow's result bucket since this isn't a result coming from Prow?
Test containers shouldn't be given access to upload directly to our GCS buckets with the pod utilities. Part of the point of the pod utilities is to avoid giving the test container those credentials. Previously with bootstrap.py there was only the test container in the pod so we had to expose GCS upload credentials to it. Unfortunately many jobs reused these Prow infrastructure credentials instead of supplying their own.

@cjwagner @stevekuznetsov 
First, the logs doesn't have to be huge. The problem is that we need the logs from 5k nodes, so simply downloading them to the test-job pod and then uploading to GCS take 5 hours instead of 5 minutes or so. The logs aren't that huge.

To continue the above, in the past (3 years ago or so), we've been talking about using LogExporter for all jobs - not just for scalability ones. And there seem to be a generic agreement on that.

LogExporter is a more native solution for k8s [we simply schedule pods on the cluster that do the job]. We didn't migrate any other jobs back then, because we want to ensure it's stable enough. But after all those years, it's pretty obvious it is.

Note that the current workflow which is:
- download logs from all nodes to me
- I will upload all of them to GCS
Simply doesn't make any sense - it is introducing obvious bottlenecks. And I strongly believe we should replace it with LogExporter in all our jobs.


Getting to your argument about uploading to different bucket - that doesn't sound like a good solution, also for the reasons I mentioned above. When I'm debugging a failed test, I want to have all logs in a single place and not to always think where I should look for a given file.
The consistency is really important for this imho.


Regarding this PR - I'm not familiar with pod-utils, especially with its security requirements, so it's hard for me to discuss. On one hand that might make sense, OTOH the alternatives mentioned above will significantly hurt any productivity here - I don't think they are really acceptable.

Is there anything else we can do here? Some random (potentially they don't make any sense after deeper thinking) thoughts:
- do the credentials differ per job? If not, can we build an image of log-exporter that will simply have them built-in?

cjwagner:
> To continue the above, in the past (3 years ago or so), we've been talking about using LogExporter for all jobs - not just for scalability ones. And there seem to be a generic agreement on that.

It currently isn't safe to give jobs direct access to Prow's GCS upload credentials because they could overwrite files they don't own if they misbehave. Some jobs still use the deprecated bootstrap.py script and mistakenly have access to those credentials, but they should not be using them. If LogExporter relies on that then we can't adopt it for all jobs currently.

We could mitigate the risks by having separate GCS buckets for Prow's results for scalability jobs so that any misbehavior would be limited to affecting the scalability job results, but that just reduces the blast area without fully addressing the problem.

I think we should have a design discussions about this and get opinions from more parties. @fejta @stevekuznetsov @alvaroaleman @spiffxp

fejta:
LogExporter needs to decide on its own where to upload buckets.

* Choose your own location (prow [injects](https://github.com/kubernetes/test-infra/blob/master/prow/jobs.md#job-environment-variables) a unique `BUILD_ID` environment variable into the test container)
* Write this chosen location into a well-known file somewhere under the directory specified by `ARTIFACTS`.
* Create a [spyglass lens](https://github.com/kubernetes/test-infra/blob/d2e2e04d1b0335335ae006973ba606e95f585af9/prow/spyglass/README.md#configuring-lenses) -- ideally a remote lens -- which knows how to look for this file and then collect information from it.

Prow is a general purpose CI system. Functionality specific to debugging kubernetes clusters does not belong in its pod utilities.

mm4tt:
Hey Folks,

Forgive me my ignorance, this PR only exposes the gcs path to the artifact dir as an env variable.  As @fejta wrote, we could technically reconstruct the gcs path in the log-exporter from `BUILD_ID`. But the conversation in the comments seems to be hooked on **exposing credentials**.

I'm not sure if I follow. All our scalability jobs use the `preset-service-account` ([example](https://github.com/kubernetes/test-infra/blob/e3875f21178fb250db2923b13843e80504c156ee/config/jobs/kubernetes/sig-scalability/sig-scalability-release-blocking-jobs.yaml#L110)) and, AFAIR, log-exporter uses the service account provided by this preset to upload files to gcs. Does it mean that with prow-utils the `preset-service-account` will stop working? 

If yes, then we have a bigger problem than just log-exporter, as some of the functionalities of ClusterLoader2 will stop working as well (e.g. prometheus snapshots, or node-killing where we use gcloud commands)

If no, then I'm not sure if I understand your arguments about security. Exporting the gcs path or not shouldn't have anything to do with test container having access to prow service account credentials.

Could someone clarify?

fejta:
Reconstructing the gcs path from BUILD_ID inside the test container would be fine -- aka doing this outside of prow.

(aside: this remains brittle -- you're still depending on the implementation detail of how and where prow uploads its logs, but this is unlikely to change. And ideally you would provide your own means of authenticating but that isn't a regression so best to handle it separately.)

Reconstructing the gcs path inside the test container so as to unblock the pod util migration is strictly an improvement upon the current state of affairs.

cjwagner:
The credentials provided by `preset-service-account` are only meant to be used by bootstrap.py to upload to GCS for jobs that still haven't migrated to the pod utilities. With bootstrap.py everything runs in one container unlike the pod utils so the credentials are unfortunately exposed to the job workload as well. That is not intended and workloads should not use these credentials themselves. Workloads should provide their own credentials and cloud storage rather than reusing the Prow infrastructure's, especially if you want to make these "universal tools" as mentioned above.

IIUC Erick is saying that using infrastructure credentials is not a regression though so it is a problem that can be addressed separately. I think thats fine, but it is good to consider now since it seems like it will impact the design.

stevekuznetsov:
I agree with Cole and Erick here. Prow exposes most (but not all) of the information you need to have in order to determine where in GCS the infra would push a job result. Specifically, we expose in `$JOB_SPEC` all of the metadata about the job, but since different Prow deployments use different schemes for uploading that, you'd also need to know how a specific deployment is configured to use GCS. The prow.k8s.io. deployment treats `kubernetes/kubernetes` as a special repository, for instance, and `kubernetes` as a special organization, and artifacts for jobs for those repos do not have the org/repo placed into the artifact path. Other deployments do things differently. 

However, as noted, recreating the *exact* path is only useful if the log exporter has access to push to the infrastructure-level GCS bucket, which is fragile and could be harmful if artifacts that Prow uploads and expects to exist are missing. The spyglass design allows for federated lenses, so the ideal process, as outlined by Erick, should look like:

1. use `$JOB_SPEC` or other sources of metadata to upload artifacts somewhere, with log exporter
2. write a lens that (at worst) simply shows an overview of the artifacts or (at best) analyzes them, digests them, or otherwise provides humans a higher-level and more useful abstraction over a file list in the job overview page

mm4tt:
Thanks for all your insights. 

I think I understand how the ultimate solution should look like. TL;DR; using our own credentials, logexporter dumping logs somewhere else than the prow infra-level gcs path (and probably making log-exporter the default to make log-dumping consistent), implementing spyglass lens for displaying the logs in the overview section.

But given our current capacity, we're going to go with the V0 solution to unblock pod-util migration, which is:
1. Changing logexporter to not depend on the GCS_ARTIFACTS_DIR and construct the gcs path from the BUILD_ID
2. Continuing using infrastructure credentials, as Eric said, it's not an regression and if we want to improve it, it can be tackled separately

@tosi3k, FYI

jkaniuk:
Log path depend on Prow settings (organization name in the path) and type of job - periodic, presubmit, postsubmit.
As previously state, we probably do not have all the info to construct full path inside a test pod.
However we can add env variable to each job with beginning of the path and append  `$BUILD_ID` to it.

mm4tt:
@jkaniuk, I think we can take most (if not all) of this info from $JOB_SPEC - see https://github.com/kubernetes/test-infra/blob/c6f0e08ab3f84b53ff0e2d4f9452d19de74948ab/prow/jobs.md#L222

PR #17217, opened 2020-04-15
Adding bugzilla severity labels

shawn-hurley:
Adding severity level as a label when the Bugzilla component is active, allows for quickly searching and categorization based on the impact of the bug.


prow/labels/labels.go
```diff
@@ -21,6 +21,11 @@ const (
 	Approved        = "approved"
 	BlockedPaths    = "do-not-merge/blocked-paths"
 	Bug             = "kind/bug"
+	BZSevUrgent     = "bugzilla/urgent"
```
stevekuznetsov, line 24: nit: can we be explicit in our names

```
BugzillaSeverityUrgent
```

prow/plugins/bugzilla/bugzilla.go
```diff
@@ -48,6 +48,11 @@ var (
 const (
 	PluginName = "bugzilla"
 	bugLink    = `[Bugzilla bug %d](%s/show_bug.cgi?id=%d)`
+	urgentSev  = "urgent"
```
stevekuznetsov, line 51: nit: can we be explicit in our names

```
urgentSeverity
```

prow/plugins/bugzilla/bugzilla.go
```diff
@@ -511,13 +517,29 @@ Comment <code>/bugzilla refresh</code> to re-evaluate validity if changes to the
...
 			hasValidLabel = true
 		}
 		if l.Name == labels.InvalidBug {
 			hasInvalidLabel = true
 		}
+		if l.Name == labels.BZSevHigh || l.Name == labels.BZSevUrgent || l.Name == labels.BZSevMed || l.Name == labels.BZSevLow || l.Name == labels.BZSevUnspec {
+			severityLabelToRemove = l.Name
+		}
+	}
+	//handle severity labels
```
stevekuznetsov, line 532: nit: the code is readable as it is, no need for the comments IMO

prow/plugins/bugzilla/bugzilla.go
```diff
@@ -511,13 +517,29 @@ Comment <code>/bugzilla refresh</code> to re-evaluate validity if changes to the
...
+	//Remove label that should not be set if set:
+	if severityLabelToRemove != "" && severityLabel != severityLabelToRemove {
+		if err := gc.RemoveLabel(e.org, e.repo, e.number, severityLabelToRemove); err != nil {
+			log.WithError(err).Error("Failed to remove severity bug label.")
+		}
+	}
+	if severityLabel != "" && severityLabel != severityLabelToRemove {
+		if err := gc.AddLabel(e.org, e.repo, e.number, severityLabel); err != nil {
+			log.WithError(err).Error(fmt.Sprintf("Failed to add severity bug label. :%v, %v", severityLabel, severityLabelToRemove))
```
stevekuznetsov, line 541: Use `.Errorf(` instead of `.Error(fmt.Sprintf`

We also don't need the labels in the error message, the `WithError()` will expose what we were trying to do with GitHub API anyway

PR #17230, opened 2020-04-16
Add tests for sig-storage-lib-external-provisioner

jsafrane:
Starting with make and make test.

cc @msau42


humblec:
lgtm. just one question on `go` version, do we need to stick to `1.13.10` or is it good to use latest `1.14.2`  @jsafrane ?

jsafrane:
go.mod of sig-storage-lib-external-provisioner says 1.13. And csi external-provisioner image is built with 1.13 too.

PR #17237, opened 2020-04-16
disable-disk-cache-partitions-by-auth-header is not yet legacy.

fejta:
We have yet to come up with a reasonable migration story

/assign @alvaroaleman @stevekuznetsov @cjwagner


fejta:
/hold

Let me switch this to take a token flag

PR #17243, opened 2020-04-16
k8s-jkns-ci-node-e2e is busted use boskos to pick run ci-kubernetes-node-kubelet

dims:
Trying a "quick" fix.


spiffxp:

PR #17247, opened 2020-04-16
k8s-jkns-ci-node-e2e is busted switch to k8s-jkns-gke-gci-soak

dims:
different experiment compared to https://github.com/kubernetes/test-infra/pull/17243

in 17243 we switched ci-kubernetes-node-kubelet to use the general boskos pool

in this, we are trying another empty project ci-kubernetes-node-kubelet-1-18


spiffxp:

PR #17250, opened 2020-04-16
Remove github flags that plank no longer uses

fejta:


fejta:
Really gotta fix these OWNERS files.

/uncc @ixdy @amwat 
/assign @cjwagner @michelle192837

PR #17251, opened 2020-04-17
jobs: pin ci-kubernetes-node-e2e-kubelet to spare project

spiffxp:
so we know it can work with boskos managed projects, but until we get
a pool set aside for node jobs, I'd rather prevent them from using
the pool used by most e2e jobs, in case a network ban happens again

/cc @dims


spiffxp:

PR #17253, opened 2020-04-17
pr-creator: configurable "maintainer_can_modify" value for PR

clarketm:
Currently, all Istio developers have repo-level **write** permission. For security reasons, there are _specific_ PRs created via `pr-creator` that we want to be immutable (e.g. `auto-merge` PRs).

**GH API docs**: _Create a pull request_ -> [`maintainer_can_modify`](https://developer.github.com/v3/pulls/#input)

https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork

cc @howardjohn

### DIFF
--- a/experiment/autobumper/main.go
+++ b/experiment/autobumper/main.go
@@ -30,6 +30,7 @@ import (
 	"k8s.io/test-infra/experiment/autobumper/bumper"
 	"k8s.io/test-infra/prow/config/secret"
 	"k8s.io/test-infra/prow/github"
+	"k8s.io/test-infra/robots/pr-creator/updater"
 )
 
 const (
@@ -174,7 +175,7 @@ func main() {
 		logrus.WithError(err).Fatal("Failed to push changes.")
 	}
 
-	if err := bumper.UpdatePR(gc, githubOrg, githubRepo, images, getAssignment(), "Update prow to", o.githubLogin+":"+remoteBranch, "master", true); err != nil {
+	if err := bumper.UpdatePR(gc, githubOrg, githubRepo, images, getAssignment(), "Update prow to", o.githubLogin+":"+remoteBranch, "master", updater.PreventMods); err != nil {
 		logrus.WithError(err).Fatal("PR creation failed.")
 	}
 }
--- a/experiment/autobumper/main.go
+++ b/experiment/autobumper/main.go
@@ -174,7 +174,7 @@ func main() {
 		logrus.WithError(err).Fatal("Failed to push changes.")
 	}
 
-	if err := bumper.UpdatePR(gc, githubOrg, githubRepo, images, getAssignment(), "Update prow to", o.githubLogin+":"+remoteBranch, "master"); err != nil {
+	if err := bumper.UpdatePR(gc, githubOrg, githubRepo, images, getAssignment(), "Update prow to", o.githubLogin+":"+remoteBranch, "master", true); err != nil {
 		logrus.WithError(err).Fatal("PR creation failed.")
 	}
 }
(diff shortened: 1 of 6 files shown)


robots/pr-creator/main.go
```diff
@@ -73,6 +74,7 @@ func optionsFromFlags() options {
 	fs.StringVar(&o.branch, "branch", "", "Repo branch to merge into")
 	fs.StringVar(&o.source, "source", "", "The user:branch to merge from")
 
+	fs.BoolVar(&o.allowMods, "allow-mods", true, "Indicates whether maintainers can modify the pull request")
```
fejta, line 77: Please name flags such that they default to false.

AKA instead of naming this `--allow-mods=true` please do something like `--prevent-mods=false`

clarketm:
> Is this a hypothetical or realized problem?
> 
> AKA what examples of members modifying PRs that we didn't want exist so far?

The realized problem is for **auto-merge** PRs (https://github.com/istio/istio/pull/23013). Any Istio developer ~200 people could technically add a commit here, push it to the fork, and  _sneak_ it into the codebase unreviewed.

robots/pr-creator/main.go
clarketm: Instead of renaming (which may cause confusion since if will be inverse of corresponding GH API param), should we just _default_ to `false`?

experiment/autobumper/main.go
```diff
@@ -174,7 +174,7 @@ func main() {
 		logrus.WithError(err).Fatal("Failed to push changes.")
 	}
 
-	if err := bumper.UpdatePR(gc, githubOrg, githubRepo, images, getAssignment(), "Update prow to", o.githubLogin+":"+remoteBranch, "master"); err != nil {
+	if err := bumper.UpdatePR(gc, githubOrg, githubRepo, images, getAssignment(), "Update prow to", o.githubLogin+":"+remoteBranch, "master", true); err != nil {
```
fejta, line 177: Consider using a const here to improve readability
```golang
const disallowMods = false
if err := bumper.UpdatePR(gc, githubOrg, ..., disallowMods) {
```

(note that this is how updater.go worked before)

robots/pr-creator/main.go
fejta: The github api is `maintainer_can_modify` so I wouldn't consider that an issue.

Sure if you want to default to `--allow-mods=false` then I'd keep the current name. If you want to default to `--disallow-mods=false` then I'd choose a name more along those lines (I'd consider something like prevent superior to disallow as don't prevent sounds less like a double negative than don't disallow)

robots/pr-creator/updater/updater.go
```diff
@@ -67,17 +67,16 @@ func UpdatePR(org, repo, title, body, matchTitle string, gc updateClient) (*int,
...
 }
 
-func EnsurePRWithLabels(org, repo, title, body, source, branch, matchTitle string, gc ensureClient, labels []string) (*int, error) {
+func EnsurePRWithLabels(org, repo, title, body, source, branch, matchTitle string, allowMods bool, gc ensureClient, labels []string) (*int, error) {
 	n, err := UpdatePR(org, repo, title, body, matchTitle, gc)
 	if err != nil {
 		return nil, fmt.Errorf("update error: %v", err)
 	}
 	if n == nil {
-		allowMods := true
```
fejta, line 80: might be useful to export values for people to use for readability, aka
```golang
const (
  AllowMods = true
  PreventMods = false
)
```

robots/pr-creator/main.go
clarketm: 👍  I will keep the current name and _default_ to `false`. This will also align with `maintainer_can_modify`.

robots/pr-creator/updater/updater.go
clarketm: Makes sense. I will add constants for these,

PR #17254, opened 2020-04-17
Spring cleaning

BenTheElder:
Doing a bit of spring cleaning on some of the things I maintain here.

Removed the following dead things:
- `images/kubekins-test`. 
  - we've migrated Kubernetes 1.16+ off of this, and 1.15 is about to be out of support. it uses the 1.14 image, which hasn't been touched in a long time
- `planter/`, spoke to @ixdy -- we're not aware of anyone actively using this, we're not actively maintaining it, and it's pretty obsolete with modern bazel version management

Ensured we have some active owners for the following not dead things
- `images/krte`
  - slimmer `kubekins-e2e` replacement the podutils world without bootstrap.py and other junk.
- `images/kubekins-e2e`
  - still necessary for most kubernetes e2e testing for now
- `kubetest2`
  - we're aiming to revamp this and drop-in-replace kubetest with something more maintainable and extensible
- `kubetest`
  - we still use this basically everywhere
- kind repo jobs
  - could use more reviewers on this too! updated

Added myself to one of @Katharine's projects that I have context on:
- `gopherage`
  - otherwise only has @spiffxp @Katharine reviewing / approving.

cc @krzyzacy 
/cc @spiffxp @ixdy @amwat


images/kubekins-e2e/OWNERS
```diff
@@ -0,0 +1,10 @@
+# See the OWNERS docs at https://go.k8s.io/owners
+
+reviewers:
+- amwat
+- spiffxp
+- BenTheElder
+approvers:
+- amwat
+- spiffxp
```
spiffxp, line 9: oh boy, now where is that elixir of knowledge that grants me all the history behind what's in here

BenTheElder: nobody actually has this :-)
if you ask sen and I about it and dig through git you might figure it out.

for this image we should just generally avoid major changes.

the krte image is simpler and I'm attempting to leave comments for everything.
we should plan to move away from the bootstrap cruft anyhow :sweat_smile:

spiffxp:
missed the planter job https://github.com/kubernetes/test-infra/pull/17271

spiffxp:
krte push failed: https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/post-test-infra-push-krte/1250963000605544449

clicked retry in deck: https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/post-test-infra-push-krte/1251192889321459712

BenTheElder:
Thanks.

Looks like ssl issues installing bazel? Hmm

spiffxp:
krte retry failed, but with a different variant, let's retry one more time: https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/post-test-infra-push-krte/1251204717929828352

spiffxp:
passed, all done

PR #17256, opened 2020-04-17
pass the env we want

BenTheElder:
ref != sha
sigh
/cc @spiffxp @amwat


BenTheElder:
tbr. trivial env swap.

PR #17261, opened 2020-04-17
Strip markdown comments from label matcher

saschagrunert:
Right now we have to prefix the labels with something (like with `> ` in
k/k) to not apply a label. With this patch we can comment them out but
still preserve their formatting, which makes enabling labels much
easier. For example:

```markdown
<!-- Choose a kind:
/kind bug
/kind feature
-->
```

### DIFF
--- a/prow/plugins/label/label.go
+++ b/prow/plugins/label/label.go
@@ -34,6 +34,7 @@ const pluginName = "label"
 
 var (
 	defaultLabels          = []string{"kind", "priority", "area"}
+	commentRegex           = regexp.MustCompile(`(?s)<!--(.*?)-->`)
 	labelRegex             = regexp.MustCompile(`(?m)^/(area|committee|kind|language|priority|sig|triage|wg)\s*(.*?)\s*$`)
 	removeLabelRegex       = regexp.MustCompile(`(?m)^/remove-(area|committee|kind|language|priority|sig|triage|wg)\s*(.*?)\s*$`)
 	customLabelRegex       = regexp.MustCompile(`(?m)^/label\s*(.*?)\s*$`)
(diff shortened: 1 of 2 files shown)


prow/plugins/label/label.go
```diff
@@ -34,6 +34,7 @@ const pluginName = "label"
 
 var (
 	defaultLabels          = []string{"kind", "priority", "area"}
+	commentRegex           = regexp.MustCompile(`(?s)<!--(.*)-->`)
```
liggitt, line 37: probably use non-greedy matching, otherwise two consecutive comments could drop content between them?

```suggestion
	commentRegex           = regexp.MustCompile(`(?s)<!--(.*?)-->`)
```

prow/plugins/label/label_test.go
```diff
@@ -575,6 +575,16 @@ func TestLabel(t *testing.T) {
...
+			name:                  "Strip markdown comments",
+			body:                  "<!--\n/kind bug\n/kind cleanup\n-->\n/area infra",
+			repoLabels:            []string{"area/infra"},
+			issueLabels:           []string{},
+			expectedNewLabels:     formatLabels("area/infra"),
+			expectedRemovedLabels: []string{},
+			commenter:             orgMember,
+			action:                github.GenericCommentActionCreated,
+		},
 	}
```
liggitt, line 588: add a test that has multiple comments and make sure this ignores commented content but still recognizes uncommented content:

```
<!--
/area comment1
-->
/area infra1
<!--
/area comment2
-->
/area infra2
```

prow/plugins/label/label.go
saschagrunert: Sounds good, fixed :+1:

prow/plugins/label/label_test.go
saschagrunert: Added the test as suggested.

saschagrunert:
> /hold for discussion on [kubernetes/kubernetes#89946](https://github.com/kubernetes/kubernetes/pull/89946)

I'm wondering which part of the discussion has to be resolved first to move forward.

In any case, I rebased on top of the latest master branch.

saschagrunert:
@justaugustus do we really need to hold this? I think this is a general improvement to the kind parser in Prow.

justaugustus:
This got lost to the sands of time.
/lgtm
/approve
/hold cancel
/assign @cblecker @fejta

PR #17264, opened 2020-04-17
jobs: pin release-informing node jobs away from k8s-jkns-ci-node-e2e

hasheddan:
release-blocking jobs were already moved to a different GCP project in #17251 and #17257 due to network issues. This also moves the release-informing jobs away until new GCP project is created (ref: https://github.com/kubernetes/kubernetes/issues/89892#issuecomment-614998826).

Ref: https://github.com/kubernetes/kubernetes/issues/89893

@spiffxp if you don't think this is necessary feel free to close this. However, it would be helpful from a CI Signal perspective for us to be getting more accurate signal from these release-informing tests in the short-term.

/cc @spiffxp @dims @kubernetes/ci-signal

### DIFF
--- a/config/jobs/kubernetes/sig-node/node-kubelet.yaml
+++ b/config/jobs/kubernetes/sig-node/node-kubelet.yaml
@@ -146,7 +146,7 @@ periodics:
       - --scenario=kubernetes_e2e
       - --
       - --deployment=node
-      - --gcp-project=k8s-jkns-ci-node-e2e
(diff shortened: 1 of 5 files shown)


spiffxp:
I would prefer not to for the time being, given that informing jobs can be routinely ignored

hasheddan:
@spiffxp that sounds reasonable to me, thanks for following up :+1:

spiffxp:
https://github.com/kubernetes/test-infra/pull/17349 should address this

PR #17269, opened 2020-04-17
jobs: use bazel 2.2.0 for test-infra bazel test jobs

spiffxp:
It's not clear to me whether these should be ignoring .bazelversion or not.

Right now the periodic is, so it's failing: https://k8s-testgrid.appspot.com/sig-testing-misc#ci-bazel is failing


fejta:
> It's not clear to me whether these should be ignoring .bazelversion or not.

We either need to change this use an image that has multiple versions of bazel installed, or we need to use the real version of bazel installed.

PR #17276, opened 2020-04-17
Revert "jobs: use bazel 2.2.0 for test-infra bazel test jobs"

fejta:
Reverts kubernetes/test-infra#17269


fejta:
Cannot use 2.2.0 until we update repo-infra to a version that supports it: https://github.com/kubernetes/repo-infra/releases

PR #17279, opened 2020-04-17
boskos-resources: shrink gke-project pool

spiffxp:
this pool is now wildly overprovisioned, I'd like to use some of it
to create some smaller canary pools


ixdy:
/lgtm

If these are in-use, they may not actually be deleted, and may require manual cleanup (i.e. deletion) from the oncall.

ixdy:
https://github.com/kubernetes/test-infra/issues/17282 for the issue I described.

spiffxp:
Looks like they were deleted, no `gke-project` resources
```
$ k get resources -n test-pods | grep k8s-jkns-gke
k8s-jkns-gke-reboot-1-4          gce-project                     free                                                                               17m
k8s-jkns-gke-slow-1-4            gce-project                     free
```

PR #17286, opened 2020-04-18
Allow kettle build to use sandbox

matthyx:
https://github.com/bazelbuild/bazel/issues/1973 and https://github.com/bazelbuild/bazel/issues/2056 will probably never be fixed, but let's see if we can use sandbox anyways!


matthyx:
Looks like it works...
/assign @spiffxp

PR #17287, opened 2020-04-18
Run e2e test TCP.CLOSE_WAIT with kind

aojea:
This e2e test is still flaky

Run it in KIND so we can diff and discard a possible environment issue

xref: https://github.com/kubernetes/kubernetes/issues/71565


BenTheElder:
> This e2e test is still flaky

I would prefer to not intentionally flake out my CI, thanks :-)

BenTheElder:
we can already see that this hits kops and cluster/ GCE, can we not fix the issue instead of making kind CI intentionally flakier..?

we put a *lot* of work into providing more reliable signal than the cloudy jobs. I still regularly spot check flakiness.

aojea:
I was not able to reproduce it locally with KIND 🤷‍♂️ 

Let me try other thing then ...
/close

PR #17293, opened 2020-04-20
Add link checker to presubmits

scottilee:
Adds a presubmit to check for broken links.

Fixes https://github.com/kubeflow/website/issues/73.

### DIFF
--- a/config/jobs/kubeflow/kubeflow-presubmits.yaml
+++ b/config/jobs/kubeflow/kubeflow-presubmits.yaml
@@ -83,6 +83,19 @@ presubmits:
       testgrid-dashboards: sig-big-data
       description: Presubmits for kubeflow/website.
       testgrid-num-columns-recent: '30'
+  - name: kubeflow-website-link-check
+    cluster: kubeflow
+    always_run: true         # Run for every PR, or only when requested.
+    labels:
+      preset-service-account: "true"
+    spec:
+      containers:
+      - image: ktbartholomew/link-checker:latest
+        imagePullPolicy: Always
+    annotations:
+      testgrid-dashboards: sig-big-data
+      description: Presubmits for kubeflow/website to check for broken links.
+      testgrid-num-columns-recent: '30'
   kubeflow/testing:
   - name: kubeflow-testing-presubmit
     cluster: kubeflow
--- a/config/jobs/kubeflow/kubeflow-presubmits.yaml
+++ b/config/jobs/kubeflow/kubeflow-presubmits.yaml
@@ -85,8 +85,7 @@ presubmits:
       testgrid-num-columns-recent: '30'
   - name: kubeflow-website-link-check
     cluster: kubeflow
-    always_run: false         # Run for every PR, or only when requested.
-    skip_report: true
+    optional: true
     labels:
       preset-service-account: "true"
     spec:
--- a/config/jobs/kubeflow/kubeflow-presubmits.yaml
+++ b/config/jobs/kubeflow/kubeflow-presubmits.yaml
@@ -90,7 +90,7 @@ presubmits:
       preset-service-account: "true"
     spec:
       containers:
-      - image: ktbartholomew/link-checker:latest
+      - image: docker.io/praqma/linkchecker:v9.3.1-154-g22449abb-10
         imagePullPolicy: Always
     annotations:
       testgrid-dashboards: sig-big-data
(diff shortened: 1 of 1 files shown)


config/jobs/kubeflow/kubeflow-presubmits.yaml
```diff
@@ -83,6 +83,19 @@ presubmits:
...
       description: Presubmits for kubeflow/website.
       testgrid-num-columns-recent: '30'
+  - name: kubeflow-website-link-check
+    cluster: kubeflow
+    always_run: true         # Run for every PR, or only when requested.
+    labels:
+      preset-service-account: "true"
+    spec:
+      containers:
+      - image: ktbartholomew/link-checker:latest
```
jlewi, line 93: Where is this image coming from?
Can we pin the image so we don't break suddenly if the latest image is updated?

scottilee: Fixed.

config/jobs/kubeflow/kubeflow-presubmits.yaml
```diff
@@ -83,6 +83,19 @@ presubmits:
       testgrid-dashboards: sig-big-data
       description: Presubmits for kubeflow/website.
       testgrid-num-columns-recent: '30'
+  - name: kubeflow-website-link-check
```
jlewi, line 86: Is there a way to mark it as non-blocking so we can verify that is working correctly before we start blocking presubmits?

scottilee: Yes, per the [docs](https://github.com/kubernetes/test-infra/blob/master/prow/jobs.md) I set `always_run: false` and `skip_report: true`. If the test is working and helpful we can flip those later.

jlewi: Why set always_run to false? Is there a way to always run it and surface the result to the user without having failures be blocking?

clarketm: @scottilee. I think the field we want here is `optional: true`. This will allow the job to automatically and appear in the PR context, but make it non-blocking.

clarketm: https://github.com/kubernetes/test-infra/blob/860d1bd5012fae1ae7330e2b59806763ce2b430e/prow/config/jobs.go#L136-L137

config/jobs/kubeflow/kubeflow-presubmits.yaml
```diff
@@ -83,6 +83,20 @@ presubmits:
...
+    cluster: kubeflow
+    always_run: false         # Run for every PR, or only when requested.
+    skip_report: true
```
clarketm, line 89:
```suggestion
    optional: true
```

scottilee: Updated.

config/jobs/kubeflow/kubeflow-presubmits.yaml
```diff
@@ -83,6 +83,19 @@ presubmits:
...
       description: Presubmits for kubeflow/website.
       testgrid-num-columns-recent: '30'
+  - name: kubeflow-website-link-check
+    cluster: kubeflow
+    optional: true
+    labels:
+      preset-service-account: "true"
+    spec:
+      containers:
+      - image: docker.io/praqma/linkchecker:v9.3.1-154-g22449abb-10
```
clarketm, line 93: Not an issue as this image is valid, but FYI in case you were unaware, images from docker hub do not require the `docker.io/` prefix.

 docker.io/praqma/linkchecker:v9.3.1-154-g22449abb-10
```suggestion
      - image: praqma/linkchecker:v9.3.1-154-g22449abb-10
```

clarketm:
/approve
lgtm from a Prow config perspective. I will defer to @jlewi for final lgtm.

config/jobs/kubeflow/kubeflow-presubmits.yaml
scottilee: Fixed.

config/jobs/kubeflow/kubeflow-presubmits.yaml
jlewi: Great thanks

jlewi:
Looks like the test is now marked as optional so it will always run but not block.

/hold cancel

PR #17296, opened 2020-04-20
[Scalability] Add extended probes timeout for containerd tests

jprzychodzen:
/sig scalability
/assign @mm4tt  
Ref. kubernetes/perf-tests#1110

Add custom probes measurement timeout for sig-scalability containerd preset.

### DIFF
--- a/config/jobs/kubernetes/sig-scalability/sig-scalability-presets.yaml
+++ b/config/jobs/kubernetes/sig-scalability/sig-scalability-presets.yaml
@@ -185,6 +185,8 @@ presets:
 - labels:
     preset-e2e-scalability-containerd: "true"
   env:
+  - name: CL2_PROBE_MEASUREMENTS_CHECK_PROBES_READY_TIMEOUT
+    value: "10m"
   - name: KUBE_CONTAINER_RUNTIME
     value: "containerd"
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes/sig-scalability/sig-scalability-presets.yaml
```diff
@@ -185,6 +185,8 @@ presets:
 - labels:
     preset-e2e-scalability-containerd: "true"
   env:
+  - name: CL2_PROBE_MEASUREMENTS_CHECK_PROBES_READY_TIMEOUT
+    value: "10m"
```
mm4tt, line 189: Are we sure that 10min will be enough?

jprzychodzen: Pull of image can timeout after 6 minutes and will try again, and current timeout is 5 minutes, becoming in total 11 minutes. We can lover this as some work is already done - all pods are scheduled, and almost all of them should already run.

Checking values from test, this can be as low as 7-8 minutes (time for DNS probes to be reported to Prometheus for 100 node experimental runs), but I do not want to go so low here. A little overhead seams reasonable.

Please unhold if this SG.

PR #17301, opened 2020-04-20
prow: use private opener interface in tide, statusreconciler and gerrit

sbueringer:
We should use private interfaces instead of the the one from `prow/io` so we don't have to implement all methods of the interface in the fakeOpeners. Also we now precisely know which methods are required in these cases and don't require other methods which won't be used here.

See also: https://github.com/kubernetes/test-infra/pull/17183#discussion_r409740786


stevekuznetsov:
However, this makes the code less direct and harder to follow. Why not make a useful mock in the `io` package so using it is simple?

PR #17302, opened 2020-04-20
Update gcloud-bazel usage

fejta:
This version has both 2.2.0 and 2.0.0 installed
Also update the gcloud-bazel/push.sh script to specify which versions to push

/assign @chases2 @spiffxp


fejta:
Thanks!

fejta:
ref https://github.com/bazelbuild/rules_k8s/pull/549

PR #17304, opened 2020-04-20
Fix Prow auto-scaling

coryrc:
Autoscaling Prow clusters intermittently fail with messages like
`Could not resolve host: github.com`. Retry clone when this is detected.

Fixes: https://github.com/knative/test-infra/issues/1082
Fixes: https://github.com/kyma-project/test-infra/issues/1198

### DIFF
--- a/prow/clonerefs/run.go
+++ b/prow/clonerefs/run.go
@@ -81,7 +81,24 @@ func (o Options) Run() error {
 		go func() {
 			defer wg.Done()
 			for ref := range input {
-				output <- cloneFunc(ref, o.SrcRoot, o.GitUserName, o.GitUserEmail, o.CookiePath, env, oauthToken)
+				// Attempt to clone the repo
+				// If the cluster is autoscaling, it seems Prow can hit DNS too hard on new nodes
+				// So we retry if have a failure we can guess is due to DNS
+				// Exponential backoff is targeted for a little over 5 minutes on the last try
+				for i := 1; i < 4; i++ {
+					record := cloneFunc(ref, o.SrcRoot, o.GitUserName, o.GitUserEmail, o.CookiePath, env, oauthToken)
+					if record.Failed {
+						failed := record.Commands[len(record.Commands)-1]
+						if strings.Contains(failed.Error, "Could not resolve host") {
+							seconds := time.Seconds * 25 * i * i
+							logrus.Infof("DNS failure detected; retrying clone in %d seconds", seconds)
+							time.Sleep(seconds)
+							continue
+						}
+					}
+					output <- record
+					break
+				}
 			}
 		}()
 	}
--- a/prow/clonerefs/run.go
+++ b/prow/clonerefs/run.go
@@ -85,15 +85,16 @@ func (o Options) Run() error {
 				// Attempt to clone the repo
 				// If the cluster is autoscaling, it seems Prow can hit DNS too hard on new nodes
 				// So we retry if have a failure we can guess is due to DNS
-				// Exponential backoff is targeted for a little over 5 minutes on the last try
-				for i := 1; i < 4; i++ {
+				// Exponential backoff is targeted for a little over 5 minutes (cumulative) on the last try
+				numIterations := 4
+				for i := 1; i <= numIterations; i++ {
 					record := cloneFunc(ref, o.SrcRoot, o.GitUserName, o.GitUserEmail, o.CookiePath, env, oauthToken)
 					if record.Failed {
 						failed := record.Commands[len(record.Commands)-1]
-						if strings.Contains(failed.Error, "Could not resolve host") {
-							seconds := time.Second * 25 * i * i
-							logrus.Infof("DNS failure detected; retrying clone in %d seconds", seconds)
-							time.Sleep(seconds)
+						if strings.Contains(failed.Error, "Could not resolve host") && i != numIterations {
+							delay := time.Second * 25 * i * i
+							logrus.Infof("DNS failure detected; retrying clone in %d seconds", delay)
+							time.Sleep(delay)
 							continue
 						}
 					}
--- a/prow/clonerefs/run.go
+++ b/prow/clonerefs/run.go
@@ -92,7 +92,7 @@ func (o Options) Run() error {
 					if record.Failed {
 						failed := record.Commands[len(record.Commands)-1]
 						if strings.Contains(failed.Error, "Could not resolve host") && i != numIterations {
-							delay := time.Second * 25 * i * i
+							delay := time.Second * time.Duration(25*i*i)
 							logrus.Infof("DNS failure detected; retrying clone in %d seconds", delay)
 							time.Sleep(delay)
 							continue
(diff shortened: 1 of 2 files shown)


cjwagner:
```
ERROR: /home/prow/go/src/github.com/kubernetes/test-infra/prow/clonerefs/BUILD.bazel:3:1: GoCompilePkg prow/clonerefs/linux_amd64_pure_stripped/go_default_library%/k8s.io/test-infra/prow/clonerefs.a failed (Exit 1)
/b/f/w/prow/clonerefs/run.go:93:19: undefined: time
/b/f/w/prow/clonerefs/run.go:95:8: undefined: time
```
/ok-to-test

prow/clonerefs/run.go
```diff
@@ -81,7 +82,24 @@ func (o Options) Run() error {
...
 			defer wg.Done()
 			for ref := range input {
-				output <- cloneFunc(ref, o.SrcRoot, o.GitUserName, o.GitUserEmail, o.CookiePath, env, oauthToken)
+				// Attempt to clone the repo
+				// If the cluster is autoscaling, it seems Prow can hit DNS too hard on new nodes
+				// So we retry if have a failure we can guess is due to DNS
+				// Exponential backoff is targeted for a little over 5 minutes on the last try
+				for i := 1; i < 4; i++ {
+					record := cloneFunc(ref, o.SrcRoot, o.GitUserName, o.GitUserEmail, o.CookiePath, env, oauthToken)
+					if record.Failed {
```
fejta, line 91: Seems like the following would be simpler:
```golang
if !record.Failed {
  output <- record
  break
}
// sleep

File: prow/clonerefs/run.go [hunk 2]
```diff
@@ -81,7 +82,24 @@ func (o Options) Run() error {
 		go func() {
 			defer wg.Done()
 			for ref := range input {
-				output <- cloneFunc(ref, o.SrcRoot, o.GitUserName, o.GitUserEmail, o.CookiePath, env, oauthToken)
```
fejta, line 84: This always writes to the output channel. The changes need to keep doing that.

prow/clonerefs/run.go
```diff
@@ -81,7 +82,24 @@ func (o Options) Run() error {
...
+				// So we retry if have a failure we can guess is due to DNS
+				// Exponential backoff is targeted for a little over 5 minutes on the last try
+				for i := 1; i < 4; i++ {
+					record := cloneFunc(ref, o.SrcRoot, o.GitUserName, o.GitUserEmail, o.CookiePath, env, oauthToken)
+					if record.Failed {
+						failed := record.Commands[len(record.Commands)-1]
+						if strings.Contains(failed.Error, "Could not resolve host") {
+							seconds := time.Second * 25 * i * i
+							logrus.Infof("DNS failure detected; retrying clone in %d seconds", seconds)
+							time.Sleep(seconds)
```
fejta, line 96: We should not sleep when `i == 3`

prow/clonerefs/run.go
```diff
@@ -81,7 +82,24 @@ func (o Options) Run() error {
...
+				// Attempt to clone the repo
+				// If the cluster is autoscaling, it seems Prow can hit DNS too hard on new nodes
+				// So we retry if have a failure we can guess is due to DNS
+				// Exponential backoff is targeted for a little over 5 minutes on the last try
+				for i := 1; i < 4; i++ {
+					record := cloneFunc(ref, o.SrcRoot, o.GitUserName, o.GitUserEmail, o.CookiePath, env, oauthToken)
+					if record.Failed {
+						failed := record.Commands[len(record.Commands)-1]
+						if strings.Contains(failed.Error, "Could not resolve host") {
+							seconds := time.Second * 25 * i * i
```
fejta, line 94: `delay` or something along those lines is a better name (the int64 in this variable is not 25 when i = 1)

stevekuznetsov:
Why should `prow` have to know about this? We auto-scale our nodes and use a DNS solution that works, always, so we never hit this. IMO this is the wrong layer.

prow/clonerefs/run.go
coryrc: I need to output when it does fail under some scenarios too

prow/clonerefs/run.go
coryrc: I'm missing the fails-3-times scenario writing-to-channel. Will figure out.

prow/clonerefs/run.go
coryrc: Why not? That's the intention.

prow/clonerefs/run.go
coryrc: delay is fine as a name

prow/clonerefs/run.go
coryrc: oh, I see; I'm doing the sleep wrong. Oh darn.

fejta: It failed all four times and will now exit. There's no need to make it sleep for 400 seconds before reporting this failure.

coryrc:
> Why should `prow` have to know about this? We auto-scale our nodes and use a DNS solution that works, always, so we never hit this. IMO this is the wrong layer.

This is the only

coryrc:
(wrong button)

coryrc:
> Why should `prow` have to know about this? We auto-scale our nodes and use a DNS solution that works, always, so we never hit this. IMO this is the wrong layer.

Because we haven't been able to figure it out.

fejta:
The clonerefs image prow injects is controlled here (does not need to be the same as the rest of prow):

https://github.com/kubernetes/test-infra/blob/a2bf5f85155043c7ca2dc01661d1d710a1bb46a6/config/prow/config.yaml#L9-L14

alvaroaleman:
> Why should prow have to know about this? We auto-scale our nodes and use a DNS solution that works, always, so we never hit this. IMO this is the wrong layer

Yeah. Wild guess would be that you use node-local-dns DaemonSet and have nothing that taints your nodes until those containers are up and ready. Please properly debug your issue instead of adding a bandaid-workaround here.

I am not  generally opposed to have retrying as everything might fail intermittently if a network is involved but:
* Please do not make that retrying specific to DNS failures, make it generic with a whitelist of known terminal errors that are not retried (e.G. 404 when cloning)
* Both the initial retry interval and the total time spent retrying are _way_ to high in my opinion

alvaroaleman:
Also, the `fixes` directives on the PR body are misleading. Kubernetes without nodelocal DNS Daemonset has known issues with DNS under load due to a conntrack race (xref https://github.com/kubernetes/kubernetes/issues/56903). That error is not specific to using git, it just happens to be that git doesn't retry DNS lookups.

cjwagner:
If you want to build your own version of clonerefs that includes your workaround there are instructions for publishing the Prow component images here: https://github.com/kubernetes/test-infra/blob/master/prow/build_test_update.md#how-to-update-the-cluster

> Update your deployment (optionally build/pushing the image) to a new image with:
> ```shell
> # export PROW_REPO_OVERRIDE=gcr.io/k8s-prow  # optionally override project
> push.sh  # Build and push the current repo state.
> bump.sh --list  # Choose a recent published version
> bump.sh v20181002-deadbeef # Use a specific version
> ```

coryrc:
Will be using private clonerefs image

Issue #17298, opened 2020-04-20
test-infra postsubmits are broken


alvaroaleman:
**What happened**:

The post-test-infra-push-prow fails consistently:
https://prow.k8s.io/?job=post-test-infra-push-prow

Error seems to be `Current Bazel version is 2.0.0; expected at least 2.2.0`

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

**Please provide links to example occurrences, if any**:

**Anything else we need to know?**:

/cc @fejta

spiffxp:
/area bazel
/cc @fejta 
`gcr.io/k8s-testimages/gcloud-bazel` needs to be updated to use bazel 2.2.0 (and then 2.2.0 images pushed and jobs updated) or https://github.com/kubernetes/test-infra/pull/17281 needs to be reverted

fejta:
Best as I can tell all the failing jobs have been rerun: https://prow.k8s.io/?repo=kubernetes%2Ftest-infra&type=postsubmit

PR #17305, opened 2020-04-20
Refactor bazelbuild to install multiple bazel versions.

fejta:
This matches the pattern kubekins-e2e uses:

https://github.com/kubernetes/test-infra/blob/42ca103703317f0db03de8803f852ad7224c26c6/images/kubekins-e2e/Dockerfile#L18-L19

https://github.com/kubernetes/test-infra/blob/42ca103703317f0db03de8803f852ad7224c26c6/images/kubekins-e2e/Dockerfile#L79-L81

ref https://github.com/kubernetes/test-infra/issues/17298

### DIFF
--- a/images/bazelbuild/Dockerfile
+++ b/images/bazelbuild/Dockerfile
@@ -13,6 +13,8 @@
 # limitations under the License.
 
 # Includes git, gcloud, and bazel.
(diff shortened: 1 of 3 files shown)


chases2:
/lgtm
/hold
Would like review from @cjwagner  if available

PR #17309, opened 2020-04-20
Fix bazel image postsubmit

fejta:
* docker tag can only tag one image at a time
* just tag GIT-CONFIG and latest-GIT-CONFIG

/assign @BenTheElder @chases2

### DIFF
--- a/images/bazel/cloudbuild.yaml
+++ b/images/bazel/cloudbuild.yaml
@@ -2,23 +2,23 @@ steps:
   - name: gcr.io/cloud-builders/docker
     args: 
     - build
-    - --tag=gcr.io/$PROJECT_ID/launcher.gcr.io/google/bazel:$_NEW_VERSION-from-$_OLD_VERSION
+    - --tag=gcr.io/$PROJECT_ID/launcher.gcr.io/google/bazel:$_GIT_TAG-$_CONFIG
     - --build-arg=NEW_VERSION=$_NEW_VERSION
     - --build-arg=OLD_VERSION=$_OLD_VERSION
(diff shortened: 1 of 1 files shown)


images/bazel/cloudbuild.yaml
```diff
@@ -2,23 +2,23 @@ steps:
...
     - build
-    - --tag=gcr.io/$PROJECT_ID/launcher.gcr.io/google/bazel:$_NEW_VERSION-from-$_OLD_VERSION
+    - --tag=gcr.io/$PROJECT_ID/launcher.gcr.io/google/bazel:$_GIT_TAG-$_CONFIG
     - --build-arg=NEW_VERSION=$_NEW_VERSION
     - --build-arg=OLD_VERSION=$_OLD_VERSION
-    - --build-arg=IMAGE_ARG=gcr.io/$PROJECT_ID/launcher.gcr.io/google/bazel:$_NEW_VERSION-from-$_OLD_VERSION
+    - --build-arg=IMAGE_ARG=gcr.io/$PROJECT_ID/launcher.gcr.io/google/bazel:$_GIT_TAG-$_CONFIG
     - .
+
   - name: gcr.io/cloud-builders/docker
```
BenTheElder, line 11: Could probably override entry point to a shell and run N tag commands?

fejta: Not actually sure if cloudbuild allows that? But nifty idea. Let me investigate later.

BenTheElder: It does -- 
https://cloud.google.com/cloud-build/docs/build-config#entrypoint

PR #17310, opened 2020-04-20
Upgrade the bazel-base image to the one test-infra uses

fejta:
/assign @chases2


BenTheElder:
/retitle Upgrade the bazel-base image to the one test-infra uses
/lgtm

PR #17312, opened 2020-04-21
kops grid: regenerate yaml

justinsb:
Fixed an extra quote and regenerated.

(Will look separately at how to verify generation)


justinsb:
cc @hakman and @rifelpet  - I forgot to regenerate the yaml after changing the image :-(

PR #17316, opened 2020-04-21
kops grid: Pass kops-image using dedicate flag

justinsb:
This should avoid problems with escaping spaces

### DIFF
--- a/config/jobs/kubernetes/kops/build-grid.py
+++ b/config/jobs/kubernetes/kops/build-grid.py
@@ -39,6 +39,7 @@ template = """
       - --extract=release/stable
       - --ginkgo-parallel
       - --kops-args={{kops_args}}
+      - --kops-image={{kops_image}}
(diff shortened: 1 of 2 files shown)


justinsb:
cc @hakman and @rifelpet 

The saga continues... I'll get this right eventually I hope!

Debating also dropping the interval to 1 hour so we can get signal faster.

rifelpet:
/lgtm
I agree with dropping to 1 hour

hakman:
@justinsb you could also take a look at the RHEL 7 in the mean time. Should be pretty much the same as CentOS.
https://testgrid.k8s.io/sig-cluster-lifecycle-kops#kops-grid-aws-flannel-rhel7

justinsb:
And @hakman you're right, but we do have that PR that targets only centos7; it would be great to show that fixing flannel + centos7, and then expand to cover rhel7 as well.

That said, it might be more important to just fix it than to have a nice clean demonstration on centos7!

hakman:
OK, let's see :)

PR #17317, opened 2020-04-21
Change project for Topology/CPU Manager CI jobs

vpickard:
The Topology Manager and CPU Manager CI jobs have been failing
for the past 2 weeks or so. The sad part is that the email notification
was not working. Discovered these jobs were failing from the recent
sig-node release pending failure from investigating failure and fixes
for https://github.com/kubernetes/kubernetes/issues/89892.

Signed-off-by: vpickard <vpickard@redhat.com>

### DIFF
--- a/config/jobs/kubernetes/sig-node/node-kubelet.yaml
+++ b/config/jobs/kubernetes/sig-node/node-kubelet.yaml
@@ -293,7 +293,7 @@ periodics:
       - --scenario=kubernetes_e2e
       - --
       - --deployment=node
-      - --gcp-project=k8s-jkns-ci-node-e2e
+      - --gcp-project=k8s-jkns-gke-gci-soak
       - --gcp-zone=us-west1-b
       - --node-args=--image-config-file=/workspace/test-infra/jobs/e2e_node/image-config-serial-cpu-manager.yaml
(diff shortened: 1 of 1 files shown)


spiffxp:
... we should get some more people in that OWNERS file

spiffxp:
I'd rather not pin anything more to this one project, using it for release-blocking jobs was a stopgap, I've some projects set aside in a boskos pool, one moment

derekwaynecarr:
@spiffxp agree that list needs to expand.  I will open a pr to sync it with sig node kubelet owners similar to what was done in another area of tree.

spiffxp:
>the sad part is that the email notificationwas not working.

it looks like these jobs were configured to alert two individuals; maybe it would be better to send to a sig-node mailing list?

spiffxp:
/approve
/lgtm
I'll let this merge but I'm going to have something to unpin it shortly

derekwaynecarr:
@vpickard thanks for following up from our sig call.

@spiffxp i will cc you on owners file follow-on.

spiffxp:
Followup PR to move to a boksos pool of projects: https://github.com/kubernetes/test-infra/pull/17349

PR #17319, opened 2020-04-21
Switch to the test-infra bazelbuild variant.

fejta:
This has both 2.2.0 (current) and 3.0.0 (potentially next) installed.

Tag from https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/post-test-infra-push-bazelbuild/1252675046963417088

/assign @chases2 @clarketm @cjwagner

### DIFF
--- a/config/jobs/kubernetes/test-infra/test-infra-trusted.yaml
+++ b/config/jobs/kubernetes/test-infra/test-infra-trusted.yaml
@@ -108,7 +108,7 @@ postsubmits:
     spec:
       serviceAccountName: pusher
       containers:
-      - image: gcr.io/k8s-testimages/bazelbuild:v20200420-42ca103-2.2.0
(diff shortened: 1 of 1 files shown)


fejta:
Also this doesn't really change much in the image -- the primary change is to make the variant name "test-infra" rather than "2.2.0"

This way when we update the image to have 3.0.0 and 2.2.0 instead of 2.2.0 and 2.0.0 it will automatically get pushed and the bumper will automatically bump it without us having to do this manually.

Coupling the image to a repo/branch (aka test-infra master) makes more sense than a bazel release (this will change in test-infra, and when test-infra changes it we'll want these images to get updated accordingly)

fejta:
"doesn't really change much" => clearly taunting the gods to make this a breaking change

PR #17320, opened 2020-04-21
Allow milestone-maintainers to restart jobs on Prow UI

dims:
Let's please allow more folks to re-run jobs as needed. picked
milestone-maintainers as i am on it :)

### DIFF
--- a/config/prow/config.yaml
+++ b/config/prow/config.yaml
@@ -67,6 +67,7 @@ deck:
     '*':
       github_team_ids:
       - 2009231 # test-infra-admins
+      - 2460384 # milestone-maintainers
 
 prowjob_namespace: default
 pod_namespace: test-pods


config/prow/config.yaml
```diff
@@ -67,6 +67,7 @@ deck:
     '*':
       github_team_ids:
       - 2009231 # test-infra-admins
+      - 2460384 # milestone-maintainers
```
fejta, line 70: I'm going to merge this for now because it isn't 100% obvious to me why this shouldn't be more broadly accessible. However can you consider scoping this down to the job/jobs you manage rather than everything?

fejta: Moving bigger discussion to https://github.com/kubernetes/test-infra/issues/17321

PR #17328, opened 2020-04-22
Rollout containerd to scalability tests

jprzychodzen:
/sig scalability
/assing @mm4tt

Rollout containerd tests following [experiments procedure](https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/docs/experiments.md).

Ref. kubernetes/perf-tests#1110

### DIFF
--- a/config/jobs/kubernetes/sig-scalability/sig-scalability-periodic-jobs.yaml
+++ b/config/jobs/kubernetes/sig-scalability/sig-scalability-periodic-jobs.yaml
@@ -92,6 +92,7 @@ periodics:
     preset-k8s-ssh: "true"
     preset-dind-enabled: "true"
     preset-e2e-kubemark-common: "true"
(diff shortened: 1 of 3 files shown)


config/jobs/kubernetes/sig-scalability/sig-scalability-periodic-jobs.yaml
```diff
@@ -92,6 +92,7 @@ periodics:
     preset-k8s-ssh: "true"
     preset-dind-enabled: "true"
     preset-e2e-kubemark-common: "true"
+    preset-e2e-scalability-containerd: "true"
```
wojtek-t, line 95: containerd doesn't make much sense for kubemark - this will only determine what is running in the base cluster (as we don't run containerd in kubemark cluster). So we don't really care about it.

I would prefer not to touch kubemark wrt containerd.

PR #17331, opened 2020-04-22
kops grid: don't run tests all at the same time

justinsb:
Spread them randomly over a 24 hour window using the cron spec, and hash
values derived from their name.

Also boost some tests of interest to run hourly, currently the flannel
on rhel tests.

### DIFF
--- a/config/jobs/kubernetes/kops/build-grid.py
+++ b/config/jobs/kubernetes/kops/build-grid.py
@@ -46,7 +46,7 @@ template = """
       - --provider=aws
       - --test_args={{test_args}}
       - --timeout=60m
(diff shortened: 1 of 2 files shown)


justinsb:
cc @hakman and @rifelpet 

Going hourly on a "hotlist" of jobs, and fixing the other jobs to be randomly distributed over 24 hours (which does mean they run less often, but maybe we can add a few more?)

hotlist == flannel on centos7, rhel7, rhel8 for now, but I imagine the hotlist will be something we change more often.

rifelpet:
this is great, it'd be nice to see this eventually extended to the remaining e2e jobs as well.

/lgtm

PR #17334, opened 2020-04-22
prow/plugins/label: consider only comment creation actions

nikhita:
In the current case, the plugin also considers Edit actions. This means
that if a comment body contained a command like `/remove-sig foo`, the
plugin will try to remove the `sig/foo` label when the comment body is
edited.

Since the `sig/foo` label would have been removed during comment creation,
the second removal during comment edit would fail with a comment, causing
spam.

See examples of spam in:
  - https://github.com/kubernetes/community/issues/4736 (due to edit of https://github.com/kubernetes/community/issues/4736#issuecomment-617791760)
  - https://github.com/kubernetes/k8s.io/issues/782 (due to edit of top-level body)

/assign @cblecker 
cc @mrbobbytables @bartsmykla

### DIFF
--- a/prow/plugins/label/label_test.go
+++ b/prow/plugins/label/label_test.go
@@ -575,6 +575,17 @@ func TestLabel(t *testing.T) {
 			expectedBotComment:    false,
 			action:                github.GenericCommentActionDeleted,
 		},
+		{
+			name:                  "Don't take action while editing body",
+			body:                  "/kind bug",
+			repoLabels:            []string{labels.Bug},
+			issueLabels:           []string{},
+			expectedNewLabels:     []string{},
+			expectedRemovedLabels: []string{},
+			commenter:             orgMember,
+			expectedBotComment:    false,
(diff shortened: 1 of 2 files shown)


prow/plugins/label/label_test.go
```diff
@@ -575,6 +575,17 @@ func TestLabel(t *testing.T) {
...
+		{
+			name:                  "Don't take action while editing body",
+			body:                  "/kind bug",
+			repoLabels:            []string{labels.Bug},
+			issueLabels:           []string{},
+			expectedNewLabels:     []string{},
+			expectedRemovedLabels: []string{},
+			commenter:             orgMember,
+			expectedBotComment:    false,
+			action:                github.GenericCommentActionEdited,
```
nikhita, line 587: changed this to `GenericCommentActionEdited` from `GenericCommentActionDeleted`

Bad copy paste :see_no_evil:

PR #17343, opened 2020-04-22
Move prow cluster configuration under configs dir

fejta:
This will prevent a prow bump from generating new prow images.

/hold


fejta:
/assign @michelle192837 

I'm doing a fair amount of due diligence here (grepping for prow/cluster and ensuring it has the config prefix, building and testing everything) but this is likely to break a push/bump/deploy job.

michelle192837:
I think this is okay, but I'm also curious if we have guidance anywhere about where job vs. cluster config should live in a repo that needs to be updated?

/lgtm

fejta:
Not sure what you mean? You mean where should we put the cluster files and where should we put the job configs?

michelle192837:
Yup. Basically, do we already recommend somewhere that 'job configs should live here, and cluster config should live here'?

fejta:
I don't think so... And the primary problem here is that we're conflating changes to prow binaries (aka a prow/crier/foo.go change) and changes to the deployments (prow/cluster/crier_deployment.yaml)

Most repos won't have this issue since the code doesn't live in their repo

prow/cmd/autobump/bump.sh
```diff
@@ -137,7 +137,7 @@ list-options() {
 
 upstream-version() {
  local branch="https://raw.githubusercontent.com/kubernetes/test-infra/master"
- local file="prow/cluster/deck_deployment.yaml"
+ local file="config/prow/cluster/deck_deployment.yaml"
```
cjwagner, line 140: We need to copy the `deck_deployment.yaml` file instead of moving it and keep the original file for a while to allow the old version of the autobump utility to bump itself to this version.

cjwagner: A temporary symlink might be a better option.

michelle192837:
Good points, thanks for explaining!

fejta:
So the silliest part here is that merging a change to prow/cluster/foo.yaml causes the push-prow job to create a new image (since something in prow changed) which causes autobump to create a PR to update prow/cluster/foo.yaml even though nothing has actually changed.

Now we'll only get new prow releases when prow code changes.

prow/cmd/autobump/bump.sh
fejta: Does the autobump utility update the autobump job?

cjwagner: Yes, it bumps itself and other jobs that use Prow component images like branchprotector, peribolos, checkconfig, etc. Example: https://github.com/istio/test-infra/blob/3a14b656a0578216498de2d5371b2948aaecb0f0/prow/cluster/jobs/istio/test-infra/istio.test-infra.trusted.master.yaml#L97-L102

PR #17346, opened 2020-04-22
Delete default_decoration_config.

fejta:
Everyone should be using `default_decoration_configs['*']` by now.


prow/config/config.go
```diff
@@ -1142,21 +1137,6 @@ func defaultPeriodics(periodics []Periodic, c *Config) error {
...
-		}
-
-		if len(c.Plank.DefaultDecorationConfigs) == 0 {
-			return errors.New("both default_decoration_config and default_decoration_configs['*'] are missing")
-
-		}
-
 		if _, ok := c.Plank.DefaultDecorationConfigs["*"]; !ok {
 			return errors.New("default_decoration_configs['*'] is missing")
 		}
```
cjwagner, line 1142: It looks like we only validate the config at the "*" key (below this line) instead of validating the config at all provided keys. That would be a good thing to fix here as well.

fejta: Done

PR #17347, opened 2020-04-22
Add deck_deployment.yaml fake

fejta:
This allow old autobumper images to find the new location, included in https://github.com/kubernetes/test-infra/pull/17338

/assign @cjwagner


cjwagner:
/retest
I don't think this will actually work as a symlink. The autobumper GETs the raw file contents from GH with `curl` instead of cloning the repo with `git`.
```bash
$ curl https://raw.githubusercontent.com/fejta/test-infra/sym/prow/cluster/deck_deployment.yaml
../../config/prow/cluster/deck_deployment.yaml
```

fejta:
Okay so I'm just going to trick autobump to read a later version for awhile

PR #17352, opened 2020-04-23
kops grid: start to populate the grid

justinsb:
We don't have a lot of data to go on at the moment when it comes to
CNI providers.  flannel is exhibiting some problems (at least with
webhooks, possibly more, on rhel family distros), but we don't really
know whether that is specific to flannel.

Start to fill out the grid with a few more networking plugins, mostly
chosen as the starting set because they are the ones I am most
familiar with.

We're starting off running these once per day by default, we'll drop
it to weekly after a few days.


justinsb:
cc @rifelpet and @hakman 

Now I've opened this can of worms, I'm now if anything more confused about what we should do about flannel.  Hoping that if we have data across the board, we can understand how flannel compares to other CNIs.

Would be good to also throw in some older k8s versions into the grid (k116, k117, k118?).  But I think we need to go weekly-by-default first, so I was hoping we could get some data before tomorrow's kops office hours with this PR...

rifelpet:
sounds good to me, hopefully the prow jobs are configured correctly for their one and only run before office hours tomorrow :)

/lgtm

justinsb:
/retest

Github gave 504s

justinsb:
Also I'm thinking we should use "k8s116", "k8s117", "k8s118" etc so we can also have another axis for "kops116", "kops117", "kops118" etc :-)  So many tests!

rifelpet:
I like that idea. hopefully my screen is big enough for all these testgrid tabs

hakman:
I like the idea also, but I think we should think about how manageable this is to keep things working for all everywhere and N versions of Kops supported.
I feel that we may need a better strategy to get people to adopt some CNI plugin or piece of code.

justinsb:
I don't disagree @hakman - we had a similar discussion in sig-cluster-lifecycle actually.  My view is that we should get the data and try to be informed by that data; it won't be persuasive or fair otherwise (and the situation may change; e.g. we've observed flannel going from bad to good to less-good).

PR #17355, opened 2020-04-23
Kops - Update all Ubuntu image versions

hakman:
Ubuntu 20.04 was officially released today so it deserves an update together with all its older brothers. This will give us a chance to see how stable it is and how CNI plugins work with it until tomorrow.
:)

/cc @rifelpet


rifelpet:
🎉 
/lgtm
/approve

PR #17365, opened 2020-04-23
Default horologium --dry-run=true

fejta:


petr-muller:
/hold

Looks like this needs fixing.

fejta:
/hold cancel

Updated the bazel rules

fejta:
Passing tests now! Can I get another review?

PR #17367, opened 2020-04-24
apply decoration config defaults in validation

droslean:
follow-up of https://github.com/kubernetes/test-infra/pull/17346

The decoration configs were validating without inheriting the default values. That cause the `checkconfig` to exit with failure.

/cc @fejta @stevekuznetsov @alvaroaleman

### DIFF
--- a/prow/config/config.go
+++ b/prow/config/config.go
@@ -1137,12 +1137,13 @@ func defaultPeriodics(periodics []Periodic, c *Config) error {
 func (c *Config) finalizeJobConfig() error {
 	if c.decorationRequested() {
 
-		if _, ok := c.Plank.DefaultDecorationConfigs["*"]; !ok {
+		def, ok := c.Plank.DefaultDecorationConfigs["*"]
+		if !ok {
 			return errors.New("default_decoration_configs['*'] is missing")
 		}
 
 		for key, valCfg := range c.Plank.DefaultDecorationConfigs {
-			if err := valCfg.Validate(); err != nil {
+			if err := valCfg.ApplyDefault(def).Validate(); err != nil {
 				return fmt.Errorf("default_decoration_configs[%q]: validation error: %v", key, err)
 			}
 		}
--- a/prow/config/config_test.go
+++ b/prow/config/config_test.go
@@ -333,11 +333,11 @@ periodics:
 			expectError: true,
 		},
 		{
-			name: "with bad repo config",
+			name: "repo should inherit from default config",
 			rawConfig: `
 plank:
   default_decoration_configs:
-    '*': # good
+    '*':
       timeout: 2h
       grace_period: 15s
       utility_images:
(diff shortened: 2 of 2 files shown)


prow/config/config.go
```diff
@@ -1141,8 +1141,9 @@ func (c *Config) finalizeJobConfig() error {
 			return errors.New("default_decoration_configs['*'] is missing")
 		}
 
+		def := c.Plank.DefaultDecorationConfigs["*"]
```
bbguimaraes, line 1144: This could reuse the lookup above.

prow/config/config_test.go
```diff
@@ -333,11 +333,11 @@ periodics:
 			expectError: true,
 		},
 		{
-			name: "with bad repo config",
+			name: "repo should inherit from default config",
```
alvaroaleman, line 336: Why are we changing a test rather than adding a new one?

prow/config/config_test.go
```diff
@@ -351,14 +351,14 @@ plank:
...
-    'org/bad': # bad
+    'org/inherit':
       timeout: 2h
       grace_period: 15s
       utility_images:
       # clonerefs: "clonerefs:default"
-        initupload: "initupload:default"
-        entrypoint: "entrypoint:default"
-        sidecar: "sidecar:default"
+      # initupload: "initupload:default"
```
alvaroaleman, line 359: Why do we comment this out rather than removing it?

prow/config/config_test.go
droslean: because this test case is wrong.

prow/config/config_test.go
droslean: I kept the old practicing. I guess this is a better way to see what is missing.

prow/config/config_test.go
droslean: also, this test case is the reason why #17346 passed.

alvaroaleman: > also, this test case is the reason why #17346 passed.

Soryr, I don't understand? I think the reason #17346 passed is that there was no test for the inheriting

prow/config/config_test.go
alvaroaleman: That is confusing,  you could also write it as `utility_images: {}`.

Also, can you please squash your commits?

prow/config/config_test.go
droslean: Indeed. If there were tests for the inheritance it would have fail.

prow/config/config_test.go
droslean: addressed

droslean:
>  is it intended that we only merge the top level decoration config and not do the merging hierarchically, i.E. default to *, default to org?

We do this when the decoration config is being used. I guess we should try to do the same with the validation as well. Let me check that.
/hold

droslean:
/hold cancel
This should be a follow-up if its needed.

PR #17372, opened 2020-04-24
kops grid: limit length of cluster names

justinsb:
We currently have an issue with really long cluster names; keep the names
shorter both for sanity in future, and to avoid the issue while we fix it.

Includes #17370


hakman:
@justinsb maybe also drop the `-aws` from name? We can safely assume that all tests are aws unless `gce` is present.

justinsb:
@hakman we're definitely going to have to keep abbreviating...  Good idea to drop `-aws` though, because the limit is specific to aws (or at least _this_ limit is!)  Sent #17437

hakman:
Great @justinsb. I think this will need a final rebase once the `-aws` one is merged.

justinsb:
Thanks @hakman - rebased!

PR #17374, opened 2020-04-24
Initially skip CI for Draft PRs

### LINKED ISSUE #17187: Allow disabling CI for WIP PRs

A way to disable CI for WIP PRs.

**Why is this needed**:

I've noticed in several WIP PRs (including my own) that there is a lot of tests being run without any use to it. Not only do they spam notifications but they also waste resources.

Additionally it would make reviewing and discussing a lot easier as one doesn't have to scroll over a huge amount of CI comments.

To fix this for one of my current PRs i tried to set the not-ok-to-test label, which didn't work.

Edit: Just to clarify, I'm not asking for CI to be generally disabled for WIP PRs, but to allow disabling trough a label in cases where author /participants see fit.

kwiesmueller:
This PR will skip CI for Draft PRs.
- `/test` commands should still work.
- Adding the `ok-to-test` label for untrusted users, should still trigger the tests.

Fixes #17187

/cc @alvaroaleman @fejta

### DIFF
--- a/prow/plugins/trigger/pull-request.go
+++ b/prow/plugins/trigger/pull-request.go
@@ -76,8 +76,13 @@ func handlePR(c Client, trigger plugins.Trigger, pr github.PullRequestEvent) err
 			return fmt.Errorf("could not check membership: %s", err)
 		}
 		if member {
+			// dedicated draft check for create to comment on the PR
+			if pr.PullRequest.Draft {
+				c.Logger.Info("Skipping all jobs for draft PR.")
+				return draftMsg(c.GitHubClient, pr.PullRequest)
+			}
 			c.Logger.Info("Starting all jobs for new PR.")
-			return buildAll(c, &pr.PullRequest, pr.GUID, baseSHA, presubmits)
+			return buildAllButDrafts(c, &pr.PullRequest, pr.GUID, baseSHA, presubmits)
 		}
 		c.Logger.Infof("Welcome message to PR author %q.", author)
 		if err := welcomeMsg(c.GitHubClient, trigger, pr.PullRequest); err != nil {
@@ -98,7 +103,7 @@ func handlePR(c Client, trigger plugins.Trigger, pr github.PullRequestEvent) err
 				}
 			}
 			c.Logger.Info("Starting all jobs for updated PR.")
-			return buildAll(c, &pr.PullRequest, pr.GUID, baseSHA, presubmits)
+			return buildAllButDrafts(c, &pr.PullRequest, pr.GUID, baseSHA, presubmits)
 		}
 	case github.PullRequestActionEdited:
 		// if someone changes the base of their PR, we will get this
@@ -132,7 +137,7 @@ func handlePR(c Client, trigger plugins.Trigger, pr github.PullRequestEvent) err
 				return fmt.Errorf("could not validate PR: %s", err)
 			} else if !trusted {
 				c.Logger.Info("Starting all jobs for untrusted PR with LGTM.")
-				return buildAll(c, &pr.PullRequest, pr.GUID, baseSHA, presubmits)
+				return buildAllButDrafts(c, &pr.PullRequest, pr.GUID, baseSHA, presubmits)
 			}
 		}
 		if pr.Label.Name == labels.OkToTest {
@@ -147,7 +152,7 @@ func handlePR(c Client, trigger plugins.Trigger, pr github.PullRequestEvent) err
 				c.Logger.Debug("Label added by the bot, skipping.")
 				return nil
 			}
-			return buildAll(c, &pr.PullRequest, pr.GUID, baseSHA, presubmits)
+			return buildAllButDrafts(c, &pr.PullRequest, pr.GUID, baseSHA, presubmits)
 		}
 	case github.PullRequestActionClosed:
 		if err := abortAllJobs(c, &pr.PullRequest); err != nil {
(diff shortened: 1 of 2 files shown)


prow/plugins/trigger/pull-request_test.go
```diff
@@ -289,95 +317,98 @@ func TestHandlePullRequest(t *testing.T) {
...
-			PullRequests: map[int]*github.PullRequest{
-				0: {
-					Number: 0,
-					User:   github.User{Login: tc.Author},
-					Base: github.PullRequestBranch{
-						Ref: "master",
-						Repo: github.Repo{
-							Owner: github.User{Login: "org"},
-							Name:  "repo",
+		t.Run(tc.name, func(t *testing.T) {
```
kwiesmueller, line 320: I added this as it was hard to see which test failed.
Would be ok with reverting this again.

prow/plugins/trigger/pull-request.go
```diff
@@ -297,6 +304,13 @@ I understand the commands that are listed [here](https://go.k8s.io/bot-commands?
 	return nil
 }
 
+func draftMsg(ghc githubClient, trigger plugins.Trigger, pr github.PullRequest) error {
+	org, repo, _ := orgRepoAuthor(pr)
+
+	comment := `Skipping CI for Draft PR`
```
kwiesmueller, line 310: Not sure if this comment is enough :thinking:

prow/plugins/trigger/pull-request_test.go
alvaroaleman: Please do keep it :)

prow/plugins/trigger/pull-request.go
alvaroaleman: Maybe extend with `If you want CI signal for your change, please convert it to an actual PR" (is there a better term for "non-draft-pr" than "actual pr"?)

prow/plugins/trigger/pull-request.go
```diff
@@ -225,6 +228,10 @@ func buildAllIfTrusted(c Client, trigger plugins.Trigger, pr github.PullRequestE
 	if err != nil {
 		return fmt.Errorf("could not validate PR: %s", err)
 	} else if trusted {
+		if pr.PullRequest.Draft && !github.HasLabel(labels.OkToTest, l) {
```
alvaroaleman, line 231: I think we should remove the second condition, because even if the PR had `ok-to-test` it wouldn't get built as long as its a draft

prow/plugins/trigger/pull-request_test.go
```diff
@@ -132,6 +133,15 @@ func TestHandlePullRequest(t *testing.T) {
 			ShouldBuild: true,
 			prAction:    github.PullRequestActionOpened,
 		},
+		{
+			name: "Trusted user open draft PR should not build",
```
alvaroaleman, line 137: Please extend the test name with ` and should comment`

prow/plugins/trigger/pull-request.go
kwiesmueller: Sure, i wanted to keep the ok-to-test idea so when the label is there, the tests run no matter what.
But i'm okay with both.

fejta: The `ok-to-test` label has no relevance to org members. It only changes behavior for people foreign to the org.

prow/plugins/trigger/pull-request.go
fejta: Also tell people that they can manually trigger with `/test all` (that will work, right?)

What is your thinking for using use ``` ` ``` here instead of `"`?

kwiesmueller: I started from the welcomeMsg until i saw that i don't need most of it.
Just forgot to change the `

prow/plugins/trigger/pull-request.go
```diff
@@ -76,6 +76,9 @@ func handlePR(c Client, trigger plugins.Trigger, pr github.PullRequestEvent) err
 			return fmt.Errorf("could not check membership: %s", err)
 		}
 		if member {
```
alvaroaleman, line 78: Wouldn't it make more sense to to write a wrapper around `buildAll` and use it for all pull request events? The current approach means that we omit the building on drafts only for certain events where we implemented it, but not for others (reopening a PR, pushing to a PR, Labeling a PR, PR base change)

kwiesmueller: The only one currently not implemented is Reopen.
I left that out on purpose with a reason that doesn't make sense, now that you mention it.
There already is buildAllIfTrusted which sounds like it, but i'm not sure why it only gets used in some places right now.
I'll look at it again.

fejta: How are things going?

kwiesmueller: I didn't have much time since last week, I'll have some more this one.
Can't wait to get this done.

kwiesmueller: Okay, i cleaned it up a little.
There still is an additional draft check for create, to comment.
I decided to only send the skip comment on create to not spam it. If you want to, i can also send it on comment.

kwiesmueller: Logging might be confusing now, as it logs that all jobs get started and then if the PR is a draft, that they get skipped.
I wouldn't see it as problem, but i can move the log message to buildAllButDrafts which would loose some details due to the different logs messages before buildAll.

alvaroaleman:
looks good, thanks for your work on this!
/hold cancel

PR #17379, opened 2020-04-24
Leave a comment on PRs that fail to merge by tide

adshmh:
Tide now leaves a comment on any PR that fails to merge for any reason.
Existing comments on the PR are checked to avoid spamming the PR with
merge failure comments.

### DIFF
--- a/prow/tide/tide.go
+++ b/prow/tide/tide.go
@@ -61,6 +61,9 @@ type githubClient interface {
 	GetRepo(owner, name string) (github.FullRepo, error)
 	Merge(string, string, int, github.MergeDetails) error
 	Query(context.Context, interface{}, map[string]interface{}) error
+	BotUserChecker() (func(candidate string) bool, error)
+	CreateComment(string, string, int, string) error
+	ListIssueComments(org, repo string, number int) ([]github.IssueComment, error)
 }
 
 type contextChecker interface {
@@ -88,7 +91,8 @@ type Controller struct {
 	// Cache entries expire if they are not used during a sync loop.
 	changedFiles *changedFilesAgent
 
-	mergeChecker *mergeChecker
+	mergeChecker     *mergeChecker
+	failureCommenter failureCommenter
 
 	History *history.History
 }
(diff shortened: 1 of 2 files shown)


alvaroaleman:
I think this would be useful. @cjwagner I remember you did not want this in order to allow GitHub to debug the problem. Did that happen?

/assign @cjwagner

prow/tide/tide.go
```diff
@@ -49,8 +49,12 @@ import (
 	"k8s.io/test-infra/prow/tide/history"
 )
 
-// For mocking out sleep during unit tests.
-var sleep = time.Sleep
+var (
+	// For mocking out sleep during unit tests.
+	sleep = time.Sleep
+
+	mergeFailureComment = "attempt to merge PR failed. Tide may retry depending on the type of failure"
```
petr-muller, line 56: It looks like we would end up commenting this string verbatim, so it should start with a capital letter and end with dot.

petr-muller (commented):
Code LGTM

adshmh:
> I am a little concerned about one scenario though. Repos that require the `tide` status context itself via branch protection often see transient merge failures when the sync controller tries to merge before the status controller sets the `tide` status to passing. Requiring the `tide` status context is not recommended because of this, but I think some repos do.

> I'm also not sure how helpful this will be without more specific and actionable information in the error comment.

Thank you for the review. I have added:
- A check to avoid commenting if the repo requires the tide context.
- The error details to the comment left on the PR.

prow/tide/tide.go
adshmh: Thank you for the review. Fixed.

prow/tide/tide.go
```diff
@@ -1824,3 +1850,54 @@ func nonFailedBatchByNameBaseAndPullsIndexFunc(obj runtime.Object) []string {
...
+			return true, nil
+		}
+	}
+	return false, nil
+}
+
+func (a *mergeFailureCommenter) CreateMergeFailureComment(pr *PullRequest, err error) error {
+	message := mergeFailureComment
+	if err != nil {
+		message = fmt.Sprintf("%s The following are the error details:\n```\n%s\n```\n", mergeFailureComment, err)
```
stevekuznetsov, line 1895: I vaguely remember this markdown formatting sort of thing not working as expected when posting via the API - can we use `<code>`?

prow/tide/tide.go
```diff
@@ -49,8 +49,12 @@ import (
 	"k8s.io/test-infra/prow/tide/history"
 )
 
-// For mocking out sleep during unit tests.
-var sleep = time.Sleep
+var (
+	// For mocking out sleep during unit tests.
+	sleep = time.Sleep
+
+	mergeFailureComment = "Attempt to merge the PR failed. Tide may retry depending on the type of failure."
```
stevekuznetsov, line 56: Is it not possible to know whether or not we will retry? I worry that this will cause a larger support load ("I'm a user, and I'm seeing this vague comment, what does it mean?")

prow/tide/tide.go
adshmh: Sorry for the delay, I missed a few updates from this repo. All reviews are now addressed.

prow/tide/tide.go
adshmh: Thank you for the review. Fixed.

PR #17381, opened 2020-04-24
bugzilla plugin: handle cherrypick bug creation

AlexNPavel:
This PR adds support for the creation of new cherrypick bugs for new PRs made by the cherrypick plugin.

/cc @stevekuznetsov


prow/bugzilla/client.go
```diff
@@ -225,7 +225,17 @@ func (c *client) CloneBug(bug *Bug) (int, error) {
 	if err != nil {
 		return 0, fmt.Errorf("failed to get parent bug's comments: %v", err)
 	}
-	return c.CreateBug(cloneBugStruct(bug, comments))
+	id, err := c.CreateBug(cloneBugStruct(bug, comments))
```
stevekuznetsov, line 228: Is this re-entrant? If we fail to update and someone tries to do another `/bugzilla refresh` we will fail here as the cloned bug already exists, no?

prow/plugins/bugzilla/bugzilla.go
```diff
@@ -228,8 +231,29 @@ func digestPR(log *logrus.Entry, pre github.PullRequestEvent, validateByDefault
...
+		cherrypickOf, err := strconv.Atoi(cherrypickMatch[1])
+		if err != nil {
+			// should be impossible based on the regex
+			log.WithError(err).Debug("Failed to parse cherrypick bugID as int - is the regex correct?")
+			return nil, err
+		}
+		e.cherrypickFromPRNum = cherrypickOf
+		cherrypickTo := cherrypickTitleMatch.FindStringSubmatch(pre.PullRequest.Title)
+		if cherrypickTo == nil {
+			log.Print("HELLO")
```
stevekuznetsov, line 248: oops :)

prow/plugins/bugzilla/bugzilla.go
```diff
@@ -228,8 +231,29 @@ func digestPR(log *logrus.Entry, pre github.PullRequestEvent, validateByDefault
...
+	if cherrypickMatch != nil {
+		e.cherrypick = true
+		cherrypickOf, err := strconv.Atoi(cherrypickMatch[1])
+		if err != nil {
+			// should be impossible based on the regex
+			log.WithError(err).Debug("Failed to parse cherrypick bugID as int - is the regex correct?")
+			return nil, err
+		}
+		e.cherrypickFromPRNum = cherrypickOf
+		cherrypickTo := cherrypickTitleMatch.FindStringSubmatch(pre.PullRequest.Title)
```
stevekuznetsov, line 246: No need for regex -- this is `pre/PullRequest.Base.Ref`

prow/plugins/bugzilla/bugzilla.go
```diff
@@ -43,6 +43,9 @@ var (
 	refreshCommandMatch  = regexp.MustCompile(`(?mi)^/bugzilla refresh\s*$`)
 	qaAssignCommandMatch = regexp.MustCompile(`(?mi)^/bugzilla assign-qa\s*$`)
 	qaReviewCommandMatch = regexp.MustCompile(`(?mi)^/bugzilla cc-qa\s*$`)
+	cherrypickPRMatch    = regexp.MustCompile(`This is an automated cherry-pick of #([0-9]+)`)
```
stevekuznetsov, line 46: can we factor this out into a constant to share with the cherrypicker

prow/plugins/bugzilla/bugzilla.go
```diff
@@ -819,6 +850,53 @@ func handleMerge(e event, gc githubClient, bc bugzilla.Client, options plugins.B
...
 
+func handleCherrypick(e event, gc githubClient, bc bugzilla.Client, options plugins.BugzillaBranchOptions, log *logrus.Entry) error {
+	comment := e.comment(gc)
+	// get the info for the PR being cherrypicked from
+	pr, err := gc.GetPullRequest(e.org, e.repo, e.cherrypickFromPRNum)
+	if err != nil {
+		log.WithError(err).Warn("Unexpected error getting title of pull request being cherrypicked from.")
+		return comment(formatError(fmt.Sprintf("checking the state of a cherrypicked pull request at https://github.com/%s/%s/pull/%d", e.org, e.repo, e.cherrypickFromPRNum), bc.Endpoint(), e.bugId, err))
+	}
+	// Attempt to identify bug from PR title
```
stevekuznetsov, line 861: please factor this out to DRY it

prow/plugins/bugzilla/bugzilla.go
```diff
@@ -819,6 +850,53 @@ func handleMerge(e event, gc githubClient, bc bugzilla.Client, options plugins.B
...
+		// should be impossible based on the regex
+		log.WithError(err).Debug("Failed to parse bug ID as int - is the regex correct?")
+		return comment(formatError(fmt.Sprintf("Failed to parse bug ID (%s) as int", mat[1]), bc.Endpoint(), e.bugId, err))
+	}
+	bug, err := getBug(bc, bugID, log, comment)
+	if err != nil {
+		log.WithError(err).Debugf("Failed to get bug %d", bugID)
+		return comment(formatError(fmt.Sprintf("Failed to get bug %d", bugID), bc.Endpoint(), e.bugId, err))
+	}
+	cloneID, err := bc.CloneBug(bug)
```
stevekuznetsov, line 878: what if the bug already has been cloned? we do not want to clone again nor fail

prow/plugins/bugzilla/bugzilla.go
```diff
@@ -819,6 +850,53 @@ func handleMerge(e event, gc githubClient, bc bugzilla.Client, options plugins.B
...
+		Version: *options.TargetRelease,
+	}
+	err = bc.UpdateBug(cloneID, update)
+	if err != nil {
+		log.WithError(err).Debugf("Unable to update version and dependencies for bug %d", cloneID)
+		return comment(formatError(fmt.Sprintf("Failed to update version and dependencies for bugzilla bug %d", cloneID), bc.Endpoint(), e.bugId, err))
+	}
+	// Replace old bugID in title with new cloneID
+	newTitle := strings.Replace(e.body, fmt.Sprintf("Bug %d", bugID), fmt.Sprintf("Bug %d", cloneID), 1)
+	response := fmt.Sprintf("Bug %d has been cloned as Bug %d. Retitling PR to link against new bug.\n/retitle %s", bugID, cloneID, newTitle)
```
stevekuznetsov, line 896: Not `Bug %d` but use the helper to create links

prow/bugzilla/fake.go
```diff
@@ -175,7 +189,17 @@ func (c *Fake) CloneBug(bug *Bug) (int, error) {
 	if err != nil {
 		return 0, fmt.Errorf("failed to get parent bug's comments: %v", err)
 	}
-	return c.CreateBug(cloneBugStruct(bug, comments))
+	id, err := c.CreateBug(cloneBugStruct(bug, comments))
```
stevekuznetsov, line 192: Does this set the version correctly the first time around?

prow/plugins/bugzilla/bugzilla_test.go
```diff
@@ -22,6 +22,7 @@ import (
 	"strconv"
 	"testing"
 
+	"github.com/google/go-cmp/cmp"
 	"github.com/sirupsen/logrus"
 	"k8s.io/apimachinery/pkg/util/diff"
```
stevekuznetsov, line 27: If you are moving to `cmp` in one place, mind moving them all in this file?

prow/bugzilla/fake.go
AlexNPavel: Nope. Need to add that to the update var.

prow/bugzilla/client.go
AlexNPavel: We identify bugs via the PR title. If we fail to clone and properly configure the bug, the bug number won't be added to the title. So I don't think that this should cause problems (other than a bug floating around in bugzilla). The user would have to do the cloning of the bug and retitling of the PR themselves in that case.

Also, this functionality isn't used by the `/bugzilla refresh` command (at least not yet). It only runs when a PR is opened with the correct description. We could also add functionality to handle on `/bugzilla refresh` instead, but that would be slightly more complicated (we would need to check whether the bug was created by the cherrypicker on every `/bugzilla refresh` and if yes, then identify what target release the bug has, creating a clone if it doesn't match the branch's target release).

prow/plugins/bugzilla/bugzilla.go
AlexNPavel: To prevent from cloning again, we could use the bugzilla search API to find bugs with the correct summary:

If the ID is different from the parent ID but the version is the same, then that bug is a clone that failed to have its version updated.

If it has the correct target release, we just failed to link it somehow, and we should link it here.

If it has anything else as the target release, it's probably a different cherry pick that we should ignore.

If we don't match anything above, that means we haven't cloned it yet, so we should just proceed to clone normally.

prow/bugzilla/client.go
stevekuznetsov: Ah, ok. I think we will get asked in the future to help devs make the cherry-pick automatically if we happen to fail on the first attempt, but I agree that as written this does not need to be re-entrant.

prow/plugins/bugzilla/bugzilla.go
stevekuznetsov: Since this code does not need to be re-entrant I think we can forgo this

prow/plugins/bugzilla/bugzilla.go
AlexNPavel: Not sure what we could share here? This is a regex and the cherrypicker plugin has an `Sprintf` string. Is there a way to combine an `Sprintf` string and regex?

prow/bugzilla/fake.go
```diff
@@ -73,6 +73,23 @@ func (c *Fake) UpdateBug(id int, update BugUpdate) error {
...
 		bug.Status = update.Status
 		bug.Resolution = update.Resolution
+		if update.Version != "" {
+			bug.Version = []string{update.Version}
+		}
+		if update.DependsOn != nil {
+			if len(update.DependsOn.Set) > 0 {
+				bug.DependsOn = update.DependsOn.Set
+			} else {
+				dependSet := sets.NewInt(bug.DependsOn...)
```
stevekuznetsov, line 83:
```go
bug.DependsOn = sets.NewInt(bug.DependsOn...).Insert(update.DependsOn.Add...).Delete(update.DependsOn.Remove...).List()
```

prow/bugzilla/fake.go
```diff
@@ -175,7 +193,17 @@ func (c *Fake) CloneBug(bug *Bug) (int, error) {
...
+	id, err := c.CreateBug(cloneBugStruct(bug, comments))
+	if err != nil {
+		return id, err
+	}
+	depends := BugUpdate{
+		DependsOn: &IDUpdate{
+			Add: []int{bug.ID},
+		},
+	}
+	err = c.UpdateBug(id, depends)
```
stevekuznetsov, line 205: Kind of problematic that you have to implement this in in the fake -- you are then testing that the fake is written correctly. not the production client. Let's try to factor this such that it's not the case.

prow/plugins/bugzilla/bugzilla.go
stevekuznetsov: Maybe break out `This is an automated cherry-pick of ` as a constant? Or better write a unit test that ensures that the cherry-picker function to generate a title is always matching the regex here.

prow/plugins/bugzilla/bugzilla.go
```diff
@@ -819,6 +827,65 @@ func handleMerge(e event, gc githubClient, bc bugzilla.Client, options plugins.B
 	return comment(fmt.Sprintf("%s %s\n%s", mergedMessage("Some"), unmergedMessage, outcomeMessage("")))
 }
 
+func handleCherrypick(e event, gc githubClient, bc bugzilla.Client, options plugins.BugzillaBranchOptions, log *logrus.Entry) error {
+	comment := e.comment(gc)
+	// get the info for the PR being cherrypicked from
+	pr, err := gc.GetPullRequest(e.org, e.repo, e.cherrypickFromPRNum)
+	if err != nil {
+		log.WithError(err).Warn("Unexpected error getting title of pull request being cherrypicked from.")
+		return comment(formatError(fmt.Sprintf("checking the state of a cherrypicked pull request at https://github.com/%s/%s/pull/%d", e.org, e.repo, e.cherrypickFromPRNum), bc.Endpoint(), e.bugId, err))
```
stevekuznetsov, line 836: We should make it clear what we were doing when we ran into this issue, since the user took no specific action to ask the bot to do anything here. 

```
Failed to create a cherry-pick bug in Bugzilla: could not get ...
```

prow/plugins/bugzilla/bugzilla.go
```diff
@@ -819,6 +827,65 @@ func handleMerge(e event, gc githubClient, bc bugzilla.Client, options plugins.B
...
+	// Attempt to identify bug from PR title
+	bugID, bugMissing, err := bugIDFromTitle(pr.Title)
+	if bugMissing {
+		log.Debugf("Parent PR %d doesn't have associated bug; not creating cherrypicked bug", pr.Number)
+		// if there is no bugzilla bug, we should simply ignore this PR
+		return nil
+	} else if err != nil {
+		// should be impossible based on the regex
+		log.WithError(err).Debugf("Failed to get bug ID from PR title \"%s\"", pr.Title)
+		return comment(formatError(fmt.Sprintf("Failed to get bug ID from PR title \"%s\"", pr.Title), bc.Endpoint(), 0, err))
```
stevekuznetsov, line 847: maybe we want to add that prefix to explain what we were doing to all of these comments actually

prow/plugins/bugzilla/bugzilla.go
```diff
@@ -819,6 +827,65 @@ func handleMerge(e event, gc githubClient, bc bugzilla.Client, options plugins.B
...
+	}
+	cloneLink := fmt.Sprintf(bugLink, cloneID, bc.Endpoint(), cloneID)
+	// Update the version of the bug to the target release
+	update := bugzilla.BugUpdate{
+		Version: *options.TargetRelease,
+	}
+	err = bc.UpdateBug(cloneID, update)
+	if err != nil {
+		log.WithError(err).Debugf("Unable to update version and dependencies for bug %d", cloneID)
+		return comment(formatError(fmt.Sprintf("Failed to update version and dependencies for bugzilla %s", cloneLink), bc.Endpoint(), e.bugId, err))
```
stevekuznetsov, line 868: This should make it clear that a clone bug *was* created

prow/plugins/bugzilla/bugzilla_test.go
```diff
@@ -1082,6 +1145,12 @@ Instructions for interacting with me using PR comments are available [here](http
 			}
 			e.missing = testCase.missing
 			e.merged = testCase.merged
+			e.cherrypick = testCase.cherryPick
+			e.cherrypickFromPRNum = testCase.cherryPickFromPRNum
+			e.cherrypickTo = testCase.cherryPickTo
+			if testCase.replaceBody != "" {
+				e.body = testCase.replaceBody
```
stevekuznetsov, line 1152: If we don't set `e.body` otherwise, why don't we just call it `testCase.body` ?

prow/plugins/bugzilla/bugzilla_test.go
```diff
@@ -1051,6 +1089,30 @@ Instructions for interacting with me using PR comments are available [here](http
 </details>`,
 			expectedBug: &bugzilla.Bug{ID: 123, Status: "CLOSED", Severity: "urgent"},
 		},
+		{
+			name:                "Cherrypick PR results in cloned bug creation",
```
stevekuznetsov, line 1093: There are a large number of error cases and messages that are not tested@

AlexNPavel:
/hold

There's some weird gitignore dependency problems going on

prow/bugzilla/client.go
```diff
@@ -161,13 +161,15 @@ func (c *client) UpdateBug(id int, update BugUpdate) error {
...
 
+// CreateBug creates a new bug on the server.
+// https://bugzilla.readthedocs.io/en/latest/api/core/v1/bug.html#create-bug
 func (c *client) CreateBug(bug *BugCreate) (int, error) {
 	logger := c.logger.WithFields(logrus.Fields{methodField: "CreateBug", "bug": bug})
 	body, err := json.Marshal(bug)
 	if err != nil {
 		return 0, fmt.Errorf("failed to marshal create payload: %v", err)
 	}
-	req, err := http.NewRequest(http.MethodPost, fmt.Sprintf("%s/rest/bug/", c.endpoint), bytes.NewBuffer(body))
```
stevekuznetsov, line 170: Why didn't this require a change in the unit tests for the client?

prow/plugins/bugzilla/bugzilla_test.go
```diff
@@ -1051,6 +1091,129 @@ Instructions for interacting with me using PR comments are available [here](http
...
+			name:                "parent PR of cherrypick not existing results in error",
+			bugs:                []bugzilla.Bug{{Product: "Test", Component: []string{"TestComponent"}, Version: []string{"v2"}, ID: 123, Status: "CLOSED", Severity: "urgent"}},
+			bugComments:         map[int][]bugzilla.Comment{123: {{BugID: 123, Count: 0, Text: "This is a bug"}}},
+			prs:                 []github.PullRequest{{Number: 2, Body: "This is an automated cherry-pick of #1.\n\n/assign user", Title: "[v1] " + base.body}},
+			body:                "[v1] " + base.body,
+			cherryPick:          true,
+			cherryPickFromPRNum: 1,
+			cherryPickTo:        "v1",
+			options:             plugins.BugzillaBranchOptions{TargetRelease: &v1},
+			expectedComment: `org/repo#1:@user: An error was encountered Failed to create a cherry-pick bug in Bugzilla: failed to checked the state of cherrypicked pull request at https://github.com/org/repo/pull/1 for bug 123 on the Bugzilla server at www.bugzilla:
```
stevekuznetsov, line 1128:
```
:@user: An error was encountered Failed to create a cherry-pick bug 
```

Is that formatted the way you want it to be? :)

prow/plugins/bugzilla/bugzilla_test.go
```diff
@@ -1051,6 +1091,129 @@ Instructions for interacting with me using PR comments are available [here](http
...
+> pull request number 1 does not exist
```
stevekuznetsov, line 1129: IMO this would read better as a `org/repo#1` link

prow/bugzilla/client.go
AlexNPavel: It did. Fixed in one of the force pushes I made trying to fix the bazel tests.

prow/plugins/bugzilla/bugzilla_test.go
AlexNPavel: That's just the error from the GitHub client. In this case from the fake client: https://github.com/kubernetes/test-infra/blob/master/prow/github/fakegithub/fakegithub.go#L226

AlexNPavel:
/hold cancel

I think bazel is happy again

Issue #17382, opened 2020-04-24
//experiment/autobumper/bumper unit tests fail on mac


fejta:
```
--- FAIL: TestCallWithWriter (0.02s)
    --- PASS: TestCallWithWriter/no_secret_in_stdout_are_working_well (0.00s)
    --- PASS: TestCallWithWriter/secret_in_stdout_are_censored (0.00s)
    --- FAIL: TestCallWithWriter/secret_in_stderr_are_censored (0.00s)
        bumper_test.go:110: Expected stderr 'ls: cannot access '/tmp/file-not-exist/CENSORED/CENSORED/file-not-exist': No such file or directory
            ', got 'ls: /tmp/file-not-exist/CENSORED/CENSORED/file-not-exist: No such file or directory
            '
    --- FAIL: TestCallWithWriter/no_secret_in_stderr_are_working_well (0.00s)
        bumper_test.go:110: Expected stderr 'ls: cannot access '/tmp/file-not-exist/aaa/file-not-exist': No such file or directory
            ', got 'ls: /tmp/file-not-exist/aaa/file-not-exist: No such file or directory
            '
```

fejta:
And config-bootstrapper fails with:

```
--- FAIL: TestRunV2 (0.28s)
    --- PASS: TestRunV2/issues/15570_is_covered (0.01s)
    --- FAIL: TestRunV2/multiple_sources (0.01s)
        main_test.go:237: multiple sources: incorrect ConfigMap state after update:   &v1.ConfigMap{
              	TypeMeta:   v1.TypeMeta{},
              	ObjectMeta: v1.ObjectMeta{Name: "other", Namespace: "default"},
              	Data: map[string]string{
            - 		"FOO.yaml":       "#FOO.yaml",
            + 		"other-foo.yaml": "#other-foo.yaml",
              	},
            - 	BinaryData: nil,
            + 	BinaryData: map[string][]uint8{},
              }
        main_test.go:237: multiple sources: incorrect ConfigMap state after update:   &v1.ConfigMap{
              	TypeMeta:   v1.TypeMeta{},
              	ObjectMeta: v1.ObjectMeta{Name: "bar", Namespace: "default"},
              	Data: map[string]string{
            - 		"BAR.yaml":       "#BAR.yaml",
            + 		"other-bar.yaml": "#other-bar.yaml",
              	},
            - 	BinaryData: nil,
            + 	BinaryData: map[string][]uint8{},
              }
[...]

PR #17384, opened 2020-04-24
Do not depend on case-sensitive filenames.

fejta:
Macs do not have case sensitive filenames by default

ref #17382


prow/cmd/config-bootstrapper/main_test.go
```diff
@@ -80,8 +80,8 @@ func testRun(clients localgit.Clients, t *testing.T) {
 	}
 
 	if err := lg.AddCommit("openshift", "other", map[string][]byte{
-		"config/FOO.yaml": []byte(`#FOO.yaml`),
```
fejta, line 83: This is the same file as L69

PR #17393, opened 2020-04-27
censor oauth token from the git commands

### LINKED ISSUE #17386: Github token leaked in clone-log.txt
When cloning a private repository, the Github token will appear in clear text in clone-log.txt file (uploaded in the logs bucket).

Here is an example of the log content (real token redacted):
```
# Cloning org/repo at master(3cecffcf737695e38c8aea5b6c04d24a1dd5223b)
# Checking out pulls:
#	14(43a3c6eaa8a416456619af8d5d29e36e088d40f7)
$ mkdir -p /home/prow/go/src/github.com/org/repo
$ git init
Initialized empty Git repository in /home/prow/go/src/github.com/org/repo/.git/
$ git config user.name ci-robot
$ git config user.email ci-robot@k8s.io
$ git fetch https://cf07....:x-oauth-basic@github.com/org/repo.git --tags --prune
From https://github.com/org/repo
 * branch            HEAD       -> FETCH_HEAD
$ git fetch https://cf07....:x-oauth-basic@github.com/org/repo.git master
From https://github.com/org/repo
```

It looks like the code responsible for this is here:

https://github.com/kubernetes/test-infra/blob/a5764e9eea3e155bc7084b0387479823b794baa3/prow/pod-utils/clone/clone.go#L124-L128

Is there something we can do to prevent this ?
Should this be considered a security risk ?

droslean:
closes #17386 

Even though we were censoring the OAuth token in the `clonerefs` output logs, the `Run` function was returning a `Record` that it's being used from the `clonerefs` binary to marshal the struct to a file that was being uploaded in GCS (see https://github.com/kubernetes/test-infra/blob/master/prow/clonerefs/run.go#L107-L114). That logic caused the OAuth token to leak in the artifacts.

### DIFF
--- a/prow/pod-utils/clone/clone.go
+++ b/prow/pod-utils/clone/clone.go
@@ -60,7 +60,7 @@ func Run(refs prowapi.Refs, dir, gitUserName, gitUserEmail, cookiePath string, e
 				message = err.Error()
 				record.Failed = true
 			}
-			record.Commands = append(record.Commands, Command{Command: formattedCommand, Output: output, Error: message})
+			record.Commands = append(record.Commands, Command{Command: censorGitCommand(formattedCommand, oauthToken), Output: output, Error: message})
 			if err != nil {
 				return err
 			}
@@ -89,6 +89,14 @@ func Run(refs prowapi.Refs, dir, gitUserName, gitUserEmail, cookiePath string, e
 	}
 
 	return record
+}
+
+func censorGitCommand(command, token string) string {
+	if token == "" {
+		return command
+	}
+	censored := bytes.ReplaceAll([]byte(command), []byte(token), []byte("CENSORED"))
+	return string(censored)
 }
 
 // PathForRefs determines the full path to where
--- a/prow/pod-utils/clone/clone_test.go
+++ b/prow/pod-utils/clone/clone_test.go
@@ -535,3 +535,33 @@ func makeFakeGitRepo(fakeTimestamp int) (string, error) {
 	}
 	return fakeGitDir, nil
 }
+
+func TestCensorGitCommand(t *testing.T) {
+	testCases := []struct {
+		id       string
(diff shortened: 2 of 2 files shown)


prow/pod-utils/clone/clone.go
```diff
@@ -60,7 +60,7 @@ func Run(refs prowapi.Refs, dir, gitUserName, gitUserEmail, cookiePath string, e
 				message = err.Error()
 				record.Failed = true
 			}
-			record.Commands = append(record.Commands, Command{Command: formattedCommand, Output: output, Error: message})
+			record.Commands = append(record.Commands, Command{Command: censorGitCommand(formattedCommand), Output: output, Error: message})
```
petr-muller, line 63: `censorGitCommand` takes 2 parameters

prow/pod-utils/clone/clone.go
```diff
@@ -91,6 +91,11 @@ func Run(refs prowapi.Refs, dir, gitUserName, gitUserEmail, cookiePath string, e
 	return record
 }
 
+func censorGitCommand(command, token string) string {
+	censored := bytes.ReplaceAll([]byte(command), []byte(token), []byte("CENSORED"))
```
petr-muller, line 95: This needs to happen only when `token` is not emtpy

eddycharly:
A simple question, it will only affect logs right ?
The command will still use the real token ?

droslean:
Yes.

alvaroaleman:
@eddycharly yes. If anyhow possible I recommend to use a SSH key for cloning of private repos, because a) it has a lower chance of being leaked as nothing will log it by default b) if it gets leaked, the access it gives is much more limited than the access given by an oauth token.

eddycharly:
@droslean Fantastic, thanks 👍

PR #17401, opened 2020-04-27
GitV2: make usage nice

alvaroaleman:
This PR does  a couple of things:

* Allow the gitv2 client to be used anonymously
* Make the gitv2 constructor functional opts based and default it to use github.com anonymously
* Expose a `Run` function on both the old and the new client to allow running arbitrary commands

### DIFF
--- a/prow/git/git.go
+++ b/prow/git/git.go
@@ -475,3 +475,13 @@ func (r *Repo) MergeCommitsExistBetween(target, head string) (bool, error) {
 	}
 	return len(b) != 0, nil
 }
+
+// ShowRef returns the commit for a commitlike. Unlike rev-parse it does not require a checkout.
+func (i *Repo) ShowRef(commitlike string) (string, error) {
+	i.logger.Infof("Getting the commit sha for commitlike %s", commitlike)
+	out, err := i.gitCommand("show-ref", "-s", commitlike).CombinedOutput()
+	if err != nil {
+		return "", fmt.Errorf("failed to get commit sha for commitlike %s: %v", commitlike, err)
+	}
+	return strings.TrimSpace(string(out)), nil
+}
--- a/prow/git/v2/client_factory.go
+++ b/prow/git/v2/client_factory.go
@@ -20,9 +20,11 @@ import (
 	"io/ioutil"
 	"os"
 	"path"
+	"runtime"
 	"sync"
 
 	"github.com/sirupsen/logrus"
+	utilpointer "k8s.io/utils/pointer"
 )
 
 // ClientFactory knows how to create clientFactory for repos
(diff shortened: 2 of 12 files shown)


prow/git/git.go
```diff
@@ -245,6 +245,11 @@ func (r *Repo) gitCommand(arg ...string) *exec.Cmd {
 	return cmd
 }
 
+// Run allows to execute arbitrary git commands
```
stevekuznetsov, line 248: Why should we expose this instead of explicit client methods? I worry that it's too easy to break some "infra" with a black-box `Run()` and impossible to review usage.

prow/git/v2/client_factory.go
```diff
@@ -41,37 +43,107 @@ type ClientFactory interface {
...
+type ClientFactoryOpt func(*ClientFactoryOpts)
+
+func defaultClientFactoryOpts(cfo *ClientFactoryOpts) {
+	if cfo.Host == "" {
+		cfo.Host = "github.com"
+	}
+	if cfo.CacheDirBase == nil {
+		switch runtime.GOOS {
+		case "linux":
+			cfo.CacheDirBase = utilpointer.StringPtr("/var/tmp")
```
stevekuznetsov, line 106: Doesn't Go have a tempdir function?

prow/git/v2/remote.go
```diff
@@ -72,24 +73,28 @@ func (f *sshRemoteResolverFactory) PublishRemote(_, repo string) RemoteResolver
 	}
 }
 
-type simpleAuthResolverFactory struct {
-	host     string
+type httpResolverFactory struct {
```
stevekuznetsov, line 76: Since the point of this is to provide simple auth, why rename?

prow/git/v2/remote_test.go
```diff
@@ -153,8 +153,29 @@ func TestSSHRemoteResolverFactory(t *testing.T) {
 	}
 }
 
-func TestSimpleAuthRemoteResolverFactory(t *testing.T) {
-	factory := simpleAuthResolverFactory{
+func TestHTTPResolverFactory_NoAuth(t *testing.T) {
```
stevekuznetsov, line 156: Ah, I see,

prow/git/v2/client_factory.go
alvaroaleman: Yeah we still use that, the problem is it returns `"/tmp"` on linux which is tmpfs backed. If this git client is used in a component that clones many, potentially big repos, this may result in gigabytes of data and either the laptop crashing or a `no space on device` error.

`"/var/tmp"` is intended for temporary files that may be bigger and might be retained longer: https://unix.stackexchange.com/questions/30489/what-is-the-difference-between-tmp-and-var-tmp

It even exists on mac

prow/git/v2/remote.go
alvaroaleman: No its not. That is something this can do, but it will also work fine anonymously.

PR #17407, opened 2020-04-27
Fix installing gotestsum in k/k unit test coverage job.

amwat:
/cc @BenTheElder @spiffxp

### DIFF
--- a/config/jobs/kubernetes/sig-testing/coverage.yaml
+++ b/config/jobs/kubernetes/sig-testing/coverage.yaml
@@ -144,8 +144,8 @@ periodics:
       args:
       - -c
       - |
-        local result=0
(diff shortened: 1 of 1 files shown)


BenTheElder:
/lgtm
/approve
whoops @dims @liggitt :/

config/jobs/kubernetes/sig-testing/coverage.yaml
```diff
@@ -144,8 +144,8 @@ periodics:
       args:
       - -c
       - |
-        local result=0
-        GO111MODULE=off go install k8s.io/kubernetes/vendor/gotest.tools/gotestsum || result=$?
+        result=0
+        ( cd hack/tools && GO111MODULE=on go install gotest.tools/gotestsum ) || result=$?
```
liggitt, line 148: does this need to work for multiple versions? a temp dir and a go get would be less entangled with the guts of k/k

PR #17413, opened 2020-04-28
Add testgrid dashboard for scalability-golang jobs

mm4tt:


mm4tt:
Looks like this needs to happen together with adding a new tab, as it's not allowed to add empty dashboard

PR #17414, opened 2020-04-28
Fix k8s commit hash in ci-build-and-push-k8s-at-golang-tip job

tosi3k:
Kudos to @mm4tt for pointing out that the prior commit hash was a bit too stale.

/sig scalability


mm4tt:
Thanks!

/lgtm
/approve

PR #17415, opened 2020-04-28
Run pull-ipvs job if ipvs proxy is changed

aojea:
Automatically run the ipvs job if some file of the proxy ipvs is changed in the PR.


config/jobs/kubernetes/sig-network/sig-network-misc.yaml
```diff
@@ -2,6 +2,7 @@ presubmits:
   kubernetes/kubernetes:
   - name: pull-kubernetes-e2e-gci-gce-ipvs
     always_run: false
+    run_if_changed: '^pkg/proxy/ipvs/'
```
andrewsykim, line 5: also `pkg/util/ipvs`?

aojea: I think this regex will take both

PR #17416, opened 2020-04-28
Adds gce-service-account flag to kops periodics

geojaz:
This flag is hopefully enough to get us through runs like https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/e2e-kops-gce-latest/1255153581443518467 . It should use the kops specific GSA as the service account for the nodes/masters because that account has RW access to the state store bucket. :crossed_fingers:

### DIFF
--- a/config/jobs/kubernetes/kops/kops-periodics-gce.yaml
+++ b/config/jobs/kubernetes/kops/kops-periodics-gce.yaml
@@ -21,6 +21,7 @@ periodics:
       - --env=KOPS_RUN_TOO_NEW_VERSION=1
       - --extract=release/latest
       - --ginkgo-parallel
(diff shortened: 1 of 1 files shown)


rifelpet:
i know in AWS there is an explicit `iam:PassRole` permission needed to assign roles to other resources. I don't know if GCP has an equivalent or if this service account has restricted permissions around that. I guess we'll fine out

/lgtm

geojaz:
In this case, I think we should be using the `pr-kubekins@kubernetes-jenkins-pull.iam.gserviceaccount.com` account to execute kops and assign itself to the nodes and masters. So it should _theoretically_ work ;) Let's see what sort of errors we can get outta this now! thanks Peter!

PR #17423, opened 2020-04-28
don't cleanup during dind up

BenTheElder:
/cc @amwat @munnerz 

will need to follow up with kubekins taking new bootstrap => new kubekins

pretty sure KRTE never had this bug

### DIFF
--- a/images/bootstrap/runner.sh
+++ b/images/bootstrap/runner.sh
@@ -78,7 +78,6 @@ if [[ "${DOCKER_IN_DOCKER_ENABLED}" == "true" ]]; then
             break
         fi
     done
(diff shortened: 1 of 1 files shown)


munnerz:
Why was this done in the first place? To ensure a clean environment when it starts (ie for binfmt_misc etc)?

BenTheElder:
> Why was this done in the first place? To ensure a clean environment when it starts (ie for binfmt_misc etc)?

it used to do binfmt_misc and docker data root cleanup (when we were persisting that)

BenTheElder:
somewhere along the line this got mangled. should be fairly sane now.

PR #17432, opened 2020-04-29
remove kind unsupported versions from conformance

aojea:
versions prior to 1.15 are unsupported, we don't need to monitor them in the conformance dashboard.

https://k8s-testgrid.appspot.com/conformance-kind

We'll keep monitoring them in the presubmits of the kind project


spiffxp:
needs rebase, I'll defer to @BenTheElder on removing these

BenTheElder:
/lgtm
/approve
thanks

PR #17433, opened 2020-04-29
fix kind presubmits for old k/k versions

aojea:
With the last bump of the kubekins image , kind jobs prior 1.15 k/k version stopped to work.

Restore previous kubekins images for those jobs.

### DIFF
--- a/config/jobs/kubernetes-sigs/kind/kind-presubmits.yaml
+++ b/config/jobs/kubernetes-sigs/kind/kind-presubmits.yaml
@@ -329,7 +329,7 @@ presubmits:
     always_run: true
     spec:
       containers:
-      - image: gcr.io/k8s-testimages/kubekins-e2e:v20200428-06f6e3b-1.14
+      - image: gcr.io/k8s-testimages/kubekins-e2e:v20200218-a445c54-1.14
         env:
         # skip serial tests and run with --ginkgo-parallel
         - name: "PARALLEL"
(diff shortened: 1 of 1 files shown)


aojea:
/assign @BenTheElder 
xref: https://github.com/kubernetes/test-infra/pull/17432

aojea:
how is the workflow, once it merges it automatically creates the new image?
one we have the new image, we have to update the jobs, right?

BenTheElder:
this was a manually done bump, the normal bump process is smarter

I think maybe it might be time to abandon these jobs, or create ones that don't build k8s (feasible in 0.9 timeframe...), the image is no longer getting built for these so e.g. the ipv6 change you filed wont be included..

aojea:
do we want to  support or at least not break this versions? I think that there are some users still asking about it, if is not a big effort I'm supportive

If we decide to support old versions, I prefer to have the signal in the presubmits than in the periodic jobs, as you say, k8s is not going to change, so the more possible cause to break those versions will be a change in kind (I guess)

BenTheElder:
so the problem is the kubernetes project will not support these.

we can't keep using these ancient images though, they lack fixes to the runner etc.

so what we can do in the future is build from released k8s binaries and just use the master branch krte image in a CI job.

for now we're not touching k8s version specific things and need to finish 0.8, we need to remove these.

BenTheElder:
we _also_ need to drop support for some versions, but I'd prefer to be more lenient than k8s

aojea:
/close
I can follow up tomorrow deleting this jobs if you don't have time

PR #17435, opened 2020-04-29
kops grid: run tests against older k8s versions

justinsb:
This should allow us to see for example whether CNI problems are
regressions or existed previously.


justinsb:
cc @hakman and @rifelpet ... this should let us see whether the CNI problems are regressions.

I went down to weekly by default so we don't overwhelm the system.  That actually gives us enough room to test older versions of kops as well - might try that tomorrow unless one of you beats me to it (I don't currently see where we would get a `- --kops-version=https://storage.googleapis.com/kops-ci/bin/latest-ci-updown-green.txt` equivalent for older kops versions).

(Also #17370, #17371 and #17372 are nice-to-haves though not critical)

rifelpet:
Nice! looks like some minor python lint issues

I wary for my laptop screen real estate :) maybe we create additional testgrid dashboards, perhaps per k8s version? `sig-cluster-lifecycle-kops-1.17` etc.

Theres quite a few ways we could split these jobs up, and we're still only testing kops' master branch. I'm not sure if we need to ask for a new dashboard or if we can just make new ones.

rifelpet:
Ah i forgot that a job can live under multiple dashboards. That definitely makes sense to have dashboards sliced up different ways.

PR #17443, opened 2020-04-30
move kind presubmits to match kubernetes upstream supported versions

BenTheElder:
drop: 1.11 ... 1.14
add: 1.18


BenTheElder:
diff is best read commit by commit, otherwise it looks crazy.

PR #17444, opened 2020-04-30
apply 17407 to cached unit test job

BenTheElder:
xref: https://github.com/kubernetes/test-infra/pull/17407/files
/cc @dims @amwat

### DIFF
--- a/config/jobs/kubernetes/sig-testing/make-test.yaml
+++ b/config/jobs/kubernetes/sig-testing/make-test.yaml
@@ -22,7 +22,7 @@ periodics:
             - |
               result=0
               # Run the tests as usual
-              GO111MODULE=off go install k8s.io/kubernetes/vendor/gotest.tools/gotestsum || result=$?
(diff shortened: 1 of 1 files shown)


BenTheElder:
/hold
sigh

BenTheElder:
this issue also applies to the job that creates the cache just up a bit in this file

config/jobs/kubernetes/sig-testing/make-test.yaml
```diff
@@ -22,7 +22,7 @@ periodics:
             - |
               result=0
               # Run the tests as usual
-              GO111MODULE=off go install k8s.io/kubernetes/vendor/gotest.tools/gotestsum || result=$?
+              ( cd hack/tools && GO111MODULE=on go install gotest.tools/gotestsum ) || result=$?
```
BenTheElder, line 25: updated to include this one as well
this should probably get DRYed out but it's a prototype soo :man_shrugging:

PR #17448, opened 2020-04-30
Add leader election to tide

### LINKED ISSUE #16153: Tide: Add leader election

Leader electiion to tide

**Why is this needed**:

Tide is not built to be running with more than one replica. In order to avoid problems during upgrades, node rotations or if someone accidentally increases the replica count, we should add leader election to get a stronger assurance that only one replica will be active at a time.

/area prow

stpabhi:
fixes #16153

### DIFF
--- a/config/prow/cluster/starter.yaml
+++ b/config/prow/cluster/starter.yaml
@@ -830,6 +830,22 @@ rules:
       - list
       - get
       - watch
+  - apiGroups:
+    - ""
+    resources:
+    - configmaps
+    resourceNames:
+    - prow-tide-leaderlock
+    verbs:
+    - get
+    - update
+  - apiGroups:
+    - ""
+    resources:
+    - configmaps
+    - events
+    verbs:
+    - create
 ---
 kind: RoleBinding
 apiVersion: rbac.authorization.k8s.io/v1beta1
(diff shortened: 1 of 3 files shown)


alvaroaleman:
@stpabhi Did you mean to close the PR?

stpabhi:
@alvaroaleman yes. My bad that I didn’t read the comment that says mgr is not used to control the lifecycle. I believe whatever changes I requested might be no-op. Also, is there any reason for doing that? I haven’t looked at the code in detail yet. Any pointers?

alvaroaleman:
So the reason why we don't use the `Manager` for the binaries lifecycle is that it doesn't support graceful termination (xref https://github.com/kubernetes-sigs/controller-runtime/issues/764). This is why we currently have the `interrupts` package even though it solves a problem that is not specific to this project in any way.

Now that I think about it though, we could probably:
* Continue to use `interrupt` to manage the binaries lifecycle
* Build a wrapper for the tide controllers that implements [`manager.Runnable`](https://godoc.org/sigs.k8s.io/controller-runtime/pkg/manager#Runnable) (This is needed so the Manager can control their lifecycle)
* Have that wrapper finish an active sync but not start a new one once its stop channel is closed
* Somehow tell `interrupts` that it has to wait (maybe we can add the controllers to both the manager and `interrupts`?) 

Strictly speaking that means that there is a short period in which two tide instances could do work since we basically ignore the lost leader lease during shutdown and finish our current work anyways, but that is certainly better than no leader election at all.

WDYT?

stpabhi:
Thanks for the input. I will take a look at the references and interrupts pkg to understand  it.

Issue #14643, opened 2019-10-07
flake data excludes pod-utils jobs


BenTheElder:
See http://storage.googleapis.com/k8s-metrics/flakes-latest.json etc. (metrics/ produced files)
and http://velodrome.k8s.io/dashboard/db/bigquery-metrics?orgId=1

The flake data is very misleading, for example `pull-kubernetes-verify` has "no flakes" which is definitely wrong.

What seems to be happening if we only include data from bootstrap.py results, not pod-utils (I think) possibly due to handling of the repos data (per @cjwagner)

We should fix this, not having flake data is a pretty big regression for managing kubernetes presubmits. I didn't realize that jobs I'd migrated were losing this.

stevekuznetsov:
What data/files is the flake analysis using? Are the pod-utils not uploading something they should?
/assign

BenTheElder:
I'm not sure I fully understand the pipeline just yet, I took a quick look at it seems something like:

- https://github.com/kubernetes/test-infra/tree/master/kettle -> fetch finished / started.json from each job, move into some schema in big query
- https://github.com/kubernetes/test-infra/tree/master/metrics -> query the table, select flake data (which appears to be reading repos in started?)

I think the issue is that repos is now a JSON blob, so @fejta suggested something like making this a field in the database and then reading that instead of this https://github.com/kubernetes/test-infra/blob/5deb5b970e73cdd55b3068b9c50962e8657bdb23/metrics/configs/flakes-config.yaml#L64

BenTheElder:
I'm not actually sure which component is at fault here or exactly why this doesn't work, but looking at the data we produce jobs that are fully pod-utils are missing and jobs that migrated to pod-utils on newer branches  have "0 flakes" when they definitely have non-zero flakes.

I would tend to suggest that the pipeline is a bit hairy and probably at fault, but the results are generally very useful for identifying sources of flakiness.

The data looks (?) present in pod-utils to me but I'm not fully familiar with that format or the big query pipeline...

https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/83519/pull-kubernetes-e2e-gce/1181359333854679040/started.json
https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/83519/pull-kubernetes-e2e-kind/1181359333888233473/started.json

https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/83519/pull-kubernetes-e2e-gce/1181359333854679040/finished.json
[...]

PR #17451, opened 2020-04-30
Migrate kind jobs to podutils

aojea:


config/jobs/kubernetes-sigs/kind/kind.yaml
```diff
@@ -212,46 +236,3 @@ periodics:
     description: Runs conformance tests using kubetest against latest kubernetes master with an IPv6 kubernetes-in-docker cluster, skipping [Serial] tests
     testgrid-alert-email: bentheelder@google.com,antonio.ojea.garcia@gmail.com
     testgrid-num-columns-recent: '3'
-# conformance test against kubernetes release-1.15 branch with `kind`
```
aojea, line 215: I think this is a leftover, is the only release is being tested here, the others are testing master only

BenTheElder:
/cc
firefighting back to this later today

config/jobs/kubernetes-sigs/kind/kind.yaml
BenTheElder: I don't think so, IIRC in 1.16 spiffxp switched some around.

aojea: anyway, it went out of support, do we want to keep testing it?

BenTheElder: we should leave it in kind presubmit. removing here is OK.

BenTheElder:
worth noting https://github.com/kubernetes/test-infra/issues/14643

BenTheElder:
needs a rebase again sadly.

aojea:
> worth noting #14643

/hold until that is solved then, flake data is important

BenTheElder:
spoke to aaron, I think we should move forward anyhow.

we can get testcase data via the triage dashboard, we just won't get data on job consistency which we're already missing where it's important (`pull-kubernetes-e2e-kind`) ... hopefully that gets resolved this quarter
/hold cancel

BenTheElder:
/lgtm
/approve
rebased it

config/jobs/kubernetes-sigs/kind/kind-presubmits.yaml
```diff
@@ -119,20 +128,13 @@ presubmits:
...
-        - "--"
-        # the script must run from kubernetes, but we're checking out kind
-        - "bash"
-        - "--"
-        - "-c"
-        - "cd ./../../k8s.io/kubernetes && ./../../sigs.k8s.io/kind/hack/ci/e2e.sh"
+        - name: BUILD_TYPE
+          value: bazel
+        command:
+        - wrapper.sh
```
BenTheElder, line 134: ahhh, this had to be moved to KRTE to use wrpaper.sh :|

BenTheElder: https://github.com/kubernetes/test-infra/pull/17593

PR #17452, opened 2020-04-30
[jobs] use node-e2e-project boskos pool for ci-containerd-node-e2e

hasheddan:
Trying same change made in https://github.com/kubernetes/test-infra/pull/17349

Ref: https://github.com/kubernetes/kubernetes/issues/89847

### DIFF
--- a/config/jobs/kubernetes/sig-node/containerd.yaml
+++ b/config/jobs/kubernetes/sig-node/containerd.yaml
@@ -233,7 +233,7 @@ periodics:
       - --
       - --node-args=--image-config-file=/workspace/test-infra/jobs/e2e_node/containerd/containerd-master/image-config.yaml
       - --deployment=node
-      - --gcp-project=cri-containerd-node-e2e
+      - --gcp-project-type=node-e2e-project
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes/sig-node/sig-node-presubmit.yaml
```diff
@@ -101,7 +101,7 @@ presubmits:
         - --scenario=kubernetes_e2e
         - --
         - --deployment=node
-        - --gcp-project=k8s-c8d-pr-node-e2e
```
spiffxp, line 104: We don't have enough capacity in this pool to handle presubmits, can you try with a ci- job that uses this project first?

hasheddan: @spiffxp hmm I don't see any other jobs that use this specific project

hasheddan: We could try this one in `cri-containerd-node-e2e` https://testgrid.k8s.io/sig-node-containerd#containerd-node-conformance

PR #17453, opened 2020-04-30
Bump up limits for verify and cross presubmits

dims:
Lots of failures for the last day or so and we haven't merged much as
well. Looking at:
https://prow.k8s.io/?repo=kubernetes%2Fkubernetes&type=batch&state=failure

pull-kubernetes-e2e-kind and pull-kubernetes-e2e-gce are hit by flakes
but pull-kubernetes-cross and pull-kubernetes-verify seem to be be
running slower and hitting timeouts.

for verify job:
- timeout up to 3 hours (override default)
- bump cpu to 6 (just like ci-kubernetes-verify-master)

for cross job:
- explicitly set 2 hours and drop the command line timeout of 90 minutes

### DIFF
--- a/config/jobs/kubernetes/sig-release/kubernetes-builds.yaml
+++ b/config/jobs/kubernetes/sig-release/kubernetes-builds.yaml
@@ -5,6 +5,8 @@ presubmits:
     labels:
       preset-service-account: "true"
       preset-dind-enabled: "true"
+    decoration_config:
+      timeout: 2h
     spec:
       containers:
       - image: gcr.io/k8s-testimages/bootstrap:v20200428-cdaab98
@@ -13,7 +15,6 @@ presubmits:
         - "--repo=k8s.io/$(REPO_NAME)=$(PULL_REFS)"
         - "--service-account=/etc/service-account/service-account.json"
         - "--upload=gs://kubernetes-jenkins/pr-logs"
-        - "--timeout=90"
         - --scenario=execute
         - --
         - --env=KUBE_RELEASE_RUN_TESTS=n
--- a/config/jobs/kubernetes/sig-testing/verify.yaml
+++ b/config/jobs/kubernetes/sig-testing/verify.yaml
@@ -14,6 +14,8 @@ presubmits:
     labels:
       preset-service-account: "true"
       preset-dind-enabled: "true"
+    decoration_config:
+      timeout: 3h
     spec:
       containers:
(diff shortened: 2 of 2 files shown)


config/jobs/kubernetes/sig-testing/verify.yaml
```diff
@@ -14,6 +14,8 @@ presubmits:
     labels:
       preset-service-account: "true"
       preset-dind-enabled: "true"
+    decoration_config:
+      timeout: 3h
```
wojtek-t, line 18: 3 hours sounds a lot to me. Isn't 2 hours really enough?

config/jobs/kubernetes/sig-testing/verify.yaml
```diff
@@ -36,7 +38,7 @@ presubmits:
           privileged: true
         resources:
           requests:
-            cpu: 4
+            cpu: 6
```
wojtek-t, line 41: out of curiosity: do you know what parts of verify use more than a single core? I though that most of these are actually single-threaded...

config/jobs/kubernetes/sig-testing/verify.yaml
spiffxp: I agree with jordan in https://github.com/kubernetes/kubernetes/pull/90639#issuecomment-621810350 we should consider splitting the verify job up into e.g. verify and verify-slow

config/jobs/kubernetes/sig-release/kubernetes-builds.yaml
```diff
@@ -5,6 +5,8 @@ presubmits:
     labels:
       preset-service-account: "true"
       preset-dind-enabled: "true"
+    decoration_config:
+      timeout: 2h
```
spiffxp, line 9: I don't think this will work, this isn't a decorated job

config/jobs/kubernetes/sig-release/kubernetes-builds.yaml
```diff
@@ -13,7 +15,6 @@ presubmits:
         - "--repo=k8s.io/$(REPO_NAME)=$(PULL_REFS)"
         - "--service-account=/etc/service-account/service-account.json"
         - "--upload=gs://kubernetes-jenkins/pr-logs"
-        - "--timeout=90"
```
spiffxp, line 16: I'm fine bumping this up to 120

dims:
will file other PRs

spiffxp:
ref: https://github.com/kubernetes/kubernetes/issues/90647

PR #17456, opened 2020-04-30
Bump timeout for pull-kubernetes-cross

dims:
Drop the extra timeout so we use the default of 2 hours


spiffxp:
ref: https://github.com/kubernetes/kubernetes/issues/90647

PR #17459, opened 2020-04-30
Azure: enforce strict JSON when parsing API model

chewong:
- Rename `azure_helpers.go` -> `aksengine_helpers.go` and `azure_test.go` -> `aksengine_test.go`
- Enforce strict JSON when parsing API model by enabling [`DisallowUnknownFields`](https://golang.org/pkg/encoding/json/#Decoder.DisallowUnknownFields) from `json.Decoder`.
- Add test cases to test the functionality above

/cc @marosset 
/assign @feiskyer


feiskyer:
> Enforce strict JSON when parsing API model by enabling DisallowUnknownFields from json.Decoder.

Do we have any reasons to enable this? I think it may introduce troubles if we want to add something new in the future.

chewong:
> > Enforce strict JSON when parsing API model by enabling DisallowUnknownFields from json.Decoder.
> 
> Do we have any reasons to enable this? I think it may introduce troubles if we want to add something new in the future.

Without enforcing strict JSON, kubetest simply omits unknown fields in badly-formatted JSON such as https://github.com/kubernetes-sigs/windows-testing/pull/168/files, which causes aks-engine not to return an error, leading to inaccurate test results (the containerd jobs were actually testing with docker due to the error). We should encourage others to modify aksengine_helpers.json if they want to add new fields in API model.

marosset:
I agree with @chewong here.
I'd rather see failures when there are fields in an aks-engine config that aren't know the kubetest or if we encounter misconfigured json over having misleading test results.

adelina-t:
I second this. As there is no trivial way to keep this in  sync with aks-engine api-model, it's better to have a signal to know when we've diverged.

PR #15596, opened 2019-12-13
Add support for empty label to ignore voting in gerrit.

amwat:
It is fairly common for users to use a non-existent label as a way to achieve "informing but not blocking" jobs. Dropping log level down to info to avoid spamming error logs.

/cc @cjwagner

PR #17460, opened 2020-04-30
Make Gerrit reporter respect report field on prow jobs

### LINKED ISSUE #17419: Gerrit controller does not respect `report: false` on ProwJobs.
**What happened**:
Prow's allows for a `skip_report` field to be specified in job config and that is inverted into a `report` field on created ProwJobs. This field is respected by GitHub reporting and prevents including the job result in comments or status contexts. The gerrit controller doesn't appear to respect this field.

- It should be a config error to configure gerrit report labels with `skip_report: true`
- The Gerrit controller should not report `report: false` jobs to Gerrit via comments nor labels.

Configure a gerrit job with `skip_report: true` and a gerrit report label annotation.

I think this can currently be worked around by omitting all gerrit report label annotation.

/help

adshmh:
Closes #17419

### DIFF
--- a/prow/config/config.go
+++ b/prow/config/config.go
@@ -47,6 +47,7 @@ import (
 	"sigs.k8s.io/yaml"
 
 	prowapi "k8s.io/test-infra/prow/apis/prowjobs/v1"
+	gerrit "k8s.io/test-infra/prow/gerrit/client"
 	"k8s.io/test-infra/prow/git/v2"
 	"k8s.io/test-infra/prow/github"
 	"k8s.io/test-infra/prow/kube"
@@ -1274,7 +1275,6 @@ func validateJobBase(v JobBase, jobType prowapi.ProwJobType, podNamespace string
 // validatePresubmits validates the presubmits for one repo
 func validatePresubmits(presubmits []Presubmit, podNamespace string) error {
 	validPresubmits := map[string][]Presubmit{}
-
 	var errs []error
 	for _, ps := range presubmits {
 		// Checking that no duplicate job in prow config exists on the same branch.
@@ -1297,7 +1297,7 @@ func validatePresubmits(presubmits []Presubmit, podNamespace string) error {
 		if err := validateTriggering(ps); err != nil {
 			errs = append(errs, err)
 		}
-		if err := validateReporting(ps.Reporter); err != nil {
+		if err := validateReporting(ps.JobBase, ps.Reporter); err != nil {
 			errs = append(errs, fmt.Errorf("invalid presubmit job %s: %v", ps.Name, err))
 		}
 		validPresubmits[ps.Name] = append(validPresubmits[ps.Name], ps)
@@ -1353,7 +1353,7 @@ func validatePostsubmits(postsubmits []Postsubmit, podNamespace string) error {
 		if err := validateJobBase(ps.JobBase, prowapi.PostsubmitJob, podNamespace); err != nil {
 			errs = append(errs, fmt.Errorf("invalid postsubmit job %s: %v", ps.Name, err))
 		}
-		if err := validateReporting(ps.Reporter); err != nil {
+		if err := validateReporting(ps.JobBase, ps.Reporter); err != nil {
 			errs = append(errs, fmt.Errorf("invalid postsubmit job %s: %v", ps.Name, err))
 		}
 		validPostsubmits[ps.Name] = append(validPostsubmits[ps.Name], ps)
@@ -1915,9 +1915,17 @@ func validateTriggering(job Presubmit) error {
 	return nil
 }
 
-func validateReporting(r Reporter) error {
+func validateReporting(j JobBase, r Reporter) error {
 	if !r.SkipReport && r.Context == "" {
 		return errors.New("job is set to report but has no context configured")
+	}
+	if !r.SkipReport {
+		return nil
+	}
+	for label, value := range j.Labels {
+		if label == gerrit.GerritReportLabel && value != "" {
+			return fmt.Errorf("Gerrit report label %s set to non-empty string but job is configured to skip reporting.", label)
+		}
 	}
 	return nil
 }
--- a/prow/gerrit/adapter/adapter.go
+++ b/prow/gerrit/adapter/adapter.go
@@ -182,7 +182,11 @@ func (c *Controller) ProcessChange(instance string, change client.ChangeInfo) er
 		return fmt.Errorf("failed to get SHA from base branch: %v", err)
 	}
 
-	triggeredJobs := []string{}
+	type triggeredJob struct {
+		name   string
+		report bool
+	}
+	var triggeredJobs []triggeredJob
 
 	refs, err := createRefs(instance, change, cloneURI, baseSHA)
 	if err != nil {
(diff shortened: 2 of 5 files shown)


fejta:
/uncc @fejta @michelle192837 
/assign @cjwagner 

Thanks for fixing this!

prow/config/config.go
```diff
@@ -1915,10 +1915,18 @@ func validateTriggering(job Presubmit) error {
 	return nil
 }
 
-func validateReporting(r Reporter) error {
+func validateReporting(j JobBase, r Reporter) error {
 	if !r.SkipReport && r.Context == "" {
 		return errors.New("job is set to report but has no context configured")
 	}
+	if !r.SkipReport {
```
fejta, line 1922: What happens when !SkipReport && len(j.Labels) == 0?

prow/gerrit/adapter/adapter.go
```diff
@@ -182,7 +182,11 @@ func (c *Controller) ProcessChange(instance string, change client.ChangeInfo) er
 		return fmt.Errorf("failed to get SHA from base branch: %v", err)
 	}
 
-	triggeredJobs := []string{}
+	type triggeredJob struct {
+		name   string
+		report bool
+	}
+	triggeredJobs := []triggeredJob{}
```
fejta, line 189: You are not initializing anything so use `var triggeredJobs []triggeredJob`

prow/gerrit/adapter/adapter_test.go
```diff
@@ -704,18 +763,18 @@ func TestProcessChange(t *testing.T) {
...
 
-		testInstance := "https://gerrit"
 		fakeProwJobClient := prowfake.NewSimpleClientset()
 		fakeLastSync := client.LastSyncState{testInstance: map[string]time.Time{}}
 
 		for _, tc := range testcases {
 			fakeLastSync[testInstance][tc.change.Project] = timeNow.Add(-time.Minute)
 		}
 
+		gc := &fgc{}
```
fejta, line 773: Similar to above I would consider `var gc fgc` (and `gc: &gc` below) to be better practice as we never set gc to nil so there's no need to have to deal with pointers.

prow/gerrit/adapter/adapter_test.go
```diff
@@ -757,5 +816,10 @@ func TestProcessChange(t *testing.T) {
 				t.Errorf("tc %s - BaseSHA should be abc, got %s", tc.name, prowjobs[0].Spec.Refs.BaseSHA)
 			}
 		}
+		if tc.shouldSkipReport {
+			if gc.reviews > 0 {
+				t.Errorf("tc %s - expected no comments, got: %d", tc.name, gc.reviews)
```
fejta, line 821: Ideally we refactor this unit test to match modern practices, aka

```golang
// deprecated
  // whatever
  tc.Errorf("tc %s ...", tc.name)
}

// expected
    t.Error("...") // name automatically appended
  }
}
```

This makes for better unit tests: cases run in parallel which make them faster as well as more likely to be hermetic.

Typically I convert test functions that still use the deprecated pattern when I'm modifying them (feel free to pass if you don't feel up to it though).

prow/gerrit/adapter/adapter.go
```diff
@@ -293,19 +297,28 @@ func (c *Controller) ProcessChange(instance string, change client.ChangeInfo) er
...
+			})
 		}
 	}
 
+	var report bool
 	if len(triggeredJobs) > 0 {
-		// comment back to gerrit
-		message := fmt.Sprintf("Triggered %d prow jobs:", len(triggeredJobs))
+		// comment back to gerrit if Report is set for any of the jobs
+		message := fmt.Sprintf("Triggered %d prow jobs (the following list only includes jobs which do not suppress reporting):", len(triggeredJobs))
```
cjwagner, line 310: nit: Something like `(%d suppressed reporting)` would be a little more concise and informative.

prow/gerrit/adapter/adapter_test.go
cjwagner: FYI I think you also need to call [`t.Parallel()`](https://golang.org/pkg/testing/#T.Parallel) at the beginning of the anonymous run function if you want the test cases to run in parallel. ref: https://golang.org/pkg/testing/#hdr-Subtests_and_Sub_benchmarks

prow/config/config.go
cjwagner: My understanding is that labels are used to control voting and grouping jobs together to aggregate job reporting. Setting the report label to the empty string causes a job to report immediately when it completes without aggregation (https://github.com/kubernetes/test-infra/pull/15596).  Omitting the report label altogether causes the default label of "Code-Review" to be used (should default to "Verified" IMO).
So for all possible states of the gerrit report label (populated, empty string, omitted) we still comment to Gerrit. I think skip_report then can control if we comment at all, but we should probably only allow skip_report when the gerrit report label is set to the empty label since we need to comment to vote on a label.

I think @amwat and @krzyzacy know the most about Gerrit reporting and labels. Is the above accurate? Do you have thoughts on how this config should work?

prow/gerrit/adapter/adapter_test.go
cjwagner: FYI I think you also need to call [`t.Parallel()`](https://golang.org/pkg/testing/#T.Parallel) at the beginning of the anonymous run function if you want the test cases to run in parallel. ref: https://golang.org/pkg/testing/#hdr-Subtests_and_Sub_benchmarks

prow/config/config.go
amwat: Yes, that summary is accurate. also +1 to default verified (that is the intent of that label: https://gerrit-review.googlesource.com/Documentation/config-labels.html#label_Verified) the only problem is that verified "-1" is blocking by default and code-review is not, maybe after we have this skip-report feature Verified could be a sane default.

In terms of config, I guess i'm not sure what we want to achieve by skip-report, specifically whether we just want to 
1. skip reporting on a label but still show up in comments, 
2. or just not report at all to gerrit in any way.

If 2. then we can just return false here when skip-report is set (and exclude it from the selector)
https://github.com/kubernetes/test-infra/blob/6e7f9252d5fae7d74fd9ffabe9175078240000ca/prow/crier/reporters/gerrit/reporter.go#L104-L105

If 1. then forcing gerrit-report label to be empty makes sense.

adshmh: Thank you for the reviews. I have updated the PR to allow `skip_report` only if the Gerrit report label is set to empty or is omitted. I think the ignoring of `skip_report` was caused by the following: 

https://github.com/kubernetes/test-infra/blob/b5246a13a917f003450af87d15a94ff2dbc711f1/prow/gerrit/adapter/adapter.go#L307-L309



Crier will not run any reporters for a job which has `skip_report` set:
https://github.com/kubernetes/test-infra/blob/b5246a13a917f003450af87d15a94ff2dbc711f1/prow/crier/controller.go#L254-L258

prow/gerrit/adapter/adapter.go
adshmh: Thank you for the review. Fixed.

prow/gerrit/adapter/adapter_test.go
adshmh: Thank you for the review. Fixed.

prow/gerrit/adapter/adapter_test.go
adshmh: Thank you for the reviews. Fixed.

PR #17340, opened 2020-04-22
Checkconfig: Add validation for tide context policy

alvaroaleman:
/assign @fejta @stevekuznetsov 

To avoid discovering this at runtime with errors like
```
{"branch":"release-4.3","component":"tide","controller":"status-update","error":"failed to set up context register: contexts ci/prow/shellcheck, ci/prow/tf-lint, ci/prow/yaml-lint are defined as required and required if present","file":"prow/tide/status.go:371","func":"k8s.io/test-infra/prow/tide.(*statusController).setStatuses.func1","level":"error","msg":"getting expected status","org":"openshift","pr":3460,"repo":"installer","sha":"e05a6363aa9e40cd3f341ada8ee3beb9dcabeda3","time":"2020-04-22T15:18:55Z"}
```

PR #17467, opened 2020-04-30
Revert "Checkconfig: Add validation for tide context policy"

cjwagner:
This reverts commit e647ad266572e105d7e1ec78fa6d1374c49af50e.
Explanation here: https://github.com/kubernetes/test-infra/pull/17340#discussion_r418211086

/assign @BenTheElder @alvaroaleman @fejta

### DIFF
--- a/prow/cmd/checkconfig/BUILD.bazel
+++ b/prow/cmd/checkconfig/BUILD.bazel
@@ -85,6 +85,5 @@ go_test(
         "@io_k8s_apimachinery//pkg/util/errors:go_default_library",
         "@io_k8s_apimachinery//pkg/util/sets:go_default_library",
         "@io_k8s_sigs_yaml//:go_default_library",
-        "@io_k8s_utils//pointer:go_default_library",
     ],
 )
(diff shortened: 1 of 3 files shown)


alvaroaleman:
/hold
Can you just disable the check?

alvaroaleman:
Tide does not support Gerrit so there is no reason to have any tide-related verifications unless its a shared prow instance for both Gerrit and Github?

We can probably disable the check by default but IMHO it makes a lot of sense to have it when Tide is being used and I would assume the majority of Prow users does use it with GitHub and Tide.

cjwagner:
Discussed out of band with Alvaro. This is problematic for Prow instance that serve both Gerrit and GitHub repos like this instance does: https://github.com/GoogleCloudPlatform/oss-test-infra

Alvaro is working on a quick forward fix to ignore non- `org/repo` format repos in the `presubmits:` section.

cjwagner:
Superseded by https://github.com/kubernetes/test-infra/pull/17469
/close

PR #17468, opened 2020-04-30
WIP: Set up service account for kops GCE instances

geojaz:
**Problem:** 
All kops GCE jobs have been broken since we switched to Workload identity on prow a couple months ago. Kops expects certain things to be able to successfully execute a test run- one is a service account that the nodes will be launched as that has RW on gs://k8s-kops-gce/. 

The current implementation in prow allows us to assume the GSA `pr-kubekins@kubernetes-jenkins-pull.iam.gserviceaccount.com` to _execute_ our job as `k8s-kops-test` KSA. Without specifying an instance service account, kops tries to use the default GSA which doesn't have the correct permissions on that bucket. 

kops has a `--gce-service-account` flag where we can provide the name of a service account for the instances to use, but this requires Service Account User or `iam.serviceAccounts.actAs` on the service account itself. So we get errors like: 
```
E0430 18:05:42.577918     242 op.go:136] GCE operation failed: googleapi: Error 400: The user does not have access to service account 'pr-kubekins@kubernetes-jenkins-pull.iam.gserviceaccount.com'.  User: 'pr-kubekins@kubernetes-jenkins-pull.iam.gserviceaccount.com'.  Ask a project owner to grant you the iam.serviceAccountUser role on the service account
```

**Proposal(s):**
We need help with the next step. I see how I can set up the relationship between a KSA and a GSA in prow, but I don't know how to set up permissions on that GSA. This PR is to help facilitate the discussion and implementation of next steps. 
~~I don't think we actually need this secondary service account `kops-instance`... it may be easier to just set up `iam.serviceAccounts.actAs` on `pr-kubekins`.~~
@fejta recommended a new service account for the instance SA and suggest to follow up with wg-k8s-infra.

### DIFF
--- a/config/prow/cluster/build_serviceaccounts.yaml
+++ b/config/prow/cluster/build_serviceaccounts.yaml
@@ -27,4 +27,13 @@ metadata:
   name: k8s-kops-test
   namespace: test-pods
 ---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  annotations:
+    # Kops GCE compute instance service account (nodes and masters)
+    iam.gke.io/gcp-service-account: kops-instance@kubernetes-jenkins-pull.iam.gserviceaccount.com
+  name: k8s-kops-instance
+  namespace: test-pods
+---
 # TODO(fejta): any other service accounts


BenTheElder:
/unassign
way over capacity, low context on this, and the robot already CCed multiple other people

geojaz:
no prob b3n, I figured you might know something, and the bot told me to add you so :)

config/prow/cluster/build_serviceaccounts.yaml
```diff
@@ -27,4 +27,13 @@ metadata:
   name: k8s-kops-test
   namespace: test-pods
 ---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  annotations:
+    # Kops GCE compute instance service account (nodes and masters)
+    iam.gke.io/gcp-service-account: kops-instance@kubernetes-jenkins-pull.iam.gserviceaccount.com
```
fejta, line 35: I would prefer we not the kubernetes-jenkins-pull project. If we need to use this project them IMO we might as well continue using pr-kubekins.

And ideally this uses a project owned by the community rather than Google (kubernetes-jenkins-pull is a Google owned project). @listx should know what the best practice is for setting this up (he did the k8s-infra-gcr-promoter GCPSA above)

On the other hand, I do have access to pr-kubekins and can run the gcloud command to grant it access to the k8s-kops-instance KSA.

geojaz: I think we're on the same page. 

I would prefer to create a new GSA (in whatever project is appropriate), and a new bucket, if needed. If we can get this GSA set up and give `pr-kubekins` `iam.serviceAccounts.actAs` on it, this should get us back on track.

fejta: Why would we need to give pr-kubekins anything?

geojaz: I was thinking that we keep the `pr-kubekins` GSA because it's used by all of our jobs and has the permissions required to execute kops and spin up a cluster to a variety of targets.

For GCE, `pr-kubekins` would execute kops. We pass a flag to kops `--gce-service-account=kops-instance@....iam.gserviceaccount.com` to be used for the VMs. So `pr-kubekins` would need serviceAccount user or permission to act as the `kops-instance` service account.

listx:
@geojaz Can you give a detailed picture of what GSA was used by Prow before the move to workload identity (WI)? That is, how did it use to work?

justinsb:
@listx and I had a look at this; we think the problem is actually that #17416 changed the service account, but actually the problem there looks to have been that kops wasn't granting permission to the underlying GCS bucket.  Going to look at what happened there...

justinsb:
I suspect (but have not yet been able to confirm) that gs://k8s-kops-gce has [UBLA](https://cloud.google.com/storage/docs/uniform-bucket-level-access) enabled , which means we don't set up permissions: https://github.com/kubernetes/kops/blob/3c85693938750c6fa5362782959b22ec17d92fd6/pkg/acls/gce/storage.go#L56-L64

That was a hotfix, but we need to raise the log level and/or verify the permissions.

That assumes that the problem is UBLA.  I'm not really sure what the best approach here is, TBH.  Perhaps moving to CNCF infra after all :-)

geojaz:
@justinsb @listx  Backstory: our kops GCE tests have been broken for a long long time. They didn't break as of #17416, that PR is a result of a long series of patches that I've been trying to iterate on to get these tests back online. I actually did https://github.com/kubernetes/kops/pull/8761/files primarily so that we could specify a GSA with appropriate permissions in those E2E tests. 

Not to be short, but I don't really care about the bucket k8s-kops-gce, I don't care about a specific account or patch. The bottom line is that I need a Google service account with a handful of permissions that we can use as the node service account for GCE tests. We also need a bucket that this service account can RW to. We're in this thread now because I can't figure out how to get those things. :)

That process is entirely opaque to me. Everyone is trying to figure out how to "get back to where we were" but there hasn't been a point in recent memory (last couple months) where our E2E tests have done anything useful. Even if we _could_make them work by changing some kops code, the E2E tests are IMO still wrong. In GCE world, we need to be able to assign a separate GSA to our nodes. 
Appreciate the help- but just please loop me into the discussion if you're trying to work through this.  I'm probably closer to the GCE stuff on kops than anyone else right now. :) thanks

geojaz:
I would also argue that https://github.com/kubernetes/kops/blob/3c85693938750c6fa5362782959b22ec17d92fd6/pkg/acls/gce/storage.go#L56-L64 isn't really what we want to do. I would prefer to manage that access somewhere other than kops. I don't have a strong justification for why, but it seems more intuitive to me from the gcp perspective /shrug

geojaz:
I think we have some clarification about how we can move forward. We raised this during our kops office hours this week:

We would like figure out how to get 1 _NEW_ google service account to be used for the GCE nodes that we spin up (kops-nodes@[hypothetical-project].iam.gserviceaccount.com). We will still need `pr-kubekins` to execute our prow jobs (and `pr-kubekins` will need to have service account actor on `kops-nodes`.

We would either need to figure out who owns `gs://k8s-kops-gce` and give `kops-nodes` RW access to it or we could create a new bucket `gs://k8s-kops-gce-[v1beta1]` and I can adjust the GCE tests to use this as state store and etcd backups. 

Hopefully that makes sense, happy to work through this together here or on slack (@geojaz)! 

cc @justinsb

geojaz:
/close 
closing in favor of #17558 to clarify that the code enclosed here is not useful and help is sought

PR #17476, opened 2020-05-01
releng: Split out building for debian-{base,iptables} and pause images

justaugustus:
Needs to merge ahead of https://github.com/kubernetes/kubernetes/pull/90665.

/assign @dims @BenTheElder

### DIFF
--- a/config/jobs/image-pushing/k8s-staging-build-image.yaml
+++ b/config/jobs/image-pushing/k8s-staging-build-image.yaml
@@ -1,12 +1,12 @@
 postsubmits:
   kubernetes/kubernetes:
-    - name: post-kubernetes-push-build-images
+    - name: post-kubernetes-push-image-debian-base
       cluster: test-infra-trusted
       annotations:
         testgrid-dashboards: sig-release-releng-informing, sig-release-master-informing, sig-release-image-pushes
         testgrid-alert-email: release-managers@kubernetes.io
       decorate: true
-      run_if_changed: '^build\/(debian-base|debian-iptables|pause)\/'
+      run_if_changed: '^build\/debian-base\/'
       branches:
         - ^master$
       spec:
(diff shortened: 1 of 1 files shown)


dims:
So here's an issue in workflows:
If we need to bump debian-base, we invariably need an update to debian-iptables. So they are joined at the hip. Which means we have to bump debian-base, ensure it gets built, promote it AND then bump debian-iptables similarly. Also in the meanwhile the 2 build systems are still using old versions of both images, so we have to have some ways of highlighting that this so we don't drop the ball (this actually happened the last time debian-base was updated, so not theoretical)

pause is fine, it does not depend on anything.
if we need to bump debian-iptables by itself, that is fine as well.

justaugustus:
@dims -- yep, I _think_ we're on the same page. See [my comment](https://github.com/kubernetes/kubernetes/pull/90665#issuecomment-622302764) on the accompanying k/k PR:

> /assign @dims @BenTheElder
> cc: @tallclair @kubernetes/release-engineering
> 
> I still need to cleanup more of `dependencies.yaml`. After that, we'll:
> 
>     * [ ]  bump the `debian-base` and `pause` image tags (in this PR)
> 
>     * [ ]  merge the updated image building jobs: [kubernetes/test-infra#17476](https://github.com/kubernetes/test-infra/pull/17476)
> 
>     * [ ]  merge this PR
> 
>     * [ ]  promote the `debian-base` and `pause` staging images
> 
> 
> Then in a follow-up PR:
> 
>     * [ ]  search/replace image references to `us.gcr.io/k8s-artifacts-prod/build-image/**` (since the [vanity domain flip](https://github.com/kubernetes/release/issues/270) hasn't happened yet)
> 
>     * [ ]  bump the `debian-iptables` tag (doesn't make sense to bump it in this PR since we'll need the `debian-base` to have been promoted first)
> 
>     * [ ]  merge the subsequent PR
> 
>     * [ ]  promote the `debian-iptables` image
> 
> 
> Some selective cherry picks may be needed as well.

In https://github.com/kubernetes/kubernetes/commit/0b22d7e22a709b39c7386c3de5cd6750ed7e038d I added the images as part of the dependencies.yaml, which means `pull-kubernetes-verify` will fail if someone misses bumping things everywhere.

You can see it failing correctly (because the `debian-base` versions are mismatched) [here](https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/90665/pull-kubernetes-verify/1256138192818540545):

```
2020/05/01 08:36:46 ERROR: /home/prow/go/src/k8s.io/kubernetes/build/dependencies.yaml indicates that k8s.gcr.io/debian-base should be at version 2.0.0, but the following files didn't match:

build/debian-base/Makefile
build/debian-iptables/Makefile

if you are changing the version of k8s.gcr.io/debian-base, make sure all of the following files are updated with the newest version including /home/prow/go/src/k8s.io/kubernetes/build/dependencies.yaml
then run ./hack/verify-external-dependencies-version.sh

exit status 1
```

dims:
Ack thanks @justaugustus

Issue #17474, opened 2020-04-30
Segfault on prow/checkconfig


jwendell:
**What happened**:

Segfault when running Prow's `checkconfig` :
```
$ checkconfig -config-path config.gen.yaml 
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x90 pc=0x5b6ba2]

goroutine 1 [running]:
regexp.(*Regexp).doExecute(0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0xc000349d70, 0xb, 0x0, 0x0, ...)
	/usr/lib/golang/src/regexp/exec.go:527 +0x562
regexp.(*Regexp).doMatch(...)
	/usr/lib/golang/src/regexp/exec.go:514
regexp.(*Regexp).MatchString(...)
	/usr/lib/golang/src/regexp/regexp.go:507
k8s.io/test-infra/prow/config.Brancher.ShouldRun(0x0, 0x0, 0x0, 0xc000255dc0, 0x2, 0x4, 0x0, 0x0, 0xc000349d70, 0xb, ...)
	/work/go/pkg/mod/k8s.io/test-infra@v0.0.0-20200430202618-cb59ea572ac9/prow/config/jobs.go:251 +0x144
k8s.io/test-infra/prow/config.Presubmit.CouldRun(...)
	/work/go/pkg/mod/k8s.io/test-infra@v0.0.0-20200430202618-cb59ea572ac9/prow/config/jobs.go:341
k8s.io/test-infra/prow/config.BranchRequirements(0xc000349d70, 0xb, 0xc0004d1500, 0x2, 0x4, 0x0, 0xc000617bb8, 0x1, 0x1, 0xc0004d1500, ...)
	/work/go/pkg/mod/k8s.io/test-infra@v0.0.0-20200430202618-cb59ea572ac9/prow/config/branch_protection.go:379 +0x199
k8s.io/test-infra/prow/config.Config.GetTideContextPolicy(0xc000346140, 0x2, 0x4, 0xc0005b7080, 0xc0005b6e70, 0x0, 0x0, 0x0, 0xc0005b7710, 0x1c6ead8, ...)
	/work/go/pkg/mod/k8s.io/test-infra@v0.0.0-20200430202618-cb59ea572ac9/prow/config/tide.go:522 +0x5aa
main.validateTideContextPolicy(0xc0001c7200, 0x0, 0x0)
[...]

PR #17477, opened 2020-05-01
Checkconfig: Prevent validateUnknownFields from manipulating config

alvaroaleman:
We re-used the existing config variables in validateUnknownFields as
    target for a yaml.Unmarshal, resulting in unrelated checks failing
    because fields we normally default get overwritten.

Fixes https://github.com/kubernetes/test-infra/issues/17474
/assign @stevekuznetsov 
/cc @jwendell

### DIFF
--- a/prow/cmd/checkconfig/main.go
+++ b/prow/cmd/checkconfig/main.go
@@ -195,6 +195,21 @@ func main() {
 		logrus.Fatalf("Error parsing options - %v", err)
 	}
 
+	if err := validate(o); err != nil {
+		switch e := err.(type) {
+		case utilerrors.Aggregate:
+			reportWarning(o.strict, e)
+		default:
+			logrus.WithError(err).Fatal("Validation failed")
+		}
+
+	} else {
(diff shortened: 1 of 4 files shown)


jwendell:
@alvaroaleman I confirm with this patch the crash is gone. Thanks!

prow/cmd/checkconfig/main.go
```diff
@@ -300,18 +315,18 @@ func main() {
 	if o.warningEnabled(unknownFieldsWarning) {
 		cfgBytes, err := ioutil.ReadFile(o.configPath)
 		if err != nil {
-			logrus.WithError(err).Fatal("Error loading Prow config for validation.")
+			return fmt.Errorf("error reading Prow config for validation: %w", err)
 		}
-		if err := validateUnknownFields(cfg, cfgBytes, o.configPath); err != nil {
```
stevekuznetsov, line 305: Would have helped to bring this out into it's own commit, took forever to figure out what was going on in here

PR #17478, opened 2020-05-01
Job validation: Check for duplicate contexts

alvaroaleman:
This PR extends the job validation to check for duplicate contexts. Also while I was it I:

* Added tests for `validatePresubmits` and `validatePostsubmits`
* Extended `validatePostsubmit` to check for invalid reporting config. We only validated that for presubmits, as we used to only report on presubmits

/assign @stevekuznetsov

### DIFF
--- a/prow/config/config_test.go
+++ b/prow/config/config_test.go
@@ -4184,8 +4184,8 @@ func TestValidatePresubmits(t *testing.T) {
 		{
 			name: "Duplicate jobname causes error",
 			presubmits: []Presubmit{
-				{JobBase: JobBase{Name: "a"}},
-				{JobBase: JobBase{Name: "a"}},
+				{JobBase: JobBase{Name: "a"}, Reporter: Reporter{Context: "foo"}},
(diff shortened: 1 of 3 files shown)


prow/config/config_test.go
```diff
@@ -4165,3 +4165,102 @@ func TestDefaultAndValidateReportTemplate(t *testing.T) {
...
+			},
+			expectedError: `[jobs b and a report to the same GitHub context "repeated", jobs a and b report to the same GitHub context "repeated"]`,
+		},
+		{
+			name: "Duplicate jobname causes error",
+			presubmits: []Presubmit{
+				{JobBase: JobBase{Name: "a"}, Reporter: Reporter{Context: "foo"}},
+				{JobBase: JobBase{Name: "a"}, Reporter: Reporter{Context: "bar"}},
+			},
+			expectedError: "duplicated presubmit job: a",
```
stevekuznetsov, line 4190: We technically allow this if they don't intersect on branch config, https://github.com/kubernetes/test-infra/issues/9948

alvaroaleman: Yeah I was lazy. Added a test for that as well.

PR #17482, opened 2020-05-01
Clarify tiny bit of documentation for Prow

coryrc:


prow/build_test_update.md
```diff
@@ -31,7 +31,7 @@ Assuming your prow components have multiple replicas, this will result in no dow
 
 Update your deployment (optionally build/pushing the image) to a new image with:
 ```shell
-# export PROW_REPO_OVERRIDE=gcr.io/k8s-prow  # optionally override project
+# export PROW_REPO_OVERRIDE=gcr.io/k8s-prow  # optionally override which registry is used
```
fejta, line 34: I'm not actually sure we want people to do anything other than override the project:
* Aside from `gcr.io`, what other registry systems will work here?
* Is there some reason I cannot set this to be `gcr.io/k8s-prow/beta`, for example?

coryrc: You mean, we should suggest the user only override the PROJECT variable (as is used in the Makefile) instead of the PROW_REPO_OVERRIDE value?

hack/print-workspace-status.sh
```diff
@@ -13,6 +13,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# This generates environment variables and is used by Bazel for that purpose
```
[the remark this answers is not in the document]
coryrc: So a better description is it takes in environment variables and outputs data which will become key-value pairs in bazel?

prow/build_test_update.md
coryrc: Actually, that won't work because while the Makefile in this directory creates PROW_REPO_OVERRIDE from constituent parts, bazel only print-workspace-status.sh which does not create the variable from parts.

coryrc: (or does it rerun that hack script on every call to bazel? so then it would work with PROJECT? this is really confusing which is why I'm trying to document it)

coryrc: `gcr.io/foo` is not a project; that's the confusion. It is a path to a registry.

`foo` does not need to be a project either; the name of the registry path is independent of paths.

PROW_REPO_OVERRIDE is [created in the Makefile](https://github.com/kubernetes/test-infra/blob/6e7f9252d5fae7d74fd9ffabe9175078240000ca/prow/Makefile#L33) from PROJECT and REGISTRY, so when you say "change the project", it sounds like you are asking that PROJECT env var should be changed.

So the comment about changing project is just confusing.

So the comment could be changed to:
`# export PROW_REPO_OVERRIDE=gcr.io/k8s-prow  # optionally change k8s-prow to your project`
which make it clearer what you are trying to say?

Issue #17463, opened 2020-04-30
bugzilla: clone does not set subcomponent when necessary, message is unclear


stevekuznetsov:
See https://github.com/openshift/machine-api-operator/pull/579#issuecomment-621903082

The bugs are that we:

1. do not set the subcomponent when we need to
2. tell the user that the failure is for "bug 0" 

/assign @AlexNPavel

alvaroaleman:
Fixed in https://github.com/kubernetes/test-infra/pull/17483
/close

PR #17483, opened 2020-05-01
bugzilla: Fix and re-enable bugzilla plugin cherrypick support

AlexNPavel:
This reverts the revert of bugzilla plugin cherrypick support and fixes the problems that caused the changes to be reverted.

/cc @stevekuznetsov

### DIFF
--- a/prow/bugzilla/client.go
+++ b/prow/bugzilla/client.go
@@ -28,6 +28,7 @@ import (
 
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/sirupsen/logrus"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 )
 
 const (
@@ -95,17 +96,18 @@ func (c *client) GetBug(id int) (*Bug, error) {
 }
 
 func getClones(c Client, bug *Bug) ([]*Bug, error) {
+	var errs []error
 	clones := []*Bug{}
 	for _, dependentID := range bug.Blocks {
 		dependent, err := c.GetBug(dependentID)
 		if err != nil {
-			return nil, fmt.Errorf("Failed to get dependent bug #%d: %v", dependentID, err)
+			errs = append(errs, fmt.Errorf("Failed to get dependent bug #%d: %v", dependentID, err))
 		}
 		if dependent.Summary == bug.Summary {
 			clones = append(clones, dependent)
 		}
 	}
-	return clones, nil
+	return clones, utilerrors.NewAggregate(errs)
 }
 
 // GetClones gets the list of bugs that the provided bug blocks that also have a matching summary.
@@ -136,6 +138,11 @@ func (c *client) GetSubComponentsOnBug(id int) (map[string][]string, error) {
 	if err := json.Unmarshal(raw, &parsedResponse); err != nil {
 		return nil, fmt.Errorf("could not unmarshal response body: %v", err)
 	}
+	// if there is no subcomponent, return an empty struct
+	if parsedResponse.Bugs == nil || len(parsedResponse.Bugs) == 0 {
+		return map[string][]string{}, nil
+	}
+	// make sure there is only 1 bug
 	if len(parsedResponse.Bugs) != 1 {
 		return nil, fmt.Errorf("did not get one bug, but %d: %v", len(parsedResponse.Bugs), parsedResponse)
 	}
--- a/prow/bugzilla/client.go
+++ b/prow/bugzilla/client.go
@@ -161,13 +161,15 @@ func (c *client) UpdateBug(id int, update BugUpdate) error {
 	return err
 }
 
+// CreateBug creates a new bug on the server.
+// https://bugzilla.readthedocs.io/en/latest/api/core/v1/bug.html#create-bug
 func (c *client) CreateBug(bug *BugCreate) (int, error) {
 	logger := c.logger.WithFields(logrus.Fields{methodField: "CreateBug", "bug": bug})
 	body, err := json.Marshal(bug)
 	if err != nil {
 		return 0, fmt.Errorf("failed to marshal create payload: %v", err)
 	}
-	req, err := http.NewRequest(http.MethodPost, fmt.Sprintf("%s/rest/bug/", c.endpoint), bytes.NewBuffer(body))
+	req, err := http.NewRequest(http.MethodPost, fmt.Sprintf("%s/rest/bug", c.endpoint), bytes.NewBuffer(body))
 	if err != nil {
 		return 0, err
 	}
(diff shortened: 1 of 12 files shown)


prow/bugzilla/client.go
```diff
@@ -205,6 +257,9 @@ func cloneBugStruct(bug *Bug, comments []Comment) *BugCreate {
 		TargetMilestone: bug.TargetMilestone,
 		Version:         bug.Version,
 	}
+	if len(subcomponents) > 0 {
```
alvaroaleman, line 260: I don't think the check is needed, you can just undonditionally set the subcomponents as you construct a new `BugCreate` struct anyways

prow/bugzilla/client.go
```diff
@@ -92,6 +94,54 @@ func (c *client) GetBug(id int) (*Bug, error) {
 	return parsedResponse.Bugs[0], nil
 }
 
+func getClones(c Client, bug *Bug) ([]*Bug, error) {
+	clones := []*Bug{}
+	for _, dependentID := range bug.Blocks {
+		dependent, err := c.GetBug(dependentID)
+		if err != nil {
+			return nil, fmt.Errorf("Failed to get dependent bug #%d: %v", dependentID, err)
```
alvaroaleman, line 102: Can we not return on the first error but instead aggregate the errors (via a `[]error` and `utilerrors.NewAggregate`) and just `continue` on err here?

prow/bugzilla/client.go
```diff
@@ -225,11 +278,15 @@ func cloneBugStruct(bug *Bug, comments []Comment) *BugCreate {
 // clone handles the bz client calls for the bug cloning process and allows us to share the implementation
 // between the real and fake client to prevent bugs from accidental discrepencies between the two.
 func clone(c Client, bug *Bug) (int, error) {
+	subcomponents, err := c.GetSubComponentsOnBug(bug.ID)
```
alvaroaleman, line 281: Will this error on a bugzilla instance that does not have subcomponents?

AlexNPavel: Yes, I need to fix that. The server doesn't return an error, but it returns the standard empty bugzilla struct (which should result in `SubComponents` = `nil`, which would result in an error in the function).

prow/bugzilla/client.go
AlexNPavel: I think this might lead to errors when creating a bug on a server that does not have the `sub_components` type. However, a `nil` check might be better.

prow/bugzilla/client.go
AlexNPavel: That being said, I'm not even sure if this works on non-Red Hat bugzilla at all at this point. I tried running this on other bugzillas, and those seem to return `component` as a `string` (matching what the documentation says it should be) vs. the `[]string` that the Red Hat bugzilla server returns, meaning that we can't unmarshal the json from other bugzilla servers...

prow/bugzilla/client.go
alvaroaleman: if the thing is nil you just put a nilpointer into the field which won't get serialized. I guess the current approach makes sense to cover cases where we have an empty but non-nil `subcomponents` map as that would get serialized?

AlexNPavel: I made the `GetSubcomponentsOnBug` function return an empty map for both nil and empty map responses from the bugzilla server. The current approach makes sense to ignore both empty and non-existent `sub_components` now.

alvaroaleman: Yeah I agree the current approach makes sense to deal with empty objects but why does that require `GetSubcomponentsOnBug` to return an empty map rather than nil?

prow/plugins/bugzilla/bugzilla.go
```diff
@@ -845,7 +845,7 @@ func handleCherrypick(e event, gc githubClient, bc bugzilla.Client, options plug
 	pr, err := gc.GetPullRequest(e.org, e.repo, e.cherrypickFromPRNum)
 	if err != nil {
 		log.WithError(err).Warn("Unexpected error getting title of pull request being cherrypicked from.")
-		return comment(formatError(fmt.Sprintf("creating a cherry-pick bug in Bugzilla: failed to check the state of cherrypicked pull request at https://github.com/%s/%s/pull/%d", e.org, e.repo, e.cherrypickFromPRNum), bc.Endpoint(), e.bugId, err))
+		return comment(fmt.Sprintf("Error creating a cherry-pick bug in Bugzilla: failed to check the state of cherrypicked pull request at https://github.com/%s/%s/pull/%d: %v.\nPlease contact an administrator to resolve this issue, then request a bug refresh with <code>/bugzilla refresh</code>.", e.org, e.repo, e.cherrypickFromPRNum, err))
```
stevekuznetsov, line 848: Why?

AlexNPavel: `formatError` needs a bug number and includes it in its error comment. This was one of the bugs here: https://github.com/kubernetes/test-infra/issues/17463

The new error message is the same as before, but with the reference to "bug 0" removed.

prow/plugins/bugzilla/bugzilla.go
```diff
@@ -819,6 +839,78 @@ func handleMerge(e event, gc githubClient, bc bugzilla.Client, options plugins.B
...
+		return nil
+	} else if err != nil {
+		// should be impossible based on the regex
+		log.WithError(err).Debugf("Failed to get bug ID from PR title \"%s\"", pr.Title)
+		return comment(fmt.Sprintf("Error creating a cherry-pick bug in Bugzilla: could not get bug ID from PR title \"%s\": %v", pr.Title, err))
+	}
+	oldLink := fmt.Sprintf(bugLink, bugID, bc.Endpoint(), bugID)
+	// Since getBug generates a comment itself, we have to add a prefix explaining that this was a cherrypick attempt to the comment
+	commentWithPrefix := func(body string) error {
+		return comment(fmt.Sprintf("Failed to create a cherry-pick bug in Bugzilla: %s", body))
```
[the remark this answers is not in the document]
AlexNPavel: This is how the rest of the bugzilla plugin is written. We could refactor this at some point to make it cleaner and more reasonable. But for that's not a task for this PR.

prow/plugins/bugzilla/bugzilla.go
```diff
@@ -819,6 +839,78 @@ func handleMerge(e event, gc githubClient, bc bugzilla.Client, options plugins.B
...
+		return comment(formatError(fmt.Sprintf("creating a cherry-pick bug in Bugzilla: encountered error cloning %s for cherrypick", oldLink), bc.Endpoint(), e.bugId, err))
+	}
+	cloneLink := fmt.Sprintf(bugLink, cloneID, bc.Endpoint(), cloneID)
+	// Update the version of the bug to the target release
+	update := bugzilla.BugUpdate{
+		Version: targetVersion,
+	}
+	err = bc.UpdateBug(cloneID, update)
+	if err != nil {
+		log.WithError(err).Debugf("Unable to update version and dependencies for bug %d", cloneID)
```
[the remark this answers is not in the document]
AlexNPavel: This is consistent with the style of the rest of the plugin. We log an error when we receive an error from a function and then we comment on the PR with that error. If commenting has an error, the hook logs that error with `Logger.WithError(err).Error("Error handling ReviewEvent.)` (substitute whatever event occurred there) and then sends the error to prometheus. Similar to my comment above, this would require a larger refactor of the code for this plugin.

PR #17490, opened 2020-05-02
Get external IPs from more sources

justinsb:
v4.ifconfig.co seems to be down, so add another source into the
mix (running on cloudflare workers)


justinsb:
cc @rifelpet  as I think you last touched this logic

PR #17492, opened 2020-05-02
enable nfs tests on kubernetes master branch

BenTheElder:
these should work now

will follow up with other branches
/cc @justaugustus @amwat


BenTheElder:
tested via temp commit in kind CI overriding the skip, and locally on my machine.

PR #17493, opened 2020-05-02
skip less tests on kind

BenTheElder:
/cc @aojea @spiffxp @amwat 

Notes:
- we still need to fix NFS tests on IPv6, but already enabled them on IPv4 in https://github.com/kubernetes/test-infra/pull/17492
- I have a tentative plan for loadbalancer / should support a proxy, but I need to prove it out after some sleep :+)
- `\[Slow\]|\[Disruptive\]|\[Flaky\]|\[Feature:.+\]` are totally standard presubmit skips
- `\[Serial\]` is skipped by virtue of being parallel testing (also standard in presubmit)
- `PodSecurityPolicy` is on the way out and not going to work on kind, and literally one test case
- `subPath should support existing` seems really slow ... will follow up on that

The delta vs `pull-kubernetes-e2e-gce` should now be down to:
`PodSecurityPolicy|LoadBalancer|load.balancer|Simple.pod.should.support.exec.through.an.HTTP.proxy|subPath.should.support.existing`

### DIFF
--- a/config/jobs/kubernetes-sigs/kind/kind-presubmits.yaml
+++ b/config/jobs/kubernetes-sigs/kind/kind-presubmits.yaml
@@ -79,7 +79,7 @@ presubmits:
         # NOTE: changes should filter down to pull-kubernetes-e2e-kind-canary
         # and then pull-kubernetes-e2e-kind
         - name: SKIP
-          value: \[Slow\]|\[Disruptive\]|\[Flaky\]|\[Feature:.+\]|PodSecurityPolicy|LoadBalancer|load.balancer|Simple.pod.should.support.exec.through.an.HTTP.proxy|subPath.should.support.existing|should.provide.basic.identity
+          value: \[Slow\]|\[Disruptive\]|\[Flaky\]|\[Feature:.+\]|PodSecurityPolicy|LoadBalancer|load.balancer|Simple.pod.should.support.exec.through.an.HTTP.proxy|subPath.should.support.existing
         - name: PARALLEL
           value: "true"
         - name: BUILD_TYPE
--- a/config/jobs/kubernetes-sigs/kind/kind-presubmits.yaml
+++ b/config/jobs/kubernetes-sigs/kind/kind-presubmits.yaml
@@ -79,7 +79,7 @@ presubmits:
         # NOTE: changes should filter down to pull-kubernetes-e2e-kind-canary
         # and then pull-kubernetes-e2e-kind
         - name: SKIP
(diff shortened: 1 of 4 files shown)


config/jobs/kubernetes-sigs/kind/kind-presubmits.yaml
```diff
@@ -79,7 +79,7 @@ presubmits:
         # NOTE: changes should filter down to pull-kubernetes-e2e-kind-canary
         # and then pull-kubernetes-e2e-kind
         - name: SKIP
-          value: \[Slow\]|\[Disruptive\]|\[Flaky\]|\[Feature:.+\]|PodSecurityPolicy|LoadBalancer|load.balancer|Network.should.set.TCP.CLOSE_WAIT.timeout|Simple.pod.should.support.exec.through.an.HTTP.proxy|subPath.should.support.existing|ReplicationController.light.Should.scale.from.1.pod.to.2.pods|should.provide.basic.identity|\[NodeFeature:PodReadinessGate\]
+          value: \[Slow\]|\[Disruptive\]|\[Flaky\]|\[Feature:.+\]|PodSecurityPolicy|LoadBalancer|load.balancer|Network.should.set.TCP.CLOSE_WAIT.timeout|Simple.pod.should.support.exec.through.an.HTTP.proxy|subPath.should.support.existing|should.provide.basic.identity|\[NodeFeature:PodReadinessGate\]
```
BenTheElder: this skip value does nothing anyhow, HPA has been a feature test since 1.16 thanks to @spiffxp

config/jobs/kubernetes-sigs/kind/kind-release-blocking.yaml
```diff
@@ -35,7 +35,7 @@ periodics:
         value: "."
       # TODO(bentheelder): reduce the skip list further
       - name: SKIP
-        value: \[Slow\]|\[Disruptive\]|\[Flaky\]|\[Feature:.+\]|PodSecurityPolicy|LoadBalancer|load.balancer|In-tree.Volumes.\[Driver:.nfs\]|PersistentVolumes.NFS|Network.should.set.TCP.CLOSE_WAIT.timeout|Simple.pod.should.support.exec.through.an.HTTP.proxy|subPath.should.support.existing|ReplicationController.light.Should.scale.from.1.pod.to.2.pods|should.provide.basic.identity|\[NodeFeature:PodReadinessGate\]
```
aojea, line 38: should we remove `Network.should.set.TCP.CLOSE_WAIT.timeout` here too?

BenTheElder: I'm only modifying presubmits for the master branch in this PR*

* Except removing the one skip that actually doesn't matter because it does nothing.

PR #17498, opened 2020-05-04
Make endpoint configurable to support GitHub Enterprise

stpabhi:


label_sync/main.go
```diff
@@ -119,8 +118,8 @@ const (
 var (
 	debug           = flag.Bool("debug", false, "Turn on debug to be more verbose")
 	confirm         = flag.Bool("confirm", false, "Make mutating API calls to GitHub.")
-	endpoint        = flagutil.NewStrings(github.DefaultAPIEndpoint)
```
fejta, line 122: This is already configurable, and we are already configuring it.

https://github.com/kubernetes/test-infra/blob/2e0f8b9ad36b1508a6477f7f0765b32076d41552/config/jobs/kubernetes/test-infra/test-infra-trusted.yaml#L1052-L1053

PR #17508, opened 2020-05-04
Deck runlocal supports listing jobs dynamically

chaodaiG:
Enhancement for running Deck locally, by listing Prow jobs dynamically.

Before this change, Deck runlocal lists jobs from a pregenerated data, so there is no way to debug any job listing functions with runlocal mode. This PR provides a way to use the same code path for listing Prow jobs as running Deck in prod. 

- Made it possible to debug hidden repo related code
- Also reduce the divergence of `localOnlyMain` and `prodOnlyMain`

Fixing: #17509

### DIFF
--- a/prow/cmd/deck/main.go
+++ b/prow/cmd/deck/main.go
@@ -323,8 +323,93 @@ func main() {
 	if runLocal {
 		localDataHandler := staticHandlerFromDir(o.pregeneratedData)
 		fallbackHandler = localDataHandler.ServeHTTP
+
+		if o.spyglass {
+			initSpyglass(cfg, o, mux, nil, nil, nil)
+		}
 	} else {
 		fallbackHandler = http.NotFound
+
+		restCfg, err := o.kubernetes.InfrastructureClusterConfig(false)
+		if err != nil {
+			logrus.WithError(err).Fatal("Error getting infrastructure cluster config.")
+		}
+		mgr, err := manager.New(restCfg, manager.Options{
+			Namespace:          cfg().ProwJobNamespace,
+			MetricsBindAddress: "0",
+			LeaderElection:     false},
+		)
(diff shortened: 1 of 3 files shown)


michelle192837:
Sorry for the delay! I think this is okay, but I'd be more confident in approving if you can split up the moved code and dynamic listing in separate commits? (Having a hard time figuring out what's existing code vs. modifications).

chaodaiG:
> Sorry for the delay! I think this is okay, but I'd be more confident in approving if you can split up the moved code and dynamic listing in separate commits? (Having a hard time figuring out what's existing code vs. modifications).

Makes sense. Inserted a commit before the last commit. Then the commit history of this PR becomes:
- Commit #1: just moving code around
- Commit #2: changes to make runlocal work with dynamic listing Prow jobs

prow/cmd/deck/main.go
```diff
@@ -320,13 +320,26 @@ func main() {
...
 
-		if o.spyglass {
-			initSpyglass(cfg, o, mux, nil, nil, nil)
+		var fjc fakePjListingClientWrapper
+		var pjs prowapi.ProwJobList
+		staticPjsPath := path.Join(o.pregeneratedData, "prowjobs.json")
+		content, err := ioutil.ReadFile(staticPjsPath)
+		if err != nil {
+			logrus.WithError(err).Fatal("Failed to read jobs from prowjobs.json.")
+		} else if err = json.Unmarshal(content, &pjs); err == nil {
```
michelle192837, line 337: I think the 'else if' should just be a new block since an error from the previous command will exit the program anyways? (I wasn't expecting to see 'err == nil' here on first look).

chaodaiG: That's right, fixed

PR #17513, opened 2020-05-05
Remove quotes from kubemark env vars

mm4tt:
I'm 90% sure they are causing the kubemark-up to fail

### DIFF
--- a/config/jobs/kubernetes/sig-scalability/sig-scalability-golang.yaml
+++ b/config/jobs/kubernetes/sig-scalability/sig-scalability-golang.yaml
@@ -51,8 +51,8 @@ periodics:
       - --cluster=gce-golang
       - --env=CL2_ENABLE_PVS=false
       - --env=CL2_LOAD_TEST_THROUGHPUT=50
-      - --env=KUBEMARK_CONTROLLER_MANAGER_TEST_ARGS="--profiling --kube-api-qps=200 --kube-api-burst=200"
-      - --env=KUBEMARK_SCHEDULER_TEST_ARGS="--profiling --kube-api-qps=200 --kube-api-burst=200"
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes/sig-scalability/sig-scalability-golang.yaml
```diff
@@ -51,8 +51,8 @@ periodics:
       - --cluster=gce-golang
       - --env=CL2_ENABLE_PVS=false
       - --env=CL2_LOAD_TEST_THROUGHPUT=50
-      - --env=KUBEMARK_CONTROLLER_MANAGER_TEST_ARGS="--profiling --kube-api-qps=200 --kube-api-burst=200"
-      - --env=KUBEMARK_SCHEDULER_TEST_ARGS="--profiling --kube-api-qps=200 --kube-api-burst=200"
+      - --env=KUBEMARK_CONTROLLER_MANAGER_TEST_ARGS=--profiling --kube-api-qps=200 --kube-api-burst=200
+      - --env=KUBEMARK_SCHEDULER_TEST_ARGS=--profiling --kube-api-qps=200 --kube-api-burst=200
```
wojtek-t, line 55: Are you sure it will work without quotes?

wojtek-t: Won't it leave only the first item?

mm4tt: Given that something like this https://github.com/kubernetes/test-infra/blob/1b7a5c47abf1554895e5aaa464c64db2f5212fa1/config/jobs/kubernetes/sig-scalability/sig-scalability-release-blocking-jobs.yaml#L35
works, I think it should work as well

PR #17523, opened 2020-05-05
Add k8s-infra-prow-build's boskos to monitoring

spiffxp:
I don't have access to the prometheus instance for
monitoring.prow.k8s.io so I'm not sure what the existing metrics look
like. This should work well with the existing boskos resources
dashboard.

I'm not sure how well this will work with the boskos server
dashboard. Do the scraped metrics automatically have a `host` or `target`
tag included in them that could be added as a parameter?

/cc @cjwagner @stevekuznetsov

### DIFF
--- a/config/prow/cluster/monitoring/mixins/grafana_dashboards/boskos.jsonnet
+++ b/config/prow/cluster/monitoring/mixins/grafana_dashboards/boskos.jsonnet
@@ -39,7 +39,7 @@ dashboard.new(
         {type: "gke-project", friendly: "GKE project"},
         {type: "gpu-project", friendly: "GPU project"},
         {type: "ingress-project", friendly: "Ingress project"},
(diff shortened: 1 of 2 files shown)


config/prow/cluster/monitoring/mixins/grafana_dashboards/boskos.jsonnet
```diff
@@ -39,7 +39,7 @@ dashboard.new(
         {type: "gke-project", friendly: "GKE project"},
         {type: "gpu-project", friendly: "GPU project"},
         {type: "ingress-project", friendly: "Ingress project"},
-        {type: "istio-project", friendly: "Istio project"},
```
spiffxp, line 42: istio-project pool is unused at this point thanks to https://github.com/kubernetes/test-infra/pull/17213

PR #17526, opened 2020-05-06
Mark pull-kubernetes-conformance-kind-ga-only-parallel required

spiffxp:
aka make the job merge-blocking

This job caught conformance tests exercising something that wasn't marked as GA yet (ref: https://github.com/kubernetes/kubernetes/pull/90367#issuecomment-624283211)

A postsubmit would have caught this eventually, but this seems like it's been flake free enough to consider making it merge-blocking.

WDYT?

/cc @BenTheElder


spiffxp:
FYI @kubernetes/ci-signal @kubernetes/release-team

PR #17527, opened 2020-05-06
Make needs-rebase respect GitHub's mergestatestatus field

DoomGerbil:
It turns out that, based on some GitHub repo settings, the `needs-rebase` service can't always tell if a PR is actually mergeable, because GitHub's API tells lies in some cases.

The response from `GetPullRequest()` has a field literally called `mergeable` that is _sometimes_ wrong - it only considers git merge-ability, not GitHub merge-ability. This means that if GitHub repo has the "Require branches to be up-to-date before merging" setting enabled, and a PR is mergeable by git, but not up-to-date with `master`, then:

- `needs-rebase` will not mark it as requiring a rebase.
- `tide` will see that the PR meets mergeability criteria, add it to a subpool, and attempt to merge it.

Unfortunately, `tide` will get back an error from GitHub when it tries to merge the PR, resulting in all PRs behind this one in the subpool being blocked from merging as well.

This PR changes the behavior of `needs-rebase` so that it will now look at the value of the `mergeable_state` field in the `GetPullRequest()` response, which contains a much more granular mergeability status, including the state `behind` to describe the scenario I outlined above.

It also turns out that there are other states in that field that equate to "Not Mergeable" as well.

The upshot of this PR is that when `needs-rebase` asks the Prow GitHub client if a PR is mergeable, the client will say "No" in the cases where git might let it merge, but GitHub won't.

I added tests around this functionality (there were no existing tests covering `IsMergeable()` at all), and we are using this functionality in `improbable`'s internal Prow deployment already.

### DIFF
--- a/prow/github/client.go
+++ b/prow/github/client.go
@@ -3215,9 +3215,9 @@ func (c *client) IsMergeable(org, repo string, number int, SHA string) (bool, er
 			switch pr.MergeableState {
 			case MergeableStateBehind, MergeableStateBlocked, MergeableStateDraft, MergeableStateUnknown:
 				c.logger.WithFields(logrus.Fields{
-					"repo":				fmt.Sprintf("%v/%v", org, repo),
-					"pr":      			number,
-					"mergeablestate": 	pr.MergeableState,
+					"repo":           fmt.Sprintf("%v/%v", org, repo),
+					"pr":             number,
+					"mergeablestate": pr.MergeableState,
 				}).Infof("PR is in unmergeable state.")
 				return false, nil
 			default:
--- a/prow/github/client.go
+++ b/prow/github/client.go
@@ -3210,7 +3210,19 @@ func (c *client) IsMergeable(org, repo string, number int, SHA string) (bool, er
 			return false, errors.New("pull request was merged while checking mergeability")
 		}
 		if pr.Mergable != nil {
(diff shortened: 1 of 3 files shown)


DoomGerbil:
Hey there - I'm not sure if there's anyone who can help with this, but I am unable to sign the CLA because the CNCF CLA singup page (https://identity.linuxfoundation.org/projects/cncf/employee-signup) seems to be broken.  My company already has a signed CLA, but I need to be added to it, I just cannot get in touch with the CNCF because when I attempt to contact them via the form on the site, I get back an error, and there is no other contact mechanism listed.

Screenshot 2020-05-06 at 18 53 35

Any idea who can help unbreak the CNCF CLA signup site?

alvaroaleman:
> This means that if GitHub repo has the "Require branches to be up-to-date before merging" setting enabled, and a PR is mergeable by git, but not up-to-date with master, then:

If you are using Tide, you should disable that setting. That setting is a hacky workaround for the fact that most (all?) of the other CI systems test a given PR once and consider its tests to be good ifthey succeeded once even though the base branch can move in the meantime, effectively invalidating the test result.

Tide however always does retesting and will only merge things that have green tests against the latest revision of the base branch. Hence, that setting isn't needed and as you noted may cause issues.

PR #17537, opened 2020-05-06
Switch temporary deck_deployment.yaml version to one that works with Gerrit.

cjwagner:
Maybe not needed at this point, but cleaner to fix.
/assign @clarketm @BenTheElder


BenTheElder:
/lgtm
/approve
thanks!

PR #17539, opened 2020-05-06
Added alert email for GCE Import/Export tests (Daisy)

zoran15:


chases2:
Can we move this dashboard into a separate subdirectory? A new 'Google' or 'google-gce' directory may be appropriate.

zoran15:
> Can we move this dashboard into a separate subdirectory? A new 'Google' or 'google-gce' directory may be appropriate.

I'm not sure. I don't have much context for moves like this but it's something we can look into within my team and any groups that might be affected. I can create an issue to track work on this effort.

In the meantime, can we get this change approved so my team is aware of any test failures? We've had situations when it's been days before we realized a test failed which caused customers to be affected.

chases2:
Moving it to a subdirectory shouldn't be too hard; there's an abbreviated how-to at the top of this config file. If you have issues, I can help on the #sig-testing channel on Slack

PR #17540, opened 2020-05-07
config: use fsnotify-based watcher for config agent

AlexNPavel:
This PR changes the watching method for the prow config agent from a polling-based watcher that checks last modification time for the files to an fsnotify event-based watcher that either watches for changes to the files if they are not configmap mounted or watches for `CREATE` events for the directory's `..data` directory if they are (a `CREATE` event on the `..data` directory is the last event that takes place when k8s updates a configmap mounted directory).

This code is based on the watcher used in the config and registry agents in `openshift/ci-tools` (https://github.com/openshift/ci-tools/blob/master/pkg/watcher/watcher.go).

The functions that create the watch functions take both event and error functions for the watcher as arguments, which allows more customization by the user (for example, `ci-tools` uses this to report errors to Prometheus).

/cc @stevekuznetsov


prow/config/agent.go
```diff
@@ -40,70 +45,186 @@ type Agent struct {
...
-				recentModTime, err := lastConfigModTime(prowConfig, jobConfig)
+	// TODO(AlexNPavel): allow empty jobConfig till fully migrate config to subdirs
+	jobIsCMMounted := false
+	if jobConfig != "" {
+		jobIsCMMounted, err = IsConfigMapMount(filepath.Dir(jobConfig))
+		if err != nil {
+			return err
+		}
+	}
+	if filepath.Dir(prowConfig) == filepath.Dir(jobConfig) && prowIsCMMounted {
```
petr-muller, line 165: Wow, that's a lot of error handling mess that really obfuscates what gets watched how and when... Maybe something like this would improve the readability and reduce duplication?

```
var cms string[]
var notcms string[]
WHAT IS WATCHED WHEN LOGIC
for cm in cms {
   runFunc, err := GetCMMountWatcher(eventFunc, errFunc, cm)
		if err != nil {
			return err
		}
		runFuncs = append(runFuncs, runFunc)
}
for notcm in notcms {
   runFunc, err := GetFileWatcher(eventFunc, errFunc, notcm)
		if err != nil {
			return err
		}
		runFuncs = append(runFuncs, runFunc)
}
```

AlexNPavel: Yeah, that was overly complicated. Simplified based on your suggestions.

AlexNPavel:
This PR should be correct and working now, so I'll mark the PR as ready for review.

However, since this is such a drastic change, I will be manually adding a `hold` on it so that the merge timing can be controlled and allow me to verify that it works properly in production ASAP after merging.

/hold

stevekuznetsov:
Can we add this in such a way that we use both loading mechanisms and report a difference in loaded config as an error?

prow/config/agent.go
```diff
@@ -40,70 +47,217 @@ type Agent struct {
...
 		}
 	}
-	return recentModTime, nil
+	// The prow config is always a single file
+	if prowIsCMMounted, err := IsConfigMapMount(filepath.Dir(prowConfig)); err != nil {
+		return err
+	} else if prowIsCMMounted {
+		cms.Insert(filepath.Dir(prowConfig))
+	} else {
+		dirs.Insert(prowConfig)
```
stevekuznetsov, line 232: You call this `dirs` but insert files

prow/config/agent.go
```diff
@@ -40,70 +47,217 @@ type Agent struct {
...
+		return nil
+	}
+	// We may need to add more directories to be watched
+	dirsEventFunc := func(w *fsnotify.Watcher) error {
+		c, err := Load(prowConfig, jobConfig, additionals...)
+		if err != nil {
+			return err
+		}
+		ca.Set(c)
+		// TODO(AlexNPavel): Is there a chance that a ConfigMap mounted directory may appear without making a new pod? If yes, handle that.
```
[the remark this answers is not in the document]
AlexNPavel: The question here is whether we need to recreate the pod the service is running in if we want to add a new configmap mount or if they can be added without recreating the pod. I don't think we can add configmaps without recreating the pod in which case this is fine, but if we can, the code will have to be updated to handle that.

prow/config/agent.go
AlexNPavel: Hmm. I have the `jobConfig` dirs added as the first items here, `prowConfig` would be the only item that's a file instead. Would `files` be a better name overall?

prow/config/agent.go
```diff
@@ -40,70 +47,249 @@ type Agent struct {
...
+		cms.Insert(filepath.Dir(prowConfig))
+	} else {
+		dirs.Insert(prowConfig)
+	}
+	for cm := range cms {
+		runFunc, err := GetCMMountWatcher(cmEventFunc, errFunc, cm)
+		if err != nil {
+			return err
+		}
+		interrupts.Run(runFunc)
```
petr-muller, line 271: Hm. This ways we start spawning watchers, then possibly hit an error and immediately return with that error, leaving some watchers running and some not. If that is acceptable and desired, we should accumulate errors and try to spawn the rest of the watchers. If not we should IMO prepare first and start spawning only if we encountered no errors.

AlexNPavel: Fixed

prow/config/agent.go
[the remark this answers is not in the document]
stevekuznetsov: No, you must edit the PodSpec and restart the Pod to mount new ConfigMaps

prow/config/agent.go
stevekuznetsov: Lol maybe `inodes` ... it's fine as is

PR #17542, opened 2020-05-07
Allow k8s-infra approval of image-pushing jobs

spiffxp:
- Fixup two jobs that couldn't previously run on k8s-infra-prow-build-trusted
- Allow people who have kubernetes.io org admin privileges to act as k8s-infra-oncall until we have a better group for it
- Allow k8s-infra-oncall to approve image-pushing jobs
- Prevent image-pushing jobs from running on test-infra-cluster (so that k8s-infra people aren't approving jobs that have access to more secrets than they know about)

/cc @fejta @cjwagner
for the test-infra side of things

/wg k8s-infra
/cc @cblecker @dims @thockin
heads up

/hold
for comment

### DIFF
--- a/config/jobs/image-pushing/OWNERS
+++ b/config/jobs/image-pushing/OWNERS
@@ -4,8 +4,9 @@ options:
 
 # Oncallers with context on this
 reviewers:
-- Katharine
+- spiffxp
 
 # All oncallers
 approvers:
-- test-infra-oncall # oncall
+- test-infra-oncall
+- wg-k8s-infra-oncall
--- a/config/tests/jobs/jobs_test.go
+++ b/config/tests/jobs/jobs_test.go
@@ -307,48 +307,68 @@ func findRequired(t *testing.T, presubmits []cfg.Presubmit) []string {
 	return required
 }
 
+// Enforce conventions for jobs that run in test-infra-trusted cluster
 func TestTrustedJobs(t *testing.T) {
 	// TODO(fejta): allow each config/jobs/kubernetes/foo/foo-trusted.yaml
 	// that uses a foo-trusted cluster
 	const trusted = "test-infra-trusted"
-	const wgK8sTrusted = "k8s-infra-prow-build-trusted"
 	trustedPath := path.Join(*jobConfigPath, "kubernetes", "test-infra", "test-infra-trusted.yaml")
-	imagePushingDir := path.Join(*jobConfigPath, "image-pushing") + "/"
 
 	// Presubmits may not use trusted clusters.
 	for _, pre := range c.AllStaticPresubmits(nil) {
-		if pre.Cluster == trusted || pre.Cluster == wgK8sTrusted {
+		if pre.Cluster == trusted {
 			t.Errorf("%s: presubmits cannot use trusted clusters", pre.Name)
 		}
 	}
 
 	// Trusted postsubmits must be defined in trustedPath
 	for _, post := range c.AllStaticPostsubmits(nil) {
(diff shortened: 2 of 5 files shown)


dims:
happy to help here. +1

config/jobs/image-pushing/OWNERS
```diff
@@ -4,8 +4,9 @@ options:
...
 # Oncallers with context on this
 reviewers:
-- Katharine
+- spiffxp
 
 # All oncallers
 approvers:
-- test-infra-oncall # oncall
+- test-infra-oncall
+- wg-k8s-infra-oncall
```
fejta, line 12: None of the jobs under config/jobs/image-pushing run under the trusted cluster right?

Can we please fix this before we allow non-oncall to approve these PRs?

https://github.com/kubernetes/test-infra/blob/5d3fe432a2c4dba8361b974bc69f315abe2a7190/config/tests/jobs/jobs_test.go#L316-L338

spiffxp: I think I did that below?

config/tests/jobs/jobs_test.go
```diff
@@ -331,6 +331,11 @@ func TestTrustedJobs(t *testing.T) {
 			if err := validateImagePushingImage(post.Spec); err != nil {
 				t.Errorf("%s defined in %s %s", post.Name, post.SourcePath, err)
 			}
+			// jobs in image-pushing have a wider approval pool than test-infra-oncall
+			// and so shouldn't be alllowed to run in test-infra-trusted cluster
+			if post.Cluster != wgK8sTrusted {
```
fejta, line 336: a) please move references to `wgK8sTrusted` and `image-pushing` out of `TestTrustedJobs`
b) Create a `TestImagePushingJobs` that validates that `image-pushing` jobs always use the `wgK8sTrusted` cluster.

config/jobs/image-pushing/OWNERS
fejta: I would like `TestTrustedJobs` to just validate the trusted jobs that test-infra-oncall approves.

I do not want to make this test more complicated (and therefore fragile) than necessary by making it validate other things (eg jobs in dir `D` use cluster `C`).

spiffxp: Updated, PTAL

config/tests/jobs/jobs_test.go
spiffxp: I opted to make a `TestK8sInfraTrusted`, as the cluster is intended to run more than jobs in image-pushing/, and I'm inclined to leave image-pushing/ where it is for discoverability rather than pull everything under one trusted root

config/tests/jobs/jobs_test.go
```diff
@@ -307,48 +307,68 @@ func findRequired(t *testing.T, presubmits []cfg.Presubmit) []string {
...
 }
 
+// Enforce conventions for jobs that run in test-infra-trusted cluster
 func TestTrustedJobs(t *testing.T) {
 	// TODO(fejta): allow each config/jobs/kubernetes/foo/foo-trusted.yaml
 	// that uses a foo-trusted cluster
 	const trusted = "test-infra-trusted"
-	const wgK8sTrusted = "k8s-infra-prow-build-trusted"
 	trustedPath := path.Join(*jobConfigPath, "kubernetes", "test-infra", "test-infra-trusted.yaml")
-	imagePushingDir := path.Join(*jobConfigPath, "image-pushing") + "/"
```
fejta, line 316: ❤️

PR #17544, opened 2020-05-07
go-runner should be in k8s-staging-build-image GCR bucket

dims:
Whoops. go-runner should be with debian-base/debian-iptables


dims:
/assign @BenTheElder 
/cc @justaugustus 

Sorry folx! got the bucket name wrong.

PR #17548, opened 2020-05-07
Add two more scalability presubmits (verify-lint and verify-dashboard)

oxddr:
1. `verify-lint` is being added via https://github.com/kubernetes/perf-tests/pull/1232
2. `verify-dashboard` is being moved to `python3` in https://github.com/kubernetes/perf-tests/pull/1236

ref https://github.com/kubernetes/perf-tests/issues/573 https://github.com/kubernetes/perf-tests/issues/552


oxddr:
/assign @mm4tt 

I suggest that we get this PR first. For some time presubmit will fail (but it will allows us to merge PRs as it's marked as optional), because there will no make targets. But that will allow us to tests them.

PR #17551, opened 2020-05-07
Make pull-kubernetes-conformance-kind-ga-only-parallel required

liggitt:
The ga-only conformance job has been release blocking and [very reliable](https://testgrid.k8s.io/conformance-all#conformance-ga-only&width=20)

The parallel presubmit has flagged accidental introduction of conformance tests depending on non-GA features, but only informationally. This makes it required.

/sig testing release
/cc @spiffxp @justaugustus 
fyi @BenTheElder


BenTheElder:
https://github.com/kubernetes/test-infra/pull/17526

spiffxp:
/close
as dupe of https://github.com/kubernetes/test-infra/pull/17526, I would have /cc'ed you on that but I wanted to double check with @BenTheElder first

PR #17553, opened 2020-05-07
bugzilla: copy all comments when cloning bug

AlexNPavel:
This changes the way that the description of the bugzilla clone is formatted to be more in line with what bugzilla's web UI clone does.

/cc @alvaroaleman

### DIFF
--- a/prow/bugzilla/client.go
+++ b/prow/bugzilla/client.go
@@ -268,18 +268,32 @@ func cloneBugStruct(bug *Bug, subcomponents map[string][]string, comments []Comm
 	if len(subcomponents) > 0 {
 		newBug.SubComponents = subcomponents
 	}
-	if len(comments) > 0 && comments[0].Count == 0 {
-		desc := comments[0]
-		newBug.Description = fmt.Sprintf("This is a clone of Bug #%d. This is the description of that bug:\n%s", bug.ID, desc.Text)
-		newBug.CommentIsPrivate = desc.IsPrivate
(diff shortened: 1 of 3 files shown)


prow/bugzilla/client.go
```diff
@@ -268,18 +268,32 @@ func cloneBugStruct(bug *Bug, subcomponents map[string][]string, comments []Comm
...
-		newBug.CommentIsPrivate = desc.IsPrivate
-		newBug.CommentTags = desc.Tags
-		newBug.IsMarkdown = desc.IsMarkdown
-	} else {
-		// This cases _shouldn't_ happen. But just in case...
-		newBug.Description = fmt.Sprintf("This is a clone of Bug #%d.", bug.ID)
-		// We don't know whether this bug should be private. Default to private in case this bug contains private information.
-		newBug.CommentIsPrivate = true
+	for _, comment := range comments {
+		if comment.IsPrivate {
```
alvaroaleman, line 272: So if the bug we are cloning contains a single private comment we make the cloned bugs description private? Is that what the UI cloning does as well or does it just omit the private comments?

AlexNPavel: The UI cloning just makes the entire description private if any comments are private.

AlexNPavel: I would've preferred to put all public comments in a public description and then maybe make a second private comment with all the private comments from the parent afterwards. But that could be a bit confusing, and for now we'll just stick with what the UI does.

PR #17565, opened 2020-05-09
aws-janitor: clean up elbv2 resources

ixdy:
I basically copied the existing elb code but updated it to use the elbv2 API. I haven't tested this yet, but I imagine it probably works.

There's lots of boilerplate here. At some point we should probably clean up some of this code to make it more maintainable...

cc @chemikadze


BenTheElder:
you should perhaps update owners here. liz is doing other things now. justin might need a ping if you're looking for his review.

ixdy:
/assign @detiber 
(if you have cycles - this should be pretty straightforward)

ixdy:
Tested this with `aws-resources-list` (since the janitor doesn't have a `--dry-run` mode) and it shows the elbv2 resources appropriately. (I don't have any elb [v1] resources in my account but it's probably still working...)

boskos/aws-janitor/resources/elb.go
```diff
@@ -26,20 +26,20 @@ import (
 	"k8s.io/klog"
 )
 
-// Clean-up ELBs
+// Clean-up Classic ELBs
 
-type LoadBalancers struct{}
+type ClassicLoadBalancers struct{}
```
chemikadze, line 31: What do you think about ELB / ELBv2 instead of (Classic|)LoadBalancers, to be consistent with object naming?

ixdy:
/close

I will reopen this against kubernetes-sigs/boskos.

PR #17566, opened 2020-05-10
Add dockerless verify script to pull-kubernetes-typecheck

dims:
New tag added in https://github.com/kubernetes/kubernetes/pull/87746


wojtek-t:
@dims - I don't have enough context about it. What that is?

dims:
@wojtek-t we have a KEP that builds kubelet without needing `docker/docker` dependency ( please see https://github.com/kubernetes/enhancements/pull/1546 ) This was implemented in https://github.com/kubernetes/kubernetes/pull/87746. As part of that we added a new verify script https://github.com/kubernetes/kubernetes/blob/master/hack/verify-typecheck-dockerless.sh which we need to run in CI

PR #17568, opened 2020-05-11
Fixes to HMAC management phase1

chizhg:
Some fixes to the HMAC management tool phase1, as per the update in the [design doc](https://docs.google.com/document/d/1m3nFfPt95kbe6BQdWp5BrZMSPkBo5mFb2nuMq1OeE24/edit?ts=5e1f65f9#heading=h.1lxac62rvsv0)

1. Configuring repo-level and org-level webhooks at the same time for a repo is not a valid use case, as there will be duplicated webhook events. So remove the related unit tests.

2. `expiry` is not needed in the secret, as we won't want a token to be expired without being noticed. Instead, add a field called `created_at` as it's needed to determine if an existing token needs to be rotated based on the new configurations.

3. Add a new check in the `checkconfig` tool to disallow org-level and repo-level webhooks to be configured at the same time on a given repo.

After these changes, the new secret file will be like:
```
'*':
  - value: key1
    created_at: 2019-10-02T15:00:00Z
  - value: key2
    created_at: 2020-10-02T15:00:00Z
foo/bar:
  - value: 123abc
    created_at: 2019-10-02T15:00:00Z
  - value: key6
    created_at: 2020-10-02T15:00:00Z
```

or 

```
'*':
  - value: key1
    created_at: 2019-10-02T15:00:00Z
  - value: key2
    created_at: 2020-10-02T15:00:00Z
foo:
  - value: 123abc
    created_at: 2019-10-02T15:00:00Z
  - value: key4
    created_at: 2020-10-02T15:00:00Z
```

/cc @cjwagner @clarketm

### DIFF
--- a/prow/ANNOUNCEMENTS.md
+++ b/prow/ANNOUNCEMENTS.md
@@ -101,6 +101,7 @@ Note: versions specified in these announcements may not include bug fixes made
 in more recent versions so it is recommended that the most recent versions are
 used when updating deployments.
 
+ - *May 18, 2020* `expiry` field has been replaced with `created_at` in the HMAC secret.
  - *April 24, 2020* Horologium now defaults to `--dry-run=true`
  - *April 23, 2020* Explicitly setting `--config-path` is now required.
  - *April 23, 2020* Update the `autobump` image to at least `v20200422-8c8546d74` before June 2020.
--- a/prow/cmd/checkconfig/main.go
+++ b/prow/cmd/checkconfig/main.go
@@ -96,6 +96,7 @@ const (
 	jobNameLengthWarning         = "long-job-names"
 	jobRefsDuplicationWarning    = "duplicate-job-refs"
 	needsOkToTestWarning         = "needs-ok-to-test"
+	managedWebhooksWarning       = "managed-webhooks"
 	validateOwnersWarning        = "validate-owners"
 	missingTriggerWarning        = "missing-trigger"
 	validateURLsWarning          = "validate-urls"
@@ -112,6 +113,7 @@ var defaultWarnings = []string{
 	jobNameLengthWarning,
 	jobRefsDuplicationWarning,
 	needsOkToTestWarning,
+	managedWebhooksWarning,
 	validateOwnersWarning,
 	missingTriggerWarning,
 	validateURLsWarning,
@@ -184,7 +186,7 @@ func (o *options) gatherOptions(flag *flag.FlagSet, args []string) error {
 		return fmt.Errorf("parse flags: %v", err)
 	}
 	if err := o.DefaultAndValidate(); err != nil {
-		return fmt.Errorf("Invalid options: %v", err)
+		return fmt.Errorf("invalid options: %v", err)
 	}
 	return nil
 }
@@ -301,6 +303,11 @@ func validate(o options) error {
 	}
 	if o.warningEnabled(needsOkToTestWarning) {
 		if err := validateNeedsOkToTestLabel(cfg); err != nil {
+			errs = append(errs, err)
+		}
+	}
+	if o.warningEnabled(managedWebhooksWarning) {
+		if err := validateManagedWebhooks(cfg); err != nil {
 			errs = append(errs, err)
 		}
 	}
@@ -862,6 +869,28 @@ func validateNeedsOkToTestLabel(cfg *config.Config) error {
 		}
 	}
 	return utilerrors.NewAggregate(queryErrors)
+}
+
+func validateManagedWebhooks(cfg *config.Config) error {
+	var errs []error
+	orgs := sets.String{}
+	for repo := range cfg.ManagedWebhooks {
+		if !strings.Contains(repo, "/") {
+			org := repo
+			orgs.Insert(org)
+		}
+	}
+	for repo := range cfg.ManagedWebhooks {
+		if strings.Contains(repo, "/") {
+			org := strings.SplitN(repo, "/", 2)[0]
+			if orgs.Has(org) {
+				errs = append(errs, fmt.Errorf(
+					"org-level and repo-level webhooks are configured together for %q, "+
+						"which is not allowed as there will be duplicated webhook events", repo))
+			}
+		}
+	}
+	return utilerrors.NewAggregate(errs)
 }
 
 func pluginsWithOwnersFile() string {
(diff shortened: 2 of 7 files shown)


fejta:
Please keep merge commits out of the PR: https://github.com/kubernetes/community/blob/master/contributors/guide/github-workflow.md#4-keep-your-branch-in-sync (aka `git rebase upstream/master && git push -f origin remote-expiry`)

Please squash your commits to the smallest number of logically distinct commits (90% of time this means a single commit) - https://github.com/kubernetes/community/blob/master/contributors/guide/github-workflow.md#squash-commits

AKA "create framework" and "refactor foo to use framework" might be two separate commits but "create framework", "fix bug I found", "typos" commits should all squashed to the same commit.

Fine to add temporary "address review comments" commits to help reviewers, but you'll need to squash these back down before merge.

chizhg:
> Please keep merge commits out of the PR: https://github.com/kubernetes/community/blob/master/contributors/guide/github-workflow.md#4-keep-your-branch-in-sync (aka `git rebase upstream/master && git push -f origin remote-expiry`)
> 
> Please squash your commits to the smallest number of logically distinct commits (90% of time this means a single commit) - https://github.com/kubernetes/community/blob/master/contributors/guide/github-workflow.md#squash-commits
> 
> AKA "create framework" and "refactor foo to use framework" might be two separate commits but "create framework", "fix bug I found", "typos" commits should all squashed to the same commit.
> 
> Fine to add temporary "address review comments" commits to help reviewers, but you'll need to squash these back down before merge.
> 
> (I would expect this PR to start with 1 commit)

Thanks for the links. Done.

prow/cmd/checkconfig/main.go
```diff
@@ -829,6 +837,28 @@ func validateNeedsOkToTestLabel(cfg *config.Config) error {
 	return utilerrors.NewAggregate(queryErrors)
 }
 
+func validateManagedWebhooks(cfg *config.Config) error {
+	var errs []error
+	orgs := sets.String{}
+	for repo := range cfg.ManagedWebhooks {
```
fejta, line 843: Does this make more sense as a CRD? Not sure why we need/want this to live in config.yaml

prow/cmd/checkconfig/main_test.go
```diff
@@ -903,6 +904,67 @@ func TestValidateStrictBranches(t *testing.T) {
 	}
 }
 
+func TestValidateManagedWebhooks(t *testing.T) {
+	var testCases = []struct {
+		name      string
+		config    config.ProwConfig
+		expectErr bool
+	}{{
```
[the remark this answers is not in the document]
chizhg: Fixed.

prow/cmd/checkconfig/main_test.go
```diff
@@ -903,6 +904,67 @@ func TestValidateStrictBranches(t *testing.T) {
...
+				"foo1":     {TokenCreatedAfter: time.Now()},
+				"foo2":     {TokenCreatedAfter: time.Now()},
+				"foo/bar":  {TokenCreatedAfter: time.Now()},
+				"foo/bar1": {TokenCreatedAfter: time.Now()},
+				"foo/bar2": {TokenCreatedAfter: time.Now()},
+			}},
+		expectErr: false,
+	},
+		{
+			name: "has duplicate webhooks",
```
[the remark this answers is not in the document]
chizhg: Thanks for pointing out. Now they are consistent.

prow/github/hmac.go
```diff
@@ -22,16 +22,17 @@ import (
...
+
+	"github.com/sirupsen/logrus"
+	"sigs.k8s.io/yaml"
 )
 
-// hmacSecret contains a hmac token and its expiration time.
+// hmacSecret contains a hmac token and the time when it's created.
 type hmacSecret struct {
-	Value  string    `json:"value"`
-	Expiry time.Time `json:"expiry"`
```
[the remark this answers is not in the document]
chizhg: The last time you checked it for k8s and istio Prow, they were still using the secret in the old format, which is a single plain-text HMAC token, so I assume no one is using it now.
But I have added it to the deprecation notice, just in case.

prow/cmd/checkconfig/main.go
chizhg: I think it will make much more sense to make it a CRD, if we are going to develop a controller to reconcile the secrets. But as we synced with Aaron last time, he was mainly considering this approach for service-account keys and ssh private keys, which is a different use case and didn't consider it as a blocker to the current HMAC rotation design.

prow/ANNOUNCEMENTS.md
```diff
@@ -101,6 +101,7 @@ Note: versions specified in these announcements may not include bug fixes made
 in more recent versions so it is recommended that the most recent versions are
 used when updating deployments.
 
+ - *May 18, 2020* `expiry` field has been removed from the HMAC secret.
```
fejta, line 104: replaced with `created_at`

chizhg: Done.

fejta: squash commits please!

chizhg: Done.
(We usually do not squash commits for Knative as developers find it harder to review new commit changes, but it seems to be a different convention for Kubernetes)

prow/github/hmac.go
```diff
@@ -94,31 +95,29 @@ func extractHmacs(repo string, tokenGenerator func() []byte) ([][]byte, error) {
 		// whole file is a single line hmac token.
 		//TODO : Once this code has been released and file has been moved to new format,
 		// we should delete this code and return error.
-		logrus.WithError(err).Info("couldn't unmarshal the hmac secret as hierarchical file. Parsing as single token format")
+		logrus.WithError(err).Trace("Couldn't unmarshal the hmac secret as hierarchical file. Parsing as single token format")
```
fejta, line 98: How often is `yaml.Unmarshal()` called? It should be at most once per 5m, not every time `ValidatePayload` is called. Will you please prioritize fixing this? Fine to do it in a subsequent PR. See how prow/config/secret package for various ways we do this sort of thing.

chizhg: It will be pretty often as it will happen on every webhook event.
Sure, I will submit a followup PR to fix it ASAP.

chizhg:
/hold cancel
Thanks!

PR #17574, opened 2020-05-12
Add external-dependencies-version check to pull-kubernetes-dependencies

justaugustus:
(Follow-up from a convo w/ @cblecker yesterday.)

tl;dr -- When we make changes to `build/dependencies.yaml`, `pull-kubernetes-verify` appears to not run `hack/verify-external-dependencies-version.sh` until much later in the job.

This PR adds the external deps check to `pull-kubernetes-dependencies`, which is a faster job, and a more appropriate place for this check.

There will be a future PR to k/k to exclude `hack/verify-external-dependencies-version.sh` from `pull-kubernetes-verify` once this has been running successfully for a bit.

/assign @cblecker 
cc: @liggitt @thockin @kubernetes/release-engineering


config/jobs/kubernetes/sig-release/release-branch-jobs/1.15.yaml
```diff
@@ -979,7 +979,7 @@ presubmits:
         - runner.sh
         env:
         - name: WHAT
-          value: vendor vendor-licenses
+          value: external-dependencies-version vendor vendor-licenses
```
cblecker, line 982: Are we still supporting 1.15 jobs?

justaugustus: @cblecker -- For now, yes. We usually blow them away for beta.0, which is right around the corner.

/hold cancel

PR #17575, opened 2020-05-12
Disable NPL in kubemark 1.15

mm4tt:
It was accidentally turned on in https://github.com/kubernetes/perf-tests/pull/1218

Fixes https://github.com/kubernetes/kubernetes/issues/90981

### DIFF
--- a/config/jobs/kubernetes/sig-release/release-branch-jobs/1.15.yaml
+++ b/config/jobs/kubernetes/sig-release/release-branch-jobs/1.15.yaml
@@ -250,6 +250,7 @@ periodics:
       - --test-cmd-args=--testconfig=testing/load/config.yaml
       - --test-cmd-args=--testconfig=testing/access-tokens/config.yaml
       - --test-cmd-args=--testoverrides=./testing/load/kubemark/500_nodes/override.yaml
+      - --test-cmd-args=--testoverrides=./testing/prometheus/not-scrape-kube-proxy.yaml
       - --test-cmd-name=ClusterLoaderV2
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes/sig-release/release-branch-jobs/1.15.yaml
```diff
@@ -250,6 +250,7 @@ periodics:
       - --test-cmd-args=--testconfig=testing/load/config.yaml
       - --test-cmd-args=--testconfig=testing/access-tokens/config.yaml
       - --test-cmd-args=--testoverrides=./testing/load/kubemark/500_nodes/override.yaml
+      - --test-cmd-args=--testoverrides=./testing/prometheus/not-scrape-kube-proxy.yaml
```
wojtek-t, line 253: It seems we don't have in 1.16 - is it possible to fix it consistently with other branches?

mm4tt: NPL is disabled in kubemark from 1.16+ - https://github.com/kubernetes/perf-tests/blob/6d4554d34c84a05070c964a4dd56ba97330b9d8d/clusterloader2/pkg/measurement/common/slos/network_programming.go#L51
It got accidentally enabled in 1.15 and this is the easiest way to disable it, once 1.15 goes away we'll have a consistent settings

wojtek-t: fair enough

PR #17595, opened 2020-05-14
fix kind periodic jobs migration

aojea:
kind periodic jobs started to fail after the migration to podutils.
The problem is that I copy-pasted blindly the job config from the PR jobs, and it was not cloning the kind repo

### DIFF
--- a/config/jobs/kubernetes-sigs/kind/kind.yaml
+++ b/config/jobs/kubernetes-sigs/kind/kind.yaml
@@ -40,7 +40,10 @@ periodics:
     repo: kubernetes
     base_ref: master
     path_alias: k8s.io/kubernetes
-  path_alias: sigs.k8s.io/kind
+  - org: kubernetes-sigs
+    repo: kind
+    base_ref: master
+    path_alias: sigs.k8s.io/kind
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes-sigs/kind/kind.yaml
```diff
@@ -40,7 +40,10 @@ periodics:
     repo: kubernetes
     base_ref: master
     path_alias: k8s.io/kubernetes
-  path_alias: sigs.k8s.io/kind
+  - org: kubernetes-sigs
+    repo: kind
+    base_ref: master
+    path_alias: sigs.k8s.io/kind
```
BenTheElder, line 46: The CI jobs should be using the latest binaries and e2e-k8s.sh like kubernetes presubmits. More clones and building is slower, flakier, etc.

aojea: ok, let me send a new PR

aojea: https://github.com/kubernetes/test-infra/pull/17596

PR #17596, opened 2020-05-14
use latest kind binaries for ci jobs

aojea:
> The CI jobs should be using the latest binaries and e2e-k8s.sh like kubernetes presubmits. More clones and building is slower, flakier, etc.

Also, remove the timeout decorations, it's a leftover from the presubmit jobs that are causing these jobs to fail


aojea:
/hold
the jobs are timing out, I'll fix it with a commit in this PR

aojea:
:crossed_fingers: this is the last PR in this migration to pod utils odyssey :sweat_smile: 
/hold cancel

PR #17597, opened 2020-05-14
Process latest report for each label separately.

fejta:
The /retest command handling in gerrit relies on parsing the messages it leaves.
When it was written prow would only ever write to one label. Now users can and do
have different sets of tests (blocking vs informing) report on different labels.

This PR changes the adapter to look for the latest report on each label,
rather than the single latest report on each label. Thus /retest should retest
jobs more correctly now.

Also involves refactoring the reporter to make Generating/Parsing reports reversible
(currently the parsed report differs from the string output of a generated report).

Added a variety of unit tests, which also included some refactoring towards this end.

### DIFF
--- a/prow/crier/reporters/gerrit/reporter.go
+++ b/prow/crier/reporters/gerrit/reporter.go
@@ -71,16 +71,19 @@ type Client struct {
 
 // Job is the view of a prowjob scoped for a report
 type Job struct {
-	Name, State, icon, url string
+	Name  string
+	State v1.ProwJobState
+	Icon  string
+	URL   string
 }
 
 // JobReport is the structured job report format
 type JobReport struct {
-	Jobs    []*Job
-	success int
-	total   int
-	message string
-	header  string
+	Jobs    []Job
+	Success int
+	Total   int
+	Message string
+	Header  string
 }
 
 // NewReporter returns a reporter client
@@ -197,8 +200,8 @@ func (c *Client) Report(pj *v1.ProwJob) ([]*v1.ProwJob, error) {
 			toReportJobs = append(toReportJobs, pjOnRevisionWithSameLabel)
 		}
 	}
-	report := generateReport(toReportJobs)
-	message := report.header + report.message
+	report := GenerateReport(toReportJobs)
+	message := report.Header + report.Message
 	// report back
 	gerritID := pj.ObjectMeta.Annotations[clientGerritID]
 	gerritInstance := pj.ObjectMeta.Annotations[clientGerritInstance]
@@ -210,7 +213,7 @@ func (c *Client) Report(pj *v1.ProwJob) ([]*v1.ProwJob, error) {
 		reportLabel = codeReview
 	}
 
-	if report.total <= 0 {
+	if report.Total <= 0 {
 		// Shouldn't happen but return if does
 		logger.Warn("Tried to report empty or aborted jobs.")
 		return nil, nil
@@ -221,7 +224,7 @@ func (c *Client) Report(pj *v1.ProwJob) ([]*v1.ProwJob, error) {
 		// Can only vote below zero before merge
 		// TODO(fejta): cannot vote below previous vote after merge
 		switch {
-		case report.success == report.total:
+		case report.Success == report.Total:
 			vote = lgtm
 		case pj.Spec.Type == v1.PresubmitJob:
 			vote = lbtm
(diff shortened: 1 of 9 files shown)


prow/gerrit/adapter/adapter.go
```diff
@@ -168,6 +168,28 @@ func createRefs(reviewHost string, change client.ChangeInfo, cloneURI *url.URL,
...
 
+func parseReports(account int, messages ...gerrit.ChangeMessageInfo) map[string]reporter.JobReport {
+	latestReports := map[string]reporter.JobReport{}
+	latestTimes := map[string]time.Time{}
+	for _, message := range messages {
+		// If message status report is not from the prow account ignore
+		if message.Author.AccountID != account {
+			continue
+		}
+		tag := message.Tag
```
amwat, line 179: I don't think we use tags, do we? should this be label instead? 
Although i don't think we get label from the API https://gerrit-review.googlesource.com/Documentation/rest-api-changes.html#change-message-info
we might need to parse that from the report too.

amwat: Also keying on the label might make things complicated if we have multiple jobs that use an empty label, since they aren't reported together in this case.
One alternative might be to key by job, gather the latest reports for each job and then see if each job was failing in their corresponding latest reports (can consider making jobs a set instead of a list in the jobreport struct, if we go down this route).

fejta: The tag is the label. 

If the label is empty we don't report.

fejta: hmm, let me investigate that...

fejta: (Neither of those things seem as immediately true as when I started down this path). Will update this when it's ready for a review

fejta:
/unassign
/assign @amwat 
@chaodaiG if you comment I can then assign you and it will show up in your PR dashboard.

prow/gerrit/adapter/adapter.go
fejta: Done

(updated to find the latest result for each job, thanks for catching this hilariously embarrassing oversight!)

chaodaiG:
ack

prow/gerrit/adapter/adapter.go
```diff
@@ -168,6 +169,43 @@ func createRefs(reviewHost string, change client.ChangeInfo, cloneURI *url.URL,
...
+// Scans messages for prow reports, which lists jobs and whether they passed.
+// Job is included in the set if the latest report has it failing.
+func failedJobs(account int, revision int, messages ...gerrit.ChangeMessageInfo) sets.String {
+	failures := sets.String{}
+	times := map[string]time.Time{}
+	for _, message := range messages {
+		if message.Author.AccountID != account { // Ignore reports from other accounts
+			continue
+		}
+		if message.RevisionNumber != revision { // Ignore reports for old commits
```
amwat, line 184: I'm not quite sure of the UX for this case, technically it's possible to have a use case where you run an optional job `/test foo` - > (it fails)  -> you send a couple of new patchsets/revisions and then want to run it again. 
Do we care if `/retest` should honor it or should we expect another `/test foo` on the newer revision?
How do we handle this in GitHub?

prow/gerrit/adapter/adapter.go
```diff
@@ -168,6 +169,43 @@ func createRefs(reviewHost string, change client.ChangeInfo, cloneURI *url.URL,
...
+			continue
+		}
+		// TODO(fejta): parse triggered job reports and remove from failure set.
+		// (alternatively refactor this whole process rely less on fragile string parsing)
```
amwat, line 188: +1 gerrit not having status contexts is annoying

prow/crier/reporters/gerrit/reporter.go
```diff
@@ -309,19 +316,27 @@ func ParseReport(message string) *JobReport {
...
-			report.Jobs = append(report.Jobs, j)
+	var report JobReport
+	report.Header = contents[start] + "\n"
+	for i := start + 1; i < len(contents); i++ {
+		if contents[i] == "" {
+			continue
 		}
+		var j Job
+		if err := deserialize(contents[i], &j); err != nil {
+			logrus.Warnf("Could not deserialize %s to a job: %v", contents[i], err)
```
amwat, line 327: add a `continue` ?
i.e. if deserialize failed we shouldn't add it to the total, success, jobs list.

prow/crier/reporters/gerrit/reporter.go
```diff
@@ -309,19 +316,27 @@ func ParseReport(message string) *JobReport {
...
+		}
+		report.Total++
+		if j.State == v1.SuccessState {
+			report.Success++
+		}
+		report.Jobs = append(report.Jobs, j)
 	}
-	report.message = strings.TrimPrefix(message, report.header)
-	return report
+	report.Message = strings.TrimPrefix(message, report.Header+"\n")
```
amwat, line 335: `report.Header` already has the extra `\n` on line 320, so this one is extraneous? (also might fail to trim properly) 
or am I missing something?

prow/gerrit/adapter/adapter.go
```diff
@@ -168,6 +169,43 @@ func createRefs(reviewHost string, change client.ChangeInfo, cloneURI *url.URL,
...
+		report := reporter.ParseReport(message.Message)
+		if report == nil {
+			continue
+		}
+		for _, job := range report.Jobs {
+			name := job.Name
+			if latest, present := times[name]; present && message.Date.Before(latest) {
+				continue
+			}
+			times[name] = message.Date.Time
```
chaodaiG, line 198: Update times[name] for every message could cause out of order problem, unless messages are already sorted by datetime. 
For example, a job passed then failed, but the messages arrives in reversed order, then this test is not considered failure

prow/gerrit/adapter/trigger.go
```diff
@@ -17,55 +17,57 @@ limitations under the License.
...
-func messageFilter(lastUpdate time.Time, change client.ChangeInfo, presubmits []config.Presubmit, latestReport *reporter.JobReport, logger *logrus.Entry) (pjutil.Filter, error) {
-	var filters []pjutil.Filter
-	currentRevision := change.Revisions[change.CurrentRevision].Number
+// presubmitContexts returns the set of failing and all job names contained in the reports.
+func presubmitContexts(failed sets.String, presubmits []config.Presubmit, logger *logrus.Entry) (sets.String, sets.String) {
+	allContexts := sets.String{}
+	for _, presubmit := range presubmits {
+		allContexts.Insert(presubmit.Name) // TODO(fejta): context, not name
+	}
+	failedContexts := allContexts.Intersection(failed)
```
chaodaiG, line 36: Just curious, is this for scenarios where a job was deleted after it failed? If yes, how can this be guarded from here to the point where it's re-kicked off?

prow/gerrit/adapter/adapter.go
fejta: That scenario is what this code guards against.

If we have already seen this job in a report from a later message, then we will skip this job for the current report. 

`TestFailedJobs` unit tests this scenario (see the redshifted scenario)

prow/gerrit/adapter/trigger.go
fejta: That is what this code does? The failedContexts set is the intersection of the set of jobs that still exist (allContexts set) and most recently reported as failed (failed set)

https://en.wikipedia.org/wiki/Intersection_(set_theory)

The point is to not rekick off deleted jobs. Retest will only retest failedContexts. By removing jobs that do not exist from the failedContexts set, we will not try retest them.

prow/gerrit/adapter/adapter.go
chaodaiG: I see, didn't notice the `continue` above

Not sure I get the idea of redshifted analog though :)

prow/gerrit/adapter/trigger.go
chaodaiG: ack

prow/gerrit/adapter/trigger.go
```diff
@@ -20,52 +20,76 @@ import (
...
+			)
+			if state := strings.ToLower(job.State); state != failure && state != errored {
+				continue
 			}
-			for _, presubmit := range presubmits {
-				j, ok := jobs[presubmit.Name]
-				if ok && (strings.ToLower(j.State) == string(v1.FailureState) || strings.ToLower(j.State) == string(v1.ErrorState)) {
-					failedContexts.Insert(presubmit.Name)
-				}
+			if failed.Has(name) && !duplicates.Has(name) {
```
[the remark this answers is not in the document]
fejta: That seems excessively noisy? Also this code path is no longer exists :)

prow/gerrit/adapter/adapter.go
fejta: Gravity slows down time. This is one (improbable) reason why a message with an earlier timestamp arrives after a message with a later timestamp

https://www.sciencealert.com/earth-s-core-is-2-5-years-younger-than-its-crust-thanks-to-the-curvature-of-space-time

https://www.youtube.com/watch?v=EXHebyV2TNM

PR #17598, opened 2020-05-14
Run [Feature:Windows] Tests in GCE

runewake2:
The ginkgo `skip` filter was previously skipping all tests flagged with a [Feature:*], this was causing tests with [Feature:Windows] to be excluded. This update enables those tests to be run while continuing to exclude other tests during Windows based tests.

This should run 4 additional tests right now:

* gMSA:
  * `[sig-windows] [Feature:Windows] GMSA Kubelet [Slow] kubelet GMSA support when creating a pod with correct GMSA credential specs passes the credential specs down to the Pod's containers`
* RunAsUsername:
  * `[sig-windows] [Feature:Windows] SecurityContext RunAsUserName should not be able to create pods with unknown usernames`
  * `[sig-windows] [Feature:Windows] SecurityContext RunAsUserName should override SecurityContext username if set`
  * `[sig-windows] [Feature:Windows] SecurityContext RunAsUserName should be able create pods and run containers with a given username`

Another gMSA e2e test is specifically excluded because it requires additional setup.

### DIFF
--- a/config/jobs/kubernetes/sig-windows/windows-gce.yaml
+++ b/config/jobs/kubernetes/sig-windows/windows-gce.yaml
@@ -79,7 +79,7 @@ periodics:
       - --gcp-nodes=2
       - --test=false
       - --test-cmd=$GOPATH/src/k8s.io/windows-testing/gce/run-e2e.sh
-      - --test-cmd-args=--ginkgo.focus=\[Conformance\]|\[NodeConformance\]|\[sig-windows\] --ginkgo.skip=\[LinuxOnly\]|\[Serial\]|\[Feature:.+\] --minStartupPods=8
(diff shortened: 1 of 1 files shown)


runewake2:
I signed it

pjh:
There are two `ci-kubernetes-e2e-windows-containerd` jobs that are unchanged - should the filters for those jobs be updated too? It would be preferable to test against containerd sooner rather than later, but maybe check if AKS + containerd is running these tests too.

pjh:
/lgtm

Thanks for writing a detailed PR description!

After this merges we should probably make the same change for our gce-windows jobs in the [release-branch-jobs](https://github.com/pjh/test-infra/tree/master/config/jobs/kubernetes/sig-release/release-branch-jobs) configs.

PR #17599, opened 2020-05-14
Fix typo in pull-kind-verify

amwat:


amwat:
/close

Fixed in https://github.com/kubernetes/test-infra/pull/17604

PR #17601, opened 2020-05-14
GCSConfiguration merging: Avoid assigning to nil map

alvaroaleman:


prow/config/BUILD.bazel
```diff
@@ -33,6 +33,10 @@ go_test(
         "//prow/pod-utils/decorate:go_default_library",
         "//prow/pod-utils/downwardapi:go_default_library",
         "@com_github_google_go_cmp//cmp:go_default_library",
+<<<<<<< HEAD
```
stevekuznetsov, line 36: yikes

PR #17608, opened 2020-05-15
Deal better with docker server/client mismatch

dims:
`pull-kubernetes-local-e2e` ends up failing because:

```
W0515 00:39:02.297] 2020/05/15 00:39:02 local.go:49: Failed to list containers: Error response from daemon: client version 1.41 is too new. Maximum supported API version is 1.40
```

So add version negotiation for better outcomes


liggitt:
picked up by https://github.com/kubernetes/test-infra/pull/17600

PR #17615, opened 2020-05-16
Cherrypicker: add support for conditional labels

clarketm:
* Add `--conditional-labels` option to the **cherrypicker** plugin to specify labels to conditionally apply to the cherrypicked PR when another label is present.
* Add new `StringToStringSlice` flag for `map[string][]string`
* Add new `StringSlice` flag for `[]string`

https://github.com/istio/istio.io/issues/6333

### DIFF
--- a/prow/flagutil/string_slice.go
+++ b/prow/flagutil/string_slice.go
@@ -1,0 +1,50 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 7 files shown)


prow/flagutil/string_slice.go
```diff
@@ -0,0 +1,50 @@
...
+
+package flagutil
+
+import (
+	"strings"
+)
+
+// StringSlice is a generic command-line flag for a []string.
+// The flag is specified using the format: value1,value2,value3
+type StringSlice struct {
```
cjwagner, line 25: We already have a `Strings` type for string slices: https://github.com/kubernetes/test-infra/blob/master/prow/flagutil/strings.go
Including this new type for comma delimited lists (instead of repeated flags) seems fine, but a name that distinguishes this type from `Strings` would be preferable since they both resolve to string slices.

prow/flagutil/string_slice_test.go
```diff
@@ -0,0 +1,63 @@
...
+		},
+		{
+			value:    "key",
+			expected: NewStringSlice([]string{"key"}),
+		},
+	}
+
+	for _, tc := range testCases {
+		actual := NewStringSlice([]string{})
+		err := actual.Set(tc.value)
```
cjwagner, line 48: Please also test that the default works if no value is specified.

PR #17616, opened 2020-05-16
Prow: Add support for yaml anchors

alvaroaleman:
This PR adds support for yaml anchors to prow. While yaml anchors are a debatable feature, it may be the lesser evil. For example [openshifts tide config](https://github.com/openshift/release/blob/245472992865a9ba0b4fca1f75b63630943f8026/core-services/prow/02_config/_config.yaml#L625-L2220) is about 1.5k lines long, most of which is an ever-repeating list of repos that all have a set of branches with a specific config.

The implementation here is basically to copy https://github.com/ghodss/yaml and use it in conjunction with `gopkg.in/yaml.v3` instead of `gopkg.in/yaml.v2`. My first approach was to use https://github.com/sevlyar/retag to create the yaml tags based on the json tags at runtime. While that works, it generates a struct that has a name with more than 32.7k chars for the prow/config package, which makes [reflect panic](https://groups.google.com/forum/#!msg/golang-nuts/0kcaYQ2_yXs/6no0f1-EAAAJ).

Another implementation option would be to simply add yaml tags everywhere where we currently have json tags.

Note that the yaml generated when marshallng has a slightly different spacing from what we had before, but is semantically identical. If we care about this, we could just continue to use the sigs.k8s.io/yaml package for marshaling.


stevekuznetsov:
Do you have a simple tool to use to migrate the config, if the `true --> "true"` changes are required by users?

alvaroaleman:
> Do you have a simple tool to use to migrate the config, if the true --> "true" changes are required by users?

No, this isn't simple to migrate because the migration tool would need to know the schema.

`checkconfig` complains about this though, so maybe we can add an Announcement for that? I am somewhat surprised this ever worked tbh, `kubectl` for example doesn't allow this either

fejta:
I do not approve of supporting yaml anchors to prow. Please generate configs using whatever method you want.

alvaroaleman:
That is a fair opinion to have. You are free to not use this. I can also
add a checkconfig check that complains about yaml anchors.

However, it is an opinion and it would be inappropriate to assume you know
what is best for every possible Prow deployment.

fejta:
This kind of thing isn't the kind of thing to just drop into a PR.

We've tried yaml anchors before and it was a maintainability disaster: the only person willing to a file is the person who wrote it.

Kubernetes objects being excessively verbose is not a unique problem to prow.

alvaroaleman:
>This kind of thing isn't the kind of thing to just drop into a PR.

Where would you like to see it?

> We've tried yaml anchors before and it was a maintainability disaster: the only person willing to a file is the person who wrote it.

Again, you are free to not use this feature and disable it for the deployments you are responsible for.

> Kubernetes objects being excessively verbose is not a unique problem to prow.

The reason for wanting this is not Kubernetes objects being verbose. Its Prows config being verbose.

fejta:
I'd like to see a proposal that we can discuss. This is unnecessary complexity that should be disabled by default and ideally kept out of prow's codebase.

alvaroaleman:
Yaml anchors already work, there is no need for this change.

PR #17618, opened 2020-05-17
prow: crier: add s3 support

sbueringer:
This PR is relatively complex because gocloud doesn't support preconditions across all storage providers

EDIT: Sorry thought his PR is independent from: [prow: deck: add s3 support #17183](https://github.com/kubernetes/test-infra/pull/17183). But the `ParsePath` func is added via the deck pr. But should be only a matter of days :)
EDIT 2: now it's merged

See: https://github.com/kubernetes/test-infra/issues/11260

### DIFF
--- a/prow/cmd/crier/main.go
+++ b/prow/cmd/crier/main.go
@@ -23,9 +23,7 @@ import (
 	"os"
 	"time"
 
-	"cloud.google.com/go/storage"
 	"github.com/sirupsen/logrus"
-	"google.golang.org/api/option"
 
 	prowapi "k8s.io/test-infra/prow/apis/prowjobs/v1"
 	prowjobinformer "k8s.io/test-infra/prow/client/informers/externalversions"
@@ -272,25 +270,13 @@ func main() {
 	}
 
 	if o.gcsWorkers > 0 || o.k8sGCSWorkers > 0 {
-		ctx := context.Background()
-		// FIXME(sbueringer): try to get rid of storage.Client in this PR
-		var s *storage.Client
-		var opts []option.ClientOption
-		if o.storage.GCSCredentialsFile != "" {
-			opts = append(opts, option.WithCredentialsFile(o.storage.GCSCredentialsFile))
-		}
-		s, err = storage.NewClient(ctx, opts...)
-		if err != nil {
-			// FIXME: handle the same as in deck, don't just ignore the error
-			logrus.WithError(err).Error("Error creating storage client for gcs workers.")
-		}
-		opener, err := io.NewOpener(ctx, o.storage.GCSCredentialsFile, o.storage.S3CredentialsFile)
+		opener, err := io.NewOpener(context.Background(), o.storage.GCSCredentialsFile, o.storage.S3CredentialsFile)
 		if err != nil {
 			logrus.WithError(err).Fatal("Error creating opener")
 		}
 
 		if o.gcsWorkers > 0 {
-			gcsReporter := gcsreporter.New(cfg, s, opener, o.dryrun)
+			gcsReporter := gcsreporter.New(cfg, opener, o.dryrun)
 			controllers = append(
 				controllers,
 				crier.NewController(
@@ -307,7 +293,7 @@ func main() {
 				logrus.WithError(err).Fatal("Error building pod client sets for Kubernetes GCS workers")
 			}
 
-			k8sGcsReporter := k8sgcsreporter.New(cfg, s, opener, coreClients, float32(o.k8sReportFraction), o.dryrun)
+			k8sGcsReporter := k8sgcsreporter.New(cfg, opener, coreClients, float32(o.k8sReportFraction), o.dryrun)
 			controllers = append(
 				controllers,
 				crier.NewController(
--- a/prow/cmd/crier/main.go
+++ b/prow/cmd/crier/main.go
@@ -122,18 +122,22 @@ func (o *options) validate() error {
 	}
 
 	if o.gcsWorkers > 0 {
-		// use gcsWorkers if blobStorageWorkers is not set
-		if o.blobStorageWorkers == 0 {
-			o.blobStorageWorkers = o.gcsWorkers
-		}
 		logrus.Warn("--gcs-workers is deprecated and will be removed in August 2020. Use --blob-storage-workers instead.")
+		// return an error when the old and new flags are both set
+		if o.blobStorageWorkers != 0 {
+			return errors.New("only one of --gcs-workers or --blog-storage-workers can be set at the same time")
+		}
+		// use gcsWorkers if blobStorageWorkers is not set
+		o.blobStorageWorkers = o.gcsWorkers
 	}
 	if o.k8sGCSWorkers > 0 {
-		// use k8sGCSWorkers if k8sBlobStorageWorkers is not set
-		if o.k8sBlobStorageWorkers == 0 {
-			o.k8sBlobStorageWorkers = o.k8sGCSWorkers
-		}
 		logrus.Warn("--kubernetes-gcs-workers is deprecated and will be removed in August 2020. Use --kubernetes-blob-storage-workers instead.")
+		// return an error when the old and new flags are both set
+		if o.k8sBlobStorageWorkers != 0 {
+			return errors.New("only one of --kubernetes-gcs-workers or --kubernetes-blog-storage-workers can be set at the same time")
+		}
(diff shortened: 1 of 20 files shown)


prow/cmd/crier/main.go
```diff
@@ -271,19 +270,13 @@ func main() {
 	}
 
 	if o.gcsWorkers > 0 || o.k8sGCSWorkers > 0 {
```
alvaroaleman, line 272: We will need to deprecate and rename this flag as well

prow/crier/reporters/gcs/internal/util/util.go
```diff
@@ -23,44 +23,49 @@ import (
 	"io"
 	"net/http"
 
-	"cloud.google.com/go/storage"
 	"github.com/sirupsen/logrus"
 	"google.golang.org/api/googleapi"
-	"k8s.io/test-infra/prow/apis/prowjobs/v1"
+	utilpointer "k8s.io/utils/pointer"
+
+	v1 "k8s.io/test-infra/prow/apis/prowjobs/v1"
```
alvaroaleman, line 30: please rename the import to prowv1

prow/io/opener.go
```diff
@@ -240,13 +240,25 @@ func (o *opener) RangeReader(ctx context.Context, path string, offset, length in
 	return reader, nil
 }
 
+var PreconditionFailedObjectAlreadyExists = fmt.Errorf("object already exists")
+
 // Writer returns a writer that overwrites the path.
 func (o *opener) Writer(ctx context.Context, p string, opts ...WriterOptions) (io.WriteCloser, error) {
+	var preconditionDoesNotExist bool
```
alvaroaleman, line 247: I think we should do this proper by:
* changing the receiver type in the `Writer` method from `WriterOptions` to a new public inteface `type WriterOption interface { Apply(wo *WriterOptions) }`
* Unexport the existing `Apply` method on the `WriterOptions`, there is no reason for it to be exported
* Make `WriterOptions` implement the `WriterOption` interface
* start this function off by doing an `options := &WriterOptions{}; for _, opt := range opts { opt.Apply(options) }`
* Later on just do an `if options.PreconditionDoesNotExist != nil && options.PreconditionDoesNotExist`

This allows us to end up with exactly one canonical `WriterOptions` struct and thus avoids having to loop three times (1. Options that can not be directly set on the corresponding gcs/blob structs, 2. for gcs, 3. for blob)

prow/io/opener.go
```diff
@@ -270,6 +282,22 @@ func (o *opener) Writer(ctx context.Context, p string, opts ...WriterOptions) (i
 	for _, opt := range opts {
 		opt.Apply(nil, &wOpts)
 	}
+
+	if preconditionDoesNotExist {
+		wOpts.BeforeWrite = func(asFunc func(interface{}) bool) error {
+			_, err := o.Reader(ctx, p)
+			// we got an error, but not file not exists
+			if err != nil && !IsNotExist(err) {
```
alvaroaleman, line 290: Can be simplified:
```
if err != nil {
  // we got an error, but not file not exists
  if !IsNotExist(err) { return err }
  // Precondition fulfilled, return nil
  return nil
}
return PreconditionFailedObjectAlreadyExists
```

prow/cmd/crier/main.go
sbueringer: So I guess that means adding new flags and deprecating:
```go
fs.IntVar(&o.gcsWorkers, "gcs-workers", 0, "Number of GCS report workers (0 means disabled)")
	fs.IntVar(&o.k8sGCSWorkers, "kubernetes-gcs-workers", 0, "Number of Kubernetes-specific GCS report workers (0 means disabled)")
```

Name of the new flags:
* I would suggest `blob-storage-workers` and `kubernetes-blob-storage-workers`
* alternative: `storage-workers` and `kubernetes-storage-workers` 

Deprecation:
* I guess I should just add a deprecation notice to both flags?
* What would be an appropriate time frame: 2-3 months?

prow/crier/reporters/gcs/internal/util/util.go
sbueringer: Done

prow/io/opener.go
sbueringer: Interesting idea. Done

prow/io/opener.go
sbueringer: Done

sbueringer:
> Some mostly stylistic nits, the change in itself looks good. I'd like to hold off with it until the ~S3~ Deck PR was deployed for a couple of days though, because this PR depends on it and merging it would make reverting more complicated

Yup agree with waiting to merge. Findings should all be addressed

prow/cmd/crier/main.go
alvaroaleman: `[kubernetes-]blob-storage-workers` SGTM. Yeah, approach would be to make the new flags default to the old flags value but emit a warning if the old flags are used and add an a line about that in `prow/Announcements.md` that they will be removed in three months

sbueringer: Thx. Done

prow/io/option.go
```diff
@@ -23,17 +23,40 @@ import (
 	"google.golang.org/api/googleapi"
 )
 
+type WriterOption interface {
+	Apply(wo *WriterOptions)
+}
+
 // WriterOptions are options for the opener Writer method
 type WriterOptions struct {
```
fejta, line 31: Is there a reason we can't just get rid of this class?

prow/cmd/crier/main.go
```diff
@@ -120,6 +121,21 @@ func (o *options) validate() error {
 		}
 	}
 
+	if o.gcsWorkers > 0 {
+		// use gcsWorkers if blobStorageWorkers is not set
+		if o.blobStorageWorkers == 0 {
+			o.blobStorageWorkers = o.gcsWorkers
```
fejta, line 127: Should this be `+=`? Do we support having gcsWorkers and k8sGCSWorkers both > 0 right now?

sbueringer: `+=` is fine for. I would have guessed that either the new or old flags are set and not both.

crier_deployment.yaml suggests it works:
```
        - --gcs-workers=1
        - --kubernetes-gcs-workers=1
```

prow/io/option.go
sbueringer: How would you replace it?

fejta: I'm fine with what is happening, but I don't see the point of the WriterOption interface.

sbueringer: Ah got it. Thought it's about the struct. The original discussion is here: https://github.com/kubernetes/test-infra/pull/17618#discussion_r426279391

Before the current implementation I had 3 for loops to "aggregate" the WriterOptions so that was the reason to change it. I'm fine with both variants.

fejta: Getting rid of the loops is great, but I don't see why that necessitates a pointless interface that has the single implementor of this interface as one of its arguments.

sbueringer: Okay so possible variants:
* current implementation
* current implementation + remove the interface and keep the Apply method

Are there other ones?

@alvaroaleman WDYT?

alvaroaleman: Yeah, I am fine with removing the interface

prow/cmd/crier/main.go
alvaroaleman: I think we should error if both are set or not? That's not a valid configuration and the user apparently did not understand that one flag replaces the other.

Also, `+= is fine for me. ` no its not. Then we silently enable `k8sgcsworkers` just because someone enabled `gcsworker`.

(As a pure anectdotial sidecoment, I find the format of setting a worker number > 0 in order to enable a controller a super bad UX. We should just provide an option to enable the given controller, set a sane default for workers and allow to optionally override that number. But that has nothing to do with this PR.)

fejta: Agree that this flag should be a boolean

Also I'm fine with erroring if both are set as that is a bad scenario (but we don't currently check for it)

sbueringer: Yup let's keep the current version and error when both old and new flags are set

prow/io/option.go
sbueringer: Done

prow/cmd/crier/main.go
sbueringer: Done

fejta:
/hold
(unhold as you please)

PR #17621, opened 2020-05-18
Enable CL2_ALLOWED_SLOW_API_CALLS everywhere

mm4tt:
Ref. https://github.com/kubernetes/perf-tests/issues/986


mm4tt:
The feature has been enabled on Friday for 100 node jobs. I took a look and they look fine, we're good to enable it everywhere

/assign @wojtekt-t
FYI, @jprzychodzen

PR #17623, opened 2020-05-18
IPVS: don't filter dashboard results

aojea:
The pr-ipvs job was failing for a long time and we didn't notice until today.

The testgrid information is misleading because is filtering the tests and is not showing all the failures

Also, skip sig-storage jobs

xref https://github.com/kubernetes/kubernetes/pull/89998#issuecomment-630229189

### DIFF
--- a/config/testgrids/kubernetes/sig-network/config.yaml
+++ b/config/testgrids/kubernetes/sig-network/config.yaml
@@ -39,13 +39,11 @@ dashboards:
         alert_mail_to_addresses: kubernetes-sig-network-test-failures@googlegroups.com
     - name: gci-gce-ipvs
       test_group_name: ci-kubernetes-e2e-gci-gce-ipvs
-      base_options: include-filter-by-regex=\[sig-network\]
       description: network gci-gce e2e tests in ipvs proxier mode
       alert_options:
         alert_mail_to_addresses: kubernetes-sig-network-test-failures@googlegroups.com, lionwei1992+alerts@gmail.com
     - name: pull-gci-gce-ipvs
       test_group_name: pull-kubernetes-e2e-gci-gce-ipvs
-      base_options: include-filter-by-regex=\[sig-network\]
       description: presubmit network gci-gce e2e tests in ipvs proxier mode
       alert_options:
         alert_mail_to_addresses: kubernetes-sig-network-test-failures@googlegroups.com


aojea:
/assign @mrhohn

:smile:

config/jobs/kubernetes/sig-network/sig-network-misc.yaml
```diff
@@ -33,7 +33,7 @@ presubmits:
         - --ginkgo-parallel=30
         - --provider=gce
         - --stage=gs://kubernetes-release-pull/ci/pull-kubernetes-e2e-gci-gce-ipvs
-        - --test_args=--ginkgo.skip=\[Slow\]|\[Serial\]|\[Disruptive\]|\[Flaky\]|\[Feature:.+\] --minStartupPods=8
+        - --test_args=--ginkgo.skip=\[Slow\]|\[Serial\]|\[Disruptive\]|\[Flaky\]|\[Feature:.+\]|\[sig-storage\] --minStartupPods=8
```
andrewsykim, line 36: We should keep this if we don't want to run storage tests cause `ginkgo.skip`?

andrewsykim: nvm ignore me, haven't had my morning coffee yet :)

andrewsykim:
@aojea I think we should hold off on skipping storage tests for now, they may be failing due to edge cases in the service data path we haven't considered.

aojea:
> @aojea I think we should hold off on skipping storage tests, for now, they may be failing due to edge cases in the service data path we haven't considered.

these tests are failing for more than 2 weeks

```
[sig-storage] Flexvolumes should be mountable when non-attachable
[sig-storage] Flexvolumes should be mountable when attachable 
```

https://github.com/kubernetes/kubernetes/pull/89998#issuecomment-630186050

aojea:
@andrewsykim what do you think if we disable the storage tests only in presubmits?
they will keep running in the periodic jobs but won't give a false signal to the people submitting PR (that's actually why I opened this PR :) )

I modify the PR with this approach

config/jobs/kubernetes/sig-network/sig-network-misc.yaml
aojea: this is skipped only for presubmit job, the periodic job will keep running the storage tests

andrewsykim:
> these tests are failing for more than 2 weeks

Btw, I agree with the change to update the filter regex on testgrid. 

> what do you think if we disable the storage tests only in presubmits?
they will keep running in the periodic jobs but won't give a false signal to the people submitting PR (that's actually why I opened this PR :) )

I'd like to keep `pull-kubernetes-e2e-gci-gce-ipvs` running the same set of tests as `pull-kubernetes-e2e-gce` if possible. 

I think the false signal is okay as long as we keep the presubmit optional? It's more motivation for us to fix the flaky/failing test anyways :P

aojea:
> I think the false signal is okay as long as we keep the presubmit optional? It's more motivation for us to fix the flaky/failing test anyways :P

fair enough :smile:

PR #17626, opened 2020-05-18
Add a Plank V2 controller

alvaroaleman:
This PR adds a featuregated and off-by-default Plank V2 implementation that is a controller and not a cronjob. Advantages over the old implementation include:
* Eventbased, thus avoids the up-to-30-seconds wait after a prowjob was created/updated and after a pod finishes
* Does not process jobs that are finished and thus avoids useless work
* Acts on individual jobs rather than a list of jobs, allowing us to get rid of all locking and simplifying the code overall
* Uses caches for both prowjobs and pods, thus reducing the load on the apiserver
* Per-Item retrying, resulting in much quicker convergence on transient errors
* Exponential backoff for retrying, rather than "retry everything forever every 30 seconds"

It passes all existing plank tests. They get copied at test runtime and Plank V2 is made to reconcile all prowjobs in the test to have something that is comparable to old plank. The only exception here is `TestOrderedJobs` which verifies that Plank orders the job, because that is not applicable as plank processes jobs as soon as it gets an event.

Apart from the added tests I also manually deployed it to verify that it actually works.

Fixes https://github.com/kubernetes/test-infra/pull/14236

/assign @stevekuznetsov @fejta @cjwagner


alvaroaleman:
The commit ordering is a bit of a hassle for such a big change, because sometimes certain issues that are discovered later on logically belong into an earlier commit, but they might not be applicable to that earlier commit which creates useless work. If possible, I would prefer to add commits on the PR on an as-need basis after feedback and just have it squashed on merge.

prow/cmd/plank/main.go
```diff
@@ -141,17 +155,54 @@ func main() {
...
-		logrus.WithError(err).Fatal("failed to add controller to manager")
+		return fmt.Errorf("failed to add controller to manager: %w", err)
 	}
-	if err := mgr.Start(interrupts.Context().Done()); err != nil {
-		logrus.WithError(err).Fatal("failed to start manager")
+
+	return nil
+}
+
+func v2Main(o options, mgr manager.Manager, cfg config.Getter) error {
```
stevekuznetsov, line 187: Why should we prefer a `plank` binary with a `--v2` flag over deploying a controller manager and enabling this controller?

prow/flagutil/kubernetes_cluster_clients.go
```diff
@@ -206,6 +208,42 @@ func (o *KubernetesOptions) BuildClusterCoreV1Clients(dryRun bool) (v1Clients ma
 	return clients, nil
 }
 
+// BuildClusterManagers returns a manager per buildCluster.
+// Per default, LeaderElection and Metrics are disabled, as we assume
```
stevekuznetsov, line 212: What does `Metrics` give us? Would be nice to see e.g. queue depth per build cluster. Unless `Metrics` just needs to be on *one* manager in the process

prow/plank/reconciler.go
```diff
@@ -0,0 +1,645 @@
...
+	totURL             string
+	clock              clock.Clock
+}
+
+func (r *reconciler) syncMetrics(stop <-chan struct{}) error {
+	for {
+		select {
+		case <-stop:
+			return nil
+		case <-time.NewTicker(30 * time.Second).C:
```
stevekuznetsov, line 130: Isn't this the form that creates new tickers and leaks them?

prow/plank/reconciler.go
```diff
@@ -0,0 +1,645 @@
...
+	if err := r.pjClient.Get(r.ctx, request.NamespacedName, pj); err != nil {
+		if !kerrors.IsNotFound(err) {
+			return reconcile.Result{}, fmt.Errorf("failed to get prowjob %s: %v", request.Name, err)
+		}
+
+		// Objects can be deleted from the API while being in our workqueue
+		return reconcile.Result{}, nil
+	}
+
+	// TODO: Terminal errors for unfixable cases like missing build clusters
```
stevekuznetsov, line 159: As long as those get bubbled up as big fires instead of being swept under the rug by not being retried

prow/plank/reconciler.go
```diff
@@ -0,0 +1,645 @@
...
+	case prowv1.TriggeredState:
+		return r.syncTriggeredJob(pj)
+	case prowv1.AbortedState:
+		return nil, r.syncAbortedJob(pj)
+	}
+
+	return nil, nil
+}
+
+func (r *reconciler) terminateDupes(pj *prowv1.ProwJob) error {
```
stevekuznetsov, line 190: What does this do if the `pj` that we pass in is the dupe?

prow/plank/reconciler.go
```diff
@@ -0,0 +1,645 @@
...
+			},
+		}
+		if err := client.Delete(r.ctx, podToDelete); err != nil {
+			return fmt.Errorf("failed to delete pod %s/%s: %w", podToDelete.Namespace, podToDelete.Name, err)
+		}
+		return nil
+	})
+}
+
+func (r *reconciler) syncPendingJob(pj *prowv1.ProwJob) error {
```
stevekuznetsov, line 214: Might be good to define what `Pending` (etc) means in the GoDoc for these reconcile methods

prow/plank/reconciler.go
```diff
@@ -0,0 +1,645 @@
...
+			pj.Status.BuildID = id
+			pj.Status.PodName = pn
+			r.log.WithFields(pjutil.ProwJobFields(pj)).Info("Pod is missing, starting a new pod")
+		}
+	} else {
+
+		switch pod.Status.Phase {
+		case corev1.PodUnknown:
+			// Pod is in Unknown state. This can happen if there is a problem with
+			// the node. Delete the old pod, we'll start a new one next loop.
```
stevekuznetsov, line 244: nit: no more loops

prow/plank/reconciler.go
```diff
@@ -0,0 +1,645 @@
...
+			client, ok := r.buildClients[pj.ClusterAlias()]
+			if !ok {
+				return fmt.Errorf("unknown pod %s: unknown cluster alias %q", pod.Name, pj.ClusterAlias())
+			}
+
+			r.log.WithField("name", pj.ObjectMeta.Name).Debug("Delete Pod.")
+			return client.Delete(r.ctx, pod)
+
+		case corev1.PodSucceeded:
+			// Pod succeeded. Update ProwJob, talk to GitHub, and start next jobs.
```
stevekuznetsov, line 255: Nit: we don't have `run_after_success` so there are no "next jobs"

prow/plank/reconciler.go
```diff
@@ -0,0 +1,645 @@
...
+			}
+			// Pod failed. Update ProwJob, talk to GitHub.
+			pj.SetComplete()
+			pj.Status.State = prowv1.FailureState
+			pj.Status.Description = "Job failed."
+
+		case corev1.PodPending:
+			maxPodPending := r.config().Plank.PodPendingTimeout.Duration
+			maxPodUnscheduled := r.config().Plank.PodUnscheduledTimeout.Duration
+			if pod.Status.StartTime.IsZero() {
```
stevekuznetsov, line 287: So we found that initcontainers running did not cause the start time to increment ... either this or the pending timeout is bugged as a result

prow/plank/reconciler.go
```diff
@@ -0,0 +1,645 @@
...
+func (r *reconciler) syncTriggeredJob(pj *prowv1.ProwJob) (*reconcile.Result, error) {
+	prevPJ := pj.DeepCopy()
+
+	var id, pn string
+
+	pod, podExists, err := r.pod(pj)
+	if err != nil {
+		return nil, err
+	}
+	// We may end up in a state where the pod exists but the prowjob is not
```
stevekuznetsov, line 361: nit: this massive comment block is for the `else` ... consider flipping the logic

prow/plank/reconciler.go
```diff
@@ -0,0 +1,645 @@
...
+			pj.SetComplete()
+			pj.Status.Description = "Job cannot be processed."
+			logrus.WithField("job", pj.Spec.Job).WithError(err).Warning("Unprocessable pod.")
+		}
+	} else {
+		id = getPodBuildID(pod)
+		pn = pod.ObjectMeta.Name
+	}
+
+	if pj.Status.State == prowv1.TriggeredState {
```
stevekuznetsov, line 390: Isn't this a no-op?

prow/plank/reconciler.go
```diff
@@ -0,0 +1,645 @@
...
+
+	pjs := &prowv1.ProwJobList{}
+	if err := r.pjClient.List(r.ctx, pjs, optNotCompletedProwJobsNamed(pj.Spec.Job)); err != nil {
+		return false, fmt.Errorf("failed listing prowjobs: %w:", err)
+	}
+	r.log.Infof("got %d not completed with same name", len(pjs.Items))
+
+	var olderMatchingPJs int
+	for _, foundPJ := range pjs.Items {
+		// Ignore self here. Second half of the condition is needed for tests.
```
stevekuznetsov, line 517: Production code that has call-outs for tests?!

prow/cmd/plank/main.go
alvaroaleman: So your suggestion here is to make this the starting point of a new `prow-controller-manager` and not touch `cmd/plank` at all?

prow/flagutil/kubernetes_cluster_clients.go
alvaroaleman: Yeah, this is just about a metric serving endpoint, not about the metrics themselves. C-R has its own metrics package with a registry, but we make our metrics serve both the ones from the prometheus registry and from the c-r registry.

I'll clarify the comment.

prow/cmd/plank/main.go
stevekuznetsov: That seems like it would create the same amount of operational change when someone wanted to opt in, and would set us up for the future.

prow/plank/reconciler.go
alvaroaleman: Yeah they should absolutely be logged but there is no point in retrying them

prow/plank/reconciler.go
alvaroaleman: This calls `pjutil.TerminateOlderJobs` which terminates only jobs that are older than the reference one passed in, hence we are not at risk that of terminating ourselves

prow/plank/reconciler.go
alvaroaleman: I can extend it, I actually just copied this code for the most part

prow/plank/reconciler.go
alvaroaleman: yeah the comment is misleading, but the pod deletion will trigger us again

prow/plank/reconciler.go
alvaroaleman: This is copied from the existing code so either its broken in both implementations or working in both. Are you okay with me opening an issue for this so we can discuss and fix this seperatly?

prow/plank/reconciler.go
alvaroaleman: No, we start off the function with the job being `Triggered` but above this, we might change it to `ErrorState`

prow/plank/reconciler.go
alvaroaleman: Yeah its terrible and I am embarrassed by this. But not sure how to else test this in a unit test:
* Including the actual Reconciliation in the test makes it more complicated
* I don't want to have a package-level variable for the Reconiler
* I don't want to make passing in the Reconciler part of the public api for this package

stevekuznetsov: I'm having a little bit of a hard time wrapping my brain around what specifically you are trying to do with this in testing, mind linking to the diff? GitHub rendering on this PR is really haywire.

alvaroaleman: Wait, my comment was for a different place of the code, I didn't look at the diff. So the basic problem here is:
* In old plank we always reconcile all prowjobs at once
* This allows us to track the pending prowjobs in memory and invalidate it on the end of the reconciliation
* Doing this here would introduce a global lock which I didn't want
* So we look at the jobs in the cluster instead and list all of them (via an index, else it would be incredible expensive)
* This list we get includes ourselves, so we have to remove ourselves via the UID
* The fakeclient does not set a UID

But you are right, the simpler and better approach is to just set a UID on the prowjob under test, I've done that.

The corresponding test is `TestMaxConcurency`

prow/cmd/plank/main.go
alvaroaleman: Done

prow/plank/reconciler.go
alvaroaleman: https://github.com/kubernetes/test-infra/issues/17688

prow/plank/reconciler.go
alvaroaleman: Done

prow/plank/reconciler.go
alvaroaleman: Updated the godoc

prow/plank/reconciler.go
alvaroaleman: Let me do this in a follow-up, PR is big enough already

prow/plank/reconciler.go
alvaroaleman: Added godocs

prow/plank/reconciler.go
alvaroaleman: Updated it

prow/plank/reconciler.go
alvaroaleman: Updated it

alvaroaleman:
@stevekuznetsov updated, PTAL

Issue #17625, opened 2020-05-18
go.k8s.io/triage serving stale data, triage update job timing out at 2h


spiffxp:
**What happened**:
From: https://storage.googleapis.com/k8s-gubernator/triage/index.html (emphasis added)

>1416 clusters of 116102 failures (7395 in last day) out of 145044 builds from 4/30/2020, 5:00:11 PM to **5/14/2020, 11:27:04 PM.**

https://k8s-testgrid.appspot.com/sig-testing-misc#triage&width=5&graph-metrics=test-duration-minutes

The job started steadily increasing in duration from ~45m to 2h in the past 2 weeks, hitting its 2h threshold around 2020-05-13

**What you expected to happen**:
This job _used_ to take 20m to  complete.

**Please provide links to example occurrences, if any**:
First timeout: https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-test-infra-triage/1260410739387011072
Latest timeout: https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-test-infra-triage/1262367914212724738

**Anything else we need to know?**:
I  wonder if the top failure cluster that started occuring around 2020-05-04 has anything to do with this?

Raising the job's timeout could ensure that fresh data lands

/area triage
/sig testing
/help
I don't have time to troubleshoot this today

aojea:
analyzing one occurrence for reference, however, I'm not familiar at all with these jobs so disregard it if it does not make sense:

```
+ pypy3 summarize.py triage_builds.json triage_tests/shard_000000000000.json triage_tests/shard_000000000001.json --output failure_data.json --output_slices slices/failure_data_PREFIX.json
I0513 03:29:11.685] done loading failed tests
I0513 03:29:11.686] Clustering failures for 2805 unique tests...
I0513 03:29:11.693]    1/2805, 16608 failures, BeforeSuite

...

I0513 03:35:52.052] Finished locally clustering 2805 unique tests (187794 failures) in 6m40s
I0513 03:36:18.969] done clustering inside each test
I0513 03:36:18.970] Combining clustered failures for 2805 unique tests...
I0513 03:36:18.989]    1/2805, 37 clusters, BeforeSuite

...

I0513 05:23:15.995] 2805/2805, 1 clusters, [GuestPolicies] [Package update] [centos-8]
I0513 05:23:15.999] Finished clustering 2805 unique tests (187670 failures) into 4470 clusters in 106m57s

[...]

PR #17629, opened 2020-05-18
bump timeout and reduce frequency of ci-test-infra-triage

dims:
Temporary fix for 17625


spiffxp:

PR #17630, opened 2020-05-19
add description text for containerd testgrid jobs

MHBauer:
I think these run to ensure that our upstream dependencies still build, but I am not sure what would fail if one of these failed.

They are very quick, usually 3-5 minutes.


MHBauer:
/assign @random-liu 
is there more of a "why" we can add to having these run every 30 minutes?

SergeyKanzhelev:
/unassign  @Random-Liu
/assign @dashpole 

/lgtm

Descriptions wouldn't hurt. We can discuss usability of these tests at https://github.com/kubernetes/test-infra/issues/18570

MHBauer:
updated with more reasoning as I understand it.

dashpole:
/approve

Thanks @MHBauer

PR #17631, opened 2020-05-19
Kops - Update to latest images for supported distros

hakman:
Mostly maintenance work.
/cc @rifelpet

### DIFF
--- a/config/jobs/kubernetes/kops/kops-periodics-distros.yaml
+++ b/config/jobs/kubernetes/kops/kops-periodics-distros.yaml
@@ -23,7 +23,7 @@ periodics:
       - --env=KOPS_RUN_TOO_NEW_VERSION=1
       - --extract=release/stable
       - --ginkgo-parallel
-      - --kops-args=--image=379101102735/debian-stretch-hvm-x86_64-gp2-2020-02-10-73984
+      - --kops-image=379101102735/debian-stretch-hvm-x86_64-gp2-2020-02-10-73984
       - --kops-priority-path=/workspace/kubernetes/platforms/linux/amd64
       - --kops-ssh-user=admin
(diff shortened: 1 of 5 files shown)


config/jobs/kubernetes/kops/kops-periodics-distros.yaml
```diff
@@ -203,11 +203,11 @@ periodics:
       - --env=KOPS_RUN_TOO_NEW_VERSION=1
       - --extract=release/stable
       - --ginkgo-parallel
-      - --kops-args=--image=ami-02eac2c0129f6376b
+      - --kops-args=
+      - --kops-image=679593333241/CentOS Linux 7 x86_64 HVM EBS ENA 2002_01-b7ee8a69-ee97-4a49-9e68-afaee216db2e-ami-0042af67f8e4dcc20.4
```
rifelpet, line 207: hm any reason to use the `--kops-image` syntax here as opposed to `--kops-args=--image` everywhere else? its probably worth being consistent one way or another (i dont have a preference either way). and maybe remove the empty `--kops-args` above if we go with `--kops-image`.

hakman: @justinsb tried using the args for this image but was not very successful. In the end he added a new field for the image. I updated all configs to use `--kops-image` now.

rifelpet:
nice, i like that we can run the centos7 test anywhere now that we arent providing an AMI ID directly :)
/lgtm
/approve

hakman:
Thanks @rifelpet. I think we should find a way to fix the grid next.

rifelpet:
yea, i tried taking a look at the [docs](https://github.com/kubernetes/test-infra/blob/master/testgrid/config.md) the other day but couldnt figure out exactly what we need for all of our jobs

PR #17632, opened 2020-05-19
Move Kops to its own testgrid dasboard

hakman:
Kops dashboard is becoming pretty busy:
https://testgrid.k8s.io/sig-cluster-lifecycle-kops#Summary

Some order would help with this: https://github.com/kubernetes/test-infra/pull/17435#issuecomment-621346432

/cc @justinsb @rifelpet


config/jobs/kubernetes/kops/OWNERS
```diff
@@ -15,5 +15,3 @@ approvers:
 reviewers:
   - joshbranham
   - robinpercy
-labels:
-- sig/cluster-lifecycle
```
rifelpet, line 19: I'm wondering if we should keep these. Even if we make kops its own testgrid dashboard, the project is still owned by sig-cluster-lifecycle

hakman: We should add some labels I guess, but not sure if we want to keep this or add a new one more explicit.

hakman: Labels are back :)

config/tests/testgrids/config_test.go
```diff
@@ -55,6 +55,7 @@ var (
 	}
 	orgs = []string{
 		"conformance",
+		"kops",
```
justinsb, line 58: I take it we have to do this to create the dashboards without the prefix?  I'm not sure how controversial this is going to be...

I wonder if we could go into companies instead?  I feel like we can justify being on par with those (e.g. istio).

justinsb: cc @BenTheElder what do you think here?

config/jobs/image-pushing/k8s-staging-kops.yaml
```diff
@@ -5,7 +5,7 @@ postsubmits:
       cluster: k8s-infra-prow-build-trusted
       annotations:
         # this is the name of some testgrid dashboard to report to.
-        testgrid-dashboards: sig-cluster-lifecycle-kops
```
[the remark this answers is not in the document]
hakman: Agreed. Done.

justinsb:
This LGTM, question for @BenTheElder on whether we should be a company or an org, but it's easy enough to move around...

/approve
/lgtm

BenTheElder:
Spoke to @spiffxp: for now :man_shrugging: but there may be more enforcement of the organization of testgrid in the future.
/approve

hakman:
Thanks @justinsb & @BenTheElder 😄

PR #17634, opened 2020-05-19
Provide the job for e2e-node hugepages tests

cynepco3hahue:
Using multiple hugepages with different page size is an alpha feature under the k8s 1.18,
to promote it to the beta, e2e tests were provided, but these tests can not run under the serial
lane, because the feature gate should be enabled and the instance that will run tests, should
have 1Gb hugepage allocated. To satisfy these requests new jobs should be introduced.

Signed-off-by: Artyom Lukianov <alukiano@redhat.com>

### DIFF
--- a/config/jobs/kubernetes/sig-node/node-kubelet.yaml
+++ b/config/jobs/kubernetes/sig-node/node-kubelet.yaml
@@ -341,3 +341,34 @@ periodics:
     testgrid-dashboards: wg-resource-management, sig-node-kubelet
     testgrid-tab-name: kubelet-serial-gce-e2e-topology-manager
     testgrid-alert-email: vpickard@redhat.com, klueska@gmail.com
+
+- name: ci-kubernetes-node-kubelet-serial-hugepages
+  interval: 4h
+  labels:
+    preset-service-account: "true"
+    preset-k8s-ssh: "true"
+  spec:
+    containers:
+      - image: gcr.io/k8s-testimages/kubekins-e2e:v20200513-7c0e31a-master
+        args:
+          - --repo=k8s.io/kubernetes=master
+          - --timeout=240
+          - --root=/go/src
+          - --scenario=kubernetes_e2e
+          - --
+          - --deployment=node
+          - --gcp-project-type=node-e2e-project
+          - --gcp-zone=us-west1-b
+          - --node-args=--image-config-file=/workspace/test-infra/jobs/e2e_node/image-config-serial-hugepages.yaml
+          - --node-test-args=--feature-gates=HugePageStorageMediumSize=true --kubelet-flags="--cgroups-per-qos=true --cgroup-root=/"
+          - --node-tests=true
+          - --provider=gce
+          - --test_args=--nodes=1 --focus="\[Feature:HugePages\]"
+          - --timeout=180m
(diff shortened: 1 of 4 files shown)


cynepco3hahue:
@dims Does it any possibility that the job is working correctly before the merge?

dims:
@cynepco3hahue this one is hard to simulate, you have the right pattern to make sure that the `pull-kubernetes-node-kubelet-serial-hugepages` job does not run unless triggered in regular workflow. So i'd just try to get this PR merged first and then file a WIP PR in kubernetes/kubernetes to trigger that pull job to kick the tires. I usually just file the pull job first but this is fine, it won't hurt.

/approve
/lgtm

/assign @BenTheElder @spiffxp

cynepco3hahue:
@dims A lot of thanks for the review!

bsdnet:
Can we also add COS image too?

Also should we create a directory like mm or hugepages for better future?

cynepco3hahue:
@bsdnet COS image has problems with 1Gb hugepages, it does not expose 1Gb hugepages folders at all(but it has the relevant CPU flag), also I unsure if it was relevant to add an additional image because you should provide test pattern to run on top of this image. For example, the CPU manager and topology manager tests never use the COS image and use only Ubuntu images.

Regarding the folder hierarchy, I can add it, but I followed the pattern under the https://github.com/kubernetes/kubernetes/tree/master/test/e2e_node/jenkins, just specify the relevant image name and functionality that it should provide.

Give me know if you still insist on the folder hierarchy, I do not have strong opinion here, so I can add it.

bsdnet:
@cynepco3hahue I saw COS enable huge pages since https://cloud.google.com/container-optimized-os/docs/release-notes#cos-dev-60-9540-0-0. Yes, you are right, though cpu flag pdpe1gb is set, but there is no 1Gb huge pages under /sys/kernel/mm/hugepages/. Will take a further look.

As to "the CPU manager and topology manager tests never use the COS image and use only Ubuntu images.". I am new and do not have the history. Do you know any reason behind this?

For the folder, we can follow current pattern for now. It may need some refactoring in future for better organizing.

/lgtm

bsdnet:
I have done some background research, and COS does not enable runtime hugepage alloc. That is why 1Gb hugepages folder is not exposed.

cynepco3hahue:
@bsdnet Thanks for the information. Does it possible to enable it via cloud-init?

cynepco3hahue:
@bsdnet Regarding the topology and CPU manager I have no idea why it runs only n top of the ubuntu :(

bsdnet:
> @bsdnet Thanks for the information. Does it possible to enable it via cloud-init?

It is possible if we append 'hugepagesz=1G hugepages=1' into the /proc/cmdline by updating grub.cfg. But this needs one reboot.  Let me ask around.

bsdnet:
Yes, it is possible to enable "hugepages" in cloud-init, but it requires a "reboot". I am looking to how to enable the kernel option for future images.

cynepco3hahue:
@bsdnet Thanks for your effort!

cynepco3hahue:
@BenTheElder @spiffxp Hi guys, can you please review the PR?

PR #17640, opened 2020-05-19
Comment accordingly if /test all cant run any jobs

### LINKED ISSUE #17580: Bot feedback on `/test all` confusing when `/test all` does not trigger any jobs
**What happened**:

We have a repo where all jobs are conditionally triggered:

```
always_run: false
run_if_changed: <regex>
```

Recently we had a PR which made a change such that no job matched the criteria, so no job was run by default. As documented in https://github.com/kubernetes/test-infra/blob/master/prow/jobs.md#standard-triggering-and-execution-behavior-for-jobs, `/test all` means "trigger all automatically triggered jobs", which for such PR means _trigger no jobs_ (no jobs are triggered automatically). The bot responded with "The specified target(s) for /test were not found." and "Use `/test all` to run all jobs."

I expect the bot to at least hint that `/test all` does not really mean ALL jobs, but only those that would automatically be triggered, and also not provide the misleading "the specified target was not found" message 

`/test all` on a org/repo/branch which only has conditionally triggered jobs, but the PR does not trigger any of them,

https://github.com/openshift/installer/pull/3542#issuecomment-624550909

adshmh:
If no jobs can be run with `/test all` command, the trigger plugin
responds accordingly in the help comment. Closes #17580

### DIFF
--- a/prow/plugins/trigger/generic-comment.go
+++ b/prow/plugins/trigger/generic-comment.go
@@ -37,9 +37,10 @@ var (
 	retestWithTargetRe  = regexp.MustCompile(`(?m)^/retest[ \t]+\S+`)
 	testWithAnyTargetRe = regexp.MustCompile(`(?m)^/test[ \t]+\S+`)
 
-	testWithoutTargetNote = "The `/test` command needs one or more targets.\n"
-	retestWithTargetNote  = "The `/retest` command does not accept any targets.\n"
-	targetNotFoundNote    = "The specified target(s) for `/test` were not found.\n"
+	testWithoutTargetNote     = "The `/test` command needs one or more targets.\n"
+	retestWithTargetNote      = "The `/retest` command does not accept any targets.\n"
+	targetNotFoundNote        = "The specified target(s) for `/test` were not found.\n"
+	thereAreNoTestAllJobsNote = "No jobs can be run with `/test all`.\n"
 )
 
 func mayNeedHelpComment(body string) bool {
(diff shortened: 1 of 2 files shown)


prow/plugins/trigger/generic-comment.go
```diff
@@ -197,20 +198,36 @@ func FilterPresubmits(honorOkToTest bool, gitHubClient GitHubClient, body string
...
-	all := func(p config.Presubmit) (bool, bool, bool) {
-		return true, true, true
+
+	runWithTestAll, err := pjutil.FilterPresubmits(pjutil.TestAllFilter(), changes, branch, presubmits, logger)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	var triggerFilters []pjutil.Filter
+	for _, ps := range presubmits {
```
alvaroaleman, line 213: This needs to filter out presubmits that can not run via `ps.CouldRun(branch)`

prow/plugins/trigger/generic-comment.go
```diff
@@ -197,20 +198,36 @@ func FilterPresubmits(honorOkToTest bool, gitHubClient GitHubClient, body string
...
+
+	var runWithTestAllNames []string
+	for _, ps := range runWithTestAll {
+		runWithTestAllNames = append(runWithTestAllNames, ps.Name)
 	}
-	var available []string
-	for _, pre := range toTest {
-		available = append(available, pre.RerunCommand)
+	var runWithTriggerNames []string
+	for _, ps := range runWithTrigger {
```
alvaroaleman, line 226: If I am reading the code correctly this means this list also includes the `runWithTestAll` presubmit. Would it maybe make sense to exclude those?

prow/plugins/trigger/generic-comment.go
adshmh: Thank you for the review. I reviewed the code again and it seems `FilterPresubmits` already excludes those by calling `ShouldRun` on each presubmit job. I have updated the PR and added a test case to make sure this is covered.

prow/plugins/trigger/generic-comment.go
adshmh: Thank you for the review. Yes, this currently works by listing each individual presubmit that can be run. The exclusion you are referring to is currently done by commeting "Use `/test all` to run all jobs." if the two lists are the same, e.g.:

```
The following commands are available to trigger jobs:

    /test presubmit-1
    /test presubmit-2


Use /test all to run all jobs.
```

Otherwise, jobs will be listed individually under `/test all`, e.g.:

```
The following commands are available to trigger jobs:

    /test presubmit-1
    /test presubmit-2

Use /test all to run the following jobs:

    presubmit-1

```

Please let me know if this needs changing and I will update the PR.

PR #17642, opened 2020-05-19
S3 Credential Chain support

### LINKED ISSUE #17197: Support for S3 default credential provider chain
This is a fork issue from #11260  due to errors encountered when using the fallback gocloud default discovery if no s3-credentials-file is specified. See https://github.com/kubernetes/test-infra/issues/11260#issuecomment-613103711
https://github.com/kubernetes/test-infra/issues/11260#issuecomment-612880075.

Support for S3 default credential provider chain

**Why is this needed**:
[Static S3 credential](https://github.com/kubernetes/test-infra/blob/master/prow/io/providers/providers.go#L98) is not possible in our case since we only have federated account with MFA enabled and we can't do something about it due to corporate IT policies/restrictions.

mikesplain:
Thanks to @sbueringer and multiple reviewers(not tagging yet to avoid un-needed pings!) for the hard work on s3 support!  Based on the discussion in https://github.com/kubernetes/test-infra/issues/17197, https://github.com/kubernetes/test-infra/issues/17197, and https://github.com/kubernetes/test-infra/pull/17201, this PR adds support for the aws credential chain. 

Tested in an environment with kiam but should work with other deployments as well. 

Please take a look and make sure it meets your needs!  Holding as well since I know other s3 PRs are pending and waiting for some soak time.  All comments welcome!

/cc @apesternikov @sbueringer
/hold

Fixes #17197

### DIFF
--- a/prow/io/providers/BUILD.bazel
+++ b/prow/io/providers/BUILD.bazel
@@ -8,6 +8,8 @@ go_library(
     deps = [
         "@com_github_aws_aws_sdk_go//aws:go_default_library",
         "@com_github_aws_aws_sdk_go//aws/credentials:go_default_library",
+        "@com_github_aws_aws_sdk_go//aws/credentials/ec2rolecreds:go_default_library",
+        "@com_github_aws_aws_sdk_go//aws/ec2metadata:go_default_library",
         "@com_github_aws_aws_sdk_go//aws/session:go_default_library",
         "@dev_gocloud//blob:go_default_library",
         "@dev_gocloud//blob/memblob:go_default_library",
--- a/prow/io/providers/providers.go
+++ b/prow/io/providers/providers.go
@@ -25,6 +25,8 @@ import (
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/credentials"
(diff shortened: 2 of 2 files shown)


sbueringer:
Tested it in my env and also works with my local Minio setup. Thx!

/lgtm

Just so that watcher of the S3/Azure issue also notice it: https://github.com/kubernetes/test-infra/issues/11260

mikesplain:
Great thanks! 

/assign @fejta @alvaroaleman

alvaroaleman:
/approve
Looks good as in I don't think it will cause existing logic to break but I can not verify if it actually works :)

Lets wait till the end of the week with merging this and the other follow-ups.

sbueringer:
@mikesplain Absolutely reasonable to add this, but dumb question, doesn't it work without specifying a s3_credential_file? The code should fallback to the auto-discovery of gocloud

mikesplain:
@sbueringer that's a fair point. As discussed in slack:

* You can use s3 without the s3-credentials-file however it requires setting your current region via something like `AWS_REGION` environment variable
* I think this still has value so users can point at different regions or endpoints utilizing the s3-credentials-file without utilizing the actual access keys of the file and instead falling back to the provider chain.

PR #17643, opened 2020-05-19
triage: bail on certain global clusters after 30s

spiffxp:
triage works by clustering test failures in two stages:
- locally: create clusters of test failures for each unique test
- globally: merge each test's clusters into a global set of clusters

The clustering/merging is done by computing edit distance between the
failure text of each test failure or failure cluster and accepting the
first pair that has an edit distance of 10% of their combined length.

This can add up in the worst case, where edit distance is going to be
computed for every existing cluster before creating a new cluster.

We've arbitrarily handled it thus far by:
- truncating failure text to ~200k~ 10k chars
- bailing out on local clustering after 60s per unique test

This PR adds:
- bailing out on global clustering of pathological / low value clusters after 30s
- more logging to see where clustering is working vs. not

This should address https://github.com/kubernetes/test-infra/issues/17625

### DIFF
--- a/triage/summarize.py
+++ b/triage/summarize.py
@@ -49,6 +49,10 @@ flakeReasonOrdinalRE = re.compile(
     r'|(?<=minion-group-|default-pool-)[-0-9a-z]{4,}'  # node names
 )
 
+LONG_OUTPUT_LEN = 10000
+TRUNCATED_SEP = '\n...[truncated]...\n'
+MAX_CLUSTER_TEXT_LEN = LONG_OUTPUT_LEN + len(TRUNCATED_SEP)
+
 
 def normalize(s):
     """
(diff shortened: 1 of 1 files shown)


spiffxp:
/test pull-test-infra-bazel 
flake
```
ERROR: An error occurred during the fetch of repository 'bazel-base':
   Pull command failed: 2020/05/20 00:14:05 Running the Image Puller to pull images from a Docker Registry...
2020/05/20 00:14:06 Image pull was unsuccessful: unable to save remote image gcr.io/k8s-testimages/launcher.gcr.io/google/bazel@sha256:cefc822f93bb3dcf272ce3e4c5162b179d5c165584ace13856afed99662b87cd: unable to write image layers: unable to write image layer: unable to write the contents of layer 0 to /bazel-scratch/.cache/bazel/_bazel_root/05618a594cb3499a4b912817865867cd/external/bazel-base/image/000.tar.gz: read tcp 10.60.151.59:53786->74.125.70.128:443: read: connection reset by peer```

spiffxp:
/cc @dims @BenTheElder 
since you were involved with https://github.com/kubernetes/test-infra/pull/17629

dims:
LGTM this definitely helps with the additional logs :)

/lgtm

PR #17646, opened 2020-05-20
Do not warning on ghproxy for mkpj

### LINKED ISSUE #17549: mkpj should not expect github caching proxy
**What happened**:

> time="2020-05-07T15:26:26Z" level=warning msg="It doesn't look like you are using ghproxy to cache API calls to GitHub! This has become a required component of Prow and other components will soon be allowed to add features that may rapidly consume API ratelimit without caching. Starting May 1, 2020 use Prow components without ghproxy at your own risk! https://github.com/kubernetes/test-infra/tree/master/ghproxy#ghproxy

I do not expect to be warned about using the mkpj CLI without a ghproxy instance.

`config/mkpj.sh`

I hope mkpj isn't going to rapidly consume API ratelimit 🙃 
/area prow

chizhg:
`mkpj` does not need many accesses to the Github APIs, so it should not require ghproxy to be set. Adding an `AllowDirectAccess` field to allow suppressing the warning. And BTW removing one redundant argument from function `genJobSpec`.

Fixes #17549 

/cc @BenTheElder @cjwagner

### DIFF
--- a/prow/cmd/deck/main.go
+++ b/prow/cmd/deck/main.go
@@ -201,6 +201,7 @@ func gatherOptions(fs *flag.FlagSet, args ...string) options {
 	o.kubernetes.AddFlags(fs)
 	o.github.AddFlags(fs)
 	o.github.AllowAnonymous = true
+	o.github.AllowDirectAccess = true
 	fs.Parse(args)
 	return o
 }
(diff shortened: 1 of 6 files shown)


BenTheElder:
this makes sense to me, but I'll defer to cole on the new github option

BenTheElder:
(also: thanks!)

chizhg:
> This LGTM. I actually was planning to do exactly this for `plank`. Since it has a GH client for legacy reasons we shouldn't really be expecting it to use GH proxy. Can you make the same change for plank please?
> The other case I noticed was that Deck has a GH client that should normally use ghproxy, but shouldn't require it when the GH token is omitted since not all Prow instances use GitHub. I think we could easily handle this in Deck as well if we only emit a warning about ghproxy being required if the github token path is non-empty.

I have made the same change for plank and deck, please take a second look. Thanks.

PR #17649, opened 2020-05-20
bump.sh checks if the images exist before updating config files

chizhg:
This is a fix to the issue 156032855

It uses the approach suggested in https://stackoverflow.com/a/55344819/13578870 to query the manifest url, to determine if the image with the given tag exists or not, which IMO is better than listing all image tags or using `docker pull`.

/assign @cjwagner @fejta

### DIFF
--- a/prow/cmd/autobump/bump.sh
+++ b/prow/cmd/autobump/bump.sh
@@ -66,7 +66,7 @@ main() {
   else
     usage
   fi
-  echo -e "Bumping: 'gcr.io/k8s-prow/' images to $(color-version ${new_version}) ..." >&2
+  echo -e "Bumping: 'gcr.io/k8s-prow/' images to $(color-version "${new_version}") ..." >&2
 
   bumpfiles=($(add_suffix "$(split_on_commas "$COMPONENT_FILE_DIR")"))
   bumpfiles+=("${CONFIG_PATH}")
@@ -74,12 +74,30 @@ main() {
     bumpfiles+=($(grep -rl -e "gcr.io/k8s-prow/" "${JOB_CONFIG_PATH}"))
   fi
 
-  echo -e "Attempting to bump the following files: ${bumpfiles[@]}" >&2
-
+  echo "Attempting to bump the following files:" >&2
+  for bf in "${bumpfiles[@]}"; do
+    echo -e "$bf"
+  done
+  local token="$(gcloud auth print-access-token)"
   # Update image tags in the identified files.
-  filter="s/gcr.io\/k8s-prow\/\([[:alnum:]_-]\+\):v[a-f0-9-]\+/gcr.io\/k8s-prow\/\1:${new_version}/I"
+  local matcher="gcr.io\/k8s-prow\/\([[:alnum:]_-]\+\):v[a-f0-9-]\+"
+  local replacer="s/${matcher}/gcr.io\/k8s-prow\/\1:${new_version}/I"
   for file in "${bumpfiles[@]}"; do
-    ${SED} -i "${filter}" ${file}
+    local images="$(grep -o "${matcher}" "${file}")"
+    local arr=(${images//\\n/})
+    # image is in the format of gcr.io/k8s-prow/[image_name]:[tag]
+    for image in ${arr[@]+"${arr[@]}"}; do
+      echo "Checking the existence of ${image}"
+      # Use the Docker Registry v2 API to query the image manifest to check if the given image tag exists or not.
+      # The manifest_url is in the format of https://gcr.io/v2/k8s-prow/[image_name]/manifests/[tag]
+      # Check more details from https://stackoverflow.com/a/55344819/13578870
+      local manifest_url=$(echo "$image" | sed "s/:/\/manifests\//" | sed "s/gcr.io/https:\/\/gcr.io\/v2/")
+      if ! curl --fail -L -H "Authorization: Bearer $token" -o /dev/null -s "${manifest_url}"; then
+        echo "The image ${image} does not exist, please double check." >&2
+        exit 1
+      fi
+    done
+    ${SED} -i "${replacer}" "${file}"
   done
 
   echo "bump.sh completed successfully!" >&2
@@ -87,13 +105,13 @@ main() {
 
 check-args() {
   if [[ -z "${COMPONENT_FILE_DIR}" ]]; then
-    echo "ERROR: $COMPONENT_FILE_DIR must be specified." >&2
+    echo "ERROR: COMPONENT_FILE_DIR must be specified as an env var." >&2
   fi
   if [[ -z "${CONFIG_PATH}" ]]; then
-    echo "ERROR: $CONFIG_PATH must be specified." >&2
+    echo "ERROR: CONFIG_PATH must be specified as an env var." >&2
   fi
   if [[ -z "${JOB_CONFIG_PATH}" ]]; then
-    echo "ERROR: $JOB_CONFIG_PATH must be specified." >&2
+    echo "ERROR: JOB_CONFIG_PATH must be specified as an env var." >&2
   fi
 }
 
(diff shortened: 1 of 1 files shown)


prow/cmd/autobump/bump.sh
```diff
@@ -66,34 +66,52 @@ main() {
...
   bumpfiles+=("${CONFIG_PATH}")
   if [[ -n "${JOB_CONFIG_PATH}" ]]; then
     bumpfiles+=($(grep -rl -e "gcr.io/k8s-prow/" "${JOB_CONFIG_PATH}"))
   fi
 
-  echo -e "Attempting to bump the following files: ${bumpfiles[@]}" >&2
+  echo -e "Attempting to bump the following files:" >&2
+  IFS=$'\n'; echo -e "${bumpfiles[*]}" >&2
 
+  local token="$(gcloud config config-helper --format 'value(credential.access_token)')"
```
fejta, line 80: `gcloud auth print-access-token`

prow/cmd/autobump/bump.sh
```diff
@@ -66,34 +66,52 @@ main() {
...
+    local images="$(grep -o "${matcher}" "${file}")"
+    local array=(${images//\\n/})
+    for i in "${!array[@]}"; do
+      # image is in the format of gcr.io/k8s-prow/[image_name]:[tag]
+      local image="${array[i]}"
+      # Use the Docker Registry v2 API to query the image manifest to check if the given image tag exists or not.
+      # The manifest_url is in the format of https://gcr.io/v2/k8s-prow/[image_name]/manifests/[tag]
+      # Check more details from https://stackoverflow.com/a/55344819/13578870
+      local manifest_url=$(echo "$image" | sed "s/:/\/manifests\//" | sed "s/gcr.io/https:\/\/gcr.io\/v2/")
+      local code=$(curl -H "Authorization: Bearer $token" -o /dev/null -s -w "%{http_code}" "${manifest_url}")
```
fejta, line 94: `--fail`? (may also want -L)

prow/cmd/autobump/bump.sh
```diff
@@ -66,34 +66,52 @@ main() {
...
+  echo -e "Bumping: 'gcr.io/k8s-prow/' images to $(color-version "${new_version}") ..." >&2
 
   bumpfiles=($(add_suffix "$(split_on_commas "$COMPONENT_FILE_DIR")"))
   bumpfiles+=("${CONFIG_PATH}")
   if [[ -n "${JOB_CONFIG_PATH}" ]]; then
     bumpfiles+=($(grep -rl -e "gcr.io/k8s-prow/" "${JOB_CONFIG_PATH}"))
   fi
 
-  echo -e "Attempting to bump the following files: ${bumpfiles[@]}" >&2
+  echo -e "Attempting to bump the following files:" >&2
```
fejta, line 77: Not using any escapes

prow/cmd/autobump/bump.sh
```diff
@@ -66,34 +66,52 @@ main() {
...
+  IFS=$'\n'; echo -e "${bumpfiles[*]}" >&2
```
fejta, line 78: wouldn't `IFS=$'\n' echo ...` be more appropriate than permanently changing IFS (via the `;`')? also

```sh
for bf in "${bumpfiles[@]}"; do
  echo -e "$bf"
done
```

is IMO easier to understand than working with IFS

prow/cmd/autobump/bump.sh
```diff
@@ -66,34 +66,52 @@ main() {
...
+  local token="$(gcloud config config-helper --format 'value(credential.access_token)')"
   # Update image tags in the identified files.
-  filter="s/gcr.io\/k8s-prow\/\([[:alnum:]_-]\+\):v[a-f0-9-]\+/gcr.io\/k8s-prow\/\1:${new_version}/I"
+  local matcher="gcr.io\/k8s-prow\/\([[:alnum:]_-]\+\):v[a-f0-9-]\+"
+  local replacer="s/${matcher}/gcr.io\/k8s-prow\/\1:${new_version}/I"
   for file in "${bumpfiles[@]}"; do
-    ${SED} -i "${filter}" ${file}
+    local images="$(grep -o "${matcher}" "${file}")"
+    local array=(${images//\\n/})
+    for i in "${!array[@]}"; do
```
fejta, line 87: No need for the index you're not doing anything with: `for image in "${array[@]}"; do`

prow/cmd/autobump/bump.sh
chizhg: Changed, thanks!

prow/cmd/autobump/bump.sh
chizhg: Added `--fail` and `-L`

prow/cmd/autobump/bump.sh
chizhg: Changed to for loop. Thanks!

prow/cmd/autobump/bump.sh
chizhg: Removed `-e` for this echo command.

prow/cmd/autobump/bump.sh
chizhg: I will get `array[*]: unbound variable` after changing to `for image in "${array[@]}"` when there is no matched image in the file. Not sure how to deal with it so I just kept as it is.

chizhg:
> Not wild about investing much into this bash script
> 
> I'd prefer we work on making https://github.com/kubernetes/test-infra/tree/master/experiment/autobumper general purpose
> 
> https://github.com/kubernetes/test-infra/tree/master/experiment/image-bumper

Totally agree with that.
This will mainly be a short term prevention to the issue we had recently. In the long term we should do it with Go. I will look into how to unify these autobump tools later.

prow/cmd/autobump/bump.sh
fejta: This works just fine for me:


```sh
 set -o errexit
 set -o nounset
 set -o pipefail
 shopt -s dotglob
 shopt -s nullglob
 
 array=()
 for image in "${array[@]}"; do
     echo hi $image
 done
```

prow/cmd/autobump/bump.sh
```diff
@@ -66,34 +66,54 @@ main() {
...
+      # image is in the format of gcr.io/k8s-prow/[image_name]:[tag]
+      local image="${array[i]}"
+      echo "Checking the existence of ${image} "
+      # Use the Docker Registry v2 API to query the image manifest to check if the given image tag exists or not.
+      # The manifest_url is in the format of https://gcr.io/v2/k8s-prow/[image_name]/manifests/[tag]
+      # Check more details from https://stackoverflow.com/a/55344819/13578870
+      local manifest_url=$(echo "$image" | sed "s/:/\/manifests\//" | sed "s/gcr.io/https:\/\/gcr.io\/v2/")
+      local code=$(curl --fail -L -H "Authorization: Bearer $token" -o /dev/null -s -w "%{http_code}" "${manifest_url}")
```
fejta, line 96: IMO there's no need for the `-w %{http_code}` the `--fail` handles this.

prow/cmd/autobump/bump.sh
chizhg: It seems for different bash versions, the behavior is different for handling empty arrays.
I have changed to `for image in ${arr[@]+"${arr[@]}"};` as suggested in https://stackoverflow.com/questions/7577052/bash-empty-array-expansion-with-set-u, and it works.

prow/cmd/autobump/bump.sh
chizhg: Thanks, done.

chizhg:
/hold cancel
It's now ready for a second review, thanks.

prow/cmd/autobump/bump.sh
```diff
@@ -66,34 +66,52 @@ main() {
...
+    ${SED} -i "${replacer}" "${file}"
   done
 
   echo "bump.sh completed successfully!" >&2
 }
 
 check-args() {
   if [[ -z "${COMPONENT_FILE_DIR}" ]]; then
-    echo "ERROR: $COMPONENT_FILE_DIR must be specified." >&2
+    echo "ERROR: COMPONENT_FILE_DIR must be specified as an env var." >&2
```
fejta, line 108: if you use ' instead of " then $COMPONENT_FILE_DIR will render correctly

PR #17650, opened 2020-05-20
Deck: support bucket/folder whitelist

clarketm:
Added _bucket_ and _folder_ whitelisting to `Deck` in order to restrict artifact requests to specific storage locations. This feature can be enabled by setting `deck.enable_whitelist` in the Prow config. 

By default, _buckets_ specified in the `plank.decoration_config[*].gcs_configuration.bucket` are whitelisted and the `pr-logs` and `logs` _folders_ are whitelisted. Additional _buckets_ can be whitelisted by adding them to `deck.additional_buckets`. Additional _folders_ can be whitelisted by adding them to `deck.additional_folders`.


prow/cmd/deck/main.go
```diff
@@ -1663,3 +1670,40 @@ func defaultLensRemoteConfig(lfc *config.LensFileConfig) error {
...
+	}
+
+	parts := strings.SplitN(path, "/", 4)
+	if len(parts) < 4 {
+		return fmt.Errorf("invalid path: %s expected <key-type>/<bucket>/<folder>/<sub-folder...>", path)
+	}
+
+	// Check bucket
+	bucket := parts[1]
+	bwl := sets.String{}
```
michelle192837, line 1687: I think this and fwl below can be a bit clearer.

prow/cmd/deck/main.go
```diff
@@ -1663,3 +1670,40 @@ func defaultLensRemoteConfig(lfc *config.LensFileConfig) error {
...
+	for _, dc := range cfg().Plank.DefaultDecorationConfigs {
+		bwl.Insert(dc.GCSConfiguration.Bucket)
+	}
+	for _, bucket := range cfg().Deck.AdditionalBuckets {
+		bwl.Insert(bucket)
```
cjwagner, line 1692: nit: You can replace the insertion loops over `[]string`s with something like this: `bwl.Insert(cfg().Deck.AdditionalBuckets...)`.

prow/cmd/deck/main.go
```diff
@@ -1004,7 +1005,9 @@ func renderSpyglass(ctx context.Context, sg *spyglass.Spyglass, cfg config.Gette
 		return "", fmt.Errorf("error when resolving real path %s: %v", src, err)
 	}
 	src = realPath
-
+	if err := ValidatePath(cfg, src); err != nil {
+		return "", err
```
cjwagner, line 1009: It looks like this error will still reach the logs and be treated as 5xx instead of being treated as a 4xx, a correct response to an invalid request. Please ensure that we are handling invalid paths as request errors in all uses.

prow/cmd/deck/main.go
```diff
@@ -1663,3 +1670,40 @@ func defaultLensRemoteConfig(lfc *config.LensFileConfig) error {
...
+	}
+	if !bwl.Has(bucket) {
+		return fmt.Errorf("bucket %q not in whitelist %v", bucket, bwl)
+	}
+
+	// Check folder
+	folder := parts[2]
+	fwl := sets.NewString(gcs.PRLogs, gcs.NonPRLogs)
+	for _, folder := range cfg().Deck.AdditionalFolders {
+		fwl.Insert(folder)
```
cjwagner, line 1702: nit: The folder whitelist is currently limited to whitelisting top-level folders. It could be made more flexibly by matching against the full path (with a prefix or regexp match) to allow specific sub folders to be whitelisted.
I don't think this matters much though.

prow/cmd/deck/main.go
```diff
@@ -1663,3 +1670,40 @@ func defaultLensRemoteConfig(lfc *config.LensFileConfig) error {
...
+	parts := strings.SplitN(path, "/", 4)
+	if len(parts) < 4 {
+		return fmt.Errorf("invalid path: %s expected <key-type>/<bucket>/<folder>/<sub-folder...>", path)
+	}
+
+	// Check bucket
+	bucket := parts[1]
+	bwl := sets.String{}
+	for _, dc := range cfg().Plank.DefaultDecorationConfigs {
+		bwl.Insert(dc.GCSConfiguration.Bucket)
```
cjwagner, line 1689: I think it would be better if we also included buckets specified in job specific decoration config so they don't have to be included in `AdditionalBuckets`. I'm not sure if that is worthwhile or quick enough to check though.

prow/ANNOUNCEMENTS.md
```diff
@@ -3,6 +3,11 @@
 ## New features
 
 New features added to each component:
+ - *May 22, 2020* Added _bucket_ and _folder_ whitelisting to Deck in order to restrict artifact requests to 
+     specific storage locations. This feature can be enabled by setting `deck.enable_whitelist` in the Prow config.
+     By default, _buckets_ specified in the `plank.decoration_config[*].gcs_configuration.bucket` are whitelisted and 
+     the `pr-logs` and `logs` _folders_ are whitelisted. Additional _buckets_ can be whitelisted by adding them to 
+     `deck.additional_buckets`. Additional _folders_ can be whitelisted by adding them to `deck.additional_folders`.
```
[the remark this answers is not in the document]
clarketm: If we want this granularity, we can configure `additional_folders` per bucket.

prow/cmd/deck/main.go
clarketm: I went with whitelisting of top-level folders because it is the most performant and simplest way to gate access. If prefix or regex matching is a need that arises then I can add it.

prow/cmd/deck/main.go
clarketm: The major downside is that on _every_ validated request each job will need to be traversed. Also, it cannot really be cached since any precomputed values will not include subsequent config updates. Let me investigate the feasibility.

prow/cmd/deck/main.go
clarketm: I agree 100% that these should be client errors (4xx), but since it requires some restructure/refactor would it be possible to do in a follow-up?

prow/cmd/deck/main.go
clarketm: I cleaned the set insertion steps up a bit.

clarketm:
@michelle192837 / @cjwagner PTAL

e-blackwelder:
I can pick up this change for clarketm

e-blackwelder:
Following up via [PR/18001](https://github.com/kubernetes/test-infra/pull/18001). I think we can close this one.

clarketm:
> Following up via [PR/18001](https://github.com/kubernetes/test-infra/pull/18001). I think we can close this one.

Thanks @e-blackwelder. I will close this PR.

PR #17617, opened 2020-05-16
Update cos-stable images to the latest versions

bart0sh:
e2e node tests fail to find cos-stable images with this error:
    Error 400: Invalid value for field 'resource.disks[0].initializeParams.sourceImage':
        'projects/cos-cloud/global/images/cos-stable-73-11647-510-0'.
    The referenced image resource cannot be found., invalid

Updating to the latest cos-stable versions should fix this issue.

Here is an example of the failed build: https://storage.googleapis.com/kubernetes-jenkins/logs/ci-kubernetes-node-kubelet-conformance/1261760411095535627/build-log.txt

PR #17656, opened 2020-05-20
add metadata to cos beta images

bart0sh:
COS beta images have /tmp mounted with 'noexec' mount option.
This makes it impossible for run_remote to execute binaries
from its workspace. This results in failed kubelet tests,
e.g. ci-kubernetes-node-kubelet-benchmark and
ci-kubernetes-node-kubelet-conformance test suites fail with
"failed to run command './ginkgo': Permission denied" error

Adding metadata pointing to cloud-init(gci-init.yaml) should solve
this as cloud-init remounts /tmp with 'exec' mount option.

### DIFF
--- a/jobs/e2e_node/benchmark-config.yaml
+++ b/jobs/e2e_node/benchmark-config.yaml
@@ -100,30 +100,35 @@ images:
     image: cos-beta-81-12871-44-0
     project: cos-cloud
     machine: n1-standard-1
+    metadata: "user-data<test/e2e_node/jenkins/gci-init.yaml,gci-update-strategy=update_disabled"
     tests:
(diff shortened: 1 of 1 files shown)


spiffxp:
Going to ask if we think we should revert https://github.com/kubernetes/test-infra/pull/17617 instead of trying to fix forward

bart0sh:
I don't think we should revert it. If we do we'll have cos-stable- images that don't exists in the configs again.

bart0sh:
I'll roll cos-77 to cos-stable-81 soon anyway.

spiffxp:
/hold cancel
this doesn't appear to cause a merge conflict, I'm fine with it

Issue #14201, opened 2019-09-05
verify-owners: misleading error message


stevekuznetsov:
Currently the `verify-owners` plugin states that users must be members of the org:

https://github.com/kubernetes/test-infra/blob/4782ba3a238e4c52de4a39d4e4b25f0822dc61ef/prow/plugins/verify-owners/verify-owners.go#L46

When in reality it is using the `trigger.TrustedUser` call:

https://github.com/kubernetes/test-infra/blob/4782ba3a238e4c52de4a39d4e4b25f0822dc61ef/prow/plugins/verify-owners/verify-owners.go#L500

This uses the `trigger` configuration to do many checks, only one of which is the organization membership one, and only conditionally:

https://github.com/kubernetes/test-infra/blob/4782ba3a238e4c52de4a39d4e4b25f0822dc61ef/prow/plugins/trigger/trigger.go#L164-L194

We should make `trigger.TrustedUser` return a reason for not being trusted so that we can have correct messages.

/help-wanted

nikhita:
Related issues: https://github.com/kubernetes/test-infra/issues/13740, https://github.com/kubernetes/test-infra/issues/13002

stevekuznetsov:
@nikhita you are so fast! <3

kikisdeliveryservice:
hi, could i work on this?

stevekuznetsov:
@kikisdeliveryservice absolutely!

cblecker:
@kikisdeliveryservice Is this something you're still looking to work on? :)

kikisdeliveryservice:
@cblecker yes apologies, I have a wip that I will clean up and push a PR in the next few days

sladyn98:
@stevekuznetsov Do we need to change the error message here or just add the reason as to why we arent trusted like not an org admin

cblecker:
Fixed in #17658

PR #17658, opened 2020-05-20
Trigger's TrustedUser returns a reason for untrusted. Consumed by verify-owners.

michaelmdresser:
Addresses https://github.com/kubernetes/test-infra/issues/14201

- Instead of a bool, Trigger's `TrustedUser` now returns a struct with the bool and the reason for a trust denial (if trust was denied).
- All calls to `TrustedUser` were updated to extract the bool from the struct.
- verify-owners's Github comment structure was updated to support the additional error messaging - the base comment pattern has more detail and the details include `TrustedUser`'s reason for denial. This addresses the ask in #14201

### DIFF
--- a/prow/plugins/trigger/trigger.go
+++ b/prow/plugins/trigger/trigger.go
@@ -48,17 +48,17 @@ const (
 	notSecondaryMember
 )
 
-// untrustedReasonString constructs a string explaining the reason for a user's denail of trust
+// String constructs a string explaining the reason for a user's denial of trust
 // from untrustedReason as described above.
-func untrustedReasonString(reasons untrustedReason) string {
+func (u untrustedReason) String() string {
 	var response string
-	if reasons&notMember != 0 {
+	if u&notMember != 0 {
 		response += "User is not a member of the org. "
 	}
-	if reasons&notCollaborator != 0 {
+	if u&notCollaborator != 0 {
 		response += "User is not a collaborator. "
 	}
-	if reasons&notSecondaryMember != 0 {
+	if u&notSecondaryMember != 0 {
 		response += "User is not a member of the trusted secondary org. "
 	}
 	response += "Satisfy at least one of these conditions to make the user trusted."
@@ -224,9 +224,9 @@ func TrustedUser(ghc trustedUserClient, onlyOrgMembers bool, trustedOrg, user, o
 	if trustedOrg == "" || trustedOrg == org {
 		// the if/else is only to improve error messaging
 		if onlyOrgMembers {
-			return TrustedUserResponse{IsTrusted: false, Reason: untrustedReasonString(notMember)}, nil // No trusted org and/or it is the same
+			return TrustedUserResponse{IsTrusted: false, Reason: notMember.String()}, nil // No trusted org and/or it is the same
 		}
-		return TrustedUserResponse{IsTrusted: false, Reason: untrustedReasonString(notMember | notCollaborator)}, nil // No trusted org and/or it is the same
+		return TrustedUserResponse{IsTrusted: false, Reason: (notMember | notCollaborator).String()}, nil // No trusted org and/or it is the same
 	}
 
 	// Check the second trusted org.
@@ -235,13 +235,13 @@ func TrustedUser(ghc trustedUserClient, onlyOrgMembers bool, trustedOrg, user, o
 		return errorResponse, fmt.Errorf("error in IsMember(%s): %v", trustedOrg, err)
 	} else if member {
 		return okResponse, nil
-	} else {
-		// the if/else is only to improve error messaging
-		if onlyOrgMembers {
-			return TrustedUserResponse{IsTrusted: false, Reason: untrustedReasonString(notMember | notSecondaryMember)}, nil
-		}
-		return TrustedUserResponse{IsTrusted: false, Reason: untrustedReasonString(notMember | notSecondaryMember | notCollaborator)}, nil
 	}
+
+	// the if/else is only to improve error messaging
+	if onlyOrgMembers {
+		return TrustedUserResponse{IsTrusted: false, Reason: (notMember | notSecondaryMember).String()}, nil
+	}
+	return TrustedUserResponse{IsTrusted: false, Reason: (notMember | notSecondaryMember | notCollaborator).String()}, nil
 }
 
 func skippedStatusFor(context string) github.Status {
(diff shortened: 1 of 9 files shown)


rancerbeta:
Bear with me, since I might be missing a bit of context. 

It looks like this function usually returns an error when the user is received as untrusted. Since that is already the case, you can have it return a special error with the reason encapsulated within that error. Otherwise it can be confusing to the func user whether they should check the object for a reason, or the error. This way of modifying means that you won't need to change the return format of the func

michaelmdresser:
There is a `return false, nil` case in multiple locations. See https://github.com/kubernetes/test-infra/pull/17658/files#diff-f1ca2651a323123714c99fa467869568L186 and https://github.com/kubernetes/test-infra/pull/17658/files#diff-f1ca2651a323123714c99fa467869568L194 (where member can be true or false)

@rancerbeta

rancerbeta:
Got it, that sgtm

prow/plugins/trigger/trigger.go
```diff
@@ -36,6 +36,11 @@ import (
 const (
 	// PluginName is the name of the trigger plugin
 	PluginName = "trigger"
+
+	nonTrustedNotMember                                  = "User is not a member of the org."
+	nonTrustedNotMemberNotCollaborator                   = "User is not a member of the org and is not a collaborator. Either will make the user trusted."
+	nonTrustedNotMemberNotSecondaryMember                = "User is not a member of the org or the trusted (secondary) org"
+	nonTrustedNotMemberNotSecondaryMemberNotCollaborator = "User is not a member of the org or the trusted (secondary) org and is not a collaborator. Any of these being true will make the user trusted."
```
amwat, line 43: There seems to be redundancies in these consts as well. It looks like there's 3 main reasons why users aren't trusted, might be worthwhile to turn those into enums and add a func that constructs a reason based on 0 or more of those reasons.

something like: https://play.golang.org/p/S27-bXXAhW-

michaelmdresser: Added. That's really cool.

cblecker:
Language looks fine to me. Code is supported by tests. Nothing controversial.
/approve

/assign @amwat 
Leaving the LGTM to you :)

prow/plugins/trigger/trigger.go
```diff
@@ -38,6 +38,32 @@ const (
...
+
+const (
+	notMember untrustedReason = 1 << iota
+	notCollaborator
+	notSecondaryMember
+)
+
+// untrustedReasonString constructs a string explaining the reason for a user's denail of trust
+// from untrustedReason as described above.
+func untrustedReasonString(reasons untrustedReason) (response string) {
```
amwat, line 53: nit: for consistency, let's not use named returns, move it to a var?

michaelmdresser: Done.

prow/plugins/verify-owners/verify-owners.go
```diff
@@ -41,14 +41,19 @@ import (
...
+	ownersFileName          = "OWNERS"
+	ownersAliasesFileName   = "OWNERS_ALIASES"
+	untrustedResponseFormat = `The following users are mentioned in %s file(s) but are untrusted for the following reasons. One way to make the user trusted is to add them as [members](%s) of the %s org. You can then trigger verification by writing ` + "`/verify-owners`" + ` in a comment.`
 )
 
+type nonTrustedReasons struct {
+	// files is a list of files they are being added in
+	files []string
+	// trigger is the reason that trigger's TrustedUser responds with for a failed trust check
+	trigger string
```
amwat, line 54: trigger can be confusing here, maybe rename it to `reason` ?

prow/plugins/trigger/trigger.go
```diff
@@ -160,38 +187,61 @@ func handlePush(pc plugins.Agent, pe github.PushEvent) error {
...
+		}
+		return TrustedUserResponse{IsTrusted: false, Reason: untrustedReasonString(notMember | notCollaborator)}, nil // No trusted org and/or it is the same
 	}
 
 	// Check the second trusted org.
 	member, err := ghc.IsMember(trustedOrg, user)
 	if err != nil {
-		return false, fmt.Errorf("error in IsMember(%s): %v", trustedOrg, err)
+		return errorResponse, fmt.Errorf("error in IsMember(%s): %v", trustedOrg, err)
+	} else if member {
```
amwat, line 236: hmm not sure why lint didn't catch this, but drop the elses since all code paths are returns?

prow/plugins/trigger/trigger.go
```diff
@@ -38,6 +38,33 @@ const (
...
+
+const (
+	notMember untrustedReason = 1 << iota
+	notCollaborator
+	notSecondaryMember
+)
+
+// untrustedReasonString constructs a string explaining the reason for a user's denail of trust
+// from untrustedReason as described above.
+func untrustedReasonString(reasons untrustedReason) string {
```
amwat, line 53: optinal nit: looking at the function name maybe a better fit would be to implement stringer on the type instead.

`func (u untrustedReason) String() string`

prow/plugins/trigger/trigger.go
```diff
@@ -38,6 +38,33 @@ const (
...
+	var response string
+	if reasons&notMember != 0 {
+		response += "User is not a member of the org. "
```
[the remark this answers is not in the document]
michaelmdresser: I would suggest opting for no `\n` because the string is actually used in a bulleted list, nested slightly.

prow/plugins/verify-owners/verify-owners.go
michaelmdresser: Called it `triggerReason` to minimize ambiguity.

prow/plugins/trigger/trigger.go
michaelmdresser: Definitely an interesting lint bug, done.

prow/plugins/trigger/trigger.go
michaelmdresser: Done.

prow/plugins/trigger/trigger.go
```diff
@@ -38,6 +38,33 @@ const (
...
+// why a user is not trusted by TrustedUser. It is used to generate messaging for users.
+type untrustedReason int
+
+const (
+	notMember untrustedReason = 1 << iota
+	notCollaborator
+	notSecondaryMember
+)
+
+// untrustedReasonString constructs a string explaining the reason for a user's denail of trust
```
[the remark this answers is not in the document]
michaelmdresser: Done.

PR #17660, opened 2020-05-20
Revert "Update cos-stable images to the latest cos lts versions"

spiffxp:
This reverts #17617 as it appears we're causing release-blocking and merge-blocking jobs to fail

ref: https://github.com/kubernetes/kubernetes/issues/91292

### DIFF
--- a/jobs/e2e_node/benchmark-config.yaml
+++ b/jobs/e2e_node/benchmark-config.yaml
@@ -1,21 +1,21 @@
 ---
 images:
   cosstable2-resource1:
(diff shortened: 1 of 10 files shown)


MHBauer:
sounds like a good idea to me, hopefully it goes back to silent failure and not blocking failure
/lgtm

MHBauer:
https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-node-kubelet/1263203433704329216 looks back to skipping .

PR #17662, opened 2020-05-20
disable node logging by default for kubetest bash provider

spiffxp:
now that we're running e2e tests in k8s-infra-prow-build we can see
the cost of resources, and it turns out k/k's cluster/kube-up.sh logs
to stackdriver by default, which costs more then the cluster under test

the thing is, most component and node logs end up in a ci job's artifacts
directory, and community members haven't ever been able to access logs in
stackdriver anyway... so these logs are redundant

there may be humans who rely on kube-up.sh's default behavior of always
logging, so instead we default to explicitly disabling logging here in
kubetest, with an option to enable it if we discover there's some job
that really needs this option

convention for flag parsing and naming was copied from kops deployer

ref: https://github.com/kubernetes/k8s.io/issues/863

### DIFF
--- a/kubetest/bash.go
+++ b/kubetest/bash.go
@@ -17,6 +17,7 @@ limitations under the License.
 package main
 
 import (
+	"flag"
 	"fmt"
 	"os"
 	"os/exec"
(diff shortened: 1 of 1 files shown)


kubetest/bash.go
```diff
@@ -30,17 +31,22 @@ type bashDeployer struct {
 	gcpZone                 string
 	gcpSSHProxyInstanceName string
 	provider                string
+	nodeLoggingEnabled      bool
 }
 
 var _ deployer = &bashDeployer{}
 
+var (
+	bashNodeLogging = flag.Bool("bash-node-logging", false, "(bash only) enable node logging to gcp")
```
BenTheElder, line 40: what's one more global + global flag, among friends? :upside_down_face: 

@amwat, we should just keep this off in kubetest2, w/o the flag.

PR #17663, opened 2020-05-20
trigger jobs when PR is switched from draft to normal

### LINKED ISSUE #17627: Trigger plugin doesn't trigger jobs when PR is switched from draft to normal.
**What happened**:
A PR was created as a draft, no tests were run because it was a draft. The PR was later switched from draft to normal, but tests were not triggered.

Tests should trigger when the PR is switched from draft to normal in the same way that tests would run if the PR was created as a normal PR.

https://github.com/kubeflow/manifests/pull/1175

There might have been some changes in the GH API/webhook behavior. ref: https://github.com/kubernetes/test-infra/issues/17583

/help

adshmh:
Trigger jobs when a draft PR is changed to normal. Closes #17627.

### DIFF
--- a/prow/plugins/trigger/pull-request.go
+++ b/prow/plugins/trigger/pull-request.go
@@ -159,7 +159,10 @@ func handlePR(c Client, trigger plugins.Trigger, pr github.PullRequestEvent) err
 			c.Logger.WithError(err).Error("Failed to abort jobs for closed pull request")
 			return err
 		}
+	case github.PullRequestActionReadyForReview:
+		return buildAllIfTrusted(c, trigger, pr, baseSHA, presubmits)
 	}
+
 	return nil
 }
 
--- a/prow/plugins/trigger/pull-request.go
+++ b/prow/plugins/trigger/pull-request.go
@@ -146,6 +146,11 @@ func handlePR(c Client, trigger plugins.Trigger, pr github.PullRequestEvent) err
 		}
 	case github.PullRequestActionReadyForReview:
 		return buildAllIfTrusted(c, trigger, pr, baseSHA, presubmits)
+	case github.PullRequestConvertedToDraft:
+		if err := abortAllJobs(c, &pr.PullRequest); err != nil {
+			c.Logger.WithError(err).Error("Failed to abort jobs for pull request converted to draft")
+			return err
+		}
 	}
 
 	return nil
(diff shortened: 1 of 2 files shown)


prow/plugins/trigger/pull-request.go
```diff
@@ -180,6 +173,7 @@ func abortAllJobs(c Client, pr *github.PullRequest) error {
 		if job.Complete() {
 			continue
 		}
+		job.SetComplete()
```
alvaroaleman, line 176: It is actually important that we do not do this, because Planks aborting ignores completed jobs.

prow/plugins/trigger/pull-request_test.go
```diff
@@ -330,6 +366,24 @@ func TestHandlePullRequest(t *testing.T) {
 			prAction:    github.PullRequestActionLabeled,
 			ShouldBuild: false,
 		},
+		{
+			name: "Abort jobs if PR is closed",
```
alvaroaleman, line 370: Thanks for adding these tests!

prow/plugins/trigger/pull-request_test.go
```diff
@@ -424,6 +478,16 @@ func TestHandlePullRequest(t *testing.T) {
 			} else if !tc.ShouldComment && len(g.IssueCommentsAdded) > 0 {
 				t.Errorf("Expected no comments to github, but got %d", len(g.IssueCommentsAdded))
 			}
+			if tc.jobToAbort != nil {
+				pj, err := fakeProwJobClient.ProwV1().ProwJobs("namespace").Get(tc.jobToAbort.Name, metav1.GetOptions{})
+				if err != nil {
+					t.Fatalf("failed to get prowjob: %v", err)
+				}
+
+				if pj.Status.State != prowapi.AbortedState {
```
alvaroaleman, line 487: Would you mind to also verify here that the job is not set to `Completed`? It seems I didn't think of adding a test for that when I built that, even though its important behavior

prow/plugins/trigger/pull-request.go
adshmh: Thank you for catching that. Fixed (It seems I was using an old CRD for prowjobs, which required `completionTime` on `Aborted` state).

prow/plugins/trigger/pull-request_test.go
adshmh: Thank you for the review.

prow/plugins/trigger/pull-request_test.go
adshmh: Thank you for the review. Done.

cjwagner:
Thanks @adshmh!

PR #17668, opened 2020-05-20
Run triage every 2h

spiffxp:
It looks like the job is now staying under 100m with the last fix
so let's get more frequent updates


spiffxp:

dims:
thanks @spiffxp

PR #17669, opened 2020-05-21
Specified instance size for flaky kubelet e2e tests to run on

alejandrox1:
kubelet e2e tests labeled as "flaky" require workloads that request at
least 15 cpus and at least 16 Gi of memory.
A n1-standard-32 instance offers 32 cpus and 120 GB of ram, enough for
all e2e tests to run.

xref: https://github.com/kubernetes/kubernetes/issues/91263

Signed-off-by: alejandrox1 <alarcj137@gmail.com>

### DIFF
--- a/config/jobs/kubernetes/sig-node/node-kubelet.yaml
+++ b/config/jobs/kubernetes/sig-node/node-kubelet.yaml
@@ -177,6 +177,7 @@ periodics:
       - --deployment=node
       - --gcp-project-type=node-e2e-project
       - --gcp-zone=us-west1-b
+      - --gcp-node-size=n1-standard-32
       - --node-args=--image-config-file=/workspace/test-infra/jobs/e2e_node/image-config.yaml
       - --node-test-args= --kubelet-flags="--cgroups-per-qos=true --cgroup-root=/"
       - --node-tests=true


alejandrox1:
/assign @yujuhong @Random-Liu @dchen1107 
ptal

spiffxp:
/hold
I don't think this is the right way to change the machine-type for node tests.  Look at the config files in jobs/node_e2e for examples

I also question whether we need this job or not.  Has it always just been restricted to these large performance tests?

alejandrox1:
> I don't think this is the right way to change the machine-type for node tests. Look at the config files in jobs/node_e2e for examples

https://github.com/kubernetes/test-infra/blob/0952907d173f7bebbf11662f4c3ef5f412bbd756/jobs/e2e_node/benchmark-config.yaml#L6

---

> I also question whether we need this job or not. Has it always just been restricted to these large performance tests? I would try digging further back in history to see if this job has ever passed. Relevant history may be in k/k's test/e2e_node directory

Doing some digging. This job came into the picture through https://github.com/kubernetes/test-infra/pull/8673.
However, it seems that the failing tests were initially categorized as becnhmark tests but then labeled as flaky because they never passed, see https://github.com/kubernetes/kubernetes/pull/76799 and https://github.com/kubernetes/kubernetes/pull/78548.
This is going back to April/May, 2019.


----


I'll put another PR with the fix in instance type in the node config, just in case, but will ask around to figure out whether these benchmarks are good and we should fix them or not.

alejandrox1:
@spiffxp thank you for the review!
Going back to the drawing board with this.
Will start by getting a better feel for the e2e tests - the original author claims these should run on n1-standard-1 machines.
But will work with the rest of sig node to get this thing moving.

PR #17672, opened 2020-05-21
Stop publishing latest-green.txt and migrate to k8s-infra-prow-build

spiffxp:
This migrates ci-kubernetes-e2e-gci-gce over to the k8s-infra owned
prow build cluster. The release branch variants of this job have
already been moved over. What was preventing this from moving over
was the --publish flag, implying that this job needed write access
to a bucket that won't allow non-google.com access.

I have looked using cs.k8s.io and can't find any code that actually
relies on the existence of latest-green.txt. I could have sworn that
it was at one time used by tooling in kubernetes/release (eg: as part
of find_green_build), but I haven't found it in history there either.

I've looked in this repo's history, and publishing to this path has
been present since inception, but I still can't find what might actually
be looking for this file.

It's referenced in docs in a few places, which I will PR to remove if
those are the only references.

/cc @fejta @ixdy
maybe you remember the history of this file?

### DIFF
--- a/config/jobs/kubernetes/sig-cloud-provider/gcp/gcp-gce.yaml
+++ b/config/jobs/kubernetes/sig-cloud-provider/gcp/gcp-gce.yaml
@@ -301,6 +301,7 @@ presubmits:
 periodics:
 - interval: 30m
   name: ci-kubernetes-e2e-gci-gce
+  cluster: k8s-infra-prow-build
   labels:
     preset-service-account: "true"
     preset-k8s-ssh: "true"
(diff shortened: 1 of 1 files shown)


spiffxp:
/hold
for comment
/cc @justaugustus 
FYI

spiffxp:

dims:
/lgtm

Let's do this!

justaugustus:
For releng, we rely on `ci/latest.txt` or `ci/latest-x.y.txt`:

https://github.com/kubernetes/release/blob/3544be5a6436265019b1e85606ba6b8e02d5ed48/pkg/release/version.go#L58-L60

which is produced from the `ci-kubernetes-build` jobs.

There are [minimal references](https://cs.k8s.io/?q=latest-green&i=nope&files=&repos=) to latest-green, so I'm in favor of finally removing it.

/lgtm
/approve

Feel free to pull the hold at will.

ixdy:
Two places that come to mind where it may have been used at some point in the distant past:
- anago (though you mention you haven't been able to find this anywhere)
- kops e2e testing, when we wanted to test latest kops vs. a green kubernetes build (or vice-versa?)

ixdy:
The latest-green.txt file was added in 2015: https://github.com/kubernetes/kubernetes/pull/10428

Amusingly, there was a recent comment about it: https://github.com/kubernetes/kubernetes/issues/10410#issuecomment-597204599

ixdy:
Also, I may have been thinking about kops' use of https://storage.googleapis.com/kops-ci/bin/latest-ci-updown-green.txt.

spiffxp:
/hold cancel
Thanks for the additional history @ixdy.  I feel reasonably confident we can drop this.  Will open up PR's to address docs references to latest-green.txt

PR #17676, opened 2020-05-21
Don't mix approvers and reviewers in blunderbuss

### LINKED ISSUE #16793: prow should recommend relevant approvers

matthyx:
Relates to #16793
Avoid using `sets.String` when keeping the order matters.

### DIFF
--- a/prow/plugins/blunderbuss/blunderbuss.go
+++ b/prow/plugins/blunderbuss/blunderbuss.go
@@ -222,10 +222,16 @@ func handle(ghc githubClient, roc repoownersClient, log *logrus.Entry, reviewerC
 				if err != nil {
 					return err
 				}
+				var added int
 				combinedReviewers := sets.NewString(reviewers...)
-				combinedReviewers.Insert(approvers...)
-				log.Infof("Added %d approvers as reviewers. %d/%d reviewers found.", combinedReviewers.Len()-len(reviewers), combinedReviewers.Len(), *reviewerCount)
-				reviewers = combinedReviewers.List()
+				for _, approver := range approvers {
+					if !combinedReviewers.Has(approver) {
+						reviewers = append(reviewers, approver)
+						combinedReviewers.Insert(approver)
+						added++
+					}
+				}
+				log.Infof("Added %d approvers as reviewers. %d/%d reviewers found.", added, combinedReviewers.Len(), *reviewerCount)
 			}
 		}
 		if missing := *reviewerCount - len(reviewers); missing > 0 {
@@ -250,7 +256,9 @@ func handle(ghc githubClient, roc repoownersClient, log *logrus.Entry, reviewerC
 
 func getReviewers(rc reviewersClient, ghc githubClient, log *logrus.Entry, author string, files []github.PullRequestChange, minReviewers int, useStatusAvailability bool) ([]string, []string, error) {
 	authorSet := sets.NewString(github.NormLogin(author))
-	reviewers := sets.NewString()
+	stage1 := sets.NewString()
+	stage2 := sets.NewString()
+	stage3 := sets.NewString()
 	requiredReviewers := sets.NewString()
 	leafReviewers := sets.NewString()
 	busyReviewers := sets.NewString()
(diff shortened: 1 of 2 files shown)


prow/plugins/blunderbuss/blunderbuss.go
```diff
@@ -251,6 +257,7 @@ func handle(ghc githubClient, roc repoownersClient, log *logrus.Entry, reviewerC
 func getReviewers(rc reviewersClient, ghc githubClient, log *logrus.Entry, author string, files []github.PullRequestChange, minReviewers int, useStatusAvailability bool) ([]string, []string, error) {
 	authorSet := sets.NewString(github.NormLogin(author))
 	reviewers := sets.NewString()
+	reviewersList := []string{}
```
fejta, line 260: `var reviewersList []string` (reviewersList = append(reviewersList, "foo") works just fine)

On the other hand I don't think we need any of the changes in `getReviewers()` -- it is just a clone of reviewers

prow/plugins/blunderbuss/blunderbuss.go
```diff
@@ -222,10 +222,16 @@ func handle(ghc githubClient, roc repoownersClient, log *logrus.Entry, reviewerC
 				if err != nil {
 					return err
 				}
+				added := 0
```
fejta, line 225: `var added int` since you're not initializing anything

https://dave.cheney.net/practical-go/presentations/qcon-china.html#_use_a_consistent_declaration_style

prow/plugins/blunderbuss/blunderbuss.go
```diff
@@ -289,11 +302,14 @@ func getReviewers(rc reviewersClient, ghc githubClient, log *logrus.Entry, autho
...
-				reviewers.Insert(r)
+				if !reviewers.Has(r) {
+					reviewersList = append(reviewersList, r)
+					reviewers.Insert(r)
+				}
 			}
 		}
 	}
-	return reviewers.List(), requiredReviewers.List(), nil
+	return reviewersList, requiredReviewers.List(), nil
```
fejta, line 312: This seems like an unnecessary (and worse) change from the existing code

matthyx: Sorry Erick but this is actually what makes this fix working... because the various loops fill reviewers in a specific order, from the leaf first up to the root.
By using the `sets.String` we are losing this information and then when we truncate the list to `maxReviewerCount` Ben goes first...

matthyx: I have decided to keep the `reviewers` set, because it is used in all the logic with unions and co.

fejta: Why do we depend on the order? This seems very wrong.

matthyx: Because the list of reviewers is constructed this way to give priority to leaf approvers when there's a need to shorten the list.
You can revert that line and see how the new test fails in suggesting Ben...

spiffxp:
/hold
Please drop the "fixes" from the description, this doesn't actually fix #16793 (ref: https://github.com/kubernetes/test-infra/issues/16793#issuecomment-631231564)

prow/plugins/blunderbuss/blunderbuss.go
```diff
@@ -272,14 +279,20 @@ func getReviewers(rc reviewersClient, ghc githubClient, log *logrus.Entry, autho
...
 		leafReviewers = leafReviewers.Union(fileUnusedLeafs)
 		if r := findReviewer(ghc, log, useStatusAvailability, &busyReviewers, &fileUnusedLeafs); r != "" {
-			reviewers.Insert(r)
+			if !reviewers.Has(r) {
+				reviewersList = append(reviewersList, r)
+				reviewers.Insert(r)
+			}
 		}
 	}
 	// now ensure that we request review from at least minReviewers reviewers. Favor leaf reviewers.
```
matthyx, line 288: @fejta see here it even says it will favor leaf reviewers...

prow/plugins/blunderbuss/blunderbuss.go
```diff
@@ -289,11 +302,14 @@ func getReviewers(rc reviewersClient, ghc githubClient, log *logrus.Entry, autho
 		fileReviewers := rc.Reviewers(file.Filename).Difference(authorSet)
```
matthyx, line 302: And the break just above ensures we only add non-leaf reviewers when we're missing some...

prow/plugins/blunderbuss/blunderbuss.go
matthyx: @fejta how would you do it?

matthyx:
@fejta here you go

prow/plugins/blunderbuss/blunderbuss.go
fejta: Ordering the string seems like it will prefer Ben to Erick, when we want each person in a given owners file assigned at roughly the same frequency.

I would use a stack of sets, one per level.

AKA if the file is:

`alpha/beta/omega/something/OWNERS`

and there are 

`alpha/OWNERS`, `alpha/beta/OWNERS` and `alpha/beta/omega/something/OWNERS`

I would represent this as: `[]sets.String`

I would choose someone by something along the lines of:

```
for i := range options {
  all := options[len(options)-i] // start from leaf/OWNERS not /OWNERS
  rand.Shuffle(all, func(i,j int) { all[j], all[i] = all[i], all[j] })
  for _, choice := range all {
    if chosen.Contains(choice) {
      continue
    }
    if enoughChoices(chosen) { break/return/whatever }
  }
  log.Warn("too few choices, oh well")
}
```

prow/plugins/blunderbuss/blunderbuss.go
fejta: Yes we should favor leaf reviewers, but we shouldn't prioritize inside the leafs reviewers

prow/plugins/blunderbuss/blunderbuss.go
```diff
@@ -222,10 +222,16 @@ func handle(ghc githubClient, roc repoownersClient, log *logrus.Entry, reviewerC
...
 					return err
 				}
+				var added int
 				combinedReviewers := sets.NewString(reviewers...)
-				combinedReviewers.Insert(approvers...)
-				log.Infof("Added %d approvers as reviewers. %d/%d reviewers found.", combinedReviewers.Len()-len(reviewers), combinedReviewers.Len(), *reviewerCount)
-				reviewers = combinedReviewers.List()
+				for _, approver := range approvers {
+					if !combinedReviewers.Has(approver) {
+						reviewers = append(reviewers, approver)
```
fejta, line 229: Given that the code below does `reviewers = reviewers[:maxReviewers]`  we cannot simply range through approvers and append to this list. There is no randomization happening here, and we need that.

prow/plugins/blunderbuss/blunderbuss.go
matthyx: Ok, thanks for having taken the time to look deeply into it. I will implement that during the rainy weekend...

prow/plugins/blunderbuss/blunderbuss.go
matthyx: indeed, and avoid mixing them when doing so

matthyx: now getReviewers returns a randomized list of approvers/reviewers, but keeping the leaf preference

prow/plugins/blunderbuss/blunderbuss.go
matthyx: randomization is now done just before return

PR #17681, opened 2020-05-21
Document how to upgrade images

fejta:


fejta:
Can I get a review please?

PR #17682, opened 2020-05-21
Update test-infra jobs to use the test-infra variant of the bazel image

fejta:


fejta:
/hold cancel

thanks!

PR #17687, opened 2020-05-22
Modified node-kubelet-flaky job and image spec

alejandrox1:
This patch aims to enable the performance benchamrks in the
node-kubelet-flaky e2e job to run on an instance with enough cpu and
memory to run performance workloads.
It also explicitly specifies the `--server-start-timeout` flag to tell
the kubelet how long to wait for healthchecks in between restarts.

xref: https://github.com/kubernetes/kubernetes/issues/91263

Signed-off-by: alejandrox1 <alarcj137@gmail.com>

/sig node
/priority important-longterm
/kind failing-test


alejandrox1:
/assign @yujuhong @Random-Liu @dchen1107
ptal

jobs/e2e_node/perf-config.yaml
```diff
@@ -0,0 +1,9 @@
+---
+images:
```
bsdnet, line 2: Can we name this file as 
perf-image-config.yaml? Which is more explicit.

alejandrox1: yes! will rename.

jobs/e2e_node/perf-config.yaml
```diff
@@ -0,0 +1,9 @@
...
+  cosbeta-resource1:
```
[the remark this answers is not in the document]
alejandrox1: and sure thing.

jobs/e2e_node/perf-config.yaml
```diff
@@ -0,0 +1,9 @@
...
+    image: cos-beta-81-12871-44-0
+    project: cos-cloud
+    machine: n1-standard-16
```
MHBauer, line 6: can see from https://github.com/kubernetes/kubernetes/pull/65251 that it is supposed to be running on this machine type.

config/jobs/kubernetes/sig-node/node-kubelet.yaml
```diff
@@ -177,8 +177,8 @@ periodics:
       - --deployment=node
       - --gcp-project-type=node-e2e-project
       - --gcp-zone=us-west1-b
-      - --node-args=--image-config-file=/workspace/test-infra/jobs/e2e_node/image-config.yaml
-      - --node-test-args= --kubelet-flags="--cgroups-per-qos=true --cgroup-root=/"
+      - --node-args=--image-config-file=/workspace/test-infra/jobs/e2e_node/perf-config.yaml
```
MHBauer, line 180: it's better to use a separate config file as well, because I have no idea why we would need to run 3 copies of this on different operating systems.

jobs/e2e_node/perf-config.yaml
MHBauer: I agree it's more explicit, but there's precedent for `bleh-config` such as `benchmark-config.yaml`.

I think it would be great to have `image` in the name.

jobs/e2e_node/perf-config.yaml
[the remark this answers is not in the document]
alejandrox1: Done

jobs/e2e_node/perf-config.yaml
alejandrox1: Done.

alejandrox1:
Thank you for the reviews @bart0sh @bsdnet @MHBauer 
patched the pr based on your comments.

config/jobs/kubernetes/sig-node/node-kubelet.yaml
```diff
@@ -177,8 +177,8 @@ periodics:
       - --deployment=node
       - --gcp-project-type=node-e2e-project
       - --gcp-zone=us-west1-b
-      - --node-args=--image-config-file=/workspace/test-infra/jobs/e2e_node/image-config.yaml
-      - --node-test-args= --kubelet-flags="--cgroups-per-qos=true --cgroup-root=/"
+      - --node-args=--image-config-file=/workspace/test-infra/jobs/e2e_node/perf-cos-81-12871-103-0-config.yaml
```
MHBauer, line 180: sorry, I'll be more explicit, I think the filename should be `perf-image-config.yaml`

alejandrox1: haha :sweat_smile: fixed it!

alejandrox1:
> I don't think we ever fully resolved the thoughts from [kubernetes/kubernetes#91263 (comment)](https://github.com/kubernetes/kubernetes/issues/91263#issuecomment-632180567) & [#17669 (comment)](https://github.com/kubernetes/test-infra/pull/17669#issuecomment-632179553)
> 
> I have an action item to write an email to the mailing list and raise the issue elsewhere to see who might want to own these.

From what I recall, there was at least 1 person who was looking at the performance jobs.
It may be worth it to make things more official and document somewhere who is doing what.
Otherwise we may run into the same issue as with "who takes care of all the cos images".

jobs/e2e_node/perf-image-config.yaml
```diff
@@ -0,0 +1,9 @@
+---
+images:
+  cosbeta-resource1:
```
spiffxp, line 3: For parity with image-config.yaml and image-config-serial.yaml
```suggestion
  cos-stable1:
```

jobs/e2e_node/perf-image-config.yaml
```diff
@@ -0,0 +1,9 @@
+---
+images:
```
spiffxp, line 2: the existing job also runs for ubuntu, can we include that here?

jobs/e2e_node/perf-image-config.yaml
```diff
@@ -0,0 +1,9 @@
...
+  cosbeta-resource1:
+    image: cos-81-12871-103-0
+    project: cos-cloud
+    machine: n1-standard-16
+    metadata: "user-data<test/e2e_node/jenkins/gci-init.yaml,gci-update-strategy=update_disabled"
+    tests:
+      - 'Node Performance Testing'
```
spiffxp, line 9: (this is me not fully understanding how all of this works)

Does this take precedence over the `[Flaky]` regex in the job? Work in combination with it?

It feels like we're overloading/misdirecting by using the term "flaky" when really this job seems to be running perf tests

jobs/e2e_node/perf-image-config.yaml
alejandrox1: Fixed

jobs/e2e_node/perf-image-config.yaml
alejandrox1: that is a good point... my decision making for this was purely based on sig node conversations about which images to run tests in and the cos ones are a definite yes.
Everything else that isn't cos seems to be up in the air at this moment (?).
During a previous meeting there was also a mention of organizing things such that each job only ran 1 image.

Right now, kubelet-flaky is running tests on this ubuntu image as well
https://github.com/kubernetes/test-infra/blob/dbd9fce65bd40199e01f6f5ee2e34ca51fa298b6/jobs/e2e_node/image-config.yaml#L6
so asking for feedback here: should this version of kubelet-flaky run ubuntu as well? and if so, should it run the same ubuntu image that it has been using until now or another?

/cc @spiffxp @MHBauer @bsdnet @vpickard

jobs/e2e_node/perf-image-config.yaml
alejandrox1: it works in combination with the focus on `[Flaky]`.

And it is overloading and misdirecting.
The long version of the story is in 
https://github.com/kubernetes/kubernetes/issues/91263
but the short version is that these tests were meant to run in an n1-standard-16 sized machine, see https://github.com/kubernetes/kubernetes/pull/65250.

But for some reason the e2e tests running in this current version of the job were never properly modified to run in that machine and have been running since the beginning in an n1-standard-1.
As such, the have always been failing.
Then they were labeled as "flaky" in https://github.com/kubernetes/kubernetes/pull/78548 with the intent of removing them from the kubelet-serial job.
The idea was to move them to their own e2e job so they didn't pollute another job.

My idea here was to explicitly specify within the image config that this whole fix is only for a handful of performance tests that were labeled as flaky becuase they were running in the wrong machine.
I think a better place for the tests in this job would be in https://github.com/kubernetes/test-infra/blob/master/jobs/e2e_node/benchmark-config.yaml but i guess one step at a time (?)

jobs/e2e_node/perf-image-config.yaml
bsdnet: @alejandrox1 Let's also update the Ubuntu since those are two images running. After we got them green, per the meeting, we may add more.
I will start a topic in sig-node to discuss how to support multiple images in a scalable way.

MHBauer: I don't know who monitors the output of this (clearly nobody), nor do I know if it surfaces in a dashboard.

jobs/e2e_node/perf-image-config.yaml
MHBauer: It happens to be the only test marked flaky left in node_e2e.

I think the idea of both `Flaky` regex and `'Node Performance Testing'` is for this to be moved out of flaky and into it's own job eventually. (It should probably be it's own job already, because nothing else is backed by a `n1-standard-16`.

spiffxp: ok, I won't block on this one as long as there's followup to either rename this job to something like `ci-kubernetes-node-kubelet-perf`, or split out a new job that does so... I think the node perf tests shouldn't be tagged as `[Flaky]` but instead something like `[Feature:Node Performance]`

we currently have no way of verifying/linking whether all e2e tests are executed by at least one CI job, so I could understand if someone wants to preemptively keep a flaky job around even if there are no flaky tests

jobs/e2e_node/perf-image-config.yaml
alejandrox1: Sweet, so im looking at all the ubuntu images for the ubuntu-os-gke-cloud project
```
gcloud compute images list --project ubuntu-os-gke-cloud
```

and `ubuntu-gke-1804-1-17-v20200605` seems to be the latest one.

As per monitoring this later on, we'll figure it out :sweat_smile:

jobs/e2e_node/perf-image-config.yaml
alejandrox1: > ok, I won't block on this one as long as there's followup to either rename this job to something like ci-kubernetes-node-kubelet-perf, or split out a new job that does so... I think the node perf tests shouldn't be tagged as [Flaky] but instead something like [Feature:Node Performance]

YES to all of this! probably split out a new job for this (or move the contents of the image config into the jobs running cenchmarks TBD with SIG Node).
And huge +1 on `[Feature:Node Performance]`

alejandrox1:
Thank you everyone for your comments!
I added an ubuntu image to run these tests on.

jobs/e2e_node/perf-image-config.yaml
```diff
@@ -0,0 +1,15 @@
+---
+images:
+  cos-stable1:
+    image: cos-81-12871-103-0
```
bsdnet, line 4: nit: cos-81-12871-119-0

MHBauer:
looks like tensorflow works okay, integersort works okay, and then embarrassingly parallel workload blows up.

PR #17693, opened 2020-05-23
Kubeflow Pipelines - Added python presubmits

Ark-kun:
These tests will be used to migrate off Travis


Bobgy:
Thanks @Ark-kun!
We should get `run_if_changed` configured for other unit tests too.

PR #17695, opened 2020-05-23
Update containerd related tests using latest kubekins image

bsdnet:

What happened:

The following tests failed continuously in the past:
https://testgrid.k8s.io/sig-node-containerd#containerd-e2e-gci-1.2
https://testgrid.k8s.io/sig-node-containerd#containerd-node-e2e-1.2
https://testgrid.k8s.io/sig-node-containerd#containerd-node-e2e-features-1.2

What you expected to happen:
They should succeed.

How to reproduce it (as minimally and precisely as possible):
See https://testgrid.k8s.io/sig-node-containerd

Please provide links to example occurrences, if any:
See https://testgrid.k8s.io/sig-node-containerd

Anything else we need to know?:
From the log of https://testgrid.k8s.io/sig-node-containerd#containerd-e2e-gci-1.2, it mentioned that failing to get k8s image 1.3, which is removed from https://github.com/kubernetes/test-infra/blob/master/images/kubekins-e2e/variants.yaml. Also, kubekins container is out of date.

Ideally, we should use the latest stable kubekins image. Also, the reason bumping to 1.15 is that it is using the same golang version with latest containered.

### DIFF
--- a/config/jobs/kubernetes/sig-node/containerd.yaml
+++ b/config/jobs/kubernetes/sig-node/containerd.yaml
@@ -143,7 +143,7 @@ periodics:
       - --env=KUBE_MASTER_EXTRA_METADATA=user-data=/workspace/github.com/containerd/cri/test/e2e/master.yaml,containerd-configure-sh=/workspace/github.com/containerd/cri/cluster/gce/configure.sh,containerd-env=/workspace/test-infra/jobs/e2e_node/containerd/containerd-release-1.2/env
       - --env=KUBE_NODE_EXTRA_METADATA=user-data=/workspace/github.com/containerd/cri/test/e2e/node.yaml,containerd-configure-sh=/workspace/github.com/containerd/cri/cluster/gce/configure.sh,containerd-env=/workspace/test-infra/jobs/e2e_node/containerd/containerd-release-1.2/env
       - --env=ENABLE_POD_SECURITY_POLICY=true
-      - --extract=ci/latest-1.13
+      - --extract=ci/latest-1.15
       - --gcp-node-image=gci
       - --gcp-nodes=4
       - --gcp-zone=us-west1-b
@@ -151,7 +151,7 @@ periodics:
       - --provider=gce
       - --test_args=--ginkgo.skip=\[Slow\]|\[Serial\]|\[Disruptive\]|\[Flaky\]|\[Feature:.+\] --minStartupPods=8
       - --timeout=50m
-      image: gcr.io/k8s-testimages/kubekins-e2e:v20191029-3656cc1-1.13
+      image: gcr.io/k8s-testimages/kubekins-e2e:v20200521-100609e-1.15
   annotations:
     testgrid-dashboards: sig-node-containerd
     testgrid-tab-name: containerd-e2e-gci-1.2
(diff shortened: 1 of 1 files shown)


MHBauer:
should these be updated as part of https://github.com/kubernetes/test-infra/pull/17715 ? how does one get into that automated flow?

bsdnet:
> should these be updated as part of #17715 ? how does one get into that automated flow?

That is a good question. I am not sure how other images are updated automatically. From the code reading, this image is special, and it use given verison golang and kubernetes. Recently, 1.3 was removed
from https://github.com/kubernetes/test-infra/blame/master/images/kubekins-e2e/variants.yaml

spiffxp:
>how does one get into that automated flow?

Use an image that gets built and pushed automatically.  In this case, use a version of kubekins that lines up with a version of kubernetes that is still built and supported.  

According the [current documented policy](https://kubernetes.io/docs/setup/release/version-skew-policy/#supported-versions), 1.15 is no longer in support.  You'll want to move to 1.16 to keep up.  In practice, the aged-out release jobs are kept around for a few weeks to cut a final patch release.  Previous release schedules have included a date for when the old jobs would be removed, but I don't see one on the [current release schedule](https://github.com/kubernetes/sig-release/tree/master/releases/release-1.19#timeline) (@kubernetes/patch-release-team @onlydole ?)

Longer term, jobs that need a per-release-branch variant should use [`config-forker` annotations](https://github.com/kubernetes/test-infra/tree/master/releng/config-forker#supported-annotations) to have a [job generated when the release-branch is cut](https://github.com/kubernetes/test-infra/tree/master/config/jobs#release-branch-jobs)

The code that does the bumping is here, fwiw: https://github.com/kubernetes/test-infra/tree/master/experiment/autobumper

spiffxp:
/approve
All the above said, I'll merge this to unblock, as it's not making things any worse.  Please address in followup PR's.

onlydole:
Thank you for the heads up, @spiffxp! It looks like the 1.15 removal dates fell off our schedule with our revisions and we’ll get them re-added 👍🏼

bsdnet:
Thanks @spiffxp  for your detailed explanation.

containered 1.2.x(https://github.com/containerd/containerd/releases/tag/v1.2.13)  is officially built with golang 1.12.17 . That is why I bump to 1.15, which is the only version using the same golang per
https://github.com/kubernetes/test-infra/blob/master/images/kubekins-e2e/variants.yaml

I will follow up to move to 1.16, and it will be also nice for auto bumper to do that(remove from variant, bump the version.) I will take a look at that too.

onlydole:
@spiffxp - to follow up on the 1.15 job removal, I synced up with @justaugustus and SIG Release purposely removed the "1.15 job removal" entry from the schedule. The 1.15 jobs don't exactly correlate to any 1.19 workstreams, and previously, the beta.0 branch was a good time to perform that removal.

Now that the branch cut is on RC.0 AND we have an extended cycle, it's a little too late to have that branch around.

TL;DR - @justaugustus will be removing the 1.15 branch jobs as there are some expectations that the config-forker tooling has, so he will need to verify it still works when someone removes branch jobs "earlier" than the tool expects.

Please let me know if that makes sense, or if there are any other questions on that front, and we can address them!

PR #17699, opened 2020-05-25
Make svcat e2e test required on pipeline

mszostok:
**Description**

In the first phase, the e2e tests were optional but e2e test proved that it is stable and should be required before merging any functionality to master


MHBauer:
I wonder if we can hook these test results up to testgrid?

/lgtm

Issue #17636, opened 2020-05-19
Prow: deck shouldn't report error message when job history symlink doesn't exist


chaodaiG:
**What happened**:
Deck logs errors when it couldn't read job history symlink for a job

**What you expected to happen**:
Deck should differentiate the reason why it failed reading job history symlink.
- If file not exist, just log as info/debug
- If any other reason, log as error

**How to reproduce it (as minimally and precisely as possible)**:
Go to https://prow.knative.dev/job-history/knative-prow/pr-logs/directory/pull-knative-client-integration-tests, and deck would log a bunch of errors as some of the symlink files don't exist

**Please provide links to example occurrences, if any**:
See https://github.com/knative/test-infra/issues/2089

**Anything else we need to know?**:

alvaroaleman:
/reopen
My fixes do not seem to work

chaodaiG:
> /reopen
> My fixes do not seem to work

Curious how you tested your changes

alvaroaleman:
By rolling it out :)

chaodaiG:
Could find the PR updating k8s prow, assume it was tested on another prow instance?

alvaroaleman:
Yeah, the openshift one: https://github.com/openshift/release/pulls?q=is%3Apr+Update+prow+and+other+images+as+necessary

We usually bump it at least daily

chaodaiG:
Glad to know. Seems like knative prow isn't the only instance suffering from this issue :)

stevekuznetsov:
Deck log spam is a feature, not a bug!

chaodaiG:
> Deck log spam is a feature, not a bug!

Agree that it's not a bug. It would be totally fine if it logs as `debug` or `info` instead of `error`, as this makes us almost have to ignore all errors from deck

alvaroaleman:
He was trolling. It is a bug. Its not an error in decks code or configuration, jobs sometimes not getting to the point of uploading logs or failing doing so will always happen sometimes.

PR #17702, opened 2020-05-25
io/opener.IsNotExist: Support wrapped errors

alvaroaleman:
Needed to actually fix https://github.com/kubernetes/test-infra/issues/17636

### DIFF
--- a/prow/io/opener.go
+++ b/prow/io/opener.go
@@ -137,13 +137,16 @@ var ErrNotFoundTest = fmt.Errorf("not found error which should only be used in t
 
 // IsNotExist will return true if the error shows that the object does not exist.
 func IsNotExist(err error) bool {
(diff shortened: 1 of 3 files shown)


prow/io/opener.go
```diff
@@ -137,15 +137,18 @@ var ErrNotFoundTest = fmt.Errorf("not found error which should only be used in t
...
+	if errors.Is(err, ErrNotFoundTest) {
 		return true
 	}
 	if errors.Is(err, os.ErrNotExist) {
 		return true
 	}
+	if errors.Is(err, storage.ErrObjectNotExist) {
+		return true
+	}
 	return gcerrors.Code(err) == gcerrors.NotFound
```
alvaroaleman, line 152: Its not possible to test this because the error type that is used in the check is in an `internal` package and there is no constructor for it

matthyx:
ok...
/lgtm
/hold

PR #17703, opened 2020-05-25
Move filteringProwJobLister into JobAgent

alvaroaleman:
There is no reason to ever use the JobAgent without the filtering
ProwJobLister and its the only thing that uses it.

This change is needed so external consumers can use the JobAgent.

/assign @stevekuznetsov


prow/deck/jobs/jobs.go
```diff
@@ -80,15 +84,75 @@ type PodLogClient interface {
...
+	showHidden  bool
+}
+
+func (c *filteringProwJobLister) ListProwJobs(selector string) ([]prowapi.ProwJob, error) {
+	prowJobList := &prowapi.ProwJobList{}
+	parsedSelector, err := labels.Parse(selector)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse selector: %v", err)
+	}
+	listOpts := &ctrlruntimeclient.ListOptions{LabelSelector: parsedSelector, Namespace: c.cfg().ProwJobNamespace}
```
alvaroaleman, line 122: Including the namespace here is the only actual change made by this PR. Everything else is just moving code around (Git can unfortunately not show that unless its the whole file).

PR #17707, opened 2020-05-26
Fix prow.k8s.io/build-id Pod label

### LINKED ISSUE #17612: prow.k8s.io/build-id Pod label is always the empty string.
**What happened**:
It looks like the `prow.k8s.io/build-id` label is always set to the empty string on ProwJob pods.
```
$ kubectl get po -n test-pods -oyaml | grep "prow.k8s.io/build-id" | sort | uniq
      prow.k8s.io/build-id: ""
```
Pods should be properly labelled with their build IDs.

/area prow
/area prow/plank

adshmh:
Closes #17612

### DIFF
--- a/prow/plank/controller.go
+++ b/prow/plank/controller.go
@@ -321,8 +321,7 @@ func (c *Controller) syncPendingJob(pj prowapi.ProwJob, pm map[string]corev1.Pod
 		c.incrementNumPendingJobs(pj.Spec.Job)
 		// Pod is missing. This can happen in case the previous pod was deleted manually or by
 		// a rescheduler. Start a new pod.
-		id, pn, err := c.startPod(pj)
-		if err != nil {
+		if err := c.startPod(&pj); err != nil {
 			if !isRequestError(err) {
 				return fmt.Errorf("error starting pod %s: %v", pod.Name, err)
 			}
@@ -331,8 +330,6 @@ func (c *Controller) syncPendingJob(pj prowapi.ProwJob, pm map[string]corev1.Pod
 			pj.Status.Description = fmt.Sprintf("Job cannot be started: %v", err)
 			c.log.WithFields(pjutil.ProwJobFields(&pj)).WithError(err).Warning("Request error starting pod.")
 		} else {
-			pj.Status.BuildID = id
-			pj.Status.PodName = pn
 			c.log.WithFields(pjutil.ProwJobFields(&pj)).Info("Pod is missing, starting a new pod")
 		}
 	} else {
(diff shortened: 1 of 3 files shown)


prow/pod-utils/decorate/podspec.go
```diff
@@ -152,18 +152,19 @@ func LabelsAndAnnotationsForJob(pj prowapi.ProwJob) (map[string]string, map[stri
 }
 
 // ProwJobToPod converts a ProwJob to a Pod that will run the tests.
-func ProwJobToPod(pj prowapi.ProwJob, buildID string) (*coreapi.Pod, error) {
-	return ProwJobToPodLocal(pj, buildID, "")
+func ProwJobToPod(pj prowapi.ProwJob) (*coreapi.Pod, error) {
```
cjwagner, line 155: Why was this function signature changed? The build ID should always be available when creating a pod for a ProwJob and including it in the function signature makes the requirement explicit.

prow/plank/controller.go
```diff
@@ -524,41 +516,39 @@ func (c *Controller) syncTriggeredJob(pj prowapi.ProwJob, pm map[string]corev1.P
...
 func (c *Controller) getBuildID(name string) (string, error) {
 	return pjutil.GetBuildID(name, c.totURL)
 }
 
 func getPodBuildID(pod *corev1.Pod) string {
-	for _, env := range pod.Spec.Containers[0].Env {
-		if env.Name == "BUILD_ID" {
-			return env.Value
-		}
+	if buildID, ok := pod.ObjectMeta.Labels[kube.ProwBuildIDLabel]; ok {
```
cjwagner, line 550: Since we've been failing to set this label, this won't work for existing jobs. We should ensure backwards compatibility.

prow/pod-utils/decorate/podspec.go
adshmh: Thank you for the review. Fixed.

prow/plank/controller.go
adshmh: Thank you for the review. Fixed and added unit tests to cover.

prow/plank/controller.go
```diff
@@ -524,42 +516,49 @@ func (c *Controller) syncTriggeredJob(pj prowapi.ProwJob, pm map[string]corev1.P
...
+	pj.Status.PodName = pod.ObjectMeta.Name
+	return nil
 }
 
 func (c *Controller) getBuildID(name string) (string, error) {
 	return pjutil.GetBuildID(name, c.totURL)
 }
 
 func getPodBuildID(pod *corev1.Pod) string {
+	if buildID, ok := pod.ObjectMeta.Labels[kube.ProwBuildIDLabel]; ok && buildID != "" {
```
cjwagner, line 551: super nit: This is slightly more idiomatic:
```go
if buildID := pod.ObjectMeta.Labels[kube.ProwBuildIDLabel]; buildID != "" {
```

adshmh: Thank you for the review. Tide merged the PR before I could correct it.

cjwagner: Yeah, I didn't want to block the PR on this since it is such a minor style change. It's just a small tip for the future.

PR #17709, opened 2020-05-26
Roll COS images to the latest available versions

bart0sh:
Roll COS images to the latest available versions
    
COS images are outdated in the benchmark configs.
    
This causes e2e node kubelet tests failures like this:
        Error 400: Invalid value for field 'resource.disks[0].initializeParams.sourceImage':
        'projects/cos-cloud/global/images/cos-stable-73-11647-510-0'.
        The referenced image resource cannot be found., invalid
    
Rolled cos-stable-73 -> cos-77 and cos-stable-77 -> cos-stable-81 to
get rid of non-existent images and to use latest available image versions.

### DIFF
--- a/jobs/e2e_node/benchmark-config.yaml
+++ b/jobs/e2e_node/benchmark-config.yaml
@@ -1,21 +1,21 @@
 ---
 images:
   cosstable2-resource1:
-    image: cos-stable-73-11647-510-0
+    image: cos-77-12371-251-0
     project: cos-cloud
     machine: n1-standard-1
     metadata: "user-data<test/e2e_node/jenkins/gci-init.yaml,gci-update-strategy=update_disabled"
     tests:
       - 'resource tracking for 0 pods per node \[Benchmark\]'
   cosstable2-resource2:
-    image: cos-stable-73-11647-510-0
+    image: cos-77-12371-251-0
     project: cos-cloud
     machine: n1-standard-1
     metadata: "user-data<test/e2e_node/jenkins/gci-init.yaml,gci-update-strategy=update_disabled"
     tests:
       - 'resource tracking for 35 pods per node \[Benchmark\]'
   cosstable2-resource3:
-    image: cos-stable-73-11647-510-0
+    image: cos-77-12371-251-0
     project: cos-cloud
     machine: n1-standard-1
     metadata: "user-data<test/e2e_node/jenkins/gci-init.yaml,gci-update-strategy=update_disabled"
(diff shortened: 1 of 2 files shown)


alejandrox1:
Assigning a couple node approvers
/assign @dchen1107 @yguo0905 @yujuhong @derekwaynecarr

jobs/e2e_node/benchmark-config.yaml
```diff
@@ -1,74 +1,74 @@
...
-    image: cos-stable-73-11647-510-0
+    image: cos-77-12371-251-0
     project: cos-cloud
     machine: n1-standard-1
     tests:
       - 'create 105 pods with 100ms interval \[Benchmark\]'
 
   cosstable1-resource1:
-    image: cos-stable-77-12371-227-0 
+    image: cos-stable-81-12871-103-0
```
bsdnet, line 57: Can we use cos-81-12871-103-0? 81 is the latest LTS.

gcloud compute images list |grep cos-cloud
cos-69-10895-385-0                                    cos-cloud            cos-69-lts                                    READY
cos-73-11647-534-0                                    cos-cloud            cos-73-lts                                    READY
cos-77-12371-273-0                                    cos-cloud            cos-77-lts                                    READY
cos-81-12871-103-0                                    cos-cloud            cos-81-lts                                    READY
cos-beta-81-12871-44-0                                cos-cloud            cos-beta                                      READY
cos-dev-84-13078-0-0                                  cos-cloud            cos-dev                                       READY
cos-stable-81-12871-103-0                             cos-cloud            cos-stable                                    READY

MHBauer: hi roy, 
would it be better to use `image_family: cos-stable`, would that consistently be the most up to date stable release?
and I'm not sure why we do 2 cos releases, but should the 77 one switch to `image_family: cos-77-lts`? I see that used in other places.

bart0sh: @bsdnet cos-81 is less expressive than cos-stable-81 in my opinion. there is at least cos-beta-81 from the same milestone.

bart0sh: @MHBauer Using image_family cos-77-lts, cos-stable, cos-beta or cos-dev is what I'm planning to do as a next step. This PR is a bugfix. Switching to image_family is an enhancement. I'd prefer not to mix them in one PR.

MHBauer: yea, not suggesting a change here, trying to get an understanding from an expert while they may notice.

bsdnet: @MHBauer  Yes, we should use cos-77-lts, cos-81-lts, and if the image_family prove working, that will reduce the frequency to update it.

@bart0sh cos-stable-81 is more expressive, but it will be short lived. For Continuous Integration, we can use it. For pre/post submit. We should use latest LTS.  See
https://cloud.google.com/container-optimized-os/docs/resources/support-policy

Issue #17714, opened 2020-05-26
node e2e tests - projects out of quota


karan:
Many sig-node tests are failing consistently due to CPU exhaustion:

```
I0526 15:56:55.801] unable to create gce instance with running docker daemon for image ubuntu-gke-1804-1-16-v20200330.  could not create instance tmp-node-e2e-076c4ac7-ubuntu-gke-1804-1-16-v20200330: [&{Code:QUOTA_EXCEEDED Location: Message:Quota 'CPUS' exceeded.  Limit: 100.0 in region us-west1. ForceSendFields:[] NullFields:[]}]
```

They are tracked in https://docs.google.com/spreadsheets/d/1mEU8B2_PmMwwgp-_xnyp7QYMBwcLoA9NNlHwDyMvO0Y/edit#gid=0.

For one of the projects, looks like it's at 100 instances already:

```
$ gcloud compute instances list --project cri-containerd-node-e2e | wc -l
100
```

A good starting point will be to:

- get a list of all these projects
- inspect all tests using these projects and how many CPUs they need
- come up with a plan for getting out of this problem (we should definitely increase quota and/or shard across more projects and/or reduce resource usage by tests)

MHBauer:
Also addresses exhaustion.

karan:
Looking at the following:

Job | Link | Project
--- | --- | ---
containerd-node-e2e-1.2 | https://testgrid.k8s.io/sig-node-containerd#containerd-node-e2e-1.2 | cri-containerd-node-e2e
containerd-node-e2e-1.3 | https://testgrid.k8s.io/sig-node-containerd#containerd-node-e2e-1.3 | cri-containerd-node-e2e
containerd-node-features | https://testgrid.k8s.io/sig-node-containerd#containerd-node-features | cri-containerd-node-e2e
containerd-node-e2e-features-1.2 | https://testgrid.k8s.io/sig-node-containerd#containerd-node-e2e-features-1.2 | cri-containerd-node-e2e
containerd-node-e2e-features-1.3 | https://testgrid.k8s.io/sig-node-containerd#containerd-node-e2e-features-1.3 | cri-containerd-node-e2e
image-validation-node-e2e | https://testgrid.k8s.io/sig-node-containerd#image-validation-node-e2e | cri-containerd-node-e2e
image-validation-node-features | https://testgrid.k8s.io/sig-node-containerd#image-validation-node-features | cri-containerd-node-e2e
[...]

PR #17717, opened 2020-05-26
Run some containerd tests in us-central1-b

karan:
For #17714

The project has a quota of 100 CPU in us-central1 and us-west1 but all these seem to run in the latter. This change shards the tests across multiple regions to better use allotted quota.

/sig node
/sig testing


MHBauer:
If those tests start running it would be great to be able to experiment with some of the configs without using a release+merge blocking test.

PR #17724, opened 2020-05-27
Update boskos pool dashboard to support same type, different instance

spiffxp:
I would like to have pools named identically across prow-builds' boksos
and k8s-infra's boskos, so there's one less thing to change when
migrating a prowjob from one build cluster to the other.

k8s-infra-gce-project vs. gce-project was a mistake I'd like to undo

I also want to be able to display a panel per-instance for the same pool
type. This is the dumb hardcoded approach


spiffxp:
I added 5 projects to the k8s-infra boskos instance ref: https://github.com/kubernetes/k8s.io/pull/898

spiffxp:
>making the instance a template variable to avoid hardcoding 

I agree, but I'd like the "all" option to enumerate dashboards, instead of aggregate.  I think there's value in seeing one pool go down and the other go up side-by-side.  I'll work on getting a setup where I can dev/test this locally.  And I'll need to migrate k8s-infra-gce-project into gce-project.

PR #17725, opened 2020-05-27
Migrate 100 node scalability release-blocking job to k8s-infra-prow-build

spiffxp:
Demonstrate use of the scalability-project pool added to k8s-infra's boskos instance added via https://github.com/kubernetes/k8s.io/pull/898

### DIFF
--- a/config/jobs/kubernetes/sig-scalability/sig-scalability-release-blocking-jobs.yaml
+++ b/config/jobs/kubernetes/sig-scalability/sig-scalability-release-blocking-jobs.yaml
@@ -101,6 +101,7 @@ periodics:
           memory: "16Gi"
 
 - interval: 30m
+  cluster: k8s-infra-prow-build
   name: ci-kubernetes-e2e-gci-gce-scalability
   tags:
   - "perfDashPrefix: gce-100Nodes-master"


wojtek-t:
/assign @mm4tt @jprzychodzen 

@mborsz - FYI as current oncall

jprzychodzen:
It seems that this change will affect access to projects, affecting possibility to debug for sig-scalability members.

Could you explain what is an expected IAM policy for projects in this pool?

spiffxp:
@jprzychodzen what IAM policy do you need?  I can create an @kubernetes.io google group for whomever scalability needs, and assign it the project viewer role for these projects, WDYT?

jprzychodzen:
Right now we have an Owner on those projects, so it's hard to provide complete list of permissions that we are requiring.

Viewer seems like a good starting point for most of the use cases, however it would be great to have elevated privileges for some small subset of people - like [Leadership](https://github.com/kubernetes/community/tree/master/sig-scalability#leadership) and [sig-scalability oncall](https://storage.googleapis.com/test-infra-oncall/oncall.html). This way we will have quick way to react during emergencies (eg. broken tests).

spiffxp:
I've opened https://github.com/kubernetes/k8s.io/issues/844 which uses owner for leads and oncall

spiffxp:
https://github.com/kubernetes/k8s.io/pull/919 has merged, I'd like to see this in during working hours today so I can rollback if there are any issues, and leave it to soak over the weekend if not

jprzychodzen:
Thanks! It looks good, it seems that there are appropriate permissions in place.

spiffxp:
Keeping an eye on runs

https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-e2e-gci-gce-scalability/1266440343058911233 - timed out

Earlier runs hover around ~50min overall

So a 40min git fetch is probably part of the problem, but doesn't explain how it got over 120min
```
I0529 18:49:00.764] Call:  git fetch --quiet --tags https://github.com/kubernetes/kubernetes master
I0529 19:28:40.363] process 66 exited with code 0 after 39.7m
```

jprzychodzen:
I do not know the details, but sig-scalability test jobs are not migrated to pod-utilities and use service account to directly access bucket and store logs. Just from a quick glance this can be related.

jprzychodzen:
Last run passed, logexporter provided artifact into [expected location](https://gcsweb.k8s.io/gcs/kubernetes-jenkins/logs/ci-kubernetes-e2e-gci-gce-scalability/1266477589925466112/artifacts/e2e-big-minion-group-0026/).

spiffxp:
FWIW this cluster uses a different service account, but it's also allowed write access into `kubernetes-jenkins`

Looks like the timeout was a blip.  Will let this run over the weekend to see how it fares before switching over the release-branch variants of this job

spiffxp:
Checking back in: https://testgrid.k8s.io/sig-release-master-blocking#gce-cos-master-scalability-100&width=5&graph-metrics=test-duration-minutes

There were some sporadic failures or timeouts over the weekend which I'm attributing to the build cluster being overloaded.

- https://github.com/kubernetes/k8s.io/pull/930 - raised the size of the cluster around 2020-05-31 3pm PT
- https://github.com/kubernetes/k8s.io/issues/931 - upgraded all of the nodes around 2020-05-31 7:30pm PT onward

All seems calm since then.

spiffxp:

PR #17727, opened 2020-05-27
Update scalability tests overrides to use new locations

wojtek-t:
/assign @mm4tt


wojtek-t:
/hold

Wait for https://github.com/kubernetes/perf-tests/pull/1280 to be merged first.

PR #17731, opened 2020-05-27
delete docker conformance test group

MHBauer:
no results for conformance after k8s-v1.14

no docker ce or ee results after v1.14 https://github.com/cncf/k8s-conformance

https://github.com/kubernetes/test-infra/blob/e3ac4d5f47937048cbb737dbe9e451abbf686fa6/kettle/buckets.yaml#L46-L48

If I check the bucket, the last results were uploaded about a year ago.

It's on the sig-node dashboard, but I think docker might be responsible party for updates, 
/cc @jeffqwb2017

### DIFF
--- a/config/testgrids/conformance/conformance-all.yaml
+++ b/config/testgrids/conformance/conformance-all.yaml
@@ -235,21 +235,6 @@ test_groups:
 - name: cloud-provider-huaweicloud-e2e-conformance-release-v1.17
   gcs_prefix: k8s-conform-huaweicloud/cloud-provider-huaweicloud/e2e-conformance-release-v1.17
 
-# Docker node conformance test group
-- name: ce18.09-ubuntu16.04-k8s1.13.x-conformance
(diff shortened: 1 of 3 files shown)


spiffxp:
/cc @derekwaynecarr 
I have no objections to removing from sig-node's dashboard (assuming sig-node is OK with this)

/hold
But I'm not sure we've gone through and actively pruned stale conformance results like this.  I took a guess at an e-mail contact and sent a message.  If we don't hear back within a week then I'm fine removing from conformance.

If you'd like to just get the results off of sig-node's dashboard in this PR and save the conformance changes for another PR, I will remove the hold

bsdnet:
/cc @dchen1107  @Random-Liu 

I have asked around within my team and adjacent team,  and looks no body knows.

Considering those tests haven't run for quite a while and also, the docker version is out of date.

I agreed that we can remove them if still no response.

MHBauer:
rebased
/remove-lifecycle stale

spiffxp (approved):
/approve
/lgtm
/hold cancel
/meow

PR #17733, opened 2020-05-27
add greenhouse-metrics for k8s-infra-prow-build

spiffxp:
we still need a dashboard that uses this


config/prow/cluster/monitoring/additional-scrape-configs_secret.yaml
```diff
@@ -16,6 +16,7 @@ stringData:
       static_configs:
         - targets:
             - "35.225.115.154" # external ip greenhouse-metrics for k8s-prow-builds
+            - "34.72.140.202" # external ip greenhouse-metrics for k8s-infra-prow-build
```
BenTheElder, line 19: not sure how this behaves but
/lgtm
/hold

PR #17734, opened 2020-05-27
use cos image for containerd node-e2e tests

MHBauer:
This should be fairly safe to experiment with. These both run successfully currently, and are not listed as blocking.

https://github.com/kubernetes/test-infra/blob/e3ac4d5f47937048cbb737dbe9e451abbf686fa6/config/jobs/kubernetes/sig-node/containerd.yaml#L219-L248

https://github.com/kubernetes/test-infra/blob/e3ac4d5f47937048cbb737dbe9e451abbf686fa6/config/jobs/kubernetes/sig-node/containerd.yaml#L309-L338

/cc @bart0sh @vpickard @bsdnet

### DIFF
--- a/jobs/e2e_node/containerd/containerd-master/image-config.yaml
+++ b/jobs/e2e_node/containerd/containerd-master/image-config.yaml
@@ -4,6 +4,6 @@ images:
     project: ubuntu-os-gke-cloud
     metadata: "user-data</go/src/github.com/containerd/cri/test/e2e_node/init.yaml,containerd-configure-sh</go/src/github.com/containerd/cri/cluster/gce/configure.sh,containerd-env</workspace/test-infra/jobs/e2e_node/containerd/containerd-master/env"
   cos-stable:
-    image: cos-77-12371-274-0
+    image_family: cos-81-lts
     project: cos-cloud
     metadata: "user-data</go/src/github.com/containerd/cri/test/e2e_node/init.yaml,containerd-configure-sh</go/src/github.com/containerd/cri/cluster/gce/configure.sh,containerd-extra-init-sh</go/src/github.com/containerd/cri/test/e2e_node/gci-init.sh,containerd-env</workspace/test-infra/jobs/e2e_node/containerd/containerd-master/env,gci-update-strategy=update_disabled"


jobs/e2e_node/containerd/containerd-master/image-config.yaml
```diff
@@ -4,6 +4,6 @@ images:
     project: ubuntu-os-gke-cloud
     metadata: "user-data</go/src/github.com/containerd/cri/test/e2e_node/init.yaml,containerd-configure-sh</go/src/github.com/containerd/cri/cluster/gce/configure.sh,containerd-env</workspace/test-infra/jobs/e2e_node/containerd/containerd-master/env"
   cos-stable:
-    image_regex: cos-stable-73-11647-510-0
+    image_family: cos-stable
```
bsdnet, line 7: I would suggest to use cos-81-lts rather than cos-stable here.

MHBauer: kind of raises the question of what lts vs stable means. If it's a floating image it's a floating image. I do not know why this runs against both ubuntu and cos.
/shrug

MHBauer:
/assign @klueska @derekwaynecarr 

right now the cos instance of these tests is not running until we fix all the out of date images and merge https://github.com/kubernetes/kubernetes/pull/91470

jobs/e2e_node/containerd/containerd-master/image-config.yaml
bsdnet: Maybe both ubuntu and cos are OSes used for early K8s development.

derekwaynecarr:
i tagged https://github.com/kubernetes/kubernetes/pull/91470

we will need this to rebase.

bart0sh:
I'd propose to wait a bit until https://github.com/kubernetes/kubernetes/pull/91854 is merged. if the recommended policy is to use 'image:' then we should probably follow it. In this particular case it would mean changing ```image: cos-77-12371-274-0``` to ```image: cos-81-12871-119-0```

jobs/e2e_node/containerd/containerd-master/image-config.yaml
```diff
@@ -4,6 +4,6 @@ images:
     project: ubuntu-os-gke-cloud
     metadata: "user-data</go/src/github.com/containerd/cri/test/e2e_node/init.yaml,containerd-configure-sh</go/src/github.com/containerd/cri/cluster/gce/configure.sh,containerd-env</workspace/test-infra/jobs/e2e_node/containerd/containerd-master/env"
   cos-stable:
-    image: cos-77-12371-274-0
+    image_family: cos-81-lts
```
SergeyKanzhelev, line 7: As per @bsdnet comment:

```suggestion
    image: cos-81-12871-119-0
```

jobs/e2e_node/containerd/containerd-master/image-config.yaml
```diff
@@ -4,6 +4,6 @@ images:
     project: ubuntu-os-gke-cloud
     metadata: "user-data</go/src/github.com/containerd/cri/test/e2e_node/init.yaml,containerd-configure-sh</go/src/github.com/containerd/cri/cluster/gce/configure.sh,containerd-env</workspace/test-infra/jobs/e2e_node/containerd/containerd-master/env"
   cos-stable:
-    image: cos-77-12371-274-0
+    image: cos-81-12871-1185-0
```
SergeyKanzhelev, line 7: why not `image_family` here as well with the `lts`?

MHBauer: https://github.com/kubernetes/test-infra/pull/17734#discussion_r469582242

SergeyKanzhelev: 🤣  ok. This comment was before we discussed it on Monday. Now I feel that consistently make it `81-lts` would be better.

MHBauer:
rebased updated to family as agreed

PR #17735, opened 2020-05-27
set grace period on some key kind jobs

BenTheElder:
/cc @amwat @aojea @spiffxp 

xref: https://github.com/kubernetes-sigs/kind/issues/1626

### DIFF
--- a/config/jobs/kubernetes-sigs/kind/kind-presubmits.yaml
+++ b/config/jobs/kubernetes-sigs/kind/kind-presubmits.yaml
@@ -64,6 +64,7 @@ presubmits:
     path_alias: sigs.k8s.io/kind
     decoration_config:
       timeout: 40m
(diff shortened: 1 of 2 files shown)


config/jobs/kubernetes-sigs/kind/kind-presubmits.yaml
```diff
@@ -64,6 +64,7 @@ presubmits:
     path_alias: sigs.k8s.io/kind
     decoration_config:
       timeout: 40m
+      terminationGracePeriodSeconds: 300
```
BenTheElder, line 67: this is wrong, bad cut and paste, it's `grace_period: 15s` (for the default anyhow)

aojea: `terminationGracePeriodSeconds`  is the pod.spec field, right?

PR #17736, opened 2020-05-28
correct grace period config and match bootstrap

BenTheElder:
the previous key was a bad copy paste, i'm surprised it passed validation, I guess we don't have strict key checking still

this also updates the value to match bootstrap, I did some thorough digging this time and i'm pretty certain that bootstrap jobs get a 15m grace period if the timeout actually is triggered (which is very buggy in bootstrap).

/cc @aojea @amwat

### DIFF
--- a/config/jobs/kubernetes-sigs/kind/kind-presubmits.yaml
+++ b/config/jobs/kubernetes-sigs/kind/kind-presubmits.yaml
@@ -64,7 +64,7 @@ presubmits:
     path_alias: sigs.k8s.io/kind
     decoration_config:
       timeout: 40m
-      terminationGracePeriodSeconds: 300
+      grace_period: 15m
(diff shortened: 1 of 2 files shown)


BenTheElder:
15m seems high, but for now i'm just going to match bootstrap, I'm most concerned with making sure we trigger cleanup etc. properly when we hit the timeout, we can narrow the period later if we wish.

BenTheElder:
To be reviewed

BenTheElder:
https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_kind/1627/pull-kind-e2e-kubernetes-1-18/1265854255458160642/build-log.txt

entrypoint logs show the 15m timeout, but not anything happening during it.

PR #17739, opened 2020-05-28
Added k8s-infra-dns-updater svcacct as trusted

bartsmykla:
This service account will be used to run dns updates using octodns when the config files under [k/k8s.io/dns/zone-configs/*.yaml](https://github.com/kubernetes/k8s.io/tree/8ab172615215f3e43fe455b5004c5a66d6df1105/dns/zone-configs) will change

More informations about whole process of automating this efforts: [k/k8s.io#786](https://github.com/kubernetes/k8s.io/issues/783)

/cc @spiffxp 

Signed-off-by: Bart Smykla <bsmykla@vmware.com>


spiffxp:
This should be added to https://github.com/kubernetes/k8s.io/blob/master/infra/gcp/clusters/projects/k8s-infra-prow-build-trusted/prow-build-trusted/resources/build-serviceaccounts.yaml

bartsmykla:
Oh, apologies I've got lost

/close

PR #17746, opened 2020-05-28
prow: sidecar: fix upload of empty files

sbueringer:
Because for empty files (io.Reader) the Write call was never executed,
no files in GCS were created. Now we always create a Writer and therefore
even empty files are written (on Close())

See also: 
* bug report: https://kubernetes.slack.com/archives/CDECRSC5U/p1590603056330600
* context: https://github.com/kubernetes/test-infra/issues/11260

### DIFF
--- a/prow/pod-utils/gcs/upload.go
+++ b/prow/pod-utils/gcs/upload.go
@@ -189,12 +189,20 @@ func (w *openerObjectWriter) Write(p []byte) (n int, err error) {
 }
 
 func (w *openerObjectWriter) Close() error {
-	if w.writeCloser != nil {
-		err := w.writeCloser.Close()
-		w.writeCloser = nil
-		return err
+	if w.writeCloser == nil {
+		// Always create a writer even if Write() was never called
+		// otherwise empty files are never created, because Write() is
+		// never called for them
+		var err error
+		w.writeCloser, err = w.Opener.Writer(w.Context, fmt.Sprintf("%s/%s", w.Bucket, w.Dest), w.opts...)
+		if err != nil {
+			return err
+		}
 	}
-	return nil
+
+	err := w.writeCloser.Close()
+	w.writeCloser = nil
+	return err
 }
 
 func (w *openerObjectWriter) ApplyWriterOptions(opts pkgio.WriterOptions) {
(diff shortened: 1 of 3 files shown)


alvaroaleman:
Why do we bother to upload empty files?

sbueringer:
> Why do we bother to upload empty files?

Good question. The old implementation did it and this apparently lead to problems in Knative (I don't know which exactly). Can you join the discussion in Slack? 

I personally don't care if we revert to the old behavior or keep the new one.This would be the fix, *if* we want to get the old behavior back.

albertomilan:
> Why do we bother to upload empty files?

Our code base make use of markers, which are empty files which are mainly used as flags.  While I could agree with you that the use case might not be extremely useful, there is no strict reason to disallow it or emit a success result for an upload that didn't happen so I would suggest to resolve this issue in order to avoid future duplications.

prow/pod-utils/gcs/upload.go
```diff
@@ -189,12 +189,20 @@ func (w *openerObjectWriter) Write(p []byte) (n int, err error) {
...
 func (w *openerObjectWriter) Close() error {
-	if w.writeCloser != nil {
-		err := w.writeCloser.Close()
-		w.writeCloser = nil
-		return err
+	if w.writeCloser == nil {
+		// Always create a writer even if Write() was never called
+		// otherwise empty files are never created, because Write() is
+		// never called for them
+		var err error
```
stevekuznetsov, line 196: Could this be 

```go
if _, err := w.Write([]byte()); err != nil {
    return err
}
err := w.writeCloser.Close()
w.writeCloser = nil
return err
```

sbueringer: Yup would be better to avoid code duplication

sbueringer: I kept the if `writeCloser == nil`. If I use the exact same you suggested it would always write an empty byte array, even if `Write()` was already used before (which should be 99% of the cases)

alvaroaleman:
Okay, sgtm. Can we have a test for this?

sbueringer:
Yup. I'll write it tomorrow

sbueringer:
> Yup. I'll write it tomorrow

Had some more time today

PR #17749, opened 2020-05-28
Crier: Reduce log spam and include reporter in logs

alvaroaleman:
This PR:
* Downgrades the logging on the `EventHandlers` to `Trace`, currently we generate one log message per event we receive. The only use of this I can see is debugging the underlying machinery and while it might not be impossible that we ever need that it is sufficiently unlikely that we should avoid logging it by default
* Re-Uses a `*logrus.Entry` throughout the sync and adds the reporter name to it
* Replaces the hand-written "single retry on conflict" with upstreams `retry.OnConflict` and adds some godocs around the rationale of doing that in the first place

### DIFF
--- a/prow/crier/controller.go
+++ b/prow/crier/controller.go
@@ -27,7 +27,7 @@ import (
 	jsonpatch "github.com/evanphx/json-patch"
 	"github.com/sirupsen/logrus"
 
-	v1 "k8s.io/test-infra/prow/apis/prowjobs/v1"
+	"k8s.io/test-infra/prow/apis/prowjobs/v1"
 
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -35,6 +35,7 @@ import (
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/util/retry"
 	"k8s.io/client-go/util/workqueue"
 	clientset "k8s.io/test-infra/prow/client/clientset/versioned"
 	pjinformers "k8s.io/test-infra/prow/client/informers/externalversions/prowjobs/v1"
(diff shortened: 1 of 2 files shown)


prow/crier/controller.go
```diff
@@ -165,7 +166,7 @@ func (c *Controller) updateReportState(pj *v1.ProwJob) error {
 	if newpj.Status.PrevReportStates == nil {
 		newpj.Status.PrevReportStates = map[string]v1.ProwJobState{}
 	}
-	newpj.Status.PrevReportStates[c.reporter.GetName()] = newpj.Status.State
+	newpj.Status.PrevReportStates[c.reporter.GetName()] = reportedState
```
alvaroaleman, line 169: Before, we just set this unconditionally to `newpj.Status.State` and due to the retrying that might have changed since we did the reporting

prow/crier/controller.go
```diff
@@ -181,7 +182,7 @@ func (c *Controller) updateReportState(pj *v1.ProwJob) error {
 		logrus.Warnf("Empty merge patch: pjData: %s, newpjData: %s", string(pjData), string(newpjData))
 	}
 
-	logrus.Infof("Created merge patch: %v", string(patch))
+	log.Infof("Created merge patch: %v", string(patch))
```
stevekuznetsov, line 185: Should this be `Info`? I sure don't want to see a merge patch .... I imagine we log what we were *trying* to do in English, which should be good enough.

alvaroaleman: But why not, they are interesting?

JK, I removed it, it was cargoed from before.

prow/crier/controller.go
```diff
@@ -264,47 +266,39 @@ func (c *Controller) processNextItem() bool {
...
-			"jobStatus": pj.Status,
-		}
-		logrus.WithError(err).WithFields(fields).Error("failed to report job")
+		log.WithError(err).Error("failed to report job")
 		return c.retry(key, err)
 	}
 
-	logrus.WithField("prowjob", keyRaw).Info("Updated job, now will update pj")
+	log.Info("Reported job, now will update pj")
+	reportedState := pj.Status.State
```
fejta, line 283: This needs to be pjob.Status.State

PR #17750, opened 2020-05-29
Kops - Add dashboards for distros and k8s versions

hakman:
Adds more dashboards to the Kops grid.


rifelpet:
/lgtm
/approve

the switch from release/stable to release/latest is because we already have jobs for k8s 1.18, and release/stable points to 1.18 so we were running the tests twice. now the `latest` jobs will test 1.19

PR #17752, opened 2020-05-29
[Windows upgrade e2e] config the tests

dntczdx:


config/testgrids/config.yaml
```diff
@@ -204,6 +204,15 @@ dashboards:
         value: 'ci-ovf-import-e2e-tests-daily issue: <test-name>'
       - key: body
         value: <test-url>
+  - name: ci-windows-upgrade-tests
```
chases2, line 207: Please add this in the `/testgrids/conformance/conformance-all.yaml` file, with the dashboard that uses it.

dntczdx: I think it has been added there. Did you review the latest commit?

spiffxp:
I'm confused why these are showing up in conformance, I don't think they have anything to do with conformance, but I'll save that for elsewhere

Issue #9628, opened 2018-09-28
Prow badge is red, should be green?


bobcatfish:
# Expected
When using the prow badge (e.g. https://prow.knative.dev/badge.svg?jobs=pull-knative-build-pipeline-unit-tests) to display the status of a PR job, it should be green when the latest run is successful, and red if it is not successful.

# Observed
In https://github.com/knative/build-pipeline we are using badges to show the Prow CI status, unfortunately these are on PR (not continuous integration) workflows, so I can understand that a failing PR might make the badge turn red, but currently the badge is red and that doesn't seem to be the case:

![image](https://user-images.githubusercontent.com/432502/46237713-2e2d2f00-c33b-11e8-8f43-7d51b8ce6960.png)

When I look at the prow status for that job (https://prow.knative.dev/?job=pull-knative-build-pipeline-build-tests), the latest status is green (though there have been some reds):

![image](https://user-images.githubusercontent.com/432502/46237736-4ef58480-c33b-11e8-9ae9-3590c1d35285.png)

BenTheElder:
/help

the code is here: https://github.com/kubernetes/test-infra/blob/master/prow/cmd/deck/badge.go

something needs fixing with that computation :^)

anwaradil:
I can work on this. I have been staring at this code for a while and I think the issue is we are not checking if the recent job has been successful in line 124 of badge .go.  

The question is should we even check the length if the recent job is successful? Assuming we turn the badge to red or green only according to the latest job but not per the last 8 runs.

Also, Maybe I need to look some more but I am not seeing where we are using or calling the func pickLatestJobs?

nzoueidi:
is this being looking into? otherwise I may want to take it :)

stevekuznetsov:
Go ahead and take it!

anwaradil:
Yes take it :)

nzoueidi:
Sorry for the delay, I have been out in vacation. 🎉  
I would like to propose an approach here before doing all of the coding..
I am thinking about adding another function in the `badge.go` file to have a sort of struct including all the latest 8 jobs and their status. Based on these status the func will have a boolean output then I can add it in the if statement (L124) in the `renderBadge` func. WDYT?
I would be very happy to have your inputs and if there is any suggestion how we could do this in a better way.

stevekuznetsov:
How do you propose to take the last eight jobs and determine the boolean output? The approach roughly sounds good to me!

BenTheElder:
[...]

PR #17759, opened 2020-05-29
Update Prow Deck Job badges display order

jonahjon:
Prow Jobs on line 91 were being ordered in the Slice with the lowest StartTime as element 0. This caused the appended job on like 103 to be the first glob matching job, and not the latest glob matching job. By reverse ordering the Slice it will perform the intended function.

### DIFF
--- a/prow/cmd/deck/badge.go
+++ b/prow/cmd/deck/badge.go
@@ -85,10 +85,10 @@ func makeShield(subject, status, color string) []byte {
 
 // pickLatestJobs returns the most recent run of each job matching the selector,
 // which is comma-separated list of globs, for example "ci-ti-*,ci-other".
-// jobs will be sorted
+// jobs will be sorted by StartTime in reverse order to display recent jobs first
 func pickLatestJobs(jobs []prowapi.ProwJob, selector string) []prowapi.ProwJob {
 	sort.Slice(jobs, func(i, j int) bool {
-		return jobs[i].Status.StartTime.Before(&jobs[j].Status.StartTime)
+		return jobs[j].Status.StartTime.Before(&jobs[i].Status.StartTime)
 	})
 	var out []prowapi.ProwJob
 	have := make(map[string]bool)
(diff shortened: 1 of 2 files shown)


jonahjon:
Closes #9628

jonahjon:
I signed it

prow/cmd/deck/badge.go
```diff
@@ -85,10 +85,10 @@ func makeShield(subject, status, color string) []byte {
 
 // pickLatestJobs returns the most recent run of each job matching the selector,
 // which is comma-separated list of globs, for example "ci-ti-*,ci-other".
-// jobs will be sorted
+// jobs will be sorted by StartTime in reverse order to display recent jobs first
 func pickLatestJobs(jobs []prowapi.ProwJob, selector string) []prowapi.ProwJob {
 	sort.Slice(jobs, func(i, j int) bool {
-		return jobs[i].Status.StartTime.Before(&jobs[j].Status.StartTime)
+		return jobs[j].Status.StartTime.Before(&jobs[i].Status.StartTime)
```
alvaroaleman, line 91: Can you please use `StartTime.After(...` rather than inversing the index usage to make this easier to read?

Also, please update `TestPickLatest` to include timestamps to make sure we don't regress here.
/ok-to-test

jonahjon: 1. Currently `k8s.io/apimachinery/pkg/apis/meta/v1` Doesn't have `.After()` only `.Before()`. Converting to `time.time` and Back to `meta.Time` might be a little slow, but can be done.

2.  Added tests!

prow/cmd/deck/badge_test.go
```diff
@@ -20,33 +20,39 @@ import (
...
 	"testing"
+	"time"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	prowapi "k8s.io/test-infra/prow/apis/prowjobs/v1"
 )
 
 func TestPickLatest(t *testing.T) {
+	earliest := metav1.Now()
+	time.Sleep(100 * time.Millisecond)
```
alvaroaleman, line 31:
```suggestion
	earliest := metav1.Time{Time: time.Now().Add(-100 * time.Millisecond)
```

jonahjon:
Took out the time import failing test after the commit suggestion

PR #17760, opened 2020-05-29
Update k8s.io group reconciliation jobs

spiffxp:
Depends on https://github.com/kubernetes/k8s.io/pull/924

PR #17762, opened 2020-05-29
Fix pull-k8sio-groups-test

spiffxp:
fix what I broke in https://github.com/kubernetes/test-infra/pull/17760


spiffxp:
ref https://github.com/kubernetes/k8s.io/issues/928

PR #17655, opened 2020-05-20
CAPZ: add storage-related periodic jobs

chewong:
Setting all the new periodic jobs to 3h to get a faster test signal before changing it to 24h.

/assign @CecileRobertMichon

PR #17763, opened 2020-05-29
CAPZ: add conformance presubmit job on MachinePool cluster

chewong:
Creating a presubmit job to test https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/659. Removing `pull-cluster-api-provider-azure-conformance-v1alpha3-1-18` since we already have some coverages in terms of using built K8s artifacts from a release branch from jobs such as https://testgrid.k8s.io/provider-azure-upstream#k8s-e2e-azure-disk-1-16-capz.

Also reducing the frequency of jobs created in https://github.com/kubernetes/test-infra/pull/17655 from 3h to 24h

/assign @CecileRobertMichon

### DIFF
--- a/config/jobs/kubernetes-sigs/cluster-api-provider-azure/cluster-api-provider-azure-presubmits.yaml
+++ b/config/jobs/kubernetes-sigs/cluster-api-provider-azure/cluster-api-provider-azure-presubmits.yaml
@@ -170,7 +170,7 @@ presubmits:
       testgrid-dashboards: sig-cluster-lifecycle-cluster-api-provider-azure
       testgrid-tab-name: verify-boilerplate
       testgrid-alert-email: k8s-infra-staging-cluster-api-azure@kubernetes.io
-  - name: pull-cluster-api-provider-azure-conformance-v1alpha3-master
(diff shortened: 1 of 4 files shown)


config/jobs/kubernetes-sigs/cluster-api-provider-azure/cluster-api-provider-azure-presubmits.yaml
```diff
@@ -241,6 +241,10 @@ presubmits:
               value: "true"
             - name: USE_CI_ARTIFACTS
               value: "true"
+            - name: FEATURE_GATE_MACHINE_POOL
+              value: "true"
+            - name: USE_MACHINE_POOL
```
CecileRobertMichon, line 246: would it make sense to use the same var for both? Or maybe set `USE_MACHINE_POOL` to the `${FEATURE_GATE_MACHINE_POOL}`? I can't think of a situation where you'd want to set one to true but not the other

chewong: I will consolidate them into `FEATURE_GATE_MACHINE_POOL`.

Issue #14718, opened 2019-10-10
owners-label plugin should say why its trying to add labels


fejta:
/help
/good-first-issue
/area prow/hook
/area prow

We sometimes see errors like this:

```yaml
{
 insertId: "1qzmc7eg10yhd2q"  
 jsonPayload: {
  author: "ddebroy"   
  component: "hook"   
  event-GUID: "2d131c80-eba1-11e9-84a3-2526b95c6daa"   
  event-type: "pull_request"   
  file: "prow/plugins/owners-label/owners-label.go:119"   
  func: "k8s.io/test-infra/prow/plugins/owners-label.handle"   
  level: "warning"   
  msg: "Unable to add nonexistent labels: ["sig/gcp"]"   
  org: "kubernetes"   
  plugin: "owners-label"   
  pr: 83098   
  repo: "kubernetes"   
  url: "https://github.com/kubernetes/kubernetes/pull/83098"   
 }
 labels: {…}  
 logName: "projects/k8s-prow/logs/hook"  
 receiveTimestamp: "2019-10-10T21:03:19.691649425Z"  
 resource: {…}  
 severity: "ERROR"  
 timestamp: "2019-10-10T21:03:18Z"  
}
```

Given the size of https://github.com/kubernetes/kubernetes/pull/83098 it is hard to figure out why this PR is trying to add a `sig/gcp` label. This is almost certainly because one or more `OWNERS` files in the kubernetes/kubernetes repo have a line like this:

https://github.com/kubernetes/test-infra/blob/203256b8c9ee04c04cd2b4c897977539b191fd4c/prow/OWNERS#L23-L24

It would be very helpful if we updated the error message: 

https://github.com/kubernetes/test-infra/blob/203256b8c9ee04c04cd2b4c897977539b191fd4c/prow/plugins/owners-label/owners-label.go#L114

To include the files that added the label:

https://github.com/kubernetes/test-infra/blob/203256b8c9ee04c04cd2b4c897977539b191fd4c/prow/plugins/owners-label/owners-label.go#L82

fejta:
How are things going? Reach out to #prow on slack if you want some help!

fejta:
Good luck :pray: :prayer_beads:

fejta:
Any progress here?

fejta:
Suspect this could still use some help!

RobertKielty:
Hi @fejta I'll a look at this today and reach out slack if I need help

/assign

[...]

PR #17767, opened 2020-05-30
Fixes 14718 implements FindConfigFileForLabel on Repoowners

RobertKielty:
Implements #14718 

When there is a problem adding a label to an issue, the logged error can report which OWNERS file contained the label that could not be added. 

`log.WithError(err).Errorf("GitHub failed to add label: %s found in %s", labelToAdd, oc.FindConfigFileForLabel(labelToAdd))`

This is implemented by mapping labels to the CONFIG files where they were found in repoowners.go

### DIFF
--- a/prow/plugins/owners-label/owners-label.go
+++ b/prow/plugins/owners-label/owners-label.go
@@ -26,6 +26,7 @@ import (
 	"k8s.io/test-infra/prow/github"
 	"k8s.io/test-infra/prow/pluginhelp"
 	"k8s.io/test-infra/prow/plugins"
+	"fmt"
 )
 
 const (
@@ -113,7 +114,11 @@ func handle(ghc githubClient, oc ownersClient, log *logrus.Entry, pre *github.Pu
 			continue
 		}
 		if err := ghc.AddLabel(org, repo, number, labelToAdd); err != nil {
-			log.WithError(err).Errorf("GitHub failed to add label: %s found in %s", labelToAdd, oc.FindConfigFileForLabel(labelToAdd))
+			log.WithError(err).WithFields(
+				logrus.Fields{
+					"label":        labelToAdd,
+					"config-files": oc.FindConfigFileForLabel(labelToAdd),
+				}).Error("Failed to add label found in config-file(s)")
 		}
 	}
 
--- a/prow/plugins/owners-label/owners-label.go
+++ b/prow/plugins/owners-label/owners-label.go
@@ -46,6 +46,7 @@ func helpProvider(config *plugins.Configuration, _ []config.OrgRepo) (*pluginhel
 
 type ownersClient interface {
 	FindLabelsForFile(path string) sets.String
+	FindConfigFileForLabel(label string) string
 }
 
 type githubClient interface {
@@ -112,7 +113,7 @@ func handle(ghc githubClient, oc ownersClient, log *logrus.Entry, pre *github.Pu
 			continue
 		}
 		if err := ghc.AddLabel(org, repo, number, labelToAdd); err != nil {
-			log.WithError(err).Errorf("GitHub failed to add the following label: %s", labelToAdd)
+			log.WithError(err).Errorf("GitHub failed to add label: %s found in %s", labelToAdd, oc.FindConfigFileForLabel(labelToAdd))
 		}
 	}
 
--- a/prow/repoowners/repoowners.go
+++ b/prow/repoowners/repoowners.go
@@ -36,6 +36,8 @@ import (
 	prowConf "k8s.io/test-infra/prow/config"
 )
 
+// Implements https://go.k8s.io/owners
+
 const (
 	ownersFileName  = "OWNERS"
 	aliasesFileName = "OWNERS_ALIASES"
@@ -216,7 +218,6 @@ type RepoOwner interface {
 	FindApproverOwnersForFile(path string) string
 	FindReviewersOwnersForFile(path string) string
 	FindLabelsForFile(path string) sets.String
-	FindConfigFileForFile(path string) sets.String
 	FindConfigFileForLabel(label string) string
 	IsNoParentOwners(path string) bool
 	LeafApprovers(path string) sets.String
@@ -894,12 +895,6 @@ func (o *RepoOwners) TopLevelApprovers() sets.String {
 	return o.entriesForFile(".", o.approvers, false)
 }
 
-
-// FindConfigFileForFile returns set of config files used to be configure
-// the RepoOwners struct for the repo file specified in path
-func (o *RepoOwners) FindConfigFileForFile(path string) sets.String {
-	return o.configSources[path][nil]
-}
 
 // FindConfigFileForLabel returns set of config files used to configure
 // the RepoOwners struct for the repo file specified in path
--- a/prow/repoowners/repoowners.go
+++ b/prow/repoowners/repoowners.go
@@ -693,7 +693,7 @@ func NormLogins(logins []string) sets.String {
 	return normed
 }
 
-var defaultDirOptions = dirOptions{}
+var defaultDirOptions dirOptions
 
 // applyConfigToPath takes a parsed Config for a path and sets up fields in RepoOwners
 func (o *RepoOwners) applyConfigToPath(path string, re *regexp.Regexp, config *Config) {
--- a/prow/repoowners/repoowners.go
+++ b/prow/repoowners/repoowners.go
@@ -499,8 +499,9 @@ func canonicalize(path string) string {
 func (o *RepoOwners) walkFunc(path string, info os.FileInfo, err error) error {
 	log := o.log.WithField("path", path)
 	log.Debugf("walkFunc(%q, %q,  %q)", path, info, err)
+
 	if err != nil {
-		log.WithError(err).Errorf("Error while walking OWNERS files while processing %q", path)
+		log.WithError(err).Error("Error while walking OWNERS files")
 		return nil
 	}
 	filename := filepath.Base(path)
--- a/prow/repoowners/repoowners.go
+++ b/prow/repoowners/repoowners.go
@@ -56,7 +56,7 @@ type Config struct {
 	Reviewers         []string `json:"reviewers,omitempty"`
 	RequiredReviewers []string `json:"required_reviewers,omitempty"`
 	Labels            []string `json:"labels,omitempty"`
-	SourcePath        string   `json:"-"`
+	sourcePath        string
 }
 
 // SimpleConfig holds options and Config applied to everything under the containing directory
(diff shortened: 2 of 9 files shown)


RobertKielty:
Working on this now.

prow/repoowners/repoowners.go
```diff
@@ -214,6 +216,9 @@ type RepoOwner interface {
 	FindApproverOwnersForFile(path string) string
 	FindReviewersOwnersForFile(path string) string
 	FindLabelsForFile(path string) sets.String
+	FindOwnersConfigForFile(path string) map[string]map[*regexp.Regexp]sets.String
```
RobertKielty, line 219: Self reviewing. This is not needed, I'm checking this now

prow/repoowners/repoowners.go
```diff
@@ -784,13 +811,17 @@ func (o *RepoOwners) IsNoParentOwners(path string) bool {
...
+// for the file specified by path.
+//
+// path - should be a full path to a filename and not directory as the final directory will be discounted if enableMDYAML is true
+// leafOnly - when true just use OWNERS deepest in the tree (closest to the path)
+//            when false retrieve all entries in all OWNERS file
 func (o *RepoOwners) entriesForFile(path string, people map[string]map[*regexp.Regexp]sets.String, leafOnly bool) sets.String {
 	d := path
+	fmt.Printf("entriesForFile : path %30v=\n", path)
+	fmt.Printf("entriesForFile : entries %v\n", people)
+	fmt.Printf("entriesForFile : %q\n", o.FindOwnersConfigForFile(path))
```
RobertKielty, line 824: These should also be removed

prow/repoowners/repoowners.go
```diff
@@ -862,6 +893,25 @@ func (o *RepoOwners) RequiredReviewers(path string) sets.String {
...
+// TopLevelApprovers return approvers for file at path
 func (o *RepoOwners) TopLevelApprovers() sets.String {
 	return o.entriesForFile(".", o.approvers, false)
 }
+
+// FindOwnersConfigForFile returns the config file as a map (OWNERS, OWNERS_ALIAS, that md file thing that I can't find out anything about)
+// that applies to the specified path
+func (o *RepoOwners) FindOwnersConfigForFile(path string) map[string]map[*regexp.Regexp]sets.String {
+	return o.configSources
+}
```
RobertKielty, line 905: To be removed.

RobertKielty:
Reproducing locally as follows ...

/usr/bin/bazel test  ...

prow/repoowners/repoowners.go
```diff
@@ -707,6 +720,20 @@ func (o *RepoOwners) applyConfigToPath(path string, re *regexp.Regexp, config *C
 		}
 		o.labels[path][re] = sets.NewString(config.Labels...)
 	}
+	if len(config.SourcePath) > 0 {
+		if o.configSources[path] == nil {
+			o.configSources[path] = make(map[*regexp.Regexp]sets.String)
```
fejta, line 725: style is generally `foo = map{}` rather than `foo = make(map)`

prow/repoowners/repoowners.go
```diff
@@ -48,11 +50,13 @@ type dirOptions struct {
 }
 
 // Config holds roles+usernames and labels for a directory considered as a unit of independent code
+// along with the path to the OWNERS file that was parsed to retrieve those labels, usernames and roles.
 type Config struct {
 	Approvers         []string `json:"approvers,omitempty"`
 	Reviewers         []string `json:"reviewers,omitempty"`
 	RequiredReviewers []string `json:"required_reviewers,omitempty"`
 	Labels            []string `json:"labels,omitempty"`
+	SourcePath        string   `json:"-"`
```
fejta, line 59: Does this need to be exported? Thus far it doesn't look like it is being used outside the package.

prow/repoowners/repoowners.go
```diff
@@ -490,8 +498,9 @@ func canonicalize(path string) string {
 
 func (o *RepoOwners) walkFunc(path string, info os.FileInfo, err error) error {
 	log := o.log.WithField("path", path)
+	log.Debugf("walkFunc(%q, %q,  %q)", path, info, err)
 	if err != nil {
-		log.WithError(err).Error("Error while walking OWNERS files.")
+		log.WithError(err).Errorf("Error while walking OWNERS files while processing %q", path)
```
fejta, line 503: Another option here would be:
`log.WithError(err).WithField("path", path).Error("Erro while walking OWNERS files")`

prow/repoowners/repoowners.go
```diff
@@ -682,6 +694,7 @@ func NormLogins(logins []string) sets.String {
 
 var defaultDirOptions = dirOptions{}
```
fejta, line 695: as long as we are near here:
`var defaultDirOptions dirOptions`

https://dave.cheney.net/practical-go/presentations/qcon-china.html#_use_a_consistent_declaration_style

prow/repoowners/repoowners.go
```diff
@@ -601,7 +610,10 @@ func (o *RepoOwners) ParseSimpleConfig(path string) (SimpleConfig, error) {
 	if err != nil {
 		return SimpleConfig{}, err
 	}
-	return LoadSimpleConfig(b)
+
+	simple, err := LoadSimpleConfig(b)
```
fejta, line 614: Can simplify this to:
`simple.SourcePath, err = LoadSimpleConfig(b)`

prow/repoowners/repoowners_test.go
```diff
@@ -1137,6 +1137,75 @@ func TestFindLabelsForPath(t *testing.T) {
...
+		}, {
+			name:                 "OWNERS in ROOT and LEAF",
+			path:                 leafDir,
+			label:                "sig/hg2tg",
+			expectedConfigSource: "./OWNERS",
+		}, {
+			name:                 "OWNERS in ROOT and LEAF",
+			path:                 "a/b/c/foo.txt",
+			label:                "sig/hg2tg",
+			expectedConfigSource: "./OWNERS",
```
fejta, line 1177: Interesting, this unit test passes?

`a/b/c/OWNERS` should take precedence over `./OWNERS` -- the former is deeper in the tree and therefore more specific.

RobertKielty:
Thanks for the review feedback @fejta I will work on this tomorrow.

fejta:
Sounds good

prow/repoowners/repoowners.go
RobertKielty: Agreed. Now sourcePath is no longer exported.

prow/repoowners/repoowners.go
RobertKielty: Ok, reviewing further,  on line 500 log already has a field Entry for path. So I think I can just drop the formatted string altogether.

prow/repoowners/repoowners.go
RobertKielty: Agreed, I changed this.

prow/repoowners/repoowners.go
RobertKielty: I was going with the flow here; I'll check to see if I can take your suggestion on both my changes and in the changes above using the unit tests and report back.

RobertKielty: Ok, so I went nuclear here and replaced *all* the map make's with map literals in the file.

Unit tests still pass. Let me know if you are ok with this?

prow/repoowners/repoowners_test.go
fejta: So I still don't understand how this works... Why doesn't this match a/b/c/OWNERS?

RobertKielty:
@fejta I apologise for the delay on getting to you on this; between CI Signal Shadow  work and CNCF prow plugin coding work I've taken my eye off this. I *can* do work on this again in the coming days.  Rob

prow/repoowners/repoowners.go
```diff
@@ -490,8 +498,10 @@ func canonicalize(path string) string {
 
 func (o *RepoOwners) walkFunc(path string, info os.FileInfo, err error) error {
 	log := o.log.WithField("path", path)
+	log.Debugf("walkFunc(%q, %q,  %q)", path, info, err)
```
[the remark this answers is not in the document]
RobertKielty: I'll remove it.

prow/repoowners/repoowners.go
```diff
@@ -490,8 +498,10 @@ func canonicalize(path string) string {
...
+
 	if err != nil {
-		log.WithError(err).Error("Error while walking OWNERS files.")
+		log.WithError(err).Error("Error while walking OWNERS files")
```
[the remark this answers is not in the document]
RobertKielty: Path has been added to the log logger instance on the first line of the `walkFunc`

prow/plugins/owners-label/owners-label.go
```diff
@@ -112,7 +113,7 @@ func handle(ghc githubClient, oc ownersClient, log *logrus.Entry, pre *github.Pu
 			continue
 		}
 		if err := ghc.AddLabel(org, repo, number, labelToAdd); err != nil {
-			log.WithError(err).Errorf("GitHub failed to add the following label: %s", labelToAdd)
+			log.WithError(err).Errorf("GitHub failed to add label: %s found in %s", labelToAdd, oc.FindConfigFileForLabel(labelToAdd))
```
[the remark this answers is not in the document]
RobertKielty: Agree with the feedback and also

I've changed existing-label to config-files in the Fields struct 
I've changed the Error message to "GHC failed to add label found in config-file(s)"

prow/repoowners/repoowners_test.go
RobertKielty: The reason this did not work as expected was that the test scaffolding had the ./OWNERS as being the source of the "sig/hg2g" label.

I change to test scafolding so that the leaf dir had a more specific label, "sig/hg2g-secondary-phase" and so now the tests pass.

I'm concerned though that this has exposed a problem with my implementation. Right now I pin a label to just a single OWNERS file.

Is it possible for two separate OWNERS files that exist on the same path to have the same label present in both files? Probably. If so how should I handle that?

It should be possible for me to build up a RepoOwners struct with data for the full repo as a report so that I can answer this question my self and I will work on that right now.

fejta:
Okay let's say we're trying to approve the file `a/b/c/foo.txt` and we have the following owners files:
* `/OWNERS`, `a/OWNERS`, `a/b/c/OWNERS`.

In this scenario:
* Any approver in `/OWNERS`, `a/OWNERS`, `a/b/c/OWNERS` can approve the change to `foo.txt`.
* We should recommend the most specific approvers, aka people in `a/b/c/OWNERS`

Now let's add a complication: `a/OWNERS` sets `no_parent_owners: true`. In this revised scenario:
* Any approver in `a/OWNERS` or `a/b/c/OWNERS` can approve the change to `foo.txt.`
  - The `/OWNERS` file is not considered at all.
* We should still recommend the most specific approvers: aka people in `a/b/c/OWNERS`.

PR #17769, opened 2020-05-31
PlankV2: Consider cache staleness for MaxConcurrency

alvaroaleman:


prow/plank/reconciler.go
```diff
@@ -111,15 +105,49 @@ func add(
...
+type shardedLock struct {
+	mapLock *sync.Mutex
+	locks   map[string]*semaphore.Weighted
+}
+
+func (s *shardedLock) getLock(key string) *semaphore.Weighted {
+	s.mapLock.Lock()
+	defer s.mapLock.Unlock()
+	if _, exists := s.locks[key]; !exists {
+		s.locks[key] = semaphore.NewWeighted(1)
```
stevekuznetsov, line 148: Why semaphore if you only have weight `1`?

alvaroaleman: Because of the `TryAcquire`, `sync.Mutex` doesn't have something like that

PR #17770, opened 2020-05-31
Bump COS image to latest version

bsdnet:

https://github.com/kubernetes/kubernetes/pull/91543 detect that stable image is gone. Now it silently failed. Instead, use the LTS image which will always succeed.

### DIFF
--- a/jobs/e2e_node/benchmark-config.yaml
+++ b/jobs/e2e_node/benchmark-config.yaml
@@ -1,21 +1,21 @@
 ---
 images:
   cosstable2-resource1:
-    image: cos-77-12371-251-0
+    image: cos-77-12371-274-0
     project: cos-cloud
     machine: n1-standard-1
     metadata: "user-data<test/e2e_node/jenkins/gci-init.yaml,gci-update-strategy=update_disabled"
     tests:
       - 'resource tracking for 0 pods per node \[Benchmark\]'
   cosstable2-resource2:
-    image: cos-77-12371-251-0
+    image: cos-77-12371-274-0
     project: cos-cloud
     machine: n1-standard-1
     metadata: "user-data<test/e2e_node/jenkins/gci-init.yaml,gci-update-strategy=update_disabled"
     tests:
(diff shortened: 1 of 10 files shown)


bsdnet:
/assign  @vpickard @bart0sh @MHBauer 
This CL uses image and get rid of image_regex, upgrade cos-73 to cos-77(cos-stable-73 does not match any image and silently fail). 
In future, we should always use LTS milestone. stable is the original channel before LTS is introduced. If there are two images, it should be the latest two LTS milestones. If it has one image, it should one of the latest two LTS. The chance for breaking using LTS is really slow because 
1.) LTS is much stable
2.) LTS has a long support period(15 months), and fully supported.
3.) Stable change every 6 weeks. This requires keep changing the image.

We can convert current images from cos-stables to cos-lts, image to image_family 
in the future changeset. That will avoid frequent configuration file change, but also
has its drawbacks. For this changeset, try to make it small to stablize the tests first.

bart0sh:
@bsdnet Would it be better to start using ```image_family: cos-77-lts``` instead? Otherwise tests will start failing every time gcp updates images that we use.

bsdnet:
> @bsdnet Would it be better to start using `image_family: cos-77-lts` instead?
I would suggest this for now until @MHBauer 's patch gets verified slowly.  There is a drawback if 
we use image_family. COS will release regularly, what should we do if the image is bad though it is
rare? We can't even rollback!  We need to think through about the situation.

> Otherwise tests will start failing every time gcp updates images that we use.

tests should not fail at all whether GCP publish new image or not. GCP never deleted images.Once the image was published, it will only be marked as DEPRECATED, and always be there.  The reason cos-stable-73-11647-510-0 failed is that the changeset 31b01ca206448881c3344be647b73c798e4e7e1b is wrong. It uses cos-stable-73 instead of co-73. 11647-510-0 is never released via the cos-stable channel, and so it does not exist at all. It is released  in cos-73-lts family.

In short, it is very safe to use latest LTS image. As to  whether to use 'image_family: cos-77-lts', let's do step by step. It also need more thinking.

PR #17772, opened 2020-05-31
prow: configurable metrics and pprof port

### LINKED ISSUE #17033: prow: make pprof and metrics port configurable
I would like to be able to configure the metrics and pprof port

**Why is this needed**:
It's not possible to run multiple prow components locally for development purposes because they are using the same metrics and pprof port. Adding a flag to these components to make it configurable should solve the problem.

sbueringer:
I only implemented it for Deck for now. After we agreed on how to implement it, I would apply this change to all callers of `ServePProf` and `ExposeMetrics`

* Implementation variants:
  * I would use a flagutil to avoid code duplication, but I'm not sure if `CommonOptions` is a good name. Alternatives:
    * `DebugOptions`
    * `ObservabilityOptions`
    * separate `MetricsOptions` and `PProfOptions`
* Should we force callers of `ServePProf` and `ExposeMetrics` to handover a port or should I add an additional func for each which allows handing over the port
* Is defaulting to the old port in flagutil enough or should we also default to the old port when port `0` is handed over?

Fixed #17033

### DIFF
--- a/prow/cmd/deck/main.go
+++ b/prow/cmd/deck/main.go
@@ -101,7 +101,7 @@ const (
 type options struct {
 	configPath            string
 	jobConfigPath         string
-	common                prowflagutil.CommonOptions
+	instrumentation       prowflagutil.InstrumentationOptions
 	kubernetes            prowflagutil.KubernetesOptions
 	github                prowflagutil.GitHubOptions
 	tideURL               string
@@ -195,7 +195,7 @@ func gatherOptions(fs *flag.FlagSet, args ...string) options {
 	fs.BoolVar(&o.allowInsecure, "allow-insecure", false, "Allows insecure requests for CSRF and GitHub oauth.")
 	fs.BoolVar(&o.dryRun, "dry-run", false, "Whether or not to make mutating API calls to GitHub.")
 	fs.StringVar(&o.pluginConfig, "plugin-config", "", "Path to plugin config file, probably /etc/plugins/plugins.yaml")
-	o.common.AddFlags(fs)
+	o.instrumentation.AddFlags(fs)
 	o.kubernetes.AddFlags(fs)
 	o.github.AddFlags(fs)
 	o.github.AllowAnonymous = true
@@ -279,7 +279,7 @@ func main() {
 	}
 
 	defer interrupts.WaitForGracefulShutdown()
-	pjutil.ServePProf(o.common.PProfPort)
+	pjutil.ServePProf(o.instrumentation.PProfPort)
 
 	// setup config agent, pod log clients etc.
 	configAgent := &config.Agent{}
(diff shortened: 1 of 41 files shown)


matthyx:
@sbueringer I think you need to update the test in `metrics_test.go` line 55.

sbueringer:
> @sbueringer I think you need to update the test in `metrics_test.go` line 55.

@matthyx Yup. I would prefer to have agreement on the implementation first. Because I also have to adjust about 20-30 callers of the `ServePProf` and `ExposeMetrics` before this PR is merge-ready :)

prow/metrics/metrics.go
```diff
@@ -31,20 +31,20 @@ import (
 	"k8s.io/test-infra/prow/interrupts"
 )
 
-const metricsPort = 9090
+const defaultMetricsPort = 9090
```
matthyx, line 34: Should we use this const everywhere instead of the few places where you still have 9090?
Same question for an hypothetical const for pprof-port (6060)?

prow/cmd/deck/main.go
```diff
@@ -102,6 +102,7 @@ const (
 type options struct {
 	configPath            string
 	jobConfigPath         string
+	common                prowflagutil.CommonOptions
```
matthyx, line 105: Also I am not sure people like generic names like `common`... maybe `port` or something related to their binding.
What do you think @alvaroaleman ?

alvaroaleman: Yeah, something like `InstrumentationOptions` would be good, the current name is too common :)

Also I am wondering if it would make sense to make this part of the config file rather than a flag or is the use case local development?

sbueringer: Use case is local development. I think it won't be used when Deck runs as Pod

sbueringer: I'll rename it to InstrumentationOptions

prow/flagutil/instrumentation.go
```diff
@@ -0,0 +1,44 @@
...
+	PProfPort int
+}
+
+// AddFlags injects common options into the given FlagSet.
+func (o *InstrumentationOptions) AddFlags(fs *flag.FlagSet) {
+	fs.IntVar(&o.MetricsPort, "metrics-port", DefaultMetricsPort, "port to serve metrics")
+	fs.IntVar(&o.PProfPort, "pprof-port", DefaultPProfPort, "port to serve pprof")
+}
+
+func (o *InstrumentationOptions) Validate(_ bool) error {
```
stevekuznetsov, line 42: Is this implementing some interface? Why do we need it?

sbueringer: It implements the flagutil.OptionGroup interface. I thought it would be a good idea so we can use it in the for loops we already have for other options groups to add flags, e.g.: https://github.com/sbueringer/test-infra/blob/efb7b016e6538072b773e2cd401188e68831058f/prow/cmd/gerrit/main.go#L93

PR #17773, opened 2020-05-31
prow: deck: add local dev doc

sbueringer:
Not sure if:
* there are other (better) ways to develop locally
* this should be documented somewhere else

### DIFF
--- a/prow/cmd/deck/README.md
+++ b/prow/cmd/deck/README.md
@@ -1,0 +1,36 @@
+
+# Deck
+
+## Running Deck locally
+
+Deck can be run locally by executing `./runlocal`. The scripts starts Deck via 
+Bazel using:
+* pre-generated data (extracted from a running Prow instance)
+* the local `config.yaml`
+* the local static files, template files and lenses
+
+
+## Debugging via Intellij / VSCode
+
+This section describes how to debug Deck locally by running it inside 
+VSCode or Intellij.
+
+```bash
+TEST_INFRA_DIR=${GOPATH}/src/k8s.io/test-infra
+
+# Prepare assets
+cd ${TEST_INFRA_DIR}
+bazel build //prow/cmd/deck:image.tar
+mkdir -p /tmp/deck
+tar -xvf ./bazel-bin/prow/cmd/deck/asset-base-layer.tar -C /tmp/deck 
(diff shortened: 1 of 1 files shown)


alvaroaleman:
Have you looked at prow/cmd/deck/runlocal, I think it more or less does whats proposed here?

sbueringer:
@alvaroaleman Looks very similar

runlocal:
* build and runs the deck image locally and uses the lenses/template/static files from inside the image
* also uses pregenerated data (looks like a pretty interesting feature, I wasn't aware of it)

doc in this PR:
* prepares the assets somewhere outside of the image (/tmp/deck)
* documents how to start it via an IDE

For me personally, it's a big advantage to be able to debug deck directly with an IDE, vs. only running it via Bazel.

Not sure if we want the documentation which this PR adds. I don't need it as I already know how to do it :).

alvaroaleman:
> For me personally, it's a big advantage to be able to debug deck directly with an IDE

Not sure I understand. How does "debug directly via an IDE work" that it cares about how the thing is started?

sbueringer:
> > For me personally, it's a big advantage to be able to debug deck directly with an IDE
> 
> Not sure I understand. How does "debug directly via an IDE work" that it cares about how the thing is started?

I'm using [Go debug configurations](https://www.jetbrains.com/help/go/creating-and-editing-run-debug-configurations.html) in Intellij which compiles the go binary with debug symbols and then starts it with dlv.

alvaroaleman:
@sbueringer okay makes sense, can you place update the document to replace "local development" with something like "integration with IntelliJ/VSCode debugging" and also add a reference to the `runlocal` script and explain that it can be used to run deck and look at it easely?

sbueringer:
> @sbueringer okay makes sense, can you place update the document to replace "local development" with something like "integration with IntelliJ/VSCode debugging" and also add a reference to the `runlocal` script and explain that it can be used to run deck and look at it easely?

@alvaroaleman Yup no problem. I'll update the PR

sbueringer:
> > @sbueringer okay makes sense, can you place update the document to replace "local development" with something like "integration with IntelliJ/VSCode debugging" and also add a reference to the `runlocal` script and explain that it can be used to run deck and look at it easely?
> 
> @alvaroaleman Yup no problem. I'll update the PR

update

PR #17775, opened 2020-05-31
prow: fix GHE pagination

### LINKED ISSUE #17776: Prow: pagination does not work on Github enterprise
**What happened**:
Hook and other Prow components do not work on Github enterprise as soon as they are used in organizations or repos with more than >100 pull requests, repos, etc.. This bug is caused by 
[`readPaginatedResultsWithValues`](https://github.com/kubernetes/test-infra/blob/master/prow/github/client.go#L1350) which doesn't handle pagination correctly for Github enterprise.

That it just works :)

Run e.g. Hook with Github enterprise with a PR with more than 100 files.

**Details**:

The `readPaginatedResultsWithValues` func paginates via `next` links returned from Github. The `next` links always contain the full URL, e.g. `api.github.com/repositories/22/pulls?per_page=100&page=2`. This works as long as the base URL of Github doesn't contain a prefix. In case of Github enterprise the base  URL can look like `<ghe-url>/api/v3`. So our current code leads to calls to URLs like `<ghe-url>/api/v3/api/v3/...` 

Proposed fix: https://github.com/kubernetes/test-infra/pull/17775

sbueringer:
Fixes #17776

### DIFF
--- a/prow/github/client_test.go
+++ b/prow/github/client_test.go
@@ -795,49 +795,87 @@ func TestUnassignIssue(t *testing.T) {
 }
 
 func TestReadPaginatedResults(t *testing.T) {
-	ts := httptest.NewTLSServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		if r.Method != http.MethodGet {
-			t.Errorf("Bad method: %s", r.Method)
-		}
-		if r.URL.Path == "/label/foo" {
-			objects := []Label{{Name: "foo"}}
-			b, err := json.Marshal(objects)
-			if err != nil {
-				t.Fatalf("Didn't expect error: %v", err)
+	type response struct {
+		labels []Label
+		next   string
+	}
+	cases := []struct {
+		name           string
+		baseSuffix     string
+		initialPath    string
+		responses      map[string]response
+		expectedLabels []Label
+	}{
+		{
+			name:        "regular pagination",
+			initialPath: "/label/foo",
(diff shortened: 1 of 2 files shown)


sbueringer:
> Please add some tests for this

Yup, no problem. So I assume the implementation itself is okay like it is?

sbueringer:
@alvaroaleman I added the test

prow/github/client_test.go
```diff
@@ -795,49 +795,88 @@ func TestUnassignIssue(t *testing.T) {
...
+			func(obj interface{}) {
+				labels = append(labels, *(obj.(*[]Label))...)
+			},
+		)
+		if err != nil {
+			t.Errorf("%s: didn't expect error: %v", tc.name, err)
 		} else {
-			t.Errorf("Bad request path: %s", r.URL.Path)
+			var expectedLabels []Label
+			for _, response := range tc.responses {
```
alvaroaleman, line 873: I am not really a fan of a test that determines its outcome at runtime (and might have bugs there) rather than having it hardcoded. I am btw also completely fine with just moving the five relevant lines here into a new function and then just testing that new functions, which is probably easier than adjusting this test.

sbueringer: Makes sense. I added `expectedLabels` to the test case.

I would keep this test as table test if it's okay. I think it's easier to see this way what the actual differences between both supported variants are. It's also easier to extend in the future compared to the old test if necessary.

PR #17778, opened 2020-06-01
Kops - Verify package hashes for containerd and Docker

hakman:
Corresponding PR for https://github.com/kubernetes/kops/pull/9234.

/cc @rifelpet


rifelpet:
i see `optional: true` which is a good way to roll this out. 

/lgtm
/approve

PR #17779, opened 2020-06-01
prow: change job url prefix format for multiple concurrent storage providers

sbueringer:
Currently, it's only possible to use one storage provider per repository. This limitation is caused by `job_url_prefix_config`. The storage provider is suffixed to the job url prefix, e.g. `https://prow.k8s.io/view/gcs/`

This PR solves the issue by deprecating job url prefix with `gcs/` suffix. The storage provider is now appended automatically so a job url prefix like `https://prow.k8s.io/view/` can be used for e.g. `GS` and `S3`. Job url prefix with `gcs/` are still handled for now.

Open Issues:
* how should the old job url prefix be deprecated? I'm handling both formats for now and I added a note to the announcements. Should I add some warnings? If yes, in which component?

xref:
* issue was noticed in the [deck s3 support PR](https://github.com/kubernetes/test-infra/pull/17183#discussion_r421700277)
* s3 support issue: #11260

### DIFF
--- a/prow/ANNOUNCEMENTS.md
+++ b/prow/ANNOUNCEMENTS.md
@@ -3,6 +3,11 @@
 ## New features
 
 New features added to each component:
+ - *July 13th, 2020* Configuring `job_url_prefix_config` with `gcs/` prefix is now deprecated. 
+    Please configure a job url prefix without the `gcs/` storage provider suffix. From now on the storage
+    provider is appended automatically so multiple storage providers can be used for builds of 
+    the same repository. For now we still handle the old configuration format, this will be removed 
+    in *September 2020*. To be clear handling of URLs with `/view/gcs` in Deck is not deprecated.
  - *June 23rd, 2020* An [hmac](/prow/cmd/hmac) tool was added to automatically reconcile webhooks and hmac
     tokens for the orgs and repos integrated with your prow instance.
  - *June 8th, 2020* A new informer-based Plank implementation was added. It can be used by deploying
--- a/prow/config/config.go
+++ b/prow/config/config.go
@@ -482,6 +482,10 @@ type Plank struct {
 	// JobURLPrefixConfig is the host and path prefix under which job details
 	// will be viewable. Use `org/repo`, `org` or `*`as key and an url as value
 	JobURLPrefixConfig map[string]string `json:"job_url_prefix_config,omitempty"`
+
+	// JobURLPrefixDisableAppendStorageProvider disables that the storageProvider is
+	// automatically appended to the JobURLPrefix
+	JobURLPrefixDisableAppendStorageProvider bool
 }
 
 func (p Plank) GetDefaultDecorationConfigs(repo string) *prowapi.DecorationConfig {
@@ -496,7 +500,20 @@ func (p Plank) GetDefaultDecorationConfigs(repo string) *prowapi.DecorationConfi
 	return def
 }
 
+// GetJobURLPrefix gets the job url prefix from the config
+// for the given refs. As we're deprecating the "gcs/" suffix
+// (to allow using multiple storageProviders within a repo)
+// we always trim the suffix here. Thus, every caller can assume
+// the job url prefix does not have a storageProvider suffix.
 func (p Plank) GetJobURLPrefix(refs *prowapi.Refs) string {
+	jobURLPrefix := p.getJobURLPrefix(refs)
+	if strings.HasSuffix(jobURLPrefix, "gcs/") {
+		return strings.TrimSuffix(jobURLPrefix, "gcs/")
+	}
+	return strings.TrimSuffix(jobURLPrefix, "gcs")
+}
+
+func (p Plank) getJobURLPrefix(refs *prowapi.Refs) string {
 	if refs == nil {
 		return p.JobURLPrefixConfig["*"]
 	}
@@ -1193,6 +1210,14 @@ func (c *Config) validateComponentConfig() error {
 	for k, v := range c.Plank.JobURLPrefixConfig {
 		if _, err := url.Parse(v); err != nil {
 			return fmt.Errorf(`Invalid value for Planks job_url_prefix_config["%s"]: %v`, k, err)
+		}
+		// TODO(@sbueringer): Remove in September 2020
+		if strings.HasSuffix(v, "gcs/") {
+			logrus.Warning(strings.Join([]string{
+				"configuring the 'gcs/' storage provider suffix in the job url prefix is now deprecated, ",
+				"please configure the job url prefix without the suffix as it's now appended automatically. Handling of the old ",
+				"configuration will be removed in September 2020",
+			}, ""))
 		}
 	}
 
--- a/prow/crier/reporters/pubsub/reporter.go
+++ b/prow/crier/reporters/pubsub/reporter.go
@@ -28,6 +28,8 @@ import (
 
 	prowapi "k8s.io/test-infra/prow/apis/prowjobs/v1"
 	"k8s.io/test-infra/prow/config"
+	"k8s.io/test-infra/prow/io/providers"
+	"k8s.io/test-infra/prow/spyglass/api"
 )
 
 const (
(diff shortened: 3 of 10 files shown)


sbueringer:
/assign @fejta 
/assign @alvaroaleman 

I'm not sure about the unit tests in the pubsub reporter of crier. The values looked pretty artificial to me. Depending on the use cases which should be supported by crier I broke something.

sbueringer:
> /assign @fejta
> /assign @alvaroaleman
> 
> I'm not sure about the unit tests in the pubsub reporter of crier. The values looked pretty artificial to me. Depending on the use cases which should be supported by crier I broke something. (EDIT: I think I got it now, working on it...)

Okay so tests should be green now. I'm pretty unsure about that Gubernator stuff.

prow/ANNOUNCEMENTS.md
```diff
@@ -3,7 +3,11 @@
 ## New features
 
 New features added to each component:
- - *May 31, 2020* '--gcs-no-auth' in Deck is deprecated and not used anymore. We always
+ - *June 1, 2020* Using `job_url_prefix_config` with `gcs/` prefix is now deprecated. 
+    Please use a job url prefix without a storage provider suffix. From now on the storage
+    provider is appended automatically so multiple storage providers can be used for builds of 
+    the same repository. Handling of the old format will be removed in *September 2020*.
```
alvaroaleman, line 9: No, this is way too short. Not sure if we can remove it all, because we need to keep it for at least as long as the retention period of logs for k8s/istio/openshift, because removing this will break old links (which doesn't mater if they are broken anyways). For us, the retention period of logs is three months. @fejta how long are logs kept for kube tests?

prow/config/config.go
```diff
@@ -496,17 +496,22 @@ func (p Plank) GetDefaultDecorationConfigs(repo string) *prowapi.DecorationConfi
...
 
+// GetJobURLPrefix gets the job url prefix from the config
+// for the given refs. As we're deprecating the "gcs/" suffix
+// (to allow using multiple storageProviders within a repo)
+// we always trim the suffix here. Thus, every caller can assume
+// the job url prefix does not have a storageProvider suffix.
 func (p Plank) GetJobURLPrefix(refs *prowapi.Refs) string {
 	if refs == nil {
-		return p.JobURLPrefixConfig["*"]
+		return strings.TrimSuffix(p.JobURLPrefixConfig["*"], "gcs/")
```
alvaroaleman, line 506: Somewhere in `prow/config.Load()` we need to check if such a prefix is present and if yes, emit a warning

prow/crier/reporters/pubsub/reporter.go
```diff
@@ -130,13 +130,36 @@ func (c *Client) generateMessageFromPJ(pj *prowapi.ProwJob) *ReportMessage {
...
+		// * storageURLPath: gs/kubernetes-jenkins/logs/ci-benchmark-microbenchmarks/1258197944759226371
+		prefix := c.config().Plank.GetJobURLPrefix(pj.Spec.Refs)
+
+		storageURLPath := strings.TrimPrefix(pj.Status.URL, prefix)
+		if strings.HasPrefix(storageURLPath, api.GCSKeyType) {
+			storageURLPath = strings.Replace(storageURLPath, api.GCSKeyType, providers.GS, 1)
+		}
+
+		if providers.HasStorageProviderPrefix(storageURLPath) {
+			storagePathSegments := strings.SplitN(storageURLPath, "/", 2)
```
alvaroaleman, line 148: So if the path is just "gs" or "s3" we will panic? Can we add a lencheck here? Shouldn't happen ofc but shouldn't kill the whole controller if it does either

prow/pjutil/pjutil_test.go
```diff
@@ -750,12 +750,12 @@ func TestJobURL(t *testing.T) {
...
 				}},
 			}},
-			expected: "https://gubernator.com/build/bucket/pr-logs/pull/org_repo/1",
+			expected: "https://gubernator.com/build/gs/bucket/pr-logs/pull/org_repo/1",
 		},
 		{
-			name: "decorated job with prefix uses gcsupload and new bucket format with gcs",
+			name: "decorated job with prefix uses gcsupload and new bucket format with gcs (deprecated job url format)",
 			plank: config.Plank{
-				JobURLPrefixConfig: map[string]string{"*": "https://prow.k8s.io/view/gcs"},
```
alvaroaleman, line 758: Can we keep a testcase without the trailing slash?

sbueringer: Yup I'll add another. Although I'm pretty sure it doesn't actually work if the trailing slash is missing I had problems a while ago when I forgot it. But maybe that was just a strange s3 fork version of mine :)

sbueringer: Okay and of course that was the case that was broken because of TrimSuffix only trimmed "gcs/" not "gcs"

prow/ANNOUNCEMENTS.md
alvaroaleman: Ah okay, then the grace period is fine but yeah, needs clarification :)

prow/config/config.go
sbueringer: Done

prow/crier/reporters/pubsub/reporter.go
sbueringer: Done

prow/pjutil/pjutil_test.go
```diff
@@ -750,10 +750,10 @@ func TestJobURL(t *testing.T) {
 					PathStrategy: prowapi.PathStrategyExplicit,
 				}},
 			}},
-			expected: "https://gubernator.com/build/bucket/pr-logs/pull/org_repo/1",
+			expected: "https://gubernator.com/build/gs/bucket/pr-logs/pull/org_repo/1",
```
fejta, line 753: Should we provide a way to turn this off? When linking to spyglass we should prepend the storage provider, but other systems like gubernator may not support it.

AKA `https://gubernator.com/build/bucket/pr-logs/pull/org_repo/1` is still the correct URL here

sbueringer: @fejta *push* :)

sbueringer: Implemented

sbueringer:
@alvaroaleman @fejta please take another look

sbueringer:
> @alvaroaleman @fejta please take another look

friendly reminder :) @fejta @alvaroaleman

prow/config/config.go
```diff
@@ -1194,6 +1211,12 @@ func (c *Config) validateComponentConfig() error {
 		if _, err := url.Parse(v); err != nil {
 			return fmt.Errorf(`Invalid value for Planks job_url_prefix_config["%s"]: %v`, k, err)
 		}
+		// TODO(@sbueringer): Remove in September 2020
+		if strings.HasSuffix(v, "gcs/") {
+			logrus.Warning("configuring the 'gcs/' storage provider suffix in the job url prefix is now deprecated, " +
```
fejta, line 1216: FYI: I'm personally a fan of stirings.join over + for long strings

strings.Join([]string{
    "line 1",
    "line 2",
}, "\n") // or "" instead of "\n" if we just want one long string
```

Comment (sbueringer) [reply prow/config/config.go]: Thx, makes sense, fixed it and rebased onto current master

File: prow/config/config.go [hunk 7]
```diff
@@ -1194,6 +1211,14 @@ func (c *Config) validateComponentConfig() error {
 		if _, err := url.Parse(v); err != nil {
 			return fmt.Errorf(`Invalid value for Planks job_url_prefix_config["%s"]: %v`, k, err)
 		}
+		// TODO(@sbueringer): Remove in September 2020
+		if strings.HasSuffix(v, "gcs/") {
+			logrus.Warning(strings.Join([]string{
+				"configuring the 'gcs/' storage provider suffix in the job url prefix is now deprecated, ",
+				"please configure the job url prefix without the suffix as it's now appended automatically. Handling of the old ",
+				"configuration will be removed in September 2020",
```
cjwagner, line 1219: Removing support for the old URL format would break existing links which doesn't seem reasonable to me. Why would we need to stop supporting the old format in order to support the new one?

sbueringer: Its only about configuration and supporting configuration with /gcs. We don't want to stop handling /gcs requests

PR #17785, opened 2020-06-01
Kops - Make pull-kops-verify-hashes pre-submit mandatory

hakman:
This should be enough to always run the job when a file from `nodeup/pkg/` is changed.

/cc @rifelpet


rifelpet:
i _think_ that when we merge this, any open PR that touches that package will have the job ran and fail because it hasnt been rebased to include the new hack/verify-hashes script. I think thats ok though, I'm guessing there arent too many affected PRs.

hakman:
If anything like that happens, it's an easy fix, so wouldn't worry.

PR #17791, opened 2020-06-01
Skip the periodic SCTP related NetworkPolicy test

janosi:
The GCP environment of the periodic test execution does not support SCTP in NetworkPolicy. So we should skip that testcase, but execute all other SCTP related tests - it makes no sense to test that in such an environment, we know that it will fail.


aojea:
/lgtm
/approve
xref https://k8s-testgrid.appspot.com/sig-network-gce#gci-gce-basic-sctp

PR #17794, opened 2020-06-01
gubernator/README: 'name' and 'value' are attributes, not child elements

wking:
These docs landed in 2f04de098e (#9182), but since we began vendoring the backing XML types in ce50d4c65c (#9887) they've been:

```go
type Property struct {
  Name  string `xml:"name,attr"`
  Value string `xml:"value,attr"`
}
```

The `attr` label gets these stored as attibutes instead of as child elements.

### DIFF
--- a/gubernator/README.md
+++ b/gubernator/README.md
@@ -92,14 +92,8 @@ The properties can be defined as:
 ```xml
 <testcase ...>
   <properties>
-    <property>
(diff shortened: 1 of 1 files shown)


wking:
/remove-lifecycle stale

Bump on a small docs cleanup :)

PR #17795, opened 2020-06-01
Update node e2e OWNERS

spiffxp:
I've noticed an uptick in organized activity on taking ownership of and maintaining the node e2e tests

I'm proposing we drop reviewers that are known to be inactive, and add as reviewers some of the folks who have been putting in effort lately

Also added a sig/node label to jobs/e2e_node while I was there


karan:
Thanks for doing this. Much needed.

MHBauer:
I praise @vpickard. Very good organizer. Happy to accept the responsibility.

dchen1107:
/lgtm
/approve

Thanks for picking up the work.

PR #17799, opened 2020-06-02
Fix wrong override locations

jkaniuk:
Fix wrong kubemark 100 override location
Tests are failing: https://k8s-testgrid.appspot.com/sig-scalability-kubemark#kubemark-100

/cc wojtek-t mm4tt

### DIFF
--- a/config/jobs/kubernetes/sig-scalability/sig-scalability-periodic-jobs.yaml
+++ b/config/jobs/kubernetes/sig-scalability/sig-scalability-periodic-jobs.yaml
@@ -133,7 +133,7 @@ periodics:
       - --test-cmd-args=--testoverrides=./testing/experiments/enable_prometheus_api_responsiveness.yaml
       - --test-cmd-args=--testoverrides=./testing/experiments/enable_restart_count_check.yaml
       - --test-cmd-args=--testoverrides=./testing/experiments/use_simple_latency_query.yaml
-      - --test-cmd-args=--testoverrides=./testing/overrides/kubemark/throughput_override.yaml
+      - --test-cmd-args=--testoverrides=./testing/overrides/kubemark_load_throughput.yaml
       - --test-cmd-name=ClusterLoaderV2
       - --timeout=240m
       - --use-logexporter


config/jobs/kubernetes/sig-scalability/sig-scalability-golang.yaml
```diff
@@ -71,10 +71,10 @@ periodics:
       - --test-cmd-args=--provider=kubemark
       - --test-cmd-args=--report-dir=/workspace/_artifacts
       - --test-cmd-args=--testconfig=testing/load/config.yaml
+      - --test-cmd-args=--testoverrides=./testing/density/5000_nodes/override.yaml
```
wojtek-t, line 74: This one is incorrect - we're moving them out from density, because overrides are actually per the whole job, not per test.
The one passed currently seems correct:
https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/testing/overrides/5000_nodes.yaml

config/jobs/kubernetes/sig-scalability/sig-scalability-periodic-jobs.yaml
```diff
@@ -133,7 +133,7 @@ periodics:
       - --test-cmd-args=--testoverrides=./testing/experiments/enable_prometheus_api_responsiveness.yaml
       - --test-cmd-args=--testoverrides=./testing/experiments/enable_restart_count_check.yaml
       - --test-cmd-args=--testoverrides=./testing/experiments/use_simple_latency_query.yaml
-      - --test-cmd-args=--testoverrides=./testing/overrides/kubemark/throughput_override.yaml
+      - --test-cmd-args=--testoverrides=./testing/overrides/kubemark_100_nodes.yaml
```
wojtek-t, line 136: Only this one is the right one - thanks!

wojtek-t: Actually no - this one is also not exactly what we have - see: https://github.com/kubernetes/perf-tests/pull/1300/files#r433645388

config/jobs/kubernetes/sig-scalability/sig-scalability-golang.yaml
jkaniuk: Problem is different - golang tests are not run from HEAD

config/jobs/kubernetes/sig-scalability/sig-scalability-periodic-jobs.yaml
jkaniuk: Fixed to kubemark_load_throughput.yaml

wojtek-t: Thanks!

config/jobs/kubernetes/sig-scalability/sig-scalability-golang.yaml
wojtek-t: Ahh - I see.

Can you maybe add a comment for these, sth like:
Change the override to ... once golang tests will migrate to 1.19+ branch?

jkaniuk: Done

mm4tt: Actually, it's just got fixed in https://github.com/kubernetes/test-infra/pull/17753, please wait till it gets merged and then most likely revert this change

jkaniuk: Reverted

wojtek-t:
/lgtm
/approve

Thanks!

PR #17800, opened 2020-06-02
run sig-windows tests on windows file change too

andyzhangx:
run following tests on windows file change:
 - pull-kubernetes-e2e-azure-disk-windows
 - pull-kubernetes-e2e-azure-file-windows

per requirements from https://github.com/kubernetes/kubernetes/pull/91468#issuecomment-636257357, let's only run tests on azure or windows related changes, otherwise it may exceed Azure testing quota easily.

/assign @chewong @liggitt @jingxu97 
The regex in this PR should be correct, right?


chewong:
/hold
for additional reviewers. Feel free to unhold if this needs to be merged ASAP @andyzhangx

PR #17801, opened 2020-06-02
Enable containerd for k/k 500 node presubmit

jprzychodzen:
Continue rollout for kubernetes/perf-tests#1110

/sig scalability
/assign @mm4tt

Rollout containerd tests following [experiments procedure](https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/docs/experiments.md).

### DIFF
--- a/config/jobs/kubernetes/sig-scalability/sig-scalability-presubmit-jobs.yaml
+++ b/config/jobs/kubernetes/sig-scalability/sig-scalability-presubmit-jobs.yaml
@@ -66,6 +66,7 @@ presubmits:
       preset-k8s-ssh: "true"
       preset-bazel-scratch-dir: "true"
       preset-e2e-scalability-common: "true"
+      preset-e2e-scalability-containerd: "true"
       preset-e2e-scalability-presubmits: "true"
     spec:
       containers:


config/jobs/kubernetes/sig-scalability/sig-scalability-presubmit-jobs.yaml
```diff
@@ -66,6 +66,7 @@ presubmits:
       preset-k8s-ssh: "true"
       preset-bazel-scratch-dir: "true"
       preset-e2e-scalability-common: "true"
+      preset-e2e-scalability-containerd: "true"
```
wojtek-t, line 69: This is not 5k presubmit - please update PR description.

wojtek-t: It's also optional, so we won't get any gain from it...

jprzychodzen: > This is not 5k presubmit - please update PR description.

Done.

> It's also optional, so we won't get any gain from it...

I'm not sure what do you mean by that,

Are you suggesting that this preset should not be rolled out to this test? Or maybe that we can safely remove this presubmit?

wojtek-t: I'm saying "rolling this out has zero value, because noone is using it".

wojtek-t: [I'm not saying "don't do that" - I'm fine with that. I'm saying that this is effectively "no-op".]

jprzychodzen: Sure, thanks for explanation.

Then I would like to merge this change to keep this config in line with other configs. If there will be an event that some will use this presubmit, then it should be configured in an expected way.

wojtek-t:
@jprzychodzen - please fix the commit message also

jprzychodzen:
Fixed commit message

PR #17812, opened 2020-06-02
Add new job to run sig-storage tests with SELinux enabled

bertinatto:
We want a periodic job that runs sig-storage tests on a host with SELinux enabled.

@hakman could you take a look at this, please?

CC @msau42 @jsafrane

### DIFF
--- a/config/jobs/kubernetes/sig-storage/sig-storage-kops-config.yaml
+++ b/config/jobs/kubernetes/sig-storage/sig-storage-kops-config.yaml
@@ -1,0 +1,35 @@
+periodics:
+- interval: 24h
+  name: ci-kubernetes-e2e-storage-selinux
+  labels:
+    preset-service-account: "true"
+    preset-aws-ssh: "true"
+    preset-aws-credential: "true"
+  decorate: true
+  decoration_config:
+    timeout: 90m
+  spec:
+    containers:
+    - command:
+      - runner.sh
+      - /workspace/scenarios/kubernetes_e2e.py
+      args:
+      - --cluster=e2e-kops-aws-k8s-storage-selinux.test-cncf-aws.k8s.io
+      - --deployment=kops
+      - --env=KUBE_SSH_USER=centos
+      - --env=KOPS_RUN_TOO_NEW_VERSION=1
+      - --extract=ci/latest
+      - --ginkgo-parallel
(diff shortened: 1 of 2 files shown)


config/jobs/kubernetes/sig-storage/sig-storage-kops-config.yaml
```diff
@@ -0,0 +1,35 @@
...
+      - --env=KUBE_SSH_USER=centos
+      - --env=KOPS_RUN_TOO_NEW_VERSION=1
+      - --extract=ci/latest
+      - --ginkgo-parallel
+      - --kops-image=679593333241/CentOS Linux 7 x86_64 HVM EBS ENA 2002_01-b7ee8a69-ee97-4a49-9e68-afaee216db2e-ami-0042af67f8e4dcc20.4
+      - --kops-priority-path=/workspace/kubernetes/platforms/linux/amd64
+      - --kops-ssh-user=centos
+      - --kops-version=https://storage.googleapis.com/kops-ci/bin/latest-ci-updown-green.txt
+      - --provider=aws
+      - --test_args=--ginkgo.focus=\[sig-storage\].*\[Slow\] --ginkgo.skip=\[Serial\]|\[Disruptive\]|\[Flaky\]|\[Feature:.+\] --minStartupPods=8
```
jsafrane, line 28: This runs only `[Slow]` tests. IMO, `[sig-storage]` should be enough.

jsafrane, line 28: It would be nice to include also SELinux tests: https://github.com/kubernetes/kubernetes/blob/2c8955fd4a26ccc799d6093f555c31d7d73cce3c/test/e2e/node/security_context.go#L137-L147

We should deflake them as part of the SELinux effort.

hakman: Most likely, the SELinux tests will fail for now, until we get a flag for the feature gate in Kops.

hakman:
@bertinatto Looks OK to me.

config/jobs/kubernetes/sig-storage/sig-storage-kops-config.yaml
bertinatto: Removed `[Slow]`

jsafrane: SELinux does not have a feature gate, it has been in Kubernetes since v1.0. I want to rework it a lot in 1.19, and for that I will have a feature gate there.

Talked to @bertinatto offline, we can add the SELinux test later, focusing on the storage tests on CentOS first.

PR #17818, opened 2020-06-03
Tag KUBE_CONTAINER_RUNTIME=docker explicitly

dims:
If we ever have to switch default runtimes in kube-up scripts, we need
to mark jobs explicitly. Let us start with the presubmit for sig-node
that run the node e2e tests. These tests depend on docker explicitly.
Note that there are specific containerd jobs that have a preset for
containerd as well if they explicitly need containerd.

### DIFF
--- a/config/jobs/kubernetes/sig-node/sig-node-presubmit.yaml
+++ b/config/jobs/kubernetes/sig-node/sig-node-presubmit.yaml
@@ -30,6 +30,7 @@ presubmits:
         - --gcp-zone=us-west1-b
         - --node-test-args=--kubelet-flags="--cgroups-per-qos=true --cgroup-root=/"
         - --node-tests=true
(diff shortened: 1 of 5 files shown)


MHBauer:
Where does the node-e2e tests read this?

dims:
@MHBauer kubernetes_e2e.py -> kubetest -> hack/ginkgo-e2e.sh is roughly the flow. 

Within kubetest, if you see bash.go, there's a call to `./hack/e2e-internal/e2e-up.sh` which eventually ends up calling cluster/kube-up.sh. that's where this env var is evaluated/used. 

Best way to see this for yourself throw in a WIP PR with `set -x` in cluster/kube-up.sh at the top of the file, and you will see all the calls happening in the e2e CI jobs (non-kind)

dims:
hmm, need to think this through a bit more. will file later.

PR #17823, opened 2020-06-03
Introduce a github event server with a http handler.

droslean:
Introduce the `githubeventserver` package which contains:

- Options with `endpoint` and `port` to allow components to bind them.
- Registration methods to allow any of the component to register its own event handler functions 
- Hook metrics and the corresponding tests have been moved to `githubeventserver` package.

**Follow-ups** PRs to change the existing external plugins and hook to use this package as well.

### DIFF
--- a/prow/cmd/hook/main.go
+++ b/prow/cmd/hook/main.go
@@ -31,6 +31,7 @@ import (
 	"k8s.io/test-infra/prow/config/secret"
 	prowflagutil "k8s.io/test-infra/prow/flagutil"
 	"k8s.io/test-infra/prow/git/v2"
+	"k8s.io/test-infra/prow/githubeventserver"
 	"k8s.io/test-infra/prow/hook"
 	"k8s.io/test-infra/prow/interrupts"
 	"k8s.io/test-infra/prow/logrusutil"
@@ -194,7 +195,7 @@ func main() {
 		BugzillaClient:            bugzillaClient,
 	}
 
-	promMetrics := hook.NewMetrics()
+	promMetrics := githubeventserver.NewMetrics()
 
 	defer interrupts.WaitForGracefulShutdown()
 
--- a/prow/githubeventserver/githubeventserver.go
+++ b/prow/githubeventserver/githubeventserver.go
@@ -1,0 +1,495 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package githubeventserver
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"strconv"
+	"sync"
+	"time"
+
+	"github.com/sirupsen/logrus"
+
+	"k8s.io/test-infra/prow/config"
+	"k8s.io/test-infra/prow/github"
+	"k8s.io/test-infra/prow/interrupts"
+	"k8s.io/test-infra/prow/pluginhelp"
+	pluginhelp_externalplugins "k8s.io/test-infra/prow/pluginhelp/externalplugins"
+	pluginhelp_hook "k8s.io/test-infra/prow/pluginhelp/hook"
+	"k8s.io/test-infra/prow/plugins"
+)
+
+const (
+	eventTypeField = "event-type"
+
+	statusEvent                   = "status"
+	pushEvent                     = "push"
+	pullRequestReviewCommentEvent = "pull_request_review_comment"
+	pullRequestReviewEvent        = "pull_request_review"
+	pullRequestEvent              = "pull_request"
+	issueCommentEvent             = "issue_comment"
+	issuesEvent                   = "issues"
+)
+
+// GitHubEventServer hold all the information needed for the
+// github event server implementation.
+type GitHubEventServer struct {
+	wg *sync.WaitGroup
+
+	endpoint string
+	port     int
+
+	serveMuxHandler *serveMuxHandler
+	httpServeMux    *http.ServeMux
+}
+
+// NewGitHubEventServer creates a new *GitHubEventServer from the given arguments.
+// It also assigns the serveMuxHandler in the http.ServeMux.
+func NewGitHubEventServer(o Options) *GitHubEventServer {
+	var wg sync.WaitGroup
+
+	httpServeMux := http.NewServeMux()
+	githubEventServer := &GitHubEventServer{
+		endpoint: o.endpoint,
+		port:     o.port,
+		wg:       &wg,
+		serveMuxHandler: &serveMuxHandler{
+			hmacTokenGenerator: o.HmacTokenGenerator,
+			log:                o.Logger,
+			metrics:            o.Metrics,
+			wg:                 &wg,
+		},
+	}
+
+	httpServeMux.Handle(o.endpoint, githubEventServer.serveMuxHandler)
+	githubEventServer.httpServeMux = httpServeMux
+
+	return githubEventServer
+}
+
+// GetHTTPServeMux returns the http.ServeMux of GitHubEventServerOptions
+func (g *GitHubEventServer) GetHTTPServeMux() *http.ServeMux {
+	return g.httpServeMux
+}
+
+// Listen runs an http server by using the existing http.ServeMux, port and the
+// given grace period duration
+func (g *GitHubEventServer) Listen(gracePeriod time.Duration) {
+	defer interrupts.WaitForGracefulShutdown()
+	interrupts.ListenAndServe(&http.Server{Addr: ":" + strconv.Itoa(g.port), Handler: g.httpServeMux}, gracePeriod)
+}
+
+// ReviewCommentEventHandler is a type of function that handles GitHub's review comment events
+type ReviewCommentEventHandler func(*logrus.Entry, github.ReviewCommentEvent)
+
+// ReviewEventHandler is a type of function that handles GitHub's review events.
+type ReviewEventHandler func(*logrus.Entry, github.ReviewEvent)
+
+// PushEventHandler is a type of function that handles GitHub's push events.
+type PushEventHandler func(*logrus.Entry, github.PushEvent)
+
+// IssueCommentEventHandler is a type of function that handles GitHub's issue comment events.
+type IssueCommentEventHandler func(*logrus.Entry, github.IssueCommentEvent)
+
+// PullRequestHandler is a type of function that handles GitHub's pull request events.
+type PullRequestHandler func(*logrus.Entry, github.PullRequestEvent)
+
+// IssueEventHandler is a type of function that handles GitHub's issue events.
+type IssueEventHandler func(*logrus.Entry, github.IssueEvent)
+
+// StatusEventHandler is a type of function that handles GitHub's status events.
+type StatusEventHandler func(*logrus.Entry, github.StatusEvent)
+
+// RegisterReviewCommentEventHandler registers an ReviewCommentEventHandler function in GitHubEventServerOptions
+func (g *GitHubEventServer) RegisterReviewCommentEventHandler(fn ReviewCommentEventHandler) {
+	g.serveMuxHandler.reviewCommentEventHandlers = append(g.serveMuxHandler.reviewCommentEventHandlers, fn)
+}
+
+// RegisterReviewEventHandler registers an ReviewEventHandler function in GitHubEventServerOptions
+func (g *GitHubEventServer) RegisterReviewEventHandler(fn ReviewEventHandler) {
+	g.serveMuxHandler.reviewEventHandlers = append(g.serveMuxHandler.reviewEventHandlers, fn)
+}
+
+// RegisterPushEventHandler registers an PushEventHandler function in GitHubEventServerOptions
+func (g *GitHubEventServer) RegisterPushEventHandler(fn PushEventHandler) {
+	g.serveMuxHandler.pushEventHandlers = append(g.serveMuxHandler.pushEventHandlers, fn)
+}
+
+// RegisterHandleIssueCommentEvent registers an IssueCommentEventHandler function in GitHubEventServerOptions
+func (g *GitHubEventServer) RegisterHandleIssueCommentEvent(fn IssueCommentEventHandler) {
+	g.serveMuxHandler.issueCommentEventHandlers = append(g.serveMuxHandler.issueCommentEventHandlers, fn)
+}
+
+// RegisterHandlePullRequestEvent registers an PullRequestHandler function in GitHubEventServerOptions
+func (g *GitHubEventServer) RegisterHandlePullRequestEvent(fn PullRequestHandler) {
+	g.serveMuxHandler.pullRequestHandlers = append(g.serveMuxHandler.pullRequestHandlers, fn)
+}
+
+// RegisterIssueEventHandler registers an IssueEventHandler function in GitHubEventServerOptions
+func (g *GitHubEventServer) RegisterIssueEventHandler(fn IssueEventHandler) {
+	g.serveMuxHandler.issueEventHandlers = append(g.serveMuxHandler.issueEventHandlers, fn)
+}
+
+// RegisterStatusEventHandler registers an StatusEventHandler function in GitHubEventServerOptions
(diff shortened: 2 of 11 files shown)


alvaroaleman:
@droslean motivation/goal?

droslean:
@alvaroaleman Removing duplicating code from the external plugins, and give the ability to use this code to anyone who wants to create a new external plugin.

prow/external-plugins/cherrypicker/server.go
```diff
@@ -105,19 +104,6 @@ type Server struct {
...
-		return
-	}
-	fmt.Fprint(w, "Event received. Have a nice day.")
-
-	if err := s.handleEvent(eventType, eventGUID, payload); err != nil {
-		logrus.WithError(err).Error("Error parsing event.")
-	}
-}
-
 func (s *Server) handleEvent(eventType, eventGUID string, payload []byte) error {
```
stevekuznetsov, line 107: I think all of this functionality should be shared, too

prow/external-plugins/needs-rebase/main.go
```diff
@@ -121,7 +125,7 @@ func main() {
 	}, o.updatePeriod)
 
 	mux := http.NewServeMux()
-	mux.Handle("/", server)
+	mux.Handle("/", handler)
 	externalplugins.ServeExternalPluginHelp(mux, log, plugin.HelpProvider)
```
stevekuznetsov, line 129: All of this is entirely shared, too -- a user ideally could create an external plugin and provide only the bits that are different

prow/external-plugins/cherrypicker/server.go
droslean: This func is different in each plugin. That's why we can pass this func in the handler.

prow/external-plugins/needs-rebase/main.go
droslean: We can add the ability to pass only the `HelpProvider` func in the handler.

prow/external-plugins/cherrypicker/server.go
droslean: Meaning that if we force the plugins to manage only the `issue_comment` events.

petr-muller: Maybe we could have an interface with different `HandleEventX` methods, and plugins that do not care about certain events would have nop implementations of the methods they don't care about?

droslean: Handling issues and pull requests should be enough. I don't like the idea of creating an interface and forcing the plugins to implement empty functions if they don't need them. A nil check in the handler should be sufficient.

droslean:
`needs-rebase` refactoring should be done in a follow-up PR since it implements its own fake GitHub client instead of https://godoc.org/k8s.io/test-infra/prow/github/fakegithub

droslean:
/hold

Holding for reviews and manually testing of hook

droslean:
@stevekuznetsov @alvaroaleman This should be ready for review.

prow/cmd/hook/main.go
```diff
@@ -193,41 +194,45 @@ func main() {
...
 		Plugins:        pluginAgent,
 		Metrics:        promMetrics,
 		TokenGenerator: secretAgent.GetTokenGenerator(o.webhookSecretFile),
 	}
+
+	logger := logrus.NewEntry(logrus.New())
+	handler := servemux.NewServeMuxHandler(server, secretAgent.GetTokenGenerator(o.webhookSecretFile), logger)
+
 	interrupts.OnInterrupt(func() {
-		server.GracefulShutdown()
```
stevekuznetsov, line 212: nit: don't rename a bunch of variables at once, it just makes your diff harder to read

prow/cmd/hook/main.go
```diff
@@ -193,41 +194,45 @@ func main() {
...
+		handler.GracefulShutdown()
 		if err := gitClient.Clean(); err != nil {
 			logrus.WithError(err).Error("Could not clean up git client cache.")
 		}
 	})
 
-	health := pjutil.NewHealth()
+	mux := servemux.InitializeServeMux(logger, "/hook", handler, nil)
```
stevekuznetsov, line 224: This is still messy. Why do I need to create a new handler, then initialize it with itself and a `/hook` string?

I want to see a library that takes an exported `Options` struct (which knows how to add itself to a `FlagSet`) and does _all_ the initalization with it. The `main()` of `hook` and `needs-rebase` for instance should be 100% identical in these regards. Adding a new external plugin should consist of:

1. importing and using the `Options`
2. adding on external items
3. registering plugins

config/prow/cluster/hook_deployment.yaml
```diff
@@ -48,6 +48,7 @@ spec:
         - --github-token-path=/etc/github/oauth
         - --config-path=/etc/config/config.yaml
         - --job-config-path=/etc/job-config
+        - --endpoint=/hook
```
alvaroaleman, line 51: I don't think we want to make this configurable and this is a backwards-incompatible change

droslean: I was thinking about that. Maybe it's better to default the flag to "/hook" and change it in the external plugins in the follow-up refactors

prow/servemux/servemux.go
```diff
@@ -0,0 +1,224 @@
...
+	"k8s.io/test-infra/prow/pluginhelp"
+	"k8s.io/test-infra/prow/pluginhelp/externalplugins"
+	"k8s.io/test-infra/prow/plugins"
+)
+
+const (
+	eventTypeField = "event-type"
+)
+
+type EventHandler interface {
```
alvaroaleman, line 41: I think the point Steve was making earlier is that we should not export an `EventHandler`, we should support a `plugin.Registry` because the `EventHandler` is just an implementation detail that doesn't really matter for plugin authors:
```
type PluginRegistry interface {
  RegisterGenericCommentHandler(name string, fn GenericCommentHandler, help HelpProvider) error
  RegisterReviewCommentEventHandler(name string, fn ReviewCommentEventHandler, help HelpProvider) error
  ...
}

func NewGitHubEventHandler(listenAddress string) EventHandler {

}
```

Furthermore, I don't think `servermux` is a good name for the package. A `ServerMux` is a generic thing for any kind of http server, but what we are building here is not, its specific for handling GithUb events.

From a reviewer perspective I would btw prefer it if you first did only the new package and not adjust hook. Right now I have to skim over 300 LOC hook changes that are irrelevant until we agree on how this here looks like

prow/servemux/servemux.go
```diff
@@ -0,0 +1,399 @@
+/*
```
alvaroaleman, line 1: The factoring looks much better now :)
Discovering this via godoc is still hard though:
```
$ go doc ./prow/servemux/


package servemux // import "k8s.io/test-infra/prow/servemux"

type Metrics struct{ ... }
    func NewMetrics() *Metrics
type PluginMetrics struct{ ... }
type ServeMuxHandlerOptions struct{ ... }
    func NewFakeServeMuxHandlerOptions() *ServeMuxHandlerOption
```

I think it would be better if we had:
```
type ServeMuxHandlerOptions struct{ ... }

type GitHubEventHandlerServer interface {
 RegisterHandleIssueCommentEvent
 RegisterHandlePullRequestEvent
 ...
}

func New(hmacTokenGenerator func() []byte, log *logrus.Entry, bindAddress string) GitHubEventHandlerServer {

}
```

Also, please rename the package. This is not a generic `servemux`. This is a very specific thing to handle github events. Maybe `githubeventserver`?

prow/servemux/servemux.go
```diff
@@ -0,0 +1,399 @@
...
+
+func (o *ServeMuxHandlerOptions) Initialize(hmacTokenGenerator func() []byte, log *logrus.Entry) {
+	var wg sync.WaitGroup
+
+	o.wg = &wg
+	o.httpServeMux = http.NewServeMux()
+	o.serveMuxHandler = &serveMuxHandler{
+		hmacTokenGenerator: hmacTokenGenerator,
+		log:                log,
+		metrics:            NewMetrics(),
```
alvaroaleman, line 77: Why is `NewMetrics` exported if it can not be set?

prow/servemux/servemux.go
```diff
@@ -0,0 +1,399 @@
...
+func (o *ServeMuxHandlerOptions) GetHTTPServeMux() *http.ServeMux {
+	return o.httpServeMux
+}
+
+func (o *ServeMuxHandlerOptions) Listen(gracePeriod time.Duration) {
+	defer interrupts.WaitForGracefulShutdown()
+	interrupts.ListenAndServe(&http.Server{Addr: ":" + strconv.Itoa(o.port), Handler: o.httpServeMux}, gracePeriod)
+}
+
+type reviewCommentEventHandler func(*logrus.Entry, github.ReviewCommentEvent, *sync.WaitGroup)
```
alvaroaleman, line 101: All of these funciton types must be exported. Right now, the servermux requires an argument of an unexported type which is annoying to use, because you have to go into the package to check the signature

prow/servemux/servemux.go
```diff
@@ -0,0 +1,399 @@
...
+		if err := json.Unmarshal(payload, &i); err != nil {
+			return err
+		}
+		i.GUID = eventGUID
+		org = i.Repo.Owner.Login
+		repo = i.Repo.Name
+
+		for _, issueEventHandler := range s.issueEventHandlers {
+			s.wg.Add(1)
+			go issueEventHandler(l, i, s.wg)
```
alvaroaleman, line 207: Instead of passing in the workgroup everywhere and relying on consumers to correctly call `wg.Done`, why do we not simply wrap it:
```
go func(){
  defer wg.Done()
  issueEventhandler(l, i)
}
```

prow/hook/server.go
```diff
@@ -17,238 +17,433 @@ limitations under the License.
...
-	ConfigAgent    *config.Agent
-	TokenGenerator func() []byte
-	Metrics        *Metrics
-
-	// c is an http client used for dispatching events
-	// to external plugin services.
-	c http.Client
-	// Tracks running handlers for graceful shutdown
-	wg sync.WaitGroup
+	ClientAgent *plugins.ClientAgent
```
alvaroaleman, line 73: Why does hook not use the `servemux` server and has its own instead?

prow/servemux/servemux_test.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2016 The Kubernetes Authors.
+Copyright 2020 The Kubernetes Authors.
```
alvaroaleman, line 2: I would like to keep these tests where they are to prove that we did not break hook

prow/servemux/servemux.go
droslean: other components can use these metrics in their handle event funcitons

prow/servemux/servemux.go
droslean: that sounds better :)

prow/hook/server.go
droslean: oops. that was a leftover of something i was trying to check. I will remove it. Thanks.

prow/servemux/servemux_test.go
droslean: Btw, this is a new file. No idea why GitHub is detecting a difference here :/

prow/servemux/servemux.go
droslean: However, on a second look, the consumers will spawn different goroutines and in case of interruption, we will kill all of those.

prow/servemux/servemux.go
droslean: I think it's better not to have a `New` function since this is going to be created with the options.

prow/hook/server.go
droslean: I deleted the previous comment since I was referring to a different thing.  So for this one, the `servemux` server shouldn't care about having this.

prow/servemux/servemux.go
alvaroaleman: I disagree. Right now what is called `GitHubEventServerOptions` is the actual implementation of the server and this is super confusing. Please make seperate the `Options` from the actual thing, export an Interface `GitHubEventServer` that holds everything that is currently on the `GitHubEventServerOptions` except for `Bind` and `Validate` and add a `New` that returns this interface.

`hook` for example doesn't need the flags.

prow/servemux/servemux.go
alvaroaleman: Where do they spawn different goroutines? They just register a plugin for an event type and we invoke that plugin for the given event, the plugin itself is not supposed to start new goroutines? And if it does it should just have its own waitGroup and not return until that is done?

droslean: You are right. I think its better to let the consumer handles its own waitgroups. Let me address this.

prow/hook/server.go
```diff
@@ -17,238 +17,430 @@ limitations under the License.
...
 }
 
-// dispatch creates a new request using the provided payload and headers
-// and dispatches the request to the provided endpoint.
-func (s *Server) dispatch(endpoint string, payload []byte, h http.Header) error {
-	req, err := http.NewRequest(http.MethodPost, endpoint, bytes.NewBuffer(payload))
-	if err != nil {
-		return err
+func (s *Server) HandleIssueEvent(l *logrus.Entry, i github.IssueEvent) {
+	l = l.WithFields(logrus.Fields{
```
alvaroaleman, line 270: Shouldn't we set these log fields (and in all other `Handle` funcs) in the `GitHubEventServer`?

prow/cmd/hook/main.go
```diff
@@ -44,18 +44,17 @@ import (
 )
 
 type options struct {
-	port int
```
alvaroaleman, line 47: Why did we remove this? It might break ppl

droslean: I won't since the `--port` flag still exists through the `githubeventserver.Options`

prow/hook/server.go
droslean: Indeed.

droslean: Addressed

droslean:
I updated this PR to not make any changes in hook for now. I will open a follow-up PR for that.

prow/githubeventserver/githubeventserver.go
```diff
@@ -0,0 +1,495 @@
...
+					"context":           se.Context,
+					"sha":               se.SHA,
+					"state":             se.State,
+					"id":                se.ID,
+				}), se)
+			}()
+		}
+
+	default:
+		l.Debug("Ignoring unhandled event type.")
```
alvaroaleman, line 390: Nit: Log the event type?

prow/githubeventserver/options.go
```diff
@@ -0,0 +1,46 @@
...
+
+import (
+	"flag"
+	"fmt"
+	"strings"
+)
+
+// Options holds all the information needed to create a
+// github event server handler.
+type Options struct {
```
alvaroaleman, line 27: The purpose of this is not really clear anymore, I would have expected this to be the only argument to `NewGitHubEventServer`

prow/githubeventserver/githubeventserver.go
droslean: This logger already holds the event type in its fields. See line 209

prow/githubeventserver/options.go
droslean: Addressed. These options are now required in `NewGitHubEventServer`

prow/githubeventserver/githubeventserver.go
```diff
@@ -0,0 +1,495 @@
...
+	endpoint string
+	port     int
+
+	serveMuxHandler *serveMuxHandler
+	httpServeMux    *http.ServeMux
+}
+
+// NewGitHubEventServer creates a new *GitHubEventServer from the given arguments.
+// It also assigns the serveMuxHandler in the http.ServeMux.
+func NewGitHubEventServer(hmacTokenGenerator func() []byte, options Options, metrics *Metrics, log *logrus.Entry) *GitHubEventServer {
```
alvaroaleman, line 66: I think the options should contain all possible options

droslean: Well, the server only needs the default endpoint and port. What other possible options you mean?

alvaroaleman: `hmacTokenGenerator`, `metrics`, `log`. It is not logical to have an options struct that holds two out of five mandatory options

droslean: Yea I guess, the consumer can just update the options after flags passed.

droslean: Also, I didn't want to have unnecessary exposed fields in Options struct. So I think passing them in the `NewGitHubEventServer` without messing around with the Options, should be sufficient. WDYT?

alvaroaleman: No. Either we don't have no options struct at all or it holds at all options. A middle ground is just super confusing.

droslean: Makes sense. Let me address this.

droslean: Addressed. Maybe it's better to add some helper functions to `Set/Get` instead of exposing them.

PR #17828, opened 2020-06-03
add note about setting $HOME=/root for buildx

aramase:
We ran into an issue while using `buildx` to build multi-arch images using cloudbuild. After some debugging figured out the `$HOME` dir need to be explicitly set to `/root` for using docker buildx (https://github.com/kubernetes-sigs/secrets-store-csi-driver/pull/219). Adding it to the docs for other users onboarding.


dims:
thanks @aramase ! this is useful tip we ran into before but neglected to doc.

/approve

PR #17833, opened 2020-06-04
Update triage more frequently

spiffxp:
The job has trended downward in duration over the past two weeks

/cc @BenTheElder @dims


spiffxp:

PR #17835, opened 2020-06-04
kops e2e: only publish latest-ci-green from one test

justinsb:
There's a pipeline:

* kops CI passes => write latest-ci.txt
* updown smoketest passes => write latest-ci-updown-green.txt
* e2e passes => write latest-ci-green.txt

Most of our e2e tests rely on latest-ci-green.txt, so that we only
test kops versions that have passed e2e.


justinsb:
Full picture is in #17840

PR #17845, opened 2020-06-04
remove planter from docs/dep.md

b01901143:
planter has been removed according to previous PR: #17254


spiffxp:
Can you update the PR description to include why you're removing the references to planter?

b01901143:
> Can you update the PR description to include why you're removing the references to planter?

edited in description

PR #17846, opened 2020-06-04
Fix updateconfig plugin bug: return nil when it could return error

chaodaiG:
updated could be 0 when all operations in previous for loop failed, thus it should return aggregated error instead of blind nil

/assign @alvaroaleman @matthyx


alvaroaleman:
@chaodaiG can we also append the error in line 340 when there is no client?

chaodaiG:
> @chaodaiG can we also append the error in line 340 when there is no client?

Good catch! Done

PR #17851, opened 2020-06-05
kops gce: try not using specified service account

### LINKED ISSUE #17558: Kops/GCE e2e tests need a new Google Service Account
**What happened**:
Kops GCE e2e tests are broken and have been for quite awhile in prow. In this context, I mean the tests are unable to make it through the `--up` phase, not that they aren't all passing. There's a bunch to unpack here, but basically they don't work because we rely on RW access to storage both during the launch of a cluster and during its operation. 

I raised a [PR](https://github.com/kubernetes/test-infra/pull/17468) on this that unfortunately obfuscates the issues, so I've closed it to begin discussion here.

We raised this during our kops office hours this week and I think we have some clarification about how we can move forward.

**Current/Existing State:** We have one GSA (`pr-kubekins`) that we [bind](https://github.com/kubernetes/test-infra/blob/master/config/prow/cluster/build_serviceaccounts.yaml) to using workload identity in prow. When this account is used to execute our jobs, the GCE nodes get assigned the default project service account (which doesn't necessarily have RW access to any GCS buckets). 

**Proposal:** We would like to learn how to get 1 NEW google service account to be used for the GCE nodes that we spin up (kops-nodes@[hypothetical-project].iam.gserviceaccount.com). We will still need `pr-kubekins` to execute our prow jobs. `pr-kubekins` will also need to have service account actor on `kops-nodes` so that it can instantiate those VMs.

In this model, we would either need to figure out who owns gs://k8s-kops-gce and give `kops-nodes` RW access to it or we could create a new bucket gs://k8s-kops-gce-[v1beta1] and I can adjust the GCE tests to use this as state store and etcd backups.

Hopefully that makes sense, happy to work through this together here or on slack (@geojaz)!

[Prow](https://prow.k8s.io/?job=*kops*gce*)

The current errors show failures because `pr-kubekins` apparently doesn't have the ability to assign itself to vms. This is just evidence of my current dead-end. Assuming we can get `kops-nodes`, I would change [this](https://github.com/kubernetes/test-infra/blob/36c6a27f0a1a52c27add2d080a63c25ad5f8c5dc/config/jobs/kubernetes/kops/kops-periodics-gce.yaml#L24) to the new `kops-nodes` account. 

Thanks!

cc @justinsb

justinsb:
Until we resolve #17558, let's make sure this works with the default
service account.

### DIFF
--- a/config/jobs/kubernetes/kops/kops-periodics-gce.yaml
+++ b/config/jobs/kubernetes/kops/kops-periodics-gce.yaml
@@ -55,7 +55,8 @@ periodics:
       - --env=KOPS_RUN_TOO_NEW_VERSION=1
       - --extract=ci/latest
       - --ginkgo-parallel
-      - --kops-args=--gce-service-account=pr-kubekins@kubernetes-jenkins-pull.iam.gserviceaccount.com
+      # Temporarily use default service account: https://github.com/kubernetes/test-infra/issues/17558
+      #- --kops-args=--gce-service-account=pr-kubekins@kubernetes-jenkins-pull.iam.gserviceaccount.com
       - --kops-priority-path=/workspace/kubernetes/platforms/linux/amd64
       - --kops-version=https://storage.googleapis.com/kops-ci/bin/latest-ci-updown-green.txt
       - --kops-zones=us-central1-c


justinsb:
cc @geojaz 

Let's get the job working again first, then create a new job that uses the service account

PR #17826, opened 2020-06-03
Stop bumping things in prow's codebase.

fejta:
* Deleted deprecated branchprotector cronjob (it is now a prowjob)
* Moved controller manager resources under config/prow
* Not using a named tag in the checkconfig unittest data.

This should hopefully reduce the number of pointless prow releases (aka releases that haven't actually changed anything in prow's codebase)

fixes https://github.com/kubernetes/test-infra/issues/17824

PR #17852, opened 2020-06-05
Fix branchprotector cronjob.yaml link

bells17:


alvaroaleman:
@bells17 the file got completely removed in https://github.com/kubernetes/test-infra/pull/17826 because this is supposed to be a periodic now so it can be viewed via the deck ui

bells17:
@alvaroaleman Thank you for you reply.
But I can't understand yet...
How should I use branchprotector now?

alvaroaleman:
Via a periodic like this: https://github.com/kubernetes/test-infra/blob/2d17640dc193b43104d631c93bb652b07d766134/config/jobs/kubernetes/test-infra/test-infra-trusted.yaml#L887

bells17:
@alvaroaleman  Sorry for the late reply.

Um... when using prow on one server, do you expect the namespace of the prowjob to work in the same namespace as the prow components?
The branchprotector needs config.yaml of the prow.
On the other hand, I think that config.yaml is supposed to be updated by the updateconfig plugin.

If it is assumed that periodic is used in a namespace different from prow components, should config update by plugin updateconfig plugin be configured to be performed in a namespace that executes a prow job separately from the namespace of prow components?

PR #17854, opened 2020-06-05
Use org variant of the bazel image.

fejta:
This has multiple versions of bazel installed, which will allow us to honor .bazelversion.
This will allow us to test the new version in presubmit and therefore safely upgrade bazel


fejta:
/assign @spiffxp @michelle192837 

Review please (adding some new people)

PR #17860, opened 2020-06-05
PlankV2: Couple of small bugs

alvaroaleman:
* Flag registered on the wrong flagset
* TerminateDupes logs an error when the pod is already gone and might use the wrong cluster if the cluster for the prowjob got changed
* startPod didn't block until the pod is in the cache, which might result in aborting the prowjob if a new reconcile request came in that tries to create it as well and fails with IsAlreadyExists

### DIFF
--- a/prow/plank/reconciler.go
+++ b/prow/plank/reconciler.go
@@ -550,6 +550,22 @@ func (r *reconciler) startPod(pj *prowv1.ProwJob) (string, string, error) {
 	if err != nil {
 		return "", "", err
 	}
+
+	// We must block until we see the pod, otherwise a new reconciliation may be triggered that tries to create
+	// the pod because its not in the cache yet, errors with IsAlreadyExists and sets the prowjob to failed
+	podName := types.NamespacedName{Namespace: pod.Namespace, Name: pod.Name}
+	if err := wait.Poll(100*time.Millisecond, 2*time.Second, func() (bool, error) {
+		if err := client.Get(r.ctx, podName, pod); err != nil {
+			if kerrors.IsNotFound(err) {
+				return false, nil
+			}
+			return false, fmt.Errorf("failed to get pod %s in cluster %s: %w", podName.String(), pj.ClusterAlias(), err)
+		}
+		return true, nil
+	}); err != nil {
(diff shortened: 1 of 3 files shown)


alvaroaleman:
/assign @fejta @petr-muller 
PR should be easy to review commit by commit :upside_down_face:

prow/plank/reconciler.go
```diff
@@ -239,22 +239,24 @@ func (r *reconciler) terminateDupes(pj *prowv1.ProwJob) error {
...
+	if !ok {
+		return fmt.Errorf("no client for cluster %q present", pj.ClusterAlias())
+	}
+	podToDelete := &corev1.Pod{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: r.config().PodNamespace,
+			Name:      pj.Name,
+		},
+	}
+	if err := client.Delete(r.ctx, podToDelete); err != nil && !kerrors.IsNotFound(err) {
```
fejta, line 256: if it is already deleted would we want to info/debug/trace this?

prow/plank/reconciler.go
```diff
@@ -550,6 +550,22 @@ func (r *reconciler) startPod(pj *prowv1.ProwJob) (string, string, error) {
 	if err != nil {
 		return "", "", err
 	}
+
+	// We must block until we see the pod, otherwise a new reconciliation may be triggered that tries to create
+	// the pod because its not in the cache yet, errors with IsAlreadyExists and sets the prowjob to failed
```
fejta, line 555: yikes

prow/plank/reconciler.go
```diff
@@ -550,6 +550,22 @@ func (r *reconciler) startPod(pj *prowv1.ProwJob) (string, string, error) {
...
+	podName := types.NamespacedName{Namespace: pod.Namespace, Name: pod.Name}
+	if err := wait.Poll(100*time.Millisecond, 2*time.Second, func() (bool, error) {
```
fejta, line 557: eventual consistency is my

prow/plank/reconciler.go
alvaroaleman: This is expected to happen sometimes if ppl push rapidly

prow/plank/reconciler.go
fejta: favorite kind

fejta: of consistency

PR #17863, opened 2020-06-05
Update to latest alpine and git images

fejta:
Now that these images are in `gcr.io/k8s-prow` and have the `vYYYYMMDD-commit` format, it should cause them to get bumped automatically as new versions appear.


fejta:
/hold cancel

Yeah as these are intermediate images I think that's fine.

PR #17864, opened 2020-06-06
Update tests after changing default image to Ubuntu 20.04

hakman:
Changes:
* defaults `KUBE_SSH_USER` to `ubuntu`
* sets `KUBE_SSH_USER` to `admin` for Kubernetes 1.16 and 1.17
* autodetects `--kops-ssh-user` in Kops
* updates default image for Ubuntu 20.04 to latest
* adds daily test for Ubuntu 20.04 in grid

/cc @justinsb @rifelpet

PR #17866, opened 2020-06-06
Use "ubuntu" user explicitly for branches 1.18 and master

hakman:
Hopefully the last change needed for now.
Ref: #17864

/cc @mikesplain @rifelpet


hakman:
Thanks also for sticking around for the approvals :)

PR #17867, opened 2020-06-06
releng: Remove 1.15 branch jobs

justaugustus:
- Remove 1.15 branch jobs
- Allow release branch jobs to be deleted before branch cuts
- Fixup annotations on e2e-gce-rbe and e2e-containerd-gce jobs


config/jobs/kubernetes/sig-cloud-provider/gcp/gcp-gce.yaml
```diff
@@ -276,6 +276,9 @@ presubmits:
     - release-\d+\.\d+ # per-release image
     annotations:
       fork-per-release: "true"
+      testgrid-alert-stale-results-hours: "24"
+      testgrid-create-test-group: "true"
+      testgrid-num-failures-to-alert: "10"
```
spiffxp, line 281: What policy are you trying to standardize to for this? Seems like something that could be enforced via tests against job configs

config/jobs/kubernetes/sig-node/sig-node-presubmit.yaml
```diff
@@ -44,6 +44,8 @@ presubmits:
     - release-\d+\.\d+   # per-release image
     annotations:
       fork-per-release: "true"
+      testgrid-create-test-group: "true"
+      testgrid-num-failures-to-alert: "10"
```
spiffxp, line 48: Same question here

justaugustus:
@spiffxp -- Less so an attempt to standardize; just trying to restore the annotations that were present in the 1.15 jobs:

https://github.com/kubernetes/test-infra/blob/9d840bab4f435f8da9edfba2597684b406758617/config/jobs/kubernetes/sig-release/release-branch-jobs/1.15.yaml#L687-L690

and

https://github.com/kubernetes/test-infra/blob/9d840bab4f435f8da9edfba2597684b406758617/config/jobs/kubernetes/sig-release/release-branch-jobs/1.15.yaml#L767-L769

I think the bare minimum that needs to be added to the main jobs would be `testgrid-create-test-group: "true"` to get the tests passing, so I can just include that instead.

spiffxp:
narrowing down to `testgrid-create-test-group: "true"` SGTM

Issue #11402, opened 2019-02-20
Repurpose k8s-merge-robot as k8s-github-robot


spiffxp:
**What should be cleaned up or changed**:
Since we've turned down the submit queue, k8s-merge-robot has sat around basically unused.  It has a github token with the same privileges as k8s-ci-robot.  Let's use it for automation that manages github configuration, and leave the k8s-ci-robot token for prow's interactions with github related to CI.  This will give us some breathing room while we figure out other ways to mitigate our github token usage limits for prow.k8s.io (like using an app instead of personal access token)

Suggest:
- renaming k8s-merge-robot to k8s-github-robot
- create a suitable avatar that makes it clear it's the evil twin of k8s-ci-robot
- use k8s-github-robot's token for utilities such as:
  - peribolos
  - label_sync
  - branch-protector?
  - migratestatus
  - etc.

**Provide any links for context**:

spiffxp:
I've successfully renamed k8s-merge-robot, we now have https://github.com/k8s-github-robot

stevekuznetsov:
Let's call out `status-reconciler` specifically?

stevekuznetsov:
All this SGTM

spiffxp:
I accepted all of the org invites, so at this point k8s-github-robot's token should be good for things.  Now to see if it's even in the prow cluster to begin with

spiffxp:
Checking in on label-sync first led me to https://github.com/kubernetes/test-infra/pull/11421

Take two...

nikhita:
> renaming k8s-merge-robot to k8s-github-robot
> create a suitable avatar that makes it clear it's the evil twin of k8s-ci-robot

These are done

> use k8s-github-robot's token for utilities such as: 

Is this done?

spiffxp:
/milestone v1.15
/assign @cjwagner 
We have talked about this a few times offline. I'm unable to push it forward directly myself.  Do we want to take care of this?  Should we include this in rate limit experiments mentioned in https://github.com/kubernetes/test-infra/issues/12510 ?

cjwagner:
Yes, lets include `k8s-github-robot` in the rate limit experiments.
The main thing we need to do to set this up is configure a new ghproxy instance for use with this token. This shouldn't be too much work, but I'm trying to prioritize some other tasks and this is low priority so I'll probably return to this after Kubecon.

spiffxp:
/area github-management
I think we should change the peribolos jobs to use this token

[...]

PR #17870, opened 2020-06-08
remove deprecated bots in approve plugin

droslean:
If those deprecated github names, will try to self-approve a PR, then the prow bot will remove the approve label.  

### DIFF
--- a/prow/plugins/approve/approve.go
+++ b/prow/plugins/approve/approve.go
@@ -51,10 +51,6 @@ var (
 	commandRegex               = regexp.MustCompile(`(?m)^/([^\s]+)[\t ]*([^\n\r]*)`)
 	notificationRegex          = regexp.MustCompile(`(?is)^\[` + approvers.ApprovalNotificationName + `\] *?([^\n]*)(?:\n\n(.*))?`)
 
-	// deprecatedBotNames are the names of the bots that previously handled approvals.
-	// Each can be removed once every PR approved by the old bot has been merged or unapproved.
-	deprecatedBotNames = []string{"k8s-merge-robot", "openshift-merge-robot"}
-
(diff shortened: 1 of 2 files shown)


alvaroaleman:
lets make this proper and remove the whole handling

alvaroaleman:
@cjwagner @stevekuznetsov can one of you confirm that its fine as I don't have the historical background?

I would like to see it gone though, because whatever problem it fixes, it only does so for k8s and openshift but not for other Prow users

cjwagner:
It has been quite a while since we've abandoned mungegithub so I think this is safe to remove now. It is certainly safe for the k8s use case since `k8s-merge-robot` has been renamed to `k8s-github-robot` already: https://github.com/kubernetes/test-infra/issues/11402#issuecomment-465776137
Steve can confirm for `openshift-merge-robot`.

stevekuznetsov:
We will have `openshift-merge-robot` running around (and use it's token for `tide`). I can't exactly grok the original issue (even though I wrote it).

cjwagner:
> I can't exactly grok the original issue (even though I wrote it).

This logic determines if the `approved` label was added by the bot or manually added by a human (in which case we don't allow the bot to remove it). It needed to know the names of the bots that handled the approval process from mungegithub to avoid treating them as humans.

`// Each can be removed once every PR approved by the old bot has been merged or unapproved.` 

Once all the instances of the `approved` label that were added by mungegithub have been handled it is safe for us to remove this special case. If there are remaining instances when we make this change those PRs will have a sticky approval that won't be removed even if the PR changes and should require additional approvers.

alvaroaleman:
> Once all the instances of the approved label that were added by mungegithub have been handled it is safe for us to remove this special case. 

Okay, that should be the case.
/lgtm
/approve

PR #17871, opened 2020-06-08
cherrypicker: wait longer for forks, comment on failure

stevekuznetsov:
Five minutes may not be enough time for a fork to show up, so we can
afford to wait for a longer time and poll less often. Also, bubble up
failures to fork to the user via comment so they understand what's going
on.

/assign @alvaroaleman @petr-muller

### DIFF
--- a/prow/external-plugins/cherrypicker/server.go
+++ b/prow/external-plugins/cherrypicker/server.go
@@ -163,7 +163,7 @@ func (s *Server) handleIssueComment(l *logrus.Entry, ic github.IssueCommentEvent
 	num := ic.Issue.Number
 	commentAuthor := ic.Comment.User.Login
 
-	l = l.WithFields(logrus.Fields{
+	logger := s.log.WithFields(logrus.Fields{
 		github.OrgLogField:  org,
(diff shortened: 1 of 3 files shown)


prow/external-plugins/cherrypicker/server.go
```diff
@@ -361,7 +361,9 @@ var cherryPickBranchFmt = "cherry-pick-%d-to-%s"
 
 func (s *Server) handle(l *logrus.Entry, requestor string, comment *github.IssueComment, org, repo, targetBranch, title, body string, num int) error {
 	if err := s.ensureForkExists(org, repo); err != nil {
-		return err
+		resp := fmt.Sprintf("cannot fork %s/%s: %v", org, repo, err)
+		s.log.WithFields(l.Data).Info(resp)
```
alvaroaleman, line 365: Why not just use `l` as the logger and instead injects its fields into the main logger?

stevekuznetsov: Copy-pasta from another log line /shrug

PR #17875, opened 2020-06-08
Updated the prow install docs

jjasghar:
- Standardized on `sh`
- standardized on the `$`
- Made some commends about `gcloud` and assuming that `gcloud` is in
  path.

Signed-off-by: JJ Asghar <jjasghar@gmail.com>

### DIFF
--- a/prow/getting_started_deploy.md
+++ b/prow/getting_started_deploy.md
@@ -35,14 +35,16 @@ You need a few things:
 
 1. [`bazel`](https://bazel.build/) build tool installed and working
 1. The prow `tackle` utility. It is recommended to use it by running `bazel run //prow/cmd/tackle` from `test-infra` directory, alternatively you can install it by running `go get -u k8s.io/test-infra/prow/cmd/tackle` (in that case you would also need go installed and working).
-1. Optionally, credentials to a Kubernetes cluster (otherwise, `tackle` will help you create one)
+**Note**: Creating the `tackle` utility assumes you have the `gcloud` application in your `$PATH`,
+if you are doing this on another cloud skip to the **Manual deployment** below.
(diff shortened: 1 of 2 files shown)


prow/getting_started_deploy.md
```diff
@@ -148,8 +152,8 @@ kubectl apply -f config/prow/cluster/starter.yaml
 
 After a moment, the cluster components will be running.
 
-```console
-$ kubectl get deployments
```
cjwagner, line 152: Removing the `$`s makes these less readable IMO.

prow/getting_started_deploy.md
```diff
@@ -57,21 +59,23 @@ See the [Next Steps](#next-steps) section after running this utility.
 
 If you do not want to use the `tackle` utility above, here are the manual set of commands tackle will run.
 
-Prow runs in a kubernetes cluster, so first figure out which cluster you want to deploy prow into. If you already have a cluster, skip to the next step.
-
-You can use the [GCP cloud console](https://console.cloud.google.com/) to set up a project and [create a new Kubernetes Engine cluster](https://console.cloud.google.com/kubernetes).
+Prow runs in a kubernetes cluster, so first figure out which cluster you want to deploy prow into.
+If you already have a cluster, skip to the **Create cluster role bindings** step.
```
cjwagner, line 63: nit: Saying what to skip instead of what to skip to will be more robust to future changes to the document.

prow/getting_started_deploy.md
jjasghar: Should I add it to all of them now?  It was inconsistent :)

cjwagner: Yes that would be great. I think it is helpful to distinguish the command that needs to be run from the command's output, but consistency is definitely most important.
Please also squash your commits.

MHBauer:
docs only
/ok-to-test

PR #17883, opened 2020-06-08
Wrap one error in deck when reading symlink

chaodaiG:
Needed for https://github.com/kubernetes/test-infra/issues/17636

Probably forgot to hit save in #17873 , this should be the last one 

/assign @alvaroaleman


alvaroaleman:
Ideally we find some tool that replaces all existing `fmt.Errorf("...something %v", err)` directives with the `%w` equivalent and find a linter that verifies this in the future.
/lgtm
/approve

PR #17884, opened 2020-06-09
Give starter.yaml some love

alvaroaleman:
* Change the namespace for Prow to `prow`, the `default` namespace
  shouldn't get polluted
* Add and use `ghproxy`, it became a requirement recently
* Rename the ingress from `ing` to `prow` to make it more meaningful
* Stop using `NodePort` services when we only need clusterIP, thats
  useless at best and a security issue at worst
* Use `github-token` and not `oauth-token` as secret name for the github
  token, the latter describes a technology, not a purpose, which is
  confusing
* Add templates for github-token and hmac-token secrets
* Set `decorate_all_jobs: true`, there is no reason to not use
  decoration for new deployments
* Make Ingress compatible with cert-manager
* Enable spyglass, Gubernator is deprecated
* Use minio, so no external object storage needs to be set up
* Add plugins config
* Add spyglass config
* Enable Inrepoconfig
* Add Tide query
* Fix Deck RBAC
* Use prow-controller-manager/plankv2
* Update decoration-utils images
* Update references to `starter.yaml`

Basically: Bring the thing up-to-date with how we expect Prow to be set up and make it something that "just works" after replacing a couple of variables rather than "Do these five manual steps, deploy `starter.yaml`, do another five manual steps.

/assign @spiffxp @fejta @sbueringer 
/cc @saschagrunert 
as you were interested in using prow in an on-prem scenario/without cloud storage.

In the future we could probably make this part of the `stable` promotion process by building it out to be used in a kind-based e2e test that uses something like `ngrok` to expose the hook endpoint.

### DIFF
--- a/config/prow/cluster/starter-gcs.yaml
+++ b/config/prow/cluster/starter-gcs.yaml
@@ -3,24 +3,123 @@
 # their own files.
 ---
 apiVersion: v1
-kind: ConfigMap
+kind: Namespace
 metadata:
-  namespace: default
-  name: plugins
-data:
-  plugins.yaml: ""
+  name: prow
 ---
 apiVersion: v1
 kind: ConfigMap
 metadata:
-  namespace: default
+  namespace: prow
+  name: plugins
+data:
+  plugins.yaml: |
+    plugins:
+      << your_github_org >>:
+      - approve
+      - assign
+      - blunderbuss
+      - cat
+      - dog
+      - help
+      - heart
+      - hold
+      - label
+      - lgtm
+      - trigger
+      - verify-owners
+      - wip
+      - yuks
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  namespace: prow
+  name: github-token
+stringData:
+  token: <<insert-token-here>>
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  namespace: prow
+  name: hmac-token
+stringData:
+  # Generate via `openssl rand -hex 20`. This is the secret used in the GitHub webhook configuration
+  hmac: << insert-hmac-token-here >>
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  namespace: prow
   name: config
 data:
   config.yaml: |
-    prowjob_namespace: default
+    prowjob_namespace: prow
     pod_namespace: test-pods
+
+    in_repo_config:
+      enabled:
+        "*": true
+
+    deck:
+     spyglass:
+       lenses:
+       - lens:
+           name: metadata
+         required_files:
+         - started.json|finished.json
+       - lens:
+           config:
+           name: buildlog
+         required_files:
+         - build-log.txt
+       - lens:
+           name: junit
+         required_files:
+         - .*/junit.*\.xml
+       - lens:
+           name: podinfo
+         required_files:
+         - podinfo.json
+
+    plank:
+      job_url_prefix_config:
+        # TODO: Remove the gcs suffix once https://github.com/kubernetes/test-infra/pull/17779 is in
+        "*": https://prow.<< your-domain.com >>/view/gcs
+      report_templates:
+        '*': >-
+            [Full PR test history](https://prow..<< your-domain.com >>/pr-history?org={{.Spec.Refs.Org}}&repo={{.Spec.Refs.Repo}}&pr={{with index .Spec.Refs.Pulls 0}}{{.Number}}{{end}}).
+            [Your PR dashboard](https://prow.<< your-domain.com >>/pr?query=is:pr+state:open+author:{{with
+            index .Spec.Refs.Pulls 0}}{{.Author}}{{end}}).
+      default_decoration_configs:
+        "*":
+          gcs_configuration:
+            bucket: gs://bucket-name
+            path_strategy: explicit
+          gcs_credentials_secret: gcs-credentials
+          utility_images:
+            clonerefs: gcr.io/k8s-prow/clonerefs:v20200710-7fa016752a
+            entrypoint: gcr.io/k8s-prow/entrypoint:v20200710-7fa016752a
+            initupload: gcr.io/k8s-prow/initupload:v20200710-7fa016752a
+            sidecar: gcr.io/k8s-prow/sidecar:v20200710-7fa016752a
+
+    tide:
+      queries:
+      - labels:
+        - lgtm
+        - approved
+        missingLabels:
+        - needs-rebase
+        - do-not-merge/hold
+        - do-not-merge/work-in-progress
+        - do-not-merge/invalid-owners-file
+        orgs:
+        - << your_github_org >>
+
+    decorate_all_jobs: true
     periodics:
-    - interval: 10m
+    - interval: 1m
       agent: kubernetes
       name: echo-test
       spec:
@@ -118,7 +217,7 @@ spec:
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  namespace: default
+  namespace: prow
   name: hook
   labels:
     app: hook
@@ -146,6 +245,9 @@ spec:
         args:
         - --dry-run=false
         - --config-path=/etc/config/config.yaml
+        - --github-token-path=/etc/github/token
+        - --github-endpoint=http://ghproxy
+        - --github-endpoint=https://api.github.com
         ports:
           - name: http
             containerPort: 8888
@@ -153,7 +255,7 @@ spec:
         - name: hmac
           mountPath: /etc/webhook
           readOnly: true
-        - name: oauth
+        - name: github-token
           mountPath: /etc/github
           readOnly: true
         - name: config
@@ -179,9 +281,9 @@ spec:
       - name: hmac
         secret:
           secretName: hmac-token
-      - name: oauth
+      - name: github-token
         secret:
-          secretName: oauth-token
+          secretName: github-token
       - name: config
         configMap:
           name: config
@@ -192,60 +294,18 @@ spec:
 apiVersion: v1
 kind: Service
 metadata:
-  namespace: default
+  namespace: prow
   name: hook
 spec:
   selector:
     app: hook
   ports:
   - port: 8888
-  type: NodePort
 ---
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  namespace: default
-  name: plank
-  labels:
-    app: plank
-spec:
-  selector:
-    matchLabels:
-      app: plank
-  replicas: 1 # Do not scale up.
-  strategy:
-    type: Recreate
-  template:
-    metadata:
-      labels:
-        app: plank
-    spec:
-      serviceAccountName: "plank"
-      containers:
-      - name: plank
-        image: gcr.io/k8s-prow/plank:v20200714-1215e87acc
-        args:
-        - --dry-run=false
-        - --config-path=/etc/config/config.yaml
-        volumeMounts:
-        - name: oauth
-          mountPath: /etc/github
-          readOnly: true
-        - name: config
-          mountPath: /etc/config
-          readOnly: true
-      volumes:
-      - name: oauth
-        secret:
-          secretName: oauth-token
-      - name: config
-        configMap:
-          name: config
----
-apiVersion: apps/v1
-kind: Deployment
-metadata:
-  namespace: default
+  namespace: prow
   name: sinker
   labels:
     app: sinker
@@ -277,7 +337,7 @@ spec:
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  namespace: default
+  namespace: prow
   name: deck
   labels:
     app: deck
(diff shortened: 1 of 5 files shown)


stevekuznetsov:
/lgtm
/hold

These changes look good to me, nice to see this modernized.

config/prow/cluster/starter.yaml
```diff
@@ -3,24 +3,149 @@
...
+apiVersion: v1
+kind: Secret
+metadata:
+  namespace: prow
+  name: s3-credentials
+stringData:
+  service-account.json: |
+    {
+      "region": "minio",
+      "access_key": "<<CHANGE_ME_MINIO_ACCESS_KEY>>",
```
fejta, line 60: It doesn't seem like we should be encouraging people to include secrets in the starter.yaml file.

config/prow/cluster/starter.yaml
```diff
@@ -876,3 +923,400 @@ roleRef:
...
+        - /data/prow-logs
+        volumeMounts:
+        - name: data
+          mountPath: "/data"
+      containers:
+      - name: minio
+        volumeMounts:
+        - name: data
+          mountPath: "/data"
+        image: minio/minio:latest
```
fejta, line 1287: Shouldn't we pick a specific version that we've tested against?

alvaroaleman:
> A lot of these changes seem overly coupled to S3. This unnecessarily increases the length/complexity of the file for people on GCP.

No. They are coupled to an object storage that can be self-hosted which happens to have a S3-compatible API as there is no such thing with a GCS-compatible API (to my knowledge). The overall goal of the `starter.yaml` file is to allow ppl to check out Prow and make it as easy as possible to get to the point of having a working Prow setup with as little work as possible.

Setting up object storage on any cloud provider is par definition always more work than applying a couple of manifests. Most ppl that stick with Prow will want that, which is why the `prow/getting_started_deploy.md` contains a section on how to do that.

config/prow/cluster/starter.yaml
alvaroaleman: Its not particularly nice but the thing is not publicly exposed, hence its not really a big deal IMHO

config/prow/cluster/starter.yaml
alvaroaleman: But then we have to regularly update it. The S3 api the thing implements can be assumed to be very stable.

fejta:
The issue I have is that all the minio volumes, secrets, deployments, etc are totally unnecessary. And we should make it easy to delete them. `kubectl delete -f minio.yaml` accomplishes that. And `kubectl create -f starter.yaml -f minio.yaml` is not substantially harder than when there is just one `-f` flag.

config/prow/cluster/starter.yaml
sbueringer: Another option could be to use some kind of envsubst magic. But no idea if we want that/ if it's better.

config/prow/cluster/starter.yaml
```diff
@@ -3,24 +3,149 @@
...
+    prowjob_namespace: prow
     pod_namespace: test-pods
+
+    in_repo_config:
+      enabled:
+        "*": true
+
+    deck:
+     spyglass:
+       lenses:
```
sbueringer, line 99: Just fyi. If we have a publicly accessible Minio URL this would also work:
gcs_browser_prefix: <minio-url>/minio/

But not sure if it's overly complicated for the starter.yaml

sbueringer:
@fejta @alvaroaleman I think if we want to decouple it from s3, it makes sense to use e.g. kustomize. Because the s3 part is not only deploying Minio but also different flags in deck etc.

config/prow/cluster/starter.yaml
```diff
@@ -3,24 +3,149 @@
...
+      queries:
+      - labels:
+        - lgtm
+        - approved
+        missingLabels:
+        - needs-rebase
+        - do-not-merge/hold
+        - do-not-merge/work-in-progress
+        - do-not-merge/invalid-owners-file
+        orgs:
```
sbueringer, line 143: I'm also setting:
```
pr_status_base_urls:
    '*': https://prow.<< your-domain.com >>/pr
```

config/prow/cluster/starter.yaml
```diff
@@ -3,24 +3,149 @@
...
+           name: junit
+         required_files:
+         - .*/junit.*\.xml
+       - lens:
+           name: podinfo
+         required_files:
+         - podinfo.json
+
+    plank:
+      job_url_prefix_config:
```
sbueringer, line 119: There's also `job_url_template` and `report_templates` were I included a template with the prow domain in my deployment. Not sure if it's important here

config/prow/cluster/starter.yaml
```diff
@@ -146,14 +271,17 @@ spec:
         args:
         - --dry-run=false
         - --config-path=/etc/config/config.yaml
```
sbueringer, line 273: Sorry it's not about fixing my deployment but don't we need `job-config-path` and the mounted jobs in hook?

sbueringer: Ah I guess we dont need it because we use in-repo config?

config/prow/cluster/starter.yaml
```diff
@@ -420,48 +516,57 @@ spec:
...
+        - --github-token-path=/etc/github/token
+        - --github-endpoint=http://ghproxy
+        - --github-endpoint=https://api.github.com
+        - --github-graphql-endpoint=http://ghproxy/graphql
```
sbueringer, line 522: Probably not necessary for starter:
```
        - --status-path=s3://tide/tide-status
        - --history-uri=s3://tide/tide-history.json
```

config/prow/cluster/starter.yaml
```diff
@@ -499,9 +608,11 @@ spec:
         - --continue-on-error=true
         - --plugin-config=/etc/plugins/plugins.yaml
         - --config-path=/etc/config/config.yaml
-        - --github-token-path=/etc/github/oauth
+        - --github-token-path=/etc/github/token
+        - --github-endpoint=http://ghproxy
+        - --github-endpoint=https://api.github.com
```
sbueringer, line 613: Probably not necessary for starter:
```
- --status-path=s3://status-reconciler/status-reconciler-status
```

config/prow/cluster/starter.yaml
alvaroaleman: So `job_url_template` is ignored when `plank.GetJobURLPrefix(pj.Spec.Refs) != ""` and should just not be used.
I'll add `report_template`. We are so bad at defaulting.

config/prow/cluster/starter.yaml
alvaroaleman: Its a single file that has both prow config and job config

config/prow/cluster/starter.yaml
alvaroaleman: In this sample the minio is not public

config/prow/cluster/starter.yaml
alvaroaleman: Yeah but this just gives an empty, non-erroring page unless you set up a github app

alvaroaleman:
> I think if we want to decouple it from s3, it makes sense to use e.g. kustomize. Because the s3 part is not only deploying Minio but also different flags in deck etc.

I really hate templating as its usually the super important glue that doesn't have any kind of testing. But yeah, I guess that has to be done

config/prow/cluster/starter.yaml
sbueringer: Ah right, makes sense

config/prow/cluster/starter.yaml
sbueringer: Should be pretty stable with Minio. We're using it since about over 2 years with any problems in this regard.

config/prow/cluster/starter.yaml
sbueringer: Good to know..

config/prow/cluster/starter.yaml
sbueringer: kk got it

config/prow/cluster/starter.yaml
```diff
@@ -3,24 +3,149 @@
 # their own files.
```
tseelbach, line 3: perhaps add to line 2/3 comment.  Something like:
"For examples see the per component yaml files in this directory."

alvaroaleman:
@fejta @sbueringer PTAL. Ive played around a bit with using kustomize, but its super finicky for our use case because we have to do things like "Adjust the embedded yaml" or "add this arg to that deployment". So I ended up just creating two files, one for minio/s3 and one for gcs.

I can create a follow-up that adds a script that checks the diff of the two files against a known-accepted diff to prevent drift. Follow-up because this PR collects a merge conflict every time prow gets bumped which is a bit annoying.

config/prow/cluster/starter-gcs.yaml
```diff
@@ -3,24 +3,123 @@
...
+        "*": https://prow.<< your-domain.com >>/view/gcs
+      report_templates:
+        '*': >-
+            [Full PR test history](https://prow..<< your-domain.com >>/pr-history?org={{.Spec.Refs.Org}}&repo={{.Spec.Refs.Repo}}&pr={{with index .Spec.Refs.Pulls 0}}{{.Number}}{{end}}).
+            [Your PR dashboard](https://prow.<< your-domain.com >>/pr?query=is:pr+state:open+author:{{with
+            index .Spec.Refs.Pulls 0}}{{.Author}}{{end}}).
+      default_decoration_configs:
+        "*":
+          gcs_configuration:
+            bucket: gcs://bucket-name
```
sbueringer, line 98: Should be `gs://` (same in other cases)

config/prow/cluster/starter-gcs.yaml
```diff
@@ -420,48 +489,60 @@ spec:
         args:
         - --dry-run=false
         - --config-path=/etc/config/config.yaml
+        - --github-token-path=/etc/github/token
+        - --github-endpoint=http://ghproxy
+        - --github-endpoint=https://api.github.com
+        - --github-graphql-endpoint=http://ghproxy/graphql
+        - --gcs-credentials-file=/etc/gcs-credentials/service-account.json
```
sbueringer, line 496: mount point for the gcs credentials is missing

config/prow/cluster/starter-gcs.yaml
```diff
@@ -499,9 +584,13 @@ spec:
         - --continue-on-error=true
         - --plugin-config=/etc/plugins/plugins.yaml
         - --config-path=/etc/config/config.yaml
-        - --github-token-path=/etc/github/oauth
+        - --github-token-path=/etc/github/token
+        - --github-endpoint=http://ghproxy
+        - --github-endpoint=https://api.github.com
+        - --gcs-credentials-file=/etc/gcs-credentials/service-account.json
```
sbueringer, line 590: mount point is missing

sbueringer:
@alvaroaleman A few nits. What do you think about calling the secret for s3 and gcs just something generic like `blob-storage-credential`. Not sure if it's worth it, but then there would almost be no diff between both deployments (only a few bucket paths, gcs-credentials-file vs s3-credentials file and the Minio depoyment I guess)

config/prow/cluster/starter-s3.yaml
```diff
@@ -0,0 +1,1346 @@
...
+  service-account.json: |
+    {
+      "region": "minio",
+      "access_key": "<<CHANGE_ME_MINIO_ACCESS_KEY>>",
+      "endpoint": "minio.prow.svc.cluster.local",
+      "insecure": true,
+      "s3_force_path_style": true,
+      "secret_key": "<<CHANGE_ME_MINIO_SECRET_KEY>>"
+    }
+---
```
sbueringer, line 1273: Should we add similar dummy credentials in the gcs yaml?

stevekuznetsov:
@enj

config/prow/cluster/starter.yaml
tseelbach: About secrets in starter.yaml:
- we have enterprise rules that we cannot commit secrets/passwords into git repos. 
- as a new Prow user, the new starter.yaml worked almost perfectly(Thank you!)  
- So following gitOps approach I wanted to check it in, so my team could start customizing, and so i could roll back when we mess up (remember we are new) But first i had to split out the secret creation.    It would have been better to have secret creation as a separate step in the first place.

alvaroaleman: I disagree. The purpose of this is to get ASAP something that works so ppl can start toying around and check if Prow fits their requirements. If and how ppl solve the problem of not having this in their yaml is not something that is supposed to get solved here. Removing a secret from a yaml manifest is trivial. Adding it and with it the correct keys might not be as trivial.

config/prow/cluster/starter-s3.yaml
alvaroaleman: IMHO its less relevant because you can just download a working credential from GCP whereas what we have for AWS is a custom format

alvaroaleman:
@fejta @sbueringer rebased and existing feedback addressed. Even if not perfect, its IMHO a lot better to get someone up and running quickly than what we had before. Since this is a rebase magnet, any chance we can merge it and do follow-ups if we find more issues?

sbueringer:
@alvaroaleman  All findings from my side addressed so 
/lgtm 

anyway :)

sbueringer:
/lgtm 
:)

PR #17886, opened 2020-06-09
Migrate 100 node release-branch blocking jobs to k8s-infra-prow-build

spiffxp:
Followup to #17725 which migrated just the master branch job


spiffxp:
/cc @mm4tt 
scalability oncall

PR #17889, opened 2020-06-09
Bugzilla: Add functions to get all clones across multi levels

GeoBK:


prow/bugzilla/client.go
```diff
@@ -117,6 +119,105 @@ func (c *client) GetClones(bug *Bug) ([]*Bug, error) {
...
+	// One option would be to return as soon as the first parent is found
+	// ideally that should be enough, although there is a check in the getRoot function to verify this
+	// Logs would need to be monitored to verify this behavior
+	for _, parentID := range bug.DependsOn {
+		parent, err := c.GetBug(parentID)
+		if err != nil {
+			errs = append(errs, fmt.Errorf("Failed to get parent bug #%d: %v", parentID, err))
+			continue
+		}
+		if parent.Summary == bug.Summary {
```
alvaroaleman, line 158: Why do we need this check?

prow/bugzilla/client.go
```diff
@@ -117,6 +119,105 @@ func (c *client) GetClones(bug *Bug) ([]*Bug, error) {
...
+
+func getRoot(c Client, bug *Bug) (*Bug, error) {
+	curr := bug
+	var errs []error
+	for len(curr.DependsOn) > 0 {
+		parent, err := getImmediateParents(c, curr)
+		if err != nil {
+			errs = append(errs, err)
+		}
+		if len(parent) > 0 {
```
alvaroaleman, line 173: Nit: Might be easier to read as `switch` statement:
```
switch n :=  len(parent) {
 case n == 1:
   curr = parent[0]
  case n > 1:
    error
  default:
    continue
}
```
}

prow/bugzilla/client.go
```diff
@@ -117,6 +119,105 @@ func (c *client) GetClones(bug *Bug) ([]*Bug, error) {
...
+			// Might be a good idea to keep this check - since we dont expect more than 1 parent
+			if len(parent) > 1 {
+				errs = append(errs, fmt.Errorf("More than one parent found for bug #%d", curr.ID))
+			}
+			curr = parent[0]
+		} else {
```
alvaroaleman, line 179: Dont we also have to check for clones and their parents here?

prow/bugzilla/client.go
```diff
@@ -117,6 +119,105 @@ func (c *client) GetClones(bug *Bug) ([]*Bug, error) {
...
+}
+
+// GetRoot returns the original bug.
+func (c *client) GetRoot(bug *Bug) (*Bug, error) {
+	return getRoot(c, bug)
+}
+
+// GetAllClones returns all the clones of the bug including itself
+// Differs from GetClones as GetClones only gets the child clones which are one level lower
+func (c *client) GetAllClones(bug *Bug) ([]*Bug, error) {
```
alvaroaleman, line 193: Can you put the explanation above in the `type Client interface`? The godoc here won't be visible from the outside as the `client` is unexported

prow/bugzilla/client.go
GeoBK: According to my understanding the depends on field can contain parent bugs which are not clone parents, and only the ones with matching summaries are clone parents. Is that wrong?

prow/bugzilla/client.go
GeoBK: I dont think so.. Since if len(parent)<=0 we are at the terminating stage i believe.

prow/bugzilla/client.go
GeoBK: I might be missing something as I dont see any of the other functions in the interface having any explanations. Do you mean line 51 of this file or did you mean some other place?

prow/bugzilla/types.go
```diff
@@ -107,6 +107,8 @@ type Bug struct {
 	Version []string `json:"version,omitempty"`
 	// Whiteboard is he value of the "status whiteboard" field on the bug.
 	Whiteboard string `json:"whiteboard,omitempty"`
+	// PRs holds the links to the pull requests associated with the bug.
+	PRs []ExternalBug
```
GeoBK, line 111: Also I wanted to call attention to this change. Since this is an incremental change I dont think it would break anything, but just wanted someones go ahead

alvaroaleman: It should be fine, but please add a `json:"fieldname,omitempty" marker

prow/bugzilla/client.go
alvaroaleman: Yeah, I mean line 51. I don't know why it was done like that before, but having godocs on the implementation and not the interface doesn't really make any sense, as you can only see them when looking at the implementation. But the implementation is unexported, the only thing that is exported is the interface and that is returned when constructing a new client.

prow/bugzilla/client.go
alvaroaleman: Hm I see, yeah it seems we do it like that elsewhere as well

prow/bugzilla/client.go
GeoBK: Yup. Got it. Should I move the explanations to the interface for the rest of the functions as well?

prow/bugzilla/client_test.go
```diff
@@ -930,3 +930,156 @@ func TestGetExternalBugPRsOnBug(t *testing.T) {
...
...e" : "If docs needed, set a value", "cf_environment" : "", "cf_fixed_in" : "", "cf_last_closed" : null, "cf_release_notes" : "", "cf_target_upstream_version" : "", "classification" : "Red Hat", "component" : ["Unknown" ], "creation_time" : "2020-06-03T09:03:41Z", "creator" : "Gal Zaidman", "creator_detail" : {"email" : "gzaidman","id" : 422274,"name" : "gzaidman","real_name" : "Gal Zaidman" }, "deadline" : null, "depends_on" : [], "docs_contact" : "", "dupe_of" : null, "groups" : [], "id" : 1843407, "is_cc_accessible" : true, "is_confirmed" : true, "is_creator_accessible" : true, "is_open" : true, "keywords" : [], "last_change_time" : "2020-06-07T11:08:27Z", "op_sys" : "Unspecified", "platform" : "Unspecified", "priority" : "unspecified", "product" : "OpenShift Container Platform", "qa_contact" : "Jianwei Hou", "qa_contact_detail" : {"email" : "jhou","id" : 332453,"name" : "jhou","real_name" : "Jianwei Hou" }, "resolution" : "", "see_also" : [], "severity" : "medium", "status" : "VERIFIED", "summary" : "[oVirt] add oVirt as a provide to openshift tests", "target_milestone" : "---", "target_release" : ["4.6.0" ], "url" : "", "version" : ["4.5" ], "whiteboard" : ""}],"faults" : []}`,
```
alvaroaleman, line 942: Please do not put real email addresses and names on github, even if they can be retrieved by anyone elsewhere

GeoBK: ok, i'll black that out

prow/bugzilla/client_test.go
```diff
@@ -930,3 +930,161 @@ func TestGetExternalBugPRsOnBug(t *testing.T) {
...
+		bugDetails     string
+		prDetails      string
+		expectedClones map[int]struct{}
+	}{
+		1843407: {
+			`{"bugs" : [{ "alias" : [], "assigned_to" : "Sudha Ponnaganti",  "blocks" : [1844101 ], "cf_clone_of" : null, "cf_doc_type" : "If docs needed, set a value", "cf_environment" : "", "cf_fixed_in" : "", "cf_last_closed" : null, "cf_release_notes" : "", "cf_target_upstream_version" : "", "classification" : "Red Hat", "component" : ["Unknown" ], "creation_time" : "2020-06-03T09:03:41Z", "creator" : "Gal Zaidman", "deadline" : null, "depends_on" : [], "docs_contact" : "", "dupe_of" : null, "groups" : [], "id" : 1843407, "is_cc_accessible" : true, "is_confirmed" : true, "is_creator_accessible" : true, "is_open" : true, "keywords" : [], "last_change_time" : "2020-06-07T11:08:27Z", "op_sys" : "Unspecified", "platform" : "Unspecified", "priority" : "unspecified", "product" : "OpenShift Container Platform", "resolution" : "", "see_also" : [], "severity" : "medium", "status" : "VERIFIED", "summary" : "[oVirt] add oVirt as a provide to openshift tests", "target_milestone" : "---", "target_release" : ["4.6.0" ], "url" : "", "version" : ["4.5" ], "whiteboard" : ""}],"faults" : []}`,
```
alvaroaleman, line 942: There are still names

prow/bugzilla/client_test.go
```diff
@@ -930,3 +931,162 @@ func TestGetExternalBugPRsOnBug(t *testing.T) {
...
+				}
+			}
+		})
+
+	}
+
+}
+
+func TestGetRootForClone(t *testing.T) {
+	bugMappings := map[int]string{
```
alvaroaleman, line 1033: These tests are pretty hard to read because they are based on embedded raw json and from looking at them its not immediately obvious which one covers which scenario like "direct parent", "root is a grandparent" etc.

A better way to do this would be to not test the whole handler but just the `getAllClones` function. The code is factored in a way that this is easely possible by passing in a fake/mock to `getAllClones`

alvaroaleman: IMHO the same applies to the other testcase, it would be much easier to read if it wasn't based on embedded json

GeoBK: I'm not sure I understand what you mean by this. As far as I understand I am testing only the GetAllClones function. Do you perhaps have an example for some other piece of code you are testing which would help me understand?
I can whittle down the size of the raw json if that would help, because not all the fields are necessary for my test.

alvaroaleman: You are testing the `GetAllClones`, not the `getAllCloens`. The latter allows to inject a client for which you can use the `Fake` and on it you can define the individual bugs it will return. You can then only define the relevant fields there like `DependsOn` which should make the tests more readable

prow/bugzilla/client_test.go
```diff
@@ -930,3 +932,95 @@ func TestGetExternalBugPRsOnBug(t *testing.T) {
...
+		t.Fatalf("Error while creating bug in Fake!\n")
+	}
+	bug1, err := fake.GetBug(bug1ID)
+	bug2ID, err := fake.CloneBug(bug1)
+	bug2, err := fake.GetBug(bug2ID)
+	bug3ID, err := fake.CloneBug(bug2)
+	bug1, err = fake.GetBug(bug1ID)
+	bug2, err = fake.GetBug(bug2ID)
+	bug3, err := fake.GetBug(bug3ID)
+	testcases := map[int]*Bug{
```
alvaroaleman, line 1009: Instead of defining the testcases as maps, could you please define them as a
`[]struct` that contains all necesary parameters and a descriptive name?

See for example:
* https://github.com/openshift/ci-tools/blob/cb0cd9370002db5f9cd474d83a6e8bae27663ebc/pkg/diffs/diffs_test.go#L205
* https://github.com/openshift/ci-tools/blob/cb0cd9370002db5f9cd474d83a6e8bae27663ebc/pkg/diffs/diffs_test.go#L386

GeoBK: Ok... I have modified the testcases to be structs instead of maps with a relevant name.

Issue #17790, opened 2020-06-01
Prow cluster nodes oversubscribed?


fejta:
Seeing some intermittent errors on prow components:

/area oncall-hotlist
/kind bug

```yaml
{
 insertId: "102ohy5f8224jo"  
 jsonPayload: {
  bucket: "kubernetes-jenkins"   
  component: "gcsreporter"   
  error: "context deadline exceeded"   
  file: "prow/crier/reporters/gcs/internal/util/util.go:63"   
  func: "k8s.io/test-infra/prow/crier/reporters/gcs/internal/util.WriteContent"   
  level: "warning"   
  msg: "Uploading info to storage failed (write)"   
  path: "logs/ci-kubernetes-e2e-gce-cos-k8sstable1-serial/1267447733355024386/prowjob.json"   
 }
 labels: {…}  
 logName: "projects/k8s-prow/logs/crier"  
 receiveTimestamp: "2020-06-01T13:28:07.998461287Z"  
 resource: {…}  
 severity: "ERROR"  
 timestamp: "2020-06-01T13:28:02Z"  
}
```

```yaml
{
 insertId: "1mu4hmig1jmjyx5"  
 jsonPayload: {
  component: "crier"   
  error: "failed to wait for updated report status to be in lister: timed out waiting for the condition"   
  file: "prow/crier/controller.go:296"   
  func: "k8s.io/test-infra/prow/crier.(*Controller).processNextItem"   
  jobName: "pull-kops-verify-generated"   
  jobStatus: "triggered"   
  level: "error"   
  msg: "Failed to update report state on prowjob"   
  prowjob: "default/d5884b6a-a436-11ea-a062-9a7eb8e8e889"   
  reporter: "github-reporter"   
 }
 labels: {…}  
 logName: "projects/k8s-prow/logs/crier"  
 receiveTimestamp: "2020-06-01T18:36:51.024275462Z"  
 resource: {…}  
 severity: "ERROR"  
 timestamp: "2020-06-01T18:36:45Z"  
}
```

PR #17891, opened 2020-06-09
Crier: Fix a bug where it waits for the wrong change

alvaroaleman:
In https://github.com/kubernetes/test-infra/commit/01fd07142c61b3f89246cbad83a23ae4251e04f5#diff-0e4877bfa35d63bade3a386f20923299L168 crier was changed to set the `prevReportedStates` field to the state that was reported rather than the last state the job itself was in, as that could change in the meantime because we retry everything and within that retying we fetch the prowjob object again.

However, I forgot to also change the section that blocks until the change is in the cache to wait for the correct change, resulting in timeouts when the prowjob status got changed after it was reported and before the reported state was in the cache (the case where we silently claimed to have reported the current state before, even though we didn't).

Fixes https://github.com/kubernetes/test-infra/issues/17790
Fixes https://github.com/openshift/release/issues/9567

### DIFF
--- a/prow/crier/controller.go
+++ b/prow/crier/controller.go
@@ -197,7 +197,7 @@ func (c *Controller) updateReportState(pj *v1.ProwJob, log *logrus.Entry, report
 			return false, err
 		}
 		if pj.Status.PrevReportStates != nil &&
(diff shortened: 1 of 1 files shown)


fejta:
I don't fully understand why but this seems to be making crier do gerrit reports forever...

PR #17892, opened 2020-06-09
github: fix the fork name field we read

stevekuznetsov:

/assign @alvaroaleman


matthyx:
I don't quite understand that fix

alvaroaleman:
> I don't quite understand that fix

The json unmarshaler only populates exported fields, as it can not see unexported fields. The field was unexported before, hence it never got populated

matthyx:
Thanks!

PR #17895, opened 2020-06-09
Switch `ListTeamMembers` to collect direct members via GitHubv4 API.

evankanderson:
Knative is attempting to adopt Peribolos, but the current team structure uses nested teams, which `peribolos --dump` reads back from the GitHub v3 API 

[Sample configuration](https://github.com/knative/community/pull/135/files#diff-9dd626e75f0cc24b1dfb2f753af55a8e)

For example, the following group definition:
```yaml
      Eventing Writers:
        description: Grants write access to eventing-related repositories.
        privacy: closed
        repos:
          eventing-kafka: write
          eventing-rabbitmq: write
        teams:
          Delivery WG Leads:
            description: The Working Group leads for Event Delivery
            members:
            - matzew
            - Harwayne
            privacy: closed
          Eventing WG Leads:
            description: The Working Group leads for Eventing
            members:
            - grantr
            - lionelvillard
            - vaikas
            privacy: closed
          Source WG Leads:
            description: The Working Group leads for Sources
            members:
            - nachocano
            - lionelvillard
            - vaikas
            - n3wscott
            privacy: closed
```
is rendered as:
```yaml
      Eventing Writers:
        description: Grants write access to eventing-related repositories.
        members:
          - matzew
          - Harwayne
          - grantr
          - lionelvillard
          - vaikas
          - nachocano
          - n3wscott
        privacy: closed
        repos:
          eventing-kafka: write
          eventing-rabbitmq: write
        teams:
          Delivery WG Leads:
            description: The Working Group leads for Event Delivery
            members:
            - matzew
            - Harwayne
            privacy: closed
          Eventing WG Leads:
            description: The Working Group leads for Eventing
            members:
            - grantr
            - lionelvillard
            - vaikas
            privacy: closed
          Source WG Leads:
            description: The Working Group leads for Sources
            members:
            - nachocano
            - lionelvillard
            - vaikas
            - n3wscott
            privacy: closed
```

(Notice the "members" under the top level group.)

Switching to the githubv4 API allows filtering memberships to direct members only.


evankanderson:
area/peribolos

prow/github/client.go
```diff
@@ -2855,27 +2855,63 @@ func (c *client) ListTeamMembers(id int, role string) ([]TeamMember, error) {
...
-		func() interface{} {
-			return &[]TeamMember{}
-		},
-		func(obj interface{}) {
-			teamMembers = append(teamMembers, *(obj.(*[]TeamMember))...)
-		},
-	)
-	if err != nil {
-		return nil, err
+	for {
```
fejta, line 2896: @cjwagner how does ghproxy work with graphql?

fejta:
> the update-bazel and update-deps targets both fail for me on MacOS 10.14:

Please `brew install coreutils`

prow/github/client.go
cjwagner: It doesn't cache, graphql requests are POSTed queries rather than simple REST requests and they don't support ETag based caching. Using graphql is only more efficient than v3+ghproxy in cases where the number of requests is small, there are a large number of resources, and the resources change frequently. Since the number of requests for this endpoint is relatively large and teams don't change frequently I suspect the v3 API + ghproxy will use less of the API ratelimit.

prow/github/client.go
```diff
@@ -2855,27 +2855,63 @@ func (c *client) ListTeamMembers(id int, role string) ([]TeamMember, error) {
...
+							Login string
+						}
+						Role string
+					}
+					PageInfo struct {
+						EndCursor   *githubql.String
+						HasNextPage bool
+					}
+					TotalCount int
+				} `graphql:"members(first: 100, after: $memberCursor, membership: IMMEDIATE)"`
```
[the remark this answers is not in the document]
evankanderson: That's a GitHub limit, you have to page beyond that point.

prow/github/client.go
evankanderson: What about collecting all the teams and members in a single/paged set of queries? For peribolos, this would probably collapse N*(team, team members * 2) to N*(members per team/100).

evankanderson:
Good point on making a new method, I'll do that.

What do you think of optionally adding a list of members by role to the Team struct?

prow/github/client.go
fejta: The resource we try to conserve here is uncached requests. So right now peribolos is essentially free because essentially 100% of the requests are cached.

AKA 10,000 cached requests is cheaper to continually run (cost ~0) than 10 uncached requests (cost 10 per run).

So ideally we keep the cost at C, not N*C which this PR does right now.

Looking through the docs though it looks like github makes it intentionally hard/impossible to only get direct members https://developer.github.com/v3/teams/members/

Could we make this a flag? AKA by default we continue using the v3 api and including children members in the parent with a flag to make this more precise by using more tokens and the v4 api?

fejta:
> optionally adding a list of members by role to the Team struct?

An issue with putting it in the shared github package is that the struct never includes this information: https://developer.github.com/v3/teams/members/#list-team-members

so how about putting this into a local struct that embeds the github.Team struct, aka something like

```golang
type peribolosTeam struct {
  github.Team
  role string
}
```

and then work with that, including having the `listTeams()` v4 thing populate it automatically. I think right now we just make two ListTeamMembers calls for each role, right?

evankanderson:
ListTeamMembers may have paginated results, but yes, there are two calls per team, and I think we can collapse the team list + per-team memberships to a single graphql query.

prow/github/client.go
evankanderson: It looks like GitHub doesn't provide an equivalent GET endpoint for GraphQL, despite the standard recommending it. You *could* enable caching for POST queries to the graphql endpoint if they have the "query" key in the JSON body, but I agree that seems like a lot more work.

I don't know if it matters that the V4 and V3 APIs have [separate rate limits](https://developer.github.com/v4/guides/resource-limitations/#:~:text=The%20GraphQL%20API%20v4%20limit,per%20hour%20for%20authenticated%20requests).

I'm moving this to a separate "ListTeamsV4" function with a new return type -- I agree that it's annoying that you can't get this information from the V3 API.

evankanderson:
Update: the refactor to use a single query for team + membership status was a bit more hairy than I expected, and I stalled out on the variety of test client interfaces in the peribolos test directory.

I don't expect I'll take another swing at this PR until end-of-week at the earliest.

PR #17900, opened 2020-06-09
remove unmaintained ARM tests from kind dashboard

BenTheElder:
I don't want these stale test results the maintainers have nothing to do with in our dashboard.

I don't know if they should be on testgrid at all, but that can be a follow-up

/cc @amwat @aojea @spiffxp


aojea:
Absolutely. 

PR is failing with following error

2020/06/09 23:10:27 FAIL: could not apply prowjob annotations: couldn't find dashboard "conformance-kind" for job "ci-kubernetes-conformance-image-1-15-test"

PR #17902, opened 2020-06-10
Add missing SELinux config to storage job

bertinatto:
Kops installs the upstream docker, and by default SELinux is not enabled.

CC @gnufied @jsafrane @hakman

### DIFF
--- a/config/jobs/kubernetes/sig-storage/sig-storage-kops-config.yaml
+++ b/config/jobs/kubernetes/sig-storage/sig-storage-kops-config.yaml
@@ -21,6 +21,7 @@ periodics:
       - --extract=ci/latest
       - --ginkgo-parallel
       - --kops-image=679593333241/CentOS Linux 7 x86_64 HVM EBS ENA 2002_01-b7ee8a69-ee97-4a49-9e68-afaee216db2e-ami-0042af67f8e4dcc20.4
+      - --kops-overrides=spec.docker.execOpt=--selinux-enabled
       - --kops-priority-path=/workspace/kubernetes/platforms/linux/amd64
       - --kops-ssh-user=centos
       - --kops-version=https://storage.googleapis.com/kops-ci/bin/latest-ci-updown-green.txt


bertinatto:
For reference, this `kops-overrides` is used here as well: https://github.com/kubernetes/test-infra/blob/a0e91d1bd566455f5c69c38baed6464e962b2e74/config/jobs/kubernetes/kops/kops-periodics-misc.yaml#L132

hakman:
I think think `spec.docker.execOpt` would have to be added to this function for this to work:
https://github.com/kubernetes/kops/blob/298f79659a2bd9c1d6548c6c95b335b5e29e6466/pkg/commands/set_cluster.go#L75

rifelpet:
That's correct, it should be a simple addition to `SetClusterFields`. Since `execOpt` is a list, you can use `cluster.spec.nodePortAccess` as an example in that function; appending values to that list. that way you could do something like `--kops-overrides=spec.docker.execOpt=foo,spec.docker.execOpt=bar` to end up with `[foo, bar]`.

Once that has been added to Kops, it should be usable by kubetest within a few hours. I'm having trouble finding docs that mention it but its possible you might want to omit the hyphens in `--selinux-enabled`.

bertinatto:
> setting the execOpt should work now that your kops PR has been merged and made it into the periodic kops jobs, I'm still not sure if you want the `--` in front of `selinux-enabled` but we can find out.
> 
> /lgtm

Thanks, @rifelpet! Do you know if there's a way we can verify that it worked? Is it possible to connect to a node from a running job to check if dockerd is running with `--selinux-enabled`?

rifelpet:
Unfortunately you can't connect to the running node. Your best bet will be any logs in the prow job or its artifacts. I'm looking at the log files from one of the nodes here: https://gcsweb.k8s.io/gcs/kubernetes-jenkins/logs/ci-kubernetes-e2e-storage-selinux/1270991318759247872/artifacts/ip-172-20-39-233.ap-southeast-1.compute.internal/ and I can't find anything that suggests that it is being set, but I also don't know exactly what to look for.

PR #17903, opened 2020-06-10
Rollout containerd to remaining jobs

jprzychodzen:
Continue rollout for kubernetes/perf-tests#1110

/sig scalability
/assign @mm4tt

Final rollout, as discussed, kubemark will not be updated.

### DIFF
--- a/config/jobs/kubernetes/sig-scalability/sig-scalability-experimental-periodic-jobs.yaml
+++ b/config/jobs/kubernetes/sig-scalability/sig-scalability-experimental-periodic-jobs.yaml
@@ -195,60 +195,44 @@ periodics:
           - --timeout=120m
           - --use-logexporter
 
-# This is a fork of ci-kubernetes-e2e-gci-gce-scalability job (as of 2020-02-18) with enabled containerd.
-- interval: 1h
-  name: ci-kubernetes-e2e-gci-gce-scalability-containerd
-  tags:
-    - "perfDashPrefix: gce-100Nodes-master-containerd"
-    - "perfDashJobType: performance"
-    - "perfDashBuildsCount: 500"
+  # Experimental job providing visibility into state of containerd.
+- interval: 6h
+  name: ci-kubernetes-e2e-gci-gce-scalability-containerd-correctness
   labels:
     preset-service-account: "true"
     preset-k8s-ssh: "true"
     preset-e2e-scalability-common: "true"
-    preset-e2e-scalability-containerd: "true"
   annotations:
     testgrid-dashboards: sig-scalability-experiments
-    testgrid-tab-name: gce-cos-master-scalability-100-containerd
+    testgrid-tab-name: gce-cos-master-scalability-100-containerd-conformance
   spec:
     containers:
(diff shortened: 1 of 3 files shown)


config/jobs/kubernetes/sig-scalability/sig-scalability-release-blocking-jobs.yaml
```diff
@@ -7,6 +7,7 @@ periodics:
     preset-service-account: "true"
     preset-k8s-ssh: "true"
     preset-e2e-scalability-common: "true"
+    preset-e2e-scalability-containerd: "true"
```
wojtek-t, line 10: I'm afraid of this one.
There may appear pushback on debugging and fixing issues by other sigs...

jprzychodzen: Instead of changing correctness 5k job to containerd, added experimental conformance 100 node job.

config/jobs/kubernetes/sig-scalability/sig-scalability-experimental-periodic-jobs.yaml
```diff
@@ -195,58 +195,37 @@ periodics:
...
-          - --test-cmd-args=--testconfig=testing/load/config.yaml
-          - --test-cmd-args=--testconfig=testing/access-tokens/config.yaml
-          - --test-cmd-args=--testoverrides=./testing/experiments/enable_prometheus_api_responsiveness.yaml
-          - --test-cmd-args=--testoverrides=./testing/experiments/enable_restart_count_check.yaml
-          - --test-cmd-args=--testoverrides=./testing/experiments/use_simple_latency_query.yaml
-          - --test-cmd-args=--testoverrides=./testing/overrides/load_throughput.yaml
-          - --test-cmd-name=ClusterLoaderV2
+          - --test=true
+          - --ginkgo-parallel=40
+          - --test_args=--ginkgo.focus=\[Conformance\] --ginkgo.skip=\[Serial\]|\[Disruptive\]|\[Flaky\]|\[Feature:.+\]|\[DisabledForLargeClusters\]
```
wojtek-t, line 227: What is the reason for making that conformance and not what we run in all other correctness tests?

jprzychodzen: I would like to slowly roll out containerd to full correctness suite (one that we run for docker), when it will have enough soaking time.

Starting as an experimental job with a limited set of tests seams like a reasonable point to start.

wojtek-t: Why can't we just make it consistent with all other correctness jobs. Given it's experimental, it's not big deal if it will start failing. And we will be able to uncover and delegate more problems at once when we find them.

jprzychodzen: Removed `focus`, making it match other correctness jobs.

PR #17904, opened 2020-06-10
[git] Add the ability to fetch from different remote

droslean:
Add the ability to fetch from a different remote using the existing user and pass.

### DIFF
--- a/prow/git/git.go
+++ b/prow/git/git.go
@@ -488,3 +488,22 @@ func (i *Repo) ShowRef(commitlike string) (string, error) {
 	}
 	return strings.TrimSpace(string(out)), nil
 }
+
+// FetchFromRemote runs a git fetch command for the specified remote.
+// Pass the destination as org/repo format since the actual URI will be constructed
+// using the existing credentials and will always use the https protocol.
+func (r *Repo) FetchFromRemote(orgRepo, branch string) error {
+	r.logger.Infof("Fetching from '%s (branch: %s)'.", orgRepo, branch)
+
+	remoteURI := fmt.Sprintf("https://%s/%s", r.host, orgRepo)
+	if r.user != "" && r.pass != "" {
+		remoteURI = fmt.Sprintf("https://%s:%s@%s/%s", r.user, r.pass, r.host, orgRepo)
+	}
+
+	co := r.gitCommand("fetch", remoteURI, branch)
+	out, err := co.CombinedOutput()
+	if err != nil {
+		return fmt.Errorf("Fetching failed, output: %q, error: %v", string(out), err)
+	}
+	return nil
+}
--- a/prow/git/v2/interactor.go
+++ b/prow/git/v2/interactor.go
@@ -278,6 +278,20 @@ func (i *interactor) FetchRef(refspec string) error {
 	return nil
 }
 
+// FetchFromRemote fetches all update from a specific remote and branch and leaves it as FETCH_HEAD.
+func (i *interactor) FetchFromRemote(remote RemoteResolver, branch string) error {
+	r, err := remote()
+	if err != nil {
+		return fmt.Errorf("couldn't get remote: %v", err)
+	}
+
+	i.logger.Infof("Fetching %s from %s", branch, r)
+	if out, err := i.executor.Run("fetch", r, branch); err != nil {
+		return fmt.Errorf("error fetching %s from %s: %v %v", branch, r, err, string(out))
+	}
+	return nil
+}
+
 // CheckoutPullRequest fetches the HEAD of a pull request using a synthetic refspec
 // available on GitHub remotes and creates a branch at that commit.
 func (i *interactor) CheckoutPullRequest(number int) error {
--- a/prow/git/v2/interactor_test.go
+++ b/prow/git/v2/interactor_test.go
@@ -18,9 +18,11 @@ package git
 
 import (
 	"errors"
-	"github.com/sirupsen/logrus"
+	"fmt"
 	"reflect"
 	"testing"
+
+	"github.com/sirupsen/logrus"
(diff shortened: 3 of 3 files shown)


prow/git/git.go
```diff
@@ -141,6 +141,43 @@ func (c *Client) unlockRepo(repo string) {
 	c.repoLocks[repo].Unlock()
 }
 
+// Init create an empty Git repository and returns a Repo
+func (c *Client) Init(organization, repository string) (*Repo, error) {
```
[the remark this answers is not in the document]
droslean: Removed. Thanks

prow/git/git.go
```diff
@@ -488,3 +525,38 @@ func (i *Repo) ShowRef(commitlike string) (string, error) {
 	}
 	return strings.TrimSpace(string(out)), nil
 }
+
+// Fetch runs a git fetch command for the specified remote. Pass the remote as org/repo format
+// since the actual URI will be constructed using the existing credentials.
+func (r *Repo) Fetch(remote, branch string) error {
```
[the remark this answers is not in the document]
alvaroaleman: Yeah it seems there is no way to use different remotes in either the old or the new clients. I would probably rename it to `FetchFromDifferentRemote` or sth but generallu think this is fine.

Should have tests though and get implemented into the v2 client as well :)

prow/git/git.go
```diff
@@ -488,3 +488,17 @@ func (i *Repo) ShowRef(commitlike string) (string, error) {
 	}
 	return strings.TrimSpace(string(out)), nil
 }
+
+// FetchFromRemote runs a git fetch command for the specified remote. Pass the remote as org/repo format
+// since the actual URI will be constructed using the existing credentials.
+func (r *Repo) FetchFromRemote(remote, branch string) error {
+	r.logger.Infof("Fetching from '%s (branch: %s)'.", remote, branch)
+	remoteURI := fmt.Sprintf("https://%s:%s@%s/%s", r.user, r.pass, r.host, remote)
```
alvaroaleman, line 496: `org/repo` is not a `remote`, can you rename the var to `orgRepo` and mention in the godocs that this will always use https?

droslean: addressed

prow/git/v2/interactor.go
```diff
@@ -278,6 +282,32 @@ func (i *interactor) FetchRef(refspec string) error {
 	return nil
 }
 
+// FetchFromRemote fetches all update from a specific remote and branch and leaves it as FETCH_HEAD.
+func (i *interactor) FetchFromRemote(remote RemoteResolver, branch string) error {
+	r, err := remote()
```
stevekuznetsov, line 287: We only read this value once and don't store it -- why not just pass the URI as a string? What benefit is it to pass a resolver?

prow/git/v2/interactor.go
```diff
@@ -278,6 +282,32 @@ func (i *interactor) FetchRef(refspec string) error {
...
+	if err != nil {
+		return fmt.Errorf("couldn't get remote: %v", err)
+	}
+
+	// Parse the remote's URI to censor any existing tokens.
+	u, err := url.Parse(r)
```
stevekuznetsov, line 293: IMO this should be on the user to do correctly by adding their tokens to the secret agent at startup

prow/git/v2/interactor.go
```diff
@@ -278,6 +282,32 @@ func (i *interactor) FetchRef(refspec string) error {
...
+	if err != nil {
+		return fmt.Errorf("couldn't parse remote url: %v", err)
+	}
+
+	if token, ok := u.User.Password(); ok {
+		i.logger.Logger.SetFormatter(logrusutil.NewCensoringFormatter(logrus.StandardLogger().Formatter, func() sets.String {
```
stevekuznetsov, line 299: As written this code _overwrites_ the other formatters.

prow/git/git.go
```diff
@@ -488,3 +488,18 @@ func (i *Repo) ShowRef(commitlike string) (string, error) {
 	}
 	return strings.TrimSpace(string(out)), nil
 }
+
+// FetchFromRemote runs a git fetch command for the specified remote.
+// Pass the destination as org/repo format since the actual URI will be constructed
+// using the existing credentials and will always use the https protocol.
+func (r *Repo) FetchFromRemote(orgRepo, branch string) error {
+	r.logger.Infof("Fetching from '%s (branch: %s)'.", orgRepo, branch)
+	remoteURI := fmt.Sprintf("https://%s:%s@%s/%s", r.user, r.pass, r.host, orgRepo)
```
alvaroaleman, line 497: Interesting, didn't know this works without the `.git` suffix
/lgtm
/approve

alvaroaleman: /hold

prow/git/v2/interactor.go
droslean: Yea, we could do that. Since I already here parse the url.

prow/git/v2/interactor.go
alvaroaleman: /lgtm cancel
Missed this

droslean: We can remove this if we rely on user to censor the secrets.

prow/git/git.go
alvaroaleman: This will break if user or password are unset, no? Do we generally assume that they are always set?

droslean: Adding a check for this won't hurt. Addressing...

droslean:
Oops, messed up the rebase.

prow/git/git.go
droslean: Addressed

PR #17912, opened 2020-06-10
Prevent release note duplicate messages

### LINKED ISSUE #16971: repeated commenting on release-note-needed when cherry pick is filed
**What happened**:

Screen Shot 2020-03-24 at 4 34 01 PM

I received emails for each of these

We should make this comment once. Besides unnecessary notifications this must be costing us API tokesn

File a cherry pick PR to kubernetes/kubernetes

https://github.com/kubernetes/kubernetes/pull/89448
https://github.com/kubernetes/kubernetes/pull/89446

/area prow

jherrera123:
Fixes #16971

**Special notes for your reviewer**:
When you open a PR, multiple plugins add labels to the PR. Each label that is added, is triggering a call to handlePR in the release note plugin, adding a new release note comment to the issue.

The way I'm fixing this issue is by limiting the label add trigger to only labels related to release note plugin.

I added unit tests, and also tested the fix on a gke prow instance connected to a github repo.

/area prow/plugins

/assign @BenTheElder

### DIFF
--- a/prow/plugins/releasenote/releasenote.go
+++ b/prow/plugins/releasenote/releasenote.go
@@ -212,32 +212,29 @@ func handlePullRequest(pc plugins.Agent, pr github.PullRequestEvent) error {
 	return handlePR(pc.GitHubClient, pc.Logger, &pr)
 }
 
-func shouldIgnorePR(pr *github.PullRequestEvent) bool {
+func shouldHandlePR(pr *github.PullRequestEvent) bool {
 	// Only consider events that edit the PR body or add a label
 	if pr.Action != github.PullRequestActionOpened &&
 		pr.Action != github.PullRequestActionEdited &&
 		pr.Action != github.PullRequestActionLabeled {
-		return true
+		return false
 	}
 
 	// Ignoring unrelated PR labels prevents duplicate release note messages
 	if pr.Action == github.PullRequestActionLabeled {
-		isRNLabelAdd := false
 		for _, rnLabel := range allRNLabels {
 			if pr.Label.Name == rnLabel {
-				isRNLabelAdd = true
+				return true
 			}
 		}
-
-		if !isRNLabelAdd {
-			return true
-		}
+		return false
 	}
-	return false
+
+	return true
 }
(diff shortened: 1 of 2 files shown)


prow/plugins/releasenote/releasenote.go
```diff
@@ -212,11 +212,32 @@ func handlePullRequest(pc plugins.Agent, pr github.PullRequestEvent) error {
 	return handlePR(pc.GitHubClient, pc.Logger, &pr)
 }
 
-func handlePR(gc githubClient, log *logrus.Entry, pr *github.PullRequestEvent) error {
+func shouldIgnorePR(pr *github.PullRequestEvent) bool {
```
alvaroaleman, line 215: Nit: The fact that a `true` return means the PR will be ignored instead of having a `shouldHandlePR` makes this hard to read for me at this time

prow/plugins/releasenote/releasenote.go
```diff
@@ -212,11 +212,32 @@ func handlePullRequest(pc plugins.Agent, pr github.PullRequestEvent) error {
...
 		pr.Action != github.PullRequestActionLabeled {
+		return true
+	}
+
+	// Ignoring unrelated PR labels prevents duplicate release note messages
+	if pr.Action == github.PullRequestActionLabeled {
+		isRNLabelAdd := false
+		for _, rnLabel := range allRNLabels {
+			if pr.Label.Name == rnLabel {
+				isRNLabelAdd = true
```
alvaroaleman, line 228: Nit: You can just `return false` here and save the variable declaration above and the check below

prow/plugins/releasenote/releasenote.go
jherrera123: Method has been changed to shouldHandlePR

prow/plugins/releasenote/releasenote.go
jherrera123: I removed the variable and I'm returning true, since the logic was changed to shouldHandlePR

PR #17916, opened 2020-06-11
Add kube-state-metrics to image-pushing job

lilic:
I tried to follow the README, if I understand correctly this is all we need, advice more than welcome! :)

Background https://github.com/kubernetes/kube-state-metrics/issues/1089

cc @tariq1890 @brancz @serathius


spiffxp:
/approve
/lgtm
I would recommend merging https://github.com/kubernetes/kube-state-metrics/pull/1155 because this job will fail until there's a cloudbuild.yaml file available

PR #17920, opened 2020-06-11
Make autobump create Gerrit PRs and support repos hosting multiple Prow instances.

cjwagner:
Instead of using a [commit message hook](https://gerrit-review.googlesource.com/Documentation/cmd-hook-commit-msg.html) to generate a random `Change-Id` for the Gerrit PR, we use a change ID deterministically generated from
```bash
$( echo "${GERRIT_HOST_REPO}; ${PROW_INSTANCE_NAME}; ${old_version}" | git hash-object --stdin)
```
so that any existing open bump PR is updated with new patchsets rather than creating a new PR every run.
An http cookiefile is used for push credentials rather than a GH personal access token.

/assign @fejta 
I've tested this to confirm that it works.

### DIFF
--- a/prow/cmd/autobump/autobump.sh
+++ b/prow/cmd/autobump/autobump.sh
@@ -19,24 +19,31 @@ set -o nounset
 set -o pipefail
 
 PLANK_DEPLOYMENT_FILE="${PLANK_DEPLOYMENT_FILE:-}"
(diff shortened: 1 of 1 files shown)


cjwagner:
Thanks for the review Matthias!
/hold cancel

PR #17924, opened 2020-06-12
Switch slow jobs to use docker due to containerd performance regression

msau42:
https://github.com/kubernetes/kubernetes/issues/92045

Tested in https://github.com/kubernetes/kubernetes/pull/92056 to verify using docker fixes the job timeout.


BenTheElder:
/lgtm
/approve
thanks!

PR #17927, opened 2020-06-12
Enable bucket whitelist in Deck

clarketm:
Enable bucket/folder whitelist feature added in https://github.com/kubernetes/test-infra/pull/17650


clarketm:
@e-blackwelder handling in #18001

PR #17930, opened 2020-06-14
Fix bugs of hmac tool and add more logs

chizhg:
1. Fix some nil errors by initializing the map correctly

2. Skip the global token when collecting repos to be removed, as we never want to delete it

3. The global hmac token got from the secret in the old format can possibly have an extra `\n`, use `strings.TrimSpace` to remove it before writing the token back to the secret

4. Change the generated token length to 20 bytes (a string of length 40), to follow what we have in the manual deployment instruction - https://github.com/kubernetes/test-infra/blob/master/prow/getting_started_deploy.md#create-the-github-secrets

5. In the org level webhook events, the `Repo` field can be possibly empty, see https://developer.github.com/webhooks/event-payloads/#webhook-payload-object-common-properties. So fall back to the `Org` field if `event.Repo.FullName` is empty to have a better support for org level webhook events.

6. Add more debug logs in deleting/updating the webhooks

This time I tested it for all possible scenarios on my local with `--dry-run=false` and it works well.

/assign @cjwagner 
/cc @cjwagner

### DIFF
--- a/prow/cmd/hmac/main.go
+++ b/prow/cmd/hmac/main.go
@@ -125,6 +125,13 @@ func main() {
 	if err := agent.Start([]string{o.github.TokenPath}); err != nil {
 		logrus.WithError(err).Fatalf("Error starting secret agent %s", o.github.TokenPath)
 	}
+
+	var configAgent config.Agent
+	if err := configAgent.Start(o.configPath, ""); err != nil {
+		logrus.WithError(err).Fatal("Error starting config agent.")
+	}
+	newHMACConfig := configAgent.Config().ManagedWebhooks
+
 	gc, err := o.github.GitHubClient(agent, o.dryRun)
 	if err != nil {
 		logrus.WithError(err).Fatal("Error creating github client")
@@ -137,6 +144,12 @@ func main() {
 
 	currentHMACMap := map[string]github.HMACsForRepo{}
 	if err := yaml.Unmarshal(currentHMACYaml, &currentHMACMap); err != nil {
+		// When the token is still a single global token, respect_legacy_global_token must be set to true before running this tool.
+		// This can prevent the global token from being deleted by mistake before users migrate all repos/orgs to use auto-generated private tokens.
+		if !newHMACConfig.RespectLegacyGlobalToken {
+			logrus.Fatal("respect_legacy_global_token must be set to true before the hmac tool is run for the first time.")
+		}
+
 		logrus.WithError(err).Error("Couldn't unmarshal the hmac secret as hierarchical file. Parsing as a single global token and writing it back to the secret.")
 		currentHMACMap["*"] = github.HMACsForRepo{
 			github.HMACToken{
@@ -144,12 +157,6 @@ func main() {
 			},
 		}
 	}
-
-	var configAgent config.Agent
-	if err := configAgent.Start(o.configPath, ""); err != nil {
-		logrus.WithError(err).Fatal("Error starting config agent.")
-	}
-	newHMACConfig := configAgent.Config().ManagedWebhooks
 
 	c := client{
 		kubernetesClient: kc,
@@ -171,7 +178,7 @@ func (c *client) handleConfigUpdate() error {
 	repoRemoved := map[string]bool{}
 	repoRotated := map[string]config.ManagedWebhookInfo{}
 
-	for repoName, hmacConfig := range c.newHMACConfig {
+	for repoName, hmacConfig := range c.newHMACConfig.OrgRepoConfig {
 		if _, ok := c.currentHMACMap[repoName]; ok {
 			repoRotated[repoName] = hmacConfig
 		} else {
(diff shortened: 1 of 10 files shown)


prow/cmd/hmac/main.go
```diff
@@ -178,6 +180,10 @@ func (c *client) handleConfigUpdate() error {
 	}
 
 	for repoName := range c.currentHMACMap {
+		// Skip the global hmac token as it should never be deleted.
+		if repoName == "*" {
+			continue
```
cjwagner, line 185: I don't think this is what we want. It should be possible to remove the global token if it is removed from the golden config. We only want to support the global token until all repos have been migrated to more appropriately scoped tokens. This is important since supporting the global token results in a small window during webhook registration where the global token is respected, even if the repo was never configured to send webhooks using the global token. This is something we'll have to live with during migration, but should be avoided after.

Here is what I'm thinking:
1. The global token value should be determined when migrating from the old hmac secret format (just the raw hmac value) to the new format assuming that `*` exists in the golden config.
1. It should not be rotated based on `tokenCreatedAfter` since we can't rotate the global token.
1. It should be removed when removed from the golden config.

chizhg: Based on our previous discussion, and the [design doc](https://docs.google.com/document/d/1we2DgZd42U0PRhB-XsvP2FJjLHX5dzJPEbXC6gqWkn0/edit#), I think we don't need to have `*` in the golden config:

1. As you mentioned, we cannot rotate the global token based on `tokenCreatedAfter` in the golden config. If we add `*` in the golden config, we'll have to add a comment saying it cannot be rotated, it can be confusing and also needs to be applied on all other Prow instances.

2. After we fully roll out this process, we can change the instructions in https://github.com/kubernetes/test-infra/blob/master/prow/getting_started_deploy.md#create-the-github-secrets to ask users create a secret with the new `*` format, and mention there are two ways to configure the webhook - one is using the global token, and the other is using the hmac management tool.

One issue with this approach is the Prow maintainers will have to manually delete the global token when it's unused. So I'm thinking about another way of handling it - how about we add a new field like `using_global_hmac_token` in the golden config? It will default to be `true`, but can be set to `false` so it's can be deleted by the hmac tool.

The format I'm thinking of is:
```
managed_webhooks:
  using_global_hmac_token: false
  config:
    qux:
      tokenCreatedAfter: 2017-10-02T15:00:00Z
    foo/bar:
      tokenCreatedAfter: 2018-10-02T15:00:00Z
    foo/baz:
      tokenCreatedAfter: 2019-10-02T15:00:00Z
```

cjwagner: Using a separate field in the golden config instead of including `*` sounds good to me. I think we should default to not supporting the global token though since that will be the desired default once existing instances have migrated. We may also want to include `legacy` in the field name to hint that it should be avoided (e.g. `respect_legacy_global_hmac`).

chizhg: Sounds good!
I'll make changes in the design doc and corresponding fix here today.

prow/github/hmac.go
```diff
@@ -55,7 +55,12 @@ func ValidatePayload(payload []byte, sig string, tokenGenerator func() []byte) b
 		return false
 	}
 
-	hmacs, err := extractHMACs(event.Repo.FullName, tokenGenerator)
+	orgRepo := event.Repo.FullName
+	// If orgRepo is empty, the event is probably org-level, so try getting org name from the Org info.
+	if orgRepo == "" {
+		orgRepo = event.Org.Login
```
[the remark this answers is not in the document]
chizhg: Added unit tests for both "org level event" and "repo level event" cases.

prow/cmd/hmac/main.go
chizhg: I have added the `respect_legacy_global_hmac` field based on our discussions here.

However, when I tested with the new format of config, I mistakenly ran the tool with an empty `managed_webhooks` configuration, and it unfortunately deleted my global token since `respect_legacy_global_hmac` is defaulted to be false. I have added some explanation for this field in the `README`, but I'm still worried that other users can possibly make the same disruptive mistake. So I'm now more inclined to my initial idea - set the field name to `using_global_token` and default it to be `true`. It is safer and will be a no-op if users do not configure anything.

WDYT?

cjwagner: I'd prefer for the default to be appropriate for the common case, not the one time migration.
I think we can prevent that mistake if we abort and warn if the `managed_webhooks` config is omitted or if the config is in the old format (just the raw global token) and `respect_legacy_global_hmac` is unset.

chizhg: Great idea! Done.

Please note I have changed `tokenCreatedAfter` to `token_created_after` in the config file as most other field names are using snake case. Now an example configuration is like below:
```yaml
managed_webhooks:
  respect_legacy_global_token: true
  org_repo_config:
    qux:
      token_created_after: 2017-10-02T15:00:00Z
    foo/bar:
      token_created_after: 2018-10-02T15:00:00Z
    foo/baz:
      token_created_after: 2019-10-02T15:00:00Z
```

PR #17933, opened 2020-06-14
kubeadm: increase job timeout to 60 minutes

neolit123:
Attempt to resolve recent, occasional flakes where the overall job
runtime of 40 minutes is not sufficient.

xref https://github.com/kubernetes/kubernetes/issues/90761#issuecomment-638877266

/sig cluster-lifecycle
/kind bug
/priority important-soon


neolit123:
looking for LGTM, thanks.
/cc @hasheddan @fabriziopandini

PR #17939, opened 2020-06-15
hook only unmarshals the hmac token secret when it's changed

chizhg:
As discussed in https://kubernetes.slack.com/archives/CDECRSC5U/p1589893627262700 and synced offline with @fejta:

1. Adding a map named `secretRevisionsMap` in the secret Agent to keep track of the revisions for each secret, and increment the revision by 1 for the secret when it's reloaded.

2. hook maintains the current revision of the hmac token map. In each http request, comparing the current and latest revision, and only reloading and unmarshalling the secret if they are not the same. This will reduce the frequency we unmarshal the secret string into the map.

/assign @fejta @alvaroaleman 

(BTW sorry for the delayed response to this issue.)

### DIFF
--- a/prow/config/secret/agent.go
+++ b/prow/config/secret/agent.go
@@ -33,7 +33,8 @@ import (
 // Agent watches a path and automatically loads the secrets stored.
 type Agent struct {
 	sync.RWMutex
-	secretsMap map[string][]byte
+	secretsMap         map[string][]byte
+	secretRevisionsMap map[string]uint32
 }
 
 // Start creates goroutines to monitor the files that contain the secret value.
@@ -46,9 +47,11 @@ func (a *Agent) Start(paths []string) error {
 	}
 
 	a.secretsMap = secretsMap
+	a.secretRevisionsMap = map[string]uint32{}
 
 	// Start one goroutine for each file to monitor and update the secret's values.
 	for secretPath := range secretsMap {
+		a.secretRevisionsMap[secretPath] = 1
 		go a.reloadSecret(secretPath)
 	}
 
@@ -65,6 +68,7 @@ func (a *Agent) Add(path string) error {
 	}
 
 	a.setSecret(path, secret)
+	a.secretRevisionsMap[path] = 1
 
 	// Start one goroutine for each file to monitor and update the secret's values.
 	go a.reloadSecret(path)
@@ -108,10 +112,10 @@ func (a *Agent) reloadSecret(secretPath string) {
 }
 
 // GetSecret returns the value of a secret stored in a map.
-func (a *Agent) GetSecret(secretPath string) []byte {
+func (a *Agent) GetSecret(secretPath string) ([]byte, uint32) {
 	a.RLock()
 	defer a.RUnlock()
-	return a.secretsMap[secretPath]
+	return a.secretsMap[secretPath], a.secretRevisionsMap[secretPath]
 }
 
 // setSecret sets a value in a map of secrets.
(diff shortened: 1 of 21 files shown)


prow/hook/server.go
```diff
@@ -53,7 +58,23 @@ type Server struct {
 
 // ServeHTTP validates an incoming webhook and puts it into the event channel.
 func (s *Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
-	eventType, eventGUID, payload, ok, resp := github.ValidateWebhook(w, r, s.TokenGenerator)
+	tokenSecret, revision := s.TokenGenerator()
+	if revision != atomic.LoadUint32(&s.currentTokenRevision) {
+		atomic.StoreUint32(&s.currentTokenRevision, revision)
```
alvaroaleman, line 63: I don't think  we should store the revision before we updated the map, especially since the unmarshaling might fail. IMHO we should simply use a `sync.RWMutex` here require the write lock for the whole update section

chizhg: The main purpose of this PR is to reduce the frequency of unmarshalling the secret string into a map - only do unmarshall when the secret changes rather than in every webhook request.

If we only use a `sync.RWMutex` but do not store the revision, we won't be able to know when the secret is changed thus cannot implement what we want above. Or are you having other ideas in mind?

alvaroaleman: I didn't mean that we should not store the revision, I meant that we should only update the revision after we have updated the value. Right now it may happen that we bump the revision and then fail or that concurrent requests will fail because we bumped the revision already but didn't yet bump the value

chizhg: Ah I see, that is a bug... I have made the change and switched to `sync.RWMutex`.

Thanks!

fejta:
Rebase please?

chizhg:
> Rebase please?

Done, thanks for reminding.

prow/github/hmac.go
```diff
@@ -23,12 +23,43 @@ import (
...
+type HMACTokenResolver struct {
+	TokenGenerator       func() ([]byte, uint32)
+	currentRepoTokenMap  map[string]HMACsForRepo
+	currentTokenRevision uint32
+	sync.RWMutex
+}
+
+// Get will return the latest hmac token map, it only unmarshalls the secret string into a map if it's changed.
+func (s *HMACTokenResolver) Get() map[string]HMACsForRepo {
+	tokenSecret, revision := s.TokenGenerator()
```
fejta, line 44: RLock guard this, please

prow/config/secret/agent.go
```diff
@@ -33,7 +33,8 @@ import (
 // Agent watches a path and automatically loads the secrets stored.
 type Agent struct {
 	sync.RWMutex
-	secretsMap map[string][]byte
+	secretsMap         map[string][]byte
+	secretRevisionsMap map[string]uint32
```
fejta, line 37: Why not just int?

This seems overspecified. If you are going to specify it, when limit it to 32 bits?

prow/config/secret/agent.go
```diff
@@ -108,22 +112,31 @@ func (a *Agent) reloadSecret(secretPath string) {
 }
 
 // GetSecret returns the value of a secret stored in a map.
-func (a *Agent) GetSecret(secretPath string) []byte {
+func (a *Agent) GetSecret(secretPath string) ([]byte, uint32) {
```
fejta, line 115: Rather than changing this method's signature, how about using a subscriber pattern?

PR #17942, opened 2020-06-15
Validate github tokens when reading and improve GitHub API error handling

JensErat:
Per accident, I stored some YAML contents instead of a GitHub token to the secrets file, and got weird log messages reading like:

```
DEBU[0001] Retrying request due to connection problem    backoff=2s client=github new-endpoint="https://git.daimler.com/api/v3" old-endpoint="https://git.daimler.com/api/v3"
DEBU[0003] Retrying request due to connection problem    backoff=4s client=github new-endpoint="https://git.daimler.com/api/v3" old-endpoint="https://git.daimler.com/api/v3"
DEBU[0007] Retrying request due to connection problem    backoff=8s client=github new-endpoint="https://git.daimler.com/api/v3" old-endpoint="https://git.daimler.com/api/v3"
DEBU[0015] Retrying request due to connection problem    backoff=16s client=github new-endpoint="https://git.daimler.com/api/v3" old-endpoint="https://git.daimler.com/api/v3"
```

I started debugging networking issues, and only succeeded in determining the wrong file contents when analyzing request and error message in the go debugger. Just extending the error message also did not seem appropriate, since this exposes the file's content in plain text in the logs, so I also added basic input validation when reading the token from a file.

### validate secret token syntax

The secret token is directly passed through from file contents to HTTP
headers (with only sanitizing by trimming space). In my case, erronously
YAML contents in there resulted in weird client-side connection issues.

From my understanding [RFC 6750] has the most narrow definition of how a
token looks like. This commit will fail early if a token is syntactically
wrong.

[RFC 6750]: https://tools.ietf.org/html/rfc6750#section-2.1

### print details on connection issues

The code expects all connection issues to be a generic network or server
side issue that can be retried. Golang also returns an error here for
client-side issues like invalid header values, hiding the actual cause.
Pass-through the error message instead to help the operator debugging,
but keep the retry approach.

### DIFF
--- a/prow/config/secret/secret.go
+++ b/prow/config/secret/secret.go
@@ -20,7 +20,12 @@ import (
 	"bytes"
 	"fmt"
 	"io/ioutil"
+	"regexp"
 )
+
+// base64token/token68 following RFC 6750
+var tokenRegex = regexp.MustCompile(`^[-\w._~+/]+$`)
+
 
 // LoadSecrets loads multiple paths of secrets and add them in a map.
 func LoadSecrets(paths []string) (map[string][]byte, error) {
@@ -42,5 +47,11 @@ func LoadSingleSecret(path string) ([]byte, error) {
 	if err != nil {
 		return nil, fmt.Errorf("error reading %s: %v", path, err)
 	}
-	return bytes.TrimSpace(b), nil
+
+	secret := bytes.TrimSpace(b)
+
+	if !tokenRegex.Match(secret) {
+		return nil, fmt.Errorf("error reading %s: invalid token format", path)
+	}
+	return secret, nil
 }
--- a/prow/config/secret/secret.go
+++ b/prow/config/secret/secret.go
@@ -26,7 +26,6 @@ import (
 // base64token/token68 following RFC 6750
 var tokenRegex = regexp.MustCompile(`^[-\w._~+/]+$`)
 
-
 // LoadSecrets loads multiple paths of secrets and add them in a map.
 func LoadSecrets(paths []string) (map[string][]byte, error) {
 	secretsMap := make(map[string][]byte, len(paths))
@@ -51,7 +50,7 @@ func LoadSingleSecret(path string) ([]byte, error) {
 	secret := bytes.TrimSpace(b)
 
 	if !tokenRegex.Match(secret) {
-		return nil, fmt.Errorf("error reading %s: invalid token format", path)
+		return nil, fmt.Errorf("error reading %s: invalid token format (does not match `%v`)", path, tokenRegex)
 	}
 	return secret, nil
 }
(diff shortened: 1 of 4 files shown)


prow/config/secret/secret_test.go
```diff
@@ -0,0 +1,48 @@
...
+	"reflect"
+	"testing"
+)
+
+func TestLoadSingleSecret(t *testing.T) {
+	tests := []struct {
+		name    string
+		content string
+		want    []byte
+		wantErr bool
```
alvaroaleman, line 16: Please do not use booleans for this but a concrete error or error message, otherwise the tests might start rotting when the Loading fails due to a different reason

prow/config/secret/secret_test.go
```diff
@@ -0,0 +1,48 @@
...
+ 	tempSecret := filepath.Join(secretDir, "tempSecret")
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if err := ioutil.WriteFile(tempSecret, []byte(tt.content), 0666); err != nil {
+				t.Fatalf("fail to write secret: %v", err)
+			}
+			got, err := LoadSingleSecret(tempSecret)
+			if (err != nil) != tt.wantErr {
+				t.Errorf("LoadSingleSecret() error = %v, wantErr %v", err, tt.wantErr)
```
alvaroaleman, line 40: Use `t.Fatalf` rather than `Errorf` and `return`

prow/config/secret/secret.go
```diff
@@ -42,5 +47,11 @@ func LoadSingleSecret(path string) ([]byte, error) {
 	if err != nil {
 		return nil, fmt.Errorf("error reading %s: %v", path, err)
 	}
-	return bytes.TrimSpace(b), nil
+
+	secret := bytes.TrimSpace(b)
+
+	if !tokenRegex.Match(secret) {
+		return nil, fmt.Errorf("error reading %s: invalid token format", path)
```
alvaroaleman, line 54: Another nit: Could you print the regex here, someting like `invalid token format: Does not match regex \`^...\``?

JensErat:
Fixed test and review findings. Do you agree with the substring test for the error messages? Comparing entire `error`s (which would enable to expect nil) is not easily possible with the path included in the message. Bazel tests run fine locally now.

JensErat:
@alvaroaleman we (to be more specific, @sbueringer) realized while the check works fine for the "local dev tools" and didn't break any tests, there are two kind of secrets: those like the github token being a plain ASCII string, and others like the HMAC token stored in YAML-structured files. This PR broke some containers like hook.

I'm currently putting together a new commit that checks in `flagutil/github.go` instead which is more specific, and will also have a look at providing at test for reading HMAC-token-like structured secrets.

alvaroaleman:
@JensErat okay, since this moves the code elsewhere anyways, can you please create a revert for this PR so we unbreak master?

PR #17943, opened 2020-06-15
Add descheduler e2e to prow jobs

damemi:
Fixes https://github.com/kubernetes-sigs/descheduler/issues/323

### DIFF
--- a/config/jobs/kubernetes-sigs/descheduler/descheduler-presubmits-master.yaml
+++ b/config/jobs/kubernetes-sigs/descheduler/descheduler-presubmits-master.yaml
@@ -52,3 +52,17 @@ presubmits:
         - make
         args:
         - test-unit
+  - name: pull-descheduler-e2e
+    annotations:
+      testgrid-dashboards: sig-scheduling
+      testgrid-tab-name: pull-descheduler-e2e
+    decorate: true
+    path_alias: sigs.k8s.io/descheduler
+    always_run: true
+    spec:
+      containers:
+      - image: golang:1.13
+        command:
+        - make
+        args:
+        - test-e2e


damemi:
Like I mentioned in https://github.com/kubernetes-sigs/descheduler/issues/323, is there a way to have this test run against multiple k8s versions?

/cc @ingvagabund @seanmalloy

ingvagabund:
Possibly a different form of https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes-sigs/kind/kind-presubmits.yaml#L298-L338 ?

damemi:
@seanmalloy @ingvagabund since we're still seeing problems with Travis running in our repo, do we want to merge this so that we are at least getting *some* e2es to run, and then we can figure out how to get additional versions? Not having any e2es running right now is critical imo

seanmalloy:
> @seanmalloy @ingvagabund since we're still seeing problems with Travis running in our repo, do we want to merge this so that we are at least getting *some* e2es to run, and then we can figure out how to get additional versions? Not having any e2es running right now is critical imo

@damemi I'm good with merging this so we at least get some e2e tests running. I'm not an org member so I can't add a lgtm.

ingvagabund:
I don't think this will be enough since `make test-e2e` assumes there's a cluster already running. See https://github.com/kubernetes-sigs/descheduler/blob/master/.travis.yml.

damemi:
> I don't think this will be enough since `make test-e2e` assumes there's a cluster already running. See https://github.com/kubernetes-sigs/descheduler/blob/master/.travis.yml.

Good point, maybe there is a test image provided that we can use that sets up a running cluster. I see a lot of other jobs use [kubekins-e2e](https://github.com/kubernetes/test-infra/search?q=kubekins-e2e&unscoped_q=kubekins-e2e) (when searching for e2e tests in this repo), but I don't know anything about this image

PR #17946, opened 2020-06-15
[wip] secret-sync-controller experiment

b01901143:
parse config into source&dest pairs
access latest version of source and dest secrets


b01901143:
Done.
Amended the previous commit.

experiment/secret-sync-controller/secret_sync_spec.go
```diff
@@ -0,0 +1,91 @@
...
+	Namespace string `yaml:"namespace"`
+	Secret    string `yaml:"secret"`
+}
+
+type SecretManagerSpec struct {
+	Project string `yaml:"project"`
+	Secret  string `yaml:"secret"`
+}
+
+func (target Target) GetLatestSecretVersion(k8s_clientset *kubernetes.Clientset, secretManager_ctx context.Context, secretManager_client *secretmanager.Client) (int, map[string][]byte) {
```
spiffxp, line 42: snake_case isn't idiomatic go, prefer camelCase
```suggestion
func (target Target) GetLatestSecretVersion(k8sClientSet *kubernetes.Clientset, secretManagerCtx context.Context, secretManagerClient *secretmanager.Client) (int, map[string][]byte) {
```

experiment/secret-sync-controller/secret_sync_spec.go
```diff
@@ -0,0 +1,91 @@
...
+
+type SecretSyncSpec struct {
+	// the target can be either a K8s secret or a SecretManager secret
+	Source      Target `yaml:"source"`
+	Destination Target `yaml:"destination"`
+}
+
+type Target struct {
+	// assert that one of the two should be `nil`
+	Kubernetes    *KubernetesSpec    `yaml:"kubernetes"`
```
spiffxp, line 28: Allow for these fields to not exist
```suggestion
	Kubernetes    *KubernetesSpec    `yaml:"kubernetes,omitempty"`
```

experiment/secret-sync-controller/secret_sync_spec.go
spiffxp: I would instead make `ctx context.Context` your first parameter, and plumb that through (instead of making a `context.TODO()` in the function below)

experiment/secret-sync-controller/client_util/client_util.go
```diff
@@ -0,0 +1,46 @@
+package client_util
```
spiffxp, line 1: avoid snake case, consider instead making this something like `secret-sync-controller/client/client.go`

experiment/secret-sync-controller/README.md
```diff
@@ -0,0 +1,71 @@
...
+## Prerequisites
+- Create a gke cluster.
+
+- [Create service account for app](https://cloud.google.com/docs/authentication/production#command-line)
+
+- Grant required permissions to the service account `service-account-name`.
+
+	- Permission to get clusters:
+
+		    gcloud projects add-iam-policy-binding k8s-jkns-gke-soak --member "serviceAccount:<service-account-name>@k8s-jkns-gke-soak.iam.gserviceaccount.com" --role "roles/container.clusterViewer"
```
spiffxp, line 18: nit: replace occurrences of k8s-jkns-gke-soak with <project-id>

Better yet could be replacing this with a script that runs all of these commands with the appropriate parameters

experiment/secret-sync-controller/main.go
```diff
@@ -0,0 +1,103 @@
...
+	if stat.IsDir() {
+		return nil, fmt.Errorf("config cannot be a dir - %s", config)
+	}
+
+	yamlFile, err := ioutil.ReadFile(config)
+	if err != nil {
+		return nil, fmt.Errorf("Error reading YAML file: %s\n", err)
+	}
+
+	specs := []SecretSyncSpec{}
```
spiffxp, line 96: I would recommend a top-level struct to hold this, you may find you want other fields at the same level, e.g.
```go
type struct SecretSyncConfig {
  Specs []SecretSyncSpec
}
```
```yaml
specs:
- source:
# ...
```

experiment/secret-sync-controller/SecretSyncs-config.yaml
```diff
@@ -0,0 +1,16 @@
+- source:
```
spiffxp, line 1: naming nit of the filename, prefer lowercase; in this case maybe just rename to `config.yaml`

experiment/secret-sync-controller/go.mod
```diff
@@ -0,0 +1,11 @@
+module b01901143.git/secret-sync
```
spiffxp, line 1: this is incorrect

experiment/secret-sync-controller/go.mod
```diff
@@ -0,0 +1,11 @@
...
+
+go 1.14
```
spiffxp, line 3: We use go1.13 in this repo, it would probably be better to follow that

experiment/secret-sync-controller/secret_sync_spec.go
b01901143: - Obtained BUILD.bazel ("experiment/BUILD.bazel", "experiment/secret-sync-controller/BUILD.bazel", "experiment/secret-sync-controller/client/BUILD.bazel")
- Renamed all snake case variables to camel case.
- The context argument is removed after switching to k8s.io/client-go v0.17.3

experiment/secret-sync-controller/client_util/client_util.go
b01901143: got rid of all snake case filename/package name/variables.
Is there any circumstance that we would use snake case?

experiment/secret-sync-controller/main.go
b01901143: Agreed. It also makes the structure more complete.

experiment/secret-sync-controller/go.mod
b01901143: changed to k8s.io/test-infra/experiment/secret-sync-controller

experiment/secret-sync-controller/go.mod
b01901143: switched to 
- go 1.13
- k8s.io/api v0.17.3
- k8s.io/apimachinery v0.17.3
- k8s.io/client-go v0.17.3

experiment/secret-sync-controller/SecretSyncs-config.yaml
b01901143: Filenames prefer lowercase with dashes?
Renamed.

experiment/secret-sync-controller/README.md
b01901143: Replaced.
Working on a script containing all these, as well as setting up workload identity.

experiment/secret-sync-controller/secret_sync_spec.go
b01901143: Add 'omitempty' tags to 
- Secret
- DenyList
- Kubernetes
- SecretManager
fields

experiment/secret-sync-controller/client/client.go
```diff
@@ -0,0 +1,48 @@
...
+	if err != nil {
+		panic(err.Error())
+	}
+	return clientset
+}
+
+func NewSecretManagerClient(ctx context.Context) *secretmanager.Client {
+	// Create the client.
+	jsonPath := "/usr/local/google/home/shanefu/gcloud_key.json"
+	client, err := secretmanager.NewClient(ctx, option.WithCredentialsFile(jsonPath))
```
spiffxp, line 43: The hardcode means I and others can't run or test this.  I would recommend using the default, and expecting the user/environment to provide the correct credentials.
```suggestion
	client, err := secretmanager.NewClient(ctx)
```

experiment/secret-sync-controller/client/client.go
```diff
@@ -0,0 +1,48 @@
...
+)
+
+func NewK8sClientset() *kubernetes.Clientset {
+	var kubeconfig *string
+	if home := os.Getenv("HOME"); home != "" {
+		kubeconfig = flag.String("kubeconfig", filepath.Join(home, ".kube", "config"), "(optional) absolute path to the kubeconfig file")
+	} else {
+		kubeconfig = flag.String("kubeconfig", "", "absolute path to the kubeconfig file")
+	}
+	flag.Parse()
```
spiffxp, line 24: If you're trying to set this up as a command line flag, you should do so in the `options` struct and follow the pattern you used for `config-path`, then turn this into `NewK8sClientset(kubconfig string)`

experiment/secret-sync-controller/client/client.go
```diff
@@ -0,0 +1,48 @@
...
+
+	// use the current context in kubeconfig
+	config, err := clientcmd.BuildConfigFromFlags("", *kubeconfig)
+	if err != nil {
+		panic(err.Error())
```
spiffxp, line 29: In general, don't panic. Instead, return the err and let the caller decide what to do. In this case that means:
- change to `func NewK8sClientset(...) (kubernetes.Interface, error)`
- change err lines like this to `return nil, err`
- change successful lines to `return clientset, nil`

https://golang.org/doc/effective_go.html#panic

experiment/secret-sync-controller/client/client.go
```diff
@@ -0,0 +1,48 @@
...
+
+	secretmanager "cloud.google.com/go/secretmanager/apiv1"
+	"google.golang.org/api/option"
+
+	"k8s.io/client-go/kubernetes"
+	_ "k8s.io/client-go/plugin/pkg/client/auth"
+	"k8s.io/client-go/tools/clientcmd"
+)
+
+func NewK8sClientset() *kubernetes.Clientset {
```
spiffxp, line 17: Consider using `kubernetes.Interface` instead

spiffxp: meta comment for the function as a whole: consider re-using https://github.com/kubernetes/test-infra/blob/master/prow/kube/cluster.go#L57 instead (you are welcome to do this as a followup PR instead)

experiment/secret-sync-controller/client/client.go
```diff
@@ -0,0 +1,48 @@
...
+	}
+	return clientset
+}
+
+func NewSecretManagerClient(ctx context.Context) *secretmanager.Client {
+	// Create the client.
+	jsonPath := "/usr/local/google/home/shanefu/gcloud_key.json"
+	client, err := secretmanager.NewClient(ctx, option.WithCredentialsFile(jsonPath))
+	if err != nil {
+		panic(err.Error())
```
spiffxp, line 45: again, don't panic, return this up to the caller

experiment/secret-sync-controller/secret-sync-config.go
```diff
@@ -0,0 +1,97 @@
...
+		return gsm.LatestSecretVersion(secretManagerCtx, secretManagerClient)
+	}
+}
+
+func (k8s *KubernetesSpec) LatestSecretVersion(k8sClientset *kubernetes.Clientset) (int, map[string][]byte) {
+	secret, err := k8sClientset.CoreV1().Secrets(k8s.Namespace).Get(k8s.Secret, metav1.GetOptions{})
+	if apierrors.IsNotFound(err) {
+		fmt.Println(err)
+	}
+	version, _ := strconv.Atoi(secret.ObjectMeta.ResourceVersion)
```
spiffxp, line 61: It looks like you're going through acrobatics to return version as an int both here and below.  I'm not sure this is a safe conversion (is ReourceVersion always guaranteed to be a string representation of an `int`?).  I would probably just keep them as strings instead.

experiment/secret-sync-controller/secret-sync-config.go
```diff
@@ -0,0 +1,97 @@
...
+	DenyList  []string `yaml:"denyList,omitempty"`
+}
+
+type SecretManagerSpec struct {
+	Project  string   `yaml:"project"`
+	Secret   string   `yaml:"secret,omitempty"`
+	DenyList []string `yaml:"denyList,omitempty"`
+}
+
+func (target Target) GetLatestSecretVersion(k8sClientset *kubernetes.Clientset, secretManagerCtx context.Context, secretManagerClient *secretmanager.Client) (int, map[string][]byte) {
```
spiffxp, line 48: As we discussed this morning, prefer instead to return a single result and an error.  Returning an untyped tuple feels like python, no go.  So this case, something like
```suggestion
type struct Secret {
  Version string // I'm using string instead of int, see comment in LatestSecretVersion
  Data map[string][]byte
}

func (target Target) GetLatestSecretVersion(k8sClientset *kubernetes.Clientset, secretManagerCtx context.Context, secretManagerClient *secretmanager.Client) (Secret, error) {
```

experiment/secret-sync-controller/secret-sync-config.go
```diff
@@ -0,0 +1,97 @@
...
+	// Call the API.
+	accResult, _ := client.AccessSecretVersion(ctx, accReq)
+	buf := make(map[interface{}]interface{})
+	yaml.Unmarshal(accResult.Payload.Data, &buf)
+
+	secret := make(map[string][]byte)
+
+	for key, val := range buf {
+		secret[fmt.Sprintf("%v", key)] = []byte(fmt.Sprintf("%v", val))
+	}
```
spiffxp, line 94: `map[interface{}]interface{}` is a red flag to me, can you help me understand what you're trying to do here?

spiffxp:
Looks like bazel's not happy, did you run `./hack/update-bazel.sh`?
```ERROR: /home/prow/go/src/github.com/kubernetes/test-infra/experiment/secret-sync-controller/BUILD.bazel:3:1: no such package '@org_golang_google_genproto//googleapis/cloud/secretmanager/v1': BUILD file not found in directory 'googleapis/cloud/secretmanager/v1' of external repository @org_golang_google_genproto. Add a BUILD file to a directory to mark it as a package. and referenced by '//experiment/secret-sync-controller:go_default_library'```

**Comment by b01901143** [comment]
> Looks like bazel's not happy, did you run `./hack/update-bazel.sh`?
> `ERROR: /home/prow/go/src/github.com/kubernetes/test-infra/experiment/secret-sync-controller/BUILD.bazel:3:1: no such package '@org_golang_google_genproto//googleapis/cloud/secretmanager/v1': BUILD file not found in directory 'googleapis/cloud/secretmanager/v1' of external repository @org_golang_google_genproto. Add a BUILD file to a directory to mark it as a package. and referenced by '//experiment/secret-sync-controller:go_default_library'`

Yes, but I ran it in the secret-sync-controller/ directory `../../hack/update-bazel.sh`
Not sure if it would cause this problem.

File: experiment/secret-sync-controller/client/client.go (hunk 9 above)
Comment (b01901143) [reply experiment/secret-sync-controller/client/client.go:43]: You're right. That was me testing if using multiple service-accounts is possible.
Modified.

By calling `client, err := secretmanager.NewClient(ctx)` if I have a KSA with workload identity binding with GSA, does it use that KSA instead of looking for a key in GOOGLE_APPLICATION_CREDENTIALS?

File: experiment/secret-sync-controller/client/client.go (hunk 10 above)
Comment (b01901143) [reply experiment/secret-sync-controller/client/client.go:24]: I was just trying to use the function `clientcmd.BuildConfigFromFlags()`
Then I realized that the input parameter doesn't have to be an output of flag.String, but just a string. Therefore now I modified this to 
`kubeconfig := filepath.Join(os.Getenv("HOME"), ".kube", "config")`

This should be later replaced with `rest.InClusterConfig()` using in-cluster-client-config.

File: experiment/secret-sync-controller/client/client.go (hunk 11 above)
Comment (b01901143) [reply experiment/secret-sync-controller/client/client.go:29]: Done

File: experiment/secret-sync-controller/BUILD.bazel [hunk 17]
```diff
@@ -0,0 +1,43 @@
+load("@io_bazel_rules_go//go:def.bzl", "go_binary", "go_library")
+
+go_library(
+    name = "go_default_library",
+    srcs = [
+        "main.go",
+        "secret_sync_spec.go",
```
spiffxp, line 7:
```suggestion
        "secret-sync-config.go",
```

experiment/secret-sync-controller/client/client.go
b01901143: It looks like https://github.com/kubernetes/test-infra/blob/19a078d68a69b0eaa948876f0e2b4f64fc9c78df/prow/kube/cluster.go#L57 tries to use in-cluster config first, and falls back to default rules if failed.
This would be optimal for me. Re-using this function.

experiment/secret-sync-controller/client/client.go
b01901143: Fixed.

experiment/secret-sync-controller/BUILD.bazel
b01901143: Oh.. sorry for that, I was being stupid.
Forgot to rerun it before I push

experiment/secret-sync-controller/secret-sync-config.go
b01901143: I was trying to parse `accResult.Payload.Data` (`[]uint8`) to a `map[string][]byte` so that the secret data for both gsm and k8s can be stored with uniform type.

experiment/secret-sync-controller/secret-sync-config.go
b01901143: You're right about this.
After looking more into resource versions:
https://kubernetes.io/docs/reference/using-api/api-concepts/#resource-versions
It turns out that they should be treated as strings.
"For example, clients must not assume resource versions are numeric, and may only compare two resource version for equality"

b01901143: According to this: https://kubernetes.io/docs/reference/using-api/api-concepts/#efficient-detection-of-changes
I think internally they are numeric, but they should be treated as opaque by clients.

experiment/secret-sync-controller/secret-sync-config.go
```diff
@@ -0,0 +1,161 @@
...
+}
+type TargetSpec struct {
+	// assert that one of the two should be `nil`
+	Kubernetes    *KubernetesSpec    `yaml:"kubernetes,omitempty"`
+	SecretManager *SecretManagerSpec `yaml:"secretManager,omitempty"`
+}
+type KubernetesSpec struct {
+	Namespace string   `yaml:"namespace"`
+	Secret    string   `yaml:"secret,omitempty"`
+	DenyList  []string `yaml:"denyList,omitempty"`
```
spiffxp, line 34: As we discussed this morning, I would drop the denyList stuff for now.  Instead, give KubernetesSpec an additional field of `Key string` or something to help the mapping between kubernetes secrets which expect key=value pairs, and secret manager secrets which don't.

experiment/secret-sync-controller/secret-sync-config.go
```diff
@@ -0,0 +1,161 @@
...
+		}
+
+		newVersion := secret.ObjectMeta.ResourceVersion
+		if newVersion != target.Kubernetes.Version {
+			updated = true
+		}
+
+		// update to latest version
+		target.Kubernetes.Version = newVersion
+		target.Kubernetes.Data = secret.Data
```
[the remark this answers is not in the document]
b01901143: The change can actually be observed, because target.Kubernetes is a pointer. I think what happens here is a new `target` is copied along with a `*KubernetesSecret` pointing to the same address. 
That said, you're right about this. It will be better to pass the pointer of `target` instead. But I guess we'll avoid storing states, and instead just return the `SecretData`

experiment/secret-sync-controller/secret-sync-config.go
```diff
@@ -0,0 +1,161 @@
...
+	DenyList []string `yaml:"denyList,omitempty"`
+}
+
+// structs for client interface
+type ClientInterface interface {
+}
+type Client struct { // actual client
+	K8sClientset        *kubernetes.Interface
+	SecretManagerClient *secretmanager.Client
+	Ctx                 context.Context
```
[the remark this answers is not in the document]
b01901143: According https://cloud.google.com/secret-manager/docs/creating-and-accessing-secrets
I think context.Background() is all we need.
But I'll use context.TODO() just in case we will want to handle them otherwise.

experiment/secret-sync-controller/secret-sync-config.go
```diff
@@ -0,0 +1,161 @@
...
+	Version   string
+	Data      map[string][]byte
+}
+type SecretManagerSecret struct {
+	Project string
+	Secret  string
+	Version string
+	Data    []uint8
+}
+
```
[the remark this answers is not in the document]
b01901143: Decided to discard storing states. Just re-fetch and compare the two secret values every time.
Deleting this struct.

experiment/secret-sync-controller/secret-sync-config.go
```diff
@@ -0,0 +1,161 @@
...
+		// update to latest version
+		target.SecretManager.Version = newVersion
+		target.SecretManager.Data = accResult.Payload.Data
+
+		return updated, nil
+	}
+}
+
+func (target Target) PrintSecret() {
```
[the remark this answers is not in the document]
b01901143: TargetSpec is also removed.
Instead I wrote a `String()` method for `SecretSyncConfig`, returning the marshaled string.

experiment/secret-sync-controller/main.go
```diff
@@ -0,0 +1,130 @@
...
+		log.Fatal(err)
+	}
+
+	for i, spec := range secretSyncConfig.Specs {
+		fmt.Printf("Secret pair [%d]:\n{\n", i)
+		err := Sync(clientInterface, spec)
+		if err != nil {
+			log.Fatal(err)
+		}
+	}
```
spiffxp, line 68: I would suggest pulling this out of main and into the controller struct I mentioned elsewhere.  Consider adding a flag to gate one-shot mode vs. running continuously

experiment/secret-sync-controller/client/client.go
```diff
@@ -0,0 +1,35 @@
...
+
+import (
+	"context"
+	"os"
+	"path/filepath"
+
+	secretmanager "cloud.google.com/go/secretmanager/apiv1"
+
+	"k8s.io/client-go/kubernetes"
+	_ "k8s.io/client-go/plugin/pkg/client/auth"
```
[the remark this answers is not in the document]
b01901143: No.
That already handled in "k8s.io/test-infra/prow/kube".

experiment/secret-sync-controller/secret-sync-config.go
```diff
@@ -0,0 +1,115 @@
...
+
+func (config SecretSyncConfig) String() string {
+	d, _ := yaml.Marshal(config)
+	return string(d)
+}
+
+type SecretData struct {
+	Data []byte
+	// Potentially more information
+}
```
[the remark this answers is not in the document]
b01901143: I defined this just in case we would want to include more data, e.g. meta and version.
But since versions would not be needed and by patching the metadata is inherited for K8s secrets,
I think I'll just use  `[]byte`

experiment/secret-sync-controller/main.go
```diff
@@ -0,0 +1,130 @@
...
+
+	// get destination secret
+	destData, destErr := clientInterface.GetKubernetesSecret(spec.Destination)
+	if destErr != nil {
+		return destErr
+	}
+
+	// update destination secret
+	if reflect.DeepEqual(*srcData, *destData) {
```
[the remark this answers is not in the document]
b01901143: Yes, `bytes.Equal` works.

experiment/secret-sync-controller/secret-sync-config.go
```diff
@@ -0,0 +1,115 @@
...
+	UpsertKubernetesSecret(KubernetesSpec, *SecretData) (*SecretData, error)
+}
+type Client struct { // actual client
+	K8sClientset        *kubernetes.Interface
+	SecretManagerClient *secretmanager.Client
+}
+
+// GetKubernetesSecret gets the K8s secret data specified in KubernetesSpec.
+// It gets a single secret values with the given spec.Key.
+func (cl Client) GetKubernetesSecret(spec KubernetesSpec) (*SecretData, error) {
```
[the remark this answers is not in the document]
b01901143: Agreed.
Should avoid copying the clients.

experiment/secret-sync-controller/main_test.go
```diff
@@ -0,0 +1,295 @@
...
+				},
+				SecretManagerSecret: map[string]map[string][]byte{
+					"projA": map[string][]byte{
+						"secretA": []byte("valueOfSecretA"),
+					},
+					"projB": map[string][]byte{
+						"secretB": []byte("valueOfSecretB"),
+					},
+				},
+			}
```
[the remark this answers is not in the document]
b01901143: @spiffxp That's what I initially did.
But then I figured that the data stored in the maps will be modified for each run (with `UpsertKubernetesSecret()`).
Or should I write a `clientMock.reset()` function to reset the data within the clientmock, and just call this `reset()` at the start of each run?

experiment/secret-sync-controller/secret-sync-config.go
```diff
@@ -0,0 +1,115 @@
...
+	secret, err := (*cl.K8sClientset).CoreV1().Secrets(spec.Namespace).Get(spec.Secret, metav1.GetOptions{})
```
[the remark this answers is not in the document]
b01901143: Yes, this bothered me for a while.
I wanted to get rid of `(*cl.K8Clientset)`, but in the same time wish to preserve a consistency between K8sClientset and SecretManagerClient being both pointers.
I think I'll change both to be values, and just make share that `Client` struct will not be copied.

experiment/secret-sync-controller/secret-sync-config.go
```diff
@@ -0,0 +1,115 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	// Get the secert value according to the key.
+	// Allow the case where the key-value pair for spec.Key is not yet create,
+	// a key-value pair will be inserted with the UpsertKubernetesSecret() function.
+	value, ok := secret.Data[spec.Key]
+	if !ok {
+		return new(SecretData), nil
```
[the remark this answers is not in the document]
b01901143: @spiffxp I had this concern too.
The answer here is no.
Part of me want to return an empty value and an `err` for key-not-found, and treat this error as acceptable in the `Sync` function. Basically using the `err` as a flag, but I'm not sure if it's the right way to utilize an `err`.

experiment/secret-sync-controller/secret-sync-config.go
```diff
@@ -0,0 +1,115 @@
...
+type SecretData struct {
+	Data []byte
+	// Potentially more information
+}
+
+// structs for client interface
+type ClientInterface interface {
+	GetKubernetesSecret(KubernetesSpec) (*SecretData, error)
+	GetSecretManagerSecret(SecretManagerSpec) (*SecretData, error)
+	UpsertKubernetesSecret(KubernetesSpec, *SecretData) (*SecretData, error)
```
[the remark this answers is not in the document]
b01901143: I was trying to avoid copying of potentially large SecretData.
But I think I'll just use []byte.
It would be better to put all these in the `client` package.

experiment/secret-sync-controller/main.go
```diff
@@ -0,0 +1,130 @@
...
+		fmt.Printf("Secret pair [%d]:\n{\n", i)
+		err := Sync(clientInterface, spec)
+		if err != nil {
+			log.Fatal(err)
+		}
+	}
+
+}
+
+func Sync(clientInterface ClientInterface, spec SecretSyncSpec) error {
```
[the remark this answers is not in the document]
b01901143: Agreed.

experiment/secret-sync-controller/main_test.go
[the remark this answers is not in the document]
spiffxp: Perhaps move the map literals up top, and keep the construction of the mock client down here?

experiment/secret-sync-controller/secret-sync-config.go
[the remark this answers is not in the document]
spiffxp: so there is prior art for using an err to say something doesn't exist, and checking for that error type, https://stackoverflow.com/questions/12518876/how-to-check-if-a-file-exists-in-go is a pretty common pattern

Let's think of it in terms of "what would I need to respond to as a caller of this method"
- if the secret doesn't exist, I'll call upsert
- if the secret key doesn't exist, I'll call upsert
- if the secret exists, I'll call upsert
- if I can't access the secret for some other reason, I can't recover

So in those terms, I guess this function should inspect the error it gets from fetching the secret before returning it

experiment/secret-sync-controller/main_test.go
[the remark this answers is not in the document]
b01901143: I wrote a constructor `NewClientMock()` above the `testSync` function,
and within each run I construct a new client with `client := NewClientMock()`.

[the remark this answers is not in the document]
b01901143: I created a function variable `NewClientMock` within `testSync`.

experiment/secret-sync-controller/main.go
b01901143: This is now implemented in the Controller struct.
```
type Controller struct {
	Client       client.Interface
	Config       *SecretSyncConfig
	RunOnce      bool
	ResyncPeriod time.Duration
}
```
`RunOnce` stands for one-shot mode.

experiment/secret-sync-controller/main_test.go
```diff
@@ -0,0 +1,295 @@
...
+		{
+			input: SecretSyncSpec{
+				Source: SecretManagerSpec{
+					Project: "projA",
+					Secret:  "secretA",
+				},
+				Destination: KubernetesSpec{
+					Namespace: "ns2",
+					Secret:    "secret2",
+					Key:       "key3",
```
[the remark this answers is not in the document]
b01901143: I'm feeling these tests should just be replaced with e2e testings on actual clients.

experiment/secret-sync-controller/main_test.go
```diff
@@ -0,0 +1,295 @@
...
+}
+
+func TestSync(t *testing.T) {
+
+	var tests = []struct {
+		input     SecretSyncSpec
+		want      []byte
+		expectErr bool
+	}{
+		{
```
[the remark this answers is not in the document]
b01901143: Added

experiment/secret-sync-controller/main.go
```diff
@@ -0,0 +1,130 @@
...
+		return nil
+	}
+	// update destination secret value
+	// inserts a key-value pair if spec.Destination does not exist yet
+	patchedData, patchErr := clientInterface.UpsertKubernetesSecret(spec.Destination, srcData)
+	if patchErr != nil {
+		return patchErr
+	}
+
+	fmt.Printf("\tSource secret: \n\t%s\n", srcData.Data)
```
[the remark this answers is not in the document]
b01901143: Turned to using `klog`.
I'm considering whether I need a in-struct `logger` for `SecretSyncController`.

experiment/secret-sync-controller/main_test.go
```diff
@@ -1,295 +1,435 @@
...
+					{
+						Name: "ns-c",
 					},
 				},
+			},
+			expectErr: true,
+		},
+	}
+	for _, tt := range tests {
+		testname := fmt.Sprintf("%s", tt.config)
```
spiffxp, line 363: consider adding a `name` field to your testcase table struct, then instead of commenting each test case you could name it

experiment/secret-sync-controller/main_test.go
```diff
@@ -1,295 +1,435 @@
...
+  - name: ns-a
+    secret:
+    - name: secret-a
+      key:
+      - name: key-a
+        data: old-token
+  - name: ns-b
+    secret:
+    - name: secret-a
+  - name: ns-c
```
spiffxp, line 34: Note you could do something a bit more concise if you assume duplicate names aren't allowed to exist
```yaml
secretManager:
  k8s-jkns-gke-soak:
    gsm-password: gsm-password-v1
    gsm-token: gsm-token-v1
kubernetes:
  ns-a:
    secret-a:
      key-a: old-token
  ns-b:
    secret-b: {}
```

experiment/secret-sync-controller/main_test.go
```diff
@@ -1,295 +1,435 @@
...
 			},
-			want:      []byte("valueOfSecretA"),
 			expectErr: false,
 		},
 		{
-			input: SecretSyncSpec{
-				Source: SecretManagerSpec{
-					Project: "projB",
-					Secret:  "secretB",
+			// syncing from <existing gsm secret> to <existing k8s secret> with <non-existing key>.
```
spiffxp, line 146: So I would be curious to see what happens if we were to sync to `secret-a`, would we see both `key-a` and `missed` in its data?  At present this could be seen as overwriting `{}` with `missed: gsm-token-v1`

experiment/secret-sync-controller/main_test.go
```diff
@@ -1,295 +1,435 @@
...
-			secret, err := mockClient.GetKubernetesSecret(tt.input)
-			if tt.error {
-				if err == nil {
-					t.Errorf("Expected error but got none.")
-				}
-			} else {
-				if err != nil {
-					t.Errorf("Expected none error but got \"Error: %s\".", err)
-				}
+			err = controller.Client.SetupTesting(setup)
```
spiffxp, line 386: As discussed above, this should not be a method on Client, but instead perhaps a method defined in this file

experiment/secret-sync-controller/main.go
```diff
@@ -0,0 +1,219 @@
...
+	if err != nil {
+		klog.Errorf("New Secret Manager client failed: %s", err)
+	}
+	clientInterface := &client.Client{
+		K8sClientset:        *k8sClientset,
+		SecretManagerClient: *secretManagerClient,
+	}
+
+	// prepare config
+	secretSyncConfig, err := LoadConfig(o.configPath)
```
spiffxp, line 68: Once you've loaded the config, you're going to want to validate it to make sure there are no invalid specs, eg:
- a destination appears twice
- there are no empty field names

You might want to create a skeleton/bare bones `SecretSyncConfig.Validate()` function to fill out later

spiffxp:
100+ comments means to me it's time to consider where we want to draw the line and call this done before iterating in another PR, let's catch up tomorrow

experiment/secret-sync-controller/client/client.go
```diff
@@ -33,3 +41,339 @@ func NewSecretManagerClient(ctx context.Context) (*secretmanager.Client, error)
...
+type Interface interface {
+	ValidateKubernetesNamespace(string) error
+	GetSecretManagerSecret(string, string) ([]byte, error)
+	GetKubernetesSecret(string, string, string) ([]byte, error)
+	UpsertKubernetesSecret(string, string, string, []byte) error
+	// For testing
+	CleanupSecretManagerSecrets(string) error
+	CleanupKubernetesSecrets(string) error
+	SetupTesting(Setup) error
+	CleanupTesting(Setup, bool) error
```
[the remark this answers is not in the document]
b01901143: I created a `tests` package, and defined a struct `TestClient` that implements the original `client.Interface` as well as the `tests.TestInterface` for fixture utilities.

experiment/secret-sync-controller/client/client.go
```diff
@@ -33,3 +41,339 @@ func NewSecretManagerClient(ctx context.Context) (*secretmanager.Client, error)
...
+func (cl *Client) UpsertKubernetesSecret(namespace string, secretID string, key string, src []byte) error {
+	// check if the namespace exists
+	_, err := cl.K8sClientset.CoreV1().Namespaces().Get(namespace, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+
+	// encode with base64 encoding
+	encodedSrc := base64.StdEncoding.EncodeToString(src)
+	patch := []byte("{\"data\":{\"" + key + "\": \"" + encodedSrc + "\"}}")
```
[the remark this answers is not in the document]
b01901143: It's weird that the result wouldn't be correct if I implement it their way.
I have to change the type of field `data` to `map[string]string`:  
```
json.Marshal(map[string]interface{}{
		"data": map[string]string{key: encodedSrc},
	})
```

experiment/secret-sync-controller/client/client.go
```diff
@@ -33,3 +41,339 @@ func NewSecretManagerClient(ctx context.Context) (*secretmanager.Client, error)
...
+	accResult, err := cl.SecretManagerClient.AccessSecretVersion(ctx, accReq)
+	if err != nil {
+		return nil, err
+	}
+
+	return accResult.Payload.Data, nil
+}
+
+// UpsertKubernetesSecret updates or inserts a key-value pair in the K8s secret with the given spec.Key.
+func (cl *Client) UpsertKubernetesSecret(namespace string, secretID string, key string, src []byte) error {
```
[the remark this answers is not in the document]
b01901143: Got it.

experiment/secret-sync-controller/main.go
```diff
@@ -1,110 +1,199 @@
...
+
+	for {
+		select {
+		case <-stopChan:
+			klog.Info("Stop signal received. Quitting...")
+			return nil
+		case <-runChan:
+			c.SyncAll()
+			if c.RunOnce {
+				c.Client.CleanupKubernetesSecrets("ns1")
```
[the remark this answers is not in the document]
b01901143: That was... not supposed to be there.
Sorry about that. Removed.

experiment/secret-sync-controller/main.go
```diff
@@ -1,110 +1,199 @@
...
+				return nil
+			}
+		}
+	}
+}
+
+// ValidateAccess validates the required access for each sync pair.
+// If finds an invlaid access, return an error and delete that sync pair.
```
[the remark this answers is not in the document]
b01901143: That makes a lot of sense.
Now this will only be tackled in `SyncAll()`. Any pair of secrets that fails to be synced/accessed will be skipped and an error message for that secret pair will pop up.

experiment/secret-sync-controller/main.go
```diff
@@ -1,110 +1,199 @@
...
+		c.Config.Specs = validSpecs
+	}
+
+	return nil
 }
 
-func Sync(clientInterface ClientInterface, spec SecretSyncSpec) error {
-	// k8s requires & yields secrets with type map[string][]byte,
-	// while SecretManager uses []uint8 (e.g. account: "foo"\n secret: "bar")
+func (c *SecretSyncController) SyncAll() error {
```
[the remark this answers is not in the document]
b01901143: Yes. It should pop error message for each secret pair that encounters failure.

experiment/secret-sync-controller/main.go
```diff
@@ -1,110 +1,199 @@
...
+			klog.Errorf("Secret sync failed for %s: %s", spec, err)
+		}
+		if updated {
+			klog.Infof("Secret %s synced from %s", spec.Destination, spec.Source)
+		}
+	}
+	return nil
+}
 
+func (c *SecretSyncController) Sync(spec SecretSyncSpec) (bool, error) {
```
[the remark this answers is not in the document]
b01901143: Got it.

experiment/secret-sync-controller/main_test.go
```diff
@@ -1,295 +1,435 @@
...
+			if err != nil {
+				t.Error(err)
 			}
-		})
-	}
-}
 
-func TestGetSecretManagerSecret(t *testing.T) {
+			var stopChan <-chan struct{}
+			controller.Start(stopChan)
```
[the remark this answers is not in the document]
b01901143: Yes.
I thought I would be testing also the continuous mode in the future.
But in that case I should write that in a different test.

experiment/secret-sync-controller/main.go
```diff
@@ -1,110 +1,199 @@
...
 	}
 
-	// get destination secret
-	destData, destErr := clientInterface.GetKubernetesSecret(spec.Destination)
-	if destErr != nil {
-		return destErr
+	//checks if the K8s namespace exists
+	err = c.Client.ValidateKubernetesNamespace(spec.Destination.Namespace)
+	if err != nil {
+		return false, err
```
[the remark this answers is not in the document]
b01901143: True.
Removed.

experiment/secret-sync-controller/client/client.go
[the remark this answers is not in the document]
b01901143: A real client only implements client.Interface.
A E2eTestClient implements tests.ClientInterface (cleanup functions + client.Interface), and actually communicates with the APIs.
A MockClient also implements tests.ClientInterface (cleanup functions + client.Interface), but only mocks the responses.

experiment/secret-sync-controller/main_test.go
b01901143: Done.
Replaced.

experiment/secret-sync-controller/client/client.go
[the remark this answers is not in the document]
spiffxp: Sounds good, it looks like you haven't pushed this commit up yet?

experiment/secret-sync-controller/main_test.go
b01901143: In this concise form, I think it's necessary to use `json.Unmarshal` instead.
`yaml.Unmarshal` always yields type of `map[interface{}]interface{}`, which I assume is a red flag.

With `json.Unmarshal`, we can just use `map[string]interface{}`.

experiment/secret-sync-controller/client/client.go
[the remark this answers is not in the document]
b01901143: No I haven't.
I'm still trying to finish up the unit tests.

experiment/secret-sync-controller/main_test.go
b01901143: Yes we would see both keys under `secret-a`.
I'll add that test case.

experiment/secret-sync-controller/main.go
b01901143: Done

experiment/secret-sync-controller/secret-sync-config.go
```diff
@@ -0,0 +1,72 @@
+package main
```
spiffxp, line 1: I would expect this to live in something other than main, I think that might get you out of this error

```
$ go test ./experiment/secret-sync-controller/main_test.go
# command-line-arguments [command-line-arguments.test]
experiment/secret-sync-controller/main_test.go:84:13: undefined: SecretSyncSpec
```

experiment/secret-sync-controller/client/client.go
```diff
@@ -0,0 +1,226 @@
...
+}
+
+// structs for client interface
+type Interface interface {
+	ValidateKubernetesNamespace(namespace string) error
+	ValidateKubernetesSecret(namespace, id string) error
+	CreateKubernetesNamespace(namespace string) error
+	GetKubernetesSecretValue(namespace, id, key string) ([]byte, error)
+	UpsertKubernetesSecret(namespace, id, key string, data []byte) error
+	CreateKubernetesSecret(namespace, id string) error
```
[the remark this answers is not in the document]
spiffxp: I guess I'm not sure I see the value in supporting the concept of a secret that has no data, which I think is what this is here for?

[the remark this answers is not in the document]
b01901143: The reason I have this function is entirely for setting up the test cases.
I didn't remove it from the `client.Interface` interface because I thought it wasn't that destructive/dangerous.

experiment/secret-sync-controller/client/client.go
```diff
@@ -33,3 +41,339 @@ func NewSecretManagerClient(ctx context.Context) (*secretmanager.Client, error)
...
+		return nil, err
+	}
+
+	return accResult.Payload.Data, nil
+}
+
+// UpsertKubernetesSecret updates or inserts a key-value pair in the K8s secret with the given spec.Key.
+func (cl *Client) UpsertKubernetesSecret(namespace string, secretID string, key string, src []byte) error {
+	// check if the namespace exists
+	_, err := cl.K8sClientset.CoreV1().Namespaces().Get(namespace, metav1.GetOptions{})
```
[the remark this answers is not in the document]
b01901143: Sorry I missed this one.
Fixed.

experiment/secret-sync-controller/main_test.go
```diff
@@ -0,0 +1,947 @@
...
+				t.Errorf("Unexpected error: %s", err)
+			}
+
+			// validate result
+			for namespace, nsItem := range tc.want {
+				// check if the namespace exists
+				err := controller.Client.ValidateKubernetesNamespace(namespace)
+				if err != nil {
+					t.Error(err)
+				}
```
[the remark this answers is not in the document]
b01901143: The initial purpose of this is to make sure that syncing doesn't delete namespaces, and also that the fixture is working as I expected.
But now I think this is rather irrelevant to the testing of `Sync()` and should be removed?

spiffxp:
I have a local copy of this rebased on https://github.com/kubernetes/test-infra/pull/18000 with an extra commit to point at my local copy of testgrid, now running into:
```
$ bazel build //experiment/secret-sync-controller
ERROR: /Users/spiffxp/w/kubernetes/test-infra/experiment/secret-sync-controller/client/BUILD.bazel:3:1: no such package '@org_golang_google_genproto//googleapis/cloud/secretmanager/v1': BUILD file not found in directory 'googleapis/cloud/secretmanager/v1' of external repository @org_golang_google_genproto. Add a BUILD file to a directory to mark it as a package. and referenced by '//experiment/secret-sync-controller/client:go_default_library'
ERROR: Analysis of target '//experiment/secret-sync-controller:secret-sync-controller' failed; build aborted: no such package '@org_golang_google_genproto//googleapis/cloud/secretmanager/v1': BUILD file not found in directory 'googleapis/cloud/secretmanager/v1' of external repository @org_golang_google_genproto. Add a BUILD file to a directory to mark it as a package.
```

experiment/secret-sync-controller/main_test.go
```diff
@@ -0,0 +1,947 @@
...
+
+				if nsItem == nil {
+					continue
+				}
+				for secret, secretItem := range nsItem.(map[string]interface{}) {
+					err = controller.Client.ValidateKubernetesSecret(namespace, secret)
+					if err != nil {
+						t.Error(err)
+					}
```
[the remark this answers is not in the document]
b01901143: I think this test is needed. It checks if the `missed` secret is upserted.

b01901143:
> ```
> $ go test -v -count=1 ./experiment/secret-sync-controller/... --e2e-client=true
> ...
> ok  	k8s.io/test-infra/experiment/secret-sync-controller	274.234s
> ```
> 
> OK so that's way slower than I was expecting. Creating/deleting namespaces for each testcase is probably overkill. Can you do that less often?

I think I can only cleanup the secrets after each test case, and cleanup the namespaces just once after each test_function?
The only thing that creates new namespaces with each run of test case is the `fixDestination` recovery, but I think I'll just remove that because it indeed makes the test cases too huge. So, there's no need to delete namespaces between test cases, at least for now.

experiment/secret-sync-controller/main_test.go
```diff
@@ -0,0 +1,947 @@
...
+		spec      SecretSyncSpec
+		want      tests.Fixture
+		update    bool
+		expectErr bool
+	}{
+		{
+			name: "Sync from <existing gsm secret> to <existing k8s secret> with <existing key> containing the <same secret values>. Should not update.",
+			spec: SecretSyncSpec{
+				Source: SecretManagerSpec{
+					Project: "k8s-jkns-gke-soak",
```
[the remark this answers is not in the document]
b01901143: Done, but did it in a sort of brute force way in `NewFixture(config []byte, project string)`

experiment/secret-sync-controller/main_test.go
```diff
@@ -0,0 +1,947 @@
...
+		t.Fatalf("Fail to parse fixture: %s", err)
+	}
+
+	var testcases = []struct {
+		name           string
+		config         SecretSyncConfig
+		resBefore      tests.Fixture
+		fixSource      tests.Fixture
+		fixDestination tests.Fixture
+		resAfter       tests.Fixture
```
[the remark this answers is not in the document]
b01901143: True. Removed.

experiment/secret-sync-controller/main_test.go
```diff
@@ -0,0 +1,947 @@
...
+
+			err = fixture.Teardown(testClient)
+			if err != nil {
+				t.Error(err)
+			}
+		})
+	}
+}
+
+func TestValidate(t *testing.T) {
```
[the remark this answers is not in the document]
b01901143: Moved this to `secret-sync-config_test.go`.

experiment/secret-sync-controller/secret-sync-config.go
b01901143: Resolved.
Put `SecretSyncConfig` into the `config` package, and  `SecretSyncController` into the `controller` package.

spiffxp:
/close
This ended up landing over in https://github.com/kubernetes-sigs/k8s-gsm-tools

PR #17949, opened 2020-06-16
Use ubuntu as ssh user for kops canaries

johngmyers:
/cc @rifelpet

### DIFF
--- a/config/jobs/kubernetes/sig-testing/kubetest-canaries.yaml
+++ b/config/jobs/kubernetes/sig-testing/kubetest-canaries.yaml
@@ -209,7 +209,8 @@ periodics:
       args:
       - --cluster=e2e-kops-aws-canary.test-cncf-aws.k8s.io
       - --deployment=kops
-      - --env=KUBE_SSH_USER=admin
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes/sig-testing/kubetest-canaries.yaml
```diff
@@ -244,6 +245,8 @@ periodics:
       args:
       - --cluster=e2e-kops-gce-canary.k8s.local
       - --deployment=kops
+      - --kops-ssh-user=ubuntu
+      - --env=KUBE_SSH_USER=ubuntu
```
rifelpet, line 249: I'm not sure this should be set on the GCE jobs (see https://github.com/kubernetes/test-infra/pull/17950#issuecomment-644726859 )

/lgtm cancel

johngmyers:
@wojtek-t Anything else you need to approve this PR?

BenTheElder:
/approve
But please move this to being owned by the kops team.
We're freezing kubetest and not actively working on these.

PR #17950, opened 2020-06-16
Use ubuntu as ssh user for kops gce tests

johngmyers:
/cc @rifelpet


rifelpet:
It looks like GCE still uses COS:

https://github.com/kubernetes/kops/blob/f3e69e85c9c010e8f344f2a5fea57409b9ae4d30/channels/alpha#L50-L55

and the failing prow jobs [still have artifacts](https://gcsweb.k8s.io/gcs/kubernetes-jenkins/logs/e2e-kops-gce-latest/1272851236973973506/) for all of the nodes, so kubetest is successfully SSHing into them. Are you sure this is needed?

johngmyers:
Okay, so that's not it.

PR #17953, opened 2020-06-16
Changes in k8sio-dns-update prow job

bartsmykla:
- Changed the name of postsubmit job which is updating the dnses from
  `k8sio-dns-update` to `post-k8sio-dns-update` which is more consistent
  with other jobs
- Updated the version of octodns image in `k8sio-dns-update` (now
  `post-k8sio-dns-update`) job for the one with `make` inside
- Updated the `k8sio-dns-update` (now `post-k8sio-dns-update`) job to
  use `make` instead of directly calling the bash scripts, which is more
  consistent with the commands used by dns-admins

/assign @spiffxp 

ref. [`k/k8s.io#961`](https://github.com/kubernetes/k8s.io/issues/961)

Signed-off-by: Bart Smykla <bsmykla@vmware.com>

### DIFF
--- a/config/jobs/kubernetes/wg-k8s-infra/trusted/wg-k8s-infra-trusted.yaml
+++ b/config/jobs/kubernetes/wg-k8s-infra/trusted/wg-k8s-infra-trusted.yaml
@@ -44,7 +44,7 @@ postsubmits:
         args:
         - -c
         - "cd groups && make run -- --confirm"
-  - name: k8sio-dns-update
+  - name: post-k8sio-dns-update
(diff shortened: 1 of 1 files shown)


bartsmykla:
Ok, now I see in logs of failing test:
```
jobs_test.go:348: pre-k8sio-dns-update: presubmits may not run in cluster: k8s-infra-prow-build-trusted
```
the problem is this job needs to use `k8s-infra-dns-updater` to work

bartsmykla:
@spiffxp I have removed the presubmit job as there was no way to make it work, so if you'll have few minutes, could you look at this updates?

PR #17960, opened 2020-06-16
GitHub GetRef: Better handle the case of getting multiple result

alvaroaleman:
The GetRef call may return multiple results because it does prefix-based
matching. Today, this means that we error on the json unmarshaling with
a `json: cannot unmarshal array into Go value of type struct { Object
map[string]string \"json:\\\"object\\\"\" }"` which is hard to handle
for callers.

This PR updates this to return an error in case of multiple refs and
allow calles to match on that error.

### DIFF
--- a/prow/github/client.go
+++ b/prow/github/client.go
@@ -2598,19 +2598,77 @@ func (c *client) ReopenPR(org, repo string, number int) error {
 // GetRef returns the SHA of the given ref, such as "heads/master".
 //
 // See https://developer.github.com/v3/git/refs/#get-a-reference
+// The gitbub api does prefix matching and might return multiple results,
+// in which case we will return a GetRefTooManyResultsError
 func (c *client) GetRef(org, repo, ref string) (string, error) {
 	durationLogger := c.log("GetRef", org, repo, ref)
 	defer durationLogger()
 
-	var res struct {
-		Object map[string]string `json:"object"`
-	}
+	res := GetRefResponse{}
 	_, err := c.request(&request{
 		method:    http.MethodGet,
 		path:      fmt.Sprintf("/repos/%s/%s/git/refs/%s", org, repo, ref),
 		exitCodes: []int{200},
 	}, &res)
-	return res.Object["sha"], err
+	if err != nil {
+		return "", nil
+	}
+
+	if n := len(res); n > 1 {
+		return "", GetRefTooManyResultsError{org: org, repo: repo, ref: ref, resultsRefs: res.RefNames()}
+	}
+	return res[0].Object.SHA, nil
+}
+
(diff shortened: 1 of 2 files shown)


prow/github/client.go
```diff
@@ -2598,19 +2598,78 @@ func (c *client) ReopenPR(org, repo string, number int) error {
...
+	}
+	return res[0].Object.SHA, nil
+}
+
+type GetRefTooManyResultsError struct {
+	org, repo, ref string
+	resultsRefs    []string
+}
+
+func (GetRefTooManyResultsError) Is(err error) bool {
```
cjwagner, line 2628: This signature seems wrong to me. Does this need to be a method of `GetRefTooManyResultsError` rather than a receiverless function? This would have to be called like so: `GetRefTooManyResultsError{}.Is(err)`.

prow/github/client.go
```diff
@@ -2598,19 +2598,78 @@ func (c *client) ReopenPR(org, repo string, number int) error {
...
+// We need a custom unmarshaler because the GetRefResult may either be a
+// single GetRefResponse or multiple
+func (grr *GetRefResponse) UnmarshalJSON(data []byte) error {
+	result := &GetRefResult{}
+	if err := json.Unmarshal(data, result); err == nil {
+		*(grr) = GetRefResponse{*result}
+		return nil
+	}
+	type responseWithoutUnmarshaler GetRefResponse
+	var response responseWithoutUnmarshaler
```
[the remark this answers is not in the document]
alvaroaleman: Good point, done

prow/github/client.go
alvaroaleman: This is not meant to be called directly but to suport `errors.Is`: https://golang.org/pkg/errors/#Is

alvaroaleman:
> I also expected some changes to handle the new error differently than other types of GetRef errors. Did you mean to include those changes here or are you planning a follow up?

Where? Right now I made this change because we need to detect this error downstream.

prow/github/client.go
alvaroaleman: Usage looks like this in the end:

```
if errors.Is(err, github.GetRefTooManyResultsError{}) { do something}
```

cjwagner: Neat, I didn't know about that.

cjwagner:
> > I also expected some changes to handle the new error differently than other types of GetRef errors. Did you mean to include those changes here or are you planning a follow up?
> 
> Where? Right now I made this change because we need to detect this error downstream.

Ah, I thought this change was needed to fix an issue with a Prow component. I didn't know there were dependencies on Prow's GH client outside k/t-i.

PR #17963, opened 2020-06-16
ci-kubernetes-e2e-gce-device-plugin-gpu-beta needs docker as default runtime

hasheddan:
A recent change in kube-up switched us to using `containerd` as default,
so we need to fix up CI jobs that depend on docker.

Same as https://github.com/kubernetes/test-infra/pull/17905

Fixes failures on https://testgrid.k8s.io/sig-release-1.18-blocking#gce-device-plugin-gpu-1.18


spiffxp:
ref: https://github.com/kubernetes/kubernetes/pull/91684

PR #17964, opened 2020-06-16
Implement kubetest2 GCE Up() for cloud-provider-gcp

michaelmdresser:
Implements the Up() method for the kubetest2 GCE deployer. This version of Up() only supports the kube-up.sh in k/cloud-provider-gcp for POC/MVP. Also sets path for kubeconfig to be in the kubetest2 artifacts directory. Chose to start with empty environment rather than inheriting from the OS to guarantee functionality on minimal information.

Relies on:
* https://github.com/kubernetes/cloud-provider-gcp/pull/153
* https://github.com/kubernetes/cloud-provider-gcp/pull/154

Tested against fresh GCP projects. If doing the same, make sure to ignore errors from the gsutil ACL command, as it will fail and does not matter.
https://github.com/kubernetes/cloud-provider-gcp/blob/bf905db434d3cfbdc345b19429a6c3a1180016b9/cluster/gce/util.sh#L244

Change:
```
gsutil -m acl ch -g all:R "${gs_url}" "${gs_url}.sha512" >/dev/null 2>&1
```
to
```
gsutil -m acl ch -g all:R "${gs_url}" "${gs_url}.sha512" >/dev/null 2>&1 || :
```

### DIFF
--- a/kubetest2/kubetest2-gce/deployer/deployer.go
+++ b/kubetest2/kubetest2-gce/deployer/deployer.go
@@ -18,6 +18,10 @@ limitations under the License.
 package deployer
 
 import (
+	"fmt"
+	"os"
+	"path/filepath"
+
 	"k8s.io/klog"
 	"k8s.io/test-infra/kubetest2/pkg/types"
 
@@ -32,13 +36,16 @@ type deployer struct {
 	// generic parts
 	commonOptions types.Options
 
-	RepoRoot string `desc:"The path to the root of the local kubernetes/cloud-provider/gcp repo. Necessary to call certain scripts. Defaults to the current directory. If operating in legacy mode, this should be set to the local kubernetes/kubernetes repo."`
+	kubeconfigPath string
+	RepoRoot       string `desc:"The path to the root of the local kubernetes/cloud-provider-gcp repo. Necessary to call certain scripts. Defaults to the current directory. If operating in legacy mode, this should be set to the local kubernetes/kubernetes repo."`
+	GCPProject     string `desc:"GCP Project to create VMs in. Must be set."`
 }
 
 // New implements deployer.New for gce
 func New(opts types.Options) (types.Deployer, *pflag.FlagSet) {
 	d := &deployer{
-		commonOptions: opts,
+		commonOptions:  opts,
+		kubeconfigPath: filepath.Join(opts.ArtifactsDir(), "kubetest2-kubeconfig"),
 	}
 
 	flagSet, err := gpflag.Parse(d)
@@ -60,14 +67,18 @@ func (d *deployer) Provider() string {
 	return Name
 }
 
-func (d *deployer) Up() error { return nil }
-
 func (d *deployer) IsUp() (up bool, err error) { return false, nil }
 
 func (d *deployer) DumpClusterLogs() error { return nil }
 
-func (d *deployer) TestSetup() error { return nil }
-
-func (d *deployer) Kubeconfig() (string, error) { return "", nil }
+func (d *deployer) Kubeconfig() (string, error) {
+	_, err := os.Stat(d.kubeconfigPath)
+	if err == nil {
+		return d.kubeconfigPath, nil
+	} else if os.IsNotExist(err) {
+		return "", fmt.Errorf("kubeconfig does not exist at: %s", d.kubeconfigPath)
+	}
+	return "", fmt.Errorf("unknown error when checking for kubeconfig at %s: %s", d.kubeconfigPath, err)
+}
 
 func (d *deployer) Down() error { return nil }
--- a/kubetest2/kubetest2-gce/deployer/up.go
+++ b/kubetest2/kubetest2-gce/deployer/up.go
@@ -1,0 +1,111 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package deployer
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"k8s.io/klog"
+	"k8s.io/test-infra/kubetest2/pkg/exec"
+)
+
+func (d *deployer) Up() error {
+	klog.Info("GCE deployer starting Up()")
+
+	if err := d.verifyFlags(); err != nil {
+		return fmt.Errorf("up failed to verify flags: %s", err)
+	}
+
+	if err := enableComputeAPI(d.GCPProject); err != nil {
+		return fmt.Errorf("up couldn't enable compute API: %s", err)
+	}
+
+	env := d.buildEnv()
+	script := filepath.Join(d.RepoRoot, "cluster", "kube-up.sh")
+	klog.Infof("About to run script at: %s", script)
+
+	cmd := exec.Command(script)
+	cmd.SetEnv(env...)
+	exec.InheritOutput(cmd)
+	err := cmd.Run()
+	if err != nil {
+		return fmt.Errorf("error encountered during %s: %s", script, err)
+	}
+
+	return nil
+}
+
+func (d *deployer) buildEnv() []string {
+	// The base env currently does not inherit the current os env (except for PATH)
+	// because (for now) it doesn't have to. In future, this may have to change when
+	// support is added for k/k's kube-up.sh and kube-down.sh which support a wide
+	// variety of environment variables. Before doing so, it is worth investigating
+	// inheriting the os env vs. adding flags to this deployer on a case-by-case
+	// basis to support individual environment configurations.
+	var env []string
+
+	// path is necessary for scripts to find gsutil, gcloud, etc
+	env = append(env, fmt.Sprintf("PATH=%s", os.Getenv("PATH")))
+
+	// kube-up.sh, kube-down.sh etc. use PROJECT as a parameter for all gcloud commands
+	env = append(env, fmt.Sprintf("PROJECT=%s", d.GCPProject))
+
+	// kubeconfig is set to tell kube-up.sh where to generate the kubeconfig
+	// we don't want this to be the default because this kubeconfig "belongs" to
+	// the run of kubetest2 and so should be placed in the artifacts directory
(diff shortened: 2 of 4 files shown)


kubetest2/kubetest2-gce/deployer/up.go
```diff
@@ -0,0 +1,96 @@
...
+
+	if err := d.enableComputeAPI(); err != nil {
+		return fmt.Errorf("up couldn't enable compute API: %s", err)
+	}
+
+	env := d.buildEnv()
+	script := filepath.Join(d.RepoRoot, "cluster", "kube-up.sh")
+
+	cmd := exec.Command(script)
+	cmd.SetEnv(env...)
```
amwat, line 40: It's awesome if we can create a hermetic env for this, but are we sure we won't need any envs passed from top down.

kubetest2/kubetest2-gce/deployer/up.go
```diff
@@ -0,0 +1,96 @@
...
+
+	return nil
+}
+
+func (d *deployer) buildEnv() []string {
+	env := []string{}
+
+	// path is necessary for scripts to find gsutil, gcloud, etc
+	env = append(env, fmt.Sprintf("PATH=%s", os.Getenv("PATH")))
+	env = append(env, fmt.Sprintf("PROJECT=%s", d.GCPProject))
```
amwat, line 55: let's add clarifying comments to all these to make our future life easier

kubetest2/kubetest2-gce/deployer/up.go
```diff
@@ -0,0 +1,96 @@
...
+	env = append(env, fmt.Sprintf("KUBECONFIG=%s", d.kubeconfigPath))
```
amwat, line 56: adding to the previous comment e.g. IIUC KUBECONFIG here is set so that kube-up will populate it for us right? and not us providing it like what kubetest2 core will do for the tester eventually.

kubetest2/kubetest2-gce/deployer/up.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+
+	if d.GCPProject == "" {
+		return fmt.Errorf("GCP Project must be set.")
+	}
+
+	return nil
+}
+
+func (d *deployer) enableComputeAPI() error {
```
amwat, line 73: nit: this method doesn't seem like it should have a deployer receiver, instead should have the project name as an argument.

kubetest2/kubetest2-gce/deployer/up.go
```diff
@@ -0,0 +1,96 @@
...
+
+	// path is necessary for scripts to find gsutil, gcloud, etc
+	env = append(env, fmt.Sprintf("PATH=%s", os.Getenv("PATH")))
+	env = append(env, fmt.Sprintf("PROJECT=%s", d.GCPProject))
+	env = append(env, fmt.Sprintf("KUBECONFIG=%s", d.kubeconfigPath))
+
+	return env
+}
+
+func (d *deployer) verifyFlags() error {
```
amwat, line 61: verifyUpFlags ? to be consistent with verifyBuildFlags?
eventually should be a method on the UpOptions whenever we move the flags to relevant structs.

kubetest2/kubetest2-gce/deployer/up.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+func (d *deployer) enableComputeAPI() error {
+	// In freshly created GCP projects, the compute API is
+	// not enabled. We need it. Enabling it after it has
+	// already been enabled is a relatively fast no-op,
+	// so this can be called without consequence.
+
+	env := os.Environ()
+	env = append(env, fmt.Sprintf("CLOUDSDK_CORE_PROJECT=%s", d.GCPProject))
```
amwat, line 80: use --project instead of setting the env?

kubetest2/kubetest2-gce/deployer/up.go
```diff
@@ -0,0 +1,96 @@
...
+
+	cmd := exec.Command(
```
amwat, line 82: optonal nit: using the golang API would be awesome here, but i know we rely on gcloud to be existing for kube-up anyway.

kubetest2/kubetest2-gce/deployer/up.go
michaelmdresser: I'm tempted to leave it as bare-bones as possible for now, understanding that it may run into issues later (when targeting k/k rather than just cloud-provider-gcp). In tests it works for now. Totally up for either direction.

kubetest2/kubetest2-gce/deployer/up.go
michaelmdresser: Clarifying this in a comment. Yes, this makes kube-up create the kubeconfig at a specified path (in the artifacts directory) so that kubetest2 core can eventually ask the deployer "where is the kubeconfig" and it can always respond with the same path (in the artifacts directory).

kubetest2/kubetest2-gce/deployer/up.go
michaelmdresser: I've gone back and forth on this. Up(), Down(), and presumably DumpClusterLogs() need the same set of flags. I'm tempted to keep this generic (not up-specific) because it is rather generic for most functionality. Let me know what you think.

kubetest2/kubetest2-gce/deployer/up.go
michaelmdresser: Agreed. However, I couldn't find services as a supported component of the golang API when I looked.

kubetest2/kubetest2-gce/deployer/up.go
```diff
@@ -0,0 +1,96 @@
...
+	if err := d.verifyFlags(); err != nil {
+		return fmt.Errorf("up could not verify flags: %s", err)
+	}
+
+	if err := d.enableComputeAPI(); err != nil {
+		return fmt.Errorf("up couldn't enable compute API: %s", err)
+	}
+
+	env := d.buildEnv()
+	script := filepath.Join(d.RepoRoot, "cluster", "kube-up.sh")
```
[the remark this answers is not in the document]
michaelmdresser: Done.

kubetest2/kubetest2-gce/deployer/up.go
michaelmdresser: Done. Let me know what you think.

kubetest2/kubetest2-gce/deployer/up.go
michaelmdresser: Definitely.

kubetest2/kubetest2-gce/deployer/up.go
michaelmdresser: Done. Good call.

kubetest2/kubetest2-gce/deployer/deployer.go
```diff
@@ -60,14 +65,12 @@ func (d *deployer) Provider() string {
...
-
 func (d *deployer) IsUp() (up bool, err error) { return false, nil }
 
 func (d *deployer) DumpClusterLogs() error { return nil }
 
-func (d *deployer) TestSetup() error { return nil }
-
-func (d *deployer) Kubeconfig() (string, error) { return "", nil }
+func (d *deployer) Kubeconfig() (string, error) {
+	return d.kubeconfigPath, nil
```
rancerbeta, line 73: maybe return an error if this kubeconfig was not populated

kubetest2/kubetest2-gce/deployer/up.go
```diff
@@ -0,0 +1,112 @@
...
+
+	"k8s.io/klog"
+	"k8s.io/test-infra/kubetest2/pkg/exec"
+)
+
+func (d *deployer) Up() error {
+	klog.Info("GCE deployer starting Up()")
+
+	if err := d.verifyFlags(); err != nil {
+		return fmt.Errorf("up could not verify flags: %s", err)
```
rancerbeta, line 32: nit: could not verify flags -> flag verification failed

kubetest2/kubetest2-gce/deployer/up.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+
+	if err := enableComputeAPI(d.GCPProject); err != nil {
+		return fmt.Errorf("up couldn't enable compute API: %s", err)
+	}
+
+	env := d.buildEnv()
+	script := filepath.Join(d.RepoRoot, "cluster", "kube-up.sh")
+	klog.Infof("About to run script at: %s", script)
+	klog.Infof("Running under environment: %+v", env)
```
rancerbeta, line 42: you might not want to do this. Be careful of dumping passwords and secrets. Optionally, you may want to use a verbosity flag + warning this could dump credentials

kubetest2/kubetest2-gce/deployer/up.go
```diff
@@ -0,0 +1,112 @@
...
+}
+
+func (d *deployer) buildEnv() []string {
+	// The base env currently does not inherit the current os env (except for PATH)
+	// because (for now) it doesn't have to. In future, this may have to change when
+	// support is added for k/k's kube-up.sh and kube-down.sh which support a wide
+	// variety of environment variables. Before doing so, it is worth investigating
+	// inheriting the os env vs. adding flags to this deployer on a case-by-case
+	// basis to support individual environment configurations.
+	env := []string{}
```
rancerbeta, line 62: its up to you, but I typically see 
`var env []string` instead of initializing the slice

kubetest2/kubetest2-gce/deployer/deployer.go
michaelmdresser: I like that. Done.

kubetest2/kubetest2-gce/deployer/up.go
michaelmdresser: Done, slightly different string.

kubetest2/kubetest2-gce/deployer/up.go
michaelmdresser: Removed it. While the env being provided does not currently inherit from the OS, it could in future and it wouldn't be clear that this could dump secrets.

kubetest2/kubetest2-gce/deployer/up.go
michaelmdresser: Sure.

rancerbeta:
lgtm, thanks mike

kubetest2/kubetest2-gce/deployer/up.go
michaelmdresser: It seems I was looking in the wrong place, or at least in the "recommended new library" that doesn't have this. Tried using it but it looks like I need to resolve project name to project number first (https://cloud.google.com/service-usage/docs/enable-disable#curl, https://cloud.google.com/resource-manager/docs/creating-managing-projects#api_1). Given the necessary existence of `gcloud` and with this being the only place we interact with the API (for now), I think its fine as is. When the deployer has to start creating firewall rules, etc. It may be worth revisiting.

PR #17968, opened 2020-06-17
Kops - Revert #17917

hakman:
Based on chat with @BenTheElder seems that the change is not quite valid.
As it is now, it blocks Kops merges.

/cc @mikesplain @rifelpet @BenTheElder

### DIFF
--- a/config/prow/config.yaml
+++ b/config/prow/config.yaml
@@ -190,15 +190,11 @@ branch-protection:
         kops:
           required_status_checks:
             contexts:
+            - build (ubuntu-18.04, 1.13)
+            - build (ubuntu-18.04, 1.14)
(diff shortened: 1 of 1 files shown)


BenTheElder:
i'm not sure if it is or isn't
/approve

hakman:
Thanks @BenTheElder. At least we can see if this was the reason or not.

rifelpet:
Thanks for setting this up
/lgtm

mikesplain:
Thanks everyone, I’ll take another look. 

This worked fine on a different prow instance. Since this merged a little bit ago and before branch protection ran again, a number of PRs have merged which means the issue is the tide config, not the branch protection config. 

Thanks for the quick revert, I’ll do more testing and we can figure out next steps.

PR #17972, opened 2020-06-17
Enable OOM tracking in GCE 100 nodes performance tests

tosi3k:
Roll out the mechanism implemented in https://github.com/kubernetes/perf-tests/pull/1309 to 100 nodes performance tests.

/sig scalability
/cc jkaniuk
/cc mm4tt
/cc wojtek-t


jkaniuk:
Please roll out periodics first, then if green - presubmits.

tosi3k:
> Please roll out periodics first, then if green - presubmits.

Done.

PR #17974, opened 2020-06-17
GHProxy: log token usage

alvaroaleman:
Right now we only log free tokens but we do not log when we use tokens. Having logs about when tokens were used might be super useful for debugging issues there.

### DIFF
--- a/prow/github/client.go
+++ b/prow/github/client.go
@@ -398,13 +398,21 @@ func (t *throttler) Do(req *http.Request) (*http.Response, error) {
 		if ghcache.CacheModeIsFree(cacheMode) {
 			// This request was fulfilled by ghcache without using an API token.
 			// Refund the throttling token we preemptively consumed.
-			log := logrus.WithFields(logrus.Fields{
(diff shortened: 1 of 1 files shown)


stevekuznetsov:
In what case does this not get captured in the Prometheus metrics? Are we going to generate gigabytes of logs with this?

alvaroaleman:
> In what case does this not get captured in the Prometheus metrics? Are we going to generate gigabytes of logs with this?

We already log every time we do not use a token which I would expect to be the more frequent case.

Metrics do not cover everything and are not granular enough for debugging, because it might for example be important to know the precise path rather than `some GetRef calls use a token`.

cjwagner:
> Are we going to generate gigabytes of logs with this?

Maybe? FWIW we are already logging the other case which is far more common on Prow instances of significant size. k8s-prow has ~90% cache efficiency so we should see this new log 9x less than the existing one. 
We could log this with `Trace` instead of `Debug` or add an option to omit those logs if we find that they are taking up too much space.

PR #17978, opened 2020-06-17
use standard image-config for conformance testing

MHBauer:
The prow job overrides the test definitions in benchmark-config
resulting in the same tests running on the same images over and over
again, way too many times.

This seems to have been a mistake when transferring into test-infra.

There are also duplicate images defined in benchmark-config which I'll make another PR to fixup.

### DIFF
--- a/config/jobs/kubernetes/sig-node/node-kubelet.yaml
+++ b/config/jobs/kubernetes/sig-node/node-kubelet.yaml
@@ -112,7 +112,7 @@ periodics:
       - --deployment=node
       - --gcp-project-type=node-e2e-project
       - --gcp-zone=us-west1-b
-      - --node-args=--image-config-file=/workspace/test-infra/jobs/e2e_node/benchmark-config.yaml --test-suite=conformance
+      - --node-args=--image-config-file=/workspace/test-infra/jobs/e2e_node/image-config.yaml --test-suite=conformance
       - --node-test-args=--feature-gates=DynamicKubeletConfig=true --kubelet-flags="--cgroups-per-qos=true --cgroup-root=/"
       - --node-tests=true
       - --provider=gce


config/jobs/kubernetes/sig-node/node-kubelet.yaml
```diff
@@ -112,7 +112,7 @@ periodics:
       - --deployment=node
       - --gcp-project-type=node-e2e-project
       - --gcp-zone=us-west1-b
-      - --node-args=--image-config-file=/workspace/test-infra/jobs/e2e_node/benchmark-config.yaml --test-suite=conformance
+      - --node-args=--image-config-file=/workspace/test-infra/jobs/e2e_node/image-config.yaml --test-suite=conformance
```
bsdnet, line 115: I saw in benchmark-config.yaml, we tried to run different tests against the same image. Is it designed differently to run a test-suite on the same image? Wonder whether it is intentional if it do "up, run test1, down", "up, run test2, down"  or "up, run test1, run test2, run test3, down"

MHBauer: I don't understand. 

benchmark config lists single test per image.

bsdnet: What I mean for benchmark-config.yaml , each test will be running:
create cluster, run test1, teardown cluster;
create cluster, runt test2, teardown cluster;

For image-config.yaml, it will run in the following:
create cluster, run test1,  run test2, teardown cluster;

If it is above case,  benchmark-config.yaml  is preferred to avoid interfering between tests.

MHBauer: Yes, exactly.

This is not running benchmarks.

derekwaynecarr:
nice catch, appreciate your continued assistance in improving testing.

/approve
/lgtm

config/jobs/kubernetes/sig-node/node-kubelet.yaml
bsdnet: Cool, this is a good fix.

MHBauer:
https://testgrid.k8s.io/sig-node-kubelet#node-kubelet-conformance tab is already blue, takes 10 runs for stale tests runs to become hidden.

PR #17981, opened 2020-06-18
Skip NodeFeature:RuntimeHandler for containerd-e2e-gci-1.2

bsdnet:

Skip the RuntimeHandler since it is not supported in containerd 1.2
https://testgrid.k8s.io/sig-node-containerd#containerd-e2e-gci-1.2

### DIFF
--- a/config/jobs/kubernetes/sig-node/containerd.yaml
+++ b/config/jobs/kubernetes/sig-node/containerd.yaml
@@ -149,7 +149,7 @@ periodics:
       - --gcp-zone=us-west1-b
       - --ginkgo-parallel=30
       - --provider=gce
-      - --test_args=--ginkgo.skip=\[Slow\]|\[Serial\]|\[Disruptive\]|\[Flaky\]|\[Feature:.+\] --minStartupPods=8
+      - --test_args=--ginkgo.skip=\[Slow\]|\[Serial\]|\[Disruptive\]|\[Flaky\]|\[Feature:.+\]|\[NodeFeature:RuntimeHandler\] --minStartupPods=8
       - --timeout=50m
       image: gcr.io/k8s-testimages/kubekins-e2e:v20200616-45495bd-1.15
   annotations:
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes/sig-node/containerd.yaml
```diff
@@ -149,7 +149,7 @@ periodics:
       - --gcp-zone=us-west1-b
       - --ginkgo-parallel=30
       - --provider=gce
-      - --test_args=--ginkgo.skip=\[Slow\]|\[Serial\]|\[Disruptive\]|\[Flaky\]|\[Feature:.+\] --minStartupPods=8
+      - --test_args=--ginkgo.skip=\[Slow\]|\[Serial\]|\[Disruptive\]|\[Flaky\]|\[Feature:.+\]|\[NodeFeature:RuntimeHandler\] --minStartupPods=8
```
MHBauer, line 152: can we add a comment here or in the commit message about *why* we're skipping the runtimehandler.

bsdnet: The difference is between test-infra/jobs/e2e_node/containerd/containerd-release-1.2/env
and test-infra/jobs/e2e_node/containerd/containerd-release-1.3/env

Containerd-1.2 does not support RuntimeHandler per:
CONTAINERD_EXTRA_RUNTIME_HANDLER: 'test-handler'
CONTAINERD_EXTRA_RUNTIME_OPTIONS: |
  BinaryName = "/home/containerd/usr/local/sbin/runc"

I am thinking where is the good place to add this comment.

bsdnet: Also, by searching, I saw a lot of test suites has |\[NodeFeature:RuntimeHandler\]  without explaining. The best place explaining maybe in the file  test-infra/jobs/e2e_node/containerd/containerd-release-1.2/env file?

bsdnet:
> I would like to read a document about this not being supported and some comment about why it's not in the test so we don't come back six months later and try to turn it on.

I do not think this is the right place to do that. Containerd is specified, and this should be documented in containerd doc. or it was documented already, but we are just not aware.

Maybe @Random-Liu  @dims  can help find the right place.

For this PR, I think fixing the tests should be our first priority. Documenting should be a separate
PR since NodeFeature:RuntimeHandler is skipped in a lot of test suite. We need a central source
for better document, IMO.

bsdnet:
https://testgrid.k8s.io/sig-node-containerd#containerd-e2e-gci-1.2 became green now!

PR #17982, opened 2020-06-18
Fix dumping logs in regional and multizonal GKE clusters

jkaniuk:
Fix dumping logs in regional and multizonal GKE clusters

/cc @mborsz 
/cc @wojtek-t 
/cc @mm4tt

### DIFF
--- a/kubetest/gke.go
+++ b/kubetest/gke.go
@@ -402,8 +402,6 @@ func (g *gkeDeployer) IsUp() error {
 //
 // TODO(zmerlynn): This whole path is really gross, but this seemed
 // the least gross hack to get this done.
-//
-// TODO(shyamjvs): Make this work with multizonal and regional clusters.
 func (g *gkeDeployer) DumpClusterLogs(localPath, gcsPath string) error {
 	// gkeLogDumpTemplate is a template of a shell script where
 	// - %[1]s is the project
@@ -412,10 +410,7 @@ func (g *gkeDeployer) DumpClusterLogs(localPath, gcsPath string) error {
 	// - %[4]s is the log-dump.sh command line
 	const gkeLogDumpTemplate = `
 function log_dump_custom_get_instances() {
-  if [[ $1 == "master" ]]; then
-    return 0
-  fi
-
+  if [[ $1 == "master" ]]; then return 0; fi
   gcloud compute instances list '--project=%[1]s' '--filter=%[4]s' '--format=get(name)'
 }
 export -f log_dump_custom_get_instances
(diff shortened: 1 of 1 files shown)


kubetest/gke.go
```diff
@@ -462,12 +458,20 @@ export KUBE_NODE_OS_DISTRIBUTION='%[3]s'
...
+	var lastError error
+	for zone, filters := range perZoneFilters {
+		err := control.FinishRunning(exec.Command("bash", "-c", fmt.Sprintf(gkeLogDumpTemplate,
+			g.project,
+			zone,
+			os.Getenv("NODE_OS_DISTRIBUTION"),
+			strings.Join(filters, " OR "),
+			dumpCmd)))
+		if err != nil {
+			log.Printf("error while dumping logs in %s zone: %v", zone, err)
```
wojtek-t, line 470: Let's not print and return the same error

kubetest/gke.go
```diff
@@ -462,12 +458,20 @@ export KUBE_NODE_OS_DISTRIBUTION='%[3]s'
 		log.Printf("error while ensuring firewall rule: %v", err)
 	}
 
-	return control.FinishRunning(exec.Command("bash", "-c", fmt.Sprintf(gkeLogDumpTemplate,
-		g.project,
-		g.zone,
-		os.Getenv("NODE_OS_DISTRIBUTION"),
-		strings.Join(filters, " OR "),
-		dumpCmd)))
+	var lastError error
```
wojtek-t, line 461: Instead of last, can we contatenate all?

kubetest/gke.go
jkaniuk: I wanted to print it, as this error message is actually discarded in the calling function:
https://github.com/kubernetes/test-infra/blob/master/kubetest/e2e.go#L122

jkaniuk: Nevermind. That call was only for cases where test failed during up. In other places error is handled.

kubetest/gke.go

kubetest/gke.go
```diff
@@ -462,12 +458,22 @@ export KUBE_NODE_OS_DISTRIBUTION='%[3]s'
...
+	var errorMessages []string
+	for zone, filters := range perZoneFilters {
+		err := control.FinishRunning(exec.Command("bash", "-c", fmt.Sprintf(gkeLogDumpTemplate,
+			g.project,
+			zone,
+			os.Getenv("NODE_OS_DISTRIBUTION"),
+			strings.Join(filters, " OR "),
+			dumpCmd)))
+		if err != nil {
+			errorMessages = append(errorMessages, fmt.Sprintf("%v", err))
```
wojtek-t, line 470: nit:
errorMessages = append(errorMessages, err.Error())

PR #17984, opened 2020-06-18
Kops - Add presubmits for network plugins

hakman:
/cc @rifelpet

### DIFF
--- a/config/jobs/kubernetes/kops/kops-presubmits-network-plugins.yaml
+++ b/config/jobs/kubernetes/kops/kops-presubmits-network-plugins.yaml
@@ -1,0 +1,314 @@
+presubmits:
+  kubernetes/kops:
+
+  - name: pull-kops-e2e-cni-amazonvpc
+    branches:
+    - master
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes/kops/kops-presubmits-network-plugins.yaml
```diff
@@ -0,0 +1,314 @@
+presubmits:
+  kubernetes/kops:
+
+  - name: pull-kops-e2e-cni-amazonvpc
+    branches:
+    - master
+    run_if_changed: '^upup\/models\/cloudup\/resources\/addons\/networking\.amazon-vpc-routed-eni\/'
+    skip_report: false
```
rifelpet, line 8: can we make these `optional: true` to start? that way we dont bombard all the open PRs with test failures if theres an issue with their setup.

hakman: it will affect only those with the path and there are none at the moment

rifelpet: ah ok, that sounds good

PR #17986, opened 2020-06-18
Support for proxy node in GKE tests

jkaniuk:
Support for proxy node (bastion) in GKE tests.
Needed for correctness tests, as some of them require SSH access to nodes.

/cc @mm4tt 
/cc @wojtek-t

### DIFF
--- a/kubetest/gke.go
+++ b/kubetest/gke.go
@@ -487,15 +487,70 @@ func (g *gkeDeployer) TestSetup() error {
 	if err := g.ensureNat(); err != nil {
 		return err
 	}
+	if err := g.setupBastion(); err != nil {
+		return err
+	}
 	if err := g.setupEnv(); err != nil {
 		return err
 	}
-	if g.sshProxyInstanceName != "" {
-		if err := setKubeShhBastionEnv(g.project, g.zone, g.sshProxyInstanceName); err != nil {
-			return err
-		}
-	}
 	g.setup = true
+	return nil
+}
+
+func (g *gkeDeployer) setupBastion() error {
+	if g.sshProxyInstanceName == "" {
+		return nil
+	}
+	var filtersToTry []string
+	// Use exact name first, VM does not have to belong to the cluster
+	exactFilter := "name=" + g.sshProxyInstanceName
+	filtersToTry = append(filtersToTry, exactFilter)
+	// As a fallback - use proxy instance name as a regex but check only cluster nodes
+	var igFilters []string
+	for _, ig := range g.instanceGroups {
+		igFilters = append(igFilters, fmt.Sprintf("(metadata.created-by ~ %s)", ig.path))
+	}
+	fuzzyFilter := fmt.Sprintf("(name ~ %s) AND (%s)",
+		g.sshProxyInstanceName,
+		strings.Join(igFilters, " OR "))
+	filtersToTry = append(filtersToTry, fuzzyFilter)
+
+	var bastion, zone string
+	for _, filter := range filtersToTry {
+		log.Printf("Checking for proxy instance with filter: %q", filter)
+		output, err := exec.Command("gcloud", "compute", "instances", "list",
+			"--filter="+filter,
+			"--format=value(name,zone)",
+			"--limit=1",
+			"--project="+g.project).Output()
+		if err != nil {
+			return fmt.Errorf("listing instances failed: %s", util.ExecError(err))
+		}
+		if len(output) == 0 {
+			continue
+		}
+		// Proxy instance found
+		fields := strings.Split(strings.TrimSpace(string(output)), "\t")
+		if len(fields) != 2 {
+			return fmt.Errorf("error parsing instances list output %q", output)
+		}
+		bastion, zone = fields[0], fields[1]
+		break
+	}
+	if bastion == "" {
+		return fmt.Errorf("proxy instance %q not found", g.sshProxyInstanceName)
+	}
+	log.Printf("Found proxy instance %q", bastion)
+
+	log.Printf("Adding NAT access config if not present")
+	control.NoOutput(exec.Command("gcloud", "compute", "instances", "add-access-config", bastion,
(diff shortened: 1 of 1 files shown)


jkaniuk:
Will do manual tests
/hold

kubetest/gke.go
```diff
@@ -487,15 +489,127 @@ func (g *gkeDeployer) TestSetup() error {
...
+	if err := g.setupBastion(); err != nil {
+		return err
+	}
 	if err := g.setupEnv(); err != nil {
 		return err
 	}
-	if g.sshProxyInstanceName != "" {
-		if err := setKubeShhBastionEnv(g.project, g.zone, g.sshProxyInstanceName); err != nil {
-			return err
+	g.setup = true
```
mm4tt, line 498: Why are we setting this now and weren't setting it before?

kubetest/gke.go
```diff
@@ -487,15 +489,127 @@ func (g *gkeDeployer) TestSetup() error {
...
+	return nil
+}
+
+func (g *gkeDeployer) setupBastion() error {
+	if g.sshProxyInstanceName == "" {
+		return nil
+	}
+	var filtersToTry []string
+	// Use exact name first, VM do not have to belong to the cluster
```
mm4tt, line 507: NIT: does not

kubetest/gke.go
```diff
@@ -102,6 +102,8 @@ type gkeDeployer struct {
 	setup          bool
 	kubecfg        string
 	instanceGroups []*ig
+	bastionIPName  string
+	zoneRegion     string
```
mm4tt, line 106: What is zoneRegion?

kubetest/gke.go
```diff
@@ -102,6 +102,8 @@ type gkeDeployer struct {
 	setup          bool
 	kubecfg        string
 	instanceGroups []*ig
+	bastionIPName  string
```
mm4tt, line 105: What is bastionIPName? What is IPName?

kubetest/gke.go
```diff
@@ -487,15 +489,127 @@ func (g *gkeDeployer) TestSetup() error {
...
+
+func (g *gkeDeployer) cleanupBastion() error {
+	if g.bastionIPName == "" {
+		return nil
+	}
+	region, err := g.getRegion()
+	if err != nil {
+		return fmt.Errorf("error finding region for proxy instance address: %v", err)
+	}
+	if err := control.FinishRunning(exec.Command("gcloud", "compute", "addresses", "delete", g.bastionIPName,
```
mm4tt, line 607: If the bastion VM had public IP then we won't have the address and the deletion will fail, right?

kubetest/gke.go
```diff
@@ -487,15 +489,127 @@ func (g *gkeDeployer) TestSetup() error {
...
+	err := setKubeShhBastionEnv(g.project, zone, bastion)
+	if err == nil {
+		// Success, finishing
+		return nil
+	}
+	if !strings.Contains(err.Error(), "doesn't have an external IP address") {
+		return fmt.Errorf("setting KUBE_SSH_BASTION variable failed: %s", util.ExecError(err))
+	}
+
+	log.Printf("Trying to reserve external IP address for proxy instance")
```
mm4tt, line 556: Can we extract the code from here to a seperate method? E.g. createPublicIpForBastion or something like that

kubetest/gke.go
jkaniuk: just GUI diff error, it was there before as well

kubetest/gke.go

kubetest/gke.go
jkaniuk: changed to regionCache - used in getRegion()

kubetest/gke.go
jkaniuk: I'm sorry, it was old version of the code (with static IPs). Somehow my next push landed in the wrong place.

kubetest/gke.go
jkaniuk: I'm sorry, it was old version of the code (with static IPs). Somehow my next push landed in the wrong place.
But address deletion was guarded by new address name, not by bastion name.

kubetest/gke.go
jkaniuk: sorry, outdated

kubetest/gke.go
```diff
@@ -487,15 +488,70 @@ func (g *gkeDeployer) TestSetup() error {
...
+	log.Printf("Adding NAT access config if not present")
+	control.NoOutput(exec.Command("gcloud", "compute", "instances", "add-access-config", bastion,
+		"--zone="+zone,
+		"--project="+g.project))
+
+	err := setKubeShhBastionEnv(g.project, zone, bastion)
+	if err != nil {
+		return fmt.Errorf("setting KUBE_SSH_BASTION variable failed: %s", util.ExecError(err))
 	}
-	g.setup = true
```
jkaniuk, line 498: @mm4tt : Here it is

jkaniuk:
After first few tests looks good
/unhold

kubetest/gke.go
```diff
@@ -692,18 +748,22 @@ func (g *gkeDeployer) ensureNat() error {
 	return nil
 }
 
-func (g *gkeDeployer) getRegion(region, zone string) (string, error) {
-	if region != "" {
-		return region, nil
+func (g *gkeDeployer) getRegion() (string, error) {
```
wojtek-t, line 751: Why do we need those changes?

Those doesn't seem to be related to the bastion changes. And if so, I would prefer them to be moved out of this PR to make is as small as possible.

jkaniuk: Right, after ditching static address handling, getRegion is not used anymore.
Cleaned up.

kubetest/gke.go
jkaniuk: removed after refactoring

wojtek-t:
/lgtm
/approve

Thanks!

PR #17988, opened 2020-06-18
Kind skips

BenTheElder:
/cc @aojea @dims 

interestingly I see CAPI is already using 6000m CPU all over the place for basically everything ...
we're going to have to take a harder look at what we run when a default CPU request / limit is set, we only have 159 build nodes (7.91 core each, ~301m currently reserved for kube-proxy / workload identity / ...).

### DIFF
--- a/config/jobs/kubernetes/sig-testing/kubernetes-kind-presubmits.yaml
+++ b/config/jobs/kubernetes/sig-testing/kubernetes-kind-presubmits.yaml
@@ -76,7 +76,7 @@ presubmits:
           value: "."
         # TODO(bentheelder): reduce the skip list further
         - name: SKIP
-          value: \[Slow\]|\[Disruptive\]|\[Flaky\]|\[Feature:.+\]|PodSecurityPolicy|LoadBalancer|load.balancer|Simple.pod.should.support.exec.through.an.HTTP.proxy|subPath.should.support.existing
+          value: \[Slow\]|\[Disruptive\]|\[Flaky\]|\[Feature:.+\]|PodSecurityPolicy|LoadBalancer|load.balancer|Simple.pod.should.support.exec.through.an.HTTP.proxy|subPath.should.support.existing|inline.execution.and.attach
         - name: PARALLEL
           value: "true"
         - name: BUILD_TYPE
--- a/config/jobs/kubernetes/sig-testing/kubernetes-kind-presubmits.yaml
+++ b/config/jobs/kubernetes/sig-testing/kubernetes-kind-presubmits.yaml
@@ -38,12 +38,8 @@ presubmits:
           privileged: true
         resources:
           requests:
-            # TODO(BenTheElder): adjust these everywhere
-            # these are both a bit below peak usage during build
-            # this is mostly for building kubernetes
             memory: "9000Mi"
-            # during the tests more like 3-20m is used
-            cpu: 2000m
+            cpu: 7500m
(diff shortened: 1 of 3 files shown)


BenTheElder:
tried to canary this but have bazel failures on the canary 🙄 
https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/46662/pull-kubernetes-e2e-kind-canary/1273626401764806656

CPU request is not too high anyhow. it's just under the machine available of 7.609
will work better when we add a default request > 100m

config/jobs/kubernetes/sig-testing/kubernetes-kind-presubmits.yaml
```diff
@@ -28,7 +28,7 @@ presubmits:
           value: "."
         # TODO(bentheelder): reduce the skip list further
         - name: SKIP
-          value: \[Slow\]|\[Disruptive\]|\[Flaky\]|\[Feature:.+\]|PodSecurityPolicy|LoadBalancer|load.balancer|Simple.pod.should.support.exec.through.an.HTTP.proxy|subPath.should.support.existing
+          value: \[Slow\]|\[Disruptive\]|\[Flaky\]|\[Feature:.+\]|PodSecurityPolicy|LoadBalancer|load.balancer|Simple.pod.should.support.exec.through.an.HTTP.proxy|subPath.should.support.existing|Conntrack|udp|UDP|NFS|nfs
```
aojea, line 31: this is the one Dan Winship reported that fails if there are icmp drops in the host

```suggestion
          value: \[Slow\]|\[Disruptive\]|\[Flaky\]|\[Feature:.+\]|PodSecurityPolicy|LoadBalancer|load.balancer|Simple.pod.should.support.exec.through.an.HTTP.proxy|subPath.should.support.existing|Conntrack|udp|UDP|NFS|nfs||should be rejected when no endpoints exist
```

aojea:
anyway 
/lgtm
this is urgent

config/jobs/kubernetes/sig-testing/kubernetes-kind-presubmits.yaml
BenTheElder: you can't use spaces here safely IIRC, update this and use `.`

config/jobs/kubernetes/sig-testing/kubernetes-kind-presubmits.yaml
```diff
@@ -80,7 +76,7 @@ presubmits:
           value: "."
         # TODO(bentheelder): reduce the skip list further
         - name: SKIP
-          value: \[Slow\]|\[Disruptive\]|\[Flaky\]|\[Feature:.+\]|PodSecurityPolicy|LoadBalancer|load.balancer|Simple.pod.should.support.exec.through.an.HTTP.proxy|subPath.should.support.existing
+          value: \[Slow\]|\[Disruptive\]|\[Flaky\]|\[Feature:.+\]|PodSecurityPolicy|LoadBalancer|load.balancer|Simple.pod.should.support.exec.through.an.HTTP.proxy|subPath.should.support.existing|inline.execution.and.attach
```
BenTheElder, line 79: this should've been on the other one @aojea :/

aojea: :upside_down_face:

PR #17991, opened 2020-06-18
lower cpu request

BenTheElder:
these are failing to schedule
we know jobs with 6000m have long existed
/cc @aojea


aojea:
/lgtm
:fire_engine: :fire: :fire_extinguisher:

BenTheElder:
we have a lot of nodes with 301m requested /  7.91 available so I don't know why this isn't working.

BenTheElder:
changed limitrange to Pod instead of container

config/prow/cluster/limitrange-testpods.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
+kind: LimitRange
+metadata:
+  name: testpods-limits
+  namespace: test-pods
+spec:
+  limits:
+  - defaultRequest:
+      cpu: 250m
+    type: Container
```
aojea, line 10: liner fails without EOL :shrug:

BenTheElder: this shouldn't be in the PR

BenTheElder:
these are scheduling again

PR #17992, opened 2020-06-18
install go in krte

BenTheElder:


images/krte/Dockerfile
```diff
@@ -24,6 +24,7 @@ FROM debian:buster
 ARG IMAGE_ARG
 # arg that specifies the bazel version to install
 ARG BAZEL_VERSION_ARG=0.23.1
+ARG GO_VERSION
```
aojea, line 27: no default?

BenTheElder: shrug, it's supplied by the build.

images/krte/Dockerfile
```diff
@@ -33,7 +34,8 @@ ARG BAZEL_VERSION_ARG=0.23.1
 # - hinting that we are in a docker container
 ENV IMAGE=${IMAGE_ARG} \
     BAZEL_VERSION=${BAZEL_VERSION_ARG} \
-    PATH=/google-cloud-sdk/bin:${PATH} \
+    GOPATH /home/prow/go \
```
BenTheElder, line 37: this requires an `=`
unfortunately nothing validates this in presubmit

BenTheElder: https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/post-test-infra-push-krte/1273674087943639041

PR #17993, opened 2020-06-18
bugzilla plugin: fix potential panic in cherrypick code

AlexNPavel:
This PR fixes an issue in the bugzilla plugin's cherrypick handling where the code was dereferencing a string pointer before checking whether it was `nil`, leading to panics when a cherrypick PR for a bug was opened in a repo where the target release was not set.

### DIFF
--- a/prow/plugins/bugzilla/bugzilla.go
+++ b/prow/plugins/bugzilla/bugzilla.go
@@ -187,7 +187,12 @@ type githubClient interface {
 	Query(ctx context.Context, q interface{}, vars map[string]interface{}) error
 }
 
-func handleGenericComment(pc plugins.Agent, e github.GenericCommentEvent) error {
+func handleGenericComment(pc plugins.Agent, e github.GenericCommentEvent) (err error) {
+	defer func() {
+		if r := recover(); r != nil {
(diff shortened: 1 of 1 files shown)


AlexNPavel:
Added deferred recover statements to the handlers.

prow/plugins/bugzilla/bugzilla.go
```diff
@@ -199,7 +208,16 @@ func handleGenericComment(pc plugins.Agent, e github.GenericCommentEvent) error
 	return nil
 }
 
-func handlePullRequest(pc plugins.Agent, pre github.PullRequestEvent) error {
+func handlePullRequest(pc plugins.Agent, pre github.PullRequestEvent) (err error) {
+	defer func() {
+		if r := recover(); r != nil {
+			if rec, ok := r.(error); ok {
```
alvaroaleman, line 214: Can this happen? I've never seen that before

AlexNPavel: Changed it to just use `fmt.Errorf` without any casting. Casting the interface seems mainly to be useful if we specifically panic with our own type/struct (`encoding/json` does this for example), but we're not using any special types here.

Issue #17997, opened 2020-06-18
Transfigure image failed to run due to bazel version mismatch


chaodaiG:
**What happened**:
Running transfigure with latest image failed due to:
```
Running kubernetes/test-infra tests...
ERROR: The project you're trying to build requires Bazel 3.0.0 (specified in /home/prow/go/src/github.com/GoogleCloudPlatform/oss-test-infra/test-infra/.bazelversion), but it wasn't found in /usr/local/lib/bazel/bin.
```

Link: https://oss-prow.knative.dev/view/gs/oss-prow/pr-logs/pull/GoogleCloudPlatform_oss-test-infra/395/pull-oss-test-infra-check-testgrid-config/1273664592261681152

**What you expected to happen**:
Transfigure run as expected

**How to reproduce it (as minimally and precisely as possible)**:

**Please provide links to example occurrences, if any**:

**Anything else we need to know?**:

BenTheElder:
Is this still the case?

chaodaiG:
It's fixed by #17999 
/close

PR #17999, opened 2020-06-18
Update bazel-base to latest test-infra variant

fejta:
Document the need to do this.

This is the one we are currently using: https://github.com/kubernetes/test-infra/blob/47fde963491bba2175a497960e677cbef338b75b/config/jobs/kubernetes/test-infra/test-infra-postsubmits.yaml#L12

### DIFF
--- a/containers.bzl
+++ b/containers.bzl
@@ -73,8 +73,8 @@ def repositories():
 
     container_pull(
         name = "bazel-base",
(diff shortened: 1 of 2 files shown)


fejta:
fixes https://github.com/kubernetes/test-infra/issues/17997

PR #18000, opened 2020-06-18
Update cloud.google.com/go past split into submodules

spiffxp:
For access to secretmanager/apiv1 client with IAM support

To unblock https://github.com/kubernetes/test-infra/pull/17946


spiffxp:
Seems like some of this was due to an older version coming in through testgrid; https://github.com/GoogleCloudPlatform/testgrid/pull/141 should address that, but I'm still bumping into things locally

PR #18001, opened 2020-06-18
Deck: supporting restrictions on buckets/folders

e-blackwelder:
_Note_: This is a rebase/follow-up to [pr/17650](https://github.com/kubernetes/test-infra/pull/17650), which was originally started by clarketm.

Added _bucket_ and _folder_ validation features to Deck to restrict artifact requests to only specific storage locations. This feature can be enabled by setting `deck.restrict_storage_paths` in the Prow config.

By default, _buckets_ specified in the `plank.decoration_config[*].gcs_configuration.bucket` and 
     the `pr-logs` and `logs` _folders_ are allowed allowed. _Additional buckets_ can be allowed by adding them to 
     `deck.additional_allowed_buckets`. _Additional folders_ can be allowed by adding them to `deck.additional_allowed_folders`.

### DIFF
--- a/prow/ANNOUNCEMENTS.md
+++ b/prow/ANNOUNCEMENTS.md
@@ -13,11 +13,11 @@ New features added to each component:
     in *September 2020*. To be clear handling of URLs with `/view/gcs` in Deck is not deprecated.
  - *June 23rd, 2020* An [hmac](/prow/cmd/hmac) tool was added to automatically reconcile webhooks and hmac
     tokens for the orgs and repos integrated with your prow instance.
-- *June 12, 2020* Added _bucket_ and _folder_ whitelisting to Deck in order to restrict artifact requests to 
-     specific storage locations. This feature can be enabled by setting `deck.enable_whitelist` in the Prow config.
-     By default, _buckets_ specified in the `plank.decoration_config[*].gcs_configuration.bucket` are whitelisted and 
-     the `pr-logs` and `logs` _folders_ are whitelisted. Additional _buckets_ can be whitelisted by adding them to 
-     `deck.additional_buckets`. Additional _folders_ can be whitelisted by adding them to `deck.additional_folders`.
+- *June 12, 2020* Added _bucket_ and _folder_ validation features to Deck to restrict artifact requests to only
+     specific storage locations. This feature can be enabled by setting `deck.restrict_storage_paths` in the Prow config.
+     By default, _buckets_ specified in the `plank.decoration_config[*].gcs_configuration.bucket` and 
+     the `pr-logs` and `logs` _folders_ are allowed allowed. _Additional buckets_ can be allowed by adding them to 
+     `deck.additional_allowed_buckets`. _Additional folders_ can be allowed by adding them to `deck.additional_allowed_folders`.
  - *June 8th, 2020* A new informer-based Plank implementation was added. It can be used by deploying
     the new [prow-controller-manager](/config/prow/experimental/controller_manager.yaml) binary.
     We plan to gradually move all our controllers into that binary, see https://github.com/kubernetes/test-infra/issues/17024
--- a/prow/ANNOUNCEMENTS.md
+++ b/prow/ANNOUNCEMENTS.md
@@ -13,6 +13,11 @@ New features added to each component:
     in *September 2020*. To be clear handling of URLs with `/view/gcs` in Deck is not deprecated.
  - *June 23rd, 2020* An [hmac](/prow/cmd/hmac) tool was added to automatically reconcile webhooks and hmac
     tokens for the orgs and repos integrated with your prow instance.
+- *June 12, 2020* Added _bucket_ and _folder_ whitelisting to Deck in order to restrict artifact requests to 
+     specific storage locations. This feature can be enabled by setting `deck.enable_whitelist` in the Prow config.
+     By default, _buckets_ specified in the `plank.decoration_config[*].gcs_configuration.bucket` are whitelisted and 
+     the `pr-logs` and `logs` _folders_ are whitelisted. Additional _buckets_ can be whitelisted by adding them to 
+     `deck.additional_buckets`. Additional _folders_ can be whitelisted by adding them to `deck.additional_folders`.
  - *June 8th, 2020* A new informer-based Plank implementation was added. It can be used by deploying
     the new [prow-controller-manager](/config/prow/experimental/controller_manager.yaml) binary.
     We plan to gradually move all our controllers into that binary, see https://github.com/kubernetes/test-infra/issues/17024
--- a/prow/ANNOUNCEMENTS.md
+++ b/prow/ANNOUNCEMENTS.md
@@ -3,6 +3,12 @@
 ## New features
 
 New features added to each component:
+  - *September 15, 2020* Added validation to Deck that will restrict artifact requests based on storage buckets.
+    Opt-out by setting `deck.skip_storage_path_validation` in your Prow config.
+    Buckets specified in job configs (`<job>.gcs_configuration.bucket`) and 
+    plank configs (`plank.default_decoration_configs[*].gcs_configuration.bucket`) are automatically allowed access.
+    Additional buckets can be allowed by adding them to the `deck.additional_allowed_buckets` list.
+    (This feature will be enabled by default ~Jan 2021. For now, you will begin to notice violation warnings in your logs.)
  - *August 31th, 2020* Added `gcs_browser_prefixes` field in spyglass configuration. `gcs_browser_prefix` will
     be deprecated in February 2021. You can now specify different values for different repositories. The
     format should be in org, org/repo or '\*' which is the default value.
@@ -13,11 +19,6 @@ New features added to each component:
     in *September 2020*. To be clear handling of URLs with `/view/gcs` in Deck is not deprecated.
  - *June 23rd, 2020* An [hmac](/prow/cmd/hmac) tool was added to automatically reconcile webhooks and hmac
     tokens for the orgs and repos integrated with your prow instance.
-- *June 12, 2020* Added _bucket_ and _folder_ validation features to Deck to restrict artifact requests to only
-     specific storage locations. This feature can be enabled by setting `deck.restrict_storage_paths` in the Prow config.
-     By default, _buckets_ specified in the `plank.decoration_config[*].gcs_configuration.bucket` and 
-     the `pr-logs` and `logs` _folders_ are allowed allowed. _Additional buckets_ can be allowed by adding them to 
-     `deck.additional_allowed_buckets`. _Additional folders_ can be allowed by adding them to `deck.additional_allowed_folders`.
  - *June 8th, 2020* A new informer-based Plank implementation was added. It can be used by deploying
     the new [prow-controller-manager](/config/prow/experimental/controller_manager.yaml) binary.
     We plan to gradually move all our controllers into that binary, see https://github.com/kubernetes/test-infra/issues/17024
--- a/prow/cmd/deck/main.go
+++ b/prow/cmd/deck/main.go
@@ -69,7 +69,6 @@ import (
 	"k8s.io/test-infra/prow/pluginhelp"
 	"k8s.io/test-infra/prow/plugins"
 	"k8s.io/test-infra/prow/plugins/trigger"
-	"k8s.io/test-infra/prow/pod-utils/gcs"
 	"k8s.io/test-infra/prow/prstatus"
 	"k8s.io/test-infra/prow/simplifypath"
 	"k8s.io/test-infra/prow/spyglass"
@@ -893,8 +892,10 @@ func handleJobHistory(o options, cfg config.Getter, opener io.Opener, log *logru
 		tmpl, err := getJobHistory(r.Context(), r.URL, cfg, opener)
 		if err != nil {
 			msg := fmt.Sprintf("failed to get job history: %v", err)
-			log.WithField("url", r.URL.String()).Warn(msg)
-			http.Error(w, msg, http.StatusInternalServerError)
+			if shouldLogHTTPErrors(err) {
+				log.WithField("url", r.URL.String()).Warn(msg)
+			}
+			http.Error(w, msg, httpStatusForError(err))
 			return
 		}
 		handleSimpleTemplate(o, cfg, "job-history.html", tmpl)(w, r)
@@ -936,9 +937,11 @@ func handleRequestJobViews(sg *spyglass.Spyglass, cfg config.Getter, o options, 
 		csrfToken := csrf.Token(r)
 		page, err := renderSpyglass(r.Context(), sg, cfg, src, o, csrfToken, log)
 		if err != nil {
-			log.WithError(err).Error("error rendering spyglass page")
-			message := fmt.Sprintf("error rendering spyglass page: %v", err)
-			http.Error(w, message, http.StatusInternalServerError)
+			msg := fmt.Sprintf("error rendering spyglass page: %v", err)
+			if shouldLogHTTPErrors(err) {
+				log.WithError(err).Error(msg)
+			}
+			http.Error(w, msg, httpStatusForError(err))
 			return
 		}
 
@@ -962,8 +965,8 @@ func renderSpyglass(ctx context.Context, sg *spyglass.Spyglass, cfg config.Gette
 		return "", fmt.Errorf("error when resolving real path %s: %v", src, err)
 	}
 	src = realPath
-	if err := ValidateStoragePath(cfg, src); err != nil {
-		return "", err // TODO(eblackwelder@): indicate it's http.StatusBadRequest (from inside ValidateStoragePath)
+	if err := validateStorageBucket(cfg, src); err != nil {
+		return "", err
 	}
 	artifactNames, err := sg.ListArtifacts(ctx, src)
 	if err != nil {
@@ -1180,8 +1183,8 @@ func handleArtifactView(o options, sg *spyglass.Spyglass, cfg config.Getter) htt
 			http.Error(w, fmt.Sprintf("Failed to parse request: %v", err), http.StatusBadRequest)
 			return
 		}
-		if err := ValidateStoragePath(cfg, request.Source); err != nil {
-			http.Error(w, fmt.Sprintf("Failed to process request: %v", err), http.StatusBadRequest)
+		if err := validateStorageBucket(cfg, request.Source); err != nil {
+			http.Error(w, fmt.Sprintf("Failed to process request: %v", err), httpStatusForError(err))
 			return
 		}
 
@@ -1632,38 +1635,28 @@ func defaultLensRemoteConfig(lfc *config.LensFileConfig) error {
 	return nil
 }
 
-// ValidateStoragePath validates the storage bucket/folder path, if the `Deck.RestrictStoragePaths` config field is true.
-// The path should match "<key-type>/<bucket>/<folder>/<sub-folder...>" pattern.
-// Additionally, the path bucket or folder should match either the built-in and default storage locations _or_
-// should be contained in the `Deck.AdditionalAllowedBuckets/Folders` config fields.
-// If the `Deck.RestrictStoragePaths` field is false, no validation is performed (and no errors are returned).
-func ValidateStoragePath(cfg config.Getter, path string) error {
-	if !cfg().Deck.RestrictStoragePaths {
-		return nil
+func validateStorageBucket(cfg config.Getter, path string) error {
+	if err := cfg().ValidateStorageBucket(path); err != nil {
+		return httpError{
+			error:      err,
+			statusCode: http.StatusBadRequest,
+		}
 	}
-
-	parts := strings.SplitN(path, "/", 4)
-	if len(parts) < 4 {
-		return fmt.Errorf("invalid path: %s expected <key-type>/<bucket>/<folder>/<sub-folder...>", path)
-	}
-
-	// Check bucket
-	bucket := parts[1]
-	allowed_buckets := sets.NewString(cfg().Deck.AdditionalAllowedBuckets...)
-	for _, dc := range cfg().Plank.DefaultDecorationConfigs {
-		allowed_buckets.Insert(dc.GCSConfiguration.Bucket)
-	}
-	if !allowed_buckets.Has(bucket) {
-		return fmt.Errorf("bucket %q not in allowed list %v", bucket, allowed_buckets)
-	}
-
-	// Check folder
-	folder := parts[2]
-	allowed_folders := sets.NewString(gcs.PRLogs, gcs.NonPRLogs)
-	allowed_folders.Insert(cfg().Deck.AdditionalAllowedFolders...)
-	if !allowed_folders.Has(folder) {
-		return fmt.Errorf("folder %q not in allowed list %v", folder, allowed_folders)
-	}
-
 	return nil
+}
+
+type httpError struct {
+	error
+	statusCode int
+}
+
+func httpStatusForError(e error) int {
+	if httpErr, ok := e.(httpError); ok {
+		return httpErr.statusCode
+	}
+	return http.StatusInternalServerError
+}
+
+func shouldLogHTTPErrors(e error) bool {
+	return httpStatusForError(e) >= http.StatusInternalServerError // 5XX
 }
--- a/prow/cmd/deck/main.go
+++ b/prow/cmd/deck/main.go
@@ -69,6 +69,7 @@ import (
 	"k8s.io/test-infra/prow/pluginhelp"
 	"k8s.io/test-infra/prow/plugins"
 	"k8s.io/test-infra/prow/plugins/trigger"
+	"k8s.io/test-infra/prow/pod-utils/gcs"
 	"k8s.io/test-infra/prow/prstatus"
 	"k8s.io/test-infra/prow/simplifypath"
 	"k8s.io/test-infra/prow/spyglass"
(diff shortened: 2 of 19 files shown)


prow/cmd/deck/main.go
```diff
@@ -948,7 +949,9 @@ func renderSpyglass(ctx context.Context, sg *spyglass.Spyglass, cfg config.Gette
 		return "", fmt.Errorf("error when resolving real path %s: %v", src, err)
 	}
 	src = realPath
-
+	if err := ValidateStoragePath(cfg, src); err != nil {
+		return "", err // TODO(eblackwelder@): indicate it's http.StatusBadRequest (from inside ValidateStoragePath)
```
e-blackwelder, line 953: I have an idea how to address this, but would rather implement it in a separate patch:
* create an `httpError` type that passes a status code along with the error message (unless we already have that type)
* create a helper method to "getHttpStatusCode" for a given `error` (by default 500, but checking for httpError's status code)
* return an `httpError` in `ValidateStoragePath`
* use the `getHttpStatusCode` method in various places: `handleArtifactView`, etc

cjwagner: That sounds reasonable. Alternatively `renderSpyglass` could just be updated to add an HTTP status code return value.
I think this should be handled in this patch though rather than add new code that we know to be incorrect.

e-blackwelder: Implemented.

prow/cmd/deck/main.go
```diff
@@ -876,7 +877,7 @@ func handleBadge(ja *jobs.JobAgent) http.HandlerFunc {
 func handleJobHistory(o options, cfg config.Getter, opener io.Opener, log *logrus.Entry) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
 		setHeadersNoCaching(w)
-		tmpl, err := getJobHistory(r.Context(), r.URL, opener)
+		tmpl, err := getJobHistory(r.Context(), r.URL, cfg, opener)
```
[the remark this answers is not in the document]
e-blackwelder: Thanks. I had missed that reference (`ValidateStoragePath` is also called in `job_history.go`). Did a thorough job this time.

prow/cmd/deck/main.go
```diff
@@ -924,7 +925,7 @@ func handleRequestJobViews(sg *spyglass.Spyglass, cfg config.Getter, o options,
 		if err != nil {
 			log.WithError(err).Error("error rendering spyglass page")
```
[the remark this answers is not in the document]
e-blackwelder: Done. Skipping the logs for these.

prow/cmd/deck/main.go
```diff
@@ -1607,3 +1618,64 @@ func defaultLensRemoteConfig(lfc *config.LensFileConfig) error {
...
+	}
+	if !allowed_buckets.Has(bucket) {
+		return httpError{
+			error:      fmt.Errorf("bucket %q not in allowed list %v", bucket, allowed_buckets),
+			statusCode: http.StatusBadRequest,
+		}
+	}
+
+	// Check folder
+	folder := parts[2]
```
alvaroaleman, line 1654: I do understand the desire to limit the bucket but I am not sure if the added value of verifying the folder warrants the added configuration complexity. If this is really needed (and I can not think of a good reason why, we assume so far that a bucket is a single trust domain) ppl could still configure this via their cloud providers IAM. WDYT?

prow/ANNOUNCEMENTS.md
```diff
@@ -3,6 +3,11 @@
 ## New features
 
 New features added to each component:
+- *June 12, 2020* Added _bucket_ and _folder_ validation features to Deck to restrict artifact requests to only
```
alvaroaleman, line 6: I think it would be good to do this by default

prow/cmd/deck/main_test.go
```diff
@@ -1050,3 +1050,121 @@ func TestCanTriggerJob(t *testing.T) {
 		}
 	}
 }
+
+func TestValidateStoragePath(t *testing.T) {
+	testCases := []struct {
+		name        string
+		config      config.ProwConfig
+		path        string
+		expectedErr bool
```
alvaroaleman, line 1059: Mind testing for the actual error/error message rather than a bool? Tests that use a bool to check for an error tend to rot over time if for some reason a different error is returned

prow/ANNOUNCEMENTS.md
e-blackwelder: Since this is a breaking change, we won't introduce it immediately. I will use a `*bool` for the "turn it on" flag so that we can detect whether it is explicitly set. I will also add warnings to the logs to indicate when the settings need to be updated.

prow/cmd/deck/main.go
e-blackwelder: Per our discussion (with Cole), I will remove folder-level validation, but leave bucket validation. I will instead introduce a file-size validator to fill a similar role.

prow/cmd/deck/main_test.go
e-blackwelder: Done.

e-blackwelder:
@alvaroaleman & @cjwagner: There is already a [Deck.Spyglass.SizeLimit](https://github.com/kubernetes/test-infra/blob/e0951ec387d72ee03fd573ca80cbf95c2f74aed7/prow/config/config.go#L616-L621) config field with _some_ validation surrounding it. I'll do a separate PR to address the gaps.

prow/cmd/deck/main.go
e-blackwelder: FYI: There is already a [Deck.Spyglass.SizeLimit](https://github.com/kubernetes/test-infra/blob/e0951ec387d72ee03fd573ca80cbf95c2f74aed7/prow/config/config.go#L616-L621) config field, with _some_ validation surrounding it. I'll do a separate PR to address artifact size limit gaps.

e-blackwelder: Done. (Removed the folder-level validation, as requested/discussed.)

prow/ANNOUNCEMENTS.md
e-blackwelder: Done. (Warnings added, "in the future" comments added, etc.)

prow/ANNOUNCEMENTS.md
```diff
@@ -5,6 +5,12 @@
 New features added to each component:
  - *June 23rd, 2020* An [hmac](/prow/cmd/hmac) tool was added to automatically reconcile webhooks and hmac
     tokens for the orgs and repos integrated with your prow instance.
+- *June 12, 2020* Added _bucket_ and _folder_ validation features to Deck to restrict artifact requests to only
+     specific storage locations. This feature can be enabled by setting `deck.restrict_storage_paths` in the Prow config.
+     By default, _buckets_ specified in the `plank.decoration_config[*].gcs_configuration.bucket` and 
+     the `pr-logs` and `logs` _folders_ are allowed allowed. _Additional buckets_ can be allowed by adding them to
```
alvaroaleman, line 11: s/allowed allowed/allowed and the folders part isn't accurate anymore

prow/cmd/deck/main.go
```diff
@@ -1607,3 +1617,68 @@ func defaultLensRemoteConfig(lfc *config.LensFileConfig) error {
 
 	return nil
 }
+
+// ValidateStoragePath validates the storage bucket, if the `Deck.RestrictStoragePaths` config field is true.
+// The path should match a generic "<key-type>/<bucket>/<folder>/<sub-folder...>" pattern.
+// Additionally, the path bucket should match either the built-in and default storage locations _or_
+// should be contained in the `Deck.AdditionalAllowedBuckets` config field.
+// If the `Deck.RestrictStoragePaths` field is false, no validation errors are returned (but there will be warnings logged).
+func ValidateStoragePath(cfg config.Getter, path string) error {
```
alvaroaleman, line 1626: Why is this exported? It is not used outside of this package or did I miss that?

prow/cmd/deck/main_test.go
```diff
@@ -1050,3 +1051,115 @@ func TestCanTriggerJob(t *testing.T) {
...
+			expectedErr: "bucket \"istio-prow\" not in allowed list",
+			httpStatus:  http.StatusBadRequest,
+		},
+	}
+
+	for _, tc := range testCases {
+		cfg := config.Config{ProwConfig: tc.config}
+		ca := config.Agent{}
+		ca.Set(&cfg)
+		expectingErr := len(tc.expectedErr) > 0
```
alvaroaleman, line 1127: While not technically incorrect its confusing to have part of the testcase in the main test and part in ths sub test, could you move all of it into the func that is called below with `t.Run`?

prow/config/config.go
```diff
@@ -659,6 +662,53 @@ type Deck struct {
...
+
+func (d *Deck) ShouldValidateStoragePaths() bool {
+	if d.RestrictStoragePaths == nil {
+		return RestrictStoragePathsByDefault
+	}
+	return *d.RestrictStoragePaths
+}
+
+func (d *Deck) ShouldWarnAboutInvalidStoragePaths() bool {
+	return RestrictStoragePathsByDefault == false && d.RestrictStoragePaths == nil
```
alvaroaleman, line 711: should just be `!RestrictStoragePathsByDefault` but wouldn't it make more sense to warn unconditionally since we want to turn this on by default?

prow/ANNOUNCEMENTS.md
e-blackwelder: Done. (Thanks for noticing; pretty sure this was an oversight after fixing a merge conflict. I've added a Markdown previewer to my IDE now.)

prow/cmd/deck/main.go
e-blackwelder: Done. (No, it doesn't seem to be used outside. It has been lowerCamelCased.)

prow/cmd/deck/main_test.go
e-blackwelder: Done.

prow/config/config.go
e-blackwelder: I think my intent was to return true *only* when they were restricted _by default_. (I figured if they've already turned off validation when it's off-by-default, they'll leave validation off when it's turned on-by-default.) I added a method comment and made the code more straightforward. Thoughts?

prow/cmd/deck/main.go
```diff
@@ -1607,3 +1617,65 @@ func defaultLensRemoteConfig(lfc *config.LensFileConfig) error {
...
+
+// validateStoragePath "validates" the storage bucket, *if* the `Deck.RestrictStoragePaths` config field is true.
+// The path should match a generic "<key-type>/<bucket>/<folder>/<sub-folder...>" pattern.
+// Additionally, the path's bucket should match either the built-in and default storage locations _or_
+// should be contained in the `Deck.AdditionalAllowedBuckets` config field.
+// If the `Deck.RestrictStoragePaths` field is false, no validation errors are returned (but there will be warnings logged).
+func validateStoragePath(cfg config.Getter, path string) error {
+	validate := func(cfg config.Getter, path string) error {
+		parts := strings.SplitN(path, "/", 4)
+		if len(parts) < 4 {
```
alvaroaleman, line 1629: I don't think this check is correct, for example for a 
* Presubmit, its seven elemtents: https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gcs/origin-ci-test/pr-logs/pull/openshift_installer/3840/pull-ci-openshift-installer-master-images/1279069653829685248
* periodic, its four elements: https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gcs/origin-ci-test/pr-logs/pull/openshift_installer/3840/pull-ci-openshift-installer-master-images/1279069653829685248
* For a batch, its six elements: https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gcs/origin-ci-test/pr-logs/pull/batch/pull-ci-integr8ly-integreatly-operator-master-manifest/1279053551028408320

On top of that there are multiple strategies that can be used and that determine the structure on the object storage so IMHO its very dangerous to make any assumptions about that here

prow/cmd/deck/main.go
```diff
@@ -1607,3 +1617,65 @@ func defaultLensRemoteConfig(lfc *config.LensFileConfig) error {
...
+		if !allowed_buckets.Has(bucket) {
+			return httpError{
+				error:      fmt.Errorf("bucket %q not in allowed list %v", bucket, allowed_buckets),
+				statusCode: http.StatusBadRequest,
+			}
+		}
+
+		return nil
+	}
+	actualErr := validate(cfg, path)
```
alvaroaleman, line 1651: Nit: Why is this named `validate`? Also a nit but I personally find code easier to read without inline funcs because with inline funcs you have to read all the logic rather than being able to say `okay we validate, if there was an error we do X etc`

prow/config/config.go
```diff
@@ -659,6 +662,57 @@ type Deck struct {
...
+	}
+
+	return nil
+}
+
+// ShouldValidateStoragePaths returns whether or not the Deck's storage path should be validated.
+// Validation could be either disabled by default or explicitly turned off.
+func (d *Deck) ShouldValidateStoragePaths() bool {
+	if d.RestrictStoragePaths == nil {
+		return RestrictStoragePathsByDefault
```
alvaroaleman, line 708: What is the reason of this additional indirection with the constant? Nothing except this func uses the constant or did I miss it? Furthermore I think nothing _should_ use this constant, if anything cares about this setting it should use this func instead.

prow/config/config.go
```diff
@@ -659,6 +662,57 @@ type Deck struct {
...
+	}
+	return *d.RestrictStoragePaths
+}
+
+// ShouldWarnAboutInvalidStoragePaths returns true unless the Deck is explicitly configured to *not* validate storage paths.
+func (d *Deck) ShouldWarnAboutInvalidStoragePaths() bool {
```
alvaroaleman, line 714: Is there a reason to put this into this func that has exactly one caller rather than doing that check in this exactly one caller? If we want to test the logging, shouldn't we instead just do that rather than testing the func that we think will cause the logging to happen?

prow/config/config.go
e-blackwelder: I started with just a constant, then refactored in the function later. I think my (vaguely intuitive) "reasons" for leaving both was to make it easier to change-then-delete in the future... `RestrictStoragePathsByDefault` is defined directly beside the `Deck` definition and named extremely similar to the `deck.RestrictStoragePaths` field. Toggling a bool named "ByDefault" would also feel "safer" than re-evaluating the contents of a method like "ShouldValidateSomething". I agree that consumers should be using the function over the constant.

But I'm admittedly _not_ a go-lang expert and I'm guessing/realizing that the additional layers/abstractions aren't idiomatic Go. I'll happily remove them, per your feedback.

prow/cmd/deck/main.go
e-blackwelder: (FYI: This part was originally Travis's change, in case you didn't know/realize.)

I _think_ you're saying that len(parts) should be somewhere between 4 and 7, depending on the type of job, but I'm not sure where you're seeing those numbers (of elements). It appears that the `SplitN(..., 4)` function limits the returned slice to _at most 4_ pieces ([golang docs](https://golang.org/pkg/strings/#SplitN)), meaning the `<sub-folder>` piece is just checked for existence.

That being said, I've personally had a bit of a problem with the pseudo-URL format of the `path` parameter -- I would have expected a "://" after the "<key-type>" (as opposed to a single "/"). I had previously noticed at least one caller (possibly the sole caller) is string-formatting the path from various pieces. I think I'd rather just pass and validate in the path as a struct, which would mean more code, but I think we'd both be satisfied.

prow/config/config.go
e-blackwelder: As far as "reasons"... just my personal experience/style, I'd say. My preference is usually something resembling "single level of abstraction per function". Because (for example), it's easier to write confidence-building unit-tests for tiny methods like this vs "did Xxxlogging framework get called with the right varargs". But (as I mention above), I'll defer to your Go-lang sensibilities. :)

prow/cmd/deck/main.go
e-blackwelder: Sorry, refactoring noise. I'll take another pass. (Not quite satisfied with where I left this, either.)

prow/cmd/deck/main.go
```diff
@@ -1607,3 +1617,65 @@ func defaultLensRemoteConfig(lfc *config.LensFileConfig) error {
...
+			return httpError{
+				error:      fmt.Errorf("invalid path: %s expected <key-type>/<bucket>/<folder>/<sub-folder...>", path),
+				statusCode: http.StatusBadRequest,
+			}
+		}
+
+		// Check bucket
+		bucket := parts[1]
+		allowed_buckets := sets.NewString(cfg().Deck.AdditionalAllowedBuckets...)
+		for _, dc := range cfg().Plank.DefaultDecorationConfigs {
```
alvaroaleman, line 1639: There is another source of valid buckets we have to consider: They can be set on jobs. So we need to walk all jobs, check if they have `.DecorationConfig != nil && .DecorationConfig.GCSConfiguration != nil && .DecorationConfig.GCSConfiguration.BucketName != "`.

Maybe it would be better to construct that set in `Load` and store it somewhere on the config?

prow/cmd/deck/main.go
e-blackwelder: I looked into the [Path Strategies](https://github.com/kubernetes/test-infra/blob/088d21b2a38643ff55e59069228f449dacd4522e/prow/apis/prowjobs/v1/types.go#L529-L531) and they do have a few different ways of formatting paths. Since I'm only concerned with bucket names now, they can all be validated in the same way -- which is just "is there a folder".

prow/config/config.go
e-blackwelder: Removed the extra `RestrictStoragePathsByDefault` const.

prow/config/config.go
e-blackwelder: Removed extraneous function `ShouldWarnAboutInvalidStoragePaths`.

prow/cmd/deck/main.go
e-blackwelder: Done.

_Note_: This can't handle buckets defined on jobs via InRepoConfig. For now (until ~Jan 2021), I'm skipping validation on InRepoConfig-enabled Prows and providing a detailed warning on how to prepare.

e-blackwelder:
> So we do not do size validation for now?

See https://github.com/kubernetes/test-infra/pull/19250

prow/cmd/deck/main_test.go
```diff
@@ -1166,3 +1166,23 @@ func TestCanTriggerJob(t *testing.T) {
 		}
 	}
 }
+
+func TestHttpStatusForError(t *testing.T) {
+	{
```
cjwagner, line 1171: nit: A slice of testcase structs `{err error, expectedCode int}` and a loop would be a bit more idiomatic and less repetitive, but it doesn't really matter here.

prow/config/config.go
```diff
@@ -693,6 +694,121 @@ type Deck struct {
...
+	}
+
+	return nil
+}
+
+var warnInRepoStorageBucketValidation time.Time
+
+// ValidateStorageBucket validates a storage bucket, *if* the `Deck.RestrictStoragePaths` config field is true.
+// The given storage path should match "<key-type>/<bucket>/<folders...>".
+// The bucket name should match either the built-in and default storage locations _or_
```
cjwagner, line 744: "the built-in and default storage locations" is not clear to me. Please elaborate here or link/reference documentations that does.

prow/config/config.go
```diff
@@ -693,6 +694,121 @@ type Deck struct {
...
+// should be contained in the `Deck.AdditionalAllowedBuckets` config field.
+// If the `Deck.RestrictStoragePaths` field is false, no validation errors are returned.
+func (c *Config) ValidateStorageBucket(path string) error {
+	parts := strings.Split(path, "/")
+	if len(parts) < 3 {
+		return fmt.Errorf("invalid path: %s expected <key-type>/<bucket>/<folders...>", path)
+	}
+
+	if len(c.InRepoConfig.Enabled) > 0 && len(c.Deck.AdditionalAllowedBuckets) == 0 {
+		logrusutil.ThrottledWarnf(&warnInRepoStorageBucketValidation, 1*time.Hour, strings.Join([]string{
```
cjwagner, line 754: nit: You don't need to use `strings.Join` here. You can
1. use ``` ` ``` for multiline raw strings
1. concatenate lines with `+` (the compiler should be smart enough to do this for literals at compile time rather than runtime)
1. use a long single line literal since this is for an error message.

I think those options are more efficient and more readable.

prow/config/config.go
```diff
@@ -693,6 +694,121 @@ type Deck struct {
...
+	}
+
+	bucket := parts[1]
+	if !c.Deck.ShouldValidateStorageBuckets() {
+		return nil
+	}
+
+	if !c.Deck.AllKnownStorageBuckets.Has(bucket) {
+		knownBuckets := []string{}
+		for b := range c.Deck.AllKnownStorageBuckets {
```
cjwagner, line 769: https://godoc.org/github.com/kubernetes/apimachinery/pkg/util/sets#String.List

prow/config/config.go
```diff
@@ -693,6 +694,121 @@ type Deck struct {
...
+	if len(c.InRepoConfig.Enabled) > 0 && len(c.Deck.AdditionalAllowedBuckets) == 0 {
+		logrusutil.ThrottledWarnf(&warnInRepoStorageBucketValidation, 1*time.Hour, strings.Join([]string{
+			"skipping storage-path validation because `in_repo_config` is enabled, but `deck.additional_allowed_buckets` empty. ",
+			"Validation will be enabled by default in January 2021. ",
+			"To disable this message, populate `deck.additional_allowed_buckets` with at least one storage bucket. ",
+			"(Note: when `deck.additional_allowed_buckets` is populated, this message will be disabled.)"}, ""))
+		return nil
+	}
+
+	bucket := parts[1]
```
petr-muller, line 762: nit: move this immediately before the usage, after the `ShouldValiate...` check which  does not use this.

prow/logrusutil/logrusutil.go
```diff
@@ -129,3 +131,33 @@ func NewCensoringFormatter(f logrus.Formatter, getSecrets func() sets.String) Ce
 		delegate:   f,
 	}
 }
+
+// ThrottledWarnf prints a warning the first time called and if at most `freq` has elapsed since the last time.
+func ThrottledWarnf(last *time.Time, freq time.Duration, format string, args ...interface{}) {
```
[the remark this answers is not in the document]
e-blackwelder: Done.

prow/plugins/config.go
```diff
@@ -1742,3 +1719,9 @@ type Override struct {
 	// members of which are allowed to override contexts
 	AllowedGitHubTeams map[string][]string `json:"allowed_github_teams,omitempty"`
 }
+
+// warnDeprecated is deprecated; use logrusutil.ThrottledWarnf instead.
+// TODO(e-blackwelder): remove <= Jan 2021
+func warnDeprecated(last *time.Time, freq time.Duration, format string) {
```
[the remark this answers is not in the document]
e-blackwelder: Done (you troll :P).

prow/config/config.go
e-blackwelder: Done.

prow/config/config.go
e-blackwelder: Done.

prow/config/config.go
e-blackwelder: Done.

prow/config/config.go
e-blackwelder: Done.

e-blackwelder (commented):
/hold

Would like to wait for Alvaro to review. (And still have a TODO from Cole.)

prow/config/config.go
```diff
@@ -693,6 +694,121 @@ type Deck struct {
 	// accepts a key of: `org/repo`, `org` or `*` (wildcard) to define what GitHub org (or repo) a particular
 	// config applies to and a value of: `RerunAuthConfig` struct to define the users/groups authorized to rerun jobs.
 	RerunAuthConfigs RerunAuthConfigs `json:"rerun_auth_configs,omitempty"`
+	// RestrictStoragePaths restricts artifact requests to specific buckets.
+	// By default, buckets listed in the GCSConfiguration are automatically allowed.
+	// Additional locations can be allowed via `AdditionalAllowedBuckets` fields.
+	// When unspecified (nil), it defaults to false (until ~Jan 2021).
+	RestrictStoragePaths *bool `json:"restrict_storage_paths,omitempty"`
```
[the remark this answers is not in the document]
e-blackwelder: Done.

prow/cmd/deck/main.go
```diff
@@ -961,7 +965,9 @@ func renderSpyglass(ctx context.Context, sg *spyglass.Spyglass, cfg config.Gette
 		return "", fmt.Errorf("error when resolving real path %s: %v", src, err)
 	}
 	src = realPath
-
+	if err := validateStorageBucket(cfg, src); err != nil {
```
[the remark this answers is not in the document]
e-blackwelder: Done. Moved validation into `newBlobStorageBucket` (job_history.go) and `StorageArtifactFetcher.newStorageJobSource` (with minor refactoring to make it work).

Balancing URL-formatting (gs/bucket/folder vs gs://bucket/folder) and bucket-vs-full-path was tricky, but method names, param names, and comments should fully reflect expectations for each.

e-blackwelder:
> Would like to wait for Alvaro to review. (And still have a TODO from Cole.)

Finished final TODO from Cole's review.

prow/spyglass/spyglass.go
```diff
@@ -85,7 +85,7 @@ func New(ctx context.Context, ja *jobs.JobAgent, cfg config.Getter, opener pkgio
 		JobAgent:               ja,
 		config:                 cfg,
 		PodLogArtifactFetcher:  NewPodLogArtifactFetcher(ja),
-		StorageArtifactFetcher: NewStorageArtifactFetcher(opener, useCookieAuth),
+		StorageArtifactFetcher: NewStorageArtifactFetcher(opener, cfg().ValidateStorageBucket, useCookieAuth),
```
cjwagner, line 88: We should pass a dynamic config getter here rather than a method of the static config struct. As is, if a new job is added with a new bucket, we'd have to restart Deck in order for it to respect the newly valid bucket.

e-blackwelder: Done.

cjwagner (approved):

PR #18004, opened 2020-06-18
Implemented Down() in GCE deployer for cloud-provider-gcp

michaelmdresser:
This is based on the MVP/POC (targeted at cloud-provider-gcp) implementation of Up() (#17964). Uses the same flags, environment, and process as Up(), but doesn't have to handle enable compute APIs because Up() should have already done that.

Tested with `--build --up --down` and verified that resources were deleted manually.

### DIFF
--- a/kubetest2/kubetest2-gce/deployer/down.go
+++ b/kubetest2/kubetest2-gce/deployer/down.go
@@ -1,0 +1,47 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


kubetest2/kubetest2-gce/deployer/down.go
```diff
@@ -0,0 +1,47 @@
...
+)
+
+func (d *deployer) Down() error {
+	klog.Info("GCE deployer starting Down()")
+
+	if err := d.verifyFlags(); err != nil {
+		return fmt.Errorf("down could not verify flags: %s", err)
+	}
+
+	env := d.buildEnv()
```
amwat, line 34: do we want to persist the project that we used in up somewhere?

have you tried running build/up/down separately?

michaelmdresser: I lean towards not persisting the project. The user has to pass the project as a flag to --up anyway and I think avoiding producing artifacts when we don't have to is a good idea.

Running build/up/down separately has been tested and does work.

PR #18005, opened 2020-06-18
Remove travis check from Kubernetes Dashboard

floreks:
Since we are migrating to Github Workflows travis check is no longer needed.

https://github.com/kubernetes/dashboard/pull/5281


floreks:
Can anyone lgtm this?

PR #18008, opened 2020-06-18
lower cpu request in k8s-infra-prow-builds

BenTheElder:
/cc @hasheddan @spiffxp


BenTheElder:
we should also consider modifying the skip list to match the presubmit, but as we're not in code freeze yet (I think?) let's see how just the CPU fix helps

BenTheElder:
this cluster also probably needs a default pod request

PR #18010, opened 2020-06-19
allow GA only conformance job as much time as GCE

BenTheElder:
recent failures appear to be timeouts
the GCE job also appears to sometimes taking >2h just to run the tests so ..

FYI @liggitt

### DIFF
--- a/config/jobs/kubernetes/sig-testing/conformance-e2e.yaml
+++ b/config/jobs/kubernetes/sig-testing/conformance-e2e.yaml
@@ -203,7 +203,7 @@ periodics:
     base_ref: master
     path_alias: k8s.io/kubernetes
   decoration_config:
-    timeout: 2h # allow plenty of time for a serial conformance run
+    timeout: 200m # allow plenty of time for a serial conformance run
   spec:
     containers:
     - image: gcr.io/k8s-testimages/krte:v20200531-b02499e-master


config/jobs/kubernetes/sig-testing/conformance-e2e.yaml
```diff
@@ -203,7 +203,7 @@ periodics:
     base_ref: master
     path_alias: k8s.io/kubernetes
   decoration_config:
-    timeout: 2h # allow plenty of time for a serial conformance run
```
liggitt, line 206: sigh

BenTheElder: I think we don't actually need this much time, but rather than pick an arbitrary timeout I _copied_ one from a different job ;-),

I think we need probably something like 2 hours + time to build / cluster up, which has recently been worse across CI (see #testing-ops, builds are not going well..)

BenTheElder: it is also still mostly passing, but failures are timeouts, and some passing runs are within a minute of timing out. (others a decent bit faster)

BenTheElder:
not sure if the presubmit has this issue too, I think that one was doing parallel tests, right?

liggitt:
Presubmit was parallel and fast

spiffxp:
/lgtm
Yeah I guess

PR #18016, opened 2020-06-19
Enable OOM tracking in periodic 1.16-1.18 perf test jobs

tosi3k:
Roll out the mechanism implemented in kubernetes/perf-tests#1309 to periodic 1.16-1.18 performance tests.

/sig scalability
/cc jkaniuk
/cc mm4tt
/cc wojtek-t


config/jobs/kubernetes/sig-release/release-branch-jobs/1.16.yaml
```diff
@@ -257,6 +258,7 @@ periodics:
       - --test-cmd-args=--testconfig=testing/load/config.yaml
       - --test-cmd-args=--testconfig=testing/access-tokens/config.yaml
       - --test-cmd-args=--testoverrides=./testing/load/kubemark/500_nodes/override.yaml
+      - --test-cmd-args=--testoverrides=./testing/overrides/enable_oom_tracking.yaml
```
wojtek-t, line 261: What this measurement does in kubemark? Are we really sure it does what we want? if it is watching OOM events from the kubemark cluster it's useless, because those will never exist (NPD is faked)...

PR #18017, opened 2020-06-19
Enable OOM tracking in periodic master perf test jobs

tosi3k:
Roll out the mechanism implemented in kubernetes/perf-tests#1309 to the rest of periodic master performance tests.

/sig scalability
/cc jkaniuk
/cc mm4tt
/cc wojtek-t

### DIFF
--- a/config/jobs/kubernetes/sig-scalability/sig-scalability-periodic-jobs.yaml
+++ b/config/jobs/kubernetes/sig-scalability/sig-scalability-periodic-jobs.yaml
@@ -117,6 +117,7 @@ periodics:
       - --kubemark
       - --kubemark-nodes=100
       - --provider=gce
+      - --env=CL2_ENABLE_CLUSTER_OOMS_TRACKER=true
       - --test=false
       - --test_args=--ginkgo.focus=xxxx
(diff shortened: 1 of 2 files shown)


config/jobs/kubernetes/sig-scalability/sig-scalability-periodic-jobs.yaml
```diff
@@ -117,6 +117,7 @@ periodics:
       - --kubemark
       - --kubemark-nodes=100
       - --provider=gce
+      - --env=CL2_ENABLE_CLUSTER_OOMS_TRACKER=true
```
wojtek-t, line 120: What does OOM tracker do in kubemark? Does it really do something useful?

IIUC, this is watching events from kubemark cluster, where NPD is faked, so it's baiscally no-op.

jkaniuk: It still tracks components on the master.

wojtek-t: ok - good point.
That said, we should probably make that explicit somewhere (measurement doc?) [that it only works for master components and we're fine with it.]

PR #18018, opened 2020-06-19
option to use iptables-nft for kind jobs

aojea:
iptables-nft has the advantage that is namespaced and does not need an external lock, that may solve a common problem for KIND jobs, were kube-proxy fails to acquire the lock and can't create the corresponding rules.


aojea:
/assign @BenTheElder 
do you think this will be enough?

PR #18020, opened 2020-06-19
Implemented DumpClusterLogs() in GCE deployer for cloud-provider-gcp

michaelmdresser:
Like the initial Up() and Down() commits, this only supports the k/cloud-provider-gcp repo. Handles checking and creation of logs directory and using kubectl and log-dump.sh to get logs. Opted to use the repo/cluster/kubectl.sh script instead of just calling the local kubectl for consistency. Had to add one environment var `$USER` for the reasons explained in the code comment.

Because Kubetest2 core does not currently support a flag to enable/disable cluster log dumps, it will currently run at the end of every up.

Tested against fresh GCP projects by calling up regularly and then manually inspecting the created logs directory.

I would like to specifically call attention to `makeLogsDir()`. Should it automatically remove the logs dir if at already exists? I think for most normal use it does not matter, as CI should be running in a fresh filesystem, but it could impact manual users.

### DIFF
--- a/kubetest2/kubetest2-gce/deployer/dumplogs.go
+++ b/kubetest2/kubetest2-gce/deployer/dumplogs.go
@@ -55,6 +55,17 @@ func (d *deployer) makeLogsDir() error {
 	_, err := os.Stat(d.logsDir)
 
 	if err == nil {
+		if d.OverwriteLogsDir {
+			if err := os.RemoveAll(d.logsDir); err != nil {
+				return fmt.Errorf("failed to delete existing logs directory: %s", err)
+			}
+
+			err := os.Mkdir(d.logsDir, os.ModePerm)
+			if err != nil {
+				return fmt.Errorf("failed to create %s: %s", d.logsDir, err)
+			}
+			return nil
+		}
 		return fmt.Errorf("cluster logs directory %s already exists, please clean up manually before continuing", d.logsDir)
 	}
 	if os.IsNotExist(err) {
--- a/kubetest2/kubetest2-gce/deployer/dumplogs.go
+++ b/kubetest2/kubetest2-gce/deployer/dumplogs.go
@@ -1,0 +1,97 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package deployer
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"k8s.io/klog"
+	"k8s.io/test-infra/kubetest2/pkg/exec"
+)
+
+func (d *deployer) DumpClusterLogs() error {
+	klog.Info("GCE deployer starting DumpClusterLogs()")
+
+	if err := d.verifyFlags(); err != nil {
+		return fmt.Errorf("dump cluster logs could not verify flags: %s", err)
+	}
+
+	if err := d.makeLogsDir(); err != nil {
+		return fmt.Errorf("couldn't make logs dir: %s", err)
+	}
+
+	env := d.buildEnv()
+	outfile, err := os.Create(filepath.Join(d.logsDir, "cluster-info.log"))
+	if err != nil {
+		return fmt.Errorf("failed to create cluster-info log file: %s", err)
+	}
+	defer outfile.Close()
+
+	command := []string{
+		d.kubectl,
+		"cluster-info",
+		"dump",
+	}
+	klog.Infof("About to run: %s", command)
+
+	cmd := exec.Command(command[0], command[1:]...)
+	cmd.SetEnv(env...)
+	exec.InheritOutput(cmd)
+	cmd.SetStderr(os.Stderr)
+	cmd.SetStdout(outfile)
+	err = cmd.Run()
+	if err != nil {
+		return fmt.Errorf("couldn't use kubectl to dump cluster info: %s", err)
+	}
+
+	env = append(env, "KUBE_GCE_INSTANCE_PREFIX=kubernetes")
+
+	command = []string{
+		filepath.Join(d.RepoRoot, "cluster", "log-dump", "log-dump.sh"),
+		d.logsDir,
+	}
+	klog.Infof("About to run: %s", command)
+
+	cmd = exec.Command(command[0], command[1:]...)
+	cmd.SetEnv(env...)
+	exec.InheritOutput(cmd)
+	err = cmd.Run()
+	if err != nil {
+		return fmt.Errorf("failed to use log-dump.sh for cluster logs: %s", err)
+	}
+
+	return nil
+}
+
+func (d *deployer) makeLogsDir() error {
+	_, err := os.Stat(d.logsDir)
+
+	if err == nil {
+		// TODO: removeall instead?
+		return fmt.Errorf("cluster logs directory %s already exists, please clean up manually before continuing", d.logsDir)
+	} else if os.IsNotExist(err) {
+		err := os.Mkdir(d.logsDir, os.ModePerm)
+		if err != nil {
+			return fmt.Errorf("failed to create %s: %s", d.logsDir, err)
+		}
+		return nil
+	}
+
+	return fmt.Errorf("unexpected exception when making cluster logs directory: %s", err)
+}
--- a/kubetest2/kubetest2-gce/deployer/dumplogs.go
+++ b/kubetest2/kubetest2-gce/deployer/dumplogs.go
@@ -36,44 +36,16 @@ func (d *deployer) DumpClusterLogs() error {
 		return fmt.Errorf("couldn't make logs dir: %s", err)
 	}
 
-	env := d.buildEnv()
-	outfile, err := os.Create(filepath.Join(d.logsDir, "cluster-info.log"))
-	if err != nil {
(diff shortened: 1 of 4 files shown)


kubetest2/kubetest2-gce/deployer/up.go
```diff
@@ -79,6 +91,8 @@ func (d *deployer) verifyFlags() error {
 		return err
 	}
 
+	d.kubectl = filepath.Join(d.RepoRoot, "cluster", "kubectl.sh")
```
amwat, line 94: perhaps we should move all the common logic, like buildEnv and deployer initialization behind a sync.Once initialization so that we don't need to do this in every step.

https://golang.org/pkg/sync/#Once

kubetest2/kubetest2-gce/deployer/dumplogs.go
```diff
@@ -0,0 +1,117 @@
...
+	args := []string{
+		d.kubectl,
+		"cluster-info",
+		"dump",
+	}
+	klog.Infof("About to run: %s", args)
+
+	cmd := exec.Command(args[0], args[1:]...)
+	cmd.SetEnv(env...)
+	exec.InheritOutput(cmd)
```
amwat, line 108: remove?

kubetest2/kubetest2-gce/deployer/dumplogs.go
```diff
@@ -0,0 +1,117 @@
...
+		}
+		return nil
+	}
+
+	return fmt.Errorf("unexpected exception when making cluster logs directory: %s", err)
+}
+
+func (d *deployer) sshDump() error {
+	env := d.buildEnv()
+	env = append(env, "KUBE_GCE_INSTANCE_PREFIX=kubernetes")
```
amwat, line 72: clarifying comment would be helpful here.
also kubernetes seems to be the default, do we want to set something up specially for kubetest2? that our Up() will do.

kubetest2/kubetest2-gce/deployer/dumplogs.go
```diff
@@ -0,0 +1,117 @@
...
+
+	return nil
+}
+
+func (d *deployer) makeLogsDir() error {
+	_, err := os.Stat(d.logsDir)
+
+	if err == nil {
+		return fmt.Errorf("cluster logs directory %s already exists, please clean up manually before continuing", d.logsDir)
+	} else if os.IsNotExist(err) {
```
amwat, line 59: redundant else

kubetest2/kubetest2-gce/deployer/dumplogs.go
```diff
@@ -0,0 +1,117 @@
...
+		filepath.Join(d.RepoRoot, "cluster", "log-dump", "log-dump.sh"),
+		d.logsDir,
+	}
+	klog.Infof("About to run: %s", args)
+
+	cmd := exec.Command(args[0], args[1:]...)
+	cmd.SetEnv(env...)
+	exec.InheritOutput(cmd)
+	err := cmd.Run()
+	if err != nil {
```
amwat, line 84:
```suggestion
	if err := cmd.Run(); err != nil {
```

kubetest2/kubetest2-gce/deployer/dumplogs.go
michaelmdresser: I don't think this is redundant? os can return more than just isnotexist as an error, which should exhibit different behavior from the program

kubetest2/kubetest2-gce/deployer/up.go
michaelmdresser: Is it worth the code complexity? Implicitly, things like buildEnv and verifyFlags should have no effect if run more than once vs just once (I suppose this is idempotency).

kubetest2/kubetest2-gce/deployer/dumplogs.go
michaelmdresser: Done.

kubetest2/kubetest2-gce/deployer/dumplogs.go
michaelmdresser: Set in buildEnv with comment. Now it will be consistent across all commands. Explained in comment.

kubetest2/kubetest2-gce/deployer/dumplogs.go
michaelmdresser: Thanks for catching that. Done.

kubetest2/kubetest2-gce/deployer/dumplogs.go
amwat: I meant, the "else" is redundant :) since the if is returning.
Dedent for maximum simplicity.

kubetest2/kubetest2-gce/deployer/up.go
amwat: Yes, but there's no reason to run them multiple times.
`sync.Once` can be used pretty easily like so:
https://play.golang.org/p/1J9ghm_eI9t

shouldn't really add that much complexity.

kubetest2/kubetest2-gce/deployer/dumplogs.go
michaelmdresser: Oh, perfect. Done.

kubetest2/kubetest2-gce/deployer/up.go
michaelmdresser: I guess I'm still not sure what is gained by using `sync.Once`, to me it seems to just add lines of code and complexity for a reader.

I don't see a potential situation where guaranteeing "run at most once" for any of the code is necessary. I doubt this code will ever be run concurrently in Go and therefore need to deal with races. The only situation (that I can see) in which we would gain some "safety" is the [completely pathological](https://unix.stackexchange.com/questions/38205/change-environment-of-a-running-process) case of a user editing the process's environment variables during its run.

If there's an argument to be made for performance, I don't think it matters because runtime is overwhelmingly dominated by kube-up.sh and company.

amwat: Well, `buildEnv()` might be simple and innocuous but stuff like `verifyFlags()` and executing `kubectl.sh` in each step seems pretty unnecessary to me when we know it's explicitly a one-time thing. 
I guess it comes down to whether or not you think `d.initialize.Do(func() { err = d.init() })` is adding significant readability complexity as compared to `err = d.init()` .  I tend towards no, but I'll leave it to you.

@rancerbeta thoughts?

/lgtm
/hold

kubetest2/kubetest2-gce/deployer/dumplogs.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	return nil
+}
+
+func (d *deployer) makeLogsDir() error {
+	_, err := os.Stat(d.logsDir)
+
+	if err == nil {
+		return fmt.Errorf("cluster logs directory %s already exists, please clean up manually before continuing", d.logsDir)
```
rancerbeta, line 58: this seems like it could get annoying for rerunning tests. You would need to delete the logs directory between test runs, is that right? 

If so, maybe you can add a flag to have it overwrite to make it easier

kubetest2/kubetest2-gce/deployer/dumplogs.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+	if os.IsNotExist(err) {
+		err := os.Mkdir(d.logsDir, os.ModePerm)
+		if err != nil {
+			return fmt.Errorf("failed to create %s: %s", d.logsDir, err)
+		}
+		return nil
+	}
+
+	return fmt.Errorf("unexpected exception when making cluster logs directory: %s", err)
```
rancerbeta, line 68: this is kind of a strange case, but the style i've seen usually wants the success case to be on the main indent. so this would be return nil instead and line 60 would get negated

kubetest2/kubetest2-gce/deployer/up.go
rancerbeta: It doesn't add too much more complexity. I think it would be a good idea to capture all your initialization at the front if possible. There's no saying what a test author may do in between each and removing that possibility is a good thing. Users are not always being logical.

kubetest2/kubetest2-gce/deployer/dumplogs.go
michaelmdresser: Done. Thanks for the suggestion, I hadn't thought of that.

kubetest2/kubetest2-gce/deployer/dumplogs.go
michaelmdresser: Switched the style of this and one other location where I had this pattern.

kubetest2/kubetest2-gce/deployer/up.go
michaelmdresser: Here is a proposal for using `sync.Once`: https://github.com/michaelmdresser/test-infra/pull/1

I opted to remove the build-specific flag verification. I could bring it back and use another `sync.Once` for it, I'm definitely interested in discussing this choice.

I didn't put `buildEnv` behind a `sync.Once` because (please, correct me if I'm wrong) the only behavior that would make sense if it was behind a `sync.Once` is to build the environment and then store it as a slice in the deployer struct. However, because it is a slice, every time a script-caller needs to modify it, that caller has to copy the slice rather than just modify it. This seems like even more of a "relying on the future programmer" than what it is now.

I suppose `buildEnv` could also modify the os environment, but I think that is even less desirable because it is more opaque.

rancerbeta: on the sync.once thing, buildenv and verifyflags look like functions that should occur once per kubetest2 CLI invocation.  it might be sufficient to check the existence of a verifiedenv/verifiedflags variable without needing to do sync.once

afaik each of the stages is never run in parallel with each other. wdyt amit?

michaelmdresser: After talking with Amit, I think I will implement this behavior alongside saving intermediate state of the deployer struct in the artifacts directory. I think this makes the most sense because that will introduce a need for a generic initialization step that should only be performed once.

rancerbeta:
lgtm

PR #18021, opened 2020-06-19
Implemented IsUp() in GCE deployer

michaelmdresser:
Uses the same simple concept from other deployers: kubectl reporting nodes means the cluster is up.

Because kubetest2 core does not specify a flag for checking "is up" this method is called at the end of Up() for initial usage until kubetest2 core implements.

Tested with a regular run of up on a fresh GCP project.

### DIFF
--- a/kubetest2/kubetest2-gce/deployer/deployer.go
+++ b/kubetest2/kubetest2-gce/deployer/deployer.go
@@ -23,6 +23,7 @@ import (
 	"path/filepath"
 
 	"k8s.io/klog"
+	"k8s.io/test-infra/kubetest2/pkg/exec"
 	"k8s.io/test-infra/kubetest2/pkg/types"
 
 	"github.com/octago/sflags/gen/gpflag"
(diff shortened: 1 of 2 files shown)


kubetest2/kubetest2-gce/deployer/up.go
```diff
@@ -79,6 +88,8 @@ func (d *deployer) verifyFlags() error {
 		return err
 	}
 
+	d.kubectl = filepath.Join(d.RepoRoot, "cluster", "kubectl.sh")
```
amwat, line 91: should we fallback to kubectl in PATH ?

michaelmdresser: I don't know. I opted for this because almost every script I've seen that needs to use kubectl uses this script in the repo (or at least defaults to it without a manual override).

kubetest2/kubetest2-gce/deployer/deployer.go
```diff
@@ -67,7 +69,32 @@ func (d *deployer) Provider() string {
...
-func (d *deployer) IsUp() (up bool, err error) { return false, nil }
+func (d *deployer) IsUp() (up bool, err error) {
+	klog.Info("GCE deployer starting IsUp()")
+
+	if err := d.verifyFlags(); err != nil {
+		return false, fmt.Errorf("is up failed to verify flags: %s", err)
+	}
+
+	env := d.buildEnv()
+	// naive assumption: nodes reported = cluster up
```
rancerbeta, line 80: interesting, they don't check status?

michaelmdresser: Yep, both the kind and gke deployers do almost exactly this block of code.

rancerbeta:
lgtm

PR #18022, opened 2020-06-19
images/gcb-docker-gcloud: Update version of docker-buildx

detiber:


detiber:
/assign @stevekuznetsov 
ptal

Issue #17922, opened 2020-06-11
Allow Plank To Create Pods In Another Namespace


midnightconman:
**What would you like to be added**:
I would like the ability to create pods in another namespace using plank. It seems fejta called this item out here (https://github.com/kubernetes/test-infra/blob/master/prow/config/config.go#L1772). It would also be nice to configure the namespace that pods are created via the kubeconfig context, so we can control the default test pod namespace via the `clusters` parameter.

**Why is this needed**:
Without this, you can't have test-pods and trusted-test-pods in the same kubernetes cluster.

alvaroaleman:
This feature already exists, you can specify a namespace on the job.

> It would also be nice to configure the namespace that pods are created via the kubeconfig context, so we can control the default test pod namespace via the clusters parameter.

That would be a pretty breaking change and I am not sure I like the idea. Maybe as a default for the global setting if unset, but should definitely never overwrite any other namespace setting.

/area prow

midnightconman:
When was it implemented, I am getting the error I linked to from
check-config?

alvaroaleman:
Which error? Not sure when this was implemented but before I was with the project.
You should be able to just set a `namespace: your-namespace` property on jobs.

midnightconman:
The one I linked to above 😄 

It says: `namespace customization requires agent: pipeline (found kubernetes)`... It also has a `TODO` item from fejta saying the we should fix this. The new line for this TODO item is (https://github.com/kubernetes/test-infra/blob/master/prow/config/config.go#L1775) and it says it would require a plank client change.

I'll take a stab at this in a PR and reference it here.

alvaroaleman:
ah okay, sorry I missed that. Should be reasonably easy.

midnightconman:
No problem and thank you... I am taking a look at plank now.

midnightconman:
Hmm, I am wondering the best way to tackle this... I think we should switch plank's rbac from a `role` to a `clusterRole` and switch this to selecting all pods with the correct selector across all namespaces. The code change would be [here](https://github.com/kubernetes/test-infra/blob/master/prow/plank/controller.go#L180) and [here](https://github.com/kubernetes/test-infra/blob/master/prow/plank/controller.go#L196).

I am not sure how this change would be made, I would think this would need to be planned out? Let me know how to proceed 😸 , or if the plan is crazy 😉 .

midnightconman:
[...]

PR #18028, opened 2020-06-20
Teach Plank V1 To Create Pods In Multiple Namespaces

midnightconman:
fixes: #17922 

This PR allows plank v1 to create pods in any namespace for the kubernetes agent, if a prowJob has overridden the config value `PodNamespace`. Plank v2 was giving me much more trouble, so I hope to be able to do that change in a followup PR.

### DIFF
--- a/config/prow/cluster/plank_rbac.yaml
+++ b/config/prow/cluster/plank_rbac.yaml
@@ -22,10 +22,9 @@ rules:
       - update
       - patch
 ---
-kind: Role
+kind: ClusterRole
 apiVersion: rbac.authorization.k8s.io/v1beta1
 metadata:
-  namespace: test-pods
   name: "plank"
 rules:
   - apiGroups:
@@ -50,10 +49,9 @@ subjects:
 - kind: ServiceAccount
   name: "plank"
 ---
-kind: RoleBinding
+kind: ClusterRoleBinding
 apiVersion: rbac.authorization.k8s.io/v1beta1
 metadata:
-  namespace: test-pods
   name: "plank"
 roleRef:
   apiGroup: rbac.authorization.k8s.io
--- a/prow/plank/controller.go
+++ b/prow/plank/controller.go
@@ -193,7 +193,7 @@ func (c *Controller) Sync() error {
 	pm := map[string]corev1.Pod{}
 	for alias, client := range c.buildClients {
 		listOpts := &ctrlruntimeclient.ListOptions{
-			Namespace: c.config().PodNamespace,
+			Namespace: "",
 			Raw:       &metav1.ListOptions{LabelSelector: selector},
 		}
 		pods := &corev1.PodList{}
@@ -529,7 +529,13 @@ func (c *Controller) startPod(pj *prowapi.ProwJob) error {
 	if err != nil {
 		return err
 	}
-	pod.Namespace = c.config().PodNamespace
+
+	// If the prowJob has a namespace specified use that value, else use the config prowJob namespace.
+	if pj.Spec.Namespace != "" {
+		pod.Namespace = pj.Spec.Namespace
+	} else {
+		pod.Namespace = c.config().PodNamespace
+	}
 
 	client, ok := c.buildClients[pj.ClusterAlias()]
 	if !ok {
--- a/prow/config/config.go
+++ b/prow/config/config.go
@@ -1771,8 +1771,7 @@ func validateAgent(v JobBase, podNamespace string) error {
 		return fmt.Errorf("error_on_eviction only applies to agent: %s (found %q)", k, agent)
 	case v.Namespace == nil || *v.Namespace == "":
 		return fmt.Errorf("failed to default namespace")
-	case *v.Namespace != podNamespace && agent != p:
-		// TODO(fejta): update plank to allow this (depends on client change)
+	case *v.Namespace != podNamespace && (agent != p || agent != k):
 		return fmt.Errorf("namespace customization requires agent: %s (found %q)", p, agent)
 	}
 	return nil


prow/plank/controller.go
```diff
@@ -529,7 +529,13 @@ func (c *Controller) startPod(pj *prowapi.ProwJob) error {
 	if err != nil {
 		return err
 	}
-	pod.Namespace = c.config().PodNamespace
+
+	// If the prowJob has a namespace specified use that value, else use the config prowJob namespace.
+	if pj.Spec.Namespace != "" {
```
alvaroaleman, line 534: IMHO we should do this during config load and not in individual components

config/prow/cluster/plank_rbac.yaml
```diff
@@ -22,10 +22,9 @@ rules:
       - update
```
alvaroaleman, line 22: This is a pretty incompatible change and will break all existing deployments. In order to actually leverage this you need to at least update `prow/cmd/plank/main.go` to not watch one but all namespaces. I believe sinker will need updating as well.

And we have to think about some kind of strategy that does not involve breaking existing deployments. We could for example in the main test if we can do a `List` with `MaxResult: 1` in a namesapce that is guaranteed to exist but not the `ProwJob` namespace, like for example `kube-system`. If we get a 403 we only use the `ProwJob` namespace and log a warning, otherwise we use all namespaces.

The unfortunate side effect is that if someone creates a job with a different namespace but uses the namespace-scoped rbac, it will silently not work because `plank` doesn't see the job.

midnightconman: Everything is better with flags?

- We could expose a flag or config parameter that would allow us to set the selector namespace... Defaulting to podnamespace?
- Or we could have a flag or config parameter called something like "allow_namespace_override"? This new config option would then require the cluster rbac.

How does that sound?

prow/plank/controller.go
midnightconman: Works for me... Let me try to move this to config.

config/prow/cluster/plank_rbac.yaml
alvaroaleman: tbh if possible I would much prefer if we not added even more configuration options :grimacing: We have way too many already

midnightconman: Two things on this 😸 

Start-up probes in kubernetes land are normally used for things that `don't change at runtime` and that are normally `environmental capabilities` (like cloud-provider capabilities or kubernetes api server api availability). RBAC in kubernetes is detached from runtime items, so we would be attaching a dynamic non-runtime item to a runtime probing technique (not ideal and would probably cause confusion). Also, we would need to restart services to change RBAC (not normally done in kubernetes land, since RBAC is decoupled from runtime).

I think if stable eco-system patterns exist for certain approaches, we should use those, that's what people are expecting... anything else will probably confuse people in the beginning and frustrate them later (when they find out you deviated from normal practices).

Thinking through this though... we can "have both". What if we leave the `namespaced` role for `podNamespace` and add another clusterRole? That should allow a more gradual change workflow? We would still need someway of changing the selector namespace (I would go with flag or config param).

alvaroaleman: We can not have both because we use a watch and we can do that either for all namespaces or for one namespace.

I am mainly opposed to introduce yet another config switch. Either we decide that giving perms for all namespaces is the right thing or we don't.

The probing for perms is done to be able to give Prow users some time to migrate and warn them that they need to do something. Compatibility is an annoying thing that does not always allow you to follow "best practices" but its also extremely important if you are building software that is being used.

midnightconman: > I am mainly opposed to introduce yet another config switch. Either we decide that giving perms for all namespaces is the right thing or we don't.

I know k8s prow currently uses a `trusted` k8s cluster for trusted workloads, do you guys have any desire to shard trust by namespace? Or would you prefer to create a whole k8s cluster for each trust shard?

Do we need a KEP for this? Should we chat more in slack? How do we move forward on this?

midnightconman: If we use a watch for pods, we need to update this manifest... it only shows `list` permissions.

alvaroaleman: Right, we don't do it right now in plank v1 but we do it in plank v2 and the former will eventually get removed (and we should not add features to one but not the others).

Yeah I am not completely sure about the best approach. We could also make the pod namespace a part of the build cluster (with default to the global default pod namespace) and then you could have a virtual cluster `mybuildcluster-other-namespace`.

And for the record, this is not something I can just decide/approve alone. Maybe create an issue first and ping erick, steve and me and bring it up in the biweekly-meeting?

midnightconman: > Right, we don't do it right now in plank v1 but we do it in plank v2 and the former will eventually get removed (and we should not add features to one but not the others).

I wasn't sure how far along plank v2 had come... should we possibly start with that one instead? It was much easier to make this change to v1, I had strange errors in v2 (I assume it's due to the informer pattern).

> Yeah I am not completely sure about the best approach. We could also make the pod namespace a part of the build cluster (with default to the global default pod namespace) and then you could have a virtual cluster mybuildcluster-other-namespace.

We spoke about this too... this was originally how it seemed something like this should be done, but I ran into the hardcoded `podNamespace` and then saw the prowJob namespace parameter, and then we arrived here :D You guys let me know what you think would be the best starting point and direction.

> And for the record, this is not something I can just decide/approve alone. Maybe create an issue first and ping erick, steve and me and bring it up in the biweekly-meeting?

No problem... we already have an issue for this (mentioned via `fixes` in the PR comment), I'll ping steve and fejta on it.

Thank you for the help, it is much appreciated!

alvaroaleman: > I wasn't sure how far along plank v2 had come... should we possibly start with that one instead? 

Yeah I think that would be better. It definitely isn't totally broken and supposed to replace plank entirely in the future, we replaced Plank with it on june fifth for openshift. If you have problems with it, please open an issue :)

PR #18030, opened 2020-06-20
Ability to dump logs in node e2e jobs

dims:
When pull-kubernetes-node-e2e-containerd fails, we don't capture logs at all:
https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/92333/pull-kubernetes-node-e2e-containerd/1274403672662478848/build-log.txt

```
W0620 19:15:54.443] 2020/06/20 19:15:54 node.go:42: Noop - Node DumpClusterLogs() - /workspace/_artifacts: 
```

### DIFF
--- a/kubetest/node.go
+++ b/kubetest/node.go
@@ -39,8 +39,7 @@ func (n nodeDeploy) IsUp() error {
 }
 
 func (n nodeDeploy) DumpClusterLogs(localPath, gcsPath string) error {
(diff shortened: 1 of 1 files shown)


spiffxp:
/approve
/lgtm
Will this actually work? I'm approving to find out, but I'm not 100% familiar with how much of the existing kube-up machinery is used for node-e2e vs. something separate

dims:
@spiffxp fully expecting to do follow ups :)

Issue #17915, opened 2020-06-11
Consistent tide squash commit message when there's only one commit


Bobgy:
**Current behavior**:
When tide is configured to `squash` by default.
If there are more than one commits in the PR, commits will be squashed and PR title will be used as the commit message.
However, if there's only one commit in the PR, e.g. https://github.com/kubeflow/pipelines/commit/7fbb7cae567d24e766ae851cc494b8db0f239e54 the commit will be untouched when merged.

**What would you like to be added**:
Even if there's only one commit in the PR, it should use PR title for the commit message.

**Why is this needed**:
To enforce consistent commit messages, so that they always come from PR title.

I'm proposing this because devs usually write arbitrary commit messages and rely on the auto squash feature to squash the commit as PR title. With current behavior, some commits will unexpectedly come from the commit message directly.

Also, I'm thinking about enforcing commit message using https://www.conventionalcommits.org/en/v1.0.0/ for kubeflow/pipelines repo, so that we can generate release notes with more metadata https://github.com/kubeflow/pipelines/issues/3920. This issue makes commit message somewhat unstable.

What do you think?

Bobgy:
Hi @cjwagner, not sure if you are the right person. From similar issues, it seems you may know related context.

Would you mind taking a look? Thanks!

cjwagner:
By default, Tide only specifies the requested merge type (merge/squash/rebase) when merging a PR and GitHub determines the commit title and body itself. This can be overridden with the `merge_commit_template` config field if you need more control over the commit text.
https://github.com/kubernetes/test-infra/blob/master/prow/cmd/tide/config.md#general-configuration
https://godoc.org/k8s.io/test-infra/prow/config#Tide

Bobgy:
Thanks @cjwagner! That answered my question perfectly.

PR #18043, opened 2020-06-22
kubeflow/pipelines: use merge_commit_template for stable commit message

Bobgy:
Part of https://github.com/kubeflow/pipelines/issues/3920

Use merge_commit_template config from https://github.com/kubernetes/test-infra/blob/master/prow/cmd/tide/config.md#general-configuration to configure a stable commit message for kubeflow/pipelines repo as a solution for https://github.com/kubernetes/test-infra/issues/17915.

### DIFF
--- a/config/prow/config.yaml
+++ b/config/prow/config.yaml
@@ -515,6 +515,9 @@ tide:
     kubernetes-sigs/minibroker: squash
     kubernetes/cloud-provider-openstack: squash
     kubernetes/dashboard: squash
+  merge_commit_template:
+    kubeflow/pipelines:
+      title: "{{ .Title }} ({{ .Number }})"
   pr_status_base_urls:
     '*': https://prow.k8s.io/pr
   blocker_label: tide/merge-blocker


Bobgy:
/assign @wojtek-t
Can you help me approve?
I'm one of the OWNERS for kubeflow/pipelines https://github.com/kubeflow/pipelines/blob/master/OWNERS, currently working on auto changelog generation that relies on commit message.

config/prow/config.yaml
```diff
@@ -515,6 +515,9 @@ tide:
     kubernetes-sigs/minibroker: squash
     kubernetes/cloud-provider-openstack: squash
     kubernetes/dashboard: squash
+  merge_commit_template:
```
jlewi, line 518: Should we do this consistently for all repos?

Bobgy: https://github.com/kubernetes/test-infra/issues/17915 has all the context why kubeflow/pipelines want this.
but I think it also has the side effect of dropping commit body all together (github default behavior is concatenating all commits' messages and bodies). It's fine for KFP, we don't care about commit body anyway.

If you think that seems reasonable for other repos too, I have no objection, but maybe it's better we first try this configuration in KFP to see if it's working as expected, (because looks like although the feature was implemented, no one was using it before).

Bobgy:
@spiffxp Can I ask for approval? Thanks!

PR #17966, opened 2020-06-17
Correctly set kubeconfig for the hmac tool

chizhg:
Fix the flag for the hmac tool to correctly set the kubeconfig.

/assign @cjwagner

PR #18046, opened 2020-06-23
Configure hmac management postsubmit job for k8s Prow

chizhg:
Configure hmac management postsubmit job for k8s Prow, which can reconcile the hmac tokens and webhooks based on the `managed_webhooks` setting in the Prow core config file.
And by the way fix two code review issues left in https://github.com/kubernetes/test-infra/pull/17966.

There are two things need to be done before getting this merged:
1. Back up the global HMAC token we currently have for k8s Prow, in case any issue happens
2. Make sure the github oauth token we use here has `admin:repo_hook` and `admin:org_hook` scopes.

/cc @cjwagner @michelle192837


chizhg:
/hold cancel

Thanks @michelle192837 !

PR #18052, opened 2020-06-23
Fix experimental containerd correctnes test definition

jprzychodzen:
/assign @mm4tt


config/jobs/kubernetes/sig-scalability/sig-scalability-experimental-periodic-jobs.yaml
```diff
@@ -220,7 +220,6 @@ periodics:
           - --gcp-node-image=gci
           - --gcp-nodes=100
           - --gcp-project-type=scalability-project
-          - --gcp-ssh-proxy-instance-name=gce-cluster-master
```
mm4tt, line 223: We should set it to the name of the master, otherwise the ssh based tests will start failing

jprzychodzen: Thanks for explanation, updated.

PR #18053, opened 2020-06-23
Add the ability to pass commit messages in the git merge operations

droslean:
Add the ability to pass commit messages in the `git merge` operations.

The change affects only the merge strategy.

### DIFF
--- a/prow/git/v2/interactor.go
+++ b/prow/git/v2/interactor.go
@@ -45,7 +45,7 @@ type Interactor interface {
 	// Merge merges the commitlike into the current HEAD
 	Merge(commitlike string) (bool, error)
 	// MergeWithStrategy merges the commitlike into the current HEAD with the strategy
-	MergeWithStrategy(commitlike, mergeStrategy string) (bool, error)
+	MergeWithStrategy(commitlike, mergeStrategy string, opts ...MergeOpt) (bool, error)
 	// MergeAndCheckout merges all commitlikes into the current HEAD with the appropriate strategy
 	MergeAndCheckout(baseSHA string, mergeStrategy string, headSHAs ...string) error
 	// Am calls `git am`
@@ -78,6 +78,21 @@ type cacher interface {
 type cloner interface {
 	// Clone clones the repository from a local path.
 	Clone(from string) error
+}
+
+// MergeOpt holds options for git merge operations.
+// Currently only commit message option is supported.
+type MergeOpt struct {
+	CommitMessage string
+}
+
+// trimMessage surrounds the message with double quotes if is a multi word message
+func trimMessage(message string) string {
+	// In case of a message with multiple words, put it into quotes
+	if len(strings.Split(message, " ")) > 1 {
+		return fmt.Sprintf("\"%s\"", message)
+	}
+	return message
 }
 
 type interactor struct {
@@ -164,11 +179,11 @@ func (i *interactor) Merge(commitlike string) (bool, error) {
 // MergeWithStrategy attempts to merge commitlike into the current branch given the merge strategy.
 // It returns true if the merge completes. if the merge does not complete successfully, we try to
 // abort it and return an error if the abort fails.
-func (i *interactor) MergeWithStrategy(commitlike, mergeStrategy string) (bool, error) {
+func (i *interactor) MergeWithStrategy(commitlike, mergeStrategy string, opts ...MergeOpt) (bool, error) {
 	i.logger.Infof("Merging %q using the %q strategy", commitlike, mergeStrategy)
 	switch mergeStrategy {
 	case "merge":
-		return i.mergeMerge(commitlike)
+		return i.mergeMerge(commitlike, opts...)
 	case "squash":
 		return i.squashMerge(commitlike)
 	default:
@@ -176,8 +191,20 @@ func (i *interactor) MergeWithStrategy(commitlike, mergeStrategy string) (bool, 
 	}
 }
 
-func (i *interactor) mergeMerge(commitlike string) (bool, error) {
-	out, err := i.executor.Run("merge", "--no-ff", "--no-stat", "-m", "merge", commitlike)
+func (i *interactor) mergeMerge(commitlike string, opts ...MergeOpt) (bool, error) {
+	args := []string{"merge", "--no-ff", "--no-stat"}
+
+	if len(opts) == 0 {
+		args = append(args, []string{"-m", "merge"}...)
+	} else {
+		for _, opt := range opts {
+			args = append(args, []string{"-m", trimMessage(opt.CommitMessage)}...)
+		}
+	}
+
+	args = append(args, commitlike)
+
+	out, err := i.executor.Run(args...)
 	if err == nil {
 		return true, nil
 	}
--- a/prow/git/v2/adapter.go
+++ b/prow/git/v2/adapter.go
@@ -62,7 +62,7 @@ func (a *repoClientAdapter) MergeAndCheckout(baseSHA string, mergeStrategy strin
 	return a.Repo.MergeAndCheckout(baseSHA, github.PullRequestMergeType(mergeStrategy), headSHAs...)
 }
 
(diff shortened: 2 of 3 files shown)


prow/git/git.go
```diff
@@ -295,6 +295,11 @@ func (r *Repo) Merge(commitlike string) (bool, error) {
...
 }
 
+// MergeWithMessage is the same with Merge() but with a message included.
+func (r *Repo) MergeWithMessage(commitlike, message string) (bool, error) {
+	return r.mergeWithMergeStrategyMergeAndMessage(commitlike, message)
+}
+
 // MergeWithStrategy attempts to merge commitlike into the current branch given the merge strategy.
 // It returns true if the merge completes. It returns an error if the abort fails.
 func (r *Repo) MergeWithStrategy(commitlike string, mergeStrategy prowgithub.PullRequestMergeType) (bool, error) {
```
alvaroaleman, line 305: @droslean can we solve this via functional opts pattern?

```
// unexported, we don;t want to allow everything
type extraArg []string

type MergeOpt func() extrArg
func WithMessage(message string) MergeOpt {
  return func() extraArg {
    return []striing{"-m", message"}
  }
}

func (r *Repo) MergeWithStrategy(commitlike string, mergeStrategy prowgithub.PullRequestMergeType, opts ...MergeOpt)
```

Otherwise the number of funcs we need will explode because we need one for each git operation + possible extra arg combination

prow/git/git.go
```diff
@@ -400,7 +426,7 @@ func (r *Repo) Push(branch string) error {
 	if r.user == "" || r.pass == "" {
 		return errors.New("cannot push without credentials - configure your git client")
 	}
-	r.logger.Infof("Pushing to '%s/%s (branch: %s)'.", r.user, r.repo, branch)
+	r.logger.Infof("Pushing to '%s/%s (branch: %s)'.", r.org, r.repo, branch)
```
stevekuznetsov, line 429: This doesn't seem right. Does it break any tests?

prow/git/git.go
```diff
@@ -346,6 +351,27 @@ func (r *Repo) mergeWithMergeStrategySquash(commitlike string) (bool, error) {
 	return true, nil
 }
 
+func (r *Repo) mergeWithMergeStrategyMergeAndMessage(commitlike, message string) (bool, error) {
```
stevekuznetsov, line 354: Shouldn't `mergeWithMergeStrategyMerge` call this?

droslean: To keep backward compability I created a new one.

prow/git/git.go
droslean: I guess the tests assumed that the user is the same with org.

prow/git/git.go
stevekuznetsov: Err... it is an unexported method, so there is no possibility of breaking anything.

prow/git/git.go
```diff
@@ -289,29 +289,46 @@ func (r *Repo) CheckoutNewBranch(branch string) error {
 	return nil
 }
 
+type extraArg []string
+
+// MergeOpt is a type of a function that returns an []string
+type MergeOpt func() extraArg
+
+// WithMessage returns an git message argument with its value
+func WithMessage(message string) MergeOpt {
```
alvaroaleman, line 298: Can we have a test? And why is this not in the v2 client?

prow/git/v2/interactor.go
```diff
@@ -45,7 +46,7 @@ type Interactor interface {
 	// Merge merges the commitlike into the current HEAD
 	Merge(commitlike string) (bool, error)
 	// MergeWithStrategy merges the commitlike into the current HEAD with the strategy
-	MergeWithStrategy(commitlike, mergeStrategy string) (bool, error)
+	MergeWithStrategy(commitlike, mergeStrategy string, opts ...git.MergeOpt) (bool, error)
```
stevekuznetsov, line 49: These should be separate types

prow/git/v2/interactor.go
```diff
@@ -25,6 +25,7 @@ import (
 	"strings"
 
 	"github.com/sirupsen/logrus"
+	"k8s.io/test-infra/prow/git"
```
stevekuznetsov, line 28: This import should not exist

prow/git/v2/interactor.go
```diff
@@ -164,20 +165,31 @@ func (i *interactor) Merge(commitlike string) (bool, error) {
...
 		return false, fmt.Errorf("merge strategy %q is not supported", mergeStrategy)
 	}
 }
 
-func (i *interactor) mergeMerge(commitlike string) (bool, error) {
-	out, err := i.executor.Run("merge", "--no-ff", "--no-stat", "-m", "merge", commitlike)
+func (i *interactor) mergeMerge(commitlike string, opts ...git.MergeOpt) (bool, error) {
+	args := []string{"merge", "--no-ff", "--no-stat"}
+
+	if len(opts) == 0 {
```
stevekuznetsov, line 183: This only works if the user passes a message in their opts

droslean: Before we were hardcoding the message to `-m merge`. I though its best to keep the same thinking. Otherwise we will rely on the user to pass a message, otherwise the command will fail.

prow/git/v2/interactor.go
droslean: Well, it seems that we actually need it since this is a different package.

prow/git/v2/interactor.go
droslean: Since the opts are optional, I don't think that it will affect anything. Do you think that its better to have a `MergeWithStrategyAndOptions` type here?

stevekuznetsov: Yes. v2 should not import v1.

prow/git/v2/interactor.go
stevekuznetsov: If the user does not pass in a message via options in the way you have written this, this fails.

prow/git/v2/interactor.go
```diff
@@ -80,6 +82,33 @@ type cloner interface {
 	Clone(from string) error
 }
 
+// MergeOpt is a type of a function that returns an []string
+type MergeOpt func() []string
+
+func hasMessageOption(opts ...MergeOpt) bool {
```
alvaroaleman, line 88: It would be easier at this point to just make `MergeOpt` a `struct` that has some fields like `CommitMessage` and then defaulting that field if unset and always doing a `-m mergeOpt.CommitMessage`

droslean: The problem here is that we don't always want to default it. The commit message is only in the Merge strategy. However that can be defaulted only in that strategy

alvaroaleman: The other strategies have a commit message as well or not? If its not implemented we should return an error at least. And having a proper struct rather than a string slice and then needing to parse the string slice seems much better to me

droslean: +1 for the struct. It makes also sense to start validating the options, but the command will return error either way because of the wrong options. For example, in this case, if you pass a commit message option in a squash command.

prow/git/v2/interactor.go
```diff
@@ -80,6 +80,20 @@ type cloner interface {
 	Clone(from string) error
 }
 
+// MergeOpt holds information of a commit message
+type MergeOpt struct {
+	CommitMessage string
+}
+
+// TrimMessage surounds the message with double quotes if is a multi word message
+func TrimMessage(message string) string {
```
alvaroaleman, line 89: Why is this exported?

prow/git/v2/interactor.go
```diff
@@ -80,6 +80,20 @@ type cloner interface {
 	Clone(from string) error
 }
 
+// MergeOpt holds information of a commit message
+type MergeOpt struct {
```
alvaroaleman, line 84: No, it holds options for Merges which happens to only be the commit message today

droslean: addressed

prow/git/v2/interactor.go
droslean: addressed

PR #18057, opened 2020-06-23
Retry git fetch commands.

fejta:
The git fetch command waits at most 2s for dns to resolve with no retries.
This can occasionally cause unnecessary flakiness.
Retry these calls with exponential backoff to mitigate

/assign @chaodaiG @michelle192837

### DIFF
--- a/prow/pod-utils/clone/clone.go
+++ b/prow/pod-utils/clone/clone.go
@@ -34,6 +34,10 @@ import (
 	"k8s.io/test-infra/prow/logrusutil"
 )
 
+type runnable interface {
+	run() (string, string, error)
+}
+
 // Run clones the refs under the prescribed directory and optionally
 // configures the git username and email in the repository as well.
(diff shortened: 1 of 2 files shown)


prow/pod-utils/clone/clone.go
```diff
@@ -142,11 +146,32 @@ func (g *gitCtx) gitCommand(args ...string) cloneCommand {
 	return cloneCommand{dir: g.cloneDir, env: g.env, command: "git", args: args}
 }
 
+var (
+	fetchRetries = []time.Duration{
+		100 * time.Millisecond,
+		200 * time.Millisecond,
+		400 * time.Millisecond,
+		800 * time.Millisecond,
+		2 * time.Millisecond,
```
michelle192837, line 155: Is this supposed to be 2 milliseconds, or 2 seconds?

fejta: Done

PR #18060, opened 2020-06-23
Prow autobump script fixing splitting function

chaodaiG:
As title, also make job config as optional since it's already functionally supported. For some reason `split_on_commas` and `add_suffix` don't work on my machine, updated these two to make them work on my machine too

### DIFF
--- a/prow/cmd/autobump/bump.sh
+++ b/prow/cmd/autobump/bump.sh
@@ -198,17 +198,18 @@ list() {
   fi
 }
 
-split_on_commas() {
-  local IFS=,
-  local array=("$1")
-  echo "${array[@]}"
-}
+ split_on_commas() {
+   local IFS=,
+   local array
+   read -ra array <<< "$1"
+   echo "${array[@]}"
+ }
 
-add_suffix() {
-  local array=("$1")
-  local suffix="${2:-/*.yaml}"
-  echo "${array[@]/%/$suffix}"
-}
+ add_suffix() {
+   local array=($1)
+   local suffix="${2:-/*.yaml}"
+   echo "${array[@]/%/$suffix}"
+ }
 
 # See https://misc.flogisoft.com/bash/tip_colors_and_formatting
 color-image() { # Bold magenta
--- a/prow/cmd/autobump/bump.sh
+++ b/prow/cmd/autobump/bump.sh
@@ -68,7 +68,10 @@ main() {
   fi
   echo -e "Bumping: 'gcr.io/k8s-prow/' images to $(color-version "${new_version}") ..." >&2
 
-  bumpfiles=($(add_suffix "$(split_on_commas "$COMPONENT_FILE_DIR")"))
+  local IFS=,
+  local component_file_dir_array
+  read -ra component_file_dir_array <<< "${COMPONENT_FILE_DIR}"
+  bumpfiles=("${component_file_dir_array[@]/%/*.yaml}")
   bumpfiles+=("${CONFIG_PATH}")
   if [[ -n "${JOB_CONFIG_PATH}" ]]; then
     bumpfiles+=($(grep -rl -e "gcr.io/k8s-prow/" "${JOB_CONFIG_PATH}"; true))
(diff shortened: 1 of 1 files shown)


prow/cmd/autobump/bump.sh
```diff
@@ -200,12 +197,13 @@ list() {
...
   local IFS=,
-  local array=("$1")
+  local array
+  read -ra array <<< "$1"
   echo "${array[@]}"
 }
 
 add_suffix() {
-  local array=("$1")
+  local array=($1)
```
michelle192837, line 206: Why no quotes around the expansion here?

chaodaiG: Because I ran a local test and it didn't work
```
split_on_commas() {
  local IFS=,
  local array
  read -ra array <<< "$1"
  echo "${array[@]}"
}

add_suffix() {
  local array=("$1")
  local suffix="${2:-/*.yaml}"
  echo "${array[@]/%/$suffix}"
}

COMPONENT_FILE_DIR="dir1,dir2"
add_suffix "$(split_on_commas "${COMPONENT_FILE_DIR}")"
```

The output is:
```
dir1 dir2/*.yaml
```

Is it expected to be?
```
dir1/*.yaml dir2/*.yaml
```

chaodaiG:
> Can you explain why this change is needed? In what scenario would we be bumping multiple core Prow config files?
> FWIW all bump files are treated the same, we could have one list of bump files, but we call out config, components, and job config explicitly because they are common and should be bumped for all Prow instances.

In the case of https://github.com/GoogleCloudPlatform/oss-test-infra, it holds the config files of two different prow instances, so this is needed. I do agree that in this case I can set up two separate bump PR to solve this problem. 

That being said, I can drop this PR, but I think the two functions that I touched in this PR don't work as expected

prow/cmd/autobump/bump.sh
cjwagner: Is that because you changed `split_on_commas`? I don't know enough about bash to understand the effect of your changes.

cjwagner: @kubernetes/bash-firefighters

prow/cmd/autobump/bump.sh
```diff
@@ -198,17 +198,18 @@ list() {
...
-}
+ split_on_commas() {
+   local IFS=,
+   local array
+   read -ra array <<< "$1"
+   echo "${array[@]}"
+ }
+
+ add_suffix() {
+   local array=($1)
```
BenTheElder, line 209: passing arrays to functions is a huge red flag that you shouldn't be using shell anymore.

chaodaiG: Agree that we should use a better language that can be tested. While in this case it's not really passing arrays to function, it passes a string with space in it, so technically it's still a single parameter

chaodaiG:
> Is that because you changed split_on_commas? I don't know enough about bash to understand the effect of your changes.

The reason I touched `split_on_commas` was because it didn't work either:

```
split_on_commas() {
  local IFS=,
  local array=("$1")
  echo "${array[@]}"
}

split_on_commas "a,b"
```

The output is `a,b` instead of `a b`

prow/cmd/autobump/bump.sh
BenTheElder: if we're relying on word splitting, it's an array :-)

BenTheElder: in this case it appears to be file paths, so this logic is broken for any path with a space in it.

chaodaiG:
> If those functions don't work as expected lets keep the fixes for that, but please remove the support for multiple config paths if that isn't needed.

Removed the multiple config paths support, and updated the PR title 

Will think about how to deal with the space in path issue as Ben mentioned

prow/cmd/autobump/bump.sh
cjwagner: FWIW I intended for this to be written in go originally, but we decided that was taking too long and revamping the bash script would be faster. Here is my proposal/design doc: https://docs.google.com/document/d/1XgmZ7lYLoYWIchpnHYw4rjfImhozpcOxqRJgGrh0sS0/edit?usp=sharing

chaodaiG: @BenTheElder , I have updated the script solving the problem of space in array. Ran a local test and it worked like this:

```
main() {
  COMPONENT_FILE_DIR="a a,b b,c c"
  local IFS=,
  local component_file_dir_array
  read -ra component_file_dir_array <<< "${COMPONENT_FILE_DIR}"
  bumpfiles=("${component_file_dir_array[@]/%//*.yaml}")
  for f in "${bumpfiles[@]}"; do
    echo "File is: $f"
  done
}

main
```

Output:
```
File is: a a/*.yaml
File is: b b/*.yaml
File is: c c/*.yaml
```

PR #18062, opened 2020-06-24
Make autobump robust to rollbacks on gerrit.

cjwagner:
Generating a deterministic change ID based on the host/repo, prow instance name, and the version we are bumping *from* allows us to maintain a single open bump PR for each Prow instance. However it incorrectly generates a Change-Id that has already been used if the Prow instance has been reverted back to a previous version that we already created and merged a bump PR for.
This PR fixes that by iteratively hashing the chosen ID until we find an unused ID. I've tested to confirm this solves the problem.

/assign @chizhg @chaodaiG

### DIFF
--- a/prow/cmd/autobump/autobump.sh
+++ b/prow/cmd/autobump/autobump.sh
@@ -128,14 +128,35 @@ create-gerrit-pr() {
   git config http.cookiefile "${creds}"
   git remote add upstream "${GERRIT_HOST_REPO}"
 
-	change="$( echo "${GERRIT_HOST_REPO}; ${PROW_INSTANCE_NAME}; ${old_version}" | git hash-object --stdin)"
+	change_id="$(get-change-id)"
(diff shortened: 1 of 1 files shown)


prow/cmd/autobump/autobump.sh
```diff
@@ -128,16 +128,37 @@ create-gerrit-pr() {
   git config http.cookiefile "${creds}"
   git remote add upstream "${GERRIT_HOST_REPO}"
 
-	change="$( echo "${GERRIT_HOST_REPO}; ${PROW_INSTANCE_NAME}; ${old_version}" | git hash-object --stdin)"
+	change_id="$(get-change-id)"
```
chizhg, line 131: Not related to this PR, but it looks the indentation in this file is pretty inconsistent...

cjwagner: Yeah looks like there is a mix of tabs and spaces. I'll fix it next time I touch it.

PR #18064, opened 2020-06-24
Fix hmac tool issue and add docs

chizhg:
1. Fix one issue for the hmac tool - revert the hmacs for a repo if its webhook fails to be created/updated. This can prevent the valid hmac token to be deleted when the tool is run for the second time. (I didn't add the unit tests here as there is not a clean way to emulate the error with the `fakehook` utility..)

2. Update some docs for the `hmac` management tool.

/assign @cjwagner

### DIFF
--- a/prow/getting_started_deploy.md
+++ b/prow/getting_started_deploy.md
@@ -180,13 +180,15 @@ to start receiving GitHub events!
 
 ## Add the webhook to GitHub
 
-You have two options to do this.
+### Add your first webhook
 
-You can do this with the `update-hook` utility:
+You have two options to do this:
+
+1. You can do this with the `update-hook` utility:
 
 ```sh
 # Note /path/to/hook/secret and /path/to/oauth/secret from earlier secrets step
-# Note the an.ip.addr.ess from previous ingres step
+# Note the an.ip.addr.ess from previous ingress step
 
 # Ideally use https://bazel.build, alternatively try:
 #   go get -u k8s.io/test-infra/experiment/update-hook && update-hook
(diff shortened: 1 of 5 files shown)


prow/getting_started_deploy.md
```diff
@@ -199,18 +201,24 @@ $ bazel run //experiment/update-hook -- \
...
-- Click `Add webhook` 
+- Change the trigger to `Send me **everything**`.
+- Click `Add webhook`.
+
+### Use `hmac` tool to manage webhooks and hmac tokens (recommended)
+
+If you need to configure webhooks for multiple orgs or repos, the manual process does not work that well as it
+can be error-prone, and it'll be painful when you want to replace the hmac token if it is accidentally leaked.
+
+In such case, it's recommended to use the [hmac][9] tool to automatically manage the webhooks
```
fejta, line 220: This is easier:

```
it's recommended to use the [hmac] tool to automatically...


[hmac]: /prow/cmd/hmac/README.md
```

No need for the pointless number IMO

chizhg: We already have [1]-[8] in https://github.com/kubernetes/test-infra/pull/18064/files#diff-11cf136e2fed06b6cb359e3025187497R601-R608 so I feel it's probably better to keep them consistent?

fejta: It is pretty confusing that [9] comes before [6]. What is the point? Why is [9] valuable to the person reading this?

fejta: IMO you should either sort these so that [6] comes before [9] or better yet drop this physical book pattern in favor of something more appropriate for a digital document.

chizhg: I see, I actually don't need this so I changed it to be just `[hmac](/prow/cmd/hmac/README.md)`..

PR #18068, opened 2020-06-24
Don't attempt to edit archived repos

evankanderson:
It turns out archived repos are read-only, so there's not much point in trying to sync more than the archived state (assuming that desired archived state and actual archived state agree).

Updated test behavior to match observed github behavior.


petr-muller:
Makes sense, thanks for the contribution!

evankanderson:
I hit this issue attempting to sync Knative repos using Peribolos configs here: https://github.com/knative/community/blob/master/peribolos/knative.yaml

evankanderson:
Awesome, thanks for the quick turn-around!

PR #18071, opened 2020-06-24
Fix infinite loop in GetRootForClone

GeoBK:
When bug has dependents but no clones in GetRootForClone we would
stay in an infinite loop since break would not change the iterator.
Added test case to handle this case as well.

### DIFF
--- a/prow/bugzilla/client.go
+++ b/prow/bugzilla/client.go
@@ -201,7 +201,7 @@ func getRootForClone(c Client, bug *Bug) (*Bug, error) {
 		}
 		switch l := len(parent); {
 		case l <= 0:
-			break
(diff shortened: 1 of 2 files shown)


prow/bugzilla/client.go
```diff
@@ -201,7 +201,7 @@ func getRootForClone(c Client, bug *Bug) (*Bug, error) {
 		}
 		switch l := len(parent); {
 		case l <= 0:
-			break
+			return curr, utilerrors.NewAggregate(errs)
```
alvaroaleman, line 204: testcase?
/ok-to-test

GeoBK: I didn't need to add a testcase to tackle it, just needed to add a dependent bug which I have added

PR #18081, opened 2020-06-25
Add ingress-nginx jobs to build e2e images

aledbf:
~~These images are used only for e2e and never will be promoted from staging~~. The goal here is to avoid building the images every time we run e2e tests.

note: the images must be promoted. As @spiffxp suggested, I will change the name adding a prefix `e2e-test-xxx`


spiffxp:
(chatted offline, summarizing here)

You're still going to want to promote these, staging repos only have 60 days of retention.  If you intend for these to be used solely for testing, I might name thing differently.

I'd also like to see the rest of the cloudbuild.yaml files for these, I could only find images/e2e here https://github.com/kubernetes/ingress-nginx/pull/5778

PR #18083, opened 2020-06-26
Kubetest2 GCE deployer supports basic boskos

michaelmdresser:
This is the simplest implementation for the GCE deployer to support automatically acquiring projects from Boskos. It does not support a decoupled up/down step yet, that is coming in a future PR. The commit messages contain most information about change specifics. `$JOB_NAME` is based on kubetest: https://github.com/michaelmdresser/test-infra/blob/7f316346539d06be7cf45f382cd7d1c70a09b5f2/kubetest/main.go#L55

Tested against a locally running boskos.

### DIFF
--- a/kubetest2/kubetest2-gce/deployer/common.go
+++ b/kubetest2/kubetest2-gce/deployer/common.go
@@ -1,0 +1,149 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package deployer
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"path/filepath"
+	"time"
+
+	"k8s.io/klog"
+	"sigs.k8s.io/boskos/client"
+	boskosCommon "sigs.k8s.io/boskos/common"
+)
+
+func (d *deployer) init() error {
+	if d.commonOptions.ShouldBuild() {
+		if err := d.verifyBuildFlags(); err != nil {
+			return fmt.Errorf("init failed to check build flags: %s", err)
+		}
+	}
+
+	if d.commonOptions.ShouldUp() {
+		if d.GCPProject == "" {
+			klog.Info("No GCP project provided, acquiring from Boskos")
+
+			if err := d.getProjectFromBoskos(); err != nil {
+				return fmt.Errorf("init failed to get project from boskos: %s", err)
+			}
+			klog.Infof("Got project %s from boskos", d.GCPProject)
+		}
+
+		if err := d.verifyFlags(); err != nil {
+			return fmt.Errorf("init failed to verify flags for up: %s", err)
+		}
+	}
+
+	if d.commonOptions.ShouldDown() {
+		if err := d.verifyFlags(); err != nil {
+			return fmt.Errorf("init failed to verify flags for up: %s", err)
+		}
+	}
+
+	return nil
+}
+
+// getProjectFromBoskos creates a boskos client, acquires a gcp project
+// and starts a heartbeat goroutine to keep the project reserved
+func (d *deployer) getProjectFromBoskos() error {
+	// TODO
+	// boskosLocation := "http://boskos.test-pods.svc.cluster.local."
+	boskosLocation := "http://localhost:8080"
+	boskos, err := client.NewClient(
+		os.Getenv("JOB_NAME")+"-kubetest2",
+		boskosLocation,
+		"",
+		"",
+	)
+	if err != nil {
+		return fmt.Errorf("failed to create boskos client: %s", err)
+	}
+
+	resourceType := "gce-project"
+	ctx, cancel := context.WithTimeout(context.Background(), time.Minute*5)
+	defer cancel()
+
+	boskosProject, err := boskos.AcquireWait(ctx, resourceType, "free", "busy")
+	if err != nil {
+		return fmt.Errorf("failed to get a %s from boskos: %s", resourceType, err)
+	}
+	if boskosProject == nil {
+		return fmt.Errorf("boksos had no %s available", resourceType)
+	}
+
+	go func(c *client.Client, resource *boskosCommon.Resource) {
+		for range time.Tick(time.Minute * 5) {
+			if err := c.UpdateOne(resource.Name, "busy", nil); err != nil {
+				klog.Warningf("[Boskos] Update of %s failed with %v", resource.Name, err)
+			}
+		}
+	}(boskos, boskosProject)
+
+	d.boskos = boskos
+	d.boskosProject = boskosProject
+	d.GCPProject = boskosProject.Name
+
+	return nil
+}
+
+func (d *deployer) verifyFlags() error {
+	if err := d.setRepoPathIfNotSet(); err != nil {
+		return err
+	}
+
+	d.kubectl = filepath.Join(d.RepoRoot, "cluster", "kubectl.sh")
+
+	if d.GCPProject == "" {
+		return fmt.Errorf("gcp project must be set")
+	}
+
+	return nil
+}
+
+func (d *deployer) buildEnv() []string {
+	// The base env currently does not inherit the current os env (except for PATH)
+	// because (for now) it doesn't have to. In future, this may have to change when
+	// support is added for k/k's kube-up.sh and kube-down.sh which support a wide
+	// variety of environment variables. Before doing so, it is worth investigating
+	// inheriting the os env vs. adding flags to this deployer on a case-by-case
+	// basis to support individual environment configurations.
+	var env []string
+
+	// path is necessary for scripts to find gsutil, gcloud, etc
+	// can be removed if env is inherited from the os
+	env = append(env, fmt.Sprintf("PATH=%s", os.Getenv("PATH")))
+
+	// used by config-test.sh to set $NETWORK in the default case
+	// if unset, bash's set -u gets angry and kills the log dump script
+	// can be removed if env is inherited from the os
+	env = append(env, fmt.Sprintf("USER=%s", os.Getenv("USER")))
+
+	// kube-up.sh, kube-down.sh etc. use PROJECT as a parameter for all gcloud commands
+	env = append(env, fmt.Sprintf("PROJECT=%s", d.GCPProject))
+
+	// kubeconfig is set to tell kube-up.sh where to generate the kubeconfig
+	// we don't want this to be the default because this kubeconfig "belongs" to
+	// the run of kubetest2 and so should be placed in the artifacts directory
+	env = append(env, fmt.Sprintf("KUBECONFIG=%s", d.kubeconfigPath))
+
+	// kube-up and kube-down get this as a default ("kubernetes") but log-dump
+	// does not. opted to set it manually here for maximum consistency
+	env = append(env, "KUBE_GCE_INSTANCE_PREFIX=kubetest2")
+	return env
+}
--- a/kubetest2/kubetest2-gce/deployer/common.go
+++ b/kubetest2/kubetest2-gce/deployer/common.go
@@ -24,6 +24,13 @@ import (
 )
 
 func (d *deployer) init() error {
+	var err error
+	d.doInit.Do(func() { err = d.initialize() })
+	return err
+}
+
+// initialize should only be called by init(), behind a sync.Once
+func (d *deployer) initialize() error {
 	if d.commonOptions.ShouldBuild() {
 		if err := d.verifyBuildFlags(); err != nil {
 			return fmt.Errorf("init failed to check build flags: %s", err)
--- a/kubetest2/kubetest2-gce/deployer/common.go
+++ b/kubetest2/kubetest2-gce/deployer/common.go
@@ -19,6 +19,7 @@ package deployer
 import (
 	"fmt"
 	"os"
+	"time"
 
 	"k8s.io/klog"
 )
@@ -45,9 +46,22 @@ func (d *deployer) initialize() error {
 		if d.GCPProject == "" {
 			klog.Info("No GCP project provided, acquiring from Boskos")
 
-			if err := d.getProjectFromBoskos(); err != nil {
+			boskos, err := makeBoskosClient(d.BoskosLocation)
+			if err != nil {
+				return fmt.Errorf("failed to make boskos client: %s", err)
+			}
+			d.boskos = boskos
+
+			projectName, err := getProjectFromBoskos(
+				d.boskos,
+				time.Duration(d.BoskosAcquireTimeoutSeconds)*time.Second,
+				d.boskosHeartbeatClose,
+			)
+
+			if err != nil {
 				return fmt.Errorf("init failed to get project from boskos: %s", err)
 			}
+			d.GCPProject = projectName
 			klog.Infof("Got project %s from boskos", d.GCPProject)
 		}
 
--- a/kubetest2/kubetest2-gce/deployer/common.go
+++ b/kubetest2/kubetest2-gce/deployer/common.go
@@ -17,15 +17,10 @@ limitations under the License.
 package deployer
 
 import (
-	"context"
 	"fmt"
 	"os"
-	"path/filepath"
-	"time"
 
 	"k8s.io/klog"
-	"sigs.k8s.io/boskos/client"
-	boskosCommon "sigs.k8s.io/boskos/common"
 )
 
 func (d *deployer) init() error {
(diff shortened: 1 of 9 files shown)


kubetest2/kubetest2-gce/deployer/common.go
```diff
@@ -0,0 +1,162 @@
...
+	if err := d.setRepoPathIfNotSet(); err != nil {
+		return err
+	}
+
+	d.kubectl = filepath.Join(d.RepoRoot, "cluster", "kubectl.sh")
+
+	// verifyUpFlags does not check for a gcp project because it is
+	// assumed that one will be acquired from boskos if it is not set
+	if d.BoskosLocation == "" {
+		d.BoskosLocation = "http://boskos.test-pods.svc.cluster.local."
```
amwat, line 112: consider making this the flag default value in general.

kubetest2/kubetest2-gce/deployer/common.go
```diff
@@ -0,0 +1,162 @@
...
+	}
+
+	return nil
+}
+
+func (d *deployer) verifyDownFlags() error {
```
amwat, line 118: move to down.go

kubetest2/kubetest2-gce/deployer/common.go
```diff
@@ -0,0 +1,162 @@
...
+		}
+	}(boskos, boskosProject)
+
+	d.boskos = boskos
+	d.boskosProject = boskosProject
+	d.GCPProject = boskosProject.Name
+
+	return nil
+}
+func (d *deployer) verifyUpFlags() error {
```
amwat, line 102: move to up.go

kubetest2/kubetest2-gce/deployer/common.go
```diff
@@ -0,0 +1,162 @@
...
+
+	if d.commonOptions.ShouldUp() {
+		if err := d.verifyUpFlags(); err != nil {
+			return fmt.Errorf("init failed to verify flags for up: %s", err)
+		}
+
+		if d.GCPProject == "" {
+			klog.Info("No GCP project provided, acquiring from Boskos")
+
+			if err := d.getProjectFromBoskos(); err != nil {
```
amwat, line 46: I think the defaulting will need some more thought here.
for example for local testing I would think that the already set gcloud default project would be the default.

Maybe it's worth differentiating the logic for local vs CI?

kubetest2/kubetest2-gce/deployer/common.go
```diff
@@ -0,0 +1,162 @@
...
+	"os"
+	"path/filepath"
+	"time"
+
+	"k8s.io/klog"
+	"sigs.k8s.io/boskos/client"
+	boskosCommon "sigs.k8s.io/boskos/common"
+)
+
+func (d *deployer) init() error {
```
amwat, line 31: sync.Once? 😬

kubetest2/kubetest2-gce/deployer/common.go
```diff
@@ -0,0 +1,162 @@
...
+
+	go func(c *client.Client, resource *boskosCommon.Resource) {
+		for range time.Tick(time.Minute * 5) {
+			if err := c.UpdateOne(resource.Name, "busy", nil); err != nil {
+				klog.Warningf("[Boskos] Update of %s failed with %v", resource.Name, err)
+			}
+		}
+	}(boskos, boskosProject)
+
+	d.boskos = boskos
```
amwat, line 96: with foresight of not needing to persist this across up and down. let's not store these in memory anyway.
shouldn't be too bad constructing the client again if we keep the same name.

kubetest2/kubetest2-gce/deployer/common.go
```diff
@@ -0,0 +1,162 @@
...
+	resourceType := "gce-project"
+	ctx, cancel := context.WithTimeout(context.Background(), time.Minute*5)
+	defer cancel()
+
+	boskosProject, err := boskos.AcquireWait(ctx, resourceType, "free", "busy")
+	if err != nil {
+		return fmt.Errorf("failed to get a %s from boskos: %s", resourceType, err)
+	}
+	if boskosProject == nil {
+		return fmt.Errorf("boksos had no %s available", resourceType)
```
amwat, line 85: typo: s/boksos/boskos

kubetest2/kubetest2-gce/deployer/common.go
michaelmdresser: If the name of the boskos resource is retained, can the client be completely thrown away? I'm not sure if the owner field of the boskos struct matters: https://github.com/kubernetes-sigs/boskos/tree/master/client

kubetest2/kubetest2-gce/deployer/common.go
michaelmdresser: Done.

kubetest2/kubetest2-gce/deployer/common.go
michaelmdresser: Done.

kubetest2/kubetest2-gce/deployer/common.go
michaelmdresser: Done.

kubetest2/kubetest2-gce/deployer/common.go
michaelmdresser: Done.

kubetest2/kubetest2-gce/deployer/up.go
```diff
@@ -135,3 +93,56 @@ func enableComputeAPI(project string) error {
...
+		d.BoskosLocation,
+		"",
+		"",
+	)
+	if err != nil {
+		return fmt.Errorf("failed to create boskos client: %s", err)
+	}
+
+	resourceType := "gce-project"
+	ctx, cancel := context.WithTimeout(context.Background(), time.Minute*5)
```
rancerbeta, line 111: generally, I've seen context should be passed in and acquired from background at the top level. This lets you share auth and other metadata easily. Gj on adding a timeout here though.

kubetest2/kubetest2-gce/deployer/up.go
```diff
@@ -135,3 +93,56 @@ func enableComputeAPI(project string) error {
...
+
+	boskosProject, err := boskos.AcquireWait(ctx, resourceType, "free", "busy")
+	if err != nil {
+		return fmt.Errorf("failed to get a %s from boskos: %s", resourceType, err)
+	}
+	if boskosProject == nil {
+		return fmt.Errorf("boskos had no %s available", resourceType)
+	}
+
+	go func(c *client.Client, resource *boskosCommon.Resource) {
```
rancerbeta, line 122: Will this goroutine ever get cleaned up? May want to use a Done channel to allow explicit closing and return a close func. Even better, issue an explicit release to boskos if closed, say at teardown

kubetest2/kubetest2-gce/deployer/up.go
rancerbeta: You'll probably want this boskos timeout to be configurable at the top level.  Or at the very least, a const at the top of the file.

kubetest2/kubetest2-gce/deployer/up.go
```diff
@@ -135,3 +93,56 @@ func enableComputeAPI(project string) error {
...
+		return fmt.Errorf("failed to create boskos client: %s", err)
+	}
+
+	resourceType := "gce-project"
+	ctx, cancel := context.WithTimeout(context.Background(), time.Minute*5)
+	defer cancel()
+
+	boskosProject, err := boskos.AcquireWait(ctx, resourceType, "free", "busy")
+	if err != nil {
+		return fmt.Errorf("failed to get a %s from boskos: %s", resourceType, err)
```
rancerbeta, line 116: nit: use %q for resourceType for nice quoting/separation from error template text

kubetest2/kubetest2-gce/deployer/common.go
```diff
@@ -0,0 +1,89 @@
...
+	if d.commonOptions.ShouldDown() {
+		if err := d.verifyDownFlags(); err != nil {
+			return fmt.Errorf("init failed to verify flags for down: %s", err)
+		}
+	}
+
+	return nil
+}
+
+func (d *deployer) buildEnv() []string {
```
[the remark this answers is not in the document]
michaelmdresser: I don't quite understand what "whitelisting via a flag means," could you explain?

kubetest2/kubetest2-gce/deployer/common.go
michaelmdresser: Done.

kubetest2/kubetest2-gce/deployer/deployer.go
```diff
@@ -37,12 +39,19 @@ type deployer struct {
 	// generic parts
 	commonOptions types.Options
 
-	kubeconfigPath   string
-	kubectl          string
-	logsDir          string
+	kubeconfigPath string
+	kubectl        string
```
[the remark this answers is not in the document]
michaelmdresser: Done, definitely agree.

kubetest2/kubetest2-gce/deployer/up.go
michaelmdresser: Done.

kubetest2/kubetest2-gce/deployer/common.go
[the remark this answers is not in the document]
rancerbeta: something like
kubetest2 gce --whitelistedEnvvars="SERVICE_SA,TEST_CONFIG"

kubetest2/kubetest2-gce/deployer/up.go
michaelmdresser: It's definitely not the cleanest way to do this. Updated with an explicit close after the project is released.

kubetest2/kubetest2-gce/deployer/up.go
michaelmdresser: Made the timeout configurable. I'm not sure about grabbing a context at the top level, given that this is currently the only place it is needed and I don't (currently) see a use case for sharing auth/other metadata outside of the deployer struct itself.

kubetest2/kubetest2-gce/deployer/common.go
michaelmdresser: I'm not sure how I feel about defaulting to the gcloud default, because its not clear behavior to the user if they don't inspect the flags. Because kubetest2 isn't actually an extension of gcloud in any way (it just happens to use it) I think its an unclear default. Happy to discuss further, we could also consider a `--local` flag but I don't like the idea of that because it could lead to some messy divergences of behavior in the future in the name of slightly better UX.

kubetest2/kubetest2-gce/deployer/up.go
rancerbeta: That sgtm, we can revisit if there are more requests being made from the deployer.

kubetest2/kubetest2-gce/deployer/common.go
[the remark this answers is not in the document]
michaelmdresser: One thing we wanted to specifically avoid was a way of passing env vars through, we wanted to make them explicit in flags instead. This doesn't seems as bad as kubetest's `--env` flag but it still sort of goes against that goal.

kubetest2/kubetest2-gce/deployer/common.go
amwat: the gcloud default is honored by kube-up so was mainly coming from extending that. But it's also fine to have our own prescriptive required flag.

For CI vs local, it should motly be auto-detect based on what envs prow sets. Mainly PROW_JOB_ID.

kubetest2/kubetest2-gce/deployer/common.go
amwat: yes we also need to use the same owner.

kubetest2/kubetest2-gce/deployer/up.go
```diff
@@ -135,3 +93,75 @@ func enableComputeAPI(project string) error {
...
+// boskos from taking the resource from the deployer while it is still in use.
+func startBoskosHeartbeat(c *client.Client, resource *boskosCommon.Resource, interval time.Duration, close chan struct{}) {
+	go func(c *client.Client, resource *boskosCommon.Resource) {
+		for {
+			select {
+			case <-close:
+				klog.Info("Boskos heartbeat func received signal to close")
+				return
+			case <-time.Tick(interval):
+				klog.Info("Sending heartbeat to Boskos")
```
amwat, line 147: I think this should be at a lower verbosity level like debug.
https://godoc.org/github.com/kubernetes/klog#V

let's change all info to V(1) and debug to V(2)

kubetest2/kubetest2-gce/deployer/up.go
```diff
@@ -135,3 +93,75 @@ func enableComputeAPI(project string) error {
 
 	return nil
 }
+
+// getProjectFromBoskos creates a boskos client, acquires a gcp project
+// and starts a heartbeat goroutine to keep the project reserved
+func (d *deployer) getProjectFromBoskos() error {
```
amwat, line 99: maybe move all boskos methods to a different file, and try to remove deployer as a receiver as much as possible.

kubetest2/kubetest2-gce/deployer/up.go
```diff
@@ -135,3 +93,75 @@ func enableComputeAPI(project string) error {
...
+	d.boskos = boskos
+	d.boskosProject = boskosProject
+	d.GCPProject = boskosProject.Name
+
+	return nil
+}
+
+// startBoskosHeartbeat starts a goroutine that sends periodic updates to boskos
+// about the provided resource until the channel is closed. This prevents
+// boskos from taking the resource from the deployer while it is still in use.
```
amwat, line 138: reaper from taking the resource ... 
to be specific

also we only care about heartbeat in the CI use case, i don't think we should really care about users running reaper/janitor locally.

kubetest2/kubetest2-gce/deployer/common.go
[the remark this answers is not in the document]
rancerbeta: got it, that sounds good to me

kubetest2/kubetest2-gce/deployer/up.go
michaelmdresser: I get the concept of V, but to set it it looks like it expects flags to be parsed, but kubetest2 uses a different flag parser. Do you know how to get around this?

https://godoc.org/github.com/kubernetes/klog#InitFlags

kubetest2/kubetest2-gce/deployer/common.go
michaelmdresser: I did some experimenting and ran into https://github.com/kubernetes-sigs/boskos/blob/master/client/client.go#L334

It seems like we won't be able to decouple up and down in the boskos case because we would have to persist boskos client state which is a private struct field.

kubetest2/kubetest2-gce/deployer/up.go
michaelmdresser: Done.

kubetest2/kubetest2-gce/deployer/up.go
michaelmdresser: Updated the comment. I'd like to avoid a split between CI and non-CI behavior for as long as I can, and this seems like a pretty harmless goroutine to leave running.

kubetest2/kubetest2-gce/deployer/common.go
michaelmdresser: I'm favoring a prescriptive flag like we are planning for all configuration of the deployer. It may be that the deployer will have to eventually detect or be told that is or is not running in Prow, but I don't think its necessary right now.

kubetest2/kubetest2-gce/deployer/common.go
amwat: that's sad to know. not sure what the motivation was for having in-memory storage in the client.
/cc @ixdy

kubetest2/kubetest2-gce/deployer/deployer.go
```diff
@@ -37,20 +40,37 @@ type deployer struct {
...
-		logsDir:        filepath.Join(opts.ArtifactsDir(), "cluster-logs"),
+		commonOptions:               opts,
+		kubeconfigPath:              filepath.Join(opts.ArtifactsDir(), "kubetest2-kubeconfig"),
+		logsDir:                     filepath.Join(opts.ArtifactsDir(), "cluster-logs"),
+		boskosHeartbeatClose:        make(chan struct{}),
+		BoskosAcquireTimeoutSeconds: 5 * 60,
+		BoskosLocation:              "http://boskos.test-pods.svc.cluster.local.",
 	}
 
 	flagSet, err := gpflag.Parse(d)
```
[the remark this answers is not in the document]
amwat: perhaps something like this?

kubetest2/kubetest2-gce/deployer/up.go
michaelmdresser: Done, was a bit confused about how to add to the flag set. Tested at all verbosity levels.

kubetest2/kubetest2-gce/deployer/deployer.go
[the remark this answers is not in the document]
michaelmdresser: Yep, didn't fully understand how the flagset stuff worked until this morning.

PR #18084, opened 2020-06-26
Remove no longer supported 1.15

jsturtevant:
While updating latest host images for sig-windows tests on https://github.com/kubernetes-sigs/windows-testing/pull/172, noticed we are still running 1.15 jobs which is no longer supported version. This removes the test configuration.

cc: @chewong @marosset 

/sig windows


marosset:
@jsturtevant can we remove the 1.14 and 1.15  aks-engine templates at https://github.com/kubernetes-sigs/windows-testing/tree/master/job-templates as well?

jsturtevant:
they were removed in https://github.com/kubernetes-sigs/windows-testing/pull/172

PR #18092, opened 2020-06-27
add the fetch from remote method to interactor's interface

droslean:
Add the fetch from remote method to interactor's interface.
It also removes it from the v1 git client.

### DIFF
--- a/prow/git/git.go
+++ b/prow/git/git.go
@@ -488,22 +488,3 @@ func (i *Repo) ShowRef(commitlike string) (string, error) {
 	}
 	return strings.TrimSpace(string(out)), nil
 }
-
-// FetchFromRemote runs a git fetch command for the specified remote.
-// Pass the destination as org/repo format since the actual URI will be constructed
-// using the existing credentials and will always use the https protocol.
-func (r *Repo) FetchFromRemote(orgRepo, branch string) error {
-	r.logger.Infof("Fetching from '%s (branch: %s)'.", orgRepo, branch)
-
-	remoteURI := fmt.Sprintf("https://%s/%s", r.host, orgRepo)
-	if r.user != "" && r.pass != "" {
-		remoteURI = fmt.Sprintf("https://%s:%s@%s/%s", r.user, r.pass, r.host, orgRepo)
-	}
-
-	co := r.gitCommand("fetch", remoteURI, branch)
-	out, err := co.CombinedOutput()
-	if err != nil {
-		return fmt.Errorf("Fetching failed, output: %q, error: %v", string(out), err)
-	}
-	return nil
-}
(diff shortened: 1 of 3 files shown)


prow/git/v2/interactor.go
```diff
@@ -54,6 +54,8 @@ type Interactor interface {
 	Fetch() error
 	// FetchRef fetches the refspec
 	FetchRef(refspec string) error
+	// FetchFromRemote fetches the branch of the given remote
+	FetchFromRemote(remote RemoteResolver, branch string) error
```
stevekuznetsov, line 58: Why isn't this remote part of the client when it's constructed?

droslean: It could already have a different remote when it is constructed. So before this, you couldn't just have a cloned remote and fetch from different remote.

droslean: One use-case is:

- Clone a repo from remote x
- Fetch a branch from remote y
- Merge
- Push branch to remote x

prow/git/v2/adapter.go
```diff
@@ -86,6 +87,19 @@ func (a *repoClientAdapter) Fetch() error {
 	return errors.New("no Fetch implementation exists in the v1 repo client")
 }
 
+func (a *repoClientAdapter) FetchFromRemote(resolver RemoteResolver, branch string) error {
```
stevekuznetsov, line 90: Where does this method signature come from? Why do we have a full `resolver`? Why is this useful? What is the method on the `git` interface?

droslean: I might have missed something in the adapter's implementation. Let me check on this.

droslean: @stevekuznetsov Are you suggesting to remove this completely from the v1 client?

prow/git/git.go
```diff
@@ -488,22 +488,3 @@ func (i *Repo) ShowRef(commitlike string) (string, error) {
 	}
 	return strings.TrimSpace(string(out)), nil
 }
-
-// FetchFromRemote runs a git fetch command for the specified remote.
-// Pass the destination as org/repo format since the actual URI will be constructed
-// using the existing credentials and will always use the https protocol.
-func (r *Repo) FetchFromRemote(orgRepo, branch string) error {
```
alvaroaleman, line 495: Why remove it?

droslean: Do we actually want this in the v1 client? I can write the adapter also. I did this in previous versions of this PR.

alvaroaleman: I would assume that if we have code, we have it for a reason and we shouldn't just randomly remove it

droslean: Yea I added this in a previous PR, but it seems its better to have the whole thing in v2 only. Shall we keep it? I can make changes.

PR #18094, opened 2020-06-27
[git v2 publisher] add the ability to push to central remote

droslean:
Since there is no way to use the git v2 client to push the changes in the configured/cloned repo this PR gives the ability to publisher to push/forcepush to either publish or central remote as well.

### DIFF
--- a/prow/git/v2/adapter.go
+++ b/prow/git/v2/adapter.go
@@ -74,8 +74,12 @@ func (a *repoClientAdapter) Commit(title, body string) error {
 	return errors.New("no Commit implementation exists in the v1 repo client")
 }
 
-func (a *repoClientAdapter) ForcePush(branch string) error {
+func (a *repoClientAdapter) PushToFork(branch string, force bool) error {
 	return a.Repo.Push(branch)
+}
+
+func (a *repoClientAdapter) PushToCentral(branch string, force bool) error {
+	return errors.New("no PushToCentral implementation exists in the v1 repo client")
 }
 
 func (a *repoClientAdapter) MirrorClone() error {
--- a/prow/git/v2/publisher.go
+++ b/prow/git/v2/publisher.go
@@ -26,16 +26,23 @@ import (
 type Publisher interface {
 	// Commit stages all changes and commits them with the message
 	Commit(title, body string) error
-	// ForcePush runs `git push -f` to the publish remote
-	ForcePush(branch string) error
+	// PushToFork pushes the local state to the fork remote
+	PushToFork(branch string, force bool) error
+	// PushToCentral pushes the local state to the central remote
+	PushToCentral(branch string, force bool) error
 }
 
 // GitUserGetter fetches a name and email for us in git commits on-demand
 type GitUserGetter func() (name, email string, err error)
 
+type remotes struct {
+	publishRemote RemoteResolver
+	centralRemote RemoteResolver
+}
+
 type publisher struct {
 	executor executor
-	remote   RemoteResolver
+	remotes  remotes
 	info     GitUserGetter
 	logger   *logrus.Entry
 }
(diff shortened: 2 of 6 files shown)


droslean:
> Pushing is a publication action, which is why the publisher (pkg/git/v2/publisher.go) already has this functionality. What are you trying to do here?

@stevekuznetsov  The publisher is pushing to `user/repo` instead of the configured `org/repo`. I think this is fine because it should push to user's fork. However, it is not possible to push a change to the cloned repo with the v2 git client. That is why I add this method in interactor by not touching the publisher at all.  Does this makes sense?

stevekuznetsov:
No. The publisher publishes. Invent a way to configure _where_ but let's not add separate push implementations to separate remotes in the different subsets of the v2 client.

prow/git/v2/adapter.go
```diff
@@ -78,6 +78,14 @@ func (a *repoClientAdapter) ForcePush(branch string) error {
 	return a.Repo.Push(branch)
 }
 
+func (a *repoClientAdapter) CentralPush(branch string, force bool) error {
+	return errors.New("no CentralPush implementation exists in the v1 repo client")
```
alvaroaleman, line 82: Why not implement this?

prow/git/v2/publisher.go
```diff
@@ -59,14 +68,49 @@ func (p *publisher) Commit(title, body string) error {
 	return nil
 }
 
-// ForcePush pushes the local state to the remote
+// ForcePush pushes the local state to the publish remote
 func (p *publisher) ForcePush(branch string) error {
-	p.logger.Infof("Pushing branch %q", branch)
-	remote, err := p.remote()
+	if err := p.PublishPush(branch, true); err != nil {
```
alvaroaleman, line 73: this is a complicated way of doing `return p.PublishPush(branch, true)`

prow/git/v2/publisher.go
```diff
@@ -28,14 +28,23 @@ type Publisher interface {
 	Commit(title, body string) error
 	// ForcePush runs `git push -f` to the publish remote
 	ForcePush(branch string) error
+	// PublishPush pushes the local state to the publish remote
+	PublishPush(branch string, force bool) error
```
alvaroaleman, line 32: `PublishPush`? What about maybe `PushToFork`? As everywhere else, please use functional opts pattern for opts rather than requiring them to be set.

prow/git/v2/publisher.go
droslean: I kept this for backward compability. I guess it should be fine to change this from the components that are using it already.

prow/git/v2/publisher.go
droslean: I kept the same naming here, since there are two remotes, central and publish ones.

prow/git/v2/adapter.go
droslean: I don't think that we should extent the v1 client in any ways.

prow/git/v2/publisher.go
alvaroaleman: same as what?

prow/git/v2/publisher.go
alvaroaleman: No, my point is that

```
return p.PublishPush(branch, true)
```

is the exact same as
```
if err := p.PublishPush(branch, true); err != nil {
    return err
}
return nil
```

Except that the former is written in one line of code wheas what you did needs four

prow/git/v2/publisher.go
alvaroaleman: Okay but this is still confusing because there is a `ForcePush` that pushes to the same fork as the `PublishPush` but the method doesn't have `Publish` in its name. Wouldn't it be better to name this just `Push` if the `PublishRemote` is the default so its consistent?

droslean: I will take a look at what it takes to completely remove the `ForcePush`. From the first look, only the cherrypicker is using that.

PR #18095, opened 2020-06-28
Configure kubeflow org to require cla/google status context

Bobgy:
/assign @jlewi @cjwagner 

Fixes https://github.com/kubeflow/pipelines/issues/1653

### DIFF
--- a/config/prow/config.yaml
+++ b/config/prow/config.yaml
@@ -412,7 +412,6 @@ tide:
     labels:
     - lgtm
     - approved
-    - "cla: yes"
     missingLabels:
     - do-not-merge
     - do-not-merge/hold
--- a/config/prow/config.yaml
+++ b/config/prow/config.yaml
@@ -404,6 +404,7 @@ tide:
     labels:
     - lgtm
     - approved
+    - "cla: yes"
     missingLabels:
     - do-not-merge
     - do-not-merge/hold
--- a/config/prow/config.yaml
+++ b/config/prow/config.yaml
@@ -395,6 +395,14 @@ branch-protection:
           required_status_checks:
             contexts:
             - continuous-integration/travis-ci
+    kubeflow:
+      protect: true
+      required_status_checks:
+        contexts:
+        - cla/google
+      exclude:
+      - "^revert-" # don't protect revert branches
+      - "^dependabot/" # don't protect branches created by dependabot
 
 tide:
   sync_period: 2m


Bobgy:
/hold

@jlewi before merging this PR, we should guarantee other kubeflow repos have `cla: yes` and `cla: no` labels appended by @googlebot automatically.

Bobgy:
I checked, it seems most repos don't have required labels. Is it possible adding labels at the organization level?

config/prow/config.yaml
```diff
@@ -404,6 +404,7 @@ tide:
     labels:
     - lgtm
     - approved
+    - "cla: yes"
```
jlewi, line 407: Why does tide need to be made aware of the cla label?
If the CLA status check isn't passing then doesn't that prevent tide from merging the PR?

Your comment: https://github.com/kubernetes/test-infra/pull/18095
made it sound like the problem was the repo didn't have these labels defined which was causing the problem.

Bobgy: Let me explain the context.

In https://github.com/kubeflow/pipelines/issues/1653

They mentioned when tide configuration doesn't exactly match github status. If an offending PR appears, tide bot will continuously try to merge the PR and fail. That will soon drain github api rate limit and affecting all other repos managed by tide.

Tide cannot read github statuses, so it must use a label to get statuses.

cblecker: Tide can check status contexts, and require them before merging, if you configure it to:
https://github.com/kubernetes/test-infra/blob/8c9d7c79573d0368573eea2ffd0403a8498f9b00/config/prow/config.yaml#L102-L104

This can be configured on a repo or org level.

cjwagner: Yes, you can use either mechanism to make Tide respect that requirement as mentioned here: https://github.com/kubeflow/pipelines/issues/1653#issuecomment-650287176

Requiring the status context rather than the label is arguably better because if the google-bot posts a context and label that conflict, using the context as the source of truth will prevent Tide from clashing with the GH settings.

/ok-to-test

Bobgy: Thanks @cblecker and @cjwagner!

I missed the later half of the comment. Then I think requiring a github status context is definitely better, will update this PR.

config/prow/config.yaml
```diff
@@ -395,6 +395,14 @@ branch-protection:
...
             contexts:
             - continuous-integration/travis-ci
+    kubeflow:
+      protect: true
+      required_status_checks:
+        contexts:
+        - cla/google
+      exclude:
+      - "^revert-" # don't protect revert branches
+      - "^dependabot/" # don't protect branches created by dependabot
```
Bobgy, line 405: These two excludes are copied from other rules.

Bobgy:
/assign @jlewi @cjwagner @cblecker 
for review

jlewi:
Thanks @Bobgy 
/lgtm
/approve

jlewi:
/assign @cblecker

@cblecker could you please review thanks?

Bobgy:
/unhold
This no longer depend on any other changes.

PR #18098, opened 2020-06-29
fix(config): fix merge_commit_template format and apply to kubeflow org

Bobgy:
Follow up on https://github.com/kubernetes/test-infra/pull/18043.

Example commit with current config: https://github.com/kubeflow/pipelines/commit/c52a73e52c64a2d1414d0294e8617da42445dfd8.
Config documentation: https://github.com/kubernetes/test-infra/blob/master/prow/cmd/tide/config.md#general-configuration

Just realized the `{{ .Number }}` is just the PR number, so I added a `#` before it.

/assign @jlewi @spiffxp

Can I get your approvals? Thanks!

### DIFF
--- a/config/prow/config.yaml
+++ b/config/prow/config.yaml
@@ -521,7 +521,7 @@ tide:
     kubernetes/cloud-provider-openstack: squash
     kubernetes/dashboard: squash
   merge_commit_template:
-    kubeflow/pipelines:
+    kubeflow:
       title: "{{ .Title }} (#{{ .Number }})"
   pr_status_base_urls:
     '*': https://prow.k8s.io/pr
--- a/config/prow/config.yaml
+++ b/config/prow/config.yaml
@@ -522,7 +522,7 @@ tide:
     kubernetes/dashboard: squash
   merge_commit_template:
     kubeflow/pipelines:
-      title: "{{ .Title }} ({{ .Number }})"
+      title: "{{ .Title }} (#{{ .Number }})"
   pr_status_base_urls:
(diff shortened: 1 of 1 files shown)


Bobgy:
@jlewi Looking at https://github.com/kubeflow/pipelines/commit/54a596abd837586e294b03c31d10ac61d9273f1a, it seems this config keeps commit content, so there isn't much unexpected information loss.

Shall we enable this across kubeflow?

jlewi:
> @jlewi Looking at kubeflow/pipelines@54a596a, it seems this config keeps commit content, so there isn't much unexpected information loss.

Can you remind me what the impact of the template is?

> Shall we enable this across kubeflow?

Probably

spiffxp:
/ok-to-test
/approve
I will leave /lgtm to @jlewi

Bobgy:
@jlewi details in https://github.com/kubernetes/test-infra/issues/17915

the template makes sure commit message is taken from PR title even If there is only one commit.

jlewi:
Thanks @Bobgy is it possible to configure this at the org level rather than the individual repo level so we can be consistent?

Bobgy:
> Thanks @Bobgy is it possible to configure this at the org level rather than the individual repo level so we can be consistent?

Yes, it's possible.
I'll update the PR then.

Bobgy:
@jlewi Updated, please review.
config documented in
https://github.com/kubernetes/test-infra/blob/master/prow/cmd/tide/config.md#general-configuration

PR #18112, opened 2020-06-30
Enable OOM tracking in presubmit master perf test jobs

tosi3k:
Roll out the mechanism implemented in kubernetes/perf-tests#1309 to presubmit master performance tests.

/sig scalability
/cc jkaniuk
/cc mm4tt
/cc wojtek-t

### DIFF
--- a/config/jobs/kubernetes/sig-scalability/sig-scalability-presubmit-jobs.yaml
+++ b/config/jobs/kubernetes/sig-scalability/sig-scalability-presubmit-jobs.yaml
@@ -35,6 +35,7 @@ presubmits:
         - --provider=gce
         - --stage=gs://kubernetes-release-pull/ci/pull-kubernetes-e2e-gce-100-performance
         - --tear-down-previous
+        - --env=CL2_ENABLE_CLUSTER_OOMS_TRACKER=true
         - --test=false
         - --test-cmd=$GOPATH/src/k8s.io/perf-tests/run-e2e.sh
         - --test-cmd-args=cluster-loader2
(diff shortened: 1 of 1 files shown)


jkaniuk:
Testgrids for periodics are green
/lgtm

wojtek-t:
/remove-lifecycle stale

@tosi3k - do we want to proceed with this?

tosi3k:
Sorry, completely forgot about this PR.

The problem here is that `pull-kubernetes-e2e-gce-100-performance` has a mutual config for all release-1.XX branches. In perf-tests 1.17 and below we use an override file to enable the tracker whereas in the newer releases we use `CL2_`-prefixed env var. This was because we lacked support for `CL2_` vars in clusterloader before 1.18 but I have cherry-picked handling them back to 1.16 ~2 months ago.

Because of that, what I would do is make the configs of older perf-tests branches use this `CL2_` env var and revisit this PR afterwards.

jkaniuk:
> The problem here is that `pull-kubernetes-e2e-gce-100-performance` has a mutual config for all release-1.XX branches. In perf-tests 1.17 and below we use an override file to enable the tracker whereas in the newer releases we use `CL2_`-prefixed env var. This was because we lacked support for `CL2_` vars in clusterloader before 1.18 but I have cherry-picked handling them back to 1.16 ~2 months ago.
> 
> Because of that, what I would do is make the configs of older perf-tests branches use this `CL2_` env var and revisit this PR afterwards.

We can fork our pulls per-release. I thought it was done like that already. WDYT?

tosi3k:
> We can fork our pulls per-release. I thought it was done like that already. WDYT?

Ah, you're right - I was just looking at `sig-scalability` subdirectory while we indeed fork the presubmits to `sig-release`. Thanks Jacek for pointing this out.

tosi3k:
Actually, OOM tracking is enabled by default in the master branch of perf-tests which will include release-1.20 onwards (FTR: https://github.com/kubernetes/perf-tests/pull/1550). Because of that I will omit turning this in our master jobs.

I noticed that 1.19 release didn't have this enabled, though. I will open a separate PR cleaning up this mess, though.

PR #18089, opened 2020-06-27
Revert "Update to containerd 1.3.5"

BenTheElder:
Reverts kubernetes/test-infra#18086

CI is failing, can't download.
this is also blocking PRs

we should test changes like this on a canary job first or make this one non-blocking @dims

PR #18114, opened 2020-06-30
Update to containerd 1.3.5 - Take #2

dims:
Previous attempt https://github.com/kubernetes/test-infra/pull/18086 was reverted in https://github.com/kubernetes/test-infra/pull/18089. A canary job was added in https://github.com/kubernetes/test-infra/pull/18091

which is now green  ✅  : 
https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/92540/pull-kubernetes-e2e-gce-ubuntu-containerd-canary/1277916511897391104/

### DIFF
--- a/config/jobs/kubernetes/sig-cloud-provider/gcp/gcp-gce.yaml
+++ b/config/jobs/kubernetes/sig-cloud-provider/gcp/gcp-gce.yaml
@@ -187,7 +187,7 @@ presubmits:
             - --build=bazel
             - --cluster=
             - --env=KUBE_CONTAINER_RUNTIME=containerd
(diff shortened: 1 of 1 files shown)


hogepodge:
This is easy enough for me to review, but since it's tagged for sig-cloud-provider I'll defer to @andrewsykim or @cheftako to lgtm. I should probably find the spots where I am in all of the reviewers and approvers file and get myself out of them.

cheftako:
I believe the key point is that dims fixed the underlying issue of the previous attempt in https://github.com/kubernetes/kubernetes/pull/92569/files. Seems like this should work and that's confirmed by the test.

PR #18119, opened 2020-06-30
Revert "bump.sh checks if the image exists"

cjwagner:
This reverts commit 0ebc4a0188aab7abf5232c11d092145be74f3929.

The image existence check is currently broken: https://prow.istio.io/view/gcs/istio-prow/logs/ci-prow-autobump/1715
(The images all exist.)
Since it is a redundant validation check that the upstream autobumper already ensures I'm reverting this to unblock other Prows.

/assign @chizhg @chaodaiG

### DIFF
--- a/prow/cmd/autobump/bump.sh
+++ b/prow/cmd/autobump/bump.sh
@@ -66,7 +66,7 @@ main() {
   else
     usage
   fi
-  echo -e "Bumping: 'gcr.io/k8s-prow/' images to $(color-version "${new_version}") ..." >&2
+  echo -e "Bumping: 'gcr.io/k8s-prow/' images to $(color-version ${new_version}) ..." >&2
 
   local IFS=,
   local component_file_dir_array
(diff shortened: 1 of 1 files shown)


chaodaiG:
/lgtm
/hold
In case @chizhg has concern

cjwagner:
I added back a couple of the UX improvements/fixes that were included in the commit I reverted. PTAL.

cjwagner:
Chi pointed out that it is possible to use bump.sh to bump to an arbitrary commit even though we don't use that feature. The validation makes sense in that case and [Chi's fix](https://github.com/kubernetes/test-infra/pull/18120) is simple so we're going to try a forward fix, but I'm not a huge fan of the extra complexity this check adds for validation that is redundant in all of our actual use cases.
Rewriting autobump.sh in go and reusing experiment/autobumper like I originally suggested would let us reuse the image existence checks written in go, write unit tests, and avoid bash quoting/splitting complexities.

PR #18129, opened 2020-07-01
Add coverage Properties to xml for ci-kubernetes-coverage-unit

MushuEE:
To display the coverage results as short text in Testgrid the xml needs <property> tags under the testcases. `gopherage junit` handles this so I added it as part of the run. Similar to earlier in the file with e2e tests (currently broken). 

This is a WIP: Still looking for best way to test this. Might be easiest to run a fake job locally that skips the testing and directly interfaces with a static results file.


MushuEE:
Hitting `level=fatal msg="Error loading config" error="default_decoration_configs['*'] is missing"` when running locally

MushuEE:
Now Signed.

MushuEE:
@MushuEE < reminder that I might need to revert this if I made a change that breaks the job.

PR #18131, opened 2020-07-01
Enable kubeflow/pipelines postsubmit tests for release branches

Bobgy:
/assign @jlewi 

part of https://github.com/kubeflow/pipelines/issues/4046


jlewi:
Why don't you just run tests on all branches?

/lgtm
/approve

Bobgy:
@jlewi when repo owners create PRs from Github UI, it automatically creates branches in the repo. We don't want to run postsubmit tests on them.

PR #18133, opened 2020-07-01
disable report for ipv6 presubmit

BenTheElder:
It's not ready to be blocking, and contributors don't understand always_run + optional.


aojea:
lgtm

PR #18136, opened 2020-07-01
Configure perf-tests presubmits to run on older release branches

jkaniuk:
Reverts #17589

Ref: https://github.com/kubernetes/perf-tests/issues/1092

/cc @mm4tt

### DIFF
--- a/config/jobs/kubernetes/sig-scalability/sig-scalability-presubmit-jobs.yaml
+++ b/config/jobs/kubernetes/sig-scalability/sig-scalability-presubmit-jobs.yaml
@@ -545,8 +545,6 @@ presubmits:
     optional: true
     decorate: true
     always_run: true
-    skip_branches:
-    - release-1.1[2-8] # Presubmit has been implemented for 1.19+ releases
     path_alias: "k8s.io/perf-tests"
     annotations:
       testgrid-dashboards: presubmits-kubernetes-scalability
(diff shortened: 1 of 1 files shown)


wojtek-t:
The "verify" scripts would have to be backported first, right?

BTW - I think the biggest value we would have from enabling cluster-loader tests on branches.

jkaniuk:
> The "verify" scripts would have to be backported first, right?

Right, just wanted to kick the tires on this one.

> BTW - I think the biggest value we would have from enabling cluster-loader tests on branches.

We also lack build and unit testing. That should be quite important as well.

wojtek-t:
> We also lack build and unit testing. That should be quite important as well.

When running clusterloader tests, we build it implicitly. It's not the whole repo, but it's the most important part.
I'm not saying we should do that - I'm saying that cl2 part is more impactful (e.g. it can break k/k release branches if we break it).

mm4tt:
Yes, we need to backport Janek changes to make it work.
I'm not sure whether it would require migrating the older release branches to go-modules. If so, then it would be non-negligible amount of work.

wojtek-t:
Good point. Let's not do that. Let's simply enable clusterloader tests on branches - this gives us good enough coverage.

jkaniuk:
Replaced by #18137

PR #18138, opened 2020-07-01
Increase resource request for kind ipv6 release blocking job

aojea:
that worked out for the same job in IPv4

https://github.com/kubernetes/kubernetes/issues/92702


BenTheElder:
OK, but we're going to have to restrict where we do this, prow.k8s.io has no capacity planning sooo

PR #18141, opened 2020-07-01
Allow /approval variant for Prow approve plugin.

wlynch:
It's common within internal Google code review to see "LGTM, Approval"
as a response to approved code reviews, so I've been somewhat conditioned to use
`/approval` with Prow, only to discover later that it didn't work.

This adds the ability for Prow to accept both `/approve` and `/approval`
as valid commands.

This also makes it easy to add other variants (e.g. `/approved`?) later
if we need to.

### DIFF
--- a/prow/plugins/approve/approve.go
+++ b/prow/plugins/approve/approve.go
@@ -40,13 +40,14 @@ const (
 	// PluginName defines this plugin's registered name.
 	PluginName = "approve"
 
-	approveCommand  = "APPROVE"
 	cancelArgument  = "cancel"
 	lgtmCommand     = "LGTM"
 	noIssueArgument = "no-issue"
 )
 
 var (
+	approveCommands = []string{"APPROVE", "APPROVAL"}
+
 	associatedIssueRegexFormat = `(?:%s/[^/]+/issues/|#)(\d+)`
 	commandRegex               = regexp.MustCompile(`(?m)^/([^\s]+)[\t ]*([^\n\r]*)`)
 	notificationRegex          = regexp.MustCompile(`(?is)^\[` + approvers.ApprovalNotificationName + `\] *?([^\n]*)(?:\n\n(.*))?`)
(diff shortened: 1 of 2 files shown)


alvaroaleman:
cool with me
/ok-to-test
/assign @fejta 
(as I can not tell if the claim of what you are using internally is accurate or not :) )

fejta:
Why not just leave an approving github review so you don't have to use /foo commands at all here? An example of my doing this: https://github.com/kubernetes/test-infra/pull/18189#pullrequestreview-444463704

I am not sure the costs associated with this sort of change are worth it:

* Nobody leaves `/approval` nor `/approve` comments inside Google. We click one of a couple buttons.
* We've tended to try and maintain a single way of doing things and expect people to learn this.
  - Humans are flexible and learn at learning a diversity of language constructs. Computers are terrible about this.
  - Once we allow this, why just these two and not also `/approved`, `/approving`? Why only English? What about what is common in other companies?
    - The marginal returns on identifying and addressing all these edge cases rapidly exceeds the costs.

But the most fundamental thing here is you should just leave an approving github review to approve, which you cannot get wrong.

To me the current behavior is clear:
* Approve by leaving an approving github review.
* Alternatively, you can leave an `/approve` comment.
* Those are the only two options (personally I would remove the case insensitivity too)


The two config knobs for this are:

https://github.com/kubernetes/test-infra/blob/49b063e089060044f9580779d77988ff248abfdc/prow/plugins/config.go#L268-L271

https://github.com/kubernetes/test-infra/blob/49b063e089060044f9580779d77988ff248abfdc/prow/plugins/config.go#L310-L312

wlynch:
> But the most fundamental thing here is you should just leave an approving github review to approve, which you cannot get wrong.

The only counterpoint I have is that the GitHub code review Approve button only shows up in the `Review Changes` reply under `/files`

image

and not under any replies in the base pull request view (this is a consequence of this PRs being represented as an Issue):

image

though perhaps this is more a feature request for GitHub to implement. :)

> Once we allow this, why just these two and not also /approved, /approving? Why only English? What about what is common in other companies?

That's fair. I'll close this out. Thanks!

PR #18145, opened 2020-07-01
Kubetest2 GCE Deployer can be configured with an arbitrary number of nodes

michaelmdresser:
Includes the cluster IP range function from original kubetest. Tested with 6 nodes, verified script output and GCP web console. Also added unit tests for verify up flags because it can be tested now.

### DIFF
--- a/kubetest2/kubetest2-gce/deployer/common.go
+++ b/kubetest2/kubetest2-gce/deployer/common.go
@@ -111,5 +111,28 @@ func (d *deployer) buildEnv() []string {
 	// kube-up and kube-down get this as a default ("kubernetes") but log-dump
 	// does not. opted to set it manually here for maximum consistency
 	env = append(env, "KUBE_GCE_INSTANCE_PREFIX=kubetest2")
+
+	// Pass through number of nodes and associated IP range. In the future,
+	// IP range will be configurable.
+	env = append(env, fmt.Sprintf("NUM_NODES=%d", d.NumNodes))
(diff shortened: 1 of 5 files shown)


rancerbeta:
lgtm

kubetest2/kubetest2-gce/deployer/common.go
```diff
@@ -111,5 +111,28 @@ func (d *deployer) buildEnv() []string {
...
+	env = append(env, fmt.Sprintf("CLUSTER_IP_RANGE=%s", getClusterIPRange(d.NumNodes)))
+
 	return env
 }
+
+// Taken from the kubetest bash (gce) deployer
+// Calculates the cluster IP range based on the no. of nodes in the cluster.
+// Note: This mimics the function get-cluster-ip-range used by kube-up script.
+func getClusterIPRange(numNodes int) string {
+	suggestedRange := "10.64.0.0/14"
```
rancerbeta, line 127: do these need to be reserved from GCP?

michaelmdresser: The 10.0.0.0/8 block is a private block so I don't believe so. These are just internal ranges and is how current kubetest already works. I don't anticipate issues unless GCP has introduced some weird behavior where IPs in private ranges have to be reserved.

rancerbeta: sgtm

rancerbeta:
lgtm with new changes

PR #18146, opened 2020-07-01
podspec decoration changed to support multiple containers

anthonyrtong:
Update decoration of pod specs to no longer assume that Pod has a single container.  
- If Pod has single container, test container name will be set automatically to `TestContainerName` (i.e. "Test").
- If Pod has multiple containers, test container names will be indexed (i.e. "Test-0", "Test-1"...)
- `InjectEntrypoint` called on all containers and generates `wrapperOptions` for each.
- `Sidecar` is given `wrapperOptions` for each container instead of for just one container.

### DIFF
--- a/prow/config/config.go
+++ b/prow/config/config.go
@@ -1264,7 +1264,7 @@ func validateJobBase(v JobBase, jobType prowapi.ProwJobType, podNamespace string
 	if err := validateAgent(v, podNamespace); err != nil {
 		return err
 	}
-	if err := validatePodSpec(jobType, v.Spec, v.Annotations["ProwMultipleContainerSupport"] == "Yes, I know what I'm doing.", v.Decorate != nil && *v.Decorate); err != nil {
+	if err := validatePodSpec(jobType, v.Spec, v.Annotations["ProwMultipleContainerSupport"] == "Yes, I know what I'm doing.", v.DecorationConfig != nil); err != nil {
 		return err
 	}
 	if err := ValidatePipelineRunSpec(jobType, v.ExtraRefs, v.PipelineRunSpec); err != nil {
--- a/prow/config/config.go
+++ b/prow/config/config.go
@@ -1264,7 +1264,7 @@ func validateJobBase(v JobBase, jobType prowapi.ProwJobType, podNamespace string
 	if err := validateAgent(v, podNamespace); err != nil {
 		return err
 	}
-	if err := validatePodSpec(jobType, v.Spec); err != nil {
+	if err := validatePodSpec(jobType, v.Spec, v.Annotations["ProwMultipleContainerSupport"] == "Yes, I know what I'm doing."); err != nil {
 		return err
 	}
 	if err := ValidatePipelineRunSpec(jobType, v.ExtraRefs, v.PipelineRunSpec); err != nil {
@@ -1282,7 +1282,12 @@ func validateJobBase(v JobBase, jobType prowapi.ProwJobType, podNamespace string
 	if err := v.UtilityConfig.Validate(); err != nil {
 		return err
 	}
-	return validateDecoration(v.Spec.Containers[0], v.DecorationConfig)
+	for i := range v.Spec.Containers {
+		if err := validateDecoration(v.Spec.Containers[i], v.DecorationConfig); err != nil {
+			return err
+		}
+	}
+	return nil
 }
 
 // validatePresubmits validates the presubmits for one repo
@@ -1852,7 +1857,7 @@ func ValidatePipelineRunSpec(jobType prowapi.ProwJobType, extraRefs []prowapi.Re
 	return nil
 }
 
-func validatePodSpec(jobType prowapi.ProwJobType, spec *v1.PodSpec) error {
+func validatePodSpec(jobType prowapi.ProwJobType, spec *v1.PodSpec, multipleContainerSupport bool) error {
 	if spec == nil {
 		return nil
 	}
@@ -1863,22 +1868,48 @@ func validatePodSpec(jobType prowapi.ProwJobType, spec *v1.PodSpec) error {
 		errs = append(errs, errors.New("pod spec may not use init containers"))
 	}
 
-	if n := len(spec.Containers); n != 1 {
+	if n := len(spec.Containers); n < 1 {
 		// We must return here to not cause an out of bounds panic in the remaining validation
+		return utilerrors.NewAggregate(append(errs, fmt.Errorf("pod spec must specify at least 1 container, found: %d", n)))
+	}
+
+	if n := len(spec.Containers); n > 1 && !multipleContainerSupport {
 		return utilerrors.NewAggregate(append(errs, fmt.Errorf("pod spec must specify exactly 1 container, found: %d", n)))
 	}
 
-	envNames := sets.String{}
-	for _, env := range spec.Containers[0].Env {
-		if envNames.Has(env.Name) {
-			errs = append(errs, fmt.Errorf("env var named %q is defined more than once", env.Name))
-		}
-		envNames.Insert(env.Name)
+	if len(spec.Containers) > 1 {
+		containerNames := sets.String{}
+		for _, container := range spec.Containers {
+			if container.Name == "" {
+				errs = append(errs, fmt.Errorf("container does not have name. all containers must have names when defining multiple containers"))
+			}
 
-		for _, prowEnv := range downwardapi.EnvForType(jobType) {
-			if env.Name == prowEnv {
-				// TODO(fejta): consider allowing this
-				errs = append(errs, fmt.Errorf("env %s is reserved", env.Name))
+			if containerNames.Has(container.Name) {
+				errs = append(errs, fmt.Errorf("container named %q is defined more than once", container.Name))
+			}
+			containerNames.Insert(container.Name)
+
+			for _, prowContainer := range decorate.PodUtilsContainerNames() {
+				if container.Name == prowContainer {
+					errs = append(errs, fmt.Errorf("container %s is a reserved for decoration", container.Name))
+				}
+			}
+		}
+	}
+
+	envNames := sets.String{}
+	for i := range spec.Containers {
+		for _, env := range spec.Containers[i].Env {
+			if envNames.Has(env.Name) {
+				errs = append(errs, fmt.Errorf("env var named %q is defined more than once", env.Name))
+			}
+			envNames.Insert(env.Name)
+
+			for _, prowEnv := range downwardapi.EnvForType(jobType) {
+				if env.Name == prowEnv {
+					// TODO(fejta): consider allowing this
+					errs = append(errs, fmt.Errorf("env %s is reserved", env.Name))
+				}
 			}
 		}
 	}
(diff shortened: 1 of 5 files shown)


prow/pod-utils/decorate/podspec.go
```diff
@@ -669,37 +677,49 @@ func decorate(spec *coreapi.PodSpec, pj *prowapi.ProwJob, rawEnv map[string]stri
...
 	spec.Volumes = append(spec.Volumes, blobStorageVolumes...)
 	if outputVolume != nil {
 		spec.Volumes = append(spec.Volumes, *outputVolume)
 	}
 
 	if len(refs) > 0 {
-		spec.Containers[0].WorkingDir = DetermineWorkDir(codeMount.MountPath, refs)
-		spec.Containers[0].VolumeMounts = append(spec.Containers[0].VolumeMounts, codeMount)
+		for i, container := range spec.Containers {
+			spec.Containers[i].WorkingDir = DetermineWorkDir(codeMount.MountPath, refs)
```
e-blackwelder, line 715: You already have access to `container`.

prow/pod-utils/decorate/podspec.go
```diff
@@ -171,7 +171,13 @@ func ProwJobToPodLocal(pj prowapi.ProwJob, buildID string, outputDir string) (*c
 
 	spec := pj.Spec.PodSpec.DeepCopy()
 	spec.RestartPolicy = "Never"
-	spec.Containers[0].Name = kube.TestContainerName
+	if len(spec.Containers) == 1 {
+		spec.Containers[0].Name = kube.TestContainerName
```
e-blackwelder, line 175: Would be nice to not handle this as a separate/special case.

prow/pod-utils/decorate/podspec.go
cjwagner: I think this is correct as is. `container` is a copy of `spec.Containers[i]` which is a struct rather than a reference in this case so changes to `container` are not reflected back to the slice.

prow/pod-utils/decorate/podspec_test.go
```diff
@@ -1893,6 +1893,360 @@ func TestProwJobToPod(t *testing.T) {
 				},
 			},
 		},
+		{
+			podName: "pod",
```
cjwagner, line 1897: This test case is for undecorated jobs with multiple containers which is something I think we should explicitly forbid during config validation. Our recommendation is that all new ProwJobs use pod utility decoration and Prow defaults to warning about undecorated jobs in `checkconfig`, we only allow jobs to be undecorated for backwards compatibility. 
If we later find that we need/want to allow undecorated jobs with multiple containers we can easily remove the restriction, but I think it will promote best practices and reduce complexity.

prow/pod-utils/decorate/podspec.go
```diff
@@ -669,37 +677,49 @@ func decorate(spec *coreapi.PodSpec, pj *prowapi.ProwJob, rawEnv map[string]stri
...
+		}
+		wrapperOptions, err := InjectEntrypoint(&spec.Containers[i], pj.Spec.DecorationConfig.Timeout.Get(), pj.Spec.DecorationConfig.GracePeriod.Get(), prefix, previous, exitZero, logMount, toolsMount)
+		if err != nil {
+			return fmt.Errorf("wrap container: %v", err)
+		}
+		wrappers = append(wrappers, *wrapperOptions)
 	}
 
-	sidecar, err := Sidecar(pj.Spec.DecorationConfig, blobStorageOptions, blobStorageMounts, logMount, outputMount, encodedJobSpec, !RequirePassingEntries, *wrapperOptions)
+	sidecar, err := Sidecar(pj.Spec.DecorationConfig, blobStorageOptions, blobStorageMounts, logMount, outputMount, encodedJobSpec, !RequirePassingEntries, wrappers...)
```
cjwagner, line 702: I would have expected corresponding changes to sidecar in order to support multiple containers, but it looks like that support is already present in some form. Could you please confirm there are unit tests for sidecar that match this use case and add or update as needed?

prow/pod-utils/decorate/podspec_test.go
```diff
@@ -1893,6 +1893,360 @@ func TestProwJobToPod(t *testing.T) {
...
+							Command: []string{"/bin/thing"},
+							Args:    []string{"some", "args"},
+							Env: []coreapi.EnvVar{
+								{Name: "MY_ENV", Value: "rocks"},
+							},
+						},
+						{
+							Image:   "tester",
+							Command: []string{"/bin/thing"},
+							Args:    []string{"some", "args"},
```
cjwagner, line 2063: Please use distinct values for the different containers in the test. If the values are the same we won't catch bugs related to order or values being unintentionally reused.

prow/pod-utils/decorate/podspec_test.go
```diff
@@ -1893,6 +1893,360 @@ func TestProwJobToPod(t *testing.T) {
...
+					},
+					GCSConfiguration: &prowapi.GCSConfiguration{
+						Bucket:       "my-bucket",
+						PathStrategy: "legacy",
+						DefaultOrg:   "kubernetes",
+						DefaultRepo:  "kubernetes",
+					},
+					GCSCredentialsSecret: "secret-name",
+					SSHKeySecrets:        []string{"ssh-1", "ssh-2"},
+					SkipCloning:          &truth,
```
cjwagner, line 2029: Please include an example that does not skip cloning as well.

stevekuznetsov:
Do we have a design document for how we expect this to work from a user's perspective? What is the `build-log`, etc?

anthonyrtong:
Here is the design doc which describes how the pod utility support for multiple containers and pipelines works:  [Multiple Container / Tekton Pipeline Support for Prow Design Doc](https://docs.google.com/document/d/1kO6CSw7Cpz6VS2dMGZY8iKudoqK3iuwdGn-bC_2Svo4/edit#heading=h.xzptrog8pyxf). It should be viewable to anyone in the kubernetes-dev google group.

Users should be able to specify multiple containers in their prowjob spec and the pod the pod utilities should upload the metadata for the job as well as a combined `build-log` for all the containers.

anthonyrtong:
ptal @cjwagner updated podspec_test

prow/config/config.go
```diff
@@ -1863,22 +1868,48 @@ func validatePodSpec(jobType prowapi.ProwJobType, spec *v1.PodSpec) error {
...
+			}
+
+			if containerNames.Has(container.Name) {
+				errs = append(errs, fmt.Errorf("container named %q is defined more than once", container.Name))
+			}
+			containerNames.Insert(container.Name)
+
+			for _, prowContainer := range decorate.PodUtilsContainerNames() {
+				if container.Name == prowContainer {
+					errs = append(errs, fmt.Errorf("container %s is a reserved for decoration", container.Name))
```
cjwagner, line 1894: I think this error could be a bit clearer. We should indicate that the *name* is reserved for decoration and direct authors to specify a different container name.

prow/config/config.go
```diff
@@ -1863,22 +1868,48 @@ func validatePodSpec(jobType prowapi.ProwJobType, spec *v1.PodSpec) error {
...
+			if container.Name == "" {
+				errs = append(errs, fmt.Errorf("container does not have name. all containers must have names when defining multiple containers"))
+			}
+
+			if containerNames.Has(container.Name) {
+				errs = append(errs, fmt.Errorf("container named %q is defined more than once", container.Name))
+			}
+			containerNames.Insert(container.Name)
+
+			for _, prowContainer := range decorate.PodUtilsContainerNames() {
```
cjwagner, line 1892: nit: It would probably be a bit more appropriate for `decorate.PodUtilsContainerNames()` to return a set rather than a slice.

prow/config/config.go
```diff
@@ -1863,22 +1868,48 @@ func validatePodSpec(jobType prowapi.ProwJobType, spec *v1.PodSpec) error {
...
+			}
 		}
-		envNames.Insert(env.Name)
+	}
 
-		for _, prowEnv := range downwardapi.EnvForType(jobType) {
-			if env.Name == prowEnv {
-				// TODO(fejta): consider allowing this
-				errs = append(errs, fmt.Errorf("env %s is reserved", env.Name))
+	envNames := sets.String{}
```
cjwagner, line 1900: This needs to be defined in the for loop since env is scoped to the container not the pod. Otherwise containers won't be able to specify an environment variable that is specified by another container.

prow/config/config.go
cjwagner: Also applies to `decorate.VolumeMounts`, `decorate.VolumeMountPaths` and `downwardapi.EnvForType`.

prow/config/config_test.go
```diff
@@ -1148,6 +1148,94 @@ func TestValidateLabels(t *testing.T) {
 	}
 }
 
+func TestValidateMultipleContainers(t *testing.T) {
+	ka := string(prowjobv1.KubernetesAgent)
+	goodSpec := v1.PodSpec{
```
cjwagner, line 1153: Please include test cases for decorated jobs. This podspec wouldn't be valid for a decorated job.

prow/config/config.go
```diff
@@ -1282,7 +1282,12 @@ func validateJobBase(v JobBase, jobType prowapi.ProwJobType, podNamespace string
 	if err := v.UtilityConfig.Validate(); err != nil {
 		return err
 	}
-	return validateDecoration(v.Spec.Containers[0], v.DecorationConfig)
+	for i := range v.Spec.Containers {
+		if err := validateDecoration(v.Spec.Containers[i], v.DecorationConfig); err != nil {
```
cjwagner, line 1286: Please include logic to forbid multiple containers if the job is not decorated.

prow/pod-utils/decorate/podspec.go
```diff
@@ -171,7 +176,13 @@ func ProwJobToPodLocal(pj prowapi.ProwJob, buildID string, outputDir string) (*c
 
 	spec := pj.Spec.PodSpec.DeepCopy()
 	spec.RestartPolicy = "Never"
-	spec.Containers[0].Name = kube.TestContainerName
+	if len(spec.Containers) == 1 {
+		spec.Containers[0].Name = kube.TestContainerName
+	} else {
+		for i := range spec.Containers {
+			spec.Containers[i].Name = fmt.Sprintf("%s-%d", kube.TestContainerName, i)
```
cjwagner, line 183: This needs to be removed now. Please add/update tests to catch this bug.

anthonyrtong:
ptal @cjwagner

prow/pod-utils/decorate/podspec.go
```diff
@@ -437,14 +450,14 @@ func processLog(log coreapi.VolumeMount, prefix string) string {
...
-	return filepath.Join(log.MountPath, fmt.Sprintf("%s-log.txt", prefix))
+	return filepath.Join(log.MountPath, fmt.Sprintf("%s-build-log.txt", prefix))
 }
 
 func markerFile(log coreapi.VolumeMount, prefix string) string {
 	if prefix == "" {
 		return filepath.Join(log.MountPath, "marker-file.txt")
 	}
-	return filepath.Join(log.MountPath, fmt.Sprintf("%s-marker.txt", prefix))
+	return filepath.Join(log.MountPath, fmt.Sprintf("%s-marker-file.txt", prefix))
```
cjwagner, line 460: nit: The changes to this function and `processLog` shouldn't be required and I don't think they improve clarity.

PR #18148, opened 2020-07-02
Refactor the experiment/autobumper tool for the future Prow staging work

chizhg:
Refactoring `experiment/autobumper` tool for Prow staging as planned in the doc - [k8s Prow auto-update with staging](https://docs.google.com/document/d/1_TdObxHL084WJbjIj83dg6-URV5wm8Oxdu-yk2WnaGY/edit?ts=5eb9b8e5#). After this change, we will be able to use this tool to bump up images versions to one of `latest`, `upstream`, `upstream-staging` and `vYYYYMMDD-deadbeef`, and we can select which files to include/exclude rather than always updating all files in the repo.

Note to reviewers:
I'm not porting the functionality related to autobumper for Gerrit repos here, as I don't want this PR to be too large, I will create a followup PR for it after this one is merged.


chaodaiG:
What's the mechanism of keeping staging prow configs in sync with prod prow configs, i.e. breaking API changes, will the staging process give false negative results when there is a breaking API change?

chizhg:
> What's the mechanism of keeping staging prow configs in sync with prod prow configs, i.e. breaking API changes, will the staging process give false negative results when there is a breaking API change?

We cannot automatically keep the staging Prow configs completely in sync with prod prow configs, like how we did for Knative prow, since the k8s prow configuration is much more complicated and has quite a few bespoke settings.

The k8s staging prow deploys all the core components required for the most critical workflows, and these configs are in sync with the production prow configs. If there is a breaking change, the person who makes the change should be responsible to update both staging and production prow configs (they need to update production prow configs anyway, and if they don't update staging prow, the staging process will be broken), in this way it will less likely cause false negatives. I will mention this process in the oncall playbook and update the doc here, after this is done.

experiment/autobumper/main.go
```diff
@@ -17,167 +17,40 @@ limitations under the License.
...
+
+	flag.BoolVar(&o.BumpProwImages, "bump-prow-images", false, "Whether to bump up version of images in gcr.io/k8s-prow/.")
+	flag.BoolVar(&o.BumpTestImages, "bump-test-images", false, "Whether to bump up version of images in gcr.io/k8s-testimages/.")
+	flag.StringVar(&o.TargetVersion, "target-version", "", "The target version to bump images version to, which can be one of latest, upstream, upstream-staging and vYYYYMMDD-deadbeef.")
+
+	flag.Var(&o.IncludedConfigPaths, "include-config-paths", "The config paths to be included in this bump, in which only .yaml files will be considered. By default all files are included.")
+	flag.Var(&o.ExcludedConfigPaths, "exclude-config-paths", "The config paths to be excluded in this bump, in which only .yaml files will be considered.")
+	flag.Var(&o.ExtraFiles, "extra-files", "The extra non-yaml file to be considered in this bump.")
+
+	flag.BoolVar(&o.CreatePullRequest, "create-pull-request", true, "Whether to create a pull request for this bump.")
```
chaodaiG, line 45: skip-pull-request makes a bit more sense to me, as the default can be false

chizhg: Makes sense. Done.

config/jobs/kubernetes/test-infra/test-infra-trusted.yaml
```diff
@@ -855,6 +855,15 @@ periodics:
       - --github-token=/etc/github-token/oauth
       - --git-name=Kubernetes Prow Robot
       - --git-email=k8s.ci.robot@gmail.com
+      - --github-org=kubernetes
```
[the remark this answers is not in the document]
chizhg: These new flags have to be included in this pull request, since the job is using `bazel run`.

chizhg:
> Left one comment on this PR, I feel it would be much better to review if this can be broken to two separate PRs. For the autobump refactoring part, it would also be better to review if you could break it into 2 commits: 1. moving parts around, 2. updates
> 
> Also, you may want to delete the binary file committed

Thanks, I have deleted the binary file.

I also realized it might be more friendly to reviewers if I split into two commits, but now it's pretty hard for me to split them..
For you convenience, line 201-265 and line 389 to the end in `bump.go` are either untouched or copied from `main.go`, so you can skip over them when you review.

experiment/autobumper/bumper/bumper.go
```diff
@@ -38,26 +42,197 @@ const (
 	testImagePrefix = "gcr.io/k8s-testimages/"
 	prowRepo        = "https://github.com/kubernetes/test-infra"
 	testImageRepo   = prowRepo
+
+	latestVersion          = "latest"
```
chaodaiG, line 46: I feel these could be enums

experiment/autobumper/bumper/bumper.go
```diff
@@ -38,26 +42,197 @@ const (
...
+	upstreamVersion        = "upstream"
+	upstreamStagingVersion = "upstream-staging"
+	tagVersion             = "vYYYYMMDD-deadbeef"
+
+	upstreamBranch           = "https://raw.githubusercontent.com/kubernetes/test-infra/master"
```
chaodaiG, line 51: Nit, this is not a branch name, rename to something like downloadingurlbase

experiment/autobumper/bumper/bumper.go
```diff
@@ -38,26 +42,197 @@ const (
...
+	if len(o.IncludedConfigPaths) == 0 {
+		return fmt.Errorf("--include-config-paths is mandatory")
+	}
+
+	return nil
+}
+
+func Run(o *Options) error {
+	if err := validateOptions(o); err != nil {
+		return fmt.Errorf("error validating options: %v", err)
```
chaodaiG, line 136: %w
And everywhere else

experiment/autobumper/bumper/bumper.go
```diff
@@ -38,26 +42,197 @@ const (
...
+			}
+			if o.GitName == "" {
+				o.GitName = user.Name
+			}
+			if o.GitEmail == "" {
+				o.GitEmail = user.Email
+			}
+		}
+
+		remoteBranch := "autobump"
```
chaodaiG, line 186: It's possible that there are multiple bumping PRs going on, i.e. oss-test-infra, can we allow more customized name?

experiment/autobumper/bumper/bumper.go
```diff
@@ -38,26 +42,197 @@ const (
...
+
+	if !changed {
+		logrus.Info("no images updated, exiting ...")
+		return nil
+	}
+
+	if o.SkipPullRequest {
+		logrus.Debugf("--skip-pull-request is set to true, won't create a pull request.")
+	} else {
+		sa := &secret.Agent{}
```
fejta, line 163: `var sa secret.Agent`

experiment/autobumper/bumper/bumper.go
```diff
@@ -261,3 +534,37 @@ func generatePRBody(images map[string]string, assignment string) string {
...
+	if req.StatusCode != http.StatusOK {
+		return "", fmt.Errorf("HTTP error %d (%q) fetching current oncaller", req.StatusCode, req.Status)
+	}
+	oncall := struct {
+		Oncall struct {
+			TestInfra string `json:"testinfra"`
+		} `json:"Oncall"`
+	}{}
+	if err := json.NewDecoder(req.Body).Decode(&oncall); err != nil {
+		return "", err
```
fejta, line 567: Please disambiguate errors when functions return an error from multiple locations.

```golang
return fmt.Errorf("get: %v", err)
return fmt.Errorf("decode response: %v", err)
```

(please apply this everywhere you are returning err directly, not just in this function)

experiment/autobumper/bumper/bumper.go
```diff
@@ -261,3 +534,37 @@ func generatePRBody(images map[string]string, assignment string) string {
...
+}
+
+func getOncaller(oncallAddress string) (string, error) {
+	req, err := http.Get(oncallAddress)
+	if err != nil {
+		return "", err
+	}
+	defer req.Body.Close()
+	if req.StatusCode != http.StatusOK {
+		return "", fmt.Errorf("HTTP error %d (%q) fetching current oncaller", req.StatusCode, req.Status)
```
fejta, line 559: IMO "fetching current oncaller" is unnecessary as it is trivial for the caller to figure this out, I'd do something more like:

`fmt.Errorf("get: HTTP error %d: %q", req.StatusCode, req.Status)`

experiment/autobumper/bumper/bumper.go
```diff
@@ -261,3 +534,37 @@ func generatePRBody(images map[string]string, assignment string) string {
 	testImagesSummary := generateSummary("test-image", testImageRepo, testImagePrefix, false, images)
 	return prowSummary + "\n\n" + testImagesSummary + "\n\n" + assignment + "\n"
 }
+
+func getAssignment(oncallAddress string) string {
+	if oncallAddress == "" {
+		return ""
+	}
+	oncaller, err := getOncaller(oncallAddress)
+	if err == nil {
```
fejta, line 543: Please indent the error flow

experiment/autobumper/bumper/bumper.go
```diff
@@ -89,28 +264,126 @@ func UpdatePullRequestWithLabels(gc github.Client, org, repo, title, body, match
...
+		return "", fmt.Errorf("error reading the response body: %v", err)
+	}
+	imagematcher := regexp.MustCompile(`(?s)^.+image:.+:(v[a-zA-Z0-9_.-]+)`)
+	res := imagematcher.FindStringSubmatch(string(body))
+	return res[1], nil
+}
+
+func isUnderPath(name string, paths []string) bool {
+	for _, p := range paths {
+		if p != "" && strings.HasPrefix(name, p) {
```
fejta, line 382: Indenting error flow would be more like:

```golang
if p == "" || !strings.HasPrefix(name, p) {
  continue
}
return true
```

experiment/autobumper/bumper/bumper.go
```diff
@@ -89,28 +264,126 @@ func UpdatePullRequestWithLabels(gc github.Client, org, repo, title, body, match
...
+			return nil, err
+		}
+	default:
+		tagPicker = func(imageHost, imageName, currentTag string) (string, error) { return tagVersion, nil }
+	}
+
+	updateFile := func(name string) error {
+		fmt.Printf("Updating file %s\n", name)
+		if err := imagebumper.UpdateFile(tagPicker, name, filterRegexp); err != nil {
+			logrus.WithError(err).Errorf("Failed to update file %q.", name)
```
fejta, line 304: Logging is unnecessary here. The caller can decide whether or not to log the error. We also do not need to pass in `name` because the caller knows this information already.

chizhg: Removed this logging line.

experiment/autobumper/bumper/bumper.go
chizhg: Done.
Since the `err` needs to be used out of the `if` block, I will have to define the two vars first:
```
var oncaller string
var err error
```

experiment/autobumper/bumper/bumper.go
chizhg: Makes sense. Changed as suggested.

experiment/autobumper/bumper/bumper.go
chizhg: Made the change in everywhere it's needed.

experiment/autobumper/bumper/bumper.go
chizhg: Done.

experiment/autobumper/bumper/bumper.go
chizhg: Good point. I have added one extra flag to allow users configure it, and the default value is `autobump`.

experiment/autobumper/bumper/bumper.go
chizhg: Done.

experiment/autobumper/bumper/bumper.go
chizhg: Changed to `upstreamURLBase`, which I feel is a better name.

experiment/autobumper/bumper/bumper.go
chizhg: I'm not sure if enums could help here, since we need to allow users to configure it via a flag, so I left this unchanged.
If you have a better approach by using enums, please let me know.

chizhg:
> Left a few comments. What's not clear to me is why there is `staging` involved in this tool, can't this be controlled by the files filter flag?

`staging` is one of the target versions. The workflow for the staging process is we update the staging Prow to the `latest` version first and then do validations, if the validations pass we update the production Prow to the `staging` version.

So `staging` will be the target version we need to update production Prow in this workflow.

experiment/autobumper/bumper/bumper.go
chizhg: I feel `if p != "" && strings.HasPrefix(name, p)` is more understandable, and this flow adds one more line, so I left this unchanged.

experiment/autobumper/bumper/bumper.go
chaodaiG: ack

chaodaiG:
> > Left a few comments. What's not clear to me is why there is `staging` involved in this tool, can't this be controlled by the files filter flag?
> 
> `staging` is one of the target versions. The workflow for the staging process is we update the staging Prow to the `latest` version first and then do validations, if the validations pass we update the production Prow to the `staging` version.
> 
> So `staging` will be the target version we need to update production Prow in this workflow.

I see. `upstream-staging` and `upstream` are ok in this case, do you think it's better to name them something like `preview` and `stable`?

chizhg:
> > > Left a few comments. What's not clear to me is why there is `staging` involved in this tool, can't this be controlled by the files filter flag?
> > 
> > 
> > `staging` is one of the target versions. The workflow for the staging process is we update the staging Prow to the `latest` version first and then do validations, if the validations pass we update the production Prow to the `staging` version.
> > So `staging` will be the target version we need to update production Prow in this workflow.
> 
> I see. `upstream-staging` and `upstream` are ok in this case, do you think it's better to name them something like `preview` and `stable`?

I agree these are better names. But before we completely roll out the staging process, I think it's better to keep the old names, since:
1. The existing autobump tool in https://github.com/kubernetes/test-infra/tree/master/prow/cmd/autobump is using `upstream`
2. The current image version for k8s Prow cannot be considered as `stable` without the staging validation

chaodaiG:
> 1. The existing autobump tool in https://github.com/kubernetes/test-infra/tree/master/prow/cmd/autobump is using `upstream`
> 2. The current image version for k8s Prow cannot be considered as `stable` without the staging validation

ack. lgtm

experiment/autobumper/bumper/bumper.go
```diff
@@ -261,3 +536,38 @@ func generatePRBody(images map[string]string, assignment string) string {
 	testImagesSummary := generateSummary("test-image", testImageRepo, testImagePrefix, false, images)
 	return prowSummary + "\n\n" + testImagesSummary + "\n\n" + assignment + "\n"
 }
+
+func getAssignment(oncallAddress string) string {
```
fejta, line 540: Thoughts on unit testing this and/or getOncaller?

experiment/autobumper/bumper/bumper.go
```diff
@@ -261,3 +536,38 @@ func generatePRBody(images map[string]string, assignment string) string {
...
+	if oncaller, err = getOncaller(oncallAddress); err == nil {
+		if oncaller != "" {
+			return "/cc @" + oncaller
+		}
+		return "Nobody is currently oncall, so falling back to Blunderbuss."
+	}
+	return fmt.Sprintf("An error occurred while finding an assignee: `%s`.\nFalling back to Blunderbuss.", err)
+}
+
+func getOncaller(oncallAddress string) (string, error) {
```
fejta, line 555: This is only ever called once, are we sure we need it?

experiment/autobumper/bumper/bumper.go
```diff
@@ -68,49 +247,145 @@ func (w HideSecretsWriter) Write(content []byte) (int, error) {
...
 	return nil
 }
 
-// UpdateReferences update the references of prow-images and testimages
-// in the files in any of "subfolders" of the current dir
+// UpdateReferences update the references of prow-images and/or testimages
+// in the files in any of "subfolders" of the includeConfigPaths but not in excludeConfigPaths
 // if the file is a yaml file (*.yaml) or extraFiles[file]=true
-func UpdateReferences(subfolders []string, extraFiles map[string]bool) (map[string]string, error) {
+func updateReferences(bumpProwImages, bumpTestImages bool, targetVersion string,
```
fejta, line 274: This seems worth unit testing

experiment/autobumper/bumper/bumper.go
```diff
@@ -68,49 +247,145 @@ func (w HideSecretsWriter) Write(content []byte) (int, error) {
...
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode != http.StatusOK {
+		return "", fmt.Errorf("HTTP error %d (%q) fetching upstream config file", resp.StatusCode, resp.Status)
+	}
+	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return "", fmt.Errorf("error reading the response body: %w", err)
+	}
+	imagematcher := regexp.MustCompile(`(?s)^.+image:.+:(v[a-zA-Z0-9_.-]+)`)
```
fejta, line 377: move this into a package-level var so we don't initialize it every call (and/or return the error)

experiment/autobumper/bumper/bumper.go
```diff
@@ -68,49 +247,145 @@ func (w HideSecretsWriter) Write(content []byte) (int, error) {
...
+	res := imagematcher.FindStringSubmatch(string(body))
+	return res[1], nil
```
fejta, line 379: res can be nil

experiment/autobumper/bumper/bumper.go
```diff
@@ -68,49 +247,145 @@ func (w HideSecretsWriter) Write(content []byte) (int, error) {
...
+		// Skip boskos images as they do not have the same image tag as other Prow components.
+		// TODO(chizhg): remove this check after all Prow instances are using boskos images not in gcr.io/k8s-prow/boskos
+		if strings.Contains(imageName, "boskos/") {
+			return currentTag, nil
+		}
+		return version, nil
+	}, nil
+}
+
+func parseUpstreamImageVersion(upstreamAddress string) (string, error) {
```
fejta, line 364: This seems worth unit testing

experiment/autobumper/bumper/bumper.go
```diff
@@ -38,26 +42,201 @@ const (
...
+
+		if err := updatePR(gc, o.GitHubOrg, o.GitHubRepo, images, getAssignment(o.OncallAddress), "Update prow to", o.GitHubLogin+":"+remoteBranch, "master", updater.PreventMods); err != nil {
+			return fmt.Errorf("failed to create the PR: %w", err)
+		}
+	}
+
+	return nil
+}
+
+func cdToRootDir() error {
```
fejta, line 205: Can we test this behavior?

experiment/autobumper/bumper/bumper.go
```diff
@@ -38,26 +42,201 @@ const (
...
+			if o.GitName == "" {
+				o.GitName = user.Name
+			}
+			if o.GitEmail == "" {
+				o.GitEmail = user.Email
+			}
+		}
+
+		remoteBranch := "autobump"
+		stdout := hideSecretsWriter{delegate: os.Stdout, censor: &sa}
```
fejta, line 191: We should test that we elide secrets correctly

experiment/autobumper/bumper/bumper.go
```diff
@@ -38,26 +42,201 @@ const (
...
+	}
+
+	if len(o.IncludedConfigPaths) == 0 {
+		return fmt.Errorf("--include-config-paths is mandatory")
+	}
+
+	return nil
+}
+
+func Run(o *Options) error {
```
fejta, line 138: This also seems like it would be useful to unit test

experiment/autobumper/bumper/bumper.go
```diff
@@ -38,26 +42,201 @@ const (
...
+	if !o.SkipPullRequest && o.RemoteBranch == "" {
+		return fmt.Errorf("--remote-branch cannot be empty when --skip-pull-request is false")
+	}
+	if (o.GitEmail == "") != (o.GitName == "") {
+		return fmt.Errorf("--git-name and --git-email must be specified together")
+	}
+
+	if o.TargetVersion != latestVersion && o.TargetVersion != upstreamVersion &&
+		o.TargetVersion != upstreamStagingVersion && !tagRegexp.MatchString(o.TargetVersion) {
+		return fmt.Errorf("invalid --target-version, must be one of %v",
```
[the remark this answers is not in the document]
chizhg: Done, changed to printing a warning.

experiment/autobumper/bumper/bumper_test.go
```diff
@@ -26,6 +26,136 @@ import (
 	"k8s.io/test-infra/prow/config/secret"
 )
 
+func Test_validateOptions(t *testing.T) {
```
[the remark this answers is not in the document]
chizhg: Fixed.

experiment/autobumper/bumper/bumper.go
chizhg: Combined it with the `getAssignment` function.

experiment/autobumper/bumper/bumper.go
chizhg: Added unit tests for this function.

experiment/autobumper/bumper/bumper.go
chizhg: Done.

experiment/autobumper/bumper/bumper.go
chizhg: Added the unit tests.

experiment/autobumper/bumper/bumper.go
chizhg: Added unit tests.

experiment/autobumper/bumper/bumper.go
chizhg: Added one extra check before this line:
```
if len(res) < 2 {
    return "", fmt.Errorf("the image tag is malformatted: %v", res)
}
``
Thanks!

experiment/autobumper/bumper/bumper.go
chizhg: There are already unit tests for it https://github.com/kubernetes/test-infra/blob/master/experiment/autobumper/bumper/bumper_test.go#L44-L115, so I didn't make changes for this comment.

experiment/autobumper/bumper/bumper.go
chizhg: This function is pretty hard to do unit testing since it will update files in place. I was thinking about adding some test files, saving their content before running this function and recovering them after test is finished, but it will be adding a lot of helper functions just for this test.

Do we have a similar test (updating a file in place and recover it after the test is finished) that I can make use of the helper functions? Or do you have other better suggestions for this?

experiment/autobumper/bumper/bumper.go
chizhg: Since this function will call the `updateReferences` function as discussed in https://github.com/kubernetes/test-infra/pull/18148#discussion_r469051656, I didn't find a good way to add unit tests for it either.

PR #18155, opened 2020-07-02
bugzilla: add new functions to identify certain bugzilla errors

AlexNPavel:
This commit adds functions to identify whether an error returned
by bugzilla is due to an invalid bug ID or is a permissions error.

/cc @stevekuznetsov

### DIFF
--- a/prow/bugzilla/client_test.go
+++ b/prow/bugzilla/client_test.go
@@ -36,8 +36,10 @@ import (
 )
 
 var (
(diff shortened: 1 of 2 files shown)


prow/bugzilla/client_test.go
```diff
@@ -84,6 +86,10 @@ func TestGetBug(t *testing.T) {
 		} else {
 			if id == 1705243 {
 				w.Write(bugData)
+			} else if id == 2 {
+				w.Write(bugAccessDenied)
+			} else if id == 3 {
```
petr-muller, line 91: I think the `2` and `3` (and possibly the valid one too) should be named constants (`deniedBugId`, `invalidBugId`) shared between here and the callsite to avoid fragility and the need to comment there.

stevekuznetsov: It's test data, so IMHO there is no utility in investing in "DRY" to make it less "fragile". Refactors here should be intentional.

PR #18157, opened 2020-07-02
PlankV2: Fix max_concurrency handling

alvaroaleman:
PlankV2 just assumed that every non-completed job is running rather
than actually checking for the Pending state. This meant that even
jobs that are kept from getting executed by the max_concurrency setting
are included in the `running` list. This causes the issue that if a
global max_concurrency setting exists and we get above that limit
and create jobs faster than they finish, we will not create any jobs
at all anymore.

### DIFF
--- a/prow/plank/reconciler.go
+++ b/prow/plank/reconciler.go
@@ -235,7 +235,7 @@ func (r *reconciler) reconcile(pj *prowv1.ProwJob) (*reconcile.Result, error) {
 
 func (r *reconciler) terminateDupes(pj *prowv1.ProwJob) error {
 	pjs := &prowv1.ProwJobList{}
-	if err := r.pjClient.List(r.ctx, pjs, optNotCompletedProwJobs()); err != nil {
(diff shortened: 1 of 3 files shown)


prow/plank/reconciler.go
```diff
@@ -235,7 +235,7 @@ func (r *reconciler) reconcile(pj *prowv1.ProwJob) (*reconcile.Result, error) {
 
 func (r *reconciler) terminateDupes(pj *prowv1.ProwJob) error {
 	pjs := &prowv1.ProwJobList{}
-	if err := r.pjClient.List(r.ctx, pjs, optNotCompletedProwJobs()); err != nil {
+	if err := r.pjClient.List(r.ctx, pjs, optPendingProwJobs()); err != nil {
```
petr-muller, line 238: Isn't the "not completed" actually correct here? I understand this is called when new job is triggered e.g. by a PR push which needs to abort existing ones. In that case we would want to remove even the previous triggers which might be still waiting in the queue, no?

alvaroaleman: Good point. I have changed it to use the `optPendingTriggeredJobsNamed` which will a) catch those as well b) only contains the jobs with the same name rather than all jobs

PR #18160, opened 2020-07-03
Allow use of vhds in aks-engine 

jsturtevant:
This helps solve issue https://github.com/kubernetes/kubernetes/issues/92771 and https://github.com/kubernetes-sigs/windows-testing/issues/177 where tests are flaking because of a node reboot.  Using a different VHD specified in the [sig-windows template](https://github.com/kubernetes-sigs/windows-testing/tree/master/job-templates) resolves this issue and speeds up the tests.

This requires a change to aks-engine to use kubectl and kubelet built from kubernetes job that @chewong is working on. 
/hold
/sig windows

### DIFF
--- a/kubetest/aksengine.go
+++ b/kubetest/aksengine.go
@@ -493,15 +493,6 @@ func (c *aksEngineDeployer) populateAPIModelTemplate() error {
 		return fmt.Errorf("No template file specified %v", err)
 	}
 
-	// set default distro so we do not use prebuilt os image
-	if v.Properties.MasterProfile.Distro == "" {
-		v.Properties.MasterProfile.Distro = "ubuntu"
-	}
-	for _, agentPool := range v.Properties.AgentPoolProfiles {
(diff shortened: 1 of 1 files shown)


kubetest/aksengine.go
```diff
@@ -493,15 +493,6 @@ func (c *aksEngineDeployer) populateAPIModelTemplate() error {
 		return fmt.Errorf("No template file specified %v", err)
 	}
 
-	// set default distro so we do not use prebuilt os image
-	if v.Properties.MasterProfile.Distro == "" {
```
feiskyer, line 497: setting distro from aks-engine.json would override the defaults here, right?

chewong: That's correct.

jsturtevant: yes that's the path we ended up taking: https://github.com/kubernetes-sigs/windows-testing/pull/178

PR #18158, opened 2020-07-02
[WIP] releng: Publish cross builds to a separate subdirectory

justaugustus:
Enables https://github.com/kubernetes/release/pull/1386 and https://github.com/kubernetes/release/pull/1387.

Signed-off-by: Stephen Augustus <saugustus@vmware.com>

/assign @BenTheElder @spiffxp 
cc: @kubernetes/release-engineering

PR #18163, opened 2020-07-03
releng: Publish fast builds to a separate subdirectory

justaugustus:
  
  (This commit matches another in #18169, so it'll get rebased out of
  whichever PR lands last.)
- kubetest: Enable extract strategy for fast (linux/amd64-only) CI builds
- scenarios: Allow kubernetes_build to recognize fast builds

Inverts the logic in https://github.com/kubernetes/test-infra/pull/18158.
Explained in more detail here: https://github.com/kubernetes/release/pull/1389

/assign @BenTheElder @spiffxp @cblecker @dims 
/hold for review
ref: https://github.com/kubernetes/sig-release/issues/850, https://github.com/kubernetes/sig-release/issues/759


BenTheElder:
I'll review this properly Monday.

I would like to request though that we settle on something and then stop moving these around, these markers are the API for our CI builds.
There's no good way to communicate moving these to all potential users and we want people to actually test these.

BenTheElder:
still not in favor of continuing to rename these markers. Doing so affects all consumers and we have no way to communicate this effectively.

where was this discussed?

spiffxp:
I've tried to follow the trail of issues and PR's and can't tell what the actual plan is, how conflicting version markers manifested, etc.  There's too much followup/revert and no "this is the desired state we are reconciling to" that I can find?

I know we chatted on slack about moving toward version markers and job names that explicitly have the version in them.  A plan of record would be useful

justaugustus:
I have no idea why the PR is pulling in a bunch of commits; tried rebasing things out, but it's not being reflected in the UI.
I'll open another PR for this.

PR #18167, opened 2020-07-03
Extend scope of kops networking e2e presubmits

johngmyers:
Extend the networking-specific presubmits to cover (most) files that currently read the respective networking spec.

The other way to go would be to run all the networking presubmits every time the model changes, but that seems excessive.

/cc @rifelpet

### DIFF
--- a/config/jobs/kubernetes/kops/kops-presubmits-network-plugins.yaml
+++ b/config/jobs/kubernetes/kops/kops-presubmits-network-plugins.yaml
@@ -4,7 +4,7 @@ presubmits:
   - name: pull-kops-e2e-cni-amazonvpc
     branches:
     - master
-    run_if_changed: '^upup\/models\/cloudup\/resources\/addons\/networking\.amazon-vpc-routed-eni\/'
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes/kops/kops-presubmits-network-plugins.yaml
```diff
@@ -160,7 +160,7 @@ presubmits:
   - name: pull-kops-e2e-cni-cilium
     branches:
     - master
-    run_if_changed: '^upup\/models\/cloudup\/resources\/addons\/networking\.cilium\.io\/'
+    run_if_changed: '^(upup\/models\/cloudup\/resources\/addons\/networking\.cilium\.io\/|pkg\/model\/(firewall|components\/cilium|iam/iam_builder).go|nodeup\/pkg\/model\/(context|networking\/cilium).go|upup\/pkg\/fi\/cloudup\/template_functions.go)'
```
rifelpet, line 163:
```suggestion
    run_if_changed: '^(upup\/models\/cloudup\/resources\/addons\/networking\.cilium\.io\/|pkg\/model\/(firewall|components\/cilium|iam\/iam_builder).go|nodeup\/pkg\/model\/(context|networking\/cilium).go|upup\/pkg\/fi\/cloudup\/template_functions.go)'
```

PR #18169, opened 2020-07-04
[WIP] releng: Add stable4 and remove beta generated jobs

justaugustus:
Part of https://github.com/kubernetes/sig-release/issues/850.

~(Needed ahead of this week's branch cut.)~

  (This commit matches another in #18163, so it'll get rebased out of
  whichever PR lands last.)

- Drop test config for beta jobs, add stable4 jobs
  
  beta markers shift depending on the stage in the release cycle and can
  end up being inaccurate without consistent care and feeding of the
  generated jobs.
  
  Here we remove the beta job configs (not the k8s-beta version marker;
  that'll happen in an upcoming commit) and establish a new `stable4`
  version config.
  
  This serves a few purposes:
  - beta releases are now cut from the main (`master`) branch, so the
    `dev` `k8sVersions`/`nodeK8sVersions` configs are now sufficient to
    describe the beta phase of the release cycle
  - now that we've expanded the support cycle for Kubernetes versions
    (from 1.19 on) to 12+ months, removing the previous jobs no longer
    falls neatly in line with the timing of the next branch cut. The
    `stable4` job configs will now allow us to maintain generated jobs
    through the newly extended support cycle.

- Regenerate jobs

- Update dashboard group/tab references

cc: @kubernetes/release-engineering @kubernetes/wg-lts


spiffxp:
/hold

>now that we've expanded the support cycle for Kubernetes versions (from 1.19 on) to 12+ months, removing the previous jobs no longer falls neatly in line with the timing of the next branch cut. The stable4 job configs will now allow us to maintain generated jobs through the newly extended support cycle.

I have not seen agreement / consensus that 1.16 is getting retroactive 12+ month support.

releng/prepare_release_branch.py
```diff
@@ -54,7 +54,7 @@ def delete_dead_branch(branch_path, current_version):
 
 
 def rotate_files(rotator_bin, branch_path, current_version):
-    suffixes = ['beta', 'stable1', 'stable2', 'stable3']
+    suffixes = ['stable1', 'stable2', 'stable3', 'stable4']
```
spiffxp, line 61: making this change without making a corresponding change to all of the non-generated jobs with the word 'beta` in them is confusing

justaugustus:
Moving this back to WIP, but rebased, so the PR doesn't fall too far behind.

> I have not seen agreement / consensus that 1.16 is getting retroactive 12+ month support.

@spiffxp -- While that's true, we have agreed on extended support for 1.19+, which means our test config needs to support that possibility, irrespective of any decisions for retroactively extending support for older branches.

Discussion is happening here and we need to close the loop on that: https://kubernetes.slack.com/archives/CDTMPM75W/p1594146739077400

justaugustus:
I'll think through this after the break.
/close

PR #18173, opened 2020-07-06
increase the time to run conformance

aojea:
the kind conformance jobs are failing because they are timing out, seems they need more than 2 hours https://testgrid.k8s.io/sig-testing-kind#conformance,%20master%20(dev)&graph-metrics=test-duration-minutes&width=20


dims:
/lgtm

band aid for now.

BenTheElder:
Thanks!

Issue #18178, opened 2020-07-06
pull-test-infra-gubernator failing


fejta:
Example: https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/test-infra/18142/pull-test-infra-gubernator/1280184099666923528

/assign

Suspect a py2 vs py3 thing

PR #18180, opened 2020-07-06
Update gubernator image

fejta:
Use apt to install gcloud.
Use a single image layer for gcloud/app-engine setup
Remove the runner

/assign @michelle192837 @cjwagner

PR #18181, opened 2020-07-06
Update gubernator tests and config

fejta:
* `update_config.sh` script needs to pass the job dir
* Teach `update_config.py` to handle empty files
* Run `update_config.sh` now that it works to update this list.

Will need an image bump (https://github.com/kubernetes/test-infra/pull/18180) in order for this to pass tests.

/assign @cjwagner @michelle192837 

fixes https://github.com/kubernetes/test-infra/issues/18178


fejta:
Will `/retest` after the image bump (should pass at that time)

fejta:
/assign @hasheddan 
in case you're still around

now passing again!

fejta:
Can I get a review please?

fejta:
🥂 Daniel!

Cole and Steve, looking you two 😇

PR #18182, opened 2020-07-06
Migrate TestGrid config from Prow -> config yaml

MushuEE:
To enable the more advanced `testgroup` config options that will allow coverage metrics to show in TestGrid Cells, I needed to move the tab/dashboard config from the prow annotations and add them into the direct TestGrid Config with a `test_group` for the one tab that displays coverage metrics.


MushuEE:
@spiffxp I agree it would be nice to extend some of the advanced features. Just not sure how the TestGroup -> Tab relationship would be handled from the annotation side.

PR #18184, opened 2020-07-06
Update sig-storage-gce-config.yaml

AndiLi99:
Increase timeout from 80mins to 120mins for job pull-kubernetes-e2e-gce-storage-snapshot.

The `ci-kubernetes-e2e-snapshot` already has a timeout of 120mins. 

Tests are failing when PR due to lack of time allotted: https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/92555/pull-kubernetes-e2e-gce-storage-snapshot/1280200772625108993/

The timeout interrupts the test suite before it can be completed.

cc: @msau42

### DIFF
--- a/config/jobs/kubernetes/sig-storage/sig-storage-gce-config.yaml
+++ b/config/jobs/kubernetes/sig-storage/sig-storage-gce-config.yaml
@@ -79,7 +79,7 @@ presubmits:
         - --repo=k8s.io/kubernetes=$(PULL_REFS)
         - --repo=k8s.io/release
         - --upload=gs://kubernetes-jenkins/pr-logs
-        - --timeout=125
+        - --timeout=150
         - --scenario=kubernetes_e2e
         - --
         - --build=bazel
@@ -91,7 +91,7 @@ presubmits:
         - --runtime-config=batch/v2alpha1=true
         - --stage=gs://kubernetes-release-pull/ci/pull-kubernetes-e2e-gce-storage-snapshot
         - --test_args=--ginkgo.focus=\[Feature:VolumeSnapshotDataSource\] --ginkgo.skip=\[Disruptive\]|\[Flaky\] --minStartupPods=8
-        - --timeout=80m
+        - --timeout=120m
         # Bazel needs privileged mode in order to sandbox builds.
         securityContext:
           privileged: true


config/jobs/kubernetes/sig-storage/sig-storage-gce-config.yaml
```diff
@@ -91,7 +91,7 @@ presubmits:
         - --runtime-config=batch/v2alpha1=true
         - --stage=gs://kubernetes-release-pull/ci/pull-kubernetes-e2e-gce-storage-snapshot
         - --test_args=--ginkgo.focus=\[Feature:VolumeSnapshotDataSource\] --ginkgo.skip=\[Disruptive\]|\[Flaky\] --minStartupPods=8
-        - --timeout=80m
+        - --timeout=120m
```
msau42, line 94: Can you also bump the top level job timeout to 150 to match the config on the serial job: https://github.com/kubernetes/test-infra/blob/3686b3f215bf471f93a5ba1bd33c0b78e7b5b9a6/config/jobs/kubernetes/sig-storage/sig-storage-gce-config.yaml#L185

msau42: And ci-kubernetes-e2e-snapshot below should also be updated

config/jobs/kubernetes/sig-storage/sig-storage-gce-config.yaml
```diff
@@ -391,7 +391,7 @@ periodics:
       - --gcp-zone=us-west1-b
       - --provider=gce
       - --test_args=--ginkgo.focus=\[Feature:VolumeSnapshotDataSource\] --ginkgo.skip=\[Disruptive\]|\[Flaky\] --minStartupPods=8
-      - --timeout=120m
+      - --timeout=160m
```
msau42, line 394: Oh sorry, I didn't realize that the ci job was already at 120. I think keeping it at 120 is fine because one difference the ci job has is it doesn't need to build kubernetes so it should be faster than the pull job.

AndiLi99: Okay, I brought it back down to what it was before.

Issue #16588, opened 2020-03-03
Multi arch for test-infra/images


lubinszARM:
This issue is for pushing multi-architecture images for test-infra/images, so that the specific jobs can be scheduled by Prow into Arm64 nodes.
 * [ ]  kubekins-test
 * [ ]  builder, google/cloud-sdk

If a typical k8s-e2e-kind-test prow job was scheduled into an Arm64 k8s cluster, a POD contains at least the following images, so we should make them to support Arm64:
 * [ ]  kubekins-e2e
 * [ ]  bootstrap
 * [ ]  sidecar
 * [ ]  initupload
 * [ ]  clonerefs

lubinszARM:
cc @BenTheElder @zhlhahaha

BenTheElder:
This seems quite a bit ahead of ourselves without any plan for Prow on such nodes.

cc @dims @mkumatag

mkumatag:
Interesting, I was thinking about this for quite some time. Can we have individual architecture's cluster running anywhere in the word and prow scheduling respective arch's job into those clusters and get things done?

BenTheElder:
I don't see why we would go to all that trouble over having jobs ssh etc.
To run test clusters on whatever arch.

We barely have people helping maintain the existing infra with relatively
self managing GKE clusters. I can't see us wanting to maintain long running
bespoke clusters on various infra on top of porting and cross building all
of these equally barely maintained CI images ...

lubinszARM:
This issue is used to track it.
We will test it in our local hybrid CI cluster firstly.
Finally we will update this issue and then decide whether to do this.
Thanks.

zhlhahaha:
Hi, @BenTheElder @mkumatag @dims , here is some update, issues and plans about prow test on ARM. Please feel free to tell me if there is anything not clear. Looking forward your suggestion.

**Update:**
- We deployed a prow cluster on x86 and a kubernetes cluster on ARM. 
- enabling test pods on ARM k8s cluster 
- Set a presubmit jobs, "pull-kubernetes-conformance-image-test", on ARM k8s cluster, it can be run successfully and all kind-conformance test passed, result has upload to [gs](https://console.cloud.google.com/storage/browser/k8s-conformance-arm/logs/test-pr/pull-kubernetes-conformance-image-test/20200319-080338-3387b4887970e1eb-7055/) 

**Issues:**
- The kubekins-e2e image for arm platform is build by ourselves.
[...]

PR #18189, opened 2020-07-07
add multi-arch support to images/alpine

lubinsz:
2 test methods have been verified:
1,   make -f images/alpine/Makefile
2, test command in prow job:
     bazel-bin/images/builder/linux_amd64_stripped/builder --project==k8s-prow  --allow-dirty=true images/alpine/

Currently, we are trying to enable arm64 prow ci for istio/kubevirt.
Both of the projects need the multi-arch images.

### DIFF
--- a/images/alpine/cloudbuild.yaml
+++ b/images/alpine/cloudbuild.yaml
@@ -1,19 +1,60 @@
 steps:
+# Step 1: Set up multiarch support.
   - name: gcr.io/cloud-builders/docker
     args:
+    - run
+    - --privileged
+    - multiarch/qemu-user-static
+    - --reset -p yes
+
+# Step 2: Create buildx builder.
+  - name: gcr.io/cloud-builders/docker
+    env:
+      - DOCKER_CLI_EXPERIMENTAL=enabled
+    args:
+    - buildx
+    - create
+    - --use
+    - --name=k8s-testimages
+
+# Step 3: Build multi-arch images, then push.
+  - name: gcr.io/cloud-builders/docker
+    env:
+      - DOCKER_CLI_EXPERIMENTAL=enabled
+    args:
+    - buildx
     - build
+    - --platform=linux/amd64,linux/arm64,linux/ppc64le,linux/s390x
+    - --push
     - -t
     - gcr.io/$PROJECT_ID/alpine:$_GIT_TAG
     - --build-arg=IMAGE_ARG=gcr.io/$PROJECT_ID/alpine:$_GIT_TAG
+    - --pull
     - .
     dir: .
+
   - name: gcr.io/cloud-builders/docker
+    env:
+      - DOCKER_CLI_EXPERIMENTAL=enabled
     args:
-    - tag
-    - gcr.io/$PROJECT_ID/alpine:$_GIT_TAG
+    - buildx
+    - build
+    - --platform=linux/amd64,linux/arm64,linux/ppc64le,linux/s390x
+    - --push
+    - -t
     - gcr.io/$PROJECT_ID/alpine:latest
+    - --build-arg=IMAGE_ARG=gcr.io/$PROJECT_ID/alpine:latest
+    - --pull
+    - .
+    dir: .
+
 substitutions:
   _GIT_TAG: '12345'
-images:
-  - 'gcr.io/$PROJECT_ID/alpine:$_GIT_TAG'
-  - 'gcr.io/$PROJECT_ID/alpine:latest'
+
+# Expected multi-arch images:
+#   gcr.io/$PROJECT_ID/alpine:$_GIT_TAG with multi-arch manifest.
+#   gcr.io/$PROJECT_ID/alpine:latest with multi-arch manifest.
+# Notice:
+# Multi-arch images can't be loaded into the local docker daemon with 'buildx --load' parameter.
+# The feature should get support for loading multi-arch with moby/moby#38738.
+# So the images field can not be used for the multi-arch case.


lubinsz:
Related issue: https://github.com/kubernetes/test-infra/issues/16588

images/alpine/cloudbuild.yaml
```diff
@@ -1,19 +1,40 @@
 steps:
-  - name: gcr.io/cloud-builders/docker
-    args:
-    - build
```
fejta, line 4: Will you please preserve this method of declaring lists?

images/alpine/cloudbuild.yaml
```diff
@@ -1,19 +1,40 @@
...
+    '-t', 'gcr.io/$PROJECT_ID/alpine:latest',
+    '--build-arg', 'IMAGE_ARG=gcr.io/$PROJECT_ID/alpine:latest',
+    '--pull', '.']
+    dir: .
+    waitFor: ['buildx-images-tag']
+
 substitutions:
   _GIT_TAG: '12345'
-images:
-  - 'gcr.io/$PROJECT_ID/alpine:$_GIT_TAG'
```
fejta, line 18: Please preserve the expected images to publish.

images/alpine/cloudbuild.yaml
```diff
@@ -1,19 +1,40 @@
...
-    - build
-    - -t
-    - gcr.io/$PROJECT_ID/alpine:$_GIT_TAG
-    - --build-arg=IMAGE_ARG=gcr.io/$PROJECT_ID/alpine:$_GIT_TAG
-    - .
+# Set up multiarch support.
+  - name: 'gcr.io/cloud-builders/docker'
+    args: ['run', '--rm', '--privileged', 'multiarch/qemu-user-static', '--reset', '-p', 'yes']
+    id: 'qemu'
+    waitFor: ['-']
```
fejta, line 6: Looks like you are defining a set of sequential steps. This is the default behavior. Thus please drop the unnecessary waitFor items.

images/alpine/cloudbuild.yaml
```diff
@@ -1,19 +1,40 @@
...
-    args:
-    - build
-    - -t
-    - gcr.io/$PROJECT_ID/alpine:$_GIT_TAG
-    - --build-arg=IMAGE_ARG=gcr.io/$PROJECT_ID/alpine:$_GIT_TAG
-    - .
+# Set up multiarch support.
+  - name: 'gcr.io/cloud-builders/docker'
+    args: ['run', '--rm', '--privileged', 'multiarch/qemu-user-static', '--reset', '-p', 'yes']
+    id: 'qemu'
```
fejta, line 5: The `id`s are not necessary (perhaps they are useful for debugging but we don't currently do anything with them.)

images/alpine/cloudbuild.yaml
lubinsz: Done.

images/alpine/cloudbuild.yaml
lubinsz: 'id' has been removed.

images/alpine/cloudbuild.yaml
lubinsz: 'waitFor' has been removed.

images/alpine/cloudbuild.yaml
lubinsz: Hi,
In step 3, those images with fat-manifest will be pushed into registry with the 'buildx --push'.
I have tested the build methods(make & prowjob) with my private gcloud account, the multi-arch images can be pushed into my private gcloud registry automatically.

On the other hand, we can't load the multi-arch images into the local docker daemon with the 'buildx --load'.
Please see the issue as reference:
https://github.com/docker/buildx/issues/59
If we keep this images field, the build will report an error

fejta: Thanks, this is my first experience with buildx and multiarch images (thanks so much for working on this).

And if the push fails then the build fails? How about then adding a comment at the bottom about what images we expect this to push and why we don't use cloud build's auto verification?

lubinszARM: > And if the push fails then the build fails? 
Yes.
> How about then adding a comment at the bottom about what images we expect this to push and why we don't use cloud build's auto verification?
The comment has been added.

lubinszARM:
Hi @dims @BenTheElder 
Any comments?
Thanks.

images/alpine/cloudbuild.yaml
```diff
@@ -1,19 +1,60 @@
...
+    - --name=k8s-testimages
+
+# Step 3: Build multi-arch images, then push.
+  - name: gcr.io/cloud-builders/docker
+    env:
+      - DOCKER_CLI_EXPERIMENTAL=enabled
+    args:
+    - buildx
     - build
+    - --platform=linux/amd64,linux/arm64,linux/ppc64le
```
BenTheElder, line 27: Any reason for this particular list of platforms?

lubinsz: Hi,
At present, kubevirt/istio is ready to do the prow ci for these 3 platforms: x86, powerpc, arm64.
And images/alpine is the basic image for other images, such as: git/sidecar/initupload...

zhlhahaha: > Any reason for this particular list of platforms?

Hi, Ben,
We are trying to enable CI test for kubevirt/kubeflow/istio project on ARM. And facing a problem that in order to trigger a prow job on ARM cluster, prow utility images need to run on ARM cluster. So we are wandering if we can make at least these utility images (sidecar/initupload/entrypoint/clonerefs) support multi-arch. We have blocked here for really a long time.
Thanks a lot!

lubinszARM: @BenTheElder 
Any comments?

fejta:
I'm inclined to merge this unless there are any outstanding issues?

srajmane:
We are looking to add s390x support to the multiarch image. While this PR is yet to be merged, 
@lubinsz  could you please help in extending support for s390x as well through this PR by adding linux/s390x to --platform=linux/amd64,linux/arm64,linux/ppc64le?

lubinszARM:
> linux/s390x

OK.

fejta (approved):

zhlhahaha (approved):

PR #18192, opened 2020-07-07
Fix failing kubectl skew periodic e2e tests

brianpursley:
**The Problem**
[Some of the periodic kubectl e2e skew tests are failing 100%](https://storage.googleapis.com/k8s-gubernator/triage/index.html?#acb0ebd30035e7aaee80)

**The Cause:**
The tests are running an older version of the test suite using a newer version of kubectl, which has had some deprecated features removed (ie. kubectl run no longer supports generators).  

The older tests are attempting to do something that the newer kubectl no longer supports.

**The Fix:**
When `--kubectl_path` is used to override the kubectl location, we should also use the `--skew` flag so kubetest runs the test suite from the kubernetes_skew directory and not from the kubernetes directory.

This PR adds the `--skew` flag whenever kubectl is overridden to run from kubernetes_skew.

Fixes https://github.com/kubernetes/kubernetes/issues/92734

### DIFF
--- a/config/jobs/kubernetes/sig-cli/sig-cli-config.yaml
+++ b/config/jobs/kubernetes/sig-cli/sig-cli-config.yaml
@@ -222,6 +222,7 @@ periodics:
       - --gcp-zone=us-west1-b
       - --ginkgo-parallel=25
       - --provider=gce
+      - --skew
       - --test_args=--ginkgo.focus=Kubectl --ginkgo.skip=\[Serial\] --kubectl-path=../../../../kubernetes_skew/cluster/kubectl.sh --minStartupPods=8
       - --timeout=120m
       image: gcr.io/k8s-testimages/kubekins-e2e:v20200623-2424179-master
@@ -249,6 +250,7 @@ periodics:
       - --gcp-node-image=gci
       - --gcp-zone=us-west1-b
       - --provider=gce
+      - --skew
       - --test_args=--ginkgo.focus=Kubectl.*\[Serial\]  --kubectl-path=../../../../kubernetes_skew/cluster/kubectl.sh --minStartupPods=8
       - --timeout=120m
       image: gcr.io/k8s-testimages/kubekins-e2e:v20200623-2424179-master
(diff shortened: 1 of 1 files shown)


brianpursley:
Putting this in WIP. I want to make sure I understand if this is the right way to handle this and if there is a way to test this locally first.

spiffxp:
/hold

Thank you for looking into this.  The skew stuff hurts my head every time.  If I go off of the description for this test config:

>stable1 e2e tests run against a stable1 gce cluster using a master kubectl binary

It sounds like running the old test suite with a new kubectl is actually the intent of this job, so I'm not sure about your fix to do the opposite.

If we run new tests against an old cluster, we risk testing functionality the old cluster doesn't have.

Can the 1.18 tests be rewritten to use whatever non-deprecated flags would exercise similar functionality instead?

spiffxp:
/cc @liggitt 
if you have any opinions on path forward

liggitt:
if we expect to run n-1 tests against version n binaries, the only option I see is to make the n-1 tests stop using deprecated things that will be removed in version n

/cc @soltysh @seans3

brianpursley:
Would it not work as long as the test suite matches the kubectl version?

cluster = n
kubectl = n-1
tests = n-1

Currently I think it is trying to do this, which doesn't work:

cluster = n
kubectl = n-1
tests = n

spiffxp:
These tests are verifying this: https://kubernetes.io/docs/setup/release/version-skew-policy/#kubectl

That is, kubectl 1.19 should be able to talk to kubernetes 1.18.  If we use 1.19's test suite against a 1.18 cluster, we may use kubectl verbs/features that are unsupported by 1.18.

So like @liggit said above, I think the 1.18 tests need to be updated to stop using the deprecated flag.

One option to help with this pattern in general might be to have a `[Deprecated]` tag for tests that are exercising known deprecated functionality.  I could then see the case made that a skew test could filter out deprecated functionality to avoid tripping over removal boundaries like this.

brianpursley:
OK, so let me close this PR and see if I can do something on the testing side. I like the idea of marking tests as `[Deprecated]` and then skipping `[Deprecated]` when skew testing w/ a newer version of kubectl.

brianpursley:
Sorry guys, I should have just kept this PR open, but I couldn't re-open after force pushing my branch :facepalm: 

Here is a PR that lets skew tests that use newer kubectl skip any test that is marked with `[Deprecated]`:
https://github.com/kubernetes/test-infra/pull/18213

PR #18195, opened 2020-07-07
Update testing to Py3 image

scottilee:
Start migrating to the Python 3 test-worker image.

/assign @jlewi

### DIFF
--- a/config/jobs/kubeflow/kubeflow-presubmits.yaml
+++ b/config/jobs/kubeflow/kubeflow-presubmits.yaml
@@ -118,7 +118,7 @@ presubmits:
       preset-service-account: "true"
     spec:
       containers:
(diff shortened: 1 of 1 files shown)


config/jobs/kubeflow/kubeflow-presubmits.yaml
```diff
@@ -118,7 +118,7 @@ presubmits:
       preset-service-account: "true"
     spec:
       containers:
-      - image: gcr.io/kubeflow-ci/test-worker:latest
+      - image: gcr.io/kubeflow-ci/test-worker-py3
```
jlewi, line 121: Actually do we have an image tagged latest?

jlewi: Added the tag

PR #18196, opened 2020-07-07
Run long tests in post-submit integration tests

liggitt:
/hold for https://github.com/kubernetes/kubernetes/pull/92879

/cc @BenTheElder

### DIFF
--- a/config/jobs/kubernetes/sig-testing/integration.yaml
+++ b/config/jobs/kubernetes/sig-testing/integration.yaml
@@ -56,6 +56,9 @@ periodics:
       - runner.sh
       args:
       - ./hack/jenkins/test-dockerized.sh
+      env:
+      - name: SHORT
+        value: --short=false
       # docker-in-docker needs privileged mode
       securityContext:
         privileged: true


BenTheElder:
/hold cancel
https://github.com/kubernetes/kubernetes/pull/92879 is in

roycaihw:
This change seems to be related to the recent failure of https://testgrid.k8s.io/sig-release-master-blocking#integration-master

Starting from [this run](https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-integration-master/1282181163326115840), two tests (`ipamperf` and `scheduler_perf`) were added to the job and were constantly failing. The time seems to match this commit. Comparing the `prowjob.json` against the previous job run, the [new run](https://storage.googleapis.com/kubernetes-jenkins/logs/ci-kubernetes-integration-master/1282181163326115840/prowjob.json) had the `SHORT` env while the [old run](https://storage.googleapis.com/kubernetes-jenkins/logs/ci-kubernetes-integration-master/1282165815772188673/prowjob.json) didn't. 

Note that the failure was reported in https://github.com/kubernetes/kubernetes/issues/92700#issuecomment-657606787, but I feel that original issue was unrelated.

roycaihw:
Edit: the added tests are failing, but the previous comment about etcd was red-herring. The "etcd already running at http://127.0.0.1:2379" is not a error message. I noticed that both tests failed after ~600s (600.522s and 600.963s). It's likely they timed-out

hasheddan:
Thanks @roycaihw! Sounds like we either need to disable these tests or more likely increase their timeout. I am having trouble finding a reference on any other jobs for how long they typically run. Any thoughts @liggitt?

liggitt:
if those tests were not being run anywhere and time out when run in CI, skipping them and opening an issue for sig-scheduling to figure out how to reintroduce them into CI would make sense to me

hasheddan:
@liggitt ack, will have PR open shortly 👍

Issue #17069, opened 2020-04-02
go.k8s.io/triage is serving stale data


spiffxp:
**What happened**:
https://go.k8s.io/triage is showing
"2216 clusters of 92485 failures (6433 in last day) out of 149213 builds from 3/18/2020, 5:00:24 PM to 3/30/2020, 10:54:32 PM."

**What you expected to happen**:
The end date should be today (04/02/2020)

**Please provide links to example occurrences, if any**:
https://storage.googleapis.com/k8s-gubernator/triage/index.html?pr=1

/area triage
/assign

spiffxp:
https://testgrid.k8s.io/sig-testing-misc#triage is all green, so I'm guessing kettle has stopped feeding data to the bigquery dataset we're using

spiffxp:
Last log entries... seems like it just got stuck?
```
spiffxp@spiffxp-macbookpro:~$ k logs pod/kettle-5df45c4dcb-fqdlf | tail -n10

==== 2020-03-30 22:55:11 PDT ========================================
PULLED 658
ACK irrelevant 655
EXTEND-ACK  3
gs://kubernetes-jenkins/logs/ci-cri-containerd-e2e-gci-gce-serial/1244861611756228612 True True 2020-03-30 22:38:48 PDT FAILURE
gs://kubernetes-jenkins/logs/ci-test-infra-benchmark-demo/1244865638594252800 True True 2020-03-30 22:54:54 PDT SUCCESS
gs://kubernetes-jenkins/logs/ci-kubernetes-e2e-gci-gce-flaky-repro/1244857333914275840 True True 2020-03-30 22:21:44 PDT FAILURE
ACK "finished.json" 3
Downloading JUnit artifacts.
```

Not out of disk space
```
spiffxp@spiffxp-macbookpro:~$ k exec -ti pod/kettle-5df45c4dcb-fqdlf df
Filesystem      1K-blocks      Used Available Use% Mounted on
overlay          98868448   6373144  92478920   7% /
tmpfs               65536         0     65536   0% /dev
tmpfs            15439876         0  15439876   0% /sys/fs/cgroup
/dev/sdb       1033024192 537385356 494573876  53% /data
/dev/sda1        98868448   6373144  92478920   7% /etc/hosts
tmpfs            15439876         4  15439872   1% /etc/service-account
shm                 65536         0     65536   0% /dev/shm
tmpfs            15439876        12  15439864   1% /run/secrets/kubernetes.io/serviceaccount
[...]

PR #18197, opened 2020-07-07
Extend Kettle build fields to be used for determining flakes

MushuEE:
#14643

Based on @spiffxp 's proposal for enabling flake detection on POD-Util jobs I am going through the process of adding new build fields. This will require a redeploy of kettle.

Also some boy-scouting of code


kettle/make_json.py
```diff
@@ -146,6 +147,8 @@ def row_for_build(path, started, finished, results):
         build['started'] = int(started['timestamp'])
         if 'node' in started:
             build['executor'] = started['node']
+        build['repo_commit'] = started.get('repo-commit', started.get('repo-version'))
+        build['repos'] = started.get('repos')
```
spiffxp, line 151: I recognize there is a dearth of comments here already, so you're just following the pattern.  But I think it might be useful for future maintainers to comment where these are expected to come from, eg: a link back to the testgrid metadata struct or something

kettle/make_json.py
```diff
@@ -176,11 +183,10 @@ def get_metadata():
...
                     # the schema specifies a string value. force it!
                     metadata[key] = json.dumps(value)
-        if not metadata:
-            return None
-        return [{'key': k, 'value': v} for k, v in sorted(metadata.items())]
+            metapairs = [{'key': k, 'value': v} for k, v in sorted(metadata.items())]
+        return metapairs
 
-    metadata = get_metadata()
+    metadata = get_metadata(build)
```
spiffxp, line 189: not sure I understand why this refactor?

kettle/make_json.py
```diff
@@ -157,13 +160,17 @@ def row_for_build(path, started, finished, results):
...
         metadata = None
+        metapairs = None
         if finished and 'metadata' in finished:
             metadata = finished['metadata']
         elif started:
             metadata = started.get('metadata')
+
         if metadata:
+            if not build.get('repos'):
+                build['repos'] = metadata.get('repos')
```
spiffxp, line 173: updating `build` in a function named `get_metadata` feels surprising, maybe we could do this after the `get_metadata` call below?

kettle/make_json.py
MushuEE: +1 I will try to make some back reference.

kettle/make_json.py
MushuEE: So I didn't like doing this but I want to access the metadata `repos` key. I have access to all key-value pairs outside of this internal method but I would have to iterate over them because I no longer have a direct map to the `repos` value. The quickest solution I saw was to edit the map object while handling the metadata.

kettle/make_json.py
MushuEE: Yeah ^ above comment kinda covers it. Just feels a little ugly to iterate over the list of pairs. I could return both the list AND the repos from get_metadata and assign that second value to `build['repos']`. Also in your doc you mention that Repos needs to be a nullable string. Do I need to join these repos with commas?

MushuEE: I majorly refactored. I decided to stay on the safe side and just remove all keys for `None` attrs so that there isn't anything crazy that might happen with nullable fields. I think this cleans things up and creates a structure for the build object so it is easier to map to the TestGrid Schema

spiffxp:
I'm going to caution that right now, restarting kettle takes for. ev. er.  So while part of me wants to say we should redeploy to verify this works, another part of me wants to suggest holding off until you've take a look at what startup involves.  Your call.

spiffxp:
You can see https://github.com/kubernetes/test-infra/issues/17069 for a depiction of what restarting looked like the last time we had to

alvaroaleman:
@MushuEE could you in the future please try to avoid merging a lot of tiny fixup commits that are not useful in the `master` branch history by either:
* Squashing them as described e.G. here https://github.com/wprig/wprig/wiki/How-to-squash-commits
* Telling Tide to squash them on merge by commenting with `/label tide/merge-method-squash`

MushuEE:
@alvaroaleman, yes thank you. I will save these options in my notes.

PR #18202, opened 2020-07-07
Make a bazel 2.1.0->2.2.0 image for testgrid

spiffxp:
Following https://github.com/kubernetes/test-infra/tree/master/images#updating-test-infra-images but for testgrid

To support https://github.com/GoogleCloudPlatform/testgrid/pull/141 if we decide we only want to upgrade to bazel 2.2.0


fejta:
(I'm supportive of updating to whatever the latest we can get working whenever we update)

PR #18203, opened 2020-07-07
Remove references to Velodrome and clean up old query filters

MushuEE:
Remove `JQ filter to make influxdb timeseries data points for Velodrome.`

Velodrome no longer exists and having these `measurement` queries in the config might just lead to confusion. I don't want to rip it out of `bigquery.py` in the case that another influx db gets spun in the future that we want to use. The `README` didn't even specify the correct query structure as it was.


MushuEE:
@spiffxp happy to nuke that code. I was hesitant just because I though there might be an other use for that in the future.

metrics/bigquery.py
```diff
@@ -273,11 +186,6 @@ def main(configs, project, bucket_path, backfill_days):
     PARSER.add_argument(
         '--bucket',
         help='Upload results to the specified gcs bucket.')
-    PARSER.add_argument(
-        '--backfill-days',
-        default=30,
-        type=int,
-        help='Number of days to backfill influxdb data.')
```
spiffxp, line 280: I think we want to keep this, I would just alter the description to avoid mentioning influxdb.  We're still passing `backfill_days` to main

MushuEE: good catch

MushuEE: Looks like that is just bing pipped though to BigQuery object and never used. I will clear them.

PR #18208, opened 2020-07-08
Address a auth race condition.

fejta:
Also add --token-path for easier local development (no need to insert this token into a cookie).

When gerrit starts up it often sees transient unauthorized errors.
This is likely because we add authentication to the handlers in a go routine.
This change makes it set up all the handlers before forking a routine to keep
the auth up to date.

Also remove some pointless functions.

### DIFF
--- a/prow/gerrit/client/client.go
+++ b/prow/gerrit/client/client.go
@@ -23,6 +23,7 @@ import (
 	"io/ioutil"
 	"sort"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/andygrunwald/go-gerrit"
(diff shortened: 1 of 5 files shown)


prow/gerrit/client/client.go
```diff
@@ -159,47 +163,79 @@ func NewClient(instances map[string][]string) (*Client, error) {
...
+			// TODO(fejta): listen for changes
+			raw, err := ioutil.ReadFile(cookiefilePath)
+			if err != nil {
+				return "", fmt.Errorf("read cookie: %w", err)
+			}
+			fields := strings.Fields(string(raw))
+			token := fields[len(fields)-1]
+			return token, nil
+		}
+	case tokenPath != "":
```
amwat, line 215: nit: do you want the override to be the first case?

fejta: Good point, looks like there's more stuff to futz with so will address in a followup

PR #18210, opened 2020-07-08
Default values in github event server options if they are nil to avoid panics

droslean:
Defaults the empty fields in the githubeventserver options to avoid panics when they are nil.

Signed-off-by: Nikolaos Moraitis nmoraiti@redhat.com

### DIFF
--- a/prow/githubeventserver/options.go
+++ b/prow/githubeventserver/options.go
@@ -44,11 +44,24 @@ type Options struct {
 	port int
 }
 
-// Validate validates the option's values.
-func (o *Options) Validate() error {
+// DefaultAndValidate validates the option's values and defaults them if they are empty.
+func (o *Options) DefaultAndValidate() error {
 	if !strings.HasPrefix(o.endpoint, "/") {
 		return fmt.Errorf("endpoint %s is not a valid url path", o.endpoint)
 	}
+
+	if o.HmacTokenGenerator == nil {
+		o.HmacTokenGenerator = func() []byte { return []byte("") }
+	}
+
+	if o.Metrics == nil {
+		o.Metrics = NewMetrics()
+	}
+
+	if o.Logger == nil {
+		o.Logger = logrus.NewEntry(logrus.New())
+	}
+
 	return nil
 }
 


prow/githubeventserver/options.go
```diff
@@ -49,6 +49,11 @@ func (o *Options) Validate() error {
 	if !strings.HasPrefix(o.endpoint, "/") {
 		return fmt.Errorf("endpoint %s is not a valid url path", o.endpoint)
 	}
+
+	o.hmacTokenGenerator = func() []byte { return []byte("") }
```
alvaroaleman, line 53: This is defaulting, not validation. `Validate` will now unconditinally overwrite all existing options. This doesn't make any sense.

prow/githubeventserver/options.go
```diff
@@ -29,14 +29,14 @@ import (
 type Options struct {
 	// HmacTokenGenerator is a function that holds a hmac token that will be used
 	// in a github event server
-	HmacTokenGenerator func() []byte
+	hmacTokenGenerator func() []byte
```
alvaroaleman, line 32: What do we gain from unexporting these and using setters?

prow/githubeventserver/options.go
droslean: This is just an initialization, to avoid those fields to be nil.

prow/githubeventserver/options.go
droslean: I don't think that those fields should be exposed. I would prefer to allow to user to set his values with a helper function instead. Also, those values will be added by the user, so technically they are already exposed to the user.

prow/githubeventserver/options.go
alvaroaleman: No. You set options, then you validate them. With what you do here, doing that will just overwrite everything a user set unconditionally. This doesn't make any sense.

prow/githubeventserver/options.go
alvaroaleman: Sorry I don't see any value in having a setter like this that only exists to write unconditionally to a private field. Having this setter doesn't make the field any less exposed, it just makes it less convenient to set.

prow/githubeventserver/options.go
droslean: In the validation point, the user won't set anything. The option flags are setting `endpoint` and `port` only. The rest stays nil. I am fine with that and rely on the user to set them later, so I can remove this change. WDYT?

alvaroaleman: Just make the fields pointers, check if they are nil, only default them if they are nil, rename the func to `DefaultAndValidate`

prow/githubeventserver/options.go
```diff
@@ -45,10 +45,23 @@ type Options struct {
 }
 
 // Validate validates the option's values.
-func (o *Options) Validate() error {
+func (o *Options) DefaultAndValidate() error {
```
alvaroaleman, line 48: godoc has the wrong name still :upside_down_face:

PR #18213, opened 2020-07-08
Skip tests marked as [Deprecated] when running skew tests

brianpursley:
What do you guys think about this one? It just adds the ability to skip tests marked with [Deprecated] when running skew tests that use a newer version of kubectl.

I'll do another PR to annotate the tests in the k/k release-1.17 branch which in combination with this should then allow these periodic e2e tests to pass.

---

**The Problem**
[Some of the periodic kubectl e2e skew tests are failing 100%](https://storage.googleapis.com/k8s-gubernator/triage/index.html?#acb0ebd30035e7aaee80)

**The Cause:**
The tests are running an older version of the test suite using a newer version of kubectl, which has had some deprecated features removed (ie. kubectl run no longer supports generators).  

The older tests are attempting to do something that the newer kubectl no longer supports.

**The Fix:**
This PR adds `--ginkgo-skip=[Deprecated]` for skew tests that override the kubectl path to use a newer version of kubectl. This will allow tests that use deprecated features to be marked as such and skipped when running skew tests that would otherwise fail.

Fixes https://github.com/kubernetes/kubernetes/issues/92734

---

This is a second attempt at fixing the problem. For reference, a previously closed PR (https://github.com/kubernetes/test-infra/pull/18192) was the first attempt.

(I probably should have just kept that one open, but oh well. I couldn't re-open it after force-pushing to my branch).

---

CC'ing the people from the other PR...

/cc @liggitt @spiffxp @soltysh @seans3 @dims

### DIFF
--- a/config/jobs/kubernetes/sig-cli/sig-cli-config.yaml
+++ b/config/jobs/kubernetes/sig-cli/sig-cli-config.yaml
@@ -103,7 +103,7 @@ periodics:
       - --ginkgo-parallel
       - --provider=gce
       - --skew
-      - --test_args=--ginkgo.focus=Kubectl --ginkgo.skip=\[Serial\] --minStartupPods=8
+      - --test_args=--ginkgo.focus=Kubectl --ginkgo.skip=\[Serial\]|\[Deprecated\] --minStartupPods=8
       - --timeout=120m
       image: gcr.io/k8s-testimages/kubekins-e2e:v20200623-2424179-master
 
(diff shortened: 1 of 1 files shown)


brianpursley:
Holding for comments...

/hold

brianpursley:
Hmm, I see what you're getting at.

If I am testing kubectl 1.18 vs cluster 1.19, but something in kubectl 1.19 was removed, shouldn't I still expect the full functionality of kubectl 1.18 to still work?

I guess I'm unclear on whether when something in kubectl is deprecated if that applies to the kubectl binary or to the overall release (kubectl + cluster).

brianpursley:
@spiffxp I updated the PR so that all skew tests will skip tests that have `[Deprecated]`.

ptal

PR #18217, opened 2020-07-08
Cache bugs for getRecursiveClones

GeoBK:
Improve performance of the GetAllClones function by reusing the
bug details fetched by getRoot in getRecursiveClones.

### DIFF
--- a/prow/bugzilla/client.go
+++ b/prow/bugzilla/client.go
@@ -24,11 +24,14 @@ import (
 	"net/http"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/sirupsen/logrus"
+	"golang.org/x/sync/errgroup"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/apimachinery/pkg/util/sets"
 )
 
 const (
@@ -75,25 +78,77 @@ type Client interface {
 	RemovePullRequestAsExternalBug(id int, org, repo string, num int) (bool, error)
 	// GetAllClones returns all the clones of the bug including itself
 	// Differs from GetClones as GetClones only gets the child clones which are one level lower
-	GetAllClones(bug *Bug) ([]*Bug, error)
+	// GetAllClones will also fetch bugDetails and the root bug details in addition to clones
+	// since we want to parallelize all the HTTP calls
+	GetAllClones(bugID int) (clones []*Bug, bug *Bug, root *Bug, err error)
 	// GetRootForClone returns the original bug.
 	GetRootForClone(bug *Bug) (*Bug, error)
+	// GetCachedClones will fetch the clones from the clones cache
+	GetCachedClones(bugID int) ([]int, bool)
+	// SetCachedClones will push the cloneIDs associated with the bug to the cache
+	SetCachedClones(bugID int, clonesList []int)
 }
 
 func NewClient(getAPIKey func() []byte, endpoint string) Client {
 	return &client{
-		logger:    logrus.WithField("client", "bugzilla"),
-		client:    &http.Client{},
-		endpoint:  endpoint,
-		getAPIKey: getAPIKey,
+		logger:      logrus.WithField("client", "bugzilla"),
+		client:      &http.Client{},
+		endpoint:    endpoint,
+		getAPIKey:   getAPIKey,
+		clonesCache: clonesCacheWithMutex{cache: map[int][]int{}},
 	}
 }
 
+type clonesCacheWithMutex struct {
+	cache map[int][]int
+	lock  sync.Mutex
+}
+
+func (cc *clonesCacheWithMutex) Get(key int) ([]int, bool) {
+	cc.lock.Lock()
+	defer cc.lock.Unlock()
+	clones, ok := cc.cache[key]
+	return clones, ok
+}
+
+func (cc *clonesCacheWithMutex) Set(key int, value []int) {
+	cc.lock.Lock()
+	defer cc.lock.Unlock()
+	cc.cache[key] = value
+}
+
+// NewBugDetailsCache is a constructor for bugDetailsCache
+func NewBugDetailsCache() *BugDetailsCache {
+	return &BugDetailsCache{cache: map[int]*Bug{}}
+}
+
+// BugDetailsCache holds the already retrieved bug details
+type BugDetailsCache struct {
+	cache map[int]*Bug
+	lock  sync.Mutex
+}
+
+// Get retrieves bug details from the cache and is thread safe
+func (bd *BugDetailsCache) Get(key int) (*Bug, bool) {
+	bd.lock.Lock()
+	defer bd.lock.Unlock()
+	bug, ok := bd.cache[key]
+	return bug, ok
+}
+
+// Set stores the bug details in the cache and is thread safe
+func (bd *BugDetailsCache) Set(key int, value *Bug) {
+	bd.lock.Lock()
+	defer bd.lock.Unlock()
+	bd.cache[key] = value
+}
+
 type client struct {
-	logger    *logrus.Entry
-	client    *http.Client
-	endpoint  string
-	getAPIKey func() []byte
+	logger      *logrus.Entry
+	client      *http.Client
+	endpoint    string
+	getAPIKey   func() []byte
+	clonesCache clonesCacheWithMutex
 }
 
 // the client is a Client impl
@@ -127,29 +182,37 @@ func (c *client) GetBug(id int) (*Bug, error) {
 	return parsedResponse.Bugs[0], nil
 }
 
-func getClones(c Client, bug *Bug) ([]*Bug, error) {
+func getClones(c Client, bug *Bug, bugCache *BugDetailsCache) ([]*Bug, error) {
 	var errs []error
 	clones := []*Bug{}
 	for _, dependentID := range bug.Blocks {
-		dependent, err := c.GetBug(dependentID)
-		if err != nil {
-			errs = append(errs, fmt.Errorf("Failed to get dependent bug #%d: %v", dependentID, err))
-			continue
+
+		dependent, ok := bugCache.Get(dependentID)
+		if !ok {
+			var err error
+			dependent, err = c.GetBug(dependentID)
+			if err != nil {
+				errs = append(errs, fmt.Errorf("Failed to get dependent bug #%d: %v", dependentID, err))
+				break
+			}
+			bugCache.Set(dependentID, dependent)
 		}
 		if dependent.Summary == bug.Summary {
 			clones = append(clones, dependent)
 		}
+
 	}
 	return clones, utilerrors.NewAggregate(errs)
 }
 
 // GetClones gets the list of bugs that the provided bug blocks that also have a matching summary.
 func (c *client) GetClones(bug *Bug) ([]*Bug, error) {
-	return getClones(c, bug)
+	bugCache := NewBugDetailsCache()
+	return getClones(c, bug, bugCache)
 }
 
 // Gets children clones recursively using a mechanism similar to bfs
-func getRecursiveClones(c Client, root *Bug) ([]*Bug, error) {
+func getRecursiveClones(c Client, root *Bug, bugCache *BugDetailsCache) ([]*Bug, error) {
 	var errs []error
 	var bug *Bug
 	clones := []*Bug{}
@@ -160,7 +223,7 @@ func getRecursiveClones(c Client, root *Bug) ([]*Bug, error) {
 	for len(childrenQ) > 0 {
 		bug, childrenQ = childrenQ[0], childrenQ[1:]
 		clones = append(clones, bug)
-		children, err := getClones(c, bug)
+		children, err := getClones(c, bug, bugCache)
 		if err != nil {
 			errs = append(errs, fmt.Errorf("Error finding clones Bug#%d: %v", bug.ID, err))
 		}
@@ -172,18 +235,25 @@ func getRecursiveClones(c Client, root *Bug) ([]*Bug, error) {
 }
 
 // getImmediateParents gets the Immediate parents of bugs with a matching summary
-func getImmediateParents(c Client, bug *Bug) ([]*Bug, error) {
+func getImmediateParents(c Client, bug *Bug, bugCache *BugDetailsCache) ([]*Bug, error) {
 	var errs []error
 	parents := []*Bug{}
 	// One option would be to return as soon as the first parent is found
 	// ideally that should be enough, although there is a check in the getRootForClone function to verify this
 	// Logs would need to be monitored to verify this behavior
 	for _, parentID := range bug.DependsOn {
-		parent, err := c.GetBug(parentID)
-		if err != nil {
-			errs = append(errs, fmt.Errorf("Failed to get parent bug #%d: %v", parentID, err))
-			continue
+
+		parent, ok := bugCache.Get(parentID)
+		if !ok {
+			var err error
+			parent, err = c.GetBug(parentID)
+			if err != nil {
+				errs = append(errs, fmt.Errorf("Failed to get parent bug #%d: %v", parentID, err))
+				continue
+			}
+			bugCache.Set(parentID, parent)
 		}
+
 		if parent.Summary == bug.Summary {
 			parents = append(parents, parent)
 		}
@@ -191,11 +261,11 @@ func getImmediateParents(c Client, bug *Bug) ([]*Bug, error) {
 	return parents, utilerrors.NewAggregate(errs)
 }
 
-func getRootForClone(c Client, bug *Bug) (*Bug, error) {
+func getRootForClone(c Client, bug *Bug, bugCache *BugDetailsCache) (*Bug, error) {
 	curr := bug
 	var errs []error
 	for len(curr.DependsOn) > 0 {
-		parent, err := getImmediateParents(c, curr)
+		parent, err := getImmediateParents(c, curr, bugCache)
 		if err != nil {
 			errs = append(errs, err)
 		}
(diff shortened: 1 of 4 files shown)


GeoBK:
@alvaroaleman - implemented a caching structure which retains the dependents across HTTP requests and another intra-cache which remembers bug details which are accessed in the same call. The way I have done the parallelization is probably not ideal, I used a fixed size slice to hold the results and errors and I'm just looping across it at the end, is there a better way?

prow/bugzilla/client.go
```diff
@@ -214,37 +231,112 @@ func getRootForClone(c Client, bug *Bug) (*Bug, error) {
...
 }
-func getAllClones(c Client, bug *Bug) ([]*Bug, error) {
-	root, err := getRootForClone(c, bug)
+
+type clonesCacheWithMux struct {
+	cache map[int][]int
+	mux   sync.Mutex
+}
+
+var clonesCache = clonesCacheWithMux{cache: map[int][]int{}}
```
alvaroaleman, line 250: This must be threadsafe, you can not use package-level variables for this. Also, lets pelase do this proper and do for all caches we have a:
```
type thingCache struct {
  cache map[string]thing
  lock sync.Mutext
}

func (tc *thingCache)Get(key string) (thing, bool) {
  tc.lock.Lock()
  defer tc.lock.Unlock()
  return tc.cache[key]
}

func (tc *thingCache) set(key string, item thing) {
  tc.lock.Lock()
  defer tc.lock.Unlock()
  tc.cache[key] = item
}
```

prow/bugzilla/client.go
```diff
@@ -214,37 +231,112 @@ func getRootForClone(c Client, bug *Bug) (*Bug, error) {
...
+	cloneErrors := make([]error, len(clonesID))
+	clones := make([]*Bug, len(clonesID))
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		// This error is being handled after the wg.Wait
+		bug, getBugErr = c.GetBug(bugID)
+	}()
+	if ok {
+		var intraCallCacheMux sync.Mutex
```
alvaroaleman, line 270: its a mutex btw, a mux is something else (using used as a term to refer to a http router)

prow/bugzilla/client.go
```diff
@@ -214,37 +231,112 @@ func getRootForClone(c Client, bug *Bug) (*Bug, error) {
...
+				defer wg.Done()
+				var clone *Bug
+				intraCallCacheMux.Lock()
+				clone, ok = intraCallCache[cloneID]
+				intraCallCacheMux.Unlock()
+				if !ok {
+					var err error
+					clone, err = c.GetBug(cloneID)
+					if err != nil {
+						cloneErrors[index] = err
```
alvaroaleman, line 283: This is not threadsafe, you can not concurrently write to a map even if its on different keys

prow/bugzilla/client.go
```diff
@@ -214,37 +231,112 @@ func getRootForClone(c Client, bug *Bug) (*Bug, error) {
...
+						return
+					}
+				}
+				clones[index] = clone
```
alvaroaleman, line 287: also not threadsafe. And it makes more sense to directly write into the cache rather than in a slice and then coipying that later in the func, because:
* Its easier to read
* You don't have to allocate the slice
* It might be re-used by one of the other goroutines and save that goroutine a call to Bugzilla

prow/bugzilla/client.go
```diff
@@ -214,37 +231,112 @@ func getRootForClone(c Client, bug *Bug) (*Bug, error) {
...
+			}(index, cloneID)
+		}
+	}
+	wg.Wait()
+	intraCallCache[bug.ID] = bug
```
alvaroaleman, line 292: This is super confusing to read, why do we assign here rather than where we retrieved the bug?

prow/bugzilla/client.go
```diff
@@ -214,37 +231,112 @@ func getRootForClone(c Client, bug *Bug) (*Bug, error) {
...
+	for _, clone := range clones {
+		intraCallCache[clone.ID] = clone
+	}
+	if getBugErr != nil {
```
alvaroaleman, line 296: Instead of doing all of this, we can just use https://godoc.org/golang.org/x/sync/errgroup because once we had one failure we essentially failed and don't need to wait for other routines

prow/bugzilla/client.go
```diff
@@ -172,30 +182,37 @@ func getRecursiveClones(c Client, root *Bug) ([]*Bug, error) {
...
 	// One option would be to return as soon as the first parent is found
 	// ideally that should be enough, although there is a check in the getRootForClone function to verify this
 	// Logs would need to be monitored to verify this behavior
 	for _, parentID := range bug.DependsOn {
-		parent, err := c.GetBug(parentID)
-		if err != nil {
-			errs = append(errs, fmt.Errorf("Failed to get parent bug #%d: %v", parentID, err))
-			continue
+
+		parent, ok := cache[parentID]
```
alvaroaleman, line 193:
```
parent, hasParent := cache[parentID]
if hasParent {
   continue
}
/// GetParent
```

prow/bugzilla/client.go
```diff
@@ -127,29 +129,37 @@ func (c *client) GetBug(id int) (*Bug, error) {
...
+func getClones(c Client, bug *Bug, cache map[int]*Bug) ([]*Bug, error) {
 	var errs []error
 	clones := []*Bug{}
 	for _, dependentID := range bug.Blocks {
-		dependent, err := c.GetBug(dependentID)
-		if err != nil {
-			errs = append(errs, fmt.Errorf("Failed to get dependent bug #%d: %v", dependentID, err))
-			continue
+
+		dependent, ok := cache[dependentID]
```
alvaroaleman, line 137: `continue` on cache hit rather than branching on cache miss

prow/bugzilla/client.go
GeoBK: This is not a map, its a slice and indexing each go routine should ensure typesafety I believe.

prow/bugzilla/client.go
GeoBK: Why would it not be threadsafe, I am writing stuff in different locations for a slice whose memory has already been allocated right? Using a slice like me would allow me to avoid using a lock in this scenario which would be needed if I were using a set/map.
1. Agreed, readability would improve if I were to assign it directly to the cache
3. I don't think so as all the clones in the list would be distinct.

prow/bugzilla/client.go
GeoBK: Assigning it here to keep it threadsafe, I would have to guard it if I were to assign it earlier.

prow/bugzilla/client.go
alvaroaleman: Okay the pre-allocation makes it threadsafe but there is no verification for this (someone might change this in the future) and its harder to read and a data structure we don't really need, we will copy it over anyways

prow/bugzilla/client.go
alvaroaleman: Yeah but whats wrong with guarding it? Just abstract that away via a Setter as explained above. The lock is only needed for accessing the cache which is super quick so I doubt this will cause any kind of lock contention

prow/bugzilla/client.go
GeoBK: I don't think I can do that unless I duplicate the code which checks if it is a clone - which needs to be done both in case its present in the cache or not. (Unless I'm missing something here)

prow/bugzilla/client.go
GeoBK: Same case as above, If i were to continue on cache hit I wouldn't execute the next if loop

prow/bugzilla/client.go
GeoBK: Yup, that seems much cleaner

prow/bugzilla/client.go
```diff
@@ -214,37 +284,95 @@ func getRootForClone(c Client, bug *Bug) (*Bug, error) {
...
+	// Remember to use a mutex to ensure thread safety
+	// Do not move this section after the section where the original bug is removed
+	cloneSet := sets.NewInt()
+	for bugID := range bugCache.cache {
+		cloneSet.Insert(bugID)
+	}
+	for _, clone := range recursedClones {
+		cloneSet.Delete(clone.ID)
+		c.SetInCache(clone.ID, cloneSet.List())
+		cloneSet.Insert(clone.ID)
```
petr-muller, line 360: Er. What is happening here? We delete an ID from the set, then set something in the cache, then insert it again, so that we can multiple times overwrite the cache with some subset of the `cloneSet`... That does not make sense to me, can you explain why we are doing this?

prow/bugzilla/client.go
```diff
@@ -214,37 +284,95 @@ func getRootForClone(c Client, bug *Bug) (*Bug, error) {
...
 	}
 	// Getting rid of the bug whose clones we are searching for
 	// we could optimize this (maybe?) if the list turns out to be too long
 	indexOfOriginal := -1
-	for index, clone := range clones {
+	for index, clone := range recursedClones {
 		if clone.ID == bug.ID {
 			// Check if there is a more elegant way to do this
```
petr-muller, line 367: Maybe just build the target slice in this loop directly by appending all members that are not the one that we want to remove? If you preallocate the slice with `make` to have a necessary size it should not be a problem performance-wise.

prow/bugzilla/client.go
```diff
@@ -214,37 +284,95 @@ func getRootForClone(c Client, bug *Bug) (*Bug, error) {
...
+	g := new(errgroup.Group)
+	g.Go(func() error {
+		var err error
+		bug, err = c.GetBug(bugID)
+		if err != nil {
+			return err
+		}
+		bugCache.Set(bugID, bug)
+		return nil
+	})
```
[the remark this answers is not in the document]
GeoBK: my bad, I added that in later and didn't notice this. Moving that out.

prow/bugzilla/client.go
GeoBK: Yup, the intention here is to set the cache to remember all the associations so that we can parallelize all the calls the next time. ie if bug1, bug2 and bug3 are clones of bug4 then bug2, bug3 and bug4 are also clones of bug1 and so on.
That is why I'm setting it in this manner.

prow/bugzilla/client.go
```diff
@@ -127,29 +182,37 @@ func (c *client) GetBug(id int) (*Bug, error) {
...
+
+		dependent, ok := bugCache.Get(dependentID)
+		if !ok {
+			var err error
+			dependent, err = c.GetBug(dependentID)
+			if err != nil {
+				errs = append(errs, fmt.Errorf("Failed to get dependent bug #%d: %v", dependentID, err))
+				break
+			}
+			bugCache.Set(dependentID, dependent)
```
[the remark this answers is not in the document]
alvaroaleman: Yeah I also thought that maybe just caching all results and having a background loop that updates everything in the cache every interval could be a suitable alternative thats less complicated and thus error prone

prow/bugzilla/client.go
```diff
@@ -75,25 +78,77 @@ type Client interface {
 	RemovePullRequestAsExternalBug(id int, org, repo string, num int) (bool, error)
 	// GetAllClones returns all the clones of the bug including itself
 	// Differs from GetClones as GetClones only gets the child clones which are one level lower
-	GetAllClones(bug *Bug) ([]*Bug, error)
+	// GetAllClones will also fetch bugDetails and the root bug details in addition to clones
+	// since we want to parallelize all the HTTP calls
+	GetAllClones(bugID int) (clones []*Bug, bug *Bug, root *Bug, err error)
 	// GetRootForClone returns the original bug.
 	GetRootForClone(bug *Bug) (*Bug, error)
+	// GetFromCache will fetch the clones from the clones cache
```
stevekuznetsov, line 86: Why do we need these two? I don't think the public interface for the client should allow for this sort of thing.

prow/bugzilla/client.go
```diff
@@ -75,25 +78,77 @@ type Client interface {
...
 	return &client{
-		logger:    logrus.WithField("client", "bugzilla"),
-		client:    &http.Client{},
-		endpoint:  endpoint,
-		getAPIKey: getAPIKey,
+		logger:      logrus.WithField("client", "bugzilla"),
+		client:      &http.Client{},
+		endpoint:    endpoint,
+		getAPIKey:   getAPIKey,
+		clonesCache: clonesCacheWithMutex{cache: map[int][]int{}},
```
stevekuznetsov, line 98: The cache is not valid to use between calls, I thought this was supposed to be only for the double-traversal when we were running a `GetAllClones()` from the middle of a tree?

prow/bugzilla/client.go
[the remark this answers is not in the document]
stevekuznetsov: This is the opposite of what I suggested -- we don't actually have any way to ensure that this cache is up-to-date, so it's of limited utility. This was not meant to be a generic cache for bugs, it was meant to hold onto data that was fetched when resolving a bug clone tree during _one_ client call.

prow/bugzilla/client.go
```diff
@@ -75,25 +78,77 @@ type Client interface {
...
 type client struct {
-	logger    *logrus.Entry
-	client    *http.Client
-	endpoint  string
-	getAPIKey func() []byte
+	logger      *logrus.Entry
+	client      *http.Client
+	endpoint    string
+	getAPIKey   func() []byte
+	clonesCache clonesCacheWithMutex
```
stevekuznetsov, line 151: Wait, is this currentl used?

prow/bugzilla/client.go
GeoBK: I added the clones list cache to the client instead of keeping it as a package variable. The only way I could access it from the client interface was to add these functions to set and get from cache. Let me know if some other method is appropriate for this.

prow/bugzilla/client.go
GeoBK: I have 2 caches now. This one should be valid between calls and holds a list of clones associated with a bug to allow for parallelizing all the GetBug calls.
The other one is for the double traversal(ie. the bugCache) and is not used between calls

prow/bugzilla/client.go
GeoBK: Yes, this is the cache used by the Get/Set function you had commented on earlier.

prow/bugzilla/client.go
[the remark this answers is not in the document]
GeoBK: No, I had not. I am not sure how I would proceed with that either. Could you please elaborate?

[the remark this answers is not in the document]
GeoBK: @stevekuznetsov  right now thats what the bugCache does, ie it holds the bug details for one client call only

prow/bugzilla/client.go
```diff
@@ -205,46 +294,91 @@ func getRootForClone(c Client, bug *Bug) (*Bug, error) {
...
+				bugCache.Set(clone.ID, clone)
+				return nil
+			})
+		}
+	}
+	if err := g.Wait(); err != nil {
+		return nil, nil, nil, fmt.Errorf("failed to get bug details: %v", err)
+	}
+
+	root, err := getRootForClone(c, bug, bugCache)
```
alvaroaleman, line 360: Can we parallelize all of this by doing something like:
```
func GetAllClones(id int) error {
 return getAllClones(id, nil, nil
}

func getAllClones(id int, result cache, errGroup *errgroup.Group) error {
       if cache == nil {
               cache = newCache()
       }
       var shouldWait bool
       if errGroup == nil {
               errGroup = newErrGroup()
               shouldWait = true
       }

       var err error
       bug, cacheHasBug := cache.Has(id)
       if !cacheHasBug() {
               bug, err = getBug(id)
               if err != nil {
                       return err
               }
       }
       errGroup.Go(traverseUp(cache.Get(id), cache, errGroup))
       errGroup.Go(traverseDown(cache.Get(id), cache, errGroup))

       if shouldWait {
               return errGroup.Wait()
       }
       return nil
}

func traverseUp(bug Bug, result cache, errGroup *errgroup.Group) error {
       for _, dependsOnID := range bug.DependsOn {
               dependsOnID := dependsOnID
               errgroup.Go(func() error {
                       if result.Has(dependsOnId) {
                               return nil
                       }
                       parent, err := getBug(dependsOn.Id)
                       if err != nil {
                               return err
                       }
                       result.Set(bug)
                       return getAllClones(parent.ID, cache, errGroup)
               })
       }
       return nil
}

func traverseDown(bug Bug, result cache, errGroup *errGroup.group) error {
       for _, childID := range bug.Blocks {
               childID := childID
               errGroup.Go(func() error {
                       if result.Has(child) {
                               return nil
                       }
                       childBug, err := getBug(childID)
                       if err != nil {
                               return err
                       }
                       result.Set(childBug)
                       return getAllClones(childBug.ID)
               })
       }
       return nil
}
```

prow/bugzilla/client.go
```diff
@@ -75,25 +79,77 @@ type Client interface {
...
-		getAPIKey: getAPIKey,
+		logger:      logrus.WithField("client", "bugzilla"),
+		client:      &http.Client{},
+		endpoint:    endpoint,
+		getAPIKey:   getAPIKey,
+		clonesCache: clonesCacheWithMutex{cache: map[int][]int{}},
 	}
 }
 
+type clonesCacheWithMutex struct {
```
alvaroaleman, line 103: The `WithMutex` in the name is a  superfluous implementation detail, the Bugzilla client is also not named `BugZillaClientWithHTTPClient`

prow/bugzilla/client.go
alvaroaleman: And probably add an incrementing counter to `traverseUp` and store the value of that together with the bug in the cache so we can find out who the root is?

prow/bugzilla/client.go
alvaroaleman: I think the `clonesCache` is something for a dedicated PR and might make more sense downstream?

prow/bugzilla/client.go
alvaroaleman: Or maybe just use `getOldestBug` to find the root from the final result?

GeoBK: Moving into separate PR as discussed offline

PR #18222, opened 2020-07-09
Kops - Build both AMD64 and ARM64 for Kubernetes presubmit

hakman:
Since ARM64 support, Kops needs Kubernetes ARM64 binaries.
Fixes: https://github.com/kubernetes/kops/issues/9526

/cc @rifelpet

### DIFF
--- a/config/jobs/kubernetes/kops/kubernetes-presubmits.yaml
+++ b/config/jobs/kubernetes/kops/kubernetes-presubmits.yaml
@@ -28,12 +28,12 @@ presubmits:
         - "--repo=k8s.io/release"
         - "--service-account=/etc/service-account/service-account.json"
         - "--upload=gs://kubernetes-jenkins/pr-logs"
-        - "--timeout=75"
+        - "--timeout=200"
         - --scenario=kubernetes_e2e
         - --
         - --aws
         - --aws-cluster-domain=test-cncf-aws.k8s.io
-        - --build=bazel
+        - --build=release
         - --cluster=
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes/kops/kubernetes-presubmits.yaml
```diff
@@ -28,12 +28,12 @@ presubmits:
...
         - "--service-account=/etc/service-account/service-account.json"
         - "--upload=gs://kubernetes-jenkins/pr-logs"
-        - "--timeout=75"
+        - "--timeout=200"
         - --scenario=kubernetes_e2e
         - --
         - --aws
         - --aws-cluster-domain=test-cncf-aws.k8s.io
-        - --build=bazel
+        - --build=release
```
BenTheElder, line 36: please do not cut a full release build in presubmit testing (!)
these are _wildly_ expensive

BenTheElder:
kops should be fixed to not depend on arm64 binaries when testing amd64.

we do not run the cross build in presubmit unless the build code is touched, it takes upwards of an hour on 8 cores with something like 50gb of ram. that's an entire CI build machine.

I am reverting this change.

config/jobs/kubernetes/kops/kubernetes-presubmits.yaml
BenTheElder: further, we use bazel in presubmit for caching for a reason.

PR #18223, opened 2020-07-09
CI tests: set up a few CI tests for kubernetes-csi/csi-driver-smb

Sakuralbj:
[Addressing](https://github.com/kubernetes-csi/csi-driver-smb/issues/54)
@andyzhangx

### DIFF
--- a/config/jobs/kubernetes-csi/csi-driver-smb/csi-driver-smb-config.yaml
+++ b/config/jobs/kubernetes-csi/csi-driver-smb/csi-driver-smb-config.yaml
@@ -1,0 +1,117 @@
+presubmits:
+  kubernetes-csi/csi-driver-smb:
+  - name: pull-csi-driver-smb-verify
+    decorate: true
+    always_run: true
+    path_alias: sigs.k8s.io/csi-driver-smb
+    branches:
+    - master
+    labels:
+      preset-service-account: "true"
+    spec:
+      containers:
+      - image: gcr.io/k8s-testimages/kubekins-e2e:v20200623-2424179-master
+        command:
+        - runner.sh
+        args:
+        - make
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes-csi/csi-driver-smb/csi-driver-smb-config.yaml
```diff
@@ -0,0 +1,117 @@
+presubmits:
+  kubernetes-csi/csi-driver-smb:
+  - name: pull-csi-driver-smb-verify
```
msau42, line 3: We actually have a script that generates pull and CI jobs for all the kubernetes-csi repos.

It involves ".prow.sh" with "UNIT", which already does build, verify, and unit tests. Do you want to use that instead?

https://github.com/kubernetes/test-infra/blob/c212d67a058661eb01c9bce29c1672826a2cddd8/config/jobs/kubernetes-csi/gen-jobs.sh#L390

Unfortunately the CI part is a bit hardcoded to the hostpath driver, so we would need to do some additional work to make it work for other drivers too cc @pohly

andyzhangx: @msau42 shall we merge this PR first? we may modify this yaml in next iteration?

msau42: Sure that's fine

andyzhangx: thanks, filed a new issue to track this: https://github.com/kubernetes-csi/csi-driver-smb/issues/56

PR #18225, opened 2020-07-09
Revert "Kops - Build both AMD64 and ARM64 for Kubernetes presubmit"

BenTheElder:
Reverts kubernetes/test-infra#18222

this is way too expensive. kops needs to use the cached amd64 build.


BenTheElder:
this is going to break the job I think but:
- the fix should be made in kops to allow arm64 to not be present when we're not testing it
- the job is not required or even run by default, changing this prevents users from hitting a "hijack a CI node for two hours" button
/hold cancel

PR #18240, opened 2020-07-09
Make autobump only update gerrit PRs when there are changes.

cjwagner:
Unlike GitHub, pushing the same changes to Gerrit multiple times causes new patchsets to be created rather than behaving as a no-op. This means that the autobump job generates a new patchset on the PR every time it runs, even if there is no change since the previous run.
To avoid this I've added some logic to look up the remote ref with the matching `Change-Id` and compare against that to determine if we actually need to push changes and update the PR. Rather than using the Gerrit API to map the `Change-Id` to the correct refspec so that we can fetch the individual PR, I've opted for fetching all the PR refs. This takes about 10 seconds on our GoB repo, but I think it is worth avoiding the complexity of providing Gerrit credentials and interacting with the Gerrit API from bash. @chizhg is working on migrating autobump to Go as which point we can use the gerrit go client if we want to make this more performant.
I've tested the new behavior with pj-on-kind
/assign @fejta

### DIFF
--- a/prow/cmd/autobump/autobump.sh
+++ b/prow/cmd/autobump/autobump.sh
@@ -35,8 +35,8 @@ PROW_INSTANCE_NAME="${PROW_INSTANCE_NAME:-prow}"
 # TODO(fejta): rewrite this in a better language REAL SOON  <-lol
 main() {
 	if [[ $# -lt 1 ]]; then
-	    echo "Usage: $(basename "$0") <path to github token or http cookiefile> [git-name] [git-email]" >&2
-	    exit 1
+			echo "Usage: $(basename "$0") <path to github token or http cookiefile> [git-name] [git-email]" >&2
+			return 1
 	fi
 	creds=$1
 	shift
@@ -52,7 +52,7 @@ main() {
 
 	if [[ "${old_version}" == "${version}" ]]; then
 		echo "Bump did not change the Prow version: it's still ${version}. Aborting no-op bump." >&2
-		exit 0
+		return 0
 	fi
 	git add -u
 	title="Bump ${PROW_INSTANCE_NAME} from ${old_version} to ${version}"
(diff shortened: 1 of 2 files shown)


prow/cmd/autobump/autobump.sh
```diff
@@ -128,9 +128,24 @@ create-gerrit-pr() {
...
-	change_id="$(get-change-id)"
+	local change_id="$(get-change-id)"
 	echo "Commit will have Change-Id: ${change_id}"
 
+	# Check for an existing open PR and see if it needs to be updated.
+	echo "Checking for an existing open Gerrit PR to update..."
+	git fetch upstream "+refs/changes/*:refs/remotes/upstream/changes/*" 2> /dev/null
+	local pr_commit="$(git log --all --grep="Change-Id: ${change_id}" -1 --format="%H")"
+	if [[ -n "${pr_commit}" ]]; then
+		local pr_version="$(git show "${pr_commit}:${PLANK_DEPLOYMENT_FILE}" | extract-version)"
```
fejta, line 139: Wouldn't `git diff $pr_commit` be better than just checking the plank version? AKA if any of the files change we want to push a new commit.

prow/cmd/autobump/autobump.sh
```diff
@@ -128,9 +128,24 @@ create-gerrit-pr() {
...
+		if [[ "${pr_version}" == "${version}" ]]; then
+			echo "Bump PR is already up to date (version ${version}). Aborting no-op update." >&2
+			exit 0
```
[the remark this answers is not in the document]
cjwagner: Fixed here and elsewhere.

prow/cmd/autobump/autobump.sh
cjwagner: Yes thats definitely better. Fixed.

PR #18244, opened 2020-07-10
[wip] rebased secret-sync-controller PR

spiffxp:
This rebases https://github.com/kubernetes/test-infra/pull/17946 on top of https://github.com/kubernetes/test-infra/pull/18000 to see if CI will repro what I'm seeing locally. Not intended for merge.

Added a commit to clean up go.mod and resulting deps.


spiffxp:
The problem being demonstrated now is that bazel (or gazelle?) seems to be seeing an older version of `google.golang.org/genproto` than I would expect

The 'version' field here
https://github.com/kubernetes/test-infra/blob/ae613f5e86ab894df63031bab1c6c616b4588f0a/repos.bzl#L1509-L1516

Matches the version here https://github.com/kubernetes/test-infra/blob/ae613f5e86ab894df63031bab1c6c616b4588f0a/go.mod#L87

So I would expect `bazel` and `go` to both see this:
https://github.com/googleapis/go-genproto/tree/7a2ca40e9dc324f00c1e7ea8786ddf6366d9e803/googleapis/cloud/secretmanager/v1

spiffxp:
It seems like we're hitting something similar to this issue: https://github.com/bazelbuild/rules_go/issues/2420

The version of `org_golang_google_genproto` we're specifying should contain `secretmanager/v1`, but instead we see
```
$ bazel query "@org_golang_google_genproto//googleapis/cloud/secretmanager/..."
@org_golang_google_genproto//googleapis/cloud/secretmanager/v1beta1:go_default_library
Loading: 0 packages loaded
```

I suspect the version we're seeing is coming in from rules_go v0.22.6 (coming in from repo-infra v0.0.6):
- https://github.com/bazelbuild/rules_go/blob/v0.22.6/go/private/repositories.bzl#L182-L201
- has a comment saying "master, as of 2020-02-21"
- which is too early to have caught secretmanager/v1 (first added 2020-02-24 via https://github.com/googleapis/go-genproto/commit/e50cd9704f63023d62cd06a1994b98227fc4d21a#diff-28e1ce177623aa3dd01dcc6df440d545)

I tried bumping the version of rules_go to v0.23.3 in my local copy of repo-infra:
- this would bring in https://github.com/bazelbuild/rules_go/blob/v0.23.3/go/private/repositories.bzl#L220-L239
- “master, as of 2020-05-21”
- however a `./hack/update-deps.sh` in repo-infra errored out with patches not applying
- `bazel build //experiments/secret-sync-controller` in test-infra led to `<snip>/execroot/io_k8s_test_infra/external/com_google_cloud_go/secretmanager/apiv1/secret_manager_client.go:495:3: undefined: "google.golang.org/genproto/googleapis/iam/v1".m`
- which corresponds to no line of code that I can see
- I suspect there are deeper reasons repo-infra hasn't migrated to rules_go v0.23.x

spiffxp:
I rolled back to rules_go v0.22.6, and tried following https://github.com/bazelbuild/rules_go/blob/master/go/dependencies.rst#overriding-dependencies to override the `org_golang_google_genproto` repo.  Specifically the "declaring a repository rule in WORKSPACE with the same name _before_ the call to go_rules_dependencies" part.
```bazel
# test-infra/WORKSPACE

load("@io_k8s_repo_infra//:repos.bzl", "configure")

# begin hack
load("@bazel_gazelle//:deps.bzl", "go_repository")
go_repository(
    name = "org_golang_google_genproto",
    build_file_generation = "on",
    build_file_proto_mode = "disable",
    importpath = "google.golang.org/genproto",
    sum = "h1:JwLN1jVnmIsfE4HkDVe2AblFAbo0Z+4cjteDSOnv6oE=",
    version = "v0.0.0-20200709005830-7a2ca40e9dc3",
)
# end hack

configure(
    go_version = "1.13",
    nogo = "@//:nogo_vet",
)
# which ends up calling go_rules_dependencies(), among other things
```

This time `bazel build //experiments/secret-sync-controller` yields: `<snip>/execroot/io_k8s_test_infra/external/org_golang_google_genproto/googleapis/cloud/secretmanager/v1/resources.pb.go:43:11: undefined: "github.com/golang/protobuf/proto".ProtoPackageIsVersion4`

Which is probably because repo-infra specifying this in `repo_infra_patches()`

https://github.com/kubernetes/repo-infra/blob/63918b0966c3f320f7af2a17d85d13f0170ba05c/repos.bzl#L43-L50

spiffxp:
/close
abandoning in favor of kubernetes-sigs/k8s-gsm-tools

PR #18252, opened 2020-07-10
Use structured logging in gerrit packages.

fejta:
This should make it easier to search for and debug logs.

Also reduce some indentation by continuing/returning early.

### DIFF
--- a/prow/gerrit/client/client.go
+++ b/prow/gerrit/client/client.go
@@ -26,7 +26,7 @@ import (
 	"sync"
 	"time"
 
-	"github.com/andygrunwald/go-gerrit"
+	gerrit "github.com/andygrunwald/go-gerrit"
 	"github.com/sirupsen/logrus"
 )
 
@@ -103,6 +103,8 @@ type gerritInstanceHandler struct {
 	accountService gerritAccount
 	changeService  gerritChange
 	projectService gerritProjects
+
+	log logrus.FieldLogger
 }
 
 // Client holds a instance:handler map
@@ -157,6 +159,7 @@ func NewClient(instances map[string][]string) (*Client, error) {
 			accountService: gc.Accounts,
 			changeService:  gc.Changes,
 			projectService: gc.Projects,
+			log:            logrus.WithField("host", instance),
 		}
 	}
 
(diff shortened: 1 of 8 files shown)


prow/gerrit/client/client.go
```diff
@@ -246,11 +257,12 @@ func (c *Client) QueryChanges(lastState LastSyncState, rateLimit int) map[string
...
-			result[h.instance] = []ChangeInfo{}
-			for _, change := range changes {
-				result[h.instance] = append(result[h.instance], change)
-			}
+		if len(changes) == 0 {
+			continue
+		}
+
+		for _, change := range changes {
+			result[h.instance] = append(result[h.instance], change)
```
chizhg, line 265: Is it safe to assume `h.instance` always exists in the `result` map? In the deleted code there is `result[h.instance] = []ChangeInfo{}`

prow/gerrit/client/client.go
```diff
@@ -366,72 +382,86 @@ func (h *gerritInstanceHandler) queryChangesForProject(project string, lastUpdat
...
-					// change has been abandoned, do nothing
+				if !newMessages && !created.After(lastUpdate) {
+					// stale commit
+					log.Info("Skipping existing change")
+					continue
 				}
-			} else {
-				logrus.Infof("Change %d, updated %s before lastUpdate %s, return", change.Number, change.Updated, lastUpdate)
-				return pending, nil
+				if !newMessages {
```
chizhg, line 458: The log message is opposite from what's meant by this if check, should it be `if newMessages`?

prow/gerrit/client/client.go
fejta: That is not necessary.

```golang
var foo []string = nil
foo = append(foo, "hello")
```

works just fine. This is in fact one of the reasons why append returns a value.

See https://play.golang.org/p/EVcqaQ0JNOI

prow/gerrit/client/client.go
fejta: If there are new messages then we got to L447 which means we already logged on line L443.

L443 covers the case where the change was updated by a comment.

L459 covers the case where someone pushed new code.

chizhg: Got it. Thanks for the explanation!

prow/gerrit/client/client.go
chizhg: I see. Thanks for the explanation and example!

PR #18254, opened 2020-07-10
Kops - Skip new tests on older branches

hakman:
Similar to #18172 .

/cc @johngmyers @rifelpet


hakman:
Example of PR where these jobs fail for `release-1.17`:
https://github.com/kubernetes/kops/pull/9549

rifelpet:
good catch 👍 
/lgtm
/approve

hakman:
> good catch 👍
more like bumping into it at full speed 😆

PR #18256, opened 2020-07-10
Kops - Fix kube-router tests

hakman:
Includes:
* more exclusions for kube-router test to become green
* increased interval for kube-router tests
* new presubmit for kuberouter
* cluster domain includes plugin name to not clash

/cc @rifelpet

### DIFF
--- a/config/jobs/kubernetes/kops/kops-presubmits-network-plugins.yaml
+++ b/config/jobs/kubernetes/kops/kops-presubmits-network-plugins.yaml
@@ -38,7 +38,7 @@ presubmits:
         - --extract=release/latest
         - --ginkgo-parallel
         - --kops-build
-        - --kops-args=--networking=amazon-vpc-routed-eni --node-size=t3.large
+        - --kops-args=--networking=amazonvpc --node-size=t3.large
         - --kops-priority-path=/home/prow/go/src/k8s.io/kops/kubernetes/platforms/linux/amd64
         - --provider=aws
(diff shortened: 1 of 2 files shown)


config/jobs/kubernetes/kops/kops-presubmits-network-plugins.yaml
```diff
@@ -79,7 +79,7 @@ presubmits:
         - /workspace/scenarios/kubernetes_e2e.py
         args:
         - --aws
-        - --aws-cluster-domain=test-cncf-aws.k8s.io
+        - --aws-cluster-domain=test-cncf-calico.k8s.io
```
rifelpet, line 82: are you sure this is safe? i don't know exactly how the AWS accounts are setup but I would imagine there is a hosted zone for test-cncf-aws.k8s.io and changing this might mean that kops can't find a hosted zone to add records to.

what exactly is the problem caused by having them all use the same aws-cluster-domain?

hakman: I was thinking they may clash if they run at same time, like the ARM64 ones last week.

hakman: You may be right, so I changed back.

rifelpet:
/lgtm
/approve

Random thought: we should consider occasionally reevaluating all the tests that we skip in case they end up getting fixed. Perhaps once per minor release.

PR #18260, opened 2020-07-10
Kops - Fix branch names for release pipelines

hakman:
Just fixing a typo :)

/cc @rifelpet

### DIFF
--- a/config/jobs/kubernetes/kops/build-pipeline.py
+++ b/config/jobs/kubernetes/kops/build-pipeline.py
@@ -88,13 +88,17 @@ def build_tests(branch, k8s_version, ssh_user):
     name = name.replace('.', '')
 
     y = template
-    y = y.replace('{{branch}}', branch)
     y = y.replace('{{extract}}', extract)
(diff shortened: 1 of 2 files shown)


rifelpet:
I ran the python scripts with these changes in it and got a few more changes in the yaml file than what is in this PR so we should make sure those are included as well.

Also is this to fix a failing job that is looking for a release marker in a different place than where this is publishing it? if so can you link to that job? just so I understand the full picture here.

hakman:
The release `ci-latest.txt` comes from this PR: https://github.com/kubernetes/kops/pull/9272 and `_PULL_BASE_REF` seems to be the branch name.

The pipeline job will use the `ci-latest.txt` file and, if it passes, will update `latest-ci-updown-green.txt`.

I think the script incorrectly assumes the branch mane as `$VERSION` instead of `release-SVERSION`.

rifelpet:
ok that makes sense to me. thanks for the explanation 👍 
/lgtm
/approve

PR #18263, opened 2020-07-10
Delete old Kyma TestGrid configuration file

Ressetkk:
We are introducing automation using [transfigure.sh](https://github.com/kubernetes/test-infra/tree/master/testgrid/cmd/transfigure) for the Kyma test infra and we need to remove the old `kyma.yaml` file.

This removal is mandatory, as right now the testgrid config validation tests are failing because of duplicated names in configs.


Ressetkk:
It’s pretty weird as you are an approver in the owners file

PR #18265, opened 2020-07-10
Use kubekins-e2e 1.18 images for Cluster API

vincepri:


vincepri:
Used `v20200710-6b3b6fe-1.18` instead of latest

PR #18270, opened 2020-07-10
Fix prow config for gce pd driver windows test

jingxu97:
Fix config issues


jingxu97:
cc @krzyzacy @BenTheElder

PR #18278, opened 2020-07-10
Sidecar Support for Multiple Containers in Parallel

anthonyrtong:
Update sidecar to support multiple containers running in parallel
- `WaitForMarker()` is now `WaitForMarkers()` and watches for multiple marker file paths.
- Sidecar's `wait()` function calls `WaitForMarkers()` to retrieve exit codes for each test container.
- Entrypoint's `Mark()` function creates a temp file outside the directory of the desired marker file.  This is necessary to prevent concurrency bugs with fsnotify's `addWatch()`.
- New test cases added to test new sidecar functionality.

### DIFF
--- a/prow/entrypoint/run.go
+++ b/prow/entrypoint/run.go
@@ -79,7 +79,7 @@ func (o Options) Run() int {
 	if err != nil {
 		logrus.WithError(err).Error("Error executing test process")
 	}
-	if err := o.mark(code); err != nil {
+	if err := o.Mark(code); err != nil {
 		logrus.WithError(err).Error("Error writing exit code to marker file")
 		return InternalErrorCode // we need to mark the real error code to safely return AlwaysZero
 	}
@@ -122,7 +122,8 @@ func (o Options) ExecuteProcess() (int, error) {
 			case <-ctx.Done():
 			}
 		}()
-		code, err := wrapper.WaitForMarker(ctx, o.PreviousMarker)
+		prevMarkerResult := wrapper.WaitForMarkers(ctx, o.PreviousMarker)[o.PreviousMarker]
+		code, err := prevMarkerResult.ReturnCode, prevMarkerResult.Err
 		cancel() // end previous go-routine when not interrupted
 		if err != nil {
 			return InternalErrorCode, fmt.Errorf("wait for previous marker %s: %v", o.PreviousMarker, err)
@@ -196,12 +197,12 @@ func (o Options) ExecuteProcess() (int, error) {
 	return returnCode, commandErr
 }
 
-func (o *Options) mark(exitCode int) error {
+func (o *Options) Mark(exitCode int) error {
 	content := []byte(strconv.Itoa(exitCode))
 
-	// create temp file in the same directory as the desired marker file
+	// create temp file outside the directory of the desired marker file
 	dir := filepath.Dir(o.MarkerFile)
-	tempFile, err := ioutil.TempFile(dir, "temp-marker")
+	tempFile, err := ioutil.TempFile("", "temp-marker")
 	if err != nil {
 		return fmt.Errorf("could not create temp marker file in %s: %v", dir, err)
 	}
--- a/prow/entrypoint/run.go
+++ b/prow/entrypoint/run.go
@@ -25,7 +25,6 @@ import (
 	"os"
 	"os/exec"
 	"os/signal"
-	"path/filepath"
 	"strconv"
 	"syscall"
 	"time"
@@ -201,10 +200,9 @@ func (o *Options) Mark(exitCode int) error {
 	content := []byte(strconv.Itoa(exitCode))
 
 	// create temp file outside the directory of the desired marker file
-	dir := filepath.Dir(o.MarkerFile)
 	tempFile, err := ioutil.TempFile("", "temp-marker")
 	if err != nil {
-		return fmt.Errorf("could not create temp marker file in %s: %v", dir, err)
+		return fmt.Errorf("could not create temp marker file: %v", err)
 	}
 	// write the exit code to the tempfile, sync to disk and close
 	if _, err = tempFile.Write(content); err != nil {
--- a/prow/pod-utils/wrapper/options.go
+++ b/prow/pod-utils/wrapper/options.go
@@ -100,11 +100,13 @@ func WaitForMarkers(ctx context.Context, paths ...string) map[string]MarkerResul
 	}
 	watcher, err := fsnotify.NewWatcher()
 	if err != nil {
+		populateMapWithError(results, fmt.Errorf("new fsnotify watch: %v", err), paths...)
 		return results
 	}
 	defer watcher.Close()
 	dir := filepath.Dir(paths[0])
 	if err := watcher.Add(dir); err != nil {
+		populateMapWithError(results, fmt.Errorf("add %s to fsnotify watch: %v", dir, err), paths...)
 		return results
 	}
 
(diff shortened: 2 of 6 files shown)


prow/pod-utils/wrapper/options.go
```diff
@@ -78,49 +83,71 @@ func (o *Options) Validate() error {
...
+	}
+
+	for _, path := range paths {
+		if _, err := os.Stat(path); !os.IsNotExist(err) {
+			results[path] = readMarkerFile(path)
 		}
+	}
+	watcher, err := fsnotify.NewWatcher()
+	if err != nil {
+		return results
```
petr-muller, line 103: I think this overall needs better error handling -- callsites expect the returned map to have a meaningful `MarkerResult` for each path they passed in, and returning it possibly empty means they can receive a `MarkerResults` zero value, which is `ReturnCode==0 Err=nil`, which is the exact opposite of the truth.

So IMO we should either populate the map with default error values (meaning something like "did not even attempt to check"), or this method should have a `error` of its own, meaning "all paths have meaningful `MarkerResults`" when not nil.

anthonyrtong: Agreed, thank you for pointing this out.  I think I am inclined to implement the "populate the map with default error values" approach; that seems like the more intuitive behavior IMO.  I'll go ahead and make this change.

prow/entrypoint/run.go
```diff
@@ -196,12 +197,12 @@ func (o Options) ExecuteProcess() (int, error) {
...
+func (o *Options) Mark(exitCode int) error {
 	content := []byte(strconv.Itoa(exitCode))
 
-	// create temp file in the same directory as the desired marker file
+	// create temp file outside the directory of the desired marker file
 	dir := filepath.Dir(o.MarkerFile)
-	tempFile, err := ioutil.TempFile(dir, "temp-marker")
+	tempFile, err := ioutil.TempFile("", "temp-marker")
 	if err != nil {
 		return fmt.Errorf("could not create temp marker file in %s: %v", dir, err)
```
e-blackwelder, line 207: I think the use of `dir` here is misleading/wrong. And it looks like the `dir` local var is now unused. (I would expect the Go compiler to warn/error about the latter.)

prow/entrypoint/run.go
```diff
@@ -122,7 +122,8 @@ func (o Options) ExecuteProcess() (int, error) {
 			case <-ctx.Done():
 			}
 		}()
-		code, err := wrapper.WaitForMarker(ctx, o.PreviousMarker)
+		prevMarkerResult := wrapper.WaitForMarkers(ctx, o.PreviousMarker)[o.PreviousMarker]
```
e-blackwelder, line 125: Is there a compelling reason to add the new local variable? I don't see `prevMarkerResult` being used elsewhere, so it seems functionally (but possibly not semantically) equivalent.

prow/entrypoint/run.go
anthonyrtong: Your right, I should remove `dir`. Go compiler didn't complain because it was accidentally left in the error msg.

prow/entrypoint/run.go
anthonyrtong: I used prevMarkerResult just because I changed `WaitForMarker()` to `WaitForMarkers()` which has different return values.  It has the same behavior, but slightly less clean because of the `WaitForMarkers()`'s more complicated use case.

anthonyrtong:
added the separate build-log uploads in this PR.  ptal @cjwagner

prow/sidecar/run_test.go
```diff
@@ -148,6 +148,152 @@ func TestWait(t *testing.T) {
...
+			markers:  []string{pass, aborted, fail},
+			abort:    true,
+			failures: 2,
+		},
+		{
+			name:     "fail when marker does not exist",
+			markers:  []string{pass},
+			missing:  true,
+			failures: 1,
+		},
```
cjwagner, line 204: Please also add a test case to ensure that the exit code we use to indicate a previous step failed is not counted as an additional failure.

prow/sidecar/run_test.go
```diff
@@ -299,28 +451,36 @@ func TestLogReader(t *testing.T) {
...
+				if err := ioutil.WriteFile(p, []byte(log), 0600); err != nil {
+					t.Fatalf("could not create log %s: %v", name, err)
 				}
 			}
 
-			buf, err := ioutil.ReadAll(logReader(entries))
-			if err != nil {
-				t.Fatalf("failed to read all: %v", err)
-			}
+			readers := logReaders((entries))
```
cjwagner, line 467: nit: Extra parenthesis here.

prow/sidecar/run_test.go
```diff
@@ -252,40 +398,46 @@ func name(idx int) string {
...
+			expected: map[string]string{
+				"test1-build-log.txt": "hello",
+				"test2-build-log.txt": "Failed to open test2-log.txt: whatever\n",
+				"test3-build-log.txt": "world",
 			},
 		},
 	}
 
-	re := regexp.MustCompile(`(?m)(Failed to open) .*log-\d.txt: .*$`)
+	re := regexp.MustCompile(`(?m)(Failed to open) .*log\.txt: .*$`)
```
cjwagner, line 440: `regexp.MustCompile` should only be used to declare package variables since it will result in a runtime panic and it is better if that is discovered at init time rather than when we try to use the regexp. I know this isn't an addition you made, but please move this outside the function in order to make it a package variable.

prow/sidecar/run.go
```diff
@@ -101,33 +112,29 @@ func (o Options) Run(ctx context.Context) (int, error) {
...
-	for i, opt := range entries {
-		ent := nameEntry(i, opt)
+func logReaders(entries []wrapper.Options) map[string]io.Reader {
+	readers := make(map[string]io.Reader)
+	for _, opt := range entries {
+		log, err := os.Open(opt.ProcessLog)
+		buildLog := "build-log.txt"
 		if len(entries) > 1 {
-			readers = append(readers, strings.NewReader(start(ent)))
+			buildLog = fmt.Sprintf("%s-build-log.txt", strings.TrimSuffix(filepath.Base(opt.ProcessLog), "-log.txt"))
```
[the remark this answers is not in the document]
anthonyrtong: Yeah I agree this would be much cleaner.  I will also need to make some very small changes to the podspec decoration to make sure to pass the container name in the wrapper options.

prow/sidecar/run_test.go
anthonyrtong: whoops

prow/sidecar/run_test.go
anthonyrtong: Would it make sense to test this for containers in parallel?  There isn't really any previous step since there isn't a deterministic ordering of containers in this case right?

prow/pod-utils/wrapper/options.go
```diff
@@ -78,49 +83,77 @@ func (o *Options) Validate() error {
...
+			results[path] = readMarkerFile(path)
 		}
+	}
+	watcher, err := fsnotify.NewWatcher()
+	if err != nil {
+		populateMapWithError(results, fmt.Errorf("new fsnotify watch: %v", err), paths...)
+		return results
+	}
+	defer watcher.Close()
+	dir := filepath.Dir(paths[0])
```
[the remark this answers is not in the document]
anthonyrtong: Yeah this is certainly assumption here.  It should be the case that all marker files and process logs are in the same directory, but yes it would be good to address this here too.

anthonyrtong:
ptal @cjwagner

prow/pod-utils/wrapper/options.go
```diff
@@ -38,6 +38,10 @@ type Options struct {
 	// Args is the process and args to run
 	Args []string `json:"args,omitempty"`
 
+	// ContainerName will contain the name of the container
+	// for the wrapped test process
+	ContainerName string `json:"container_name,omitempty"`
```
cjwagner, line 43: This new field is never set. Please update the decoration logic and tests.

PR #18282, opened 2020-07-10
Update presubmit config for gce pd repo

jingxu97:
change build from ci/latest to ci/k8s-master to make sure windows
binaries are build

### DIFF
--- a/config/jobs/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/gcp-compute-persistent-disk-csi-driver-windows.yaml
+++ b/config/jobs/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/gcp-compute-persistent-disk-csi-driver-windows.yaml
@@ -66,7 +66,7 @@ presubmits:
         args:
         - --check-leaked-resources
         - --cluster=
-        - --extract=ci/latest
+        - --extract=ci/k8s-master
         - --gcp-zone=us-west1-b
         - --ginkgo-parallel=8
         - --provider=gce


config/jobs/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/gcp-compute-persistent-disk-csi-driver-windows.yaml
```diff
@@ -66,7 +66,7 @@ presubmits:
         args:
         - --check-leaked-resources
         - --cluster=
-        - --extract=ci/latest
```
amwat, line 69: /lgtm

note that they aren't the same version, so i'm not entirely sure what this test aims to target.
this change will start using https://storage.googleapis.com/kubernetes-release-dev/ci/k8s-master.txt version
instead of https://storage.googleapis.com/kubernetes-release-dev/ci/latest.txt

Side note, since this looks like a presubmit, should this instead try to build binaries from source?
some more testing info here: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-testing/e2e-tests.md#extracting-a-specific-version-of-kubernetes


/hold

jingxu97: This presubmit job is for gce pd csi driver repo. It does not change anything in k/k repo, just use the e2e tests there. It will make build for gce-pd-csi repo. So I think k8s-master version should be good for this purpose.

amwat: /hold cancel

PR #18288, opened 2020-07-12
Trigger krte image push on kubekins-e2e variants change

hasheddan:
Updates post-test-infra-push-krte to be triggered every time
kubekins-e2e variants change because they are symlinked from the krte
image dir.

/assign @BenTheElder 
/cc @justaugustus  @kubernetes/ci-signal

### DIFF
--- a/config/jobs/kubernetes/test-infra/test-infra-trusted.yaml
+++ b/config/jobs/kubernetes/test-infra/test-infra-trusted.yaml
@@ -462,7 +462,7 @@ postsubmits:
         - images/kubekins-e2e/
   - name: post-test-infra-push-krte
     cluster: test-infra-trusted
-    run_if_changed: '^images/krte/'
+    run_if_changed: '^images/(krte/|kubekins-e2e/variants.yaml)'
     annotations:
       testgrid-dashboards: "sig-testing-images"
       testgrid-tab-name: "krte"
--- a/images/krte/wrapper.sh
+++ b/images/krte/wrapper.sh
@@ -39,10 +39,10 @@ printf '%0.s=' {1..80} >&2; echo >&2
 
 cleanup(){
   if [[ "${DOCKER_IN_DOCKER_ENABLED:-false}" == "true" ]]; then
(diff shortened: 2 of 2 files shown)


BenTheElder:
https://github.com/kubernetes/test-infra/blob/a4e899b6392d755057329df44ffacfe4653c1472/config/jobs/kubernetes/test-infra/test-infra-trusted.yaml#L465

Just needs to include the variants.yaml under kubekins.
The symlink does not change so the regex is WAI, it's just the wrong regex, which is why this feature is an anti-pattern.

This should have been changed when the symlink was added, but it's not intuitive.

hasheddan:
@BenTheElder thanks! will update!

hasheddan:
@BenTheElder will we also need a trivial change to the `krte` image dir to actually trigger this job?

config/jobs/kubernetes/test-infra/test-infra-trusted.yaml
```diff
@@ -462,7 +462,7 @@ postsubmits:
         - images/kubekins-e2e/
   - name: post-test-infra-push-krte
     cluster: test-infra-trusted
-    run_if_changed: '^images/krte/'
+    run_if_changed: '^(images/)(krte/|kubekins-e2e/variants.yaml)'
```
BenTheElder, line 465: The first capture group here is unnecessary, the second part is incorrectly split into two groups:
```suggestion
    run_if_changed: '^images/(krte/)|(kubekins-e2e/variants.yaml)'
```

hasheddan: @BenTheElder I believe the proposed change here would incorrectly match `kubekins-e2e/variants.yaml` with any preceding path. The first capture group is unnecessary, but it looks like your change is splitting the second part into two capture groups?

hasheddan: I removed first capture group and added trivial change in `images/krte/`

BenTheElder: Ack

justaugustus:
Great catch, @hasheddan!
/lgtm

PR #15612, opened 2019-12-16
releng: Remove bazel version markers

justaugustus:
~A previous PR changed the names and locations of the bazel version
markers and build artifacts. We fix that here to ensure the job is
properly forked on branch creation and update the 1.14 bazel build job
to adhere to the convention of the newer branches.~

~Figuring out the correct approach for this. I'll update the description once we decide.~

#### Update 3/15/20:
Bazel version markers appear to be unused.
In general, we should opt to use the version markers produced from the
ci-kubernetes-build* jobs instead, so we stop publishing the bazel ones.

Signed-off-by: Stephen Augustus <saugustus@vmware.com>

cc: @kubernetes/release-engineering 
ref: https://github.com/kubernetes/test-infra/pull/15610, https://github.com/kubernetes/kubernetes/issues/88553

PR #18290, opened 2020-07-13
releng: Publish fast builds to a separate subdirectory

justaugustus:
(https://github.com/kubernetes/test-infra/pull/18163 got thrashed w/ a bunch of old commits, so opening a new PR.)

  (This commit matches another in #18169, so it'll get rebased out of
  whichever PR lands last.)
- kubetest: Enable extract strategy for fast (linux/amd64-only) CI builds
- scenarios: Allow kubernetes_build to recognize fast builds

Inverts the logic in https://github.com/kubernetes/test-infra/pull/18158.
Explained in more detail here: https://github.com/kubernetes/release/pull/1389

/assign @BenTheElder @spiffxp @cblecker @dims 
/hold for review
ref: https://github.com/kubernetes/sig-release/issues/850, https://github.com/kubernetes/sig-release/issues/759

### DIFF
--- a/docs/kubernetes-versions.md
+++ b/docs/kubernetes-versions.md
@@ -1,71 +1,452 @@
-# Kubernetes Versions
+# Kubernetes Version Markers
 
-## CI Versions
+Version markers are text files which act as sort of a public API for accessing
+Kubernetes builds.
 
-Many periodic jobs related to `kubernetes/kubernetes` use an approximation of
-channel based versions.
+They are artifacts of a successful Kubernetes build which are stored in a
+Google Cloud Storage bucket alongside the builds they make reference to.
 
-At the time of writing (pre-`release-1.19` branch creation), these map as follows:
+You may see version markers leveraged in a variety of places, including:
 
-| channel    | branch       |
-| ---------- | ------------ |
-| ci/master  | master       |
-| ci/beta    | release-1.18 |
-| ci/stable1 | release-1.17 |
-| ci/stable2 | release-1.16 |
-| ci/stable3 | release-1.15 |
+- extraction strategies for e2e tests
+- Release Engineering tooling
+- external user-maintained scripts
 
-The names correspond to file pointers in the `kubernetes-release-dev` GCS
-bucket that are updated by CI
+## Table of Contents <!-- omit in toc -->
 
-```shell
-for suffix in master beta stable1 stable2 stable3; do
-  echo ci/k8s-$suffix: $(gsutil cat gs://kubernetes-release-dev/ci/k8s-$suffix.txt);
-done
+- [tl;dr](#tldr)
+  - [CI - cross build](#ci---cross-build)
+  - [CI - linux/amd64 (fast) build](#ci---linuxamd64-fast-build)
+  - [Release - Official release build](#release---official-release-build)
+  - [Release - Pre-release build](#release---pre-release-build)
+- [Usage](#usage)
+  - [Format](#format)
+  - [Access](#access)
+    - [gsutil](#gsutil)
+    - [curl/wget](#curlwget)
+    - [gcsweb](#gcsweb)
+  - [Querying a build](#querying-a-build)
+- [Marker types](#marker-types)
+  - [CI](#ci)
+    - [latest](#latest)
+    - [latest-fast](#latest-fast)
+    - [**DEPRECATED** - generic](#deprecated---generic)
+  - [Release](#release)
+    - [Official](#official)
+    - [Pre-release](#pre-release)
+- [Future Plans](#future-plans)
+- [Known Issues](#known-issues)
+  - [Generic version markers are not explicit](#generic-version-markers-are-not-explicit)
+  - [Manually created jobs using generic version markers can be inaccurate](#manually-created-jobs-using-generic-version-markers-can-be-inaccurate)
+- [Previous Issues](#previous-issues)
+  - [linux/amd64 version markers are colliding with cross builds](#linuxamd64-version-markers-are-colliding-with-cross-builds)
+  - [Cross builds are stored in a separate GCS bucket](#cross-builds-are-stored-in-a-separate-gcs-bucket)
+  - [Generated jobs may not represent intention](#generated-jobs-may-not-represent-intention)
+  - [bazel version markers appear to be unused](#bazel-version-markers-appear-to-be-unused)
 
-ci/k8s-master: v1.19.0-beta.2.607+4c853bb28f57f8
-ci/k8s-beta: v1.18.6-rc.0.15+e38139724f8f00
-ci/k8s-stable1: v1.17.9-rc.0.13+fdec14cd4c84b7
-ci/k8s-stable2: v1.16.13-rc.0.5+24212641c73bc0
-ci/k8s-stable3: v1.15.13-beta.0.1+a34f1e483104bd
+## tl;dr
+
+You need a...
+
+### CI - cross build
+
+Use `gsutil cat gs://kubernetes-release-dev/ci/latest.txt` (`master` branch)
+
+OR
+
+`gsutil cat gs://kubernetes-release-dev/ci/latest-x.y.txt`, where `x` is the
+Kubernetes major version and `y` is the Kubernetes minor version (release branches).
+
+### CI - linux/amd64 (fast) build
+
+Use `gsutil cat gs://kubernetes-release-dev/ci/latest-fast.txt` (**_only available
+on `master` branch_**).
+
+### Release - Official release build
+
+Use `gsutil cat gs://kubernetes-release/release/stable-x.y.txt`, where `x` is the
+Kubernetes major version and `y` is the Kubernetes minor version.
+
+### Release - Pre-release build
+
+Use `gsutil cat gs://kubernetes-release/release/latest-x.y.txt`, where `x` is the
+Kubernetes major version and `y` is the Kubernetes minor version.
+
+## Usage
+
+### Format
+
+All version markers have similar endpoints:
+
+```console
+<gcs-bucket>/<directory>/<marker>
 ```
 
-## Release Versions
+Expected output is [semver](https://semver.org/spec/v2.0.0.html)-compliant
+version, prepended with a `v`.
 
-There are also file pointers for release builds, these correspond to file 
-pointers in the `kubernetes-release` GCS bucket.
+Example:
 
-For example, for the latest release at the time of writing (pre-`release-1.19` branch creation):
+```console
+v1.20.0-alpha.0.391+575c4925be8c39
+```
 
-| channel             | tag                 |
-| ------------------- | ------------------- |
-| release/latest-1.15 | v1.15.13-beta.0 |
-| release/stable-1.15 | v1.15.12 |
-| release/latest-1.16 | v1.16.13-rc.0 |
-| release/stable-1.16 | v1.16.12 |
-| release/latest-1.17 | v1.17.9-rc.0 |
-| release/stable-1.17 | v1.17.8 |
-| release/latest-1.18 | v1.18.6-rc.0 |
-| release/stable-1.18 | v1.18.5 |
-| release/latest-1.19 | v1.19.0-beta.2 |
-| release/stable-1.19 | n/a |
+### Access
+
+Version markers are accessible via HTTP, so there are several ways to get them,
+depending on your use case.
+
+#### [gsutil](https://cloud.google.com/storage/docs/gsutil)
 
 ```shell
-for n in $(seq 15 19); do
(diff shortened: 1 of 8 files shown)


justaugustus:
Still some questions to answer from the previous PR:

From @BenTheElder:

> I'll review this properly Monday.
> 
> I would like to request though that we settle on something and then stop moving these around, these markers are the API for our CI builds.
> There's no good way to communicate moving these to all potential users and we want people to actually test these.

> still not in favor of continuing to rename these markers. Doing so affects all consumers and we have no way to communicate this effectively.
> 
> where was this discussed?

From @spiffxp:

> I've tried to follow the trail of issues and PR's and can't tell what the actual plan is, how conflicting version markers manifested, etc. There's too much followup/revert and no "this is the desired state we are reconciling to" that I can find?
> 
> I know we chatted on slack about moving toward version markers and job names that explicitly have the version in them. A plan of record would be useful

hakman:
@justaugustus Thanks for trying to find a better solution for this.
At the moment the current cross-builds marker `k8s-master` is pointing to a 2 weeks old build. The race condition between fast and cross builds pretty much prevents any cross-build from running, which makes multi-arch testing with CI builds impossible.
```
> curl -s https://storage.googleapis.com/kubernetes-release-dev/ci/k8s-master.txt
v1.19.0-beta.2.607+4c853bb28f57f8
```

Is there any way to force cross-builds to run at least on daily basis until a decision is reached here?

justaugustus:
> Is there any way to force cross-builds to run at least on daily basis until a decision is reached here?

@hakman -- I'd prefer to solve this instead of thinking of a workaround. I think we're close here.

justaugustus:
@BenTheElder @spiffxp -- More details, as requested.

Version markers are text files stored in the root of various GCS buckets:

- CI: https://gcsweb.k8s.io/gcs/kubernetes-release-dev/ci
- Release: https://gcsweb.k8s.io/gcs/kubernetes-release/release

They represent the results of different types of Kubernetes build jobs and act as sort of a public API for accessing builds. One can see them leveraged in extraction strategies for e2e tests, release engineering tooling, and user-created scripts.

Unfortunately, the way certain version markers are generated and utilized can at best be confusing, and at worst, disruptive.

There are a variety of problems, some of which are symptoms of the other ones...

### FIXED - Cross builds are stored in a separate GCS bucket

(Fixed in https://github.com/kubernetes/test-infra/pull/14030.)

This makes long-term usage of cross builds a little more difficult, since scripts utilizing version markers tend to consider only the version marker filename, while the GCS bucket name remains unparameterized.

### FIXED - Generated jobs may not represent intention

(Fixed in https://github.com/kubernetes/test-infra/pull/15564.)

As the generic version markers can shift throughout the release cycle, every time we regenerate jobs, they may not represent what we intend to test.

The best examples of this are pretty much every job using the `k8s-beta` version marker, and more specifically, skew and upgrade jobs.

### FIXED - bazel version markers appear to be unused

(Fixed in https://github.com/kubernetes/test-infra/pull/15612.)

### Generic version markers are not explicit

We publish a set of additional generic version markers:

- `k8s-master`
- `k8s-beta`
- `k8s-stable1`
- `k8s-stable2`
- `k8s-stable3`

Depending on the point in the release cycle, the meaning of these markers can change.

- `k8s-master` always points to the version on `master`.
- `k8s-beta` may represent:
  - `master`s build version (pre-branch cut)
  - a to-be-released build version (post-branch cut)
  - a recently released build version (post-release)

Knowing what these markers mean at any one time presumes knowledge of the build/release process or a correct interpretation of the [Kubernetes versions docs](https://github.com/kubernetes/test-infra/blob/master/docs/kubernetes-versions.md), which is often out of date and in low-visibility location.

### Manually created jobs using generic version markers can be inaccurate

Non-generated jobs using generic version markers do not get the same level of scrunity as ones that are generated via [`releng/test_config.yaml`][test_config.yaml].

This leads to inaccuracies between the versions presumed to be used in test and the versions that may be displayed in testgrid.

`ci-kubernetes-e2e-gce-beta-stable1-gci-kubectl-skew` is a great example:

https://github.com/kubernetes/test-infra/blob/96e08f4be2a86189f59c72055785f817ac346d30/config/jobs/kubernetes/sig-cli/sig-cli-config.yaml#L85-L112

All variants of that prowjob have landed on the `sig-release-job-config-errors` dashboard for various misconfiguration issues that are the result of generic version markers

### linux/amd64 version markers are colliding with cross builds

"Fast" (linux/amd64-only) builds run every 5 minutes, while cross builds run every hour.
They also write to the same version markers (`latest.txt`, `latest-<major>.txt`, `latest-<major>.<minor>.txt`).

The Kubernetes build jobs have a mechanism for checking if a build already exists and will exit early to save on test cycles.

What this means is if a "fast" build has already happened for a commit, then the corresponding cross build will exit without building.

This has been happening pretty consistently lately, so cross build consumers are using much older versions of Kubernetes than intended.

(Note that this condition only happens on `master`.)

---

I'd like to establish a rough plan of record to continue iteratively fixing some of these issues.

### Plan of record

- [x] (https://github.com/kubernetes/test-infra/pull/15564) Use explicit (`latest-x.y`) version markers in generated jobs
- [ ] (https://github.com/kubernetes/test-infra/pull/18290, https://github.com/kubernetes/release/pull/1389) Publish fast builds to separate subdirectory to prevent collisions
- [ ] (https://github.com/kubernetes/test-infra/pull/18169) Add a new `stable4` field to the kubernetes version lists in [`releng/test_config.yaml`][test_config.yaml] to remove confusion around jobs with `beta` in their name
- [ ] Refactor any non-generated jobs using generic version markers
- [ ] Refactor any config-forked release jobs using generic version markers
- [ ] Refactor any jobs using `fork-per-release-generic-suffix: "true"` annotations
- [ ] Disable usage of `fork-per-release-generic-suffix: "true"` annotations
- [ ] Rewrite the [Kubernetes versions doc](https://github.com/kubernetes/test-infra/blob/master/docs/kubernetes-versions.md) and put it in a more visible location
- [ ] Refactor [`releng/test_config.yaml`][test_config.yaml] to remove references to generic versions (e.g., prefer `ci-kubernetes-e2enode-ubuntu1-latest-1-19-gkespec` over `ci-kubernetes-e2enode-ubuntu1-k8sbeta-gkespec`)

[test_config.yaml]: https://github.com/kubernetes/test-infra/blob/master/releng/test_config.yaml

justaugustus:
> still not in favor of continuing to rename these markers. Doing so affects all consumers and we have no way to communicate this effectively.
> 
> where was this discussed?

@BenTheElder -- it's kind of scattered across Slack across multiple release cycles :sob: 

The previous recent PRs were specifically meant to not impact existing consumers.

This one:
- Moves the fast builds to a subdirectory (`/fast`)
- Enables an extraction strategy for fast builds

For people using `latest.txt`, after merge they'd be pointing at a cross build, instead of a fast build.
This could cause a transient failure in jobs using that marker while the next cross build is in progress, though I doubt it (tests using the latest marker are linux/amd64-only already).

For people using generic version markers, there should be no impact as we haven't changed the generic markers.

After the cross builds have started building again, I'd update jobs using generic version markers to use `latest` or `latest-x.y` markers instead.

justaugustus:
Recent Slack convo: https://kubernetes.slack.com/archives/C2C40FMNF/p1593720698263900
arm64 user report: https://kubernetes.slack.com/archives/C0BP8PW9G/p1594608658055800

dims:
Let's move the info / checklist to an issue (pretty please!) will get lost here. Also next time we should try to open an issue early so we have a record (i hate slack search!)

thanks,
Dims

justaugustus:
@dims -- Already opened and ref'ed in the PR description: https://github.com/kubernetes/sig-release/issues/850#issuecomment-659230693

I opened that last November and have been threading incremental fixes to it since.
(The plan of record and additional details are recent additions though.)

spiffxp:
Trying to recap what was discussed over zoom Thursday:
- I would like to see this PR update versions.md to reflect the current/desired state (in anticipation of this content moving somewhere more visible)
- We need to decide whether `latest.txt` should correspond to the results of a fast build (amd64-only) or a cross-build for the in-development release of kubernetes
  - cross build means waiting for ~80-120min after a PR merges before release-blocking CI jobs will start using a build containing that PR's changes - this is what all other release branches do
  - fast build means waiting only ~20-30min, but is the opposite of what all other release branches do
- We didn't have to make this choice before because a race condition was benefitting us? Fast builds would show up first and eventually get overwritten by cross builds? Couldn't quite remember this part...

detiber:
It would be great to get more generic descriptions of when to use which marker documented.

For example, It's not quite clear to me when I would use latest.txt rather than master.txt

It's also not overly clear to me when the beta transitions from a released version back to the state of master, is that only after beta builds are started for the next release?

Generic descriptions for `stablen` could likely be something like current stable release, n-1, etc
Though, I'm wondering why we couldn't use something a bit more clear and descriptive like:
- stable
- stable-1
- stable-2
- stable-3 (since it looks like the plan is to add another stable marker)

Then the naming itself might make the use cases a bit clearer.

Last thought, would it be possible to have a marker that represents:
- current state of master if all current release branches are "stable"
- current state of latest release branch if release is "pending"

That would allow for jobs to be defined in a way that they represent the latest build for the "next minor" version of kubernetes without having to potentially swap it between the yet to be released release branch and `master`, and target the "next patch" versions of supported released kubernetes versions using the "stable" markers.

jsturtevant:
I also ran into the issue with older builds using `k8s-master` when trying to fix the windows 1.19 release master-informing tests. We are currently failing due to older tests that have been removed but are not in the latest builds on `k8s-master`.  This build process is new to me but if I understand correctly I believe this would fix the issue by letting us use the `latest.txt`

> Depending on the point in the release cycle, the meaning of these markers can change.
> 
>     * `k8s-master` always points to the version on `master`.

I might be confused here but it seems like `k8s-master` is pointing to something 2 weeks old.  I was expecting it to point to builds on the master branch.   Adding to @detiber's comment, I think it would useful for the [release documentation](https://github.com/kubernetes/test-infra/blob/master/docs/kubernetes-versions.md) mentioned  to have use cases.  For example (I may not have these exactly right):

- if you want latest builds use `ci/master`
- if you want to the latest builds from a version (ex 1.19) use `ci/latest-1.19`

justaugustus:
I'm updating this PR today to include some guidance on version marker usage.

liggitt:
> I'm updating this PR today to include some guidance on version marker usage.

is waiting for this PR still the most reasonable thing for https://github.com/kubernetes/kubernetes/issues/93309#issuecomment-662596786 to do?

justaugustus:
@BenTheElder @spiffxp -- Version marker doc has been updated. PTAL.

docs/kubernetes-versions.md
```diff
@@ -1,71 +1,452 @@
...
+
+**kubekins-e2e `extract`:** `ci/k8s-<generic-version>.txt`
+
+The following generic markers are available:
+
+- `k8s-master`
+- `k8s-beta`
+- `k8s-stable1`
+- `k8s-stable2`
+- `k8s-stable3`
```
hakman, line 302: I think @detiber 's suggestion is a bit more intuitive:
```suggestion
- `k8s-stable`
- `k8s-stable-1`
- `k8s-stable-2`
...
```

justaugustus: This documents the current state, where "current" is what exists today in addition to what would happen once this PR merges.

There is no `k8s-stable` marker.

hakman: Got it. Thanks.

docs/kubernetes-versions.md
```diff
@@ -1,71 +1,452 @@
...
+
+Version markers are broken into two primary types: `CI` and `release` markers
+
+### CI
+
+**GCS Bucket:** `kubernetes-release-dev`
+
+**Directory:** `ci`
+
+#### latest
```
spiffxp, line 268: Formatting nit: it's visually difficult to tell the difference between these.  Maybe bullet the bolded items?

justaugustus:
@detiber @jsturtevant -- Can you read through the docs in this PR and let me know if they're sufficient for your needs?

Note: I'm only aiming to unblock cross builds here... any other topics around this should be discussed in https://github.com/kubernetes/sig-release/issues/850.

justaugustus:
Thanks @spiffxp!
I'm going to lift this so we can get the builds started, but also so I can reference the rewritten doc in the email to k-dev.

PR to follow to swap the markers on blocking jobs.
/hold cancel

justaugustus:
Opened https://github.com/kubernetes/test-infra/pull/18517 to fix up jobs using the `ci/k8s-master` version marker.

PR #18292, opened 2020-07-13
prow: descheduler: deploy 1.18.2 cluster and run make test-e2e

ingvagabund:
SSIA

PR #18298, opened 2020-07-13
Set KIND_E2E env var for e2e test script

damemi:
Following https://github.com/kubernetes/test-infra/pull/18292, and working with the changes in https://github.com/kubernetes-sigs/descheduler/pull/340, this sets an env var to specify that the Prow tests should be run with kind.


ingvagabund:
-lgtm

PR #18304, opened 2020-07-14
trigger: /retest should force failed jobs to rerun

petr-muller:
The interaction between `/retest` and conditionally executed jobs is
currently confusing. At least some users expect `/retest` to trigger all
failed jobs, even those that only run conditionally (`always_run: false`),
which was not true: the trigger criteria still applied.

I think that the expectation is valid though: once the job was for any
reason triggered and contributed a status to a PR, I find the
expectation that `/retest` will retest it reasonable.

This also makes the interaction with retest bot more intuitive. That bot
searches for jobs with failing tests and posts `/retest` on them. We had
saw a PR with a single contitionally executed job (manual trigger only)
failed, and the bot just continued to attempt to `/retest` without any
effect:

https://github.com/openshift/origin/pull/25195

### DIFF
--- a/prow/pjutil/filter.go
+++ b/prow/pjutil/filter.go
@@ -97,7 +97,8 @@ func FilterPresubmits(filter Filter, changes config.ChangedFilesProvider, branch
 // RetestFilter builds a filter for `/retest`
 func RetestFilter(failedContexts, allContexts sets.String) Filter {
 	return func(p config.Presubmit) (bool, bool, bool) {
-		return failedContexts.Has(p.Context) || (!p.NeedsExplicitTrigger() && !allContexts.Has(p.Context)), false, true
+		failed := failedContexts.Has(p.Context)
+		return failed || (!p.NeedsExplicitTrigger() && !allContexts.Has(p.Context)), false, failed
 	}
 }
 
--- a/prow/plugins/trigger/generic-comment_test.go
+++ b/prow/plugins/trigger/generic-comment_test.go
@@ -545,31 +545,6 @@ func TestHandleGenericComment(t *testing.T) {
 			},
 		},
 		{
-			name: "/retest of RunIfChanged job that doesn't need to run and hasn't run",
-
-			Author: "trusted-member",
-			Body:   "/retest",
-			State:  "open",
-			IsPR:   true,
-			Presubmits: map[string][]config.Presubmit{
-				"org/repo": {
-					{
-						JobBase: config.JobBase{
-							Name: "jeb",
-						},
-						RegexpChangeMatcher: config.RegexpChangeMatcher{
-							RunIfChanged: "CHANGED2",
-						},
-						Reporter: config.Reporter{
-							Context: "pull-jeb",
-						},
-						Trigger:      `(?m)^/test (?:.*? )?jeb(?: .*?)?$`,
-						RerunCommand: `/test jeb`,
-					},
-				},
-			},
-		},
-		{
 			name: "explicit /test for RunIfChanged job that doesn't need to run",
 
 			Author: "trusted-member",
--- a/prow/plugins/trigger/generic-comment_test.go
+++ b/prow/plugins/trigger/generic-comment_test.go
@@ -419,7 +419,7 @@ func TestHandleGenericComment(t *testing.T) {
 					},
 				},
 			},
-			ShouldBuild: false,
+			ShouldBuild: true,
 		},
 		{
 			name:   "Run if changed job triggered by /ok-to-test",
@@ -1070,7 +1070,7 @@ func TestRetestFilter(t *testing.T) {
 					},
 				},
 			},
-			expected: [][]bool{{true, false, true}, {false, false, true}},
+			expected: [][]bool{{true, false, true}, {false, false, false}},
 		},
 		{
 			name:           "retest filter matches jobs that would run automatically and haven't yet ",
@@ -1095,7 +1095,7 @@ func TestRetestFilter(t *testing.T) {
 					},
 				},
 			},
-			expected: [][]bool{{false, false, true}, {true, false, true}},
+			expected: [][]bool{{false, false, false}, {true, false, false}},
 		},
 	}
 
--- a/prow/config/jobs.go
+++ b/prow/config/jobs.go
@@ -354,12 +354,8 @@ func (ps Presubmit) ShouldRun(baseRef string, changes ChangedFilesProvider, forc
 	if forced {
 		return true, nil
 	}
-	if determined, shouldRun, err := ps.RegexpChangeMatcher.ShouldRun(changes); err != nil {
-		return false, err
-	} else if determined {
-		return shouldRun, nil
-	}
-	return defaults, nil
+	determined, shouldRun, err := ps.RegexpChangeMatcher.ShouldRun(changes)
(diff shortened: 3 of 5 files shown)


petr-muller:
I think this behavior is also consistent with the [documentation](https://github.com/kubernetes/test-infra/blob/master/prow/jobs.md#triggering-jobs-with-comments) saying:

> /retest : When posting /retest, two types of jobs will be triggered:
> - all jobs that have run and failed will run unconditionally
> - any not-yet-executed automatically run jobs will run conditionally

alvaroaleman:
Not sure about this.
/cc @fejta @cblecker

fejta:
Will this make `/retest` rerun all failed jobs? AKA optional ones, ones that do not report, ones that trigger on a branch/file/etc? Or just a subset (what subset)?

/assign @cjwagner 
for comment

cjwagner:
> What about retired contexts with failures?

These should be handled by the status reconciler.

It was definitely intended for`/retest` to rerun any jobs that were previously manually triggered and failed (perhaps with the exception of jobs that skip reporting, but I'm not sure if that case matters much). Basically if there are any red ProwJob statuses on the PR we should rerun them.

prow/pjutil/filter.go
```diff
@@ -97,7 +97,8 @@ func FilterPresubmits(filter Filter, changes config.ChangedFilesProvider, branch
 // RetestFilter builds a filter for `/retest`
 func RetestFilter(failedContexts, allContexts sets.String) Filter {
 	return func(p config.Presubmit) (bool, bool, bool) {
-		return failedContexts.Has(p.Context) || (!p.NeedsExplicitTrigger() && !allContexts.Has(p.Context)), false, true
+		failed := failedContexts.Has(p.Context)
+		return failed || (!p.NeedsExplicitTrigger() && !allContexts.Has(p.Context)), failed, true // force the trigger for failed jobs
```
cjwagner, line 101: The final return value is
```
//  - what the default behavior should be if the presubmit
//    runs conditionally and does not match trigger conditions
```
So I would have expected returning `true` as the last return value would cause the jobs to run if the filter matches (first return value) even if the trigger conditions are not met which could point to a bug elsewhere. @stevekuznetsov does that sound right?

petr-muller:
> Will this make /retest rerun all failed jobs? AKA optional ones, ones that do not report, ones that trigger on a branch/file/etc? Or just a subset (what subset)?

@fejta My intent is to trigger all jobs that have a failing status context on a PR, that *can* trigger on a PR (so not the ones not matching the brancher) , but would not necessarily be automatically triggered by e.g. a forced push of the current commits after an amend. The prime example would be an `always_run: false` job that was triggered manually via `/test` and failed.

I understand the set of triggered jobs would not include:
- ones that do not report (would not have a failing status context), unless they reported in the past on the PR
- ones that run on a different branch (would be filtered out by the `CouldRun` [check](https://github.com/kubernetes/test-infra/blob/ad24686e38736c3d111a07d0a96dd1291f2a0947/prow/config/jobs.go#L348-L350))

prow/pjutil/filter.go
petr-muller: Oh, I see what you mean! I misunderstood the code a bit, you are right. The case we were observing was indeed a job with `run_if_changed`, where *that* criteria were ruling the job out. If it were a job without `run_if_changed`, it would be triggered even with `always_run: false`, I guess.

In that case, the bug would be here I think:

https://github.com/kubernetes/test-infra/blob/ad24686e38736c3d111a07d0a96dd1291f2a0947/prow/config/jobs.go#L357-L362

The change regex matcher will explicity rule that job out because the job has a regex which does not match. The `default` will only apply when `determined==false`, which will only happen when the job does not have `run_if_changed`.

petr-muller: So... `return shouldRun || default` ?

fejta:
SGTM, Cole seems to have more context on the details of how this works so delegating it his way

prow/pjutil/filter.go
cjwagner: I looked through this again and that seems right to me.

petr-muller: Cool, I'll submit that change.

petr-muller:
@cjwagner PTAL. I still had to change even the `RetestFilter` for `default` to be `true` only for the failing jobs, not for the ones that get triggered by the "would run automatically but didn't yet" part of the `RetestFilter` condition.

matthyx:
/lgtm
@cjwagner PTAL

prow/pjutil/filter.go
```diff
@@ -97,7 +97,8 @@ func FilterPresubmits(filter Filter, changes config.ChangedFilesProvider, branch
 // RetestFilter builds a filter for `/retest`
 func RetestFilter(failedContexts, allContexts sets.String) Filter {
 	return func(p config.Presubmit) (bool, bool, bool) {
-		return failedContexts.Has(p.Context) || (!p.NeedsExplicitTrigger() && !allContexts.Has(p.Context)), false, true
+		failed := failedContexts.Has(p.Context)
+		return failed || (!p.NeedsExplicitTrigger() && !allContexts.Has(p.Context)), false, failed
```
cjwagner, line 101: > I still had to change even the `RetestFilter` for `default` to be `true` only for the failing jobs, not for the ones that get triggered by the "would run automatically but didn't yet" part of the `RetestFilter` condition.

I don't understand why this would be needed.

petr-muller: This is needed because `/retest` triggers two kinds of jobs ([doc](https://github.com/kubernetes/test-infra/blob/master/prow/jobs.md#triggering-jobs-with-comments), [comment](https://github.com/kubernetes/test-infra/blob/4afd9ba744d65d6a3ff028118a4adb1a00026e7d/prow/plugins/trigger/generic-comment.go#L171-L174)):

1. The ones that executed in the past and posted a failing status. We **do** want to ignore usual trigger conditions b/c we want to get rid of the failing status.
2. The ones that did not execute in the past (didn't post a status) but would be now triggered b/c of the trigger conditions. This part is basically `/test all` but only for jobs that did not run yet. We *don't* want to ignore trigger conditions for these, otherwise we would trigger **all** jobs.

I don't think the second case actually happens that often -- the case should be preempted by `statusreconciler`, but I guess it can apply for some corner cases after GH outages where statuses are not reported correctly and the like.

cjwagner: Ok, I see how this makes sense now when paired with the change to `Presubmit.ShouldRun`.

prow/plugins/trigger/generic-comment_test.go
```diff
@@ -544,31 +544,6 @@ func TestHandleGenericComment(t *testing.T) {
 				},
 			},
 		},
-		{
-			name: "/retest of RunIfChanged job that doesn't need to run and hasn't run",
```
cjwagner, line 548: ~Shouldn't we keep this test case to ensure jobs like this are not triggered?~ This was a duplicate test case, good catch.

cjwagner:
[flake](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/test-infra/18304/pull-test-infra-bazel/1293601141833076737)
/retest

PR #18307, opened 2020-07-14
create new Prow job to run promoter vulnerability presubmit check

yodahekinsew:
Creating the template for a new Prow job that will later be used to run the promoter's vulnerability presubmit check (https://github.com/kubernetes-sigs/k8s-container-image-promoter/issues/235) on pull requests that edit the promoter manifests. The template job won't be running over anything right now (only runs for the directory 'foo/bar'), but once the vulnerability check is finished the job can be used on a test directory for testing before running the presubmit on all the promoter manifests.

### DIFF
--- a/config/jobs/kubernetes/sig-release/cip/container-image-promoter.yaml
+++ b/config/jobs/kubernetes/sig-release/cip/container-image-promoter.yaml
@@ -21,6 +21,27 @@ presubmits:
         # /home/prow/go/src/github.com/{{.Org}}/{{.Repo}}, so just '.' should
         # suffice, but it's nice to be explicit.
         - -thin-manifest-dir=/home/prow/go/src/github.com/kubernetes/k8s.io/k8s.gcr.io
+  # Check that images to be promoted are free of vulnerabilities
+  - name: pull-k8sio-cip-vuln
+    annotations:
+      testgrid-create-test-group: 'true'
+      testgrid-dashboards: wg-k8s-infra-k8sio
+    decorate: true
+    skip_report: false
+    run_if_changed: 'k8s.gcr.io/((images/.*/images\.yaml)|(manifests/.*/promoter-manifest\.yaml))'
+    max_concurrency: 10
+    branches:
+    - ^vuln-check-test$
+    spec:
+      serviceAccountName: k8s-infra-gcr-vuln-scanning
+      containers:
+      - image: us.gcr.io/k8s-artifacts-prod/artifact-promoter/cip:20200723-v2.3.1-242-g7c328ba
+        command:
+        - cip
+        args:
+        - -thin-manifest-dir=/home/prow/go/src/github.com/kubernetes/k8s.io/k8s.gcr.io
+        - -vuln-severity-threshold=3
+        - -use-service-account
   # Check that changes to backup scripts are valid.
   - name: pull-k8sio-backup
     annotations:
--- a/config/prow/cluster/build_serviceaccounts.yaml
+++ b/config/prow/cluster/build_serviceaccounts.yaml
@@ -23,8 +23,8 @@ kind: ServiceAccount
 metadata:
   annotations:
     # Used by container image promoter vulnerability scanning presubmit check (pull-k8sio-cip-vuln)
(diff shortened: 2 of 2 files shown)


config/jobs/kubernetes/sig-release/cip/container-image-promoter.yaml
```diff
@@ -21,6 +21,26 @@ presubmits:
...
+      testgrid-create-test-group: 'true'
+      testgrid-dashboards: wg-k8s-infra-k8sio
+    decorate: true
+    skip_report: false
+    run_if_changed: 'k8s.gcr.io/((images/.*/images\.yaml)|(manifests/.*/promoter-manifest\.yaml))'
+    max_concurrency: 10
+    branches:
+    - ^master$
+    spec:
+      serviceAccountName: k8s-infra-gcr-promoter-vuln-scanning
```
spiffxp, line 36: /cc @fejta 
I don't think we allow presubmits to run as trusted service accounts that have special access.  Could you explain more about what this job is doing?

fejta: What cluster should this pod run in? Where is the k8s-infra-gcr-promoter-vuln-scanning service account defined?

yodahekinsew: @spiffxp this presubmit will check new images to be promoted for any vulnerabilities by looking up the image through the Container Analysis service. This is why the Container Analysis API was enabled on all of the image staging projects last week.

config/jobs/kubernetes/sig-release/cip/container-image-promoter.yaml
```diff
@@ -21,6 +21,27 @@ presubmits:
...
+      testgrid-create-test-group: 'true'
+      testgrid-dashboards: wg-k8s-infra-k8sio
+    decorate: true
+    skip_report: false
+    run_if_changed: 'k8s.gcr.io/((images/.*/images\.yaml)|(manifests/.*/promoter-manifest\.yaml))'
+    max_concurrency: 10
+    branches:
+    - ^vuln-check-test$
+    spec:
+      serviceAccountName: k8s-infra-gcr-vuln-scanning
```
spiffxp, line 36: https://github.com/kubernetes/test-infra/blob/master/config/prow/cluster/build_serviceaccounts.yaml#L21-L28 needs to be updated to use this shorter name

config/jobs/kubernetes/sig-release/cip/container-image-promoter.yaml
spiffxp: The kubernetes service account is defined for the (default) build cluster here: https://github.com/kubernetes/test-infra/blob/master/config/prow/cluster/build_serviceaccounts.yaml#L21-L28

The google service account is defined here: https://github.com/kubernetes/k8s.io/blob/master/infra/gcp/ensure-prod-storage.sh#L346-L350

It's given a role for each k8s-staging-* project in the kubernetes.io org here:
https://github.com/kubernetes/k8s.io/blob/4adcbd780a71d502a0d8fceb456f57ef7e7353af/infra/gcp/ensure-staging-storage.sh#L150-L153

That role appears to be `roles/containeranalysis.occurrences.viewer`: https://github.com/kubernetes/k8s.io/blob/8d556981bb18d33d9353d7b350b71794a0ae2021/infra/gcp/lib.sh#L186-L201

Which is defined here https://cloud.google.com/iam/docs/understanding-roles#container-analysis-roles

It's still not clear to me whether this is something I would want to allow any job from the public to run as? But it doesn't expose any destructive operations at least

spiffxp:
Manually pushed a vuln-check-test branch to k8s.io to support testing this https://kubernetes.slack.com/archives/CCK68P2Q2/p1595539448409600

config/jobs/kubernetes/sig-release/cip/container-image-promoter.yaml
yodahekinsew: Done

PR #18311, opened 2020-07-14
Remove Validate from Deserialize

chases2:
Cause Configurator to, when serializing to YAML only, not validate for completeness.

This is necessary since something that's an "obviously wrong, common error" (`days_of_results = 0`) in a proto is perfectly valid in YAML (days of results is not set; it should be the default value of the repository).

Needs testing as a Prow Job with the broken config (xref https://github.com/GoogleCloudPlatform/oss-test-infra/issues/437) to see if it resolves the issue.


chases2:
Tested on transfigure from that repository, and it seems to resolve the issue.

PR #18316, opened 2020-07-15
Slack reporter only report jobs if the state matches

chizhg:
According to the current logic, if one job has `reporter_config.slack.channel` field set on its config, Slack reporter will report the job status in every state even if it's not included in `job_states_to_report` in the Prow core config. With this logic, users won't be able to configure which states they want to report for a specific job.

This PR changes the logic to also allow overriding `job_states_to_report` and `report_template` in the job level, and only report the job that has `reporter_config.slack.channel` if it has a matched state. I have searched `reporter_config` in this repo and there is no job currently using it, so this change should not impact any Prow users.


chaodaiG:
> According to the current logic, if one job has `reporter_config.slack.channel` field set on its config, Slack reporter will report the job status in every state even if it's not included in `job_states_to_report` in the Prow core config. With this logic, users won't be able to configure which states they want to report for a specific job.

This change is a bit intrusive for existing users who expect slack messages on every single state of a job. Maybe start with `reporter_config.slack.states:[STATE1, STATE2..STATEn]`, and if this is not defined then assume all?

chizhg:
> > According to the current logic, if one job has `reporter_config.slack.channel` field set on its config, Slack reporter will report the job status in every state even if it's not included in `job_states_to_report` in the Prow core config. With this logic, users won't be able to configure which states they want to report for a specific job.
> 
> This change is a bit intrusive for existing users who expect slack messages on every single state of a job. Maybe start with `reporter_config.slack.states:[STATE1, STATE2..STATEn]`, and if this is not defined then assume all?

It looks there is no existing user for this, at least it's not used by k8s Prow.

I was considering the similar thing when I worked on the PR, just want to double check, when you say `not defined`, do you mean `not defined in the job config` or `not defined in the job config AND prow config`? My preference would be the latter since it provides consistent logic of overwriting.

chaodaiG:
> I was considering the similar thing when I worked on the PR, but just want to double check, when you say `not defined`, do you mean `not defined in the job config` or `not defined in the job config AND prow config`? My preference would be the latter since it provides consistent logic of overwriting.

My intention is the config defined in job could override prow config, but rethinking about it I don't even think prow level config makes sense for this, what would be the default?

chizhg:
> > I was considering the similar thing when I worked on the PR, but just want to double check, when you say `not defined`, do you mean `not defined in the job config` or `not defined in the job config AND prow config`? My preference would be the latter since it provides consistent logic of overwriting.
> 
> My intention is the config defined in job could override prow config, but rethinking about it I don't even think prow level config makes sense for this, what would be the default?

The Prow config is explained in https://github.com/kubernetes/test-infra/tree/master/prow/crier#slack-reporter (search for `slack_reporter_configs`), is it what you asked?

chaodaiG:
> The Prow config is explained in https://github.com/kubernetes/test-infra/tree/master/prow/crier#slack-reporter (search for `slack_reporter_configs`), is it what you asked?

I'm asking the use case, what if someone expects alerts on all states but the default is `failure`

chizhg:
> > The Prow config is explained in https://github.com/kubernetes/test-infra/tree/master/prow/crier#slack-reporter (search for `slack_reporter_configs`), is it what you asked?
> 
> I'm asking the use case, what if someone expects alerts on all states but the default is `failure`

I think I got confused on what you meant by `default`... 

The logic I'm considering now is:
1. The `job_states_to_report` in Prow config is empty by default, so no jobs will be reported.
2. Users can configure `job_states_to_report` as non-empty in */org/repo level, then all job states within that scope will be reported.
3. The logic I think worth adding: support configuring `job_states_to_report` in job level, if it's configured, it will overwrite the Prow config.

chaodaiG:
> > > The Prow config is explained in https://github.com/kubernetes/test-infra/tree/master/prow/crier#slack-reporter (search for `slack_reporter_configs`), is it what you asked?
> > 
> > 
> > I'm asking the use case, what if someone expects alerts on all states but the default is `failure`
> 
> I think I got confused on what you meant by `default`...
> 
> What I'm considering now is:
> 
> 1. The `job_states_to_report` in Prow config is empty by default, so no jobs will be reported.
> 2. Users can configure `job_states_to_report` as non-empty in */org/repo level in Prow config, then all job states within that scope will be reported.
> 3. _Here is the logic I think worth adding_: support configuring `job_states_to_report` in job level, if it's configured, it will overwrite the Prow config.

Agree with the logic you proposed above

chizhg:
> > > > The Prow config is explained in https://github.com/kubernetes/test-infra/tree/master/prow/crier#slack-reporter (search for `slack_reporter_configs`), is it what you asked?
> > > 
> > > 
> > > I'm asking the use case, what if someone expects alerts on all states but the default is `failure`
> > 
> > 
> > I think I got confused on what you meant by `default`...
> > What I'm considering now is:
> > 
> > 1. The `job_states_to_report` in Prow config is empty by default, so no jobs will be reported.
> > 2. Users can configure `job_states_to_report` as non-empty in */org/repo level in Prow config, then all job states within that scope will be reported.
> > 3. _Here is the logic I think worth adding_: support configuring `job_states_to_report` in job level, if it's configured, it will overwrite the Prow config.
> 
> Agree with the logic you proposed above

Done as we discussed.

prow/crier/reporters/slack/reporter.go
```diff
@@ -19,7 +19,7 @@ package slack
 import (
 	"bytes"
 	"fmt"
-	"text/template"
+	templatePkg "text/template"
```
chaodaiG, line 22: I thought package name should not contain capital letter

prow/crier/reporters/slack/reporter.go
```diff
@@ -86,32 +95,38 @@ func (sr *slackReporter) GetName() string {
...
 func (sr *slackReporter) ShouldReport(pj *v1.ProwJob) bool {
 	logger := sr.logger.WithFields(pjutil.ProwJobFields(pj))
-	// if a user specifically put a channel on their job, they want
-	// it to be reported regardless of what other settings exist
-	if _, set := jobChannel(pj); set {
-		logger.Debugf("reporting as channel is explicitly set")
-		return true
+	jobCfg := jobConfig(pj)
+	prowCfg := sr.config(pj.Spec.Refs)
+	// If a user specifically put a channel on their job, they want
```
chaodaiG, line 100: Is this comment still true?

prow/crier/reporters/slack/reporter.go
chizhg: It's fine to contain capital letter in the package alias I think, but generally it looks we should always use lower-case, so I changed the alias to `texttemplate`.

prow/crier/reporters/slack/reporter.go
chizhg: I updated the comment: `regardless of what other settings exist` => `regardless of the job types setting`

prow/crier/reporters/slack/reporter.go
```diff
@@ -39,25 +39,34 @@ type slackReporter struct {
 	dryRun bool
 }
 
-func jobChannel(pj *v1.ProwJob) (string, bool) {
-	if pj.Spec.ReporterConfig != nil && pj.Spec.ReporterConfig.Slack != nil && pj.Spec.ReporterConfig.Slack.Channel != "" {
-		return pj.Spec.ReporterConfig.Slack.Channel, true
+func jobConfig(pj *v1.ProwJob) *v1.SlackReporterConfig {
+	if pj.Spec.ReporterConfig != nil && pj.Spec.ReporterConfig.Slack != nil {
```
petr-muller, line 43: The ` && pj.Spec.ReporterConfig.Slack != nil` part is redundant - we just return that value, which can as well be `nil`.

chaodaiG:
other than the comment from petr-muller, lgtm from my perspective

prow/crier/reporters/slack/reporter.go
chizhg: Removed, thanks!

prow/crier/reporters/slack/reporter.go
```diff
@@ -98,32 +107,38 @@ func (sr *slackReporter) GetName() string {
...
-	// it to be reported regardless of what other settings exist
-	if _, set := jobChannel(pj); set {
-		logger.Debugf("reporting as channel is explicitly set")
-		return true
+	jobCfg := jobConfig(pj)
+	prowCfg := sr.getConfig(pj)
+	// If a user specifically put a channel on their job, they want
+	// it to be reported regardless of the job types setting.
+	jobShouldReport := false
+	if jobCfg != nil && jobCfg.Channel != "" {
```
alvaroaleman, line 115: Nit: Wouldn't all of this be easier expressed as `if jobCfg != nil && jobCfg.Channel != "" && (len(jobCfg.StatesToReport == 0 || stateSliceHasState(jobCfg.StatesToReport, job.Status.State) { return true }`?

A bit of a nitpicky comment but I had to read the tests to understand what you want to do

chizhg: It actually should be
```
if jobCfg != nil && jobCfg.Channel != "" && ((len(jobCfg.StatesToReport) == 0 && stateSliceHasState(prowCfg.StatesToReport, job.Status.State)) || stateSliceHasState(jobCfg.StatesToReport, job.Status.State)) { return true }
```
, which looks too long and not that expressive to me. I have updated the comments and hopefully it can explain the logic more clearly.

prow/crier/reporters/slack/reporter.go
```diff
@@ -19,7 +19,7 @@ package slack
 import (
 	"bytes"
 	"fmt"
-	"text/template"
+	texttemplate "text/template"
```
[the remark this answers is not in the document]
chizhg: Makes sense, renamed the function and removed the package alias.

prow/crier/reporters/slack/reporter_test.go
```diff
@@ -95,6 +96,66 @@ func TestShouldReport(t *testing.T) {
...
+					},
+				},
+				Status: v1.ProwJobStatus{
+					State: v1.FailureState,
+				},
+			},
+			expected: true,
+		},
+		{
+			name: "Job with channel config but does not have matched state should not report",
```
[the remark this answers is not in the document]
chizhg: Added the test. Thanks!

stevekuznetsov:
Love this change!

prow/crier/reporters/slack/reporter.go
alvaroaleman: You mean `len(jobCfg.StatesToReport) == 0 || stateSliceHasState(prowCfg.StatesToReport, job.Status.State)`?

The `|| stateSliceHasState(jobCfg.StatesToReport, job.Status.State)` part means its breaking though? Because today we report every job that has a channel regardless of state. I guess changing that makes sense, just wondering if it will surprise ppl

alvaroaleman: Could you maybe add something to prow/ANNOUNCEMENTS.md in the breaking change section? I agree that changing this makes sense but I have no idea if there are maybe ppl who expect todays behavior.
cc @xrstf

chizhg: I have added this as a breaking change to `prow/ANNOUNCEMENTS.md`.

To provide more context, our use case is we only want to report the `failure` and `error` states for a few specific Prow jobs, without this change we won't be able to do so.

alvaroaleman: Yeah, makes perfect sense. Thanks.

PR #18339, opened 2020-07-16
gce-ipvs: use gci node image

andrewsykim:
Some of the flexvolume tests are failing and aside from the proxy mode, the node image is the only difference
compared to other jobs where the flexvolume tests pass. Seeing if changing the node image to gci would fix the
flex volume tests

More details about test failures here: https://github.com/kubernetes/kubernetes/issues/92708 

Signed-off-by: Andrew Sy Kim <kim.andrewsy@gmail.com>

### DIFF
--- a/config/jobs/kubernetes/sig-network/sig-network-misc.yaml
+++ b/config/jobs/kubernetes/sig-network/sig-network-misc.yaml
@@ -30,7 +30,7 @@ presubmits:
         - --cluster=
         - --env=KUBE_PROXY_MODE=ipvs
         - --extract=local
-        - --gcp-node-image=ubuntu
+        - --gcp-node-image=gci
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes/sig-network/sig-network-misc.yaml
```diff
@@ -30,7 +30,7 @@ presubmits:
         - --cluster=
         - --env=KUBE_PROXY_MODE=ipvs
         - --extract=local
-        - --gcp-node-image=ubuntu
+        - --gcp-node-image=gci
```
aojea, line 33: @dims does this ring a bell? 
e2e flexvolumes tests and node images are related somehow?

aojea:
/lgtm
/approve
let's see :crossed_fingers:

PR #18340, opened 2020-07-16
Parallelize tree traversal for getAllClones

GeoBK:
Parallelize the upwards/downwards tree traversal for getAllClones

### DIFF
--- a/prow/bugzilla/client.go
+++ b/prow/bugzilla/client.go
@@ -24,10 +24,12 @@ import (
 	"net/http"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/sirupsen/logrus"
+	"golang.org/x/sync/errgroup"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 )
 
@@ -87,6 +89,43 @@ func NewClient(getAPIKey func() []byte, endpoint string) Client {
 		endpoint:  endpoint,
 		getAPIKey: getAPIKey,
 	}
+}
+
+// newBugDetailsCache is a constructor for bugDetailsCache
+func newBugDetailsCache() *bugDetailsCache {
+	return &bugDetailsCache{cache: map[int]Bug{}}
+}
+
+// bugDetailsCache holds the already retrieved bug details
+type bugDetailsCache struct {
+	cache map[int]Bug
+	lock  sync.Mutex
+}
+
+// get retrieves bug details from the cache and is thread safe
+func (bd *bugDetailsCache) get(key int) (bug Bug, exists bool) {
+	bd.lock.Lock()
+	defer bd.lock.Unlock()
+	entry, ok := bd.cache[key]
+	return entry, ok
+}
+
+// set stores the bug details in the cache and is thread safe
+func (bd *bugDetailsCache) set(key int, value Bug) {
+	bd.lock.Lock()
+	defer bd.lock.Unlock()
+	bd.cache[key] = value
+}
+
+// list returns a slice of all bugs in the cache
+func (bd *bugDetailsCache) list() []Bug {
+	bd.lock.Lock()
+	defer bd.lock.Unlock()
+	result := make([]Bug, 0, len(bd.cache))
+	for _, bug := range bd.cache {
+		result = append(result, bug)
+	}
+	return result
 }
 
 type client struct {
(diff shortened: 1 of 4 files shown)


GeoBK:
cc @alvaroaleman

prow/bugzilla/client.go
```diff
@@ -220,31 +275,106 @@ func (c *client) GetRootForClone(bug *Bug) (*Bug, error) {
...
-	clones = append(clones[:indexOfOriginal], clones[indexOfOriginal+1:]...)
-	return clones, nil
+	return nil
+}
+
+func traverseUp(c Client, bug *Bug, bugCache *BugDetailsCache, errGroup *errgroup.Group) error {
+	for _, dependsOnID := range bug.DependsOn {
+		dependsOnID := dependsOnID
+		errGroup.Go(func() error {
+			parent, isVisited, alreadyFetched := bugCache.Get(dependsOnID)
```
alvaroaleman, line 331: Why do we need `isVisisted`? Either its in the cache then we fetched it or its not, then we didn't fetch it

prow/bugzilla/client.go
```diff
@@ -89,6 +91,59 @@ func NewClient(getAPIKey func() []byte, endpoint string) Client {
...
+	return &BugDetailsCache{cache: map[int]cacheEntry{}}
+}
+
+type cacheEntry struct {
+	bug       *Bug
+	isVisited bool
+}
+
+// BugDetailsCache holds the already retrieved bug details
+type BugDetailsCache struct {
```
alvaroaleman, line 105: Nothing of the cache should be exported, this is a private implementation detail

prow/bugzilla/client.go
GeoBK: If in case we get the cache prepopulated with data(which would be the case if we do a prefetch, also the case if we do a GetBug for just the main bug), then isVisited is used to mark whether or not the the getLinkedBugs needs to be called on it.
ie to identify the case where the bug is fetched and we need to still call getLinkedBugs

alvaroaleman: Sorry I don't understand. There is no way to pre-populate the cache. Lets keep this scoped to just the traversal optimization. We always call `getLinkedBugs`. We should use the `alreadyFetched` check for short-circuiting where you are currently using the `isVisited`.

GeoBK:
@alvaroaleman - FYI I had a test failure saying that a race was detected. I think it was because I was passing pointers to the objects instead of the actual objects.
I have not had that issue since, but I would appreciate you keeping an eye out for that while reviewing the code. Thanks

prow/bugzilla/client.go
```diff
@@ -220,31 +259,105 @@ func (c *client) GetRootForClone(bug *Bug) (*Bug, error) {
...
+	var shouldWait bool
+	if errGroup == nil {
+		shouldWait = true
+		errGroup = new(errgroup.Group)
+	}
+	var err error
+	var bug *Bug
+	bugObj, cacheHasBug := bugCache.get(bugID)
+	if !cacheHasBug {
+		bug, err = c.GetBug(bugID)
```
alvaroaleman, line 294: just re-use the `bugObj` variable here and save yourself the allocation and the `else` branch

prow/bugzilla/client.go
```diff
@@ -220,31 +259,105 @@ func (c *client) GetRootForClone(bug *Bug) (*Bug, error) {
...
+
+func traverseUp(c Client, bug *Bug, bugCache *bugDetailsCache, errGroup *errgroup.Group) error {
+	for _, dependsOnID := range bug.DependsOn {
+		dependsOnID := dependsOnID
+		errGroup.Go(func() error {
+			parentObj, alreadyFetched := bugCache.get(dependsOnID)
+			if alreadyFetched {
+				return nil
+			}
+			parent := &parentObj
```
alvaroaleman, line 322: why use a new variable for the same pointer?

prow/bugzilla/client.go
```diff
@@ -220,31 +259,105 @@ func (c *client) GetRootForClone(bug *Bug) (*Bug, error) {
...
+			var err error
+			parent, err = c.GetBug(dependsOnID)
```
alvaroaleman, line 324: I believe you can simply do a
```
parentObj, err := c.GetBug(dependsOnID)`
```

prow/bugzilla/client_test.go
```diff
@@ -1040,27 +1040,28 @@ func TestGetAllClones(t *testing.T) {
 		{
 			"Clones including multiple children",
 			bug1,
```
alvaroaleman, line 1042: Can we make these tests readable by defining all the bugs and their relationship in the testcases rather than having them reference some bugs with indistinguishable names elsewhere?

prow/bugzilla/client.go
GeoBK: bugObj is type Bug, bug is of type *Bug

prow/bugzilla/client.go
GeoBK: parentObj is also of type Bug and parent is of type *Bug

prow/bugzilla/client.go
alvaroaleman: So its `*bugObj, err = c.GetBug(bugID)`

GeoBK: it spits out an error message saying its an invalid indirect for the above...

prow/bugzilla/client.go
```diff
@@ -89,6 +91,43 @@ func NewClient(getAPIKey func() []byte, endpoint string) Client {
...
+	bd.lock.Lock()
+	defer bd.lock.Unlock()
+	bd.cache[key] = value
+}
+
+// list returns a slice of all bugs in the cache
+func (bd *bugDetailsCache) list() []Bug {
+	bd.lock.Lock()
+	defer bd.lock.Unlock()
+	result := []Bug{}
```
petr-muller, line 124: Not sure how large these should be in practice but we can allocate a large enough slice right away and possibly avoid reallocation.

prow/bugzilla/client.go
alvaroaleman:
```diff
diff --git a/prow/bugzilla/client.go b/prow/bugzilla/client.go
index f5c15ee5e4..49350af88f 100644
--- a/prow/bugzilla/client.go
+++ b/prow/bugzilla/client.go
@@ -287,22 +287,19 @@ func getAllLinkedBugs(c Client, bugID int, bugCache *bugDetailsCache, errGroup *
                shouldWait = true
                errGroup = new(errgroup.Group)
        }
-       var err error
-       var bug *Bug
        bugObj, cacheHasBug := bugCache.get(bugID)
        if !cacheHasBug {
-               bug, err = c.GetBug(bugID)
+               bug, err := c.GetBug(bugID)
                if err != nil {
                        return err
                }
-       } else {
-               bug = &bugObj
+               bugObj = *bug
        }
        errGroup.Go(func() error {
-               return traverseUp(c, bug, bugCache, errGroup)
+               return traverseUp(c, &bugObj, bugCache, errGroup)
        })
        errGroup.Go(func() error {
-               return traverseDown(c, bug, bugCache, errGroup)
+               return traverseDown(c, &bugObj, bugCache, errGroup)
        })
```

prow/bugzilla/client.go
GeoBK: I have no way to know how large the clones set would be, but I guess 10 would be a reasonable number beyond which most would not go. I can make that change then

prow/bugzilla/client.go
GeoBK: Why is it better to do it in this way?

GeoBK: also bug wont be in scope for the else loop if I do that

alvaroaleman: There is no `else` block anymore. Its better because you save the `else` block and the preacllication of `var err, bug error *Bug`

prow/bugzilla/client.go
alvaroaleman: `len(cache.cache)` is the max size of the set.

prow/bugzilla/client_test.go
```diff
@@ -1004,74 +1004,78 @@ func errorChecker(err error, t *testing.T) {
...
-		},
-		{
-			"Clones when not directly related",
-			bug4,
-			sets.NewInt(bug1ID, bug2ID, bug3ID),
+			[]Bug{
+				{
+					Summary:   "Dummy bug to test getAllClones",
+					ID:        1,
+					DependsOn: []int{},
```
alvaroaleman, line 1019: You can just leave empty fields out. You can also leave the `Summary` out for everything that should match, empty string will match empty string. You can then probably just define the bug in a single line

prow/bugzilla/client_test.go
```diff
@@ -1004,74 +1004,78 @@ func errorChecker(err error, t *testing.T) {
...
+					ID:        4,
+					DependsOn: []int{1},
+				},
+				{
+					Summary:   "Dummy bug to test getAllClones",
+					ID:        5,
+					DependsOn: []int{1},
+				},
+			},
+			map[int]sets.Int{
```
alvaroaleman, line 1045: Why is this keyed by bug? It should be sufficient to compare a set of bugIDs (without any keying) to the result of `getClones` or not?

GeoBK: For the case of bug 4 which doesn't have any clones in this configuration, we would need a separate set of clones - which is why it is keyed

alvaroaleman: I don't follow. It just shouldn't be returned by `getCloenes`, because its not a clone?

GeoBK: Yes, bug4 is not a clone

alvaroaleman: Yeah so whats the point of having a set of each bug that is a clone that contains all bugs that are a clone and a set of nothing for the bug that is not a clone? Just use a set of all bugs you expect `getAllClones` to return intstead.

GeoBK: I would then have to add additional logic inside the test to identify the one bug which is not a clone and which ones are clones. With this approach I can just configure the clones in the map and not have to do anything else

alvaroaleman: No. Here on the testcase you define a set of expected clones (`sets.NewInt(1,2,3,5)` and in the test logic you construct a new set and compare:
```
				clones, err := getAllClones(fake, &bug, bugCache)
￼				if err != nil {
￼					t.Errorf("Error occurred when none was expected: %v", err)
￼				}
        actualClonesSet := sets.Int{}
        for _, clone := range clones { actualClonesSet.Insert(clone.ID }
        if !tc.expectedClones.Equals(actualCloneSet) { t.Errorf("returned clones do not match expected ones")}

GeoBK: Right now I am testing for all the bugs which I am passing in []bugs, do you want me to change that to test for only one bug per testcase? In this case I wouldn't need key the expected clones.

alvaroaleman: There is only one testcase already. I don't undrstand the question. The testcase tests the in and out of the func. Right now you use a super complicated way to test the out of the function. There is no need for that.

PR #18342, opened 2020-07-16
Undo image update

scottilee:
Undoing previous commit https://github.com/kubernetes/test-infra/pull/18195 until I can figure out a proper py3 container setup.

/assign @jlewi


jlewi:
@scottilee Can you rebase the change and fix the commit so that the CLA check passes?

jlewi:
@scottilee Didn't work; look at the commits
https://github.com/kubernetes/test-infra/pull/18342/commits

there are two commits. Notice how the first one isn't associated with a GitHub account? That's probably because the email on the commit doesn't match the one associated with your github account.

You need to rebase and squash the commits.

scottilee:
I could not find the commit in my history. I just created a new PR which passed the check here: https://github.com/kubernetes/test-infra/pull/18353. Closing this commit 😠

PR #18345, opened 2020-07-16
bugzilla: restrict automation for bug in groups

AlexNPavel:
This PR adds a new `allowed_groups` config to the bugzilla plugin
configuration and changes the behavior of the plugin to ignore bugs
that are part of groups not in the `allowed_groups` config.

/cc @stevekuznetsov

### DIFF
--- a/prow/plugins/bugzilla/bugzilla.go
+++ b/prow/plugins/bugzilla/bugzilla.go
@@ -446,6 +446,18 @@ func handle(e event, gc githubClient, bc bugzilla.Client, options plugins.Bugzil
 		}
 		if !isBugAllowed(bug, options.AllowedGroups) {
 			// ignore bugs that are in non-allowed groups for this repo
+			if refreshCommandMatch.MatchString(e.body) {
+				response := fmt.Sprintf(bugLink+" is in a bug group that is not in the allowed groups for this repo.", e.bugId, bc.Endpoint(), e.bugId)
+				if len(options.AllowedGroups) > 0 {
+					response += "\nAllowed groups for this repo are:"
+					for _, group := range options.AllowedGroups {
+						response += "\n- " + group
+					}
+				} else {
+					response += " There are no allowed bug groups configured for this repo."
+				}
+				return comment(response)
+			}
 			return nil
 		}
 	}
--- a/prow/plugins/bugzilla/bugzilla.go
+++ b/prow/plugins/bugzilla/bugzilla.go
@@ -438,6 +438,17 @@ func processQuery(query *emailToLoginQuery, email string, log *logrus.Entry) str
 
 func handle(e event, gc githubClient, bc bugzilla.Client, options plugins.BugzillaBranchOptions, log *logrus.Entry) error {
 	comment := e.comment(gc)
+	// check if bug is part of a restricted group
+	if !e.missing {
+		bug, err := getBug(bc, e.bugId, log, comment)
+		if err != nil || bug == nil {
+			return err
+		}
+		if !isBugAllowed(bug, options.AllowedGroups) {
+			// ignore bugs that are in non-allowed groups for this repo
+			return nil
+		}
+	}
 	// merges follow a different pattern from the normal validation
 	if e.merged {
 		return handleMerge(e, gc, bc, options, log)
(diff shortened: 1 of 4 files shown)


prow/plugins/bugzilla/bugzilla.go
```diff
@@ -438,6 +438,17 @@ func processQuery(query *emailToLoginQuery, email string, log *logrus.Entry) str
...
 func handle(e event, gc githubClient, bc bugzilla.Client, options plugins.BugzillaBranchOptions, log *logrus.Entry) error {
 	comment := e.comment(gc)
+	// check if bug is part of a restricted group
+	if !e.missing {
+		bug, err := getBug(bc, e.bugId, log, comment)
+		if err != nil || bug == nil {
+			return err
+		}
+		if !isBugAllowed(bug, options.AllowedGroups) {
+			// ignore bugs that are in non-allowed groups for this repo
```
stevekuznetsov, line 448: We don't have to leak which group the bug is in, but we should comment for the user to know that the bug is not in the allowed groups list, and list which groups are allowed. That way they can see feedback and we leak nothing.

prow/plugins/bugzilla/bugzilla_test.go
```diff
@@ -1121,8 +1121,8 @@ Instructions for interacting with me using PR comments are available [here](http
 			}},
 			prs:     []github.PullRequest{{Number: base.number, Merged: true}},
 			options: plugins.BugzillaBranchOptions{StateAfterMerge: &modified}, // no requirements --> always valid
-			expectedComment: `org/repo#1:@user: An error was encountered searching for external tracker bugs for bug 123 on the Bugzilla server at www.bugzilla:
-> injected error adding external bug to bug
```
stevekuznetsov, line 1125: Why did unrelated tests change?

AlexNPavel: Because the fake client is extremely non-specific with injected errors. We simply mark a bug to inject errors for and then that error occurs for any operation (except create bug). Since the first thing we do in handle now is get the bug, we hit the "getting bug" error before the error for adding an external bug, which used to be the first error we would hit in this unit test.

We would have to change the way the fake client injects errors and make it more specific/targeted based on operation to make prevent this test from changing.

prow/plugins/bugzilla/bugzilla.go
AlexNPavel: This would cause us to comment every time there is a change though. Would it make sense to only have it respond about the bug group if it was triggered by a `/bugzilla refresh` and ignore in all other cases?

stevekuznetsov: Yeah, that sounds good

stevekuznetsov:
@petr-muller however, a deny-list allows leaks and would require that the sensitive names be listed in configuration. An allow-list is a fail-safe.

petr-muller:
> @petr-muller however, a deny-list allows leaks and would require that the sensitive names be listed in configuration. An allow-list is a fail-safe.

@stevekuznetsov Okay. Out of curiosity, how would I express "no group"? Or is it not a thing in BZ?

AlexNPavel:
> @stevekuznetsov Okay. Out of curiosity, how would I express "no group"? Or is it not a thing in BZ?

By default, bugzilla bugs have no group (unless they are marked as private, which sets the group to "redhat" for redhat bugzilla) and I believe they can only be added to a group after they are created. Adding groups restricts which users are allowed to view the bug. Bugs with no group are considered public.

petr-muller:
@AlexNPavel that's what I expected, and that's why I'm wondering. Are we sure that bugs that need to be linked somewhere are always in a group? Sounds like many of them will not have a group, and with allowlist we have no way to allow such bugs to be linked...

AlexNPavel:
> @AlexNPavel that's what I expected, and that's why I'm wondering. Are we sure that bugs that need to be linked somewhere are always in a group? Sounds like many of them will not have a group, and with allowlist we have no way to allow such bugs to be linked...

This change doesn't block bugs that aren't in a group. It only blocks bugs if they are part of a group that's not allowed. The `isBugAllowed` function returns true for bugs with no groups.

PR #18348, opened 2020-07-17
Kubeflow - Requiring the Pipelines presubmits

Ark-kun:


Bobgy:
How does it work when `always_run` and `run_if_changed` are both present?

Ark-kun:
> How does it work when `always_run` and `run_if_changed` are both present?

Hmm. You're probably right. Removed `always_run`.

jlewi:
Rubber stamp
/lgtm
/approve

/hold In case you want to wait for @Bobgy to actually review.

Bobgy:
Thanks!
/lgtm
/unhold

PR #18349, opened 2020-07-17
Fixed space gap

Tej-Singh-Rana:


metrics/README.md
```diff
@@ -66,7 +66,7 @@ jqfilter: |
 To add a new metric, create a PR that adds a new yaml config file
 specifying the metric name (`metric`), the bigquery query to execute (`query`), and a
 jq filter to filter the data for the daily and latest files (`jqfilter`).
-*Optionally*: Include a jqfilter to extract influxdb timeseries measurements
+*Optionally*: Include a jq filter to extract influxdb timeseries measurements
```
spiffxp, line 69: The lack of space is intentional, since it's the name of the field used. Can you reformat to `jqfilter` with backticks instead?

Tej-Singh-Rana: I have to delete this entire "Optionally: Include a jqfilter to extract influxdb timeseries measurements from the raw query results (jqmeasurements)" ?

spiffxp: yes

PR #18352, opened 2020-07-17
Kops - Add Kubernetes 1.19 periodic e2e

hakman:
... and switch back to `ci/latest` and `KOPS_ARCH=amd64`.

/cc @rifelpet


rifelpet:
/lgtm
/approve 

We should also consider dropping older k8s version jobs when we add new ones

hakman:
> We should also consider dropping older k8s version jobs when we add new ones

Unfortunately, we still support all old versions. Getting rid of them is very slow...

PR #18362, opened 2020-07-17
Remove --backfill from metrics-bigquery job

spiffxp:
This was removed in an earlier PR, we forgot to update the job

ref: https://github.com/kubernetes/test-infra/pull/18203


spiffxp:
I happened to notice this failing https://testgrid.k8s.io/sig-testing-misc#metrics-bigquery

I feel like I should have seen e-mail for this, but that's probably a separate thing

PR #18372, opened 2020-07-19
Use test_args instead of cmd-test-args for windows-gce jobs

hasheddan:
Currently windows-gce jobs are providing ginkgo runtime config values
with the test-cmd-args flag, which is causing some skip and focus
directives to be ignored. This updates those jobs to use the test_args
flag.

Note: we will want to follow this with updates to release branch jobs if it proves successful.

Ref: https://github.com/kubernetes/kubernetes/issues/87389
Ref: https://github.com/kubernetes/kubernetes/issues/88974


yliaog:
thanks for the quick fix
/lgtm

PR #18377, opened 2020-07-20
enable gke presubmit job for sig-storage-local-static-provisioner

cofyc:


cofyc:
/assign @msau42 
This can be merged after https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/pull/207.

PR #18381, opened 2020-07-20
report ipv6 status on presubmits

aojea:
we had to remove the job because the job was unstable, confusing
developers.

KIND IPv6 jobs are more stable now, we also don't run storage tests
to avoid flakiness.

Fixes: https://github.com/kubernetes/kubernetes/issues/92886


aojea:
/hold
depends on https://github.com/kubernetes/kubernetes/issues/92965
/assign @BenTheElder @liggitt

aojea:
/hold cancel
this is quite stable and without the storage tests it will be in a better shape
feel free to let it soak for some days though

spiffxp:
/approve
I'd like to defer the /lgtm to one of the assignees

PR #18383, opened 2020-07-20
Auto redirect to spyglass for jobs with Spyglass link

MushuEE:
Spyglass is now the de facto view for jobs, it is no longer "New" I don't see a reason to not direct traffic there for jobs that have the option. I don't think many are using Gubernator for this anymore either. I can follow up with a PR that also directs the job history to the Spyglass version as well.

### DIFF
--- a/gubernator/view_build.py
+++ b/gubernator/view_build.py
@@ -259,6 +259,8 @@ class BuildHandler(view_base.BaseHandler):
         if external_config is not None:
             if external_config.get('spyglass'):
                 spyglass_link = 'https://' + external_config['prow_url'] + '/view/gcs/' + build_dir
+                self.redirect(spyglass_link)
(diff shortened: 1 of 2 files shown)


MushuEE:
/hold

not sure if that needs to be on it's own line

gubernator/view_build.py
```diff
@@ -259,6 +259,8 @@ def get(self, prefix, job, build):
         if external_config is not None:
             if external_config.get('spyglass'):
                 spyglass_link = 'https://' + external_config['prow_url'] + '/view/gcs/' + build_dir
+                self.redirect(spyglass_link)
```
spiffxp, line 262: If I knew for sure that nobody was using gubernator I would be fine with this.  There may be some internal users though?

PR #18387, opened 2020-07-20
Add preset-k8s-ssh to kubetest2 canary to fix log dump script issues

michaelmdresser:
The kubetest2 canary job currently functions fine, except for the log dump step. It runs into SSH/SCP permission issues, which can almost definitely be fixed by adding the `preset-k8s-ssh: "true"` label to the job config. This is based on [this job](https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/cloud-provider-gcp/cloud-provider-gcp-presubmits.yaml#L33) which ends up calling the same script that `kubetest2 gce` does for log dump.

Example failing job: https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_kubetest2/5/gce-build-up-down/1285269272859250692
Example success for the similar job that has the label: https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/cloud-provider-gcp/157/cloud-provider-gcp-e2e-create/1283917939208097793

This PR also moves the job description to be under the `kubernetes-sigs` folder to match the new location of the kubetest2 code.


amwat:
/cc @spiffxp @BenTheElder 
for approval

I'll add an OWNERS entry for these jobs as a follow up to make things easier. or do you want to add it as part of this PR itself?

PR #18396, opened 2020-07-20
Revert "Kettle build and build job are hitting timeouts with the default 10m"

MushuEE:
Reverts kubernetes/test-infra#18376

Apparently the max time allowed is [10min...](https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-test-infra-push-kettle/1285277075426512896)


spiffxp:
https://cloud.google.com/cloud-build/docs/api/reference/rest/v1/projects.builds#resource-build - I think you need to make timeout a top level field, instead of in the steps

MushuEE:
#18398

PR #18405, opened 2020-07-21
Kops - Use "kubekins-e2e" image "master" for all version tests

hakman:
This should fix https://github.com/kubernetes/kops/issues/9604, unless using master breaks something else.


mikesplain:
Thanks @hakman 

/lgtm
/approve

PR #18412, opened 2020-07-21
Remove issue comment on pull-kubernetes-node-e2e-containerd 

hasheddan:
This job was previously disabled in #18356 due to consistent timeout, which has
now been addressed by moving it to a new GCP project in #18409.

/cc @dims @liggitt

### DIFF
--- a/config/jobs/kubernetes/sig-node/sig-node-presubmit.yaml
+++ b/config/jobs/kubernetes/sig-node/sig-node-presubmit.yaml
@@ -82,7 +82,6 @@ presubmits:
   - name: pull-kubernetes-node-e2e-containerd
     branches:
     - master
-    # disabled pending resolution of https://github.com/kubernetes/kubernetes/issues/89847
     always_run: false
     optional: true
     skip_report: true


hasheddan:
@spiffxp @BenTheElder could one of you give a lgtm if this looks okay?

BenTheElder:
cc @liggitt 

we've been working to minimize the number of default presubmit jobs (as opposed to postsubmit, periodic) which are especially resource heavy (run on every pushed commit) and increase the odds of a flake blocking merges

BenTheElder:
I'm not sure what the cost/benefit is on this job specifically, or if we'd also wind up with yet another one for cri-o.

see also discussion in https://github.com/kubernetes/kubernetes/issues/92937

liggitt:
starting with this in a periodic job with good alerting, and a run_always: false PR job that can be manually triggered when desired would be my recommendation

if we see the periodic job getting broken frequently due to merged changes the other node e2e doesn't catch we can consider ways to prevent that

hasheddan:
@liggitt sounds good, will update

hasheddan:
@dims should this still have name `pull-kubernetes-node-e2e-containerd` as a periodic job?

dims:
@hasheddan we have a periodic CI job already if i remember right

hasheddan:
@dims there are a few e2e containerd node tests it looks like. If those are sufficient is any action needed here? I guess we could just drop the comment about re-enabling for now

dims:
@hasheddan +1

hasheddan:
@dims okay I think this is ready now :)

Issue #18397, opened 2020-07-20
Branchprotector is failing to update repos in the kubeflow org.


cjwagner:
**What happened**:
Branchprotector started failing between 
[1278462001802121218](https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/ci-test-infra-branchprotector/1278462001802121218) | 2020-07-01 22:54:51 +0000 UTC 
[1278446903859613702](https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/ci-test-infra-branchprotector/1278446903859613702) | 2020-07-01 21:54:52 +0000 UTC

with an error message that suggests many (all?) updates to the kubeflow repo have failed due to 404s while GETing the existing branch protection settings from GH.

[...]

PR #18415, opened 2020-07-21
Revert kubeflow branch protection config.

cjwagner:
Reverts https://github.com/kubernetes/test-infra/pull/18095 to fix https://github.com/kubernetes/test-infra/issues/18397

/assign @michelle192837 
cc @jlewi @Bobgy

### DIFF
--- a/config/prow/config.yaml
+++ b/config/prow/config.yaml
@@ -407,6 +407,7 @@ tide:
     labels:
     - lgtm
     - approved
+    - "cla: yes"
     missingLabels:
     - do-not-merge
     - do-not-merge/hold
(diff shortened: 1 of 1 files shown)


jlewi:
/lgtm
/approve

/hold 
@cjwagner  THe PR was supposed to address kubeflow/pipelines#1653 ; do you know if that problem was causing a different type of spam for prow oncall and if so is it worse then not rolling back? Since prow oncall is the one being negatively impacted I'll defer to whatever you want.

cjwagner:
> @cjwagner THe PR was supposed to address [kubeflow/pipelines#1653](https://github.com/kubeflow/pipelines/issues/1653) ; do you know if that problem was causing a different type of spam for prow oncall and if so is it worse then not rolling back? Since prow oncall is the one being negatively impacted I'll defer to whatever you want.

The PR causes a different type of spam and doesn't resolve the original spam so reverting it until the permissions are resolved seems like the right thing to do.

PR #18416, opened 2020-07-21
build cluster default CPU requests

BenTheElder:
/hold
cc @amerai @spiffxp @liggitt 
xref: https://github.com/kubernetes/kubernetes/issues/92937

### DIFF
--- a/config/prow/cluster/BUILD.bazel
+++ b/config/prow/cluster/BUILD.bazel
@@ -26,6 +26,11 @@ release(
         cluster = BUILD_CLUSTER,
     ),
     component(
+        "cpu-limit-range",
+        "limitrange",
+        cluster = BUILD_CLUSTER,
+    ),
+    component(
         "create-loop-devs",
         "daemonset",
         cluster = BUILD_CLUSTER,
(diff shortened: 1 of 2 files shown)


BenTheElder:
this may cause issues with jobs trying to request the whole node due to sidecar now having requests.
sidecar should probably have it's own lower request set, if it isn't already.

if not for that, it might even be reasonable for the test container to have a default of 1CPU, but we don't have a great way to target only that container.

spiffxp:
Is it worth identifying those jobs which might have issues ahead of time?

BenTheElder:
I'm not certain if they will

liggitt:
I may be imagining it, but I think kind pod schedule timeouts immediately started failing ~100% of the time on my PRs. Do we need to increase the scheduling timeout? Does the scheduler place the oldest pending pods?

BenTheElder:
we may need to adjust the CPU request, I intended to lift the hold when I
was monitoring for this.

BenTheElder:
https://github.com/kubernetes/test-infra/pull/18420

BenTheElder:
presubmit actually running again https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/92316/pull-kubernetes-e2e-kind/1285708641734758400

PR #18417, opened 2020-07-21
Cache for bugzilla client

GeoBK:


GeoBK:
@alvaroaleman

prow/bugzilla/client.go
```diff
@@ -82,10 +82,12 @@ type Client interface {
 	GetRootForClone(bug *Bug) (*Bug, error)
 }
 
-func NewClient(getAPIKey func() []byte, endpoint string) Client {
+// NewClient returns a bugzilla client. The httpclient is taken as an argument to allow
+// custom implementations of the http.Client
+func NewClient(getAPIKey func() []byte, endpoint string, httpclient *http.Client) Client {
```
alvaroaleman, line 87: There can not be a custom implementation of a `http.Client`, as its a concrete type, so the godoc is misleading.

Also, please allow to set a `http.RoundTripper` that will be used as transport rather than the whole client.

Lastly, what you are doing here is a breaking change, we can not do that. Please add a `SetRoundTripper` function to the Bugzilla Client interface instead.

GeoBK: I have added a SetRoundTripper to the Client interface, but this would cause breaking changes in any other implementations of Client interface since SetRoundTripper won't be implemented. 
lmk if this is not what you meant.

PR #18419, opened 2020-07-21
Make autobump with gerrit use Prow version instead of git diff to check if updates are needed.

cjwagner:
`git diff` will frequently indicate changes even if the bump PR doesn't need to be updated because it includes changes from other PRs that have merged in the repo since the last update to the bump PR.
This PR reverts the changes I made to address this comment: https://github.com/kubernetes/test-infra/pull/18240#discussion_r452505405

/assign @fejta


cjwagner:
Thanks Matthias!
/hold cancel

PR #18420, opened 2020-07-21
Default requests for podutils, adjust test-container requests

BenTheElder:
follow up to #18416 

/cc @spiffxp @liggitt


BenTheElder:
see also: https://github.com/openshift/release/blob/2e9d02df3e81caf26de674f4f1cb2425f687da64/core-services/prow/02_config/_config.yaml#L474

BenTheElder:
presubmit actually running again https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/92316/pull-kubernetes-e2e-kind/1285708641734758400

PR #18421, opened 2020-07-22
don't always run pull-kubernetes-dependencies-canary

BenTheElder:
see discussion at https://github.com/kubernetes/kubernetes/issues/92937#issuecomment-662177037

AFAICT @cblecker added this to test GOPROXY and we've not had a real user for it since we enabled that by default anyhow.


BenTheElder:
we could probably delete it, but it won't harm anything by being available to manually run

cblecker:
Retroactive LGTM

PR #18423, opened 2020-07-22
Order clones returned from getAllClones

GeoBK:


GeoBK:
@AlexNPavel - this change would also be needed to for the tests to pass - since the parallel traversal would make the order of  getAllClones random, a sort needs to be done to make the order deterministic

PR #18426, opened 2020-07-22
add hmac generator and a logger as an argument when creating a new github event server

droslean:
Since all users will always want the  hmac generator, this PR adds it as arg when creating a new github event server and removing them from the github event server options.

/cc @stevekuznetsov 

### DIFF
--- a/prow/githubeventserver/options.go
+++ b/prow/githubeventserver/options.go
@@ -20,23 +20,14 @@ import (
 	"flag"
 	"fmt"
 	"strings"
-
-	"github.com/sirupsen/logrus"
 )
 
 // Options holds the endpoint and port information that can be used
 // to create a new github event server
 type Options struct {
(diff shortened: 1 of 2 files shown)


prow/githubeventserver/options.go
```diff
@@ -20,48 +20,22 @@ import (
...
+func (o *Options) Validate() error {
 	if !strings.HasPrefix(o.endpoint, "/") {
 		return fmt.Errorf("endpoint %s is not a valid url path", o.endpoint)
 	}
-
-	if o.HmacTokenGenerator == nil {
-		o.HmacTokenGenerator = func() []byte { return []byte("") }
-	}
-
-	if o.Metrics == nil {
```
stevekuznetsov, line 57: If you were already doing this, then you do not need to change anything, right? Passing a `NewMetrics()` manually does nothing on top of the default.

droslean: updated.

PR #18428, opened 2020-07-22
add ListenAndServe and Shutdown functions to github event server

droslean:
Instead of running the HTTP server by using the interrupts library, implement the `ListenAndServe` and `Shutdown` functions to allow users to run directly the struct with `interrupts.ListenAndServe`
Add and define the HTTP server when creating a new GitHub event server and remove the `port` and `endpoint` fields from the struct.

/cc @stevekuznetsov 

### DIFF
--- a/prow/githubeventserver/githubeventserver.go
+++ b/prow/githubeventserver/githubeventserver.go
@@ -18,6 +18,7 @@ package githubeventserver
 
 import (
 	"bytes"
+	"context"
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
@@ -30,7 +31,6 @@ import (
 
 	"k8s.io/test-infra/prow/config"
 	"k8s.io/test-infra/prow/github"
-	"k8s.io/test-infra/prow/interrupts"
 	"k8s.io/test-infra/prow/pluginhelp"
 	pluginhelp_externalplugins "k8s.io/test-infra/prow/pluginhelp/externalplugins"
 	pluginhelp_hook "k8s.io/test-infra/prow/pluginhelp/hook"
@@ -54,11 +54,10 @@ const (
 type GitHubEventServer struct {
 	wg *sync.WaitGroup
 
-	endpoint string
-	port     int
-
 	serveMuxHandler *serveMuxHandler
 	httpServeMux    *http.ServeMux
+
+	httpServer *http.Server
 }
 
 // New creates a new GitHubEventServer from the given arguments.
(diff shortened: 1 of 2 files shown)


prow/githubeventserver/githubeventserver.go
```diff
@@ -92,9 +93,35 @@ func (g *GitHubEventServer) GetHTTPServeMux() *http.ServeMux {
 
 // Listen runs an http server by using the existing http.ServeMux, port and the
 // given grace period duration
-func (g *GitHubEventServer) Listen(gracePeriod time.Duration) {
+func (g *GitHubEventServer) Listen(ctx context.Context, gracePeriod time.Duration) {
 	defer interrupts.WaitForGracefulShutdown()
-	interrupts.ListenAndServe(&http.Server{Addr: ":" + strconv.Itoa(g.port), Handler: g.httpServeMux}, gracePeriod)
+
+	server := &http.Server{Addr: ":" + strconv.Itoa(g.port), Handler: g.httpServeMux}
+	interrupts.Run(func(ctx context.Context) {
```
stevekuznetsov, line 100: The caller uses `interrupts.Run()`. The server here just runs. If you are literally just running an HTTP server, expose it and allow `interrupts.ListenAndServe()`.

stevekuznetsov:
If the GitHubEventServer is simply running an HTTP server, this is all a caller should need to do in their `main.go`:

https://github.com/kubernetes/test-infra/blob/526277bee528e7308459b210517d1679f3944e8f/prow/cmd/hook/main.go#L235

prow/githubeventserver/githubeventserver.go
```diff
@@ -90,11 +89,9 @@ func (g *GitHubEventServer) GetHTTPServeMux() *http.ServeMux {
 	return g.httpServeMux
 }
 
-// Listen runs an http server by using the existing http.ServeMux, port and the
-// given grace period duration
-func (g *GitHubEventServer) Listen(gracePeriod time.Duration) {
-	defer interrupts.WaitForGracefulShutdown()
-	interrupts.ListenAndServe(&http.Server{Addr: ":" + strconv.Itoa(g.port), Handler: g.httpServeMux}, gracePeriod)
+// GetHTTPServer returns an http server using the github event server's port and http serve mux handler
+func (g *GitHubEventServer) GetHTTPServer() *http.Server {
```
stevekuznetsov, line 93: `interrupts.ListenAndServe()` just expects an interface -- can we make `GitHubEventServer` implement that interface, so no function is necessary to expose the internal server?

droslean: Thanks @stevekuznetsov. I should have done this from the beginning. Addressed

PR #18429, opened 2020-07-22
Spyglass multiple build logs

anthonyrtong:
- `ListArtifacts()` now supports listing the artifact names of multiple build-logs from multiple containers instead of a hardcoded `"build-log.txt"`.  `FetchArtifacts()` now supports fetching multiple `PodLogArtifact` artifacts.  
  - We can check the ProwJob CRD to find the names of containers in the ProwJob.  Pod utils will upload build-log's with the naming scheme:  `"<containername>-build-log.txt"`.  This way we can determine the names of all the build-logs that Spyglass needs to display.  
  - Once we determine the names of all of the ProwJob's build-logs, we can check whether these build-logs exist in gcs storage.  If the logs exist, then we do not have to stream them from the running Pod.
  - If we cannot access the ProwJob CRD because it has been garbage collected, then we can assume that all the build-logs for the ProwJob already exist in GCS since the job has finished running.
- `PodLogArtifact` can now support logs for any container not just `kube.TestContainerName`.  This allows us to retrieve the log artifacts for multiple containers.


prow/cmd/deck/main.go
```diff
@@ -511,8 +511,8 @@ type podLogClient struct {
 	client corev1.PodInterface
 }
 
-func (c *podLogClient) GetLogs(name string) ([]byte, error) {
-	reader, err := c.client.GetLogs(name, &coreapi.PodLogOptions{Container: kube.TestContainerName}).Stream()
+func (c *podLogClient) GetLogs(name string, container string) ([]byte, error) {
```
e-blackwelder, line 514: You can/should combine the types when they're the same. Something like this:
`... GetLogs(name, container string) ...`

prow/cmd/deck/main.go
```diff
@@ -1282,7 +1282,7 @@ func handlePluginHelp(ha *helpAgent, log *logrus.Entry) http.HandlerFunc {
 }
 
 type logClient interface {
-	GetJobLog(job, id string) ([]byte, error)
+	GetJobLog(job, id string, container string) ([]byte, error)
```
e-blackwelder, line 1285: Ditto. (You'll notice both "job" and "id" already share the same type.)

prow/spyglass/artifacts.go
```diff
@@ -50,29 +50,51 @@ func (s *Spyglass) ListArtifacts(ctx context.Context, src string) ([]string, err
...
-		artifactNames = append(artifactNames, "build-log.txt")
+
+	job, err := s.jobAgent.GetProwJob(jobName, buildID)
+	if err != nil {
+		return artifactNames, nil
+	}
+
+	jobContainers := job.Spec.PodSpec.Containers
+
+	if len(jobContainers) == 1 {
```
e-blackwelder, line 69: I think it would be nice to combine the single container use-case with the for-loop. (At the least, you could move the if/else _inside_ the for-loop.)

prow/spyglass/artifacts.go
```diff
@@ -50,29 +50,51 @@ func (s *Spyglass) ListArtifacts(ctx context.Context, src string) ([]string, err
...
 func (*Spyglass) KeyToJob(src string) (jobName string, buildID string, err error) {
-	src = strings.Trim(src, "/")
-	parsed := strings.Split(src, "/")
-	if len(parsed) < 2 {
-		return "", "", fmt.Errorf("expected at least two path components in %q", src)
-	}
-	jobName = parsed[len(parsed)-2]
-	buildID = parsed[len(parsed)-1]
-	return jobName, buildID, nil
+	return common.KeyToJob(src)
```
e-blackwelder, line 97: Nice.

prow/spyglass/podlogartifact_fetcher.go
```diff
@@ -34,11 +36,23 @@ func NewPodLogArtifactFetcher(ja jobAgent) *PodLogArtifactFetcher {
 }
 
 // artifact constructs an artifact handle for the given job build
-func (af *PodLogArtifactFetcher) Artifact(_ context.Context, jobName, buildID string, sizeLimit int64) (api.Artifact, error) {
-
-	podLog, err := NewPodLogArtifact(jobName, buildID, sizeLimit, af.jobAgent)
+func (af *PodLogArtifactFetcher) Artifact(_ context.Context, key string, artifactName string, sizeLimit int64) (api.Artifact, error) {
```
e-blackwelder, line 39: Can combine these method args, too.

prow/spyglass/podlogartifact_fetcher.go
```diff
@@ -34,11 +36,23 @@ func NewPodLogArtifactFetcher(ja jobAgent) *PodLogArtifactFetcher {
...
+	common.KeyToJob(key)
```
e-blackwelder, line 40: BUG: unused return value/called twice.

prow/spyglass/podlogartifact_fetcher.go
```diff
@@ -34,11 +36,23 @@ func NewPodLogArtifactFetcher(ja jobAgent) *PodLogArtifactFetcher {
...
 		return nil, fmt.Errorf("error accessing pod log from given source: %w", err)
 	}
 	return podLog, nil
 }
+
+func containerName(artifactName string) string {
+	if artifactName == "build-log.txt" {
+		return kube.TestContainerName
+	}
+	return artifactName[0 : len(artifactName)-len("-build-log.txt")]
```
e-blackwelder, line 57: Is there something simpler (and safer) that can do the same? (Maybe a "replace trailing `"-build-log.txt"` with `""`"?)

[Hint, hint](https://golang.org/pkg/strings/#TrimSuffix)

prow/spyglass/podlogartifact_fetcher_test.go
```diff
@@ -51,7 +52,7 @@ func TestFetchArtifacts_Prow(t *testing.T) {
 	}
 
 	for _, tc := range testCases {
-		artifact, err := goodFetcher.Artifact(context.Background(), tc.job, tc.buildID, maxSize)
+		artifact, err := goodFetcher.Artifact(context.Background(), fmt.Sprintf("%s/%s", tc.job, tc.buildID), "build-log.txt", maxSize)
```
e-blackwelder, line 55: I'm guessing the fmt.Sprintf returns the something-something-"key", right? Is there a (helper) function in use somewhere that already has this functionality? (If not, should there be?)

anthonyrtong:
ptal @michelle192837 @chaodaiG @cjwagner

prow/spyglass/podlogartifact_fetcher.go
```diff
@@ -34,11 +36,22 @@ func NewPodLogArtifactFetcher(ja jobAgent) *PodLogArtifactFetcher {
...
+	containerName := containerName(artifactName)
+	podLog, err := NewPodLogArtifact(jobName, buildID, artifactName, containerName, sizeLimit, af.jobAgent)
 	if err != nil {
 		return nil, fmt.Errorf("error accessing pod log from given source: %w", err)
 	}
 	return podLog, nil
 }
+
+func containerName(artifactName string) string {
+	if artifactName == "build-log.txt" {
```
chaodaiG, line 53: This should be a const defined somewhere else too

prow/spyglass/artifacts.go
anthonyrtong: I am keeping a single container use-case just for backwards compatibility.  I will move it inside the for loop to be cleaner.

prow/spyglass/podlogartifact_fetcher.go
anthonyrtong: Whoops

anthonyrtong:
ptal @alvaroaleman

matthyx:
/lgtm
@anthonyrtong you can rename you PR to drop WIP if it's ready

prow/spyglass/lenses/common/common.go
```diff
@@ -206,16 +204,16 @@ func FetchArtifacts(
 			_, err = art.Size()
 		}
 		if err != nil {
-			if name == "build-log.txt" {
-				podLogNeeded = true
+			if matched, err := regexp.MatchString(".*build-log.txt", name); err == nil && matched {
```
alvaroaleman, line 207: The common way for such a static regex would be to define it as a package level variable like this, then you don't need the error checking and will discover during tests if it fails to compile:

`var buildLogRegex = regexp.MustCompile("regex string")`

alvaroaleman: Also, not introduced by you, but why do we have both this and the `required_files` option? Do we ignore the `required_files` option or why is it possible to see the buildlog lense before there is anything in gcs?

prow/spyglass/artifacts.go
```diff
@@ -50,29 +50,47 @@ func (s *Spyglass) ListArtifacts(ctx context.Context, src string) ([]string, err
...
+	}
+
+	jobContainers := job.Spec.PodSpec.Containers
+
+	for _, container := range jobContainers {
+		logName := SingleLogName
+		if len(jobContainers) > 1 {
+			logName = fmt.Sprintf("%s-build-log.txt", container.Name)
+		}
+		if !contains(artifactNames, logName) {
```
alvaroaleman, line 74: Not sure if it makes sense here already, but generally there is `k8s.io/apimachinery/pkg/util/sets` which makes working with sets easier

prow/spyglass/lenses/common/common.go
anthonyrtong: > Also, not introduced by you, but why do we have both this and the `required_files` option? Do we ignore the `required_files` option or why is it possible to see the buildlog lense before there is anything in gcs?

Yeah, so AFAICT we use this `ListArtifacts()` function in `cmd/deck/main.go` [here](https://github.com/kubernetes/test-infra/blob/e883d5a5273deafe70aa54754e1d4cb73f307c3a/prow/cmd/deck/main.go#L952) in order to list all the artifacts that either exist in gcs _or_ that we assume will exist in gcs in the future.  We use these artifact names to check against the `required_files` [here](https://github.com/kubernetes/test-infra/blob/e883d5a5273deafe70aa54754e1d4cb73f307c3a/prow/cmd/deck/main.go#L966).  We are able to see the buildlog lense even if there is no build-log in gcs because we explicitly add the buildlog artifact names in `ListArtifacts()`.  Before, this was done by adding a just a hardcoded `"build-log.txt"` and in this PR, we add buildlog artifact names by getting the container names.  Regardless, this logic is used so that we still render the buildlog lense even if there is not build-logs in gcs because we want still try to stream in the build-logs from the prowjob.

prow/spyglass/artifacts.go
anthonyrtong: Yes I agree it would probably makes sense to use sets instead. I'll make that change.

prow/spyglass/podlogartifact_fetcher_test.go
```diff
@@ -27,31 +27,37 @@ func TestFetchArtifacts_Prow(t *testing.T) {
...
+			artifact:  "build-log.txt",
+			expectErr: true,
+		},
+		{
+			name:      "Fetch log with no artifact name",
+			key:       "BFG/435",
+			artifact:  "",
 			expectErr: true,
 		},
 	}
```
cjwagner, line 57: Please also include tests for the multiple container case.

cjwagner: Same for `podlogartifact_test.go` and `spyglass_test.go`.

prow/spyglass/artifacts_test.go
```diff
@@ -66,9 +66,7 @@ func TestSpyglass_ListArtifacts(t *testing.T) {
 			args: args{
 				src: "gs/test-bucket/logs/example-ci-run/404",
 			},
-			want: []string{
-				"build-log.txt",
-			},
+			want: []string{},
```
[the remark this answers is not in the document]
anthonyrtong: I changed this testcase because `ListArtifacts()` gets the build log name from the job instead of assuming the name is "build-log.txt".  Because of this I thought the unit test couldn't actually test adding the missing build log name when it doesn't exist in persistent storage, but I just realized I can probably test this properly if I change the fake job agent to handle this.

prow/spyglass/artifacts.go
```diff
@@ -50,29 +51,40 @@ func (s *Spyglass) ListArtifacts(ctx context.Context, src string) ([]string, err
...
+
+	jobContainers := job.Spec.PodSpec.Containers
+
+	for _, container := range jobContainers {
+		logName := SingleLogName
+		if len(jobContainers) > 1 {
+			logName = fmt.Sprintf("%s-build-log.txt", container.Name)
+		}
+		if artifactNamesSet.Has(logName) {
+			artifactNamesSet.Insert(logName)
```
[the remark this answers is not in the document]
anthonyrtong: Yeah it was supposed to be reverse. I'll be sure to fix and test this.

prow/spyglass/artifacts.go
anthonyrtong: Oh yeah that would make way more sense.

prow/cmd/deck/main.go
```diff
@@ -1292,12 +1292,16 @@ func handleLog(lc logClient, log *logrus.Entry) http.HandlerFunc {
 		w.Header().Set("Access-Control-Allow-Origin", "*")
 		job := r.URL.Query().Get("job")
 		id := r.URL.Query().Get("id")
-		logger := log.WithFields(logrus.Fields{"job": job, "id": id})
+		container := r.URL.Query().Get("container")
```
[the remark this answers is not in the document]
anthonyrtong: I see.  I'm not entirely sure what the desired behavior should be for this if we have multiple logs.  If there's only 1 icon for a job log, what would we want to display to a user when they click on it?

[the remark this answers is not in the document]
cjwagner: We probably should update that button to have a drop down list for multiple containers or something like that. In the meantime we could assume spyglass is enabled and use a spyglass link, but that isn't ideal.

alvaroaleman:
@anthonytong nit: Could you either squash your commit or ass the tide squash label so it does that on merge?

prow/cmd/deck/static/api/prow.ts
```diff
@@ -122,3 +122,10 @@ export interface ProwJobStatus {
   jenkins_build_id?: string;
   prev_report_states?: { [key: string]: ProwJobState };
 }
+
+// PodSpec is a description of a pod.
+// PodSpec mirrors the PodSpec struct defined in k8s.io/api/core/v1
+// Podspec interface only holds containers right now since no other values are used
+export interface PodSpec {
+  containers: object[];
+}
```
anthonyrtong, line 131: I added this so that deck would know how many containers a job has. This way it'll know whether the build-log icon should lead to a single container build-log or to spyglass where users can see multiple. I'm not sure if I should add all of the other podspec fields to this just for consistency.

anthonyrtong:
ptal @cjwagner

anthonyrtong:
ptal @matthyx

matthyx:
Seems like you have addressed all @cjwagner remarks...
/lgtm

prow/cmd/deck/static/prow/prow.ts
```diff
@@ -571,7 +572,17 @@ function redraw(fz: FuzzySearch): void {
         r.appendChild(cell.state(state));
         if ((agent === "kubernetes" && pod_name) || agent !== "kubernetes") {
             const logIcon = icon.create("description", "Build log");
-            logIcon.href = `log?job=${job}&id=${build_id}`;
+            if (pod_spec == null || pod_spec.containers.length <= 1) {
+                logIcon.href = `log?job=${job}&id=${build_id}`;
+            } else {
+                const buildIndex = url.indexOf('/build/');
```
cjwagner, line 578: When does the URL contain `/build/`? I don't see any examples of that on https://prow.k8s.io/data.js.

prow/cmd/deck/template/index.html
```diff
@@ -2,7 +2,7 @@
 
 {{define "scripts"}}
 <script type="text/javascript" src="/static/prow_bundle.min.js"></script>
-<script type="text/javascript" src="prowjobs.js?var=allBuilds&omit=annotations,labels,decoration_config,pod_spec"></script>
+<script type="text/javascript" src="prowjobs.js?var=allBuilds&omit=annotations,labels,decoration_config"></script>
```
cjwagner, line 5: Including the pod_spec in the results seems to increase the GET time by >50% on average according to some checks I did with `time curl`. That seems a little excessive given that we only need to know if there are multiple containers. Can we determine this more efficiently or perhaps only check if the logs button is actually clicked?

prow/spyglass/podlogartifact.go
```diff
@@ -71,23 +78,27 @@ func (a *PodLogArtifact) CanonicalLink() string {
 		"job": []string{a.name},
 		"id":  []string{a.buildID},
 	}
+	if a.artifactName != singleLogName {
+		q.Set("container", a.container)
+	}
```
cjwagner, line 83: Shouldn't we also use the `buildLogRegex` here (or the `containerName` func) so that we can set the `container` query parameter in the multiple container case as well?

prow/spyglass/podlogartifact_fetcher_test.go
```diff
@@ -26,37 +28,76 @@ func TestFetchArtifacts_Prow(t *testing.T) {
...
 			expectErr: true,
 		},
+		{
+			name:      "Fetch log with no artifact name",
+			key:       "BFG/435",
+			artifact:  "",
+			expectErr: true,
+		},
+		{
+			name:         "Fetch custom build-log with no artifact name",
```
cjwagner, line 65: This example includes an artifact name.

prow/spyglass/spyglass_test.go
```diff
@@ -1240,22 +1279,22 @@ func TestFetchArtifactsPodLog(t *testing.T) {
 	}
 
 	for _, key := range testKeys {
-		result, err := sg.FetchArtifacts(context.Background(), key, "", 500e6, []string{"build-log.txt"})
+		containers := []string{"test-1", "test-2"}
+		result, err := sg.FetchArtifacts(context.Background(), key, "", 500e6, []string{fmt.Sprintf("%s-%s", containers[0], singleLogName), fmt.Sprintf("%s-%s", containers[1], singleLogName)})
```
cjwagner, line 1283: It looks like the passed string slice is only appropriate for the multiple container case, but this includes tests for the single container case as well.

prow/spyglass/podlogartifact.go
anthonyrtong: This sets the container query parameter _only_ in the multiple container use case, otherwise it leaves the container query parameter as empty, in which case we automatically assume that the container name is kube.TestContainerName.  Regardless, this if statement seems unnecessary and I should remove it.

prow/cmd/deck/static/prow/prow.ts
anthonyrtong: It doesn't seem like any of our urls for jobs have `/build/`, however I saw this logic for the [spyglass button](https://github.com/kubernetes/test-infra/blob/af45e5598c0395ad9b249e0031cbc1270b0e068e/prow/cmd/deck/static/prow/prow.ts#L618-L625) here and thought that it might be useful for some backwards compatibility reason.  If that's not the case, then yeah it should probably be removed since it doesn't seem to be relevant now.

prow/spyglass/podlogartifact_fetcher_test.go
anthonyrtong: Typo. Not sure why I thought it even makes sense to have a custom build log with no name 😅

prow/spyglass/spyglass_test.go
anthonyrtong: It seemed like this test was just testing that FetchArtifacts() would fetch PodLogArtifacts when the build-logs don't exist in gcs storage.  I kinda assumed that if it worked for multiple containers, then it would work for a single container so I changed it to test just multiple containers.  For completeness, it makes sense that I add back the single container use case.  Additionally, this made me just realize that it would be good to add a testcase for cases with multiple build logs where some build logs are in gcs and some need to be streamed, so I'll go ahead and add that as well.

prow/cmd/deck/template/index.html
anthonyrtong: I think I found a way to solve this issue by changing the way handleProwJobs() works in `deck/main.go.`  If instead of completely setting the podspec to nil when omitted we just set the podspec to be nothing but a list of empty containers, this gives us the information we need in order to achieve the same behavior.  From the testing that I've done locally with `time curl` it seems like the GET time is not nearly as horrendous as retrieving the entire podspec.

prow/cmd/deck/main.go
```diff
@@ -798,7 +798,17 @@ func handleProwJobs(ja *jobs.JobAgent, log *logrus.Entry) http.HandlerFunc {
 					jobs[i].Spec.DecorationConfig = nil
 				}
 				if set.Has(PodSpec) {
-					jobs[i].Spec.PodSpec = nil
+					// when we omit the podspec, we don't set it completely to nil
+					// instead, we set it to a new podspec that just has an empty container for each container that exists in the actual podspec
+					// this is so we can determine how many containers there are for a given prowjob without fetching all of the podspec details
+					// this is necessary for prow/cmd/deck/static/prow/prow.ts to determine whether the logIcon should link to a log endpoint or to spyglass
+					emptyContainers := []coreapi.Container{}
+					for range jobs[i].Spec.PodSpec.Containers {
```
cjwagner, line 806: It isn't safe to assume the `PodSpec` field is non-nil. We should only initialize a `PodSpec` with empty containers if a PodSpec was specified. Please add a test case for this.

prow/cmd/deck/static/prow/prow.ts
```diff
@@ -571,7 +572,18 @@ function redraw(fz: FuzzySearch): void {
...
+            } else {
+                // this logic exists for legacy jobs that are configured for gubernator compatibility
+                const buildIndex = url.indexOf('/build/');
+                if (buildIndex !== -1) {
+                    const gcsUrl = `${window.location.origin}/view/gcs/${url.substring(buildIndex + '/build/'.length)}`;
+                    logIcon.href = gcsUrl;
+                } else if (url.includes('/view/')) {
+                    logIcon.href = url;
+                }
+            }
```
cjwagner, line 586: Please ensure that we always set `logIcon.href`. We should probably default to using the `log?job=${job}&id=${build_id}` format if we can't determine an appropriate spyglass link.

prow/cmd/deck/main.go
anthonyrtong: Not sure if I'm just not able to find it, but it doesn't seem like we have any tests for this handler at all.  I'll go ahead and add some in main_test.go

PR #18431, opened 2020-07-22
reduce CPU request for job in wg-k8s-infra cluster

BenTheElder:
should resolve https://github.com/kubernetes/kubernetes/issues/93334

### DIFF
--- a/config/jobs/kubernetes/sig-testing/conformance-e2e.yaml
+++ b/config/jobs/kubernetes/sig-testing/conformance-e2e.yaml
@@ -226,8 +226,11 @@ periodics:
           # these are both a bit below peak usage during build
           # this is mostly for building kubernetes
           memory: "9000Mi"
-          # during the tests more like 3-20m is used
-          cpu: 7400m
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes/sig-testing/conformance-e2e.yaml
```diff
@@ -227,7 +227,7 @@ periodics:
           # this is mostly for building kubernetes
           memory: "9000Mi"
           # during the tests more like 3-20m is used
-          cpu: 7400m
+          cpu: 7300m
```
liggitt, line 230: any chance of a comment explaining how we arrived here?

BenTheElder: this approximately is the remaining CPU on the node after the sidecar started using 100m and the GKE overhead.
visibility into this is poor. this was not meant to be a lasting plan.

liggitt:
ok...

/lgtm

BenTheElder:
I wrote up a quick comment on this.
It's not great.

PR #18437, opened 2020-07-22
add golangci-lint verification for vSphere CSI Driver

divyenpatel:
Adding `golangci-lint` verification for vSphere CSI Driver Pull Requests.

We have recently added support for validating vSphere CSI Driver code using  `golangci-lint` and `make golangci-lint` target is added to do this verification. - https://github.com/kubernetes-sigs/vsphere-csi-driver/blob/master/Makefile#L363-L364


divyenpatel:
@codenrhoden @dvonthenen Can you help with the review?

codenrhoden:
Verified that this `make` target existed in the current tip of the repo and it passed successfully.

/lgtm
/approve

PR #18438, opened 2020-07-22
Increase controller-runtime PR tests CPU/memory

vincepri:
/assign @alvaroaleman

### DIFF
--- a/config/jobs/kubernetes-sigs/controller-runtime/controller-runtime-periodics-master.yaml
+++ b/config/jobs/kubernetes-sigs/controller-runtime/controller-runtime-periodics-master.yaml
@@ -13,7 +13,7 @@ periodics:
       - ./hack/ci-check-everything.sh
       resources:
         requests:
-          cpu: 6000m
+          cpu: "7000m"
   annotations:
     testgrid-dashboards: sig-api-machinery-kubebuilder
     testgrid-tab-name: controller-runtime-periodic-master
--- a/config/jobs/kubernetes-sigs/controller-runtime/controller-runtime-presubmits-master.yaml
+++ b/config/jobs/kubernetes-sigs/controller-runtime/controller-runtime-presubmits-master.yaml
@@ -14,7 +14,7 @@ presubmits:
         - ./hack/ci-check-everything.sh
         resources:
           requests:
-            cpu: 6000m
+            cpu: "7000m"
     annotations:
       testgrid-dashboards: sig-api-machinery-kubebuilder
       testgrid-tab-name: controller-runtime-master
--- a/config/jobs/kubernetes-sigs/controller-runtime/controller-runtime-presubmits-release-0.6.yaml
+++ b/config/jobs/kubernetes-sigs/controller-runtime/controller-runtime-presubmits-release-0.6.yaml
@@ -14,7 +14,7 @@ presubmits:
         - ./hack/ci-check-everything.sh
         resources:
           requests:
-            cpu: 4000m
+            cpu: "7000m"
     annotations:
       testgrid-dashboards: sig-api-machinery-kubebuilder
       testgrid-tab-name: controller-runtime-release-0.5


config/jobs/kubernetes-sigs/controller-runtime/controller-runtime-periodics-master.yaml
```diff
@@ -2,18 +2,22 @@ periodics:
...
+  labels:
+      preset-service-account: "true"
   extra_refs:
   - org: kubernetes-sigs
     repo: controller-runtime
     base_ref: master
   spec:
     containers:
-    - image: golang:1.13
+    - image: gcr.io/k8s-testimages/kubekins-e2e:v20200720-4766100-1.18
```
alvaroaleman, line 13: why kubekind, why service account?

vincepri:
I also changed the base images to match what we're using in Cluster API, which will allow us to be in line with the upstream dependencies as well

config/jobs/kubernetes-sigs/controller-runtime/controller-runtime-periodics-master.yaml
```diff
@@ -2,18 +2,22 @@ periodics:
...
       command:
+      - runner.sh
       - ./hack/ci-check-everything.sh
       resources:
         requests:
-          cpu: 6000m
+          cpu: "6000m"
```
alvaroaleman, line 19: The nodes have 7400M allocatable and 16 gig ram, I am reasonable sure this won't make a difference because we get an exclusive node anyways: https://kubernetes.slack.com/archives/C09QZ4DQB/p1594929742176700?thread_ts=1594858427.174300&cid=C09QZ4DQB

config/jobs/kubernetes-sigs/controller-runtime/controller-runtime-periodics-master.yaml
vincepri: I matched the same that we're using on the Cluster API side, I forget why the service account was needed, although I could look tomorrow if necessary (logs or something?). The main reason that `kubekins` is nice is that the images are hosted on GCR, and the embedded go version is the same that Kubernetes has been built with. For example `1.18` here uses Go 1.13. When in the future we change the built-in dependencies to 1.19, we should build with Go 1.15, the images would already have it.

config/jobs/kubernetes-sigs/controller-runtime/controller-runtime-periodics-master.yaml
vincepri: That's interesting, Should we try to allocate ~7 then to make sure nothing else is scheduled? AFAIK, there is lots of jobs that don't have explicit CPU/memory requests that could still get scheduled, unless there is a higher default sometwhere

config/jobs/kubernetes-sigs/controller-runtime/controller-runtime-periodics-master.yaml
vincepri: I don't think we need the service account for this project, will remove it.

alvaroaleman: Yeah but we pull a gigantic image we don't need and use wrapping that makes reproducing everything harder. I think updating the go minor version on the job when we update the kube dependencies shouldn't be too difficult

config/jobs/kubernetes-sigs/controller-runtime/controller-runtime-periodics-master.yaml
alvaroaleman: Yeah from what I can see we should be requesting 7400m to guarantee exlusivity: a81e294b95401dddeb1c8fea5514f8a6b1ef11af

config/jobs/kubernetes-sigs/controller-runtime/controller-runtime-periodics-master.yaml
vincepri: Fixed

config/jobs/kubernetes-sigs/controller-runtime/controller-runtime-periodics-master.yaml
vincepri: Placed it to 7000m to have some headroom, happy to bump to 7300 or 7400

PR #18445, opened 2020-07-23
Fixed bug causing dir level to always show 100% coverage

### LINKED ISSUE #18443: Gopherage junit reports top level dir coverage as 100% not average of subdir
**What happened**:
`./gopherage junit --threshold 0.05 filtered.cov > junit_coverage.xml` will build junit xml for individual files without issue, however reporting the toplevel dir always resolves to 100%, not the average of the sub file coverage.

```
    <testcase class_name="go_coverage" name="k8s.io/kubernetes/cluster/images/conformance/go-runner/env.go" time="0">
        <properties>
            <property name="coverage" value="85.7"></property>
        </properties>
    </testcase>
    <testcase class_name="go_coverage" name="k8s.io/kubernetes/cluster/images/conformance/go-runner/tar.go" time="0">
        <properties>
            <property name="coverage" value="81.5"></property>
        </properties>
    </testcase>
    <testcase class_name="go_coverage" name="k8s.io/kubernetes/cluster/images/conformance/go-runner" time="0">
        <properties>
            <property name="coverage" value="100.0"></property>
        </properties>
    </testcase>
```
```
    <testcase class_name="go_coverage" name="k8s.io/kubernetes/cluster/images/conformance/go-runner" time="0">
        <properties>
            <property name="coverage" value="83.6"></property>
        </properties>
    </testcase>
```
Run `gopherage junit` against any subset of a `.cov` file with < 100% coverage values

[Here](https://testgrid.k8s.io/sig-testing-canaries#ci-kubernetes-coverage-unit) with [this](https://storage.googleapis.com/kubernetes-jenkins/logs/ci-kubernetes-coverage-unit/1286322423590817792/artifacts/junit_coverage.xml)
/assign

MushuEE:
The input was being edited, not the new CoverageList... leading to
0 values on all attrs and causing 100% to be reported. Unsure how this even worked before

#18443

### DIFF
--- a/gopherage/pkg/cov/junit/calculation/coveragelist.go
+++ b/gopherage/pkg/cov/junit/calculation/coveragelist.go
@@ -44,8 +44,8 @@ func (covList *CoverageList) Ratio() float32 {
 // summarize summarizes all items in the Group and stores the result
 func (covList *CoverageList) summarize() {
 	for _, item := range covList.Group {
-		covList.Coverage.NumCoveredStmts += item.NumCoveredStmts
-		covList.Coverage.NumAllStmts += item.NumAllStmts
+		covList.NumCoveredStmts += item.NumCoveredStmts
+		covList.NumAllStmts += item.NumAllStmts
 	}
 }
 
--- a/gopherage/pkg/cov/junit/calculation/coveragelist.go
+++ b/gopherage/pkg/cov/junit/calculation/coveragelist.go
@@ -43,6 +43,8 @@ func (covList *CoverageList) Ratio() float32 {
 
 // summarize summarizes all items in the Group and stores the result
 func (covList *CoverageList) summarize() {
+	covList.NumCoveredStmts = 0
+	covList.NumAllStmts = 0
 	for _, item := range covList.Group {
 		covList.NumCoveredStmts += item.NumCoveredStmts
 		covList.NumAllStmts += item.NumAllStmts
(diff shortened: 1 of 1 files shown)


MushuEE:
@michaelkolber

MushuEE:
/hold
really did something ugly in Git. Fixing

gopherage/pkg/cov/junit/calculation/coveragelist.go
```diff
@@ -43,8 +43,6 @@ func (covList *CoverageList) Ratio() float32 {
 
 // summarize summarizes all items in the Group and stores the result
 func (covList *CoverageList) summarize() {
-	covList.NumCoveredStmts = 0
```
BenTheElder, line 46: if we don't zero this, repeated calls are going to keep summing higher and higher?

MushuEE: Each object is new each time `s := newCoverageList("Filtered Summary: " + prefix)` so we will never iterate over the same one twice. That being said. If we did, I would think we wouldn't want to re-zero so that a subdir contains the results for all coverage results with that matching prefix.

MushuEE: https://github.com/kubernetes/test-infra/blob/2035c42220f2ab250958f28c78255d24f895e949/gopherage/pkg/cov/junit/xmlwriter.go#L85-L86

BenTheElder: > Each object is new each time s := newCoverageList("Filtered Summary: " + prefix) so we will never iterate over the same one twice. 

In the current code. But as an API it doesn't make sense to repeatedly add to the same fields if called twice.

> // summarize summarizes all items in the Group and stores the result

I would not expect two calls to summarize() to produce a 2x count when accessed.
One way or another this clearly needs refactoring.

MushuEE: I added them back in. based on the doc for that method I think reseting makes sense. Just because it currently isn't recalled. Doesn't mean it wont ever be!

Issue #18386, opened 2020-07-20
pull-test-infra-gubernator	failing


michelle192837:
From Job History, looks like it's been failing since the 17th (https://prow.k8s.io/job-history/gs/kubernetes-jenkins/pr-logs/directory/pull-test-infra-gubernator).

Example @ https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/test-infra/18355/pull-test-infra-gubernator/1285260858326257665

/assign @fejta 
/area gubernator

spiffxp:
Will be fixed by https://github.com/kubernetes/test-infra/pull/18391

michelle192837:
Looks like this is still failing :(

https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/test-infra/18414/pull-test-infra-gubernator/1285989995366060032
https://github.com/kubernetes/test-infra/pull/18414

michelle192837:
Oh, though there _are_ some passing runs, so it's not just failing.

spiffxp:
https://github.com/kubernetes/test-infra/pull/18448 should fix and cause future PRs to fail this check prior to merging

spiffxp:
/close
Closing in favor of https://github.com/kubernetes/test-infra/issues/18770

PR #18446, opened 2020-07-23
Remove gcs/ suffix from job_url_prefix_config.

cjwagner:
Update to address the change here: https://github.com/kubernetes/test-infra/pull/17779

I don't think we should be deprecating the old job url format, but I'm raising my concerns on that PR. Changing this for now since it is causing tons of log spam.

/assign @chaodaiG @e-blackwelder


cjwagner:
/override pull-test-infra-gubernator
https://github.com/kubernetes/test-infra/issues/18386

PR #18448, opened 2020-07-23
fix gubernator presubmit, try catching this in the future

spiffxp:
https://github.com/kubernetes/test-infra/pull/18421 caused pull-test-infra-gubernator to start failing again

this should hopefully cause presubmits to fail if a similar job config change occurs


hack/verify-all.sh
```diff
@@ -19,3 +19,4 @@ set -o pipefail
 
 cd "$(git rev-parse --show-toplevel)"
 find hack -name 'verify-*.sh' -not -name "$(basename "$0")" \( -print -exec '{}' ';' -o -quit \)
+./guberantor/verify_config.sh
```
spiffxp, line 22: guberANTor

PR #18451, opened 2020-07-23
increase default grace priod to match bootstrap.py

BenTheElder:
we've been moving more jobs over to `decroate: true`

currently the default grace period for handling timeout interrupt is only 15s, on an overloaded cluster / with non trivial cleanup this is too short, and we wind up falling back on e.g. boskos.

bootstrap.py had a fixed grace period of 15m, and plenty of things rely on that. 
for kubernetes prow.k8s.io only I am moving the default grace period to match bootstrap.py.

individual jobs can still choose to have shorter grace periods, and well behaved signal handlers with trivial cleanup should exit quickly anyhow

/cc @amwat @MushuEE 
FYI @liggitt

### DIFF
--- a/config/prow/config.yaml
+++ b/config/prow/config.yaml
@@ -9,7 +9,7 @@ plank:
   default_decoration_configs:
     '*':
       timeout: 2h
-      grace_period: 15s
+      grace_period: 15m
       utility_images:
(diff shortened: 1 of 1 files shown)


liggitt:
+1

PR #18453, opened 2020-07-23
Migrate cloud-provider-gcp e2e-create test to use kubetest2

michaelmdresser:
Because no current test container includes a kubetest2 binary, this test must clone and build kubetest2 before running the test. In future, the kubekins container will be replaced by one that already includes a kubetest2 binary.

The current kubetest2 exec tester does not properly handle the kubeconfig set by the deployer, probably because it can run any command and thus doesn't always need a kubeconfig. Therefore, when `kubectl` is run normally it looks for the default config location instead of the deployer-generated config location and then it fails because it tries to connect to the default (`localhost:8080`). Thus we have to manually specify the path of the deployer-generated kubeconfig to the exec test. The kubeconfig path is specific to the artifacts dir specified in prow and specific to the GCE deployer.

Tested using mkpj and mkpod on a local Kind cluster. Used a custom service account and targeted a specific project. Therefore, this did not test working with Boskos but Boskos interaction has been proven in other tests.

### DIFF
--- a/config/jobs/kubernetes/cloud-provider-gcp/cloud-provider-gcp-presubmits.yaml
+++ b/config/jobs/kubernetes/cloud-provider-gcp/cloud-provider-gcp-presubmits.yaml
@@ -55,4 +55,4 @@ presubmits:
           cd "${REPO_ROOT}/../kubetest2" || exit 1;
           make install;
           make install-deployer-gce;
-          kubetest2 gce -v 2 --repo-root $REPO_ROOT --build --up --down --test=exec -- kubectl get all --all-namespaces;
+          kubetest2 gce -v 2 --repo-root $REPO_ROOT --build --up --down --test=exec -- kubectl get all --all-namespaces --kubeconfig="${ARTIFACTS}/kubetest2-kubeconfig";
--- a/config/jobs/kubernetes/cloud-provider-gcp/cloud-provider-gcp-presubmits.yaml
+++ b/config/jobs/kubernetes/cloud-provider-gcp/cloud-provider-gcp-presubmits.yaml
@@ -30,15 +30,29 @@ presubmits:
     path_alias: k8s.io/cloud-provider-gcp
     labels:
       preset-service-account: "true"
-      preset-k8s-ssh: "true"
     annotations:
       testgrid-dashboards: provider-gcp-presubmits
-      description: End to end cluster creation test using cluster/kube-up based on kubernetes/cloud-provider-gcp.
+      description: End to end cluster creation test using kubetest2 based on kubernetes/cloud-provider-gcp.
       testgrid-num-columns-recent: '30'
+    extra_refs:
+    - org: kubernetes-sigs
+      repo: kubetest2
+      base_ref: master
+      path_alias: k8s.io/kubetest2
     spec:
       containers:
       - image: gcr.io/k8s-testimages/kubekins-e2e:v20200720-4766100-master
         command:
         - runner.sh
         args:
-        - test/e2e.sh
+        - "/bin/bash"
+        - "-c"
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes/cloud-provider-gcp/cloud-provider-gcp-presubmits.yaml
```diff
@@ -30,15 +30,29 @@ presubmits:
...
+        - "-c"
+        - set -o errexit;
+          set -o nounset;
+          set -o pipefail;
+          set -o xtrace;
+          REPO_ROOT=$(git rev-parse --show-toplevel);
+          cd "${REPO_ROOT}/../kubetest2" || exit 1;
+          make install;
+          make install-deployer-gce;
+          kubetest2 gce -v 2 --repo-root $REPO_ROOT --build --up --down;
```
amwat, line 58: would recommend atleast using the exec tester to run basic validation like
https://github.com/kubernetes/cloud-provider-gcp/blob/master/test/e2e.sh#L24


so something like
`kubetest2 gce -v 2 --repo-root $REPO_ROOT --build --up --down --test=exec -- kubectl get all --all-namespaces`

michaelmdresser: Added. When tested outside of Kind, this works fine. When tested in Kind in the Prow job, it fails. When investigating, I noticed that `kubectl cluster-info dump` also fails. The error reporting around the exec tester isn't very verbose, but the error during `cluster-info dump` was `The connection to the server localhost:8080 was refused - did you specify the right host or port?` which suggests that its just a messed up config. Given that the `cluster-info` step does not fail in other test runs (see https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_kubetest2/5/gce-build-up-down/1285669256284868611) I am inclined to believe this is an odd interaction with Kind and not an actual bug.

michaelmdresser: Working on getting a test running in GKE to validate.

michaelmdresser: Had to manually specify a kubeconfig path for the test command, see the updated PR text for details. It was working locally before the change because the default kubeconfig on my machine points to a valid cluster but this is not the case in the Prow pod. Tested with a Prow job on a GKE cluster.

config/jobs/kubernetes/cloud-provider-gcp/cloud-provider-gcp-presubmits.yaml
```diff
@@ -30,15 +30,29 @@ presubmits:
...
+        - "-c"
+        - set -o errexit;
+          set -o nounset;
+          set -o pipefail;
+          set -o xtrace;
+          REPO_ROOT=$(git rev-parse --show-toplevel);
+          cd "${REPO_ROOT}/../kubetest2" || exit 1;
+          make install;
+          make install-deployer-gce;
+          kubetest2 gce -v 2 --repo-root $REPO_ROOT --build --up --down --test=exec -- kubectl get all --all-namespaces --kubeconfig=/logs/artifacts/kubetest2-kubeconfig;
```
amwat, line 58: nit: maybe $ARTIFACTS instead of /logs/artifacts

michaelmdresser: Done and tested.

amwat:
/lgtm

/cc @BenTheElder 
for approval

PR #18458, opened 2020-07-23
enable oncall updater to handle multiple rotations

BenTheElder:
/cc @spiffxp @amwat @MushuEE 

also adds @MushuEE to the mapping, and trims stale entries

### DIFF
--- a/experiment/slack-oncall-updater/main.go
+++ b/experiment/slack-oncall-updater/main.go
@@ -26,6 +26,34 @@ import (
 	"strings"
 )
 
+// githubToSlack maps (lowercase!) GitHub usernames to Slack user IDs.
+// Add yourself when you join the oncall rotation.
+// You can mostly easily find your user ID by visiting your Slack profile, clicking "...",
+// and clicking "Copy user ID".
+var githubToSlack = map[string]string{
+	"amwat":          "U9B1P2UGP",
+	"chases2":        "UJ9R0FWD6",
+	"cjwagner":       "U4QFZFMCM",
+	"clarketm":       "UKMR3JF29",
+	"bentheelder":    "U1P7T516X",
+	"fejta":          "U0E2KHQ13",
+	"katharine":      "UBTBNJ6GL",
+	"krzyzacy":       "U22Q65CTG",
+	"mirandachrist":  "UJR3XHHNF",
+	"michelle192837": "U3TRY5WV7",
+	"spiffxp":        "U09R2FL93",
+	"tony-yang":      "UK3MVSP3J",
+}
+
+// rotationToSlack group maps the rotation in go.k8s.io/oncall to the slack
+// group ID for the oncall alias group for this rotation
+var rotationToSlackGroup = map[string]string{
+	"test-infra":         "SGLF0GUQH",   // @test-infra-oncall
+	"google-build-admin": "S017N31TLNN", // @google-build-admin
+	// TODO: create this slack group?
+	// "scalability": "",
+}
+
 func getJSON(url string, v interface{}) error {
 	resp, err := http.Get(url)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


experiment/slack-oncall-updater/main.go
```diff
@@ -26,6 +26,31 @@ import (
...
+	"spiffxp":        "U09R2FL93",
+}
+
+// rotationToSlack group maps the rotation in go.k8s.io/oncall to the slack
+// group ID for the oncall alias group for this rotation
+var rotationToSlackGroup = map[string]string{
+	"test-infra":         "SGLF0GUQH",   // @test-infra-oncall
+	"google-build-admin": "S017N31TLNN", // @google-build-admin
+	// TODO: create this slack group?
+	// "scalability": "",
```
BenTheElder, line 51: FYI @wojtek-t, this tool enables go.k8s.io/oncall data to map to a slack group so instead of pinging random specific people the release team etc. can ping `@test-infra-oncall` => the current test-infra-oncall

You could enable this for scalability oncall by:
- have each member of the rotation add themselves to `githubToSlack` above, following the instructions above
- sending a PR like https://github.com/kubernetes/community/pull/4989 to add an external-managed group with some name to the slack config
- asking #slack-admins to manually create the group with this name
- `@mention` the group from the slack web client, alt/right click & inspect the name. Grab the `data-user-group-id` property value, add it to this mapping on this line.

experiment/slack-oncall-updater/main.go
```diff
@@ -26,6 +26,31 @@ import (
...
 
+// githubToSlack maps (lowercase!) GitHub usernames to Slack user IDs.
+// Add yourself when you join the oncall rotation.
+// You can mostly easily find your user ID by visiting your Slack profile, clicking "...",
+// and clicking "Copy user ID".
+var githubToSlack = map[string]string{
+	"amwat":          "U9B1P2UGP",
+	"chases2":        "UJ9R0FWD6",
+	"cjwagner":       "U4QFZFMCM",
+	"bentheelder":    "U1P7T516X",
```
MushuEE, line 37: Should we alphabetize on the off chance that this gets that large?

experiment/slack-oncall-updater/main.go
```diff
@@ -127,21 +122,42 @@ func main() {
...
-	userID, err := mapGitHubToSlack(oncall)
-	if err != nil {
-		log.Fatalf("Failed to get Slack ID: %v", err)
+	log.Printf("Current oncallers: %s\n", oncallers)
+
+	failed := false
+	for rotation, user := range oncallers {
+		// skip rotations that have not yet added their slack ID
+		// this tool is not required
+		groupID, exists := rotationToSlackGroup[rotation]
```
amwat, line 135: FYI currently the json key doesn't map to the rotation.
i.e. `test-infra` is named `testinfra` in the json blob.

we will also need to change that export alias.

BenTheElder: we don't, I just have the wrong key in the map above then.

amwat: that works as well

experiment/slack-oncall-updater/main.go
BenTheElder: done.

experiment/slack-oncall-updater/main.go
BenTheElder: should be fixed

BenTheElder:
fixed alphabetization and corrected the test-infra => testinfra json key

BenTheElder:
for reference the current json data:
`{"Oncall":{"google-build-admins":"bentheelder","scalability":"tosi3k","testinfra":"cjwagner"}}`

experiment/slack-oncall-updater/main.go
```diff
@@ -26,6 +26,31 @@ import (
...
+	"michelle192837": "U3TRY5WV7",
+	"mushuee":        "U013TPFJWC8",
+	"spiffxp":        "U09R2FL93",
+}
+
+// rotationToSlack group maps the rotation in go.k8s.io/oncall to the slack
+// group ID for the oncall alias group for this rotation
+var rotationToSlackGroup = map[string]string{
+	"testinfra":          "SGLF0GUQH",   // @test-infra-oncall
+	"google-build-admin": "S017N31TLNN", // @google-build-admin
```
MushuEE, line 49: This should also be plural `"google-build-admins":"bentheelder"`

BenTheElder: cl/322878818 should take effect at 4:27

BenTheElder: it didn't take effect yet, so I added a second entry for both.

BenTheElder:
Noting that this worked as intended so far.

PR #18460, opened 2020-07-24
Tabularize the oncall page

amwat:
Also add a postsubmit to upload the changes whenever it changes.

/cc @BenTheElder @spiffxp

### DIFF
--- a/config/jobs/kubernetes/test-infra/test-infra-trusted.yaml
+++ b/config/jobs/kubernetes/test-infra/test-infra-trusted.yaml
@@ -575,6 +575,32 @@ postsubmits:
         - --project=k8s-testimages
         - --build-dir=.
         - gencred/
+  - name: post-test-infra-upload-oncall
+    cluster: test-infra-trusted
+    branches:
+    - master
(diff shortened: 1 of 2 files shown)


config/jobs/kubernetes/test-infra/test-infra-trusted.yaml
```diff
@@ -575,6 +575,32 @@ postsubmits:
...
         - --build-dir=.
         - gencred/
+  - name: post-test-infra-upload-oncall
+    cluster: test-infra-trusted
+    branches:
+    - master
+    run_if_changed: '^maintenance/oncall.html$'
+    decorate: true
+    spec:
+      serviceAccountName: pusher
```
amwat, line 585: not sure if this is the right one

BenTheElder: this is also not going to upload it I think because it's in the same PR

BenTheElder: we'll need to do a follow-up change to the file. I plan to at some point anyhow

PR #18461, opened 2020-07-24
add more info to oncall page

BenTheElder:
- clarify scope of rotations
- add more fields to each including the slack alias, timezone, description
- remove release managers (it's confusingly unrelated / not an oncall rotation, seems to be a legacy thing going back years)
- minor nits / cleanup / CSS

it's still not the prettiest page, but it's a bit more informative.

/cc @amwat @spiffxp @MushuEE 
Screen Shot 2020-07-24 at 12 29 25 AM

### DIFF
--- a/maintenance/oncall.html
+++ b/maintenance/oncall.html
@@ -2,6 +2,25 @@
 <!-- To "deploy": gsutil cp -Z oncall.html gs://test-infra-oncall/ -->
 <html lang="en">
 <script>
+  var rotationsMetadata = {
+    "google-build-admin": {
+      "prettyName": "Google Build Admin",
+      "slack": "@google-build-admin",
+      "timezone": '<a href="https://time.is/PT">PT</a>',
+      "description": 'Debian / RPM package build & publishing for releases',
+    },
+    "testinfra": {
+      "prettyName": "Test Infra",
+      "slack": "@test-infra-oncall",
+      "timezone": '<a href="https://time.is/PT">PT</a>',
+      "description": "Prow, TestGrid, and other Google hosted infrastructure"
+    },
+    "scalability": {
+      "prettyName": "Scalability",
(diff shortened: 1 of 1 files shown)


maintenance/oncall.html
```diff
@@ -2,6 +2,25 @@
...
+    "testinfra": {
+      "prettyName": "Test Infra",
+      "slack": "@test-infra-oncall",
+      "timezone": '<a href="https://time.is/PT">PT</a>',
+      "description": "Prow, TestGrid, and other Google hosted infrastructure"
+    },
+    "scalability": {
+      "prettyName": "Scalability",
+      "timezone": '<a href="https://time.is/CET">CET</a>',
+      // TODO: description
```
BenTheElder, line 21: @wojtek-t @tosi3k we could use a short description of the scalability rotation here
for slack see: https://github.com/kubernetes/test-infra/pull/18458#discussion_r459772745

BenTheElder:
also not sure this is 100% phrased correctly etc., but it should be a better starting point than the near complete lack of written-down details currently for these rotations.

maintenance/oncall.html
```diff
@@ -2,6 +2,25 @@
...
+    },
+    "testinfra": {
+      "prettyName": "Test Infra",
+      "slack": "@test-infra-oncall",
+      "timezone": '<a href="https://time.is/PT">PT</a>',
+      "description": "Prow, TestGrid, and other Google hosted infrastructure"
+    },
+    "scalability": {
+      "prettyName": "Scalability",
+      "timezone": '<a href="https://time.is/CET">CET</a>',
```
BenTheElder, line 20: this is a best-guess based on WAW / teams/ timezones. we can correct later if this is somehow wrong.

BenTheElder:
@amwat your auto push job works

PR #18470, opened 2020-07-24
fix github-token key name in secret

stpabhi:

This PR fixes the github-token key name in README to match with [starter-s3.yaml](https://github.com/kubernetes/test-infra/blob/master/config/prow/cluster/starter-s3.yaml#L369) and [starter-gcs.yaml](https://github.com/kubernetes/test-infra/blob/master/config/prow/cluster/starter-gcs.yaml#L248)


alvaroaleman:
Thanks!

Issue #18411, opened 2020-07-21
Replay velodrome clean from #18394 without breaking Metrics Test


MushuEE:
**What should be cleaned up or changed**:
#18394 was a revert of #18362 that cause failures in https://k8s-testgrid.appspot.com/sig-testing-misc#metrics-bigquery
The is work should be repeated but the 

**Provide any links for context**:

https://k8s-testgrid.appspot.com/sig-testing-misc#metrics-bigquery

https://storage.googleapis.com/kubernetes-jenkins/logs/metrics-bigquery/1285038502559354882/build-log.txt

bq is returning a non-0 exit code
/assign @MushuEE

PR #18472, opened 2020-07-24
Redo BigQuery influx purge without breaking metric job

MushuEE:
This was done previously as part of #18394 but had an issue with config update
That issue was fixed and tested as well as extra cleanup
/assign @spiffxp 

@spiffxp you were correct the LAST_DATA_TIME was not being replace and it was causing the query to break

### DIFF
--- a/metrics/bigquery.py
+++ b/metrics/bigquery.py
@@ -17,9 +17,7 @@
 """Runs bigquery metrics and uploads the result to GCS."""
 
 import argparse
-import calendar
 import glob
-import json
 import os
 import pipes
 import re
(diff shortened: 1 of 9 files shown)


MushuEE:
#18411

metrics/bigquery.py
```diff
@@ -28,10 +26,10 @@
 import time
 import traceback
 
-import influxdb
 import requests
 import ruamel.yaml as yaml
 
+BACKFILL_DAYS = 30
```
spiffxp, line 32: I personally think I'll want this to be a flag again at some point but this is good for now

PR #18473, opened 2020-07-24
Remove no longer needed netd jobs

MrHohn:
Per https://github.com/kubernetes/kubernetes/issues/93377#issuecomment-663136265.

Fixes kubernetes/kubernetes#93377

/cc @cadmuxe @aojea


BenTheElder:
FYI @liggitt @spiffxp

PR #18477, opened 2020-07-26
pass s3-credentials-file arg to deck

stpabhi:

This PR adds missing s3-credentials-file in deck args to starter-s3.yaml.

Error:
```error rendering spyglass page: error when resolving real path s3/prow-logs/logs/echo-test/1287312016553283584: blob (key "logs/echo-test/1287312016553283584.txt") (code=Unknown): MissingRegion: could not find region configuration```


mkumatag:
Thanks for the fix @stpabhi, I was exactly hitting this issue :P

PR #18478, opened 2020-07-26
Purge Kettle Monitor.py

MushuEE:
This has been unused for 2 years, influx is gone with the Velodrome removal as well. Cleaned up.


MushuEE:
Bad git state, starting over

PR #18481, opened 2020-07-27
presubmit (on demand) dual stack job

aojea:
This will allow to run an e2e test with a temporal kind version
that supports dual stack. It just need to add a comment in the PR

`/test pull-kubernetes-e2e-kind-conformance-dual`

Adding the job will help to review PRs related to dual stack and
avoid regressions.

### DIFF
--- a/config/jobs/kubernetes/sig-network/sig-network-kind.yaml
+++ b/config/jobs/kubernetes/sig-network/sig-network-kind.yaml
@@ -1,3 +1,57 @@
+presubmits:
+  kubernetes/kubernetes:
+  - name: pull-kubernetes-e2e-kind-dual-canary
+    optional: true
+    always_run: false
+    skip_report: false
+    decorate: true
+    labels:
+      preset-bazel-scratch-dir: "true"
+      preset-bazel-remote-cache-enabled: "true"
+      preset-dind-enabled: "true"
+      preset-kind-volume-mounts: "true"
+    decoration_config:
+      timeout: 60m
+      grace_period: 15m
+    path_alias: k8s.io/kubernetes
+    spec:
+      containers:
+      - image: gcr.io/k8s-testimages/krte:v20200720-4766100-master
+        command:
+        - wrapper.sh
+        - bash
+        - -c
+        - curl -sSL https://github.com/aojea/kind/releases/download/dualstack/linux-amd64.tgz | tar xvfz - -C "${PATH%%:*}/" && e2e-k8s.sh
+        env:
+        - name: FOCUS
+          value: \[Conformance\]|DualStack
+        - name: PARALLEL
+          value: "true"
+        - name: BUILD_TYPE
+          value: bazel
+        # enable IPV6 in bootstrap image
+        - name: "DOCKER_IN_DOCKER_IPV6_ENABLED"
+          value: "true"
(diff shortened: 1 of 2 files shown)


config/jobs/kubernetes/sig-network/sig-network-kind.yaml
```diff
@@ -1,3 +1,56 @@
...
+      grace_period: 15m
+    path_alias: k8s.io/kubernetes
+    spec:
+      containers:
+      - image: gcr.io/k8s-testimages/krte:v20200720-4766100-master
+        command:
+        - wrapper.sh
+        - bash
+        - -c
+        - curl -sSL https://github.com/aojea/kind/releases/download/dualstack/linux-amd64.tgz | tar xvfz - -C "${PATH%%:*}/" && e2e-k8s.sh
```
aojea, line 24: this was to live in my fork until it is more stable
cc: @BenTheElder

MrHohn:
> ERROR: /home/prow/go/src/github.com/kubernetes/test-infra/config/tests/testgrids/BUILD.bazel:25:1: Executing genrule //config/tests/testgrids:testgrid_config failed (Exit 1)
2020/07/27 09:59:38 FAIL: could not write config: 1 error occurred:
	* configuration error for (TestGroup) pull-kubernetes-e2e-kind-conformance-dual: Each Test Group must be referenced by at least 1 Dashboard Tab.

From the bazel failure, looks like test group needs to be included in a dashboard.

BenTheElder:
it should be in presubmits-kubernetes-non-blocking dashboard. though I don't think this is useful, that's a separate discussion from this PR

config/jobs/kubernetes/sig-network/sig-network-kind.yaml
BenTheElder: ACK, we do have a bit of a chicken and egg problem here, I suppose this is reasonable in the immediate term.
We've had similar things in CI before.

aojea: we are blind against a PR with 73 files changes and +9,512 −3,250 :shrug:

BenTheElder: long term we don't keep things like this, but notably early windows support required a similar fork for kube-up, temporarily.

config/jobs/kubernetes/sig-network/sig-network-kind.yaml
```diff
@@ -1,3 +1,60 @@
+presubmits:
+  kubernetes/kubernetes:
+  - name: pull-kubernetes-e2e-kind-conformance-dual
```
BenTheElder, line 3: maybe drop conformance from the name to avoid being misleading, the subset of tests is related to dualstack.

config/jobs/kubernetes/sig-network/sig-network-kind.yaml
aojea: this is temporary, of course

aojea:
@MrHohn do you mind lgtmed again :sweat_smile:

MrHohn:
(Sorry that I missed it.)
/lgtm

Issue #18424, opened 2020-07-22
General build caching support (via pod utilities)


saschagrunert:
_I know that there we're some similar discussions before, so I'd like to bring-up the topic again to solve the question about **how** to achieve the goal._

For example, caching go modules in a project-specific GCS Bucket would speed-up the build process if no modules have changed between two separate runs. The overall approach could be similar to the solution [CircleCI provides for caching dependencies](https://circleci.com/docs/2.0/caching):

1. Define a set of files/directories which should be cached from the build system, for example the `$GOCACHE`
1. Define a set of files/directories in the repository which will be hashed to identify the target cache
    - A name for the cache could be inherited by the job name and the hash string
1. Save the cache (if not already present) in a configurable Google Cloud Bucket location
    - The bucket retention policy will also define the cache expiry
1. Load the cache right after cloning if existing

My implementation approach would be to directly add this functionality to prow via the pod utilities (in a similar fashion like `clonerefs`). The `decoration_config` can be used to specify the caches as well as their destination bucket. WDYT? 

/cc @kubernetes/release-engineering

dims:
@saschagrunert i am sure you have run into this already if not, please see some prior discussion/poc:
- https://github.com/kubernetes/kubernetes/issues/88553
- https://github.com/kubernetes/test-infra/pull/16623

chizhg:
Not sure if it could be another option or not - Istio recently did a bunch of work on adding caching support via Prow job configurations by mounting shared volumes, and it seemed to be pretty effective, see https://github.com/istio/test-infra/pulls?q=is%3Apr+cache

Probably it can be achieved if Prow can supports configuring volumes with some dirs/files cached by default, and then inject the volume into the `test` container spec?

BenTheElder:
> Probably it can be achieved if Prow can support configuring volumes with some dirs/files cached by default, and then inject the volume into the test container?

Prow could get pretty far by mounting a persistent volume of some type to $XDG_CACHE_HOME (and setting this), but the devil is in the details.

In Kubernetes there's substantial divergence between multiple actively developed branches, so you want something aware of this for example. (I.E. don't just mount a generic go cache volume to all jobs).

Uploading a cache archive to GCS in postsubmit and downloading it in presubmit is relatively straightforward to do without any prow changes (https://github.com/kubernetes/test-infra/pull/16623), as another alternative to what istio did.

For Kubernetes it's somewhat less straightforward because of the complex build system / additional containerization of some steps.

[...]

PR #18483, opened 2020-07-27
RFC: Add cache support to prow

saschagrunert:
This commits adds basic caching support to prow by the usage of the
already available Object Storage bucket credentials.

Caching works only if job decorations are enabled, because it will hook
into the already existing `clonerefs` init container as well as the
`sidecar`. This means that we try to copy the cache from the Object
Storage bucket **after cloning the sources** and try to upload the cache
**after the main jobs process has finished successfully**.

We use `emptyDir` volumes to share the data during the execution, which
means that we create a new Volume for each `Cache` entry and mount them
into the `clonerefs`, `sidecar` and possible job containers.

A single cache entry can be defined in the `decoration_config` via:

```yaml
decoration_config:
  caches:
  - path: /tmp/gopath
    keys:
    - go.sum
    - version-marker.txt
```

One caveat right now is that the `path` has to be unique because it will
be used as normalized reference for the volume and mounts. Another point
worth mentioning is that the `path` has to be absolute to be mount-able.

All `Cache.Keys` will be translated into a uniquely identifiable
tarball (see `cache.TarBallName()`), which stores the cache data and
therefore indirectly invalidates the cache if any of the `keys` change.

Things still do be done:

- [ ] Integration testing
- [ ] Downloading caches via a new GCS download helper
- [ ] Compressing and extracting the cache tarball

/cc @kubernetes/release-engineering 
Closes https://github.com/kubernetes/test-infra/issues/18424


prow/cache/types/types.go
```diff
@@ -0,0 +1,104 @@
...
+	corev1 "k8s.io/api/core/v1"
+)
+
+// Caches is the basic abstraction over multiple Cache entries
+type Caches []*Cache
+
+// Validate verifies if every Cache contains a name, keys and a path.
+func (c Caches) Validate() error {
+	for _, cache := range c {
+		if cache.Name == "" {
```
MushuEE, line 35: Do we need enforcement that the Cache Name is unique?

saschagrunert: Yes, I think I can add it by iterating over all caches and checking for uniqueness. :+1:

MushuEE: That or Caches can be a struct with both `[]*Cache` and a `CacheSet := make(map[string]bool)` to have the quick check. But would force a change in all `Caches`

alvaroaleman:
@saschagrunert is there any kind of design doc around this? How is the object storage to local path mapping done?

It doesn't need to be super exhaustive but I would like to see a document outlining the overall idea end to end before we merge code.

saschagrunert:
> @saschagrunert is there any kind of design doc around this? How is the object storage to local path mapping done?
> 
> It doesn't need to be super exhaustive but I would like to see a document outlining the overall idea end to end before we merge code.

Yep I agree, I’ll add a documentation to `cache/README.md` and link it to the right places if that fit our needs. 👍

saschagrunert:
Two questions:

- Would it be better to implement the download functionality in a separate PR?
- Does uploading also support directories? Or is it just about files?

alvaroaleman:
@saschagrunert lets please not implement anything until we have agreed on a design. I see no such agreement in https://github.com/kubernetes/test-infra/issues/18424

prow/caching.md
```diff
@@ -0,0 +1,50 @@
+# Caching dependencies in Prow
```
howardjohn, line 1: I am curious about the choice to use GCS as the cache store rather than local disk. There are definitely some trade offs, seems disk is faster (not sure how much), simpler, but has potential size constraints (for Kubernetes scale - no where near a problem for Istio yet), and higher cache miss (new node is empty).

alvaroaleman: @howardjohn that assumes thread safety of the cache, is this the case for golang? I think it would be a bit safer to generally not assume that

prow/caching.md
```diff
@@ -0,0 +1,50 @@
...
+because the cache should be available between multiple Prow runs and even
+between different Prow jobs.
+
+A single cache entry can be defined in the `decoration_config` via:
+
+```yaml
+decoration_config:
+  caches:
+    - name: my-cache
+      path: /gocache
```
alvaroaleman, line 31: So `path` is both the target (in case of download) and the source (in case of upload)?

prow/caching.md
```diff
@@ -0,0 +1,50 @@
...
+      keys:
+        - go.mod
+        - go.sum
```
alvaroaleman, line 34: I think it should also be possible to out-out of the cache uploading. The keys are a very nice idea to allow invalidating the cache based on dependencies but they don't help with invalidating based on your own code.

More practically speaking, I would for example assume that presubmits should not upload the cache but instead a postsubmit should do that. Otherwise presubmits will race with each other for writing the cache, if they don't happen to change dependencies.

prow/caching.md
alvaroaleman: Could we get rid of the `name` attribute by using a sanitzed version of the `path` as name?

prow/caching.md
```diff
@@ -0,0 +1,50 @@
+# Caching dependencies in Prow
+
+Caching dependencies in prow works only when the [Pod Utilities][0] are enabled. A
+valid GCS Bucket has to be configured via `gcs_configuration` either in the Prow
```
alvaroaleman, line 4: Please do a `s/GCS/Object Storage/g`, we also support S3 :)

prow/caching.md
```diff
@@ -0,0 +1,50 @@
...
+Please ensure that the `name` has to be unique because it will be used as
+reference for the internally used volumes and mounts. Another point worth
+mentioning is that the `path` has to be absolute to be mount-able.
+
+The `keys` contain repository-relative paths, where a SHA256 is calculated for
+each of the specified files. The complete cache path on the remote bucket will
+then be assembled like this:
+
+```
+/prow/cache/{{ name }}-{{ sha256 keys[0] }}-{{ sha256 keys[1] }}…
```
alvaroaleman, line 46: I think it would make sense to upload this is a `tar.gz` rather than individually. My local gocache which I cleared yesterday has for example more then 30k files already, uploading all of this individually is very inefficient

prow/caching.md
```diff
@@ -0,0 +1,50 @@
...
+```
+
+This means that we indirectly invalidate the cache if its name or any of the
```
alvaroaleman, line 49: Conceptionally what would be great is to be able to delete all caches older than $some_timeframe. I guess this is possible by using metadata from the object storage?

prow/caching.md
```diff
@@ -0,0 +1,50 @@
...
+[1]: /prow/cmd/plank/README.md
+
+Saving and restoring the cache will be done in the `clonerefs` and `sidecar` pod
+utility containers, where:
+
+- `clonerefs` tries to restore the cache after the clone has been done. Multiple
+  `keys` can be specified to reference files inside the repository. Those files
+  will be hashed and serve as unique identifier for the cache.
+
+- `sidecar` tries to save the cache after the main test container has been
```
alvaroaleman, line 17: There is currently ongoing work to have more than one test container, cc @anthonyrtong . I guess this would mean the cache needs a `containerName` to associate it correctly?

prow/caching.md
```diff
@@ -0,0 +1,50 @@
...
+valid GCS Bucket has to be configured via `gcs_configuration` either in the Prow
+jobs `decoration_config` or in [Planks][1] `default_decoration_configs`.
+
+[0]: /prow/pod-utilities.md
+[1]: /prow/cmd/plank/README.md
+
+Saving and restoring the cache will be done in the `clonerefs` and `sidecar` pod
+utility containers, where:
+
+- `clonerefs` tries to restore the cache after the clone has been done. Multiple
```
alvaroaleman, line 13: Please clarify what happens if there is no cache and what happens if there was an error accessing the cache

prow/caching.md
howardjohn: I am fairly sure it is for go, but may not be for other cache usages

prow/caching.md
```diff
@@ -0,0 +1,50 @@
...
+
+A single cache entry can be defined in the `decoration_config` via:
+
+```yaml
+decoration_config:
+  caches:
+    - name: my-cache
+      path: /gocache
+      keys:
+        - go.mod
```
howardjohn, line 33: I don't think you want to invalidate on go.mod changing. If I change 1 dependency, I should only fetch that new one, not everything. Go is smart enough to handle a partially filled cache

alvaroaleman: FTR we had something built around Prow ourselves before that worked by just doing a `go test ./...` and uploading the resulting goacche in a postsubmit and then downloading that in all jobs on the same branch before doing any builds or tests. This worked very well

prow/caching.md
alvaroaleman: I've opened https://github.com/golang/go/issues/40461 to find out if golangs build and test cache is threadsafe

saschagrunert: To what are we referring when speaking about "local disk"? A `PersistentVolume`? I think this would work as well, but the benefit of using a GCS bucket I saw were:
- we have the chance to publicly download the cache if we want to debug the build
- we can enable some sort of lifecycle to cleanup old caches automatically

prow/caching.md
saschagrunert: Yes, I'll change the implementation to use the `Path` as indirect name. :+1:

prow/caching.md
saschagrunert: Done :+1:

prow/caching.md
saschagrunert: I agree that makes sense.

prow/caching.md
saschagrunert: Yeah I was thinking about some global cleanup either via some life-cycle setting in the object storage or a dedicated job taking care of removing caches older than x. WDYT?

prow/caching.md
saschagrunert: I thought about this as well and my idea was to share a cache across all containers (current implementation).

prow/caching.md
saschagrunert: Done :+1:

prow/caching.md
saschagrunert: Yeah that example is misleading, I'll change it to still outline that we can cache based on multiple keys.

prow/caching.md
saschagrunert: Sounds good, added a new `download_only` option to the doc as well as the API fields.

prow/caching.md
```diff
@@ -0,0 +1,58 @@
...
+
+If `download_only` is true, then it will not upload the cache if the test
+process succeeded and no cache is already available.
+
+The `keys` contain repository-relative paths, where a SHA256 is calculated for
+each of the specified files. The complete cache path on the remote bucket will
+then be assembled like this:
+
+```
+/prow/cache/{{ normalized path }}-{{ sha256 keys[0] }}-{{ sha256 keys[1] }}.tar.gz
```
alvaroaleman, line 54: As pointed out before, a file in the repository is probably not very useful as cache key. What about including `org-repo-branch` and suggesting to download only in presubmits/periodics and upload in a postsubmit? In that case it would also make sense to remove the `download_only` attribute and instead use an `upload: true`, because its the less common

fejta:
Why go with GCS rather than an image?

My thought is that we would deconstruct clonerefs, which currently:
* Initializes all the git repositories
* Checks out the base branches and merges in any PRs.

We could move the initialization out to another container, which could then more intelligently setup the repos, for example:
* configuring a reference repository
* copying temporary files, etc.

The image for this container can be extended per repo/prow deployment to hold common information such as:
* the git history for repos
* go cache
* pre-built files, etc.

The container can then copy this from the image into the workspace.

The advantage of using an image to store this data over GCS is that we do not have to redownload everything every time -- so long as the image remains the same we'll only have to download it once per node.

alvaroaleman:
> We could move the initialization out to another container, which could then more intelligently setup the repos, for example:

So basically another init container that does whatever it wants and can declare `emptyDir` volumes who will be mounted on the same path as they were mounted on that init container?

fejta:
Something like that. But I actually don't see the need for any new emptyDir volumes?

We just need this new initcontainer to prepare the codeMount emptyDir volume such that clonerefs and the test containers can do their stuff faster.

Currently after `clonerefs` exits there will be a dir that has done a `git init`, `git fetch` and 0-n `git checkout/merge` commands for each Refs in the prow job to the inside the `codeMount` volume:

https://github.com/kubernetes/test-infra/blob/817c06d1f81f30e9ad1ad69d9cda16498502d432/prow/pod-utils/decorate/podspec.go#L427

We then inject this `codeMount` volume into all the test containers:

https://github.com/kubernetes/test-infra/blob/817c06d1f81f30e9ad1ad69d9cda16498502d432/prow/pod-utils/decorate/podspec.go#L718-L724

So this would involve moving the `git init` responsibility out of the clonerefs and into the whatever-cache initcontainer that runs immediately prior to the clonerefs container.

It would be responsible for:
* Running git init for each ref in the prow job
* Any additional stuff it wants to do to the codeMount volume
  - Speed up cloneref's git fetch call by prepopulating .git with files from GCS and/or somewhere in the image
  - Speed up `go mod download by copying stuff into the go module cache
  - Speed up `bazel build` and/or `go build` by copying prebuilt .o etc files.

alvaroaleman:
> Any additional stuff it wants to do to the codeMount volume

Yeah but that won't do it if you want to have something else cached, like the `GOCACHE` which seems to be the usecase here and for istio

fejta:
(how is that different from use case 2?)

alvaroaleman:
the fact that we might want to touch other locations than the git repo is the reason we might need additional emptyDir Volumes (What usecase 2 are you referencing?)

fejta:
The second bullet for "stuff it wants to do", aka speeding up go mod download.

We should not design around istio's implementation which is targeting the same problem. We should choose the right abstraction.

IMO Everything relevant to the job's workspace should be within the codeMount. I don't think we need to support an arbitrary number of volumes. If someone needs that then we should focus on allowing multiple containers -- they can use the downward API to do whatever they want on their random 50 emptydir volumes.

The go module cache already lives inside the codeMount, we explicitly put it here:

https://github.com/kubernetes/test-infra/blob/817c06d1f81f30e9ad1ad69d9cda16498502d432/prow/pod-utils/decorate/podspec.go#L609

howardjohn:
Also note for go specifically - go mod cache is configurable, and go build cache will be in go 1.15.

saschagrunert:
My intention was to take the least invasive path first, but I guess we have to refactor some of the current code parts before actually supporting caching. I think the proposed changes by @fejta would also imply documentation and test adaptions, so we're probably good writing a KEP around that topic.

PR #18487, opened 2020-07-27
Plank: Verify all containers finished before declaring a job a success

alvaroaleman:
This job https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_multus-cni/71/pull-ci-openshift-multus-cni-master-e2e-aws/1287051359769595904 finished and was declared a success when it wasn't because of https://github.com/kubernetes/kubernetes/issues/58711.

This PR adds an additional verification that all containers in the pod finished successfully before declaring a job a success and declares it an error otherwise, as this is a bug in the underlying kube cluster.

### DIFF
--- a/prow/plank/controller.go
+++ b/prow/plank/controller.go
@@ -350,10 +350,17 @@ func (c *Controller) syncPendingJob(pj prowapi.ProwJob, pm map[string]corev1.Pod
 			return client.Delete(c.ctx, &pod)
 
 		case corev1.PodSucceeded:
-			// Pod succeeded. Update ProwJob, and start next jobs.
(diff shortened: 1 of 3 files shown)


smarterclayton:
From a safety thing, anyone running prow before 1.18 may be vulnerable to this, and prow on an autoscaling cluster is very likely to get graceful deletions or other evictions triggered, thus increasing the chance that a failing / incomplete pod is reported as passed.  So I think this is an important safety check that is reasonable to add (prior to 1.18 we can't trust exitCode = 0 to mean "pod passed", now we can)

PR #18486, opened 2020-07-27
Setup image pushing prowjob

b01901143:
image pushing prowjob for sigs.k8s.io/k8s-gsm-tools

PR #18488, opened 2020-07-27
Fix typo in k8s-gsm-tools job

spiffxp:
Fixes https://github.com/kubernetes/test-infra/pull/18486/files#r461192139


b01901143:
Thanks for catching this

/lgtm

Issue #18159, opened 2020-07-02
Add presubmits in testgrid for repos/modules


ZhiFeng1993:
Below are all repos that don't have their presubmits in testgrid.
This issue is to track moving this down to just kubernetes-security/
```
var noPresubmitsInTestgridPrefixes = []string{
	"containerd/cri",
	"GoogleCloudPlatform/k8s-multicluster-ingress",
	"kubeflow/pipelines",
	"kubernetes-sigs/cluster-capacity",
	"kubernetes-sigs/gcp-compute-persistent-disk-csi-driver",
	"kubernetes-sigs/gcp-filestore-csi-driver",
	"kubernetes-sigs/kind",
	"kubernetes-sigs/kubebuilder-declarative-pattern",
	"kubernetes-sigs/scheduler-plugins",
	"kubernetes-sigs/service-catalog",
	"kubernetes-sigs/sig-storage-local-static-provisioner",
	"kubernetes-sigs/slack-infra",
	"kubernetes-sigs/testing_frameworks",
	"kubernetes/client-go",
	"kubernetes/cloud-provider-openstack",
	"kubernetes/dns",
	"kubernetes/enhancements",
	"kubernetes/ingress-gce",
	"kubernetes/kubeadm",
	"kubernetes/minikube",
	// This is the one entry that should be here
	"kubernetes-security/",
}
```
This used to be tracked in the code. Make it an independent issue for easier tracking.  
https://github.com/kubernetes/test-infra/blob/master/config/tests/testgrids/config_test.go

BenTheElder:
I'm not sure why we'd want presubmits in testgrid.
Presubmits are very noisy and the heatmap in prow gives better insight (e.g. PRs that don't build are clearly distinct by runtime)

BenTheElder:
cc @spiffxp

chases2:
I got tripped up on this as well. Why do we want to display these tests on TestGrid? They run at variable frequencies against code that may not actually merge or work.

spiffxp:
It's about ensuring that testgrid can be a consistent gateway to all jobs executed by prow.k8s.io, instead of sometimes having to go view prow's deck, or spyglass' job-history page.

Presubmits aren't guaranteed to pass reliably, but testgrid can still provide a useful visual signal if one set of tests starts failing consistently across presubmits

chases2:
We discussed this during today's sig-testing meeting. We didn't come to a consensus, but I do want to note what I heard:

[...]

PR #18493, opened 2020-07-28
Add basic presubmits for kubetest2

amwat:
/cc @BenTheElder


config/tests/testgrids/config_test.go
```diff
@@ -356,6 +356,7 @@ var noPresubmitsInTestgridPrefixes = []string{
 	"kubernetes-sigs/cluster-capacity",
 	"kubernetes-sigs/gcp-filestore-csi-driver",
 	"kubernetes-sigs/kind",
+	"kubernetes-sigs/kubetest2",
```
amwat, line 359: +1 to https://github.com/kubernetes/test-infra/issues/18159#issuecomment-663402819

having a testgrid dashboard for non e2e presubmits just adds noise.

BenTheElder: testgrid also can't really keep up with k/k where it might be useful, prow.k8s.io / deck is good enough for presubmits (along with go.k8s.io/triage)

config/jobs/kubernetes-sigs/kubetest2/kubetest2-presubmits.yaml
```diff
@@ -0,0 +1,92 @@
...
+      containers:
+      - image: gcr.io/k8s-testimages/krte:v20200726-f8d6253-master
+        command:
+        - wrapper.sh
+        args:
+        - make
+        - install-deployer-gce
+        securityContext:
+          privileged: true
+  - name: pull-kubetest2-build-gke
```
BenTheElder, line 63: is there a reason to not have a single build job?
go caching should save some compute if we combine these.

amwat: these will run only when the specific deployer code is touched. 
alternatively building all the deployers every time does add a non-trivial amount of time to each build.
unless we can add a smart build job that figures out the changed deployers, multiple job in parallel seems like the best approach (at the cost of having multiple jobs)?

BenTheElder: this is a bad idea, because the deployers also depend on code not in these paths, you will miss breaking changes from that direction.

a single build job will avoid re-building the same packages reapeatedly.
if it gets too slow we'll get persistent go caching

amwat: unified into a single job. let' give that a go and see if we really need a go cache.

PR #18496, opened 2020-07-28
Increase memory limit for 5k correctness test

jkaniuk:
Ref: https://github.com/kubernetes/kubernetes/issues/93506

/cc wojtek-t


BenTheElder:
https://github.com/kubernetes/test-infra/issues/18530

PR #18500, opened 2020-07-28
Kettle parsing fails if 'time' key is ''

MushuEE:
Quick fix, hit
```
Traceback (most recent call last):
  File "make_json.py", line 278, in make_rows
    yield rowid, row_for_build(path, started, finished, results)
  File "make_json.py", line 218, in row_for_build
    for test in parse_junit(result):
  File "make_json.py", line 139, in parse_junit
    time, failure_text, skipped = parse_result(child)
  File "make_json.py", line 129, in parse_result
    time = float(child_node.attrib.get('time', 0))
ValueError: could not convert string to float: ''
```
If time is empty str
/assign @spiffxp

### DIFF
--- a/kettle/make_json.py
+++ b/kettle/make_json.py
@@ -126,7 +126,7 @@ def parse_junit(xml):
     # isn't very interesting.
 
     def parse_result(child_node):
-        time = float(child_node.attrib.get('time', 0))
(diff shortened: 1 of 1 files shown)


BenTheElder:
thanks

PR #18501, opened 2020-07-28
Kettle repos was getting overwrote if metadata didn't exist

MushuEE:
Small bug, code was overwriting the `repos` attr after being set in started, so if `repos` was None from Metadata, it would be set back to None and not populated in BQ

### DIFF
--- a/kettle/make_json_test.py
+++ b/kettle/make_json_test.py
@@ -19,6 +19,8 @@ import json
 import time
 import unittest
 
+from parameterized import parameterized
+
 import make_json
 import model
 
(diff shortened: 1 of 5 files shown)


MushuEE:
> Can you add test coverage for this?

Done, used actual results for all four run types, @spiffxp if you could just validate that the output makes sense in the `expected` part of the test, that would be great. https://docs.google.com/document/d/1SeAvk6f5LIbzOM3b-tscscSOj9e2KiQQLU3zjVGSRxM/edit?usp=sharing

kettle/make_json_test.py
```diff
@@ -35,6 +37,257 @@ def test_buckets(self):
...
+            {
+                "k8s.io/kubernetes": "master:5a529",
+                "k8s.io/release": "master"
+            },
+            {
+                'job': 'pr:pr-logs',
+                'executor': '0790211c-cacb-11ea-a4b9-4a19d9b965b2',
+                'path': 'gs://kubernetes-jenkins/pr-logs/path',
+                'test': [{'name': 'Test1', 'failed': True}],
+                'passed': True,
```
spiffxp, line 176: I did a double-take here, but I think this is right... it's not like kettle interprets the test results, it just translates what it sees

PR #18512, opened 2020-07-28
parse cloneUri and clonePath from repoLink to support GHE

stpabhi:

This PR fixes repo clone issues in presubmit jobs when using GitHub Enterprise

### DIFF
--- a/prow/pod-utils/clone/clone.go
+++ b/prow/pod-utils/clone/clone.go
@@ -110,7 +110,12 @@ func PathForRefs(baseDir string, refs prowapi.Refs) string {
 	if refs.PathAlias != "" {
 		clonePath = refs.PathAlias
 	} else {
-		clonePath = fmt.Sprintf("github.com/%s/%s", refs.Org, refs.Repo)
+		repoUrl, err := url.Parse(refs.RepoLink)
+		if err != nil {
+			clonePath = fmt.Sprintf("github.com/%s/%s", refs.Org, refs.Repo)
+		} else {
+			clonePath = fmt.Sprintf("%s/%s/%s", repoUrl.Host, refs.Org, refs.Repo)
+		}
 	}
 	return path.Join(baseDir, "src", clonePath)
 }
@@ -127,7 +132,7 @@ func gitCtxForRefs(refs prowapi.Refs, baseDir string, env []string, oauthToken s
 	g := gitCtx{
 		cloneDir:      PathForRefs(baseDir, refs),
 		env:           env,
-		repositoryURI: fmt.Sprintf("https://github.com/%s/%s.git", refs.Org, refs.Repo),
+		repositoryURI: fmt.Sprintf("%s.git", refs.RepoLink),
 	}
 	if refs.CloneURI != "" {
 		g.repositoryURI = refs.CloneURI


prow/pod-utils/clone/clone.go
```diff
@@ -110,7 +110,12 @@ func PathForRefs(baseDir string, refs prowapi.Refs) string {
 	if refs.PathAlias != "" {
 		clonePath = refs.PathAlias
 	} else {
-		clonePath = fmt.Sprintf("github.com/%s/%s", refs.Org, refs.Repo)
+		repoUrl, err := url.Parse(refs.RepoLink)
```
alvaroaleman, line 113: cc @sbueringer 

I don't think we should ever do an `iff err != nil { assume user intended to use GHE }`.

Also, please add some testcases demonstrating what you want to achieve.

stpabhi: Can we confidently assume repoLink is a valid url? If so, we don't have to handle err. I will add test case as well. Thanks

alvaroaleman: My point being: Don't make assumptions based on `if err != nil`. If you want to make assumptions, test for those.

As for the concrete change, its not clear to me what this does that you can't do with `refs.PathAlias`

stpabhi: I agree we can achieve the same with ```refs.PathAlias``` but that would require all presubmit jobs to have ```refs.PathAlias``` and ```refs.CloneURI``` set other than ```github.com```. I'm wondering if we can achieve using the ```github.link_url``` or any other field

stpabhi: I could also add it here [pjutil](https://github.com/kubernetes/test-infra/blob/master/prow/pjutil/pjutil.go#L72) where repoLink can be parsed. If it's done here no assumptions are needed as it's a required field. Thoughts?

alvaroaleman: Would it make sense the other way round to default `PathAlias` based on the `CloneURI` (which would require a parsing that works with all possible git url types)?

stpabhi: CloneURI is also optional field and is not set [here](https://github.com/kubernetes/test-infra/blob/master/prow/pjutil/pjutil.go#L72).  Would it be more appropriate to set CloneURI and PathAlias when new NewPresubmit is triggered in ```pjutil```? Like this.

```
func createRefs(pr github.PullRequest, baseSHA string) prowapi.Refs {
...
repoLink := pr.Base.Repo.HTMLURL
...
repoUrl, _ := url.Parse(repoLink)
return prowapi.Refs{
  ...
  CloneURI:  fmt.Sprintf("%s.git", repoLink),
  PathAlias: fmt.Sprintf("%s/%s/%s", repoUrl.Host, org, repo),
  ...
}

This would guarantee both cloneURI and pathAlias are always set. Let me know.

**Comment by sbueringer** [comment]
@stpabhi Just to understand what you're trying to do. It's not about fixing something that doesn't work, it's just about making it easier to use Prow with Github enterprise right?

I'm just asking because I tried to "fix" it myself a while ago until I discovered I have to set `clone_uri` and `path_alias` on every job and then it works:
```
   clone_uri: https://<ghe-url>/<org>/<repo>.git
   path_alias: <ghe-url>/<org>/<repo>
```

stpabhi:
@sbueringer Yes, I’m trying to make it work with GitHub Enterprise. I realized that repoLink is not always set. My above suggestion to parse in pjutil work with GHE. `repoLink := pr.Base.Repo.HTMLURL` I believe HTMLURL is a [required](https://github.com/kubernetes/test-infra/blob/master/prow/github/types.go#L246) field.

sbueringer:
@stpabhi But you are aware that it already works with Github Enterprise when you explicitly configure `clone_uri` and `path_alias`?

stpabhi:
> @stpabhi But you are aware that it already works with Github Enterprise when you explicitly configure `clone_uri` and `path_alias` on the presubmit jobs?

Yes, I do. It’s just to avoid setting explicitly if possible.

PR #18514, opened 2020-07-28
Remove duplicate fast flag from kubernetes build

hasheddan:
The --fast flag is specified twice in kubernetes_build.py, causing
argparse to error with conflicts when passing arguments.

See https://github.com/kubernetes/test-infra/pull/18290/commits/751e5de21c0724fd607b1d0ebc96376035b9a744 for context.

Causing failure in `build-master-fast`: https://testgrid.k8s.io/sig-release-master-blocking#build-master-fast

/cc @justaugustus


justaugustus:
The job passed, but the context is stuck?
/test pull-test-infra-bazel

PR #18516, opened 2020-07-29
Make copying slices safer in berghelroach.go, make a small semantic fix

michaelkolber:
This is one in a series of fixes to ensure that slices are copied/assigned safely.

/cc bentheelder spiffxp


spiffxp:

PR #18521, opened 2020-07-29
testgrids/openshift: Update to latest job definitions

smarterclayton:
Include changes to split OKD jobs up from OCP jobs (which was
confusing) and a number of test renames since then.

### DIFF
--- a/config/testgrids/openshift/redhat-openshift-ocp-release-4.5-informing.yaml
+++ b/config/testgrids/openshift/redhat-openshift-ocp-release-4.5-informing.yaml
@@ -883,6 +883,33 @@ dashboards:
       - key: comment
         value: 'test: <test-name> failed, see job: <link>'
       url: https://bugzilla.redhat.com/enter_bug.cgi
+    name: release-openshift-origin-installer-e2e-aws-4.5
+    open_bug_template:
+      url: https://github.com/openshift/origin/issues/
+    open_test_template:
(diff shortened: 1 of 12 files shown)


config/testgrids/openshift/redhat-openshift-ocp-release-4.5-informing.yaml
```diff
@@ -1574,6 +1601,9 @@ test_groups:
 - days_of_results: 60
   gcs_prefix: origin-ci-test/logs/release-openshift-ocp-installer-e2e-vsphere-upi-serial-4.5
   name: release-openshift-ocp-installer-e2e-vsphere-upi-serial-4.5
+- days_of_results: 60
+  gcs_prefix: origin-ci-test/logs/release-openshift-origin-installer-e2e-aws-4.5
+  name: release-openshift-origin-installer-e2e-aws-4.5
```
bparees, line 1606: why are origin install jobs moving from "blocking" to "informing"?

do we need a separate dashboard for origin vs ocp, too?

smarterclayton: Because they are not referenced by the release controller.  Azure is.

No, CI and OCP are the same.  They are one product.  OKD is a separate product.  Bulk of testing of OKD comes from CI and OCP

PR #18524, opened 2020-07-29
Move remake verify job to periodic

liggitt:
The verify-remake job is expensive and tests extremely rarely-changed files (Makefile.generated_files and make-rules)

This switches the presubmit to optional and only runs when those paths are changed, since the periodic release-blocking verify job already covers this

xref https://github.com/kubernetes/kubernetes/issues/87807#issuecomment-662532855

### DIFF
--- a/config/jobs/kubernetes/sig-testing/files-remake.yaml
+++ b/config/jobs/kubernetes/sig-testing/files-remake.yaml
@@ -3,7 +3,9 @@ presubmits:
   - name: pull-kubernetes-files-remake
     path_alias: "k8s.io/kubernetes"
     decorate: true
-    always_run: true
+    optional: true
+    run_if_changed: 'Makefile.generated_files|make-rules'
+    always_run: false
     skip_report: false
     # branched per release (older go versions)
     skip_branches:
--- a/config/jobs/kubernetes/sig-testing/verify.yaml
+++ b/config/jobs/kubernetes/sig-testing/verify.yaml
@@ -25,7 +25,7 @@ presubmits:
         env:
         - name: EXCLUDE_TYPECHECK
           value: "y"
-        # this is now covered by pull-kubernetes-files-remake
+        # this is now covered by the optional pull-kubernetes-files-remake that runs when Makefile.generated_files / make-rules changes
         - name: EXCLUDE_FILES_REMAKE
           value: "y"
         - name: EXCLUDE_GODEP


config/jobs/kubernetes/sig-testing/files-remake.yaml
```diff
@@ -1,9 +1,47 @@
+periodics:
+- interval: 6h
+  kubernetes/kubernetes:
+  - cluster: k8s-infra-prow-build
```
spiffxp, line 4: you'll need to set resource limits/requests to run here

liggitt: hoisted with my own petard

liggitt: do you have links to the dashboard for current resource use, and the group you need to PR into to get access?

spiffxp: (need to pr this into a doc someplace)

group: https://github.com/kubernetes/k8s.io/blob/master/groups/groups.yaml#L681-L692

The only visibility we can provide right now is for the verify job as a whole (unclear which part is generated-files-remake) https://console.cloud.google.com/monitoring/metrics-explorer?project=k8s-infra-prow-build&timeDomain=6h&pageState=%7B%22xyChart%22:%7B%22dataSets%22:%5B%7B%22timeSeriesFilter%22:%7B%22filter%22:%22metric.type%3D%5C%22kubernetes.io/container/memory/used_bytes%5C%22+resource.type%3D%5C%22k8s_container%5C%22+metadata.user_labels.%5C%22prow.k8s.io/job%5C%22%3Dmonitoring.regex.full_match(%5C%22ci-kubernetes-verify.*%5C%22)%22,%22minAlignmentPeriod%22:%223600s%22,%22unitOverride%22:%22By%22,%22aggregations%22:%5B%7B%22perSeriesAligner%22:%22ALIGN_MEAN%22,%22crossSeriesReducer%22:%22REDUCE_SUM%22,%22groupByFields%22:%5B%22metadata.user_labels.%5C%22prow.k8s.io/job%5C%22%22%5D%7D,%7B%22crossSeriesReducer%22:%22REDUCE_NONE%22%7D%5D%7D,%22targetAxis%22:%22Y1%22,%22plotType%22:%22LINE%22%7D%5D,%22options%22:%7B%22mode%22:%22COLOR%22%7D,%22constantLines%22:%5B%5D,%22timeshiftDuration%22:%220s%22,%22y1Axis%22:%7B%22label%22:%22y1Axis%22,%22scale%22:%22LINEAR%22%7D%7D,%22isAutoRefresh%22:true,%22timeSelection%22:%7B%22timeRange%22:%221d%22%7D%7D

Would also probably be a good idea to exclude this out of the main verify job

spiffxp: you could guess high, and then look at limit utilization metrics to see how much is actually needed

config/jobs/kubernetes/sig-testing/files-remake.yaml
```diff
@@ -1,9 +1,47 @@
+periodics:
+- interval: 6h
+  kubernetes/kubernetes:
```
spiffxp, line 3: checkconfig is complaining about this, you'll want to do something more like https://github.com/kubernetes/test-infra/blob/a4f5d39d933a5ce10cf101515322ec19d308c0fd/config/jobs/kubernetes/sig-testing/integration.yaml#L31-L40

BenTheElder:
xref: https://github.com/kubernetes/kubernetes/issues/92937

BenTheElder:
also xref: #18530, one less job to deal with.

config/jobs/kubernetes/sig-testing/files-remake.yaml
```diff
@@ -3,7 +3,9 @@ presubmits:
   - name: pull-kubernetes-files-remake
     path_alias: "k8s.io/kubernetes"
     decorate: true
-    always_run: true
+    optional: true
```
BenTheElder, line 6: `run_if_changed: foo` + `optional: false` is well behaved (required to pass for merge if it runs, considered OK if it's skipped), I suspect that's the behavior we actually want?

PR #18529, opened 2020-07-29
Kettle image now pushed to k8s-testimages  registry

MushuEE:
It should be pointed to k8s-gubernator project registry, I also extended the docs to help anyone that hits an image pull issue

### DIFF
--- a/kettle/Makefile
+++ b/kettle/Makefile
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-IMG = gcr.io/k8s-testimages/kettle
+IMG = gcr.io/k8s-gubernator/kettle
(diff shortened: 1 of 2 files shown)


kettle/Makefile
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-IMG = gcr.io/k8s-testimages/kettle
+IMG = gcr.io/k8s-gubernator/kettle
```
spiffxp, line 15: I feel like It'd be preferable to keep it in k8s-testimages? It's where most (if not all?) of our other test-infra-built images land

MushuEE: Ok I will need to play around with the push because rn `make push` puts them in guber

PR #18531, opened 2020-07-29
Add python3, python3-distutils, and python3-pip to the bootstrap image

detiber:
Currently we install the python2 equivalents, but not the python3 equivalents, this improves the ability to use this image for tooling that no longer supports python2.

/assign @hasheddan


spiffxp:
This appears to have broken pull-kubernetes-verify https://github.com/kubernetes/kubernetes/issues/93623

dims:
all the python3 stuff should be back in kubekins image now.

PR #18532, opened 2020-07-29
image-builder: add jobs config for Azure presubmit

CecileRobertMichon:
This adds a new job config for https://github.com/kubernetes-sigs/image-builder.

For now only an optional presubmit is added in order to test https://github.com/kubernetes-sigs/image-builder/pull/313. In the future, the goal is to make the job and other builder tests run by default and be required on PRs.

/assign @codenrhoden 
/cc @timothysc @justinsb @luxas @moshloop @figo @akutz

### DIFF
--- a/config/jobs/kubernetes-sigs/image-builder/OWNERS
+++ b/config/jobs/kubernetes-sigs/image-builder/OWNERS
@@ -1,0 +1,11 @@
+# See the OWNERS docs at https://go.k8s.io/owners
+
+approvers:
+  - akutz
+  - CecileRobertMichon
+  - codenrhoden
+  - figo
+  - justinsb
+  - luxas
+  - moshloop
+  - timothysc
--- a/config/jobs/kubernetes-sigs/image-builder/image-builder-presubmits.yaml
+++ b/config/jobs/kubernetes-sigs/image-builder/image-builder-presubmits.yaml
@@ -1,0 +1,24 @@
+presubmits:
+  kubernetes-sigs/image-builder:
+  - name: pull-azure-all
+    labels:
+      preset-azure-cred: "true"
+    always_run: false
+    optional: true
+    decorate: true
+    decoration_config:
+      timeout: 1h
+    max_concurrency: 12
+    path_alias: sigs.k8s.io/image-builder
+    spec:
+      containers:
+      - image: gcr.io/k8s-testimages/kubekins-e2e:v20200728-3610e67-1.18
+        args:
+          - runner.sh
+          - "./images/capi/packer/azure/scripts/ci-azure-e2e.sh"
+        resources:
+          requests:
+            cpu: 1000m
+    annotations:
+      testgrid-dashboards: sig-cluster-lifecycle-image-builder
+      testgrid-tab-name: pr-azure-all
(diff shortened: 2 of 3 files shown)


config/jobs/kubernetes-sigs/image-builder/OWNERS
```diff
@@ -0,0 +1,11 @@
+# See the OWNERS docs at https://go.k8s.io/owners
+
+approvers:
```
CecileRobertMichon, line 3: list taken from https://github.com/kubernetes-sigs/image-builder/blob/master/OWNERS (alphabetized)

config/jobs/kubernetes-sigs/image-builder/image-builder-presubmits.yaml
```diff
@@ -0,0 +1,26 @@
+presubmits:
+  kubernetes-sigs/image-builder:
+  - name: pull-azure-ubuntu-1804
+    labels:
+      preset-azure-cred: "true"
+    always_run: false
+    optional: true
+    decorate: true
+    decoration_config:
+      timeout: 1h
```
CecileRobertMichon, line 10: 1h should be plenty but might adjust later especially if we start doing more stuff like run node conformance down the line

config/jobs/kubernetes-sigs/image-builder/image-builder-presubmits.yaml
```diff
@@ -0,0 +1,23 @@
+presubmits:
+  kubernetes-sigs/image-builder:
+  - name: pull-azure-all
+    labels:
+      preset-azure-cred: "true"
+    always_run: false
```
detiber, line 6: Assuming the intent is that this will get flipped to true after https://github.com/kubernetes-sigs/image-builder/pull/313 merges?

detiber: Do we need to worry about setting max_concurrency for this job, or is it relatively safe to kick these off any number of times concurrently?

config/jobs/kubernetes-sigs/image-builder/OWNERS
CecileRobertMichon: We should update them in https://github.com/kubernetes-sigs/image-builder/blob/master/OWNERS first since that's the source of truth. cc @codenrhoden

config/jobs/kubernetes-sigs/image-builder/image-builder-presubmits.yaml
CecileRobertMichon: > Assuming the intent is that this will get flipped to true after kubernetes-sigs/image-builder#313 merges?

correct, for now I want to use this to test kubernetes-sigs/image-builder#313 without running it on other PRs

> Do we need to worry about setting max_concurrency for this job, or is it relatively safe to kick these off any number of times concurrently?

the limit is the number of Azure resources it consumes but it's only 1 vm per job so relatively low cost. I can set a max concurrency just to be safe.

CecileRobertMichon: Added concurrency

CecileRobertMichon:
This is ready to go, hoping to get it merged so I can start testing kubernetes-sigs/image-builder#313 in prow

PR #18533, opened 2020-07-30
Use fast builds on release-master-blocking

spiffxp:
Fast (`linux/amd64`-only) builds used to land at `ci/latest`. Now they live at `ci/latest-fast`, and cross builds live at `ci/latest`. This means jobs that used to pick up the latest merged PR within 20-30m are now taking 90m-2h to do so.

Unless a job needs to use the cross build, it should switch to using `ci/latest-fast`.

Updates the follopwing release-master-blocking jobs:
- ci-kubernetes-e2e-gce-master-new-gci-kubectl-skew-serial
- ci-kubernetes-gce-conformance-latest
- ci-kubernetes-e2e-gci-gce
- ci-kubernetes-e2e-ubuntu-gce-containerd
- ci-kubernetes-e2e-gci-gce-alpha-features
- ci-kubernetes-e2e-gci-gce-reboot
- ci-kubernetes-e2e-gce-device-plugin-gpu
- ci-kubernetes-e2e-gci-gce-ingress
- ci-kubernetes-e2e-gci-gce-scalability

Updates the following release-master-informing jobs:
- ci-kubernetes-e2e-gce-scale-correctness
- ci-kubernetes-e2e-gce-scale-performance

ref: https://github.com/kubernetes/test-infra/pull/18290#discussion_r461858097

PR #18534, opened 2020-07-30
Enforce use of ci/latest-fast on in-development release

spiffxp:
Wanted to make sure I hadn't missed anything in https://github.com/kubernetes/test-infra/pull/18533


spiffxp:
@justaugustus I think if such a situation arises we can decide to update this test to check against a list of exceptions

justaugustus:
Fair point!
/lgtm
/hold cancel

PR #18535, opened 2020-07-30
Ensure fast builds match fast extraction strategy

hasheddan:
Currently ci/*-fast extraction strategies may incorrectly match to ci
mode instead of ciFast mode due to the regex being used. This ensures
that ci/*-fast extraction strategies will keep checking for ciFast if
they first match ci.

Fixes https://github.com/kubernetes/kubernetes/issues/93561

/cc @justaugustus @spiffxp 

### DIFF
--- a/kubetest/extract_k8s.go
+++ b/kubetest/extract_k8s.go
@@ -96,6 +96,10 @@ func (l *extractStrategies) Set(value string) error {
 		if mat == nil {
 			continue
 		}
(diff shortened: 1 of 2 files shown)


justaugustus:
/lgtm
/approve
/assign @wojtek-t @feiskyer 
/priority critical-urgent

(Assigning some non-US Pacific approvers to hopefully see if we can recover before they get in the office.)

Issue #18528, opened 2020-07-29
"Pod scheduling timeout." is displayed as a job failure without additional context by spyglass' metadata lens.


cjwagner:
**What happened**:
ProwJobs that error out due to "Pod scheduling timeout." are incorrectly described by the metadata lens:
```
Test started last Monday at 6:22 PM failed after 1m26s. (more info)
```
These failures are particularly opaque to users because no build log or pod info is available. The only way to determine that the job actually errored out due to a pod scheduling timeout instead of failing normally is to look at the ProwJob YAML/JSON and check the `status` section.

**What you expected to happen**:
The metadata lens should indicate the correct job state (`error` rather than `failure`) and it should include the status description ("Pod scheduling timeout.")

**How to reproduce it (as minimally and precisely as possible)**:
Create a ProwJob that can't be scheduled (e.g. requires more CPU than any node can provide) and try to run it.

**Please provide links to example occurrences, if any**:
https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-conformance-kind-ga-only/1288101820190691330

/help
/area prow prow/deck prow/spyglass

BenTheElder:
@hasheddan I think you fixed this in https://github.com/kubernetes/test-infra/pull/18537 ?

hasheddan:
@BenTheElder a `failure` is still reported instead of `error` even though the appropriate hint is now being displayed. I have a follow up for that and may also see if I can surface a little more info to troubleshoot why a pod scheduling failure occurred 👍

Issue #18530, opened 2020-07-29
Kubernetes CI Policy: critical jobs must be Guaranteed Pod QOS


BenTheElder:
Part of https://github.com/kubernetes/test-infra/issues/18551

This is a policy action Item out of [Policies to improve Kubernetes CI] discussed at SIG-Testing yesterday.

- **Why it’s necessary**:
  - We believe jobs are getting starved of the resources they require, because the scheduler is not able to effectively place them on a node with sufficient resources
  - We believe that jobs that declare resource requests have a better chance at being placed on a node with sufficient resources for them (at _scheduling_ time)
  - We believe that jobs that declare resource limits are going to succeed because they have accurately declared resources, not because they are getting lucky and bursting above their requested resources
- “Critical” here is release-blocking and merge-blocking for kubernetes/kubernetes
  - trying to avoid needing to tackle the long tail of O(hundreds) of job configurations that do not specify resource requests/limits today, and focus on the release/merge blocking jobs first
  - beyond Kubernetes release blocking / presubmit blocking tests figuring out priority fairly would be tricky. We've already started moving these specifically to CNCF owned GCP projects, so the tentative "priority" plan here is to just have dedicated cluster(s) for these workloads, versus general CI. If not for that, I think we should definitely consider it on those tasks at least.
- Guaranteed QOS means
  - Non-zero CPU & memory resource requests to avoid starvation (+1)
  - Matching CPU & memory resource limits to ensure successes aren’t just due to burst (+1)
  - For more info, see: https://kubernetes.io/docs/tasks/configure-pod-container/quality-service-pod/

Checklist of release-blocking jobs: (h/t @tpepper)
- [x] https://github.com/kubernetes/test-infra/pull/18471 - everything else (initial proof of concept)
- [x] https://github.com/kubernetes/test-infra/issues/18577 - ci-kubernetes-build
- [x] https://github.com/kubernetes/test-infra/issues/18578 - ci-kubernetes-build-fast
- [x] https://github.com/kubernetes/test-infra/issues/18579 - ci-kubernetes-e2e-gce-device-plugin-gpu
- [x] https://github.com/kubernetes/test-infra/issues/18580 - ci-kubernetes-e2e-gci-gce-ingress
- [x] https://github.com/kubernetes/test-infra/issues/18581 - periodic-kubernetes-bazel-build
- [x] https://github.com/kubernetes/test-infra/issues/18582 - periodic-kubernetes-bazel-test

[...]

PR #18537, opened 2020-07-30
Pass prowjob.json to metadata lens

hasheddan:
The metadata lens will pass hints from an errored prow job, but it
currently never gets the prowjob.json passed as an artifact so hints are
never surfaced.

From running locally, you can see the hint is surfaced for the same job in which it is not on prow.k8s.io (https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/periodic-kubernetes-bazel-build-master/1288775099830243328)

error-hint

Partially addresses https://github.com/kubernetes/test-infra/issues/18528 but I am going to work on surfacing a little more information and reporting `error` instead of `failed` as suggested in the issue.

/cc @cjwagner @cblecker 

### DIFF
--- a/config/prow/config.yaml
+++ b/config/prow/config.yaml
@@ -53,7 +53,7 @@ deck:
       required_files:
       - started.json|finished.json
       optional_files:
-      - podinfo.json
+      - podinfo.json|prowjob.json
     - lens:
         name: buildlog
       required_files:


dims:
looks legit :) https://cs.k8s.io/?q=optional_files&i=nope&files=&repos=

/approve
/lgtm

hasheddan:
https://github.com/kubernetes/test-infra/issues/18530

BenTheElder:
thank you!

PR #18539, opened 2020-07-30
Add presets to seccomp-operator jobs

saschagrunert:
The tests complain that the docker socket is not available so we add two
more presets to fix that issue.

/cc @rhafer @hasheddan @pjbgf 

/hold


saschagrunert:
I think this should be fine now,
/hold cancel

PR #18541, opened 2020-07-30
Fix seccomp-operator bucket name

saschagrunert:
The suffix should not be needed for the bucket.

Ref: https://prow.k8s.io/log?job=post-seccomp-operator-push-image&id=1288855630882279424
/assign @hasheddan


saschagrunert:
/assign @clarketm
May I ask for approval here? :innocent:

PR #18545, opened 2020-07-30
Raise monitoring.prow.k8s.io prometheus retention to 90d

spiffxp:
I don't have any insight into how much disk prometheus is using right now at 30d retention. Ultimately I'd like to raise to a year

/cc @cjwagner @BenTheElder

### DIFF
--- a/config/prow/cluster/monitoring/prow_prometheus.yaml
+++ b/config/prow/cluster/monitoring/prow_prometheus.yaml
@@ -16,7 +16,7 @@ spec:
         resources:
           requests:
             storage: 100Gi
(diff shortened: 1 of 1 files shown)


BenTheElder:
amusingly while looking for this I found we still have a 1gb tot-storage disk in k8s-prow

I think we're using 23.7G / 100GB PV (check `du -h /prometheus` in the pod)

spiffxp:
/cc @cjwagner 
since I mentioned this yesterday

spiffxp:
/cc @chases2 
FYI as test-infa-oncall

PR #18556, opened 2020-07-31
Add Pod QOS Guaranteed tests for critical jobs

spiffxp:
Logf instead of Errorf to start with, so we can get a list of jobs that
aren't yet passing

This isn't the most user-friendly way of displaying a report, but this is what I would consider the authoritative list of jobs that have yet to match the new policy

### DIFF
--- a/config/tests/jobs/jobs_test.go
+++ b/config/tests/jobs/jobs_test.go
@@ -28,6 +28,7 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"regexp"
 	"sort"
 	"strings"
 	"testing"
@@ -967,10 +968,7 @@ func allStaticJobs() []cfg.JobBase {
 	return jobs
 }
 
-// isPodQOSGuaranteed returns true if the PodSpec's containers have non-zero
-// resource limits that are equal to their resource requests
-func isPodQOSGuaranteed(spec *coreapi.PodSpec) bool {
-	isGuaranteed := true
+func verifyPodQOSGuaranteed(spec *coreapi.PodSpec) (errs []error) {
 	resourceNames := []coreapi.ResourceName{
 		coreapi.ResourceCPU,
 		coreapi.ResourceMemory,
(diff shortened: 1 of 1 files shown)


config/tests/jobs/jobs_test.go
```diff
@@ -979,12 +977,78 @@ func isPodQOSGuaranteed(spec *coreapi.PodSpec) bool {
...
+	if !ok {
+		return false
+	}
+	return re.MatchString(dashboards)
+}
+
+// TODO: s/Should/Must and s/Logf/Errorf when all jobs pass
+func TestKubernetesMergeBlockingJobsShouldHavePodQOSGuaranteed(t *testing.T) {
+	repo := "kubernetes/kubernetes"
+	jobs := c.AllStaticPresubmits([]string{repo})
```
MushuEE, line 1023: Is this not also captured in `allStaticJobs?`

config/tests/jobs/jobs_test.go
```diff
@@ -979,12 +977,78 @@ func isPodQOSGuaranteed(spec *coreapi.PodSpec) bool {
...
+		for _, err := range errs {
+			t.Logf("%v (%v): %v", job.Name, branches, err)
+		}
+	}
+}
+
+// TODO: s/Should/Must and s/Logf/Errorf when all jobs pass
+func TestKubernetesReleaseBlockingJobsShouldHavePodQOSGuaranteed(t *testing.T) {
+	for _, job := range allStaticJobs() {
+		// Only consider Pods that are merge-blocking
```
MushuEE, line 1043: `release-blocking`

config/tests/jobs/jobs_test.go
spiffxp: `JobBase` doesn't provide info about what repo that job is for if it's a presubmit or a postubmit, so if I use `allStaticJobs` I have no way of filtering for jobs that run against "kubernetes/kubernetes"

PR #17784, opened 2020-06-01
prow: hook: approvers plugin: configurable command help page

sbueringer:
xref: https://github.com/kubernetes/test-infra/issues/17774

PR #18571, opened 2020-07-31
Make PR Process link configurable

mikesplain:
Similar to https://github.com/kubernetes/test-infra/pull/17784, we'd like to configure the PR Process Link.  Thanks!

### DIFF
--- a/prow/plugins/config.go
+++ b/prow/plugins/config.go
@@ -274,6 +274,10 @@ type Approve struct {
 	// The default value is "https://go.k8s.io/bot-commands". The command help page is served by Deck
 	// and available under https://<deck-url>/command-help, e.g. "https://prow.k8s.io/command-help"
 	CommandHelpLink string `json:"commandHelpLink"`
+
+	// PrProcessLink is the link to the help page which explains the code review process.
+	// The default value is "https://git.k8s.io/community/contributors/guide/owners.md#the-code-review-process".
+	PrProcessLink string `json:"pr_process_link,omitempty"`
 }
 
 var (
@@ -853,6 +857,9 @@ func (c *Configuration) setDefaults() {
 	for i, approve := range c.Approve {
 		if approve.CommandHelpLink == "" {
 			c.Approve[i].CommandHelpLink = "https://go.k8s.io/bot-commands"
+		}
+		if approve.PrProcessLink == "" {
+			c.Approve[i].PrProcessLink = "https://git.k8s.io/community/contributors/guide/owners.md#the-code-review-process"
 		}
 	}
 	if c.Blunderbuss.ReviewerCount == nil {
--- a/prow/plugins/config_test.go
+++ b/prow/plugins/config_test.go
@@ -216,16 +216,22 @@ func TestSetApproveDefaults(t *testing.T) {
 
 		commandHelpLink         string
 		expectedCommandHelpLink string
+		prProcessLink           string
+		expectedPrProcessLink   string
 	}{
 		{
-			name:                    "set commandHelpLink default",
+			name:                    "set commandHelpLink and prProcessLink default",
 			commandHelpLink:         "",
+			prProcessLink:           "",
 			expectedCommandHelpLink: "https://go.k8s.io/bot-commands",
+			expectedPrProcessLink:   "https://git.k8s.io/community/contributors/guide/owners.md#the-code-review-process",
 		},
 		{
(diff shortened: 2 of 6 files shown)


prow/plugins/config.go
```diff
@@ -274,6 +274,10 @@ type Approve struct {
 	// The default value is "https://go.k8s.io/bot-commands". The command help page is served by Deck
 	// and available under https://<deck-url>/command-help, e.g. "https://prow.k8s.io/command-help"
 	CommandHelpLink string `json:"commandHelpLink"`
+
+	// PrProcessLink is the link to the help page which explains the code review process.
+	// The default value is "https://git.k8s.io/community/contributors/guide/owners.md#the-code-review-process".
+	PrProcessLink string `json:"prProcessLink"`
```
alvaroaleman, line 280: Can we please use camel case like for all other attributes and `omitempty`? I know I missed this in the review of `CommandHelpLink` but thats no reason to keep doing the wrong thing :)

mikesplain: Absolutely! Fair point, although I'm slightly confused and assume you're referencing the json attributes that all seem to use underscores instead of Camel Case? So something like this? Sorry for the follow up question!

```suggestion
	PrProcessLink string `json:"pr_process_link,omitempty"`
```

alvaroaleman: Yeah exactly

mikesplain: Cool will do thanks!

mikesplain:
Updated and gofmt'ed

prow/plugins/config_test.go
```diff
@@ -216,16 +216,22 @@ func TestSetApproveDefaults(t *testing.T) {
...
 			name:                    "set commandHelpLink default",
 			commandHelpLink:         "",
+			prProcessLink:           "",
 			expectedCommandHelpLink: "https://go.k8s.io/bot-commands",
+			expectedPrProcessLink:   "https://git.k8s.io/community/contributors/guide/owners.md#the-code-review-process",
 		},
 		{
 			name:                    "keep commandHelpLink value",
 			commandHelpLink:         "https://prow.k8s.io/command-help",
+			prProcessLink:           "https://github.com/kubernetes/community/blob/master/contributors/guide/owners.md#the-code-review-process",
```
alvaroaleman, line 232: mind changing the `prProcessLink` here? This test would pass even if the functionality of setting it was broken as the input value is the same as the default

prow/plugins/config_test.go
```diff
@@ -245,6 +252,10 @@ func TestSetApproveDefaults(t *testing.T) {
 		if c.Approve[0].CommandHelpLink != test.expectedCommandHelpLink {
 			t.Errorf("unexpected commandHelpLink: %s, expected: %s", c.Approve[0].CommandHelpLink, test.expectedCommandHelpLink)
 		}
+
+		if c.Approve[0].PrProcessLink != test.expectedPrProcessLink {
+			t.Errorf("unexpected commandHelpLink: %s, expected: %s", c.Approve[0].PrProcessLink, test.expectedPrProcessLink)
```
alvaroaleman, line 257: The error is misleading

prow/plugins/config_test.go
mikesplain: Sure, I figured the default being `https://git.k8s.io/community` was different enough from `https://github.com/kubernetes/community` but I made it more explicitly different than the default now.

prow/plugins/config_test.go
```diff
@@ -216,16 +216,22 @@ func TestSetApproveDefaults(t *testing.T) {
 
 		commandHelpLink         string
 		expectedCommandHelpLink string
+		prProcessLink           string
+		expectedPrProcessLink   string
 	}{
 		{
 			name:                    "set commandHelpLink default",
```
alvaroaleman, line 223: Please rename the two testcases to clarify that they test both commandHelpLink and PrProcessLink (or create new testcases for the latter)

prow/plugins/config_test.go
mikesplain: Doh.... thank you, how embarrassing.

mikesplain: Fixed, thanks

prow/plugins/config_test.go
mikesplain: Renamed, thanks!

PR #18573, opened 2020-07-31
image-builder: use the latest kubekins image

CecileRobertMichon:
https://github.com/kubernetes-sigs/image-builder/pull/313#issuecomment-667315046

/assign @detiber @codenrhoden


codenrhoden:
/lgtm

I thought the versions were updated automatically? Might be the `-1.18` vs `-master`...

CecileRobertMichon:
@codenrhoden they are, but I used 1.18 at first because that's what cluster api was using (because there were issues with go 1.14 in the -master image)

PR #18575, opened 2020-07-31
Fix wheel spinner to be centered

rothgar:
@BenTheElder This probably fixes the spinner to be centered.

I'm not familiar with the code though. How can I run this locally to verify?


BenTheElder:
I think you want [prow/cmd/deck/runlocal](https://github.com/kubernetes/test-infra/blob/master/prow/cmd/deck/runlocal) + view `/view/gs/kubernetes-jenkins/pr-logs/pull/93589/pull-kubernetes-e2e-kind/1289324492274798593` on the local instance (the data is public read so it should load).

BenTheElder:
so I got this running and the results were amusing, it seems to swing in a wide arc instead of about the center.

you'll need bazel at the right version, I use bazelisk in path as bazel https://github.com/bazelbuild/bazelisk

after runlocal http://localhost:8080/view/gs/kubernetes-jenkins/pr-logs/pull/93589/pull-kubernetes-e2e-kind/1289324492274798593 serves a sample page

rothgar:
I realized my hard coded px is relative to the page so once I get the local version up I'll play with the values more.

BenTheElder:
thanks!
compiling may take ... a bit.

rothgar:
So I think this is the easiest option. I forced the image to be square and then use 50% spin so it's relative to the resized height. Technically it's stretched but there's no more spin wobble.

BenTheElder:
looks good, thank you!
/lgtm
/approve

PR #18602, opened 2020-08-01
Revert verify jobs to last known good kubekins version (v20200730-e636ee2)

hasheddan:
verify-publishing-bot is failing in the verify jobs due to issue with
pyyaml import. This gets the jobs running again until that issue is
resolved.

/cc @liggitt @spiffxp 

/priority critical-urgent

### DIFF
--- a/config/jobs/kubernetes/sig-release/release-branch-jobs/1.16.yaml
+++ b/config/jobs/kubernetes/sig-release/release-branch-jobs/1.16.yaml
@@ -427,7 +427,7 @@ periodics:
         value: release-1.16
       - name: REPO_DIR
         value: /workspace/k8s.io/kubernetes
-      image: gcr.io/k8s-testimages/kubekins-e2e:v20200731-2efbce0-1.16
+      image: gcr.io/k8s-testimages/kubekins-e2e:v20200730-e636ee2-1.16
       imagePullPolicy: Always
       name: ""
       resources:
@@ -1202,7 +1202,7 @@ presubmits:
           value: release-1.16
         - name: REPO_DIR
           value: /workspace/k8s.io/kubernetes
-        image: gcr.io/k8s-testimages/kubekins-e2e:v20200731-2efbce0-1.16
+        image: gcr.io/k8s-testimages/kubekins-e2e:v20200730-e636ee2-1.16
         imagePullPolicy: Always
         name: ""
         resources:
--- a/config/jobs/kubernetes/sig-release/release-branch-jobs/1.17.yaml
+++ b/config/jobs/kubernetes/sig-release/release-branch-jobs/1.17.yaml
@@ -488,7 +488,7 @@ periodics:
         value: release-1.17
       - name: REPO_DIR
         value: /workspace/k8s.io/kubernetes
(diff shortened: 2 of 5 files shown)


liggitt:
Both python scripts passed with the change in https://github.com/kubernetes/kubernetes/pull/93627

It seems strange to have to pick CI script changes to release branches though. Making dramatic changes to release branch test images seems not great

config/jobs/kubernetes/sig-release/release-branch-jobs/1.16.yaml
```diff
@@ -427,7 +427,7 @@ periodics:
         value: release-1.16
       - name: REPO_DIR
         value: /workspace/k8s.io/kubernetes
-      image: gcr.io/k8s-testimages/kubekins-e2e:v20200731-2efbce0-1.16
+      image: gcr.io/k8s-testimages/kubekins-e2e:v20200730-e636ee2-1.16
```
spiffxp, line 430: I expect subsequent autobump PRs will try to change these back to whatever the latest built version is. If we still haven't found a fix-forward by then, we'll want to consider a manual tag

spiffxp:
I also think it's completely fair to revert https://github.com/kubernetes/test-infra/pull/18531

hasheddan:
@liggitt @spiffxp I could go either way. It seems like `python3` is desirable for the CAPI jobs, and it likely makes sense to move to `python3` at some point in other areas as well. However, I agree that picking the CI change onto release branches feels strange, but that seems like more of a product of using the same image building for every release branch. Has there ever been discussion of maintaining separate test images from the image building perspective (e.g. not just the 1.18 flavor of the new kubekins image, but literally maintaining a separate 1.18 kubekins image build)?

liggitt:
Reverting back to known good state (reverting #18531) seems like a good first step

Then considering how to reintroduce that in a non-disruptive way (making python3 available only as python3, but keeping `python` referring time python2, etc) is reasonable

hasheddan:
@liggitt just to be sure, `python` still is referring to python2 I believe. I think the issue is that there are two versions of `pyyaml` installed so when using `python` (i.e. python2) it is using the python3 source and having issues. Not entirely sure though (was looking at https://stackoverflow.com/questions/36055100/pyyaml-with-python-3-x)

liggitt:
> @liggitt just to be sure, `python` still is referring to python2 I believe. I think the issue is that there are two versions of `pyyaml` installed so when using `python` (i.e. python2) it is using the python3 source and having issues. Not entirely sure though (was looking at https://stackoverflow.com/questions/36055100/pyyaml-with-python-3-x)

That seems like a good thing to figure out in a canary job with master and release branches not in a broken state. I'd recommend a revert of https://github.com/kubernetes/test-infra/pull/18531 to unblock the critical path for release

hasheddan:
@liggitt will do 👍

liggitt:
(separately, I'd love to see removing the dependency on kubekins python in kubernetes verify scripts - looks like boilerplate, flags_underscore, and publishing-bot are the three verify scripts we use python for)

PR #18603, opened 2020-08-01
Revert "Add python3, python3-distutils, and python3-pip to the bootstrap image"

hasheddan:
This is causing https://github.com/kubernetes/kubernetes/issues/93623. Will reintroduce when it can be done safely.

See https://github.com/kubernetes/test-infra/pull/18602 and https://github.com/kubernetes/kubernetes/pull/93627

Reverts kubernetes/test-infra#18531

/cc @liggitt @spiffxp @detiber

PR #18605, opened 2020-08-01
Bump bootstrap base image in kubekins to v20200801-23e4b0e

hasheddan:
Bump bootstrap image to version with reverted python3 addition.

See https://github.com/kubernetes/test-infra/pull/18603

/cc @liggitt @spiffxp


hasheddan:
After `kubekins-e2e` image is build will open a PR to bump jobs :+1:

PR #18572, opened 2020-07-31
Fix ci/fast URL for setting k8s release

hasheddan:
`Currently CI jobs are being pointed to /ci/fast/*.txt but the correct
URL is /ci/*-fast.txt.`

URL currently being used: https://storage.googleapis.com/kubernetes-release-dev/ci/fast/latest.txt

Correct URL: https://storage.googleapis.com/kubernetes-release-dev/ci/latest-fast.txt

Example job run: https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-e2e-gci-gce-ingress/1289287120283766784

Fixes https://github.com/kubernetes/kubernetes/issues/93561

Signed-off-by: hasheddan <georgedanielmangum@gmail.com>

/cc @justaugustus @BenTheElder @spiffxp 
/priority critical-urgent

PR #18608, opened 2020-08-02
Supply get-kube.sh with correct fast build source URL

hasheddan:
Fast builds publish the lastest version marker to
`kubernetes-release-dev/ci/<version>-fast.txt` but the build source is
actually at
`kubernetes-release-dev/ci/fast/<version>/kubernetes.tar.gz`. This
updates extraction to use the correct URL for each purpose.

Example failed run: https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-e2e-gci-gce-ingress/1289724502317469699

Follow up to #18572 

Note: jobs will only fail intermittently right now (with low traffic on k/k) because eventually the regular CI build (i.e. not fast build) appears at the source URL currently being used. This will fix the issue after a `kubekins` image bump. I would also like to circle back and clean up some of the logic here / potentially have fast build latest version marker published to `kubernetes-release-dev/ci/fast/<version>.txt`

/cc @justaugustus @dims @spiffxp @BenTheElder

### DIFF
--- a/kubetest/extract_k8s.go
+++ b/kubetest/extract_k8s.go
@@ -344,15 +344,14 @@ var httpCat = func(url string) ([]byte, error) {
 	return release, nil
 }
 
-func setReleaseFromHTTP(prefix, suffix string, getSrc bool) error {
+func setReleaseFromHTTP(prefix, suffix string) (string, string, error) {
 	url := fmt.Sprintf("https://storage.googleapis.com/%s", prefix)
 	catURL := fmt.Sprintf("%s/%s.txt", url, suffix)
 	release, err := httpCat(catURL)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


justaugustus:
> I would also like to circle back and clean up some of the logic here / potentially have fast build latest version marker published to `kubernetes-release-dev/ci/fast/<version>.txt`

@hasheddan -- I actually purposely decided not to go that route.

Every extraction script I've seen expects the version marker to be located at `<bucket>/<folder>/<marker>.txt`, so I didn't want us to create a situation where the marker wouldn't be where it's "supposed to be".

As of https://github.com/kubernetes/test-infra/pull/18290, we've essentially accepted `latest` to mean, "most recent cross build". I think it might be confusing to have this marker name mean different things in different places.

PR #18610, opened 2020-08-02
Add resource requests and limits for pull-kubernetes-verify presubmits

### LINKED ISSUE #18597: Make critical jobs Guaranteed Pod QOS: pull-kubernetes-verify

This is part of https://github.com/kubernetes/test-infra/issues/18530

The following jobs should be Guaranteed Pod QOS, meaning they should have CPU and memory resource limits, and matching resource requests:
- pull-kubernetes-verify

These jobs run on (google.com only) k8s-prow-build, so @spiffxp has provided the following guess:
- suggest 7 cpu, 10Gi mem

General steps to follow:
- update the job definitions in [/config/jobs](/config/jobs) to have matching CPU and memory limits and requests
  - e.g. ([ci-kubernetes-e2e-gci-gce has a `resources:` field with matching entries](https://github.com/kubernetes/test-infra/blob/2eac54f721ed58479a3126e4f9ca5cbfcc73821b/config/jobs/kubernetes/sig-cloud-provider/gcp/gcp-gce.yaml#L357-L394))
- open a pull request, include a link to this issue and cc `@kubernetes/ci-signal` in the description
- keep an eye on the jobs for the next few days, and if they start failing more than usual, open followup pull requests to raise resources
  - can look at recent prow.k8s.io runs (e.g. https://prow.k8s.io/?job=ci-kubernetes-e2e-gci-gce)
  - can look at job history (e.g. https://prow.k8s.io/job-history/gs/kubernetes-jenkins/logs/ci-kubernetes-e2e-gci-gce)
  - can look at testgrid dashboards (e.g. https://testgrid.k8s.io/sig-release-master-blocking#gce-cos-master-default)
- after the job has remained stable for a few days, declare victory

/sig testing
/sig release
/area jobs
/area release-eng

hasheddan:
Sets limits and requests for  cpu to 7000m and memory to 10Gi as part of
guaranteed pod QOS effort for critical jobs.

Fixes #18597
xref #18530 

Went ahead and added these since I have already been monitoring `pull-kubernetes-verify` regularly as part of performance improvement efforts with @liggitt 

/cc @spiffxp @BenTheElder @tpepper

### DIFF
--- a/config/jobs/kubernetes/sig-testing/verify.yaml
+++ b/config/jobs/kubernetes/sig-testing/verify.yaml
@@ -38,8 +38,12 @@ presubmits:
         securityContext:
           privileged: true
         resources:
+          limits:
+            cpu: 7
+            memory: 10Gi
           requests:
-            cpu: 4
+            cpu: 7
+            memory: 10Gi
 periodics:
 - interval: 1h
   cluster: k8s-infra-prow-build
--- a/config/jobs/kubernetes/sig-release/release-branch-jobs/1.16.yaml
+++ b/config/jobs/kubernetes/sig-release/release-branch-jobs/1.16.yaml
@@ -1206,7 +1206,11 @@ presubmits:
         imagePullPolicy: Always
         name: ""
         resources:
+          limits:
+            cpu: "7"
+            memory: 10Gi
           requests:
-            cpu: "4"
+            cpu: "7"
+            memory: 10Gi
         securityContext:
           privileged: true
--- a/config/jobs/kubernetes/sig-release/release-branch-jobs/1.17.yaml
+++ b/config/jobs/kubernetes/sig-release/release-branch-jobs/1.17.yaml
@@ -1214,8 +1214,12 @@ presubmits:
         imagePullPolicy: Always
         name: ""
         resources:
+          limits:
+            cpu: "7"
+            memory: 10Gi
           requests:
-            cpu: "4"
+            cpu: "7"
+            memory: 10Gi
         securityContext:
           privileged: true
   - always_run: false
--- a/config/jobs/kubernetes/sig-release/release-branch-jobs/1.18.yaml
+++ b/config/jobs/kubernetes/sig-release/release-branch-jobs/1.18.yaml
@@ -1449,7 +1449,11 @@ presubmits:
         imagePullPolicy: Always
         name: ""
         resources:
+          limits:
+            cpu: "7"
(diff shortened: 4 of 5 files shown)


config/jobs/kubernetes/sig-testing/verify.yaml
```diff
@@ -38,8 +38,12 @@ presubmits:
         securityContext:
           privileged: true
         resources:
+          limits:
+            cpu: 7
```
liggitt, line 42: Is this the peak or the average? Not sure we want to reserve the peak CPU

hasheddan: I was going off of recommended amount from @spiffxp. I _think_ I don't have access to metrics on these jobs, but could be wrong

spiffxp: (really need to find a way to export this data)
Screen Shot 2020-08-02 at 10 15 27 AM

hasheddan: @liggitt @spiffxp ok to remove hold here?

spiffxp: I'm ok with it as long as you're around to watch and adjust as needed if the jobs starts failing

hasheddan: @spiffxp I will be :+1:

spiffxp:
Suspicious job runs:
- https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/directory/pull-kubernetes-verify/1290329991950635009
- https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/directory/pull-kubernetes-verify/1290404087937896449

They list all passing tests, but fewer than the usual amount, and register as failed.  This feels like getting killed for Reasons™, the question is whether those are due to resources?

spiffxp:
All passing tests but "failed" is usually the sign of a job that was killed/interrupted due to timeout, e.g.
- https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/directory/pull-kubernetes-verify/1290209323099623425

hasheddan:
@spiffxp this behavior seems like the standard force push / abort behavior. A quick glance at the first one suggests that is what happened. Do you see differently?

spiffxp:
I hadn't looked that far, that also sounds reasonable

PR #18611, opened 2020-08-02
Report error if job fails due to pod scheduling timeout

### LINKED ISSUE #18528: "Pod scheduling timeout." is displayed as a job failure without additional context by spyglass' metadata lens.

hasheddan:
Updates metadata spyglass lens to show error instead of failed if a job
was not run successfully due to a pod scheduling timeout or other
infrastructure issue.

Example screenshot:

err

Fixes #18528 
xref #18530 

/cc @cjwagner @dims

### DIFF
--- a/prow/spyglass/lenses/metadata/template.html
+++ b/prow/spyglass/lenses/metadata/template.html
@@ -8,6 +8,8 @@
 <p class="test-summary">Test started <abbr id="summary-start-time" title="{{.StartTime}}">{{.StartTime}}</abbr> {{if .Finished -}}
 {{- if .Passed -}}
   <span class="passed">passed</span>
+{{- else if .Errored -}}
+  <span class="failed">error</span>
 {{- else -}}
   <span class="failed">failed</span>
 {{- end -}}
--- a/prow/spyglass/lenses/metadata/lens.go
+++ b/prow/spyglass/lenses/metadata/lens.go
@@ -89,6 +89,7 @@ func (lens Lens) Body(artifacts []api.Artifact, resourceDir string, data string,
 		FinishedTime time.Time
 		Finished     bool
 		Passed       bool
+		Errored      bool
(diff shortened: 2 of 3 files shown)


prow/spyglass/lenses/metadata/template.html
```diff
@@ -8,6 +8,8 @@
 <p class="test-summary">Test started <abbr id="summary-start-time" title="{{.StartTime}}">{{.StartTime}}</abbr> {{if .Finished -}}
 {{- if .Passed -}}
   <span class="passed">passed</span>
+{{- else if .Errored -}}
+  <span class="failed">error</span>
```
cjwagner, line 12: nit: This doesn't seem grammatically consistent, but I would have put `errored` and apparently that isn't even a word 🙃

hasheddan: @cjwagner fair enough :) I am hoping to circle back and hopefully surface additional information about why a scheduling timeout happens, but there isn't much surfaced on the `prowjob.json` at this point so it'll need some work

PR #18607, opened 2020-08-01
Alter bazel-test-canary to run on k8s-infa-prow-build

spiffxp:
Need to make some changes to do so:
- stop using RBE
- take a guess at resource limits (will likely need tuning since the
  guess is based on runs that used RBE)

PR #18613, opened 2020-08-02
Swap periodic-kubernetes-bazel-test-master for canary

ameukam:
Migrate:

https://testgrid.k8s.io/sig-release-master-blocking#bazel-build-master
https://testgrid.k8s.io/sig-release-master-blocking#bazel-test-master

/assign @spiffxp @BenTheElder
/wg k8s-infra

### DIFF
--- a/config/jobs/kubernetes/sig-testing/bazel-build-test.yaml
+++ b/config/jobs/kubernetes/sig-testing/bazel-build-test.yaml
@@ -275,13 +275,12 @@ periodics:
       - 2241179 # release-managers
 
 # periodic bazel test jobs
-- name: periodic-kubernetes-bazel-test-master
+- name: periodic-kubernetes-bazel-test-canary
   annotations:
     fork-per-release: "true"
     fork-per-release-periodic-interval: 6h
-    testgrid-alert-email: kubernetes-sig-testing-alerts@googlegroups.com, k8s-infra-oncall@google.com
-    testgrid-dashboards: sig-release-master-blocking, google-unit
-    testgrid-tab-name: bazel-test-master
+    testgrid-dashboards: sig-testing-canaries
+    testgrid-tab-name: bazel-test
     description: "runs bazel test //... //hack:verify-all -//build/... -//vendor/..."
   interval: 5m
   decorate: true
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes/sig-testing/bazel-build-test.yaml
```diff
@@ -197,6 +197,7 @@ periodics:
     testgrid-tab-name: bazel-build-master
     description: run bazel build //... -//vendor/...
   interval: 5m
+  cluster: k8s-infra-prow-build
```
hasheddan, line 200: @ameukam just a heads up on the test failure, all `k8s-infra-prow-build` jobs require resource limits and requests as part of https://github.com/kubernetes/test-infra/issues/18530

spiffxp:
/hold
These can't be migrated as-is.  See the description of https://github.com/kubernetes/k8s.io/issues/846 for why build can't be migrated, and it's going to require more effort than releng has bandwidth for in 1.19

I'm confirming test can run without RBE via https://github.com/kubernetes/test-infra/pull/18607; if that works, then you would do the same for test

BenTheElder:
/hold
double-hold :-)

super appreciate the effort, the issue is that a major side-effect of these jobs is keeping the bazel cache warm, for use in presubmits
that cache is not cross cluster, and most of the presubmit jobs using the cache can't be moved yet.

we should probably coordinate merging this with being ready to move the rest of the bazel using presubmits

config/jobs/kubernetes/sig-testing/bazel-build-test.yaml
ameukam: Thanks for the heads up. I'll update the PR when we don't need to hold it.

spiffxp:
>a major side-effect of these jobs is keeping the bazel cache warm, for use in presubmits

Hmm, TIL.  Are there any other jobs that could serve a similar role?  Like pull-kubernetes-e2e-gce or something? Or would we want to consider triggering a non-reporting equivalent of this job on k8s-infra-prow-build?

spiffxp:
(I guess I'm effectively keeping k8s-infra-prow-build's bazel cache warm by having migrated periodic-kubernetes-bazel-test-canary over there)

BenTheElder:
it's important that it be a frequent post-submit.

the cache is not aware of presubmit vs potsubmit usage, so we want
postsubmit / periodic to be reasonably frequent, so as the cache is not
drowned out by random PR changes.

eviction is based on least recently accessed.

BenTheElder:
we might be able to disable writes from presubmit, but that is not
currently in effect today.

spiffxp:
Well, at this point then, I'd like to suggest @ameukam change his PR to do the following:
- leave periodic-kubernetes-bazel-build-master untouched
- swap the names of the bazel-test jobs such that
  - periodic-kubernetes-bazel-test-canary = the job that runs on k8s-prow-builds, uses RBE, and is on sig-testing-canaries dashboard
  - periodic-kubernetes-bazel-test-master = the job that runs on k8s-infra-prow-build and is release-blocking

This will leave k8s-prow-builds' bazel cache getting the exact same amount of coverage, while moving the release-blocking test job to a dedicated cluster

spiffxp:
@ameukam if you redo this as I described, that should take care of https://github.com/kubernetes/test-infra/issues/18652

ameukam:
@spiffxp I updated the PR based on my understanding of what you came up with.

config/jobs/kubernetes/sig-testing/bazel-build-test.yaml
ameukam: Added in #18629 #18630.

spiffxp:
/retitle Swap periodic-kubernetes-bazel-test-master for canary

PR #17869, opened 2020-06-07
Update kube dependencies to 1.18.6

alvaroaleman:
This includes some changes to the generated clients, all their methods now take a `context.Context` as first and a `metav1.<<METHOD>>Options` as last argument. Those were fixed via https://github.com/kubernetes-sigs/clientgofix or manually when we re-defined a sub interface.

PR #18618, opened 2020-08-03
fix go-toml issue

andyzhangx:
fixed issue mentioned 
 - https://github.com/kubernetes/kubernetes/pull/93607#issuecomment-667694666
 - https://github.com/kubernetes-sigs/azuredisk-csi-driver/pull/470#issuecomment-667544365

which is caused by `go-toml` version upgrade from v1.6.0 to 1.8.0(introduced by https://github.com/kubernetes/test-infra/pull/17869  3 days ago), let's revert it back to make it work first.

cc @feiskyer @chewong

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -12,6 +12,7 @@ replace (
 	cloud.google.com/go/pubsub => cloud.google.com/go/pubsub v1.3.1
 	github.com/Azure/go-autorest => github.com/Azure/go-autorest v12.2.0+incompatible
 	github.com/docker/docker => github.com/docker/docker v1.4.2-0.20200203170920-46ec8731fbce
+	github.com/pelletier/go-toml => github.com/pelletier/go-toml v1.6.0
 	golang.org/x/lint => golang.org/x/lint v0.0.0-20190409202823-959b441ac422
 	k8s.io/api => k8s.io/api v0.18.6
 	k8s.io/apiextensions-apiserver => k8s.io/apiextensions-apiserver v0.18.6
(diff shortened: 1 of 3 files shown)


BenTheElder:
I don't see any bugs upstream related to this. I do see some stricter parsing in the release notes.

Sounds to me like the secret contents were always invalid?

chewong:
I tested locally with both go-toml v1.8.0 and v1.6.0 and they both work with the expected credential format. The expected credential format is:

```toml
[Creds] 
ClientID = "..."
ClientSecret = "..."
StorageAccountKey = "..."
StorageAccountName = "..."
SubscriptionID = "..."
TenantID = "..."
```

and it was working fine until 2 days ago. I was able to repro the error `error parsing credentials file /etc/azure-cred/credentials (1, 59): parsing error: no value can start with C` when the credential file becomes:

```toml
[Creds] ClientID = "..." ClientSecret = "..." StorageAccountKey = "..." StorageAccountName = "..." SubscriptionID = "..." TenantID = "..."
```

So maybe the content got corrupted? (all new line characters become spaces for some reasons)

chewong:
I just checked that the secret that prow has was indeed in the format of 
```
[Creds] ClientID = "..." ClientSecret = "..." StorageAccountKey = "..." StorageAccountName = "..." SubscriptionID = "..." TenantID = "..."
```
 which works for go-toml v1.6.0 but not v1.8.0.

chewong:
Working with test-infra-on-call to get the secret updated. We could close this PR for now.
/close

Issue #18577, opened 2020-08-01
Make critical jobs Guaranteed Pod QOS: ci-kubernetes-build


spiffxp:
**What should be cleaned up or changed**:

This is part of https://github.com/kubernetes/test-infra/issues/18530

The following jobs should be Guaranteed Pod QOS, meaning they should have CPU and memory resource limits, and matching resource requests:
- ci-kubernetes-build
- ci-kubernetes-build-1-19
- ci-kubernetes-build-stable1
- ci-kubernetes-build-stable2
- ci-kubernetes-build-stable3

These jobs run on (google.com only) k8s-prow-build, so @spiffxp has provided the following guess:
- 7 cpu (or as near 8 as you can get), slightly above 32 Gi mem (34?)

General steps to follow:
- update the job definitions in [/config/jobs](/config/jobs) to have matching CPU and memory limits and requests
  - e.g. ([ci-kubernetes-e2e-gci-gce has a `resources:` field with matching entries](https://github.com/kubernetes/test-infra/blob/2eac54f721ed58479a3126e4f9ca5cbfcc73821b/config/jobs/kubernetes/sig-cloud-provider/gcp/gcp-gce.yaml#L357-L394))
- open a pull request, include a link to this issue and cc `@kubernetes/ci-signal` in the description
- keep an eye on the jobs for the next few days, and if they start failing more than usual, open followup pull requests to raise resources
  - can look at recent prow.k8s.io runs (e.g. https://prow.k8s.io/?job=ci-kubernetes-e2e-gci-gce)
  - can look at job history (e.g. https://prow.k8s.io/job-history/gs/kubernetes-jenkins/logs/ci-kubernetes-e2e-gci-gce)
  - can look at testgrid dashboards (e.g. https://testgrid.k8s.io/sig-release-master-blocking#gce-cos-master-default)
- after the job has remained stable for a few days, declare victory

/sig testing
/sig release
/area jobs
/area release-eng

ZhiFeng1993:
https://prow.k8s.io/?job=ci-kubernetes-build

spiffxp:
/remove-help
since @ZhiFeng1993  has it

ZhiFeng1993:
/close

I think it's safe to close this one. cc. @RobertKielty

spiffxp:
Looking at testgrids

https://testgrid.k8s.io/sig-release-master-blocking#build-master&graph-metrics=test-duration-minutes&width=20 - looks mostly ok

[...]

PR #18620, opened 2020-08-03
Make critical jobs Guaranteed Pod QOS for ci-kubernetes-build-*

ZhiFeng1993:
Add cpu and resource usage as suggested in: https://github.com/kubernetes/test-infra/issues/18577

ref: https://github.com/kubernetes/test-infra/issues/18577

EDIT by spiffxp to change fixes: to ref, don't want to auto-close the issue this corresponds to until we've confirmed the job continues to run healthily

### DIFF
--- a/config/jobs/kubernetes/sig-release/kubernetes-builds.yaml
+++ b/config/jobs/kubernetes/sig-release/kubernetes-builds.yaml
@@ -48,9 +48,12 @@ periodics:
       securityContext:
         privileged: true
       resources:
+        limits:
+          cpu: 7300m
+          memory: "34Gi"
         requests:
-          cpu: 4
-          memory: "8Gi"
+          cpu: 7300m
+          memory: "34Gi"
   rerun_auth_config:
     github_team_ids:
       - 2241179 # release-managers
--- a/config/jobs/kubernetes/sig-release/release-branch-jobs/1.16.yaml
+++ b/config/jobs/kubernetes/sig-release/release-branch-jobs/1.16.yaml
@@ -167,9 +167,12 @@ periodics:
       image: gcr.io/k8s-testimages/bootstrap:v20200729-ee80a02
       name: ""
       resources:
+        limits:
+          cpu: 7300m
+          memory: "34Gi"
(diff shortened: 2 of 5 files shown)


config/jobs/kubernetes/sig-release/kubernetes-builds.yaml
```diff
@@ -48,9 +48,12 @@ periodics:
       securityContext:
         privileged: true
       resources:
+        limits:
+          cpu: 8
```
hasheddan, line 52: This will likely cause pod scheduling failure every time as there is overhead for a few other processes on a node, suggest something like `7300m` (ref: https://github.com/kubernetes/test-infra/pull/18420)

BenTheElder: the values are different for this cluster, because the build jobs are still in google.com prow, and cannot easily be moved (due to the release bucket)

BenTheElder:
> the values are different for this cluster, because the build jobs are still in google.com prow, and cannot easily be moved (due to the release bucket)


cc @spiffxp 
This was not a good first issue, did we label the issue as such?

config/jobs/kubernetes/sig-release/kubernetes-builds.yaml
hasheddan: @BenTheElder ah gotcha, I wonder if we could add some of the basic stats of that cluster to the umbrella issue

BenTheElder:
Daniel: the cluster config is public for the k8s-infra cluster (it's in
k8s.io), the trick is figuring out how much CPU is available on a node.
Currently it's something like 100m less than the google.com cluster,
because network policy is enabled and that addon runs a 100m cpu request
pod on each node (calico).

spiffxp:
/ok-to-test
@BenTheElder labeled as help, not good-first

I'm fine merging this and finding out whether we've asked for too much the hard way

justaugustus:
FYI @kubernetes/ci-signal @kubernetes/release-engineering

PR #18623, opened 2020-08-03
Test using kubetest2 uses go get instead of build from source

michaelmdresser:
Stops test runners from needing kubetest2's build dependencies. Overall more robust. Tested on a GKE cluster with mkpj and mkpod.

Solves a test problem occuring in https://github.com/kubernetes/cloud-provider-gcp/pull/160. Addresses https://github.com/kubernetes/cloud-provider-gcp/issues/161.

### DIFF
--- a/config/jobs/kubernetes/cloud-provider-gcp/cloud-provider-gcp-presubmits.yaml
+++ b/config/jobs/kubernetes/cloud-provider-gcp/cloud-provider-gcp-presubmits.yaml
@@ -34,11 +34,6 @@ presubmits:
       testgrid-dashboards: provider-gcp-presubmits
       description: End to end cluster creation test using kubetest2 based on kubernetes/cloud-provider-gcp.
       testgrid-num-columns-recent: '30'
-    extra_refs:
-    - org: kubernetes-sigs
-      repo: kubetest2
-      base_ref: master
-      path_alias: k8s.io/kubetest2
     spec:
       containers:
       - image: gcr.io/k8s-testimages/kubekins-e2e:v20200802-ff98b58-master
@@ -52,7 +47,9 @@ presubmits:
           set -o pipefail;
           set -o xtrace;
           REPO_ROOT=$(git rev-parse --show-toplevel);
-          cd "${REPO_ROOT}/../kubetest2" || exit 1;
-          make install;
-          make install-deployer-gce;
+          cd;
+          export GO111MODULE=on;
+          go get sigs.k8s.io/kubetest2@latest;
+          go get sigs.k8s.io/kubetest2/kubetest2-gce@latest;
+          go get sigs.k8s.io/kubetest2/kubetest2-tester-exec@latest;
           kubetest2 gce -v 2 --repo-root $REPO_ROOT --build --up --down --test=exec -- kubectl get all --all-namespaces --kubeconfig="${ARTIFACTS}/kubetest2-kubeconfig";


amwat:
/cc @spiffxp @BenTheElder 
for approval

config/jobs/kubernetes/cloud-provider-gcp/cloud-provider-gcp-presubmits.yaml
```diff
@@ -52,7 +47,9 @@ presubmits:
           set -o pipefail;
           set -o xtrace;
           REPO_ROOT=$(git rev-parse --show-toplevel);
-          cd "${REPO_ROOT}/../kubetest2" || exit 1;
-          make install;
-          make install-deployer-gce;
+          cd;
```
MushuEE, line 50: This feels like it really should be part of `extra_refs`. Is this now the recommended way of handling this case?

michaelmdresser: As I understand it, `extra_refs` clones a git repo, which is what we have been doing. However, this doesn't give us built binaries meaning every test runner using kubetest2 also depends on kubetest2's build dependencies (in this case, the issue arose because the test runner does not have Docker). Using `go get` we get built binaries, avoiding this issue.

BenTheElder: working directory is your first extra ref, i'm not sure what this CD is for.

michaelmdresser: `cd` in case of modules path weirdness

config/jobs/kubernetes/cloud-provider-gcp/cloud-provider-gcp-presubmits.yaml
```diff
@@ -52,7 +47,9 @@ presubmits:
...
+          export GO111MODULE=on;
+          go get sigs.k8s.io/kubetest2@latest;
```
BenTheElder, line 52: you can do `go get sigs/k8s.io/kubetest2/...` if they're the same module I think

michaelmdresser: FWIW doing so would install all deployers and testers, yes?

michaelmdresser: (not just the ones we need)

BenTheElder: good point. a concern though is trying to make this easy to replicate locally.

in the future we should not be using HEAD (use a tagged vesion) and we should expect users to be able to user this locally.

BenTheElder: I'd expect most upstream deployers to be cheap to install or moved out of the main module.

michaelmdresser: As far as using HEAD goes, we avoid this problem with the standard e2e workflow by using tagged versions of kubekins with a certain build of kubetest. Autobumper solves updating versions. We don't yet have an equivalent for kubetest2. @amwat mentioned wanting to put together a much smaller e2e test container that provides kubetest2 with similar semantics.

For install, I agree it should be easy. Happy to go either way with it. The eventual new test container should solve this problem completely.

amwat: well this job is meant to be at head intentionally so we want this either way. but yes eventually we can go get tags and or reuse what's baked in the test image. but probably go getting tags is more reproducible anyway.

amwat: @michaelmdresser let's also get a dupe of this job running against kubetest2 as a presubmit.

PR #18629, opened 2020-08-03
critical jobs Guaranteed Pod QOS: periodic-kubernetes-bazel-build

tpepper:
Adding basic resource request and limit for
periodic-kubernetes-bazel-build, which is in the
sig-release-master-blocking dashboard.

Signed-off-by: Tim Pepper <tpepper@vmware.com>

/cc @kubernetes/ci-signal

Fixes: #18581

### DIFF
--- a/config/jobs/kubernetes/sig-release/release-branch-jobs/1.16.yaml
+++ b/config/jobs/kubernetes/sig-release/release-branch-jobs/1.16.yaml
@@ -325,7 +325,13 @@ periodics:
       - bash
       image: gcr.io/k8s-testimages/kubekins-e2e:v20200802-ff98b58-1.16
       name: ""
-      resources: {}
+      resources:
+        limits:
+          cpu: 4
+          memory: "12Gi"
+        requests:
+          cpu: 4
+          memory: "12Gi"
 - annotations:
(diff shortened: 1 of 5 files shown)


config/jobs/kubernetes/sig-release/release-branch-jobs/1.16.yaml
```diff
@@ -325,7 +325,13 @@ periodics:
       - bash
       image: gcr.io/k8s-testimages/kubekins-e2e:v20200802-ff98b58-1.16
       name: ""
-      resources: {}
+      resources:
+        limits:
+          cpu: 4
+          memory: "10Gi"
```
hasheddan, line 331: @tpepper noticed @spiffxp recommended a little more than 10 here. Were you wanting to try this and then bump up?

tpepper: Aaaah I was reading that as start with 10 and see if we need to go up more to 12 based on results.  I'll switch it to 12.

tpepper: amended

PR #18635, opened 2020-08-04
Decoration Utils: Pass on decoration config

alvaroaleman:
This allows pod-creating jobs to re-use their decoration config so it can be passed on.

/assign @stevekuznetsov


prow/pod-utils/decorate/podspec.go
```diff
@@ -60,6 +61,10 @@ const (
 	outputMountName         = "output"
 	outputMountPath         = "/output"
 	oauthTokenFilename      = "oauth-token"
+
+	// DecorationConfigEnv holds a serialized version
+	// of the decoration config for this job.
+	DecorationConfigEnv = "DECORATION_CONFIG"
```
cjwagner, line 67: Rather than creating a new env var, what do you think about updating our existing job spec env var to include the `DecorationConfig` field? That seems cleaner to me given that `DecorationConfig` is part of the `ProwJob.Spec`.
https://github.com/kubernetes/test-infra/blob/f7585b830722bf47c66eb1a39402f76569c00dc9/prow/pod-utils/downwardapi/jobspec.go#L50
https://github.com/kubernetes/test-infra/blob/f7585b830722bf47c66eb1a39402f76569c00dc9/prow/pod-utils/downwardapi/jobspec.go#L103

alvaroaleman:
@cjwagner updated to use the existing env var. I lost patience when adjusting the expected output of the tests, so I made them fixture based (First commit so diff is visible). Let me know what you think.

Issue #18550, opened 2020-07-30
Kubernetes CI Policy: merge-blocking jobs must run in dedicated cluster


spiffxp:
Part of https://github.com/kubernetes/test-infra/issues/18551

**Why this is necessary:**
- we believe declaring Guaranteed Pod QOS jobs may not be defended against Best Effort or Burstable Pods hogging all resources on the same node
- a cluster that only has Guaranteed pods is far more likely to respect resource requirements

**Decisions to make:**
- For all of the jobs being migrated to k8s-infra-prow-build, we're going to use the same node pool as everything else.  To pin to a dedicated node pool will require much more boilerplate (or possibly augmenting prow's preset feature).  If after migrating everything we find we _do_ need a dedicated nodepool, then we'll pay the added cost.
- Jobs that can't be migrated quickly will remain in the default google.com-owned k8s-prow-builds cluster until such time as they can be migrated.  They will continue to compete for resources until they have migrated, and we will be reliant on test-infra-oncall to provide any visibility into the behavior of such jobs.

**Jobs to migrate:**
- These jobs don't need any e2e projects
  - [x] https://github.com/kubernetes/test-infra/issues/18846 - pull-kubernetes-dependencies
  - [x] https://github.com/kubernetes/test-infra/issues/18811 - pull-kubernetes-integration
  - [x] https://github.com/kubernetes/test-infra/issues/18848 - pull-kubernetes-typecheck
  - [x] https://github.com/kubernetes/test-infra/issues/18849 - pull-kubernetes-verify
- These jobs use kind, also don't need e2e projects
  - [x] https://github.com/kubernetes/test-infra/issues/18850 - pull-kubernetes-conformance-kind-ga-only-parallel
  - [x] https://github.com/kubernetes/test-infra/issues/18812 - pull-kubernetes-e2e-kind
  - [x] https://github.com/kubernetes/test-infra/issues/18812 - pull-kubernetes-e2e-kind-ipv6
- These jobs can use standard gce-project e2e projects
  - [x] https://github.com/kubernetes/test-infra/issues/18852 - pull-kubernetes-e2e-gce
  - [x] https://github.com/kubernetes/test-infra/issues/18851 - pull-kubernetes-node-e2e
  - [x] https://github.com/kubernetes/test-infra/issues/18813 - pull-kubernetes-e2e-gce-ubuntu-containerd
[...]

Issue #18549, opened 2020-07-30
Kubernetes CI Policy: release-blocking jobs must run in dedicated cluster


spiffxp:
Part of https://github.com/kubernetes/test-infra/issues/18551

Why it's necessary:
- we believe declaring Guaranteed Pod QOS jobs may not be defended against Best Effort or Burstable Pods hogging all resources on the same node
- a cluster that only has Guaranteed pods is far more likely to respect resource requirements

What decisions do we need to make:
- For all of the jobs being migrated to k8s-infra-prow-build, we're going to use the same node pool as everything else.  To pin to a dedicated node pool will require much more boilerplate (or possibly augmenting prow's preset feature).  If after migrating everything we find we _do_ need a dedicated nodepool, then we'll pay the added cost.
- Jobs that can't be migrated quickly will remain in google.com-owned k8s-prow-builds until we can overcome whatever obstacles are keeping them there.  This means no community visibility into resource consumption, and test-infra-oncall will have to be relied upon.

Boskos projects to be added:
- [x] https://github.com/kubernetes/k8s.io/issues/1095 - gpu-project

Jobs to be migrated
- [x] https://github.com/kubernetes/test-infra/issues/19483 - ci-kubernetes-build (more involved due to https://github.com/kubernetes/k8s.io/issues/846; do this _after_ ci-kubernetes-build-fast)
- [x] https://github.com/kubernetes/test-infra/issues/19484 - ci-kubernetes-build-fast (more involved due to https://github.com/kubernetes/k8s.io/issues/846)
- [x] https://github.com/kubernetes/test-infra/pull/17674 - ci-kubernetes-conformance-kind-ga-only
- [x] https://github.com/kubernetes/test-infra/issues/18650 - ci-kubernetes-e2e-gce-device-plugin-gpu 
- [x] https://github.com/kubernetes/test-infra/pull/17488 - ci-kubernetes-e2e-gce-master-new-gci-kubectl-skew
- [x] https://github.com/kubernetes/test-infra/pull/17672 - ci-kubernetes-e2e-gci-gce
- [x] https://github.com/kubernetes/test-infra/pull/17488 - ci-kubernetes-e2e-gci-gce-alpha-features
- [x] https://github.com/kubernetes/test-infra/issues/18651 - ci-kubernetes-e2e-gci-gce-ingress
- [x] https://github.com/kubernetes/test-infra/pull/17488 - ci-kubernetes-e2e-gci-gce-reboot
[...]

Issue #18507, opened 2020-07-28
ci-kubernetes-conformance-kind-ga-only is experiencing pod scheduling timeout


hasheddan:
**What happened**:

`ci-kubernetes-conformance-kind-ga-only` is experiencing pod scheduling timeout since https://github.com/kubernetes/test-infra/pull/18471

**What you expected to happen**:

For the pod to be scheduled.

**Please provide links to example occurrences, if any**:

https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-conformance-kind-ga-only/1288101820190691330

**Anything else we need to know?**:

@spiffxp I am having trouble determining what is causing the scheduling timeout based on looking at the metrics. Do you have any insight?

/assign

hasheddan:
Pod was scheduled successfully this morning: https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-conformance-kind-ga-only/1288464839429066754

BenTheElder:
https://github.com/kubernetes/test-infra/issues/18530

RobertKielty:
For ref [conformance-all#conformance-ga-only](https://testgrid.k8s.io/conformance-all#conformance-ga-only)

spiffxp:
https://prow.k8s.io/?job=ci-kubernetes-conformance-kind-ga-only

- https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-conformance-kind-ga-only/1290097710933741573
  - There are no nodes that your pod can schedule to - check your requests, tolerations, and node selectors (0/22 nodes are available: 19 Insufficient cpu, 3 node(s) had taints that the pod didn't tolerate.)
- https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-conformance-kind-ga-only/1290279157099401216
  - Job execution failed: Pod scheduling timeout.

Perhaps we should bring the CPU limit down a bit? Or should we somehow be forcing the cluster to be larger?

/assign @BenTheElder

hasheddan:
@spiffxp also seeing this same issue with https://testgrid.k8s.io/sig-release-master-blocking#kind-master-parallel. It appears both of these jobs are requesting a full node (i.e. `7300m` CPU). Is this necessary?

BenTheElder:
The full node is to avoid noisy neighbors.
We cannot schedule for disk IO
[...]

PR #18637, opened 2020-08-04
Give k8s-infra-prow-build cluster time to scale-up

spiffxp:
We have k8s-infra-prow-build setup to autoscale if it doen't have enough capacity. Anecdotally I saw it take 5sec to decide that it needed to scale, but another ~2min for a node to come online. Unfortunately plank deleted the pod after 1min, so we never got to benefit from the added capacity.

Let's assume it could take longer to scale up, so wait up to 5min before calling an unscheduled pod safe to delete.

This is based on investigation I did for period-kubernetes-bazel-test-canary refusing to schedule, ref: https://github.com/kubernetes/test-infra/pull/18607#issuecomment-668287397

### DIFF
--- a/config/prow/config.yaml
+++ b/config/prow/config.yaml
@@ -5,7 +5,7 @@ plank:
   job_url_prefix_config:
     '*': https://prow.k8s.io/view/
   pod_pending_timeout: 15m
-  pod_unscheduled_timeout: 1m
+  pod_unscheduled_timeout: 5m
   default_decoration_configs:
     '*':
       timeout: 2h


spiffxp:
/cc @BenTheElder @hasheddan 
I think this may improve some Pod scheduling timeout woes in general

/cc @chases2 
FYI as test-infra oncall

spiffxp:
This should help with migrate release-blocking jobs to k8s-infra-prow-build (https://github.com/kubernetes/test-infra/issues/18549) and especially help with merge-blocking https://github.com/kubernetes/test-infra/issues/18550

spiffxp:
The job I did this for hasn't failed due to pod scheduling timeout since this landed https://testgrid.k8s.io/sig-testing-canaries#bazel-test&width=20

spiffxp:
I think this has taken care of https://github.com/kubernetes/test-infra/issues/18507

Issue #18596, opened 2020-08-01
Make critical jobs Guaranteed Pod QOS: pull-kubernetes-typecheck


spiffxp:
**What should be cleaned up or changed**:

This is part of https://github.com/kubernetes/test-infra/issues/18530

The following jobs should be Guaranteed Pod QOS, meaning they should have CPU and memory resource limits, and matching resource requests:
- pull-kubernetes-typecheck

These jobs run on (google.com only) k8s-prow-build, so @spiffxp has provided the following guess:
- suggest 5 cpu, slightly above 34 Gi (36?) mem

General steps to follow:
- update the job definitions in [/config/jobs](/config/jobs) to have matching CPU and memory limits and requests
  - e.g. ([ci-kubernetes-e2e-gci-gce has a `resources:` field with matching entries](https://github.com/kubernetes/test-infra/blob/2eac54f721ed58479a3126e4f9ca5cbfcc73821b/config/jobs/kubernetes/sig-cloud-provider/gcp/gcp-gce.yaml#L357-L394))
- open a pull request, include a link to this issue and cc `@kubernetes/ci-signal` in the description
- keep an eye on the jobs for the next few days, and if they start failing more than usual, open followup pull requests to raise resources
  - can look at recent prow.k8s.io runs (e.g. https://prow.k8s.io/?job=ci-kubernetes-e2e-gci-gce)
  - can look at job history (e.g. https://prow.k8s.io/job-history/gs/kubernetes-jenkins/logs/ci-kubernetes-e2e-gci-gce)
  - can look at testgrid dashboards (e.g. https://testgrid.k8s.io/sig-release-master-blocking#gce-cos-master-default)
- after the job has remained stable for a few days, declare victory

/sig testing
/sig release
/area jobs
/area release-eng

spiffxp:
/remove-help
since @RobertKielty has it

RobertKielty:
Checked in with @spiffxp and let him know that I attempted to use phaino to check my job config changes. phaino does not work on jobs that have decorate set to true which this one does.

spiffxp:
This appears to have gotten less noisy since we introduced limits

PR #18648, opened 2020-08-04
adds resource limits/requests to pull-kubernetes-typecheck #18596

RobertKielty:
Pod spec has matching resource requests and limits on both CPU and memory
    and thus will be assigned a Guaranteed QoS class


spiffxp:
Can you update the description to include a link to https://github.com/kubernetes/test-infra/issues/18596 ?

PR #18649, opened 2020-08-04
replace pull-kubernetes-e2e-gce with pull-kubernetes-e2e-gce-ubuntu-c…

BenTheElder:
…ontainerd

see: https://github.com/kubernetes/kubernetes/issues/92937

### DIFF
--- a/config/jobs/kubernetes/sig-cloud-provider/gcp/gcp-gce.yaml
+++ b/config/jobs/kubernetes/sig-cloud-provider/gcp/gcp-gce.yaml
@@ -16,105 +16,7 @@ presets:
 
 presubmits:
   kubernetes/kubernetes:
-  - name: pull-kubernetes-e2e-gce
-    always_run: true
-    skip_branches:
-    - release-\d+\.\d+ # per-release image
-    annotations:
-      fork-per-release: "true"
-      testgrid-alert-stale-results-hours: "24"
-      testgrid-create-test-group: "true"
-      testgrid-num-failures-to-alert: "10"
-    labels:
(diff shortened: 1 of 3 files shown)


BenTheElder:
these jobs are close enough that i'm comfortable doing the replacement in one step (which is also why we're removing one)

for very different jobs we should avoid swapping names.

BenTheElder:
> ERROR: /home/prow/go/src/github.com/kubernetes/test-infra/config/tests/testgrids/BUILD.bazel:25:1: Executing genrule //config/tests/testgrids:testgrid_config failed (Exit 1)
2020/08/04 19:05:29 FAIL: could not write config: 2 errors occurred:
	* could not find the referenced (TestGroup) pull-kubernetes-e2e-gce
	* could not find the referenced (TestGroup) pull-kubernetes-e2e-gce-canary

hrm, these errors are not super useful, that is a whole lot of logs without telling me where the problem is :/

https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/test-infra/18649/pull-test-infra-bazel/1290724701928165376

liggitt:
is there still a way to address this job via a manually triggered presubmit? it would be super handy if that was possible for most postsubmits/periodics to avoid needing to speculatively merge a fix for them without being able to exercise the job

BenTheElder:
agree, I'll rework this
/hold

BenTheElder:
> What is canary in some names represent?

those are manually run jobs used to trial some slight deviation in the config from the one we actually block on (e.g. newer image from test-infra).

that's not specific to this PR

PR #18658, opened 2020-08-04
Revert "Fix seccomp-operator bucket name"

hasheddan:
Reverts kubernetes/test-infra#18541
Fixes https://github.com/kubernetes-sigs/seccomp-operator/issues/64

This change was made thinking we were targeting the wrong bucket. It appears the the correct bucket to use here is the one with the `-gcb` postfix.

/hold to determine this is the correct fix

/cc @saschagrunert @dims

### DIFF
--- a/config/jobs/image-pushing/k8s-staging-seccomp-operator.yaml
+++ b/config/jobs/image-pushing/k8s-staging-seccomp-operator.yaml
@@ -14,7 +14,7 @@ postsubmits:
               - /run.sh
             args:
               - --project=k8s-infra-staging-seccomp-operator
-              - --scratch-bucket=gs://k8s-infra-staging-seccomp-operator
+              - --scratch-bucket=gs://k8s-infra-staging-seccomp-operator-gcb
               - .
             env:
               - name: LOG_TO_STDOUT
--- a/config/jobs/image-pushing/k8s-staging-seccomp-operator.yaml
+++ b/config/jobs/image-pushing/k8s-staging-seccomp-operator.yaml
@@ -13,8 +13,8 @@ postsubmits:
             command:
               - /run.sh
             args:
-              - --project=k8s-infra-staging-seccomp-operator
-              - --scratch-bucket=gs://k8s-infra-staging-seccomp-operator-gcb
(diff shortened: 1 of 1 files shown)


dims:
let's give it a shot :)

hasheddan:
/hold cancel
weeee!

dims:
/hold

hmmm `gs://k8s-staging-seccomp-operator-gcb`

config/jobs/image-pushing/k8s-staging-seccomp-operator.yaml
```diff
@@ -14,7 +14,7 @@ postsubmits:
               - /run.sh
             args:
               - --project=k8s-infra-staging-seccomp-operator
-              - --scratch-bucket=gs://k8s-infra-staging-seccomp-operator
+              - --scratch-bucket=gs://k8s-infra-staging-seccomp-operator-gcb
```
dims, line 17: `gs://k8s-staging-seccomp-operator-gcb`

need to drop `-infra`

hasheddan: Ah nice! Will update

hasheddan: @dims just checking, the fact that `infra` is used here won't be a problem? https://github.com/kubernetes/k8s.io/pull/1017#issuecomment-660949271

hasheddan:
Should be ready now!

config/jobs/image-pushing/k8s-staging-seccomp-operator.yaml
dims: can you please peek at the script output to verify the name of the acl vs name of bucket etc. https://github.com/kubernetes/k8s.io/pull/1045#issuecomment-668820036

(i can check after the call i am in)

dims:
looks legit. please recheck the script output before removing hold :)
/lgtm

hasheddan:
script looks good :+1:  Thanks @dims!

/hold cancel

PR #18667, opened 2020-08-05
Use rebase merge method for k-sigs/seccomp-operator

saschagrunert:
We decided that we wanna use the rebased merge approach for prow to
enforce a linear git history.

/cc @hasheddan @pjbgf


saschagrunert:
Any reason that I had to update the gubernator config? I mean this seems to fail the CI but does not seem to related to my change...

spiffxp:
🤦 yes, the two jobs that were removed are no longer merge-blocking; I don't know how these changes weren't caught in the PRs that made these changes though... thank you for updating

spiffxp:
Opened https://github.com/kubernetes/test-infra/issues/18770 to track the gubernator issue

Issue #18586, opened 2020-08-01
Make critical jobs Guaranteed Pod QOS: pull-kubernetes-cross


spiffxp:
**What should be cleaned up or changed**:

This is part of https://github.com/kubernetes/test-infra/issues/18530

The following jobs should be Guaranteed Pod QOS, meaning they should have CPU and memory resource limits, and matching resource requests:
- pull-kubernetes-cross

These jobs run on (google.com only) k8s-prow-build, so @spiffxp has provided the following guess:
- suggest 7 (or as near 8 as you can get) cpu, slightly above 34 Gi (36?) mem
- jobs that want as much cpu as possible on a 8cpu node can't ask for 8 cpu's: not all is allocatable, and some overhead must be budgeted for 100m init containers that are patched onto the pod by prow (I see 7300m-7500m used in some places, would start on the low end)

General steps to follow:
- update the job definitions in [/config/jobs](/config/jobs) to have matching CPU and memory limits and requests
  - e.g. ([ci-kubernetes-e2e-gci-gce has a `resources:` field with matching entries](https://github.com/kubernetes/test-infra/blob/2eac54f721ed58479a3126e4f9ca5cbfcc73821b/config/jobs/kubernetes/sig-cloud-provider/gcp/gcp-gce.yaml#L357-L394))
- open a pull request, include a link to this issue and cc `@kubernetes/ci-signal` in the description
- keep an eye on the jobs for the next few days, and if they start failing more than usual, open followup pull requests to raise resources
  - can look at recent prow.k8s.io runs (e.g. https://prow.k8s.io/?job=ci-kubernetes-e2e-gci-gce)
  - can look at job history (e.g. https://prow.k8s.io/job-history/gs/kubernetes-jenkins/logs/ci-kubernetes-e2e-gci-gce)
  - can look at testgrid dashboards (e.g. https://testgrid.k8s.io/sig-release-master-blocking#gce-cos-master-default)
- after the job has remained stable for a few days, declare victory

/sig testing
/sig release
/area jobs
/area release-eng

cpanato:
/assign
i will take this one
/remove-help

liggitt:
fyi: pull-kubernetes-cross is optional as of https://github.com/kubernetes/test-infra/pull/18612

PR #18673, opened 2020-08-05
critical jobs Guaranteed Pod QOS: pull-kubernetes-cross

cpanato:
Adding basic resource request and limit for pull-kubernetes-cross
which we can see in the presubmits-kubernetes-nonblocking
dashboards.

 - Fixes: https://github.com/kubernetes/test-infra/issues/18586

/cc @kubernetes/ci-signal

### DIFF
--- a/config/jobs/kubernetes/sig-release/kubernetes-builds.yaml
+++ b/config/jobs/kubernetes/sig-release/kubernetes-builds.yaml
@@ -19,8 +19,11 @@ presubmits:
         securityContext:
           privileged: true
         resources:
+          limits:
+            cpu: 7300m
+            memory: "41Gi"
           requests:
-            cpu: 7
+            cpu: 7300m
             memory: "41Gi"
     annotations:
       testgrid-create-test-group: 'true'


config/jobs/kubernetes/sig-release/kubernetes-builds.yaml
```diff
@@ -19,8 +19,11 @@ presubmits:
         securityContext:
           privileged: true
         resources:
+          limits:
+            cpu: 7300m
+            memory: "41Gi"
```
hasheddan, line 24: Any reason for going higher than recommended `36Gi`?

cpanato: actually not, will fix that

cpanato: done

config/jobs/kubernetes/sig-release/kubernetes-builds.yaml
```diff
@@ -19,9 +19,12 @@ presubmits:
         securityContext:
           privileged: true
         resources:
+          limits:
+            cpu: 7300m
+            memory: "36Gi"
           requests:
-            cpu: 7
-            memory: "41Gi"
```
liggitt, line 24: so this is lowering the request? I think the job runs differently on a 40GB memory boundary (serial vs parallel)

https://github.com/kubernetes/kubernetes/blob/aa0632208ec0a2191e956d86f3406afbb497b881/hack/lib/golang.sh#L307-L310

not sure what we want here

spiffxp: I’m ok if we keep it at 41 for now. It looked like it never hit that peak but we can find out later once we move it off k8s-prow-builds

cpanato: done @spiffxp

PR #18674, opened 2020-08-05
critical jobs Guaranteed Pod QOS: pull-kubernetes-e2e-gce-100-performance

cpanato:
Adding basic resource request and limit for
pull-kubernetes-e2e-gce-100-performance,
which we can see in the presubmits-kubernetes-blocking, 
presubmits-kubernetes-scalability dashboards.

- Fixes: https://github.com/kubernetes/test-infra/issues/18589

/cc @kubernetes/ci-signal

### DIFF
--- a/config/jobs/kubernetes/sig-release/release-branch-jobs/1.16.yaml
+++ b/config/jobs/kubernetes/sig-release/release-branch-jobs/1.16.yaml
@@ -964,8 +964,12 @@ presubmits:
         image: gcr.io/k8s-testimages/kubekins-e2e:v20200802-ff98b58-master
         name: ""
         resources:
+          limits:
+            cpu: 4
+            memory: "14Gi"
           requests:
-            memory: 6Gi
+            cpu: 4
+            memory: "14Gi"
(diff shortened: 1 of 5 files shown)


config/jobs/kubernetes/sig-release/release-branch-jobs/1.16.yaml
```diff
@@ -964,8 +964,12 @@ presubmits:
         image: gcr.io/k8s-testimages/kubekins-e2e:v20200802-ff98b58-master
         name: ""
         resources:
+          limits:
+            cpu: 4
+            memory: "13Gi"
```
hasheddan, line 969: I think we need a bit over 13 here per https://github.com/kubernetes/test-infra/issues/18589

cpanato: ok, i did not fully get that :) will change

cpanato: done

PR #18676, opened 2020-08-05
Implements 18592 QoS Guarantee for pull-kubernetes-files-remake

RobertKielty:
Adds matching limits and requests for cpu and memory : 1500m cpu, 7Gi mem

Part of #18530
cc @kubernetes/ci-signal

### DIFF
--- a/config/jobs/kubernetes/sig-testing/files-remake.yaml
+++ b/config/jobs/kubernetes/sig-testing/files-remake.yaml
@@ -27,3 +27,10 @@ presubmits:
         # Space separated list of the checks to run
         - name: WHAT
           value: generated-files-remake
+        resources:
+          limits:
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes/sig-testing/kubernetes-kind-presubmits.yaml
```diff
@@ -283,13 +283,12 @@ presubmits:
         securityContext:
           privileged: true
         resources:
+          limits:
+            cpu: 4
+            memory: 9Gi
```
spiffxp, line 288: This should be a different PR

RobertKielty: I'll amend

PR #18685, opened 2020-08-05
Add extra-version-markers support to kubernetes_build.py

hasheddan:
Add the ability to specify --extra-version-markers when pushing a build,
which is replacing the deprecated --extra-publish-file arg.

Follow up to https://github.com/kubernetes/release/pull/1385. Will follow with migrating build jobs to `--extra-version-markers`. Motivation is https://github.com/kubernetes/kubernetes/issues/93694#issuecomment-669209941. Should also fix `push-build` to still honor `--extra-publish-file`.

/hold for explicit approval from @justaugustus 

/cc @justaugustus @kubernetes/release-engineering

### DIFF
--- a/scenarios/kubernetes_build.py
+++ b/scenarios/kubernetes_build.py
@@ -124,6 +124,8 @@ def main(args):
         push_build_args.append('--docker-registry=%s' % args.registry)
     if args.extra_publish_file:
         push_build_args.append('--extra-publish-file=%s' % args.extra_publish_file)
+    if args.extra_version_markers:
+        push_build_args.append('--extra-version-markers=%s' % args.extra_version_markers)
     if args.fast:
         push_build_args.append('--fast')
     if args.allow_dup:
@@ -162,6 +164,8 @@ if __name__ == '__main__':
         '--registry', help='Push images to the specified docker registry')
     PARSER.add_argument(
         '--extra-publish-file', help='Additional version file uploads to')
+    PARSER.add_argument(
+        '--extra-version-markers', help='Additional version file uploads to')
     PARSER.add_argument(
         '--fast', action='store_true', help='Specifies a fast build')
     PARSER.add_argument(
--- a/config/jobs/kubernetes/sig-release/kubernetes-builds.yaml
+++ b/config/jobs/kubernetes/sig-release/kubernetes-builds.yaml
@@ -42,7 +42,7 @@ periodics:
       - --scenario=kubernetes_build
       - --
       - --allow-dup
-      - --extra-publish-file=k8s-master
+      - --extra-version-markers=k8s-master
       - --registry=gcr.io/kubernetes-ci-images
       # docker-in-docker needs privileged mode
       securityContext:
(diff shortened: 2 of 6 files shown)


hasheddan:
/cc @tpepper 

Saw you comment on the issue

hasheddan:
@justaugustus any thoughts here?

justaugustus:
Spoke w/ Dan here: https://kubernetes.slack.com/archives/CJH2GBF7Y/p1596734676114900?thread_ts=1596636797.056600&cid=CJH2GBF7Y

The current logic in push-build.sh is meant to honor both the new and old flag: https://github.com/kubernetes/release/blob/8e0a83384d70a90826e275eeec318a821fc69962/push-build.sh#L176

This leads me to believe there's an issue with how the extra version marker array is being built, not something in k/test-infra.
Let's try to discover that issue w/ some debug logging before moving forward with this.

justaugustus:
> This leads me to believe there's an issue with how the extra version marker array is being built, not something in k/test-infra.

There was. Fixed in https://github.com/kubernetes/release/pull/1481. Thanks Dan!

> Let's try to discover that issue w/ some debug logging before moving forward with this.

Added in https://github.com/kubernetes/release/pull/1471.

---

This should be good to approve once https://github.com/kubernetes/test-infra/pull/18688 is pulled into the PR.

hasheddan:
@justaugustus should be good now :)

justaugustus:
Thanks!
/lgtm
/approve
/assign @cjwagner @fejta 
/hold cancel

scenarios/kubernetes_build.py
```diff
@@ -162,6 +164,8 @@ def main(args):
         '--registry', help='Push images to the specified docker registry')
     PARSER.add_argument(
         '--extra-publish-file', help='Additional version file uploads to')
+    PARSER.add_argument(
+        '--extra-version-markers', help='Additional version file uploads to')
```
MushuEE, line 168: Does this help need to be updated? Or should it really match `extra-publish-file`?

hasheddan: It is serving the same purpose as `--extra-publish-file` (i.e. it will just overwrite `--extra-publish-file` if present), so it is likely fine to match unless we just want a new description altogether

hasheddan:
@fejta @cjwagner could you take a quick look here?

scenarios/kubernetes_build.py
cjwagner: It is unclear to me how adding another flag with identical behavior will fix anything. I don't have the context to understand the motivation linked in the PR body though. Could you clarify what the problem is and how this fixes it? Why can't the callers of the script just use `--extra-publish-file`?

hasheddan: @cjwagner adding the flag is not fixing anything, but it enables consumers to use the flag instead of `--extra-publish-file`, which has been deprecated in [push-build.sh](https://github.com/kubernetes/release/blob/b29efe949dfebf3ce4dbd277111924d089940bc3/push-build.sh#L49). Eventually I would expect `--extra-publish-file` to be removed.

cjwagner: Got it. LGTM then.

PR #18688, opened 2020-08-05
Replace extra-publish-file with extra-version-markers in build jobs

hasheddan:
Updates build jobs to use --extra-version-markers instead of the
deprecated --extra-publish-file.

/hold for #18685 
/cc @justaugustus @kubernetes/release-engineering


hasheddan:
/close

in favor of https://github.com/kubernetes/test-infra/pull/18685

Issue #18591, opened 2020-08-01
Make critical jobs Guaranteed Pod QOS: pull-kubernetes-e2e-kind


spiffxp:
**What should be cleaned up or changed**:

This is part of https://github.com/kubernetes/test-infra/issues/18530

The following jobs should be Guaranteed Pod QOS, meaning they should have CPU and memory resource limits, and matching resource requests:
- pull-kubernetes-e2e-kind

These jobs run on (google.com only) k8s-prow-build, so @spiffxp has provided the following guess:
- suggest 4 cpu, slightly above 8 Gi (9?) mem

General steps to follow:
- update the job definitions in [/config/jobs](/config/jobs) to have matching CPU and memory limits and requests
  - e.g. ([ci-kubernetes-e2e-gci-gce has a `resources:` field with matching entries](https://github.com/kubernetes/test-infra/blob/2eac54f721ed58479a3126e4f9ca5cbfcc73821b/config/jobs/kubernetes/sig-cloud-provider/gcp/gcp-gce.yaml#L357-L394))
- open a pull request, include a link to this issue and cc `@kubernetes/ci-signal` in the description
- keep an eye on the jobs for the next few days, and if they start failing more than usual, open followup pull requests to raise resources
  - can look at recent prow.k8s.io runs (e.g. https://prow.k8s.io/?job=ci-kubernetes-e2e-gci-gce)
  - can look at job history (e.g. https://prow.k8s.io/job-history/gs/kubernetes-jenkins/logs/ci-kubernetes-e2e-gci-gce)
  - can look at testgrid dashboards (e.g. https://testgrid.k8s.io/sig-release-master-blocking#gce-cos-master-default)
- after the job has remained stable for a few days, declare victory

/sig testing
/sig release
/area jobs
/area release-eng

spiffxp:
/remove-help
since @ameukam has it

spiffxp:
https://prow.k8s.io/?job=pull-kubernetes-e2e-kind&state=error

Seeing jobs in error state this morning.  Most timing out after ~5min (which I'm assuming means they couldn't get the resources they wanted).

Two timing out at 15-17min

ameukam:
With #18817 merged, things we can watch :
[...]

PR #18691, opened 2020-08-05
Configure Guaranteed Pod QoS for pull-kubernetes-e2e-kind

ameukam:
Set resources limits and requests as suggested.

Ref: https://github.com/kubernetes/test-infra/issues/18591
Part of: https://github.com/kubernetes/test-infra/issues/18530

/assign @spiffxp @BenTheElder 
cc @kubernetes/ci-signal

### DIFF
--- a/config/jobs/kubernetes/sig-release/release-branch-jobs/1.16.yaml
+++ b/config/jobs/kubernetes/sig-release/release-branch-jobs/1.16.yaml
@@ -1203,8 +1203,11 @@ presubmits:
         image: gcr.io/k8s-testimages/kubekins-e2e:v20200802-ff98b58-1.16
         name: ""
         resources:
+          limits:
+            cpu: 4
+            memory: 9000Mi
           requests:
-            cpu: "2"
+            cpu: 4
             memory: 9000Mi
         securityContext:
           privileged: true
--- a/config/jobs/kubernetes/sig-testing/kubernetes-kind-presubmits.yaml
+++ b/config/jobs/kubernetes/sig-testing/kubernetes-kind-presubmits.yaml
@@ -37,9 +37,12 @@ presubmits:
         securityContext:
           privileged: true
         resources:
-          requests:
-            memory: "9000Mi"
+          limits:
             cpu: 7400m
+            memory: 9000Mi
+          requests:
+            cpu: 7400m
+            memory: 9000Mi
     annotations:
       testgrid-num-failures-to-alert: '10'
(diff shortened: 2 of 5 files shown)


config/jobs/kubernetes/sig-testing/kubernetes-kind-presubmits.yaml
```diff
@@ -37,9 +37,12 @@ presubmits:
         securityContext:
           privileged: true
         resources:
+          limits:
+            cpu: 4
+            memory: 9000Mi
           requests:
-            memory: "9000Mi"
-            cpu: 7400m
+            cpu: 4
```
BenTheElder, line 44: while kind jobs still have to do a build we should not limit the CPU this low
these jobs are equivilant to the presubmit in resource usage, unlike most CI jobs

RobertKielty:
Fixes #18591

RobertKielty:
Last comment added  so as to link this PR to it's corresponding Issue on the project board https://github.com/orgs/kubernetes/projects/46

ameukam:
@BenTheElder Once this is merged, what is the success rate we should expect to consider this done? Just want to know how many days we need to monitor these jobs.

BenTheElder:
It's hard to say, presubmit success rate somewhat depends on the PRs incoming, and it varies by branch.

config/jobs/kubernetes/sig-release/release-branch-jobs/1.16.yaml
```diff
@@ -1203,8 +1203,11 @@ presubmits:
         image: gcr.io/k8s-testimages/kubekins-e2e:v20200802-ff98b58-1.16
         name: ""
         resources:
+          limits:
+            cpu: 4
+            memory: 9000Mi
           requests:
-            cpu: "2"
+            cpu: 4
```
BenTheElder, line 1210: what CPU are we giving the other bazel jobs currently?

spiffxp: 4 for the jobs that use RBE, I'm still trying to find something for the jobs that don't

BenTheElder: hmm ok
I'm nervous that limiting the CPU is going to cause build timeouts, but I guess I'm willing to try it

BenTheElder:
/approve
/lgtm
/hold
please be prepared to increase the resources on the older releases if we see issues, I think we might

PR #18692, opened 2020-08-05
Adding Job with multiple more complicated test containers

anthonyrtong:
Job is a combination of ci-test-infra-bazel and ci-test-infra-benchmark-demo to test more complicated multi container behavior.

### DIFF
--- a/config/jobs/kubernetes/test-infra/test-infra-canaries.yaml
+++ b/config/jobs/kubernetes/test-infra/test-infra-canaries.yaml
@@ -53,3 +53,40 @@ periodics:
     testgrid-dashboards: sig-testing-canaries
     testgrid-tab-name: multiple-container-test
     description: echo from two different containers
+- interval: 1h
+  name: ci-test-infra-larger-multiple-container-test
+  decorate: true
+  extra_refs:
+  - org: kubernetes
+    repo: test-infra
+    base_ref: master
+    path_alias: k8s.io/test-infra
+  labels:
+    preset-service-account: "true"
+    preset-bazel-scratch-dir: "true"
+  spec:
+    containers:
+    - name: benchmark-junit-container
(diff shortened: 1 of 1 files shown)


anthonyrtong:
ptal @e-blackwelder

config/jobs/kubernetes/test-infra/test-infra-canaries.yaml
```diff
@@ -53,3 +53,40 @@ periodics:
...
+- interval: 1h
+  name: ci-test-infra-larger-multiple-container-test
+  decorate: true
+  extra_refs:
+  - org: kubernetes
+    repo: test-infra
+    base_ref: master
+    path_alias: k8s.io/test-infra
+  labels:
+    preset-service-account: "true"
```
cjwagner, line 65: This should not be required. It is rampantly misused for legacy reasons, but should only be required for old undecorated jobs that are still using the deprecated `bootstrap.py` instead of the pod utilities.

config/jobs/kubernetes/test-infra/test-infra-canaries.yaml
```diff
@@ -53,3 +53,40 @@ periodics:
...
+      command:
+      - hack/bazel.sh
+      args:
+      - test
+      - --config=ci
+      - --nobuild_tests_only
+      - //...
+  annotations:
+    testgrid-alert-email: anthonytong@google.com
+    ProwMultipleContainerSupport: Yes, I know what I'm doing.
```
cjwagner, line 89: This is no longer required, right?

anthonyrtong: Yeah, that PR to remove the annotation is merged now.  I'll remove it for both of the two jobs.

PR #18695, opened 2020-08-06
Centralize the create-build-cluster.sh script and make merge_kubeconfig_secret.py handle secret keys intelligently.

cjwagner:
I recommend reviewing by commit as the second commit just copies [`create-build-cluster.sh`](https://github.com/GoogleCloudPlatform/oss-test-infra/blob/master/prow/create-build-cluster.sh) from GoogleCloudPlatform/oss-test-infra. This script is generally useful for any Prow instance so I'd like to upstream it and centralize it like `pj-on-kind.sh`.

The changes to `merge_kubeconfig_secret.py` have two effects:
- A required `--context` flag was added so that users must explicitly identify the cluster they wish to update.
- An `--auto` mode was added to automatically determine the `--dest-key` and `--src-key` values assuming that keys are formatted like `config-20200730`. This will make it simpler for oncall to update the secret and promote a consistent key formatting pattern that makes roll backs easy.

Addresses https://github.com/GoogleCloudPlatform/oss-test-infra/issues/400

/assign @chaodaiG @michelle192837 @alvaroaleman

### DIFF
--- a/gencred/README.md
+++ b/gencred/README.md
@@ -143,10 +143,13 @@ users:
 If you store kubeconfig files in kubernetes secrets to allow pods to access other kubernetes clusters (like many of Prow's components require) consider using [`merge_kubeconfig_secret.py`](/gencred/merge_kubeconfig_secret.py) to merge the kubeconfig produced by `gencred` into the secret.
 
 ```shell
-# Generate a kubeconfig.yaml as described and shown above, then run something like:
-./merge_kubeconfig_secret.py --src-key=config-old --dest-key=config-new kubeconfig.yaml
-# Update references (e.g. `--kubeconfig` flags) to point to config-new instead of config-old.
+# Generate a kubeconfig.yaml as described and shown above.
+./merge_kubeconfig_secret.py --auto --context=my-kube-context kubeconfig.yaml
+# Note: The first time the script is used you may be prompted to rerun it with --src-key specified.
+# Finish by updating references (e.g. `--kubeconfig` flags in Prow deployment files) to point to the updated secret key. The script will indicate which key was updated in its output.
 ```
+
+The script exposes optional flags to override the secret namespace, name, keys, and pruning behavior. Run `./merge_kubeconfig_secret.py --help` to view all options.
 
 ### Library
 
--- a/gencred/merge_kubeconfig_secret.py
+++ b/gencred/merge_kubeconfig_secret.py
@@ -28,9 +28,13 @@ precedence to the secret."""
 
 import argparse
 import os
+import re
 import subprocess
 import sys
 import tempfile
+import time
+
+reAutoKey = re.compile('^config-(\\d{8})$')
 
 def call(cmd, **kwargs):
     print('>>> %s' % cmd)
@@ -47,11 +51,31 @@ def call(cmd, **kwargs):
 
 def main(args):
     print(args)
+    validateArgs(args)
+    if args.auto:
+        # We need to determine the dest key automatically.
+        args.dest_key = time.strftime('config-%Y%m%d')
+        if not args.src_key:
+            # Also try to automatically determine the src key.
+            cmd = 'kubectl --context="%s" get secret --namespace "%s" "%s" -o go-template="{{range \\$key, \\$content := .data}}{{\\$key}};{{end}}"' % (args.context, args.namespace, args.name) #pylint: disable=line-too-long
+            keys = call(cmd).stdout.rstrip(";").split(";")
+            matches = [key for key in keys if reAutoKey.match(key)]
+            matches.sort(reverse=True)
+            if len(matches) == 0:
+                raise ValueError('The %s/%s secret does not contain any keys matching the "config-20200730" format. Please try again with --src-key set to the most recent key. Existing keys: %s' % (args.namespace, args.name, keys)) #pylint: disable=line-too-long
+
+            args.src_key = matches[0]
+        # Only enable pruning if we won't overwrite the source key.
+        # This ensures that a second update on the same day will still have a
+        # key to roll back to if needed.
+        args.prune = args.src_key != args.dest_key
+        print('Automatic mode: --src-key=%s  --dest-key=%s' % (args.src_key, args.dest_key))
+
     with tempfile.TemporaryDirectory() as tmpdir:
         orig = '%s/original' % (tmpdir)
         merged = '%s/merged' % (tmpdir)
         # Copy the current secret contents into a temp file.
-        cmd = 'kubectl get secret --namespace "%s" "%s" -o go-template="{{index .data \\"%s\\"}}" | base64 -d > %s' % (args.namespace, args.name, args.src_key, orig) #pylint: disable=line-too-long
+        cmd = 'kubectl --context="%s" get secret --namespace "%s" "%s" -o go-template="{{index .data \\"%s\\"}}" | base64 -d > %s' % (args.context, args.namespace, args.name, args.src_key, orig) #pylint: disable=line-too-long
         call(cmd)
 
         # Merge the existing and new kubeconfigs into another temp file.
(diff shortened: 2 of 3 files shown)


gencred/README.md
```diff
@@ -143,11 +143,14 @@ users:
 If you store kubeconfig files in kubernetes secrets to allow pods to access other kubernetes clusters (like many of Prow's components require) consider using [`merge_kubeconfig_secret.py`](/gencred/merge_kubeconfig_secret.py) to merge the kubeconfig produced by `gencred` into the secret.
 
 ```shell
-# Generate a kubeconfig.yaml as described and shown above, then run something like:
-./merge_kubeconfig_secret.py --src-key=config-old --dest-key=config-new kubeconfig.yaml
-# Update references (e.g. `--kubeconfig` flags) to point to config-new instead of config-old.
+# Generate a kubeconfig.yaml as described and shown above.
+./merge_kubeconfig_secret.py --auto --context=my-kube-context kubeconfig.yaml
```
alvaroaleman, line 147: To avoid the need for this, the `KUBECONFIG` env var can be used to supply mulitple kubeconfigs: https://github.com/openshift/release/blob/d534072d3f09e94d1a070469dd27a7da2b1d1b63/clusters/app.ci/prow/03_deployment/sinker.yaml#L36

Has the nice sideeffect that adding/removing a cluster causes a rollout and secret size is not a concern, as each kubeconfig can get its own secret

prow/create-build-cluster.sh
```diff
@@ -0,0 +1,139 @@
...
+    authed="true"
+  fi
+}
+function createProject() {
+  # Create project, configure billing, enable GKE, add IAM rule for oncall team.
+  echo "Creating project '${PROJECT}' (this may take a few minutes)..."
+  gcloud projects create "${PROJECT}" --name="${PROJECT}" --folder="${FOLDER_ID}"
+  gcloud beta billing projects link "${PROJECT}" --billing-account="${BILLING_ACCOUNT_ID}"
+  gcloud services enable "container.googleapis.com" --project="${PROJECT}"
+  gcloud projects add-iam-policy-binding "${PROJECT}" --member="group:mdb.cloud-kubernetes-engprod-oncall@google.com" --role="roles/owner"
```
alvaroaleman, line 82: should `member` here  be configurable? Or is the script intended for google-managed clusters only?

gencred/README.md
cjwagner: That is something I considered originally and that is how the script actually does the merge, but I figured that a single kubeconfig would be easier to manage. Here are my thoughts, let me know what you think:
- I prefer using flags rather than using env vars since it is clear how flags are consumed.
- Order matters when using the `KUBECONFIG` env var and accidentally adding a kubeconfig to the front of the list instead of the back would cause lots of problems due to the `current-context` changing.

> Has the nice sideeffect that adding/removing a cluster causes a rollout and secret size is not a concern, as each kubeconfig can get its own secret

- Adding a cluster using `./merge_kubeconfig_secret.py` and updating the `--kubeconfig` flag causes a rollout in the same way that adding/updating a secret and updating the `KUBECONFIG` env var would.
- Having a separate secret for every kubeconfig would require a lot of extra mounts in the deployment files which isn't a problem, but would be more annoying to read than just a single kubeconfig secret mount.
- Secret size doesn't really seem like a concern for our use cases at least. The k8s-prow kubeconfig is only 36KB and the secret size limit is 1MB. If an instance does eventually reach the secret size limit we could start using the `KUBECONFIG` env var with multiple secrets, but even then I think we'd want to consolidate into a small number of kubeconfigs rather than a kubeconfig per cluster since we'd have a huge number of clusters if we are hitting the size limit.

prow/create-build-cluster.sh
cjwagner: It should definitely be configurable. Fixed.
I left `group:mdb.cloud-kubernetes-engprod-oncall@google.com` as the default to provide an example value.

gencred/README.md
```diff
@@ -143,11 +143,14 @@ users:
...
+# Note: The first time the script is used you may be prompted to rerun it with --src-key specified.
+# Finish by updating references (e.g. `--kubeconfig` flags in Prow deployment files) to point to the updated secret key. The script will indicate which key was updated in its output.
 ```
 
+You script exposes optional flags to override the secret namespace, name, keys, and pruning behavior. Run `./merge_kubeconfig_secret.py --help` to view all options.
```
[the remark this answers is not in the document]
cjwagner: 😄 Yes I changed the wording, but apparently not completely.

gencred/merge_kubeconfig_secret.py
```diff
@@ -108,7 +145,12 @@ def main(args):
...
         '--prune',
         action='store_true',
-        help='Remove all secret keys besides the source and dest. This should be used periodically to delete old kubeconfigs and keep the secret size under control.' #pylint: disable=line-too-long
-        )
+        help='Remove all secret keys besides the source and dest. This should be used periodically to delete old kubeconfigs and keep the secret size under control.', #pylint: disable=line-too-long
+    )
+    parser.add_argument(
+        '--auto',
+        action='store_true',
+        help='Automatically determine --dest-key and optionally --src-key assuming keys are of the form "config-20200730". Pruning is enabled.', #pylint: disable=line-too-long
```
chases2, line 153: "config-YYYYMMDD" will be clearer after time passes

gencred/merge_kubeconfig_secret.py
```diff
@@ -68,19 +92,34 @@ def main(args):
...
-            call('kubectl create secret generic --namespace "%s" "%s" --from-file="%s=%s" %s --dry-run -oyaml | kubectl replace -f -' % (args.namespace, args.name, args.dest_key, merged, srcflag)) #pylint: disable=line-too-long
+            call('kubectl --context="%s" create secret generic --namespace "%s" "%s" --from-file="%s=%s" %s --dry-run -oyaml | kubectl replace -f -' % (args.context, args.namespace, args.name, args.dest_key, merged, srcflag)) #pylint: disable=line-too-long
         else:
             content = ''
             with open(merged, 'r') as mergedFile:
                 yamlPad = '    '
                 content = yamlPad + mergedFile.read()
                 content = content.replace('\n', '\n' + yamlPad)
-            call('kubectl patch --namespace "%s" "secret/%s" --patch "stringData:\n  %s: |\n%s\n"' % (args.namespace, args.name, args.dest_key, content)) #pylint: disable=line-too-long
+            call('kubectl --context="%s" patch --namespace "%s" "secret/%s" --patch "stringData:\n  %s: |\n%s\n"' % (args.context, args.namespace, args.name, args.dest_key, content)) #pylint: disable=line-too-long
```
chases2, line 102: I think large templates like this should make it clear what arguments it's taking with a line break.
```
call('very long template'
    % (arg1, arg2))
```

gencred/README.md
alvaroaleman: > Order matters when using the KUBECONFIG env var and accidentally adding a kubeconfig to the front of the list instead of the back would cause lots of problems due to the current-context changing.

I would have expected us to never care about `current-context?



Generally I prefer to have it explicitly and see in the config which clusters are referenced rather than having to look into the contents of a secret. But at the end of the day its probably a question of taste, both work.

/lgtm

prow/create-build-cluster.sh
```diff
@@ -0,0 +1,154 @@
...
+  # Create a service account for uploading to GCS.
+  gcloud beta iam service-accounts create "${sa}" --project="${PROJECT}" --description="SA for Prow's pod utilities to use to upload job results to GCS." --display-name="Prow Pod Utilities"
+  # Generate private key and attach to the service account.
+  gcloud iam service-accounts keys create "sa-key.json" --project="${PROJECT}" --iam-account="${saFull}"
+  kubectl create secret generic "service-account" -n "test-pods" --from-file="service-account.json=sa-key.json"
+  echo
+  echo "Please ask the test-infra oncall (https://go.k8s.io/oncall) to run the following:"
+  echo "  gsutil acl ch -u \"${saFull}:O\" \"gs://${GCS_BUCKET}\""
+  echo
+  echo "Press any key to aknowledge (this doesn't need to be completed to continue this script, but it needs to be done before uploading will work)..."
```
chases2, line 120: Consider splitting into two shorter log lines:
```
echo "This must be done before uploading will work"
echo "However, the script can continue (press any key)..."
```
Consider splitting the occurrence in `gencreds()` also.

nit: missing 'c' in acknowledge

cjwagner:
Thanks for the reviews! I'm going to ignore the clarity nits and merge this since it is needed now and I think the UX is good enough given that this is for Prow admins. I'll incorporate them if we need to make more changes later.
/hold cancel

PR #18699, opened 2020-08-06
Add podman label to the kind project

aojea:
KIND currently has 2 providers, docker as default and podman as experimental.

We add a label for podman provider In order to differentiate issues and PRs

Fixes: https://github.com/kubernetes-sigs/kind/issues/1765

### DIFF
--- a/label_sync/labels.yaml
+++ b/label_sync/labels.yaml
@@ -1543,3 +1543,15 @@ repos:
         name: kind/user-story
         target: issues
         addedBy: humans
+  kubernetes-sigs/kind:
+    labels:
+      - color: 19abff
+        description: Issues or PRs related to docker
+        name: area/provider/docker
+        target: both
+        addedBy: humans
+      - color: 970fff
+        description: Issues or PRs related to podman
+        name: area/provider/podman
+        target: both
+        addedBy: humans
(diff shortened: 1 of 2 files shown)


label_sync/labels.yaml
```diff
@@ -1543,3 +1543,10 @@ repos:
         name: kind/user-story
         target: issues
         addedBy: humans
+  kubernetes-sigs/kind:
+    labels:
+      - color: 970fff
+        description: Issues or PRs related to podman
+        name: area/provider/podman
```
BenTheElder, line 1550: we should add both?

aojea: I though about that, but this is very specific and the other is a catch all.
are we going to use the docker label it at the end of the day ?

spiffxp: Is "provider" the right term here?  I see it overloaded/used in so many contexts throughout the project (e.g. cluster provider, cloud provider, credentials provide, test/e2e/frameworks' "provider).  Is there a more specific term that would make more sense?

aojea: thing is that in this case it matches perfectly :smile: 
https://github.com/kubernetes-sigs/kind/tree/master/pkg/cluster/internal/providers/podman

BenTheElder: > are we going to use the docker label it at the end of the day ?

yes, when there are docker provider specific bugs

aojea: What about aaron comment? do we want another term instead of provider? I'm very bad a naming so I will copy it verbatim

BenTheElder: "node provider" is the full term but I don't think that's helpful here.
it's OK because area/ is repo specific normally anyhow

BenTheElder:
`/area provider/foo` will work once this merges

BenTheElder:
btw example of docker https://github.com/kubernetes-sigs/kind/issues/1732

PR #18707, opened 2020-08-06
Print output of push build call

hasheddan:
Prints the output of push build when called as a subprocess in
kubernetes_build.py scenario.

I _think_ this will print the output --> `check_output` runs command and returns output

/cc @justaugustus

### DIFF
--- a/scenarios/kubernetes_build.py
+++ b/scenarios/kubernetes_build.py
@@ -142,7 +142,8 @@ def main(args):
         check('make', 'quick-release')
     else:
         check('make', 'release')
-    check(args.push_build_script, *push_build_args)
+    output = check_output(args.push_build_script, *push_build_args)
+    print >>sys.stderr, 'Push build result: ', output
 
 if __name__ == '__main__':
     PARSER = argparse.ArgumentParser(


scenarios/kubernetes_build.py
```diff
@@ -142,7 +142,7 @@ def main(args):
         check('make', 'quick-release')
     else:
         check('make', 'release')
-    check(args.push_build_script, *push_build_args)
+    print(check_output(args.push_build_script, *push_build_args))
```
justaugustus, line 145:
```suggestion
    check_output(args.push_build_script, *push_build_args)
```

?
I'm not sure, tbh.

hasheddan: @dims please bless us with your python knowledge :)

justaugustus: @hasheddan -- given the usage elsewhere, I think just `check_output` should work.

hasheddan: @justaugustus I see instances of this being set to a var, but I don't see any examples of this being called in a place where it is obvious that the returned value is printed

hasheddan:
Thanks @dims! @justaugustus ready for another review

dims:
/approve

leaving lgtm for @justaugustus

BenTheElder:
we should really stop using scenarios / bootstrap

justaugustus:
> we should really stop using scenarios / bootstrap

One day, my friend. One day.
_Looks wistfully off into the distance_

/lgtm
/approve

Issue #18708, opened 2020-08-06
post-test-infra-push-triage incorrectly reporting URL of built images


michaelkolber:
**What happened**:

A `post-test-infra-push-triage` job reported that an image was pushed to gcr.io/k8s-prow.

**What you expected to happen**:

The job should report that the image was pushed to gcr.io/k8s-testimages, where it was actually pushed to, instead of incorrectly saying it was pushed to gcr.io/k8s-prow. This looks like it may be a larger issue with [GCB Builder](https://github.com/kubernetes/test-infra/tree/master/images/builder).

**How to reproduce it (as minimally and precisely as possible)**:

Run `post-test-infra-push-triage`.

**Please provide links to example occurrences, if any**:

[Build log](https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-test-infra-push-triage/1291448901563322368):
- Note line 43, with the incorrect project name/URL:
```
2020/08/06 19:07:14 Successfully built image: gcr.io/k8s-prow/triage:v20200806-9940234 gcr.io/k8s-prow/triage:latest gcr.io/k8s-prow/triage:go-v20200806-9940234 gcr.io/k8s-prow/triage:go-latest  
```
- Note line 31, with `--project` set to the correct project name:
```
INFO: Running command line: bazel-bin/images/builder/linux_amd64_stripped/builder '--log-dir=/logs/artifacts' '--scratch-bucket=gs://k8s-testimages-scratch' '--project=k8s-testimages' '--build-dir=.' triage/
```

Furthermore, https://gcr.io/k8s-prow/triage doesn't exist, while https://gcr.io/k8s-testimages/triage does exist, and contains the built images.

**Anything else we need to know?**:

/cc @BenTheElder

BenTheElder:
> 2020/08/06 22:41:44 Successfully built image: gcr.io/k8s-testimages/triage:v20200806-781c64c gcr.io/k8s-testimages/triage:latest gcr.io/k8s-testimages/triage:go-v20200806-781c64c gcr.io/k8s-testimages/triage:go-latest  

https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-test-infra-push-triage/1291502926757367808

PR #18709, opened 2020-08-06
fix projectID in built image logs

BenTheElder:
fixes https://github.com/kubernetes/test-infra/issues/18708


images/builder/main.go
```diff
@@ -67,7 +67,15 @@ func getImageName(o options, tag string, config string) (string, error) {
 	if err := yaml.Unmarshal(buf, &cloudbuildyamlFile); err != nil {
 		return "", fmt.Errorf("failed to get image name: %v", err)
 	}
-	var projectID, _ = getProjectID()
```
BenTheElder, line 70: always assuming we should discover it from the current project is wrong when we have most builds run in another project ...

BenTheElder:
note: this is untested :man_shrugging:

PR #18711, opened 2020-08-06
Remove ProwMultipleContainerSupport annotation config requirement

anthonyrtong:
removing the required annotation for multiple container prowjobs that was for testing.


anthonyrtong:
just removing the old annotation for testing
/assign @cjwagner

PR #18712, opened 2020-08-06
Add kube-up/down and dump logs dependencies to KRTE

michaelmdresser:
The recently-released GCE deployer for kubetest2 will be run in KRTE containers in CI. KRTE is currently missing a couple of necessary dependencies:

The deployer calls out to `kube-up.sh`, which needs `gcloud` alpha and beta components.

The deployer also uses a log dumping script which relies on `gcloud compute ssh` and `scp`. This adds a necessary SSH client.

### DIFF
--- a/images/krte/Dockerfile
+++ b/images/krte/Dockerfile
@@ -66,6 +66,7 @@ RUN echo "Installing Packages ..." \
             kmod \
             lsb-release \
             mercurial \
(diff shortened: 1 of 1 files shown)


BenTheElder:
discussing out of band: I'd actually like to avoid this in KRTE over time, but OK with it to unblock for the moment.

PR #18714, opened 2020-08-06
Add periodic conformance test run against cloud-provider-gcp using kubetest2

michaelmdresser:
Adds a new periodic job to run the conformance ginkgo suite against the [cloud-provider-gcp](https://github.com/kubernetes/cloud-provider-gcp) repository. Test job is based on
[ci-kubernetes-gce-conformance-latest](https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/sig-cloud-provider/gcp/gce-conformance.yaml#L4), with the main difference being using the new [kubetest2](https://github.com/kubernetes-sigs/kubetest2) GCE deployer instead of kubetest. Because cloud-provider-gcp does not have published release tars, this job has to clone and build the cloud-provider-gcp repository. The job also does not include the forking and the alerting in the job it is based on because it needs to be validated in production prow before meaningful decisions can be made using it.

Additional note: due to https://github.com/kubernetes/test-infra/issues/18706, this job uses KRTE instead of kubekins-e2e as the test runner.

Tested in a GKE cluster using mkpj and mkpod.

### DIFF
--- a/config/jobs/kubernetes/cloud-provider-gcp/cloud-provider-gcp-periodics.yaml
+++ b/config/jobs/kubernetes/cloud-provider-gcp/cloud-provider-gcp-periodics.yaml
@@ -4,6 +4,7 @@ periodics:
   name: ci-cloud-provider-gcp-conformance-latest
   annotations:
     testgrid-tab-name: Conformance - Cloud Provider GCP - master
+    testgrid-dashboards: provider-gcp-periodics
     description: Runs conformance tests using kubetest2 against cloud-provider-gcp master on GCE
   labels:
     preset-service-account: "true"
--- a/config/jobs/kubernetes/cloud-provider-gcp/cloud-provider-gcp-periodics.yaml
+++ b/config/jobs/kubernetes/cloud-provider-gcp/cloud-provider-gcp-periodics.yaml
@@ -11,6 +11,13 @@ periodics:
   spec:
     containers:
     - image: gcr.io/k8s-testimages/krte:v20200806-28035c4-1.16
+      resources:
+        limits:
+          cpu: 2000m
+          memory: 2G
+        requests:
+          cpu: 2000m
+          memory: 2G
       command:
       - wrapper.sh
       args:
--- a/config/jobs/kubernetes/cloud-provider-gcp/cloud-provider-gcp-periodics.yaml
+++ b/config/jobs/kubernetes/cloud-provider-gcp/cloud-provider-gcp-periodics.yaml
@@ -1,0 +1,32 @@
+periodics:
+- interval: 3h
+  cluster: k8s-infra-prow-build
+  name: ci-cloud-provider-gcp-conformance-latest
+  annotations:
+    testgrid-tab-name: Conformance - Cloud Provider GCP - master
+    description: Runs conformance tests using kubetest2 against cloud-provider-gcp master on GCE
+  labels:
+    preset-service-account: "true"
+  spec:
+    containers:
+    - image: gcr.io/k8s-testimages/krte:v20200806-28035c4-1.16
+      command:
+      - wrapper.sh
+      args:
+      - "/bin/bash"
+      - "-c"
+      # TODO: Use published release tars for cloud-provider-gcp if/once they exist
+      - set -o errexit;
+        set -o nounset;
+        set -o pipefail;
+        set -o xtrace;
+        git clone https://github.com/kubernetes/cloud-provider-gcp;
+        cd cloud-provider-gcp;
+        REPO_ROOT=$(git rev-parse --show-toplevel);
+        cd;
+        go get -u github.com/onsi/ginkgo/ginkgo;
+        export GO111MODULE=on;
+        go get sigs.k8s.io/kubetest2@latest;
+        go get sigs.k8s.io/kubetest2/kubetest2-gce@latest;
+        go get sigs.k8s.io/kubetest2/kubetest2-tester-ginkgo@latest;
+        kubetest2 gce -v 2 --repo-root $REPO_ROOT --build --up --down --test=ginkgo -- --focus-regex='\[Conformance\]' --parallel=30
--- a/config/jobs/kubernetes/cloud-provider-gcp/cloud-provider-gcp-periodics.yaml
+++ b/config/jobs/kubernetes/cloud-provider-gcp/cloud-provider-gcp-periodics.yaml
@@ -19,11 +19,11 @@ periodics:
     - image: gcr.io/k8s-testimages/krte:v20200806-28035c4-1.16
       resources:
         limits:
-          cpu: 2000m
-          memory: 2G
+          cpu: 4
+          memory: 14Gi
         requests:
-          cpu: 2000m
-          memory: 2G
+          cpu: 4
+          memory: 14Gi
       command:
       - wrapper.sh
       args:
--- a/config/jobs/kubernetes/cloud-provider-gcp/cloud-provider-gcp-periodics.yaml
+++ b/config/jobs/kubernetes/cloud-provider-gcp/cloud-provider-gcp-periodics.yaml
@@ -2,12 +2,18 @@ periodics:
 - interval: 3h
   cluster: k8s-infra-prow-build
   name: ci-cloud-provider-gcp-conformance-latest
+  decorate: true
   annotations:
     testgrid-tab-name: Conformance - Cloud Provider GCP - master
     testgrid-dashboards: provider-gcp-periodics
     description: Runs conformance tests using kubetest2 against cloud-provider-gcp master on GCE
   labels:
     preset-service-account: "true"
+  extra_refs:
+  - org: kubernetes
+    repo: cloud-provider-gcp
+    base_ref: master
+    path_alias: k8s.io/cloud-provider-gcp
   spec:
     containers:
     - image: gcr.io/k8s-testimages/krte:v20200806-28035c4-1.16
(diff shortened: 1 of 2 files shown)


config/jobs/kubernetes/cloud-provider-gcp/cloud-provider-gcp-periodics.yaml
```diff
@@ -0,0 +1,32 @@
...
+  cluster: k8s-infra-prow-build
+  name: ci-cloud-provider-gcp-conformance-latest
+  annotations:
+    testgrid-tab-name: Conformance - Cloud Provider GCP - master
+    description: Runs conformance tests using kubetest2 against cloud-provider-gcp master on GCE
+  labels:
+    preset-service-account: "true"
+  spec:
+    containers:
+    - image: gcr.io/k8s-testimages/krte:v20200806-28035c4-1.16
```
michaelmdresser, line 12: I'm not sure which tag of this KRTE release I should be using. [This](https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/post-test-infra-push-krte/1291480780802887680) is the postsubmit that built it.

config/jobs/kubernetes/cloud-provider-gcp/cloud-provider-gcp-periodics.yaml
```diff
@@ -0,0 +1,32 @@
+periodics:
+- interval: 3h
```
michaelmdresser, line 2: Maybe this interval should be longer. Conformance tests take a while.

config/jobs/kubernetes/cloud-provider-gcp/cloud-provider-gcp-periodics.yaml
```diff
@@ -0,0 +1,32 @@
...
+        git clone https://github.com/kubernetes/cloud-provider-gcp;
+        cd cloud-provider-gcp;
+        REPO_ROOT=$(git rev-parse --show-toplevel);
+        cd;
+        go get -u github.com/onsi/ginkgo/ginkgo;
+        export GO111MODULE=on;
+        go get sigs.k8s.io/kubetest2@latest;
+        go get sigs.k8s.io/kubetest2/kubetest2-gce@latest;
+        go get sigs.k8s.io/kubetest2/kubetest2-tester-ginkgo@latest;
+        kubetest2 gce -v 2 --repo-root $REPO_ROOT --build --up --down --test=ginkgo -- --focus-regex='\[Conformance\]' --parallel=30
```
michaelmdresser, line 32: The original job doesn't configure a parallel amount for the tester. Doing so speeds up tests. I would appreciate perspective on potential issues this could cause.

config/jobs/kubernetes/cloud-provider-gcp/cloud-provider-gcp-periodics.yaml
amwat: we need your fix so this is fine, but eventually we should use the auto generated tag.

config/jobs/kubernetes/cloud-provider-gcp/cloud-provider-gcp-periodics.yaml
amwat: 3h seems fine, that's also the cadence for the existing job.

config/jobs/kubernetes/cloud-provider-gcp/cloud-provider-gcp-periodics.yaml
amwat: Not all jobs can be safely run in parallel and usually causes failures as compared to when run serially.
Let's stick to serial for now, or we can have another job for parallel.

config/jobs/kubernetes/cloud-provider-gcp/cloud-provider-gcp-periodics.yaml
```diff
@@ -0,0 +1,32 @@
+periodics:
+- interval: 3h
+  cluster: k8s-infra-prow-build
+  name: ci-cloud-provider-gcp-conformance-latest
+  annotations:
+    testgrid-tab-name: Conformance - Cloud Provider GCP - master
```
amwat, line 6: you'll need a dashboard here: https://github.com/kubernetes/test-infra/blob/master/config/testgrids/kubernetes/sig-cloud-provider/gcp/config.yaml

michaelmdresser: Added, hopefully that's the correct setup.

config/jobs/kubernetes/cloud-provider-gcp/cloud-provider-gcp-periodics.yaml
michaelmdresser: Dropped the parallel

config/jobs/kubernetes/cloud-provider-gcp/cloud-provider-gcp-periodics.yaml
michaelmdresser: Added, hopefully that's the correct syntax.

config/jobs/kubernetes/cloud-provider-gcp/cloud-provider-gcp-periodics.yaml
```diff
@@ -0,0 +1,33 @@
...
+      - wrapper.sh
+      args:
+      - "/bin/bash"
+      - "-c"
+      # TODO: Use published release tars for cloud-provider-gcp if/once they exist
+      - set -o errexit;
+        set -o nounset;
+        set -o pipefail;
+        set -o xtrace;
+        git clone https://github.com/kubernetes/cloud-provider-gcp;
```
amwat, line 24: can we use `extra_refs` instead of manually git cloning.

config/jobs/kubernetes/cloud-provider-gcp/cloud-provider-gcp-periodics.yaml
```diff
@@ -0,0 +1,33 @@
...
+        cd cloud-provider-gcp;
+        REPO_ROOT=$(git rev-parse --show-toplevel);
+        cd;
+        go get -u github.com/onsi/ginkgo/ginkgo;
```
amwat, line 28: fyi the ginkgo binary also exists in the same tarball as `e2e.test` .
Let's start with this.
and as a follow up add the logic inside the tester to use that one and then remove this from here.

michaelmdresser: Good idea

config/jobs/kubernetes/cloud-provider-gcp/cloud-provider-gcp-periodics.yaml
michaelmdresser: Periodics don't get access to `extra_refs`. I believe this is also indicated by the fact that there is no `clonerefs` containers as part of periodic job specs.

amwat: I'm not sure if that's true. `extra_refs` is a [ProwJob field](https://godoc.org/k8s.io/test-infra/prow/apis/prowjobs/v1#ProwJobSpec) so it's independent of whether it's a presubmit or a periodic.
We use it in a bunch of jobs https://github.com/kubernetes/test-infra/blob/70c45962f9a72e8c8141243612b15e3b6aaf6141/config/jobs/kubernetes-sigs/kind/kind.yaml#L5-L9

michaelmdresser: Hmm. I tried adding it to my periodic, regenerating it, and then looking for the repo and couldn't find it.

amwat: you'll need `decorate: true`

michaelmdresser: `decorate: true` solves this problem. Updated the PR to use `extra_refs` to get the `cloud-provider-gcp` repo.

amwat:
/cc @BenTheElder @cheftako 
approval for adding the dashboard

config/jobs/kubernetes/cloud-provider-gcp/cloud-provider-gcp-periodics.yaml
```diff
@@ -0,0 +1,44 @@
...
+  spec:
+    containers:
+    - image: gcr.io/k8s-testimages/krte:v20200806-28035c4-1.16
+      resources:
+        limits:
+          cpu: 2000m
+          memory: 2G
+        requests:
+          cpu: 2000m
+          memory: 2G
```
spiffxp, line 26: The closest analogue I would have for this is maybe pull-kubernetes-e2e-gce, which is currently at `cpu: 4, memory: 14Gi`.

amwat:
/lgtm
@spiffxp can you approve the dashboard changes as well?

liggitt:
it looks like this hasn't passed since it was added: https://testgrid.k8s.io/provider-gcp-periodics#Conformance%20-%20Cloud%20Provider%20GCP%20-%20master&width=5

the failing tests look exclusively connected to the new APIs added in 1.19... what configuration is this running with, and are we sure it's running against the 1.19 artifacts?

liggitt:
opened kubernetes/kubernetes#94051 to track

PR #18718, opened 2020-08-07
make pull-kubernetes-e2e-kind-ipv6 required

BenTheElder:
This job has comparable stability to the other e2e presubmits, we've made significant strides by removing problematic tests for now (in line with the ipv4 job) and making some fixes to the networking stack.

Right now is our opportunity to make it required before the PR floodgates open and we regress on IPv6 again.

> Success rate over time: 3h: 80%, 12h: 79%, 48h: 64%

https://prow.k8s.io/?job=pull-kubernetes-e2e-kind-ipv6
> Success rate over time: 3h: 67%, 12h: 74%, 48h: 68%

https://prow.k8s.io/?job=pull-kubernetes-e2e-gce
> Success rate over time: 3h: 75%, 12h: 70%, 48h: 68%

(NOTE: 3h rate is meaningless this time of night, but you can see the 12h / 48h are actually comparable to e2e-gce)

This job also takes only ~1/2 the time of GCE, and can be done with less resources*

We already run it by default on all PRs, and it is currently reporting to PRs.

### DIFF
--- a/config/jobs/kubernetes/sig-testing/kubernetes-kind-presubmits.yaml
+++ b/config/jobs/kubernetes/sig-testing/kubernetes-kind-presubmits.yaml
@@ -94,7 +94,7 @@ presubmits:
       testgrid-create-test-group: 'true'
 
   - name: pull-kubernetes-e2e-kind-ipv6
-    optional: true
+    optional: false
     always_run: true
     skip_report: false
     decorate: true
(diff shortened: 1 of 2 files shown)


BenTheElder:
/cc @liggitt @spiffxp 
FYI @aojea

BenTheElder:
Additional NOTE: This does NOT affect branches other than master. We should consider 1.19, but I'm most concerned with getting this into mainline before we open development back up.

I do not expect this to negatively impact the release.
I do expect failing to get it in before development opens back up will leave us with the ipv6 stability chicken-and-egg issues we've been wrestling with.

BenTheElder:
/hold
for feedback

aojea:
this is progress :smile: 
/lgtm

PR #18721, opened 2020-08-07
fix kind ipv6 network job regex

aojea:
disable the IPv6 that require internet connectivity, they are going
to fail until the CI infra supports IPv6.

Also, fix the regex removing `Experimental` from the feature tag,
since it was never used there

xref: https://k8s-testgrid.appspot.com/sig-network-kind#sig-network-kind,%20IPv6,%20master

Fixes: https://github.com/kubernetes/kubernetes/issues/93527


aojea:
> /hodl cancel

/hold cancel 

🙃

PR #18728, opened 2020-08-07
demote pull-kuberentes-e2e-gce-device-plugin-gpu from merge-blocking

spiffxp:
It runs a single test, and based on the last two weeks, the only time the job fails is when it runs out of GPU quota (the project used doesn't have committed GPU's).  There is likely sufficient signal being obtained from the CI equivalent to this job.

This is part of https://github.com/kubernetes/test-infra/issues/18729


spiffxp:
https://testgrid.k8s.io/presubmits-kubernetes-nonblocking - it's already on the nonblocking dashboard, not sure how it became a blocking job

BenTheElder:
the per-release-branch versions need updating too

spiffxp:
/cc @vishh @mindprince @yguo0905 @jiayingz 
FYI

spiffxp:
FYI @liggitt

liggitt:
always_run defaults to false, and optional defaults to true?

Issue #18726, opened 2020-08-07
Rewrite triage summarizer in go


spiffxp:
ref: https://github.com/kubernetes/test-infra/issues/12508#issuecomment-499731728
ref: https://github.com/kubernetes/test-infra/issues/17625

The triage summarizer is too slow for the data we're throwing at it.  It's written in python.  Let's rewrite it in go.

There's been work going on for this for a while but I realized we didn't actually have an umbrella issue to track it.  So, now we do.

/assign @michaelkolber 

/area triage
/sig testing

spiffxp:
Proposal presented at SIG Testing meeting 2020-06-30: https://docs.google.com/document/d/1Co9_0Oc5iCtSFiNNVvOe4QpysINPyW6hFHLB4t2xRyU/edit?usp=sharing

michaelkolber:
Files/jobs that will need to be cleaned up after switching over to Go triage:
- [ ] triage/*.py
- [ ] triage/update_summaries-go.sh -> triage/update_summaries.sh
  - [ ] Also in Dockerfile
  - [ ] Also in test-infra-periodics.yaml
- [ ] triage/Dockerfile-go -> triage/Dockerfile
  - [ ] Also in cloudbuild.yaml
- [ ] triage/cloudbuild.yaml:
  - [ ] Remove python builds
  - [ ] Remove `go-` from build tags
- [ ] [test-infra-periodics.yaml](https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/test-infra/test-infra-periodics.yaml):
  - [ ] `ci-test-infra-triage-go` -> `ci-test-infra-triage`
  - [ ] Correctly set `testgrid-alert-email`

BenTheElder:
https://storage.googleapis.com/k8s-gubernator/triage-go/index.html <-- the rewrite
https://storage.googleapis.com/k8s-gubernator/triage/index.html

looks like the `(N today)` disagree substantially? but at least we're seeing similar clusters

spiffxp:
I took a quick side-by-side comparison

![Screen Shot 2020-08-07 at 3 18 21 PM](https://user-images.githubusercontent.com/49258/89693993-14f55200-d8c5-11ea-9472-0595a5551d4b.png)

Two things jump out:
- The top cluster is the same failure text, but somehow we're seeing way fewer failures?
- The go rewrite has a number of "Invalid Date" entries

BenTheElder:
[...]

PR #18730, opened 2020-08-07
Use relative paths when fetching triage results

michaelkolber:
ref: #18726

Uses the bucket name when we're confident we can determine it

/cc BenTheElder spiffxp


BenTheElder:
/lgtm
/approve
/hold
need to figure out how we've been uploading this

BenTheElder:
unholding after https://github.com/kubernetes/test-infra/pull/18733

Issue #18651, opened 2020-08-04
release-blocking jobs must run in dedicated cluster: ci-kubernetes-e2e-gci-gce-ingress


spiffxp:
Part of https://github.com/kubernetes/test-infra/issues/18549

Need to figure out what special quotas "ingress-project" projects have: https://github.com/kubernetes/k8s.io/issues/1093.  Need google.com access to compare against the existing google.com projects, so can't make this a help-wanted

Once ingress-project is setup in k8s-infra-prow-builds' cluster, this should be a straightforward "set the `cluster:` field" PR

/sig testing
/sig release
/wg k8s-infra
/area jobs

spiffxp:
Looks like ingress-project actually isn't needed for the release-blocking jobs? The canary ran just fine with default quotas.  So a PR to move the release-blocking jobs, using default projects: https://github.com/kubernetes/test-infra/pull/18734

spiffxp:
Now that https://github.com/kubernetes/test-infra/pull/18734 has merged, keep an eye on the release-blocking jobs:
- https://testgrid.k8s.io/sig-release-master-blocking#gci-gce-ingress
- https://testgrid.k8s.io/sig-release-1.19-blocking#gce-cos-k8sbeta-ingress
- https://testgrid.k8s.io/sig-release-1.18-blocking#gce-cos-k8sstable1-ingress
- https://testgrid.k8s.io/sig-release-1.17-blocking#gce-cos-k8sstable2-ingress
- https://testgrid.k8s.io/sig-release-1.16-blocking#gce-cos-k8sstable3-ingress

spiffxp:
/close
I've seen runs for everything but 1.16 and anticipate that will behave the same as the others

PR #18734, opened 2020-08-07
Migrate release-blocking ingress jobs to k8s-infra-prow-build

spiffxp:
Based on https://testgrid.k8s.io/sig-testing-canaries#gci-gce-ingress it appears these jobs do just fine with the default set of quotas available in the GCP projects in the default gce-project pool

A followup change will remove the canary job once we have sufficient history/signal from the release-blocking jobs

Part of https://github.com/kubernetes/test-infra/issues/18651


spiffxp:
/cc @BenTheElder @hasheddan 
FYI @kubernetes/ci-signal

PR #18738, opened 2020-08-07
bump new container image promoter image

yodahekinsew:
Bumping a new image for the Container Image Promoter for both the pull-k8sio-cip Prow job and the pull-k8sio-cip-vuln Prow job


yodahekinsew:
@listx @prathak07

listx:
@yodahekinsew Can you update all promoter images? Just grep for the old sha and replace with the one in this PR.

PR #18739, opened 2020-08-07
add jobs to deploy k8s-infra prow-build resources

spiffxp:
One job for each cluster, but both jobs run in the k8s-infra trusted cluster, which is the only cluster that has the prow-deployer service account

part of https://github.com/kubernetes/k8s.io/issues/845


spiffxp:
/retitle add jobs to deploy k8s-infra prow-build resources

BenTheElder:
/lgtm
/approve
seems legit

PR #18751, opened 2020-08-08
Enforce release-blocking jobs must be Pod QOS Guaranteed

spiffxp:
All existing release-blocking jobs now have limits set, so fail
if any job is ever added to release-blocking that doesn't have
limits set

Part of https://github.com/kubernetes/test-infra/issues/18530


BenTheElder:
/lgtm
/approve
yesss

PR #18752, opened 2020-08-08
Update pull-kubernetes-files-remake 1.19 job to non-blocking

spiffxp:
The non-release-branch variant was changed to an optional,
run_if_changed job, but the copy that was forked for 1.19
was not. Set resource limits while we're here

Part of https://github.com/kubernetes/test-infra/issues/18592
Followup to https://github.com/kubernetes/test-infra/pull/18676


BenTheElder:
/lgtm
/approve
thanks

BenTheElder:
FYI @liggitt

PR #18604, opened 2020-08-01
Ensure pyyaml is available for python3

dims:
See https://github.com/kubernetes/test-infra/pull/18603 for some context, basically the verify job was failing because pyyaml was not available

    Pull in the reverted commit 9f4fb3d8a8874449b7f4a9bfd4d818df324077de and
    add the package for python3 pyyaml.

Signed-off-by: Davanum Srinivas <davanum@gmail.com>

PR #18754, opened 2020-08-08
Fix ingress-controller-conformance pip dependency

aledbf:
xref: https://github.com/kubernetes-sigs/ingress-controller-conformance/pull/58

### DIFF
--- a/config/jobs/kubernetes-sigs/ingress-controller-conformance/ingress-controller-conformance.yaml
+++ b/config/jobs/kubernetes-sigs/ingress-controller-conformance/ingress-controller-conformance.yaml
@@ -59,7 +59,7 @@ presubmits:
         args:
         - bash
         - -c
-        - apt update && apt-get install -y python3-pip && pip3 install reformat-gherkin && make verify-gherkin
+        - pip3 install reformat-gherkin==2.0.0 --ignore-installed PyYAML && make verify-gherkin
     annotations:
       testgrid-dashboards: sig-network-ingress-controller-conformance
       testgrid-tab-name: verify-gherkin


config/jobs/kubernetes-sigs/ingress-controller-conformance/ingress-controller-conformance.yaml
```diff
@@ -59,7 +59,7 @@ presubmits:
         args:
         - bash
         - -c
-        - apt update && apt-get install -y python3-pip && pip3 install reformat-gherkin && make verify-gherkin
+        - apt update && apt-get install -y python3-pip && pip3 install reformat-gherkin==2.0.0 --ignore-installed PyYAML && make verify-gherkin
```
hasheddan, line 62: @aledbf I think we probably broke this [here](https://github.com/kubernetes/test-infra/pull/18604) :/ I _think_ you can skip installing `python3-pip` here though now as it should be present in the image. However, it [looks like](https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/directory/pull-ingress-controller-conformance-gherkin/1292184216368320512#1:build-log.txt%3A32) installing it again will just be a no op so probably fine to keep it around :)

aledbf: done :)

aledbf: @hasheddan can you approve this change?

hasheddan: Sure! Although I'm not in approver here so you'll need someone else as well :)

config/jobs/kubernetes-sigs/ingress-controller-conformance/ingress-controller-conformance.yaml
```diff
@@ -59,7 +59,7 @@ presubmits:
         args:
         - bash
         - -c
-        - apt update && apt-get install -y python3-pip && pip3 install reformat-gherkin && make verify-gherkin
+        - pip3 install reformat-gherkin==2.0.0 --ignore-installed PyYAML && make verify-gherkin
```
BenTheElder, line 62: we really shouldn't need a workaround like this in a job, the image changes are clearly broken.

PR #18756, opened 2020-08-09
rename blobfuse repo

andyzhangx:
`blobfuse-csi-driver` --> `blob-csi-driver`

the `--test-blobfuse-csi-driver` test parameter should also be renamed, will address that in another PR.

details about repo renaming: https://github.com/kubernetes-sigs/blob-csi-driver/pull/209

/assign @feiskyer @chewong


andyzhangx:
@dims @wojtek-t @cblecker could you help approve? thanks.

PR #18759, opened 2020-08-09
Fix typo in resource requests for files-remake

odinuge:
The task fails when 1memory is used instead of memory.


odinuge:
Example of failing PR: https://github.com/kubernetes/kubernetes/pull/93834

PR #18698, opened 2020-08-06
Make pull-kubernetes-kubemark-e2e-gce-big job guaranteed Pod QOS

Evalle:
PR Summary:
- Makes pull-kubernetes-kubemark-e2e-gce-big job guaranteed Pod QOS
- Fixes https://github.com/kubernetes/test-infra/issues/18594

@kubernetes/ci-signal

PR #18762, opened 2020-08-10
[scalability][presubmits][k/k] Update resources

jprzychodzen:
/sig scalability
/cc @wojtekt-t
/cc @tosi3k 
/cc @jkaniuk 

* Add `--flush-mem-after-build` to release memory after build phase.
* Increase CPU to at least 6, to speed up building build phase
* If limit does not exists, provide the value
  * If possible, copied it from relase-blockingjobs.yaml
  * `pull-kubernetes-kubemark-e2e-gce-scale` job was manually checked (please see discussion in #18698)
  * Few rare values were extrapolated from jobs with higher node number (for optional and not running presubmits)

### DIFF
--- a/config/jobs/kubernetes/sig-scalability/sig-scalability-presubmit-jobs.yaml
+++ b/config/jobs/kubernetes/sig-scalability/sig-scalability-presubmit-jobs.yaml
@@ -29,6 +29,7 @@ presubmits:
         - --build=bazel
         - --cluster=
         - --extract=local
+        - --flush-mem-after-build=true
         - --gcp-nodes=100
         - --gcp-project-type=scalability-presubmit-project
         - --gcp-zone=us-east1-b
@@ -54,11 +55,12 @@ presubmits:
         - --use-logexporter
         image: gcr.io/k8s-testimages/kubekins-e2e:v20200807-c10eea0-master
         resources:
+          # Using 6 CPU to speed up bazel build phase (2 is enough for the test itself)
           limits:
-            cpu: 4
+            cpu: 6
             memory: "14Gi"
           requests:
-            cpu: 4
+            cpu: 6
             memory: "14Gi"
 
   - name: pull-kubernetes-e2e-gce-big-performance
@@ -88,6 +90,7 @@ presubmits:
         - --cluster=
         - --env=HEAPSTER_MACHINE_TYPE=n1-standard-4
         - --extract=local
+        - --flush-mem-after-build=true
         - --gcp-nodes=500
         - --gcp-project=k8s-presubmit-scale
         - --gcp-zone=us-east1-b
@@ -109,8 +112,13 @@ presubmits:
         - --use-logexporter
         image: gcr.io/k8s-testimages/kubekins-e2e:v20200807-c10eea0-master
         resources:
+          limits:
+            # Using 6 CPU to speed up bazel build phase (4 is enough for the test itself)
+            cpu: 6
+            memory: "16Gi"
           requests:
-            memory: "6Gi"
+            cpu: 6
+            memory: "16Gi"
 
   # This is an equivalent of the ci-kubernetes-e2e-gce-scale-correctness test
   # at 100 node scale. It's an optional presubmit to simplify testing changes
@@ -155,8 +163,12 @@ presubmits:
         - --use-logexporter
         image: gcr.io/k8s-testimages/kubekins-e2e:v20200807-c10eea0-master
         resources:
+          limits:
+            cpu: 6
+            memory: "16Gi"
           requests:
-            memory: "6Gi"
+            cpu: 6
+            memory: "16Gi"
 
   - name: pull-kubernetes-e2e-gce-large-performance
     always_run: false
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes/sig-scalability/sig-scalability-presubmit-jobs.yaml
```diff
@@ -55,10 +56,10 @@ presubmits:
         image: gcr.io/k8s-testimages/kubekins-e2e:v20200807-c10eea0-master
         resources:
           limits:
-            cpu: 4
+            cpu: 6
```
wojtek-t, line 59: Is 4 really not enough?

Do we allocate 6 because of bazel overhead? If so, might be good to add an explicit comment (here an n other places).

config/jobs/kubernetes/sig-scalability/sig-scalability-presubmit-jobs.yaml
```diff
@@ -207,8 +218,12 @@ presubmits:
         - --use-logexporter
         image: gcr.io/k8s-testimages/kubekins-e2e:v20200807-c10eea0-master
         resources:
+          limits:
+            cpu: 6
+            memory: "8Gi"
```
wojtek-t, line 223: Is 8GB enough for 2000-node presubmit?

config/jobs/kubernetes/sig-scalability/sig-scalability-presubmit-jobs.yaml
```diff
@@ -29,6 +29,7 @@ presubmits:
         - --build=bazel
         - --cluster=
         - --extract=local
+        - --flush-mem-after-build=true
```
wojtek-t, line 32: Can you also open another PR to adjust what was done in: https://github.com/kubernetes/test-infra/pull/18698 (to make everything more consistent).

jprzychodzen: As I understand you are referring to files like `config/jobs/kubernetes/sig-release/release-branch-jobs/1.19.yaml`?

Sure, after we will have a consensus here, I will 'backport' changes to release-branch-jobs. SG?

config/jobs/kubernetes/sig-scalability/sig-scalability-presubmit-jobs.yaml
jprzychodzen: Yes, it's because of bazel. We could go for lower CPU, but that would increase a duration of presubmit jobs, so probably we want to optimize for time, rather than resource consumption.

I will add comment

config/jobs/kubernetes/sig-scalability/sig-scalability-presubmit-jobs.yaml
jprzychodzen: IIRC When I was tracking Kubemark manually I've seen an usage of a little above 1 GB. Assuming that it will grow linearly, we can go for with 8GB. Also, for Kubemark 5000 nodes test we are using 16 GB, so again it suggest that we are good with half of that for 2000 nodes.

However, we can go for 'better safe than sorry' way and increase this request, especially that it's an optional presubmit that is rarely triggered. I will change to 16GB.

config/jobs/kubernetes/sig-scalability/sig-scalability-presubmit-jobs.yaml
wojtek-t: and other release branches yes

config/jobs/kubernetes/sig-scalability/sig-scalability-presubmit-jobs.yaml
```diff
@@ -54,11 +55,12 @@ presubmits:
         - --use-logexporter
         image: gcr.io/k8s-testimages/kubekins-e2e:v20200807-c10eea0-master
         resources:
+          # Using 6 CPU to speed up bazel build phase
```
wojtek-t, line 58: ... phase (4 is enough for the test itself).

Same below.

config/jobs/kubernetes/sig-scalability/sig-scalability-presubmit-jobs.yaml
```diff
@@ -277,11 +297,12 @@ presubmits:
         # value: "y"
         resources:
           limits:
+            # Using 6 CPU to speed up bazel build phase
             cpu: 6
-            memory: 24Gi
+            memory: 8Gi
```
wojtek-t, line 302: Given you set big-performance to 16, we should set it to 16 here too (or to 8 there). Those two should be consistent.

config/jobs/kubernetes/sig-scalability/sig-scalability-presubmit-jobs.yaml
jprzychodzen: Done.

config/jobs/kubernetes/sig-scalability/sig-scalability-presubmit-jobs.yaml
jprzychodzen: Done.

Issue #18770, opened 2020-08-10
Demoting jobs from kubernetes/kubernetes merge-blocking causes pull-test-infra-gubernator to fail on subsequent PRs 


spiffxp:
**What happened**:
https://github.com/kubernetes/test-infra/pull/18667 had to update gubernator/config.yaml to make pull-test-infra-gubernator pass

The PR had nothing to do with gubernator

**What you expected to happen**:
The PR(s) that demoted pull-kubernetes-files-remake and pull-kubernetes-e2e-gce-device-plugin-gpu should have failed to merge until they had updated gubernator/config.yaml

**Please provide links to example occurrences, if any**:
https://github.com/kubernetes/test-infra/pull/18667 - is the PR that caught the failures
https://github.com/kubernetes/test-infra/pull/18728 - is a PR that should have failed

**Anything else we need to know?**:
Should the gubernator job be triggering on more files than it currently is?
Why does this list of merge blocking jobs exist? Do we still need it?

spiffxp:
https://github.com/kubernetes/test-infra/issues/18386 has links to other occurrences, closing that in favor of this

spiffxp:
https://github.com/kubernetes/test-infra/pull/18448 attempted to fix this, but didn't?

another PR that ended up breaking pull-test-infra-gubernator: https://github.com/kubernetes/test-infra/pull/18421

spiffxp:
Proved that #18771 fixed the problem by opening a PR that changed a job's merge-blocking status without updating `gubernator/config.yaml`, which resulted in this test failure: https://github.com/kubernetes/test-infra/pull/18788#issuecomment-672331686

/close

PR #18771, opened 2020-08-10
Try to catch gubernator-breaking job changes at presubmit time

spiffxp:
Another attempt at fixing https://github.com/kubernetes/test-infra/issues/18770

The previous attempt at triggering the relevant tests via `bazel test //hack/...` (https://github.com/kubernetes/test-infra/pull/18448) didn't work because:
- the script was mispelled...
- ...actually bazel doesn't invoke verify-all.sh
- ...actually gubernator isn't in bazel's view of the world at all

Rather than tackle that last problem, let's just trigger the job that ends up failing more often

I won't be able to verify this as part of this PR, so look for a dummy PR to verify this post-merge

### DIFF
--- a/hack/verify-all.sh
+++ b/hack/verify-all.sh
@@ -19,4 +19,3 @@ set -o pipefail
 
 cd "$(git rev-parse --show-toplevel)"
 find hack -name 'verify-*.sh' -not -name "$(basename "$0")" \( -print -exec '{}' ';' -o -quit \)
-./guberantor/verify_config.sh
--- a/config/jobs/kubernetes/test-infra/test-infra-presubmits.yaml
+++ b/config/jobs/kubernetes/test-infra/test-infra-presubmits.yaml
@@ -25,7 +25,7 @@ presubmits:
   - name: pull-test-infra-gubernator
     branches:
     - master
-    run_if_changed: 'gubernator|config/prow/config.yaml' # TODO(fejta): |config/jobs/.+\.yaml
+    run_if_changed: 'gubernator|config/prow/config.yaml|config/jobs/.+\.yaml'
     decorate: true
     labels:
       preset-service-account: "true"


spiffxp:
/wip
TIL a single [wip] commit won't mark the PR as work in progress, good to know

BenTheElder:
> TIL a single [wip] commit won't mark the PR as work in progress, good to know

the only way a commit helps is if your PR has a single commit the title is auto-populated from the commit by default.

hack/verify-all.sh
```diff
@@ -19,4 +19,4 @@ set -o pipefail
 
 cd "$(git rev-parse --show-toplevel)"
 find hack -name 'verify-*.sh' -not -name "$(basename "$0")" \( -print -exec '{}' ';' -o -quit \)
-./guberantor/verify_config.sh
+./gubernator/verify_config.sh
```
BenTheElder, line 22: lol, how long has this been broken?

config/jobs/kubernetes/sig-testing/bazel-build-test.yaml
```diff
@@ -2,7 +2,7 @@ presubmits:
   kubernetes/kubernetes:
   - name: pull-kubernetes-bazel-build
     decorate: true
-    always_run: true
+    always_run: false
```
BenTheElder, line 5: this seems unrelated?

spiffxp: this is a change that would require I also update gubernator's config, so this PR _should_ be failing

this is how I've discovered that bazel doesn't invoke ./hack/verify-all.sh

hack/verify-all.sh
spiffxp: not long, I only introduced this within the last N weeks

turns out this won't work anyway...

chases2:
/lgtm
/hold
(For manual merging)

PR #18781, opened 2020-08-11
azure: add preset label for azure jobs that need anonymous pull

devigned:
Create a new preset label derived from `preset-azure-cred` which contains a registry with anonymous pull access enabled. This will be used initially in the Cluster API Provider Azure jobs.

/cc @chewong @CecileRobertMichon

### DIFF
--- a/config/prow/config.yaml
+++ b/config/prow/config.yaml
@@ -727,6 +727,32 @@ presets:
   - name: azure-ssh
     mountPath: /etc/azure-ssh
     readOnly: true
+- labels:
+    preset-azure-anonymous-pull: "true"
+  env:
+    - name: REGISTRY
+      value: capzci.azurecr.io
+- labels:
+    preset-azure-cred-only: "true"
+  env:
+    - name: AZURE_CREDENTIALS
+      value: /etc/azure-cred/credentials
+    - name: AZURE_SSH_PUBLIC_KEY_FILE
+      value: /etc/azure-ssh/azure-ssh-pub
+  volumes:
+    - name: azure-cred
+      secret:
+        secretName: azure-cred
+    - name: azure-ssh
+      secret:
+        secretName: azure-ssh
+  volumeMounts:
+    - name: azure-cred
+      mountPath: /etc/azure-cred
+      readOnly: true
+    - name: azure-ssh
+      mountPath: /etc/azure-ssh
+      readOnly: true
 # storage / caching presets
 - labels:
     preset-bazel-scratch-dir: "true"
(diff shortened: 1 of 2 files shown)


config/prow/config.yaml
```diff
@@ -727,6 +727,31 @@ presets:
...
+- labels:
+    preset-azure-cred-anonymous-pull: "true"
+  env:
+    - name: AZURE_CREDENTIALS
+      value: /etc/azure-cred/credentials
+    - name: AZURE_SSH_PUBLIC_KEY_FILE
+      value: /etc/azure-ssh/azure-ssh-pub
+    - name: REGISTRY
+      value: capzci.azurecr.io
+    - name: AZ_STORAGE_CONTAINER_NAME
```
CecileRobertMichon, line 739: do we need this?

devigned: I don't think so. 

@chewong is this in use anywhere in CAPZ related jobs?

devigned: I removed the storage bucket and broke out `preset-azure-anonymous-pull` and `preset-azure-cred-only` which will allow easier composition going forward.

chewong: Only `AZURE_CREDENTIALS` is used by CAPZ e2e test. We might not even need `AZURE_SSH_PUBLIC_KEY_FILE` and allow the script to generate one during test run.

config/prow/config.yaml
```diff
@@ -727,6 +727,32 @@ presets:
   - name: azure-ssh
     mountPath: /etc/azure-ssh
     readOnly: true
+- labels:
+    preset-azure-anonymous-pull: "true"
+  env:
+    - name: REGISTRY
+      value: capzci.azurecr.io
+- labels:
+    preset-azure-cred-only: "true"
```
CecileRobertMichon, line 736: is the idea to reuse cred-only for the other jobs in the future? IMO it would make more sense to either 1) have one label preset `preset-azure-cred-anonymous-pull` for capz that has both creds and registry and not have any reuse between capz and other azure jobs like you had before or 2) have one `preset-azure-cred` label that is shared amongst all the Azure jobs (capz + others) that only contains creds and then `preset-azure-acr-anonymous-pull` + `preset-azure-acr` labels for registry/storage stuff

devigned: I was shooting for 2. I thought this would be a good way to start the discussion about decomposing `preset-azure-cred`. I didn't want to do that in this PR. In this PR, I just wanted to setup the example for use in capz.

CecileRobertMichon: great, that's my preference as well

Issue #18729, opened 2020-08-07
Re-evaluate set of merge-blocking jobs for kubernetes/kubernetes


spiffxp:
Pulling this out of https://github.com/kubernetes/kubernetes/issues/92937#issuecomment-662178519

Related to but not explicitly part of https://github.com/kubernetes/test-infra/issues/18551

One thing that's come up in discussion over why kubernetes PR's are so hard to merge (https://github.com/kubernetes/kubernetes/issues/92937) is whether we really need so many jobs to run for each and every single PR that is opened against kubernetes/kubernetes.  There is a desire to see if we can trim the number of jobs down without sacrificing (too much) coverage.

Reasons for this are:
- if we assume each job has a non-zero chance of flaking, fewer jobs means fewer chances for a PR to encounter a flake
- if we assume jobs are flaking due to resource contention, fewer jobs running means more resources available for jobs to consume

/priority important-soon
/area jobs
/sig testing
/sig release

FYI @BenTheElder @liggitt @kubernetes/ci-signal

spiffxp:
Comment from @liggitt 
> I'd divide those into categories for improvement like this:

> immediate: should probably be post-submits or periodics set to notify relevant owners:
>- pull-kubernetes-dependencies-canary
>- pull-kubernetes-e2e-gce-device-plugin-gpu (could keep a presubmit with always_run: false for manual triggering/testing if we want)
>- pull-kubernetes-files-remake

>significantly overlapping tests ... can we move one to post-submit + notification:
> - pull-kubernetes-e2e-gce (cos+docker)
> - pull-kubernetes-e2e-gce-ubuntu-containerd (ubuntu+containerd ... already have containerd coverage via kind presubmits)

> pair of performance tests (can we have just one performance presubmit?):
> - pull-kubernetes-kubemark-e2e-gce-big
> - pull-kubernetes-e2e-gce-100-performance

>several tests that seem to just check ability to build... can we collapse these somehow:
>- pull-kubernetes-bazel-build
>- pull-kubernetes-typecheck
>- pull-kubernetes-cross

spiffxp:
https://github.com/kubernetes/test-infra/pull/18728 - to address pull-kubernetes-e2e-gce-device-plugin-gpu

spiffxp:
https://github.com/kubernetes/test-infra/pull/18649 - proposes demoting pull-kubernetes-e2e-gce in favor of pull-kubernetes-e2e-gce-ubuntu-containerd

[...]

PR #18788, opened 2020-08-11
Make pull-kubernetes-kubemark-e2e-gce-big manually triggered, optional

spiffxp:
part of https://github.com/kubernetes/test-infra/issues/18729

ref: https://github.com/kubernetes/test-infra/issues/18729#issuecomment-671252325

/cc @wojtek-t @mm4tt

### DIFF
--- a/gubernator/config.yaml
+++ b/gubernator/config.yaml
@@ -34,7 +34,6 @@ jobs:
   - pull-kubernetes-e2e-kind
   - pull-kubernetes-e2e-kind-ipv6
   - pull-kubernetes-integration
-  - pull-kubernetes-kubemark-e2e-gce-big
   - pull-kubernetes-node-e2e
   - pull-kubernetes-typecheck
   - pull-kubernetes-verify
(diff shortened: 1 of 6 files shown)


spiffxp:
pull-test-infra-gubernator failed!  the system works! (closing https://github.com/kubernetes/test-infra/issues/18770)

spiffxp:
FYI @BenTheElder @liggitt

gubernator/config.yaml
```diff
@@ -34,7 +34,6 @@ jobs:
   - pull-kubernetes-e2e-kind
   - pull-kubernetes-e2e-kind-ipv6
   - pull-kubernetes-integration
-  - pull-kubernetes-kubemark-e2e-gce-big
```
wojtek-t, line 37: Given the job is not being removed, just no longer required, shouldn't we keep this as is?

spiffxp: The recommended way to navigate to your job's logs would be to use spyglass.  Gubernator's list is generated based on `always_run` jobs.  We have basically stopped supporting gubernator at this point.

PR #18791, opened 2020-08-11
Use multi os hyperv aks-engine package

jsturtevant:
The current aks-engine package used for hyperv does not support server 2004 and the testgrid job is failing: https://testgrid.k8s.io/sig-windows-containerd#aks-engine-azure-windows-2004-master-containerd-hyperv

This unblocks the test failure with a ask-engine package that enables multiple OS based on https://github.com/Azure/aks-engine/pull/3688 until we are able to get merge and get a new aks-engine release.

/cc @chewong @marosset @adelina-t 

/sig windows


jsturtevant:
The short url would be just for this subset of tests and will eventually make it to the main release.  This is a temporary fix until we have the work merged and released in aks-engine at that time we can move to short url https://aka.ms/aks-engine/aks-engine-k8s-e2e.tar.gz that is used for the rest of the jobs.

PR #18794, opened 2020-08-12
[gcsweb] Add the ability to use an OAuth token to access private buckets

droslean:
Currently, the `gcsweb` tool can access only public buckets. In order to give the ability to access also private buckets, GCS already supports using OAuth token in the API calls. ref https://cloud.google.com/storage/docs/authentication 

With this change, the user will be able to provide the OAuth token in a file using the flag `--oauth-token-file`. Then all the API calls will use the `"Authorization: Bearer OAUTH2_TOKEN"` header as well.

This PR:

- Refactor option flags
- Improving the logging
- Include the OAuth token file support

### DIFF
--- a/gcsweb/cmd/gcsweb/gcsweb.go
+++ b/gcsweb/cmd/gcsweb/gcsweb.go
@@ -31,24 +31,21 @@ import (
 	"strings"
 	"time"
 
+	"github.com/sirupsen/logrus"
+
 	"k8s.io/test-infra/gcsweb/pkg/version"
 )
 
-// The base URL for GCS's HTTP API.
-const gcsBaseURL = "https://storage.googleapis.com"
-const gcsPath = "/gcs" // path for GCS browsing on this server
-
-// The base URL for GCP's GCS browser.
-const gcsBrowserURL = "https://console.cloud.google.com/storage/browser"
-
-var flPort = flag.Int("p", 8080, "port number on which to listen")
-var flIcons = flag.String("i", "/icons", "path to the icons directory")
-var flStyles = flag.String("s", "/styles", "path to the styles directory")
-var flVersion = flag.Bool("version", false, "print version and exit")
-var flUpgradeProxiedHTTPtoHTTPS = flag.Bool("upgrade-proxied-http-to-https", false,
-	"upgrade any proxied request (e.g. from GCLB) from http to https")
-
 const (
+	// The base URL for GCS's HTTP API.
+	gcsBaseURL = "https://storage.googleapis.com"
+
+	// path for GCS browsing on this server
+	gcsPath = "/gcs"
+
+	// The base URL for GCP's GCS browser.
+	gcsBrowserURL = "https://console.cloud.google.com/storage/browser"
+
 	iconFile = "/icons/file.png"
 	iconDir  = "/icons/dir.png"
 	iconBack = "/icons/back.png"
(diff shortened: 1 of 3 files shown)


petr-muller:
Can you provide description of how this is supposed to work? What OAuth token is the user (the one who deploys gcsweb) supposed to provide to access a private bucket?

gcsweb/cmd/gcsweb/gcsweb.go
```diff
@@ -67,26 +67,94 @@ func (ss *strslice) Set(value string) error {
...
+	}
+
+	if o.oauthTokenFile != "" {
+		secretAgent := &secret.Agent{}
+		if err := secretAgent.Start([]string{o.oauthTokenFile}); err != nil {
+			logrus.WithError(err).Fatal("Error starting secrets agent.")
+		}
+
+		oauthTokenFileGenerator := secretAgent.GetTokenGenerator(o.oauthTokenFile)
+		logrus.SetFormatter(logrusutil.NewCensoringFormatter(logrus.StandardLogger().Formatter, func() sets.String {
```
stevekuznetsov, line 143: Why? `secretAgent.Start()` does this.

gcsweb/cmd/gcsweb/gcsweb.go
```diff
@@ -216,17 +292,25 @@ func gcsRequest(w http.ResponseWriter, r *http.Request) {
 	if markers, found := r.URL.Query()["marker"]; found {
 		url += "&marker=" + markers[0]
 	}
+	urlLogger := logger.WithField("url", url)
+
+	// Create a new request using http
+	req, err := http.NewRequest("GET", url, nil)
```
stevekuznetsov, line 298: `s/"GET"/http.MethodGet/`

gcsweb/cmd/gcsweb/gcsweb.go
```diff
@@ -216,17 +292,25 @@ func gcsRequest(w http.ResponseWriter, r *http.Request) {
 	if markers, found := r.URL.Query()["marker"]; found {
 		url += "&marker=" + markers[0]
 	}
+	urlLogger := logger.WithField("url", url)
```
stevekuznetsov, line 295: more useful:

```go
logger = logger.WithFields(logrus.Fields{
  "bucket": bucket,
  "object": object,
  "markers": markers,
}
```

gcsweb/cmd/gcsweb/gcsweb.go
```diff
@@ -216,17 +292,25 @@ func gcsRequest(w http.ResponseWriter, r *http.Request) {
...
+
+	if s.tokenGenerator != nil {
+		req.Header.Add("Authorization", fmt.Sprintf("Bearer %s", string(s.tokenGenerator())))
+	}
 
-	resp, err := http.Get(url)
+	resp, err := s.httpClient.Do(req)
 	if err != nil {
-		logger.Printf("GET %s: %s", url, err)
+		urlLogger.WithError(err).Error("error occurred")
```
stevekuznetsov, line 306: `s/error occurred/failed to GET from GCS/`

gcsweb/cmd/gcsweb/gcsweb.go
```diff
@@ -216,17 +292,25 @@ func gcsRequest(w http.ResponseWriter, r *http.Request) {
...
 		w.WriteHeader(http.StatusInternalServerError)
 		fmt.Fprintf(w, "http.Get: %v", err)
 		return
 	}
 	defer resp.Body.Close()
 
-	logger.Printf("GET %s: %s", url, resp.Status)
+	urlLogger.WithField("status", resp.Status).Info("URL processed")
```
stevekuznetsov, line 313: Don't change these error messages?

gcsweb/cmd/gcsweb/gcsweb.go
```diff
@@ -67,26 +67,94 @@ func (ss *strslice) Set(value string) error {
...
+	flStyles       string
+	oauthTokenFile string
+
+	flVersion bool
+
+	// Only buckets in this list will be served.
+	allowedBuckets strslice
+}
+
+var flUpgradeProxiedHTTPtoHTTPS bool
```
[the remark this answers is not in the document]
droslean: I left this as global like before, because there are many funcs that use this variable in some checks, and I didn't want to turn them all to methods and pass this variable around.

gcsweb/cmd/gcsweb/gcsweb.go
droslean: Well, I didn't change the message because all the information from the old message is now in the logger's field.

PR #18795, opened 2020-08-12
Make some small fixes to Triage

michaelkolber:
ref: #18726

As I come across small mistakes, I'll add them to this PR


triage/summarize/files.go
```diff
@@ -145,12 +145,14 @@ string, no message is printed.
 */
 func memoizeResults(filepath string, message string, v interface{}) {
```
MushuEE, line 146: nit: (filepath, message string, v interface{})

triage/summarize/files.go
```diff
@@ -143,14 +143,16 @@ prints a warning if the results could not be memoized.
...
-		logInfo("Done " + message)
+	if err != nil {
+		logWarning("Could not memoize results to '%s': %s", filepath, err)
 		return
 	}
 
-	logWarning("Could not memoize results to '%s': %s", filepath, err)
+	if message != "" {
+		logInfo("Done " + message)
+	}
```
michaelkolber, line 155: This change will eventually need to use klog, so it should only be merged after #18858

michaelkolber:
Use klog for the logging fix

PR #18796, opened 2020-08-12
Add CAPI job to verify external links

benmoss:
The other part of https://github.com/kubernetes-sigs/cluster-api/pull/3474. With that PR we won't be doing link verification in `make verify`, so adding this as an optional additional PR presubmit.

@vincepri @fabriziopandini

### DIFF
--- a/config/jobs/kubernetes-sigs/cluster-api/cluster-api-presubmits.yaml
+++ b/config/jobs/kubernetes-sigs/cluster-api/cluster-api-presubmits.yaml
@@ -81,6 +81,25 @@ presubmits:
     annotations:
       testgrid-dashboards: sig-cluster-lifecycle-cluster-api
       testgrid-tab-name: pr-verify
+  - name: pull-cluster-api-verify-external-links
+    optional: true
+    decorate: true
+    path_alias: sigs.k8s.io/cluster-api
+    always_run: true
+    labels:
+      preset-service-account: "true"
+    skip_branches:
+    - gh-pages
+    spec:
+      containers:
+      - image: gcr.io/k8s-testimages/kubekins-e2e:v20200802-ff98b58-1.18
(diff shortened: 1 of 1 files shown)


vincepri:
/hold

Until we discuss and merge the other PR next week, want to make sure we don't merge too early

benmoss:
/hold cancel
@ncdc @detiber this is how I anticipated we'd get feedback on broken links

detiber:
@benmoss it might be good to also add a periodic job in case any existing links break.

benmoss:
Just updated the make target to `verify-book-links`

benmoss:
@detiber i think with this we'll get "periodic" coverage in that PRs will all have a failing job until one merges to fix the broken link. It's not ideal but it'll also help surface them better than a dashboard on testgrid would. We could do periodic as well but my fear would be that it'd go unnoticed.

config/jobs/kubernetes-sigs/cluster-api/cluster-api-presubmits.yaml
```diff
@@ -81,6 +81,28 @@ presubmits:
...
+    spec:
+      containers:
+      - image: gcr.io/k8s-testimages/kubekins-e2e:v20200802-ff98b58-1.18
+        command:
+        - "runner.sh"
+        - "make"
+        - "verify-book-links"
+        resources:
+          requests:
+            cpu: 7300m
```
ncdc, line 102: Do we need this much CPU for this job?

benmoss: I have no idea! Probably not, I just copied it from the other adjacent jobs. I see `apidiff` has no resource requests, so I'll drop it here and see what happens.

PR #18797, opened 2020-08-12
Enforce presence of jobs in kubernetes-presubmits-blocking testgrid

spiffxp:
https://testgrid.k8s.io/presubmits-kubernetes-blocking has been maintained by hand, which has allowed it to drift from the set of jobs that are actually merge-blocking today

I noticed this since we've been changing that set of jobs lately (ref: https://github.com/kubernetes/test-infra/issues/18729)


spiffxp:
FYI @kubernetes/ci-signal

PR #18800, opened 2020-08-12
Fix post-test-infra-upload-triage

michaelkolber:
ref: #18726


michaelkolber:
/uncc dims wojtek-t
/cc BenTheElder spiffxp
Sorry for the mixup!

Issue #18808, opened 2020-08-12
ci-kubernetes-build jobs may upload incomplete set of artfacts


spiffxp:
**What happened**:
ci-kubernetes-build jobs that timeout may upload an incomplete set of artifacts, and subsequent runs against the same commit don't rebuild / push the rest of the artifacts

The two timeouts on the left are examples of this
![](https://user-images.githubusercontent.com/49258/90074249-a5f66f80-dcaf-11ea-816e-e5442f8d29e0.png)

**What you expected to happen**:
I expect if ci-kubernetes-build (and its release-branch variants) times out or uploads an incomplete set of artifacts, the next run of the job will rebuild / push the rest of the artifacts

**How to reproduce it (as minimally and precisely as possible)**:
Set the timeout for a ci-kubernetes-build job low enough that it times out during artifact upload

**Please provide links to example occurrences, if any**:

https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-build-stable1/1291744141339791362 (`repo-commit 236b9e7fcda25d9b28afa81305ab23f98e622461`)
```
I0807 17:06:38.191] +++ [0807 17:06:38] Waiting on test tarballs
I0807 17:06:38.196] +++ [0807 17:06:38] Starting tarball: test linux-arm64
I0807 17:06:38.201] +++ [0807 17:06:38] Starting tarball: test darwin-amd64
I0807 17:06:38.203] +++ [0807 17:06:38] Starting tarball: test windows-amd64
I0807 17:08:37.091] +++ [0807 17:08:37] Building tarball: test portable
W0807 17:08:37.720] Run: ('../release/push-build.sh', '--nomock', '--verbose', '--ci', '--release-kind=kubernetes', '--docker-registry=gcr.io/kubernetes-ci-images', '--extra-publish-file=k8s-stable1', '--allow-dup')
I0807 17:33:29.168] Terminate 485 on timeout
```
It's not clear what did or did not get pushed

https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-build-stable1/1291191706741379076
```
W0806 01:59:55.439] Run: ('gsutil', 'ls', 'gs://kubernetes-release-dev/ci/v1.18.7-rc.0.36+2a3a36842f8ab9')
[...]

PR #18809, opened 2020-08-12
Raise ci-kubernetes-build timeout to 240min

spiffxp:
This is kicking the can down the road. I'm seeing timeouts at 180min
that may not have completed upload of all release artifacts. The next
run only checks a few artifacts before deciding the build's already
been done.

So, let's hopefully encounter this less, and tune back down when we
have a higher fidelity "did it already build?" check

ref: https://github.com/kubernetes/test-infra/issues/18808


spiffxp:

PR #18810, opened 2020-08-13
refresh some owners

BenTheElder:
signing up my team for more reviews / approves for things we work on after discussing with them, and cleaning up some stale entries

- @spiffxp and I are current root approvers, and none of these are no_parent_owners type things...
- @amwat is an emeritus root approver
- @MushuEE is neither, only signing up for reviews, except for kettle, which he's been reviewing / working on, bumping up to approver

- @cjwagner reached out about scenarios/ specifically, it's deprecated but needs nominal maintainership until kubernetes no longer depends on it. shifting the load on that to kubernetes engprod, following the points above

- @Katharine and @ixdy are working on other things, rotating them to emeritus for any files I touched along the way

/cc @amwat @spiffxp @MushuEE @cjwagner @Katharine @ixdy

### DIFF
--- a/images/builder/OWNERS
+++ b/images/builder/OWNERS
@@ -1,0 +1,6 @@
+# See the OWNERS docs at https://go.k8s.io/owners
+
+reviewers:
+- test-infra-oncall
+approvers:
+- test-infra-oncall
--- a/kettle/OWNERS
+++ b/kettle/OWNERS
@@ -1,12 +1,17 @@
 # See the OWNERS docs at https://go.k8s.io/owners
 
 reviewers:
-- Katharine
 - spiffxp
 - MushuEE
+- amwat
+- BenTheElder
 approvers:
-- Katharine
 - spiffxp
+- MushuEE
+- amwat
+- BenTheElder
+emeritus_approvers:
+- Katharine
 
 labels:
 - area/kettle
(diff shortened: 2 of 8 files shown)


images/builder/OWNERS
```diff
@@ -0,0 +1,6 @@
+# See the OWNERS docs at https://go.k8s.io/owners
+
+reviewers:
+- test-infra-oncall
+approvers:
+- test-infra-oncall
```
spiffxp, line 6: wg-k8s-infra-oncall should be here too

BenTheElder, line 6: oh also clarified that test-infra-oncall should own this for now, as this tooling is expected to be trusted in the trusted cluster

kettle/OWNERS
```diff
@@ -1,12 +1,17 @@
...
 reviewers:
-- Katharine
 - spiffxp
 - MushuEE
+- amwat
+- BenTheElder
 approvers:
-- Katharine
 - spiffxp
+- MushuEE
```
spiffxp, line 10: 💯 for approver

images/builder/OWNERS
BenTheElder: i'm not sure on this, can we do this in a follow-up?

spiffxp: sure

PR #18823, opened 2020-08-13
spyglass: buildlog: allow hiding the raw log link

stevekuznetsov:
There's nothing more annoying that getting a raw GCS txt link from a
user instead of the full Spyglass page.

/assign @alvaroaleman

### DIFF
--- a/prow/spyglass/lenses/buildlog/lens.go
+++ b/prow/spyglass/lenses/buildlog/lens.go
@@ -44,6 +44,12 @@ const (
 
 type config struct {
 	HighlightRegexes []string `json:"highlight_regexes"`
+	HideRawLog       bool     `json:"hide_raw_log,omitempty"`
+}
+
+type parsedConfig struct {
+	highlightRegex *regexp.Regexp
+	showRawLog     bool
 }
 
 var _ api.Lens = Lens{}
@@ -62,7 +68,8 @@ func (lens Lens) Config() lenses.LensConfig {
 
 // Header executes the "header" section of the template.
 func (lens Lens) Header(artifacts []api.Artifact, resourceDir string, config json.RawMessage) string {
-	return executeTemplate(resourceDir, "header", BuildLogsView{})
+	conf := getConfig(config)
+	return executeTemplate(resourceDir, "header", BuildLogsView{ShowRawLog: conf.showRawLog})
 }
 
 // defaultErrRE matches keywords and glog error messages.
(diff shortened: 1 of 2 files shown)


alvaroaleman:
May be but this in turn is super annoying for logs that are big enough to kill the users browser when displayed in spyglass

BenTheElder:
I actually just had a conversation with someone wanting the raw link without the page taking so long to load /shrug
As long as it's still available on prow.k8s.io, I don't have a strong opinion on enabling other instances to change this.

aside: We should update the reviewers here.

prow/spyglass/lenses/buildlog/lens.go
```diff
@@ -124,29 +131,37 @@ type BuildLogsView struct {
 	LogViews           []LogArtifactView
 	RawGetAllRequests  map[string]string
 	RawGetMoreRequests map[string]string
```
cjwagner, line 133: I know you didn't add these, but it looks like they are unused. Can we remove them?

prow/spyglass/lenses/buildlog/template.html
```diff
@@ -7,7 +7,7 @@
 {{range $log := .LogViews}}
   <div>
     <button class="show-all-button" data-artifact="{{$log.ArtifactName}}">Show all hidden lines</button>
-    <a href="{{$log.ArtifactLink}}" style="padding-left:15px;">Raw {{$log.ArtifactName}}<i class="material-icons" style="font-size: 1em; vertical-align: middle; padding-left: 3px;">open_in_new</i></a>
+    {{if .ShowRaw}}<a href="{{$log.ArtifactLink}}" style="padding-left:15px;">Raw {{$log.ArtifactName}}<i class="material-icons" style="font-size: 1em; vertical-align: middle; padding-left: 3px;">open_in_new</i></a>{{end}}
```
cjwagner, line 10:
```suggestion
    {{if .ShowRawLog}}<a href="{{$log.ArtifactLink}}" style="padding-left:15px;">Raw {{$log.ArtifactName}}<i class="material-icons" style="font-size: 1em; vertical-align: middle; padding-left: 3px;">open_in_new</i></a>{{end}}
```

stevekuznetsov:
@BenTheElder I'd argue then we should make the view better for large logs. Having a raw link and not being able to get back to the job page to look at artifacts, trends in the job, anything ... it's pretty terrible UX and I think we can do better.

cjwagner:
> @BenTheElder I'd argue then we should make the view better for large logs. Having a raw link and not being able to get back to the job page to look at artifacts, trends in the job, anything ... it's pretty terrible UX and I think we can do better.

That sounds like the best option to me, but we probably shouldn't remove the raw log link across the board until spyglass can display large logs reasonably. Lets make it opt out in the meantime (like this PR does).

stevekuznetsov:
@cjwagner sgtm! Updated the PR to address your comments

BenTheElder:
> I'd argue then we should make the view better for large logs. 

Sure, until that exists there are reasons to go straight to the log file though.

I know some users are downloading the artifacts to explore locally as well, often starting with just the build log (AIUI).

I intermittently do these myself. Spyglass is neat, but just grabbing the text log is often faster and gives full text search.

> Having a raw link and not being able to get back to the job page to look at artifacts, trends in the job, anything ... 

The browser provides a back button, history etc ...?

PR #18824, opened 2020-08-13
describe node-kubelet orphans job

MHBauer:
See 3815dfe478695d58c82e84c4e8097ca29ee58df7 in k/k#64130 for classification PR.

Node-exclusive tags described in:
https://docs.google.com/document/d/1BdNVUGtYO6NDx10x_fueRh_DLT-SVdlPC_SsXjYCHOE/edit#

It's not clear what this job is for without looking back about 2 years in history. This is my attempt at adding context. 

We should make a goal of our node test/ci to classify these jobs and get rid of this bucket that is a miscellaneous grab bag.
/cc @vpickard @SergeyKanzhelev


MHBauer:
Example output now, 
Screen Shot 2020-08-13 at 13 18 42
working on that failing test in kubernetes/kubernetes/pull/93868 if any of y'all want to follow along.

Issue #18812, opened 2020-08-13
Migrate merge-blocking jobs to dedicated cluster: pull-kubernetes-e2e-kind


spiffxp:
Part of https://github.com/kubernetes/test-infra/issues/18550

This is a trial balloon of a specific type of job (kind)

I'd like to get a sense of what sort of scaling jobs of this type are going to cause k8s-infra-prow-build.  I'd also like to see whether this interferes with existing release-blocking job load.  If so, we should consider a more boilerplate-filled approach of setting up a separate nodepool, and using taints/tolerations/nodeselectors for merge-blocking vs. release-blocking jobs.

I'm choosing pull-kubernetes-e2e-kind because I've seen it hit scheduling errors due to lack of resources lately.  Getting it over to k8s-infra-prow-build will allow more community visibility, and ideally more resources thanks to autoscaling.

/area jobs
/wg k8s-infra
/sig testing
/sig release

spiffxp:
/assign @ameukam 
who opened https://github.com/kubernetes/test-infra/pull/18817

spiffxp:
Change merged at 2020-08-13 10:24am PT and then again at 12:17pm PT

Things to watch
- https://prow.k8s.io/?job=pull-kubernetes-e2e-kind
- https://testgrid.k8s.io/presubmits-kubernetes-blocking#pull-kubernetes-e2e-kind&graph-metrics=test-duration-minutes
- https://storage.googleapis.com/k8s-gubernator/triage/index.html?pr=1&job=pull-kubernetes-e2e-kind
- https://prow.k8s.io/job-history/gs/kubernetes-jenkins/pr-logs/directory/pull-kubernetes-e2e-kind
- [prow-build dashboard 1w](https://console.cloud.google.com/monitoring/dashboards/custom/10925237040785467832?project=k8s-infra-prow-build&timeDomain=1w)
- [prowjobs-experiment 1w](https://console.cloud.google.com/monitoring/dashboards/custom/10925237040785467832?project=k8s-infra-prow-build&timeDomain=1w)
[...]

PR #18825, opened 2020-08-13
Migrate pull-kubernetes-e2e-kind-ipv6 to k8s-infra-prow-build

spiffxp:
I'm also dropping its cpu down from 7400m to 4

The 7400m was an attempt to reserve an entire node on k8s-prow since
other jobs weren't declaring their resource usage up front. Since we're
moving to k8s-infra-prow-build which requires resource limits, I expect
this to be less of a problem.

I've also noticed that pull-kuberentes-conformance-kind-ga-only-parallel
has been running with `cpu: 4` pretty happily. The bulk of the CPU usage
is for building kubernetes.

The ci- variants of these jobs use 7300m CPU, I'm inclined to bump them
down once we see how these behave

ref: https://github.com/kubernetes/test-infra/issues/18812

/cc @BenTheElder

### DIFF
--- a/config/jobs/kubernetes/sig-testing/kubernetes-kind-presubmits.yaml
+++ b/config/jobs/kubernetes/sig-testing/kubernetes-kind-presubmits.yaml
@@ -39,10 +39,10 @@ presubmits:
           privileged: true
         resources:
           limits:
-            cpu: 7400m
+            cpu: 4
             memory: 9000Mi
           requests:
-            cpu: 7400m
+            cpu: 4
             memory: 9000Mi
     annotations:
       testgrid-num-failures-to-alert: '10'
@@ -52,6 +52,7 @@ presubmits:
 
 
   - name: pull-kubernetes-e2e-kind-canary
+    cluster: k8s-infra-prow-build
     optional: true
     always_run: false # manual for testing new configurations
     decorate: true
@@ -89,15 +90,19 @@ presubmits:
         securityContext:
           privileged: true
         resources:
+          limits:
+            cpu: 4
+            memory: 9Gi
           requests:
-            memory: "9000Mi"
-            cpu: 7400m
+            cpu: 4
+            memory: 9Gi
     annotations:
       testgrid-num-failures-to-alert: '10'
       testgrid-alert-stale-results-hours: '24'
       testgrid-create-test-group: 'true'
 
   - name: pull-kubernetes-e2e-kind-ipv6
+    cluster: k8s-infra-prow-build
     optional: false
     always_run: true
     skip_report: false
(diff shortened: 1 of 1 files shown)


spiffxp:
/hold
I'm going to rebase once https://github.com/kubernetes/test-infra/pull/18817 merges

config/jobs/kubernetes/sig-testing/kubernetes-kind-presubmits.yaml
```diff
@@ -39,10 +39,10 @@ presubmits:
           privileged: true
         resources:
           limits:
-            cpu: 7400m
+            cpu: 4
```
spiffxp, line 42: This brings the main presubmit in line with the release-branch jobs

config/jobs/kubernetes/sig-testing/kubernetes-kind-presubmits.yaml
```diff
@@ -52,6 +52,7 @@ presubmits:
 
 
   - name: pull-kubernetes-e2e-kind-canary
+    cluster: k8s-infra-prow-build
```
spiffxp, line 55: Make the canary run in the same place w/ same limits as pull-kubernetes-e2e-kind

BenTheElder:
How much slower is building with only 4 CPU?
What are we giving GCE?

GCE also gets the boskos rented temporary nodes, which add CPU, I'd like to avoid complaints from test owners that kind has less CPU than GCE, which is an entirely artificial problem.

spiffxp:
>What are we giving GCE?

pull-kubernetes-e2e-gce is `cpu: 4` at present

>How much slower is building with only 4 CPU?

Since these tests don't break out build into their own phase, let's use pull-kubernetes-conformance-kind-ga-only-parallel as an example
- https://github.com/kubernetes/test-infra/pull/18671 set pull-kubernetes-conformance-kind-ga-only-parallel's limit to `cpu: 4` (previously it requested `cpu: 2`).  
- It landed 2020-08-05 11:15am PT
- https://testgrid.k8s.io/presubmits-kubernetes-blocking#pull-kubernetes-conformance-kind-ga-only-parallel&graph-metrics=test-duration-minutes&width=5 shows no significant change in Overall duration

>I'd like to avoid complaints from test owners that kind has less CPU than GCE, which is an entirely artificial problem.

I also want to get the tuning right, and would like to avoid having to spin up a new build cluster node for each kind job if we can help it.

BenTheElder:
> I also want to get the tuning right, and would like to avoid having to spin up a new build cluster node for each kind job if we can help it.

I guess, but we literally spin up N nodes for e2e GCE, and these are harder to manage correctly.

BenTheElder:
/lgtm
/approve
I think we're going to need to bump this in the future.

BenTheElder:
failures like https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/93980/pull-kubernetes-e2e-kind-ipv6/1294016707383267329 seem to have spiked. I wasn't seeing these frequently before

also in the other job https://github.com/kubernetes/kubernetes/pull/93888#issuecomment-673535234

BenTheElder:
not sure if related.

Issue #18826, opened 2020-08-13
testgrid tab groups for sig-node informing/blocking jobs


MHBauer:
`Problem we are solving: if a person is not a part of SIG node - can I ask quickly whether things are OK now or not?`

**What should be cleaned up or changed**:
Add a testgrid tab-group config consisting of existing jobs used for determining whether sig-node tests are in good condition.

We want two new top-level testgrid tab groups:
 1. one for informing (maybe pull/pr/presubmits?)
 2. one to match blocking (ci/periodics?)
   - master
   - appropriate release versioned

Here are some examples from existing testgrid groups,

for pr presubmits: 
https://testgrid.k8s.io/presubmits-kubernetes-nonblocking

for blocking:
https://testgrid.k8s.io/presubmits-kubernetes-blocking
https://testgrid.k8s.io/presubmits-kubernetes-blocking#pull-kubernetes-node-e2e

--
These are the same job showing up multiple test groups, as an example of that this can be done
https://testgrid.k8s.io/sig-release-master-blocking
https://testgrid.k8s.io/sig-release-master-blocking#node-kubelet-master
https://testgrid.k8s.io/sig-node-kubelet#node-kubelet-master

--
https://testgrid.k8s.io/sig-release-1.19-blocking
https://testgrid.k8s.io/sig-release-1.19-blocking#node-kubelet-1.19
+ other versioned releases

**Provide any links for context**:
https://docs.google.com/document/d/1fb-ugvgdSVIkkuJ388_nhp2pBTy_4HEVg5848Xy7n5U/edit#
Go to 2020-08-10 meeting notes, look for
`Problem we are solving: if a person is not a part of SIG node - can I ask quickly whether things are OK now or not?`

/cc @vpickard @SergeyKanzhelev @dims

alejandrox1:
> We want two new top-level testgrid tab groups:
>
> 1. one for informing (maybe pull/pr/presubmits?)
> 2. one to match blocking (ci/periodics?)

I think it would be better to organize jobs by whether they should be used to gauge the quality of kubernetes/kubelet, i.e., a blocking job would mean that the job is maintained and if a failure occurs there then something must be fixed.
For anon-blocking job, it could be that it is a "work in progress" and a job failure (or a test failure) does not reflect on the quality of kubernetes/kubelet/etc.

MHBauer:
I think I confused the issue by mentioning periodics and pulls. I should not have done that.

[...]

PR #18831, opened 2020-08-13
new sig-node testgrid tab-group sig-node-blocking

MHBauer:
add development and release blocking jobs

first crack at  #18826

It seems to me there are two ways to add a job report to a tab-group.
1. by adding all of the jobs to config.yaml
2. by annotating individual jobs with the dashboard they should belong to.

I don't know which is better/more-supported/suggested.

Goal is to be one stop shop for `if a person is not a part of SIG node - can I ask quickly whether things are OK now or not?`

Questions:
Is 'sig-node-blocking' understandable enough as a name? 
Did I do this correctly? 
How do I communicate to release that this dashboard exists in the future?
Should this group contain periodics and pull request jobs?
What else needs to go in here?

/cc @kubernetes/sig-node-test-failures
/cc @dims


MHBauer:
Hmn, maybe it's best to limit changes to the one config.yaml so as to keep the number of people with permissions to a minimum.

alejandrox1:
/assign @SergeyKanzhelev 
ptal

config/jobs/kubernetes/sig-node/node-kubelet.yaml
```diff
@@ -5,7 +5,7 @@ periodics:
   annotations:
     fork-per-release: "true"
     fork-per-release-periodic-interval: 1h 2h 6h 24h
-    testgrid-dashboards: sig-release-master-blocking, sig-node-kubelet
+    testgrid-dashboards: sig-release-master-blocking, sig-node-kubelet, sig-node-blocking
```
alejandrox1, line 8: what about putting it in "sig-node-kubelet-blocking"?
thatway each area(kubelet, containerd, cri, etc) can have a blocking and a "nonblocking" (jobs that are maintained and working as intended versus jobs that need attention).

MHBauer:
This is a good discussion to have. I think we need to balance 'single dashboard of all the things' vs 'dashboard explosion'.

I can see that maybe we have a cube with axes of versions, area, blocking/informing. In that, the only one I wouldn't collapse would be blocking/informing, those must always stay separate. There may even be a fourth axis, of run time, such as Periodic vs Pull/Merge.

I don't think it would be a terrible thing to have a summary board that duplicates release-blocking lists, as we want (As node) to have our own "node blocking" list of tasks.

Given the areas criterion, We've got them split up now as a combination of blockign and non-blocking, which creates some fatigue when looking at the top level as some are always red, providing no signal to act on at a glance.

So I would consider two things, unless we can get one  more level of indirection on tabs, which seems like a bad idea immediately to me.

1. split up an existing area into into blocking and informing
2. create a top-level blocking that acts as signal for 'everything node' at a glance. 

I think I was going more after 2, to start, but 1 is more what others were thinking?

SergeyKanzhelev:
The one stop shop for entire node would be ideal. The per-area tests as they are today working fine to narrow down specific feature that is being tested.

@MHBauer I really like the approach you took marking stable tests as a blocking for now.

Perhaps the plan can be:

- Mark green and blocking as blocking
- Mark flacky/failed, but desired to be blocking as WIP
  - Work on fixing those
- Do we need a single pane of glass view for "other"?

config/testgrids/kubernetes/sig-node/config.yaml
```diff
@@ -12,9 +12,11 @@ dashboard_groups:
     - sig-node-node-feature-discovery
     - sig-node-node-problem-detector
     - sig-node-seccomp-operator
+    - sig-node-blocking
```
SergeyKanzhelev, line 15: Can we make it first?

config/testgrids/kubernetes/sig-node/config.yaml
```diff
@@ -12,9 +12,11 @@ dashboard_groups:
...
 
 dashboards:
 - name: sig-node-cadvisor
+- name: sig-node-blocking # tabs included in this group are defined in the jobs that are included
```
SergeyKanzhelev, line 19: it would be so nice to have a way to add a description to the tab to show in Summary view. Do you know if this is supported?

SergeyKanzhelev:
/lgtm
with the minor suggestions

config/testgrids/kubernetes/sig-node/config.yaml
MHBauer: huh, I guess so. I didn't realize the order represented the actual order. lol

config/testgrids/kubernetes/sig-node/config.yaml
MHBauer: https://github.com/GoogleCloudPlatform/testgrid/blob/96bf6f1712d54474d69f483ccbe49fa4f9ae0204/pb/config/config.proto#L539-L540

Maybe? 

I've not encountered any existing things that look like that. Have you seen one? It would be easy to find  with an example?

SergeyKanzhelev:
@alejandrox1 are you ok with this approach? Do you have a concern?
@MHBauer can you please make blocking dashboard first? I didn't find use of a description for the dashboard. Not sure if we want to experiment in this PR.

SergeyKanzhelev:
/lgtm
/assign @dashpole

@dashpole we need your approval here. We are introducing a new dashboard with all the e2e tests that are considered blocking for sig node. This is a starting point.

(I will re-lgtm after ordering will change)

MHBauer:
Rebased and updated with hopefully the ideas being more clearly communicated.

I want a duplicate of release-blocking jobs in the sig-node dashboards, so I don't have to go look at sig-release dashboards and figure out which jobs I might be interested in, as a sig-node contributor.

I want as well a short list of jobs that should always be green as they relate to overall sig-node health.

> If we bundle all "good" jobs in a blocking dashboard then we may be replicating the sig release's blocking dashboard (in spirit https://testgrid.k8s.io/sig-release-master-blocking) as it would be a way of telling the community that all of these jobs are well maintained and a failure in them means that something is up.
> And node contributors would still have to switch back and forth between sig-node-blocking and sig-node-kubelet (as an example) to figure out how to interpret the job they are looking at.

I think the duplication is required and good to get people to avoid switching between the two major groups.


---

I see improvements we can make by adding organization and descriptions/comments to existing jobs, as well as maybe some renames. I have left some comments in our group doc.

dashpole:
Thanks for the effort cleaning these up @MHBauer and @SergeyKanzhelev 

/approve

PR #18832, opened 2020-08-13
pull-kubernetes-e2e-gce always_run: false

BenTheElder:
to be replaced with `pull-kubernetes-e2e-gce-ubuntu-containerd`.

we can consider renaming the jobs in a followup, this way doesn't confuse the result history on running PRs

xref: https://github.com/kubernetes/test-infra/issues/18729

### DIFF
--- a/config/jobs/kubernetes/sig-cloud-provider/gcp/gcp-gce.yaml
+++ b/config/jobs/kubernetes/sig-cloud-provider/gcp/gcp-gce.yaml
@@ -17,7 +17,7 @@ presets:
 presubmits:
   kubernetes/kubernetes:
   - name: pull-kubernetes-e2e-gce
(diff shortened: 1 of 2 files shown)


BenTheElder:
cc @liggit @spiffxp @dims @justaugustus 
/hold

spiffxp:
FYI @kubernetes/ci-signal

BenTheElder:
BTW: IIRC this is also duplicated in at least 1.18, but I'm less concerned with locked-in release branches at the moment.

I didn't want 1.19 to deviate from the main branch while we haven't released yet.

spiffxp:
1.18 the ubuntu-containerd job is `always_run: false` so I think it makes sense for this to be blocking for 1.18 and earlier

PR #18835, opened 2020-08-14
Make Triage local clustering multithreaded

michaelkolber:
ref: #18672

/cc BenTheElder spiffxp

We'll want to compare both timing difference and accuracy difference. I've saved the current version of the results (as they get overwritten every time Triage runs within a day).

### DIFF
--- a/triage/summarize/cluster.go
+++ b/triage/summarize/cluster.go
@@ -20,11 +20,17 @@ Contains functions that cluster failures.
 
 package summarize
 
-import "time"
+import (
+	"sync"
+	"time"
+)
 
 /*
 clusterLocal clusters together the failures for each test. Failures come in grouped
 by the test they belong to. These groups are subdivided into clusters.
+
+numWorkers determines how many goroutines to spawn to simultaneously process the failure
+groups. If numWorkers <= 0, the value is set to 1.
 
 Takes:
     {
@@ -49,7 +55,7 @@ Returns:
 		...
 	}
 */
-func clusterLocal(failuresByTest failuresGroup) nestedFailuresGroups {
+func clusterLocal(failuresByTest failuresGroup, numWorkers int) nestedFailuresGroups {
 	const memoPath string = "memo_cluster_local.json"
 	const memoMessage string = "clustering inside each test"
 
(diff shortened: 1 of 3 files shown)


triage/summarize/cluster.go
```diff
@@ -60,17 +67,81 @@ func clusterLocal(failuresByTest failuresGroup) nestedFailuresGroups {
...
+		Each worker will pull from the work queue and perform a clustering operation, and then push
+		the result to the done queue.
+	*/
+
+	var workerWG sync.WaitGroup
+	var doneWG sync.WaitGroup // Prevents this function from returning results before they are all written to clustered
+
+	numWorkers := threads
+	if threads <= 0 {
+		numWorkers = runtime.NumCPU() - 1 // Dedicate one CPU to pushing to the work queue/pulling from the done queue/main thread
```
BenTheElder, line 89: suggestion: I usually recommend hauling references to globals up the stack, it makes it easier to unit test etc. later

I would put this as the default for a flag that allows tuning this.

also: what if NumCPU is 1?

michaelkolber: I set the flag default as max of NumCPU-1 or 1, and then also set a (I think sensible) default of 1 for the function if numWorkers is passed in as <= 0.

BenTheElder:
/lgtm
/approve
once this merges, the next step is to set CPU requests and limits on the job (matching to guarantee CPU) and set the flag to match

triage/summarize/cluster.go
```diff
@@ -60,17 +66,79 @@ func clusterLocal(failuresByTest failuresGroup) nestedFailuresGroups {
...
+
+	// Read from the done queue
+	doneQueueWG.Add(1)
+	go func() {
+		defer doneQueueWG.Done()
+
+		for dg := range doneQueue {
+			numFailures += len(dg.input.Failures)
+			logInfo("%4d/%4d tests, %5d failures, %s", len(clustered)+1, len(failuresByTest), len(dg.input.Failures), dg.input.Key)
+			clustered[dg.input.Key] = dg.output
```
BenTheElder, line 117: ah, I'm surprised this wasn't caught by the unit tests running with race detector enabled,

I missed that this is the same map, you can't have concurrent map read/write without a mutex.

BenTheElder: Actually it's in clusterTest, so it's not obvious in the patchset.
https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/ci-test-infra-triage/1294155497288503297

michaelkolber: Ah, yeah, I should have followed the function calls all the way down.

michaelkolber: Addressed in #18845

PR #18858, opened 2020-08-14
Switch from log to klog

michaelkolber:
ref: #18726

Also:
- Gracefully handlesome previously fatal errors
- Make some fatal errors fail fast

/cc BenTheElder spiffxp

### DIFF
--- a/triage/summarize/cluster.go
+++ b/triage/summarize/cluster.go
@@ -23,6 +23,8 @@ package summarize
 import (
 	"sync"
 	"time"
+
+	"k8s.io/klog/v2"
 )
 
 /*
@@ -68,7 +70,7 @@ func clusterLocal(failuresByTest failuresGroup, numWorkers int) nestedFailuresGr
 
 	numFailures := 0 // The number of failures processed so far
 	start := time.Now()
-	logInfo("Clustering failures for %d unique tests...", len(failuresByTest))
+	klog.V(2).Infof("Clustering failures for %d unique tests...", len(failuresByTest))
 
 	/*
 		Since local clustering is done within each test, there is no interdependency among the clusters,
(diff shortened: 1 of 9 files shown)


triage/update_summaries.sh
```diff
@@ -72,7 +72,9 @@ mkdir -p slices
   --builds triage_builds.json \
   --tests triage_tests/*.json \
   --output failure_data.json \
-  --output_slices slices/failure_data_PREFIX.json
+  --output_slices slices/failure_data_PREFIX.json \
+  --logtostderr=true \
```
michaelkolber, line 76: I'm pretty sure this is wrong but I'm not sure how else to pass in the flag

BenTheElder: binaries frequently set this themselves / hide the flag

BenTheElder: https://cs.k8s.io/?q=logtostderr%22&i=nope&files=&repos=

michaelkolber: Setting it internally does not make the flag work

michaelkolber: I switched to klog/v2 _and_ called InitFlags, that should do it

michaelkolber: Added a custom flagset as per https://github.com/kubernetes/klog/blob/master/examples/coexist_glog/coexist_glog.go

michaelkolber: Ok, it seems to have passed without any issues

michaelkolber:
This is ready for review

triage/summarize/cluster.go
```diff
@@ -68,7 +70,7 @@ func clusterLocal(failuresByTest failuresGroup, numWorkers int) nestedFailuresGr
 
 	numFailures := 0 // The number of failures processed so far
 	start := time.Now()
-	logInfo("Clustering failures for %d unique tests...", len(failuresByTest))
+	klog.V(2).Infof("Clustering failures for %d unique tests...", len(failuresByTest))
```
MushuEE, line 73: What is enabling the V-level logging? Will it log nothing if not enabled? https://godoc.org/github.com/kubernetes/klog#Verbose.Enabled

michaelkolber: I think by default it logs at V2, but regardless it's being set here:  
https://github.com/kubernetes/test-infra/pull/18858/files#diff-e209704c6b63bec066319227a3d98099R94

PR #18862, opened 2020-08-14
Add resource requests to Triage

michaelkolber:
This should make the results more consistent
ref: #18726
/cc BenTheElder spiffxp

### DIFF
--- a/config/jobs/kubernetes/test-infra/test-infra-periodics.yaml
+++ b/config/jobs/kubernetes/test-infra/test-infra-periodics.yaml
@@ -45,6 +45,13 @@ periodics:
       - name: service
         mountPath: /etc/service-account
         readOnly: true
+      resources:
+        requests:
+          cpu: 4
+          memory: 2Gi
+        limits:
+          cpu: 4
+          memory: 2Gi
     volumes:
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes/test-infra/test-infra-periodics.yaml
```diff
@@ -45,6 +45,13 @@ periodics:
       - name: service
         mountPath: /etc/service-account
         readOnly: true
+      resources:
+        requests:
+          cpu: 4
```
BenTheElder, line 50: we should set the number of workers flag

michaelkolber: Why not let it pick it up automatically?

BenTheElder: Is it correctly detecting it?
There are various mechanisms for finding this info on linux, and IIRC not all of them are correctly masked by containers.

michaelkolber: I can add some logging so we can be sure

PR #18864, opened 2020-08-14
description of some kind pull/pr jobs

MHBauer:
Focused on pull/pr jobs because I see these most often when I am
looking at a PR, and I have not yet learned to decode the name.

### DIFF
--- a/config/jobs/kubernetes/sig-testing/kubernetes-kind-presubmits.yaml
+++ b/config/jobs/kubernetes/sig-testing/kubernetes-kind-presubmits.yaml
@@ -49,6 +49,7 @@ presubmits:
       testgrid-alert-stale-results-hours: '24'
       testgrid-create-test-group: 'true'
       fork-per-release: "true"
+      description: "Run a quick subset of e2e tests, skipping those that are flaky."
 
 
   - name: pull-kubernetes-e2e-kind-canary
@@ -156,6 +157,7 @@ presubmits:
       testgrid-num-failures-to-alert: '10'
       testgrid-alert-stale-results-hours: '24'
       testgrid-create-test-group: 'true'
+      description: "Run a quick subset of e2e tests, with IPv6 networking, skipping those that are flaky."
 
   - name: pull-kubernetes-e2e-kind-ipv6-canary
     cluster: k8s-infra-prow-build
@@ -256,6 +258,7 @@ presubmits:
       testgrid-alert-stale-results-hours: '24'
       testgrid-create-test-group: 'true'
       testgrid-dashboards: sig-testing-kind
+      description: "Run all conformance tests with only GA APIs enabled to ensure that conformance tests do not rely on beta APIs or features."
 
   - name: pull-kubernetes-conformance-kind-ga-only-parallel
     always_run: true
@@ -301,4 +304,5 @@ presubmits:
       testgrid-alert-stale-results-hours: '24'
       testgrid-create-test-group: 'true'
       testgrid-dashboards: sig-testing-kind
+      description: "Run only parallel non-[SERIAL] conformance tests with only GA APIs enabled to ensure that conformance tests do not rely on beta APIs or features."
     cluster: k8s-infra-prow-build


config/jobs/kubernetes/sig-testing/kubernetes-kind-presubmits.yaml
```diff
@@ -256,6 +256,7 @@ presubmits:
       testgrid-alert-stale-results-hours: '24'
       testgrid-create-test-group: 'true'
       testgrid-dashboards: sig-testing-kind
+      description: "Run all conformance tests with only GA APIs enabled to ensure that conformance tests do not rely on beta APIs or features."
```
BenTheElder, line 259: I forget, but I think this skips the serial tests

config/jobs/kubernetes/sig-testing/kubernetes-kind-presubmits.yaml
```diff
@@ -49,7 +49,7 @@ presubmits:
       testgrid-alert-stale-results-hours: '24'
       testgrid-create-test-group: 'true'
       fork-per-release: "true"
-
```
BenTheElder, line 52: nit: inconsistent newline removal

config/jobs/kubernetes/sig-testing/kubernetes-kind-presubmits.yaml
```diff
@@ -49,7 +49,7 @@ presubmits:
...
+      description: "Run a quick subset of e2e tests, skipping those that are flaky."
```
BenTheElder, line 52: there is also the -ipv6 job

config/jobs/kubernetes/sig-testing/kubernetes-kind-presubmits.yaml
MHBauer: huh, weird, looks like there was two blanks before. I'll put it back?

config/jobs/kubernetes/sig-testing/kubernetes-kind-presubmits.yaml
```diff
@@ -156,6 +157,7 @@ presubmits:
       testgrid-num-failures-to-alert: '10'
       testgrid-alert-stale-results-hours: '24'
       testgrid-create-test-group: 'true'
+      description: "Run a quick subset of e2e tests, with IPv6 networking, skipping those that are flaky."
```
MHBauer, line 160: I'm not too happy with this one. Can I get any more explicit behind the "why" ipv6 to fill in here?

I'll do some research on the history real quick. Be back soon.

spiffxp: This is the only ipv6 e2e coverage the project has, everything else is ipv4

spiffxp: By that I mean it's important that the project have e2e ipv6 coverage, since it's supported, and so far this is the sole mechanism we have to enforce it upstream (that I am aware of anyway)

MHBauer: Other questions I have:
Is it IPv6 only? Is IPv4 never used, or disabled somehow? Is it some error if I were to see IPv4 addresses used during a run somehow?
Is this the only place it's checked?
Are there tests that ought to behave differently, or not run, or only run, when IPv6 is on?
Should everything be IP version agnostic, as far as we know? That is a desired end-state and this is checking to ensure it by running the same suite of tests?

spiffxp: I think @aojea can probably answer these questions

spiffxp: /cc @aojea

aojea: > Is it IPv6 only? Is IPv4 never used, or disabled somehow? Is it some error if I were to see IPv4 addresses used during a run somehow?
The environment were it runs is the same that runs for the IPv4 only, dual stack, the host has IPv4 and IPv6 addresses, but Kubernetes only uses the corresponding to the IP family it has been configured, if not it should be a bug, but there are certain exceptions

> Is this the only place it's checked?
there are periodic jobs with KIND and another one in Azure
> Are there tests that ought to behave differently, or not run, or only run, when IPv6 is on?
Those have a tag [Networking-IPv6]
> Should everything be IP version agnostic, as far as we know? That is a desired end-state and this is checking to ensure it by running the same suite of tests?
I did a lot of job on that area, I have implemented a method in the e2e framework that "identifies" the IP family of the cluster (based in the IP family of the "kubernetes" service). Ideally, the test should be agnostic . Actually I submitted a PR related today https://github.com/kubernetes/kubernetes/pull/94433

spiffxp:
#20498 (comment)

PR #18868, opened 2020-08-14
code of conduct committee update

tpepper:
Rotating out our emeriti members and adding the 2020 elected members.

Signed-off-by: Tim Pepper <tpepper@vmware.com>


BenTheElder:
manually syncrhonizing these seems terrible

@cjwagner could we support multiple OWNERS_ALIASES files?
we could place the primary one in github.com/kubernetes/.github and only edit it once, and then have the repo specific aliases in some secondary file.

PR #18870, opened 2020-08-15
Kopsgce: create a statestore bucket for each test

geojaz:
We've been having issues with our kops GCE E2E tests (for example: https://storage.googleapis.com/kubernetes-jenkins/logs/e2e-kops-gce-latest/1290333264250146817/artifacts/104.154.223.210/etcd.log)

I believe this is due to permissions on the state store bucket. This PR is a temporary effort to get the tests working again before we check out kubetest2.  

This patch creates a bucket to use for the state store for each job instead of trying to adjust ACLs or anything else on the "normal/default" e2e bucket. It will create the bucket when in kops/gce mode only.

TODO/Questions:
* Do I need to delete the bucket after the test? does it get cleaned out automatically?
* We may need to grant the default compute service account object admin on the newly created bucket.

### DIFF
--- a/kubetest/BUILD.bazel
+++ b/kubetest/BUILD.bazel
@@ -45,6 +45,7 @@ go_library(
         "@com_github_pelletier_go_toml//:go_default_library",
         "@com_github_satori_go_uuid//:go_default_library",
         "@com_github_spf13_pflag//:go_default_library",
+        "@com_google_cloud_go_storage//:go_default_library",
         "@io_k8s_sigs_boskos//client:go_default_library",
         "@org_golang_x_crypto//ssh:go_default_library",
     ],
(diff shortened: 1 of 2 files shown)


scenarios/kubernetes_e2e.py
```diff
@@ -272,6 +272,8 @@ def set_up_kops_gce(workspace, args, mode, cluster, runner_args):
     for path in [args.gce_ssh, args.gce_pub]:
         if not os.path.isfile(os.path.expandvars(path)):
             raise IOError(path, os.path.expandvars(path))
+    
+
```
rifelpet, line 276: unnecessary diff here?

geojaz: must have been some linting. i'll undo.

PR #18877, opened 2020-08-17
GKE: Disable shielded nodes if using custom image

karan:
In 1.18+ Shielded Nodes are enabled by default in GKE. However, custom images are not supported with shielded nodes. This disables the feature when using custom image type.

In this case, the cluster will not come up successfully, and the kubelet will just fail to connect to the apiserver.

### DIFF
--- a/kubetest/gke.go
+++ b/kubetest/gke.go
@@ -336,6 +336,8 @@ func (g *gkeDeployer) Up() error {
 		// gcloud enables node auto-upgrade by default, which doesn't work with CUSTOM image.
 		// We disable auto-upgrade explicitly here.
 		args = append(args, "--no-enable-autoupgrade")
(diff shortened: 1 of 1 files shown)


kubetest/gke.go
```diff
@@ -336,6 +336,7 @@ func (g *gkeDeployer) Up() error {
 		// gcloud enables node auto-upgrade by default, which doesn't work with CUSTOM image.
 		// We disable auto-upgrade explicitly here.
 		args = append(args, "--no-enable-autoupgrade")
+		args = append(args, "--no-enable-shielded-nodes")
```
MHBauer, line 339: Commit message is good for this line if I look at the history in the future.

Might want to add to the comment above or create a separate inline comment with the reason.

karan: Done

PR #18880, opened 2020-08-17
Pin test version to v1.18.0 for ci-cloud-provider-gcp-conformance

amwat:
ref: https://github.com/kubernetes/kubernetes/issues/94051

/cc @liggitt @BenTheElder

### DIFF
--- a/config/jobs/kubernetes/cloud-provider-gcp/cloud-provider-gcp-periodics.yaml
+++ b/config/jobs/kubernetes/cloud-provider-gcp/cloud-provider-gcp-periodics.yaml
@@ -40,4 +40,4 @@ periodics:
         go get sigs.k8s.io/kubetest2@latest;
         go get sigs.k8s.io/kubetest2/kubetest2-gce@latest;
         go get sigs.k8s.io/kubetest2/kubetest2-tester-ginkgo@latest;
-        kubetest2 gce -v 2 --repo-root $REPO_ROOT --build --up --down --test=ginkgo -- --focus-regex='\[Conformance\]'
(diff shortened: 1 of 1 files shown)


amwat:
Do we want to rename the job to say v1.18 instead of latest? That would also lose testgrid history though.

config/jobs/kubernetes/cloud-provider-gcp/cloud-provider-gcp-periodics.yaml
```diff
@@ -40,4 +40,4 @@ periodics:
         go get sigs.k8s.io/kubetest2@latest;
         go get sigs.k8s.io/kubetest2/kubetest2-gce@latest;
         go get sigs.k8s.io/kubetest2/kubetest2-tester-ginkgo@latest;
-        kubetest2 gce -v 2 --repo-root $REPO_ROOT --build --up --down --test=ginkgo -- --focus-regex='\[Conformance\]'
+        kubetest2 gce -v 2 --repo-root $REPO_ROOT --build --up --down --test=ginkgo -- --test-package-version=v1.18.0 --focus-regex='\[Conformance\]'
```
liggitt, line 43: is testing 1.18 with 1.18 tests that useful? I'm most interested in changes in dev/1.19 requiring reaction

amwat: updated the PR description. cloud provider gcp is pinned to v1.18
https://github.com/kubernetes/cloud-provider-gcp/blob/master/WORKSPACE#L108

amwat: but yeah I think the intention of this job should be to use it as a signal to fix the cloud provider gcp builds to be able to support kubernetes at HEAD.
I would say failures are WAI in that case?

PR #18886, opened 2020-08-18
use gh workflow YAML compliant names for prow checks

tariq1890:
Just realised that we need to use the `job` yaml object name as the check over here and the value of the `job` object's field `name`.


tariq1890:
The checks can be found here: https://github.com/kubernetes/kube-state-metrics/blob/master/.github/workflows/ci.yml#L11

tariq1890:
Thanks for the review @BenTheElder !

PR #18891, opened 2020-08-18
Enable per-zone logexporter in multizonal and regional GKE clusters

tosi3k:
Leverage the mechanism introduced in https://github.com/kubernetes/kubernetes/pull/94056 to prevent logexporter from duplicating work for multizonal/regional clusters when conducting log dumping from the tests.

Also fix the comment describing `gkeLogDumpTemplate` const on the occasion.

/sig scalability
/cc mm4tt
/cc jkaniuk

### DIFF
--- a/kubetest/gke.go
+++ b/kubetest/gke.go
@@ -408,8 +408,10 @@ func (g *gkeDeployer) DumpClusterLogs(localPath, gcsPath string) error {
 	// gkeLogDumpTemplate is a template of a shell script where
 	// - %[1]s is the project
 	// - %[2]s is the zone
-	// - %[3]s is a filter composed of the instance groups
-	// - %[4]s is the log-dump.sh command line
+	// - %[3]s is the OS distribution of nodes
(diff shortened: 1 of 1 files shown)


kubetest/gke.go
```diff
@@ -437,6 +440,7 @@ export KUBE_NODE_OS_DISTRIBUTION='%[3]s'
 		filter := fmt.Sprintf("(metadata.created-by ~ %s)", ig.path)
 		perZoneFilters[ig.zone] = append(perZoneFilters[ig.zone], filter)
 	}
+	isMultizonalOrRegional := len(g.instanceGroups) > 1
```
mm4tt, line 443: What if we have two instance groups in the same zone?

tosi3k: Whoops... that was supposed to be:
```
isMultizonalOrRegional := len(perZoneFilters) > 1
```
Good catch!

tosi3k: Done.

PR #18892, opened 2020-08-18
Azure: bump aks-engine orchestrator release to 1.19

chewong:
/assign @marosset @feiskyer 
/cc @jsturtevant


config/jobs/kubernetes-sigs/sig-windows/sig-windows-config.yaml
```diff
@@ -234,7 +234,7 @@ periodics:
 #       # Azure-specific test args
 #       - --deployment=aksengine
 #       - --provider=skeleton
-#       - --aksengine-orchestratorRelease=1.18
+#       - --aksengine-orchestratorRelease=1.19
```
jsturtevant, line 237: I think we should remove this commented section all together. The changes in any config might not have made it to this and we should start from scratch if we want to add a new job.

jsturtevant:
The config updates lgtm.  I will let other look at owns/reviewers.

To double check my understanding the flag `- --aksengine-orchestratorRelease` is used by aks-engine to set defaults on kubelet and other components and so should be set to the latest possible, even though these jobs are are technically building for 1.20 (master branch at this time)?  Any additional flags that would be required for new 1.20+ features would need to be added directly to the api model in https://github.com/kubernetes-sigs/windows-testing/tree/master/job-templates?

marosset:
@jsturtevant your understanding of how orchestorRelease works is the same as mine.
Normally it is used to pick the Kubernetes components too but in this case we are using binaries built in the PR job.

chewong:
> The config updates lgtm. I will let other look at owns/reviewers.
> 
> To double check my understanding the flag `- --aksengine-orchestratorRelease` is used by aks-engine to set defaults on kubelet and other components and so should be set to the latest possible, even though these jobs are technically building for 1.20 (master branch at this time)? Any additional flags that would be required for new 1.20+ features would need to be added directly to the api model in https://github.com/kubernetes-sigs/windows-testing/tree/master/job-templates?

That's correct. `- --aksengine-orchestratorRelease` is also used to determine whether we are deploying the cluster with hyperkube or individual kube images. Once 1.20 orchestrator release is available in aks-engine, we could switch the flag value to 1.20.

I am not aware of any flags that need to be added to the API model ATM.

PR #18896, opened 2020-08-19
Update Metric with query info

MushuEE:
Added small block about how to write and test the queries.

Just boy-scouting as I investigate flake query updates


MushuEE:
/hold
something happened...

PR #18899, opened 2020-08-19
Extend kettle unit testing for build object

MushuEE:
I also reorg some of the methods to be more visible in setup and to make more sense with the flow of the Build object

I will extend unit tests for more methods in other PRs. Don't want to have super massive reviews with all the other things in our pipeline

### DIFF
--- a/kettle/make_json.py
+++ b/kettle/make_json.py
@@ -86,7 +86,7 @@ class Build:
         self.repos = None
         self.metadata = None
         self.elapsed = None
-        self.path_to_job_and_number()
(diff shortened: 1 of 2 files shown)


kettle/make_json.py
```diff
@@ -75,6 +97,25 @@ def generate(cls, path, tests, started, finished, metadata, repos):
         build.set_elapsed()
         return build
 
+    def path_to_job_and_number(self):
```
amwat, line 100: should this be renamed to `populate_path_to_job_and_number` now?

PR #18902, opened 2020-08-19
Temporarily increase master-size to 8vCPU in scalability 100 node jobs

mm4tt:
Ref. https://github.com/kubernetes/kubernetes/issues/94098

### DIFF
--- a/config/jobs/kubernetes/sig-scalability/sig-scalability-presubmit-jobs.yaml
+++ b/config/jobs/kubernetes/sig-scalability/sig-scalability-presubmit-jobs.yaml
@@ -31,6 +31,8 @@ presubmits:
         - --extract=local
         - --flush-mem-after-build=true
         - --gcp-nodes=100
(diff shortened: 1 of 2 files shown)


config/jobs/kubernetes/sig-scalability/sig-scalability-presubmit-jobs.yaml
```diff
@@ -31,6 +31,8 @@ presubmits:
         - --extract=local
         - --flush-mem-after-build=true
         - --gcp-nodes=100
+        # TODO(k/k#94098): Remove the line below once we get enough data.
+        - --gcp-master-size=n1-standard-8
```
wojtek-t, line 35: Let's do the same for post submit in master branch.

mm4tt: Done

wojtek-t:
/lgtm
/approve

Thanks

PR #18903, opened 2020-08-19
add NFS CSI driver sanity test

andyzhangx:
/assign @msau42

after this PR(https://github.com/kubernetes-csi/csi-driver-nfs/pull/49) merged, we could run sanity test in CI.

### DIFF
--- a/config/jobs/kubernetes-csi/csi-driver-nfs/OWNERS
+++ b/config/jobs/kubernetes-csi/csi-driver-nfs/OWNERS
@@ -1,0 +1,8 @@
+# See the OWNERS docs at https://go.k8s.io/owners
+
+reviewers:
+- andyzhangx
+- msau42
+approvers:
+- andyzhangx
+- msau42
--- a/config/jobs/kubernetes-csi/csi-driver-nfs/csi-driver-nfs-unmanaged.yaml
+++ b/config/jobs/kubernetes-csi/csi-driver-nfs/csi-driver-nfs-unmanaged.yaml
@@ -1,0 +1,26 @@
+presubmits:
+  kubernetes-csi/csi-driver-nfs:
+  - name: pull-csi-driver-nfs-sanity
+    decorate: true
+    always_run: true
+    path_alias: sigs.k8s.io/csi-driver-nfs
+    branches:
+    - master
+    labels:
+      preset-service-account: "true"
+      preset-azure-cred: "true"
+    spec:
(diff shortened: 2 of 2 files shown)


config/jobs/kubernetes-csi/csi-driver-nfs/csi-driver-nfs-config.yaml
```diff
@@ -29,3 +29,27 @@ presubmits:
         resources:
           requests:
             cpu: 2000m
+  - name: pull-csi-driver-nfs-sanity
```
msau42, line 32: This file is auto generated by gen-jobs.sh. @pohly do you have any suggestions on how to add custom tests?

andyzhangx: and what about e2e tests support in the future, some driver supports, some driver does not support that. Cannot be all generated by gen-jobs.sh right?

andyzhangx: I figured it out, just move this config to another new file.

andyzhangx:
@msau42 could you approve? thanks.

msau42:
This config is going to get lost the next time we update the csi jobs. To get unblocked, I guess we can remove the repo from gen-jobs.sh. @pohly if you have suggestions on how we can better support non-hostpath csi drivers running their own sanity and e2e tests, that would be greatly appreciated.

andyzhangx:
> This config is going to get lost the next time we update the csi jobs. To get unblocked, I guess we can remove the repo from gen-jobs.sh. @pohly if you have suggestions on how we can better support non-hostpath csi drivers running their own sanity and e2e tests, that would be greatly appreciated.

@msau42 I created a new file `csi-driver-nfs-config-presubmits.yaml`, would that file still be overwritten by gen-jobs.sh?

msau42:
That should work for now. Can you name the file something like "csi-driver-nfs-unmanaged.yaml" or something like that to be more clear this is not the normal generated file? And to avoid confusion with the other file that also has presubmits.

andyzhangx:
> That should work for now. Can you name the file something like "csi-driver-nfs-unmanaged.yaml" or something like that to be more clear this is not the normal generated file? And to avoid confusion with the other file that also has presubmits.

done, PTAL.

PR #18905, opened 2020-08-19
Rename default git user for clonerefs to `clonerefs`

alvaroaleman:
This helps identifying commits created by this user, the generic
`ci-robot` we currently use makes this hard.

This change will break scripts that expect to find a commit authored by
the old name.

### DIFF
--- a/prow/clonerefs/options.go
+++ b/prow/clonerefs/options.go
@@ -109,9 +109,9 @@ const (
 	// in when run.
 	JSONConfigEnvVar = "CLONEREFS_OPTIONS"
 	// DefaultGitUserName is the default name used in git config
-	DefaultGitUserName = "ci-robot"
+	DefaultGitUserName = "clonerefs"
 	// DefaultGitUserEmail is the default email used in git config
-	DefaultGitUserEmail = "ci-robot@k8s.io"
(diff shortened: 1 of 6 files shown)


cjwagner:
> @cjwagner do we want some sort of deprecation period for something like this?

Probably? I don't know of anything that depends on this, but I wouldn't be surprised if this breaks something.

What else creates commits as `ci-robot@k8s.io` that you are trying to distinguish from?

alvaroaleman:
> What else creates commits as ci-robot@k8s.io that you are trying to distinguish from?

The main issue is the title of `ci-robot`. I was debugging an issue where we accidentally merged a PR, looked at the commit history and saw the `ci-robot` and thought it was just tide.

cjwagner:
Tide should always create commits with the bot account since it uses the GH v3 API to merge PRs. I agree that `clonerefs@k8s.io` would be clearer than `ci-robot@k8s.io`, but is that worth a breaking change? 
I don't feel strongly about this though and I suspect most automation that is expecting a specific bot name will care about Tide merge commits, not whatever clonerefs is doing, so I don't think this will break much if anything. How did you end up with a commit from clonerefs in the commit history?

PR #18907, opened 2020-08-19
bugzilla: instrument event-local logging

stevekuznetsov:
Similar to how it's done for the GitHub client, the Bugzilla client now
exposes all of the event-specific data fields when logging, as well as
passing a specific and explicit user agent header.

/assign @alvaroaleman

### DIFF
--- a/prow/bugzilla/client.go
+++ b/prow/bugzilla/client.go
@@ -32,6 +32,8 @@ import (
 	"github.com/sirupsen/logrus"
 	"golang.org/x/sync/errgroup"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+
+	"k8s.io/test-infra/prow/version"
 )
 
 const (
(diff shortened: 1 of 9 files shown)


prow/bugzilla/client.go
```diff
@@ -137,8 +145,53 @@ func (bd *bugDetailsCache) list() []Bug {
...
+
+// ForPlugin clones the client, keeping the underlying delegate the same but adding
+// a plugin identifier and log field
+func (c *client) ForPlugin(plugin string) Client {
+	return c.forKeyValue("plugin", plugin)
+}
+
+// ForSubcomponent clones the client, keeping the underlying delegate the same but adding
+// an identifier and log field
+func (c *client) ForSubcomponent(subcomponent string) Client {
```
alvaroaleman, line 165: This is unused, isn't it?

stevekuznetsov: I expect users of the client downstream to make use of it, same as with the GitHub one

alvaroaleman: okay. LGTM modulo the bazel failures then

PR #18908, opened 2020-08-19
Recommend number of janitor replicas to use

coryrc:
When I first setup boskos I was very confused on how to set this
number. Knative had a ratio of 1:6 for a long time, which is a little
ridiculous.

/assign ixdy
/cc ixdy

### DIFF
--- a/config/prow/cluster/boskos-janitor.yaml
+++ b/config/prow/cluster/boskos-janitor.yaml
@@ -6,7 +6,7 @@ metadata:
     app: boskos-janitor
   namespace: test-pods
 spec:
-  replicas: 4  # 4 distributed janitor instances for gke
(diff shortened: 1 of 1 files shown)


config/prow/cluster/boskos-janitor.yaml
```diff
@@ -6,7 +6,7 @@ metadata:
     app: boskos-janitor
   namespace: test-pods
 spec:
-  replicas: 4  # 4 distributed janitor instances for gke
+  replicas: 4  # 4 distributed janitor instances for gke; recommend min 2, 1 for each 30 projects
```
ixdy, line 9: this documentation might be better suited for the kubernetes-sigs/boskos project.

Issue #18851, opened 2020-08-14
Migrate merge-blocking jobs to dedicated cluster: pull-kubernetes-node-e2e


spiffxp:
**What should be cleaned up or changed**:

This is part of #18550

To properly monitor the outcome of this, you should be a member of k8s-infra-prow-viewers@kubernetes.io.  PR yourself into https://github.com/kubernetes/k8s.io/blob/master/groups/groups.yaml#L603-L628 if you're not a member.

Migrate pull-kubernetes-node-e2e to k8s-infra-prow-build by adding a `cluster: k8s-infra-prow-build` field to the job:
- Make sure any release-branch variants are also updated
- Reference this issue, but do not "fix" it; instead leave this open to monitor the behavior of the job before/after
- Here's an example PR that did this for pull-kubernetes-integration: https://github.com/kubernetes/test-infra/pull/18814

NOTE: migrating this job is not as straightforward as some of the other #18550 issues, because we also need to:
- Switch away from a fixed GCP project to boskos-managed projects
  - Replace `--gcp-project=k8s-jkns-pr-node-e2e` with `--gcp-project-type=gce-project`
- If this turns out to break things, revert and ask for help

Once the PR has merged, note the date/time it merged.  This will allow you to compare before/after behavior.

Things to watch for the job
- https://prow.k8s.io/?job=pull-kubernetes-node-e2e
  - does the job start failing more often?
  - does the job start going into error state?
- https://testgrid.k8s.io/presubmits-kubernetes-blocking#pull-kubernetes-node-e2e&graph-metrics=test-duration-minutes
  - does the job duration look worse than before? spikier than before?
- https://storage.googleapis.com/k8s-gubernator/triage/index.html?pr=1&job=pull-kubernetes-node-e2e
  - do more failures show up than before?
- https://prow.k8s.io/job-history/gs/kubernetes-jenkins/pr-logs/directory/pull-kubernetes-node-e2e
  - (can be used to answer some of the same questions as above)
[...]

PR #18915, opened 2020-08-19
Migrate pull-kubernetes-node-e2e to k8s-infra-prow-build

spiffxp:
Drop the use of k8s-jkns-pr-node-e2e, which means the jobs will
instead use the gce-project pool from boskos, which is the same
pool used by the ci equivalent of this job.

Part of https://github.com/kubernetes/test-infra/issues/18851


spiffxp:
/cc @BenTheElder 
FYI @SergeyKanzhelev @vpickard @kubernetes/ci-signal

Issue #18852, opened 2020-08-14
Migrate merge-blocking jobs to dedicated cluster: pull-kubernetes-e2e-gce


spiffxp:
**What should be cleaned up or changed**:

This is part of #18550

To properly monitor the outcome of this, you should be a member of k8s-infra-prow-viewers@kubernetes.io.  PR yourself into https://github.com/kubernetes/k8s.io/blob/master/groups/groups.yaml#L603-L628 if you're not a member.

Migrate pull-kubernetes-e2e-gce to k8s-infra-prow-build by adding a `cluster: k8s-infra-prow-build` field to the job:
- Make sure any release-branch variants are also updated
- Reference this issue, but do not "fix" it; instead leave this open to monitor the behavior of the job before/after
- Here's an example PR that did this for pull-kubernetes-integration: https://github.com/kubernetes/test-infra/pull/18814

NOTE: migrating this job is not as straightforward as some of the other #18550 issues, because:
- it's being demoted from merge-blocking on release-1.19 and the main branch (as of https://github.com/kubernetes/test-infra/pull/18832)
- so in terms of appropriate amount of PR traffic, either manually trigger this job, or look for release-1.18 (and earlier) PRs

Once the PR has merged, note the date/time it merged.  This will allow you to compare before/after behavior.

Things to watch for the job
- https://prow.k8s.io/?job=pull-kubernetes-e2e-gce
  - does the job start failing more often?
  - does the job start going into error state?
- https://testgrid.k8s.io/presubmits-kubernetes-blocking#pull-kubernetes-e2e-gce&graph-metrics=test-duration-minutes
  - does the job duration look worse than before? spikier than before?
- https://storage.googleapis.com/k8s-gubernator/triage/index.html?pr=1&job=pull-kubernetes-e2e-gce
  - do more failures show up than before?
- https://prow.k8s.io/job-history/gs/kubernetes-jenkins/pr-logs/directory/pull-kubernetes-e2e-gce
  - (can be used to answer some of the same questions as above)
[...]

PR #18916, opened 2020-08-19
Migrate pull-kubernetes-e2e-gce to k8s-infra-prow-build

spiffxp:
Even though 1.19 and the main branch variants are no longer
merge blocking, let's go ahead and move them all at once

Part of https://github.com/kubernetes/test-infra/issues/18852


spiffxp:
FYI @kubernetes/ci-signal

PR #18931, opened 2020-08-20
Fix Triage binary not running

michaelkolber:
ref: #18726

Add GOOS env variable to Triage build step

/cc BenTheElder spiffxp

### DIFF
--- a/triage/Dockerfile
+++ b/triage/Dockerfile
@@ -22,7 +22,7 @@ COPY . /test-infra/
 
 # Build Triage
 RUN cd /test-infra/triage/ && \
-    go build -v -o triage ./main.go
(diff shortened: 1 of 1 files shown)


triage/Dockerfile
```diff
@@ -22,7 +22,7 @@ COPY . /test-infra/
 
 # Build Triage
 RUN cd /test-infra/triage/ && \
-    go build -v -o triage ./main.go
+    env GOOS=linux go build -v -o triage ./main.go
```
MushuEE, line 25: Do we still need the `apk add python2` below?

MushuEE: nvm!

michaelkolber: Yes, for the Cloud SDK

PR #18933, opened 2020-08-20
Michael is the local triage expert should be OWNER

MushuEE:
He also expressed interest in staying active. Thanks @michaelkolber

/assign @BenTheElder @spiffxp 
cc/ @michaelkolber


BenTheElder:
/approve
/lgtm
/hold

usually the process is member => reviewer => approver, but seeing as you wrote it all :-)

MushuEE:
@BenTheElder do we not have a hack that checks if owner changes meet the member requirement?

BenTheElder:
Michael is a member.

PR #18941, opened 2020-08-21
Add more fields to prowjob-report

spiffxp:
- generate consistently sorted output for ease of diffing
- add ProwJobReportRow type, update json/html formats to use it
- use total resources for all containers
- add always_run
- guess periodic repo/branch
- guess presubmit/postsubmit branches
- add a date column (e.g. when was this last run, or config as of when)
- reorder csv output

[example of csv imported into sheets](https://docs.google.com/spreadsheets/d/1dgSxfa0jdaYk76S6DqJeazl2TKVAvZcCg-7Yx3U0j-I/edit#gid=1043959272&fvid=711578984) using a filter view to restrict to critical jobs


MushuEE:
Thanks @spiffxp! I will update html to reflect this after merge. There are some nits but as this still lives in experimental I don't think boy-scouting is needed. 
/lgtm

PR #18951, opened 2020-08-22
Mkpod: Create pods that don't get cleaned up by sinker

alvaroaleman:


prow/cmd/mkpod/main.go
```diff
@@ -131,6 +132,17 @@ func main() {
 		}
 	}
 
+	// We need to remove the created-by-prow label, otherwise sinker will promptly clean this
```
fejta, line 135: I guess kubernetes will eventually clean it up automatically and we expect the volume of these pods created within that duration to be insignificant?

PR #8131, opened 2018-05-24
[WIP] Command plugin framework and 'merge-blocker' plugin that uses it.

cjwagner:
Design doc: https://docs.google.com/document/d/1Lo1jn6DORjbj8szkE-kc2nXwooQMmrvcYpbfhr7fFnE/edit?usp=sharing (kubernetes-dev and kubernetes-sig-testing google groups can view).

This PR introduces a simple framework for defining and using modular plugin handling logic for command based plugins like `/hold`, `/milestone foo`, etc. It exposes functions for multiple levels of handling logic abstraction; simple commands that control a single label can be declared in a few lines while more complex commands could just use the command regexp matching provided by `Command`s.
I only implemented some of the possible modular components. More can be added as needed.
This framework should generalize much of the common handling logic in Prow plugins that implement commands such as regexp matching for the command, validating the access rights of the commenter, and adding or removing labels.

I'm thinking that once there are enough modular components I'll add a new Prow plugin that consumes a config that declaratively defines commands. Different command types would have different configs. A simple one might look like this:
```yaml
single-label-command:
  name: merge-blocker
  label: merge-blocker
  policy:
    team-access:
      id: 12345
      name: kubernetes-sig-testing
  # Maybe some fields to allow the plugin to be documented
```
This would let us define simple commands in config at runtime without updating or redeploying code. 

The `merge-blocker` plugin implements a command that manages a `merge-blocker` label and is only usable by members of a specific GitHub team. Tide will eventually search for this label to find issues that should block PRs from merging.
I may switch some of the simpler plugins over to using this framework in a follow up PR.

Tests still need to be added.

/area prow
/cc @kargakis @krzyzacy @BenTheElder  @fejta @stevekuznetsov @rmmh

PR #18955, opened 2020-08-24
Allow applying tide/merge-blocker using label command

nikhita:
Ref: https://github.com/kubernetes/website/issues/23379#issuecomment-678849678

/assign @spiffxp 

cc @jimangel @zacharysarah

### DIFF
--- a/config/prow/plugins.yaml
+++ b/config/prow/plugins.yaml
@@ -125,6 +125,7 @@ label:
     - platform/gcp
     - platform/minikube
     - platform/other
+    - tide/merge-blocker
     - tide/merge-method-merge
     - tide/merge-method-rebase
     - tide/merge-method-squash


nikhita:
cc @savitharaghunathan

justaugustus:
/hold

I just want to get some clarification on who can apply this label before proceeding.
I don't think we want to allow anyone who can apply labels to be able to block Tide (we probably want this restricted to repo admins/writers and approvers)? Maybe this requires its own plugin?

jimangel:
It would be awesome to create an allow list like milestone maintainers: https://github.com/kubernetes/org/blob/master/config/kubernetes/sig-release/teams.yaml#L2

Or potentially add the ability to block to the milestone maintainers group.

Not sure the feasibility of that type of restriction...

MushuEE:
Quick reminder to also update [docs](https://github.com/kubernetes/test-infra/blob/master/label_sync/labels.md#labels-that-apply-to-all-repos-only-for-issues) if a plugin is added and the `Added By` group is restricted.

chases2:
How does applying this label differ from a hold? It seems like both would stop this PR from merging.

justaugustus:
> How does applying this label differ from a hold? It seems like both would stop this PR from merging.

The `tide/merge-blocker` label blocks **ALL** merges, not just the PR in question.
For a practical example --> https://github.com/kubernetes/release/issues/1503

What you'll see on active PRs (example: https://github.com/kubernetes/release/pull/1502):
Screenshot from 2020-08-25 17-55-58

What you'll see on Tide (example: https://github.com/kubernetes/website/issues/23399):
Screenshot from 2020-08-25 17-56-38

spiffxp:
>I don't think we want to allow anyone who can apply labels to be able to block Tide (we probably want this restricted to repo admins/writers and approvers)? 

By keeping this to a label that can only be manually applied, we were effectively restricting it to repo admins/writers.  Except now also anyone using Triage or Maintain roles (which I don't think we disallow via automated enforcement, but to my knowledge nobody is using them).  

For the specific use-case that prompted this, maybe the Release Team Docs lead should be added to @kubernetes/website-maintainers to fulfill their duties?

Rather than yet another plugin that restricts to a group of people, I wonder if we have enough prior art to make this a more general purpose thing that can be mixed in? `/override` looks like a good starting point

spiffxp:
_looks wistfully off in the distance at https://github.com/kubernetes/test-infra/pull/8131_

zacharysarah:
@spiffxp 👋 
> For the specific use-case that prompted this, maybe the Release Team Docs lead should be added to @kubernetes/website-maintainers to fulfill their duties?

That seems like a reasonable solution. 👍 Also, much gratitude for `tide/merge-blocker`. It's a significant quality of life improvement for freezing k/website during releases.

/cc @jimangel @kbarnard10 @sftim @kbhawkey @onlydole

PR #18956, opened 2020-08-24
Disable PerformanceDNS test

aojea:
This test was very useful to understand several limits of Kubernetes using KIND
- etcd storage size, https://github.com/kubernetes/kubernetes/issues/94029
- controller manager QPS limit, https://kubernetes.slack.com/archives/C09QZTRH7/p1597992971004900

However, it is very disruptive and affects the other jobs, causing them to fail, so better to skip it

### DIFF
--- a/config/jobs/kubernetes/sig-network/sig-network-kind.yaml
+++ b/config/jobs/kubernetes/sig-network/sig-network-kind.yaml
@@ -87,7 +87,7 @@ periodics:
       - name: FOCUS
         value: \[sig-network\]
       - name: SKIP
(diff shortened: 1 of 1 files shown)


aojea:
/hold cancel
Thanks, I like to have the tesgrid green :)

BenTheElder:
> This test was very useful to understand several limits of Kubernetes using KIND

We should really just increase the storage limit to match GCE and check the QPS limit too, these limits are not KIND specific and are wholly arbitrary.

This test did not teach us about kind limits, it taught us about the GCE cluster config vs kubeadm relying on defaults.

BenTheElder:
We avoided the storage limit for now by changing one other default (the service CIDR), if this test is too disruptive it should be tagged in the test instead.

IMO this is just surfacing that kubeadm's configured limits are too low for test at least, we advertise that kubernetes can handle 10k services and we need to test that.

Issue #18966, opened 2020-08-24
Kettle pypy source no longer supported by BitBucket


MushuEE:
**What happened**:
https://k8s-testgrid.appspot.com/sig-testing-images#kettle started failing with a 403 on `https://bitbucket.org/squeaky/portable-pypy/downloads/pypy3.6-7.1.1-beta-linux_x86_64-portable.tar.bz2` download

This is because BitBucket suspended support for Mercurial? < guessing based off the alert I see.

![image](https://user-images.githubusercontent.com/7189267/91094086-2a7db200-e60f-11ea-8e2b-3a562c892cd0.png)

**What you expected to happen**:
Kettle docker file to build with pypy, without error.
**How to reproduce it (as minimally and precisely as possible)**:
Use `docker build` or `bazel run //images/builder -- --project=k8s-testimages --allow-dirty --build-dir=. kettle/`

**Please provide links to example occurrences, if any**:
https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/post-test-infra-push-kettle/1297936252569391104

**Anything else we need to know?**:
Currently hunting for a replacement maybe from pypy directly? 

/cc @spiffxp

MushuEE:
Verified Passing

PR #18967, opened 2020-08-24
Update kettle pypy source as bitbucket no longer hosts

MushuEE:
Updating the pypy version and source as BB has removed hosting

@clarketm It looks like https://github.com/squeaky-pl/portable-pypy version 7.3.0+ can just use pypy from source because the portable parts were added upstream. So I bumped the version here and changed the source. 

#18966 

/area kettle


MushuEE:
/hold cancel

7.3.1 updated

Issue #18929, opened 2020-08-20
spyglass: Include junit's classname in UI.


EricEdens:
(I'm happy to help with implementation; this request is to discuss whether the feature is desirable, and if so, how it should look when it's finished.)

**Problem**:
spyglass displays the test's name (via junit), but doesn't include its class. Sometimes the same test name is used in multiple classes.

**What would you like to be added**:
The ability to differentiate between tests with the same name. A couple of ideas:
1. In the UI, prepend the class name to the test name.
   - This could happen always, or only when there's a collision.
2. Group results by class name

**Why is this needed**:
The current workaround is to visually inspect the junit XML files.

chizhg:
Big +1 to this.
We also have tests that have the same name but are under different packages, including the class name in the UI would definitely help to differentiate them.

/area prow/spyglass

PR #18970, opened 2020-08-25
Include Junit result class in Junit Lens

MushuEE:
image

#18929
/cc @EricEdens @chizhg 

I think this is all we need for the lens... though I am not sure how to test these 
/assign @cjwagner

### DIFF
--- a/prow/spyglass/lenses/junit/template.html
+++ b/prow/spyglass/lenses/junit/template.html
@@ -29,7 +29,7 @@
         <td colspan="2" style="padding: 0;">
           <table class="failed-layout">
             <tr class="failure-name">
-              <td class="mdl-data-table__cell--non-numeric test-name">{{$firstTest.Name}}&nbsp;<i class="icon-button material-icons arrow-icon">expand_more</i></td>
+              <td class="mdl-data-table__cell--non-numeric test-name">{{$firstTest.ClassName}}:{{$firstTest.Name}}&nbsp;<i class="icon-button material-icons arrow-icon">expand_more</i></td>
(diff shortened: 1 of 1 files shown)


chizhg:
Thanks Grant!
LGTM from my side.
/lgtm

prow/spyglass/lenses/junit/template.html
```diff
@@ -29,7 +29,7 @@
         <td colspan="2" style="padding: 0;">
           <table class="failed-layout">
             <tr class="failure-name">
-              <td class="mdl-data-table__cell--non-numeric test-name">{{$firstTest.Name}}&nbsp;<i class="icon-button material-icons arrow-icon">expand_more</i></td>
+              <td class="mdl-data-table__cell--non-numeric test-name">{{$firstTest.ClassName}}: {{$firstTest.Name}}&nbsp;<i class="icon-button material-icons arrow-icon">expand_more</i></td>
```
chaodaiG, line 32: This only changes test names for failed tests, is it intended?

chaodaiG:
I think you can test with https://github.com/kubernetes/test-infra/blob/master/prow/cmd/deck/runlocal

prow/spyglass/lenses/junit/template.html
MushuEE: @EricEdens did you want all tests to show class or only failing in lens?

cjwagner: I would expect the naming scheme to be consistent across tests regardless of their outcome.

MushuEE: 👍 will update across all outcomes

cjwagner (approved):

PR #18971, opened 2020-08-25
remove mention of StartupProbe feature flag as it is being GA'd

SergeyKanzhelev:
As the StartupProve being GA'd this feature flag needs to be removed.  See PR: https://github.com/kubernetes/kubernetes/pull/94160

This tab was broken for a while as this tab was querying `[NodeAlphaFeature:..]` tests. And this was removed from the test definition for a while now as feature was moved to the beta. 

So this PR can be merged right away. There is very little point in making test show up in this tab again or add this feature flag to the `node-kubelet-serial` tab to make tests work again. Presumably this flag will be removed in master next week.


SergeyKanzhelev:
CC: @matthyx

Issue #18972, opened 2020-08-25
Investigate node-kubelet-serial-alpha tab - why test not showing up and whether we need this tab


SergeyKanzhelev:
**What should be cleaned up or changed**:
The tab [node-kubelet-serial-alpha](https://testgrid.k8s.io/sig-node-kubelet#node-kubelet-serial-alpha) is empty - doesn't run any tests. 

CPUManager tests seems to be matching the definition - has `[Serial]` and `[NodeAlphaFeature:]` [tags](https://github.com/kubernetes/test-infra/blob/bd3b81b6c985971cf10665af396962f1c3136785/config/jobs/kubernetes/sig-node/node-kubelet.yaml#L287):

https://github.com/kubernetes/kubernetes/blob/5708511499fe500ae3b4bbd40204cef382f652e8/test/e2e_node/cpu_manager_test.go#L635-L636

But doesn't show up in the tab and test does not set up the feature flag to run it.

So either need to understand why tests wouldn't run or remove the tab altogether.

SergeyKanzhelev:
Like @alejandrox1 said this test indeed didn't run because:

```
I0824 16:58:49.178] Aug 24 16:58:44.299: INFO: Skipping CPU Manager tests since the CPU capacity < 2
```

SergeyKanzhelev:
So CPU manager tests already configured to run in dedicated tab [kubelet-serial-gce-e2e-cpu-manager](https://testgrid.k8s.io/sig-node-kubelet#kubelet-serial-gce-e2e-cpu-manager). So this tab has no tests to run. I'd suggest to delete this tab altogether.

SergeyKanzhelev:
/close

it was resolved with the #18974

PR #18974, opened 2020-08-25
remove node-kubelet-serial-alpha as it doesn't contain any tests

SergeyKanzhelev:
See some thoughts on it here: https://github.com/kubernetes/test-infra/issues/18972

This tab has no tests to run.

/sig node
/kind cleanup

### DIFF
--- a/config/jobs/kubernetes/sig-node/node-kubelet.yaml
+++ b/config/jobs/kubernetes/sig-node/node-kubelet.yaml
@@ -263,36 +263,6 @@ periodics:
     testgrid-dashboards: sig-node-kubelet
     testgrid-tab-name: node-kubelet-serial
 
-- name: ci-kubernetes-node-kubelet-serial-alpha
-  interval: 3h30m
-  labels:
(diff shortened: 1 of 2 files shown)


bart0sh:
@SergeyKanzhelev thank you for the PR. Good catch!
/lgtm

MHBauer:
Do we intend to never use NodeAlphaFeature any more? Seems like we'd have to put this right back if we use that FOCUS tag again.

Otherwise I'd leave this and add a description that "if it's empty, we don't have any alpha features currently".

SergeyKanzhelev:
> Do we intend to never use NodeAlphaFeature any more? Seems like we'd have to put this right back if we use that FOCUS tag again.
> 
> Otherwise I'd leave this and add a description that "if it's empty, we don't have any alpha features currently".

This tab is `Alpha` features that needs to run `Serial`. I'm fine keeping it if this is the policy. Curious, would adding it back be hard? Should it be commented out with the message to uncomment when needed?

PR #18982, opened 2020-08-25
Update prow playbook

spiffxp:
WIP because many links / context left to add

### DIFF
--- a/docs/playbooks/greenhouse.md
+++ b/docs/playbooks/greenhouse.md
@@ -122,14 +122,14 @@ Consider asking for help from someone familiar with Bazel if you suspect this
 issue.
 
 <!--URLS-->
-[OWNERS]: ./OWNERS 
-[README]: ./README.md 
-[playbooks]: ./../docs/playbooks.md
+[OWNERS]: /greenhouse/OWNERS 
+[README]: /greenhouse/README.md 
+[playbooks]: /docs/playbooks/README.md
 <!--Additional URLS-->
 [bazel]: https://bazel.build/
 [remote-build-cache]: https://docs.bazel.build/versions/master/remote-caching.html
-[deployment.yaml]: ./deployment.yaml
-[service.yaml]: ./deployment.yaml
+[deployment.yaml]: /greenhouse/deployment.yaml
+[service.yaml]: /greenhouse/service.yaml
 [prow-k8s-io]: https://prow.k8s.io
 [bazel#4558]: https://github.com/bazelbuild/bazel/issues/4558
 [velodrome]: http://velodrome.k8s.io/dashboard/db/bazel-cache?refresh=1m&orgId=1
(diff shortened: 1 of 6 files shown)


alvaroaleman:
@spiffxp Can we maybe find a location for this that is not in /prow as this is a prow.k8s.io specific thing and not valid for any prow instance?

spiffxp:
@alvaroaleman yeah I think that's fair, /docs/playboook.md linked here but I think this probably belongs over in /docs

/cc @cjwagner 
WDYT?

spiffxp:
/remove-lifecycle rotten

Will work on getting this out of wip this week

docs/playbooks/greenhouse.md
```diff
@@ -0,0 +1,136 @@
+# Greenhouse Playbook
```
fejta, line 1: Thoughts on making this show up as a move rather than a new file?

fejta (commented):
LGTM

spiffxp:
I redid this as series of commits to just move the greenhouse and prow playbooks to /docs/playbooks, fix any broken links, and add tombstones in case there are links to the old files (I couldn't find anything via https://cs.k8s.io)

So, no new content, until I see what the diffs / commits look like for a reviewer

docs/playbooks/greenhouse.md
spiffxp: The addition of tombstones in the old locations prevents git from seeing this as a move when looking at the entire branch

I'm going to keep the addition of tombstones as the last commit, and you can use the following compare link to see what's actually changed: https://github.com/kubernetes/test-infra/compare/master...spiffxp:doc-prow-monitoring%5E

Or look at just the first commit to see the move

docs/playbooks/prow.md
```diff
@@ -0,0 +1,146 @@
...
+
+- The google.com-owned build cluster for trusted jobs that need access to sensitive secrets
+- Is the kubernetes prow service cluster, under a different name
+
+### k8s-infra-prow-build
+
+- The community-owned prow build cluster
+- Lives in kubernetes.io GCP project k8s-infra-prow-build
+- Infra managed via terraform in k8s.io/infra/gcp/clusters/projects/k8s-infra-prow-build/prow-build
+- Kubernetes manifests live in k8s.io/infra/gcp/clusters/projects/k8s-infra-prow-build/prow-build/resources
```
spiffxp, line 69: These are now out of date
```suggestion
- Infra managed via terraform in k8s.io/infra/gcp/terraform/k8s-infra-prow-build
- Kubernetes manifests live in k8s.io/infra/gcp/terraform/k8s-infra-prow-build/prow-build/resources
```

spiffxp:
Perfect is the enemy of good here, I'm dropping the WIP tag to get this merged.

I don't know that this is all organized in the clearest way, but there's more content about how prow is currently deployed and where to go look to evaluate its health than before.

/cc @ameukam 
I'm specifically excluding k8s-infra-prow.k8s.io for now, we can do that as a followup (or link to other docs for it)

/cc @listx 
Since I believe you've been looking at docs recently

Issue #18983, opened 2020-08-25
ci-test-infra-triage (triage clustering) is failing


BenTheElder:
**What happened**:

https://prow.k8s.io/?job=*triage* is failing consistently

**What you expected to happen**:

triage should be succeeding

**How to reproduce it (as minimally and precisely as possible)**:

**Please provide links to example occurrences, if any**:

**Anything else we need to know?**:

https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/ci-test-infra-triage/1298307826921246720

```
Operation completed over 3 objects/74.1 MiB.                                     
+ gzip -df triage_tests/shard_000000000000.json.gz triage_tests/shard_000000000001.json.gz triage_tests/shard_000000000002.json.gz
+ mkdir -p slices
+ /triage --builds triage_builds.json --tests triage_tests/shard_000000000000.json triage_tests/shard_000000000001.json triage_tests/shard_000000000002.json --output failure_data.json --output_slices slices/failure_data_PREFIX.json --num_workers 7
F0825 17:16:24.059704     141 summarize.go:65] 'PREFIX' not in output_slices flag
goroutine 1 [running]:
k8s.io/klog/v2.stacks(0xc0000a8001, 0xc0000f2000, 0x52, 0xa2)
	/go/pkg/mod/k8s.io/klog/v2@v2.0.0/klog.go:972 +0xb8
k8s.io/klog/v2.(*loggingT).output(0x669500, 0xc000000003, 0x0, 0x0, 0xc0000d64d0, 0x64db91, 0xc, 0x41, 0x0)
	/go/pkg/mod/k8s.io/klog/v2@v2.0.0/klog.go:921 +0x19d
k8s.io/klog/v2.(*loggingT).printf(0x669500, 0x3, 0x0, 0x0, 0x56b62a, 0x22, 0x0, 0x0, 0x0)
	/go/pkg/mod/k8s.io/klog/v2@v2.0.0/klog.go:733 +0x17a
k8s.io/klog/v2.Fatalf(...)
	/go/pkg/mod/k8s.io/klog/v2@v2.0.0/klog.go:1427
k8s.io/test-infra/triage/summarize.parseFlags(0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
	/test-infra/triage/summarize/summarize.go:65 +0x4bc
k8s.io/test-infra/triage/summarize.Main()
[...]

PR #18984, opened 2020-08-25
move tests to arguments

BenTheElder:
We were passing a glob to a flag, in shell this expands to N arguments when there are N matching files so we wind up with (hand converted

`'/triage' '--builds' 'triage_builds.json' '--tests' 'triage_tests/shard_000000000000.json' 'triage_tests/shard_000000000001.json' 'triage_tests/shard_000000000002.json' '--output failure_data.json' '--output_slices slices/failure_data_PREFIX.json' '--num_workers' '7'`

Note particularly: `'--tests' 'triage_tests/shard_000000000000.json' 'triage_tests/shard_000000000001.json' 'triage_tests/shard_000000000002.json'`

Because there's an argument without a flag, go's flag package stops parsing flags, so we fail to set the flags that come after it.

I moved this to the binaries' arguments instead since it's the only repeated flag, and go native flags don't offer repeated flags so we were space splitting.

Another option would be to use a different flag parser and craft the bash to pass `--tests $file` once per file.

### DIFF
--- a/triage/update_summaries.sh
+++ b/triage/update_summaries.sh
@@ -70,10 +70,10 @@ mkdir -p slices
 
 /triage \
   --builds triage_builds.json \
-  --tests triage_tests/*.json \
   --output failure_data.json \
   --output_slices slices/failure_data_PREFIX.json \
   --num_workers 7  # Go incorrectly determines the number of CPUs in a pod, set manually to (2*CPUs-1) corresponding to test-infra-periodics.yaml
+  triage_tests/*.json
 
 gsutil_cp() {
   gsutil -h 'Cache-Control: no-store, must-revalidate' -m cp -Z -a public-read "$@"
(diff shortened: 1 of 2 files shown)


BenTheElder:
fixes: https://github.com/kubernetes/test-infra/issues/18983

triage/update_summaries.sh
```diff
@@ -70,10 +70,10 @@ mkdir -p slices
 
 /triage \
   --builds triage_builds.json \
-  --tests triage_tests/*.json \
   --output failure_data.json \
   --output_slices slices/failure_data_PREFIX.json \
   --num_workers 7  # Go incorrectly determines the number of CPUs in a pod, set manually to (2*CPUs-1) corresponding to test-infra-periodics.yaml
+  triage_tests/*.json
```
michaelkolber, line 76: Are you missing a `\` here?

BenTheElder: we are on the line above :upside_down_face:
:this-is-fine:

hasheddan: Ah sorry for missing that :man_facepalming:

BenTheElder: https://github.com/kubernetes/test-infra/pull/18986

PR #18986, opened 2020-08-25
fix update_summaries.sh missing slash, move num_workers setting to job config

BenTheElder:
fixes: https://github.com/kubernetes/test-infra/pull/18984/files#r476656811
also move --num_workers value out into the job config so we can keep it in sync properly

### DIFF
--- a/config/jobs/kubernetes/test-infra/test-infra-periodics.yaml
+++ b/config/jobs/kubernetes/test-infra/test-infra-periodics.yaml
@@ -36,6 +36,9 @@ periodics:
       env:
       - name: GOOGLE_APPLICATION_CREDENTIALS
         value: /etc/service-account/service-account.json
+      # Go incorrectly determines the number of CPUs in a pod, set manually to (2*CPUs-1) corresponding to test-infra-periodics.yaml
+      - name: NUM_WORKERS
+        value: "7"
       command:
       - "timeout"
       args:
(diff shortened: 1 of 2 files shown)


triage/update_summaries.sh
```diff
@@ -72,7 +72,7 @@ mkdir -p slices
   --builds triage_builds.json \
   --output failure_data.json \
   --output_slices slices/failure_data_PREFIX.json \
-  --num_workers 7  # Go incorrectly determines the number of CPUs in a pod, set manually to (2*CPUs-1) corresponding to test-infra-periodics.yaml
+  ${NUM_WORKERS:+"--num_workers=${NUM_WORKERS}"} \
```
BenTheElder, line 75: this expands to null if NUM_WORKERS isn't set, or `--num_workers=${NUM_WORKERS}` if it is

config/jobs/kubernetes/test-infra/test-infra-periodics.yaml
```diff
@@ -36,6 +36,9 @@ periodics:
       env:
       - name: GOOGLE_APPLICATION_CREDENTIALS
         value: /etc/service-account/service-account.json
+      # Go incorrectly determines the number of CPUs in a pod, set manually to (2*CPUs-1) corresponding to test-infra-periodics.yaml
```
BenTheElder, line 39: this comment is now amusingly self-referential, but I'd prefer to get this fix merged, as it's not overly harmful, just amusing, and triage has been down for a while

michaelkolber: Why did it suddenly stop working?

PR #18988, opened 2020-08-25
Kops - Decrease parallelism from 25 to 15 for ARM64 Conformance

hakman:
... to reduce flakes.
/cc @rifelpet


rifelpet:
We're seeing pods stuck in pending due to insufficient cpu. we'll first try dropping the parallelism and if that doesnt help we can increase the cluster size.

/lgtm

Issue #18652, opened 2020-08-04
release-blocking jobs must run in dedicated cluster: periodic-kubernetes-bazel-test


spiffxp:
Part of https://github.com/kubernetes/test-infra/issues/18549

k8s-infra-prow-build doesn't have RBE, so this job needs to be configured not to use RBE (I proved this works and 
determined the resource limits that should be used in https://github.com/kubernetes/test-infra/pull/18607)

Per https://github.com/kubernetes/test-infra/pull/18613#issuecomment-668178288 it is important we keep some variant of "bazel test" running periodically in the k8s-prow-builds cluster

So, per https://github.com/kubernetes/test-infra/pull/18613#issuecomment-668703640 I suggested we rename jobs:
>- swap the names of the bazel-test jobs such that
>   - periodic-kubernetes-bazel-test-canary = the job that runs on k8s-prow-builds, uses RBE, and is on sig-testing-canaries dashboard
>   - periodic-kubernetes-bazel-test-master = the job that runs on k8s-infra-prow-build and is release-blocking

/sig testing
/sig release
/wg k8s-infra
/area jobs

/assign @ameukam 
assigning since @ameukam started down this path with https://github.com/kubernetes/test-infra/pull/18613

spiffxp:
https://github.com/kubernetes/test-infra/pull/18613 landed 2020-08-05 ~5pm PT
![Screen Shot 2020-08-08 at 1 18 02 PM](https://user-images.githubusercontent.com/49258/89719093-98707b00-d979-11ea-9f2e-5be7ed344c03.png)

It's unclear how much compute the RBE variant of the job ended up consuming, but it looks like we run slower.  This is still well within release-blocking bounds, but good to know.

I missed that we changed the interval when we made this change, we should change that back.  And the release-branch variants need to be changed over as well.

ameukam:
It's unclear to me what's the correct value to set for the interval of the release-branch variants. Is it still necessary to make this change?

spiffxp:
>It's unclear to me what's the correct value to set for the interval of the release-branch variants. Is it still necessary to make this change?

Both jobs have a `fork-per-release-periodic-interval: 6h` annotation set, I think their intervals should be left as-is.

spiffxp:
What needs to be done for the release-branch variants is removal of the `--config=remote` and `--remote_instance_name=...` flags so that their `args` look like the `args` of `periodic-kubernetes-bazel-test-master`

[...]

PR #18997, opened 2020-08-26
Migrate periodic-kubernetes-bazel-test job variants to k8s-infra-prow-build

ameukam:
Ref: https://github.com/kubernetes/test-infra/issues/18652

Adjust resources of the job variants based on the master job.

/assign @spiffxp @BenTheElder 
cc @kubernetes/ci-signal 

/hold


spiffxp:
FYI @liggitt presubmits are still running in RBE, so if we notice the CI variants start behaving wildly differently, this is why.  https://testgrid.k8s.io/sig-release-master-blocking#bazel-test-master has been running this way since 2020-08-05 (ref: https://github.com/kubernetes/test-infra/issues/18652#issuecomment-670970119)

PR #18998, opened 2020-08-26
Correct intervals for periodic-kubernetes-bazel-test jobs

spiffxp:
The canary and main jobs were swapped, but their intervals were not.
The main job should be running more frequently than the canary

This addresses https://github.com/kubernetes/test-infra/issues/18652#issuecomment-670970119

Followup to https://github.com/kubernetes/test-infra/pull/18613
Part of https://github.com/kubernetes/test-infra/issues/18652


spiffxp:
FYI @kubernetes/ci-signal 
/cc @ameukam @BenTheElder

PR #19006, opened 2020-08-26
ConfigUpdater: Allow targeting a named list of clusters

alvaroaleman:
This drastically simplifies the configuration when a number of identical
clusters is targeted.

### DIFF
--- a/prow/ANNOUNCEMENTS.md
+++ b/prow/ANNOUNCEMENTS.md
@@ -119,7 +119,8 @@ state and no claims of backwards compatibility are made for any external API.
 Note: versions specified in these announcements may not include bug fixes made
 in more recent versions so it is recommended that the most recent versions are
 used when updating deployments.
- 
+
+ - *September 2, 2020* The already deprecated `namespace` and `additional_namespaces` settings in the config updater will be removed in October, 2020
  - *August 28, 2020* `tide` now ignores archived repositories in queries.
  - *August 28, 2020* The `Clusters` format and associated `--build-cluster` flag has been removed.
  - *August 24, 2020* The deprecated reporting functionality has been removed from Plank, use crier with `--github-workers=1` instead
--- a/prow/plugins/config.go
+++ b/prow/plugins/config.go
@@ -904,7 +904,7 @@ func (c *ConfigUpdater) SetDefaults() {
 
 	for name, spec := range c.Maps {
 		if spec.Namespace != "" || len(spec.AdditionalNamespaces) > 0 {
-			logrus.Warnf("'namespace' and 'additional_namespaces' are deprecated for config-updater plugin, use 'clusters' instead")
+			logrus.Warn("'namespace' and 'additional_namespaces' are deprecated for config-updater plugin and will be removed in October, 2020, use 'clusters' instead")
 		}
 		// as a result, namespaces will never be an empty slice (namespace in the slice could be empty string)
 		// and clusters will never be an empty map (map[cluster] could be am empty slice)
--- a/prow/plugins/config.go
+++ b/prow/plugins/config.go
@@ -17,6 +17,7 @@ limitations under the License.
 package plugins
 
 import (
+	"encoding/json"
 	"errors"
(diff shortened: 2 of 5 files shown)


prow/plugins/config.go
```diff
@@ -427,10 +428,21 @@ type ConfigMapSpec struct {
...
+
+// A ClusterGroup is a list of clusters with namespaces
+type ClusterGroup struct {
+	Clusters   []string `json:"clusters,omitempty"`
+	Namespaces []string `json:"namespaces,omitempty"`
 }
 
 // ConfigUpdater contains the configuration for the config-updater plugin.
 type ConfigUpdater struct {
+	// ClusterGroups is a map of ClusterGropups that can be used as a target
```
fejta, line 443: spelling: `ClusterGropups`

prow/plugins/config.go
```diff
@@ -427,10 +428,21 @@ type ConfigMapSpec struct {
 	// Clusters is a map from cluster to namespaces
 	// which specifies the targets the configMap needs to be deployed, i.e., each namespace in map[cluster]
 	Clusters map[string][]string `json:"clusters"`
+	// ClusterGroup is a list of named cluster_groups to target. Mutually exclusive with clusters.
```
fejta, line 431: It seems like you could achieve this same effect with a yaml anchor and keep this complexity out of prow's codebase.

Is there a reason to prefer moving this complexity inside prow?

prow/plugins/config.go
```diff
@@ -427,10 +428,21 @@ type ConfigMapSpec struct {
 	// Clusters is a map from cluster to namespaces
 	// which specifies the targets the configMap needs to be deployed, i.e., each namespace in map[cluster]
 	Clusters map[string][]string `json:"clusters"`
```
fejta, line 430: Apparently we now have
* `Namespace`
* `Namespaces`
* `AdditionalNamespaces`
* `Clusters`
* `ClusterGroups`

This seems a little redundant. Thoughts on combining namespace and cluster into context?

prow/plugins/config.go
alvaroaleman: From my understanding, yaml anchors do not work for maps: https://github.com/go-yaml/yaml/issues/184

Generally, I am happy with anything that works

prow/plugins/config.go
alvaroaleman: Yeah, this is super convoluted and would make sense to clean up, however I am a bit concerned that that might break ppl. 

> Thoughts on combining namespace and cluster into context?

`namespaces` is not actually user-settable and `clusters` is a `map[clustername][]clusterNamespace` so pretty much what you meant by `context` already?

fejta: Could we just drop namespace and additional namespaces? Either now or in the future? Thoughts on naming clusters context since it is essentially a context already?

alvaroaleman: > Could we just drop namespace and additional namespaces? 

Funnily they have been deprecated already in october 2019, but without removal timeline. I have updated that to say that they will be removed in october 2020.

> Thoughts on naming clusters context since it is essentially a context already?
Its not completely the same, because a cluster allows to specify multiple namespaces, which a context would normally not allow. Given that and that the renaming is a hassle with little benefits, I would keep that as-is.

prow/plugins/config.go
alvaroaleman: Updated

prow/ANNOUNCEMENTS.md
```diff
@@ -119,7 +119,9 @@ state and no claims of backwards compatibility are made for any external API.
 Note: versions specified in these announcements may not include bug fixes made
 in more recent versions so it is recommended that the most recent versions are
 used when updating deployments.
- 
+<<<<<<< HEAD
```
fejta, line 122: Bad merge

alvaroaleman: Fixed

fejta (approved):

PR #19004, opened 2020-08-26
Update prow's monitoring mixins

spiffxp:
I want to be able to specify a dashboard template / variable of type "custom". The currently used version of grafonnet assumes type "query".  The latest version of grafonnet has many more options available.

I did this via the following:
```sh
cd config/prow/cluster/monitoring/mixins
mv vendor/grafonnet/BUILD.bazel .
make clean-vendor
make install
cp BUILD.bazel vendor/grafonnet
```

The resulting dashboard json that is generated has one new field but is otherwise identical

/cc @cjwagner @stevekuznetsov

Issue #18785, opened 2020-08-11
Kubernetes CI Policy: define metrics/reports that allow us to track whether the situation is getting better


spiffxp:
Part of https://github.com/kubernetes/test-infra/issues/18551

We were experiencing a lot of obvious pain as humans when https://github.com/kubernetes/kubernetes/issues/92937 was opened.

There are a number of theories as to why that pain was being experienced, and we're now acting based on some of those theories.

What we are lacking is:
- metrics that show that pain is being experienced
- metrics that prove the theories as to why the pain was being experienced
- metrics/reports that prove the action we are taking is having positive/negative impact on overall CI health

This issue is intended to cover brainstorming, exploring and implementing metrics / reports that help guide us in the right direction.

Some suggestions / questions I'm pulling up from below
- [ ] can we / does it make sense to implement an alert when nothing has merged into kubernetes/kubernetes for a while, and should have (non-empty tide pool)?
- [ ] can we / does it make sense to implement an alert when the kubernetes/kubernetes non-release-branch tide pool is above a certain threshold? what should that threshold be?
- [ ] can we identify which job runs were due to new commits vs. `/retest` spam?

spiffxp:
"We are experiencing pain" - https://github.com/kubernetes/kubernetes/issues/92937 

- PR's weren't merging into kubernetes/kubernetes' main branch for too long
  - can we have an alert for this?
  - need to account for periods when there's nothing to merge (e.g. weekends, holidays)
- one way we could measure/observe PR's not merging is by looking at tide pool size for kubernetes/kubernetes master
  - plotted at https://monitoring.prow.k8s.io/d/d69a91f76d8110d3e72885ee5ce8038e/tide-dashboard?orgId=1&fullscreen&panelId=5&from=now-90d&to=now
  - only moves at the speed of the slowest presubmit job
- PR's weren't merging due to jobs failing
  - these were occurring in tide's batch jobs
  - these were occurring on each PR in presubmit (e.g. trying to get "all green checkmarks" before tide would try merging)
  - do we have any flake / failure data over time for each of these?
  - can we say "this job is the one that's been causing batches to fail the most"?
  - etc.
- jobs were failing due to (flakes? scheduling errors? resource contention?)
  - jobs that failed due to test failure (flake or legitimate) end up in "failure" state
  - jobs that failed to schedule end up in "error" state

spiffxp:
"Theories as to why the pain is being experienced"

- we suspect we may have encountered resource contention
[...]

PR #19007, opened 2020-08-26
Update plank dashboard to filter and group by more things

spiffxp:
Allow someone to filter and group by:
- cluster
- org
- repo
- state
- type

Add a third graph panel, and change each graph panel to group by
user-selectable labels

I intially had a graph panel for each of these, but:
- it's a lot of clutter
- it appeared to be timing out prometheus at the 90d range

It turns out each template re-queries prometheus (serially) anytime the
dashboard's time-range changes, and only after templates are refreshed do
queries fired off for the dashboard's panels in parallel. This is necessary
because the `label_values` prometheus query is time-range dependent.

To mitigate this I changed the "type" and "state" templates to use a hardcoded
list of values, since I think it unlikely we're going to add new prowjob types
or states.

This required https://github.com/kubernetes/test-infra/pull/19004 to
merge first so I could use `template.custom`

### DIFF
--- a/config/prow/cluster/monitoring/mixins/grafana_dashboards/plank.jsonnet
+++ b/config/prow/cluster/monitoring/mixins/grafana_dashboards/plank.jsonnet
@@ -2,6 +2,7 @@ local grafana = import 'grafonnet/grafana.libsonnet';
 local dashboard = grafana.dashboard;
 local graphPanel = grafana.graphPanel;
 local prometheus = grafana.prometheus;
+local template = grafana.template;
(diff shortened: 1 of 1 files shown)


spiffxp:
Part of https://github.com/kubernetes/test-infra/issues/18785

spiffxp:
e.g.
Screen Shot 2020-08-26 at 4 50 02 PM

spiffxp:
FYI @kubernetes/ci-signal

PR #19009, opened 2020-08-27
Prepare config-forker for unique names in prow

### LINKED ISSUE #19005: status-reconciler gets confused when jobs for different branches have the same name
**What happened**:

We had a job on master which was optional, non-reporting and alwaysrun. It also got forked to two release branches. Finally we made it required and reporting on master. This seemed to confuse status-reconciler. It was comparing the optional release-branch job with the job on master. Therefore it always thought that the last saved state had the job as not required, since it compared the master-job with the release-job.

statusreconciler recreated the jobs on every reconcile loop again and again.

I am not entirely sure in which direction this should be solved. Discussing this with @alvaroaleman and @stevekuznetsov, they tend to think that jobs should just never have the same name. On the other hand kubernetes/sig-release uses this right now, and it seems to be the only way to keep consistent status context namings across branches, since @alvaroaleman meant that duplicating status contexts is prohibited now.

rmohr:
Create unique names on forked jobs and modify the reporting context in
an expected way. If the original context did not include `-master`, keep
it in the original state, to not invent new context values on release
branches. Otherwise replace -master with the version to make the github
context distinct from the master jobs.

Fixes #19005 when using the config-forker. `kubernetes/sig-release` may not have experienced the issue yet, but I think it would be affected by this issue if it would make optional forked jobs suddenly required on master.

### DIFF
--- a/releng/config-forker/main_test.go
+++ b/releng/config-forker/main_test.go
@@ -20,6 +20,8 @@ import (
 	"reflect"
 	"testing"
 
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/util/diff"
 	prowapi "k8s.io/test-infra/prow/apis/prowjobs/v1"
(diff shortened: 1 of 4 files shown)


rmohr:
@davidvossel FYI

releng/config-forker/main_test.go
```diff
@@ -439,6 +439,39 @@ func TestGeneratePresubmits(t *testing.T) {
...
+					Context: "mycontext",
+				},
+			},
+			{
+				JobBase: config.JobBase{
+					Name:        "pull-kubernetes-e2e-replace-context",
+					Annotations: map[string]string{forkAnnotation: "true"},
+				},
+				Brancher: config.Brancher{
+					SkipBranches: []string{`release-\d\.\d`},
```
hasheddan, line 469: Hmm looks like `cmp` does not like the unexported fields in `config.Brancher` for comparison: https://github.com/kubernetes/test-infra/blob/882eb3f17e8e4f1344f7198ee161fb51ba471f2f/prow/config/jobs.go#L215

Do you want to add a custom comparator?

rmohr: On it. thanks.

rmohr: Done.

rmohr:
Let me know if I can do anything to move this PR forward.

justaugustus:
Thanks!
/lgtm
/approve

rmohr:
@justaugustus should I propose the PR again?

PR #19012, opened 2020-08-27
Add sig-cli-master dashboard for all jobs defined in sig-cli-config.yaml

brianpursley:
Some of the test jobs defined in sig-cli-config.yaml are not on the sig-cli-master testgrid dashboard.

This PR adds `sig-cli-master` to the list of dashboards so that SIG-CLI can more easily see the overall status of these jo


soltysh (approved):
/lgtm
/approve

PR #19013, opened 2020-08-27
image-builder: change Azure presubmit to run if capi directory has changed

CecileRobertMichon:
Now that we've collected data over the past few days of the job running in optional mode, and after discussing at office hours, change the job to no longer be optional and run if anything in capi has changed. Also adjusts the max concurrency to 12.

/assign @codenrhoden @moshloop @detiber


codenrhoden:
Thanks @CecileRobertMichon !
/lgtm

Issue #19015, opened 2020-08-27
Bugs in Crier gcsk8sreporter result in stuck pods


alvaroaleman:
**What happened**:

If for whatever reason the Crier gcsk8sreporter  doesn't remove its finalizer but reports success, we end up with stuck pods because nothing ever cleans up that finalizer.

While ending up in this situation in the first place is a bug, we should handle this case better by making something else (sinker?) detect this condition, log an error and clean up.

/area prow
/area prow/crier

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

**Please provide links to example occurrences, if any**:

**Anything else we need to know?**:

spiffxp:
ref: https://github.com/kubernetes/test-infra/pull/18952 for the PR that introduced the finalizer

PR #19018, opened 2020-08-27
Sinker: Cleanup kubernetes reporter finalizer to prevent stuck pods

alvaroaleman:
This PR updates sinker to remove the finalizer from the kubernetes reporter from pods in cases where the reporter didn't do that and won't to it anymore in order to prevent stuck pods.

Fixes https://github.com/kubernetes/test-infra/issues/19015


hasheddan:
@alvaroaleman I was chatting with @RobertKielty this morning about the failures were are seeing in https://testgrid.k8s.io/sig-release-master-blocking#conformance-ga-only where tests are passing and they are not being reported in testgrid. This PR jumped out to me as something that might affect that. Any insight?

alvaroaleman:
Not really. This PR just results in pods with a finalizer getting the finalizer removed. It doesn't change anything about what Plank does. Neither did the original change that introduced the finalizer. Plank is the controller that sets the report status.

spiffxp:
ref: https://github.com/kubernetes/test-infra/pull/18952 for the PR that introduced the finalizer

PR #19020, opened 2020-08-27
Plank: Mark jobs whose pod is deleted as done

alvaroaleman:
Currently its possible that a pod gets deleted, hangs around because of the kubernetes reporter finalizer but the job is not marked as completed, resulting in the pod staying around and the job getting stuck.

This PR changes Plank to mark jobs for which the pod got deleted but still exists as errored.

A sample pod manifest for this can be found here: https://gist.github.com/alvaroaleman/1864aa808958a2cb1cdab4aed869466f

### DIFF
--- a/prow/plank/controller.go
+++ b/prow/plank/controller.go
@@ -415,9 +415,11 @@ func (c *Controller) syncPendingJob(pj prowapi.ProwJob, pm map[string]corev1.Pod
 				}
 				break
 			}
-			// Pod is running. Do nothing.
-			c.incrementNumPendingJobs(pj.Spec.Job)
-			return nil
+			if pod.DeletionTimestamp == nil {
+				// Pod is running. Do nothing.
+				c.incrementNumPendingJobs(pj.Spec.Job)
+				return nil
+			}
 		case corev1.PodRunning:
 			maxPodRunning := c.config().Plank.PodRunningTimeout.Duration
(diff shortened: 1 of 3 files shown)


petr-muller (approved):
/lgtm
/hold

Holding for more reviews, if desired.

prow/plank/controller.go
```diff
@@ -415,9 +415,11 @@ func (c *Controller) syncPendingJob(pj prowapi.ProwJob, pm map[string]corev1.Pod
 				}
 				break
 			}
-			// Pod is running. Do nothing.
-			c.incrementNumPendingJobs(pj.Spec.Job)
-			return nil
+			if pod.DeletionTimestamp == nil {
```
fejta, line 418: Is deleted different from an eviction?

alvaroaleman: Yes, an eviction is indicated in the pod status and does not result in the pod object being deleted.

The only case I am aware of where a cluster could delete a pod is cluster autoscaler downscaling. To prevent that, we have a PDB in openshift (I'll create a PR to add this to the config here as well, because its often used as a reference).

So far, we went by the assumption that if a pod vanishes, this means we have to just re-create it and are good. Since "pod vanishes" is not an expected condition and indicates that someone or something messed with the pods, I would like to make this an error condition, because it can easily hide issues. Should we in the future find other cases where we expect a pod to get deleted by some other entity, we can try to handle those explicitly.

PR #19023, opened 2020-08-27
Crier: Pass on logger onto reporters

alvaroaleman:
Today, all the crier reporters have their own logger that may or may not
have certain fields which makes them extremely hard to debug.

This PR changes the reporter interface to include a logger and makes the
reporters use that.

### DIFF
--- a/prow/crier/controller.go
+++ b/prow/crier/controller.go
@@ -42,9 +42,13 @@ import (
 )
 
 type reportClient interface {
-	Report(pj *v1.ProwJob) ([]*v1.ProwJob, error)
+	// Report reports a Prowjob. The provided logger is already populated with the
+	// prowjob name and the reporter name.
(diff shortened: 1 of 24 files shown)


prow/crier/controller.go
```diff
@@ -219,7 +223,7 @@ func (c *Controller) processNextItem() bool {
 
 	// assert the string out of the key (format `namespace/name`)
 	keyRaw := key.(string)
-	log := logrus.WithField("reporter", c.reporter.GetName()).WithField("prowjob", keyRaw)
+	log := logrus.WithField("reporter", c.reporter.GetName()).WithField("key", keyRaw)
```
alvaroaleman, line 226: `keyraw` is `namespace/name` notation, so using it here as value for `prowjob` is just wrong

petr-muller (approved):
/lgtm
/hold

Holding for more reviews, if needed.

alvaroaleman:
This should be safe to merge as it only changes logging
/hold cancel

PR #19026, opened 2020-08-28
Add kettle unit tests for populate_finish

MushuEE:
Splitting up unit test PRs so that the focus is more singular for reviewers but generally trying to understand kettle via extending unit tests. 

/area Kettle


spiffxp (approved):
/approve
/lgtm

PR #19029, opened 2020-08-28
Make the order of test cases shown in JUnit deterministic

### LINKED ISSUE #18175: JUnit lense changes test-render order randomly
**What happened**:

When I look at a job with multiple failures I see that the rendered tests are randomly changing their order on page reloads.

Best for us would be if they keep the order which they have in the junit.xml. This makes it easy to log additional info for tests without having to create files with very interesting names.

Look for a job with multiple test failures and reload the page multiple times.

Here an example: https://prow.apps.ovirt.org/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/3642/pull-kubevirt-e2e-k8s-1.17/1279367128155164672. Just reload the page a few times.

tosi3k:
Currently the test cases listed in the JUnit lens are ordered in a random sequence per artifact file. This is because of nondeterministic iteration order over maps in Golang - per Golang's documentation:

> The iteration order over maps is not specified and is not guaranteed to be the same from one iteration to the next.

The PR would make the order of the test cases shown in the JUnit lens deterministic and it would reflect the sequence that is contained inside the artifact files.

Fixes #18175

### DIFF
--- a/prow/spyglass/lenses/junit/lens.go
+++ b/prow/spyglass/lenses/junit/lens.go
@@ -145,6 +145,7 @@ func (lens Lens) getJvd(artifacts []api.Artifact) JVD {
 	for _, artifact := range artifacts {
 		go func(artifact api.Artifact) {
 			groups := make(map[testIdentifier][]JunitResult)
+			var testsSequence []testIdentifier
 			result := testResults{
 				link: artifact.CanonicalLink(),
 				path: artifact.JobPath(),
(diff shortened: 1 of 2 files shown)


prow/spyglass/lenses/junit/lens.go
```diff
@@ -145,6 +145,7 @@ func (lens Lens) getJvd(artifacts []api.Artifact) JVD {
 	for _, artifact := range artifacts {
 		go func(artifact api.Artifact) {
 			groups := make(map[testIdentifier][]JunitResult)
+			testsSequence := make([]testIdentifier, 0)
```
matthyx, line 148: Please prefer a slice initialization with var, as suggested [here](https://github.com/golang/go/wiki/CodeReviewComments#declaring-empty-slices):
`var testsSequence []testIdentifier`

matthyx:
> @tosi3k please provide a testcase demontsrating the issue and that your fix works. From a quick glance, I do not see any sorting so its not clear to me how this can fix the described issue.

This should work because the test order is taken from `junit.xml`, let me try to provide a test...

prow/spyglass/lenses/junit/lens.go
tosi3k: Done.

tosi3k:
> @tosi3k please provide a testcase demontsrating the issue and that your fix works. From a quick glance, I do not see any sorting so its not clear to me how this can fix the described issue.

Done. I confirmed that without my change to `lens.go` the unit test fails.

matthyx:
Thanks @tosi3k !

matthyx:
can you add this to the description:
> fixes #18175

tosi3k:
> Thanks @tosi3k !

Thanks for pointing out the place with the unit tests!

Sorry I didn't respond earlier here - I was on vacation for the last two weeks :).

alvaroaleman (approved):
Thanks!

PR #19030, opened 2020-08-28
WIP: kops staging build: Build tagged builds not as CI

justinsb:
Continuous builds have a tag that includes the git sha; for tagged
builds we only want to use the tag.


justinsb:
Looks like we don't need this (currently)

Issue #18853, opened 2020-08-14
Migrate merge-blocking jobs to dedicated cluster: pull-kubernetes-e2e-gce-network-proxy-grpc


spiffxp:
**What should be cleaned up or changed**:

This is part of #18550

To properly monitor the outcome of this, you should be a member of k8s-infra-prow-viewers@kubernetes.io.  PR yourself into https://github.com/kubernetes/k8s.io/blob/master/groups/groups.yaml#L603-L628 if you're not a member.

Migrate pull-kubernetes-e2e-gce-network-proxy-grpc to k8s-infra-prow-build by adding a `cluster: k8s-infra-prow-build` field to the job:
- Make sure any release-branch variants are also updated
- Reference this issue, but do not "fix" it; instead leave this open to monitor the behavior of the job before/after
- Here's an example PR that did this for pull-kubernetes-integration: https://github.com/kubernetes/test-infra/pull/18814

NOTE: migrating this job is not as straightforward as some of the other #18550 issues, because:
- it doesn't run for every single PR
- so in terms of appropriate amount of PR traffic, either manually trigger this job, or look for open PRs that have triggered this job (e.g. using the job-history link below), and explicitly `/test <this-job-name>` on them

Once the PR has merged, note the date/time it merged.  This will allow you to compare before/after behavior.

Things to watch for the job
- https://prow.k8s.io/?job=pull-kubernetes-e2e-gce-network-proxy-grpc
  - does the job start failing more often?
  - does the job start going into error state?
- https://testgrid.k8s.io/presubmits-kubernetes-blocking#pull-kubernetes-e2e-gce-network-proxy-grpc&graph-metrics=test-duration-minutes
  - does the job duration look worse than before? spikier than before?
- https://storage.googleapis.com/k8s-gubernator/triage/index.html?pr=1&job=pull-kubernetes-e2e-gce-network-proxy-grpc
  - do more failures show up than before?
- https://prow.k8s.io/job-history/gs/kubernetes-jenkins/pr-logs/directory/pull-kubernetes-e2e-gce-network-proxy-grpc
  - (can be used to answer some of the same questions as above)
[...]

PR #19033, opened 2020-08-28
Make network-proxy presubmit optional

liggitt:
The job is [solidly failing](https://testgrid.k8s.io/presubmits-kubernetes-blocking#pull-kubernetes-e2e-gce-network-proxy-grpc) and blocking merges (e.g. https://github.com/kubernetes/kubernetes/pull/93095)

It looks like work on the job is in progress (https://github.com/kubernetes/kubernetes/pull/94127), but it should not be blocking.

cc @BenTheElder 
cc @bartsmykla


bartsmykla:
@spiffxp could you help, as I don't have enough permissions?

spiffxp:
@bartsmykla are you going to have time to get to https://github.com/kubernetes/test-infra/issues/18853#issuecomment-678396071 ?

BenTheElder:
I feel like we should have an allowlist config test of blocking kubernetes presubmits to prevent this sort of surprise.

bartsmykla:
@spiffxp apologies, missed that comment. On it

PR #19036, opened 2020-08-28
Plank: Don't overwrite buildIDs

alvaroaleman:
This results in the gcs target dir changing which means the history page
contains jobs that will never get upated.

### DIFF
--- a/prow/plank/controller.go
+++ b/prow/plank/controller.go
@@ -526,14 +526,18 @@ func (c *Controller) syncTriggeredJob(pj prowapi.ProwJob, pm map[string]corev1.P
 }
 
 func (c *Controller) startPod(pj *prowapi.ProwJob) error {
-	buildID, err := c.getBuildID(pj.Spec.Job)
-	if err != nil {
-		return fmt.Errorf("error getting build ID: %v", err)
-	}
+	// Must never be changed, that will mess up reporting. Unfortunately, CRDs
+	// don't support immutable fields yet: https://github.com/kubernetes/kubernetes/issues/65973
+	if pj.Status.BuildID == "" {
+		buildID, err := c.getBuildID(pj.Spec.Job)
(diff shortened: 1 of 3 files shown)


alvaroaleman:
/hold
Today its impossible that two jobs runs have the same buildid. After this change, this could happen if the pod just completely vanishes (its unclear to me why that in turn happens, but it does). After https://github.com/kubernetes/test-infra/pull/18952 this shouldnt be possible anymore I think?

Are we comfortable with two runs possibly having different buildIDs? If not what do we do here? 


/cc @BenTheElder

alvaroaleman:
FTR, this is not a hypothetical case but something that happened here:
* First build id: https://storage.googleapis.com/origin-ci-test/pr-logs/pull/kubevirt_node-maintenance-operator/118/pull-ci-kubevirt-node-maintenance-operator-master-openshift-e2e/1299279202926727168/prowjob.json
* Second build id (there is also a build id label that does not get overwritten): https://storage.googleapis.com/origin-ci-test/pr-logs/pull/kubevirt_node-maintenance-operator/118/pull-ci-kubevirt-node-maintenance-operator-master-openshift-e2e/1299279538907254784/prowjob.json

Alternatively, maybe we can error the job should this ever happen?

alvaroaleman:
Okay, the case where this is expected to happen is if a pod got deleted due to eviction, because we want to handle this transparently by deleting the pod and then re-creating it afterwards: https://github.com/kubernetes/test-infra/blob/2e51e452150284c5dc7a233368fa02dde3373c53/prow/plank/reconciler.go#L334

This messes with the assumption of "A prowjob is a jobrun" because it means that a prowjob is actually a job run request that can have multiple job run instances, in case of eviction. And there are other parts of the system that can not cope well with this but that is probably a different bug.

Issue #18480, opened 2020-07-27
ci-test-infra-benchmark-demo spams triage dashboard


BenTheElder:
**What happened**:

https://storage.googleapis.com/k8s-gubernator/triage/index.html?pr=1#d2fcbd76899acda45017

**What you expected to happen**:

i expect failures in triage to be things we legitimately need to look into

**How to reproduce it (as minimally and precisely as possible)**:

discover that this job exists while looking through our CI failures in go.k8s.io/triage

**Please provide links to example occurrences, if any**:

**Anything else we need to know?**:

BenTheElder:
FYI @liggitt @spiffxp @MushuEE @michaelkolber

BenTheElder:
also .. does a "demo" job need to run every 20 minutes ...?
https://github.com/kubernetes/test-infra/blob/691fd981fd87549368893aa960e2d1c36af4e38e/config/jobs/kubernetes/test-infra/test-infra-canaries.yaml#L15

MushuEE:
Yeah a 10 min and 20 min canary seems excessive, was it designed to be an early/frequent warning system? Looks like it came to be from a big migration

BenTheElder:
We should still do something about this issue.
I forget if we made the change to kettle... Even if we did we should probably reduce the frequency here.

cc @cjwagner @spiffxp

michaelkolber:
We can do what we did with #18322 and add a condition to the SQL query, but I feel like the right way to do it would be to stop this test from reporting its failures to BigQuery? If that's possible.

MushuEE:
So that is Kettle's job, uploading to BQ. We could include a list of exclusions for kettle? And remove them from Triage. Adding them in triage is just pushing them down the stack.

MushuEE:
Also... for that benchmark job. Can we expect Fail/Error? Instead of letting them fail?

MushuEE:
Ah ok so just found https://github.com/kubernetes/test-infra/blob/1f2f32cb9e0bffdb73252025887ffd3b82e2a12a/kettle/make_db.py#L170-L171

I will put up a PR to add both coverage and banchmark

cjwagner:
> also .. does a "demo" job need to run every 20 minutes ...?
> https://github.com/kubernetes/test-infra/blob/691fd981fd87549368893aa960e2d1c36af4e38e/config/jobs/kubernetes/test-infra/test-infra-canaries.yaml#L15

[...]

PR #19037, opened 2020-08-28
Increase benchmark-junit demo period from 20m to 24h.

cjwagner:
ref: https://github.com/kubernetes/test-infra/issues/18480#issuecomment-682961944

The current frequency is excessive.


BenTheElder:
/lgtm
/approve
thanks!

PR #19038, opened 2020-08-28
Add more jobs to kettle's garbage job list

MushuEE:
#18480

This should exclude some more jobs from Triage board / BQ
... I really now want to rewrite this in Golang 

/area kettle
/cc @BenTheElder @michaelkolber @cjwagner 

Also will create a PR to remove #18325 changes and hold till kettle redeployed and verified not uploading Garbage

PR #19039, opened 2020-08-28
Remove job filtering from triage query, should be in kettle

MushuEE:
Reverting #18325 and replacing with #19038

/hold need to update kettle and verify that the change is working prior to merging. 

/cc @michaelkolber

### DIFF
--- a/triage/update_summaries.sh
+++ b/triage/update_summaries.sh
@@ -42,8 +42,7 @@ bq --headless --format=json query --max_rows 1000000 \
   from
     [k8s-gubernator:build.all]
   where
-    timestamp_to_sec(started) > TIMESTAMP_TO_SEC(DATE_ADD(CURRENT_DATE(), -14, 'DAY'))
-    and job != 'ci-kubernetes-coverage-unit'" \
+    timestamp_to_sec(started) > TIMESTAMP_TO_SEC(DATE_ADD(CURRENT_DATE(), -14, 'DAY')) \
   > triage_builds.json
 
 bq query --allow_large_results --headless --max_rows 0 --replace --destination_table k8s-gubernator:temp.triage \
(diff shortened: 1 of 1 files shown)


triage/update_summaries.sh
```diff
@@ -42,8 +42,7 @@ bq --headless --format=json query --max_rows 1000000 \
   from
     [k8s-gubernator:build.all]
   where
-    timestamp_to_sec(started) > TIMESTAMP_TO_SEC(DATE_ADD(CURRENT_DATE(), -14, 'DAY'))
-    and job != 'ci-kubernetes-coverage-unit'" \
+    timestamp_to_sec(started) > TIMESTAMP_TO_SEC(DATE_ADD(CURRENT_DATE(), -14, 'DAY')) \
```
michaelkolber, line 45: This is missing a closing quote.

```suggestion
    timestamp_to_sec(started) > TIMESTAMP_TO_SEC(DATE_ADD(CURRENT_DATE(), -14, 'DAY'))" \
```

MushuEE: thanks

spiffxp (commented):
I'm not sure how I feel about this.  Even if we're filtering certain jobs out in kettle going forward, there is noisy data already in this dataset in the past.  You wouldn't want to deploy this until kettle has filtered data out for at least 14 days.

Whereas if we decided there are more noisy jobs that should be filtered out, we could make the change here and have it take effect immediately.

michaelkolber:
I feel like it would make more sense to have it in a single location, regardless of where it is. It's more efficient to have it filtered through kettle, and it's faster through here, but I can definitely see someone in the future trying to figure out why their job isn't appearing in Triage only to find out that it was filtered out in a different place than they thought.

Issue #18267, opened 2020-07-10
benchmarkjunit: allow streaming output to stdout


howardjohn:
**What would you like to be added**:

It would be nice if benchmarkjunit can stream output to stdout. For example, something like `go run ./pkg/benchmarkjunit ./experiment/dummybenchmarks -l -` where output goes to stdout, and its not buffered.

**Why is this needed**:

This way when running we can see progress, rather than waiting for the entire test to finish.

BenTheElder:
cc @cjwagner

cjwagner:
Here's a PR: https://github.com/kubernetes/test-infra/pull/19042

PR #19042, opened 2020-08-28
Add support for streaming benchmark output while running benchmarkjunit.

cjwagner:
It is now possible to specify `--log-file -` or `-l -` to stream benchmark output while the benchmarks are being run. Note that junit results will only be omitted once the benchmarks complete.
I tested this locally and it works as expected.

fixes https://github.com/kubernetes/test-infra/issues/18267
/assign @howardjohn

### DIFF
--- a/pkg/benchmarkjunit/main.go
+++ b/pkg/benchmarkjunit/main.go
@@ -17,9 +17,12 @@ limitations under the License.
 package main
 
 import (
+	"bytes"
 	"encoding/xml"
 	"fmt"
+	"io"
 	"io/ioutil"
+	"os"
 	"os/exec"
 
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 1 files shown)


pkg/benchmarkjunit/main.go
```diff
@@ -66,11 +69,23 @@ func run(opts *options, args []string) {
...
+		testErr = testCmd.Run()
+		testOutput = buf.Bytes()
+	} else {
+		testOutput, testErr = testCmd.CombinedOutput()
+	}
 	if testErr != nil {
 		logrus.WithError(testErr).Error("Error(s) executing benchmarks.")
 	}
-	if len(opts.logFile) > 0 {
+	if len(opts.logFile) > 0 && opts.logFile != "-" {
```
MushuEE, line 88: Nit: maybe move this to a switch, might also make it more flexible if more options are added later on?
```
case opts.logFile == "-":
		// Stream command output to stdout.
		var buf bytes.Buffer
		writer := io.MultiWriter(os.Stdout, &buf)
		testCmd.Stdout = writer
		testCmd.Stderr = writer
		testErr = testCmd.Run()
		testOutput = buf.Bytes()
case len(opts.logFile) > 0:
       testOutput, testErr = testCmd.CombinedOutput()
       if err := ioutil.WriteFile(opts.logFile, testOutput, 0666); err != nil {
			logrus.WithError(err).Fatalf("Failed to write to log file %q.", opts.logFile)
       }
```

MushuEE: nvm, that doesn't quite work with `if testErr != nil`

PR #19051, opened 2020-08-28
Tweaking the rw permissions on the temporary bucket to drill down on …

geojaz:
…kops/GCE failures


geojaz:
This continues with effort in #19031

geojaz:
@rifelpet this is a punt with way too permission to get a test to run. :)

rifelpet:
@geojaz It looks like the gce tests are now launching the cluster successfully, only a few e2e test failures:

https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/e2e-kops-gce-latest/1300263636203212804

is this still needed?

geojaz:
@rifelpet woot! let's work on those errors instead 😄 
/close

PR #19053, opened 2020-08-29
Crier github reporter: Avoid lock contention issues

alvaroaleman:
This PR:
* Extends crier so reporters can defer their reporting ( Same commit as in https://github.com/kubernetes/test-infra/pull/19048)
* Fixes a lock contention issue in the github reporter:

    Crier github reporter: Don't block workers waiting for lock
    
    In 00562873e0c597be9d186ebc3887c3fa0690400c we introduced PR-Level
    locking for presubmits to the crier github reporter in order to allow
    running it with multiple workers without having the workers race with
    each other when creating the GitHub comment.
    
    The implementation naively used a map of sync.Mutex, blocking workers
    that are waiting for their lock. This can result in all workers being
    blocked, which is worse than the initial problem of only having one
    worker.
    
    This commit updates that to instead use a semaphore.Weightened and
    TryAcquire, returning a RequeueAfter if the lock is occupied in order to
    minimize lock contention.

### DIFF
--- a/prow/crier/BUILD.bazel
+++ b/prow/crier/BUILD.bazel
@@ -47,11 +47,13 @@ go_test(
     embed = [":go_default_library"],
     deps = [
         "//prow/apis/prowjobs/v1:go_default_library",
+        "@com_github_google_go_cmp//cmp:go_default_library",
         "@com_github_sirupsen_logrus//:go_default_library",
         "@io_k8s_apimachinery//pkg/runtime:go_default_library",
         "@io_k8s_apimachinery//pkg/types:go_default_library",
         "@io_k8s_sigs_controller_runtime//:go_default_library",
         "@io_k8s_sigs_controller_runtime//pkg/client:go_default_library",
         "@io_k8s_sigs_controller_runtime//pkg/client/fake:go_default_library",
+        "@io_k8s_sigs_controller_runtime//pkg/reconcile:go_default_library",
     ],
 )
--- a/prow/crier/controller.go
+++ b/prow/crier/controller.go
@@ -39,7 +39,8 @@ import (
 type ReportClient interface {
 	// Report reports a Prowjob. The provided logger is already populated with the
 	// prowjob name and the reporter name.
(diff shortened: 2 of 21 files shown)


MushuEE:
Cool! Is there any concern over a worker getting starved out just by happen chance that every 5 seconds the locks are taken? I don't know much about Crier so it might not be a big deal if a lock is not taken for a while.

alvaroaleman:
> Cool! Is there any concern over a worker getting starved out just by happen chance that every 5 seconds the locks are taken? I don't know much about Crier so it might not be a big deal if a lock is not taken for a while.

If the locks are taken, the Reconciliation will return with a `RequeueAfter: 5 * time.Seconds`. This puts the given work item (a prowjob) back into the `workqueue` and doesn't hand it out to workers in this timeframe. If there is any other item in the workqueue, the workers will take it. Does that answer your question (I am not sure if I understood it correctly)?

MushuEE:
Thanks @alvaroaleman, I was more curious if the same job could get put back in the queue multiple times in a row due to happenstance where resources are locked each time it is pulled off the queue. So the same job keeps getting pushed back after 5 seconds say, 10 times in a row. I am not sure if that ~50 second wait will impact anything.

alvaroaleman:
> Thanks @alvaroaleman, I was more curious if the same job could get put back in the queue multiple times in a row due to happenstance where resources are locked each time it is pulled off the queue. So the same job keeps getting pushed back after 5 seconds say, 10 times in a row. I am not sure if that ~50 second wait will impact anything.

Ah. So yeah, this is possible but that would mean that either the reporting of a different job for the same PR takes very long or that in those 50 seconds we don't get our turn but other jobs for the same PR do. The five seconds are intended to approximate the duration of a single successful report. Admittedly, its a somewhat pessimistic assumption (and we seem to not have metrics for this, we only have them for non-mutating requests :/ ). Maybe reduce it to two seconds, wdyt?

MushuEE:
I default to your expertise. It sounds like the likelihood of major delay is low and even if there is delay, it isn't high impact. I really was more curious for me own sake of understanding our systems a bit better. Quite a lot of moving parts to ramp up with! Thanks!

/lgtm

PR #19055, opened 2020-08-29
Return error when no artifacts found during fetching in spyglass

### LINKED ISSUE #18999: podinfo Body() called with no artifacts, which should never happen.
**What happened**:
```yaml
{
  component: "deck"   
  file: "prow/spyglass/lenses/podinfo/podinfo.go:82"   
  func: "k8s.io/test-infra/prow/spyglass/lenses/podinfo.Lens.Body"   
  level: "error"   
  msg: "podinfo Body() called with no artifacts, which should never happen."   
}
```
For this to never happen

/help

habibrosyad:
Fixes #18999

### DIFF
--- a/prow/spyglass/spyglass.go
+++ b/prow/spyglass/spyglass.go
@@ -403,9 +403,14 @@ func (sg *Spyglass) RunToPR(src string) (string, string, int, error) {
 // ExtraLinks fetches started.json and extracts links from metadata.links.
 func (sg *Spyglass) ExtraLinks(ctx context.Context, src string) ([]ExtraLink, error) {
 	artifacts, err := sg.FetchArtifacts(ctx, src, "", 1000000, []string{prowv1.StartedStatusFile})
+	// Failing to parse src, that's an error.
+	if err != nil {
+		return nil, err
+	}
+
 	// Failing to find started.json is okay, just return nothing quietly.
-	if err != nil || len(artifacts) == 0 {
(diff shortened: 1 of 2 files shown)


prow/spyglass/spyglass.go
```diff
@@ -404,7 +404,7 @@ func (sg *Spyglass) RunToPR(src string) (string, string, int, error) {
 func (sg *Spyglass) ExtraLinks(ctx context.Context, src string) ([]ExtraLink, error) {
 	artifacts, err := sg.FetchArtifacts(ctx, src, "", 1000000, []string{prowv1.StartedStatusFile})
 	// Failing to find started.json is okay, just return nothing quietly.
-	if err != nil || len(artifacts) == 0 {
+	if err != nil {
 		logrus.WithError(err).Debugf("Failed to find started.json while looking for extra links.")
```
MushuEE, line 408: It seems that there are other errors that can be returned from `FetchArtifacts` should it always be saying "Failed to find started.json..."? 
`return arts, fmt.Errorf("error parsing src: %v", err)` < is that also "okay, just return nothing"?
If so we might want to update the comment?

habibrosyad: For the parsing error I do think that it needs to return error instead of nothing. I wonder if it's intended or not, will check it later.

Imho, the proper fix would be to return error on parsing failure, and return nothing when no artifacts found (with defined error).

habibrosyad: Done in d56b996, could you review again @MushuEE?

PR #19062, opened 2020-08-31
prow/plugins/override: normalize user before looking up in OWNERS

### LINKED ISSUE #15043: Override not allowed, even though username is listed in OWNERS
**What happened**:
`@LorbusChris: LorbusChris unauthorized: /override is restricted to Repo administrators, approvers in top level OWNERS file.`

see https://github.com/openshift/machine-config-operator/pull/1220#issuecomment-548383786

Prow let's me do the override 

In MCO repo, it's reproducible for my user on all PRs, another example: https://github.com/openshift/machine-config-operator/pull/1229#issuecomment-548364714

https://github.com/openshift/machine-config-operator/pull/1229#issuecomment-548364714
https://github.com/openshift/machine-config-operator/pull/1220#issuecomment-548383786

Maybe it's casued by the upper case letters in my username?

steveeJ:
- [x] Add a test case to verify the failure
- [x] Fix the failure

Fixes #15043.

### DIFF
--- a/prow/plugins/override/override_test.go
+++ b/prow/plugins/override/override_test.go
@@ -598,6 +598,27 @@ func TestHandle(t *testing.T) {
 			},
 		},
 		{
+			name:      "override with allow_top_level_owners works for uppercase user",
+			comment:   "/override job",
+			user:      "Code_owner",
+			options:   plugins.Override{AllowTopLevelOwners: true},
+			approvers: []string{"code_owner"},
+			contexts: map[string]github.Status{
+				"job": {
+					Context:     "job",
+					Description: "failed",
+					State:       github.StatusFailure,
+				},
+			},
+			expected: map[string]github.Status{
+				"job": {
+					Context:     "job",
+					Description: description("Code_owner"),
+					State:       github.StatusSuccess,
+				},
+			},
+		},
+		{
 			name:    "override with allow_top_level_owners fails if user is not in OWNERS file",
 			comment: "/override job",
 			user:    "non_code_owner",
--- a/prow/plugins/override/override.go
+++ b/prow/plugins/override/override.go
@@ -204,7 +204,7 @@ func authorizedTopLevelOwner(oc ownersClient, allowTopLevelOwners bool, log *log
 			log.WithError(err).Warnf("cannot determine whether %s is a top level owner of %s/%s", user, org, repo)
 			return false
 		}
-		return owners.TopLevelApprovers().Has(user)
+		return owners.TopLevelApprovers().Has(github.NormLogin(user))
 	}
 	return false
 }


alvaroaleman:
@steveeJ I'll unassign myself for now, can you assign me once this is ready for a review? Thanks!
/uncc

steveeJ:
@alvaroaleman Thanks for approving the tests. The test case I added failed as expected and demonstrates the bug:

```
    --- FAIL: TestHandle/override_with_allow_top_level_owners_works_for_uppercase_user (0.00s)
        override_test.go:740: bad statuses: actual map[string]github.Status{"job":github.Status{State:"failure", TargetURL:"", Description:"failed", Context:"job"}} != expected map[string]github.Status{"job":github.Status{State:"success", TargetURL:"", Description:"Overridden by code_owner", Context:"job"}} 
```

I'll add a commit which fixes this case to this PR shortly.

steveeJ:
This is ready for review.

/assign @alvaroaleman

prow/plugins/override/override_test.go
```diff
@@ -597,6 +597,27 @@ func TestHandle(t *testing.T) {
 				},
 			},
 		},
+		{
+			name:      "override with allow_top_level_owners works for uppercase user",
```
alvaroaleman, line 601: Whats with the other way round, the user in the owners file is not all lowercase?

steveeJ: The test function currently differs from the tested code in that aspect. The real code effectively runs `github.NormLogin` on all users which are loaded from the _OWNERS_ file. The test doesn't do that. Being aware of that we don't need to test the case you're mentioning, although I'm open to rewriting the test function to be more similar to the tested code.

alvaroaleman: okay sgtm. Can youy sign the CLA and squash your commits?

steveeJ: I squashed the first two commits and added a fixup commit on top that makes the changes to the test function as described above. I think it makes sense to keep that commit and squash it as well, WDTY?
Meanwhile I'll read up on the CLA signature.

alvaroaleman: What is a fixup commit? I don't think there is value in having the testcase declare an uppercase user just so the test definition normalizes it, that is effectively just testcode testing testcode

steveeJ: > What is a fixup commit?

It's what I call a commit that was created using `git commit --fixup=<some-rev>`.

> I don't think there is value in having the testcase declare an uppercase user just so the test definition normalizes it, that is effectively just testcode testing testcode

I see that point. Unfortunately the real code is not suited to be called from the test function and I'm currently not motivated to refactor that. I'll drop the fixup commit then.

PR #19069, opened 2020-08-31
Migrate pull-kubernetes-bazel-test-canary to k8s-infra

spiffxp:
Start with the resources used by the RBE variant of the job

Part of https://github.com/kubernetes/test-infra/issues/19070


spiffxp:
FYI @kubernetes/ci-signal 
/cc @BenTheElder

PR #19071, opened 2020-08-31
Revert tag-match flag for image builder

damemi:
This flag was introduced in #18506 and later attempted to be fixed in #18540. However,
this command won't work with Go's cmd libraries as this Git flag takes a `glob`, which are
not expanded by Go.

While this would be a nice feature to have, we should continue searching for a way to implement it.
Though in order to get image builds working for the Descheduler again, we need to remove
it for now.


damemi:
/cc @spiffxp 
Could you please take a look at this? Thank you for approving the other PRs to add it but it seems unfortunately too broken for me to figure out right now (this is currently blocking the descheduler release because we have no recent image builds to promote).

A quick search looks like no one else started using it, so it should be safe to remove (though if anyone was using it, their builds would also be broken).
/priority critical-urgent

PR #19074, opened 2020-08-31
Add a PDB to prevent cluster autoscaler from evicting test workloads

alvaroaleman:
IIRC the prow.k8s.io build cluster doesn't use the cluster autoscaler but I think its still good to put the PDB in its config, because its often used as a reference by other prow setups (and without CA its just a no-op).

/assign @fejta

### DIFF
--- a/config/prow/cluster/test_pods_pdb.yaml
+++ b/config/prow/cluster/test_pods_pdb.yaml
@@ -1,0 +1,31 @@
+# Copyright 2020 The Kubernetes Authors All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# The purpose of the PodDisruptionBudget here is to never allow evicting pods created by prow.
+# Eviction of pods can happen for one of two reasons:
+# * Cluster autoscaler downscaling
+# * Someome/Something using `kubectl drain`
+#
+# It is still possible to delete the pods via a normal delete call. See https://kubernetes.io/docs/tasks/administer-cluster/safely-drain-node/#the-eviction-api
+
+apiVersion: policy/v1beta1
+kind: PodDisruptionBudget
+metadata:
+  name: prow-pods
+  namespace: test-pods
+spec:
+  maxUnavailable: 0
+  selector:
+    matchLabels:
+      created-by-prow: "true"
--- a/config/prow/cluster/BUILD.bazel
+++ b/config/prow/cluster/BUILD.bazel
@@ -77,6 +77,15 @@ release(
         MULTI_KIND,
         cluster = BUILD_CLUSTER,
     ),
+    component(
+        "test_pods",
+        "pdb",
+    ),
+    component(
+        "test_pods",
+        "pdb",
+        cluster = BUILD_CLUSTER,
+    ),
 )
 
 # TODO(fejta): do we want to auto-apply this?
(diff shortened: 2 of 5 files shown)


config/prow/cluster/test_pods_pdb.yaml
```diff
@@ -0,0 +1,31 @@
...
+
+# The purpose of the PodDisruptionBudget here is to never allow evicting pods created by prow.
+# Eviction of pods can happen for one of two reasons:
+# * Cluster autoscaler downscaling
+# * Someome/Something using `kubectl drain`
+#
+# It is still possible to delete the pods via a normal delete call. See https://kubernetes.io/docs/tasks/administer-cluster/safely-drain-node/#the-eviction-api
+
+apiVersion: policy/v1beta1
+kind: PodDisruptionBudget
```
fejta, line 23: Need to declare the usage of this yaml file, something like:

https://github.com/kubernetes/test-infra/blob/1606c2c0f03778659bdb25e72c390f673b0597a4/config/prow/cluster/BUILD.bazel#L74-L78

(and also one in the service cluster, which is where we run test-infra-trusted jobs)

alvaroaleman:
> Can we just have plank automatically create/manage this pod disruption budget?

But plank is the wrong entity for this, this would basically belong into a `prow-operator`, if there was one

cjwagner:
I agree with Alvaro, but also agree with Erick's point about the current inconvenience of needing to deploy this to every build cluster.
While a prow-operator definitely seems like the most correct place to put this logic, I don't think it would be harmful for it to exist in plank until we have a prow-operator. Incorporating this into a binary rather than just our deployment automation will also have the nice side effect of being more likely to be adopted by existing instances. What do you think Alvaro?

alvaroaleman:
>I agree with Alvaro, but also agree with Erick's point about the current inconvenience of needing to deploy this to every build cluster.
While a prow-operator definitely seems like the most correct place to put this logic, I don't think it would be harmful for it to exist in plank until we have a prow-operator. Incorporating this into a binary rather than just our deployment automation will also have the nice side effect of being more likely to be adopted by existing instances. What do you think Alvaro?

We already require rbac that also has changed regularly in the past to be deployed for each and every build cluster context. Just like rbac, the PDB is a dependency of plank (albeit a weaker one as it doesn't stop working, it just may result in interruptions). I don't really see why the PDB should get a special treatment. Furthermore, I heavily dislike the idea of mixing layers here because:

* It makes reasoning hard, as the component suddenly sets up _some_ of its dependencies by itself
* It requires plank to have too many privileges (we have to add a dependency (more rbac) so it can set up a different dependeny by itself)
* Its guaranteed to violate expectations (or result in super weird code) because ppl rightfully expect a controller to constantly reconcile whatever it reconciles, which plank won't do for a PDB because it only reacts to pods and prowjobs

fejta:
Keeping it out of plank seems like the right separation of concerns.

We could do this with a separate controller/operator: read the kubeconfig to determine all the contexts, apply the desired pdb in each context.

alvaroaleman:
> We could do this with a separate controller/operator: read the kubeconfig to determine all the contexts, apply the desired pdb in each context.

I don't disagree, but thats a lot more work than adding a manifest and its work that doesn't really have a great ROI (because we already have existing mechanics to deploy things to all build clusters downstream and I would expect prow.k8s.io to have that as well, otherwise build clusters would break frequently)

cjwagner:
Alvaro's reasoning makes sense to me. I'm happy to stamp this after this comment is addressed: https://github.com/kubernetes/test-infra/pull/19074#discussion_r481304237

config/prow/cluster/test_pods_pdb.yaml
alvaroaleman: Fixed, PTAL

cjwagner: We still need a copy for the service cluster where trusted jobs run.

alvaroaleman: Can you point me to a sample as how to do that? I tried to do it by copying the rule without the `cluster = BUILD_CLUSTER` argument, but that always failed with a rule name clash

fejta: Looks like the abstraction doesn't currently support this https://github.com/kubernetes/test-infra/blob/e04e876f2a35a7770af7ee0ec500301b1ae830c6/prow/def.bzl#L155-L162

Each thing needs to have a unique name, but right now the name is a function of the file name. You could add a `clusters=` And then do for cluster in clusters: n = cluster+"_"+n or something

PR #19076, opened 2020-08-31
Migrate kubernetes kubemark presubmits to k8s-infra-prow-build

spiffxp:
Drop the canary job since it's proven to work (ref: https://github.com/kubernetes/kubernetes/pull/92316#issuecomment-684007312)

Followup to https://github.com/kubernetes/test-infra/pull/19058
Part of https://github.com/kubernetes/k8s.io/issues/851


spiffxp:
FYI @kubernetes/ci-signal

PR #18904, opened 2020-08-19
migrates merge-blocking job to k8s-infra-prow-build

RobertKielty:
fixes #18854
  migrated job is pull-kubernetes-e2e-gce-network-proxy-http-connect

PR #18889, opened 2020-08-18
Migrating e2e-gce-network-proxt-grpc job

bartsmykla:
Migration of the job to the new cluster: `k8s-infra-prow-build` as a
result of https://github.com/kubernetes/test-infra/issues/18853

Signed-off-by: Bart Smykla <bsmykla@vmware.com>

PR #19079, opened 2020-09-01
Drop --gcp-project= flag from kubernetes network-proxy presubmits

spiffxp:
Part of https://github.com/kubernetes/test-infra/issues/18853
Followup to https://github.com/kubernetes/test-infra/pull/18889

Let's find out if the gce-project pool in k8s-infra-prow-build meets the needs of these jobs. 

Fixing grpc variant directly (ref: https://github.com/kubernetes/test-infra/issues/18853#issuecomment-678396071). If this works, we should revert https://github.com/kubernetes/test-infra/pull/19033

Trying out http-connect variant via a canary (ref: https://github.com/kubernetes/test-infra/pull/18904#pullrequestreview-479180944)

### DIFF
--- a/config/jobs/kubernetes/sig-api-machinery/sig-api-machinery-config.yaml
+++ b/config/jobs/kubernetes/sig-api-machinery/sig-api-machinery-config.yaml
@@ -121,6 +121,51 @@ presubmits:
             memory: 6Gi
     annotations:
       testgrid-dashboards: sig-api-machinery-network-proxy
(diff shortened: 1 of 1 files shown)


spiffxp:
FYI @kubernetes/ci-signal

spiffxp:
Made the canary job optional so tests don't yell at me to put it on kuberntes-presubmits-blocking testgrid

Issue #19080, opened 2020-09-01
(ci|pull)-kubernetes-kind-e2e.* jobs are failing


spiffxp:
**What happened**:
- https://testgrid.k8s.io/sig-release-1.19-blocking#kind-1.19-parallel&width=20 
  - 2020-08-31 11:56 PDT - started failing continuously 
  - 2020-08-28 13:47 PDT started seeing intermitted incompletes ("still running" according to spyglass) 
- https://testgrid.k8s.io/sig-release-1.19-blocking#kind-ipv6-1.19-parallel&width=20
  - 2020-08-31 11:34 PDT - started failing continuously 
  - 2020-08-28 15:24 PDT started seeing intermitted incompletes ("still running" according to spyglass) 

Seeing a similar pattern for all other kind release-blocking jobs.

Some, but not all of the failures, are of the variety "There are no nodes that your pod can schedule to" according to spyglass. 

I suspect what is happening is the following:
- these jobs are demanding about an entire node's worth of CPU, so any nodes that have even one job running have insufficient CPU
- cluster autoscaling spins up new nodes to meet the demand
- new nodes become available, but other jobs with smaller resource requests get scheduled on them, so they have insufficient CPU
- pod is eventually terminated by plank due to remaining in Pending for too long

**What you expected to happen**:

I expected cluster autoscaling to spin up new nodes if none were available with sufficient capacity for these jobs.

**How to reproduce it (as minimally and precisely as possible)**:

**Please provide links to example occurrences, if any**:
e.g. for https://testgrid.k8s.io/sig-release-master-blocking#kind-master-parallel&width=20
- https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-kind-e2e-parallel/1300592620015194112 - 3 tests fail
- https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-kind-e2e-parallel/1300577268589924352 - There are no nodes that your pod can schedule to - check your requests, tolerations, and node selectors (skip schedule deleting pod: test-pods/332044eb-ebe2-11ea-aff6-5a14b0d7f3dc)
[...]

PR #19081, opened 2020-09-01
Tune release-blocking kind job resources down

spiffxp:
The kind release-blocking prowjobs have been ending up in error
state over the past few week days, often due to pods being unable
to schedule due to insufficient CPU. We've been seeing increased
PR traffic due to v1.19 getting released, and a concerted attempt
to clear the backlog of v1.20 PRs that were waiting for code thaw.

The kind release-blocking jobs were the only jobs migrated to
k8s-infra-prow-build that reserved 7300m CPU. The next highest
cpu limit is 7, used by the verify presubmit and bazel-test periodic
jobs. These jobs have not been ending up in error state.

I was originally going to match the kind presubmit jobs, which use
4 cpu, but they have been failing a lot today during times I would
correlate with high PR volume. I'm still tuning the memory to match
based on graphs in the issue linked to this PR.

Hoping to address https://github.com/kubernetes/test-infra/issues/19080

### DIFF
--- a/config/jobs/kubernetes-sigs/kind/kind-release-blocking.yaml
+++ b/config/jobs/kubernetes-sigs/kind/kind-release-blocking.yaml
@@ -46,14 +46,13 @@ periodics:
         privileged: true
       resources:
         limits:
-          memory: 14Gi
-          cpu: 7300m
+          memory: 9Gi
+          cpu: 7
         requests:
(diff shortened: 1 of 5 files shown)


spiffxp:
/cc @BenTheElder 
FYI @kubernetes/ci-signal

Issue #18866, opened 2020-08-14
Plank crash loops on bad build cluster instead of continuing


michelle192837:
Seeing fatal errors for plank due to a bad build cluster in kubeconfig. Error message `"Error creating build cluster clients." `

Instead of failing to start/falling over, plank should log the error in creating a build cluster client and continue.

/area/prow
/area/prow/plank
/kind oncall-hotlist

PR #19088, opened 2020-09-01
Change defunct kubeconfig error level to Error

chases2:
Changes error level for issues caused similar to #18866

The change was also moved to the Prow controller manager, since it looked similar. What does that do differently, and should we be moving to that from plank? I didn't see a readme for that component, so I'm not sure.

/assign @alvaroaleman

### DIFF
--- a/prow/cmd/plank/main.go
+++ b/prow/cmd/plank/main.go
@@ -118,7 +118,7 @@ func main() {
 
 	buildClusterClients, err := o.kubernetes.BuildClusterUncachedRuntimeClients(o.dryRun)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


alvaroaleman:
/lgtm


> The change was also moved to the Prow controller manager, since it looked similar. What does that do differently, and should we be moving to that from plank? I didn't see a readme for that component, so I'm not sure.

Its mostly a more modern implementation and yes, it should be used, I did create a PR for that already but oncall will need to have a look when it gets rolled out in case I missed something: https://github.com/kubernetes/test-infra/pull/18949

PR #19090, opened 2020-09-01
Sinker: Use cache for pods

alvaroaleman:
This improves the efficiency and speed of sinker and reduces the load put on build cluster api, because we don't do a list every sync iteration anymore.


stevekuznetsov (approved):

prow/cmd/sinker/main.go
```diff
@@ -137,21 +137,28 @@ func main() {
...
 
-	podClients := map[string]podInterface{}
-	for cluster, client := range buildClusterClients {
-		podClients[cluster] = client
+	buildClusterClients := map[string]ctrlruntimeclient.Client{}
+	for clusterName, buildManager := range buildManagers {
+		if err := mgr.Add(buildManager); err != nil {
+			logrus.WithError(err).Fatal("Failed to add build cluster manager to main manager")
+		}
+		buildClusterClients[clusterName] = mgr.GetClient()
```
cjwagner, line 154: I'm pretty sure this should be `buildManager.GetClient()` rather than `mgr.GetClient()`. It is still unclear to me why Openshift's Prow instance doesn't have issues with this, but k8s-prow does.

alvaroaleman: Openshifts Prow instance did have the same issue, but for some reason our alert didn't get triggered

stevekuznetsov: Our `sinker` *was* deleting .. something? From Friday:

```
$ oc --context app.ci --namespace ci logs --tail=-1 --selector component=sinker | grep 'Deleted old completed pod' | jq .cluster --raw-output | sort | uniq
api.ci
app.ci
build01
build02
vsphere
```

alvaroaleman: Yeah, as cole pointed out it did get a client for all clusters, however the client was always the client to the cluster the prowjobs are in but stored under all build clusters names

stevekuznetsov: Oh, and of course NotFound is not an error when we're deleting ... hahaha perfect.

cjwagner: Sounds like we should maybe have a separate metric for NotFound so that we can alert if we can't find a significant fraction of the pods.

Issue #13032, opened 2019-06-14
skip gce* tests when only k/cmd/kubeadm is touched


neolit123:
so i had the idea for a while, to make the following jobs skip if a PR is only touching `cmd/kubeadm` in `kubernetes/kubernetes`.

```
pull-kubernetes-e2e-gce-100-performance
pull-kubernetes-e2e-gce
pull-kubernetes-e2e-gce-device-plugin-gpu
pull-kubernetes-kubemark-e2e-gce-big
```

my rationale is that these jobs run kube-up tests  on kubeadm deployer changes, which doesn't make sense. i think ideally we should skip them if the PR is **only** touching `cmd/kubeadm` and run `kind` presubmit instead.

this can be in place until kubeadm moves out of tree.

to my understanding negating a match however is not that easy.
[run_if_changed](https://github.com/kubernetes/test-infra/blob/master/prow/config/jobs.go#L207-L213) runs a job if one of the touched paths matches.

but you cannot use negative lookahead in golang `^((?!foo).)*$`,
which means that a new prow config field is required?

quick idea is: add `skip_if_only_changed` (not a great name) and error out if both that and `run_if_changed` are used.

#### so the tasks would be:
- extend prow to support such a functionality
- add the `skip_if_only_changed` fields to the above presubmit jobs + set `always_run: false`
- enable the kubeadm/kind presubmits to always run on `cmd/kubeadm` changes.

pros:
- skips running heavy tests on kubeadm PRs
- enables kubeadm presubmits

cons:
- complicates the prow config with additional regexp and fields.

@spiffxp please feel free to close if this is a no-go.

/assign @spiffxp @timothysc @neolit123
cc @fabriziopandini @BenTheElder 

/sig cluster-lifecycle
/sig testing
/kind feature
/priority backlog

BenTheElder:
It's possible to accomplish this with a clever run_if_changed regex without lookahead :-)

I will say it has generally been a footgun to run tests based on file path regexes because they're generally not precise enough but this sounds simple enough.

neolit123:
> It's possible to accomplish this with a clever run_if_changed regex without lookahead

i know, but from what i tested it ain't going to be pretty. :)

BenTheElder:
from last time this came up:
https://play.golang.org/p/cmPlf-Ro_y9
```
# don't run if only docs/*
[...]

PR #19092, opened 2020-09-01
[cluster-api] Enable smoke test for pre-submit job

sedefsavas:
This PR modifies pre-submit script to run only smoke tests.

cc @neolit123 @fabriziopandini @wfernandes @ncdc @vincepri 

/hold until https://github.com/kubernetes-sigs/cluster-api/pull/3582 and https://github.com/kubernetes-sigs/cluster-api/pull/3583 are merged.

### DIFF
--- a/config/jobs/kubernetes-sigs/cluster-api/cluster-api-presubmits.yaml
+++ b/config/jobs/kubernetes-sigs/cluster-api/cluster-api-presubmits.yaml
@@ -117,6 +117,9 @@ presubmits:
         args:
           - runner.sh
           - "./scripts/ci-e2e.sh"
+        env:
+          - name: GINKGO_FOCUS
+            value: "[PR-Blocking]"
         # we need privileged mode in order to do docker in docker
         securityContext:
           privileged: true


config/jobs/kubernetes-sigs/cluster-api/cluster-api-presubmits.yaml
```diff
@@ -116,7 +117,7 @@ presubmits:
       - image: gcr.io/k8s-testimages/kubekins-e2e:v20200831-9529f9c-1.18
         args:
           - runner.sh
-          - "./scripts/ci-e2e.sh"
+          - "./scripts/ci-e2e.sh smoke"
```
randomvariable, line 120: Instead of adding conditional logic to ci-e2e.sh, why not set the GINKGO_FOCUS environment variable?

fabriziopandini: +1 Also we should start to introduce labels to the test name, like [Conformance], [Slow], and eventually something CAPI specific like [Upgrades], [SelfHosting], ... so we can choose what fits in the smoke test by label instead of by name

config/jobs/kubernetes-sigs/cluster-api/cluster-api-ci.yaml
```diff
@@ -24,6 +24,7 @@ periodics:
     testgrid-num-failures-to-alert: "2"
 - name: periodic-cluster-api-e2e-test
   interval: 1h
+  run_if_changed: '^(docs/|.*\.md$)'
```
neolit123, line 27: this would do the reverse, it would run the tests **only** on /docs and *.md changes.
see here:
https://github.com/kubernetes/test-infra/issues/13032#issuecomment-501953200

wfernandes: That's correct. I misread the initial `^` as a negation whereas in this case it indicates the start of the line.

vincepri: We don't have to do this today, we can optimize these down the line — There isn't really a good way to do this because go regexp (used here) doesn't support negative lookahead

neolit123: due to the language limitations i can see e.g. `skip_if_changed` that takes precedence and errors out of both are in the prow config. but i vaguely recall arguments against this feature. #prow on k8s slack should know more.

ncdc: If we want to implement only running on docs/md changes, we'd probably need to model after the CAPV configuration, which explicitly lists paths (go packages + files like main.go, go.mod/sum, etc) for code changes for unit/e2e jobs. This becomes a maintenance burden, however, because it needs to be kept up to date if/when we add new packages or files that affect compilation.

config/jobs/kubernetes-sigs/cluster-api/cluster-api-presubmits.yaml
wfernandes: @sedefsavas I believe this is what @randomvariable is suggesting, https://github.com/kubernetes/test-infra/blob/e6b25da2c1bc8057d51e80828af0cfe4e08a022d/config/jobs/kubernetes-sigs/cluster-api-provider-aws/cluster-api-provider-aws-presubmits.yaml#L169-L178

with `GINKGO_FOCUS` thus not having to make changes to our ci-e2e.sh script as well.

config/jobs/kubernetes-sigs/cluster-api/cluster-api-ci.yaml
sedefsavas: Will create another PR for this.

sedefsavas: https://github.com/kubernetes/test-infra/pull/19103

config/jobs/kubernetes-sigs/cluster-api/cluster-api-presubmits.yaml
```diff
@@ -117,6 +117,9 @@ presubmits:
         args:
           - runner.sh
           - "./scripts/ci-e2e.sh"
+        env:
+          - name: GINKGO_FOCUS
```
wfernandes, line 121: I wonder if the `GINKGO_FOCUS=` [specified in `./scripts/ci-e2e.sh`](https://github.com/kubernetes-sigs/cluster-api/blob/9a1948ee0bc7226f26276d778b140630de88a04f/scripts/ci-e2e.sh#L54) will override the env var specified here.

In the `test/e2e/Makefile` we have a [conditional variable assignment operator](https://github.com/kubernetes-sigs/cluster-api/blob/9a1948ee0bc7226f26276d778b140630de88a04f/test/e2e/Makefile#L52) `GINKGO_FOCUS ?= ` which defaults to empty.

So we may not need the variable specified in `./scripts/ci-e2e.sh`.

Info on Makefile conditional variable assignment.
https://www.gnu.org/software/make/manual/make.html#index-_003f_003d

randomvariable: Yes, it should be deleted in ci-e2e.sh.

sedefsavas: Yes, good catch @wfernandes .
Created a PR for this: https://github.com/kubernetes-sigs/cluster-api/pull/3583

fabriziopandini (commented):
/approve

fabriziopandini:
/approve cancel
missing square brackets

vincepri (approved):
/approve

PR #19094, opened 2020-09-02
Use RepoLink in pod-utilities/clonerefs

zzq889:
clonerefs doesn't treat `repo_link` like other utilities. So I just add one.

### DIFF
--- a/prow/pod-utils/clone/clone.go
+++ b/prow/pod-utils/clone/clone.go
@@ -131,6 +131,8 @@ func gitCtxForRefs(refs prowapi.Refs, baseDir string, env []string, oauthToken s
 	}
 	if refs.CloneURI != "" {
 		g.repositoryURI = refs.CloneURI
+	} else if refs.RepoLink != "" {
+		g.repositoryURI = fmt.Sprintf("%s.git", refs.RepoLink)
 	}
 
 	if len(oauthToken) > 0 {


zzq889:
@spiffxp may I ask when will this PR be merged?

prow/pod-utils/clone/clone.go
```diff
@@ -131,6 +131,8 @@ func gitCtxForRefs(refs prowapi.Refs, baseDir string, env []string, oauthToken s
 	}
 	if refs.CloneURI != "" {
 		g.repositoryURI = refs.CloneURI
+	} else if refs.RepoLink != "" {
+		g.repositoryURI = fmt.Sprintf("%s.git", refs.RepoLink)
```
cjwagner, line 135: I'm not sure if this is valid for an arbitrary RepoLink since RepoLinks are meant to be human usable links to the source code, not necessarily a place that `git` can clone from. In particular I'd be concerned about this being problematic for Gerrit repos.
This is a pretty sensitive bit of code to change so we would need to be certain that this would not cause regressions for anyone, but unfortunately I don't think that is the case.

zzq889: Sounds reasonable. But I saw some other components like pipeline also has this kind of logic.

cjwagner: The pipeline component is barely used in its current state, thats probably not the behavior we want. I agree that we should use these fields consistently across components.

zzq889:
I think I should close this PR this time.

PR #19109, opened 2020-09-02
Fix oncall slack handle

chaodaiG:


chases2 (approved):

e-blackwelder (approved):
I now see the "(lowercase!)" comment. Perhaps we could manually lower-case them inside the tool in the future?

BenTheElder:
> I now see the "(lowercase!)" comment. Perhaps we could manually lower-case them inside the tool in the future?

these are the keys in the map, so we could but we'd have to take the map and iterate the keys inserting map[lowercase(key)] = map[key] which seems a bit silly when you need to read the following sentences to know what to put into the map anyhow.

PR #19114, opened 2020-09-03
Allow gcsweb to use a GCS service account credentials for authentication

droslean:
Currently, the tool can use an OAuth token to access private buckets.
However, if the user doesn't have any static google API key and instead
a token from a specific service account must be used for security
reasons, google API needs to dynamically generate an expirable token using
the service account's private key.
This commit also refactor the way how we access a private bucket by
introducing the storage client.


gcsweb/cmd/gcsweb/gcsweb.go
```diff
@@ -107,11 +124,27 @@ func (o *options) validate() error {
 	if _, err := os.Stat(o.flStyles); os.IsNotExist(err) {
 		return fmt.Errorf("styles path '%s' doesn't exists.", o.flStyles)
 	}
+	if o.oauthTokenFile != "" && o.gcsCredentialsFile != "" {
```
alvaroaleman, line 127: Why are we building all of this ourselves instead of using `cloud.google.com/go/storage`?

stevekuznetsov: Not sure I agree - "rewriting" it to build an authenticated client and use it is simple and we don't maintain any code that way. Let's take that approach.

droslean: Updated

droslean:
/hold

The way how the tool displays the files in this PR, required the user to be already authenticated in the google cloud project. Holding the PR until I will make the changes to fix this issue.

droslean:
/hold cancel

Displaying files is now including.

stevekuznetsov (commented):
@cjwagner this looks good to me and I was able to look at a deployed version of it in action to verify it's functioning. Do you want to either ACK the review or do your own, and then we can likely get you access to a test env as well?

gcsweb/cmd/gcsweb/gcsweb.go
```diff
@@ -107,14 +117,54 @@ func (o *options) validate() error {
...
+		return errors.New("specifying both --oauth-token-file and --gcs-credentials-file is not allowed.")
+	}
+
+	resolveGCSBaseURL = func() string { return gcsBaseURL }
```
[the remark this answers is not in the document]
droslean: well that was in an older commit. It doesn't exist now, since I made the storage client to get the files and output to the screen, we don't need any of the google URLs

gcsweb/cmd/gcsweb/gcsweb.go
```diff
@@ -107,14 +111,49 @@ func (o *options) validate() error {
...
+func getStorageClient(o options) (*storage.Client, error) {
+	ctx := context.Background()
+	clientOption := option.WithoutAuthentication()
+
+	if o.oauthTokenFile != "" {
+		secretAgent := &secret.Agent{}
+		if err := secretAgent.Start([]string{o.oauthTokenFile}); err != nil {
+			return nil, fmt.Errorf("Error starting secrets agent: %v", err)
+		}
+		clientOption = option.WithAPIKey(string(secretAgent.GetSecret(o.oauthTokenFile)))
```
cjwagner, line 141: I don't think there is any value in using the secret agent if you only actually load the secret once while initializing.

gcsweb/cmd/gcsweb/gcsweb.go
```diff
@@ -270,72 +303,73 @@ func (s *server) gcsRequest(w http.ResponseWriter, r *http.Request) {
...
+
+	objectLogger.Info("Processing request...")
+	// Getting the object attributes directly will determine if is a folder or a file.
+	objAttrs, _ := s.storageClient.Bucket(bucket).Object(object).Attrs(context.Background())
+
+	// This means that the object is a file.
+	if objAttrs != nil {
+		obj := s.storageClient.Bucket(bucket).Object(object)
+
+		r, err := obj.NewReader(context.Background())
```
cjwagner, line 315: Please use a different variable name. `r` is already used for the `*http.Request` parameter which makes the code after this a bit confusing.

gcsweb/cmd/gcsweb/gcsweb.go
```diff
@@ -270,72 +303,73 @@ func (s *server) gcsRequest(w http.ResponseWriter, r *http.Request) {
...
-
-	markers, found := r.URL.Query()["marker"]
-	if found {
-		url += "&marker=" + markers[0]
+		if _, err := io.Copy(w, r); err != nil {
+			objectLogger.WithError(err).Error("couldn't output object")
+			w.WriteHeader(http.StatusInternalServerError)
+			fmt.Fprintf(w, "Error: %v", err)
+			return
+		}
```
cjwagner, line 329: If the object is a file rather than a dir this outputs the file then proceeds to render the dir as well? I'm confused why this function doesn't return after successfully copying the file contents.

droslean: Nice catch. Addressed. Thanks!

gcsweb/cmd/gcsweb/gcsweb.go
droslean: Addressed. I removed the secret agent. Let's just read the file contents.

gcsweb/cmd/gcsweb/gcsweb.go
```diff
@@ -616,33 +457,27 @@ func (dir *gcsDir) Render(out http.ResponseWriter, inPath string) {
...
-	if err == nil {
-		mtime = ts.Format("02 Jan 2006 15:04:05")
-	}
-	var url, size string
-	if rec.isDir {
-		url = gcsPath + inPath + rec.Name
-		size = "-"
-	} else {
-		url = gcsBaseURL + inPath + rec.Name
-		size = fmt.Sprintf("%v", rec.Size)
```
cjwagner, line 638: Why/how is this logic ok to omit? It looks like the `isDir` field isn't used at all anymore and this always constructs the URL assuming the record is a directory. (but formats the size as if its a file?)

cjwagner (commented):
It'd be nice to see some unit tests and/or a demo instance since it seems like this PR includes substantial changes to how gcsweb works beyond just adding the storage client (e.g. redirect and rendering behavior) and I'm not really familiar with this code.

gcsweb/cmd/gcsweb/gcsweb.go
droslean: Before, the URL of the files was pointing directly to the google link. That was feasible because all the links were public. However now we need to get the contents of the files using a potential authenticated storage client.  So we omit the current logic and we can make the files point to the same gcsweb instance, and with its turn, it will handle the object as a file and output its contents. The storage client can understand if the requested object is a file or not by getting the attributes [here](https://github.com/kubernetes/test-infra/blob/a5a26e60f03f147274cf9fa2056c687244e8b29e/gcsweb/cmd/gcsweb/gcsweb.go#L311)

droslean:
> It'd be nice to see some unit tests and/or a demo instance since it seems like this PR includes substantial changes to how gcsweb works beyond just adding the storage client (e.g. redirect and rendering behavior) and I'm not really familiar with this code.

I will try to find the best way to unittest the whole functionality.

droslean:
@cjwagner @stevekuznetsov I added some unittests as well. Please check.

gcsweb/cmd/gcsweb/gcsweb.go
```diff
@@ -616,33 +470,27 @@ func (dir *gcsDir) Render(out http.ResponseWriter, inPath string) {
 
 // Record represents a single "Contents" entry in a GCS bucket.
 type Record struct {
-	Name  string `xml:"Key"`
-	MTime string `xml:"LastModified"`
-	Size  int64  `xml:"Size"`
+	Name  string
+	MTime time.Time
+	Size  int64
 	isDir bool
```
cjwagner, line 476: nit: This field still needs to be omitted.

cjwagner (approved):

PR #19120, opened 2020-09-03
Add requests/limits for all sig-scalability presubmits

jprzychodzen:
/hold
/cc @mm4tt 

I could not verify some of those values in Metrics Explorer, as those jobs are rarely triggered. However, all those presubmits are relatively cheap. It would be best to submit it and monitor, leaving merge for Monday.

### DIFF
--- a/config/jobs/kubernetes/sig-scalability/sig-scalability-presubmit-jobs.yaml
+++ b/config/jobs/kubernetes/sig-scalability/sig-scalability-presubmit-jobs.yaml
@@ -432,62 +432,12 @@ presubmits:
         - --use-logexporter
         image: gcr.io/k8s-testimages/kubekins-e2e:v20200903-fd282bc-master
         resources:
-          requests:
+          limits:
+            cpu: 2
             memory: "6Gi"
-
-  # clone of pull-perf-tests-clusterloader2 with enabled node killer and using only load test
-  - name: pull-perf-tests-clusterloader2-nodekiller-experimental
-    always_run: false
-    skip_report: false
-    max_concurrency: 1
-    branches:
-      - master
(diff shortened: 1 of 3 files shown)


mm4tt:
/lgtm
/approve
/test pull-test-infra-bazel

/hold
Let's submit this on Monday

MushuEE:
Does the TG config need to be removed as well?
https://github.com/kubernetes/test-infra/blob/2e51e452150284c5dc7a233368fa02dde3373c53/config/testgrids/kubernetes/presubmits/config.yaml#L170

jprzychodzen:
@MushuEE Right, removed from there. Also from `config/testgrids/conformance/conformance-all.yaml`

config/jobs/kubernetes/sig-scalability/sig-scalability-presubmit-jobs.yaml
```diff
@@ -432,62 +432,12 @@ presubmits:
         - --use-logexporter
         image: gcr.io/k8s-testimages/kubekins-e2e:v20200901-eeeadc5-master
         resources:
+          limits:
+            cpu: 2
+            memory: "8Gi"
```
wojtek-t, line 437: The reason to have 8Gi here and 6Gi for periodic is that we need more for bazel?

jprzychodzen: Bazel is not used here to build K8s - it's perf-tests presubmit.

When filling this value I've done manual inspection of Metrics Explorer and just rounded it to 8Gi instead of 6Gi.

Changing it to 6Gi to match periodic.

PR #19122, opened 2020-09-03
Log cmd in bumper in order to making debugging easier

alvaroaleman:


experiment/autobumper/bumper/bumper.go
```diff
@@ -229,6 +229,16 @@ func cdToRootDir() error {
 }
 
 func Call(stdout, stderr io.Writer, cmd string, args ...string) error {
+	(&logrus.Logger{
+		Out:       stderr,
+		Formatter: logrus.StandardLogger().Formatter,
```
stevekuznetsov, line 234: Will this carry forward the secretAgent/censor?

alvaroaleman: yeah, that's why it uses the passed-in stderr rather than `os.stderr`

stevekuznetsov (approved):

PR #19125, opened 2020-09-03
image-pushing/k8s-staging-nfd: add PULL_BASE_SHA as a gcb substitution

marquiz:
Makes it possible to have per-commit unique container image tag that is
re-producible in other jobs running against the same branch/commit. It
will be used by post-submit end-to-end test job that will test the very
image built and published by the image-pushing job.


ArangoGutierrez (approved):
/lgtm

marquiz:
That was quick, thanks @chaodaiG !!

PR #19127, opened 2020-09-03
prow/override: Trim whitespace in /override  <context>

cgwalters:
It's really confusing if one types e.g.
`/override  ci/prow/blah`
because Markdown compresses two spaces into one when rendered,
so the error is not obvious.  Let's be nice to the humans
and trim whitespace.

Signed-off-by: Colin Walters <walters@verbum.org>

### DIFF
--- a/prow/plugins/override/override.go
+++ b/prow/plugins/override/override.go
@@ -286,7 +286,7 @@ func handle(oc overrideClient, log *logrus.Entry, e *github.GenericCommentEvent,
 			log.Debug(resp)
 			return oc.CreateComment(org, repo, number, plugins.FormatResponseRaw(e.Body, e.HTMLURL, user, resp))
 		}
-		overrides.Insert(m[2])
+		overrides.Insert(strings.TrimSpace(m[2]))
 	}
 
 	authorized := authorizedUser(oc, log, org, repo, user)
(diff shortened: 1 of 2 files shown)


prow/plugins/override/override.go
```diff
@@ -286,7 +286,7 @@ func handle(oc overrideClient, log *logrus.Entry, e *github.GenericCommentEvent,
 			log.Debug(resp)
 			return oc.CreateComment(org, repo, number, plugins.FormatResponseRaw(e.Body, e.HTMLURL, user, resp))
 		}
-		overrides.Insert(m[2])
+		overrides.Insert(strings.TrimSpace(m[2]))
```
alvaroaleman, line 289: @cgwalters please add a tetscase for this
/ok-to-test

cgwalters: Done!

cgwalters:
I've hit this a number of times, here's one example https://github.com/ostreedev/ostree/pull/2189#issuecomment-686681560 - you can see the issue by viewing the raw markdown.

stevekuznetsov (approved):

cgwalters:
Hah, fixed the extra whitespace in my test for extra whitespace...

alvaroaleman (approved):

alvaroaleman:
failure looks transient
/retest

PR #19137, opened 2020-09-05
Include spelling to transfigure.sh bazel tests

### LINKED ISSUE #18990: transfigure.sh fails if there is a misspelling in the config
**What happened**:
Transfigure.sh tests for Testgrid and Config lint tests, but does not test for misspellings.

If a user names their dashboard something that the common misspelling tester doesn't like, it will cause transfigure.sh to fail only AFTER the PR is made

Instead, testing jobs (when transfigure.sh is run in its dry-mode) should run this misspelling test also

https://github.com/kubernetes/test-infra/pull/18895
https://github.com/GoogleCloudPlatform/oss-test-infra/pull/479

/assign
/help-wanted

matthyx:
Fixes #18990

### DIFF
--- a/testgrid/cmd/transfigure/transfigure.sh
+++ b/testgrid/cmd/transfigure/transfigure.sh
@@ -77,7 +77,7 @@ main() {
   fi
 
   echo "Running kubernetes/test-infra tests..."
-  bazel test //config/tests/...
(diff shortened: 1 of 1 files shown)


chases2 (approved):
Looks good, thank you!

PR #19142, opened 2020-09-07
Update local-e2e.yaml

jayunit100:
Skip the `SchedulerPreemption` tests , as they break local-e2e tests on hack/local-up-cluster.sh (i gues bc only one node?)

### DIFF
--- a/config/jobs/kubernetes/sig-testing/local-e2e.yaml
+++ b/config/jobs/kubernetes/sig-testing/local-e2e.yaml
@@ -28,7 +28,7 @@ presubmits:
         - --deployment=local
         - --ginkgo-parallel=1
         - --provider=local
-        - --test_args=--ginkgo.focus=\[Conformance\] --ginkgo.skip=SchedulerPreemption|\[sig\-apps\]\sDaemon\sset\s\[Serial\]\sshould\srollback\swithout\sunnecessary\srestarts\s\[Conformance\]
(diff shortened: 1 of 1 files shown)


dims:
@jayunit100 nothing has changed in a while in that rest https://github.com/kubernetes/kubernetes/commits/master/test/e2e/scheduling/preemption.go so i would not suggest we do this.

Also the local e2e CI job is optional, so your PR should be able to merge without this passing.

jayunit100:
so, as is - - the scheduler tests *don't* pass because of 

```
error during ./hack/ginkgo-e2e.sh --ginkgo.focus=\[Conformance\] --ginkgo.skip=\[sig\-apps\]\sDaemon\sset\s\[Serial\]\sshould\srollback\swithout\sunnecessary\srestarts\s\[Conformance\] --report-dir=/workspace/_artifacts --disable-log-dump=true (interrupted): exit status 1
```

The root cause is:

```
I0907 15:33:53.751] 
7603
I0907 15:33:53.751] [Fail] [sig-scheduling] SchedulerPreemption [Serial] [It] validates lower priority pod preemption by critical pod [Conformance] 
7604
I0907 15:33:53.751] /go/src/k8s.io/kubernetes/_output/local/go/src/k8s.io/kubernetes/test/e2e/scheduling/preemption.go:253
7605
I0907 15:33:53.751] 
```

So, I suspect this test has *never* passed?

jayunit100:
hiya @dims ,  since we know the Pre-emption tests will always fail on local-up-cluster (bc its single node), what's the motivation for running them ?  maybe im missing context... 

Specifically, these are *gauranteed to always fail* because of ... 
```
Sep  9 12:17:45.367: FAIL: We need at least two pods to be created but all nodes are already heavily utilized, so preemption tests cannot be run
```

dims:
@jayunit100 good point. Can you please update the PR with the full name of the test (just like the other one)?

jayunit100:
oh race condition in the lgtm loop I think, :)

PR #19149, opened 2020-09-09
Fix prod-push make command for Kettle

MushuEE:
Indentation and made command match what is used in CI job

/label cleanup
/area kettle


amwat (commented):
/lgtm

PR #19151, opened 2020-09-09
Add GCS credential path option to deck in starter-*.yaml

zuiurs:
Add GCS credential path option to deck to use intended credential not anonymous, and remove duplicate option in `starter-*.yaml`.


chases2:
What's the context for this change? Are we switching credentials or deprecating a flag?

/test all

PR #19159, opened 2020-09-09
Bumper: Allow user to specify working directory for git push

GeoBK:
This change would 
1. add the RunAndCommitIfChanged function which would run the passed in commands and commit the changes if any.
2. allow the user to specify the working directory of a GitPush

### DIFF
--- a/experiment/autobumper/bumper/bumper.go
+++ b/experiment/autobumper/bumper/bumper.go
@@ -20,6 +20,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"errors"
+	"flag"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -105,6 +106,26 @@ type Options struct {
 	ExtraFiles          fileArrayFlag
 
 	SkipPullRequest bool
+}
+
+// GitAuthorOptions is specifically to read the author info for a commit
+type GitAuthorOptions struct {
+	GitName  string
+	GitEmail string
+}
+
+// AddFlags will read the author info from the command line parameters
+func (o *GitAuthorOptions) AddFlags(fs *flag.FlagSet) {
+	fs.StringVar(&o.GitName, "git-name", "", "The name to use on the git commit.")
+	fs.StringVar(&o.GitEmail, "git-email", "", "The email to use on the git commit.")
+}
+
+// Validate will validate the input GitAuthorOptions
+func (o *GitAuthorOptions) Validate() error {
+	if (o.GitEmail == "") != (o.GitName == "") {
+		return fmt.Errorf("--git-name and --git-email must be specified together")
+	}
+	return nil
 }
 
 // GitCommand is used to pass the various components of the git command which needs to be executed
--- a/experiment/autobumper/bumper/bumper.go
+++ b/experiment/autobumper/bumper/bumper.go
@@ -58,6 +58,8 @@ const (
 
 	errOncallMsgTempl = "An error occurred while finding an assignee: `%s`.\nFalling back to Blunderbuss."
 	noOncallMsg       = "Nobody is currently oncall, so falling back to Blunderbuss."
+
+	gitCmd = "git"
 )
 
 var (
@@ -264,7 +266,7 @@ func cdToRootDir() error {
 		}
 		return nil
 	}
-	cmd := exec.Command("git", "rev-parse", "--show-toplevel")
+	cmd := exec.Command(gitCmd, "rev-parse", "--show-toplevel")
 	output, err := cmd.Output()
 	if err != nil {
 		return fmt.Errorf("failed to get the repo's root directory: %w", err)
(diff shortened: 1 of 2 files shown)


prow/flagutil/git.go
```diff
@@ -28,8 +29,8 @@ import (
 // GitOptions holds options for interacting with git.
 type GitOptions struct {
 	host          string
-	user          string
-	email         string
+	User          string
```
stevekuznetsov, line 32: So the use from `bumper` is not the `git` library. Although the options _seem_ the same, they are different. We should have a separate struct for them that's not tied to this, unless we expect a single binary to use both?

GeoBK: I dont understand what you mean by "So the use from bumper is not the git library".
I am using both git and bumper here(https://github.com/openshift/ci-tools/pull/1211/files#diff-4440cf6f38e6026b7a8028aa83c9f664R33) 
Am I doing that wrong?

experiment/autobumper/bumper/bumper.go
```diff
@@ -518,14 +518,59 @@ func GitCommitAndPush(remote, remoteBranch, name, email, message string, stdout,
...
+	}
+
+	if !changed {
+		logrus.WithField("command", fullCommand).Info("No changes to commit")
+		return false, nil
+	}
+
+	gitCmd := "git"
+	gitArgs := []string{}
+	if workingDir != "" {
```
MushuEE, line 556: If this might be used across even more functions in the future is it worth making something like (below)?
```
type GitCommand struct {
         baseCommand       string
         workingDir               string
         args                          []string{}
}

func (gc GitCommand) Call(stdout, stderr io.Writer) error {
        	if gc.workingDir != "" {
		gc.args = append(gc.args, fmt.Sprintf("--git-dir=%s/.git", gc.workingDir), fmt.Sprintf("--work-tree=%s", gc.workingDir))
	}
	c := exec.Command(gc.baseCommand, gc.args...)
	c.Stdout = stdout
	c.Stderr = stderr
	return c.Run()
}
```

experiment/autobumper/bumper/bumper.go
```diff
@@ -518,14 +518,59 @@ func GitCommitAndPush(remote, remoteBranch, name, email, message string, stdout,
 }
 
 // GitPush push the changes to the given remote and branch.
-func GitPush(remote, remoteBranch string, stdout, stderr io.Writer) error {
+func GitPush(remote, remoteBranch string, stdout, stderr io.Writer, workingDir string) error {
 	logrus.Info("Pushing to remote...")
-	if err := Call(stdout, stderr, "git", "push", "-f", remote, fmt.Sprintf("HEAD:%s", remoteBranch)); err != nil {
+	gitArgs := []string{}
+	if workingDir != "" {
+		gitArgs = append(gitArgs, fmt.Sprintf("--git-dir=%s/.git", workingDir), fmt.Sprintf("--work-tree=%s", workingDir))
```
petr-muller, line 525: I think we can simplify the `--git-dir=%s/.git --work-tree=%s` to `-C %s`.

experiment/autobumper/bumper/bumper.go
```diff
@@ -518,14 +518,59 @@ func GitCommitAndPush(remote, remoteBranch, name, email, message string, stdout,
...
+	}
+	gitPushArgs := append(gitArgs, "push", "-f", remote, fmt.Sprintf("HEAD:%s", remoteBranch))
```
petr-muller, line 527: IMO no need to introduce a new variable, just use `gitArgs` (and because we're in `GitPush` function, I'd even call it just `args`.

experiment/autobumper/bumper/bumper.go
```diff
@@ -518,14 +518,59 @@ func GitCommitAndPush(remote, remoteBranch, name, email, message string, stdout,
...
+	if err := Call(stdout, stderr, "git", gitPushArgs...); err != nil {
 		return fmt.Errorf("failed to git push: %w", err)
 	}
 	return nil
 }
 
+// RunAndCommitIfNeeded makes a commit in the workingDir if there are
+// any changes resulting from the command execution. Returns true if a commit is made
+func RunAndCommitIfNeeded(stdout, stderr io.Writer, author, cmd string, args []string, workingDir string) (bool, error) {
+	fullCommand := fmt.Sprintf("%s %s", filepath.Base(cmd), strings.Join(args, " "))
```
petr-muller, line 537: Same comment like I had in the ci-tools PR - `fullCommand` is used in logging. While reading logs, I want to see what the tool really does (calls `cmd`) without prettyfing it with `filepath.Base`.

prow/flagutil/git.go
GeoBK: moved this from git options into new class GitAuthorOptions in bumper.go

stevekuznetsov (approved):

Issue #19161, opened 2020-09-09
Crier: gerrit reporter delayed by 50 minutes, and duplicate reporting


chaodaiG:
**What happened**:
After bumping prow from https://github.com/kubernetes/test-infra/compare/967ee0e48b...b5ff4681c3, crier gerrit reporter reported on gerrit label ~50 minutes after all jobs voting for the label finished. Also, the reporter reported N times the same message when there are N jobs voting on the same label

**What you expected to happen**:
Crier gerrit reporter delay no more than a couple of minutes, and no duplicate reports

**How to reproduce it (as minimally and precisely as possible)**:

**Please provide links to example occurrences, if any**:

**Anything else we need to know?**:
After filtering all the changes between the bump, the most relevant PR is https://github.com/kubernetes/test-infra/pull/18950

chaodaiG:
cc @alvaroaleman

chaodaiG:
cc @cjwagner

alvaroaleman:
Fixed by https://github.com/kubernetes/test-infra/pull/19171
/close

PR #19162, opened 2020-09-09
Revert "Make crier controller-runtime based #18950"

chaodaiG:
This change had possibly caused crier gerrit reporter delayed reporting, reverting it while investigating the root cause
See bug https://github.com/kubernetes/test-infra/issues/19161

/hold
/assign @alvaroaleman 
/cc @cjwagner


prow/crier/controller_test.go
```diff
@@ -190,17 +271,14 @@ func TestController_Run(t *testing.T) {
...
-				pjclientset: cs,
-				reporter:    &rp,
+			for _, job := range test.knownJobs {
+				prowjobs = append(prowjobs, job)
 			}
+			cs := fake.NewSimpleClientset(prowjobs...)
+			nmwrk := 2
+			c := NewController(cs, q, inf, &rp, nmwrk)
 
+<<<<<<< HEAD
```
cjwagner, line 281: There are 2 unresolved merge conflicts.

cjwagner (changes requested):

cjwagner:
We've discussed out of band and it doesn't seem fair to revert this given that other instances may depend on it now. Reverting may have been the right move if we'd reacted to the problem earlier, but its been 12 days now so we're planning to continue using an old version of Prow until we can root cause this and make a forward fix.
/close

PR #19165, opened 2020-09-09
kubemark: add 5000 node scheduler job

adtac:
/sig scheduling
/cc @mm4tt

### DIFF
--- a/config/jobs/kubernetes/sig-scalability/sig-scalability-periodic-jobs.yaml
+++ b/config/jobs/kubernetes/sig-scalability/sig-scalability-periodic-jobs.yaml
@@ -260,7 +260,9 @@ periodics:
   tags:
   - "perfDashPrefix: kubemark-5000Nodes"
   - "perfDashJobType: performance"
-  interval: 12h
+  # Run twice a day (at 00:01 and 16:01 UTC) on the odd days of each month. The
+  # job is expected to take ~12-14h, hence the 16 hour gap.
+  cron: '1 0,16 1-31/2 * *'
   labels:
     preset-service-account: "true"
     preset-k8s-ssh: "true"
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes/sig-scalability/sig-scalability-periodic-jobs.yaml
```diff
@@ -260,7 +260,7 @@ periodics:
   tags:
   - "perfDashPrefix: kubemark-5000Nodes"
   - "perfDashJobType: performance"
-  interval: 12h
+  interval: 48h
```
adtac, line 263: Not sure how to make each run on alternate days as we discussed, if you have any suggestions, please let me know.

config/jobs/kubernetes/sig-scalability/sig-scalability-periodic-jobs.yaml
```diff
@@ -319,6 +319,67 @@ periodics:
...
+  labels:
+    preset-service-account: "true"
+    preset-k8s-ssh: "true"
+    preset-dind-enabled: "true"
+    preset-e2e-kubemark-common: "true"
+    preset-e2e-kubemark-gce-scale: "true"
+  annotations:
+    testgrid-dashboards: sig-scalability-kubemark
+    testgrid-tab-name: kubemark-5000
```
MushuEE, line 335: Maybe `kubemark-5000-scheduler`? This tab is already claimed by `ci-kubernetes-kubemark-gce-scale`

config/jobs/kubernetes/sig-scalability/sig-scalability-periodic-jobs.yaml
mm4tt: Use `cron` instead of `interval`. Here is an example - https://github.com/kubernetes/test-infra/blob/939c608dd2d9e6c974c2946a8c206e74f64a023a/config/jobs/kubernetes/sig-scalability/sig-scalability-release-blocking-jobs.yaml#L3

config/jobs/kubernetes/sig-scalability/sig-scalability-periodic-jobs.yaml
adtac: Done.

config/jobs/kubernetes/sig-scalability/sig-scalability-periodic-jobs.yaml
adtac: Done.

config/jobs/kubernetes/sig-scalability/sig-scalability-periodic-jobs.yaml
```diff
@@ -319,6 +319,67 @@ periodics:
...
+    containers:
+    - image: gcr.io/k8s-testimages/kubekins-e2e:v20200901-eeeadc5-master
+      args:
+      - --repo=k8s.io/kubernetes=master
+      - --repo=k8s.io/perf-tests=master
+      - --root=/go/src
+      - --timeout=1100
+      - --scenario=kubernetes_e2e
+      - --
+      - --cluster=kubemark-5000
```
adtac, line 348: As per https://github.com/kubernetes/perf-tests/pull/1464#discussion_r487918757, I have concerns about whether using the same cluster is a good idea. While we only have plans to have 3 workloads (anti-affinity, affinity, spreading), assuming that each takes at least ~13 hours, I think it might be prudent to:

1. Use a different cluster so that we won't have false positives from the jobs overlapping. While this is a less of a concern with the 100-node job, I think it might be a good idea to do it there as well.
2. Run the 5k job weekly - the scheduler team/sig doesn't need daily data for the 5k job, weekly results for the 5k job are sufficient. There might be more workloads added to the scheduler suite in the future.

cc @ahg-g for opinions on (2)?

ahg-g: I think we should aim for at least two runs per week.

mm4tt: I asked about this already in the other PR, but I'd like to understand why a single test takes 13h. It seems super long comparing to the vanilla density test which takes less than 2h

ahg-g:
Thanks Adi, anything left on this one?

adtac:
Nope, should be good to go, the job doesn't take 13h as feared (should take ~1 hour per workload + 1 hour for cluster set up and tear down).

/cc @mm4tt

config/jobs/kubernetes/sig-scalability/sig-scalability-periodic-jobs.yaml
```diff
@@ -260,7 +260,7 @@ periodics:
   tags:
   - "perfDashPrefix: kubemark-5000Nodes"
   - "perfDashJobType: performance"
-  interval: 12h
+  cron: '1 12 1-31/2 * *' # Run every second day at 12:01 UTC starting on the 1st of each month
```
mm4tt, line 263: Previously we had two runs per day, we this we'll have only one.
I know that cron is not flexible enough to express what we'd want (e.g. keep running twice a day, every 3rd run run scheduler suite instead of regular test). Ideally, we'd be able to specify multiple cron expressions, but prow doesn't support it. But, can we at least do something like:

1. Run the regular test twice a day every second (odd) day, given that it takes ~14h, let's run it 00:01 and 16:01
2. Run the scheduler test once a day every second (even) day, to give the previous regular tests 14h +2h buffer to finish let's start it at 08:01

WDYT?

adtac: sg :+1: but instead of 08:01 UTC for the scheduler suite job, I opted for a more conservative 10:01 to give it a bit more time. This shouldn't be an issue because the scheduler suite job is expected to finish in 6-8 hours, which is well before the next kubemark-5000Nodes job.

mm4tt:
/approve

Adding the scalability oncall for this week as FYI
/cc @tosi3k 

@adtac, please keep an eye on the kubemark jobs and let us know if something goes wrong

PR #19169, opened 2020-09-09
add transistion images for bazel 3.4.1

mikedanese:
While I'm in the neighborhood...

Ideally, this does not switch the effective bazel version of jobs until
we change the .bazelversion file in each repo, but I would appreciate a
confirmation on that assumption.

I'm not touching images/bazelbuild/variants.yaml because I'm not sure
what they are for.

/assign @fejta @BenTheElder 
cc @justaugustus

### DIFF
--- a/images/kubekins-e2e/variants.yaml
+++ b/images/kubekins-e2e/variants.yaml
@@ -3,21 +3,21 @@ variants:
     CONFIG: experimental
     GO_VERSION: 1.15
     K8S_RELEASE: stable
-    BAZEL_VERSION: 2.2.0
-    OLD_BAZEL_VERSION: 0.28.1
+    BAZEL_VERSION: 3.4.1
+    OLD_BAZEL_VERSION: 2.2.0
     UPGRADE_DOCKER: 'true'
   go-canary:
     CONFIG: go-canary
     GO_VERSION: 1.15
     K8S_RELEASE: stable
-    BAZEL_VERSION: 2.2.0
-    OLD_BAZEL_VERSION: 0.23.2
+    BAZEL_VERSION: 3.4.1
+    OLD_BAZEL_VERSION: 2.2.0
   master:
     CONFIG: master
     GO_VERSION: 1.15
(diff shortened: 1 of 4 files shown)


BenTheElder:
> Ideally, this does not switch the effective bazel version of jobs until we change the .bazelversion file in each repo, but I would appreciate a confirmation on that assumption.

AIUI it should not, but it's possible some repo uses these w/o a .bazelversion (we should tell them to add one then anyhow)

images/kubekins-e2e/variants.yaml
```diff
@@ -3,8 +3,8 @@ variants:
     CONFIG: experimental
     GO_VERSION: 1.15
     K8S_RELEASE: stable
-    BAZEL_VERSION: 2.2.0
-    OLD_BAZEL_VERSION: 0.28.1
+    BAZEL_VERSION: 3.4.1
+    OLD_BAZEL_VERSION: 2.2.0
```
BenTheElder, line 7: we should probably update the master key below if we're updating images/bazel/variants for the kubernetes-master config above?

mikedanese: Ah ya, fixed.

fejta (approved):
/hold

mikedanese:
@fejta why hold?

fejta:
Just to give control over when it merges

fejta:
(so feel free to unhold whenever you want)

mikedanese:
Thanks @fejta !

/hold cancel

justaugustus:
@fejta @mikedanese -- Saw the comments flying by...
Just noting that [bazel 3.7.0](https://github.com/bazelbuild/bazel/releases/tag/3.7.0) is out. Do we want to try to bump to that? :upside_down_face:

mikedanese:
@justaugustus I'd recommend we do a fast follow. I've done a fair amount of testing at this version already.

justaugustus:
SGTM!

PR #19171, opened 2020-09-10
Fix Crier handling of reporters that report multiple PJs.

cjwagner:
This PR includes fixes for 2 distinct bugs and some cleanup.

/assign @alvaroaleman @fejta @chaodaiG

### DIFF
--- a/prow/crier/controller.go
+++ b/prow/crier/controller.go
@@ -101,13 +101,41 @@ func (r *reconciler) updateReportState(pj *prowv1.ProwJob, log *logrus.Entry, re
 			return false, err
 		}
 		if pj.Status.PrevReportStates != nil &&
-			newpj.Status.PrevReportStates[r.reporter.GetName()] == reportedState {
+			pj.Status.PrevReportStates[r.reporter.GetName()] == reportedState {
 			return true, nil
 		}
 		return false, nil
 	}); err != nil {
 		return fmt.Errorf("failed to wait for updated report status to be in lister: %w", err)
 	}
+	return nil
+}
+
+func (r *reconciler) updateReportStateWithRetries(pj *prowv1.ProwJob, log *logrus.Entry) error {
+	reportState := pj.Status.State
+	log = log.WithFields(logrus.Fields{
+		"prowjob":   pj.Name,
+		"jobName":   pj.Spec.Job,
+		"jobStatus": reportState,
+	})
+	// We have to retry here, if we return we lose the information that we already reported this job.
+	if err := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
+		// Get it first, this is very cheap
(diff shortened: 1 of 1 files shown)


prow/crier/controller.go
```diff
@@ -101,7 +101,7 @@ func (r *reconciler) updateReportState(pj *prowv1.ProwJob, log *logrus.Entry, re
 			return false, err
 		}
 		if pj.Status.PrevReportStates != nil &&
-			newpj.Status.PrevReportStates[r.reporter.GetName()] == reportedState {
+			pj.Status.PrevReportStates[r.reporter.GetName()] == reportedState {
```
cjwagner, line 104: Bug 1.

prow/crier/controller.go
```diff
@@ -166,27 +193,12 @@ func (r *reconciler) reconcile(log *logrus.Entry, req reconcile.Request) (*recon
...
 	}
 
-	log.Info("Reported job, now will update pj")
+	log.Info("Reported job(s), now will update pj(s)")
 	for _, pjob := range pjs {
-		reportedState := pjob.Status.State
-		// We have to retry here, if we return we lose the information that we already reported this job.
-		if err := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
-			// Get it first, this is very cheap
-			if err := r.pjclientset.Get(r.ctx, req.NamespacedName, &pj); err != nil {
```
cjwagner, line 175: Bug 2. This `Get` call gets the original PJ identified by `req.NamespacedName` rather than getting the `pjob` PJ.

prow/crier/controller.go
```diff
@@ -166,27 +194,12 @@ func (r *reconciler) reconcile(log *logrus.Entry, req reconcile.Request) (*recon
 		return requeue, nil
 	}
 
-	log.Info("Reported job, now will update pj")
+	log.Info("Reported job(s), now will update pj(s)")
 	for _, pjob := range pjs {
```
fejta, line 198: Please move this for loop out of this function. This is the second time reusing a stale `pj` variable has caused an outage (previously this was from https://github.com/kubernetes/test-infra/pull/18065)

cjwagner: I moved the loop contents into its own function to help with that. I figured putting the loop itself in the helper function would increase the chance of reusing a stale PJ rather than reduce it since multiple PJs would be accessible from the function.

WDYT? I'm happy to move the loop into the function if that is still preferable.

fejta: I'd prefer to put it in its own function but can do that later. This is much better. Thanks.

fejta (approved):

fejta:
fixes https://github.com/kubernetes/test-infra/pull/19171

PR #19172, opened 2020-09-10
Make crier use exponential backoff when waiting for lister update.

cjwagner:
Previously the initial wait was 1 second which throttled the report rate significantly by blocking the lister. Recent issues with excessive reporting exacerbated this to the point that reporting was 50 mins delayed on some PRs. While we don't need to optimize for the excessive reporting bug, this delay seems unreasonably large even for normal operation as it necessitates blocking reporting for 1 second per job which can be significant when reporting multiple jobs as a group like the gerrit reporter does.

This PR switches to use exponential backoff. Instead of 2 attempts 1 second apart, we now make 4 attempts with intervals of 0.1, 0.3, 0.9, and 2.7 seconds.
/assign @alvaroaleman @fejta

### DIFF
--- a/prow/crier/controller.go
+++ b/prow/crier/controller.go
@@ -92,11 +92,17 @@ func (r *reconciler) updateReportState(pj *prowv1.ProwJob, log *logrus.Entry, re
 		return fmt.Errorf("failed to patch: %w", err)
 	}
 
+	name := types.NamespacedName{Namespace: pj.Namespace, Name: pj.Name}
 	// Block until the update is in the lister to make sure that events from another controller
 	// that also does reporting dont trigger another report because our lister doesn't yet contain
 	// the updated Status
-	name := types.NamespacedName{Namespace: pj.Namespace, Name: pj.Name}
-	if err := wait.Poll(time.Second, 3*time.Second, func() (bool, error) {
+	backoff := wait.Backoff{
+		Duration: 100 * time.Millisecond,
+		Factor:   3.0,
+		Steps:    4,
+		Cap:      3 * time.Second,
+	}
+	if err := wait.ExponentialBackoff(backoff, func() (bool, error) {
 		if err := r.pjclientset.Get(r.ctx, name, pj); err != nil {
 			return false, err
 		}


prow/crier/controller.go
```diff
@@ -92,11 +92,17 @@ func (r *reconciler) updateReportState(pj *prowv1.ProwJob, log *logrus.Entry, re
...
 	}
 
+	name := types.NamespacedName{Namespace: pj.Namespace, Name: pj.Name}
 	// Block until the update is in the lister to make sure that events from another controller
 	// that also does reporting dont trigger another report because our lister doesn't yet contain
 	// the updated Status
-	name := types.NamespacedName{Namespace: pj.Namespace, Name: pj.Name}
-	if err := wait.Poll(time.Second, 3*time.Second, func() (bool, error) {
+	backoff := wait.Backoff{
+		Duration: 100 * time.Millisecond,
```
fejta, line 100: Is 100ms a reasonable time to expect to have already received the update to the watch?

fejta (approved):
/hold

prow/crier/controller.go
cjwagner: I don't have any data on this, but I'd expect so.  The check should be very cheap anyways and optimizing for reducing the locking time seems more important than reducing the number of cheap attempts. 
If you'd like I could add some debug logs as well to indicate how often we have to retry so that we can tweak these parameters, but I think these values will work just fine.

petr-muller (approved):

prow/crier/controller.go
alvaroaleman: Not a very strong opinion one way or another, but we could maybe just stick with the `wait.Poll` but do it every 100ms? Since this just attempts to get data from the cache its a super cheap thing to do

cjwagner: I usually prefer exponential backoff, but that seems reasonable here since this is so cheap and since there will only be one thread in a polling loop per controller. That should also reduce the average amount of time we spend unnecessarily waiting compared to exponential backoff which is important because this blocks the controller. I'll close this PR and open another to simply decrease the poll period.
/close

PR #19173, opened 2020-09-10
add conformance gate prow job

zachmandeville:
## Summary
This PR introduces a job to check whether any stable endpoints promoted in the release are without conformance tests.  
## Background
It is [a requirement that GA/stable API's have conformance tests](https://github.com/kubernetes/community/pull/1806).  This requirement is hard to enforce and multiple releases have had stable endpoints promoted without tests.  This creates technical debt for conformance,with test writers working on new tests to target API's from older releases.  You can see this in this graph from [APISnoop](https://apisnoop.cncf.io/conformance-progress):
visualization(3)
where the orange bars in each release mark newly promoted endpoints without tests, and the accumulation of technical debt.
## This job
This prow job uses the same [APISnoop database](https://github.com/cncf/apisnoop/tree/master/apps/snoopdb) to pull in the audit logs from the most recent runs of these two e2e jobs:
- [ci-kubernetes-e2e-gci-gce](https://testgrid.k8s.io/google-gce#gce-cos-master-default)
- [ci-kubernetes-gce-conformance-latest](https://testgrid.k8s.io/conformance-all#Conformance%20-%20GCE%20-%20master)
It compares the endpoints hit by conformance tests in the audit logs to the  endpoints defined in the open API spec, and lists any new GA endpoints that are not hit by a conformance test.  If the list is empty, the job succeeds.  If not, the job fails.  One can then look at the bottom of the logs to find a list of untested endpoints.
You can see an example of this job on [prow.cncf.io](https://prow.cncf.io).

This job is intended to ultimately work as a release blocking job, and so its configured to fit  the [release blocking criteria](https://github.com/kubernetes/sig-release/blob/master/release-blocking-jobs.md#release-blocking-criteria-and-dashboard).

### DIFF
--- a/config/jobs/kubernetes/sig-arch/conformance-gate.yaml
+++ b/config/jobs/kubernetes/sig-arch/conformance-gate.yaml
@@ -1,0 +1,16 @@
+periodics:
+- interval: 3h
+  name: apisnoop-conformance-gate
+  annotations:
+    testgrid-dashboards: sig-arch-conformance
+    testgrid-tab-name: apisnoop-conformance-gate
+    test-grid-alert-email: kubernetes-sig-arch-conformance-test-failures@googlegroups.com
+    testgrid-num-failures-to-alert: '1'
+    description: 'Uses APISnoop to check that new GA endpoints are conformance tested in latest e2e test run'
+  decorate: true
+  spec:
+    containers:
+    - name: apisnoop-gate
+      image: gcr.io/k8s-staging-apisnoop/snoopdb:v20200923-v2020.09.23-1-gb3afc41
+      args:
+      - postgres
--- a/config/jobs/kubernetes/sig-arch/OWNERS
+++ b/config/jobs/kubernetes/sig-arch/OWNERS
@@ -1,0 +1,16 @@
+# See the OWNERS docs at https://go.k8s.io/owners
+
+# This is the owner of the test code.  The test data itself is owned by sig-architecture.
+reviewers:
+  - cheftako
+  - oomichi
+  - johnbelamaric
+  - bobymcbobs
+approvers:
+  - cheftako
+  - spiffxp
+  - johnbelamaric
+  - hh
+labels:
+  - area/conformance
+  - sig/architecture
--- a/config/testgrids/kubernetes/sig-arch/OWNERS
+++ b/config/testgrids/kubernetes/sig-arch/OWNERS
@@ -1,0 +1,16 @@
+# See the OWNERS docs at https://go.k8s.io/owners
+
+# This is the owner of the test code.  The test data itself is owned by sig-architecture.
+reviewers:
+  - cheftako
+  - oomichi
+  - johnbelamaric
+  - bobymcbobs
+approvers:
+  - cheftako
+  - spiffxp
+  - johnbelamaric
+  - hh
+labels:
+  - area/conformance
+  - sig/architecture
--- a/config/testgrids/kubernetes/sig-arch/config.yaml
+++ b/config/testgrids/kubernetes/sig-arch/config.yaml
@@ -1,0 +1,11 @@
+# Dashboard Group
+#
+dashboard_groups:
+- name: sig-arch
+  dashboard_names:
+  - sig-arch-conformance
+
+# Dashboards
+#
+dashboards:
+- name: sig-arch-conformance


hh:
@zz Go ahead and specify the command as your startup script.

config/jobs/kubernetes/sig-release/conformance-gate.yaml
```diff
@@ -0,0 +1,14 @@
+periodics:
+- interval: 4h
+  name: conformance-gate
+  decorate: true
+  spec:
+    containers:
+    - name: apisnoop-gate
+      image: zachboyofdestiny/snoopdb:gate
```
feiskyer, line 8: could we avoid using a personal image here?

spiffxp:
>You can see an example of this job on prow.cncf.io:
> - the logs from a passing job( as 1.20.0 currently has no new untested endpoints)
> - the logs from a purposefully failing job(it looks at untested endpoints from prior releases)

Do you have a spyglass or gcsweb uri for those?  They expired

config/jobs/kubernetes/sig-release/conformance-gate.yaml
```diff
@@ -0,0 +1,14 @@
...
+      ports:
+      - name: postgres
+        containerPort: 5432
+      env:
+      - name: K8S_JOB
```
BenTheElder, line 13: FWIW there are a number of standard environment variables present in kubernetes CI
https://github.com/kubernetes/test-infra/blob/master/prow/jobs.md#job-environment-variables

aojea:
2 questions:

1. can we have another job to gate per percentage? i.e. fail if GA less than 75%, beta less than 50%
2. can we use a different terminology to differentiate between API endpoints and the endpoints API?

config/jobs/kubernetes/sig-release/conformance-gate.yaml
zachmandeville: @feiskyer , yep!  We just put in a change to APISnoop so that the new database it uses can be a part of the existing test-infra image builds for APISnoop.  I opened a PR for this change here: https://github.com/kubernetes/test-infra/pull/19234

When merged, I can update the container here to use this image.

zachmandeville:
> > You can see an example of this job on prow.cncf.io:
> > 
> > * the logs from a passing job( as 1.20.0 currently has no new untested endpoints)
> > * the logs from a purposefully failing job(it looks at untested endpoints from prior releases)
> 
> Do you have a spyglass or gcsweb uri for those? They expired

@spiffxp , I do!  The logs are coming from jobs running at https://prow.cncf.io
Here's a link to [the logs for the most recent job](https://prow.cncf.io/log?job=snoop-gate&id=1306080617657536512)

(note: this job is now failing as 1.20.0 has a new, untested stable endpoint with `getInternalApiserverAPIGroup`)

config/jobs/kubernetes/sig-release/conformance-gate.yaml
zachmandeville: Good point, @BenTheElder!  I updated the underlying image to use a standard env var, and removed my redundant one.  This is part of commit a0c307c431521046333762492c402edab4e00c61

zachmandeville:
> 2 questions:
> 
>     1. can we have another job to gate per percentage? i.e. fail if GA less than 75%, beta less than 50%
> 
>     2. can we use a different terminology to differentiate between API endpoints and the endpoints API?
 
To answer your questions, @aojea:
1. This would be straightforward to do with the foundations set in this job.
2. It would be straightforward to update the messages given in the logs to use different terminology, if this is what you mean?

config/jobs/kubernetes/sig-release/conformance-gate.yaml
zachmandeville: @feiskyer , This job now uses an image from `gcr.io/k8s-staging-apisnoop`.

hh:
It's owned by SIG-Arch, specifically the Conformance Subproject, but I the folks that need to be notified are SIG-Release.

dims:
Confirmed what @hh says

justaugustus:
Okay, so we should probably have this under a SIG Arch directory, so we can assign conformance OWNERS to review/approve.

This was about to merge as a SIG Release test w/o SIG Release approval.

zachmandeville:
I created a `sig-arch` directory at the same level as `sig-release` and moved this job there.  Would this work?
I also added an email for job fail notifications, but am keen for feedback on the best email to use for this!

hh:
I'm trying to make sure we include @kubernetes/sig-release in the configuration of this job, but in crossing SIGs with regards to who's the consumer of this job vs who 'OWNERS' owns the reconfiguration via git.

The intent of the job is to be sure that beta to stable API promotions include conformance test coverage. This should be a release blocking job, so we don't create any more technical/testing debt.

The notification target / email should probably be sig-release rather than sig-arch, while the maintenance / ownership of the job itself should probably be sig-arch.

If this job fails, sig-release should communicate to the SIG/KEP/API introducing any newly stable feature their requirement to making this job pass by writing tests (which in turn get approved by SIG-Arch) within the release.

If the tests can't be written and promoted to satisfy sig-arch, then the new endpoints should be reverted to delayed to a later release.

config/jobs/kubernetes/sig-arch/conformance-gate.yaml
```diff
@@ -0,0 +1,18 @@
+periodics:
+- interval: 4h
+  name: conformance-gate
+  annotations:
+    testgrid-dashboards: conformance-apisnoop
+    testgrid-tab-name: conformance-gate
+    test-grid-alert-email: en-server-prow-test-alert@google.com
```
neolit123, line 7: This should be the sig arch mailing list?

neolit123 (commented):
Please keep the commits sqashed to 1.

config/jobs/kubernetes/sig-arch/conformance-gate.yaml
neolit123: There should be a description annotation explaining with a couple of sentences what the job does. Example https://github.com/kubernetes/test-infra/blob/b660a5251071db1e5aca5a9e1ca5b08064691eb7/config/jobs/kubernetes/sig-node/node-kubelet.yaml#L5

config/jobs/kubernetes/sig-arch/conformance-gate.yaml
```diff
@@ -0,0 +1,18 @@
...
+    testgrid-num-failures-to-alert: '1'
+  decorate: true
+  spec:
+    containers:
+    - name: apisnoop-gate
+      image: gcr.io/k8s-staging-apisnoop/snoopdb:v20200923-v2020.09.23-1-gb3afc41
+      ports:
+      - name: postgres
+        containerPort: 5432
```
spiffxp, line 16: Can this be dropped?

config/jobs/kubernetes/sig-arch/conformance-gate.yaml
spiffxp: Agree with @neolit123's comments, although I might hold off on alerting the arch mailing list until this is proven to run and not fail spuriously.  The existing e-mail address is not valid though, I'd maybe suggest apisnoop@ii.coop to start with?

If the goal is to say sig arch owns this job, it would be helpful to setup a sig-arch dashboard group, eg: sig-arch/sig-arch-conformance/apisnoop-conformance-gate or something similar.

spiffxp (changes requested):
Some comments/questions

config/jobs/kubernetes/sig-arch/conformance-gate.yaml
zachmandeville: ah, yep!  This has been dropped, @spiffxp.

config/jobs/kubernetes/sig-arch/conformance-gate.yaml
zachmandeville: @spiffxp and @neolit123 : I've included the creation of a sig-arch dashboard into this pr and updated the job's annotations accordingly.  The dashboard group is `sig-arch`, and the dashboard-name is `sig-arch-conformance`.

spiffxp:
/cc @johnbelamaric
LGTM, but I'd like to get the /lgtm from @johnbelamaric as sig-arch chair

config/jobs/kubernetes/sig-arch/conformance-gate.yaml
```diff
@@ -0,0 +1,16 @@
+periodics:
+- interval: 3h
+  name: apisnoop-conformance-gate
+  annotations:
+    testgrid-dashboards: sig-arch-conformance
+    testgrid-tab-name: apisnoop-conformance-gate
+    test-grid-alert-email: apisnoop@ii.coop
```
johnbelamaric, line 7: Change to newly created group: kubernetes-sig-arch-conformance-test-failures@googlegroups.com

johnbelamaric (changes requested):
LGTM in general, please update to use the test failures group alias I just created.

config/jobs/kubernetes/sig-arch/conformance-gate.yaml
zachmandeville: Ah, thank you for that!  The email has been updated.

harche:
sorry my bad. Wrong repo.

spiffxp (approved):
/approve
/lgtm

PR #19170, opened 2020-09-10
Migrate pull-kubernetes-bazel-test to k8s-infra-prow-build

spiffxp:
Part of https://github.com/kubernetes/test-infra/issues/19070

Per https://github.com/kubernetes/test-infra/issues/19070#issuecomment-689899166 I'm leaving the resource requests as-is since the canary job seemed to behave acceptably

Issue #19070, opened 2020-08-31
Migrate merge-blocking jobs to dedicated cluster: pull-kubernetes-bazel-test


spiffxp:
**What should be cleaned up or changed**:

This is part of #18550

To properly monitor the outcome of this, you should be a member of k8s-infra-prow-viewers@kubernetes.io.  PR yourself into https://github.com/kubernetes/k8s.io/blob/master/groups/groups.yaml#L603-L628 if you're not a member.

Migrate pull-kubernetes-bazel-test to k8s-infra-prow-build by adding a `cluster: k8s-infra-prow-build` field to the job:
- Make sure any release-branch variants are also updated
- Reference this issue, but do not "fix" it; instead leave this open to monitor the behavior of the job before/after
- Here's an example PR that did this for pull-kubernetes-integration: https://github.com/kubernetes/test-infra/pull/18814

NOTE: migrating this job is not as straightforward as some of the other #18550 issues, because:
- other flags need to be removed to migrate it off of RBE
```
- --config=remote
- --remote_instance_name=projects/k8s-prow-builds/instances/default_instance
```
- its behavior off of RBE may be substantially different or consume more resources than with RBE (ref: https://docs.bazel.build/versions/master/remote-execution.html)

Once the PR has merged, note the date/time it merged.  This will allow you to compare before/after behavior.

Things to watch for the job
- https://prow.k8s.io/?job=pull-kubernetes-bazel-test
  - does the job start failing more often?
  - does the job start going into error state?
- https://testgrid.k8s.io/presubmits-kubernetes-blocking#pull-kubernetes-bazel-test&graph-metrics=test-duration-minutes
  - does the job duration look worse than before? spikier than before?
- https://storage.googleapis.com/k8s-gubernator/triage/index.html?pr=1&job=pull-kubernetes-bazel-test
  - do more failures show up than before?
- https://prow.k8s.io/job-history/gs/kubernetes-jenkins/pr-logs/directory/pull-kubernetes-bazel-test
  - (can be used to answer some of the same questions as above)
[...]

PR #19179, opened 2020-09-10
Bump cpu from 4 to 7 for pull-kubernetes-bazel-test

spiffxp:
To match the resources used by the CI version of this job; previous
resources were for running bazel using RBE, we're no longer doing that

Followup to https://github.com/kubernetes/test-infra/pull/19170
Part of https://github.com/kubernetes/test-infra/issues/19070


spiffxp:
/cc @liggitt 
FYI @kubernetes/ci-signal 
Let's see if flakes go down with this limit

hasheddan (approved):
/lgtm

PR #19181, opened 2020-09-10
[KEP-1933] Prow Job

mlevesquedion:
## Summary

This PR introduces a new Prow Job for manually running the [go-flow-levee](https://github.com/google/go-flow-levee/) analyzer on `kubernetes/kubernetes`, as part of the Alpha phase of [KEP-1933](https://github.com/kubernetes/enhancements/tree/master/keps/sig-instrumentation/1933-secret-logging-static-analysis). The test target was introduced in `kubernetes/kubernetes` in [this PR](https://github.com/kubernetes/kubernetes/pull/94661).

The intent is to be able to manually trigger a run of the analyzer on a PR by commenting `/test pull-kubernetes-verify-govet-levee`.

I have tested the Job locally using `pj-on-kind.sh`.

The YAML for the job is based on `config/jobs/kubernetes/sig-testing/verify.yaml`.

### DIFF
--- a/config/jobs/kubernetes/sig-instrumentation/verify-govet-levee.yaml
+++ b/config/jobs/kubernetes/sig-instrumentation/verify-govet-levee.yaml
@@ -1,0 +1,49 @@
+presubmits:
+  kubernetes/kubernetes:
+  - name: pull-kubernetes-verify-govet-levee
+    cluster: k8s-infra-prow-build
+    always_run: false
+    decorate: true
+    skip_branches:
+      - release-\d+.\d+ # per-release job
+    optional: true
+    skip_report: false
+    path_alias: k8s.io/kubernetes
+    labels:
+      preset-service-account: "true"
+      preset-dind-enabled: "true"
+    spec:
+      containers:
+      - image: gcr.io/k8s-testimages/kubekins-e2e:v20200831-9529f9c-master
+        imagePullPolicy: Always
+        command:
+        - make
+        args:
+        - verify
+        env:
+        # Space separated list of the checks to run
+        - name: WHAT
+          value: govet-levee
+        - name: EXCLUDE_TYPECHECK
+          value: "y"
+        # this is now covered by the optional pull-kubernetes-files-remake that runs when Makefile.generated_files / make-rules changes
+        - name: EXCLUDE_FILES_REMAKE
+          value: "y"
+        - name: EXCLUDE_GODEP
+          value: "y"
+        - name: KUBE_VERIFY_GIT_BRANCH
+          value: master
+        - name: REPO_DIR
+          value: /workspace/k8s.io/kubernetes
+        # docker-in-docker needs privileged mode
+        securityContext:
+          privileged: true
+        resources:
+          # Consider reducing memory limits after govet memory usage issue is
+          # addressed in https://github.com/kubernetes/kubernetes/issues/93822
+          limits:
+            cpu: 7
+            memory: 12Gi
+          requests:
+            cpu: 7
+            memory: 12Gi
--- a/config/jobs/kubernetes/sig-instrumentation/verify-govet-levee.yaml
+++ b/config/jobs/kubernetes/sig-instrumentation/verify-govet-levee.yaml
@@ -9,6 +9,8 @@ presubmits:
     optional: true
     skip_report: false
     path_alias: k8s.io/kubernetes
+    annotations:
+      testgrid-create-test-group: "true"
     labels:
       preset-service-account: "true"
       preset-dind-enabled: "true"
--- a/config/testgrids/kubernetes/presubmits/config.yaml
+++ b/config/testgrids/kubernetes/presubmits/config.yaml
@@ -147,6 +147,9 @@ dashboards:
   - name: pull-kubernetes-files-remake
     test_group_name: pull-kubernetes-files-remake
     base_options: width=10
+  - name: pull-kubernetes-verify-govet-levee
+    test_group_name: pull-kubernetes-verify-govet-levee
+    base_options: width=10
 - name: presubmits-kubernetes-scalability
   dashboard_tab:
   - name: pull-kubernetes-e2e-gce-100-performance


mlevesquedion:
I've signed the CLA.

mlevesquedion:
/assign @spiffxp @BenTheElder 

Aaron and Ben, I'm assigning you since you have context from https://github.com/kubernetes/kubernetes/pull/94661.

spiffxp (changes requested):
SIG Testing is not the appropriate owner for this job.  I would suggest either SIG Instrumentation (based on the KEP), or perhaps SIG Security longer term.  Please move to one of their directories

I would also suggest following the pattern that [pull-kubernetes-typecheck uses](https://github.com/kubernetes/test-infra/blob/e711c265201915ffc732296770c0e804c89d44b7/config/jobs/kubernetes/sig-testing/typecheck.yaml), `make verify WHAT=govet-levee`

You're not skipping older release branches via `skip_branches`, if that's intentional for testing that's fine

Tests may fail because this should show up in testgrid, you should add to https://github.com/kubernetes/test-infra/blob/9c663ce1bbf49680e86afe0287accabe44023b77/config/testgrids/kubernetes/presubmits/config.yaml#L80

mlevesquedion:
> SIG Testing is not the appropriate owner for this job. I would suggest either SIG Instrumentation (based on the KEP), or perhaps SIG Security longer term. Please move to one of their directories

Moved to SIG-Instrumentation for now.

> I would also suggest following the pattern that pull-kubernetes-typecheck uses, make verify WHAT=govet-levee

Thanks for the suggestion! Done.

> You're not skipping older release branches via skip_branches, if that's intentional for testing that's fine

Oops. We do want to skip those. Done.

> Tests may fail because this should show up in testgrid, you should add to ...

Done.

spiffxp:
```
could not find the referenced (TestGroup) pull-kubernetes-verify-govet-levee
```

ah right, you need an annotation to auto-generate that testgroup for you https://github.com/kubernetes/test-infra/blob/e711c265201915ffc732296770c0e804c89d44b7/config/jobs/kubernetes/sig-testing/typecheck.yaml#L15

config/jobs/kubernetes/sig-instrumentation/verify-govet-levee.yaml
```diff
@@ -0,0 +1,49 @@
...
+  - name: pull-kubernetes-verify-govet-levee
+    cluster: k8s-infra-prow-build
+    always_run: false
+    decorate: true
+    skip_branches:
+      - release-\d+.\d+ # per-release job
+    optional: true
+    skip_report: false
+    path_alias: k8s.io/kubernetes
+    labels:
```
spiffxp, line 12:
```suggestion
    annotations:
      testgrid-create-test-group: "true"
    labels:
```

chases2:
> ```
> could not find the referenced (TestGroup) pull-kubernetes-verify-govet-levee
> ```
> 
> ah right, you need an annotation to auto-generate that testgroup for you
> 
> https://github.com/kubernetes/test-infra/blob/e711c265201915ffc732296770c0e804c89d44b7/config/jobs/kubernetes/sig-testing/typecheck.yaml#L15

Is this a useful TestGrid row? This is a test that is only supposed to run while manually invoked, so there's no guarantee how frequently it will run or how reasonable the code it tests is.

I don't see any issues with having the test, I'm just trying to understand the use case here.

mlevesquedion:
On closer examination, I don't think we need this in testgrid.

However in @spiffxp 's original comment it sounded like not having it in testgrid could cause issues :

> Tests may fail because this should show up in testgrid

(My interpretation may be wrong.)

The intent is to be able to run the test manually on a number of PR's to confirm that everything is working properly so we can graduate to beta. We don't need testgrid for that.

mlevesquedion:
I tried removing the testgrid for the job and running `test-infra/config/tests/testgrids/config_test.go` and I got the following failure:

```
=== RUN   TestKubernetesProwInstanceJobsMustHaveMatchingTestgridEntries
--- FAIL: TestKubernetesProwInstanceJobsMustHaveMatchingTestgridEntries (0.01s)
    config_test.go:493: Job pull-kubernetes-verify-govet-levee does not have a matching testgrid testgroup
```

So it does look like it's preferable to have the testgrid.

Let me know if I need to make additional changes to the job.

BenTheElder:
https://testgrid.k8s.io/presubmits-kubernetes-blocking -- jobs that block PRs from merging
https://testgrid.k8s.io/presubmits-kubernetes-nonblocking -- jobs that do not block PRs from merging

Kubernetes/Kubernetes repo presubmits should be in one of these.

mlevesquedion:
Sounds good, @BenTheElder. It is in `non-blocking` right now.

chases2:
> I tried removing the testgrid for the job and running `test-infra/config/tests/testgrids/config_test.go` and I got the following failure:
> 
> ```
> === RUN   TestKubernetesProwInstanceJobsMustHaveMatchingTestgridEntries
> --- FAIL: TestKubernetesProwInstanceJobsMustHaveMatchingTestgridEntries (0.01s)
>     config_test.go:493: Job pull-kubernetes-verify-govet-levee does not have a matching testgrid testgroup
> ```
> 
> So it does look like it's preferable to have the testgrid.
> 
> Let me know if I need to make additional changes to the job.

On further reflection, you should keep the TestGrid configuration. I'm confusing what TestGrid and Configurator do, and what this repository does.

There's an [ongoing effort](https://github.com/kubernetes/test-infra/blob/6c49475cfc14f8697fb78a6075c5a529ecdd790c/config/tests/testgrids/config_test.go#L357-L359) specifically in this instance to put presubmits in TestGrid. The issue for discussion is at https://github.com/kubernetes/test-infra/issues/18159. It seems contentious, but not something to address in this PR.

You did it right the first time. Sorry about that!

spiffxp (approved):
/approve
/lgtm

Issue #19124, opened 2020-09-03
conformance-ga-only (ci-kubernetes-conformance-kind-ga-only) scheduling timeout


hasheddan:
**What happened**:

`conformance-ga-only` has been regularly experiencing pod scheduling timeouts since 8/24/20.

**Please provide links to example occurrences, if any**:

https://testgrid.k8s.io/sig-release-master-blocking#conformance-ga-only
https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-conformance-kind-ga-only/1301526736504098817

**Anything else we need to know?**:

Error message:

```
There are no nodes that your pod can schedule to - check your requests, tolerations, and node selectors (0/93 nodes are available: 3 node(s) had taints that the pod didn't tolerate, 5 Insufficient memory, 90 Insufficient cpu.)
```

/cc @kubernetes/ci-signal

spiffxp:
/close
https://github.com/kubernetes/test-infra/pull/19183 should fix this, and in the future please file against kubernetes/kubernetes

PR #19183, opened 2020-09-10
Reduce ci-conformance-kind-ga-only cpu to 7

spiffxp:
This was opened in response to https://k8s-testgrid.appspot.com/sig-release-master-blocking#conformance-ga-only being red; when it fails, it's because it can't schedule

I think we're running into the same issue as kind presubmits that were asking for 7300m and failing to find a node with sufficient CPU. If we want "the whole node", 7 cpu is still enough, since everything here asks for cpu, and nothing asks for <1 cpu

ref: https://github.com/kubernetes/test-infra/pull/19081 for the presubmits

### DIFF
--- a/config/jobs/kubernetes/sig-testing/conformance-e2e.yaml
+++ b/config/jobs/kubernetes/sig-testing/conformance-e2e.yaml
@@ -182,17 +182,10 @@ periodics:
       resources:
         limits:
           memory: 12Gi
-          cpu: 7300m
+          cpu: 7
         requests:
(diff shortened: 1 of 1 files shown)


spiffxp:
/cc @BenTheElder 
FYI @kubernetes/ci-signal

hasheddan (approved):
/lgtm

spiffxp:
This should fix https://github.com/kubernetes/test-infra/issues/19124

PR #19187, opened 2020-09-11
Add exclusion list to buckets.yaml in Kettle

MushuEE:
This adds Tests, Docs, and code to ignore a list of jobs pulled from that bucket URL. This does not require kettle redeploy to update the exclude list, just merge changes in `buckets.yaml` to master. 

/area kettle
/cc @spiffxp @michaelkolber 
/kind feature

### DIFF
--- a/kettle/README.md
+++ b/kettle/README.md
@@ -70,6 +70,22 @@ It might take a couple of hours to be fully functional and start updating BigQue
 
 To add fields to the BQ table, Visit the [k8s-gubernator:build BigQuery dataset](https://bigquery.cloud.google.com/dataset/k8s-gubernator:build) and Select the table (Ex. Build > All). Schema -> Edit Schema -> Add field. As well as update [schema.json](./schema.json)
 
+## Adding Buckets
+
+To add a new GCS bucket to Kettle, simply update [buckets.yaml](./buckets.yaml) in `master`, it will be auto pulled by Kettle on the next cycle.
+
+```yaml
+gs://<bucket path>: #bucket url
+  contact: "username" #Git Hub Username
+  prefix: "abc:" #the identifier prefixed to jobs from this bucket (ends in :).
+  sequential: (bool) #an optional boolean that indicates whether test runs in this
+  #                  bucket are numbered sequentially
+  exclude_jobs: [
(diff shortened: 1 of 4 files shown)


kettle/README.md
```diff
@@ -70,6 +70,22 @@ It might take a couple of hours to be fully functional and start updating BigQue
...
+```yaml
+gs://<bucket path>: #bucket url
+  contact: "username" #Git Hub Username
+  prefix: "abc:" #the identifier prefixed to jobs from this bucket (ends in :).
+  sequential: (bool) #an optional boolean that indicates whether test runs in this
+  #                  bucket are numbered sequentially
+  exclude_jobs: [
+    'job_name1',
+    'job_name2',
+  ] #List of jobs to explicitly exclude from kettle data collection
```
spiffxp, line 86: For a multi-line list I would probably use a different syntax unless there's a good reason not too
```suggestion
  exclude_jobs: # list of jobs to explicitly exclude from kettle data collection
    - job_name1
    - job_name2
```

kettle/buckets.yaml
```diff
@@ -6,11 +6,16 @@
...
+#   exclude_jobs: A list of job names that will not be uploaded to BQ
 
 gs://kubernetes-jenkins/logs/:
   contact: "fejta"
   prefix: ""
   sequential: false
+  exclude_jobs: [
+    'ci-test-infra-benchmark-demo',
+    'ci-kubernetes-coverage-unit',
+  ]
```
spiffxp, line 18:
```suggestion
  exclude_jobs:
    - ci-test-infra-benchmark-demo
    - ci-kubernetes-coverage-unit
```

spiffxp (changes requested):
One syntax nit

kettle/README.md
MushuEE: ah thanks, I just went with the first option I saw when searching `adding a list to yaml`

kettle/buckets.yaml
MushuEE: Validated works 👍

spiffxp (approved):
/approve
/lgtm

PR #19188, opened 2020-09-11
spyglass - make the order of JUnit test cases deterministic

### LINKED ISSUE #18175: JUnit lense changes test-render order randomly

matthyx:
fixes #18175
this PR supercedes #19029 

/cc @tosi3k 
/assign @alvaroaleman

### DIFF
--- a/prow/spyglass/lenses/junit/lens_test.go
+++ b/prow/spyglass/lenses/junit/lens_test.go
@@ -350,13 +350,13 @@ func TestGetJvd(t *testing.T) {
 						<testcase classname="fake_class_0" name="fake_test_0"></testcase>
 					</testsuite>
 					<testsuite>
-						<testcase classname="fake_class_0" name="fake_test_0"></testcase>
+						<testcase classname="fake_class_1" name="fake_test_1"></testcase>
 					</testsuite>
 				</testsuites>
 				`),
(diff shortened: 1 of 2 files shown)


prow/spyglass/lenses/junit/lens_test.go
```diff
@@ -350,13 +350,13 @@ func TestGetJvd(t *testing.T) {
 						<testcase classname="fake_class_0" name="fake_test_0"></testcase>
 					</testsuite>
 					<testsuite>
-						<testcase classname="fake_class_0" name="fake_test_0"></testcase>
```
tosi3k, line 353: Could we create a separate test case instead of modifying the existing one?

PR #19192, opened 2020-09-11
kops: Run a special JWKS test job daily

justinsb:
We want particular e2e coverage while we're developing the feature.


justinsb:
cc @rifelpet I think you have the most context here?

Issue #13432, opened 2019-07-12
Test result ingestion into BigQuery via Kettle is hiccuping increasingly frequently


spiffxp:
**What happened**:
http://velodrome.k8s.io/dashboard/db/bigquery-metrics?panelId=12&fullscreen&orgId=1&from=now-6M&to=now

Lots of alerts happening.  We've only had one or two weeks in Q2 where an alert hasn't fired.

**What you expected to happen**:

No alerts to fire.

**How to reproduce it (as minimally and precisely as possible)**:

**Please provide links to example occurrences, if any**:

**Anything else we need to know?**:

I suspect this isn't due to Kettle being written in python, but rather the pattern that Kettle follows.  It may require a redesign or rewrite but I'm not necessarily sure it's mandated that we switch languages for performance.  I suspect i/o is our problem here (due to unbounded growth of a sqlite database)

/area kettle
/area metrics

spiffxp:
/priority important-soon
I am concerned this is going to tip over at some point

<img width="921" alt="Screen Shot 2019-08-14 at 1 05 16 PM" src="https://user-images.githubusercontent.com/49258/63052449-35e2bd00-be94-11e9-976b-f98dd3f8930e.png">

spiffxp:
/remove-lifecycle rotten
It's still hiccuping, but alerts are no longer firing as often
![Screen Shot 2019-12-12 at 3 03 38 PM](https://user-images.githubusercontent.com/49258/70756140-996ce200-1cf0-11ea-8e11-30e2aad6cb67.png)

MushuEE:
@spiffxp thanks, will start looking into it.
/assign

MushuEE:
Span is increasing deltas, not sure if kettle is taking longer to process all the data or not https://k8s-testgrid.appspot.com/sig-testing-misc#metrics-kettle

It might be worth it to start publishing the time that steps take to run within commands. Will add logging shortly

MushuEE:
```
 resource: {…}  
 severity: "ERROR"  
 textPayload: "Traceback (most recent call last):
  File "/kettle/update.py", line 59, in <module>
    main()
  File "/kettle/update.py", line 47, in main
    call(bq_cmd + bq_ext + ' k8s-gubernator:build.week build_week.json.gz schema.json')
  File "/kettle/update.py", line 25, in call
    raise OSError('invocation failed')
OSError: invocation failed
" 
```
[...]

PR #19195, opened 2020-09-11
Add call timing and error dumping to kettle

MushuEE:
To better debug kettle failures and stale build tables, I added some more debugging information to the updater. #13432 investigation 

/area kettle

/cc @spiffxp 

Currently output only looks like :
```
  File "/kettle/update.py", line 59, in <module>
    main()
  File "/kettle/update.py", line 47, in main
    call(bq_cmd + bq_ext + ' k8s-gubernator:build.week build_week.json.gz schema.json')
  File "/kettle/update.py", line 25, in call
    raise OSError('invocation failed')
OSError: invocation failed
" 
```
I chose the redirect because subprocess lib sometimes has memory issue with large calls and I didn't want to hit that.

### DIFF
--- a/kettle/update.py
+++ b/kettle/update.py
@@ -15,9 +15,6 @@
 # limitations under the License.
 
 
-
-
-
 import os
 from datetime import datetime
 
@@ -34,7 +31,6 @@ def print_dump(file):
 
 
 def call(cmd, dump=False):
-
     started = datetime.now()
 
     redirect = ' > {DUMP}}' if dump else ''
--- a/kettle/update.py
+++ b/kettle/update.py
@@ -15,14 +15,36 @@
 # limitations under the License.
 
 
+
+
+
 import os
+from datetime import datetime
+
+DUMP = 'dump.txt'
+
+def print_dump(file):
+    if os.path.exists(file):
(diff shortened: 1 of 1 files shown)


kettle/update.py
```diff
@@ -16,13 +16,32 @@
...
+        print(output)
+    else:
+        print(f'unable to find dump file: {file}')
+
+
+def call(cmd, dump=False):
+    started = datetime.now()
+
+    redirect = ' > {DUMP}}' if dump else ''
+    status = os.system(f'{cmd}{redirect}')
```
spiffxp, line 37: I don't think literal expansion happens recursively?
```
>>> foo = 'monkeys'
>>> bar = '{foo}'
>>> print(f'{bar}')
{foo}
```
Reassigning to cmd means we can include the redirect in further logging
```suggestion
    cmd = f'{cmd} > {DUMP}' if dump else cmd
    status = os.system(cmd)
```

spiffxp (changes requested):
Some nits

kettle/update.py
MushuEE: Good catch

spiffxp (approved):
/approve
/lgtm

PR #19198, opened 2020-09-13
run e2e test on minimum k8s version

andyzhangx:
per current azure csi drivers doc, 1.14 is the minimum supported version, adjust some test configs, run on k8s 1.14 (one e2e test run on 1.14 is enough)

/assign  @ZeroMagic @chewong


ZeroMagic (approved):
/lgtm

PR #19194, opened 2020-09-11
Revert "Sinker: Use cache for pods #19090"

chaodaiG:
Revert #19090
For #19157 

/cc @cjwagner @spiffxp

Issue #19157, opened 2020-09-09
Sinker stopped deleting pods after prow bump


chaodaiG:
**What happened**:
Sinker stopped deleting pods after bumping prow https://github.com/kubernetes/test-infra/pull/19105

**What you expected to happen**:
Sinker work properly

**How to reproduce it (as minimally and precisely as possible)**:

**Please provide links to example occurrences, if any**:

**Anything else we need to know?**:

chaodaiG:
Reverting the bump for now: https://github.com/kubernetes/test-infra/pull/19156

chaodaiG:
The reverted changes are: https://github.com/kubernetes/test-infra/compare/b5ff4681c3...94d8b4dfcd

chaodaiG:
Based on sinker logs, there was no error listing pods or deleting pods, so it seems like the problem is at listing pods, which could possibly related to https://github.com/kubernetes/test-infra/pull/19090
+ @alvaroaleman for insights

fejta:
> The reverted changes are: b5ff468...94d8b4d

Try and use a PR next time :)

chaodaiG:
/reopen

This appeared after the bump today

alvaroaleman:
Fixed in #19194 
/close

PR #19204, opened 2020-09-14
Sinker: Use cache for pods

alvaroaleman:
This PR re-adds the changes from https://github.com/kubernetes/test-infra/pull/19090 that got reverted in https://github.com/kubernetes/test-infra/pull/19194 due to https://github.com/kubernetes/test-infra/issues/19157 and fixes the issue with the original change where it would only deleted pods in the cluster the prowjobs are in (xref https://github.com/kubernetes/test-infra/pull/19090#discussion_r487294693)

/assign @cjwagner @chaodaiG


prow/cmd/sinker/main.go
```diff
@@ -137,21 +137,28 @@ func main() {
...
 
-	podClients := map[string]podInterface{}
-	for cluster, client := range buildClusterClients {
-		podClients[cluster] = client
+	buildClusterClients := map[string]ctrlruntimeclient.Client{}
+	for clusterName, buildManager := range buildManagers {
+		if err := mgr.Add(buildManager); err != nil {
+			logrus.WithError(err).Fatal("Failed to add build cluster manager to main manager")
+		}
+		buildClusterClients[clusterName] = buildManager.GetClient()
```
cjwagner, line 154: Please update the unit tests to validate that we clean pods from all build clusters.

cjwagner (commented):
Looks right to me, but please update the unit tests to validate the bug is fixed and prevent future regressions.

prow/cmd/sinker/main.go
alvaroaleman: We already have unittests that validate that, the bug here is because we didn't insert the correct client, not because we didn't use the correct client. Testing this is pretty difficult and would require us to build a `fakeManager` that is then used in unittests.

cjwagner: Fair enough.
/lgtm
/approve

config/prow/cluster/sinker_rbac.yaml
```diff
@@ -49,6 +49,8 @@ rules:
     verbs:
       - delete
       - list
+      - watch
```
fejta, line 52: This is a breaking change, right? Please add to ANNOUNCEMENTS.md that sinker now needs `watch` and `get` access on pods.

PR #19207, opened 2020-09-14
Switch from 81-lts to 85-lts image family

bsdnet:


bsdnet:
/held

Other than python2 issue, we found other issues.

PR #19209, opened 2020-09-14
Update query to use new `repos` field and `repo_commit`

MushuEE:
This is part of the effort to integrate [@spiffxp 's proposal](https://docs.google.com/document/d/139R-4VJjt9_RF2tcNNnH_fi35qQPdCot8CIhaG91HOU/edit#heading=h.x9snb54sjlu9) #14643. These changes are the `Update the query` element of the doc. 

Kettle has been updated so that every job with `repos` in either `Started.json` or `finished.json.metadata` will get populated as a json Blob string. So both Bootstrap and Decorated jobs will populate it. Kettle now also populates `repo-commit` field which was updated in the query to be the default and fall back to `version` if `null`.

*Kettle was only updated to make these changes about a month and a half ago. If this is merged we will not be looking at flakes that far back. Because we only look a week back here... I think this is ok.

/hold
make sure this looks correct... 

/area metrics
/cc @cjwagner @spiffxp 

Reference queries
```
          SELECT
            job,
            if(substr(job, 0, 3) = 'pr:', 'pull', 'ci') kind,  /* pull or ci */
            ifnull(repo_commit, version) version, /* bootstrap git version, empty for ci  */
            if(substr(job, 0, 3) = 'pr:',
              regexp_extract(
                repos,
                r'[^,]+,\d+:([a-f0-9]+)"'
              ),
              version
            ) commit,  /* repo commit for PR or version for CI */
            result,  /* SUCCESS if the build passed */
            test  /* repeated tuple of tests */
          FROM `k8s-gubernator.build.week` as t
          where
            datetime(started) > datetime_sub(current_datetime(), interval 7 DAY)
            and version != 'unknown'
            and (
              (substr(job, 0, 3) = 'ci-' and version != 'unknown') or
              array_length(split(replace(repos,', ', ''), ',')) = 2 /*serial pr jobs only (# of PR refs +1 == 2)*/
            )
```
and
```
SELECT * FROM `k8s-gubernator.build.week` WHERE started > '2020-09-04 23:05:29'
``` 
Just to grab some recent runs

### DIFF
--- a/metrics/configs/flakes-config.yaml
+++ b/metrics/configs/flakes-config.yaml
@@ -55,14 +55,10 @@ query: |
           SELECT
             job,
             if(substr(job, 0, 3) = 'pr:', 'pull', 'ci') kind,  /* pull or ci */
-            version, /* bootstrap git version, empty for ci  */
+            ifnull(repo_commit, version) version, /* bootstrap git version, empty for ci  */
             if(substr(job, 0, 3) = 'pr:',
               regexp_extract(
-                (
-                  select i.value
-                  from t.metadata i
-                  where i.key = 'repos'
-                ),
+                repos,
                 r'[^,]+,\d+:([a-f0-9]+)"'
               ),
               version
@@ -75,13 +71,7 @@ query: |
             and version != 'unknown'
             and (
               (substr(job, 0, 3) = 'ci-' and version != 'unknown') or
-              exists(
-                select as struct
-                  i
-                from t.metadata i
-                where i.key = 'repos' and
-                array_length(split(replace(i.value,', ', ''), ',')) = 2 /*serial pr jobs only (# of PR refs +1 == 2)*/
-              )
+              array_length(split(replace(repos,', ', ''), ',')) = 2 /*serial pr jobs only (# of PR refs +1 == 2)*/
             )
         )
         group by job, commit


cjwagner (approved):

spiffxp (commented):
It looks like we fallback to version if repo_commit is null, is it possible to do the equivalent for repos?

I'm thinking about this in the context of the `flakes-daily` query, and trying to backfill

MushuEE:
I actually originally had it that way but it was just very ugly as both conditions needed `ifnull` and the whole metadata loop. There might be a cleaner way but I do not know SQL well enough yet to know for sure.

spiffxp:
you could push into yet one more subquery to hide the logic of "where should repos come from" from outer queries; it might make performance a bit worse but we're not exactly running this super frequently

how ugly are we talking here? would extra comments help?

MushuEE:
I think the extended query language might allow me to set a var that can be used by the rest of the query

spiffxp:
/close
I'm going to close this given that we merged an "experiment" version of this query in October 2020 (ref: https://github.com/kubernetes/test-infra/pull/19669).  We are currently comparing results of the experiment to the existing query, and iterating on discrepancies

PR #19200, opened 2020-09-14
Update Legacy Endpoints for Delete and Edit

MushuEE:
Started work on updating Legacy endpoints for prow Github client. Updated references also. 

There was a unit test missing for Delete so I added that as well.

/area Prow
/kind cleanup

see #19186 for more context. More end point TODO

image

PR #19211, opened 2020-09-14
Part2: Api endpoint updates for Team and Repo endpoints

MushuEE:
This is the second Part of change #19200. This updates a few more end points but was more complicated due to OrgName not being available in`prow/apis/prowjobs/v1/types.go` Loop. However the functionality in that loop seems to repeat in the following loops so it seemed safe to remove. 

This change consisted mostly of adding `org` args to all endpoints. 

**Question** we were using `org` in place of `owner` (See [ref](https://docs.github.com/en/rest/reference/teams#add-or-update-team-repository-permissions)) in [main.go](https://docs.github.com/en/rest/reference/teams#add-or-update-team-repository-permissions). It looks like the `owner` we use is also `orgName` in code... is that correct?

### DIFF
--- a/prow/apis/prowjobs/v1/types.go
+++ b/prow/apis/prowjobs/v1/types.go
@@ -228,23 +228,14 @@ func (rac *RerunAuthConfig) IsAuthorized(user string, cli prowgithub.RerunClient
 			return true, nil
 		}
 	}
-	for _, ght := range rac.GitHubTeamIDs {
-		member, err := cli.TeamHasMember(ght, user)
-		if err != nil {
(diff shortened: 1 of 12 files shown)


MushuEE:
Looks like I missed some references, will fix tomorrow.

prow/apis/prowjobs/v1/types.go
```diff
@@ -228,23 +228,14 @@ func (rac *RerunAuthConfig) IsAuthorized(user string, cli prowgithub.RerunClient
 			return true, nil
 		}
 	}
-	for _, ght := range rac.GitHubTeamIDs {
```
MushuEE, line 231: What is the difference between a TeamID and TeamSlug?

MushuEE:
Sorry got pulled into other things, will circle back when load lightens

PR #19212, opened 2020-09-14
Update Spyglass docs and configure buildlog lens to highlight start of bazel output.

cjwagner:
Bazel output in the `buildlog` spyglass lens has been a pain to view since `go` started outputing all this garbage: ([e.g.](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/test-infra/19204/pull-test-infra-bazel/1305620250233933830))
```
...
go: downloading github.com/docker/libnetwork v0.5.6
go: extracting github.com/docker/libnetwork v0.5.6
go: finding github.com/docker/libkv v0.2.1
...
```

This PR configures [prow.k8s.io](https://prow.k8s.io) to highlight lines like
> INFO: Analyzed 2772 targets (1666 packages loaded, 26131 targets configured). 

which appear just before the start of the interesting bazel output after all 4000+ lines of go importing noise. Users can then choose to expand the logs below this point rather than only being provided with an option to expand all 4682 lines.

I also updated the docs and improved the scope of some of the other regexes in the lens configs.

/assign @fejta

### DIFF
--- a/config/prow/config.yaml
+++ b/config/prow/config.yaml
@@ -51,27 +51,36 @@ deck:
     - lens:
         name: metadata
       required_files:
-      - started.json|finished.json
+      - ^(?:started|finished)\.json$
       optional_files:
-      - podinfo.json|prowjob.json
+      - ^(?:podinfo|prowjob)\.json$
     - lens:
         name: buildlog
+        config:
+          highlight_regexes:
+          - timed out
+          - "ERROR:"
+          - (FAIL|Failure \[)\b
+          - panic\b
+          - ^E\d{4} \d\d:\d\d:\d\d\.\d\d\d]
+          # This highlights the start of bazel tests/runs to skip go importing noise.
+          - "^INFO: Analyzed \\d+ targets"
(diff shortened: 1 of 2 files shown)


cjwagner:
/assign @cblecker 
In case there are contribex concerns about highlighting this text across all jobs. I think this should be OK to try at least though given that
- the worse case scenario is undesired auto-expansion of extraneous logs
- we can quickly revert if needed using the `highlight_regexes` section from the config

If this does have too broad of an impact we could make the `buildlog` lens accept org/repo specific config like we have done for other types of configs.

fejta (approved):

cjwagner:
Much nicer 😄 https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/test-infra/19204/pull-test-infra-bazel/1305620250233933830

fejta:
very nice!

cblecker:
This looks fine to me.

cc @nikhita if she has any concerns

PR #19216, opened 2020-09-15
Add on demand e2e test for CAPI

fabriziopandini:
https://github.com/kubernetes/test-infra/pull/19092 reduced the number of e2e test we are running on each CAPI PR.
 
However, in some cases it might be necessary/useful to run all the e2e, so this PR makes this possible by adding a new on demand job that could be invoked by adding `/test pull-cluster-api-e2e-full` to comments in the PR.

The job is **optional**, so eventual flakes should not prevent merging PRs

/assign @vincepri 
/cc @sedefsavas 
/cc @wfernandes


fabriziopandini:
CC @sedefsavas

vincepri (approved):
/approve
/lgtm

PR #19226, opened 2020-09-15
kind-conformance-image: use latest kind like the rest of CI

mikedanese:
And find the right kubectl with aquery.

cc @justaugustus @BenTheElder

### DIFF
--- a/experiment/kind-conformance-image-e2e.sh
+++ b/experiment/kind-conformance-image-e2e.sh
@@ -18,15 +18,6 @@ set -o nounset
 set -o pipefail
 set -o xtrace
 
-# set a fixed version so that users of this script manually upgrade kind
-# in a controlled fashion along with the script contents (config, flags...)
-#
-# NOTE: temporarily we are using a specific HEAD commit so that we
-# - get some fixes related to Kubernetes's build changing
-# - don't get surprised when kind changes between now and the next stable release
-# We should switch back to a release tag at the next release.
-STABLE_KIND_VERSION=v0.6.0
-
 # our exit handler (trap)
 cleanup() {
     # always attempt to dump logs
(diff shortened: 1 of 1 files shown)


experiment/kind-conformance-image-e2e.sh
```diff
@@ -48,7 +39,7 @@ cleanup() {
 install_kind() {
     # install `kind` to tempdir
     TMP_DIR=$(mktemp -d)
-    curl -sLo "${TMP_DIR}/kind" https://github.com/kubernetes-sigs/kind/releases/download/${STABLE_KIND_VERSION}/kind-linux-amd64
+    curl -sLo "${TMP_DIR}/kind" https://kind.sigs.k8s.io/dl/latest/kind-linux-amd64
```
mikedanese, line 42: @BenTheElder wdyt with aligning with other CI jobs here? Alternatively, we can tag a new kind release.

BenTheElder:
FYI this job is not blocking for a reason, it's something the sonobuoy folks more or less set up to get coverage for the e2e.test wrapper image but it's not something anyone has been looking after properly. Upstream CI does not use the e2e wrapper image otherwise for various reasons.

I would be in favor of demoting this to CI given the limited need and the lack of upkeep.

experiment/kind-conformance-image-e2e.sh
BenTheElder: There was a new release tagged last night, but switching to latest here is reasonable.

I think this job should be demoted out of release blocking and presubmit given the lack of maintainership, it's frequently failing.
https://testgrid.k8s.io/conformance-all#conformance-image,%20master%20(dev)

For some additional context:

All of the other jobs using this `latest` binary the KIND maintainers actively support and we block kind PRs on running mirrors of the jobs. 

The `latest` (nightly / continuous) binary is only supported for upstream k/k testing where we need to move quicker to work around upstream breaking changes like this.

This image is really only consumed by sonobuoy, which is not upstream.

mikedanese:
@bentheelder thanks. Looks like it only runs if conformance changed:

https://github.com/kubernetes/test-infra/blob/4612a9ca313b199daa4fc782863ccb8d8723a482/config/jobs/kubernetes/sig-testing/conformance-e2e.yaml#L7-L11

cc @dims

experiment/kind-conformance-image-e2e.sh
BenTheElder: Actually, it appears the CI version of this job is not even in release informing currently, let alone blocking.

BenTheElder:
thanks @mikedanese :+1:

PR #19227, opened 2020-09-15
Set global cookiefile once, before running parallel clones.

fejta:
/assign @krzyzacy @cjwagner 

We run the clone for each ref in parallel, which represents a race condition if both attempt to write the file concurrently:

```
$ git config --global http.cookiefile /secrets/cookiefile/cookies
error: could not lock config file /root/.gitconfig: File exists
# Error: exit status 255
```

Even without the race this is wasteful as we only need to do it once. So move this into clonerefs:

### DIFF
--- a/prow/clonerefs/run.go
+++ b/prow/clonerefs/run.go
@@ -65,6 +65,10 @@ func (o Options) Run() error {
 		}
 	}
 
+	if p := needsGlobalCookiePath(o.CookiePath, o.GitRefs...); p != "" {
+		configureGlobalCookiefile(p)
+	}
+
 	var numWorkers int
 	if o.MaxParallelWorkers != 0 {
 		numWorkers = o.MaxParallelWorkers
@@ -117,6 +121,29 @@ func (o Options) Run() error {
 		return fmt.Errorf("one or more of the records are in failed state")
 	}
 
+	return nil
+}
+
+func needsGlobalCookiePath(cookieFile string, refs ...prowapi.Refs) string {
+	if cookieFile == "" || len(refs) == 0 {
+		return ""
+	}
+
+	for _, r := range refs {
+		if !r.SkipSubmodules {
+			return cookieFile
+		}
+	}
+	return ""
+}
+
+// configureGlobalCookiefile ensures git authenticates submodules correctly.
+//
+// Since this is a global setting, we do it once and before running parallel clones.
+func configureGlobalCookiefile(cookiePath string) error {
+	if err := exec.Command("git", "config", "--global", "http.cookiefile", cookiePath).Run(); err != nil {
+		return fmt.Errorf("git config --global http.cookiefile %q: %w", cookiePath, err)
+	}
 	return nil
 }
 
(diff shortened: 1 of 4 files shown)


prow/pod-utils/clone/clone.go
```diff
@@ -180,13 +180,8 @@ func (g *gitCtx) commandsForBaseRef(refs prowapi.Refs, gitUserName, gitUserEmail
...
 	}
-	if cookiePath != "" {
-		if refs.SkipSubmodules {
-			commands = append(commands, g.gitCommand("config", "http.cookiefile", cookiePath))
-		} else {
-			// --global to ensure that all submodules can use this auth
-			commands = append(commands, g.gitCommand("config", "--global", "http.cookiefile", cookiePath))
-		}
+	if cookiePath != "" && refs.SkipSubmodules {
+		commands = append(commands, g.gitCommand("config", "http.cookiefile", cookiePath))
```
cjwagner, line 184: Why do we need to configure this locally when skipping submodules? I don't see how that changes the behavior compared to always configuring globally, especially in the case where multiple refs cause us to configure the cookiefile both globally and locally.

krzyzacy (approved):

prow/clonerefs/run.go
```diff
@@ -120,6 +124,29 @@ func (o Options) Run() error {
...
+	}
+	return ""
+}
+
+// configureGlobalCookiefile ensures git authenticates submodules correctly.
+//
+// Since this is a global setting, we do it once and before running parallel clones.
+func configureGlobalCookiefile(cookiePath string) error {
+	if err := exec.Command("git", "config", "--global", "http.cookiefile", cookiePath).Run(); err != nil {
+		return fmt.Errorf("git config --global http.cookiefile %q: %w", cookiePath, err)
```
cjwagner, line 145: This won't be included in the text clone log or structured clone record anymore (https://github.com/kubernetes/test-infra/blob/master/prow/pod-utils/clone/types.go) which is unfortunate. Including it would require some changes to support an initialization `Record` that isn't associated with refs.

prow/clonerefs/run.go
```diff
@@ -65,6 +65,10 @@ func (o Options) Run() error {
 		}
 	}
 
+	if p := needsGlobalCookiePath(o.CookiePath, o.GitRefs...); p != "" {
+		configureGlobalCookiefile(p)
```
cjwagner, line 69: Handle the error.

prow/pod-utils/clone/clone.go
fejta: If we are not using submodules, we do not need to overwrite the global config.

The local config takes precedence over the global one, and is less likely to have side effects.

prow/clonerefs/run.go
fejta: Addressed in https://github.com/kubernetes/test-infra/pull/19231

prow/clonerefs/run.go
fejta: Addressed in https://github.com/kubernetes/test-infra/pull/19231

PR #19230, opened 2020-09-16
Add Prometheus recording rules and grafana dashboards for SLO compliance.

cjwagner:
This PR provides a mechanism for easily defining and tracking SLO compliance across Prow. Compliance is defined per-component by labelling prometheus alerts as SLO violation indicators. Compliance will be recorded per-component and overall using prometheus recording rules to define synthetic metrics. The SLO metrics will have a value of 1 if SLO compliance is met or a value of 0 if there are alerts indicating an SLO violation. In the case of a violation, the metrics will include the labels from the source alert(s). This will help us answer questions like "What percentage of the time was hook working?", "What percentage of outages were caused by problems with hook?", and "What percentage of the outages caused by hook were due to a specific symptom?".

The PR also includes a dashboard for the SLO metrics. New graph panels will be automatically generated when alerts are marked as SLO indicators for new components.

I've marked some Tide alerts as SLIs to demo and test this. Once I've confirmed things are working well I'll extend this to track SLO for other components as well.

/assign @hongkailiu @MushuEE

### DIFF
--- a/config/prow/cluster/monitoring/mixins/grafana_dashboards/slo.jsonnet
+++ b/config/prow/cluster/monitoring/mixins/grafana_dashboards/slo.jsonnet
@@ -1,0 +1,41 @@
+local config =  import 'config.libsonnet';
+local grafana = import 'grafonnet/grafana.libsonnet';
+local dashboard = grafana.dashboard;
+local graphPanel = grafana.graphPanel;
+local prometheus = grafana.prometheus;
+
+
+local dashboardConfig = {
+        uid: config._config.grafanaDashboardIDs['slo.json'],
+    };
+
+dashboard.new(
+        'SLO Compliance Dashboard',
+        time_from='now-7d',
+        schemaVersion=18,
+      )
+.addPanel(
+    graphPanel.new(
+        'Prow overall SLO compliance',
+        description='slo_prow_ok',
+        datasource='prometheus',
+        legend_rightSide=true,
+    )
+    .addTarget(prometheus.target(
+        'slo_prow_ok'
+    )),
+    gridPos={h: 4, w: 24, x: 0, y: 0})
+.addPanels([
+    graphPanel.new(
+        '%s SLO compliance' % comp,
+        description='slo_component_ok{slo="%s"}' % comp,
+        datasource='prometheus',
+        legend_rightSide=true,
+    )
+    .addTarget(prometheus.target(
+        'slo_component_ok{slo="%s"}' % comp
+    ))
+
+    for comp in config._config.slo.components
+])
++ dashboardConfig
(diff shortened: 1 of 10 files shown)


MushuEE:
QQ, 

> The SLO metrics will have a value of 1 if SLO compliance is met or a value of 0 if there are alerts indicating an SLO violation

Typically I expect 0 to be the rule and 1 to be the exception to the rule like with rc. Was there any particular reason that meeting compliance is 1 and not the other way around?

config/prow/cluster/monitoring/mixins/grafana_dashboards/slo.jsonnet
```diff
@@ -0,0 +1,45 @@
...
+.addPanel(
+    graphPanel.new(
+        'Prow overall SLO compliance',
+        description='slo_prow_ok',
+        datasource='prometheus',
+        legend_rightSide=true,
+    )
+    .addTarget(prometheus.target(
+        'slo_prow_ok'
+    )), gridPos={
```
MushuEE, line 26: Not sure if this was auto formatted but I normally see the Pos params on one line.

config/prow/cluster/monitoring/mixins/grafana_dashboards/slo.jsonnet
```diff
@@ -0,0 +1,45 @@
...
+        uid: config._config.grafanaDashboardIDs['slo.json'],
+    };
+
+dashboard.new(
+        'SLO Compliance Dashboard',
+        time_from='now-7d',
+        schemaVersion=18,
+      )
+.addPanel(
+    graphPanel.new(
```
MushuEE, line 18: Can we add a [link](https://grafana.com/docs/grafana/latest/linking/panel-links/) to anything that defines what might trigger SLO compliance is? I am not sure what docs we have on failures, or maybe a link to a view we can see failures that were triggered?

cjwagner: The time series on the graphs will be labelled with the fields from the relevant alert(s) when not SLO compliant. I think thats should be sufficient for viewing failures.
Once this is demoed I'll add and link some brief docs about how to add alerts to an SLO, how to read the dashboard, how to run more interesting/complex compliance queries, and how to root cause SLO violations. I think the dashboards will be pretty self explanatory and clearly identify the culprit alerts though.

cjwagner:
> Typically I expect 0 to be the rule and 1 to be the exception to the rule like with rc. Was there any particular reason that meeting compliance is 1 and not the other way around?

I don't know what `rc` refers to.
I actually originally wrote queries for both cases (e.g. `slo_component_ok` and `slo_component_violated`), but only one case needs to be recorded because it is trivial to invert the signal by multiplying by -1 and adding 1. I chose recording 1 as compliant because that seemed more appropriate and consistent with other health metrics to me.  I included `_ok` in the metric name to try to indicate that a positive (1, true) signal indicates SLO compliance.

I can change this if it is unexpected, but this seems consistent with other health/liveness metrics to me. Either way should be fine since it is trivial to invert.

config/prow/cluster/monitoring/mixins/grafana_dashboards/slo.jsonnet
cjwagner: The jsonnet files are all manually formatted. I reformatted the `gridPos` parameter to a single line.

MushuEE:
/lgtm
/approve
Thanks!

PR #19240, opened 2020-09-16
community: Use slack-infra staging project images for tempelis jobs

justaugustus:
Switch to `gcr.io/k8s-staging-slack-infra/tempelis:v20200909-1eb97f5`, as
that's the most recent image k-sigs/slack-infra has built (as opposed to
gcr.io/k8s-tempelis/tempelis).

Affects: https://github.com/kubernetes/community/pull/5150

### DIFF
--- a/config/jobs/kubernetes/community/community-presubmit.yaml
+++ b/config/jobs/kubernetes/community/community-presubmit.yaml
@@ -25,7 +25,7 @@ presubmits:
       testgrid-dashboards: sig-contribex-slack-infra
     spec:
       containers:
-      - image: gcr.io/k8s-tempelis/tempelis:v20190412-7c9b6e51
+      - image: gcr.io/k8s-staging-slack-infra/tempelis:v20200909-1eb97f5
         command:
         - /tempelis
         args:
--- a/config/jobs/kubernetes/test-infra/test-infra-trusted.yaml
+++ b/config/jobs/kubernetes/test-infra/test-infra-trusted.yaml
@@ -744,7 +744,7 @@ postsubmits:
       testgrid-dashboards: sig-contribex-slack-infra
     spec:
       containers:
-      - image: gcr.io/k8s-tempelis/tempelis:v20190412-7c9b6e51
+      - image: gcr.io/k8s-staging-slack-infra/tempelis:v20200909-1eb97f5
         command:
         - /tempelis
         args:


config/jobs/kubernetes/community/community-presubmit.yaml
```diff
@@ -25,7 +25,7 @@ presubmits:
       testgrid-dashboards: sig-contribex-slack-infra
     spec:
       containers:
-      - image: gcr.io/k8s-tempelis/tempelis:v20190412-7c9b6e51
+      - image: gcr.io/k8s-staging-slack-infra/tempelis:latest
```
justaugustus, line 28: I opted for `latest` instead of `v20200909-1eb97f5` to prevent recurrences of using an old image, as there is no autobump functionality for this image.

mrbobbytables: Still not super keen on latest, but it might be a good stop gap. I just fear that it'll bite us later. What do you think @nikhita?

justaugustus: While I'm not a fan of using `latest` in general, I think it's a lot more likely that you'll run into an issue w/ outdated images (`v20190412-7c9b6e51`).
Using `latest` here is going to expose errors faster / fix already-patched errors like: https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/community/5150/pull-community-tempelis-check/1306224014775029760

An alternative is image promotion, but someone will still have to keep the config up-to-date here.

nikhita: I'm not a fan of using `latest` either. If we update to `latest` here, we'll need to update the actual postsubmit as well - https://github.com/kubernetes/test-infra/blob/3c707a0f70e8c49df8dbb009327ef980d018b297/config/jobs/kubernetes/test-infra/test-infra-trusted.yaml#L747

Given that the tempelis postsubmit runs in a trusted cluster, I'm not comfortable moving to `latest`... 

Do we know who has access to push images for tempelis?

justaugustus: @nikhita @mrbobbytables -- I've updated both references to use `gcr.io/k8s-staging-slack-infra/tempelis:v20200909-1eb97f5`.

Contributors with access to push to Slack Infra staging are defined here
https://github.com/kubernetes/k8s.io/blob/4595f0515a5534412a1df5cca6689f986c73f5be/groups/groups.yaml#L1267-L1278

nikhita:
Lgtm from me, but will let @mrbobbytables or @jeefy tag since they are OWNERS for slack-infra

justaugustus:
For Slack Infra:
/assign @jeefy

For `config/`:
/assign @dims

dims:
/approve

leaving lgtm to @mrbobbytables

PR #19244, opened 2020-09-16
Github: Add trace-level logging of curl command to easen debugging

alvaroaleman:


stevekuznetsov (approved):
/hold
Do we want to find out if anything cares about there being a space in that log field?

alvaroaleman:
@stevekuznetsov not sure, I can also rename it to just `curl`

alvaroaleman:
@stevekuznetsov updated it to use `curl` as field name rather than `curl command`

petr-muller (approved):

PR #19250, opened 2020-09-16
Validate all Spyglass artifact requests against size limits

e-blackwelder:
Limits can be modified via `deck.spyglass.size_limit` Prow config field.

(Previously, only `ReadAll` operations were validated against the size limit.)

### DIFF
--- a/prow/spyglass/podlogartifact.go
+++ b/prow/spyglass/podlogartifact.go
@@ -96,6 +96,9 @@ func (a *PodLogArtifact) JobPath() string {
 
 // ReadAt implements reading a range of bytes from the pod logs endpoint
 func (a *PodLogArtifact) ReadAt(p []byte, off int64) (n int, err error) {
+	if int64(len(p)) > a.sizeLimit {
+		return 0, lenses.ErrRequestSizeTooLarge
+	}
 	logs, err := a.jobAgent.GetJobLog(a.name, a.buildID, a.container)
 	if err != nil {
 		return 0, fmt.Errorf("error getting pod log: %v", err)
@@ -129,6 +132,9 @@ func (a *PodLogArtifact) ReadAll() ([]byte, error) {
 
 // ReadAtMost reads at most n bytes
 func (a *PodLogArtifact) ReadAtMost(n int64) ([]byte, error) {
+	if n > a.sizeLimit {
+		return nil, lenses.ErrRequestSizeTooLarge
+	}
 	logs, err := a.jobAgent.GetJobLog(a.name, a.buildID, a.container)
 	if err != nil {
 		return nil, fmt.Errorf("error getting pod log: %v", err)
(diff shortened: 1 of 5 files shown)


prow/spyglass/podlogartifact_test.go
```diff
@@ -375,3 +376,79 @@ func TestReadAll_PodLog(t *testing.T) {
...
+}
+
+func (f *fakeAgent) GetJobLog(job, id, container string) ([]byte, error) {
+	return []byte(f.contents), nil
+}
+
+func TestPodLogArtifact_RespectsSizeLimit(t *testing.T) {
+	contents := "Supercalifragilisticexpialidocious"
+	sizeLimit := int64(5)
+	tooManyBytes := int64(sizeLimit * 2)
```
MushuEE, line 395: Shouldn't need to cast this * maintains type afaik

prow/spyglass/podlogartifact_test.go
```diff
@@ -375,3 +376,79 @@ func TestReadAll_PodLog(t *testing.T) {
...
+
+func (f *fakeAgent) GetProwJob(job, id string) (prowapi.ProwJob, error) {
+	return prowapi.ProwJob{}, nil
+}
+
+func (f *fakeAgent) GetJobLog(job, id, container string) ([]byte, error) {
+	return []byte(f.contents), nil
+}
+
+func TestPodLogArtifact_RespectsSizeLimit(t *testing.T) {
```
MushuEE, line 392: Some positive testcases would probably be good just to validate that we are not always erring out

prow/spyglass/podlogartifact.go
```diff
@@ -152,6 +158,9 @@ func (a *PodLogArtifact) ReadAtMost(n int64) ([]byte, error) {
 
 // ReadTail reads the last n bytes of the pod log
 func (a *PodLogArtifact) ReadTail(n int64) ([]byte, error) {
+	if n > a.sizeLimit {
```
MushuEE, line 161: if n > a.sizeLimit but there are less than `a.sizeLimit` total logs, should we still return error?

e-blackwelder: That is a good question and I waffled on the implementation. I finally decided to just keep the implementation simple. Thoughts? Reasons to change it?

If we really wanted to be an awesome API, we would let users query for the size limit directly. (But for reference, we're talking about size limits around 100MB, so ... just don't ask for gobs of data.)

prow/spyglass/podlogartifact_test.go
e-blackwelder: Done.

prow/spyglass/podlogartifact.go
MushuEE: 👍  good with leaving it, simple works and being draconian with rules seems to be the standard so I think it does make sense to not have this fuzzy functionality where things may or may error depending on the underlying logs

MushuEE:
Thanks, Unit test logic is cool

/lgtm

prow/spyglass/podlogartifact.go
e-blackwelder: Agreed. Thanks!

PR #19251, opened 2020-09-16
Dialing down spyglass artifact size limits.

e-blackwelder:
Half a GB per artifact is quite a lot of data to pipe through Deck.

(And it turns out 100 MB is the default sizeLimit anyway.)


spiffxp (approved):
/approve
/lgtm

Issue #5962, opened 2017-12-14
Prow support for triggering jobs based on tag changes


perotinus:
It would be useful to the cluster-registry if Prow could support triggering jobs based on GitHub `CreateEvent"`s.

Concrete use case: I would like to be able to do cluster-registry releases without having to run tools locally on my machine, based on tags created by a GitHub release. If Prow can respond to GitHub `CreateEvent`s, then creating the tag can run the release process/upload binaries/etc. without requiring the person. This allows releases to be done by people who may not have write access to the GCP project that hosts the binaries, simply by creating a 

Subnote: it might be useful to know if a tag is verified, and to only start a release based on a verified, signed tag. cf https://help.github.com/articles/signing-tags-using-gpg/

BenTheElder:
Also CC @kargakis @stevekuznetsov @sebastienvas for potentially interested parties.

As discussed on slack probably the most tricky part to supporting tag-triggered-jobs is deciding how we want the config to look.

I'm not sure about the GPG checking details wit github but this also sounds potentially useful

/area prow
/kind feature

stevekuznetsov:
If we want to use these jobs for building and pushing releases we should probably get a trusted build cluster up and running otherwise your ProwJob running on a new tag won't have the credentials to do anything meaningful with whatever it builds.

BenTheElder:
Yeah, it would be preferable to have #5848 first. We definitely plan to implement that one though.

BenTheElder:
this would be nice to have for github releases though to further automate publishing of tagged/named release versions, the main reason we can't implement this is how cludgey the triggering logic is now :( 

this should basically just be a post-submit triggered on tags being pushed.

/remove-lifecycle rotten

BenTheElder:
see this thread: https://github.com/kubernetes/test-infra/pull/10802#discussion_r248715913

cjwagner:
So we already (inadvertently) support triggering jobs based on tag changes by using a postsubmit job with a branch regexp matching `refs/tags/<tag-name>`?
That is an interesting find and a pleasant surprise. I think that means we can close this issue.

zhouhaibing089:
> So we already (inadvertently) support triggering jobs based on tag changes by using a postsubmit job with a branch regexp matching refs/tags/<tag-name>?

I think `refs/tags/` is [trimmed][1]..

[...]

Issue #19242, opened 2020-09-16
Triggering image build job on new tags.


prameshj:
Support for building images from k8s.io/dns was added in - https://github.com/kubernetes/test-infra/pull/17123
The build job runs everytime there is a new PR merged into master branch. The _GIT_TAG is of the form - "v20200416-1.15.12-2-g232a189"
This is the cloudbuild yaml - https://github.com/kubernetes/dns/blob/master/cloudbuild.yaml

I want to configure this so that, when there is a new tag added to the repo, a build id triggered with _GIT_TAG=<value of the newly added tag>
Is this supported? 
I am able to manually submit a build with the tag value like:
```
gcloud builds submit ~/src/k8s.io/dns --config cloudbuild.yaml --substitutions _PULL_BASE_REF=master,_GIT_TAG=1.15.14 --project k8s-staging-dns --gcs-log-dir gs://k8s-staging-dns-gcb/logs --gcs-source-staging-dir gs://k8s-staging-dns-gcb/source
```
It would be great to automate this.
Any thoughts on how to achieve this?

@fejta @listx

justaugustus:
I think adding a SemVer regex in the `branches` list of your jobs should work.

ref: https://github.com/kubernetes/test-infra/blob/f316ca27125ebdcfe93ef92cc9cd119269f09fad/config/jobs/image-pushing/k8s-staging-cluster-api.yaml#L14

PR #19254, opened 2020-09-16
Update DNS image push yaml to trigger on new tags

prameshj:
Using '^\d+\.\d+\.\d+$' instead of 'refs/tags/\d+\.\d+\.\d+$' since it appears the prefix might be getting trimmed:
https://github.com/kubernetes/test-infra/issues/5962#issuecomment-657715030

/assign @fejta @BenTheElder @cjwagner

### DIFF
--- a/config/jobs/image-pushing/k8s-staging-dns.yaml
+++ b/config/jobs/image-pushing/k8s-staging-dns.yaml
@@ -7,8 +7,10 @@ postsubmits:
         # this is the name of some testgrid dashboard to report to.
         testgrid-dashboards: sig-network-dns
       decorate: true
+      # match on regex '^\d+\.\d+\.\d+$' to trigger job on new tags
       branches:
         - ^master$
+        - ^\d+\.\d+\.\d+$
       spec:
         serviceAccountName: gcb-builder
         containers:


prameshj:
Fixes issue - https://github.com/kubernetes/test-infra/issues/19242

fejta (approved):

prameshj:
When would these changes take effect? I pushed a new tag to the DNS repo, but i am not seeing a new job kick in yet. - https://k8s-testgrid.appspot.com/sig-network-dns#dns-push-images
Does it require an update to pick up these changes?
@fejta

spiffxp:
testgrid doesn't necessarily update as quickly, if you need to watch how things are going as quickly as possible keep an eye on https://prow.k8s.io (e.g. https://prow.k8s.io/?job=dns-push-images)

spiffxp:
(and it looks like it did trigger, for tag 1.2.3)

prameshj:
Thanks Aaron! I see it too.. sorry for the false alarm.
I see the image tag generated is "v20200918-1.2.3", I was hoping to get just "1.2.3". Is any other config change required? This is the cloudbuild file i used - https://github.com/kubernetes/dns/blob/master/cloudbuild.yaml

spiffxp:
Based on https://storage.googleapis.com/kubernetes-jenkins/logs/dns-push-images/1306945097224425472/podinfo.json I'm going to guess you want to use `PULL_BASE_REF`

PR #19258, opened 2020-09-17
Move dashpole to emeritus for node

dashpole:
Follow-up to https://github.com/kubernetes/kubernetes/pull/94840.  I am refocusing on instrumentation for the time being.

I would remove myself from configs/jobs/cadvisor/OWNERS, but i'm the only one in that file.  :)  I'll keep it for now.

/assign @SergeyKanzhelev @derekwaynecarr  

cc @dims @dchen1107 @sjenning


spiffxp (approved):
/approve
/lgtm
Thanks for all you've done for node!

PR #19260, opened 2020-09-17
disable pull-azurefile-csi-driver-e2e-migration

andyzhangx:
there is a bug in azure file csi migration, disable this test first


ZeroMagic (approved):
/lgtm

PR #19261, opened 2020-09-17
Revert memory bump in ci-kubernetes-e2e-gce-scale-correctness

mborsz:
After revert https://github.com/kubernetes/kubernetes/pull/94800, the test takes ~12-15GiB now, so 48GiB should be really safe.

/assign @mm4tt


mm4tt:
/lgtm

Nice!

PR #19268, opened 2020-09-17
Add containerd to managed_webhooks

dmcgowan:
Use managed webhooks for containerd/containerd

containerd/cri already has a manually created webhook here. Going forward we only intend to add containerd/containerd and not other under the containerd org.

@dims @dmcgowan @fejta


fejta (approved):
/ok-to-test

dmcgowan:
cla should be good now

PR #19271, opened 2020-09-18
Switch config_updater configuration to use new format

### LINKED ISSUE #15662: config-updater needs to use clusters instead of namespaces
**What happened**:
The logs for `hook`, `deck`, and `status-reconciler` are all very loud about a deprecated format:
```
jsonPayload: {
  component: "hook"   
  file: "prow/plugins/config.go:865"   
  func: "k8s.io/test-infra/prow/plugins.(*ConfigUpdater).SetDefaults"   
  level: "warning"   
  msg: "'namespace' and 'additional_namespaces' are deprecated for config-updater plugin, use 'clusters' instead"   
 }
```

It looks like there was an attempt to move to a new format, but there were issues with it. See https://github.com/kubernetes/test-infra/pull/15632

These warnings are crowding out most other warnings. I don't expect this many warnings about the same thing over and over again.

matthyx:
fixes #15662

In order to work we need to mount kubeconfig into `hook` https://github.com/kubernetes/test-infra/issues/15662#issuecomment-694417485

### DIFF
--- a/config/prow-staging/cluster/hook_deployment.yaml
+++ b/config/prow-staging/cluster/hook_deployment.yaml
@@ -47,6 +47,7 @@ spec:
         - --github-token-path=/etc/github/oauth
         - --config-path=/etc/config/config.yaml
         - --job-config-path=/etc/job-config
+        - --kubeconfig=/etc/kubeconfig/config
         ports:
           - name: http
             containerPort: 8888
@@ -65,6 +66,9 @@ spec:
           readOnly: true
         - name: plugins
           mountPath: /etc/plugins
+          readOnly: true
+        - name: kubeconfig
+          mountPath: /etc/kubeconfig
           readOnly: true
         livenessProbe:
           httpGet:
@@ -95,3 +99,7 @@ spec:
       - name: plugins
         configMap:
           name: plugins
+      - name: kubeconfig
+        secret:
+          defaultMode: 420
+          secretName: kubeconfig
--- a/config/prow/plugins.yaml
+++ b/config/prow/plugins.yaml
@@ -446,7 +446,9 @@ config_updater:
   maps:
     label_sync/labels.yaml:
       name: label-config
-      namespace: test-pods
+      clusters:
+        test-infra-trusted:
+          - test-pods
     config/prow/config.yaml:
       name: config
     config/prow/plugins.yaml:
--- a/config/prow/plugins.yaml
+++ b/config/prow/plugins.yaml
@@ -451,14 +451,26 @@ config_updater:
           - test-pods
     config/prow/config.yaml:
       name: config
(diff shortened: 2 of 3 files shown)


matthyx:
/hold
need confirmation from Cole
/cc @cjwagner

config/prow/plugins.yaml
```diff
@@ -446,7 +446,9 @@ config_updater:
   maps:
     label_sync/labels.yaml:
       name: label-config
-      namespace: test-pods
+      clusters:
+        test-infra-trusted:
+          - test-pods
     config/prow/config.yaml:
       name: config
```
cjwagner, line 453: I think we actually need to make this change for all these entries.
> Currently things are working because the default cluster is used which is the in-cluster context since the kubeconfig is omitted. ... I'm pretty sure mounting the kubeconfig will change what the default context points to because that will become the build cluster.

So I think this PR would currently cause the other configmaps managed here to start deploying to the build cluster instead of the service cluster. To avoid that we should probably specify the following for each:
```suggestion
    config/prow/config.yaml:
      name: config
      clusters:
        test-infra-trusted:
          - default
```
I'd prefer if we had an explicit in-cluster context, but I'm not sure how that would work with our current kubeconfig parsing logic.

matthyx: I see... Maybe we should update the kubecofig parsing logic first?

matthyx: > I think we actually need to make this change for all these entries.

Like that?

config/prow-staging/cluster/hook_deployment.yaml
```diff
@@ -66,6 +66,9 @@ spec:
         - name: plugins
           mountPath: /etc/plugins
           readOnly: true
+        - name: kubeconfig
+          mountPath: /etc/kubeconfig
```
cjwagner, line 70: `--kubeconfig` flag is also needed.

config/prow/plugins.yaml
cjwagner: Yes, I think so.

config/prow-staging/cluster/hook_deployment.yaml
matthyx: done

cjwagner (approved):
Holding for oncall to merge. Once this is deployed we should observe an update to one of the configmaps and ensure the update is deployed to the correct location. If there is problem with the new config we'll need to revert and redeploy the old config by manually editing the configmap or by using the [`config-bootstrapper`](https://github.com/kubernetes/test-infra/tree/master/prow/cmd/config-bootstrapper#config-bootstrapper).
/hold
/assign @chases2 @michelle192837

PR #19275, opened 2020-09-18
Update Azure e2e focus to allow parallel Ginkgo nodes

mboersma:
I've tested this change locally and through CAPZ's e2e. Focusing on the top-level `Describe` container prevents ginkgo from spawning multiple nodes, but finding the `Context`s under that works as expected:

```
Running in parallel across 3 nodes
```

### DIFF
--- a/config/jobs/kubernetes-sigs/cluster-api-provider-azure/cluster-api-provider-azure-presubmits.yaml
+++ b/config/jobs/kubernetes-sigs/cluster-api-provider-azure/cluster-api-provider-azure-presubmits.yaml
@@ -81,7 +81,7 @@ presubmits:
         - "./scripts/ci-e2e.sh"
         env:
           - name: GINKGO_FOCUS
-            value: "Workload cluster creation"
+            value: "Creating .* control-plane cluster"
         # docker-in-docker needs privileged mode
         securityContext:
           privileged: true


config/jobs/kubernetes-sigs/cluster-api-provider-azure/cluster-api-provider-azure-presubmits.yaml
```diff
@@ -81,7 +81,7 @@ presubmits:
         - "./scripts/ci-e2e.sh"
         env:
           - name: GINKGO_FOCUS
-            value: "Workload cluster creation"
+            value: "Creating .* control-plane cluster"
```
CecileRobertMichon, line 84: does this need to have control-plane in there? If not `Creating .* cluster` might be more suited in some cases like "create a managed cluster"

CecileRobertMichon (commented):
Does https://github.com/kubernetes/test-infra/blob/c0b620f67a0d7622747becad65d2aefa5f5e39cf/config/jobs/kubernetes-sigs/cluster-api-provider-azure/cluster-api-provider-azure-periodics.yaml#L66 also need to change or is it good as is?

config/jobs/kubernetes-sigs/cluster-api-provider-azure/cluster-api-provider-azure-presubmits.yaml
mboersma: No, it doesn't need to have "control-plane," I was just imagining possible accidental matches so I made it as specific as possible. But we can relax it--originally I had `Creating .* cluster`.

mboersma:
Oops, missed that periodic one. I don't see the parallel nodes message when I use that as `-focus`. Let me see if I can tweak that one as well.

CecileRobertMichon:
there's also https://github.com/kubernetes/test-infra/blob/c0b620f67a0d7622747becad65d2aefa5f5e39cf/config/jobs/kubernetes-sigs/cluster-api-provider-azure/cluster-api-provider-azure-presubmits.yaml#L112

mboersma:
I am realizing this entire thing boils down to lack of quoting in our Makefile. With `-focus="Workload cluster creation"` we are getting parallel nodes, but with `-focus=Workload cluster creation` we do not. Regardless, it looks like the fix should be entirely in CAPZ, since I can make all the `-focus` arguments you pointed to run with parallel nodes if quoted properly.

Humiliated? No. Humbled? Yes. For lack of quotation marks, the kingdom was nearly lost. Back to work!

Issue #18546, opened 2020-07-30
Increase monitoring.prow.k8s.io retention to 1y


spiffxp:
It has been set to 30d, which means we don't have the ability to look at trends over a full kubernetes release cycle, let alone comparing against a prior release cycle.

Trends I'm particularly interested in
- what's the size of the tide pool like over time?
- how many pods have hit error state over time?
- what's our boskos pool usage over time?

https://github.com/kubernetes/test-infra/pull/18545 increased to 90d

Will likely need to increase PV size to fit 1y of data.

/area prow

BenTheElder:
cc @cjwagner @MushuEE

MushuEE:
I am not super familiar with PVC using VCT, does that dynamically scale? I think I have only used the VolumeClaimTemplate in the context of StatefulSets

PR #19276, opened 2020-09-18
Raise prow prometheus retention to 180d

spiffxp:
Part of https://github.com/kubernetes/test-infra/issues/18546

/cc @cjwagner

### DIFF
--- a/config/prow/cluster/monitoring/prow_prometheus.yaml
+++ b/config/prow/cluster/monitoring/prow_prometheus.yaml
@@ -16,7 +16,7 @@ spec:
         resources:
           requests:
             storage: 100Gi
-  retention: "90d"
+  retention: "180d"
   serviceAccountName: prometheus-prow
   alerting:
     alertmanagers:


MushuEE:
/hold

not sure if you want @cjwagner to verify

config/prow/cluster/monitoring/prow_prometheus.yaml
```diff
@@ -16,7 +16,7 @@ spec:
         resources:
           requests:
             storage: 100Gi
```
cjwagner, line 18: Do we have any idea how much of this we're using?

cjwagner (approved):
/hold

spiffxp:
I don't have access to check in on the PV itself.  Going off of prometheus' own metrics, storage consumed for its data has grown as follows:

Screen Shot 2020-09-22 at 4 57 07 PM

We're at 40gb now after a little under 90d.  Doesn't look like we'll hit our 100Gi limit in 180d.  We'll need to resize if we want retention any further than that.

WDYT @cjwagner

config/prow/cluster/monitoring/prow_prometheus.yaml
spiffxp: I don't have access to check in on the PV itself.  Going off of prometheus' own metrics, storage consumed for its data has grown as follows:

Screen Shot 2020-09-22 at 4 57 07 PM

We're at 40gb now after a little under 90d.  Doesn't look like we'll hit our 100Gi limit in 180d.  We'll need to resize if we want retention any further than that.

WDYT @cjwagner

cjwagner: Sounds safe to retain 180d then.
If resizing necessitates restarting with no data then we should resize sooner rather than later, but presumably we can instead migrate the data to new PVs or resize the PVs in place? I'm not sure what the process would be for a PV managed by the prometheus operator.

spiffxp: It sounds like it's doable manually, and there may be a stalled upstream KEP preventing it from happening automatically (ref: https://github.com/prometheus-operator/prometheus-operator/issues/2357)

spiffxp: Opened https://github.com/kubernetes/test-infra/issues/19325 for followup

PR #19283, opened 2020-09-19
Add experimental query to trial run flake using new repos field

MushuEE:
This is just a port of #19209 to trial run query without altering existing query.

/area kettle
/assign @spiffxp


MushuEE:
/will try to fix CLA, maybe email mis-match between machines

cjwagner (approved):

PR #19302, opened 2020-09-22
Upgrade to COS M85 and use image_family

bsdnet:

For some reason PR #19294 's patch is overrided.

Something wrong with my github workflow :)

### DIFF
--- a/jobs/e2e_node/image-config-serial.yaml
+++ b/jobs/e2e_node/image-config-serial.yaml
@@ -6,7 +6,7 @@ images:
     image: ubuntu-gke-1804-1-16-v20200330 # docker 17.03
     project: ubuntu-os-gke-cloud
   cos-stable2:
-    image: cos-77-12371-274-0 # docker v19.03.1, deprecated after 2020-12-17
+    image_family: cos-81-lts # deprecated after 2021-06-24
     project: cos-cloud
     metadata: "user-data<test/e2e_node/jenkins/gci-init-gpu.yaml,gci-update-strategy=update_disabled"
     resources:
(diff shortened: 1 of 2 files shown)


bsdnet:
//assign @dims  @karan

jobs/e2e_node/image-config-serial.yaml
```diff
@@ -6,14 +6,14 @@ images:
     image: ubuntu-gke-1804-1-16-v20200330 # docker 17.03
     project: ubuntu-os-gke-cloud
   cos-stable2:
-    image: cos-77-12371-274-0 # docker v19.03.1, deprecated after 2020-12-17
+    image_family: cos-81-lts # docker v19.03.6, deprecated after 2021-06-24
```
SergeyKanzhelev, line 9: Last PR we removed the docker version from the comment as it doesn't affect anything. Can we do the same here?

jobs/e2e_node/perf-image-config.yaml
```diff
@@ -1,7 +1,7 @@
 ---
 images:
   cos-stable1:
-    image: cos-81-12871-119-0 # docker 19.03.6, current latest LTS.
+    image_family: cos-85-lts # docker 19.03.9, current latest LTS.
```
SergeyKanzhelev, line 4: let's add deprecated after instead

SergeyKanzhelev:
/lgtm 

maybe improve some comments

jobs/e2e_node/image-config-serial.yaml
bsdnet: Sure, I can remove it in next patch. 
I do not see docker, or deprecation matters, since those info are changing and also listed in COS release notes

jobs/e2e_node/perf-image-config.yaml
bsdnet: I am thinking to remove all the comments, WDYT?

PR #19319, opened 2020-09-23
Upgrade cos from 77 to 81 and use image_family

bsdnet:

cos 77 will deprecate soon, and move cos-stable2 to cos-81


MHBauer (approved):
I'm super happy we've fixed all the bugs and ended up here!

/lgtm

PR #19344, opened 2020-09-24
[gcsweb] add the content-type in headers when handling an object

droslean:
/cc @stevekuznetsov 

### DIFF
--- a/gcsweb/cmd/gcsweb/gcsweb.go
+++ b/gcsweb/cmd/gcsweb/gcsweb.go
@@ -287,7 +287,14 @@ type server struct {
 	storageClient *storage.Client
 }
 
-func (s *server) handleObject(w http.ResponseWriter, bucket, object string) error {
+type objectHeaders struct {
+	contentType        string
+	contentEncoding    string
+	contentDisposition string
+	contentLanguage    string
+}
+
+func (s *server) handleObject(w http.ResponseWriter, bucket, object string, headers objectHeaders) error {
 	obj := s.storageClient.Bucket(bucket).Object(object)
 
 	objReader, err := obj.NewReader(context.Background())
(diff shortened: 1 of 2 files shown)


gcsweb/cmd/gcsweb/gcsweb.go
```diff
@@ -371,7 +373,7 @@ func (s *server) gcsRequest(w http.ResponseWriter, r *http.Request) {
 
 	// This means that the object is a file.
 	if objAttrs != nil {
-		if err := s.handleObject(w, bucket, object); err != nil {
+		if err := s.handleObject(w, bucket, object, objAttrs.ContentType); err != nil {
```
wking, line 376: Probably want to pass through `ContentEncoding` too?

droslean: I include also the encoding.

gcsweb/cmd/gcsweb/gcsweb_test.go
```diff
@@ -209,7 +209,7 @@ func TestHandleObject(t *testing.T) {
 
 			s := server{storageClient: client}
 
-			err = s.handleObject(w, tc.bucket, tc.object)
+			err = s.handleObject(w, tc.bucket, tc.object, objectHeaders{})
```
stevekuznetsov, line 212: Ideally we would check that headers are set  correctly, too, in the test

droslean: Yea that will need to actually have different file content and check the desired headers as well.

droslean: We will need to change the mock server also to generate the headers as well. I will need to study their docs etc. This sounds like a nice follow-up

droslean: addressed

stevekuznetsov (approved):

PR #19350, opened 2020-09-24
Move capi link verification to periodic

benmoss:
Fast feedback that this job is providing too much noise as a presubmit. It doesn't help that it's failing since we're getting rate-limited by GitHub, but I will have a separate PR to introduce a GitHub token for this job to consume once that is made available in the Prow environment.

@ncdc @vincepri


vincepri (approved):
/approve

PR #19353, opened 2020-09-24
add BoundServiceAccountTokenVolume upgrade test config

zshihang:
add a periodic job config for alpha feature BoundServiceAccountTokenVolume upgrade test.

the test is added in https://github.com/kubernetes/kubernetes/pull/94835.

### DIFF
--- a/config/jobs/kubernetes/sig-auth/serviceaccount-admission-controller-migration-config.yaml
+++ b/config/jobs/kubernetes/sig-auth/serviceaccount-admission-controller-migration-config.yaml
@@ -1,0 +1,33 @@
+periodics:
+- name: sig-auth-serviceaccount-admission-controller-migration
+  annotations:
+    testgrid-dashboards: sig-auth-gce
+    testgrid-tab-name: upgrade-tests
+    # TODO: use sig-auth mailing list
+    testgrid-alert-email: zshihang@google.com
+    testgrid-num-failures-to-alert: "3"
+  labels:
+    preset-service-account: "true"
+    preset-k8s-ssh: "true"
+  decorate: true
+  # TODO: lower to once a day
+  interval: 2h
+  spec:
+    containers:
+      - image: gcr.io/k8s-testimages/kubekins-e2e:v20201014-d8fda77-master
+        args:
+        - --timeout=920
+        - --bare
+        - --scenario=kubernetes_e2e
+        - --
+        - --check-leaked-resources
+        - --check-version-skew=false
+        - --env=STORAGE_MEDIA_TYPE=application/vnd.kubernetes.protobuf
+        # TODO: change to release/stable-x.y
+        - --extract=ci/latest
+        - --extract=ci/latest-1.19
+        - --gcp-node-image=gci
+        - --gcp-zone=us-west1-b
+        - --provider=gce
+        - --upgrade_args="--ginkgo.focus=\[Feature:BoundServiceAccountTokenVolume\] --upgrade-target=ci/latest --upgrade-image=gci"
+        - --timeout=60m
--- a/config/jobs/kubernetes/sig-auth/OWNERS
+++ b/config/jobs/kubernetes/sig-auth/OWNERS
@@ -1,0 +1,9 @@
+# See the OWNERS docs at https://go.k8s.io/owners
+approvers:
+- liggitt
+- mikedanese
+reviewers:
+- liggitt
+- mikedanese
+labels:
+- sig/auth


config/jobs/kubernetes/sig-auth/serviceaccount-admission-controller-migration-config.yaml
```diff
@@ -0,0 +1,23 @@
...
+    preset-k8s-ssh: "true"
+  decorate: true
+  interval: 2h
+  spec:
+    containers:
+      image: gcr.io/k8s-testimages/kubekins-e2e:v20200918-58bdf8d-master
+    - args:
+      - --timeout=920
+      - --bare
+      - --scenario=kubernetes_e2e
```
liggitt, line 16: is this testing an upgrade or just a regular e2e test?

liggitt:
I'm a little confused about whether this is actually setting up an upgrade test... I expected something more like this:

https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/sig-cloud-provider/gcp/upgrade-gce.yaml#L18-L25

config/jobs/kubernetes/sig-auth/serviceaccount-admission-controller-migration-config.yaml
zshihang: it is a upgrade test even though there is no upgrade_args specified.

since `--test_args="--ginkgo.focus=\[Feature:BoundServiceAccountTokenVolume\]"` will select the upgrade test for this feature, it will run the upgrade test framework. the first target would be `--extract=ci/latest-1.19` and the unspecified target would be `ci/latest`.

liggitt: I see... can we make the upgrade target more explicit in the job config?

zshihang: done

config/jobs/kubernetes/sig-auth/serviceaccount-admission-controller-migration-config.yaml
```diff
@@ -0,0 +1,27 @@
+periodics:
+- name: serviceaccount-admission-controller-migration
+  annotations:
+    testgrid-dashboards: sig-auth-gce
+  labels:
+    preset-service-account: "true"
+    preset-k8s-ssh: "true"
+  decorate: true
+  interval: 2h
```
liggitt, line 9: I'm ok with this to start until we get the job green, then we should probably lower it to ~1 time a day

config/jobs/kubernetes/sig-auth/serviceaccount-admission-controller-migration-config.yaml
```diff
@@ -0,0 +1,27 @@
+periodics:
+- name: serviceaccount-admission-controller-migration
```
liggitt, line 2: not sure if there's a convention for naming these. I'd tend to prefix with sig-auth. Thoughts, @BenTheElder ?

config/jobs/kubernetes/sig-auth/serviceaccount-admission-controller-migration-config.yaml
```diff
@@ -0,0 +1,27 @@
...
+  annotations:
+    testgrid-dashboards: sig-auth-gce
+  labels:
+    preset-service-account: "true"
+    preset-k8s-ssh: "true"
+  decorate: true
+  interval: 2h
+  spec:
+    containers:
+      - image: gcr.io/k8s-testimages/kubekins-e2e:v20200918-58bdf8d-master
```
liggitt, line 12: double check this is the current kubekins image

config/jobs/kubernetes/sig-auth/serviceaccount-admission-controller-migration-config.yaml
```diff
@@ -0,0 +1,27 @@
...
+        args:
+        - --timeout=920
+        - --bare
+        - --scenario=kubernetes_e2e
+        - --
+        - --check-leaked-resources
+        - --check-version-skew=false
+        - --env=STORAGE_MEDIA_TYPE=application/vnd.kubernetes.protobuf
+        - --extract=ci/latest
+        - --extract=ci/latest-1.19
```
liggitt, line 22: what will this do once ci/latest is 1.21? Should we be using stable1 markers? cc @justaugustus

liggitt (commented):
A few questions on the job name and whether we should be using pinned versions (latest-1.19) or floating versions (stable1 and latest), but looks good in general

config/jobs/kubernetes/sig-auth/serviceaccount-admission-controller-migration-config.yaml
zshihang: ack

config/jobs/kubernetes/sig-auth/serviceaccount-admission-controller-migration-config.yaml
zshihang: prepended with sig-auth

config/jobs/kubernetes/sig-auth/serviceaccount-admission-controller-migration-config.yaml
zshihang: done

config/jobs/kubernetes/sig-auth/serviceaccount-admission-controller-migration-config.yaml
zshihang: use ci/latest-1.19 and ci/latest so that the hash and coredns fix are included. we can change them later if appropriate.

liggitt: Sounds good, let's make sure that's in an issue to track

config/jobs/kubernetes/sig-auth/serviceaccount-admission-controller-migration-config.yaml
```diff
@@ -0,0 +1,28 @@
+periodics:
+- name: sig-auth-serviceaccount-admission-controller-migration
+  annotations:
+    testgrid-dashboards: sig-auth-gce
+    testgrid-tab-name: gce
+  labels:
+    preset-service-account: "true"
+    preset-k8s-ssh: "true"
+  decorate: true
+  interval: 2h
```
liggitt, line 10: Oh, I forgot, we want to set up notifications for this job. This can be individuals until we get the job green, then should transition to the sig-auth list. See use of testgrid-alert-email in other jobs for examples, and include the mailing list transition in the follow-up tracking issue

zshihang: done

config/jobs/kubernetes/sig-auth/serviceaccount-admission-controller-migration-config.yaml
zshihang: done. https://github.com/kubernetes/test-infra/issues/19575

liggitt:
/lgtm

To testing reviewers/approvers

spiffxp (approved):
/approve

liggitt:
> /hold for #19353 to merge

@dims... that's a self-reference... is the thing you meant to refer to merged? :-)

zshihang:
/hold cancel

the test is cherry picked into 1.19

config/jobs/kubernetes/sig-auth/serviceaccount-admission-controller-migration-config.yaml
```diff
@@ -0,0 +1,33 @@
...
+  # TODO: lower to once a day
+  interval: 2h
+  spec:
+    containers:
+      - image: gcr.io/k8s-testimages/kubekins-e2e:v20201014-d8fda77-master
+        args:
+        - --timeout=920
+        - --bare
+        - --scenario=kubernetes_e2e
+        - --
```
spiffxp, line 22: I missed this... these are bootstrap.py parameters, they shouldn't be here if `decorate: true` is being used.

It's more preferable to use `decorate: true` than to revert to bootstrap.py, as bootstrap has been deprecated for years

config/jobs/kubernetes/sig-auth/serviceaccount-admission-controller-migration-config.yaml
```diff
@@ -0,0 +1,33 @@
...
+        - --check-leaked-resources
+        - --check-version-skew=false
+        - --env=STORAGE_MEDIA_TYPE=application/vnd.kubernetes.protobuf
```
spiffxp, line 25: This isn't consumed by kubetest, you would want to set this as an `env:` on the container instead

config/jobs/kubernetes/sig-auth/serviceaccount-admission-controller-migration-config.yaml
spiffxp: See https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/sig-cloud-provider/azure/sig-azure-config.yaml#L73-L82 for an example of what `command:` and `args:` should be instead

spiffxp (commented):
I missed that this was using `decorate: true` and bootstrap.py at the same time.  I left review comments on how to migrate away from bootstrap.  I haven't personally tested them, so if you'd rather drop `decorate: true` that's fine too.

config/jobs/kubernetes/sig-auth/serviceaccount-admission-controller-migration-config.yaml
zshihang: done in the followup

config/jobs/kubernetes/sig-auth/serviceaccount-admission-controller-migration-config.yaml
zshihang: done. thank you for the review.

dims:
> > /hold for #19353 to merge
> 
> @dims... that's a self-reference... is the thing you meant to refer to merged? :-)

Whoops!

PR #19354, opened 2020-09-24
post-test-infra-push-prow: run when GCSWeb changes

stevekuznetsov:
The job publishes the GCSWeb image, so it needs to run when the source
changes.


petr-muller (approved):
/hold

PR #19355, opened 2020-09-24
Add github token to CAPI periodic link check job

benmoss:
This will keep us from getting rate-limited by GitHub, hopefully :smile_cat: 

@vincepri


vincepri (commented):
/assign @ncdc
/approve

vincepri:
I assume we've taken the necessary precaution to make sure the token isn't leaked?

benmoss:
> I assume we've taken the necessary precaution to make sure the token isn't leaked?

The make target doesn't print the token, and I can't think of what else would. It is a completely unprivileged token from an account that doesn't have access to any repositories. I'll share the credentials so others can manage/regenerate the token if need be.

Issue #19359, opened 2020-09-25
Malformed Junit in Prow job breaking Kettle Parsing 


MushuEE:
**What happened**:
Kettle command:
`make_json.py  --days 7 | pv | gzip > build_week.json.gz` fails with 
```
ERROR:root:error on gs://kubernetes-jenkins/logs/ci-kubernetes-e2e-gci-gce-network-proxy-grpc/1307184419391410176
Traceback (most recent call last):
  File "make_json.py", line 274, in make_rows
    yield rowid, row_for_build(path, started, finished, results)
  File "make_json.py", line 214, in row_for_build
    for test in parse_junit(result):
  File "make_json.py", line 152, in parse_junit
    tree = ET.fromstring(xml)
  File "/usr/lib/python3.6/xml/etree/ElementTree.py", line 1315, in XML
    return parser.close()
  File "<string>", line None
xml.etree.ElementTree.ParseError: unclosed token: line 305, column 6
```

due to malformed JSON at https://storage.googleapis.com/kubernetes-jenkins/logs/ci-kubernetes-e2e-gci-gce-network-proxy-grpc/1307184419391410176/artifacts/junit_01.xml

![image](https://user-images.githubusercontent.com/7189267/94224025-4098bf80-fea6-11ea-8c7e-b53e12ed8002.png)
[JOB](https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/ci-kubernetes-e2e-gci-gce-network-proxy-grpc/1307184419391410176)

**What you expected to happen**:
Correct XML format in job data and Kettle to pass on Malformed data.

**How to reproduce it (as minimally and precisely as possible)**:
`kubectl exec -it kettle-<running hash> bash` 
`python3 make_json.py  --days 7 | pv | gzip > build_week.json.gz`

**Please provide links to example occurrences, if any**:
ABOVE

**Anything else we need to know?**:

/cc @spiffxp @BenTheElder @RobertKielty

PR #19364, opened 2020-09-25
Fix several issues in Kettle

MushuEE:
First, kettle would die if any malformed xml was collected in a job
This is now skipped for failing jobs

Pypy was being installed as pypy3

Also fully reverted cmd change incase it was an issue in writing to BQ
```
root@kettle-57f5b7f9f-h4ph2:/kettle# ln -s /opt/pypy*/bin/pypy3 /usr/bin
root@kettle-57f5b7f9f-h4ph2:/kettle# pypy
bash: pypy: command not found
```
```
root@kettle-57f5b7f9f-h4ph2:/kettle# pypy3
Python 3.6.9 (2ad108f17bdb, Apr 07 2020, 02:59:05)
[PyPy 7.3.1 with GCC 7.3.1 20180303 (Red Hat 7.3.1-5)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>>>
```
```
root@kettle-57f5b7f9f-h4ph2:/kettle# ls -lart /usr/bin/pypy*
lrwxrwxrwx 1 root root 19 Sep 24 14:25 /usr/bin/pypy -> '/opt/pypy*/bin/pypy'
lrwxrwxrwx 1 root root 37 Sep 25 10:08 /usr/bin/pypy3 -> /opt/pypy3.6-v7.3.1-linux64/bin/pypy3
```

#19359
/area kettle
/cc @RobertKielty @spiffxp

### DIFF
--- a/kettle/Dockerfile
+++ b/kettle/Dockerfile
@@ -33,7 +33,7 @@ RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime \
 RUN pip3 install requests google-cloud-pubsub==0.25.0 google-cloud-bigquery==0.27.0 influxdb ruamel.yaml==0.16
 
 RUN curl -fsSL https://downloads.python.org/pypy/pypy3.6-v7.3.1-linux64.tar.bz2 | tar xj -C opt  && \
-    ln -s /opt/pypy*/bin/pypy /usr/bin
+    ln -s /opt/pypy*/bin/pypy3 /usr/bin
 
 RUN curl -o installer https://sdk.cloud.google.com && \
     bash installer --disable-prompts --install-dir=/ && \
--- a/kettle/make_json.py
+++ b/kettle/make_json.py
@@ -149,7 +149,13 @@ class Build:
 def parse_junit(xml):
     """Generate failed tests as a series of dicts. Ignore skipped tests."""
     # NOTE: this is modified from gubernator/view_build.py
-    tree = ET.fromstring(xml)
+
+    try:
+        tree = ET.fromstring(xml)
+    except ET.ParseError:
+        print("Malformed xml, skipping")
+        return [] #return empty itterator to skip results for this test
+
 
     # pylint: disable=redefined-outer-name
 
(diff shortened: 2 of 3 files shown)


kettle/make_json.py
```diff
@@ -149,7 +149,13 @@ def set_elapsed(self):
 def parse_junit(xml):
     """Generate failed tests as a series of dicts. Ignore skipped tests."""
     # NOTE: this is modified from gubernator/view_build.py
-    tree = ET.fromstring(xml)
+
+    try:
+        tree = ET.fromstring(xml)
+    except ET.ParseError:
+        print("Malformed xml, skipping")
```
spiffxp, line 156: For followup: do we get logging of which tests / jobs had malformed XML?

kettle/Dockerfile
```diff
@@ -33,7 +33,7 @@ RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime \
 RUN pip3 install requests google-cloud-pubsub==0.25.0 google-cloud-bigquery==0.27.0 influxdb ruamel.yaml==0.16
 
 RUN curl -fsSL https://downloads.python.org/pypy/pypy3.6-v7.3.1-linux64.tar.bz2 | tar xj -C opt  && \
-    ln -s /opt/pypy*/bin/pypy /usr/bin
+    ln -s /opt/pypy*/bin/pypy3 /usr/bin
```
spiffxp, line 36: I couldn't find anything in release notes for 7.1 - 7.3 that would explain this, super weird

spiffxp (approved):
/approve
/lgtm

kettle/make_json.py
MushuEE: right before hand we log the bucket but within this method we don't have info if it is malformed. 

I will follow up with a PR to make the logging better 👍

kettle/Dockerfile
MushuEE: neither could I, and I have tried it in a few env and still only get `pypy3`

spiffxp (approved):
/lgtm

Issue #19368, opened 2020-09-25
Kettle BQ load failure


MushuEE:
**What happened**:
```
Killed
BigQuery error in load operation: Could not connect with BigQuery server due to:
RedirectMissingLocation('Redirected but the response is missing a Location:
header.',)
+ time python3 make_db.py --buckets buckets.yaml --junit --threads 32
+ pypy3 make_json.py --days 1 --assert-oldest 1.9
+ pypy3 make_json.py --reset-emitted --days 1 | pv | gzip > build_day.json.gz
+ bq load --source_format=NEWLINE_DELIMITED_JSON --max_bad_records=1000 --replace k8s-gubernator:build.day build_day.json.gz schema.json
+ pypy3 make_json.py --reset-emitted --days 7 | pv | gzip > build_week.json.gz
+ bq load --source_format=NEWLINE_DELIMITED_JSON --max_bad_records=1000 --replace k8s-gubernator:build.week build_week.json.gz schema.json
Traceback (most recent call last):
  File "/kettle/update.py", line 71, in <module>
    main()
  File "/kettle/update.py", line 59, in main
    call(bq_cmd + bq_ext + ' k8s-gubernator:build.week build_week.json.gz schema.json')
  File "/kettle/update.py", line 37, in call
    raise OSError('invocation failed')
OSError: invocation failed
```
Hit on invocation of `pypy3 make_json.py  --days 7 | pv | gzip > build_week.json.gz`

I do see `/bigquery?project=k8s-gubernator&redirect_from_classic=true`
So possibly there is a real issue. This does not happen for Day table (I believe)
**What you expected to happen**:
No issue loading data to build.week
**How to reproduce it (as minimally and precisely as possible)**:
run above from container
**Please provide links to example occurrences, if any**:

**Anything else we need to know?**:

/assign
/area kettle

MushuEE:
https://stackoverflow.com/questions/59815620/gcloud-upload-httplib2-redirectmissinglocation-redirected-but-the-response-is-m might be on to the solution. Possibly something changed with gcloud packages recently?

MushuEE:
Fixed in container with
```
curl https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-sdk-294.0.0-linux-x86_64.tar.gz | tar xfz - -C <dir>

./<dir>/google-cloud-sdk/install.sh -q
```

PR #19369, opened 2020-09-26
Pin google-cloud-sdk to 294.0.0

MushuEE:
Versions greater than this have http redirect issue with 308s

This causes the bq load to fail for the redirect urls

#19368
Fix was seen in internal discussions. 
/cc @spiffxp 
@spiffxp I am not sure if I should untar to `.` or `/tmp` etc. Will trouble shoot locally before deploying 
/hold
/area kettle


spiffxp (approved):
/approve
/lgtm
/hold cancel

PR #19375, opened 2020-09-28
Kettle dropping out of path population early

MushuEE:
This error was causing builds to be existed to early if they were not prs, else was missed in the Build class migration. Will avoid lots of log spam as well

/area kettle

### DIFF
--- a/kettle/make_json.py
+++ b/kettle/make_json.py
@@ -36,7 +36,7 @@ SECONDS_PER_DAY = 86400
 
 def buckets_yaml():
     import ruamel.yaml as yaml  # pylint: disable=import-outside-toplevel
-    with open(os.path.dirname(os.path.abspath('__file__'))+'/buckets.yaml') as fp:
+    with open(os.path.dirname(os.path.abspath(__file__))+'/buckets.yaml') as fp:
         return yaml.safe_load(fp)
 
 # pypy compatibility hack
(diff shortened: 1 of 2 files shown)


kettle/make_json.py
```diff
@@ -103,11 +107,11 @@ def populate_path_to_job_and_number(self):
             if self.path.startswith(bucket):
                 prefix = meta['prefix']
                 break
-
+        else:
```
spiffxp, line 110: is this supposed to be part of the `if` in the for loop?

kettle/make_json.py
```diff
@@ -103,11 +107,11 @@ def populate_path_to_job_and_number(self):
...
             if self.path.startswith('gs://kubernetes-jenkins/pr-logs'):
                 prefix = 'pr:'
             else:
-                raise ValueError('unknown build path')
+                raise BuildError(f'unknown build path for {self.path} in known bucket paths')
```
spiffxp, line 114: The new message lgtm.  What's the rationale behind changing the type? Personally I would have kept it the same to reduce blast radius

spiffxp (changes requested):
Two questions

kettle/make_json.py
MushuEE: It actually is not. It is one of those odd `for - else` statements. I never see them and it is probably the reason it got lost in translation.

kettle/make_json.py
MushuEE: I have just found it more maintainable to use custom exceptions because they are usually more descriptive, if called from else where can be specifically caught from a known origin vs base exceptions that can be triggered by multiple lower level things. 

I am ok removing it to reduce the number of changes though.

kettle/make_json.py
spiffxp: sounds like it tripped us both up, maybe add a comment that this is intentional and/or why?

kettle/make_json.py
MushuEE: done

spiffxp (approved):
/approve
/lgtm

kettle/make_json.py
```diff
@@ -99,15 +100,17 @@ def generate(cls, path, tests, started, finished, metadata, repos):
 
     def populate_path_to_job_and_number(self):
         assert not self.path.endswith('/')
+        prefix = ''
```
MushuEE, line 103:
```suggestion
```
This was for a different approach that I chose not to take

amwat (commented):
/lgtm

PR #19377, opened 2020-09-28
Enable debsity test in load test

marseel:


mborsz:
FTR: In https://github.com/kubernetes/perf-tests/pull/1485:
* CL2_POD_STARTUP_LATENCY_THRESHOLD = 1h by default
* CL2_SCHEDULER_THROUGHPUT_THRESHOLD = 0 by default

This way, even if we enable this in 100 node scale this shouldn't fail the test even if it should (it simply starts gathering data).
It can still have some side effects affecting the test, but given results in experimental job, we should be fine.

/lgtm
/approve

wojtek-t:
> This way, even if we enable this in 100 node scale this shouldn't fail the test even if it should (it simply starts gathering data).
It can still have some side effects affecting the test, but given results in experimental job, we should be fine.

+1
we should also expose them in perf-dash as soon as we have some data

PR #19378, opened 2020-09-28
sig-scalability][presubmit k/k] Remove access-tokens

jprzychodzen:
Fix kubernetes/perf-tests#1489

/assign @wojtek-t 

Removing access tokens from submit to speed up test. It is still possible to test access tokens if needed with optional `pull-kubernetes-e2e-gce-big-performance` presubmit:

```
/test pull-kubernetes-e2e-gce-big-performance
```

### DIFF
--- a/config/jobs/kubernetes/sig-scalability/sig-scalability-presubmit-jobs.yaml
+++ b/config/jobs/kubernetes/sig-scalability/sig-scalability-presubmit-jobs.yaml
@@ -1,5 +1,7 @@
 presubmits:
   kubernetes/kubernetes:
+  # This job is lighweight version of release blocking periodic job 'ci-kubernetes-e2e-gci-gce-scalability'.
+  # Some elements (like access tokens) are removed to speed up testing.
   - name: pull-kubernetes-e2e-gce-100-performance
     cluster: k8s-infra-prow-build
     always_run: true
@@ -48,7 +50,6 @@ presubmits:
         - --test-cmd-args=--report-dir=/workspace/_artifacts
         - --test-cmd-args=--testconfig=testing/density/config.yaml
         - --test-cmd-args=--testconfig=testing/load/config.yaml
-        - --test-cmd-args=--testconfig=testing/access-tokens/config.yaml
         - --test-cmd-args=--testoverrides=./testing/experiments/enable_prometheus_api_responsiveness.yaml
         - --test-cmd-args=--testoverrides=./testing/experiments/enable_restart_count_check.yaml
         - --test-cmd-args=--testoverrides=./testing/experiments/use_simple_latency_query.yaml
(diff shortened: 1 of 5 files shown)


config/jobs/kubernetes/sig-scalability/sig-scalability-presubmit-jobs.yaml
```diff
@@ -48,7 +48,9 @@ presubmits:
         - --test-cmd-args=--report-dir=/workspace/_artifacts
         - --test-cmd-args=--testconfig=testing/density/config.yaml
         - --test-cmd-args=--testconfig=testing/load/config.yaml
-        - --test-cmd-args=--testconfig=testing/access-tokens/config.yaml
+        # 'access-tokens' test is explicitly missing from this presubmit, as it is a very targeted test.
```
wojtek-t, line 51: I'm not sure this comment really gives us much. I would remove it.

jprzychodzen: We should mark here with this test is different than CI test - otherwise we would need to transfer this knowledge in person in case a new engineer joins oncall. Could you suggest a rephrase of this comment?

wojtek-t: I actually claim we will be diverging more and more in the future.
I would like to run just load as part of presubmit and more tests in postsubmit (there will be more and more tests coming).
Mentioning all of them everywhere doesn't make much sense.

jprzychodzen: I do not want to lose a relationship between tests and explanation why there are differences. Some comment here to document can really help.

I'm okay with changing this comment to a more verbose version of 'presubmit test is a lighweight version of periodic test to save time'.

wojtek-t: Let's comment at the job level, not here then.

jprzychodzen: Done.

config/jobs/kubernetes/sig-scalability/sig-scalability-presubmit-jobs.yaml
```diff
@@ -1,5 +1,7 @@
 presubmits:
   kubernetes/kubernetes:
+  # This job is lighweight version of release blocking periodic job 'ci-kubernetes-e2e-gci-gce-scalability'.
+  # Some elements (like access tokens) are removed to speed up testing.
```
wojtek-t, line 4: Thanks!

Let's do the same for release branches, e.g.:
https://github.com/kubernetes/test-infra/blob/9d8ab99aabf76f00432689bcae10836add0f6726/config/jobs/kubernetes/sig-release/release-branch-jobs/1.16.yaml

jprzychodzen: Done. I agree that there we do not really need comments for eng onboarding.

wojtek-t:
/lgtm
/approve

Thanks!

Issue #19360, opened 2020-09-25
Jobs timing out must fail


aojea:
**What happened**:

A job that timed, but it was still trying to run the tests. At that moment, none of the test had failed,  ginkgo output indicates PASS and the job is GREEN

**What you expected to happen**:

The job can not be GREEN :smile: 

**How to reproduce it (as minimally and precisely as possible)**:

See link below

**Please provide links to example occurrences, if any**:

https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-e2e-ubuntu-gce-basic-sctp/1309438967824781312

**Anything else we need to know?**:

I think that is the ginkgo behavior to not fail if you interrupt the test and no test has failed, I don't know who parses the output of the job (prow?, testgrid?) but it should not indicate that the job run is OK

BenTheElder:
https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-e2e-ubuntu-gce-basic-sctp/1309438967824781312 is marked failed?

aojea:
ahh, then it is testgrid not reporting correctly?
https://testgrid.k8s.io/sig-network-gce#ubuntu-gce-basic-sctp

BenTheElder:
testgrid only reads the junit produced by the job, currently prow does not
produce any at the job level, only ARTIFACTS created by the job command

BenTheElder:
Correction: testgrid DOES read the metadata, and produces a synthetic "Overall" row, but this dashboard filters out the "Overall" status with the regex.

this is WAI.

aojea:
duh, thanks for finding it

BenTheElder:
I didn't 🙃, reached out to some of the prow/testgrid team and they
explained.

PR #19390, opened 2020-09-29
remove sig-network testgrid base filter

aojea:
TIL that if you use a regex filter the overall status in testgrid is the one from obtained from the filter, not from the job.

https://github.com/kubernetes/test-infra/issues/19360

We should check that the jobs are healthy, noty that some test pass only, we can filter not working test if necessary.


MrHohn (approved):
Ha, this is tricky.
/lgtm

aojea:
/hold cancel
:crossed_fingers:

PR #19392, opened 2020-09-29
Enable density test in load test on 5k cluster

marseel:


mborsz:
/ok-to-test
/lgtm
/approve

FTR: This only enables data gathering in 5k scale (in 100 scale it looks good). Any slo violation in a new "density test" will not cause test failure.

The current timeout is 1050m, the current test (load + density) takes ~850m, so we have ~200m left. The current load test takes ~120m and we know that steps we are adding to load test should be faster than the density test (this is the point of this change), so we should still have > 80m left.

Issue #19386, opened 2020-09-28
Extract test step periodically failing


eddiezane:
**What happened**:

The `Extract` test step is periodically failing across pipelines since 2020-09-11.

The extract step downloads binaries from https://storage.googleapis.com/kubernetes-release-dev/ci/fast/v1.xxxxxx using curl with a retry of 3. The download makes progress and then fails 3 times with: 

```
curl: (56) OpenSSL SSL_read: SSL_ERROR_SYSCALL, errno 104
```

![image](https://user-images.githubusercontent.com/959573/94475307-df6e3600-019c-11eb-839a-99b315ee597d.png)

**What you expected to happen**:

I expect the download to succeed.

**How to reproduce it (as minimally and precisely as possible)**:

Still flaking as of opening this.

**Please provide links to example occurrences, if any**:

[Timeline of aggregated failures](https://storage.googleapis.com/k8s-gubernator/triage/index.html?date=2020-09-23&text=error%20during%20.%2Fget-kube.sh%3A%20exit%20status%2056).

The following boards run this test (still more to add - will update):

* [Conformance - GCE - master](https://testgrid.k8s.io/sig-release-master-blocking#Conformance%20-%20GCE%20-%20master&include-filter-by-regex=Extract&width=90)
* [gce-cos-master-alpha-features](https://testgrid.k8s.io/sig-release-master-blocking#gce-cos-master-alpha-features&include-filter-by-regex=Extract&width=90)
* [gce-cos-master-default](https://testgrid.k8s.io/sig-release-master-blocking#gce-cos-master-default&include-filter-by-regex=Extract&width=90)
* [gce-cos-master-reboot](https://testgrid.k8s.io/sig-release-master-blocking#gce-cos-master-reboot&include-filter-by-regex=Extract&width=90)
* [gce-cos-master-scalability-100](https://testgrid.k8s.io/sig-release-master-blocking#gce-cos-master-scalability-100&include-filter-by-regex=Extract&width=90)
* [gce-ubuntu-master-containerd](https://testgrid.k8s.io/sig-release-master-blocking#gce-ubuntu-master-containerd&include-filter-by-regex=Extract&width=90)
* [gci-gce-ingress](https://testgrid.k8s.io/sig-release-master-blocking#gci-gce-ingress&include-filter-by-regex=Extract&width=90)
[...]

PR #19394, opened 2020-09-29
Retry get-kube.sh up to 5 times during extraction

hasheddan:
Jobs are currently flaking on ECONNRESET errors when extracting k8s
which are not retried by curl. As opposed to adding the
--retry-all-errors flag, which can cause unintended behavior, to
curl in get-kube.sh, we just bump the retry of the whole script in an
attempt to reduce flakiness.

Attempt at remedying #19386 

/cc @eddiezane @thejoycekung @RobertKielty


eddiezane (approved):
This seems like a good temporary mitigation while we figure out what changed around 2020-09-11 or what's interrupting the download.

/lgtm

amwat (commented):
/approve

PR #19395, opened 2020-09-29
[scalability][presubmit] Add optional Python specific test

jprzychodzen:
/cc @wojtek-t 

Ref. kubernetes/perf-tests#1470

Marking this as an optional test to verify

### DIFF
--- a/config/jobs/kubernetes/sig-scalability/sig-scalability-presubmit-jobs.yaml
+++ b/config/jobs/kubernetes/sig-scalability/sig-scalability-presubmit-jobs.yaml
@@ -603,6 +603,33 @@ presubmits:
             cpu: 1
             memory: "2Gi"
 
+  - name: pull-perf-tests-verify-all-python
+    decorate: true
+    always_run: true
+    # TODO(jprzychodzen): Make this non-optional after verification that it works
+    # See discussion in kubernetes/perf-tests#1470 for context
+    optional: true
+    skip_branches:
+      - release-1.1[2-9] # Presubmit has been implemented for 1.20+ releases
+    path_alias: "k8s.io/perf-tests"
+    annotations:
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes/sig-scalability/sig-scalability-presubmit-jobs.yaml
```diff
@@ -603,6 +603,31 @@ presubmits:
             cpu: 1
             memory: "2Gi"
 
+  - name: pull-perf-tests-verify-all-python
+    decorate: true
+    always_run: true
+    optional: true
```
wojtek-t, line 609: Please add a TODO to remove this once we prove it's stable.

config/jobs/kubernetes/sig-scalability/sig-scalability-presubmit-jobs.yaml
```diff
@@ -603,6 +603,31 @@ presubmits:
...
+    skip_branches:
+      - release-1.1[2-8] # Presubmit has been implemented for 1.19+ releases
```
wojtek-t, line 611: We don't have this script for 1.19 too.

config/jobs/kubernetes/sig-scalability/sig-scalability-presubmit-jobs.yaml
jprzychodzen: Done.

config/jobs/kubernetes/sig-scalability/sig-scalability-presubmit-jobs.yaml
jprzychodzen: Good catch! Thanks

PR #19408, opened 2020-09-30
Add Gardener test results for kubernetes v1.19

dguendisch:
Adds gardener test results for new kubernetes version 1.19 to the testgrid.

cc @schrodit


dguendisch:
@chases2 can you please also `/lgtm` that PR? Seems like otherwise the tide check will not run and hence prow will not merge.

chases2 (approved):

Issue #19414, opened 2020-09-30
Kettle google-cloud-sdk should not need force pinning.


MushuEE:
**What should be cleaned up or changed**:
#19369 was added to deal with #19365 
```
BigQuery error in load operation: Could not connect with BigQuery server due to:
RedirectMissingLocation('Redirected but the response is missing a Location:
header.',)
```

The pinning of the version should be a short term fix.

**Provide any links for context**:

MushuEE:
Revert #19415 on close

BenTheElder:
🛸 scanning for signs of life 🛸

MushuEE:
This will take a little digging, but I will circle back. I found this issue even internal. 
/remove-lifecycle stale

BenTheElder:
https://github.com/kubernetes/test-infra/pull/21120 actually fixed here I think?

MushuEE:
@BenTheElder Correct I believe it no longer needs to be back pinned as the issue was fixed.

PR #19415, opened 2020-09-30
Remove auto-update from gcloud run step from Kettle

MushuEE:
This step will auto update `google-cloud-sdk` versions of the sdk > 294.0.0 have an http2lib conflict and causes `bq load` to fail out. This was fixed for pulling in #19369

This should be added back if #19414 is completed


spiffxp (approved):
/approve
/lgtm
yikes, good catch

PR #19419, opened 2020-10-01
disable sctp connectivity tests on gce

aojea:
It hangs on the test that tries to access the NodePort SCTP service, but is able to access the ClusterIP service, so I assume that is GCE or deployment specific, but not a big deal, so better skip it and reenable if we are more interested later


aojea:
/assign @MrHohn 
sorry for the back and forth but I'd like to have the jobs green in sig-network and this does not seem worth the effort by now

MrHohn (approved):
/lgtm
/approve

PR #19428, opened 2020-10-01
Fix a race in planks aborting and make it more robust

alvaroaleman:
Currently, its possible for the logic that aborts older instances of a
given prowjob to race and first set a prowjob to aborted, then to error:

{"component":"prow-controller-manager","controller":"plank","event-GUID":"bba62c80-03f6-11eb-86e5-e53548611fe1","file":"prow/pjutil/abort.go:126","from":"pending","func":"k8s.io/test-infra/prow/pjutil.TerminateOlderJobs","job":"pull-ci-kubevirt-hyperconverged-cluster-operator-release-1.2-hco-e2e-aws","level":"info","msg":"Transitioning states","name":"bdd0d20a-03f6-11eb-8b65-0a580a800dea","org":"kubevirt","pr":852,"repo":"hyperconverged-cluster-operator","severity":"info","time":"2020-10-01T15:01:38Z","to":"aborted","type":"presubmit"}
{"component":"prow-controller-manager","controller":"plank","event-GUID":"bba62c80-03f6-11eb-86e5-e53548611fe1","file":"prow/plank/reconciler.go:455","from":"pending","func":"k8s.io/test-infra/prow/plank.(*reconciler).syncPendingJob","job":"pull-ci-kubevirt-hyperconverged-cluster-operator-release-1.2-hco-e2e-aws","level":"info","msg":"Transitioning states.","name":"bdd0d20a-03f6-11eb-8b65-0a580a800dea","org":"kubevirt","pr":852,"repo":"hyperconverged-cluster-operator","severity":"info","time":"2020-10-01T15:01:38Z","to":"error","type":"presubmit"}

This happens because:
* The abort logic deletes the pod via a callback when processing a newer
  instance of the job
* Then updates the ProwJobs status to aborted and completes it
* The pod deletion causes an event which we might start processing
  before we have the prowjob status update in our cache
* We now see that the pod is gone when the ProwJobs status is still
  pending and mark it as error

This PR changes the logic to instead simply set the prowjobs status to
aborted, without marking it as complete. Plank will then get an event,
delete the pod and mark the job as completed. This also has the added
side-effect that its more robust, if the deletion fails we retry instead
of just logging a warning as before.

### DIFF
--- a/prow/pjutil/BUILD.bazel
+++ b/prow/pjutil/BUILD.bazel
@@ -76,7 +76,6 @@ go_test(
         "@io_k8s_apimachinery//pkg/runtime:go_default_library",
         "@io_k8s_apimachinery//pkg/util/diff:go_default_library",
         "@io_k8s_apimachinery//pkg/util/sets:go_default_library",
-        "@io_k8s_sigs_controller_runtime//pkg/client:go_default_library",
         "@io_k8s_sigs_controller_runtime//pkg/client/fake:go_default_library",
     ],
 )
--- a/prow/pjutil/abort.go
+++ b/prow/pjutil/abort.go
@@ -44,12 +44,6 @@ type prowClient interface {
 	Patch(ctx context.Context, name string, pt ktypes.PatchType, data []byte, o metav1.PatchOptions, subresources ...string) (result *prowapi.ProwJob, err error)
 }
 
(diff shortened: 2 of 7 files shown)


stevekuznetsov (approved):
Since the controller already handles the case where a ProwJob transitions to aborted state, this flow makes a lot more sense than what we were doing previously.

PR #19437, opened 2020-10-02
Tide print debug instead of error message when it failed to merge due to user errors

chaodaiG:
These should be errors surfaced somewhere on users side, instead of logs on operator side. For example, when there are github actions mixed with prow jobs in a repo, tide would spit out way too many errors because it couldn't merge

/cc @cjwagner @alvaroaleman

### DIFF
--- a/prow/tide/tide.go
+++ b/prow/tide/tide.go
@@ -23,6 +23,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"net/http"
 	"sort"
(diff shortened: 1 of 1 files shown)


prow/tide/tide.go
```diff
@@ -1089,13 +1090,12 @@ func (c *Controller) mergePRs(sp subpool, prs []PullRequest) error {
...
-		keepTrying, err := tryMerge(func() error {
+		keepTrying, hasMerged := c.tryMerge(func() error {
 			ghMergeDetails := c.prepareMergeDetails(commitTemplates, pr, mergeMethod)
 			return c.ghc.Merge(sp.org, sp.repo, int(pr.Number), ghMergeDetails)
 		})
-		if err != nil {
-			log.WithError(err).Error("Merge failed.")
-			errs = append(errs, err)
+		if !hasMerged {
+			errs = append(errs, errors.New("Failed merging"))
```
alvaroaleman, line 1098: Why not simply log it on Debug level here? The errors in `tryMerge` are completely out of context and don't include the info that whatever happened happend when trying to merge

chaodaiG: Good idea, updated

alvaroaleman (approved):

PR #19440, opened 2020-10-02
Update kube dependencies to v1.19.3

alvaroaleman:
This PR updates the kube and controller-runtime dependencies to their v1.19 version.

~WIP because there is no controller-runtime release yet and because there is an unresolved issue with an ingress conversion.~

Lets just merge this even without using a released controller-runtime version as that is happening very slow this time around. We can bring in a released version of it once it exists.


prow/plank/reconciler.go
```diff
@@ -751,15 +727,13 @@ func predicates(additionalSelector string, callback func(bool)) (predicate.Predi
...
-	return &handler.EnqueueRequestsFromMapFunc{
-		ToRequests: handler.ToRequestsFunc(func(mo handler.MapObject) []controllerruntime.Request {
-			return []controllerruntime.Request{{NamespacedName: ctrlruntimeclient.ObjectKey{
-				Namespace: prowJobNamespace,
-				Name:      mo.Meta.GetName(),
-			}},
-			}
-		}),
-	}
+	// TODO alvaroaleman: This shouldn't be taking a MapObject but an Object
```
e-blackwelder, line 730: It looks like you already made that change here. (So, is the TODO still applicable?)

e-blackwelder (approved):
/hold

Looks good to me, but waiting on the WIP (and a "nit" about your TODO comment).

prow/plank/reconciler.go
alvaroaleman: good catch, removed it

petr-muller (approved):

PR #19442, opened 2020-10-02
Start tracking SLO for deck, hook, sinker, and the monitoring stack.

cjwagner:
This PR adds SLO tracking for components with existing alerts that identify SLO violations. In particular I used alerts with severity `high` or `critical` (the ones that route to #testing-ops) that indicate an active outage, not an impending outage.

I chose to use a single combined SLO compliance metric for `grafana`, `prometheus`, and `alertmanager` because these are the constituent components of the monitoring stack and we don't really need to track their SLO as granularly (these components are pretty stable and we don't own their source code). Since we do have uptime, availability, and blackbox prober alerts for the monitoring stack I thought we mind as well track SLO for it, but I don't want to crowd the SLO dashboard. I'm happy to change this if people feel otherwise.

I did not configure SLO to be recorded for components that currently only have uptime alerts. It seems more appropriate to wait until after adding more precise alerts for those components. 

### DIFF
--- a/config/prow/cluster/monitoring/mixins/lib/config.libsonnet
+++ b/config/prow/cluster/monitoring/mixins/lib/config.libsonnet
@@ -6,12 +6,30 @@
       'ghproxy.json': 'd72fe8d0400b2912e319b1e95d0ab1b3',
       'slo.json': 'ea313af4b7904c7c983d20d9572235a5',
     },
+    // Component name constants
     components: {
-      tide: 'Tide'
+      // Values should be lowercase for use with prometheus 'job' label.
+      crier: 'crier',
+      deck: 'deck',
+      ghproxy: 'ghproxy',
+      hook: 'hook',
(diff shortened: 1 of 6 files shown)


config/prow/cluster/monitoring/mixins/lib/config.libsonnet
```diff
@@ -6,12 +6,30 @@
...
+      tide: 'tide',
     },
-    local components = self.components,
+    local comps = self.components,
+
+    // SLO compliance tracking config
     slo: {
-      components: [components.tide],
+      components: [
+        comps.deck,
```
chaodaiG, line 27: LGTM overall, a quick question, are these values defined somewhere else?

cjwagner: The component constants are defined directly above (line 10 with a reference for convenience on line 22).

PR #15010, opened 2019-10-29
Fix ineffectual assignments

ibrasho:
Fix ineffectual assignments reported in https://goreportcard.com/report/github.com/kubernetes/test-infra#ineffassign

Fixes #14943

PR #19444, opened 2020-10-04
Various simplifications, bug fixes and removals of deprecated APIs

bvwells:
Various simplifications, bug fixes and removals of deprecated APIs.

### DIFF
--- a/config/tests/jobs/jobs_test.go
+++ b/config/tests/jobs/jobs_test.go
@@ -753,11 +753,6 @@ func checkScenarioArgs(jobName, imageName string, args []string) error {
 	}
 
 	if scenario == "" {
-		entry := jobName
-		if strings.HasPrefix(jobName, "pull-security-kubernetes") {
-			entry = strings.Replace(entry, "pull-security-kubernetes", "pull-kubernetes", -1)
-		}
-
 		if !scenarioArgs {
 			if strings.Contains(imageName, "kubekins-e2e") ||
 				strings.Contains(imageName, "bootstrap") ||
@@ -1115,8 +1110,8 @@ func TestK8sInfraProwBuildJobsMustNotExceedTotalCapacity(t *testing.T) {
 		}
 	}
 	for _, r := range resourceNames {
-		total, _ := totalLimit[r]
-		max, _ := maxLimit[r]
+		total := totalLimit[r]
+		max := maxLimit[r]
 		if total.Cmp(max) > -1 {
 			t.Errorf("Total %s limit %s greater than expected limit %s", r, total.String(), max.String())
 		}
--- a/kubetest/aksengine.go
+++ b/kubetest/aksengine.go
@@ -214,7 +214,11 @@ func (c *aksEngineDeployer) SetCustomCloudProfileEnvironment() error {
 			Timeout: 30 * time.Second,
 		}
 		endpointsresp, err := httpClient.Get(metadataURL)
-		if err != nil || endpointsresp.StatusCode != 200 {
+		if err != nil {
+			return fmt.Errorf("%s . apimodel invalid: failed to retrieve Azure Stack endpoints from %s", err, metadataURL)
+		}
+		defer endpointsresp.Body.Close()
+		if endpointsresp.StatusCode != 200 {
 			return fmt.Errorf("%s . apimodel invalid: failed to retrieve Azure Stack endpoints from %s", err, metadataURL)
 		}
 
@@ -696,6 +700,9 @@ func (c *aksEngineDeployer) loadARMTemplates() error {
 func (c *aksEngineDeployer) getAzureClient(ctx context.Context) error {
 	// instantiate Azure Resource Manager Client
 	env, err := azure.EnvironmentFromName(c.azureEnvironment)
+	if err != nil {
+		return err
+	}
 	var client *AzureClient
 	if c.isAzureStackCloud() && strings.EqualFold(c.azureIdentitySystem, ADFSIdentitySystem) {
 		if client, err = getAzureClient(env,
@@ -758,7 +765,6 @@ func (c *aksEngineDeployer) createCluster() error {
 func (c *aksEngineDeployer) dockerLogin() error {
 	cwd, _ := os.Getwd()
 	log.Printf("CWD %v", cwd)
-	cmd := &exec.Cmd{}
 	username := ""
 	pwd := ""
 	server := ""
@@ -781,7 +787,7 @@ func (c *aksEngineDeployer) dockerLogin() error {
 		pwd = c.credentials.ClientSecret
 		server = imageRegistry
 	}
-	cmd = exec.Command("docker", "login", fmt.Sprintf("--username=%s", username), fmt.Sprintf("--password=%s", pwd), server)
+	cmd := exec.Command("docker", "login", fmt.Sprintf("--username=%s", username), fmt.Sprintf("--password=%s", pwd), server)
 	if err = cmd.Run(); err != nil {
 		return fmt.Errorf("failed Docker login with error: %v", err)
 	}
--- a/prow/flagutil/github.go
+++ b/prow/flagutil/github.go
@@ -18,7 +18,6 @@ package flagutil
 
 import (
 	"bytes"
-	"errors"
 	"flag"
 	"fmt"
 	"net/url"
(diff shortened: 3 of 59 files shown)


bvwells:
Hope this is useful!

prow/spyglass/lenses/lenses.go
```diff
@@ -91,9 +91,6 @@ func RegisterLens(lens Lens) error {
 	if config.Title == "" {
 		return errors.New("empty title field in view metadata")
 	}
-	if config.Priority < 0 {
```
alvaroaleman, line 94: Why did you remove this?

config/tests/jobs/jobs_test.go
```diff
@@ -753,11 +753,6 @@ func checkScenarioArgs(jobName, imageName string, args []string) error {
 	}
 
 	if scenario == "" {
-		entry := jobName
-		if strings.HasPrefix(jobName, "pull-security-kubernetes") {
```
alvaroaleman, line 757: Why did you remove this? Do we not ready the `entry` variable anymore?

prow/kube/client.go
```diff
@@ -214,8 +214,6 @@ func (c *Client) requestRetry(r *request) ([]byte, error) {
 		return nil, NewConflictError(fmt.Errorf("body: %s", string(rb)))
 	} else if resp.StatusCode == 422 {
 		return nil, NewUnprocessableEntityError(fmt.Errorf("body: %s", string(rb)))
-	} else if resp.StatusCode == 404 {
```
alvaroaleman, line 217: Why did you remove this?

prow/repoowners/repoowners.go
```diff
@@ -794,7 +794,7 @@ func (o *RepoOwners) entriesForFile(path string, people map[string]map[*regexp.R
 	if !o.enableMDYAML || !strings.HasSuffix(path, ".md") {
 		// if path is a directory, this will remove the leaf directory, and returns "." for topmost dir
 		d = filepath.Dir(d)
-		d = canonicalize(path)
```
alvaroaleman, line 797: This changes the result, are you sure this is correct?

alvaroaleman (commented):
/ok-to-test
Thanks for your work! Some questions, but most of it looks good

prow/spyglass/lenses/lenses.go
bvwells: Priority is a uint so can't be less than zero.

prow/kube/client.go
bvwells: 404 is already handled just above. Sadly GitHub diff doesn't show it. :-)

prow/repoowners/repoowners.go
bvwells: Not 100%. The value for d from 
```
d = filepath.Dir(d)
```
Is never used as it is overwritten by
```
d = canonicalize(path)
```
And is hence redundant. There is some code in the same file which does 
```
                d = filepath.Dir(d)
		d = canonicalize(d)
```
So I assume this should be the same...

bvwells:
Thanks for the review @alvaroaleman!

config/tests/jobs/jobs_test.go
bvwells: Yes, ```entry``` is no longer used.

prow/repoowners/repoowners.go
alvaroaleman: Can you instead change it to just `d = canonicalize(d)` (without the `filepath.Dir(d)`)? That would preserve the current behavior. Unfortunately we don't have tests for this it seems

bvwells: Sure, no problems. I've updated and removed the comment as it refers to ```filepath.Dir(d)```.

alvaroaleman (approved):
/assign @fejta

kubetest/aksengine.go
```diff
@@ -214,7 +214,11 @@ func (c *aksEngineDeployer) SetCustomCloudProfileEnvironment() error {
 			Timeout: 30 * time.Second,
 		}
 		endpointsresp, err := httpClient.Get(metadataURL)
-		if err != nil || endpointsresp.StatusCode != 200 {
+		if err != nil {
```
spiffxp, line 217: I don't think this matches the intent of what was here before.  Is there a test that demonstrates the behavior this is simplifying?

https://golang.org/pkg/net/http/#Client.Get

> A non-2xx response doesn't cause an error.

prow/flagutil/github.go
```diff
@@ -57,9 +57,9 @@ func (o *GitHubOptions) AddFlags(fs *flag.FlagSet) {
 // Validate validates GitHub options.
 func (o *GitHubOptions) Validate(dryRun bool) error {
 	endpoints := o.endpoint.Strings()
-	for _, uri := range endpoints {
+	for i, uri := range endpoints {
 		if uri == "" {
-			uri = github.DefaultAPIEndpoint
+			endpoints[i] = github.DefaultAPIEndpoint
```
spiffxp, line 62: can you add a test that demonstrates the bug that this is fixing?

prow/github/helpers.go
```diff
@@ -34,7 +34,7 @@ const ImageSizeLimit = 5242880
 // HasLabel checks if label is in the label set "issueLabels".
 func HasLabel(label string, issueLabels []Label) bool {
 	for _, l := range issueLabels {
-		if strings.ToLower(l.Name) == strings.ToLower(label) {
+		if strings.EqualFold(l.Name, label) {
```
spiffxp, line 37: does this more closely match GitHub's equality than what was there before?

prow/github/types.go
```diff
@@ -745,7 +745,7 @@ func (i Issue) IsPullRequest() bool {
 // HasLabel checks if an issue has a given label.
 func (i Issue) HasLabel(labelToFind string) bool {
 	for _, label := range i.Labels {
-		if strings.ToLower(label.Name) == strings.ToLower(labelToFind) {
+		if strings.EqualFold(label.Name, labelToFind) {
```
spiffxp, line 748: does this more closely match GitHub's equality than what was there before?

prow/pluginhelp/hook/hook.go
```diff
@@ -362,13 +362,13 @@ func (ha *HelpAgent) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		http.Error(w, "405 Method not allowed", http.StatusMethodNotAllowed)
 		return
 	}
-	b, err := ioutil.ReadAll(r.Body)
+	_, err := ioutil.ReadAll(r.Body)
```
spiffxp, line 365: Should we consider removing this block entirely? ref: https://github.com/kubernetes/test-infra/pull/15010

spiffxp (changes requested):
A few questions/nits

Out of curiosity, what tool were you using for this?

prow/pluginhelp/hook/hook.go
bvwells: Sure, removed.

prow/github/helpers.go
bvwells: https://golang.org/pkg/strings/#EqualFold
```
EqualFold reports whether s and t, interpreted as UTF-8 strings, are equal under Unicode case-folding, which is a more general form of case-insensitivity.
```

You would perhaps never hit any issues as the labels you are using are in English. Use of strings.EqualFold is often encouraged as a best practice.

prow/github/types.go
bvwells: See https://github.com/kubernetes/test-infra/pull/19444#discussion_r499798993

prow/flagutil/github.go
bvwells: Added a number of tests for the Validate function. Cleaned up Validate slightly as some of the code there is unreachable. Hope it is all okay.

bvwells:
@spiffxp thanks for the review. I was using https://github.com/golangci/golangci-lint which runs many community linters. 

Not sure if this is something you'd be interested in running on CI. I can raise an issue if it is of interest.

prow/flagutil/github.go
bvwells: The bug here is that the range loop over endpoints takes a copy of each slice element. The copy of the slice element ```uri``` is set to the default API endpoint which has absolutely no effect.

I think that the expected behaviour here should be that if any of the string elements in the endpoints slice is empty its value should be defaulted to the default API endpoint.

alvaroaleman:
/lgtm

ping @spiffxp anything open from your side?

btw @bvwells does this fix all erros golangci-lint finds (in which configuration?)? We could consider enabling it, as its pretty good at finding subtile bugs and dead code

bvwells:
@alvaroaleman there are still more issues to be fixed. Some of them would depend on which linters you'd like to enable. I could raise an issue if you like.

fejta (approved):

alvaroaleman:
> @alvaroaleman there are still more issues to be fixed. Some of them would depend on which linters you'd like to enable. I could raise an issue if you like.

Yeah, if you can do that, that would be great. Its a topic that's perfectly suited for newcomes and opening an issue will tell ppl that this is something we want.

PR #19446, opened 2020-10-04
run sctp tests in kind

aojea:
I tested it locally and works.
 We run SCTP tests on a GCE job, and the only issue was intra-VM communication.
Since KIND does not use  intra-VM communication it should be fine


MrHohn (approved):
/lgtm
/approve

PR #19447, opened 2020-10-04
Added configuration to run tests on kind cluster for multi-tenancy project (kubectl-mtb)

Divya063:
Tried to check locally using phaino, but dind was not properly setup as there's no `/sys/fs/cgroup`
Environment: macOS


spiffxp (changes requested):
In order to ensure future PR's to config/jobs/kubernetes-sigs/wg-multitenancy don't have to go all the way down to config/OWNERS for approval, can you please add an OWNERS file?

Divya063:
@spiffxp Done.

PR #19448, opened 2020-10-04
Kettle add option to limit number of builds pulled

MushuEE:
This will be primarily used for staging to make the process faster and to limit the space used on the staging deployment

/area kettle


spiffxp (approved):
/approve
/lgtm

Issue #9079, opened 2018-08-16
prow: deck: serve doc for configuration options for plugins


stevekuznetsov:
Would be awesome to be able to serve doc from the `deck` plugin catalog for the YAML configuration that each plugin accepts.

/area prow
/area prow/deck
/kind documentation
/cc @cjwagner @BenTheElder @fejta

cjwagner:
This would be a nice addition. PluginHelp already includes details about the current configuration so it makes sense to extend that to describe the configuration options more generally and potentially provide example YAML snippets.

We should consider providing a link to the `config.yaml` source for the instance so that it is obvious which file holds the config for the Prow instance.

stevekuznetsov:
> We should consider providing a link to the config.yaml source for the instance so that it is obvious which file holds the config for the Prow instance.

+100

clarketm:
@stevekuznetsov / @cjwagner / @idealhack  - regarding implementation (more specifically UX/UI) is the goal here to simply provide a href out to a static *example* yaml configuration for each plugin (either in Github repo or hosted via deck), or should we take this an additional step and add a snippet of `yaml` to the UI component for each plugin?

![image](https://user-images.githubusercontent.com/9957358/61399104-4a766a00-a882-11e9-96bc-7ed6e28d8fd6.png)

cjwagner:
Are you asking if we should embed the YAML in the UI instead of just providing a link to it? Embedding in the UI would certainly be preferable.

clarketm:
@cjwagner - Yes. At the minimum, I was planning on linking to the hosted file, but I wanted to ensure that additionally adding a formatted `yaml` snippet to the UI was the desired intent as that is how I envision this being the most useful.

spiffxp:
/help
This has been done for the approve plugin, if someone wanted to help they could put in similar snippets for other plugins by following @clarketm's example https://github.com/kubernetes/test-infra/pull/13507/files#diff-fafcffa198880e235687b288b620bcb8

matthyx:
- [ ] blockade
- [ ] blunderbuss
- [ ] bugzilla
- [ ] cat
- [ ] cherrypickunapproved
- [ ] dco
- [ ] golint
- [ ] goose
- [ ] heart
- [ ] label
- [ ] lgtm
- [ ] milestoneapplier
- [ ] override
- [ ] project
- [ ] projectmanager
- [ ] require-matching-label
- [ ] requiresig
- [ ] retitle
- [ ] slack
[...]

PR #19450, opened 2020-10-05
Add configuration snippet to blockade plugin

matthyx:
relates to #9079

### DIFF
--- a/prow/plugins/blockade/blockade.go
+++ b/prow/plugins/blockade/blockade.go
@@ -74,9 +74,30 @@ func helpProvider(config *plugins.Configuration, enabledRepos []config.OrgRepo) 
 		}
 		blockConfig[repo.String()] = buf.String()
 	}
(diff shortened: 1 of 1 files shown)


fejta (approved):

prow/plugins/blockade/blockade.go
```diff
@@ -74,9 +74,30 @@ func helpProvider(config *plugins.Configuration, enabledRepos []config.OrgRepo)
...
+			{
+				Repos: []string{
+					"ORGANIZATION",
+					"ORGANIZATION/REPOSITORY",
+				},
+				BlockRegexps: []string{
+					"^examples/",
+				},
+				ExceptionRegexps: []string{
+					"^excepts/",
```
cjwagner, line 88: > ExceptionRegexps are regular expressions matching the file paths that are exceptions to the BlockRegexps.

Thus ExceptionRegexps only make sense if they match a subset of what the BlockRegexps match. So this example is misleading.

PR #19460, opened 2020-10-06
Add a prometheus metric for missing pods in sinker.

cjwagner:
Adding a metric for the number of 'pod not found' errors sinker sees so that we can add an alert for when sinker cannot find a significant fraction of the pods it is trying to delete.
ref https://github.com/kubernetes/test-infra/pull/19090#discussion_r488080386

I can alternatively reuse the existing `sinker_pod_removal_errors` metric and distinguish the 'not found' errors with the reason label. I was concerned about that being a breaking change for existing consumers of that metric, but if that is not a concern it may be preferable. Let me know.

/assign @alvaroaleman @stevekuznetsov

### DIFF
--- a/prow/cmd/sinker/main.go
+++ b/prow/cmd/sinker/main.go
@@ -462,11 +462,14 @@ func (c *controller) deletePod(log *logrus.Entry, pod *corev1api.Pod, reason str
 	if err := client.Delete(c.ctx, pod); err == nil {
 		log.WithFields(logrus.Fields{"pod": name, "reason": reason}).Info("Deleted old completed pod.")
 		m.podsRemoved[reason]++
-	} else if !k8serrors.IsNotFound(err) {
-		log.WithField("pod", name).WithError(err).Error("Error deleting pod.")
+	} else {
 		m.podRemovalErrors[string(k8serrors.ReasonForError(err))]++
+		if k8serrors.IsNotFound(err) {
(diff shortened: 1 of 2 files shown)


alvaroaleman:
> I can alternatively reuse the existing sinker_pod_removal_errors metric and distinguish the 'not found' errors with the reason label. I was concerned about that being a breaking change for existing consumers of that metric, but if that is not a concern it may be preferable. Let me know.

I lean a bit towards that, at the end of the day we don't really expect many 404s during normal operation, do we? The only case in which these happens is if something/someone deletes a pod in between the pod LIST call and before we get to process it, right?

cjwagner:
> I lean a bit towards that, at the end of the day we don't really expect many 404s during normal operation, do we? The only case in which these happens is if something/someone deletes a pod in between the pod LIST call and before we get to process it, right?

Yes, I would expect more 404s than other error types, but still a small number. I'll make the change.

cjwagner:
@alvaroaleman Updated. Short and sweet now.

alvaroaleman:
@cjwagner You need to update the 404 toleration test to not panic on metric write :grimacing:

cjwagner:
> @cjwagner You need to update the 404 toleration test to not panic on metric write 😬

Whoops. I've now updated the test to check that 404 is *not* tolerated and that existing pods that are deleted are recorded as such.

alvaroaleman (approved):

PR #19463, opened 2020-10-06
add sig-network ipvs jobs

aojea:
There is practically no coverage for IPVS, only one peridoc and presubmit job in GCE.
This will provide additional signal and an easier way to reproduce issues

### DIFF
--- a/config/jobs/kubernetes/sig-network/sig-network-kind.yaml
+++ b/config/jobs/kubernetes/sig-network/sig-network-kind.yaml
@@ -161,6 +161,118 @@ periodics:
     description: Runs network tests using KIND against latest kubernetes master with an IPv6 kubernetes-in-docker cluster
     testgrid-alert-email: antonio.ojea.garcia@gmail.com, kubernetes-sig-network-test-failures@googlegroups.com
     testgrid-num-columns-recent: '3'
+# network test against kubernetes master branch with `kind`
+- interval: 6h
+  name: ci-kubernetes-kind-network-ipvs
+  labels:
+    preset-service-account: "true"
+    preset-bazel-scratch-dir: "true"
+    preset-bazel-remote-cache-enabled: "true"
+    preset-dind-enabled: "true"
+    preset-kind-volume-mounts: "true"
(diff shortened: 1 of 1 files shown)


aojea:
/assign @MrHohn @BenTheElder 
/cc @cmluciano @danwinship 
Dan this will run the SCTP tests, so it can meet the requirement to go with SCTP to HA
/hold
Depends on KIND

danwinship:
> There is practically no coverage for IPVS, only one peridoc and presubmit job in GCE.
> This will provide additional signal and an easier way to reproduce issues

How is this different from the existing periodic and presubmit? Is it actually useful to have both?

aojea:
> How is this different from the existing periodic and presubmit? Is it actually useful to have both?

Indeed is useful to have both, because they are complementary:

- gce jobs runs in a real scenario, a cluster with VMs in a cloud provider, you may have higher latency or sporadic packet drops. You can exercise more the reliability here. It also is more conservative regarding the components, in terms of upgrades to containerd, cniplugins, image version, ...

- KIND represents a more ideal scenario, since all the network is emulated the latency and bw are impressive, and packet drops can happen (I think) only if kernel buffers runs out. You exercise more the performance here Also, KIND receives practically immediately all the fixes, to containerd, runc, cniplugins, ...

Why you should have both and not one or another:
- If a test fails only in one place and not in the other, you can assume is something specific of one scenario (or component version, ie. runc)
- If a test fails constantly in both, well, something is wrong with the test or with the feature
- If a test flakes in both, you can do a diff, but you have an advantage that it should be possible to reproduce it locally
...

Sorry for the long answer, but since is a common question I just wanted to drop it here ;)

cmluciano (commented):
+1 This will be super helpful for quickly referencing different kube-proxy mode testing during triage.

MrHohn (approved):
LGTM per the rationale from https://github.com/kubernetes/test-infra/pull/19463.
/lgtm

aojea:
/hold cancel
KIND PR merged

PR #19467, opened 2020-10-06
Fix checkconfig cluster test to tolerate defaulted cluster field

alvaroaleman:
Currently we fail if a jenkins job has the cluster field set to
anything, however we check a config after it got defaulted and the
defaulting sets the cluster to "default", meaning this check will always
fail for jenkins type jobs.


hongkailiu (commented):
/lgtm

Thanks.

PR #19470, opened 2020-10-07
Add a checkconfig ci job to use as a heartbeat for alerts.

cjwagner:
The intention is to run this job every 10mins and alert if we don't see at least 1 successful completion in the past 20mins. This will notify us if horologium, plank, or clonerefs are not working.

/assign @chases2 @e-blackwelder

PR #19471, opened 2020-10-07
Add an alert to detect missing successful results from heartbeat jobs.

cjwagner:
/hold
Holding as this is a follow up to https://github.com/kubernetes/test-infra/pull/19470 (The first commit is from that PR)

### DIFF
--- a/config/prow/cluster/monitoring/mixins/prometheus/plank_alerts.libsonnet
+++ b/config/prow/cluster/monitoring/mixins/prometheus/plank_alerts.libsonnet
@@ -1,0 +1,35 @@
+{
+  prometheusAlerts+:: {
+    local componentName = $._config.components.plank,
+    groups+: [
+      {
+        name: 'Heartbeat ProwJobs',
+        # To add more heartbeat PJs add entries to `heartbeatJobs` in config.libsonnet
+        # NOTE: These alerts are associated with plank, but may be
+        #       triggered by problems with horologium or the pod utils.
+        rules: [
+          {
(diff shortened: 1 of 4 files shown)


config/prow/cluster/monitoring/mixins/prometheus/plank_alerts.libsonnet
```diff
@@ -0,0 +1,26 @@
...
+  prometheusAlerts+:: {
+    local componentName = $._config.components.plank,
+    groups+: [
+      {
+        name: 'Heartbeat ProwJobs',
+        rules: [
+          {
+            alert: 'No recent successful runs: `%s`' % job.name,
+            expr: |||
+              increase(prowjob_state_transitions{job_name="%s", state="success"}[%s]) < 0.5
```
e-blackwelder, line 11: Please add a comment explaining what this does.

cjwagner: Done.

e-blackwelder (approved):

cjwagner:
The value of `increase(prowjob_state_transitions{job_name="ci-test-infra-prow-checkconfig", state="success"}[20m]) ` is dropping down to 1 occasionally due to the alertInterval being exactly double the job interval. I'm going to decrease the job interval to 9 mins to avoid this without increasing the alert interval. As is we might accidentally fire an alert on a single transient failure.
Screen Shot 2020-10-07 at 1 58 28 PM

e-blackwelder (approved):

PR #19478, opened 2020-10-07
Tide: don't print error on user errors

chaodaiG:
https://github.com/kubernetes/test-infra/pull/19437 had silented this in `mergePR` function, but forgot to delete the error propagation, which lead to controller still prints out error messages


alvaroaleman (approved):

PR #19482, opened 2020-10-07
Crier: don't print error message when failed to report status on commit disappeared due to force push

chaodaiG:


prow/github/report/report.go
```diff
@@ -143,7 +143,10 @@ func Report(ghc GitHubClient, reportTemplate *template.Template, pj prowapi.Prow
 		return nil
 	}
 
-	if err := reportStatus(ghc, pj); err != nil {
+	if err := reportStatus(ghc, pj); err != nil &&
+		!strings.Contains(err.Error(), "\"message\":\"Not Found\"") {
```
alvaroaleman, line 147: That error would also occur when the bot doesn't have access, for example because its a private repo. Swallowing the error makes a lot of sense, but can we maybe print something on `debug` level when this happens?

alvaroaleman: Also, could you add a test for this? Either here or in prow/crier/reporters/github/reporter_test.go . The latter already has a test for a similiar scenario

chaodaiG: Make sense, both. Done

prow/crier/reporters/github/reporter.go
```diff
@@ -150,10 +150,18 @@ func (c *Client) Report(_ *logrus.Entry, pj *v1.ProwJob) ([]*v1.ProwJob, *reconc
 
 	// TODO(krzyzacy): ditch ReportTemplate, and we can drop reference to config.Getter
 	err := report.Report(c.gc, c.config().Plank.ReportTemplateForRepo(pj.Spec.Refs), *pj, c.config().GitHubReporter.JobTypesToReport)
-	if err != nil && strings.Contains(err.Error(), "This SHA and context has reached the maximum number of statuses") {
-		// This is completely unrecoverable, so just swallow the error to make sure we wont retry, even when crier gets restarted.
-		err = nil
+	if err != nil {
+		if strings.Contains(err.Error(), "This SHA and context has reached the maximum number of statuses") {
+			// This is completely unrecoverable, so just swallow the error to make sure we wont retry, even when crier gets restarted.
+			log.WithError(err).Debug("Encountered unrecoverable error")
```
cjwagner, line 156: nit: If I saw this in the logs I'd think there is a critical error with Prow 😈 It would be nice to clarify that this just means we are skipping retries for this PJ.

prow/crier/reporters/github/reporter.go
```diff
@@ -150,10 +150,18 @@ func (c *Client) Report(_ *logrus.Entry, pj *v1.ProwJob) ([]*v1.ProwJob, *reconc
...
+			err = nil
+		} else if strings.Contains(err.Error(), "\"message\":\"Not Found\"") {
+			// "message":"Not Found" error occurs when someone force push, which is not a crier error
+			log.WithError(err).Debug("Cannot find endpoint")
```
cjwagner, line 160: "Cannot find endpoint" doesn't seem like an accurate message. We wouldn't have received a 404 if that was the case, right? Something like "Could not find PR commit, skipping retries" would be more accurate, or just combine this with the previous conditional and use a generic debug message.

prow/github/client.go
```diff
@@ -1869,7 +1869,7 @@ func (c *client) ListReviews(org, repo string, number int) ([]Review, error) {
 
 // CreateStatus creates or updates the status of a commit.
 //
-// See https://developer.github.com/v3/repos/statuses/#create-a-status
+// See https://docs.github.com/en/free-pro-team@latest/rest/reference/repos#statuses
```
cjwagner, line 1872: nit
```suggestion
// See https://docs.github.com/en/free-pro-team@latest/rest/reference/repos#create-a-commit-status
```

cjwagner (commented):
Thanks for fixing this Chao. I have a few nits, but otherwise LGTM.

prow/crier/reporters/github/reporter.go
chaodaiG: done

prow/crier/reporters/github/reporter.go
chaodaiG: done

prow/github/client.go
chaodaiG: done

cjwagner (approved):

Issue #19484, opened 2020-10-07
release-blocking jobs must run in dedicated cluster: ci-kubernetes-build-fast


spiffxp:
**What should be cleaned up or changed**:

This is part of #18549 

To properly monitor the outcome of this, you should be a member of k8s-infra-prow-viewers@kubernetes.io.  PR yourself into https://github.com/kubernetes/k8s.io/blob/master/groups/groups.yaml#L603-L628 if you're not a member.

NOTE: I am not tagging this as "help wanted" because it is blocked on https://github.com/kubernetes/k8s.io/issues/846.  Here is my guess at how we could do this:
- create a duplicate job that pushes to the new bucket writable by k8s-infra-prow-build
- ensure it's building and pushing appropriately
- update a release-blocking job to pull from the new bucket
- if no problems, roll out changes progressively
  - a few more jobs in release-blocking
  - all jobs in release-blocking that use this job's results
  - a job that still runs in the "default" cluster
  - all jobs that use this job's results
- rename jobs / get rid of the job that runs on the "default" cluster

It will be helpful to note the date/time that PR's merge.  This will allow you to compare before/after behavior.

Things to watch for the job
- https://prow.k8s.io/?job=ci-kubernetes-build-fast
  - does the job start failing more often?
  - does the job start going into error state?
- https://testgrid.k8s.io/presubmits-kubernetes-blocking#ci-kubernetes-build-fast&graph-metrics=test-duration-minutes
  - does the job duration look worse than before? spikier than before?
- https://storage.googleapis.com/k8s-gubernator/triage/index.html?pr=1&job=ci-kubernetes-build-fast
  - do more failures show up than before?
- https://prow.k8s.io/job-history/gs/kubernetes-jenkins/pr-logs/directory/ci-kubernetes-build-fast
  - (can be used to answer some of the same questions as above)
[...]

PR #19487, opened 2020-10-08
create ci-kubernetes-build-fast canary job to validate job running in the cluster

cpanato:
create duplicate `ci-kubernetes-build-fast` to run in the `k8s-infra-prow-build` cluster to validate the execution build

ref: https://github.com/kubernetes/test-infra/issues/19484 and https://github.com/kubernetes/kubernetes/issues/95173

/cc @spiffxp Dont know if there is anything else that we need to change, but this is an initial PR, let me know if this is correct.

### DIFF
--- a/config/jobs/kubernetes/sig-release/kubernetes-builds.yaml
+++ b/config/jobs/kubernetes/sig-release/kubernetes-builds.yaml
@@ -103,6 +103,44 @@ periodics:
     testgrid-alert-email: release-team@kubernetes.io
     description: 'Ends up running: make quick-release'
 
+- interval: 5m
+  name: ci-kubernetes-build-fast-canary
+  cluster: k8s-infra-prow-build
+  labels:
+    preset-service-account: "true"
+    preset-dind-enabled: "true"
+  spec:
+    containers:
+    - image: gcr.io/k8s-testimages/bootstrap:v20200924-7fcf543
+      args:
+      - --repo=k8s.io/kubernetes
+      - --repo=k8s.io/release
+      - --root=/go/src
+      - --timeout=40
+      - --scenario=kubernetes_build
+      - --
+      - --allow-dup
+      - --fast
+      - --release=k8s-release-dev/ci-canary
+      # docker-in-docker needs privileged mode
+      securityContext:
+        privileged: true
+      resources:
+        limits:
+          cpu: 6
+          memory: "12Gi"
+        requests:
+          cpu: 6
+          memory: "12Gi"
+  rerun_auth_config:
+    github_team_ids:
+      - 2241179 # release-managers
+  annotations:
+    testgrid-dashboards: sig-release-master-informing, sig-testing-canaries
+    testgrid-tab-name: build-master-fast-canary
+    testgrid-alert-email: release-team@kubernetes.io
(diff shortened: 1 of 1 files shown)


spiffxp (changes requested):
This needs to be modified to push to gs://k8s-release-dev instead of gs://kubernetes-release-dev

I'm not sure off the top of my head which flag needs to be set where for this

config/jobs/kubernetes/sig-release/kubernetes-builds.yaml
```diff
@@ -103,6 +103,43 @@ periodics:
...
+          cpu: 6
+          memory: "12Gi"
+        requests:
+          cpu: 6
+          memory: "12Gi"
+  rerun_auth_config:
+    github_team_ids:
+      - 2241179 # release-managers
+  annotations:
+    testgrid-dashboards: sig-release-master-blocking, google-unit
```
spiffxp, line 138: I would suggest leaving it off of blocking until we've gained confidence it's working as intended
```suggestion
    testgrid-dashboards: sig-release-master-informing, sig-testing-canaries
```

config/jobs/kubernetes/sig-release/kubernetes-builds.yaml
```diff
@@ -103,6 +103,43 @@ periodics:
...
+    testgrid-tab-name: build-master-fast-canary
+    testgrid-alert-email: release-team@kubernetes.io
+    description: 'Ends up running: make quick-release - temporary job to validate it running in a new cluster'
```
spiffxp, line 141:
```suggestion
    description: 'Ends up running: make quick-release. Canary: runs in k8s-infra-prow-build, pushes to gs://k8s-release-dev'
```

spiffxp (changes requested):
Some suggested changes as well

cpanato:
@spiffxp thanks for your review and input! i will check how to change that and apply here.

cpanato:
@spiffxp done, PTAL

config/jobs/kubernetes/sig-release/kubernetes-builds.yaml
```diff
@@ -103,6 +103,44 @@ periodics:
...
+      args:
+      - --repo=k8s.io/kubernetes
+      - --repo=k8s.io/release
+      - --root=/go/src
+      - --timeout=40
+      - --scenario=kubernetes_build
+      - --
+      - --allow-dup
+      - --fast
+      - --release=k8s-release-dev/ci-canary
```
spiffxp, line 124: I'll merge this as-is just to see if the job runs successfully, but for followup this should be different bucket / same path.  I think what you have here is different bucket / different path

spiffxp (approved):
/approve
/lgtm
I'm going to assume you've got the flag correct.  This will need changes before we can start moving forward, but I'll merge this to get early signal.

config/jobs/kubernetes/sig-release/kubernetes-builds.yaml
cpanato: ok, followup PR: https://github.com/kubernetes/test-infra/pull/19518

PR #19490, opened 2020-10-08
fix the modification time format in gcsweb tool

droslean:
`2020-10-01 12:47:48.208 +0000 UTC` with the current format becomes `10 Oct 2020 12:47:48`  /shrug

RFC1123 format will make it `Thu, 01 Oct 2020 12:47:48 UTC`  


petr-muller (approved):
No more custom format, TZ info included. Not only it fixes a problem, but it's better, too.

stevekuznetsov (approved):

PR #19493, opened 2020-10-08
Add build-gh-pages presubmit job for node-feature-discovery

marquiz:
Verify buildability of gh-pages site.

### DIFF
--- a/config/jobs/kubernetes-sigs/node-feature-discovery/node-feature-discovery-presubmits.yaml
+++ b/config/jobs/kubernetes-sigs/node-feature-discovery/node-feature-discovery-presubmits.yaml
@@ -33,7 +33,6 @@ presubmits:
       description: "Build container image of node-feature-discovery"
     labels:
       preset-dind-enabled: "true"
-      preset-kind-volume-mounts: "true"
     spec:
       containers:
       - image: gcr.io/k8s-testimages/kubekins-e2e:v20201014-882eb3f-master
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes-sigs/node-feature-discovery/node-feature-discovery-presubmits.yaml
```diff
@@ -43,3 +43,22 @@ presubmits:
         - runner.sh
         args:
         - scripts/test-infra/build-image.sh
+  - name: pull-node-feature-discovery-build-gh-pages
+    always_run: true
+    decorate: true
+    annotations:
+      testgrid-dashboards: sig-node-node-feature-discovery
+      testgrid-tab-name: build-gh-pages
+      description: "Test build of gh-pages documentation"
```
fejta, line 52: Why does this require kind and/or a privileged security context?

fejta (approved):

config/jobs/kubernetes-sigs/node-feature-discovery/node-feature-discovery-presubmits.yaml
marquiz: Thanks for the review @fejta !

Kind is certainly not needed and was a copy-paste mistake. This is now fixed in the new version of the PR. Also dropped kind from the build-image job. Docs are built in docker, thus (AFAIU) dind and privileged context is needed. Please correct me if I'm wrong or if some other image than kubekins would be more appropriate.

fejta (approved):
/hold cancel

PR #19495, opened 2020-10-08
Use nightly aks-engine builds for windows + hyper-v isolated container periodics

marosset:
Use nightly aks-engine builds for windows + hyper-v isolated container periodic jobs so we don't need to increase test velocity.


chewong (approved):
/lgtm
/approve

PR #19498, opened 2020-10-08
Clear env var so Azure e2e generates a local keypair

mboersma:
To implement [`ClusterLogCollector`](https://github.com/kubernetes-sigs/cluster-api/blob/59faa7b1c63ba301cb780836891e7b10098fce34/test/framework/cluster_proxy.go#L81) for CAPZ using an SSH mechanism, Azure e2e tests need access to the private key as well. By leaving `AZURE_SSH_PUBLIC_KEY_FILE` empty, testing will fall back to [generating a local keypair](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/scripts/ci-e2e.sh#L78-L86) just for that run.

This works well as demonstrated in kubernetes-sigs/cluster-api-provider-azure#976.

Since the key materials aren't accessible outside of the prow CI environment, I don't think this introduces a security risk. But please let me know if you disagree or have any other feedback. cc: @nader-ziada @CecileRobertMichon @devigned

### DIFF
--- a/config/jobs/kubernetes-sigs/cluster-api-provider-azure/cluster-api-provider-azure-presubmits.yaml
+++ b/config/jobs/kubernetes-sigs/cluster-api-provider-azure/cluster-api-provider-azure-presubmits.yaml
@@ -38,7 +38,7 @@ presubmits:
     labels:
       preset-dind-enabled: "true"
       preset-kind-volume-mounts: "true"
-      preset-azure-cred: "true"
+      preset-azure-cred-only: "true"
     spec:
       containers:
       - image: gcr.io/k8s-testimages/kubekins-e2e:v20201005-0b243c0-1.18
--- a/config/prow/config.yaml
+++ b/config/prow/config.yaml
@@ -722,21 +722,13 @@ presets:
   env:
     - name: AZURE_CREDENTIALS
       value: /etc/azure-cred/credentials
-    - name: AZURE_SSH_PUBLIC_KEY_FILE
-      value: /etc/azure-ssh/azure-ssh-pub
   volumes:
     - name: azure-cred
       secret:
         secretName: azure-cred
-    - name: azure-ssh
-      secret:
-        secretName: azure-ssh
   volumeMounts:
     - name: azure-cred
       mountPath: /etc/azure-cred
-      readOnly: true
-    - name: azure-ssh
-      mountPath: /etc/azure-ssh
       readOnly: true
 # storage / caching presets
 - labels:
@@ -801,4 +793,3 @@ presets:
 - env:
   - name: GOPROXY
     value: "https://proxy.golang.org"
-


devigned (approved):
/lgtm

The key is short lived. I feel good about this change.

devigned:
/cancel lgtm

Looks like there's more work to be done.

devigned:
/lgtm cancel

Looks like there is more work to be done.

mboersma:
I thought overriding the env var was the least disruptive way, but apparently that won't work.

I think either we could remove the `preset-azure-cred` label from the e2e task, or remove usage of AZURE_SSH_PUBLIC_KEY_FILE entirely. I don't see how either approach would break things, but let me look deeper.

CecileRobertMichon:
Pretty sure we need `preset-azure-cred` for the service principal creds

mboersma:
I made a new config label that retains everything e2e appears to need, but removes the `AZURE_SSH_PUBLIC_KEY_FILE` env var.

devigned:
@mboersma what do you think about reusing `preset-azure-cred-only`? Just remove the ssh pub key from it. I don't think it's needed there.

mboersma:
> what do you think about reusing `preset-azure-cred-only`? 

Sure, but it appears we need `REGISTRY` as well (and may be expecting `KUBE_VERBOSE=0`). I can add those to `preset-azure-cred-only` instead if that's ok.

devigned:
I think we should be ok with the default registry via https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/2e100d19787016a78130fd5b9440dedbad754c85/scripts/ci-e2e.sh#L53-L64.

devigned (approved):
/lgtm

config/jobs/kubernetes-sigs/cluster-api-provider-azure/cluster-api-provider-azure-presubmits.yaml
```diff
@@ -38,7 +38,7 @@ presubmits:
     labels:
       preset-dind-enabled: "true"
       preset-kind-volume-mounts: "true"
-      preset-azure-cred: "true"
```
CecileRobertMichon, line 41: are there any other jobs (eg. periodic CAPI e2e) that we're expecting to get logs from that should also be updated?

mboersma: Yes, looks like we should do the same for `pull-cluster-api-provider-azure-capi-e2e`--good thinking. Those are the two places that `./scripts/ci-e2e.sh` is called.

mboersma: Derp, nevermind--we are already there: that was why `preset-azure-cred-only` was added in the first place. So I *think* that covers it.

PR #19500, opened 2020-10-08
Add a Jira plugin to prow

alvaroaleman:
This PR adds a simple jira plugin that will create a link in jira issue that are mentioned in PRs or issues and update comments to insert a link to the corresponding jira issue where mentioned. Apart from the included unit tests I have also deployed this in a demo env  to verify that it actually works.


prow/cmd/hook/main.go
```diff
@@ -152,6 +155,21 @@ func main() {
 		bugzillaClient = &bugzilla.Fake{}
 	}
 
+	defer func() {
```
stevekuznetsov: ?

prow/plugins/jira/jira.go
```diff
@@ -0,0 +1,201 @@
...
+		go func(issue string) {
+			defer wg.Done()
+			if err := upsertGithubLinkToIssue(log, issue, jc, e); err != nil {
+				log.WithField("Issue", issue).WithError(err).Error("Failed to ensure Github link on Jira issue")
+			}
+		}(issue)
+	}
+
+	if err := updateComment(e, referencedIssues.UnsortedList(), jc.JiraURL(), ghc); err != nil {
+		logrus.WithError(err).Error("Failed to insert links into body")
```
stevekuznetsov, line 105: `s/logrus/log/`

prow/plugins/jira/jira_test.go
```diff
@@ -0,0 +1,233 @@
...
+			expectedNewLinks: []jira.RemoteLink{{Object: &jira.RemoteLinkObject{
+				URL:   "https://github.com/org/repo/issues/3",
+				Title: "org/repo#3: Some issue",
+				Icon: &jira.RemoteLinkIcon{
+					Url16x16: "https://github.com/favicon.ico",
+					Title:    "GitHub",
+				},
+			},
+			}},
+			expectedCommentUpdates: map[string]string{"/:1": "Some text and also [ABC-123](https://my-jira.com/browse/ABC-123)"},
```
stevekuznetsov, line 117: Why not specify org&repo to get sensible comments here?

prow/plugins/jira/jira.go
```diff
@@ -0,0 +1,201 @@
...
+		wg.Add(1)
+		go func(issue string) {
+			defer wg.Done()
+			if err := upsertGithubLinkToIssue(log, issue, jc, e); err != nil {
+				log.WithField("Issue", issue).WithError(err).Error("Failed to ensure Github link on Jira issue")
+			}
+		}(issue)
+	}
+
+	if err := updateComment(e, referencedIssues.UnsortedList(), jc.JiraURL(), ghc); err != nil {
```
stevekuznetsov, line 104: Isn't this needlessly called on a title change?

prow/plugins/jira/jira.go
```diff
@@ -0,0 +1,201 @@
...
+		logrus.WithError(err).Error("Failed to insert links into body")
+	}
+	wg.Wait()
+
+	return nil
+}
+
+func updateComment(e *github.GenericCommentEvent, validIssues []string, jiraBaseURL string, ghc githubClient) error {
```
stevekuznetsov, line 112: This is new behavior for Prow - editing user's comments - what is wrong with responding with our own comment from the bot that we added a link?

prow/jira/jira.go
```diff
@@ -0,0 +1,232 @@
...
+	}
+}
+
+func WithFields(fields logrus.Fields) Option {
+	return func(o *Options) {
+		o.LogFields = fields
+	}
+}
+
+func NewClient(endpoint string, opts ...Option) (Client, error) {
```
stevekuznetsov, line 60: Can you copy the logger/delegate design from the GitHub, Bugzilla clients so that we can inject event-specific loggers into these clients when we're using them in a `hook` goroutine?

prow/plugins/jira/jira.go
alvaroaleman: I don't think it will even get title changes as it only listens to genericCommentEvents

prow/plugins/jira/jira.go
alvaroaleman: Well, because its a lot more practical. Someone uses a link in their comment, you can just click on it. Thats a lot nicer than having to scroll to find some bot message. The bot message makes a lot of sense for for example approval and tests, because the message for those always belongs at the end, but this isn't really the case here. You can see it btw in action here https://github.com/alvaro-bot/test-prow/pull/2 and https://github.com/alvaro-bot/test-prow/issues/3

prow/jira/jira.go
alvaroaleman: This is pretty tricky, because the only place where we actually use the logger is in the `retryable` http client. This in turn is wrapped with the basic auth transport and we can not replace the transport on an existing `*jiraClient`. So the only way I can think of that would allow us to make this happen is by going through the constructor again. Unfortunately, the constructor can error which makes this hard to use.

prow/plugins/jira/jira_test.go
alvaroaleman: Done

prow/jira/jira.go
stevekuznetsov: Why not log the HTTP calls the way we do for other clients?

PR #19502, opened 2020-10-08
Override audit policy for conformance informing jobs

hh:
An alternative to https://github.com/kubernetes/kubernetes/pull/95388

### DIFF
--- a/config/jobs/kubernetes/sig-cloud-provider/gcp/gce-conformance.yaml
+++ b/config/jobs/kubernetes/sig-cloud-provider/gcp/gce-conformance.yaml
@@ -37,6 +37,179 @@ periodics:
         requests:
           cpu: 1
           memory: 3Gi
+      env:
+      # cncf/apisnoop needs to consume audit logs from the cluster under test in order to
+      # effectively report on conformance test coverage of api operations. The default audit
+      # policy excludes events from being logged due to performance reasons. For this
+      # job we will accept the performance hit and allow events to be logged so that we can
+      # measure coverage of api operations that involve events.
+      #
+      # We do this by setting ENABLE_APISERVER_ADVANCED_AUDIT which tells
+      # cluster/gce/gci/configure-kubeapiserver.sh to override the default audit policy with the
+      # contents of the ADVANCED_AUDIT_POLICY env var.  We've copy-pasted the default
+      # audit policy as output by https://github.com/kubernetes/kubernetes/blob/b1e3a2ac7a80d49ef6153457eaa35dac4fc98444/cluster/gce/gci/configure-helper.sh#L1077-L1235
+      # and then removed the block that says "don't log events"
+      #
+      # NOTE: if the default audit policy changes, this will need to be updated accordingly
+      # TODO(https://github.com/kubernetes/test-infra/issues/19613)
+      # ^^ ideally we could do this via a patch instead
+      - name: ENABLE_APISERVER_ADVANCED_AUDIT
+        value: "true"
+      - name: ADVANCED_AUDIT_POLICY
+        value: |-
+          apiVersion: audit.k8s.io/v1
+          kind: Policy
+          rules:
+            # The following requests were manually identified as high-volume and low-risk,
+            # so drop them.
+            - level: None
+              users: ["system:kube-proxy"]
+              verbs: ["watch"]
+              resources:
+                - group: "" # core
+                  resources: ["endpoints", "services", "services/status"]
+            - level: None
+              # Ingress controller reads 'configmaps/ingress-uid' through the unsecured port.
+              # TODO(#46983): Change this to the ingress controller service account.
+              users: ["system:unsecured"]
+              namespaces: ["kube-system"]
+              verbs: ["get"]
+              resources:
+                - group: "" # core
+                  resources: ["configmaps"]
+            - level: None
+              users: ["kubelet"] # legacy kubelet identity
+              verbs: ["get"]
+              resources:
+                - group: "" # core
+                  resources: ["nodes", "nodes/status"]
+            - level: None
+              userGroups: ["system:nodes"]
+              verbs: ["get"]
+              resources:
+                - group: "" # core
+                  resources: ["nodes", "nodes/status"]
+            - level: None
+              users:
+                - system:kube-controller-manager
+                - system:kube-scheduler
+                - system:serviceaccount:kube-system:endpoint-controller
+              verbs: ["get", "update"]
+              namespaces: ["kube-system"]
+              resources:
+                - group: "" # core
+                  resources: ["endpoints"]
+            - level: None
+              users: ["system:apiserver"]
+              verbs: ["get"]
+              resources:
+                - group: "" # core
+                  resources: ["namespaces", "namespaces/status", "namespaces/finalize"]
+            - level: None
+              users: ["cluster-autoscaler"]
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes/sig-cloud-provider/gcp/gcp-gce.yaml
```diff
@@ -499,12 +505,163 @@ periodics:
...
-  annotations:
-    testgrid-dashboards: sig-release-master-blocking, google-gce, google-gci
-    testgrid-tab-name: gce-cos-master-default
-    testgrid-num-failures-to-alert: '6'
-    testgrid-alert-email: release-team@kubernetes.io
-    description: Uses kubetest to run e2e tests (-Slow|Serial|Disruptive|Flaky|Feature) against a cluster created with cluster/kube-up.sh
+      env:
+      - name: ENABLE_APISERVER_ADVANCED_AUDIT
+        value: "true"
+      - name: ADVANCED_AUDIT_POLICY
```
spiffxp, line 511: I don't think we should change the default audit policy for the default e2e job

config/jobs/kubernetes/sig-cloud-provider/gcp/gce-conformance.yaml
```diff
@@ -37,6 +37,164 @@ periodics:
         requests:
           cpu: 1
           memory: 3Gi
+      env:
+      - name: ENABLE_APISERVER_ADVANCED_AUDIT
+        value: "true"
+      - name: ADVANCED_AUDIT_POLICY
```
spiffxp, line 43: Can you comment why this is here?  In an ideal world, this job would be rewritten to patch the existing default policy .yaml, and explain why the patch.  Pretend you're a future maintainer trying to understand what this is, whether this is safe to remove, or when/why it needs to be changed.

spiffxp (changes requested):

config/jobs/kubernetes/sig-cloud-provider/gcp/gcp-gce.yaml
hh: Currently we need the results of a full test run (for all tests, not just the conformance runs, so we can see tests that we could promote, or emulate to hit with conformance tests.

Is there another job we could modify that has similar perodicity that's not the default job?

config/jobs/kubernetes/sig-cloud-provider/gcp/gce-conformance.yaml
hh: It seems to be the correct place to over-ride based on the underlying scripts.

https://github.com/kubernetes/kubernetes/blob/397466dec4c3f8cbf577cb7378a081fe9dad4c6c/cluster/gce/gci/configure-kubeapiserver.sh#L146-L150

Should we instead update the cmd/args of the job to run a series of commands to first run 'sed/awk' to patch / remove this, before we run the existing job:

https://github.com/kubernetes/kubernetes/blob/b1e3a2ac7a80d49ef6153457eaa35dac4fc98444/cluster/gce/gci/configure-helper.sh#L1178-L1182

hh: The patch, if we found a way to cleanly apply it within the job:

```
    diff default-policy.json new-policy.json 2>&1
```

```
  71,75d70
  <   # Don't log events requests.
  <   - level: None
  <     resources:
  <       - group: "" # core
  <         resources: ["events"]
```

hh: I updated the PR to only update the conformance job, and added a comment as to why we override the audit policy, including that it would be better to patch rather than override if possible.

config/jobs/kubernetes/sig-cloud-provider/gcp/gce-conformance.yaml
```diff
@@ -37,6 +37,170 @@ periodics:
         requests:
           cpu: 1
           memory: 3Gi
+      env:
+      # cluster/gce/gci/configure-kubeapiserver.sh allows us to override
+      # the default audit-policy
+      # We need to remove the events filter to ensure we can audit
+      # conformance tests are actually hitting event endpoint operations
+      # Ideally we could patch the audit policy within this job definiton
+      # instead of overriding the policy via ADVANCED_AUDIT_POLICY
```
spiffxp, line 46: Maybe this is too much context that instead belongs in the commit message? But this would be my suggestion:
```suggestion
      # cncf/apisnoop needs to consume audit logs from the cluster under test in order to
      # effectively report on conformance test coverage of api operations. The default audit
      # policy excludes events from being logged due to performance reasons. For this
      # job we will accept the performance hit and allow events to be logged.
      # 
      # We do this by setting ENABLE_APISERVER_ADVANCED_AUDIT which tells 
      # cluster/gce/gci/configure-kubeapiserver.sh to override the default audit policy with the
      # contents of the ADVANCED_AUDIT_POLICY env var.  We've copy-pasted the default
      # audit policy as output by https://github.com/kubernetes/kubernetes/blob/b1e3a2ac7a80d49ef6153457eaa35dac4fc98444/cluster/gce/gci/configure-helper.sh#L1077-L1235
      # and then removed the block that says "don't log events"
      # 
      # NOTE: if the default audit policy changes, this will need to be updated accordingly
      # TODO(a-link-to-a-github-issue): ideally we could do this via a patch instead
```

config/jobs/kubernetes/sig-cloud-provider/gcp/gcp-gce.yaml
hh: I'm fine leaving it alone for now. It means we might be lacking non-conformance test coverage, but it's not as important as conformance coverage itself.

config/jobs/kubernetes/sig-cloud-provider/gcp/gce-conformance.yaml
```diff
@@ -37,6 +37,178 @@ periodics:
...
+      # measure coverage of api operations that involve events.
+      #
+      # We do this by setting ENABLE_APISERVER_ADVANCED_AUDIT which tells
+      # cluster/gce/gci/configure-kubeapiserver.sh to override the default audit policy with the
+      # contents of the ADVANCED_AUDIT_POLICY env var.  We've copy-pasted the default
+      # audit policy as output by https://github.com/kubernetes/kubernetes/blob/b1e3a2ac7a80d49ef6153457eaa35dac4fc98444/cluster/gce/gci/configure-helper.sh#L1077-L1235
+      # and then removed the block that says "don't log events"
+      #
+      # NOTE: if the default audit policy changes, this will need to be updated accordingly
+      # TODO(link-to-a-github-issue): ideally we could do this via a patch instead
```
spiffxp, line 54: I meant to imply you should make a github issue and link to it here

hh: Sorry I missed that: https://github.com/kubernetes/test-infra/issues/19613 created and added to the comment.

spiffxp (approved):
/approve
/lgtm

spiffxp:
This caused the job to start failing (ref: https://github.com/kubernetes/test-infra/issues/19613#issuecomment-711723310) so I opened a revert https://github.com/kubernetes/test-infra/pull/19618

PR #19506, opened 2020-10-09
Add configuration snippet to project, projectmanager, require-matchin…

matthyx:
…g-label, requriesig and retitle
also normalize log message with `Warnf`
relates to #9079

### DIFF
--- a/prow/plugins/blockade/blockade.go
+++ b/prow/plugins/blockade/blockade.go
@@ -92,7 +92,7 @@ func helpProvider(config *plugins.Configuration, enabledRepos []config.OrgRepo) 
 		},
 	})
 	if err != nil {
-		logrus.WithError(err).Warn("cannot generate comments for blockade plugin")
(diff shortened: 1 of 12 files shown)


prow/plugins/blockade/blockade.go
```diff
@@ -74,25 +74,9 @@ func helpProvider(config *plugins.Configuration, enabledRepos []config.OrgRepo)
...
-				BlockRegexps: []string{
-					"^examples*",
-				},
-				ExceptionRegexps: []string{
-					"^examples-keep/",
-				},
-				Explanation: "examples have moved elsewhere",
-			},
-		},
-	})
```
petr-muller, line 93: Why is this removed? The description does not mention it.

matthyx: Good catch... I did wrong with my rebase.

petr-muller (approved):

cjwagner (approved):

PR #19507, opened 2020-10-09
add optional ipvs presubmit job for dual stack

aojea:
It turns out that the optional presubmit job that we added for dual stack uses kube-proxy in IPVS mode, hence we can not compare the results with the current kind jobs that uses iptables mode.

Adding another optional job, will allow us to compare the results and help to troubleshoot issues

xref: https://github.com/kubernetes/kubernetes/pull/91824#issuecomment-706106538


MrHohn (approved):
SG - needs a presubmit owner to approve though
/lgtm
/approve

spiffxp (approved):
/approve
/lgtm

PR #19508, opened 2020-10-09
sig-network jobs should test conformance too

aojea:
Kubernetes is a distributed system, that means that everything uses the network in one or another way.
Testing sig-network gives the sig an understanding of the test status that he owns, but running them alone we can not understand if it is causing any issue or affecting other tests.
We must run at minimum conformance tests in parallel to be sure we don't break another tests.


MrHohn (commented):
Make sense, thanks for improving the sig-network jobs :)

PR #19509, opened 2020-10-09
Temporary increase memory for kubemark-500 job

jprzychodzen:
Let's check if test if flaky because it's memory throttled.

Ref. #95401

/cc @tosi3k 
/assign @wojtek-t


jprzychodzen:
Ref kubernetes/kubernetes#95401

PR #19510, opened 2020-10-09
CAPZ: update conformance test k8s version to 1.18.9

CecileRobertMichon:
Updated all conformance jobs to use 1.18.9 instead of 1.18.6.

/assign @nader-ziada @mboersma


mboersma (approved):
lgtm

PR #19513, opened 2020-10-09
Sum heartbeat job metrics across plank pods.

cjwagner:
Without this sum we get a time series for each plank pod which causes us to incorrectly fire an alert twice. Once as the new pod starts with no recent successful runs and once for the old pod as recent runs drop to 0.
Screen Shot 2020-10-09 at 12 58 36 PM

/assign @e-blackwelder


cjwagner:
/unassign @e-blackwelder 
^ OOO
/assign @chases2 @chaodaiG

PR #19516, opened 2020-10-10
fix enable-workload-identity.sh based on the latest doc

chizhg:
According to https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity, some parameters in the `gcloud` command that enables workload identity for existing clusters and node pools have been updated. Fix enable-workload-identity.sh based on it.

### DIFF
--- a/workload-identity/README.md
+++ b/workload-identity/README.md
@@ -43,7 +43,7 @@ bind-service-accounts.sh \
   SOMEBODY@PROJECT.iam.gserviceaccount.com
 ```
 
-This script assumes the same person can access both `K8S\_PROJECT` and
+This script assumes the same person can access both `K8S_PROJECT` and
 `PROJECT`. If that is not true then the `PROJECT` owner can just run this
 command directly:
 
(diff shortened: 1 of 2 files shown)


workload-identity/README.md
```diff
@@ -43,7 +43,7 @@ bind-service-accounts.sh \
   SOMEBODY@PROJECT.iam.gserviceaccount.com
 ```
 
-This script assumes the same person can access both `K8S\_PROJECT` and
```
fejta, line 46: This is to escape the _ character (as opposed to starting italics)

fejta (approved):
/hold

workload-identity/README.md
chizhg: Hmm I checked the preview page in https://github.com/kubernetes/test-infra/blob/deef8f2a358b813446f177a4d2f8c0d53fac9fbb/workload-identity/README.md#gcp-service-account and the formatting looks correct without the escape, so it's actually needed I think.

fejta: Yeah, my editor must be doing something weird/wrong.

PR #19517, opened 2020-10-10
increase sig-network jobs timeouts

aojea:
after adding the conformance tests to the jobs, they need more time to execute and times out.
Previous sig-network jobs running in seral show that it takes about 50 minutes, and conformance about 120 minutes.
Bumping another 50 minutes, and using 200 minutes should be enoug


MrHohn (approved):
/lgtm

Issue #15611, opened 2019-12-16
[prow] GitHub Check Runs API support


morvencao:
**What happened**:

We're using **prow** and **travis-ci.com** combination in our repo, to make **tide** manage PR auto-merging, we add the following `branch-protection` configuration to `config.yam` to make sure travis jobs success before auto-merging.
```
tide:
  ...
  context_options:
    # Use branch protection options to define required and optional contexts
    from-branch-protection: true
    # Treat unknown contexts as optional
    skip-unknown-contexts: true

branch-protection:
  orgs:
    XXX:
      repos:
        xxxxx:
          protect: true
          exclude:
            - "^revert-" # don't protect revert branches
            - "^dependabot/" # don't protect branches created by dependabot
            - "^greenkeeper/" # don't protect branches created by greenkeeper
          required_status_checks:
            contexts:
            - "Travis CI - Pull Request"
``` 

However, the PR is always pending for the reason of `Not mergeable. Job Travis CI - Pull Request has not succeed...`:
![image](https://user-images.githubusercontent.com/5468682/70886482-87ec4880-2016-11ea-8425-a0dfc1fe0681.png)

After investigation, the **travis-ci.com** has deprecated [Github status check API](https://developer.github.com/v3/repos/statuses/), instead, all the repos managed by **travis-ci.com** update status report with the new [GitHub Check Runs API](https://developer.github.com/v3/checks/runs/), see: https://blog.travis-ci.com/2018-09-27-deprecating-github-commit-status-api-for-github-apps-managed-repositories

As a result, tide can't get the status of travis jobs from **travis-ci.com**, that should be the reason of pending PR.

**The question is that do we have plugins that support **travis-ci.com** check run API?** 

Looks like Travis recommend to use `travis-ci.com` for new users and projects (both private and public).

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

**Please provide links to example occurrences, if any**:

**Anything else we need to know?**:

/cc @gyliu513 @clyang82

morvencao:
@BenTheElder Any insight for this?

alvaroaleman:
hey @morvencao , thanks for raising the issue. It is a problem we are aware of and I thought we already had an issue for it, but I cant find it :thinking: 

I am not sure what the current plan is, as the checkruns api is in developer preview and may `change without advance notice`. 

[...]

PR #19520, opened 2020-10-10
Tide: Add checkrun support

alvaroaleman:
This PR adds checkrun support to tide by extending its github graphql query to also get checkruns and then translating them back to status contexts to keep the changeset small. This allows tide to respect statuses posted by github apps like for example travis.

Apart from the included unittests I also deployed this to verify it actually works.

Whats not entirely clear to me is the impact this has on token usage, however in Openshift, the available tokens for the merge robot on the v4 api never went below 3k in the last 30 days so I assume this won't be a problem.

Not marking https://github.com/kubernetes/test-infra/issues/15611 as done yet as the PR leaves one todo that I want to fix in a follow-up.

/assign @cjwagner

### DIFF
--- a/prow/tide/status.go
+++ b/prow/tide/status.go
@@ -218,9 +218,10 @@ func requirementDiff(pr *PullRequest, q *config.TideQuery, cc contextChecker) (s
 
 	// fixing label issues takes precedence over status contexts
 	var contexts []string
+	log := logrus.WithFields(pr.logFields())
 	for _, commit := range pr.Commits.Nodes {
 		if commit.Commit.OID == pr.HeadRefOID {
-			for _, ctx := range unsuccessfulContexts(commit.Commit.Status.Contexts, cc, logrus.New().WithFields(pr.logFields())) {
+			for _, ctx := range unsuccessfulContexts(append(commit.Commit.Status.Contexts, checkRunNodesToContexts(log, commit.Commit.StatusCheckRollup.Contexts.Nodes)...), cc, log) {
 				contexts = append(contexts, string(ctx.Context))
 			}
 		}
--- a/prow/tide/status_test.go
+++ b/prow/tide/status_test.go
@@ -53,6 +53,7 @@ func TestExpectedStatus(t *testing.T) {
 		secondQueryAuthor string
 		milestone         string
 		contexts          []Context
+		checkRuns         []CheckRun
 		inPool            bool
 		blocks            []int
 		prowJobs          []runtime.Object
@@ -191,6 +192,136 @@ func TestExpectedStatus(t *testing.T) {
 			desc:  fmt.Sprintf(statusNotInPool, " Job job-name has not succeeded."),
 		},
 		{
+			name:              "single bad checkrun",
+			labels:            neededLabels,
+			checkRuns:         []CheckRun{{Name: githubql.String("job-name"), Status: checkRunStatusCompleted, Conclusion: githubql.String(githubql.StatusStateFailure)}},
+			author:            "batman",
+			firstQueryAuthor:  "batman",
+			secondQueryAuthor: "batman",
+			milestone:         "v1.0",
+			inPool:            false,
+
+			state: github.StatusPending,
+			desc:  fmt.Sprintf(statusNotInPool, " Job job-name has not succeeded."),
+		},
+		{
+			name:              "single good checkrun",
+			labels:            neededLabels,
+			checkRuns:         []CheckRun{{Name: githubql.String("job-name"), Status: checkRunStatusCompleted, Conclusion: githubql.String(githubql.StatusStateSuccess)}},
+			author:            "batman",
+			firstQueryAuthor:  "batman",
+			secondQueryAuthor: "batman",
+			milestone:         "v1.0",
+			inPool:            true,
+
+			state: github.StatusSuccess,
+			desc:  statusInPool,
+		},
+		{
+			name:   "multiple good checkruns",
+			labels: neededLabels,
+			checkRuns: []CheckRun{
+				{Name: githubql.String("job-name"), Status: checkRunStatusCompleted, Conclusion: githubql.String(githubql.StatusStateSuccess)},
+				{Name: githubql.String("another-job"), Status: checkRunStatusCompleted, Conclusion: githubql.String(githubql.StatusStateSuccess)},
+			},
+			author:            "batman",
+			firstQueryAuthor:  "batman",
+			secondQueryAuthor: "batman",
+			milestone:         "v1.0",
+			inPool:            true,
+
+			state: github.StatusSuccess,
+			desc:  statusInPool,
+		},
+		{
+			name:   "mix of good and bad checkruns",
+			labels: neededLabels,
+			checkRuns: []CheckRun{
+				{Name: githubql.String("job-name"), Status: checkRunStatusCompleted, Conclusion: githubql.String(githubql.StatusStateSuccess)},
+				{Name: githubql.String("another-job"), Status: checkRunStatusCompleted, Conclusion: githubql.String(githubql.StatusStateFailure)},
+			},
+			author:            "batman",
+			firstQueryAuthor:  "batman",
+			secondQueryAuthor: "batman",
+			milestone:         "v1.0",
+			inPool:            false,
+
+			state: github.StatusPending,
+			desc:  fmt.Sprintf(statusNotInPool, " Job another-job has not succeeded."),
+		},
+		{
+			name:   "mix of good status contexts and checkruns",
+			labels: neededLabels,
+			checkRuns: []CheckRun{
+				{Name: githubql.String("job-name"), Status: checkRunStatusCompleted, Conclusion: githubql.String(githubql.StatusStateSuccess)},
+			},
+			contexts: []Context{
+				{Context: githubql.String("other-job-name"), State: githubql.StatusStateSuccess},
+			},
+			author:            "batman",
+			firstQueryAuthor:  "batman",
+			secondQueryAuthor: "batman",
+			milestone:         "v1.0",
+			inPool:            true,
+
+			state: github.StatusSuccess,
+			desc:  statusInPool,
+		},
+		{
+			name:   "mix of bad status contexts and checkruns",
+			labels: neededLabels,
+			checkRuns: []CheckRun{
+				{Name: githubql.String("job-name"), Status: checkRunStatusCompleted, Conclusion: githubql.String(githubql.StatusStateFailure)},
+			},
+			contexts: []Context{
+				{Context: githubql.String("other-job-name"), State: githubql.StatusStateFailure},
+			},
+			author:            "batman",
+			firstQueryAuthor:  "batman",
+			secondQueryAuthor: "batman",
+			milestone:         "v1.0",
+			inPool:            false,
+
+			state: github.StatusPending,
+			desc:  fmt.Sprintf(statusNotInPool, " Jobs job-name, other-job-name have not succeeded."),
+		},
+		{
+			name:   "good context, bad checkrun",
+			labels: neededLabels,
+			checkRuns: []CheckRun{
+				{Name: githubql.String("job-name"), Status: checkRunStatusCompleted, Conclusion: githubql.String(githubql.StatusStateFailure)},
+			},
+			contexts: []Context{
+				{Context: githubql.String("other-job-name"), State: githubql.StatusStateSuccess},
+			},
+			author:            "batman",
+			firstQueryAuthor:  "batman",
+			secondQueryAuthor: "batman",
+			milestone:         "v1.0",
+			inPool:            false,
+
+			state: github.StatusPending,
+			desc:  fmt.Sprintf(statusNotInPool, " Job job-name has not succeeded."),
+		},
+		{
+			name:   "bad context, good checkrun",
+			labels: neededLabels,
+			checkRuns: []CheckRun{
+				{Name: githubql.String("job-name"), Status: checkRunStatusCompleted, Conclusion: githubql.String(githubql.StatusStateSuccess)},
+			},
+			contexts: []Context{
+				{Context: githubql.String("other-job-name"), State: githubql.StatusStateFailure},
+			},
+			author:            "batman",
+			firstQueryAuthor:  "batman",
+			secondQueryAuthor: "batman",
+			milestone:         "v1.0",
+			inPool:            false,
+
+			state: github.StatusPending,
+			desc:  fmt.Sprintf(statusNotInPool, " Job other-job-name has not succeeded."),
+		},
+		{
 			name:              "multiple bad contexts",
 			labels:            neededLabels,
 			author:            "batman",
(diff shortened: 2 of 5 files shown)


matthyx:
> @alvaroaleman: The following test **failed**, say `/retest` to rerun all failed tests:
> 
> Test name	Commit	Details	Rerun command
> pull-test-infra-bazel	[81929d8](https://github.com/kubernetes/test-infra/commit/81929d82428bfd3397750c82a3d2c6b15b094955)	[link](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/test-infra/19520/pull-test-infra-bazel/1315025200789590016)	`/test pull-test-infra-bazel`
> [Full PR test history](https://prow.k8s.io/pr-history?org=kubernetes&repo=test-infra&pr=19520). [Your PR dashboard](https://prow.k8s.io/pr?query=is%3Apr%20state%3Aopen%20author%3Aalvaroaleman). Please help us cut down on flakes by [linking to](https://git.k8s.io/community/contributors/devel/sig-testing/flaky-tests.md#filing-issues-for-flaky-tests) an [open issue](https://github.com/kubernetes/test-infra/issues?q=is:issue+is:open) when you hit one in your PR.

ok, my English isn't great...

chizhg:
This is a feature that we wanted for Knative Prow 😄 , since the Knative community started running some workloads with GitHub Actions, which also post status as GitHub checks.

Another problem that Tide does not work well with other CI systems like Travis and GitHub Actions is, Tide cannot rebase the PRs on top of the most recent base branch commit and retrigger these jobs (like how it handles Prow jobs) before merging them. Do you have insights on how/if Tide can support this for other CI systems?

alvaroaleman:
> Another problem that Tide does not work well with other CI systems like Travis and GitHub Actions is, Tide cannot rebase the PRs on top of the most recent base branch commit and retrigger these jobs (like how it handles Prow jobs) before merging them. Do you have insights on how/if Tide can support this for other CI systems?

This isn't really "tide not working well with other CI systems" this is more "Prow/Tide has a feature none of the other CI systems has ;)

GitHubs support for this problem is extremely limited, neither the status check api nor the check run api have the notion of a base commit and for example travis doesn't merge at all, it just checks out the PR. So the only way to ensure that the code that ends up on the base branch after merging does actually have passing tests is to enforce that PRs are based on the latest HEAD of their base prior to allowing merging, for which github has an option.

Adding a setting to tide to refuse merging PRs if they are not based on the latest head should be reasonable easy.

Automatically rebasing PRs is more complex. PRs do have a setting to allow maintainers to update them, but ppl can uncheck that. If I understand the docs correctly, PRs can be updated to allow this by someone who has write access to the target repo. After that it would be possible to push to the PRs. One big concern I have though is that a CI system updating my PR is something I would never expect to happen :grimacing: Also, I am not sure if I like the idea of having this in tide, as Tide is already pretty complicated and that would complicate it further.

Maybe a different approach to this problem could be to have a plugin (or a GH action) that reacts to pushes in the repo by finding all open PRs that are now not based on the latest head anymore and adds a merge blocking label?

prow/tide/tide.go
```diff
@@ -1731,6 +1750,8 @@ func headContexts(log *logrus.Entry, ghc githubClient, pr *PullRequest) ([]Conte
 	org := string(pr.Repository.Owner.Login)
 	repo := string(pr.Repository.Name)
 	// Log this event so we can tune the number of commits we list to minimize this.
+	// TODO alvaroaleman: Add checkrun support here. Doesn't seem to happen often though,
+	// openshift doesn't have a single occurrence of this in the past seven days.
```
cjwagner, line 1754: ~The failure mode for this is pretty bad if the checkruns are not required for merge by branch protection. We'd merge the PR without passing tests. In particular someone could intentionally bypass tests by pushing a large number of PR commits.~ Looks like we would instead filter the PR out of the pool which isn't ideal, but isn't dangerous.

prow/tide/tide.go
```diff
@@ -1823,3 +1844,42 @@ func nonFailedBatchByNameBaseAndPullsIndexFunc(obj runtime.Object) []string {
 
 	return []string{nonFailedBatchByNameBaseAndPullsIndexKey(pj.Spec.Job, pj.Spec.Refs)}
 }
+
+func checkRunNodesToContexts(log *logrus.Entry, nodes []CheckRunNode) []Context {
+	var result []Context
+	for _, node := range nodes {
+		// GitHub gives us an empty checkrun per status context. In theory they could
```
cjwagner, line 1851: Wow that is unexpected and useless behavior... We should consider filing a support ticket to GitHub about that, it seems unlikely to be intentional.

prow/tide/tide.go
```diff
@@ -1823,3 +1844,42 @@ func nonFailedBatchByNameBaseAndPullsIndexFunc(obj runtime.Object) []string {
...
+	return result
+}
+
+// checkRunToContext translates a checkRun to a classic context
+// ref: https://developer.github.com/v3/checks/runs/#parameters
+func checkRunToContext(checkRun CheckRun) Context {
+	context := Context{
+		Context: checkRun.Name,
+	}
+	if checkRun.Status != "COMPLETED" {
```
cjwagner, line 1873: nit: Using constants for `Status` and `Conclusion` values would avoid typos.

cjwagner (commented):
Do we need to make this an opt-in feature? This could be a breaking change for any repo currently using Tide that also has GitHub checks since this change will make those checks required for merge now.
Since Tide didn't behave well with GH checks previously there might not be any cases of this anymore, but I'm not sure.

alvaroaleman:
> Do we need to make this an opt-in feature? This could be a breaking change for any repo currently using Tide that also has GitHub checks since this change will make those checks required for merge now.

So far, Tide just ignores github checks. It is a breaking change, but I am not sure I like the idea of opting in. IMHO its safe to assume that if a repo has a given status check they care about its result. If not, they can still opt-out in their branch protection config. Furthermore, the tide status context will show `check blabla has not succeeded`, so the issue is made very visible. WDYT?

cjwagner:
I also don't like the idea of opting-in. If users can still opt-out via config then that satisfies my concern 👍 
Specifically the `optional-contexts` fields of Tide's `context_options` config should address this problem: https://godoc.org/k8s.io/test-infra/prow/config#TideContextPolicy

prow/tide/status_test.go
```diff
@@ -206,6 +220,22 @@ func TestExpectedStatus(t *testing.T) {
 			state: github.StatusPending,
 			desc:  fmt.Sprintf(statusNotInPool, " Jobs job-name, other-job-name have not succeeded."),
 		},
+		{
+			name:              "multiple bad checkruns",
```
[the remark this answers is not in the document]
alvaroaleman: Done

prow/tide/tide.go
alvaroaleman: done

prow/tide/status.go
```diff
@@ -218,9 +218,10 @@ func requirementDiff(pr *PullRequest, q *config.TideQuery, cc contextChecker) (s
 
 	// fixing label issues takes precedence over status contexts
 	var contexts []string
+	log := logrus.New().WithFields(pr.logFields())
```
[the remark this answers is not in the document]
alvaroaleman: done

chizhg:
>> Automatically rebasing PRs is more complex. PRs do have a setting to allow maintainers to update them, but ppl can uncheck that. If I understand the docs correctly, PRs can be updated to allow this by someone who has write access to the target repo. After that it would be possible to push to the PRs. One big concern I have though is that a CI system updating my PR is something I would never expect to happen 😬 Also, I am not sure if I like the idea of having this in tide, as Tide is already pretty complicated and that would complicate it further.

Got some interesting observations for GitHub Actions - the default behavior for [checkout GitHub Actions](https://github.com/actions/checkout) is checking out the PR merge branch, which is `refs/pull/:prNumber/merge` - https://github.community/t/questions-about-pr-workflows-and-actions-checkout-v2/122347/4. So as long as the GitHub Actions can be retriggered and validated when Tide is trying to merge a PR, the PR will be safe to merge.

Not sure how other CI systems like `travis-ci` works, but I guess it's a pretty common pattern for workflows triggered on pull requests.

alvaroaleman:
>  So as long as the GitHub Actions can be retriggered and validated when Tide is trying to merge a PR, the PR will be safe to merge.

I don't think there is an API to trigger github actions. The only way to do that is to push to the PR.

prow/tide/tide.go
stevekuznetsov: This is a property of a PR and something leaking from the backend - if it happens for a PR once, it will happen every time, we usually notice. We do see this from time to time and detecting it will be really tough - the first time someone has to figure out why a PR is not merging when it's got failing check runs and hitting this API edge case, they will likely spend an enormous amount of time debugging. We should consider doing this in the current PR or a quick follow-up, I do not want to be the clueless `tide` admin who gets the question about this.

stevekuznetsov (approved):

prow/tide/tide.go
alvaroaleman: Yeah, i'll be doing a follow-up soon for this

chizhg:
> > So as long as the GitHub Actions can be retriggered and validated when Tide is trying to merge a PR, the PR will be safe to merge.
> 
> I don't think there is an API to trigger github actions. The only way to do that is to push to the PR.

It looks there are ways to retrigger successful GitHub Actions now :-)
https://developer.github.com/v3/actions/workflow-runs/#re-run-a-workflow
https://github.com/sindresorhus/refined-github/issues/3113#issuecomment-633963172

Haven't personally tried it yet.

alvaroaleman:
> It looks there are ways to retrigger successful GitHub Actions now :-)

Oh interesting, but it dosn't really help because:
* The pr doesn't get merged in the checkrun, so what is tested differs from what will be in the base branch after merge
* We can not find out the base revision a checkrun ran against if it would be merging

I think the only way to ensure no broken code can merge via actions is to enforce that prs must be based on the latest head of the basebranch before they get merged

chizhg:
> > It looks there are ways to retrigger successful GitHub Actions now :-)
> 
> Oh interesting, but it dosn't really help because:
> 
> * The pr doesn't get merged in the checkrun, so what is tested differs from what will be in the base branch after merge
> * We can not find out the base revision a checkrun ran against if it would be merging
> 
> I think the only way to ensure no broken code can merge via actions is to enforce that prs must be based on the latest head of the basebranch before they get merged

Hmm sorry if I misunderstood, but wouldn't my comment in https://github.com/kubernetes/test-infra/pull/19520#issuecomment-716995553 solve this issue?

The default behavior for `checkout GitHub Actions` is checking out the PR merge branch, which is refs/pull/:prNumber/merge - https://github.community/t/questions-about-pr-workflows-and-actions-checkout-v2/122347/4.

alvaroaleman:
@chizhg oh sorry, I misread before. The question how we find out if a checkrun ran against the latest merge branch still remains though or not?

chizhg:
I think we don't need to find out that as long as the GitHub Action is configured correctly, since every time it reruns it will automatically run against the latest merge branch.

alvaroaleman:
> I think we don't need to find out that as long as the GitHub Action is configured correctly, since every time it reruns it will automatically run against the latest merge branch.

True but we need to track this somehow. At the end of the day tide needs to be able to say "This PR is ready for merge" and in order to determine that, it needs to determine "Did this checkrun run against the latest revision of the base branch" which it currently can not do from all I see.

chizhg:
> > I think we don't need to find out that as long as the GitHub Action is configured correctly, since every time it reruns it will automatically run against the latest merge branch.
> 
> True but we need to track this somehow. At the end of the day tide needs to be able to say "This PR is ready for merge" and in order to determine that, it needs to determine "Did this checkrun run against the latest revision of the base branch" which it currently can not do from all I see.

Yeah agreed. One way to handle this is checking the GitHub Actions config files to make sure they are configured to "run against the latest revision of the base branch", but I'm almost certain it's not something we want to add to Tide 😅

chizhg:
And thanks for your insights and patience here @alvaroaleman ! 

The reason that I'm so interested in this is I feel since we are adding checkrun support here, we have a general consensus that users do have desires to use other CI systems side by side with Prow, so making them integrate more seamlessly might be a reasonable goal.

alvaroaleman:
Reverted in https://github.com/kubernetes/test-infra/pull/19520 because this drives query cost up too much. Maybe the query can be improved?

sbueringer:
@alvaroaleman Just fyi. CheckRun does not seem to work on Github Enterprise (yet). Even with the latest version ([2.22.2](https://enterprise.github.com/releases)). I rolled back to a earlier version of beginning of this March. Just when other GHE users are reporting this

alvaroaleman:
> Just fyi. Querying for CheckRun via GraphQL does not seem to work on Github Enterprise (yet).

According to their docs, GHE does have the types that are being used in the new part of the query so I would assume that using them is supposed to work: https://docs.github.com/en/enterprise-server@2.22/graphql/reference/objects#statuscheckrollup 

They also mention github actions in their [release notes](https://enterprise.github.com/releases/2.22.2/notes), for example:
> If the storage account settings failed to validate while configuring GitHub Actions, running ghe-actions-teardown was required before making a new attempt.


Is this maybe a setting in your instance? Could you in doubt open a ticket with their support and ask?

sbueringer:
@alvaroaleman  Thx. Yup, I'll open a ticket via our internal GHE team. I would assume that we don't have Github actions enabled (yet) which probably explains the errors

sbueringer:
> @alvaroaleman Thx. Yup, I'll open a ticket via our internal GHE team. I would assume that we don't have Github actions enabled (yet) which probably explains the errors. Or it's this preview: https://docs.github.com/en/free-pro-team@latest/graphql/overview/schema-previews#checks-preview

I tried around a bit. Looks like it's necessary to send an accept header to be able to access types of the Checks schema preview.

```bash
func (s *SomeClient) RoundTrip(r *http.Request) (*http.Response, error) {
	r.Header.Set("Accept", "application/vnd.github.antiope-preview+json")
	return s.Transport.RoundTrip(r)
}
```

As the Checks API is not in the preview list on latest (only on 2.22.2), I assume it will be GA with the next GHE release in about a week. As our GHE team seems to update quiet often I assume in my internal setup it will work in a few weeks.

alvaroaleman:
The correct docs link is https://docs.github.com/en/enterprise-server@2.22/graphql/overview/schema-previews

And yeah, it looks like setting the header requires to write a small roundtripper for the v4 client as it has no option for that. If you want, you can do a PR for that, to my knowledge there are no disadvantage to setting that header when its not needed anymore and it might help ppl that are not as quick in updating their GHE

sbueringer:
@alvaroaleman 

I'm not sure how big that change would be. 

In my local tests I just added the header to every request send by the GraphQL client. I think it would be better to only do this in case the GrapQL query uses the specific API (not sure about side effects at other places where the GrapQL client is used). 

As far as I could see there is no option in the GraphQL library we're using to set headers for a specific request.

alvaroaleman:
> n my local tests I just added the header to every request send by the GraphQL client. I think it would be better to only do this in case the GrapQL query uses the specific API (not sure about side effects at other places where the GrapQL client is used).

Yeah, that's what I head in mind. Our V3 client has accept headers that aren't needed anymore for years and to my knowledge this never caused any issues. You should probably make such a roundtripper append to an existing `accept` header if exists, but other than that I do not see any problem with adding it to all requests. Figuring out if a query uses this specific api OTOH is pretty tricky I think.

sbueringer:
> > n my local tests I just added the header to every request send by the GraphQL client. I think it would be better to only do this in case the GrapQL query uses the specific API (not sure about side effects at other places where the GrapQL client is used).
> 
> Yeah, that's what I head in mind. Our V3 client has accept headers that aren't needed anymore for years and to my knowledge this never caused any issues. You should probably make such a roundtripper append to an existing `accept` header if exists, but other than that I do not see any problem with adding it to all requests. Figuring out if a query uses this specific api OTOH is pretty tricky I think.

Sounds reasonable safe with append. I'll open a pr for it.

Issue #14925, opened 2019-10-23
Improve admissions controller to prevent Prow jobs with bad config from taking down Prow.


clarketm:
**What would you like to be added**:
Improve admissions controller to prevent Prow jobs with bad config from taking down Prow. Countless times, misconfigured jobs I have created have caused Prow to crash. 

**Why is this needed**:
Reduce downtime and Increase reliability.

/assign

stevekuznetsov:
While this is valuable I do want to ask how this happens "countless times" -- are people posting YAML they wrote in an editor manually? Use `mkpj` and this does not happen -- may be nice for your deployment to not crash while the (more challenging) work is done to reject those ProwJobs.

stevekuznetsov:
Also, depending on the version of k8s the validation available from the APIServer for CRDs may be better, should look into that before writing an admission controller.

clarketm:
@stevekuznetsov 

> While this is valuable I do want to ask how this happens "countless times" -- are people posting YAML they wrote in an editor manually?

I would say most, if not all, of the *countless times* I am referring to are from editor mishaps (i.e. taking an existing PJ, tweaking it slightly in editor, then applying it). 

> Also, depending on the version of k8s the validation available from the APIServer for CRDs may be better, should look into that before writing an admission controller.

I was *aware* that an admission controller could assist with *validation*, but I have yet to explore other options. I will investigate alternative validation options.

stevekuznetsov:
Yeah whoever is editing PJ YAML by hand should have their hand slapped and told to use `mkpj` ;)

clarketm:
/remove-lifecycle stale

I have not prioritized this but will get to it when I have time.

clarketm:
/assign @cjwagner 

He is working on a formal design document for this.

jprzychodzen:
Adding information from #23283 

**What happened**:
I've applied on the cluster manually edited job. Plank was unable to unmarshal the object, and entered into the restart loop. It seems that CRD definition [1] is much more accepting than Go object[2].

[1] https://github.com/kubernetes/test-infra/blob/master/config/prow/cluster/starter-gcs.yaml#L153
[2] https://github.com/kubernetes/test-infra/blob/master/prow/apis/prowjobs/v1/types.go#L115

**What you expected to happen**:
Malformed prowjobs.prow.k8s.io should be validated on kube-apiserver level by appropriate CRD. If not possible, then malformed objects should be logged then ignored, instead of crashlooping Prow components.

[...]

PR #19522, opened 2020-10-11
Generate a schema for the ProwJob crd

alvaroaleman:
This PR adds codegeneration that uses controller-gen to generate a CRD
schema for ProwJobs based on the go types. This provides us with two
advantages:
    
* It is not possible anymore to submit jobs with an invalid schema.
   Doing this could result in Prow going down completely.
* `kubectl explain` can be used to get info about the fields in a
   prowjob

The only drawback of this change is that we lose the validation that .status.completionTime must be set if `.status.state` is in `success, error, failure`,as controller-gen doesn't seem to support cross-field validation.

Fixes https://github.com/kubernetes/test-infra/issues/14925


alvaroaleman:
/hold
It seems the generated schema is not actually valid:
```
The CustomResourceDefinition "prowjobs.prow.k8s.io" is invalid: 
* metadata.annotations: Too long: must have at most 262144 bytes
* spec.validation.openAPIV3Schema.properties[spec].properties[pipeline_run_spec].properties[params].items.properties[value].type: Unsupported value: "Any": supported values: "array", "boolean", "integer", "number", "object", "string"
* spec.validation.openAPIV3Schema.properties[spec].properties[pipeline_run_spec].properties[pipelineSpec].properties[params].items.properties[default].type: Unsupported value: "Any": supported values: "array", "boolean", "integer", "number", "object", "string"
* spec.validation.openAPIV3Schema.properties[spec].properties[pipeline_run_spec].properties[pipelineSpec].properties[tasks].items.properties[params].items.properties[value].type: Unsupported value: "Any": supported values: "array", "boolean", "integer", "number", "object", "string"
* spec.validation.openAPIV3Schema.properties[spec].properties[pipeline_run_spec].properties[pipelineSpec].properties[tasks].items.properties[conditions].items.properties[params].items.properties[value].type: Unsupported value: "Any": supported values: "array", "boolean", "integer", "number", "object", "string"
* spec.validation.openAPIV3Schema.properties[spec].properties[pipeline_run_spec].properties[pipelineSpec].properties[tasks].items.properties[taskSpec].properties[steps].items.properties[ports].items.properties[protocol].default: Required value: this property is in x-kubernetes-list-map-keys, so it must have a default or be a required property
* spec.validation.openAPIV3Schema.properties[spec].properties[pipeline_run_spec].properties[pipelineSpec].properties[tasks].items.properties[taskSpec].properties[stepTemplate].properties[ports].items.properties[protocol].default: Required value: this property is in x-kubernetes-list-map-keys, so it must have a default or be a required property
* spec.validation.openAPIV3Schema.properties[spec].properties[pipeline_run_spec].properties[pipelineSpec].properties[tasks].items.properties[taskSpec].properties[inputs].properties[params].items.properties[default].type: Unsupported value: "Any": supported values: "array", "boolean", "integer", "number", "object", "string"
* spec.validation.openAPIV3Schema.properties[spec].properties[pipeline_run_spec].properties[pipelineSpec].properties[tasks].items.properties[taskSpec].properties[params].items.properties[default].type: Unsupported value: "Any": supported values: "array", "boolean", "integer", "number", "object", "string"
* spec.validation.openAPIV3Schema.properties[spec].properties[pipeline_run_spec].properties[pipelineSpec].properties[tasks].items.properties[taskSpec].properties[sidecars].items.properties[ports].items.properties[protocol].default: Required value: this property is in x-kubernetes-list-map-keys, so it must have a default or be a required property
* spec.validation.openAPIV3Schema.properties[spec].properties[pod_spec].properties[containers].items.properties[ports].items.properties[protocol].default: Required value: this property is in x-kubernetes-list-map-keys, so it must have a default or be a required property
* spec.validation.openAPIV3Schema.properties[spec].properties[pod_spec].properties[initContainers].items.properties[ports].items.properties[protocol].default: Required value: this property is in x-kubernetes-list-map-keys, so it must have a default or be a required property
* spec.validation.openAPIV3Schema.properties[spec].properties[decoration_config].properties[timeout].type: Unsupported value: "Any": supported values: "array", "boolean", "integer", "number", "object", "string"
* spec.validation.openAPIV3Schema.properties[spec].properties[decoration_config].properties[grace_period].type: Unsupported value: "Any": supported values: "array", "boolean", "integer", "number", "object", "string"
```

alvaroaleman:
/hold cancel
/cc @chaodaiG 

This passed integration testing and is ready for review :tada: It unfortunately means that `kubectl apply -f crd.yaml` doesn't work, because the annotation `kubectl` created for client-side apply exceeds the max annotation size so we must use server side apply. Is the prow.k8s.io infra cluster new enough to allow us using server-side apply?

chaodaiG:
I love this change! And it's awesome to see that it passed integration test!

> The only drawback of this change is that we lose the validation that .status.completionTime must be set if .status.state is in success, error, failure,as controller-gen doesn't seem to support cross-field validation.

Doesn't seem to be supported yet https://book.kubebuilder.io/reference/markers/crd-validation.html, can we raise an issue over there? 
Also, imho there is probably not much to lose if we don't have this validation, iirc there are a couple of places where we check `completionTime` as a signal of job completion, so they need to be changed. If we do want to keep existing behavior, one possible option could be a few `sed` command in `hack/update-codegen.sh`. WDYT?

> Is the prow.k8s.io infra cluster new enough to allow us using server-side apply?

this is a beta feature since [1.18](https://kubernetes.io/blog/2020/04/01/kubernetes-1.18-feature-server-side-apply-beta-2/#:~:text=Server%2Dside%20Apply%20is%20a,to%20edit%20the%20same%20field), prow.k8s.io is on 1.20 atm, so we should be fine

alvaroaleman:
> Doesn't seem to be supported yet https://book.kubebuilder.io/reference/markers/crd-validation.html, can we raise an issue over there?

Created https://github.com/kubernetes-sigs/controller-tools/issues/605

> Also, imho there is probably not much to lose if we don't have this validation, iirc there are a couple of places > where we check completionTime as a signal of job completion, so they need to be changed. If we do want to > keep existing behavior, one possible option could be a few sed command in hack/update-codegen.sh. WDYT?

So generally `completionTime` is the only reliable indicator for completion, so we can not really get rid of that (We used to do it through the state but that doesn't work anymore, because there is a transient `aborted` state (where its in .status not not yet realized) and a final one (when it was realized) and they can only be differentiated through the `completionTime` being set). I can try to add the validation with `sed` as well.

> this is a beta feature since 1.18, prow.k8s.io is on 1.20 atm, so we should be fine

Ok great. Do you know if and if yes where we need to update the deployment postsubmit to use server-side apply?

cjwagner:
> It unfortunately means that `kubectl apply -f crd.yaml` doesn't work, because the annotation `kubectl` created for client-side apply exceeds the max annotation size so we must use server side apply. Is the prow.k8s.io infra cluster new enough to allow us using server-side apply?

If I understand correctly this would be a breaking change to some existing Prow deployment mechanisms (probably all of them). Is that the case? If so I'm hesitant to approve of this change. Is the value add worth breaking every existing Prow instance? How exactly do the deployment mechanisms need to be changed to use server side apply?

chaodaiG:
> I can try to add the validation with `sed` as well.

That would be appreciated!
 
> Ok great. Do you know if and if yes where we need to update the deployment postsubmit to use server-side apply?

I can only provide what I know:

The deployment is defined at https://github.com/kubernetes/test-infra/blob/fcbb70627fd3df525278e8f23c4b5df237a88cf9/config/prow/cluster/BUILD.bazel#L76, which references `component` as defined at https://github.com/kubernetes/test-infra/blob/fcbb70627fd3df525278e8f23c4b5df237a88cf9/prow/def.bzl#L219, and `kwargs` are passed down to `object` at https://github.com/kubernetes/test-infra/blob/fcbb70627fd3df525278e8f23c4b5df237a88cf9/prow/def.bzl#L192, which references `k8s_object` at https://github.com/bazelbuild/rules_k8s/blob/e27fcafdd51cb2d24aeee8c2bc0e2ffe69941946/k8s/object.bzl#L477, args are then unpacked by key `args` at https://github.com/bazelbuild/rules_k8s/blob/e27fcafdd51cb2d24aeee8c2bc0e2ffe69941946/k8s/object.bzl#L507. 

So my guess is that modifying https://github.com/kubernetes/test-infra/blob/fcbb70627fd3df525278e8f23c4b5df237a88cf9/config/prow/cluster/BUILD.bazel#L76 to:
```
component("prowjob", "customresourcedefinition", args="--server-side=true")
```
would probably work

alvaroaleman:
> If I understand correctly this would be a breaking change to some existing Prow deployment mechanisms (probably all of them). Is that the case? If so I'm hesitant to approve of this change. Is the value add worth breaking every existing Prow instance? How exactly do the deployment mechanisms need to be changed to use server side apply?

Yes but to get the breaking change in the first place, they need to update the CRD manifest (except in the case of prow.k8s.io). It also means that if we change the prowjob types in the future, ppl need to update the CRD before code that has those changes will work (the api will just reject it until then).

If its worth it? I honestly don't know. In the associated issue it says "Countless times, misconfigured jobs I have created have caused Prow to crash." and there are infrequent comments by other ppl about the same, @jkaniuk was the last one. I personally never saw this happening, but I am also not in the habit of manually editing prowjobs.

jprzychodzen:
Is it possible to handle this change as optional? Leave two CRDs in repository, one with and the other without validation? This way older K8s cluster will not get broken, and new clusters can opt in for increased reliability.

From my perspective this change would be extremely useful for private Prow instances. As an owner of a private Prow instance I would like to have some kind validation that will prevent cluster users from breaking the instance.

I'm aware that there is a `mkpj` tool, but this tool assumes that underlying configuration is available. However, it's not always a true and when discussing specific jobs, our unit of abstraction is a single ProwJob run. This means that there are happening manual edits on ProwJob level and malformed ProwJob might be applied to cluster.

alvaroaleman:
@cjwagner WDYT?

cjwagner:
> Yes but to get the breaking change in the first place, they need to update the CRD manifest (except in the case of prow.k8s.io).

Gotcha, this addresses most of my concern.
Im fine with this as long as we mention the breaking change in the `announcements.md` file and point to a copy of the existing CRD manifest in the repo. I don't want to add a k8s version requirement just for this feature.

jprzychodzen:
@alvaroaleman Could you address @cjwagner comment?

alvaroaleman:
@chaodaiG @cjwagner ptal, I believe I have addressed all remarks now

config/prow/cluster/BUILD.bazel
```diff
@@ -73,7 +73,11 @@ release(
     component("prow-canary-k8s-io", "managedcertificate"),
     component("prow-k8s-io", "managedcertificate"),
     component("prow-kubernetes-io", "managedcertificate"),
-    component("prowjob", "customresourcedefinition"),
+    component(
+        "prowjob",
+        "customresourcedefinition",
+        args = ["--server-side=true"],
```
chaodaiG, line 79: nit: we're just speculating that this will work, but not sure. might suggest move this change to a separate PR just in case revert is needed we don't need to revert this large PR

chaodaiG (approved):
/hold

LGTM other than my nit

config/prow/cluster/BUILD.bazel
alvaroaleman: Well, not doing this will definitely not work as the PR changes the CRD file :grimacing:

chaodaiG: well, finger crossed then

PR #19525, opened 2020-10-12
Enable density test in load test for presubmits

marseel:
/assign @mborsz

### DIFF
--- a/config/jobs/kubernetes/sig-scalability/sig-scalability-presubmit-jobs.yaml
+++ b/config/jobs/kubernetes/sig-scalability/sig-scalability-presubmit-jobs.yaml
@@ -451,7 +451,8 @@ presubmits:
         - --provider=gce
         - --tear-down-previous
         - --env=CL2_ENABLE_DNS_PROGRAMMING=true
-        - --env=CL2_ENABLE_VIOLATIONS_FOR_SCHEDULING_THROUGHPUT=false
+        - --env=CL2_SCHEDULER_THROUGHPUT_THRESHOLD=0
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes/sig-scalability/sig-scalability-presubmit-jobs.yaml
```diff
@@ -451,15 +451,14 @@ presubmits:
         - --provider=gce
         - --tear-down-previous
         - --env=CL2_ENABLE_DNS_PROGRAMMING=true
-        - --env=CL2_ENABLE_VIOLATIONS_FOR_SCHEDULING_THROUGHPUT=false
```
mborsz, line 454: I think we need to understand why it was disabled here. 

Setting CL2_ENABLE_DENSITY_TEST=true will enable violations here, which may not be allowed for some reason.

marseel: Looks like it was turned off, because on 1.18 and 1.19 SchedulingThroughput was below 100. Right now on master, it's much higher than 100, but it's due to increased master size from 4 cores to 8 cores.

mborsz:
/ok-to-test
/lgtm
/approve

I think we can include other presubmits as well (including k/k presubmit)

mborsz:
https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/perf-tests/1529/pull-perf-tests-clusterloader2/1315962499295612928/ is an example presubmit with this change

PR #19529, opened 2020-10-12
Make buildlimit max if set to 0 (per --help)

MushuEE:
Changed it and it no longer matched the comment
I don't think we ever want to run and collect nothing.
This also make the logic to handle staging vs prod cleaner


amwat (commented):
/lgtm

PR #19531, opened 2020-10-12
Add BUILD_LIMIT env that can be set via kubectl

MushuEE:
This should allow `--env "BUILD_LIMIT=100"` to be set with the staging image

It will limit the number of builds collected from GCS per job
/area Kettle

### DIFF
--- a/kettle/update.py
+++ b/kettle/update.py
@@ -38,7 +38,7 @@ def call(cmd):
 
 
 def main():
-    call('time python3 make_db.py --buckets buckets.yaml --junit --threads 32')
+    call('time python3 make_db.py --buckets buckets.yaml --junit --threads 32 --buildlimit ' + os.getenv('BUILD_LIMIT', '0'))
 
     bq_cmd = 'bq load --source_format=NEWLINE_DELIMITED_JSON --max_bad_records=1000'
     mj_cmd = 'pypy3 make_json.py'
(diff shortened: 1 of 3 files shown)


kettle/update.py
```diff
@@ -38,7 +38,7 @@ def call(cmd):
 
 
 def main():
-    call('time python3 make_db.py --buckets buckets.yaml --junit --threads 32')
+    call('time python3 make_db.py --buckets buckets.yaml --junit --threads 32 --buildlimit ' + os.getenv('BUILD_LIMIT', '0'))
```
amwat, line 41: the defaulting logic should probably be where the flag is declared, instead of where it's used?

MushuEE: default logic is there, but if the ENV is unset I think it will fail because the flag isn't an integer. would just be `TypeError: can only concatenate str (not "NoneType") to str`

amwat: Wouldn't 

```python3
default=int(os.getenv('BUILD_LIMIT', '0'))
``` 
at

https://github.com/kubernetes/test-infra/blob/1579b01bab959eb44dbae5e410ef16546c3d45ff/kettle/make_db.py#L352

do the trick? or am i missing something

not a big deal though

/lgtm
/hold

MushuEE: Ah I see what you mean, attempting to get default from ENV inside made_db.py?

MushuEE: done

amwat (commented):
/lgtm

amwat (commented):
/lgtm

PR #19540, opened 2020-10-13
Enable density in load for all nonkubemark tests

marseel:
This PR removes all usages of density test in non kubemark tests. 

@jkaniuk agreed to migrate calico test too.

/assing @mborsz

### DIFF
--- a/config/jobs/kubernetes/sig-scalability/sig-scalability-experimental-periodic-jobs.yaml
+++ b/config/jobs/kubernetes/sig-scalability/sig-scalability-experimental-periodic-jobs.yaml
@@ -77,7 +77,8 @@ periodics:
       - --gcp-zone=us-east1-b
       - --provider=gce
       - --env=CL2_ENABLE_CLUSTER_OOMS_TRACKER=true
-      - --env=CL2_ENABLE_VIOLATIONS_FOR_SCHEDULING_THROUGHPUT=false
+      - --env=CL2_SCHEDULER_THROUGHPUT_THRESHOLD=0
+      - --env=CL2_ENABLE_DENSITY_TEST=true
       - --test=false
       - --test-cmd=$GOPATH/src/k8s.io/perf-tests/run-e2e.sh
       - --test-cmd-args=cluster-loader2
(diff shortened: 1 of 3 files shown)


config/jobs/kubernetes/sig-scalability/sig-scalability-experimental-periodic-jobs.yaml
```diff
@@ -77,7 +77,8 @@ periodics:
       - --gcp-zone=us-east1-b
       - --provider=gce
       - --env=CL2_ENABLE_CLUSTER_OOMS_TRACKER=true
-      - --env=CL2_ENABLE_VIOLATIONS_FOR_SCHEDULING_THROUGHPUT=false
+      - --env=CL2_SCHEDULER_THROUGHPUT_THRESHOLD=0
```
wojtek-t, line 80: Can we move that to preset if we enable it everywhere?

mborsz: I think it makes sense to set CL2_ENABLE_DENSITY_TEST=true in preset-e2e-scalability-common and CL2_ENABLE_DENSITY_TEST=false in preset-e2e-kubemark-common to (temporarily disable) this in kubemark until it stabilizes.

  For CL2_SCHEDULER_THROUGHPUT_THRESHOLD, I think we should set this in each job separately (only few jobs disable scheduling throughput violations, so using preset isn't possible)

mborsz: For CL2_SCHEDULER_THROUGHPUT_THRESHOLD, I would suggest simply changing CL2_ENABLE_VIOLATIONS_FOR_SCHEDULING_THROUGHPUT=false to CL2_SCHEDULER_THROUGHPUT_THRESHOLD=0 everywhere, without moving to presets. This is a state as is now, and potential cleanup (moving to presets) can happen in the next PR.

mborsz: Sorry, I don't think we should set  CL2_ENABLE_DENSITY_TEST=true in preset-e2e-scalability-common. This will affect all k8s versions, including 1.18, 1.17 and 1.16. While CL2_ENABLE_DENSITY_TEST flag isn't implemented there, so technically should be safe, the current state is way simpler to reason about.

So overall, I suggest keeping this PR in the current state (no presets at all + mechanically changing --testconfig=testing/density/config.yaml to --env=CL2_ENABLE_DENSITY_TEST=true and similarly for CL2_SCHEDULER_THROUGHPUT_THRESHOLD), as this is the easiest to follow. We may want to do flag cleanup at some point, but please do not make it in the current PR.

wojtek-t: OK - the previous releases argument is a good one. Thanks!

mborsz:
/ok-to-test
/lgtm
/approve
/hold
Let's submit this tomorrow morning to monitor affected jobs and rollback if needed.

PR #19543, opened 2020-10-13
[sig-scalability][kubemark-500] Increase timeout and decrease memory req/limit

jprzychodzen:
Rollback changes from #19509 . Decrease flakiness by increasing timeouts by 20 minutes

Ref. kubernetes/kubernetes#95401

/assign @wojtek-t  
/cc @tosi3k

### DIFF
--- a/config/jobs/kubernetes/sig-release/release-branch-jobs/1.16.yaml
+++ b/config/jobs/kubernetes/sig-release/release-branch-jobs/1.16.yaml
@@ -261,7 +261,7 @@ periodics:
       - --repo=k8s.io/kubernetes=release-1.16
       - --repo=k8s.io/perf-tests=release-1.16
       - --root=/go/src
-      - --timeout=140
+      - --timeout=120
       - --scenario=kubernetes_e2e
       - --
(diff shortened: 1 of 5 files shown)


config/jobs/kubernetes/sig-release/release-branch-jobs/1.16.yaml
```diff
@@ -261,7 +261,7 @@ periodics:
       - --repo=k8s.io/kubernetes=release-1.16
       - --repo=k8s.io/perf-tests=release-1.16
       - --root=/go/src
-      - --timeout=140
+      - --timeout=160
```
wojtek-t, line 264: Why do we have such a huge difference between these two timeouts?
We generally tend to have much smaller difference - 20m (for 5k node jobs it's 30m):
https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/sig-scalability/sig-scalability-release-blocking-jobs.yaml#L137

Let's set it 20m above the timeout below (so 120 here, 100 below as is).

Same everywhere else.

PR #19544, opened 2020-10-13
Rerun prow job: log debug instead of error on user errors

chaodaiG:
These errors are already surfaced to users as http errors, no need to bring immediate attention of prow operator


petr-muller:
I'm not that sure about these. If my Prow is somehow misconfigured I think I'd like to know from the logs and not rely on users.

chaodaiG:
Makes sense, dropping the PR
/close

PR #19512, opened 2020-10-09
secrets-store-csi-driver: config for gcp provider tests

tam7t:
Companion to https://github.com/kubernetes-sigs/secrets-store-csi-driver/pull/340

Add integration tests for the gcp plugin to https://github.com/kubernetes-sigs/secrets-store-csi-driver

PR #19548, opened 2020-10-13
secrets-store-csi-driver: remove preset creds

tam7t:
Per comment on https://github.com/kubernetes/test-infra/pull/19512#discussion_r503456849

Remove `preset-service-account: "true"` as this:
1. gives the running job access to modify its own output
2. should not be needed by the providers

The gcp tests are now passing on https://github.com/kubernetes-sigs/secrets-store-csi-driver/pull/340 without this label.


aramase (commented):
/lgtm

aramase (commented):
/approve

PR #19554, opened 2020-10-13
Make missing webhook alert interval configurable.

cjwagner:
Some Prow instances don't have enough activity to expect a webhook every 10mins.

/assign @chaodaiG


config/prow/cluster/monitoring/mixins/prometheus/hook_alert.libsonnet
```diff
@@ -12,13 +12,13 @@
               (sum(increase(prow_webhook_counter[1m])) == 0 or absent(prow_webhook_counter))
               and ((day_of_week() > 0) and (day_of_week() < 6) and (hour() >= 16))
             |||,
-            'for': '10m',
+            'for': $._config.webhookMissingAlertInterval,
```
chaodaiG, line 15: Understand it's not from this PR, but curious why there is single quote around `for` here

cjwagner: Perhaps because `for` is a keyword in jsonnet? It might not be needed, but shouldn't hurt either.

PR #19558, opened 2020-10-13
Split Kettle staging execution out

MushuEE:
Break up kettle execution paths so that staging reports to it's own table

@amwat Do you think it would be worth it to replicate all three tables here? I think 'all' encompasses the logic for `day` and `week` but I am unsure. 

/area kettle

### DIFF
--- a/kettle/update.py
+++ b/kettle/update.py
@@ -55,10 +55,7 @@ def main():
         bq_ext = ' --replace'
         mj_ext = ' --reset-emitted'
 
-    if os.getenv('DEPLOYMENT', 'prod') == "staging":
-        call(f'{mj_cmd} | pv | gzip > build_staging.json.gz')
-        call(f'{bq_cmd} k8s-gubernator:build.staging build_staging.json.gz schema.json')
-    else:
+    if os.getenv('DEPLOYMENT', 'staging') == "prod":
         call(f'{mj_cmd} {mj_ext} --days 1 | pv | gzip > build_day.json.gz')
         call(f'{bq_cmd} {bq_ext} k8s-gubernator:build.day build_day.json.gz schema.json')
 
@@ -70,6 +67,9 @@ def main():
 
         call('python3 stream.py --poll kubernetes-jenkins/gcs-changes/kettle '
             ' --dataset k8s-gubernator:build --tables all:0 day:1 week:7 --stop_at=1')
+    else:
+        call(f'{mj_cmd} | pv | gzip > build_staging.json.gz')
+        call(f'{bq_cmd} k8s-gubernator:build.staging build_staging.json.gz schema.json')
 
 
 if __name__ == '__main__':
--- a/kettle/update.py
+++ b/kettle/update.py
@@ -63,7 +63,7 @@ def main():
         call(bq_cmd + bq_ext + ' k8s-gubernator:build.day build_day.json.gz schema.json')
 
         call(f'{mj_cmd} {mj_ext} --days 7 | pv | gzip > build_week.json.gz')
-        call(bq_cmd + bq_ext + ' k8s-gubernator:build.week build_week.json.gz schema.json')
+        call(f'{bq_cmd} {bq_ext} k8s-gubernator:build.week build_week.json.gz schema.json')
 
         call(f'{mj_cmd} | pv | gzip > build_all.json.gz')
         call(f'{bq_cmd} k8s-gubernator:build.all build_all.json.gz schema.json')
--- a/kettle/update.py
+++ b/kettle/update.py
@@ -41,7 +41,7 @@ def call(cmd):
 
 
 def main():
-    call('time python3 make_db.py --buckets buckets.yaml --junit --threads {THREADS}')
+    call(f'time python3 make_db.py --buckets buckets.yaml --junit --threads {THREADS}')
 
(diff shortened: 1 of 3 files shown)


kettle/update.py
```diff
@@ -38,31 +41,35 @@ def call(cmd):
 
 
 def main():
-    call('time python3 make_db.py --buckets buckets.yaml --junit --threads 32')
+    call('time python3 make_db.py --buckets buckets.yaml --junit --threads {THREADS}')
```
spiffxp, line 44:
```suggestion
    call(f'time python3 make_db.py --buckets buckets.yaml --junit --threads {THREADS}')
```

kettle/update.py
```diff
@@ -38,31 +41,35 @@ def call(cmd):
...
 
-    bq_cmd = 'bq load --source_format=NEWLINE_DELIMITED_JSON --max_bad_records=1000'
+    bq_cmd = 'bq load --source_format=NEWLINE_DELIMITED_JSON --max_bad_records={MAX_BAD_RECORDS}'
```
spiffxp, line 46:
```suggestion
    bq_cmd = f'bq load --source_format=NEWLINE_DELIMITED_JSON --max_bad_records={MAX_BAD_RECORDS}'
```

kettle/update.py
```diff
@@ -38,31 +41,35 @@ def call(cmd):
...
     mj_cmd = 'pypy3 make_json.py'
 
     mj_ext = ''
     bq_ext = ''
     try:
-        call(mj_cmd + ' --days 1 --assert-oldest 1.9')
+        call(mj_cmd + ' --days 1 --assert-oldest {DAYS_OLD}')
```
spiffxp, line 52: Missing an f on the second string, or could also be
```suggestion
        call(f'{mj_cmd} --days 1 --assert-oldest {DAYS_OLD}')
```

kettle/update.py
```diff
@@ -38,31 +41,35 @@ def call(cmd):
...
         mj_ext = ' --reset-emitted'
 
-    call(mj_cmd + mj_ext + ' --days 1 | pv | gzip > build_day.json.gz')
-    call(bq_cmd + bq_ext + ' k8s-gubernator:build.day build_day.json.gz schema.json')
+    if os.getenv('DEPLOYMENT', 'prod') == "staging":
+        call(f'{mj_cmd} | pv | gzip > build_staging.json.gz')
+        call(f'{bq_cmd} k8s-gubernator:build.staging build_staging.json.gz schema.json')
+    else:
+        call(f'{mj_cmd} {mj_ext} --days 1 | pv | gzip > build_day.json.gz')
+        call(bq_cmd + bq_ext + ' k8s-gubernator:build.day build_day.json.gz schema.json')
```
spiffxp, line 63: style nit, maybe stick with f strings for consistency?
```suggestion
        call(f'{bq_cmd} {bq_ext} k8s-gubernator:build.day build_day.json.gz schema.json')
```

kettle/update.py
```diff
@@ -38,31 +41,35 @@ def call(cmd):
...
 
-    call(mj_cmd + mj_ext + ' --days 7 | pv | gzip > build_week.json.gz')
-    call(bq_cmd + bq_ext + ' k8s-gubernator:build.week build_week.json.gz schema.json')
+        call(f'{mj_cmd} {mj_ext} --days 7 | pv | gzip > build_week.json.gz')
+        call(bq_cmd + bq_ext + ' k8s-gubernator:build.week build_week.json.gz schema.json')
```
spiffxp, line 66:
```suggestion
        call(f'{bq_cmd} {bq_ext} k8s-gubernator:build.week build_week.json.gz schema.json')
```

kettle/update.py
```diff
@@ -38,31 +41,35 @@ def call(cmd):
...
-        call(mj_cmd + ' --days 1 --assert-oldest 1.9')
+        call(mj_cmd + ' --days 1 --assert-oldest {DAYS_OLD}')
     except OSError:
         # cycle daily/weekly tables
         bq_ext = ' --replace'
         mj_ext = ' --reset-emitted'
 
-    call(mj_cmd + mj_ext + ' --days 1 | pv | gzip > build_day.json.gz')
-    call(bq_cmd + bq_ext + ' k8s-gubernator:build.day build_day.json.gz schema.json')
+    if os.getenv('DEPLOYMENT', 'prod') == "staging":
```
spiffxp, line 58: I get that this means we don't have to change the prod deployment, but it also means someone could accidentally the prod deployment if they used a word other than "staging" for this env var.

I might at least flip it to default to staging, or go further and raise early on if this is an unrecognized value.

spiffxp (changes requested):
Some style nits, one concern about a sharp edge

amwat (commented):
I think it depends on what we are aiming for with the staging kettle. 
Most probably we just want assurance that the kettle workflow is working fine so a small representative subset of tables seems fine.

kettle/update.py
MushuEE: Sounds good, I was thinking about it and didn't want to change the default behavior. Because in Prod we just use "set image" that ENV isn't there currently and I didn't want to change the deployment.

MushuEE:
#19592

PR #19566, opened 2020-10-14
mkpj: Allow triggering and monitoring of prow jobs as well

GeoBK:
Setting the `--trigger-job` flag to `true` would load the cluster config and trigger the job. 
The tool would also wait for the job to succeed/fail and report the status of the job along with a link to the details of the job run.


prow/cmd/mkpj/main.go
```diff
@@ -204,6 +211,9 @@ func gatherOptions() options {
 	fs.IntVar(&o.pullNumber, "pull-number", 0, "Git pull number under test")
 	fs.StringVar(&o.pullSha, "pull-sha", "", "Git pull SHA under test")
 	fs.StringVar(&o.pullAuthor, "pull-author", "", "Git pull author under test")
+	fs.BoolVar(&o.triggerJob, "trigger-job", false, "Submit the job to Prow and wait for results")
+	fs.StringVar(&o.outputPath, "output-path", "", "File to which output of the triggered job would be printed")
```
alvaroaleman, line 215: What for do we need that? Doesn't it make more sense to simply print the job on stdout and print the progress stuff on stderr? Thats how other popular tools like e.G. `curl` work.

If really wanted, ppl can redirect the `stderr` stream to a file

GeoBK: If no file is provided, I am printing it by default onto stdout, should I remove this option altogether?

alvaroaleman: yeah please, that will simplify the code and as mentioned, if ppl really want this, they can simply redirect the stream

prow/pjutil/trigger.go
```diff
@@ -0,0 +1,147 @@
...
+	"k8s.io/test-infra/prow/pod-utils/downwardapi"
+)
+
+// getJobArtifactsURL returns the artifacts URL for the given job
+func getJobArtifactsURL(prowJob *pjapi.ProwJob, config *prowconfig.Config) string {
+	var identifier string
+	if prowJob.Spec.Refs != nil {
+		identifier = fmt.Sprintf("%s/%s", prowJob.Spec.Refs.Org, prowJob.Spec.Refs.Repo)
+	} else {
+		identifier = fmt.Sprintf("%s/%s", prowJob.Spec.ExtraRefs[0].Org, prowJob.Spec.ExtraRefs[0].Repo)
```
alvaroaleman, line 41: ExtraRefs might be nil, you need to access this conditionally

prow/cmd/mkpj/main.go
GeoBK: alright i'll do that

prow/pjutil/trigger.go
```diff
@@ -0,0 +1,147 @@
...
+	"k8s.io/apimachinery/pkg/fields"
+	pjapi "k8s.io/test-infra/prow/apis/prowjobs/v1"
+	prowconfig "k8s.io/test-infra/prow/config"
+	prowflagutil "k8s.io/test-infra/prow/flagutil"
+	"k8s.io/test-infra/prow/gcsupload"
+	"k8s.io/test-infra/prow/pod-utils/downwardapi"
+)
+
+// getJobArtifactsURL returns the artifacts URL for the given job
+func getJobArtifactsURL(prowJob *pjapi.ProwJob, config *prowconfig.Config) string {
```
alvaroaleman, line 36: Hm we should probably store an `artifactsURL` on the job to not need this

prow/pjutil/trigger.go
```diff
@@ -0,0 +1,147 @@
...
+	jobBasePath, _, _ := gcsupload.PathsForJob(config.Plank.GetDefaultDecorationConfigs(identifier).GCSConfiguration, &spec, "")
+	return fmt.Sprintf("%s%s/%s",
+		config.Deck.Spyglass.GCSBrowserPrefix,
+		config.Plank.GetDefaultDecorationConfigs(identifier).GCSConfiguration.Bucket,
+		jobBasePath,
+	)
+}
+
+// Calls toJSON method on a jobResult type and writes it to the output path
+func writeResultOutput(prowjobResult JobResult, outputPath string, fileSystem afero.Fs) error {
```
alvaroaleman, line 53: You can save yourself the interface definition and implementation by just accepting an empty interface (`interface{}`) here, just like `json.Marshal` does.

Also, I would suggest to yaml marshal this instead for readability?

prow/pjutil/trigger.go
```diff
@@ -0,0 +1,147 @@
...
+			return fmt.Errorf("failed to create watch for ProwJobs: %w", err)
+		}
+
+		for event := range w.ResultChan() {
+			prowJob, ok := event.Object.(*pjapi.ProwJob)
+			if !ok {
+				return fmt.Errorf("received an unexpected object from Watch: object-type %s", fmt.Sprintf("%T", event.Object))
+			}
+
+			prowJobArtifactsURL := getJobArtifactsURL(prowJob, config)
```
alvaroaleman, line 117: I don't think this needs to be in the loop and called on every event

prow/pjutil/trigger.go
```diff
@@ -0,0 +1,147 @@
...
+	Status       pjapi.ProwJobState `json:"status"`
+	ArtifactsURL string             `json:"prowjob_artifacts_url"`
+	URL          string             `json:"prowjob_url"`
+}
+
+func (p *prowjobResult) toJSON() ([]byte, error) {
+	return json.MarshalIndent(p, "", "    ")
+}
+
+func TriggerProwJob(o prowflagutil.KubernetesOptions, prowjob *pjapi.ProwJob, config *prowconfig.Config, envVars map[string]string, fileSystem afero.Fs, dryRun bool, outputPath string) error {
```
stevekuznetsov, line 87: This seems complicated. Why not

```go
	// ...
	var result prowJobResult
	var continue bool
	for {
		result, continue = resultForJob(client, selector.String())
		if !continue {
			break
		}
	}

	result.ArtifactsURL = getJobArtifactsURL(prowJob, config)
	out, err := json.Marshal(result)
	if err != nil {
	  logrus.WithError(err).Fatal("Could not marshal result)
	}
	fmt.Println(out)
}

func resultForJob(client *v1.Client, selector string) (result prowJobResult, continue bool) {
	w, err := client.Watch(context.TODO(), metav1.ListOptions{FieldSelector: selector})
	if err != nil {
		return fmt.Errorf("failed to create watch for ProwJobs: %w", err)
	}

	for event := range w.ResultChan() {
		prowJob, ok := event.Object.(*pjapi.ProwJob)
		if !ok {
			return fmt.Errorf("received an unexpected object from Watch: object-type %s", fmt.Sprintf("%T", event.Object))
		}

		switch prowJob.Status.State {
		case pjapi.FailureState, pjapi.AbortedState, pjapi.ErrorState, pjapi.SuccessState: // these are terminal
			return &prowjobResult{
				Status:       prowJob.Status.State,
				URL:          prowJob.Status.URL,
			}, false
		default:
			return nil, true
		}
	}
}
```

stevekuznetsov (commented):
I also don't really think we care about the artifact link. It's fine to just give a link to Spyglass.

alvaroaleman (commented):
/ok-to-test

prow/pjutil/trigger.go
alvaroaleman: No, there is only the spyglass url (from which we might parse the artifact url)

prow/cmd/mkpj/main.go
```diff
@@ -250,13 +258,21 @@ func main() {
...
+		b, err := yaml.Marshal(&pj)
+		if err != nil {
+			logrus.WithError(err).Fatal("Error marshalling YAML.")
+		}
+		fmt.Print(string(b))
+		if o.local {
+			logrus.Info("Use 'bazel run //prow/cmd/phaino' to run this job locally in docker")
+		}
+
+		os.Exit(0)
```
alvaroaleman, line 271: Just `return`

prow/pjutil/trigger.go
```diff
@@ -0,0 +1,125 @@
...
+			return fmt.Errorf("failed to watch job: %w", err)
+		}
+		if !shouldContinue {
+			break
+		}
+	}
+	if result.Status != pjapi.SuccessState {
+		logrus.Warn("job failed")
+	}
+	result.ArtifactsURL = getJobArtifactsURL(prowJob, config)
```
alvaroaleman, line 118: Can we only use the spyglass url in `.satus.url` as steve suggested and save ourselves the calculation of the artifacts url? There is a link to the latter in the former

GeoBK: I dont know why we were calculating it, but I left it in since it was because it was explicitly calculated in the cvp-trigger tool. I am removing it for now but we might need to revisit this later if it is necessary in cvp-trigger

prow/pjutil/trigger_test.go
```diff
@@ -0,0 +1,21 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package pjutil
+
+type fakeJobResult struct {
+	err error
```
alvaroaleman, line 20: Remainder?

GeoBK: We removed all the functions which we were testing in the earlier iteration so I had to remove the corresponding tests as well. I'll add the tests for the new functions, just want to confirm that the overall code looks good before that.

prow/cmd/mkpj/main.go
```diff
@@ -189,6 +191,10 @@ func (o *options) Validate() error {
 		return err
 	}
 
+	if err := o.kubeOptions.Validate(false); err != nil {
```
stevekuznetsov, line 194: This is only required when `--trigger-job` is set, right?

prow/pjutil/trigger.go
```diff
@@ -0,0 +1,101 @@
...
+			return &prowjobResult{
+				Status: prowJob.Status.State,
+				URL:    prowJob.Status.URL,
+			}, false, nil
+		}
+	}
+	return nil, true, nil
+}
+
+// TriggerProwJob would trigger the job provided by the prowjob parameter
```
stevekuznetsov, line 62: nit: Generally `trigger` just means to start it, not also to follow. This could be either broken into two methods or named a bit better

prow/pjutil/trigger.go
```diff
@@ -0,0 +1,101 @@
...
+func TriggerProwJob(o prowflagutil.KubernetesOptions, prowjob *pjapi.ProwJob, config *prowconfig.Config, envVars map[string]string, dryRun bool) error {
+	logrus.Info("getting cluster config")
+	pjclient, err := o.ProwJobClient(config.ProwJobNamespace, dryRun)
+	if err != nil {
+		return fmt.Errorf("failed getting prowjob client: %w", err)
+	}
+
+	logrus.WithFields(ProwJobFields(prowjob)).Info("submitting a new prowjob")
+	created, err := pjclient.Create(context.TODO(), prowjob, metav1.CreateOptions{})
```
stevekuznetsov, line 71: Might as well use `context.Background()` since this explicitly not passing a context

prow/pjutil/trigger.go
```diff
@@ -0,0 +1,101 @@
...
+	prowflagutil "k8s.io/test-infra/prow/flagutil"
+)
+
+type prowjobResult struct {
+	Status pjapi.ProwJobState `json:"status"`
+	URL    string             `json:"prowjob_url"`
+}
+
+func resultForJob(pjclient prowv1.ProwJobInterface, selector string) (*prowjobResult, bool, error) {
+	w, err := pjclient.Watch(context.TODO(), metav1.ListOptions{FieldSelector: selector})
```
stevekuznetsov, line 40: Might as well use `context.Background()` since this explicitly not passing a context

prow/pjutil/trigger.go
```diff
@@ -0,0 +1,103 @@
...
+		prowJob, ok := event.Object.(*pjapi.ProwJob)
+		if !ok {
+			return nil, false, fmt.Errorf("received an unexpected object from Watch: object-type %s", fmt.Sprintf("%T", event.Object))
+		}
+
+		switch prowJob.Status.State {
+		case pjapi.FailureState, pjapi.AbortedState, pjapi.ErrorState, pjapi.SuccessState:
+			return &prowjobResult{
+				Status: prowJob.Status.State,
+				URL:    prowJob.Status.URL,
```
alvaroaleman, line 54: The url is part of the status so I think there isn't really much reason for the `prowjobResultType`, you can just use `prowjob.Status` instead

prow/pjutil/trigger_test.go
alvaroaleman: This is still unreachable code

prow/pjutil/trigger_test.go
```diff
@@ -0,0 +1,191 @@
...
+		t.Run(tc.name, func(t *testing.T) {
+			cs := fake.NewSimpleClientset(&tc.args.pj)
+			cs.Fake.PrependWatchReactor("prowjobs", func(action coretesting.Action) (bool, watch.Interface, error) {
+				ret := watch.NewFakeWithChanSize(len(tc.args.watchResults), true)
+				for _, res := range tc.args.watchResults {
+					ret.Modify(&res)
+				}
+				return true, ret, nil
+			})
+			pjr, shouldContinue, err := resultForJob(cs.ProwV1().ProwJobs("prowjobs"), tc.args.selector)
```
alvaroaleman, line 179: You can infer the selector from `tc.args.pj`

prow/cmd/mkpj/main.go
```diff
@@ -38,14 +38,16 @@ type options struct {
...
-
-	baseRef    string
-	baseSha    string
-	pullNumber int
-	pullSha    string
-	pullAuthor string
-	org        string
-	repo       string
+	triggerJob    bool
+	outputPath    string
```
alvaroaleman, line 42: This is unused

alvaroaleman (approved):

PR #19569, opened 2020-10-14
Delete wrong servicemonitor from monitoring

chaodaiG:


cjwagner (approved):
The correct properly named service monitors for crier and horologium are farther down in the file.

PR #19576, opened 2020-10-14
Update Kettle Staging docs with more detail

MushuEE:
Just update and add detail to staging docs. This is in an effort to have a way to test kettle changes without going live in Prod.

/area kettle
/hold.  cancel when live


MushuEE:
#19592

spiffxp:
needs rebase

PR #19578, opened 2020-10-15
Simplify sig-arch conformance gate

hh:
Attempt to remove mount point and security contexts from conformance-job


oomichi:
Why do we need to revert the commit 04ef0532749efc93aa3b77777a8419320b6bc4f2 ?

hh:
We seem to be having trouble with entry point wrapper:
https://prow.k8s.io/log?job=apisnoop-conformance-gate&id=1316546413642911744

A similar job definition on prow.cncf.io (sans the decorators) seems to work
https://prow.cncf.io/log?job=cg-b3&id=1316541295774142464

spiffxp (approved):
/approve
/lgtm

Issue #19534, opened 2020-10-12
Plank errors listing pods cause plank to try to recreate pods for all running ProwJobs.


cjwagner:
**What happened**:
Plank failed to list pods and proceeded to try to create a pod for every running ProwJob that it failed to list pods for. This resulted in many `pod <pod-name> already exists` errors.

**What you expected to happen**:
If Plank fails to list pods for a build cluster, ProwJobs from that build cluster should be ignored for that sync loop.

/help

cjwagner:
> We could add a check to only create pods when getting them before that returns a 404? The cache not being synced returns a distinct error.

That sounds sensible to me.

matthyx:
No longer an issue with the `prow-controller-manager`.
/close

PR #19581, opened 2020-10-15
Plank: only create the pod when it is missing

matthyx:
fixes https://github.com/kubernetes/test-infra/issues/19534

### DIFF
--- a/prow/plank/reconciler.go
+++ b/prow/plank/reconciler.go
@@ -608,21 +608,27 @@ func (r *reconciler) startPod(pj *prowv1.ProwJob) (string, string, error) {
 		return "", "", err
 	}
 	pod.Namespace = r.config().PodNamespace
+	podName := types.NamespacedName{Namespace: pod.Namespace, Name: pod.Name}
 
 	client, ok := r.buildClients[pj.ClusterAlias()]
 	if !ok {
 		// TODO: Terminal error to prevent requeuing
 		return "", "", fmt.Errorf("unknown cluster alias %q", pj.ClusterAlias())
 	}
-	err = client.Create(r.ctx, pod)
-	r.log.WithFields(pjutil.ProwJobFields(pj)).Debug("Create Pod.")
-	if err != nil {
-		return "", "", err
+
+	// only create the pod when it is really missing
+	if err := client.Get(r.ctx, podName, pod); err != nil {
+		if kerrors.IsNotFound(err) {
+			r.log.WithFields(pjutil.ProwJobFields(pj)).Debug("Create Pod.")
+			err := client.Create(r.ctx, pod)
+			if err != nil {
+				return "", "", err
+			}
+		}
 	}
 
 	// We must block until we see the pod, otherwise a new reconciliation may be triggered that tries to create
 	// the pod because its not in the cache yet, errors with IsAlreadyExists and sets the prowjob to failed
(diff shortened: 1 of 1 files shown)


alvaroaleman:
@matthyx this won't fix the issue, prow.k8s.io uses the prow controller manager whos plank version is implemented in plank/reconciler.go

matthyx:
> @matthyx this won't fix the issue, prow.k8s.io uses the prow controller manager whos plank version is implemented in plank/reconciler.go

like that @alvaroaleman ?

prow/plank/reconciler.go
```diff
@@ -608,21 +608,27 @@ func (r *reconciler) startPod(pj *prowv1.ProwJob) (string, string, error) {
...
 		// TODO: Terminal error to prevent requeuing
 		return "", "", fmt.Errorf("unknown cluster alias %q", pj.ClusterAlias())
 	}
-	err = client.Create(r.ctx, pod)
-	r.log.WithFields(pjutil.ProwJobFields(pj)).Debug("Create Pod.")
-	if err != nil {
-		return "", "", err
+
+	// only create the pod when it is really missing
+	if err := client.Get(r.ctx, podName, pod); err != nil {
```
alvaroaleman, line 620: This fixes the issue but causes some new ones (and I am very sorry for suggesting this without properly reflecting about the implications):
* The wait.Poll below will always timeout if the cache was unable to sync
* The only indication of having a buildcluster with misconfigured rbac is now klog when it logs the cache sync errors, however this is unstructured and will hence be dropped by many log collectors, meaning a Prow admin will never see those errors in their aggregated logging
* I generally am not confident that the whole state model plank has will work properly if it can not see what it did

Another idea would be to wait in the `main` for the build cluster managers to start and cachesync, however that isn't really trivial either:
* We would need to start the build cluster managers or at least their cache, else the cache will never sync
* But the main `mgr` will also start them and they do not support getting started multiple times
* The `Start` is blocking so maybe we could write a shim to make them support getting started multiple times by blocking subsequent starts on the first start? It is a bit weird though because `Start` takes a channel that is supposed to stop it so that shim would need to make sure that it gets stopped if any of the stop channels get closed

Another idea would be a goroutine that checks if the caches are synced and signals plank to ignore clusters whose caches haven't synced? Seems a lot of work though.

alvaroaleman: @cjwagner actually, I am getting confused about how this was possible. The `Controller` that wraps the `Reconciler` in plankv2 should block indefinitely if it can not sync a cache: https://github.com/kubernetes-sigs/controller-runtime/blob/5118ae568d16e0b29bee8b82c58114b50d82574d/pkg/internal/controller/controller.go#L153  (tbf that's not great behavior either)

Did this issue happen with the prow controller manager or the old plank? What version exactly? What was the underlying issue?

chaodaiG: I don't have enough context for https://github.com/kubernetes/test-infra/issues/19534, so probably won't able to provide much help on where and when cjwagner saw the issue. Would like to let you know that cjwagner is OOO for multiple weeks, so probably won't get back to us until he came back

cjwagner: I'm back!
Looks like that error came from plank based on the timing. Since then all our Prow instances have been updated to use the prow-controller-manager instead. If this is not a concern with the prow-controller-manager then we can just consider this done?

matthyx:
Welcome back @cjwagner !
/close

PR #19584, opened 2020-10-15
make local-e2e-yaml pass and easy to read

jayunit100:
Might as well just skip all `Serial` tests.  local-e2e doesn't work otherwise.
To add all the escape sequences in here would be almost impossible to read: 
- `[sig-scheduling] SchedulerPreemption [Serial] validates basic preemption works`
- ` [sig-scheduling] SchedulerPreemption [Serial] validates lower priority pod preemption by critical pod [Conformance] `
and so on.

Currently, these tests fail conformance on a local-e2e ... I assume that nobody is using this signal since it automatically skips them... 

```
Kubernetes e2e suite: [sig-scheduling] SchedulerPreemption [Serial] validates basic preemption works [Conformance] expand_more1m0s | Kubernetes e2e suite: [sig-scheduling] SchedulerPreemption [Serial] validates basic preemption works [Conformance] expand_more | 1m0s
-- | -- | --
Kubernetes e2e suite: [sig-scheduling] SchedulerPreemption [Serial] validates basic preemption works [Conformance] expand_more | 1m0s
Kubernetes e2e suite: [sig-scheduling] SchedulerPreemption [Serial] validates lower priority pod preemption by critical pod [Conformance] expand_less | Kubernetes e2e suite: [sig-scheduling] SchedulerPreemption [Serial] validates lower priority pod preemption by critical pod [Conformance] expand_less
Kubernetes e2e suite: [sig-scheduling] SchedulerPreemption [Serial] validates lower priority pod preemption by critical pod [Conformance] expand_less
```

It will be alot simpler if we just skip all `Serial` tests, to get this job passing.  We can make the filter a little more detailed later, but I guess the priority for now should be "the simplest thing we can do to make this test pass".

### DIFF
--- a/config/jobs/kubernetes/sig-testing/local-e2e.yaml
+++ b/config/jobs/kubernetes/sig-testing/local-e2e.yaml
@@ -28,7 +28,7 @@ presubmits:
         - --deployment=local
         - --ginkgo-parallel=1
         - --provider=local
-        - --test_args=--ginkgo.focus=\[Conformance\] --ginkgo.skip=\[sig\-scheduling\]\sSchedulerPreemption\s\[Serial\]\s\[It\]\svalidates\slower\spriority\spod\spreemption\sby\scritical\spod\s\[Conformance\]|\[sig\-apps\]\sDaemon\sset\s\[Serial\]\sshould\srollback\swithout\sunnecessary\srestarts\s\[Conformance\]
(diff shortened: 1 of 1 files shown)


jayunit100:
follow up to #19142 @dims , this should just make this test pass, finally.... Other `Serial` tests which break also.

Issue #19588, opened 2020-10-15
Enable golangci-lint for static analysis


bvwells:
**What should be cleaned up or changed**:

This is a follow up to PR #19444. [golangci-lint](https://github.com/golangci/golangci-lint) can be used to identify various static analysis issues such as dead code, poor formatting, inefficient assignment and use of deprecated APIs and ultimately bugs.

I think that this tool would be useful in identifying issues in the code base. 

**Provide any links for context**:

https://github.com/golangci/golangci-lint

bvwells:
As promised @alvaroaleman, here is the suggestion to add golangci-lint. I'm happy to have a go at enabling it.

alvaroaleman:
> here is the suggestion to add golangci-lint. I'm happy to have a go at enabling it.

That would be great :) It requires adding a job in config/jobs/kubernetes/test-infra/test-infra-presubmits.yaml

bvwells:
@alvaroaleman something like this?

```

  - name: pull-test-infra-verify-golangci-lint
    always_run: true
    run_if_changed: '\.go'
    decorate: true
    spec:
      containers:
        - image: golangci/golangci-lint:v1.29.0
          command:
            - make
          args:
            - golangci-lint
    annotations:
      testgrid-dashboards: presubmits-test-infra
      testgrid-tab-name: verify-golangci-lint
```

alvaroaleman:
Yes, except for the `run_if_changed`. You also need to add a config for it so this actually passes

bvwells:
Sounds good. I haven't used Bazel much, but was trying to workout how to add golangci-lint to the build chain. Does this need to be done as well?

alvaroaleman:
@fejta is the one who would have the last word on that, I personally think its fine without and just adding a new job

fejta:
Does `bazelisk run @io_k8s_repo_infra//hack:verify-golangci-lint` already work? https://github.com/kubernetes/repo-infra/blob/master/hack/BUILD.bazel#L139

bvwells:
I'll have a play and double check.

bvwells:
I might be doing something wrong, but I get:

```
% bazelisk run @io_k8s_repo_infra//hack:verify-golangci-lint       
INFO: Analyzed target @io_k8s_repo_infra//hack:verify-golangci-lint (1 packages loaded, 483 targets configured).
INFO: Found 1 target...
[...]

PR #19589, opened 2020-10-15
Add golangci-lint to presubmit checks with minimal configuration

bvwells:
Add golangci-lint to presubmit commits to address #19588.

### DIFF
--- a/config/jobs/kubernetes/test-infra/test-infra-presubmits.yaml
+++ b/config/jobs/kubernetes/test-infra/test-infra-presubmits.yaml
@@ -97,3 +97,17 @@ presubmits:
         - --warnings=duplicate-job-refs
     annotations:
       testgrid-dashboards: presubmits-test-infra
+
+  - name: pull-test-infra-verify-golangci-lint
+    decorate: true
+    run_if_changed: '\.go'
+    spec:
+      containers:
+        - image: golangci/golangci-lint:v1.31.0
+          command:
+            - golangci-lint
+          args:
+            - run
+    annotations:
+      testgrid-dashboards: presubmits-test-infra
+      testgrid-tab-name: verify-golangci-lint
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,0 +1,8 @@
+run:
+  deadline: 5m
+
+linters:
+  disable-all: true
+  enable:
+    - gofmt
+    - misspell


alvaroaleman (commented):
/ok-to-test

bvwells:
@alvaroaleman is added the precommit hook to use v1.31.0 of golangci. Is this okay or should it match the version defined in the repo?

config/jobs/kubernetes/test-infra/test-infra-presubmits.yaml
```diff
@@ -97,3 +97,17 @@ presubmits:
         - --warnings=duplicate-job-refs
     annotations:
       testgrid-dashboards: presubmits-test-infra
+
+  - name: pull-test-infra-verify-golangci-lint
+    decorate: true
+    run_if_changed: '\.go'
+    spec:
+      containers:
+        - image: golangci/golangci-lint:v1.31.0
```
spiffxp, line 106: Is this in sync with the version used via bazel and the repo-infra route? ref: https://github.com/kubernetes/test-infra/issues/19588#issuecomment-709593545

I'm guessing @fejta would rather see us upgrade it in repo-infra

alvaroaleman: What I don't understand is the value of pulling this in via repo infra. We want to specify a version of this we use, not get a specific version pulled in transiently by a dependency. 

Additionally, linting needs a lot of ram, downstream we currently request 10 gig. Coupling this with the generic bazel job means that we need to request as much for the bazel job, even though its not needed for most of its runtime.

bvwells: @spiffxp I'm more than happy to have a go at upgrading golang-ci in repo-infra, but not sure on the knock effect this will have...

spiffxp:
/cc @fejta 
for your opinion

config/jobs/kubernetes/test-infra/test-infra-presubmits.yaml
fejta: Upgrading golangci-lint in repo-infra is not a problem

Ideally everything needed to verify the repo happens inside `bazel test //...`

Ideally we have a single version of golangci-lint to worry about across the org.

alvaroaleman: @fejta can you either provide guidance on how to continue here with bazel or let us add the job outside of it?

FWIW, I don't see any value in having the same version of a linter across the org. Everyone should be able to update at their own pace.

alvaroaleman:
We added golang-ci lint in https://github.com/kubernetes/test-infra/pull/21751 so this isn't needed anymore. Thanks for your work, though!
/close

PR #19594, opened 2020-10-16
[sig-scalability] Run 5k performance test with e2-medium instances

jprzychodzen:
Change VM instance type for 5k performance test to `e2-medium`.

According to https://cloud.google.com/products/calculator e2-medium is more cost-effective in 12 hours  x 7 days model.

I've run a one-off test to confirm that test passes, it took similar time as n1-standard-1 version

/assign @wojtek-t


wojtek-t:
/lgtm
/approve

@mborsz - FYI (since you're oncall this week)

PR #19597, opened 2020-10-16
kubernetes-sigs/k8s-container-image-promoter: add prow plugins

cpanato:
Add common prow plugins to kubernetes-sigs/k8s-container-image-promoter repo

ref: https://github.com/kubernetes-sigs/k8s-container-image-promoter/issues/266

/cc @justaugustus @nikhita

### DIFF
--- a/config/prow/plugins.yaml
+++ b/config/prow/plugins.yaml
@@ -947,6 +947,11 @@ plugins:
   kubernetes-sigs/security-profiles-operator:
   - release-note
 
+  kubernetes-sigs/k8s-container-image-promoter:
+  - mergecommitblocker
+  - override
+  - release-note
+
   containerd/cri:
   - assign
   - cla


config/prow/plugins.yaml
```diff
@@ -947,6 +947,15 @@ plugins:
   kubernetes-sigs/security-profiles-operator:
   - release-note
 
+  kubernetes-sigs/k8s-container-image-promoter:
+  - blockade
```
nikhita, line 951: blockade, milestone, milestoneapplier, require-matching-label plugins would also need be specifically configured for the repo.

For example, for milestone_applier, you'll need to tell what milestones to apply:

https://github.com/kubernetes/test-infra/blob/12c654eaf2c5aa900963d63233df08e2fcb91d94/config/prow/plugins.yaml#L327-L335

cpanato: ahh, thanks! I think we can drop those, for now, don't think this is a hard requirement for the repo


thanks for your review @nikhita

spiffxp (approved):
/approve
/lgtm

PR #19600, opened 2020-10-16
[scalability] Migrate performance tests to E2

jprzychodzen:
/cc @mborsz 
/assign @wojtek-t 

Migrate performance tests to E2 instance family.

### DIFF
--- a/config/jobs/kubernetes/sig-release/release-branch-jobs/1.16.yaml
+++ b/config/jobs/kubernetes/sig-release/release-branch-jobs/1.16.yaml
@@ -210,6 +210,7 @@ periodics:
       - --cluster=e2e-big
       - --extract=ci/latest-1.16
       - --gcp-node-image=gci
+      - --gcp-node-size=e2-medium
       - --gcp-nodes=100
       - --gcp-project-type=scalability-project
       - --gcp-zone=us-east1-b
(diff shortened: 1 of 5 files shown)


jprzychodzen:
I've tested manually that master job passes with comparable time

config/jobs/kubernetes/sig-scalability/sig-scalability-presubmit-jobs.yaml
```diff
@@ -33,6 +33,7 @@ presubmits:
         - --cluster=
         - --extract=local
         - --flush-mem-after-build=true
+        - --gcp-node-size=e2-medium
```
wojtek-t, line 36: I'm a bit reluctant to changing presubmit immediately with postsubmits. We should split them into two separate PR.

Also @mborsz as current and next-week oncall - FYI

jprzychodzen: Sure, removed presubmit from this PR, I will create another one for presubmits.

jprzychodzen: Presubmit only PR is available in #19602

config/jobs/kubernetes/sig-release/release-branch-jobs/1.16.yaml
```diff
@@ -939,6 +940,7 @@ presubmits:
         - --cluster=
         - --extract=local
         - --flush-mem-after-build=true
+        - --gcp-node-size=e2-medium
```
wojtek-t, line 943: Sorry - missed that. I don't think we should be changing presubmits especially for release branches in the first step...

wojtek-t:
Closing in favor of https://github.com/kubernetes/test-infra/pull/20043

/close

Issue #19592, opened 2020-10-15
Build Kettle Staging Workflow


MushuEE:
# Project Plan: Kettle Staging Env

## 1. Introduction 
Currently [Kettle](https://github.com/kubernetes/test-infra/tree/master/kettle) is a service that runs in GCP, scrapes a collection of GCS buckets for Job info/status. It takes the job results, stores them in a local SQL db and then passes the data to [BQ tables](https://bigquery.cloud.google.com/dataset/k8s-gubernator:build). This process is performed for every job in storage which is currently in the range of 600G of job data. It takes up to a few hours to complete an upload cycle.

Kettle has no current e2e tests or integration tests, the only way to validate functionality is to test live. The biggest customer of kettle is Triage, these are not critical services but down time does remove visibility to job metics (primarily flake data). 

Kettle would greatly benefit from having a staging environment that can be used to trial changes and dirty builds without impacting services.

### 2.1 Overview
Kettle staging will mirror Kettle production functionality and use the production image. This should limit issue triage onus. However Kettle Staging should be MUCH faster than Production for quick iteration. This can be accomplished by limiting the number of builds collected per job. Staging should also be limited in disk usage, there is no need to test the full breath of job collection. *There might be a case where there are failures based on malformed job data that staging could miss.  

Infra needs:
- [x] New SSD for staging PCV
- [x] New table in BQ build for staging to publish to
- [x] Create new staging PVC
- [x] Create new staging Deployment 

Code Changes:
- [x] `README.md` updated with information on staging
- [x] `Makefile` updated with commands to update staging env
- [x] Split `update.py` functionality to write to different BQ tables based on `DEPLOYMENT` (type) env
- [x] Create new `deployment.yaml` and `pvc.yaml` for staging

### 2.2 Details
Env vars will be used to fork functionality and set the max number of builds to pull for each job. There are several layers to pass information through. Kubernetes yaml -> Docker -> `runner.sh` -> `update.py` -> helper scripts. Env vars are the single interface that can be accessed by and set at multiple levels of execution and can be enforced at the Kubernetes deployment level. 

#### Roles:
/assign 
/area kettle
## 3. Estimates and Deadlines
Week of 10/16 deploy first instance. Week of 10/23 test with new images and validate functionality, builds, logs etc. 
- [x] Deployed
- [x] Validated
- [x] Tested 
/cc @BenTheElder @amwat @spiffxp

PR #19609, opened 2020-10-16
Fix yaml issues in Kettle staging files

MushuEE:
dedupe env
make int str
specify pvc

/assign @spiffxp 
/area kettle


spiffxp (approved):
/approve
/lgtm

MushuEE:
#19592

PR #19610, opened 2020-10-16
do not run other e2e tests for upgrade test

zshihang:
kubenetes_e2e.py is wrapper over kubetest where we need to explicitly
tell it not to run other e2e tests.


config/jobs/kubernetes/sig-auth/serviceaccount-admission-controller-migration-config.yaml
```diff
@@ -19,6 +19,7 @@ periodics:
         - runner.sh
         - /workspace/scenarios/kubernetes_e2e.py
         args:
+        - --test=false
```
liggitt, line 24: a comment here would help future us understand what seems nonsensical at first

zshihang: done

PR #19615, opened 2020-10-17
Increase memory for pull-kubernetes-node-e2e

eddiezane:
The presubmit pull-kubernetes-node-e2e has been consistently flaking due to early build failures with a reason of `OOMKilled`.

The job is currently set to 3 gigs of memory which seems a bit low.

This PR bumps the resource limit/request to 6 gigs to match other jobs.

History:

https://prow.k8s.io/job-history/gs/kubernetes-jenkins/pr-logs/directory/pull-kubernetes-node-e2e

Recent:

https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/95625/pull-kubernetes-node-e2e/1317239449473519616/
https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/94115/pull-kubernetes-node-e2e/1317172507119194112
https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/95640/pull-kubernetes-node-e2e/1317184022786347008

Note: I could have totally misinterpreted just about everything here.

### DIFF
--- a/config/jobs/kubernetes/sig-node/sig-node-presubmit.yaml
+++ b/config/jobs/kubernetes/sig-node/sig-node-presubmit.yaml
@@ -37,10 +37,10 @@ presubmits:
         resources:
           limits:
             cpu: 4
-            memory: 3Gi
+            memory: 6Gi
           requests:
             cpu: 4
-            memory: 3Gi
+            memory: 6Gi
   - name: pull-kubernetes-e2e-containerd-gce
     always_run: false
     optional: true


hasheddan (approved):
This seems reasonable to me, thanks @eddiezane! Would be good to follow up by taking a peek at metrics if you have access.

/lgtm

spiffxp:
Hot take: blindly raising resource requests and timeouts is the wrong approach here.  This could be a sign of a performance regression

spiffxp:
The reason this job had less memory than other jobs is its execution model is different.  Where other e2e jobs run their tests within the pod, node e2e jobs ssh to vm's and run their tests there.

spiffxp:
Historical context: https://github.com/kubernetes/test-infra/issues/18851 - there are links here that can be used to see what the performance of the job's cpu and memory is over time (search for "metrics explorer")

eddiezane:
@spiffxp thanks for the great insight. Do you recommend rolling this back then?

In the meantime I'll keep an eye on the metrics you linked. It's hard to tell right now because of the expected Sunday drop off.

What else can I do to help?

dims:
@eddiezane let's revert please given @spiffxp's advice! :)

spiffxp:
Encountered again in https://github.com/kubernetes/kubernetes/pull/96108#issuecomment-720593001

I think I found the culprit (ref: https://gist.github.com/spiffxp/01990c3bb2779c29feffecf76902e905)

When I looked 10 days ago:
- prior to 2020-10-09, percentage of builds oomkilled was within the range 0-4%
- from 2020-10-09 onward, percentage of builds oomkilled was within the range 7-22%

The likely culprit is https://github.com/kubernetes/kubernetes/pull/95239, which updated a number kubelet and node-related package builds while adding dual stack support.

From 2020-10-18 - 2020-10-20 while this PR was merged, percentage of builds oomkilled was within the range 0-6%

It's not immediately clear to me how we could trim the build down while keeping dual stack, so I think @eddiezane's approach of raising the memory limit is correct.

PR #19620, opened 2020-10-19
Add --extract-{ci,release}-bucket flags to kubetest

spiffxp:
This is to support migrating jobs to community-owned buckets:

- ci: kubernetes-release-dev -> k8s-release-dev
- release: kubernetes-release -> k8s-release

The flags default to the google-owned kubernetes buckets, so jobs
will require config changes to migrate

kubetest ends up setting KUBERNETES_RELEASE_URL and calling out
to kubernetes/kubernetes/cluster/get-kube.sh, which will then
use the env var instead of any of its built-in defaults

ref:
https://github.com/kubernetes/k8s.io/issues/846#issuecomment-705722852oref

### DIFF
--- a/kubetest/main.go
+++ b/kubetest/main.go
@@ -57,32 +57,34 @@ var (
 )
 
 type options struct {
-	build              buildStrategy
-	charts             bool
-	checkLeaks         bool
-	checkSkew          bool
-	cluster            string
-	clusterIPRange     string
-	deployment         string
-	down               bool
-	dump               string
-	dumpPreTestLogs    string
-	extract            extractStrategies
-	extractSource      bool
-	flushMemAfterBuild bool
-	focusRegex         string
-	gcpCloudSdk        string
-	gcpMasterImage     string
-	gcpMasterSize      string
-	gcpNetwork         string
-	gcpNodeImage       string
-	gcpImageFamily     string
-	gcpImageProject    string
-	gcpNodes           string
-	gcpNodeSize        string
-	gcpProject         string
-	gcpProjectType     string
-	gcpServiceAccount  string
+	build                buildStrategy
+	charts               bool
+	checkLeaks           bool
+	checkSkew            bool
+	cluster              string
+	clusterIPRange       string
+	deployment           string
+	down                 bool
+	dump                 string
+	dumpPreTestLogs      string
+	extract              extractStrategies
+	extractCIBucket      string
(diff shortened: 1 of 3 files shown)


kubetest/main.go
```diff
@@ -135,6 +137,8 @@ func defineFlags() *options {
 	flag.StringVar(&o.dump, "dump", "", "If set, dump bring-up and cluster logs to this location on test or cluster-up failure")
 	flag.StringVar(&o.dumpPreTestLogs, "dump-pre-test-logs", "", "If set, dump cluster logs to this location before running tests")
 	flag.Var(&o.extract, "extract", "Extract k8s binaries from the specified release location")
+	flag.StringVar(&o.extractCIBucket, "extract-ci-bucket", "kubernetes-release-dev", "Extract k8s CI binaries from the specified GCS bucket")
+	flag.StringVar(&o.extractReleaseBucket, "extract-release-bucket", "kubernetes-release", "Extract k8s CI binaries from the specified GCS bucket")
```
amwat, line 141:
```suggestion
	flag.StringVar(&o.extractReleaseBucket, "extract-release-bucket", "kubernetes-release", "Extract k8s release binaries from the specified GCS bucket")
```

amwat: This looks more like we need a way to override the two default hardcoded values, and not as much we need a way to customize both values in a specific run? Do we want to consider using experimental envs instead?

A more involved change would modify the corresponding strategies, but given the current maintenance state this minimal change LGTM.

amwat (commented):
/lgtm
/hold

kubetest/main.go
spiffxp: >This looks more like we need a way to override the two default hardcoded values, and not as much we need a way to customize both values in a specific run? 

Migrating kubernetes-release is going to be more involved, so I was anticipating upgrade jobs for an existing release to the in-development release pulling from kubernetes-release and k8s-release-dev in the interim.

>Do we want to consider using experimental envs instead?

I'd rather avoid envs and would prefer flags for less magic.  Though admittedly kubernetes/kubernetes scripts by themselves will need envs eventually.

> A more involved change would modify the corresponding strategies, but given the current maintenance state this minimal change LGTM.

Agreed.  The current strategies parse themselves from the string passed to extract.  I considered something like `k8s-release-dev:ci/latest`, but as you said, that seemed like it more be more invasive.

PR #19628, opened 2020-10-20
Add docker provider label for CAPI

benmoss:
Adds a label for the Docker provider for CAPI

@detiber @vincepri @fabriziopandini

### DIFF
--- a/label_sync/labels.yaml
+++ b/label_sync/labels.yaml
@@ -1515,6 +1515,11 @@ repos:
         target: both
         prowPlugin: label
         addedBy: anyone
+      - color: 0052cc
+        description: Issues or PRs related to docker provider
+        name: area/provider/docker
+        target: both
+        addedBy: label
   kubernetes-sigs/cluster-api-provider-azure:
     labels:
       - color: c7def8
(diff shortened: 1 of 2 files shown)


label_sync/labels.yaml
```diff
@@ -49,6 +49,11 @@ default:
       name: area/provider/digitalocean
       target: both
       addedBy: label
+    - color: 0052cc
+      description: Issues or PRs related to docker provider
+      name: area/provider/docker
+      target: both
+      addedBy: label
```
detiber, line 56: I believe these labels are intended for k8s cloud-provider integration rather than cluster-api providers. It might be better to add it to the `cluster-api` specific configuration below to avoid syncing this to other repositories (such as k/k)

cpanato:
@benmoss need to follow this https://github.com/kubernetes/test-infra/tree/master/label_sync#usage to make bazel happy

benmoss:
@detiber @cpanato fixed both! thanks

cpanato (approved):
/lgtm

vincepri:
Looking at the other labels, they're all generic and not specific to any provider. Is this something that we actually want?

benmoss:
Not sure what you mean, we already have labels for every other provider: https://github.com/kubernetes-sigs/cluster-api/labels?q=area

benmoss:
Ah, I get it now! The other providers are referring to the in-tree k8s cloud providers. I thought it was strange that we were putting all these CAPI providers in the general labels area but now I realize it's the k8s providers that are leaking into CAPI. Hmm, I guess I am open that maybe we don't need to categorize issues relating to CAPD? I just think it would be handy. Open to alternative label names too.

fabriziopandini (commented):
/lgtm

vincepri:
No strong opinions here, just wanted to point that out, I'll let others chime in

cc @CecileRobertMichon @fabriziopandini

CecileRobertMichon:
No objections to adding a docker label from my side, especially since it's scoped to the CAPI repo and the docker provider is special since it lives in the core repo.

/lgtm

fejta (approved):
/hold

(anyone in cluster-api is free to unhold)

detiber:
I don't see any comments that strongly object to this, so removing the hold
/hold cancel

PR #19632, opened 2020-10-20
Update ci-kubernetes-e2e-gci-gce to latest kubekins

spiffxp:
This is to verify https://github.com/kubernetes/test-infra/pull/19620 doesn't break anything

Next PR will be to use the new --extract-ci-bucket flag

ref: https://github.com/kubernetes/k8s.io/issues/846#issuecomment-705722852oref


hasheddan (approved):
/lgtm

spiffxp:
https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-e2e-gci-gce/1318654183234080768 is using the new image and was able to successfully standup a cluster

Opened https://github.com/kubernetes/test-infra/pull/19634 for the next change

PR #19633, opened 2020-10-20
CAPZ: update CAPI periodic job to use cred-only preset

CecileRobertMichon:
This updates the periodic CAPI + Azure e2e job to match the pull job so it:
1) runs the self hosted spec
2) collects workload cluster logs


mboersma (approved):
lgtm

devigned (approved):

PR #19634, opened 2020-10-20
Make ci-kubernetes-e2e-gci-gce use k8s-release-dev artifacts

spiffxp:
This uses the `--extract-ci-bucket` flag introduced in https://github.com/kubernetes/test-infra/pull/19620

Followup to https://github.com/kubernetes/test-infra/pull/19632

ref: https://github.com/kubernetes/k8s.io/issues/846#issuecomment-705722852oref


hasheddan (approved):
/lgtm

spiffxp:
The job has stayed green since this PR merged and I confirmed by looking at the build log for the most recent run that it's downloading from the new bucket instead of the old (https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-e2e-gci-gce/1318913898769813504)

PR #19635, opened 2020-10-20
Update approval configuration for k/website

sftim:
Match configuration to current, typical review process for SIG Docs.

[k/website](https://github.com/kubernetes/website) uses lgtm label to indicate technical review and approved label to indicate that the content change is acceptable.

@kubernetes/sig-docs-leads FYI

### DIFF
--- a/config/prow/plugins.yaml
+++ b/config/prow/plugins.yaml
@@ -67,6 +67,10 @@ approve:
   require_self_approval: false
   lgtm_acts_as_approve: true
 - repos:
+  - kubernetes/website
+  require_self_approval: true
+  lgtm_acts_as_approve: false
+- repos:
   - kubernetes/enhancements
   - kubernetes/kops
   - kubernetes/kubernetes


config/prow/plugins.yaml
```diff
@@ -66,6 +66,10 @@ approve:
   - kubernetes/utils
   require_self_approval: false
   lgtm_acts_as_approve: true
+- repos:
+  - kubernetes/website
+  require_self_approval: true
+  lgtm_acts_as_approve: false
```
sftim, line 72: - even if you're an approver, someone needs to approve your PR (it's not automatically approved)
- we don't treat _/lgtm_ as approval if you're an approver; that could be surprising - it has caught me out
- approvers can still explicitly /lgtm and /approve the same PR - eg fine for typo fixes

sftim: For context: https://kubernetes.slack.com/archives/C1J0BPD2M/p1603222339171000 & following thread

jimangel:
Thanks Tim!
/lgtm
/approve

PR #19641, opened 2020-10-21
Increase interval to every 3h

BobyMCbobs:
Undo change to 10mins


hh:
/hold
Let's look at this in the morning.

johnbelamaric:
I am missing context here. 10m sounds like WAY to frequently though.

johnbelamaric:
/approve

Leave it to Hippie to remove the hold.

PR #19647, opened 2020-10-21
Deprecate plank v1

alvaroaleman:
Prow controller manager/Plank v2 is in use by Openshift since June and by K8s since September and seems to work fine. We do not want to maintain two implementations long term.

/assign @fejta


petr-muller (approved):
LGTM, this works flawlessly in OpenShift for some time now.

fejta (approved):
*prow controller manager (in PR description)

PR #19656, opened 2020-10-21
Use command: instead of args: for apisnoop-gate entrypoint + arg

hh:
This is a duplicate of #19655 which seems stuck without `cncf-cla`


hh:
I think there's an issue with the job itself:

https://prow.k8s.io/job-history/gs/kubernetes-jenkins/pr-logs/directory/pull-test-infra-bazel

hh:
/test pull-test-infra-bazel

>

PR #19660, opened 2020-10-22
releng: promote ci-kubernetes-build-fast-canary to be the official job

cpanato:
- remove the canary job, the test shows the job is stable running in the new cluster.
- set the `ci-kubernetes-build-fast` job to use the new cluster

fixes: https://github.com/kubernetes/test-infra/issues/19484
fixes: https://github.com/kubernetes/kubernetes/issues/95173

/cc @spiffxp @justaugustus

### DIFF
--- a/config/jobs/kubernetes/sig-release/kubernetes-builds.yaml
+++ b/config/jobs/kubernetes/sig-release/kubernetes-builds.yaml
@@ -69,42 +69,6 @@ periodics:
 
 - interval: 5m
   name: ci-kubernetes-build-fast
-  labels:
-    preset-service-account: "true"
-    preset-dind-enabled: "true"
-  spec:
-    containers:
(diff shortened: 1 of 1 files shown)


justaugustus (changes requested):
Needs

```yaml
      - --release=k8s-release-dev
```

after `--fast` and the testgrid alerts should include release-managers@kubernetes.io for all build jobs.

cpanato:
@justaugustus one question, when do I know we need to use the `k8s-release-dev` bucket ?
will update the PR based on your feedback, thanks!

cpanato:
updated :)

mborsz:
Could you comment on the context of using k8s-release-dev? It seems to break some tests including scalability tests (see https://github.com/kubernetes/test-infra/issues/19838).

Should we revert this change (k8s-release-dev part) or are you planning to migrate existing jobs to k8s-release-dev bucket?

Issue #19662, opened 2020-10-22
pull-test-infra-bazel broken


alvaroaleman:
Related: https://github.com/bazelbuild/continuous-integration/issues/1055

**What happened**:

pull-test-infra-bazel permanently fails with:
```
ERROR: /home/prow/go/src/github.com/kubernetes/test-infra/hack/BUILD.bazel:16:1:  failed (Exit 34): UNAVAILABLE: HTTP/2 error code: NO_ERROR
```

Samples:
* https://github.com/kubernetes/test-infra/pull/19647
* https://github.com/kubernetes/test-infra/pull/19655
* https://github.com/kubernetes/test-infra/pull/19659

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

**Please provide links to example occurrences, if any**:

**Anything else we need to know?**:

rifelpet:
[My PR](https://prow.k8s.io/pr-history/?org=kubernetes&repo=test-infra&pr=19652) was failing with this same issue about 15 hours ago. Retrying it now and it appears to be [stuck](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/test-infra/19652/pull-test-infra-bazel/1319254306397884416).

There are other PRs with [jobs in the same state](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/test-infra/19656/pull-test-infra-bazel/1319239467524952064) as well.

thejoycekung:
Similar issues happening for the `periodic-kubernetes-bazel-build-master` job - kubernetes#95795

fejta:
This looks related to RBE. One option is to reduce paralleism. A second is to turn off RBE. Going to try out both approaches.

spiffxp:
FWIW

**pull-test-infra-bazel** uses RBE (but perhaps without a remote cache)
- first failing job started 2020-10-21 20:39 UTC (https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/test-infra/19651/pull-test-infra-bazel/1319015417548640256)
- pull-test-infra-bazel runs in the default build cluster (there is no `cluster:` field)
- the --config=ci flag here... https://github.com/kubernetes/test-infra/blob/d0a341486174028a3f6267cbad5e4a40d45758dc/config/jobs/kubernetes/test-infra/test-infra-presubmits.yaml#L18
[...]

PR #19664, opened 2020-10-22
Reduce RBE parallelism

fejta:
ref https://github.com/kubernetes/test-infra/issues/19662

This might address these errors


alvaroaleman (approved):

alvaroaleman:
/shrug
wont mege anyways if it doesn't fix the issue

Issue #19666, opened 2020-10-22
Pod-Util BQ job supported flake query


MushuEE:
**What would you like to be added**:
Current [flake queries](https://github.com/kubernetes/test-infra/blob/master/metrics/configs/flakes-config.yaml) exclude pod-utils job recently added to BQ by kettle. The query needs to be updated to support fetching of not only past jobs but new `bootstrap` jobs and new `decorated` jobs. 

**Why is this needed**:
As a follow up to #19592. Without adding this feature there is no visibility into metric on Pod-Util jobs which is the new conical job type (post bootstrap deprecation). 

# Notes + TODO:
The current flake query involves four stages of queries
> From inside create [
>- job,
>- build_consistency (percentage of runs that did not flake),
>- commit_consistency (percentage of commits that did not flake),
>- flakes (times job flaked),
>- runs (times job ran),
>- commits (commits tested),
> -tests (array of flaked tests in the job)]
>> From inside create [
>>- job, 
>>- commit, 
>>- flaked (if not all passed), 
>>- passed, runs (as int), 
>>- tests (ordered by number of failures)]
>>> From inside create [
>>>- job, 
>>>- commit, 
>>>- sum(`SUCCESS` result), 
>>>- count (all result), 
>>>- tests (array of all tests)]
>>>>Select from the week table where `version` (null for decorated) is not 'unknown' AND (ci job OR it was a serial PR job as is defined by `repos`) AND in last 7 days. ![image](https://user-images.githubusercontent.com/7189267/96927902-a27c2300-146c-11eb-9dbe-bdaa87f1ef36.png) Create something of the following:
>>>>- job, 
>>>>- kind(`pull` or `ci`, version (needs to be updated for both version and repo_commit (decorated), 
>>>>- commit (hash pulled from repos or version), 
>>>>- result, 
>>>>- test (tuple of tests)]

  
Some of these fields exist in `decorated` and not in `bootstrap` and vice versa. All need jobs being updated (new kettle)
- [x] TODO: Update Production Kettle now that staging is stable and tested

Because old jobs will not be updated with new fields, the query must support BOTH old schema and new [schema](https://github.com/kubernetes/test-infra/blob/master/kettle/schema.json) with respective fields being updated from the two job types.
- [x] TODO: Breakdown the current query elements and validate that they collect both `decorated` and `bootstrap` jobs and pass them up the query chain as expected. (You might want to pick a job of each type and start with those to validate each step without to much noise from other jobs) 

/assign
[...]

PR #19669, opened 2020-10-23
Update Experimental query for Pod-Util Support

MushuEE:
This (I believe) adds pod utils support for collecting decorated jobs. *Only validated for ci jobs so far.

If you look at comment on #19666 I had found a decorated job that was flaking by the following query:
`SELECT * FROM `k8s-gubernator.build.staging` where repo_commit is not null and result!='SUCCESS'`

Using this I created queries for that specific job:

```
          SELECT
            job,
            if(substr(job, 0, 3) = 'pr:', 'pull', 'ci') kind,  /* pull or ci */
            ifnull(repo_commit, version) version, /* git version, empty for ci  */
            if(substr(job, 0, 3) = 'pr:',
              regexp_extract(
                repos,
                r'[^,]+,\d+:([a-f0-9]+)"'
              ),
              ifnull(repo_commit, version)
            ) commit,  /* repo commit for PR or version for CI */
            result,  /* SUCCESS if the build passed */
            test  /* repeated tuple of tests */
          FROM `k8s-gubernator.build.staging` as t
          where
            datetime(started) > datetime_sub(current_datetime(), interval 7 DAY)
            and (version != 'unknown' or repo_commit	is not null)
            and (
              (substr(job, 0, 3) = 'ci-' and (version != 'unknown' or repo_commit	is not null)) or
              array_length(split(replace(repos,', ', ''), ',')) = 2 /*serial pr jobs only (# of PR refs +1 == 2)*/
            ) and job='ci-test-infra-bazel'
```
This validated that the flake job was getting collected I also ran the original query and validated the diff (see image).

image

/area metrics
/assign @spiffxp 
/area kettle

### DIFF
--- a/metrics/configs/flakes-experiment-config.yaml
+++ b/metrics/configs/flakes-experiment-config.yaml
@@ -58,7 +58,7 @@ query: |
             ifnull(repo_commit, version) version, /* git version, empty for ci  */
             if(substr(job, 0, 3) = 'pr:',
               regexp_extract(
-                repos,
+                ifnull(repos, (select i.value from t.metadata i where i.key = 'repos')),
                 r'[^,]+,\d+:([a-f0-9]+)"'
               ),
               ifnull(repo_commit, version)
@@ -71,7 +71,13 @@ query: |
             and (version != 'unknown' or repo_commit is not null)
             and (
               substr(job, 0, 3) = 'ci-' or
-              array_length(split(replace(repos,', ', ''), ',')) = 2 /*serial pr jobs only (# of PR refs +1 == 2)*/
+              exists(
+                select as struct
+                  i
+                from t.metadata i
+                where (i.key = 'repos' or repos is not null) and
+                array_length(split(replace(ifnull(t.repos, i.value),', ', ''), ',')) = 2 /*serial pr jobs only (# of PR refs +1 == 2)*/
+              )
             )
         )
         group by job, commit
--- a/metrics/configs/flakes-experiment-config.yaml
+++ b/metrics/configs/flakes-experiment-config.yaml
@@ -65,7 +65,7 @@ query: |
             ) commit,  /* repo commit for PR or version for CI */
             result,  /* SUCCESS if the build passed */
             test  /* repeated tuple of tests */
-          FROM `k8s-gubernator.build.week` as t
+          FROM `k8s-gubernator.build.staging` as t
           where
             datetime(started) > datetime_sub(current_datetime(), interval 7 DAY)
             and version != 'unknown'
--- a/metrics/configs/flakes-experiment-config.yaml
+++ b/metrics/configs/flakes-experiment-config.yaml
@@ -55,22 +55,22 @@ query: |
           SELECT
             job,
             if(substr(job, 0, 3) = 'pr:', 'pull', 'ci') kind,  /* pull or ci */
-            ifnull(repo_commit, version) version, /* bootstrap git version, empty for ci  */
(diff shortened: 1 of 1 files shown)


MushuEE:
Also validated for decorated `pr:*` 
image

metrics/configs/flakes-experiment-config.yaml
```diff
@@ -55,22 +55,22 @@ query: |
...
             test  /* repeated tuple of tests */
-          FROM `k8s-gubernator.build.week` as t
+          FROM `k8s-gubernator.build.staging` as t
           where
             datetime(started) > datetime_sub(current_datetime(), interval 7 DAY)
-            and version != 'unknown'
+            and (version != 'unknown' or repo_commit is not null)
             and (
-              (substr(job, 0, 3) = 'ci-' and version != 'unknown') or
+              (substr(job, 0, 3) = 'ci-' and (version != 'unknown' or repo_commit	is not null)) or
```
spiffxp, line 73: nit for followup (since this is also in the original and you're just following what's there): since we check `(version != 'unknown' or repo_commit is not null)` above, I don't think we need to do it again
```suggestion
              (substr(job, 0, 3) = 'ci-' or
```

spiffxp (approved):
Trying to compare this against https://docs.google.com/document/d/139R-4VJjt9_RF2tcNNnH_fi35qQPdCot8CIhaG91HOU/edit#heading=h.gga1mrjainkr I think I'll need to explore kettle changes to confirm if repos is populated correctly, won't hold merge for that

spiffxp:
comparing against: https://docs.google.com/document/d/139R-4VJjt9_RF2tcNNnH_fi35qQPdCot8CIhaG91HOU/edit

>Kettle would populate this using started.json[‘repos’] field, falling back to finished.json[‘metadata’][‘repos’] if empty. This field would thus be populated by both bootstrap and decorated jobs.

kettle does this

> For PR jobs, the query would be updated to use build.repos instead of build.metadata.repos, falling back to build.metadata.repos if the field doesn’t exist.

The query is not doing this, it's using repos by default, which means it won't work against data that's earlier than the kettle change.  I think merging this PR to iterate is fine.  But we should followup to replace repos with something like
```
ifnull(repos, (select i.value from t.metadata i where i.key = 'repos'))
```

MushuEE:
> comparing against: https://docs.google.com/document/d/139R-4VJjt9_RF2tcNNnH_fi35qQPdCot8CIhaG91HOU/edit
> 
> > Kettle would populate this using started.json[‘repos’] field, falling back to finished.json[‘metadata’][‘repos’] if empty. This field would thus be populated by both bootstrap and decorated jobs.
> 
> kettle does this
> 
> > For PR jobs, the query would be updated to use build.repos instead of build.metadata.repos, falling back to build.metadata.repos if the field doesn’t exist.
> 
> The query is not doing this, it's using repos by default, which means it won't work against data that's earlier than the kettle change. I think merging this PR to iterate is fine. But we should followup to replace repos with something like
> 
> ```
> ifnull(repos, (select i.value from t.metadata i where i.key = 'repos'))
> ```

Yes this is the last thing to solve and is the ugly bit. I will add it in this PR 
/hold

MushuEE:
Done
/hold cancel

MushuEE:
This will enable #14643 to be closed

e-blackwelder (approved):
Note: Only reviewed changes in `Eye bleed backwards query support - cb4fd04`. (spiffxp approved the previous changes.)

MushuEE:
@spiffxp mind re-lgtm?

PR #19675, opened 2020-10-23
Move log dump script and logexporter config out of kubekins directory

tosi3k:
Conduct a small cleanup of the kubekins-e2e directory. This way, changes to log dumping script/logexporter won't need `kubekins-e2e` owners engagement.

/sig scalability
/assign @wojtek-t @jkaniuk @BenTheElder


images/kubekins-e2e/Dockerfile
```diff
@@ -98,8 +98,8 @@ ENV IMAGE=${IMAGE_ARG}
 ADD ["kops-e2e-runner.sh", \
     "kubetest", \
     "https://raw.githubusercontent.com/kubernetes/kubernetes/master/cluster/get-kube.sh", \
-    "log-dump.sh", \
-    "logexporter-daemonset.yaml", \
+    "https://raw.githubusercontent.com/kubernetes/test-infra/master/logexporter/cluster/log-dump.sh", \
+    "https://raw.githubusercontent.com/kubernetes/test-infra/master/logexporter/cluster/logexporter-daemonset.yaml", \
```
BenTheElder, line 102: please don't do that (live pulling from HEAD via magical ADD, we need to stop doing that). 
if your only goal is a separate set of approvers, you can instead add a subdirectory for logdup with OWNERS

jkaniuk: The goal here is to move log-dump.sh to a more generic place, as it should not be tied to kubekins.
How to do that with Dockerfile?

jkaniuk: discussed offline

Benjamin Elder:
> the actual build logic is a cloudbuild https://github.com/kubernetes/test-infra/blob/master/images/kubekins-e2e/cloudbuild.yaml
> so eg kubetest itself is not in that directory
> we could copy it over, or we could change the build context for the dockerfile to be the repo scope
some other images are built from the repo root even though the dockerfile is not there

changing repo root seems easier

tosi3k: Changed the repo root accordingly.
I tested `kubekins-e2e` image built on top of the current change by running a simple Prow job leveraging that image and it works as expected.

jkaniuk: Great!

jkaniuk:
@BenTheElder - PTAL

PR #19659, opened 2020-10-22
CAPZ: add anonymous pull to PR E2E job

CecileRobertMichon:
This is needed for https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/974 to enable the private cluster management cluster to pull the capz manager image.

/assign @devigned

PR #19678, opened 2020-10-23
CAPZ: set LOCAL_ONLY to false for PR e2e

CecileRobertMichon:
This should have been part of #19659 

As a side note, this will cause all e2e to pull images instead of using the local versions on kind, maybe we should look into making it more granular and only pull from acr when needed to save time?

/assign @devigned


devigned (approved):
/lgtm

PR #19680, opened 2020-10-24
Show commit  of current building jobs

### LINKED ISSUE #19232: Prow's PR history page doesn't properly identify commit until job completes.
**What happened**:
Job runs seem to always appear in the "Unknown" column until the job completes.
Screen Shot 2020-09-15 at 5 22 01 PM

Jobs should be listed under the correct column while they are still running.

1. Find a running presubmit job: https://prow.k8s.io/?type=presubmit&state=pending
1. Then click the "PR History" button and notice the job(s) in the "Unknown" column.

Any examples I link won't be useful once the job completes. Use the reproduction steps above to find a live example.

/help
/good-first-issue
/area prow prow/deck

vicsufer:
Fix #19232

### DIFF
--- a/prow/cmd/deck/job_history.go
+++ b/prow/cmd/deck/job_history.go
@@ -345,6 +345,7 @@ func getBuildData(ctx context.Context, bucket storageBucket, dir string) (buildD
 	if err != nil {
 		return b, fmt.Errorf("failed to read started.json: %v", err)
 	}
+	b.Started = time.Unix(started.Timestamp, 0)
 	finished := gcs.Finished{}
 	err = readJSON(ctx, bucket, path.Join(dir, prowv1.FinishedStatusFile), &finished)
 	if err != nil {
--- a/prow/cmd/deck/job_history.go
+++ b/prow/cmd/deck/job_history.go
@@ -345,16 +345,23 @@ func getBuildData(ctx context.Context, bucket storageBucket, dir string) (buildD
 	if err != nil {
 		return b, fmt.Errorf("failed to read started.json: %v", err)
 	}
-	b.Started = time.Unix(started.Timestamp, 0)
-	if commitHash, err := getPullCommitHash(started.Pull); err == nil {
-		b.commitHash = commitHash
-	}
 	finished := gcs.Finished{}
 	err = readJSON(ctx, bucket, path.Join(dir, prowv1.FinishedStatusFile), &finished)
 	if err != nil {
 		b.Result = "Pending"
+		for _, ref := range started.Repos { 
+			if strings.Contains(ref, "," + started.Pull + ":"){
+				started.Pull = ref
+				break;
+			}
+		}
 		logrus.Debugf("failed to read finished.json (job might be unfinished): %v", err)
 	}
+
+	if commitHash, err := getPullCommitHash(started.Pull); err == nil {
+		b.commitHash = commitHash
+	}
+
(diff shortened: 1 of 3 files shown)


prow/cmd/deck/job_history.go
```diff
@@ -346,15 +346,23 @@ func getBuildData(ctx context.Context, bucket storageBucket, dir string) (buildD
...
-	if commitHash, err := getPullCommitHash(started.Pull); err == nil {
-		b.commitHash = commitHash
-	}
 	finished := gcs.Finished{}
 	err = readJSON(ctx, bucket, path.Join(dir, prowv1.FinishedStatusFile), &finished)
 	if err != nil {
 		b.Result = "Pending"
+		for _, ref := range started.Repos { 
+			if strings.Contains(ref, "," + started.Pull + ":"){
+				started.Pull = ref
```
vicsufer, line 355: Im not completely happy with this approch, `started.Pull `field only has the PR number when the build is in pending state thus I decided to the Repos field from started.json.

prow/cmd/deck/job_history.go
```diff
@@ -346,15 +346,23 @@ func getBuildData(ctx context.Context, bucket storageBucket, dir string) (buildD
...
+				break;
+			}
+		}
 		logrus.Debugf("failed to read finished.json (job might be unfinished): %v", err)
 	}
+
+	if commitHash, err := getPullCommitHash(started.Pull); err == nil {
```
vicsufer, line 362: If I do not update the started.Pull field with the found reference, using solely the PR Number this method fails to commit and thus sets build commitHash to "Unknown"

prow/cmd/deck/job_history.go
MushuEE: Here is a reference to the conical source of truth for these fields. I would make sure these fields stay consistent.

prow/cmd/deck/pr_history.go
```diff
@@ -39,7 +39,7 @@ import (
 	"k8s.io/test-infra/prow/pod-utils/downwardapi"
 )
 
-var pullCommitRe = regexp.MustCompile(`^[-\w]+:\w{40},\d+:(\w{40})$`)
+var pullCommitRe = regexp.MustCompile(`^[-\.\w]+:\w{40},\d+:(\w{40})$`)
```
vicsufer, line 42: Fixed error when branch names have dot in its name.

prow/cmd/deck/job_history.go
vicsufer: Thanks for the heads up with that file @MushuEE , I've reviewed it and the approach seems to be consistent with it, that why I iterate the repository to find the reference instead of ussing the first one, just in case the order changes somehow.

vicsufer:
Hey @fejta, @clarketm!
Did you have the time to review this? 
Thanks!

fejta (commented):
/ok-to-test

fejta:
Thanks for the ping. This is super cool, great contribution. Can we extend the unit test coverage? Otherwise things LGTM

prow/cmd/deck/pr_history.go
vicsufer: Sure!  I'll try to extend overall coverage.

fejta (approved):

prow/cmd/deck/job_history.go
fejta: Would also to see unit test coverage for this scenario

Issue #19682, opened 2020-10-26
Kettle new xml parse traceback on 'testsuite" tag


MushuEE:
**What happened**:
Prod Kettle 10/25: 
```
ERROR:root:error on gs://kubernetes-jenkins/pr-logs/pull/81478/pull-kubernetes-e2e-gce-rbe/1162056366320259075
Traceback (most recent call last):
  File "make_json.py", line 282, in make_rows
    yield rowid, row_for_build(path, started, finished, results)
  File "make_json.py", line 222, in row_for_build
    for test in parse_junit(result):
  File "make_json.py", line 187, in parse_junit
    name = child.attrib['name']
KeyError: 'name'
```
**What you expected to happen**:
No traceback on this form of xml. When I try to access this data I get:
```
https://prow.k8s.io/job-history/gs/kubernetes-jenkins/pr-logs/directory/81478
```
>failed to get job history: failed to locate build data: failed to read pr-logs/directory/81478/latest-build.txt: creating reader for object pr-logs/directory/81478/latest-build.txt: storage: object doesn't exist

Seems that job no longer exists. This causes kettle to hang.
**How to reproduce it (as minimally and precisely as possible)**:

**Please provide links to example occurrences, if any**:

**Anything else we need to know?**:
/area kettle
/assign

spiffxp:
Maybe related: I see nothing in testgrid for this job? https://testgrid.k8s.io/presubmits-kubernetes-nonblocking#pull-kubernetes-e2e-gce-rbe

MushuEE:
> Maybe related: I see nothing in testgrid for this job? https://testgrid.k8s.io/presubmits-kubernetes-nonblocking#pull-kubernetes-e2e-gce-rbe

Seems like I can't even find that job in prow anymore.

MushuEE:
Maybe after https://github.com/kubernetes/test-infra/pull/18027, they stopped running and the results have timed out? @BenTheElder

spiffxp:
I think the testgrid issue is related but orthogonal, turns out there are more jobs affected, so I opened https://github.com/kubernetes/test-infra/issues/19696

PR #19687, opened 2020-10-26
Kettle keeps hitting "malformed" xml 

MushuEE:
Kettle parses xml from GCS buckets, there is no set standard for that XML format and some jobs seem to produce xml that lacks important information: Ex. Test and Test Suite name. 

Instead of trying to handle this and confusing the situation with unknown tests names, I prefer to log an error on the bucket and skip the parsing as to avoid confusing half results.

#19682 
/area Kettle


spiffxp:
Failing verify-pylint
```
==================== Test output for //hack:verify-pylint:
Linting python...
************* Module kettle.make_json
kettle/make_json.py:204:8: W1202: Use % formatting in logging functions and pass the % parameters as arguments (logging-format-interpolation)
```

kettle/make_json_test.py
```diff
@@ -634,27 +634,15 @@ class ParseJsonTest(unittest.TestCase):
...
-           'time': 0.0}],
+         [],
         ),
         ('Failure as Message',
          'kettle/testdata/failure_message.xml',
-         [{'failed': True,
-           'failure_text': 'simple_test.go:186: Running tests in file plumbing',
-           'name': 'unknown TestSimpleFile',
-           'time': 0.0}],
+         [],
```
spiffxp, line 641: I would update testdata to be valid here, give testsuite a name; this test case shouldn't be exercising the no-suite-name case, that's done above

kettle/make_json_test.py
```diff
@@ -634,27 +634,15 @@ class ParseJsonTest(unittest.TestCase):
...
         ),
         ('Failure as Text',
          'kettle/testdata/failure_text.xml',
-         [{'failed': True,
-           'failure_text': 'simple_test.go:186: Running tests in file plumbing',
-           'name': 'unknown TestSimpleFile',
-           'time': 0.0}],
+         [],
```
spiffxp, line 645: same here

kettle/make_json.py
```diff
@@ -182,24 +181,28 @@ def parse_result(child_node):
...
-            if skipped:
-                continue
-            yield make_result(name, time, failure_text)
-    elif tree.tag == 'testsuites':
-        for testsuite in tree:
-            suite_name = testsuite.attrib.get('name', 'unknown')
-            for child in testsuite.findall('testcase'):
-                name = '%s %s' % (suite_name, child.attrib['name'])
+    try:
+        if tree.tag == 'testsuite':
```
spiffxp, line 185: The name attribute validation needs to be here instead of within the `testsuites` case if you want to be as strict as [the junit5 xsd](https://github.com/junit-team/junit5/blob/main/platform-tests/src/test/resources/jenkins-junit.xsd) which has `name` as a required attribute

My read of what is in this PR is:
- if I pass in a bare `<testsuite>` I don't need  `name` attribute (but it's worth noting [testgrid's junit parsing](https://github.com/GoogleCloudPlatform/testgrid/blob/481a3cbf271b6b36e32113024c6b7c3162bd66c2/metadata/junit/junit.go) assumes `<testsuites>` as the root element)
- if I pass in a `<testsuites><testsuite>...` I do need a `name` attribute on the testsuite

spiffxp (commented):
Part of me wants to be more forgiving and opt for a `.get('name', 'key_is_missing')` approach for missing keys, but I agree being more strict may reduce support surface area.

kettle/make_json_test.py
MushuEE: I am fine with making it more robust, I opted for fail out because it made it a little difficult to track down failures with an unknown back reference, but within a job it seems there are typically very few tests with the same name across test suites. I will update, thanks.

kettle/make_json.py
MushuEE: I changed it to always make a try for name and default to unspecified.

spiffxp (approved):
/approve
/lgtm

Issue #19607, opened 2020-10-16
Verify-owners plugin does not remove the do-not-merge label if PR is updated to remove all OWNERS files.


cjwagner:
**What happened**:
1. A PR included the creation of some invalid OWNERS files (approvers who were not members).
1. The [verify-owners](https://github.com/kubernetes/test-infra/tree/master/prow/plugins/verify-owners) plugin added the `do-not-merge/invalid-OWNERS-file`
1. The PR was updated to remove all the OWNERS files.
1. The verify-owners plugin did not remove the do-not-merge label because there were no OWNERS file changes in the PR so it never ran validation.

**What you expected to happen**:
PRs that have the `do-not-merge/invalid-OWNERS-file` label should always be validated by the verify-owners plugin when they change even if they no longer have OWNERS file changes so that we can remove the do-not-merge label when needed.

**Please provide links to example occurrences, if any**:
https://github.com/GoogleCloudPlatform/oss-test-infra/pull/554#issuecomment-710343565

/help
/good-first-issue

matthyx:
Oh wow, never thought about that one...
@joostas let me know if you need a hand at this one!

alvaroaleman:
Fixed in https://github.com/kubernetes/test-infra/pull/19691
/close

PR #19691, opened 2020-10-26
Verify owners remove label if OWNERS file was removed

joostas:
Ref. #19607 
Verify owners plugin: remove label if OWNERS file was removed.
Initial implementation and tests

/kind bug
/assign @matthyx

### DIFF
--- a/prow/plugins/verify-owners/verify-owners.go
+++ b/prow/plugins/verify-owners/verify-owners.go
@@ -205,13 +205,9 @@ func handle(ghc githubClient, gc git.ClientFactory, roc repoownersClient, log *l
 
 	// List modified OWNERS files.
 	var modifiedOwnersFiles []github.PullRequestChange
-	var ownerModified bool
 	for _, change := range changes {
-		if filepath.Base(change.Filename) == ownersFileName {
-			ownerModified = true
-			if change.Status != github.PullRequestFileRemoved {
-				modifiedOwnersFiles = append(modifiedOwnersFiles, change)
-			}
+		if filepath.Base(change.Filename) == ownersFileName && change.Status != github.PullRequestFileRemoved {
+			modifiedOwnersFiles = append(modifiedOwnersFiles, change)
 		}
 	}
 
@@ -226,7 +222,13 @@ func handle(ghc githubClient, gc git.ClientFactory, roc repoownersClient, log *l
 		}
 	}
 
-	if !ownerModified && !ownerAliasesModified {
+	issueLabels, err := ghc.GetIssueLabels(org, repo, number)
+	if err != nil {
+		return err
+	}
+	hasInvalidOwnersLabel := github.HasLabel(labels.InvalidOwners, issueLabels)
+
+	if len(modifiedOwnersFiles) == 0 && !ownerAliasesModified && !hasInvalidOwnersLabel {
 		return nil
 	}
 
@@ -281,13 +283,6 @@ func handle(ghc githubClient, gc git.ClientFactory, roc repoownersClient, log *l
 			}
 		}
 	}
-
-	// React if there are files with incorrect configs or non-trusted users.
-	issueLabels, err := ghc.GetIssueLabels(org, repo, number)
-	if err != nil {
-		return err
-	}
-	hasInvalidOwnersLabel := github.HasLabel(labels.InvalidOwners, issueLabels)
 
 	if len(wrongOwnersFiles) > 0 {
 		s := "s"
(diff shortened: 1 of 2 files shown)


joostas:
I have signed CLA.

matthyx:
@joostas can you amend your commit to change the user as described here:
https://docs.github.com/en/free-pro-team@latest/github/committing-changes-to-your-project/why-are-my-commits-linked-to-the-wrong-user

prow/plugins/verify-owners/verify-owners.go
```diff
@@ -222,10 +222,6 @@ func handle(ghc githubClient, gc git.ClientFactory, roc repoownersClient, log *l
 		}
 	}
 
-	if len(modifiedOwnersFiles) == 0 && !ownerAliasesModified {
```
matthyx, line 225: if you remove this condition, then you will clone the repo for every PR...

matthyx: it would be better to add a condition to still verify owners whenever `ownersFileName` or `ownersAliasesFileName` have a `change.Status` of `github.PullRequestFileRemoved`

joostas: Hi @matthyx. Thanks for your quick response!
Good point about repo cloning, Returned back that short-circuiting with an additional check if there were any owner files modifications.

prow/plugins/verify-owners/verify-owners_test.go
```diff
@@ -142,7 +142,7 @@ var ownerAliasesFiles = map[string][]byte{
 `),
 }
 
-func IssueLabelsAddedContain(arr []string, str string) bool {
+func IssueLabelsContain(arr []string, str string) bool {
```
matthyx, line 145: I don't get why you changed the name of that func?

matthyx:
@joostas also please squash both commits

prow/plugins/verify-owners/verify-owners_test.go
joostas: Previously in tests, it worked only with the collection `fghc.IssueLabelsAdded`, now I added test which works with `fghc.IssueLabelsRemoved` therefore I decided maybe more generic name `IssueLabelsContain` would fit better here.

matthyx:
perfect, thanks!
/lgtm
/assign @fejta

joostas:
Updated tests

prow/plugins/verify-owners/verify-owners.go
```diff
@@ -205,9 +205,13 @@ func handle(ghc githubClient, gc git.ClientFactory, roc repoownersClient, log *l
 
 	// List modified OWNERS files.
 	var modifiedOwnersFiles []github.PullRequestChange
+	var ownerModified bool
 	for _, change := range changes {
-		if filepath.Base(change.Filename) == ownersFileName && change.Status != github.PullRequestFileRemoved {
-			modifiedOwnersFiles = append(modifiedOwnersFiles, change)
+		if filepath.Base(change.Filename) == ownersFileName {
```
alvaroaleman, line 210: This doesn't change anything about he behavior of the plugin and does not address the issue mentioned in https://github.com/kubernetes/test-infra/issues/19607

The issue mentioned in https://github.com/kubernetes/test-infra/issues/19607 is that a PR at some point changed an OWNERS file and that change was invalid, the invalid owners label gets added. Later on the PR gets updated to not include any OWNERS file at all anymore, but the plugin does not remove that label.

What we need here is:
* A check if the invalid owners file exists
* Code to remove it in the `len(modifiedOwnersFiles) == 0 && !ownerAliasesModified {` condition if it exists

joostas: Ah, ok. I interpreted this task in the wrong direction. Thanks for pointing out.
Now I updated code, to only short circuit if PR does not have invalid `do-not-merge/invalid-OWNERS-file` label.
```go
if len(modifiedOwnersFiles) == 0 && !ownerAliasesModified && !hasInvalidOwnersLabel {
	return nil
}
```
Decided don't duplicate the label removal code. Hope this will be correct solution.

prow/plugins/verify-owners/verify-owners_test.go
```diff
@@ -1255,13 +1255,153 @@ func testHandleGenericComment(clients localgit.Clients, t *testing.T) {
...
+	if err := addFilesToRepo(lg, []string{"OWNERS"}, "valid"); err != nil {
+		t.Fatalf("Adding base commit: %v", err)
+	}
+
+	if err := addFilesToRepo(lg, []string{"OWNERS_ALIASES"}, "collaborators"); err != nil {
+		t.Fatalf("Adding base commit: %v", err)
+	}
+
+	for i, test := range tests {
+		pr := i + 1
```
alvaroaleman, line 1320: Can you please make these subtests:

```
for _, test := range tests {
  t.Run(test.name, func(t  *testing.T){
  ....
}
}
```

Other than that PR lgtm

alvaroaleman (approved):

PR #19695, opened 2020-10-26
Add instruction for updating vendor of monitoring stack

chaodaiG:
Was only able to find the instructions from https://github.com/kubernetes/test-infra/pull/19004, put it in README so that it's better discoverable

/assign @spiffxp


spiffxp (approved):
Thanks, I swore I had documented this but apparently only in the git commit where I needed to do this.  I wish this was a make target or automated instead, but this is better than before

/approve
/lgtm

PR #19702, opened 2020-10-27
Genyaml: Add PopulateStruct func to autopopulate structs

alvaroaleman:
Currently, genyaml must be fed with a struct whose pointer fields are
manually populated with data, otherwise no documentation will be
generated for them. This is very cumbersome for more complicated types.

This PR adds a PopulateStruct function to genyaml that will do this at
runtime via reflection.

/assign @clarketm

### DIFF
--- a/prow/config/prow-config-documented.yaml
+++ b/prow/config/prow-config-documented.yaml
@@ -1,0 +1,575 @@
+branch-protection:
+    allow_deletions: false
+    allow_force_pushes: false
+    enforce_admins: false
+    exclude:
+      - ""
+    orgs:
+        "":
+            allow_deletions: false
+            allow_force_pushes: false
+            enforce_admins: false
+            exclude:
+              - ""
+            protect: false
+            repos:
+                "":
+                    allow_deletions: false
+                    allow_force_pushes: false
+                    branches:
+                        "":
+                            allow_deletions: false
+                            allow_force_pushes: false
+                            enforce_admins: false
+                            exclude:
+                              - ""
+                            protect: false
+                            required_linear_history: false
+                            required_pull_request_reviews:
+                                dismiss_stale_reviews: false
+                                dismissal_restrictions:
+                                    teams:
+                                      - ""
+                                    users:
+                                      - ""
+                                require_code_owner_reviews: false
+                                required_approving_review_count: 0
+                            required_status_checks:
+                                contexts:
+                                  - ""
+                                strict: false
+                            restrictions:
+                                teams:
+                                  - ""
+                                users:
+                                  - ""
+                    enforce_admins: false
+                    exclude:
+                      - ""
+                    protect: false
+                    required_linear_history: false
+                    required_pull_request_reviews:
+                        dismiss_stale_reviews: false
+                        dismissal_restrictions:
+                            teams:
+                              - ""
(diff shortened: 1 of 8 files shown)


prow/config/testdata/prow-config-documented.yaml
```diff
@@ -0,0 +1,575 @@
+branch-protection:
```
alvaroaleman, line 1: Something is buggy here in genyaml, this does have comments in the source. It works when directly generating for the `BranchProtection` type. I would like to defer looking into this to a follow-up though.

prow/config/prow-config-documented.yaml
```diff
@@ -0,0 +1,575 @@
+branch-protection:
```
alvaroaleman, line 1: Something is buggy here in genyaml, this does have comments in the source. It works when directly generating for the `BranchProtection` type. I would like to defer looking into this to a follow-up though as whatever it doesn't get introduced by this PR.

prow/config/prow-config-documented.yaml
```diff
@@ -0,0 +1,575 @@
...
+
+    # HiddenRepos is a list of orgs and/or repos that should not be displayed by Deck.
+    hidden_repos:
+      - ""
+
+    # Deprecated: RerunAuthConfig specifies who is able to trigger job reruns if that feature is enabled.
+    # The permissions here apply to all jobs.
+    # This option will be removed in favor of RerunAuthConfigs in July 2020.
+    rerun_auth_config:
+        github_orgs:
```
stevekuznetsov, line 153: Are these not commented in the source?

prow/config/prow-config-documented.yaml
```diff
@@ -0,0 +1,575 @@
...
+        "": null
+plank:
+    # DefaultDecorationConfigs holds the default decoration config for specific values.
+    # This config will be used on each Presubmit and Postsubmit's corresponding org/repo, and on Periodics
+    # if extraRefs[0] exists.
+    # Use `org/repo`, `org` or `*` as a key.
+    default_decoration_configs:
+        "":
+            cookiefile_secret: ' '
+            gcs_configuration:
```
stevekuznetsov, line 352: I know these have comments!

prow/config/prow-config-documented.yaml
alvaroaleman: Yeah, see https://github.com/kubernetes/test-infra/pull/19702#discussion_r512399362

The basic issue is that genyaml considers each file to be its own package which means that if you have a struct in file a that references another struct in file b it won't work. Ref: https://github.com/kubernetes/test-infra/blob/30c5d23c40dbf10ce2b4ed9232a6199c385923b1/pkg/genyaml/genyaml.go#L154

Thats a genyaml bug though that wasn't introduced by this PR, so I'd like to fix it in a follow-up.

stevekuznetsov:
Results and tests LGTM, I am not great at reviewing reflect code. Since it's not supposed to be in any critical path and users know to defer a panic recovery, seems reasonable.

/lgtm
/hold
/approve

alvaroaleman:
Cancelling the hold, the code added here is only used in a test so very unlikely to cause harm
/hold cancel

PR #19705, opened 2020-10-27
add periodic jobs for dual stack

aojea:
Now the the dual-stack PR merged we need to be sure we don't regress adding periodic jobs

### DIFF
--- a/config/jobs/kubernetes/sig-network/sig-network-kind.yaml
+++ b/config/jobs/kubernetes/sig-network/sig-network-kind.yaml
@@ -48,7 +48,7 @@ presubmits:
             cpu: 2000m
     annotations:
       testgrid-dashboards: sig-network-kind
-      testgrid-tab-name: sig-network-kind, dual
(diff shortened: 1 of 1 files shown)


aojea:
/assign @MrHohn 

@aramase can you please do the same for the Azure ones that you added as presubmit?
 also add a new regex to run conformance test

aramase:
> @aramase can you please do the same for the Azure ones that you added as presubmit?
also add a new regex to run conformance test

@aojea https://github.com/kubernetes/test-infra/pull/19720 adds the periodic for iptables mode and enables conformance testing too.

aojea:
kindly begging for a lgtm @bowei @thockin  :)

MrHohn (approved):
Sorry for the delay. Feel free to unhold if you don't need an additional LGTM.
/lgtm
/hold

aojea:
/hold cancel
no worries @MrHohn an many thanks.
The rush is just because  that I don't want that some new code breaks dualstack and we don't notice it, bear in mind we are not running any dual stack presubmit job

PR #19711, opened 2020-10-27
prow/bugzilla: allow specifying the tracker id

stevekuznetsov:
prow/bugzilla: use cmp.Diff in tests

The previous diff library is deprecated and this one creates better
output anyway.

---

prow/bugzilla: allow specifying the tracker id

While the URL of the external tracker is meant to be a convenient,
unique identifier for adding external bugs, it may be the case that more
than one external tracker bug types are configured on the Bugzilla
server that track against one server. In this case, the URL is no longer
unique and the internal unique identifier must be used. Normal API keys
may not have the privileges necessary to list all possible trackers, so
we expect the user to simply provide the identifier at startup.

---

/assign @alvaroaleman

### DIFF
--- a/prow/flagutil/bugzilla.go
+++ b/prow/flagutil/bugzilla.go
@@ -29,13 +29,15 @@ import (
 
 // BugzillaOptions holds options for interacting with Bugzilla.
 type BugzillaOptions struct {
-	endpoint   string
-	ApiKeyPath string
+	endpoint                string
+	githubExternalTrackerId uint
+	ApiKeyPath              string
 }
 
 // AddFlags injects Bugzilla options into the given FlagSet.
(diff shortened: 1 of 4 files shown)


prow/flagutil/bugzilla.go
```diff
@@ -29,13 +29,15 @@ import (
 
 // BugzillaOptions holds options for interacting with Bugzilla.
 type BugzillaOptions struct {
-	endpoint   string
-	ApiKeyPath string
+	endpoint                string
+	githubExternalTrackerId int
```
alvaroaleman, line 33: use an `uint` and safe yourself the check

alvaroaleman (approved):

PR #19716, opened 2020-10-27
Adding make install to service-apis config

robscott:
This is intended to verify that examples match CRDs.


robscott:
Need to rework this.

/hold

robscott:
Ended up solving this in the repo with https://github.com/kubernetes-sigs/service-apis/pull/428.

/close

PR #19726, opened 2020-10-28
Cleanup experimental jobs

wojtek-t:
Cleanup 3 experimental scalability jobs:
- ci-kubernetes-e2e-scalability-experimental-load - supposed to validate merging density with load, which has already been checked in to the main config and enabled in most jobs 
- ci-kubernetes-e2e-scalability-experimental-load-containerd - containerd is already a default
- ci-kubernetes-e2e-gci-gce-scalability-containerd-correctness - containerd is already a default

Also cleanup: enable_prometheus_api_responsiveness.yaml as the value of this env is no longer used anywhere.

/assign @jkaniuk @mborsz


jkaniuk:
Please remove ENABLE_PROMETHEUS_API_RESPONSIVENESS from all places then.

jkaniuk:
Right, ENABLE_PROMETHEUS_API_RESPONSIVENESS is in another repo, so that would be a follow-up PR

/lgtm

PR #19729, opened 2020-10-28
Use 50 load test throughput in 5k scale test

mborsz:
This change starts using 50 load test throughput (5x increase) and 30 deletion test throughput (3x increase) in our 5k master load test.

The expected latency growth is approx. 2x on simple metrics which is still within our SLO.

Ref https://github.com/kubernetes/perf-tests/issues/561

/assign @wojtek-t


wojtek-t:
/lgtm

I'm waiting to see our 5k node tests taking 4 hours or so after this change...

PR #19730, opened 2020-10-28
Revert "Tide: Add checkrun support"

alvaroaleman:
This reverts commit 1b5f13772461c1200b87c050a91f97f1daa934bd.

This change increases the V4 token usage in Openshift from about 1.7k
tokens per hour to about 4.6k.

Below a graph of the token usage, you can clearly see when we updated tide and when we reverted the update:

v4


alvaroaleman:
FYI @chizhg as you were interested in the initial change

stevekuznetsov:
:(

PR #19731, opened 2020-10-28
Add deletion annotation handling for job forking

tosi3k:
This would come in handy when one has a master branch-specific preset that is not desired to be copied when a job is forked to the past releases.

/sig scalability
/assign @jkaniuk

### DIFF
--- a/releng/config-forker/main.go
+++ b/releng/config-forker/main.go
@@ -39,6 +39,7 @@ const (
 	periodicIntervalAnnotation   = "fork-per-release-periodic-interval"
 	cronAnnotation               = "fork-per-release-cron"
 	replacementAnnotation        = "fork-per-release-replacements"
+	deletionAnnotation           = "fork-per-release-deletions"
 	testgridDashboardsAnnotation = "testgrid-dashboards"
 	testgridTabNameAnnotation    = "testgrid-tab-name"
 	descriptionAnnotation        = "description"
@@ -154,6 +155,7 @@ func generatePeriodics(conf config.JobConfig, version string) ([]config.Periodic
 		if err != nil {
 			return nil, fmt.Errorf("%s: %v", periodic.Name, err)
 		}
+		p.Labels = performDeletion(p.Labels, p.Annotations[deletionAnnotation])
 		p.Annotations = cleanAnnotations(fixTestgridAnnotations(p.Annotations, version, false))
 		newPeriodics = append(newPeriodics, p)
 	}
@@ -163,7 +165,7 @@ func generatePeriodics(conf config.JobConfig, version string) ([]config.Periodic
 func cleanAnnotations(annotations map[string]string) map[string]string {
 	result := map[string]string{}
 	for k, v := range annotations {
-		if k == forkAnnotation || k == replacementAnnotation {
+		if k == forkAnnotation || k == replacementAnnotation || k == deletionAnnotation {
 			continue
 		}
 		if k == periodicIntervalAnnotation && v == "" {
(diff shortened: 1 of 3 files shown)


releng/config-forker/main.go
```diff
@@ -39,6 +39,7 @@ const (
 	periodicIntervalAnnotation   = "fork-per-release-periodic-interval"
 	cronAnnotation               = "fork-per-release-cron"
 	replacementAnnotation        = "fork-per-release-replacements"
+	deletionAnnotation           = "fork-per-release-deletions"
```
hasheddan, line 42: nitpick: what would you think about calling this something like `fork-per-release-remove-labels`? It took me a moment to determine exactly what this was doing (could just be me :sweat_smile:). We could also consider building this in with `fork-per-release-replacements` by defining a new operator other than `->` (maybe `X`), but I tend to like what you have added here as a simpler solution :+1:

tosi3k: >what would you think about calling this something like fork-per-release-remove-labels

I imagine one would like to use `fork-per-release-deletions` in some scenarios other than just labels in the future. Currently the need is just for the labels but somebody might need this for other job fields as well (`fork-per-release-replacements` already performs substitutions on the container args and tags) so limiting it to just the labels might result in some maintenance problems in the future. OTOH I do not have a strong opinion here :).

>We could also consider building this in with fork-per-release-replacements by defining a new operator other than -> (maybe X)

Thought about it as well, but that was a bit more complicated and went for the simpler solution.

Thanks for the prompt peek into that!

tosi3k:
@hasheddan do we want someone else to take a look at that or are we good to go here?

hasheddan (approved):
I think we are good to go, feel free to tag me on any PRs that introduce this functionality (i.e. add the deletion annotation) in sig-release jobs :+1: 

/lgtm

tpepper (changes requested):
Generally looks good to me.  It would be beneficial though to amend the commit message to have a body describing when/why scenarios for performing deletions.  This would also be good in the README.md, which only implicitly covers the forking.  Why we have "fork-per-release" isn't in that doc and should be.  With it there then it's more natural to additionally discuss the types of mutations such as deletion which happen after forking.  This is almost a nit, but it is actually very important for maintainability to establish architectural reasoning.

@hasheddan would it make sense to go ahead and merge this (with a commit message amendment) and separately update the readme with the high level purpose of forking and expected mutation scenarios?  The https://github.com/kubernetes/test-infra/tree/master/releng/config-forker#actions-taken is almost pseudocode details of what the forker does.  But is somebody trying to use or review the forker in the context of their branches or their test variations going to understand how they should make their configs be well forkable or what replacements and deletions are commonly needed during forkings?

tosi3k:
> Generally looks good to me. It would be beneficial though to amend the commit message to have a body describing when/why scenarios for performing deletions. This would also be good in the README.md, which only implicitly covers the forking. Why we have "fork-per-release" isn't in that doc and should be. With it there then it's more natural to additionally discuss the types of mutations such as deletion which happen after forking. This is almost a nit, but it is actually very important for maintainability to establish architectural reasoning.

Good point, thanks. I made the entry in README.md about the deletions a bit more verbose and amended the commit message accordingly (its body is almost a copy-paste from the README file entry).

tosi3k:
@hasheddan I think that @tpepper is expecting an answer from you in https://github.com/kubernetes/test-infra/pull/19731#pullrequestreview-519913046 :) 

If we're good with the current change, could we get the missing "lgtm" label here?

tosi3k:
@hasheddan friendly ping :).

hasheddan:
@tosi3k sorry for the delay here! This looks good to me now, thanks :)

/lgtm

PR #19734, opened 2020-10-28
sig-storage images: periodically rebuild csi-driver-host-path

pohly:
We are using the "canary" images from the staging repo for some tests.
If we don't push changes into a repo for a while, there's a risk that
the existing canary image from the post-submit job becomes so old that
it gets removed.

Periodically rebuilding the image avoids that. It also ensures that we
detect issues like not being compatible with the cloud build service
before we really need it.


pohly:
I *think* that this might work, but I have no way of testing this other than getting the PR merged...

config/jobs/image-pushing/k8s-staging-sig-storage.yaml
```diff
@@ -19,7 +19,7 @@ postsubmits:
       spec:
         serviceAccountName: gcb-builder
         containers:
-          - image: gcr.io/k8s-testimages/image-builder:v20200901-ab141a0
+          - image: gcr.io/k8s-testimages/image-builder:v20200612-cd781f9
```
fejta, line 22: Are you intentionally downgrading the image-builder image?

fejta (approved):
/hold

config/jobs/image-pushing/k8s-staging-sig-storage.yaml
pohly: No, that must have crept in while starting from an older base revision. Fixed.

fejta (approved):

PR #19735, opened 2020-10-28
Fix NPD e2e node tests

tosi3k:
Context: https://github.com/kubernetes/kubernetes/issues/95955.


knight42:
/hold

I think this would not solve the problem, because we have disabled insecure port in apiserver. I would think of a solution ASAP, sorry for breaking this test.

tosi3k:
>I think this would not solve the problem, because we have disabled insecure port in apiserver.

Hmm, after taking a second look think you're right. At first, I thought that the PR was only setting the default access via HTTPS that could be easily overridden like in the change I propose but now I see that we use TLS in every execution path in the node tests.

>I would think of a solution ASAP, sorry for breaking this test.

Thanks a lot! And I don't blame you, on the contrary - your change has uncovered a problem with the test :). Let's move the discussion to the original issue.

Issue #19358, opened 2020-09-24
Kettle google.cloud requirements are far out of date


MushuEE:
**What should be cleaned up or changed**:
Kettle [DockerFile](https://github.com/kubernetes/test-infra/blob/master/kettle/Dockerfile) pins versions of:
`google-cloud-pubsub==0.25.0 google-cloud-bigquery==0.27.0` These are long out of date (~3 years) and should be brought back up to date with newer API calls.

**Provide any links for context**:
https://github.com/googleapis/google-cloud-python

/area kettle
/assign

MushuEE:
#19356 is small step

PR #19740, opened 2020-10-29
Add sub stream results to staging

MushuEE:
Add in the stream.py calls for staging

This is a big step that was left out of staging

I have seen some issue with stream.py in prod, I am not sure how the sub works or if it even still exits today.

I need to look more into what is going on here, issues are related to google-cloud-bigquery versions. However, these are long outdated #19358 and the source has moved/changed. The [Table](https://googleapis.dev/python/bigquery/latest/generated/google.cloud.bigquery.table.Table.html#google.cloud.bigquery.table.Table) no longer has a method we are using and I can't find why/how that changed


spiffxp (approved):
/approve
/lgtm

PR #19744, opened 2020-10-29
Revert explicit bumps of master sizes

wojtek-t:
This is no longer needed after https://github.com/kubernetes/kubernetes/pull/95947

Fix https://github.com/kubernetes/kubernetes/issues/94098

/assign @jkaniuk


jkaniuk:
/lgtm

@mborsz

PR #19747, opened 2020-10-29
Allow Kubemark master log dumping without k/k

tosi3k:
When the new log dumping mechanism was enabled for Kubemark test jobs, it only used it for dumping logs of nodes - masters were still getting log dumped using a script from k/k (namely `k/k/test/kubemark/master-log-dump.sh` that was wrapping `k/k/cluster/log-dump/log-dump.sh`).

I ran a Prow job using a `kubekins-e2e` image based on this PR to confirm that it works as expected for Kubemark jobs.

/sig scalability
/assign @jkaniuk @wojtek-t

### DIFF
--- a/kubetest/e2e.go
+++ b/kubetest/e2e.go
@@ -232,7 +232,7 @@ func run(deploy deployer, o options) error {
 	var kubemarkDownErr error
 	if o.down && o.kubemark {
 		kubemarkWg.Add(1)
-		go kubemarkDown(&kubemarkDownErr, &kubemarkWg, dump)
+		go kubemarkDown(&kubemarkDownErr, &kubemarkWg, o.provider, dump)
 	}
 
 	if o.charts {
@@ -737,10 +737,11 @@ func kubemarkGinkgoTest(testArgs []string, dump string) error {
 }
 
 // Brings down the kubemark cluster.
-func kubemarkDown(err *error, wg *sync.WaitGroup, dump string) {
+func kubemarkDown(err *error, wg *sync.WaitGroup, provider, dump string) {
 	defer wg.Done()
 	control.XMLWrap(&suite, "Kubemark MasterLogDump", func() error {
-		cmd := exec.Command("./cluster/log-dump/log-dump.sh", dump)
+		logDumpPath := logDumpPath(provider)
+		cmd := exec.Command(logDumpPath, dump)
 		masterName := os.Getenv("MASTER_NAME")
 		cmd.Env = append(
 			os.Environ(),


kubetest/e2e.go
```diff
@@ -737,9 +739,21 @@ func kubemarkGinkgoTest(testArgs []string, dump string) error {
 }
 
 // Brings down the kubemark cluster.
-func kubemarkDown(err *error, wg *sync.WaitGroup, dump string) {
+func kubemarkDown(err *error, wg *sync.WaitGroup, provider, dump string) {
 	defer wg.Done()
 	control.XMLWrap(&suite, "Kubemark MasterLogDump", func() error {
+		if useTestInfraLogDump(provider) {
```
jkaniuk, line 745: Why not to pass updated logDumpPath to master-log-dump.sh instead?

jkaniuk: master-log-dump.sh and other scripts handle setting many variables
how do you know none of them is needed on gce?

jkaniuk: to be more specific:
1. would not allowing to change any of variables handled in master-log-dump break anyone on gce?
2. do we change in master-log-dump any defaults for variables used in log-dump

tosi3k: >would not allowing to change any of variables handled in master-log-dump break anyone on gce?

Unlikely. In `master-log-dump.sh`, we use `log-dump.sh` to just dump logs from masters using `dump_masters` function which doesn't use any covert env vars. 

>do we change in master-log-dump any defaults for variables used in log-dump

We call `export KUBEMARK_MASTER_NAME="${MASTER_NAME}"` which we already handle in the PR's logic.

Second env var is `DUMP_ONLY_MASTER_LOGS` which we also set to `true` in the PR as well as in the script.

In the sourced `k/k/test/kubemark/cloud-provider-config.sh` we set `PROJECT="${PROJECT:-}"` but that is already handled later on by `detect-project` function called in `log-dump.sh`. No relevant env vars in there.

Two other indirectly sourced files, namely `k/k/cluster/kubemark/gce/config-default.sh` and `k/k/cluster/gce/config-common.sh` do not introduce any relevant env vars used for masters log dumping as well.

TL;DR I have a strong belief that those scripts are sourced redundantly - they are used when a Kubemark cluster is being brought up and do not introduce anything important for masters' log dumping.

>is there anything to change? do we need master-log-dump at all?

I guess we don't need it... ran git log on master-log-dump.sh and apart from the script introduction in 2017 by us there were only two changes touching that file:
* shebang update in Apr 2018,
* shellcheck lint error fix in Feb 2019.

Therefore I believe we can get rid of using that script completely. I'd suggest doing that in a separate PR, though, as it would touch non-scalability jobs.

jkaniuk: I would suggest then to first merge PR removing this `master-log-dump.sh` completely (using logic from this PR) and later add support for kubekins log-dump

@wojtek-t - WDYT?

tosi3k: >I would suggest then to first merge PR removing this master-log-dump.sh completely (using logic from this PR) and later add support for kubekins log-dump

+1, such sequence of PRs seems more appropriate to me.

wojtek-t: Yes - I like this proposal.

PR #12724, opened 2019-05-21
Re-Enable 'Run older jobs first when concurrency is limited'

alvaroaleman:
/assign @cjwagner @stevekuznetsov 

Re-enables the ordering by creation date for prowJobs with limited concurrency, which got disabled in #12706.

PR #19751, opened 2020-10-29
post-test-infra-bazel: Set max_concurrency 1 to prevent races

alvaroaleman:
Right now multiple runs of this race with each other if a batch merge happens

/assign @fejta

### DIFF
--- a/config/jobs/kubernetes/test-infra/test-infra-trusted.yaml
+++ b/config/jobs/kubernetes/test-infra/test-infra-trusted.yaml
@@ -40,6 +40,7 @@ postsubmits:
     decorate: true
     branches:
     - ^master$
+    max_concurrency: 1
     spec:
       serviceAccountName: pusher
       containers:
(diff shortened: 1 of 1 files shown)


stevekuznetsov:
I can't remember, do we always trigger the oldest one first when there's a backup? Or is there a possibility that we trigger out-of-order?

alvaroaleman:
We trigger in order, I added that ~1.5 years ago: https://github.com/kubernetes/test-infra/pull/12724

fejta:
Why do we care if multiple post-test-infra-bazel jobs run in parallel?

alvaroaleman:
> Why do we care if multiple post-test-infra-bazel jobs run in parallel?

After batch merge this happens and might result in the the older revision finishing building last, resulting in the bumper to bump prow to a version that is not the latest

fejta:
This job just runs unit tests

fejta:
Maybe these jobs are what you want to change? https://github.com/kubernetes/test-infra/blob/02acbdcb97e36b76caa9808eb4bf6e4dd4a56657/config/jobs/kubernetes/test-infra/test-infra-trusted.yaml#L33-L52

alvaroaleman:
@fejta updated

fejta (approved):

PR #19753, opened 2020-10-29
prow: Preserve exclude_defaults in bugzilla plugin

jarrpa:
This may be a naive PR, but I'm running into problems with the Bugzilla plugin with the dependent bug configurations not being overridden.

Signed-off-by: Jose A. Rivera <jarrpa@redhat.com>

### DIFF
--- a/prow/plugins/config.go
+++ b/prow/plugins/config.go
@@ -1492,6 +1492,9 @@ func ResolveBugzillaOptions(parent, child BugzillaBranchOptions) BugzillaBranchO
 
 	if child.ExcludeDefaults == nil || !*child.ExcludeDefaults {
 		// populate with the parent
+		if parent.ExcludeDefaults != nil {
+			output.ExcludeDefaults = parent.ExcludeDefaults
+		}
 		if parent.ValidateByDefault != nil {
 			output.ValidateByDefault = parent.ValidateByDefault
 		}
@@ -1553,6 +1556,9 @@ func ResolveBugzillaOptions(parent, child BugzillaBranchOptions) BugzillaBranchO
 	}
 
 	// override with the child
+	if child.ExcludeDefaults != nil {
+		output.ExcludeDefaults = child.ExcludeDefaults
+	}
 	if child.ValidateByDefault != nil {
 		output.ValidateByDefault = child.ValidateByDefault
 	}
--- a/prow/plugins/config_test.go
+++ b/prow/plugins/config_test.go
@@ -615,6 +615,24 @@ func TestResolveBugzillaOptions(t *testing.T) {
 				StateAfterMerge:            &preState,
 			},
 		},
+		{
+			name:     "parent target release is excluded on child",
+			parent:   BugzillaBranchOptions{TargetRelease: &one},
+			child:    BugzillaBranchOptions{ExcludeDefaults: &yes},
+			expected: BugzillaBranchOptions{ExcludeDefaults: &yes},
+		},
+		{
+			name:     "parent target release is excluded on child with other options",
+			parent:   BugzillaBranchOptions{DependentBugTargetReleases: &[]string{one}},
+			child:    BugzillaBranchOptions{TargetRelease: &one, ExcludeDefaults: &yes},
+			expected: BugzillaBranchOptions{TargetRelease: &one, ExcludeDefaults: &yes},
+		},
+		{
+			name:     "parent exclude merges with child options",
+			parent:   BugzillaBranchOptions{DependentBugTargetReleases: &[]string{one}, ExcludeDefaults: &yes},
+			child:    BugzillaBranchOptions{TargetRelease: &one},
+			expected: BugzillaBranchOptions{DependentBugTargetReleases: &[]string{one}, TargetRelease: &one, ExcludeDefaults: &yes},
+		},
 	}
 	for _, testCase := range testCases {
 		t.Run(testCase.name, func(t *testing.T) {
@@ -776,7 +794,16 @@ orgs:
             - VERIFIED
             validate_by_default: true
             status_after_validation: MODIFIED
-            status_after_merge: NOTABUG`
+            status_after_merge: NOTABUG
+          "my-special-branch":
+            exclude_defaults: true
+            validate_by_default: false
+      another-repo:
+        branches:
+          "*":
+            exclude_defaults: true
+          "my-org-branch":
+            target_release: my-repo-branch`
 	var config Bugzilla
 	if err := yaml.Unmarshal([]byte(rawConfig), &config); err != nil {
 		t.Fatalf("couldn't unmarshal config: %v", err)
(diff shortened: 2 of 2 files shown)


prow/plugins/config.go
```diff
@@ -1492,6 +1492,9 @@ func ResolveBugzillaOptions(parent, child BugzillaBranchOptions) BugzillaBranchO
 
 	if child.ExcludeDefaults == nil || !*child.ExcludeDefaults {
 		// populate with the parent
+		if parent.ExcludeDefaults != nil {
```
alvaroaleman, line 1495: Please add tests

jarrpa: On it.

jarrpa: Test added.

AlexNPavel:
I don't believe that the `exclude_defaults` option is supposed to propagate downwards. I believe it's only supposed to apply to the current child being processed. For instance, if `exclude_defaults` is set for a specific repo, I would just expect the fields set in higher levels (like organization) to be excluded and then for all children (branches) of this repo to use what is set for the repo (unless a branch specifically sets `exclude_defaults` itself). Propagating down as this PR does would mean that branches in the repo would exclude what is set in the repo's settings, which it shouldn't.

jarrpa:
> I don't believe that the `exclude_defaults` option is supposed to propagate downwards. I believe it's only supposed to apply to the current child being processed. For instance, if `exclude_defaults` is set for a specific repo, I would just expect the fields set in higher levels (like organization) to be excluded and then for all children (branches) of this repo to use what is set for the repo (unless a branch specifically sets `exclude_defaults` itself). Propagating down as this PR does would mean that branches in the repo would exclude what is set in the repo's settings, which it shouldn't.

My specific problem is that there are settings in the OpenShift bugzilla defaults stanza for a given branch (release-4.6) that I want to not have apply to the branch of the same name in my repo in the OpenShift org. I am setting `exclude_defaults` in the branch config and it is being effectively ignored. I'll see if removing the parent stanza still lets this work.

jarrpa:
@AlexNPavel Indeed, if I remove the parent setting then the following test cases fail: https://github.com/kubernetes/test-infra/pull/19753/files#diff-d008f69794c584e7319b049f76339c7f808d68e456c715c749751a8d4130f80bR1423-R1448

These are the very test cases that are affecting me and that need to be resolved. To be even more specific, it's these lines configured in `bugzilla.default["release-4.6"]` that I want to override as nil: https://github.com/openshift/release/blob/master/core-services/prow/02_config/_plugins.yaml#L489-L490

Just from looking at the code and what little documentation exists, I think what I'm proposing is what should be correct behavior.

AlexNPavel:
@jarrpa I've looked into the actual behavior of the bugzilla option resolver more and I agree that `exclude_defaults` doesn't actually work the way I thought it did and your fix makes it behave correctly.

One thing that needs to be changed with this PR before it can merge is the test. Instead of making a new test function, the test functions `TestResolveBugzillaOptions` and `TestOptionsForBranch` should be updated to include tests for `exclude_defaults`.

AlexNPavel:
Actually, wait. This may affect other repos in unexpected ways due to the way they use `exclude_defaults` currently. Let me talk with the rest of the dptp tomorrow to see how this should be handled.

jarrpa:
@AlexNPavel Test comments addressed.

prow/plugins/config.go
```diff
@@ -1492,6 +1492,9 @@ func ResolveBugzillaOptions(parent, child BugzillaBranchOptions) BugzillaBranchO
...
+			output.ExcludeDefaults = parent.ExcludeDefaults
```
AlexNPavel, line 1496: I don't believe this should be included. The updated test still passes without this line

jarrpa: I added some more tests to more accurately illustrate the problem I'm facing:
```diff
diff --git a/prow/plugins/config_test.go b/prow/plugins/config_test.go
index 5fc7c569b2..f6abe7138d 100644
--- a/prow/plugins/config_test.go
+++ b/prow/plugins/config_test.go
@@ -797,7 +797,13 @@ orgs:
             status_after_merge: NOTABUG
           "my-special-branch":
             exclude_defaults: true
-            validate_by_default: false`
+            validate_by_default: false
+      another-repo:
+        branches:
+          "*":
+            exclude_defaults: true
+          "my-org-branch":
+            target_release: my-repo-branch`
        var config Bugzilla
        if err := yaml.Unmarshal([]byte(rawConfig), &config); err != nil {
                t.Fatalf("couldn't unmarshal config: %v", err)
@@ -857,6 +863,13 @@ orgs:
                        branch:   "my-special-branch",
                        expected: BugzillaBranchOptions{ValidateByDefault: &no, ExcludeDefaults: &yes},
                },
+               {
+                       name:     "exclude branch on repo cascades to branch config",
+                       org:      "my-org",
+                       repo:     "another-repo",
+                       branch:   "my-org-branch",
+                       expected: BugzillaBranchOptions{TargetRelease: &repoBranch, ExcludeDefaults: &yes},
+               },
        }
        for _, testCase := range testCases {
                t.Run(testCase.name, func(t *testing.T) {
@@ -941,6 +954,15 @@ orgs:
                                },
                        },
                },
+               {
+                       name: "excluded repo gets no defaults",
+                       org:  "my-org",
+                       repo: "another-repo",
+                       expected: map[string]BugzillaBranchOptions{
+                               "*":             {ExcludeDefaults: &yes},
+                               "my-org-branch": {ExcludeDefaults: &yes, TargetRelease: &repoBranch},
+                       },
+               },
        }
        for _, testCase := range repoTestCases {
                t.Run(testCase.name, func(t *testing.T) {
```

These fail without the above parent lines, the org defaults propagate into `my-org-branch` when I don't want them to.

AlexNPavel: Ok, that makes sense. To prevent `exclude_defaults` in `org` or `global` from being passed to lower levels (and thus causing themselves to be excluded), can you add this change?

```
diff --git a/prow/plugins/config.go b/prow/plugins/config.go
index 5ef8bda48f..722cbd52d6 100644
--- a/prow/plugins/config.go
+++ b/prow/plugins/config.go
@@ -1649,12 +1649,16 @@ func (b *Bugzilla) OptionsForBranch(org, repo, branch string) BugzillaBranchOpti
        if !exists {
                return options
        }
+       // do not propagate exclude_defaults downwards
+       options.ExcludeDefaults = nil
        options = ResolveBugzillaOptions(options, OptionsForItem(branch, orgOptions.Default))
 
        repoOptions, exists := orgOptions.Repos[repo]
        if !exists {
                return options
        }
+       // do not propagate exclude_defaults downwards
+       options.ExcludeDefaults = nil
        options = ResolveBugzillaOptions(options, OptionsForItem(branch, repoOptions.Branches))
 
        return options
```

This will prevent downwards propagation while still allowing the wildcard repo branch (`*`) to set `exclude_defaults` for other branches in the repo.

AlexNPavel (approved):
/lgtm

PR #19715, opened 2020-10-27
CAPZ: add "e2e-full" job that includes GPU-based nodes

mboersma:
kubernetes-sigs/cluster-api-provider-azure#1002 adds a Ginkgo e2e spec to provision and test a GPU-enabled Azure cluster, but skips that test by default. This adds a new `pull-cluster-api-provider-azure-e2e-full` job that can be run on demand to include the GPU-enabled test spec.

cc: @CecileRobertMichon @devigned

PR #19756, opened 2020-10-29
fix: remove run_if_changed line from CAPZ e2e-full job

mboersma:
#19715 mistakenly included a `run_if_changed` line, but was intended to be triggered only manually.

cc: @CecileRobertMichon


CecileRobertMichon (approved):
/ok-to-test
/lgtm

cpanato:
We don't want that thing anymore? Or it is not working as expected?

@mboersma @CecileRobertMichon

CecileRobertMichon:
@cpanato we do, just not on `pull-cluster-api-provider-azure-e2e-ful` which is meant to not run by default (it's a new job @mboersma just added)

cpanato:
Ahhh sorry my mistake then

PR #19757, opened 2020-10-29
add APIPriorityAndFairness to presubmit and periodic alpha jobs

adtac:


adtac:
cc @yue9944882 @ahg-g @MikeSpreitzer

adtac:
note: https://github.com/kubernetes/kubernetes/pull/94925 will need to add "Feature:APIPriorityAndFairness" to the ginkgo test description

yue9944882:
@adtac thanks, i added `[Feature:APIPriorityAndFairness]` to the e2e test name.

config/jobs/kubernetes/sig-cloud-provider/gcp/gcp-gce.yaml
```diff
@@ -457,7 +457,7 @@ presubmits:
         - --provider=gce
         - --runtime-config=api/all=true
         - --stage=gs://kubernetes-release-pull/ci/pull-kubernetes-e2e-gce-alpha-features
-        - --test_args=--ginkgo.focus=\[Feature:(ServiceAccountIssuerDiscovery|StorageVersionHash|Audit|PodPreset|RunAsGroup|TTLAfterFinished|NodeLease|CSIStorageCapacity|GenericEphemeralVolume)\]|Networking --ginkgo.skip=\[Feature:(SCTPConnectivity|Volumes|Networking-Performance)\]|IPv6|csi-hostpath-v0 --minStartupPods=8
+        - --test_args=--ginkgo.focus=\[Feature:(ServiceAccountIssuerDiscovery|StorageVersionHash|Audit|PodPreset|RunAsGroup|TTLAfterFinished|NodeLease|CSIStorageCapacity|GenericEphemeralVolume|APIPriorityAndFairness)\]|Networking --ginkgo.skip=\[Feature:(SCTPConnectivity|Volumes|Networking-Performance)\]|IPv6|csi-hostpath-v0 --minStartupPods=8
```
MikeSpreitzer, line 460: Is this enough of a change?  Remember that the APIPriorityAndFairness feature gate does not work unless the "runtime config" of `flowcontrol.apiserver.k8s.io/v1alpha1=true` is also given.

yue9944882: > Is this enough of a change? 

i think so, if you notice, the line above sets `--runtime-config=api/all=true`, i guess that's for the e2e target server

adtac: yes, `--runtime-config=api/all=true` should take care of that. However, `--enable-priority-and-fairness` won't be included in the API server's cmdline flags, which is a concern. Since we're slowly graduating this to beta, perhaps we can make `--enable-priority-and-fairness` default to true?

adtac: after discussing this on slack, I now understand that `--enable-priority-and-fairness` is true by default, so it shouldn't be a concern

andrewsykim (commented):
/approve
/lgtm

MikeSpreitzer:
Results can be seen at https://testgrid.k8s.io/google-gce#gce-cos-master-alpha-features

PR #19758, opened 2020-10-29
update canary bazel jobs to latest image

mikedanese:
/assign @BenTheElder @fejta 
/cc @justaugustus


fejta (approved):

fejta:
Isn't there a master-latest or latest-master? Might want to change it to those tags. Otherwise these jobs are not really doing anything different than the non-canary.

BenTheElder:
yeah, IIRC the other kubekins canaries use a `$variant-latest` + `imagePullPolicy: always`

PR #19719, opened 2020-10-27
Create a new plugin supports /trick-or-treat

chaodaiG:
Collaborated with mpherman2@ on this project, and figured it might bring some fun to k8s developers as we approach halloween

/cc @ mpherman2 @michelle192837 @e-blackwelder

PR #19759, opened 2020-10-29
Import trick-or-treat plugin in hook

chaodaiG:
These were missing from #19719 

By the way renaming the folder and file names to make them consistent with other plugins

/cc @fejta @michelle192837


chaodaiG:
For awareness @mpherman2

fejta (approved):

PR #19772, opened 2020-10-30
sig-storage images: fix canary rebuild job

pohly:
The build rules rely on PULL_BASE_REF for determining what the
tag name should be, so it has to be set also for the periodic job.


pohly:
/assign @fejta 

This is a small fix for PR https://github.com/kubernetes/test-infra/pull/19734

fejta (approved):

PR #19779, opened 2020-10-30
Add more notes to prow/phony README

chaodaiG:


prow/cmd/phony/README.md
```diff
@@ -13,7 +13,11 @@ go run prow/cmd/hook/main.go
 --config-path=config/prow/config.yaml
 --plugin-config=config/prow/plugins.yaml
 --hmac-secret-file=path/to/hmac
--github-token-path=path/to/github-token
+--github-token-path=path/to/github-token
+
+# Note:
+# --deck-url is required because --dry-run flag requires it
```
alvaroaleman, line 19: Are you sure? That flag is only used for the `github` client

chaodaiG: That's what is says when I didn't supply it. The error message might be wrong then

chaodaiG: Why does `github` client need it btw?

alvaroaleman: github requires it so we don't do mutating api calls (we just do nothing instead).

And I was looking at the wrong binary (deck), where we indeed don't require it

/lgtm
/approve

Issue #19477, opened 2020-10-07
Mitigate dockerhub changes rolling out Nov 1st


spiffxp:
Creating this as a placeholder based on discussion in Slack and yesterday's SIG Testing meeting.  I'll sketch in what I remember but to defer to @BenTheElder for a plan

Dockerhub is going to rate-limit pulls starting Nov 1st.  See https://www.docker.com/pricing/resource-consumption-updates

> A pull request is up to two GET requests to the registry URL path ‘/v2/*/manifests/*’. [...] Note that all pull requests, including ones for images you already have, are counted by this method. This is the trade-off for not counting individual layers.

> Rate limits for Docker image pulls are based on the account type of the user requesting the image - not the account type of the image’s owner.

> Because we are limiting on manifest requests, the number of layers (blob requests) related to a pull is unlimited at this time.

> Please see https://docs.docker.com/registry/recipes/mirror/ to run docker/distribution as a mirror

Pull limit is:
- anonymous (rate limited per IP) - 100 per 6h (~16/h)
- authenticated - 200 per 6h (~33/h)
- paid plans - unlimited

Ideas:
- invest in a pool of paid dockerhub accounts that can be distributed to all CI nodes and all e2e clusters under test
- pull through cache
- audit our tests and image builds, and reduce or remove our dependence on any dockerhub-hosted images
- steer people toward k8s.gcr.io and reduce friction

We will likely need to fan out audit/change jobs to all SIGs / subprojects.  

I think to start with we should ensure merge-blocking kubernetes/kubernetes jobs are safe, since they represent significant CI volume

spiffxp:
/assign @BenTheElder 
/sig testing
/priority important-soon
We should survey soon and evaluate priority based on that.

spiffxp:
/area images
/area jobs
for changing jobs
/area prow
for the idea of a pull-through cache

aojea:
If all the images used are here

https://github.com/kubernetes/kubernetes/blob/00b6d7ccee30b435509e1547916ab1b42738ff21/test/utils/image/manifest.go#L204-L244

There are only 8 images, it seems easy to move them to gcr.io

```
        configs[BusyBox] = Config{dockerLibraryRegistry, "busybox", "1.29"}
        configs[GlusterDynamicProvisioner] = Config{dockerGluster, "glusterdynamic-provisioner", "v1.0"}
        configs[Httpd] = Config{dockerLibraryRegistry, "httpd", "2.4.38-alpine"}
[...]

PR #19780, opened 2020-10-30
Configure kubekins dockerd to use mirror.gcr.io

spiffxp:
Follow instructions from https://cloud.google.com/container-registry/docs/pulling-cached-images

This is intended for jobs that use docker-in-docker via the `preset-dind-enabled` preset and also use the kubekins image

ref: https://github.com/kubernetes/test-infra/issues/19477

### DIFF
--- a/images/kubekins-e2e/Dockerfile
+++ b/images/kubekins-e2e/Dockerfile
@@ -90,6 +90,13 @@ RUN [ "${UPGRADE_DOCKER_ARG}" = "true" ] && \
     rm -rf /var/lib/apt/lists/* && \
     sed -i 's/cgroupfs_mount$/#cgroupfs_mount\n/' /etc/init.d/docker || true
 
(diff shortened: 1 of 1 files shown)


spiffxp:
Waiting on postsubmit to build https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-test-infra-push-kubekins-e2e/1322240454648926208

spiffxp:
Successfully pushed
2020-10-30 18:23:26 UTC - image pushed by https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-test-infra-push-kubekins-e2e/1322240454648926208
2020-10-30 18:50:10 UTC - image used by https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-e2e-gce-canary/1322249448511770624  (which passed)

So this hasn't broken things

PR #19791, opened 2020-10-31
vulndash: adding cpu/mem requests

cpanato:
defining cpu/mem requests for the job, because this execution timeout: https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/ci-release-vulndash-update/1322405073002696704

and running locally in my machine took around 1:30 minute (2,3 GHz Quad-Core Intel Core i7 / 32 GB )

/assing @justaugustus


hasheddan (commented):
/assign @justaugustus 

Probably good to set some limits on this as well 👍

justaugustus (changes requested):
Agreed with @hasheddan; @cpanato -- would you mind setting identical `limits` here as well?

cpanato:
yes! makes totally sense, thanks for the review @hasheddan @justaugustus 
did the change, PTAL

PR #19796, opened 2020-11-02
Remove out of date comment

hi-rustin:
I think the `no-issue` is outdated.

And seems no logic related to `no-issue`(I don't know if this is only for consistency with the approve plugin, but I find it weird and confusing).

### DIFF
--- a/prow/plugins/lgtm/lgtm.go
+++ b/prow/plugins/lgtm/lgtm.go
@@ -373,11 +373,10 @@ func handle(wantLGTM bool, config *plugins.Configuration, ownersClient repoowner
 	return nil
 }
 
-func stickyLgtm(log *logrus.Entry, gc githubClient, config *plugins.Configuration, lgtm *plugins.Lgtm, author, org, repo string) bool {
+func stickyLgtm(log *logrus.Entry, gc githubClient, _ *plugins.Configuration, lgtm *plugins.Lgtm, author, org, repo string) bool {
 	if len(lgtm.StickyLgtmTeam) > 0 {
 		if teams, err := gc.ListTeams(org); err == nil {
 			for _, teamInOrg := range teams {
-				// lgtm.TrustedAuthorTeams is supposed to be a very short list.
 				if strings.Compare(teamInOrg.Name, lgtm.StickyLgtmTeam) == 0 {
 					if members, err := gc.ListTeamMembers(teamInOrg.ID, github.RoleAll); err == nil {
 						for _, member := range members {


hi-rustin:
Hi @fejta  @alvaroaleman, I think I will continue to contribute to test-infra for a long time.  Could you become my sponsors?

prow/plugins/lgtm/lgtm.go
```diff
@@ -373,11 +373,10 @@ func handle(wantLGTM bool, config *plugins.Configuration, ownersClient repoowner
 	return nil
 }
 
-func stickyLgtm(log *logrus.Entry, gc githubClient, config *plugins.Configuration, lgtm *plugins.Lgtm, author, org, repo string) bool {
+func stickyLgtm(log *logrus.Entry, gc githubClient, _ *plugins.Configuration, lgtm *plugins.Lgtm, author, org, repo string) bool {
 	if len(lgtm.StickyLgtmTeam) > 0 {
 		if teams, err := gc.ListTeams(org); err == nil {
 			for _, teamInOrg := range teams {
-				// lgtm.TrustedAuthorTeams is supposed to be a very short list.
```
fejta, line 380: Why are we dropping this comment?

fejta (changes requested):
/ok-to-test

Will you also fix this in the approve plugin? https://github.com/kubernetes/test-infra/blob/6cc991bea964f51ee828aa3f300d404a4715a0c4/prow/plugins/approve/approve.go#L46?

fejta:
Awesome, thanks so much for the fix here!

And nice to meet you :) Yes, please keep contributing for a bit and I'll happily sponsor.

prow/plugins/lgtm/lgtm.go
hi-rustin: TrustedAuthorTeams outdated.

hi-rustin:
> Will you also fix this in the approve plugin?

I think `no-issue` working on `approve`, because we have a `IssueRequired` configuration. So we cannot remove it.

prow/plugins/lgtm/lgtm.go
hi-rustin: Now we only have `StickyLgtmTeam`.

fejta:
The issue is you can configure `/lgtm no-issue` to do the same thing as `/approve no-issue` so if we recognize the approve we should also lgtm?  https://github.com/kubernetes/test-infra/blob/master/prow/plugins/approve/approve.go#L365

hi-rustin:
> The issue is you can configure `/lgtm no-issue` to do the same thing as `/approve no-issue` so if we recognize the approve we should also lgtm? https://github.com/kubernetes/test-infra/blob/master/prow/plugins/approve/approve.go#L365

Yes, you are right. Thanks for your review.

fejta (approved):

fejta:
I started a discussion on the #prow slack channel about removing support for `/lgtm` == `/approve`, if we think this is a good idea would you want to work on that?

hi-rustin:
> I started a discussion on the #prow slack channel about removing support for `/lgtm` == `/approve`, if we think this is a good idea would you want to work on that?

Of course 😄

hi-rustin:
> Awesome, thanks so much for the fix here!
> 
> And nice to meet you :) Yes, please keep contributing for a bit and I'll happily sponsor.

@fejta  @alvaroaleman How about now? 😄  
I have created 15 PRs for test-infra, and 12 PRs have been merged.

PR #19799, opened 2020-11-02
Fix autobumper: replace plank with prow controller manager

chaodaiG:
Plank was already replaced by prow controller manager, but autobumper.sh still uses plank as reference, updating it to make it work

/assign @fejta @michelle192837 @chases2

### DIFF
--- a/prow/cmd/autobump/autobump.sh
+++ b/prow/cmd/autobump/autobump.sh
@@ -18,7 +18,7 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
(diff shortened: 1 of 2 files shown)


spiffxp (commented):
/lgtm
/hold
remove when you're ready or have the reviews you want

~Part of me wonders if we should continue bumping both until an announced deadline.~

~I see we announced a deprecation October 2020, but gave no deadline.  Is that enough notice for anyone who happens to be consuming the deployment as an artifact?~

Nevermind, I read more closely, this is about treating the new binary as the reference version

chaodaiG:
Thank you Aaron! This is already broken, let's merge it and fix if there is any problem

/unhold

PR #19801, opened 2020-11-02
Enable test failure on logexporter failures in scalability test jobs

tosi3k:
/sig scalability
/assign @jkaniuk


jkaniuk:
Could we have it for presubmits first?
I'm fine with either: master only or all presubmits.

tosi3k:
> Could we have it for periodics first?
> I'm fine with either: master only or all periodics, as it is a low risk change.

Done - I propose we enable it for all periodics and then proceed with enabling this for all jobs provided we don't come across any problems.

jkaniuk:
/lgtm
@mm4tt

PR #19803, opened 2020-11-02
Revert "Revert "Increase memory for pull-kubernetes-node-e2e""

eddiezane:
Further analysis by @spiffxp found that the increase in memory consumption is
most likely related to https://github.com/kubernetes/kubernetes/pull/95239. It
is not immediately clear how to trim down usage here so bumping the resource
limit is the agreed approach.

ref: https://github.com/kubernetes/test-infra/pull/19615#issuecomment-720615200


spiffxp (approved):
/approve
/lgtm
Thanks!  Sorry again for the flail on this

For posterity's sake, we don't need to make this change on existing release branches because they don't have the culprit PR.

PR #19813, opened 2020-11-03
Add custom label prefix support for cherry picker

hi-rustin:
We want to use a custom prefix to be consistent with the label style we currently use.

I think other communities also have needs. Maybe we can migrate these labels, but I think it is more reasonable to support custom label style.

### DIFF
--- a/prow/external-plugins/cherrypicker/main.go
+++ b/prow/external-plugins/cherrypicker/main.go
@@ -46,6 +46,7 @@ type options struct {
 	prowAssignments   bool
 	allowAll          bool
 	issueOnConflict   bool
+	labelPrefix       string
 }
 
 func (o *options) Validate() error {
@@ -68,6 +69,7 @@ func gatherOptions() options {
 	fs.BoolVar(&o.prowAssignments, "use-prow-assignments", true, "Use prow commands to assign cherrypicked PRs.")
 	fs.BoolVar(&o.allowAll, "allow-all", false, "Allow anybody to use automated cherrypicks by skipping GitHub organization membership checks.")
 	fs.BoolVar(&o.issueOnConflict, "create-issue-on-conflict", false, "Create a GitHub issue and assign it to the requestor on cherrypick conflict.")
+	fs.StringVar(&o.labelPrefix, "label-prefix", defaultLabelPrefix, "Set a custom label prefix.")
 	for _, group := range []flagutil.OptionGroup{&o.github} {
 		group.AddFlags(fs)
 	}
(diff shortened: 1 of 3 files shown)


prow/external-plugins/cherrypicker/lib/lib.go
```diff
@@ -19,10 +19,10 @@ package cherrypicker
 import "fmt"
 
 // CreateCherrypickBody creates the body of a cherrypick PR
-func CreateCherrypickBody(num int, requestor, note string) string {
+func CreateCherrypickBody(num int, requester, note string) string {
 	cherryPickBody := fmt.Sprintf("This is an automated cherry-pick of #%d", num)
-	if len(requestor) != 0 {
-		cherryPickBody = fmt.Sprintf("%s\n\n/assign %s", cherryPickBody, requestor)
+	if len(requester) != 0 {
+		cherryPickBody = fmt.Sprintf("%s\n\n/assign %s", cherryPickBody, requester)
```
petr-muller, line 25: Please remove these, `requestor` is a valid variant of the [word](https://dictionary.cambridge.org/dictionary/english/requester) so this just pollutes the change.

/ok-to-test

petr-muller (changes requested):

petr-muller:
What's the rationale? Why should cherrypick-related labels have a different prefix?

hi-rustin:
> What's the rationale? Why should cherrypick-related labels have a different prefix?

Because we want to use a custom prefix to be consistent with the label style we currently use.

I think other communities also have needs. Maybe we can migrate these labels, but I think it is a more reasonable way to support custom label style.

prow/external-plugins/cherrypicker/server_test.go
```diff
@@ -570,13 +580,13 @@ func testCherryPickPRWithLabels(clients localgit.Clients, t *testing.T) {
 			},
 			prLabels: []github.Label{
 				{
-					Name: "cherrypick/release-1.5",
+					Name: labelPrefix + "release-1.5",
```
stevekuznetsov, line 583: nit: Prefer not to parameterize your test cases. This test is not only testing the output of the plugin but also testing that you correctly create test data here at runtime. Passing static data would be preferred.

prow/external-plugins/cherrypicker/main.go
```diff
@@ -68,10 +69,11 @@ func gatherOptions() options {
 	fs.BoolVar(&o.prowAssignments, "use-prow-assignments", true, "Use prow commands to assign cherrypicked PRs.")
 	fs.BoolVar(&o.allowAll, "allow-all", false, "Allow anybody to use automated cherrypicks by skipping GitHub organization membership checks.")
 	fs.BoolVar(&o.issueOnConflict, "create-issue-on-conflict", false, "Create a GitHub issue and assign it to the requestor on cherrypick conflict.")
+	fs.StringVar(&o.labelPrefix, "label-prefix", "", "Set a custom label prefix.")
```
stevekuznetsov, line 72: You could expose the default prefix as a `const` and make it the default value of this flag, so you don't need the `len() != 0` defaulting logic later.

hi-rustin: Fixed.

prow/external-plugins/cherrypicker/server_test.go
hi-rustin: Fixed.

hi-rustin:
@stevekuznetsov  How about now?
PTAL~ Thanks!

hi-rustin:
@stevekuznetsov ping~

kargakis (approved):

PR #19819, opened 2020-11-03
Fix security-profiles-operator image pushing bucket

saschagrunert:
We changed the bucket name to shorthand `security-profile` to `sp` to
fix the name length limitation. This is now reflected here, too.

cc @pjbgf @hasheddan


hasheddan (approved):
/lgtm

pjbgf (approved):
/lgtm

PR #19824, opened 2020-11-03
cloud-provider-openstack: set ignore_review_state to true

lingxiankong:
Sets ignore_review_state to true for kubernetes/cloud-provider-openstack to avoid the auto approval for the PRs submitted by the person with 'approve' permission.

### DIFF
--- a/config/prow/plugins.yaml
+++ b/config/prow/plugins.yaml
@@ -99,6 +99,10 @@ approve:
   - helm/charts
   require_self_approval: false
   lgtm_acts_as_approve: true
+- repos:
+  - kubernetes/cloud-provider-openstack
+  require_self_approval: false
+  ignore_review_state: true
 
 # Lower bounds in number of lines changed; XS is assumed to be zero.
 size:


config/prow/plugins.yaml
```diff
@@ -99,6 +99,10 @@ approve:
   - helm/charts
   require_self_approval: false
   lgtm_acts_as_approve: true
+- repos:
+  - kubernetes/cloud-provider-openstack
```
dims, line 103: you can add this line after 99 before 100 i think. since that section has the flags you need ( require_self_approval - false and ignore_review_state - true)

lingxiankong: it doesn't include `ignore_review_state`, does it?

jichenjc:
the purpose of this PR lgtm , thanks for this @lingxiankong

config/prow/plugins.yaml
dims: whoops. you are right

PR #19840, opened 2020-11-04
windows servercore cache: Adds kubernetes extra_refs

claudiubelu:
Currently the job fails because the periodic jobs do not have projects cloned into them. We can fix this by having extra_refs with the
required project. This is similarly done by other periodic image building jobs: [1]

[1] https://github.com/kubernetes/test-infra/blob/f022c95bc87531c16db48b5cde151916fa78464d/config/jobs/image-pushing/k8s-staging-sig-storage-periodic.yaml


spiffxp (approved):
/approve
/lgtm

Issue #19836, opened 2020-11-04
The getting_started_deploy.md confused beginners


hi-rustin:
**What would you like to be added**:
Update getting_started_deploy.md to help beginners get started quickly.

**Why is this needed**:
It seems that our starter deploys prow into the `prow` namespace. But `getting_started_deploy.md` still uses the default namespace.

hh:
## Usage of `oauth` when referring to `gitlab-token` causes confusion

I think it's beyond the docs and also in the command line args and mount-points for the `github-token` in the example manifests.

### `github-oauth` is used for Prow and PR status

https://github.com/kubernetes/test-infra/blob/master/prow/getting_started_deploy.md#set-up-github-oauth

### `oauth` is confusingly also used in the examples and paths referring to `github-token`.

```
3. Set this token aside for later (we'll assume you wrote it to a file on your workstation at /path/to/oauth/secret)
```

```
- Configuring GitHub to send prow webhooks for your repos. This is where you'll provide the absolute /path/to/oauth/secret
```

https://github.com/kubernetes/test-infra/blob/master/prow/getting_started_deploy.md#create-the-github-secrets

```
The github-token is the OAuth2 token you created above for the [GitHub bot account]. If you need to create one, go to https://github.com/settings/tokens.

> kubectl create secret generic github-token --from-file=token=/path/to/oauth/secret
```

```
$ bazel run //experiment/update-hook -- \
  --hmac-path=/path/to/hook/secret \
  --github-token-path=/path/to/oauth/secret
```

hh:
Just wanting to make sure I'm not missing something with regards to referring to the token as an oauth secret of some sort.

hh:
Seems the internal name within the prowjob k8s api type also refers to an `OauthTokenSecret`:
https://github.com/kubernetes/test-infra/blob/master/prow/apis/prowjobs/v1/types.go#L364-L366

I'm suspecting this is where the nomenclature for overlap may be stemming from.

hh:
Opening this (just for doc updates regarding oauth api tokens vs webui tokens):
- https://github.com/kubernetes/test-infra/pull/19847

Mainly because I found it very confusing, since we generate, use, and configure both kinds.

PR #19849, opened 2020-11-05
Update the docs to handle prow running its own namespace

hi-rustin:
Please refer to https://github.com/kubernetes/test-infra/issues/19836

### DIFF
--- a/prow/getting_started_deploy.md
+++ b/prow/getting_started_deploy.md
@@ -132,14 +132,14 @@ randomness-generator, eg `openssl rand -hex 20`.
 
 ```sh
 $ openssl rand -hex 20 > /path/to/hook/secret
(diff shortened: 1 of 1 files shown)


spiffxp:
My preference would be that we not do this and instead update the docs to handle prow running its own namespace.

/cc @fejta @alvaroaleman 
I'm going to defer to those with more prow ownership than I

alvaroaleman:
> My preference would be that we not do this and instead update the docs to handle prow running its own namespace.

Yeah I agree. Having Prow in its own namespace allows to easily clean it up if not needed anymore and to more easily check what's workloads and other objects associated to Prow exist.

hi-rustin:
Got it. I will try to update the docs.

fejta (approved):

PR #19858, opened 2020-11-05
capdo: use the same image as the presubmit job

cpanato:
the only difference for the presubmit job is the image

/assign @MorrisLaw @xmudrii @prksu


xmudrii (approved):
/lgtm
/approve

PR #19861, opened 2020-11-05
Tide: Deduplicate checkrun results and use the best only

alvaroaleman:
Checkruns can be restarted, resulting in mulitiple checkruns of the same
name being reported on a given commit with potentially different
results.

This PR makes tide de-duplicate them and use the best, similiar to what
we already do for ProwJobs.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -51,6 +51,7 @@ require (
 	github.com/gomodule/redigo v1.7.0
 	github.com/google/go-cmp v0.5.0
 	github.com/google/go-github v17.0.0+incompatible
+	github.com/google/gofuzz v1.1.0
 	github.com/google/uuid v1.1.1
 	github.com/gorilla/csrf v1.6.2
 	github.com/gorilla/mux v1.7.4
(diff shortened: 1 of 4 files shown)


matthyx:
super clean implementation!
/lgtm
/hold

stevekuznetsov (approved):
I'm not super familiar with how check runs show up in the UI - but if we expect a success to beat out all other statuses, we will merge even if the latest run has failed, is this a case where the UI shows a checkrun failure?

/hold

alvaroaleman:
> I'm not super familiar with how check runs show up in the UI - but if we expect a success to beat out all other statuses, we will merge even if the latest run has failed, is this a case where the UI shows a checkrun failure?

Hypothetically this could happen, the UI from what it looks always showed the latest run. But that would require a successful run to happen, then someone trigger a re-run and then that re-run to fail. While possible, its unlikely because why would you re-run a test that succeeded?

It is the very same with prowjobs btw, tide just checks if it finds any successful prowjob for the given commit, if a prowjob first succeeds, then gets manually retriggered and fails, tide would still merge.

PR #19862, opened 2020-11-06
Sinker supports excluding build cluster

chaodaiG:
Sinker treats all build clusters stored in kubeconfig secret as managed, majority of times this is valid assumption. But it could be handy to allow sinker ignore certain build clusters. For example,  migrating workloads from prow instance A to prow instance B, to achieve seemless transition (aka no down time), there will be certain amount of time that sinker instances of prow A and prow B both are working, which will cause sinker A delete pods scheduled by prow B, vice versa.

This PR provides a solution to this problem, so that the build cluster can be temporarily ignored by sinker A and sinker B, thus prow A and prow B can both work at the same time.

/cc @spiffxp @amwat @alvaroaleman

### DIFF
--- a/prow/cmd/sinker/main.go
+++ b/prow/cmd/sinker/main.go
@@ -363,6 +363,17 @@ func (c *controller) clean() {
 	// Now clean up old pods.
 	for cluster, client := range c.podClients {
 		log := c.logger.WithField("cluster", cluster)
+		var isClusterExcluded bool
+		for _, excludeCluster := range c.config().Sinker.ExcludeClusters {
+			if excludeCluster == cluster {
+				isClusterExcluded = true
+				break
+			}
+		}
+		if isClusterExcluded {
+			log.Debugf("Cluster %q is excluded, skipping pods deletion.", cluster)
+			break
+		}
 		var pods corev1api.PodList
 		if err := client.List(c.ctx, &pods, ctrlruntimeclient.MatchingLabels{kube.CreatedByProw: "true"}, ctrlruntimeclient.InNamespace(c.config().PodNamespace)); err != nil {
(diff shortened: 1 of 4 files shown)


alvaroaleman:
Why not simply remove the kubeconfig?

chaodaiG:
> Why not simply remove the kubeconfig?

The idea is for seemless transition between two prow instances, to avoid downtime. It would be expected that prow A still functions at the moment prow B starts managing workloads, and at that time we would like prow A still being able to schedule pods with prow controller manager, which requires the existence of kubeconfig.

If it's acceptable to have a short period of downtime, then yes we can just delete the kubeconfig, but that requires prow operator onboarding prow B as soon as possible to unblock PRs.

chaodaiG:
@alvaroaleman any thoughts?

alvaroaleman (approved):
I am not really a fan of this and think ppl should just remove the kubeconfig (I remember there was that discussion with cole where he wanted a merged kubeconfig rather than one kubeconfig per cluster, that decision is presumably why you need this now), but I guess it doesn't have a high maintenance cost

chaodaiG:

Fixed test error due to updates in master branch, @alvaroaleman please add another /lgtm

chaodaiG:
There is another legit error, it didn't fail go test on my local though. Let me see whether I can/should fix it

chaodaiG:
ok I introduced a flaky test, let me deflake it first.

chaodaiG:
The flaky test turned out to be a code logic error, should be fixed now

chaodaiG:
@alvaroaleman please add another lgtm

alvaroaleman (approved):

PR #19870, opened 2020-11-06
Return stderr output in kubernetes_build called subprocesses

hasheddan:
Redirects stderr to stdout such that it will be included in the output
of a subprocess that is executed in kubernetes_build. This can be useful
for troubleshooting errors.

See [subprocess docs](https://docs.python.org/2/library/subprocess.html) for info.

xref: https://github.com/kubernetes/release/issues/1693

/cc @justaugustus @ameukam 
/assign @dims (for his python wisdom)


saschagrunert (approved):
/lgtm

spiffxp (approved):
/approve
/lgtm

PR #19871, opened 2020-11-06
config/tests/jobs: add check for use of k8s-release-dev bucket

spiffxp:
enforce that anything extracting from ci/latest-fast should be
using the k8s-release-dev bucket

drop ci/latest-fast from skew job; I would prefer it use latest-fast,
but ci/stable1 hasn't migrated over yet, and extract-ci-bucket applies
to all extracts

This should close out https://github.com/kubernetes/test-infra/issues/19484
but I'll do that manually after looking things over


spiffxp:
FYI @kubernetes/ci-signal

cpanato (approved):
thanks for that Aaron

/lgtm

PR #19878, opened 2020-11-06
Trigger: Abort jobs on pushes

alvaroaleman:
Plank already has logic to abort older jobs, which causes it to
terminate all older instances of a given job after a push. This won't
work if the new revision doesn't trigger this job (e.G. due to
`run_if_changed`). This change makes trigger abort all jobs for a given
PR on a push event prior to creating new ones.


matthyx:
How come this doesn't require a fix in tests?

PR #19888, opened 2020-11-09
Use logrusutil.ComponentInit replace manual setup 

hi-rustin:
Use logrusutil.ComponentInit replace manual setup.

### DIFF
--- a/prow/external-plugins/refresh/main.go
+++ b/prow/external-plugins/refresh/main.go
@@ -28,6 +28,7 @@ import (
 
 	"github.com/sirupsen/logrus"
 	"k8s.io/test-infra/prow/interrupts"
+	"k8s.io/test-infra/prow/logrusutil"
 	"k8s.io/test-infra/prow/pjutil"
 
 	"k8s.io/test-infra/pkg/flagutil"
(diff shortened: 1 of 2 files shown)


prow/external-plugins/refresh/main.go
```diff
@@ -84,8 +84,18 @@ func main() {
 	}
 
 	logrus.SetFormatter(&logrus.JSONFormatter{})
-	// TODO: Use global option from the prow config.
-	logrus.SetLevel(logrus.DebugLevel)
```
alvaroaleman, line 88: Please use `logrusutil.ComponentInit`. Also, the whole level setting code isn't needed, we do that during config load already: https://github.com/kubernetes/test-infra/blob/8e5725a1d2c4186b7cc5f46cae1a132559a4d0af/prow/config/config.go#L1850

hi-rustin: Got it. Thanks for your review!

alvaroaleman (commented):
/ok-to-test
You will need to update bazel

alvaroaleman (approved):

PR #19844, opened 2020-11-04
releng: Add the contact information in case of job issue

cpanato:
To send an email notification to the release-team or
release-managers+alerts

Fixes: https://github.com/kubernetes/sig-release/issues/1216

/hold not 100% sure if the emails are correct to the jobs

/assign @alejandrox1 @hasheddan @saschagrunert @justaugustus

Issue #18599, opened 2020-08-01
Kubernetes CI Policy: all prow.k8s.io jobs must have testgrid alerts configured


spiffxp:
Part of https://github.com/kubernetes/test-infra/issues/18551

**Why this is important:**
- In order to ensure effective use of community resources, we expect them to be spent on jobs that provide useful signal, and are actively maintained
- Configuring testgrid alerts requires setting an e-mail address, which gives us a point of contact to escalate to if a job is deemed an ineffective use of community resources
- We'll use this to implement a policy where we reserve the right to remove/disable jobs that are deemed an ineffective use of resources (e.g. perma-failing for O(weeks)) if the point of contact is unresponsive

TODO:
- come up with test to enforce this policy (logging only)
- come up with report to identify jobs and likely candidate sig owners
  - eg: if a job is on "sig-foo"'s dashboard, it's likely they want to own it
  - who to contact for jobs that aren't sigs?
- send out notice to all sigs, give a deadline of N weeks
- any jobs that don't have e-mail addresses configured should be removed
- flip test to failing once all jobs meet the policy

Other thoughts / notes:
- need to parse both prowjob config and testgrid config for this
  - there are some testgrids not populated by prow, we can probably ignore these?
  - there are some prowjobs that don't have all their testgrid config in annotations

/sig testing

spiffxp:
Some prior art to serve as starting points:
- `go run ./experiment/prowjob-report/main.go --config ./config/prow/config.yaml --job-config ./config/jobs --format csv > jobs.csv` which I periodically reimport into [this spreadsheet](https://docs.google.com/spreadsheets/d/1dgSxfa0jdaYk76S6DqJeazl2TKVAvZcCg-7Yx3U0j-I/edit#gid=1043959272)
  - the `owner_dash` column is a guess on who should own the job based on which sig/wg-prefixed testgrid dashboard the job lives under
  - could run with `--format json` and pipe through to jq
  - there are some gaps; the big one I'm aware of is that it misses testgrid info that is solely in testgrid config
- This test (which does more than its name lets on) should serve as a good starting point for how to fail or log jobs/tabs that don't have alerts configured: https://github.com/kubernetes/test-infra/blob/2293860ee4926f2c474188805007f71462966edc/config/tests/testgrids/config_test.go#L506-L544

RobertKielty:
@spiffxp mysel @ScrapCodes and @raydeeam will do work together on the test. Coordinating with them both now.

[...]

PR #19892, opened 2020-11-09
Add test for testrgrid alert emails on release-blocking jobs

hasheddan:
Adds a check that release-blocking jobs have testgrid alert emails. Logs
violations for now, but should be updated to error on violations when
all jobs pass.

xref: https://github.com/kubernetes/sig-release/issues/1216 https://github.com/kubernetes/test-infra/pull/19844#issuecomment-721867705

/assign @justaugustus @cpanato

### DIFF
--- a/config/tests/jobs/jobs_test.go
+++ b/config/tests/jobs/jobs_test.go
@@ -969,6 +969,13 @@ func verifyPodQOSGuaranteed(spec *coreapi.PodSpec) (errs []error) {
 	return errs
 }
 
+func hasAlertEmails(job cfg.JobBase) bool {
+	if _, ok := job.Annotations["testgrid-alert-email"]; !ok {
+		return false
+	}
+	return true
+}
+
 // A job is merge-blocking if it:
 // - is not optional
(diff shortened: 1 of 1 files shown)


chases2 (approved):

cpanato:
Thanks for this @hasheddan

spiffxp:
Sorry I missed this.  We should revert this, because there is already a test that does this, and catches more corner cases.  It is possible to configure test grid alerts outside of prowjob files, so just checking prow job annotations is insufficient.

Instead, we should build on this test:

https://github.com/kubernetes/test-infra/blob/2293860ee4926f2c474188805007f71462966edc/config/tests/testgrids/config_test.go#L506-L544

It requires a testgrid.proto file to be generated (using test grid files and prow job files as source), and then looks at the test grid options.  If you invoke this with bazel, that proto file is generated automatically (see https://github.com/kubernetes/test-infra/issues/18599#issuecomment-695762966)

hasheddan:
@spiffxp thanks for the info and apologies for missing this existing test. I'll revert and then update the existing :+1:

PR #19900, opened 2020-11-10
Config-Updater: Remove deprecated namespace and additional_namespaces opts

alvaroaleman:
/assign @fejta


chaodaiG:
Please also update https://github.com/kubernetes/test-infra/blob/6c49475cfc14f8697fb78a6075c5a529ecdd790c/prow/plugins/plugin-config-documented.yaml#L320

PR #19906, opened 2020-11-10
Fix rbac after controller-runtime update

alvaroaleman:
In #19440 controller-runtime was updated, changing its default for the
leader election to be both a configmap (which were never intended to be
used for this) and the lease resource. The latter requires additional
rbac which is added here.

The reason for using two apigroups for this in parallel is to allow safe
transitions. We should after some grace period start using the lease
resource only.

### DIFF
--- a/config/prow/cluster/prow_controller_manager_rbac.yaml
+++ b/config/prow/cluster/prow_controller_manager_rbac.yaml
@@ -25,6 +25,21 @@ metadata:
   name: "prow-controller-manager"
 rules:
 - apiGroups:
+  - coordination.k8s.io
+  resources:
+  - leases
+  resourceNames:
+  - prow-controller-manager-leader-lock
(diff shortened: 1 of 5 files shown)


chaodaiG:
@michelle192837 for awareness

michelle192837 (approved):
/hold

LGTM from my perspective, but holding in case you want review from one of the assigned folks as well. Hold cancel at will if not.

alvaroaleman:
> LGTM from my perspective, but holding in case you want review from one of the assigned folks as well. Hold cancel at will if not

No its fine, its a copy of what we already deployed in Openshift

/hold cancel

alvaroaleman:
@chaodaiG I had to rebase because there was a mergeconflict in ANNOUNCEMENTS.md, could you give a new lgtm?

PR #19913, opened 2020-11-11
Make go autobumper configurable so it can work with more tools

mpherman2:


experiment/autobumper/bumper/bumper.go
```diff
@@ -94,25 +82,32 @@ func (af *fileArrayFlag) Set(value string) error {
 
 // Options is the options for autobumper operations.
 type Options struct {
```
chaodaiG, line 84: Some of the options are optional, or has some other custom restrictions as defined in `validateOptions` function below. So you may want to add `omitempty` on some of the fields

experiment/autobumper/bumper/bumper.go
```diff
@@ -94,25 +82,32 @@ func (af *fileArrayFlag) Set(value string) error {
...
-	ExtraFiles          fileArrayFlag
-
-	SkipPullRequest bool
+	GitHubOrg           string   `yaml:"gitHubOrg"`
+	GitHubRepo          string   `yaml:"gitHubRepo"`
+	GitHubLogin         string   `yaml:"gitHubLogin"`
+	GitHubToken         string   `yaml:"gitHubToken"`
+	GitName             string   `yaml:"gitName"`
+	GitEmail            string   `yaml:"gitEmail"`
+	RemoteBranch        string   `yaml:"remoteBranch"`
```
chaodaiG, line 91: This one could have a default value

experiment/autobumper/bumper/bumper.go
```diff
@@ -94,25 +82,32 @@ func (af *fileArrayFlag) Set(value string) error {
...
+	UpstreamURLBase     string   `yaml:"upstreamURLBase"`
+	IncludedConfigPaths []string `yaml:"includedConfigPaths"`
+	ExcludedConfigPaths []string `yaml:"excludedConfigPaths"`
+	ExtraFiles          []string `yaml:"extraFiles"`
+	TargetVersion       string   `yaml:"targetVersion"`
+	Prefixes            []Prefix `yaml:"prefixes"`
+}
+
+// Prefix is the information needed for each prefix being bumped.
+type Prefix struct {
```
chaodaiG, line 104: Same `omitempty` here

experiment/autobumper/bumper/bumper.go
```diff
@@ -94,25 +82,32 @@ func (af *fileArrayFlag) Set(value string) error {
...
-	BumpBoskosImages bool
-	BumpTestImages   bool
-	TargetVersion    string
-
-	IncludedConfigPaths fileArrayFlag
-	ExcludedConfigPaths fileArrayFlag
-	ExtraFiles          fileArrayFlag
-
-	SkipPullRequest bool
+	GitHubOrg           string   `yaml:"gitHubOrg"`
```
chaodaiG, line 85: Technically in Go these fields should all have comments since they are exposed. The reason why they were not initially, as I guessed, is that they were documented as flag parsing logic, such as:
```
flag.StringVar(&o.GitHubToken, "github-token", "", "The path to the GitHub token file.")
```

I think you also need to copy/paste the comments here as you have deleted all flag parsing

experiment/autobumper/bumper/bumper.go
```diff
@@ -161,36 +156,39 @@ func (gc GitCommand) getCommand() string {
...
-		return fmt.Errorf("at least one of --bump-prow-images, --bump-boskos-images and --bump-test-images must be specified")
-	}
-	if (o.BumpProwImages || o.BumpBoskosImages) && o.BumpTestImages && o.TargetVersion != latestVersion {
-		return fmt.Errorf("--target-version must be latest if you want to bump both prow/boskos and test images")
-	}
-	if o.BumpTestImages && (o.TargetVersion == upstreamVersion || o.TargetVersion == upstreamStagingVersion) {
-		return fmt.Errorf("%q and %q versions can only be specified to bump prow/boskos images", upstreamVersion, upstreamStagingVersion)
+	if o.TargetVersion == upstreamVersion {
+		for _, prefix := range o.Prefixes {
+			if prefix.RefConfigFile == "" || prefix.StagingRefConfigFile == "" {
```
chaodaiG, line 185: Not quite sure that I understand the logic of why StagingRefConfigFile is required for all cases. Could you add document somewhere?

experiment/autobumper/bumper/bumper.go
```diff
@@ -161,36 +156,39 @@ func (gc GitCommand) getCommand() string {
...
+				return fmt.Errorf("target Version can't be 'upstream' without RefConfigFile and StagingRefConfigFile for each prefix")
+			}
+		}
 	}
-
-	if len(o.IncludedConfigPaths) == 0 {
-		return fmt.Errorf("--include-config-paths is mandatory")
+	if o.TargetVersion == upstreamVersion && o.UpstreamURLBase == "" {
```
chaodaiG, line 190: Can we default this UpstreamURLBase as current repo?

chaodaiG (changes requested):
Good work so far! Left a few comments

experiment/autobumper/bumper/bumper.go
```diff
@@ -555,24 +527,38 @@ func HasChanges() (bool, error) {
 	return len(strings.TrimSuffix(string(combinedOutput), "\n")) > 0, nil
 }
 
-func makeCommitSummary(images map[string]string) (string, error) {
-	version, err := getNewProwVersion(images)
-	if err != nil {
-		return "", err
+func getPrefixesString(prefixes []Prefix) string {
```
MushuEE, line 530: worth it to reuse `getAllPrefixes`? 
`strings.Join(getAllPrefixes(prefixes), ", ")`

experiment/autobumper/bumper/bumper.go
```diff
@@ -555,24 +527,38 @@ func HasChanges() (bool, error) {
...
+func makeCommitSummary(prefixes []Prefix, versions map[string][]string, consistency bool) string {
+	if !consistency || len(versions) == 0 {
+		return fmt.Sprintf("Update %s images as necessary", getPrefixesString(prefixes))
+	}
+	res := fmt.Sprintf("Update")
+	for _, prefix := range prefixes {
+		for tag, imageList := range versions {
+			//Should not be possible for tag to be in map with empty imageList
+			if strings.HasPrefix(imageList[0], prefix.Prefix) {
+				res = fmt.Sprintf("%s %s to %s,", res, prefix.Prefix, tag)
```
MushuEE, line 547: This only breaks out of the inner loop correct? So the result might change depending on the order of prefixes. Is that intended?

experiment/autobumper/bumper/bumper.go
mpherman2: It seems that any entry that is not in the yaml is automatically set to the default value even without omitempty so omitempty should not be necessary.

experiment/autobumper/bumper/bumper.go
mpherman2: see above

experiment/autobumper/bumper/bumper.go
```diff
@@ -161,36 +156,39 @@ func (gc GitCommand) getCommand() string {
...
 	if !o.SkipPullRequest && o.RemoteBranch == "" {
-		return fmt.Errorf("--remote-branch cannot be empty when --skip-pull-request is false")
-	}
-	if (o.GitEmail == "") != (o.GitName == "") {
-		return fmt.Errorf("--git-name and --git-email must be specified together")
+		return fmt.Errorf("remote-branch cannot be empty when skip-pull-request is false")
 	}
 
+	if !o.SkipPullRequest && (o.GitHubOrg == "" || o.GitHubRepo == "") {
+		return fmt.Errorf("github-org and github-repo are mandatory when skip-pull-request is false")
```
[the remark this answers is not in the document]
mpherman2: Changed the error message to make this more clear. If skipPullRequest is unspecified it defaults to false

experiment/autobumper/bumper/bumper.go
```diff
@@ -344,48 +342,43 @@ func UpdatePullRequestWithLabels(gc github.Client, org, repo, title, body, match
...
-	}
-	if bumpBoskosImages {
-		filters = append(filters, boskosPrefix)
-	}
-	if bumpTestImages {
-		filters = append(filters, testImagePrefix)
-	}
-	filterRegexp := regexp.MustCompile(strings.Join(filters, "|"))
-
+	filterRegexp := regexp.MustCompile(strings.Join(getAllPrefixes(o.Prefixes), "|"))
```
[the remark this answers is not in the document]
mpherman2: I believe this is correct

experiment/autobumper/bumper/bumper.go
```diff
@@ -433,47 +426,44 @@ func updateReferences(imageBumperCli imageBumper, filterRegexp *regexp.Regexp, t
...
-	}
 	return func(imageHost, imageName, currentTag string) (string, error) {
-		imageFullPath := imageHost + "/" + imageName
-		if strings.HasPrefix(imageFullPath, prowPrefix) {
-			return prowVersion, nil
-		} else if strings.HasPrefix(imageFullPath, boskosPrefix) {
-			return boskosVersion, nil
-		} else {
-			return currentTag, nil
+		imageFullPath := imageHost + "/" + imageName + "/" + currentTag
```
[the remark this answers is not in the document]
mpherman2: Since this is not actually a filepath, and the separator will always be "/", I don't want to use filepath.Join just in case it runs on an OS with a different separator. Was there another reason why I might want to use filepath?

experiment/autobumper/bumper/bumper.go
mpherman2: Good catch!

experiment/autobumper/bumper/bumper.go
```diff
@@ -505,42 +495,24 @@ func isUnderPath(name string, paths []string) bool {
...
-	case 0:
-		return "", nil
-	case 1:
-		for version := range found {
-			return version, nil
+		// If we found more than 1 new version for this prefix, there is a inconsistent prefix
+		if len(versions)-found > 1 {
+			consistent = false
 		}
+		found = len(versions)
```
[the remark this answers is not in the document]
mpherman2: Yea I wrote the code and had to read it twice it to understand what was going on haha. Not sure what I was thinking when I wrote it. Good idea!

experiment/autobumper/bumper/bumper.go
mpherman2: I believe so. This is supposed to generate a string with all of the prefixes that have been bumped and what they have been bumped to. The order of the prefixes should not matter in the commit summary. Is that what you meant?

experiment/autobumper/bumper/bumper.go
chaodaiG: ack

experiment/autobumper/bumper/bumper.go
chaodaiG: Sorry for not making it clear, I meant set the default to `https://raw.githubusercontent.com/kubernetes/test-infra/master` since this is where the tool lives

experiment/autobumper/bumper/bumper.go
chaodaiG: sgtm

experiment/autobumper/bumper/bumper.go
mpherman2: That makes sense. Done

chaodaiG:
Cool, lgtm from my end.
/lgtm
/approve
/hold
In case @MushuEE had more comments

PR #19918, opened 2020-11-11
Remove unused funcs in blunderbuss

hi-rustin:
Remove unused funcs.

Another problem:
Is the `max_request_count` useful? I think maybe it's enough that we have the `request_count`.


hi-rustin:
/cc @fejta 

Is the `max_request_count` useful? I think maybe it's enough that we have the `request_count`?

alvaroaleman (commented):
/ok-to-test

alvaroaleman (approved):

PR #19841, opened 2020-11-04
releng: Teach jobs using latest-fast marker to extract from K8s Infra

justaugustus:
As we continue to migrate release-blocking jobs to a dedicated K8s Infra
cluster, jobs that use the latest-fast marker need to extract builds
from `gs://k8s-release-dev`, which is the K8s Infra equivalent of
`gs://kubernetes-release-dev`.

A new flag (`--extract-ci-bucket=k8s-release-dev`) was added to support
this transitional use case, so we employ it here.

Exceptions:
Of note, the ci-kubernetes-e2e-gce-master-new-gci-kubectl-skew-serial is
not included in this PR.

This job does two extractions:
- `--extract=ci/k8s-stable1`
- `--extract=ci/latest-fast`

As the generic version markers (like `k8s-stable1`) have not been
migrated to K8s Infra, we cannot take advantage of this flag.

We'll plan to fix this job in a follow-up.

(May mitigate https://github.com/kubernetes/test-infra/issues/19838.)

This is part of migrating release-blocking jobs to K8s Infra (ref: https://github.com/kubernetes/test-infra/issues/19484, https://github.com/kubernetes/test-infra/issues/18549).

---

Previous thought process (https://github.com/kubernetes/test-infra/commit/2b282b3c0a01cb02609ed27ac8f73c5d7e3567a8) is explained here:

~This reverts commit f1fd4142d87e877832be19d89037214d917b76aa. (ref: https://github.com/kubernetes/test-infra/pull/19660).~

~tl;dr of what's happening is that any existing job that extracts the `latest-fast` version marker is extracting the marker from `gs://kubernetes-release-dev/ci/fast` (Google Infra) instead of `gs://k8s-release-dev/ci/fast` (K8s Infra), which means they are using stale builds.~

~We need to make an accompanying change to the extract logic, which will take some time to roll out new images for.
This is the quickest course of action in the meantime.~

/assign @cpanato @saschagrunert @hasheddan 
/priority critical-urgent
cc: @kubernetes/sig-scalability @kubernetes/release-engineering

PR #19920, opened 2020-11-11
ci-kubernetes-e2e-ubuntu-gce: Extract latest-fast marker

justaugustus:
Match the version marker extraction choice with
ci-kubernetes-e2e-ubuntu-gce-containerd.

(Also, alpha-sorts the job flags for sanity.)

/assign @hasheddan @dims @andrewsykim 
ref: https://github.com/kubernetes/kubernetes/issues/96463, https://github.com/kubernetes/test-infra/pull/19841

### DIFF
--- a/config/jobs/kubernetes/sig-cloud-provider/gcp/gcp-gce.yaml
+++ b/config/jobs/kubernetes/sig-cloud-provider/gcp/gcp-gce.yaml
@@ -522,7 +522,6 @@ periodics:
           - --
           - --check-leaked-resources
           - --env=ENABLE_POD_SECURITY_POLICY=true
-          - --extract=ci/latest
           - --env=KUBE_MASTER_OS_DISTRIBUTION=ubuntu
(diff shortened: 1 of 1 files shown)


hasheddan (approved):
/lgtm
/hold

I think this is a good idea, but I don't think it will address the issue seen in https://github.com/kubernetes/kubernetes/issues/96463. I think we should move ahead with this change and I'll add more context to why we may be seeing a failure in that issue :+1:

justaugustus:
> I think this is a good idea, but I don't think it will address the issue seen in [kubernetes/kubernetes#96463](https://github.com/kubernetes/kubernetes/issues/96463).

Yep, yep. Mostly opened this for y'all to get a faster feedback loop for debugging any subsequent merges to k/k.

justaugustus:
/hold cancel
ref: https://kubernetes.slack.com/archives/CN0K3TE2C/p1605110336207000?thread_ts=1605107595.204000&cid=CN0K3TE2C

PR #19929, opened 2020-11-12
Remove TaintBasedEvictions from master branch tests

marseel:
Currently kubemark master-branch tests are failing, because TaintBasedEvictions feature gate was removed. This PR fixes it.

### DIFF
--- a/config/jobs/kubernetes/sig-scalability/sig-scalability-presets.yaml
+++ b/config/jobs/kubernetes/sig-scalability/sig-scalability-presets.yaml
@@ -38,11 +38,6 @@ presets:
   # Reduce etcd compaction frequency to match production.
   - name: KUBEMARK_ETCD_COMPACTION_INTERVAL_SEC
     value: "150"
-  # Use Taint based evictions to control hollow node recreation in case of
-  # node VM restart.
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes/sig-scalability/sig-scalability-presets.yaml
```diff
@@ -38,11 +38,6 @@ presets:
   # Reduce etcd compaction frequency to match production.
   - name: KUBEMARK_ETCD_COMPACTION_INTERVAL_SEC
     value: "150"
-  # Use Taint based evictions to control hollow node recreation in case of
-  # node VM restart.
-  # See https://github.com/kubernetes/kubernetes/issues/67120 for context.
-  - name: KUBE_FEATURE_GATES
-    value: "TaintBasedEvictions=true"
```
wojtek-t, line 45: TaintBasedEviction was enabled by default already in 1.16:
https://github.com/kubernetes/kubernetes/blob/release-1.16/pkg/features/kube_features.go#L520

So you can just safely remove it from everywhere - it was a noop for quite some time...

marseel: ah cool, thanks

PR #19934, opened 2020-11-12
Explicitly enable probing container test for dockershim job

hasheddan:
Explicitly enables `[k8s.io] Probing container should be restarted startup probe
fails` to test against a job using dockershim nodes to guard against
failure see in https://github.com/kubernetes/kubernetes/issues/96463.

/cc @andrewsykim @justaugustus 
/assign @derekwaynecarr

### DIFF
--- a/config/jobs/kubernetes/sig-node/sig-node-presubmit.yaml
+++ b/config/jobs/kubernetes/sig-node/sig-node-presubmit.yaml
@@ -31,7 +31,8 @@ presubmits:
         - --node-test-args=--kubelet-flags="--cgroups-per-qos=true --cgroup-root=/"
         - --node-tests=true
         - --provider=gce
-        - --test_args=--nodes=8 --focus="\[NodeConformance\]" --skip="\[Flaky\]|\[Slow\]|\[Serial\]"
+        # Explicitly enable probing container test to make sure we run against dockershim (https://github.com/kubernetes/kubernetes/issues/96463)
+        - --test_args=--nodes=8 --focus="\[NodeConformance\]|\[k8s\.io\] Probing container should be restarted startup probe fails" --skip="\[Flaky\]|\[Slow\]|\[Serial\]"
         - --timeout=65m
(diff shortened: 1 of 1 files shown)


hasheddan:
Note: I think this will cause all PRs to start failing this presubmit until https://github.com/kubernetes/kubernetes/pull/96495 fixes it

config/jobs/kubernetes/sig-node/sig-node-presubmit.yaml
```diff
@@ -31,7 +31,8 @@ presubmits:
         - --node-test-args=--kubelet-flags="--cgroups-per-qos=true --cgroup-root=/"
         - --node-tests=true
         - --provider=gce
-        - --test_args=--nodes=8 --focus="\[NodeConformance\]" --skip="\[Flaky\]|\[Slow\]|\[Serial\]"
+        # Explicitly enable probing container test to make sure we run against dockershim (https://github.com/kubernetes/kubernetes/issues/96463)
+        - --test_args=--nodes=8 --focus="\[NodeConformance\]|\[k8s\.io\] Probing container should be restarted startup probe fails" --skip="\[Flaky\]|\[Slow\]|\[Serial\]"
```
andrewsykim, line 35: Is there a more generic filter we can use here instead of specifcally the startup probe error?

hasheddan: We could test more things, but just went on recommendation from @derekwaynecarr. Might be worth adding a new `[dockershim]` test label which would allow us to be more generic here.. thoughts?

andrewsykim:
> Note: I think this will cause all PRs to start failing this presubmit until kubernetes/kubernetes#96495 fixes it

ack, I think I found the problem, will push a fix soon

andrewsykim:
@hasheddan I think we may not need this PR, I can temporarily add the `[NodeConformance]` tag on the failing tests in my upcoming PR to test this behavior. Those tests should be in node conformance anyways IMO but we can have that discussion after v1.20

hasheddan:
@andrewsykim SGTM

/close

PR #19937, opened 2020-11-13
sig storage: refresh canary images once per week

pohly:
This removes the manually created job in
k8s-staging-sig-storage-periodic.yaml with a generated job in
k8s-staging-sig-storage.sh and also enables the same for all
SIG-Storage repos which build (or will build) images.


pohly:
/assign @msau42 

Does it make sense to have the periodic job for repos which do not build images successfully yet? I'm undecided between "let's add them so that they are there when needed" and "it's just additional noise and work".

chases2:
> Does it make sense to have the periodic job for repos which do not build images successfully yet? I'm undecided between "let's add them so that they are there when needed" and "it's just additional noise and work".

I think I'm on the side of "it's additional noise and work". The churn from failing periodic test that are "supposed to fail" seems worse to me than submitting a PR or rerunning a script.

chases2 (commented):
The change looks good as a whole. Waiting for an LGTM from someone with a bit more context on the specific tests.
/assign @spiffxp 
/approve

config/jobs/image-pushing/k8s-staging-sig-storage.sh
```diff
@@ -74,3 +74,54 @@ for repo in "${REPOS[@]}"; do
               - .
 EOF
 done
+
+cat >>"${OUTPUT}" <<EOF
+
+# Canary images are used by some Prow jobs to ensure that the upcoming releases
+# of the sidecars work together. We don't promote those canary images.
+# To avoid getting them evicted from the staging area, we have to rebuild
+# them periodically. One additional benefit is that build errors show up
```
msau42, line 83: I'm not sure I understand the comment about build errors showing up early? Shouldn't we see build errors in the post submit job whenever a PR merges?

msau42 (commented):
I agree we should not enable the periodic for repos that have not migrated successfully to the new build process.

config/jobs/image-pushing/k8s-staging-sig-storage.sh
pohly: Suppose something changes in the cloud build server or the cloud build image that gets patched into these jobs periodically such that an image build fails. If a repo doesn't merge any PRs for a while, then we won't notice until we try to do a release.

pohly:
I excluded those repos which had failures or no results in https://k8s-testgrid.appspot.com/sig-storage-image-build

PR #19939, opened 2020-11-13
Kubernete-CSI: set snapshot CRD version

pohly:
Some jobs need the "master" version of the snapshotter CRD and
controller. For others it makes sense to select the version that gets
tested in the Prow job instead of having to bump the prow.sh script in
all affected repos.
    
The caveat is that all future versions of external-snapshotter must
install through exactly the YAML files that are currently used by
prow.sh:
https://github.com/kubernetes-csi/csi-release-tools/blob/5d874cce4e649dfd254d01b9b44179ffa72aee75/prow.sh#L702-L722


pohly:
/assign @msau42 
/cc @xing-yang 

Here's a different approach for bumping the CRD version.

If the install URLs are stable across releases, then this should be viable. I'll let you guys decide whether that is acceptable.

config/jobs/kubernetes-csi/csi-driver-nfs/csi-driver-nfs-config.yaml
```diff
@@ -23,56 +23,12 @@ presubmits:
...
+        env:
+        - name: CSI_SNAPSHOTTER_VERSION
+          value: "v3.0.2"
         # docker-in-docker needs privileged mode
         securityContext:
           privileged: true
         resources:
           requests:
             cpu: 2000m
-  - name: pull-csi-driver-nfs-e2e
```
xing-yang, line 32: Is this change required for the snapshot CRD version changes?

config/jobs/kubernetes-csi/external-health-monitor/external-health-monitor-config.yaml
```diff
@@ -23,6 +23,9 @@ presubmits:
         - runner.sh
         args:
         - ./.prow.sh
+        env:
+        - name: CSI_SNAPSHOTTER_VERSION
+          value: "v3.0.2"
```
xing-yang, line 28: Should this be "master"?

xing-yang:
I'm fine with this approach.
@msau42 what do you think?

config/jobs/kubernetes-csi/csi-driver-nfs/csi-driver-nfs-config.yaml
pohly: I had to fix a mistake from someone's prior PR, see the first commit in this PR: 1145b61d36404380468a184c2722363a8908a6fd

config/jobs/kubernetes-csi/external-health-monitor/external-health-monitor-config.yaml
pohly: This job runs with the Kubernetes and sidecar version that is specified inside the prow.sh job, which shouldn't depend on an unreleased snapshotter. So using the fixed version seems fine here.

One could also argue that this particular job shouldn't override `CSI_SNAPSHOTTER_VERSION`.

config/jobs/kubernetes-csi/external-snapshotter/external-snapshotter-config.yaml
```diff
@@ -36,6 +36,8 @@ presubmits:
           value: "1.17"
         - name: CSI_PROW_DRIVER_VERSION
           value: "v1.4.0"
+        - name: CSI_SNAPSHOTTER_VERSION
```
msau42, line 39: does this mean that changes to the snapshot-controller in external-snapshotter prs are untested?

xing-yang: It is tested: https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-csi_external-snapshotter/425/pull-kubernetes-csi-external-snapshotter-1-17-on-kubernetes-1-17/1326879789335711744/artifacts/default/csi-hostpath-snapshotter-0/csi-snapshotter.log

xing-yang: https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-csi_external-snapshotter/425/pull-kubernetes-csi-external-snapshotter-1-17-on-kubernetes-1-17/1326879789335711744/artifacts/default/snapshot-controller-0/snapshot-controller.log

xing-yang: CSI_SNAPSHOTTER_VERSION is set in prow.sh currently.
https://github.com/kubernetes-csi/csi-release-tools/blob/master/prow.sh#L340

msau42: What I mean is that it looks like prow.sh installs a version of the snapshot controller that has already merged and been released? https://github.com/kubernetes-csi/csi-release-tools/blob/master/prow.sh#L722

So if you make changes to snapshot-controller in a PR, then those changes won't get tested as part of the pull job.

xing-yang: Looks like that's the case.  We probably need to replace the snapshot-controller image with the one built from the PR.

pohly: For images used by the csi-driver-host-path deploy script, overriding the images that were built as part of the repo under test is defined here:
https://github.com/kubernetes-csi/csi-release-tools/blob/5d874cce4e649dfd254d01b9b44179ffa72aee75/prow.sh#L1047-L1079

Something similar is needed for the snapshot controller and prow.sh itself.

But this is unrelated to this PR, right?

xing-yang: Yes, I think so.  That will be a separate PR to fix prow.sh.

Issue #18338, opened 2020-07-16
Create a plugin that tests job changes


alvaroaleman:
**What would you like to be added**:

A way to test job changes. Openshift does this via a CLI that checks out the config repo, finds changed jobs and creates a so-called "rehearse" job": https://github.com/openshift/ci-tools/tree/master/cmd/pj-rehearse. What openshift does is highly specific because many jobs there rely on configmaps that then also need to be deployed. This shouldn't be the case for most jobs of other prow instances.

So basically a plugin that on config-repo (test-infra in kube casse) PRs:
* Finds added/changed jobs
* Creates an instance of them and reports that back to the job change pr
* Potentially this needs opt out logic and/or a dedicated command to trigger this
* Some railguards i.E. create at most N rehearse jobs

**Why is this needed**:

Because currently job config changes are untestable if access to the prow.k8s.io build cluster is not given.

/area prow
/cc @BenTheElder @fejta

BenTheElder:
We usually just ask that users create a manually triggered "canary" copy of
the job where risky changes are first trialed for things that depend on CI
resources.

For things that don't there's the local running options (mkpod, pj-on-kind,
phaino ...)

alvaroaleman:
Yeah in my experience jobs that do not actually run in the real cluster are just different. `mkpod` requires to set up a different decoration config or know other intimate details. Many jobs generally assume the presence of certain secrets.

I just ran into the issue with my "increase CPU requests oh damn the cluster doesn't have nodes this size" change and since I am used to getting feedback on job changes, I wanted to leave the idea here.

I am not going to build this, it is just intended as an idea for the ppl that have to regularly deal with the fun of this.

PR #19942, opened 2020-11-13
Create a node e2e testjob with Podutils to validate migration

rikatz:

This will create a new Node e2e job, using podutil.

The Job is optional, and it was created to validate the migration of the critical job to podutils

### DIFF
--- a/config/jobs/kubernetes/sig-node/sig-node-presubmit.yaml
+++ b/config/jobs/kubernetes/sig-node/sig-node-presubmit.yaml
@@ -41,6 +41,49 @@ presubmits:
           requests:
             cpu: 4
             memory: 6Gi
+  - name: pull-kubernetes-node-e2e-podutil
+    skip_branches:
+    - release-\d+\.\d+  # per-release image
+    annotations:
+      fork-per-release: "true"
+      testgrid-alert-stale-results-hours: "24"
+      testgrid-create-test-group: "true"
+      testgrid-num-failures-to-alert: "10"
+    always_run: false
+    decorate: true
+    cluster: k8s-infra-prow-build
+    max_concurrency: 12
+    labels:
+      preset-service-account: "true"
+      preset-k8s-ssh: "true"
+    spec:
+      containers:
+      - image: gcr.io/k8s-testimages/kubekins-e2e:v20201111-a263fd7-master
+        command:
+        - runner.sh
+        - kubetest
+        args:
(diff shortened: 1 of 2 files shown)


chaodaiG:
/cc @BenTheElder @krzyzacy 
For help on kubetest

config/jobs/kubernetes/sig-node/sig-node-presubmit.yaml
```diff
@@ -9,6 +9,10 @@ presubmits:
       testgrid-create-test-group: "true"
       testgrid-num-failures-to-alert: "10"
     always_run: true
+    decorate: true
+    extra_refs:
```
alvaroaleman, line 13: presubmits don't need the extra_refs for the repo they run on and I think adding it will result in undefined behavior

rikatz: Great, thanks! Removing!

rikatz: Removed.

@alvaroaleman if this is right, I can migrate all the jobs from sig-node here to podsubmits (it seems pretty 'trivial', as they use node configs).

Just don't know if there's a way to test like it is in production without actually merging it

alvaroaleman: to my knowledge its not possible to test job changes on prow.k8s.io without actually merging them

alvaroaleman: (You could copy the job, set optional: true on the copy, create a PR on k/k and check, then remove the copy and update the original here)

alvaroaleman: Yeah, seems that's the only way: https://github.com/kubernetes/test-infra/issues/18338#issuecomment-659530553

rikatz: Damn, right then :)

So I can just migrated the required ``pull-kubernetes-node-e2e`` to not have a big change.

I've followed your idea and created a similar job, with other name to test if it works :)

If this is ok we can follow this way and then I can start migrating some other jobs to podutils

chaodaiG:
/lgtm
/approve
Looks good to me generally, since it's not blocking other PRs, let's try it out

PR #19946, opened 2020-11-13
Removing helm/charts from test-infra

mattfarina:
The charts repo is no longer supported. It no longer gets new
contributions. So, this change is to remove test-infra config
that has been used to support the charts repo. See
https://helm.sh/blog/helm-2-becomes-unsupported/


scottrigby:
Cross-linking https://github.com/helm/charts/issues/23944

chases2 (approved):

Issue #19754, opened 2020-10-29
History page times out for certain jobs


alvaroaleman:
**What happened**:

Opening the job history for certain jobs results in a timeout: https://prow.ci.openshift.org/job-history/gs/origin-ci-test/logs/release-openshift-origin-installer-e2e-aws-upgrade

While it works for others: https://prow.ci.openshift.org/job-history/gs/origin-ci-test/logs/release-openshift-ocp-installer-e2e-aws-ovn-4.7

It appears that we might not properly limit the amount of data we collect from gcs

/area prow
/area prow/deck

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

**Please provide links to example occurrences, if any**:

**Anything else we need to know?**:

PR #19950, opened 2020-11-13
Deck job history: Limit time to find all builds to 10s

alvaroaleman:
Fixes https://github.com/kubernetes/test-infra/issues/19754


prow/cmd/deck/job_history.go
```diff
@@ -231,12 +232,9 @@ func (bucket blobStorageBucket) listAll(ctx context.Context, prefix string) ([]s
 
 // Gets all build ids for a job.
 func (bucket blobStorageBucket) listBuildIDs(ctx context.Context, root string) ([]int64, error) {
-	ids := []int64{}
+	var ids []int64
 	if strings.HasPrefix(root, logsPrefix) {
 		dirs, err := bucket.listSubDirs(ctx, root)
```
MushuEE, line 237: Feels like it is against the Go patterns to have an `err` that only stays in scope on like 248 when dirs happens to be `nil` and `err` is not redefined. However I am still getting to know Go!

prow/cmd/deck/job_history.go
```diff
@@ -434,8 +435,11 @@ func getJobHistory(ctx context.Context, url *url.URL, cfg config.Getter, opener
 		tmpl.LatestLink = linkID(url, emptyID)
 	}
 
-	buildIDs, err := bucket.listBuildIDs(ctx, root)
-	if err != nil {
+	// Don't spend an unbound amount of time finding a potentially huge history
+	buildIDListCtx, cancel := context.WithTimeout(ctx, 10*time.Second)
```
MushuEE, line 439: Is this best effort or will timeouts produce no results?

prow/cmd/deck/job_history.go
alvaroaleman: Deferring error checking like this is absolutely not idiomatic go, but unavoidable here, otherwise we will always either get all build ids successful (which for example for the job linked in the issue report takes multiple minutes) or return none. That's why I added the testcase, otherwise the next person will move that code back to where it was.

The redefining/shadowing is indeed an issue though, good catch.

prow/cmd/deck/job_history.go
alvaroaleman: Best effort, that is why the other change was required. It seems there is no way to limit the number of result, the timeout seems to be the only way to somehow limit this.

prow/cmd/deck/job_history.go
alvaroaleman: Updated to rename the `err` here to `listErr` to avoid the shadowing

cjwagner:
Is updating cloud.google.com/go/storage to 1.12.0 necessary for this effort? It doesn't look like the functions are used any differently.
I ask because we have been having issues with crier on one of our internal instances and just determined that crier has been consuming way too much memory and running far too many goroutines. Most of the threads are stuck waiting for a semaphore in a grpc callstack, but it isn't clear what it causing the grpc threads to be created. I'm not aware of anywhere we use grpc directly so I'm guessing this is a dependency. Memory usage graphs show that the problem began after bumping Prow with this change diff: https://github.com/kubernetes/test-infra/compare/dbd60ab192...dbc92a6558 I don't see how any of the commits in that range could impact crier besides the one in this PR that updates the storage client and grpc.

If it isn't inconvenient it'd be nice to revert the version update to see if that was the culprit. This has repeatedly impacted the affected Prow instance, but none of our other instances are misbehaving or consuming excess memory.

alvaroaleman:
@cjwagner no it is not required, I bumped it because it adds a startAt or something like that property to the list call, but I ended up not using that so feel free to revert it back

PR #19945, opened 2020-11-13
feat: Add milestone applier

irvifa:
This is required for milestone applier to work on k/website.

PR #19952, opened 2020-11-14
fix: Add prefix v

irvifa:
Possible follow up of https://github.com/kubernetes/test-infra/pull/19945

### DIFF
--- a/config/prow/plugins.yaml
+++ b/config/prow/plugins.yaml
@@ -352,7 +352,7 @@ milestone_applier:
     release-0.4: v0.4
     release-0.3: v0.3
   kubernetes/website:
-    dev-1.20: 1.20
+    dev-1.20: v1.20
 
 repo_milestone:
   # Default maintainer


irvifa:
https://github.com/kubernetes/website/issues/24960

irvifa:
I understand that we can use milestone applier for release against certain branch in various repo but it seems like it doesn’t applied for my PR.
https://github.com/kubernetes/website/pull/25033
While we already added the config in:
https://github.com/kubernetes/test-infra/blob/f09ff0e8053579b045424785867bc2a80a4b963a/config/prow/plugins.yaml#L355
https://github.com/kubernetes/test-infra/blob/f09ff0e8053579b045424785867bc2a80a4b963a/config/prow/plugins.yaml#L801

Also based on https://github.com/kubernetes/test-infra/blob/ea76c130fa6d97b07aeceb817d9d1016c2c7338e/prow/plugins/milestoneapplier/milestoneapplier.go#L101-L110 the PR I opened should be marked with milestone 1.20. However it doesn't. When i take a look into other format details it seems like the milestones is always have the prefix of `v`.

hasheddan (approved):
/lgtm

annajung:
@irvifa It wouldn't work because there is no `v` in milestones under `k/website`.
I think it might be better to file an issue instead of changing the config to match a milestone that doesn't exist.

hasheddan:
@annajung I believe you can change the name of the milestone on k/website and it will still keep all those that have been manually added

annajung:
Others have `v` prefixed because other repos milestones are all prefixed with `v`. 

Also [linking a conversation](https://kubernetes.slack.com/archives/C01672LSZL0/p1604609283077400) I had with @spiffxp  about the milestone applier

annajung:
hey @hasheddan I did not change the [k/website milestone](https://github.com/kubernetes/website/milestones) for this release. I however brought it up and is being discussed w/ sig docs to change it to match other k repos, but no action was taken on this.

hasheddan:
@annajung gotcha 👍 thanks for the clarity and apologies for the preemptive approval here! Sounds like we can just update appropriately next cycle

/hold

annajung:
Just to clarify, we can't simply change the milestone because of all the docs that are currently referencing the milestone. I think a bit more discussion with Sig Docs are required for us to prefix the milestone with `v`. But, I don't think the milestone applier functionality should work regardless, therefore, I propose filing an issue instead.

annajung:
@irvifa feel free to assign me to the issue, I can help take a look! :)

irvifa:
@annajung I see.. I also have a conversation with @justaugustus about this, since dev-1.20 is not default branch and when a PR opened against that branch the plugins doesn’t work as I expect it to be. Wondering if I missed something. Afaik milestonestatus is deprecated as well so the only difference we have with other repos using milestoneapplier is only override plugins but I’m not sure either. 🤔

hasheddan:
> Just to clarify, we can't simply change the milestone because of all the docs that are currently referencing the current format of the milestone.

@annajung  Not sure if this is what you are saying, but if you change the milestone name (i.e. `1.20` --> `v1.20`) all the issues and PRs that were in `1.20` will automatically be in `v1.20`. Not sure how prow will react to the `/milestone 1.20` comments that already exist, but I am guessing they would just be ignored. In addition, any docs that link to the milestone should automatically go to `v1.20` as the name is not the unique identifier of a milestone ([https://github.com/kubernetes/website/milestone/46](https://github.com/kubernetes/website/milestone/46)). Anyway, will certainly leave this up to you, but just wanted to make that known just in case :)

annajung:
I think we might be able to help either fix this bug or extend this functionality to make sure that non default branches work as well? 

We can wait to hear more feedback from @justaugustus and @spiffxp

annajung:
Thanks @hasheddan , I think Sig Docs are more worried about the consistency across all docs that reference the milestone. I think there are places that need to be manually changed and they wanted to make sure that when the milestone format changes, all existing docs/reference to the milestone changes with it! :)

Nothing against prefixing `v` in the. milestone, I think the idea is to make sure all other docs change with it.

PR #19953, opened 2020-11-14
Remove DEPRECATED configuration from updateconfig README.md

hi-rustin:
The DEPRECATED configuration already removed by https://github.com/kubernetes/test-infra/commit/62b5ded808847f7ae5ec0a7a14d60bdfd89270e7.

So we need to update the README.md

### DIFF
--- a/prow/plugins/updateconfig/README.md
+++ b/prow/plugins/updateconfig/README.md
@@ -7,7 +7,7 @@
 ## Usage
 
 Update your `plugins.yaml` file to something along the following lines:
-```
+```yaml
 plugins:
   my-github/repo:
   - config-updater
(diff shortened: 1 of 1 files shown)


prow/plugins/updateconfig/README.md
```diff
@@ -17,17 +17,14 @@ config_updater:
     # Update the thing-config configmap whenever thing changes
     path/to/some/other/thing:
       name: thing-config
-      # Using ProwJobNamespace by default.
+    # Specify the clusters and namespaces
```
alvaroaleman, line 20: Why is this comment here when there is no cluster and namespace specification afterwards?
Please instead mention that if cluster and namespace configuration are unset, it will be put into the `default` cluster in the `prowjob` namespace

prow/plugins/updateconfig/README.md
```diff
@@ -17,17 +17,14 @@ config_updater:
...
-      # DEPRECATED: Please use "clusters" below
-      namespace: otherNamespace
-      # specify the clusters and namespaces that the configmap targets
+      # Specify the clusters and namespaces that the configmap targets
       # which requires that the --kubeconfig arg is enabled for Hook
       # https://github.com/kubernetes/test-infra/blob/master/prow/getting_started_deploy.md#run-test-pods-in-different-clusters
       # if not set or empty, it uses the cluster where prow components are running
       # and the specified namespace(s)
-      # if defined, the above namespace will be ignored
       clusters:
```
hi-rustin, line 28: @alvaroaleman  We set up clusters and namespaces here.

prow/plugins/updateconfig/README.md
alvaroaleman: > @alvaroaleman We set up clusters and namespaces here.

Yes but you are removing the information that this is getting defaulted and the `cluster and namespaces` are optional config options: https://github.com/kubernetes/test-infra/blob/b2cf66401ce642c5e726fdb5cfd1d43eb67a88be/prow/plugins/config.go#L844-L846

hi-rustin: Replaced. Thanks!

alvaroaleman (approved):

PR #19956, opened 2020-11-15
avoid "Failed to clean up repo" by disabling git-gc

### LINKED ISSUE #18768: Recurring prow log errors "Failed to clean up repo"
**What happened**: istio prow generates lots of recurring errors with error:
```
jsonPayload: {
  component: "hook"   
  error: "unlinkat /tmp/git236299396/.git: directory not empty"   
  file: "prow/config/inrepoconfig.go:76"   
  func: "k8s.io/test-infra/prow/config.defaultProwYAMLGetter.func1"   
  level: "error"   
  msg: "Failed to clean up repo."   
  repo: "istio/istio"   
 }
```

**What you expected to happen**: no errors

Check https://pantheon.corp.google.com/logs/viewer?project=istio-testing&minLogLevel=0&expandAll=false&timestamp=2020-08-10T18%3A45%3A25.475000000Z&customFacets=jsonPayload.level%2CjsonPayload.component%2CjsonPayload.msg&limitCustomFacetWidth=true&interval=CUSTOM&resource=container%2Fcluster_name%2Fprow&scrollTimestamp=2020-08-10T18%3A40%3A36.000000000Z&pli=1&rapt=AEjHL4OIudnQvHfkwsjeHVKiuCNtyCK-tlShRt38O1NCSXeK1nyc4TYLdL8D7C7VIIxwkX4_JCEZzEoIjZvVAJUWcTgRZHSklw&advancedFilter=resource.type%3D%22container%22%0Aresource.labels.cluster_name%3D%22prow%22%0Aresource.labels.zone%3D%22us-west1-a%22%0Aresource.labels.pod_id%3D%22hook-7f87496b48-qb29h%22%0Aresource.labels.project_id%3D%22istio-testing%22%0Aresource.labels.container_name%3D%22hook%22%0Aresource.labels.instance_id%3D%22509889749234183265%22%0Aresource.labels.namespace_id%3D%22default%22%0Atimestamp%3D%222020-08-10T18%3A40%3A36.000000000Z%22%0AinsertId%3D%2210x1srgg2bi8ijd%22&dateRangeStart=2020-08-07T18%3A41%3A18.000Z&dateRangeEnd=2020-08-10T18%3A41%3A18.000Z

The error comes from https://github.com/kubernetes/test-infra/blob/7ada03969d444ea084935fd465e50605b45d9284/prow/config/inrepoconfig.go#L74, as a best effort cleanup, maybe this should just be a `debug` logging?

matthyx:
relates to #18768
@alvaroaleman do you think it's a good idea to silence that error?

### DIFF
--- a/prow/git/git.go
+++ b/prow/git/git.go
@@ -184,7 +184,7 @@ func (c *Client) Clone(organization, repository string) (*Repo, error) {
 	if b, err := exec.Command(c.git, "clone", cache, t).CombinedOutput(); err != nil {
 		return nil, fmt.Errorf("git repo clone error: %v. output: %s", err, string(b))
 	}
-	return &Repo{
+	r := &Repo{
 		dir:    t,
 		logger: c.logger,
 		git:    c.git,
@@ -194,7 +194,12 @@ func (c *Client) Clone(organization, repository string) (*Repo, error) {
 		repo:   repository,
 		user:   user,
 		pass:   pass,
-	}, nil
+	}
+	// disable git GC
+	if err := r.Config("gc.auto", "0"); err != nil {
+		return nil, err
+	}
+	return r, nil
 }
 
 // Repo is a clone of a git repository. Create with Client.Clone, and don't


alvaroaleman:
> @alvaroaleman do you think it's a good idea to silence that error?

Not sure, from a quick look it seems we log it on error-level everywhere we log it. The error seems to indicate that something was written in that dir after `.Clean()` was called which seems like a bug

chaodaiG:
> > @alvaroaleman do you think it's a good idea to silence that error?
> 
> Not sure, from a quick look it seems we log it on error-level everywhere we log it. The error seems to indicate that something was written in that dir after `.Clean()` was called which seems like a bug

The same error is not appearing on tide, And the error is very similar:
```
jsonPayload: {
func: "k8s.io/test-infra/prow/config.defaultProwYAMLGetter.func1"
msg: "Failed to clean up repo."
error: "unlinkat /tmp/git367912785/.git: directory not empty"
repo: "istio/istio"
component: "tide"
level: "error"
file: "prow/config/inrepoconfig.go:76"
}
```

Could it be tide and hook both try to access the same directory?

alvaroaleman:
> Could it be tide and hook both try to access the same directory?

Hook and tide run in different pods and hence have different mount namespaces. Apart from that, the client creates a distinct temp dir each time `Clone` is called: https://github.com/kubernetes/test-infra/blob/91b8575a1f1300e0c3b612d466a880d3141c34e3/prow/git/git.go#L180-L188

My first theory here was that we might incorrectly stack defers and close the file after trying to clean, but that doesn't seem to be the case, we only access the file via `ioutil.ReadFile` which internally closes it. Does `git` maybe launch background tasks sometimes?

alvaroaleman:
https://git-scm.com/docs/git-gc#Documentation/git-gc.txt-gcauto seems like it could cause background tasks to be run. Obviously, this is pure speculation but I do not see how our code could cause something like this.

matthyx:
So then maybe it's fine to silence it?

alvaroaleman:
I would rather disable the git garbage collection to prove the theory. I don't think it provides any value for the very shortlived mirror clones we create via the git client.

chaodaiG:
> I would rather disable the git garbage collection to prove the theory. I don't think it provides any value for the very shortlived mirror clones we create via the git client.

I kind of agree both :)

This certainly is a best effort clean up and the the failure here doesn't alter any outcome. On the other hand, I'm also very interested in why this happened. 

So why not do both?

matthyx:
Now that I think of it, do we need the clone at all? In many of my CI builds it's sufficient to download the tar archive at the reference, no need for the history...

alvaroaleman:
> Now that I think of it, do we need the clone at all? In many of my CI builds it's sufficient to download the tar archive at the reference, no need for the history...

The problem with that is that it counts towards the api token budget, the `git clone` does not.

matthyx:
Damn

matthyx:
@chaodaiG @alvaroaleman I have added a call to git config just before the actual cleanup, I suggest we keep the error log as it is and check if this modification helps.

chaodaiG:
Sgtm

alvaroaleman:
> @chaodaiG @alvaroaleman I have added a call to git config just before the actual cleanup, I suggest we keep the error log as it is and check if this modification helps.

The way I understand [the docs on this](https://git-scm.com/docs/git-gc#Documentation/git-gc.txt-gcauto) this will kick in immediately after some thresholds are met unless disabled. So I would think that we have to set this globally asap (unless git actually stops a running gc when you disable the auto functionality, but that would surprise me tbh)

alvaroaleman (approved):

alvaroaleman:
Thanks!

PR #19959, opened 2020-11-16
Cleanup density test

marseel:
Cleans up all remaining tests that contain density job.


mborsz:
/lgtm

Looks good to me, but I would like the current oncall to review this as well.
/assign @jkaniuk

Issue #19961, opened 2020-11-16
Kettle entering CrashLoopBackOff with "Bad GCS path" 


MushuEE:
**What happened**:
11/16 Checked on state of Kettle, Looks like starting 11/11 errors started occurring in `make_db.py`
![image](https://user-images.githubusercontent.com/7189267/99285546-8447df80-27ec-11eb-818c-fbc1f2c14ac8.png)
```
"20.16user 18.85system 0:37.85elapsed 103%CPU (0avgtext+0avgdata 1465472maxresident)k
"
```
```
Traceback (most recent call last):
  File "make_db.py", line 346, in <module>
    OPTIONS.junit)
  File "make_db.py", line 310, in main
    threads, client_class)
  File "make_db.py", line 246, in get_builds
    for n, (build_dir, started, finished) in enumerate(builds_iterator):
  File "/usr/lib/python3.6/multiprocessing/pool.py", line 735, in next
    raise value
ValueError: Bad GCS path
Command exited with non-zero status 1
20.74user 22.63system 0:42.67elapsed 101%CPU (0avgtext+0avgdata 1465524maxresident)k
0inputs+88outputs (0major+603132minor)pagefaults 0swaps
+ time python3 make_db.py --buckets buckets.yaml --junit --threads 32
Traceback (most recent call last):
  File "/kettle/update.py", line 59, in <module>
    main()
  File "/kettle/update.py", line 29, in main
    call('time python3 make_db.py --buckets buckets.yaml --junit --threads 32')
  File "/kettle/update.py", line 25, in call
    raise OSError('invocation failed')
OSError: invocation failed
```

Unsure if this is a CPU limit/memory issue or if it is another failure.

**What you expected to happen**:
make_db to pass successfully 

**How to reproduce it (as minimally and precisely as possible)**:
Start kettle prod container. 

**Please provide links to example occurrences, if any**:
https://console.cloud.google.com/logs/query;query=resource.type%3D%22k8s_container%22%0Aresource.labels.cluster_name%3D%22g8r%22%0Aresource.labels.namespace_name%3D%22default%22%0Aresource.labels.container_name%3D%22kettle%22%0A%22%2B%22;timeRange=2020-11-08T17:10:19.418Z%2F2020-11-16T17:10:19.418215Z;summaryFields=true:true:32:beginning?scrollTimestamp=2020-09-08T13:49:38.595881274Z&dateRangeUnbound=both&project=k8s-gubernator
**Anything else we need to know?**:

[...]

PR #19962, opened 2020-11-16
[Kettle] Extend error message with information on unfound path

MushuEE:
#19961 is causing failures in Kettle. There is little context around the path it was trying.

### DIFF
--- a/kettle/make_db.py
+++ b/kettle/make_db.py
@@ -71,7 +71,7 @@ class GCSClient:
     @staticmethod
     def _parse_uri(path):
         if not path.startswith('gs://'):
-            raise ValueError("Bad GCS path")
+            raise ValueError("Bad path: %s, should start with 'gs://" % path)
         bucket, prefix = path[5:].split('/', 1)
         return bucket, prefix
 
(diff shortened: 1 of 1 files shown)


amwat:
```
ERROR:root:failed to get tests for kubeflow-pipeline-backend-test/1326401600935170048
```

https://prow.k8s.io/job-history/gs/kubernetes-jenkins/pr-logs/directory/kubeflow-pipeline-backend-test?buildId=1326401600935170048

that run definitely seems erroneous in general
```
started at: 0001-01-01 00:00:00 +0000 UTC
```


/lgtm

MushuEE:
Bad path it is looking for is `kubeflow-pipeline-backend-test/*` it is for some reason not prepended with the `gs://` parts. 
I do not know what caused this.
Workers are attempting to run: `WORKER_CLIENT.get_started_finished(job, build)` 
Which should make the build_dir `self.get('%s/directory/%s/%s.txt' % (self.jobs_dir, job, build)).strip()` this implies that there is no `gs://` in jobs_dir

here job=`kubeflow-pipeline-backend-test` and build = `some number`

This behavior implies `build_dir` is not correct or nil which can be caused by `if 400 <= resp.status_code < 500 and resp.status_code != 429:`

PR #19963, opened 2020-11-16
Fix NPD batch merging.

Random-Liu:
Use `batch` for batch merging.

### DIFF
--- a/config/jobs/kubernetes/node-problem-detector/node-problem-detector-presubmits.yaml
+++ b/config/jobs/kubernetes/node-problem-detector/node-problem-detector-presubmits.yaml
@@ -79,7 +79,7 @@ presubmits:
         - bash
         - -c
         - >-
-          ./test/build.sh -p $(PULL_NUMBER) pr &&
+          ./test/build.sh -p $(BUILD_ID) pr &&
           source pr.env &&
           cd $(GOPATH)/src/k8s.io/kubernetes &&
           /workspace/scenarios/kubernetes_e2e.py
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes/node-problem-detector/node-problem-detector-presubmits.yaml
```diff
@@ -79,7 +79,7 @@ presubmits:
         - bash
         - -c
         - >-
-          ./test/build.sh -p $(PULL_NUMBER) pr &&
+          ./test/build.sh -p ${PULL_NUMBER:-batch} pr &&
```
BenTheElder, line 82: I think this is only safe if we make the (bad) assumption that there's only one batch at a time ever.
https://github.com/kubernetes/node-problem-detector/blob/35bfe697a5fd9b058f86a3fb704b7e6d7b68457c/test/build.sh#L130

Maybe e.g. BUILD_ID

BenTheElder:
```suggestion
          ./test/build.sh -p ${BUILD_ID} pr &&
```

Random-Liu: Done

Random-Liu: Thanks for the suggestion!

wangzhen127:
Probably update the PR description.

/lgtm
/hold

Unhold when the NPD side change is ready.

Random-Liu:
@wangzhen127 We don't need to change NPD side. This should just work. :)

The NPD side is more a cleanup or an improvement.

wangzhen127:
Ah, I see.

/unhold

Issue #19629, opened 2020-10-20
Add configuration for prow plugin's respond


hi-rustin:
**What would you like to be added**:
Could we add some configurations to the https://github.com/kubernetes/test-infra/blob/6ad9eb5f38fc7ca9867b2700f1eced5a2e5454ff/prow/plugins/respond.go#L30

and 

https://github.com/kubernetes/test-infra/blob/6ad9eb5f38fc7ca9867b2700f1eced5a2e5454ff/prow/plugins/respond.go#L27

**Why is this needed**:

Because I want to use these plugins in my project or community, but we have different links and documentation. So I hope it can be a config.

If this is possible, I would be happy to help.

hi-rustin:
@BenTheElder  @fejta @stevekuznetsov  Do you have any suggestions?

BenTheElder:
Some similar messages have been made configurable before but the decision is up to the prow owners https://github.com/kubernetes/test-infra/blob/master/prow/OWNERS

matthyx:
@hi-rustin this is not quite as easy as it sounds...
These 2 are consts defined for the whole prow plugins, so if you want to make them configurable they need to be parameterized in `test-infra/prow/plugins/config.go` with a default value reflecting the current links.
As @BenTheElder said, some plugins have this link configurable (like `approve`) and therefore you will have to migrate by supporting both configs for a while and then finally only use the common one.
You can work on that if you want, I can help you.
/assign @hi-rustin

BenTheElder:
https://github.com/kubernetes/test-infra/pull/19965 is this sufficient?

hi-rustin:
> #19965 is this sufficient?

It feels like all the plugins have to be modified, and I want to put the configuration in config and pass it to each plugin to use. I don't know if this is possible?

BenTheElder:
I think the ones in the issue body were done, as for others I don't see why not but I won't personally be approving them.

hi-rustin:
@fejta @alvaroaleman  What do you think?

hi-rustin:
> as for others I don't see why not but I won't personally be approving them (see prow OWNERS)

Because contributors may open detail, they will be directed to https://go.k8s.io/bot-commands. But may actually use not k8s prow but their own deployment.

PR #19965, opened 2020-11-17
Add HelpGuidelinesURL configuration for help

hi-rustin:
Add `HelpGuidelinesURL` configuration for help plugin.

### DIFF
--- a/prow/plugins/config.go
+++ b/prow/plugins/config.go
@@ -59,7 +59,6 @@ type Configuration struct {
 	Owners Owners `json:"owners,omitempty"`
 
 	// Built-in plugins specific configuration.
-
 	Approve              []Approve                    `json:"approve,omitempty"`
 	Blockades            []Blockade                   `json:"blockades,omitempty"`
 	Blunderbuss          Blunderbuss                  `json:"blunderbuss,omitempty"`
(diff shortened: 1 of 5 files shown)


matthyx:
@hi-rustin I think it's similar to https://github.com/kubernetes/test-infra/issues/19629 we somehow should make this global to prow instead of plugin specific...

I would make a global configuration listing all urls that should be configurable (and potentially shared between plugins) in prow installations.

Any thoughts @alvaroaleman ?

hi-rustin:
@alvaroaleman  Could you please take a look?

hi-rustin:
@matthyx  @alvaroaleman  ping~

hi-rustin:
> I would make a global configuration listing all urls that should be configurable (and potentially shared between plugins) in prow installations.

I think some generic configurations can be put into the global configuration share, but such links that are only relevant to a specific plugin can be put into the plugin's own configuration.

prow/plugins/config.go
```diff
@@ -85,6 +84,19 @@ type Configuration struct {
...
+}
+
+type Help struct {
+	// HelpGuidelinesURL is the URL of the help page, which provides guidance on how and when to use the help wanted and good first issue labels.
+	// The default value is "https://git.k8s.io/community/contributors/guide/help-wanted.md".
+	HelpGuidelinesURL string `json:"help_guidelines_url,omitempty"`
+}
+
+func (h *Help) SetDefaults() {
+	if len(h.HelpGuidelinesURL) == 0 {
```
fejta, line 97: `if h.HelpGuidelinesURL == "" {`

prow/plugins/config.go
```diff
@@ -85,6 +84,19 @@ type Configuration struct {
...
+	Help                 Help                         `json:"help,omitempty"`
+}
+
+type Help struct {
+	// HelpGuidelinesURL is the URL of the help page, which provides guidance on how and when to use the help wanted and good first issue labels.
+	// The default value is "https://git.k8s.io/community/contributors/guide/help-wanted.md".
+	HelpGuidelinesURL string `json:"help_guidelines_url,omitempty"`
+}
+
+func (h *Help) SetDefaults() {
```
fejta, line 96: Who do we need to export this to?

fejta (approved):
/hold

thanks!

prow/plugins/config.go
hi-rustin: Fixed.

prow/plugins/config.go
hi-rustin: Fixed.

PR #19966, opened 2020-11-17
Replace SingleCommit with RepositoryCommit

hi-rustin:
Because `RepositoryCommit` also meets the structural requirements in the https://developer.github.com/v3/repos/commits/#get-a-single-commit link.  So I think we should replace it with  RepositoryCommit.


hi-rustin:
@alvaroaleman  Could you please take a look?

petr-muller (approved):
Looks safe according to the docs

alvaroaleman (approved):

PR #19969, opened 2020-11-17
add sctp periodic job for azure

aramase:
- Adds periodic job for testing SCTP feature

### DIFF
--- a/config/jobs/kubernetes/sig-cloud-provider/azure/sig-azure-config.yaml
+++ b/config/jobs/kubernetes/sig-cloud-provider/azure/sig-azure-config.yaml
@@ -363,3 +363,54 @@ periodics:
     testgrid-alert-email: kubernetes-provider-azure@googlegroups.com
     description: Runs conformance & node conformance tests latest kubernetes master with cluster-api-provider-azure
     testgrid-num-columns-recent: '30'
+- interval: 24h
+  name: ci-kubernetes-e2e-sctp-master
+  decorate: true
+  decoration_config:
+    timeout: 3h
+  labels:
+    preset-service-account: "true"
+    preset-azure-cred: "true"
+    preset-dind-enabled: "true"
+  extra_refs:
+  - org: kubernetes
+    repo: kubernetes
+    base_ref: master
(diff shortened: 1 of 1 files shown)


aramase:
/hold
(finishing up testing locally)

cc @aojea

config/jobs/kubernetes/sig-cloud-provider/azure/sig-azure-config.yaml
```diff
@@ -363,3 +363,54 @@ periodics:
     testgrid-alert-email: kubernetes-provider-azure@googlegroups.com
     description: Runs conformance & node conformance tests latest kubernetes master with cluster-api-provider-azure
     testgrid-num-columns-recent: '30'
+- interval: 24h
+  name: ci-kubernetes-e2e-sctp-master
```
aojea, line 367: sctp does not have a large amount of tests, maybe you want to run a "full" e2e sig-network suite, you can see an example here
https://testgrid.k8s.io/sig-network-kind#sig-network-kind,%20master
https://github.com/kubernetes/test-infra/blob/9828977f38e50bcb800629aee9301f681cb9b06a/config/jobs/kubernetes/sig-network/sig-network-kind.yaml#L140-L143

aramase: @aojea The reason was for just confining to small subset was because of this comment in the KEP - 
```
These will be tagged "[Feature:SCTPConnectivity]". They can be run by network plugin developers but will not run as part of any of the standard Kubernetes test runs. They should all be marked [Disruptive] since they may cause the SCTP kernel module to be loaded, which may interfere with the Basic Tests' "make sure the SCTP kernel module doesn't get loaded" checks.
```

I'll change this to be similar to the kind network tests

aojea: just try an error, IIRC that is because you can run SCTP in userspace or in kernel, and  the tests that use the userspace SCTP check if the module was loaded ...  just do whatever is simpler for you ;)

PR #19973, opened 2020-11-17
[Kettle] Clean up some logging 

MushuEE:
This is just an intermediate cleanup while I lock in some other fixes related to #19961
/area kettle


amwat (commented):
/lgtm

PR #19974, opened 2020-11-18
[Kettle] Add in seconds to tag

MushuEE:
If you are testing and building multiple images in succession, they will overwrite pervious images from the same day

Should I use %s which is more granular?
/area kettle

### DIFF
--- a/kettle/Makefile
+++ b/kettle/Makefile
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 IMG = gcr.io/k8s-testimages/kettle
-TAG := $(shell date +v%Y%m%d)-$(shell git describe --tags --always --dirty)
+TAG := $(shell date +v%Y%m%d%S)-$(shell git describe --tags --always --dirty)
 
 # These are the usual GKE variables.
 PROJECT       ?= k8s-gubernator


kettle/Makefile
```diff
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 IMG = gcr.io/k8s-testimages/kettle
-TAG := $(shell date +v%Y%m%d)-$(shell git describe --tags --always --dirty)
```
amwat, line 16: the TAG also includes the commit so even if you build many images they would have different commits right?
unless you are building and pushing from local changes in which case they would have the dirty tag and fine to be overwritten.

MushuEE: The image will be overwritten but if you try to re-pull from `kubectl set image` it will think it already has the correct image. Is there a force?

MushuEE: `imagePullPolicy: Always` does not seem to impact `set image`

amwat: not sure what you mean, wouldn't a new image have a new tag if it's based off of a new commit?
why will it think it already has the correct image?

MushuEE: Here is the process that was bugging me:
- make a change on a `dirty` branch
- upload change with tag: `123abc`
- `kubectl set image kettle:<image>:v123abc`
- check, something was broken.
- make fix and re-upload (this will have the same tag `123abc`
- `kubectl set image kettle:<image>:v123abc` <<< This now does nothing because it doesn't try to repull

amwat: ok so this is for local testing, not prod manifests.
can you just push it with a different tag?

MushuEE: I could force the tag, or what I have been doing is setting it to another image than switching back

PR #19980, opened 2020-11-18
[Kettle] Skip jobs with malformed json data

MushuEE:
Some jobs seem to have malformed or non-existent: job data Ex. `<build>.txt`, `started.json`, or `finished.json`. These causes failures in Kettle when attempts are made to pull. These failures causes kettle to completely break 11/11 causing #19961

I have this image running in staging and so far it has not hit the same issue and passed the make_db steps. I will keep watch with this change and make sure there are not jobs we expect to be added getting left out. 

Also changed some loggers back to prints because the loggers are suppressed till failure.

### DIFF
--- a/kettle/make_db.py
+++ b/kettle/make_db.py
@@ -17,6 +17,7 @@
 
 import argparse
 import logging
+import json
 import os
 import random
 import re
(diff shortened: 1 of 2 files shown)


kettle/make_db.py
```diff
@@ -231,11 +236,11 @@ def get_all_builds(db, jobs_dir, metadata, threads, client_class, build_limit):
     """
     gcs = client_class(jobs_dir, metadata)
 
-    logging.info('Loading builds from %s', jobs_dir)
+    print(f'Loading builds from {jobs_dir}')
```
amwat, line 239: why switch to print instead of logging?

amwat (commented):
/lgtm

PR #19988, opened 2020-11-19
releng/vulndash: increase job timeout for debugging

cpanato:
Increasing the job timeout for debugging purposes since the job is failing due timeout and adding debug level

Will add more debug logs in the service

Discussion here: https://github.com/kubernetes/release/issues/1665#issuecomment-730526470

/assign @spiffxp


hasheddan (approved):
/lgtm

Issue #19732, opened 2020-10-28
Kettle Prod failing to decode data


MushuEE:
**What happened**:
Some file paths seem to have corrupt data with the database:
```
Error while reading data, error message: JSON parsing error in row starting at position 1286510561: Parser terminated before end of string

Error while reading data, error message: JSON parsing error in row starting at position 1945315794: Could not parse value

ERROR:root:error on gs://kubernetes-jenkins/pr-logs/pull/57600/pull-kubernetes-verify/84123
Traceback (most recent call last):
  File "make_json.py", line 280, in make_rows
    results = db.test_results_for_build(path)
  File "/kettle/model.py", line 167, in test_results_for_build
    data = zlib.decompress(dataz).decode('utf-8')
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xc0 in position 1147: invalid start byte
ERROR:root:error on gs://kubernetes-jenkins/pr-logs/pull/57600/pull-kubernetes-verify/84147
Traceback (most recent call last):
  File "make_json.py", line 280, in make_rows
    results = db.test_results_for_build(path)
  File "/kettle/model.py", line 167, in test_results_for_build
    data = zlib.decompress(dataz).decode('utf-8')
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xc0 in position 2874: invalid start byte
ERROR:root:error on gs://pivotal-e2e-results/kubo-windows-2019/1553782223
Traceback (most recent call last):
  File "make_json.py", line 281, in make_rows
    yield rowid, row_for_build(path, started, finished, results)
  File "make_json.py", line 254, in row_for_build
    build = Build.generate(path, tests, started, finished, metadata, repos)
  File "make_json.py", line 94, in generate
    build = cls(path, tests)
  File "make_json.py", line 90, in __init__
    self.populate_path_to_job_and_number()
  File "make_json.py", line 112, in populate_path_to_job_and_number
    raise ValueError(f'unknown build path for {self.path} in known bucket paths')
ValueError: unknown build path for gs://pivotal-e2e-results/kubo-windows-2019/1553782223 in known bucket paths
ERROR:root:error on gs://kubernetes-jenkins/pr-logs/pull/77766/pull-kubernetes-dependencies-canary/1127001638385487872
Traceback (most recent call last):
  File "make_json.py", line 280, in make_rows
    results = db.test_results_for_build(path)
  File "/kettle/model.py", line 167, in test_results_for_build
[...]

PR #19991, opened 2020-11-19
Skip adding results for malformed test data

MushuEE:
/area kettle

#19732 fix (the malformed jobs only exist in the Prod DB so I can't test in staging without injecting bad data)

The malformed data is with good data so I try to run a replace on the bad characters to get what we can.

All of these jobs seem extremely old and I can not find them in prow history.

### DIFF
--- a/kettle/model.py
+++ b/kettle/model.py
@@ -166,9 +166,13 @@ class Database:
         for dataz, in self.db.execute(
                 'select data from file where path between ? and ?',
                 (path, path + '\x7F')):
-            data = zlib.decompress(dataz).decode('utf-8')
-            if data:
-                results.append(data)
+            try:
+                data = zlib.decompress(dataz).decode('utf-8', 'replace')
+                if data:
+                    results.append(data)
+            except UnicodeDecodeError:
+                print(f'Failed to decode data for {path}')
+                break
         return results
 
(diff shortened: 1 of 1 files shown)


kettle/model.py
```diff
@@ -166,9 +166,13 @@ def test_results_for_build(self, path):
...
-            data = zlib.decompress(dataz).decode('utf-8')
-            if data:
-                results.append(data)
+            try:
+                data = zlib.decompress(dataz).decode('utf-8', 'replace')
+                if data:
+                    results.append(data)
+            except UnicodeDecodeError:
+                print(f'Failed to decode data for {path}')
+                break
```
amwat, line 175: do we want to break here? or continue on to next data?

MushuEE: This is all data within the same job, all test results. If one part of it is corrupt it looks like it can cascade. Because the decompress takes so long and the jobs with this issue are so old, I thought it would be more efficient to give up than skip that shard of data. Down to remove it again though.

amwat: /lgtm

MushuEE: *also the `len` of data is almost always 1, AFAIK. Have not ever had a cursor object come back with more than one entry

PR #19998, opened 2020-11-20
Remove useless minio PersistentVolumeClaim

hi-rustin:
It seems that we are not using this `minio` PersistentVolumeClaim. So I deleted it.

### DIFF
--- a/config/prow/cluster/starter-gcs.yaml
+++ b/config/prow/cluster/starter-gcs.yaml
@@ -1269,15 +1269,3 @@ subjects:
 - kind: ServiceAccount
   name: crier
   namespace: prow
----
-apiVersion: v1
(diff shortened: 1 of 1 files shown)


chases2:
This looks correct to me, but I'd like oncall to be aware of this merge.
/ok-to-test
/assign @e-blackwelder 
/assign @chaodaiG

chaodaiG:
Same here, I don't think the volume claim will be deleted by postsubmit as `kubectl apply` doesn't take care of deleted stuff, so there should be no impact imo.

/approve
Just to be safe leave the lgtm to oncall

alvaroaleman (approved):

alvaroaleman:
This manifest is not deployed for prow.k8s.io

PR #20003, opened 2020-11-20
Run master-informing capg against release-1.19

hasheddan:
This is a temporary fix to make the capg master-informing job run
against the release-1.19 conformance tests while it is using the latest
stable kubernetes version from 1.19.

This should be followed by investigating whether CAPI jobs should be on
any master dashboards for testgrid when they are running against older
branches.

Follow up to https://github.com/kubernetes/test-infra/pull/19752

Fixes https://github.com/kubernetes/kubernetes/issues/96055

/assign @cpanato @detiber

### DIFF
--- a/config/jobs/kubernetes-sigs/cluster-api-provider-gcp/cluster-api-provider-gcp-ci.yaml
+++ b/config/jobs/kubernetes-sigs/cluster-api-provider-gcp/cluster-api-provider-gcp-ci.yaml
@@ -208,7 +208,7 @@ periodics:
       path_alias: "sigs.k8s.io/image-builder"
     - org: kubernetes
       repo: kubernetes
-      base_ref: master
+      base_ref: release-1.19
       path_alias: k8s.io/kubernetes
   spec:
     containers:


cpanato (approved):
/lgtm

I think we need to take a look at all jobs for GCP and perform a cleanup, wdyt @detiber ?

cpanato:
thanks for the fix @hasheddan 3

hasheddan (commented):
/hold

looks like the CAPA job is running against `master`.. doing some more digging

hasheddan:
@cpanato I think this should be a good short term fix https://github.com/kubernetes-sigs/cluster-api-provider-gcp/pull/330

cpanato:
@hasheddan I think need to change the job `pull-cluster-api-provider-gcp-make-conformance` as well

hasheddan:
@cpanato to be clear, I think this should be closed in favor of https://github.com/kubernetes-sigs/cluster-api-provider-gcp/pull/332 because we are updating CAPG to use `master` instead of 1.19

PR #20012, opened 2020-11-23
[Metrics] Make Query with Pod Util jobs default

MushuEE:
Bringing Metric for weekly flakes up to date. Kettle now reports jobs with fields that allow them to be consumed by the Pod Util metrics. 

Deprecated the non-podutil query
/area kettle
/area metrics
/cc @BenTheElder @spiffxp @amwat


spiffxp:
```AssertionError: test-infra/metrics/README.md includes metrics that are missing config files: flakes-deprecated.```

spiffxp:
What validation has been done to ensure the new flakiness metrics are correct?

MushuEE:
I performed the new query and the old query, validated that the old query values existed with the same values. I don't know if I have a way to verify that the additional jobs are correct, however I validated they are from PodUtils.

MushuEE:
Done in #20500

Issue #19034, opened 2020-08-28
[prow] Job History for pull-kubernetes-e2e-gce has incorrect data


RobertKielty:
**What happened**:
Loaded the following in Chrome
https://prow.k8s.io/job-history/gs/kubernetes-jenkins/pr-logs/directory/pull-kubernetes-e2e-gce

**What you expected to happen**:
I expect the page to contain Job run results 

but it is mainly comprised of yellow rows with the following data

Build Id : blank
Started : 0001-01-01 00:00:00 +0000 UTC
Duration : 0s
Result : blank

**How to reproduce it (as minimally and precisely as possible)**:
Just load the link in a browser, visit Older Runs and Newer Runs

**Please provide links to example occurrences, if any**:

**Anything else we need to know?**:
Page can take up to 40s to load

Bottom of page today reports "Showing 20/40879 results"

spiffxp:
I got there via:
- https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/directory/pull-kubernetes-e2e-gce/1300326047228628993
- Click "Job History"

And am seeing the same thing

spiffxp:
This job migrated to k8s-infra-prow-build by #18916 which merged 2020-08-19 16:40 PT, I would like to understand if this is related at all

A known good to compare against:
- https://prow.k8s.io/job-history/gs/kubernetes-jenkins/pr-logs/directory/pull-kubernetes-e2e-gce-ubuntu-containerd
- migrated to k8s-infra-prow-build via https://github.com/kubernetes/test-infra/pull/18815 which merged 2020-08-12 20:20 PT
- another dimension to check: this job remains always_run and merge-blocking for all branches, while pull-kubernetes-e2e-gce doesn't always run for the main branch and release-1.19

RobertKielty:
I think next step here is to track down the source data in the bucket that job_history reads and see why is parsed oddly.

I'll see if I can find the data and report back.

spiffxp:
/help
It might be worth looking at the output of `gsutil ls gs://kubernetes-jenkins/pr-logs/directory/pull-kubernetes-e2e/` and comparing some of the later entries against what is visible in prow's job history view.

Is this bad data? Is this a bug in prow / spyglass?

RobertKielty:
@snowmanstark , who is new to the project and keen to contribute, pinged me about working on this.

Building on the signposting offered by @spiffxp (many thanks, much appreicated) we formulated the following plan 

[...]

PR #20018, opened 2020-11-23
Add trailing slash to job_history root.

### LINKED ISSUE #19721: Prow "Job History" Spammed with 0s pending runs
**What happened**:
Deck `Job History` is flooded with jobs that look like 
> <No Build ID>   0001-01-01 00:00:00 +0000 UTC     0s   <no result>

for [pull-kubernetes-e2e-gce](https://prow.k8s.io/job-history/gs/kubernetes-jenkins/pr-logs/directory/pull-kubernetes-e2e-gce)

image

Only true runs show in job history.

View Job History link above.

https://prow.k8s.io/job-history/gs/kubernetes-jenkins/pr-logs/directory/pull-kubernetes-e2e-gce

/cc @chaodaiG

vicsufer:
As noted in #19034 (comment) adding a trailing / to the job root fixes the issue.
After extracting the root field from the given URL, if it has no trailing `/` I add it.

fix #19721

/cc @MushuEE


prow/cmd/deck/job_history_test.go
```diff
@@ -44,63 +44,63 @@ func TestJobHistURL(t *testing.T) {
...
-			root:            "logs/bar-e2e",
+			root:            "logs/bar-e2e/",
 			id:              emptyID,
 		},
 		{
 			address:         "http://www.example.com/job-history/s3/foo-bucket/logs/bar-e2e?buildId=123456789123456789",
 			bktName:         "foo-bucket",
 			storageProvider: providers.S3,
-			root:            "logs/bar-e2e",
+			root:            "logs/bar-e2e/",
```
MushuEE, line 103: Should we include a test where the root doesn't have `/` to cover your added `/` case?

vicsufer: Actually every test case checks that, the root is extracted from the `address` property, if it won't work as expected the result would be a root with no trailing `/`.

MushuEE: haha got it thanks

prow/cmd/deck/job_history.go
```diff
@@ -307,6 +307,10 @@ func parseJobHistURL(url *url.URL) (storageProvider, bucketName, root string, bu
 		err = fmt.Errorf("invalid path for job: %v", url.Path)
 		return
 	}
+	// Append / to root if it does not end with it.
+	if !strings.HasSuffix(root, "/") {
```
alvaroaleman, line 311: @vicsufer please fill out the PR body and explain your fix. From what I see right now it has somewhere between no and a breaking effect:

* here its a no-op: https://github.com/kubernetes/test-infra/blob/7553d5a08c3361877bed3c3dfadb4bcf36740ac4/prow/cmd/deck/job_history.go#L427
* Here its again a no-op: https://github.com/kubernetes/test-infra/blob/7553d5a08c3361877bed3c3dfadb4bcf36740ac4/prow/cmd/deck/job_history.go#L237
* Here its breaking: https://github.com/kubernetes/test-infra/blob/7553d5a08c3361877bed3c3dfadb4bcf36740ac4/prow/cmd/deck/job_history.go#L251

alvaroaleman:
@vicsufer this change is way too broad. Please find out where exactly we are missing a trailing slash rather than this brute force method that may break other things along the way. In prow/cmd/deck/README.md is a description on how to run Deck locally.

vicsufer:
@alvaroaleman you're correct, I'll try to find a workaround.

Should this PR be closed?

alvaroaleman:
@vicsufer well whatever you prefer, updating this PR is also completely fine

PR #20023, opened 2020-11-23
add 100-node higher QPS limit scheduler job

adtac:
/hold until https://github.com/kubernetes/kubernetes/pull/96813 is merged

So sig/scheduling has made numerous improvements to the scheduler over the past few releases, but these improvements have not been realised in practice due to the scheduler's QPS limits. With API priority and fairness being graduated to beta in 1.20 (https://github.com/kubernetes/kubernetes/pull/96527), we now have a path towards safely removing the scheduler's rate limiting in the 1.21+.

Of course, this would need to be done slowly. @ahg-g and I discussed this offline and we believe that increasing the scheduler's QPS limit (not removing) in 1.21 is the first step. Once things are observed to be stable over a couple of releases, the client-side rate limiting will be entirely limit and kube-scheduler will depend on APF to make sure it doesn't overload the API server.

While I have done some preliminary benchmarking of APF with higher scheduler QPS limits (not public yet, will share with everyone in a short while), there would be more confidence in this approach if there were public, periodic benchmarks that run on test-infra. The results could be displayed in perf-dash. These benchmarks should run on the higher QPS limit. The job should be carefully observed for signs of instability. The job would be temporary and is only expected to run for a 3-4 releases; once the scheduler's client-side QPS limits are removed, just the regular job (`ci-kubernetes-kubemark-100-gce`) would be sufficient.

A separate job is needed to do this because all changes happen during cluster creation. As a result, existing jobs cannot be re-used for this.

/sig scheduling
cc @ahg-g

### DIFF
--- a/config/jobs/kubernetes/sig-scalability/sig-scalability-periodic-jobs.yaml
+++ b/config/jobs/kubernetes/sig-scalability/sig-scalability-periodic-jobs.yaml
@@ -190,7 +190,7 @@ periodics:
       - --repo=k8s.io/kubernetes=master
       - --repo=k8s.io/perf-tests=master
       - --root=/go/src
-      - --timeout=260
+      - --timeout=170
       - --scenario=kubernetes_e2e
       - --
       - --cluster=kubemark-100
@@ -219,7 +219,7 @@ periodics:
       - --test-cmd-args=--testoverrides=./testing/experiments/enable_restart_count_check.yaml
       - --test-cmd-args=--testoverrides=./testing/experiments/use_simple_latency_query.yaml
       - --test-cmd-name=ClusterLoaderV2
-      - --timeout=240m
+      - --timeout=150m
       - --use-logexporter
       # docker-in-docker needs privileged mode
       securityContext:
@@ -270,10 +270,8 @@ periodics:
       - --kubemark
       - --kubemark-nodes=100
       - --provider=gce
-      - --env=KUBE_SCHEDULER_API_QPS=300
-      - --env=KUBE_SCHEDULER_API_BURST=300
-      - --env=KUBE_CONTROLLER_MANAGER_API_QPS=300
-      - --env=KUBE_CONTROLLER_MANAGER_API_BURST=300
+      - --env=CONTROLLER_MANAGER_TEST_ARGS=--profiling --kube-api-qps=300 --kube-api-burst=300
+      - --env=SCHEDULER_TEST_ARGS=--profiling --kube-api-qps=300 --kube-api-burst=300
       - --env=CL2_ENABLE_CLUSTER_OOMS_TRACKER=true
       - --test=false
       - --test_args=--ginkgo.focus=xxxx
--- a/config/jobs/kubernetes/sig-scalability/sig-scalability-periodic-jobs.yaml
+++ b/config/jobs/kubernetes/sig-scalability/sig-scalability-periodic-jobs.yaml
@@ -232,6 +232,74 @@ periodics:
           cpu: 2
           memory: "2Gi"
 
+- name: ci-kubernetes-kubemark-100-gce-scheduler-highqps
+  tags:
+  - "perfDashPrefix: kubemark-100Nodes-scheduler-highqps"
+  - "perfDashJobType: performance"
+  interval: 24h
+  labels:
+    preset-service-account: "true"
+    preset-k8s-ssh: "true"
+    preset-dind-enabled: "true"
+    preset-e2e-kubemark-common: "true"
+    preset-e2e-scalability-periodics: "true"
+    preset-e2e-scalability-periodics-master: "true"
+  annotations:
+    testgrid-dashboards: sig-scalability-kubemark
+    testgrid-tab-name: kubemark-100-scheduler-highqps
+    testgrid-num-failures-to-alert: '1'
+  spec:
+    containers:
+    - image: gcr.io/k8s-testimages/kubekins-e2e:v20201111-a263fd7-master
+      args:
+      - --repo=k8s.io/kubernetes=master
+      - --repo=k8s.io/perf-tests=master
+      - --root=/go/src
+      - --timeout=170
+      - --scenario=kubernetes_e2e
+      - --
+      - --cluster=kubemark-100-scheduler-highqps
+      - --build=bazel
+      - --extract=bazel
+      - --gcp-master-size=e2-standard-2
+      - --gcp-node-image=gci
+      - --gcp-node-size=e2-standard-4
+      - --gcp-nodes=4
+      - --gcp-project-type=scalability-project
+      - --gcp-zone=us-east1-b
+      - --kubemark
+      - --kubemark-nodes=100
+      - --provider=gce
+      - --env=KUBE_SCHEDULER_API_QPS=300
+      - --env=KUBE_SCHEDULER_API_BURST=300
+      - --env=KUBE_CONTROLLER_MANAGER_API_QPS=300
+      - --env=KUBE_CONTROLLER_MANAGER_API_BURST=300
+      - --env=CL2_ENABLE_CLUSTER_OOMS_TRACKER=true
+      - --test=false
+      - --test_args=--ginkgo.focus=xxxx
+      - --test-cmd=$GOPATH/src/k8s.io/perf-tests/run-e2e.sh
+      - --test-cmd-args=cluster-loader2
+      - --test-cmd-args=--nodes=100
+      - --test-cmd-args=--prometheus-scrape-node-exporter=true
+      - --test-cmd-args=--provider=kubemark
+      - --test-cmd-args=--report-dir=/workspace/_artifacts
+      - --test-cmd-args=--testsuite=testing/density/scheduler-suite.yaml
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes/sig-scalability/sig-scalability-periodic-jobs.yaml
```diff
@@ -232,6 +232,76 @@ periodics:
           cpu: 2
           memory: "2Gi"
 
+- name: ci-kubernetes-kubemark-100-gce-scheduler-highqps
```
wojtek-t, line 235: Let me clarify one thing very explicitly.

While I agree this test makes sense to have (though probably I would put that in experimental file), as it will allow us to see where we are, track regressions, etc.
we are NOT close to be able to apply it in production clusters (at least by default).

We are not able to keep up with 100pods/s throughput in some usecases, and the P&F in the current shape doesn't address many of the issues yet (both handling treating lists differently than other requests, as well as addressing long-running requests (watch in particular) are on the roadmap, but they aren't done).

So I would like to make it very explicit that it's not something that I will agree (as SIG Scalability TL) to changing defaults any time soon.

config/jobs/kubernetes/sig-scalability/sig-scalability-periodic-jobs.yaml
```diff
@@ -232,6 +232,76 @@ periodics:
...
+  interval: 24h
+  labels:
+    preset-service-account: "true"
+    preset-k8s-ssh: "true"
+    preset-dind-enabled: "true"
+    preset-e2e-kubemark-common: "true"
+    preset-e2e-scalability-periodics: "true"
+    preset-e2e-scalability-periodics-master: "true"
+  annotations:
+    testgrid-dashboards: sig-scalability-kubemark
```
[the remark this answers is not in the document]
adtac: Done, I've modified the other two scheduler jobs to be in the new dashboard as well.

config/jobs/kubernetes/sig-scalability/sig-scalability-periodic-jobs.yaml
adtac: (adding some context for anyone stumbling upon this thread: we discussed this offline and everybody's on the same page -- this job is just one of many scalability tests that needs to be done before QPS limits are removed on the client-side)

config/jobs/kubernetes/sig-scalability/sig-scalability-periodic-jobs.yaml
```diff
@@ -232,6 +232,74 @@ periodics:
...
+    testgrid-tab-name: kubemark-100-highqps
+    testgrid-num-failures-to-alert: '1'
+  spec:
+    containers:
+    - image: gcr.io/k8s-testimages/kubekins-e2e:v20201111-a263fd7-master
+      args:
+      - --repo=k8s.io/kubernetes=master
+      - --repo=k8s.io/perf-tests=master
+      - --root=/go/src
+      - --timeout=260
```
wojtek-t, line 258: Seems fairly high - how long it really takes (it's BTW same comment for the job above).

config/jobs/kubernetes/sig-scalability/sig-scalability-periodic-jobs.yaml
```diff
@@ -180,8 +180,8 @@ periodics:
     preset-e2e-scalability-periodics: "true"
     preset-e2e-scalability-periodics-master: "true"
   annotations:
-    testgrid-dashboards: sig-scalability-kubemark
-    testgrid-tab-name: kubemark-100-scheduler
+    testgrid-dashboards: sig-scalability-scheduler
```
wojtek-t, line 183: It seems the amount of scheduler jobs is increasing over time. Can we extract them to a separate file, e.g.
sig-scalability-scheduler-periodic-jobs.yaml
or sth like that?

config/jobs/kubernetes/sig-scalability/sig-scalability-periodic-jobs.yaml
```diff
@@ -232,6 +232,74 @@ periodics:
...
+      - --root=/go/src
+      - --timeout=260
+      - --scenario=kubernetes_e2e
+      - --
+      - --cluster=kubemark-100-scheduler-highqps
+      - --build=bazel
+      - --extract=bazel
+      - --gcp-master-size=n1-standard-2
+      - --gcp-node-image=gci
+      - --gcp-node-size=n1-standard-4
```
wojtek-t, line 266: Let's migrate to e2 VMs.

config/jobs/kubernetes/sig-scalability/sig-scalability-periodic-jobs.yaml
wojtek-t: Also - please move these changes to a separate PR.

config/jobs/kubernetes/sig-scalability/sig-scalability-periodic-jobs.yaml
adtac: the average seems to be around 1 hour (occasionally taking 1-2 hours)

my reasoning behind 260 was: 20 mins to set up the job + 4*60 mins to for each test in the scheduler suite -- none of the tests in the suite should take 60 mins with just 100 nodes, so there was also some breathing space built into the 260

I'm fine with reducing this and the job's timeout to 150 mins

config/jobs/kubernetes/sig-scalability/sig-scalability-periodic-jobs.yaml
adtac: Done, removed the dashboard change from this PR, I'll open a separate PR for that + a new jobs YAML file.

config/jobs/kubernetes/sig-scalability/sig-scalability-periodic-jobs.yaml
adtac: Done. Note: I only updated the new job's machine type; I didn't update the 100Nodes-scheduler job to e2, not sure what the scalability team's policy is on updating the machine type of existing jobs.

config/jobs/kubernetes/sig-scalability/sig-scalability-periodic-jobs.yaml
adtac: Updated both jobs to 150 minutes.

adtac:
@wojtek-t updated this PR to use `--env=CONTROLLER_MANAGER_TEST_ARGS` and `--env=SCHEDULER_TEST_ARGS`

config/jobs/kubernetes/sig-scalability/sig-scalability-periodic-jobs.yaml
```diff
@@ -219,7 +219,73 @@ periodics:
...
+- name: ci-kubernetes-kubemark-100-gce-scheduler-highqps
+  tags:
+  - "perfDashPrefix: kubemark-100Nodes-scheduler-highqps"
+  - "perfDashJobType: performance"
+  interval: 24h
+  labels:
+    preset-service-account: "true"
+    preset-k8s-ssh: "true"
+    preset-dind-enabled: "true"
+    preset-e2e-kubemark-common: "true"
```
wojtek-t, line 244: I'm fine with starting with kubemark, but eventually we should consider migrating to real clusters. Kubemark is still visibly different than real clusters (easier from the point-of-view of control-plane).

wojtek-t:
@adtac - next time please ping me if I'm not responding for week+.

/lgtm
/approve

config/jobs/kubernetes/sig-scalability/sig-scalability-periodic-jobs.yaml
```diff
@@ -219,7 +219,73 @@ periodics:
...
+    - image: gcr.io/k8s-testimages/kubekins-e2e:v20201111-a263fd7-master
+      args:
+      - --repo=k8s.io/kubernetes=master
+      - --repo=k8s.io/perf-tests=master
+      - --root=/go/src
+      - --timeout=170
+      - --scenario=kubernetes_e2e
+      - --
+      - --cluster=kubemark-100-scheduler-highqps
+      - --build=bazel
```
BenTheElder, line 262: is there a reason to be checking out kubernetes and doing a source build here?
most CI jobs consume our existing builds (--extract=ci/latest or similar)

this is cheaper and faster

adtac: not really, but IIRC to test the PR I had to build from my fork and I must have forgotten to change back to --extract=ci/latest. I'll open a PR to switch to reuse builds.

BenTheElder: Ack thanks!
I'm working on switching over how we build (bazel=>quick) but suspected this one shouldn't be building anyhow

BenTheElder: https://github.com/kubernetes/test-infra/pull/21062 covers this

adtac: thanks @BenTheElder!

PR #19284, opened 2020-09-19
Remove Kops unenforceable required checks

hakman:
The GitHub checks cannot be enforced due to https://github.com/kubernetes/test-infra/issues/15611.

PR #20029, opened 2020-11-24
Add back the GH jobs the required status check

hakman:
Re-adding the changes from #19284 now that https://github.com/kubernetes/test-infra/issues/15611 was fixed.

/cc @rifelpet @mikesplain @justinsb


spiffxp (approved):
/approve
/lgtm

Issue #19981, opened 2020-11-18
Tide error: broken pipe


chases2:
```
jsonPayload: {
file: "prow/tide/history/history.go:183"
error: "write tcp 10.108.11.15:8888->10.108.6.28:47268: write: broken pipe"
level: "error"
func: "k8s.io/test-infra/prow/tide/history.(*History).ServeHTTP"
component: "tide"
msg: "Writing JSON history response."
}
```

chases2:
Occurred twice in a row (1 minute apart), but then seemed to stop.

e-blackwelder:
It looks like the browser's side of the HTTP/TCP connection is closing "unexpectedly": users could be canceling the page load, navigate away, refresh the page, etc.

Per:
https://github.com/kubernetes/test-infra/blob/master/prow/tide/history/history.go#L182

... This is not a bug. However, we should lower the "severity" of the logs so we don't panic about this again. :)

PR #20040, opened 2020-11-24
Tide: Degrade error writing http response to debug

alvaroaleman:
This is expected to happen if the client vanishes, e.G. because we are
currently updating Prow so an old Deck instance gets removed.

Fixes https://github.com/kubernetes/test-infra/issues/19981


hi-rustin (approved):
/hold

Because I am not a reviewer for the Prow, but it looks good to me.

PR #19603, opened 2020-10-16
[sig-scalability] Migrate ci-kubernetes-e2e-gce-scale-correctness to E2

jprzychodzen:
Migrate 5k scale correctness test to e2 instances.

Tested manually. Tested that test passed in 2h38m29s, so it's similar time.

/assign @wojtek-t 
/cc @mborsz

PR #20043, opened 2020-11-25
Migrate nodes of scalability master periodic jobs to e2- family

tosi3k:
Let's start with master periodics and then proceed with release branches periodics and with presubmits in the end.

/sig scalability
/assign @wojtek-t

### DIFF
--- a/config/jobs/kubernetes/sig-scalability/sig-scalability-golang.yaml
+++ b/config/jobs/kubernetes/sig-scalability/sig-scalability-golang.yaml
@@ -67,7 +67,7 @@ periodics:
       - --env=KUBEMARK_CONTROLLER_MANAGER_TEST_ARGS=--profiling --kube-api-qps=200 --kube-api-burst=200
       - --env=KUBEMARK_SCHEDULER_TEST_ARGS=--profiling --kube-api-qps=200 --kube-api-burst=200
       - --extract=gs://k8s-scale-golang-build/ci/latest-1.18.txt
-      - --gcp-node-size=n1-standard-8
+      - --gcp-node-size=e2-standard-8
       - --gcp-nodes=50
       - --gcp-project=k8s-presubmit-scale
       - --gcp-zone=us-east1-b
--- a/config/jobs/kubernetes/sig-scalability/sig-scalability-periodic-jobs.yaml
+++ b/config/jobs/kubernetes/sig-scalability/sig-scalability-periodic-jobs.yaml
@@ -131,7 +131,7 @@ periodics:
       - --extract=ci/latest
       - --gcp-master-size=n1-standard-2
       - --gcp-node-image=gci
-      - --gcp-node-size=n1-standard-4
+      - --gcp-node-size=e2-standard-4
       - --gcp-nodes=4
       - --gcp-project-type=scalability-project
       - --gcp-zone=us-east1-b
@@ -196,7 +196,7 @@ periodics:
       - --extract=ci/latest
       - --gcp-master-size=n1-standard-2
       - --gcp-node-image=gci
-      - --gcp-node-size=n1-standard-4
+      - --gcp-node-size=e2-standard-4
       - --gcp-nodes=4
       - --gcp-project-type=scalability-project
       - --gcp-zone=us-east1-b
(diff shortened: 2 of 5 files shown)


tosi3k:
CCing @marseel for visibility as the current scalability oncall.

config/jobs/kubernetes/sig-scalability/sig-scalability-golang.yaml
```diff
@@ -66,8 +66,9 @@ periodics:
       - --env=CL2_LOAD_TEST_THROUGHPUT=50
       - --env=KUBEMARK_CONTROLLER_MANAGER_TEST_ARGS=--profiling --kube-api-qps=200 --kube-api-burst=200
       - --env=KUBEMARK_SCHEDULER_TEST_ARGS=--profiling --kube-api-qps=200 --kube-api-burst=200
+      - --env=HEAPSTER_MACHINE_TYPE=e2-standard-8
```
wojtek-t, line 69: If it has the same size as the other machines, it isn't needed, right?

config/jobs/kubernetes/sig-scalability/sig-scalability-periodic-jobs.yaml
```diff
@@ -330,7 +330,8 @@ periodics:
       - --cluster=kubemark-5000
       - --extract=ci/latest
       - --gcp-node-image=gci
-      - --gcp-node-size=n1-standard-8
+      - --gcp-node-size=e2-standard-8
+      - --env=HEAPSTER_MACHINE_TYPE=e2-standard-8
```
wojtek-t, line 334: Same here and below

config/jobs/kubernetes/sig-scalability/sig-scalability-release-blocking-jobs.yaml
```diff
@@ -76,7 +76,7 @@ periodics:
       - --scenario=kubernetes_e2e
       - --
       - --cluster=gce-scale-cluster
-      - --env=HEAPSTER_MACHINE_TYPE=n1-standard-32
+      - --env=HEAPSTER_MACHINE_TYPE=e2-standard-32
```
wojtek-t, line 79: Let's not migrate those large ones at this point.

tosi3k: I think that the large ones were already migrated in https://github.com/kubernetes/test-infra/pull/19594 and https://github.com/kubernetes/test-infra/pull/19603 but we forgot to change this env var appropriately in 5k performance job, hence my change here.

Shall I include this amendment in a separate PR?

config/jobs/kubernetes/sig-scalability/sig-scalability-golang.yaml
tosi3k: Makes sense. OTOH, we include `preset-e2e-kubemark-gce-scale` preset here where we have such an entry:
```
- name: HEAPSTER_MACHINE_TYPE
    value: "n1-standard-8"
```
hence my addition here. Shall we get rid of this entry in that preset altogether then?

config/jobs/kubernetes/sig-scalability/sig-scalability-periodic-jobs.yaml
tosi3k: Ditto here and below.

config/jobs/kubernetes/sig-scalability/sig-scalability-golang.yaml
wojtek-t: Let's change the preset then (I know it will affect other jobs, but that's probably fine).

tosi3k: Done.

config/jobs/kubernetes/sig-scalability/sig-scalability-periodic-jobs.yaml
tosi3k: Done.

config/jobs/kubernetes/sig-scalability/sig-scalability-release-blocking-jobs.yaml
tosi3k: Leaving it as it is after offline discussion.

mborsz:
FTR: This PR has caused ~10% pod startup degradation:
4sSNwqbbkQ3iCnU

I think it's acceptable as we are still way within SLO, but I think it's good to know that there is some performance diff between n1-standard-1 and e2-medium nodes.

wojtek-t:
Which isn't completely unexepcted, given that e2-medium are shared core VMs.

Just to confirm that, we don't see any impact on "run-to-watch" phase, but there is impact on "schedule-to-watch", which clearly means that the time to start is what is impacted. Which matches the fact that something on the node changed.

PR #20050, opened 2020-11-26
Change k/release block regex to only include push-build.sh

saschagrunert:
We moved all release logic into our internal library. The only legacy
code is the now self-contained push-build.sh script. This means that we
now only block on modification to push-build.sh, which should be removed
in the future as well.

cc @kubernetes/release-engineering


xmudrii (approved):
/lgtm

hasheddan (approved):
/lgtm

PR #20054, opened 2020-11-26
Migrate nodes of scalability presubmit jobs to e2- family

tosi3k:
Finish what we did in https://github.com/kubernetes/test-infra/pull/20047 and https://github.com/kubernetes/test-infra/pull/20043.

/sig scalability


tosi3k:
@wojtek-t FYI: putting this on hold until we get all the (hopefully positive) results from release-1.XX branches jobs which should be O(tomorrow). Then we could proceed with merging this on Monday.

/hold

hasheddan (approved):
/lgtm

PR #20055, opened 2020-11-27
added secondary github-graphql-endpoint

dudicoco:
If ghproxy is down tide/deck should target the github endpoint.

### DIFF
--- a/config/prow/cluster/starter-gcs.yaml
+++ b/config/prow/cluster/starter-gcs.yaml
@@ -371,6 +371,7 @@ spec:
         - --github-endpoint=http://ghproxy
         - --github-endpoint=https://api.github.com
         - --github-graphql-endpoint=http://ghproxy/graphql
+        - --github-graphql-endpoint=https://api.github.com/graphql
         - --plugin-config=/etc/plugins/plugins.yaml
         - --spyglass=true
         ports:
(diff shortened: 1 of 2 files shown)


hi-rustin (approved):

hi-rustin:
/lgtm cancel

Better squash your PR to one commit.

I don’t know whether you should add this endpoint to other places as well?

alvaroaleman:
/hold
The V4 client supports only one endpoint

chases2:
Should it target GitHub itself if GHproxy is down? That would fix the issue for a little while, but if we exhaust our API limit, fixing GHProxy won't fix the outage.

dudicoco:
@alvaroaleman i'm not sure what the V4 client is, but it worked for me when testing with two endpoints and scaling ghproxy down to 0.
@chases2 how would that be different than the backup endpoint for the `github-endpoint` arg?

chases2:
> @chases2 how would that be different than the backup endpoint for the `github-endpoint` arg?

It's not. I guess I'm questioning that decision, too. Maybe we've decided that an outage of up to an hour for that reason isn't as big a deal as an outage due to ghProxy failing.

Having the graphql endpoint do the same thing the API endpoint looks like a good idea to me, though. I'll let @alvaroaleman remove the hold once they're satisfied.

/approve

alvaroaleman:
>  i'm not sure what the V4 client is, but it worked for me when testing with two endpoints and scaling ghproxy down to 0.

Presumably because passing the arg the second time completely overrides the first value of the arg. We definitely do not support more than one endpoint for the V4/graphql api: https://github.com/kubernetes/test-infra/blob/4d2c0d9ae722adc2212edd2bc159bd051568dbba/prow/github/client.go#L488

alvaroaleman:
Just to clarify again: The graphql client in our github client does not support multiple endpoints, that is only supported by the v3 client. If the flag can be set multiple times, that's easier an annoying nuance of how go flags work or a bug in our code, but it is not going to result in the graphql client using more than one endpoint.

/close

PR #20061, opened 2020-11-30
Parametrize gke instance group prefix

marseel:
This PR allows us to modify prefix of instance group.

### DIFF
--- a/kubetest/gke.go
+++ b/kubetest/gke.go
@@ -55,16 +55,17 @@ var (
 	gkeSubnetMode                  = flag.String("gke-subnet-mode", "auto", "(gke only) subnet creation mode of the GKE cluster network.")
 	gkeReleaseChannel              = flag.String("gke-release-channel", "", "(gke only) if specified, bring up GKE clusters from that release channel.")
 	gkeSingleZoneNodeInstanceGroup = flag.Bool("gke-single-zone-node-instance-group", true, "(gke only) Add instance groups from a single zone to the NODE_INSTANCE_GROUP env variable.")
+	gkeInstanceGroupPrefix         = flag.String("gke-instance-group-prefix", "gke", "(gke only) Use a different instance group prefix.")
 	gkeNodePorts                   = flag.String("gke-node-ports", "", "(gke only) List of ports on nodes to open, allowing e.g. master to connect to pods on private nodes. The format should be 'protocol[:port[-port]],[...]' as in gcloud compute firewall-rules create --allow.")
 	gkeCreateNat                   = flag.Bool("gke-create-nat", false, "(gke only) Configure Cloud NAT allowing outbound connections in cluster with private nodes.")
 	gkeNatMinPortsPerVm            = flag.Int("gke-nat-min-ports-per-vm", 64, "(gke only) Specify number of ports per cluster VM for NAT router. Number of ports * number of nodes / 64k = number of auto-allocated IP addresses (there is a hard limit of 100 IPs).")
 
-	// poolRe matches instance group URLs of the form `https://www.googleapis.com/compute/v1/projects/some-project/zones/a-zone/instanceGroupManagers/gke-some-cluster-some-pool-90fcb815-grp`. Match meaning:
+	// poolReTemplate matches instance group URLs of the form `https://www.googleapis.com/compute/v1/projects/some-project/zones/a-zone/instanceGroupManagers/gke-some-cluster-some-pool-90fcb815-grp`. Match meaning:
 	// m[0]: path starting with zones/
 	// m[1]: zone
 	// m[2]: pool name (passed to e2es)
 	// m[3]: unique hash (used as nonce for firewall rules)
-	poolRe = regexp.MustCompile(`zones/([^/]+)/instanceGroupManagers/(gke-.*-([0-9a-f]{8})-grp)$`)
+	poolReTemplate = `zones/([^/]+)/instanceGroupManagers/(%s-.*-([0-9a-f]{8})-grp)$`
 
 	urlRe = regexp.MustCompile(`https://.*/`)
 )
@@ -98,6 +99,7 @@ type gkeDeployer struct {
 	createCommand               []string
 	singleZoneNodeInstanceGroup bool
 	sshProxyInstanceName        string
+	poolRe                      *regexp.Regexp
 
 	setup          bool
 	kubecfg        string
@@ -152,6 +154,12 @@ func newGKE(provider, project, zone, region, network, image, imageFamily, imageP
 			return nil, fmt.Errorf("--image-family and --image-project must be set for GKE deployment if --gcp-node-image=CUSTOM")
 		}
 	}
+
+	poolRe, err := regexp.Compile(fmt.Sprintf(poolReTemplate, *gkeInstanceGroupPrefix))
+	if err != nil {
+		return nil, fmt.Errorf("Couldn't compile regex %v. prefix: %s", err, *gkeInstanceGroupPrefix)
+	}
+	g.poolRe = poolRe
 	g.imageFamily = imageFamily
 	g.imageProject = imageProject
 	g.image = image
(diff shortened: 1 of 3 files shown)


jkaniuk:
In general, lgtm, but could you add some unit tests for this regex?
Like: zonal / regional / "test" prefix?

marseel:
I've added tests. It's in separate commit to make it easier to read. I will squash commits once it's ok for merging.

kubetest/gke_test.go
```diff
@@ -0,0 +1,174 @@
...
+		"imageFamily",
+		"imageProject",
+		"cluster",
+		"sshProxyInstanceName",
+		new(string),
+		new(string),
+	)
+}
+
+func TestParseInstanceGroupsFromGCloud(t *testing.T) {
```
jkaniuk, line 51: With  all those variables, I  have problems tracking what is being actually tested :-/

Why not use plain strings in those tests?

kubetest/gke_test.go
```diff
@@ -0,0 +1,174 @@
...
+				{
+					path: expectedPathCDefault,
+					zone: zoneC,
+					name: expectedNameDefault,
+					uniq: expectedUniq,
+				},
+			},
+		},
+		{
+			input:                 strings.Join([]string{urlZoneADefault, urlZoneBDefault, urlZoneCDefault}, ";"),
```
jkaniuk, line 134: nit: that could be zonal - same test, less code

marseel: Zonal is also implemented in line 92. I guess I could remove this test.

kubetest/gke_test.go
marseel: I'm pretty sure that with the length of these strings it would be quite hard to read it. Variables' names kind of summarize what's important in string without reading whole string.

marseel: I've replaces variables with strings.

marseel (commented):
Please take another look

kubetest/gke_test.go
```diff
@@ -0,0 +1,164 @@
...
+		"imageProject",
+		"cluster",
+		"sshProxyInstanceName",
+		new(string),
+		new(string),
+	)
+}
+
+func TestParseInstanceGroupsFromGCloud(t *testing.T) {
+	urlTemplate := `https://www.googleapis.com/compute/v1/projects/test-project/zones/%s/instanceGroupManagers/%s-io-default-pool-6f8ec327-grp`
```
jkaniuk, line 52: nit: this `-io-` part, where does it come from?

kubetest/gke.go
```diff
@@ -152,6 +154,8 @@ func newGKE(provider, project, zone, region, network, image, imageFamily, imageP
 			return nil, fmt.Errorf("--image-family and --image-project must be set for GKE deployment if --gcp-node-image=CUSTOM")
 		}
 	}
+
+	g.poolRe = regexp.MustCompile(fmt.Sprintf(poolTemplateRe, *gkeInstanceGroupPrefix))
```
jkaniuk, line 158: After moving regex compiling from package initialization to main code, please handle compile errors, as they would no longer be so obvious to catch.

kubetest/gke.go
```diff
@@ -657,14 +657,26 @@ func (g *gkeDeployer) getInstanceGroups() error {
 	if len(g.instanceGroups) > 0 {
 		return nil
 	}
+	igs, err := g.getInstanceGroupsFromGCloud()
+	if err != nil {
+		return err
+	}
+	return g.parseInstanceGroupsFromGCloud(igs)
```
jkaniuk, line 664: nit: s/GCloud/Gcloud/

kubetest/gke.go
```diff
@@ -657,14 +657,26 @@ func (g *gkeDeployer) getInstanceGroups() error {
 	if len(g.instanceGroups) > 0 {
 		return nil
 	}
+	igs, err := g.getInstanceGroupsFromGCloud()
```
jkaniuk, line 660: nit: s/GCloud/Gcloud/ - original name is not with caps

kubetest/gke.go
```diff
@@ -657,14 +657,26 @@ func (g *gkeDeployer) getInstanceGroups() error {
...
 		g.location)...).Output()
 	if err != nil {
-		return fmt.Errorf("instance group URL fetch failed: %s", util.ExecError(err))
+		return "", fmt.Errorf("instance group URL fetch failed: %s", util.ExecError(err))
 	}
-	igURLs := strings.Split(strings.TrimSpace(string(igs)), ";")
+	return string(igs), nil
+}
+
+func (g *gkeDeployer) parseInstanceGroupsFromGCloud(igs string) error {
```
jkaniuk, line 678: nit: It could be more explicit and readable to return the result instead of handling assignment internally - also from tests point of view.

jkaniuk (commented):
Few small comments, apart from that: ok

kubetest/gke_test.go
marseel: I've removed it. I took it from some default instance group I got on GCP

kubetest/gke.go
marseel: Done.

kubetest/gke.go
marseel: Done

kubetest/gke.go
marseel: I was also thinking about that, but didn't want to change previous behaviour ;) Done

kubetest/gke_test.go
```diff
@@ -0,0 +1,180 @@
...
+					zone: "us-central1-c",
+					name: `test-default-pool-6f8ec327-grp`,
+					uniq: `6f8ec327`,
+				},
+			},
+			expectedError: false,
+		},
+	}
+
+	for _, tc := range cases {
```
jkaniuk, line 168: having test case names, t.Run should be used with this name
https://blog.golang.org/subtests

marseel: Done.

kubetest/gke.go
```diff
@@ -55,16 +55,17 @@ var (
...
 	gkeNatMinPortsPerVm            = flag.Int("gke-nat-min-ports-per-vm", 64, "(gke only) Specify number of ports per cluster VM for NAT router. Number of ports * number of nodes / 64k = number of auto-allocated IP addresses (there is a hard limit of 100 IPs).")
 
-	// poolRe matches instance group URLs of the form `https://www.googleapis.com/compute/v1/projects/some-project/zones/a-zone/instanceGroupManagers/gke-some-cluster-some-pool-90fcb815-grp`. Match meaning:
+	// poolTemplateRe matches instance group URLs of the form `https://www.googleapis.com/compute/v1/projects/some-project/zones/a-zone/instanceGroupManagers/gke-some-cluster-some-pool-90fcb815-grp`. Match meaning:
 	// m[0]: path starting with zones/
 	// m[1]: zone
 	// m[2]: pool name (passed to e2es)
 	// m[3]: unique hash (used as nonce for firewall rules)
-	poolRe = regexp.MustCompile(`zones/([^/]+)/instanceGroupManagers/(gke-.*-([0-9a-f]{8})-grp)$`)
+	poolTemplateRe = `zones/([^/]+)/instanceGroupManagers/(%s-.*-([0-9a-f]{8})-grp)$`
```
wojtek-t, line 68: poolReTemplate ?

[it's not regexp for poolTemplate but its template for "pool regexp"

kubetest/gke_test.go
```diff
@@ -0,0 +1,182 @@
+/*
+Copyright 2017 The Kubernetes Authors.
```
wojtek-t, line 2: 2020

kubetest/gke.go
marseel: Done.

kubetest/gke_test.go
marseel: Done.

PR #20065, opened 2020-11-30
Metrics also collects prow version timestamp

chaodaiG:
This is useful in prometheus alerting, so that one can set up prometheus rule based on the freshness of the running binary. I.e. if the image is older than 2 weeks, ping oncall to bring their attention

### DIFF
--- a/prow/kube/metrics.go
+++ b/prow/kube/metrics.go
@@ -18,6 +18,7 @@ package kube
 
 import (
 	"github.com/prometheus/client_golang/prometheus"
+	"github.com/sirupsen/logrus"
 
 	prowapi "k8s.io/test-infra/prow/apis/prowjobs/v1"
 	"k8s.io/test-infra/prow/version"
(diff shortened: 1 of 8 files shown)


prow/kube/metrics.go
```diff
@@ -118,6 +123,15 @@ func GatherProwJobMetrics(current []prowapi.ProwJob) {
 	// stale metrics like triggered or pending jobs that are now complete.
 	prowJobs.Reset()
 
+	// record prow version
+	version, err := version.VersionTimestamp()
+	if err != nil {
+		// Not worth panicking
```
fejta, line 129: You don't need to panic, but please log the error.

prow/kube/metrics.go
```diff
@@ -113,11 +119,21 @@ var previousStates map[jobIdentifier]prowapi.ProwJobState
 
 // GatherProwJobMetrics gathers prometheus metrics for prowjobs.
 // Not threadsafe, ensure this is called serially.
-func GatherProwJobMetrics(current []prowapi.ProwJob) {
+func GatherProwJobMetrics(l *logrus.Entry, current []prowapi.ProwJob) {
```
fejta, line 122: You can just do logrus.WithError(err).Debug()" if you don't want to pass in an entry.

fejta (approved):
/hold

prow/kube/metrics.go
chaodaiG: That's good to know. In this case I do want log with more context, so that this debug log can tell which component it's calling from.

prow/kube/metrics.go
chaodaiG: done

PR #20084, opened 2020-12-01
capg: update interval for periodic conformance tests to 12 hours

cpanato:
Increase the interval of the periodic conformance tests for CAPG to 12 hours, i think we don't need to run every 4 hours

/assign @detiber


detiber (commented):
/lgtm
/approve

PR #20085, opened 2020-12-01
Add vi to kettle for local debugging ease

MushuEE:
Just adding `vi`, right now there is not text editor and viewing files needs to be done with cat and editing is not possible without writing directly `>`

/area kettle
/cc @spiffxp


MushuEE:
Will not add VI to prod image

Issue #20024, opened 2020-11-23
[Kettle] Process hangs when generating json.gz for 'all' table


MushuEE:
**What would you like to be added**:
Add more compute to Kettle instances. Or figure out if CPU is impacting speed/lockup.

**Why is this needed**:
```
top - 10:34:45 up 2 days,  5:18,  0 users,  load average: 1.16, 1.19, 1.18
Tasks:   8 total,   2 running,   6 sleeping,   0 stopped,   0 zombie
%Cpu(s): 13.4 us,  0.3 sy,  0.0 ni, 86.2 id,  0.0 wa,  0.0 hi,  0.1 si,  0.0 st
KiB Mem : 53588024 total,  2290884 free,  7649364 used, 43647776 buff/cache
KiB Swap:        0 total,        0 free,        0 used. 45492108 avail Mem

    PID USER      PR  NI    VIRT    RES    SHR S  %CPU %MEM     TIME+ COMMAND
    619 root      20   0 5952940 5.110g  42432 R 100.0 10.0   2897:56 pypy3
      1 root      20   0   18384   3096   2820 S   0.0  0.0   0:00.02 runner.sh
     40 root      20   0   26768   9056   5204 S   0.0  0.0   0:00.03 python3
    618 root      20   0    4636    836    768 S   0.0  0.0   0:00.00 sh
    620 root      20   0    4700    828    768 S   0.0  0.0   1:05.94 pv
    621 root      20   0    4792   1528   1252 S   0.0  0.0   0:01.64 gzip
    643 root      20   0   18512   3404   3028 S   0.0  0.0   0:00.00 bash
    652 root      20   0   36628   3092   2644 R   0.0  0.0   0:00.03 top
```
It seems that Kettle Prod is hitting cpu limits when trying to build json. It takes extremely long to complete an update cycle and now seems to freeze at point, not updating at all and "catching" on specific builds *logs seem to end there*
```
Error while reading data, error message: JSON parsing error in row starting at position 605377307: Parser terminated before end of string

Error while reading data, error message: JSON parsing error in row starting at position 752722019: Parser terminated before end of string

Error while reading data, error message: JSON parsing error in row starting at position 1126381032: Parser terminated before end of string

ERROR:root:error on gs://pivotal-e2e-results/kubo-windows-2019/1553782223
Traceback (most recent call last):
  File "make_json.py", line 281, in make_rows
    yield rowid, row_for_build(path, started, finished, results)
[...]

PR #20091, opened 2020-12-02
[Kettle] Limit days of builds to `all` table to avoid deadlock

MushuEE:
Due to #20024 we are seeing kettle prod hand indefinitely on `make_json` for the `all` table. This seems to not be an issue if the number of builds collected and pushed is limited to 30 days. (Hung on 90+ days).

30 days will back update kettle to before the `all` table stopped being populated and will limit the collection time. This does not write over existing older builds and will be safe for BQ. This should be reverted when there is a more permanent solution or a change in architecture.

This change was tested in Kettle and successfully updated `all`.

I also added some logging to better catch the odd exception I had seen in #19732

/cc @amwat @spiffxp @roberthbailey 
/area Kettle


MushuEE:
/hold this might be counting dupes without including the --replace. (We only want to replace data from the last 30 days)

- Can we break the make_json into multiple X day blocks?

MushuEE:
UPDATE: The way kettle works, it creates a unique table based on the days specified. Previously there was no 30 day table. This caused builds that had been uploaded to be double counted in the `build.all` 
`SELECT * FROM k8s-gubernator.build.all WHERE started between '2020-11-1 22:52:14 UTC' and '2020-11-8 22:52:14 UTC' order by number LIMIT 1000`

This should not repeat going forward but will cause an issue if reverting back to `all` and not `--30` I can remove duplicates however (i believe) @spiffxp 

/hold cancel

spiffxp (changes requested):
Can you rebase to avoid the merge commits?

kettle/Dockerfile
```diff
@@ -28,6 +28,7 @@ RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime \
     python \
     python3 \
     python3-pip \
+    vim \
```
amwat, line 31: we usually avoid adding text editors to prod images, i'm guessing this was needed during debugging?

amwat (commented):
/lgtm
/hold

feel free to cancel the hold if the duplicate data issue is resolved.

kettle/Dockerfile
MushuEE: Yes it is just generally helpful but I can remove

MushuEE:
@spiffxp 30 day issue will not repeat duplicates but I would like to get this in soon if possible. I am tempted to backup the older database and either clear it or start fresh with a new one to hopefully give kettle back some life.

amwat:
/lgtm
/hold

@MushuEE Is this good to merge?
feel free to cancel the hold if all the issues are resolved.

spiffxp:
The triage job has started failing because it's getting zero results, I'm assuming merging/deploying this might bring it back to green (though we should fix triage code to handle zero results) https://github.com/kubernetes/test-infra/issues/20259

PR #19853, opened 2020-11-05
images(kubekins-e2e,krte): Stop building 1.16 images

justaugustus:
Kubernetes 1.16 is out of support, so we stop building the kubekins-e2e and krte 1.16 images.

The 1.16 cleanup PR (https://github.com/kubernetes/test-infra/pull/19850) is quickly becoming a rebase magnet, so I'm pulling this out to get it merged first.

Signed-off-by: Stephen Augustus <saugustus@vmware.com>

/assign @cpanato @saschagrunert @hasheddan 
cc: @kubernetes/release-engineering

PR #20094, opened 2020-12-02
kubeadm jobs: update manifests.yaml to not use latest-master

neolit123:
update manifests.yaml to use the latest kubekins tag instead of `latest-x` and to not always pulling the image.

### DIFF
--- a/config/jobs/kubernetes/sig-cluster-lifecycle/manifests.yaml
+++ b/config/jobs/kubernetes/sig-cluster-lifecycle/manifests.yaml
@@ -13,8 +13,7 @@ periodics:
       - --scenario=execute
       - --
       - ./tests/e2e/manifests/verify_manifest_lists.sh
-      image: gcr.io/k8s-testimages/kubekins-e2e:latest-master
-      imagePullPolicy: Always
+      image: gcr.io/k8s-testimages/kubekins-e2e:v20201203-4778e22-master
   annotations:
     testgrid-dashboards: sig-cluster-lifecycle-all
     testgrid-tab-name: periodic-manifest-lists


neolit123:
/sig cluster-lifecycle
/cc @fabriziopandini @SataQiu 

looking for LGTM on this one. thanks.

neolit123:
one problem with this current PR is that it requires always pulling the latest image which is heavy on the infra.
looks like the bot did an update here:
https://github.com/kubernetes/test-infra/pull/20064

but missed one entry:
```
kubeadm-kinder-x-on-y.yaml:191:    - image: gcr.io/k8s-testimages/kubekins-e2e:v20201031-122dc79-1.16
```
since it was using an outdated tag.

i will change this PR to only fix that entry and hope the bot mostly handles that instead of using "latest"

config/jobs/kubernetes/sig-cluster-lifecycle/kubeadm-kinder-x-on-y.yaml
```diff
@@ -188,7 +188,7 @@ periodics:
     path_alias: k8s.io/kubeadm
   spec:
     containers:
-    - image: gcr.io/k8s-testimages/kubekins-e2e:v20201031-122dc79-1.16
+    - image: gcr.io/k8s-testimages/kubekins-e2e:v20201201-1941b8f-1.16
```
justaugustus, line 191: This image doesn't exist.

justaugustus (changes requested):
@neolit123 -- We stopped building the 1.16 kubekins/krte variants in https://github.com/kubernetes/test-infra/pull/19853, so the 1.16 change here needs to be dropped.

config/jobs/kubernetes/sig-cluster-lifecycle/kubeadm-kinder-x-on-y.yaml
neolit123: i guess this is another reason to use `latest-x.yy` everywhere, but since kubekins is ~2GB that's a lot of traffic for all the jobs as they need "Always" as the pull policy.

we can try to keep track of this image tag discrepancy manually for the time being.

neolit123: updated.

config/jobs/kubernetes/sig-cluster-lifecycle/manifests.yaml
```diff
@@ -13,8 +13,7 @@ periodics:
       - --scenario=execute
       - --
       - ./tests/e2e/manifests/verify_manifest_lists.sh
-      image: gcr.io/k8s-testimages/kubekins-e2e:latest-master
-      imagePullPolicy: Always
+      image: gcr.io/k8s-testimages/kubekins-e2e:v20201201-1941b8f-master
```
SataQiu, line 16: The bot seems to update the image tag automatically #20115 
Now the latest image is `gcr.io/k8s-testimages/kubekins-e2e:v20201203-4778e22-master`

neolit123: good catch @SataQiu 
updated to `v20201203-4778e22-master`

neolit123:
CI failure looks unrelated:

> ERROR: /home/prow/go/src/github.com/kubernetes/test-infra/testgrid/cmd/transfigure/BUILD.bazel:3:1: //testgrid/cmd/transfigure:image depends on @bazel-base//image:image in repository @bazel-base which failed to fetch. no such package '@bazel-base//image': Pull command failed: 2020/12/07 15:07:25 Running the Image Puller to pull images from a Docker Registry...

https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/test-infra/20094/pull-test-infra-bazel/1335963729568731136

/retest

PR #20097, opened 2020-12-02
Add tests for containerd

justinsb:
Preparing for the great deprecation!


justinsb:
cc @hakman I think this is where we can get the test coverage to inform when containerd can be the default.

PR #20098, opened 2020-12-02
azure: remove 1.20 containerd jobs

chewong:

/assign @marosset


chewong:
ping @marosset

marosset:
/lgtm
These are duplicate jobs and we are only testing with containerd for Windows (at least on Azure) in release jobs for v.120+

PR #20100, opened 2020-12-02
Switch kubernetes 1.20 bazel jobs to an image that exists

spiffxp:
This is intended to be a temporary workaround until a proper 1.20
variant exists, if that is the direction we intend to head

ref: https://github.com/kubernetes/kubernetes/issues/97025#issuecomment-737530281


hasheddan (approved):
/lgtm
/hold for others to weigh in

xmudrii (approved):
This seems like an okay workaround.
/lgtm

spiffxp:
https://github.com/kubernetes/test-infra/issues/20101 is the issue for what we'd like to do as a proper fix

saschagrunert (approved):
/lgtm

PR #20109, opened 2020-12-03
Use an options to gather the flags for label_sync

hi-rustin:
Fixes https://github.com/kubernetes/test-infra/blob/7ad5b27f2631eb09d15db54390c144ba6ae9e300/label_sync/main.go#L114


fejta (approved):
thanks!

PR #20117, opened 2020-12-03
Continue to test stable version of kops

justinsb:
We can continue to test older versions of kops that are still in
widespread use, to catch regressions.

Can also give signal about when problems were introduced.

### DIFF
--- a/config/jobs/kubernetes/kops/build-grid.py
+++ b/config/jobs/kubernetes/kops/build-grid.py
@@ -354,6 +354,7 @@ k8s_versions = [
 
 kops_versions = [
     None, # maps to latest
+    "1.18",
     "1.19",
 ]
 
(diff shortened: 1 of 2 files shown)


justinsb:
WIP because I think we should do this, but I want to run it by @rifelpet and @hakman who removed 1.18 previously.

I think having more testing in the grid is good.  I want to consider the resource requirements separately (i.e. if we need to halve the rate at which these tests run, we can do that).  Another effect is that some tests will run less often, particularly on the master branch which is more likely; if we are concerned with that we could also start to split into buckets and/or special-case a few.

But first: should we be testing the older versions as well? :-)

rifelpet:
Continuing to test 1.18 is fine with me. Once the conflicts are resolved I'm happy to lgtm this

hakman:
Same here. :)

justinsb:
Thanks @hakman and @rifelpet ... rebased and removed WIP

PR #20122, opened 2020-12-03
Bump go to 1.15

alvaroaleman:
/assign @fejta 

Not sure if this is correct, I just noticed that the go version we use in CI seems to be different from what I use locally which is annoying if it fails on things that work perfectly fine locally.


alvaroaleman:
I have no idea how to interpret that java stacktrace, if someone who knows bazel could chime in, that would be great.

stevekuznetsov (approved):

PR #20124, opened 2020-12-04
kops-grid: add one-off test for AWS External CloudController

justinsb:
We know there are problems, so we're not introducing it into the
broader set of permutations yet.


justinsb:
WIP because we need the cluster-name length limiting  from #20097

justinsb:
cc @nckturner

justinsb:
Required PR merged, so rebased and removed WIP!

mikesplain (approved):

PR #20131, opened 2020-12-04
kops-grid: build annotations using code instead of templates

justinsb:
This will allow us to put our own annotations in there, which is
important for analyzing the results without string-parsing the name.


justinsb:
/retest

prow networking flake

mikesplain (approved):

PR #20151, opened 2020-12-07
Setting lgtm_acts_as_approve to false for service-apis

robscott:
Although it appears to be more common for `lgtm` to be equivalent to `approve` for approvers, this seems like it could result in some accidental approves. Disabling this will help ensure all service-apis PRs are intentionally approved. More context in https://kubernetes.slack.com/archives/CR0H13KGA/p1607361291247300.

/cc @hbagdi @bowei


hbagdi (approved):
My approval probably doesn't count here but I want to say yes to this change.
Thanks Rob for the quickly tracing this down and fixing in a matter of minutes!

PR #20158, opened 2020-12-07
azure: remove 2004 containerd hyperv jobs

chewong:

/assign @marosset


jsturtevant:
Can we just remove them instead of commenting?  If needed we can always look back at the older commits

chewong:
> Can we just remove them instead of commenting? If needed we can always look back at the older commits

done!

PR #20161, opened 2020-12-08
Add support for defaulting Prow monitoring config, decouple from prow.k8s.io, and update README port-forwarding instructions to include grafana.

cjwagner:
/assign @chaodaiG 

From now on we should avoid hardcoding values specific to the prow.k8s.io instance and instead add new config fields as needed.  Config defaulting logic and helper functions can be added to the `defaultConfig` function in the `config_util.libsonnet` file.
New config fields should provide any defaulting logic needed to ensure that prometheus rules and grafana dashboards that use them behave well when no value is specified. This will allow us to safely update the monitoring stack for different instances by simply copying the latest rules and dashboards from upstream without having to update config.

### DIFF
--- a/config/prow/cluster/monitoring/mixins/lib/config.libsonnet
+++ b/config/prow/cluster/monitoring/mixins/lib/config.libsonnet
@@ -1,80 +1,77 @@
-{
-  _config+:: {
-    // Grafana dashboard IDs are necessary for stable links for dashboards
-    grafanaDashboardIDs: {
-      'boskos-http.json': 'eec46c579cbf4a518e5bbcbbf4913de9',
-      'ghproxy.json': 'd72fe8d0400b2912e319b1e95d0ab1b3',
-      'slo.json': 'ea313af4b7904c7c983d20d9572235a5',
-    },
-    // Component name constants
-    components: {
-      // Values should be lowercase for use with prometheus 'job' label.
-      crier: 'crier',
-      deck: 'deck',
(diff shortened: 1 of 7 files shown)


papuneftali (approved):

papuneftali (approved):

config/prow/cluster/monitoring/mixins/lib/config.libsonnet
```diff
@@ -1,80 +1,77 @@
...
+  // How long we go during work hours without seeing a webhook before alerting.
+  webhookMissingAlertInterval: '10m',
+
+  // How many days prow hasn't been bumped.
+  prowImageStaleByDays: 7,
+};
+
+// Generate the real config by adding in constant fields and defaulting where needed.
+{
+  _config+:: util.defaultConfig(util.consts + config),
```
chaodaiG, line 75: Nit, this seems a bit excessive to me to call method from another module with the value from the other module, is there a way to make `defaultConfig` always use `consts`?

chaodaiG:
Overall looks good to me, left a nit

/lgtm
/approve
/hold
If you want to deal with my nit

config/prow/cluster/monitoring/mixins/lib/config.libsonnet
cjwagner: Yeah thats part of the reason I moved the constants to the new file. I forgot to change this. Fixed.

PR #20163, opened 2020-12-08
Code cleanup in projectmanager

hi-rustin:
Clean up projectmanager useless code and fix typos.

### DIFF
--- a/prow/plugins/projectmanager/projectmanager.go
+++ b/prow/plugins/projectmanager/projectmanager.go
@@ -35,7 +35,7 @@ const (
 
 var (
 	failedToAddProjectCard = "Failed to add project card for the issue/PR"
-	issueAlreadyInProject  = "The issue/PR %d already assigned to the project %s"
+	issueAlreadyInProject  = "The issue/PR %s already assigned to the project %s"
 
 	handleIssueActions = map[github.IssueEventAction]bool{
 		github.IssueActionOpened:    true,
@@ -66,7 +66,7 @@ org/repos:
                   area/sig-testing
 */
 // TODO Handle Label deletion, pr/issue should be removed from the project when label criteria does  not meet
-// TODO Pr/issue state change, pr/iisue is on project board only if its state is listed in the configuration
+// TODO Pr/issue state change, pr/issue is on project board only if its state is listed in the configuration
 func init() {
 	plugins.RegisterIssueHandler(pluginName, handleIssueOrPullRequest, helpProvider)
 }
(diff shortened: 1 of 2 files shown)


prow/plugins/projectmanager/projectmanager.go
```diff
@@ -35,7 +35,7 @@ const (
 
 var (
 	failedToAddProjectCard = "Failed to add project card for the issue/PR"
-	issueAlreadyInProject  = "The issue/PR %d already assigned to the project %s"
+	issueAlreadyInProject  = "The issue/PR %s already assigned to the project %s"
```
petr-muller, line 38: This looks like this should be a number?

prow/plugins/projectmanager/projectmanager_test.go
```diff
@@ -747,15 +747,15 @@ func checkCards(expectedColumnCards, projectColumnCards map[int][]github.Project
...
+				return fmt.Errorf("unable to find project card: %v under column: %d", card, columnID)
 			}
 		}
 	}
 	return nil
 }
 
 func TestHelpProvider(t *testing.T) {
-	var i int = 0
+	var i = 0
```
petr-muller, line 758: nit: `var i int` is IMO a bit more idiomatic

prow/plugins/projectmanager/projectmanager.go
petr-muller: Ah ok, it's a URL. Nevermind.

petr-muller (approved):

fejta (approved):
/hold

petr-muller:
@hi-rustin also, a suggestion - calling code "useless" is a little loaded. Not entirely hostile, but it is still an unnecessary judgement. I would suggest to use a little bit neutral "unused", or simply call it "Code cleanup in projectmanager" or something similar.

prow/plugins/projectmanager/projectmanager_test.go
fejta: https://dave.cheney.net/practical-go/presentations/qcon-china.html#_use_a_consistent_declaration_style is the source I cite for this

fejta:
> I would suggest to use a little bit neutral "unused", or simply call it "Code cleanup in projectmanager" or something similar.

"Simplify unnecessary `(foo) to foo`" is how I would bikeshed the wording. I'd consider the differences in English between unused/unnecessary/useless to be pretty subtle :)

prow/plugins/projectmanager/projectmanager_test.go
hi-rustin: Fixed

hi-rustin:
@petr-muller  @fejta  Thanks for your review. I changed the commit message to `Code cleanup in projectmanager`.

petr-muller (approved):
/hold cancel

PR #20164, opened 2020-12-08
[KEP-1933] Alpha -> Beta Graduation

mlevesquedion:
This PR updates the `verify-govet-levee` check to run automatically as a non-blocking presubmit. This represents the "Beta" step in [KEP-1933](https://github.com/kubernetes/enhancements/tree/master/keps/sig-instrumentation/1933-secret-logging-static-analysis): 

> Analysis runs as a non-blocking presubmit check, warning developers of any findings in their changes.

Apha -> Beta Graduation criteria described in the [KEP](https://github.com/kubernetes/enhancements/tree/master/keps/sig-instrumentation/1933-secret-logging-static-analysis) are as follows:
> - Test is validated as running soundly.
> - Analysis configuration consumes fields tags introduced by [KEP-1753](https://github.com/kubernetes/enhancements/blob/master/keps/sig-instrumentation/1753-logs-sanitization/README.md) for identification of material that should not be logged. Configuration should also identify other sensitive data not bound to fields, if any are identified during KEP-1753's investigation.

@PurelyApplied and I have validated that the analysis is running soundly: the analysis is not currently producing any reports, but it produces a (non-blocking) test failure if a developer introduces a new credentials leak (demonstration PR [here](https://github.com/kubernetes/kubernetes/pull/97213)). The current configuration has recently been updated (merged PR [here](https://github.com/kubernetes/kubernetes/pull/96997)) and contains both the field tags and other sensitive data identified during KEP-1753's investigation.

### DIFF
--- a/config/jobs/kubernetes/sig-instrumentation/verify-govet-levee.yaml
+++ b/config/jobs/kubernetes/sig-instrumentation/verify-govet-levee.yaml
@@ -7,7 +7,7 @@ presubmits:
     skip_branches:
       - release-\d+.\d+ # per-release job
     optional: true
-    skip_report: true
+    skip_report: false
     path_alias: k8s.io/kubernetes
     annotations:
       testgrid-create-test-group: "true"
--- a/config/jobs/kubernetes/sig-instrumentation/verify-govet-levee.yaml
+++ b/config/jobs/kubernetes/sig-instrumentation/verify-govet-levee.yaml
@@ -2,12 +2,12 @@ presubmits:
   kubernetes/kubernetes:
   - name: pull-kubernetes-verify-govet-levee
     cluster: k8s-infra-prow-build
-    always_run: false
+    always_run: true
     decorate: true
     skip_branches:
       - release-\d+.\d+ # per-release job
     optional: true
-    skip_report: false
+    skip_report: true
     path_alias: k8s.io/kubernetes
     annotations:
       testgrid-create-test-group: "true"
--- a/gubernator/config.yaml
+++ b/gubernator/config.yaml
@@ -37,4 +37,5 @@ jobs:
   - pull-kubernetes-node-e2e
   - pull-kubernetes-typecheck
   - pull-kubernetes-verify
+  - pull-kubernetes-verify-govet-levee
 recursive_artifacts: false


PurelyApplied:
Sig-security is in the process of taking ownership of this from sig-instrumentation.  Adding sig tag.

/sig security

ehashman:
/ok-to-test

LGTM

mlevesquedion:
Thanks for giving the ok-to-test!

I've fixed the broken test involving gubernator.

mlevesquedion:
@BenTheElder do these changes look ok to you?

config/jobs/kubernetes/sig-instrumentation/verify-govet-levee.yaml
```diff
@@ -2,7 +2,7 @@ presubmits:
   kubernetes/kubernetes:
   - name: pull-kubernetes-verify-govet-levee
     cluster: k8s-infra-prow-build
-    always_run: false
+    always_run: true
```
BenTheElder, line 5: this should actually be marked `optional: true` if it has not been always_run / we don't have data that this specific configuration passes reliably (sorry if this was miscommunicated but testing it once on one PR isn't really what we're aiming for there).

the reason is, switching this to always_run true / implicit optional: false will make it required for merge.
even if that's your intended end-goal, it should only be true once it's been running reliably. we usually soak for at least a week.

it can switch to non-optional later.
both need to be outside of the test-freeze window (could not have done this much sooner than now)

mlevesquedion: It is marked `optional: true` a couple lines below (on line 9, specifically), but it's not apparent in the diff shown by GitHub. It was `optional: true` before I made the change to `always_run`.

Is this ok?

I agree that this should not block PRs for now. Blocking PRs would not be coherent with the Beta target, which states:

> Analysis runs as a non-blocking presubmit check

BenTheElder: thanks 😅

BenTheElder:
/lgtm
/approve

I don't want to block this further, but I see `imagePullPolicy: Always` reviewing the full job config, and I don't think that should be the case anymore / it's not using a mutable tag. We should drop that.

PR #20185, opened 2020-12-09
Update config validation checks to be more correct

stevekuznetsov:
config: correctly validate mount paths

Only some of the possible decoration mounts will be makde available to
the test container provided by the user, so only check for those.

---

config: fix volume definition check

If a user's mounting a volume we add to the container in decoration, it
will be valid.

---

/assign @alvaroaleman

### DIFF
--- a/prow/config/config_test.go
+++ b/prow/config/config_test.go
@@ -884,7 +884,7 @@ func TestValidatePodSpec(t *testing.T) {
 			spec: func(s *v1.PodSpec) {
 				s.Containers[0].VolumeMounts = append(s.Containers[0].VolumeMounts, v1.VolumeMount{
 					Name:      "fun",
-					MountPath: decorate.VolumeMountPaths().List()[0],
+					MountPath: decorate.VolumeMountPathsOnTestContainer().List()[0],
 				})
 			},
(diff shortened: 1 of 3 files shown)


prow/config/config_test.go
```diff
@@ -902,7 +902,25 @@ func TestValidatePodSpec(t *testing.T) {
...
 					Name:      "foo",
-					MountPath: filepath.Join(decorate.VolumeMountPaths().List()[0], "extra"),
+					MountPath: filepath.Join(decorate.VolumeMountPathsOnTestContainer().List()[0], "extra"),
+				})
+			},
+		},
+		{
+			name: "accept non-conflicting mount path",
+			spec: func(s *v1.PodSpec) {
+				s.Containers[0].VolumeMounts = append(s.Containers[0].VolumeMounts, v1.VolumeMount{
```
alvaroaleman, line 912: These two tests test the same thing, did you mean to set the `Volume` here?

stevekuznetsov: The meaning was different but yeah it's the same underlying data

alvaroaleman (approved):

Issue #20184, opened 2020-12-09
StatusReconciler Panic: index out of range [1] with length 1


clarkmcc:
**What happened**:
Panic out of bounds error

**What you expected to happen**:
No panic

**How to reproduce it (as minimally and precisely as possible)**:
No idea. All other services/deployments appear to be running fine. I'm happy to provide any other information that would be helpful.

**Please provide links to example occurrences, if any**:
```shell
panic: runtime error: index out of range [1] with length 1 goroutine 75 [running]: 
    k8s.io/test-infra/prow/statusreconciler.(*Controller).retireRemovedContexts(0xc000360f60, 0xc000582f60, 0xc000582f60, 0x0) 
    prow/statusreconciler/controller.go:285 +0x79d k8s.io/test-infra/prow/statusreconciler.(*Controller).reconcile(0xc000360f60, 0x0, 0x0, 0x0, 0xc0004aa3f0, 0x0, 0xc00045d900, 0x1, 0x4, 0x0, ...) 
    prow/statusreconciler/controller.go:191 +0xcc k8s.io/test-infra/prow/statusreconciler.(*Controller).Run(0xc000360f60, 0x20cfb60, 0xc0003b2a40) 
    prow/statusreconciler/controller.go:170 +0x325 main.main.func1(0x20cfb60, 0xc0003b2a40) 
    prow/cmd/status-reconciler/main.go:164 +0x3e k8s.io/test-infra/prow/interrupts.Run.func1(0xc0003d6390, 0x20cfb60, 0xc0003b2a40) 
    prow/interrupts/interrupts.go:142 +0x6e created by k8s.io/test-infra/prow/interrupts.Run prow/interrupts/interrupts.go:140 +0xb0
```

**Anything else we need to know?**:
I've been getting this issue repeatedly for the last 24 hours or so. Everything else appears to be working. Jobs are running fine, PRs are being merged correctly, and chatops is working great as well.

clarkmcc:
After further investigation, it looks like the error is coming from

https://github.com/kubernetes/test-infra/blob/3846aa0d7dba8b2ce27c58e9751a9f4047ec9b89/prow/statusreconciler/controller.go#L281-L285

I only have one presubmit in_repo_config that looks like this

```yaml
presubmits:
- name: go-test
  decorate: true
  spec:
    containers:
      - image: golang:1.15-alpine
        command:
          - /bin/sh
        args:
          - ./hack/test-go.sh
```

[...]

PR #20187, opened 2020-12-09
Statusreconciler: Don't panic on invalid orgrepo string

alvaroaleman:
Shouldn't be happening in the first place but we shouldn't panic if it does

Fixes https://github.com/kubernetes/test-infra/issues/20184


alvaroaleman:
Fixes https://github.com/kubernetes/test-infra/issues/20184

PR #20188, opened 2020-12-09
Fix secret token path to be absolute and move autobump config file

mpherman2:
The token in the autobump-config file was using a relative path instead of an absolute one.


fejta (approved):
/ok-to-test

Issue #18973, opened 2020-08-25
node-kubelet-master and node-kubelet-conformance seem like duplicates


SergeyKanzhelev:
CC: @MHBauer 

**What should be cleaned up or changed**:

node-kubelet-master and node-kubelet-conformance seem like duplicates. Need to investigate on whether we need both or some restructure is needed.

**Provide any links for context**:

Discussed that needs to be investigated during today's meting: https://docs.google.com/document/d/1fb-ugvgdSVIkkuJ388_nhp2pBTy_4HEVg5848Xy7n5U/edit#

knabben:
my 2 cents, the `node-kubelet-conformance` uses the `--test-suite=conformance` flag when starting the test. 

Without the flag `node-kubelet-master` the command only copy the binaries ("kubelet", "e2e_node.test", "ginkgo", "mounter") to the tar directory (the tar file is created for the e2e node, after copying it).

On the other hand, the flag --test-suite=conformance as value, builds the entire docker image with --conformance=true on Ginkgo as the image Entrypoint, what is not very clear is the usage of a docker image to build this.

mmerkes:
Below is a diff (without analysis) of the two different jobs. I'll update this post or follow up with a separate one with more information on the differences.

**node-kubelet-master**

*+annotations*

- `fork-per-release: 'true'`
- `fork-per-release-periodic-interval: 1h 2h 6h 24h`
- `testgrid-alert-email: kubernetes-sig-node+testgrid@googlegroups.com`
- `testgrid-dashboards:`
  - `sig-release-master-blocking`
  - `sig-node-release-blocking`

*interval*

`interval: 1h` vs `interval: 2h` for conformance

*container args*

- `--test_args=--nodes=8` vs. `--test_args=--nodes=8`
- `--skip="\[Flaky\]|\[Serial\]"` vs. `--skip="\[Flaky\]"`

*+containers.resources*

```
      resources:
        limits:
          cpu: 6
          memory: 6Gi
        requests:
          cpu: 6
          memory: 6Gi
```

**node-kubelet-conformance**

*+container args*

- `--gcp-project-type=node-e2e-project`
- `--test-suite=conformance`
- `--node-test-args=--feature-gates=DynamicKubeletConfig=true`

*+containers.securityContext*

```
      # docker-in-docker needs privileged mode
      securityContext:
        privileged: true
```

mmerkes:
I found a little more information on the reasoning behind running conformance tests in Docker:

[...]

PR #20198, opened 2020-12-10
Removes the --test-suite=conformance flag from node-kubelet-conformance job

mmerkes:
While [investigating a job failure](https://github.com/kubernetes/kubernetes/issues/97130), we found out that the `node-kubelet-conformance` job wasn't actually running the correct tests. Setting the `--test-suite=conformance` resulted in ginkgo flags getting ignored and using the `[Conformance]` focus instead. More details [here](https://github.com/kubernetes/kubernetes/issues/97130).

This should fix the current job failures, and I will discuss longer term steps with the SIG-Node CI subgroup.

### DIFF
--- a/config/jobs/kubernetes/sig-node/node-kubelet.yaml
+++ b/config/jobs/kubernetes/sig-node/node-kubelet.yaml
@@ -88,7 +88,7 @@ periodics:
       - --deployment=node
       - --gcp-project-type=node-e2e-project
       - --gcp-zone=us-west1-b
-      - --node-args=--image-config-file=/workspace/test-infra/jobs/e2e_node/image-config.yaml --test-suite=conformance
+      - --node-args=--image-config-file=/workspace/test-infra/jobs/e2e_node/image-config.yaml
       - --node-test-args=--feature-gates=DynamicKubeletConfig=true --kubelet-flags="--cgroups-per-qos=true --cgroup-root=/"
       - --node-tests=true
       - --provider=gce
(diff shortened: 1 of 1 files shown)


mmerkes:
I signed it

knabben:
Removing the `--test-suite=conformance` will make it equivalent with `kubelet-master`, for that matter makes more sense to delete the entire test.

The question is: Should we be testing only the `NodeConformance` tests in a `conformance` suite?

mmerkes:
> Removing the `--test-suite=conformance` will make it equivalent to `kubelet-master`, in this case makes more sense to delete the entire test.
> 
> The question is: Should we have a job running `NodeConformance` tests in a `conformance` suite?

While I don't disagree, I see this PR as the first step. This will make the job green and buy us time without doing anything too drastic. We can take this back to SIG-Node or the CI subgroup and make a decision on the longer term plan, which should either be to remove the job entirely (and clean up all of the related code) or figure out the identity and role of the node conformance tests and make that a reality.

knabben:
/lgtm

agreed on the short term plan for unblocking the lane

MHBauer:
My immediate reaction is "If it doesn't have that flag, it's not running the conformance suite, which does a bunch of different operations for conformance".

I need to read into the details in the issue and refresh myself.

Otherwise my reaction is the same as https://github.com/kubernetes/test-infra/pull/20198#issuecomment-742703140, and I believe I asked why there are multiple different "conformance" jobs some months ago, with the answer being some version of "they do different things".

mmerkes:
@MHBauer Yes, you did ask that very question. That issue is [here](https://github.com/kubernetes/test-infra/issues/18973). However, we never closed the loop on that one and made a decision. This issue is bringing up that conversation again.

My immediate concern is turning the job green. It doesn't actually do what we want today given that it ignores the `NodeConformance` tag anyway, so this feels ok to me. Anything tagged `[Conformance]` will already be run in other conformance jobs.

That being said, if we're concerned about the final desired state not being tied to this PR, we can sit on this one and either remove it entirely or make it do what we want, once we figure out what exactly that is.

SergeyKanzhelev (approved):

SergeyKanzhelev:
/lgtm

as discussed on the meeting today let's make it green first and will work on plan

SergeyKanzhelev:
/ok-to-test

(to clear up label)

PR #20205, opened 2020-12-10
Github: Add apps auth support for the v4 client

alvaroaleman:
This change adds apps auth support to the Github V4 client. This is
achieved by adding a new method that takes an additional org argument,
then using the context to pass this through to the underlying
roundtripper. For consistency, the client was changed to use the Context
for this everywhere instead of a header on the V3 api.

We use the context to pass the org on because the code that constructs
the V4 requests is part of a dependency, giving us no way to use a
Header. (Ab-)using the context for this allows us to get away without
requiring changes to our dependencies.

The second commit makes blunderbuss use this by means of a wrapper that sets the "org" attribute based off of the event the request came from.

### DIFF
--- a/prow/github/client.go
+++ b/prow/github/client.go
@@ -238,6 +238,7 @@ type Client interface {
 
 	Throttle(hourlyTokens, burst int)
 	Query(ctx context.Context, q interface{}, vars map[string]interface{}) error
+	QueryWithGitHubAppsSupport(ctx context.Context, q interface{}, vars map[string]interface{}, org string) error
 
 	SetMax404Retries(int)
 
@@ -352,7 +353,7 @@ type httpClient interface {
 
 // Interface for how prow interacts with the graphql client, which we may throttle.
 type gqlClient interface {
-	Query(ctx context.Context, q interface{}, vars map[string]interface{}) error
+	QueryWithGitHubAppsSupport(ctx context.Context, q interface{}, vars map[string]interface{}, org string) error
 }
 
 // throttler sets a ceiling on the rate of GitHub requests.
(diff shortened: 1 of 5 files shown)


prow/github/client.go
```diff
@@ -238,6 +238,7 @@ type Client interface {
 
 	Throttle(hourlyTokens, burst int)
 	Query(ctx context.Context, q interface{}, vars map[string]interface{}) error
+	QueryWithGithubAppsSupport(ctx context.Context, q interface{}, vars map[string]interface{}, org string) error
```
alvaroaleman, line 241: Definitely open for a better name btw :shrug:

matthyx:
> We use the context to pass the org

Not sure I like it, but if there is no other way...
Dumb question, you cannot use injection like it would be done in SpringBoot?

alvaroaleman:
> Dumb question, you cannot use injection like it would be done in SpringBoot?

Its years since I have last used Java and I've never used SpringBoot so not sure if I can give a good answer to that ;)

The basic problem we have is that we have a `http.RoundTripper` that is used inside all github clients and that adds authentication to all requests. To be able to do that, it needs to know the github organization the request is for (meaning either it targets that org or it was made because of an event in that org).

The `http.RoundTripper` gets a `http.Request` which means it has to infer the org from the request (Wrapping the whole client and then passing that info from the wrapper into the roundTripper outside of the request would require a global lock for thread safety and that in turn would heavily degrade performance). 

In the original implementation for the v3 github client I just made the client set a header, which was then read by the RoundTripper. In the V4 case, we import the client from a dependency (and wrap it a bit for ratelimiting and similiar). The interface it offers is:
```
func (c *Client) Query(ctx context.Context, q interface{}, variables map[string]interface{}) error
```

That imported client then constructs the request, giving us no way of setting headers. The only other way I see  to pass the org on would be to put it somehow into the query or into `variables`, but then the RoundTripper would need to deserialize the request body (which for all it knows is an arbitrary `map[string]interface{}`), extract the `org` from there and construct a new body without the org and that seems a lot worse to me.

If I've missed anything there or you have another idea on how we could do this, I'm all ears :)

matthyx:
nah, fine 👍 
/lgtm

alvaroaleman:
@matthyx I had to rebase, could you re-lgtm?

PR #20208, opened 2020-12-11
Back up kubeconfig secret in GCP secret manager

chaodaiG:
This script overrides kubeconfig secret in cluster, and since kubernetes cluster doesn't backup secret, the old value could get lost. Utilize the versioning of GCP secret manager for backing up, so that old history won't get lost

### DIFF
--- a/gencred/BUILD.bazel
+++ b/gencred/BUILD.bazel
@@ -1,3 +1,4 @@
+load("@py3_deps//:requirements.bzl", "requirement")
 load("@io_bazel_rules_go//go:def.bzl", "go_binary", "go_library")
 
 filegroup(
@@ -33,4 +34,13 @@ go_binary(
     name = "gencred",
     embed = [":go_default_library"],
     visibility = ["//visibility:public"],
+)
+
+py_binary(
+    name = "merge_kubeconfig_secret",
+    srcs = ["merge_kubeconfig_secret.py"],
+    python_version = "PY3",
+    deps = [
+        requirement("google-cloud-secret-manager"),
+    ],
 )
--- a/gencred/merge_kubeconfig_secret.py
+++ b/gencred/merge_kubeconfig_secret.py
@@ -17,7 +17,11 @@
 """Merges a kubeconfig file into a kubeconfig file in a k8s secret giving
 precedence to the secret."""
 
-# Requirements: kubectl (pointed at the correct cluster) and base64
+# Requirements:
+#   - kubectl (pointed at the correct cluster)
+#   - base64
+#   - google-cloud-secret-manager authenticated via ADC by running:
+#    `unset GOOGLE_APPLICATION_CREDENTIALS && gcloud auth login && gcloud auth application-default login`
 # Example usage:
 # ./merge_kubeconfig_secret.py \
 #   --name=mykube \
@@ -38,8 +42,8 @@ import time
 # Requires `pip install google-cloud-secret-manager`
 from google.cloud import secretmanager
 
-reAutoKey = re.compile('^config-(\\d{8})$')
-gcp_secret_key = "prowkubeconfigbackup"
+RE_AUTO_KEY = re.compile('^config-(\\d{8})$')
+GCP_SECRET_KEY = "prowkubeconfigbackup"
 
 
 def call(cmd, **kwargs):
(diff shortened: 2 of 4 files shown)


gencred/merge_kubeconfig_secret.py
```diff
@@ -34,7 +34,13 @@
 import tempfile
 import time
 
+# Import the Secret Manager client library.
+# Requires `pip install google-cloud-secret-manager`
+from google.cloud import secretmanager
+
 reAutoKey = re.compile('^config-(\\d{8})$')
+gcp_secret_key = "prowkubeconfigbackup"
```
michelle192837, line 42: nit: Should this be ALL_CAPS?

chaodaiG:
POC
Screen Shot 2020-12-10 at 4 31 04 PM

Screen Shot 2020-12-10 at 4 31 21 PM

gencred/merge_kubeconfig_secret.py
```diff
@@ -44,11 +50,31 @@ def call(cmd, **kwargs):
...
+        timeout=10,  # seconds
         universal_newlines=True,
         **kwargs,
     )
 
+
+def save_secret_to_gcp(project_id, data, secret_id=gcp_secret_key):
+    dst = f"projects/{project_id}/secrets/{secret_id}"
+    print('Saving secret to %s' % dst)
+    client = secretmanager.SecretManagerServiceClient()
```
michelle192837, line 62: Huh, no auth needed?

gencred/merge_kubeconfig_secret.py
```diff
@@ -114,12 +151,18 @@ def validateArgs(args):
 
 
 if __name__ == '__main__':
-    parser = argparse.ArgumentParser(description='Merges the provided kubeconfig file into a kubeconfig file living in a kubernetes secret in order to add new cluster contexts to the secret. Requires kubectl and base64.') #pylint: disable=line-too-long
+    parser = argparse.ArgumentParser(
+        description='Merges the provided kubeconfig file into a kubeconfig file living in a kubernetes secret in order to add new cluster contexts to the secret. Requires kubectl and base64.')  # pylint: disable=line-too-long
```
michelle192837, line 155: This should be updated to note the new google-cloud-secret-manager requirement.

chaodaiG: This is good to know, updated

gencred/merge_kubeconfig_secret.py
chaodaiG: Actually it's needed, it requires ADC auth which is done by gcloud, I have added the required command in the description below, as well as `requirement` section on top

gencred/merge_kubeconfig_secret.py
chaodaiG: Done, to be more careful maybe they should be `_` prefixed, but I'm not imagining anyone importing this file, so let's keep them as they are

chaodaiG:
Something is not right in the test, it failed with:
```
Collecting google-cloud-secret-manager==2.1.0 (from -r /home/prow/go/src/github.com/kubernetes/test-infra/requirements3.txt (line 13))
 (  Could not find a version that satisfies the requirement google-cloud-secret-manager==2.1.0 (from -r /home/prow/go/src/github.com/kubernetes/test-infra/requirements3.txt (line 13)) (from versions: 0.1.0, 0.1.1, 0.2.0, 1.0.0)
```

But this was added as python3 requirements, which 2.1.0 is [officially supported](https://pypi.org/project/google-cloud-secret-manager/) version. The official document also says `The last version of this library compatible with Python 2.7 is google-cloud-secret-manager==1.0.0.`,  the error message `from versions: 0.1.0, 0.1.1, 0.2.0, 1.0.0` indicates that the test was fetching python3 dependencies using python2.

gencred/BUILD.bazel
```diff
@@ -34,3 +35,12 @@ go_binary(
     embed = [":go_default_library"],
     visibility = ["//visibility:public"],
 )
+
+py_binary(
+    name = "merge_kubeconfig_secret",
+    srcs = ["merge_kubeconfig_secret.py"],
+    python_version = "PY3",
+    deps = [
+        requirement("google-cloud-secret-manager"),
```
fejta, line 44: google-cloud-secret-manager depends on asteroid, thus you need to add `requirement('asteroid`)` in here (as well as any other dependencies

chaodaiG: Had an offline discussion, the error message with `asteroid` was probably a misleading error, which very likely was a truncated error message meant to print https://github.com/kubernetes/test-infra/blob/master/requirements3.txt but only got to the first line.

chaodaiG:
No clue how to resolve the bazel test error, will park this PR for here for now

fejta:
My three thoughts are:
1) Golang is generally going to be a better choice around k8s than python
2) Despite the error message, I suspect a missing transitive dependency in the requirements -- either in requirements.txt or more likely the require() build rule
3) Perhaps there's a stale version of the pip project list

PR #20211, opened 2020-12-11
Fix cri-tools post summit job which is failing

hickeyma:
Fixes failing post summit job https://testgrid.k8s.io/sig-node-cri-tools#post-cri-tools-push-image-user-multi-arch

Also, renames the prow config file to standardize with other files.

/kind bug


hickeyma:
Thanks @chaodaiG

PR #20224, opened 2020-12-13
skip e2e test with hostNetwork pods

aojea:
the dns jobs using nodecachei, are failing because the test:

[sig-network] Networking Granular Checks: Services should
function for service endpoints using hostNetwork

has already a service listening on the ports used by the test,
causing that the pods under tests fail to be deployed due to a
port conflict and failing the test.

Since there is no guarantee to choose a port for the test that
doesn't conflict with any other service in the cluster, better we
skip the test in this job.

These are the jobs that are failing
https://testgrid.k8s.io/sig-network-gce#gci-gce-coredns-nodecache
https://testgrid.k8s.io/sig-network-gce#gci-gce-kube-dns-nodecache

### DIFF
--- a/config/jobs/kubernetes/sig-network/sig-network-misc.yaml
+++ b/config/jobs/kubernetes/sig-network/sig-network-misc.yaml
@@ -242,7 +242,7 @@ periodics:
       - --gcp-zone=us-west1-b
       - --ginkgo-parallel=30
       - --provider=gce
-      - --test_args=--ginkgo.skip=\[Serial\]|\[Disruptive\]|\[Flaky\]|\[Feature:.+\] --minStartupPods=8
+      - --test_args=--ginkgo.skip=\[Serial\]|\[Disruptive\]|\[Flaky\]|\[Feature:.+\]|hostNetwork --minStartupPods=8
       - --timeout=150m
       image: gcr.io/k8s-testimages/kubekins-e2e:v20201208-8b354d9-master
 
@@ -443,7 +443,7 @@ periodics:
       - --gcp-zone=us-west1-b
       - --ginkgo-parallel=30
       - --provider=gce
-      - --test_args=--ginkgo.skip=\[Serial\]|\[Disruptive\]|\[Flaky\]|\[Feature:.+\] --minStartupPods=8
+      - --test_args=--ginkgo.skip=\[Serial\]|\[Disruptive\]|\[Flaky\]|\[Feature:.+\]|hostNetwork --minStartupPods=8
       - --timeout=150m
       image: gcr.io/k8s-testimages/kubekins-e2e:v20201208-8b354d9-master
 


prameshj:
Looks like these tests have been failing for a while.

From one of the logs, i see the webserver container in the client pod is not Ready - https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-e2e-gci-gce-coredns-nodecache/1338208494687358976

Is this due to contention for the "8080" port?

aojea:
> Looks like these tests have been failing for a while.
> 
> From one of the logs, i see the webserver container in the client pod is not Ready - https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-e2e-gci-gce-coredns-nodecache/1338208494687358976
> 
> Is this due to contention for the "8080" port?

yeah, it seems so it fails in 
> k8s.io/kubernetes/test/e2e/framework/network.(*NetworkingTestConfig).createNetProxyPods(0xc007a361c0, 0x4db7759, 0x9, 0xc002ef4f30, 0x0, 0x203000, 0x0)

https://github.com/kubernetes/kubernetes/blob/87984d84d18c7816835e5efedebb17ed13aa2509/test/e2e/framework/network/utils.go#L835-L839

I previously thought, wrongly, that it was failing when creating the service, so I tried to skip if that was the problem
https://github.com/kubernetes/kubernetes/pull/96453

I have to revert that one, since a service is never going to conflict, are the pod when trying to be scheduled the ones that are going to conflict, as we can see here

prameshj:
> > Looks like these tests have been failing for a while.
> > From one of the logs, i see the webserver container in the client pod is not Ready - https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-e2e-gci-gce-coredns-nodecache/1338208494687358976
> > Is this due to contention for the "8080" port?
> 
> yeah, it seems so it fails in
> 
> > k8s.io/kubernetes/test/e2e/framework/network.(*NetworkingTestConfig).createNetProxyPods(0xc007a361c0, 0x4db7759, 0x9, 0xc002ef4f30, 0x0, 0x203000, 0x0)
> 
> https://github.com/kubernetes/kubernetes/blob/87984d84d18c7816835e5efedebb17ed13aa2509/test/e2e/framework/network/utils.go#L835-L839
> 

Thanks, I see this is using port 8080 - https://github.com/kubernetes/kubernetes/blob/87984d84d18c7816835e5efedebb17ed13aa2509/test/e2e/framework/network/utils.go#L54

I guess changing the test to use a different Endpoint Port is a more risky change.  We could change that constant to a different value - 8888, but i see "8080" hard-coded in quite a few YAMLs. Other option is to change NodeLocalDNS healthcheck port in the yaml. I am not sure if that will break any real deployments if they upgrade. Skipping test seems like the best option here.

> I previously thought, wrongly, that it was failing when creating the service, so I tried to skip if that was the problem
> [kubernetes/kubernetes#96453](https://github.com/kubernetes/kubernetes/pull/96453)
> 
> I have to revert that one, since a service is never going to conflict, are the pod when trying to be scheduled the ones that are going to conflict, as we can see here

PR #20229, opened 2020-12-14
Remove tag from firewall rules

marseel:
Sometime there might be a case when we don't have access to get tags from instance groups. In this case we also want to create firewall rules without tag.

### DIFF
--- a/kubetest/gke.go
+++ b/kubetest/gke.go
@@ -622,19 +622,6 @@ func (g *gkeDeployer) ensureFirewall() error {
 	}
 	log.Printf("Couldn't describe firewall '%s', assuming it doesn't exist and creating it", firewall)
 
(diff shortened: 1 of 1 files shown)


mikedanese:
> Sometime there might be a case when we don't have access to get tags from instance groups.

Can you be more specific? What's the scenario?

marseel:
/reopen
/remove-lifecycle rotten
/assign @mikedanese 
If cluster is Autopilot cluster https://cloud.google.com/kubernetes-engine/docs/concepts/autopilot-overview
then we might not even have access to nodes and kubetest can't even get list of instances within nodepool. I think that the only way to create firewall rule is to get rid of tag so the rule is applied to whole network. Do you think it poses security risk, taking into account that it's used only for testing?

Issue #20236, opened 2020-12-14
[Kettle] Streaming build data sometimes results in JSON error


MushuEE:
**What happened**:
During the data stream process, 
```
ERROR:root:Failed to decode request for kubernetes-jenkins/o/pr-logs%2Fpull%2Fkubeflow_pipelines%2F4885%2Fkubeflow-pipeline-backend-test%2F1336900731344523264%2Ffinished.json
Traceback (most recent call last):
  File "/kettle/make_db.py", line 67, in _request
    return resp.json()
  File "/usr/local/lib/python3.6/dist-packages/requests/models.py", line 900, in json
    return complexjson.loads(self.text, **kwargs)
  File "/usr/lib/python3.6/json/__init__.py", line 354, in loads
    return _default_decoder.decode(s)
  File "/usr/lib/python3.6/json/decoder.py", line 339, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/lib/python3.6/json/decoder.py", line 357, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```
**What you expected to happen**:
Build that fail to decode Json are logged and skipped

**How to reproduce it (as minimally and precisely as possible)**:
Attempt to pull from above path `kubernetes-jenkins/o/pr-logs%2Fpull%2Fkubeflow_pipelines%2F4885%2Fkubeflow-pipeline-backend-test%2F1336900731344523264%2Ffinished.json`

**Please provide links to example occurrences, if any**:

**Anything else we need to know?**:
Looks like there is a possibility the paths are getting poorly decoded. 
`https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubeflow_pipelines/4861/kubeflow-pipeline-backend-test/1338602622315139072/finished.json`
vs
`kubernetes-jenkins/o/pr-logs%2Fpull%2Fkubeflow_pipelines%2F4885%2Fkubeflow-pipeline-backend-test%2F1336900731344523264%2Ffinished.json
Traceback (most recent call last):`

MushuEE:
It looks like the request if getting nothing:
```
>>> import requests
>>> r = requests.Session()
>>> r.get(uri, params={'alt': 'media'})
<Response [200]>
>>> resp = r.get(uri, params={'alt': 'media'})
>>> resp.json()
Failure
>>> resp.text
''
```
[...]

PR #20238, opened 2020-12-15
[Kettle] Don't log traceback when empty json object is hit

MushuEE:
/hold this needs to be tested in staging

#20236

Seems that for some paths, pulling with `alt:media` causes no data to be returned for specific job paths. I do not know the root cause but removing this option fixes it. (I also believe it might be time to update the REST uri path)

/area kettle

### DIFF
--- a/kettle/make_db.py
+++ b/kettle/make_db.py
@@ -67,7 +67,7 @@ class GCSClient:
                         return resp.json()
                     except json.decoder.JSONDecodeError:
                         logging.exception('Failed to decode request for %s',
-                                           urllib.parse.unquote(url))
+                                          urllib.parse.unquote(url))
                         return None
                 return resp.text
             except requests.exceptions.RequestException:
--- a/kettle/make_db.py
+++ b/kettle/make_db.py
@@ -66,7 +66,8 @@ class GCSClient:
                     try:
                         return resp.json()
                     except json.decoder.JSONDecodeError:
-                        logging.exception('Failed to decode request for %s', path)
+                        logging.exception('Failed to decode request for %s',
+                                           urllib.parse.unquote(url))
                         return None
                 return resp.text
             except requests.exceptions.RequestException:
@@ -84,7 +85,7 @@ class GCSClient:
         """Get an object from GCS."""
         bucket, path = self._parse_uri(path)
         return self._request(f'{bucket}/o/{urllib.parse.quote(path, "")}',
-                             {'alt': 'media'}, as_json=as_json)
+                             {}, as_json=as_json)
 
     def ls(self,
            path,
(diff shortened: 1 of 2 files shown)


MushuEE:
```
>>> r = x.get(url)
>>> r.json()
{'kind': 'storage#object', 'id': 'kubernetes-jenkins/logs/ci-cri-containerd-cri-validation-windows/1338954069217120256/started.json/1608066507327679', 'selfLink': 'https://www.googleapis.com/storage/v1/b/kubernetes-jenkins/o/logs%2Fci-cri-containerd-cri-validation-windows%2F1338954069217120256%2Fstarted.json', 'mediaLink': 'https://www.googleapis.com/download/storage/v1/b/kubernetes-jenkins/o/logs%2Fci-cri-containerd-cri-validation-windows%2F1338954069217120256%2Fstarted.json?generation=1608066507327679&alt=media', 'name': 'logs/ci-cri-containerd-cri-validation-windows/1338954069217120256/started.json', 'bucket': 'kubernetes-jenkins', 'generation': '1608066507327679', 'metageneration': '1', 'contentType': 'text/plain; charset=utf-8', 'storageClass': 'STANDARD', 'size': '175', 'md5Hash': 'mpUJ33MEQxpPyV4UqrHLUg==', 'crc32c': 'cmSNhg==', 'etag': 'CL+Rx8fy0O0CEAE=', 'timeCreated': '2020-12-15T21:08:27.327Z', 'updated': '2020-12-15T21:08:27.327Z', 'timeStorageClassUpdated': '2020-12-15T21:08:27.327Z'}
>>> r2 = x.get(url, params={'alt':'media'})
>>> r2.json()
{'timestamp': 1608066507, 'repos': {'containerd/cri': 'master'}, 'repo-commit': '68b61297b59e38c1088db10fbd19807a4ffbad87', 'repo-version': '68b61297b59e38c1088db10fbd19807a4ffbad87'}
```

Seems like there are two cases that need to be handled. One is where there is object data, and one where there is not. This make the problem more complicated.

kettle/make_db.py
```diff
@@ -66,7 +66,8 @@ def _request(self, path, params, as_json=True):
                     try:
                         return resp.json()
                     except json.decoder.JSONDecodeError:
-                        logging.exception('Failed to decode request for %s', path)
+                        logging.error('Failed to decode request for %s',
+                                      urllib.parse.unquote(url))
```
spiffxp, line 70: AFAICT this is the only real change in this PR, it's unclear to me how it's related to the description?

MushuEE: Previously it was logging the stack with the log. It made me believe that there was a stack trace being hit because there was no real indication that it was part of the log message. The error is common to all jobs where they have a blank started/finsihed.json. So spamming the logs with traces was obnoxious and I changes it. When I say "trace back logging" i just mean that the trace was being logged to console.

spiffxp: 🤦 I missed the change from `error` to `exception` and was only looking at the parameters.  This makes sense

spiffxp (approved):
/approve
/lgtm

PR #20239, opened 2020-12-15
Add default values for generic autobumper that work for k8s test-infra bump

mpherman2:


experiment/autobumper/bumper/bumper.go
```diff
@@ -139,6 +185,16 @@ type GitAuthorOptions struct {
 	GitEmail string
 }
 
+// UnmarshallWithDefaults unmarshalls yaml config file using default options for unspecified fields
+func UnmarshallWithDefaults(data []byte) (Options, error) {
```
alvaroaleman, line 189: Please unexport this and export `defaultOptions` instead. Ppl may or may not want to use a configfile for this, but they will definitely want to get the defaults, as half of it is essentially constant

/ok-to-test

mpherman2: I left this exported so that I can use it in main.go, but I exported DefaultOptions as well now.

alvaroaleman (approved):

cjwagner:
I think we should restrict defaulting to fields that are not critical dependencies and have a reasonable value that would be suitable for most instances.
- Defaulting fields that represent dependencies like the GitHub token path makes it hard to track important dependencies. Prow components used to default the GH token path, but we switched to having them explicitly specify their token path with flags because this makes the dependency obvious and tells oncall and others where to look. In these cases I think explicit config is preferable to defaults.
- Defaulting fields that don't have a value suitable for all instances doesn't provide utility and makes it possible for an incorrect value to be silently used. For example, the bot login and repo details will need to be different for different instances/use-cases. Defaulting these fields to k8s-prow specific values doesn't help any other instance and doesn't really make things easier for k8s-prow either compared to explicitly defining them in config. Additionally, other instances that omit these fields will incorrectly use the k8s-prow values rather than emit errors as would be desired.

Defaulting is great for things like `consistentImages: true` since that is likely to be suitable across instances and doesn't correspond with a dependency.

fejta:
FWIW I am eager for us to start using this for testgrid releases

alvaroaleman:
Yeah, where I really want the defaulting is for the prefixes (Potentially modulo `{Staging,}RefConfigFile`, I had to look into the code to understand what it does last time and forgot since)

mpherman2:
RefConfigFile is used when the target version is set "upstream" it looks for that file in the upstream version (in this case k8s test-infra) to grab the correct image version to bump to. If you are bumping to targetVersion = "latest" you do not need to provide RefConfigFile. This is probably also falls into the category of something we do not want to default as it is not suitable for all instances. I can give the variable a better name and/or write some better comments to clear up that confusion.

alvaroaleman:
@mpherman2 that is the kind of information that should be a godoc on the fields :)

mpherman2:
> @mpherman2 that is the kind of information that should be a godoc on the fields :)

I will work on improving godoc in another PR. closing this one.

PR #20246, opened 2020-12-15
Do not make logexporter logs public by default

tosi3k:
Remove the ACL argument from logexporter's `gsutil` command invocations so that logs are private by default.

Additionally fix the logexporter's Dockerfile:
* add explicit download of systemd package,
* fix the base image tag (instead of using `latest`) and switch to Debian which is lighter than Fedora,
* download Python 3 instead of deprecated Python 2.

/sig scalability
/assign @jkaniuk @wojtek-t

### DIFF
--- a/logexporter/cmd/Dockerfile
+++ b/logexporter/cmd/Dockerfile
@@ -15,12 +15,13 @@
 # This file builds an image for the log exporter tool. For more info,
 # have a look at the tool's README.md file.
 
-# Base image must have 'journalctl' binary.
-FROM fedora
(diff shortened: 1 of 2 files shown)


tosi3k:
I ran a test with a custom logexporter image built on top of this PR and it works as intended.

jkaniuk:
Could you split it into two commits or even PRs?
Change of base image and fix.

tosi3k:
> Could you split it into two commits or even PRs?
> Change of base image and fix.

Split it into separate commits. Done.

Issue #14130, opened 2019-08-29
"github.com" hardcoded in git client breaks use of internal github appliance


hampus77:
**What happened**:
Specified internal github endpoint for deck and hook and tried to execute a tekton pipeline

**What you expected to happen**:
A git resource to be created and cloned from our internal github

**How to reproduce it (as minimally and precisely as possible)**:
Create a prowjob that has a presubmit that executes tekton 

**Please provide links to example occurrences, if any**:

**Anything else we need to know?**:
The custom endpoint flags work as expected, however, "github.com" is hardcoded at https://github.com/kubernetes/test-infra/blob/master/prow/git/git.go#L87

cjwagner:
The git client you linked to isn't used by the pipeline controller at all. I suspect you are seeing "github.com" used because you are not setting a `clone_uri` field: https://github.com/kubernetes/test-infra/blob/c69eb89915fca776c67b632049d9b834aaeb74f7/prow/cmd/pipeline/controller.go#L558-L567
Example here: https://github.com/kubernetes/test-infra/blob/cab76bdbd3760143a7b45dc35cf0b4d8d43b6c77/prow/pod-utilities.md#how-to-configure

Please reopen this and link to an example occurrence if you are specifying a `clone_uri` that is not being respected.
/close

hampus77:
I´m sorry, I wasn´t really specific but this problem arise before any pipelines are run. This is in response to an pull_request event when I enter the /trigger command:
Hook logs:

{"author":"xxx","component":"hook","event-GUID":"9b415440-ca38-11e9-87c3-41615571cc0b","event-type":"pull_request","file":"prow/plugins/trigger/generic-comment.go:50","func":"k8s.io/test-infra/prow/plugins/trigger.handleGenericComment","level":"debug","msg":"Comment is made by the bot, skipping.","org":"sebshift","plugin":"trigger","pr":3,"repo":"multitenant-logger","time":"2019-08-29T08:40:08Z","url":"https://<CORRECT_URL>/sebshift/multitenant-logger/pull/3 "}

 {"client":"github","component":"hook","file":"prow/github/client.go:522","func":"k8s.io/test-infra/prow/github.(*client).log","level":"info","msg":"GetRef(sebshift, multitenant-logger, heads/master)","time":"2019-08-29T08:40:08Z"}

[...]

PR #20249, opened 2020-12-15
clonerefs: Use repo link from refs if available

taharah:
Dynamically sets the clone URI and path to flex based on the repo link from the ref if it is available. This helps further mitigate hard coded instances of "github.com" as noted on #14130. 

Signed-off-by: Trevor Wood <Trevor.G.Wood@gmail.com>

### DIFF
--- a/prow/pod-utils/clone/clone.go
+++ b/prow/pod-utils/clone/clone.go
@@ -22,7 +22,6 @@ import (
 	"net/url"
 	"os/exec"
 	"path"
-	"regexp"
 	"strconv"
 	"strings"
 	"time"
@@ -111,8 +110,9 @@ func PathForRefs(baseDir string, refs prowapi.Refs) string {
 	if refs.PathAlias != "" {
 		clonePath = refs.PathAlias
 	} else if refs.RepoLink != "" {
-		re := regexp.MustCompile(`^https?://`)
-		clonePath = re.ReplaceAllString(refs.RepoLink, "")
+		// Drop the protocol from the RepoLink
+		parts := strings.Split(refs.RepoLink, "://")
+		clonePath = parts[len(parts)-1]
 	} else {
 		clonePath = fmt.Sprintf("github.com/%s/%s", refs.Org, refs.Repo)
 	}
--- a/prow/pod-utils/clone/clone.go
+++ b/prow/pod-utils/clone/clone.go
@@ -22,6 +22,7 @@ import (
 	"net/url"
 	"os/exec"
 	"path"
+	"regexp"
 	"strconv"
 	"strings"
 	"time"
@@ -109,6 +110,9 @@ func PathForRefs(baseDir string, refs prowapi.Refs) string {
 	var clonePath string
 	if refs.PathAlias != "" {
 		clonePath = refs.PathAlias
+	} else if refs.RepoLink != "" {
+		re := regexp.MustCompile(`^https?://`)
+		clonePath = re.ReplaceAllString(refs.RepoLink, "")
 	} else {
 		clonePath = fmt.Sprintf("github.com/%s/%s", refs.Org, refs.Repo)
 	}
@@ -124,10 +128,17 @@ type gitCtx struct {
 
 // gitCtxForRefs creates a gitCtx based on the provide refs and baseDir.
 func gitCtxForRefs(refs prowapi.Refs, baseDir string, env []string, oauthToken string) gitCtx {
+	var repoURI string
+	if refs.RepoLink != "" {
+		repoURI = fmt.Sprintf("%s.git", refs.RepoLink)
+	} else {
+		repoURI = fmt.Sprintf("https://github.com/%s/%s.git", refs.Org, refs.Repo)
+	}
+
 	g := gitCtx{
 		cloneDir:      PathForRefs(baseDir, refs),
 		env:           env,
-		repositoryURI: fmt.Sprintf("https://github.com/%s/%s.git", refs.Org, refs.Repo),
+		repositoryURI: repoURI,
 	}
 	if refs.CloneURI != "" {
 		g.repositoryURI = refs.CloneURI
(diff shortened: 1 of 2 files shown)


prow/pod-utils/clone/clone.go
```diff
@@ -109,6 +110,9 @@ func PathForRefs(baseDir string, refs prowapi.Refs) string {
 	var clonePath string
 	if refs.PathAlias != "" {
 		clonePath = refs.PathAlias
+	} else if refs.RepoLink != "" {
+		re := regexp.MustCompile(`^https?://`)
+		clonePath = re.ReplaceAllString(refs.RepoLink, "")
```
alvaroaleman, line 115: Do I understand correctly that the purpose of the regex is to strip the `http://` prefix? If so, `strings.TrimPrefix` would be more appropriate
/ok-to-test

taharah: Yes, that is the intent; however, I used regex instead of `strings.TrimPrefix` in order to remove either `http://` or `https://` since more than likely there are still enterprise deployments without TLS enabled.

alvaroaleman: Sure, so use two `string.Replace`s or just split by `://` and drop the first part? This problem seems a bit too trivial to use a regex

taharah:
@alvaroaleman I've addressed your comment. Additionally, just to provide some anecdotal evidence, I deployed my fork internally to our Prow cluster that uses GitHub Enterprise and the clonerefs utility worked exactly as expected.

Thank you for reviewing and happy holidays!

alvaroaleman (approved):

prow/pod-utils/clone/clone.go
```diff
@@ -109,6 +109,10 @@ func PathForRefs(baseDir string, refs prowapi.Refs) string {
 	var clonePath string
 	if refs.PathAlias != "" {
 		clonePath = refs.PathAlias
+	} else if refs.RepoLink != "" {
```
fejta, line 112: In the future can we get these sort of changes behind flags? This broke gerrit integration (checking the repo out this way makes more sense, but is a backwards incompatible change that we need to be aware of)

alvaroaleman: The issue is that very few ppl know what an incompatible change from gerrits POV is and those tend to not actively review changes. From githubs POV this is a backwards-compatible change.

fejta: Change from "either use path_alias or else github.com/$ORG/$REPO" to now make repo_link take precedence is not a backwards compatible change.

alvaroaleman: For github it is, because `RepoLink` without the protcol prefix and delimiter is the same as what is in the else block except for when it is a private GHE intstance, then it's the url to that instance and using that is more correct than always using github.com

fejta: https://github.com/kubernetes/test-infra/blob/2abb7ff26579325f6a335990003f665755c96d7a/prow/gerrit/adapter/adapter.go#L154

fejta: "I'm going to make this depend on RepoLink without a flag and also refuse to check whether anything depends on RepoLink" doesn't seem like a cooperative attitude

alvaroaleman: This is not about attitude, this is about not being aware of how the gerrit integration uses things or what it even uses. I didn't even know it uses the decoration utils, so I didn't think about grepping anything, I just validated that its compatible from Githubs POV (which to all my knowledge it is).

Can you create a list of functionality that gerrit uses? I genuinely have no idea about that. I absolutely didn't expect a struct that is very clearly designed to hold data from Github to be used for something else.

fejta: It uses very little in hook (aka plugins, etc) and basically all features of everything else.

fejta: Is leaving a github comment unique to github? Yes, gerrit is unlikely to touch a comment call.

Is checking out a git repo unique to github? Not unique to github, something gerrit will likely depend on.

taharah: @fejta I have to admit that I was also looking at this change solely from the perspective of Github and was not aware of the implications on the integrations of other git providers. Do you have an open issue tracking the impact of this change on Gerrit? I would like to help implement any fix(es) or solution that works for all providers.

fejta: Thanks, and there's no issue to track since we just updated our paths to correct it.

This was a great contribution (thanks for making it!) and the correct solution. Nothing we need to change about the solution. We just needed to update our jobs to use the new path (which we've done).

Also I wouldn't expect you to be aware of these gotchas. Most people just use github, and core maintainers have been struggling to avoid these edge cases.

Ideally I'd like us to either:
* Be a little bit more diligent about checking the impacts of these changes
  - The new behavior changes RepoLink and ClonePath, which will gerrit depending on it
* Recognizing that this is often hard to, perhaps gating this behavior change behind a flag.
  - https://github.com/kubernetes/test-infra/pull/20339 is a good example of something which adds new functionality without immediately foisting it on everyone

fejta: PS: the problem here is really subtle, so not super surprising to me that we missed it (both on impl, review, our deployment):
* It changes the checked out location for gerrit repos from `/root/github.com/somewhere.googlesource.com/bar` to `/root/somewhere.googlesource.com/bar`
  - The new path is obviously the more correct one
  - Not triggered if the ref specifies path_alias.


So if the job has problems if and only if it:
1) Uses gerrit
2) has gerrit refs that don't specify path_alias
3) Runs code that requires source code path to be /root/github.com/gerrit rather than /root/gerrit

As you might imagine most of our jobs fail at least one of these criteria -- they either aren't on gerrit, use path_alias and/or do not care so precisely where the code is located. Therefore it wasn't immediately obvious the new release was bad since most jobs were working.

PR #20251, opened 2020-12-16
Create a test-infra presubmit test job for interation testing prow

chaodaiG:
This job is set to not always run, optional for tide, as well as skipping report on github, will change these once the integration test is verified to work

### DIFF
--- a/config/jobs/kubernetes/test-infra/test-infra-presubmits.yaml
+++ b/config/jobs/kubernetes/test-infra/test-infra-presubmits.yaml
@@ -97,3 +97,23 @@ presubmits:
         - --warnings=duplicate-job-refs
     annotations:
       testgrid-dashboards: presubmits-test-infra
+  - name: pull-test-infra-integration
+    branches:
+    - master
+    always_run: false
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes/test-infra/test-infra-presubmits.yaml
```diff
@@ -97,3 +97,23 @@ presubmits:
         - --warnings=duplicate-job-refs
     annotations:
       testgrid-dashboards: presubmits-test-infra
+  - name: pull-test-infra-integration
+    branches:
+    - master
+    always_run: false
```
chases2, line 103: If always_run: false and there's no "run when this file is changed" regex, does that mean it'll never run unless you `/test` for it?

chaodaiG: Yeah that's intended, this job will be used for testing my initial PR of seeding integration tests and before that no other PR should ever trigger this job since they are bound to fail

alvaroaleman (approved):

PR #20256, opened 2020-12-16
Prow alert: make stale prow image alert configurable and set it to fire only when the staleness discovered a whole day

chaodaiG:
This alert fired this morning due to a temporary prow-controller-manager outage for 6 minutes, de-sensitise it since it's not meant to be that urgent

### DIFF
--- a/config/prow/cluster/monitoring/mixins/lib/config.libsonnet
+++ b/config/prow/cluster/monitoring/mixins/lib/config.libsonnet
@@ -67,7 +67,7 @@ local config = {
   webhookMissingAlertInterval: '10m',
 
   // How many days prow hasn't been bumped.
-  prowImageStaleByDays: 7,
+  prowImageStaleByDays: {daysStale: 7, hoursToTrigger: '24h'},
 };
 
 // Generate the real config by adding in constant fields and defaulting where needed.
--- a/config/prow/cluster/monitoring/mixins/lib/config.libsonnet
+++ b/config/prow/cluster/monitoring/mixins/lib/config.libsonnet
@@ -67,7 +67,7 @@ local config = {
   webhookMissingAlertInterval: '10m',
 
   // How many days prow hasn't been bumped.
-  prowImageStaleByDays: {daysStale: 7, hoursToTrigger: '24h'},
+  prowImageStaleByDays: {daysStale: 7, eventDuration: '24h'},
 };
 
 // Generate the real config by adding in constant fields and defaulting where needed.
--- a/config/prow/cluster/monitoring/mixins/lib/config_util.libsonnet
+++ b/config/prow/cluster/monitoring/mixins/lib/config_util.libsonnet
@@ -39,6 +39,6 @@
           ),
         },
     }
-    + default(config, 'prowImageStaleByDays', 7)
+    + default(config, 'prowImageStaleByDays', {daysStale: 7, hoursToTrigger: '24h'})
   ),
 }
--- a/config/prow/cluster/monitoring/mixins/lib/config_util.libsonnet
+++ b/config/prow/cluster/monitoring/mixins/lib/config_util.libsonnet
@@ -38,7 +38,8 @@
               '%s (Requires <https://github.com/kubernetes/test-infra/tree/master/config/prow/cluster/monitoring#access-components-web-page|port-forwarding the grafana service> and accessing path "%s")' % [text, path]
           ),
         },
+      prowImageStaleByDays+: default(config.prowImageStaleByDays, 'daysStale', 7)
+        + default(config.prowImageStaleByDays, 'hoursToTrigger', '24h'),
     }
-    + default(config, 'prowImageStaleByDays', {daysStale: 7, hoursToTrigger: '24h'})
   ),
 }
--- a/config/prow/cluster/monitoring/mixins/lib/config_util.libsonnet
+++ b/config/prow/cluster/monitoring/mixins/lib/config_util.libsonnet
@@ -39,7 +39,7 @@
           ),
         },
       prowImageStaleByDays+: default(config.prowImageStaleByDays, 'daysStale', 7)
-        + default(config.prowImageStaleByDays, 'hoursToTrigger', '24h'),
+        + default(config.prowImageStaleByDays, 'eventDuration', '24h'),
     }
   ),
 }
--- a/config/prow/cluster/monitoring/mixins/prometheus/stale_alerts.libsonnet
+++ b/config/prow/cluster/monitoring/mixins/prometheus/stale_alerts.libsonnet
@@ -8,13 +8,13 @@
             alert: 'Prow images are stale',
             expr: |||
               time()-max(prow_version) > %d * 24 * 3600
-            ||| % $._config.prowImageStaleByDays,
-            'for': '5m',
+            ||| % $._config.prowImageStaleByDays.daysStale,
+            'for': '%s' % $._config.prowImageStaleByDays.hoursToTrigger,
             labels: {
               severity: 'critical',
             },
             annotations: {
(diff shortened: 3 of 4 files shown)


chaodaiG:
I know where the problem is, will fix shortly

config/prow/cluster/monitoring/mixins/lib/config.libsonnet
```diff
@@ -67,7 +67,7 @@ local config = {
   webhookMissingAlertInterval: '10m',
 
   // How many days prow hasn't been bumped.
-  prowImageStaleByDays: 7,
+  prowImageStaleByDays: {daysStale: 7, hoursToTrigger: '24h'},
```
cjwagner, line 70: This is a breaking change for every other Prow instance if we just want to be able to update the alert files without requiring config updates. Thats probably ok for now since we haven't automated the update and can manually fix this for every other instance, but please avoid in the future.

nit: `prowImageStaleByDays` is now kind of incorrectly named. Please drop the `Days` suffix now that this is a struct. Similarly `hoursToTrigger` is a bit of a misnomer since the value contains units.

nit: The `hoursToTrigger` field would be a good field to add defaulting logic for. I think `24h` is suitable for most every instance.

config/prow/cluster/monitoring/mixins/prometheus/stale_alerts.libsonnet
```diff
@@ -8,13 +8,13 @@
             alert: 'Prow images are stale',
             expr: |||
               time()-max(prow_version) > %d * 24 * 3600
-            ||| % $._config.prowImageStaleByDays,
-            'for': '5m',
+            ||| % $._config.prowImageStaleByDays.daysStale,
+            'for': '%s' % $._config.prowImageStaleByDays.hoursToTrigger,
```
cjwagner, line 12: Nit: No formatting is needed here, you can use the variable directly.

config/prow/cluster/monitoring/mixins/prometheus/stale_alerts.libsonnet
```diff
@@ -8,13 +8,13 @@
...
             labels: {
               severity: 'critical',
             },
             annotations: {
-              message: '@test-infra-oncall The prow images are older than %d days.' % $._config.prowImageStaleByDays,
+              message: '@test-infra-oncall The prow images are older than %d days.' % $._config.prowImageStaleByDays.daysStale,
```
cjwagner, line 17: Please remove `@test-infra-oncall` that is only appropriate for k8s-prow and only appropriate for reporting sinks that route to slack so it should be included in the receiver config https://github.com/kubernetes/test-infra/blob/a38a2c68c3701f197bbf2386746ab704d4a01e55/config/prow/cluster/monitoring/alertmanager-prow_secret.yaml#L28

config/prow/cluster/monitoring/mixins/lib/config.libsonnet
chaodaiG: I have added defaults for both `daysStale` and `hoursToTrigger` at https://github.com/kubernetes/test-infra/pull/20256/files#diff-f0b2fe3f3f67579243edc6515cac8a84522abbe9633dcb4c62a786c52ac517bdR42, did you mean to default `hoursToTrigger` only?

cjwagner: Defaulting both is great. `hoursToTrigger` just seemed a little more likely to be the same across multiple instances so I called that out.

chaodaiG: ack

config/prow/cluster/monitoring/mixins/prometheus/stale_alerts.libsonnet
chaodaiG: Done

config/prow/cluster/monitoring/mixins/prometheus/stale_alerts.libsonnet
chaodaiG: Done

chaodaiG:

config/prow/cluster/monitoring/mixins/lib/config_util.libsonnet
```diff
@@ -39,6 +39,6 @@
           ),
         },
     }
-    + default(config, 'prowImageStaleByDays', 7)
+    + default(config, 'prowImageStaleByDays', {daysStale: 7, hoursToTrigger: '24h'})
```
cjwagner, line 42: nit: This won't default the fields individually, only the entire struct if it is omitted.

config/prow/cluster/monitoring/mixins/lib/config.libsonnet
cjwagner: `hoursToTrigger` is still a misnomer and inconsistent with the `daysStale` field because it is actually a duration since it includes the `h` unit specifier.

config/prow/cluster/monitoring/mixins/prometheus/stale_alerts.libsonnet
```diff
@@ -8,13 +8,13 @@
...
-            ||| % $._config.prowImageStaleByDays,
-            'for': '5m',
+            ||| % $._config.prowImageStaleByDays.daysStale,
+            'for': $._config.prowImageStaleByDays.hoursToTrigger,
             labels: {
               severity: 'critical',
             },
             annotations: {
-              message: '@test-infra-oncall The prow images are older than %d days.' % $._config.prowImageStaleByDays,
+              message: 'The prow images are older than %d days.' % $._config.prowImageStaleByDays.daysStale,
```
cjwagner, line 17: nit: Could also mention the `hoursToTrigger` in the description. (Otherwise it might seem like this alert is firing a day late.)

config/prow/cluster/monitoring/mixins/lib/config_util.libsonnet
chaodaiG: Make sense, done

config/prow/cluster/monitoring/mixins/prometheus/stale_alerts.libsonnet
chaodaiG: Done

chaodaiG:

config/prow/cluster/monitoring/mixins/lib/config_util.libsonnet
```diff
@@ -38,7 +38,8 @@
               '%s (Requires <https://github.com/kubernetes/test-infra/tree/master/config/prow/cluster/monitoring#access-components-web-page|port-forwarding the grafana service> and accessing path "%s")' % [text, path]
           ),
         },
+      prowImageStaleByDays+: default(config.prowImageStaleByDays, 'daysStale', 7)
+        + default(config.prowImageStaleByDays, 'hoursToTrigger', '24h'),
```
cjwagner, line 42: This defaults the wrong field.

cjwagner (commented):
One previous comment also still needs to be addressed (kind of hidden on a previous review) https://github.com/kubernetes/test-infra/pull/20256#discussion_r556200013

config/prow/cluster/monitoring/mixins/lib/config.libsonnet
chaodaiG: makes sense, updated to `eventDuration`

cjwagner (approved):

config/prow/cluster/monitoring/mixins/lib/config_util.libsonnet
cjwagner: 🤦 No it doesn't.

PR #20257, opened 2020-12-16
kops AWS CCM: run with kubernetes 1.19

justinsb:
The AWS CCM doesn't currently have a build for 1.21; also we likely
want to validate with a stable k8s version first.

### DIFF
--- a/config/jobs/kubernetes/kops/kops-periodics-grid.yaml
+++ b/config/jobs/kubernetes/kops/kops-periodics-grid.yaml
@@ -47145,7 +47145,7 @@ periodics:
     testgrid-dashboards: sig-cluster-lifecycle-kops, google-aws, kops-grid, kops-distro-u2004, kops-k8s-latest
     testgrid-tab-name: kops-grid-scenario-public-jwks
 
-# {"cloud": "aws", "container_runtime": "docker", "distro": "u2004", "extra_flags": "--override=cluster.spec.cloudControllerManager.cloudProvider=aws", "feature_flags": "EnableExternalCloudController,SpecOverrideFlag", "k8s_version": null, "kops_version": null, "networking": null}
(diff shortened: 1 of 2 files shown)


justinsb:
cc @nckturner I think this will address the problems.  The invalid manifest problems have also been fixed, and were because we were returning an empty tag for an unknown k8s version, which the yaml parser was treating as a map (I think!)

config/jobs/kubernetes/kops/kops-periodics-grid.yaml
```diff
@@ -47189,10 +47189,10 @@ periodics:
     test.kops.k8s.io/distro: u2004
     test.kops.k8s.io/extra_flags: --override=cluster.spec.cloudControllerManager.cloudProvider=aws
     test.kops.k8s.io/feature_flags: EnableExternalCloudController,SpecOverrideFlag
-    test.kops.k8s.io/k8s_version: ''
+    test.kops.k8s.io/k8s_version: '1.19'
     test.kops.k8s.io/kops_version: ''
     test.kops.k8s.io/networking: ''
-    testgrid-dashboards: sig-cluster-lifecycle-kops, google-aws, kops-grid, kops-distro-u2004, kops-k8s-latest
+    testgrid-dashboards: sig-cluster-lifecycle-kops, google-aws, kops-grid, kops-distro-u2004, kops-k8s-1.19
```
nckturner, line 47195: Should we also add it to the `sig-aws-cloud-provider-aws` dashboard?

nckturner:
Nice!

config/jobs/kubernetes/kops/kops-periodics-grid.yaml
justinsb: Good idea -- I'll do that in another PR though as it will require some additional wiring up

justinsb: Done in #20260 - thanks for the suggestion!

PR #20258, opened 2020-12-16
kops arm64 scenario: pass zone explicitly using kops-zones flag

justinsb:
Otherwise kubetest injects a different zone, causing a
conflict (usually).


justinsb:
cc @hakman I hope this will fix the arm64 scenario test

PR #20265, opened 2020-12-16
test images: Adds dockerhub images used in tests

claudiubelu:
Adds the prow config jobs for images that are currently hosted on dockerhub but are being used in E2E tests. The image Dockerfiles are mostly noop since the purpose of these jobs is to mirror those images into our own registries, so we can then use them in tests instead of the dockerhub ones.

The PR that adds the images: https://github.com/kubernetes/kubernetes/pull/95567


BenTheElder:
(lgtm pending kubernetes/kubernetes discussion, I'm +1 on these but there's an outstanding comment not in favor, let's give it a moment before proceeding with either)

Issue #20275, opened 2020-12-17
Move logs in scalability logs to a separate bucket


mborsz:
Partially discussed in https://github.com/kubernetes/test-infra/pull/17215#issuecomment-621084436

The goal here is to modify scalability tests to dump logs to a separate bucket to:
a) unblock pod-utils migration (see https://github.com/kubernetes/test-infra/pull/17215 for explanation why it's needed)
b) by moving most of the files out of artifacts directory, making prow faster (see https://github.com/kubernetes/test-infra/issues/17505)

/assign

mborsz:
This is done
/close

PR #20276, opened 2020-12-17
Make it possible to override --logexporter-gcs-path

mborsz:
This PR introduces a way to set --logexporter-gcs-path in kubetest to something else than GCS artifacts dir.

* If --logexporter-gcs-path arg in kubernetes_e2e is set, it will pass this value down to kubetests
* otherwise it will fallback to the old behavior

The goal is to make it possible to move scalability test logs outside gcs

Ref https://github.com/kubernetes/test-infra/issues/20275

### DIFF
--- a/scenarios/kubernetes_e2e.py
+++ b/scenarios/kubernetes_e2e.py
@@ -536,8 +536,7 @@ def main(args):
     runner_args.extend(args.kubetest_args)
 
     if args.use_logexporter:
-        # TODO(fejta): Take the below value through a flag instead of env var.
-        runner_args.append('--logexporter-gcs-path=%s' % os.environ.get('GCS_ARTIFACTS_DIR', ''))
+        runner_args.append('--logexporter-gcs-path=%s' % args.logexporter_gcs_path)
 
     if args.aws:
         # Legacy - prefer passing --deployment=kops, --provider=aws,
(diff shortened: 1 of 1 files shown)


scenarios/kubernetes_e2e.py
```diff
@@ -537,7 +537,7 @@ def main(args):
 
     if args.use_logexporter:
         # TODO(fejta): Take the below value through a flag instead of env var.
-        runner_args.append('--logexporter-gcs-path=%s' % os.environ.get('GCS_ARTIFACTS_DIR', ''))
+        runner_args.append('--logexporter-gcs-path=%s' % os.environ.get('LOG_DUMP_DIR', os.environ.get('GCS_ARTIFACTS_DIR', '')))
```
spiffxp, line 540: I would prefer to see the TODO addressed, and the custom path passed in explicitly as a flag
```python
runner_args.append('--logexporter-gcs-path=%s' % args.logexporter_gcs_path)
# ...
parser.add_argument(
        '--logexporter-gcs-path',
        default=os.environ.get('GCS_ARTIFACTS_DIR',''),
        help='GCS path where logexporter tool will upload logs if enabled')
```

spiffxp (changes requested):
Does it need to be an env var?

scenarios/kubernetes_e2e.py
mborsz: Thanks! Looks much better now!

PTAL

mborsz:
Hi!

Friendly ping! :)

PR #20282, opened 2020-12-17
Customize column header on testgrid for ppc64le conformance jobs

Rajalakshmi-Girish:
Test-Grid : https://k8s-testgrid.appspot.com/conformance-ppc64le#Periodic%20ppc64le%20conformance%20test%20on%20local%20cluster

The current commit-id that is being display in test-grid for ppc64le conformance jobs is from the https://github.com/ppc64le-cloud/kubetest2-plugins. 

It needs to be the version of k8s against which the conformance tests are run.
Hence customizing the column header to point to new key `k8s-build-version `which has been added to finished.json file in logs folder of gcs bucket from where the logs are picked up.
Reference: https://github.com/kubernetes/test-infra/blob/master/testgrid/config.md#column-headers


spiffxp (approved):
/ok-to-test
/approve
/lgtm

PR #20291, opened 2020-12-18
Allow k8s-ci-robot and kube-state-metrics-admins/maintainers to push to gh-pages branch

cpanato:
trying to fix the release issue: https://github.com/kubernetes/kube-state-metrics/pull/1334

/assign @scottrigby @lilic

### DIFF
--- a/config/prow/config.yaml
+++ b/config/prow/config.yaml
@@ -276,6 +276,14 @@ branch-protection:
             - ci-validate-docs
             - ci-validate-go-modules
             - ci-validate-manifests
+          branches:
+            gh-pages:
+              restrictions:
+                users:
+                - k8s-ci-robot
+                teams:
+                - kube-state-metrics-admins
+                - kube-state-metrics-maintainers
         kube-scheduler:
           restrictions:
             users: []


scottrigby (approved):
Thank you!

wonder-twins-power-activate

cpanato:
@scottrigby @dims I think instead use the exclude rule that removes the protect branch, we can add a restriction allowing the bot and the admin/maintainers to be able to push to the branch.


Dims, this is to release the Helm chart, and there is a Github action that will run every time we merge any helm chart code, and right now it is failing because the bot does not have permission to push the release to the branch.

I think with this we keep the branch protected but allow trusted users/team to push the release

scottrigby:
@cpanato Only one thought: that approach would require updating this file whenever a new maintainer for each project is added (basically, anyone who is able to merge a PR for a chart on a k8s project), because the chart-releaser project works by updating the index file in gh-pages using the GITHUB_TOKEN of whichever user merges the PR.

scottrigby:
@cpanato question: is there ever a time when maintainers need to merge PRs manually rather than waiting for k8s-ci-robot to do it through prow commands?

cblecker:
/approve

The approach as of 4f04de9 looks fine to me.

@cpanato If you change approaches, please re-ping for approval.

cpanato:
@scottrigby the Teams defined they exist and refer to the admins/maintainers of the project/repository this exist for all projects and you can see the definition in `k/org` repo, so if we need to add/remove a new admin/maintainer this is already part of the process today, however, we don't need to edit, only if you want to add a new team or the team change the name.

We should not add individual users here, except the bot that will merge the PR.


This way we keep the branch protected, the previous approach we remove the branch from the protected which people can delete de branch by mistake (admins/maintainers)

I think if this works as expected, will be the recommended approach because keeps the branch protected. If does not work (😄 ) then we can try to remove the branch from being protected.

thoughts? makes sense? or am I too confusing :)


ps. sorry for the delay to reply

scottrigby:
@cpanato  yes that makes sense, let's try it 🙂

PR #20295, opened 2020-12-19
Always run the containerd pre-submit job on master

hakman:
As discussed during the 18.12.2020 office hours, the containerd pre-submit job should always run and eventually replace the Docker job.

/cc @olemarkus @rifelpet


matthyx:
@hakman do you need a lgtm here?

hakman:
Thanks @matthyx. That would be nice. 😄

PR #20303, opened 2020-12-21
Set days_of_results to 7 ppc64le-conformance and ppc64le-conformance-containerd jobs

Rajalakshmi-Girish:
The ppc64le-test-grid is showing internal error. 
Currently, it is throwing the following message:

```
Internal Error: Refresh, and file an issue if problem persists. (Your table may be too big; consider lower `days_of_results` or `enable_test_methods = false`).
```
Hence adding days_of_results: 7 to the ppc64le-conformance jobs.


spiffxp (approved):
/approve
/lgtm

PR #20304, opened 2020-12-21
Add 'LOGDUMP_TO_GCS_ONLY' which can be used to upload all logs to the same dir

mborsz:
Without this change, master logs and nodes that have been fetched using ssh fallback are fetched locally, but logexporter pushes to '${gcs_artifacts_dir}'. This makes it impossible to set gcs_artifacts_dir to something else than artifacts as this will lead to splitting logs to two directories depending on the way they have been fetched.

With 'LOGDUMP_TO_GCS_ONLY', all logs are fetched to the same directory.

Ref https://github.com/kubernetes/test-infra/issues/20275

/assign @wojtek-t

### DIFF
--- a/logexporter/cluster/log-dump.sh
+++ b/logexporter/cluster/log-dump.sh
@@ -21,7 +21,8 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
-readonly report_dir="${1:-_artifacts}"
+readonly local_report_dir="${1:-_artifacts}"
+report_dir=""
 readonly gcs_artifacts_dir="${2:-}"
 readonly logexporter_namespace="${3:-logexporter}"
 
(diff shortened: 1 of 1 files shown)


mborsz:
pull-test-infra-bazel failure is flake caused by failure to fetch some remote bazel repository.
/retest

logexporter/cluster/log-dump.sh
```diff
@@ -853,6 +852,47 @@ function main() {
...
+  fi
+
+  dump_logs
+
+  if [[ -n "${LOGDUMP_TO_GCS_ONLY:-}" ]]; then
+    echo "Uploading '${report_dir}' to '${gcs_artifacts_dir}'"
+
+    # Normally, result of gsutil uploading data will be different when target directory exists
+    # and when it's missing. `gsutil cp -r /tmp/dir/logs gs://target-bucket/X/1234/`:
+    # 1. will create `gs://target-bucket/X/1234/logs` if '1234' existed.
```
wojtek-t, line 881: just for my own education: it will just create the directory and not upload it?

I think I don't fully understand the difference between 1 and 2 here.

wojtek-t:
Thanks

/lgtm
/approve

PR #20312, opened 2020-12-21
Hook: Allow to enable/disable on org/repo

alvaroaleman:
This change allows to enable or disable hook on org or repo level. The
intended use case for this is a migration of Prow from personal access
tokens to a Github App.


alvaroaleman (commented):
/hold cancel

Issue #20307, opened 2020-12-21
[Kettle] Streaming BQ results issuing "illegal requests"


MushuEE:
**What happened**:
Kettle was updated to surpass the production lockup it was hitting due to Database growth and it has been some time since it was successful in executing a stream PubSub process. It appears this process is now failing due to `Your client has issued a malformed or illegal request.  <ins>That\u2019s all we know.</ins>`

```
Traceback (most recent call last):
  File "stream.py", line 310, in <module>
    stop=StopWhen(OPTIONS.stop_at))
  File "stream.py", line 204, in main
    emitted = insert_data(bq_client, table, make_json.make_rows(db, builds))
  File "stream.py", line 124, in insert_data
    errors = retry(bq_client.insert_rows, table, rows, skip_invalid_rows=True)
  File "stream.py", line 95, in retry
    return func(*args, **kwargs)
  File "/usr/local/lib/python3.6/dist-packages/google/cloud/bigquery/client.py", line 2869, in insert_rows
    return self.insert_rows_json(table, json_rows, **kwargs)
  File "/usr/local/lib/python3.6/dist-packages/google/cloud/bigquery/client.py", line 3016, in insert_rows_json
    timeout=timeout,
  File "/usr/local/lib/python3.6/dist-packages/google/cloud/bigquery/client.py", line 636, in _call_api
    return call()
  File "/usr/local/lib/python3.6/dist-packages/google/api_core/retry.py", line 286, in retry_wrapped_func
    on_error=on_error,
  File "/usr/local/lib/python3.6/dist-packages/google/api_core/retry.py", line 184, in retry_target
    return target()
  File "/usr/local/lib/python3.6/dist-packages/google/cloud/_http.py", line 438, in api_request
    raise exceptions.from_http_response(response)
google.api_core.exceptions.BadRequest: 400 POST https://bigquery.googleapis.com/bigquery/v2/projects/k8s-gubernator/datasets/build/tables/all/insertAll?prettyPrint=false: <!DOCTYPE html>
<html lang=en>
  <meta charset=utf-8>
  <meta name=viewport content="initial-scale=1, minimum-scale=1, width=device-width">
  <title>Error 400 (Bad Request)!!1</title>
  <style>
[...]

PR #20314, opened 2020-12-22
[Kettle] Sync Stream with 30 day change and chunk stream

MushuEE:
*I forgot to sync the stream with the load change on the 30 day all table

Also if you send very large row requests we break the request. Added chunking to avoid this issue.

/cc @amwat 
#20307
/area kettle

### DIFF
--- a/kettle/stream.py
+++ b/kettle/stream.py
@@ -40,7 +40,7 @@ import model
 import make_db
 import make_json
 
-MAX_ROW_UPLOAD=25
+MAX_ROW_UPLOAD = 25
 
 def process_changes(results):
     """Split GCS change events into trivial ack_ids and builds to further process."""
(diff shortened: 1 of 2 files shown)


kettle/stream.py
```diff
@@ -112,22 +113,29 @@ def insert_data(bq_client, table, rows_iter):
         rows_iter: row_id, dict representing a make_json.Build
     Returns the row_ids that were inserted.
     """
+    def divide_chunks(l, n=MAX_ROW_UPLOAD):
```
amwat, line 116: probably use a more descriptive argument

kettle/stream.py
```diff
@@ -40,6 +40,7 @@
 import make_db
 import make_json
 
+MAX_ROW_UPLOAD=25
```
amwat, line 43: nit: spacing lint

amwat (commented):
/approve

PR #20315, opened 2020-12-22
[Kettle] Make day vars Global to help enforce sync between make_json and stream 

MushuEE:
If these change in one place and not the other the emit tables will be out of sync and can cause dupe data
/area kettle


kettle/update.py
```diff
@@ -22,6 +22,9 @@
 THREADS = 32
 MAX_BAD_RECORDS = 1000
 DAYS_OLD = 1.9
+DAY = 1
+WEEK = 2
```
amwat, line 26: 7 ?

MushuEE: no idea how I did that lol must have been late

MushuEE, line 26:
```suggestion
WEEK = 7
```

PR #20317, opened 2020-12-22
distill the networkpolicy job targets

jayunit100:
Adding `Slow` and `L7` skips to the networkpolicy suite so that we can increase the S/N ratio and increase its relevance to on-prem testing scenarios.


jayunit100:
im going to close this, ricardo has taken over the porting of this....

PR #20325, opened 2020-12-22
skip the l7 test for networkpolicies

jayunit100:
Skip the L7 test for networkpolicies


jayunit100:
@aojea have a PR in flight to add this tag 

https://github.com/kubernetes/kubernetes/pull/91592 will be green then

jayunit100:
Ultimately we'll want to consolidate all L7 tags under the `L7` label, and all L3 tags under the `L3` label, and then simplify these tags.

jayunit100:
xref, eventually https://github.com/kubernetes/kubernetes/issues/97472 will obviate all these myriad tags if we can accomplish unifying the l3 and l7 on all loadbalancer tests

jayunit100:
will let ricardo/knabben handle these skip flags

PR #20326, opened 2020-12-22
Update cloud-provider-aws-config.yaml

ayberk:
- Skip presubmits for gh-pages branch. 
- Don't protect gh-pages so github actions can push to it.

/cc @nckturner


nckturner:
@ayberk can you also add something akin to https://github.com/kubernetes/test-infra/pull/20323 which will un-protect the gh-pages branch?  This should fix the mkdocs job: https://github.com/kubernetes/cloud-provider-aws/runs/1598685921

spiffxp (approved):
/approve
/lgtm

Issue #20324, opened 2020-12-22
Migrate sig-network jobs to podutils


rikatz:
**What should be cleaned up or changed**: Change sig-network jobs to use podutils

**Provide any links for context**: https://gist.github.com/dims/c1296f8ed42238baea0a5fcae45f4cf4#file-notes-md

rikatz:
@aojea wanna close this one and wait like...3 months and we can start migrating again?

aojea:
since you already migrated one, and podutils fixed the problems with the graphs, I think we can go ahead ... if there are volunteers to do it, I have low bandwidth these days :)

rikatz:
me and @jayunit100 will volunteer some folks ;x

@mattfenwick @knabben let's put this in our backlog. But I can deal with that also tomorrow :)

knabben:
I can help in the effort to migrate them.. maybe split the workload
I'm curious about this migration procedure.

rikatz:
yes, let's talk about that on sunday triage party, sounds good?

spiffxp:
/remove-lifecycle rotten
/reopen
Would be cool to see https://github.com/kubernetes/test-infra/pull/20864 picked back up and use this as a pilot to write a guide on migrating from bootstrap to pod-uytils

aojea:
> I can help in the effort to migrate them.. maybe split the workload
> I'm curious about this migration procedure.

@rikatz @knabben things changed since this comment, right ?

/help wanted

PR #20308, opened 2020-12-21
add job to test network policies

knabben:
We don't have any job testing network policies, the goal of this PR is to have signal on network policy before merging the new testsuite, so we are confident with the new code.
It leverage current k/k scripts to install network policies in GCE

continuing the rebase from @aojea work on #19817

PR #20336, opened 2020-12-23
Migrate Network Policy testjob to podutils

rikatz:

Started migrating sig-net jobs to podutils, for now let's see if the network policy job works fine :)

Part of: #20324

### DIFF
--- a/config/jobs/kubernetes/sig-network/sig-network-misc.yaml
+++ b/config/jobs/kubernetes/sig-network/sig-network-misc.yaml
@@ -6,6 +6,8 @@ presubmits:
     always_run: false
     run_if_changed: '^(test/e2e/network/|pkg/apis/networking)'
     optional: true
+    decorate: true
+    path_alias: k8s.io/kubernetes
     skip_report: false
     labels:
       preset-service-account: "true"
@@ -16,7 +18,10 @@ presubmits:
       preset-pull-kubernetes-e2e-gce: "true"
     spec:
       containers:
-      - args:
+      - command:
+        - runner.sh
+        - /workspace/scenarios/kubernetes_e2e.py
+        args:
         - --timeout=170
         - --root=/go/src
         - --repo=k8s.io/kubernetes=$(PULL_REFS)


jayunit100:
Looks ok to me ! Should we merge and iterate on this ? I don’t see any logic changes so as long as it’s plugged in to the new stuff properly we’re good !!!!!

rikatz:
The biggest change here is that we don't rely on bootstrap.py anymore to create the resources, pull artifacts, etc etc.

It's not a huge change, but sig-testing has deprecated bootstrap.py and asked to migrate some jobs, so I guess this is a good start :)

config/jobs/kubernetes/sig-network/sig-network-misc.yaml
```diff
@@ -6,6 +6,13 @@ presubmits:
     always_run: false
     run_if_changed: '^(test/e2e/network/|pkg/apis/networking)'
     optional: true
+    decorate: true
+    path_alias: k8s.io/kubernetes
+    extra_refs:
+    - org: kubernetes
+      repo: test-infra
+      base_ref: master
+      path_alias: k8s.io/test-infra
```
aojea, line 15: why do you need to pull k8s.io/test-infra?

rikatz: Because of `/workspace/scenarios/kubernetes_e2e.py` which is in test-infra repo :)

rikatz: Wait, I'm trying to remember why actually we used this, and maybe wasn't because of the scenarios, but because of https://github.com/kubernetes/test-infra/commit/67e14ce72e5f63600e02b40b9780fbf7e49c7ccd#diff-9495ba672589316a9265fd090ecb9c3c74fd414a11386acafda27df4bd409409

As we do not use anything at test-infra I'll remove it

rikatz: Also, I'm wondering if --stage=gs://kubernetes-release-pull/ci/pull-kubernetes-e2e-ubuntu-gce-network-policies is necessary or if this is also covered by podutils, BUT I think for now will leave this way

rikatz:
Removed the extra-args, left the stage=gs:// because I guess podutils only deals with the artifacts pushing, not the staging :)

aojea:
let´s tackle this next week, the jobs look stable but we should let them soak

rikatz:
@aojea let me know when I can rebase this so we can test if with podutils is better :)

Seems that it's already done with https://github.com/kubernetes/test-infra/pull/20308 so just need now to see if the behavior keeps the same :)

spiffxp:
#20498 (comment)

aojea:
/lgtm
:crossed_fingers: 
Thanks Ricardo

PR #20339, opened 2020-12-23
Tide: Allow constructing per-org queries

alvaroaleman:
This change allows to get per-org query strings from a tide query. This
is the first step towards adding github apps support for Tide and was
extracted from the full changeset to make reviewing easier.

The old code is left, because we want only want to do this when github
apps auth is used, as it might result in higher token usage.

FTR, the full changeset for tide github apps support that appears to generally be working but needs some prettying and tests is here: https://github.com/kubernetes/test-infra/compare/master...alvaroaleman:tide-github-apps?expand=1

### DIFF
--- a/prow/config/tide.go
+++ b/prow/config/tide.go
@@ -213,38 +213,80 @@ type TideQuery struct {
 	ReviewApprovedRequired bool `json:"reviewApprovedRequired,omitempty"`
 }
 
-// Query returns the corresponding github search string for the tide query.
-func (tq *TideQuery) Query() string {
-	toks := []string{"is:pr", "state:open", "archived:false"}
+// constructQuery returns a map[org][]orgSpecificQueryParts (org, repo, -repo), remainingQueryString
+func (tq *TideQuery) constructQuery() (map[string][]string, string) {
+	// map org->repo directives (if any)
+	orgScopedIdentifiers := map[string][]string{}
 	for _, o := range tq.Orgs {
-		toks = append(toks, fmt.Sprintf("org:\"%s\"", o))
+		if _, ok := orgScopedIdentifiers[o]; !ok {
+			orgScopedIdentifiers[o] = []string{fmt.Sprintf(`org:"%s"`, o)}
+		}
 	}
 	for _, r := range tq.Repos {
-		toks = append(toks, fmt.Sprintf("repo:\"%s\"", r))
+		if org, _, ok := splitOrgRepoString(r); ok {
+			orgScopedIdentifiers[org] = append(orgScopedIdentifiers[org], fmt.Sprintf("repo:\"%s\"", r))
+		}
 	}
+
 	for _, r := range tq.ExcludedRepos {
-		toks = append(toks, fmt.Sprintf("-repo:\"%s\"", r))
+		if org, _, ok := splitOrgRepoString(r); ok {
+			orgScopedIdentifiers[org] = append(orgScopedIdentifiers[org], fmt.Sprintf("-repo:\"%s\"", r))
+		}
 	}
+
+	queryString := []string{"is:pr", "state:open", "archived:false"}
 	if tq.Author != "" {
-		toks = append(toks, fmt.Sprintf("author:\"%s\"", tq.Author))
+		queryString = append(queryString, fmt.Sprintf("author:\"%s\"", tq.Author))
 	}
 	for _, b := range tq.ExcludedBranches {
-		toks = append(toks, fmt.Sprintf("-base:\"%s\"", b))
+		queryString = append(queryString, fmt.Sprintf("-base:\"%s\"", b))
 	}
 	for _, b := range tq.IncludedBranches {
-		toks = append(toks, fmt.Sprintf("base:\"%s\"", b))
+		queryString = append(queryString, fmt.Sprintf("base:\"%s\"", b))
 	}
 	for _, l := range tq.Labels {
-		toks = append(toks, fmt.Sprintf("label:\"%s\"", l))
+		queryString = append(queryString, fmt.Sprintf("label:\"%s\"", l))
 	}
 	for _, l := range tq.MissingLabels {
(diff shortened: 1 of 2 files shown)


prow/config/tide_test.go
```diff
@@ -33,33 +34,83 @@ import (
...
+}
+
 func TestTideQuery(t *testing.T) {
 	q := " " + testQuery.Query() + " "
-	checkTok := func(tok string) {
-		if !strings.Contains(q, " "+tok+" ") {
-			t.Errorf("Expected query to contain \"%s\", got \"%s\"", tok, q)
-		}
-	}
+	checkTok := checkTok(t, q)
```
alvaroaleman, line 59: I am btw inclined to change the tests to simply test for a literal search string. It has the drawback that ordering will matter so if someone changes the code in a way that the query is the same but with a different order, those tests will fail incorrectly but:
* It makes the tests much easier to understand (Expected query to be X, was Y)
* It also verifies that there are no additional incorrect parts in the search string, this may introduce bugs that will go unnoticed with the current tests (I had a bug around that initially)

alvaroaleman: Does anyone have objections to that?

prow/config/tide.go
```diff
@@ -213,38 +213,83 @@ type TideQuery struct {
...
+	}
+
+	return orgScopedIdentifiers, strings.Join(queryString, " ")
+}
+
+func splitOrgRepoString(orgRepo string) (string, string, bool) {
+	split := strings.Split(orgRepo, "/")
+	if len(split) != 2 {
+		// Just do it like the github search itself and ignore invalid orgRepo identifiers
+		logrus.WithField("OrgRepoStrig", orgRepo).Warn("OrgRepo string is invalid, will be ignored")
```
stevekuznetsov, line 266:
```suggestion
		logrus.WithField("OrgRepoString", orgRepo).Warn("OrgRepo string is invalid, will be ignored")
```

prow/config/tide.go
```diff
@@ -213,38 +213,83 @@ type TideQuery struct {
...
+		return "", "", false
+	}
+	return split[0], split[1], true
+}
+
+// OrgQueries returns the github search string for the query, sharded
```
stevekuznetsov, line 272:
```suggestion
// OrgQueries returns the GitHub search string for the query, sharded
```

(everywhere)

prow/config/tide.go
```diff
@@ -213,38 +213,83 @@ type TideQuery struct {
 	ReviewApprovedRequired bool `json:"reviewApprovedRequired,omitempty"`
 }
 
-// Query returns the corresponding github search string for the tide query.
-func (tq *TideQuery) Query() string {
-	toks := []string{"is:pr", "state:open", "archived:false"}
+func (tq *TideQuery) constructQuery() (map[string][]string, string) {
```
stevekuznetsov, line 216: Can we GoDoc the output of this func now

prow/config/tide.go
```diff
@@ -213,38 +213,83 @@ type TideQuery struct {
...
+	}
+	return split[0], split[1], true
+}
+
+// OrgQueries returns the github search string for the query, sharded
+// by org.
+func (tq *TideQuery) OrgQueries() map[string]string {
+	orgRepoIdentifiers, queryString := tq.constructQuery()
+	result := map[string]string{}
+	for org, repoIdentifiers := range orgRepoIdentifiers {
```
stevekuznetsov, line 277: nit: couldn't this just be

```go
for org, repoIdentifiers := range orgRepoIdentifiers {
    result[org] = queryString + " " + strings.Join(repoIdentifiers, " ")
}
```

alvaroaleman: err yeah

prow/config/tide_test.go
alvaroaleman: okay the static string doesn't work because the result string depends on a map order and that is not deterministic. I have added a check for duplicates though

alvaroaleman:
@stevekuznetsov updated, ptal

fejta:
(FYI tide mostly involves clicking the github merge button and is another thing that does not involve gerrit)

prow/config/tide.go
```diff
@@ -213,38 +213,81 @@ type TideQuery struct {
...
+	}
+
+	return orgScopedIdentifiers, strings.Join(queryString, " ")
+}
+
+func splitOrgRepoString(orgRepo string) (string, string, bool) {
+	split := strings.Split(orgRepo, "/")
+	if len(split) != 2 {
+		// Just do it like the github search itself and ignore invalid orgRepo identifiers
+		logrus.WithField("OrgRepoString", orgRepo).Warn("OrgRepo string is invalid, will be ignored")
```
fejta, line 267: Isn't this going to be excessively spammy if something is misconfigured?

prow/config/tide_test.go
```diff
@@ -33,33 +34,100 @@ import (
...
+	"label:\"lgtm\"",
+	"label:\"approved\"",
+	"-label:\"foo\"",
+	"author:\"batman\"",
+	"milestone:\"milestone\"",
+	"review:approved",
+}
+
 func TestTideQuery(t *testing.T) {
 	q := " " + testQuery.Query() + " "
```
fejta, line 58: It doesn't look like this covers the case where multiple org scopes are put into the same string? That seems important to cover here.

fejta (approved):
/hold

prow/config/tide_test.go
alvaroaleman: It does, because the `testQuery` refers to two orgs: `org` and `k`, the latter via the `Repos`/`ExcludedRepos` attributes

prow/config/tide.go
alvaroaleman: Yeah, removed if. Ignoring something like that is effectively what already happens today, as github just won't find anything in that case

PR #20341, opened 2020-12-24
Improve use of contexts/timeouts in Crier.

cjwagner:
I tried to pick reasonable timeouts, but I'm happy to change them or make them configurable if that would be preferable.
Also includes an update to the OWNERS file.
/assign @fejta @alvaroaleman

### DIFF
--- a/prow/crier/controller.go
+++ b/prow/crier/controller.go
@@ -145,6 +145,12 @@ func (r *reconciler) updateReportStateWithRetries(ctx context.Context, pj *prowv
 func (r *reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
 	log := logrus.WithField("reporter", r.reporter.GetName()).WithField("key", req.String()).WithField("prowjob", req.Name)
 	log.Debug("processing next key")
+	// Limit reconciliation time to 15 minutes. This should more than enough time
+	// for any reasonable reporter. Most reporters should set a stricter timeout
+	// themselves. This mainly helps avoid leaking reconciliation threads that
+	// will never complete.
+	ctx, cancel := context.WithTimeout(ctx, 15*time.Minute)
+	defer cancel()
 	result, err := r.reconcile(ctx, log, req)
 	if err != nil {
 		log.WithError(err).Error("Reconciliation failed")
@@ -157,8 +163,10 @@ func (r *reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 
 func (r *reconciler) reconcile(ctx context.Context, log *logrus.Entry, req reconcile.Request) (*reconcile.Result, error) {
 
+	pjContext, pjCancel := context.WithTimeout(ctx, 30*time.Second)
+	defer pjCancel()
 	var pj prowv1.ProwJob
-	if err := r.pjclientset.Get(ctx, req.NamespacedName, &pj); err != nil {
+	if err := r.pjclientset.Get(pjContext, req.NamespacedName, &pj); err != nil {
 		if errors.IsNotFound(err) {
 			log.Debug("object no longer exist")
 			return nil, nil
@@ -166,6 +174,7 @@ func (r *reconciler) reconcile(ctx context.Context, log *logrus.Entry, req recon
 
 		return nil, fmt.Errorf("failed to get prowjob %s: %w", req.String(), err)
 	}
+	pjCancel()
 
 	if !r.shouldHandle(&pj) {
 		return nil, nil
@@ -199,9 +208,14 @@ func (r *reconciler) reconcile(ctx context.Context, log *logrus.Entry, req recon
 		return requeue, nil
 	}
 
-	log.Info("Reported job(s), now will update pj(s)")
+	log.Infof("Reported job(s), now will update %d pj(s)", len(pjs))
+	// Spend up to 10 mins attempting to update report state since the behavior
+	// is poor if we fail. This long of a delay is needed for when the API
+	// server is temporarily unavailable.
+	updateContext, updateCancel := context.WithTimeout(ctx, 10*time.Minute)
+	defer updateCancel()
 	for _, pjob := range pjs {
-		if err := r.updateReportStateWithRetries(ctx, pjob, log); err != nil {
+		if err := r.updateReportStateWithRetries(updateContext, pjob, log); err != nil {
 			log.WithError(err).Error("Failed to update report state on prowjob")
 			return nil, err
 		}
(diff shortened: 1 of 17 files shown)


alvaroaleman:
@cjwagner our of curiosity, what motivated this change?

cjwagner:
> @cjwagner our of curiosity, what motivated this change?

We are investigating excessive memory usage (20GB+) and goroutines (200k+) in Crier on one of our instances. The goroutines are mostly gRPC threads that we can't identify the source of. Properly using timeout contexts throughout Crier might help terminate dangling threads that are currently using background contexts. We don't have great confidence that this will fix the problem, but it can't hurt and seems more robust either way.
While working on this I also tried to look for memory leaks and I think I identified the real source of our problems. The fix is here: https://github.com/kubernetes/test-infra/pull/20342

prow/crier/controller.go
```diff
@@ -157,23 +163,26 @@ func (r *reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 
 func (r *reconciler) reconcile(ctx context.Context, log *logrus.Entry, req reconcile.Request) (*reconcile.Result, error) {
 
+	pjContext, pjCancel := context.WithTimeout(ctx, 30*time.Second)
+	defer pjCancel()
 	var pj prowv1.ProwJob
-	if err := r.pjclientset.Get(ctx, req.NamespacedName, &pj); err != nil {
+	if err := r.pjclientset.Get(pjContext, req.NamespacedName, &pj); err != nil {
```
alvaroaleman, line 169: This just retrieves it from the cache, I don't think that logic can even respect a context

prow/crier/controller.go
```diff
@@ -199,9 +208,14 @@ func (r *reconciler) reconcile(ctx context.Context, log *logrus.Entry, req recon
...
 	}
 
-	log.Info("Reported job(s), now will update pj(s)")
+	log.Infof("Reported job(s), now will update %d pj(s)", len(pjs))
 	for _, pjob := range pjs {
-		if err := r.updateReportStateWithRetries(ctx, pjob, log); err != nil {
+		// Spend up to 10 mins attempting to update report state since the behavior
+		// is poor if we fail. This long of a delay is needed for when the API
+		// server is temporarily unavailable.
+		updateContext, updateCancel := context.WithTimeout(ctx, 10*time.Minute)
```
alvaroaleman, line 216: Similarly, this is unlikely to do anything because the real cap is the `retry.DefaultBackoff` so I think this will only make a difference if a single UPDATE call takes 10 minutes?

prow/crier/reporters/gcs/reporter.go
```diff
@@ -43,8 +43,8 @@ type gcsReporter struct {
 	author util.Author
 }
 
-func (gr *gcsReporter) Report(log *logrus.Entry, pj *prowv1.ProwJob) ([]*prowv1.ProwJob, *reconcile.Result, error) {
-	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second) // TODO: pass through a global context?
+func (gr *gcsReporter) Report(ctx context.Context, log *logrus.Entry, pj *prowv1.ProwJob) ([]*prowv1.ProwJob, *reconcile.Result, error) {
+	ctx, cancel := context.WithTimeout(ctx, 20*time.Second) // TODO: pass through a global context?
```
alvaroaleman, line 47: I think we can remove the TODO now

prow/crier/controller.go
```diff
@@ -199,9 +208,14 @@ func (r *reconciler) reconcile(ctx context.Context, log *logrus.Entry, req recon
...
+		defer updateCancel()
```
fejta, line 217: Queuing up 10,000 `defer updateCancel()` call to run when we return seems a little excessive. Consider just a single one of these outside the for loop.

prow/crier/controller.go
fejta: Not clear to me whether `updateReportStateWithRetries()` spawns off a go-routine that can take however long it wants.

If it does then it seems like just having a single `ctx, cancel := context.WithTimeout(parent, 10 * minute) // or hour or whatever` seems sufficient to me.

AKA what we really want to avoid is a situation where a single reconcile loop takes 7 hours because of some oddity -- in that situation we can just error out and fix it up on the next reconcile loop.

prow/crier/controller.go
cjwagner: The controller runtime client accepts the context so I assumed it was respected. I think it makes sense to supply something besides the background context so that we don't have to make assumptions about the implementation of the client, but I'm happy to omit the stricter 30 second timeout or switch back to the background context if you don't think this is reasonable.

prow/crier/controller.go
cjwagner: > so I think this will only make a difference if a single UPDATE call takes 10 minutes?

I was thinking this is what would happen if the API server was unavailable due to a master upgrade or reconfiguration, but maybe the requests would just time out?

> Not clear to me whether `updateReportStateWithRetries()` spawns off a go-routine that can take however long it wants.

I don't see us doing that in our code, but I'm not sure if the controller runtime client spawns go-routines to GET or PATCH. That seems possible, but a bit unlikely.
I agree that a single timeout context for updating the report state across all PJs seems sufficient here.

prow/crier/controller.go
alvaroaleman: Yeah, I mostly would like to avoid the additional subcontext here to keep things simple, the global timeout is a good thing to have and there is no reason not to use that context here

prow/crier/controller.go
```diff
@@ -199,9 +208,14 @@ func (r *reconciler) reconcile(ctx context.Context, log *logrus.Entry, req recon
 		return requeue, nil
 	}
 
-	log.Info("Reported job(s), now will update pj(s)")
+	log.Infof("Reported job(s), now will update %d pj(s)", len(pjs))
```
alvaroaleman, line 211: Please use a log field rather than a `printf`-style message so ppl can filter for this log line

prow/crier/controller.go
fejta: +1 to there being a single `reconcile()` level context with a guaranteed reasonable/conservative timeout -- 30m? 1hr?

This obviates the need for `updateContext, pjContext` and should also be used for `ShouldReport(), Report()` etc.

cjwagner: 👍 Switched to a single 30m context at the `reconcile()` level.

prow/crier/controller.go
cjwagner: Done.

cjwagner:
Comments addressed, thanks for the review.

fejta (approved):
/hold

PR #20342, opened 2020-12-24
Fix memory leaks in pubsub reporter.

cjwagner:
I think this is the root cause of the memory leak we've observed in Crier. I looked through the controller and all the reporters, this is the only memory leak I can spot, but I think it explains why we've only had these leaks grow out of control on an instance where pubsub is more heavily used.

/assign @fejta @chaodaiG 
/kind bug

### DIFF
--- a/prow/crier/reporters/pubsub/reporter.go
+++ b/prow/crier/reporters/pubsub/reporter.go
@@ -98,12 +98,18 @@ func (c *Client) Report(_ *logrus.Entry, pj *prowapi.ProwJob) ([]*prowapi.ProwJo
 	message := c.generateMessageFromPJ(pj)
 
 	ctx := context.Background()
+	// TODO: Consider caching the pubsub client.
 	client, err := pubsub.NewClient(ctx, message.Project)
 
 	if err != nil {
 		return nil, nil, fmt.Errorf("could not create pubsub Client: %v", err)
 	}
+	defer func() {
+		logrus.WithError(client.Close()).Debug("Closed pubsub client.")
+	}()
+
 	topic := client.Topic(message.Topic)
+	defer topic.Stop() // Sends remaining messages then stops goroutines.
 
 	d, err := json.Marshal(message)
 	if err != nil {


prow/crier/reporters/pubsub/reporter.go
```diff
@@ -98,12 +98,16 @@ func (c *Client) Report(_ *logrus.Entry, pj *prowapi.ProwJob) ([]*prowapi.ProwJo
 	message := c.generateMessageFromPJ(pj)
 
 	ctx := context.Background()
+	// TODO: Consider caching the pubsub client.
 	client, err := pubsub.NewClient(ctx, message.Project)
```
alvaroaleman, line 102: The client also has a `Close()`, mostly to shutdown its connection pool apparently, maybe `defer` that as well?

cjwagner: Yes, I'm already (sneakily) doing that on line 107 within the log statement. 😄

alvaroaleman: Ah, missed that. I am pretty sure this doesn't do what you think though, the log statement itself does get deferred, but the resolving of its argument does not, hence you will close the client in line 107 and not via the `defer`

prow/crier/reporters/pubsub/reporter.go
```diff
@@ -98,12 +98,16 @@ func (c *Client) Report(_ *logrus.Entry, pj *prowapi.ProwJob) ([]*prowapi.ProwJo
...
 
 	if err != nil {
 		return nil, nil, fmt.Errorf("could not create pubsub Client: %v", err)
 	}
+	defer logrus.WithError(client.Close()).Debug("Closed pubsub client.")
```
fejta, line 107:
```golang
defer func() {
  logrus.WithError().Debug()
}()
```

should address Alvaro's point

prow/crier/reporters/pubsub/reporter.go
```diff
@@ -98,12 +98,16 @@ func (c *Client) Report(_ *logrus.Entry, pj *prowapi.ProwJob) ([]*prowapi.ProwJo
 	message := c.generateMessageFromPJ(pj)
 
 	ctx := context.Background()
+	// TODO: Consider caching the pubsub client.
```
fejta, line 101: This definitely seems like an excellent idea for a subsequent PR. Current design is incredibly wasteful.

prow/crier/reporters/pubsub/reporter.go
cjwagner: Good catch, I thought the arguments were evaluated lazily. Fixed.

cjwagner:
PTAL
/cc @chases2

chases2 (approved):

PR #20343, opened 2020-12-24
Fix an extremely rare race in the GitHub reporter.

cjwagner:
Previously there was a possible race between the cleanup thread and two threads (**A** and **B**) attempting to lock the same PR.
1. **A**: Calls `getLock()` for PR X and is returned a reference to the lock, but has not yet acquired it with `lock.Lock()`.
1. **Cleanup**: Locks the map.
1. **Cleanup**: Acquires PR X's lock, then removes the lock from the map and releases it.
1. **Cleanup**: Finishes cleaning and release the map lock.
1. **A**: Calls `lock.Lock()` on the PR X lock that is no longer in the map.
1. **B**: Calls `getLock()` for PR X and is returned a reference to the **new** PR X lock. **B** acquires the lock and now may conflict with **A**.

Returning the lock from `getLock()` already acquired ensures that the cleanup thread cannot remove the lock from the map before it is used. In other words step 3 would not be possible since step 5 would have already happened as part of step 1 and PR X's lock would already be held.

We'd probably never hit this in a million years, but it is technically possible 🙃 
/kind bug
/assign @fejta @chaodaiG @alvaroaleman

### DIFF
--- a/prow/crier/reporters/github/reporter.go
+++ b/prow/crier/reporters/github/reporter.go
@@ -53,16 +53,32 @@ type simplePull struct {
 }
 
 type shardedLock struct {
-	mapLock *sync.Mutex
+	mapLock *sync.RWMutex
 	locks   map[simplePull]*sync.Mutex
 }
 
 func (s *shardedLock) getLock(key simplePull) *sync.Mutex {
-	s.mapLock.Lock()
-	defer s.mapLock.Unlock()
-	if _, exists := s.locks[key]; !exists {
-		s.locks[key] = &sync.Mutex{}
+	s.mapLock.RLock()
+	defer s.mapLock.RUnlock()
+
+	_, exists := s.locks[key]
+	for !exists {
+		// The key is not in the map, we need to switch to the write lock, add the
+		// new key (if still needed) and then switch back to the read lock.
+		// The key might have been deleted while we switched back to the read lock
+		// though so we do this in a loop until we hold the read lock and the key
+		// exists.
+		s.mapLock.RUnlock()
+		s.mapLock.Lock()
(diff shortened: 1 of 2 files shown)


prow/crier/reporters/github/reporter.go
```diff
@@ -63,6 +63,7 @@ func (s *shardedLock) getLock(key simplePull) *sync.Mutex {
 	if _, exists := s.locks[key]; !exists {
 		s.locks[key] = &sync.Mutex{}
 	}
+	s.locks[key].Lock()
```
alvaroaleman, line 66: I am not really fond of the approach, because it means that we have to hold the map lock (which is at the end of the day a global lock around all presubmit reporting) while waiting for lock, causing a lot of contention.

If we can not think of a way to avoid that, I'd prefer to keep this the way it is because:
* The race you described is generally not very likely to happen
* Even if it ends up happening, the actual race that this is preventing can only happen for failed jobs when they modify the github comment in parallel, discarding each others changes

cjwagner: Thats a good point, I overlooked that. That problem can be significantly reduced by using a read-write mutex so that we only hold a global read lock when waiting for the individual locks. I thought it would eliminate the problem by limiting contention to the once per hour cleanup, but I just remembered there would still be contention when adding a key for a new PR.
Since the `sync.RWMutex` doesn't support atomic upgrade/downgrade, using a RWMutex also required a locking loop which isn't as straight forward as a `Lock` and deferred `Unlock`.

I agree that this race is too unlikely to occur to be worth the reduced performance and any extra complexity. The expected behavior if the race does somehow occur isn't critical either.
Thanks for the review, Alvaro!
/close

PR #20347, opened 2020-12-24
GithubV4: Properly set User-Agent header

alvaroaleman:
This change makes the V4 portion of our github client properly set its
User-Agent header by passing it through the upstream githubql client via
it's context and then setting it via the headerSetting transport.


matthyx:
/lgtm
/hold
Merry Christmas!

alvaroaleman:
Merry Christmas :)

alvaroaleman (commented):
/hold cancel

PR #20352, opened 2020-12-28
Explicitly disable device manager test for windows

jsturtevant:
The device manager test should not run on nodes that are not gpu configured.  

Partially fixes https://github.com/kubernetes/kubernetes/issues/97554


marosset:
/lgtm
/approve

If we have more than one test case we should consider adding a specific identifier in brackets to allow for better discovery/re-use but this is OK for a single test case.

PR #20356, opened 2020-12-29
resource request for network policy job

aojea:
the periodic job that test network policies is not stable, random
test fail but we couldn't observe any pattern.

Some of the test fail because the endpoints are not ready, to discard
first that we have a possible problem with resources constraints, we
allocate memory to the nodes, matching the current presubmit job.

### DIFF
--- a/config/jobs/kubernetes/sig-network/sig-network-misc.yaml
+++ b/config/jobs/kubernetes/sig-network/sig-network-misc.yaml
@@ -634,6 +634,9 @@ periodics:
       - --extract=ci/latest
       - --timeout=50m
       image: gcr.io/k8s-testimages/kubekins-e2e:v20201225-59e70a3-master
+      resources:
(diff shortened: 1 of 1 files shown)


jayunit100:
so the hypothesis here is that 'slowness' of pods spinning down is happening due to memory constraints?  ok.  Seems though like we should have that in the metadata somehow right?

aojea:
> so the hypothesis here is that 'slowness' of pods spinning down is happening due to memory constraints? ok. Seems though like we should have that in the metadata somehow right?

I couldn't find a pattern in current failures or correlate with an specific test, so it seems a good step to try to discard the common first, the last 2 test failures were because the endpoints were not ready on time, it is based on experience, this happened before ...

aojea:
@jayunit100 I had to rebase, we need a new lgtm :)

PR #20357, opened 2020-12-29
skip cloud provider tests for network policies

aojea:
There are several tests that depend on the Cloud Provider, we must
skip those tests when testing general features, because those
should be owned by the specific Cloud Provider.

This is needed because of:

 - the test failure can be related to the Cloud Provider infra, and that
will be not possible to debug for the Kubernetes community.

 - the features are not standard, so we can extrapolate that what works
   in A will work in B.


jayunit100:
+1000 yes, please lets remove as much noise from the netpol signal as possible thanks ! 

/lgtm

Issue #20279, opened 2020-12-17
[Kettle] Staging ENV most likely consuming PubSub from Prod


MushuEE:
**What happened**:
Both Kettle Staging and Prod consume PubSub events from the same *kettle* subscription. They ACK the messages from the PubSub events independently. After a message is consumed it is not resent.

These events are being triggered by GCS and I have a strong suspicion that though the event is probably sent out simultaneously, if staging is consuming that event prior to Prod, Prod will never see it. 

**What you expected to happen**:
Staging and Prod do not consume PubSub events from each-other.

**How to reproduce it (as minimally and precisely as possible)**:
Create two subscribers ACKing on the same subscription.

**Please provide links to example occurrences, if any**:

**Anything else we need to know?**:
This can be solved but creating a duplicate PubSub point for staging. 
/area kettle
/cc @amwat

MushuEE:
/close
This is being handled

PR #20361, opened 2020-12-30
[Kettle] Make staging PubSub that mirrors kettle subscription

MushuEE:
Built a new subscription:
image

This is an exact duplicate but will not consume from Prod allowing Staging to run again #20279

/cc @amwat 
/area kettle

### DIFF
--- a/kettle/update.py
+++ b/kettle/update.py
@@ -74,7 +74,7 @@ def main():
     else:
         call(f'{mj_cmd} | pv | gzip > build_staging.json.gz')
         call(f'{bq_cmd} k8s-gubernator:build.staging build_staging.json.gz schema.json')
-        call('python3 stream.py --poll kubernetes-jenkins/gcs-changes/kettle ' \
+        call('python3 stream.py --poll kubernetes-jenkins/gcs-changes/kettle-staging ' \
             '--dataset k8s-gubernator:build --tables staging:0 --stop_at=1')
 
 if __name__ == '__main__':


kettle/update.py
```diff
@@ -74,7 +74,7 @@ def main():
     else:
         call(f'{mj_cmd} | pv | gzip > build_staging.json.gz')
         call(f'{bq_cmd} k8s-gubernator:build.staging build_staging.json.gz schema.json')
-        call('python3 stream.py --poll kubernetes-jenkins/gcs-changes/kettle ' \
+        call('python3 stream.py --poll kubernetes-jenkins/gcs-changes/kettle-staging ' \
```
amwat, line 77: won't this change also affect prod?

The subscription name should ideally be a parameter

MushuEE: There are two subs now, so it will not. They should not cross contaminate.

MushuEE: But i can make it a param but there should really be one per instance.

amwat: I mean if we make the hardcoded code change, whenever we deploy to prod next it will also pick up the staging subscription instead right ?

MushuEE: Prod has it's own call to stream on line 72, which uses the unchanged subscription. So should be 100% fine. Only staging will use the experimental subscription.

MushuEE: Agreed, will update as params

PR #20362, opened 2020-12-31
[Kettle Experiment] Add Cloud Function to Stream build results to BQ

MushuEE:
This is a WIP and does not need to be reviewed

### DIFF
--- a/kettle/CloudFunctions/README.md
+++ b/kettle/CloudFunctions/README.md
@@ -1,0 +1,25 @@
+# Kettle Streaming Big Query Insert
+
+## Google Cloud Requirements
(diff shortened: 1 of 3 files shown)


MushuEE:
@amwat @spiffxp if we are ok with merging WIPs, I would like to keep PRs ~200 lines or less so reviewing isn't painful. I did deploy and check this works. Next I will add some deployment scripts and start adding in code for generating the row info and BQ inserts.

amwat:
I'm fine merging WIP PRs, not sure if we care about where though.
maybe under `experimental/ ` ?

PR #20367, opened 2021-01-04
Decreasing netpol parallel jobs

knabben:
Netpol tests are flaky, decreasing the concurrent runs to see if stay less flaky. We are still investigating the cases here.

### DIFF
--- a/config/jobs/kubernetes/sig-network/sig-network-misc.yaml
+++ b/config/jobs/kubernetes/sig-network/sig-network-misc.yaml
@@ -17,7 +17,7 @@ presubmits:
     spec:
       containers:
       - args:
-        - --timeout=70
+        - --timeout=170
         - --root=/go/src
         - --repo=k8s.io/kubernetes=$(PULL_REFS)
         - --repo=k8s.io/release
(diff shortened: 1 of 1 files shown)


aojea:
IIRC the netpol tests also have some concurrency internally, I was trying to find and tweak the netpol concurrency before the overall concurrency to see if this solves the problem, otherwise it will be a blocker for a wider adoption of netpol

aojea:
@knabben let's go with this  we need to increase the timeouts too, take into account that there are 2 timeouts per job
I will go with the ones in the `  name: ci-kubernetes-e2e-gci-gce-kube-dns` job, 170 and 150

      - --timeout=170
      - --bare
      - --scenario=kubernetes_e2e
      - --
      - --check-leaked-resources
      - --cluster=
      - --env=CLUSTER_DNS_CORE_DNS=false
      - --extract=ci/latest
      - --gcp-zone=us-west1-b
      - --ginkgo-parallel=30
      - --provider=gce
      - --test_args=--ginkgo.skip=\[Serial\]|\[Disruptive\]|\[Flaky\]|\[Feature:.+\] --minStartupPods=8
      - --timeout=150m

aojea:
please update the interval of the periodic job to run every 4 hours please too

PR #20375, opened 2021-01-05
test: Avoid running e2e tests on doc change

nearora-msft:
This PR is to avoid running e2e tests when doc(.md) files are changed

Fixes : https://github.com/kubernetes-sigs/azuredisk-csi-driver/issues/427


chewong (changes requested):
The regex is not supported by Go (See https://groups.google.com/g/golang-nuts/c/7qgSDWPIh_E). Could you try `run_if_changed: .*[^m][^d]$`?

andyzhangx (approved):
/lgtm
/approve

Issue #20363, opened 2020-12-31
[Kettle] PubSub subscriptions are not filtered and it causes more overhead


MushuEE:
**What happened**:
Current Kettle PubSub subscriptions trigger on ALL gcs changes. This causes many more messages to be processed than needed. The filtering is a newer feature and can be added now.

**What you expected to happen**:
Only the `finished.json` object should trigger a pubsub message Pull.

**How to reproduce it (as minimally and precisely as possible)**:

**Please provide links to example occurrences, if any**:

**Anything else we need to know?**:

/assign

MushuEE:
Not possible to filter on Suffix currently

PR #20383, opened 2021-01-05
[Kettle] Filter on subscription to avoid check in code

MushuEE:
In researching a solution to #20363 I found that we are unable to filter with a suffix. However, we can filter by eventType. I updated the staging subscription to filter non `OBJECT_FINALIZE` events. This should save at least some of the message processing overhead. 

/hold need to test in staging. 
/area kettle
/cc @amwat 

```
Subscription filter 
attributes.eventType = "OBJECT_FINALIZE"
```


amwat (commented):
/lgtm

does this auto deploy to prod?
would like to see this in staging only first

MushuEE:
It is deployed to staging right now. I checked that it is no longer getting only OBJECT_FINALIZE events now. But I need to rebuild staging with the other PR using the staging subscription to test.

amwat:
is this still on?

MushuEE:
Yes, I made this change in staging and never completed because I didn't feel comfortable nuking the existing prod Subscription.

I created a new one : `kettle-filtered` and will switch over before removing the old unfiltered one.

@amwat

MushuEE:
/hold
got this branch in a bad state because it was so old

PR #20384, opened 2021-01-05
Trigger: Retry ProwJob creation

alvaroaleman:
Triggers Prowjob creation is entriely eventbased and if it encounters a
transient error, the user has to manually issue a /retest or /test
jobame to recover from that situation. This change makes us retry with
an exponential backoff instead.

### DIFF
--- a/prow/plugins/trigger/trigger.go
+++ b/prow/plugins/trigger/trigger.go
@@ -20,11 +20,16 @@ import (
 	"context"
 	"fmt"
 	"strings"
+	"time"
 
 	"github.com/sirupsen/logrus"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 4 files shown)


prow/plugins/trigger/trigger.go
```diff
@@ -322,3 +327,33 @@ func getPostsubmits(log *logrus.Entry, gc git.ClientFactory, cfg *config.Config,
 	}
 	return postsubmits
 }
+
+// createWithRetry will retry the cration of a ProwJob. The Name must be set, otherwise we might end up creating it multiple times
```
matthyx, line 331: do we need to check for the name to be set?

alvaroaleman: I think it's fine, we don't have any code that uses GenerateName today but use the UUID lib for that everywhere.
/hold cancel

PR #20387, opened 2021-01-06
Skip presubmits for gh-pages branch in fsx, efs csi repos

wongma7:
I marked the gh-pages branches unprotected in https://github.com/kubernetes/test-infra/pull/20323/files but there are still presubmits configured to run there, so branchprotector sees that as a conflict.

Fix by removing the presubmits from gh-pages branch, same as how it was done for ebs. https://github.com/kubernetes/test-infra/pull/20026

REF: https://prow.k8s.io/?job=ci-test-infra-branchprotector


ayberk:
LGTM

PR #20389, opened 2021-01-06
Fix CAPI e2e-k8s-latest-main using env variables

fabriziopandini:
This PR fixes a temporary job created for testing https://github.com/kubernetes-sigs/cluster-api/pull/3916 by adding env variables.
If this works, then it will be easier to create all the permutations for the Kubernetes version in scope for the CAPI test matrix.

/cc @vincepri


vincepri (approved):
/approve
/lgtm

PR #20391, opened 2021-01-06
Drop kubeflow-pipelines-components-gcp-python27 from kubeflow-presubmits.yaml

chensun:
We won't support Python 2.7 any more post https://github.com/kubeflow/pipelines/pull/4960


chensun:
obsolete.

PR #20392, opened 2021-01-06
Add git to the image for the generic autobumper

mpherman2:
The autobumper uses git, and it is not available by default in the image, so it needs to be installed.


spiffxp (approved):
/approve
/lgtm

PR #20394, opened 2021-01-06
sig-windows: Remove windows tests for 1.16

jsturtevant:
1.16 branch is no longer active: https://github.com/kubernetes/sig-release/blob/master/releases/patch-releases.md#non-active-branch-history

/sig windows
/cc @chewong @marosset


chewong (approved):
/lgtm
/approve

Issue #19575, opened 2020-10-14
TODOs for sig-auth-serviceaccount-admission-controller-migration upgrade test 


zshihang:
**What should be cleaned up or changed**:

- use sig auth mailing list as `testgrid-alert-email`
- use release/stable-x.y as extracts
- lower frequency to 1/day

**Provide any links for context**:

BenTheElder:
still TODO?

liggitt:
yup - https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/sig-auth/serviceaccount-admission-controller-migration-config.yaml

PR #20395, opened 2021-01-07
finish TODO in upgrade test config

zshihang:
fix: #19575

### DIFF
--- a/config/jobs/kubernetes/sig-auth/serviceaccount-admission-controller-migration-config.yaml
+++ b/config/jobs/kubernetes/sig-auth/serviceaccount-admission-controller-migration-config.yaml
@@ -3,15 +3,13 @@ periodics:
   annotations:
     testgrid-dashboards: sig-auth-gce
     testgrid-tab-name: upgrade-tests
-    # TODO: use sig-auth mailing list
-    testgrid-alert-email: zshihang@google.com
+    testgrid-alert-email: kubernetes-sig-auth+testgrid@googlegroups.com
     testgrid-num-failures-to-alert: "3"
   labels:
     preset-service-account: "true"
     preset-k8s-ssh: "true"
   decorate: true
-  # TODO: lower to once a day
-  interval: 2h
+  interval: 24h
   spec:
     containers:
       - image: gcr.io/k8s-testimages/kubekins-e2e:v20210108-3c85f1a-master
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes/sig-auth/serviceaccount-admission-controller-migration-config.yaml
```diff
@@ -24,9 +22,8 @@ periodics:
         - --test=false
         - --check-leaked-resources
         - --check-version-skew=false
-        # TODO: change to release/stable-x.y
-        - --extract=ci/latest
-        - --extract=ci/latest-1.19
+        - --extract=ci/stable
+        - --extract=ci/stable-1.19
```
liggitt, line 26: I'm fuzzy about what these represent... is ci/stable the latest 1.20.x release or the latest commit on release-1.20 branch? Same question for ci/stable-1.19.

If we want to test upgrade from the latest release to current dev branch, should this be `ci/latest` and `ci/stable`?

zshihang: ci/latest | ci/stable 
-|-
v1.21.0-alpha.0 | v1.20.1

ci/latest-1.20 | ci/stable-1.20
-|-
v1.20.2-rc.0 | v1.20.1

yea, ci/stable is the latest 1.20.x while ci/latest has the largest tag number. e.g. v1.21.0-alpha.0

with bound beta, latest or stable will eventually have bound enabled by default and this upgrade will not have much value.

my current plan is fix the initial version to versions with bound alpha (e.g. ci/stable-1.19) and upgrade to ci/stable.

liggitt: unless we want this job to tell us something broke **after** we cut a release, I'm pretty sure the version we're upgrading to should be using **latest** instead of **stable**

given we also didn't enable BoundServiceAccountTokenVolume by default in 1.20, I think this should probably just test from `ci/latest-1.20` to `ci/latest` for now.

zshihang: sounds good.

spiffxp:
#20498 (comment)

PR #20396, opened 2021-01-07
sig-windows: Do not run device test on jobs that don't have gpu

jsturtevant:
This is a follow up to https://github.com/kubernetes/test-infra/pull/20352. I missed the Periodic in the commit and they are causing the same failures in the pre-submits on the PR https://github.com/kubernetes/kubernetes/pull/97491

/sig windows
/cc @marosset @chewong


chewong (approved):
/lgtm
/approve

PR #20397, opened 2021-01-07
[Kettle Stream] Deployment Docs

MushuEE:
/hold this is based off the template PR and will need to be rebased after merging the base.

This is a deployment script. I realized that the stream.py script only takes build changes in the kubernetes-jenkins bucket. So it will not stream from the other buckets listed.

/area kettle

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -24,6 +24,7 @@ replace (
 )
 
 require (
+	cloud.google.com/go v0.66.0
 	cloud.google.com/go/pubsub v1.4.0
 	cloud.google.com/go/storage v1.12.0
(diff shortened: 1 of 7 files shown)


alvaroaleman:
Weird, why did blunderbuss choose me as a reviewer, I am not a root reviewer or approver :thinking: 

I don't really have much context on kettle, so will uncc. Thanks for your work!

/uncc

MushuEE:
@alvaroaleman was my bad, I didn't rebase correctly and it added every tag and sourced two random reviewers because I had so many commit pulled in. Thanks!

PR #20405, opened 2021-01-07
[Kettle] Make subscription path a deployment variable

MushuEE:
@amwat quick pr follow up making subscription a deployment env.

### DIFF
--- a/kettle/deployment-staging.yaml
+++ b/kettle/deployment-staging.yaml
@@ -20,6 +20,8 @@ spec:
           value: "1000"
         - name: DEPLOYMENT
           value: staging
+        - name: SUBSCRIPTION_PATH
+          value: kubernetes-jenkins/gcs-changes/kettle-staging
         volumeMounts:
         - name: service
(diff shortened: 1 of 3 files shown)


spiffxp:
pylint error

I'm also beginning to wonder if 3 env vars for passing in parameters is a sign that we should switch to using command line flags, but won't block on that

amwat:
+1 to making them flags instead of envs, fine as a follow up.
/lgtm

MushuEE:
@spiffxp, not sure how we would get around that. The entry point is runner.sh, would 
```
      containers:
      - name: kettle
        image: gcr.io/k8s-testimages/kettle:latest
        command: "/kettle/runner.sh"]
        args: ["some", "args"]
```
we do it like that per deployment and not use the default entry point?

amwat:
we can leave the entrypoint as is and just provide the args.
https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/#notes

PR #20407, opened 2021-01-07
Adding additional presets to get correct VMC creds with extra permissions

kkeshavamurthy:
CC: @codenrhoden


codenrhoden:
/ok-to-test
/lgtm

We want these vSphere CCM creds for testing.

PR #20415, opened 2021-01-08
Scrape node exporter in 5k tests

mborsz:
It is already enabled in 100 node test. I also started 5k manual test run and it works there as well.

/assign @wojtek-t


wojtek-t:
/lgtm

[Though let's merge on Monday, not Friday afternoon.]
/hold

PR #20417, opened 2021-01-08
Rewrite merge_kubernetes_secret.py in Go, plus GCP secret manager backup

chaodaiG:
Rewrite the python script in Go, with:

- Use k8s client library for interacting with cluster instead of shelling out to kubectl.
- Add a step of backing up secret to GCP secret manager before updating.
- Add unit tests.

/assign @cjwagner @fejta 
/cc @alvaroaleman


alvaroaleman:
We are not going to use this in Openshift, so I am not going to comment on the code, I can just repeat what I already said when Cole updated the python script for this: Merging kubeconfigs is a super bad idea because:
* You are unable to find out in your config git history when a certain cluster was added/removed, because it does not result in a config change
* Since there is no change in configs, components won't get redeployed, so if there is an issue with a kubeconfig, you will notice the next time a component gets restarted (most likely Prow upgrade) which is a different point in time, which makes correlating hard
* You have to solve the problem of merging kubeconfig

I can just strongly suggest to use the env var and have a dedicated kubeconfig per cluster to avoid all of this like e.G. here: https://github.com/openshift/release/blob/4b6b42276b95d8688c9327158b28aaa5dcdfd2c5/clusters/app.ci/prow/03_deployment/hook.yaml#L94-L96

chaodaiG:
Thank you @alvaroaleman for the input! I have synced with Cole and we agree that we are suffering from no version controlled history of adding/removing certain build cluster, and more importantly we're going to run into the 1Mb limit of a secret. So I think I'll drop this PR and we'll adopt the openshift model instead.

PR #20419, opened 2021-01-08
golang version of transfigure command

joshua-bone:


testgrid/cmd/transfigure/cmd/transfigure.go
```diff
@@ -0,0 +1,363 @@
...
+	"strings"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/spf13/cobra"
+)
+
+const Branch = "transfigure-branch"
+
+var (
+	dryRun         bool
```
chaodaiG, line 37: Can we put these in a struct like https://github.com/kubernetes/test-infra/blob/6d943af9d0a0bc93da838ba0ab19414e0370306c/gencred/cmd/gencred/main.go#L45, to avoid so many global vars?

testgrid/cmd/transfigure/cmd/transfigure.go
```diff
@@ -0,0 +1,363 @@
...
+	remoteForkRepo string
+	repoSubdir     string
+	tokenContents  string
+	testgridDir    string
+	testgridYaml   string
+	workingDir     string
+)
+
+// rootCmd represents the base command when called without any subcommands
+var rootCmd = &cobra.Command{
```
chaodaiG, line 53: I'm afraid that using cobra is an overkill in this case, since there is no subcommand for this tool, use standard flag parsing as https://github.com/kubernetes/test-infra/blob/6d943af9d0a0bc93da838ba0ab19414e0370306c/gencred/cmd/gencred/main.go#L63 should be suffice

testgrid/cmd/transfigure/cmd/transfigure.go
```diff
@@ -0,0 +1,363 @@
...
+		err := run()
+		if err != nil {
+			log.Fatal(err)
+		}
+	},
+}
+
+// Execute adds all child commands to the root command and sets flags appropriately.
+// This is called by main.main(). It only needs to happen once to the rootCmd.
+func Execute() {
```
chaodaiG, line 69: Generally in Go it's the caller's responsibility to decide whether to kill the program or not. If it's really desired to kill in helper method I would suggest rename it with `ExecuteMustSucceed`

chaodaiG: By the way this method doesn't have to be exported as it's only consumed within the package

testgrid/cmd/transfigure/cmd/transfigure.go
```diff
@@ -0,0 +1,363 @@
...
+
+func run() error {
+	if !dryRun && githubToken == "" {
+		return errors.New("Run was not a dry run, and flag 'github_token' was not set.")
+	}
+
+	err := createTempWorkingDir()
+	if err != nil {
+		return err
+	}
```
chaodaiG, line 84: The Go way:
```suggestion
	if err := createTempWorkingDir(); err != nil {
		return err
	}
```

And same below

testgrid/cmd/transfigure/cmd/transfigure.go
```diff
@@ -0,0 +1,363 @@
...
+	rootCmd.PersistentFlags().StringVar(&gitEmail, "git_email", "", "The GitHub email to use. Pulled from github_token if not specified.")
+	rootCmd.PersistentFlags().BoolVar(&dryRun, "dry_run", false, "If true, no PR will be created.")
+}
+
+func createTempWorkingDir() error {
+	tempDir, err := ioutil.TempDir(".", "transfigure_")
+	if err != nil {
+		return wrapErrorOrNil("Error creating temp dir", err)
+	}
+	workingDir, err = filepath.Abs(tempDir)
```
chaodaiG, line 174: Would be better return `workingDir` in this function to avoid global var

testgrid/cmd/transfigure/cmd/transfigure.go
```diff
@@ -0,0 +1,363 @@
...
+	if err != nil {
+		return wrapErrorOrNil("Error getting abs path for working dir", err)
+	}
+	log.Printf("Created temp directory %v/", tempDir)
+	return nil
+}
+
+func removeTempWorkingDir() {
+	os.RemoveAll(workingDir)
+	log.Printf("Removed temp directory %v/", workingDir)
```
chaodaiG, line 184: This will execute even if previous step failed, you might want to handle error above

testgrid/cmd/transfigure/cmd/transfigure.go
```diff
@@ -0,0 +1,363 @@
...
+}
+
+func cloneK8sTestInfraRepo() error {
+	log.Printf("Cloning kubernetes/test-infra into temp directory")
+	cmd := exec.Command("git", "clone", "https://github.com/kubernetes/test-infra.git")
```
chaodaiG, line 189: Can we make this value `https://github.com/kubernetes/test-infra.gits` as const?

testgrid/cmd/transfigure/cmd/transfigure.go
```diff
@@ -0,0 +1,363 @@
...
+	return nil
+}
+
+func removeTempWorkingDir() {
+	os.RemoveAll(workingDir)
+	log.Printf("Removed temp directory %v/", workingDir)
+}
+
+func cloneK8sTestInfraRepo() error {
+	log.Printf("Cloning kubernetes/test-infra into temp directory")
```
chaodaiG, line 188: There is no format here, use `Print` maybe?

testgrid/cmd/transfigure/cmd/transfigure.go
```diff
@@ -0,0 +1,363 @@
...
+	cmd := exec.Command(
+		"/Users/joshuabone/go/src/k8s.io/test-infra/testgrid/cmd/transfigure/configurator",
+		fmt.Sprintf("--prow-config=%s", prowConfig),
+		fmt.Sprintf("--prow-job-config=%s", prowJobConfig),
+		"--output-yaml",
+		fmt.Sprintf("--yaml=%s", yamlPath),
+		"--oneshot",
+		fmt.Sprintf("--output=%s/gen-config.yaml", testgridDir),
+	)
+	_, err = runCmd(cmd)
```
chaodaiG, line 223: Might want to print `configurator` output for debugging purpose

testgrid/cmd/transfigure/cmd/transfigure.go
```diff
@@ -0,0 +1,363 @@
...
+func gitDiffExists() bool {
+	cmd := exec.Command("git", "diff", "--cached", "--quiet", "--exit-code")
+	cmd.Dir = workingDir
+	err := cmd.Run()
+	return err != nil
+}
+
+func runBazelTests() error {
+	log.Print("Running kubernetes/test-infra tests...")
+	cmd := exec.Command("bazel", "test", "test-infra/config/tests/...", "test-infra/hack:verify-spelling")
```
chaodaiG, line 241: Make these consts as well

testgrid/cmd/transfigure/cmd/transfigure.go
```diff
@@ -0,0 +1,363 @@
...
+	out, err := runCmd(cmd)
+	if err != nil {
```
chaodaiG, line 243: Why not simply return err? `out` should be printed no matter there is error or not imo

testgrid/cmd/transfigure/cmd/transfigure.go
```diff
@@ -0,0 +1,363 @@
...
+		return wrapErrorOrNil("Error running bazel test: "+out, err)
+	} else {
+		log.Print("Tests successful!")
+		return nil
+	}
+}
+
+func ensureGitUserAndEmail() error {
+	log.Print("Checking Git Config user and email")
+	runCmd(exec.Command("git", "config", "user.name", gitUser))
```
chaodaiG, line 253: If intentionally opting out of error checking there should be a comment

testgrid/cmd/transfigure/cmd/transfigure.go
```diff
@@ -0,0 +1,363 @@
...
+	runCmd(exec.Command("git", "config", "user.email", gitEmail))
+
+	output, err := runCmd(exec.Command("git", "config", "user.name"))
```
chaodaiG, line 256: If step above succeeded then you shouldn't need to do this check

testgrid/cmd/transfigure/cmd/transfigure.go
```diff
@@ -0,0 +1,363 @@
...
+		return wrapErrorOrNil("Error on git config user.email", err)
+	}
+	if diff := cmp.Diff(output, gitEmail); diff != "" {
+		return errors.New("Unexpected Git Email: (-got +want)\n" + diff)
+	}
+	return nil
+}
+
+func gitCommitAndPush() error {
+	prTitle = "Update TestGrid for " + repoSubdir
```
chaodaiG, line 275: This should be a global const imo

testgrid/cmd/transfigure/cmd/transfigure.go
joshua-bone: I really like Cobra for things like automatic help generation and would rather not write those things myself. Is it ok to leave as is?

testgrid/cmd/transfigure/cmd/transfigure.go
joshua-bone: Done

testgrid/cmd/transfigure/cmd/transfigure.go
joshua-bone: Moved the os.exit call to main(). I got errors when trying to make Execute() non exported, so leaving as is.

testgrid/cmd/transfigure/cmd/transfigure.go
joshua-bone: Done.

testgrid/cmd/transfigure/cmd/transfigure.go
joshua-bone: Done.

testgrid/cmd/transfigure/cmd/transfigure.go
joshua-bone: Done.

testgrid/cmd/transfigure/cmd/transfigure.go
joshua-bone: Added error checking.

testgrid/cmd/transfigure/cmd/transfigure.go
```diff
@@ -0,0 +1,362 @@
...
+}
+
+func generateTestgridYAML() error {
+	yamlPath, err := filepath.Abs(o.testgridYaml)
+	if err != nil {
+		return wrapErrorOrNil("Invalid testgrid yaml path "+o.testgridYaml, err)
+	}
+
+	cmd := exec.Command(
+		"/Users/joshuabone/go/src/k8s.io/test-infra/testgrid/cmd/transfigure/configurator",
```
chases2, line 213: This can't be the correct path for all users, right? Does this work from an image?

testgrid/cmd/transfigure/cmd/transfigure.go
```diff
@@ -0,0 +1,362 @@
...
+	// Create a new HTTP request.
+	req, err := http.NewRequest("GET", "https://api.github.com/user", nil)
+	if err != nil {
+		return err
+	}
+	// Append the token we just read from file, and send the request.
+	req.Header.Add("Authorization", "token "+o.tokenContents)
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return wrapErrorOrNil("Error sending request", err)
```
chases2, line 321: If you know that err isn't nil, I would use a direct `fmt.Errorf("Error sending GitHub request: %v", err)` call instead.

testgrid/cmd/transfigure/main.go
```diff
@@ -0,0 +1,30 @@
...
+
+import (
+	"fmt"
+	"os"
+
+	transfigure "k8s.io/test-infra/testgrid/cmd/transfigure/cmd"
+)
+
+func main() {
+	if err := transfigure.Execute(); err != nil {
```
chases2, line 26: Having a "main.go" and a "transfigure.go" file is a good idea, but I would change what goes into which file.

main.go would be a good place to handle flags and options. Summarizer has an [example](https://github.com/GoogleCloudPlatform/testgrid/blob/1844fdad5f0b8ddc2101e52069cca0209403e765/cmd/summarizer/main.go#L54) of this pattern.

Handling flags is complex, since users might specify these options with flags or in the old "list everything in the right order" way. Using Cobra for this seems great if that's what you're comfortable with, but the options parsing should pass the data it cares about (or an options struct) in the functions. Without that, everything is too coupled together to test easily.

testgrid/cmd/transfigure/cmd/transfigure.go
joshua-bone: Done.

chases2 (commented):
There are some nits throughout, but my big questions are:
- What sort of testing has been done to show that this works?
- Do you intend to have this image replace the existing transfigure image, or be a new "Transfigure 2" that users have to migrate to? I can't tell which you intend just from looking at the code.

testgrid/cmd/transfigure/cmd/transfigure.go
joshua-bone: Done

chaodaiG:
Overall lgtm to me. 
/approve

Please fix tests, these commands might be helpful:
bazel run @io_k8s_repo_infra//hack:update-bazel
bazel run @io_k8s_repo_infra//hack:update-deps

Please also answer @chases2 's comments

chases2:
@joshua-bone and I spoke about this PR. The idea is to rewrite transfigure (as a separate version) in a way that would make it run more nicely with RBE. Its arguments are not meant to be backwards-compatible; users would have to consciously switch if they want to.

In that case, I'm fine with the code so long as a README is added (or the [transfigure readme](https://github.com/kubernetes/test-infra/blob/master/testgrid/cmd/transfigure/README.md) changed) that explains a bit about the two versions of the code.

You will also need to run the bazel update commands chaodaiG mentioned to make the presubmit happy. Thanks so much for your work on this! :D

testgrid/cmd/transfigure/main.go
joshua-bone: Done

testgrid/cmd/transfigure/cmd/transfigure.go
joshua-bone: Done

testgrid/cmd/transfigure/cmd/transfigure.go
joshua-bone: Done

testgrid/cmd/transfigure/cmd/transfigure.go
joshua-bone: Done

testgrid/cmd/transfigure/cmd/transfigure.go
joshua-bone: Done

Issue #20421, opened 2021-01-08
client-go and boskos imports lead to go.mod replace hacks


howardjohn:
**What should be cleaned up or changed**:

The go.mod file imports client-go@v11, but replaces it. The result is anyone importing k8s.io/test-infra needs to do the same - essentially we are polluting the downstream entire golang ecosystem, and propagating this `replace` requirement. In go 1.16, `go install` doesn't work if a go.mod has `replace` in it as well, making it even more annoying.

The fix, however, is not trivial. k8s.io/test-infra depends on sigs.k8s.io/boskos, and sigs.k8s.io/boskos depends on k8s.io/test-infra. This means we need both dependencies to change at the same time to break the "poisoned" client-go@v11 version.

I don't know a good way to actually change these in tandem. Its easy locally with temporary `replace ..` directives. Otherwise, it seems like it may require to push a broken commit of one of the repos, referring to the not-yet-existing fixed version of the other, but I am not sure that would actually work in practice.

Another alternative is to not have a circular dependency, by dropping one of the imports or merging to a monorepo. I am not sure the work involved there though.

**Provide any links for context**:

Similar issue: https://github.com/knative/test-infra/issues/2414

howardjohn:
test-infra depends on boskos only for `kubetest`. It seems there is a kubetest2 in its own repo, so one option is to move out kubetest, or remove it if its been replaced with kubetest2.

Boskos deps:
```
$ rg --no-filename --no-line-number k8s.io/test-infra | sort | uniq -c | rg -v v0.0.0 | sort
      1         "k8s.io/test-infra/prow/config/secret"
      1         "k8s.io/test-infra/prow/flagutil"
      1         "k8s.io/test-infra/prow/simplifypath"
      2         "k8s.io/test-infra/pkg/flagutil"
      2         "k8s.io/test-infra/prow/pjutil"
      2         prowflagutil "k8s.io/test-infra/prow/flagutil"
      3         "k8s.io/test-infra/prow/config"
      3         "k8s.io/test-infra/prow/interrupts"
      3         prowmetrics "k8s.io/test-infra/prow/metrics"
      8         "k8s.io/test-infra/prow/logrusutil"
```
Seem more involved, but also mostly just smaller utilities that could probably be dropped with some investment

BenTheElder:
https://github.com/kubernetes-sigs/boskos/issues/89
https://github.com/kubernetes/test-infra/pull/20422#issuecomment-871735014

[...]

PR #20422, opened 2021-01-08
Drop boskos import and client-go@v11

howardjohn:
This contains a `replace` for tektoncd, will not be needed when
https://github.com/tektoncd/pipeline/pull/3668 is merged. I am opening
the PR now for early feedback.

This is an attempt at fixing
https://github.com/kubernetes/test-infra/issues/20421. This is done by
making kubetest a submodule. Alternatives are to remove it entirely (if
its fully replaced with kubetest2) or move it to its own repo

If we don't want to make kubetest a submodule, its also possible to merge this, update sigs.k8s.io/boskos to the latest k8s.io/test-infra, then send a new PR to make kubetest not a submodule.


config/prow/cluster/boskos_test.go
```diff
@@ -1,34 +0,0 @@
...
-
-package cluster
-
-import (
-	"testing"
-
-	"sigs.k8s.io/boskos/common"
-)
-
-func TestConfig(t *testing.T) {
```
alvaroaleman, line 25: this should probably be replaced with a presubmit that verifies the config: https://github.com/openshift/release/blob/94da441c473adf1dc6e4c76231b64b6f4a4f9d49/ci-operator/jobs/openshift/release/openshift-release-master-presubmits.yaml#L41-L65

howardjohn: Ah I just realized what this test was actually doing - you are right, looks like we shouldn't just remove it.

BTW, an alternative is to remove it now, then once boskos updates we can add it back. It just depends if we want to stop importing boskos forever, or just to avoid the bad client-go import temporarily. Probably long term its good to not have a circular dependency.

BenTheElder:
kubetest2 won't be able to replace kubetest in all jobs for some time (it's not a super high priority to eliminate, but kubetest2 has been much easier to extend to new tasks), but being a distinct go module SGTM. The rest of test-infra should not be importing kubetest and vice versa.

BenTheElder:
I don't think this trivially maps to bazel though, because bazel has a single set of dependencies and includes the go dependencies.

I'm not sure we need to be able to build kubetest with bazel given it's basically frozen except for bug-fixes, but only if there's a new presubmit covering changes to kubetest/ that ensures it still builds etc since right now we just bazel build //... + bazel test //... the whole repo.

Alternatively if someone has a strong opinion about kuebtest and bazel, we could maybe do a sub-WORKSPACE.

BenTheElder:
> If we don't want to make kubetest a submodule, its also possible to merge this, update sigs.k8s.io/boskos to the latest k8s.io/test-infra, then send a new PR to make kubetest not a submodule.

We can't just update boskos as-is? Do we have in import cycle? :yikes:

Can we not import test-infra in boskos?

howardjohn:
@BenTheElder yes there is a cycle. Info in https://github.com/kubernetes/test-infra/issues/20421. Its _probably_ not that hard to not import test-infra in boskos but I am not familiar with it much - the functions here don't seem to major: https://github.com/kubernetes/test-infra/issues/20421#issuecomment-757018208

alvaroaleman:
> Can we not import test-infra in boskos?

We can, but that would be a bit annoying, as it currently imports a set of utility code from test-infra. I would rather make test-infra not depend on boskos, by replacing the current golang-based config test with the boskos checkconfig binary (which has the sideeffect that we test the config against the same boskos version we actually use, rather than some boskos version from the last time someone bothered to bump the dependency) and make kubetest its own submodule. WDYT?

BenTheElder:
> and make kubetest its own submodule. WDYT?

Have to solve the part where presubmit build + test uses bazel which does not have submodules.

I don't think test-infra is a great place to store utility packages imported by other repos, itself being a mono-repo, and it's not surprising that test-infra tools would want to interact with boskos.

BenTheElder:
I think maybe we can switch kubetest back to a simple local boskos client instead of pulling in the real client. I don't think multi-go-module is going to work as expected with bazel.

BenTheElder:
> I think maybe we can switch kubetest back to a simple local boskos client instead of pulling in the real client. I don't think multi-go-module is going to work as expected with bazel.

I think this is the best path forward, if anyone wants to take this up. Because the boskos libraries are similarly part of the Kubernetes project and under the same ownership, license, etc., we could even more or less just copy in the bits kubetest needs.

It also might be a good thought, on the other hand, to update the boskos project to not publish a client that requires transitively depending on Kubernetes, however I don't know that anyone is actively maintaining there just this moment.

Issue #18789, opened 2020-08-11
Cleanup usage of kubernetes-release-pull in kubernetes presubmits


amwat:
**What should be cleaned up or changed**:
We stage builds to `gs://kubernetes-release-pull` in almost every presubmit job.
But from what I can tell nothing is actually consuming those builds since the jobs also use `extract=local` .
It's a non-trivial overhead to upload the release tars in every presubmit and we should remove all the non-required usages.

**Provide any links for context**:
https://cs.k8s.io/?q=kubernetes-release-pull&i=nope&files=&repos=
https://github.com/kubernetes/test-infra/blob/c4628a3a1c2e4c149f669348f351fe78ebd1f258/kubetest/extract_k8s.go#L449-L466
Random GCE provider job: https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/directory/pull-kubernetes-e2e-gce/1293275406807339008#1:build-log.txt%3A903

/cc @spiffxp @BenTheElder @MushuEE 

---
EDIT(@spiffxp): I made a list of the offending jobs going off the criteria `--extract=local` and `--stage=gs://kubernetes-release-pull/*`
- if the job triggers for a single branch it's labeled as `job@branch`
- if the job triggers for all branches it's labeled as `job`
- there are no presubmits that trigger for N branches (where all > N > 1)
- there are no periodics or postsubmits that touch gs://kubernetes-release-pull
- this picks up some `--provider=aws` jobs (kops), it remains to be seen whether they need `--stage` or not

EDIT(@BenTheElder): I removed the outdated checklist and instead i'm going to provide a search: https://github.com/search?q=repo%3Akubernetes%2Ftest-infra+%22--stage%3Dgs%3A%2F%2Fkubernetes-release-pull%22&type=code

BenTheElder:
we should test this in a canary just because this stuff is old and brittle and I can't remember why we were doing this anymore 🙃

amwat:
update: 
https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/92316/pull-kubernetes-e2e-gce-no-stage/1294345203125063682/#1:build-log.txt%3A355

the local path
https://github.com/kubernetes/test-infra/blob/c4628a3a1c2e4c149f669348f351fe78ebd1f258/kubetest/extract_k8s.go#L450

comes from 
[...]

PR #20427, opened 2021-01-08
Match extract and build strategy for no-stage canary job.

amwat:
ref: #18789 

/cc @spiffxp @BenTheElder


spiffxp (approved):
/approve
/lgtm

PR #20428, opened 2021-01-09
Add check to make sure fork exists for autobump and if not create the fork

mpherman2:
In order for the autobumper to work, the robot doing the autobump must have a fork of the repo on their account. This checks to make sure they do, and if they do not it creates one.

### DIFF
--- a/experiment/autobumper/bumper/bumper.go
+++ b/experiment/autobumper/bumper/bumper.go
@@ -286,6 +286,13 @@ func Run(o *Options) error {
 			}
 		}
 
+		// Check to see if the proper fork exists and if it does not, create one.
+		// TODO(mpherman): Handle case where account has repo of same name as one we are trying to autobump but is not a fork of it (e.g. test-infra)
+		_, err := gc.EnsureFork(o.GitHubLogin, o.GitHubOrg, o.GitHubRepo)
+		if err != nil {
+			return fmt.Errorf("fork needed for autobump does not exist. unable to create new fork. %v", err)
+		}
+
 		stdout := HideSecretsWriter{Delegate: os.Stdout, Censor: &sa}
 		stderr := HideSecretsWriter{Delegate: os.Stderr, Censor: &sa}
 		if err := MakeGitCommit(fmt.Sprintf("git@github.com:%s/%s.git", o.GitHubLogin, o.RemoteName), o.HeadBranchName, o.GitName, o.GitEmail, o.Prefixes, stdout, stderr, versions); err != nil {
--- a/prow/external-plugins/cherrypicker/server.go
+++ b/prow/external-plugins/cherrypicker/server.go
@@ -51,6 +51,7 @@ type githubClient interface {
 	CreateFork(org, repo string) (string, error)
 	CreatePullRequest(org, repo, title, body, head, base string, canModify bool) (int, error)
 	CreateIssue(org, repo, title, body string, milestone int, labels, assignees []string) (int, error)
+	EnsureFork(forkingUser, org, repo string) (string, error)
 	GetPullRequest(org, repo string, number int) (*github.PullRequest, error)
 	GetPullRequestPatch(org, repo string, number int) ([]byte, error)
 	GetPullRequests(org, repo string) ([]github.PullRequest, error)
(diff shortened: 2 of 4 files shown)


experiment/autobumper/bumper/bumper.go
```diff
@@ -286,6 +286,17 @@ func Run(o *Options) error {
 			}
 		}
 
+		// Check to see if the proper fork exists and if it does not, create one.
+		// TODO(mpherman): Handle case where account has repo of same name as one we are trying to autobump but is not a fork of it (e.g. test-infra)
+		_, err := gc.GetRepo(o.GitHubLogin, o.GitHubRepo)
```
chaodaiG, line 291: Can we do better error checking to differentiate from repo not exist vs. random errors? If the error type is not easy to get, maybe try use ListRepo then filter out the names(need to be careful of pagination)?

mpherman2: Done. I originally tried to get just the error code (404) but that isn't piped to the response as far as I can tell, but I got it working with the error string instead.

experiment/autobumper/bumper/bumper.go
```diff
@@ -286,6 +287,21 @@ func Run(o *Options) error {
...
+		if err != nil {
+			if err.Error() == getRepo404ErrMsg {
+				logrus.Infof("fork needed for autobump does not exist, creating new fork of %s/%s for user %s.", o.GitHubOrg, o.GitHubRepo, o.GitHubLogin)
+				_, err := gc.CreateFork(o.GitHubOrg, o.GitHubRepo)
```
chaodaiG, line 296: nit, you can combine this line with the line below

chaodaiG (commented):
overall looks good to me
/lgtm

/hold

experiment/autobumper/bumper/bumper.go
```diff
@@ -286,6 +287,21 @@ func Run(o *Options) error {
 			}
 		}
 
+		// Check to see if the proper fork exists and if it does not, create one.
```
alvaroaleman, line 290: This doesn't consider the fact that a fork needs some time to be ready after being created. Also, we already have code for this problem, can we somehow re-use the logic of one for the other?

https://github.com/kubernetes/test-infra/blob/299a5a1244b05692963ae1380966ca76e9315f18/prow/external-plugins/cherrypicker/server.go#L521

experiment/autobumper/bumper/bumper.go
mpherman2: Done

experiment/autobumper/bumper/bumper.go
```diff
@@ -286,6 +287,20 @@ func Run(o *Options) error {
...
+		// TODO(mpherman): Handle case where account has repo of same name as one we are trying to autobump but is not a fork of it (e.g. test-infra)
+		_, err := gc.GetRepo(o.GitHubLogin, o.GitHubRepo)
+		if err != nil {
+			if err.Error() == getRepo404ErrMsg {
```
alvaroaleman, line 294: Please write [left-aligned code](https://blog.codinghorror.com/flattening-arrow-code/) to avoid hard to read passages like this, i.E. here:
```
if err.Error() != getRepo404ErrMsg {
  return fmt.Errorf("failed to check")
}
// create fork, need need for an else clause
```

chaodaiG: Thanks for the link, that's pretty useful to learn

experiment/autobumper/bumper/bumper.go
mpherman2: Done. I move most of the ensureFork logic to the github client so both cherrypicker and autobumper could use it. PTAL

experiment/autobumper/bumper/bumper.go
```diff
@@ -52,6 +52,7 @@ const (
 
 	errOncallMsgTempl = "An error occurred while finding an assignee: `%s`.\nFalling back to Blunderbuss."
 	noOncallMsg       = "Nobody is currently oncall, so falling back to Blunderbuss."
+	getRepo404ErrMsg  = "status code 404 not one of [200], body: {\"message\":\"Not Found\",\"documentation_url\":\"https://docs.github.com/rest/reference/repos#get-a-repository\"}"
```
alvaroaleman, line 55: This is unused now

prow/github/client.go
```diff
@@ -3561,6 +3562,70 @@ func (c *client) CreateFork(owner, repo string) (string, error) {
 	return resp.Name, err
 }
 
+// EnsureFork checks to see that there is a fork of org/repo in the forkedUsers repositories.
+// If there is not, it makes one, and waits for the fork to be created before returning.
+func (c *client) EnsureFork(forkingUser, org, repo string) (string, error) {
+	// Fork repo if it doesn't exist.
+	fork := forkingUser + "/" + repo
+	repos, err := c.GetRepos(forkingUser, true)
+	logrus.Info(repos)
```
alvaroaleman, line 3571: Debugging remainder?

prow/github/client.go
```diff
@@ -3561,6 +3562,70 @@ func (c *client) CreateFork(owner, repo string) (string, error) {
 	return resp.Name, err
 }
 
+// EnsureFork checks to see that there is a fork of org/repo in the forkedUsers repositories.
+// If there is not, it makes one, and waits for the fork to be created before returning.
+func (c *client) EnsureFork(forkingUser, org, repo string) (string, error) {
```
alvaroaleman, line 3567: Nit: Could you godoc that the returned string is the repos name and that it might be different from the original name due to a naming conflict?

prow/external-plugins/cherrypicker/server.go
```diff
@@ -516,61 +517,12 @@ func (s *Server) createIssue(org, repo, title, body string, num int, comment *gi
...
-		if !r.Fork {
-			continue
-		}
-		if r.FullName == repo {
-			return true
-		}
-	}
-	return false
+	// fork repo if it doesn't exsit
+	s.ghc.EnsureFork(s.botUser.Login, org, repo)
```
alvaroaleman, line 523: Can we handle the error, please?

alvaroaleman (approved):

chaodaiG (approved):

stevekuznetsov:
@chaodaiG @mpherman2 every single time that there's a `_, err :=` for these calls in this PR, it's 100% wrong. I'm a bit surprised those bugs are here, since they ignore the case where the fork is of a different name from the original, which is the entire point of these changes, no?

experiment/autobumper/bumper/bumper.go
```diff
@@ -286,6 +286,13 @@ func Run(o *Options) error {
 			}
 		}
 
+		// Check to see if the proper fork exists and if it does not, create one.
+		// TODO(mpherman): Handle case where account has repo of same name as one we are trying to autobump but is not a fork of it (e.g. test-infra)
+		_, err := gc.EnsureFork(o.GitHubLogin, o.GitHubOrg, o.GitHubRepo)
```
stevekuznetsov, line 291: This _cannot_ be ignored, you must keep track of what this is called if you're going to interact with it.

prow/external-plugins/cherrypicker/server.go
```diff
@@ -516,61 +517,15 @@ func (s *Server) createIssue(org, repo, title, body string, num int, comment *gi
...
-			}
-			ghErr = ""
-			if repoExists(owner+"/"+name, []github.Repo{repo.Repo}) {
-				return nil
-			}
-		case <-after:
-			return fmt.Errorf("timed out waiting for %s to appear on GitHub%s", owner+"/"+name, ghErr)
-		}
+	// fork repo if it doesn't exsit
+	if _, err := s.ghc.EnsureFork(s.botUser.Login, org, repo); err != nil {
```
stevekuznetsov, line 523: This _cannot_ be ignored, this change breaks cherry-picker in cases where the fork is not named the same as the repo.

mpherman2: https://github.com/kubernetes/test-infra/pull/20826 Thanks for pointing this out.

PR #20434, opened 2021-01-10
disable iperf test for network policy jobs

aojea:
The IPerf test is the only flake remaining
https://testgrid.k8s.io/sig-network-gce#presubmit-network-policies,%20google-gce

Let's skip it by now, it is an area that we didn't cover in CI yet, so it may require more work before promoting it to run in presubmit jobs

### DIFF
--- a/config/jobs/kubernetes/sig-network/sig-network-misc.yaml
+++ b/config/jobs/kubernetes/sig-network/sig-network-misc.yaml
@@ -47,7 +47,7 @@ presubmits:
         - --provider=gce
         # Skipping snat tests probably GCE related? https://github.com/kubernetes/test-infra/issues/20321
         # Skipping Cloud Provider specific tests: LoadBalancer, ESIPP (Source IP Preservation)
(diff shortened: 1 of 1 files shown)


jayunit100:
@aojea ironically matt thinks is PR https://github.com/kubernetes/kubernetes/pull/94015/files may fix this .... would it be ok to just wait 24 hours before we disable these tests from the netpol job ?

mattfenwick:
Thanks @aojea, I just reported a flake for this -- https://github.com/kubernetes/kubernetes/issues/97890. 
We also have https://github.com/kubernetes/kubernetes/pull/94015 which might address this flake.

Which would you rather move forward with -- disable the job, or fix the flake?

aojea:
ahh, somehow I assumed Matt's was already merged, cool
/hold

aojea:
fix the flake better, of course :)

mattfenwick:
👍 okay sounds good, hopefully we can get the fix merged soon and then I'll keep an eye to see if that addresses the flake!

PR #20437, opened 2021-01-11
sig-scalability: add experimental kubeadm job

neolit123:
/kind feature
/sig scalability

/hold for https://github.com/kubernetes/perf-tests/pull/1649

### DIFF
--- a/config/jobs/kubernetes/sig-scalability/sig-scalability-experimental-periodic-jobs.yaml
+++ b/config/jobs/kubernetes/sig-scalability/sig-scalability-experimental-periodic-jobs.yaml
@@ -146,3 +146,34 @@ periodics:
           - --test-cmd-name=ClusterLoaderV2
           - --timeout=120m
           - --use-logexporter
+
+- name: ci-kubernetes-kubeadm-scalability
+  interval: 60m # TODO: adjust
+  labels:
+    preset-service-account: "true"
+  decorate: true
+  decoration_config:
+    timeout: 60m # TODO: adjust
+  extra_refs:
+  - org: kubernetes
+    repo: perf-tests
+    base_ref: master
+    path_alias: k8s.io/perf-tests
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes/sig-scalability/sig-scalability-experimental-periodic-jobs.yaml
```diff
@@ -146,3 +146,34 @@ periodics:
...
+    base_ref: master
+    path_alias: k8s.io/kubernetes
+  spec:
+    containers:
+    - image: gcr.io/k8s-testimages/kubekins-e2e:v20210107-e4af2d6-master
+      env:
+      - name: NUM_WORKERS
+        value: "10" # TODO: adjust
+      command:
+      - "runner.sh"
```
mm4tt, line 173: Repeating the question from the perf-test PR. Why cannot we use kubetest to take advantage of the boilerplate it implements, e.g. boskos integration? Using kubetest will also make sure that we haven't forgotten about anything, e.g. log dumping. Here is the bash deployer that is based on the kube-up.sh and used by the scale tests - https://github.com/kubernetes/test-infra/blob/ace236a80ac1eab0235507c5ea4cec2761fc123d/kubetest/bash.go

neolit123: this is using the so called "pod utils", i think having the boskos support integrated would be nicer. i cannot seem to find examples on how to use the `--deployer=node ` that you've mentioned.

lyzs90: you're looking for `--deployment=node` i believe

neolit123:
https://github.com/kubernetes/perf-tests/pull/1649#issuecomment-856098997

Issue #20439, opened 2021-01-11
Prow prometheus ran out of disk space


chaodaiG:
**What happened**:

Prometheus alert fired with error: 
```
[FIRING:1] ServiceLostHA prometheus (prow-monitoring/prow critical monitoring)
The service prometheus has at most 1 instance for 5 minutes.
```
See https://kubernetes.slack.com/archives/C7J9RP96G/p1610377648195000

**What you expected to happen**:

Prometheus works without problem.

**Anything else we need to know?**:

It's expected to have two prometheus pods running, but one of them kept on crash looping: 

```
NAME                                   READY   STATUS             RESTARTS   AGE
prometheus-prow-0                      3/3     Running            0          4d1h
prometheus-prow-1                      2/3     CrashLoopBackOff   74         4d1h
```

Checking it's logs and got:

```
level=info ts=2021-01-11T17:01:45.688403124Z caller=main.go:489 msg="Stopping scrape discovery manager..."
level=info ts=2021-01-11T17:01:45.688416615Z caller=main.go:503 msg="Stopping notify discovery manager..."
level=info ts=2021-01-11T17:01:45.688423764Z caller=main.go:525 msg="Stopping scrape manager..."
level=info ts=2021-01-11T17:01:45.688431383Z caller=main.go:499 msg="Notify discovery manager stopped"
level=info ts=2021-01-11T17:01:45.688458895Z caller=main.go:485 msg="Scrape discovery manager stopped"
level=info ts=2021-01-11T17:01:45.688474027Z caller=manager.go:736 component="rule manager" msg="Stopping rule manager..."
level=info ts=2021-01-11T17:01:45.688488811Z caller=manager.go:742 component="rule manager" msg="Rule manager stopped"
level=info ts=2021-01-11T17:01:45.688499755Z caller=notifier.go:521 component=notifier msg="Stopping notification manager..."
level=info ts=2021-01-11T17:01:45.688510184Z caller=main.go:679 msg="Notifier manager stopped"
level=info ts=2021-01-11T17:01:45.688601397Z caller=main.go:519 msg="Scrape manager stopped"
level=error ts=2021-01-11T17:01:45.688785224Z caller=main.go:688 err="opening storage failed: zero-pad torn page: write /prometheus/wal/00016631: no space left on device"
```

chaodaiG:
[...]

PR #20440, opened 2021-01-11
Decrease prometheus retention to 90 days from 180 days

chaodaiG:
Prometheus is crash looping due to disk full: https://github.com/kubernetes/test-infra/issues/20439

Decrease the retention for now before we figure out how to increase the volume size appropriately

### DIFF
--- a/config/prow/cluster/monitoring/prow_prometheus.yaml
+++ b/config/prow/cluster/monitoring/prow_prometheus.yaml
@@ -16,7 +16,7 @@ spec:
         resources:
           requests:
             storage: 100Gi
-  retention: "180d"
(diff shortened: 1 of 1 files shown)


cjwagner (approved):
/hold

chaodaiG:
@spiffxp , could you please take a look and confirm whether this works for you?

spiffxp:
/hold cancel
I'll live with 90d

chaodaiG:
Unfortunately this doesn't seem to have solved the problem, the pod is still failing to start with the same reason, it doesn't seem like that prometheus does pruning before doing it's business

PR #20442, opened 2021-01-11
WIP: Add docs and stage Kettle Metics Charts

MushuEE:
/area metrics
/cc @amwat @BenTheElder @spiffxp @RobertKielty 

DataStudio has an `<iframe>` element that will place the Chart as flat HTML
I added a new bucket folder with the other metrics that can Host this element and shows the real updated Graph. To create the illusion that the chart is in the README, I took a screenshot and linked to it. 

This will all change as the chart changes. Also I can not make the Chart view 100% public so I will need to work on that.


MushuEE:
image
Will see this if not made public

deads2k:
> Will see this if not made public

Thanks, I still can't access it.

MushuEE:
TODO: create from CNCF account that allows public sharing.

MushuEE:
/hold
Need to do this from a public account

amwat:
are we tracking the community owned datastudio anywhere?

spiffxp:
https://github.com/kubernetes/k8s.io/issues/1590

metrics/reports/metrics-report.html
```diff
@@ -0,0 +1,2 @@
+<!-- This file can be redployed with `gsutil cp ./metrics-report.html gs://k8s-metrics/k8s-metrics-report/metrics-report.html`-->
```
BenTheElder, line 1: for some of these other files we now have prowjobs do this

BenTheElder:
I think we still want this, but we are blocked on https://github.com/kubernetes/k8s.io/issues/1590

PR #20447, opened 2021-01-12
First pass; creating a new gce periodic for ubuntu

geojaz:
I'm working on the test case which will land in https://github.com/kubernetes/kops/pull/10555


geojaz:
/hold cancel 
I think this is just about ready. Dumb question- how do I add args to kOps to my kubetest2 invocation? I want to set the image, etc

hakman:
I am not sure if this is easier as periodic or a presubmit test.

geojaz:
@hakman @rifelpet I'd be happy to do this as a presubmit- I worried that we might not have a green build on it immediately and didn't wanna block other work... is that the case? Also, I'm not totally sure that I'm able to tweak the appropriate options yet (at least via kubetest2). Is there something I'm missing where I can pass flags and such?

hakman:
@geojaz I was proposing a presubmit job that has `always_run: false`. This means that you need to write a comment like `/test job-name` to trigger it and won't block any other job. The benefit would be that you won't need to merge various iterations of the fix to see if if it works.

geojaz:
Cool thanks for the info @hakman, that makes sense to me. I just wanted to make sure I wasn't missing something on the args. And thanks for explaining the flags for presubmits! I'll update it accordingly and get the job started in this PR and I'll do some iteration on kubetest2 to get the right handles to tweak options in another :)

rifelpet:
I created a dedicated makefile for e2e jobs [here](https://github.com/kubernetes/kops/blob/master/tests/e2e/e2e.mk). I have plans to templatize it so that we can define many test jobs without a bunch of copy/pasted make targets, but for now feel free to add the new target there.

geojaz:
Sweet more pieces of the puzzle! :D

PR #20451, opened 2021-01-12
First prow integration test: sinker

chaodaiG:
This is the first integration test added for prow, as designed in https://docs.google.com/document/d/1hIHIoApoR4OUs_esBDE7A778wi-jUEZcr2-a0zVTqW0/edit. 

The integration test deploys prow components in KIND cluster and test prow functions inside the cluster.

/assign @cjwagner @alvaroaleman @fejta


chaodaiG:
The integration test took 6 min 40 sec: https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/test-infra/20451/pull-test-infra-integration/1349025501456371712

matthyx:
awesome
/lgtm
/hold

prow/test/integration/setup-cluster.sh
```diff
@@ -0,0 +1,90 @@
...
+  namespace: kube-public
+data:
+  localRegistryHosting.v1: |
+    host: "localhost:${reg_port}"
+    help: "https://kind.sigs.k8s.io/docs/user/local-registry/"
+EOF
+
+# Install nginx and wait for it ready
+echo "Install nginx on kind cluster"
+kubectl --context=${CONTEXT} apply -f https://raw.githubusercontent.com/kubernetes/ingress-nginx/master/deploy/static/provider/kind/deploy.yaml
```
petr-muller, line 88: Can we rely on this URL to stay this way? How likely is this to break?

prow/test/integration/setup-cluster.sh
```diff
@@ -0,0 +1,90 @@
...
+metadata:
+  name: local-registry-hosting
+  namespace: kube-public
+data:
+  localRegistryHosting.v1: |
+    host: "localhost:${reg_port}"
+    help: "https://kind.sigs.k8s.io/docs/user/local-registry/"
+EOF
+
+# Install nginx and wait for it ready
```
petr-muller, line 86: We don't seem to wait for it here though.

prow/test/integration/setup-prow.sh
```diff
@@ -0,0 +1,55 @@
...
+  kubectl --context=${CONTEXT} wait pod \
+    --for=condition=ready \
+    --selector=app=${pod} \
+    --timeout=180s
+done
+
+echo "Push test image to registry"
+docker pull busybox
+docker tag busybox:latest localhost:5000/busybox:latest
+docker push localhost:5000/busybox:latest
```
petr-muller, line 55: Won't we hit the dockerhub rate limits?

prow/test/integration/test/sinker_test.go
```diff
@@ -0,0 +1,183 @@
...
+
+			clusterContext := getClusterContext()
+			t.Logf("Creating client for cluster: %s", clusterContext)
+
+			// defaultKubeconfig := os.Getenv("KUBECONFIG")
+			// stat, err := os.Lstat(defaultKubeconfig)
+			// if err != nil {
+			// 	t.Fatalf("Failed stat %q: %v", defaultKubeconfig, err)
+			// }
+			// t.Logf("Stat of %q: %v\n\n%v\n\n%v", defaultKubeconfig, stat.Mode(), stat.Sys(), stat)
```
petr-muller, line 66: debug code? do we need to keep this?

prow/test/integration/prow/cluster/100_namespace.yaml
```diff
@@ -0,0 +1,99 @@
...
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  namespace: default
+  name: oauth-token
+data:
+  oauth: ZmFrZW9hdXRodG9rZW4K # From 'fakeoauthtoken'
+---
+apiVersion: apiextensions.k8s.io/v1beta1
```
alvaroaleman, line 14: Would it be possible to use the manifests from the config directory (maybe the starter-s3.yaml) to make sure they are correct?

prow/test/integration/prow/config.yaml
```diff
@@ -0,0 +1,9 @@
+sinker:
+  resync_period: 5s
+  max_prowjob_age: 48h
+  max_pod_age: 48h
+  terminated_pod_ttl: 30m
+
+prowjob_namespace: default
+pod_namespace: test-pods
+log_level: debug
```
alvaroaleman, line 9: You can just omit the job config file, it isn't mandatory (I can not comment on an empty file)

prow/test/integration/setup-cluster.sh
alvaroaleman: Yeah agreed, we should refer a concrete revision here rather than just `master`

prow/test/integration/test.sh
```diff
@@ -0,0 +1,38 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit
+
+CURRENT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd -P )"
+
+if [[ -n "${CI:-}" ]]; then
```
alvaroaleman, line 20: I would just check for the presence of the `kind` binary rather than making assumptions about where it is and isn't present

prow/test/integration/test.sh
```diff
@@ -0,0 +1,38 @@
...
+  # TODO(chaodaiG): remove this once kind is installed in test image
+  echo "Install KIND for prow"
+  curl -Lo /usr/bin/kind https://kind.sigs.k8s.io/dl/v0.9.0/kind-linux-amd64
+  chmod +x /usr/bin/kind
+
+  # TODO(chaodaiG): remove this once bazel is installed in test image
+  echo "Install bazel for prow"
```
alvaroaleman, line 27: Same here, checking for thhe presence of bazel rather than for running in CI makes IMHO more sense

prow/test/integration/test/sinker_test.go
```diff
@@ -0,0 +1,183 @@
...
+		{
+			"orphaned-pod",
+			false,
+			true,
+		},
+	}
+
+	for _, tt := range tests {
+		tt := tt
+		name := tt.name
```
alvaroaleman, line 54: this isn't needed, since you already capture `tt`

prow/test/integration/test/setup.go
```diff
@@ -0,0 +1,103 @@
...
+	// If KUBECONFIG env var isn't set then look for $HOME/.kube/config
+	if defaultKubeconfig == "" {
+		if usr, err := user.Current(); err == nil {
+			defaultKubeconfig = path.Join(usr.HomeDir, ".kube/config")
+		}
+	}
+	return defaultKubeconfig
+}
+
+func NewClients(configPath, clusterName string) (*kubernetes.Clientset, *prow.Clientset, error) {
```
alvaroaleman, line 63: I can only recommend to use the [controller-runtime client](https://godoc.org/sigs.k8s.io/controller-runtime/pkg/client#Client) for this, as it is one client that allows you to interact with all object kinds

prow/test/integration/test/sinker_test.go
```diff
@@ -0,0 +1,183 @@
...
+				if err != nil {
+					return false, fmt.Errorf("Failed listing pods: %v", err)
+				}
+				return (p.Status.Phase == corev1.PodRunning), nil
+			}); err != nil {
+				t.Fatalf("Pod was not created successfully: %v", err)
+			}
+			t.Logf("Pod is running: %s", name)
+
+			// Make sure pod is deleted, it'll take roughly 2 minutes
```
alvaroaleman, line 151: This takes two minutes with a five second resync period, are you sure?

prow/test/integration/test/sinker_test.go
```diff
@@ -0,0 +1,183 @@
...
+			ctx := context.Background()
+
+			t.Cleanup(func() {
+				prowjobClient.ProwV1().ProwJobs("default").
+					Delete(ctx, name, v1.DeleteOptions{})
+				kubeClient.CoreV1().Pods(testpodNamespace).
+					Delete(ctx, name, v1.DeleteOptions{})
+			})
+
+			if tt.hasCRD {
```
alvaroaleman, line 81: err, please rename to `hasCR`, all tests have the CRD.

prow/test/integration/test/sinker_test.go
```diff
@@ -0,0 +1,183 @@
...
+				}
+				var exist bool
+				for _, p := range pods.Items {
+					if p.Name == name {
+						exist = true
+					}
+				}
+				return !exist, nil
+			})
+			pods, err := kubeClient.CoreV1().Pods(testpodNamespace).List(ctx, v1.ListOptions{})
```
alvaroaleman, line 167: Can we just capture the `exists` variable in this scope and avoid the second list, pod iteration etc and end the test right here?

alvaroaleman (commented):
Awesome to see some progress here :)

prow/test/integration/setup-prow.sh
chaodaiG: Good point. What do you think about craning this image to gcr?

prow/test/integration/prow/config.yaml
chaodaiG: done

prow/test/integration/prow/cluster/100_namespace.yaml
chaodaiG: That would also be what I prefer, since majority of the manifests are identical. But there will be slight difference in deployment config, such as the image path, and future github-endpoint for github related integration tests, as well as other services that need mock(s). These can be achieved though, by various different method, such as kustomize, but might need some maintenance. What do you think?

prow/test/integration/test/sinker_test.go
```diff
@@ -0,0 +1,178 @@
...
+		{
+			"running-pod",
+			true,
+			false,
+		},
+		{
+			"orphaned-pod",
+			false,
+			true,
+		},
```
cjwagner, line 51: We'll also want to test some more scenarios like the following:
1. completed, non-orphaned pods are deleted after the `terminatedPodTTL` expires.
1. pods not created by prow are not deleted.
1. prowjobs (not pods) are deleted after `maxProwJobAge` has passed.

I figure this PR is more an initial prototype for integration testing though so we don't need to add these just yet if you'd rather focus on just validating this integration testing pattern.

prow/test/integration/test/sinker_test.go
```diff
@@ -0,0 +1,178 @@
...
+				t.Logf("Finished creating CRD: %s", tt.name)
+			}
+			// Create pod
+			t.Logf("Creating pod: %s", tt.name)
+			if err := kubeClient.Create(ctx, &pod); err != nil {
+				t.Fatalf("Failed creating pod: %v", err)
+			}
+			t.Logf("Finished creating pod: %s", tt.name)
+
+			// Make sure pod is running
```
cjwagner, line 134: This races with sinker deleting the pod. To safely test the orphaned pod case, I'd expect a PJ to be created before the pod is created, wait for the pod to start, then delete the PJ to orphan the pod. That should prevent sinker from seeing an orphaned pod until after we've confirmed the pod was successfully created.

prow/test/integration/setup-prow.sh
cjwagner: We already publish an alpine image with Prow. I think that is suitable for this. [gcr.io/k8s-prow/alpine](https://gcr.io/k8s-prow/alpine)

prow/test/integration/prow/config.yaml
cjwagner, line 9: It would be handy to allow dumping the Prow component logs into an output dir (`$ARTIFACTS` in CI) so that we can more easily debug integration test failures.

cjwagner (commented):
Thanks for working on this Chao, very exciting!

prow/test/integration/test/setup.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	return kubeClient, nil
+}
+
+func BuildClientConfig(kubeConfigPath, clusterName string) (*rest.Config, error) {
+	overrides := clientcmd.ConfigOverrides{}
+	// Override the cluster name if provided.
+	if clusterName != "" {
+		overrides.Context.Cluster = clusterName
```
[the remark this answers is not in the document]
chaodaiG: Agreed that `clusterName` is actually a context, this just happens to also be named as clustername in `clientcmd.ConfigOverrides`. The documentation didn't clearly mention which one to use, but the code examples I have seen somewhere else uses `Context.Cluster`, I have stepped in debugging mode and confirmed that it worked.

prow/test/integration/test/sinker_test.go
```diff
@@ -0,0 +1,178 @@
...
+				kubeClient.Delete(ctx, &pod)
+			})
+
+			if tt.hasCR {
+				t.Logf("Creating CRD: %s", tt.name)
+
+				if err := kubeClient.Create(ctx, &prowjob); err != nil {
+					t.Fatalf("Failed creating prowjob: %v", err)
+				}
+				t.Logf("Finished creating CRD: %s", tt.name)
```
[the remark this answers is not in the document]
chaodaiG: updated

prow/test/integration/prow/config.yaml
alvaroaleman: It can also be done with client-go, but it might be easier with `kubectl`. We don't need to stream it though, we can just as some kind of a post step dump the log of all pods in a file that has the pod name or sth like that

prow/test/integration/test/setup.go
```diff
@@ -0,0 +1,89 @@
...
+	testpodNamespace = "test-pods"
+)
+
+var clusterContext = flag.String("cluster", "kind-kind-prow-integration", "The context of cluster to use for test")
+
+func getClusterContext() string {
+	return *clusterContext
+}
+
+func getDefaultKubeconfig(cfg string) string {
```
alvaroaleman, line 48: all of this is something the `clientcfg.ConfigLoader` already does with its default ruleset

prow/test/integration/test.sh
cjwagner: I still don't think this works? If I have bazel installed, but don't have the correct bazel version, this will download the correct version, but then continue to use the version I originally had installed.
What is the need for requiring such a specific bazel version?

prow/test/integration/test/setup.go
chaodaiG: Good to learn, done

prow/test/integration/prow/config.yaml
chaodaiG: done

prow/test/integration/test/setup.go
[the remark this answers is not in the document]
cjwagner: Doesn't this only work because they happen to be the same and the current-context is already set? I would expect this to work without this override. Overriding just the cluster name without overriding the user probably wouldn't work if it was actually making a change. Here are the godoc: https://godoc.org/k8s.io/client-go/tools/clientcmd#ConfigOverrides

prow/test/integration/prow/config.yaml
cjwagner: Yes I'd expect logs to be dumped with kubectl at the end if $ARTIFACTS is populated (or better yet use a CLI arg/flag).

chaodaiG: And have verified that dumped logs are under artifacts: https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/test-infra/20451/pull-test-infra-integration/1349801426494164992

prow/test/integration/test.sh
chaodaiG: This solves the problem of the test image has bazel installed but not at the version required, it felt to me that bazel is smart enough to figure out which version to use?

prow/test/integration/test/setup.go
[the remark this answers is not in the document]
chaodaiG: I see your point, also override currentcontext

chaodaiG:
@cjwagner , instead of using build tag for integration test, a test flag `--run-integration-test` was added for the test suite, if not provided the tests won't run.

prow/test/integration/test/sinker_test.go
chaodaiG: ack, done

chaodaiG:
@petr-muller , @alvaroaleman , @cjwagner , I believe I have addressed all comments, could you take another look?

alvaroaleman (approved):
/hold

It would be nice to enable reporting for the new presubmit though, so that it appears below PRs where ppl explicitly triggered it

prow/test/integration/test/setup.go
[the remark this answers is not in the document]
cjwagner: I'm pretty sure overwriting the `Context.Cluster` would be problematic if the values actually differed since the cluster needs to be associated with the correct user (`AuthInfo`). That being said I don't know if the values will ever differ in practice so this might be fine anyways.

cjwagner (approved):
/hold cancel

PR #20459, opened 2021-01-13
Allow jobs to skip_fetch_head and only fetch the base ref.

fejta:
Some multi-headed repos have no root HEAD ref, thus git fetch origin will fail (needs to specify a head ref).
Support these repos by allowing them to specify a skip_fetch_head on either the job base and/or extra_ref level.

### DIFF
--- a/prow/pod-utils/clone/clone.go
+++ b/prow/pod-utils/clone/clone.go
@@ -195,13 +195,23 @@ func (g *gitCtx) commandsForBaseRef(refs prowapi.Refs, gitUserName, gitUserEmail
 		commands = append(commands, g.gitCommand("config", "http.cookiefile", cookiePath))
 	}
 
-	if refs.CloneDepth > 0 {
-		commands = append(commands, g.gitFetch(g.repositoryURI, "--tags", "--prune", "--depth", strconv.Itoa(refs.CloneDepth)))
-		commands = append(commands, g.gitFetch("--depth", strconv.Itoa(refs.CloneDepth), g.repositoryURI, refs.BaseRef))
-	} else {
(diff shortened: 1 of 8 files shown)


cjwagner (approved):
This LGTM, but I'm not familiar with multi-headed repos so I can't validate that this addresses the core problem. Sounds like you've confirmed that avoiding an unqualified `fetch` resolves the problem though?
/hold

prow/pod-utils/clone/clone.go
```diff
@@ -195,13 +195,23 @@ func (g *gitCtx) commandsForBaseRef(refs prowapi.Refs, gitUserName, gitUserEmail
...
-		commands = append(commands, g.gitFetch("--depth", strconv.Itoa(refs.CloneDepth), g.repositoryURI, refs.BaseRef))
-	} else {
-		commands = append(commands, g.gitFetch(g.repositoryURI, "--tags", "--prune"))
-		commands = append(commands, g.gitFetch(g.repositoryURI, refs.BaseRef))
+	var depthArgs []string
+	if d := refs.CloneDepth; d > 0 {
+		depthArgs = append(depthArgs, "--depth", strconv.Itoa(d))
+	}
+
+	if !refs.SkipFetchHead {
```
alvaroaleman, line 203: Since we explicitly fetch the base and HEAD we care about anyways, can't we simply omit fetching the repos head altogether?

Also, terminology is confusing here, becaue the head this refers to is not e.G. the pull head but the repo head

fejta: I am anticipating this causing problems for repos like k/k which need refs to previous branch tags to exist during testing. Thoughts?

PR #20460, opened 2021-01-13
Fix bash script autobumper to use new branch based EnsurePR

mpherman2:
Changes were made in a pervious PR to use author and branch name to detect if a PR already exists instead of author and PR title. We did not realize the pr-creator tool was also used by the autobump.sh script. This fixes the autobumper script as well as the pr-creator tool's main function which was not updated along with the change to the EnsurePR method.


alvaroaleman (approved):
/lgtm
This is actually not something we import, as it is in a `package main` :)

/assign @chaodaiG 
I can not approve here

cjwagner (approved):

Issue #20377, opened 2021-01-05
Support retrieving a directory in Prow's GitHub client


rahulbharuka:
**What would you like to be added**:
Add support to retrieve a directory in Prow's GitHub client.

**Why is this needed**:
Prow's Github client currently does not support retrieving a directory's content. Supporting it would unlock new use cases.
e.g. Searching a pattern within same directory files.

https://github.com/kubernetes/test-infra/blob/b3a0c9272beaa59d7ee00ef992383783fa0b40b2/prow/github/client.go#L3000

rahulbharuka:
Closing the issue as the support is now added by PR #20461.

PR #20461, opened 2021-01-13
Support directory retrieval in Prow's Github client

rahulbharuka:


rahulbharuka:
implements issue #20377

prow/github/types.go
```diff
@@ -943,10 +943,13 @@ type DraftReviewComment struct {
...
 }
 
-// Content is some base64 encoded github file content
+// Content contains information about a github file/directory.
 type Content struct {
 	Content string `json:"content"`
 	SHA     string `json:"sha"`
+	Type    string `json:"type"`
+	Name    string `json:"name"`
+	Path    string `json:"path"`
```
petr-muller, line 952: I wonder if we should have different struct types for different responses. Looking at the different responses in https://docs.github.com/en/free-pro-team@latest/rest/reference/repos#get-repository-content--code-samples it seem that e.g. for directories the `Content` member would not exist, so code that wants to work with `Content`  struct would always need to check the `Type` member which is IMO not that great.

So maybe we should have a `DirectoryContent` struct type that `GetDirectory` would always return?

prow/github/client.go
```diff
@@ -4096,3 +4097,40 @@ func (c *client) GetApp() (*App, error) {
...
+//
+// See https://developer.github.com/v3/repos/contents/#get-contents
+func (c *client) GetDirectory(org, repo, dirpath, commit string) ([]Content, error) {
+	durationLogger := c.log("GetDirectory", org, repo, dirpath, commit)
+	defer durationLogger()
+
+	url := fmt.Sprintf("/repos/%s/%s/contents/%s", org, repo, dirpath)
+	if commit != "" {
+		url = fmt.Sprintf("%s?ref=%s", url, commit)
+	}
```
petr-muller, line 4112: I think this should at least escape the input (https://golang.org/pkg/net/url/#QueryEscape) but to me it looks that a query should be really handled by a lower layer - we'd be passing something that is no longer a "path" into a `path` member and rely on the implementation to handle this. 

So maybe it's time to add a `query` member into the `request`?

petr-muller (commented):
/ok-to-test

prow/github/types.go
rahulbharuka: Sure, that would be cleaner. I have added `DirectoryContent` struct type now.

prow/github/client.go
rahulbharuka: Thanks. I have escaped the query-string for now.

To handle query-string escaping in lower layer, we would need to support multiple query params. (example use-case: https://github.com/kubernetes/test-infra/blob/master/prow/github/client.go#L2974)

So, shall we take up this task separately ?

petr-muller: Ah, I missed this prior art...  I guess it's fine to just escape that for now.

prow/github/types.go
```diff
@@ -943,13 +943,10 @@ type DraftReviewComment struct {
 	Body     string `json:"body"`
 }
 
-// Content contains information about a github file/directory.
+// Content is some base64 encoded github file content
```
petr-muller: Can we link to the piece of doc that shows what GH response does this represent?

prow/github/types.go
```diff
@@ -1328,3 +1325,11 @@ type AppInstallationToken struct {
 	Permissions  InstallationPermissions `json:"permissions,omitempty"`
 	Repositories []Repo                  `json:"repositories,omitempty"`
 }
+
+// DirectoryContent contains information about a github directory.
+type DirectoryContent struct {
```
petr-muller, line 1330: Here too, can we link to the piece of doc that shows what GH response does this represent?

petr-muller (approved):
LGTM, please consider adding some links the GoDocs

petr-muller (approved):

spiffxp:
#20498 (comment)

/cc @alvaroaleman @cjwagner

alvaroaleman (approved):

cjwagner (approved):
This LGTM, but we should carefully consider every use of this new GH client function to evaluate if we should be using the `git` client to checkout the repo for inspection instead. Checking out the repo with `git` would be slower (for small numbers of directory requests at least), but it would be free in terms of GH API rate limit consumption.

PR #20467, opened 2021-01-13
vulndash: remove debug logs from the job

cpanato:
remove the debug logs from the job to reduce the log file

/assign @spiffxp


hasheddan (approved):
This seems fine to me given that debug logs were enabled when we were having timeout issues and that hasn't been an issue as of late. Thanks @cpanato!

/lgtm

cpanato:
@hasheddan we are discussing it in slack, if you want to follow here is the link: https://kubernetes.slack.com/archives/CJH2GBF7Y/p1605861516125600

PR #20468, opened 2021-01-13
allow specifying labels to prioritize PRs in merge pools

### LINKED ISSUE #18985: Prioritize PRs with certain labels in merge pool

Currently, of the pool of PRs that are eligible to merge, PR number is used to order merges so oldest eligible PRs merge first.

It would be useful to be able to give an ordered list of label queries that would be prioritized before falling back to PR number sort.

For example:
```yaml
priority:
# prioritize PRs fixing flaking/failing tests
- labels: ["kind/failing-test"]
- labels: ["area/deflake"]
# prioritize critical bugfixes
- labels: ["kind/bug", "priority/critical-urgent"]
```

matthyx:
fixes #18985

### DIFF
--- a/prow/config/tide.go
+++ b/prow/config/tide.go
@@ -73,6 +73,11 @@ type TideMergeCommitTemplate struct {
 	Body  *template.Template `json:"-"`
 }
 
+// TidePriority contains a list of labels used to prioritize PRs in the merge pool
+type TidePriority struct {
+	Labels []string `json:"labels,omitempty"`
+}
+
 // Tide is config for the tide pool.
 type Tide struct {
 	// SyncPeriod specifies how often Tide will sync jobs with GitHub. Defaults to 1m.
@@ -146,6 +151,10 @@ type Tide struct {
 	//  0 => unlimited batch size
 	// -1 => batch merging disabled :(
 	BatchSizeLimitMap map[string]int `json:"batch_size_limit,omitempty"`
+
+	// Priority is an ordered list of labels that would be prioritized before other PRs
+	// PRs should match all labels contained in a list to be prioritized
+	Priority []TidePriority `json:"priority,omitempty"`
 }
 
 func (t *Tide) BatchSizeLimit(repo OrgRepo) int {
--- a/prow/tide/tide.go
+++ b/prow/tide/tide.go
@@ -766,23 +766,43 @@ func unsuccessfulContexts(contexts []Context, cc contextChecker, log *logrus.Ent
 	return failed
 }
 
-func pickSmallestPassingNumber(log *logrus.Entry, ghc githubClient, prs []PullRequest, cc map[int]contextChecker) (bool, PullRequest) {
+func hasAllLabels(pr PullRequest, labels []string) bool {
+	if len(labels) == 0 {
+		return true
+	}
+	prLabels := sets.NewString()
+	for _, l := range pr.Labels.Nodes {
+		prLabels.Insert(string(l.Name))
+	}
+	requiredLabels := sets.NewString(labels...)
+	return prLabels.Intersection(requiredLabels).Equal(requiredLabels)
+}
+
+func pickHighestPriorityPR(log *logrus.Entry, ghc githubClient, prs []PullRequest, cc map[int]contextChecker, isPassingTestsFunc func(*logrus.Entry, githubClient, PullRequest, contextChecker) bool, priorities []config.TidePriority) (bool, PullRequest) {
 	smallestNumber := -1
 	var smallestPR PullRequest
-	for _, pr := range prs {
-		if smallestNumber != -1 && int(pr.Number) >= smallestNumber {
-			continue
+	for _, p := range append(priorities, config.TidePriority{}) {
+		for _, pr := range prs {
+			if !hasAllLabels(pr, p.Labels) {
+				continue
+			}
+			if smallestNumber != -1 && int(pr.Number) >= smallestNumber {
+				continue
+			}
+			if len(pr.Commits.Nodes) < 1 {
+				continue
+			}
+			if !isPassingTestsFunc(log, ghc, pr, cc[int(pr.Number)]) {
+				continue
+			}
+			smallestNumber = int(pr.Number)
+			smallestPR = pr
 		}
-		if len(pr.Commits.Nodes) < 1 {
-			continue
+		if smallestNumber > -1 {
+			return true, smallestPR
 		}
-		if !isPassingTests(log, ghc, pr, cc[int(pr.Number)]) {
-			continue
-		}
-		smallestNumber = int(pr.Number)
-		smallestPR = pr
 	}
-	return smallestNumber > -1, smallestPR
+	return false, smallestPR
 }
 
 // accumulateBatch looks at existing batch ProwJobs and, if applicable, returns:
@@ -1258,7 +1278,7 @@ func (c *Controller) takeAction(sp subpool, batchPending, successes, pendings, m
 	// Do not merge PRs while waiting for a batch to complete. We don't want to
 	// invalidate the old batch result.
 	if len(successes) > 0 && len(batchPending) == 0 {
-		if ok, pr := pickSmallestPassingNumber(sp.log, c.ghc, successes, sp.cc); ok {
+		if ok, pr := pickHighestPriorityPR(sp.log, c.ghc, successes, sp.cc, isPassingTests, c.config().Tide.Priority); ok {
 			return Merge, []PullRequest{pr}, c.mergePRs(sp, []PullRequest{pr})
 		}
 	}
@@ -1278,7 +1298,7 @@ func (c *Controller) takeAction(sp subpool, batchPending, successes, pendings, m
 	}
 	// If we have no serial jobs pending or successful, trigger one.
 	if len(missings) > 0 && len(pendings) == 0 && len(successes) == 0 {
-		if ok, pr := pickSmallestPassingNumber(sp.log, c.ghc, missings, sp.cc); ok {
+		if ok, pr := pickHighestPriorityPR(sp.log, c.ghc, missings, sp.cc, isPassingTests, c.config().Tide.Priority); ok {
 			return Trigger, []PullRequest{pr}, c.trigger(sp, missingSerialTests[int(pr.Number)], []PullRequest{pr})
 		}
 	}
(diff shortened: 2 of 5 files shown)


prow/config/tide.go
```diff
@@ -146,6 +151,9 @@ type Tide struct {
 	//  0 => unlimited batch size
 	// -1 => batch merging disabled :(
 	BatchSizeLimitMap map[string]int `json:"batch_size_limit,omitempty"`
+
+	// Priority
+	Priority []TidePriority `json:"priority,omitempty"`
```
alvaroaleman, line 156: If we ever want to add something like this, it needs a lot more documentation than this

prow/config/tide.go
```diff
@@ -146,6 +151,9 @@ type Tide struct {
 	//  0 => unlimited batch size
 	// -1 => batch merging disabled :(
 	BatchSizeLimitMap map[string]int `json:"batch_size_limit,omitempty"`
+
+	// Priority
```
liggitt, line 155: add a little more documentation here about how it is used?

prow/config/tide.go
```diff
@@ -73,6 +73,11 @@ type TideMergeCommitTemplate struct {
 	Body  *template.Template `json:"-"`
 }
 
+// TidePriority
+type TidePriority struct {
+	Labels []string `json:"labels,omitempty"`
```
liggitt, line 78: add a little more documentation here?

prow/tide/tide.go
```diff
@@ -1544,7 +1557,14 @@ func sortPools(pools []Pool) {
 	})
 
 	sortPRs := func(prs []PullRequest) {
-		sort.Slice(prs, func(i, j int) bool { return int(prs[i].Number) < int(prs[j].Number) })
+		sort.Slice(prs, func(i, j int) bool {
```
liggitt, line 1560: what happens to the merge queue if labels change in a way that modifies priority while a batch job is running? hoping that doesn't interrupt an already-running batch or merge job

prow/tide/tide.go
```diff
@@ -1532,7 +1532,20 @@ func prMeta(prs ...PullRequest) []prowapi.Pull {
 	return res
 }
 
-func sortPools(pools []Pool) {
+// TODO: should we memoize this?
+func getPRPriority(pr PullRequest, priorities []config.TidePriority) int {
+	for i, p := range priorities {
+		labels := sets.NewString(p.Labels...)
+		for _, l := range pr.Labels.Nodes {
+			if labels.Has(string(l.Name)) {
```
alvaroaleman, line 1540: Wouldn't it be more logical to require all labels in the priority setting?

prow/tide/tide.go
```diff
@@ -437,7 +437,7 @@ func (c *Controller) Sync() error {
 	for pool := range poolChan {
 		pools = append(pools, pool)
 	}
-	sortPools(pools)
+	sortPools(pools, c.config().Tide.Priority)
```
alvaroaleman, line 440: The only way in which the data we now sort differently is used is in serving tides status via http which is probably not want you want :)

alvaroaleman: I _think_ what you intended to change was `pickSmallestPassingNumber` which is used to determine which PR to merge if there are multiple candidates and which PR to serially retest, if there are multiple candidates

alvaroaleman (commented):
@liggitt just to be sure what you described in https://github.com/kubernetes/test-infra/issues/18985 is understood correctly: Was the idea to create some kind of "virtual" query for the priorities (i.E. as long as anything matches the priority labels on top of meeting a normal query, we are pretty much going to ignore everything else) or was it to priorize merging/serial retesting of PRs according to the prorities?

Right now this PR (I think) intends to do the latter, but we would for example still wait for a running batch before merging. Implementing the former is probably going to be a bit tricky.

/cc @cjwagner @stevekuznetsov
for comments on the idea

prow/tide/tide.go
liggitt: yes, the priority should only match if all of the labels in the priority are on the PR

liggitt:
> @liggitt just to be sure what you described in #18985 is understood correctly: Was the idea to create some kind of "virtual" query for the priorities (i.E. as long as anything matches the priority labels on top of meeting a normal query, we are pretty much going to ignore everything else) or was it to priorize merging/serial retesting of PRs according to the prorities?

The latter, to influence the sort order used to construct the merge pool with labels

> Right now this PR (I think) intends to do the latter, but we would for example still wait for a running batch before merging. 

That matches my expectations.

The main use case is:
1. merge pool has many PRs that meet requirements (e.g. 100 PRs with lgtm+approve+green CI)
2. a flake is found that is making the merge queue flake badly
3. a fix is made for that flake in a new PR, which is reviewed+approved+passes tests

Because the flake fix PR is new, it currently goes at the end of the merge pool. We would want to be able to label that PR to make it jump in line rather than having the queue limp through all the other ready PRs flaking/failing tests frequently.

prow/tide/tide.go
matthyx: this time on the right method...

matthyx:
@alvaroaleman if you have time pls

prow/tide/tide.go
```diff
@@ -766,23 +766,42 @@ func unsuccessfulContexts(contexts []Context, cc contextChecker, log *logrus.Ent
...
-func pickSmallestPassingNumber(log *logrus.Entry, ghc githubClient, prs []PullRequest, cc map[int]contextChecker) (bool, PullRequest) {
+func hasAllLabels(pr PullRequest, labels []string) bool {
+	if len(labels) == 0 {
+		return true
+	}
+	prLabels := sets.NewString()
+	for _, l := range pr.Labels.Nodes {
+		prLabels.Insert(string(l.Name))
+	}
+	return len(prLabels.Intersection(sets.NewString(labels...))) == len(labels)
```
alvaroaleman, line 777: Nit: Sets have an `Equal` you could use here

prow/tide/tide.go
```diff
@@ -766,23 +766,42 @@ func unsuccessfulContexts(contexts []Context, cc contextChecker, log *logrus.Ent
...
+}
+
+func pickSmallestPassingNumber(log *logrus.Entry, ghc githubClient, prs []PullRequest, cc map[int]contextChecker, isPassingTestsFunc func(*logrus.Entry, githubClient, PullRequest, contextChecker) bool, priorities []config.TidePriority) (bool, PullRequest) {
```
alvaroaleman, line 780: Could you rename this func to something like `pickHighestPriorityPR` or sth like that to account for the changed behavior?

prow/tide/tide.go
```diff
@@ -1258,7 +1278,7 @@ func (c *Controller) takeAction(sp subpool, batchPending, successes, pendings, m
 	// Do not merge PRs while waiting for a batch to complete. We don't want to
 	// invalidate the old batch result.
 	if len(successes) > 0 && len(batchPending) == 0 {
-		if ok, pr := pickSmallestPassingNumber(sp.log, c.ghc, successes, sp.cc); ok {
+		if ok, pr := pickHighestPriorityPR(sp.log, c.ghc, successes, sp.cc, isPassingTests, c.config().Tide.Priority); ok {
```
alvaroaleman, line 1281: Why the added isPassingTests? The fact that this is in `successes` implies that it has passing tests I think

prow/tide/tide.go
```diff
@@ -1278,7 +1298,7 @@ func (c *Controller) takeAction(sp subpool, batchPending, successes, pendings, m
 	}
 	// If we have no serial jobs pending or successful, trigger one.
 	if len(missings) > 0 && len(pendings) == 0 && len(successes) == 0 {
-		if ok, pr := pickSmallestPassingNumber(sp.log, c.ghc, missings, sp.cc); ok {
+		if ok, pr := pickHighestPriorityPR(sp.log, c.ghc, missings, sp.cc, isPassingTests, c.config().Tide.Priority); ok {
```
alvaroaleman, line 1301: Same here, `missings` means that is has missing up-to-date tests

prow/tide/tide_test.go
```diff
@@ -3942,3 +3951,74 @@ func TestDeduplicateContestsDoesntLoseData(t *testing.T) {
...
+				testPRWithLabels("org", "repo", "A", 9, githubql.MergeableStateMergeable, []string{"kind/failing-test"}),
+				testPRWithLabels("org", "repo", "A", 10, githubql.MergeableStateMergeable, []string{"kind/bug", "priority/critical-urgent"}),
+			},
+			expected: 9,
+		},
+	}
+	alwaysTrue := func(*logrus.Entry, githubClient, PullRequest, contextChecker) bool { return true }
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			_, got := pickHighestPriorityPR(nil, nil, tc.prs, nil, alwaysTrue, priorities)
```
matthyx, line 4018: @alvaroaleman for easier unit tests, I have moved the func `isPassingTests` to a parameter that I can mock more easily than the reall one

prow/tide/tide.go
matthyx: Same reason

prow/tide/tide.go
matthyx: Look at my comment https://github.com/kubernetes/test-infra/pull/20468/files#r559936185

alvaroaleman (approved):
Thanks a lot for your work on this!

matthyx:
> Thanks a lot for your work on this!

let's see how it helps 1.21 near code freeze

liggitt:
capability lgtm, I didn't review the implementation

prow/config/tide.go
```diff
@@ -146,6 +151,10 @@ type Tide struct {
 	//  0 => unlimited batch size
 	// -1 => batch merging disabled :(
 	BatchSizeLimitMap map[string]int `json:"batch_size_limit,omitempty"`
+
+	// Priority is an ordered list of labels that would be prioritized before other PRs
+	// PRs should match all labels contained in a list to be prioritized
```
cjwagner, line 156: nit: This doesn't seem quite accurate.
```suggestion
	// Priority is an ordered list of sets of labels that would be prioritized before other PRs
	// PRs should match all labels contained in a set to be prioritized
```

cjwagner (approved):
LGTM, just one nit that can be ignored.

I am wondering about changing labels on PRs causing churn that could slow the merge rate, but I don't think that is actually a concern. Tide should merge serially tested PRs that have up to date tests even if they are no longer the highest priority. In other words it looks like we only use priority to decide what to test next; higher priority PRs entering the queue should not prevent merging lower priority PRs that have been successfully retested and can be immediately merged.

PR #20469, opened 2021-01-13
Update KubeVirt TestGrid config.

fgimenez:
* Removed non-compatible jobs
* Added dashboard tab for periodics

### DIFF
--- a/config/testgrids/kubevirt/gen-config.yaml
+++ b/config/testgrids/kubevirt/gen-config.yaml
@@ -1,6 +1,7 @@
 dashboard_groups:
 - dashboard_names:
   - kubevirt-presubmits
+  - kubevirt-periodics
   name: kubevirt
 dashboards:
 - dashboard_tab:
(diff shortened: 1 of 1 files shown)


chases2:
Is the kubevirt repository using transfigure to copy the config over? If so, then a manual update PR will just be overwritten.

Also, I see that this directory used to have an [OWNERS file](https://github.com/kubernetes/test-infra/commit/4ddb62ffc060b3e143833284871d47b9bd6a2911), but it was removed. Was there an issue with it?

fgimenez:
> Is the kubevirt repository using transfigure to copy the config over? If so, then a manual update PR will just be overwritten.

Nope, so far we have created the PRs manually, will take a look at transfigure

> Also, I see that this directory used to have an [OWNERS file](https://github.com/kubernetes/test-infra/commit/4ddb62ffc060b3e143833284871d47b9bd6a2911), but it was removed. Was there an issue with it?

Yep, initially after looking at the testgrid configuration of other projects I added an OWNERS file but the bot complained bc none of the users on it was a k8s member. I'd love to start contributing but haven't seen any opportunities yet :)

chases2:
It looks like that repository is already using testgrid-* annotations, so I wasn't sure. [Transfigure](https://github.com/kubernetes/test-infra/blob/master/testgrid/cmd/transfigure/README.md) (and Configurator) read from those annotations.

In that case, this all looks good! Thanks!

/lgtm
/approve

Issue #17097, opened 2020-04-04
Fix config forker annotations in scalability jobs


wojtek-t:
Our templates were incorrectly using perf-tests branches.See https://github.com/kubernetes/test-infra/pull/17089 for more context

We need to double check our templates to confirm that this won't happen in the future again.

/assign @jkaniuk 

/cc @mm4tt @kubernetes/sig-scalability-bugs

jkaniuk:
We've decided to not automatically use perf-tests branches in presubmits, as creation of new perf-tests branch is a manual process and happens later than roll of kubernetes versions.

So presubmit flow right now is:
- k/k master jobs points to perf-tests/master

after forking:
- presubmit k/k/release-1.18 still point to perf-tests/master
- periodic on k/k/release-1.18 point to perf-tests/release-1.18 (triggering alerts)

manual work is needed:
- creation of new perf-tests branch to fix alerts
- when periodics are good - switching presubmits to release branch

jkaniuk:
That last step is easy to forget unfortunately.
We have an internal issue (b/146629137) with reminder to update perf-tests branches.
Any idea how to do that better?

wojtek-t:
Thanks for looking into it.

For now, I was simply manually cutting branches in github UI.
But maybe what we should do is to create a script that will:
- create branch
- file an issue (and assign it to whoever is running that script) to switch stuff in test-infra

It would make it much harder to forget about the second step.

jkaniuk:
Looks like pull-kubernetes-kubemark-e2e-gce-big template is still broken (and 1.20 presubmit as well):

https://github.com/kubernetes/test-infra/blob/a112810d6a51fff217455593e6d085c5ab1e2a0a/config/jobs/kubernetes/sig-release/release-branch-jobs/1.20.yaml#L1309

jkaniuk:
@marseel as an oncall

jkaniuk:
They are definitely not good.

1.20 was fixed manually, but 1.21 is wrong

https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/sig-release/release-branch-jobs/1.21.yaml
```
   - base_ref: master
      org: kubernetes
      path_alias: k8s.io/perf-tests
      repo: perf-tests
```

I'm not sure how config forker works now, after switching to podutils.

We have `--repo=k8s.io/perf-tests=master -> --repo=k8s.io/perf-tests=release-{{.Version}}`
[...]

PR #20476, opened 2021-01-14
Fix performance presubmit test job to release-1.20 branch of k/k

tosi3k:
`pull-kubernetes-e2e-gce-100-performance` presubmit to `release-1.20` branch was failing as it was using `master` branch of perf-tests (instead of `release-1.20`) which in turn recently ingested an incompatible change in https://github.com/kubernetes/perf-tests/pull/1647.

This should fix https://github.com/kubernetes/kubernetes/issues/98039.

/sig scalability
/assign @wojtek-t


wojtek-t:
/lgtm
/approve

Thanks

@tosi3k - s a follow up, can we fix the template, so that we won't hit the same problem in 1.21?

cpanato (approved):
/lgtm

jprzychodzen:
It seems that this is related to #17097. We hit this almost every release.

PR #20478, opened 2021-01-14
Deprecate 'preset-e2e-scalability-containerd' as containerd is defauft CRI

jprzychodzen:
We cannot delete preset, as there are existing tests for 1.19 and 1.20 release that depend on this preset

`preset-e2e-scalability-containerd` set following envs:

1. Setting runtime to containerd (already a default)
2. Setting timeout for perf-test probes to 10 min (we want to remove that)
3. Dumping containerd logs (we should do that for all tests)

/cc @wojtek-t 

Result from [this comment](https://github.com/kubernetes/kubernetes/issues/94256#issuecomment-753912951)

### DIFF
--- a/config/jobs/kubernetes/sig-scalability/sig-scalability-presets.yaml
+++ b/config/jobs/kubernetes/sig-scalability/sig-scalability-presets.yaml
@@ -226,17 +226,18 @@ presets:
     value: 50
   - name: PERF_TESTS_PRINT_COMMIT_HISTORY
     value: true
+  - name: LOG_DUMP_SAVE_SERVICES
+    value: "containerd"
 
   ###### Scalability Envs
   ### Common env variables for containerd scalability-related suites.
+  # Deprecated. Should be removed after 1.20 branch will be removed from testing
 - labels:
     preset-e2e-scalability-containerd: "true"
   env:
   - name: CL2_PROBE_MEASUREMENTS_CHECK_PROBES_READY_TIMEOUT
     value: "10m"
   - name: KUBE_CONTAINER_RUNTIME
-    value: "containerd"
-  - name: LOG_DUMP_SAVE_SERVICES
     value: "containerd"
 
 ###### Scalability Envs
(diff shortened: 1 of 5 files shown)


config/jobs/kubernetes/sig-scalability/sig-scalability-presets.yaml
```diff
@@ -226,18 +226,19 @@ presets:
     value: 50
   - name: PERF_TESTS_PRINT_COMMIT_HISTORY
     value: true
+  - name: LOG_DUMP_SAVE_SERVICES
+    value: "containerd"
 
   ###### Scalability Envs
   ### Common env variables for containerd scalability-related suites.
+  # Deprecated. Should be removed after 1.20 branch will be removed from testing
```
wojtek-t, line 234: We cannot remove the preset, but we can remove the CL2_PROBE_MEASUREMENTS_CHECK_PROBES_READY_TIMEOUT, right?

jprzychodzen: I do not think so? I guess that mitigation `net.netfilter.nf_conntrack_tcp_be_liberal=1` is not backported to branches 1.19 and 1.20. Is this correct?

wojtek-t: Let's set it only in periodics and presubmits then maybe:
https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/sig-scalability/sig-scalability-presets.yaml#L294

I would really like to avoid having that in master jobs.

jprzychodzen: In this PR I'm removing `preset-e2e-scalability-containerd: "true"` from all master jobs, effectively removing CL2_PROBE_MEASUREMENTS_CHECK_PROBES_READY_TIMEOUT from all master jobs.

I do not think that we need CL2_PROBE_MEASUREMENTS_CHECK_PROBES_READY_TIMEOUT in presubmits/periodics. Could you elaborate on that?

wojtek-t: OK - you're right, it's only in this preset.

So yeah - I'm fine with this PR.

wojtek-t:
/lgtm
/approve

/hold
@jprzychodzen - Let's merge on Monday morning.

jprzychodzen:
It's Monday morning, merging.
/unhold

PR #20482, opened 2021-01-14
prow: allow configuring the filenames used for OWNERS(_ALIASES)

stevekuznetsov:
When a repository is a fork of an upstream repository with OWNERS, or
vendors a repository that also uses OWNERS, it is generally not intended
for users who are OWNERS upstream to have rights in the downstream.
Furthermore, changing the OWNERS files leads to merge conflicts as
changes from upstream are pulled in, which leads to headaches. Allowing
users to configure the files used by Prow to denote owners and aliases
side-steps this issue.


stevekuznetsov:
Seventeen tries later I think `bazel` is finally happy ...

cjwagner:
The `owners_dir_blacklist` configuration field is designed to address this issue: https://github.com/kubernetes/test-infra/blob/351ae39320b208108582b59e193bac4fb7fcc190/prow/config/config.go#L143-L145

Can that satisfy your use case? I'd prefer to avoid making the OWNERS file names configurable. I'd rather have consistency across repos so that repo admins can choose whether OWNERS files sourced from other repos should apply (and even allow some to apply while ignoring others).

stevekuznetsov:
@cjwagner no, in the case of a fork that's maintained actively, it's required that the entire codebase be changed (not only vendor) and that sub-directories, etc, in the codebase have reviewers and approvers as needed.

prow/plugins/config.go
```diff
@@ -167,6 +168,22 @@ type Owners struct {
 	// OWNERS file, preventing their automatic addition by the owners-label plugin.
 	// This check is performed by the verify-owners plugin.
 	LabelsBlackList []string `json:"labels_blacklist,omitempty"`
+
+	// Filenames allows configuring repos to use a separate set of filenames for
+	// any plugin that interacts with these files.
+	Filenames map[string]ownersconfig.Filenames `json:"filenames,omitempty"`
```
alvaroaleman, line 174: Could you mention in the godoc that the key is in `org/repo` format?

prow/plugins/ownersconfig/config.go
```diff
@@ -0,0 +1,42 @@
...
+
+type Resolver func(org, repo string) Filenames
+
+// FakeResolver fills in for tests that use a resolver but aren't testing it.
+func FakeResolver(_, _ string) Filenames {
+	return FakeFilenames
+}
+
+// FakeFilenames fills in for tests that need a Filenames but aren't testing them.
+var FakeFilenames = Filenames{
```
alvaroaleman, line 39: Can we rename this to `DefaultFileNames`? The current naming is a bit confusing, as it makes for example tests look like they use a non-default set of names

stevekuznetsov: I explicitly didn't name it that way since I do *not* want production code to use these, they exist explicitly for testcases.

stevekuznetsov: I can make that more explicit in the doc.

alvaroaleman: Sure, but that doesn't change anything about the fact that they are the default and not a `Fake`

stevekuznetsov:
@alvaroaleman updated YAML

stevekuznetsov:
@cjwagner @alvaroaleman PTAL

alvaroaleman (approved):

PR #20483, opened 2021-01-14
Update kube dependencies to 1.20.2

alvaroaleman:


fejta (commented):
Are you sure we still need all these replace statements? Looks like many of these are replaced to their current values? (AKA k8s.io/api v0.20.2 is replaced with... v0.20.2)

alvaroaleman:
@fejta good point, I removed all that did a no-op replacement, lets see if tests pass

alvaroaleman:
@fejta all tests green, seems everything works just fine without those :)

fejta (approved):

PR #20487, opened 2021-01-15
Add chaodaiG to OWNERS of a few prow components

chaodaiG:
So that code reviews can be redirected to chaodaiG, also by the way added reviewers and approvers sections under `prow/cmd` for each components.

/assign @fejta @cjwagner @michelle192837 @e-blackwelder

### DIFF
--- a/prow/cmd/crier/OWNERS
+++ b/prow/cmd/crier/OWNERS
@@ -1,5 +1,10 @@
 # See the OWNERS docs at https://go.k8s.io/owners
-
+reviewers:
+- chaodaiG
+- fejta
+approvers:
+- chaodaiG
+- fejta
 emeritus_approvers:
 - krzyzacy
 labels:
--- a/prow/cmd/crier/OWNERS
+++ b/prow/cmd/crier/OWNERS
@@ -1,9 +1,10 @@
 # See the OWNERS docs at https://go.k8s.io/owners
 reviewers:
 - chaodaiG
+- cjwagner
 - fejta
 approvers:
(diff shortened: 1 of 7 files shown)


chaodaiG:
@cjwagner , I have removed you from crier, let me know if you still want to be on the list

cjwagner (approved):

petr-muller:
Hm, won't this lead to a much smaller set of people asked for reviews? Previously review requests were quite reasonably spread between prow root reviewers, this now got much more limited.

cjwagner:
@petr-muller Thats a good point. I think we'd still like this granularity so that we can route reviews for specific components to the individuals that own (or want to gain experience with) those components rather than just having the broader set of Prow OWNERS.
Does that sound reasonable? If so I think we can address your concern by distributing users currently in Prow's root OWNERS file to more granular OWNERS files.

petr-muller:
I don't have a really strong opinion. I think granularity is good to make sure the assigned reviewers are able to review an actually do care, but with the amount of subcomponents that Prow has, maintaining this many OWNERS sounds like a lot of work and the files will get stale. I think that Prow has mostly general components where most Prow root reviewers care, and just few special ones (gerrit, spyglass) that benefit from having specialized reviewers (not everyone speaks typescript). On the other hand, it does not welcome new contributors that want to learn by starting doing reviews for a component or two, which sounds like a good argument for granular ownership.

So I guess distributing root OWNERS to the granular ones sounds like a good solution.

PR #20490, opened 2021-01-15
Switch ruamel.yaml.load() to safe_load()

sfowl:


stevekuznetsov (approved):

chases2:
/ok-to-test

It looks like you'll need a small indent change to make the linter happy. Other than that, looks good!

I did a little reading into the difference between load() and safe_load(), but could only find a [doc](https://yaml.readthedocs.io/en/latest/basicuse.html) which says that it "load[s] the doucment without resolving unknown tags". I think by tags it means [this metadata](https://stackoverflow.com/questions/15233335/i-dont-understand-what-a-yaml-tag-is), but is that all that "safe" does?

sfowl:
> /ok-to-test
> 
> It looks like you'll need a small indent change to make the linter happy. Other than that, looks good!

Thanks. Fixing that revealed that `safe_load()` doesn't support the same params so I change the Loader to SafeLoader instead, which has the same result.

> I did a little reading into the difference between load() and safe_load(), but could only find a [doc](https://yaml.readthedocs.io/en/latest/basicuse.html) which says that it "load[s] the doucment without resolving unknown tags". I think by tags it means [this metadata](https://stackoverflow.com/questions/15233335/i-dont-understand-what-a-yaml-tag-is), but is that all that "safe" does?

safe_load()/SafeLoader is also a hardening measure against crafted yaml that can lead to code execution, e.g.

https://nvd.nist.gov/vuln/detail/CVE-2019-20478

spiffxp (approved):
/lgtm

PR #20493, opened 2021-01-15
vulndash: increase page size to get the vulnerabilities and increase the resource spec

cpanato:
this is to try to debug why it is timing out because we don't have
enough permissions to test it locally

while we wait to see if we can get some credentials to try locally

/assign @spiffxp @hasheddan


hasheddan (approved):
@cpanato I am up for giving this a shot, but it may be worth also investigating if we can parallelize some of the listing. I am not super familiar with grafeas api and how it handles pagination, so this could be more challenging than I am expecting.

/lgtm

cpanato:
@hasheddan yes, I'm checking here on how to do that

PR #20498, opened 2021-01-15
Prow integration test job report to github

chaodaiG:
The integration test introduced in https://github.com/kubernetes/test-infra/pull/20451 works as expected, make it reporting to github before make it required for presubmit

### DIFF
--- a/config/jobs/kubernetes/test-infra/test-infra-presubmits.yaml
+++ b/config/jobs/kubernetes/test-infra/test-infra-presubmits.yaml
@@ -102,7 +102,6 @@ presubmits:
     - master
     always_run: false
     decorate: true
-    skip_report: true
     optional: true
     labels:
       preset-service-account: "true"


config/jobs/kubernetes/test-infra/test-infra-presubmits.yaml
```diff
@@ -102,7 +102,6 @@ presubmits:
     - master
     always_run: false
     decorate: true
-    skip_report: true
```
alvaroaleman, line 105: skip_report: true is completely independent from the `optional` setting btw

alvaroaleman (approved):

spiffxp:
FYI: this caused a couple of PR's to get stuck.  They passed their tests, then this context was made required, then the PR was approved... but the new context was never automatically triggered.  I'm manually triggering now.

How could we have added this in a less disruptive way?

alvaroaleman:
@spiffxp by finally deploying the status reconciler: https://github.com/kubernetes/test-infra/blob/86f41aa3149f107c8ac5872b14c0d80881ce52f3/prow/cmd/status-reconciler/README.md#L1

It is in use by Openshift for O(years) and we are not aware of any issues

spiffxp:
It's present at https://github.com/kubernetes/test-infra/blob/master/config/prow/cluster/statusreconciler_deployment.yaml

And appears to be deployed
```
$ k --context=gke_k8s-prow_us-central1-f_prow get pods -l app=statusreconciler
NAME                                READY   STATUS    RESTARTS   AGE
statusreconciler-84c7757876-wtc85   1/1     Running   0          121m
```

It's documented as handling this case
>a new blocking presubmit exists and should be triggered for every trusted pull request in flight

But it appears it didn't?

alvaroaleman:
@spiffxp actually the change that should have triggered that is https://github.com/kubernetes/test-infra/pull/20501
If I got the timings right, it is the first bunch of results in https://cloudlogging.app.goo.gl/712ifCvvwd4hVUYK6

It just shows that it didn't do anything :upside_down_face:

chaodaiG:
Not 100% sure, but have a hypothesis. I think statusreconciler figures out a new required prowjob by looking up previously stored data in memory, which tends to be lost when the pod is restarted, for example autobump. The PR makes this job required https://github.com/kubernetes/test-infra/pull/20501 was merged almost right before prow bump https://github.com/kubernetes/test-infra/pull/20497, so this could be the reason why statusreconciler didn't figure out this new job

alvaroaleman:
That would indeed explain it. I've created a PR to include the config git revision in the config itself, so the statusreconciler can add that to its log, which would allow us to actually debug a situations like this, right now its just guessing: #20527

PR #20500, opened 2021-01-15
Clean up and update the prod flake query to support PodUtil

MushuEE:
@spiffxp I used another select to simplify.
Tested against BQ and everything looks correct.

This CL is a final update to the flake-config that will not include PodUtil + Bootstrap jobs. There should no longer be any need to delay bootstrap change over
/area metrics


spiffxp (approved):
/approve
/lgtm

Manually ran query, agree this looks good

I would announce this to kubernetes-dev@ and cc kubernetes-sig-testing@ and kubernetes-sig-release@.  A heads up that it's happening with link to example of new/experimental results, and followup on Monday.

PR #20505, opened 2021-01-15
Move prow version prometheus metrics out of prowjob metrics, more prow components report this metric

chaodaiG:
Previously the version metrics was mixed with prowjob metrics, which is not ideal for other prow components to use it. Moving it out to version package so that it's easier to be used by others


chaodaiG:
/hold
Don't merge before weekend, to avoid alerts in case the metrics is missing after this PR

petr-muller (approved):

chaodaiG:
/unhold

FYI @michelle192837 , if prow image stale alert fired abnormally, this might be related

PR #20506, opened 2021-01-16
Update autobump description to correctly handle commits past a tag.

fejta:
Testgrid tags tend to look like v20210115-v0.0.38-14-g4c88977
This is standard git describe --tags output, so handle it correctly.

/assign @mpherman2


e-blackwelder (approved):

fejta:
> /lgtm trying again now that I have joined k8s

a) Just use github approve on this repo
b) `/lgtm` needs to be on its own line `/lgtm my additional thoughts` is not a command

PR #20514, opened 2021-01-18
Allow bugzilla plugin to match multiple target releases

jarrpa:
A bit of a naive implementation, but it might work.

This is to facilitate the use case where the target release for a given
branch may change over time but follow a specific pattern (e.g. "1.0.*")

Signed-off-by: Jose A. Rivera <jarrpa@redhat.com>


alvaroaleman (commented):
@jarrpa can you add a test, please?
/ok-to-test

jarrpa:
@alvaroaleman Done. I wasn't able to run it locally right away, since it seems I need go 1.15, but the tests appear to pass on the PR.

stevekuznetsov (changes requested):
I think we need a new field, and I would propose that it's a list of items, not a regex. Changing the interpretation of the existing field will be surprising to users and we should make this opt-in.

jarrpa:
> I think we need a new field, and I would propose that it's a list of items, not a regex. Changing the interpretation of the existing field will be surprising to users and we should make this opt-in.

Sounds fair, I wasn't sure how far to take the implementation. I can add on to the APi for sure.

PR #20515, opened 2021-01-18
Disable branch protection on gh-pages for cloud-provider-openstack

brtknr:
We need to disable branch protection on gh-pages branch for chart-releaser-action to function correctly.


lingxiankong:
/lgtm

Thanks!

PR #20521, opened 2021-01-19
Prow integration test: test crier with fake github server

chaodaiG:
This PR adds integration test for testing crier reporting on github context. This is a seed for demonstrating how to use a fake client to test prow.

- Add a fakeghserver component listening for `http://fakeghserver` in cluster.
- Crier uses this fakeghserver for all it's github reporting in integration test.
- Add an ingress for test runner to interact with fakeghserver for validating changes done by prow.


chaodaiG:
/uncc @chases2 @matthyx 
Let me fix the test first

chaodaiG:
Test passed. 

prow/test/integration/prow/cluster/crier.yaml
```diff
@@ -0,0 +1,137 @@
+---
+kind: ServiceAccount
+apiVersion: v1
+metadata:
+  annotations:
+    iam.gke.io/gcp-service-account: control-plane@k8s-prow.iam.gserviceaccount.com
```
cjwagner, line 6: Let's remove this, it's misleading.

prow/test/integration/test/crier_test.go
```diff
@@ -0,0 +1,187 @@
...
+						"prow.k8s.io/job":       podName,
+						"prow.k8s.io/refs.org":  "fake-org",
+						"prow.k8s.io/refs.pull": "0",
+						"prow.k8s.io/refs.repo": "fake-repo",
+						"prow.k8s.io/type":      "presubmit",
+					},
+					Labels: map[string]string{
+						"created-by-prow":  "true",
+						"prow.k8s.io/type": "presubmit",
+						"name":             podName,
```
cjwagner, line 101: nit: This isn't a label that we have.

prow/test/integration/test/crier_test.go
```diff
@@ -0,0 +1,187 @@
...
+				kubeClient.Delete(ctx, &prowjob)
+			})
+
+			t.Logf("Creating prowjob: %s", podName)
+			if err := kubeClient.Create(ctx, &prowjob); err != nil {
+				t.Fatalf("Failed creating prowjob: %v", err)
+			}
+			t.Logf("Finished creating prowjob: %s", podName)
+
+			var waitStatus func(*testing.T, string)
```
cjwagner, line 140: nit: Not needed.

prow/test/integration/fakeghserver/fakeghserver.go
```diff
@@ -0,0 +1,219 @@
...
+	if err := o.validate(); err != nil {
+		logrus.WithError(err).Fatal("Invalid arguments.")
+	}
+	defer interrupts.WaitForGracefulShutdown()
+	ghClient := fakegithub.NewFakeClient()
+
+	mux := http.NewServeMux()
+	// setup common handlers for local and deployed runs
+	mux.Handle("/", reposHandler(ghClient))
+	mux.Handle("/fakeghserver/", http.StripPrefix("/fakeghserver", reposHandler(ghClient)))
```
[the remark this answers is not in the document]
cjwagner: Ah I see now. This is just to expose the same interface cluster-externally without using the root path externally (since that will be used by Deck later).

Another option would be to deploy testing services like `fakeghserver` to their own namespace with their own ingress on a different port. This would separate http routing for the testing services from the components under test and let us use a more standard ingress for the Prow components. (This wouldn't work with GKE ingresses, but should with ingress-nginx here.)

We will never need to serve a `/fakeghserver` path from Deck so the current solution should be fine, just food for thought.

prow/test/integration/test/crier_test.go
```diff
@@ -0,0 +1,187 @@
...
+			waitStatus = func(t *testing.T, s string) {
+				if err := wait.Poll(200*time.Microsecond, 10*time.Second, func() (bool, error) {
+					e := fmt.Sprintf("http://localhost/fakeghserver/repos/fake-org/fake-repo/statuses/%s", sha)
+					resp, err := http.Get(e)
```
cjwagner, line 144: Can we instead use our real GH client (pointed at `http://localhost/fakeghserver`) for this?

prow/test/integration/fakeghserver/fakeghserver.go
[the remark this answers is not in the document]
alvaroaleman: just standard name based virtual hosting should also work, no need for multiple ports

prow/test/integration/fakeghserver/fakeghserver.go
```diff
@@ -0,0 +1,219 @@
...
+	var ok int
+	// By default ok is 200
+	ok = http.StatusOK
+	var err error
+	var msg string
+	parts := strings.Split(r.URL.Path, "/")
+	if strings.TrimSpace(parts[0]) == "" {
+		parts = parts[1:]
+	}
+	switch {
```
alvaroaleman, line 119: can we just use a http.ServeMux for the routing rather than implementing it with 200 lines of nested switch statements that will only grow bigger over time?

prow/test/integration/fakeghserver/fakeghserver.go
```diff
@@ -0,0 +1,219 @@
...
+			msg = string(content)
+		}
+	case parts[0] == "repos":
+		switch {
+		case len(parts) == 4 && parts[3] == "issues":
+			// Create status is 201
+			ok = http.StatusCreated
+			data := github.Issue{}
+			if err := unmarshal(w, r, &data); err != nil {
+				return fmt.Errorf("failed processing payload: %v", err)
```
alvaroaleman, line 136: Why do we sometimes return and sometimes set some variables that then triggers behavior at the end of the function?

prow/test/integration/fakeghserver/fakeghserver.go
alvaroaleman: (if http.ServeMux lacks features, please look for another pre-defined router. Go is famous for having a not so powerfull router in the stdlib and plenthoa of OSS ones to make up for that)

prow/test/integration/test/crier_test.go
```diff
@@ -0,0 +1,187 @@
...
+					},
+					Report: true,
+				},
+				Status: prowjobv1.ProwJobStatus{
+					State: prowjobv1.TriggeredState,
+				},
+			}
+
+			t.Cleanup(func() {
+				kubeClient.Delete(ctx, &prowjob)
```
alvaroaleman, line 131: Can we at least log the error?

prow/test/integration/test/crier_test.go
```diff
@@ -0,0 +1,187 @@
...
+								Author: "fake_author",
+								Number: 0,
+								SHA:    sha,
+							},
+						},
+					},
+					Report: true,
+				},
+				Status: prowjobv1.ProwJobStatus{
+					State: prowjobv1.TriggeredState,
```
alvaroaleman, line 126: Instead of implicitly testing the `Triggered` state in every job, why don't we do that explicitly with a testcase?

prow/test/integration/test/crier_test.go
```diff
@@ -0,0 +1,187 @@
...
+				if s == prowjobv1.SuccessState || s == prowjobv1.FailureState {
+					changes = append(changes, fmt.Sprintf("\"completionTime\": \"%s\"", time.Now().Format(time.RFC3339)))
+				}
+				patch := []byte(fmt.Sprintf("{\"status\": {%s}}", strings.Join(changes, ", ")))
+				if err := kubeClient.Patch(ctx, &prowjob, client.RawPatch(types.MergePatchType, patch)); err != nil {
+					t.Fatalf("Failed patching prowjob: %v", err)
+				}
+			}
+
+			for i, s := range tt.states {
```
alvaroaleman, line 181: I am a bit confused by this:
* Why don't we just create the job with the state we want to test and then check the reporting?
* Why do allow more than one state per testcase?

chaodaiG: Updated to do just one state per testcase

prow/test/integration/test/crier_test.go
chaodaiG: Done

prow/test/integration/test/crier_test.go
chaodaiG: Removed

prow/test/integration/test/crier_test.go
chaodaiG: Done

prow/test/integration/test/crier_test.go
chaodaiG: Updated to do just one state per testcase

prow/test/integration/test/crier_test.go
chaodaiG: Removed

prow/test/integration/fakeghserver/fakeghserver.go
chaodaiG: This was completely rewritten

prow/test/integration/fakeghserver/fakeghserver.go
chaodaiG: Updated to use routing from gorilla.mux

prow/test/integration/fakeghserver/fakeghserver.go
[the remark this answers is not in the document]
chaodaiG: Updated to use gorilla.mux to make this easier

chaodaiG:
@alvaroaleman @cjwagner , updated based on code review comments, please take another look. 

Fun fact: spent some time debugging a test failure caused by prowjob deleted by sinker before reporting, which I guess is good and exactly what's expected from writing integration tests.

chaodaiG:
The timeout error seems legit, I think some of the unit tests are not very happy with the mutex I added in fakegithub Client, still looking into why this had happened

prow/github/client.go
```diff
@@ -2061,7 +2061,7 @@ func (c *client) ListReviews(org, repo string, number int) ([]Review, error) {
 func (c *client) CreateStatus(org, repo, SHA string, s Status) error {
 	durationLogger := c.log("CreateStatus", org, repo, SHA, s)
 	defer durationLogger()
-
+	logrus.Infof("Creating status: %s", fmt.Sprintf("/repos/%s/%s/statuses/%s", org, repo, SHA))
```
alvaroaleman, line 2064: Remainder? The duration logger should include this info

prow/github/fakegithub/fakegithub.go
```diff
@@ -264,15 +326,48 @@ func (f *FakeClient) GetIssue(owner, repo string, number int) (*github.Issue, er
...
+	new := &github.Issue{
+		ID:        f.IssueID,
+		Title:     title,
+		Body:      body,
+		Milestone: github.Milestone{Number: milestone},
+		Labels:    ls,
+		Assignees: as,
+	}
+	f.Issues[f.IssueID] = new
+	f.IssueComments[f.IssueID] = make([]github.IssueComment, 0)
```
alvaroaleman, line 363: nit: Slices do not need initialization

prow/plugins/blockade/blockade_test.go
```diff
@@ -296,13 +296,12 @@ func TestHandle(t *testing.T) {
...
-		fakeClient := &fakegithub.FakeClient{
-			RepoLabelsExisting: []string{labels.BlockedPaths, otherLabel},
-			IssueComments:      make(map[int][]github.IssueComment),
-			PullRequestChanges: make(map[int][]github.PullRequestChange),
-			IssueLabelsAdded:   []string{},
-			IssueLabelsRemoved: []string{},
-		}
+		fakeClient := fakegithub.NewFakeClient()
+		fakeClient.RepoLabelsExisting = []string{labels.BlockedPaths, otherLabel}
+		fakeClient.IssueComments = make(map[int][]github.IssueComment)
```
alvaroaleman, line 301: Nit: The two maps now get initialized in the constructor and the slices do not need it

prow/plugins/bugzilla/bugzilla_test.go
```diff
@@ -689,13 +689,12 @@ Instructions for interacting with me using PR comments are available [here](http
...
-				PullRequests: map[int]*github.PullRequest{
-					1: {Base: github.PullRequestBranch{Ref: "branch"}, Title: testCase.title, Merged: testCase.merged},
-				},
-				IssueComments: map[int][]github.IssueComment{},
+			client := fakegithub.NewFakeClient()
+			client.PullRequests = map[int]*github.PullRequest{
+				1: {Base: github.PullRequestBranch{Ref: "branch"}, Title: testCase.title, Merged: testCase.merged},
 			}
-			event, err := digestComment(&client, logrus.WithField("testCase", testCase.name), testCase.e)
+			client.IssueComments = map[int][]github.IssueComment{}
```
alvaroaleman, line 696: Nit: Not needed

prow/test/integration/fakeghserver/fakeghserver.go
```diff
@@ -0,0 +1,258 @@
...
+
+import (
+	"encoding/json"
+	"flag"
+	"fmt"
+	"net/http"
+	"strconv"
+	"time"
+
+	"github.com/google/go-github/github"
```
alvaroaleman, line 28: Why are we using a different github lib here?

prow/test/integration/fakeghserver/fakeghserver.go
```diff
@@ -0,0 +1,258 @@
...
+	w.Header().Set("Content-Type", "application/json")
+	err := d.Decode(&data)
+	if err != nil {
+		w.WriteHeader(http.StatusInternalServerError)
+		fmt.Fprintf(w, "{\"error\": \"Failed unmarshal request: %v\"}", err.Error())
+	}
+	return err
+}
+
+func response(w http.ResponseWriter, r *http.Request, msg string, ok int, err error) error {
```
alvaroaleman, line 104: Maybe rename the `ok` arg to `statusCode`?

prow/test/integration/fakeghserver/fakeghserver.go
```diff
@@ -0,0 +1,258 @@
...
+		var ok int
+		var err error
+		ok = http.StatusOK
+		vars := mux.Vars(r)
+		org, repo, SHA := vars["org"], vars["repo"], vars["sha"]
+		if r.Method == http.MethodPost {
+			ok = http.StatusCreated
+			data := prowgh.Status{}
+			err = unmarshal(w, r, &data)
+			if err == nil {
```
alvaroaleman, line 163: In order to avoid the weird `if err == nil { err = doNextThing() }` nesting, it might be worth to implement `response` as a wrapper that takes the actual handler as a `func(r *http.Request) (statusCode int, err error)`

prow/test/integration/test/crier_test.go
```diff
@@ -0,0 +1,153 @@
...
+			if err := wait.Poll(200*time.Millisecond, 5*time.Minute, func() (bool, error) {
+				ss, err := githubClient.GetCombinedStatus("fake-org", "fake-repo", sha)
+				if err != nil {
+					return false, fmt.Errorf("failed listing statues: %w", err)
+				}
+				if want, got := 1, len(ss.Statuses); want != got {
+					// Wait until it's ready
+					return false, nil
+				}
+				return tt.want == ss.Statuses[0].State, nil
```
alvaroaleman, line 147: Might be worth to also verify that the prowjobs `status.prev_report_states.github-reporter` gets set

alvaroaleman (commented):
There is a bigger bunch of now uneeded map initializations respective slice initialiations that were never needed. Thanks god for GitHubs `mark file as viewed` feature.

The server looks much nicer with the mux now :)

prow/test/integration/fakeghserver/fakeghserver.go
chaodaiG: Good idea, implemented, and looks a lot nicer

prow/test/integration/fakeghserver/fakeghserver.go
```diff
@@ -0,0 +1,249 @@
...
+	}
+}
+
+func userHandler(ghc *fakegithub.FakeClient) func(*http.Request) (string, int, error) {
+	return func(r *http.Request) (string, int, error) {
+		logrus.Infof("Serving: %s, %s", r.URL.Path, r.Method)
+		var msg string
+		var statusCode int
+		var err error
+		statusCode = http.StatusOK
```
alvaroaleman, line 133: Nit, applies to all handlers: The code becomes easier to read if you just `return string(content), http.StatusOK, err`, rather than defining the value of the returned variable 10 lines above where you return it

alvaroaleman (approved):
/hold

Some stylistic nits, other than that looks good. Holding so you can fix them and cole can have a look

prow/test/integration/fakeghserver/fakeghserver.go
chaodaiG: Good suggestion! This was a left over effect of the `if err == nil{}` nesting logic, since it's not doing that any more, I think the suggested approach works much better. Done.

prow/test/integration/fakeghserver/fakeghserver.go
```diff
@@ -0,0 +1,249 @@
...
+		logrus.Infof("Not supported: %s, %s", r.URL.Path, r.Method)
+		return "", http.StatusNotFound,
+			fmt.Errorf("{\"error\": \"API not supported\"}, %s, %s", r.URL.Path, r.Method)
+	}
+}
+
+func userHandler(ghc *fakegithub.FakeClient) func(*http.Request) (string, int, error) {
+	return func(r *http.Request) (string, int, error) {
+		logrus.Infof("Serving: %s, %s", r.URL.Path, r.Method)
+		var msg string
```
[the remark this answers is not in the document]
chaodaiG: Prefer the other proposed approach over this one.

chaodaiG:
@cjwagner could you take another look please?

cjwagner (approved):

PR #20527, opened 2021-01-20
Config: Insert git sha when using configupdater

alvaroaleman:
This change makes the configupdater always add a key named VERSION to
the configmap which contains the git sha that triggered said update.
The configloading in turn will pick it up if present. This allows
components to include the config version in their logs, which can be
useful for debugging.

The second commit shows the reason I started thinking about this, namely the fact that the statusreconciler log is near useless, it tells us what it did but not for what config version, which makes debugging issues impossible and forces us to guess (ref https://github.com/kubernetes/test-infra/pull/20498#issuecomment-763139545)

This is technically a breaking change in that we now always include a key in the configmaps that we didn't include before (but won't override if it gets populated). This is fine for prow itself, as it only looks at all files in the context of job configs and there it ignores files that do not end with `.yaml` or `.yml`: https://github.com/kubernetes/test-infra/blob/6a328dba0fabd4a5b11ccf6d245c513bb528e2e8/prow/config/config.go#L1026

If someone uses the configupdater for something other than the prow config and then looks at all files and expects all of them to have a certain format, that would break now. But this seems rather unlikely to me. WDYT @fejta ?

/assign @fejta

### DIFF
--- a/prow/cmd/config-bootstrapper/main.go
+++ b/prow/cmd/config-bootstrapper/main.go
@@ -106,7 +106,20 @@ func run(sourcePaths []string, defaultNamespace string, configUpdater plugins.Co
 	var errors int
 	// act like the whole repo just got committed
 	var changes []github.PullRequestChange
+	var version string
+
 	for _, sourcePath := range sourcePaths {
+
+		versionFilePath := filepath.Join(sourcePath, config.ConfigVersionFileName)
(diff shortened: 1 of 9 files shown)


stevekuznetsov (approved):
Seems safe to me. Could make it `PROW-VERSION` or something if this ends up breaking something.

PR #20079, opened 2020-12-01
releng: Generate 1.20 release branch jobs

justaugustus:
- Revert "Prepare config-forker for unique names in prow"
  This reverts commit 9def652.
  The config update was causing some errors I haven't seen before, touching gubernator. Overall, deprioritizing in favor of getting the branch jobs in.
- releng/prepare_release_branch.py: Fixups
  - Parameterize min and max allowed config count
  - Add some print() statements for each function
  - Actually delete the oldest branch config before rotation
  - Block rotation if the incorrect config count is found
- Set periodic-kubernetes-bazel-test-master to fork, not the canary
- releng: Update 1.16 `prowImage` to workaround rotation bug 
- releng: Generate 1.20 release branch jobs
  generated: Run `bazel run //releng:prepare_release_branch`
- releng: Add 1.20 dashboard groups, remove 1.16

/assign @saschagrunert @xmudrii 
cc: @kubernetes/release-engineering 

/hold for some job cleanups

PR #20535, opened 2021-01-20
Prepare config-forker for unique names in prow, second try

rmohr:
Create unique names on forked jobs and modify the reporting context in
an expected way. If the original context did not include `-master`, keep
it in the original state, to not invent new context values on release
branches. Otherwise replace -master with the version to make the github
context distinct from the master jobs.

This is another try to bring in https://github.com/kubernetes/test-infra/pull/19009 after it had to be reverted in https://github.com/kubernetes/test-infra/pull/20079.

### DIFF
--- a/releng/config-forker/main.go
+++ b/releng/config-forker/main.go
@@ -85,6 +85,8 @@ func generatePresubmits(c config.JobConfig, version string) (map[string][]config
 			p := presubmit
 			p.SkipBranches = nil
 			p.Branches = []string{"release-" + version}
+			p.Context = generatePresubmitContextVariant(p.Name, p.Context, version)
+			p.Name = generatePresubmitNameVariant(p.Name, version)
 			if p.Spec != nil {
 				for i := range p.Spec.Containers {
 					c := &p.Spec.Containers[i]
@@ -347,6 +349,23 @@ func generateNameVariant(name, version string, generic bool) string {
 	}
 	if !strings.HasSuffix(name, "-master") {
 		return name + suffix
+	}
+	return strings.ReplaceAll(name, "-master", suffix)
+}
+
+func generatePresubmitNameVariant(name, version string) string {
+	suffix := "-" + version
+	if !strings.HasSuffix(name, "-master") {
+		return name + suffix
+	}
+	return strings.ReplaceAll(name, "-master", suffix)
+}
+
+func generatePresubmitContextVariant(name, context, version string) string {
+	suffix := "-" + version
+
+	if context != "" {
+		return strings.ReplaceAll(context, "-master", suffix)
 	}
 	return strings.ReplaceAll(name, "-master", suffix)
 }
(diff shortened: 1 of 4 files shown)


releng/config-forker/main.go
```diff
@@ -351,6 +353,25 @@ func generateNameVariant(name, version string, generic bool) string {
...
+
+func generatePresubmitContextVariant(name, context, version string) string {
+	suffix := "-" + version
+
+	if context != "" {
+		return strings.ReplaceAll(context, "-master", suffix)
+	}
+
+	context = name
+	return strings.ReplaceAll(context, "-master", suffix)
```
MushuEE, line 372:
```suggestion
func generatePresubmitContextVariant(name, context, version string) string {
	suffix := "-" + version

	if context != "" {
		context = name
	}

	return strings.ReplaceAll(context, "-master", suffix)
```
Does that work? The `context = name` line seems like it isn't totally needed if you return `return strings.ReplaceAll(name, "-master", suffix)` or use suggestion. Thanks!

rmohr: > The `context = name` line seems like it isn't totally needed if you return `return strings.ReplaceAll(name, "-master", suffix)` or use suggestion. Thanks!


Definitely looks better. Done. :+1:

releng/config-forker/main.go
```diff
@@ -351,6 +353,23 @@ func generateNameVariant(name, version string, generic bool) string {
 	return strings.ReplaceAll(name, "-master", suffix)
 }
 
+func generatePresubmitNameVariant(name, version string) string {
+	suffix := "-" + version
+	if !strings.HasSuffix(name, "-master") {
+		return name + suffix
```
rmohr, line 359: Note that I follow in this line the example of `generateNameVariant` which works exactly like this.
I would have expected that we would append the suffix and then still do a `ReplaceAll`:

```go
func generatePresubmitNameVariant(name, version string) string {
	suffix := "-" + version
	if !strings.HasSuffix(name, "-master") {
		name = name + suffix
	}
	return strings.ReplaceAll(name, "-master", suffix)
}
```

rmohr:
anything I can help with?

releng/config-forker/main.go
```diff
@@ -351,6 +353,23 @@ func generateNameVariant(name, version string, generic bool) string {
...
+	}
+	return strings.ReplaceAll(name, "-master", suffix)
```
saschagrunert, line 361: I recommend making this a little helper function since it is used in three places, or making `"-master"` a global const.

saschagrunert (approved):
Just a nit, otherwise LGTM

saschagrunert:
Cleanup in https://github.com/kubernetes/test-infra/pull/22490

rmohr:
> Cleanup in #22490

Thanks. Prow was too fast for me to react :)

PR #20538, opened 2021-01-20
Enabling Alpha API internal.apiserver.k8s.io/v1alpha1 on node-kubelet-alpha

knabben:
SIG Node Alpha tab is using AllAlpha=true, and `StorageVersionAPI` started to break the suite in the bootstrap around 70 days ago.

https://testgrid.k8s.io/sig-node-kubelet#node-kubelet-alpha

xRefs https://github.com/kubernetes/kubernetes/issues/98220

### DIFF
--- a/config/jobs/kubernetes/sig-node/node-kubelet.yaml
+++ b/config/jobs/kubernetes/sig-node/node-kubelet.yaml
@@ -62,6 +62,7 @@ periodics:
       - --node-tests=true
       - --provider=gce
       - --test_args=--nodes=8 --focus="\[NodeConformance\]|\[NodeAlphaFeature:.+\]" --skip="\[Flaky\]|\[Serial\]"
+      - --runtime-config=api/all=true
       - --timeout=65m
       env:
       - name: GOPATH


liggitt:
rather than disable this, shouldn't we figure out why it is breaking and fix it? cc @roycaihw

roycaihw:
It looks like the server doesn't have the required alpha API `internal.apiserver.k8s.io/v1alpha1` enabled. The API endpoint didn't exist. [E.g.](https://storage.googleapis.com/kubernetes-jenkins/logs/ci-kubernetes-node-kubelet-alpha/1351860935962136576/artifacts/tmp-node-e2e-b08b69c7-cos-81-12871-1230-3/services.log)

```
I0120 12:04:43.528242    1846 httplog.go:89] "HTTP" verb="POST" URI="/apis/internal.apiserver.k8s.io/v1alpha1/storageversions" latency="115.503µs" userAgent="e2e_node.test/v1.21.0 (linux/amd64) kubernetes/e99f60f" srcIP="127.0.0.1:48560" resp=404
```

I think it's plausible for some alpha features to depend on alpha APIs. @knabben You may want to enable alpha APIs using `--runtime-config=api/all=true` like [this config](https://github.com/kubernetes/test-infra/blob/0ce49155473e93c7a487ab216bdee09ea3b64f10/config/jobs/kubernetes/sig-cloud-provider/gcp/gcp-gce.yaml#L459). Otherwise similar issues may happen in the future.

knabben:
Enabling the api/all in the runtime config worked here, thanks @roycaihw

/assign @sjenning

knabben:
/retitle Enabling Alpha API internal.apiserver.k8s.io/v1alpha1 on node-kubelet-alpha

PR #20543, opened 2021-01-20
re-add support for match-title in pr creator

mpherman2:
We removed match-title from the pr creator in favor of headBranch, not realizing it was being used outside of test-infra and needed to be backwards compatible. This change re-adds support for match-title while keeping the updater library backwards compatible as well.

### DIFF
--- a/robots/pr-creator/main.go
+++ b/robots/pr-creator/main.go
@@ -43,6 +43,7 @@ type options struct {
 
 	title      string
 	headBranch string
+	matchTitle string
 	body       string
 }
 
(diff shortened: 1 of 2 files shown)


robots/pr-creator/main.go
```diff
@@ -100,7 +102,14 @@ func main() {
 		logrus.WithError(err).Fatal("Failed to create github client")
 	}
 
-	n, err := updater.EnsurePR(o.org, o.repo, o.title, o.body, o.source, o.branch, o.headBranch, o.allowMods, gc)
+	var queryTokensString string
+	// Prioritize using headBranch as it is less flakey
+	if o.headBranch != "" {
```
MushuEE, line 107: Should there be a check added to validate they are not both set? Or mention one has priority in the options?

mpherman2: Added to the options that headBranch overrides match-title.

fejta (approved):
/lint
/hold

PR #20522, opened 2021-01-19
Clean and update Metrics Query and Filter

MushuEE:
- Remove filtering of non-pr jobs as well as show all flake tests
- Align formatting on all configs
- Update Flakes-Daily to also include PodUtils

This will update results to look like:
```
{
  "pr:pull-kubernetes-e2e-gce-ubuntu-containerd": {
    "consistency": 0.933,
    "flakes": 25,
    "flakiest": {
      "[k8s.io] [sig-node] NodeProblemDetector [DisabledForLargeClusters] should run without error": 6,
      "[sig-network] Conntrack should be able to preserve UDP traffic when server pod cycles for a NodePort service": 5,
      "[sig-storage] In-tree Volumes [Driver: local][LocalVolumeType: tmpfs] [Testpattern: Pre-provisioned PV (default fs)] volumes should store data": 3,
      "[sig-storage] PersistentVolumes-local  [Volume type: dir] One pod requesting one prebound PVC should be able to mount volume and write from pod1": 3,
      "[sig-storage] In-tree Volumes [Driver: local][LocalVolumeType: blockfs] [Testpattern: Pre-provisioned PV (default fs)] subPath should support existing directory": 2,
      "Build": 2,
      "[sig-storage] PersistentVolumes-local  [Volume type: blockfswithformat] Two pods mounting a local volume at the same time should be able to write from pod1 and read from pod2": 2,
      "[sig-storage] In-tree Volumes [Driver: hostPathSymlink] [Testpattern: Inline-volume (default fs)] subPath should support file as subpath [LinuxOnly]": 2,
      "[sig-network] Networking Granular Checks: Services should function for service endpoints using hostNetwork": 2,
      "[k8s.io] Container Lifecycle Hook when create a pod with lifecycle hook should execute poststart exec hook properly [NodeConformance] [Conformance]": 2
    }
  },
  "ci-kubernetes-e2e-gci-gce-slow": {
    "consistency": 0.611,
    "flakes": 21,
    "flakiest": {
      "[sig-storage] In-tree Volumes [Driver: gluster] [Testpattern: Pre-provisioned PV (default fs)] volumeIO should write files of various sizes, verify size, validate content [Slow][LinuxOnly]": 6,
      "[sig-storage] In-tree Volumes [Driver: gluster] [Testpattern: Inline-volume (default fs)] volumeIO should write files of various sizes, verify size, validate content [Slow][LinuxOnly]": 4,
[...]

PR #20546, opened 2021-01-20
Prow integration test: deflake sinker test by waiting on an earlier signal of pod deletion

chaodaiG:
sinker test was found flaky in https://github.com/kubernetes/test-infra/pull/20522#issuecomment-764008120, which was likely due to the fact of pod deletion takes various amount of time. Try to deflake this by determining pod deletion based on the existence of `ObjectMeta.DeletionTimestamp`, this had worked pretty well during my local testing. By the way reducing the time of waiting from 2 minutes to 1 minute to make the test shorter, we can make this longer if this turned out to be flaky.

### DIFF
--- a/prow/test/integration/test/sinker_test.go
+++ b/prow/test/integration/test/sinker_test.go
@@ -282,11 +282,11 @@ func TestDeletePod(t *testing.T) {
 				}
 			}
 
-			// Make sure pod is deleted, it'll take roughly 3 minutes
+			// Make sure pod is deleted, it'll take roughly 1 minutes
 			// Don't care about the outcome, will check later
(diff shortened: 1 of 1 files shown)


alvaroaleman (approved):

alvaroaleman:
@chaodaiG another thing that is likely to help a lot in keeping this test flakefree it to set requests on the pod. I believe @spiffxp knows how to find out historic data of what a job uses, using that as basis to set some requests would be good.

Right now the pod can be scheduled to any node that has the capacity to fullfill the request set by the `limitrange` which is
```
     requests:
        cpu: 250m
        memory: 1Gi
```

and unlikely to be enough.

chaodaiG:
Good idea, added: https://github.com/kubernetes/test-infra/pull/20552

Issue #14343, opened 2019-09-16
Move kubeflow jobs to https://github.com/GoogleCloudPlatform/oss-test-infra/tree/master/prow/prowjobs


fejta:
@jlewi what do you think about moving the kubeflow jobs over to https://github.com/GoogleCloudPlatform/oss-test-infra/tree/master/prow/prowjobs and prow.gflocks.com?

jlewi:
@fejta That seems fine with me.

What else would need to change? i.e

Do testgrids move?
Do we use a different GCS bucket for prow artifacts?

@scottilee is this something you could help with?

jlewi:
@fejta How urgent is this on your end?

fejta:
Not urgent.

Testgrid can stay the same.
Are you still not using pod utils? If so then yes, you'll upload to a different bucket (maybe prow specifies where to upload it?)

@chases2 we should probably set up GCP/oss-test-infra to work like istio -- where we can annotate prowjobs there and have them show up in this testgrid.

jlewi:
@fejta correct we manually upload our files to GCS right now; but we could probably switch to use pod_utils.

scottilee:
@fejta Can you share some more info (e.g., a link to a ticket or document with explanation if available) on why the move from prow.k8s.io to prow.gflocks.com?

Also, would it just be creating a "kubeflow" folder in https://github.com/GoogleCloudPlatform/oss-test-infra/tree/master/prow/prowjobs and moving the files in https://github.com/kubernetes/test-infra/tree/master/config/jobs/kubeflow to there?

Lastly, I'm not familiar with pod-utils. I'm assuming it's this https://github.com/kubernetes/test-infra/tree/master/prow/pod-utils? Any more info anywhere so I can read up on it?

fejta:
> why the move 

prow.k8s.io is for kubernetes (or at least CNCF project)
prow.gflocks.com is for public google projects.

Eventually we want to migrate all non-CNCF projects out of prow.k8s.io

And yes, it is ideally
a) creating a GKE cluster to run jobs (provides you with isolation from other jobs)
b) configuring prow to schedule kubeflow jobs into that cluster
  b1) also moving any secrets, configmaps, etc that jobs use
c) moving jobs to that prow instance

> pod-utils

Let's not worry about this for now, see https://github.com/kubernetes/test-infra/blob/master/prow/jobs.md#pod-utilities for more detail.

[...]

PR #20555, opened 2021-01-21
test(kubeflow): clean up no longer used tests

Bobgy:
Part of https://github.com/kubernetes/test-infra/issues/14343
/assign @chaodaiG @jlewi 

Before moving tests for other kubeflow repos, let me first clean up ones that we no longer use.


spiffxp:
Looks like the corresponding testgroups need to be removed as well

Bobgy:
Thanks, I've removed corresponding testgroups. Let's see

Bobgy:
All passing now

Bobgy:
UPDATE: removed some more tests that I confirmed to be no longer used

PR #20557, opened 2021-01-21
Add the missing extraref into netpol job

rikatz:

Last attempt to make netpol job work with podutils, otherwise will revert and wait for some more migration steps :D

/cc @aojea


aojea:
/lgtm
/approve
This is going to be the good one :muscle:

PR #20558, opened 2021-01-21
Skipping SCTPConnectivity tests on node-kubelet-orphans tab

knabben:
There's a clean up of SCTP tests in progress (the tracking issue is: https://github.com/kubernetes/kubernetes/issues/96717)

For this reason, this PR is skipping the flake tests under `node-kubelet-orphans` until we get the SCTP task closed.

xRefs https://github.com/kubernetes/kubernetes/issues/98265

### DIFF
--- a/config/jobs/kubernetes/sig-node/node-kubelet.yaml
+++ b/config/jobs/kubernetes/sig-node/node-kubelet.yaml
@@ -191,7 +191,7 @@ periodics:
       - --node-test-args=--kubelet-flags="--cgroups-per-qos=true --cgroup-root=/"
       - --node-tests=true
       - --provider=gce
-      - --test_args=--nodes=1 --skip="\[Flaky\]|\[NodeConformance\]|\[NodeFeature:.+\]|\[NodeSpecialFeature:.+\]|\[NodeAlphaFeature:.+\]|\[Legacy:.+\]|\[Benchmark\]"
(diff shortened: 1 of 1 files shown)


aojea:
seems that the refactor in k/k is going to take time and doesn´t make sense these tests in this job
/lgtm

MHBauer:
Skipping would mean they aren't run anywhere if there isn't a specific job focusing on them. As part of fixing them they should be given some Feature flags so they are enabled elsewhere, and not run as part of this job.

knabben:
> Skipping would mean they aren't run anywhere if there isn't a specific job focusing on them. As part of fixing them they should be given some Feature flags so they are enabled elsewhere, and not run as part of this job.

@MHBauer I think they are being skipped in other places as well (until the final fix) instead of having a feature flag.

aojea:
> Skipping would mean they aren't run anywhere if there isn't a specific job focusing on them. As part of fixing them they should be given some Feature flags so they are enabled elsewhere, and not run as part of this job.

SCTP is GA and is enable by default, and tests are running in sig-network already, https://testgrid.k8s.io/sig-network-kind#sig-network-kind,%20master&include-filter-by-regex=sctp 

I lgtmed because the name of the job doesn't seem related to sctp at all and the tests are not orphan

PR #20559, opened 2021-01-21
kubernetes-csi: add Kubernetes 1.20

pohly:
The prow.sh script has ignored the patch version in the Kubernetes
version number for a while now, therefore we can simplify the
generator script and drop the patch version from all version numbers.

/cc @msau42


config/jobs/kubernetes-csi/gen-jobs.sh
```diff
@@ -20,12 +20,14 @@
...
-# All the Kubernetes versions we're testing.
-# Must have pre-built kind images (see https://hub.docker.com/r/kindest/node/tags)
+# All the Kubernetes versions we're testing. The patch version is
+# irrelevant because the prow.sh script will pick a suitable KinD
+# image or build from source.
 k8s_versions="
-1.17.0
-1.18.0
-1.19.1
+1.17
```
msau42, line 27: we can remove 1.17 as it's not supported anymore.

config/jobs/kubernetes-csi/gen-jobs.sh
```diff
@@ -288,7 +280,7 @@ snapshotter_version() {
         # All other jobs test against the latest stable snapshotter
         # release. Additional jobs could be created to cover version
         # skew, if desired.
-        echo '"v3.0.2"'
+        echo '"v3.0.3"'
```
msau42, line 283: hm now we have 4.0 but it requires 1.20 as a min version cc @xing-yang

config/jobs/kubernetes-csi/gen-jobs.sh
pohly: I was wondering about that too. I kept it because I first wanted to add 1.20 (i.e. always test on three releases), but I agree that combining the removal of the old release with the addition of the new (even if not enabled immediately) cuts down the overall effort - I'll change it.

config/jobs/kubernetes-csi/gen-jobs.sh
pohly: I suppose we need a switch statement somewhere then?

xing-yang: Yeah, maybe set it to 4.0 for 1.20 and higher and set it to 3.0.3 for 1.19 or lower.

pohly: So we don't need (or want) to test 3.0.3 on 1.20? Some users might run it there.

That's fine with me, I just wanted to be sure.

pohly: Added a Kubernetes check.

config/jobs/kubernetes-csi/gen-jobs.sh
msau42: Has this been removed yet?

msau42:
/lgtm
/approve

1.17 can be removed as a followup

PR #16125, opened 2020-02-04
Increase timeout/interval for Windows containerd job to 4h

pjh:
Test runs for this job are regularly failing at the 2 hour mark: "Test started today at 7:42 AM failed after 2h0m17s. (more info)"
https://testgrid.k8s.io/google-windows#gce-windows-2019-containerd-master&width=20

This PR updates the test timeout and interval to 4 hours instead of 2. containerd is known to be significantly slower than dockershim on Windows for now.

PR #20560, opened 2021-01-21
Increase Windows test timeout

jingxu97:
Increase timeout to see whether it help downloading log files

See issue for some details kubernetes/kubernetes#90224

### DIFF
--- a/config/jobs/kubernetes/sig-windows/windows-gce.yaml
+++ b/config/jobs/kubernetes/sig-windows/windows-gce.yaml
@@ -58,7 +58,7 @@ periodics:
     repo: windows-testing
     base_ref: master
     path_alias: k8s.io/windows-testing
-  interval: 2h
+  interval: 4h
   labels:
     preset-k8s-ssh: "true"
     preset-service-account: "true"
@@ -82,7 +82,7 @@ periodics:
       - --test_args=--ginkgo.focus=\[Conformance\]|\[NodeConformance\]|\[sig-windows\] --ginkgo.skip=\[LinuxOnly\]|\[Serial\]|\[Feature:.+\] --minStartupPods=8
       - --test-cmd-args=--ginkgo.focus=\[Conformance\]|\[NodeConformance\]|\[sig-windows\] --ginkgo.skip=\[LinuxOnly\]|\[Serial\]|\[Feature:.+\] --minStartupPods=8
       - --test-cmd-args=--node-os-distro=windows
-      - --timeout=120m
+      - --timeout=180m
       env:
       - name: WINDOWS_NODE_OS_DISTRIBUTION
         value: "win2019"
@@ -103,7 +103,7 @@ periodics:
     repo: windows-testing
     base_ref: master
     path_alias: k8s.io/windows-testing
-  interval: 2h
+  interval: 4h
   labels:
     preset-k8s-ssh: "true"
     preset-service-account: "true"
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes/sig-windows/windows-gce.yaml
```diff
@@ -82,7 +82,7 @@ periodics:
       - --test_args=--ginkgo.focus=\[Conformance\]|\[NodeConformance\]|\[sig-windows\] --ginkgo.skip=\[LinuxOnly\]|\[Serial\]|\[Feature:.+\] --minStartupPods=8
       - --test-cmd-args=--ginkgo.focus=\[Conformance\]|\[NodeConformance\]|\[sig-windows\] --ginkgo.skip=\[LinuxOnly\]|\[Serial\]|\[Feature:.+\] --minStartupPods=8
       - --test-cmd-args=--node-os-distro=windows
-      - --timeout=120m
+      - --timeout=180m
```
pjh, line 85: I think you need to change the `interval` field up above too.

pjh (commented):
My expectation is that you'll also need to add this configuration for the jobs you're changing:

```
  decoration_config:
    timeout: 3h
```

This is based on https://github.com/kubernetes/test-infra/pull/16125/files, and the `ci-kubernetes-e2e-windows-gce-serial`, `ci-kubernetes-e2e-windows-containerd*` and `ci-kubernetes-e2e-windows-node-throughput*` jobs already in this file.

pjh:
We'll also need to update the configs in these files, though that can probably be done in a separate PR, after this one is merged and the results look good.

https://github.com/kubernetes/test-infra/blob/c094f147c10f9b667472e81a5b36e12a61446e37/config/jobs/kubernetes/sig-release/release-branch-jobs/1.17.yaml
https://github.com/kubernetes/test-infra/blob/c094f147c10f9b667472e81a5b36e12a61446e37/config/jobs/kubernetes/sig-release/release-branch-jobs/1.18.yaml
https://github.com/kubernetes/test-infra/blob/c094f147c10f9b667472e81a5b36e12a61446e37/config/jobs/kubernetes/sig-release/release-branch-jobs/1.19.yaml
https://github.com/kubernetes/test-infra/blob/ca33369cd86ca3faa01d3d64fbd61528a3bee3f5/config/jobs/kubernetes/sig-release/release-branch-jobs/1.20.yaml

pjh (commented):
Please reference https://github.com/kubernetes/kubernetes/issues/90224 in the PR description.

config/jobs/kubernetes/sig-windows/windows-gce.yaml
jingxu97: oh, got it. But I am wondering whether it is actually allowed to have interval shorten than timeout. I assume multiple tests can be run concurrently?

pjh: I'm guessing that with interval > timeout this will work fine, but I don't have any detailed knowledge of how these configs work. If something breaks, we'll fix it.

pjh:
The adjusted interval looks good. Please take a look at the other comments I added.

jingxu97:
> The adjusted interval looks good. Please take a look at the other comments I added.

I update msg during commit, but it is not shown up. Directly update it here. Thansk!

pjh:
What about `decoration_config` (see comment above)?

I don't exactly know what that field does, but since other jobs have it I'd go ahead and add it for these too.

jingxu97:
> My expectation is that you'll also need to add this configuration for the jobs you're changing:
> 
> ```
>   decoration_config:
>     timeout: 3h
> ```
> 
> This is based on https://github.com/kubernetes/test-infra/pull/16125/files, and the `ci-kubernetes-e2e-windows-gce-serial`, `ci-kubernetes-e2e-windows-containerd*` and `ci-kubernetes-e2e-windows-node-throughput*` jobs already in this file.

I searched a little bit, I think by default the timeout is 24 hours https://github.com/kubernetes/test-infra/blob/5a22a6073063eb110523bc93ee31f5c5b4994714/prow/config/config.go#L62
So I think it is fine that we don't set this value.

pjh:
It looks like decoration_config is not for the test job itself, but for the pods run during tests: https://github.com/kubernetes/test-infra/blob/master/prow/apis/prowjobs/v1/types.go#L163

```
	// DecorationConfig holds configuration options for
	// decorating PodSpecs that users provide
	DecorationConfig *DecorationConfig `json:"decoration_config,omitempty"`
```

Anyway, seems fine to omit.

PR #20563, opened 2021-01-22
Fix autobumper PR title contains info about bumps that did not occur.

mpherman2:
Two issues were causing inaccurate PR titles in the autobumper. The first is that getVersionsAndCheckConsistency was supposed to return a map of tags:images for all bumped images, but if the image was not bumped, the unbumped tag and image were added to the map. This was causing the PR title to say "Prefix was bumped to x" when Prefix was already at version x and was not bumped in the PR.

The second issue was that for images that are not bumped consistently, we just say "Prefix was bumped as needed", but this also included if it was not bumped at all. This has now been changed so that if the prefix was not bumped at all it is not included in the PR title.

### DIFF
--- a/experiment/autobumper/bumper/bumper.go
+++ b/experiment/autobumper/bumper/bumper.go
@@ -539,18 +539,31 @@ func isUnderPath(name string, paths []string) bool {
 	return false
 }
 
+// getVersionsAndCheckConisistency takes a list of Prefixes and a map of
+// all the images found in the code before the bump : their versions after the bump
+// For example {"gcr.io/k8s-prow/test1:tag": "newtag", "gcr.io/k8s-prow/test2:tag": "newtag"},
+// and returns a map of new versions resulted from bumping : the images using those versions.
+// It will error if one of the Prefixes was bumped inconsistently when it was not supposed to
 func getVersionsAndCheckConsistency(prefixes []Prefix, images map[string]string) (map[string][]string, error) {
 	// Key is tag, value is full image.
 	versions := map[string][]string{}
 	for _, prefix := range prefixes {
 		newVersions := 0
+		unbumped_found := false
 		for k, v := range images {
 			if strings.HasPrefix(k, prefix.Prefix) {
 				if _, ok := versions[v]; !ok {
 					newVersions++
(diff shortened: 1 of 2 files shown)


alvaroaleman (approved):

experiment/autobumper/bumper/bumper.go
```diff
@@ -544,13 +544,21 @@ func getVersionsAndCheckConsistency(prefixes []Prefix, images map[string]string)
...
+		unbumped_found := false
 		for k, v := range images {
 			if strings.HasPrefix(k, prefix.Prefix) {
 				if _, ok := versions[v]; !ok {
 					newVersions++
 				}
-				versions[v] = append(versions[v], k)
-				if prefix.ConsistentImages && newVersions > 1 {
+				//Only add the bumped images to the new versions map
+				if !strings.Contains(k, v) {
```
cjwagner, line 554: Why would the image name ever contain the tag? Does `k` include the existing tag (not bump tag) as well?
It would be very helpful to have function comments explaining the format of the parameters. It is difficult to trace the source of the `images` parameter to determine its key and value format.

experiment/autobumper/bumper/bumper.go
```diff
@@ -587,7 +595,13 @@ func makeCommitSummary(prefixes []Prefix, versions map[string][]string) string {
 	var consistentBumps []string
 	for _, prefix := range prefixes {
 		if !prefix.ConsistentImages {
-			inconsistentBumps = append(inconsistentBumps, prefix.Name)
+			for _, imageList := range versions {
+				//Should not be possible for tag to be in map with empty imageList
+				if strings.HasPrefix(imageList[0], prefix.Prefix) {
```
cjwagner, line 600: This won't work if multiple prefixes bump to the same version, right? I think in that case the first entry in `imageList` may not contain the correct prefix, but a later entry would. Please add a test to cover this case as well.

mpherman2: Good catch. I did not consider this possibility. Fixed and added unit test

cjwagner (approved):

PR #20564, opened 2021-01-22
feat(github): add AddLabels function to RepositoryClient interface

taharah:
Add a method to the GitHub client RepositoryClient interface to add multiple labels to a PR with a single API call.

Signed-off-by: Trevor Wood <Trevor.G.Wood@gmail.com>

### DIFF
--- a/prow/github/fakegithub/fakegithub.go
+++ b/prow/github/fakegithub/fakegithub.go
@@ -373,6 +373,29 @@ func (f *FakeClient) AddLabel(owner, repo string, number int, label string) erro
 	return fmt.Errorf("cannot add %v to %s/%s/#%d", label, owner, repo, number)
 }
 
+// AddLabels adds a list of labels
+func (f *FakeClient) AddLabels(owner, repo string, number int, labels ...string) error {
+	for _, label := range labels {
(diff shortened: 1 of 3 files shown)


prow/github/fakegithub/fakegithub.go
```diff
@@ -373,6 +373,29 @@ func (f *FakeClient) AddLabel(owner, repo string, number int, label string) erro
 	return fmt.Errorf("cannot add %v to %s/%s/#%d", label, owner, repo, number)
 }
 
+// AddLabels adds a list of labels
+func (f *FakeClient) AddLabels(owner, repo string, number int, labels ...string) error {
```
cjwagner, line 377: Please deduplicate code shared with the `AddLabel` function above. (i.e. make one of these functions call the other like you did for the real client functions.)

cjwagner (commented):
This looks reasonable to me, but what is the motivation? Where will the new function be used?

taharah:
> This looks reasonable to me, but what is the motivation? Where will the new function be used?

I was working on developing an external plugin that needs to apply multiple labels. Since the API supports adding more than one label in a single call, I thought it would be useful to add the method to the client rather than having to loop over the list of labels and make multiple calls.

cjwagner (approved):

cjwagner (approved):

PR #20576, opened 2021-01-22
Enable tide priority labels in config

matthyx:


matthyx:
@cjwagner so checkconfig should pass tomorrow once we have bumped prow?

cjwagner:
> @cjwagner so checkconfig should pass tomorrow once we have bumped prow?

Yes, we can `/retest` after the next Prow bump on Monday (which I'll do since I'm oncall).

/assign @cblecker 
^ For approval since this changes how Tide will prioritize PRs.

matthyx:
@cblecker the PR is ready for approval, thx!

matthyx:
ping @cblecker @liggitt as code freeze is coming in less than a month :)

matthyx:
> ping @cblecker @liggitt as code freeze is coming in less than a month :)

(I mean tide pools will start to grow again)

config/prow/config.yaml
```diff
@@ -597,13 +597,12 @@ tide:
             skip-unknown-contexts: true
   batch_size_limit:
     "kubernetes/kubernetes": 15
-#  TODO uncomment after merging #20468
```
BenTheElder, line 600: FYI @liggitt @dims @spiffxp

cblecker (approved):
I have no concerns about this.

/lgtm
/approve
/hold

@nikhita Any thoughts from you?

config/prow/config.yaml
```diff
@@ -597,13 +597,12 @@ tide:
...
-#  priority:
-#  # prioritize PRs fixing flaking/failing tests
-#  - labels: [ "kind/failing-test" ]
-#  - labels: [ "area/deflake" ]
-#  # prioritize critical bugfixes
-#  - labels: [ "kind/bug", "priority/critical-urgent" ]
+  priority:
+  # prioritize PRs fixing flaking/failing tests
+  - labels: [ "kind/failing-test" ]
+  - labels: [ "area/deflake" ]
```
nikhita, line 603: I think it'd be good to add `priority/critical-urgent` here.

Just to have consistency on how we define tide priority for k/k - probably have `kind/*`, `priority/*` labels as mandatory and `area/*` as optional

nikhita (commented):
Left one suggestion. Thanks for implementing the feature and following this through with the config change! :heart:

config/prow/config.yaml
matthyx: Hi @nikhita :)

So you only want to prioritize `area/deflake` PRs that are also labelled as `priority/critical-urgent`?

nikhita: > So you only want to prioritize area/deflake PRs that are also labelled as priority/critical-urgent?

Yes :+1: 

The intent behind it is to ensure that if we are bumping the tide priority, it should also have a priority label on it :)

matthyx: there you go

config/prow/config.yaml
```diff
@@ -597,13 +597,12 @@ tide:
...
-#  priority:
-#  # prioritize PRs fixing flaking/failing tests
-#  - labels: [ "kind/failing-test" ]
-#  - labels: [ "area/deflake" ]
-#  # prioritize critical bugfixes
-#  - labels: [ "kind/bug", "priority/critical-urgent" ]
+  priority:
+  # prioritize PRs fixing flaking/failing tests
+  - labels: [ "kind/failing-test" ]
+  - labels: [ "area/deflake", "priority/critical-urgent" ]
```
nikhita, line 603: Sorry, I should have been clearer :)

I think we should have this priority but will let @liggitt confirm since he opened https://github.com/kubernetes/test-infra/issues/18985

```yaml
  - labels: [ "kind/flake", "area/deflake", "priority/critical-urgent" ]
  - labels: [ "kind/failing-test", "priority/critical-urgent" ]
  - labels: [ "kind/bug", "priority/critical-urgent" ]
```

matthyx: ah ok go you now :)
yeah, if Jordan could step in...

matthyx: /assign @liggitt

spiffxp: I have one concern here, but I'm not sure if it should block.

My read is that this is the same priority _for all tide queries_.  Is there any reason this wasn't implemented as a per-query setting?

I'm not sure that what `kubernetes/kubernetes` wants in terms of priority should dictate what `kubernetes/website` gets, for example.

liggitt: I don't have a strong opinion on requiring the priority labels either way, but I don't object. requiring both `"kind/flake"` and `"area/deflake"` is a bit tedious (it's not really clear to someone opening a PR to fix a bad flake they should add that area)

liggitt: I'm happy with any of the above iterations

/unassign

matthyx: > I'm not sure that what `kubernetes/kubernetes` wants in terms of priority should dictate what `kubernetes/website` gets, for example.

Good question... I don't think it's that critical because it's mostly k/k that has a huge tide backlog?

spiffxp: >requiring both "kind/flake" and "area/deflake" is a bit tedious

Agreed. Drop the area label. Keep it to kind and priority. Those are labels consistent across the org

Otherwise yeah I think this is worth trying out

matthyx: @spiffxp done

spiffxp (approved):
/approve
/lgtm
/hold cancel
/meowvie
:shipit: psyched to see this

PR #20596, opened 2021-01-25
kubernetes-csi 1.20, II

pohly:
This is a update for https://github.com/kubernetes/test-infra/pull/20559 which fixes two things:
- Prow jobs need major/minor/patch Kubernetes version (https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-csi_csi-driver-host-path/235/pull-kubernetes-csi-csi-driver-host-path-1-17-on-kubernetes-1-17/1353777478430101504)
- drop 1.17


config/jobs/kubernetes-csi/gen-jobs.sh
```diff
@@ -344,7 +342,7 @@ EOF
         # unrelated to the PR. Testing against the latest Kubernetes is covered
         # by periodic jobs (see https://k8s-testgrid.appspot.com/sig-storage-csi-ci#Summary).
         - name: CSI_PROW_KUBERNETES_VERSION
-          value: "$kubernetes"
+          value: "$kubernetes.0"
```
msau42, line 345: This might be ok for now, but sometimes there are bug fixes or changes to tests made in patch versions that we may want.

msau42 (commented):
/lgtm
/approve

config/jobs/kubernetes-csi/gen-jobs.sh
pohly: Yes. Addressing that then will be a bigger change to the `gen-jobs.sh` script again - more that I wanted to tackle now.

PR #20598, opened 2021-01-25
Fix upstream bumps not showing correct commit diff

mpherman2:
The PR Summary generates the commit diff by using the images map created during the image bump. The upstreamImageResolver was creating this map incorrectly breaking this part of the pr summary when targetVersion was upstream

### DIFF
--- a/experiment/autobumper/bumper/bumper.go
+++ b/experiment/autobumper/bumper/bumper.go
@@ -477,7 +477,7 @@ func upstreamImageVersionResolver(
 	}
 
 	return func(imageHost, imageName, currentTag string) (string, error) {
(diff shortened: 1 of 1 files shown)


fejta (approved):

experiment/autobumper/bumper/bumper.go
```diff
@@ -477,7 +477,7 @@ func upstreamImageVersionResolver(
 	}
 
 	return func(imageHost, imageName, currentTag string) (string, error) {
-		imageFullPath := imageHost + "/" + imageName + "/" + currentTag
+		imageFullPath := imageHost + "/" + imageName + ":" + currentTag
```
fejta, line 480: Consider adding a unit test to cover the scenario we were having problems with.

PR #20601, opened 2021-01-25
update template url to aks-e for dual-stack

aramase:
- Switches to using `aks-engine` published templates for dual-stack jobs.


MrHohn (approved):
/lgtm
/approve

PR #20603, opened 2021-01-26
Add GCE Windows 20H2 test job

pjh:
This change adds a new test job using Windows 20H2 nodes on GCE. It will run against K8s 1.21 at head.

### DIFF
--- a/config/jobs/kubernetes/sig-windows/windows-gce.yaml
+++ b/config/jobs/kubernetes/sig-windows/windows-gce.yaml
@@ -141,6 +141,51 @@ periodics:
     testgrid-tab-name: gce-windows-1909-master
     description: Runs tests on a Kubernetes cluster with Windows 1909 nodes on GCE
 
+- name: ci-kubernetes-e2e-windows-gce-20h2
+  decorate: true
+  extra_refs:
+  - org: kubernetes-sigs
+    repo: windows-testing
+    base_ref: master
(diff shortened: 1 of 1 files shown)


pjh:
/cc @jeremyje 

I'm running an e2e test job manually against a 20H2 cluster now to verify that this job will work.
/hold

pjh:
I'm struggling to run a full set of e2e tests against the cluster I brought up with Windows 20H2 nodes (see https://github.com/kubernetes/kubernetes/pull/98487), but of the 42 tests that did run successfully ~75% of them passed. It's possible that the failing tests took so long that it caused the entire test run to time out - I need to look into that further.

Anyway, it seems like adding the job now will start to provide us some useful 20H2 signal, so I'd like to submit this now. If for some reason it fails hard or it ends up being very flaky then we can revert this and re-submit after making some fixes. @jeremyje what do you think?

/unhold

pjh:
[agnhost:2.26](https://hub.docker.com/layers/e2eteam/agnhost/2.26/images/sha256-8502ab55bae3eccd057ab7b18e14f5f617c1b101a57443cd4b454c24df646eca?context=explore) has been rebuilt to support 20H2. Since agnhost is use for the majority of the e2e test cases, that also suggests we'll start to get useful signal from this test job.

We're chatting in Slack to determine how + when the other Windows test containers can be updated for 20H2.

PR #20607, opened 2021-01-26
Statusreconciler: add an option denylist-all

chaodaiG:
Denylist is only used for triggerring new jobs, but not retire or migrate. This makes it impossible to stop statusreconciler from mark contexts as 'retired' when the workloads are moved from prow instance A to prow instance B.

Related: https://github.com/kubernetes/test-infra/issues/14343

### DIFF
--- a/prow/statusreconciler/controller_test.go
+++ b/prow/statusreconciler/controller_test.go
@@ -867,7 +867,7 @@ func TestControllerReconcile(t *testing.T) {
 		expectErr bool
 	}{
 		{
-			name: "ignored org skips creation, still does retire and migrate",
+			name: "ignored org skips creation, retire and migrate",
 			generator: func() (Controller, func(*testing.T)) {
 				fpjt := newfakeProwJobTriggerer()
 				fghc := newFakeGitHubClient(orgRepoKey)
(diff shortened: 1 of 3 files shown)


chaodaiG:
Context: after removing all prow jobs, trigger plugin, tide from kubeflow/pipelines repo in https://github.com/kubernetes/test-infra/pull/20553, plus add kubeflow/pipelines to statusreconciler denylist in https://github.com/kubernetes/test-infra/pull/20591, "context retired" still happened in kubeflow/pipelines repo, which made the tests "green" and pretty scary.

prow/statusreconciler/controller_test.go
```diff
@@ -867,7 +867,7 @@ func TestControllerReconcile(t *testing.T) {
 		expectErr bool
 	}{
 		{
-			name: "ignored org skips creation, still does retire and migrate",
```
alvaroaleman, line 870: This testcase makes it sound like the current behavior was intended. @stevekuznetsov do you recall why the denylist only applies to job creations and not to retirements and migraitons?

stevekuznetsov: IIRC the only real use for the deny-list was to not cause thundering herds of new jobs on `kubernetes/kubernetes` - migrating and retiring just use GH tokens.

chaodaiG: The initial purpose of not causing spamming of new jobs totally makes sense, not quite sure what `migrating and retiring just use GH tokens` means though.

stevekuznetsov: There was always enough tokens in order to retire old jobs and migrate contexts - even if the changing jobs were for `kubernetes/kubernetes`. So, these code paths were not part of the deny-list. This is in contrast to compute resources, which is what made triggering new jobs for `kubernetes/kubernetes` impossible.

chaodaiG: Gotcha. The question is, does `kubernetes/kubernetes` still need the retire and migrate behaviors? If that's the case, then we may need to have more granular control such as breaking `denylist` to the following:

- `denylist-retire`
- `denylist-migrate`
- `denylist-trigger`

Alternatively, we can make `denylist` take in a map, which would be a breaking change though.

chaodaiG: @stevekuznetsov , any thoughts?

stevekuznetsov: I imagine it will be very sorely missed if someone changes a presubmit on `k/k` and the stale contexts are not cleaned up. cc @BenTheElder 

I'd maybe be in favor of a new deny-list field that does everything, and keep the current one as-is?

chaodaiG: sgtm, added a new option and kept the current one as is. @stevekuznetsov please take another look

chaodaiG:
Unrelated test failure
/retest

stevekuznetsov (approved):

PR #20610, opened 2021-01-26
[Kettle] Handle case when build table is empty

MushuEE:
In monitoring Kettle last night I saw a new failure

```
Traceback (most recent call last):
  File "make_json.py", line 328, in <module>
    sys.exit(main(DB, OPTIONS, sys.stdout))
  File "make_json.py", line 296, in main
    if oldest < time.time() - opts.assert_oldest * SECONDS_PER_DAY:
TypeError: '<' not supported between instances of 'NoneType' and 'float'
```
This should not impact the behavior much as far as I understand, but want to target this behavior so there are no stacks in prod

/area kettle
/type bug

### DIFF
--- a/kettle/make_json.py
+++ b/kettle/make_json.py
@@ -293,8 +293,10 @@ def main(db, opts, outfile):
 
     if opts.assert_oldest:
         oldest = db.get_oldest_emitted(incremental_table)
+        if oldest is None:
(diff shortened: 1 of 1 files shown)


spiffxp (approved):
/approve
/lgtm

Issue #20595, opened 2021-01-25
Prow: malformed prowjob should have been GCed


chaodaiG:
**What happened**:

A malformed prowjob that would have never been scheduled a pod due to a typo in `cluster` field, has been hanging around for more than 4 days, and prow-controller-manager keeps on reporting errors on this prowjob.

**What you expected to happen**:

The prowjob should have been GCed earlier, maybe the same as max pod age (48 hours)?

**How to reproduce it (as minimally and precisely as possible)**:

Create a prowjob with `cluster` defined to be a cluster not exist.

**Please provide links to example occurrences, if any**:

**Anything else we need to know?**:

chaodaiG:
An initial thought, as discussed with @cjwagner , is that sinker can delete prowjob whose age is over the max prowjob age. Quoted from offline discussion:

```
PJ **runtime** should really be controlled with the decoration timeout and graceperiod since that can be configured granularly. The max PJ age is to allow the PJ resources to exist beyond the lifetime of the pod.
```

chaodaiG:
CC @cjwagner @alvaroaleman

alvaroaleman:
I would argue this is working as intended. Maybe it is not a typo and someone is planning on adding that cluster? Prow can't know and should IMHO not make assumptions

chaodaiG:
Agree that prow shouldn't make assumptions. This behavior of a prowjob hanging for more than 4 days, however, is concerning. I still think it should be GCed automatically instead of prow operator manually remove them.

alvaroaleman:
>  This behavior of a prowjob hanging for more than 4 days, however, is concerning. I still think it should be GCed automatically instead of prow operator manually remove them.
Well, Sinker will eventually do that.

We could probably fix this TODO to make it spam the log less: https://github.com/kubernetes/test-infra/blob/e3a8e137c7b928824d358ee55be35806f7aaa8fa/prow/plank/reconciler.go#L552-L553

chaodaiG:
Agreed that fixing this TODO will do, so I'm onboard with it. @cjwagner what do you think?

cjwagner:
> > This behavior of a prowjob hanging for more than 4 days, however, is concerning. I still think it should be GCed automatically instead of prow operator manually remove them.
> 
> Well, Sinker will eventually do that.

[...]

PR #20612, opened 2021-01-26
Plank: don't requeue when build cluster doesn't exist

chaodaiG:
Plank loads build clusters at start up, and stores them in memory. If a build cluster is not found in this map, then it will not find it in next reconcilation cycle. There is no reason to requeue prowjobs failed with this reason since it's bound to failure

Fixes https://github.com/kubernetes/test-infra/issues/20595

### DIFF
--- a/prow/plank/error.go
+++ b/prow/plank/error.go
@@ -1,0 +1,31 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package plank
+
+import "errors"
+
+var ClusterClientNotExistError *clusterClientNotExistError
+
+type clusterClientNotExistError struct{}
+
+func (ce *clusterClientNotExistError) Error() string {
+	return "Cluster client not exist"
+}
+
+func (ce *clusterClientNotExistError) Unwrap() error {
+	return errors.New("Cluster client not exist")
+}
--- a/prow/plank/error.go
+++ b/prow/plank/error.go
@@ -16,16 +16,29 @@ limitations under the License.
 
 package plank
 
-import "errors"
+import (
+	"errors"
+	"fmt"
+)
 
-var ClusterClientNotExistError *clusterClientNotExistError
-
-type clusterClientNotExistError struct{}
-
-func (ce *clusterClientNotExistError) Error() string {
-	return "Cluster client not exist"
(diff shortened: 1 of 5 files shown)


prow/plank/error.go
```diff
@@ -0,0 +1,31 @@
...
+
+var ClusterClientNotExistError *clusterClientNotExistError
+
+type clusterClientNotExistError struct{}
+
+func (ce *clusterClientNotExistError) Error() string {
+	return "Cluster client not exist"
+}
+
+func (ce *clusterClientNotExistError) Unwrap() error {
```
alvaroaleman, line 29: I don't think you need an `Unwrap`, you need an `Is` and we need to make sure that the `%w` verb is used with `fmt.Errorf` everywhere (There are linters for that).

A nicer pattern for triggering behavior via errors is to do it as wrapper, because that allows you to easily re-use the triggering behavior part for other use cases (say: Podspec is invalid, that will never succeed either). Something like this: https://github.com/openshift/ci-tools/blob/e253c1cb44c7eed7468086efb3812a02c283c6ee/pkg/controller/util/error.go#L10

chaodaiG: Looking at the `TerminalError` implementation, agree that it's pretty elegant implementation, but feel like this would exactly be how go1.13 implemented `Is` and `Unwrap` in my imagination, which makes me think it's like re-inventing a very tiny wheel. IMHO using `%w` and `Is` is more `go` way.

For the reuse scenario, I think I can make it more general by renaming the error to be something like `nonRetryableError` so that it can be used by other cases as well.

alvaroaleman: > Looking at the TerminalError implementation, agree that it's pretty elegant implementation, but feel like this would exactly be how go1.13 implemented Is and Unwrap in my imagination, which makes me think it's like re-inventing a very tiny wheel. IMHO using %w and Is is more go way.

What I linked is entirely based on golangs 1.13 errors.Is, not sure I understand the comment?

> For the reuse scenario, I think I can make it more general by renaming the error to be something like nonRetryableError so that it can be used by other cases as well.

That would be great. Also, if we want to keep this working we should probably enforce `%w` usage via a linter

chaodaiG: > What I linked is entirely based on golangs 1.13 errors.Is, not sure I understand the comment?

Yeah that's right, sorry I overlooked.

 
> > For the reuse scenario, I think I can make it more general by renaming the error to be something like nonRetryableError so that it can be used by other cases as well.
> 
> That would be great. Also, if we want to keep this working we should probably enforce `%w` usage via a linter

I just looked up and couldn't find where golint is configured for this repo, did I overlook something?

alvaroaleman: > I just looked up and couldn't find where golint is configured for this repo, did I overlook something?

We do not have any linting today, it got stuck on @fejta insisting to do that within bazel via repo-infra and no one from the non-googlers knows how do do that: https://github.com/kubernetes/test-infra/pull/19589#discussion_r507889662

chaodaiG: Gotcha, thanks for all the contexts! Switched to go with TerminalError as wrapper function approach

alvaroaleman (approved):

chaodaiG:
Test failure related to https://github.com/kubernetes/test-infra/issues/20628

PR #20614, opened 2021-01-26
Add support to autobumper for multiple repos with the same name

mpherman2:
The autobumper would run into issues if we wanted to bump a repo with a name that was already in use by the robot's GH account. This change allows configuring the name of the repo we wish the robot to use for the remote fork and will create the fork if it does not exist.

### DIFF
--- a/experiment/autobumper/bumper/bumper.go
+++ b/experiment/autobumper/bumper/bumper.go
@@ -111,7 +111,8 @@ type Options struct {
 	ExtraFiles []string `yaml:"extraFiles"`
 	// The target version to bump images version to, which can be one of latest, upstream, upstream-staging and vYYYYMMDD-deadbeef.
 	TargetVersion string `yaml:"targetVersion"`
-	// The name used in the address when creating remote. Format will be git@github.com:{GitLogin}/{RemoteName}.git
+	// The name used in the address when creating remote. This should be the same name as the fork. If fork does not exist this will be the name of the fork that is created.
+	// If it is not the same as the fork, the robot will change the name of the fork to this. Format will be git@github.com:{GitLogin}/{RemoteName}.git
 	RemoteName string `yaml:"remoteName"`
 	// The name of the branch that will be used when creating the pull request. If unset, defaults to "autobump".
 	HeadBranchName string `yaml:"headBranchName"`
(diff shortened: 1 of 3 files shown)


experiment/autobumper/bumper/bumper.go
```diff
@@ -288,11 +288,22 @@ func Run(o *Options) error {
...
 			return fmt.Errorf("fork needed for autobump does not exist. unable to create new fork. %v", err)
 		}
+		// If a new fork was created with a name other than o.RemoteName
+		if forkName != o.RemoteName {
+			var updateRequest = github.RepoUpdateRequest{
+				RepoRequest: github.RepoRequest{
+					Name: &o.RemoteName,
+				},
+			}
+			_, err := gc.UpdateRepo(o.GitHubLogin, forkName, updateRequest)
```
chaodaiG, line 302: After moving repo from `orgA/repoA` to `orgA/repoB`, Github redirects all requests against `orgA/repoA` to `orgA/repoB`, does this affect this use case? The reason for this question it that I  imagine the reason for this improvement is that the `forkName` is supposed to be repurposed to be used by something else, would later on querying `forkName` from somewhere else get you unexpected results?

experiment/autobumper/bumper/bumper.go
```diff
@@ -288,11 +288,22 @@ func Run(o *Options) error {
...
+			if err != nil {
+				return fmt.Errorf("unable to change name of forked repo from %s to %s due to error: %v. forked repo needs to be named %s. Either make this fix manually or change o.RemoteName", forkName, o.RemoteName, err, o.RemoteName)
```
chaodaiG, line 304: nit, use %w for error in `fmt.Errorf`

prow/github/client.go
```diff
@@ -3582,7 +3582,8 @@ func (c *client) EnsureFork(forkingUser, org, repo string) (string, error) {
 	if err != nil {
 		return repo, fmt.Errorf("could not fetch all existing repos: %v", err)
 	}
-	if !repoExists(fork, repos) {
+	// if the repo does not exist, or it does, but is not a fork of the repo we want
+	if exists, response := repoExists(fork, repos); !exists || response.Parent.FullName != fmt.Sprintf("%s/%s", org, repo) {
```
chaodaiG, line 3586: `repoExists` doesn't need to return two variables. One possible alternative is:

```
func getFork(fork string, repos []Repo) *Repo {
  // If not found, return nil
  // return pointer of the repo
}
```

Then this line can be:

```
if repo := getFork(fork, repos); repo == nil || repo.Parent.FullName != fmt.Sprintf("%s/%s", org, repo) {
```

experiment/autobumper/bumper/bumper.go
mpherman2: forkName is returned by EnsureFork. EnsureFork checks to see if a fork already exists for a given repo, if not creates one, and then returns the name of the fork. So if we had a istio-test-infra repo that was a fork of istio/test-infra Ensure fork would return "istio-test-infra." If we did not already have a istio fork, but had another test-infra fork, Ensurefork would create a fork and return "test-infra-1". forkName is only used to compare this with what the config file says the remote name is and if they don't match it will change the name of the fork with UpdateRepo. forkName is not used ever again. Most of the time forkName == remoteName. The only case where this won't be true is on the first run of the autobumper if the fork does not already exist.

experiment/autobumper/bumper/bumper.go
mpherman2: Done

prow/github/client.go
mpherman2: Done

experiment/autobumper/bumper/bumper.go
mpherman2: I added a log info when the fork name gets changed, and elaborated on the Options.RemoteName comment to hopefully make this more clear as well.

chaodaiG: ah. I see, totally makes sense now. Thank you!

chaodaiG (approved):
/hold
In case @alvaroaleman wants to take a look

prow/github/client.go
```diff
@@ -3582,7 +3582,8 @@ func (c *client) EnsureFork(forkingUser, org, repo string) (string, error) {
 	if err != nil {
 		return repo, fmt.Errorf("could not fetch all existing repos: %v", err)
 	}
-	if !repoExists(fork, repos) {
+	// if the repo does not exist, or it does, but is not a fork of the repo we want
+	if forkedRepo := getFork(fork, repos); forkedRepo == nil || forkedRepo.Parent.FullName != fmt.Sprintf("%s/%s", org, repo) {
```
fejta, line 3586: In what scenario will the full name be wrong?

fejta (approved):
/retest

prow/github/client.go
mpherman2: The situation is we are bumping "istio/test-infra", and the robot's gh account has a repo called "test-infra" but that repo was forked from k8s/test-infra, so instead of just checking if "robot/test-infra" exists, we also check that it's parent's full name is "istio/test-infra", if not, the fork we want might not exist.

fejta: Very clever, thanks!

Issue #19264, opened 2020-09-17
Use alternate names for blacklist/whitelist in prow


spiffxp:
**What should be cleaned up or changed**:
There are sundry files within prow that can probably use more specific terms, or the generic terms "allowlist/denylist"

```
$ ag -l whitelist\|blacklist prow | sort
prow/ANNOUNCEMENTS.md # ignore, deprecation announcements
prow/cmd/exporter/collector.go
prow/cmd/exporter/collector_test.go
prow/cmd/hook/main.go
prow/cmd/status-reconciler/main.go # offending flag has been deprecated
prow/cmd/status-reconciler/main_test.go
prow/config/config.go
prow/config/prow-config-documented.yaml
prow/hook/hook_test.go
prow/plugins/approve/approve_test.go
prow/plugins/blunderbuss/blunderbuss_test.go
prow/plugins/config.go
prow/plugins/lgtm/lgtm_test.go
prow/plugins/plugin-config-documented.yaml
prow/plugins/verify-owners/verify-owners.go
prow/plugins/verify-owners/verify-owners_test.go
prow/repoowners/repoowners.go
prow/repoowners/repoowners_test.go
```

EDIT: pulling up from https://github.com/kubernetes/test-infra/issues/19264#issuecomment-767913564

All instances of `whitelist` have been removed from prow (thanks @chases2!).  `blacklist` appears in two config fields that would need to be deprecated:

- [ ] `OwnersDirBlackList`, in the Prow config
- [ ] `LabelsBlacklist`, in the Prow plugins config

**Provide any links for context**:

/area prow
/wg naming
/sig testing

justaugustus:
Needs a rec proposed for replacement.
cc: @kubernetes/wg-naming

justaugustus:
For ADR:
/assign @celestehorgan 

For advice on impl:
/assign @justaugustus

justaugustus:
ADR up for review: https://github.com/kubernetes/community/pull/5341

chases2:
I went through the files in the prow dir and renamed the easier ones. I managed to remove all of the instances of "whitelist", but "blacklist" has two configurations that would need to be deprecated:

- [ ] `OwnersDirBlackList`, in the Prow config
- [ ] `LabelsBlacklist`, in the Prow plugins config

Also notable:
- The file ANNOUNCEMENTS.md contains deprecation announcements, which shouldn't be changed.
[...]

PR #20616, opened 2021-01-26
Use alternate internal comments/vars from whitelist/blacklist

chases2:
xref #19264

I tried to be context-sensitive in choosing new names, trying to best reflect the idea the comment/variable is trying to convey.

This PR doesn't cover files that change Prow Config options; those changes will require deprecations.

### DIFF
--- a/prow/apis/prowjobs/v1/types.go
+++ b/prow/apis/prowjobs/v1/types.go
@@ -259,10 +259,10 @@ func (rac *RerunAuthConfig) Validate() error {
 		return nil
 	}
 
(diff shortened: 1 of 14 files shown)


chases2:
Running `bazel test` in verify-owners fails flagrantly when I try to run it locally, but the same failures don't show up when Prow runs the tests (even without my changes).

Is the git stub being used in them not working locally? Or do I have some build artifact out of place?
`verify-owners_test.go:1242: Switching to master branch: /usr/bin/git [checkout master]: exit status 1, error: pathspec 'master' did not match any file(s) known to git`

michelle192837 (approved):

PR #20619, opened 2021-01-27
Add 'DefaultServiceAccountName' field to decoration config.

cjwagner:
This will allow us to specify a default K8s service account to apply to ProwJob pods that don't specify their own in the podspec. In particular, this will be used to default the service account to one with sufficient permissions (via GCP Workload Identity) for the pod utilities to upload results to GCS.
DD: https://docs.google.com/document/d/1VtamQuaR-TFVud3NjCRUV8Nqtp4WmBK1na7fNNiYHXU/edit?ts=600ae8e0&resourcekey=0-VXcyCZ9kcLJxQnnrUx4p0Q#heading=h.6dyyhz5krl9v

/assign @chaodaiG @alvaroaleman

### DIFF
--- a/prow/apis/prowjobs/v1/types.go
+++ b/prow/apis/prowjobs/v1/types.go
@@ -344,10 +344,13 @@ type DecorationConfig struct {
 	GCSConfiguration *GCSConfiguration `json:"gcs_configuration,omitempty"`
 	// GCSCredentialsSecret is the name of the Kubernetes secret
 	// that holds GCS push credentials.
-	GCSCredentialsSecret string `json:"gcs_credentials_secret,omitempty"`
+	GCSCredentialsSecret *string `json:"gcs_credentials_secret,omitempty"`
 	// S3CredentialsSecret is the name of the Kubernetes secret
 	// that holds blob storage push credentials.
-	S3CredentialsSecret string `json:"s3_credentials_secret,omitempty"`
+	S3CredentialsSecret *string `json:"s3_credentials_secret,omitempty"`
+	// DefaultServiceAccountName is the name of the Kubernetes service account
+	// that should be used by the pod if one is not specified in the podspec.
+	DefaultServiceAccountName *string `json:"default_service_account_name,omitempty"`
 	// SSHKeySecrets are the names of Kubernetes secrets that contain
 	// SSK keys which should be used during the cloning process.
 	SSHKeySecrets []string `json:"ssh_key_secrets,omitempty"`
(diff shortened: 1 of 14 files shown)


prow/apis/prowjobs/v1/types.go
```diff
@@ -344,10 +344,13 @@ type DecorationConfig struct {
 	GCSConfiguration *GCSConfiguration `json:"gcs_configuration,omitempty"`
 	// GCSCredentialsSecret is the name of the Kubernetes secret
 	// that holds GCS push credentials.
-	GCSCredentialsSecret string `json:"gcs_credentials_secret,omitempty"`
+	GCSCredentialsSecret *string `json:"gcs_credentials_secret,omitempty"`
 	// S3CredentialsSecret is the name of the Kubernetes secret
 	// that holds blob storage push credentials.
 	S3CredentialsSecret string `json:"s3_credentials_secret,omitempty"`
```
chaodaiG, line 350: nit, make this a pointer to make things consistent

prow/apis/prowjobs/v1/types_test.go
```diff
@@ -23,9 +23,13 @@ import (
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
-	"github.com/google/gofuzz"
+	fuzz "github.com/google/gofuzz"
 )
 
+func pStr(str string) *string {
+	return &str
```
chaodaiG, line 30: Initially thought this could cause invalid memory address as `str` is passed in as a copy of the value, so the entire memory address in only valid within the scope of this function. then later on figured out how go is smart on this https://www.ardanlabs.com/blog/2017/05/language-mechanics-on-escape-analysis.html, it's good to learn

prow/pod-utils/decorate/podspec.go
```diff
@@ -746,6 +746,11 @@ func decorate(spec *coreapi.PodSpec, pj *prowapi.ProwJob, rawEnv map[string]stri
 		spec.TerminationGracePeriodSeconds = &gracePeriodSeconds
 	}
 
+	defaultSA := pj.Spec.DecorationConfig.DefaultServiceAccountName
+	if spec.ServiceAccountName == "" && defaultSA != nil {
```
chaodaiG, line 750: Do we also need to do the `*defaultSA != ""` check like above?

chaodaiG (commented):
Looks good to me, left a few nits.

chaodaiG:
/test pull-test-infra-bazel
I think there is something going on with RBE since this morning, a little context: https://github.com/kubernetes/test-infra/pull/20626#issuecomment-768491717

prow/apis/prowjobs/v1/types.go
cjwagner: Good call, it's needed for more than just consistency. If someone wants to override a broadly scoped decoration config using S3 credentials in favor of a K8s service account with workload identity they'll need this in order to unset the S3 credentials.
Fixed.

fejta (approved):
/hold

cjwagner:
Word on the street is that the bazel issue is resolved.
/retest
/hold cancel

PR #20620, opened 2021-01-27
Add the sender field to IssueEvent

hi-rustin:
Currently we can't get the sender in the issue event, so add the sender field.

### DIFF
--- a/prow/github/types.go
+++ b/prow/github/types.go
@@ -663,7 +663,8 @@ type IssueEvent struct {
 	Issue  Issue            `json:"issue"`
 	Repo   Repo             `json:"repository"`
 	// Label is specified for IssueActionLabeled and IssueActionUnlabeled events.
-	Label Label `json:"label"`
+	Label  Label `json:"label"`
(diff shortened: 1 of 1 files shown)


alvaroaleman:
@hi-rustin as this has the potential to break pretty much all event processing, please do the due diligence and provide a link to the github docs to proof that this is correct and safe, so your reviewer doesn't have to do that for you

hi-rustin:
We can see the definition of sender in the documentation. For more information, see: https://docs.github.com/en/developers/webhooks-and-events/webhook-events-and-payloads#issues

And the definition of this sender is the same as in the [pull request event](https://docs.github.com/en/developers/webhooks-and-events/webhook-events-and-payloads#pull_request). So we can define it like in the pull request event.

https://github.com/kubernetes/test-infra/blob/b2d0efd2bb4afb200162cd78d669fd0294c6c247/prow/github/types.go#L236

alvaroaleman (approved):

hi-rustin:
@alvaroaleman  Thank you for your review.

Issue #20628, opened 2021-01-27
RBE bazel jobs are timing out


chaodaiG:
**What happened**:

There are several prow jobs started failing timing out after 2 hours, inspecting the logs and they were hanging at bazel test phase.

https://testgrid.k8s.io/sig-release-master-blocking#bazel-build-master
https://prow.k8s.io/?repo=kubernetes%2Ftest-infra&job=pull-test-infra-integration
https://prow.k8s.io/?repo=kubernetes%2Ftest-infra&job=pull-test-infra-bazel

One noticeable common factor among them is the usage of RBE

chaodaiG:
CC @fejta

fejta:
Lots of CI uses bazel, and bazel almost always uses RBE.

cjwagner:
Some more ongoing examples:
- https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/test-infra/20619/pull-test-infra-bazel/1354526643548327936
- https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/test-infra/20619/pull-test-infra-integration/1354526643657379840

chaodaiG:
Not using RBE in https://github.com/kubernetes/test-infra/pull/20626 can make `pull-test-infra-integration` job [pass](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/test-infra/20626/pull-test-infra-integration/1354541217169281024), while the same job is now consistently timing out in other PRs https://prow.k8s.io/?repo=kubernetes%2Ftest-infra&job=pull-test-infra-integration

CecileRobertMichon:
running into this in #20630

alvaroaleman:
This appears to block all merges into this repo, last merge was 9 hours ego and there is a bunch of merge-eligible PRs stuck in retesting: https://github.com/kubernetes/test-infra/pulls?q=is%3Apr+is%3Aopen+label%3Aapproved+label%3Algtm+-label%3A%22do-not-merge%2Fhold%22

justaugustus:
FYI @kubernetes/release-engineering

chaodaiG:
Current hypothesis is different bazel versions in k8s/test-infra and k8s/kubernetes caused confusion to RBE, and #20632 is an attempt to fix

chaodaiG:
Seeing tests reliably going through now:
[...]

PR #20632, opened 2021-01-27
Use bazel version 3.1.0, update variants to 3.4.1

fejta:
Once these images are pushed I should be able to bump to 3.4.1.


chaodaiG:
Please also update https://github.com/kubernetes/test-infra/blob/7e74b33534f950170801fb3ce015c1e96c975471/prow/test/integration/test.sh#L38 and https://github.com/kubernetes/test-infra/blob/7e74b33534f950170801fb3ce015c1e96c975471/prow/test/integration/test.sh#L42

images/bazel/variants.yaml
```diff
@@ -9,8 +9,8 @@ variants:
     OLD_VERSION: 0.29.1
   test-infra:
     CONFIG: test-infra # Used by test-infra repo
+    OLD_VERSION: 3.4.1
```
chaodaiG, line 12: Should 3.4.1 be `NEW_VERSION`?

fejta: Sure, although doesn't really matter

fejta:
> Please also update

Done

michelle192837 (approved):

chaodaiG (approved):

chaodaiG:
RBE disk quota increased, try again

chaodaiG:
`pull-test-infra-integration` is blocked by #20628 , which is what this PR is trying to fix. Overriding it to make this fix go through.
/override pull-test-infra-integration

chaodaiG:
Tide doesn't seem to handle the override very well in this scenario, overriding all to see if it'll work
/override pull-test-infra-verify-file-perms pull-test-infra-yamllint

chaodaiG:
Tide doesn't seem to be merging this PR even with all contextes overridden, manually merging the PR

cjwagner:
Tide works with the override plugin and is behaving as expected here. Tide prefers batches over serial merges so it won't merge an individual PR while a batch is being tested to avoid invalidating the batch. Since the bazel jobs are taking 2h15mins to complete the batches are blocking the merge queue for that amount of time.
https://prow.k8s.io/?repo=kubernetes%2Ftest-infra&type=batch

PR #20635, opened 2021-01-28
Automatically determine bazel version to install

fejta:
We need to run whatever is in `.bazelversion`, not 3.0.0


chaodaiG (approved):

matthyx:
@prow YUNo merge?

fejta:
Probably because of https://prow.k8s.io/?repo=kubernetes%2Ftest-infra&type=batch

Issue #20542, opened 2021-01-20
Clean up sig-windows-signal 


jayunit100:
**What should be cleaned up or changed**:

```
https://testgrid.k8s.io/sig-windows-releases
```

Has ~ 8 tests that havent passed, in any recent history.

Lets remove the GCE tests from there, or else, see if folks from GCE wanted to fix them ? 

**Provide any links for context**:

https://testgrid.k8s.io/sig-windows-releases

Our goal here is to have a canonical endpoint (https://testgrid.k8s.io/sig-windows-releases) where we can go as the windows sig which give us clear test signal.  

This is espeically important as 
- containerd windows jobs get more common and we remove dockershim,  
- CSI proxy goes GA, 
- multiple CNIs with network policies begin to get integrated into jobs 

Is this ok with you @pjh  @jeremyje

pjh:
Consistently failing tests that I see today under the GCE jobs at https://testgrid.k8s.io/sig-windows-releases:

Kubernetes e2e suite.[sig-network] Services should be able to change the type from ExternalName to ClusterIP [Conformance]
Kubernetes e2e suite.[sig-network] Services should be able to change the type from ExternalName to NodePort [Conformance]
Kubernetes e2e suite.[sig-network] Services should be able to create a functioning NodePort service [Conformance]

Since those are part of the [Conformance] suite, is removing them possible/advisable? Our team needs to look into the failures, not sure how long it will take. If removing them temporarily from these dashboards would be useful then I'm fine with that.

@jayunit100 are there more failing tests that you see somewhere?

Also, looks like gce-windows-1.17 job is consistently failing to start - we'll look at that too.

pjh:
It looks like some "Kubernetes e2e suite.[sig-storage] In-tree Volumes [Driver: windows-gcepd] ..." tests are mostly (but not entirely) failing, as also mentioned in Slack.

This test is also mostly-failing: [sig-cli] Kubectl client Guestbook application should create and stop a working application [Conformance]

neolit123:
> https://testgrid.k8s.io/sig-windows-releases

fix for `kubeadm-windows-gcp-k8s-stable` is here:
https://github.com/kubernetes-sigs/sig-windows-tools/pull/140

claudiubelu:
> 
> 
> Consistently failing tests that I see today under the GCE jobs at https://testgrid.k8s.io/sig-windows-releases:
> 
[...]

PR #20653, opened 2021-01-29
Organize the sig-windows board

jsturtevant:
As part of [cleaning up the Windows tests](https://github.com/kubernetes/test-infra/issues/20542) we discussed in our sig meeting this week of creating a smaller dashboard that would be the "signal" for windows. We discussed keeping this dashboard smaller than the current https://testgrid.k8s.io/sig-windows-releases board and having criteria for how a test can be added to the new dash.

This is a proposal for adding only our `release-informing` (https://testgrid.k8s.io/sig-release-master-informing) jobs initially to a new `sig-windows-signal` dashboard so we can make these consistently green and then evaluate adding other tests (maybe presubmits or post submits?)

Going to hold for comment from for others who maintain tests: @jayunit100 @ravisantoshgudimetla @pjh @jeremyje @marosset @claudiubelu @chewong 

/hold

fyi - I dropped our 1.17 tests as it is now longer actively maintained: https://github.com/kubernetes/sig-release/blob/master/releases/patch-releases.md#non-active-branch-history


config/jobs/kubernetes/sig-windows/windows-gce.yaml
```diff
@@ -92,7 +92,7 @@ periodics:
   annotations:
     fork-per-release: "true"
     fork-per-release-replacements: "--extract=ci/latest -> --extract=ci/latest-{{.Version}}"
-    testgrid-dashboards: google-windows, sig-windows-gce, sig-windows-releases, sig-release-master-informing
+    testgrid-dashboards: google-windows, sig-windows-gce, sig-release-master-informing, sig-windows-signal
```
jsturtevant, line 95: Do we want to move the GCE configs into the ` config/jobs/kubernetes-sigs/sig-windows` folder? I think this is the folder where sig jobs are prefered to live.

config/testgrids/kubernetes/sig-windows/config.yaml
```diff
@@ -10,6 +11,7 @@ dashboard_groups:
     - sig-windows-networking
 
 dashboards:
+- name: sig-windows-signal
```
jsturtevant, line 14: I called it `signal` so we could add non informing jobs to it later if we wanted.  Not tied to the name

config/jobs/kubernetes-sigs/sig-windows/release-1.17-windows.yaml
```diff
@@ -1,227 +0,0 @@
-presubmits:
```
chewong, line 1: Is 1.17 deprecated now?

jsturtevant: https://github.com/kubernetes/sig-release/blob/master/releases/patch-releases.md#non-active-branch-history

chewong: Didn't see the comment above

chewong (approved):
/lgtm

config/jobs/kubernetes/sig-windows/windows-gce.yaml
jsturtevant: @jeremyje @pjh Any thoughts on if these should be consolidated?

PR #20656, opened 2021-01-29
Add support for alternate default_decoration_configs format that allows defaulting by cluster.

cjwagner:
I'm opening this as WIP because I still need to add some more tests and docs.

Design Doc: https://docs.google.com/document/d/1VtamQuaR-TFVud3NjCRUV8Nqtp4WmBK1na7fNNiYHXU/edit?resourcekey=0-VXcyCZ9kcLJxQnnrUx4p0Q


cjwagner:
I don't know why I'm getting this error, I didn't touch anything related to this field AFAICT and this field was already unexported...
```
--- FAIL: TestGeneratePeriodics (0.00s)
panic: cannot handle unexported field at {[]config.Periodic}[0].interval:
	"k8s.io/test-infra/prow/config".Periodic
consider using a custom Comparer; if you control the implementation of type, you can also consider using an Exporter, AllowUnexported, or cmpopts.IgnoreUnexported [recovered]
	panic: cannot handle unexported field at {[]config.Periodic}[0].interval:
	"k8s.io/test-infra/prow/config".Periodic
consider using a custom Comparer; if you control the implementation of type, you can also consider using an Exporter, AllowUnexported, or cmpopts.IgnoreUnexported
goroutine 123 [running]:
testing.tRunner.func1.1(0x2302dc0, 0xc0001c9810)
	GOROOT/src/testing/testing.go:1057 +0x46a
testing.tRunner.func1(0xc00050e900)
	GOROOT/src/testing/testing.go:1060 +0x636
panic(0x2302dc0, 0xc0001c9810)
	GOROOT/src/runtime/panic.go:975 +0x3e9
github.com/google/go-cmp/cmp.validator.apply(0xc000453c20, 0x2476fc0, 0xc0001b4128, 0x1a6, 0x2476fc0, 0xc0000ecba8, 0x1a6)
	external/com_github_google_go_cmp/cmp/options.go:244 +0x70e
github.com/google/go-cmp/cmp.(*state).tryOptions(0xc000453c20, 0x2967f80, 0x2476fc0, 0x2476fc0, 0xc0001b4128, 0x1a6, 0x2476fc0, 0xc0000ecba8, 0x1a6, 0x0)
	external/com_github_google_go_cmp/cmp/compare.go:303 +0x15a
github.com/google/go-cmp/cmp.(*state).compareAny(0xc000453c20, 0x29417c0, 0xc000590700)
	external/com_github_google_go_cmp/cmp/compare.go:258 +0x434
github.com/google/go-cmp/cmp.(*state).compareStruct(0xc000453c20, 0x2967f80, 0x2482860, 0x2482860, 0xc0001b4000, 0x199, 0x2482860, 0xc0000eca80, 0x199)
	external/com_github_google_go_cmp/cmp/compare.go:427 +0x58e
github.com/google/go-cmp/cmp.(*state).compareAny(0xc000453c20, 0x2941780, 0xc00007c900)
	external/com_github_google_go_cmp/cmp/compare.go:286 +0x17a8
github.com/google/go-cmp/cmp.(*state).statelessCompare(0xc000453c20, 0x2941780, 0xc00007c900, 0x2967f80, 0x2967f80)
	external/com_github_google_go_cmp/cmp/compare.go:229 +0xfc
github.com/google/go-cmp/cmp.(*state).compareSlice.func2(0x0, 0x0, 0x0, 0x353a1c0)
	external/com_github_google_go_cmp/cmp/compare.go:493 +0x105
github.com/google/go-cmp/cmp/internal/diff.Difference(0x5, 0x5, 0xc000597210, 0x5, 0x5, 0xc000041880)
	external/com_github_google_go_cmp/cmp/internal/diff/diff.go:241 +0x2b4
github.com/google/go-cmp/cmp.(*state).compareSlice(0xc000453c20, 0x2967f80, 0x22c8f20, 0x22c8f20, 0xc00000e340, 0x97, 0x22c8f20, 0xc00000e360, 0x97)
	external/com_github_google_go_cmp/cmp/compare.go:492 +0x852
github.com/google/go-cmp/cmp.(*state).compareAny(0xc000453c20, 0x2932480, 0xc0002a5ec0)
	external/com_github_google_go_cmp/cmp/compare.go:288 +0x117f
github.com/google/go-cmp/cmp.Diff(0x22c8f20, 0xc00000e340, 0x22c8f20, 0xc00000e360, 0x0, 0x0, 0x0, 0x6, 0x0)
	external/com_github_google_go_cmp/cmp/compare.go:119 +0x105
k8s.io/apimachinery/pkg/util/diff.legacyDiff(...)
	external/io_k8s_apimachinery/pkg/util/diff/diff.go:51
k8s.io/apimachinery/pkg/util/diff.ObjectDiff(...)
	external/io_k8s_apimachinery/pkg/util/diff/diff.go:58
k8s.io/test-infra/releng/config-forker.TestGeneratePeriodics(0xc00050e900)
	releng/config-forker/main_test.go:704 +0x19b2
testing.tRunner(0xc00050e900, 0x2657c78)
	GOROOT/src/testing/testing.go:1108 +0x203
created by testing.(*T).Run
	GOROOT/src/testing/testing.go:1159 +0x797
```

cjwagner:
Looks like the diff handling was already broken in that test case, but only exposed as a problem if the test case fails. I fixed that.

cjwagner:
More apparently unrelated errors for boilerplate in files I didn't touch...
```
2 files have incorrect boilerplate headers:
prow/test/integration/test/crier_test.go
prow/test/integration/test/overall_test.go
```
It is unclear to me why this started failing now, but didn't fail on the PR that introduced the boilerplate: https://github.com/kubernetes/test-infra/pull/20521

prow/config/config.go
```diff
@@ -499,16 +508,123 @@ type Plank struct {
...
-	def := p.DefaultDecorationConfigs["*"]
-	if dcByRepo, ok := p.DefaultDecorationConfigs[repo]; ok {
-		return dcByRepo.ApplyDefault(def)
+type DefaultDecorationConfigEntry struct {
+	// Matching/filtering fields
+
+	// Repo matches against the "org/repo" that the presubmit or postsubmit is
+	// associated with. If the job is a periodic, extra_refs[0] is used. If the
+	// job is a periodic without extra_refs, the empty string will be used.
+	Repo *regexp.Regexp `json:"-"`
```
chaodaiG, line 517: Is there a particular reason must use regex instead of existing `*` style? Ask user to learn something new is fine imo as long as it's well communicated. I'm interested in the reasoning behind this change.

releng/config-forker/main.go
```diff
@@ -104,6 +104,13 @@ func generatePresubmits(c config.JobConfig, version string) (map[string][]config
 	return newPresubmits, nil
 }
 
+func shouldDecorate(c *config.JobConfig, util config.UtilityConfig) bool {
+	if util.Decorate != nil {
+		return *util.Decorate
+	}
+	return c.DecorateAllJobs
```
chaodaiG, line 111: why not do the same as https://github.com/kubernetes/test-infra/pull/20656/files?file-filters%5B%5D=.go#diff-88e6b42a331a7aa9260f9a830961e3ba3c77e0e6d781465378b510d7abc1517cR1364?

chaodaiG (commented):
Generally looks good to me. Will come back again when there are more unit tests and documentation

prow/config/config.go
cjwagner: I don't think there is an existing `*` (glob? wildcard?) matching style. The global config used key `*`, but that was just a special case literal match.
I think regexp matching is overkill here, but I figured it would be most appropriate for consistency with other Prow config fields. `run_if_changed`, `branches`, `exclude_branches`, `trigger`, etc. all use regular expressions rather than `*` matching. 

I also considered naming this field `OrgRepo` and just checking for string equality with the org or org/repo. This is consistent with other parts of Prow config, but I thought it more important that all the matching/filtering fields in this struct use a consistent matching system. I think it could be confusing (especially if we add more filter fields) if some things are matching literally (possibly against multiple fields), some are using `*` matching, and some are using regexp.
WDYT?

cjwagner (commented):
> Will come back again when there are more unit tests and documentation

Thanks, I ran into some more unrelated issues last night, but I have the majority of the tests done now. I'll notify when this is ready for another review.

cjwagner:
This is ready for review now.

prow/config/prow-config-documented.yaml
```diff
@@ -472,139 +472,10 @@ owners_dir_blacklist:
     repos:
         "": null
 plank:
-    # DefaultDecorationConfigs holds the default decoration config for specific values.
```
chaodaiG, line 475: Don't we still want to keep the documentation for these fields?

releng/config-forker/main.go
chaodaiG: I see, thank you for the context. I think it would be nice to make the generated config more explicit for better visibility, unless it somehow has a problem

chaodaiG (commented):
Looks great! Unit test looks comprehensive to me, and documentations look accurate. Left a couple of comments

releng/config-forker/main.go
cjwagner: I agree, but I think that should be done in a separate PR if is is desired. That code is [owned by releng](https://github.com/kubernetes/test-infra/blob/492140b4514ce0b410cb3f0681d0941a67dbc394/OWNERS_ALIASES#L229-L235) so we should get their approval for any behavioral changes, which I don't want to mix into this large PR.

prow/config/config.go
cjwagner: > And for cluster filed, I think the only reason using regex instead of literal is because of global default right?

I was thinking it might be nice to reduce config size by using the cluster names to apply config across common clusters. e.g. All clusters with `-trusted` suffix or all clusters containing `istio`. This isn't that important though.

/assign @alvaroaleman 
WDYT Alvaro?

prow/config/prow-config-documented.yaml
cjwagner: This is autogenerated and I don't think it can be easily fixed since the field now has type `json.RawMessage`. I don't think there is a good way to display 2 alternative formats or a good way to indicate that to the generator.

stevekuznetsov (commented):
My quick read says no, but does this change affect a round-trip deserialize/default/serialize loop on job configuration? We load, format and dump all config in a presubmit and I don't want to see that output end up with resolved defaults in every job.

cjwagner:
> My quick read says no, but does this change affect a round-trip deserialize/default/serialize loop on job configuration? We load, format and dump all config in a presubmit and I don't want to see that output end up with resolved defaults in every job.

@stevekuznetsov We already resolve the default to every job at load time today: https://github.com/kubernetes/test-infra/blob/5fd357428d61beb431d99870fc79e7ecb64b9be2/prow/config/config.go#L1251-L1256
Do you not see this behavior already?

alvaroaleman:
> Do you not see this behavior already?

Currently, we have code that loads -> Serializes- > Writes to disk, which is how everything post-defaulting ends up getting written out. That won't work anymore after this change, we will just directly write back to disk what we put into the field.

prow/config/config.go
```diff
@@ -499,16 +508,137 @@ type Plank struct {
...
+			errs = append(errs, err)
+		}
+	}
+	return utilerrors.NewAggregate(errs)
+}
+
+// DefaultDecorationMapToSliceTesting is a convenience function that is exposed
+// to allow unit tests to convert the old map format to the new slice format.
+// It should only be used in testing.
+func DefaultDecorationMapToSliceTesting(m map[string]*prowapi.DecorationConfig) []*DefaultDecorationConfigEntry {
```
alvaroaleman, line 611: Nit: Can we have a `Must` somewhere in the name so it becomes clear that this might panic?

prow/config/config.go
```diff
@@ -1293,26 +1425,27 @@ func defaultPresubmits(presubmits []Presubmit, c *Config, repo string) error {
...
+// defaultPeriodics defaults c.Periodics
+func defaultPeriodics(c *Config) error {
+	c.defaultPeriodicFields(c.Periodics)
 	var errs []error
-	c.defaultPeriodicFields(periodics)
-	for _, periodic := range periodics {
-		if err := resolvePresets(periodic.Name, periodic.Labels, periodic.Spec, c.Presets); err != nil {
+	for i := range c.Periodics {
+		setPeriodicDecorationDefaults(c, &c.Periodics[i])
+		if err := resolvePresets(c.Periodics[i].Name, c.Periodics[i].Labels, c.Periodics[i].Spec, c.Presets); err != nil {
```
alvaroaleman, line 1448: What, this never worked? :D

prow/config/config_test.go
```diff
@@ -3584,12 +3653,369 @@ func TestRefGetterForGitHubPullRequest(t *testing.T) {
...
+			}
+		})
+	}
+}
+
+// complexConfig is shared by multiple test cases that test DefaultDecorationConfig
+// merging logic. It configures the upload bucket based on the org/repo and
+// uses either a GCS secret or k8s SA depending on the cluster.
+// A specific 'override' org overrides some fields in the trusted cluster only.
+var complexConfig = &Config{
```
alvaroaleman, line 3938: small nit: Its nicer to put something like this behind a getter func (`func complexConfig() *Config`) to avoid sharing state between tests

prow/config/config_test.go
```diff
@@ -4808,30 +5468,56 @@ func TestInRepoConfigAllowsCluster(t *testing.T) {
 	}
 }
 
-func TestGetDefaultDecorationConfigsThreadSafety(t *testing.T) {
-	const repo = "repo"
-	p := Plank{DefaultDecorationConfigs: map[string]*prowapi.DecorationConfig{
-		"*": {
-			GCSConfiguration: &prowapi.GCSConfiguration{
-				MediaTypes: map[string]string{"text": "text"},
+// func TestFinalizeDefaultDecorationConfigs(t *testing.T) {
```
alvaroaleman, line 5471: ?

prow/cmd/deck/pr_history.go
```diff
@@ -227,7 +227,11 @@ func getStorageDirsForPR(c *config.Config, gitHubClient deckGitHubClient, gitCli
 			gcsConfig = presubmit.DecorationConfig.GCSConfiguration
 		} else {
 			// for undecorated jobs assume the default
-			gcsConfig = c.Plank.GetDefaultDecorationConfigs(fullRepo).GCSConfiguration
+			def := c.Plank.MergeDefaultDecorationConfig(fullRepo, presubmit.Cluster, nil)
```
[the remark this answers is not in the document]
cjwagner: Merging is done at load time. This is a special case where the job is not decorated at all (so there was nothing to merge) and we are attempting to guess a reasonable default config in order to make assumptions about blob storage upload location.

prow/cmd/plank/README.md
```diff
@@ -21,7 +21,9 @@ plank:
   job_url_template: 'https://<domain>/view/<bucket-name>/pr-logs/pull/{{.Spec.Refs.Repo}}/{{with index .Spec.Refs.Pulls 0}}{{.Number}}{{end}}/{{.Spec.Job}}/{{.Status.BuildID}}'
   report_template: '[Full PR test history](https://<domain>/pr-history?org={{.Spec.Refs.Org}}&repo={{.Spec.Refs.Repo}}&pr={{with index .Spec.Refs.Pulls 0}}{{.Number}}{{end}})'
   default_decoration_configs:
-    '*':
+  # All entries that match a job are used, later entries override previous values.
```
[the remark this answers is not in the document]
cjwagner: Yeah, the alternative would be for both formats to be included under different names which wouldn't be ideal either. To be honest the generated config doc doesn't seem to provide any value over what the godocs provide since all the field values are empty. Godocs + real world examples or samples are what I would reference today.

prow/config/config.go
```diff
@@ -484,11 +484,20 @@ type Plank struct {
...
-	// Use `org/repo`, `org` or `*` as a key.
-	DefaultDecorationConfigs map[string]*prowapi.DecorationConfig `json:"default_decoration_configs,omitempty"`
+	// Each entry in the slice specifies Repo and Cluster regexp filter fields to
+	// match against jobs and a corresponding DecorationConfig. All entries that
+	// match a job are used. Later matching entries override the fields of earlier
+	// matching entries.
+	// Alternatively this field may be type `map[string]*prowapi.DecorationConfig`
+	// Use `org/repo`, `org` or `*` as a key to match against jobs. (Periodics
+	// use extra_refs[0] for matching if present.)
+	DefaultDecorationConfigs []*DefaultDecorationConfigEntry `json:"-"`
```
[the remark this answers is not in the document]
cjwagner: The godocs detail the types that are accepted. I wanted to use a single field to avoid allowing multiple sources of truth, but we could enforce mutual exclusivity with additional validation if that is preferred. I don't think having separate top level config fields for each type will make the code simpler to read, but I don't feel strongly about it and am happy to change it if preferred.

prow/config/config.go
```diff
@@ -499,16 +508,137 @@ type Plank struct {
...
+
+// Matches returns true iff all the filters for the entry match a job.
+func (d *DefaultDecorationConfigEntry) Matches(repo, cluster string) bool {
+	return d.Repo.MatchString(repo) && d.Cluster.MatchString(cluster)
+}
+
+// MergeDefaultDecorationConfig finds all matching DefaultDecorationConfigEntry
+// for a job and merges them sequentially before merging into the job's own
+// DecorationConfig. Configs merged later override values from earlier configs.
+func (p *Plank) MergeDefaultDecorationConfig(repo, cluster string, jobDC *prowapi.DecorationConfig) *prowapi.DecorationConfig {
```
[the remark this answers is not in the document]
cjwagner: Merging is done once on load, this is exposed for special use cases where we need a decoration config for a job that didn't request decoration. That is needed to make assumptions about the upload location.

prow/config/config.go
cjwagner: I think it worked. The `periodic.Spec` field is all that `resolvePresets` modifies and it is a pointer type. I only changed this function signature because `setPeriodicDecorationDefaults` needs it and I wanted them to share the for loop for conciseness.

cjwagner:
>  I don't want to see that output end up with resolved defaults in every job.

> Currently, we have code that loads -> Serializes- > Writes to disk, which is how everything post-defaulting ends up getting written out.

I'm confused as to whether or not you want the resolved decoration config to be included on jobs after reserialization. If I understand correctly, Steve's response suggests that it shouldn't be and Alvaro's suggests that it should. Based on the existing code I would expect it to include the resolved decoration config today and I don't think this PR changes that behavior.

prow/config/config.go
cjwagner: Ok that all makes good sense to me.
What are your thoughts on the cluster field then? I had hoped to use a consistent matching system, but the more I think about it, the more I'm thinking we should just have `cluster` be a literal string match, at least initially. That is simpler and likely sufficient for existing use cases. With this new config slice pattern we can easily add more matching fields like `repo_re` or `cluster_re` down the road if we find a need for them.

cjwagner:
Thank you for the detailed reviews. I have responded to all comments (but haven't made changes yet).

prow/cmd/plank/README.md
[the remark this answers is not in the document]
alvaroaleman: The generated config contains the very same docs the godoc fields have while using the json field names, so IMHO it is a lot more useful than godocs.

prow/config/config.go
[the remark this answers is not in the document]
alvaroaleman: Well right now its not enough to look at the types, you need to look at the loading code to understand the possible formats. Having two mutually exclusive fields would restore the ability to just look at the type to understand the format, so I would argue its simpler or WDYT?

prow/config/config.go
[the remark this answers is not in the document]
alvaroaleman: So from what I can see (please tell me if I missed something!) this function has zero callsites outside of the `prow/config` package that call it with a non-nil `jobDC` which makes it look like its currently conflating two seperate things: 1) Merge configs during load 2) Look up config for a given repo+cluster

prow/config/config.go
alvaroaleman: I don't really have an opinion regarding regex or not for `cluster`. It makes IMHO more sense than using it for `repo`, because for clusters you probably could have some kind of naming scheme that's based on trust domain but at the end of the day it boils down to what you need there and you know that better than me

alvaroaleman:
> I'm confused as to whether or not you want the resolved decoration config to be included on jobs after reserialization. If I understand correctly, Steve's response suggests that it shouldn't be and Alvaro's suggests that it should. Based on the existing code I would expect it to include the resolved decoration config today and I don't think this PR changes that behavior.

For the main prow config, we do Prows config load -> Some tweaks -> Serialize + write, for the jobs we just do a `yaml.Unmarshal` -> `yaml.Marshal`. This means that today the defaulting is not included after reserialization, because it only happens on job level. After this change, the "change to new format" would be included because it happens in the main config (But I don't think it's possible to avoid that)

cjwagner:
> After this change, the "change to new format" would be included because it happens in the main config (But I don't think it's possible to avoid that)

The change to new format would not be included as currently written because the new format field is not serialized. Only the raw JSON is serialized back. If I switch to having separate fields for the different config formats as requested in the other comments the new format would then be used in the serialized output rather than using whatever was originally specified.

So if serializing back to the same format that was originally specified is important to you I think we need to continue using a single raw JSON field rather than 2 fields with the different accepted types. I'm happy to go with whichever option you prefer, but I think we need to decide whether consistent reserialization or clearer generated config documentation is more important.

prow/config/config.go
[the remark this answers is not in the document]
cjwagner: FWIW you don't need to look at the loading code, the accepted types are included in the godoc. Still happy to change if you prefer, but now I think this might be problematic for your reserialization requirements so please confirm: https://github.com/kubernetes/test-infra/pull/20656#issuecomment-785428469

prow/config/config.go
cjwagner: Ok I'm going to go with literal matching then, I think that will be sufficient for now. I like the idea of expanding functionality with something like `cluster_re` if we find that we need alternative matching systems for any of these fields later on.

alvaroaleman:
> So if serializing back to the same format that was originally specified is important to you I think we need to continue using a single raw JSON field rather than 2 fields with the different accepted types. I'm happy to go with whichever option you prefer, but I think we need to decide whether consistent reserialization or clearer generated config documentation is more importa

Well, so the actual reason why we do the reserialization is to get ordered yaml back, which we won't when using `json.RawMessage` anyways as that preserves just the original content as []byte so I would prefer having both fields but making them mutually exclusive.

cjwagner:
PTAL, I think the way I implemented this preserves the reserialization behavior while still using 2 separate fields to make the generated docs more clear.
Only the newest commit contains substantive code changes since the last review, other commits were just rebased.

alvaroaleman:
Sorry, a bit underwater atm, will look early next week

alvaroaleman:
Yeah, the way it's implemented now looks good :) There are still a few outstanding comments from a previous review, though

prow/config/config.go
cjwagner: This can no longer panic after removing the regexp compilation.

cjwagner:
> There are still a few outstanding comments from a previous review, though

My apologies, I forgot to implement those changes. I believe I've addressed everything now.

cjwagner:
@chaodaiG @alvaroaleman PTAL. Changes since the last review are in their own commit for convenience.

alvaroaleman (approved):
/lgtm

Looks good, thanks for all the work! Could you squash it though (or apply the tide squash label)?

chaodaiG (approved):

PR #20663, opened 2021-01-29
ci-kubernetes-build-canary: Migrate from bootstrap to krel

justaugustus:
The [`no-bootstrap` job](https://prow.k8s.io/job-history/gs/kubernetes-jenkins/logs/ci-kubernetes-build-no-bootstrap) has been running successfully for some time, so
let's see if we can promote it to the canary phase.

There are some recent failures that seem related to recognizing buildx,
so the `--configure-docker` is added here to hopefully clear that up.

/assign @hasheddan @cpanato @saschagrunert 

ref: https://github.com/kubernetes/release/issues/1711#issuecomment-767167222
cc: @spiffxp @kubernetes/ci-signal


hasheddan (approved):
/lgtm

PR #20665, opened 2021-01-29
prep kubernetes/org jobs for default branch rename

spiffxp:
specificalky
- add main to any pre/postsubmit triggers that reference master
- try removing explicit master base_ref, see if that gets us default branch

Part of https://github.com/kubernetes/org/issues/2466

### DIFF
--- a/config/jobs/kubernetes/test-infra/test-infra-trusted.yaml
+++ b/config/jobs/kubernetes/test-infra/test-infra-trusted.yaml
@@ -4,6 +4,7 @@ postsubmits:
     cluster: test-infra-trusted
     decorate: true
     branches:
+    - ^main$
     - ^master$
     max_concurrency: 1
     spec:


spiffxp:
My read of the clone part of pod-utils is that they assume `base_ref` is present.  https://github.com/kubernetes/test-infra/blob/94c2ba78474027a100b2da7a733f6a1fd5eef930/prow/pod-utils/clone/clone.go#L203-L229

It's not clear to me whether something further up the stack does something to default base_ref if not set.

Locally
- `git fetch origin ''` is equivalent to fetching all branches, and `git checkout FETCH_HEAD` puts me on whichever branch was alphabetically first
- `git branch --force "" FETCH_HEAD` errors with `fatal: '' is not a valid branch name.`

Tests are happy though!

spiffxp:
For purposes of migrating kubernetes/org today, I'm going to defer further experiments on making periodics resilient to remote branch renames to a job that runs more frequently than 24h.

Putting the `base_ref` back...

Issue #20606, opened 2021-01-26
Kettle seeing "Bad Request" in BQ insert 


MushuEE:
**What happened**:
Kettle staging is seeing `Bad Requests` when trying to stream data to BQ, this is not the behavior seen within Prod.
 
```
Traceback (most recent call last):
  File "stream.py", line 318, in <module>
    stop=StopWhen(OPTIONS.stop_at))
  File "stream.py", line 212, in main
    emitted = insert_data(bq_client, table, make_json.make_rows(db, builds))
  File "stream.py", line 132, in insert_data
    errors = retry(bq_client.insert_rows, table, chunk, skip_invalid_rows=True)
  File "stream.py", line 96, in retry
    return func(*args, **kwargs)
  File "/usr/local/lib/python3.6/dist-packages/google/cloud/bigquery/client.py", line 2869, in insert_rows
    return self.insert_rows_json(table, json_rows, **kwargs)
  File "/usr/local/lib/python3.6/dist-packages/google/cloud/bigquery/client.py", line 3016, in insert_rows_json
    timeout=timeout,
  File "/usr/local/lib/python3.6/dist-packages/google/cloud/bigquery/client.py", line 636, in _call_api
    return call()
  File "/usr/local/lib/python3.6/dist-packages/google/api_core/retry.py", line 286, in retry_wrapped_func
    on_error=on_error,
  File "/usr/local/lib/python3.6/dist-packages/google/api_core/retry.py", line 184, in retry_target
    return target()
  File "/usr/local/lib/python3.6/dist-packages/google/cloud/_http.py", line 438, in api_request
    raise exceptions.from_http_response(response)
google.api_core.exceptions.BadRequest: 400 POST https://bigquery.googleapis.com/bigquery/v2/projects/k8s-gubernator/datasets/build/tables/staging/insertAll?prettyPrint=false: <!DOCTYPE html>
<html lang=en>
  <meta charset=utf-8>
  <meta name=viewport content="initial-scale=1, minimum-scale=1, width=device-width">
  <title>Error 400 (Bad Request)!!1</title>
  <style>
[...]

PR #20666, opened 2021-01-29
[Kettle] Decrease Max chunk size and catch Over Size error

MushuEE:
We are getting a cryptic message in the streaming process for BQ inserts:

See #20606

Based on https://github.com/googleapis/google-cloud-go/issues/2855

It seems this can happen if the Payload is far too large. It might be worth attempting to split the chunks into "byte sized" pieces but for a quicker fix I would like to just make this change. 

/area kettle
/sig testing
/cc @spiffxp

### DIFF
--- a/kettle/stream.py
+++ b/kettle/stream.py
@@ -29,6 +29,7 @@ import time
 import multiprocessing.pool
 
 try:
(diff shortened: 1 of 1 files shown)


spiffxp (approved):
/lgtm

PR #20675, opened 2021-01-29
Prep kubernetes/k8s.io for default branch rename

spiffxp:
Part of https://github.com/kubernetes/k8s.io/issues/1597

/hold
The periodic changes here are disruptive, they will break until the branch is renamed. Will remove the hold when it's not late on a friday.


saschagrunert (approved):
/lgtm

PR #20679, opened 2021-01-30
[tide] Allow PRs to be picked for batch while still pending tests

arturo-skydio:
Allows tide to select PRs without the need for a first successful presubmit(s) build, this is intended to increase the likelihood of batch PRs. This logic disqualifies any failed individual builds and just include pending or passing.

Primary usecase is where teams have quick fixes that can be added to batch while still pending their individual builds (where build times are longer than usual).

Introducing a boolean config `batchAllowPending`, if set to `true` while enable this functionality.

### DIFF
--- a/prow/tide/tide.go
+++ b/prow/tide/tide.go
@@ -778,6 +778,23 @@ func isPassingTests(log *logrus.Entry, ghc githubClient, pr PullRequest, cc cont
 	return len(unsuccessful) == 0
 }
 
+// isFailingTests returns whether one or more contexts are failing.
+func isFailingTests(log *logrus.Entry, ghc githubClient, pr PullRequest, cc contextChecker) bool {
+	log = log.WithFields(pr.logFields())
+	contexts, err := headContexts(log, ghc, &pr)
+	if err != nil {
+		log.WithError(err).Error("Getting head commit status contexts.")
+		// If we can't get the status of the commit, assume that it is failing.
+		return true
+	}
+	for _, ctx := range unsuccessfulContexts(contexts, cc, log) {
+		if ctx.State == githubql.StatusStateFailure || ctx.State == githubql.StatusStateError {
+			return true
+		}
+	}
+	return false
+}
+
 // unsuccessfulContexts determines which contexts from the list that we care about are
 // failed. For instance, we do not care about our own context.
 // If the branchProtection is set to only check for required checks, we will skip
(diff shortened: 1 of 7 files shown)


petr-muller:
/ok-to-test

The code looks straightforward, but what's the use case? If the PR is failing tests what's the point of including in a batch?

arturo-skydio:
> /ok-to-test
> 
> The code looks straightforward, but what's the use case? If the PR is failing tests what's the point of including in a batch?

Hi, thanks for reviewing, added some description to the PR, not sure if this usecase applies to other teams using Prow but for us it was something that helped with our builds being around 2 hours

prow/tide/tide.go
```diff
@@ -1011,9 +1031,13 @@ func (c *Controller) pickBatch(sp subpool, cc map[int]contextChecker) ([]PullReq
 	// we must choose the oldest PRs for the batch
 	sort.Slice(sp.prs, func(i, j int) bool { return sp.prs[i].Number < sp.prs[j].Number })
 
+	batchSkipTests := false
+	if c.config().Tide.BatchSkipTests != nil && *c.config().Tide.BatchSkipTests {
+		batchSkipTests = true
```
petr-muller, line 1036:
```suggestion
	var batchSkipTests bool
	if c.config().Tide.BatchSkipTests != nil {
		batchSkipTests = *c.config().Tide.BatchSkipTests 
```

is IMO a bit more idiomatic and less wordy, but I would not even be afraid of doing this in a single line (`bst := (ptr != nil && *ptr`)

prow/tide/tide.go
```diff
@@ -1011,9 +1031,13 @@ func (c *Controller) pickBatch(sp subpool, cc map[int]contextChecker) ([]PullReq
...
+	}
 	var candidates []PullRequest
 	for _, pr := range sp.prs {
-		if isPassingTests(sp.log, c.ghc, pr, cc[int(pr.Number)]) {
+		if (batchSkipTests && isPendingTests(sp.log, c.ghc, pr, cc[int(pr.Number)])) || isPassingTests(sp.log, c.ghc, pr, cc[int(pr.Number)]) {
```
petr-muller, line 1040: I think that requiring all checks to be pending may be too strict when there are jobs that run very fast. I think you'd want all checks to be either passing or pending?

arturo-skydio: As far as I understand it I'm only checking to see if there are failing tests because I'm looping over unsuccessfulContext pr's which includes failures and pending, so maybe whats needed here is a better function name like `failingTests` ?

prow/tide/tide.go
arturo-skydio: will try the one liner thanks!

prow/tide/tide.go
arturo-skydio: Renamed to `isFailingTests` , let me know if that works

petr-muller: Yeah, thanks for explanation, I misread the function :)

petr-muller (approved):

arturo-skydio:
@petr-muller sorry, just renamed `batchSkipTests` to `batchAllowPending` and added docs that are more in line with actual behavior, can you reapprove? Thanks again

petr-muller (approved):

arturo-skydio:
Any feedback here? Or suggestions on how to get this merged?

petr-muller (approved):

PR #20686, opened 2021-02-01
[Kettle] Catch and log the chunks that fail for future debugging

MushuEE:
Follow-up to #20666

That change retried with the same chunk and would fail. Unfortunately this is not easy to debug. This change will skip that failure and  log the issue to help identify the problem. 
/area kettle


spiffxp (approved):
/approve
/lgtm

MushuEE:
Darn haha, just hit an error when there is wrong encoding.

PR #20688, opened 2021-02-01
skip hostport tests on jobs with Docker runtime

aojea:
Kubernetes deployments that use docker as runtime, aka dockershim, have a
known issue on the HostPort implementation, pods can not be scheduled on
the same node because the code tries to bind a socket to avoid hijacking
services on the host, but it doesn´t take into account the HostIP field.

The test failure is legit, previusly it was skipped becauese it was
tagged as Serial when it did belong to sig-scheduling

Due to the dockershim deprecation it was decided to not move forward with the fix
xref. https://github.com/kubernetes/kubernetes/pull/94382

Fixes: https://github.com/kubernetes/kubernetes/issues/98648

### DIFF
--- a/config/jobs/kubernetes/sig-cloud-provider/gcp/gcp-gce.yaml
+++ b/config/jobs/kubernetes/sig-cloud-provider/gcp/gcp-gce.yaml
@@ -538,7 +538,7 @@ periodics:
           - --gcp-zone=us-west1-b
           - --ginkgo-parallel=30
           - --provider=gce
-          - --test_args=--ginkgo.skip=\[Slow\]|\[Serial\]|\[Disruptive\]|\[Flaky\]|\[Feature:.+\] --minStartupPods=8
+          - --test_args=--ginkgo.skip=\[Slow\]|\[Serial\]|\[Disruptive\]|\[Flaky\]|\[Feature:.+\]|HostPort --minStartupPods=8
           - --timeout=50m
         image: gcr.io/k8s-testimages/kubekins-e2e:v20210129-3799a64-master
         resources:


oomichi:
> Due to the dockershim deprecation it was decided to not move forward with the fix xref. kubernetes/kubernetes#94382

That makes sense.
One question: The skipped test is a part of conformance test. Is it nice to drop conformance tag from the test?

aojea:
> One question: The skipped test is a part of conformance test. Is it nice to drop conformance tag from the test?

The test is ok, and it keeps testing the same scenario than before, if it was Conformance before I think it has to continue being conformance.

However,  the test implementation wasn't correct hence some implementations with bugs, like in this case kubenet/hostport, were passing the test when they should be failing

I couldn't find a better solution, in theory, this job will not run in next releases :shrug:

BenTheElder:
Conformance tests may not skip, period.
Tests can be removed from conformance, but we should tag the conformance owners.

aojea:
> Conformance tests may not skip, period.
> Tests can be removed from conformance, but we should tag the conformance owners.

the test is right, I'm against removing a valid test from conformance just because kubenet, that is going to be deprecated has a bug that we don't want to fix

BenTheElder:
agreed.
We can instead skip running it on kubenet from the CI config side in jobs not targeting conformance. And by moving CI away from kubenet.

aojea:
I'm lost now, is this job targeting conformance? it doesn't mention it

BenTheElder:
It's not, but we have lots of GCE jobs and probably want some upstream conformance signal that is not just KIND, I forget which use kubenet but suspect many do?

aojea:
> It's not, but we have lots of GCE jobs and probably want some upstream conformance signal that is not just KIND, I forget which use kubenet but suspect many do?

send an email to sig-network today and will bring it up on this week meeting, I can followup after that ...

aojea:
I forget to mention that this test was already skipped in this job before @BenTheElder , becaues it was tagged as Serial

liggitt:
> The test is ok, and it keeps testing the same scenario than before, if it was Conformance before I think it has to continue being conformance.

the test was changed during the 1.20 timeframe to add the curl checks

BenTheElder:
> I forget to mention that this test was already skipped in this job before @BenTheElder , because it was tagged as Serial

Sure, but skipping it only in these particular jobs only punts the problem around. We should only do that if we think it's acceptable to fail in all jobs with this configuration.

aojea:
> > The test is ok, and it keeps testing the same scenario than before, if it was Conformance before I think it has to continue being conformance.
> 
> the test was changed during the 1.20 timeframe to add the curl checks

Because the test was wrong, was passing despite the ports were not forwarded. ... The scheduling worked because some of the pods were not using the hostport ... That is exactly what the test has to assert

liggitt:
It seems wrong to modify a conformance test without fixing an in tree component we ship to continue being conformant

aojea:
The fix was pushed back https://github.com/kubernetes/kubernetes/pull/94382

aojea:
If we want the test working again I can fix kubenet, just let me know

dims:
@aojea looks like we can close this out ... let's continue in https://github.com/kubernetes/kubernetes/pull/98755

aojea:
> @aojea looks like we can close this out ... let's continue in [kubernetes/kubernetes#98755](https://github.com/kubernetes/kubernetes/pull/98755)

absolutely

BenTheElder:
thanks!

PR #20701, opened 2021-02-02
Directly write to gzip from make_json

MushuEE:
Kettle currently runs something like `pypy3 make_json.py --days {DAY} | pv | gzip > build_day.json.gz`, tons of data is sent to stdout and I have a suspicion is contributing to the recent lockups we have been seeing.

This change writes the data directly to the `.gz` byte encoded file.

/hold complete staging testing

### DIFF
--- a/kettle/make_json.py
+++ b/kettle/make_json.py
@@ -17,6 +17,7 @@
 """Generate JSON for BigQuery importing."""
 
 import argparse
+import gzip
 import logging
 import json
 import os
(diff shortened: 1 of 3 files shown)


spiffxp:
Needs to pass tests

amwat:
is this still required?

spiffxp:
Same question

MushuEE:
This might no longer be needed, will close for now but will return if we again hit freeze

PR #20689, opened 2021-02-01
releng: add presubmit job to test kube cross image builder

cpanato:
Adding the first job to test if the setup is correct and not to make big changes 😄
If works fine will add the other jobs in a follow-up PR(s)

This PR adds a pre-submit job to test the image build for kube cross

Part of https://github.com/kubernetes/release/issues/1850

/assign @justaugustus @saschagrunert @hasheddan 
cc @kubernetes/release-engineering

Issue #20652, opened 2021-01-28
Mitigate image-pushing jobs hitting GetRequestsPerMinutePerProject quota for prow build cluster project


spiffxp:
**What happened**:
Context: https://github.com/kubernetes/k8s.io/issues/1576#issuecomment-767691409

As the volume of image-pushing jobs running on the prow build cluster in k8s-infra-prow-build-trusted has grown, we're starting to bump into a GCB service quota (GetRequestsPerMinutePerProject) for the project.  This isn't something we can request to raise like other quota (e.g. max gcp instances per region)

**What you expected to happen**:
Have GCB service requests charged to the project running the GCB builds instead of a central shared project.  Avoid bumping into API-related quota.

**How to reproduce it (as minimally and precisely as possible)**:
Merge a PR to kubernetes/kubernetes that updates multiple test/images subdirectories, or otherwise induce a high volume of image-pushing jobs on k8s-infra-prow-build-trusted

Ignore whether you bump into the concurrent builds quota (also a GCB service quota)

Can visualize usage (and whether quota is hit) here if a member of k8s-infra-prow-viewers@kubernetes.io: https://console.cloud.google.com/apis/api/cloudbuild.googleapis.com/quotas?orgonly=true&project=k8s-infra-prow-build-trusted&supportedpurview=project&pageState=(%22duration%22:(%22groupValue%22:%22P30D%22,%22customValue%22:null))

**Please provide links to example occurrences, if any**:
Don't have link to jobs that encountered this specifically, but https://github.com/kubernetes/k8s.io/issues/1576 describes the issue, and the metric explorer link above shows roughly when we've bumped into quota.

**Anything else we need to know?**:
Parent issue: https://github.com/kubernetes/release/issues/1869

My guess is that we need to move away from using a shared service account in the build cluster's project (gcb-builder@k8s-infra-prow-build-trusted), and instead setup service accounts per staging project.

It's unclear to me whether these would all need access to something in the build cluster project.

A service-account-per-project would add a bunch of boilerplate to the service accounts loaded into the build cluster, and add another field to job configs that needs to be set manually vs. copy-pasted.  We could offset this by verifying configs are correct via presubmit enforcement.

I'm open to other suggestions to automate the boilerplate away, or a solution that involves image-builder consuming less API quota.

/milestone v1.21
/priority important-soon
/wg k8s-infra
/sig testing
/area images
/sig release
/area release-eng
/assign @cpanato @justaugustus 
as owners of parent issue

[...]

PR #20703, opened 2021-02-02
releng: update pull-release-image-kube-cross to set the cluster where it should run

cpanato:
follow up of https://github.com/kubernetes/test-infra/pull/20689

the default cluster does not have the service account that is needed for this job, so adding the cluster that contains the setup

The error when the job tried to run

```
Pod can not be created: pods "63ae2683-6532-11eb-82fa-d61530c4f03a" ... count test-pods/gcb-builder: serviceaccount "gcb-builder" not found
```

/assign @saschagrunert @hasheddan @xmudrii @puerco 

### DIFF
--- a/config/jobs/kubernetes/release/release-config.yaml
+++ b/config/jobs/kubernetes/release/release-config.yaml
@@ -85,11 +85,12 @@ presubmits:
       testgrid-num-columns-recent: '30'
   - name: pull-release-image-kube-cross
     optional: true
+    cluster: k8s-infra-prow-build
     decorate: true
     run_if_changed: '^images\/build\/cross\/'
     path_alias: k8s.io/release
     spec:
-      serviceAccountName: gcb-builder
+      serviceAccountName: gcb-builder-releng-test
       containers:
         - image: gcr.io/k8s-testimages/image-builder:v20210204-b06ec78
           command:
@@ -102,6 +103,13 @@ presubmits:
           env:
             - name: LOG_TO_STDOUT
               value: "y"
+          resources:
+            requests:
+              cpu: 1000m
+              memory: 1Gi
+            limits:
+              cpu: 1000m
+              memory: 1Gi
     annotations:
       testgrid-dashboards: sig-release-releng-presubmits
       testgrid-tab-name: release-image-kube-cross


cpanato:
`jobs_test.go:351: pull-release-image-kube-cross: presubmits may not run in cluster: k8s-infra-prow-build-trusted`

/hold to see how to fix that

cpanato:
@spiffxp @saschagrunert @hasheddan 

We need to trigger a GCB build for PRs to test the images, but looks like the `gcb-builder` service account is not available in pre-submits, I did not find another service account, does that exist? If not how can we add that?

hasheddan (commented):
@cpanato I am guessing this is by design as typically we wouldn't be using this SA in a presubmit. However, in this specific case we have set up a testing repo so I would think the optimal scenario here would be to put a scoped token in the untrusted cluster if we want to maintain the idea of not running presubmits in trusted. Would defer to @spiffxp to weigh in though.

cpanato:
@hasheddan agree with you on this, lets wait for Aaron to see other options

spiffxp:
The reason the clusters are called "trusted" is that we don't want them running "untrusted" code, e.g. presubmits.  We treat the act of merging as the line of trust... code has passed tests, and the relevant humans have reviewed it.  So postubmits/periodics on trusted clusters are fine, but not presubmits.

@hasheddan's suggestion of scoped SA in untrusted cluster SGTM, PR to kubernetes/k8s.io would look something like:
- update scripts to make the GCP service account, bind via workload identity to an account of the same name on k8s-infra-prow-build (something like https://github.com/kubernetes/k8s.io/blob/master/infra/gcp/ensure-main-project.sh#L144-L155)
- give it privs to write to gcr for the project you care about (something like https://github.com/kubernetes/k8s.io/blob/35c51085f1096c01516021a8bcb34b9cfda656bd/infra/gcp/ensure-staging-storage.sh#L285-L295)
- add the service account resource to https://github.com/kubernetes/k8s.io/blob/master/infra/gcp/clusters/projects/k8s-infra-prow-build/prow-build/resources/build-serviceaccounts.yaml (eg: https://github.com/kubernetes/k8s.io/blob/master/infra/gcp/clusters/projects/k8s-infra-prow-build-trusted/prow-build-trusted/resources/build-serviceaccounts.yaml)

Now might be a good time to start adding some enforcement around which jobs are allowed to use which service accounts via tests in config/tests/jobs (ssuggested this in https://github.com/kubernetes/k8s.io/pull/1393#issuecomment-722002594)

This approach would also be baby steps toward "we should probably be using a service account per staging project" which might help address gcb quota issues https://github.com/kubernetes/test-infra/issues/20652

config/jobs/kubernetes/release/release-config.yaml
```diff
@@ -84,6 +84,7 @@ presubmits:
       testgrid-alert-email: release-managers+alerts@kubernetes.io
       testgrid-num-columns-recent: '30'
   - name: pull-release-image-kube-cross
+    cluster: k8s-infra-prow-build-trusted
```
spiffxp, line 87: change this to `k8s-infra-prow-build`

change L93 to `serviceAccountName: gcb-builder-releng-test` (ref: https://github.com/kubernetes/k8s.io/pull/1626)

cpanato: thanks @spiffxp. updated

config/jobs/kubernetes/release/release-config.yaml
```diff
@@ -102,6 +103,13 @@ presubmits:
           env:
             - name: LOG_TO_STDOUT
               value: "y"
+          resources:
+            requests:
+              cpu: 1000m
```
cpanato, line 108: the job will run on GCB so I don't think we need many resources

spiffxp (approved):
/lgtm
/hold cancel

PR #20707, opened 2021-02-02
Add support of excluded_repos for plugins

### LINKED ISSUE #20631: Consider supporting `ExcludedRepos` for plugins

Support `ExcludedRepos` like https://github.com/kubernetes/test-infra/blob/7e74b33534f950170801fb3ce015c1e96c975471/config/prow/config.yaml#L488. 

Currently we can only do:

```
plugins:
  orgA:
  - pluginCommon
  orgA/repoA:
  - pluginNotForRepoB
  orgA/repoC:
  - pluginNotForRepoB
```

**Why is this needed**:

Would be good if we can do:

```
plugins:
  orgA:
    plugins:
    - pluginCommon
  orgA:
    excluded_repos:
    - repoB
    plugins:
    - pluginNotForRepoB
```

This would be especially helpful in scenarios where a large org want to exclude one or two repos from prow without duplicating all plugins for all other repos.

matthyx:
Fixes #20631
/cc @chaodaiG

### DIFF
--- a/prow/plugins/config.go
+++ b/prow/plugins/config.go
@@ -26,6 +26,8 @@ import (
 	"strings"
 	"time"
 
+	"sigs.k8s.io/yaml"
+
 	"github.com/sirupsen/logrus"
 
 	"k8s.io/apimachinery/pkg/util/sets"
@@ -47,10 +49,10 @@ const (
 type Configuration struct {
 	// Plugins is a map of repositories (eg "k/k") to lists of
 	// plugin names.
-	// You can find a comprehensive list of the default avaulable plugins here
+	// You can find a comprehensive list of the default available plugins here
 	// https://github.com/kubernetes/test-infra/tree/master/prow/plugins
 	// note that you're also able to add external plugins.
-	Plugins map[string][]string `json:"plugins,omitempty"`
+	Plugins Plugins `json:"plugins,omitempty"`
 
 	// ExternalPlugins is a map of repositories (eg "k/k") to lists of
 	// external plugins.
@@ -106,6 +108,13 @@ type Golint struct {
 	// in (0,1] over which problems will be printed. Defaults to
 	// 0.8, as does the `go lint` tool.
 	MinimumConfidence *float64 `json:"minimum_confidence,omitempty"`
+}
+
+type Plugins map[string]OrgPlugins
+
+type OrgPlugins struct {
+	ExcludedRepos []string `json:"excluded_repos,omitempty"`
+	Plugins       []string `json:"plugins,omitempty"`
 }
 
 // ExternalPlugin holds configuration for registering an external
@@ -802,11 +811,37 @@ func (c *Configuration) DcoFor(org, repo string) *Dco {
 	return &Dco{}
 }
 
+func OldToNewPlugins(oldPlugins map[string][]string) Plugins {
+	newPlugins := make(Plugins)
+	for repo, plugins := range oldPlugins {
+		newPlugins[repo] = OrgPlugins{
+			Plugins: plugins,
+		}
+	}
+	return newPlugins
+}
+
+type pluginsWithoutUnmarshaler Plugins
+
+func (p *Plugins) UnmarshalJSON(d []byte) error {
+	var oldPlugins map[string][]string
+	if err := yaml.Unmarshal(d, &oldPlugins); err == nil {
+		logrus.Warn("plugins declaration uses a deprecated config style, please migrate it")
+		*p = OldToNewPlugins(oldPlugins)
+		return nil
+	}
+	var target pluginsWithoutUnmarshaler
+	err := yaml.Unmarshal(d, &target)
+	*p = Plugins(target)
+	return err
+}
+
 // EnabledReposForPlugin returns the orgs and repos that have enabled the passed plugin.
-func (c *Configuration) EnabledReposForPlugin(plugin string) (orgs, repos []string) {
+func (c *Configuration) EnabledReposForPlugin(plugin string) (orgs, repos []string, orgExceptions map[string]sets.String) {
+	orgExceptions = make(map[string]sets.String)
 	for repo, plugins := range c.Plugins {
 		found := false
-		for _, candidate := range plugins {
+		for _, candidate := range plugins.Plugins {
 			if candidate == plugin {
 				found = true
 				break
@@ -817,8 +852,17 @@ func (c *Configuration) EnabledReposForPlugin(plugin string) (orgs, repos []stri
 				repos = append(repos, repo)
 			} else {
 				orgs = append(orgs, repo)
+				orgExceptions[repo] = sets.NewString()
+				for _, excludedRepo := range plugins.ExcludedRepos {
+					orgExceptions[repo].Insert(fmt.Sprintf("%s/%s", repo, excludedRepo))
+				}
 			}
 		}
+	}
+	// <plugin> plugin might be declared in both org and org/repo
+	// in that case, remove repo from org's orgExceptions despite the excluded_repo in org
+	for _, repo := range repos {
+		orgExceptions[strings.Split(repo, "/")[0]].Delete(repo)
 	}
 	return
 }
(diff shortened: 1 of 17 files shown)


matthyx:
I need to add test cases for the new format, and fix some CI checks, but could you check already if that would do it?

matthyx:
This is the first PR where I need to support both old and new config styles... after some time we can simplify it by ditching the old format.

chaodaiG:
Would worth mentioning in https://github.com/kubernetes/test-infra/blob/master/prow/ANNOUNCEMENTS.md as well I think

chaodaiG:
> I need to add test cases for the new format, and fix some CI checks, but could you check already if that would do it?

Sorry overlooked this comment last time. This looks exactly like what's in my mind!

prow/plugins/config.go
```diff
@@ -817,9 +853,15 @@ func (c *Configuration) EnabledReposForPlugin(plugin string) (orgs, repos []stri
 				repos = append(repos, repo)
 			} else {
 				orgs = append(orgs, repo)
+				orgExceptions[repo] = sets.NewString(plugins.ExcludedRepos...)
```
matthyx, line 856: I am not sure `orgExceptions` is the right way to go... do you have an idea @alvaroaleman ?

matthyx:
@chaodaiG PTAL

prow/plugins/config.go
chaodaiG: My feeling is that this is very explicit, like `orgA/repoB`, so a slice of repos should do?

prow/plugins/plugins.go
```diff
@@ -431,8 +431,8 @@ func (pa *ConfigAgent) getPlugins(owner, repo string) []string {
 	var plugins []string
 
 	fullName := fmt.Sprintf("%s/%s", owner, repo)
-	plugins = append(plugins, pa.configuration.Plugins[owner]...)
-	plugins = append(plugins, pa.configuration.Plugins[fullName]...)
+	plugins = append(plugins, pa.configuration.Plugins()[owner].Plugins...)
+	plugins = append(plugins, pa.configuration.Plugins()[fullName].Plugins...)
```
chaodaiG, line 435: Shouldn't we process excluded repos here?

chaodaiG (changes requested):

prow/plugins/config.go
```diff
@@ -817,9 +853,15 @@ func (c *Configuration) EnabledReposForPlugin(plugin string) (orgs, repos []stri
...
 			}
 		}
 	}
+	// if org/repo has explicitly plugin enabled, remove its exception
```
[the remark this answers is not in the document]
matthyx: yes that's what I meant, alright will rephrase

prow/plugins/config.go
```diff
@@ -802,11 +809,40 @@ func (c *Configuration) DcoFor(org, repo string) *Dco {
...
+	newPlugins := make(map[string]OrgPlugins)
+	for repo, plugins := range oldPlugins {
+		newPlugins[repo] = OrgPlugins{
+			Plugins: plugins,
+		}
+	}
+	return newPlugins
+}
+
+func ToRawMessage(plugins interface{}) json.RawMessage {
```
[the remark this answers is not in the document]
matthyx: yes in all tests, as during the migration time we need to use a `json.RawMessage` to allow deserialization with our own code until we completely switch to the new form

prow/plugins/plugins.go
matthyx: yes

prow/plugins/config.go
matthyx: I am just trying to integrate with the existing mechanisms

prow/plugins/config.go
```diff
@@ -802,11 +809,40 @@ func (c *Configuration) DcoFor(org, repo string) *Dco {
...
+	raw, _ := yaml.Marshal(plugins)
+	return raw
+}
+
+func (c *Configuration) Plugins() map[string]OrgPlugins {
+	var oldPlugins map[string][]string
+	if err := yaml.Unmarshal(c.PluginsRaw, &oldPlugins); err != nil {
+		var newPlugins map[string]OrgPlugins
+		if err := yaml.Unmarshal(c.PluginsRaw, &newPlugins); err != nil {
+			logrus.WithError(err).Error("error unmarshalling Plugins config to oldPlugins and newPlugins formats")
```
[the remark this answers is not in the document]
matthyx: I see, but then what do I do with this error in the caller side?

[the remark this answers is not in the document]
chaodaiG: I think checkconfig need to catch this error and not letting the change go through. And for hook I would prefer to let it crash, with an error message asking for running checkconfig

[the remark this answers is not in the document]
matthyx: This custom unmarshaling is just for the transition period, to let people migrate from old to new format... otherwise such error is thrown when the yaml file is parsed, so maybe I should validate at that very beginning and not every time I need to access plugins?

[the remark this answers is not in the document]
matthyx: Ok, I will do like `ConfigUpdater` and redefine my own unmarshall method...

prow/plugins/config.go
```diff
@@ -802,11 +811,37 @@ func (c *Configuration) DcoFor(org, repo string) *Dco {
...
+	newPlugins := make(Plugins)
+	for repo, plugins := range oldPlugins {
+		newPlugins[repo] = OrgPlugins{
+			Plugins: plugins,
+		}
+	}
+	return newPlugins
+}
+
+type pluginsWithoutUnmarshaler Plugins
```
matthyx, line 824: This is just so elegant... thanks @alvaroaleman for the tip in the same file!

prow/plugins/config_test.go
```diff
@@ -1441,19 +1441,25 @@ func TestConfigUpdaterResolve(t *testing.T) {
 }
 
 func TestEnabledReposForPlugin(t *testing.T) {
```
chaodaiG, line 1443: I would expect a test case that cannot be unmarshalled

prow/plugins/plugins_test.go
```diff
@@ -144,6 +144,80 @@ func TestGetPlugins(t *testing.T) {
 			expectedPlugins: []string{"plugin3"},
 		},
 	}
+	for _, tc := range testcases {
+		pa := ConfigAgent{configuration: &Configuration{Plugins: OldToNewPlugins(tc.pluginMap)}}
+
+		plugins := pa.getPlugins(tc.owner, tc.repo)
+		if len(plugins) != len(tc.expectedPlugins) {
```
chaodaiG, line 151: Is it possible to use cmp.Diff instead? Same below

chaodaiG (approved):
Overall looks awesome to me. Left one last nit

/hold
In case you want to address

matthyx:
> In case you want to address

Squashed commits as well. Thanks for your help!

chaodaiG (approved):
Looks awesome! Thank you for the contribution!

@michelle192837 and @alvaroaleman  for awareness

prow/plugins/config_test.go
```diff
@@ -1437,3 +1439,84 @@ func TestConfigUpdaterResolve(t *testing.T) {
...
+ plugins:
+ - pluginCommon
+ - pluginNotForRepoB
+orgA/repoB:
+ plugins:
+ - pluginCommon
+ - pluginOnlyForRepoB
+`)
+	var p Plugins
+	err := yaml.Unmarshal(badPluginsYaml, &p)
```
chaodaiG, line 1518: Sorry I must haven't got enough coffee for the morning, didn't notice this test didn't use `UnmarshalJSON` method. Please modify the test to use it

chaodaiG (commented):
Just noticed that there is still a hold on the PR, left a comment. No big deal, so please feel free address it in current PR or in future PR

prow/plugins/config_test.go
matthyx: Sorry, fixed!

chaodaiG (approved):
Thank you!

cjwagner:
This is handy, thanks Matthias!

matthyx:
> This is handy, thanks Matthias!

Now I will rework my never ending contribution for config trees... I'm almost sure by the time I get it merged go will support generics 😂

fejta:
What documentation am I supposed to follow to do this migration?

fejta:
This is somewhat unhelpful:
```
msg: "plugins declaration uses a deprecated config style, please migrate it"
file: "prow/plugins/config.go:843"
```

fejta:
Also can you please update this logging to only say this once per process and/or once per hour? Right now this is the only thing visible in our logs.

PR #20710, opened 2021-02-02
Update presets to use DO spaces for kops e2e tests

srikiz:
Updating presets here to include DO Spaces access key and secret.
Once this gets in, I am planning to create a new job that KOPS job can use it.

Please let me know in case I am missing something here, thanks !
FYI - @MorrisLaw @rifelpet


rifelpet:
If it is acceptable for cluster-api-provider-digitalocean to also have access to the S3 keys then this looks correct to me. Otherwise we could create a second preset just for the S3 keys and have the kops jobs use both presets.

prksu:
agree with @rifelpet, have different preset between DO space keys and access token would be better.

prksu (approved):
/approve

PR #20715, opened 2021-02-02
azure: remove unsupported Kubernetes versions

chewong:

Removing jobs with unsupported Kubernetes versions according to https://github.com/kubernetes/sig-release/blob/master/releases/patch-releases.md#non-active-branch-history.


feiskyer (approved):
/lgtm

PR #20717, opened 2021-02-03
Refactor hook to use the github event server

droslean:
The github event server was created following the hook's concept. So currently hook is using so much duplicated code. 
This PR removes all the standalone event server functionality and use the github event server instead.


droslean:
/hold

I am holding it to do some manually end-to-end testing

prow/cmd/hook/main.go
```diff
@@ -223,42 +224,43 @@ func main() {
...
+		ClientAgent: clientAgent,
+		ConfigAgent: configAgent,
+		Plugins:     pluginAgent,
+		Metrics:     promMetrics,
+		RepoEnabled: o.githubEnablement.EnablementChecker(),
 	}
+
+	eventServer := githubeventserver.New(o.githubEventServerOptions, secretAgent.GetTokenGenerator(o.webhookSecretFile), logrus.NewEntry(logrus.New()))
+
+	eventServer.RegisterReviewEventHandler(server.HandleReviewEvent)
```
stevekuznetsov, line 243: The event server should just handle multiple plugins for one event, not needing the second-level server to do the multiplexing.

droslean: This functionality is registered via the external plugins in [here](https://github.com/kubernetes/test-infra/pull/20717/files#diff-25f3adc74c147e59649ac9242dc3b9f462de7682b53810df1f8afe06485004fdR252)

PR #20719, opened 2021-02-03
labels_sync: allow setting org-level labels

petr-muller:
In OpenShift we cargo-culted us into thinking the tool actually supports
org-level labels but it doesn't. The *proper* structure for doing this
is probably more hierarchic (like some of the Prow plugins, e.g.
branch protector), but I wanted to avoid larger changes.

### DIFF
--- a/label_sync/main.go
+++ b/label_sync/main.go
@@ -84,6 +84,7 @@ type Label struct {
 // There is also a Default list of labels applied to every Repo
 type Configuration struct {
 	Repos   map[string]RepoConfig `json:"repos,omitempty"`
+	Orgs    map[string]RepoConfig `json:"orgs,omitempty"`
 	Default RepoConfig            `json:"default"`
 }
 
@@ -239,6 +240,9 @@ func stringInSortedSlice(a string, list []string) bool {
 func (c Configuration) Labels() []Label {
 	var labelarrays [][]Label
 	labelarrays = append(labelarrays, c.Default.Labels)
+	for _, org := range c.Orgs {
+		labelarrays = append(labelarrays, org.Labels)
+	}
 	for _, repo := range c.Repos {
 		labelarrays = append(labelarrays, repo.Labels)
 	}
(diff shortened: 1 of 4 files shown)


label_sync/main.go
```diff
@@ -265,29 +269,43 @@ func (c Configuration) Labels() []Label {
...
+
+	// Check org-level labels for duplicities with default labels
+	orgSeen := map[string]map[string]string{}
+	for org, orgConfig := range c.Orgs {
+		if orgSeen[org], err = validate(orgConfig.Labels, org, defaultSeen); err != nil {
+			return fmt.Errorf("invalid config: %v", err)
+		}
+	}
+
+	// Check other repos labels for
```
stevekuznetsov, line 289: for ... ?

stevekuznetsov (approved):

label_sync/main.go
petr-muller: I wonder what I wanted to say here :/

petr-muller:
Blah it's reassuring to see we're not the only org suffering from random flakes

```
Image pull was unsuccessful: unable to save remote image gcr.io/k8s-prow/git@sha256:1527341aff1003b6b27c8ed935e6f0200258bee55b6eb178ca3ef124196384fe: unable to write image layers: unable to write image layer: unable to write the contents of layer 2 to /bazel-scratch/.cache/bazel/_bazel_root/05618a594cb3499a4b912817865867cd/external/git-base/image/002.tar.gz: read tcp 10.60.45.105:43820->172.217.214.128:443: read: connection reset by peer
```

/retest

fejta (approved):
Do we want to support the ability to migrate a repo label to an org label?

/hold

fejta:
(unhold at will)

petr-muller:
> Do we want to support the ability to migrate a repo label to an org label?

It's possible that I missed something important. Where's the problem in that? IIUC both default and now org-level labels work by iterating over repos and treat them individually. So migrating a repo label to an org. you could just add it to an org, remove it from the repo (this is enforced by the duplicate check) and that is all. When running, the label would be added to all other repos in the org, and the operation would be a nop for the repo that already had the label.

GH has a "default labels" tied to an org, which controls the labels that newly created repos have, but `label_sync` currently does not interact with these at all. An interesting question whether maintaining these would belong to label_sync or peribolos

/hold cancel

PR #20721, opened 2021-02-03
kubernetes-csi: fix snapshotter for periodic on-kubernetes-1-20

pohly:
Those jobs use the release-1.20 branch and that wasn't detected as
Kubernetes 1.20 where we want to use the v4.0.0 snapshotter.

### DIFF
--- a/config/jobs/kubernetes-csi/gen-jobs.sh
+++ b/config/jobs/kubernetes-csi/gen-jobs.sh
@@ -265,6 +265,34 @@ pull_alwaysrun() {
     fi
 }
 
+# version_gt returns true if arg1 is greater than arg2.
+#
+# This function expects versions to be one of the following formats:
+#   X.Y.Z, release-X.Y.Z, vX.Y.Z
+#
+#   where X,Y, and Z are any number.
+#
+# Partial versions (1.2, release-1.2) work as well.
+# The follow substrings are stripped before version comparison:
+#   - "v"
+#   - "release-"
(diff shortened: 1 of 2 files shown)


config/jobs/kubernetes-csi/gen-jobs.sh
```diff
@@ -281,7 +281,7 @@ snapshotter_version() {
         # Additional jobs could be created to cover version
         # skew, if desired.
         case "$kubernetes" in
-            1.20)
+            latest|release-1.20|1.20)
```
msau42, line 284: is "master" also a possible value?

pohly: Not in the current version of the script. There is one `snapshotter_version latest` invocation and one with `snapshotter $actual` where actual is set to `latest`.

But it can't hurt to include it, just in case.

I was also toying with the idea to use `1.[2-9][0-9]` and `release-1.[2-9][0-9]` as pattern to avoid having to update this when adding 1.21 and future version, but wasn't sure - might be overkill?

msau42: Hmmm that's a good point. I think having a > check would be nice so we don't have to keep remembering to update this

pohly: I've copied the `version_gt` from the hostpath deploy script and added "master".

PR #20725, opened 2021-02-03
Install bazel 3.1.0 on kubekins-e2e images

chaodaiG:


images/kubekins-e2e/variants.yaml
```diff
@@ -16,8 +16,8 @@ variants:
     CONFIG: master
```
fejta, line 16: please create a test-infra variant, do not corrupt the kubernetes/kubernetes variant

chaodaiG: Done.

fejta:
why are we using kubekins rather than krte?

chaodaiG:
It's symlinked I believe, this file is used by both images

fejta:
THis LGTM but I don't see how krte has anything to do with anything.

I can't actually identifying anything using krte (including this PR). Should we just delete this image?

fejta:
> I can't actually identifying anything using krte (including this PR). Should we just delete this image?

Ah, no it looks like things that do not use bazel sometimes use this image: https://github.com/search?q=org%3Akubernetes+krte&type=code

Can you remove the references to krte here? It is unrelated to your task.

chaodaiG:
Sounds good, updated PR title

fejta (approved):

PR #20729, opened 2021-02-03
Make prometheus alerts rule for ci absent configurable

chaodaiG:
Not all prow instances have the same set of core components to be monitored, make this configurable so it's easier to upgrade.

### DIFF
--- a/config/prow/cluster/monitoring/mixins/prometheus/ci_absent_alerts.libsonnet
+++ b/config/prow/cluster/monitoring/mixins/prometheus/ci_absent_alerts.libsonnet
@@ -1,6 +1,5 @@
 {
   prometheusAlerts+:: {
-    local comps = $._config.components,
     groups+: [
       {
         name: 'ci-absent',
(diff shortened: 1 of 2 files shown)


config/prow/cluster/monitoring/mixins/prometheus/ci_absent_alerts.libsonnet
```diff
@@ -19,16 +18,7 @@
...
-              comps.crier,
-              comps.deck,
-              comps.ghproxy,
-              comps.hook,
-              comps.horologium,
-              comps.prowControllerManager,
-              comps.sinker,
-              comps.tide,
-          ]
+          for name in $._config.ciAbsents.components
```
cjwagner, line 21: nit: This is not backwards compatible without defaulting. Not sure if we care yet given that updates are currently manual anyways.

cjwagner (approved):
Feel free to unhold.
/hold

config/prow/cluster/monitoring/mixins/prometheus/ci_absent_alerts.libsonnet
chaodaiG: ack, will be careful when updating other prow monitoring stacks

Issue #20627, opened 2021-01-27
Prow's unit tests assume `master` as the default branch which is not true for newer versions of git.


cjwagner:
**What happened**:
Newer versions of `git` default to naming the initial branch `main` rather than `master`.
Prow unit tests make assumptions about the default branch name so most unit tests involving git result in errors like the following:
- `Error checking out master: /usr/bin/git [checkout master]: exit status 1, error: pathspec 'master' did not match any file(s) known to git`
- `failed to get baseSHA: /usr/bin/git [rev-parse master]: exit status 128, fatal: ambiguous argument 'master': unknown revision or path not in the working tree.`

**What you expected to happen**:
See the slack discussion here: https://kubernetes.slack.com/archives/CDECRSC5U/p1611770593024000
In particular, from @alvaroaleman:
> in [that stackoverflow](https://stackoverflow.com/questions/42871542/how-can-i-create-a-git-repository-with-the-default-branch-name-other-than-maste/42871621#42871621) they mention that you can explicitly specific that branch as arg to git init via --initial-branch=name. Since we apparently make assumptions about that branches name, passing that arg when doing git init sounds like a good idea to me

**How to reproduce it (as minimally and precisely as possible)**:
Update to a new version of git or configure the global `init.defaultBranch` configuration option to `main` rather than master and then run the Prow unit tests.

cc @alvaroaleman

alvaroaleman:
Fixed in #20730
/close

PR #20730, opened 2021-02-03
Allow unit tests to pass when init.defaultBranch is set.

fejta:
Currently a lot of unit tests expect the default branch to be master. This is now a runtime
configurable variable, often configured to main. Update tests to handle this situation --
note that one test was behaving oddly with this, so configured this branch to force master as the initial branch.

### DIFF
--- a/prow/git/git_test.go
+++ b/prow/git/git_test.go
@@ -30,6 +30,8 @@ import (
 	"k8s.io/test-infra/prow/github"
 )
 
+var defaultBranch = localgit.DefaultBranch("")
+
 func TestClone(t *testing.T) {
 	testClone(localgit.New, t)
 }
@@ -204,12 +206,12 @@ func testMergeCommitsExistBetween(clients localgit.Clients, t *testing.T) {
 			}
 		}
 		mergeMaster = func() {
-			if _, err := lg.Merge("foo", "bar", "master"); err != nil {
+			if _, err := lg.Merge("foo", "bar", defaultBranch); err != nil {
 				t.Fatalf("Rebasing commit: %v", err)
 			}
 		}
 		rebaseMaster = func() {
-			if _, err := lg.Rebase("foo", "bar", "master"); err != nil {
+			if _, err := lg.Rebase("foo", "bar", defaultBranch); err != nil {
 				t.Fatalf("Rebasing commit: %v", err)
 			}
 		}
@@ -248,7 +250,7 @@ func testMergeCommitsExistBetween(clients localgit.Clients, t *testing.T) {
 		checkoutPR(tt.prNum)
 	}
 	// switch back to master and create a new commit 'ouch'
-	checkoutBranch("master")
+	checkoutBranch(defaultBranch)
 	addCommit("ouch")
 	masterSHA, err := lg.RevParse("foo", "bar", "HEAD")
 	if err != nil {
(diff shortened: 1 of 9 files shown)


prow/git/localgit/localgit.go
```diff
@@ -32,13 +32,25 @@ import (
 
 type Clients func() (*LocalGit, v2.ClientFactory, error)
 
+func DefaultBranch(dir string) string {
```
alvaroaleman, line 35: Why scatter invocations of this all over the place into package-level vars rather than just hardcoding any value for `InitialBranch` and exposing it via a `const`?

alvaroaleman:
Ref https://github.com/kubernetes/test-infra/issues/20627

prow/git/localgit/localgit.go
fejta: Because the default branch is not a constant

fejta: It varies per system, per git repo, per version of git.

alvaroaleman: I am aware, that is why `localgit` should always explicitly set it (aka default `InitialBranch` to the value of an exported `const` and use that `const` rather than the current `var`)

fejta: nah

fejta: I do not want to depend on the `git init --initial-branch=foo` flag being available, as it was only added in July 2020:

* https://git-scm.com/docs/git-init/2.27.0
* https://git-scm.com/docs/git-init/2.28.0

alvaroaleman: But you do already by having added an option for that. Besides, why not? Developers tend to have fairly current versions of git and CI obviously works as well

fejta: I didn't though. The flag is only used on one test and only when the default branch is not master, which is only possible when an old version of git is used

prow/git/git_test.go
```diff
@@ -30,6 +30,8 @@ import (
 	"k8s.io/test-infra/prow/github"
 )
 
+var defaultBranch = localgit.DefaultBranch("")
```
cjwagner, line 33: Why is `DefaultBranch` always called with the empty string? If using the current directory is always appropriate can we remove the parameter?
Another benefit of this would be that we could determine the default branch once as part of the initialization of the `localgit` package and then expose the value as a global rather than many package level variables. This would be more similar to what Alvaro is suggesting.

fejta: This is not a global value though. It is something that that can be defined anywhere git can be configured, aka at the system, global, repo and/or submodule levels. 

The change here was not to switch the HEAD branch from master to foo but to make the head branch something configurable, and IMO we should reflect this reality in the code rather than just choose a different global constant.

fejta: That's also reason I have this as a string: the value of `git config init.defaultBranch` depends on the git repo so allow that to be passed into this exported method (defaulting to the current dir)

fejta:
> it makes local testing annoying.

Running `git config --global init.defaultBranch master` is a way to work around the issue until this merges. Agree it makes unit testing annoying.

cjwagner:
> Running `git config --global init.defaultBranch master` is a way to work around the issue until this merges.

I know this should work, but it doesn't seem to for me. I think both my global and repo specific config are set correctly:
```console
$ git config --get init.defaultBranch
master
$ git config --global --get init.defaultBranch
master

```

PR #20736, opened 2021-02-04
Check leaked resources in presubmits

mborsz:
Ref https://github.com/kubernetes/kubernetes/issues/98757

It's a second time some change in kubernetes/kubernetes makes our periodic job fail on diffResources. The reason for that is that k/k presubmit doesn't run diffResources tests, so we will not block such changes.

This PR enables leaked resources checking in both k/k and perf-tests presubmits to match our CI job.

/assign @wojtek-t @jkaniuk @mm4tt

### DIFF
--- a/config/jobs/kubernetes/sig-scalability/sig-scalability-presubmit-jobs.yaml
+++ b/config/jobs/kubernetes/sig-scalability/sig-scalability-presubmit-jobs.yaml
@@ -28,6 +28,7 @@ presubmits:
         - --timeout=120
         - --scenario=kubernetes_e2e
         - --
+        - --check-leaked-resources
         - --build=bazel
         - --cluster=
         - --extract=local
@@ -427,6 +428,7 @@ presubmits:
         - --timeout=120
         - --scenario=kubernetes_e2e
         - --
+        - --check-leaked-resources
         - --cluster=
         - --extract=ci/latest
         - --gcp-nodes=100


mborsz:
/hold
Let's wait until the diffResources starts passing to avoid blocking presubmit

https://github.com/kubernetes/perf-tests/pull/1700 is expected to fix the issue

wojtek-t:
Test failure seems related - PTAL

mborsz:
Hmmm

```
    jobs_test.go:907: Invalid Scenario Args : presubmit job pull-kubernetes-e2e-gce-100-performance should not check for resource leaks
```

mborsz:
The check has been added in https://github.com/kubernetes/test-infra/pull/8642

/cc @krzyzacy @BenTheElder

Why presubmits are not allowed to check if they leak resources? What is a reasoning behind this rule?
What is a suggested way of preventing a case that happened twice already: some change (here change of label names) made our presubmit leaking resources and we haven't blocked that PR as the presubmit doesn't check for resource leak?

BenTheElder:
It's too far back that I don't recall anything about the origin of this, I may not have even been involved, however:

Why are we not letting boskos cleanup resources?
When we're testing kubernets/kubernetes, why do we care if they leak? Leaking resources is a property of cluster teardown, not kubernetes.

jkaniuk:
> Why are we not letting boskos cleanup resources?

That is just masking a problem, not eliminating it.
Not everybody run Prow tests or kubetest using Boskos. 
Boskos has its own bugs/issues regarding cleanup - not everything is supported.

BenTheElder:
> That is just masking a problem, not eliminating it.

Kubernetes/Kubernetes has no reason to care if the provisioner used for testing leaks resources, that's a bug in that project.
It doesn't make sense to block/flake Kubernetes/Kubernetes PRs on such a failure and the PR author is unlikely to do anything about this.

> Not everybody run Prow tests or kubetest using Boskos.

They do when we're talking about our presubmits ...

> Boskos has its own bugs/issues regarding cleanup - not everything is supported.

Do you have a concrete example? That's an unacceptable situation to be in. Jobs can be terminated without reaching completion and boskos must still resiliently cleanup after them.

BenTheElder:
cc @spiffxp @ixdy

PR #20737, opened 2021-02-04
Add a script that updates bazel using the bazel image

droslean:
Recently I found myself straggling while trying to update bazel and I was able to do the work by using the bazel image. It seems that bazel needs a lot of dependencies which is hard to be used by the contributors. 

This PR adds a script to update bazel using the `gcr.io/k8s-testimages/launcher.gcr.io/google/bazel` image. 

/cc @fejta 

### DIFF
--- a/hack/update-bazel.sh
+++ b/hack/update-bazel.sh
@@ -17,10 +17,43 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
-if ! command -v bazel &>/dev/null; then
-  echo "Install bazel at https://bazel.build" >&2
-  exit 1
+
+bazel-from-image() {
+	if [ -x "$(command -v docker)" ]; then
+	    CONTAINER_ENGINE=docker
+	elif [ -x "$(command -v podman)" ]; then
+	    CONTAINER_ENGINE=podman
+	else
+		echo "There is no docker or podman installed. Please use hack/update-bazel.sh script instead."
+		exit 1
+	fi
+
+	WORKDIR=$(pwd)
+	TEST_INFRA_PATH=${WORKDIR%/hack}
+
+
+	$CONTAINER_ENGINE run --user $(id -u):$(id -g) --volume ${TEST_INFRA_PATH}:/test-infra:Z \
+	--workdir /test-infra --rm gcr.io/k8s-testimages/launcher.gcr.io/google/bazel:latest-test-infra $@
+}
+
+bazel-direct() {
+  bazel $@
+}
+
+bazel-from-bazelisk() {
+  bazelisk $@
+}
+
+if [[ "$1" == "--from-image" ]]; then
+	bazel=bazel-from-image
+elif [ -x "$(command -v bazelisk)" ]; then
+  bazel=bazel-from-bazelisk
+elif [ -x "$(command -v bazel)" ]; then
+  bazel=bazel-direct
+else
+  bazel=bazel-from-image
 fi
 
-set -o xtrace
-bazel run @io_k8s_repo_infra//hack:update-bazel
+"$bazel" run @io_k8s_repo_infra//hack:update-bazel
+
+exit $?


hack/update-bazel-with-image.sh
```diff
@@ -0,0 +1,40 @@
...
+	exit 1
+fi
+
+WORKDIR=$(pwd)
+TEST_INFRA_PATH=${WORKDIR%/hack}
+
+
+$CONTAINER_ENGINE run --volume ${TEST_INFRA_PATH}:/test-infra:Z \
+--workdir /test-infra \
+-it gcr.io/k8s-testimages/launcher.gcr.io/google/bazel:latest-test-infra \
```
alvaroaleman, line 36: Can you also fix perms to use the userid so we don't end up creating files owned by root?

alvaroaleman (commented):
/assign @fejta

hack/update-bazel-with-image.sh
droslean: well, in theory, the developer could work as root and commit his changes with these perms. I don't think that this script will affect anything. That depends on the developer.

alvaroaleman: no it does not depend on the developer. This script needs to include `--user=$UID` so it will run as the same UID as the user that is calling it

hack/update-bazel-with-image.sh
```diff
@@ -0,0 +1,40 @@
...
+else
+	echo "There is no docker or podman installed. Please use hack/update-bazel.sh script instead."
+	exit 1
+fi
+
+WORKDIR=$(pwd)
+TEST_INFRA_PATH=${WORKDIR%/hack}
+
+
+$CONTAINER_ENGINE run --volume ${TEST_INFRA_PATH}:/test-infra:Z \
```
fejta, line 34: What do we need -it for?
Will your add --rm=true?

hack/update-bazel-with-image.sh
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env bash
+# Copyright 2016 The Kubernetes Authors.
```
fejta, line 2: Resetting the last 5 years certainly has its appeal, alas...

hack/update-bazel-with-image.sh
```diff
@@ -0,0 +1,40 @@
...
+fi
+
+WORKDIR=$(pwd)
+TEST_INFRA_PATH=${WORKDIR%/hack}
+
+
+$CONTAINER_ENGINE run --volume ${TEST_INFRA_PATH}:/test-infra:Z \
+--workdir /test-infra \
+-it gcr.io/k8s-testimages/launcher.gcr.io/google/bazel:latest-test-infra \
+run @io_k8s_repo_infra//hack:update-bazel
```
fejta, line 37: How do you feel about moving all the contents of this file inside https://github.com/kubernetes/test-infra/blob/master/hack/update-bazel.sh?

hack/update-bazel-with-image.sh
droslean: So the best practice here would be `--user $(id -u):$(id -g)` which for some reason doesn't work on me. I guess the problem is on my side. I would appreciate it if you check it too...

hack/update-bazel.sh
```diff
@@ -17,10 +17,23 @@ set -o errexit
...
-if ! command -v bazel &>/dev/null; then
-  echo "Install bazel at https://bazel.build" >&2
-  exit 1
+
+if [ -x "$(command -v docker)" ]; then
+    CONTAINER_ENGINE=docker
+elif [ -x "$(command -v podman)" ]; then
+    CONTAINER_ENGINE=podman
+else
+	echo "There is no docker or podman installed. Please use hack/update-bazel.sh script instead."
```
fejta, line 26: Ideally we just merge these two scripts, aka something along the lines of:

* If bazel or bazelisk is installed, use that
* Otherwise use an image.


```sh
bazel-from-image() {
  $CONTAINER_ENGINE run --user $(id -u) ... $@
}

bazel-direct() {
  bazel $@
}

bazel-from-bazelisk() {
  bazelisk $@
}


if [ -x $(command -v bazelisk) ]; then
  bazel=bazel-from-bazelisk
elif [ -x $(command -v bazel) ]; then
  bazel=bazel-direct
else
  bazel=bazel-from-image
fi

"$bazel" run @io_k8s_repo_infra//hack:update-bazel
```

droslean: But what if a user wants to run from an image even if he has bazel already installed? That was my case at least. I guess adding an argument will be ok.

droslean: @fejta updated.

hack/update-bazel-with-image.sh
alvaroaleman: works fine:
```
 docker run --rm  --user $(id -u):$(id -g)  debian id
uid=1000 gid=1000 groups=1000
```

fejta (approved):

PR #20740, opened 2021-02-04
kops: move grid to alpha channel

justinsb:
As discussed, we should get the (broad) grid on the alpha channel, to
give us earlier signal.

We can have PR tests and "headline" periodics on the stable channel.


justinsb:
Rebased, I kept the distros jobs on stable (though I don't know whether they should be on alpha; easy enough to change in any case!  cc @rifelpet )

justinsb:
/test pull-test-infra-bazel

infra flake

PR #20743, opened 2021-02-04
Approvers: Add AutoApproveUnownedSubfolders option

alvaroaleman:
This change adds an allow_folder_creation option to the approvers. When
opted-in, it will not require approval for subfolders in a folder with
this set. It is implemented by purging the files that fall into this
category from the list of approval-needing files in the approve plugin.

Note for reviewers: In two occasions I have changed a set of table-driven tests to use subtests, please disable whitespace on the diff, that drastically reduces the diff size.

### DIFF
--- a/prow/plugins/approve/approvers/owners.go
+++ b/prow/plugins/approve/approvers/owners.go
@@ -46,11 +46,18 @@ type Repo interface {
 	LeafApprovers(path string) sets.String
 	FindApproverOwnersForFile(file string) string
 	IsNoParentOwners(path string) bool
+	IsAutoApproveUnownedSubfolders(directory string) bool
 	Filenames() ownersconfig.Filenames
 }
 
 // Owners provides functionality related to owners of a specific code change.
 type Owners struct {
+	// filenamesUnfiltered contains all files in a given PR, including those
+	// that do not need approval because of IsAutoApproveUnownedSubfolders
+	filenamesUnfiltered []string
+	// filenames refers to the files in a given PR, not to OWNERS files. Files
+	// that are a directory below an Owners file with IsAutoApproveUnownedSubfolders
+	// are excluded here but kept in filenamesUnfiltered.
 	filenames []string
 	repo      Repo
 	seed      int64
@@ -60,15 +67,15 @@ type Owners struct {
 
 // NewOwners consturcts a new Owners instance. filenames is the slice of files changed.
 func NewOwners(log *logrus.Entry, filenames []string, r Repo, s int64) Owners {
-	return Owners{filenames: filenames, repo: r, seed: s, log: log}
+	return Owners{filenamesUnfiltered: filenames, filenames: filenames, repo: r, seed: s, log: log}
 }
 
 // GetApprovers returns a map from ownersFiles -> people that are approvers in them
 func (o Owners) GetApprovers() map[string]sets.String {
 	ownersToApprovers := map[string]sets.String{}
 
-	for fn := range o.GetOwnersSet() {
-		ownersToApprovers[fn] = o.repo.Approvers(fn).Set()
+	for ownersFilename := range o.GetOwnersSet() {
+		ownersToApprovers[ownersFilename] = o.repo.Approvers(ownersFilename).Set()
 	}
 
 	return ownersToApprovers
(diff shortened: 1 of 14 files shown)


prow/repoowners/repoowners.go
```diff
@@ -46,6 +46,10 @@ const (
 
 type dirOptions struct {
 	NoParentOwners bool `json:"no_parent_owners,omitempty"`
+	// AllowFolderCreation will result in changes to a subpath of a given path
+	// that does not have an OWNERS file being auto-approved. This should be
+	// enabled with caution.
+	AllowFolderCreation bool `json:"allow_folder_creation,omitempty"`
```
fejta, line 52: Can we choose a different name here? `AutoApprove`?

I don't understand why "allow folder creation" would mean "auto approve files under this folder" or "do not require files under here to need approval"

prow/plugins/approve/approvers/owners.go
```diff
@@ -176,13 +183,28 @@ func (o Owners) GetSuggestedApprovers(reverseMap map[string]sets.String, potenti
 // GetOwnersSet returns a set containing all the Owners files necessary to get the PR approved
 func (o Owners) GetOwnersSet() sets.String {
 	owners := sets.NewString()
-	for _, fn := range o.filenames {
-		owners.Insert(o.repo.FindApproverOwnersForFile(fn))
+	for idx, toApprove := range o.filenames {
+		ownersFile := o.repo.FindApproverOwnersForFile(toApprove)
+		// If the ownersfile for toApprove is in the parrent folder and has AllowFolderCreation enabled, we purge
```
fejta, line 188: parent

fejta (commented):
Nice feature. Can I bike shed this name slightly?

prow/repoowners/repoowners.go
fejta: AutoAllow, AllowFolder, whatever. I'm primarily getting caught on the "Creation" word

alvaroaleman: Generally I am definitely open for suggestions on how to improve the name. The Creation stems from the fact that this option is not supposed to get inherited, i.E. if you have folder/OWNERS with this enabled and folder/subfolder/OWNERS, then changes to folder/subfolder should _not_ get autoapproved. 

Maybe `AutoApproveSubfoldersIfNoOwners`?

fejta: So if I set AllowFolderCreation for folder/OWNERS then folder/foo/whatever.go gets autoapproved because there is no folder/foo/OWNERS whereas folder/bar/something.go needs approval from someone (/OWNERS, folder/OWNERS, folder/bar/OWNERS) because folder/bar/OWNERS exists?

And furthermore that folder/bar/more/this.go will need the same approval (unless folder/bar/OWNERS or folder/bar/more/OWNERS.go sets this flag?)

alvaroaleman: Yes, exactly. The underlying usecase is to allow ppl to PR new config in without requiring approval but not to allow them to mess with other ppls config. In most cases this will be completed by additional verification that checks that the new config doesn't override existing config. It would for example help with the delegation story in https://docs.google.com/document/d/1S8DFXrvFw27xYyHMDZmtUkVvd0cbDzFI8T5H35FkCyU/edit# (currently listed as "possibly openshift-specific" but it appears that ppl think it's a good idea to do it upstream)

fejta: Gotcha! This feature sounds great.

fejta: I do not at all understand the link between "Creation" and "Not inherited by children." Can we drop that word?

Personally I think just documenting that a child OWNERS file (and the files owned by that child file) does not inherit this feature is sufficient (without adding Creating to the word).

Your doc comment calls this "auto-approved", how about using that word?

AKA `AutoApprove` `AutoApproveSubfolders` or `AutoApproveChildren` whatever.

fejta (approved):
/hold

prow/repoowners/repoowners.go
alvaroaleman: Renamed to `AutoApproveSubfolders`

alvaroaleman:
` unable to save remote image gcr.io/k8s-testimages/launcher.gcr.io/google/bazel@sha256:e006f1c3658dd11d5176d5f7a862df4d5b9c06cfe014d5b8a86bb64b20a6f8be: unable to write image layers:`

/retest

alvaroaleman:
/retest

Same error again

alvaroaleman:
```
exec ${PAGER:-/usr/bin/less} "$0" || exit 1
Executing tests from @io_k8s_repo_infra//hack:verify-boilerplate
-----------------------------------------------------------------------------
2 files have incorrect boilerplate headers:
prow/test/integration/test/crier_test.go
prow/test/integration/test/overall_test.go
```

The heck, this PR doesn't touch those files at all
/retest

alvaroaleman:
@chaodaiG I am getting the above-mentioned error even on retest, any idea what is going on? This change doesn't touch the files the check complains about at all

chaodaiG:
> @chaodaiG I am getting the above-mentioned error even on retest, any idea what is going on? This change doesn't touch the files the check complains about at all

This is the same error Cole encountered https://github.com/kubernetes/test-infra/pull/20656#issuecomment-770113854. I have looked on Cole's PR but didn't get any clue

chaodaiG:
Something mystery going on here. The PR that I reverted https://github.com/kubernetes/test-infra/pull/20759/files#diff-725d3ad3307a7361f8b1f5e1d815433c31333c8511cf142a7f46ce3045c24184R8 apparently touched the boilerplate of these 2 files, and passed presubmit test, so either the one before or the one after is correct, or both are correct?

chaodaiG:
very confusing, how did https://github.com/kubernetes/test-infra/pull/20759 pass while this one doesn't

alvaroaleman:
/retest
The boilerplate issue was fixed in https://github.com/kubernetes/test-infra/pull/20763

alvaroaleman:
image pull failure again
/retest

fejta (approved):

prow/repoowners/repoowners.go
alvaroaleman: I've renamed it yet again to `AutoApproveUnownedSubfolders` to clarify that it will not apply when a subfolder has an owners file, WDYT?

PR #20754, opened 2021-02-05
Fix cri-o download urls

odinuge:
This fixes all the cri-o builds using these configs..

cri-o use 'git rev-parse --short HEAD' as the filename, and the length
of the sha can vary. (in theory from 4 to 40 chars, but less in
practice). The output can vary between machines due to config, so it is
not stable in that sense.

This tries with length 7, 8 and 9, and works as a temporary solution to
the problem. Will connect to the cri-o people to see if they can come up
with a stable solution.

Diff is easier to understand by using `git show  --word-diff=color`


dims:
Ugh! looks like we don't have a choice in the short term ... expanding here for the record

```
/usr/bin/bash -c '
/usr/bin/curl --fail --retry 5 --retry-delay 3 --silent --show-error -o /usr/local/crio.tar.gz  https://storage.googleapis.com/k8s-conform-cri-o/artifacts/crio-$(git ls-remote https://github.com/cri-o/cri-o master | cut -c1-7).tar.gz ||
/usr/bin/curl --fail --retry 5 --retry-delay 3 --silent --show-error -o /usr/local/crio.tar.gz  https://storage.googleapis.com/k8s-conform-cri-o/artifacts/crio-$(git ls-remote https://github.com/cri-o/cri-o master | cut -c1-8).tar.gz || 
/usr/bin/curl --fail --retry 5 --retry-delay 3 --silent --show-error -o /usr/local/crio.tar.gz  https://storage.googleapis.com/k8s-conform-cri-o/artifacts/crio-$(git ls-remote https://github.com/cri-o/cri-o master | cut -c1-9).tar.gz
'
```

/approve
/lgtm

odinuge:
> Ugh! looks like we don't have a choice in the short term ... expanding here for the record

Wow, that was quick @dims, thank you so much! Tests are now working again :tada:

PR #20766, opened 2021-02-05
Tide: Limit duration of searches to 5 mins

alvaroaleman:
We see occurences were status udpate time takes more than 3.8k seconds
(about 63 minutes). Not entirely sure why this happens, but normal sync
durations for everything is < 60 seconds: https://monitoring.prow.k8s.io/d/d69a91f76d8110d3e72885ee5ce8038e/tide-dashboard?panelId=7&fullscreen&orgId=1
    
FGprof shows that its spending the time querying github. Now sure how
that can take so long, but we should under no cirumstances block the
entire controller for longer durations.

/assign @stevekuznetsov @chaodaiG 

Graph for the durations today in Openshift:
tide-wtf


chaodaiG:
lgtm. looks likely exponential backoff was under effect.

/lgtm
/approve
/hold
In case @stevekuznetsov wants to take a look

alvaroaleman:
The odds of this breaking anything are very slim and it fixes a very concrete issue so
/hold cancel

PR #20768, opened 2021-02-06
Phaino: also detect repo under GOPATH

chaodaiG:
When k8s/test-infra was not cloned under GOPATH, running phaino cannot find the source code, make it also checking under GOPATH

### DIFF
--- a/prow/cmd/phaino/local.go
+++ b/prow/cmd/phaino/local.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"go/build"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -86,7 +87,10 @@ func readRepo(ctx context.Context, path string) (string, error) {
 	}
 	def, err := findRepo(wd, path)
 	if err != nil { // If k8s/test-infra is not under GOPATH, find under GOPATH.
-		def, err = findRepo(filepath.Join(os.Getenv("GOPATH"), "src"), path)
+		pkg, err := build.Default.Import(path, build.Default.GOPATH, build.FindOnly|build.IgnoreVendor)
+		if err == nil {
+			def = pkg.Dir
+		}
 	}
 	if err != nil {
 		logrus.WithError(err).WithField("repo", path).Warn("could not find repo")
--- a/prow/cmd/phaino/local.go
+++ b/prow/cmd/phaino/local.go
@@ -35,6 +35,9 @@ import (
 )
 
 func realPath(p string) (string, error) {
+	if p == "" {
+		return "", errors.New("cannot find repo")
+	}
 	return filepath.Abs(os.ExpandEnv(p))
 }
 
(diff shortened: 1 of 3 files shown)


prow/cmd/phaino/local.go
```diff
@@ -85,6 +85,9 @@ func readRepo(ctx context.Context, path string) (string, error) {
 		return "", fmt.Errorf("workingDir: %v", err)
 	}
 	def, err := findRepo(wd, path)
+	if err != nil { // If k8s/test-infra is not under GOPATH, find under GOPATH.
+		def, err = findRepo(filepath.Join(os.Getenv("GOPATH"), "src"), path)
```
BenTheElder, line 89: GOPATH env doesn't have to be set to have a GOPATH (it has a default, try `go env GOPATH` in an environment without the env, it will be ~`~/go`).

you can locate a package how go does with: https://golang.org/pkg/go/build/
```golang
// pkg.Dir will be the path if found
pkg, err := build.Default.Import(path, build.Default.GOPATH, build.FindOnly|build.IgnoreVendor)
```

chaodaiG: Great! Done.

prow/cmd/phaino/local.go
```diff
@@ -85,6 +86,12 @@ func readRepo(ctx context.Context, path string) (string, error) {
 		return "", fmt.Errorf("workingDir: %v", err)
 	}
 	def, err := findRepo(wd, path)
+	if err != nil { // If k8s/test-infra is not under GOPATH, find under GOPATH.
+		pkg, err := build.Default.Import(path, build.Default.GOPATH, build.FindOnly|build.IgnoreVendor)
```
BenTheElder, line 90: this err shadows the err above, given the block below I think you want something like:
```suggestion
		pkg, err2 := build.Default.Import(path, build.Default.GOPATH, build.FindOnly|build.IgnoreVendor)
		err = err2
```
https://play.golang.org/p/qzBp525zKbW

BenTheElder: should add: I think the more idiomatic approach is to streamline this (perhaps breaking into some helpers), but this is at least functionally correct

fejta:
Is it common for people to check out every repo to the gopath location? This seems archaic to me.

BenTheElder:
> Is it common for people to check out every repo to the gopath location? This seems archaic to me.

I'm not sure exactly how common it is, but I still see some major go projects not really using modules yet, so probably somewhat common?

I personally still do checkout in a gopath structure just because I'm used to it / have muscle memory and it makes repo organization clearer to me than a flat hierarchy (k8s.io/test-infra vs github.io/istio/test-infra is easier for me to predict and mentally map than renaming like `k8s-test-infra` or something).

chaodaiG:
@fejta , the changes from this PR is to help users of this tool figure out where the source code lives, is best effort imo since the output will still need to be verified by user prompt. So I think it's still worthy checking GOPATH to make this step more helpful to users of prow other than k8s prow. What do you think?

fejta (approved):

PR #20765, opened 2021-02-05
Kind job updates: update test target, add more branches

BenTheElder:
1.19/1.20 have been an oversight.
unit => test happened when we made a distinction for integration tests.
these are overdue.

PR #20769, opened 2021-02-06
fix errors in new kind presubmits

BenTheElder:
https://github.com/kubernetes/test-infra/pull/20765/files#r571382897

though there's also still the odd behavior that these jobs don't show up https://github.com/kubernetes-sigs/kind/pull/648#issuecomment-774421917 🤔


BenTheElder:
TBR: really minor CI fix, blocking a repro attempt.

PR #20770, opened 2021-02-06
drop branch restriction from kind presubmits

BenTheElder:
🤦 

copy paste inheritance is the worst.


BenTheElder:
to be reviewed. tiny kind CI config fix.

PR #19856, opened 2020-11-05
Configure post-submit test for HNC

GinnyJI:
Note: this changed has not been tested, I don't know how to test it. If
this change break then post-submit tests for HNC will simply not run,
but developers can still merge in their changes because it's not
supposed to affect pre-submit tests.

Here's what I referenced for this change: https://github.com/kubernetes/test-infra/blob/9b4e473e58850ac47bcaa4b464cc6f2c4ee20c0b/config/jobs/kubernetes/sig-network/ingress-nginx.yaml#L166-L195

PR #20773, opened 2021-02-07
Add HNC postsubmits and periodics

adrianludwin:
Run HNC's end-to-end tests in a Kind cluster, using a custom-built image
defined in
https://github.com/kubernetes-sigs/multi-tenancy/tree/master/incubator/hnc/hack/prow-e2e.

Note that I've set the periodics to run every 12h; this is temporary
until we've seen a few passing runs. After that point, I'll reduce the
frequency to 48h (i.e. every two days).

Tested: on my local workstation, the custom-built image successfully
runs the latest e2e tests on Kind following the instructions in the HNC
repo (see README at the link above). I have no way to test that the Prow
configs are correct as we've cobbled them together from other tests'
examples.

### DIFF
--- a/config/jobs/kubernetes-sigs/wg-multi-tenancy/hnc-e2e.yaml
+++ b/config/jobs/kubernetes-sigs/wg-multi-tenancy/hnc-e2e.yaml
@@ -1,0 +1,66 @@
+# For more information on these tests (especially how to generate the images),
+# see
+# https://github.com/kubernetes-sigs/multi-tenancy/tree/master/incubator/hnc/hack/prow-e2e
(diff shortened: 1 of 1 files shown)


adrianludwin:
This replaces #19856

rjbez17:
Giving this a shot :)

/ok-to-test
/lgtm
/approve

adrianludwin:
/cc @rjbez17 

Looks like I've fixed all the presubmits now, can you please re-lgtm? Thanks!

rjbez17:
/lgtm

Nice work!

PR #20779, opened 2021-02-08
Rename seccomp-operator directory to security-profiles-operator

saschagrunert:
This part of the migration seemed to be missed, so we rename the
directory to follow the convention.

/assign @pjbgf @hasheddan


hasheddan (approved):
/lgtm

Issue #18943, opened 2020-08-22
pull-kubernetes-e2e-windows-gce uses excessive resources


BenTheElder:
**What happened**: 

We got a report in slack about issues with this job timing out. Upon inspection it is apparent that this job is doing a full and complete cross build of Kubernetes for all platforms / architectures in presubmit.

We do a complete _typecheck_ (already decently expensive) and have not run cross in presubmit for some time.

`make release` needs something like 50GB X 8 Cores over an hour (and a LOT of disk space), this uses way too much to be running in presubmit. In post-submit it can use the existing cross build results.

**What you expected to happen**:

Presubmits should only build what they need.

**How to reproduce it (as minimally and precisely as possible)**:

See the job definition and logs https://prow.k8s.io/?job=pull-kubernetes-e2e-windows-gce

**Please provide links to example occurrences, if any**:

**Anything else we need to know?**:

/sig windows

Please do not use this job until it is rewritten to avoid this excessively expensive build.

BenTheElder:
The full cross release build has been very expensive for O(years), we replaced it with typecheck ~2-3 years ago.

Building without bazel is acceptable but still does not require doing a complete cross build.

BenTheElder:
/remove-lifecycle rotten
please fix this or remove the job, this is still true: https://github.com/kubernetes/test-infra/blob/94bde9b43d8af13762777ee060954f2a90ea1d5d/config/jobs/kubernetes/sig-windows/windows-gce.yaml#L472
@kubernetes/sig-windows-bugs

BenTheElder:
@kubernetes/sig-windows-bugs I've filed a nominal fix myself https://github.com/kubernetes/test-infra/pull/21068

PR #20789, opened 2021-02-08
Disable unit tests for Windows e2e

jingxu97:
disable unit tests when running e2e tests for Windows to avoid it
blocking e2e tests

### DIFF
--- a/config/jobs/kubernetes/sig-windows/windows-gce.yaml
+++ b/config/jobs/kubernetes/sig-windows/windows-gce.yaml
@@ -581,6 +581,10 @@ presubmits:
           value: "win2019"
         - name: PREPULL_YAML
           value: "prepull-head.yaml"
+        # Disable unit test for Windows for now.
+        # TODO: Should remove it after unit tests are fixed.
+        - name: KUBE_RELEASE_RUN_TESTS
+          value: "n"
         image: gcr.io/k8s-testimages/kubekins-e2e:v20210302-a6bf478-master
         # docker-in-docker needs privileged mode
         securityContext:


jingxu97:
cc @pjh

config/jobs/kubernetes/sig-windows/windows-gce.yaml
```diff
@@ -579,6 +579,8 @@ presubmits:
           value: "win2019"
         - name: PREPULL_YAML
           value: "prepull-head.yaml"
+        - name: KUBE_RELEASE_RUN_TESTS
```
BenTheElder, line 582: please stop running this with `--build=release` in presubmit, it's extremely expensive and slow:  https://github.com/kubernetes/test-infra/issues/18943

BenTheElder:
This now uses quick release anyhow. https://github.com/kubernetes/test-infra/pull/21068

config/jobs/kubernetes/sig-windows/windows-gce.yaml
```diff
@@ -581,6 +581,10 @@ presubmits:
           value: "win2019"
         - name: PREPULL_YAML
           value: "prepull-head.yaml"
+        # Disable unit test for Windows for now.
+        # TODO: Should remove it after unit tests are fixed.
```
pjh, line 585: What is broken about the unit tests? Is there an issue tracking this? Will they ever be fixed, or will this never work for Windows?

BenTheElder: 1. they're broken on windows, there is an issue in k/k
2. when running e2e we should not be running the unit tests
3. these run when building a release by default. they do not when running a quick release

BenTheElder:
this PR should be unnecessary now https://github.com/kubernetes/test-infra/pull/21068

dchen1107:
@BenTheElder Thanks!

Issue #20751, opened 2021-02-05
[BUG] Deck is NOT working properly after the new changes.


droslean:
We recently bump prow to run with changes including in https://github.com/kubernetes/test-infra/compare/567fc1ab64...b06ec788ea and it caused `deck` to stop being functional.

```sh
$ curl https://prow.ci.openshift.org/static/prow_bundle.min.js
404 page not found
```

/cc @kubernetes/test-infra-admins @kubernetes/test-infra-maintainers

alvaroaleman:
Bisect shows its 81c5adb48839d6af0e96b07ba2956f7b727f8c94

chaodaiG:
@fejta , the bazel 3.4.1 PR and seemed affecting deck

chaodaiG:
Bumped k8s prow https://github.com/kubernetes/test-infra/pull/20741 after reverting of #20651, everything looks fine to me

droslean:
It seems that https://github.com/kubernetes/test-infra/commit/4d6791131bd487a3b9c0375dfff7c5c46328d06c broke it again.

droslean:
Error from browsers console:
```sh
Uncaught SyntaxError: import declarations may only appear at top level of a module
```

alvaroaleman:
@fejta next time you introduce this, can you please start a deck instance built with the new bazel version (locally via docker or behind a canary ingress in your production) and click through it to make sure that we do not break it a third time?

cjwagner:
Trying again: https://github.com/kubernetes/test-infra/pull/20824

alvaroaleman:
everything appears to be working now, thanks!
/close

PR #20790, opened 2021-02-08
Use bazel 3.4.1

fejta:
Second commit includes fix for #20751

/assign @chaodaiG


prow/cmd/deck/validate-static-files.sh
```diff
@@ -0,0 +1,45 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+# check-file ensures that all static links are valid
+check-file() {
+  for path in $(grep -h -o -E '/static[^"]+' $1); do
```
chaodaiG, line 22: Do you need a `-r` here, seems like needed for grepping `/static` directory

fejta: No, it is grepping `$1`, looking for strings that start with `/static`, such as `/static/prow_bundle.min.js` that it couldn't find previously.

chaodaiG: `$1` is from https://github.com/kubernetes/test-infra/pull/20790/commits/87fea4ce8dabcbfef39fb009f3e99ff4c48601f7#diff-4dbb2eb4801e310971dae75e09a6bd4c65a171b48b7f317ac1722ac0cb7ef44cR51-R52, the second one is the absolute path of `//prow/cmd/deck/static`, which is a directory. I have tried this command on my local machine and it complained:

```
$ grep -h -o -E '/static[^"]+' ./prow/cmd/deck/static
grep: ./prow/cmd/deck/static: Is a directory
```

Does bazel do something differently?

fejta: The `locations` function in the `$(locations //prow/cmd/deck/static)` line you are referring to translates the `//prow/cmd/deck/static` label into the list of paths that label contains.

See https://docs.bazel.build/versions/master/be/general.html#genrule_examples for some documentation on this.

chaodaiG: Ah. I see, good to learn.

fejta: Looks like this is the better link: https://docs.bazel.build/versions/master/be/make-variables.html#predefined_label_variables

PR #20792, opened 2021-02-08
Skip K8s 1.20 tests for external-snapshotter@release-2.1 branch

ialidzhikov:
As discussed in https://github.com/kubernetes-csi/external-snapshotter/pull/471, the K8s 1.20 e2e tests are not appropriate for `release-2.1` branch of `kubernetes-csi/external-snapshotter` - the tests are assuming apiversion v1 but the only available apiversion in kubernetes-csi/external-snapshotter@release-2.1 is v1beta1.

### DIFF
--- a/config/jobs/kubernetes-csi/external-snapshotter/external-snapshotter-config.yaml
+++ b/config/jobs/kubernetes-csi/external-snapshotter/external-snapshotter-config.yaml
@@ -183,7 +183,7 @@ presubmits:
     optional: true
     decorate: true
     skip_report: false
-    skip_branches: ["^(k8s_1.12.0-beta.1|release-0.4|release-1.0)$"]
+    skip_branches: ["^(k8s_1.12.0-beta.1|release-0.4|release-1.0|release-2.1)$"]
     labels:
       preset-service-account: "true"
       preset-dind-enabled: "true"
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes-csi/external-snapshotter/external-snapshotter-config.yaml
```diff
@@ -183,7 +183,7 @@ presubmits:
     optional: true
     decorate: true
     skip_report: false
-    skip_branches: ["^(k8s_1.12.0-beta.1|release-0.4|release-1.0)$"]
+    skip_branches: ["^(k8s_1.12.0-beta.1|release-0.4|release-1.0|release-2.1)$"]
```
pohly, line 186: These files are generated automatically. The https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes-csi/gen-jobs.sh script has to be updated to exclude branches on a per-job basis.

There's currently no mechanism in the script to configure that and I don't have a good idea for how to add it.

/hold

msau42: Sorry I don't think skipping the job entirely is the right fix. Older releases using the beta api should still function. Can we fix the kubernetes e2e version to 1.19?

ialidzhikov:
I don't have capacity to continue looking into this issue. I created https://github.com/kubernetes-csi/external-snapshotter/pull/466 that contains some basic information with PRs in which we discussed the issue.

/close

config/jobs/kubernetes-csi/external-snapshotter/external-snapshotter-config.yaml
pohly: That should work. This could go into .prow.sh as in https://github.com/kubernetes-csi/external-snapshotter/pull/471

ialidzhikov: Thanks @pohly . Can you file a new PR that is based on kubernetes-csi/external-snapshotter#471?

PR #20600, opened 2021-01-25
Add containerd presubmits

ravisantoshgudimetla:
Add containerd presubmits for pull-kubernetes-e2e-azure-disk-windows, pull-kubernetes-e2e-azure-file-windows

PR #20795, opened 2021-02-09
Make containerd tests mandatory on csi-disk and csi-file for azure

ravisantoshgudimetla:
As of now, we're trying to catch azure csi issues in k8s/k8s. I think catching them earlier in csi disk and file repos would be ideal. This change also makes the tests mandatory. Let me know if you think otherwise.

@jsturtevant @jayunit100 @andyzhangx @aravindhp

### DIFF
--- a/config/jobs/kubernetes-sigs/azuredisk-csi-driver/azuredisk-csi-driver-config.yaml
+++ b/config/jobs/kubernetes-sigs/azuredisk-csi-driver/azuredisk-csi-driver-config.yaml
@@ -601,7 +601,7 @@ presubmits:
       testgrid-num-columns-recent: '30'
   - name: pull-azuredisk-csi-driver-e2e-windows-containerd
     decorate: true
-    always_run: false
+    always_run: true
     path_alias: sigs.k8s.io/azuredisk-csi-driver
     branches:
     - master
--- a/config/jobs/kubernetes-sigs/azurefile-csi-driver/azurefile-csi-driver-config.yaml
+++ b/config/jobs/kubernetes-sigs/azurefile-csi-driver/azurefile-csi-driver-config.yaml
@@ -317,7 +317,7 @@ presubmits:
       testgrid-num-columns-recent: '30'
   - name: pull-azurefile-csi-driver-e2e-windows-containerd
     decorate: true
-    always_run: false
+    always_run: true
     path_alias: sigs.k8s.io/azurefile-csi-driver
     branches:
     - master


jayunit100:
hiya ok - so who owns the overall health of CSI ? 
Whats the history of these tests ?
Are they known to be stable?

ravisantoshgudimetla:
cc @ddebroy @ksubrmnn @marosset @jingxu97

config/jobs/kubernetes-sigs/azuredisk-csi-driver/azuredisk-csi-driver-config.yaml
```diff
@@ -601,7 +601,7 @@ presubmits:
       testgrid-num-columns-recent: '30'
   - name: pull-azuredisk-csi-driver-e2e-windows-containerd
     decorate: true
-    always_run: false
```
andyzhangx, line 604: pls keep set as `false` for now since there is issue in `pull-azuredisk-csi-driver-e2e-windows-containerd` test:
https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/98522/pull-kubernetes-e2e-azure-disk-windows-containerd/1355067374688014336

andyzhangx: wait, this is a wrong place, it's for azure disk csi driver, there is still issue running this driver on windows containerd

andyzhangx:
@ravisantoshgudimetla  the Azure Disk & File CS drivers are already running e2e tests on Linux containerd node, while it still has issue running on Windows containerd.

config/jobs/kubernetes-sigs/azuredisk-csi-driver/azuredisk-csi-driver-config.yaml
ravisantoshgudimetla: Can you please elaborate on the issue? I want to understand if this was already known.

ravisantoshgudimetla:
> @ravisantoshgudimetla the Azure Disk & File CS drivers are already running e2e tests on Linux containerd node, while it still has issue running on Windows containerd.

My idea in opening this PR to catch the issues early in the development by having these presubmits in azure csi repos instead of catching them in k8s/k8s.

xref: https://github.com/kubernetes/test-infra/pull/20600

config/jobs/kubernetes-sigs/azuredisk-csi-driver/azuredisk-csi-driver-config.yaml
andyzhangx: Azure disk & file CSI driver does not work on windows containerd now, need to fix that first.

here are the tracking issues:
https://github.com/kubernetes-sigs/azuredisk-csi-driver/issues/547
https://github.com/kubernetes-sigs/azurefile-csi-driver/issues/382

andyzhangx:
BTW, we could enable in-tree azure file test on windows containerd now, it's working and stable: https://github.com/kubernetes/test-infra/pull/20800

PR #20797, opened 2021-02-09
Cherrypicker: Improve Logging

alvaroaleman:
* Initialize log so it gets standard component field and json formatting
* Actually propagate logger back up by adding fields to the value of the
  logger rather than assigning a new pointer to the pointer var which
  does not do anything wrt backpropagating
* Use fields instead of printf-style logging
* Add some error wrapping
* Aggregate errors rather than swallowing them when getting follow-up
  errors
* Parallelize tests, because they are uber slow _and_ executed serially

### DIFF
--- a/prow/external-plugins/cherrypicker/server.go
+++ b/prow/external-plugins/cherrypicker/server.go
@@ -30,6 +30,7 @@ import (
 
 	"github.com/sirupsen/logrus"
 
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/test-infra/prow/config"
 	cherrypicker "k8s.io/test-infra/prow/external-plugins/cherrypicker/lib"
 	"k8s.io/test-infra/prow/git/v2"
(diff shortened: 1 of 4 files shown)


prow/external-plugins/cherrypicker/server.go
```diff
@@ -168,7 +167,7 @@ func (s *Server) handleIssueComment(l *logrus.Entry, ic github.IssueCommentEvent
 	commentAuthor := ic.Comment.User.Login
 
 	// Do not create a new logger, its fields are re-used by the caller in case of errors
-	l = l.WithFields(logrus.Fields{
+	*l = *l.WithFields(logrus.Fields{
```
alvaroaleman, line 170: This didn't do what the comment above claimed, `l` holds a pointer, the assignment before made it hold a new pointer, but that won't change the original pointer. Changing it like this will instead change the value behind the pointer, so the caller will actually get these fields as well.

matthyx:
Nice, @alvaroaleman should we create an umbrella task and improve logging in every prow component?
/lgtm
/hold

petr-muller:
LGTM

matthyx:
> Nice, @alvaroaleman should we create an umbrella task and improve logging in every prow component?

What do you think?

alvaroaleman:
> Nice, @alvaroaleman should we create an umbrella task and improve logging in every prow component

In most (all?) components its pretty good already I believe. All the core components (Deck, Hook, Sinker, Crier, Plank) definitely have good logging. I think this was an outlier because its its own binary and not something ppl want to look at often :upside_down_face: 

/hold cancel

alvaroaleman:
Why does the `needs-rebase` label not get removed :(

alvaroaleman:
I did a no-op force push and the label is still there :/

PR #20802, opened 2021-02-09
WIP: Preload scratch images for kubemark tests

lyzs90:
Details: https://github.com/kubernetes/kubernetes/issues/90833

cc @mm4tt @wojtek-t 

I'm not sure if you can trigger this job from a non-master branch / fork. I can duplicate this presubmit so it wouldn't block anyone else if it goes wrong.


mm4tt:
I think it should be fine to gradually enable it in all kubemark jobs. But, to be on the safe side, let's send first a "DO NOT SUBMIT" PR against perf-tests repo and enable the preloading in this line - https://github.com/kubernetes/perf-tests/blob/6c217264a3a76ad86a13d004045849349aca8080/clusterloader2/run-e2e.sh#L37

I'd like to see how it will work with the existing presubmits

PR #20813, opened 2021-02-09
Move autobumper out of experimental directory

mpherman2:
The autobumper code is used regularly both within k8s/test-infra, and externally, so we are moving it out of the experimental directory to make clear that changes to the code should make an effort to remain backwards compatible.


alvaroaleman (approved):

spiffxp (approved):
/approve
/lgtm
Remove /hold when you're ready for this to merge

mpherman2:
/hold
we discussed and we decided to move this to prow/cmd instead of robots/

cjwagner (approved):
This LGTM, but I'd like to have Alvaro sign off on this again since his approval is old, this breaks external import paths, we've changed the destination of this migration, and we've changed the release process to include this with the Prow components.
/hold

mpherman2:
> This LGTM, but I'd like to have Alvaro sign off on this again since his approval is old, this breaks external import paths, we've changed the destination of this migration, and we've changed the release process to include this with the Prow components.
> /hold

@alvaroaleman can you take a look at this again.

alvaroaleman (approved):

stevekuznetsov:
Does Go have a way to chase these refs? Weird to see a break on import after bumping vendor :|

PR #20817, opened 2021-02-10
Enable testing huge services in master performance jobs

tosi3k:
Enable testing huge services in CI (ref https://github.com/kubernetes/perf-tests/pull/1681).

/sig scalability
/assign @wojtek-t

### DIFF
--- a/config/jobs/kubernetes/sig-scalability/sig-scalability-release-blocking-jobs.yaml
+++ b/config/jobs/kubernetes/sig-scalability/sig-scalability-release-blocking-jobs.yaml
@@ -89,6 +89,7 @@ periodics:
       - --metadata-sources=cl2-metadata.json
       - --env=CL2_LOAD_TEST_THROUGHPUT=50
       - --env=CL2_DELETE_TEST_THROUGHPUT=30
+      - --env=CL2_ENABLE_HUGE_SERVICES=true
       # Overrides CONTROLLER_MANAGER_TEST_ARGS from preset-e2e-scalability-periodics.
       - --env=CONTROLLER_MANAGER_TEST_ARGS=--profiling --kube-api-qps=100 --kube-api-burst=100 --endpointslice-updates-batch-period=500ms
       - --test=false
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes/sig-scalability/sig-scalability-release-blocking-jobs.yaml
```diff
@@ -89,6 +89,7 @@ periodics:
       - --metadata-sources=cl2-metadata.json
       - --env=CL2_LOAD_TEST_THROUGHPUT=50
       - --env=CL2_DELETE_TEST_THROUGHPUT=30
+      - --env=CL2_ENABLE_HUGE_SERVICES=true
```
wojtek-t, line 92: Why only 5k job? Why not also 100-node one?

tosi3k: At 100 nodes scale we have 1000 scheduler throughput pods which are distributed among 10 namespaces as deployments containing 100 pods each ([ref](https://github.com/kubernetes/perf-tests/blob/35b6c7a0e7f13bd0095e48968355f38a14d164e1/clusterloader2/testing/load/config.yaml#L82-L84)).

This means the "huge services" will front 100 pods which corresponds to a single EndpointSlice object per service, hence it's not really a huge service anymore (and the 250 pods services fronting the big deployments are probably a better way of testing the EPSes here).

wojtek-t: I know.
I would say - it's "huge" when compared to cluster size.

I'm more asking - not what happens, but is there a good reason to not enable it there?

tosi3k: Discussed this offline and the conclusion is that we want to enable that at 100 nodes scale to have a uniform setup for all scales.

PR #20824, opened 2021-02-10
Update to bazel 3.4.1

fejta:
Also specify "window" as the top-level "this" argument.

Second commit shows delta from previous attempt. Validated that the main page loads and seems to work correctly as well as other things (plugin help, tide history, etc)

/assign @cjwagner


cjwagner (approved):
This looks right to me and sounds like it has been sufficiently validated. Any concerns with this before we try it again Alvaro?
/assign @alvaroaleman 
/hold

alvaroaleman:
Lets give it another shot
/hold cancel

Issue #20760, opened 2021-02-05
Migrate jobs away from deprecated bootstrap usage


spiffxp:
This is an umbrella issue to track migrating job configs away from bootstrap and instead use [pod-utils](https://github.com/kubernetes/test-infra/blob/master/prow/pod-utilities.md).

Bootstrap has been deprecated for a few years at this point, and jobs that use bootstrap.py have been unsupported by test-infra-oncall for at least 1.5 years (ref: https://github.com/kubernetes/test-infra/pull/14392).  Pod-utils are the documented and supported way of writing prowjobs: https://github.com/kubernetes/test-infra/tree/master/config/jobs#job-cookbook

One way to be helpful: PR in a guide on how to do this!  Relevant content to source from:
- @dims put together a guide https://gist.github.com/dims/c1296f8ed42238baea0a5fcae45f4cf4
- @rikatz went through this for pull-kubernetes-node-e2e
  - https://github.com/kubernetes/test-infra/pull/19942
  - https://github.com/kubernetes/test-infra/pull/19947
  - https://github.com/kubernetes/test-infra/pull/19948
  - https://github.com/kubernetes/test-infra/pull/19949
  - https://github.com/kubernetes/test-infra/pull/19951

Another way to help: produce a tool, script or dashboard that keeps track of what has and has not been migrated.  We could either link to its results from this issue, or periodically update this issue with the results.

The most concrete way to help: migrate the jobs you or your SIG owns

/lifecycle frozen
/priority important-longterm
/sig testing
Once we have tracking of which sigs own jobs that use bootstrap, we can tag with appropriate sigs.
/area jobs
/area bootstrap
/area prow/pod-utils
/help

spiffxp:
As an FYI, jobs that migrate away from bootstrap will lack an Infra-Commit column in testgrid: https://github.com/kubernetes/test-infra/issues/20650.  This is not a blocker to migration, and is probably a good thing (one less thing to bisect).

I strongly suspect kubernetes/test-infra code being checked out in a job's container isn't actually necessary for the vast majority of jobs that make the switch away from bootstrap.

spiffxp:
Things to help track our status:
- https://github.com/kubernetes/test-infra/pull/20825 added tests to log jobs that are deprecated
- https://github.com/kubernetes/test-infra/pull/22838 added summary lines to a number of tests

dims:
xref: https://github.com/kubernetes/test-infra/issues/30846

[...]

PR #20825, opened 2021-02-10
Add tests to log jobs using deprecated bootstrap, scenarios

spiffxp:
Same pattern as CI policy changes... start with "should" tests that log jobs using deprecated things.

This can be used to build a help-wanted list, or shard out issues to specific job owners, etc.

/cc @amwat @MushuEE

### DIFF
--- a/config/tests/jobs/jobs_test.go
+++ b/config/tests/jobs/jobs_test.go
@@ -1109,3 +1109,86 @@ func TestKubernetesE2eJobsMustExtractFromK8sInfraBuckets(t *testing.T) {
 		}
 	}
 }
+
+// Prow jobs should use pod-utils instead of relying on bootstrap
+// https://github.com/kubernetes/test-infra/issues/20760
+func TestKubernetesProwJobsShouldUsePodUtils(t *testing.T) {
+	total := 0
+	fails := 0
+	for _, job := range allStaticJobs() {
+		total++
+		// Only consider Pods
+		// TODO(https://github.com/kubernetes/test-infra/issues/14343): remove kubeflow exemption when job configs migrated
+		if job.Spec == nil || strings.HasPrefix("kubeflow", job.Name) {
+			continue
+		}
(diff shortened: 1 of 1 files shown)


spiffxp:
FWIW, here's what I found running tests locally
```
jobs_test.go:1150: 403/2136 jobs do not use pod-utils
jobs_test.go:1192: 835/2136 jobs using deprecated scenarios
```

spiffxp:
/test pull-test-infra-bazel
flake `read tcp 10.60.209.54:51456->74.125.70.128:443: read: connection reset by peer`

spiffxp:
ref: https://github.com/kubernetes/test-infra/issues/20760 for bootstrap migration

MushuEE (approved):
/lgtm

If I can find time I would love to build a struct for each job that can hold a little more metadata and help determine owners to see that Sigs have the most depreciated jobs.

spiffxp:
>If I can find time I would love to build a struct for each job that can hold a little more metadata and help determine owners to see that Sigs have the most depreciated jobs.

I was sortof headed in that direction in experiment/prowjob-report.  There are a number of policy changes we want rolled out across all jobs, but we'd rather delegate that work out to the appropriate job owners. e.g.
- setting alert e-mails for all jobs
- setting pod resource limits for all jobs that run in k8s-infra
- jobs shouldn't use bootstrap
- jobs shouldn't use scenarios
- jobs shouldn't use kubetest

We have a bunch of distinct things that could imply ownership, I'm not sure how consistent they are:
- testgrid dashboard (but there can be many, I was guessing first sig- dashboard that wasn't sig-release)
- testgrid alert e-mail (but there are many non-sig e-mails)
- path for job / OWNER label for job (but this misses ownership of job results not produced by prow.kl8s.io that we ingest with kettle and testgrid)

ATM I wonder if it'd be best to try making path source of truth, and automatically added some labels at push-to-prow-time that could then end up in finished.json's metadata

PR #20826, opened 2021-02-11
Fix cherrypicker when the fork is not named the same as the repo

mpherman2:
If the fork name was not the same as the repo name, we were not returning the right value for ensureForkExists

### DIFF
--- a/prow/external-plugins/cherrypicker/server.go
+++ b/prow/external-plugins/cherrypicker/server.go
@@ -556,7 +556,8 @@ func (s *Server) ensureForkExists(org, repo string) (string, error) {
 	fork := s.botUser.Login + "/" + repo
 
 	// fork repo if it doesn't exsit
-	if _, err := s.ghc.EnsureFork(s.botUser.Login, org, repo); err != nil {
+	repo, err := s.ghc.EnsureFork(s.botUser.Login, org, repo)
+	if err != nil {
 		return repo, err
 	}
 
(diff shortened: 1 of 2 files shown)


chaodaiG:
Looks good to me, thank you for fixing this!
/lgtm
/approve
/hold
So that @stevekuznetsov can take a look

prow/external-plugins/cherrypicker/server.go
```diff
@@ -556,7 +556,7 @@ func (s *Server) ensureForkExists(org, repo string) (string, error) {
 	fork := s.botUser.Login + "/" + repo
 
 	// fork repo if it doesn't exsit
-	if _, err := s.ghc.EnsureFork(s.botUser.Login, org, repo); err != nil {
+	if repo, err := s.ghc.EnsureFork(s.botUser.Login, org, repo); err != nil {
```
alvaroaleman, line 559: The only case in which this code will use the name from `EnsureFork` is if there was an error in which case we don't do anything anyways, this doesn't address Steves comment at all

alvaroaleman: I suggest you write some tests for this method to be confident your changes do what you think they do

mpherman2: Done. Thanks for the feedback

alvaroaleman (approved):
Thanks!

Issue #20816, opened 2021-02-10
jobs erroring out due to failure to fetch token


BenTheElder:
**What happened**:

Seeing jobs fail early on with no logs more and more often lately, and not signs we've seen before like "failed scheduling" (need to increase nodes) or github clone flakes (kubernetes/kubernetes is a huge clone, still need to mitigate that #https://github.com/kubernetes/test-infra/issues/18226).

Instead we're seeing: 

```
 - containerID: containerd://2372ffea321a2dec129d222dc4b43ee8774b4c0528a2e36d9271c6a51819b35e
    image: gcr.io/k8s-prow/initupload:v20210209-12c3370f77
    imageID: gcr.io/k8s-prow/initupload@sha256:4108bc2d7ddb6a2272d525b060d305edd5f9575430ca022ceef22475140ce713
    lastState: {}
    name: initupload
    ready: false
    restartCount: 0
    state:
      terminated:
        containerID: containerd://2372ffea321a2dec129d222dc4b43ee8774b4c0528a2e36d9271c6a51819b35e
        exitCode: 1
        finishedAt: "2021-02-10T07:02:42Z"
        message: |
[...]

PR #20827, opened 2021-02-11
Deck reports init container failure

chaodaiG:
Part of #20816

When a prowjob failed due to init container failure, there is no log or any hint in deck for user to figure out what's going on. Make deck report this.


alvaroaleman (approved):
This is great, thanks!

PR #20833, opened 2021-02-11
add SuspendJob to presubmit and periodic alpha jobs

adtac:
See KEP: kubernetes/enhancements#2234

This PR enables the Feature:SuspendJob e2e tests added in https://github.com/kubernetes/kubernetes/pull/98727 which adds an alpha feature gate.

/sig apps
/hold until https://github.com/kubernetes/kubernetes/pull/98727 is merged
/cc @ahg-g @alculquicondor


adtac:
/hold cancel

since we don't need to wait for the implementation PR to merge. In fact, merging this before that PR will tell us if the alpha e2e test passes.

alculquicondor:
Wouldn't the binary fail to start if the feature is not recognized because it hasn't been merged yet?

adtac:
No, this change simply adds "[Feature:SuspendJob]" to the regex filter for tests. The feature gate is implicitly enabled by the AllAlpha=True which will include the implementation PR's new feature gate when it merges.

alculquicondor:
gotcha

/lgtm

PR #20837, opened 2021-02-11
Cherrypicker: Do not react to unrelated label events

alvaroaleman:
If a PR was labeled but doesn't have the cherrypick label, we should just do nothing


hi-rustin (approved):
/hold 

Is there a way to test this change?

PR #20839, opened 2021-02-11
Deck: surface all initcontainer errors

chaodaiG:
#20827 enabled deck to report initcontainer error, but it only reported the last failed container instead of all, correcting that in this PR


chaodaiG:
See an example https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/98596/pull-kubernetes-bazel-test/1359930862556155904. It failed both `entrypoint` and `initupload`, but only reported `entrypoint` which is less relevant in this case.

spiffxp (approved):
/approve
/lgtm

Issue #20840, opened 2021-02-11
Surface image pull failures in spyglass metadata lens.


cjwagner:
**What would you like to be added**:
If a ProwJob fails with a pod pending timeout due to not being able to pull the test image, the metadata lens only hints:
> Job execution failed: Pod pending timeout.

From the pod lens we can see the pod events which indicate that the failed image pull is the culprit, but it would be preferable to improve the metadata lens hint to point to this problem more directly. We should also update the generic "Pod pending timeout." message to recommend inspecting the pod lens since that will likely contain events that explain the nature of the pod pending timeout.

**Why is this needed**:
`Job execution failed: Pod pending timeout.` is not actionable, pointing to the pod lens will help that. Failure to pull images is a particularly common failure mode that would be helpful to expose prominently.

/help
/area prow/spyglass prow/deck

PR #20841, opened 2021-02-11
Deck: also capture 'ErrImagePull' as reason for prow job failures

chaodaiG:
Fixes: https://github.com/kubernetes/test-infra/issues/20840


cjwagner (approved):
Thanks Chao, I forgot that we try to expose image pull failures already.

PR #20845, opened 2021-02-11
Update terraform config for k8s-prow based on current state

chaodaiG:
node types were switched from n1-* to e2-*, also deleted the n1-standard-4 nodepool which is a duplicate of n1-standard-8 nodepool


spiffxp (approved):
/lgtm
/hold
in case you want anyone else to review, remove if not

PR #20857, opened 2021-02-12
Update test-info post submit to build the setcap image.

vinayakankugoyal:
PR https://github.com/kubernetes/release/pull/1684 added a new setcap image. We need to start building this image and pushing it to k8s.gcr.io along with the other build-images. We plan to use it in https://github.com/kubernetes/kubernetes/pull/96134.


BenTheElder:
/lgtm
/approve
thanks!

PR #20858, opened 2021-02-13
prow/deck: fix pr status (GHE)

sbueringer:
Previously the check header was only send for GraphQL calls. The pr
status page also needs calls to the regular check endpoint. Thus, we
also have to send the header in this case.

xref: https://github.com/kubernetes/test-infra/pull/19806

### DIFF
--- a/prow/github/client.go
+++ b/prow/github/client.go
@@ -4109,6 +4109,7 @@ func (c *client) ListCheckRuns(org, repo, ref string) (*CheckRunList, error) {
 
 	var checkRunList CheckRunList
 	_, err := c.request(&request{
+		accept:    "application/vnd.github.antiope-preview+json",
 		method:    http.MethodGet,
 		path:      fmt.Sprintf("/repos/%s/%s/commits/%s/check-runs", org, repo, ref),
 		org:       org,


prow/github/client.go
```diff
@@ -562,7 +562,7 @@ func newClient(
 				Timeout: maxRequestTime,
 				Transport: &oauth2.Transport{
 					Source: newReloadingTokenSource(getToken),
-					Base:   graphQLTransport,
+					Base:   baseRoundTripperWithPreviewHeaders,
```
alvaroaleman, line 565: In the v3 client we usually set them on the methods that need them rather than globally: https://github.com/kubernetes/test-infra/blob/2efad71f085991710643e572dbde606377a807b1/prow/github/client.go#L4039

sbueringer: Okay, so I assume this means in tide when we call the check method.

I could also add it here globally and only add the header when the calls goes to the check API. This would make it easier to avoid future problems when the checks API is called from another place.

I assume you prefer the first option. WDYT?

alvaroaleman: Yeah, please just add it to the relevant methods

sbueringer: @alvaroaleman PTAL :)

alvaroaleman (approved):

PR #20860, opened 2021-02-15
kops grid: skip Kubectl cluster-info dump should check if cluster-info dump succeeds

justinsb:
It may be the cause of our timeouts.

### DIFF
--- a/config/jobs/kubernetes/kops/build_grid.py
+++ b/config/jobs/kubernetes/kops/build_grid.py
@@ -320,6 +320,12 @@ def build_test(cloud='aws',
     if networking == "cilium":
         # https://github.com/cilium/cilium/issues/10002
         skip_regex += r'|TCP.CLOSE_WAIT'
+
+    # Test may be causing timeouts:
+    #  Kubectl cluster-info dump should check if cluster-info dump succeeds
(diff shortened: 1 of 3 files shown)


justinsb:
Here's why I think it is the cause....

https://k8s-testgrid.appspot.com/kops-distro-u2004#kops-grid-calico-u2004-k19-docker has a lot of timeouts.

Tests that timeout (and sadly any other tests on that ginkgo process) will show up as white boxes.  The ginkgo process assignment is random IIUC.  But the test which is slow and causes the timeout should be a consistent white row, as it will always be in the shard containing itself :-)

Visually, that looks to be "Kubernetes e2e suite.[sig-cli] Kubectl client Kubectl cluster-info dump should check if cluster-info dump succeeds"

I did dig into one log by hand, and there is additional information in there; we can see most of the other "white box" tests passing in the log.  There are three(?) others that also don't show up, but they aren't consistent across other runs, so this one seems likely.

We see _nothing_ in the logs from the slow test, because ginkgo buffers the test output AIUI.  It does this because otherwise the test results would be interleaved in the output file.

cc @rifelpet

rifelpet:
I was looking for similar patterns in our other jobs and noticed that test stand out as well, I think this is as good of a place as any to start the test bisection :)

/lgtm

PR #20868, opened 2021-02-16
prow/plugins: allow label-based automatic posting of comments

nikhita:
This commit allows automatically posting a comment based on the presence
of certain labels (labels matching a particular regexp).

Some examples where this feature can be useful:

1. If an issue contains the `kind/feature` label, the bot can add a
comment pointing to the KEPs process (https://github.com/kubernetes/kubernetes/pull/98867).

2. If an issue contains the `kind/support` label, the bot can add a
comment pointing to the support resources. The bot can also automatically
close the issue with a `/close` comment (https://github.com/kubernetes/community/issues/5435#issuecomment-774974714).

This feature is being added to the `require-matching-label` plugin.
While this feature isn't inherently related to requiring a particular
label, porting this feature to a new/different plugin requires
duplication of code.

/kind feature
/sig contributor-experience
/cc @mrbobbytables @cblecker 
/assign @cblecker

### DIFF
--- a/prow/plugins/require-matching-label/require-matching-label.go
+++ b/prow/plugins/require-matching-label/require-matching-label.go
@@ -90,13 +90,21 @@ func helpProvider(config *plugins.Configuration, _ []config.OrgRepo) (*pluginhel
 				MissingComment: "Please add a label referencing the kind.",
 				GracePeriod:    "5s",
 			},
+			{
+				Org:             "org",
+				Repo:            "repo",
+				Issues:          true,
+				Regexp:          "^kind/feature$",
+				MatchingComment: "Please create a KEP instead.",
+				GracePeriod:     "5s",
+			},
 		},
 	})
 	if err != nil {
 		logrus.WithError(err).Warnf("cannot generate comments for %s plugin", pluginName)
 	}
 	return &pluginhelp.PluginHelp{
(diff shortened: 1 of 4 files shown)


petr-muller (approved):

prow/plugins/require-matching-label/require-matching-label.go
```diff
@@ -236,11 +244,23 @@ func handle(log *logrus.Entry, ghc githubClient, cp commentPruner, configs []plu
...
-					log.WithError(err).Error("Failed to create comment.")
+					log.WithError(err).Error("Failed to create missing comment.")
 				}
 			}
 		}
 
+		if cfg.MatchingComment != "" {
+			if hasMatchingLabel {
+				msg := plugins.FormatSimpleResponse(e.author, cfg.MatchingComment)
+				if err := ghc.CreateComment(e.org, e.repo, e.number, msg); err != nil {
```
cjwagner, line 255: We need to avoid creating a comment if one has already been created, otherwise this will spam the PR on unrelated changes.

cjwagner (commented):
> This feature is being added to the `require-matching-label` plugin.
While this feature isn't inherently related to requiring a particular
label, porting this feature to a new/different plugin requires
duplication of code.

We could share code between two plugins rather than combining the logic for separate use cases into one. In particular, the `require-matching-label` plugin config requires a `MissingLabel` to be specified: https://github.com/kubernetes/test-infra/blob/5e490580b8a60705d24c62079a1f65ec1531cf5b/prow/plugins/config.go#L668-L670 but that doesn't really make sense for the use case of commenting when a specific label is present.

Issue #17505, opened 2020-05-04
Prow's /view is a slow to load GCE 5k results


mborsz:
**What happened**:
I'm trying to debug failing gce 5000 nodes test, e.g.: https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-e2e-gce-scale-performance/1256992291344617478

I noticed that it takes ~20 seconds to load a single page like that.
While the debugging failing test usually involves checking multiple test runs, this is quite inconvenient to wait for each view to load.
**What you expected to happen**:
I expected that this page would load in < 1s.
**How to reproduce it (as minimally and precisely as possible)**:
Just go to https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-e2e-gce-scale-performance/1256992291344617478
**Please provide links to example occurrences, if any**:

**Anything else we need to know?**:
I checked logs of deck component and it looks like most of the time we spend in fetching all files from GCE bucket.
```
{
  component: "deck"   
  duration: "25.539879839s"   
  file: "prow/spyglass/gcsartifact_fetcher.go:146"   
  func: "k8s.io/test-infra/prow/spyglass.(*GCSArtifactFetcher).artifacts"   
  level: "info"   
  msg: "Listed 55930 artifacts."   
 }
```

I checked a code and it looks like we fetch 55930 artifact names (which takes 23 seconds) to filter out ~8 files out of them:
https://github.com/kubernetes/test-infra/blob/2bf808d3b5658669f7b0a1bdf93d1c9b73938dd7/config/prow/config.yaml#L36-L60

I would like to improve this to make sig-scalability life easier. Any suggestion how to do that?

mborsz:
I checked and Google Storage has no API that allows to fetch objects in given bucket that match some regexp (kind of expected). 

I was thinking about migrating out from regexp in `LensFileConfig` to something that can be easier handled by GCS API (e.g. prefix + extension filter or some globs?) and then for each lens, do a separate GCS API call that fetches required_files + optional_fields for that len.
 
@Katharine, WDYT?

mborsz:
Alternatively, we can keep regexps, but say that only "filename" part of the object path will be matched, so e.g.:
* artifacts/junit.*.xml will match all junit*.xml files in artifacts/ directory instead of potentially matching all xml files in all junit* directories
* arti.*facts/*.xml will be prohibited
* .*/*.xml will be prohibited as well

[...]

PR #20869, opened 2021-02-16
Implement 'logs' lens

mborsz:
Context: https://github.com/kubernetes/test-infra/issues/20275

We are moving logs to a separate bucket (to unblock pod utils migration and solve deck performance issues caused by listing huge number of objects from GCS: https://github.com/kubernetes/test-infra/issues/17505). To be able to easily find logs, I would like to add logs deck that will show link to logs if 'master-and-node-logs.txt' file is present (it's dumped here: https://github.com/kubernetes/test-infra/blob/master/logexporter/cluster/log-dump.sh#L870).

In the next PR (once this change is released), I will add change to config/prow/config.yaml like:
```
    - lens:
        name: logs
      required_files:
        - ^artifacts/master-and-node-logs\.txt$
```

to enable this code

I have already tested this in my local setup and it seems to work.

/assign @BenTheElder


petr-muller (approved):
Looks simple and LGTM, although I am not that familiar with lenses so it'd be better if someone more familiar would look, too.

I generally like when new stuff is already scaffolded with even trivial tests --  if only so it is not intimidating to add ones when actually needed, but there's very little logic here yet so there's probably really nothing to really test.

mborsz:
pull-test-infra-bazel has failed with:

```
ERROR: An error occurred during the fetch of repository 'git-base':
   Pull command failed: 2021/02/16 18:32:51 Running the Image Puller to pull images from a Docker Registry...
2021/02/16 18:32:52 Image pull was unsuccessful: unable to save remote image gcr.io/k8s-prow/git@sha256:1527341aff1003b6b27c8ed935e6f0200258bee55b6eb178ca3ef124196384fe: unable to write image layers: unable to write image layer: unable to write the contents of layer 2 to /bazel-scratch/.cache/bazel/_bazel_root/05618a594cb3499a4b912817865867cd/external/git-base/image/002.tar.gz: read tcp 10.61.51.56:33448->209.85.200.128:443: read: connection reset by peer
```
/retest

mborsz:
Now pull-test-infra-bazel failed with tslint. Fixed.

mborsz:
I fixed the presubmit. Can we get this submitted today? I would like to migrate more of our jobs to a separate bucket logging but without nice deck support this will be a significant UX issue.

e-blackwelder (approved):
Looks reasonable to me.

PR #20877, opened 2021-02-17
stop using  HugePageStorageMediumSize feature flag as it is on by default

bart0sh:
This is a preparation for graduating HugePageStorageMediumSize feature to GA.


derekwaynecarr:
the feature is beta and on-by default, so this is good.

/approve
/lgtm

PR #20879, opened 2021-02-17
Drop batch/v2alpha1

soltysh:
Follow-up to https://github.com/kubernetes/test-infra/pull/20865 to drop `batch/v2alpha1` also from k8s 1.18, 1.19 and 1.20. None of them should be using this API for long time. 
Ref: https://github.com/kubernetes/kubernetes/pull/96987


saschagrunert (approved):
/lgtm

spiffxp:
I'm almost inclined to revert this branch until we can figure why/if this prevented a PR merging to k/k master.  Something is really wrong if this PR is what unblocked https://github.com/kubernetes/kubernetes/pull/96987

PR #20885, opened 2021-02-17
Log or enforce kubernetes jobs pull from k8s-infra buckets

spiffxp:
To help close out https://github.com/kubernetes/k8s.io/issues/846#issuecomment-765714845

Now that we have community-owned CI artifacts, jobs that use them should
be pulling CI from gs://k8s-release-dev instead of the google.com-owned
gs://kubernetes-release-dev.

Setup to enforce/fail for smaller subsets of jobs, and warn for larger
sets of jobs over time. This is why there's currently unused code to
verify use of gs://k8s-release, and TODOs to enable later.

As a start:

- Fail any kubetest jobs not extracting from gs://k8s-release-dev if:
  - they use --extract=ci/latest-fast
  - they use --extract=ci/* and are release-blocking<sup>1</sup>
- Warn any other kubetest jobs not extracting from gs://k8s-release-dev

<sup>1</sup>also catches --extract=gci/ and --extract=v1.2.3+456 but these aren't in active use here

Issue #19922, opened 2020-11-11
Enforce / monitor / document version marker usage


hasheddan:
**What would you like to be added**:

sig-release maintains [`build`](https://testgrid.k8s.io/sig-release-master-blocking#build-master) and [`fast-build`](https://testgrid.k8s.io/sig-release-master-blocking#build-master-fast) jobs to periodically build and publish kubernetes artifacts. These jobs also update their respective [version markers](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-release/kubernetes-versions.md#ci). When other jobs run, they determine the version of Kubernetes to download by fetching the latest version marker for a particular version. Some of these jobs use the latest "fast" version and some use just the latest. In practice, the only time we need to use a non-fast version is when we are testing an os/arch combo other than `linux/amd64`, but there are a number of jobs that use non-fast version even though it is not required (example: https://github.com/kubernetes/kubernetes/issues/96463#issuecomment-725478355).

**Why is this needed**:

Using a non-fast build is not necessarily wrong, but it does result in a stale build being used when a perfectly valid new one is available. We should always use fast-build when appropriate. However, it is difficult to enforce this as the version marker being used is not always surfaced in the prowjob config. In order to enforce usage of a particular version marker, jobs must utilize frameworks such as [kubetest](https://github.com/kubernetes/test-infra/tree/master/kubetest) where the version is selected in a standardized manner.

Work in this area is related to https://github.com/kubernetes/release/issues/1711 and https://github.com/kubernetes/community/pull/5014.

/cc @justaugustus @saschagrunert @kubernetes/release-engineering 
/assign

justaugustus:
This also has some heavy overlap/interaction with [assessing the version markers in release branch job configs](https://github.com/kubernetes/sig-release/issues/850), so assigning myself as well.

/assign
/remove-kind feature
/kind cleanup
/sig release
/area release-eng

justaugustus:
Also of note: migrating release-blocking jobs to K8s Infra (ref: #19484, #18549).

spiffxp:
/priority important-soon
Version markers are off again

stable1 should be v1.20, since it's the most recent release
```
$ gsutil cat gs://kubernetes-release-dev/ci/k8s-stable1.txt
v1.19.8-rc.0.74+2fffed102ddbb6
```
This means "master/stable1" skew jobs that think they're going most-recent release to master (v1.20/v1.21), are in fact skewing by two versions (v1.19/v1.21)

[...]

PR #20887, opened 2021-02-17
reconcile kubernetes/kubernetes stableN version markers

spiffxp:
Addresses https://github.com/kubernetes/test-infra/issues/19922#issuecomment-780789250

When the `gs://k8s-release-dev/ci/k8s-stableN.txt' version markers were rolled forward, the equivalents in `gs://kubernetes-release-dev` were missed.

The config for auto-generated release jobs in `releng/` was also missed, so some `stable1` jobs were pointed at `v1.19` (two releases ago instead of 1)

No longer WIP now that:
- ~I haven't verified whether the stableN job names match up properly, and confirmed everything on a given release-1.y-blocking testgrid dashboard actually uses release-1.y~ lgtm
- ~This contains commits from https://github.com/kubernetes/test-infra/pull/20885, I'll rebase once that merges~ rebased

What I consider out of scope for this PR, but am open to for followup:
- replace 'stableN' with '1-y' in job names (we lose job history in e.g. testgrid for a bit, but we're early enough in release cycle to make this change IMO)
- replace uses of '--extract=ci/k8s-stableN' with more concrete version markers

I'm less certain we can drop populating the `ci/k8s-stableN` version markers without a deprecation window


spiffxp:
Needed to rebase due to https://github.com/kubernetes/test-infra/pull/20880 merging

cpanato (approved):
/lgtm

PR #20892, opened 2021-02-18
[tide] Ensure that tide github context is success before attempting to merge.

arturo-skydio:
This ensure the statusContext "tide" is set to success before attempting to merge.

It allows users to rely on the "tide" context branch protection without having to set all the other as required contexts.

Another behavior I believe this fixes is where a presubmit job gets triggered for a PR that is about to be merged.

This is a very needed improvement if you're also using this other setting still not merge in PROW #20679 . Restarting builds for the PRs that didn't get merged in time and get picked again for a batch.

### DIFF
--- a/prow/tide/tide.go
+++ b/prow/tide/tide.go
@@ -1136,11 +1136,22 @@ func (c *Controller) mergePRs(sp subpool, prs []PullRequest) error {
 	var errs []error
 	log := sp.log.WithField("merge-targets", prNumbers(prs))
 	tideConfig := c.config().Tide
+
 	for i, pr := range prs {
 		log := log.WithFields(pr.logFields())
 		mergeMethod, err := prMergeMethod(tideConfig, &pr)
 		if err != nil {
 			log.WithError(err).Error("Failed to determine merge method.")
+			errs = append(errs, err)
+			failed = append(failed, int(pr.Number))
+			continue
+		}
+
+		// Ensure tide context has success state, otherwise PR merge will fail if branch protection
+		// in github is enabled and the loop to change tide context hasn't done it already
+		err = setTideStatusSuccess(pr, c.ghc, c.config(), log)
+		if err != nil {
+			log.WithError(err).Error("Unable to set tide context to SUCCESS.")
 			errs = append(errs, err)
 			failed = append(failed, int(pr.Number))
 			continue
@@ -1181,6 +1192,23 @@ func (c *Controller) mergePRs(sp subpool, prs []PullRequest) error {
 		}
 	}
 	return fmt.Errorf("failed merging %v%s: %v", failed, batch, utilerrors.NewAggregate(errs))
+}
+
+// setTideStatusSuccess calls github api to change tide status context to success
+func setTideStatusSuccess(pr PullRequest, ghc githubClient, cfg *config.Config, log *logrus.Entry) error {
+	if err := ghc.CreateStatus(
+		string(pr.Repository.Owner.Login),
+		string(pr.Repository.Name),
+		string(pr.HeadRefOID),
+		github.Status{
+			Context:   statusContext,
+			State:     "success",
+			TargetURL: targetURL(cfg, &pr, log),
+		}); err != nil {
+		return err
+	}
+
+	return nil
 }
 
 // tryMerge attempts 1 merge and returns a bool indicating if we should try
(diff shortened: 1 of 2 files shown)


prow/tide/tide.go
```diff
@@ -1183,6 +1195,37 @@ func (c *Controller) mergePRs(sp subpool, prs []PullRequest) error {
...
+		return err
+	}
+	for _, ctx := range contexts {
+		if string(ctx.Context) == "tide" {
+			actualState = ctx.State
+		}
+	}
+
+	if string(actualState) != "success" {
+		if err := c.ghc.CreateStatus(
```
alvaroaleman, line 1213: I think it would preferable to just do this unconditionally, because:
* Uses always one token vs uses always one and up to two tokens
* Quicker, because we save an api call
* Simpler code

prow/tide/tide.go
```diff
@@ -1146,6 +1147,17 @@ func (c *Controller) mergePRs(sp subpool, prs []PullRequest) error {
 			continue
 		}
 
+		// Ensure tide context if enabled
+		if tideConfig.TideGithubContext != nil && *tideConfig.TideGithubContext {
+			err = c.ensureTideContext(pr, log)
```
alvaroaleman, line 1152: Please add tests for this

prow/config/prow-config-documented.yaml
```diff
@@ -861,3 +861,8 @@ tide:
     # We can consider allowing this to be set separately for separate repos, or
     # allowing it to be a template.
     target_url: ' '
+
+    # TideGithubContext ensures that before attempting to merge PRs tide updates the github
+    # context to "success", this setting is useful to turn on if you use github feature to
+    # protect branch on tide context, which can then have failures on tide when attempting merge
+    tide_github_context: false
```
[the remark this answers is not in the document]
arturo-skydio: Gotcha, so just to confirm we probably don't need to care about having an option but always set the context to success?, Also not sure if my assessment that there will always be the "tide" context if you have your repo setup with prow hooks is correct

prow/tide/tide.go
arturo-skydio: makes sense, will remove the check for current state

prow/config/prow-config-documented.yaml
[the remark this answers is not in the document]
chaodaiG: > Cc @chaodaiG why does prow.k8s.io not track the v4 api tokens?

I'm not quite sure, @cjwagner might be able to answer this question better

prow/tide/tide.go
arturo-skydio: Added some, let me know if it makes sense, followed examples from other tests on using the fgc

prow/tide/tide.go
```diff
@@ -1146,6 +1147,16 @@ func (c *Controller) mergePRs(sp subpool, prs []PullRequest) error {
 			continue
 		}
 
+		// Ensure tide context has success state, otherwise PR merge will fail if branch protection
+		// in github is enabled and the loop to change tide context hasn't done it already
+		err = setTideStatusSuccess(pr, c.ghc, c.config(), log)
```
cjwagner, line 1152: Without synchronization with the status controller there will be a race here. The status controller could set the status back to pending before or after we try to merge.

prow/config/prow-config-documented.yaml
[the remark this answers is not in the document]
cjwagner: I'm not sure, I thought we did based on the dashboard configuration, but it looks like we might not be routing v4 traffic through the proxy. I don't see any use of this flag: https://github.com/kubernetes/test-infra/blob/b769e42a6f5cdef9c30bce331f62080926fdb9af/prow/flagutil/github.go#L61

prow/tide/tide.go
arturo-skydio: Hi thanks for reviewing, were you thinking maybe set another gh context 'tide-merging' and have the loop ignore those PRs? Or how is this 'state' regularly handled in prow

cjwagner: Adding another GH context would not be appropriate for this. We only need once context, we just need the status and sync controllers to be in agreement.
The state of the status context is currently entirely managed by the tide status controller which is a separate set of goroutines from the tide sync controller that handles retesting and merging. If the sync controller needs a status to be set to green immediately before merge, it needs to synchronize with the status controller to either 1) have the status controller set the context on its behalf or 2) ensure the status controller is not and will not interfere with the status on that PR so that the sync controller can set the status to green itself.

PR #20897, opened 2021-02-18
Rename 'logs' lens to 'links'

mborsz:
In https://github.com/kubernetes/test-infra/pull/20869 I added a 'logs' lens.

After merging that PR, @e-blackwelder pinged me on chat and asked for the following changes:
* using `html/template` instead of `text/template`
* Renaming 'logs', to avoid confusion with 'buildlog'
* Introducing some limit for files we read

I propose renaming this to 'link' and in fact implement slightly more generic lens that allows to display any links. The expected usage is:
```
    - lens:
        name: links
      required_files:
        - ^artifacts/.*\.link\.txt$
        - ^artifacts/master-and-node-logs\.txt$ # Temporary, until we migrate all log-dump to use 'master-and-node-logs.link.txt'

```

This way, we can also generate files like e.g. artifacts/some-fancy-debugging-dashboard.link.txt or artifacts/grafana.link.txt and we will display that link to e.g. metrics for that run.

/assign @e-blackwelder


prow/spyglass/lenses/links/links.go
```diff
@@ -74,21 +77,44 @@ func renderTemplate(resourceDir, block string, params interface{}) (string, erro
...
+		if err != nil {
+			logrus.WithError(err).Warnf("Failed to read artifact file: %q", artifact.JobPath())
+		}
+		links = append(links, link{
+			Name: humanReadableName(filepath.Base(artifact.JobPath())),
+			URL:  strings.TrimSpace(string(content)),
+		})
 	}
 
+	sort.Slice(links, func(i, j int) bool { return links[i].Name < links[i].Name })
```
cjwagner, line 112:
```suggestion
	sort.Slice(links, func(i, j int) bool { return links[i].Name < links[j].Name })
```

cjwagner (commented):
Thanks for the improvements! I spotted one issue with the sorting, but otherwise looks good to me.

prow/spyglass/lenses/links/links.go
mborsz: Thanks! :) Fixed

cjwagner (approved):

BenTheElder:
This seems handy 😄  👍

PR #20899, opened 2021-02-18
Stop suggesting already assigned approvers

matthyx:
Relates to #16793
This is a complete departure from the existing logic, it probably needs confirmation from @cjwagner before merging...
/cc @BenTheElder 
/cc @cjwagner

### DIFF
--- a/prow/plugins/approve/approve_test.go
+++ b/prow/plugins/approve/approve_test.go
@@ -277,8 +277,7 @@ Approvers can cancel approval by writing ` + "`/approve cancel`" + ` in a commen
 			expectedComment: `[APPROVALNOTIFIER] This PR is **NOT APPROVED**
 
 This pull-request has been approved by:
-To complete the [pull request process](https://git.k8s.io/community/contributors/guide/owners.md#the-code-review-process), please assign **cjwagner** after the PR has been reviewed.
-You can assign the PR to them by writing ` + "`/assign @cjwagner`" + ` in a comment when ready.
+To complete the [pull request process](https://git.k8s.io/community/contributors/guide/owners.md#the-code-review-process), please ask for approval from **cjwagner** after the PR has been reviewed.
 
 The full list of commands accepted by this bot can be found [here](https://go.k8s.io/bot-commands?repo=org%2Frepo).
 
@@ -723,8 +722,7 @@ Approvers can cancel approval by writing ` + "`/approve cancel`" + ` in a commen
 			expectedComment: `[APPROVALNOTIFIER] This PR is **NOT APPROVED**
 
 This pull-request has been approved by:
-To complete the [pull request process](https://git.k8s.io/community/contributors/guide/owners.md#the-code-review-process), please assign **cjwagner** after the PR has been reviewed.
-You can assign the PR to them by writing ` + "`/assign @cjwagner`" + ` in a comment when ready.
+To complete the [pull request process](https://git.k8s.io/community/contributors/guide/owners.md#the-code-review-process), please ask for approval from **cjwagner** after the PR has been reviewed.
 
 The full list of commands accepted by this bot can be found [here](https://go.k8s.io/bot-commands?repo=org%2Frepo).
 
@@ -793,8 +791,7 @@ Approvers can cancel approval by writing ` + "`/approve cancel`" + ` in a commen
 			expectedComment: `[APPROVALNOTIFIER] This PR is **NOT APPROVED**
 
 This pull-request has been approved by:
-To complete the [pull request process](https://git.k8s.io/community/contributors/guide/owners.md#the-code-review-process), please assign **cjwagner** after the PR has been reviewed.
-You can assign the PR to them by writing ` + "`/assign @cjwagner`" + ` in a comment when ready.
+To complete the [pull request process](https://git.k8s.io/community/contributors/guide/owners.md#the-code-review-process), please ask for approval from **cjwagner** after the PR has been reviewed.
 
 The full list of commands accepted by this bot can be found [here](https://go.k8s.io/bot-commands?repo=org%2Frepo).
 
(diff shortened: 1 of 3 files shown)


cjwagner:
/hold
I don't think this is what we want. This is the concern:
> Seems a bit silly to suggest assigning someone who is already assigned, if nothing else? Perhaps we can just not suggest then?

I think the core of the concern is that it is confusing to be asked to assign someone who is already assigned, not that we should avoid suggesting getting approval from assignees.
If a user with sufficient approval permissions is already assigned to a PR I think we want to continue suggesting getting approval from them rather than suggesting another person who isn't assigned to the PR. BUT we should recognize when the user is already assigned and suggest prompting them for approval rather than `/assign`ing them again since that does nothing and is confusing/silly.

matthyx:
> we should recognize when the user is already assigned and suggest prompting them for approval rather than `/assign`ing them again since that does nothing and is confusing/silly.

ok, so I keep the current behavior of `GetCCs` but in the comment template itself I treat this information differently if the suggested person is already assigned

matthyx:
Like that @cjwagner ?

prow/plugins/approve/approve_test.go
```diff
@@ -942,8 +937,7 @@ Approvers can cancel approval by writing ` + "`/approve cancel`" + ` in a commen
 			expectedComment: `[APPROVALNOTIFIER] This PR is **NOT APPROVED**
 
 This pull-request has been approved by:
-To complete the [pull request process](https://git.k8s.io/community/contributors/guide/owners.md#the-code-review-process), please assign **cjwagner** after the PR has been reviewed.
-You can assign the PR to them by writing ` + "`/assign @cjwagner`" + ` in a comment when ready.
+To complete the [pull request process](https://git.k8s.io/community/contributors/guide/owners.md#the-code-review-process), please ask an approval from **cjwagner** after the PR has been reviewed.
```
chaodaiG, line 940: Is there a test case that cjwagner is not assigned on this PR yet, and the comment still suggests `/assign`?

prow/plugins/approve/approvers/owners.go
```diff
@@ -556,6 +556,14 @@ func (ap Approvers) ListNoIssueApprovals() []Approval {
 	return approvals
 }
 
+func (ap Approvers) AssignedCCs() []string {
```
chaodaiG, line 559: Should we have unit test for these 2 new functions?

chaodaiG (commented):
Overall looks good to me, feel like there are some unit tests missing

prow/plugins/approve/approvers/owners.go
```diff
@@ -633,8 +641,13 @@ Approval requirements bypassed by manually added approval.
 This pull-request has been approved by:{{range $index, $approval := .ap.ListApprovals}}{{if $index}}, {{else}} {{end}}{{$approval}}{{end}}
 
 {{- if (and (not .ap.AreFilesApproved) (not (call .ap.ManuallyApproved))) }}
-To complete the [pull request process]({{ .prProcessLink }}), please assign {{range $index, $cc := .ap.GetCCs}}{{if $index}}, {{end}}**{{$cc}}**{{end}} after the PR has been reviewed.
-You can assign the PR to them by writing `+"`/assign {{range $index, $cc := .ap.GetCCs}}{{if $index}} {{end}}@{{$cc}}{{end}}`"+` in a comment when ready.
+{{ if len .ap.SuggestedCCs -}}
+To complete the [pull request process]({{ .prProcessLink }}), please assign {{range $index, $cc := .ap.SuggestedCCs}}{{if $index}}, {{end}}**{{$cc}}**{{end}} after the PR has been reviewed.
+You can assign the PR to them by writing `+"`/assign {{range $index, $cc := .ap.SuggestedCCs}}{{if $index}} {{end}}@{{$cc}}{{end}}`"+` in a comment when ready.
+{{- end}}
+{{- if len .ap.AssignedCCs -}}
```
cjwagner, line 648: If `GetCCs()` returns multiple users and some are assignees while others are not, we'll output both of these sections which seems wrong to me. Please add a unit test for this case too.

matthyx: It was on purpose... What do you suggest then?

cjwagner: Correction: I meant confusing not wrong. I think we **do** want to recommend both assigning the SuggestedCCs and requesting approval from the AssignedCCs, but I think it is confusing to have two lines with different instructions for "To complete the pull request process...".

To me it would be clearer to have one set of instructions when we have both suggested and assigned CCs. Something like this:
> To complete the pull request process, please ask for approval from cjwagner and additionally assign mattyx after the PR has been reviewed.
You can assign the PR to them by writing `/assign matthyx` in a comment when ready

matthyx: done this part for review, although I need to add some test cases suggested by @chaodaiG

prow/plugins/approve/approve_test.go
matthyx: like prow/plugins/approve/approvers/approvers_test.go:866 ?

chaodaiG: Yeah that would work. More specifically:
-Without- the line containing `/approve cancel`
-With- the line containing `"someone else", "github.ReviewStateApproved` or `"someone else", "/approve"

Also you can test when there is only self-approval

prow/plugins/approve/approvers/owners.go
MushuEE: nit: quick doc string for public functions

matthyx:
Sorry for the delay, PTAL!

prow/plugins/approve/approve_test.go
```diff
@@ -1089,6 +1085,39 @@ Approvers can cancel approval by writing ` + "`/approve cancel`" + ` in a commen
...
+			reviewActsAsApprove: true,
+			githubLinkURL:       &url.URL{Scheme: "https", Host: "github.com"},
+
+			expectDelete:  false,
+			expectToggle:  false,
+			expectComment: true,
+			expectedComment: `[APPROVALNOTIFIER] This PR is **NOT APPROVED**
+
+This pull-request has been approved by: *<a href="" title="Approved">Alice</a>*
+To complete the [pull request process](https://git.k8s.io/community/contributors/guide/owners.md#the-code-review-process), please ask for approval from **cjwagner** after the PR has been reviewed.
```
chaodaiG, line 1106: it took me a while to figure out the logic of this unit test:
- cjwagner is the author of all test cases in this test
- this test case has `selfApprove: false`, so cjwagner didn't approve their own test
- then the PR suggests ask for approval from cjwagner

Logically speaking this is all correct, and this is perhaps the best we can do for this scenario in prod. Might not be in this PR, but maybe we should make the PR author configurable to cover more scenarios?

chaodaiG (approved):
LGTM, left a comment

/hold
In case you want @cjwagner to take another look

cjwagner (approved):

PR #20902, opened 2021-02-18
configurator: prepopulate fields depending on Prow instance

### LINKED ISSUE #20859: Configurator should detect the Prow instance being used and populate `open_test_template`
Prow configuration knows about where its hosted (or more specifically, what prefix it should add to access the job viewer) from `job_url_prefix_config`: https://github.com/kubernetes/test-infra/blob/80cdac19030375fd7ee17fca11ddd5a0ec772cd0/config/prow/config.yaml#L6.

The key is either "*", an org, or an org/repo, with the most specific being used.

TestGrid's configuration should populate at least the Dashboard Tab Option [`open_test_template`](https://github.com/GoogleCloudPlatform/testgrid/blob/2ece0ecafcd93a6454d535055c9f22b478b6da65/pb/config/config.proto#L522) (and potentially `results_url_template`) using this information.

Currently, these are being populated for all users using a default setting in TestGrid: https://github.com/kubernetes/test-infra/blob/d34feb5af1fd5a07bd963ea04637f7039b3b00fb/config/testgrids/default.yaml#L22-L25

**Why is this needed**:
There are some use cases where TestGrid configuration is being imported from a different Prow instance (see [Transfigure](https://github.com/kubernetes/test-infra/tree/master/testgrid/cmd/transfigure)), requiring users to set their own local defaults to display correctly.

Example: https://github.com/kubernetes/test-infra/blob/master/config/testgrids/kubevirt/default.yaml

Unfortunately (and unrelated), nested defaults don't actually work.  https://github.com/kubernetes/test-infra/issues/20756
This enhancement would reduce our reliance on nested defaults and would make Prow and TestGrid operate more independently of their specific instance.

/help-wanted

fgimenez:
Fixes #20859 

Changes to populate `OpenTestTemplate` in dashboard tab configs based on `JobBase`'s `JobURLPrefix`. It reuses the logic in `Plank.GetJobURLPrefix`, so some of the tests might be redundant. Also, in order to be able to use that logic a Periodic job CR is created (simplest from what I've seen), let me know if there's a better way.

/cc @chases2 @spiffxp 


chases2 (approved):
This looks good to me! Feel free to cancel the hold when you're ready.
/hold

Thanks!

testgrid/cmd/configurator/prow_test.go
```diff
@@ -632,6 +634,216 @@ func Test_applySingleProwjobAnnotation_WithDefaults(t *testing.T) {
 
 }
 
+func Test_applySingleProwjobAnnotations_AutomaticFieldPopulation(t *testing.T) {
+	tests := []*struct {
+		name                 string
+		prowConfig           *prowConfig.Config
```
spiffxp, line 640:
```suggestion
		jobURLPrefixConfig           map[string]string
```
You're not changing anything else about `prowConfig`, so build it in your test harness, e.g.
```golang
prowConfig := fakeProwConfig()
prowConfig.Plank.JobURLPrefixConfig = test.jobURLPrefixConfig
```

testgrid/cmd/configurator/prow_test.go
```diff
@@ -632,6 +634,216 @@ func Test_applySingleProwjobAnnotation_WithDefaults(t *testing.T) {
...
+		expectedDashboardTab config.DashboardTab
+	}{
+		{
+			name: "job url prefix without specific sufix",
```
spiffxp, line 644:
```suggestion
			name: "job url prefix without specific suffix",
```

testgrid/cmd/configurator/prow_test.go
```diff
@@ -632,6 +634,216 @@ func Test_applySingleProwjobAnnotation_WithDefaults(t *testing.T) {
 
 }
 
+func Test_applySingleProwjobAnnotations_AutomaticFieldPopulation(t *testing.T) {
```
spiffxp, line 637: Seems like you're only interested in verifying  `OpenTestTemplate` here, I would consider modifying your test case to only check that field, or folding into test cases above that are checking all fields
```suggestion
func Test_applySingleProwjobAnnotations_OpenTestTemplate(t *testing.T) {
```

spiffxp (changes requested):
I'd like to see a few changes, suggestions below

testgrid/cmd/configurator/prow_test.go
spiffxp: blocking nit: I would like to see a test case confirming this overwrites test grid's default OpenTestTemplate

testgrid/cmd/configurator/prow_test.go
```diff
@@ -632,6 +634,216 @@ func Test_applySingleProwjobAnnotation_WithDefaults(t *testing.T) {
...
+				Description:   ProwJobDefaultDescription,
+				TestGroupName: ProwJobName,
+				CodeSearchUrlTemplate: &config.LinkTemplate{
+					Url: "https://github.com/test/repo/compare/<start-custom-0>...<end-custom-0>",
+				},
+				OpenBugTemplate: &config.LinkTemplate{
+					Url: "https://github.com/test/repo/issues/",
+				},
+				OpenTestTemplate: &config.LinkTemplate{
+					Url: path.Join(ProwJobURLPrefixConfig, "view", "<gcs_prefix>", "<changelist>"),
```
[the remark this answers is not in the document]
fgimenez: Awesome thx, this uncovered a problem mixing path.Join and URLs, fixed.

testgrid/cmd/configurator/prow_test.go
fgimenez: > Non-blocking nit: I would consider modifying your test case to only check that field, or folding into test cases above that are checking all fields

makes total sense, I've gone for the second option.

fgimenez: > blocking nit: I would like to see a test case confirming this overwrites test grid's default OpenTestTemplate

Done, added as part of Test_applySingleProwjobAnnotations_OpenTestTemplate

testgrid/cmd/configurator/prow.go
```diff
@@ -191,12 +192,32 @@ func (pac *prowAwareConfigurator) applySingleProwjobAnnotations(c *configpb.Conf
...
+			// create simple ProwJob from JobBase including repo info to get the job url prefix
+			pjSpec := pjutil.PeriodicSpec(prowConfig.Periodic{JobBase: j})
+			if repo != "" {
+				items := strings.Split(repo, "/")
+				pjSpec.ExtraRefs = []prowapi.Refs{{Org: items[0]}}
+				if len(items) > 1 {
+					pjSpec.ExtraRefs[0].Repo = items[1]
+				}
+			}
+			pj := pjutil.NewProwJob(pjSpec, nil, nil)
```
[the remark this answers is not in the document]
fgimenez: Absolutely, thanks a lot for the suggestion, much nicer now :heart: Already pushed, PTAL

chases2 (commented):
Looks good! Please merge your commits together for this change.
/lint

spiffxp (approved):
/approve
/lgtm
/hold cancel
I'll let tide do the squashing

spiffxp:
>I'll let tide do the squashing

ok, fair enough, looking at the merge commit description https://github.com/kubernetes/test-infra/commit/a0aaab97de8ca2b6acd0edfa3fd623e4331a142d I'll opt to ask for an author rebase next time

spiffxp:
Think this caused https://github.com/kubernetes/test-infra/issues/20988, followup PR to fix: https://github.com/kubernetes/test-infra/pull/20987

PR #20903, opened 2021-02-18
Prow dashboard: rename boskos job name to avoid wrong matching

chaodaiG:
Before https://github.com/kubernetes/test-infra/pull/20888, both k8s-prow-builds-boskos and k8s-infra-prow-builds-boskos use job name of k8s-prow-builds-boskos, so when #20888 changes the matching to , this condition  matches  from both clusters. Renaming it so that the history data is also accurate


spiffxp (approved):
/approve
/lgtm

PR #20908, opened 2021-02-19
Support list file commits

hi-rustin:
We need to get the history of contributors for each file when implementing an external plugin. So add this API to get the commits against the file.

GitHub documentation: https://docs.github.com/en/rest/reference/repos#commits

Example: https://api.github.com/repos/ti-community-infra/tichi/commits?path=README.md

The result is an array of RepositoryCommits.

### DIFF
--- a/prow/github/client.go
+++ b/prow/github/client.go
@@ -145,6 +145,7 @@ type CommitClient interface {
 	ListCheckRuns(org, repo, ref string) (*CheckRunList, error)
 	GetRef(org, repo, ref string) (string, error)
 	DeleteRef(org, repo, ref string) error
+	ListFileCommits(org, repo, path string) ([]RepositoryCommit, error)
 }
 
 // RepositoryClient interface for repository related API actions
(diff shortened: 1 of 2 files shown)


prow/github/client.go
```diff
@@ -2969,6 +2970,35 @@ func (c *client) DeleteRef(org, repo, ref string) error {
 	return err
 }
 
+// ListFileCommits returns the commits for this file path.
+//
+// See https://developer.github.com/v3/repos/#list-commits
+func (c *client) ListFileCommits(org, repo, filePath string) ([]RepositoryCommit, error) {
```
MushuEE, line 2976: Would be nice to throw in a quick test for this, the GetRef test is probably a nice framework for it

hi-rustin: Added.

hi-rustin:
@alvaroaleman Could you please take a look?

alvaroaleman (approved):

PR #20909, opened 2021-02-19
user apiserver /healthz instead of cs check

pacoxu:
As componentstatus as deprecated in 1.19+, https://github.com/kubernetes/kubernetes/pull/93570 conformance test should not use it at all.

### DIFF
--- a/kubetest/conformance/BUILD.bazel
+++ b/kubetest/conformance/BUILD.bazel
@@ -8,8 +8,6 @@ go_library(
     deps = [
         "//kubetest/e2e:go_default_library",
         "//kubetest/process:go_default_library",
-        "@io_k8s_api//core/v1:go_default_library",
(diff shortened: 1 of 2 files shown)


amwat:
Thanks for the change!

I think this code path is no longer used anywhere in CI (not sure if folks are using it locally) since it's under 
`--deployment=conformance`

https://cs.k8s.io/?q=--deployment%3Dconformance&i=nope&files=&excludeFiles=&repos=

Note that kubetest is deprecated in general, so we are only accepting critical bug fixes which affect existing CI jobs
https://github.com/kubernetes/test-infra/tree/master/kubetest#deprecation-notice

Not to mention we currently still officially support testing for 1.18

pacoxu:
@amwat thanks for the information. I see the message in conformance testing log, if that is deprecated, no need to change this.

Let me close it. 
/close

PR #20912, opened 2021-02-19
Make links clickable

mborsz:
Let's go one step further. Let's make our links clickable!

/assign @e-blackwelder
/assign @cjwagner

### DIFF
--- a/prow/spyglass/lenses/links/links.go
+++ b/prow/spyglass/lenses/links/links.go
@@ -21,6 +21,8 @@ import (
 	"encoding/json"
 	"fmt"
 	"html/template"
+	"io"
+	"net/url"
 	"path/filepath"
 	"sort"
 	"strings"
@@ -91,23 +93,50 @@ func humanReadableName(name string) string {
 	return strings.Join(words, " ")
 }
 
+func clickableLink(link string, spyglassConfig config.Spyglass) string {
+	if strings.HasPrefix(link, "gs://") {
+		return spyglassConfig.GCSBrowserPrefix + strings.TrimPrefix(link, "gs://")
+	}
+	if strings.HasPrefix(link, "http://") || strings.HasPrefix(link, "https://") {
+		// Use https://google.com/url?q=[url] redirect to avoid leaking referer.
+		url := url.URL{
+			Scheme: "https",
+			Host:   "google.com",
+			Path:   "/url",
+		}
+		q := url.Query()
+		q.Add("q", link)
+		url.RawQuery = q.Encode()
+		return url.String()
+	}
+	return ""
+}
+
+type link struct {
+	Name string
+	URL  string
+	Link string
+}
+
+func toLink(jobPath string, content []byte, spyglassConfig config.Spyglass) link {
+	url := strings.TrimSpace(string(content))
+	return link{
+		Name: humanReadableName(filepath.Base(jobPath)),
+		URL:  url,
+		Link: clickableLink(url, spyglassConfig),
+	}
+}
+
 // Body renders link to logs.
 func (lens Lens) Body(artifacts []api.Artifact, resourceDir string, data string, config json.RawMessage, spyglassConfig config.Spyglass) string {
-	type link struct {
-		Name string
-		URL  string
-	}
-
 	var links []link
 	for _, artifact := range artifacts {
 		content, err := artifact.ReadAtMost(bytesLimit)
-		if err != nil {
+		if err != nil && err != io.EOF {
 			logrus.WithError(err).Warnf("Failed to read artifact file: %q", artifact.JobPath())
+			continue
 		}
-		links = append(links, link{
-			Name: humanReadableName(filepath.Base(artifact.JobPath())),
-			URL:  strings.TrimSpace(string(content)),
-		})
+		links = append(links, toLink(artifact.JobPath(), content, spyglassConfig))
 	}
 
 	sort.Slice(links, func(i, j int) bool { return links[i].Name < links[j].Name })
--- a/prow/spyglass/lenses/links/links_test.go
+++ b/prow/spyglass/lenses/links/links_test.go
@@ -16,7 +16,11 @@ limitations under the License.
 
 package links
 
-import "testing"
+import (
+	"testing"
+
+	"k8s.io/test-infra/prow/config"
+)
 
 func TestHumanReadableName(t *testing.T) {
 	tests := []struct {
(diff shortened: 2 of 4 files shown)


wojtek-t:
/lgtm
/approve

thanks!

prow/spyglass/lenses/links/links.go
```diff
@@ -36,6 +37,8 @@ const (
 	priority = 20
 
 	bytesLimit = 10 * 1024
+
+	defaultGCSBrowserPrefix = "https://gcsweb.k8s.io/gcs/"
```
alvaroaleman, line 41: There is config.spyglass.gcs_browser_prefix, please use that instead of  a hardcoded value

prow/spyglass/lenses/links/links.go
```diff
@@ -90,22 +93,56 @@ func humanReadableName(name string) string {
...
+		if err != nil && err != io.EOF {
 			logrus.WithError(err).Warnf("Failed to read artifact file: %q", artifact.JobPath())
+			continue
 		}
+		url := strings.TrimSpace(string(content))
 		links = append(links, link{
 			Name: humanReadableName(filepath.Base(artifact.JobPath())),
-			URL:  strings.TrimSpace(string(content)),
+			URL:  url,
+			Link: clickableLink(url, c),
```
alvaroaleman, line 145: tests?

alvaroaleman (changes requested):

prow/spyglass/lenses/links/links.go
mborsz: I wanted to use that value initially, but it looks like lens doesn't have access to spyglass config (only to lens config)

alvaroaleman: So lets plug it through?

prow/spyglass/lenses/links/links.go
```diff
@@ -90,22 +93,56 @@ func humanReadableName(name string) string {
...
+}
+
+func parseConfig(raw json.RawMessage) (*Config, error) {
+	c := &Config{
+		GCSBrowserPrefix: defaultGCSBrowserPrefix,
+	}
+	if len(raw) == 0 {
+		return c, nil
+	}
+	return c, json.Unmarshal(raw, c)
```
e-blackwelder, line 108: Agree that tests would be nice... I'm _guessing_ that an empty JSON object (something like `{ }`) or one with an empty-string `gcs_browser_prefix` value would break the clicableLink-generation.

prow/spyglass/lenses/links/links.go
mborsz: I opened https://github.com/kubernetes/test-infra/pull/21373. Please review this change :)

prow/spyglass/lenses/links/links.go
```diff
@@ -90,22 +93,56 @@ func humanReadableName(name string) string {
...
+}
+
+func clickableLink(url string, c *Config) string {
```
fejta, line 111: Log files are generated by arbitrary untrusted code. Thus the URI can be anything, potentially something malicious. 

Can we do a better job protecting the user from harm here?

1) Make sure the link is sanitized and less likely to be malicious
2) Redirect to something first, so that we do not leak the person's internal URLs to the destination site.
  - AKA a link to google.com in a prow.k8s.io/view/gcs/whatever log will first route through prow.k8s.io/redirect?dest=google.com
  - Otherwise we will be sending view/gcs/whatever to google.com
3) Otherwise (or maybe in addition) putting this functionality behind a config variable that defaults to off

fejta (changes requested):

prow/spyglass/lenses/links/links.go
mborsz: Done

prow/spyglass/lenses/links/links.go
mborsz: Added

prow/spyglass/lenses/links/links.go
mborsz: Removed this section competely

prow/spyglass/lenses/links/links.go
mborsz: I added redirect service. The lens is already optional and requires enabling it in prow's config.

mborsz:
I addressed all the comments.  PTAL

prow/spyglass/lenses/links/links_test.go
```diff
@@ -55,3 +59,40 @@ func TestHumanReadableName(t *testing.T) {
...
+			content: []byte("   http://website.com/asdasds?adssasds   "),
+			want: link{
+				Name: "Dashboard",
+				URL:  "http://website.com/asdasds?adssasds",
+				Link: "https://google.com/url?q=http%3A%2F%2Fwebsite.com%2Fasdasds%3Fadssasds",
+			},
+		},
+		{
+			jobPath: "artifacts/master-and-node-logs.txt",
+			content: []byte("gs://bucket/asdasd/asdasd\n"),
```
fejta, line 83: Wouldn't more likely content be something like:

```
foo.go:38 did something
foo.go:100 did something else see http://www.google.com for more info
foo.go:38 did something
```

prow/spyglass/lenses/links/links.go
```diff
@@ -91,23 +93,50 @@ func humanReadableName(name string) string {
...
+		URL:  url,
+		Link: clickableLink(url, spyglassConfig),
+	}
+}
+
+// Body renders link to logs.
+func (lens Lens) Body(artifacts []api.Artifact, resourceDir string, data string, config json.RawMessage, spyglassConfig config.Spyglass) string {
 	var links []link
 	for _, artifact := range artifacts {
 		content, err := artifact.ReadAtMost(bytesLimit)
```
fejta, line 134: It seems weird to me that we're reading the first bytesLimit of every artifact and then encoding it as URL without any other validation/extraction. Is that really the intended behavior?

fejta (commented):
Looks good but I feel like I'm missing something about what artifacts are sent to body.

prow/spyglass/lenses/links/links_test.go
mborsz: The expected content is either gs:// path or link. Example here: https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/ci-kubernetes-e2e-gce-scale-performance/1401947321612111872

prow/spyglass/lenses/links/links.go
mborsz: Yes, the artifact body is expected to be a single line url or gs:// path. Only some artifacts (the ones with 'link.txt' suffix) will reach this area.

Example test using this feature: https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/ci-kubernetes-e2e-gce-scale-performance/1401947321612111872

The "debugging links" section shows an example "master-and-node-logs.link.txt" file.

fejta (approved):

PR #20915, opened 2021-02-19
Ignore DUMP_TO_GCS_ONLY if ${gcs_artifacts_dir} is not set.

mborsz:
If ${gcs_artifacts_dir} is not set, we have no bucket to push to.

Without that, if we add DUMP_TO_GCS_ONLY to preset, jobs not using exporter will fail.

/assign @wojtek-t 
/hold (I'm testing this fix right now)


mborsz:
/hold cancel

Tested. It works.

PR #20917, opened 2021-02-19
Enable override plugin in kubernetes-sigs/kubebuilder

Adirio:
According to:

https://github.com/kubernetes/test-infra/blob/b023bd33f7999be8fefe25900a59b44defe3271f/config/prow/plugins.yaml#L643-L645

This is how `override` is enabled for https://github.com/kubernetes-sigs/kubebuilder.

As an example,  https://github.com/kubernetes-sigs/kind is doing the same:

https://github.com/kubernetes/test-infra/blob/b023bd33f7999be8fefe25900a59b44defe3271f/config/prow/plugins.yaml#L912-L914

There doesn't seem to be any order, so I placed it the last of `kubernetes-sigs/*` repos.

/cc @DirectXMan12

### DIFF
--- a/config/prow/plugins.yaml
+++ b/config/prow/plugins.yaml
@@ -990,6 +990,10 @@ plugins:
     - override
     - release-note
 
+  kubernetes-sigs/kubebuilder:
+    plugins:
+    - override
+
   containerd/cri:
     plugins:
     - assign


camilamacedo86 (approved):
/lgtm
Great

Adirio:
Friendly ping @cblecker

Adirio:
Tests keep failing but the PR seems to follow the exact same behavior as other projects/repos.

PR #20920, opened 2021-02-19
Cherrypicker: Allow pushing to a differently-named fork

alvaroaleman:
Currently, if the fork is named differently we use that forks name together with the base organization to pull, which is wrong. This changes it to use the original name for pulling and the forks name for pushing.

### DIFF
--- a/prow/external-plugins/cherrypicker/server.go
+++ b/prow/external-plugins/cherrypicker/server.go
@@ -88,7 +88,7 @@ type Server struct {
 
 	gc git.ClientFactory
 	// Used for unit testing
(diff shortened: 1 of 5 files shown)


michelle192837 (approved):
/hold

Looks good to me, but holding in case you want someone more familiar with Prow to double-check. Unhold at will if not!

alvaroaleman:
The PR body correctly described the right behavior while the change didn't actually implement that and continued to clone the primary repo with the fork name :upside_down_face: fixed that

hi-rustin (approved):

PR #20923, opened 2021-02-19
pod-utils: be more intelligent with grace period

stevekuznetsov:
When a user sets a grace period, today they have to consider they're
setting one for the gracful termination of their process *and* the
upload that follows it. This is error-prone and causes artifact losses.

### DIFF
--- a/prow/pod-utils/decorate/podspec.go
+++ b/prow/pod-utils/decorate/podspec.go
@@ -742,7 +742,11 @@ func decorate(spec *coreapi.PodSpec, pj *prowapi.ProwJob, rawEnv map[string]stri
 	spec.Containers = append(spec.Containers, *sidecar)
 
 	if spec.TerminationGracePeriodSeconds == nil && pj.Spec.DecorationConfig.GracePeriod != nil {
-		gracePeriodSeconds := int64(pj.Spec.DecorationConfig.GracePeriod.Seconds())
+		// Unless the user's asked for something specific, we want to set the grace period on the Pod to
+		// a reasonable value, as the overall grace period for the Pod must encompass both the time taken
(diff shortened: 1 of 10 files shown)


prow/pod-utils/decorate/podspec.go
```diff
@@ -742,7 +742,11 @@ func decorate(spec *coreapi.PodSpec, pj *prowapi.ProwJob, rawEnv map[string]stri
 	spec.Containers = append(spec.Containers, *sidecar)
 
 	if spec.TerminationGracePeriodSeconds == nil && pj.Spec.DecorationConfig.GracePeriod != nil {
-		gracePeriodSeconds := int64(pj.Spec.DecorationConfig.GracePeriod.Seconds())
+		// Unless the user's asked for something specific, we want to set the grace period on the Pod to
+		// a reasonable value, as the overall grace period for the Pod must encompass both the time taken
+		// to gracefully terminate the test process *and* the time taken to process and upload the resulting
+		// artifacts to the cloud. As a reasonable rule of thumb, assume a 80/20 split between these tasks.
+		gracePeriodSeconds := int64(pj.Spec.DecorationConfig.GracePeriod.Seconds()) * 5 / 4
```
alvaroaleman, line 749: Because expressing it as `* 1.25` would be too simple to read :joy:

alvaroaleman (approved):

prow/pod-utils/decorate/podspec.go
stevekuznetsov: https://play.golang.org/p/kAfDXDOcQ-7

```
./prog.go:10:18: constant 1.25 truncated to integer
```

1.25 is not an integer :)

PR #20927, opened 2021-02-20
kops repos to require self-approval, approve button = lgtm

olemarkus:


config/prow/plugins.yaml
```diff
@@ -72,7 +72,7 @@ approve:
   - kubernetes-client
   - kubernetes-csi
   - kubernetes-sigs
-  require_self_approval: false
+  require_self_approval: true
```
spiffxp, line 75: you probably don't mean to be changing this for `kubernetes/kubernetes`, every repo in `kubernetes-sigs`, etc.

move `kubernetes/kops` to the entry above that contains `kubernetes/website`

spiffxp (changes requested):
/hold
don't want this to accidentally merge

config/prow/plugins.yaml
olemarkus: oups. I most certainly did not.

Thanks. Moved.

spiffxp (approved):
/approve
/lgtm
/hold cancel

PR #20932, opened 2021-02-21
prow: add Github Enterprise documentation

sbueringer:
Finally got around to writing down how to configure Prow for Github enterprise.

I'm not sure if it's the right place to document it.

### DIFF
--- a/prow/getting_started_deploy.md
+++ b/prow/getting_started_deploy.md
@@ -242,6 +242,44 @@ can be error-prone, and it'll be painful when you want to replace the hmac token
 In such case, it's recommended to use the [hmac](/prow/cmd/hmac/README.md) tool to automatically manage the webhooks
 and hmac tokens for you via declarative configuration.
 
+## Deploying with GitHub Enterprise
+
+When using GitHub Enterpise (GHE), Prow must be configured slightly differently. It's possible to run GHE with or 
+without the `api` subdomain:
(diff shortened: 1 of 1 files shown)


prow/getting_started_deploy.md
```diff
@@ -242,6 +242,44 @@ can be error-prone, and it'll be painful when you want to replace the hmac token
...
+without the `api` subdomain:
+* with the `api` subdomain the endpoints are:
+   * v3: `https://api.<<github-hostname>>`
+   * graphql: `https://api.<<github-hostname>>/graphql`
+* without the `api` subdomain the endpoints are:
+  * v3: `https://<<github-hostname>>/api/v3`
+  * graphql: `https://<<github-hostname>>/api/graphql`
+
+Prow component configuration:
+* `ghproxy`:
```
alvaroaleman, line 257: Out of curiosity, why are you using ghproxy with GHE, it doesn't have the api request limiting or does it?

sbueringer: I'm not aware of the API limit (and I"m certain we wouldn't hit it).

Just configured it because I thought it might help performance a bit and because it's part of the "default" deployment.

But yeah good point, I can alternatively document it without `ghproxy` it should simplify the explanations a bit. WDYT?

alvaroaleman (approved):

PR #20937, opened 2021-02-22
Provide jobs for e2e-node graceful node shutdown tests

wzshiming:
ref https://github.com/kubernetes/kubernetes/pull/98658#issuecomment-783101360


SergeyKanzhelev (approved):
/lgtm

PR #20941, opened 2021-02-22
Add support for empty output

mborsz:
Without that, if there are no logs (e.g. gke cluster with node log dumping disabled), it fails with:

```
CommandException: One or more URLs matched no objects.
```
on the attempt to call gsutil cp on empty dir.

FTR: I tried preparing some tests for log-dump.sh, but after ~1 hour fighting with kubetest/bash I failed so for now we need to rely on manual testing.

Tested manually for both cases: empty result and nonempty result.

/assign @wojtek-t


wojtek-t:
/lgtm

thanks!

PR #20944, opened 2021-02-22
azure: add KUBE_BUILD_CONFORMANCE=n to preset-azure-cred

chewong:

Addresses https://github.com/kubernetes/kubernetes/pull/99241#issuecomment-782544503.


spiffxp (approved):
/approve
/lgtm

WDYT about moving some of these azure presets into a subdir with more relevant OWNERS? maybe `config/jobs/kubernetes/sig-cloud-provider/azure` or `config/jobs/kubernetes/sig-windows`?

PR #20946, opened 2021-02-22
test: restrict the use of specific presets to certain jobs

chewong:

Addresses the second bullet point mentioned in https://github.com/kubernetes/k8s.io/issues/1673#issuecomment-780942164

/assign @spiffxp

### DIFF
--- a/config/tests/jobs/jobs_test.go
+++ b/config/tests/jobs/jobs_test.go
@@ -388,15 +388,59 @@ func validateImagePushingImage(spec *coreapi.PodSpec) error {
 	return nil
 }
 
-// Unit test only postsubmit/periodic jobs in config/jobs/<org>/<project>/<project>-trusted.yaml can use
-// secrets for <org>/<project> in default public cluster.
+// Restrict the use of specific secrets to certain jobs in config/jobs/<org>/<project>/<basename>.yaml
 func TestTrustedJobSecretsRestricted(t *testing.T) {
-	secretsRestricted := map[string]sets.String{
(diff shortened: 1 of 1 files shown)


config/tests/jobs/jobs_test.go
```diff
@@ -1279,3 +1279,56 @@ func TestKubernetesProwJobsShouldNotUseDeprecatedScenarios(t *testing.T) {
 		t.Logf("%v/%v jobs using deprecated scenarios", fails, total)
 	}
 }
+
+// Only allow certain prow jobs to use certain labels/presets
```
spiffxp, line 1283:
```suggestion
// Restrict use of specific presets to jobs defined in {org}/{project} paths
```

spiffxp (commented):
Thank you for doing this!

This is a great start, but doesn't prevent other job definitions from using the secrets defined in these presets.

I think you could blend this with the secret-based logic in `TestTrustedJobSecretsRestricted` if you add something that gathers secret names from a preset.

spiffxp (changes requested):
Getting close...

config/tests/jobs/jobs_test.go
```diff
@@ -388,15 +388,54 @@ func validateImagePushingImage(spec *coreapi.PodSpec) error {
...
+					match = false
+					break
+				}
+			}
+			if match {
+				for _, v := range preset.Volumes {
+					if v.VolumeSource.Secret != nil {
+						secrets.Insert(v.VolumeSource.Secret.SecretName)
+					}
+				}
```
spiffxp, line 411: Please check `preset.Env` as well

spiffxp (approved):
/approve
/lgtm

PR #20947, opened 2021-02-22
Prow integration test job: use test-infra variant

chaodaiG:


fejta:
What is the point of this? The previous version seems cleaner -- more logic is encapsulated in the bash script. This is moving logic into the yaml file.

chaodaiG:
> What is the point of this? The previous version seems cleaner -- more logic is encapsulated in the bash script. This is moving logic into the yaml file.

Having been looking for examples of how the other repos are doing it, and looks like it's pretty common to install kind in prow jobs (The reason might just be like me seeing others doing it so follow the same way), an example is https://github.com/kubernetes/test-infra/blob/1a96fc6301c4b48ae45102222c200896c48631c2/config/jobs/kubernetes/sig-network/sig-network-kind.yaml#L133. 

+ @BenTheElder , what would be the recommended approach of installing kind in prow jobs? In script (I.e. ./hack/ensure-kind.sh), bazel rule (i.e. `bazel run //ensure-kind`), or in prowjob?

fejta:
a) The entrypoint to prow should be a trivially executable script that is source controlled.

Meets the criteria:
```
prow/test/integration/test.sh --config=ci
```

Does not meet this criteria:
```
bash -c curl -sSL https://kind.sigs.k8s.io/dl/latest/linux-amd64.tgz | tar xvfz - -C "${PATH%%:*}/" && ./prow/test/integration/integration-test.sh  --config=ci
```

A person trying to invoke this manually is far more likely to mess up the latter than the former. Please move this logic into a script and out of the prowjob config.

b) You shouldn't be curling/untarring the same thing thousands of times a day. Please download a specific version of kind, and do this when creating the image, not when invoking testing.

chaodaiG:
Created https://github.com/kubernetes/test-infra/pull/20957 for installing kind in kubekins-e2e image. Will update this PR to use the new image once #20957 is merged.

/hold

BenTheElder:
> @BenTheElder , what would be the recommended approach of installing kind in prow jobs? In script (I.e. ./hack/ensure-kind.sh), bazel rule (i.e. bazel run //ensure-kind), or in prowjob?

For test-infra I would think either:
- A `WORKSPACE` dependency (with a `select` rule for the host platform)
- Pre-install in the image being used in CI (and perhaps check the host version is current stable in the test script?)

Of the stable (`v0.10.0` currently) release.

Not curling the unadvertised latest tgz in a prowjob, please don't use that unless you're writing a job to test Kubernetes @ HEAD (and even there it's ... a tradeoff).

---

Semi-off topic/long-winded 😞  on why that other approach exists at all:
<details>
The "latest" binary/compressed tarball you were previously using is *unadvertised* and expressly for the purpose of testing Kubernetes itself at unreleased versions.  


For that purpose only we frequently need to make changes to keep up with the requirements for developing / running unreleased Kubernetes at `HEAD`. So those Kubernetes jobs download a ~2.5MB (compressed, we've minimized the binary size etc. a *ton*) tgz with the latest kind build, which we've tested against the same Kubernetes tests before merge.

Anything else that is **not** trying to test Kubernetes at `HEAD` should use the stable kind release and should ideally cache it (in the workspace cache, in the image, something ...). This is more efficient and will not be surprised by kind changes in-between releases.

Also that shell one-liner is not in a script because we haven't gotten agreement to add such a script to k/k yet (k/k doesn't want to wind up with a proliferation of these for various tools ...), but we also don't want to pay the penalty of cloning an additional repo (which will cost more than even downloading this tarball) for that one-liner when testing Kubernetes.
That does not make sense here, and arguably doesn't make a lot of sense in Kubernetes either.

In the future when testing Kubernetes we will probably emit a `REPRODUCING.txt` artifact (been drafting a doc for this) with instructions to obtain the exact `kind` version used and run the tests the same way locally since Kubernetes e2e is generally a bit fiddly. But ideally you just want a simple bazel/make/shell target to invoke and there's no blockers for that here.
</details>

fejta (approved):
/hold

BenTheElder (commented):
Title doesn't seem to match the diff?
(I expected to see the `image` updated?)

chaodaiG:
> Title doesn't seem to match the diff?
> (I expected to see the `image` updated?)

yeah it was intended, I was waiting for the fresh image to be pushed :)

chaodaiG:
> > Title doesn't seem to match the diff?
> > (I expected to see the `image` updated?)
> 
> yeah it was intended, I was waiting for the fresh image to be pushed :)

@bentheelder, the PR is updated now with updated image tag

fejta (approved):

chaodaiG:
Fixed merge conflict due to autobump, need another approve. @fejta 
/unhold

fejta (approved):

PR #20949, opened 2021-02-22
Use generic autobump image for autobump job

mpherman2:
Replacing the current way that autobumper is run with the way that uses the generic_autobump images. This will allow us to use the generic_autobump image in k8s/t-i as a reference when bumping the autobump jobs in other repos.


spiffxp (approved):
/approve
/lgtm
/hold
remove /hold when you're ready to deploy

michelle192837 (approved):

PR #20957, opened 2021-02-23
Install kind on kubekins-e2e image in test-infra variant

chaodaiG:
Prow integration test requires kind installation, install it in kubekins-e2e image so that the test doesn't need to download kind every time


fejta (approved):
/hold

chaodaiG:
@BenTheElder , any concern here?

PR #20963, opened 2021-02-23
Move periodic-kubernetes-bazel-build jobs to release-informing

spiffxp:
Part of https://github.com/kubernetes/test-infra/issues/18549

The intent here is to remove these jobs from the list of release-blocking jobs that need to be migrated over.

Given https://github.com/kubernetes/enhancements/issues/2420 I think we should stop considering these as release-blocking, since we already have release-blocking non-bazel build jobs


spiffxp:
/cc @ameukam @justaugustus @hasheddan 
/hold
for comment from @kubernetes/release-engineering

hasheddan (approved):
/lgtm

Issue #19483, opened 2020-10-07
release-blocking jobs must run in dedicated cluster: ci-kubernetes-build


spiffxp:
**What should be cleaned up or changed**:

This is part of #18549 

To properly monitor the outcome of this, you should be a member of k8s-infra-prow-viewers@kubernetes.io.  PR yourself into https://github.com/kubernetes/k8s.io/blob/master/groups/groups.yaml#L603-L628 if you're not a member.

NOTE: I am not tagging this as "help wanted" because it is blocked on https://github.com/kubernetes/k8s.io/issues/846.  I would also recommend doing ci-kubernetes-build-fast first.  Here is my guess at how we could do this:
- create a duplicate job that pushes to the new bucket writable by k8s-infra-prow-build
- ensure it's building and pushing appropriately
- update a release-blocking job to pull from the new bucket
- if no problems, roll out changes progressively
  - a few more jobs in release-blocking
  - all jobs in release-blocking that use this job's results
  - a job that still runs in the "default" cluster
  - all jobs that use this job's results
- rename jobs / get rid of the job that runs on the "default" cluster
- do the same for release-branch variants, can probably do a faster rollout

It will be helpful to note the date/time that PR's merge.  This will allow you to compare before/after behavior.

Things to watch for the job
- https://prow.k8s.io/?job=ci-kubernetes-build
  - does the job start failing more often?
  - does the job start going into error state?
- https://testgrid.k8s.io/presubmits-kubernetes-blocking#ci-kubernetes-build&graph-metrics=test-duration-minutes
  - does the job duration look worse than before? spikier than before?
- https://storage.googleapis.com/k8s-gubernator/triage/index.html?pr=1&job=ci-kubernetes-build
  - do more failures show up than before?
- https://prow.k8s.io/job-history/gs/kubernetes-jenkins/pr-logs/directory/ci-kubernetes-build
  - (can be used to answer some of the same questions as above)
[...]

PR #20964, opened 2021-02-23
Deprecate google-hosted kubernetes-build jobs

spiffxp:
Part of https://github.com/kubernetes/test-infra/issues/18549 (more specifically part of https://github.com/kubernetes/test-infra/issues/19483)

Now that all release-blocking jobs use artifacts stored in community-owned `gs://k8s-release-dev` (ref: https://github.com/kubernetes/test-infra/pull/20885), it's time to call the jobs that produce those artifacts release-blocking, and demote the jobs that write to `gs://kubernetes-release-dev` as release-informing.

Deprecate jobs that write to `gs://kuebrnetes-release-dev` by:
- appending deprecated to their job name and testgrid tab name
- moving from release-blocking to release-informing

Promote jobs that write to `gs://k8s-release-dev` by:
- removing canary from their job name and testgrid tab name
- moving from release-informing to release-blocking
- removing from sig-testing-canaries

/hold
This is a "breaking" change that will split job history across the rename boundary since tools like testgrid and kettle don't understand the concept of job renames.

Would like consensus from SIG Release subprojects

### DIFF
--- a/config/jobs/kubernetes/sig-release/kubernetes-builds.yaml
+++ b/config/jobs/kubernetes/sig-release/kubernetes-builds.yaml
@@ -30,7 +30,7 @@ presubmits:
 
 periodics:
 - interval: 1h
-  name: ci-kubernetes-build
+  name: ci-kubernetes-build-deprecated
   labels:
     preset-service-account: "true"
     preset-dind-enabled: "true"
@@ -63,11 +63,11 @@ periodics:
   annotations:
     fork-per-release: "true"
     fork-per-release-replacements: "k8s-master -> k8s-beta"
-    testgrid-dashboards: sig-release-master-blocking
-    testgrid-tab-name: build-master
+    testgrid-dashboards: sig-release-master-informing
+    testgrid-tab-name: build-master-deprecated
     testgrid-alert-email: release-managers+alerts@kubernetes.io, release-team@kubernetes.io
 
-- name: ci-kubernetes-build-canary
+- name: ci-kubernetes-build
   interval: 1h
   cluster: k8s-infra-prow-build
   decorate: true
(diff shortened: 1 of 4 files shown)


spiffxp:
FYI @kubernetes/release-engineering @kubernetes/ci-signal @kubernetes/release-team-leads

hasheddan (commented):
I'm very happy to see this change, thanks @spiffxp! My one concern here is that there are still some critical components (such as dl.k8s.io) that utilize the builds pushed to the old infra. Do we feel good about making this change before all consumers have moved over?

justaugustus:
For SIG Release and associated subprojects:
/lgtm
/approve

> My one concern here is that there are still some critical components (such as dl.k8s.io) that utilize the builds pushed to the old infra. Do we feel good about making this change before all consumers have moved over?

Discussed and approved at [today's SIG Release meeting](https://docs.google.com/document/d/1Fu6HxXQu8wl6TwloGUEOXVzZ1rwZ72IAhglnaAMCPqA/edit#heading=h.nxhqj64hvnos).

Two things to note:
1. We're deprecating the jobs but not removing them here, so there should be no impact for dl.k8s.io consumers
2. We're going to lose job history (or rather, job history will move to a different place for new runs), so it's better to do this sooner rather than later

Will leave @hasheddan to lifting the hold.

hasheddan (approved):
/lgtm
/hold cancel

spiffxp:
>Do we feel good about making this change before all consumers have moved over?

FWIW I took a quick look at what builds are present where: https://gist.github.com/spiffxp/9a339a93ae889d1378e6f2259a733895

>We're deprecating the jobs but not removing them here, so there should be no impact for dl.k8s.io consumers

Exactly.  I _would_ like to switch dl.k8s.io/ci to point to k8s-release-dev at some point, but not until we have a better handle on ensuring build parity between the buckets (maybe using STS to sync, etc.)

PR #20965, opened 2021-02-23
kubernetes-sigs/blob-csi-driver: add new job for e2e test using blobfuse proxy

boddumanohar:
https://github.com/kubernetes-sigs/blob-csi-driver/pull/349  In this PR, I have added support for blob-csi-driver NodeStage calls to be able to mount via a proxy that is running either or node or by using a daemonset.

This new jobs will test if NodeStage calls through blobfuse-proxy are working as expected.

In the above PR, blob-csi-driver will looks for an env: `ENABLE_BLOBFUSE_PROXY` if that is available, then it will install blobfuse-proxy as a daemonset.

### DIFF
--- a/config/jobs/kubernetes-sigs/blob-csi-driver/blob-csi-driver-config.yaml
+++ b/config/jobs/kubernetes-sigs/blob-csi-driver/blob-csi-driver-config.yaml
@@ -190,3 +190,55 @@ presubmits:
       testgrid-tab-name: pr-blob-csi-driver-e2e-vmss
       description: "Run E2E tests on a VMSS-based cluster for Azure Blob Storage CSI driver."
       testgrid-num-columns-recent: '30'
+  - name: pull-blob-csi-driver-e2e-proxy
+    decorate: true
+    always_run: true
+    path_alias: sigs.k8s.io/blob-csi-driver
+    branches:
+    - master
+    labels:
+      preset-service-account: "true"
+      preset-azure-cred: "true"
+      preset-dind-enabled: "true"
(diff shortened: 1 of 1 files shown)


andyzhangx (approved):
/lgtm
/approve

config/jobs/kubernetes-sigs/blob-csi-driver/blob-csi-driver-config.yaml
```diff
@@ -190,3 +190,55 @@ presubmits:
...
+        - --ginkgo-parallel=1
+        - --timeout=420m
+        securityContext:
+          privileged: true
+        env:
+          - name: ENABLE_BLOBFUSE_PROXY
+            value: "true"
+    annotations:
+      testgrid-dashboards: provider-azure-blobfuse-csi-driver, provider-azure-presubmit
+      testgrid-tab-name: pr-blob-csi-driver-e2e
```
andyzhangx, line 242: testgrid-tab-name: pull-blob-csi-driver-e2e-proxy

config/jobs/kubernetes-sigs/blob-csi-driver/blob-csi-driver-config.yaml
```diff
@@ -190,3 +190,55 @@ presubmits:
...
+      description: "Run E2E tests for Azure Blob Storage CSI driver."
```
andyzhangx, line 243: Run E2E tests for Azure Blob Storage CSI driver based on blobfuse-proxy

PR #20966, opened 2021-02-23
bump kind presubmit jobs resources

aojea:
Since these jobs are flaking, mainly because pods times out waiting to be running, one guess is that this is due to resources constraint.

With no evidence, maybe try something and observe how it affects is a good approach

xref: https://github.com/kubernetes/test-infra/pull/18825#issuecomment-673632240

### DIFF
--- a/config/jobs/kubernetes/sig-testing/kubernetes-kind-presubmits.yaml
+++ b/config/jobs/kubernetes/sig-testing/kubernetes-kind-presubmits.yaml
@@ -39,10 +39,10 @@ presubmits:
           privileged: true
         resources:
           limits:
-            cpu: 4
+            cpu: 7
             memory: 9000Mi
           requests:
-            cpu: 4
+            cpu: 7
             memory: 9000Mi
     annotations:
       testgrid-num-failures-to-alert: '10'


spiffxp (changes requested):
I personally would rather see this done for one job first, whichever is highest traffic, to confirm whether this works

spiffxp:
https://prow.k8s.io/?type=presubmit&job=*kind*&cluster=k8s-infra-prow-build behavior over the last 12h

spiffxp:
https://storage.googleapis.com/k8s-gubernator/triage/index.html?job=kubernetes.*kind
Screen Shot 2021-02-23 at 12 17 37 PM

There does seem to be a rise in this failure since ~2/19.  The k8s-infra-prow-build cluster was updated to use a node-local dns cache add-on, and upgraded to v1.17 at that time (ref: https://github.com/kubernetes/k8s.io/issues/1541#issuecomment-782314737 and https://github.com/kubernetes/k8s.io/issues/1541#issuecomment-783527464)

It's worth noting, this is happening for ci jobs too, which are already at 7 cpu, e.g. https://storage.googleapis.com/kubernetes-jenkins/logs/ci-kubernetes-kind-e2e-parallel-1-20/1364237532232945664/prowjob.json

I would try looking into logs on the cluster e.g. [`resource.type=~"k8s_(cluster|node|pod)" resource.labels.project_id="k8s-infra-prow-build"`](https://console.cloud.google.com/logs/query;query=resource.type%3D~%22k8s_%2528cluster%7Cnode%7Cpod%2529%22;timeRange=P5D?project=k8s-infra-prow-build) and trying to filter down to see if you can identify a change in behavior in the last 5 days

BenTheElder:
> I personally would rather see this done for one job first, whichever is highest traffic, to confirm whether this works

+1, also we should be more aggressive _for the one job_, we should schedule up ~the whole node (7.X) so we can say "yes it works without noisy io neighbors" or "no we still see throttling + pod startup flakes". If it works then we can follow up with less aggressively tuning another job. If it doesn't this is only going to make it harder to schedule and there's some evidence that this may not be sufficient.

We also need to look into https://github.com/kubernetes/k8s.io/issues/1187

aojea:
/hold
this sounds like a plan

spiffxp:
>~the whole node (7.X)

This has gotten us into trouble almost every time someone does it.  Stick to 7 cpu for the test container, the additional overhead of pod utils puts you at 7.4.  That's more than enough to prevent other jobs from trying to sneak in.

BenTheElder:
> This has gotten us into trouble almost every time someone does it. Stick to 7 cpu for the test container, the additional overhead of pod utils puts you at 7.4. That's more than enough to prevent other jobs from trying to sneak in.

7.4 / ~* sounds sufficient, I forgot about the podutils requests.

BenTheElder:
that particular error string increased ~6 days ago?

https://github.com/kubernetes/kubernetes/pull/99147#event-4346325171 🙃

aojea:
> that particular error string increased ~6 days ago?
> 
> [kubernetes/kubernetes#99147 (comment)](https://github.com/kubernetes/kubernetes/pull/99147#event-4346325171)

hahah, at least we know the reasons, pods are not being able to be running and ready

spiffxp:
https://github.com/kubernetes/test-infra/pull/20966#issuecomment-784371223 - as a reminder, this graph is for CI jobs _only_, which already have 7 cpu.  

https://github.com/kubernetes/k8s.io/issues/1703 raised quota, if you want to modify this to update just `pull-kubernetes-e2e-kind` I'm willing to merge

spiffxp:
(images from local grafana dashboard, someday I'll get this public, or setup a datastudio dashboard)

Is this due to increase in traffic? Probably not. Looking back 6 months, failure rate wasn't this bad pre-holiday lull.
Screen Shot 2021-02-25 at 12 58 48 PM

Looking at the last 90 days, for just the kind jobs, it's pretty clear pull-kubernetes-kind-e2e jumped sometime at the beginning of february
Screen Shot 2021-02-25 at 1 00 47 PM

I would suggest trying to figure out what changed between 2020-02-04 and 2020-02-07
Screen Shot 2021-02-25 at 1 03 22 PM

Same story for the CI jobs
Screen Shot 2021-02-25 at 1 05 13 PM

spiffxp:
What I don't understand is why triage doesn't tell the same story:
https://storage.googleapis.com/k8s-gubernator/triage/index.html?date=2020-02-11&job=kind
Screen Shot 2021-02-25 at 1 09 20 PM

aojea:
> What I don't understand is why triage doesn't tell the same story:
> https://storage.googleapis.com/k8s-gubernator/triage/index.html?date=2020-02-11&job=kind
> Screen Shot 2021-02-25 at 1 09 20 PM

it is actually the opposite, the failure rate decreased after
> tl;dr I would suggest trying to figure out what changed between 2020-02-04 and 2020-02-07

aojea:
/hold cancel
just only kind-presubmit

spiffxp (approved):
/approve 
/lgtm

spiffxp:
Triage filters out failures that don't fit into clusters of a certain size by failure text. So it's still the case that the jobs are failing more. Just not due to easily identified common error messages

PR #20968, opened 2021-02-23
provider-openstack: migrated gs bucket and cleanup

chrigl:
We are requested to migrate our GS bucket to k8s-federated-conformance
ones. See
https://github.com/kubernetes/k8s.io/issues/1311#issuecomment-778507615

While I was there, I moved the provider-openstack related dashboards
under sig-cloud-provider and alongside azure, aws and gcp. In the
future, I would like to integrate csi related signals as well. But
that's something for a separate task.

This PR also removes tests for kubernetes v1.16, replaced by v1.18,
v1.19 and v1.20.

### DIFF
--- a/config/testgrids/conformance/conformance-all.yaml
+++ b/config/testgrids/conformance/conformance-all.yaml
@@ -5,7 +5,6 @@ dashboard_groups:
     - conformance-apisnoop
     - conformance-gce
     - conformance-kind
-    - conformance-cloud-provider-openstack
     - conformance-cloud-provider-huaweicloud
     - conformance-cloud-provider-vsphere
     - conformance-vsphere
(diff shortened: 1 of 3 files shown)


config/testgrids/kubernetes/sig-cloud-provider/openstack/OWNERS
```diff
@@ -0,0 +1,13 @@
+# See the OWNERS docs at https://go.k8s.io/owners
+
+reviewers:
+- chrigl
+- lingxiankong
+- ramineni
+approvers:
```
spiffxp, line 7: thank you!

config/testgrids/conformance/conformance-all.yaml
```diff
@@ -104,28 +96,6 @@ dashboards:
 
 - name: conformance-apisnoop
 - name: conformance-gce
-- name: conformance-cloud-provider-openstack
```
spiffxp, line 107: FWIW I have found it useful to see conformance results aggregated in this one dashboard group (e.g. https://testgrid.k8s.io/conformance) but we can revisit later

spiffxp (approved):
/approve
/lgtm

PR #20970, opened 2021-02-23
Ignore Linux only test (uses hostnetwork)

jsturtevant:
This test started using `hostnetwork` in windows during 1.20 and we are not able to backport it due to conformance issues: https://github.com/kubernetes/kubernetes/pull/99081/#issuecomment-780944375

This was fixed in 1.21.

/assign @chewong 

/cc @ravisantoshgudimetla


chewong (approved):
/lgtm

PR #20978, opened 2021-02-23
Always run migration ebs csi test

wongma7:
Since it seems to be passing consistently with inline bug fix , make it required.

https://testgrid.k8s.io/provider-aws-ebs-csi-driver#migration-test-latest


ayberk:
/lgtm
/retest

Nice!

PR #20979, opened 2021-02-23
kubekins-e2e dockerfile: fix bash weirdness by using if instead of &&

chaodaiG:
`&&` style condition makes the docker build exit 1 when the variable evaluated is an empty string, this was verified locally with docker build. The fix is using explicit `if` statement. See where it failed https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-test-infra-push-kubekins-e2e/1364342488625582080


chaodaiG:
fyi @BenTheElder @fejta

spiffxp (approved):
/approve
/lgtm

chaodaiG:
flaky test
/retest

PR #20986, opened 2021-02-24
Add config merger lists for k8s

chases2:
The [config merger](https://github.com/GoogleCloudPlatform/testgrid/tree/master/cmd/config_merger) utility allows TestGrid configurations to be merged together without a PR request (compare to [Transfigure](https://github.com/kubernetes/test-infra/tree/master/testgrid/cmd/transfigure)) after being initially registered.

### DIFF
--- a/config/mergelists/BUILD.bazel
+++ b/config/mergelists/BUILD.bazel
@@ -1,0 +1,22 @@
+package(default_visibility = ["//visibility:public"])
+
+filegroup(
+    name = "mergelists",
+    srcs = glob([
+        "*.yaml",
+    ]),
+)
+
+filegroup(
+    name = "package-srcs",
+    srcs = glob(["**"]),
+    tags = ["automanaged"],
+    visibility = ["//visibility:private"],
+)
+
+filegroup(
+    name = "all-srcs",
+    srcs = [":package-srcs"],
+    tags = ["automanaged"],
+    visibility = ["//visibility:public"],
+)
--- a/config/mergelists/canary.yaml
+++ b/config/mergelists/canary.yaml
@@ -1,0 +1,4 @@
+target: "gs://k8s-testgrid-canary/config"
+sources:
+- name: ""
+  location: "gs://k8s-testgrid-canary/configs/k8s/config"
--- a/config/mergelists/prod.yaml
+++ b/config/mergelists/prod.yaml
@@ -1,0 +1,4 @@
+target: "gs://k8s-testgrid/config"
+sources:
+- name: ""
+  location: "gs://k8s-testgrid/configs/k8s/config"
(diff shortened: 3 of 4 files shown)


chases2:
Context: https://github.com/GoogleCloudPlatform/testgrid/issues/326

config/jobs/kubernetes/test-infra/test-infra-trusted.yaml
```diff
@@ -1031,6 +1031,28 @@ periodics:
...
     description: Runs label_sync to synchronize GitHub repo labels with the label config defined in label_sync/labels.yaml.
+- name: test-infra-merge-testgrid-config-canary
+  interval: 30m
+  cluster: test-infra-trusted
+  max_concurrency: 1
+  decorate: true
+  spec:
+    serviceAccountName: testgrid-config-updater
+    containers:
+    - image: gcr.io/k8s-testgrid/config_merger:latest
```
fejta, line 1042: Can we please pin an actual version, rather than a floating tag like latest?

config/jobs/kubernetes/test-infra/test-infra-trusted.yaml
```diff
@@ -1031,6 +1031,28 @@ periodics:
...
+      command:
+      - /config_merger
```
fejta, line 1044: should this be /app/whatever/config_merge.binary by chance?

config/jobs/kubernetes/test-infra/test-infra-trusted.yaml
```diff
@@ -1031,6 +1031,28 @@ periodics:
...
+      args:
```
fejta, line 1045: Thoughts on having this follow the updater pattern of:
* Adding a `--wait=30m` flag that causes it to run forever
* Making this a deployment rather than a periodic release that
  - lives in the testgrid repo
  - becomes part of the testgrid release

chases2: Having this be part of the TestGrid cluster is a good idea. There will still need to be some changes to the Configurator job here.

config/jobs/kubernetes/test-infra/test-infra-trusted.yaml
chases2: Fixed in TG review

chases2:
A separate PR to the TestGrid repo will onboard the actual config-merger controllers. This PR now sets up the configs in cloud storage.

chases2:
I have plans to add verification tests for the mergelists, but implementing them requires us to update our version of TestGrid from 0.0.30 -> 0.0.54, which has some implications for Configurator.

It also appeared to make subdirectory defaulting start working, so the update isn't trivial. It'll be covered in a later PR.

michelle192837 (commented):
/hold
Unhold if you're good to go.

michelle192837 (approved):

fejta (approved):

PR #20987, opened 2021-02-24
configurator: prefix gs/ to prow-derived open_test_templates

spiffxp:
Followup to https://github.com/kubernetes/test-infra/pull/20902

https://testgrid.k8s.io/conformance-all#Conformance%20-%20GCE%20-%201.20

click on a cell, get sent to https://prow.k8s.io/view/kubernetes-jenkins/logs/ci-kubernetes-gce-conformance-latest-1-20/1364346208910315520

```
error rendering spyglass page: error when resolving real path kubernetes-jenkins/logs/ci-kubernetes-gce-conformance-latest-1-20/1364346208910315520: 
  error opening file bucket: 
  open blob.Bucket: 
  no driver registered for "kubernetes-jenkins" for URL "kubernetes-jenkins://logs"; 
  available schemes: mem, s3
```

We should be defaulting to `gs://` since that's the only place testgrid updater scrapes from IIRC

Fixes https://github.com/kubernetes/test-infra/issues/20988

### DIFF
--- a/config/testgrids/default.yaml
+++ b/config/testgrids/default.yaml
@@ -20,7 +20,7 @@ default_test_group:
 
 default_dashboard_tab:
   open_test_template: # The URL template to visit after clicking on a cell
-    url: https://prow.k8s.io/view/gcs/<gcs_prefix>/<changelist>
+    url: https://prow.k8s.io/view/gs/<gcs_prefix>/<changelist>
   file_bug_template: # The URL template to visit when filing a bug
     url: https://github.com/kubernetes/kubernetes/issues/new
     options:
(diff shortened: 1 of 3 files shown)


spiffxp:
/cc @amwat @chases2 
you noticed
/cc @fgimenez 
FYI

config/testgrids/default.yaml
```diff
@@ -20,7 +20,7 @@ default_test_group:
 
 default_dashboard_tab:
   open_test_template: # The URL template to visit after clicking on a cell
-    url: https://prow.k8s.io/view/gcs/<gcs_prefix>/<changelist>
+    url: https://prow.k8s.io/view/gs/<gcs_prefix>/<changelist>
```
spiffxp, line 23: per @fejta 
>btw should now be gs not gcs
>
>Yes:
>https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-test-infra-deploy-prow/1364279379974164480
>
>Deprecated:
>https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/post-test-infra-deploy-prow/1364279379974164480
>
>(gs is the official scheme for gcs)
>
>(previously gcs was meant to distinguish from "query the api server rather than blob storage")

chases2 (approved):

PR #20989, opened 2021-02-24
jobs: enforce kubernetes CI policy for blocking jobs

spiffxp:
- First commit enables "release-blocking must run on k8s-infra"
- Next commit refactors into "CI Policy" tests instead of individual tests for each policy

I think this is the last thing to do for https://github.com/kubernetes/test-infra/issues/18549 but am going to avoid auto-closing with this PR

### DIFF
--- a/config/tests/jobs/jobs_test.go
+++ b/config/tests/jobs/jobs_test.go
@@ -1036,17 +1036,14 @@ func TestKubernetesMergeBlockingJobMustRunOnK8sInfraProwBuild(t *testing.T) {
 	}
 }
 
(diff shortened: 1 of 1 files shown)


spiffxp:
/uncc @wojtek-t 
/cc @thejoycekung 
for @kubernetes/ci-signal 
/cc @ameukam @hasheddan 
SIG Release TL's who've helped with https://github.com/kubernetes/test-infra/issues/18549

hasheddan (approved):
/lgtm
Thanks @spiffxp!

PR #20991, opened 2021-02-24
kubernetes cached make test: disable -race, simplify running

BenTheElder:
follow up to https://github.com/kubernetes/test-infra/pull/20977

we can revert back to enabling the race detector later if we want, but I would like to collect data on how it performs when we *are* able to cache this and run it in presubmit, which won't happen with `-race` enabled.

### DIFF
--- a/config/jobs/kubernetes/sig-testing/make-test.yaml
+++ b/config/jobs/kubernetes/sig-testing/make-test.yaml
@@ -25,7 +25,7 @@ presubmits:
               mkdir -p _output/local/go/
               time tar -xzf cache.tar.gz -C _output/local/go
               # Run tests as usual
-              time make test KUBE_RACE=-race KUBE_TIMEOUT=--timeout=240s
+              time make test KUBE_TIMEOUT=--timeout=240s
           # TODO: direct copy from pull-kubernetes-bazel-test, tune these
           resources:
             limits:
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes/sig-testing/make-test.yaml
```diff
@@ -58,8 +58,7 @@ periodics:
             - |
               result=0
               # Run the tests as usual
-              ( cd hack/tools && GO111MODULE=on go install gotest.tools/gotestsum ) || result=$?
```
amwat, line 61: Don't we still need this ?

amwat (commented):
/lgtm
/approve
/hold 
 One question, feel free to remove the hold

config/jobs/kubernetes/sig-testing/make-test.yaml
amwat: nvm just saw the commit description

/hold cancel

BenTheElder: liggitt fixed this when we started migrating things to `make test`, it does the sane thing and installs a copy from hack/tools/go.mod's version into a local GOBIN and uses that instead of depending on one in system PATH

BenTheElder: (in 1.21 devel)

PR #21000, opened 2021-02-24
Make the 1.20 pre-submit and 1.20 periodic suite skip the same tests

olemarkus:


olemarkus:
/cc @hakman 

Hope this will make the 1.20 presubmit tests work. They seem to be stuck on the RuntimeClass stuff.

hakman:
> Hope this will make the 1.20 presubmit tests work. They seem to be stuck on the RuntimeClass stuff.

Strange because should not be triggered for Docker

PR #21003, opened 2021-02-24
Update to newest golang version 1.16

dims:
needed for https://github.com/kubernetes/client-go/pull/932

### DIFF
--- a/config/jobs/kubernetes/client-go/client-go-presubmits.yaml
+++ b/config/jobs/kubernetes/client-go/client-go-presubmits.yaml
@@ -6,7 +6,7 @@ presubmits:
     path_alias: k8s.io/client-go
     spec:
       containers:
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes/client-go/client-go-presubmits.yaml
```diff
@@ -6,7 +6,7 @@ presubmits:
     path_alias: k8s.io/client-go
     spec:
       containers:
-      - image: golang:1.14.4
+      - image: golang:1.16
```
liggitt, line 9: I'd probably leave this at go1.15 for now, since that is what k8s.io/kubernetes is still on

dims: fixing.

PR #21008, opened 2021-02-24
start running make test on all Kubernetes PRs, drop caching to match bazel test …

BenTheElder:
…-race

/cc @spiffxp @dims @liggitt

### DIFF
--- a/config/jobs/kubernetes/sig-testing/make-test.yaml
+++ b/config/jobs/kubernetes/sig-testing/make-test.yaml
@@ -1,6 +1,6 @@
 presubmits:
   kubernetes/kubernetes:
-  - name: pull-kubernetes-make-test
+  - name: pull-kubernetes-unit
     annotations:
       testgrid-dashboards: sig-testing-misc
     decorate: true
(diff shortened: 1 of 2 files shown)


config/jobs/kubernetes/sig-testing/make-test.yaml
```diff
@@ -10,22 +10,15 @@ presubmits:
     labels:
       preset-service-account: "true"
     optional: true
-    always_run: false
+    always_run: true
```
BenTheElder, line 13: I've been testing this, it behaves as well as the bazel job.
We can start running it everywhere but *not* require it, then transition which job is required, then phase out the old job.

config/jobs/kubernetes/sig-testing/make-test.yaml
```diff
@@ -10,22 +10,15 @@ presubmits:
...
     spec:
       containers:
         - image: gcr.io/k8s-testimages/kubekins-e2e:v20210224-a27864c-master
           command:
             - runner.sh
-            - bash
           args:
-            - -c
-            - |
-              # Restore the cache
```
BenTheElder, line 23: this is not helping us currently, prefer to match bazel for now (-race)

config/jobs/kubernetes/sig-testing/make-test.yaml
```diff
@@ -10,22 +10,15 @@ presubmits:
...
-              time curl https://storage.googleapis.com/kubernetes-jenkins/cache/poc/k8s-test-cache.tar.gz -o cache.tar.gz
-              mkdir -p _output/local/go/
-              time tar -xzf cache.tar.gz -C _output/local/go
-              # Run tests as usual
-              time make test KUBE_TIMEOUT=--timeout=240s
+            # TODO: stop depending on custom timeout
+            - make test KUBE_RACE=-race KUBE_TIMEOUT=--timeout=240s
```
BenTheElder, line 21: we should reduce this down to `make test KUBE_RACE=-race` but that should come later, priority is smooth transition

gubernator/config.yaml
```diff
@@ -34,6 +34,7 @@ jobs:
   - pull-kubernetes-e2e-kind
   - pull-kubernetes-e2e-kind-ipv6
   - pull-kubernetes-integration
+  - pull-kubernetes-make-test
```
spiffxp, line 37: Silly nit I didn't catch before the PR introducing this job merged: WDYT about `pull-kubernetes-test` or `pull-kubernetes-unit` instead? Avoid baking the name of how they're invoked into the job

liggitt: +1 for pull-kubernetes-unit

BenTheElder: Sure

BenTheElder: done

spiffxp (approved):
/approve
/lgtm

PR #21009, opened 2021-02-24
kubernetes kind: update ipv6 job ...

BenTheElder:
... to use make.

This job is extremely close to the ipv4 job, we can see how the timing behaves and then look to migrate the rest.

/cc @spiffxp @dims @liggitt @aojea

### DIFF
--- a/config/jobs/kubernetes/sig-testing/kubernetes-kind-presubmits.yaml
+++ b/config/jobs/kubernetes/sig-testing/kubernetes-kind-presubmits.yaml
@@ -110,8 +110,6 @@ presubmits:
     skip_branches:
     - release-\d+\.\d+ # per-release settings
     labels:
-      preset-bazel-scratch-dir: "true"
-      preset-bazel-remote-cache-enabled: "true"
       preset-dind-enabled: "true"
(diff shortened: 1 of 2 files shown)


config/jobs/kubernetes/sig-testing/kubernetes-kind-presubmits.yaml
```diff
@@ -134,8 +132,6 @@ presubmits:
           value: \[Slow\]|\[Disruptive\]|\[Flaky\]|\[Feature:.+\]|\[sig-storage\]|PodSecurityPolicy|LoadBalancer|load.balancer|In-tree.Volumes.\[Driver:.nfs\]|PersistentVolumes.NFS|Simple.pod.should.support.exec.through.an.HTTP.proxy|subPath.should.support.existing
         - name: PARALLEL
           value: "true"
-        - name: BUILD_TYPE
```
BenTheElder, line 137: the default is make, the logic is "if set to bazel use bazel, else use make", so we can just drop this and have one less magic CI setting to clean up.

aojea:
hahah bazel don't want to die, is fighting back :rofl: 
/retest

PR #21010, opened 2021-02-24
Upgrade COS from 85 to 89

bsdnet:
Update COS from current LTS 85 to next LTS 89


dims:
/approve
/lgtm

let's roll the dice. feel free to remove `/hold` when you feel ready.

PR #21016, opened 2021-02-25
kubeadm: add jobs to test the kubelet n-1 and n-2 skew

neolit123:
see discussion here:
https://groups.google.com/d/msgid/kubernetes-sig-architecture/CAH1uJ6U4qftRnmWthjYKtBKAzRFFnX7XWLMkSbyVWHUsX8%3DBBg%40mail.gmail.com?utm_medium=email&utm_source=footer

and explanation here:
https://github.com/kubernetes/kubeadm/pull/2396

https://github.com/kubernetes/kubeadm/pull/2396 is required before this merges.

### DIFF
--- a/config/jobs/kubernetes/sig-cluster-lifecycle/kubeadm-kinder-kubelet-x-on-y.yaml
+++ b/config/jobs/kubernetes/sig-cluster-lifecycle/kubeadm-kinder-kubelet-x-on-y.yaml
@@ -1,0 +1,282 @@
+# periodic jobs
+
+periodics:
+- name: ci-kubernetes-e2e-kubeadm-kinder-kubelet-1-20-on-latest
+  interval: 2h
+  decorate: true
+  labels:
+    preset-dind-enabled: "true"
+    preset-kind-volume-mounts: "true"
+  annotations:
+    testgrid-dashboards: sig-cluster-lifecycle-kubeadm, sig-node-kubelet
+    testgrid-tab-name: kubeadm-kinder-kubelet-1-20-on-latest
+    testgrid-alert-email: kubernetes-sig-cluster-lifecycle+testgrid@googlegroups.com, kubernetes-sig-node+testgrid@googlegroups.com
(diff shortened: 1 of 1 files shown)


kfox1111:
I think there used to be a check for kubeadm upgrade that would prevent an upgrade of the control plane with n-2 kubelets. Is that still there? Should it be disabled once these checks are enabled?

neolit123:
> I think there used to be a check for kubeadm upgrade that would prevent an upgrade of the control plane with n-2 kubelets. Is that still there? Should it be disabled once these checks are enabled?

these test jobs don't include `kubeadm upgrade`, as this was not discussed.
`kubeadm upgrade` does have a check for unsupported current kubelet version on the node which can be skipped with `-f`.

kfox1111:
Should we file a separate issue for kubeadm to update the check once this is in?

neolit123:
> Should we file a separate issue for kubeadm to update the check once this is in?

the kubeadm kubelet skew will remain limited to N - 1 of the kubeadm version. these tests are making an exception.

neolit123:
this is ready for review / LGTM as the k/kubeadm PR that has the test job configuration is merging:
https://github.com/kubernetes/kubeadm/pull/2396

cynepco3hahue:
I do not see any problems with jobs, the single question who will be responsible to analyze failures(in the case when we will have any)

neolit123:
> I do not see any problems with jobs, the single question who will be responsible to analyze failures(in the case when we will have any)

my proposal:

- if the failures are deployer related (kubeadm) - the kubeadm team will triage and fix.
- if the kubeadm team sees potential problems in the kubelet skew we will ping sig-node, but sig node will see the failures on their mailing list too.
- if kubelet skew failures are not resolved by the sig-node team after a long period of time, the kubeadm team will pause / remove affected jobs.

/hold cancel

cynepco3hahue:
@neolit123 Thanks, works for me.

neolit123:
/retest
(job stuck on status?)

PR #21020, opened 2021-02-25
phaino supports using local gcloud credentials and kubeconfig

chizhg:
1. Add `--use-local-gcloud-credentials` and `--use-local-kubeconfig` to simplify the steps to use local credentials
2. Add `--skip-volume-mounts`,`--extra-volume-mounts`, `--skip-envs`, `--extra-envs`  to allow more customizations on how to run the Prow job
3. Loose the restriction for `checkPrivilege` by removing the user confirmation and print a warning message instead
4. A few other small improvements

/cc @chaodaiG

### DIFF
--- a/prow/cmd/phaino/README.md
+++ b/prow/cmd/phaino/README.md
@@ -33,10 +33,15 @@ volumes that the Prow Job may require.
 * `--print` the command that runs each job without running it
 * `--privileged` jobs are allowed to run instead of rejected
 * `--timeout=10m` controls how long to allow jobs to run before interrupting them
-* `--repo=k8s/test-infra=/go/src/k8s-test-infra` local path where prow job required repos are cloned at, can be passed in repeatedly
+* `--gopath=/go` provides the GOPATH that is used in the container
+* `--skip-volume-mounts=volume1,volume2` includes the unwanted volume mounts that are defined in the job spec
+* `--extra-volume-mounts=/go/src/k8s.io/test-infra=/Users/xyz/k8s-test-infra` includes the extra volume mounts needed for the container. Key is the mount path and value is the local path
+* `--skip-envs=env1,env2` includes the unwanted env vars that are defined in the job spec
+* `--extra-envs=env1=val1,env2=val2` includes the extra env vars needed for the container
+* `--use-local-gcloud-credentials` controls whether to use the same gcloud credentials as local or not
+* `--use-local-kubeconfig` controls whether to use the same kubeconfig as local or not
 
 See `bazel run //prow/cmd/phaino -- --help` for full option list.
-
 
 ### Usage examples
 #### URL example
--- a/prow/cmd/phaino/local.go
+++ b/prow/cmd/phaino/local.go
@@ -23,6 +23,7 @@ import (
 	"go/build"
 	"os"
 	"os/exec"
+	"path"
 	"path/filepath"
 	"strings"
 	"sync"
@@ -30,14 +31,34 @@ import (
 
 	"github.com/sirupsen/logrus"
 	coreapi "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
 
 	prowapi "k8s.io/test-infra/prow/apis/prowjobs/v1"
 )
+
+const (
+	// The well-known env name for gcloud credentials.
+	gcloudCredEnvName = "GOOGLE_APPLICATION_CREDENTIALS"
+	// The path to mount the gcloud default config files to the container.
+	gcloudDefaultConfigMountPath = "/root/.config/gcloud"
+	// The well-known env name for kubeconfig.
+	kubeconfigEnvKey = "KUBECONFIG"
+	// The path to mount the kubectl default config files to the container.
+	kubectlDefaultConfigMountPath = "/root/.kube"
+	// The default GOPATH in the container.
+	defaultGOPATH = "/go"
+)
+
+var baseArgs = []string{"docker", "run", "--rm=true"}
 
 func realPath(p string) (string, error) {
 	if p == "" {
 		return "", errors.New("cannot find repo")
 	}
+	if _, err := os.Stat(p); os.IsNotExist(err) {
+		return "", fmt.Errorf("%q does not exist on local", p)
+	}
+
 	return filepath.Abs(os.ExpandEnv(p))
 }
 
@@ -80,17 +101,14 @@ func pathAlias(r prowapi.Refs) string {
 	if r.PathAlias != "" {
 		return r.PathAlias
 	}
-	path := fmt.Sprintf("%s/%s", r.Org, r.Repo)
+	repoPath := fmt.Sprintf("%s/%s", r.Org, r.Repo)
 	if !strings.HasPrefix(r.Org, "http://") && !strings.HasPrefix(r.Org, "https://") {
-		path = fmt.Sprintf("github.com/%s", path)
+		repoPath = fmt.Sprintf("github.com/%s", repoPath)
 	}
-	return path
+	return repoPath
 }
 
-func readRepo(ctx context.Context, path string, repoPathMap map[string]string, readUserInput func(string, string) (string, error)) (string, error) {
-	if repo, ok := repoPathMap[path]; ok {
-		return repo, nil
-	}
+func readRepo(path string, readUserInput func(string, string) (string, error)) (string, error) {
 	wd, err := workingDir()
 	if err != nil {
 		return "", fmt.Errorf("workingDir: %v", err)
@@ -103,15 +121,16 @@ func readRepo(ctx context.Context, path string, repoPathMap map[string]string, r
 	if err != nil { // Fall back to find repo from local
 		def, err = findRepoFromLocal(wd, path)
 	}
+	if err == nil && def != "" {
+		return realPath(def)
+	}
 	if err != nil {
 		logrus.WithError(err).WithField("repo", path).Warn("could not find repo")
 	}
+
 	out, err := readUserInput(path, def)
 	if err != nil {
 		return "", fmt.Errorf("scan: %v", err)
-	}
-	if out == "" {
-		out = def
 	}
 	return realPath(out)
 }
@@ -143,9 +162,6 @@ func findRepoFromLocal(wd, path string) (string, error) {
 	opwd, err := realPath(wd)
 	if err != nil {
 		return "", fmt.Errorf("wd not found: %v", err)
-	}
-	if strings.HasPrefix(path, "github.com/kubernetes/") {
-		path = strings.Replace(path, "github.com/kubernetes/", "k8s.io/", 1)
 	}
 
 	var old string
@@ -180,9 +196,7 @@ func findRepoFromLocal(wd, path string) (string, error) {
 	return "", errors.New("cannot find repo")
 }
 
-var baseArgs = []string{"docker", "run", "--rm=true"}
-
-func checkPrivilege(ctx context.Context, cont coreapi.Container, allow bool) (bool, error) {
+func checkPrivilege(cont coreapi.Container, pjName string, allow bool) (bool, error) {
 	if cont.SecurityContext == nil {
 		return false, nil
 	}
@@ -192,22 +206,15 @@ func checkPrivilege(ctx context.Context, cont coreapi.Container, allow bool) (bo
 	if !*cont.SecurityContext.Privileged {
 		return false, nil
 	}
-	fmt.Fprint(os.Stderr, "Privileged jobs are unsafe. Remove from local run? [yes]: ")
-	out, err := scanln(ctx)
-	if err != nil {
-		return false, fmt.Errorf("scan: %v", err)
+	if !allow {
+		return false, errors.New("privileged jobs are disallowed")
 	}
-	if out == "no" || out == "n" {
-		if !allow {
-			return false, errors.New("privileged jobs are disallowed")
-		}
-		logrus.Warn("DANGER: privileged containers are unsafe security risks. Please refactor")
-		return true, nil
-	}
-	return false, nil
+
+	logrus.Warningf("WARNING: running privileged job %q can allow nearly all access to the host, please be careful with it", pjName)
+	return true, nil
 }
 
-func convertToLocal(ctx context.Context, log *logrus.Entry, pj prowapi.ProwJob, repoPathMap map[string]string, name string, allowPrivilege bool) ([]string, error) {
+func (opts *options) convertToLocal(ctx context.Context, log *logrus.Entry, pj prowapi.ProwJob, name string) ([]string, error) {
 	log.Info("Converting job into docker run command...")
 	var localArgs []string
 	localArgs = append(localArgs, baseArgs...)
@@ -228,12 +235,8 @@ func convertToLocal(ctx context.Context, log *logrus.Entry, pj prowapi.ProwJob, 
 		localArgs = append(localArgs, "--entrypoint="+entrypoint)
 	}
 
-	for _, env := range container.Env {
-		localArgs = append(localArgs, "-e", env.Name+"="+env.Value)
-	}
-
 	// TODO(fejta): capabilities
-	priv, err := checkPrivilege(ctx, container, allowPrivilege)
+	priv, err := checkPrivilege(container, pj.Spec.Job, opts.priv)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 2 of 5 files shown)


prow/cmd/phaino/main.go
```diff
@@ -57,16 +61,19 @@ func gatherOptions() options {
 	fs.DurationVar(&o.timeout, "timeout", time.Hour, "Maximum duration for each job (0 for unlimited)")
 	fs.DurationVar(&o.totalTimeout, "total-timeout", 0, "Maximum duration for all jobs (0 for unlimited)")
 	fs.DurationVar(&o.grace, "grace", 10*time.Second, "Terminate timed out jobs after this grace period (1s minimum)")
-	fs.StringToStringVar(&o.repoPath, "repo", map[string]string{}, "Local path of repos, can be passed in repeatedly")
+	fs.StringToStringVar(&o.repoPathMap, "repo", map[string]string{}, "Local path of repos, can be passed in repeatedly")
+	fs.BoolVar(&o.mountJobVolumes, "mount-job-volumes", false, "Mount volumes defined in the job spec")
```
chaodaiG, line 65: Should this default to true? Or in another word, this var should be `skip-mount-job-volumes` so that it can be default to be false

prow/cmd/phaino/local.go
```diff
@@ -246,23 +254,66 @@ func convertToLocal(ctx context.Context, log *logrus.Entry, pj prowapi.ProwJob,
...
+				}
+				localArgs = append(localArgs, "-v", arg)
+			}
 		}
-		if vol.EmptyDir != nil {
-			localArgs = append(localArgs, "-v", mount.MountPath)
+	}
+
+	// Setup gcloud credentials
+	if opt.useLocalGcloudCredentials {
```
chaodaiG, line 280: In the case the prow job already defined a volume mount + envvar `GOOGLE_APPLICATION_CREDENTIALS`, the desired behavior would be phaino is smart enough to mount for me and not ask at previous mount step, this also applies to KUBECONFIG. WDYT?

prow/cmd/phaino/local.go
```diff
@@ -246,23 +254,66 @@ func convertToLocal(ctx context.Context, log *logrus.Entry, pj prowapi.ProwJob,
...
-			if err != nil {
-				return nil, fmt.Errorf("bad mount %q: %v", mount.Name, err)
+			// We only want to use the default gcloud credentials if GOOGLE_APPLICATION_CREDENTIALS env var is not explicitly set.
+			defaultGcloudConfigPath := path.Join(os.Getenv("HOME"), ".config/gcloud")
+			if _, err := os.Stat(defaultGcloudConfigPath); !os.IsNotExist(err) {
+				localArgs = append(localArgs, "-v", defaultGcloudConfigPath + ":" + gcloudDefaultConfigMountPath)
+				// Overwrite the gcloud config path, as per https://stackoverflow.com/a/48343135
+				localArgs = append(localArgs, "-e", "CLOUDSDK_CONFIG=" + gcloudDefaultConfigMountPath)
+			}
+		}
```
chaodaiG, line 295: There is a chance both these failed, probably deserve a warning here

prow/cmd/phaino/local.go
```diff
@@ -246,23 +254,66 @@ func convertToLocal(ctx context.Context, log *logrus.Entry, pj prowapi.ProwJob,
...
+		// to the container
+		if gcloudKey != "" {
+			localArgs = append(localArgs, "-e", "GOOGLE_APPLICATION_CREDENTIALS="+gcloudKey)
+			localArgs = append(localArgs, "-v", gcloudKey + ":" + gcloudKey + ":ro")
 		} else {
-			local, err := readMount(ctx, mount)
-			if err != nil {
-				return nil, fmt.Errorf("bad mount %q: %v", mount.Name, err)
+			// We only want to use the default gcloud credentials if GOOGLE_APPLICATION_CREDENTIALS env var is not explicitly set.
+			defaultGcloudConfigPath := path.Join(os.Getenv("HOME"), ".config/gcloud")
```
chaodaiG, line 289: A documentation point to `.config/gcloud` would be nice

chaodaiG (commented):
Thank you for the improvement, this will make use of phaino easier imo. Left a few comments.

chaodaiG:
I think the refactor of method signature also breaks some unit tests, please fix as well.

prow/cmd/phaino/main.go
chizhg: Other flags are all default to false, I think it's better to be consistent.

I just had a second thought on this - since sometimes users might want to only mount some of the volumes but not all, would it be better to define the flag as a `StringSliceVar` and users can choose which volumes to mount?

prow/cmd/phaino/local.go
chizhg: That is a good idea! 

I can change the line to `if opt.useLocalGcloudCredentials || jobSpec.GetEnv("GOOGLE_APPLICATION_CREDENTIALS") != ""`, since if the env var is defined, it assumes there must be a volume mount. Does it make sense to you?

prow/cmd/phaino/main.go
chaodaiG: yeah I would prefer the idea of skipping certain volumes, and by default no one is skipped

prow/cmd/phaino/local.go
chaodaiG: I would want the other way around: don't ask me to mount GCP SA key at previous step because I already said use default credentials.

prow/cmd/phaino/main.go
chizhg: Added the below 4 flags as we discussed offline:
- `--skip-volume-mounts`
- `--extra-volume-mounts`
- `--skip-envs`
- `--extra-envs`

prow/cmd/phaino/local.go
chizhg: Done.

prow/cmd/phaino/local.go
chizhg: Added the warning log.

prow/cmd/phaino/local.go
chizhg: Added the documentation link.

prow/cmd/phaino/local.go
```diff
@@ -34,10 +35,31 @@ import (
...
+	gcloudCredEnvName = "GOOGLE_APPLICATION_CREDENTIALS"
+	// The path to mount the gcloud default config files to the container.
+	gcloudDefaultConfigMountPath = "/root/.config/gcloud"
+	// The well-known env name for kubeconfig.
+	kubeconfigEnvKey = "KUBECONFIG"
+	// The path to mount the kubectl default config files to the container.
+	kubectlDefaultConfigMountPath = "/root/.kube"
+	// The default GOPATH in the container.
+	GOPATH = "/go"
+	// The default GOSRCPATH in the container.
```
chaodaiG, line 49: Would be nice to also document this is from prow job, aka where cloneref clone source code

prow/cmd/phaino/local.go
```diff
@@ -246,7 +250,87 @@ func convertToLocal(ctx context.Context, log *logrus.Entry, pj prowapi.ProwJob,
 		log.Warn("Ignoring resource requirements")
 	}
 
+	// Setup gcloud credentials
```
chaodaiG, line 253: Document why doing this prior to other steps would be helpful

prow/cmd/phaino/local.go
```diff
@@ -246,7 +250,87 @@ func convertToLocal(ctx context.Context, log *logrus.Entry, pj prowapi.ProwJob,
...
+	if opt.useLocalGcloudCredentials || envExists(container.Env, gcloudCredEnvName) {
+		gcloudKey := os.Getenv(gcloudCredEnvName)
+		// If GOOGLE_APPLICATION_CREDENTIALS is not empty, also mount the key file
+		// to the container
+		if gcloudKey != "" {
+			if _, err := os.Stat(gcloudKey); !os.IsNotExist(err) {
+				localArgs = append(localArgs, "-e", gcloudCredEnvName+"="+gcloudKey)
+				localArgs = append(localArgs, "-v", gcloudKey+":"+gcloudKey+":ro")
+			} else {
+				logrus.Warning("The GOOGLE_APPLICATION_CREDENTIALS file does not exist on your local machine, thus gloud authentication won't work in the container")
```
chaodaiG, line 263: I think it would be cleaner to just print a general warning saying "failed using default gcloud credential" after the outer if condition

prow/cmd/phaino/local.go
```diff
@@ -246,7 +250,87 @@ func convertToLocal(ctx context.Context, log *logrus.Entry, pj prowapi.ProwJob,
 		log.Warn("Ignoring resource requirements")
 	}
 
+	// Setup gcloud credentials
+	if opt.useLocalGcloudCredentials || envExists(container.Env, gcloudCredEnvName) {
```
chaodaiG, line 254: Probably should drop `envExists(container.Env, gcloudCredEnvName)`, as I feel this is a bit over smart, would it possible that user does want to attach a different service account? I would expect user explicitly asking for this

prow/cmd/phaino/local.go
```diff
@@ -246,7 +250,87 @@ func convertToLocal(ctx context.Context, log *logrus.Entry, pj prowapi.ProwJob,
...
+				// Overwrite the gcloud config path, as per https://stackoverflow.com/a/48343135
+				localArgs = append(localArgs, "-e", "CLOUDSDK_CONFIG="+gcloudDefaultConfigMountPath)
+			} else {
+				logrus.Warning("The default gcloud credentials does not exist on your local machine, thus gloud authentication won't work in the container")
+			}
+		}
+	}
+
+	// Setup kubeconfig
+	if opt.useLocalKubeconfig || envExists(container.Env, kubeconfigEnvKey) {
```
chaodaiG, line 280: ditto `envExists(container.Env, kubeconfigEnvKey)`

prow/cmd/phaino/local.go
```diff
@@ -246,7 +250,87 @@ func convertToLocal(ctx context.Context, log *logrus.Entry, pj prowapi.ProwJob,
...
+		if env.Name == gcloudCredEnvName || env.Name == kubeconfigEnvKey {
+			continue
+		}
+		if strExists(opt.skippedEnvVars, env.Name) {
+			continue
+		}
+		localArgs = append(localArgs, "-e", env.Name+"="+env.Value)
+	}
+	for envKey, envValue := range opt.extraEnvVars {
+		localArgs = append(localArgs, "-e", envKey+"="+envValue)
```
chaodaiG, line 324: What would happen if this is meant to be an override? would `-e foo=bar -e foo=baz` work?

prow/cmd/phaino/local.go
```diff
@@ -265,6 +349,9 @@ func convertToLocal(ctx context.Context, log *logrus.Entry, pj prowapi.ProwJob,
 			localArgs = append(localArgs, "-v", arg)
 		}
 	}
+	for pathInContainer, localPath := range opt.extraVolumesMounts {
+		localArgs = append(localArgs, "-v", localPath+":"+pathInContainer)
```
chaodaiG, line 353: ditto override

prow/cmd/phaino/local.go
```diff
@@ -278,31 +365,35 @@ func convertToLocal(ctx context.Context, log *logrus.Entry, pj prowapi.ProwJob,
...
 		refs = append(refs, pj.Spec.ExtraRefs...)
 		for _, ref := range refs {
-			path := pathAlias(ref)
-			repo, err := readRepo(ctx, path, repoPathMap, readUserInput)
-			if err != nil {
-				return nil, fmt.Errorf("bad repo(%s): %v", path, err)
+			repoPath := pathAlias(ref)
+			dest := filepath.Join(GOSRCPATH, repoPath)
+			// The repo hasn't been mounted.
+			if _, ok := opt.extraVolumesMounts[filepath.Join(GOSRCPATH, repoPath)]; !ok {
```
chaodaiG, line 379: nit: do we also want to support full path like `/go/src/k8s/test-infra` aside from `k8s/test-infra`?

chaodaiG: Technically user should expect to pass in `/go/src/k8s/test-infra=my/local/clone/path` tbh since this is a volume mount

prow/cmd/phaino/README.md
```diff
@@ -30,10 +30,16 @@ volumes that the Prow Job may require.
 ### Common options
 
 * `--grace=5m` controls how long to wait for interrupted jobs before terminating
-* `--print` the command that runs each job without running it
+* `--dry-run` will only print out the command that runs each job without actually running it
```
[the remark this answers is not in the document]
chizhg: Sure, changed it back.

prow/cmd/phaino/README.md
```diff
@@ -30,10 +30,16 @@ volumes that the Prow Job may require.
...
 * `--privileged` jobs are allowed to run instead of rejected
+* `--mount-job-volumes` controls whether to mount the volumes defined in the job spec or not
```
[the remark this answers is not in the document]
chizhg: I updated the code but forgot the README.
Now fixed.

prow/cmd/phaino/local.go
chizhg: Fixed.

prow/cmd/phaino/local.go
chizhg: Ditto fixed.

prow/cmd/phaino/local.go
chizhg: I have changed to use a map to keep track of all the envs, and overwrite whenever needed, as we chatted offline. So there won't be duplication.

prow/cmd/phaino/local.go
chizhg: Ditto.

prow/cmd/phaino/local.go
chizhg: Hmm it was already supported in this way - users are always expected to provide full path via the `--extra-volume-mounts` flag.

prow/cmd/phaino/local.go
chizhg: I have adjusted the sequence and moved this after the general env vars and volumes setup.

prow/cmd/phaino/local.go
chizhg: I read the code and got some interesting findings:
- The root path used by `clonerefs` is `/home/prow/go` by default - https://github.com/kubernetes/test-infra/blob/master/prow/pod-utils/decorate/podspec.go#L52
- The `GOPATH` is set as `/go` in the `kubekins-e2e` image - https://github.com/kubernetes/test-infra/blob/master/images/kubekins-e2e/Dockerfile#L26

I guess the reason that there are no issues due to the inconsistency is, repos do not have to be under GOPATH as long as they are on go modules. But some projects might still be relying on GOPATH, so I have made it configurable and set its default value to `/go`.

Please let me know if it makes sense to you.

prow/cmd/phaino/local.go
chizhg: I don't quite get what you meant for this comment so I'm leaving the logic unchanged. Let's chat about it later.

chaodaiG:
From high level I'm good with this approach, thank you for the contribution!

Please add unit tests for these new logics.

/cc @fejta

prow/cmd/phaino/README.md
```diff
@@ -33,11 +33,16 @@ volumes that the Prow Job may require.
 * `--print` the command that runs each job without running it
 * `--privileged` jobs are allowed to run instead of rejected
 * `--timeout=10m` controls how long to allow jobs to run before interrupting them
-* `--repo=k8s/test-infra=/go/src/k8s-test-infra` local path where prow job required repos are cloned at, can be passed in repeatedly
+* `--gopath=/go` overwrites the GOPATH to use in the container
```
fejta, line 36: This seems odd to me, why is this helpful/necessary?

prow/cmd/phaino/local.go
```diff
@@ -246,23 +248,108 @@ func convertToLocal(ctx context.Context, log *logrus.Entry, pj prowapi.ProwJob,
...
+		if strExists(opt.skippedEnvVars, env.Name) {
+			continue
+		}
+		envs[env.Name] = env.Value
+	}
+	for name, value := range opt.extraEnvVars {
+		envs[name] = value
+	}
+	// Overwrite GOPATH in the container.
+	envs["GOPATH"] = opt.gopath
```
fejta, line 294: Shouldn't we only do this when it `!= ""` 

I'm also unhappy that we're baking GOPATH so deeply into everything when it is an archaic golang feature.

prow/cmd/phaino/local.go
```diff
@@ -246,23 +248,108 @@ func convertToLocal(ctx context.Context, log *logrus.Entry, pj prowapi.ProwJob,
...
+	if opt.useLocalGcloudCredentials {
+		gcloudKey := os.Getenv(gcloudCredEnvName)
+		// If GOOGLE_APPLICATION_CREDENTIALS is not empty, also mount the key file
+		// to the container
+		if gcloudKey != "" {
+			if _, err := os.Stat(gcloudKey); !os.IsNotExist(err) {
+				volumeMounts[gcloudKey+":ro"] = gcloudKey
+				envs[gcloudCredEnvName] = gcloudKey
+			} else {
+				logrus.Warningf("The GOOGLE_APPLICATION_CREDENTIALS file does not exist on your local machine, thus gcloud authentication won't work in the container")
```
fejta, line 306: idea: How does `gcloud auth application-default login` work? Might be nice to enable this to work eventually (I'll sometimes use this with testgrid development, simulates workload identity)

prow/cmd/phaino/local.go
```diff
@@ -278,33 +365,47 @@ func convertToLocal(ctx context.Context, log *logrus.Entry, pj prowapi.ProwJob,
 		return scanln(ctx)
 	}
 
+	// Mount repositories.
 	if decoration != nil {
+		gosrcpath := filepath.Join(opt.gopath, "src")
```
fejta, line 370: goSrcPath seems more approropiate naming here.

prow/cmd/phaino/local.go
```diff
@@ -316,6 +417,26 @@ func convertToLocal(ctx context.Context, log *logrus.Entry, pj prowapi.ProwJob,
...
+	for _, env := range envs {
+		if env.Name == envName {
+			return true
+		}
+	}
+	return false
+}
+
+// strExists checks if the given string exists in the string list
+func strExists(strs []string, s string) bool {
```
fejta, line 431: consider sets.String for this, not sure if it is worth it here

fejta (approved):
Nice to see these improvements to phaino!
/hold

prow/cmd/phaino/local.go
chizhg: Done. Thanks.

prow/cmd/phaino/local.go
chizhg: Fixed.

prow/cmd/phaino/local.go
chizhg: I reconsidered about this, it looks we don't actually need to overwrite the GOPATH, so I removed this line.

prow/cmd/phaino/README.md
chizhg: I have removed the logic to overwrite the `GOPATH` env var, but it's still sometimes useful to allow users pass the `GOPATH` that will be used in the container. 

The reason we need it is when resolving the refs, the repos need to be cloned under the `GOPATH` (if they depend on it), but we cannot know what the `GOPATH` is without running the container, so we should allow users to tell where to clone the repo to.

chizhg:
/hold cancel
Thanks @chaodaiG and @fejta !

PR #21021, opened 2021-02-25
blockade: allow configuring blockades per branch

nikhita:
Ref: slack discussion - https://kubernetes.slack.com/archives/C78F00H99/p1614202310015700?thread_ts=1613501041.053200&cid=C78F00H99

This commit adds a new field to the blockade config to specify the
branch regexp for which the blockade should apply to.

If no branch regexp is specified, it considers all branches (current
behaviour).

This is needed in cases where we want to block changes to certain files
(e.g. test/conformance/testdata/) on release branches.

cc @hh @dims @justaugustus 
folks on original thread

cc @alejandrox1 @hasheddan @saschagrunert @jeremyrickard 
other sig-release leads

### DIFF
--- a/prow/plugins/config.go
+++ b/prow/plugins/config.go
@@ -264,6 +264,11 @@ type Size struct {
 type Blockade struct {
 	// Repos are either of the form org/repos or just org.
 	Repos []string `json:"repos,omitempty"`
+	// BranchRegexp is the regular expression for branches that the the blockade applies to.
+	// If BranchRegexp is not specified, the blockade applies to all branches by default.
+	// Compiles into BranchRe during config load.
+	BranchRegexp *string        `json:"branchregexp,omitempty"`
+	BranchRe     *regexp.Regexp `json:"-"`
 	// BlockRegexps are regular expressions matching the file paths to block.
 	BlockRegexps []string `json:"blockregexps,omitempty"`
 	// ExceptionRegexps are regular expressions matching the file paths that are exceptions to the BlockRegexps.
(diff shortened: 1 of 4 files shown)


justaugustus:
Ahhh, my favorite plugin gets some new powers!

For SIG Release:
/lgtm
/approve

/hold for SIG Testing review

prow/plugins/config.go
```diff
@@ -1161,6 +1166,17 @@ func compileRegexpsAndDurations(pc *Configuration) error {
 	}
 	pc.CherryPickUnapproved.BranchRe = branchRe
 
+	for _, blockade := range pc.Blockades {
+		if blockade.BranchRegexp == nil {
+			continue
+		}
+		branchRe, err := regexp.Compile(*blockade.BranchRegexp)
```
matthyx, line 1173: maybe we could add to `config_test` a failing regex to ensure we catch it?

matthyx: otherwise LGTM

nikhita: > maybe we could add to config_test a failing regex to ensure we catch it?

@matthyx do you mind if we move this to a follow-up? We currently don't have any tests in `config_test.go` that check for failing regexes so adding a test for `compileRegexpsAndDurations` would also require validating other plugin configs with regex (not just blockade).

matthyx (approved):

spiffxp (approved):
/approve
/lgtm
/hold cancel

PR #21022, opened 2021-02-25
Revert "Upgrade COS from 85 to 89"

pacoxu:
Reverts kubernetes/test-infra#21010

https://prow.k8s.io/?job=pull-kubernetes-node-e2e
It started to fail several hours ago

I suspect it caused by kubernetes/test-infra#21010

### DIFF
--- a/config/jobs/kubernetes/node-problem-detector/node-problem-detector-ci.yaml
+++ b/config/jobs/kubernetes/node-problem-detector/node-problem-detector-ci.yaml
@@ -266,7 +266,7 @@ periodics:
       - name: ZONE
         value: us-central1-a
       - name: IMAGE_FAMILY
-        value: cos-89-lts
+        value: cos-85-lts
       - name: IMAGE_PROJECT
         value: cos-cloud
       - name: BOSKOS_PROJECT_TYPE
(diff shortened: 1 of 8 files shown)


pacoxu:
/cc @dims @bsdnet

I suggest reverting it before investigating it.


I found this: https://cloud.google.com/container-optimized-os/docs/release-notes/#cos-beta-89-16108-0-180
https://cloud.google.com/container-optimized-os/docs/release-notes/#lts_image_families
89 is not lts, but beta here. However, I am not sure about it.

pacoxu:
We may run something like https://github.com/kubernetes/kubernetes/pull/99443/ before merging cos next upgrading. That would be great if it can work in CI in test-infra.

odinuge (approved):
/lgtm

Yes, you are right! 85 is latest lts, while 89 is beta.

/priority critical-urgent

pacoxu:
/assign @dchen1107 
for approval

wojtek-t:
Yes - all node tests are failing with:
```
W0225 09:04:04.762] F0225 09:04:04.762509    5582 run_remote.go:242] Could not retrieve a image based on image regex "" and family "cos-89-lts": found zero images based on regex "" and family "cos-89-lts" in project "cos-cloud"
W0225 09:04:04.763] goroutine 1 [running]:
```

This definitely broke everything.

/lgtm
/approve

wojtek-t:
@dims - can you help with approving?

dims:
/approve
/lgtm

well rolling the dice did not work out :)

PR #21028, opened 2021-02-25
images/bootstrap: clone test-infra in build, fetch in run

spiffxp:
Gated behind `BOOTSTRAP_FETCH_TEST_INFRA` which defaults to false

### DIFF
--- a/images/bootstrap/entrypoint.sh
+++ b/images/bootstrap/entrypoint.sh
@@ -17,8 +17,24 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
-# get test-infra for latest bootstrap etc
-git clone https://github.com/kubernetes/test-infra
+BOOTSTRAP_FETCH_TEST_INFRA=${BOOTSTRAP_FETCH_TEST_INFRA:-"true"}
+
+# clone-or-fetch latest test-infra repo
+if ! [[ -d test-infra/.git ]] || [[ "${BOOTSTRAP_FETCH_TEST_INFRA}" != "true" ]]; then
+  echo "cloning https://github.com/kubernetes/test-infra"
+  rm -rf test-infra
+  git clone https://github.com/kubernetes/test-infra
+else
+  echo "fetching https://github.com/kubernetes/test-infra"
+  pushd test-infra >/dev/null
+  remote=origin
+  git remote set-head ${remote} -a
+  # auto-detect default remote branch name
+  branch=$(<".git/refs/remotes/${remote}/HEAD" sed -e "s|ref: refs/remotes/${remote}/||")
+  git fetch "${remote}" "${branch}" && git reset --hard "${remote}/${branch}"
+  # but don't bother setting local branch name yet; scenario-based jobs still assume 'master'
+  popd >/dev/null
+fi
 
 BOOTSTRAP_UPLOAD_BUCKET_PATH=${BOOTSTRAP_UPLOAD_BUCKET_PATH:-"gs://kubernetes-jenkins/logs"}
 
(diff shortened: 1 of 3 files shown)


spiffxp:
/hold
to trial via https://github.com/kubernetes/test-infra/pull/21029

spiffxp:
/hold cancel
```
fetching https://github.com/kubernetes/test-infra
origin/HEAD set to master
From https://github.com/kubernetes/test-infra
 * branch                  master     -> FETCH_HEAD
   6252cf69ff..cce6effe4a  master     -> origin/master
HEAD is now at cce6effe4a Merge pull request #21021 from nikhita/blockade-per-branch
Docker in Docker enabled, initializing...
```
Yeah it works, the time saved doesn't show up in logs because, of course, bootstrap jobs aren't decorated, so stdout isn't getting put into `build-log.txt`.

But I'm pretty sure we'll see clones of this repo go down once this rolls out

spiffxp:
This reverts https://github.com/kubernetes/test-infra/pull/21029 now that I've confirmed this works

images/bootstrap/entrypoint.sh
```diff
@@ -17,8 +17,24 @@ set -o errexit
...
 
-# get test-infra for latest bootstrap etc
-git clone https://github.com/kubernetes/test-infra
+BOOTSTRAP_FETCH_TEST_INFRA=${BOOTSTRAP_FETCH_TEST_INFRA:-"true"}
+
+# clone-or-fetch latest test-infra repo
+if ! [[ -d test-infra/.git ]] || [[ "${BOOTSTRAP_FETCH_TEST_INFRA}" != "true" ]]; then
+  echo "cloning https://github.com/kubernetes/test-infra"
+  rm -rf test-infra
+  git clone https://github.com/kubernetes/test-infra
```
hasheddan, line 26: @spiffxp it seems unlikely we would hit this case, but in the event that we want latest and local repo does not exist (so we must clone it), can we just do a shallow clone?

images/bootstrap/entrypoint.sh
```diff
@@ -17,8 +17,24 @@ set -o errexit
...
+else
+  echo "fetching https://github.com/kubernetes/test-infra"
```
sladyn98, line 28: Suggestion: This URL is used multiple times could it be put in a variable

sladyn98 (approved):
Apart from that looks good to me.

dims:
Whoops! One question 

/hold

images/bootstrap/entrypoint.sh
spiffxp: This is here for fallback to previous behavior in the event we see something weird.  Given that bootstrap is deprecated I'd rather not fallback to different clever behavior.

We could consider a followup where a shallow clone is done by default instead of a fetch.  I'm slightly wary of shallow clones since at our scale they can put a hurt on github if there is followup fetching (this is really more of a kubernetes/kubernetes problem).

We could also consider a plan where there is _no_ fetching done at runtime.  What's in the image is what you get.  This is more in line with how `kubetest` changes work today.

It would also be helpful to get stats out of this.  Right now bootstrap seems to be swallowing my attempts to prepend to `build-log.txt`.  Another idea might be to pre-create whatever artifacts dir it uses and dump something in there.

But I consider these all followup.  I'd like to get this included in the kubekins that gets included in today's autobump

hasheddan: SGTM

spiffxp:
https://github.com/kubernetes/test-infra/pull/21041 - job configs were updated to use a kubekins-e2e images with this change as of 2021-02-26 14:20:00 PT

spiffxp:
From https://github.com/kubernetes/test-infra/graphs/traffic
Screen Shot 2021-03-01 at 12 23 22 PM

So that's the result of updating our CI.  Will be curious to see if there are any downstream project that pick this change up.

PR #21030, opened 2021-02-25
Enable using built binaries on the kubetest2 gce conformance job.

amwat:
follow up to : https://github.com/kubernetes-sigs/kubetest2/pull/95

xref: https://github.com/kubernetes/enhancements/issues/2464

/cc @BenTheElder @spiffxp

### DIFF
--- a/config/jobs/kubernetes/sig-cloud-provider/gcp/gce-conformance.yaml
+++ b/config/jobs/kubernetes/sig-cloud-provider/gcp/gce-conformance.yaml
@@ -92,4 +92,5 @@ periodics:
           --down \;
           --test=ginkgo \;
           -- \;
-          --focus-regex='\[Conformance\]'
+          --focus-regex='\[Conformance\]';
+          --use-built-binaries


config/jobs/kubernetes/sig-cloud-provider/gcp/gce-conformance.yaml
```diff
@@ -92,4 +92,5 @@ periodics:
           --down \;
           --test=ginkgo \;
           -- \;
-          --focus-regex='\[Conformance\]'
+          --focus-regex='\[Conformance\]';
+          --use-built-binaries
```
BenTheElder, line 96: IMHO once this is validated this should actually be the default unless test binaries have not been built.

BenTheElder (approved):
lgtm

config/jobs/kubernetes/sig-cloud-provider/gcp/gce-conformance.yaml
amwat: +1, yeah i didn't change it in the initial PR since that would break the current default.

PR #21033, opened 2021-02-25
Add support to autobumper for upstream RefConfigFile with multiple image types

mpherman2:
When bumping to upstream images, the autobumper assumed the refConfigFile contained only images with the given prefix. This change makes the parser more robust and checks that the image it is grabbing from refConfigFile has the same prefix. This allows us to use a refConfigFile that contains different image prefixes.

### DIFF
--- a/prow/cmd/generic-autobumper/bumper/bumper.go
+++ b/prow/cmd/generic-autobumper/bumper/bumper.go
@@ -60,7 +60,7 @@ const (
 
 var (
 	tagRegexp    = regexp.MustCompile("v[0-9]{8}-[a-f0-9]{6,9}")
-	imageMatcher = regexp.MustCompile(`(?s)^.+image:.+:(v[a-zA-Z0-9_.-]+)`)
+	imageMatcher = regexp.MustCompile(`(?s)^.+image:(.+):(v[a-zA-Z0-9_.-]+)`)
 )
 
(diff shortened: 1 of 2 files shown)


prow/cmd/generic-autobumper/bumper/bumper.go
```diff
@@ -534,7 +534,20 @@ func upstreamConfigVersions(upstreamVersionType string, o *Options, parse func(u
 	return versions, nil
 }
 
-func parseUpstreamImageVersion(upstreamAddress string) (string, error) {
+func findExactMatch(body, prefix string) (string, error) {
+	for _, line := range strings.Split(strings.TrimSuffix(body, "\n"), "\n") {
+		if strings.Contains(line, prefix) {
+			res := imageMatcher.FindStringSubmatch(string(line))
```
chaodaiG, line 540: Instead of string splitting, I would prefer `imageMatcher.FindAllStringSubmatch` and filter through the matched results to find the first one containing `prefix`, this feels more accurate to me

mpherman2: Discussed offline, but FindAllStringSubmatch would not work without major changes to the regex. We added another capture group to the regex to make the string splitting method more accurate

chaodaiG (approved):

PR #21039, opened 2021-02-26
Flip Kubernetes jobs to quick build

BenTheElder:
This PR looks indimidating, but it's not quite so:

- A lot of these jobs don't even run, they're manually invoked
- Only jobs targeting github.com/kubernetes/kubernetes on the master branch are updated
- The mechanical approach to update job config follows simple rules:
  - remove `preset-bazel.*` labels
  - ensure a `preset-dind-enabled: "true"` label
  - ensure securityContext for running dind
  - s/--build=bazel/--build=quick/ for kubetest jobs
  - drop `BUILD_TYPE` env from kind jobs (which default to quick otherwise)
- Each commit targets a specific subset for easier review

The remainder of the diff is cleaning up the conformance image test script, ~and dropping some unused 1.13 EKS jobs and kubernetes rbe jobs.~ (moved out the latter).

We've already been testing jobs using the quick build, as expected it's fine.

### DIFF
--- a/config/jobs/kubernetes/sig-api-machinery/sig-api-machinery-config.yaml
+++ b/config/jobs/kubernetes/sig-api-machinery/sig-api-machinery-config.yaml
@@ -89,8 +89,7 @@ presubmits:
     labels:
       preset-service-account: "true"
       preset-k8s-ssh: "true"
-      preset-bazel-scratch-dir: "true"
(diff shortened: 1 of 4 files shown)


BenTheElder:
breaking out portions of this.

BenTheElder:
first one: https://github.com/kubernetes/test-infra/pull/21045 (stale EKS cleanup)

PR #21047, opened 2021-02-26
move sig testing Kubernetes presubmits to quick

BenTheElder:
build
/cc @spiffxp @dims 
local-up-cluster and conformance image jobs. simplifies the conformance-image script.
broken out of https://github.com/kubernetes/test-infra/pull/21039

### DIFF
--- a/experiment/kind-conformance-image-e2e.sh
+++ b/experiment/kind-conformance-image-e2e.sh
@@ -47,21 +47,12 @@ install_kind() {
 
 # build kubernetes / node image, e2e binaries
 build() {
-    # possibly enable bazel build caching before building kubernetes
-    BAZEL_REMOTE_CACHE_ENABLED=${BAZEL_REMOTE_CACHE_ENABLED:-false}
-    if [[ "${BAZEL_REMOTE_CACHE_ENABLED}" == "true" ]]; then
-        # run the script in the kubekins image, do not fail if it fails
-        /usr/local/bin/create_bazel_cache_rcs.sh || true
-    fi
-
     # build the node image w/ kubernetes
-    kind build node-image --type=bazel --kube-root="${PWD}"
+    kind build node-image --kube-root="${PWD}"
 
-    # try to make sure the kubectl we built is in PATH
+    # ensure kubectl is in path
     local maybe_kubectl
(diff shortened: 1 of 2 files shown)


experiment/kind-conformance-image-e2e.sh
```diff
@@ -47,25 +47,8 @@ install_kind() {
...
-    BAZEL_REMOTE_CACHE_ENABLED=${BAZEL_REMOTE_CACHE_ENABLED:-false}
-    if [[ "${BAZEL_REMOTE_CACHE_ENABLED}" == "true" ]]; then
-        # run the script in the kubekins image, do not fail if it fails
-        /usr/local/bin/create_bazel_cache_rcs.sh || true
-    fi
-
     # build the node image w/ kubernetes
-    kind build node-image --type=bazel --kube-root="${PWD}"
-
-    # try to make sure the kubectl we built is in PATH
```
BenTheElder, line 60: we may or may not need to do something like this again (probably not)
this job doesn't really depend on the local kubectl except to install the conformance pod anyhow though.

this snippet is based on the other kind jobs which actually do need to use the just-built kubectl for tests.

BenTheElder: implemented this, tested this part locally.

experiment/kind-conformance-image-e2e.sh
```diff
@@ -47,21 +47,12 @@ install_kind() {
...
-    kind build node-image --type=bazel --kube-root="${PWD}"
+    kind build node-image --kube-root="${PWD}"
 
-    # try to make sure the kubectl we built is in PATH
+    # ensure kubectl is in path
     local maybe_kubectl
-    maybe_kubectl="$(bazel aquery 'mnemonic("GoLink", //cmd/kubectl)' \
-      | grep '^  Outputs: \[.*\]$' \
-      | sed 's/^  Outputs: \[\(.*\)\]$/\1/')"
+    maybe_kubectl="$(find "${PWD}/_output" -name "kubectl" -type f)"
```
BenTheElder, line 55: this is the previous implementation we already had, just `s/bazel-bin/_output` this works fine for our purposes.

PR #21049, opened 2021-02-26
update Kubernetes node presubmits

BenTheElder:
- switch to quick build
- drop stray presets

/cc @dims @yujuhong

### DIFF
--- a/config/jobs/kubernetes/sig-node/sig-node-presubmit.yaml
+++ b/config/jobs/kubernetes/sig-node/sig-node-presubmit.yaml
@@ -62,8 +62,6 @@ presubmits:
     max_concurrency: 12
     labels:
       preset-service-account: "true"
-      preset-bazel-scratch-dir: "true"
-      preset-bazel-remote-cache-enabled: "true"
       preset-dind-enabled: "true"
       preset-kind-volume-mounts: "true"
       preset-k8s-ssh: "true"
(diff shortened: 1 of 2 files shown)


BenTheElder:
broken out of https://github.com/kubernetes/test-infra/pull/21039

config/jobs/kubernetes/sig-node/sig-node-presubmit.yaml
```diff
@@ -62,8 +62,6 @@ presubmits:
     max_concurrency: 12
     labels:
       preset-service-account: "true"
-      preset-bazel-scratch-dir: "true"
```
dims, line 65: don't need dind preset here? (like the other two)

BenTheElder: it was already here on line 67/65 below

config/jobs/kubernetes/sig-node/sig-node-presubmit.yaml
```diff
@@ -62,8 +62,6 @@ presubmits:
...
-      preset-bazel-remote-cache-enabled: "true"
       preset-dind-enabled: "true"
```
BenTheElder, line 65: @dims here ^^

dims: whoops!

PR #21052, opened 2021-02-26
Add configurable labels to autobumper

mpherman2:
The autobumper config now has a new field "labels" which you can use to add labels to the autobump PR

### DIFF
--- a/experiment/autobumper/bumper/bumper.go
+++ b/experiment/autobumper/bumper/bumper.go
@@ -116,6 +116,8 @@ type Options struct {
 	RemoteName string `yaml:"remoteName"`
 	// The name of the branch that will be used when creating the pull request. If unset, defaults to "autobump".
 	HeadBranchName string `yaml:"headBranchName"`
+	// Optional list of labels to add to the bump PR
+	Labels []string `yaml:"labels"`
 	// List of prefixes that the autobumped is looking for, and other information needed to bump them. Must have at least 1 prefix.
 	Prefixes []Prefix `yaml:"prefixes"`
 }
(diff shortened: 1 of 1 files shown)


experiment/autobumper/bumper/bumper.go
```diff
@@ -376,8 +378,11 @@ func (w HideSecretsWriter) Write(content []byte) (int, error) {
 // "images" contains the tag replacements that have been made which is returned from "updateReferences([]string{"."}, extraFiles)"
 // "images" and "extraLineInPRBody" are used to generate commit summary and body of the PR
 func UpdatePR(gc github.Client, org, repo string, images map[string]string, extraLineInPRBody, login, baseBranch, headBranch string, allowMods bool, prefixes []Prefix, versions map[string][]string) error {
```
chaodaiG, line 380: Is this still used?

experiment/autobumper/bumper/bumper.go
```diff
@@ -116,6 +116,8 @@ type Options struct {
 	RemoteName string `yaml:"remoteName"`
 	// The name of the branch that will be used when creating the pull request. If unset, defaults to "autobump".
 	HeadBranchName string `yaml:"headBranchName"`
+	// Optional list of labels to add to the bump PR
+	Labels []string `yaml:"labels"`
```
chaodaiG, line 120: We probably have discussed about this before 😅 , is `omitempty` needed?

chaodaiG (commented):
Awesome!

experiment/autobumper/bumper/bumper.go
mpherman2: I don't believe it is needed. The default value is []. I tested running without labels in the config and it worked as intended. I think omitempty helps when going from a golang struct -> json, but it doesn't seem to do anything when going from json -> struct.

experiment/autobumper/bumper/bumper.go
mpherman2: Not in the autobumper, and it is redundant; however it is exported, and I wanted to keep the change backwards compatible.
/assign @alvaroaleman 
Can we get rid of this?

chaodaiG: Good thoughts!

chaodaiG (approved):
Approve since this is already backward compatible. 

/hold
If you want to clean it up in current PR let's wait for @alvaroaleman to confirm. Otherwise please document the purpose why keeping this function, would be even better log a warning for anyone who invoke this function with a deprecation date

experiment/autobumper/bumper/bumper.go
alvaroaleman: Yeah, we are still using this

PR #21054, opened 2021-02-26
Prow job labeled with prow version

chaodaiG:
As proposed in https://github.com/kubernetes/enhancements/pull/2540, k8s prow will be bumped more frequently than once per work day, adding this label to the job so that it could help associate failures with specific prow versions

### DIFF
--- a/prow/kube/prowjob.go
+++ b/prow/kube/prowjob.go
@@ -41,6 +41,9 @@ const (
 	// job names can be arbitrarily long, this is added as
 	// an annotation instead of a label.
 	ProwJobAnnotation = "prow.k8s.io/job"
+	// ProwVersionLabel is added in resources created by prow and
+	// carries the version of prow that decorated this job.
+	ProwVersionLabel = "prow.k8s.io/version"
 	// OrgLabel is added in resources created by prow and
 	// carries the org associated with the job, eg kubernetes-sigs.
 	OrgLabel = "prow.k8s.io/refs.org"
--- a/prow/kube/prowjob.go
+++ b/prow/kube/prowjob.go
@@ -41,9 +41,9 @@ const (
 	// job names can be arbitrarily long, this is added as
 	// an annotation instead of a label.
 	ProwJobAnnotation = "prow.k8s.io/job"
-	// ProwVersionLabel is added in resources created by prow and
+	// PlankVersionLabel is added in resources created by prow and
 	// carries the version of prow that decorated this job.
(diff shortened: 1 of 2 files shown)


fejta (approved):
/hold

prow/kube/prowjob.go
```diff
@@ -41,6 +41,9 @@ const (
 	// job names can be arbitrarily long, this is added as
 	// an annotation instead of a label.
 	ProwJobAnnotation = "prow.k8s.io/job"
+	// ProwVersionLabel is added in resources created by prow and
+	// carries the version of prow that decorated this job.
+	ProwVersionLabel = "prow.k8s.io/version"
```
spiffxp, line 46: We deploy all prow components as a bunch of micro services in lockstep today. It may be worth considering the versions of the components themselves though.

In terms of components that would be relevant to why pods landed on a cluster and how they ended up in their current form.  Ideally we could have:
- prow.k8s.io/hook-version is responsible for triggering a job in response to GitHub events
- prow.k8s.io/plank-version is responsible for making a pod out of a prowjob
- prow is using a specific config when creating this pod (sha of the repo from which the config was deployed? hash of configmap contents? revision of the configmap resource?), or maybe specific component and job configs

spiffxp (approved):
/lgtm
/hold
If you want to change the label name at all. Otherwise we'll save that for future improvements

prow/kube/prowjob.go
alvaroaleman: A better approach to this would be leverage the `managedFields`. We would need to figure out how to set a correct fieldmanager (xref https://github.com/kubernetes-sigs/controller-runtime/issues/1215), then something like [`kubect-blame`](https://github.com/knight42/kubectl-blame) can be used to figure out what component in what version set which fields. Requires kube 1.18 though.

alvaroaleman: Although hm, they are intended as user identifier in server side apply, which means if we put a version inside that, we would prevent ourselves from starting to use server side apply, if we ever wanted.

spiffxp: The nice thing about labels in my experience is they automatically get attached to metrics within google cloud monitoring, and I'm pretty sure [kubernetes/kube-state-metrics](https://github.com/kubernetes/kube-state-metrics) does the same

chaodaiG: Good point and done

spiffxp (approved):
/approve
/lgtm

PR #21056, opened 2021-02-26
update kind jobs

BenTheElder:
update kubernetes/kubernetes and kubernetes-sigs/kind jobs to use quick build

/cc @amwat @spiffxp

### DIFF
--- a/config/jobs/kubernetes/sig-arch/conformance-audit.yaml
+++ b/config/jobs/kubernetes/sig-arch/conformance-audit.yaml
@@ -6,8 +6,6 @@ periodics:
   name: ci-audit-kind-conformance
   labels:
     preset-service-account: "true"
-    preset-bazel-scratch-dir: "true"
-    preset-bazel-remote-cache-enabled: "true"
     preset-dind-enabled: "true"
     preset-kind-volume-mounts: "true"
   decorate: true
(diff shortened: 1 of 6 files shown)


config/jobs/kubernetes/sig-arch/conformance-audit.yaml
```diff
@@ -34,7 +32,7 @@ periodics:
       - name: GINKGO_TOLERATE_FLAKES
         value: "n"
       - name: BUILD_TYPE
-        value: bazel
+        value: docker
```
BenTheElder, line 35: this one is not just dropped because it's using a custom script. cc @hh

amwat (commented):
/lgtm

config/jobs/kubernetes/sig-arch/conformance-audit.yaml
hh: @zachmandeville can you ensure our jobs are still happy?

zachmandeville: this job is happy! https://k8s-testgrid.appspot.com/sig-arch-conformance#kind-conformance-audit

BenTheElder: and importantly building with make as intended, we are dropping bazel in k/k (notice going out tomorrow)

Issue #21053, opened 2021-02-26
Config presets validation does not fail on unknown fields


chemikadze:
**What happened**:

Attempt to add `envFrom` (which turned out to be not supported) to config does not produce config parsing errors.

**What you expected to happen**:

Config to be detected as invalid.

**How to reproduce it (as minimally and precisely as possible)**:

Preset snippet:

```
- labels:
    preset-example: "true"
  envFrom:
    - secretRef:
        name: secret-example
```

**Please provide links to example occurrences, if any**:

Private project, N/A.

**Anything else we need to know?**:

BenTheElder:
https://github.com/kubernetes/test-infra/pull/10812 implemented what you want I think, but was never approved and I got busy with other things.

BenTheElder:
actually this seems like a different bug, ref https://github.com/kubernetes/test-infra/issues/11879#issuecomment-514427424

BenTheElder:
I tried https://github.com/kubernetes/test-infra/pull/21075
Perhaps someone can rebase it and try again.

PR #12450, opened 2019-05-02
Add a yamllint presubmit to test-infra

alvaroaleman:
This PR adds a yamllint presubmit to test-infra and fixes up the issues yamllint found.

I used a Docker images from ours (quay.io/kubermatic/yamllint:0.1) because I don't know the process of adding a new docker-image for test-infra usage. Its Dockerfile looks like this:

```Dockerfile
FROM python:3.7.1-alpine3.8

RUN pip3 install yamllint==1.13.0
```

If someone tells me how image publishing works I am happy to add that in this PR too.

Some of the default linters that I didn't disable can be argued about, specifically:
* trailing spaces
* trailing blank lines
* no new line character at the end of file

But since the amount of issues that had to be fixed because of them was still limited and it increases readability I left them enabled. No strong opinion here thought.

The linters that are enabled now are:

```
rules:
  braces: enable
  brackets: enable
  colons: enable
  commas: enable
  comments-indentation:
    level: warning
  empty-lines: enable
  empty-values: enable
  hyphens: enable
  key-duplicates: enable
  new-line-at-end-of-file: enable
  new-lines: enable
  octal-values: enable
  quoted-strings: disable
  trailing-spaces: enable
  truthy:
    level: warning
```

More details about the meaning of the rules can be found [in the yamllint docs](https://yamllint.readthedocs.io/en/stable/rules.html)

/assign @fejta

PR #21073, opened 2021-02-28
use UnmarshalStrict for prow config [redux]

BenTheElder:
rebased https://github.com/kubernetes/test-infra/pull/10812 based on https://github.com/kubernetes/test-infra/pull/12450#issuecomment-489026169, https://github.com/kubernetes/test-infra/issues/21053

### DIFF
--- a/prow/ANNOUNCEMENTS.md
+++ b/prow/ANNOUNCEMENTS.md
@@ -129,6 +129,8 @@ Note: versions specified in these announcements may not include bug fixes made
 in more recent versions so it is recommended that the most recent versions are
 used when updating deployments.
 
+ - *February 28th, 2021* config loading now uses `UnmarshalStrict` to ensure that all
+   keys are for valid fields. This may break deployments using invalid prow configs.
  - *January 24th, 2021* Planks Pod pending and Pod scheduling timeout defaults where changed from 24h each to the more reasonable 10 minutes/5 minutes, respectively.
  - *January 1, 2021* Support for `whitelist` and `branch_whitelist` fields in Slack merge warning configuration is discontinued. You can use `exempt_users` and `exempt_branches` fields instead.
  - *November 24, 2020* The `requiresig` plugin has been removed in favor of the `require-matching-label` plugin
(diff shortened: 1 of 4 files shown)


BenTheElder:
AFAICT https://github.com/kubernetes/test-infra/pull/10812 was functional, and discussion ultimately trended towards accepting it, but it was never approved so I eventually closed it (busy doing other things?)

We may need to add this to more places now, I haven't checked yet.

BenTheElder:
1. This change is potentially quite controversial still so:

/approve cancel

- Not trying to use my root approval permission. I will let someone in `prow/OWNERS` approve this one.

2. This caught a minor wrong configuration (`always_run: true` on a postsubmit), but only a minor one.
- See: https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/test-infra/21073/pull-test-infra-bazel/1366103861428752384
- Fix in: https://github.com/kubernetes/test-infra/pull/21074

3. We can get quite a lot of benefit from strict-checking configuration in presubmit at least, and we have since better decoupled `config/` and `prow/`, filed an alternate PR that allows us to start trying this for `config/` without doing it in prow components at runtime (yet?). We can discuss where else we should do this if anywhere.
- Filed alternative PR: https://github.com/kubernetes/test-infra/pull/21075

stevekuznetsov:
@petr-muller suggested this and I think he's right - tools that operate on configuration to provide users feedback should use a strict unmarshal. Tools that run in production should not.

So, in this case, I'd love to see `checkconfig` run with a strict unmarshal (does it not?) and no control-plane services to do so. We recently removed a field in one of our configs (not a Prow config) and we had a controller crash. What's the utility in that? Or the utility in Prow not loading the new config indefinitely? Ideally the code doing this load will run in a presubmit on this repo and give value by disallowing merges, but not in production.

stevekuznetsov:
Ah, that's https://github.com/kubernetes/test-infra/pull/21075 - perfect <3

BenTheElder:
The controller wouldn't crash here it just wouldn't load the invalid config (config agent will retain previous config and log an error). I'm not convinced loading invalid config at runtime is a good idea, but to best handle that we really should only be adding fields unless we revision the config per component config which would be a safer pattern. That's a bigger change though.

Anyhow for this project my primary concern is validating at config time. Forgot to close this in favor of the other, which is in progress.

PR #21075, opened 2021-02-28
optionally strict parse prow config

BenTheElder:
... initially enable it for `config/test/jobs` and as a more powerful `--warning=unknown-fields` in `prow/cmd/checkconfig`.

alternative to https://github.com/kubernetes/test-infra/pull/21073

### DIFF
--- a/prow/cmd/checkconfig/main.go
+++ b/prow/cmd/checkconfig/main.go
@@ -330,20 +330,7 @@ func validate(o options) error {
 		}
 	}
 	if o.warningEnabled(unknownFieldsWarning) {
-		cfgBytes, err := ioutil.ReadFile(o.configPath)
-		if err != nil {
-			return fmt.Errorf("error reading Prow config for validation: %w", err)
-		}
-		if err := validateUnknownFields(&config.Config{}, cfgBytes, o.configPath); err != nil {
-			errs = append(errs, err)
-		}
-	}
-	if pcfg != nil && o.warningEnabled(unknownFieldsWarning) {
-		pcfgBytes, err := ioutil.ReadFile(o.pluginConfig)
-		if err != nil {
-			return fmt.Errorf("error reading Prow plugin config for validation: %w", err)
-		}
-		if err := validateUnknownFields(&plugins.Configuration{}, pcfgBytes, o.pluginConfig); err != nil {
+		if _, err := config.LoadStrict(o.configPath, o.jobConfigPath); err != nil {
 			errs = append(errs, err)
 		}
 	}
@@ -461,14 +448,6 @@ func validateURLs(c config.ProwConfig) error {
 	}
 
 	return utilerrors.NewAggregate(validationErrs)
-}
-
-func validateUnknownFields(cfg interface{}, cfgBytes []byte, filePath string) error {
-	err := yaml.Unmarshal(cfgBytes, &cfg, yaml.DisallowUnknownFields)
-	if err != nil {
-		return fmt.Errorf("unknown fields or bad config in %s: %v", filePath, err)
-	}
-	return nil
 }
 
 func validateJobRequirements(c config.JobConfig) error {
--- a/prow/config/config_test.go
+++ b/prow/config/config_test.go
@@ -344,10 +344,11 @@ func TestGetGCSBrowserPrefix(t *testing.T) {
 func TestDecorationRawYaml(t *testing.T) {
 	t.Parallel()
 	var testCases = []struct {
-		name        string
-		expectError bool
-		rawConfig   string
-		expected    *prowapi.DecorationConfig
+		name              string
+		expectError       bool
+		expectStrictError bool
+		rawConfig         string
+		expected          *prowapi.DecorationConfig
 	}{
 		{
 			name:        "no default",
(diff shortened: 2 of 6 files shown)


BenTheElder:
This failure is expected and means the PR works:

https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/test-infra/21075/pull-test-infra-bazel/1366128045840142336


> INFO: From Testing //config/tests/jobs:go_default_test:
==================== Test output for //config/tests/jobs:go_default_test:
Could not load config: error unmarshaling ../../jobs/image-pushing/k8s-staging-multitenancy.yaml: error unmarshaling JSON: while decoding JSON: json: unknown field "always_run"================================================================================

The fix for the bad config is in https://github.com/kubernetes/test-infra/pull/21074

cjwagner:
This would be great to have! I'm thinking it might make more sense to colocate this logic with the existing unknown fields checks in the checkconfig tool: https://github.com/kubernetes/test-infra/blob/a2740c4267c5352705590340f2406834fec59207/prow/cmd/checkconfig/main.go#L332-L339

The current `unknown-fields` check only validates config.yaml. I think it would make sense to add another check like `unknown-fields-jobs` and enable it here: https://github.com/kubernetes/test-infra/blob/d74af0f356ee386c9c1d33b3fd1c6368900d83d6/config/jobs/kubernetes/test-infra/test-infra-presubmits.yaml#L90-L97

I think the benefit of including this in the checkconfig tool is that it makes it easier to reuse this validation in other repos. WDYT?

BenTheElder:
> I think the benefit of including this in the checkconfig tool is that it makes it easier to reuse this validation in other repos. WDYT?

Well we still need a call like `LoadStrict` somewhere to enable this, I don't think we lose anything by also strict parsing in `config/jobs/test`.

Unfortunately it's going to be more slightly annoying to add this to checkconfig because it uses a config agent instead of one-shot loading for some reason.

BenTheElder:
Looking at optionally enabling in the agent.

BenTheElder:
Oh I see we just parse *again* in `validateUnknownFields.

cjwagner:
> Unfortunately it's going to be more slightly annoying to add this to checkconfig because it uses a config agent instead of one-shot loading for some reason.

I doubt that was an intentional design decision, I cant imagine why we'd need a config agent rather than a single load here. Feel free to change if that is helpful.


> Oh I see we just parse _again_ in `validateUnknownFields`.

Yes. It is easy for config.yaml and plugin.yaml because there is a single file which is not the case for job config. We could copy the recursive file lookup from the normal config loading code or make a `LoadStrict` function like you were doing. The later might be a bit cleaner? Up to you though I think either is fine.

prow/cmd/checkconfig/main.go
```diff
@@ -330,20 +330,7 @@ func validate(o options) error {
...
-			errs = append(errs, err)
-		}
-	}
-	if pcfg != nil && o.warningEnabled(unknownFieldsWarning) {
-		pcfgBytes, err := ioutil.ReadFile(o.pluginConfig)
-		if err != nil {
-			return fmt.Errorf("error reading Prow plugin config for validation: %w", err)
-		}
-		if err := validateUnknownFields(&plugins.Configuration{}, pcfgBytes, o.pluginConfig); err != nil {
+		if _, err := config.LoadStrict(o.configPath, o.jobConfigPath); err != nil {
```
BenTheElder, line 333: The current validateUnknownFields was all fields except job fields, that seems like more of an oversight / artificial limitation / easier to implement rather than an intent to allow individually warning on unknown fields in only one of (plugins, prow, jobs).

just using `LoadStrict` here dedupes the logic if we ever decide to strict load in other components and covers all 3

I also have a diff stashed to enable it optionally in the config agent, but I think that's possibly another PR / probably this tool should switch to one-shot `Load` anyhow, but I want to keep the diff for this PR down.

BenTheElder:
/retest
[config fix landed]

BenTheElder:
there are more of these bad job configs ... lots of them.

I'm going to fix some of them but some of them are in config generators I haven't touched before and I'm running out of time for the moment.

BenTheElder:
fixing _some_ of them in https://github.com/kubernetes/test-infra/pull/21080
it's already +330 / −342

prow/config/config_test.go
```diff
@@ -344,10 +344,11 @@ func TestGetGCSBrowserPrefix(t *testing.T) {
 func TestDecorationRawYaml(t *testing.T) {
```
cjwagner, line 344: I think these test cases would make more sense in `TestValidConfigLoading`. That is also set up to test loading from multiple files which we need to test that strict validation works for job config.

prow/cmd/checkconfig/main.go
BenTheElder: is it not actually a warning then?

BenTheElder: ah I see, there's only an option to convert all warnings to errors, there's no distinction between warning and error.
makes sense then.

BenTheElder:
I'll come back to this soon with the requested changes, I need to take care of a couple other things today/tomorrow first.

BenTheElder:
There hasn't been much interest in this, I haven't been able to get reviews on PRs fixing the errors. Will potentially come back to at some point but pretty low priority for me ...

PR #21085, opened 2021-03-01
kubernetes: update required presubmits

BenTheElder:
disable bazel jobs per KEP https://github.com/kubernetes/enhancements/issues/2420

/cc @dims @liggitt @spiffxp


liggitt:
both of these are only for the master branch; release branches are still covered by individual configs

https://github.com/kubernetes/enhancements/issues/2420#issuecomment-785663043 has the analysis showing all the tests being run by these jobs are covered by other required jobs

BenTheElder:
updated to only disable required, will disable always_run in a follow up once branchprotector catches up.
it's taking nearly two hours to complete a run these days and we don't want a required job to stop reporting.

PR #21086, opened 2021-03-01
update always_run kubernetes jobs

BenTheElder:
disable the bazel presubmits per https://github.com/kubernetes/enhancements/issues/2420
/hold
need to wait for the protected branch state to catchup I think
/cc @spiffxp @dims @liggitt


liggitt:
/lgtm
leaving hold in place until protected branches are adjusted

BenTheElder:
replaced by https://github.com/kubernetes/test-infra/pull/21088

PR #21087, opened 2021-03-01
Update pj-on-kind images

chases2:
Updates docker images as part of running script
Also updates hard-coded API version of kind used

/assign @cjwagner


cjwagner (approved):
Thanks Sean!

Issue #21090, opened 2021-03-01
presubmits were not triggering for kubernetes/kubernetes


BenTheElder:
**What happened**:

When a PR is pushed | opened in Kubernetes we're not seeing jobs trigger, just the automatic github statuses for required jobs like:
> pull-kubernetes-conformance-kind-ga-only-parallel Expected — Waiting for status to be reported

If you comment `/test all` manually jobs are triggered and run as expected.

**What you expected to happen**:

Tests should start when PRs that do not need `ok-to-test` are opened / pushed

**How to reproduce it (as minimally and precisely as possible)**:

Push to or open a PR in github.com/kubernetes/kubernetes

**Please provide links to example occurrences, if any**:

https://github.com/kubernetes/kubernetes/pull/96968#issuecomment-788243206

**Anything else we need to know?**:

Seems to be happening to all new PRs in this repo at least.
/area prow

BenTheElder:
> 2021-03-01 12:30:58.351 PST
[...]

PR #21082, opened 2021-03-01
Block conformance changes on release branches in k/k

nikhita:
Ref: slack discussion - https://kubernetes.slack.com/archives/C78F00H99/p1614202310015700?thread_ts=1613501041.053200&cid=C78F00H99

Follow up to https://github.com/kubernetes/test-infra/pull/21021

cc @hh @dims @justaugustus 
folks on original thread

cc @alejandrox1 @hasheddan @saschagrunert @jeremyrickard 
other sig-release leads

PR #21092, opened 2021-03-01
Revert "Block conformance changes on release branches in k/k"

spiffxp:
This reverts commit 1f152d6a89bc1ec4bb5dd4e9cdb0d6e664fe7ab6.

Reverts PR https://github.com/kubernetes/test-infra/pull/21082

We're hitting nil pointer exceptions in hook now, blocking presubmits from trigger in kubernetes/kubernetes (ref: https://github.com/kubernetes/test-infra/issues/21090)

Faster to roll this config change back than update prow


BenTheElder:
/lgtm
looking into the NPE

PR #21093, opened 2021-03-01
fix NPE in blockade

BenTheElder:
I think this is sufficient anyhow, see: https://github.com/kubernetes/test-infra/issues/21090#issuecomment-788253621

We probably should be doing this even if it's not the whole fix, we should check the actual value we want to gate for nil instead of it's source of truth / a proxy value.

### DIFF
--- a/prow/plugins/blockade/blockade.go
+++ b/prow/plugins/blockade/blockade.go
@@ -218,7 +218,7 @@ func compileApplicableBlockades(org, repo, branch string, log *logrus.Entry, blo
 			continue
 		}
 		// if branchregexp is specified, only consider blockades that apply to this branch
-		if raw.BranchRegexp != nil && !raw.BranchRe.MatchString(branch) {
+		if raw.BranchRe != nil && !raw.BranchRe.MatchString(branch) {
 			continue
 		}
 
--- a/prow/plugins/blockade/blockade_test.go
+++ b/prow/plugins/blockade/blockade_test.go
@@ -81,6 +81,12 @@ var (
 		BlockRegexps: []string{`test/conformance/testdata/.*`},
 		Explanation:  "6",
 	}
+	blockBadBranchRe = plugins.Blockade{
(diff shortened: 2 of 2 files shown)


BenTheElder:
see also: https://github.com/kubernetes/test-infra/pull/21021/files#r582803114

prow/plugins/blockade/blockade.go
```diff
@@ -218,7 +218,7 @@ func compileApplicableBlockades(org, repo, branch string, log *logrus.Entry, blo
 			continue
 		}
 		// if branchregexp is specified, only consider blockades that apply to this branch
-		if raw.BranchRegexp != nil && !raw.BranchRe.MatchString(branch) {
+		if raw.BranchRe != nil && !raw.BranchRe.MatchString(branch) {
```
spiffxp, line 221: Can we add a test that would have caught this?

BenTheElder: It's a bit unclear how we got into this state, I think maybe you have a regex that won't compile?

BenTheElder: but looking at https://github.com/kubernetes/test-infra/pull/21092/files, the regex there is valid ..?
and we appear to have tests that don't set this BranchRegexp field

BenTheElder: done

prow/plugins/blockade/blockade_test.go
```diff
@@ -92,6 +98,12 @@ func TestCalculateBlocks(t *testing.T) {
 		config          []plugins.Blockade
 		expectedSummary summary
 	}{
+		{
```
BenTheElder, line 101: this case NPEs if I revert the patch to blockade.go

spiffxp (approved):
/approve
/lgtm

nikhita:
Thanks, @BenTheElder!

PR #21100, opened 2021-03-01
Fix blockade regexp compilation.

cjwagner:
We were compiling the branch regexp, but incorrectly storing the result in a struct scoped to the for loop rather than updating the actual configuration.
/assign @BenTheElder @spiffxp 
fixes https://github.com/kubernetes/test-infra/issues/21090

### DIFF
--- a/prow/plugins/config.go
+++ b/prow/plugins/config.go
@@ -1166,15 +1166,15 @@ func compileRegexpsAndDurations(pc *Configuration) error {
 	}
 	pc.CherryPickUnapproved.BranchRe = branchRe
 
-	for _, blockade := range pc.Blockades {
-		if blockade.BranchRegexp == nil {
+	for i := range pc.Blockades {
+		if pc.Blockades[i].BranchRegexp == nil {
 			continue
 		}
-		branchRe, err := regexp.Compile(*blockade.BranchRegexp)
+		branchRe, err := regexp.Compile(*pc.Blockades[i].BranchRegexp)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


prow/plugins/config.go
```diff
@@ -1166,15 +1166,15 @@ func compileRegexpsAndDurations(pc *Configuration) error {
...
+		if pc.Blockades[i].BranchRegexp == nil {
 			continue
 		}
-		branchRe, err := regexp.Compile(*blockade.BranchRegexp)
+		branchRe, err := regexp.Compile(*pc.Blockades[i].BranchRegexp)
 		if err != nil {
-			return fmt.Errorf("failed to compile blockade branchregexp: %q, error: %v", *blockade.BranchRegexp, err)
+			return fmt.Errorf("failed to compile blockade branchregexp: %q, error: %v", *pc.Blockades[i].BranchRegexp, err)
 		}
-		blockade.BranchRe = branchRe
```
chaodaiG, line 1177: Not super convinced that this is the fix. `blockade` is a reference, this line actually change the value of reference struct

alvaroaleman: `blockade` is not a pointer (if that is what you mean by reference). This code never changed anything in `pc.Blockades`, as the `blockade` variable is scoped to the loop.

/lgtm
/approve
/hold

in case I missunderstood your comment

chaodaiG: You're right, I misread, thought the `*` was for blockade. 

/unhold

BenTheElder:
Thanks!

PR #20369, opened 2021-01-04
sig-cloud-provider: Add APIServerTracing focus to pull-kubernetes-e2e-gce-alpha-features

dashpole:
This enables the test for apiserver distributed tracing being added in https://github.com/kubernetes/kubernetes/pull/94942.

The ENABLE_APISERVER_TRACING env var creates the config file on the master during cluster creation.

PR #21102, opened 2021-03-01
Fix typo in alpha feature list

dashpole:
I added a typo when I added my feature gate to the alpha feature gates list, so I broke the config for the ServiceAccountIssuerDiscovery feature.

https://github.com/kubernetes/test-infra/pull/20369 is where it was introduced.

@ehashman deserves credit for finding it.

cc @mtaufen 
who added the feature gate initially.


ehashman (commented):
/lgtm
:sweat_smile:

PR #21103, opened 2021-03-01
ProwJob endpoint: Strip managed fields

alvaroaleman:
I doubt anyone looking at their prowjob ever wanted to see this, they
make it required to scroll a lot and provide no value in this context.


alvaroaleman:
The kube version used for the prow.k8s.io cluster appears to be too old for this, but look for example at https://prow.ci.openshift.org/prowjob?prowjob=a25aa2da-7add-11eb-81d7-0a580a800547, that is just super annoying to read

stevekuznetsov (approved):

PR #20755, opened 2021-02-05
releng: Migrate all build canary jobs to `krel`

justaugustus:
Outside of a few transient cluster scheduling issues, the `krel`-based `master` canary build job on Kubernetes Community infrastructure has been [running pretty reliably](https://prow.k8s.io/?job=ci-kubernetes-build-canary) for a few days, so let's cut the rest of the release branches over! 

- Fix incorrectly transposed testgrid annotations for build jobs

  Release branch build jobs were named as K8s infra ("canary") build jobs
  and vice versa. This fixes that.

  Took the opportunity to also properly name the 1.18 jobs by version,
  instead of generic suffix.

- build-canary jobs are releng-blocking, but release-informing

  Adds a fork annotation as well.

- Migrate all build canary jobs to `krel` (ref: https://github.com/kubernetes/release/issues/1711)

- Drop EOL 1.17 release branch jobs (ref: https://github.com/kubernetes/test-infra/pull/20589)

/assign @hasheddan @cpanato @saschagrunert
cc: @kubernetes/release-engineering @kubernetes/ci-signal @spiffxp

PR #20589, opened 2021-01-25
Cleanup release-1.17 slack mergewarnings and milestone applier

cpanato:
### Cleanup
 - remove release branch 1.17 from the slack mergewarnings 
 ~- remove release-1.17 jobs~
 - remove milestone applier for release-1.17

since this branch is not receiving updates anymore.

/kind cleanup

/assign @justaugustus @saschagrunert @hasheddan  @dims 
cc: @kubernetes/release-engineering

PR #21106, opened 2021-03-02
images: stop building 1.17 variant of kubekins-e2e

spiffxp:
1.17 is no longer receiving patch releases
- https://github.com/kubernetes/test-infra/pull/20755 dropped the 1.17 jobs
- https://github.com/kubernetes/test-infra/pull/20589 dropped some config info

this drops the 1.17 variant of kubekins-e2e

is there an umbrella issue for these sorts of "turn down infra for old releases" PRs?


spiffxp:

PR #21108, opened 2021-03-02
kubernetes-verify: Update go-canary jobs to use go1.16 images

justaugustus:
Follow-up to https://github.com/kubernetes/test-infra/pull/21107.
The releng-ci image does not have some of the runner.sh bootstrap logic, so we switch back to the kubekins go1.16 `go-canary` variant being built in https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-test-infra-push-kubekins-e2e/1366592695471968256.

Signed-off-by: Stephen Augustus <foo@auggie.dev>

/assign @spiffxp @dims @BenTheElder 
cc: @kubernetes/release-engineering 
/hold until the [image build postsubmit for kubekins-e2e](https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-test-infra-push-kubekins-e2e/1366592695471968256) completes

PR #21109, opened 2021-03-02
kubernetes-verify: Use go-canary kubekins-e2e variants for go1.16 update

justaugustus:
Follow-up to https://github.com/kubernetes/test-infra/pull/21108.

- kubernetes-verify: Use go-canary kubekins-e2e variants for go1.16 update
- kubekins-e2e: Build `master` and `experimental` variants with go1.16
- kubekins-e2e: Build go1.15 variants using go1.15.8

`pull-kubernetes-dependencies-go-canary` (go1.16 job) now passes on https://github.com/kubernetes/kubernetes/pull/98572: https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/98572/pull-kubernetes-dependencies-go-canary/1366599409453764608

/assign @spiffxp @dims @BenTheElder 
/hold until [`pull-kubernetes-verify-go-canary`](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/98572/pull-kubernetes-verify-go-canary/1366599409504096256) passes as well

### DIFF
--- a/config/jobs/kubernetes/sig-testing/verify.yaml
+++ b/config/jobs/kubernetes/sig-testing/verify.yaml
@@ -17,7 +17,10 @@ presubmits:
       preset-dind-enabled: "true"
     spec:
       containers:
-      - image: gcr.io/k8s-testimages/kubekins-e2e:v20210226-c001921-master
+      # TODO(releng): Use the 'master' variant once kubernetes/kubernetes has
+      #               been updated to go1.16
+      #               ref: https://github.com/kubernetes/kubernetes/pull/98572
+      - image: gcr.io/k8s-testimages/kubekins-e2e:v20210302-c18942a-go-canary
         imagePullPolicy: Always
         command:
         - runner.sh
--- a/config/jobs/kubernetes/sig-testing/dependencies.yaml
+++ b/config/jobs/kubernetes/sig-testing/dependencies.yaml
@@ -21,7 +21,10 @@ presubmits:
       - name: main
         command:
         - runner.sh
-        image: gcr.io/k8s-testimages/kubekins-e2e:v20210226-c001921-master
+        # TODO(releng): Use the 'master' variant once kubernetes/kubernetes has
+        #               been updated to go1.16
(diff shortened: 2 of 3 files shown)


hasheddan (approved):
/lgtm

saschagrunert (approved):
/lgtm

justaugustus:
go1.16 canary verify jobs are passing:
-  `pull-kubernetes-dependencies-go-canary`: https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/98572/pull-kubernetes-dependencies-go-canary/1366634510489227264
- `pull-kubernetes-verify-go-canary`: https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/98572/pull-kubernetes-verify-go-canary/1366631602737647616

/hold cancel

config/jobs/kubernetes/sig-testing/verify.yaml
```diff
@@ -17,7 +17,10 @@ presubmits:
       preset-dind-enabled: "true"
     spec:
       containers:
-      - image: gcr.io/k8s-testimages/kubekins-e2e:v20210226-c001921-master
+      # TODO(releng): Use the 'master' variant once kubernetes/kubernetes has
+      #               been updated to go1.16
+      #               ref: https://github.com/kubernetes/kubernetes/pull/98572
+      - image: gcr.io/k8s-testimages/kubekins-e2e:v20210302-c18942a-go-canary
```
BenTheElder, line 23: we should not be using the canary tag for the real job?

we should be updating the real image, after testing in the canary with the latest version in a canary/experimental job. we auto push -latest images.

these changes left other jobs not updated.

BenTheElder: if we want to have a special image to try go early for an extended period that should get it's own job

we should have a manually invoked job that uses the standard image track at the -latest tag with always pull instead of the date-sha version (which lets us test the latest version of the standard image prior to an image bump here)

the primary job should keep using the date-sha version so we can roll forward/back without waiting for an image push, and track the versions in git.

PR #21112, opened 2021-03-02
add prow job to check if all json files are sorted

kkeshavamurthy:
Job to check if json files are sorted

CC: @codenrhoden @CecileRobertMichon

### DIFF
--- a/config/jobs/kubernetes-sigs/image-builder/image-builder-presubmits.yaml
+++ b/config/jobs/kubernetes-sigs/image-builder/image-builder-presubmits.yaml
@@ -46,3 +46,23 @@ presubmits:
     annotations:
       testgrid-dashboards: sig-cluster-lifecycle-image-builder
       testgrid-tab-name: pr-azure-sigs
+  - name: json-sort-check
+    decorate: true
+    always_run: false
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes-sigs/image-builder/image-builder-presubmits.yaml
```diff
@@ -46,3 +46,26 @@ presubmits:
     annotations:
       testgrid-dashboards: sig-cluster-lifecycle-image-builder
       testgrid-tab-name: pr-azure-sigs
+  - name: json-sort-check
+    labels:
```
codenrhoden, line 50: I don't think these two labels are necessary for this job, since we don't need docker-in-docker to run that test.

kkeshavamurthy: Removed.

codenrhoden:
Nice. So I think this is good to merge, since it's set as optional and `always_run` is false, it will only run if specifically triggered at the moment. That way we can wait until the PR to clean up the JSON files is merged.

/lgtm

PR #21113, opened 2021-03-02
kubernetes-csi: disable Bazel for master and >= 1.21

pohly:
Bazel support was removed, so we must not attempt to build Kubernetes
with that as it will just fail.

### DIFF
--- a/config/jobs/kubernetes-csi/gen-jobs.sh
+++ b/config/jobs/kubernetes-csi/gen-jobs.sh
@@ -47,7 +47,7 @@ latest_stable_k8s_version="1.20"
 hostpath_driver_version="v1.5.0"
 
 # We need this image because it has Docker in Docker and go.
-dind_image="gcr.io/k8s-testimages/kubekins-e2e:v20210113-cc576af-master"
+dind_image="gcr.io/k8s-testimages/kubekins-e2e:v20210226-c001921-master"
 
 # All kubernetes-csi repos which are part of the hostpath driver example.
 # For these repos we generate the full test matrix. For each entry here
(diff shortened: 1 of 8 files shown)


config/jobs/kubernetes-csi/gen-jobs.sh
```diff
@@ -316,6 +316,16 @@ snapshotter_version() {
...
+use_bazel() {
+    local kubernetes="$1"
+
+    # Kubernetes 1.21 removed support for building with Bazel.
+    if version_gt "$kubernetes" "1.21"; then
+        echo "false"
+    else
+        echo "true"
+    fi
+}
```
pohly, line 328: We could avoid adding this function and simple pass `CSI_PROW_USE_BAZEL=false` to all jobs. Then jobs which currently build images for Kubernetes 1.20 would also stop using Bazel although at that time, using Bazel was still the recommended way of building. I felt it was safer to follow recommendations for each release, hence this version check.

spiffxp (approved):
/approve
/lgtm

BenTheElder:
Sorry I missed this!
I thought I'd found all CI usage 😞

pohly:
No problem. This one was rather indirect and not obvious from the job.

Thanks for removing Bazel. I lost track of how many times one of my PRs failed initially in the Prow jobs because I had forgotten to refresh a BUILD file after a minor import change.

config/jobs/kubernetes-csi/gen-jobs.sh
BenTheElder: this is pretty much what we're doing with rest of CI for now, and especially kind jobs have this env based approach.

though because we carry some extra complexity in kind for building w/ bazel and are re-working image builds I may drop this between now and the next kind release and either:
- pin kubernetes release branch CI to kind v0.10.0
- go ahead and switch the release branch CI over to make for kind jobs at least (lower impact the further back you go anyhow)

PR #21114, opened 2021-03-02
Update all Cluster API main jobs to use kubekins with Go 1.16

vincepri:


justaugustus (approved):
From RelEng, it'll be great to have CAPI using the `go-canary` variant:
- It roughly runs parallel to the `master` kubekins-e2e variant
- When it doesn't, CAPI becomes a feedback loop for RelEng, which is ultimately what we want more of

/lgtm
/hold if any feedback is required from other CAPI maintainers; release when ready

ref: https://kubernetes.slack.com/archives/C09QZ4DQB/p1614703470050200?thread_ts=1614614854.041700&cid=C09QZ4DQB

fabriziopandini:
/lgtm
from my side.
+100 for CAPI stepping into feedback loop for RelEng

PR #21115, opened 2021-03-02
secrets: censor more efficiently

stevekuznetsov:
The `bytes.ReplaceAll()` call will copy the input, so the previous
implementation of the secret agent's censoring would take up O(MN)
memory where M is the size of the input and N is the number of secrets
that are loaded. There's no need for that. The `bytereplacer` utility
will operate in-place (unless it needs to reallocate for a longer
replacement) and is faster, as well. I've kept the old `Censor()`
function around and it will continue not to mutate input. The content of
the censored text is changing, though, from the literal `"CENSORED"`
string to repeated `"*"` characters as long as the input.

### DIFF
--- a/prow/config/secret/agent.go
+++ b/prow/config/secret/agent.go
@@ -18,22 +18,22 @@ limitations under the License.
 package secret
 
 import (
-	"bytes"
 	"os"
 	"sync"
 	"time"
 
 	"github.com/sirupsen/logrus"
-
 	"k8s.io/apimachinery/pkg/util/sets"
 
 	"k8s.io/test-infra/prow/logrusutil"
+	"k8s.io/test-infra/prow/secretutil"
 )
 
 // Agent watches a path and automatically loads the secrets stored.
 type Agent struct {
 	sync.RWMutex
 	secretsMap map[string][]byte
+	*secretutil.ReloadingCensorer
 }
 
 // Start creates goroutines to monitor the files that contain the secret value.
@@ -46,13 +46,15 @@ func (a *Agent) Start(paths []string) error {
 	}
 
 	a.secretsMap = secretsMap
+	a.ReloadingCensorer = secretutil.NewCensorer()
+	a.refreshCensorer()
 
 	// Start one goroutine for each file to monitor and update the secret's values.
 	for secretPath := range secretsMap {
 		go a.reloadSecret(secretPath)
 	}
 
-	logrus.SetFormatter(logrusutil.NewCensoringFormatter(logrus.StandardLogger().Formatter, a.getSecrets))
+	logrus.SetFormatter(logrusutil.NewFormatterWithCensor(logrus.StandardLogger().Formatter, a.ReloadingCensorer))
 
 	return nil
 }
@@ -119,6 +121,16 @@ func (a *Agent) setSecret(secretPath string, secretValue []byte) {
 	a.Lock()
 	defer a.Unlock()
 	a.secretsMap[secretPath] = secretValue
+	a.refreshCensorer()
+}
+
+// refreshCensorer should be called when the lock is held and the secrets map changes
+func (a *Agent) refreshCensorer() {
+	var secrets [][]byte
+	for _, value := range a.secretsMap {
+		secrets = append(secrets, value)
+	}
+	a.ReloadingCensorer.RefreshBytes(secrets...)
 }
 
 // GetTokenGenerator returns a function that gets the value of a given secret.
@@ -128,17 +140,16 @@ func (a *Agent) GetTokenGenerator(secretPath string) func() []byte {
 	}
 }
 
-const censored = "CENSORED"
-
-var censoredBytes = []byte(censored)
-
 // Censor replaces sensitive parts of the content with a placeholder.
 func (a *Agent) Censor(content []byte) []byte {
-	for sKey := range a.secretsMap {
-		secret := a.GetSecret(sKey)
-		content = bytes.ReplaceAll(content, secret, censoredBytes)
+	a.RLock()
+	defer a.RUnlock()
+	if a.ReloadingCensorer == nil {
+		// there's no constructor for an Agent so we can't ensure that everyone is
+		// trying to censor *after* actually loading a secret ...
+		return content
 	}
-	return content
+	return secretutil.AdaptCensorer(a.ReloadingCensorer)(content)
 }
 
 func (a *Agent) getSecrets() sets.String {
--- a/prow/logrusutil/logrusutil.go
+++ b/prow/logrusutil/logrusutil.go
@@ -18,14 +18,13 @@ limitations under the License.
 package logrusutil
 
 import (
-	"bytes"
-	"strings"
 	"sync"
 	"time"
 
 	"github.com/sirupsen/logrus"
 	"k8s.io/apimachinery/pkg/util/sets"
 
+	"k8s.io/test-infra/prow/secretutil"
 	"k8s.io/test-infra/prow/version"
 )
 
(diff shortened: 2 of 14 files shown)


stevekuznetsov:
```
 Validating bazel rules...
gazelle: finding module path for import go4.org/bytereplacer: exec: "go": executable file not found in $PATH
Current rules (-) do not match expected (+):
gazelle diff:
--- prow/config/secret/BUILD.bazel	1970-01-01 00:00:00.000000000 +0000
+++ prow/config/secret/BUILD.bazel	1970-01-01 00:00:00.000000000 +0000
@@ -14,7 +14,6 @@
         "//prow/logrusutil:go_default_library",
         "@com_github_sirupsen_logrus//:go_default_library",
         "@io_k8s_apimachinery//pkg/util/sets:go_default_library",
-        "@org_go4//bytereplacer:go_default_library",
     ],
 ) 
```

What does this mean? I ran `hack/update-bazel.sh` locally ...

prow/config/secret/agent.go
```diff
@@ -128,17 +136,24 @@ func (a *Agent) GetTokenGenerator(secretPath string) func() []byte {
...
-		secret := a.GetSecret(sKey)
-		content = bytes.ReplaceAll(content, secret, censoredBytes)
+	out := make([]byte, len(content))
+	copy(out, content)
+	return a.CensorInPlace(out)
+}
+
+// CensorInPlace replaces sensitive parts of the content with a placeholder,
+// mutating the input content.
+func (a *Agent) CensorInPlace(content []byte) []byte {
```
alvaroaleman, line 148: Is there any usecase where we want this? I would prefer not to have two seperate `Censor` methods in the agents api. Also, lets not return something, that makes misuse too easy as it creates the impression that was a copy.

stevekuznetsov: We want this always? I don't see when we *don't* want this.

stevekuznetsov:
@alvaroaleman updated to have stronger guarantees on `CensorInPlace()`

prow/config/secret/agent.go
alvaroaleman: Then lets remove the old Censor?

stevekuznetsov: People vendor this code. I want to not break them :)

stevekuznetsov:
Guess we had a test depend on the literal censor string, fixed

stevekuznetsov:
/hold

WTF, the `logrus` formatter copies this logic? ...

prow/logrusutil/logrusutil.go
```diff
@@ -99,36 +98,26 @@ func (f CensoringFormatter) Format(entry *logrus.Entry) ([]byte, error) {
...
-	}
+	f.censorer.Censor(content)
 	return content
 }
 
 // NewCensoringFormatter generates a `CensoringFormatter` with
 // a formatter as delegate and a set of strings to censor
 func NewCensoringFormatter(f logrus.Formatter, getSecrets func() sets.String) CensoringFormatter {
+	censorer := secretutil.NewCensorer()
+	censorer.Refresh(getSecrets().List()...)
```
stevekuznetsov, line 111: Technically speaking this will not refresh the list of secrets, but the previous implementation would. A quick survey of the users of this shows it's people shoving static lists in there ...

```go
logrus.SetFormatter(logrusutil.NewCensoringFormatter(logrus.StandardLogger().Formatter, func() sets.String {
	return sets.NewString(oauthToken)
}))
```

stevekuznetsov (commented):
Updated to unify this with the `logrusutil` package, creating a unified censoring implementation that will be useful for both. Not sure why we had this logic duplicated ...

stevekuznetsov:
The _unit tests_ got duplicated, too? What on earth am I even looking at

prow/secretutil/censor.go
```diff
@@ -0,0 +1,103 @@
...
+		asStrings = append(asStrings, string(secret))
+	}
+	c.Refresh(asStrings...)
+}
+
+// Refresh refreshes the set of secrets that we censor.
+func (c *ReloadingCensorer) Refresh(secrets ...string) {
+	var replacements []string
+	addReplacement := func(s string) {
+		replacements = append(replacements, s, strings.Repeat(`*`, len(s)))
```
alvaroaleman, line 75: Instead of always telling the length of the original secret, can we make this the `min(len(s), 3)` to avoid exposing that info in the 99% of cases where the secret is longer than three characters?

prow/secretutil/censor.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+	for _, secret := range secrets {
+		if secret == "" {
+			continue
+		}
+		if trimmed := strings.TrimSpace(secret); trimmed != secret {
+			secret = trimmed
+		}
+		addReplacement(secret)
+		encoded := base64.StdEncoding.EncodeToString([]byte(secret))
```
alvaroaleman, line 85: Lets mention this in the godocs

prow/secretutil/censor_test.go
```diff
@@ -0,0 +1,88 @@
...
+	"github.com/google/go-cmp/cmp"
+)
+
+func copyBytes(input []byte) []byte {
+	output := make([]byte, len(input))
+	copy(output, input)
+	return output
+}
+
+var text = []byte("secret SECRET c2VjcmV0 sEcReT")
```
alvaroaleman, line 31: You can replace this and copyBytes with a `func() []byte`. Lets scope this to the testcase where its actually used though to not make it pop up in IDE autocompletion

prow/secretutil/censor.go
stevekuznetsov: In the context of the streaming censoring algorithm for artifacts, allowing this to mutate the length of the output is not ideal. A quick overview of the algorithm again:

1. read in 2N bytes where N is the longest secret to censor
2. censor all secrets in those 2N bytes
3. write the first N bytes to the output file
4. move the last N bytes to the start of the buffer
5. read N more bytes into the buffer
6. loop from 2

So if censoring changes the length, things get math-y and hard there. I agree that your suggestion is slightly nicer - do you think it's worth the extra complexity it will incur later?

petr-muller: Leaking length is technically still a leak b/c it reduces the searching space significantly. I think the complexity is worth it.

petr-muller: Leaking length limits the search space significantly so I think the added complexity would be worth it.

stevekuznetsov: The library can't mutate the length of the slice itself, so it returns e.g.:

```
secret SECRET c2VjcmV0 sEcReT
*** SECRET *** sEcReT0 sEcReT
<--- replacement ---><suffix>
```

I'd need to post-process the slice walking backwards to trim it. This seems like needless complexity but I'll give it a shot.

stevekuznetsov: @petr-muller how does it matter that an attacker sees that a token is 30 chars long for e.g. a GitHub OAuth token? I don't buy the "search space" argument - not only would they need to, after seeing an artifact with a censored bit, determine *what* it's censoring, what system it is for, then they have ... no information?

Like, all GitHub OAuth tokens are of the same length. Knowing that our token is also that length gives them info enough to search .. the entire space. While in a purely technical sense sure it leaks something, I just don't see the utility of it.

stevekuznetsov: And, actually, post-processing the slice to trim the suffix requires us to re-allocate the memory anyway, since we need to keep the old slice around for the traversal. I'm leaning hard against doing this.

stevekuznetsov: After more communication, it's clear that it will not be possible to achieve this without changing the byte replacer itself. That's orders of magnitude harder than this pull. I've left the signature here with `*[]byte` to allow it to be done in the future, but it won't be part of this initial pull.

petr-muller (approved):

PR #10515, opened 2018-12-20
Allow cuj, platform, community, and team labels to be set via bot.

jlewi:
* The Kubeflow project would like to be able to use the K8s bot to add labels
  prefixed by cuj and platform

* cuj stands for critical user journey. We'd like to use this to group related
  issues to delivering a core user experience.

* We use platform to group issues by platform; e.g.
  /platform/gcp
  /platform/minikube.

* There is an open issue #8648 is fixed to allow whitelisting additional prefixes. In the interim we can whitelist individual labels for use with /label.

PR #21119, opened 2021-03-02
Re-add additional labels

alvaroaleman:
They got removed in 6edfa06cea600aba8df5342a3b2c3716d8cede8f when cleaning up kubeflow.

Its not entirely clear for me which of these only belonged to kubeflow, so I've re-added all of them. Removing them later should be easy. The tide labels are definitely needed everywhere.

/assign @fejta @chaodaiG

### DIFF
--- a/config/prow/plugins.yaml
+++ b/config/prow/plugins.yaml
@@ -117,6 +117,23 @@ label:
     # This label is used by the API review process
     # https://git.k8s.io/community/sig-architecture/api-review-process.md#mechanics
     - api-review
+    # These labels are used by Kubeflow
+    # TODO(https://github.com/kubernetes/test-infra/issues/8648): Switch
+    # to configuring prefixes and not individual labels once  that's supported.
+    - community/discussion
+    - community/maintenance
+    - community/question
+    - cuj/build-train-deploy
+    - cuj/multi-user
+    - platform/aws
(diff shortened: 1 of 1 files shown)


cjwagner (approved):

config/prow/plugins.yaml
```diff
@@ -117,6 +117,23 @@ label:
...
+    - community/maintenance
+    - community/question
+    - cuj/build-train-deploy
+    - cuj/multi-user
+    - platform/aws
+    - platform/azure
+    - platform/gcp
+    - platform/minikube
+    - platform/other
+    - tide/merge-method-merge
```
BenTheElder, line 133: I think only the tide labels from here down are used outside kubeflow.

Labels like `platform/*` would be `area/foo` in a kubernetes project. `community/foo` would be `kind/foo`. the `cuj/*` labels seem very specific to kuebflow.

cjwagner: Thats correct, I found the PR from the blame: https://github.com/kubernetes/test-infra/pull/10515

cjwagner: Annnnd its already been fixed: https://github.com/kubernetes/test-infra/pull/21131 :man_facepalming: 
Sorry for the noise.

PR #21120, opened 2021-03-02
[Kettle] Unpin sdk version of google-cloud

MushuEE:
#19414

BigQuery error in load operation: Could not connect with BigQuery server due to:
RedirectMissingLocation('Redirected but the response is missing a Location:
header.',) was introduced in 294+ version and was fixed in 312+ versions, we are at 320 today. 

/area kettle

### DIFF
--- a/kettle/Dockerfile
+++ b/kettle/Dockerfile
@@ -35,7 +35,7 @@ RUN pip3 install requests google-cloud-pubsub==2.1.0 google-cloud-bigquery==2.2.
 RUN curl -fsSL https://downloads.python.org/pypy/pypy3.6-v7.3.1-linux64.tar.bz2 | tar xj -C opt  && \
     ln -s /opt/pypy*/bin/pypy3 /usr/bin
 
-RUN curl https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-sdk-294.0.0-linux-x86_64.tar.gz | tar xfz - -C . && \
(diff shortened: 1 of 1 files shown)


kettle/Dockerfile
```diff
@@ -35,8 +35,9 @@ RUN pip3 install requests google-cloud-pubsub==2.1.0 google-cloud-bigquery==2.2.
 RUN curl -fsSL https://downloads.python.org/pypy/pypy3.6-v7.3.1-linux64.tar.bz2 | tar xj -C opt  && \
     ln -s /opt/pypy*/bin/pypy3 /usr/bin
 
-RUN curl https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-sdk-294.0.0-linux-x86_64.tar.gz | tar xfz - -C . && \
-    ./google-cloud-sdk/install.sh -q && \
+RUN curl -o installer https://sdk.cloud.google.com && \
```
amwat, line 38: we should just update this to a newer version instead

amwat (commented):
If anything this means we should pin harder?
so that casual changes don't break us

kettle/Dockerfile
spiffxp: Completely agree

MushuEE: Works for me

amwat (commented):
/lgtm
/approve

PR #21123, opened 2021-03-03
Fix graceful-node-shutdown focus

wzshiming:
Fixes kubernetes/kubernetes#99667


bobbypage:
thanks for quick fix!

/lgtm

PR #21125, opened 2021-03-03
update ci-k8sio-audit image and logic

hh:
Based on the recent work to create a bunch of small audit PRs.

- Using an image that already includes jq, golang, and tooling to build pr-creator
- Simplify commit variables and pr-creation


spiffxp (approved):
/approve
/lgtm

PR #21055, opened 2021-02-26
Switch containerd cri e2e to the right project

dims:
sig-node/containerd.yaml is using `cri-containerd-node-e2e` lets switch
away from the `cri-c8d-pr-node-e2e` which seems to be broken

See https://github.com/containerd/containerd/pull/5095#issuecomment-786908342

Signed-off-by: Davanum Srinivas <davanum@gmail.com>

PR #21127, opened 2021-03-03
Trying to fix pull-containerd-node-e2e

adisky:
Updating image from cos-81 to cos-85
getting error `unable to create gce instance with running docker daemon for image cos-81-12871-1245-15` and we are using cos-85 everywhere else.

Related to #21055


adisky:
cc @dims @SergeyKanzhelev

PR #21128, opened 2021-03-03
tide: add Message and MessageBody feilds to Commit

hi-rustin:
We would like to be able to add only the sign off message to the commit body when tide squash merges, but it seems that tide does not currently support that feature. So I would like to add these two fields and use the message body of the first commit in the template.

These fields are defined in commit, for more information see: https://docs.github.com/en/graphql/reference/objects#commit

### DIFF
--- a/prow/tide/tide.go
+++ b/prow/tide/tide.go
@@ -1731,6 +1731,8 @@ type Commit struct {
 		Contexts []Context
 	}
 	OID               githubql.String `graphql:"oid"`
+	Message           githubql.String
+	MessageBody       githubql.String
 	StatusCheckRollup StatusCheckRollup
 }
 


hi-rustin:
/cc @alvaroaleman 

Could you please take a look?  Thanks!

prow/tide/tide.go
```diff
@@ -1731,6 +1731,8 @@ type Commit struct {
 		Contexts []Context
 	}
 	OID               githubql.String `graphql:"oid"`
+	Message           githubql.String
```
alvaroaleman, line 1734: I don't want to add something like this before we actually start using it and have agreed that using it in a certain way is something we want to do.

hi-rustin: So do you have any suggestions to solve this problem we have now? Because our community is using squash mode to merge pr, but we don't want to use the default squash body provided by GitHub because that would take a lot of commit titles with it. But we need the sign off information. That's why I thought to get the body of the first commit.

alvaroaleman: I don't think that is possible. Squashing commit conceptionally conflicts with preserving signed-off. It also waters the meaning of signed-off, because having signed-off one commit in a PR doesn't imply having signed-off all. I would suggest to just tell your users to do proper commits and squash themselves and then merge with a different merge method that keeps the commits.

hi-rustin: /close

Thanks for the reply, it makes sense.

PR #21136, opened 2021-03-03
Report prow deployment job status on Slack

chaodaiG:
Related https://github.com/kubernetes/enhancements/pull/2540

Report prow deployment job status on Slack instead of oncall manually posting

### DIFF
--- a/config/jobs/kubernetes/test-infra/test-infra-trusted.yaml
+++ b/config/jobs/kubernetes/test-infra/test-infra-trusted.yaml
@@ -99,6 +99,15 @@ postsubmits:
     branches:
     - ^master$
     max_concurrency: 1
+    reporter_config:
+      slack:
+        channel: "prow-alerts"
+        job_states_to_report:
+        - success
+        - failure
+        - aborted
+        - error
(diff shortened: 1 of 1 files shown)


chaodaiG:
/cc @spiffxp @cjwagner @stevekuznetsov 

The current change proposes to send these to `prow-alerts` channel instead of `testing-ops` to avoid spamming the latter and make it less valuable. However currently `prow-alerts` is pretty spammy, are we sure the current alerts are looked at?

spiffxp:
> /cc @spiffxp @cjwagner @stevekuznetsov 
> 
> 
> 
> The current change proposes to send these to `prow-alerts` channel instead of `testing-ops` to avoid spamming the latter and make it less valuable. However currently `prow-alerts` is pretty spammy, are we sure the current alerts are looked at?

I think we should vastly increase the threshold for or remove the ghproxy alerts entirely. Informational alerts are good, but those seem to be pure noise that be adequately tracked via ghproxy graphs

But I don't think that precludes adding these to that channel

spiffxp (approved):
/approve
/lgtm

spiffxp:
Opened https://github.com/kubernetes/test-infra/issues/21090#issuecomment-790027245 for followup on prow alert noise

Issue #21137, opened 2021-03-03
Decouple test image bumping from prow bumping


spiffxp:
https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/99655/pull-kubernetes-unit/1367006331575209984
=> prowjob yaml button =>
https://prow.k8s.io/prowjob?prowjob=cb9cbe1e-7bed-11eb-97bd-22fa521bf179
=> look for image: =>
gcr.io/k8s-testimages/kubekins-e2e:v20210226-c001921-master

yeah that's not going to have the go1.16, looks like we haven't had a prow bump since 02/26

we need to decouple these update loops (images for prow itself, test images) as we discussed at the last SIG testing. cc @spiffxp

__Originally posted by @BenTheElder in https://github.com/kubernetes/kubernetes/pull/99655#issuecomment-789494480__

BenTheElder:
cc @chaodaiG

chaodaiG:
cc @mpherman2 as well

spiffxp:
Pulling out of https://github.com/kubernetes/test-infra/pull/21144#pullrequestreview-604489193

One question for followup: is there some way we could have merges of just these autobump PRs alert to slack?

A longer term thing: someplace to see all deploys and all config changes pushed to prow.k8s.io to aid "what changed when"

But for now being filter down just to "big PRs that change a lot" matches the current experience people are used to. I'd rather not take away what they're used to before having something better.

spiffxp:
Let's figure out what it would look like to hand this off to community.  My sketch:
- if they can, community members in appropriate OWNERS can approve these now
- wait until after code freeze, formally handoff ASAP
- use existing groups (ci-signal or release-eng) or use this to bootstrap a prowjob-oncall
- making sure the group knows
  - start with empty oncall field and hardcode ping to appropriate github team?
  - move to something generated by and stored in community
- document that this is how jobs configs are bumped for prow.k8s.io (a linkable subheading of https://github.com/kubernetes/test-infra/tree/master/config/jobs if there isn't a better place)

spiffxp:
/reopen
the job landed, but there's still more to do

chaodaiG:
Also https://github.com/kubernetes/test-infra/pull/21144 probably had left out some of the images in prow jobs that's part of prow but not `gcr.io/k8s-testimages`. I will take a look at the new bumping PRs generated and see what's missing

BenTheElder:
[...]

PR #21144, opened 2021-03-03
Split prow bump

chaodaiG:
Fixed https://github.com/kubernetes/test-infra/issues/21137

Note that this change still results in both PRs assigned to testinfra oncalls. And #21142 was created to address this problem

### DIFF
--- a/config/jobs/kubernetes/test-infra/test-infra-trusted.yaml
+++ b/config/jobs/kubernetes/test-infra/test-infra-trusted.yaml
@@ -906,7 +906,7 @@ periodics:
       command:
       - /generic_autobump
       args:
-      - --config=config/prow/autobump-config.yaml
+      - --config=config/prow/autobump-config/prow-component-autobump-config.yaml
       volumeMounts:
       - name: github
         mountPath: /etc/github-token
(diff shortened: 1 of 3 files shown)


chaodaiG:
/hold

Would like to wait for @BenTheElder and @spiffxp taking a look
/cc @BenTheElder @spiffxp

spiffxp (approved):
/approve
/lgtm
Keeping to oncall sounds good for now.  Let's wait til after code freeze to propose formally handing off to community.

One question for followup: is there some way we could have merges of just these autobump PRs alert to slack?  

Someplace to see all deploys and all config changes pushed to prow.k8s.io sounds like a good followup that would aid troubleshooting.

But being able to filter down just to "big PRs that change a lot" matches the current experience people are used to.  I'd rather not take away what they're used to before having something better.

BenTheElder:
SGTM, I do think we should relieve test-infra oncall of being concerned with these though, the community should be maintaining the images we test with in k8s.io without depending on the google infra oncall (see: recent go1.16 bump being in a bad state while people just waited for the oncall to upgrade ...)

PR #21151, opened 2021-03-04
GracefulNodeShutdown skip ubuntu

wzshiming:
xref kubernetes/kubernetes#99766

Skip testing GracefulNodeShutdown on ubuntu, because it does not have gdbus installed by default.


SergeyKanzhelev:
/close

in favor of https://github.com/kubernetes/kubernetes/pull/99805

PR #21155, opened 2021-03-04
[pull-containerd-node-e2e-job] Better logging and increase timeout

dims:
- The current setup for node e2e does not capture logs for all nodes that were started. So trying the other new alternate toggle to see if that works better
- Bump timeout just in case the cluster is underpowered

Yes, this is just for diagnosis, will flip them right back if things don't work


mikebrow (approved):
LGTM
Thanks!

PR #21157, opened 2021-03-04
Gerrit label processing: include job name and id in debug logs

chaodaiG:
We are seeing very strange behavior of gerrit label processing, that the value for label 'prow.k8s.io/gerrit-patchset' is a whitespace. The only place this value being set is https://github.com/kubernetes/test-infra/blob/cdbc8651410c316098479c03fbf54d22fe6db8b8/prow/gerrit/adapter/adapter.go#L306, and 'strconv.Itoa' technically should never return a whitespace. The current log doesn't tell which prowjob it is from and made it hard to debug, add these fields in debug message to help debug

### DIFF
--- a/prow/pod-utils/decorate/podspec.go
+++ b/prow/pod-utils/decorate/podspec.go
@@ -92,11 +92,14 @@ func PodUtilsContainerNames() sets.String {
 // User-provided extraLabels and extraAnnotations values will take precedence over auto-provided values.
 func LabelsAndAnnotationsForSpec(spec prowapi.ProwJobSpec, extraLabels, extraAnnotations map[string]string) (map[string]string, map[string]string) {
 	jobNameForLabel := spec.Job
+	log := logrus.WithFields(logrus.Fields{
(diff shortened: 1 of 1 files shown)


chaodaiG:
Example log:
```
jsonPayload: {
component: "gerrit"
func: "k8s.io/test-infra/prow/pod-utils/decorate.LabelsAndAnnotationsForSpec"
msg: "Removing invalid label"
errors: [
0: "a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')"
]
key: "prow.k8s.io/gerrit-patchset"
level: "warning"
file: "prow/pod-utils/decorate/podspec.go:135"
value: " "
}
```

alvaroaleman (approved):

PR #21159, opened 2021-03-04
Dump nodes better when instance groups are not used (node-e2e)

dims:
In node e2e tests, we don't end up capturing any logs directly from
nodes so it is very difficult to diagnose issues. This happends because
the node e2e harness does not create/use instance groups, so in this
case we just do a simple query to list instances and use that.


BenTheElder:
cc @amwat

amwat (commented):
/lgtm
/approve

we don't have node tests support in kubetest2 yet, will keep this in mind to start with :)

PR #21162, opened 2021-03-04
kubernetes: stop using bazel in dualstack jobs

BenTheElder:
/cc @aojea


aojea:
/lgtm
/approve
thanks for catching it

PR #21164, opened 2021-03-04
sidecar: censor artifacts and logs before upload

stevekuznetsov:
Often, the output of the `sidecar` tool is a public cold storage bucket.
It's never appropriate to push secret data to these buckets, so
`sidecar` should be able to censor the data it's operating over before
upload. This patch adds the framework for the mechanism. Users opt into
this mechanism by passing the `sidecar` options for secrets they wish to
censor. This may trivially be employed in Prow by mounting secret
volumes on the Pod to this container, too. When secrets are configured,
`sidecar` will censor them. The algorithm uses a small sliding window
frame buffer so as to not increase the memory footprint of this process
when users create enormous logs or artifacts.

Today, this process looks for literal secret values as well as
base64-encoded values, which are common in k8s secrets. Users may
accidentally `echo` a secret or a `Secret`, and this approach catches
both. The algorithm today only handles plain text files, but I plan to
improve it to seamlessly untar and unzip archives as well to censor
their contents before reassembling the archives for upload.


stevekuznetsov:
Go 1.15 does not have `io.NopCloser`, oops!

prow/sidecar/run.go
```diff
@@ -115,6 +115,11 @@ func (o Options) Run(ctx context.Context) (int, error) {
 	// uploading, so we ignore the signals.
 	signal.Ignore(os.Interrupt, syscall.SIGTERM)
 
+	if len(o.Secrets) > 0 {
+		if err := o.censor(); err != nil {
+			logrus.Warnf("Failed to censor data: %v", err)
```
stevekuznetsov, line 120: Or we could refuse to upload anything .. ?

prow/sidecar/censor.go
```diff
@@ -0,0 +1,206 @@
...
+	}
+	censorer := secretutil.NewCensorer()
+	censorer.RefreshBytes(secrets...)
+
+	// We want to keep as little of the file in memory as possible in order for
+	// this operation to be reasonably performant in space. However, to guarantee
+	// that we censor every instance of every secret, our frame buffer must be at
+	// least two times larger than the largest secret we are about to censor.
+	frameSize := defaultFrameSize
+	if largest := censorer.LargestSecret(); 2*largest > frameSize {
```
alvaroaleman, line 71: this logic is effectively untested, maybe add a private variadic option to override the default?

prow/sidecar/censor.go
```diff
@@ -0,0 +1,206 @@
...
+			}(path)
+			return nil
+		}); err != nil {
+			return fmt.Errorf("could not walk items to censor them: %w", err)
+		}
+	}
+
+	wg.Wait()
+	close(errors)
+	return kerrors.NewAggregate(errs)
```
alvaroaleman, line 111: This races with the goroutine that writes to `errs`

prow/sidecar/censor.go
```diff
@@ -0,0 +1,206 @@
...
+	if err := os.Rename(output.Name(), path); err != nil {
+		return fmt.Errorf("could not overwrite file after censoring: %w", err)
+	}
+
+	return nil
+}
+
+// censor censors input data and streams it to the output. We have a memory footprint of frameSize bytes.
+func censor(input io.ReadCloser, output io.WriteCloser, censorer secretutil.Censorer, frameSize int) error {
+	if frameSize%2 != 0 {
```
alvaroaleman, line 140: what you call `framesize` is buffer size, frame size is the half of it

prow/sidecar/censor_test.go
```diff
@@ -0,0 +1,141 @@
...
+			secrets:   []string{"younger", "my"},
+			output:    "In ** ******* and more vulnerable years ** father gave me some advice that I’ve been turning over ",
+			frameSize: 16,
+		},
+	}
+	for _, testCase := range testCases {
+		t.Run(testCase.name, func(t *testing.T) {
+			censorer := secretutil.NewCensorer()
+			censorer.Refresh(testCase.secrets...)
+			input := ioutil.NopCloser(bytes.NewBufferString(testCase.input))
```
alvaroaleman, line 65: Can you also verify that `Close` is being called?

prow/sidecar/censor.go
```diff
@@ -0,0 +1,206 @@
...
+	go func() {
+		for err := range errors {
+			errs = append(errs, err)
+		}
+	}()
+
+	secrets, err := loadSecrets(o.Secrets)
+	if err != nil {
+		return fmt.Errorf("could not load secrets: %w", err)
+	}
```
alvaroaleman, line 62: Lets just return here when `len(secrets) == 0`. The whole thing needs featuregating and that seems like a good way to do it. It also makes sense to move all the setup like error reading, semaphore, concurrency etc below this.

prow/sidecar/options.go
```diff
@@ -75,6 +75,16 @@ type Options struct {
 	// longer than the `entrypoint` grace period for the test process and the time
 	// taken by `sidecar` to upload all relevant artifacts.
 	IgnoreInterrupts bool `json:"ignore_interrupts,omitempty"`
+
+	// Secrets are paths to directories containing secret data. The contents of these
+	// secret data files will be censored from the logs and artifacts uploaded to the
+	// cloud.
+	Secrets []string `json:"secrets,omitempty"`
```
alvaroaleman, line 82: Maybe `SecretDirectories`? This requires reading the godoc to understand what is supposed to get in there

prow/testutil/fixtures.go
```diff
@@ -0,0 +1,84 @@
...
+		if err := ioutil.WriteFile(golden, actual, 0644); err != nil {
+			t.Fatalf("failed to write updated fixture: %v", err)
+		}
+	}
+	expected, err := ioutil.ReadFile(golden)
+	if err != nil {
+		t.Fatalf("failed to read testdata file: %v", err)
+	}
+
+	diff := difflib.UnifiedDiff{
```
alvaroaleman, line 69: nit: Why not `cmp.Diff`?

prow/sidecar/censor.go
```diff
@@ -0,0 +1,206 @@
...
+	if o.CensoringConcurrency == nil {
+		concurrency = int64(runtime.GOMAXPROCS(0))
+	} else {
+		concurrency = *o.CensoringConcurrency
+	}
+	sem := semaphore.NewWeighted(concurrency)
+	wg := &sync.WaitGroup{}
+	errors := make(chan error)
+	var errs []error
+	go func() {
```
alvaroaleman, line 53: You can address the race of this with the return by using the WaitGroup

prow/sidecar/censor.go
stevekuznetsov: I don't even call this if someone does not pass secrets to the `sidecar`.

prow/testutil/fixtures.go
stevekuznetsov: Copied your code from `ci-tools` ;)

prow/sidecar/censor_test.go
stevekuznetsov: Not trivially, no, I don't think I can. Why?

stevekuznetsov:
@alvaroaleman updated

stevekuznetsov:
/hold

I will squash after review

alvaroaleman (approved):

alvaroaleman (approved):

PR #21165, opened 2021-03-04
Prow component autobumper: also include prow images in prow jobs

chaodaiG:
They will be restricted to only bump gcr.io/k8s-prow/ and gcr.io/k8s-staging-boskos/

Part of https://github.com/kubernetes/test-infra/issues/21137


spiffxp (approved):
/approve
/lgtm

PR #21170, opened 2021-03-05
Add kube-state-metrics service in prow cluster config directory

chaodaiG:
Inspired by https://github.com/kubernetes/test-infra/issues/21090#issuecomment-790651066, adding kube-state-metrics so that we can capture prow pods crashlooping metrics, which will allow us set prometheus alert for issues like https://github.com/kubernetes/test-infra/issues/21090

Note: merging of this PR will not result it deployed in prow yet, will do it in a separate PR

### DIFF
--- a/config/prow/cluster/kube-state-metrics_deployment.yaml
+++ b/config/prow/cluster/kube-state-metrics_deployment.yaml
@@ -1,0 +1,46 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  labels:
+    app.kubernetes.io/name: kube-state-metrics
+    app.kubernetes.io/version: 2.0.0-beta
+  name: kube-state-metrics
+  namespace: kube-system
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app.kubernetes.io/name: kube-state-metrics
+  template:
+    metadata:
+      labels:
+        app.kubernetes.io/name: kube-state-metrics
+        app.kubernetes.io/version: 2.0.0-beta
+    spec:
+      containers:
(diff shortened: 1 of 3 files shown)


chaodaiG:
/cc @alvaroaleman @cjwagner 
CC @BenTheElder 

@alvaroaleman , I didn't find the deployment file from https://github.com/openshift/release, so created this based on instructions from https://github.com/kubernetes/kube-state-metrics, please take a look to see if there is anything suspicious. Thanks!

alvaroaleman (approved):
Looks good so far, you won't find a manifest in our repo because openshift itself sets up k-s-m :)
Only thing missing seems to be prometheus configuration to actually scrape it, but I am not sure how that is handled here?

Holding in case you want to control when its getting rolled out or more feedback, feel free to unhold
/hold

chaodaiG:
@cjwagner could you also take a look please?

cjwagner (approved):
This LGTM, but I think this comment is incorrect:
> Note: merging of this PR will not result it deployed in prow yet, will do it in a separate PR

Isn't everything under config/prow/cluster automatically deployed?

Also, as Alvaro noted, we'll need a servicemonitor resource as well.
/hold

chaodaiG:
> Isn't everything under config/prow/cluster automatically deployed?

As far as I know only components explicitly listed here are automatically deployed https://github.com/kubernetes/test-infra/blob/8d98c7dbeb781332c49d84532344e786dc06a102/config/prow/cluster/BUILD.bazel#L38

> Also, as Alvaro noted, we'll need a servicemonitor resource as well.

Yes will do it in a separate PR

PR #21181, opened 2021-03-05
Use built binaries in pull-kubernetes-e2e-gce-kubetest2

amwat:
xref: https://github.com/kubernetes/enhancements/issues/2464
/cc @BenTheElder @spiffxp


spiffxp (approved):
/approve
/lgtm
For followup at some point, I think not having "optional: true" means this will be a blocking job for prs where this is manually triggered

PR #21186, opened 2021-03-05
Fix the testgrid url for Kyma dashboards 

adamwalach:
We're testing to fix our dashboard. Unless this is connected to the issue with failing podinfo.json hopefully this will work.


tehcyx:
Missing descritipion: We're testing to fix our dashboard. Unless this is connected to the issue with failing podinfo.json hopefully this will work.

Issue #21190, opened 2021-03-05
Coverage Job in constant fail state due to go_test


MushuEE:
**What happened**:
The `ci-kubernetes-coverage-unit` job executes coverage for all `*_test.go` under k8s root. If a test fails, the whole job fails. There are several tests I see failing: [Job](https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/ci-kubernetes-coverage-unit/1367751836915732480)

I plan to 1) Always pass the coverage Job, within the TestGrid results we should see "pass/fail" coverage data. That is meant to be actionable. 

2) I will look into the test failures and see where they originate from and if there is work to be done on stability. 

**What you expected to happen**:
Coverage should surface coverage numbers and not corrupt Triage results with mass failures.

**How to reproduce it (as minimally and precisely as possible)**:
Run `make test KUBE_COVER=y KUBE_COVER_REPORT_DIR="${ARTIFACTS}" KUBE_TIMEOUT="--timeout=300s"` within k8s and see if tests fail.

**Please provide links to example occurrences, if any**:

**Anything else we need to know?**:

/area metrics
/sig testing
/cc @liggitt @spiffxp

BenTheElder:
> Coverage should surface coverage numbers and not corrupt Triage results with mass failures.

why is triage corrupted by mass failures? are tests failing more often under coverage?
shouldn't we fix the tests? or exclude the junit from triage?

MushuEE:
I had blanked that we have previously seen this and these jobs should be excluded but are not, making a new issues for the kettle exclude failure.

PR #21191, opened 2021-03-05
Remove ci-coverage from test results

MushuEE:
Coverage will fail if any test fails #21190

The coverage job is designed to report coverage, it should not be forced to fail if tests under it fail.

### DIFF
--- a/triage/update_summaries.sh
+++ b/triage/update_summaries.sh
@@ -42,7 +42,8 @@ bq --headless --format=json query --max_rows 1000000 \
   from
     [k8s-gubernator:build.all]
   where
-    timestamp_to_sec(started) > TIMESTAMP_TO_SEC(DATE_ADD(CURRENT_DATE(), -14, 'DAY'))" \
+    timestamp_to_sec(started) > TIMESTAMP_TO_SEC(DATE_ADD(CURRENT_DATE(), -14, 'DAY'))
+    and job != 'ci-kubernetes-coverage-unit'" \
   > triage_builds.json
 
 bq query --allow_large_results --headless --max_rows 0 --replace --destination_table k8s-gubernator:temp.triage \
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes/sig-testing/coverage.yaml
```diff
@@ -148,7 +148,7 @@ periodics:
       - |
         result=0
         ( cd hack/tools && GO111MODULE=on go install gotest.tools/gotestsum ) || result=$?
-        make test KUBE_COVER=y KUBE_COVER_REPORT_DIR="${ARTIFACTS}" KUBE_TIMEOUT="--timeout=300s" || result=$?
```
BenTheElder, line 151: I'm not sure I agree with this. cc @liggitt

liggitt: The job is solid red currently, and not driving any action. if there's value in getting coverage results even if all tests don't pass, this makes sense. I don't have context to know how the results are being used here to know.

MushuEE: The way we actually "use coverage" is that TestGrid shows red/green for high and low areas of coverage. Our tests are too flakey to rely on them all passing on each run.

BenTheElder: https://github.com/kubernetes/test-infra/issues/21190#issuecomment-791599161

BenTheElder: IIRC triage will scrape the junit results, rather than the job status? So changing the exit code does not exclude it.
But we can exclude jobs from the scrape.

MushuEE: *Ben just reminded me that I also need to change the threshold or exclude the job from Kettle

BenTheElder: https://github.com/kubernetes/test-infra/pull/19038
https://github.com/kubernetes/test-infra/issues/18480

BenTheElder: we should add it to the "garbage jobs" list in triage, but continue to report failures.
ideally we should get the job healthy and out of that list, but it makes sense to add it for now.

MushuEE: I will build a better mechanism for excluding jobs in Triage.

MushuEE:
Created and issue for the exclude jobs not working

BenTheElder:
I think this is OK but want-lgtm=spiffxp

spiffxp (approved):
/approve
/lgtm
If we get to N=3 jobs we should avoid the copy-paste

Issue #21138, opened 2021-03-03
Make #prow-alerts less noisy


spiffxp:
> /cc @spiffxp @cjwagner @stevekuznetsov 
> 
> 
> 
> The current change proposes to send these to `prow-alerts` channel instead of `testing-ops` to avoid spamming the latter and make it less valuable. However currently `prow-alerts` is pretty spammy, are we sure the current alerts are looked at?

I think we should vastly increase the threshold for or remove the ghproxy alerts entirely. Informational alerts are good, but those seem to be pure noise that be adequately tracked via ghproxy graphs

But I don't think that precludes adding these to that channel

__Originally posted by @spiffxp in https://github.com/kubernetes/test-infra/pull/21136#issuecomment-789999648__

spiffxp:
https://github.com/kubernetes/test-infra/pull/21192 merged, need to see if these changes are live yet. The channel still has a bunch of ghproxy alerts

chaodaiG:
The changes from https://github.com/kubernetes/test-infra/pull/21192 are not live yet, the config from the PR was applied, but somehow it didn't land on prometheus alerter, so it's still using the old configmap. The weird thing is that `kind: PrometheusRule` in the cluster was updated with the new values from #21192, but the configmap `prometheus-prow-rulefiles-0` which was mounted on prometheus pod was not updated (It should have been generated from `PrometheusRule`.)

My suspicion was that previous upgrade https://github.com/kubernetes/test-infra/pull/20456 and the revert https://github.com/kubernetes/test-infra/pull/20457 had messed up something. Planning to roll forward with https://github.com/kubernetes/test-infra/pull/20484 to see if this can be fixed

cc @cjwagner for awareness

chaodaiG:
There are recurring errors in prometheus operator:
```
E0309 15:27:16.435071       1 reflector.go:134] github.com/coreos/prometheus-operator/pkg/alertmanager/operator.go:237: Failed to list *v1beta2.StatefulSet: the server could not find the requested resource
```

v1beta2 was deprecated in k8s v1.16 per https://www.pulumi.com/docs/reference/pkg/kubernetes/apps/v1beta2/statefulset/#statefulset, so actually it has nothing to do with the previous upgrade and revert, and I think the correct fix is actually pull prometheus operator forward to make it compatible with k8s v1.16

chaodaiG:
This was done
/close

PR #21192, opened 2021-03-05
Relax ghproxy alert rules by extending duration of status code starting with 4 or 5

chaodaiG:
Part of: #21138 

Currently prow-alerts channel is pretty much filled with warnings from ghproxy, they might provide some value but it's so spammy that if any other useful alerts were surfaced they would easily be overlooked.

Simply inspecting the warnings, majority of them were 422 and 405 codes. And some of them were actually expected, for example:

Expecting 422 from https://github.com/kubernetes/test-infra/blob/7f3fe6c5ffbd8a665d68e38b6ddbd865e06716a6/prow/github/client.go#L2707
Expecting 405 from https://github.com/kubernetes/test-infra/blob/7f3fe6c5ffbd8a665d68e38b6ddbd865e06716a6/prow/github/client.go#L3472

Extending to 30 minutes and hopefully it would dilute the impact enough to eliminate the noise. If this doesn't work will need to figure out how to correctly set warnings for these two status codes

### DIFF
--- a/config/prow/cluster/monitoring/mixins/prometheus/ghproxy_alerts.libsonnet
+++ b/config/prow/cluster/monitoring/mixins/prometheus/ghproxy_alerts.libsonnet
@@ -7,9 +7,9 @@
         name: 'ghproxy',
         rules: [
           {
-            alert: 'ghproxy-specific-status-code-abnormal',
+            alert: 'ghproxy-specific-status-code-5xx',
             expr: |||
-              sum(rate(github_request_duration_count{status=~"[45]..",status!="404",status!="410"}[30m])) by (status,path) / ignoring(status) group_left sum(rate(github_request_duration_count[5m])) by (path) * 100 > 10
+              sum(rate(github_request_duration_count{status=~"5.."}[5m])) by (status,path) / ignoring(status) group_left sum(rate(github_request_duration_count[5m])) by (path) * 100 > 10
             |||,
             labels: {
               severity: 'warning',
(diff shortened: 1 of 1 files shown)


config/prow/cluster/monitoring/mixins/prometheus/ghproxy_alerts.libsonnet
```diff
@@ -9,7 +9,7 @@
           {
             alert: 'ghproxy-specific-status-code-abnormal',
             expr: |||
-              sum(rate(github_request_duration_count{status=~"[45]..",status!="404",status!="410"}[5m])) by (status,path) / ignoring(status) group_left sum(rate(github_request_duration_count[5m])) by (path) * 100 > 10
+              sum(rate(github_request_duration_count{status=~"[45]..",status!="404",status!="410"}[30m])) by (status,path) / ignoring(status) group_left sum(rate(github_request_duration_count[5m])) by (path) * 100 > 10
```
cjwagner, line 12: I skimmed through some of the #prow-alerts channel and didn't see any 5xx alerts.  WDYT about keeping the tighter alerting window for 5xx and only loosening 4xx?
5mins to 30mins is a pretty big change and delays the alerts for so long that they we'd be likely to notice serious problems ourselves before the alert fires. Could we choose a smaller time window if we also increased the error threshold to more than 10%?

chaodaiG: good point about 5xx, updated to keep 5xx still alert with 5m window events.

I have seen 100% for example https://kubernetes.slack.com/archives/CQP3RC68Y/p1614920511052900, which imo would be pretty hard to brought down to 20% even with doubled time, so I think diluting it out is more effective, still finger crossed on this though

config/prow/cluster/monitoring/mixins/prometheus/ghproxy_alerts.libsonnet
```diff
@@ -19,9 +19,33 @@
...
+              severity: 'warning',
+            },
+            annotations: {
+              message: '{{ $value | humanize }}%% of all API requests through the GitHub proxy are errorring with code {{ $labels.status }}. Check %s.' % [monitoringLink('/d/%s/github-cache?orgId=1&refresh=1m&fullscreen&panelId=8' % [dashboardID], 'the ghproxy dashboard')],
+            },
+          },
+          {
+            alert: 'ghproxy-specific-status-code-4xx',
+            expr: |||
+              sum(rate(github_request_duration_count{status=~"4..",status!="404",status!="410"}[30m])) by (status,path) / ignoring(status) group_left sum(rate(github_request_duration_count[5m])) by (path) * 100 > 10
```
cjwagner, line 36: The second time interval needs to be updated to `30m` as well. Same with the other 4xx alert.

chaodaiG: good catch, done

cjwagner (approved):

Issue #21027, opened 2021-02-25
Presubmit test-infra bazel test: reduce flakiness caused by image pull failures


chaodaiG:
**What should be cleaned up or changed**:

It has been lately observed that presubmit bazel test has been pretty flaky on image pulling failures from code paths that are totally unrelated with the PR. Would be nice to reduce the flakiness by:

- Understand why bazel test needs to pull images, whether it's possible to avoid the pulling.
- If not possible, is there any chance to cache the image to make it less flaky?

**Provide any links for context**:

From PR: https://github.com/kubernetes/test-infra/pull/21026
- https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/test-infra/21026/pull-test-infra-bazel/1364984456091275264
- https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/test-infra/21026/pull-test-infra-bazel/1364983366922801152

From PR: https://github.com/kubernetes/test-infra/pull/20943
- https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/test-infra/20943/pull-test-infra-bazel/1364685688871587840
- https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/test-infra/20943/pull-test-infra-bazel/1364684891010109440
- https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/test-infra/20943/pull-test-infra-bazel/1364687057552674816

chaodaiG:
CC @alvaroaleman

BenTheElder:
> Understand why bazel test needs to pull images, whether it's possible to avoid the pulling.

IIRC: `WORKSPACE` depencencies are resolved before doing any computation of the targets. Container images are workspace dependencies.

> If not possible, is there any chance to cache the image to make it less flaky?

Bazel doesn't support remote cache for this (`WORKSPACE`) using HTTP remote cache or RBE. It could be done by persisting the local `~/.cache/bazel`, though that can grow unbounded (normally a user might periodically delete it with e.g. `bazel clean --expunge`) and you'll need a solution for persisting it.

IIRC one of the issues is that these container pulls do not retry.

alvaroaleman:
https://github.com/bazelbuild/rules_docker/issues/673 is the so far ignored bazel issue for this

BenTheElder:
[...]

PR #21195, opened 2021-03-05
retry fetching bazel dependencies

BenTheElder:
This is basically https://github.com/kubernetes/test-infra/pull/11605 but for pull not push

Fixes https://github.com/kubernetes/test-infra/issues/21027

https://docs.bazel.build/versions/master/external.html#fetching-dependencies

### DIFF
--- a/hack/bazel.sh
+++ b/hack/bazel.sh
@@ -20,6 +20,19 @@ set -o errexit
 set -o pipefail
 
 code=0
+# Remove retries after
+# https://github.com/bazelbuild/bazel/issues/12599
+if [ -n "${BAZEL_FETCH_PLEASE:-}" ]; then
+  for i in {1..3}; do
+    if (set -o xtrace && bazel fetch "${BAZEL_FETCH_PLEASE}"); then
+      break
+    elif [[ "$i" == 3 ]]; then
+      echo "Failed"
+      exit 1
+    fi
+    echo "Failed fetch attempt $i, retrying ..."
+  done
+fi
 (set -o xtrace && bazel "$@") || code=$?
 coalesce=$(dirname "${BASH_SOURCE[0]}")/coalesce.py
 (set -o xtrace && "$coalesce") || true
--- a/config/jobs/kubernetes/test-infra/test-infra-periodics.yaml
+++ b/config/jobs/kubernetes/test-infra/test-infra-periodics.yaml
@@ -19,6 +19,9 @@ periodics:
       - --config=ci
       - --nobuild_tests_only
       - //...
+      env:
+      - name: BAZEL_FETCH_PLEASE
+        value: //...
   annotations:
     testgrid-dashboards: sig-testing-misc
     testgrid-tab-name: ci-bazel
--- a/config/jobs/kubernetes/test-infra/test-infra-postsubmits.yaml
+++ b/config/jobs/kubernetes/test-infra/test-infra-postsubmits.yaml
@@ -17,6 +17,9 @@ postsubmits:
         - --config=ci
         - --nobuild_tests_only
         - //...
+        env:
+        - name: BAZEL_FETCH_PLEASE
+          value: //...
     annotations:
       testgrid-dashboards: sig-testing-misc
       testgrid-tab-name: post-bazel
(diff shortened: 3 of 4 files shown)


hack/bazel.sh
```diff
@@ -20,6 +20,20 @@ set -o errexit
 set -o pipefail
 
 code=0
+# Remove retries after
+# https://github.com/bazelbuild/bazel/issues/12599
+BAZEL_SYNC_PLEASE='true'
+if [ "${BAZEL_SYNC_PLEASE:-}" = 'true' ]; then
+  for i in {1..3}; do
+    if (set -o xtrace && bazel sync); then
```
BenTheElder, line 28: note: this fetches all deps. we could instead `bazel fetch <target pattern>` but then we'd need to supply just the target pattern and keep that in sync (flags to other commands are not compatible with this one)

hack/bazel.sh
```diff
@@ -20,6 +20,20 @@ set -o errexit
 set -o pipefail
 
 code=0
+# Remove retries after
+# https://github.com/bazelbuild/bazel/issues/12599
+BAZEL_SYNC_PLEASE='true'
```
BenTheElder, line 25: forcing on test in this PR temporarily.
final PR would opt-in specific jobs to this initially (test-infra) since it's used in multiple places

BenTheElder:
[running repeatedly to collect data, these passed]
/test pull-test-infra-bazel

BenTheElder:
[running repeatedly to collect data, these passed]
/test pull-test-infra-bazel

BenTheElder:
[running repeatedly to collect data, these passed]
/test pull-test-infra-bazel

BenTheElder:
These have pretty comparable runtimes as before (as expected), because bazel is smart enough to recognize that we've already populated the dependencies after the `bazel sync` call and not bother trying to obtain them again once we move on to the `bazel test` call.

Given that we pretty much always use `//...` (all targets) in CI, `bazel sync` is suitable (ensure all dependencies). 

One more test run, then I'll drop the temp commit and add a commit to enable this in test-infra jobs.

/test pull-test-infra-bazel

hack/bazel.sh
BenTheElder: also: in test-infra our build/test target is `//...` (all targets) and this is generally true, so it's pretty unlikely to be worth the complexity of fetch instead of sync.

fejta: Given that this is behind a flag anyway, why not `bazel fetch $BAZEL_SYNC_PLEASE` and change it to `BAZEL_SYNC_PLEASE=//...` instead of `=true`?

fejta: Things like pushing images, deploying k8s etc can suffer these types of flakes too, will benefit from this retry pattern but are scoped to something smaller than `//...`

BenTheElder: > Given that this is behind a flag anyway, why not `bazel fetch $BAZEL_SYNC_PLEASE` and change it to BAZEL_SYNC_PLEASE=//... instead of =true?

This SGTM, I was initially concerned about:
- Two sources of truth for the targets 
  - We're going to fallback to ensuring whatever dependencies we need in the `bazel test` / `bazel build` call anyhow, so not a _big_ deal, just another place to try to keep this up to date
- Target patterns can be like `//... -//foo` (multiple args / an array). If bazel doesn't allow spaces in targets (seems likely actually?) then we can just space/word split the string.

Agree that this is desirable for smaller targets, my initial perusal found `//...` in use.
I'll verify patterns and no-spaces support and rework this.

BenTheElder: Does not appear that spaces are valid in targets so splitting should be fine.

https://docs.bazel.build/versions/master/build-ref.html#name

> Target names must be composed entirely of characters drawn from the set a–z, A–Z, 0–9, and the punctuation symbols !%-@^_` "#$&'()*-+,;<=>?[]{|}~/.

While target patterns appear to be a superset consisting ~only of adding the subtractive targets (preceding `-`), and the `...` wildcarding.

https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/runtime/commands/target-syntax.txt

BenTheElder: even better: we don't need to split at all. `bazel fetch '//... -//experiment/...'` turns out to be perfectly valid.

BenTheElder: done.

fejta (approved):
/hold

BenTheElder:
/hold cancel
will keep an eye on this and see about rolling it out to other use cases

PR #21198, opened 2021-03-05
prow: Add ignore-space-change flag to git am

sebsoto:
This commit adds the ignore-space-change flag to git am calls in the git
client. This is being done as certain patches are not able to applied.
This is unexpected behavior as the commits the patches have been created
from are able to be cherry-picked. This behavior may be caused by
carriage returns being present in the contents of the patch.

Signed-off-by: Sebastian Soto <ssoto@redhat.com>

### DIFF
--- a/prow/git/git.go
+++ b/prow/git/git.go
@@ -384,7 +384,7 @@ func (r *Repo) MergeAndCheckout(baseSHA string, mergeStrategy prowgithub.PullReq
 // an error if the patch cannot be applied.
 func (r *Repo) Am(path string) error {
 	r.logger.Infof("Applying %s.", path)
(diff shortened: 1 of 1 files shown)


petr-muller:
Hrm, the code is obviously fine, but I wonder if this is always safe to do (if it were, it would be git default?) and does not result in something pathological in some cases?

Personally I think the automation should not try to be smart and add such relaxing parameters.

aravindhp:
@petr-muller @sebsoto could using `--ignore-space-change` be considered a fall back option if the first patch attempt fails?

petr-muller:
@aravindhp that sounds good to me

PR #21201, opened 2021-03-06
sidecar: censor files inside of archives

stevekuznetsov:
Now, `sidecar` will transparently unarchive artifacts where necessary,
censoring the files they contain before re-archiving them prior to
upload.

Two things of note:
- frustratingly, Go does not have high-level utilities for unarchiving
  and archiving files. There are hundreds of examples of people copying
  an implmenetation that exists in an internal package of the standard
  library. While the implementation seems to handle the one obvious CVE
  related to unzipping, there may be more. There doesn't seem to be an
  obvious third-party library to use for this, either.
- comparing two tarballs is not trivial as they encode metadata about
  the files they contain. In a perfect world, the input and output
  directories in the test data for the integration test would simply
  contain tarballs. However, as we're unarchiving and rearchiving things
  with abandon during this process, it is really tricky to get
  everything identical about the metadata, and it's not really important
  for the actual use-case of censoring artifacts. Handling all the edge
  cases to make the tests nice seemed like a bad investment of time.
  Similarly, it would have required a more intelligent diff approach
  anyway as nobody wants to see the diff between two binary files. By
  archiving before running the tests and unarchiving after, we keep the
  test and comparison lightweight while allowing UPDATE=true to work.

/assign @alvaroaleman @petr-muller @cjwagner


stevekuznetsov:
Hmmm ... as written, I guess this only actually tests that my `unarchive()` is an inverse of `archive()` ... however I got to this state by first trying to commit as test data real `.tar.gz` files, when I ran into the myriad ways two tarballs may have files inside with identical content but be different byte-for byte ... thoughts on how to best test if this is not acceptable appreciated.

prow/sidecar/censor.go
```diff
@@ -112,6 +116,204 @@ func (o Options) censor() error {
...
+		case mode.IsDir():
+			if err := os.MkdirAll(abs, 0755); err != nil {
+				return fmt.Errorf("could not create directory while unpacking archive: %w", err)
+			}
+		case mode.IsRegular():
+			file, err := os.OpenFile(abs, os.O_RDWR|os.O_CREATE|os.O_TRUNC, mode.Perm())
+			if err != nil {
+				return err
+			}
+			n, err := io.Copy(file, tarReader)
```
alvaroaleman, line 237: Any idea how the memory consumption of this looks like? Depending on that it might be worth parallellizing, since CPU is a compressible resource so grabbing as much as we can helps. Also I imagine it includes a lot of syscall, which would be another good reason to parallelize

prow/sidecar/censor.go
```diff
@@ -112,6 +116,204 @@ func (o Options) censor() error {
...
+			if closeErr := file.Close(); closeErr != nil && err == nil {
+				err = closeErr
```
alvaroaleman, line 239: Lets just use an error slice and utilerrors, potentially swallowing errors seems like it might result in hard to debug issues

prow/sidecar/censor_test.go
```diff
@@ -122,6 +122,14 @@ func TestCensorIntegration(t *testing.T) {
 	}); err != nil {
 		t.Fatalf("failed to copy input to temp dir: %v", err)
 	}
+	// also, tar the input - it's not trivial to diff two tarballs while only caring about
+	// file content, not metadata, so this test will tar up the archive from the input and
+	// untar it after the fact for simple diffs and updates
+	archiveDir := filepath.Join(tempDir, "artifacts/archive")
```
alvaroaleman, line 128: Can we shell out to `tar` for this so we can be confident that our untargzip/targzip implementation is correct?

prow/sidecar/censor.go
```diff
@@ -71,35 +74,36 @@ func (o Options) censor() error {
 	if largest := censorer.LargestSecret(); 2*largest > bufferSize {
 		bufferSize = 2 * largest
 	}
+	censorFile := fileCensorer(sem, wg, errors, censorer, bufferSize)
+	censor := func(file string) {
+		wg.Add(1)
+		go censorFile(file)
```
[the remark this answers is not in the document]
stevekuznetsov: Yeah, I went back and forth on this, would love to hear your opinion:

1. I originally had the closure do the add as well, but we need access to the goroutine itself for the control flow management in the archival code
2. I went back and forth on bringing this out into a named function at all, since it's minimal, I just had N places where we did the exact same thing - wait group, semaphore, file handler - and knew that doing it wrong in any one case would be easy, so decided to DRY.

What do you think is the better option?

prow/sidecar/censor.go
```diff
@@ -112,6 +116,204 @@ func (o Options) censor() error {
...
+		return fmt.Errorf("could not unpack archive: %w", err)
+	}
+
+	children := &sync.WaitGroup{}
+	if err := filepath.Walk(outputDir, func(absPath string, info os.FileInfo, err error) error {
+		if info.IsDir() {
+			return nil
+		}
+
+		wg.Add(1)
```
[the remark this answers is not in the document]
stevekuznetsov: I guess implicitly, sure, but at the same time I wanted all file handling goroutines to be part of the same worker pool - with the same group and the same semaphore/

prow/sidecar/censor.go
stevekuznetsov: IIRC the internal buffers are pretty small for `io.Copy()` between two `os.File` - so it's efficient in space. Yes, we could make the file writes out parallel

prow/sidecar/censor.go
stevekuznetsov: Good catch, this was copy pasta, can change it

prow/sidecar/censor_test.go
stevekuznetsov: Yeah, that was a thought.

stevekuznetsov:
@alvaroaleman updated to add tests that verify `archive()` and `unarchive()` properly function w.r.t. `tar` and also tests for round-tripping. I did *not* change the central integration test to use tarballs, since the diff on those if it's wrong is awful to look at, and the bugs from that should be in the content of the files, not the headers in the archive. I did not make unpacking parallel, it might be premature?

stevekuznetsov:
Only fails in `bazel` on the CI server wooo

stevekuznetsov:
OK, had to set (permissive) permissions hard-coded in the archives, too ... likely not an issue for this use-case?

stevekuznetsov:
That seems to have done it. @alvaroaleman ready for review

prow/sidecar/censor.go
```diff
@@ -112,6 +116,218 @@ func (o Options) censor() error {
...
+
+	if err := filepath.Walk(srcDir, func(absPath string, info os.FileInfo, err error) error {
+		header, err := tar.FileInfoHeader(info, info.Name())
+		if err != nil {
+			return fmt.Errorf("could not create tar header: %w", err)
+		}
+		// the header won't get nested paths right
+		relpath, _ := filepath.Rel(srcDir, absPath) // err happens when there's no rel path, but we know there must be
+		header.Name = relpath
+		// in order to get reproducible output archives, we need to remove variance
```
stevekuznetsov, line 286: IMHO this section is not contentious, but it *does* change the archive as given by the user. I claim it's not contentious as the modification time, user bits etc for an artifact are not really that interesting. Who cares if `g+w` on the file in GCS?

prow/sidecar/censor.go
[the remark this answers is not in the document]
alvaroaleman: I think the new approach of doing both the wg.Add and defer in `fileCensorer` is perfect

prow/sidecar/censor.go
```diff
@@ -112,6 +116,218 @@ func (o Options) censor() error {
 	return kerrors.NewAggregate(errs)
 }
 
+// fileCensorer returns a closure over all of our synchronization for a clean handler signature
+func fileCensorer(sem *semaphore.Weighted, errors chan<- error, censorer secretutil.Censorer, bufferSize int) func(wg *sync.WaitGroup, file string) {
+	return func(wg *sync.WaitGroup, file string) {
+		wg.Add(1)
+		go func() {
+			if err := sem.Acquire(context.Background(), 1); err != nil {
```
alvaroaleman, line 124: Not related to this PR but I forgot to mention that before: Instead of defaulting the concurrency to the number of cores, can we default it to some fixed const (4? 8?) so we do not have a different memory profile depending on how many cores the node we get scheduled to has?

prow/sidecar/censor.go
alvaroaleman: yeah doesn't really matter but why does this need to be reproducible? I think for tests it would be easiest to just let `tar` create an archive from some data we provide, then use that. That most closely mirrors reality IMHO

stevekuznetsov: So only test that `archive()` creates an archive that `tar` can extract into the original input, but test nothing about the actual shape of the archive?

alvaroaleman: What is important here is IMHO a) We can cope with archives created by tar b) tar can cope with archives created by us. I don't think it is super important to check that the archive we create looks the same as one tar creates plus such a test might be pretty brittle and e.G. not work on different tar versions because some default changed or whatnot

stevekuznetsov: Sounds good to me

prow/sidecar/censor.go
stevekuznetsov: The pattern pretty much everywhere has benn to default to max cores, I don't really have a strong opinion other than it's likely better to be self-consistent with all the other places we do this same thing.

stevekuznetsov:
@alvaroaleman removed the file metadata mangling and removed the test for the shape of the `.tar.gz` archive, retained tests for `archive()` and `tar`

prow/sidecar/censor.go
alvaroaleman: Yeah but the reason for that is usually to limit file descriptors or such. Here, the reason is to have a limited and predictable memory profile. We need that to be able to set proper requests. We can not autochange requests based on the number of cores on the node a given pod gets scheduled to.

stevekuznetsov: Makes sense.

alvaroaleman (approved):

PR #21206, opened 2021-03-07
report pull-kubernetes-node-e2e-containerd to github

dims:
Not sure why we switched this off. it's optional and does not run unless triggered, we should at least let this show up in GH so the link to the runs are easy to find.


BenTheElder (approved):
It can be confusing to contributors (who don't understand Required vs not) so we tend to do this for jobs that will be always run / required but, particularly for a totally manually invoked job it's fine.

PR #21210, opened 2021-03-07
kOps - Always use latest images in periodic tests

hakman:
/cc @rifelpet


config/jobs/kubernetes/kops/build_grid.py
```diff
@@ -141,37 +143,68 @@ def should_skip_newer_k8s(k8s_version, kops_version):
...
+    'rhel8': latest_aws_image('309956199498', 'RHEL-8.*_HVM-*-x86_64-0-Hourly2-GP2'),
+    'u1804': latest_aws_image('099720109477', 'ubuntu/images/hvm-ssd/ubuntu-bionic-18.04-amd64-server-*'), # pylint: disable=line-too-long
+    'u2004': latest_aws_image('099720109477', 'ubuntu/images/hvm-ssd/ubuntu-focal-20.04-amd64-server-*'), # pylint: disable=line-too-long
+}
+
 def distro_info(distro):
     if distro == 'amzn2':
         kops_ssh_user = 'ec2-user'
-        kops_image = '137112412989/amzn2-ami-hvm-2.0.20201126.0-x86_64-gp2'
+        kops_image = distro_images[distro]
```
rifelpet, line 180: can we just consolidate all of these to a single `kops_image = distro_images[distro]` at the top of this function?

rifelpet: in fact we could add the ssh user mapping to distro_images and then get rid of this function entirely

hakman: I think this last commit should do it :).

config/jobs/kubernetes/kops/build_grid.py
```diff
@@ -206,7 +217,13 @@ def build_test(cloud='aws',
     if should_skip_newer_k8s(k8s_version, kops_version):
         return None
 
-    kops_image, kops_ssh_user = distro_info(distro)
+    kops_image = distro_images[distro]
+    if kops_image is None:
+        raise Exception(f"missing required distro image for {distro}")
```
rifelpet, line 222: This code will never be reached because `distro_images[distro]` will raise a KeyError if the key doesn't exist.  I think the generic KeyError is fine and we dont need to make error message any friendlier.

Same with the ssh user exception below.

hakman: My python is a bit rusty...

rifelpet (approved):

PR #21213, opened 2021-03-07
[pull-kubernetes-node-e2e-containerd] capture additional logs related to containerd

dims:
There is a containerd-installation service as well that we need to pull logs from.


nikhita (approved):
/lgtm

PR #21214, opened 2021-03-07
wg-k8s-infra: add pull-k8s.io-verify job

nikhita:
For https://github.com/kubernetes/k8s.io/pull/1765

/hold
this should not merge until https://github.com/kubernetes/k8s.io/pull/1765 has merged

/assign @spiffxp


spiffxp (approved):
/approve
/lgtm
Let's find out if status reconciler does it's thing for the rename here

spiffxp:
@nikhita @ameukam doh, this broke merges to k8s.io since verify-terraform expects a cluster. A followup PR to exclude it by default here is probably quickest way to unblock

PR #21215, opened 2021-03-07
Autobumper: PR title handles cases where only inconsistent image tag is bumped

chaodaiG:


mpherman2:
/lgtm

Thanks for this change!

BenTheElder:
thanks :-)

PR #21218, opened 2021-03-08
Fix ci-containerd-build-1-3 and ci-containerd-build-1-2 to work with go 1.16

dims:
See failures in:

https://testgrid.k8s.io/sig-node-containerd#containerd-build-1.2
and https://testgrid.k8s.io/sig-node-containerd#containerd-build-1.3

show issues because go modules are switched on by default now as the
kubekins image is now at 1.16, so let's switch it off


nikhita (approved):
Makes sense :+1: 

/lgtm

PR #21229, opened 2021-03-08
Add gce pd driver tests for 2004 and 20h2 windows image

jingxu97:
Add gce pd driver e2e tests and migratino tests for 2004 and 20h2
windows image


jingxu97:
cc @mauriciopoppe

PR #21232, opened 2021-03-08
Forwarding runtime-config to remote.go

knabben:
Runtime-config is not being used on node e2e tests. It's required for API alpha enablement.
This should fix the failures of `ci-kubernetes-node-kubelet-alpha` tab.

NOTE: This one must come first - Node E2E config-runtime handler - https://github.com/kubernetes/kubernetes/pull/99960
xref https://github.com/kubernetes/kubernetes/issues/98220

### DIFF
--- a/kubetest/e2e.go
+++ b/kubetest/e2e.go
@@ -172,7 +172,7 @@ func run(deploy deployer, o options) error {
 		} else if o.nodeTests {
 			nodeArgs := strings.Fields(o.nodeArgs)
 			errs = util.AppendError(errs, control.XMLWrap(&suite, "Node Tests", func() error {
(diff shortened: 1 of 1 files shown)


amwat (commented):
This seems like a scoped enough change, which is a fix for existing tests failing.

/approve

/cc @spiffxp 
/lgtm
/hold

spiffxp (approved):
/approve
/lgtm
/hold cancel
Agreed. And dependent PR has merged

spiffxp:
@knabben is this intended only for v1.21 onward or is this something that needs to be backported to earlier release branches as well?

knabben:
@spiffxp this fix, in particular, is for a job running on master only.. in principle do not need to be backported

PR #21241, opened 2021-03-09
Fix Peribolos log by correctly parsing the OAuth scopes in the response header

chizhg:
The logic is wrong when Peribolos determines the `403` error reason by parsing the `OAuth` headers in the GitHub API response, which results in a misleading log message - https://prow.knative.dev/view/gs/knative-prow/logs/post-knative-peribolos/1369002331068174336#1:build-log.txt%3A9.

This PR will fix it and print out the response body when it's unable to determine the error from the headers. For this case, the response body is `{"message":"Blocked","documentation_url":"https://docs.github.com/rest/reference/orgs#set-organization-membership-for-a-user"}`

/cc @fejta @evankanderson

### DIFF
--- a/prow/github/client.go
+++ b/prow/github/client.go
@@ -883,12 +883,21 @@ func (c *client) requestRetry(method, path, accept, org string, body interface{}
 						resp.Body.Close()
 						break
 					}
-				} else if oauthScopes := resp.Header.Get("X-Accepted-OAuth-Scopes"); len(oauthScopes) > 0 {
+				} else {
+					acceptedScopes := resp.Header.Get("X-Accepted-OAuth-Scopes")
 					authorizedScopes := resp.Header.Get("X-OAuth-Scopes")
 					if authorizedScopes == "" {
 						authorizedScopes = "no"
 					}
-					err = fmt.Errorf("the account is using %s oauth scopes, please make sure you are using at least one of the following oauth scopes: %s", authorizedScopes, oauthScopes)
+
+					want := sets.NewString(strings.Split(acceptedScopes, ",")...)
+					got := strings.Split(authorizedScopes, ",")
+					if !want.HasAny(got...) {
+						err = fmt.Errorf("the account is using %s oauth scopes, please make sure you are using at least one of the following oauth scopes: %s", authorizedScopes, acceptedScopes)
+					} else {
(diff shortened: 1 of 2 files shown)


prow/github/client.go
```diff
@@ -883,12 +883,26 @@ func (c *client) requestRetry(method, path, accept, org string, body interface{}
...
-					err = fmt.Errorf("the account is using %s oauth scopes, please make sure you are using at least one of the following oauth scopes: %s", authorizedScopes, oauthScopes)
+					oauthScopeAccepted := false
+					for _, scope := range strings.Split(authorizedScopes, ",") {
+						if acceptedOAuthScopes.Has(scope) {
+							oauthScopeAccepted = true
+							break
+						}
+					}
+
+					if !oauthScopeAccepted {
```
evankanderson, line 900:
```suggestion
					if acceptedOauthScopes.HasAny(strings.Split(authorizedScopes, ",")...) {
```

evankanderson: I might also `s/acceptedOauthScopes/acceptedScopes/` or even just `want`.

prow/github/client.go
```diff
@@ -883,12 +883,26 @@ func (c *client) requestRetry(method, path, accept, org string, body interface{}
...
+						err = fmt.Errorf("the account is using %s oauth scopes, please make sure you are using at least one of the following oauth scopes: %s", authorizedScopes, acceptedOAuthScopes)
+					} else {
+						body, _ := ioutil.ReadAll(resp.Body)
```
evankanderson, line 903: It would be nice if there was a string in the error message we could key off, rather than needing to compare accepted & supplied scopes for that error.

evankanderson (commented):
/lgtm

With a couple suggestions/improvements.

prow/github/client.go
chizhg: Thanks, done.

petr-muller (approved):

fejta (approved):

fejta (commented):
Please add unit tests for this scenario
/hold

chizhg:
> Please add unit tests for this scenario
> /hold

Done.

fejta (approved):

PR #21244, opened 2021-03-09
kubernetes: run experimental unit test job as non-root user with latest image

BenTheElder:
https://github.com/kubernetes/kubernetes/issues/99881

### DIFF
--- a/config/jobs/kubernetes/sig-testing/make-test.yaml
+++ b/config/jobs/kubernetes/sig-testing/make-test.yaml
@@ -48,7 +48,12 @@ presubmits:
     path_alias: k8s.io/kubernetes
     spec:
       containers:
-        - image: gcr.io/k8s-testimages/kubekins-e2e:v20210302-a6bf478-master
+        - image: gcr.io/k8s-testimages/kubekins-e2e:latest-master
+          imagePullPolicy: Always
+          # let's see if we can run unit tests as a non-root user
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes/sig-testing/make-test.yaml
```diff
@@ -48,7 +48,12 @@ presubmits:
     path_alias: k8s.io/kubernetes
     spec:
       containers:
-        - image: gcr.io/k8s-testimages/kubekins-e2e:v20210302-a6bf478-master
+        - image: gcr.io/k8s-testimages/kubekins-e2e:latest-master
+          imagePullPolicy: always
+          # let's see if we can run unit tests as a non-root user
```
BenTheElder, line 53: (I know we can't but I'm not yet sure what else will break in CI, this job is optionally run so we can find out live and work through it)

spiffxp (approved):
/approve
/lgtm

config/jobs/kubernetes/sig-testing/make-test.yaml
BenTheElder: yeah ... there's a lot more besides the tests themselves to sort out https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/92316/pull-kubernetes-unit-experimental/1369159862277115904

PR #21246, opened 2021-03-09
kubernetes: iterate on experimental rootless unit test job

BenTheElder:
follow up to https://github.com/kubernetes/test-infra/pull/21244

### DIFF
--- a/config/jobs/kubernetes/sig-testing/make-test.yaml
+++ b/config/jobs/kubernetes/sig-testing/make-test.yaml
@@ -53,12 +53,11 @@ presubmits:
             runAsUser: 2000
             allowPrivilegeEscalation: false
           command:
-            - runner.sh
-            - bash
-            - -c
-          args:
+            - make
+            - test
+            - KUBE_RACE=-race
             # TODO: stop depending on custom timeout
-            - 'make test KUBE_RACE=-race KUBE_TIMEOUT=--timeout=240s'
+            - KUBE_TIMEOUT=--timeout=240s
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes/sig-testing/make-test.yaml
```diff
@@ -41,26 +41,23 @@ presubmits:
     cluster: k8s-infra-prow-build
     skip_branches:
     - release-\d+.\d+ # per-release job
-    labels:
-      preset-service-account: "true"
```
BenTheElder, line 45: this should only be necessary for bootstrap jobs
this job is decorated (and the non-experimental) we should be able to drop this from both but starting here

this is relevant here because this + runner.sh cause us to try to activate the credentials / use gcloud, which fails (but we shouldn't need to use gcloud, so just stop using it instead of trying to make that available to non-root)

config/jobs/kubernetes/sig-testing/make-test.yaml
```diff
@@ -41,26 +41,23 @@ presubmits:
...
     optional: true
     always_run: false
     path_alias: k8s.io/kubernetes
     spec:
+      # let's see if we can run unit tests as a non-root user
+      securityContext:
```
BenTheElder, line 49: it appears we can't write to the code mount when we just set the test container, so that's an issue.
setting this on the pod level should default it for all containers.

BenTheElder:
made it to running the code generators 🎉

PR #21248, opened 2021-03-09
Enable dind on pull-kubernetes-e2e-gce-kubetest2

amwat:
/cc @BenTheElder @spiffxp

### DIFF
--- a/config/jobs/kubernetes/sig-cloud-provider/gcp/gcp-gce.yaml
+++ b/config/jobs/kubernetes/sig-cloud-provider/gcp/gcp-gce.yaml
@@ -111,6 +111,7 @@ presubmits:
       timeout: 80m # hard cap, based on original pre-kubetest2 job but moved to a prow-level timeout
     labels:
       preset-service-account: "true"
+      preset-dind-enabled: "true"
     annotations:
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes/sig-cloud-provider/gcp/gcp-gce.yaml
```diff
@@ -124,6 +125,8 @@ presubmits:
           limits:
             cpu: 4
             memory: 14Gi
+        securityContext:
+          privileged: true
         command:
         - wrapper.sh
```
BenTheElder, line 131: please migrate non-kind jobs off of krte at some point, so we can continue to iterate on this image more tightly

BenTheElder (approved):
/lgtm
/approve

config/jobs/kubernetes/sig-cloud-provider/gcp/gcp-gce.yaml
amwat: ACK, currently they are using krte only for dind, go and to ensure we aren't somehow dependent on bootstrap (via kubekins) (atleast to start with).
Eventually, will move to a kubekins(esque) image with kubetest2.

PR #21258, opened 2021-03-09
Turn on json sort check ci job

kkeshavamurthy:
Now that sort files are merged in, turning on the check for all future PRs. 

/assign @codenrhoden

### DIFF
--- a/config/jobs/kubernetes-sigs/image-builder/image-builder-presubmits.yaml
+++ b/config/jobs/kubernetes-sigs/image-builder/image-builder-presubmits.yaml
@@ -48,8 +48,7 @@ presubmits:
       testgrid-tab-name: pr-azure-sigs
   - name: json-sort-check
     decorate: true
-    always_run: false
-    optional: true
+    run_if_changed: 'images/capi/.*\.json$'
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes-sigs/image-builder/image-builder-presubmits.yaml
```diff
@@ -48,8 +48,7 @@ presubmits:
       testgrid-tab-name: pr-azure-sigs
   - name: json-sort-check
     decorate: true
-    always_run: false
-    optional: true
+    run_if_changed: 'images/capi/.*'
```
codenrhoden, line 51: I think this would be better as `images/capi/*\.json$` to only run when a JSON file is changed.

kkeshavamurthy: Yup. Changed.

codenrhoden:
/ok-to-test
/lgtm
/hold

I think we'll want to wait for https://github.com/kubernetes-sigs/image-builder/pull/368 to merge first?

Issue #21126, opened 2021-03-03
Testgrid complains on "podinfo.json" not found and marks conformance tests as failed


dguendisch:
**What happened**:
[Gardener](https://github.com/gardener/gardener) reports conformance tests (and others) since some years into testgrid for various cloudproviders and various k8s versions.
Since yesterday all of our conformance tests are marked as failed because of a test named `Pod` and the error `podinfo.json not found...`
See e.g. https://testgrid.k8s.io/conformance-all#Gardener,%20v1.20%20AWS

What's the issue with the tests and what is this presumably new `Pod` test (not marked as conformance) about?

**What you expected to happen**:
Tests to be marked as succeeded when all conformance tests passed.

cc @schrodit

thejoycekung:
This is also happening on the [Conformance - Openstack](https://testgrid.k8s.io/sig-release-master-informing#Conformance%20-%20OpenStack) job as well. 

/area provider/openstack

thejoycekung:
for visibility: 
/remove-area provider/openstack
/cc @kubernetes/provider-openstack-misc

BenTheElder:
`podinfo.json` is an artifact prow produces about the pod it ran tests in. asking the testgrid owners about this.

BenTheElder:
see: https://github.com/GoogleCloudPlatform/testgrid/issues/393

dguendisch:
I see that GoogleCloudPlatform/testgrid#393 got fixed by adding the option to disable the prow analysis with https://github.com/GoogleCloudPlatform/testgrid/pull/396
Does that mean I could just add `disable_prowjob_analysis: true` to every entry like [this one](https://github.com/kubernetes/test-infra/blob/master/config/testgrids/conformance/conformance-all.yaml#L25)?
In the same line, could I add this config as well [here](https://github.com/kubernetes/test-infra/blob/master/config/testgrids/gardener/config.yaml) or probably on a more global place?

chases2:
If there's a subdirectory of them, it might make sense to add a default.yaml in a subdirectory, but this isn't working right yet. See https://github.com/kubernetes/test-infra/issues/20756

dguendisch:
[...]

PR #21259, opened 2021-03-09
[cpo] added disable_prowjob_analysis

chrigl:
See https://github.com/GoogleCloudPlatform/testgrid/issues/393
and https://github.com/kubernetes/test-infra/issues/21126


dguendisch:
Hi @chrigl, I want to disable prowjob analysis for gardener dashboards (https://github.com/kubernetes/test-infra/pull/21280) and therefore stumbled upon this PR here.
If your intent is as well as my intent to disable the prowjob row, then I think you mixed the actual value. According to https://github.com/GoogleCloudPlatform/testgrid/issues/393#issuecomment-793204421 you would need to add `disable_prowjob_analysis: true` instead of `false`. Sorry if I misunderstood the intent of this PR here :)

chrigl:
@dguendisch 🤣  Thanks so much, I was wondering why it didn't work.

Issue #21260, opened 2021-03-09
Make the sig-windows-gce tests green.


jeremyje:
**What should be cleaned up or changed**:
Test Grid for sig-windows-gce is fairly red. We are unblocked to move forward with the fixes to clean up the redness since the repo lists have stablized.

**Provide any links for context**:
https://testgrid.k8s.io/sig-windows-gce

BenTheElder:
Usually we track test / job health in https://github.com/kubernetes/kubernetes FWIW (the CI signal team in particular files bugs there / keeps tabs on these).

jeremyje:
They are pretty green at this point so I'm closing this issue.

PR #21261, opened 2021-03-09
Use -master repo list for GCE Windows tests.

jeremyje:
This change is first of a set of changes to make the sig-windows-gce tests green.

https://github.com/kubernetes/test-infra/issues/21260


jsturtevant:
@chewong could you take a quick look?

config/jobs/kubernetes/sig-windows/windows-gce.yaml
```diff
@@ -200,7 +202,7 @@ periodics:
     preset-service-account: "true"
     preset-common-gce-windows: "true"
     preset-e2e-gce-windows: "true"
-    preset-windows-repo-list-2004: "true"
+    preset-windows-repo-list-master: "true"
```
chewong, line 205: For 2004 job, I think we want to use image-repo-list-2004 instead because the test image in the default gcAuthenticatedRegistry does not support 2004, so we have to manually point it to e2eprivate dockerhub.

jsturtevant, line 205: I think we need to keep this one at 2004: https://github.com/kubernetes/test-infra/blob/3e9f5cf5f61318e03de3ffc0e16af8926b658b49/config/jobs/kubernetes-sigs/sig-windows/sig-windows-sac.yaml#L120

is that right @chewong ?

jsturtevant: reviewed at the same time ignore this...

jsturtevant: do we have an open PR to get the latest versions into gcAuthenticated?

chewong: @claudiubelu  do you have any update on this? Ideally, we would only want to maintain one image-repo-list for master branch instead of two.

PR #21263, opened 2021-03-09
[sig-node-containerd] Drop perma-fail windows related ci jobs

dims:
- containerd/containerd does not support the shells scripts here
- failures galore as well:
  - https://testgrid.k8s.io/sig-node-containerd#windows-build
  - https://testgrid.k8s.io/sig-node-containerd#cri-validation-windows
 
Also there are other CI jobs maintained by containerd and sig-windows as well. We don't need to do this here


ehashman (commented):
/lgtm

RIP broken tests

PR #21269, opened 2021-03-09
Revert "Revert "Prometheus operator: upgrade from 0.29 to 0.45

chaodaiG:
This reverts commit ae60af3c187f856a8e07e0eab7c2c207f1deb4c5.

Fix:
- Also add RBAC permission for prometheus operator to create CRDs

### DIFF
--- a/config/prow/cluster/monitoring/prometheus_operator_deployment.yaml
+++ b/config/prow/cluster/monitoring/prometheus_operator_deployment.yaml
@@ -18,12 +18,11 @@ spec:
     spec:
       containers:
       - args:
+        - --log-level=all
         - --kubelet-service=kube-system/kubelet
-        - --logtostderr=true
-        - --config-reloader-image=quay.io/coreos/configmap-reload:v0.0.1
-        - --prometheus-config-reloader=quay.io/coreos/prometheus-config-reloader:v0.29.0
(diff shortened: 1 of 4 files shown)


config/prow/cluster/monitoring/prometheus_operator_deployment.yaml
```diff
@@ -18,12 +18,11 @@ spec:
     spec:
       containers:
       - args:
+        - --log-level=all
         - --kubelet-service=kube-system/kubelet
-        - --logtostderr=true
-        - --config-reloader-image=quay.io/coreos/configmap-reload:v0.0.1
```
cjwagner, line 23: This is the only removal that stands out to me, but I figure it is intentional. Please proceed if so.

cjwagner (approved):
/hold

config/prow/cluster/monitoring/prometheus_operator_deployment.yaml
chaodaiG: Yes it is, and it's so obviously noted on the branch
Screen Shot 2021-03-09 at 3 05 13 PM


https://github.com/prometheus-operator/prometheus-operator/blob/release-0.45/cmd/operator/main.go

Issue #21193, opened 2021-03-05
[Kettle] Exclude Job list not excluding results


MushuEE:
**What happened**:
Kettle has a job exclusion list: https://github.com/kubernetes/test-infra/tree/master/kettle#adding-buckets

Jobs in this list should not be added to metrics, however they are: ci-kubernetes-coverage-unit

**What you expected to happen**:
ci-kubernetes-coverage-unit job is not included in metrics values

**How to reproduce it (as minimally and precisely as possible)**:
View BQ table for job results and see the ci-kubernetes-coverage-unit job is being uploaded.

**Please provide links to example occurrences, if any**:
`SELECT * FROM `k8s-gubernator.build.day` where job = "ci-kubernetes-coverage-unit" LIMIT 1000` 

**Anything else we need to know?**:
/assign

BenTheElder:
@MushuEE did you already fix this?

MushuEE:
It has a temporary fix from the Triage side. We were going back and forth on where the exclusion should be: Should kettle pick up everything and Triage filters or vice versa. Will close for now

PR #21271, opened 2021-03-10
[Kettle] Skip stream of jobs that are excluded

MushuEE:
This is not the most pretty but there was no clean way I could think of without changing a whole lot more. I also added tests that give a better picture of how everything is handled

#21193

/area kettle

### DIFF
--- a/kettle/stream.py
+++ b/kettle/stream.py
@@ -27,6 +27,7 @@ import traceback
 import time
 
 import multiprocessing.pool
+import ruamel.yaml as yaml
 
 try:
     from google.api_core import exceptions as api_exceptions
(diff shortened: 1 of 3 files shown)


kettle/stream.py
```diff
@@ -41,23 +42,30 @@
 import make_db
 import make_json
 
+
 MAX_ROW_UPLOAD = 10 # See https://github.com/googleapis/google-cloud-go/issues/2855
 
-def process_changes(results):
+
+def should_exclude(objectID, bucketID, buckets):
+    job_name = objectID.rsplit('/')[-3] # See tests for refernce
```
amwat, line 50: i think a comment of a sample objectID would also go well here [-3] looks like magic :)

MushuEE, line 50:
```suggestion
   # Objects of form a/b/c/<jobname>/<hash>/<objectFile>'
    job_name = objectID.rsplit('/')[-3] # See tests for refernce
```

MushuEE, line 51:
```suggestion
    job_name = objectID.rsplit('/')[-3] # See tests for reference
```

MushuEE: done

PR #21276, opened 2021-03-10
Kubetest Version Addition

### LINKED ISSUE #20886: Make it clear which version of kubetest or kubetest2 is being used
We may have `deprecated` kubetest, but it's still in use.  Whatever we do here should support `kubetest2` or be included in `kubetest2`

https://github.com/kubernetes/kubernetes/pull/96987#issuecomment-780646063 is an example of confusion over the fact that changes to `kubetest` don't take effect until job configs are updated to a new image (e.g. `kubekins`)

We could stand to document how changes flow a better.  But it might help for troubleshooting if we make it clearer what version of kubetest is being used.

Ideas include:
- log the version kubetest was built at
- have kubetest log its version
- store version info as an artifact
- include kubetest's version in `metadata.json` so it can be displayed as a custom testgrid column (maybe related to https://github.com/kubernetes/test-infra/issues/20650)

supriya-premkumar:
Changes:
- Adds version flag for kubetest
- Adds kubetest version in metadata.json
- Fixes part of #20886

### DIFF
--- a/images/kubekins-e2e/cloudbuild.yaml
+++ b/images/kubekins-e2e/cloudbuild.yaml
@@ -15,6 +15,7 @@ steps:
   - name: gcr.io/cloud-builders/docker
     args:
     - build
+    - --ldflags="-X=main.gitTag=$_GIT_TAG"
     - --tag=gcr.io/$PROJECT_ID/kubekins-e2e:$_GIT_TAG-$_CONFIG
     - --build-arg=BAZEL_VERSION_ARG=$_BAZEL_VERSION
     - --build-arg=CFSSL_VERSION=$_CFSSL_VERSION
--- a/kubetest/main.go
+++ b/kubetest/main.go
@@ -48,16 +48,18 @@ const defaultGinkgoParallel = 25
 
 var (
 	artifacts = filepath.Join(os.Getenv("WORKSPACE"), "_artifacts")
-	interrupt = time.NewTimer(time.Duration(0)) // interrupt testing at this time.
-	terminate = time.NewTimer(time.Duration(0)) // terminate testing at this time.
-	verbose   = false
-	timeout   = time.Duration(0)
 	boskos, _ = client.NewClient(os.Getenv("JOB_NAME"), "http://boskos.test-pods.svc.cluster.local.", "", "")
 	control   = process.NewControl(timeout, interrupt, terminate, verbose)
+	gitTag    = ""                              // initializing default zero value. ldflags will populate this during build time.
+	interrupt = time.NewTimer(time.Duration(0)) // interrupt testing at this time.
+	terminate = time.NewTimer(time.Duration(0)) // terminate testing at this time.
+	timeout   = time.Duration(0)
+	verbose   = false
 )
 
 type options struct {
 	build                buildStrategy
+	boskosWaitDuration   time.Duration
 	charts               bool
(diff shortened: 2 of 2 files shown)


supriya-premkumar:
> store version info as an artifact

@spiffxp @dims  I was not able to figure artifact stuff from looking at the code.

Is metadata.json and artifacts different? Appreciate any pointers.

kubetest/main.go
```diff
@@ -274,6 +277,7 @@ func validateFlags(o *options) error {
 
 func main() {
 	log.SetFlags(log.LstdFlags | log.Lshortfile)
+	log.Printf("Running kubetest version: %s", gitTag)
```
amwat, line 280: +1 to printing it always!

Additionally we can just exit after `validateFlags` (line 295) if `version` is true
i.e. when `kubetest --version` is invoked

amwat (commented):
Thank you for adding this!

just a couple of minor comments, but the approach lgtm!

images/kubekins-e2e/cloudbuild.yaml
```diff
@@ -15,6 +15,7 @@ steps:
   - name: gcr.io/cloud-builders/docker
     args:
     - build
+    - --ldflags=-X main.gitTag=$_GIT_TAG
```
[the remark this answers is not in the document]
supriya-premkumar: Ack. That makes sense.

kubetest/main.go
supriya-premkumar: Returning on version makes sense. I think we should do it as a part of `run()`  on line 59 after printing the version

amwat: I think the earlier we terminate, the better.
Usually `--version` would be requested explicitly and wouldn't be coupled with other flags like `--up` `--down` (and/or version gets priority) (similar to `--help`)

so simple invocations like`kubetest --version` can avoid going through `complete -> (acquireKubernetes/validWorkingDirectory) ->run`

supriya-premkumar: Oh. I wanted to keep the handler for version command inside of `run()` for code consistency.
But it makes sense to return early. Will inline this right after `validateFlags`

kubetest/main.go
```diff
@@ -289,6 +293,11 @@ func main() {
 		log.Fatalf("Flags validation failed. err: %v", err)
 	}
 
+	if o.version {
+		log.Printf("kubetest version: %s", gitTag)
```
amwat, line 297: super nit: since this is the expected output of the command this should be written to stdout :)
(and a newline) e.g. for something like`kubetest --version 2>/dev/null`

maybe `fmt.Printf("kubetest version: %s\n", gitTag)` ?

amwat (commented):
Thanks a lot for working on this change!
/approve

just a final comment before lgtm

kubetest/main.go
supriya-premkumar: Ack

amwat: might've missed `log` vs `fmt`

log defaults to stderr, fmt defaults to stdout.

spiffxp (approved):
/approve
/lgtm
Thanks for working on this @supriya-premkumar!

spiffxp:
Edited to drop the fixes part. We'll need more to plumb this through in kettle and testgrid

PR #21280, opened 2021-03-10
Disable prowjob analysis for Gardener testgroups

dguendisch:
Disabling the prowjob analysis for gardener testgroups, see https://github.com/kubernetes/test-infra/issues/21126
I didn't got an answer to https://github.com/kubernetes/test-infra/issues/21126#issuecomment-793730188, so I hope I disabled it in the correct location.

cc @schrodit 
/area testgrid


dguendisch:
cc fyi @chrigl

chases2 (approved):
/ok-to-test

Yeah, adding them to the testgroup like this looks correct! (At least, according to https://github.com/GoogleCloudPlatform/testgrid/issues/393#issuecomment-793204421).

PR #21282, opened 2021-03-10
periodic job to catch unit test flakes

aojea:
We can run the unit tests 10 times in a row to catch flakes, this is impossible to do in a presubmit, but doing it periodically we can correlate presubmit flakes with this job results and identify flakes sooner


dims:
I like the idea, will approve it. will let the others verify if the params are correct/enough ( looks good to me and we can always iterate )

/approve

spiffxp (approved):
/approve
/lgtm

BenTheElder:
thank you!

PR #21291, opened 2021-03-10
Revert "Revert "Prometheus operator: update to 0.45"

chaodaiG:
This reverts commit ddc935d8d75c565a9ff13b2f038215b95fd14727. Corresponding to PR https://github.com/kubernetes/test-infra/pull/21290

Improvements on top of previous attmpt:
- Explicitly define each individual CRDs (The previous attempt used `bundle.yaml` from https://github.com/prometheus-operator/prometheus-operator/blob/release-0.45/bundle.yaml, github UI tricked me by not fully displaying the entire file, and there were many stuff that we don't need in the not displayed part).
- Removed unneeded RBAC since operator doesn't create CRDs any more.

### DIFF
--- a/config/prow/cluster/monitoring/BUILD.bazel
+++ b/config/prow/cluster/monitoring/BUILD.bazel
@@ -27,31 +27,59 @@ release(
 load("@io_bazel_rules_k8s//k8s:object.bzl", "k8s_object")
 
 k8s_object(
-    name = "alertmanager_crd",
+    name = "prometheus_crd_alertmanagerconfigs",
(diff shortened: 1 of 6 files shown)


chaodaiG:
/cc @cjwagner @spiffxp 

/cc @chases2 @mpherman2 
Since oncall

spiffxp (approved):
/approve
/lgtm
/hold
Remove when ready to deploy

chaodaiG:
/unhold
Let's try it the third time

PR #21300, opened 2021-03-11
Prometheus alert: reduce noise by not alerting on expected errors

chaodaiG:
https://github.com/kubernetes/test-infra/pull/21192 doesn't seem to be sufficient on reducing the noise level. There were already 2 alerts showed up in 2 hours after the changes went in effect. Also retrospectively queried previous data and revealed that it'll still be spammy with expected errors. Explicitly stop alerting on these errors in this PR.

### DIFF
--- a/config/prow/cluster/monitoring/mixins/prometheus/ghproxy_alerts.libsonnet
+++ b/config/prow/cluster/monitoring/mixins/prometheus/ghproxy_alerts.libsonnet
@@ -32,8 +32,49 @@
           },
           {
             alert: 'ghproxy-specific-status-code-4xx',
+            // Paths that contains error codes expected by prow(Grabbed from previous prow alerts):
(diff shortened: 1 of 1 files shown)


chaodaiG:
/cc @spiffxp 

/hold
Let's wait a bit longer and see what's going on in https://kubernetes.slack.com/archives/CQP3RC68Y

chaodaiG:
/unhold
The channel is still pretty spammy. @cjwagner @spiffxp , may I get an approval please?

cjwagner (approved):
LGTM, but if this gets any more complicated we should consider using some jsonnet magic to generate these alerts from a list of tuples `<path, code>` that we want to ignore.
It may also be easier to express these exceptions in the form of alertmanager silences. It doesn't seem like great practice to generate alerts that will always be silenced though.
/shrug

chaodaiG:
> LGTM, but if this gets any more complicated we should consider using some jsonnet magic to generate these alerts from a list of tuples `<path, code>` that we want to ignore.
> It may also be easier to express these exceptions in the form of alertmanager silences. It doesn't seem like great practice to generate alerts that will always be silenced though.
> /shrug

Totally agreed. TBH I had also realized that this is not very easily extendable and not scalable, and spent like 10 minutes trying to do the tuple magic with jsonnet but decided might not be this time :)

PR #21301, opened 2021-03-11
add ci-kubernetes-unit

BenTheElder:
xref: https://github.com/kubernetes/enhancements/issues/2420

cc @spiffxp @dims

### DIFF
--- a/config/jobs/kubernetes/sig-testing/make-test.yaml
+++ b/config/jobs/kubernetes/sig-testing/make-test.yaml
@@ -68,6 +68,39 @@ presubmits:
               memory: "36Gi"
 periodics:
   - interval: 1h
+    name: ci-kubernetes-unit
+    annotations:
+      fork-per-release: "true"
+      testgrid-num-failures-to-alert: '3'
+      testgrid-alert-stale-results-hours: '24'
+      testgrid-create-test-group: 'true'
+      testgrid-dashboards: sig-release-master-informing
+      testgrid-alert-email: kubernetes-sig-testing-alerts@googlegroups.com
+    decorate: true
+    cluster: k8s-infra-prow-build
+    extra_refs:
+      - org: kubernetes
+        repo: kubernetes
+        base_ref: master
+        path_alias: k8s.io/kubernetes
+    spec:
+      containers:
+        - image: gcr.io/k8s-testimages/kubekins-e2e:v20210309-0688fde-master
+          command:
+            - make
+            - test
+            - KUBE_RACE=-race
+            # TODO: stop depending on custom timeout
+            - KUBE_TIMEOUT=--timeout=240s
+          # TODO: direct copy from pull-kubernetes-bazel-test, tune these
+          resources:
+            limits:
+              cpu: 4
+              memory: "36Gi"
+            requests:
+              cpu: 4
+              memory: "36Gi"
+  - interval: 1h
     name: ci-kubernetes-generate-make-test-cache
     annotations:
       testgrid-dashboards: sig-testing-canaries


config/jobs/kubernetes/sig-testing/make-test.yaml
```diff
@@ -67,6 +67,38 @@ presubmits:
...
               memory: "36Gi"
 periodics:
+  - interval: 1h
+    name: ci-kubernetes-unit
+    annotations:
+      fork-per-release: "true"
+      testgrid-num-failures-to-alert: '3'
+      testgrid-alert-stale-results-hours: '24'
+      testgrid-create-test-group: 'true'
+      testgrid-dashboards: sig-release-master-informing
```
BenTheElder, line 77: this really ought to be in blocking, but given that the job doesn't exist yet and we had a gap in this versus the bazel based previous job ...

config/jobs/kubernetes/sig-testing/make-test.yaml
```diff
@@ -67,6 +67,38 @@ presubmits:
...
+            # TODO: stop depending on custom timeout
+            - KUBE_TIMEOUT=--timeout=240s
+          # TODO: direct copy from pull-kubernetes-bazel-test, tune these
+          resources:
+            limits:
+              cpu: 4
+              memory: "36Gi"
+            requests:
+              cpu: 4
+              memory: "36Gi"
```
dims, line 101: wowza! 36 :)

config/jobs/kubernetes/sig-testing/make-test.yaml
dims: agree

spiffxp, line 77:
```suggestion
      testgrid-dashboards: sig-release-master-blocking
      testgrid-tab-name: unit-test-master
```

spiffxp: ref: https://github.com/kubernetes/sig-release/blob/master/release-blocking-jobs.md#promoting-or-demoting-jobs

/sig release
/assign @hasheddan @saschagrunert @thejoycekung @palnabarun 
I'd like to defer to SIG Release TLs, v1.21 CI Signal and Lead here ... but here's my pitch to start at blocking.

IMO this is ~equivalent to the `bazel-test` tab that _was_ on release-blocking (eg: https://testgrid.k8s.io/sig-release-1.20-blocking#bazel-test-1.20).  Given that we already have a merge-blocking variant of this job, we're already likely going to see @kubernetes/ci-signal keeping tabs on unit test flakes (community heroics aside)

So while ordinarily I'd advocate for N days/weeks in informing before promoting prior to test freeze, I feel like we don't save much effort by doing so.

spiffxp (approved):
/hold
for comment

config/jobs/kubernetes/sig-testing/make-test.yaml
spiffxp: ... that said a quick trial in informing followed by promote is probably the more reasonable thing I'm asking for vs. just slam into blocking

config/jobs/kubernetes/sig-testing/make-test.yaml
spiffxp: sooo... are these coming from what the bazel jobs used? or have we taken a look at resource consumption and this is still basically where we're at when trying to do as much in parallel as possible?

config/jobs/kubernetes/sig-testing/make-test.yaml
hasheddan: @spiffxp I would be in favor of a short time on `informing` before promotion to `blocking` :+1: Per the [promotion guidelines](https://github.com/kubernetes/sig-release/blob/master/release-blocking-jobs.md#promoting-or-demoting-jobs), we'll need @palnabarun to officially sign off on this.

hasheddan (changes requested):

config/jobs/kubernetes/sig-testing/make-test.yaml
```diff
@@ -67,6 +67,38 @@ presubmits:
               cpu: 4
               memory: "36Gi"
 periodics:
+  - interval: 1h
+    name: ci-kubernetes-unit
+    annotations:
```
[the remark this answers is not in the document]
BenTheElder: the bazel job went to `testgrid-alert-email: kubernetes-sig-testing-alerts@googlegroups.com, k8s-infra-oncall@google.com` the latter no longer makes sense, will add the former.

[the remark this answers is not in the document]
hasheddan: SGTM, thanks!

hasheddan (approved):
/lgtm

config/jobs/kubernetes/sig-testing/make-test.yaml
BenTheElder: I lean towards following the procedure so as not to open a path to exceptions in the future. I think we can afford to add this to release blocking after a short time given we only _just_ entered code freeze and have blocking presubmit coverage (which did go through a promotion path as well, even if theoretically equivilant)

config/jobs/kubernetes/sig-testing/make-test.yaml
BenTheElder: > sooo... are these coming from what the bazel jobs used? or have we taken a look at resource consumption and this is still basically where we're at when trying to do as much in parallel as possible?

These are coming from the bazel jobs (see comment on line 95 above). We run CI on highmem nodes where 52gb : 8 cores is the ratio.

We can likely reduce this, and the timeout, and enable `-race` by default in `make test`, but it's more important to get it stable and blocking.

In 1.21 I will be tuning these, including a push to require unit tests to pass as a non-root user without privileges (in progress but probably landing in 1.21 realistically).

There's a lot wrong with how we've been running unit tests. Starting with lift and shift from bazel 🙃

config/jobs/kubernetes/sig-testing/make-test.yaml
palnabarun: Keeping this in `informing` for a short period of time makes sense. Let's promote this to `blocking` after soaking in for a few days.

BenTheElder:
if someone decides this should move to -blocking immediately, that can easily be done in a follow-up.
it should be fine to merge for now and get running so we have the history to point to when moving it to -blocking otherwise

/hold cancel

config/jobs/kubernetes/sig-testing/make-test.yaml
spiffxp: >it's more important to get it stable and blocking

completely agree, I don't think I phrased well... it was purely curiosity from my end, I didn't know if we'd bothered tuning anything yet (I know there's been a lot of work bringing individual test times down)

BenTheElder: just individual test packages, nothing else has received any tuning yet.
it's very difficult to safely tune resource usage as a prowjob author, we don't point you to anything to observe consumption, biasing on the side of asking for too much is safer.

BenTheElder:
FWIW there is a flake in the second run, but it's not of the test setup AFAICT, just a bad unit test ... https://github.com/kubernetes/kubernetes/issues/100112

PR #21306, opened 2021-03-11
make test timeout is proportional to count

aojea:
We don't want unit test to go beyond a specific threshold, I think that 30 seconds for an unit test is a good and reasonable limit, that is what current parameters does, but it is unfeasible to do it in one shot.
Let's iterate and give now a 1 min per execution margin, once we got there we cut it to the finally 30 seconds per test

### DIFF
--- a/config/jobs/kubernetes/sig-testing/make-test.yaml
+++ b/config/jobs/kubernetes/sig-testing/make-test.yaml
@@ -173,4 +173,4 @@ periodics:
             - bash
             - -c
           args:
-            - 'make test KUBE_RACE=-race KUBE_TIMEOUT=--timeout=300s GOFLAGS=-count=10'
+            - 'make test KUBE_RACE=-race KUBE_TIMEOUT=--timeout=600s GOFLAGS=-count=10'


config/jobs/kubernetes/sig-testing/make-test.yaml
```diff
@@ -173,4 +173,4 @@ periodics:
             - bash
             - -c
           args:
-            - 'make test KUBE_RACE=-race KUBE_TIMEOUT=--timeout=300s GOFLAGS=-count=10'
+            - 'make test KUBE_RACE=-race KUBE_TIMEOUT=--timeout=3000s GOFLAGS=-count=10'
```
BenTheElder, line 176: xref: https://github.com/golang/go/issues/17337

BenTheElder: This is 50 minutes per test though .... I'm not sure this is reasonable.
We should probably not let the tests take so long? 🤔

aojea: let's put 600s, so we can start to cut at the 1 min per execution now, and in the next iteration we limit all tests to 30 seconds max, does it make sense?

BenTheElder:
let's try this anyhow
we can iterate further
FYI @liggitt 
/lgtm
/approve

PR #21307, opened 2021-03-11
releng: Update cross image presubmit to push to the correct staging project

cpanato:
as discussed here: https://github.com/kubernetes/k8s.io/issues/1772 we should try to push the image to the project that is running on.
Later I will check about the `--local` flag

/assign @spiffxp @justaugustus

### DIFF
--- a/config/jobs/kubernetes/release/release-config.yaml
+++ b/config/jobs/kubernetes/release/release-config.yaml
@@ -108,6 +108,8 @@ presubmits:
           env:
             - name: LOG_TO_STDOUT
               value: "y"
+            - name: REGISTRY
+              value: gcr.io/k8s-staging-releng-test
           resources:
             requests:
               cpu: 1000m


justaugustus (changes requested):
@cpanato -- Let's mimic the post-submit here:   https://github.com/kubernetes/test-infra/blob/ff6abef0af0c0644b9085e9fea07ea71322a68ba/config/jobs/image-pushing/releng/k8s-staging-build-image.yaml#L113-L138

Namely: use the `--project` flag to set the project, instead of env vars

justaugustus (approved):
Errr, nevermind. I see it's set further up in the config.

/retitle releng: Update cross image presubmit to push to the correct staging project
/lgtm

hasheddan (approved):
/lgtm

cpanato:
the `--project`  does not set the registry variable

justaugustus:
> the `--project` does not set the registry variable

Yep, see my second review: https://github.com/kubernetes/test-infra/pull/21307#pullrequestreview-609679553

PR #21311, opened 2021-03-11
Check --runtime-config before appending on Node tests runner

knabben:
The flag needs to be backported for older release branches before being forwarded

Fixes https://github.com/kubernetes/kubernetes/issues/100124


aojea:
/lgtm
but I'm not familiar enough with this code so it will be good another pair of eyes

aojea:
/assign @BenTheElder @spiffxp 
someone from sig-node?

dims (approved):
/approve
/lgtm

spiffxp (approved):
/approve 
/lgtm

PR #21315, opened 2021-03-11
Prow autobump job run every 6 hours instead of every hour

chaodaiG:
In preparing for continuous deploying prow.k8s.io. Prow will be updated every 6 hours in the beginning so that it's easier to identify which version to roll back to if something bad happened.

Part of: https://github.com/kubernetes/enhancements/tree/master/keps/sig-testing/2539-continuously-deploy-k8s-prow


spiffxp (approved):
/approve
/lgtm
/hold
Remove when ready to deploy

Issue #21317, opened 2021-03-11
Prow log errors: Failed to create k8s-tpu-operator pipeline client


chaodaiG:
**What happened**:

Seeing repeated errors in k8s prow log:

```
jsonPayload: {
func: "main.main"
msg: "Failed to create k8s-tpu-operator pipeline client"
component: "pipeline"
file: "prow/cmd/pipeline/main.go:160"
level: "fatal"
error: "Get "https://35.227.184.110/apis/tekton.dev/v1alpha1/pipelineruns?limit=1": dial tcp 35.227.184.110:443: i/o timeout"
}
resource: {
type: "k8s_container"
labels: {
pod_name: "prow-pipeline-5bb7586468-cccl5"
namespace_name: "default"
project_id: "k8s-prow"
container_name: "pipeline"
location: "us-central1-f"
cluster_name: "prow"
}
```

chaodaiG:
This is actually causing prow-pipeline pod crashlooping:
```
prow-pipeline-5bb7586468-cccl5             0/1     CrashLoopBackOff   253        22h
```

chaodaiG:
The build cluster named `k8s-tpu-operator` isn't reachable to other prow components as well, very likely the cluster was not reachable from prow any more. Pipeline shouldn't crashloop when this happened, #21318 to make it not panic.

Not sure which team owns `k8s-tpu-operator`.

cjwagner:
@yguo0905 Deleted the cluster. I don't see any job config for that cluster so we should probably remove the kubeconfig entry now.

chaodaiG:
> @yguo0905 deleted the cluster. I don't see any job config for that cluster so we should probably remove the kubeconfig entry now.

Done, deleted it from kubeconfig, @yguo0905 please let us know if this is still needed

chaodaiG:
Pipeline stopped crashlooping after this, mark it fixed for now
/close

PR #21318, opened 2021-03-11
prow/pipeline: don't panic when a build cluster is not reachable

chaodaiG:
Part of: https://github.com/kubernetes/test-infra/issues/21317

The build cluster named k8s-tpu-operator had been unreachable in the past day, this shouldn't make pipeline crashlooping

### DIFF
--- a/prow/cmd/pipeline/main.go
+++ b/prow/cmd/pipeline/main.go
@@ -156,8 +156,10 @@ func main() {
 			logrus.WithError(err).Infof("Ignoring cluster context %s: tekton pipeline CRD not deployed", context)
 			continue
 		}
+		// Don't panic when a build cluster cannot be reached
 		if err != nil {
-			logrus.WithError(err).Fatalf("Failed to create %s pipeline client", context)
+			logrus.WithError(err).Warningf("Failed to create %s pipeline client", context)
+			continue
 		}
 		pipelineConfigs[context] = *bc
 	}


prow/cmd/pipeline/main.go
```diff
@@ -150,6 +152,11 @@ func main() {
 
 	pipelineConfigs := map[string]pipelineConfig{}
 	for context, cfg := range configs {
+		// Don't panic when a build cluster cannot be reached
```
alvaroaleman, line 155: The client creation runs a full api discovery which is an expensive way to check if a cluster is reachable. Wouldn't it be simpler to instead replace the `if err != nil { Fatalf()}` statement below with a Warning and a `continue`?

chaodaiG: Agreed and I like your approach better. Would like to hear what's @fejta think about this approach

fejta: sgtm

Though in an ideal world a transient error when starting the pipeline controller would not cause this context to be permanently unavailable. AKA we'd retry every minute or 10 or whatever and add it to the map once available -- but that'll be a more involved change.

chaodaiG: Done

fejta (approved):
/hold

prow/cmd/pipeline/main.go
```diff
@@ -156,8 +156,10 @@ func main() {
 			logrus.WithError(err).Infof("Ignoring cluster context %s: tekton pipeline CRD not deployed", context)
 			continue
 		}
+		// Don't panic when a build cluster cannot be reached
 		if err != nil {
-			logrus.WithError(err).Fatalf("Failed to create %s pipeline client", context)
+			logrus.WithError(err).Warningf("Failed to create %s pipeline client", context)
```
cjwagner, line 161: nit: I think this should be error level rather than warning.

cjwagner (approved):
/hold

prow/cmd/pipeline/main.go
chaodaiG: Other major prow components already reported this as error, I think it's ok to make it a warning since we'll always get it from other prow components already

PR #21325, opened 2021-03-11
Add 'validate-all' test to k-sigs/image-builder

codenrhoden:
This patch adds a new pre-submit for `kubernetes-sigs/image-builder` that runs `make validate-all` upon relevant file changes.

Some env vars needs to be set to fake values for the tests to pass.

/assign @CecileRobertMichon 
/cc @perithompson @kkeshavamurthy

### DIFF
--- a/config/jobs/kubernetes-sigs/image-builder/image-builder-presubmits.yaml
+++ b/config/jobs/kubernetes-sigs/image-builder/image-builder-presubmits.yaml
@@ -65,3 +65,32 @@ presubmits:
     annotations:
       testgrid-dashboards: sig-cluster-lifecycle-image-builder
       testgrid-tab-name: pr-json-sort-check
+  - name: packer-validate
+    decorate: true
+    always_run: false
+    optional: true
+    decoration_config:
+      timeout: 20m
+    max_concurrency: 5
+    path_alias: sigs.k8s.io/image-builder
+    spec:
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes-sigs/image-builder/image-builder-presubmits.yaml
```diff
@@ -65,3 +65,31 @@ presubmits:
     annotations:
       testgrid-dashboards: sig-cluster-lifecycle-image-builder
       testgrid-tab-name: pr-json-sort-check
+  - name: packer-validate
+    decorate: true
+    run_if_changed: 'images/capi/Makefile|images/capi/packer/.*'
+    decoration_config:
```
CecileRobertMichon, line 71: we should set this to 

```
always_run: false
optional: true
```

so it can be merged and used to validate https://github.com/kubernetes-sigs/image-builder/pull/546, and then once https://github.com/kubernetes-sigs/image-builder/pull/546 merges we can change it to required and run_if_changed

codenrhoden: @CecileRobertMichon Thanks for keeping me honest. PTAL?

CecileRobertMichon (commented):
/lgtm
/approve

PR #21328, opened 2021-03-11
Prow prometheus monitors kube-state-metrics

chaodaiG:
This was from https://github.com/openshift/release/blob/ac1b4f17255011592a2fb104d121668fd6b85ef5/clusters/app.ci/prow-monitoring/kube-state-metrics.yaml and https://github.com/prometheus-operator/kube-prometheus/blob/release-0.7/manifests/kube-state-metrics-serviceMonitor.yaml, not sure whether this will work or not. But shouldn't break anything, I think


alvaroaleman (approved):
/hold

Cancel at your convenience

PR #21329, opened 2021-03-12
Prometheus alert: reduce global ghproxy alert noise, add missing doc

chaodaiG:
One line of doc was missing from https://github.com/kubernetes/test-infra/pull/21300, add here. Also adjust global alert threshold for known error types to make it less noisy


config/prow/cluster/monitoring/mixins/prometheus/ghproxy_alerts.libsonnet
```diff
@@ -86,7 +87,19 @@
...
+              severity: 'warning',
+            },
+            annotations: {
+              message: '{{ $value | humanize }}%% of all API requests through the GitHub proxy are errorring with code {{ $labels.status }}. Check %s.' % [monitoringLink('/d/%s/github-cache?orgId=1&refresh=1m&fullscreen&panelId=8' % [dashboardID], 'the ghproxy dashboard')],
+            },
+          },
+          {
+            alert: 'ghproxy-global-status-code-403-405-422',
+            expr: |||
+              sum(rate(github_request_duration_count{status=~"403|405|422"}[30m])) by (status) / ignoring(status) group_left sum(rate(github_request_duration_count[30m])) * 100 > 10
```
cjwagner, line 102: Hopefully we can tighten the time window for the global alerts more than this now that we are excluding the noisy path+status combos. 30mins is a very long delay for an alert about this kind of problem.

cjwagner (approved):

PR #21349, opened 2021-03-12
Migrate SIG Scalability k/k presubmit jobs to pod-utils

tosi3k:


tosi3k:
Let's hold this until Monday.

/hold

mborsz:
I think it looks ok, but I would prefer to migrate perf-tests presubmits first to double check that everything works as expected. In the next wave we can migrate k/k presubmits

tosi3k:
> I think it looks ok, but I would prefer to migrate perf-tests presubmits first to double check that everything works as expected. In the next wave we can migrate k/k presubmits

k8s.io/perf-tests presubmits were migrated in https://github.com/kubernetes/test-infra/pull/21372. I changed this PR to change just the k/k presubmits in the meantime.

Let me invoke the presubmits to perf-tests to see whether we succeeded and revisit this PR afterwards.

tosi3k:
> Let me invoke the presubmits to perf-tests to see whether we succeeded and revisit this PR afterwards.

Working as intended: https://github.com/kubernetes/perf-tests/pull/1767.

mborsz:
Thanks!

PR #21351, opened 2021-03-12
update dual-stack regex to remove alpha

aramase:

- Update `gingko.focus` regex to match on `IPv6DualStack` as the alpha tag is going be removed.

Reference PR: https://github.com/kubernetes/kubernetes/pull/100189

This change should continue to work with current master as the regex will match with `IPv6DualStackAlphaFeature` as well. 

/cc @aojea


aojea:
/lgtm
/approve
Thanks

By the way , we should work on having the dual-stack jobs before the release https://testgrid.k8s.io/sig-network-dualstack-azure-e2e , I can help if needed, but pretty busy these days

aramase:
Test failed because of
```
ERROR: no such package '@io_k8s_repo_infra//': java.io.IOException: Error downloading [https://github.com/kubernetes/repo-infra/archive/v0.2.0.tar.gz] to /bazel-scratch/.cache/bazel/_bazel_root/05618a594cb3499a4b912817865867cd/external/io_k8s_repo_infra/v0.2.0.tar.gz: GET returned 500 Internal Server Error
```

Looks like there is an ongoing github incident

PR #21369, opened 2021-03-14
Ignoring KubeProxyDaemonSetMigration tests on netpol jobs

knabben:
Flake tests on due to `kube-proxy` folder migration: https://github.com/kubernetes/kubernetes/pull/100214

https://testgrid.k8s.io/sig-network-gce#network-policies,%20google-gce

### DIFF
--- a/config/jobs/kubernetes/sig-network/sig-network-misc.yaml
+++ b/config/jobs/kubernetes/sig-network/sig-network-misc.yaml
@@ -51,7 +51,7 @@ presubmits:
         - --provider=gce
         # Skipping snat tests probably GCE related? https://github.com/kubernetes/test-infra/issues/20321
         # Skipping Cloud Provider specific tests: LoadBalancer, ESIPP (Source IP Preservation)
-        - --test_args=--ginkgo.focus=\[sig-network\]|\[Conformance\]|\[Feature:NetworkPolicy\] --ginkgo.skip=\[Feature:(Networking-IPv6|Example|Federation|PerformanceDNS)\]|DualStack|GCE|SCTP|Disruptive|Serial|SNAT|LB.health.check|LoadBalancer|load.balancer|ESIPP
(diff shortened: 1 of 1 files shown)


knabben:
https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/ci-kubernetes-e2e-ubuntu-gce-network-policies/1371101056292884480

```
Fetching the previously installed CoreDNS version
STORAGE_MEDIA_TYPE must be specified when run non-interactively.
The default etcd storage media type in 1.6 has changed from application/json to application/vnd.kubernetes.protobuf.
Documentation about the change can be found at https://kubernetes.io/docs/admin/etcd_upgrade.

ETCD2 DOES NOT SUPPORT PROTOBUF: If you wish to have to ability to downgrade to etcd2 later application/json must be used.

It's HIGHLY recommended that etcd be backed up before this step!!

To enable using json, before running this script set:
export STORAGE_MEDIA_TYPE=application/json

To enable using protobuf, before running this script set:
export STORAGE_MEDIA_TYPE=application/vnd.kubernetes.protobuf

Mar 14 14:19:35.975: FAIL: Unexpected error:
```

aojea:
We shouldn't be running this test here, on mobile now, but maybe the test should be tagged as disruptive or serial?

aojea:
/hold
better ask sig-testing folks , the test also seems to panic, it belongs to upgrade as @knabben pointed out it comes from https://github.com/kubernetes/kubernetes/pull/100214

@spiffxp @BenTheElder @wojtek-t what is the best way to proceed here?

- we skip the test in this job
- we tag the test as disruptive/serial/...?
- ..

liggitt:
I'm not familiar with the disruptive or panic aspects mentioned here, but the etcd2 / proto storage warning is woefully outdated, that hasn't been relevant since ~1.12. That stanza of the upgrade script can probably be deleted.

wojtek-t:
Yes - the media-type is unrelated.

The upgrade tests shouldn't be running as part of this suite.
The test-args should be modified to exclude the "Feature:KubeProxyDaemonSetMigration" instead.

aojea:
> The test-args should be modified to exclude the "Feature:KubeProxyDaemonSetMigration" instead.

@knabben do you mind changing the PR to filter by this tag instead?

knabben:
/retitle Ignoring KubeProxyDaemonSetMigration tests on netpol jobs

knabben:
@aojea sure, filtering the tags instead.

PR #21373, opened 2021-03-15
Propagate config.Spyglass to lens

mborsz:
As requested in https://github.com/kubernetes/test-infra/pull/20912, I'm propagating config.Spyglass to lens.

/assign @alvaroaleman

### DIFF
--- a/prow/spyglass/api/spyglass.go
+++ b/prow/spyglass/api/spyglass.go
@@ -18,6 +18,8 @@ package api
 
 import (
 	"encoding/json"
+
+	"k8s.io/test-infra/prow/config"
 )
 
 // Key types specify the way Spyglass will fetch artifact handles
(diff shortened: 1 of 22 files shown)


alvaroaleman (approved):

prow/spyglass/api/spyglass.go
```diff
@@ -29,13 +31,13 @@ const (
 // Lens defines the interface that lenses are required to implement in order to be used by Spyglass.
 type Lens interface {
 	// Header returns a a string that is injected into the rendered lens's <head>
-	Header(artifacts []Artifact, resourceRoot string, config json.RawMessage) string
+	Header(artifacts []Artifact, resourceRoot string, config json.RawMessage, spyglassConfig config.Spyglass) string
```
fejta, line 34: What is in `config json.RawMessage`? Why do remote lenses need to receive prow's entire internal spyglass config datastructure?

mborsz: config.RawMessage is a config for a given lens. In https://github.com/kubernetes/test-infra/pull/20912 I wanted to add a per-lens gcs bucket prefix to a lens config and I received a feedback to reuse the one that is already in spyglass config, which is reasonable to me.

Instead of propagating only gcs bucket prefix field from the spyglass config (which would be very specific and unlikely to be useful in other lenses) I decided to pass entire config.Spyglass as other lenses may want to use some other field.

@fejta, Do you want to pass only GCS bucket prefix or you disagree with this approach?

@alvaroaleman

mborsz: config json.RawMessage is a config for a given lens. In https://github.com/kubernetes/test-infra/pull/20912 I wanted to add a per-lens gcs bucket prefix to a lens config and I received a feedback to reuse the one that is already in spyglass config, which is reasonable to me.

Instead of propagating only gcs bucket prefix field from the spyglass config (which would be very specific and unlikely to be useful in other lenses) I decided to pass entire config.Spyglass as other lenses may want to use some other field.

@fejta, Do you want to pass only GCS bucket prefix or you disagree with this approach?

@alvaroaleman

alvaroaleman: So generally, I do not have a super strong opinion here one way or another. Conceptionally I think that the `config.Spyglass` contains both global settings which might or might not be useful for multiple lenses like the prefix and lens-specific settings that other lenses should not be using (if they do, that is a sign that it shouldn't be a lens-specific setting).

But I also see that its more work to add a new config structure that contains only the global settings and that passing only the one field the one lens happens to need to all lenses seems weird.

fejta: Fair enough, and it looks like this is only relevant for in-binary lenses (not remote lenses), right? If so SGTM

fejta (approved):
/hold

prow/spyglass/api/spyglass.go
mborsz: Yes, I understand this code path is used only for in-binary lenses.

mborsz:
/hold cancel

Thanks for review!

PR #21375, opened 2021-03-15
Prometheus: service monitor also monitors label app.kubernetes.io/name

chaodaiG:
This is the default label of kube state metrics. Prefer to make prometheus monitor this label instead of renaming this label to avoid complication on the kube state metrics service itself

### DIFF
--- a/config/prow/cluster/monitoring/prow_prometheus.yaml
+++ b/config/prow/cluster/monitoring/prow_prometheus.yaml
@@ -32,6 +32,8 @@ spec:
     matchExpressions:
     - key: app
       operator: Exists
+    - key: app.kubernetes.io/name
+      operator: Exists
   version: v2.7.1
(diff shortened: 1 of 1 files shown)


cjwagner (approved):
> Prefer to make prometheus monitor this label instead of renaming this label to avoid complication on the kube state metrics service itself

FWIW using the `app` label like we do for the rest of the components wouldn't require any changes to the kube state metrics service. Only the label on the service monitor itself (not it's matcher) would need to be updated. https://github.com/kubernetes/test-infra/blob/198593ae42013ee1a0b04e756b8f9fea33b6e649/config/prow/cluster/monitoring/kube-state-metrics_servicemonitors.yaml#L5

I think this is ok too though.
/hold

chaodaiG:
/unhold
Let's give it a shot

PR #21378, opened 2021-03-15
Revert "Prometheus: service monitor also monitors label app.kubernete…

chaodaiG:
…s.io/name"

This reverts commit 4667228810f0fdb70864e121775e651a36fc0978.

Saw multiple alerts `grafanaDown`, `prometheusDown` etc. right after the PR #21375 , reverting it for now


chaodaiG:
Clarify: they are not really down, the alerts were fired because prometheus didn’t get metrics from these components

alvaroaleman (approved):

PR #21382, opened 2021-03-15
kube-state-metrics: Add 'app:' label so that prometheus can access it…

chaodaiG:
…s metrics

### DIFF
--- a/config/prow/cluster/monitoring/kube-state-metrics_servicemonitors.yaml
+++ b/config/prow/cluster/monitoring/kube-state-metrics_servicemonitors.yaml
@@ -4,6 +4,7 @@ metadata:
   labels:
     app.kubernetes.io/name: kube-state-metrics
     app.kubernetes.io/version: 1.9.7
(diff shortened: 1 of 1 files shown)


config/prow/cluster/monitoring/kube-state-metrics_servicemonitors.yaml
```diff
@@ -4,6 +4,7 @@ metadata:
   labels:
     app.kubernetes.io/name: kube-state-metrics
     app.kubernetes.io/version: 1.9.7
+    app: kube-state-metrics
```
cjwagner, line 7: Why are both this and https://github.com/kubernetes/test-infra/pull/21375 needed?

cjwagner: Oh nevermind, I see the revert now: https://github.com/kubernetes/test-infra/pull/21378

cjwagner (approved):

PR #21383, opened 2021-03-15
Prometheus servcie monitor: use consistent ports names as kube-state-metrics service

chaodaiG:
Prometheus still not get metrics from kube-state-metrcis after #21382 , this might be the reason

### DIFF
--- a/config/prow/cluster/monitoring/kube-state-metrics_servicemonitors.yaml
+++ b/config/prow/cluster/monitoring/kube-state-metrics_servicemonitors.yaml
@@ -14,14 +14,14 @@ spec:
       key: ""
     honorLabels: true
     interval: 2m
-    port: https-main
+    port: http-metrics
     scheme: https
     scrapeTimeout: 2m
   - bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
     bearerTokenSecret:
       key: ""
     interval: 2m
-    port: https-self
+    port: telemetry
     scheme: https
     scrapeTimeout: 2m
   jobLabel: app.kubernetes.io/name
--- a/config/prow/cluster/monitoring/kube-state-metrics_servicemonitors.yaml
+++ b/config/prow/cluster/monitoring/kube-state-metrics_servicemonitors.yaml
@@ -15,14 +15,14 @@ spec:
     honorLabels: true
     interval: 2m
     port: http-metrics
-    scheme: https
+    scheme: http
     scrapeTimeout: 2m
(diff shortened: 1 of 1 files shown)


chaodaiG:
See https://github.com/kubernetes/test-infra/blob/786971a8d8db9b677f8dce9d56bd726b688bc774/config/prow/cluster/kube-state-metrics_service.yaml#L12 for ports names

chaodaiG:
/hold
I think it still need tls setting, will work on it

chaodaiG:
Added tls config. Based on what I read, the cert used by servicemonitor could be any tls cert. @alvaroaleman, did I get it correct?

alvaroaleman:
TBH I am not familiar with tls for k-s-m. In Openshift, this gets set up automatically and the corresponding controller probably can mint CAs trusted by the cluster CA.

In my private kube cluster that probably more closely resembles this, I just make prometheus scrape port 8080 http without tls, as k-s-m doesn't have a serving cert. If you really need that, you would probably need something like cert-manager to mint it for you.

chaodaiG:
> TBH I am not familiar with tls for k-s-m. In Openshift, this gets set up automatically and the corresponding controller probably can mint CAs trusted by the cluster CA.
> 
> In my private kube cluster that probably more closely resembles this, I just make prometheus scrape port 8080 http without tls, as k-s-m doesn't have a serving cert. If you really need that, you would probably need something like cert-manager to mint it for you.

I see, thank you! Prometheus is getting `Get "https://10.108.19.102:8080/metrics": http: server gave HTTP response to HTTPS client`, my initial suspicion was tls, this probably is not the case if your private cluster could work without tls. I'll take more look

alvaroaleman:
You probably want to change this to http: https://github.com/kubernetes/test-infra/blob/fe4f16f621305342bfea37f9f6bcce84d9a523e1/config/prow/cluster/monitoring/kube-state-metrics_servicemonitors.yaml#L24

alvaroaleman:
and here: https://github.com/kubernetes/test-infra/blob/fe4f16f621305342bfea37f9f6bcce84d9a523e1/config/prow/cluster/monitoring/kube-state-metrics_servicemonitors.yaml#L17

The tls might be a valid expectation in Openshift but not here

chaodaiG:
> and here:
> 
> https://github.com/kubernetes/test-infra/blob/fe4f16f621305342bfea37f9f6bcce84d9a523e1/config/prow/cluster/monitoring/kube-state-metrics_servicemonitors.yaml#L17
> 
> The tls might be a valid expectation in Openshift but not here

omg! That definitely should be it

alvaroaleman (approved):

PR #21386, opened 2021-03-16
Make gerrit adapter's job triggered comments link to results on deck.

cjwagner:
Currently we only link to job results when reporting job completion and if users want to inspect job results while a job is running they must find the jobs by filtering on the Prow front end manually. This change updates the "Triggered %d jobs" comment to also link to the Prow front end with a filter for all jobs run on the PR. 
Linking to every job's spyglass page individually was considered, but without proper link formatting it was considered too verbose/difficult to read. Additionally, at trigger time we don't have a build ID so we cannot link to individual job runs in spyglass at that time, we'd need to instead have the gerrit reporter provide that information when the job transitions to the pending state.

/assign @fejta @chaodaiG

### DIFF
--- a/prow/gerrit/adapter/adapter.go
+++ b/prow/gerrit/adapter/adapter.go
@@ -328,20 +328,49 @@ func (c *Controller) processChange(logger logrus.FieldLogger, instance string, c
 
 	// comment back to gerrit if Report is set for any of the jobs
 	var reportingJobs int
-	var message string
+	var jobList string
 	for _, job := range triggeredJobs {
 		if job.report {
-			message += fmt.Sprintf("\n  * Name: %s", job.name)
+			jobList += fmt.Sprintf("\n  * Name: %s", job.name)
 			reportingJobs++
 		}
 	}
 
 	if reportingJobs > 0 {
-		message = fmt.Sprintf("Triggered %d prow jobs (%d suppressed reporting):", len(triggeredJobs), len(triggeredJobs)-reportingJobs) + message
+		message := fmt.Sprintf("Triggered %d prow jobs (%d suppressed reporting): ", len(triggeredJobs), len(triggeredJobs)-reportingJobs)
+		// If we have a Deck URL, link to all results for the CL, otherwise list the triggered jobs.
+		link, err := deckLinkForPR(c.config().Gerrit.DeckURL, refs, change.Status)
+		if err != nil {
+			logger.WithError(err).Error("Failed to generate link to job results on Deck.")
+		}
+		if link != "" && err == nil {
+			message = message + link
+		} else {
+			message = message + jobList
+		}
 		if err := c.gc.SetReview(instance, change.ID, change.CurrentRevision, message, nil); err != nil {
 			return err
 		}
 	}
 
 	return nil
+}
+
+func deckLinkForPR(deckURL string, refs prowapi.Refs, changeStatus string) (string, error) {
+	if deckURL == "" || changeStatus == client.Merged {
+		return "", nil
+	}
+
+	parsed, err := url.Parse(deckURL)
+	if err != nil {
+		return "", fmt.Errorf("failed to parse gerrit.deck_url (impossible: this should have been caught at load time): %v", err)
+	}
+	query := parsed.Query()
+	query.Set("repo", fmt.Sprintf("%s/%s", refs.Org, refs.Repo))
+	if len(refs.Pulls) != 1 {
+		return "", fmt.Errorf("impossible: triggered jobs for a Gerrit change, but refs.pulls was empty")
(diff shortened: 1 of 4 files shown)


prow/gerrit/adapter/adapter.go
```diff
@@ -328,20 +328,49 @@ func (c *Controller) processChange(logger logrus.FieldLogger, instance string, c
...
 
 	if reportingJobs > 0 {
-		message = fmt.Sprintf("Triggered %d prow jobs (%d suppressed reporting):", len(triggeredJobs), len(triggeredJobs)-reportingJobs) + message
+		message := fmt.Sprintf("Triggered %d prow jobs (%d suppressed reporting): ", len(triggeredJobs), len(triggeredJobs)-reportingJobs)
+		// If we have a Deck URL, link to all results for the CL, otherwise list the triggered jobs.
+		link, err := deckLink(c.config().Gerrit.DeckURL, refs)
+		if err != nil {
+			logger.WithError(err).Error("Failed to generate link to job results on Deck.")
+		}
+		if link != "" {
```
chaodaiG, line 346: Should we rely on error instead of link? My feeling is it probably doesn't make too much sense to display link if `deckLink` returns error, since it could either be a wrong or partial link

prow/gerrit/adapter/adapter.go
```diff
@@ -328,20 +328,49 @@ func (c *Controller) processChange(logger logrus.FieldLogger, instance string, c
...
+		}
 		if err := c.gc.SetReview(instance, change.ID, change.CurrentRevision, message, nil); err != nil {
 			return err
 		}
 	}
 
 	return nil
 }
+
+func deckLink(deckURL string, refs prowapi.Refs) (string, error) {
```
chaodaiG, line 359: nit: deckLinkForPR

prow/gerrit/adapter/adapter.go
cjwagner: If `deckLink` returns an error the link will be empty, I think that's a pretty common contract with go functions that return a value and an error, but I can make this more explicit.
We can't just rely on the error because the link can be empty without an error, but I can gate this on both the error and the link.

prow/gerrit/adapter/adapter.go
```diff
@@ -328,20 +328,49 @@ func (c *Controller) processChange(logger logrus.FieldLogger, instance string, c
...
 	if reportingJobs > 0 {
-		message = fmt.Sprintf("Triggered %d prow jobs (%d suppressed reporting):", len(triggeredJobs), len(triggeredJobs)-reportingJobs) + message
+		message := fmt.Sprintf("Triggered %d prow jobs (%d suppressed reporting): ", len(triggeredJobs), len(triggeredJobs)-reportingJobs)
+		// If we have a Deck URL, link to all results for the CL, otherwise list the triggered jobs.
+		link, err := deckLinkForPR(c.config().Gerrit.DeckURL, refs)
+		if err != nil {
+			logger.WithError(err).Error("Failed to generate link to job results on Deck.")
+		}
+		if link != "" && err == nil {
+			message = message + link
```
chaodaiG, line 347: Sorry missed this yesterday: should message include jobList also in this case?

cjwagner: I intentionally omitted the list of job names in favor of the link. We print the count of triggered jobs in either case, but I figured if we have the link we can reduce the size of the bot comments to make things more readable. I don't think users typically care which presubmits are triggered, just that they are triggered. Following the link will lead to specifics if they are needed.
WDYT?

cjwagner:
/hold
I realized these links won't work for postsubmit jobs because postsubmits unfortunately are not associated with the PR. I'll see if there is a way to generate a link to filter the appropriate postsubmits, otherwise I'll keep the old behavior for postsubmits.

prow/gerrit/adapter/adapter.go
chaodaiG: Makes sense. It probably doesn't make that much sense printing out each job if they are not clickable, so lgtm

chaodaiG (approved):

cjwagner:
We don't have a way to filter by SHA or baseref which is what we'd need to link to the appropriate postsubmits. I think the best we could do with the existing filter fields is link to all postsubmits for the repo which seems too broad so I kept the existing behavior of listing the job names without links for postsubmits.

prow/gerrit/adapter/adapter.go
fejta: This also initially seemed odd to me but I think that's just my change aversion bias. Making these messages shorter SGTM.

fejta (approved):
/hold

PR #21388, opened 2021-03-16
Drop specific zone flag for arm builds

olemarkus:
/cc @hakman


hakman:
Please leave the zo.ne and set it to eu-central-1

olemarkus:
done

hakman (approved):

PR #21394, opened 2021-03-16
Prometheus alert for pod crashlooping

chaodaiG:
Add alert catching pod crashlooping

### DIFF
--- a/config/prow/cluster/monitoring/mixins/prometheus/prow_alerts.libsonnet
+++ b/config/prow/cluster/monitoring/mixins/prometheus/prow_alerts.libsonnet
@@ -1,0 +1,23 @@
+{
+  prometheusAlerts+:: {
+    local componentName = $._config.components.monitoring,
+    groups+: [
+      {
+        name: 'prow',
+        rules: [
+          {
+            alert: 'prow-pod-crashlooping',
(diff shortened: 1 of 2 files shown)


chaodaiG:
The alert is currently set to be pretty sensitive, any pod restart would trigger it. Will adjust later if we feel it's too loud

alvaroaleman (approved):

config/prow/cluster/monitoring/mixins/prometheus/prow_alerts.libsonnet
```diff
@@ -0,0 +1,23 @@
+{
+  prometheusAlerts+:: {
+    local componentName = $._config.components.monitoring,
+    groups+: [
+      {
+        name: 'prow',
+        rules: [
+          {
+            alert: 'prow-pod-crashlooping',
+            expr: 'rate(kube_pod_container_status_restarts_total{namespace=~"default|prow-monitoring",job="kube-state-metrics"}[5m]) * 60 * 5 > 0',
```
cjwagner, line 10: FYI for the future: Prometheus has a convenient `increase(<metric>[<duration>])` function that can be used instead of `rate(<metric>[<duration>]) * <duration in seconds>`. It handles breaks in monotonicity just like `rate()`.

Issue #20756, opened 2021-02-05
Configurator only respects the root default.yaml, not local default.yaml


chases2:
**What happened**:
Configurator, currently being used to manage the TestGrid configuration at testgrid.k8s.io via [this script](https://github.com/kubernetes/test-infra/blob/master/testgrid/config-upload.sh), is uploading configurations with defaults populated by the root default file only.

**What you expected to happen**:
Configurator should upload a TestGrid configuration with default settings based on local default.yaml files.

**How to reproduce it (as minimally and precisely as possible)**:
Run `./testgrid/config_upload.sh` with the current checked-in configuration

**Please provide links to example occurrences, if any**:
Kubevirt default settings: https://github.com/kubernetes/test-infra/blob/master/config/testgrids/kubevirt/default.yaml
Kubevirt loaded configuration: https://testgrid.k8s.io/q/config/kubevirt-periodics/periodic-kubevirt-e2e-k8s-prev

**Anything else we need to know?**:
Necessary improvements to yamlcfg libraries in TestGrid:
https://github.com/GoogleCloudPlatform/testgrid/issues/7

/cc @michelle192837 @mpherman2 @fgimenez

chases2:
The flow here is convoluted, since Configurator is being used both to move settings from Kubevirt's Prow instance to the k8s/test-infra repository (via Transfigure) and to upload from that repository to cloud.

I see two ways forward:
1) Change Configurator to have a "seek default" flag instead of a specific default, and ensure that Transfigure does not apply defaults while ./config-upload.sh does.
2) Replace our use of Transfigure with a [config-merger](https://github.com/GoogleCloudPlatform/testgrid/pull/349), leaving Kubevirt to run Configurator on their own Prow instance with their own default settings.

spiffxp:
>Change Configurator to have an "apply default" flag instead of a specific default, and ensure that Transfigure does not apply defaults while ./config-upload.sh does.

Instead of a new flag, if we dropped the assumption in this if, not specifying `--default path/to/default.yaml` would mean "don't apply defaults"
https://github.com/kubernetes/test-infra/blob/9188da5e58d4d947b23cef3b863e340092e24d5c/testgrid/cmd/configurator/main.go#L103-L106

fgimenez:
/assign 
/remove-help

I'd like to give it a try, will follow @spiffxp suggestion and ping you when ready :)

spiffxp:
[...]

PR #21400, opened 2021-03-16
Add config test to mergelists

chases2:
Updates TestGrid version as well.
The version update caused unit testing to enforce requirements on local default files. This may fix https://github.com/kubernetes/test-infra/issues/20756

/assign @fgimenez 
I'm looking for feedback on the default file. Now that https://github.com/kubernetes/test-infra/pull/20902 is in, should the default file be updated (as shown) or removed entirely?

/assign @mpherman2 
The strict default config option was set to "false" for now.

Follow-up to https://github.com/kubernetes/test-infra/pull/20986

/lint

### DIFF
--- a/prow/spyglass/lenses/junit/lens.go
+++ b/prow/spyglass/lenses/junit/lens.go
@@ -158,7 +158,7 @@ func (lens Lens) getJvd(artifacts []api.Artifact) JVD {
 				resultChan <- result
 				return
 			}
-			var suites junit.Suites
+			var suites *junit.Suites
 			suites, result.err = junit.Parse(contents)
 			if result.err != nil {
 				logrus.WithError(result.err).WithField("artifact", artifact.CanonicalLink()).Info("Error parsing junit file.")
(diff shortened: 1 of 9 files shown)


fgimenez:
> I'm looking for feedback on the default file. Now that #20902 is in, should the default file be updated (as shown) or removed entirely?

@chases2 thanks for the heads up! I think it is ok to completely remove `config/testgrids/kubevirt/default.yaml`, we only needed the `open_test_template.url` customization and is now working fine with #20902

chases2:
Deleted the unused default.yaml

Update-deps updates TestGrid and changes the go.mod and go.sum files, but it does so in a way that verify-deps won't accept as accurate.

chases2:
Found the root issue: https://github.com/kubernetes/test-infra/issues/21463

This PR should be good to go now

prow/spyglass/lenses/junit/lens.go
```diff
@@ -158,7 +158,7 @@ func (lens Lens) getJvd(artifacts []api.Artifact) JVD {
 				resultChan <- result
 				return
 			}
-			var suites junit.Suites
+			var suites *junit.Suites
```
spiffxp, line 161: ???

this seems unrelated

PR #21401, opened 2021-03-16
fix image-builder validate-all test

codenrhoden:
Need to point to Makefile in images/capi folder

/assign @CecileRobertMichon 

Me working constantly with a `CWD` of `images/capi` bites me again...


CecileRobertMichon:
/lgtm

good thing we tested it 😄

PR #21101, opened 2021-03-01
[wip] autobumper: support base branches other than master

spiffxp:
default to `main`, updated autobump-config.yaml for test-infra to use `master`

needed this to use autobumper to open https://github.com/kubernetes/k8s.io/pull/1740

PR #21403, opened 2021-03-16
Autobumper: support target branch name other than master branch

chaodaiG:


cjwagner (approved):
/hold cancel

prow/cmd/generic-autobumper/bumper/bumper.go
```diff
@@ -89,6 +90,8 @@ type Options struct {
 	GitHubOrg string `yaml:"gitHubOrg"`
 	// The target GitHub repo name where the autobump PR will be created. Only required when SkipPullRequest is false.
 	GitHubRepo string `yaml:"gitHubRepo"`
+	// GitHubTargetBranch is the target GitHub branch where the autobump PR will be created against with. Default is master.
+	GitHubTargetBranch string `yaml:"githubGitHubTargetBranch"`
```
spiffxp, line 94: Just came across this, I was getting back around to https://github.com/kubernetes/test-infra/pull/21101 and noticed  you opted to default to `master` instead of auto-detect.  I can try adding that.

How locked are we into this API?  I think would have preferred `gitHubBaseBranch` for consistency with GitHub's terminology and the other fields in this struct

chaodaiG: Ah. sorry totally forgot #21101 when implementing this. It was mainly for unblocking myself from bumping knative prow jobs since they already switched over to main.

So afaict it's only used here https://github.com/knative/test-infra/blob/5273590479dfaa642f6a37e6dbb0bcb182047847/config/prod/prow/knative-autobump-config.yaml#L6, feel free making any change to it

PR #21405, opened 2021-03-16
Add basic sharding for branchprotection config

alvaroaleman:
This change adds the ability to read additional prow config from additional folders. It will only work for branchprotection config, and everything else will error, because there is no merging logic implemented yet. The merging logic doesn't allow any redundant config. This is the first step towards https://docs.google.com/document/d/1S8DFXrvFw27xYyHMDZmtUkVvd0cbDzFI8T5H35FkCyU/edit#heading=h.1sfy56v9ifz6

What is not yet included at all and will be added in the future is a new validation that checks that the additional configs are in a org/repo hirarchy and that settings are only set at the appropriate level in that hirarchy.


prow/config/config.go
```diff
@@ -2386,3 +2428,14 @@ func StringsToOrgRepos(vs []string) []OrgRepo {
 	}
 	return vsm
 }
+
+// mergeFrom merges two prow configs. It must be called _before_ doing any
+// defaulting.
+func (c *ProwConfig) mergeFrom(additional *ProwConfig) error {
```
chaodaiG, line 2434: nit: `pc` instead of `c`

prow/config/config.go
```diff
@@ -1132,7 +1171,10 @@ func yamlToConfig(path string, nc interface{}) error {
 		jc = v
 	case *Config:
 		jc = &v.JobConfig
+	default:
```
chaodaiG, line 1174: Could you add a comment explaining why return nil instead of error?

prow/config/config.go
```diff
@@ -1070,6 +1071,44 @@ func loadConfig(prowConfig, jobConfig string) (*Config, error) {
...
+		var errs []error
+		errs = append(errs, filepath.Walk(additionalProwConfigDirs, func(path string, info os.FileInfo, err error) error {
+			if err != nil {
+				// Finish walking and handle all errors in bulk at the end, otherwise this is annoying as a user
+				errs = append(errs, err)
+				return nil
+			}
+			if strings.HasPrefix(info.Name(), "..") {
+				// kubernetes volumes also include files we
+				// should not look be looking into for keys
```
chaodaiG, line 1084: Not quite sure I understand this comment

prow/config/branch_protection.go
```diff
@@ -176,15 +178,88 @@ type BranchProtection struct {
...
+func isPolicyEmpty(p Policy) bool {
+	return apiequality.Semantic.DeepEqual(p, Policy{})
+}
+
+func (bp *BranchProtection) merge(additional *BranchProtection) error {
+	var errs []error
+	if !isPolicyEmpty(bp.Policy) && !isPolicyEmpty(additional.Policy) {
+		errs = append(errs, errors.New("both brachprotection configs set a top-level policy"))
+	} else {
+		bp.Policy = additional.Policy
```
chaodaiG, line 201: Is it expected to replace `bp.Policy` with empty `additional.Policy`?

prow/config/config.go
alvaroaleman: Fixed

prow/config/config.go
alvaroaleman: Done

prow/config/config.go
alvaroaleman: I've extended the comment, is it sufficent?

prow/config/branch_protection.go
alvaroaleman: Do you mean if its expected that if both `bp.Policy` and `additional.Policy` are empty we will overwrite the former with the latter? If yes, well its not explicitly wanted but it effectively a no-op and not extra handling this case keeps the code simpler

chaodaiG: I meant when `bp.Policy` is not empty and `additional.Policy` is empty, the `else` condition is executed, which would override `bp.Policy` with the empty `additional.Policy`

alvaroaleman:
Updated the PR description to be a bit less terse and also added some test that the actual loading works. This PR ended up being quite big, because I had to insert a new flag into almost all binaries we have and we have lots of them. It should be relatively straight forward to review, though.

prow/config/branch_protection.go
chaodaiG: I might be wrong, `&&` condition for two negations is a little bit mind twisting :)

alvaroaleman: Oh you are right, that was absolutely not intended, very good catch. Fixed it and added a `TestBranchprotectionMergeRemainsUnchangedWhenMergingEmptyIn`. 

Also renamed `isPolicyEmpty` to `isPolicySet` to avoid all the negations.

Only good thing about this messy code is that it's very easy to test via fuzzing/property based testing.

chaodaiG (approved):
lgtm

/hold
/cc @cjwagner
For awareness of oncall

PR #21408, opened 2021-03-16
Prow bot automatic accept admin invitation by hmac tool

chaodaiG:
As part of prow onboarding steps, prow oncall need to login into prow robot account on github UI and accept the invitation as org/repo admin on behalf of prow robot, this is toilsome. 

Extending hmac tool(Prow hmac token auto rotation tool that also sets up webhooks) to accept pending invitations to eliminate oncall doing this manually. This feature is turned off by default so it's backward compatible, it requires a prow config change to turn on


chaodaiG:
Half of the credit goes to @cjwagner since we pair programmed most of the parts.

/cc @cjwagner @alvaroaleman

prow/github/client.go
```diff
@@ -1347,6 +1351,87 @@ func (c *client) ListOrgInvitations(org string) ([]OrgInvitation, error) {
...
+func (c *client) ListCurrentUserRepoInvitations() ([]UserRepoInvitation, error) {
+	c.log("ListCurrentUserRepoInvitations")
+	if c.fake {
+		return nil, nil
+	}
+	path := "/user/repository_invitations"
+	var ret []UserRepoInvitation
+	err := c.readPaginatedResults(
+		path,
+		acceptNone,
```
cjwagner, line 1366: Please ensure the accept headers match the docs. e.g. this one should be `application/vnd.github.v3+json`.

prow/github/client.go
```diff
@@ -1347,6 +1351,87 @@ func (c *client) ListOrgInvitations(org string) ([]OrgInvitation, error) {
...
+		method:    http.MethodPatch,
+		path:      fmt.Sprintf("/user/repository_invitations/%d", invitationID),
+		org:       "",
+		exitCodes: []int{204},
+	}, nil)
+
+	return err
+}
+
+func (c *client) ListCurrentUserOrgInvitations() ([]UserOrgInvitation, error) {
```
cjwagner, line 1397: We need to include the `state=pending` query parameter otherwise we will get both invitations and already active org memberships.

prow/cmd/hmac/fakeghhook/fakeghhook.go
```diff
@@ -123,6 +126,52 @@ func (f *FakeClient) DeleteRepoHook(org, repo string, id int, req github.HookReq
...
+	// Must exist
+	found := -1
+	for idx, iv := range f.UserRepoInvitations {
+		if iv.InvitationID == invitationID {
+			found = idx
+			break
+		}
+	}
+	if found != -1 {
+		f.UserRepoInvitations = append(f.UserRepoInvitations[:found], f.UserRepoInvitations[found+1:]...)
```
alvaroaleman, line 147: this can panic or not? If we have exactly one, we will end up calling `f.UserRepoInvitations[1:]...` on an one-element slice, right?

prow/cmd/hmac/fakeghhook/fakeghhook.go
```diff
@@ -123,6 +126,52 @@ func (f *FakeClient) DeleteRepoHook(org, repo string, id int, req github.HookReq
...
+	// Must exist
+	found := -1
+	for idx, iv := range f.UserOrgInvitations {
+		if iv.Org.Login == org {
+			found = idx
+			break
+		}
+	}
+	if found != -1 {
+		f.UserOrgInvitations = append(f.UserOrgInvitations[:found], f.UserOrgInvitations[found+1:]...)
```
alvaroaleman, line 170: Same here regarding potential out of bounds

prow/cmd/hmac/main_test.go
```diff
@@ -653,3 +654,166 @@ func TestBatchOnboardNewTokenForRepos(t *testing.T) {
...
+func TestHandleInvitation(t *testing.T) {
+	tests := []struct {
+		name          string
+		urivs         []github.UserRepoInvitation
+		uoivs         []github.UserOrgInvitation
+		errs          []error
+		newHMACConfig config.ManagedWebhooks
+		wantUrivs     []github.UserRepoInvitation
+		wantUoivs     []github.UserOrgInvitation
+		wantErr       bool
```
alvaroaleman, line 667: Using booleans to indicate an expected error instead of substring matching or type matching has the issue that it has a lot of potential to rot over time if it ends up erroring for a different reason, resulting in the functionality breaking but the test still passing

prow/cmd/hmac/main.go
```diff
@@ -176,11 +176,70 @@ func main() {
...
+			}
+		}
+	}
+	// Accept org invitation
+	orgIvs, err := c.githubHookClient.ListCurrentUserOrgInvitations()
+	if err != nil {
+		return err
+	}
+	for _, iv := range orgIvs {
+		if github.RepoPermissionLevel(iv.Role) != github.Admin {
```
cjwagner, line 225: nit: Repository permissions are different from  organization roles.
[Repository permissions](https://docs.github.com/en/rest/reference/repos#get-repository-permissions-for-a-user) can be `admin`, `write`, `read`, or `none`.
[Organization roles](https://docs.github.com/en/rest/reference/orgs#get-organization-membership-for-a-user) can be `admin`, `member`, and maybe some more options.

So this will work in practice since both have an `admin` option, but it would be a bit better to cast this to a different type to distinguish this from repository permissions.

prow/github/client_test.go
```diff
@@ -2761,7 +2761,7 @@ func TestAllMethodsThatDoRequestSetOrgHeader(t *testing.T) {
 		t.Run(clientType.Method(i).Name, func(t *testing.T) {
 
 			checkingRoundTripper := testRoundTripper{func(r *http.Request) (*http.Response, error) {
-				if !strings.HasPrefix(r.URL.Path, "/app") {
+				if !strings.HasPrefix(r.URL.Path, "/app") && !strings.HasPrefix(r.URL.Path, "/user/repository_invitations") && !strings.HasPrefix(r.URL.Path, "/user/memberships/orgs") {
```
alvaroaleman, line 2764: Could we instead add the method names into the toSkip set and a small go doc (I guess a gh app can not be invited, right?)

prow/cmd/hmac/main.go
```diff
@@ -176,11 +176,70 @@ func main() {
...
+			// reasons for not considering org matching:
+			// 1. The FullName is org/repo
+			// 2. If an org is defined as managed webhook but only invite
+			// bot as admin on repo level, the webhook setup will fail
+			// 3. Also we are not ready to receive spamming webhook from the
+			// org if it only configured a repo in hmac
+			if iv.Repository.FullName == repoName {
+				if err := c.githubHookClient.AcceptUserRepoInvitation(iv.InvitationID); err != nil {
+					return fmt.Errorf("failed accepting repo invitation: %w", err)
+				}
```
cjwagner, line 215: nit: Consider adding an Info level log message for successful invitation acceptance.

prow/github/client.go
```diff
@@ -1347,6 +1351,87 @@ func (c *client) ListOrgInvitations(org string) ([]OrgInvitation, error) {
...
+}
+
+func (c *client) AcceptUserOrgInvitation(org string) error {
+	c.log("AcceptUserOrgInvitation", org)
+
+	_, err := c.request(&request{
+		method:      http.MethodPatch,
+		path:        fmt.Sprintf("/user/memberships/orgs/%s", org),
+		org:         "",
+		requestBody: &struct{ org, state string }{org: org, state: "active"},
```
cjwagner, line 1428: The org should not be part of the request body. It is specified via the path: https://docs.github.com/en/rest/reference/orgs#update-an-organization-membership-for-the-authenticated-user

prow/github/client.go
```diff
@@ -1347,6 +1351,87 @@ func (c *client) ListOrgInvitations(org string) ([]OrgInvitation, error) {
...
+	return ret, nil
+}
+
+func (c *client) AcceptUserOrgInvitation(org string) error {
+	c.log("AcceptUserOrgInvitation", org)
+
+	_, err := c.request(&request{
+		method:      http.MethodPatch,
+		path:        fmt.Sprintf("/user/memberships/orgs/%s", org),
+		org:         "",
```
cjwagner, line 1427: nit: We can populate the org here.

prow/github/client_test.go
chaodaiG: Moved

prow/cmd/hmac/fakeghhook/fakeghhook.go
```diff
@@ -123,6 +126,52 @@ func (f *FakeClient) DeleteRepoHook(org, repo string, id int, req github.HookReq
...
+	found := -1
+	for idx, iv := range f.UserRepoInvitations {
+		if iv.InvitationID == invitationID {
+			found = idx
+			break
+		}
+	}
+	if found != -1 {
+		f.UserRepoInvitations = append(f.UserRepoInvitations[:found], f.UserRepoInvitations[found+1:]...)
+	}
```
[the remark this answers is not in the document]
cjwagner: Hmm, I don't really see how this `getNextError` pattern is better then. Doesn't that require us to anticipate every error in the test cases rather than catching errors we didn't anticipate? For example, if we try to accept an invite that does not exist in a unit test and we didn't instrument the test case to emit an error because we didn't expect it, we would incorrectly pass the test case. Basically it seems like making the error generation entirely manual causes us to skip validation everywhere we don't expect to need it.

prow/github/client.go
cjwagner: See the docs here: https://docs.github.com/en/rest/reference/orgs#list-organization-memberships-for-the-authenticated-user
> Indicates the state of the memberships to return. Can be either active or pending. If not specified, the API returns both active and pending memberships.

We have a version of the pagination handler that handles query parameters. Please use that instead: https://github.com/kubernetes/test-infra/blob/ea8f981918d11fec80d037ad90768487ce18780a/prow/github/client.go#L1724-L1725

prow/cmd/hmac/fakeghhook/fakeghhook.go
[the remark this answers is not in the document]
chaodaiG: `getNextError` can better control what error is returned for these github API wrappers, for example it can return error types other than 404 not found. But I agree that it's an overkill here, will remove `getnextError` here

prow/cmd/hmac/main.go
```diff
@@ -178,11 +178,72 @@ func main() {
...
+		return nil
+	}
+	// Accept repos invitations first
+	repoIvs, err := c.githubHookClient.ListCurrentUserRepoInvitations()
+	if err != nil {
+		return err
+	}
+	for _, iv := range repoIvs {
+		if iv.Permission != github.Admin {
+			logrus.Errorf("invalid invitation from %s not accepted. Want: %v, got: %s",
```
chizhg, line 202: Super nit: it'll be more clear to state the error is due to permission
```suggestion
			logrus.Errorf("invalid invitation from %s is not accepted. Permission want: %v, got: %s",
```

prow/github/client.go
chaodaiG: cool, done.

cjwagner (approved):
Looks awesome, I'm excited to see it in action.

PR #21410, opened 2021-03-17
recover from panics in plugins of hook

### LINKED ISSUE #21098: Hook should recover panics in plugins

Currently, a panic in a hook plugin will bring down the whole binary. We should instead recover plugin panics and log them as errors.

/area hook
/help

**Why is this needed**:

ykakarap:
This PR is aims to fix #21098 

This PR captures if any of the hook plugins panics and returns them as errors. With this if one plugin panics all other plugins will not stop working.

### DIFF
--- a/prow/hook/events.go
+++ b/prow/hook/events.go
@@ -17,6 +17,8 @@ limitations under the License.
 package hook
 
 import (
+	"fmt"
+	"runtime/debug"
 	"time"
 
 	"github.com/prometheus/client_golang/prometheus"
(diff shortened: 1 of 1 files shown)


nikhita:
Thanks @ykakarap! 

/assign @alvaroaleman

prow/hook/events.go
```diff
@@ -430,10 +432,22 @@ func (s *Server) handleGenericComment(l *logrus.Entry, ce *github.GenericComment
...
 			start := time.Now()
 			labels := prometheus.Labels{"event_type": l.Data[eventTypeField].(string), "action": string(ce.Action), "plugin": p}
-			if err := h(agent, *ce); err != nil {
+			if err := func() (err error) {
+				defer func() {
+					if r := recover(); r != nil {
+						err = fmt.Errorf("plugin '%v' paniced with %v. stack is : \n%s", p, r, debug.Stack())
+					}
+				}()
+				err = h(agent, *ce)
```
alvaroaleman, line 441: Nit: You can just `return err` here

alvaroaleman: And yeah, given that we have a big bunch of handlers it might make sense to build a reusable func for this that itself takes a `func() error`

alvaroaleman: I think `events.go` is fine

ykakarap:
@alvaroaleman this PR is now ready for review 🙂

alvaroaleman (approved):
thanks!

PR #21420, opened 2021-03-17
[Kettle] Edit the error handling of large requests

MushuEE:
I am seeing errors in handling the BadRequest Exception:
```
UnicodeEncodeError: 'ascii' codec can't encode characters in position 4717021-4717023: ordinal not in range(128)
```
The fact that the character position is that large reenforces the thought that we are attempting to send some chunks that have FAR too much data

/area kettle

at minimum this should stop kettle from erroring out. I should think of a way to report what jobs are getting this much data collected


amwat (commented):
/lgtm
/approve

amwat (commented):
/lgtm

PR #21422, opened 2021-03-17
Retry GetRepos() as for user when it fails for org

petr-muller:
The code assumes the "owner" is an org, but nothing prevents it from
being a username instead, where the API call is unfortunately different
and we would fail on it.

We can retry the call for a user if it fails for an org. This increases
consumption of tokens when there are real errors and/or when there are
user repos configured (but these previously failed, so IMO it's worth
it).

Took the opportunity and removed an unused param from an unexported
method.

### DIFF
--- a/prow/pluginhelp/hook/hook.go
+++ b/prow/pluginhelp/hook/hook.go
@@ -31,8 +31,9 @@ import (
 	"time"
 
 	"github.com/sirupsen/logrus"
-
+	"k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/sets"
+
 	prowconfig "k8s.io/test-infra/prow/config"
 	"k8s.io/test-infra/prow/github"
 	"k8s.io/test-infra/prow/pluginhelp"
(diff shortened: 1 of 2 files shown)


prow/pluginhelp/hook/hook.go
```diff
@@ -303,6 +304,22 @@ func (oa *orgAgent) orgToReposMap(config *plugins.Configuration) map[string]sets
...
+func reposForOrgOrUser(ghc githubClient, orgOrUser string) ([]github.Repo, error) {
+	var errs [2]error
+	inError := map[bool]string{true: "user", false: "org"}
+
+	// Check as an org first, it is more likely in normal use case
+	for i, isUser := range []bool{false, true} {
+		if repos, err := ghc.GetRepos(orgOrUser, isUser); err == nil {
+			return repos, nil
+		} else {
+			errs[i] = fmt.Errorf("failed to get repos for %s %s: %w", inError[isUser], orgOrUser, err)
```
chaodaiG, line 316: Should this error be swallowed when the use case is user? Or in another word, does `GetRepos(<user>, false)` return error?

petr-muller: I'm not sure I understand the question correctly. When we call `GetRepos(<user>, false)` (that's what the current code does) we get a 404 error. The `false` results in getting repos for an **org**, and there isn't an org with that name. In the current state this gets logged as an error (that's how I noticed it) and it is skipped.

So the new wrapper always does one of the following: Either one of the two attempts passes and no error is returned, or both fail and we return an aggregate of the two errors that we hit: one of them will always be a 404 and one will the the "real" one.

chaodaiG: I see, sorry missed the return above.

chaodaiG (approved):

PR #21431, opened 2021-03-18
create an assisted-installer ownership

osherdp:
/hold
Just so we can make changes more freely.


osherdp:
I'll get to this later :)

PR #21436, opened 2021-03-18
sidecar: emit debug logs by default

stevekuznetsov:
Anyone looking at the logs for this utility will need them anyway, we
run by default with logging at debug in production for the rest of Prow
today so this matches that pattern.


AlexNPavel (approved):
/lgtm

PR #21437, opened 2021-03-18
sidecar: place temp files on the same device

stevekuznetsov:
We want to be able to do a simple rename to overwrite the source file
with the censored data, and we can't do that across device boundaries,
which exist between volume mounts in a k8s Pod. Placing temporary files
in the same directory as the files they will eventually replace ensures
that renames never happen across device boundaries.

### DIFF
--- a/prow/sidecar/censor.go
+++ b/prow/sidecar/censor.go
@@ -260,7 +260,9 @@ func validRelPath(p string) bool {
 
 // archive re-packs the dir into the destination
 func archive(srcDir, destArchive string) error {
-	output, err := ioutil.TempFile("", "tmp-archive")
+	// we want the temporary file we use for output to be in the same directory as the real destination, so
+	// we can be certain that our final os.Rename() call will not have to operate across a device boundary
+	output, err := ioutil.TempFile(filepath.Dir(destArchive), "tmp-archive")
 	if err != nil {
 		return fmt.Errorf("failed to create temporary file for archive: %w", err)
 	}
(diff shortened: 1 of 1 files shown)


prow/sidecar/censor.go
```diff
@@ -260,7 +260,9 @@ func validRelPath(p string) bool {
 
 // archive re-packs the dir into the destination
 func archive(srcDir, destArchive string) error {
-	output, err := ioutil.TempFile("", "tmp-archive")
+	// we want the temporary file we use for output to be in the same directory as the real destination, so
+	// we can be certain that our final os.Rename() call will not have to operate across a device boundary
+	output, err := os.Create(fmt.Sprintf("%s.tmp-archive", destArchive))
```
BenTheElder, line 265: why not:
```suggestion
	output, err := ioutil.TempFile(filepath.Dir(destArchive), "tmp-archive"))
```
`TempFile` can use the dest dir but still retain the non-conflicting random filename suffix behavior

stevekuznetsov: Ah, gotcha - misread the `ioutil.Tempdir` doc, thought it put stuff under `$TMP` no matter what

alvaroaleman (approved):

prow/sidecar/censor.go
```diff
@@ -260,7 +260,9 @@ func validRelPath(p string) bool {
 
 // archive re-packs the dir into the destination
 func archive(srcDir, destArchive string) error {
-	output, err := ioutil.TempFile("", "tmp-archive")
+	// we want the temporary file we use for output to be in the same directory as the real destination, so
+	// we can be certain that our final os.Rename() call will not have to operate across a device boundary
+	output, err := ioutil.TempFile(filepath.Dir(destArchive), "tmp-archive"))
```
BenTheElder, line 265: oops, these have an extra trailing `)` which fails the build.

alvaroaleman (approved):

stevekuznetsov:
Now with 100% fewer closing parentheses

alvaroaleman (approved):

PR #21424, opened 2021-03-17
stop specifying master branch in optional kubernetes kind presubmits

BenTheElder:
- use kind from artifact
- use decorated job instead of bootstrap
- use skip_branches instead of branches
xref: https://github.com/kubernetes-sigs/kind/issues/2120

PR #21439, opened 2021-03-18
add missing path_alias

BenTheElder:
follow up to https://github.com/kubernetes/test-infra/pull/21424


config/jobs/kubernetes/sig-testing/conformance-e2e.yaml
```diff
@@ -47,6 +47,7 @@ presubmits:
 
   - name: pull-kubernetes-conformance-kind-ipv6-parallel
     decorate: true
+    path_alis: k8s.io/kubernetes
```
aojea, line 50: Typo Ali's

PR #21441, opened 2021-03-18
Bugzilla: Copy blocks field when cloning bugs

alvaroaleman:


prow/bugzilla/types.go
```diff
@@ -226,6 +226,7 @@ type Flag struct {
 type BugUpdate struct {
 	// DependsOn specifies the bugs that this bug depends on
 	DependsOn  *IDUpdate `json:"depends_on,omitempty"`
+	Blocks     *IDUpdate `json:"blocks,omitempty"`
```
stevekuznetsov, line 229: Of note: even though this is the inverse of the `DependsOn` field, they are entirely unrelated in the API.

stevekuznetsov (approved):
This is a good first pass, hopefully we don't need a more targeted approach.

PR #21442, opened 2021-03-18
sidecar: first-class support for registry credentials

stevekuznetsov:
sidecar: compare fixtures in the other direction

Previously, we just ensured that all of the output in the golden
directory existed in the output from the test. This obviously fails to
work when the test changes and produces new content. Instead, check that
everything in the output matched the golden files.

---

sidecar: first-class support for registry credentials

Kubernetes has two first-class Secret types that store registry
credentials. While in the common case we cannot know which specific part
of a secret is important, in these cases it is possible to censor not
only the whole registry credential, but also each individual item in the
credentials file that is sensitive in its own right.

---

The first commit really shows how annoying it is to do `diff a b` in Go as opposed to writing these in shell :|


prow/sidecar/censor.go
```diff
@@ -418,10 +419,79 @@ func loadSecrets(paths []string) ([][]byte, error) {
...
+			extra, parseErr := parser(raw)
+			if parseErr != nil {
+				return fmt.Errorf("could not read %s as a docker secret: %v", path, parseErr)
+			}
+			// It is important that these are added to the list of secrets *after* their parent data
+			// as we will censor in order and this will give a reasonable guarantee that the parent
+			// data (a superset of any of these fields) will be censored in its entirety, first. It
+			// remains possible that the sliding window used to censor pulls in only part of the
+			// superset and some small part of it is censored first, making the larger superset no
+			// longer match the file being censored.
```
stevekuznetsov, line 440: We could double-parse the file if we *really* wanted to enforce this, or we could make this behavior opt-in. However, I think it's probably best as-is ... if we hit this case, we will still censor all of the "sensitive" parts of the larger secret and users have and will continue to leak just the important part of a pull secret, not the whole thing in all of it's JSON glory.

prow/testutil/fixtures.go
```diff
@@ -29,19 +29,19 @@ import (
 
 // CompareWithFixtureDir will compare all files in a directory with a corresponding test fixture directory.
 func CompareWithFixtureDir(t *testing.T, golden, output string) {
-	if walkErr := filepath.Walk(golden, func(path string, info os.FileInfo, err error) error {
+	if walkErr := filepath.Walk(output, func(path string, info os.FileInfo, err error) error {
 		if err != nil {
 			return err
 		}
-		if info.IsDir() {
+		if info.IsDir() || info.Mode()&os.ModeSymlink == os.ModeSymlink {
```
stevekuznetsov, line 36:
```
 --- FAIL: TestCensorIntegration (0.01s)
    fixtures.go:68: failed to read testdata file: open testdata/output/artifacts/first/link: no such file or directory 
```

I explicitly do *not* look at symlinks in this test, so seeing this failure makes me think that somehow this isn't a symlink in bazel-land. Sigh.

stevekuznetsov:
@cjwagner @fejta ok cool so `bazel` reads my symlinks and gets rid of them. By the time the test runs, there are no symlinks, just normal files. How do I get around this? My code handles symlinks specifically.

stevekuznetsov:
@BenTheElder took your suggestion of writing the symlinks at runtime to circumvent the build sandbox thing.

prow/sidecar/censor.go
```diff
@@ -418,10 +419,79 @@ func loadSecrets(paths []string) ([][]byte, error) {
...
 			return nil
 		}); err != nil {
 			return nil, err
 		}
 	}
 	return secrets, nil
 }
+
+// loadDockercfgAuths parses auth values from a kubernetes.io/dockercfg secret
+func loadDockercfgAuths(content []byte) ([]string, error) {
```
alvaroaleman, line 453: Could you explain why we need this? In case someone parses and extracts auth items from a pull secret?

prow/sidecar/censor.go
```diff
@@ -418,10 +419,79 @@ func loadSecrets(paths []string) ([][]byte, error) {
 				return err
 			}
 			secrets = append(secrets, raw)
+			var parser = func(bytes []byte) ([]string, error) {
+				return []string{}, nil
```
alvaroaleman, line 423: Nit:  No reason to construct an empty slice, you can just return nil

prow/sidecar/censor.go
stevekuznetsov: Explain here in a comment or in the code? Yes, it's to guard the items if they're extracted. Or imagine a tool that shows part of the output of a test, which cuts a secret in half - this helps us censor the important bits even if we can't do the whole thing.

alvaroaleman (approved):

PR #21443, opened 2021-03-19
KRTE cleanup

BenTheElder:
/cc @amwat


images/krte/Dockerfile
```diff
@@ -26,7 +26,7 @@ ARG GO_VERSION
 # - adding gsutil etc. to path (where we will install them)
 # - disabling prompts when installing gsutil etc.
 # - hinting that we are in a docker container
-ENV IMAGE=${IMAGE_ARG} \
+ENV KRTE_IMAGE=${IMAGE_ARG} \
```
BenTheElder, line 29: kubekins should also do that. will leave to another PR.
(`IMAGE` is commonly used in makefiles etc.)

amwat (commented):
/lgtm
/hold

This doesn't auto bump existing jobs does it? else we'll need to move the current kubetest2 jobs to kubekins simultaneously.

BenTheElder:
> This doesn't auto bump existing jobs does it? else we'll need to move the current kubetest2 jobs to kubekins simultaneously.

Nope, we can file another PR to migrate before the autobump

BenTheElder:
/hold cancel
@amwat  moved the jobs off

images/krte/Dockerfile
```diff
@@ -16,32 +16,25 @@
 # NOTE: we attempt to avoid unnecessary tools and image layers while
 # supporting kubernetes builds, kind installation, etc.
 
-ARG OLD_BAZEL_VERSION
-FROM launcher.gcr.io/google/bazel:${OLD_BAZEL_VERSION} as old-bazel
-FROM debian:buster
```
BenTheElder, line 21: we should still have this, oops

PR #21453, opened 2021-03-19
Sinker: Instantly clean duplicate prowjobs

alvaroaleman:
This change adds a feature gated addition to sinker to make it delete
duplicate prowjobs. Prowjobs that are not periodics, have finished and
have the same name, refs, extrarefs and result are considered
duplicates.

The reason for wanting this is to allow increasing the prowjob retention
without overloading the api. The prowjob retention is important,
because we always have to repeat tests for prs after max_prowjob_age.

/cc @chaodaiG 

If we don't find any issues with this, I would also like to enable it by default and remove the feature gate. WDYT?

Edit: There are also a couple of additional ways to get rid of more useless prowjob instances in the system:
* We can always delete completed aborted jobs
* If we have multiple results for the same job and revision, we can delete everything except the best 
* We can check if the base/head revision of a job are current and if not, delete it (that will cost api budget, though)

### DIFF
--- a/prow/cmd/sinker/main.go
+++ b/prow/cmd/sinker/main.go
@@ -18,9 +18,12 @@ package main
 
 import (
 	"context"
+	"encoding/json"
 	"flag"
 	"fmt"
 	"os"
+	"sort"
+	"strings"
 	"time"
 
 	"github.com/pkg/errors"
@@ -50,6 +53,7 @@ import (
 
 type options struct {
 	runOnce                    bool
+	cleanDuplicates            bool
 	configPath                 string
 	jobConfigPath              string
 	supplementalProwConfigDirs flagutil.Strings
@@ -64,12 +68,14 @@ const (
 	reasonPodTTLed    = "ttled"
 
 	reasonProwJobAged         = "aged"
+	reasonDuplicate           = "duplicate"
 	reasonProwJobAgedPeriodic = "aged-periodic"
 )
 
 func gatherOptions(fs *flag.FlagSet, args ...string) options {
 	o := options{}
 	fs.BoolVar(&o.runOnce, "run-once", false, "If true, run only once then quit.")
+	fs.BoolVar(&o.cleanDuplicates, "clean-duplicate-prowjobs", false, "If sinker should instantly clean up duplicate prowjobs. Jobs that are not a periodic, did complete and have the same name, refs, extra refs and state are considered duplicates.")
 	fs.StringVar(&o.configPath, "config-path", "", "Path to config.yaml.")
 	fs.StringVar(&o.jobConfigPath, "job-config-path", "", "Path to prow job configs.")
 	fs.Var(&o.supplementalProwConfigDirs, "supplemental-prow-config-dir", "An additional directory from which to load prow configs. Can be used for config sharding but only supports a subset of the config. The flag can be passed multiple times.")
@@ -191,6 +197,12 @@ type controller struct {
 	podClients    map[string]ctrlruntimeclient.Client
 	config        config.Getter
 	runOnce       bool
+	// cleanDuplicates makes sinker delete all duplicate jobs
+	// that are completed and not a periodic. Jobs are deemed
+	// a duplicate if their .spec.job, .spec.refs, .spec.extra_refs
+	// and .status.state fields are identical. Keeping these
+	// doesn't provide any value.
+	cleanDuplicates bool
 }
 
 func (c *controller) Start(ctx context.Context) error {
@@ -313,8 +325,17 @@ func (c *controller) clean() {
 	metrics.prowJobsCreated = len(prowJobs.Items)
 
 	// Only delete pod if its prowjob is marked as finished
-	pjMap := map[string]*prowapi.ProwJob{}
+	pjMap := make(map[string]*prowapi.ProwJob, len(prowJobs.Items))
 	isFinished := sets.NewString()
+
+	// Sort by creationTimestamp so we always keep the youngest if we deduplicate jobs.
+	// We must sort before we construct the pjMap, otherwise the pjMap will be wrong
+	// if the sorting swapped anything. This is because its values are pointers pointing
+	// to elements in prowjob.Items which is a slice of values and those values are
+	// changed when swapping.
+	sort.Slice(prowJobs.Items, func(i, j int) bool {
+		return prowJobs.Items[i].CreationTimestamp.After(prowJobs.Items[j].CreationTimestamp.Time)
+	})
 
 	maxProwJobAge := c.config().Sinker.MaxProwJobAge.Duration
 	for i, prowJob := range prowJobs.Items {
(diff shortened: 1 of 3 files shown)


chaodaiG (commented):
Generally makes sense to me. One thing worth discussion, is that deck depends on prowjobs, so test failed due to flaky tests will not be surfaced on deck, which is a little bit strange and sometimes could even be problematic if someone is tracking flaky tests on deck. Do we have a solution for this? Like where else can we find the same info?

alvaroaleman:
> Do we have a solution for this? Like where else can we find the same info?

The job history page is gcs-backed: https://prow.k8s.io/job-history/gs/kubernetes-jenkins/pr-logs/directory/pull-test-infra-bazel

chaodaiG:
> The job history page is gcs-backed: https://prow.k8s.io/job-history/gs/kubernetes-jenkins/pr-logs/directory/pull-test-infra-bazel

The job history page isn't very user friendly:
```
1373007811180498944 | 2021-03-19 20:26:14 +0000 UTC | 5s | SUCCESS
```
This doesn't tell me much, but yeah it's better than nothing. We can think of modernize that page in the future

alvaroaleman:
> This doesn't tell me much, but yeah it's better than nothing. We can think of modernize that page in the future

Yeah, I do agree that page isn't as useful.

The fundamental issue we have is that we use kube crds to store a lot of data and they are simply not made for that. This wouldn't be a big deal if they were just for humans after a job finished, but since they are also used to check if retesting is needed, it is an issue. In openshift, we currently have a 24h retention only and still peak at more than 17k crs. This small retention means that we almost always have to retest, because reviews in most cases take more than 24h. I wasn't really consciously aware of this unfortunate connection either tbh.

prow/cmd/sinker/main.go
```diff
@@ -532,3 +560,78 @@ func podNeedsKubernetesFinalizerCleanup(log *logrus.Entry, pj *prowapi.ProwJob,
...
+
+		duplicates[key] = append(duplicates[key], pj.Name)
+	}
+
+	var result []string
+	for _, duplicateSlice := range duplicates {
+		if len(duplicateSlice) == 1 {
+			continue
+		}
+		result = append(result, duplicateSlice[1:]...)
```
chaodaiG, line 603: I assume this is based on the assumption of pjs are sorted based on start timestamp, is this always true? Just want to make sure we are not deleting latest run by accident

chaodaiG:
> > This doesn't tell me much, but yeah it's better than nothing. We can think of modernize that page in the future
> 
> Yeah, I do agree that page isn't as useful.
> 
> The fundamental issue we have is that we use kube crds to store a lot of data and they are simply not made for that. This wouldn't be a big deal if they were just for humans after a job finished, but since they are also used to check if retesting is needed, it is an issue. In openshift, we currently have a 24h retention only and still peak at more than 17k crs. This small retention means that we almost always have to retest, because reviews in most cases take more than 24h. I wasn't really consciously aware of this unfortunate connection either tbh.

Thank you for the context. Due to this unfortunate connection I totally agree that we should start with opting out by default, and we can decide whether to switch the default in the future

prow/cmd/sinker/main.go
```diff
@@ -532,3 +560,78 @@ func podNeedsKubernetesFinalizerCleanup(log *logrus.Entry, pj *prowapi.ProwJob,
...
+		if pj.Spec.Type == prowapi.PeriodicJob || !pj.Complete() {
+			continue
+		}
+
+		key.job = pj.Spec.Job
+		key.refs, err = comparableRefsString(pj.Spec.Refs)
+		if err != nil {
+			return nil, err
+		}
+		key.extraRefs, err = comparableExtraRefsString(pj.Spec.ExtraRefs)
```
cjwagner, line 589: nit: All this marshalling will probably be pretty expensive. I don't really see a better alternative, but it might be worth logging the time we spend in `getDuplicates`.

cjwagner (commented):
Just 1 nit and a +1 for Chao's review comment.

Due to the negative impact on the front end UX that Chao mentioned I don't think we'd want to enable this on Prow instances that we manage as is. I think it would be too confusing for previous job runs to disapear from the front end when they are rerun. The positive impact on Tide retesting would be very significant though so gating this feature with a flag sounds great. I'd like to take advantage of this eventually if we can work around the UI problems.

Having a structured datastore for job metadata would solve this and other problems, but is a more involved change. @fejta is looking into something like this.
A simpler solutions that would mitigate just this problem might be to start annotating ProwJobs with the buildIDs of their duplicates. That would allow the front end to indicate that a job has previous duplicate runs and link to them in Spyglass/GCS without having a ProwJob resource for the duplicates.

cjwagner:
> * We can check if the base/head revision of a job are current and if not, delete it (that will cost api budget, though)

Just a minor note, but if we did this from a plugin we could do it for free using webhook data.


Another way we could more aggressively GC ProwJobs that aren't helpful for tide would be to allow shorter age limits for periodic and postsubmit ProwJobs. Maybe even only keep the last X (1?) results for postsubmit and periodic jobs? I think most users mainly care about presubmits so that might have less overall effect on average front end UX. It looks like 1/3 of the current ProwJob resources for https://prow.ci.openshift.org/ are postsubmits or periodics.

prow/cmd/sinker/main.go
```diff
@@ -18,11 +18,17 @@ package main
...
 	"context"
+	"encoding/json"
 	"flag"
 	"fmt"
 	"os"
+	"sort"
+	"strings"
 	"time"
 
+	//"strings"
```
stevekuznetsov, line 29: oops

prow/cmd/sinker/main.go
```diff
@@ -532,3 +573,80 @@ func podNeedsKubernetesFinalizerCleanup(log *logrus.Entry, pj *prowapi.ProwJob,
...
+		if err != nil {
+			return "", err
+		}
+		b.WriteString(res)
+	}
+
+	return b.String(), nil
+}
+
+func comparableRefsString(r *prowapi.Refs) (string, error) {
```
stevekuznetsov, line 643: Surely there's something from `cmp` that gives us this comparison without a marshal? Or, can't we get a `DeepEqual` generated for us by client-go stuff since this is part of the CR? Or, barring that, just write an equality func by hand and a fuzzing test for it so it stays up to date? This seems expensive to be putting into such a hot callpath.

prow/cmd/sinker/main.go
alvaroaleman: Yeah I thought a bit about this, but I think we do not care a lot about how long this takes, because in most cases sinker sync takes a lot less time than its resync period. It might be possible to optimize this a bit by handcrafting something to construct a string out of a `ref` but that adds some complexity and I wanted to avoid this unless we are sure the performance is actually an issue.

prow/cmd/sinker/main.go
alvaroaleman: This is a map, maps fundamentally need a key without providing any input, otherwise they would always need to iterate over all elements on fetch and iterate over all possible keys when you insert a new element.

I thought about using a combination of a map and a slice by using  a map of jobnames with a two dimensional job array, first dimension is the unique refs, then iterating and using a comparison func to compare. But this is definitely a lot more complex and might not even be faster.

prow/cmd/sinker/main.go
alvaroaleman: Tbh I didn't think about which to keep at all :) I've added some sorting so youngest comes first so the one we keep will be the younged

prow/cmd/sinker/main.go
stevekuznetsov: Sorry, I'm not following. Why are we choosing to do JSON serialization and string comparison over some dynamic comparison from e.g. `cmp` or a comparison func written for this purpose?

alvaroaleman: We group all prowjobs that have the same (.spec.name, spec.refs, spec.extraRefs, status.State). Using a comparison func would require us to call the comparison func for all groups for all prowjobs, unless we find a match earlier. Since we will have a lot of groups this can be assumed to be more expensive than what we currently do. Additionally, it is more complicated. So for now, this is the simplest solution. If we end up finding out that we need a more efficient approach, we will have to profile alternate approaches.

PR #21458, opened 2021-03-19
[Kettle] Open files in context

MushuEE:
@amwat found we were leaking file descriptors in opening the BQ schema.

Dropped file into context to avoid that

/area kettle


amwat (commented):
/lgtm
/approve

PR #21459, opened 2021-03-19
azure: use lower case for resource group name

chewong:

Fixes failure in https://testgrid.k8s.io/provider-azure-master-signal#aks-engine-azure-file. Looks like the azure file volume name (which has the resource group name as the prefix during the test) only accepts a lower case string.

/assign @nilo19 @feiskyer


feiskyer (commented):
/lgtm
/approve

PR #21460, opened 2021-03-19
Move kubetest2 jobs off of krte

amwat:
/cc @BenTheElder @spiffxp


BenTheElder:
/lgtm
/approve
thank you!

PR #21461, opened 2021-03-19
add back missing FROM to krte

BenTheElder:
missed in https://github.com/kubernetes/test-infra/pull/21443/files#r597994859


amwat (commented):
/lgtm

Issue #21426, opened 2021-03-18
Plugin-help cannot be served when plugins are enabled for organizations with many repos.


cjwagner:
**What happened**:
Generating plugin-help information currently requires listing all repos in organizations that have plugins configured in order to allow complete plugin-help information to be generated for every repo. This is problematic when organizations have huge numbers of repos which may be the case if plugins are configured for such an org, but webhooks are only enabled for some repos in the org. For example, the https://oss-prow.knative.dev instance serves repos in the `GoogleCloudPlatform` and `google` organizations.

**What you expected to happen**:
Plugin help to be served even if plugins are configured for a large organization.

**How to reproduce it (as minimally and precisely as possible)**:
Configure plugins for a large GitHub organization (with or without setting up webhooks) and try to access plugin help via the `/plugins` or `/command-help` Deck paths.

**Please provide links to example occurrences, if any**:
https://oss-prow.knative.dev/plugins

**Anything else we need to know?**:
The original issue is here and has more details: https://github.com/GoogleCloudPlatform/oss-test-infra/issues/362

petr-muller:
/remove-lifecycle stale

We still suffer from this :(

PR #21466, opened 2021-03-21
[Prow] Speedup GetRepos Query

MushuEE:
WIP

This is in efforts to stop #21426

Using the v4 endpoint collecting only `Name` seems to bring queries down > 50% this might not. be sufficient to fix the issue but will help in the process. 

@cjwagner  I am still trying to learn Go and am not sure if this is the best way of forking this functionality. Seems more verbose than required

/area prow
/hold WIP

### DIFF
--- a/prow/github/client.go
+++ b/prow/github/client.go
@@ -2315,6 +2315,104 @@ func (c *client) GetRepos(org string, isUser bool) ([]Repo, error) {
 	return repos, nil
 }
 
+// Repoer is an interface for GraphQL repo queries
+type Repoer interface {
+	Repos() Repositories
+}
+
+// Repositories represents a sub element of the GraphQL query for repos
+type Repositories struct {
+	Edges []struct {
+		Node RepoName
+	}
+	PageInfo struct {
+		EndCursor   githubv4.String
+		HasNextPage githubv4.Boolean
+	}
+}
+
+// OrgRepo represnets a query for repos under an organization
+type OrgRepo struct {
+	Organization struct {
+		Repositories `graphql:"repositories(first: $first, after: $cursor)"`
+	} `graphql:"organization(login: $login)"`
+	RateLimit Rates
+}
+
+// UserRepo represnets a query for repos under a user
+type UserRepo struct {
+	User struct {
+		Repositories `graphql:"repositories(first: $first, after: $cursor)"`
+	} `graphql:"user(login: $login)"`
+	RateLimit Rates
+}
+
+// Repositories collects the subelement from a OrgRepo query
+func (o *OrgRepo) Repositories() Repositories {
+	return o.Organization.Repositories
+}
+
+// Repositories collects the subelement from a UserRepo query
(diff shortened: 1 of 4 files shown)


prow/github/client.go
```diff
@@ -2315,6 +2315,112 @@ func (c *client) GetRepos(org string, isUser bool) ([]Repo, error) {
 	return repos, nil
 }
 
+// GetRepoNames returns names of all repos in an org.
+//
+// This call uses multiple API tokens when results are paginated.
+//
+// See https://docs.github.com/en/graphql/reference/objects#repository
+func (c *client) GetRepoNames(org string, isUser bool) ([]Repo, error) {
```
petr-muller, line 2323: Maybe this would be the good opportunity to get rid of the `isUser` modal and simply provide `GetRepoNamesForOrg` and `GetRepoNamesForUser` to stress out the fact that these calls are different in GH API.

MushuEE: Updated to do this, I will need to hunt down where all these are called and replace them. Before had wanted to make sure everyone was ok with the structure of the code. I am getting the feel for Golang still and was not sure if I should put some of these in `types.go`

prow/github/client.go
```diff
@@ -2315,6 +2315,104 @@ func (c *client) GetRepos(org string, isUser bool) ([]Repo, error) {
...
+func (c *client) getPaginatedRepoNames(org string, r Repoer) ([]RepoName, error) {
+	repoNames := []RepoName
+
+	if c.fake {
+		return repoNames, nil
+	}
+
+	variables := map[string]interface{}{
+		"login":  graphql.String(org),
+		"first":  graphql.Int(100),
```
cjwagner, line 2398: Can we get more than 100?

prow/github/types.go
```diff
@@ -346,6 +346,13 @@ type Repo struct {
 	Parent      ParentRepo      `json:"parent"`
 }
 
+// RepoName is a limited use version of Repo that holds only `name`
+// it is used in the v4 query to collect only names, this is faster than v3
+// See also https://docs.github.com/en/graphql/reference/objects#repository
+type RepoName struct {
```
cjwagner, line 352: nit: Since this type is specific to the `GetRepoNamesFor*` functions I think it makes sense to colocate it with those functions and the rest of the related types. Unfortunately the types used for graphql queries are not typically as reusable as the REST API types.

prow/github/client.go
```diff
@@ -2315,6 +2315,104 @@ func (c *client) GetRepos(org string, isUser bool) ([]Repo, error) {
...
+
+	return repoNames, nil
+}
+
+// GetRepoNamesForUser returns names of all repos under a user.
+//
+// This call uses multiple API tokens when results are paginated.
+//
+// See https://docs.github.com/en/graphql/reference/objects#repository
+func (c *client) GetRepoNamesForUser(org string) ([]Repo, error) {
```
cjwagner, line 2379: It would be more convenient for these to return `([]string, error)`.

PR #21468, opened 2021-03-22
kubernetes-csi: more jobs for distributed provisioning and storage capacity

pohly:
The periodic job runs in the hostpath driver repo, like the others,
but uses the "distributed provisioning" YAML files for deployment.

The optional pre-merge job for external-provisioner is important
because otherwise we have no coverage of the distributed provisioning
and storage capacity tracking code before merging a PR.

### DIFF
--- a/config/jobs/kubernetes-csi/external-provisioner/external-provisioner-manual-job-config.yaml
+++ b/config/jobs/kubernetes-csi/external-provisioner/external-provisioner-manual-job-config.yaml
@@ -1,0 +1,56 @@
+# Derived from external-provisioner-config.yaml, modified and maintained manually.
+
+presubmits:
+  kubernetes-csi/external-provisioner:
+  - name: pull-kubernetes-csi-external-provisioner-distributed-on-kubernetes-master
+    # Explicitly needs to be started with /test.
+    # This cannot be enabled by default because there's always the risk
+    # that something changes in master which breaks the pre-merge check.
+    always_run: false
+    optional: true
+    decorate: true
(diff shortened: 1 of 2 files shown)


config/jobs/kubernetes-csi/external-provisioner/external-provisioner-manual-job-config.yaml
```diff
@@ -0,0 +1,56 @@
...
+        - name: CSI_PROW_DEPLOYMENT
+          value: "kubernetes-distributed"
+        # docker-in-docker needs privileged mode
+        securityContext:
+          privileged: true
+      resources:
+        requests:
+          # these are both a bit below peak usage during build
+          # this is mostly for building kubernetes
+          memory: "9000Mi"
```
xing-yang, line 54: I see this memory setting is removed from other places, but kept here.  Can you explain why?

pohly: The gen script for the other jobs uses a heuristic to determine whether Kubernetes must be built from source: the unstable Kubernetes is assumed to be built from source, everything else is assumed to be supported by KinD images. That heuristic is currently wrong, 1.20 still gets built from source. I'm changing that in https://github.com/kubernetes-csi/csi-release-tools/pull/138 so it didn't seem worthwhile to override the heuristic.

Here I kept the higher memory requirement because these jobs will be updated soonish anyway - I want to move them into the automatically generated jobs.

PR #21469, opened 2021-03-22
kubernetes-csi: promote testing on 1.20 to mandatory

pohly:
The various components should now pass testing on 1.20.

### DIFF
--- a/config/jobs/kubernetes-csi/csi-driver-host-path/csi-driver-host-path-config.yaml
+++ b/config/jobs/kubernetes-csi/csi-driver-host-path/csi-driver-host-path-config.yaml
@@ -188,7 +188,7 @@ presubmits:
           cpu: 2000m
   - name: pull-kubernetes-csi-csi-driver-host-path-1-20-on-kubernetes-1-20
     always_run: true
-    optional: true
+    optional: false
     decorate: true
     skip_report: false
     skip_branches: []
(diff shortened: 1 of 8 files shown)


config/jobs/kubernetes-csi/csi-driver-host-path/csi-driver-host-path-config.yaml
```diff
@@ -188,7 +188,7 @@ presubmits:
           cpu: 2000m
   - name: pull-kubernetes-csi-csi-driver-host-path-1-20-on-kubernetes-1-20
     always_run: true
-    optional: true
+    optional: false
```
xing-yang, line 191: Why is this changed to "false" by default now?

config/jobs/kubernetes-csi/csi-driver-host-path/csi-driver-host-path-config.yaml
```diff
@@ -229,13 +229,9 @@ presubmits:
         # docker-in-docker needs privileged mode
         securityContext:
           privileged: true
-      resources:
-        requests:
-          # these are both a bit below peak usage during build
-          # this is mostly for building kubernetes
-          memory: "9000Mi"
```
xing-yang, line 236: Can you explain why this is removed?

config/jobs/kubernetes-csi/csi-driver-host-path/csi-driver-host-path-config.yaml
pohly: Because we no longer want to tolerate test failures on 1.20.

config/jobs/kubernetes-csi/csi-driver-host-path/csi-driver-host-path-config.yaml
pohly: See https://github.com/kubernetes/test-infra/pull/21468#discussion_r598804630

PR #21482, opened 2021-03-23
bump pull-kind-e2e timeout

aojea:
The job is timing out, the last 3 success were achieved in 39m54s, 38m12s and 38m42s

Bumping the timeout to 60 mins, same that the k/k presubmit jobs

### DIFF
--- a/config/jobs/kubernetes-sigs/kind/kind-presubmits.yaml
+++ b/config/jobs/kubernetes-sigs/kind/kind-presubmits.yaml
@@ -62,7 +62,7 @@ presubmits:
       path_alias: k8s.io/kubernetes
     path_alias: sigs.k8s.io/kind
     decoration_config:
-      timeout: 40m
+      timeout: 60m
       grace_period: 15m
     spec:
       containers:


config/jobs/kubernetes-sigs/kind/kind-presubmits.yaml
```diff
@@ -62,7 +62,7 @@ presubmits:
       path_alias: k8s.io/kubernetes
     path_alias: sigs.k8s.io/kind
     decoration_config:
-      timeout: 40m
+      timeout: 50m
```
BenTheElder, line 65: We should bump k/k as well. This job is meant to give us the same experience as k/k presubmit.

aojea: hahha you are 2 years ahead of the rest
https://github.com/kubernetes/test-infra/commit/ad9017ab1cb85d122e782781ab692e35610dd3cd

aojea: let's have the same timeout in all of them

BenTheElder (approved):

PR #21479, opened 2021-03-22
azure: add storage test on CAPZ to 1.18, 1.19, 1.20 and master signal dashboard

chewong:
Signed-off-by: Ernest Wong <chuwon@microsoft.com>

Follow-up PR for https://github.com/kubernetes/test-infra/pull/21323. 
- Adding the following storage test on CAPZ clusters for 1.18, 1.19, 1.20, and master:
  - capz-azure-file
  - capz-azure-file-machinepool
  - capz-azure-disk
  - capz-azure-disk-machinepool
- Using [krte](https://github.com/kubernetes/test-infra/tree/master/images/krte) instead of kubekins-e2e for as the test image due to slightly small image footprint

/assign @CecileRobertMichon

PR #21483, opened 2021-03-23
azure: update capz test image and aks-engine build

chewong:

Follow-up PR for https://github.com/kubernetes/test-infra/pull/21479.

- jobs were incorrectly checking out `release-master` branch instead of `master` for Kubernetes
- `krte` does not have `ssh-keygen`. Let's switch back to `kubekins-e2e` for now
- Use aks-engine nightly build for jobs that are using aks-engine

/assign @CecileRobertMichon

### DIFF
--- a/config/jobs/kubernetes/sig-cloud-provider/azure/release-1.19.yaml
+++ b/config/jobs/kubernetes/sig-cloud-provider/azure/release-1.19.yaml
@@ -39,7 +39,7 @@ presubmits:
         - --aksengine-location=westus2
         - --aksengine-public-key=$(AZURE_SSH_PUBLIC_KEY_FILE)
         - --aksengine-template-url=https://raw.githubusercontent.com/kubernetes-sigs/cloud-provider-azure/master/tests/k8s-azure/manifest/kubernetes.json
-        - --aksengine-download-url=https://aka.ms/aks-engine/aks-engine-k8s-e2e.tar.gz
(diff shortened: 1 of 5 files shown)


config/jobs/kubernetes/sig-cloud-provider/azure/release-1.19.yaml
```diff
@@ -437,9 +438,9 @@ periodics:
     path_alias: k8s.io/kubernetes
   spec:
     containers:
-    - image: gcr.io/k8s-testimages/krte:v20210319-e46e31c-master
+    - image: gcr.io/k8s-testimages/kubekins-e2e:v20210312-67f589a-master
```
CecileRobertMichon, line 441: I think you need go-canary to get go 1.16

CecileRobertMichon: https://github.com/kubernetes/test-infra/blob/5209ee96af140395dbc80f46e7f78f2782862051/config/jobs/kubernetes-sigs/cluster-api-provider-azure/cluster-api-provider-azure-presubmits.yaml#L13

chewong: `gcr.io/k8s-testimages/kubekins-e2e:v20210312-67f589a-master` also has Go 1.16.1. The reason why it didn't work in the past is that we use the kubekins-e2e based on the Kubernetes release, which has a different Go version from master

PR #21485, opened 2021-03-23
Use 2004 Image Repo List for GCE's 20H2 test runs.

jeremyje:
This change effectively moves the private image repo from gcr.io to e2eprivate which supports 20H2.


jeremyje:
https://github.com/kubernetes-sigs/windows-testing/pull/259 is the other option.

jeremyje:
/hold
/sig windows

Holding because looks like https://github.com/kubernetes-sigs/windows-testing/pull/259 will be accepted instead.

PR #21486, opened 2021-03-23
prow: pod-utils: sidecar: allow configuring censor paths

stevekuznetsov:
It is now possible to configure the paths that will and will not have
their contents censored with globs. In order to fail into an overly
cautious configuration, if a path matches both include and exclude
directives, we will censor it. If no include directives are given, the
default behavior is to include everything.

/assign @BenTheElder @cjwagner @alvaroaleman


prow/sidecar/censor.go
```diff
@@ -93,6 +94,14 @@ func (o Options) censor() error {
 				return nil
 			}
 			logger := logrus.WithField("path", absPath)
+			relpath, _ := filepath.Rel(item, absPath) // err happens when there's no rel path, but we know there must be
```
cjwagner, line 97: nit: If we're somehow wrong it is better to get a "Programmer error!" log rather than silence.

cjwagner (commented):
> In order to fail into an overly cautious configuration, if a path matches both include and exclude directives, we will censor it.

I think the only case in which one would specify both includes and excludes would be to exclude paths that match the includes directive. Paths that don't match the include directive would be automatically excluded so the exclude directive wouldn't be needed.
So I think either the opposite behavior should be used or specifying both includes and excludes should be forbidden since there would be no use for it.


Also it doesn't look like this can actually be configured from the job config, these options are never provided by the decoration logic.

stevekuznetsov:
@cjwagner thanks for looking - I think I handled all of your comments, PTAL :)

stevekuznetsov:
The number of things a dev has to do when updating an api/config type is kind of annoying and not consistent at all:

```
UPDATE=true go test ./prow/config/... -run TestGenYamlDocs
hack/update-codegen.sh
```

prow/apis/prowjobs/v1/types.go
```diff
@@ -368,9 +368,72 @@ type DecorationConfig struct {
...
+	// IncludeDirectories are directories which should have their content censored. If
+	// present, only content in these directories will be censored. Entries in this list
+	// are parsed with the go-zglob library, allowing for globbed matches.
+	IncludeDirectories []string `json:"include_directories,omitempty"`
+
+	// ExcludeDirectories are directories which should not have their content censored. If
+	// present, content in these directories will not be censored even if the directory also
+	// matches a glob in IncludeDirectories. Entries in this list are parsed with the go-
+	// zglob library, allowing for globbed matches.
+	ExcludeDirectories []string `json:"exclude_directories,omitempty"`
```
spiffxp, line 402: Coming at this with very little context... as in, if I'm just looking at these options, I don't understand this feature:
- what path are these directories relative to?
- what's getting censored when/where again? this is artifacts and logs right? happening concurrently as the test is running? or before artifacts are uploaded?
- is it per container? per job?
- is this on the node running the job? is this only supported by `agent: kubernetes` jobs?

stevekuznetsov: > what path are these directories relative to?

The only path that users ever do anything relative to, the `$ARTIFACTS` path - but I can clarify that in the GoDoc. I honestly don't think it's worth trying to include or exclude files from this, I imagine the compute/time taken to load & censor even large build files is tiny and the mental overhead of figuring out what might and might not contain sensitive data and to what the paths are relative to, etc, outweigh the costs ... but the feature was requested, so here is is :)

> what's getting censored when/where again?

This is a two-fold answer. First, I imagine it's best if this were opaque to the user. Then, they can just use the thing and forget about it. Second, it depends on how well the system can auto-detect this for people. Right now, the impl will censor any data present in k8s secrets mounted to the containers in the Pod that execute user code.

> this is artifacts and logs right? happening concurrently as the test is running? or before artifacts are uploaded?

This is an option of `sidecar`, which handles artifacts and logs pushed to GCS, so I think in context you kind of know the answers to all of these if you're actually configuring it. Censoring happens before it's uploaded, but from a user's perspective a Prow Job has no artifacts until they are uploaded so that question is kind of moot.

> is it per container? per job?

All Prow configuration is per job. No ability to configure options per container exists for any part of the system. IIRC only one `sidecar` runs even when more than one container exists in the job anyway, so there's no real clear answer to this question. Again, I'd wager the user does not care about this distinction.

> is this on the node running the job?

This is part of Prow k8s job decoration, so yes like with any other part of the larger system it runs on Kubernetes jobs only.

stevekuznetsov:
Added a documentation header in the pod utilities doc.

stevekuznetsov:
@cjwagner did this need anything else?

prow/apis/prowjobs/v1/types.go
cjwagner: Please clarify godocs to indicate that paths are relative to `$ARTIFACTS`.

cjwagner (commented):
1 comment nit, otherwise LGTM.

cjwagner (approved):

PR #21498, opened 2021-03-24
Enable API availability measurement in 100 and 5k test

tosi3k:
Let's enable this in order to collect some data which would support us in choosing a valid API availability SLO.

/sig scalability
/assign @wojtek-t


wojtek-t:
/lgtm
/approve

/hold
Let's merge on Monday.

PR #21502, opened 2021-03-24
chore: switch to aks-e nightly build for dual-stack jobs

aramase:

- Switches to using nightly aks-engine build from `https://github.com/Azure/aks-engine/releases/download/nightly/aks-engine-nightly-linux-amd64.tar.gz` as the fix for OIDC issuer is not yet available in the default release.

```bash
2021/03/24 11:20:30 OK: Got token
2021/03/24 11:20:30 OK: got issuer kubernetes.default.svc
2021/03/24 11:20:30 Full, not-validated claims: 
openidmetadata.claims{Claims:jwt.Claims{Issuer:"kubernetes.default.svc", Subject:"system:serviceaccount:svcaccounts-283:default", Audience:jwt.Audience{"oidc-discovery-test"}, Expiry:1616585428, NotBefore:1616584828, IssuedAt:1616584828, ID:""}, Kubernetes:openidmetadata.kubeClaims{Namespace:"svcaccounts-283", ServiceAccount:openidmetadata.kubeName{Name:"default", UID:"38eca534-bf8e-4425-85ae-052ffa53265a"}}}
2021/03/24 11:20:30 Get "kubernetes.default.svc/.well-known/openid-configuration": unsupported protocol scheme ""
```

After there is an aks-engine release with the fix, we'll switch back to using a fixed version

/assign @aojea

### DIFF
--- a/config/jobs/kubernetes/sig-network/dualstack-e2e.yaml
+++ b/config/jobs/kubernetes/sig-network/dualstack-e2e.yaml
@@ -141,7 +141,7 @@ periodics:
       - --aksengine-location=westus2
       - --aksengine-public-key=$(AZURE_SSH_PUBLIC_KEY_FILE)
       - --aksengine-template-url=https://raw.githubusercontent.com/Azure/aks-engine/master/examples/dualstack/kubernetes.json
-      - --aksengine-download-url=https://aka.ms/aks-engine/aks-engine-k8s-e2e.tar.gz
+      - --aksengine-download-url=https://github.com/Azure/aks-engine/releases/download/nightly/aks-engine-nightly-linux-amd64.tar.gz
       # Specific test args
       # Skipping "HostPort validates that there is no conflict between pods with same hostPort but different hostIP and protocol" because of kubenet bug in HostPort impl
       - --test_args=--ginkgo.focus=\[Feature:IPv6DualStackAlphaFeature\]|\[Conformance\] --ginkgo.skip=\[Slow\]|\[Serial\]|\[Flaky\]|Should.recreate.evicted.statefulset|HostPort.validates.that.there.is.no.conflict.between.pods.with.same.hostPort.but.different.hostIP.and.protocol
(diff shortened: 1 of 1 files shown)


aramase:
Opened https://github.com/kubernetes-sigs/cloud-provider-azure/pull/571 to add to the job manifests as well

config/jobs/kubernetes/sig-network/dualstack-e2e.yaml
```diff
@@ -141,7 +141,7 @@ periodics:
       - --aksengine-location=westus2
       - --aksengine-public-key=$(AZURE_SSH_PUBLIC_KEY_FILE)
       - --aksengine-template-url=https://raw.githubusercontent.com/Azure/aks-engine/master/examples/dualstack/kubernetes.json
-      - --aksengine-download-url=https://aka.ms/aks-engine/aks-engine-k8s-e2e.tar.gz
+      - --aksengine-download-url=https://github.com/Azure/aks-engine/releases/download/nightly/aks-engine-nightly-linux-amd64.tar.gz
```
aojea, line 144: can we pin to a specific nightly? or is this ok?
I'm afraid we break for some unrelated change

aramase: @aojea I share the same concern. The locked version unfortunately doesn't have the fix yet. Once https://github.com/kubernetes-sigs/cloud-provider-azure/pull/571 merges later today we can change back to using the old version. I want to unblock today's test run until the other PR is merged. Or we can just wait for the other one to merge and not do this. WDYT?

aojea: I think that is better to wait then, so we can traces the changes, if we do several changes at the same time, and after that the job fails,  we'll not be able to reconstruct what happened

aramase: sgtm! I'll close this PR and we'll just get the other one merged.

PR #21503, opened 2021-03-24
Update test package version for cloud-provider-gcp conformance test

cici37:
Since cloud-provider-gcp is updated with v1.20, update the related test package version as well.


cici37:
Hi @mikedanese Would you mind to take a look when you have time? Thank you🙏

Issue #21507, opened 2021-03-24
BranchProtection merging test is flaky.


cjwagner:
**What happened**:
`TestBranchProtectionMergeMergesAnythingWithoutAnError` flaked. The fuzzing seems to have detected a bug: https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/test-infra/21505/pull-test-infra-bazel/1374784788686180352
The expected result diff is pretty tough to read due to the fuzzing.

**What you expected to happen**:
No flaking unit tests and for the underlying bug to be resolved if there is one.

**Anything else we need to know?**:
The tests were recently introduced in this PR: https://github.com/kubernetes/test-infra/pull/21405

/assign @alvaroaleman 
/area prow/branchprotector

PR #21510, opened 2021-03-24
Persist job runs past max_prowjob_age by storing them in the GH status description

alvaroaleman:
Currently, Tide always has to retest if a test result is older than
max_prowjob_age because it can not determine anymore what baserevision
the test run ran against.

While it is possible to optimize Sinker to be able to keep presubmits
longer, the bigger Prow instances like the Kubernetes or Openshift one
will never be able to keep ProwJobs for more than a couple of days.
This is not sufficient for for example backports, where PRs regularly
wait for a week or more until they can be merged.

This PR makes us (ab-)use the GitHub status context description to store
the baseSHA and makes Tide's accumulate construct virtual ProwJob
objects out of these. This allows us to store past presubmit results
pretty much indefinitely.

Apart from the included unittests I also manually deployed this and verified that Tide merges without creating a new Prowjob when the baseSHA is current and continues to create the ProwJob if not.

The status now looks like this on GitHub:
gh_basesha

This approach was discussed in Slack: https://kubernetes.slack.com/archives/CDECRSC5U/p1616509899019900
Supersedes and thereby closes https://github.com/kubernetes/test-infra/pull/21453

I recommend disabling whitespace changes to reduce the diff. Most of the change in this PR is tests, because the PR object tide uses has very deeply nested substructs.

/assign @chaodaiG @cjwagner @stevekuznetsov

### DIFF
--- a/prow/config/config.go
+++ b/prow/config/config.go
@@ -2536,3 +2536,42 @@ func (pc *ProwConfig) mergeFrom(additional *ProwConfig) error {
 	}
 	return pc.BranchProtection.merge(&additional.BranchProtection)
 }
+
+// ContextDescriptionWithBaseSha is used by the GitHub reporting to store the baseSHA of a context
+// in the status context description. Tide will read this if present using the BaseSHAFromContextDescription
+// func. Storing the baseSHA in the status context allows us to store job results pretty much forever,
+// instead of having to rerun everything after sinker cleaned up the ProwJobs.
+func ContextDescriptionWithBaseSha(humanReadable, baseSHA string) string {
+	var suffix string
+	if baseSHA != "" {
+		suffix = contextDescriptionBaseSHADelimiter + baseSHA
+	}
+	return truncate(humanReadable, contextDescriptionMaxLen-len(suffix)) + suffix
+}
+
+// BaseSHAFromContextDescription is used by Tide to decode a baseSHA from a github status context
+// description created via ContextDescriptionWithBaseSha. It will return an empty string if no
+// valid sha was found.
+func BaseSHAFromContextDescription(description string) string {
+	split := strings.Split(description, contextDescriptionBaseSHADelimiter)
+	// SHA1s are always 40 digits long
+	if len(split) != 2 || len(split[1]) != 40 {
+		return ""
+	}
+	return split[1]
+}
+
+const (
+	contextDescriptionBaseSHADelimiter = " Basesha:"
+	contextDescriptionMaxLen           = 140 // https://developer.github.com/v3/repos/deployments/#parameters-2
+	elide                              = " ... "
+)
+
+// truncate converts "really long messages" into "really ... messages".
+func truncate(in string, maxLen int) string {
+	half := (maxLen - len(elide)) / 2
+	if len(in) <= maxLen {
+		return in
+	}
+	return in[:half] + elide + in[len(in)-half:]
+}
--- a/prow/tide/tide.go
+++ b/prow/tide/tide.go
@@ -932,12 +932,54 @@ func (c *Controller) accumulateBatch(sp subpool) (successBatch []PullRequest, pe
 	return successBatch, pendingBatch
 }
 
+// prowJobsFromContexts constructs ProwJob objects from all successful presubmit contexts that include a baseSHA.
+// This is needed because otherwise we would always need retesting for results that are older than sinkers
+// max_prowjob_age.
+func prowJobsFromContexts(l *logrus.Entry, ghc githubClient, pr *PullRequest, baseSHA string) ([]prowapi.ProwJob, error) {
+	headContexts, err := headContexts(l, ghc, pr)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get head contexts: %w", err)
+	}
+	var passingCurrentContexts []string
+	for _, headContext := range headContexts {
+		if headContext.State != githubql.StatusStateSuccess {
+			continue
+		}
+		if baseSHAForContext := config.BaseSHAFromContextDescription(string(headContext.Description)); baseSHAForContext != "" && baseSHAForContext == baseSHA {
+			passingCurrentContexts = append(passingCurrentContexts, string((headContext.Context)))
+		}
+	}
+
+	var prowjobsFromContexts []prowapi.ProwJob
+	for _, passingCurrentContext := range passingCurrentContexts {
+		prowjobsFromContexts = append(prowjobsFromContexts, prowapi.ProwJob{
+			Spec: prowapi.ProwJobSpec{
+				Context: passingCurrentContext,
+				Refs:    &prowapi.Refs{Pulls: []prowapi.Pull{{Number: int(pr.Number), SHA: string(pr.HeadRefOID)}}},
+				Type:    prowapi.PresubmitJob,
+			},
+			Status: prowapi.ProwJobStatus{
+				State: prowapi.SuccessState,
+			},
+		})
+	}
+
+	return prowjobsFromContexts, nil
+}
+
 // accumulate returns the supplied PRs sorted into three buckets based on their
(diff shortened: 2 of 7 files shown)


prow/tide/tide.go
```diff
@@ -932,12 +932,54 @@ func (c *Controller) accumulateBatch(sp subpool) (successBatch []PullRequest, pe
...
+			passingCurrentContexts = append(passingCurrentContexts, string((headContext.Context)))
+		}
+	}
+
+	var prowjobsFromContexts []prowapi.ProwJob
+	for _, passingCurrentContext := range passingCurrentContexts {
+		prowjobsFromContexts = append(prowjobsFromContexts, prowapi.ProwJob{
+			Spec: prowapi.ProwJobSpec{
+				Context: passingCurrentContext,
+				Refs:    &prowapi.Refs{Pulls: []prowapi.Pull{{Number: int(pr.Number), SHA: string(pr.HeadRefOID)}}},
```
stevekuznetsov, line 958: This is a bit confusing - instead of only creating ProwJobs that pass against the current HEAD (I assume we append this to the filtered list?) why don't we just always create the synthetic ProwJob and have it specify the base ref it tested against, then have the filtering etc happen as normal after?

alvaroaleman: Because the list this gets appended to is already filtered by baseSHA

prow/config/config.go
```diff
@@ -2536,3 +2536,42 @@ func (pc *ProwConfig) mergeFrom(additional *ProwConfig) error {
...
+	return truncate(humanReadable, contextDescriptionMaxLen-len(suffix)) + suffix
+}
+
+// BaseSHAFromContextDescription is used by Tide to decode a baseSHA from a github status context
+// description created via ContextDescriptionWithBaseSha. It will return an empty string if no
+// valid sha was found.
+func BaseSHAFromContextDescription(description string) string {
+	split := strings.Split(description, contextDescriptionBaseSHADelimiter)
+	// SHA1s are always 40 digits long
+	if len(split) != 2 || len(split[1]) != 40 {
```
chaodaiG, line 2558: This assumes `Basesha:<SHA>` always the last element of context description, is there test guarding this?

prow/tide/tide.go
```diff
@@ -932,12 +932,54 @@ func (c *Controller) accumulateBatch(sp subpool) (successBatch []PullRequest, pe
...
 // accumulate returns the supplied PRs sorted into three buckets based on their
 // accumulated state across the presubmits.
-func accumulate(presubmits map[int][]config.Presubmit, prs []PullRequest, pjs []prowapi.ProwJob, log *logrus.Entry) (successes, pendings, missings []PullRequest, missingTests map[int][]config.Presubmit) {
+func accumulate(presubmits map[int][]config.Presubmit, prs []PullRequest, pjs []prowapi.ProwJob, log *logrus.Entry, baseSHA string, ghc githubClient) (successes, pendings, missings []PullRequest, missingTests map[int][]config.Presubmit) {
 
 	missingTests = map[int][]config.Presubmit{}
 	for _, pr := range prs {
+
+		if prowjobsFromContexts, err := prowJobsFromContexts(log, ghc, &pr, baseSHA); err != nil {
+			logrus.WithError(err).Error("failed to get prowjobs from contexts")
```
chaodaiG, line 978: can we use `log` instead of `logrus` for more info?

prow/tide/tide.go
```diff
@@ -932,12 +932,54 @@ func (c *Controller) accumulateBatch(sp subpool) (successBatch []PullRequest, pe
...
+		} else {
+			pjs = append(pjs, prowjobsFromContexts...)
```
chaodaiG, line 980: I think appending means the priority of status context is higher than prowjobs, so if status context says pass and prowjobs says fail, tide will still merge the PR. Is there any potential risk here? For example outdated status context?

prow/config/config.go
alvaroaleman: There is `TestContextDescriptionWithBaseShaRoundTripping` and we will only return it if the second element has the correct length (40 characters). The worst case for this is that this functionality breaks but not that we incorrectly merge something, because Tide checks if the basesha is correct

prow/tide/tide.go
alvaroaleman: good catch, fixed

prow/tide/tide.go
alvaroaleman: The status context are retrieved for a sha, so they can not be outdated (except for the basesha which is why we are doing the dance here). Tide always tries to find the best result and will use that (i.E. if you have one failure, one success, the job is considered a success) so the order shouldn't be relevant: https://github.com/kubernetes/test-infra/blob/eeb7afb52810b459ddc23b33890e69cbf8efb61f/prow/tide/tide.go#L983

alvaroaleman:
Urg #21507
/retest

prow/tide/tide.go
chaodaiG: I see, this makes sense. Since they are the same sha and baseSha, the order shouldn't matter. Even in worst case, say a flaky job passed then failed, tide will still respect the passed context if the failed job hasn't been reported by crier yet, which is totally fine with me, and I would even consider it a bonus. So SGTM

chaodaiG (approved):
/hold

prow/tide/tide.go
stevekuznetsov: Hm - I guess this entire change would be less dangerous etc if we always created fake ProwJobs with whatever baseSHA and status we see in GitHub, then we'd only have one filtering code for the list and not potentially introduce two areas that, if bugged, have large impacts on tide.

prow/tide/tide.go
stevekuznetsov: Yeah see like *this* conversation also does not have to happen if we generate all contexts with correct timestamps etc and do filtering after.

stevekuznetsov (approved):
The code as written looks correct. While I still think there's a more defensive way to code this, if we're happy with this impl please feel free to cancel the hold and merge.

/hold

prow/tide/tide.go
alvaroaleman: Updated it to create ProwJobs instead. It will still only create prowjobs for the matching baseSHA and successes to not overload the apiserver, but it will take the created baseSHA from the github context.

cjwagner: Why do the ProwJobs need to be added to the API server? That seems like unnecessary load on the API server given that Tide is the only thing that will consume these. There are also other catches like avoiding creating the same PJs every sync loop and ensuring other Prow components handle the fake PJs appropriately (e.g. Deck should ignore them, sinker would need to handle them differently).
I interpreted Steve's request to mean create ProwJob structs so that we can filter them the same way we filter real PJs, not create real ProwJob CRs on the API server.

alvaroaleman: > Why do the ProwJobs need to be added to the API server?

Because the filtering is done via [an index over the cache](https://github.com/kubernetes/test-infra/blob/eeb7afb52810b459ddc23b33890e69cbf8efb61f/prow/tide/tide.go#L1716) in order to be efficient and I don't think trying to trick the cache into thinking there was data that isn't on the server is a great idea.

> There are also other catches like avoiding creating the same PJs every sync loop and ensuring other Prow components handle the fake PJs appropriately 

True, I've reverted the pull to the original version

> I interpreted Steve's request to mean create ProwJob structs so that we can filter them the same way we filter real PJs, not create real ProwJob CRs on the API server.

It already did create prowjob structs in the first version and as mentioned above, re-using the filtering isn't really possible.

fejta:
LGTM
/uncc
Will defer to Cole (and Chao) for this PR

stevekuznetsov:
/lgtm
/hold cancel

The last commit reverted back to the changes that had been LGTMed in the past, this seems good to go.

PR #21513, opened 2021-03-24
Phaino: update instruction with advanced usages

chaodaiG:


prow/cmd/phaino/README.md
```diff
@@ -41,6 +41,37 @@ volumes that the Prow Job may require.
 * `--use-local-gcloud-credentials` controls whether to use the same gcloud credentials as local or not
 * `--use-local-kubeconfig` controls whether to use the same kubeconfig as local or not
 
+#### Common options usage scenarios
+
+Phaino is smart at prompting for where repo is located, volume mounts etc., if
+it's desired to save the prompts, use the following tricks instead:
+
+- If the repo is not cloned under GOPATH, use:
```
chizhg, line 49: This should be changed to `If the repo needs to be cloned under GOPATH`.

The usage of the `--gopath` flag here is a bit tricky, since when handling the `extra_refs`, we need the repos to be always under the `GOPATH` and skip the repos that are already mounted via `--extra-volume-mounts`, `--gopath` is needed to make sure the logic is correct.

prow/cmd/phaino/README.md
```diff
@@ -41,6 +41,37 @@ volumes that the Prow Job may require.
...
+  ```
+  --use-local-kubeconfig
+  --skip-volume-mounts=kubeconfig
+  ```
+- If job requires mounting gcloud default credentials, assume the mount is named `service-account`,use:
+  ```
+  --use-local-gcloud-credentials
+  --skip-volume-mounts=service-account
+  ```
+- If job requires mounting something else like `name:foo; mountPath: /bar`,use:
```
chizhg, line 64: Hmm is this example correct?

chizhg (commented):
Thanks Chao! Left some comments.

prow/cmd/phaino/README.md
chaodaiG: yes I believe so, it's not very intuitive to me tbh

prow/cmd/phaino/README.md
chaodaiG: that's right, done

PR #21524, opened 2021-03-25
Make Cluster API make-main target optional

vincepri:
After moving golangci-lint to github actions, which have an internal
cache and can run in parallel, we can move this target to optional and
then removing it a bit later.


CecileRobertMichon:
/lgtm
/hold

until https://github.com/kubernetes-sigs/cluster-api/pull/4374 merges

PR #21525, opened 2021-03-25
Deflake TestBranchProtectionMergeMergesAnythingWithoutAnError

alvaroaleman:
The merging logic used to only check if a policy was set for a Branch
and if not, do nothing. This made the test fail when an empty policy was
set, because a map with a key and an empty value differs from a map with
no key.

I also have to say I am disappointed at the fuzzing here, I didn't manage to reproduce it without the explicit fuzzing funcs even when running 10k rounds of the test (maybe the seed which is `time.Now().UnixNano()` by default has too little entropy?).

Fixes https://github.com/kubernetes/test-infra/issues/21507
/cc @stevekuznetsov @chaodaiG @stevekuznetsov

### DIFF
--- a/prow/config/branch_protection.go
+++ b/prow/config/branch_protection.go
@@ -221,7 +221,7 @@ func (bp *BranchProtection) merge(additional *BranchProtection) error {
 		}
 		if isPolicySet(bp.Orgs[org].Policy) && isPolicySet(additional.Orgs[org].Policy) {
 			errs = append(errs, fmt.Errorf("both branchprotection configs define a policy for org %s", org))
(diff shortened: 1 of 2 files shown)


stevekuznetsov (approved):

chaodaiG:
Thank you for fixing this! Have a little idea on your question

> I also have to say I am disappointed at the fuzzing here, I didn't manage to reproduce it without the explicit fuzzing funcs even when running 10k rounds of the test (maybe the seed which is `time.Now().UnixNano()` by default has too little entropy?).

Would it be possible to make the test runnable with a given seed? If that's possible we can let the test print out the random seed used in the failed test, and use the same seed to repro the test.

PR #21528, opened 2021-03-25
Improve fuzztest by printing seed and re-using fuzzer

alvaroaleman:
Printing the seed allows us to reproduce issues we see in CI, which
should make understanding them easier. Suggested by Chao in
https://github.com/kubernetes/test-infra/pull/21525#issuecomment-807109286.

Re-using the fuzzer should give us more entropy, because it internally
uses a PRNG that gets initialized with a seed. The PRNG should yield
more entropy when we continue using it, rather than re-initializing it
all the time with a seed that is only slightly different.

/assign @chaodaiG

### DIFF
--- a/prow/apis/prowjobs/v1/types_test.go
+++ b/prow/apis/prowjobs/v1/types_test.go
@@ -245,10 +245,14 @@ func TestDecorationDefaultingDoesntOverwrite(t *testing.T) {
 
 func TestApplyDefaultsAppliesDefaultsForAllFields(t *testing.T) {
 	t.Parallel()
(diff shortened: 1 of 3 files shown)


chaodaiG (approved):
Awesome!

/hold
A nit: would it be easier to print the seed in error message so you don't have to scroll through test logs in prow?

alvaroaleman:
> A nit: would it be easier to print the seed in error message so you don't have to scroll through test logs in prow?

`go test` will suppress the log unless the test fails so I think its fine

PR #21537, opened 2021-03-26
kind dualstack presubmit

aojea:
We use "dual" in the API intead of "DualStack" to be consistent with the other ipfamily modes "ipv4" and "ipv6"


neolit123 (commented):
/lgtm
/hold

PR #21539, opened 2021-03-26
Security Profiles Operator: Increase memory limit for test pod

JAORMX:
We're hitting some issues regarding the pod running out of memory.
This tries to address that.

Signed-off-by: Juan Antonio Osorio Robles <jaosorior@redhat.com>


saschagrunert (approved):
/lgtm

PR #21446, opened 2021-03-19
publishing-bot: add job to run go tests

cpanato:
Fixes: https://github.com/kubernetes/publishing-bot/issues/247

part of: https://github.com/kubernetes/publishing-bot/pull/249

/assign @nikhita

PR #21551, opened 2021-03-29
pull-publishing-bot-validate: revert to use go1.13

nikhita:
See update to go1.16 in https://github.com/kubernetes/test-infra/pull/21446. The job failed after the update to go1.16.

Failure - https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/100616/pull-publishing-bot-validate/1376464585715879936

I'd like to revert to use go1.13 and investigate why the move to go1.16 broke as a follow-up to unblock https://github.com/kubernetes/kubernetes/pull/100616

/assign @xmudrii 
cc @cpanato

### DIFF
--- a/config/jobs/kubernetes/publishing-bot/publishing-bot-presubmits.yaml
+++ b/config/jobs/kubernetes/publishing-bot/publishing-bot-presubmits.yaml
@@ -40,7 +40,7 @@ presubmits:
       path_alias: k8s.io/publishing-bot
     spec:
       containers:
-      - image: golang:1.16
+      - image: golang:1.13
         command:
         - go
         args:


cpanato (approved):
/lgtm

cpanato:
@nikhita how can I reproduce locally the test failure?

nikhita:
For the record - this job just validates that the rules are written correctly, go versions follow a regex, etc so moving to go1.13 doesn't really have a huge impact on the test itself.

nikhita:
> @nikhita how can I reproduce locally the test failure?

maybe running the test in a container or even locally...if it's a go1.16 error, I think it should fail locally as well

xmudrii (approved):
/lgtm

PR #21554, opened 2021-03-29
Update docker version from 19.03.x to 20.10.x

dims:
Moving up!! will help some local-up-cluster testing as well with
containerd.


BenTheElder:
/lgtm
/approve
this will require manual follow-up to switch kubekins-e2e to it.

we already use 20.10.5 in krte

PR #21549, opened 2021-03-29
Bump golang version used in gcb-docker-gcloud to 1.16.2

sbueringer:
Hi,

I would like to create a new verison of the image with Golang 1.16.2. Not sure how the image is build after this PR has been merged.

xref: https://github.com/kubernetes-sigs/cluster-api/issues/4383

PR #21566, opened 2021-03-30
Fix gcb-docker-gcloud image build

sbueringer:
Fixup PR for #21549

Apparently there's only a python2 or python3 package available in Alpine 3.13.3 (which is the one used by the go image). I assume python3 is okay.

Build log of my local test: https://gist.github.com/sbueringer/0df661a1e2a04a9460ee8e8b7d10471f


sbueringer:
Worked this time: https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-test-infra-push-gcb-docker-gcloud/1377330199879749632

PR #21568, opened 2021-03-30
Stop using extensions/v1beta1 in scalability tests

wojtek-t:
This is to fix issues spotted in experimental PR removing extensions/v1beta1: https://github.com/kubernetes/kubernetes/pull/99840

/assign @mborsz

### DIFF
--- a/config/jobs/kubernetes/sig-scalability/sig-scalability-presets.yaml
+++ b/config/jobs/kubernetes/sig-scalability/sig-scalability-presets.yaml
@@ -217,7 +217,7 @@ presets:
       apps/v1/Deployment
       apps/v1/ReplicaSet
       apps/v1/StatefulSet
-      extensions/v1beta1/Ingress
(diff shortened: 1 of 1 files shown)


mborsz:
Wojtek, just to confirm: which version of k8s introduces networking.k8s.io/v1beta1/Ingress? Is this change compatible with e.g. 1.17?

wojtek-t:
> Wojtek, just to confirm: which version of k8s introduces networking.k8s.io/v1beta1/Ingress? Is this change compatible with e.g. 1.17?

yes - it was added in 1.16.

We should eventually migrate to v1/ingress, but that was introduced in 1.19.

wojtek-t:
[And extensions v1beta1 is going to be removed in 1.22, so we need to do that anyway.]

mborsz:
Perfect, thanks!
/lgtm

PR #21574, opened 2021-03-30
crio: run the kubelet in its own cgroup

giuseppe:
add --kubelet-cgroups= to the kubelet args so that it is moved to its
own cgroup.

Signed-off-by: Giuseppe Scrivano <gscrivan@redhat.com>

### DIFF
--- a/config/jobs/kubernetes/sig-node/crio.yaml
+++ b/config/jobs/kubernetes/sig-node/crio.yaml
@@ -17,7 +17,7 @@ periodics:
       - --env=KUBE_SSH_USER=core
       - --gcp-project-type=node-e2e-project
       - --gcp-zone=us-west1-b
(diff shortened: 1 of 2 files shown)


giuseppe:
@harche PTAL

giuseppe:
I've noticed locally that without this change the `[sig-node] Summary API [NodeConformance] when querying /stats/summary should report resource usage through the stats api expand_more` test often fails with error:

```
.Node.SystemContainers[kubelet].CPU:
	Expected
	    <string>: CPUStats
	to match fields: {
	.UsageNanoCores:
		Expected
		    <*uint64 | 0xc0006bf190>: 0
		To satisfy at least one of these matchers: [%!s(*matchers.BeZeroMatcher=&{}) %!s(*gstruct.PointerMatcher=&{0xc0011a4780 Expected
		    <uint64>: 0
		to be >=
		    <int>: 10000})]
	.UsageCoreNanoSeconds:
		Expected
		    <uint64>: 0
		to be >=
		    <int>: 10000000
	}
```

harche:
Thanks @giuseppe

PR #21578, opened 2021-03-30
Status reconciler: officially stop supporting blacklist

chaodaiG:
Part of https://github.com/kubernetes/test-infra/issues/19264

### DIFF
--- a/prow/cmd/status-reconciler/main.go
+++ b/prow/cmd/status-reconciler/main.go
@@ -18,7 +18,6 @@ package main
 
 import (
 	"context"
-	"errors"
 	"flag"
 	"os"
 	"time"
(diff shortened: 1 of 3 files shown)


prow/cmd/status-reconciler/main.go
```diff
@@ -80,7 +78,6 @@ func gatherOptions(fs *flag.FlagSet, args ...string) options {
 	fs.BoolVar(&o.continueOnError, "continue-on-error", false, "Indicates that the migration should continue if context migration fails for an individual PR.")
 	fs.Var(&o.addedPresubmitDenylist, "denylist", "Org or org/repo to ignore new added presubmits for, set more than once to add more.")
 	fs.Var(&o.addedPresubmitDenylistAll, "denylist-all", "Org or org/repo to ignore reconciling, set more than once to add more.")
-	fs.Var(&o.addedPresubmitBlacklist, "blacklist", "[Will be deprecated after May 2021] Org or org/repo to ignore new added presubmits for, set more than once to add more.")
```
petr-muller, line 83: I absolutely do not mind this being gone but if we set a later date ("after May 2021") then we should keep the promise?

chaodaiG: huh. Didn't notice that there is an in consistency between ANNOUNCEMENT.md and here, let's keep both promises. Thank you for pointing out!

chaodaiG:
As @petr-muller , that the warning says end supporting after May, will do it afterwards
/close

PR #21579, opened 2021-03-30
[Kettle] Typo in Error string of api_exceptions.BadRequest

MushuEE:
Saw this snuck in. Tried to write tests for it but because `from google.api_core import exceptions as api_exceptions` is not an import that can be downloaded from pip I can't raise it in the test. I tried mocking but it got increasingly complicated to attempt to mock the import.


amwat (commented):
/lgtm
/approve

PR #21589, opened 2021-03-31
return err from `batchOnboardNewTokenForRepos`

### LINKED ISSUE #21171: HMAC reconciler does not exit non-zero on failure => Postsubmit succeeds on failure.
**What happened**:
The [HMAC reconciler](https://github.com/kubernetes/test-infra/tree/master/prow/cmd/hmac) encountered an error, but the postsubmit job still succeeded: https://oss-prow.knative.dev/view/gs/oss-prow/logs/post-oss-test-infra-reconcile-hmacs/1367626368740757504

If any errors occur that cannot be recovered from, the tool should exit with a non-zero exit code so that the postsubmit ProwJob reports failure.

/help
/good-first-issue

MadhavJivrajani:
- To exit HMAC reconciler with non-zero error code on non-recoverable error
    - Non-zero exit takes place through `handleConfigUpdate()` in `main()` using `.Fatal()`

Fix #21171

### DIFF
--- a/prow/cmd/hmac/main.go
+++ b/prow/cmd/hmac/main.go
@@ -425,6 +425,7 @@ func (c *client) batchOnboardNewTokenForRepos() error {
 			} else {
 				delete(c.currentHMACMap, repo)
 			}
+			return err
 		}
 	}
 


alvaroaleman (approved):
thanks!

prow/cmd/hmac/main.go
```diff
@@ -425,6 +425,7 @@ func (c *client) batchOnboardNewTokenForRepos() error {
 			} else {
 				delete(c.currentHMACMap, repo)
 			}
+			return err
```
cjwagner, line 428: Erroring out early here rather than completing the onboarding for other tokens is problematic. If the onboarding isn't completed for some reason we need to revert the hmac changes.

Rather than erroring out early I'd expect the error to be recorded, but for execution to continue. We should only report this error after cleanup has completed. The same applies to any possible validation errors from line 416.

cjwagner: No worries.  I'll make a fix PR in a bit.

cjwagner: https://github.com/kubernetes/test-infra/pull/21599

PR #21590, opened 2021-03-31
plugins/bugzilla: improve handleClose behavior

AlexNPavel:
This PR changes the behavior of the bugzilla plugin to prevent reopening
closed bugs when a linked PR for the bug is closed. It also adds the
ability to create comments to the bugzilla client and uses this to leave
a comment on bugs when their state is changed due to a closed and
unlinked PR.

/cc @stevekuznetsov

### DIFF
--- a/prow/plugins/bugzilla/bugzilla.go
+++ b/prow/plugins/bugzilla/bugzilla.go
@@ -1187,22 +1187,33 @@ func handleClose(e event, gc githubClient, bc bugzilla.Client, options plugins.B
 			return comment(formatError("removing this pull request from the external tracker bugs", bc.Endpoint(), e.bugId, err))
 		}
 		if options.StateAfterClose != nil {
-			links, err := bc.GetExternalBugPRsOnBug(e.bugId)
+			bug, err := bc.GetBug(e.bugId)
 			if err != nil {
-				log.WithError(err).Warn("Unexpected error getting external tracker bugs for Bugzilla bug.")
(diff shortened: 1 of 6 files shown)


prow/plugins/bugzilla/bugzilla.go
```diff
@@ -1187,22 +1187,33 @@ func handleClose(e event, gc githubClient, bc bugzilla.Client, options plugins.B
...
-				log.WithError(err).Warn("Unexpected error getting external tracker bugs for Bugzilla bug.")
-				return comment(formatError("getting external tracker bugs", bc.Endpoint(), e.bugId, err))
+				log.WithError(err).Warn("Unexpected error getting Bugzilla bug.")
+				return comment(formatError("getting bug", bc.Endpoint(), e.bugId, err))
 			}
-			if len(links) == 0 {
-				bug, err := getBug(bc, e.bugId, log, comment)
-				if err != nil || bug == nil {
-					return err
+			if bug.Status != "CLOSED" {
```
stevekuznetsov, line 1195: Is it safe to assume this is the only such state? Should this be configurable? If not, feel free to cancel hold

/lgtm
/hold

AlexNPavel: I think technically we shouldn't move to `state_after_close` for `POST`, `ON_QA`, or `VERIFIED` either, but those are only possible if there's another PR that's merged already, so it shouldn't happen.

PR #21591, opened 2021-03-31
remove tensorflow/minigo from config as it is archived

BenTheElder:
xref: https://github.com/kubernetes/test-infra/issues/12863 https://github.com/tensorflow/minigo

EDIT: also removing tensorflow/k8s, which is now a redirect to a kubeflow repo and using their bots.


BenTheElder:
CI is happy now.

PR #21592, opened 2021-03-31
Update presubmit tests targets for PD CSI driver and windows

mauriciopoppe:
~~Our objective is to have the following optional presubmit tests in k/k~~:

```
/test pull-kubernetes-e2e-gcp-compute-persistent-disk-csi-driver-windows-2019
/test pull-kubernetes-e2e-gcp-compute-persistent-disk-csi-driver-windows-20h2
```

After a discussion with @jingxu97 we realized that this is not going to work, we're only going to change the target to include windows 20h2

ref https://github.com/kubernetes/kubernetes/pull/100641

### DIFF
--- a/config/jobs/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/gcp-compute-persistent-disk-csi-driver-windows.yaml
+++ b/config/jobs/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/gcp-compute-persistent-disk-csi-driver-windows.yaml
@@ -318,7 +318,7 @@ presubmits:
       description: Kubernetes Integration tests for Kubernetes Master branch and Driver latest build
       testgrid-alert-email: jinxu@google.com
 
-  - name: pull-gcp-compute-persistent-disk-csi-driver-e2e-windows-1-19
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/gcp-compute-persistent-disk-csi-driver-windows.yaml
```diff
@@ -270,8 +270,8 @@ periodics:
     testgrid-tab-name: ci-win2019-provider-gcp-compute-persistent-disk-csi-driver-migration
     description: Kubernetes Integration tests for Kubernetes Master branch and Driver latest
 presubmits:
-  kubernetes-sigs/gcp-compute-persistent-disk-csi-driver:
-  - name: pull-gcp-compute-persistent-disk-csi-driver-e2e-win2019
+  kubernetes-sigs/pull-kubernetes-e2e-gcp-persistent-disk-csi-driver-windows:
```
jingxu97, line 273: this should not be changed, this is the repo name

PR #21594, opened 2021-03-31
azure: add provider-azure-presubmit dashboard

chewong:

- switch test images from versioned to latest so we can pick up the latest kubetest change without waiting for it to get bumped manually
- create provider-azure-presubmit dashboard for master presubmit jobs

/assign @feiskyer


feiskyer (commented):
/lgtm
/approve

chewong:
friendly ping @dims

dims:
/approve
/lgtm

@chewong please confirm all these are optional jobs that run only when someone triggers it?

/hold for question

chewong:
> /approve
> /lgtm
> 
> @chewong please confirm all these are optional jobs that run only when someone triggers it?
> 
> /hold for question

I've confirmed that all of the presubsmit jobs have `always_run: false` and `optional: true`, although we have `run_if_changed: 'azure.*\.go'`. Is that fine?

PR #21605, opened 2021-04-01
add kind dual presubmits

aojea:
Now that dualstack is supported in KIND "officially" , and that those jobs were running for a while without issues we can run them in presubmits.

https://testgrid.k8s.io/sig-network-kind#sig-network-kind,%20dual,%20master&width=20

There are 2 dual-stack presubmit jobs depending on the kube-proxy implementation.
The ipvs one will run only if some ipvs code has changed.
The iptables one will run on every commit, but it will be optional without reporting status to let it soak.

xref: https://github.com/kubernetes-sigs/kind/issues/2172

### DIFF
--- a/config/jobs/kubernetes/sig-network/sig-network-kind.yaml
+++ b/config/jobs/kubernetes/sig-network/sig-network-kind.yaml
@@ -1,10 +1,12 @@
 presubmits:
   kubernetes/kubernetes:
-  - name: pull-kubernetes-e2e-kind-dual-canary
+  - name: pull-kubernetes-e2e-kind-dual
     optional: true
-    always_run: false
-    skip_report: false
+    always_run: true
+    skip_report: true
     decorate: true
+    skip_branches:
+    - release-\d+\.\d+ # per-release settings
     labels:
       preset-dind-enabled: "true"
       preset-kind-volume-mounts: "true"
@@ -49,11 +51,14 @@ presubmits:
       testgrid-tab-name: pr-sig-network-kind, dual
       description: Runs tests against a Dual Stack Kubernetes in Docker cluster
       testgrid-alert-email: antonio.ojea.garcia@gmail.com
-  - name: pull-kubernetes-e2e-kind-ipvs-dual-canary
+  - name: pull-kubernetes-e2e-kind-ipvs-dual
     optional: true
     always_run: false
     skip_report: false
     decorate: true
+    run_if_changed: '^pkg/.*/ipvs/'
+    skip_branches:
+    - release-\d+\.\d+ # per-release settings
     labels:
       preset-dind-enabled: "true"
       preset-kind-volume-mounts: "true"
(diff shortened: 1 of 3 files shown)


aojea:
is this something am I doing wrong?
pull-test-infra-gubernator — Job failed.

```
...............WARNING:root:label event with no labels (multiple changes?)
...ERROR:root:webhook failed signature check
.ERROR:root:webhook failed signature check
........................ERROR:root:unable to fetch 'http://testbed.example.com/_ah/gcs/foo/quux': status code 404
....ERROR:root:failed to get user name: status 403
.ERROR:root:failed to vend token: status 403
```

aojea:
Ben is out this week 
/assign @spiffxp

config/jobs/kubernetes/sig-network/sig-network-kind.yaml
```diff
@@ -1,10 +1,12 @@
 presubmits:
   kubernetes/kubernetes:
-  - name: pull-kubernetes-e2e-kind-dual-canary
+  - name: pull-kubernetes-e2e-kind-dual
     optional: true
-    always_run: false
-    skip_report: false
+    always_run: true
+    skip_report: true
```
spiffxp, line 6: Three always run kind jobs is a lot. Can we replace the other two with this? Or at least one of them? We should definitely not skip reporting unless the intent is to gain confidence to replace

spiffxp (changes requested):

config/jobs/kubernetes/sig-network/sig-network-kind.yaml
aojea: this is to gain confidence ...
we need the ipv6 only because is the only job that gates on ipv6 ...
the dual is a superset of the ipv4, and we have all the other jobs for ipv4 ... 
I'd like to hear from Ben too

aojea:
/hold
let's wait until dual-stack goes GA, tentative this cycle, and we can switch pull-kind for pull-kind-dual

BenTheElder:
> let's wait until dual-stack goes GA, tentative this cycle, and we can switch pull-kind for pull-kind-dual

I feel like of the 3 scenarios people are most likely to have ipv4, or dual, not ipv6 and dual. Let's switch the ipv6 presubmit for dual?

aojea:
> I feel like of the 3 scenarios people are most likely to have ipv4, or dual, not ipv6 and dual. Let's switch the ipv6 presubmit for dual?

the dual job test exercises the same codepath that the ipv4, since its first ipFamily is IPv4 ... we don't have any other gate for ipv6. I prefer switching the ipv4 to dual, so we don't loose any coverage ... and gce and the others are still ipv4 only

BenTheElder:
> and gce and the others are still ipv4 only

yes, which is why I'm not thrilled to not have a closely matching kind job, we need to be able to answer the question if flakes are due to kind (or GCE!) or something else (like totally different cluster mode / tests)

BenTheElder:
longterm I personally still lean towards GCE -> not in k/k presubmit, except maybe the moderate scale tests. we want to drive down test time and avoid testing vendors versus k8s itself.

but until then, we have to compare kind to GCE as the defacto prior presubmit.

aojea:
we can use the periodics job to get signal, if we start to see bugs or regressions I'll reopen the debate
Thanks

PR #21610, opened 2021-04-01
External manager: add a placeholder yaml and update instructions to make things clearer

chaodaiG:


chaodaiG:
The existing instructions don't seem to be very descriptive to users, as learned from @chewong offline
/cc @chewong

chases2 (approved):

PR #21611, opened 2021-04-01
[Kettle] Fix how excluded jobs are found

MushuEE:
Every once in a while we get an object ID that doesn't match expected.

This method should be much safer in finding correct jobs to exclude.

/area Kettle


spiffxp (approved):
/approve
/lgtm

PR #21614, opened 2021-04-01
Prow autobumper: support labels override as cmd arg

chaodaiG:
The problem we might run into while rolling back prow autobump PR, is that there is no existing autobump PR for oncall to put hold on for pausing prow auto deployment process. Since the autobumper runs every 6 hours, oncall might need to wait for 6 hours or wake up at exactly the time when autobumper runs next morning. As I can imagine this will be very annoying.

One way to solve this (The only way I can think about right now) is having autobumper job runs every hour, and only adds 'skip-review' label on 9:05 and 15:05 each day. Instead of wrangling this logic in autobumper code, I would perfer creating two separate prow jobs, one runs continuously without labeling, while the other runs 9:05 and 15:05 on week days and add label on PR. By going this approach, there should always be a new autobump PR created within an hour after rollback

### DIFF
--- a/prow/cmd/generic-autobumper/BUILD.bazel
+++ b/prow/cmd/generic-autobumper/BUILD.bazel
@@ -11,6 +11,7 @@ go_library(
     deps = [
         "//prow/cmd/generic-autobumper/bumper:go_default_library",
         "@com_github_sirupsen_logrus//:go_default_library",
+        "@com_github_spf13_pflag//:go_default_library",
         "@in_gopkg_yaml_v2//:go_default_library",
     ],
 )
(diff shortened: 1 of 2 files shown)


chaodaiG:
/assign @mpherman2 

/cc @alvaroaleman 
Does this make sense to you? How does openshift solve this problem?

alvaroaleman:
> Does this make sense to you? How does openshift solve this problem?

We never used the bumper binary, just its internal logic: https://github.com/openshift/ci-tools/blob/ee2e70a37df1646433c8337435bb9f9504e971dc/cmd/autobumper/main.go#L91-L149

This predates me so I don't know what the original reason for that was (vs using the bumper binary). Our query merges by author, so we don't need the labeling.

PR #21619, opened 2021-04-01
Add another prow autobump job that runs continuously without skip-review label

chaodaiG:
So that test-infra-oncall can always find an open autobump PR to put hold on to pause the auto deployment process


chaodaiG:
/hold
This won't succeed until the next bump of prow

/cc @mpherman2 @cjwagner

cjwagner (approved):
/hold

PR #21621, opened 2021-04-01
Increase pull-kubernetes-e2e-gce-csi-serial timeout to 150m

mauriciopoppe:
We've been seeing some timeouts in the executions of the testsuite, it takes a time close to 2 hours now for successful tests https://prow.k8s.io/job-history/gs/kubernetes-jenkins/pr-logs/directory/pull-kubernetes-e2e-gce-csi-serial and as we're enabling more tests we need to increase the timeout, I increased to 2:30hours

/assign @jingxu97

### DIFF
--- a/config/jobs/kubernetes/sig-storage/sig-storage-gce-config.yaml
+++ b/config/jobs/kubernetes/sig-storage/sig-storage-gce-config.yaml
@@ -136,7 +136,7 @@ presubmits:
         - --provider=gce
         - --stage=gs://kubernetes-release-pull/ci/pull-kubernetes-e2e-gce-csi-serial
         - --test_args=--ginkgo.focus=CSI.*(\[Serial\]|\[Disruptive\]) --ginkgo.skip=\[Flaky\]|\[Feature:.+\]|\[Slow\] --minStartupPods=8
-        - --timeout=120m
+        - --timeout=150m
         image: gcr.io/k8s-testimages/kubekins-e2e:v20210330-fadf59c-master
         resources:
           requests:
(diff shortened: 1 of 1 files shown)


msau42:
Can you give an example of a test run where the job timed out? I want to see if there's opportunity for reducing the time spent in any tests first before we bump the timeout.

I think one thing that would help is if we can run the pd tests in parallel. I think installing pdcsi driver by default in kube-up rather than per testcase will signficantly speed up this test job. cc @mattcary

mauriciopoppe:
Yes, I saw two timeouts while working on https://github.com/kubernetes/kubernetes/pull/100641

- https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/100743/pull-kubernetes-e2e-gce-csi-serial/1377528969095548928
- https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/100641/pull-kubernetes-e2e-gce-csi-serial/1377420215398174720

msau42:
Thanks, it seems like with the tests taking ~80mins and build/up taking ~30minutes we do end up running close to the 2hr limit. I don't see any tests taking a particularly long time, so it doesn't look like there's much opportunity to reduce the test time. Increasing the timeout is ok for now so that we reduce the need for folks to need to retest, but once the branch opens for 1.22, I think we should target the change to install the pdcsi driver by default so we can start running most of these tests in parallel.

/lgtm
/approve

jingxu97:
I think after some new tests (eg. generic epheral storage tests) were added, the total runtime increased from about 1.5 hours to 2 hours since last month

mauriciopoppe:
@msau42 could you please lgtm this one again? there was a failure in a presubmit test, thanks :)

PR #21623, opened 2021-04-02
Add script for creating GCP service account with appropriate pod util permissions.

cjwagner:
Also can be used to add the permissions to an existing SA.
This is similar to [part of the create_build_cluster.sh script](https://github.com/kubernetes/test-infra/blob/fcc93433c62fc98bc8e25d5fc96fff4ff2d7f8f8/prow/create-build-cluster.sh#L107-L122), but it is intended
 - for use with Workload Identity rather than relying on service account key secrets
 - to support tenant/build cluster specific job result buckets

I've tested this and it seems to work properly.
/assign @chaodaiG

### DIFF
--- a/prow/make_prowjob_sa.sh
+++ b/prow/make_prowjob_sa.sh
@@ -1,0 +1,81 @@
+#!/usr/bin/env bash
+# Copyright 2021 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This script is used to create a new GCP service account with permissions need by pod utilities to upload job results to GCS.
+# ProwJobs can be configured to use this identity by associating the GCP SA with a K8s SA via workload identity, then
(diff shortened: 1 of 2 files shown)


prow/make_prowjob_sa.sh
```diff
@@ -0,0 +1,81 @@
...
+createSA() {
+  gcloud beta iam service-accounts create \
+    ${SA_NAME} \
+    --project="${PROJECT_ID}" \
+    --description="${SA_DESCRIPTION}" \
+    --display-name="${SA_DISPLAY_NAME}"
+}
+
+authorizeUpload() {
+  gsutil acl ch -u "${SA}:O" "${BUCKET}"
```
chaodaiG, line 78: I think there is another command for uniform bucket level

chaodaiG (commented):
https://github.com/kubernetes/test-infra/blob/e7f321167701f741e32174b041bd009adbce5a4a/prow/create-build-cluster.sh#L112

We might want to update ./create-build-cluster.sh in this PR as well, do you see any blockers?

cjwagner:
> We might want to update ./create-build-cluster.sh in this PR as well, do you see any blockers?

No real blockers, but I'd like to do this in a separate PR once I've tested everything to confirm I'm not missing any steps. I'll also include WI setup.

I did fix the `gsutil` command in `create-build-cluster.sh` for uniform bucket ACLs.

prow/make_prowjob_sa.sh
cjwagner: :+1: Seems like the `iam` command is needed instead of `acl` for this case.
https://cloud.google.com/storage/docs/uniform-bucket-level-access
https://cloud.google.com/storage/docs/gsutil/commands/iam
https://cloud.google.com/storage/docs/gsutil/commands/acl

chaodaiG (approved):

prow/make_prowjob_sa.sh
chaodaiG: Feel free do it in separate PR if preferred

PR #21624, opened 2021-04-02
Crier gerrit reporter: skip reporting if job from newer patchset already exists

chaodaiG:
In a gerrit PR, when a new change is pushed to it, the previous batch of presubmit jobs were aborted, if all jobs haven't finished by the time of abortion then gerrit won't report because aborted state wouldn't trigger a report. But if there was a single job failed before the abortion then crier wouldn't report the failed prowjob because by the time of failure the other jobs were still running, this is also expected, this job wouldn't be reported when the abortion happen either, this is also expected. However, I have observed lately that these earlier failed jobs were reported long time later, in the range of 3-12 hours according to my observation. It perhaps is related to prow bump that restarted crier which makes it reconcile all prowjobs and triggered the reporting of prowjob failed long time ago.

This PR is fixing this problem by stopping crier from reporting if there are prowjobs exist for the same PR that are from a later patchset

### DIFF
--- a/prow/crier/reporters/gerrit/reporter.go
+++ b/prow/crier/reporters/gerrit/reporter.go
@@ -137,7 +137,10 @@ func (c *Client) ShouldReport(ctx context.Context, log *logrus.Entry, pj *v1.Pro
 		return true
 	}
 
-	allPJAgreeReport := func(labels []string, singlePRAgree func(pj *v1.ProwJob) bool) bool {
+	// allPJsAgreeToReport is a helper function that queries all prowjobs based
+	// on provided labels and run each one through singlePJAgreeToReport,
+	// returns false if any of the prowjob doesn't agree.
+	allPJsAgreeToReport := func(labels []string, singlePJAgreeToReport func(pj *v1.ProwJob) bool) bool {
 		selector := map[string]string{}
 		for _, l := range labels {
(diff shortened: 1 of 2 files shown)


chaodaiG:
/cc @chizhg 
Who originally reported this problem

prow/crier/reporters/gerrit/reporter.go
```diff
@@ -136,28 +137,58 @@ func (c *Client) ShouldReport(ctx context.Context, log *logrus.Entry, pj *v1.Pro
...
+			return -1
+		}
+		intPs, err := strconv.Atoi(ps)
+		if err != nil {
+			log.Warnf("Found non integer label for %s: %s in prowjob %s", client.GerritPatchset, ps, pj.Name)
+			return -1
+		}
+		return intPs
 	}
+	patchsetNum := patchsetNumFromPJ(pj)
```
chizhg, line 179: The code logic is a bit confusing since all the prowjob vars are named as `pj`.. Probably better to name them as `curtPJ` and `otherPJ`, whichever that can be clearer.

chaodaiG: Added a comment explaining this, hopefully it helps

chizhg:
Thanks a lot @chaodaiG !!

/lgtm
/hold
In case Cole and Erick have other comments.

PR #21638, opened 2021-04-02
Actually pass the variants KUBETEST2_VERSION as a docker build-arg.

amwat:
/cc @BenTheElder @MushuEE @spiffxp


amwat:
Thanks @dims :)

PR #21641, opened 2021-04-02
Switch ci-cri-containerd-node-e2e-benchmark to correct containerd master

dims:
`containerd/cri` was deprecated and the contents of it moved to `containerd/containerd` a few months ago. Hoping this will help bring https://testgrid.k8s.io/sig-node-containerd#node-e2e-benchmark back to life or at least on a path where we can fix it.


hasheddan (approved):
/lgtm

Issue #21167, opened 2021-03-04
Sidecar doesn't properly handle best effort upload and races with entrypoint.


cjwagner:
We've recently had multiple presubmits report failure in spyglass and finished.json even though they were actually aborted due to more recent commits being pushed. The ProwJob resource disagrees with finished.json and properly indicates the aborted state.

I think we are encountering this race: https://github.com/kubernetes/test-infra/blob/2a5710e4ad2ee2942634c60e5c0f753f4012a877/prow/sidecar/run.go#L90-L95

This option is designed to help mitigate when used in conjunction with an appropriate graceperiod timeout: https://github.com/kubernetes/test-infra/blob/2a5710e4ad2ee2942634c60e5c0f753f4012a877/prow/sidecar/options.go#L50-L77

However, the option is always set to false, we don't have a way to configure it: https://github.com/kubernetes/test-infra/blob/276f55e384c73100f0a3c2e8f27cf126104f93d6/prow/pod-utils/decorate/podspec.go#L723

<hr>

I think the comment describing the race is misleading and the logic could be improved here. In particular we don't actually try to perform the upload twice if an interrupt is received, we just immediately begin the upload then terminate. I would expect sidecar to immediately begin the upload, but then check/wait for the marker files to be written, and reupload when the markers are present.
Based on the comment this sounds like what was actually intended.

/help

viveksyngh:
I would like to work on this issue

stevekuznetsov:
I think that's what the behavior did when I originally wrote it @cjwagner lol. Perhaps that's been broken since. Unfortunately hard to e2e test this component.

stevekuznetsov:
@viveksyngh sure thing - shoot me an /assign and I'll review when you are ready

stevekuznetsov:
Actually @cjwagner as of https://github.com/kubernetes/test-infra/pull/20923 we should intelligently set the Pod's `terminationGracePeriodSeconds` to be larger than the `entrypoint`'s grace period, so perhaps it would be safe to default to ignoring interrupts if we think that the default buffer we add is enough to do the upload.

cjwagner:
> so perhaps it would be safe to default to ignoring interrupts if we think that the default buffer we add is enough to do the upload.

That would be one way to mitigate this problem and it would probably be sufficient most of the time, but I don't think it is as robust as doing a best effort upload then continuing to wait and try the real upload. The double upload seems strictly better since it ensures that we upload something even if the main upload takes too long to complete.

[...]

PR #21644, opened 2021-04-03
prow: sidecar: allow configuring ignore interrupts 

viveksyngh:
Add a field `upload_ignores_interrupts` to decoration config spec to configure ignore interrupts while performing
upload. It also tries to peform best-effort upload when `upload_ignores_interrupts` is set to `false` and interrupt
is received. While performing best-effort upload if main upload is ready it interrupts best-effort upload and switches
to main upload.

Fixes: #21167

### DIFF
--- a/prow/apis/prowjobs/v1/types.go
+++ b/prow/apis/prowjobs/v1/types.go
@@ -373,6 +373,10 @@ type DecorationConfig struct {
 
 	// CensoringOptions exposes options for censoring output logs and artifacts.
 	CensoringOptions *CensoringOptions `json:"censoring_options,omitempty"`
+
+	// UploadIgnoresInterrupts causes sidecar to ignore interrupts for the upload process in
+	// hope that the test process exits cleanly before starting an upload.
+	UploadIgnoresInterrupts *bool `json:"upload_ignores_interrupts,omitempty"`
 }
 
 type CensoringOptions struct {
@@ -434,6 +438,7 @@ func (g *CensoringOptions) ApplyDefault(def *CensoringOptions) *CensoringOptions
 		merged.ExcludeDirectories = def.ExcludeDirectories
 	}
 	return &merged
+
 }
 
 // Resources holds resource requests and limits for
@@ -531,6 +536,10 @@ func (d *DecorationConfig) ApplyDefault(def *DecorationConfig) *DecorationConfig
 	}
 	if merged.CensorSecrets == nil {
 		merged.CensorSecrets = def.CensorSecrets
+	}
+
+	if merged.UploadIgnoresInterrupts == nil {
+		merged.UploadIgnoresInterrupts = def.UploadIgnoresInterrupts
 	}
 
 	return &merged
--- a/prow/pod-utils/decorate/podspec.go
+++ b/prow/pod-utils/decorate/podspec.go
@@ -742,7 +742,9 @@ func decorate(spec *coreapi.PodSpec, pj *prowapi.ProwJob, rawEnv map[string]stri
 		wrappers = append(wrappers, *wrapperOptions)
 	}
 
-	sidecar, err := Sidecar(pj.Spec.DecorationConfig, blobStorageOptions, blobStorageMounts, logMount, outputMount, encodedJobSpec, !RequirePassingEntries, !IgnoreInterrupts, secretVolumeMounts, wrappers...)
+	ignoreInterrupts := pj.Spec.DecorationConfig.UploadIgnoresInterrupts != nil && *pj.Spec.DecorationConfig.UploadIgnoresInterrupts
+
+	sidecar, err := Sidecar(pj.Spec.DecorationConfig, blobStorageOptions, blobStorageMounts, logMount, outputMount, encodedJobSpec, !RequirePassingEntries, ignoreInterrupts, secretVolumeMounts, wrappers...)
 	if err != nil {
 		return fmt.Errorf("create sidecar: %v", err)
 	}
@@ -793,8 +795,6 @@ func DetermineWorkDir(baseDir string, refs []prowapi.Refs) string {
 const (
 	// RequirePassingEntries causes sidecar to return an error if any entry fails. Otherwise it exits cleanly so long as it can complete.
 	RequirePassingEntries = true
-	// IgnoreInterrupts causes sidecar to ignore interrupts and hope that the test process exits cleanly before starting an upload.
-	IgnoreInterrupts = true
 )
 
 func Sidecar(config *prowapi.DecorationConfig, gcsOptions gcsupload.Options, blobStorageMounts []coreapi.VolumeMount, logMount coreapi.VolumeMount, outputMount *coreapi.VolumeMount, encodedJobSpec string, requirePassingEntries, ignoreInterrupts bool, secretVolumeMounts []coreapi.VolumeMount, wrappers ...wrapper.Options) (*coreapi.Container, error) {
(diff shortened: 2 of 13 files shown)


alvaroaleman:
@viveksyngh can you explain the motivation/usecase for this new knob?

viveksyngh:
@alvaroaleman Sorry I didn't link the issue but I have created to fix this issue #21167

prow/apis/prowjobs/v1/types.go
```diff
@@ -373,6 +373,10 @@ type DecorationConfig struct {
 
 	// CensoringOptions exposes options for censoring output logs and artifacts.
 	CensoringOptions *CensoringOptions `json:"censoring_options,omitempty"`
+
+	// IgnoreInterrupts causes sidecar to ignore interrupts and
+	// hope that the test process exits cleanly before starting an upload.
+	IgnoreInterrupts *bool `json:"ignore_interrupts,omitempty"`
```
cjwagner, line 379: I think we should rename this to something that indicates that this applies to the job result upload only and that interrupts will still be forwarded to the test process, e.g. `upload_ignores_interrupts`

prow/pod-utils/decorate/podspec.go
```diff
@@ -742,7 +742,14 @@ func decorate(spec *coreapi.PodSpec, pj *prowapi.ProwJob, rawEnv map[string]stri
 		wrappers = append(wrappers, *wrapperOptions)
 	}
 
-	sidecar, err := Sidecar(pj.Spec.DecorationConfig, blobStorageOptions, blobStorageMounts, logMount, outputMount, encodedJobSpec, !RequirePassingEntries, !IgnoreInterrupts, secretVolumeMounts, wrappers...)
+	var ignoreInterrupts bool
+	if pj.Spec.DecorationConfig.IgnoreInterrupts == nil { // if nil, set it to false
+		ignoreInterrupts = false
+	} else {
+		ignoreInterrupts = *pj.Spec.DecorationConfig.IgnoreInterrupts
+	}
```
cjwagner, line 750: nit: This is a bit simpler
```suggestion
	ignoreInterrupts := pj.Spec.DecorationConfig.IgnoreInterrupts != nil && *pj.Spec.DecorationConfig.IgnoreInterrupts
```

prow/sidecar/run.go
```diff
@@ -108,6 +108,10 @@ func (o Options) Run(ctx context.Context) (int, error) {
 		return 0, fmt.Errorf("could not resolve job spec: %v", err)
 	}
 
+	entries := o.entries()
+	buildLogs := logReaders(entries)
+	metadata := combineMetadata(entries)
```
cjwagner, line 113: `logReaders` and `combineMetadata` cannot be called this early. They need to be called immediately before `doUpload` so that we don't open and read the files before they've been written.

prow/sidecar/run.go
```diff
@@ -125,7 +129,12 @@ func (o Options) Run(ctx context.Context) (int, error) {
 				// second upload but we can tolerate this as we'd rather get SOME
 				// data into GCS than attempt to cancel these uploads and get none.
 				logrus.Errorf("Received an interrupt: %s, cancelling...", s)
-				cancel()
+
+				err = o.doUpload(spec, false, true, metadata, buildLogs)
```
cjwagner, line 133: This case skips some of the steps before `doUpload` is called, in particular checking for the deprecated wrapper options and censoring the data before upload. Please put all this common logic in a function that can be called in both places.

cjwagner, line 133: We need this preemptive upload to be interrupted if we see that the marker files are written and we're able to begin the main upload.
To achieve this we should instrument the call stack of `doUpload` with a context that can be used to cancel the upload early and then pass the `ctx` context here. That way if the `wait` function returns on line 148 we'll cancel the preemptive upload on line 150 and begin the main upload.

prow/sidecar/run.go
```diff
@@ -125,7 +129,12 @@ func (o Options) Run(ctx context.Context) (int, error) {
...
+				if err != nil {
+					logrus.Errorf("Failed to perform best-effort upload: %v", err)
+				}
```
cjwagner, line 136: Please also log an `Info` level log if the best-effort upload is successful. That will make it easier to determine if the best-effort upload overlaps with the main upload temporally while debugging.

cjwagner (changes requested):
Thanks for working on this. Exposing the `ignore_interrupts` option is nice, but the fix to the best effort upload logic is even better. With that fix the option hopefully won't be needed at all.

prow/apis/prowjobs/v1/types.go
viveksyngh: Updated it to `upload_ignores_interrupts`

prow/pod-utils/decorate/podspec.go
viveksyngh: Updated it

prow/sidecar/run.go
viveksyngh: Move these calls to immediately before `doUpload` and `bestEffortUpload` call.

prow/sidecar/run.go
viveksyngh: Added a new method `preUpload` which performs steps required before upload and call it before both upload process.

viveksyngh: Handled this by adding a new method `doBestEffortUpload` which uses `doUpload` but also switches context when main upload is ready.

prow/sidecar/run.go
viveksyngh: Added this inside `doBestEffortUpload` method.

cjwagner (commented):
Sorry for the delayed review.

prow/sidecar/run.go
```diff
@@ -213,6 +214,45 @@ func combineMetadata(entries []wrapper.Options) map[string]interface{} {
...
+
+	go func() {
+		errChan <- o.doUpload(spec, passed, aborted, metadata, logReaders)
+	}()
+
+	for {
+		select {
+		case <-ctx.Done():
+			logrus.Infof("Interrupting best-effort upload as main upload is ready")
+			return
```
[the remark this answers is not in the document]
viveksyngh: Update the functions to pass the context in stack and removed separate go routine

alvaroaleman:
@cjwagner is this change really addressing the issue in question? From your original [comment](https://github.com/kubernetes/test-infra/issues/21167):
```
I think the comment describing the race is misleading and the logic could be improved here. In particular we don't actually try to perform the upload twice if an interrupt is received, we just immediately begin the upload then terminate. I would expect sidecar to immediately begin the upload, but then check/wait for the marker files to be written, and reupload when the markers are present
```

cjwagner (changes requested):
Looks good to me except for one thread safety issue.

cjwagner:
> @cjwagner is this change really addressing the issue in question?

Yes, I believe so. It isn't very clear from PR title/description, but the main change in this PR is the fix of the best effort upload to actually do what the comment described. The PR also makes the interrupt ignoring behavior configurable which may not be useful if the main fix solves the problem, but it shouldn't hurt though.

prow/sidecar/run.go
```diff
@@ -146,14 +156,12 @@ func (o Options) Run(ctx context.Context) (int, error) {
...
-		if err := o.censor(); err != nil {
-			logrus.Warnf("Failed to censor data: %v", err)
-		}
-	}
+	o.preUpload()
+
 	buildLogs := logReaders(entries)
 	metadata := combineMetadata(entries)
-	return failures, o.doUpload(spec, passed, aborted, metadata, buildLogs)
+	ctx = context.Background()
```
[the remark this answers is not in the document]
viveksyngh: Fixed this.

cjwagner (approved):
Thank you for the contribution, Vivek!

Issue #21006, opened 2021-02-24
`needs-rebase` application is delayed, only seems to be triggered on comment


ehashman:
**What happened**:

For many PRs in k/k that I've triaged, there were conflicts present for days, but the needs-rebase label is not getting applied automatically. The bot only seems to apply it upon prompting.

**What you expected to happen**:

`needs-rebase` should be applied as soon as a merge conflict is present (with some reasonable buffer).

**How to reproduce it (as minimally and precisely as possible)**:

Here's an example of something that needs a rebase but is missing the label: https://github.com/kubernetes/kubernetes/pull/88172

**Please provide links to example occurrences, if any**:

- https://github.com/kubernetes/kubernetes/pull/92888#issuecomment-762539759
- https://github.com/kubernetes/kubernetes/pull/98825
- https://github.com/kubernetes/kubernetes/pull/99000
- etc.

**Anything else we need to know?**:

Right now, when we put a comment to trigger the bot into adding a needs-rebase label, it resets the stale/rotten timer on PRs with "activity". Ideally we'd like a way to avoid that.

alvaroaleman:
The basic issue is that a state change to needs-rebase doesn't have an event associated, as its the result of the base branch changing (usually). So the only way to figure this out is to scan all PRs, which is expensive in terms of github api quota.

What happens today is that we check all PRs every 24 hours for this: https://github.com/kubernetes/test-infra/blob/ae42f5b65b6577cf20f0c58d238d96524ab6927c/prow/external-plugins/needs-rebase/main.go#L73

I would assume that the plugin [gets heavily throttled when it does its check](https://github.com/kubernetes/test-infra/blob/ae42f5b65b6577cf20f0c58d238d96524ab6927c/prow/external-plugins/needs-rebase/main.go#L108) and Prow gets bumped before it makes much progress, resulting in a reset of the 24h timer. This would also explain why it usually does something on the weekend (no bumps, so even if slow, it makes so progress). Maybe just reduce that interval to 12h?

ehashman:
I'm certain that PRs are currently being missed right now. The example PR of https://github.com/kubernetes/kubernetes/pull/88172 has needed a rebase for well over 24h -- I think the incompatible change was made 5mo ago.

ehashman:
https://github.com/kubernetes/kubernetes/pull/68726 another example of a PR that needed a rebase for months but it only got applied after an action (in this case, me closing the PR)

nikhita:
[...]

PR #21645, opened 2021-04-03
Needs rebase: Better cope with 1k Github search result size limit

alvaroaleman:
GitHub has a 1k results per query hard limit:
https://github.community/t/graphql-github-api-how-to-get-more-than-1000-pull-requests/13838/10

We currently hit this in the prow.k8s.io instance, which is why the
periodic rescan doesn't properly update all PRs. This change fixes that
by:
* Using a distinct query per org and one for all repos (if any)
* Making k/k its special snowflake that doesn't get only one but two
  unique queries, as the repo alone has currently 1.1k open PRs

Fixes https://github.com/kubernetes/test-infra/issues/21006
/assign @chaodaiG 
/cc @ehashman @nikhita

### DIFF
--- a/prow/external-plugins/needs-rebase/plugin/plugin.go
+++ b/prow/external-plugins/needs-rebase/plugin/plugin.go
@@ -25,6 +25,7 @@ import (
 
 	githubql "github.com/shurcooL/githubv4"
 	"github.com/sirupsen/logrus"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 
 	"k8s.io/test-infra/prow/config"
 	"k8s.io/test-infra/prow/github"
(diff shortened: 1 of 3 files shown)


prow/external-plugins/needs-rebase/plugin/plugin.go
```diff
@@ -133,17 +136,45 @@ func HandleAll(log *logrus.Entry, ghc githubClient, config *plugins.Configuratio
...
 	}
-	var buf bytes.Buffer
-	fmt.Fprint(&buf, "archived:false is:pr is:open")
+	// GitHub hard caps queries at 1k results, so do one query per org and one for
+	// all repos. Ref: https://github.community/t/graphql-github-api-how-to-get-more-than-1000-pull-requests/13838/11
+	var queries []string
 	for _, org := range orgs {
-		fmt.Fprintf(&buf, " org:\"%s\"", org)
+		// https://img.17qq.com/images/crqhcuueqhx.jpeg
+		if org == "kubernetes" {
```
chaodaiG, line 144: This is unfortunate for k8s/kubernetes, and hopefully won't happen in any other repo. When it does happen we might want to make this configurable like `needs-rebase: large-org: true`, but for now 🤷‍♂️

chaodaiG (approved):
/hold

prow/external-plugins/needs-rebase/plugin/plugin.go
alvaroaleman: Yeah, so obviously this is pretty clunky, but it would be more than just a bool to plug through, we'd need to tell which repo to exclude from the org query and how to shard the query for that repo

alvaroaleman (commented):
/hold cancel

PR #21646, opened 2021-04-03
ci-cri-containerd-node-e2e-benchmark should be checking out containerd/containerd

dims:
Missed a spot in https://github.com/kubernetes/test-infra/pull/21641


hasheddan (approved):
/retitle ci-cri-containerd-node-e2e-benchmark should be checking out containerd/containerd

/lgtm

PR #21649, opened 2021-04-04
pull-release-cluster-up: Only run when push-build.sh is modified

justaugustus:
IIUC, the original intent of this job was to ensure changes to the
k/release shell library did not result in breakages in jobs using
bootstrap scenarios.

Now that push-build.sh is legacy (and self-contained), we only need to
trigger this cluster job when that file is changed.

/assign @hasheddan @puerco


hasheddan (approved):
/lgtm

Issue #18782, opened 2020-08-11
Prow appears to merge before passing tests due to a Crier and Tide race 


chases2:
I think there's a race condition where Crier marks GitHub's status as Mergeable at the same time that Tide tries to merge it.
If Tide wins, then Crier won't update a merged PR and it _appears_ that a PR has merged before all the GitHub statuses are green.

Note that this does not result in Tide merging a PR that shouldn't be merged; it seems to only affect the status contexts that appear.

Examples:
https://github.com/kubernetes/website/pull/22643
https://github.com/kubernetes/test-infra/pull/17874

Related Issues:
https://github.com/kubernetes/test-infra/issues/16472

alvaroaleman:
This isn't crier, this is the tide status controller which runs independently of the tide sync controller and might not update the tide status to green before a PR gets merged, ref: https://github.com/kubernetes/test-infra/issues/16472#issuecomment-640874553

Issue #16472, opened 2020-02-25
Allow protecting the tide github status context


chaodaiG:
**What would you like to be added**:
When reporting Prow jobs status for PRs in Github UI, also note whether it's required or optional as defined by Prow job itself

<img width="723" alt="Screen Shot 2020-02-25 at 8 08 33 AM" src="https://user-images.githubusercontent.com/45011425/75265812-47434500-57a6-11ea-974a-b0ce27513923.png">

**Why is this needed**:
This can help remove the ambiguity introduced by hidden `required` jobs that only required by `tide`but not explicitly marked as `required` by admins. Also, this will make it possible to only require `tide` as requirement

Related: https://github.com/knative/test-infra/issues/1742

stevekuznetsov:
Looks like you have branch protection configured and it's clearly marking the job as required on the right of your screenshot -- is what you're asking for a separate feature?

chaodaiG:
The example was from a PR in the current repo. The goal of this feature request is to make us being able to only add tide on branch protection instead of adding each individual job

chaodaiG:
The example was from a PR in the current repo. The goal of this feature request is to make us being able to only add tide on branch protection instead of adding each individual job

chaodaiG:
To be more clear, this is just a UI change request. As functionally we can already only add `tide` to branch protection, since `tide` already implicitly requires all jobs that are not marked `optional` to pass. So the ask is just labeling each job more explicitly in Github UI. 

IMHO this is achievable if the status updater already knows about `optional`, otherwise it may not worth the effort to wire them together

alvaroaleman:
/retitle Allow protecting the tide github status context

> The example was from a PR in the current repo. The goal of this feature request is to make us being able to only add tide on branch protection instead of adding each individual job

This can be done today _but_ will result in an elevated error rate from tide, because the part that updates the status is async from the part that merges, which means that the merging may fail because the github status context wasn't updated yet.

Generally this is a huge +1 from me, as its conceptionally much saner than protecting each individual github status context, because that requires a giant amount of tokens and the settings are only eventual consistent with job config changes.

chases2:
Related to this asynchronousness, sometimes PRs will merge before their tide status is updated. This leads to some PRs (like https://github.com/kubernetes/test-infra/pull/17874) that will merge even while the Tide status says the PR is "not mergeable". This has been observed to happen on multiple repositories.

/remove-lifecycle stale

matthyx:
[...]

Issue #6289, opened 2018-01-14
If tide is enabled on a repo, the tide status context should be required by branch protection


cblecker:
/area prow
cc: @fejta @spxtr

cblecker:
@fejta Any update on this?

fejta:
/unassign
/assign @cjwagner 
/area prow

Making the tide context required breaks tide.

cblecker:
Wouldn't this be a race condition, then?

What about for repos where there isn't other prow testing? If it's not required, then anyone with write access would be able to manually merge the PR as long as the CLA context passes (this is why we require the SQ context itself)

fejta:
True true, which are both good reasons for considering to address it rather than closing... although removing everyone's write access would solve that problem :)

Cole has the most state here and will let him weigh in.

cjwagner:
Making the tide context required doesn't work well with Tide, but it doesn't completely break us. It just causes us to (almost always) fail the first attempt at every merge and have to retry the next sync loop.

The issue is that statuses are updated from a separate controller thread instead of being updated as part of the main sync controller loop and that statuses are updated with lower priority than syncing (since we care more about actually merging things than updating statuses correctly).
In order to allow the tide status to be required, the sync controller would need to wait for the status to be updated to green before merging the PR, otherwise the merge fails because a required status is not green.

An easy, but mildly hacky solution would be to have the sync controller set the status to green before merging when necessary. There is still potential for a race with the status controller (especially if the status controller is 'lapped' by the sync controller), but i think a low chance of setting a red status on a merged PR is a better alternative to failing the first attempt at every merge.

spiffxp:
/area prow/tide
Since we're paying attention to status contexts, is this worth address or do we close this as good enough for now?

cblecker:
I would *still* really like this fwiw

stevekuznetsov:
Isn't this as noted by Cole fundamentally going to introduce weird races with the two `tide` controllers and therefore very unlikely to actually happen?

cblecker:
/remove-lifecycle rotten

The biggest use case for this is where you want to require using tide for merging, and not show the green button when tests are passing, but other conditions (like required/missing labels) aren't met.

sftim:
[...]

PR #21650, opened 2021-04-05
Tide: Set github status to success prior to merging

alvaroaleman:
This PR changes Tide to set its status context to success prior to merging. It is https://github.com/kubernetes/test-infra/pull/20892 plus a second commit to synchronize this with the status controller, ref https://github.com/kubernetes/test-infra/pull/20892#discussion_r578693503

Unblocks https://github.com/kubernetes/test-infra/issues/6289
Supersedes and thereby closes https://github.com/kubernetes/test-infra/pull/20892
Fixes https://github.com/kubernetes/test-infra/issues/16472
Fixes https://github.com/kubernetes/test-infra/issues/18782

/assign @cjwagner

### DIFF
--- a/prow/tide/status.go
+++ b/prow/tide/status.go
@@ -80,6 +80,11 @@ type statusController struct {
 	// the minimum status update period.
 	lastSyncStart time.Time
 
+	// dontUpdateStatus contains all PRs for which the Tide sync controller
+	// updated the status to success prior to merging. As the name suggests,
+	// the status controller must not update their status.
+	dontUpdateStatus threadSafePRSet
+
 	sync.Mutex
 	poolPRs          map[string]PullRequest
 	requiredContexts map[string][]string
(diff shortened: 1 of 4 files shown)


prow/tide/status.go
```diff
@@ -642,3 +647,36 @@ func contextCheckerGetterFactory(cfg *config.Config, gc git.ClientFactory, org,
 		return contextPolicy, nil
 	}
 }
+
+type minimalPR struct {
```
stevekuznetsov, line 651: nit: what is minimal about this? maybe `pullRequestIdentifier`?

prow/tide/tide.go
```diff
@@ -408,6 +408,7 @@ func (c *Controller) Sync() error {
 	c.sc.requiredContexts = requiredContextsMap(filteredPools)
 	select {
 	case c.sc.newPoolPending <- true:
+		c.sc.dontUpdateStatus.reset()
```
stevekuznetsov, line 411: What kind of guarantees do we have about the overall memory footprint of the set given how often we expect to call this?

alvaroaleman: This set only contains PRs that we merged and is reset after status controller sync finished + tide sync finished it querying. The mem usage of this is absolutely negligible.

fejta (approved):
/hold

Worth putting this functionality behind a flag?

alvaroaleman:
> Worth putting this functionality behind a flag?

I don't really think that is needed as the worst case scenario is that it doesn't work, which means we stay were we are today. And I would prefer to not increase the already big configuration surface further.

/hold cancel

PR #21652, opened 2021-04-05
Define a team/katacoda label for k/website

sftim:
[Katacoda](https://www.katacoda.com/) provide hosted Kubernetes learning environments for https://kubernetes.io/docs/tutorials/ - this label will filtering on those specific issues.

We already have the label, but it's manually-managed and Prow doesn't recognize it.

### DIFF
--- a/label_sync/labels.yaml
+++ b/label_sync/labels.yaml
@@ -1494,6 +1494,12 @@ repos:
         target: both
         prowPlugin: label
         addedBy: anyone
+      - color: d2b48c
+        description: Issues with the Katacoda infrastructure for tutorials
+        name: team/katacoda
+        target: issues
+        prowPlugin: label
+        addedBy: anyone
   kubernetes-sigs/cluster-api:
     labels:
       - color: c7def8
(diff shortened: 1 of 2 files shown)


label_sync/labels.yaml
```diff
@@ -1494,6 +1494,12 @@ repos:
         target: both
         prowPlugin: label
         addedBy: anyone
+      - color: d2b48c
+        description: Issues with the Katacoda infrastructure for tutorials
+        name: team/katacoda
```
spiffxp, line 1499: Note AFAIK you also need to update prow config to allow people to /label team/katacoda

spiffxp: It's likely plugins.yaml

spiffxp (approved):
Need to fix tests and then I'll lgtm for real
/hold
Remove when updated

sftim:
Might take a while before I can get this right, unless there's a container made that has the relevant tools ready to use.

fejta:
/uncc
LGTM

sftim:
/hold cancel
as generated Markdown is now fixed

PR #21654, opened 2021-04-05
Skips GPUDevicePlugin test for Windows test runs

claudiubelu:
Most jobs are already skipping the test [1] by specifying it by name in their regex. Recently [2], the tag ``[Feature:GPUDevicePlugin]`` was added to the test name, which means that most of the GCE tests will also skip it since they have ``\[Feature:.+\]`` in their
skip regex, with the exception of the ``ci-kubernetes-e2e-windows-gce-alpha-features`` job.

This commit adds ``\[Feature:GPUDevicePlugin\]`` to the mentioned job's regex, so it will be skipped there as well.

[1]: ``[sig-windows] Device Plugin should be able to create a functioning device plugin for Windows``
[2]: https://github.com/kubernetes/kubernetes/pull/100815


pjh:
/lgtm

Thanks Claudiu, and thanks for the helpful PR description.

PR #21657, opened 2021-04-05
Adjust prow stale alert to be day of week based, and more strict

chaodaiG:
Since k8s prow is auto-deployed now, it's much more predictable that there should be at least one bump in 48 hours. Adjusting the prometheus alert rule to 2 days, and let it relax +2 days if Monday or Tuesday


cjwagner (approved):

chaodaiG:
Just rebased, may I have another approve please ?

cjwagner (approved):

PR #21664, opened 2021-04-05
Deploy kubernetes external secrets controller and CRD in k8s-prow-bui…

chaodaiG:
…lds cluster

### DIFF
--- a/config/prow/cluster/build_kubernetes-external-secrets_crd.yaml
+++ b/config/prow/cluster/build_kubernetes-external-secrets_crd.yaml
@@ -1,0 +1,1 @@
+./kubernetes-external-secrets_crd.yaml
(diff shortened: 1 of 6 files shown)


config/prow/cluster/build_kubernetes-external-secrets_crd.yaml
```diff
@@ -0,0 +1 @@
+./kubernetes-external-secrets_crd.yaml
```
fejta, line 1: Shouldn't this be a yaml file?

chaodaiG: This is a symlink. I meant to use the same file, but this rule:
```
 component(
        "kubernetes-external-secrets",
        "deployment",
        "rbac",
        "service",
        "crd",
        cluster = BUILD_CLUSTER,
    ),
```
kept on complaining the rule `kubernetes-external-secrets-deployment` already exists, and creating this symlink seems solving the problem

fejta (approved):

PR #21669, opened 2021-04-06
Add disable_prowjob_analysis: true to Windows containerd-integration

adelina-t:
Adds disable_prowjob_analysis: true to Windows containerd-integration
tests as these tests are not run from a prow instance and do not
have any pod info.


marosset (approved):
/lgtm
/approve

PR #21670, opened 2021-04-06
Fix ci-kubernetes-e2e-windows-containerd-gce-1-21 to reflect the master job

dims:
https://testgrid.k8s.io/sig-node-containerd#gce-windows-2019-containerd-1.21 is running a different set of things since we cloned it from 1.20 ... instead let's make it reflect the master job ( https://testgrid.k8s.io/sig-node-containerd#gce-windows-2019-containerd-master ). the only difference from the master job would be the which branch we pull k8s release from and the image we use to run the job.


adisky (commented):
/lgtm

PR #21672, opened 2021-04-06
Fix CAPI upgrade tests

fabriziopandini:
This PR updates CAPI upgrade jobs to the changes introduced by https://github.com/kubernetes-sigs/cluster-api/pull/4397 and more specifically
- all the jobs are pulling Kubernetes, so we can build kinder images if those are not published yer
- remove the unnecessary BUILD_NODE_IMAGE_TAG variable

We are also handing the fact that in kubernetes CI there is already the v1.22 release even if 1.21 is not published yet

/assign @vincepri 
@srm09


vincepri (approved):
/approve
/lgtm

PR #21653, opened 2021-04-05
Slack token secret synced from google secret manager

chaodaiG:
https://github.com/kubernetes/test-infra/pull/21610 added a place holder file for kubernetes external secret, and failed prow deployment job as `kubectl apply -f` doesn't like empty file. This PR fixes forward, and by the way make it possible for automatically deploying prow monitoring stack without manual effort.

PR #21675, opened 2021-04-06
Prometheus alertmanager use alertmanagerconfig CR instead of plain text secret

chaodaiG:
Fixes: 21665

This change eliminates the need of manual replacing secret values before deployment

The CRD was already deployed while upgrading prometheus operator. I have tested this on my personal cluster and it seemed working. The only catch is I haven't figured out how to use template yet, but since the template is just a simple form, I think inline could be good(This part isn't tested yet)


chaodaiG:
/cc @cjwagner @hongkailiu @alvaroaleman 

/hold
Might need to delete the secret `alertmanager-prow` before the deployment of this PR

config/prow/cluster/monitoring/BUILD.bazel
```diff
@@ -19,6 +19,7 @@ release(
     component("alertmanager_rbac", MULTI_KIND),
     component("alertmanager_expose", MULTI_KIND),
     component("prow", "alertmanager"),
+    component("alertmanager-prow", "config"),
```
cjwagner, line 22: I think this assumes/requires the contents of `alertmanager-prow_config.yaml` to only be `kind: Config`. I'm not certain that this would cause problems, but either way it would probably be better to specify `alertmanagerconfig` here and change the file name to match.

config/prow/cluster/monitoring/alertmanager-prow_config.yaml
```diff
@@ -0,0 +1,41 @@
...
+          name: 'prometheus-alert-slack-post-prow-alerts-secret-url'
+          key: 'url'
+        iconURL: https://avatars3.githubusercontent.com/u/3380462
+        text: '@test-infra-oncall {{ .CommonAnnotations.message }}'
+        linkNames: true
+    - name: 'slack-alerts'
+      slackConfigs:
+      - channel: '#testing-ops'
+        apiURL:
+          name: 'prometheus-alert-slack-post-testing-ops-secret-url'
```
cjwagner, line 37: Have both of these secrets been created in the `prow-monitoring` namespace?

cjwagner (commented):
> Might need to delete the secret `alertmanager-prow` before the deployment of this PR

That shouldn't be necessary and I think we should wait to do that until we are certain the new config is working properly in case we need to revert.

config/prow/cluster/monitoring/alertmanager-prow_config.yaml
cjwagner: Looks like yes: https://github.com/kubernetes/test-infra/pull/21653

config/prow/cluster/monitoring/BUILD.bazel
chaodaiG: good point, done

config/prow/cluster/monitoring/alertmanager-prow_config.yaml
chaodaiG: yeah they both exist

config/prow/cluster/monitoring/prow_alertmanagerconfig.yaml
```diff
@@ -0,0 +1,41 @@
+# PROW_INSTANCE_SPECIFIC
+# Slack endpoint, or even different methods of alerting
+apiVersion: monitoring.coreos.com/v1alpha1
+kind: AlertmanagerConfig
+metadata:
+  name: alertmanager-prow
```
cjwagner, line 6: Based on the file name this should be "prow". Again, I'm not sure if this actually matters.

cjwagner (approved):

config/prow/cluster/monitoring/prow_alertmanagerconfig.yaml
chaodaiG: afaict the first part of the file name is pretty much used for naming the generated bazel targets as https://github.com/kubernetes/test-infra/blob/985ef5e7217e4bc0d8abbe01e7e4d2ab439bd43e/prow/def.bzl#L158 and https://github.com/bazelbuild/rules_k8s/blob/f1c6399cdd691b7aca90073398e8f690ec8992c6/k8s/object.bzl#L481. Will unhold tomorrow

chaodaiG:
This worked. But prometheus operator generated `alertmanager-prow-generated` secret by merging the output from both `alertmanager-prow` secret and `alertmanagerconfig` CR, deleting the `alertmanager-prow` secret and the problem was gone

PR #21679, opened 2021-04-06
Allow setting team/katacoda label

sftim:
Follows on from https://github.com/kubernetes/test-infra/pull/21652

If this merges then I believe you can 
/label team/katacoda

for repos that have a Katacoda team label.

(Only the [k/website](https://github.com/kubernetes/website) repo actually has such a label)


spiffxp (approved):
/approve
/lgtm

PR #21680, opened 2021-04-06
Grafana secret deployed from google secret manager

chaodaiG:
The secret currently need to be deployed manually if there is an update, make it backed up by google secret manager so that it can be automatically generated


config/prow/cluster/kubernetes_external_secrets.yaml
```diff
@@ -27,3 +26,16 @@ spec:
...
+apiVersion: kubernetes-client.io/v1
+kind: ExternalSecret
+metadata:
+  name: grafana
+  namespace: prow-monitoring
+spec:
+  backendType: gcpSecretsManager
+  projectId: k8s-prow
+  data:
+  - key: gke_k8s-prow_us-central1-f_prow__prow-monitoring__grafana
```
chaodaiG, line 39: This secret already exists in k8s-prow project, value was grabbed from the secret located in cluster

cjwagner (approved):

cjwagner:
/hold
Will this work with the existing secret? I'd expect this to fail if a secret already exists and is not managed by KES.

chaodaiG:
> /hold
> Will this work with the existing secret? I'd expect this to fail if a secret already exists and is not managed by KES.

I have tested this on my dev cluster, and KES just went ahead override existing secret. I think this is WAI, can prevent someone from manually messing up a secret

PR #21695, opened 2021-04-07
E2E test images: Adds cuda-vector-add-old postsubmit job

claudiubelu:
Currently, the only image left in ``gcr.io/kubernetes-e2e-test-images`` is the ``cuda-vector-add:1.0`` image. This adds a postsubmit job for it, so we can then promote it to the ``k8s.gcr.io/e2e-test-images`` registry.

Related: https://github.com/kubernetes/kubernetes/issues/96770
Related: https://github.com/kubernetes/kubernetes/pull/100887


spiffxp (approved):
/approve 
/lgtm

PR #21704, opened 2021-04-07
Leftpad GitHub status description before adding basesha

alvaroaleman:
This leftpads the Github status description before the basesha which
makes the thing look a lot nicer. Roundtripping works for both unpadded
and padded descriptions.

Now:
basesha_padded

vs before:

basesha_unpadded

### DIFF
--- a/prow/config/config.go
+++ b/prow/config/config.go
@@ -2546,6 +2546,16 @@ func ContextDescriptionWithBaseSha(humanReadable, baseSHA string) string {
 	var suffix string
 	if baseSHA != "" {
 		suffix = contextDescriptionBaseSHADelimiter + baseSHA
+		// Leftpad the baseSHA suffix so its shown at a stable position on the right side in the GitHub UI.
+		// The GitHub UI will also trim it on the right side and replace some part of it with '...'. The
+		// API always returns the full string.
+		if len(humanReadable+suffix) < contextDescriptionMaxLen {
+			for i := 0; i < contextDescriptionMaxLen-len(humanReadable+suffix); i++ {
+				// This looks like a standard space but is U+2001, because GitHub seems to deduplicate normal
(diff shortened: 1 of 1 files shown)


prow/config/config.go
```diff
@@ -2546,6 +2546,13 @@ func ContextDescriptionWithBaseSha(humanReadable, baseSHA string) string {
 	var suffix string
 	if baseSHA != "" {
 		suffix = contextDescriptionBaseSHADelimiter + baseSHA
+		if len(humanReadable+suffix) < 140 {
```
chaodaiG, line 2549: 140 is a const at https://github.com/kubernetes/test-infra/pull/21704/files#diff-88e6b42a331a7aa9260f9a830961e3ba3c77e0e6d781465378b510d7abc1517cR2574

chaodaiG (commented):
Awesome!

prow/config/config.go
alvaroaleman: good point, updated. I've also added some godocs to clarify the reason because it might be confusing from just looking at the code.

chaodaiG (approved):
/hold
Since no unit test is broken I would suspect there is no unit test for this function, would be nice to have imho

alvaroaleman:
There are unittests that do property-based testing for the things we care about from a functional perspective: We can [roundtrip](https://github.com/alvaroaleman/test-infra/blob/a00f119df2eac8d5a2570fd0ab698ff8264dc5a6/prow/config/config_test.go#L6079-L6081) and we do [not exceed the 140 char limit](https://github.com/alvaroaleman/test-infra/blob/a00f119df2eac8d5a2570fd0ab698ff8264dc5a6/prow/config/config_test.go#L6075-L6077). This change doesn't break that, so they still pass :)
/hold cancel

PR #21705, opened 2021-04-07
ci-kubernetes-e2e-windows-containerd-gce* is not really testing containerd

dims:
Looks like we never adjusted the job to really use `containerd` after https://github.com/kubernetes/kubernetes/pull/92063/files landed.

a quick look at logs confirms that the job is really using docker.
https://gcsweb.k8s.io/gcs/kubernetes-jenkins/logs/ci-kubernetes-e2e-windows-containerd-gce/1379879151552434176/artifacts/

/cc @jingxu97 @marosset 

### DIFF
--- a/config/jobs/kubernetes/sig-windows/windows-gce.yaml
+++ b/config/jobs/kubernetes/sig-windows/windows-gce.yaml
@@ -18,7 +18,7 @@ presets:
 - labels:
     preset-e2e-gce-windows-containerd: "true"
   env:
-  - name: KUBE_CONTAINER_RUNTIME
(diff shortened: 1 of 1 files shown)


dims:
cc @ibabou @jeremyje @pjh

marosset (approved):
/lgtm
/approve

ibabou:
@dims Thanks for the update! I just came across this today as I hit an issue running a cluster through kube-up with KUBE_WINDOWS_CONTAINER_RUNTIME=containerd as usual (node weren't able to download containerd binaries), but turned we don't pass that by looking at the gce-windows-2019-containerd-1.21's jobs. I was just looking to update and validate.

PR #21718, opened 2021-04-08
Use consistent sig-windows-networking jobs names

ionutbalutoiu:
This is a quality of life change to bring more clarity into the
`sig-windows-networking` Prow jobs history.

The logs and artifacts stored into the previous buckets were
copied into the new locations.

Therefore, the upstream testgrid links for the older jobs will not break.


ionutbalutoiu:
cc: @mkostersitz @marosset @adelina-t

PR #21719, opened 2021-04-08
test-infra: add golangci-lint Job

sbueringer:
Bazel-based implementation of golangci-lint

Hope this helps to resolve the deadlock from: https://github.com/kubernetes/test-infra/pull/19589#discussion_r507889662

xref: 
* https://github.com/kubernetes/test-infra/pull/19589
* https://github.com/kubernetes/test-infra/issues/19588

Notes:
* idea is to start with an additional optional job for now and make it mandatory and add additional linter over time
* do we want to use the same golangci-lint version as in repo infra:
  * let's use the same version for now. 1.34.1 is reasonable up-to-date and we can always move away from repo-infra in case we really have to 
* do we want to run golangci-lint in a separate job
  * I tested to run it with the regular bazel job and I always hit a 10m timeout. Even if we adjust the timeout the bazel Job would get significantly slower

I'm not sure how to get the tests green as we basically first need the changes to the `pull-test-infra-bazel ` job. Should I open a PR to exclude the golangci-linter first so this PR get's green before merge, wdyt?

### DIFF
--- a/config/jobs/kubernetes/test-infra/test-infra-presubmits.yaml
+++ b/config/jobs/kubernetes/test-infra/test-infra-presubmits.yaml
@@ -17,13 +17,41 @@ presubmits:
         - test
         - --config=ci
         - --nobuild_tests_only
+        - --
         - //...
+        - -//hack:verify-golangci-lint
         env:
         - name: BAZEL_FETCH_PLEASE
           value: //...
     annotations:
       testgrid-dashboards: presubmits-test-infra
       testgrid-tab-name: bazel
+
+  - name: pull-test-infra-golangci-lint
+    branches:
+    - master
+    always_run: true
+    optional: true
+    decorate: true
+    labels:
+      preset-service-account: "true"
+      preset-bazel-scratch-dir: "true"
+    spec:
+      containers:
+      - image: gcr.io/k8s-testimages/launcher.gcr.io/google/bazel:v20210204-b06ec78-test-infra
+        command:
+        - hack/bazel.sh
+        args:
+        - test
+        - --config=ci
+        - --nobuild_tests_only
+        - //hack:verify-golangci-lint
+        env:
+        - name: BAZEL_FETCH_PLEASE
+          value: //...
+    annotations:
+      testgrid-dashboards: presubmits-test-infra
+      testgrid-tab-name: golangci-lint
 
   - name: pull-test-infra-gubernator
     branches:
--- a/hack/BUILD.bazel
+++ b/hack/BUILD.bazel
@@ -223,6 +223,13 @@ test_suite(
 )
 
 test_suite(
+    name = "verify-golangci-lint",
+    tests = [
+        "@io_k8s_repo_infra//hack:verify-golangci-lint",
+    ],
+)
+
+test_suite(
     name = "verify-linters",
     tags = ["lint"],  # picks up all non-manual targets with this tag
 )
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,0 +1,8 @@
+run:
+  deadline: 10m
+
+linters:
+  disable-all: true
+  enable:
+  - gofmt
+  - misspell
--- a/hack/verify-golangci-lint.sh
+++ b/hack/verify-golangci-lint.sh
@@ -1,0 +1,26 @@
+#!/usr/bin/env bash
+# Copyright 2021 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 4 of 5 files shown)


config/jobs/kubernetes/test-infra/test-infra-presubmits.yaml
```diff
@@ -18,13 +18,40 @@ presubmits:
         - --config=ci
         - --nobuild_tests_only
         - //...
+        - -//hack:verify-golangci-lint
```
alvaroaleman, line 21: I think this needs to be a distinct PR, otherwise the presubmit won't pass here, right? Will it fail if that target doesn't exist yet?

config/jobs/kubernetes/test-infra/test-infra-presubmits.yaml
```diff
@@ -18,13 +18,40 @@ presubmits:
...
+    spec:
+      containers:
+      - image: gcr.io/k8s-testimages/launcher.gcr.io/google/bazel:v20210204-b06ec78-test-infra
+        command:
+        - hack/bazel.sh
+        args:
+        - test
+        - --config=ci
+        - --nobuild_tests_only
+        - //hack:verify-golangci-lint
```
alvaroaleman, line 47: Maybe also put this into a distinct PR so we can check here if everything works?

config/jobs/kubernetes/test-infra/test-infra-presubmits.yaml
sbueringer: Just tested it. I can ignore whole folders but not a single non-existing target. I could open a PR which changes this jobs and adds `verify-golangci-lint` with an empty test-suite. I guess that would work

config/jobs/kubernetes/test-infra/test-infra-presubmits.yaml
sbueringer: So do you mean 2 PRs:
* make changes to `pull-test-infra-bazel` and add an empty `verify-golangci-lint` test-suite
* add the optional Job and the implementation of `verify-golangci-lint`

Or to split this into 3 PRs?

config/jobs/kubernetes/test-infra/test-infra-presubmits.yaml
alvaroaleman: yeah that sounds good, too

config/jobs/kubernetes/test-infra/test-infra-presubmits.yaml
alvaroaleman: One PR to add the exclude, an empty target and the new job as optional, and then verify everything on this PR

sbueringer: Got it, So my version + also add the new optional Job in the first PR (just to confirm)

alvaroaleman: yeah, exactly

fejta:
Ideally we could just get `bazel test @io_k8s_repo_infra//hack:verify-golangci-lint` to work in this repo.

alvaroaleman:
> Ideally we could just get bazel test @io_k8s_repo_infra//hack:verify-golangci-lint to work in this repo.

Isn't that exactly what this PR is doing?

hack/BUILD.bazel
```diff
@@ -222,6 +222,13 @@ test_suite(
     ],
 )
 
+test_suite(
+    name = "verify-golangci-lint",
+    tests = [
+        "@io_k8s_repo_infra//hack:verify-golangci-lint",
```
fejta, line 228: Please just add this to the `verify-all` suite

fejta:
> Isn't that exactly what this PR is doing?

No it seems to do all kinds of unnecessary stuff

fejta:
https://github.com/kubernetes/test-infra/pull/21751 this passed on my workstation, maybe it will also work in CI

fejta:
https://docs.bazel.build/versions/master/be/common-definitions.html#common-attributes-tests suspect the timeout is getting enforced at the bazel level

fejta:
Will send you two a repo-infra PR that increases the timeout to 15m in a second.

sbueringer:
> > Isn't that exactly what this PR is doing?
> 
> No it seems to do all kinds of unnecessary stuff

Most of what this PR did was based on the assumptions:
* we don't want to slow down the bazel job that much
* we want to add a hack/verify script like for (looks to me) everything else:
image


But anyway. Looks like the linter now works on the main branch and as that was also the goal of these PRs I'll close them.

/close

PR #21732, opened 2021-04-09
test-infra: add (no-op) golangci-lint job

sbueringer:
xref: https://github.com/kubernetes/test-infra/pull/21719#discussion_r609993039

As discussed. I would change it to `always_run` true in the 2nd PR.

I tested it locally by executing:
```
./hack/bazel.sh test --nobuild_tests_only -- //... -//hack:verify-golangci-lint
./hack/bazel.sh test --nobuild_tests_only //hack:verify-golangci-lint
```

### DIFF
--- a/config/jobs/kubernetes/test-infra/test-infra-presubmits.yaml
+++ b/config/jobs/kubernetes/test-infra/test-infra-presubmits.yaml
@@ -17,13 +17,41 @@ presubmits:
         - test
         - --config=ci
         - --nobuild_tests_only
+        - --
         - //...
+        - -//hack:verify-golangci-lint
         env:
         - name: BAZEL_FETCH_PLEASE
           value: //...
     annotations:
       testgrid-dashboards: presubmits-test-infra
       testgrid-tab-name: bazel
+
+  - name: pull-test-infra-golangci-lint
+    branches:
+    - master
+    always_run: false
+    optional: true
+    decorate: true
+    labels:
+      preset-service-account: "true"
+      preset-bazel-scratch-dir: "true"
+    spec:
+      containers:
+      - image: gcr.io/k8s-testimages/launcher.gcr.io/google/bazel:v20210204-b06ec78-test-infra
+        command:
+        - hack/bazel.sh
+        args:
+        - test
+        - --config=ci
+        - --nobuild_tests_only
+        - //hack:verify-golangci-lint
+        env:
+        - name: BAZEL_FETCH_PLEASE
+          value: //...
+    annotations:
+      testgrid-dashboards: presubmits-test-infra
+      testgrid-tab-name: golangci-lint
 
   - name: pull-test-infra-gubernator
     branches:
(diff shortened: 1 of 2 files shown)


alvaroaleman (approved):
Looks good, but I do not have approval powers here.
/assign @chaodaiG

chaodaiG (commented):
This looks good to me, but I know @fejta cares about newly added presubmit jobs, so
/assign @fejta

alvaroaleman:
Just to expand on the reasoning for making this a distinct job: This takes about four minutes and a half minutes  and shares nothing with the standard bazel job. Adding it into the standard bazel job thus would make it about four minutes slower (and right now that just makes it fail, because it seems to have an internal 10m timeout that @sbueringer couldn't figure out)

fejta:
> Adding it into the standard bazel job thus would make it about four minutes slower

That makes zero sense. Why would it do that?

fejta:
Having 17,000 contexts per presubmit is both wasteful and impossible for an engineer to reproduce off github/prow. This should be part of the standard `bazel test //...` suite.

alvaroaleman:
> That makes zero sense. Why would it do that?

Because it takes four minutes to run the linter

fejta:
If you click the `pull-test-infra-bazel` link, you'll notice it already takes 8m. Therefore I would expect adding a 4m test to the suite to add ~0 additional seconds.

alvaroaleman:
> If you click the pull-test-infra-bazel link, you'll notice it already takes 8m. Therefore I would expect adding a 4m test to the suite to add ~0 additional seconds.

If you check reality, you will notice it disagrees: https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/test-infra/21719/pull-test-infra-bazel/1380224499546329088#1:build-log.txt%3A6185

fejta:
I don't think that's measure what you think it is

9m without: https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/test-infra/21726/pull-test-infra-bazel/1380368015446511616

4m with: https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/test-infra/21719/pull-test-infra-bazel/1380206596231008256

alvaroaleman:
@fejta you seem to have an outdated link. The latest run on https://github.com/kubernetes/test-infra/pull/21719 failed with a bazel timeout error.

I don't really care how many jobs we have unless having few jobs means waiting longer. If this can be solved without an additional job, please advice in https://github.com/kubernetes/test-infra/pull/21719. @sbueringer didn't manage to get past the linked timeout error

fejta:
This means that we will try to do up to 500 things in parallel: https://github.com/kubernetes/test-infra/blob/47f708efabda0eed3e0da8900bb5358720697ea9/.bazelrc#L24

So I would be somewhat surprised if this is causing things to take meaningfully longer.

sbueringer:
I'll close the PR as it's not necessary anymore.
/close

PR #21733, opened 2021-04-09
cluster-api-provider-openstack: send stale test alerts after 48h

sbueringer:
After we added the alerts yesterday there where already alerts today. Problem is that we run the tests every 12h hours and we get the first alert after 12h without new results.

xref: https://kubernetes.slack.com/archives/C8TSNPY4T/p1617911492444300


sbueringer:
/assign @jichenjc 

Quick approve/lgtm would be appreciatd to stop the alert spam :)

sbueringer:
Wow that was quick, thank you :)

PR #21734, opened 2021-04-09
sig-storage: grant team permission to trigger jobs

pohly:
We occassionally need to rerun an image build job for a tagged release
because QEMU sometimes is flaky. By using a team defined in k8s.io/org
we only need to change a single line when adding other people.


pohly:
The corresponding org change is in https://github.com/kubernetes/org/pull/2626

It should be merged first.

pohly:
@spiffxp: the team exists. Can you check this PR? I hope I understood your example correctly.

fejta (approved):

PR #21735, opened 2021-04-09
Soak time for containerd 1.5.0 using release candidate rc.0

dims:
Note that the matching runc version is now documented in a file here:
https://github.com/containerd/containerd/blob/v1.5.0-rc.0/script/setup/runc-version


aojea:
this is the runc with the required rootless changes, right?
/lgtm

aojea:
answering myself , is the `apply caps: operation not permitted"`  thing

dims:
@aojea yep indeed.

PR #21743, opened 2021-04-09
Switch image-builder packer-validate to script

codenrhoden:
The packer-validate test can now be run through a script, simlifying the
Prow definition and making it changeable without modifying the Prow job
definition.

/assign @CecileRobertMichon @detiber

### DIFF
--- a/config/jobs/kubernetes-sigs/image-builder/image-builder-presubmits.yaml
+++ b/config/jobs/kubernetes-sigs/image-builder/image-builder-presubmits.yaml
@@ -76,23 +76,9 @@ presubmits:
     spec:
       containers:
         - image: gcr.io/k8s-testimages/kubekins-e2e:v20210403-e49d2c6-master
-          command:
-          - make
           args:
-          - -C
-          - ./images/capi
-          - validate-all
-          env:
-          - name: AZURE_LOCATION
-            value: "fake"
-          - name: RESOURCE_GROUP_NAME
-            value: "fake"
-          - name: STORAGE_ACCOUNT_NAME
-            value: "fake"
-          - name: DIGITALOCEAN_ACCESS_TOKEN
-            value: "fake"
-          - name: GCP_PROJECT_ID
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes-sigs/image-builder/image-builder-presubmits.yaml
```diff
@@ -76,23 +76,9 @@ presubmits:
...
-          - name: RESOURCE_GROUP_NAME
-            value: "fake"
-          - name: STORAGE_ACCOUNT_NAME
-            value: "fake"
-          - name: DIGITALOCEAN_ACCESS_TOKEN
-            value: "fake"
-          - name: GCP_PROJECT_ID
-            value: "fake"
+          - runner.sh
+          - "./images/capi/ci/packer-validate.sh"
```
CecileRobertMichon, line 81: nit: what do you think about `./images/capi/scripts/ci-packer-validate.sh` instead?

https://github.com/golang-standards/project-layout#scripts

codenrhoden: I'm fine with that. It's been a while since I've read through that link. 

My only complaint there is that I think if a project has a `scripts` directory, I get confused between what goes in that and what goes in `hack`, the common dir for this stuff within K8s projects.

CecileRobertMichon: >  I get confused between what goes in that and what goes in hack, the common dir for this stuff within K8s projects.

true, I think k/k is using `hack` as an all-purpose dev scripts directory. Some projects like CAPI have `hack` for dev scripts https://github.com/kubernetes-sigs/cluster-api/tree/master/hack and `scripts` for ci scripts. I don't know if there is a `hack` convention defined anywhere...

codenrhoden:
Updated as requested, @CecileRobertMichon

PR #21745, opened 2021-04-09
Onboarding prow build cluster: using gsm to stop sharing plain text secret

chaodaiG:
Also make the script smarter by helping creating PR

Note: previous implementation was done by kustomize, which unfortunately was not able to preserve boilerplate or inline comments, thus not suitable for k8s prow, since both exist, so fallback to use less strict `sed` replacing, luckily it works, at least for now..

### DIFF
--- a/prow/create-build-cluster.sh
+++ b/prow/create-build-cluster.sh
@@ -25,6 +25,11 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
+ROOT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd -P)"
+
+readonly PROW_SECRET_ACCESSOR_SA="kubernetes-external-secrets-sa@k8s-prow.iam.gserviceaccount.com"
+readonly APPS_CONSUME_KUBECONFIG="crier deck hook pipeline prow_controller_manager sinker"
+
 # Specific to Prow instance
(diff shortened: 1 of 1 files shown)


fejta (approved):
/hold

matthyx:
OMG, hate `sed`...
Do you need a LGTM?
/lgtm
/hold

matthyx:
> /hold

oops, didn't refresh, sorry for the duplicate

prow/create-build-cluster.sh
```diff
@@ -123,28 +139,125 @@ function createUploadSASecret() {
...
+apiVersion: kubernetes-client.io/v1
+kind: ExternalSecret
+metadata:
+  name: kubeconfig-build-${TEAM}
+  namespace: default
+spec:
+  backendType: gcpSecretsManager
+  projectId: ${PROJECT}
+  data:
+  - key: ${gsm_secret_name}
```
cjwagner, line 219: This is a local variable of the caller so this won't be set, right?

chaodaiG: good catch! Didn't know how it worked in my local test, let me fix it

chaodaiG: https://github.com/kubernetes/test-infra/pull/21770

PR #21749, opened 2021-04-09
Move sig-windows presubmits for easier maintaince

jsturtevant:
Editing the main file was cumbersome, having less in the file should make it easier.

/assign @chewong


chewong (approved):
/lgtm

PR #21751, opened 2021-04-09
Turn on golangci-lint

fejta:
does this work

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,0 +1,24 @@
+run:
+  deadline: 13m
+  skip-dirs:
+  - .git
+  - .tool
+  - vendor
+  - verify
+
+linters-settings:
+  dupl:
+    threshold: 100
+  gocyclo:
+    min-complexity: 50
+
+linters:
+  enable:
+  - govet
+  disable-all: true
+
+issues:
+  exclude-rules:
+  - linters:
+    - golint
+    text: ".*should not use dot dot imports"
--- a/hack/BUILD.bazel
+++ b/hack/BUILD.bazel
@@ -214,11 +214,7 @@ test_suite(
     name = "verify-all",
     tests = [
         ":verify-linters",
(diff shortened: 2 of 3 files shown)


.golangci.yml
```diff
@@ -0,0 +1,28 @@
...
+
+linters-settings:
+  dupl:
+    threshold: 100
+  gocyclo:
+    min-complexity: 50
+
+linters:
+  enable:
+  - govet
```
alvaroaleman, line 18: The only one of the linters you enabled that does not fail in this repo is govet

fejta: Are you sure? This set seems to pass (at least locally).

I just copied the file that is in repo-infra, don't really care about these values atm, just trying to get things working.

alvaroaleman: Yes, if it passes, the bazel target doesn't work: Checking out your PR and running
```
golangci-lint run ./... 2>&1|c
```

gives  these https://gist.github.com/alvaroaleman/6ef68adef5aaa01633467c8eb8120772

fejta: Hmm, what is golangci-lint's exit code? 

The implementation doesn't check output, just relies on the exit code to determine pass/fail: https://github.com/kubernetes/repo-infra/blob/ee974c891e8162941a25135e6c90575fbd35c93a/hack/verify-golangci-lint.sh#L49

fejta: Hmm, exit code is 1

alvaroaleman:
```
$ golangci-lint run ./... &>/dev/null; echo $?
1
```

fejta: Good news! It failed

fejta: https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/test-infra/21751/pull-test-infra-bazel/1380673894922326016

alvaroaleman (approved):
It does take significantly longer now btw, from ~7 mins to ~11 minutes.

PR #21752, opened 2021-04-10
Require GOSS populate job for image-builder

codenrhoden:
Only if GOSS, config, or either scripts involved have changed.

/hold
^ until https://github.com/kubernetes-sigs/image-builder/pull/585 is merged

### DIFF
--- a/config/jobs/kubernetes-sigs/image-builder/image-builder-presubmits.yaml
+++ b/config/jobs/kubernetes-sigs/image-builder/image-builder-presubmits.yaml
@@ -111,8 +111,7 @@ presubmits:
       testgrid-tab-name: pr-pull-image-builder-gcp-all
   - name: pull-goss-populate
     path_alias: sigs.k8s.io/image-builder
-    always_run: false
-    optional: true
+    run_if_changed: 'images/capi/packer/config/.*|images/capi/packer/goss/.*|images/capi/scripts/ci-goss-populate\.sh|images/capi/hack/generate-goss-specs\.py'
     decorate: true
     decoration_config:
       timeout: 20m


config/jobs/kubernetes-sigs/image-builder/image-builder-presubmits.yaml
```diff
@@ -111,8 +111,7 @@ presubmits:
       testgrid-tab-name: pr-pull-image-builder-gcp-all
   - name: pull-goss-populate
     path_alias: sigs.k8s.io/image-builder
-    always_run: false
-    optional: true
+    run_if_changed: 'images/capi/packer/config/.*|images/capi/packer/goss/.*|images/capi/scripts/ci-goss-populate\.sh|images/capi/hack/generate-goss-specs\.py'
```
CecileRobertMichon, line 114: How are planning to use this job? is it meant to be validation or generation or both? If it's validation/both, shouldn't it error if there is a diff? I might be missing something but I don't see what part of the script makes the presubmit fail if a diff was generated

codenrhoden: It doesn't generate a diff, because we don't check the final GOSS tests into the repo (they are dynamically generated based on variables like the K8s version, containerd version, etc). The files in the repo are basically templates, and there are for loops that iterate over values in `goss-vars.yaml`. It's entirely possible that typos or missing data causes the dynamic generation of these tests to fail, and that's what this new set of tests catch. It basically checks if we *can* generate the tests without failure.

CecileRobertMichon: makes sense, thanks for the explanation!

PR #21755, opened 2021-04-10
Revert "Soak time for containerd 1.5.0 using release candidate rc.0"

dims:
This reverts commit f3ab2eb09b41ee3571f1391512feda65109a2dae.

looks like the older v1 config.toml breaks newer containerd.


aojea:
will keep it in hold to see if I can see the root cause
/hold
sooner or later we have to go through this

PR #21758, opened 2021-04-11
Remove dead code and enable linting for it

alvaroaleman:
The linter only ever considers unexported code to be possibly dead, so this should not break any third-party consumers.


chaodaiG (approved):
Awesome!
/hold

PR #21759, opened 2021-04-11
Remove plank v1

alvaroaleman:
This change removes plank v1, which has been deprecated since October
2020 and was originally slated for removal in February 2021.

Users are advised to instead use the prow-controller-manager.


prow/plank/reconciler.go
```diff
@@ -52,6 +52,11 @@ import (
 
 const ControllerName = "plank"
 
+// PodStatus constants
```
alvaroaleman, line 55: The additions here are copies of code that was in controller.go but also used by the code in reconciler.go

petr-muller (approved):

PR #21760, opened 2021-04-11
Horologium: Use a cached prowjob client

alvaroaleman:
This change makes Horologium use a cached prowjob client instead of
doing a List every minute against the API. This significantly
reduces the load both on the kube api and on horologium, because they
don't have to (de-) serialize a couple of hundred MB of json every
Minute.

### DIFF
--- a/prow/cmd/horologium/main.go
+++ b/prow/cmd/horologium/main.go
@@ -25,9 +25,9 @@ import (
 	"time"
 
 	"github.com/sirupsen/logrus"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/sets"
+	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/cluster"
 
 	prowapi "k8s.io/test-infra/prow/apis/prowjobs/v1"
 	"k8s.io/test-infra/prow/config"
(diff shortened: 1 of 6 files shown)


prow/cmd/horologium/main.go
```diff
@@ -93,9 +93,27 @@ func main() {
...
+	cfg, err := o.kubernetes.InfrastructureClusterConfig(o.dryRun)
 	if err != nil {
-		logrus.WithError(err).Fatal("Error getting Kubernetes client.")
+		logrus.WithError(err).Fatal("Failed to get prowjob kubeconfig")
+	}
+	cluster, err := cluster.New(cfg, func(o *cluster.Options) { o.Namespace = configAgent.Config().ProwJobNamespace })
+	if err != nil {
+		logrus.WithError(err).Fatal("failed to construct prowjob client")
+	}
+	if _, err := cluster.GetCache().GetInformer(interrupts.Context(), &prowapi.ProwJob{}); err != nil {
```
chaodaiG, line 104: Is this just for sanity check? Or `GetInformer` actually does something like `SetInformer` in the `Cache`? Would be nice to document it imo

alvaroaleman: Added some godocs

chaodaiG (approved):
/hold
unhold when ready

I think horologium should be easily tested with integration test

chaodaiG:
> I think horologium should be easily tested with integration test

I meant I'll give it a stab some time if no one beats me to it

alvaroaleman:
I had to rebase due to a conflict in ANNOUNCEMENTS.md, ptal

chaodaiG (approved):

PR #21762, opened 2021-04-12
labels: refine release-note plugin name

hi-rustin:
It should be release-note instead of releasenote.


label_sync/labels.md
```diff
@@ -157,16 +157,16 @@ larger set of contributors to apply/remove them.
...
 | <a id="do-not-merge/invalid-owners-file" href="#do-not-merge/invalid-owners-file">`do-not-merge/invalid-owners-file`</a> | Indicates that a PR should not merge because it has an invalid OWNERS file in it.| prow |  [verify-owners](https://git.k8s.io/test-infra/prow/plugins/verify-owners) |
-| <a id="do-not-merge/release-note-label-needed" href="#do-not-merge/release-note-label-needed">`do-not-merge/release-note-label-needed`</a> | Indicates that a PR should not merge because it's missing one of the release note labels. <br><br> This was previously `release-note-label-needed`, | prow |  [releasenote](https://git.k8s.io/test-infra/prow/plugins/releasenote) |
+| <a id="do-not-merge/release-note-label-needed" href="#do-not-merge/release-note-label-needed">`do-not-merge/release-note-label-needed`</a> | Indicates that a PR should not merge because it's missing one of the release note labels. <br><br> This was previously `release-note-label-needed`, | prow |  [release-note](https://git.k8s.io/test-infra/prow/plugins/release-note) |
```
fejta, line 160: The path is `releasenote`, not release-note

The PluginName is `release-note`

https://github.com/kubernetes/test-infra/blob/cc94620ed96811c9f698e16f7d539608c6544448/prow/plugins/releasenote/releasenote.go#L36

fejta (commented):
The package name is `releasenote`

label_sync/labels.md
hi-rustin: So the path is used here as expected? Or do we need to rename the path?

hi-rustin:
/close

If prowPlugin is the package name, then I think it's fine. If it's a plugin name then we need to change it. If it's the latter, feel free to /reopen the PR.

Issue #21633, opened 2021-04-02
1.21 release branch is missing presubmits


BenTheElder:
https://github.com/kubernetes/test-infra/issues/21606#issuecomment-812094062

sorry I don't have time to dig into this, I need to wrap up some other things and then I'm OOO for a week, but I want to make sure this is tracked.

/sig release
/area jobs
/priority critical-urgent

justaugustus:
@kubernetes/release-managers -- Someone please volunteer to investigate, assign yourself, and treat this as your highest (SIG Release) priority this week.

cpanato:
I will work on that
/assign

cpanato:
jobs are running now 
see a PR for example: https://github.com/kubernetes/kubernetes/pull/101385

cpanato:
i guess we can close it now, if it is missing more jobs we can reopen or create a new issue
/close

puerco:
Awesome job, thanks @cpanato  :D :tada:

BenTheElder:
thank you!!

PR #21764, opened 2021-04-12
update release configuration jobs with missing jobs

cpanato:
Add `fork-per-release: "true"` for the following jobs:
- `pull-kubernetes-unit`
- `pull-kubernetes-conformance-kind-ga-only-parallel`
- `pull-kubernetes-conformance-kind-ipv6-parallel`
- `pull-kubernetes-e2e-kind-ipv6`
- `pull-kubernetes-verify-govet-levee`

and after that ran

```
bazel run //releng/config-forker -- \
  --job-config $(pwd)/config/jobs \
  --version 1.21 \
  --output $(pwd)/config/jobs/kubernetes/sig-release/release-branch-jobs/1.21.yaml
```
and edit to add only the missing jobs and not update other jobs

Part of https://github.com/kubernetes/test-infra/issues/21633

/assign @justaugustus @BenTheElder @saschagrunert 
cc @kubernetes/release-managers

### DIFF
--- a/config/jobs/kubernetes/sig-instrumentation/verify-govet-levee.yaml
+++ b/config/jobs/kubernetes/sig-instrumentation/verify-govet-levee.yaml
@@ -10,6 +10,7 @@ presubmits:
     skip_report: false
     path_alias: k8s.io/kubernetes
     annotations:
+      fork-per-release: "true"
       testgrid-create-test-group: "true"
     labels:
       preset-service-account: "true"
--- a/config/jobs/kubernetes/sig-release/release-branch-jobs/1.21.yaml
+++ b/config/jobs/kubernetes/sig-release/release-branch-jobs/1.21.yaml
@@ -1706,3 +1706,183 @@ presubmits:
             memory: 12Gi
         securityContext:
           privileged: true
+  - always_run: true
+    branches:
+    - release-1.21
+    cluster: k8s-infra-prow-build
+    decorate: true
+    labels:
+      preset-dind-enabled: "true"
+      preset-service-account: "true"
+    name: pull-kubernetes-verify-govet-levee
+    path_alias: k8s.io/kubernetes
+    spec:
+      containers:
+      - args:
+        - verify
+        command:
+        - make
+        env:
+        - name: WHAT
+          value: govet-levee
+        - name: EXCLUDE_TYPECHECK
+          value: "y"
+        - name: EXCLUDE_FILES_REMAKE
+          value: "y"
+        - name: EXCLUDE_GODEP
+          value: "y"
+        - name: KUBE_VERIFY_GIT_BRANCH
+          value: release-1.21
+        - name: REPO_DIR
+          value: /workspace/k8s.io/kubernetes
+        image: gcr.io/k8s-testimages/kubekins-e2e:v20210412-176e4b6-1.21
+        imagePullPolicy: IfNotPresent
+        name: ""
+        resources:
+          limits:
+            cpu: "7"
+            memory: 12Gi
+          requests:
+            cpu: "7"
+            memory: 12Gi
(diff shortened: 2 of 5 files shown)


config/jobs/kubernetes/sig-instrumentation/verify-govet-levee.yaml
```diff
@@ -5,7 +5,7 @@ presubmits:
     always_run: true
     decorate: true
     skip_branches:
-      - release-\d+.\d+ # per-release job
+    - release-\d+.(1[1-9]|20) # per-release job, skip release < 1.20, for  > 1.21 this will run + main branch
```
saschagrunert, line 8: What is the reason to allow only 1.21 and not the other maintained release branches?

cpanato: It was skipping all release branched before, which I don't know the reason, but Ben said (https://kubernetes.slack.com/archives/CJH2GBF7Y/p1617470563282300?thread_ts=1617394767.280300&cid=CJH2GBF7Y) because of bazel.

Then applied this rule to block the job to run when the release branch is 1.20 or older and allow to 1.21 and newer like 1.22.... including the main branch

Not sure if this applies to all jobs in this PR or just to a few.

should we make a canary job for those and try out in PRs for each release branch?

saschagrunert: No I think that's fine. 

Let's add those reasons for skipping older release branches as a yaml comment.

BenTheElder: This is incorrect. All jobs with this rule should be forked as they use a per-release branched image with different go version etc.

BenTheElder: Line 19 below is why this has the "per release job" comment and it should have an instance per release as well as the non specific instance for master/main/feature branches/

cpanato: I think i understood, will update those tomorrow.
Thbaks for your feedback

cpanato:
it change a lot of things, I will redo and push one job only at time to check

config/jobs/kubernetes/sig-release/release-branch-jobs/1.18.yaml
```diff
@@ -120,12 +73,59 @@ periodics:
           cpu: "1"
           memory: 3Gi
 - annotations:
-    fork-per-release-periodic-interval: ""
+    fork-per-release-periodic-interval: 2h 6h 24h
+    testgrid-alert-email: bentheelder@google.com,siarkowicz@google.com
+    testgrid-dashboards: sig-instrumentation-tests, sig-testing-kind, sig-release-job-config-errors
+    testgrid-num-columns-recent: "6"
+    testgrid-tab-name: kind-json-logging-1.18
```
BenTheElder, line 80: this job should not exist for < 1.21

BenTheElder:
we should not be modifying the old release branches, just 1.21 which missed jobs prior to release

e.g. the structured logging and unit test jobs are only for 1.21+, in some previous release we didn't have the relevant kubelet flags and we still use bazel for unit tests on older releases (we do not modify builds for older branches except where necessary).

cpanato:
@BenTheElder @justaugustus Let me know if this is correct now

cpanato:
@BenTheElder gently ping

BenTheElder (approved):
pretty sure this is right 🤞 

sorry, I've been intermittently unavailable this week (e.g. vaccine -> fatigue ...)

glad to finally get this in 🔥

PR #21765, opened 2021-04-12
enable storage-bucket validation by default

### LINKED ISSUE #19366: ~Jan 2021: enable storage-bucket validation by default
Storage bucket validation was introduced in #18001. It is currently opt-in only (via deck configuration).

Validation should be enabled by default -- but no sooner than ~Jan 2021. (This grace period allows Prow instance maintainers time to inspect logs and configure additional allowed storage buckets.)

matthyx:
fix #19366 
/cc @e-blackwelder 
/assign @alvaroaleman

### DIFF
--- a/prow/ANNOUNCEMENTS.md
+++ b/prow/ANNOUNCEMENTS.md
@@ -3,6 +3,8 @@
 ## New features
 
 New features added to each component:
+  - *April 12th, 2021* End of grace period for storage bucket validation, additional buckets have to be allowed
+    by adding them to the `deck.additional_allowed_buckets` list.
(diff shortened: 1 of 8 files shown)


alvaroaleman:
You also need to update `ShouldValidateStorageBuckets` (and maybe unexport it? its not used outside of the config package from what I can see)

matthyx:
I see *a lot* of other expired TODOs in that file and thorough prow code base... maybe I'll do a bunch Spring cleaning PRs when I'm bored.

alvaroaleman (approved):

PR #21771, opened 2021-04-12
Add config path location for using external secrets

chaodaiG:
Also by the way fixing an error: the cluster secret will be deleted if external secret CR was deleted

### DIFF
--- a/prow/prow_secrets.md
+++ b/prow/prow_secrets.md
@@ -40,11 +40,13 @@ where the prow service/build clusters are located.
    ```
    The above command ensures that the service account used by prow can only
    access the secret name `<my-gsm-secret-name>` in the GCP project owned by
-   clients. The service account used for prow.k8s.io is defined in
+   clients. The service account used for prow.k8s.io (aka `test-infra-trusted`
+   build cluster) is defined in
    [`trusted_serviceaccounts.yaml`](https://github.com/kubernetes/test-infra/blob/1b2153ebe2809727a45c5b930647b2a3609dd7e7/config/prow/cluster/trusted_serviceaccounts.yaml#L46),
    and the secrets are defined in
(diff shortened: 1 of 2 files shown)


chaodaiG:
/cc @cjwagner 
CC @aramase @tam7t

prow/prow_secrets.md
```diff
@@ -41,7 +41,13 @@ where the prow service/build clusters are located.
    The above command ensures that the service account used by prow can only
    access the secret name `<my-gsm-secret-name>` in the GCP project owned by
    clients. The service account used for prow.k8s.io is defined in
```
cjwagner, line 43: Please clarify that this first pair of serviceaccount+externalsecret is for the service cluster and anything running in the `test-infra-trusted` build cluster. It would also be good to clarify that the `k8s-prow-builds` cluster is the default build cluster.

prow/prow_secrets.md
```diff
@@ -75,6 +81,5 @@ data:
 ```
 
 The `Secret` will be updated automatically when the secret value in gsm changed
-or the `ExternalSecret` is changed. The secret will not be deleted by kubernetes
-external secret, even in the case of `ExternalSecret` CR is deleted from the
-cluster. Deletion of `Secret` will need to be handled separately if desired.
+or the `ExternalSecret` is changed. when `ExternalSecret` CR is deleted from the
+cluster, the secret will be also be deleted by kubernetes external secret.
```
cjwagner, line 85: Note that deleting the ExternalSecret from git source does not cause our deployment automation to delete the resource in cluster, so this still requires manual intervention.

cjwagner (commented):
We should reference this doc from the job config cookbook: https://github.com/kubernetes/test-infra/blob/master/config/jobs/README.md

prow/prow_secrets.md
chaodaiG: done

prow/prow_secrets.md
chaodaiG: done

cjwagner (approved):

PR #21772, opened 2021-04-12
Needs rebase: Don't always require github token

alvaroaleman:
When using apps auth, the token isn't needed.

I've manually deployed this with apps auth to check if it works now, which it does.


hongkailiu (commented):
/lgtm

PR #21773, opened 2021-04-12
Enable misspell and ineffassign linters

alvaroaleman:
And fix the issues they found.

### DIFF
--- a/def/configmap/main.go
+++ b/def/configmap/main.go
@@ -26,7 +26,7 @@ import (
 	"os"
 	"strings"
 
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	"sigs.k8s.io/yaml"
 )
 
@@ -60,11 +60,7 @@ func (mkv *multiKeyValue) Set(v string) error {
 	if len(p) != 2 {
 		return fmt.Errorf("%s does not match label=value", v)
 	}
-	if mkv == nil {
-		mkv = &multiKeyValue{
-			p[0]: p[1],
-		}
-	} else {
+	if mkv != nil {
 		(*mkv)[p[0]] = p[1]
 	}
 	return nil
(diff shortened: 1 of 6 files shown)


def/configmap/main.go
```diff
@@ -60,11 +60,7 @@ func (mkv *multiKeyValue) Set(v string) error {
 	if len(p) != 2 {
 		return fmt.Errorf("%s does not match label=value", v)
 	}
-	if mkv == nil {
```
alvaroaleman, line 63: The `mkv` pointer is scoped to the method, so this does nothing outside of the method

prow/cmd/generic-autobumper/bumper/bumper.go
```diff
@@ -393,9 +393,9 @@ func Run(o *Options) error {
 		// If failed to push because a closed PR already exists with this change ID (the PR was abandoned). Hash the ID again and try one more time.
 		if err != nil && strings.Contains(err.Error(), "failed to push some refs") && strings.Contains(err.Error(), "closed") {
 			logrus.Warn("Error pushing CR due to already used ChangeID. PR may have been abandoned. Trying again with new ChangeID.")
-			changeId, err := getChangeId(o.Gerrit.Author, o.Gerrit.AutobumpPRIdentifier, changeId)
-			if err != nil {
-				return err
+			changeId, subErr := getChangeId(o.Gerrit.Author, o.Gerrit.AutobumpPRIdentifier, changeId)
```
alvaroaleman, line 396: A couple of lines below is a statement that wants to set the `err` variable outside of the conditions scope, which didn't work because we shadowed the variable in the conditions scope

prow/cmd/hmac/fakeghhook/fakeghhook.go
```diff
@@ -169,10 +169,8 @@ func deleteHook(m map[string][]github.Hook, key string, req github.HookRequest)
 		return nil
 	}
 
-	hookExists := false
 	for i, hook := range m[key] {
 		if hook.Config.URL == req.Config.URL {
-			hookExists = true
```
alvaroaleman, line 175: We always `return` when we enter this condition, so it is impossible for any other code go see the `hookExists` variable being true

alvaroaleman (commented):
/cc @chaodaiG

def/configmap/main.go
chaodaiG: 😰

prow/cmd/generic-autobumper/bumper/bumper.go
chaodaiG: @mpherman2 for awareness

chaodaiG (approved):
/hold
I'm just finger crossed that none of these errors became intended behavior and this correction action breaks it. This probably only applies to the autobumper code

alvaroaleman:
/hold cancel
I think we can fix the autobumper afterwards if this here isn't correct

PR #21775, opened 2021-04-12
Add check on upstream bumps to make sure image exists

mpherman2:
When bumping from an upstream reference, the autobumper uses 1 tag as the reference for all of the images being bumped. This happened under the assumption that the tag existed for all images using that prefix. This turns out to not always be the case, so this PR adds a check to make sure that the tag being bumped to exists for each image.

/assign @cjwagner

### DIFF
--- a/prow/cmd/generic-autobumper/bumper/bumper.go
+++ b/prow/cmd/generic-autobumper/bumper/bumper.go
@@ -524,6 +524,7 @@ type imageBumper interface {
 	UpdateFile(tagPicker func(imageHost, imageName, currentTag string) (string, error), path string, imageFilter *regexp.Regexp) error
 	GetReplacements() map[string]string
 	AddToCache(image, newTag string)
+	TagExists(imageHost, imageName, currentTag string) (bool, error)
 }
 
 func updateReferences(imageBumperCli imageBumper, filterRegexp *regexp.Regexp, o *Options) (map[string]string, error) {
@@ -596,8 +597,17 @@ func upstreamImageVersionResolver(
 		imageFullPath := imageHost + "/" + imageName + ":" + currentTag
 		for prefix, version := range upstreamVersions {
 			if strings.HasPrefix(imageFullPath, prefix) {
-				imageBumperCli.AddToCache(imageFullPath, version)
-				return version, nil
+				exists, err := imageBumperCli.TagExists(imageHost, imageName, version)
+				if err != nil {
+					return "", err
+				}
+				if exists {
+					imageBumperCli.AddToCache(imageFullPath, version)
+					return version, nil
+				} else {
+					imageBumperCli.AddToCache(imageFullPath, currentTag)
+					return "", fmt.Errorf("Unable to bump to %s, image tag %s does not exist for %s", imageFullPath, version, imageName)
+				}
 			}
 		}
 		return currentTag, nil
--- a/prow/cmd/generic-autobumper/bumper/bumper_test.go
+++ b/prow/cmd/generic-autobumper/bumper/bumper_test.go
@@ -534,6 +534,13 @@ func (c *fakeImageBumperCli) AddToCache(image, newTag string) {
 	c.tagCache[image] = newTag
 }
 
+func (cli *fakeImageBumperCli) TagExists(imageHost, imageName, currentTag string) (bool, error) {
+	if currentTag == "DNE" {
+		return false, nil
+	}
+	return true, nil
+}
+
 func TestUpdateReferences(t *testing.T) {
 	cases := []struct {
 		description        string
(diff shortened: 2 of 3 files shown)


prow/cmd/generic-autobumper/bumper/bumper.go
```diff
@@ -596,8 +597,17 @@ func upstreamImageVersionResolver(
 		imageFullPath := imageHost + "/" + imageName + ":" + currentTag
 		for prefix, version := range upstreamVersions {
 			if strings.HasPrefix(imageFullPath, prefix) {
-				imageBumperCli.AddToCache(imageFullPath, version)
-				return version, nil
+				exists, err := imageBumperCli.TagExists(imageHost, imageName, version)
```
cjwagner, line 600: Would it make more sense to do this at a higher level than the upstream image version resolver so that this logic applies to all resolvers? We should never bump to an image that doesn't exist.

prow/cmd/generic-autobumper/bumper/bumper_test.go
```diff
@@ -534,6 +534,10 @@ func (c *fakeImageBumperCli) AddToCache(image, newTag string) {
 	c.tagCache[image] = newTag
 }
 
+func (cli *fakeImageBumperCli) TagExists(imageHost, imageName, currentTag string) (bool, error) {
+	return true, nil
```
cjwagner, line 538: Please test the new functionality with test cases where the tag does not exist.

prow/cmd/generic-autobumper/bumper/bumper.go
mpherman2: The three possibilities for TargetVersion are "upstream" "latest" and "other". This handles "upstream". "latest" already handles this issue implicitly. Anything other than upstream or latest is a hard override which will just set the image to whatever the TargetVersion is, even if it is not a valid tag. None of the previous autobumpers performed any checks for this option. I assumed we wanted users to be able to use this hard override target version to change the image tag to whatever they wished without restriction. If this was just an oversight, I can add the check there as well?

prow/cmd/generic-autobumper/bumper/bumper_test.go
mpherman2: Done

prow/cmd/generic-autobumper/bumper/bumper.go
cjwagner: The only time I can imagine wanting to bump to a tag that doesn't exist is if the image was private and the tag did actually exist, but the autobumper didn't have permission to list tags. The right way to solve that would probably be to authorize the autobumper, but I guess it doesn't matter much since we don't use the "other" mode anyways. There might be other reasons for a hard override I'm not thinking of too.
/shrug

prow/cmd/generic-autobumper/bumper/bumper_test.go
```diff
@@ -786,6 +804,16 @@ func TestUpstreamImageVersionResolver(t *testing.T) {
...
 		},
+		{
+			description:         "tag does not exist",
+			upstreamVersionType: upstreamVersion,
+			expectError:         false,
+			prefixes:            []Prefix{tagDoesNotExistPrefix},
+			imageHost:           doesNotExistPrefix,
+			currentTag:          "doesNotExist",
+			expectedTargetTag:   "",
+			resolverError:       true,
```
cjwagner, line 815: Note that while we do generate an error from the tagPicker in this case, the error is just printed to the log. https://github.com/kubernetes/test-infra/blob/f7e21a3c18f4f4bbc7ee170675ed53e4544a0632/experiment/image-bumper/bumper/bumper.go#L221
This will not cause the job to fail so this will be kind of invisible to users. Is that the preferred behavior here? I think it would be nice to know that an image couldn't be bumped, but we won't want to block the autobumper on that in all cases. Some sort of excusable notification would be nice, but I don't think we have a suitable mechanism. This might be the best option for now, just something to consider.

cjwagner (approved):
I have some more thoughts, but this might be the best solution for now. Feel free to `/hold cancel`.
/hold

prow/cmd/generic-autobumper/bumper/bumper.go
mpherman2: I mostly use the hard override mode for testing. I think it might be best to go back in another PR and formalize this by adding a "enforce image exists" option. For now I think this is fine though.

prow/cmd/generic-autobumper/bumper/bumper_test.go
mpherman2: Yea I noticed this as well. It helps that If "consistentImages" is set, then being unable to bump one of the images will cause the autobump job to fail because the images will not be consistent after the bump. The user would still need to go through the logs to find the specific cause though. I figured this was best because it would not interfere with bumps where we don't care if the images are consistent.

Issue #21655, opened 2021-04-05
All MULTIZONE CI jobs are br0ken!


dims:
- https://testgrid.k8s.io/sig-network-ingress-gce-e2e#ingress-gce-e2e-multi-zone
- https://testgrid.k8s.io/google-gce#gce-multizone
- https://testgrid.k8s.io/sig-node-containerd#e2e-cos-multizone
- https://testgrid.k8s.io/google-gce#gce-ha-master

basically all the ones marked `MULTIZONE=true` are broken - https://cs.k8s.io/?q=MULTIZONE%3Dtrue&i=nope&files=&excludeFiles=&repos=kubernetes/test-infra

Question:
- Do we just clean them up and drop MULTIZONE related stuff in the scripts/kubetest etc?

dims:
cc @spiffxp @BenTheElder @thockin @cheftako

BenTheElder:
I've been OOO and don't know much about what this is. cc @robscott @bowei

robscott:
Looks like there's more discussion on the PR now and we can safely remove the jobs: https://github.com/kubernetes/test-infra/pull/21777

BenTheElder:
these were removed https://github.com/kubernetes/test-infra/pull/21777

PR #21777, opened 2021-04-13
Drop MULTIZONE CI jobs

dims:
Drops CI jobs mentioned in https://github.com/kubernetes/test-infra/issues/21655


dchen1107:
I remembered the test was introduced initially as a regression tests, but I couldn't trace down the details. cc/ @BenTheElder 
@MrHohn  to see if they know more on this. 

On another hand since the CI jobs failed for a long time without much attention, I am totally ok to remove it from OSS dashboard.

dchen1107:
Quickly sync up with @MrHohn. The test was added to capture issues (such as misconfigured LB, etc.) when running cluster in multiple zones initially. But we all agreed that the CI is still valid for OSS K8s. Meanwhile, the initial test cases were migrated to use internal CIs already.

MrHohn:
> I remembered the test was introduced initially as a regression tests, but I couldn't trace down the details. cc/ @BenTheElder
@MrHohn to see if they know more on this.

+1, the ingress-gce one [was added](https://github.com/kubernetes/test-infra/pull/10416) to cover a regression when cluster runs on multiple zones. I believe we have internal CIs that cover the same cases already, so I'm fine with removing this CI.

dchen1107:
/lgtm
/approve

Thanks for cleaning up this.

BenTheElder:
> 2021/04/14 20:06:06 FAIL: could not write config: 1 error occurred:
>	* could not find the referenced (TestGroup) ci-kubernetes-e2e-gce-multizone

dims:
@BenTheElder fixed now.

PR #21783, opened 2021-04-13
Prow integration test: add horologium test ensure prowjob scheduled

chaodaiG:
Testing basic scheduling functions of horologium.

Also btw:
- Symlink cluster configs that are identical with prod instead of copy/paste

### DIFF
--- a/prow/test/integration/test/horologium_test.go
+++ b/prow/test/integration/test/horologium_test.go
@@ -1,0 +1,72 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package integration
+
+import (
+	"context"
+	"fmt"
+	"testing"
+	"time"
+
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/util/wait"
+	prowjobv1 "k8s.io/test-infra/prow/apis/prowjobs/v1"
+	"k8s.io/test-infra/prow/kube"
+	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func TestLaunchProwJob(t *testing.T) {
+	const existJobName = "horologium-schedule-test-job"
+	t.Parallel()
+
+	tests := []struct {
+		name string
+	}{
+		{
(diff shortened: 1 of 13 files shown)


prow/config/config.go
```diff
@@ -690,6 +691,12 @@ type Gerrit struct {
 	DeckURL string `json:"deck_url,omitempty"`
 }
 
+// Horologium is config for the Horologium.
+type Horologium struct {
+	// TickInterval is how often we do a sync with binded gerrit instance
```
alvaroaleman, line 696:
```suggestion
	// TickInterval is the interval in which we check if new jobs need to be created. Defaults to one minute.
```

prow/test/integration/test/horologium_test.go
```diff
@@ -0,0 +1,114 @@
...
+			if err != nil {
+				t.Fatalf("Failed creating clients for cluster %q: %v", clusterContext, err)
+			}
+			ctx := context.Background()
+
+			getNextRunOrFail := func(t *testing.T, jobName string, lastRun *v1.Time) *prowjobv1.ProwJob {
+				var res *prowjobv1.ProwJob
+				if err := wait.Poll(time.Second, 70*time.Second, func() (bool, error) {
+					pjs := &prowjobv1.ProwJobList{}
+					err = kubeClient.List(ctx, pjs, ctrlruntimeclient.InNamespace(defaultNamespace))
```
alvaroaleman, line 60: Instead of doing a full list, can we just list with a label selector of the job name? That would allow us to not block so much, horologium will create the job as soon as its up and if the job exist, we know the test passed.

I'd also be in favor of moving the prow-cm-tests into a dedicated PR and scoping this to "make sure horologium creates a prowjob"

chaodaiG: Done, both

prow/config/config.go
chaodaiG: Will do in next PR

prow/test/integration/test/horologium_test.go
```diff
@@ -0,0 +1,67 @@
...
+			clusterContext := getClusterContext()
+			t.Logf("Creating client for cluster: %s", clusterContext)
+			kubeClient, err := NewClients("", clusterContext)
+			if err != nil {
+				t.Fatalf("Failed creating clients for cluster %q: %v", clusterContext, err)
+			}
+			ctx := context.Background()
+
+			pjs := &prowjobv1.ProwJobList{}
+			err = kubeClient.List(ctx, pjs, &ctrlruntimeclient.ListOptions{
```
alvaroaleman, line 55: Is it safe to assume that horologium always finished starting before we run this test?

chaodaiG: Good point, for prow this probably is true, as it pushes test images before running tests https://github.com/kubernetes/test-infra/blob/b15f63b0d8faba951d40b6f11b9acf2d42934be5/prow/test/integration/setup-cluster.sh#L142. But I think it's better to wait for 1 minute just to be safe

chaodaiG: Done

alvaroaleman (approved):
thanks!

Issue #21615, opened 2021-04-01
Windows jobs on master/1.21-informing are timing out


thejoycekung:
**What happened**:
ci-kubernetes-e2e-windows-gce-2004
ci-kubernetes-e2e-windows-gce-2019
ci-kubernetes-e2e-windows-gce-20h2
ci-kubernetes-e2e-windows-gce-2004-1-21
ci-kubernetes-e2e-windows-gce-2019-1-21
ci-kubernetes-e2e-windows-gce-20h2-1-21

are all timing out. On the master side, they are timing out at 3h (recently bumped from 2h on this PR: https://github.com/kubernetes/test-infra/pull/21584); on the 1.21 side, they are timing out at 2h.

**What you expected to happen**:
I suspect that these jobs should not be taking this long to run, as similar jobs from 1.20 and 1.19 take less than half the time.

For example, these take ~1h30m to run:
- [ci-kubernetes-e2e-windows-gce-2019-1-20](https://testgrid.k8s.io/sig-release-1.20-informing#gce-windows-2019-1.20)
  - this run took 1h20m: https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/ci-kubernetes-e2e-windows-gce-2019-1-20/1377214071648030720
- [ci-kubernetes-e2e-windows-gce-2019-1-19](https://testgrid.k8s.io/sig-release-1.19-informing#gce-windows-2019-1.19)
  - this run took 1h14m: https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/ci-kubernetes-e2e-windows-gce-2019-1-19/1377663123665719296

If the jobs on master and 1.21 _are_ supposed to be taking this long to run, perhaps we should just keep bumping the timeout. But the fact that they take so much longer than the 1.20 and 1.19 jobs seems rather suspicious.

**Please provide links to example occurrences, if any**:
Here's one example from each job I listed as timing out:
- https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/ci-kubernetes-e2e-windows-gce-2004/1377600712464142336
- https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/ci-kubernetes-e2e-windows-gce-2019/1377544085862092800
- https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/ci-kubernetes-e2e-windows-gce-20h2/1377582841705009152
[...]

PR #21785, opened 2021-04-13
tests: Skips device plugin for Windows

claudiubelu:
The test ``[sig-windows] Device Plugin should be able to create a functioning device plugin for Windows`` is supposed to run on nodes with GPUs, which is not a common case, thus, it should be excluded from most test runs. The test pulls a 10+ GB container image, running for 100+ minutes, which can lead to test job timeouts.

This has been addressed on master, the test has been tagged with ``[Feature:GPUDevicePlugin]``, and most test jobs are already skipping Feature tests, but that's not the case for 1.21.

Fixes: #21615


marosset (approved):
/lgtm
/approve

PR #21787, opened 2021-04-14
Prow integration test: horologium test involves prow-controller-manager

chaodaiG:


prow/cmd/horologium/main.go
```diff
@@ -40,6 +40,10 @@ import (
 	"k8s.io/test-infra/prow/pjutil"
 )
 
+const (
+	defaultTickInterval = 1 * time.Minute
```
alvaroaleman, line 44: Nit: Multiplicating thing with one is a super useless thing to do :P

prow/test/integration/test/horologium_test.go
```diff
@@ -54,18 +57,63 @@ func TestLaunchProwJob(t *testing.T) {
...
+			}
+
+			t.Logf("Wait for the next run of %q", existJobName)
+			pj := getNextRunOrFail(t, existJobName, nil)
+			// Enforce that the previous run was created 30-60 seconds ago, so
+			// that the next run won't happen in the following 30 seconds.
+			if pj.CreationTimestamp.Add(30 * time.Second).After(time.Now()) {
+				pj = getNextRunOrFail(t, existJobName, &pj.CreationTimestamp)
+			}
+
```
alvaroaleman, line 95: OK, so I have to say I find the logic in here and what we want to test quite confusing. What exactly do we want to assert, that Horologium creates multiple ProwJob when time passes? Wouldn't it be simpler to do something like:
1. Probably wait a bit
2. List all ProwJobs with the given job name
3. Write a loop that checks for each job when it was created and expects another job if time.Now after (cron + horologium tick interval)

Also, this doesn't really require the prow controller manager, as long as max_concurrency of the job is zero.

chaodaiG: In local test I have found that the start time of each prow job was exactly 60 seconds apart from each other, I guess I can make it assert this behavior first. The logic I really wanted to test here is that when a job started out of schedule as https://github.com/kubernetes/test-infra/pull/21787/files#diff-90d950be47dbcefc00fe33839bbba6b6a2d3c0c137ec6dec99baf22c14eda471R107, the next prowjob would respect the new schedule. I can add more docs to clarify. WDYT?

> Also, this doesn't really require the prow controller manager, as long as max_concurrency of the job is zero.

That's what I thought too, but the current behavior of horologium does care whether previous job had finished or not: https://github.com/kubernetes/test-infra/blob/45dfeedbcec3c8319c9672ad7d5259d498eabe37/prow/cmd/horologium/main.go#L167

alvaroaleman: Ok interesting, so we just ignore the `max_concurrency` setting there and have it implicitly always be 1. Some more godocs would be great, yeah.

chaodaiG: Yeah it surprised me as well, but have to admit that writing integration test is a good way of learning how prow really works

chaodaiG: Added more doc, hopefully it's clearer

alvaroaleman (approved):
thanks!

PR #21806, opened 2021-04-14
add tam7t as reviewer for secrets-store-csi-driver test config

aramase:

@tam7t is already a reviewer for the Secrets Store CSI Driver sig-auth subproject. Adding Tommy as reviewer to the test config as well.

Membership request for kubernetes org is complete: https://github.com/kubernetes/org/issues/2638

/assign @ritazh 
/cc @tam7t


aramase:
> The following users are mentioned in OWNERS file(s) but are untrusted for the following reasons. One way to make the user trusted is to add them as members of the kubernetes org. You can then trigger verification by writing /verify-owners in a comment.

We can run `/verify-owners` when Tommy gets and accepts the invite.

PR #21807, opened 2021-04-14
Make supplemental prow config filename configurable

alvaroaleman:
In order to actually be able to put the prow config into an org/repo
hirarchy, we need to filter it by filename, because otherwise:
* We need to pass a flag per org, as we'd otherwise include the main
  config which would fail when trying to merge
* We eventually want to also put the plugin config into the same
  hirarchy and that one will obviously not be parseable as prow config

Because this is the second time I have to touch every single binary that
reads the Prow config, I have created a new flagutil for this. This new
flagutil got its own package because:
* Otherwise there is an import cycle because prow/config transiently
  imports it
* All consumers that import the flagutil package would also transiently
  import the config package which has _a lot_ of imports itself

Because we didn't have a flagutil for this yet, we were using this
quite inconsistently:
* The flag names are sometimes different
* Some binaries had defaults, most don't

Thus a lot of options needed to be exported.

Reviewing this is tedious but at the end boils down to:
* If the flag names were not `config-path`/`job-config-path` they must be explicitly in the `ConfigOptions`
* If one of the flags had a default, that must be explicitly configured now
* `AddFlags` must be called

I've done a round myself today and _think_ I got everything.


experiment/ci-janitor/main.go
```diff
@@ -139,10 +131,11 @@ func main() {
 		logrus.Fatalf("Invalid options: %v", err)
 	}
 
-	conf, err := config.Load(o.configPath, o.jobConfigPath, o.supplementalProwConfigDirs.Strings())
+	agent, err := o.prowConfig.ConfigAgent()
```
stevekuznetsov, line 134: nit: technically `prowConfig.ConfigAgent` stutters, Go-ism would be `prowConfig.Agent()`

experiment/prowjob-report/main.go
```diff
@@ -38,43 +38,39 @@ import (
...
 // Loaded at TestMain.
 var prowConfig *cfg.Config
 
 func main() {
-	flag.Var(&supplementalProwConfigDirs, "supplemental-prow-config-dir", "An additional directory from which to load prow configs. Can be used for config sharding but only supports a subset of the config. The flag can be passed multiple times.")
-	flag.Parse()
-	if *configPath == "" {
-		fmt.Println("--config must set")
-		os.Exit(1)
+	configOpts := configflagutil.ConfigOptions{
```
stevekuznetsov, line 55: It's a lot of follow-up work but do you think / does @cjwagner think we should collapse these at some point so that there aren't e.g. different defaults across the tooling ?

prow/cmd/checkconfig/main_test.go
```diff
@@ -1326,7 +1337,7 @@ func TestOptions(t *testing.T) {
 			case actualErr != nil:
 				t.Errorf("unexpected error: %v", actualErr)
 			case !reflect.DeepEqual(&actualOptions, tc.expectedOptions):
-				t.Errorf("actual %#v != expected %#v", actualOptions, *tc.expectedOptions)
+				t.Errorf("actual \n%#v\n != expected \n%#v\n", actualOptions, *tc.expectedOptions)
```
stevekuznetsov, line 1340: nit: `cmp.Diff` ?

stevekuznetsov (approved):
Changes here look totally reasonable, it's a shame this was not DRY before. "High risk" change because manually validating that 35 default options did not change is hard.

PR #20581, opened 2021-01-23
Explicitly disable InTreePluginGCEUnregister for all alpha jobs

Jiawei0227:
Follow this change of adding InTreePluginGCEUnregister flag in https://github.com/kubernetes/kubernetes/pull/98243, we will need to explicitly disable InTreePluginGCEUnregister for all alpha jobs since all the tests using gce-pd intree plugin will fail because CSI Driver have not been installed yet.

/cc @msau42

PR #21809, opened 2021-04-14
Remove InTreePluginGCEUnregister feature flag from previous branches

hasheddan:
Updates alphafeatures jobs on branches where InTreePluginGCEUnregister
has not yet been added to exclude the flag for the time-being. It should
be re-enabled when 1.21 reaches each respective version marker.

Partial revert of https://github.com/kubernetes/test-infra/pull/20581

/assign @puerco @justaugustus 
/cc @Jiawei0227

### DIFF
--- a/config/jobs/kubernetes/generated/generated.yaml
+++ b/config/jobs/kubernetes/generated/generated.yaml
@@ -62,7 +62,7 @@ periodics:
       - --timeout=180m
       - --env=KUBE_PROXY_DAEMONSET=true
       - --env=ENABLE_POD_PRIORITY=true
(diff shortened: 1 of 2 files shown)


justaugustus (changes requested):
@hasheddan -- put a quick `/hold` on this to beat the bots.

We'll need to regenerate the jobs or manually remove the flag in the relevant branch jobs.
The latter makes sense here.

hasheddan:
@justaugustus thanks! Just updated :)

justaugustus (approved):
Thanks!
/hold cancel

PR #21823, opened 2021-04-16
Update plugins formatting & ingress to v1

fmuyassarov:
Remove some duplication from `args`, update the plugins.yaml formatting as suggested [here](https://github.com/kubernetes/test-infra/issues/20631#issuecomment-787693609), update the ingress api to v1.

### DIFF
--- a/config/prow/cluster/starter-s3.yaml
+++ b/config/prow/cluster/starter-s3.yaml
@@ -16,20 +16,21 @@ data:
   plugins.yaml: |
     plugins:
       << your_github_org >>:
-      - approve
-      - assign
-      - blunderbuss
-      - cat
-      - dog
-      - help
-      - heart
-      - hold
-      - label
-      - lgtm
-      - trigger
-      - verify-owners
-      - wip
-      - yuks
+        plugins:
+        - approve
+        - assign
+        - blunderbuss
+        - cat
+        - dog
+        - help
+        - heart
+        - hold
+        - label
+        - lgtm
+        - trigger
+        - verify-owners
+        - wip
+        - yuks
 ---
 apiVersion: v1
 kind: Secret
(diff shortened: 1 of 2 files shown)


config/prow/cluster/starter-s3.yaml
```diff
@@ -532,7 +532,7 @@ spec:
   - port: 80
     targetPort: 8888
 ---
-apiVersion: extensions/v1beta1
+apiVersion: networking.k8s.io/v1
```
fejta, line 535: I suspect this will break kubectl apply, right?

fejta (approved):
/hold

config/prow/cluster/starter-s3.yaml
fmuyassarov: Do you mean because there might be `extensions/v1beta1` version still available on the cluster?

fejta: Oh sorry these are both just starter examples? Nevermind.
/hold cancel

fejta (approved):
/ok-to-test

config/prow/cluster/starter-s3.yaml
fejta: (my concern is that changing the api version of our production ingress resources will cause our production deployments to start failing as I would expect that field to be immutable (could be wrong, but not something I want to merge until the morning)

fmuyassarov: Yes I agree, I was hesitating too to update it, but went ahead considering these are starters for someone like me who is trying to setup a prow cluster.

fejta: Totally, thanks so much for cleaning this up!!

PR #21827, opened 2021-04-16
Flagutil: return error if --github-token-path isn't set

### LINKED ISSUE #21797: Flagutil: return error if `--github-token-path` isn't set
missing required flag: please set to --github-token-path=%s before June 2020

**Provide any links for context**:
https://github.com/kubernetes/test-infra/blob/3bea7c721703e690e4c974d6d8feaf60db4dec32/prow/flagutil/github.go#L87

matthyx:
fix #21797

### DIFF
--- a/prow/flagutil/github.go
+++ b/prow/flagutil/github.go
@@ -83,12 +83,6 @@ func (o *GitHubOptions) Validate(bool) error {
 		return errors.New("--app-id and --app-private-key-path must be set together")
 	}
 
-	if o.TokenPath == "" && o.AppID == "" && !o.AllowAnonymous {
-		// TODO(fejta): just return error after May 2020
-		logrus.Warnf("missing required flag: please set to --github-token-path=%s before June 2020", DefaultGitHubTokenPath)
-		o.TokenPath = DefaultGitHubTokenPath
-	}
-
 	if o.TokenPath != "" && len(endpoints) == 1 && endpoints[0] == github.DefaultAPIEndpoint && !o.AllowDirectAccess {
 		logrus.Warn("It doesn't look like you are using ghproxy to cache API calls to GitHub! This has become a required component of Prow and other components will soon be allowed to add features that may rapidly consume API ratelimit without caching. Starting May 1, 2020 use Prow components without ghproxy at your own risk! https://github.com/kubernetes/test-infra/tree/master/ghproxy#ghproxy")
 	}
--- a/prow/ANNOUNCEMENTS.md
+++ b/prow/ANNOUNCEMENTS.md
@@ -140,6 +140,7 @@ Note: versions specified in these announcements may not include bug fixes made
 in more recent versions so it is recommended that the most recent versions are
 used when updating deployments.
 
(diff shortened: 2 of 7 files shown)


alvaroaleman (approved):

prow/flagutil/github.go
```diff
@@ -84,9 +84,7 @@ func (o *GitHubOptions) Validate(bool) error {
 	}
 
 	if o.TokenPath == "" && o.AppID == "" && !o.AllowAnonymous {
-		// TODO(fejta): just return error after May 2020
-		logrus.Warnf("missing required flag: please set to --github-token-path=%s before June 2020", DefaultGitHubTokenPath)
-		o.TokenPath = DefaultGitHubTokenPath
+		return fmt.Errorf("missing required flag: please set to --github-token-path=%s", DefaultGitHubTokenPath)
```
fejta, line 87: I'm not sure that this is appropriate any longer.

Components like crier, deck, mkpj are all used in githubless environments and should not require `--github-token-path != ""`

matthyx: alright, then should we close this and only remove the TODO?

fejta: Well I do think we should remove `DefaultGitHubTokenPath` but not return an error if it is empty as some binaries (crier) sometimes but not always use github.

matthyx: Alright, but then when an app that needs it doesn't provide the option, we get such error message:
`error starting secret agent : error reading : open : no such file or directory`

matthyx: Maybe we should specify somehow that this config is optional for (crier) and others?

maintenance/migratestatus/migratestatus_test.go
```diff
@@ -19,16 +19,20 @@ package main
 import (
 	"strings"
 	"testing"
+
+	"k8s.io/test-infra/prow/flagutil"
 )
 
 const validURL = "http://example.org"
 
+var gitHubOptions = flagutil.GitHubOptions{TokenPath: flagutil.DefaultGitHubTokenPath}
```
fejta, line 28: Should we be deleting DefaultGithubTokenPath? That is the goal here.

prow/flagutil/github.go
fejta: Making the GitHubOptions struct more complicated so it can distinguish between token path being required or optional works for sure!

I'm okay them getting an `error stating secret agent` -- likely their path is not the default one so they set it anyway and even if they are using the default they've a super long time to start explicitly setting it.

matthyx: Alright, I chose the easiest of the 2 :)

fejta:
Looks like some unit tests need an update?

matthyx:
Indeed, sorry Erick. I'll correct that tomorrow morning.

matthyx:
> Looks like some unit tests need an update?

done

fejta (approved):
Thanks so much for working through the multiple rounds here!

PR #21833, opened 2021-04-16
test images: Adds private registry images

claudiubelu:
Currently, we're pulling the ``alpine`` and ``windows-nanoserver`` images from the ``gcr.io/authenticated-image-pulling`` registry. We can add jobs that will mirror those images into own own private registry.

Depends On: https://github.com/kubernetes/k8s.io/pull/1929
Related: https://github.com/kubernetes/k8s.io/issues/1528


BenTheElder:
still depends on https://github.com/kubernetes/k8s.io/pull/1929

PR #21835, opened 2021-04-16
Spyglass: Add a lens to display html from the artifacts

alvaroaleman:
This change adds a lens to spyglass that if setup will display html
files from the artifacts dir. This makes it very easy for job owners to
extend the view with custom output of their own.

Sample view when including [this](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/logs/periodic-ci-openshift-release-master-ci-4.8-upgrade-from-stable-4.7-e2e-aws-ovn-upgrade/1382306626584711168/artifacts/e2e-aws-ovn-upgrade/openshift-e2e-test/artifacts/e2e-intervals_20210414-125331.html) (Cut off on the screenshot because its too large, but its displayed properly until the end):

expanded

Collapsed:

collapsed

### DIFF
--- a/prow/cmd/deck/BUILD.bazel
+++ b/prow/cmd/deck/BUILD.bazel
@@ -157,6 +157,7 @@ go_library(
         "//prow/spyglass/lenses/buildlog:go_default_library",
         "//prow/spyglass/lenses/common:go_default_library",
         "//prow/spyglass/lenses/coverage:go_default_library",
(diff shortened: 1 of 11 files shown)


stevekuznetsov (approved):

stevekuznetsov:
The little resizing snippet is clever and likely a bit insane but seems reasonable given the tiling of iFrames we're doing anyway ...

PR #21847, opened 2021-04-19
Temporarily disable `secping` job

sfowl:
See: https://github.com/kubernetes/security/issues/56

### DIFF
--- a/config/jobs/kubernetes/test-infra/fejta-bot-periodics.yaml
+++ b/config/jobs/kubernetes/test-infra/fejta-bot-periodics.yaml
@@ -330,40 +330,43 @@ periodics:
       secret:
         secretName: fejta-bot-token
 
+# XXX Temporarily disabled during the transition to `security_contacts` in `OWNERS`
+# See: https://github.com/kubernetes/security/issues/56
+#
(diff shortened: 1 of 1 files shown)


chases2:
/ok-to-test

Do we want to disable secping like this before the new security contact solution is working?

sfowl:
> Do we want to disable secping like this before the new security contact solution is working?

I opened this with the do-not-merge label while the spec change is being reviewed in kubernetes/community. I'll remove the label once the kubernetes/community change is merged, then once `secping` is disabled I'll begin on the bulk SECURITY_CONTACTS updates.

There's a bit more detail in the comments of kubernetes/community#5398.

fejta:
/uncc @wojtek-t 
/cc
/cc @IanColdwater @tabbysable 

a) Have you run this idea by sig-security? Adding sig-security leads to review and/or find reviewers. Not clear if this is a proposal or a PR to implement a decision.

b) Please consider moving this into a sig-security directory with an OWNERS file that references sig-security, which will property let you make these decisions without involving sig-testing: https://github.com/kubernetes/test-infra/tree/master/config/jobs/kubernetes

PR #21857, opened 2021-04-19
Spyglass html lense: Hide documents by default

alvaroaleman:
This allows us to both not show documents by default and not break their
rendering if they fail expect a non-zero size.


alvaroaleman:
/assign @petr-muller 
Could you have a look?

stevekuznetsov (approved):

PR #21870, opened 2021-04-20
Configupdater: Add use_full_path_as_key option

alvaroaleman:
Currently the configupdater will always use the basename of a filename
as the configmay key if none is explicitly configured. Because of that,
all filenames have to be unique, which is very inconvenient.

This change adds a use_full_path_as_key option, which if enabled will
use the full path, replacing dashes with slashes.

### DIFF
--- a/prow/plugins/plugin-config-documented.yaml
+++ b/prow/plugins/plugin-config-documented.yaml
@@ -351,7 +351,9 @@ config_updater:
             gzip: false
 
             # Key is the key in the ConfigMap to update with the file contents.
-            # If no explicit key is given, the basename of the file will be used.
+            # If no explicit key is given, the basename of the file will be used unless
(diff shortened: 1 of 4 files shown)


prow/plugins/plugin-config-documented.yaml
```diff
@@ -351,7 +351,9 @@ config_updater:
             gzip: false
 
             # Key is the key in the ConfigMap to update with the file contents.
-            # If no explicit key is given, the basename of the file will be used.
+            # If no explicit key is given, the basename of the file will be used unless
+            # use_full_path_as_key: true is set, in which case the full filepath relative
+            # to the repository root will be used, replacing slashes with dashes.
```
alvaroaleman, line 356: There seems to be a bug in genyaml here, it should have produced something.

stevekuznetsov (approved):

PR #21877, opened 2021-04-20
Return error when `--github-token-path` is missed

kskumgk63:
Hi,
I've found the `TODO` comment below, and fixed it.

https://github.com/kubernetes/test-infra/blob/0f3eef4736c0519f009e1e71e06b13e91af09b2b/prow/flagutil/github.go#L86-L90


matthyx:
/hold

@kskumgk63 looks like a duplicate for https://github.com/kubernetes/test-infra/pull/21827

Please check if there is already an issue for it, and if it has been assigned, in that case https://github.com/kubernetes/test-infra/issues/21797.

PR #21882, opened 2021-04-21
Update blob-csi-driver-config.yaml

andyzhangx:
set minimum k8s version as 1.17
```
Error: loading API model in generateCmd: error parsing the api model: the following OrchestratorProfile configuration is not supported: OrchestratorType: "Kubernetes", OrchestratorRelease: "1.16", OrchestratorVersion: "". Please use one of the following versions: [1.17.17 1.18.17 1.19.9 1.20.5 1.21.0]
Usage:
  aks-engine generate [flags]
```


ZeroMagic (approved):
/lgtm

PR #21885, opened 2021-04-21
Cherrypick: support multiple calls per comment

### LINKED ISSUE #21714: cherrypick plugin doesn't response to all commands in a github comment
**What happened**:
cherry-pick plugin didn't detect multiple uses of `/cherry-pick` command in the same message

ie. from https://github.com/knative/serving/pull/11127#issuecomment-814917770
```
/cherry-pick release-0.22
/cherry-pick release-0.21
```
It would schedule two cherry picks

matthyx:
fix #21714

### DIFF
--- a/prow/external-plugins/cherrypicker/server.go
+++ b/prow/external-plugins/cherrypicker/server.go
@@ -286,15 +286,13 @@ func (s *Server) handlePullRequest(l *logrus.Entry, pre github.PullRequestEvent)
 	for i := range comments {
 		c := comments[i]
 		cherryPickMatches := cherryPickRe.FindAllStringSubmatch(c.Body, -1)
-		if len(cherryPickMatches) == 0 || len(cherryPickMatches[0]) != 2 {
-			continue
+		for _, match := range cherryPickMatches {
+			targetBranch := strings.TrimSpace(match[1])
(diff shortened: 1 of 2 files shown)


prow/external-plugins/cherrypicker/server.go
```diff
@@ -289,12 +289,13 @@ func (s *Server) handlePullRequest(l *logrus.Entry, pre github.PullRequestEvent)
 		if len(cherryPickMatches) == 0 || len(cherryPickMatches[0]) != 2 {
```
alvaroaleman, line 289: Don't we need to move the second check into the loop?

matthyx: I think we can remove it completely... either it matches and match[1] exists, or it doesn't and we won't iterate.

matthyx: I mean this would be silly, right?
```
			if len(match) != 2 {
				continue
			}
```

alvaroaleman: I guess you are right, if it matches the capture group will have a value so we can just remove the check

alvaroaleman (approved):

PR #13627, opened 2019-07-26
checkconfig: add optional check to verify OWNERS presence

petr-muller:
When a repo enables an `approve` plugin but does not contain OWNERS
file, it does not make sense and the situation pollutes the logs.

The file presence is checked using `RepositoryClient.GetFile()` method
which transfers the content that we do not actually use, but given
OWNERS files are generally very small, I think it is not worth
implementing something like a `HasFile` with a HEAD request.

Because this check is not local and cheap, I did not make it run by
default and introduced a new option to enable "expensive" checks.

PR #21891, opened 2021-04-21
gh: expose throttler parameters as options

petr-muller:
Tools using the client may want to configure the behavior of their
throttling (these values are currently usually hardcoded). GH client
parameters are configured in a shared structure, but different tools
that use the client may want to specify different defaults depending on
their use case.

Add a new method that allows to build a flag parser using specified
defaults. I added a method using the functional options pattern so it is
extensible and quite easily usable from callsites.

The change makes all clients constructed from GH options call
`Throttle`, but the defaults are zeroes, which result in unthrottled
clients. The existing users of `Throttle` will not be broken, they will
just call `Throttle` twice, once with defaults, once with their values,
so their behavior should not change. The existing users can be ported to
defaults by removing their hardcoded `Throttle` calls and providing
custom defaults for options. I included such change for the
`needs-rebase` plugin and will port other callsites in test-infra if the
PR is accepted.

### DIFF
--- a/prow/cmd/branchprotector/protect.go
+++ b/prow/cmd/branchprotector/protect.go
@@ -47,9 +47,12 @@ type options struct {
 	config             configflagutil.ConfigOptions
 	confirm            bool
 	verifyRestrictions bool
-	tokens             int
-	tokenBurst         int
-	github             flagutil.GitHubOptions
+
+	// TODO(petr-muller): Remove after August 2021, replaced by github.ThrottleHourlyTokens
+	tokens     int
+	tokenBurst int
+
+	github flagutil.GitHubOptions
 }
 
 func (o *options) Validate() error {
@@ -61,6 +64,21 @@ func (o *options) Validate() error {
 		return err
 	}
 
+	if o.tokens != defaultTokens {
+		if o.github.ThrottleHourlyTokens != defaultTokens {
+			return fmt.Errorf("--tokens cannot be specified together with --github-hourly-tokens: use just the latter")
+		}
+		logrus.Warn("--tokens is deprecated: use --github-hourly-tokens instead")
+		o.github.ThrottleHourlyTokens = o.tokens
+	}
+	if o.tokenBurst != defaultBurst {
+		if o.github.ThrottleAllowBurst != defaultBurst {
+			return fmt.Errorf("--token-burst cannot be specified together with --github-allowed-burst: use just the latter")
+		}
+		logrus.Warn("--token-burst is deprecated: use --github-allowed-burst instead")
+		o.github.ThrottleAllowBurst = o.tokenBurst
+	}
+
 	return nil
 }
 
@@ -69,10 +87,10 @@ func gatherOptions() options {
 	fs := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
 	fs.BoolVar(&o.confirm, "confirm", false, "Mutate github if set")
 	fs.BoolVar(&o.verifyRestrictions, "verify-restrictions", false, "Verify the restrictions section of the request for authorized collaborators/teams")
-	fs.IntVar(&o.tokens, "tokens", defaultTokens, "Throttle hourly token consumption (0 to disable)")
-	fs.IntVar(&o.tokenBurst, "token-burst", defaultBurst, "Allow consuming a subset of hourly tokens in a short burst")
+	fs.IntVar(&o.tokens, "tokens", defaultTokens, "Throttle hourly token consumption (0 to disable) DEPRECATED: use --github-hourly-tokens")
+	fs.IntVar(&o.tokenBurst, "token-burst", defaultBurst, "Allow consuming a subset of hourly tokens in a short burst. DEPRECATED: use --github-allowed-burst")
 	o.config.AddFlags(fs)
-	o.github.AddFlags(fs)
+	o.github.AddCustomizedFlags(fs, flagutil.ThrottlerDefaults(defaultTokens, defaultBurst))
 	fs.Parse(os.Args[1:])
 	return o
 }
@@ -121,7 +139,6 @@ func main() {
 	if err != nil {
 		logrus.WithError(err).Fatal("Error getting GitHub client.")
 	}
-	githubClient.Throttle(o.tokens, o.tokenBurst)
 
 	p := protector{
 		client:             githubClient,
--- a/prow/cmd/checkconfig/main.go
+++ b/prow/cmd/checkconfig/main.go
@@ -109,6 +109,9 @@ const (
 	validateClusterFieldWarning                   = "validate-cluster-field"
 	validateSupplementalProwConfigOrgRepoHirarchy = "validate-supplemental-prow-config-hirarchy"
 	validateUnmanagedBranchConfigHasNoSubconfig   = "validate-unmanaged-branchconfig-has-no-subconfig"
+
+	defaultHourlyTokens = 3000
+	defaultAllowedBurst = 100
 )
 
 var defaultWarnings = []string{
@@ -133,6 +136,8 @@ var defaultWarnings = []string{
 var expensiveWarnings = []string{
 	verifyOwnersFilePresence,
 }
+
+var throttlerDefaults = flagutil.ThrottlerDefaults(defaultHourlyTokens, defaultAllowedBurst)
 
 func getAllWarnings() []string {
 	var all []string
@@ -188,7 +193,7 @@ func (o *options) gatherOptions(flag *flag.FlagSet, args []string) error {
 	flag.Var(&o.excludeWarnings, "exclude-warning", "Warnings to exclude. Use repeatedly to provide a list of warnings to exclude")
 	flag.BoolVar(&o.expensive, "expensive-checks", false, "If set, additional expensive warnings will be enabled")
 	flag.BoolVar(&o.strict, "strict", false, "If set, consider all warnings as errors.")
-	o.github.AddFlags(flag)
+	o.github.AddCustomizedFlags(flag, throttlerDefaults)
 	o.github.AllowAnonymous = true
 	o.config.AddFlags(flag)
 	if err := flag.Parse(args); err != nil {
@@ -271,7 +276,6 @@ func validate(o options) error {
 		if err != nil {
 			return fmt.Errorf("error loading GitHub client: %w", err)
 		}
-		githubClient.Throttle(3000, 100) // 300 hourly tokens, bursts of 100
 		// 404s are expected to happen, no point in retrying
 		githubClient.SetMax404Retries(0)
 
--- a/prow/external-plugins/needs-rebase/main.go
+++ b/prow/external-plugins/needs-rebase/main.go
@@ -47,18 +47,30 @@ type options struct {
 	github                 prowflagutil.GitHubOptions
 	instrumentationOptions prowflagutil.InstrumentationOptions
 	logLevel               string
-	hourlyTokens           int
+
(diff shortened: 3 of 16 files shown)


prow/flagutil/github.go
```diff
@@ -47,25 +47,62 @@ type GitHubOptions struct {
...
 // AddFlags injects GitHub options into the given FlagSet.
 func (o *GitHubOptions) AddFlags(fs *flag.FlagSet) {
-	fs.StringVar(&o.Host, "github-host", github.DefaultHost, "GitHub's default host (may differ for enterprise)")
-	o.endpoint = NewStrings(github.DefaultAPIEndpoint)
+	defaults := githubDefaults()
+	o.addFlags(fs, defaults)
+}
+
+// AddFlagsWithCustomDefaults injects GitHub options into the given FlagSet,
+func (o *GitHubOptions) AddFlagsWithCustomDefaults(fs *flag.FlagSet, defaulters ...CustomGithubDefaulter) {
```
alvaroaleman, line 83: How do we deal with tide which constructs two clients with different defaults for their throttler settings? And also I would have thought we have more components with pre-existing flags for this

petr-muller: > How do we deal with tide which constructs two clients with different defaults for their throttler settings?

The change should at least not break it, `Throttle` is smart enough and it can handle multiple subsequent `Throttle` calls, so Tide should just continue working same as it did before. The problematic part is probably that the components that do something smart like this will have confusing `--help` output and values passed in this new param would not be respected. Would that be a blocker?

> And also I would have thought we have more components with pre-existing flags for this

We probably do, but I haven't invested the effort in porting them all. I will do the legwork if the general approach is accepted.

alvaroaleman: > The problematic part is probably that the components that do something smart like this will have confusing --help output and values passed in this new param would not be respected. Would that be a blocker?

Can we maybe just add an option to not add the new throttler flags if requested and use that?

> We probably do, but I haven't invested the effort in porting them all. I will do the legwork if the general approach is accepted.

sounds reasonable

petr-muller: One way how to make Tide conform is to infer its partial throttlers from the totals given by the params. That way Tide will first create the clients trottled to the specified values (right now the clients are unthrottled) and then subsequent `Throttle` calls cal further restrict the individual clients.

petr-muller: > Can we maybe just add an option to not add the new throttler flags if requested and use that?

Yep. It does not fit into the current pattern but it should not be hard to bend it for that purpose.

prow/flagutil/github.go
```diff
@@ -47,25 +47,62 @@ type GitHubOptions struct {
...
-	fs.StringVar(&o.AppPrivateKeyPath, "github-app-private-key-path", "", "Path to the private key of the github app. If set, requires --github-app-id to bet set and --github-token-path to be unset")
+	fs.StringVar(&o.graphqlEndpoint, "github-graphql-endpoint", defaults.graphqlEndpoint, "GitHub GraphQL API endpoint (may differ for enterprise).")
+	fs.StringVar(&o.TokenPath, "github-token-path", defaults.TokenPath, "Path to the file containing the GitHub OAuth secret.")
+	fs.StringVar(&o.AppID, "github-app-id", defaults.AppID, "ID of the GitHub app. If set, requires --github-app-private-key-path to be set and --github-token-path to be unset.")
+	fs.StringVar(&o.AppPrivateKeyPath, "github-app-private-key-path", defaults.AppPrivateKeyPath, "Path to the private key of the github app. If set, requires --github-app-id to bet set and --github-token-path to be unset")
+	fs.IntVar(&o.ThrottleHourlyTokens, "github-hourly-tokens", defaults.ThrottleHourlyTokens, "If set to a value larger than zero, enable client-side throttling to limit hourly token consumption. If set, --github-allowed-burst must be positive too.")
```
alvaroaleman, line 100: I was kind of expecting the flag adding to be conditional

petr-muller: Why? I can make it so, but is there a good reason to make them so?

alvaroaleman: Yes, something like tide will have multiple flags for this and one of it just doesn't do anything which is confusing at best

petr-muller: Maybe I misunderstood, but I thought you suggested the https://github.com/kubernetes/test-infra/pull/21891#discussion_r617678748 to allow opting out.

alvaroaleman: I wrote:
> Can we maybe just add an option to not add the new throttler flags if requested and use that?

Adding a bool option to the flagutil to not add these two flags if explicitly requested is what I meant by that

petr-muller:
I ported all users of `Throttle()`. Tide is the only thing that wants to do its own thing and suppress the common options. The only other leftover is `label_sync` which does not use the common option group at all, so it would need to be ported to it entirely and I do not intend to do that in the PR.

1. Need to fix a bunch of tests, mostly related to existing same-purpose options in existing tools. 
2. Need to fix the old + new option resolution, what I wrote originally for needs-rebase is trash

prow/flagutil/github.go
```diff
@@ -47,25 +47,87 @@ type GitHubOptions struct {
...
+		o.defaults.ThrottleHourlyTokens = hourlyTokens
+		o.defaults.ThrottleAllowBurst = allowedBursts
+	}
+}
+
+// DenyThrottlerOptions suppresses the presence of throttler-related flags,
+// effectively disallowing external users to parametrize default throttling
+// behavior. This is useful mostly when a program creates multiple GH clients
+// with different behavior.
+func DenyThrottlerOptions() FlagParameter {
```
alvaroaleman, line 84:
```suggestion
func DisableThrottlerOptions() FlagParameter {
```

prow/cmd/checkconfig/main.go
```diff
@@ -271,7 +276,6 @@ func validate(o options) error {
 		if err != nil {
 			return fmt.Errorf("error loading GitHub client: %w", err)
 		}
-		githubClient.Throttle(3000, 100) // 300 hourly tokens, bursts of 100
```
alvaroaleman, line 274: I think this has a zero too much :D

prow/cmd/branchprotector/protect.go
```diff
@@ -47,9 +47,12 @@ type options struct {
 	config             configflagutil.ConfigOptions
 	confirm            bool
 	verifyRestrictions bool
-	tokens             int
-	tokenBurst         int
-	github             flagutil.GitHubOptions
+
+	// TODO(petr-muller): Remove after August 2021, replaced by github.ThrottleHourlyTokens
```
alvaroaleman, line 51: Maybe add a deprecation warning to ANNOUNCEMENTS.md?

prow/flagutil/github.go
```diff
@@ -93,6 +155,18 @@ func (o *GitHubOptions) Validate(bool) error {
 		return fmt.Errorf("invalid -github-graphql-endpoint URI: %q", o.graphqlEndpoint)
 	}
 
+	if (o.ThrottleHourlyTokens > 0) != (o.ThrottleAllowBurst > 0) {
+		if o.ThrottleHourlyTokens == 0 {
+			// Tolerate `--github-hourly-tokens=0` alone to disable throttling
```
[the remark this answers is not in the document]
petr-muller: Well spotted, this is a little ugly.

Initially I had it like you say: either both are positive, or both are zero. I changed it to the current state to not break users. The existing tools that 1) throttled by default 2) had custom throttling settings that allowed to set hourly tokens to zero to disable throttling allowed a call like this, without setting the burst option to zero too:

```
$ peribolos --hourly-tokens=0 ...
```

In this PR, these legacy options are translated to the new common ones, so existing usage like above would be broken if we enforced that both params are zero.

prow/external-plugins/needs-rebase/main.go
```diff
@@ -73,9 +85,11 @@ func gatherOptions() options {
 	fs.DurationVar(&o.updatePeriod, "update-period", time.Hour*24, "Period duration for periodic scans of all PRs.")
 	fs.StringVar(&o.webhookSecretFile, "hmac-secret-file", "/etc/webhook/hmac", "Path to the file containing the GitHub HMAC secret.")
 	fs.StringVar(&o.logLevel, "log-level", "debug", fmt.Sprintf("Log level is one of %v.", logrus.AllLevels))
-	fs.IntVar(&o.hourlyTokens, "hourly-tokens", 360, "The number of hourly tokens need-rebase may use")
+	fs.IntVar(&o.hourlyTokens, "hourly-tokens", defaultHourlyTokens, "The number of hourly tokens need-rebase may use. DEPRECATED: use --github-allowed-burst")
```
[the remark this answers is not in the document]
petr-muller: We do, on L72

prow/flagutil/github.go
petr-muller: #drunkcoding but I do not drink anymore so /shrug

prow/cmd/checkconfig/main.go
petr-muller: Heh. It's this way since 7 Aug 2019 when I made that mistake in https://github.com/kubernetes/test-infra/pull/13627 so maybe let's keep it if nobody complained until now?

prow/cmd/branchprotector/protect.go
petr-muller: Sounds good!

prow/flagutil/github.go
petr-muller: Fixed everywhere.

prow/flagutil/github.go
```diff
@@ -47,25 +47,87 @@ type GitHubOptions struct {
...
+	fs.StringVar(&o.graphqlEndpoint, "github-graphql-endpoint", defaults.graphqlEndpoint, "GitHub GraphQL API endpoint (may differ for enterprise).")
+	fs.StringVar(&o.TokenPath, "github-token-path", defaults.TokenPath, "Path to the file containing the GitHub OAuth secret.")
+	fs.StringVar(&o.AppID, "github-app-id", defaults.AppID, "ID of the GitHub app. If set, requires --github-app-private-key-path to be set and --github-token-path to be unset.")
+	fs.StringVar(&o.AppPrivateKeyPath, "github-app-private-key-path", defaults.AppPrivateKeyPath, "Path to the private key of the github app. If set, requires --github-app-id to bet set and --github-token-path to be unset")
+
+	if !params.denyThrottlerOptions {
+		fs.IntVar(&o.ThrottleHourlyTokens, "github-hourly-tokens", defaults.ThrottleHourlyTokens, "If set to a value larger than zero, enable client-side throttling to limit hourly token consumption. If set, --github-allowed-burst must be positive too.")
+		fs.IntVar(&o.ThrottleAllowBurst, "github-allowed-burst", defaults.ThrottleAllowBurst, "Size of token consumption bursts. If set, --github-hourly-tokens must be positive too and set to a higher or equal number.")
```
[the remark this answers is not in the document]
petr-muller: In the current implementation that default will either be zero (and `ThrottleHourlyTokens` will be too) when the program does not customize the defaults, or `ThrottleAllowBurst` will be explicitly set to something via `ThrottlerDefaults`.

petr-muller:
Thanks for the review!  Addressed all feedback by either code changes (resolved convos) or I responded with explaining why some parts are done the way they are.

alvaroaleman (approved):
thanks! I am going to squash this as none of the commits can be reverted individually

Issue #21869, opened 2021-04-20
Prow controller manager: recreate pod failed with 'pod already exist'


chaodaiG:
**What happened**:

It happens not very frequently but was observed on google owned prow instance, that a job was reported failed with `pod already exist`, after logs tracing, prow-controller-manager:

1. [Time: 0] Scheduled the job pod on build cluster.
2. [Time: 2s] Didn't find the pod schedule at step 1.
3. [Time: 2s] Recreated the pod.

The pod scheduled at step 3 failed with `pod already exist`. It's very likely that the pod created in step 1 became available right after the recreating of the pod. 

**What you expected to happen**:

Prow-controller-manager should be more strict on rescheduling the pod

alvaroaleman:
So what should have happened on step 1 is that plank blocks until it can see the pod (AKA it arrived back in its cache): https://github.com/kubernetes/test-infra/blob/827b4f4f702ae7435abf992709a9f22e82e0251a/prow/plank/reconciler.go#L625-L626

Did that time out?

chaodaiG:
yeah it timed out in 2 seconds:
```
jsonPayload: {
component: "prow-controller-manager"
controller: "plank"
level: "error"
error: "error starting pod: failed waiting for new pod test-pods/039bc087-a12d-11eb-9175-de64ce24b86b in cluster build  appear in cache: timed out waiting for the condition"
msg: "Reconciliation failed"
file: "prow/plank/reconciler.go:202"
name: "039bc087-a12d-11eb-9175-de64ce24b86b"
func: "k8s.io/test-infra/prow/plank.(*reconciler).defaultReconcile"
}
```

alvaroaleman:
Hm okay, maybe bump it to say 10 seconds or something? That should absolutely be enough and if not there is maybe an issue with your cluster.

chaodaiG:
> Hm okay, maybe bump it to say 10 seconds or something? That should absolutely be enough and if not there is maybe an issue with your cluster.

I think that would be good. Since majority of times the performance is the same as of today, and it would guard against this bug as well under certain circumstances

PR #21893, opened 2021-04-21
Plank wait for 10 seconds for a pod to come up

chaodaiG:
Fixes: #21869

Majority of times the performance is the same as of today, and it would guard against this bug as well under certain circumstances


alvaroaleman:
lgtm modulo the invalid commit message

alvaroaleman (approved):

PR #21895, opened 2021-04-21
Hmac tool: add more debug messages for easier debugging

chaodaiG:


cjwagner (approved):
FWIW I think the error we're seeing is coming from here so I don't think these debugging messages will give us any more info with the problem we're looking into now. https://github.com/kubernetes/test-infra/blob/5f1e2e6921bcc1e4244b13bf0e2ece7257bb4abb/prow/cmd/hmac/main.go#L220-L222

chaodaiG:
/unhold
Let's try it since it's only used by hmac tool and this part of function is already buggy

chaodaiG:
/hold
@cjwagner  is correct, let me see if I can add logs at correct place

cjwagner (approved):

chaodaiG:
Allowing `unaffiliated` state so that hmac tool can go further
/unhold

Issue #21563, opened 2021-03-29
Empty coverage lens


amwat:
**What happened**:
Empty coverage lens

**What you expected to happen**:
Coverage lens shows coverage treemap

**How to reproduce it (as minimally and precisely as possible)**:
already happening live in prow.k8s.io

**Please provide links to example occurrences, if any**:
https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/ci-kubernetes-coverage-unit/1346830331453378560

**Anything else we need to know?**:
I don't have access to prow.k8s.io logs, but this is also happening in other prows
```
{component: deck, file: prow/simplifypath/simplify.go:48, func: k8s.io/test-infra/prow/simplifypath.(*simplifier).Simplify, level: debug, msg: Path not handled. This is a bug, please open an issue against the kubernetes/test-infra repository with this error message., path: /static/spyglass/lens.css}
```

/cc @michelle192837 

fyi @stevekuznetsov @BenTheElder

cjwagner:
I think the `simplifypath` errors are only related to metrics so that probably isn't the cause of this problem.
The prow.k8s.io logs don't include anything from the file `prow/spyglass/lenses/coverage/coverage.go` in the past week. The HTML template seems to execute without error, but I did notice this line was missing: https://github.com/kubernetes/test-infra/blob/f7e21a3c18f4f4bbc7ee170675ed53e4544a0632/prow/spyglass/lenses/coverage/template.html#L9

When I inspect the spyglass page I see multiple occurences of the following error in the console (seems like 1 occurrence per spyglass lens on the page).

> ​Uncaught SecurityError: Sandbox access violation: Blocked a frame at "null" from accessing a frame at "https://prow.k8s.io".  The frame requesting access is sandboxed and lacks the "allow-same-origin" flag.

I also see this error, but only on the spyglass page with the coverage lens:

> Uncaught SyntaxError: Cannot use import statement outside a module     script_bundle.min.js:1

This seems like a likely culprit, but I don't know how to find where that import statement would be (`script_bundle.min.js:1` is not very helpful).

chaodaiG:
@amwat , do you have a working example?

amwat:
the link in the first issue comment still fails for me

chaodaiG:
> the link in the first issue comment still fails for me

Same here. Just want to see whether this is failing for all spyglass pages that use coverage lens or just this specific instance

amwat:
[...]

PR #21897, opened 2021-04-21
re-enable workspace root importfor coverage lense

BenTheElder:
see: https://github.com/kubernetes/test-infra/issues/21563, https://github.com/bazelbuild/rules_nodejs/wiki#no-longer-link-the-workspace-root

### DIFF
--- a/prow/spyglass/lenses/coverage/BUILD.bazel
+++ b/prow/spyglass/lenses/coverage/BUILD.bazel
@@ -31,6 +31,7 @@ ts_library(
 rollup_bundle(
     name = "script_bundle",
     entry_point = ":coverage.ts",
+    link_workspace_root = True,
     deps = [
         ":script",
         "//gopherage/cmd/html/static:parser",


BenTheElder:
needs testing, but theoretically this seems to bring back the previously relied upon behavior before we upgraded from 0.36 to 3.1

BenTheElder:
unable to verify this change currently due to https://github.com/bazelbuild/bazel/issues/13379 + python (2.x) being removed from our workstations. i'll look into a workaround

BenTheElder:
curling `spyglass/static/coverage/script_bundle.min.js` contains the minified `parseCoverage` method exported by the imported parser module, whereas the one on prow.k8s.io has a literal `import {parseCoverage} from "io_k8s_test_infra/gopherage/cmd/html/static/parser";` presumably that wasn't resolved at build time and won't work.

this should fix the issue.

amwat:
/lgtm
/approve

was able to verify this locally

PR #21904, opened 2021-04-22
cherrypicker: remove useless assert error msg of test

hi-rustin:


matthyx (approved):
it's not an assert, but a hint about what's missing... but ok it can go
overall the `external-plugins` would require some cleanups imho

cjwagner:
What is the motivation for removing this hint? It seems preferable to keep it to me, but it doesn't seem to be very important either way.

hi-rustin:
> What is the motivation for removing this hint? It seems preferable to keep it to me, but it doesn't seem to be very important either way.

Because this is an invalid output, it should not be release-1.5. And this information does not help it.

cjwagner (approved):

PR #21914, opened 2021-04-22
Make Cluster API make-main target optional

CecileRobertMichon:
After moving golangci-lint to github actions, which have an internal cache and can run in parallel, we can move this target to optional and then removing it a bit later.

Re-apply https://github.com/kubernetes/test-infra/pull/21524 now that the github action is running on every PR.

/assign @vincepri


vincepri (approved):
/approve
/lgtm

PR #21923, opened 2021-04-23
Add do-not-merge/flake-free-test-needed for Conformance Promotion PR's

Riaankl:
**What this PR does / why we need it:**
To increase participation on Conformance Promotion PR's.
The Promotion PR will be created directly after the e2e test PR merge and the `do-not-merge/flake-free-test-needed` label will be added manually.

This increases the exposure time of the Promotion PR in the community before merging.

The label is added and removed by humans with no automation currently.

**Fixes:**
#22063

/sig testing
/sig architecture
/area conformance

### DIFF
--- a/label_sync/labels.md
+++ b/label_sync/labels.md
@@ -354,6 +354,7 @@ larger set of contributors to apply/remove them.
 | Name | Description | Added By | Prow Plugin |
 | ---- | ----------- | -------- | --- |
 | <a id="do-not-merge/contains-merge-commits" href="#do-not-merge/contains-merge-commits">`do-not-merge/contains-merge-commits`</a> | Indicates a PR which contains merge commits.| prow |  [mergecommitblocker](https://git.k8s.io/test-infra/prow/plugins/mergecommitblocker) |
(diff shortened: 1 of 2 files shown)


fejta:
Is there some issue that this is associated with?

nikhita:
/hold
needs to be notified broadly

> Is there some issue that this is associated with?

+1 with the contribex hat on. Has this been discussed somewhere? :)

nikhita:
> Should I create an issue to go with the PR?

Yes, please create an issue with details on how the label should be applied. If automatically, what conditions should be satisfied, etc.

> I can take it to next week's Conformance and SIG Arch meetings.
Any other places you think we should get buy-in?

Thanks, conformance and SIG Arch sound perfect :)

dims:
@Riaankl are there bot commands to add and remove this label?

stevekuznetsov:
We chatted about this in the SIG Testing meeting today, some thoughts:

 - adding the label to PRs changing the file should be possible today with the existing OWNERS label mechanism
 - requiring users to remove the label by using the GitHub UI would require that those members have write access to the repo, which is rare
 -  Riaan was hoping that only OWNERS in the file would be allowed to remove the label, which seems like a reasonable approach but would require new code to be written (in a Prow plugin?) for this mechanism

Would love to hear your input on this @spiffxp  :)

PR #21925, opened 2021-04-23
disable testing on eastus2euap region

andyzhangx:
/hold


feiskyer:
any reasons to disable them?

andyzhangx:
> any reasons to disable them?

the donated CNCF subs does not support canary region after filing a support ticket, so this test never works in canary.

feiskyer:
@andyzhangx do we still need this?

PR #21926, opened 2021-04-23
[phaino] format env vars properly for docker

rmohr:
Env variables passed to docker should be formated like

```
-e ENV1=val1 -e ENV2=val2
```

instead of

```
-e ENV1:val1 -e ENV2:val2
```

This  fixes an issue where env variables defined on the containers were not passed through.

Signed-off-by: Roman Mohr <rmohr@redhat.com>


chaodaiG:
Appears to be a bug introduced in https://github.com/kubernetes/test-infra/pull/21020, thank you @rmohr for the fix!

/ok-to-test
/lgtm
/approve

PR #21932, opened 2021-04-23
Stop deploying tot component.

cjwagner:
From `tot` logs: ``` "Invalid options: --config-path is mandatory" ```
`tot` is not mandatory or used any more so lets delete it. I'll delete the service and deployment from the cluster after this merges.

/assign @fejta


fejta (approved):
/hold

cjwagner:
Deployment and service have been deleted.

PR #21936, opened 2021-04-24
Revert "skip e2e test with hostNetwork pods"

prameshj:
Reverts kubernetes/test-infra#20224

We have now changed the port number for the e2e test in https://github.com/kubernetes/kubernetes/pull/101381

RE-enabling the hostNetwork tests to run with NodeLocalDNS enabled.


aojea:
/lgtm
/approve

🤞

aojea:
it is working @prameshj 
https://testgrid.k8s.io/sig-network-gce#gci-gce-coredns-nodecache&include-filter-by-regex=hostNetwork
:clap:

PR #21938, opened 2021-04-24
instrumentation: expose a nicer interface for memory profiles

stevekuznetsov:
memory profile analysis: better parsing

Sometimes, nodes are dropped from the graph and a line like the
following exists in the preamble:

Dropped 62 nodes (cum <= 4.86MB)

In these cases, the `[5:]` was naieve - instead, we can look for the
headers that precede the data.

---

instrumentation: expose a nicer interface for memory profiles

Periodically sampling the memory profile of a process is an extremely
useful tool for undertanding usage. This patch extracts the prior work
for `sidecar` and allows it to be used in any binary that has
instrumentation options.

---

/assign @alvaroaleman


stevekuznetsov:
note: I had to pull out the decoration-dependent validation into a separate package to break an import cycle

ghproxy/ghproxy.go
```diff
@@ -155,6 +155,9 @@ func main() {
 	}
 
 	pjutil.ServePProf(o.instrumentationOptions.PProfPort)
+	if o.instrumentationOptions.ProfileMemory {
```
alvaroaleman, line 158: Lets just move this and `pjutil.ServePprof` into a method of the instrumentation options (`Instrument`?) instead of repeating it everywhere?

stevekuznetsov: Yeah, we can do that too.

stevekuznetsov:
I might have to abandon this particular approach, the cycle refactor is impossible. yikes.

stevekuznetsov:
OK. @alvaroaleman ready for review.

prow/cmd/pipeline/main.go
```diff
@@ -28,7 +28,9 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
+	"k8s.io/test-infra/prow/pjutil/pprof"
 
+	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp" // support gcp users in .kube/config
```
alvaroaleman, line 33: How you manage to always get imports wrong is beyond me lol

alvaroaleman (approved):

PR #21940, opened 2021-04-25
Simplifypath: Add greedy option and use it

alvaroaleman:
We need to be able to match greedily to cope with label names that have a slash in them. Additionally it is useful for cases where we don't actually care after a given element, for example the static assets in deck/spyglass.


stevekuznetsov (approved):
Nice, this is a super clean way to handle the issue.

PR #21943, opened 2021-04-25
Enable gosimple linter

alvaroaleman:
It requires to:
* Replace literal regex strings in `"` to use backticks instead so backslashes inside do not need double escaping
* Replacing loops that append from a slice to another slice to be replaced with `toAppendSlice = append(toAppendSlice, otherSlice...)`
* `time.TIme.Sub(time.Now())` to be replaced with `time.Until(time.Time)`
* Variable declarations that are immediately followed by variable assignments to use the short form of doing both (`a := val`)

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -19,6 +19,7 @@ linters:
   - misspell
   - ineffassign
   - staticcheck
+  - gosimple
   disable-all: true
 
 issues:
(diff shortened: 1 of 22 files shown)


chaodaiG:
LGTM, please fix merge conflict and I'll approve

chaodaiG (approved):

BenTheElder (commented):
LGTM

PR #21944, opened 2021-04-25
Enable staticcheck linter

alvaroaleman:
Prevents among others:
* Variables that are appended to but never read from
* Shadowed+dropped errors, most notably in a bunch of filepath.Read
* Invalid xml struct tags
* strings.TrimLeft used as strings.TrimPrefix, which is wrong


stevekuznetsov:
All of this seems reasonable to me
/approve
/lgtm

PR #21952, opened 2021-04-26
Refactor generic-autobumper: separate prcreator as a standalone package

chaodaiG:
So that prcreator package can be used by other tools.


chaodaiG:
The current implementation keeps https://github.com/kubernetes/test-infra/tree/master/prow/cmd/generic-autobumper/bumper, for backward compatibility purpose, especially for downstream consumer of `bumper` package.  This refactor changes the signature of some previously exposed method, which is not very straightforward to avoid this, for example https://github.com/kubernetes/test-infra/pull/21952/files#diff-1155bc13648e69007570e6f4cf58e67dacbae1702366f44ecc75703dd7145791R400.

Also a note for code reviewers: the unit tests are missing, will add once we agree that this is the direction to go

prow/cmd/generic-autobumper/prcreator/prcreator.go
```diff
@@ -0,0 +1,712 @@
...
+	Gerrit *Gerrit `yaml:"gerrit"`
+	// The name used in the address when creating remote. This should be the same name as the fork. If fork does not exist this will be the name of the fork that is created.
+	// If it is not the same as the fork, the robot will change the name of the fork to this. Format will be git@github.com:{GitLogin}/{RemoteName}.git
+	RemoteName string `yaml:"remoteName"`
+	// The name of the branch that will be used when creating the pull request. If unset, defaults to "autobump".
+	HeadBranchName string `yaml:"headBranchName"`
+	// Optional list of labels to add to the bump PR
+	Labels []string `yaml:"labels"`
+	// Dryrun skips PR creation
+	Dryrun bool `yaml:"dryrun"`
```
mpherman2, line 104: This is a redundant option as we already have SkipPullRequest.

chaodaiG: Good point, the behavior of Dryrun is slightly different though, it still creates fork, does commit, just not git push or pullrequest creation, would you mind if I change the behavior of `SkipPullRequest`?

chaodaiG:
The error message said `bazel run @io_k8s_repo_infra//hack:update-deps` could fix, I ran it locally but it didn't make any difference, might have run into bazel cache, will see how to make it work

chaodaiG:
> The error message said `bazel run @io_k8s_repo_infra//hack:update-deps` could fix, I ran it locally but it didn't make any difference, might have run into bazel cache, will see how to make it work

`bazel run @io_k8s_repo_infra//hack:update-deps` made a difference on linux machine, not sure what's going on, but at least this error should be fixed

prow/cmd/generic-autobumper/prcreator/prcreator.go
mpherman2: Yea that is fine. I think this is mostly used for testing anyway

chaodaiG:
> The current implementation keeps https://github.com/kubernetes/test-infra/tree/master/prow/cmd/generic-autobumper/bumper, for backward compatibility purpose, especially for downstream consumer of `bumper` package. This refactor changes the signature of some previously exposed method, which is not very straightforward to avoid this, for example https://github.com/kubernetes/test-infra/pull/21952/files#diff-1155bc13648e69007570e6f4cf58e67dacbae1702366f44ecc75703dd7145791R400.
> 
> Also a note for code reviewers: the unit tests are missing, will add once we agree that this is the direction to go

@alvaroaleman please take a look, especially around the idea of keeping exposed methods compatible with openshift

alvaroaleman:
> @alvaroaleman please take a look, especially around the idea of keeping exposed methods compatible with openshift

Can you give me a day or two to check if we can just use the autobumper tool? I _think_ that should work and would be preferrable for everyone.

We also use the bumper package in a bunch of other places, but that is very organically grown and inconsistent, so collecting our usecases (I _think_ its mostly: "run command", "git commit if changes", "upsert PR") and using it as input for your work here (and then using it downstream as well) and ideally ending up with a minimal public api would make things better, too.

chaodaiG:
> Can you give me a day or two to check if we can just use the autobumper tool? I _think_ that should work and would be preferrable for everyone.

Not big rush on this work, so please take your time.
 
> We also use the bumper package in a bunch of other places, but that is very organically grown and inconsistent, so collecting our usecases (I _think_ its mostly: "run command", "git commit if changes", "upsert PR") and using it as input for your work here (and then using it downstream as well) and ideally ending up with a minimal public api would make things better, too.

I can see that there are something in common here, the purpose of this PR is to create a `prcreator` package, that takes "run command" function as an argument, and create PR if there is anything changed

alvaroaleman:
Ok, so to close the loop here:
* We using the upstream bumper now to bump prow
* Our current usage after that for the bumper package is:
```
rg -F bumper. pkg/ cmd/|sed -e 's/.*bumper\.//g' -e 's/{.*//g' -e 's/(.*//g'|sort|uniq -c
      7 Call
      2 GitAuthorOptions
      4 GitCommitAndPush
      1 GitCommitSignoffAndPush
      1 GitPush
      5 HasChanges
     13 HideSecretsWriter
      6 UpdatePullRequestWithLabels
```

> I can see that there are something in common here, the purpose of this PR is to create a prcreator package, that takes "run command" function as an argument, and create PR if there is anything changed

That sounds almost ideal, but:
* We also need the possibility to run zero commands to use this as a lib
* We also need the possibility to run multiple commands and commit after each

Maybe instead of giving the thing a `run command` function, it could take a slice of `type CommandDef struct { command []string; commitMsg string}` and then:
* If command is not given, don't run any command and only commit (if changes)
* If commitMsg is not given, use the command as commit msg

WDYT?

chaodaiG:
> Maybe instead of giving the thing a `run command` function, it could take a slice of `type CommandDef struct { command []string; commitMsg string}` and then:
> 
> * If command is not given, don't run any command and only commit (if changes)
> * If commitMsg is not given, use the command as commit msg

This is better than what's in this PR, so yeah!
One thing worth noting is:
* There might be more complicated scenarios than a few commands, so a slice of `[]func() (err error, commitMsg string)` maybe?

+ @mpherman2  for thoughts or concerns here

alvaroaleman:
> There might be more complicated scenarios than a few commands, so a slice of []func() (err error, commitMsg string) maybe?

Oh yeah, that is probably even better!

chaodaiG:
/uncc @alvaroaleman @cjwagner @mpherman2 
Will add back once it's ready

chaodaiG:
It's ready for review now.

/assign @alvaroaleman @mpherman2 
/cc @cjwagner

prow/cmd/generic-autobumper/bumper/bumper.go
```diff
@@ -477,24 +454,33 @@ func (w HideSecretsWriter) Write(content []byte) (int, error) {
...
 // UpdatePullRequest updates with github client "gc" the PR of github repo org/repo
-// with "title" and "body" of PR matching author and headBranch from "source" to "baseBranch"
-func UpdatePullRequest(gc github.Client, org, repo, title, body, source, baseBranch, headBranch string, allowMods bool) error {
-	return UpdatePullRequestWithLabels(gc, org, repo, title, body, source, baseBranch, headBranch, allowMods, nil)
+// with "title" and "body" of PR matching author and headBranch from "source" to
+// "baseBranch"
+//
+// Keeping for backward compatibility reason:
+// https://github.com/kubernetes/test-infra/pull/21952#issuecomment-829562870
+func UpdatePullRequest(gc github.Client, org, repo, title, body, source, baseBranch, headBranch string, allowMods bool, dryrun bool) error {
```
chaodaiG, line 470: Added a `dryrun` switch here which breaks the exposed signature. @alvaroaleman , if this is not desired I can try to set it a global var instead

prow/cmd/generic-autobumper/main.go
```diff
@@ -19,47 +19,467 @@ package main
...
-	var o bumper.Options
+	var o Options
+	var pro bumper.Options
 	data, err := ioutil.ReadFile(config)
 	if err != nil {
-		return nil, fmt.Errorf("Failed to read in config file, %s", config)
+		return nil, nil, fmt.Errorf("Failed to read in config file, %s", config)
+	}
+
+	err = yaml.Unmarshal(data, &o)
```
alvaroaleman, line 143: Wouldn't it be more straight-forward to embedd `bumper.Options` into `Options` rather than unmarshaling the same data into two types?

alvaroaleman (commented):
So I don't think I have any conceptional issues with this, but this is just too big to properly review it at once, could we maybe split this up?

Maybe something like:
1. Add the reusable lib for creating PRs, copying code as needed
1. Start using it in the prow bumper
2. Deduplicate the bumper config with the one from the lib

chaodaiG:
First PR of breakdowns: https://github.com/kubernetes/test-infra/pull/22097

prow/cmd/generic-autobumper/main.go
```diff
@@ -267,16 +307,13 @@ func parseUpstreamImageVersion(upstreamAddress, prefix string) (string, error) {
...
-		if p != "" && strings.HasPrefix(name, p) {
-			return true
+	for _, line := range strings.Split(strings.TrimSuffix(string(body), "\n"), "\n") {
+		res := imageMatcher.FindStringSubmatch(string(line))
+		if len(res) > 2 && strings.Contains(res[1], prefix) {
+			return res[2], nil
 		}
 	}
-	return false
+	return "", fmt.Errorf("unable to find match for %s in upstream refConfigFile", prefix)
```
fejta, line 347: IMO this is equivalent to `return "", errors.New("not found")` as all of the information in here is trivially known to the caller and/or incorrect:
* prefix is given to the caller
* upstream is both part of the function name as well as the variable that fetches this information

So I would expect the caller to be the one to decide whether to log its information:

```golang
if tag, err := parseUpstreamImageVersion(url, prefix); err != nil {
  return fmt.Errorf("failed to parse %s for %q: %w", url, prefix, err)
}
```

prow/cmd/generic-autobumper/main.go
```diff
@@ -309,35 +346,32 @@ func getVersionsAndCheckConsistency(prefixes []Prefix, images map[string]string)
...
-// and returns a summary of what was bumped for use in the commit message
-func makeCommitSummary(prefixes []Prefix, versions map[string][]string) string {
 	if len(versions) == 0 {
-		return fmt.Sprintf("Update %s images as necessary", getPrefixesString(prefixes))
+		return fmt.Sprintf("Update %s images as necessary", strings.Join(allPrefixes, ", "))
 	}
 	var inconsistentBumps []string
 	var consistentBumps []string
+	// isBumpedPrefix takes a prefix and a map of new tags resulted from bumping
+	// : the images using those tags and itterates over the map to find if the
```
fejta: iterates

fejta (approved):
Basically LGTM but I think we should fix github.com name (or maybe that isn't a regression?)
/hold

prow/cmd/generic-autobumper/bumper/bumper.go
```diff
@@ -357,19 +277,48 @@ func Run(o *Options) error {
 			}
 		}
 
-		if err := MakeGitCommit(fmt.Sprintf("https://%s:%s@github.com/%s/%s.git", o.GitHubLogin, string(sa.GetTokenGenerator(o.GitHubToken)()), o.GitHubLogin, o.RemoteName), o.HeadBranchName, o.GitName, o.GitEmail, o.Prefixes, stdout, stderr, versions); err != nil {
+		// Make change, commit and push
+		for _, f := range prh.Changes() {
```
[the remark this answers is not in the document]
chaodaiG: Addressed in https://github.com/kubernetes/test-infra/pull/22112

prow/cmd/generic-autobumper/bumper/bumper.go
```diff
@@ -357,19 +277,48 @@ func Run(o *Options) error {
...
+			msg, err := f()
+			if err != nil {
+				return fmt.Errorf("failed to process provided function: %w", err)
```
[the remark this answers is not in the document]
chaodaiG: Addressed in #22112

prow/cmd/generic-autobumper/bumper/bumper.go
```diff
@@ -357,19 +277,48 @@ func Run(o *Options) error {
...
+			}
+
+			changed, err := HasChanges()
+			if err != nil {
+				return fmt.Errorf("error occurred when checking changes: %w", err)
```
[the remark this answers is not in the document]
chaodaiG: Addressed in #22112

prow/cmd/generic-autobumper/bumper/bumper.go
```diff
@@ -357,19 +277,48 @@ func Run(o *Options) error {
...
+				logrus.Info("Nothing changed, exiting ...")
+				return nil
+			}
+
+			if err := gitCommit(o.GitName, o.GitEmail, msg, stdout, stderr, false); err != nil {
+				return fmt.Errorf("failed git commit: %w", err)
+			}
+		}
+
+		if err := gitPush(fmt.Sprintf("git@github.com:%s/%s.git", o.GitHubLogin, o.RemoteName),
```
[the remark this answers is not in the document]
chaodaiG: This package supports only github and gerrit so far, and this part is nested under `if o.Gerrit == nil {`, so it's intended behavior

prow/cmd/generic-autobumper/bumper/bumper.go
```diff
@@ -357,19 +277,48 @@ func Run(o *Options) error {
...
 		if o.GitHubBaseBranch == "" {
 			repo, err := gc.GetRepo(o.GitHubOrg, o.GitHubRepo)
 			if err != nil {
 				return fmt.Errorf("failed to detect default remote branch for %s/%s: %w", o.GitHubOrg, o.GitHubRepo, err)
 			}
 			o.GitHubBaseBranch = repo.DefaultBranch
 		}
-
-		if err := updatePRWithLabels(gc, o.GitHubOrg, o.GitHubRepo, images, getAssignment(o.AssignTo, o.OncallAddress, o.OncallGroup), o.GitHubLogin, o.GitHubBaseBranch, o.HeadBranchName, updater.PreventMods, o.Prefixes, versions, o.Labels); err != nil {
+		if err := updatePRWithLabels(gc, o.GitHubOrg, o.GitHubRepo,
```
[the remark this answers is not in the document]
chaodaiG: Addressed in #22112, went with one line

prow/cmd/generic-autobumper/helper_test.go
```diff
@@ -0,0 +1,99 @@
+/*
+Copyright 2019 The Kubernetes Authors.
```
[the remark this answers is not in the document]
chaodaiG: Addressed in #22112

chaodaiG (commented):
Sorry my bad, should have closed this PR to avoid confusion.

This PR is broken down into several smaller PRs:
- create reusable functions for pr creation: https://github.com/kubernetes/test-infra/pull/22097
- generic-autobumper uses pr creator functions: https://github.com/kubernetes/test-infra/pull/22112
- clean up PR (to be added)

Luckily all the latest PR comments also applies on the broken down PRs, have addressed them in #22112 

Closing this PR to avoid confusion

/close

PR #21953, opened 2021-04-26
Azure: Set LOCAL_ONLY as part of anonymous pull preset

CecileRobertMichon:
When using the `preset-azure-anonymous-pull` preset which sets `REGISTRY` to "capzci.azurecr.io", we should always set `LOCAL_ONLY` to false. This PR adds it to the preset so we don't have to add it to every single job that uses the preset.

/assign @chewong @devigned


chewong (commented):
/lgtm

PR #21958, opened 2021-04-26
Always report status when pending, crier doesnt know when the the link for the jen…

arturo-skydio:
…kins job is available and is not able to post pending

Currently, if `--skip-report` setting is turned on in jenkins-operator, crier is not able to set the link to jenkins when it is available.  There maybe a way to tell crier to report now that we have the jenkins job url, but couldn't find a way there


petr-muller:
/ok-to-test

Have no idea about jenkins-operator but let's get this tested

PR #21972, opened 2021-04-27
Abort batch jobs that are invalid because of PR head changed or left pool

arturo-skydio:
Any reason to keep running presubmit jobs for invalid batch jobs?, I couldn't figure out the logic here, let me know if there is something obvious I'm missing

### DIFF
--- a/prow/tide/tide.go
+++ b/prow/tide/tide.go
@@ -867,18 +867,23 @@ func (c *Controller) accumulateBatch(sp subpool) (successBatch []PullRequest, pe
 					state.prs = append(state.prs, pr)
 				} else if !ok {
 					state.validPulls = false
-					sp.log.WithField("batch", ref).WithFields(pr.logFields()).Debug("batch job invalid, PR left pool")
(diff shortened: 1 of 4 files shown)


prow/tide/tide.go
```diff
@@ -867,18 +867,23 @@ func (c *Controller) accumulateBatch(sp subpool) (successBatch []PullRequest, pe
...
 				}
 			}
 			states[ref] = state
 		}
 		if !states[ref].validPulls {
-			// The batch contains a PR ref that has changed. Skip it.
+			// The batch contains a PR ref that has changed. Skip it and set state to abort.
+			prevPJ := pj.DeepCopy()
+			pj.Status.State = prowapi.AbortedState
+			if err := c.prowJobClient.Patch(c.ctx, pj.DeepCopy(), ctrlruntimeclient.MergeFrom(prevPJ)); err != nil {
```
alvaroaleman, line 884: The `DeepCopy()` here is not needed. Also, please add tests.

PR #21977, opened 2021-04-27
Gerrit presubmit reports will not vote negative after merge

mpherman2:
This is a fix for log spam when presubmit jobs fail after PR is already merged. Gerrit does not allow negative votes after merge.
Pair Programmed with @chaodaiG 
/assign @fejta


fejta (approved):
/hold

Thanks!

Issue #21980, opened 2021-04-27
Warning logs in prow: /job-history 'object doesn't exist'


chaodaiG:
**What happened**:

Recurring warning logs in prow like:

```
jsonPayload: {
file: "prow/cmd/deck/main.go:887"
handler: "/job-history"
url: "/job-history/gs/kubernetes-jenkins/pr-logs/directory/pull-cluster-api-verify-external-links"
component: "deck"
level: "warning"
msg: "failed to get job history: failed to locate build data: failed to read pr-logs/directory/pull-cluster-api-verify-external-links/latest-build.txt: creating reader for object pr-logs/directory/pull-cluster-api-verify-external-links/latest-build.txt: storage: object doesn't exist"
func: "main.handleJobHistory.func1"
}
```

This is happening pretty frequently, roughly in the order of ~10 per hour.

**What you expected to happen**:

If this is a bug we should fix it, otherwise these are not warning logs

spiffxp:
/sig testing
@chaodaiG is this still happening?

chaodaiG:
haven't seen it lately
/close

PR #21981, opened 2021-04-27
Spyglass job history page error: correctly detecting http response code

chaodaiG:


prow/cmd/deck/main_test.go
```diff
@@ -1189,6 +1189,14 @@ func TestHttpStatusForError(t *testing.T) {
 			},
 			expectedStatus: http.StatusGone,
 		},
+		{
```
chaodaiG, line 1192: The current way of retrieving http response code always return 5XX for any wrapped errors, see this test case

chaodaiG:
Fixes: https://github.com/kubernetes/test-infra/issues/21980

prow/cmd/deck/main_test.go
alvaroaleman: Shouldn't we fix that then?

chaodaiG: Agree that's the best approach, thought about it for a few minutes but didn't have a good way. Will update here if I can find a better way. Also let me know if you have an idea

alvaroaleman:
```
diff --git a/prow/cmd/deck/main.go b/prow/cmd/deck/main.go
index dec42e1b0e..94ad16678c 100644
--- a/prow/cmd/deck/main.go
+++ b/prow/cmd/deck/main.go
@@ -1651,7 +1651,8 @@ type httpError struct {
 }
 
 func httpStatusForError(e error) int {
-       if httpErr, ok := e.(httpError); ok {
+       var httpErr httpError
+       if ok := errors.As(e, &httpErr); ok {
                return httpErr.statusCode
        }
        return http.StatusInternalServerErro
```

chaodaiG: Awesome, good to learn! I'll give it a shot

chaodaiG: Done. And btw replaced a few `%v` to `%w`

chaodaiG: Also updated both PR title and commit msg

alvaroaleman: Thanks. As an aside, we have verification in our downstream repo to enforce error wrapping: https://github.com/openshift/ci-tools/blob/9f710b04bc1319b5374ea7ef26a8dd1e8dd97f8b/hack/lint.sh#L24

And also something to fix it up automatically that works okayish: https://github.com/openshift/ci-tools/blob/9f710b04bc1319b5374ea7ef26a8dd1e8dd97f8b/hack/fixup-error-wrapping.sh#L10

However, I do not know how to integrate new stuff with bazel :upside_down_face:

alvaroaleman (approved):

Issue #21983, opened 2021-04-27
Flooding warnings in tide: BranchProtectionWarnings


chaodaiG:
**What happened**:

Tide runs every minute, it emits the same warning every time it runs:

```
jsonPayload: {
func: "k8s.io/test-infra/prow/config.(*Config).BranchProtectionWarnings"
file: "prow/config/branch_protection.go:451"
component: "tide"
msg: "The following repos define a policy or require context(s), but have one or more branches with protect: false"
controller: "sync"
level: "warning"
repos: "kubernetes-sigs/aws-ebs-csi-driver=gh-pages,kubernetes-sigs/aws-efs-csi-driver=gh-pages,kubernetes-sigs/aws-fsx-csi-driver=gh-pages,kubernetes-sigs/nfs-subdir-external-provisioner=gh-pages,kubernetes-sigs/node-feature-discovery-operator=gh-pages,kubernetes-sigs/node-feature-discovery=gh-pages,kubernetes/cloud-provider-aws=gh-pages,kubernetes/cloud-provider-openstack=gh-pages,kubernetes/ingress-nginx=gh-pages,kubernetes/kube-state-metrics=gh-pages"
}
```

**What you expected to happen**:

This is not an actionable warning for prow operator, doesn't make much sense alert every minute.

chaodaiG:
For example in https://github.com/kubernetes/test-infra/pull/20157, it's intended not to protect aws-ebs-csi-driver gh-pages branch

PR #21984, opened 2021-04-27
Tide branch protection warnings turn to debug

chaodaiG:
These are pretty much intended behavior from users, not worth log as warnings

Fixes https://github.com/kubernetes/test-infra/issues/21983

### DIFF
--- a/prow/config/branch_protection.go
+++ b/prow/config/branch_protection.go
@@ -444,11 +444,10 @@ func (c *Config) unprotectedBranches(presubmits map[string][]Presubmit) []string
 //     because any branches not explicitly specified in the configuration will be unprotected.
 func (c *Config) BranchProtectionWarnings(logger *logrus.Entry, presubmits map[string][]Presubmit) {
 	if warnings := c.reposWithDisabledPolicy(); len(warnings) > 0 {
-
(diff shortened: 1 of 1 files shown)


prow/config/branch_protection.go
```diff
@@ -444,11 +444,10 @@ func (c *Config) unprotectedBranches(presubmits map[string][]Presubmit) []string
 //     because any branches not explicitly specified in the configuration will be unprotected.
 func (c *Config) BranchProtectionWarnings(logger *logrus.Entry, presubmits map[string][]Presubmit) {
 	if warnings := c.reposWithDisabledPolicy(); len(warnings) > 0 {
-
-		logger.WithField("repos", strings.Join(warnings, ",")).Warn("The following repos define a policy, but have protect: false")
```
alvaroaleman, line 448: Are you sure? AFAIK those are essentially misconfigurations
/cc @petr-muller

chaodaiG: yeah I believe so, see https://github.com/kubernetes/test-infra/issues/21983#issuecomment-827969398

petr-muller: I think the PR is correct, I think "policy" here is usually inferred from the presence of required jobs, and the fact that you have required jobs does not always mean you want to manage BP.

PR #21991, opened 2021-04-28
Phaino: add a new flag code-mount-path replacing gopath arg

chaodaiG:
The use of gopath can lead to the confusion of something related to golang, which is not very friendly with repos that don't primarily use golang

### DIFF
--- a/prow/cmd/phaino/local.go
+++ b/prow/cmd/phaino/local.go
@@ -45,9 +45,9 @@ const (
 	kubeconfigEnvKey = "KUBECONFIG"
 	// The path to mount the kubectl default config files to the container.
 	kubectlDefaultConfigMountPath = "/root/.kube"
-	// The default GOPATH in the container.
+	// The default code mount path in the container.
 	// To be consistent with https://github.com/kubernetes/test-infra/blob/19829768bb8ff2a9bb8de76e4dbcc1e520aaeb18/prow/pod-utils/decorate/podspec.go#L52
-	defaultGOPATH = "/home/prow/go"
+	defaultCodeMountpath = "/home/prow/go/src"
 )
 
 var baseArgs = []string{"docker", "run", "--rm=true"}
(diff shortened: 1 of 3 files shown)


prow/cmd/phaino/local.go
```diff
@@ -418,7 +418,7 @@ func (opts *options) resolveRefs(ctx context.Context, volumeMounts map[string]st
 		return scanln(ctx)
 	}
 
-	goSrcPath := filepath.Join(opts.gopath, "src")
+	goSrcPath := filepath.Join(opts.codeMountPath, "src")
```
chizhg, line 421: It'll become confusing to append `src` and use it as the final mount path.. Probably better to also ask users to add `src` as part of the --code-mount-path?

chaodaiG: good point, updated.

prow/cmd/phaino/main.go
```diff
@@ -66,9 +67,10 @@ func gatherOptions() options {
 	fs.DurationVar(&o.timeout, "timeout", time.Hour, "Maximum duration for each job (0 for unlimited)")
 	fs.DurationVar(&o.totalTimeout, "total-timeout", 0, "Maximum duration for all jobs (0 for unlimited)")
 	fs.DurationVar(&o.grace, "grace", 10*time.Second, "Terminate timed out jobs after this grace period (1s minimum)")
-	fs.StringVar(&o.gopath, "gopath", defaultGOPATH, "The GOPATH that is used in the container. "+
-		"Default is /home/prow/go, need to be changed if the repository depends on GOPATH and it's is set to a different value in the container.")
-
+	fs.StringVar(&o.gopath, "gopath", defaultCodeMountpath, "The path that is used in the container. "+
```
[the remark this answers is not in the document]
chaodaiG: Since this is a command line tool and not versioned, I might prefer keep the legacy arg until it's hard to maintain. Did update the PR with better handling of both args though

PR #21993, opened 2021-04-28
Inrepoconfig: allow explicitly unset empty global cluster

chaodaiG:


prow/config/config_test.go
```diff
@@ -2408,6 +2407,21 @@ in_repo_config:
...
 		},
+		{
+			name: "InRepoConfigAllowedClusters respects explicit empty default",
+			prowConfig: `
+in_repo_config:
+  allowed_clusters:
+    "*": []
+`,
+			verify: func(c *Config) error {
+				if len(c.InRepoConfig.AllowedClusters) != 1 ||
```
alvaroaleman, line 2418: This shouldn't tolerate one only zero or not?

chaodaiG: The value of `c.InRepoConfig.AllowedClusters` is `map["*"]{}`, did you mean we should explicitly make it nil?

alvaroaleman: ah wait, I just misread the diff and thought the first half of the condition was `c.InRepoConfig.AllowedClusters["*"]) != 1`

alvaroaleman (approved):

PR #21999, opened 2021-04-28
Branchprotection: Add a new unmanaged setting

alvaroaleman:
This is helpful for two reasons:
* protect: false changes its meaning depending on if we have a mandatory
  prowjob or not: If we do, it means "Do not touch bp settings", if we
  don't it means "Remove all Branchprotection config". This is extremely
  confusing
* The new setting avoids doing any api calls for unmanaged config

Contrary to all other branchprotection settings, this one can not be
overriden on a lower level, which is why a new checkconfig check was
added.

/cc @petr-muller

### DIFF
--- a/prow/cmd/branchprotector/protect_test.go
+++ b/prow/cmd/branchprotector/protect_test.go
@@ -24,6 +24,7 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/google/go-cmp/cmp"
 	"k8s.io/apimachinery/pkg/util/diff"
 	"sigs.k8s.io/yaml"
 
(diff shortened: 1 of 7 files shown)


prow/cmd/checkconfig/main.go
```diff
@@ -1220,6 +1229,54 @@ func validateAdditionalProwConfigIsInOrgRepoDirectoryStructure(root string, file
...
+			errs = append(errs, errors.New("branch protection is globally set to unmanaged, but has configuration"))
+		}
+		for org := range bp.Orgs {
+			errs = append(errs, fmt.Errorf("branch protection config is globally set to unmanaged but has configuration for org %s", org))
+		}
+	}
+	for orgName, orgConfig := range bp.Orgs {
+		if orgConfig.Unmanaged != nil && *orgConfig.Unmanaged {
+			if doesUnmanagedBranchprotectionPolicyHaveSettings(orgConfig.Policy) {
+				errs = append(errs, fmt.Errorf("branch protection config for rg %s is set to unmanaged, but it defines settings", orgName))
```
stevekuznetsov, line 1245:
```suggestion
				errs = append(errs, fmt.Errorf("branch protection config for org %s is set to unmanaged, but it defines settings", orgName))
```

stevekuznetsov (commented):
/cc @fejta

petr-muller (approved):
I like the decision to forbid the override on lower level, reduces the complexity. I *think* this should be fine but I will not be surprised if this breaks something once we start to use it.

I am sad that `protect: false` continues to have the ambiguous meaning but this is a great step forward.

petr-muller:
/hold
(to give Erick a chance)

alvaroaleman:
Last push is fixing an error message to reference the string `org` instead of `rg` after changing the code to the same effect

petr-muller (approved):

prow/cmd/branchprotector/protect_test.go
```diff
@@ -1162,6 +1163,87 @@ branch-protection:
 				},
 			},
 		},
+		{
+			name:     "Global unmanaged: true makes us not do anything",
+			branches: []string{"cfgdef/repo1=master", "cfgdef/repo1=branch", "cfgdef/repo2=master"},
+			config: `
+branch-protection:
+  unmanaged: true
```
fejta, line 1171: Isn't this specific scenario rather pointless? Shouldn't we handle this by just dropping the cfgdef org?

fejta (approved):
Tide now sets the context to passed before merging right?

If so then it seems like we can just make the tide context required, and never any of the jobs. Would that help make protect flag more coherent?

/hold

alvaroaleman:
> Tide now sets the context to passed before merging right?

> If so then it seems like we can just make the tide context required and never any of the jobs.

Yes, we could do that change now. Ppl however use the `Required` attribute on Github to infer if a job is required or optional. I am not sure if we can just change the branchprotector for everyone to only protect the `tide` status

> Would that help make protect flag more coherent?

No, because right now `protect: false`  with no mandatory job means "delete all branch protection" but with a mandatory job it means "do not manage branch protection". IMHO we should just deprecate and remove the flag, because its ambigious (that would require to first add another "disable_branch_protection" setting)

>  If so then it seems like we can just make the tide context required

prow/cmd/branchprotector/protect_test.go
alvaroaleman: The test verifies that the global unmanaged setting makes us not do anything. If we have no config, we wouldn't do anything regardless, so we need to have config.

From a practical point of view this config is pointless and the new checkconfig check would complain about it, yes.

PR #22000, opened 2021-04-28
kubernetes-csi: pass directory to test against via env variable

pohly:
The pull-test.sh script should be told where to find the additional
source code repo instead of having to guess.

/assign @msau42


pohly:
@msau42: still needs approval

PR #22004, opened 2021-04-28
Bumper: Use https to push to github

alvaroaleman:
The bumper currently both requires a github token with write access
(because it checks if a fork exists and creates one if that is not the
case) _and_ a ssh key which is used for pushing. This is needlessly
complicated, this change makes us use the the token and push via https
instead.

/assign @chaodaiG


chaodaiG (approved):
This is awesome! 

@mpherman2 for awareness

PR #22008, opened 2021-04-28
Add Tide support for multiple TargetURLs.

cjwagner:
Adds support for a `TargetURLs` map that is mutually exclusive with the existing `TargetURL` field. This will be used to make Tide status context links point to the correct domain when running a separate private deck instance.
This parallels the `PRStatusBaseURL` and `PRStatusBaseURLs` fields.

/assign @alvaroaleman @chaodaiG

### DIFF
--- a/prow/config/config.go
+++ b/prow/config/config.go
@@ -1943,6 +1943,17 @@ func parseProwConfig(c *Config) error {
 		return fmt.Errorf("tide has invalid max_goroutines (%d), it needs to be a positive number", c.Tide.MaxGoroutines)
 	}
 
+	if len(c.Tide.TargetURLs) > 0 && c.Tide.TargetURL != "" {
+		return fmt.Errorf("tide.target_url and tide.target_urls are mutually exclusive")
+	}
+
+	if c.Tide.TargetURLs == nil {
+		c.Tide.TargetURLs = map[string]string{}
+	}
+	if c.Tide.TargetURL != "" {
+		c.Tide.TargetURLs["*"] = c.Tide.TargetURL
+	}
+
 	if c.Tide.PRStatusBaseURLs == nil {
 		c.Tide.PRStatusBaseURLs = map[string]string{}
 	}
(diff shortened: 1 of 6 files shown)


prow/config/tide.go
```diff
@@ -224,6 +229,16 @@ func (t *Tide) GetPRStatusBaseURL(repo OrgRepo) string {
 	return t.PRStatusBaseURLs["*"]
 }
 
+func (t *Tide) GetTargetURL(repo OrgRepo) string {
+	if byOrgRepo, ok := t.TargetURLs[repo.String()]; ok {
+		return byOrgRepo
+	} else if byOrg, ok := t.TargetURLs[repo.Org]; ok {
```
chaodaiG, line 235: else is not needed since already returned

chaodaiG (approved):
/hold
For @alvaroaleman to take a look

Also left a nit

prow/config/config.go
```diff
@@ -1943,6 +1943,14 @@ func parseProwConfig(c *Config) error {
 		return fmt.Errorf("tide has invalid max_goroutines (%d), it needs to be a positive number", c.Tide.MaxGoroutines)
 	}
 
+	if len(c.Tide.TargetURLs) > 0 && c.Tide.TargetURL != "" {
+		return fmt.Errorf("tide.target_url and tide.target_urls are mutually exclusive")
+	}
+
+	if c.Tide.TargetURLs == nil {
+		c.Tide.TargetURLs = map[string]string{}
+	}
```
alvaroaleman, line 1952: I think you also want to copy c.Tide.TargetURL into ci.Tide.TargetURLs if the former is non-empty

cjwagner: Fixed and added unit tests.

prow/config/tide.go
cjwagner: Changed here and `GetPRStatusBaseURL`

chaodaiG (approved):

PR #22010, opened 2021-04-28
Remove awly from cloud-provider-gcp OWNERS

awly:


awly:
/assign @amwat 
/assign @jpbetz 
Ping @mikedanese

PR #22014, opened 2021-04-29
blob-csi-driver: add new prow job to test external e2e tests

boddumanohar:
with this https://github.com/kubernetes-sigs/blob-csi-driver/pull/395 PR, `blob-csi-driver` introduces kubernetes external e2e tests. To tests the external e2e tests, we will require a new job.

thanks,
Manohar.


andyzhangx (approved):
/lgtm
/approve

PR #22019, opened 2021-04-29
capz: add additional presets to conformance jobs

chewong:

Replacing `preset-azure-cred` with `preset-azure-cred-only` and `preset-azure-anonymous-pull` to prevent `AZURE_SSH_PUBLIC_KEY_FILE` from being overwritten 

/assign @CecileRobertMichon

### DIFF
--- a/config/jobs/kubernetes/sig-cloud-provider/azure/release-master.yaml
+++ b/config/jobs/kubernetes/sig-cloud-provider/azure/release-master.yaml
@@ -517,7 +517,7 @@ periodics:
   labels:
     preset-dind-enabled: "true"
     preset-kind-volume-mounts: "true"
-    preset-azure-cred: "true"
+    preset-azure-cred-only: "true"
   extra_refs:
   - org: kubernetes-sigs
     repo: cluster-api-provider-azure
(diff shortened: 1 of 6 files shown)


config/jobs/kubernetes/sig-cloud-provider/azure/release-master.yaml
```diff
@@ -517,7 +517,8 @@ periodics:
   labels:
     preset-dind-enabled: "true"
     preset-kind-volume-mounts: "true"
-    preset-azure-cred: "true"
+    preset-azure-cred-only: "true"
+    preset-azure-anonymous-pull: "true"
```
CecileRobertMichon, line 521: right now we're using a local registry for these jobs, do we need the external registry? Are we building k8s artifacts for these ones?

CecileRobertMichon: If you just want to remove the SSH var you can use `preset-azure-cred-only` without `preset-azure-anonymous-pull`

chewong: Switched back to local registry

PR #22020, opened 2021-04-29
capz: run HA conformance tests in serial

jackfrancis:
This PR changes the ginkgo configuration of capz HA control plane conformance tests so that they run in serial.


chewong (approved):
/lgtm

Issue #21736, opened 2021-04-09
Remove canonical dashboard


dims:
Need i say more? (see below)

![image](https://user-images.githubusercontent.com/23304/114195906-e382c980-991e-11eb-91aa-8179855848d5.png)

dims:
This is even worse! :( https://testgrid.k8s.io/sig-release-generated#Summary

BenTheElder:
cc @spiffxp

BenTheElder:
+1

BenTheElder:
I agree we should do this. but also
/priority backlog

dims:
cc @alxchalkias

dims:
Done!

PR #22023, opened 2021-04-29
Drop broken canonical/ubuntu stuff

dims:
Fix for https://github.com/kubernetes/test-infra/issues/21736


hasheddan (approved):
Thank you for cleaning this up @dims! Do we have history on these jobs? Sounds like we need a better process for regularly evaluating generated configs 👍
/lgtm

PR #22024, opened 2021-04-29
Autobumper: github noOp Changes do not cause PR

mpherman2:
There does not appear to be an easy way to check if there is an autobump will be the same as the current PR without rebasing the current PR. This change fetches the current PR, rebases it, and then does a diff. If the rebase fails, it aborts and just pushes the changes without checking if it is a noOp change.

### DIFF
--- a/prow/cmd/generic-autobumper/bumper/bumper.go
+++ b/prow/cmd/generic-autobumper/bumper/bumper.go
@@ -786,6 +786,54 @@ func GitCommitAndPush(remote, remoteBranch, name, email, message string, stdout,
 func GitCommitSignoffAndPush(remote, remoteBranch, name, email, message string, stdout, stderr io.Writer, signoff bool) error {
 	logrus.Info("Making git commit...")
 
+	if err := Call(stdout, stderr, gitCmd, "remote", "add", forkRemoteName, remote); err != nil {
+		return fmt.Errorf("failed to add remote: %w", err)
+	}
+	// Stash changes and create a branch the commit will live
+	if err := Call(stdout, stderr, gitCmd, "stash"); err != nil {
+		return fmt.Errorf("failed to git stash: %w", err)
+	}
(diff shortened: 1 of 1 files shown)


matthyx:
wow so much pain... is there a way to show and cover the improvement in bumper_test?
(probably not with the mocked githubs)

prow/cmd/generic-autobumper/bumper/bumper.go
```diff
@@ -786,44 +786,81 @@ func GitCommitAndPush(remote, remoteBranch, name, email, message string, stdout,
...
 	if err := Call(stdout, fetchStderr, gitCmd, "fetch", forkRemoteName, remoteBranch); err != nil {
 		logrus.Info("fetchStderr is : ", fetchStderr.String())
 		if !strings.Contains(strings.ToLower(fetchStderr.String()), fmt.Sprintf("couldn't find remote ref %s", remoteBranch)) {
 			return fmt.Errorf("failed to fetch from fork: %w", err)
+		} else {
+			needToDiff = false
 		}
 	} else {
-		var err error
-		remoteTreeRef, err = getTreeRef(stderr, fmt.Sprintf("refs/remotes/%s/%s", forkRemoteName, remoteBranch))
```
alvaroaleman, line 814: The precise reason of the whole fetching remote tree ref and comparing it with the local one is to not push no-op changes. What exactly is the problem with that you are seeing?

chaodaiG:
I believe this will work and serve its purpose, but I doubt whether it worths the complexity. Had synced with @mpherman2 offline and we couldn't hash out a better solution.

matthyx:
can we at least ensure there is no regression in the future by adding some tests?

mpherman2:
Just to summarize the conversation had here and offline: Given that the only real problem with the current implementation is that the reviewer receives emails more often, the problem is not problematic enough to warrant such added complexity until the autobumper has been refactored to use a more unit-testable git client. This will be revisited after that refactor has occurred

PR #22035, opened 2021-04-30
Ensure the DiableCloudProviders featureGate is not set by default

Danil-Grigorev:
That fix is required for https://github.com/kubernetes/kubernetes/pull/100136 to pass.

### DIFF
--- a/config/jobs/kubernetes/sig-cloud-provider/gcp/gcp-gce.yaml
+++ b/config/jobs/kubernetes/sig-cloud-provider/gcp/gcp-gce.yaml
@@ -413,7 +413,7 @@ presubmits:
         - --
         - --build=quick
         - --cluster=
-        - --env=KUBE_FEATURE_GATES=AllAlpha=true,CSIMigration=false,InTreePluginGCEUnregister=false
+        - --env=KUBE_FEATURE_GATES=AllAlpha=true,CSIMigration=false,InTreePluginGCEUnregister=false,DisableCloudProviders=false
         - --env=KUBE_PROXY_DAEMONSET=true
         - --env=ENABLE_POD_PRIORITY=true
         - --env=ENABLE_APISERVER_TRACING=true


config/jobs/kubernetes/sig-cloud-provider/gcp/gcp-gce.yaml
```diff
@@ -413,7 +413,7 @@ presubmits:
         - --
         - --build=quick
         - --cluster=
-        - --env=KUBE_FEATURE_GATES=AllAlpha=true,CSIMigration=false,InTreePluginGCEUnregister=false
+        - --env=KUBE_FEATURE_GATES=AllAlpha=true,CSIMigration=false,InTreePluginGCEUnregister=false,DisableCloudProviders=false
```
cheftako, line 416: /cc @jpbetz @cici37 
Not quite the build cloud provider flag which should never be disabled for this repo that I warned about. However this is a similar control but this one is a feature gate flag. This is another one we need to be aware of, although with finer grained controls, we may find this one useful for testing KAS+KCM+Kubelet.

elmiko (commented):
thanks Danil!

i am not adding the lgtm tag since i am not as familiar with this code, but this is what we need for our testing.

cheftako:
Fundamentally I do not expect all DisableCloudProviders=true to pass in K/K for GCE. The goal is to get it passing in K/cloud-provider-gcp.

config/jobs/kubernetes/sig-cloud-provider/gcp/gcp-gce.yaml
liggitt: the alpha job is broken since this merged with this error:

```
E0514 03:54:23.631068    2633 server.go:218] "Failed to set feature gates from initial flags-based config" err="unrecognized feature gate: DisableCloudProviders"
```

justaugustus: @liggitt -- I think we've got a chicken-and-egg.

Looks like https://github.com/kubernetes/kubernetes/pull/100136 needs to merge as well.

PR #22036, opened 2021-04-30
hack/analyze-memory-profiles: misc improvements

stevekuznetsov:
This patch adds:

 - use the creation timestamp of the profile files instead of the time
   in the output as the output time only has resolution to the minute
 - add the maximum of a series to its legend entry
 - plot the overall aggregate usage as well

usage

### DIFF
--- a/hack/analyze-memory-profiles.py
+++ b/hack/analyze-memory-profiles.py
@@ -22,14 +22,15 @@
 # visualize them.
 
 import os
+import pathlib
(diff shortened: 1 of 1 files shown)


stevekuznetsov:
Timing in example is messy since I copied files around so they all have basically the same `mtime` but it works as expected in the normal case ;)

petr-muller (approved):

PR #22038, opened 2021-04-30
Clonerefs pipes stdin in exec

chaodaiG:
Majority of times this will not be used, since clonerefs has been working. Lately I have observed that  hangs in prow job for several hours. Local test suggests:

- In a repo where http cookiefile is required,  prompts for username when cookiefile expired
- Doing the same thing in GKE pod with plain bash, and it fail right away, suggests tty detected

So the suspicion is that shelling out to  in Go stuck around this, https://groups.google.com/g/golang-nuts/c/pJe7TVb-68w suggests stdin might be related

### DIFF
--- a/prow/pod-utils/clone/clone.go
+++ b/prow/pod-utils/clone/clone.go
@@ -20,6 +20,7 @@ import (
 	"bytes"
 	"fmt"
 	"net/url"
+	"os"
 	"os/exec"
 	"path"
 	"strconv"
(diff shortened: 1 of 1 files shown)


fejta:
Neither this solution nor the explanation make a lot of sense to me.

What are you expecting to send what input into what clonerefs clone command?

chaodaiG:
Had an offline discussion with @fejta , agreed that this might not be the best way of solving this problem, shelving it for now

PR #22044, opened 2021-05-02
Allow kubekins image to build images in trusted cluster

dims:
Trying to get a CAPG node image working. The image-builder image does
not have all the things needed, so let's allow kubekins as well.
kubekins works fine in existing CAPG jobs where this node image is being
built.

### DIFF
--- a/config/jobs/image-pushing/k8s-staging-cluster-api.yaml
+++ b/config/jobs/image-pushing/k8s-staging-cluster-api.yaml
@@ -287,7 +287,7 @@ periodics:
   spec:
     serviceAccountName: gcb-builder
     containers:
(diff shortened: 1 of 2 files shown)


hasheddan (approved):
/lgtm

spiffxp:
/cc @BenTheElder @justinsb @cheftako  @vincepri 
I don't think I'm OK with this @dims.  The kubekins image is a kitchen sink.  I would rather have small, well-known images running in a cluster that someone might want to exfiltrate secrets from.  Also, why are we running the image build directly on the cluster?  This should be happening in GCB?

dims:
@spiffxp i had also logged some context here: https://github.com/kubernetes/k8s.io/issues/1993

PR #22056, opened 2021-05-03
nginx-ingress: update pull-ingress-nginx-test to use make test instead of cover

cpanato:
the cover rule was removed here: https://github.com/kubernetes/ingress-nginx/commit/a1c662a9a85a4b314b3c0d1835e2e11e6147b532#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52

so updating the job to run the unit tests without the cover for now.

/assign @rikatz @ElvinEfendi @strongjz

### DIFF
--- a/config/jobs/kubernetes/sig-network/ingress-nginx.yaml
+++ b/config/jobs/kubernetes/sig-network/ingress-nginx.yaml
@@ -143,22 +143,8 @@ presubmits:
       containers:
       - image: k8s.gcr.io/ingress-nginx/e2e-test-runner:v20210104-g81a8d5cd8@sha256:bfd55f589ea998f961825a9d09158d766cf621d1b8fc5d8c905aba07d9794e08
         command:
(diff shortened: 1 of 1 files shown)


rikatz:
/lgtm
/approve

Thanks!!!

config/jobs/kubernetes/sig-network/ingress-nginx.yaml
```diff
@@ -145,20 +145,7 @@ presubmits:
         command:
         - /bin/bash
         - -c
-        - "GIT_COMMIT=${PULL_PULL_SHA} make cover"
-        env:
```
aojea, line 149: should it be
```
- make
- test
```

cpanato: updated @aojea thanks for your review

PR #22060, opened 2021-05-03
Print debug message for http cookiefile used during git clone

chaodaiG:
We use http cookiefile for authenticating during cloning with http protocol, the cookiefile is rotated by either https://github.com/kubernetes/test-infra/blob/master/prow/cmd/grandmatriarch or https://github.com/istio/test-infra/tree/master/authentikos. When the cloning failed for any reason, half of the time the suspect tured to be staled cookiefile being used, but there is no immediate easy way for us to figure this out. Adding this debug line would help greatly in this scenario as grandmatriarch already prints md5sum.

### DIFF
--- a/prow/clonerefs/run.go
+++ b/prow/clonerefs/run.go
@@ -17,6 +17,7 @@ limitations under the License.
 package clonerefs
 
 import (
+	"crypto/md5"
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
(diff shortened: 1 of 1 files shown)


alvaroaleman (approved):

prow/clonerefs/run.go
```diff
@@ -83,6 +84,16 @@ func (o *Options) createRecords() []clone.Record {
 		oauthToken = token
 	}
 
+	// Print md5 sum of cookiefile for debugging purpose
+	if len(o.CookiePath) > 0 {
+		l := logrus.WithField("http-cookiefile", o.CookiePath)
+		f, err := ioutil.ReadFile(o.CookiePath)
+		if err != nil {
+			l.WithError(err).Warn("Cannot read http cookiefile")
```
fejta, line 92: What is the path that we cannot read?

prow/clonerefs/run.go
```diff
@@ -83,6 +84,16 @@ func (o *Options) createRecords() []clone.Record {
...
+		} else {
+			l.WithField("md5sum", fmt.Sprintf("%x", md5.Sum(f))).Debug("Http cookiefile md5 sum")
```
fejta, line 94: Info() seems more appropriate here.

fejta (commented):
/hold

prow/clonerefs/run.go
chaodaiG: line 89 includes the path in logger

prow/clonerefs/run.go
chaodaiG: Already merged. I can create another PR updating this if still desired

fejta: No need,LGTM!

PR #22066, opened 2021-05-03
kubernetes-csi: single pod jobs

pohly:
The prow jobs for https://github.com/kubernetes-csi/csi-driver-host-path/pull/282

Includes PR https://github.com/kubernetes/test-infra/pull/22065

/assign @msau42

### DIFF
--- a/config/jobs/kubernetes-csi/gen-jobs.sh
+++ b/config/jobs/kubernetes-csi/gen-jobs.sh
@@ -48,7 +48,7 @@ latest_stable_k8s_version="1.20" # TODO: bump to 1.21 after testing a pull job
 hostpath_driver_version="v1.6.0"
 
 # We need this image because it has Docker in Docker and go.
-dind_image="gcr.io/k8s-testimages/kubekins-e2e:v20210418-e5f251e-master"
+dind_image="gcr.io/k8s-testimages/kubekins-e2e:v20210428-a1a20d1-master"
 
 # All kubernetes-csi repos which are part of the hostpath driver example.
 # For these repos we generate the full test matrix. For each entry here
@@ -256,8 +256,12 @@ expand_tests () {
 pull_optional() {
     local tests="$1"
     local kubernetes="$2"
+    local deployment_suffix="$3"
 
-    if [ "$tests" == "alpha" ]; then
+    # https://github.com/kubernetes-csi/csi-driver-host-path/pull/282 has not been merged yet,
+    # therefore pull jobs which depend on the new deployment flavors have to be optional.
+    # TODO: remove this check once merged.
+    if [ "$tests" == "alpha" ] || [ "$deployment_suffix" ] ; then
         echo "true"
     elif [ "$kubernetes" == "$experimental_k8s_version" ]; then
         # New k8s versions may require updates to kind or release-tools.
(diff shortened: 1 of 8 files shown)


pohly:
Sorry, this needs further work.

pohly:
The new jobs are currently optional because I wanted to play it safe.

Enabling them also for old branches and making them mandatory there is not going to be a problem: the deployment suffix will be ignored and the two Prow jobs (with and without suffix) will run the same tests in the same configuration. This is a bit wasteful and could be avoided by skipping branches that are known to not have the additional deployments. The problem is that this list depends on the repo - I am not sure whether we care enough to configure that?

config/jobs/kubernetes-csi/gen-jobs.sh
```diff
@@ -345,6 +349,14 @@ use_bazel() (
     fi
 )
 
+additional_deployment_suffices () (
+    local repo="$1"
+
+    case "$repo" in
+        csi-driver-host-path) echo "-prow";;
```
msau42, line 356: would using a suffix like "-singlepod" be more self-descriptive?

pohly: I preferred to name the additional deployment after their purpose ("testing with Prow - don't use them") instead of what they currently contain ("one pod per sidecar"). My thinking was that the content might change so that it is more than just "single pod", in which case that name would become unsuitable.

But I don't feel strongly about this.

msau42: My thought was that the purpose of single pod deployment could extend beyond just testing with prow.

msau42: Oh sorry I thought this suffix was for the new model where all sidecars are in a pod. I guess my preference is still to name it after the deployment model. We have 3 in the repo now: distributed, combinedpod, separatepods?

pohly: Yes. But the deployment that "normal" users should use doesn't have a suffix ("kubernetes-1.18").

How about "kubernetes-1.18", "kubernetes-1.18-test", "kubernetes-distributed" (no version because there is only one)?

pohly: I.e. "-test" instead of "-prow"?

msau42: Sure that's fine. Regarding distributed, should we have the k8s version in it? It's only one now but we're adding 1.21 and 1.22 in the future

pohly: "distributed" is only used in one job at the moment. I think having just one version is fine, in particular because it works across all supported Kubernetes versions.

pohly: I've pushed a version with `-test` as suffix. Okay now?

msau42: I have a feeling that we will need multi k8s versions for distributed deployment for the same reasons why we have multi versions for normal deployments.

For now this is fine.

pohly: For normal deployments the reason currently is the need to test different snapshotter releases. That's not needed for distributed deployment because it doesn't support snapshotting. But I agree, eventually we might need it. We'll cross that bridge when we get there...

PR #22071, opened 2021-05-04
releng: Update Slack branch warnings

justaugustus:
- Move branch alerts to #release-management
- Remove RelMgr exemptions in kubernetes/kubernetes Slack warnings

  We don't actually require exemption anymore, as Release Managers no longer fast-forward branches

- Remove Slack exemptions from k/release `prototype` branch 

Continuation of https://github.com/kubernetes/release/pull/2042 and discussion in https://kubernetes.slack.com/archives/C2C40FMNF/p1620060717262100.

I would've liked to exempt dependabot from triggering Slack warnings, but it's an app, not a GitHub user, and I'm unsure how that will behave w/ the merge warning Prow plugin.
EDIT: Instead, we'll compromise and move the alerts to the #release-management channel.


hasheddan (approved):
/lgtm

config/prow/plugins.yaml
```diff
@@ -261,43 +261,15 @@ slack:
...
     - k8s-ci-robot # future home of tide
     - k8s-release-robot # anago
+    - cpanato # Release Manager
+    - feiskyer # Release Manager
+    - hasheddan # subproject owner / Release Manager
+    - idealhack # Release Manager
+    - justaugustus # subproject owner / Release Manager
+    - puerco # Release Manager
+    - saschagrunert # subproject owner / Release Manager
+    - xmudrii # Release Manager
```
cblecker, line 271: This would exempt these users from all branches, including master/default. Why would we want this?

cblecker (changes requested):
/hold

config/prow/plugins.yaml
justaugustus: @cblecker -- Thanks for the review!

That actually got me thinking...
We don't actually require exemption anymore, as Release Managers no longer fast-forward branches. 🙃 

I've dropped all of the Release Manager-specific branch exemptions.
/hold cancel

cblecker:
/lgtm
/approve

🎉

PR #22075, opened 2021-05-04
Security contacts validation

sfowl:
To support proposed change in https://github.com/kubernetes/security/issues/56

Marked this PR as a draft, as first the spec change to OWNERS/SECURITY_CONTACTS should be agreed upon and merged:

https://github.com/kubernetes/community/pull/5398

### DIFF
--- a/prow/plugins/verify-owners/verify-owners.go
+++ b/prow/plugins/verify-owners/verify-owners.go
@@ -367,6 +367,7 @@ func parseOwnersFile(oc ownersClient, path string, c github.PullRequestChange, l
 	var reviewers []string
 	var approvers []string
 	var labels []string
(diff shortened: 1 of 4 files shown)


sfowl:
Slack IDs don't seem to have a rigid format:

> We recommend considering the format of the string to be an opaque value, and not to rely on a particular structure.

https://api.slack.com/types/user

However in practice, I haven't seen any that don't follow the regex pattern of `[UW][A-Z0-9]+`. See https://github.com/kubernetes/community/blob/master/communication/slack-config/users.yaml for some example values.

I chose not to add validation for E-mail addresses as I wanted to leave the door open for obfuscate email addresses e.g. using " at " rather than "@".

PR #22081, opened 2021-05-04
Pluginconfig: Start sharding support

alvaroaleman:
This PR starts the sharding support for plugin config. It:
* Adds merging for the list of enabled plugins
* Plugs loading supplemental files throught the PluginAgent

It does not yet expose this on the binaries or add validation that its structured in org/repo format, I will create dedicated follow-up PRs for those.


prow/plugins/config.go
```diff
@@ -1766,3 +1767,40 @@ type Override struct {
...
+	}
+	if err := c.Plugins.mergeFrom(&other.Plugins); err != nil {
+		errs = append(errs, fmt.Errorf("failed to merge .plugins from supplemental config: %w", err))
+	}
+
+	return utilerrors.NewAggregate(errs)
+}
+
+func (p *Plugins) mergeFrom(other *Plugins) error {
+	if other == nil {
```
stevekuznetsov, line 1788: Doesn't the `for` implicitly handle this?

alvaroaleman: `other` can be nil and we dererence it in the loop, which will panic if its nil

prow/plugins/config.go
```diff
@@ -1766,3 +1767,40 @@ type Override struct {
...
+		return nil
+	}
+	if len(*p) == 0 {
```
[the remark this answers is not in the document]
alvaroaleman: Yeah, this is just a fastpath

prow/plugins/config_test.go
```diff
@@ -1520,3 +1523,138 @@ orgA/repoB:
...
+			},
+		},
+	}
+
+	seed := time.Now().UnixNano()
+	// Print the seed so failures can easily be reproduced
+	t.Logf("Seed: %d", seed)
+	fuzzer := fuzz.NewWithSeed(seed)
+
+	// Do not parallelize, the PRNG used by the fuzzer is not threadsafe
```
chases2, line 1588: Is this still true? It looks like _someone_ made this fuzzer thread-safe recently
https://github.com/google/gofuzz/pull/61

alvaroaleman: Updated gofuzz and parallelized the test

prow/plugins/config.go
```diff
@@ -1766,3 +1767,40 @@ type Override struct {
...
+		*p = *other
```
fejta, line 1792: This seems mildly surprising

fejta (approved):
/hold
/cc

Is there a link to an issue/proposal etc about the sharding effort?

fejta:
/uncc @e-blackwelder 
/cc @chaodaiG @cjwagner 
This something either of you two want to look at chao/cole?

prow/plugins/config_test.go
```diff
@@ -1520,3 +1523,141 @@ orgA/repoB:
...
+					t.Errorf("merging a config into itself did not produce an error. Serialized config:\n%s", string(serialized))
+				}
+			},
+		},
+	}
+
+	seed := time.Now().UnixNano()
+	// Print the seed so failures can easily be reproduced
+	t.Logf("Seed: %d", seed)
+	fuzzer := fuzz.NewWithSeed(seed)
```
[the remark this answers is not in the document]
alvaroaleman: The testing is very reliable actually, we only had one case where it was flaky in the past and that was due to the fuzzer only creating empty, non-nil structs with a very low probability: https://github.com/kubernetes/test-infra/issues/21507

It is IMHO important to have it in a presubmit, because there a bunch of things you simply can not test without, like the "merging doesn't lose data". There is no other test you could write that would start failing if someone adds another field to a struct that is being merged.

FWIW we use fuzz testing in t-i since November already, it was first added in 7fe0b85d44a

We can ofc reconsider this if we see that these tests get unreliable/flaky, but that wasn't the case so far.

prow/plugins/config.go
alvaroaleman: Why?

alvaroaleman:
> Is there a link to an issue/proposal etc about the sharding effort?

Yes, here: https://docs.google.com/document/d/1S8DFXrvFw27xYyHMDZmtUkVvd0cbDzFI8T5H35FkCyU/edit

prow/plugins/plugins.go
```diff
@@ -251,7 +254,7 @@ func NewFakeConfigAgent() ConfigAgent {
 // the file can't be read or the configuration is invalid.
 // If checkUnknownPlugins is true, unrecognized plugin names will make config
 // loading fail.
-func (pa *ConfigAgent) Load(path string, checkUnknownPlugins bool) error {
+func (pa *ConfigAgent) Load(path string, supplementalPluginConfigDirs []string, supplementalPluginConfigFileSuffix string, checkUnknownPlugins bool) error {
```
chaodaiG, line 257: This looks to me that the config will be something like:
```
--supplementalPluginConfigDirs=dir_a
--supplementalPluginConfigDirs=dir_b
--supplementalPluginConfigFileSuffix=custom.yaml 
```

This should work majority of cases, is there any chance in the future that the suffix might differ between directories? If so might want to consider changing this to a map instead

prow/plugins/config_test.go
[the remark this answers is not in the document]
fejta: If it is passing it isn't doing anything useful.

If it fails someone is going to do `/retest` and then because of the randomness it will pass.

So I'm not really seeing the scenario where this is a useful presubmit.

> There is no other test you could write that would start failing if someone adds another field to a struct that is being merged.

Reflection?

chaodaiG (approved):
LGTM to me, left a nit, feel free ignore

prow/plugins/config_test.go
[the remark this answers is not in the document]
alvaroaleman: > If it is passing it isn't doing anything useful.

It won't pass, it will instantly start failing. For a simple example, do a change like this:
```
diff --git a/prow/apis/prowjobs/v1/types.go b/prow/apis/prowjobs/v1/types.go
index 5fabefdbb7..b62ff84c35 100644
--- a/prow/apis/prowjobs/v1/types.go
+++ b/prow/apis/prowjobs/v1/types.go
@@ -325,6 +325,7 @@ func (d *Duration) MarshalJSON() ([]byte, error) {
 // This is primarily used to provide automatic integration with gubernator
 // and testgrid.
 type DecorationConfig struct {
+       TestField string
        // Timeout is how long the pod utilities will wait
        // before aborting a job with SIGINT.
        Timeout *Duration `json:"timeout,omitempty"
```

Then run
```
go test ./prow/apis/prowjobs/v1 -run TestApplyDefaultsAppliesDefaultsForAllFields
```

I'll buy you a beer if you manage to get that test to pass just a single time.

> Reflection?

That is this lib, we'd just be rebuilding it.

prow/plugins/plugins.go
alvaroaleman: I think this is easy to change, should we actually end up having that use-case

prow/plugins/config_test.go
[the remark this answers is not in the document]
fejta: oh no, I've been nerd sniped!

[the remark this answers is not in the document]
fejta: I am defeated:
```
//prow/apis/prowjobs/v1:go_default_test                                  FAILED in 5000 out of 5000 in 1.4s
  Stats over 5000 runs: max = 1.4s, min = 0.3s, avg = 0.6s, dev = 0.2s
```

stevekuznetsov:
/lgtm

From me as well, so I think we can 

/hold cancel

PR #22092, opened 2021-05-06
Add milestone plugin to HNC

adrianludwin:


adrianludwin:
Not sure if this is the right way to do it but Ryan, wdyt?

/assign @rjbez17

rjbez17:
/ok-to-test
/lgtm
/approve

This seems correct to me, we can iterate if it's not.

PR #22098, opened 2021-05-06
dco: Retrieve default branch for CONTRIBUTING link

dhaiducek:
With our repos migrating to a default branch called `main` and not `master`, the `dco` link to the `CONTRIBUTING.md` is broken. This change detects the default branch to use in the URL.

(I'm not super comfortable with Go yet, so let me know if something could be done better. I also haven't worked out yet how to write a test that would mock `GetRepo()` calls so that the code could be tested, but the current tests pass so far.)


prow/plugins/dco/dco.go
```diff
@@ -202,7 +203,18 @@ func checkExistingLabels(gc gitHubClient, l *logrus.Entry, org, repo string, num
 // takeAction will take appropriate action on the pull request according to its
 // current state.
 func takeAction(gc gitHubClient, cp commentPruner, l *logrus.Entry, org, repo string, pr github.PullRequest, commitsMissingDCO []github.RepositoryCommit, existingStatus string, hasYesLabel, hasNoLabel, addComment bool) error {
-	targetURL := fmt.Sprintf("https://github.com/%s/%s/blob/master/CONTRIBUTING.md", org, repo)
+	// Get default branch for repo to construct link to CONTRIBUTING.md (default: master)
+	repoObj, err := gc.GetRepo(org, repo)
```
alvaroaleman, line 207: This will have a terrible cache hit rate, as the response contains details like `pushed_at`, number of stars and similiar. Can we build a cached version of this in `prow/github`?

alvaroaleman: "this" being "GetDefaultBranch(org, repo)"

dhaiducek: I was wondering whether grabbing the entire repo object every time would be detrimental. Yes, I'll work on improving that today.

dhaiducek: Sorry for the delay. Implementing a cache is a bit new for me, but I've implemented a basic in-memory cache for `GetDefaultBranch` in `client.go`. Please take a look when you have a moment! Thanks!

prow/github/client.go
```diff
@@ -348,6 +350,11 @@ func (c *client) WithFields(fields logrus.Fields) Client {
 	}
 }
 
+// In-memory caching (https://github.com/patrickmn/go-cache)
+var (
+	memCache = cache.New(20*time.Minute, 30*time.Minute)
```
alvaroaleman, line 355: Package-level vars are a big no-no, please make this a property of `client`

prow/github/client.go
```diff
@@ -2475,6 +2482,34 @@ func (c *client) GetBranches(org, repo string, onlyProtected bool) ([]Branch, er
...
+	cacheValue, found := memCache.Get(repoKey)
+	if found {
+		defaultBranch = cacheValue.(string)
+	} else {
+		repoObj, err := c.GetRepo(org, repo)
+		if err != nil {
+			return "", err
+		}
+		defaultBranch = repoObj.Repo.DefaultBranch
+		memCache.Set(repoKey, defaultBranch, 0)
```
alvaroaleman, line 2507: So if you don't expire them this isn't worth a new lib, you can implement a cache as a `map[string]string` with a `sync.Mutex` for thread safety

prow/github/client.go
```diff
@@ -2475,6 +2482,34 @@ func (c *client) GetBranches(org, repo string, onlyProtected bool) ([]Branch, er
 	return branches, nil
 }
 
+// GetDefaultBranch returns default branch of the repo.
+// Uses a memory cache for performance, storing repos with key `org_repo_defaultBranch`.
+//
+// See https://docs.github.com/en/rest/reference/repos#list-organization-repositories
+func (c *client) GetDefaultBranch(org, repo string) (string, error) {
```
alvaroaleman, line 2489: Nit, could we make sure to clarify that this is cached?
```suggestion
func (c *client) GetDefaultBranchCached(org, repo string) (string, error) {
```

prow/github/client.go
```diff
@@ -2475,6 +2482,34 @@ func (c *client) GetBranches(org, repo string, onlyProtected bool) ([]Branch, er
...
+		repoObj, err := c.GetRepo(org, repo)
+		if err != nil {
+			return "", err
+		}
+		defaultBranch = repoObj.Repo.DefaultBranch
+		memCache.Set(repoKey, defaultBranch, 0)
+	}
+
+	return defaultBranch, nil
+}
```
alvaroaleman, line 2511: Please add some tests

dhaiducek:
Thanks for the reminder stalebot--I do plan to return to complete this. 🙂

Issue #22099, opened 2021-05-06
Revisit root OWNERS in this repo, and whether test-infra-oncall still needs to be in it


spiffxp:
**What should be cleaned up or changed**:

There are a number of people in test-infra-oncall who don't regularly interact with this repo, and thus have insufficient context to review/approve random PRs that come their way.

A stopgap measure is to take test-infra-oncall out of root reviewers, and explicitly add in some of its members who are knowledgeable enough as root reviewers.

Longer term, some things to consider:
- could we push test-infra-oncall approve privileges to specific directories instead of root?
- how many files / PRs are hitting root OWNERS for reviewers?
- how many files / PRs are suggesting root OWNERS for approval?

Based on that, we should consider:
- adding more subdirectory or leaf OWNERS, and growing reviewers there
- consider updating prow to not suggest root OWNERS if leaf OWNERS provide sufficient coverage

FYI @fejta @BenTheElder @stevekuznetsov 

/sig testing
/priority important-longterm

spiffxp:
https://github.com/kubernetes/test-infra/pull/22100 is the stopgap measure

PR #22100, opened 2021-05-06
Reduce oncall top-level reviewers.

fejta:
Oncall has people with different amounts of involvement in test-infra.
It is useful to have oncall be able to approve these PRs when necessary,
but we should try to direct reviews to the people actively involved in
this repo. So reduce to that subset.

For feedback:
/assign @BenTheElder @spiffxp @stevekuznetsov 

Do you three want to get these kinds of reviews?
/cc @chaodaiG @cjwagner @mpherman2

ref https://github.com/kubernetes/test-infra/issues/22099
/hold

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -4,7 +4,10 @@ reviewers:
 - BenTheElder # lead
 - spiffxp # lead
 - stevekuznetsov # lead
-- test-infra-oncall # oncall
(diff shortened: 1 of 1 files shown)


e-blackwelder:
Is there a `/woot` command?? :)

cjwagner:
+1 from me

PR #22103, opened 2021-05-07
Additional kOps upgrade tests

justinsb:
To test the scenario in kops #11370

### DIFF
--- a/config/jobs/kubernetes/kops/kops-periodics-misc.yaml
+++ b/config/jobs/kubernetes/kops/kops-periodics-misc.yaml
@@ -114,3 +114,107 @@ periodics:
     testgrid-dashboards: sig-cluster-lifecycle-kops, google-aws, kops-misc, kops-kubetest2
     testgrid-days-of-results: "30"
     testgrid-tab-name: kops-aws-kops-upgrade
+- interval: 1h # TODO(justinsb): reduce once working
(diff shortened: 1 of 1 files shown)


justinsb:
Depends on https://github.com/kubernetes/kops/pull/11403

justinsb:
/assign @rifelpet 

I didn't realize you were actively working on these, so this may overlap with your work!

config/jobs/kubernetes/kops/kops-periodics-misc.yaml
```diff
@@ -114,3 +114,107 @@ periodics:
...
+      - name: KOPS_STATE_STORE
+        value: s3://k8s-kops-prow
+      - name: KOPS_VERSION_A
+        value: 1.18.3
+      - name: KOPS_VERSION_B
+        value: 1.19.2
+      - name: K8S_VERSION_A
+        value: 1.18.18
+      - name: K8S_VERSION_B
+        value: 1.18.18
```
rifelpet, line 159: I was wondering if we could somehow utilize channels so that we don't need to specify patch versions for all of these

justinsb: Using channels is a good idea.  I wasn't sure what we should be testing though e.g. 1.18.0 to 1.19.<latest> ?  So I punted on "how do we choose these values" problem ... for now!

PR #22107, opened 2021-05-07
Network Performance tc's: changes for fetching worker pod logs

VivekThrivikraman-est:
Currently the worker pod logs are not extracted by logexporter, this change is for fetching the worker pods log as well along with other logs.

### DIFF
--- a/config/jobs/kubernetes/sig-scalability/sig-scalability-periodic-jobs.yaml
+++ b/config/jobs/kubernetes/sig-scalability/sig-scalability-periodic-jobs.yaml
@@ -949,6 +949,7 @@ periodics:
       - --gcp-nodes=102
       - --provider=gce
       - --test=false
+      - --env=LOG_DUMP_EXTRA_FILES=containers/worker*
       - --test-cmd=$GOPATH/src/k8s.io/perf-tests/run-e2e.sh
       - --test-cmd-args=cluster-loader2
       - --test-cmd-args=--nodes=100


config/jobs/kubernetes/sig-scalability/sig-scalability-periodic-jobs.yaml
```diff
@@ -949,6 +949,7 @@ periodics:
       - --gcp-nodes=102
       - --provider=gce
       - --test=false
+      - --env=LOG_DUMP_EXTRA_FILES=containers/worker*
```
marseel, line 952: Do we have some example run that confirms it works?

mborsz: Do we have any experimental run that confirms that it works?

Also I guess you may want to set this to --env=LOG_DUMP_EXTRA_FILES=cl2-*containers/worker* to keep cl2-* logs (as in preset).

mborsz: I have started manual run and it failed to gather worker* logs.

In logexporter's logs I see:

W0512 09:44:49.182043       1 main.go:189] Failed to copy any logfiles with pattern '/var/log/containers/worker*.log*': exit status 1

Most likely the link in /var/log/containers/ no longer exists when we run logexporter (as clusterloader has already finished and deleted that pod).

Could you verify that by. creating few node cluster and seeing what happens to /var/log/containers:
a) when pod exists
b) when pod is deleted?

Is there. any path that can be used to. reach container's logs after container is deleted?

VivekThrivikraman-est: As you mentioned once a test is run and pod deleted the files are also removed, so logexporter can't get those logs. Planning to try out the option of writing log file directly to /var/log using log_file flag.

PR #22108, opened 2021-05-07
Add a package for replicating improvements made in k8s prow monitoring stack

chaodaiG:
This package will be used for replicating all improvements made in https://github.com/kubernetes/test-infra/tree/master/config/prow/cluster/monitoring/mixins to all other prow instances.

This package can be invoked manually for now. Once refactoring of generic-autobumper is completed will add PR creation logic to this package so that it could be used by prow jobs just like how prow images are bumped


mpherman2:
/lgtm
/hold
Holding if you want to wait for Cole to look at it or if you want to wait for the PR creator

chaodaiG:
/unhold
This PR is harmless, we can discuss in follow up PR that integrates it with prcreator

PR #22112, opened 2021-05-07
Generic autobumper: use bumper package as prcreator instead

chaodaiG:
This is the second PR as breaking down of https://github.com/kubernetes/test-infra/pull/21952. This PR moves the prow images bumping logic into the main package, haven't done any cleaning up yet for maintaining backwards compatibility.

### DIFF
--- a/prow/cmd/generic-autobumper/main.go
+++ b/prow/cmd/generic-autobumper/main.go
@@ -17,6 +17,7 @@ limitations under the License.
 package main
 
 import (
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"net/http"
@@ -52,7 +53,7 @@ var (
 var _ bumper.PRHandler = (*client)(nil)
 
 type client struct {
-	o        *Options
+	o        *options
 	images   map[string]string
 	versions map[string][]string
 }
@@ -63,13 +64,11 @@ func (c *client) Changes() []func() (string, error) {
 	return []func() (string, error){
 		func() (string, error) {
 			var err error
-			c.images, err = UpdateReferences(c.o)
-			if err != nil {
+			if c.images, err = updateReferencesWrapper(c.o); err != nil {
 				return "", fmt.Errorf("failed to update image references: %w", err)
 			}
 
-			c.versions, err = getVersionsAndCheckConsistency(c.o.Prefixes, c.images)
-			if err != nil {
+			if c.versions, err = getVersionsAndCheckConsistency(c.o.Prefixes, c.images); err != nil {
 				return "", err
 			}
 
@@ -88,8 +87,8 @@ func (c *client) PRTitleBody() (string, string, error) {
 	return makeCommitSummary(c.o.Prefixes, c.versions), "", nil
 }
 
-// Options is the options for autobumper operations.
-type Options struct {
+// options is the options for autobumper operations.
+type options struct {
 	// The URL where upstream image references are located. Only required if Target Version is "upstream" or "upstreamStaging". Use "https://raw.githubusercontent.com/{ORG}/{REPO}"
 	// Images will be bumped based off images located at the address using this URL and the refConfigFile or stagingRefConigFile for each Prefix.
 	UpstreamURLBase string `yaml:"upstreamURLBase"`
@@ -102,11 +101,11 @@ type Options struct {
 	// The target version to bump images version to, which can be one of latest, upstream, upstream-staging and vYYYYMMDD-deadbeef.
 	TargetVersion string `yaml:"targetVersion"`
 	// List of prefixes that the autobumped is looking for, and other information needed to bump them. Must have at least 1 prefix.
-	Prefixes []Prefix `yaml:"prefixes"`
+	Prefixes []prefix `yaml:"prefixes"`
 }
 
-// Prefix is the information needed for each prefix being bumped.
-type Prefix struct {
+// prefix is the information needed for each prefix being bumped.
+type prefix struct {
 	// Name of the tool being bumped
 	Name string `yaml:"name"`
 	// The image prefix that the autobumper should look for
@@ -123,7 +122,7 @@ type Prefix struct {
 	ConsistentImages bool `yaml:"consistentImages"`
 }
 
-func parseOptions() (*Options, *bumper.Options, error) {
+func parseOptions() (*options, *bumper.Options, error) {
 	var config string
 	var labelsOverride []string
 	var skipPullRequest bool
@@ -133,20 +132,19 @@ func parseOptions() (*Options, *bumper.Options, error) {
 	flag.BoolVar(&skipPullRequest, "skip-pullrequest", false, "")
 	flag.Parse()
 
-	var o Options
+	var o options
 	var pro bumper.Options
 	data, err := ioutil.ReadFile(config)
 	if err != nil {
-		return nil, nil, fmt.Errorf("Failed to read in config file, %s", config)
+		return nil, nil, fmt.Errorf("read %q: %w", config, err)
 	}
 
-	err = yaml.Unmarshal(data, &o)
-	if err != nil {
-		return nil, nil, fmt.Errorf("Failed to parse yaml file, %s", err)
+	if err = yaml.Unmarshal(data, &o); err != nil {
+		return nil, nil, fmt.Errorf("unmarshal %q: %w", config, err)
 	}
 
 	if err := yaml.Unmarshal(data, &pro); err != nil {
-		return nil, nil, fmt.Errorf("Failed to parse yaml file, %s", err)
+		return nil, nil, fmt.Errorf("unmarshal %q: %w", config, err)
 	}
 
 	if labelsOverride != nil {
@@ -156,17 +154,17 @@ func parseOptions() (*Options, *bumper.Options, error) {
 	return &o, &pro, nil
 }
 
-func validateOptions(o *Options) error {
+func validateOptions(o *options) error {
 	if len(o.Prefixes) == 0 {
-		return fmt.Errorf("Must have at least one Prefix specified")
+		return errors.New("must have at least one Prefix specified")
 	}
 	if len(o.IncludedConfigPaths) == 0 {
-		return fmt.Errorf("includedConfigPaths is mandatory")
+		return errors.New("includedConfigPaths is mandatory")
 	}
 	if o.TargetVersion != latestVersion && o.TargetVersion != upstreamVersion &&
 		o.TargetVersion != upstreamStagingVersion && !tagRegexp.MatchString(o.TargetVersion) {
-		logrus.Warnf("Warning: targetVersion is not one of %v so it might not work properly.",
-			[]string{latestVersion, upstreamVersion, upstreamStagingVersion, tagVersion})
+		logrus.WithField("allowed", []string{latestVersion, upstreamVersion, upstreamStagingVersion, tagVersion}).Warn(
+			"Warning: targetVersion mot in allowed so it might not work properly.")
 	}
 	if o.TargetVersion == upstreamVersion {
 		for _, prefix := range o.Prefixes {
@@ -190,16 +188,19 @@ func validateOptions(o *Options) error {
 	return nil
 }
 
-// UpdateReferences update the references of prow-images and/or boskos-images and/or testimages
+// updateReferencesWrapper update the references of prow-images and/or boskos-images and/or testimages
 // in the files in any of "subfolders" of the includeConfigPaths but not in excludeConfigPaths
 // if the file is a yaml file (*.yaml) or extraFiles[file]=true
-func UpdateReferences(o *Options) (map[string]string, error) {
+func updateReferencesWrapper(o *options) (map[string]string, error) {
 	logrus.Info("Bumping image references...")
 	var allPrefixes []string
 	for _, prefix := range o.Prefixes {
 		allPrefixes = append(allPrefixes, prefix.Prefix)
 	}
-	filterRegexp := regexp.MustCompile(strings.Join(allPrefixes, "|"))
+	filterRegexp, err := regexp.Compile(strings.Join(allPrefixes, "|"))
+	if err != nil {
+		return nil, fmt.Errorf("bad regexp %q: %w", strings.Join(allPrefixes, "|"), err)
+	}
 	imageBumperCli := imagebumper.NewClient()
 	return updateReferences(imageBumperCli, filterRegexp, o)
 }
@@ -212,16 +213,15 @@ type imageBumper interface {
 	TagExists(imageHost, imageName, currentTag string) (bool, error)
 }
 
-func updateReferences(imageBumperCli imageBumper, filterRegexp *regexp.Regexp, o *Options) (map[string]string, error) {
+func updateReferences(imageBumperCli imageBumper, filterRegexp *regexp.Regexp, o *options) (map[string]string, error) {
 	var tagPicker func(string, string, string) (string, error)
-	var err error
 
 	switch o.TargetVersion {
 	case latestVersion:
 		tagPicker = imageBumperCli.FindLatestTag
 	case upstreamVersion, upstreamStagingVersion:
-		tagPicker, err = upstreamImageVersionResolver(o, o.TargetVersion, parseUpstreamImageVersion, imageBumperCli)
-		if err != nil {
+		var err error
+		if tagPicker, err = upstreamImageVersionResolver(o, o.TargetVersion, parseUpstreamImageVersion, imageBumperCli); err != nil {
 			return nil, fmt.Errorf("failed to resolve the %s image version: %w", o.TargetVersion, err)
 		}
 	default:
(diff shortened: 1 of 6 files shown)


prow/cmd/generic-autobumper/main.go
```diff
@@ -19,47 +19,467 @@ package main
...
+	var skipPullRequest bool
 
 	flag.StringVar(&config, "config", "", "The path to the config file for the autobumber.")
 	flag.StringSliceVar(&labelsOverride, "labels-override", nil, "Override labels to be added to PR.")
+	flag.BoolVar(&skipPullRequest, "skip-pullrequest", false, "")
 	flag.Parse()
 
-	var o bumper.Options
+	var o Options
+	var pro bumper.Options
```
chaodaiG, line 137: @alvaroaleman , saw your comment in https://github.com/kubernetes/test-infra/pull/21952#discussion_r627712028, this is less easier to do in this PR due to duplicated fields for backwards compatibility(If deleting these field from bumper.Options then it'll break the old set of functions). I think it's doable and will do it in the next PR for cleaning up

chaodaiG:
/cc @alvaroaleman @mpherman2 

/hold
Let's not merge it late Friday

prow/cmd/generic-autobumper/main_test.go
```diff
@@ -0,0 +1,843 @@
...
+		extraFiles         []string
+		expectedRes        map[string]string
+		expectError        bool
+	}{
+		{
+			description:        "update the images to the latest version",
+			targetVersion:      latestVersion,
+			includeConfigPaths: []string{"testdata/dir/subdir1", "testdata/dir/subdir2"},
+			expectedRes: map[string]string{
+				"testdata/dir/subdir1/test1-1.yaml": "fake-latest",
```
fejta, line 211: It looks like all these files are empty. Instead of checking in these files can we just create them at the start of the test?

https://github.com/kubernetes/test-infra/blob/176e4b678ce11d0c4640d818d57034e9ae523403/testgrid/cmd/transfigure/cmd/transfigure.go#L84-L88

prow/cmd/generic-autobumper/main.go
```diff
@@ -19,47 +19,467 @@ package main
...
+		logrus.Warnf("targetVersion can't be 'upstream' or 'upstreamStaging` without upstreamURLBase set. Default upstreamURLBase is %q", defaultUpstreamURLBase)
+	}
+
+	return nil
+}
+
+// UpdateReferences update the references of prow-images and/or boskos-images and/or testimages
+// in the files in any of "subfolders" of the includeConfigPaths but not in excludeConfigPaths
+// if the file is a yaml file (*.yaml) or extraFiles[file]=true
+func UpdateReferences(o *Options) (map[string]string, error) {
```
fejta, line 196: What package are you exporting UpdateReferences to?

If this is only used by the main package, please do not export this function unless another package needs it or you are trying to satisfy an interface (applies to all exported functions in the main package)

prow/cmd/generic-autobumper/main.go
```diff
@@ -19,47 +19,467 @@ package main
...
+	var o Options
+	var pro bumper.Options
 	data, err := ioutil.ReadFile(config)
 	if err != nil {
-		return nil, fmt.Errorf("Failed to read in config file, %s", config)
+		return nil, nil, fmt.Errorf("Failed to read in config file, %s", config)
+	}
+
+	err = yaml.Unmarshal(data, &o)
+	if err != nil {
```
fejta, line 144: merge with previous line

prow/cmd/generic-autobumper/main.go
```diff
@@ -19,47 +19,467 @@ package main
...
+		return nil, nil, fmt.Errorf("Failed to parse yaml file, %s", err)
```
fejta, line 145: errors should be lower case
prefer %w over %s

Generally I think of errors are more like stack traces, less human readable sentences (that is for logging).

So I would write this as `fmt.Errorf("unmarshal %q: %w", config, err)`

(applies everywhere)

prow/cmd/generic-autobumper/main.go
```diff
@@ -19,47 +19,467 @@ package main
...
+		pro.Labels = labelsOverride
 	}
-	return &o, nil
+	pro.SkipPullRequest = skipPullRequest
+	return &o, &pro, nil
+}
+
+func validateOptions(o *Options) error {
+	if len(o.Prefixes) == 0 {
+		return fmt.Errorf("Must have at least one Prefix specified")
```
fejta, line 161: Prefer `errors.New()` when you are not formatting a string

(applies everywhere)

prow/cmd/generic-autobumper/main.go
```diff
@@ -19,47 +19,467 @@ package main
...
+// UpdateReferences update the references of prow-images and/or boskos-images and/or testimages
+// in the files in any of "subfolders" of the includeConfigPaths but not in excludeConfigPaths
+// if the file is a yaml file (*.yaml) or extraFiles[file]=true
+func UpdateReferences(o *Options) (map[string]string, error) {
+	logrus.Info("Bumping image references...")
+	var allPrefixes []string
+	for _, prefix := range o.Prefixes {
+		allPrefixes = append(allPrefixes, prefix.Prefix)
+	}
+	filterRegexp := regexp.MustCompile(strings.Join(allPrefixes, "|"))
```
fejta, line 202: Please handle the error

prow/cmd/generic-autobumper/main.go
```diff
@@ -19,47 +19,467 @@ package main
...
+	FindLatestTag(imageHost, imageName, currentTag string) (string, error)
+	UpdateFile(tagPicker func(imageHost, imageName, currentTag string) (string, error), path string, imageFilter *regexp.Regexp) error
+	GetReplacements() map[string]string
+	AddToCache(image, newTag string)
+	TagExists(imageHost, imageName, currentTag string) (bool, error)
+}
+
+func updateReferences(imageBumperCli imageBumper, filterRegexp *regexp.Regexp, o *Options) (map[string]string, error) {
+	var tagPicker func(string, string, string) (string, error)
+	var err error
```
fejta, line 217: not necessary to declare an error here, I'd move to just before the `upstreamImageVersionResolver()` call

prow/cmd/generic-autobumper/main.go
```diff
@@ -19,47 +19,467 @@ package main
...
+
+	switch o.TargetVersion {
+	case latestVersion:
+		tagPicker = imageBumperCli.FindLatestTag
+	case upstreamVersion, upstreamStagingVersion:
+		tagPicker, err = upstreamImageVersionResolver(o, o.TargetVersion, parseUpstreamImageVersion, imageBumperCli)
+		if err != nil {
```
fejta, line 224: since you're using `=` rather than `:=` you can combine these two lines

prow/cmd/generic-autobumper/main.go
```diff
@@ -19,47 +19,467 @@ package main
...
+			return nil, fmt.Errorf("failed to resolve the %s image version: %w", o.TargetVersion, err)
+		}
+	default:
+		tagPicker = func(imageHost, imageName, currentTag string) (string, error) { return o.TargetVersion, nil }
+	}
+
+	updateFile := func(name string) error {
+		logrus.Infof("Updating file %s", name)
```
fejta, line 232: Prefer structured logging, so `logrus.WithField("file", name).Info("Updating")` or something along those lines.

prow/cmd/generic-autobumper/main.go
```diff
@@ -19,47 +19,467 @@ package main
...
+			return updateFile(name)
+		}
+		return nil
+	}
+
+	// Updated all .yaml files under the included config paths but not under excluded config paths.
+	for _, path := range o.IncludedConfigPaths {
+		info, err := os.Stat(path)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get the file info for %q", path)
```
fejta, line 249: Is there a reason you're not including err?

prow/cmd/generic-autobumper/main.go
```diff
@@ -19,47 +19,467 @@ package main
...
+	o *Options, upstreamVersionType string, parse func(upstreamAddress, prefix string) (string, error), imageBumperCli imageBumper) (func(imageHost, imageName, currentTag string) (string, error), error) {
+	upstreamVersions, err := upstreamConfigVersions(upstreamVersionType, o, parse)
+	if err != nil {
+		return nil, err
+	}
+
+	return func(imageHost, imageName, currentTag string) (string, error) {
+		imageFullPath := imageHost + "/" + imageName + ":" + currentTag
+		for prefix, version := range upstreamVersions {
+			if strings.HasPrefix(imageFullPath, prefix) {
```
fejta, line 286: Please try to minimize indenting, so I would rewrite this as:

```golang
if !strings.HasPrefix(imageFullPath, prefix) {
  continue
}
exists, err := imageBumperCli...
...
```

prow/cmd/generic-autobumper/main_test.go
```diff
@@ -0,0 +1,843 @@
+/*
+Copyright 2019 The Kubernetes Authors.
```
fejta, line 2: Should this be 2021?

prow/cmd/generic-autobumper/main.go
```diff
@@ -19,47 +19,467 @@ package main
...
+}
+
+// Changes returns a slice of functions, each one does some stuff, and
+// returns commit message for the changes
+func (c *client) Changes() []func() (string, error) {
+	return []func() (string, error){
+		func() (string, error) {
+			var err error
+			c.images, err = UpdateReferences(c.o)
+			if err != nil {
```
fejta, line 67: I'd combine with the previous line, since you're using using `=` (applies everywhere, think I already wrote this somewhere else once)

prow/cmd/generic-autobumper/main.go
```diff
@@ -19,47 +19,467 @@ package main
...
+	}
+}
+
+// PRTitleBody returns the body of the PR, this function runs after each commit
+func (c *client) PRTitleBody() (string, string, error) {
+	return makeCommitSummary(c.o.Prefixes, c.versions), "", nil
+}
+
+// Options is the options for autobumper operations.
+type Options struct {
```
fejta, line 92: suspect this doesn't need to be exported either

prow/cmd/generic-autobumper/main.go
```diff
@@ -19,47 +19,467 @@ package main
...
+func validateOptions(o *Options) error {
+	if len(o.Prefixes) == 0 {
+		return fmt.Errorf("Must have at least one Prefix specified")
+	}
+	if len(o.IncludedConfigPaths) == 0 {
+		return fmt.Errorf("includedConfigPaths is mandatory")
+	}
+	if o.TargetVersion != latestVersion && o.TargetVersion != upstreamVersion &&
+		o.TargetVersion != upstreamStagingVersion && !tagRegexp.MatchString(o.TargetVersion) {
+		logrus.Warnf("Warning: targetVersion is not one of %v so it might not work properly.",
```
fejta, line 168: I rarely use formatted strings with logrus. Using fields is the preferred style.

prow/cmd/generic-autobumper/main_test.go
```diff
@@ -0,0 +1,843 @@
...
+		Prefix:               prowPrefix,
+		RefConfigFile:        prowRefConfigFile,
+		StagingRefConfigFile: prowStagingRefConfigFile,
+	}
+	boskosPrefixStruct := Prefix{
+		Prefix:               boskosPrefix,
+		RefConfigFile:        boskosRefConfigFile,
+		StagingRefConfigFile: boskosStagingRefConfigFile,
+	}
+	//Prefix used to test when a tag does not exist. This is used to have parser return a tag that will make TagExists return false
```
fejta, line 390: I'd prefer a space between `//` and `Prefix`

prow/cmd/generic-autobumper/main_test.go
```diff
@@ -0,0 +1,843 @@
...
+			}
+			if !tc.expectError && err != nil {
+				t.Errorf("Expected to not get an error but got one: %v", err)
+			}
+		})
+	}
+}
+
+func TestUpstreamImageVersionResolver(t *testing.T) {
+	prowProdFakeVersion := "v-prow-prod-version"
```
fejta, line 366: a) these can all be consts, right?

```golang
const (
  prowProdFakeVersion = "v-prow-prod-version"
  prowStagingFakeVersion = "v-prow-staging-version"
  ...
)
```

b) While it can be more verbose, when it comes times to read and maintain unit tests IMO it tends to be easier if you put as much info entirely inside the table as as little as possible outside it. Although this can be an art. 

AKA it is much easie to read `expectedTargetTag: "v-prow-prod-version"` than it is to hunt down what `prowProdFakeVersion` is set to.

prow/cmd/generic-autobumper/main_test.go
```diff
@@ -0,0 +1,843 @@
...
+		StagingRefConfigFile: boskosStagingRefConfigFile,
+	}
+	//Prefix used to test when a tag does not exist. This is used to have parser return a tag that will make TagExists return false
+	tagDoesNotExistPrefix := Prefix{
+		Prefix:               doesNotExistPrefix,
+		RefConfigFile:        doesNotExist,
+		StagingRefConfigFile: doesNotExist,
+	}
+
+	fakeImageVersionParser := func(upstreamAddress, prefix string) (string, error) {
```
fejta, line 397: Consider making this part of the table, which allows me to provide an image parser laser focused for the scenario, which tends to be easier to read.

It will be harder for me 3 months from now to try and figure out if it is safe for me to modify `case fakeUpstreamURLBase + "/" + prowRefConfigFile:` right now as every unit test depends on it.

So instead you can do this


```golang
cases := []struct {
  parser func(string, string) (string, error)
  currentTag string
}{
  {
      description: "whatever",
      currentTag: "cute-joke-to-make-this-more-fun",
      parser: func(upAddr, pref string) (string, error) {
         if upAddr == "cute-joke-to-make-this-more-fun" {
            return pref, nil
         }
         return "", errors.New("try again, you screwed up!")
    },
}
```

This tends to be easier to maintain -- easier to read, less brittle. Easier to read since I only need to care about the parser function inside the scope of the table case. Less brittle because I can add a new case and add a new parser for that case, rather than trying to figure out the best way to reuse and/or update the global parser to handle my additional scenario.

fejta:
If you don't actually want this PR to be reviewed would it be useful to add a second commit with stuff you do want to be reviewed?

Since this PR is entirely new lines of code it is not obvious to me that anything is getting moved or copied.

chaodaiG:
> If you don't actually want this PR to be reviewed would it be useful to add a second commit with stuff you do want to be reviewed?
> 
> Since this PR is entirely new lines of code it is not obvious to me that anything is getting moved or copied.

Good point, I'll restructure this PR first to make it better reviewable. By the way, I think your review feedbacks are all valuable, will address them on top of the restructuring

prow/cmd/generic-autobumper/main.go
chaodaiG: Made it private

prow/cmd/generic-autobumper/main.go
chaodaiG: done

prow/cmd/generic-autobumper/main.go
chaodaiG: done

prow/cmd/generic-autobumper/main.go
chaodaiG: done

prow/cmd/generic-autobumper/main.go
chaodaiG: done

prow/cmd/generic-autobumper/main.go
chaodaiG: renamed and made it private

prow/cmd/generic-autobumper/main.go
chaodaiG: done

prow/cmd/generic-autobumper/main.go
chaodaiG: done

prow/cmd/generic-autobumper/main.go
chaodaiG: done

prow/cmd/generic-autobumper/main.go
chaodaiG: done

prow/cmd/generic-autobumper/main.go
chaodaiG: added

prow/cmd/generic-autobumper/main.go
chaodaiG: done

prow/cmd/generic-autobumper/main_test.go
chaodaiG: done

prow/cmd/generic-autobumper/main_test.go
chaodaiG: fixed

prow/cmd/generic-autobumper/main.go
chaodaiG: done

prow/cmd/generic-autobumper/main_test.go
chaodaiG: done

chaodaiG:
Done, the PR had been broken down into 3 commits:
1. Blind copy/paste functions from `generic-autobumper/bumper` package
2. `generic-autobumper` main package does all local files manipulations, and use prcreator functions from `generic-autobumper/bumper` for PRs
3. Improvements based on comments

PR #22122, opened 2021-05-09
kops-upgrade-ab: Expect versions to have "v" prefix

justinsb:
This better supports versions that are URLs, and is more consistent
with how we specify versions elsewhere.


justinsb:
/assign @hakman 

This should fix things (after I broke them!)

hakman (approved):

PR #22126, opened 2021-05-10
fix: make expectedHeadSha optional

hi-rustin:
This is an optional parameter. It has default value.


prow/github/client.go
```diff
@@ -3976,7 +3976,7 @@ func (c *client) ListPRCommits(org, repo string, number int) ([]RepositoryCommit
 //
 // GitHub API docs: https://developer.github.com/v3/pulls#update-a-pull-request-branch
 func (c *client) UpdatePullRequestBranch(org, repo string, number int, expectedHeadSha *string) error {
-	durationLogger := c.log("UpdatePullRequestBranch", org, repo, *expectedHeadSha)
```
alvaroaleman, line 3979: Fix for what? Please provide a proper description and a testcase.

hi-rustin: So I think using the default value might help a bit. But the method will be Panic anyway when it is nil , so the main thing is to make it really optional.

alvaroaleman (approved):

PR #22134, opened 2021-05-10
[jenkins-operator] Add ability to trigger jobs not in the root path o…

arturo-skydio:
…f jenkins

Currently, there is no way to trigger a jenkins job that has a subpath, this change allows the user to specify, if not there, then it defaults to the prow job name


matthyx:
is this still a valid PR?

PR #22135, opened 2021-05-10
[jenkins-operator] Allow specifying additional params for downstream …

arturo-skydio:
…jenkins jobs


matthyx:
Hi @arturo-skydio could you add a simple test case, making sure we don't break.

matthyx:
@arturo-skydio can you check if you didn't rebase from the wrong branch... 11 commits looking like a mistake.

PR #22145, opened 2021-05-11
Checkconfig: Validate additional plugins configs are in org/repo hirarchy

alvaroaleman:
This extends checkconfig to also be able to check if additional plugin configs are in an org repo hirarchy. This doesn't have any actual use yet, because flags to actually configure that aren't plugged through yet, I will do that in a follow-up.

### DIFF
--- a/prow/cmd/checkconfig/main.go
+++ b/prow/cmd/checkconfig/main.go
@@ -368,7 +368,7 @@ func validate(o options) error {
 
 	if o.warningEnabled(validateSupplementalProwConfigOrgRepoHirarchy) {
 		for _, supplementalProwConfigDir := range o.config.SupplementalProwConfigDirs.Strings() {
-			errs = append(errs, validateAdditionalProwConfigIsInOrgRepoDirectoryStructure(supplementalProwConfigDir, os.DirFS("./"), o.config.SupplementalProwConfigsFileNameSuffix))
+			errs = append(errs, validateAdditionalProwConfigIsInOrgRepoDirectoryStructure(supplementalProwConfigDir, os.DirFS("./"), o.config.SupplementalProwConfigsFileNameSuffix, "not/implemented"))
 		}
 	}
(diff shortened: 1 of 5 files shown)


prow/cmd/checkconfig/main.go
```diff
@@ -1283,3 +1276,26 @@ func prefixWithAndIfNeeded(s string, needsAnd bool) string {
...
 }
+
+func getSupplementalConfigScope(path string, filesystem fs.FS, supplementalProwConfigsFileNameSuffix string) (global bool, orgs sets.String, repos sets.String, err error) {
+
+	var cfg interface {
+		HasConfigFor() (bool, sets.String, sets.String)
+	}
+	if strings.HasSuffix(path, supplementalProwConfigsFileNameSuffix) {
+		cfg = &config.Config{}
+	} else {
```
stevekuznetsov, line 1287: This is a weak inference and requires that the caller continue to run the validation before calling this - might be nice to pass in an enum or the type to marshal into itself and not expose this function to the suffixes?

prow/plugins/config.go
```diff
@@ -1804,3 +1805,20 @@ func (p *Plugins) mergeFrom(other *Plugins) error {
 
 	return utilerrors.NewAggregate(errs)
 }
+
+func (c *Configuration) HasConfigFor() (global bool, orgs sets.String, repos sets.String) {
+	if !reflect.DeepEqual(c, &Configuration{Plugins: c.Plugins}) {
+		global = true
```
stevekuznetsov, line 1811: Not sure I follow - a config is global if it contains more than just plugins? Why? Since that's the only sub-section we've added sharding for thus far?

alvaroaleman: Yeah, because the plugins are the only part we implemented merging for

stevekuznetsov (approved):

Issue #22148, opened 2021-05-11
Prometheus alert didn't fire as expected


chaodaiG:
As discussed in https://github.com/kubernetes/test-infra/issues/22147, that prometheus indicated an alert fired, but it was not posted on https://kubernetes.slack.com/archives/C7J9RP96G as expected

chaodaiG:
https://github.com/kubernetes/test-infra/pull/22149 had merged, if it does fix then there should be alerts in https://kubernetes.slack.com/archives/C7J9RP96G some time tonight

chaodaiG:
It appears that alertmanagerconfig has an enforcement of namespace matching, which doesn't work with blackbox alerts yet. prometheus-operator/prometheus-operator#3737 is the discussion and the implementation is under work.

PR #22160, opened 2021-05-12
Revert "Prometheus alertmanager use alertmanagerconfig CR instead of …

chaodaiG:
…plain text secret"

Fixes: https://github.com/kubernetes/test-infra/issues/22148

Reverts https://github.com/kubernetes/test-infra/pull/22149 and  https://github.com/kubernetes/test-infra/pull/21675

It appears that alertmanagerconfig has an enforcement of namespace matching, which doesn't work with blackbox alerts yet. https://github.com/prometheus-operator/prometheus-operator/issues/3737 is the discussion and the implementation is under work.

Revert the PRs to restore the blackbox alerting behavior

/hold
I'll need to manually create the secret after this PR merges


cjwagner (approved):

chaodaiG:
/unhold
Will manually restore the secret once PR merged

PR #22173, opened 2021-05-13
Prow: Plug supplemental plugin config files through

alvaroaleman:
This PR adds a new pluginsconfig flagutil that includes flags for
supplemental plugins config files. It is in its own package to avoid
making everything that uses the standarf flagutil depend transiently on
the plugins.


prow/cmd/checkconfig/main.go
```diff
@@ -1118,7 +1123,24 @@ func validateCluster(cfg *config.Config) error {
 	return utilerrors.NewAggregate(errs)
 }
 
-func validateAdditionalProwConfigIsInOrgRepoDirectoryStructure(root string, filesystem fs.FS, supplementalProwConfigsFileNameSuffix, supplementalPluginsConfigFileNameSuffix string) error {
+func validateAdditionalProwConfigIsInOrgRepoDirectoryStructure(filesystem fs.FS, supplementalProwConfigDirs, supplementalPluginsConfigDirs []string, supplementalProwConfigsFileNameSuffix, supplementalPluginsConfigFileNameSuffix string) error {
+	var errs []error
+
+	for _, supplementalProwConfigDir := range supplementalPluginsConfigDirs {
+		if err := validateAdditionalConfigIsInOrgRepoDirectoryStructure(supplementalProwConfigDir, filesystem, func() hirarchicalConfig { return &config.Config{} }, supplementalProwConfigsFileNameSuffix); err != nil {
```
e-blackwelder, line 1130: [nit] spelling: "hierarchical"

prow/cmd/checkconfig/main.go
```diff
@@ -1118,7 +1123,24 @@ func validateCluster(cfg *config.Config) error {
...
+			errs = append(errs, err)
+		}
+	}
+	for _, supplementalPluginsConfigDir := range supplementalPluginsConfigDirs {
+		if err := validateAdditionalConfigIsInOrgRepoDirectoryStructure(supplementalPluginsConfigDir, filesystem, func() hirarchicalConfig { return &plugins.Configuration{} }, supplementalPluginsConfigFileNameSuffix); err != nil {
```
e-blackwelder, line 1135: ditto

e-blackwelder (commented):
/lgtm
/hold

prow/cmd/checkconfig/main.go
alvaroaleman: fixed

prow/flagutil/plugins/plugins.go
```diff
@@ -0,0 +1,62 @@
...
+	PluginConfigPathDefault                  *string
+	SupplementalPluginsConfigDirs            flagutil.Strings
+	SupplementalPluginsConfigsFileNameSuffix string
+	CheckUnknownPlugins                      bool
+}
+
+func (o *PluginOptions) AddFlags(fs *flag.FlagSet) {
+	defaultConfigPath := DefaultConfigPath
+	if o.PluginConfigPathDefault != nil {
+		defaultConfigPath = *o.PluginConfigPathDefault
```
cjwagner, line 44: Nit: I think `AddFlags` should default to not specifying a `defaultConfigPath` so that `PluginConfigPathDefault` only needs to be changed to set a default. Since we want to encourage explicitly setting the plugin config path it seems better to not provide a default value unless the binary explicitly requests otherwise.

prow/cmd/checkconfig/main.go
```diff
@@ -346,11 +351,11 @@ func validate(o options) error {
 		}
 	}
 	if pcfg != nil && o.warningEnabled(unknownFieldsWarning) {
-		pcfgBytes, err := ioutil.ReadFile(o.pluginConfig)
+		pcfgBytes, err := ioutil.ReadFile(o.pluginsConfig.PluginConfigPath)
 		if err != nil {
 			return fmt.Errorf("error reading Prow plugin config for validation: %w", err)
 		}
-		if err := validateUnknownFields(&plugins.Configuration{}, pcfgBytes, o.pluginConfig); err != nil {
+		if err := validateUnknownFields(&plugins.Configuration{}, pcfgBytes, o.pluginsConfig.PluginConfigPath); err != nil {
```
cjwagner, line 358: Unrelated to this PR, but note that the unknown fields check doesn't look at supplemental prow or plugin config.

cjwagner (approved):
I have one nit about the plugin config path defaulting.
/hold

prow/cmd/checkconfig/main.go
alvaroaleman: Good point

prow/flagutil/plugins/plugins.go
alvaroaleman: Good point, it was mostly there because I was mirroring the status quo where we have it. I've inversed it so that everything that wants a default needs to explicitly configure the PluginOptions.

cjwagner (approved):

PR #22174, opened 2021-05-13
bumpmonitoring package uses prcreator package for automatically creating PR

chaodaiG:


experiment/bumpmonitoring/main.go
```diff
@@ -46,19 +51,92 @@ var (
...
+	var config string
+	var labelsOverride []string
+	var skipPullRequest bool
+
+	flag.StringSliceVar(&labelsOverride, "labels-override", nil, "Override labels to be added to PR.")
+	flag.BoolVar(&skipPullRequest, "skip-pullrequest", false, "")
+	flag.Parse()
+
+	var o options
+	data, err := ioutil.ReadFile(config)
```
cjwagner, line 69: `config` is unset when used here so this will always fail.

experiment/bumpmonitoring/main.go
```diff
@@ -46,19 +51,92 @@ var (
...
+	return nil
+}
+
+var _ bumper.PRHandler = (*client)(nil)
+
 type client struct {
-	srcPath string
-	dstPath string
+	SrcPath string
+	DstPath string
```
cjwagner, line 99: nit: I don't think these need to be exported.

cjwagner (commented):
Looks good except that the config file cannot be specified.

experiment/bumpmonitoring/main.go
chaodaiG: good point, done

experiment/bumpmonitoring/main.go
chaodaiG: nice catch, fixed

cjwagner (approved):

PR #22176, opened 2021-05-13
Fix bind-service-account.sh to work with GCP default compute engine service accounts.

cjwagner:
The assumption that GCP service accounts are all of the form `<someone>@<project>.iam.gserviceaccount.com` doesn't hold for default compute engine service accounts which are of the form `<project-number>-compute@developer.gserviceaccount.com`.
/assign @fejta


chaodaiG (approved):
/hold

PR #22177, opened 2021-05-13
Update create-build-cluster.sh to make pod utils use workload identit…

cjwagner:
…y and generate default decoration config.

/assign @mpherman2

### DIFF
--- a/prow/create-build-cluster.sh
+++ b/prow/create-build-cluster.sh
@@ -25,12 +25,9 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
-ROOT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd -P)"
-
 # Specific to Prow instance
 PROW_INSTANCE_NAME="${PROW_INSTANCE_NAME:-}"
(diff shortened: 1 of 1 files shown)


prow/create-build-cluster.sh
```diff
@@ -168,26 +191,37 @@ function createUploadSASecret() {
...
   fi
 }
 
-origdir="$( pwd -P )"
-tempdir="$( mktemp -d )"
-echo
-echo "Temporary files produced are stored at: ${tempdir}"
-echo
+function genConfig() {
+  # TODO: Automatically inject this into config.yaml at the same time as kubeconfig credential setup (which auto creates a CL we can include this in).
```
chaodaiG, line 200:
```suggestion
  # TODO: Automatically inject this into config.yaml at the same time as kubeconfig credential setup (which auto creates a PR we can include this in).
```

chaodaiG (approved):
/hold
left a nit

prow/create-build-cluster.sh
cjwagner: Committed.

chaodaiG (approved):
/retest

PR #22185, opened 2021-05-14
bumper: restore GH login/mail detection

petr-muller:
Looks like this snippet from `Run()` was left out for some reason when `Run2()`
was created: deriving name is especially useful b/c it is needed to assemble the
push target remote URL.

/cc @alvaroaleman @chaodaiG


alvaroaleman (approved):

chaodaiG:
Thank you @petr-muller !

PR #22193, opened 2021-05-15
drop csi-hostpath driver

aojea:
We dopped them from the presubmits, but the periodic in release blocking keep running them

Long history
https://github.com/kubernetes/test-infra/pull/22025
https://github.com/kubernetes/kubernetes/issues/101275

The release blocking jobs are a christmas tree
https://testgrid.k8s.io/sig-release-master-blocking#kind-master-parallel&width=5

### DIFF
--- a/config/jobs/kubernetes-sigs/kind/kind-release-blocking.yaml
+++ b/config/jobs/kubernetes-sigs/kind/kind-release-blocking.yaml
@@ -34,7 +34,7 @@ periodics:
         value: "."
       # TODO(bentheelder): reduce the skip list further
       - name: SKIP
-        value: \[Slow\]|\[Disruptive\]|\[Flaky\]|\[Feature:.+\]|PodSecurityPolicy|LoadBalancer|load.balancer|In-tree.Volumes.\[Driver:.nfs\]|PersistentVolumes.NFS|Network.should.set.TCP.CLOSE_WAIT.timeout|Simple.pod.should.support.exec.through.an.HTTP.proxy|subPath.should.support.existing|should.provide.basic.identity|\[NodeFeature:PodReadinessGate\]
+        value: \[Slow\]|\[Disruptive\]|\[Flaky\]|\[Feature:.+\]|PodSecurityPolicy|LoadBalancer|load.balancer|In-tree.Volumes.\[Driver:.nfs\]|PersistentVolumes.NFS|Network.should.set.TCP.CLOSE_WAIT.timeout|Simple.pod.should.support.exec.through.an.HTTP.proxy|subPath.should.support.existing|should.provide.basic.identity|\[NodeFeature:PodReadinessGate\]|Driver:.csi-hostpath
       - name: PARALLEL
         value: "true"
       # we need privileged mode in order to do docker in docker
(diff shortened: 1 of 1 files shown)


BenTheElder:
Please see the discussion in the tracking issue and in release CI signal. We are still early in the release and the intent was to get these healthy again.

If we think progress is going too slowly we need to add them to a new job where progress can still be made.

aojea:
My concern is that this can mask new failures and we loose all the history, since right now the jobs are randomly failing

BenTheElder:
https://github.com/kubernetes/kubernetes/issues/101913#issuecomment-839986455

cc @msau42 @pohly 

I agree that there is this concern, but hopefully this will be resolved. If we near freeze we will need to drop these though.

If you want you can drop these _but setup a new job(s) to cover them_ seperately?

msau42:
We have not been able to make much progress with root causing. After doing some manual runs, we found that the tests seemed very flaky even before the drivers were updated. But also from the manual runs, we can't get a good sense of flake percentage.

So far our efforts are being spent to triage each failure. They're revealing some very unusual behavior, some known, some new, but they're complex so getting to the root cause will take some time and fixing them may also take a while.

So that being said, I am ok with removing it from the blocking tests, and adding a new non-blocking job where we can continue triaging. In general, we are still getting CSI signal with the [gce jobs](https://k8s-testgrid.appspot.com/sig-storage-kubernetes#gce), which are nowhere near as flaky. My only guess is that perhaps the kind environment is more resource constrained so more races are being hit.

pohly:
Is the new job going to run just the CSI tests or everything? The problem is that the CSI tests in isolation seem less flaky, so a job just running those will not be very useful and to some extend redundant - we already have https://testgrid.k8s.io/sig-storage-csi-ci#canary-on-master

aojea:
> The problem is that the CSI tests in isolation seem less flaky,

I recommend always to run the tests at least with the conformance suite in parallel, the interaction with other tests matter as we can see here

> So that being said, I am ok with removing it from the blocking tests, and adding a new non-blocking job where we can continue triaging. In general, we are still getting CSI signal with the gce jobs, which are nowhere near as flaky. My only guess is that perhaps the kind environment is more resource constrained so more races are being hit.

Kind environment adds much more stress since everything is "faster", i.e. for the network you have an ideal environment since there is practically no latency nor packet drops, the bandwidth is about 37 gpbs ... so we hit bottlenecks that are not common in other environment where we are protected by the slowness of the network



> I agree that there is this concern, but hopefully this will be resolved. If we near freeze we will need to drop these though.

let's put it on hold, at least a few more weeks, if we move it to other place and considering the enormous backlog that people here handle it is likely we "deprioritise" it :upside_down_face: , at least thinking in me

aojea:
/close

this has improved a lot since 05-18
https://testgrid.k8s.io/sig-release-master-blocking#kind-master-parallel&width=5
@msau42 @pohly does it ring a bell?

msau42:
nope... we have not made any changes to storage tests or storage bug fixes in the past week that could impact this. I think the hostpath tests may have potentially be impacted by general system issues.

BenTheElder:
iinnnnteresting. something clearly changed. I don't think it was the infra 🤔 

This is a rough changeset from testgrid over the days it changed:
https://github.com/kubernetes/kubernetes/compare/1d3808493...2d58b7288

maybe:
- https://github.com/kubernetes/kubernetes/pull/101888
- https://github.com/kubernetes/kubernetes/pull/101712

liggitt:
the runc bump was intended to address a severe performance regression

BenTheElder:
And actually ... https://github.com/kubernetes-sigs/kind/commits/main

On may 18th in kind: https://github.com/kubernetes-sigs/kind/commit/86054ae3ecbfb14b3806b71c37b5bbb5dc1932dc we picked up runc94 there with containerd 1.5.1

We've since picked up runc rc95 + containerd 1.5.2

aojea:
the runc94 was reverted by jordan a few hours later https://github.com/kubernetes/kubernetes/pull/102094

liggitt:
and came back in in runc r95 in https://github.com/kubernetes/kubernetes/pull/102147

aojea:
yeah, but in May 21st, the improvement started on May 18th

BenTheElder:
I'm not clear what all it affected but we updated the actual runc _binary_ and some other binary dependencies in kind around the same time and that was not reverted.

BenTheElder:
we should probably bring back this suite in presubmit.

BenTheElder:
https://prow.k8s.io/?job=pull-kubernetes-e2e-kind is pretty green

> Success rate over time: 3h: 78%, 12h: 89%, 48h: 90%

PR #22197, opened 2021-05-16
add e2e for aws-lb-controller

olemarkus:
/cc @rifelpet


rifelpet (approved):
can you drop the frequency of the CSI test now that it is working?

PR #22200, opened 2021-05-17
added periodical for depstat

RinkiyaKeDad:

Added a periodical job for depstat based on the [previous one](https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/sig-arch/kubernetes-depstat.yaml). Please let me know if [name](https://github.com/kubernetes/test-infra/compare/master...RinkiyaKeDad:depstat_periodical?expand=1#diff-9ae83e83dcb890a759fd567fc08d0b8c2bd4671f11ece4fe79e1ad5b62b58514R4), [description](https://github.com/kubernetes/test-infra/compare/master...RinkiyaKeDad:depstat_periodical?expand=1#diff-9ae83e83dcb890a759fd567fc08d0b8c2bd4671f11ece4fe79e1ad5b62b58514R30) or something else needs to be changed.

/assign @dims

### DIFF
--- a/config/jobs/kubernetes/sig-arch/kubernetes-depstat-periodical.yaml
+++ b/config/jobs/kubernetes/sig-arch/kubernetes-depstat-periodical.yaml
@@ -1,0 +1,30 @@
+periodics:
+  kubernetes/kubernetes:
+  - interval: 6h
+    name: check-dependency-stats-periodical
+    decorate: true
+    path_alias: k8s.io/kubernetes
+    always_run: false
+    optional: true
+    spec:
+      containers:
+      - image: golang
+        command:
+        - /bin/bash
+        args:
+        - -c
+        - |
+          set -euo; \
+          export PATH=$PATH:$GOPATH/bin; \
+          mkdir -p "${ARTIFACTS}"; \
+          pushd "$ARTIFACTS"; \
+          go install github.com/kubernetes-sigs/depstat@latest; \
+          popd; \
+          depstat stats --json > "${ARTIFACTS}/stats.json"; \
+          git checkout -b base "${PULL_BASE_SHA}"; \
+          depstat stats --json > "${ARTIFACTS}/stats-base.json"; \
+          diff -s --ignore-all-space "${ARTIFACTS}"/stats-base.json "${ARTIFACTS}"/stats.json || true;
+    annotations:
+      testgrid-create-test-group: "true"
+      testgrid-dashboards: sig-testing-misc
+      description: Generates dependency statistics by running depstat for base and current PR periodically
--- a/config/jobs/kubernetes/sig-arch/kubernetes-depstat-periodical.yaml
+++ b/config/jobs/kubernetes/sig-arch/kubernetes-depstat-periodical.yaml
@@ -1,9 +1,12 @@
 periodics:
-  kubernetes/kubernetes:
   - interval: 6h
     name: check-dependency-stats-periodical
     decorate: true
-    path_alias: k8s.io/kubernetes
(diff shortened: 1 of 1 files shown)


RinkiyaKeDad:
Test failing with the error:
```
ERROR: /home/prow/go/src/github.com/kubernetes/test-infra/config/tests/testgrids/BUILD.bazel:25:8: Executing genrule //config/tests/testgrids:testgrid_config failed (Exit 1)
6274
2021/05/17 13:25:10 FAIL: couldn't load prow config: error unmarshaling config/jobs/kubernetes/sig-arch/kubernetes-depstat-periodical.yaml: error unmarshaling JSON: while decoding JSON: json: cannot unmarshal object into Go struct field JobConfig.periodics of type []config.Periodic
```

I think it's because the way I specified it is incorrect. Will try the way it has been done [here](https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/sig-testing/coverage.yaml#L9-L13).

config/jobs/kubernetes/sig-arch/kubernetes-depstat-periodical.yaml
```diff
@@ -0,0 +1,33 @@
+periodics:
+  - interval: 6h
+    name: check-dependency-stats-periodical
+    decorate: true
+    extra_refs:
+    - org: kubernetes
+      repo: kubernetes
+      base_ref: master
```
RinkiyaKeDad, line 8: This will be `master`, right? Confirming cause this might mean that initially we are checked into master instead of the PR commit.

config/jobs/kubernetes/sig-arch/kubernetes-depstat-periodical.yaml
```diff
@@ -0,0 +1,33 @@
...
+      path_alias: k8s.io/kubernetes
+    always_run: false
+    optional: true
```
RinkiyaKeDad, line 11: Are these two needed here? I didn't see them in [this](https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/sig-testing/coverage.yaml#L2-L23) periodical?

config/jobs/kubernetes/sig-arch/kubernetes-depstat-periodical.yaml
```diff
@@ -0,0 +1,33 @@
...
+    spec:
```
RinkiyaKeDad, line 12: I'm not sure what [`decoration_config`](https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/sig-testing/coverage.yaml#L22-L23) means. Do we need that here right before spec?

RinkiyaKeDad (commented):
@dims it's all green but I have some doubts on if I've done this correctly or not.

config/jobs/kubernetes/sig-arch/kubernetes-depstat-periodical.yaml
dims: yes master is good

config/jobs/kubernetes/sig-arch/kubernetes-depstat-periodical.yaml
dims: you don't need these 2

config/jobs/kubernetes/sig-arch/kubernetes-depstat-periodical.yaml
dims: if we used it in the presubmit then keep it, else remove it.

PR #22210, opened 2021-05-17
Add tool for assiting onboarding to gerrit Prow

mpherman2:
This was getting quite long, so I will add functionality that automatically creates a PR for the change in my next update. 

/assign @chaodaiG 
/assign @cjwagner

### DIFF
--- a/experiment/gerrit-onboarder/onboarder.go
+++ b/experiment/gerrit-onboarder/onboarder.go
@@ -1,0 +1,357 @@
+package main
+
+import (
+	"errors"
+	"flag"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"strings"
+
+	"github.com/sirupsen/logrus"
+
+	"k8s.io/test-infra/prow/cmd/generic-autobumper/bumper"
+)
+
+const (
+	UUID                = 0
+	GROUP_NAME          = 1
+	GROUPS_FILE         = "groups"
+	PROJECT_CONFIG_FILE = "project.config"
+
+	ACCESS_HEADER            = "[access \"refs/*\"]"
+	PROW_READ_ACCESS_FORMAT  = "read = group %s"
+	PROW_LABEL_ACCESS_FORMAT = "label-Verified = -1..+1 group %s"
+	LABEL_HEADER             = "[label \"Verified\"]"
+)
+
+type Options struct {
+	Host       string
+	Repo       string
+	UUID       string
+	GroupName  string
+	BranchName string
+}
+
+func getAllLabelLines() []string {
+	return []string{
+		"function = MaxWithBlock",
+		"value = -1 Failed",
+		"value = 0 No score",
+		"value = +1 Verified",
+		"copyAllScoresIfNoCodeChange = true",
+		"defaultValue = 0",
+	}
+}
+
+func parseAndValidateOptions() (*Options, error) {
+	var o Options
+	flag.StringVar(&o.Host, "host", "", "The gerrit host.")
+	flag.StringVar(&o.Repo, "repo", "", "The gerrit Repo.")
+	flag.StringVar(&o.UUID, "uuid", "", "The UUID to be added to the file.")
+	flag.StringVar(&o.GroupName, "group", "", "The corresponding group name for the UUID.")
+	flag.StringVar(&o.BranchName, "branch", "", "The name of the branch where refs/meta/config will live")
+	flag.Parse()
+
+	if o.Host == "" || o.Repo == "" || o.UUID == "" || o.GroupName == "" || o.BranchName == "" {
+		return &o, errors.New("all flags are required")
+	}
+
+	return &o, nil
+}
+
+func intMax(x, y int) int {
+	if x > y {
+		return x
+	}
+	return y
+}
+
+func maxIDLen(values []string) int {
+	max := 0
+	for _, item := range values {
+		max = intMax(max, len(item))
+	}
+	return intMax(max, len("# UUID"))
+}
+
+func getFormatString(maxLine int) string {
+	return "%-" + fmt.Sprintf("%d", maxLine) + "v\t%s\n"
+}
+
+func mapToGroups(groupsMap map[string]string, orderedUUIDs []string) string {
+	maxLine := maxIDLen(orderedUUIDs)
+	groups := fmt.Sprintf(getFormatString(maxLine), "# UUID", "Group Name") + "#\n"
+
+	for _, id := range orderedUUIDs {
+		groups = groups + fmt.Sprintf(getFormatString(maxLine), id, groupsMap[id])
+	}
+	return groups
+}
+
+func groupsToMap(groupsFile string) (map[string]string, []string) {
+	orderedKeys := []string{}
+	groupsMap := map[string]string{}
+	lines := strings.Split(groupsFile, "\n")
+	for _, line := range lines {
+		if !strings.HasPrefix(line, "#") && line != "" {
+			pair := strings.Split(line, "\t")
+			orderedKeys = append(orderedKeys, strings.TrimSpace(pair[UUID]))
+			groupsMap[strings.TrimSpace(pair[UUID])] = strings.TrimSpace(pair[GROUP_NAME])
+		}
+	}
+	return groupsMap, orderedKeys
+}
+
+func ensureUUID(groupsFile, id, group string) string {
+	groupsMap, orderedKeys := groupsToMap(groupsFile)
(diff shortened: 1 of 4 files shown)


experiment/gerrit-onboarder/onboarder.go
```diff
@@ -0,0 +1,357 @@
...
+		return fmt.Errorf("failed to switch to new branch, %w", err)
+	}
+
+	return nil
+}
+
+func main() {
+	o, err := parseAndValidateOptions()
+	if err != nil {
+		logrus.WithError(err).Fatal("Failed to run onboarding tool")
```
chaodaiG, line 326: Would expect a better targeted error, otherwise just `logrus.Fatal(err)`, same below

experiment/gerrit-onboarder/onboarder.go
```diff
@@ -0,0 +1,357 @@
...
+	}
+}
+
+func parseAndValidateOptions() (*Options, error) {
+	var o Options
+	flag.StringVar(&o.Host, "host", "", "The gerrit host.")
+	flag.StringVar(&o.Repo, "repo", "", "The gerrit Repo.")
+	flag.StringVar(&o.UUID, "uuid", "", "The UUID to be added to the file.")
+	flag.StringVar(&o.GroupName, "group", "", "The corresponding group name for the UUID.")
+	flag.StringVar(&o.BranchName, "branch", "", "The name of the branch where refs/meta/config will live")
```
chaodaiG, line 53: branch is not a gerrit thing imo, it's a local ref instead of remote. can we use a random string here instead of asking user to supply?

experiment/gerrit-onboarder/onboarder.go
```diff
@@ -0,0 +1,357 @@
...
+		if !strings.HasPrefix(line, "#") && line != "" {
+			pair := strings.Split(line, "\t")
+			orderedKeys = append(orderedKeys, strings.TrimSpace(pair[UUID]))
+			groupsMap[strings.TrimSpace(pair[UUID])] = strings.TrimSpace(pair[GROUP_NAME])
+		}
+	}
+	return groupsMap, orderedKeys
+}
+
+func ensureUUID(groupsFile, id, group string) string {
```
chaodaiG, line 106: nit: rename `id` to `uuid`

experiment/gerrit-onboarder/onboarder.go
```diff
@@ -0,0 +1,357 @@
...
+	groupsMap, orderedKeys := groupsToMap(groupsFile)
+	if value, ok := groupsMap[id]; ok && group == value {
```
chaodaiG, line 108: This means override later if group not equal to value, when would this happen?

experiment/gerrit-onboarder/onboarder.go
```diff
@@ -0,0 +1,357 @@
...
+		return groupsFile
+	}
+	groupsMap[id] = group
+	orderedKeys = append(orderedKeys, id)
```
chaodaiG, line 112: Would there be duplicated keys when override?

experiment/gerrit-onboarder/onboarder.go
```diff
@@ -0,0 +1,357 @@
...
+	}
+	return groups
+}
+
+func groupsToMap(groupsFile string) (map[string]string, []string) {
+	orderedKeys := []string{}
+	groupsMap := map[string]string{}
+	lines := strings.Split(groupsFile, "\n")
+	for _, line := range lines {
+		if !strings.HasPrefix(line, "#") && line != "" {
```
chaodaiG, line 97: Comments are ignored here and lost later, I think they should be kept

experiment/gerrit-onboarder/onboarder.go
```diff
@@ -0,0 +1,357 @@
...
+		return false
+	}
+}
+
+func verifyInTree(host, cur_branch string, configMap map[string][]string, verify func(map[string][]string) bool) (bool, error) {
+	if verify(configMap) {
+		return true, nil
+	} else if inheritance := getInheritedRepo(configMap); inheritance != "" {
+		parent_branch := cur_branch + "_parent"
+		if err := fetchMetaConfig(host, inheritance, parent_branch); err != nil {
```
chaodaiG, line 231: This is fetching in the same directory, which might not be easy to understand/debug, can we do it in a separate directory?

experiment/gerrit-onboarder/onboarder.go
```diff
@@ -0,0 +1,357 @@
...
+		}
+		return false
+	}
+}
+
+func labelAccessExistsFunc(groupName string) func(map[string][]string) bool {
+	return func(configMap map[string][]string) bool {
+		for _, value := range configMap {
+			for _, item := range value {
+				if strings.HasPrefix(strings.TrimSpace(item), "label-Verified =") && strings.HasSuffix(strings.TrimSpace(item), fmt.Sprintf("group %s", groupName)) {
```
chaodaiG, line 217: A comment here would be helpful to readers, briefly explaining that this tool is best effort and not to be intrusive. So when target label is already associated with prow robot, then assume it's already done, and no PR is needed.

Also, please use either a global var or an option for the name of the label

experiment/gerrit-onboarder/onboarder.go
```diff
@@ -0,0 +1,357 @@
...
+	return false, nil
+}
+
+func ensureProjectConfig(config, host, cur_branch, groupName string) (string, error) {
+	configMap, orderedKeys := configToMap(config)
+
+	// Check that prow automation robot has access to refs/*
+	accessLines := []string{}
+	readAccessLine := fmt.Sprintf(PROW_READ_ACCESS_FORMAT, groupName)
+	prowReadAccess, err := verifyInTree(host, cur_branch, configMap, lineInRightHeaderFunc(ACCESS_HEADER, readAccessLine))
```
chaodaiG, line 260: Other than `[access "refs/*"]`, it could be anything else like `[access "refs/heads/*"]`, or `[access "refs/heads/main"]`, or even more. Would suggest make it more flexible, like anything matching `[access "refs/.*`

experiment/gerrit-onboarder/onboarder.go
mpherman2: This would happen if the UUID is already being used, but with a different group name. I think this case should probably just error. Adding that now.

experiment/gerrit-onboarder/onboarder.go
mpherman2: See above. Now erroring on dupe keys.

experiment/gerrit-onboarder/onboarder.go
mpherman2: Done

experiment/gerrit-onboarder/onboarder.go
mpherman2: Do you mean clone the repo in a temp dir and create the PR there? I prefer doing it in their own copy of the repo because that way users can make changes to the PR easily if needed. Why would doing it in another dir be easier to debug?

chaodaiG: IIUC this is fetching a different repo, generally not the best practice in git to clone different repos under the same path. But since `refs/meta/configs` is so different, I think it might be fine. So up to you to decide

experiment/gerrit-onboarder/onboarder.go
```diff
@@ -0,0 +1,418 @@
...
+	groupsFile        = "groups"
+	projectConfigFile = "project.config"
+
+	accessHeader          = `[access "refs/*"]`
+	prowReadAccessFormat  = "read = group %s"
+	prowLabelAccessFormat = "label-Verified = -1..+1 group %s"
+	labelHeader           = `[label "Verified"]`
+	labelEquals           = "label-Verified ="
+
+	accessRefsRegex = `^\[access "refs\/.*"\]`
```
chaodaiG, line 47: `regexp.MustCompile` this as global var to avoid runtime regexp error

experiment/gerrit-onboarder/onboarder.go
mpherman2: My original plan was to create a temp dir, clone the parent repo, and fetching from there, but I realized this was a lot easier. Given that we are just reading and not writing to the other repos I don't think we will run into issues. If we do, we can make the change in a future PR.

experiment/gerrit-onboarder/onboarder.go
mpherman2: Done

PR #22212, opened 2021-05-17
Adjust the timeout to 40m for scheduler-plugins integration test

Huang-Wei:
The scheduler-plugins doesn't store /vendor folder, also the integration test may need to download and install etcd. These steps take a certain amount of execution time in the integration test suite. Recently we added several integration tests, and noticed the test times out frequent than before.

This PR specifies the timeout as 40m (default value is 10m?) to help us narrow down the root cause: if the timeout is due to coding bugs, or pure time out.

### DIFF
--- a/config/jobs/kubernetes-sigs/scheduler-plugins/scheduler-plugins-presubmits-master.yaml
+++ b/config/jobs/kubernetes-sigs/scheduler-plugins/scheduler-plugins-presubmits-master.yaml
@@ -42,6 +42,8 @@ presubmits:
         - unit-test
   - name: pull-scheduler-plugins-integration-test-master
     decorate: true
+    decoration_config:
+      timeout: 40m
     path_alias: sigs.k8s.io/scheduler-plugins
     branches:
     - ^master$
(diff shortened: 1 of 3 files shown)


seanmalloy (approved):

seanmalloy:
The changes look good me to, but I can't approve because I'm not an org member.

/assign @denkensk

Huang-Wei:
Hmm, it seems the integration test still timed out immediately after 10 mins: https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_scheduler-plugins/193/pull-scheduler-plugins-integration-test-master/1397241162464497664

@BenTheElder could you kindly shed some light? Is that the timeout config isn't configured properly?

```yaml
    decoration_config:
      timeout: 40m
```

BenTheElder:
The decoration config timeout sets the timeout for the jobs at the prow level, our default timeout on prow.k8s.io is actually like 2 hours I think?

> panic: test timed out after 10m0s

That log line isn't coming from prow, I assume the integration test also has some timeout?

Huang-Wei:
Aha, I guess the default 10m timeout is coming from `go test`:

```
	-timeout d
	    If a test binary runs longer than duration d, panic.
	    If d is 0, the timeout is disabled.
	    The default is 10 minutes (10m).
```

Huang-Wei:
> our default timeout on prow.k8s.io is actually like 2 hours I think?

I will revert this PR as it's not related to prow setting after https://github.com/kubernetes-sigs/scheduler-plugins/pull/202 gets merged.

PR #22219, opened 2021-05-18
fix for depstat's periodical prow job

RinkiyaKeDad:

I modified the original script to just run install depstat and run it with the `--json` flag (instead of redirecting the output to a file like earlier). Is this what you had in mind @dims?

/assign @dims

### DIFF
--- a/config/jobs/kubernetes/sig-arch/kubernetes-depstat-periodical.yaml
+++ b/config/jobs/kubernetes/sig-arch/kubernetes-depstat-periodical.yaml
@@ -21,7 +21,7 @@ periodics:
           pushd "$ARTIFACTS"; \
           go install github.com/kubernetes-sigs/depstat@latest; \
           popd; \
-          depstat stats --json || true \
+          depstat stats --json || true; 
     annotations:
       testgrid-create-test-group: "true"
       testgrid-dashboards: sig-testing-misc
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes/sig-arch/kubernetes-depstat-periodical.yaml
```diff
@@ -21,11 +21,8 @@ periodics:
           pushd "$ARTIFACTS"; \
           go install github.com/kubernetes-sigs/depstat@latest; \
           popd; \
-          depstat stats --json > "${ARTIFACTS}/stats.json"; \
```
dims, line 24: we should print the output and store it into a file as well. will be useful for later automation, look at the `tee` command in linux

config/jobs/kubernetes/sig-arch/kubernetes-depstat-periodical.yaml
```diff
@@ -21,7 +21,7 @@ periodics:
           pushd "$ARTIFACTS"; \
           go install github.com/kubernetes-sigs/depstat@latest; \
           popd; \
-          depstat stats --json || true;
```
dims: do you really need `|| true`? is the depstat command expected to fail, i'd rather see RED when it does.

RinkiyaKeDad:
Forgot to ping you on this one @dims 😅 can we merge this now?

PR #22222, opened 2021-05-18
add optional gce job for L7/ingress presubmits

aojea:
We are bling trying to bump the ingress to v1, or at least I couldn't find a job with the whole suite of tests, 

https://github.com/kubernetes/kubernetes/pull/102030

so I think we can add this one to run optionally if some of the networking e2e tests changes


puerco:
/lgtm
Thanks a lot Antonio :+1:

PR #22226, opened 2021-05-18
Fix cgroup since gke is using systemd for cgroup manager

azhao155:
Verify it could run on gke.

### DIFF
--- a/config/jobs/kubernetes/minikube/minikube.yaml
+++ b/config/jobs/kubernetes/minikube/minikube.yaml
@@ -28,7 +28,7 @@ presubmits:
         - wrapper.sh
         - bash
         - -c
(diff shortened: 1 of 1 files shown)


azhao155:
@medyagh

config/jobs/kubernetes/minikube/minikube.yaml
```diff
@@ -28,7 +28,7 @@ presubmits:
         - wrapper.sh
         - bash
         - -c
-        - make && ./out/minikube start --force && kubectl get pods -A
+        - make && ./out/minikube start --force-systemd --force && kubectl get pods -A
```
medyagh, line 31: maybe add a comment that says --force because we are runnnig as root

medyagh (approved):

PR #22227, opened 2021-05-18
prow dashboard: add a panel showing current active prowjobs CR count

chaodaiG:
Current panels show cumulative counts of prowjobs CR, which doesn't provide too much value

### DIFF
--- a/config/prow/cluster/monitoring/mixins/grafana_dashboards/plank.jsonnet
+++ b/config/prow/cluster/monitoring/mixins/grafana_dashboards/plank.jsonnet
@@ -101,9 +101,10 @@ dashboard.new(
 )
 .addPanel(
     (graphPanel.new(
-        'number of Prow jobs by ${group_by_1}',
+        'number of Prow jobs (stacked) by ${group_by_1}',
         description='sum(prowjobs{...}) by (${group_by_1})',
         datasource='prometheus',
+        stack=true,
         legend_alignAsTable=true,
         legend_rightSide=true,
 
@@ -119,9 +120,10 @@ dashboard.new(
   })
 .addPanel(
     (graphPanel.new(
-        'number of Prow jobs by ${group_by_2}',
+        'number of Prow jobs (stacked) by ${group_by_2}',
         description='sum(prowjobs{...}) by (${group_by_2})',
         datasource='prometheus',
+        stack=true,
         legend_alignAsTable=true,
         legend_rightSide=true,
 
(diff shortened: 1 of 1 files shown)


config/prow/cluster/monitoring/mixins/grafana_dashboards/plank.jsonnet
```diff
@@ -119,7 +119,25 @@ dashboard.new(
...
-        'number of Prow jobs by ${group_by_2}',
+        'number of Prow jobs by ${group_by_1}',
+        description='prowjobs{...} by (${group_by_1})',
+        datasource='prometheus',
+        legend_alignAsTable=true,
+        legend_rightSide=true,
+
+    ) + legendConfig)
+    .addTarget(prometheus.target(
+        'prowjobs{job=~"plank|prow-controller-manager",cluster=~"${cluster}",org=~"${org}",repo=~"${repo}",state=~"${state}",type=~"${type}"} by (${group_by_1})',
```
cjwagner, line 130: What does ` by (${group_by_1})` do without an aggregation function? This doesn't look valid to me.

cjwagner (commented):
One alternative if we just want to be able to see the total number of ProwJob resources in the cluster would be to make the graph stacked.

config/prow/cluster/monitoring/mixins/grafana_dashboards/plank.jsonnet
chaodaiG: Changed to `sum(increase` in 5 minutes, seems to me this captures new jobs in the past 5 minutes, does this make sense?

cjwagner: Not quite, I guess I'm unclear what you are actually trying to show that the current graph does not show. I thought you just want the total number of PJs at a given time that match the filters (without the grouping), but now I'm unsure.

`increase` is not a valid function to use on the metric though. [`prowjobs`](https://github.com/kubernetes/test-infra/blob/f7e21a3c18f4f4bbc7ee170675ed53e4544a0632/prow/kube/metrics.go#L45-L48) is a Gauge metric and `increase` is only appropriate for Counter metrics. [ref](https://prometheus.io/docs/prometheus/latest/querying/functions/#increase)

chaodaiG: Actually I have misread https://monitoring.prow.k8s.io/d/e1778910572e3552a935c2035ce80369/plank-dashboard?orgId=1, the graph does show what I want, the only thing missing is stack, will update the PR

chaodaiG: Updated to make all 3 graphs stacking, let me know if this makes sense or not, would be happy to add new ones just for stack if the current behavior is desired

cjwagner: I think just the stacked graphs are fine, but could you make it clear they are stacked, please? e.g. indicate it in the graph title. Stacked graphs can be pretty misleading if it isn't clear they are stacked.

chaodaiG: done

cjwagner (approved):

PR #22236, opened 2021-05-18
enable dind for ingress presubmit

aojea:
fails to build because is not able to get to the docker socket

```
+++ [0518 22:20:43] Verifying Prerequisites....
Can't connect to 'docker' daemon.  please fix and retry.

````


aojea:
ping @cpanato

PR #22243, opened 2021-05-19
running depstat prow job automatically when dependencies change

RinkiyaKeDad:

As discussed in the last SIG-testing meeting, added `run_if_changed` for the depstat prow job so that it runs automatically whenever a PR with dependency changes is made.

/assign @dims @spiffxp

### DIFF
--- a/config/jobs/kubernetes/sig-arch/kubernetes-depstat.yaml
+++ b/config/jobs/kubernetes/sig-arch/kubernetes-depstat.yaml
@@ -5,6 +5,7 @@ presubmits:
     path_alias: k8s.io/kubernetes
     always_run: false
     optional: true
+    run_if_changed: '/go.(mod|sum)|vendor/'
     spec:
       containers:
       - image: golang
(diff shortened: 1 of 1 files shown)


RinkiyaKeDad:
@dims it's green, please merge :)

config/jobs/kubernetes/sig-arch/kubernetes-depstat.yaml
```diff
@@ -5,6 +5,7 @@ presubmits:
     path_alias: k8s.io/kubernetes
     always_run: false
     optional: true
+    run_if_changed: '/go.(mod|sum)|vendor/'
```
dims, line 8: did you see the slashes anywhere else @RinkiyaKeDad?

Can we keep it simple?
```
run_if_changed: '^(go.mod|go.sum|vendor)'
```

RinkiyaKeDad: > did you see the slashes anywhere else @RinkiyaKeDad?

I saw a trailing slash [here](https://github.com/kubernetes/test-infra/blob/master/config/jobs/image-pushing/k8s-staging-ingress-controller-conformance.yaml#L8) and at a couple of other places so I thought I should keep them both 😅 

Will update! :)

PR #22249, opened 2021-05-19
don't run IngressScale on presubmits

aojea:
The job is taking more than 2 hours, we don't need to run them here


ameukam:
/lgtm
/hold
Remove hold when ready to deploy.

aojea:
/hold cancel
go go go thanks

PR #22254, opened 2021-05-19
Crier gerrit reporter: don't ignore aborted jobs

chaodaiG:
When there are 5 jobs, 4 passed, 1 aborted, gerrit reports +1 on the PR, while it should report -1 instead. This PR fixes the problem.

The concern of reporting all aborted jobs from previous batch was addressed by ShouldReport function

### DIFF
--- a/prow/crier/reporters/gerrit/reporter.go
+++ b/prow/crier/reporters/gerrit/reporter.go
@@ -237,9 +237,6 @@ func (c *Client) Report(ctx context.Context, logger *logrus.Entry, pj *v1.ProwJo
 			}
 		}
 		for _, pjOnRevisionWithSameLabel := range mostRecentJob {
-			if pjOnRevisionWithSameLabel.Status.State == v1.AbortedState {
-				continue
-			}
 			toReportJobs = append(toReportJobs, pjOnRevisionWithSameLabel)
 		}
 	}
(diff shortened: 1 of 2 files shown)


chaodaiG:
CC @amwat

fejta (approved):

prow/crier/reporters/gerrit/reporter.go
```diff
@@ -237,9 +237,6 @@ func (c *Client) Report(ctx context.Context, logger *logrus.Entry, pj *v1.ProwJo
 			}
 		}
 		for _, pjOnRevisionWithSameLabel := range mostRecentJob {
-			if pjOnRevisionWithSameLabel.Status.State == v1.AbortedState {
-				continue
```
amwat, line 241: note that this might bring in more spam now.
Originally this was intended to skip legitimate aborted jobs. i.e. whenever a new patchset is uploaded

chaodaiG: This case was handled by `ShouldReport` as https://github.com/kubernetes/test-infra/blob/4f436d0b9431be68934a37799ed7069099391de2/prow/crier/reporters/gerrit/reporter.go#L198, so we should be safe removing this condition

chaodaiG: The race I was talking about was between:
- jobs from patchset `n` aborted when patchset `n+1` was pushed
- jobs from patchset `n+1` hasn't been triggered yet

In this race the condition of ` return patchsetNumFromPJ(pj) <= patchsetNum ` will not prevent spamming.

However, https://github.com/kubernetes/test-infra/blob/4f436d0b9431be68934a37799ed7069099391de2/prow/crier/reporters/gerrit/reporter.go#L123 does gate on aborted batches, when all jobs are aborted then no report will be generated

PR #22260, opened 2021-05-19
GitHub: Allow setting up throttling per org

alvaroaleman:
When using github apps auth, the token budget is scoped to a github organization and not global like with personal access tokens. Thus the throttling also needs to be scoped to an org.

This PR adds that by making the throttler internally use an org map and default to a global key if no throttler for the given org was found. 

/cc @petr-muller


petr-muller (approved):
Neat!

PR #14423, opened 2019-09-20
remove k-s/k

BenTheElder:
cc @liggitt

PR #22270, opened 2021-05-20
Add a presubmit job for kubernetes-security/cvelist-public

dims:
related to:
- https://github.com/kubernetes-security/cvelist-public/pull/4
- https://github.com/kubernetes/security/issues/127

### DIFF
--- a/config/jobs/kubernetes/sig-security/cvelist-public.yaml
+++ b/config/jobs/kubernetes/sig-security/cvelist-public.yaml
@@ -1,0 +1,21 @@
+presubmits:
+  kubernetes-security/cvelist-public:
+  - name: validate-cve-files
+    decorate: true
+    always_run: false
+    optional: true
+    run_if_changed: 'CVE.*json$'
+    spec:
+      containers:
+      - image: golang
+        command:
+        - /bin/bash
+        args:
+        - -c
+        - |
+          set -euo pipefail; \
+          git diff --name-only --diff-filter=d $PULL_BASE_SHA...$PULL_PULL_SHA | grep "\/CVE.*json$" | xargs validate-k8s.sh
+    annotations:
+      testgrid-create-test-group: "true"
+      testgrid-dashboards: sig-security-cvelist-public
+      description: Validate the CVE json
--- a/config/jobs/kubernetes/sig-security/OWNERS
+++ b/config/jobs/kubernetes/sig-security/OWNERS
@@ -1,0 +1,9 @@
+# See the OWNERS docs at https://go.k8s.io/owners
+
+# This is the owner of the security related code.  The test data itself is owned by sig-security.
+reviewers:
+  - tallclair
+approvers:
+  - tallclair
+labels:
+  - sig/security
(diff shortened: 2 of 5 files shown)


config/jobs/kubernetes/sig-security/cvelist-public.yaml
```diff
@@ -0,0 +1,26 @@
...
+      containers:
+      - image: golang
+        command:
+        - /bin/bash
+        args:
+        - -c
+        - |
+          set -euo; \
+          files=( $(git diff --name-only --diff-filter=d $PULL_BASE_SHA...$PULL_PULL_SHA | grep "\/CVE.*json$") ); \
+          for f in ${files[@]}:; \
```
tallclair, line 18: If you want to just pass in all the files, I'll update the validate script to take multiple files.

dims: @tallclair let’s leave it as is now, (one file per invocation of the shell script). once we have the skeleton working, we can tweak

tallclair: up to you. I already merged the updated script. The advantage being that it will check all files before failing.

config/jobs/kubernetes/sig-security/cvelist-public.yaml
```diff
@@ -0,0 +1,26 @@
...
+    run_if_changed: 'CVE.*json$'
+    spec:
+      containers:
+      - image: golang
+        command:
+        - /bin/bash
+        args:
+        - -c
+        - |
+          set -euo; \
```
tallclair, line 16: doesn't `set -o` need an argument? I think this should be `set -euo pipefail`, but I could be wrong...

dims: fixed.

config/jobs/kubernetes/sig-security/cvelist-public.yaml
dims: fixed now.

config/jobs/kubernetes/sig-security/cvelist-public.yaml
```diff
@@ -0,0 +1,21 @@
+presubmits:
+  kubernetes/kubernetes:
```
nikhita, line 2: should be ` kubernetes-security/cvelist-public `?

dims: thanks @nikhita fixed.

dims:
Here's confirmation that we do NOT have a cluster named `security` though we do have one called `k8s-infra-prow-build-trusted`
- https://cs.k8s.io/?q=cluster%3A.*trusted&i=fosho&files=&excludeFiles=&repos=kubernetes/test-infra
- https://cs.k8s.io/?q=cluster%3A.*security&i=fosho&files=&excludeFiles=&repos=kubernetes/test-infra

nikhita:
The security cluster test looks like a stale reference from https://github.com/kubernetes/test-infra/pull/14423#issuecomment-533710023

/lgtm

PR #22275, opened 2021-05-20
[cvelist] install jq

dims:


tallclair:
/lgtm
Alternatively, it looks like we publish a "bootstrap" image with jq included.

PR #22279, opened 2021-05-20
kinder: test jobs to use a different SIG node alerts ML

neolit123:
Switch the kubelet skew jobs to report to:
sig-node-test-failures+testgrid@googlegroups.com

/cc @ehashman


ehashman (commented):
/lgtm

PR #22284, opened 2021-05-20
[sig-node] Add presubmit for kubelet-serial

ehashman:
Mostly copied from https://github.com/kubernetes/test-infra/blob/29f8f503b402de8c4642d2cc74ddba79551d739e/config/jobs/kubernetes/sig-node/node-kubelet.yaml#L205-L233

/cc @dims @BenTheElder 
/sig node

### DIFF
--- a/config/jobs/kubernetes/sig-node/sig-node-presubmit.yaml
+++ b/config/jobs/kubernetes/sig-node/sig-node-presubmit.yaml
@@ -204,6 +204,39 @@ presubmits:
             memory: "6Gi"
     annotations:
       testgrid-create-test-group: 'true'
+  - name: pull-kubernetes-node-kubelet-serial
+    always_run: false
+    optional: true
+    skip_report: false
+    skip_branches:
+    - release-\d+\.\d+  # per-release image
+    annotations:
+      testgrid-dashboards: sig-node-kubelet
+      testgrid-tab-name: pr-node-kubelet-serial
+    labels:
+      preset-service-account: "true"
+      preset-k8s-ssh: "true"
+    spec:
+      containers:
+      - image: gcr.io/k8s-testimages/kubekins-e2e:v20210512-b8d1b30-master
+        args:
+        - --repo=k8s.io/kubernetes=master
+        - --timeout=440
+        - --root=/go/src
+        - --scenario=kubernetes_e2e
+        - --
+        - --deployment=node
+        - --gcp-project-type=node-e2e-project
+        - --gcp-zone=us-west1-b
+        - --node-args=--image-config-file=/workspace/test-infra/jobs/e2e_node/image-config-serial.yaml
+        - --node-test-args=--feature-gates=DynamicKubeletConfig=true,LocalStorageCapacityIsolation=true --kubelet-flags="--cgroups-per-qos=true --cgroup-root=/"
+        - --node-tests=true
+        - --provider=gce
+        - --test_args=--nodes=1 --focus="\[Serial\]" --skip="\[Flaky\]|\[Benchmark\]|\[NodeSpecialFeature:.+\]|\[NodeAlphaFeature:.+\]"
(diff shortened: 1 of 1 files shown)


dims:
/hold

let's please explore a containerd version of this job!

ehashman:
> /hold
> 
> let's please explore a containerd version of this job!

@dims while I agree we really need a non-docker version of this job, right now this is just a presubmit duplicate of the job that's currently running as a periodic. I don't think that we should block being able to trigger the existing job on a PR.

dims:
@ehashman that's fair, i am just pushing on us to come up with a containerd based job.

ehashman:
We definitely need to do it :pray: But I would like to see the hold cancelled here so we can at least test this on PRs rather than merge first test as periodic :sweat_smile:

harche:
I will add similar job with crio + systemd configuration.

harche:
@ehashman @dims https://github.com/kubernetes/test-infra/pull/22290

config/jobs/kubernetes/sig-node/sig-node-presubmit.yaml
```diff
@@ -204,6 +204,39 @@ presubmits:
...
+        - --
+        - --deployment=node
+        - --gcp-project-type=node-e2e-project
+        - --gcp-zone=us-west1-b
+        - --node-args=--image-config-file=/workspace/test-infra/jobs/e2e_node/image-config-serial.yaml
+        - --node-test-args=--feature-gates=DynamicKubeletConfig=true,LocalStorageCapacityIsolation=true --kubelet-flags="--cgroups-per-qos=true --cgroup-root=/"
+        - --node-tests=true
+        - --provider=gce
+        - --test_args=--nodes=1 --focus="\[Serial\]" --skip="\[Flaky\]|\[Benchmark\]|\[NodeSpecialFeature:.+\]|\[NodeAlphaFeature:.+\]"
+        - --timeout=300m
```
odinuge, line 236: I am pretty sure this timeout is the reason the period job is failing/we get no output. Bump this to `420` or `360` in order to see if it makes any difference? We can always go back and bump the value later in case it is needed.

odinuge (approved):
Except for the timeout stuff;

/lgtm

ref. @dims' comment, I think it is best to keep this "similar" to the old `node-kubelet-serial` for now, and then add/change it to use containerd later. @harche is doing a cri-o version with systemd, which is way more important!

config/jobs/kubernetes/sig-node/sig-node-presubmit.yaml
ehashman: Bumped it to 420m.

odinuge:
`Job execution failed: Pod got deleted unexpectedly`
/retest

BenTheElder:
>     jobs_test.go:894: Invalid Scenario Args : jobs pull-kubernetes-node-kubelet-serial - kubetest timeout(7h0m0s), >bootstrap timeout(5h20m0s): bootstrap timeout need to be 20min more than kubetest timeout!
 
Also ideally new jobs should not be using https://github.com/kubernetes/test-infra/blob/master/jenkins/bootstrap.py, instead using `decorate: true` and the podutils, but that's another matter. Something to put on the backlog.

SergeyKanzhelev:
@ehashman can you please fix the broken build. I agree that we can unhold and address the containerd comment later

ehashman:
Should be fixed now :crossed_fingers:

ehashman:
Should be GTG now! @SergeyKanzhelev @BenTheElder

SergeyKanzhelev:
/hold cancel

@dims are you ok with the plan to enable it first and later we can either add another job with containerd or change this one?

dims:
@SergeyKanzhelev sounds good!

BenTheElder (approved):
LGTM

PR #22288, opened 2021-05-21
Add PR creation to the gerrit onboarder

mpherman2:
We are unable to use the same PR Creator that we created for the autobumper, because that pr creator assumes that the PR will be created by prowbot during a prowjob. This is much simpler as we are just running the tool locally for now, so we added simple "commit and push" calls to git.

/assign @chaodaiG 
/assign @cjwagner

### DIFF
--- a/experiment/gerrit-onboarder/onboarder.go
+++ b/experiment/gerrit-onboarder/onboarder.go
@@ -17,12 +17,16 @@ limitations under the License.
 package main
 
 import (
+	"bytes"
 	"errors"
 	"flag"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"math/rand"
 	"os"
+	"os/exec"
+	"path"
 	"regexp"
 	"strings"
 	"time"
@@ -63,6 +67,8 @@ type options struct {
 	repo      string
 	uuID      string
 	groupName string
+	dryRun    bool
+	local     bool
 }
 
 func parseAndValidateOptions() (*options, error) {
@@ -71,10 +77,12 @@ func parseAndValidateOptions() (*options, error) {
 	flag.StringVar(&o.repo, "repo", "", "The gerrit Repo.")
 	flag.StringVar(&o.uuID, "uuid", "", "The UUID to be added to the file.")
 	flag.StringVar(&o.groupName, "group", "", "The corresponding group name for the UUID.")
+	flag.BoolVar(&o.dryRun, "dry_run", false, "If dry_run is true, PR will not be created")
+	flag.BoolVar(&o.local, "local", false, "If local is true, changes will be made to local repo instead of new temp dir.")
 	flag.Parse()
 
 	if o.host == "" || o.repo == "" || o.uuID == "" || o.groupName == "" {
-		return &o, errors.New("all flags are required")
+		return &o, errors.New("host, repo, uuid, and group are all required fields")
 	}
 
 	return &o, nil
@@ -155,23 +163,23 @@ func ensureUUID(groupsFile, uuid, group string) (string, error) {
 	return mapToGroups(groupsMap, orderedKeys), nil
 }
 
-func updateGroups(uuid, group string) (bool, error) {
-	data, err := ioutil.ReadFile(groupsFile)
+func updateGroups(workDir, uuid, group string) error {
+	data, err := ioutil.ReadFile(path.Join(workDir, groupsFile))
 	if err != nil {
-		return false, fmt.Errorf("faield to read groups file: %w", err)
+		return fmt.Errorf("failed to read groups file: %w", err)
 	}
 
 	newData, err := ensureUUID(string(data), uuid, group)
 	if err != nil {
-		return false, fmt.Errorf("faield to ensure group exists: %w", err)
+		return fmt.Errorf("failed to ensure group exists: %w", err)
 	}
 
-	err = ioutil.WriteFile(groupsFile, []byte(newData), 0755)
+	err = ioutil.WriteFile(path.Join(workDir, groupsFile), []byte(newData), 0755)
 	if err != nil {
-		return false, fmt.Errorf("faield to write groups file: %w", err)
+		return fmt.Errorf("failed to write groups file: %w", err)
 	}
 
-	return newData != string(data), nil
+	return nil
 }
 
 func configToMap(configFile string) (map[string][]string, []string) {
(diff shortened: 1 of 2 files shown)


experiment/gerrit-onboarder/onboarder.go
```diff
@@ -408,6 +454,22 @@ func main() {
...
+	if diff == "" {
+		logrus.Info("No changes made. Returning without creating CR")
+		return
+	}
+
+	if err = createCR(); err != nil {
+		logrus.Fatal(err)
+	}
+
+	os.Chdir(curDir)
```
chaodaiG, line 470: Debug?

experiment/gerrit-onboarder/onboarder.go
```diff
@@ -373,12 +374,57 @@ func fetchMetaConfig(host, repo, branch string) error {
...
+		logrus.Fatal(err)
+	}
+
+	workDir, err := ioutil.TempDir("", "gerrit_onboarding")
+	if err != nil {
+		logrus.Fatal(err)
+	}
+	defer os.RemoveAll(workDir)
+
+	err = os.Chdir(workDir)
```
chaodaiG, line 414: This looks bash to me, can we not using `chDir`? If it's for shelling out to git, `exec` package does support `Dir` https://golang.org/pkg/os/exec/#Cmd

experiment/gerrit-onboarder/onboarder.go
```diff
@@ -408,6 +454,22 @@ func main() {
...
 
-	//Create PR
+	diff, err := getDiff()
+	if err != nil {
+		logrus.Fatal(err)
+	}
+	if diff == "" {
+		logrus.Info("No changes made. Returning without creating CR")
+		return
+	}
```
chaodaiG, line 464: This can be part of `createCR`

mpherman2: Done

experiment/gerrit-onboarder/onboarder.go
mpherman2: Created a execInDir function and replaced all uses of Call with this to avoid having to chDir.

experiment/gerrit-onboarder/onboarder.go
```diff
@@ -341,73 +345,134 @@ func ensureProjectConfig(config, host, cur_branch, groupName string) (string, er
...
-	groupsChanged, err := updateGroups(o.uuID, o.groupName)
+	groupsChanged, err := updateGroups(workDir, o.uuID, o.groupName)
 	if err != nil {
 		logrus.Fatal(err)
 	}
 
 	if groupsChanged {
 		logrus.Info("groups file did not change. Skipping stash step")
-		if err := bumper.Call(os.Stdout, os.Stderr, "git", "stash"); err != nil {
+		if err := execInDir(os.Stdout, os.Stderr, workDir, "git", "stash"); err != nil {
```
chaodaiG, line 459: This doesn't match the log. Also, I assume `git stash` is because of checking out different repos in the same dir, should we consider checking out in different dir to save one shelling out to git?

mpherman2: Yea. The stash _was_ necessary to traverse the inheritance tree when checking the project config, but I realized that if we do that first we never need to stash changes. That is much cleaner.

mpherman2 (commented):
Responded to your comment, and also added two new flags. One to run the code in a local repo instead of creating a temp dir, and one to run the code without creating a PR. This is useful for testing, but also will be useful if people want to run the tool and then make some edits before submitting the PR

experiment/gerrit-onboarder/onboarder.go
```diff
@@ -341,73 +349,131 @@ func ensureProjectConfig(config, host, cur_branch, groupName string) (string, er
...
+	}
+	if diff == "" {
+		logrus.Info("No changes made. Returning without creating CR")
+		return nil
+	}
+	commitMessage := fmt.Sprintf("Grant the Prow cluster read and label permissions\n\nChange-Id: I%s", bumper.GitHash(fmt.Sprintf("%d", rand.Int())))
+	if err := execInDir(os.Stdout, os.Stderr, workDir, "git", "commit", "-a", "-v", "-m", commitMessage); err != nil {
+		return fmt.Errorf("unable to commit: %w", err)
+	}
+	if err := execInDir(os.Stdout, os.Stderr, workDir, "git", "push", "origin", "HEAD:refs/for/refs/meta/config"); err != nil {
```
chaodaiG, line 416: Would recommend moving dryrun check here for getting most out of dryrun

experiment/gerrit-onboarder/onboarder.go
```diff
@@ -341,73 +349,131 @@ func ensureProjectConfig(config, host, cur_branch, groupName string) (string, er
...
-	if err := bumper.Call(os.Stdout, os.Stderr, "git", "switch", "-c", branch); err != nil {
+	if err := execInDir(os.Stdout, os.Stderr, workDir, "git", "switch", "-c", branch); err != nil {
 		return fmt.Errorf("failed to switch to new branch, %w", err)
 	}
 
 	return nil
 }
 
+func execInDir(stdout, stderr io.Writer, dir string, cmd string, args ...string) error {
+	(&logrus.Logger{
```
chaodaiG, line 385: If feels a bit weird to initialize this every time. Also, could you explain why setting all these logging configs?

experiment/gerrit-onboarder/onboarder.go
```diff
@@ -341,73 +349,131 @@ func ensureProjectConfig(config, host, cur_branch, groupName string) (string, er
...
 	branchName := fmt.Sprintf("gerritOnboarding_%d", rand.Int())
 
-	if err = fetchMetaConfig(o.host, o.repo, branchName); err != nil {
+	if err = fetchMetaConfig(o.host, o.repo, branchName, workDir); err != nil {
 		logrus.Fatal(err)
 	}
 
-	groupsChanged, err := updateGroups(o.uuID, o.groupName)
-	if err != nil {
+	if err = updatePojectConfig(workDir, o.host, branchName, o.groupName); err != nil {
```
chaodaiG, line 464: branch is already checked out in `fetchMetaConfig`, so I'm not super convinced that it's required here

chaodaiG: In another word, probably should just make branchname a local var in `fetchMetaConfig`

experiment/gerrit-onboarder/onboarder.go
mpherman2: Done

experiment/gerrit-onboarder/onboarder.go
mpherman2: Using https://github.com/kubernetes/test-infra/blob/d9d68773cbb3f396035b49b545b8bcefa367155d/prow/cmd/generic-autobumper/bumper/bumper.go#L398 as my reference when creating this function. This feels cleaner than passing the logger around to every function. Should I create the logger as a global var?

experiment/gerrit-onboarder/onboarder.go
mpherman2: I believe it is required we pass the branch name to updateProjectConfig for use in VerifyInTree. We keep track of the current branch so that we can create new branches when moving down the inheritance tree and be able to return to the original branch.

experiment/gerrit-onboarder/onboarder.go
chaodaiG: Ack, nvm then

experiment/gerrit-onboarder/onboarder.go
chaodaiG: I thought we are cloning in different dirs now, so might not need to switch branch again?

experiment/gerrit-onboarder/onboarder.go
```diff
@@ -341,73 +349,131 @@ func ensureProjectConfig(config, host, cur_branch, groupName string) (string, er
...
-	if groupsChanged {
-		logrus.Info("groups file did not change. Skipping stash step")
-		if err := bumper.Call(os.Stdout, os.Stderr, "git", "stash"); err != nil {
-			logrus.Fatal(err)
-		}
-	}
-
-	if err = updatePojectConfig(o.host, branchName, o.groupName); err != nil {
+	err = updateGroups(workDir, o.uuID, o.groupName)
+	if err != nil {
```
chaodaiG, line 471: combine these two lines as well

mpherman2: Done

chaodaiG (approved):

PR #22292, opened 2021-05-21
Use kubernetes/test-infra log_dump.sh to dump logs.

YangLu1031:
The log_dump.sh in k/k repo is migrated to k/test-infra: https://github.com/kubernetes/kubernetes/blob/master/cluster/log-dump/README.md. 
So according to the migration steps, need to add `USE_TEST_INFRA_LOG_DUMPING=true` to use it.

On the other hand, I use this PR for verification as well, as running kubetest locally seems not easy. And we can run `/test pull-kubernetes-e2e-windows-gce` to trigger a test to verify whether expected logs are collected in artifacts.

### DIFF
--- a/config/jobs/kubernetes/sig-windows/windows-gce.yaml
+++ b/config/jobs/kubernetes/sig-windows/windows-gce.yaml
@@ -8,6 +8,8 @@ presets:
     value: "windows"
   - name: KUBELET_TEST_ARGS
     value: "--feature-gates=KubeletPodResources=false"
+  - name: USE_TEST_INFRA_LOG_DUMPING
+    value: "true"
 - labels:
     preset-e2e-gce-windows: "true"
   env:


YangLu1031:
Ok, seems `/test pull-kubernetes-e2e-windows-gce` doesn't work under this repo.

pjh:
This relates to https://github.com/kubernetes/kubernetes/issues/96915.

pjh:
This looks like https://github.com/kubernetes/test-infra/blob/f1b528b015de88fccf7aee59610ac5ec5aee342a/config/jobs/kubernetes/sig-scalability/sig-scalability-presets.yaml#L88 so that's good enough for me.

/lgtm

YangLu1031:
Thanks @pjh, /unhold

PR #22300, opened 2021-05-21
added timeout for depstat jobs

RinkiyaKeDad:

Added a timeout of 5 mins for both the depstat jobs. For context, running on k/k currently takes ~1 min. Timeout will help detect if we face the long time to produce results problem for k/k also at some point.

/assign @dims


oomichi:
This timeout seems the lowest one in all jobs which specify timeout.
According to `For context, running on k/k currently takes ~1 min.` it is fine to add this at this time.
Then we can increase it if necessary.

/ok-to-test
/lgtm

PR #17771, opened 2020-05-31
prow: deck: S3 refactoring cleanup

sbueringer:
This PR cleans up some struct/variable names we didn't want to refactor during the previous. S3 deck PR. It also removes `gcs-no-auth` flag (not sure if it's too early for that)

xref: https://github.com/kubernetes/test-infra/issues/11260

PR #22302, opened 2021-05-21
Remove --gcs-no-auth flag

YuviGold:
On https://github.com/kubernetes/test-infra/pull/17771 this flag was
announced as going to be deprecated.

At the moment this flag does nothing except for a log:
'--gcs-no-auth' is deprecated and is not used anymore.
We always fall back to an anonymous client now, if all other options fail.

/cc sbueringer


sbueringer:
@YuviGold Thank you very much!

/lgtm

/cc @alvaroaleman

alvaroaleman (approved):

PR #22305, opened 2021-05-21
Set runlocal script absolute path

YuviGold:
The `runlocal` script is relying on being executed from the root
directory. Hence an absolute path is more appropriate.

Additionally, the URL to the deployed server was added to the readme.


petr-muller (approved):
/ok-to-test

alvaroaleman (approved):

Issue #22285, opened 2021-05-20
gcupload should upload finished.json last


fejta:
https://kubernetes.slack.com/archives/CDECRSC5U/p1621546950011500

Uploading finished.json signifies the job is done and should happen as close as possible to the job actually finishing.

Currently gcsupload just gathers all the artifacts together and uploads them concurrently. If I have many GBs of artifacts, this can result in us uploading finished.json a long time before we actually finish (because we're still uploading all the other artifacts).

I want to change up this to only upload finished.json at the end.

This will help systems like testgrid better understand when the logs have been fully uploaded to GCS.

fejta:
/help
/good-first-issue

I'm planning to work on this soon unless someone else fancies themselves to work on it.

One idea would be to change assembleTargets to return `[]map[string]gcs.UploadFunc` rather than `map[string]gcs.UploadFunc` and do each map of items sequentially (putting just the finished.json in the second set)

https://github.com/kubernetes/test-infra/blob/ac9ffd6caea2caf244b277a025635640f6095544/prow/gcsupload/run.go#L48

Another idea would be to just look specifically for finished.json in uploadTargets and pull it out, and then call this sequence twice: https://github.com/kubernetes/test-infra/blob/ac9ffd6caea2caf244b277a025635640f6095544/prow/gcsupload/run.go#L53-L71

A third idea would be to make "final target" an option in https://github.com/kubernetes/test-infra/blob/master/prow/gcsupload/options.go#L118 and then do one of the above things based on that value (and default it to finished.json)

PR #22310, opened 2021-05-21
Upload extra targets after the primary ones

fejta:
This will allow us to upload the finished.json after the other targets,
which will minimize the amount of time between when finished.json is
uploaded and the prowjob completes.

fixes https://github.com/kubernetes/test-infra/issues/22285


fejta:
Moved the bumpmonitoring fix to https://github.com/kubernetes/test-infra/pull/22311

chaodaiG (approved):
/hold

PR #22317, opened 2021-05-22
Show pull info on job-history

YuviGold:
Add a new Pull pointer field for buildData struct.
The pull is being fetched from the prowjob.json file from gcs.
    
Each build on job history page would present pull data (if there is):
<build_id> (#<pull> by <author)

Screenshot from 2021-05-22 18-42-43

### DIFF
--- a/prow/cmd/deck/job_history.go
+++ b/prow/cmd/deck/job_history.go
@@ -366,9 +366,9 @@ func getBuildData(ctx context.Context, bucket storageBucket, dir string) (buildD
 	}
 
 	pj := prowapi.ProwJob{}
-	err = readJSON(ctx, bucket, path.Join(dir, "prowjob.json"), &pj)
+	err = readJSON(ctx, bucket, path.Join(dir, prowv1.ProwJobFile), &pj)
 	if err != nil {
-		logrus.Debugf("failed to read prowjob.json: %v", err)
+		logrus.WithError(err).Debugf("failed to read %s", prowv1.ProwJobFile)
 	} else {
 		if pj.Spec.Refs != nil && len(pj.Spec.Refs.Pulls) > 0 {
 			b.Pull = &pj.Spec.Refs.Pulls[0]
--- a/prow/cmd/deck/job_history.go
+++ b/prow/cmd/deck/job_history.go
@@ -33,6 +33,7 @@ import (
 
 	"github.com/sirupsen/logrus"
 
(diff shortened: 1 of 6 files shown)


prow/cmd/deck/job_history.go
```diff
@@ -385,6 +391,19 @@ func getBuildData(ctx context.Context, bucket storageBucket, dir string) (buildD
 	return b, nil
 }
 
+func getPullRequest(jobAgent *jobs.JobAgent, jobName, buildID string) (*prowv1.Pull, error) {
+	j, err := jobAgent.GetProwJob(jobName, buildID)
```
alvaroaleman, line 395: For the bigger Prow instances like k8s and openshift, this will only work for about 24 hours, after which the prowjobs are deleted from the kubernetes cluster. The purpose of the view you are changing is to view historical data from GCS, which means that 99% of the jobs you are seeing there won't be in the Kubernetes cluster anymore and thus this will not work.

YuviGold: Thanks Alvaro, I wasn't aware of this limitation.
I changed the implementation to read `prowjob.json` from GCS instead of using JobAgent cache.

chaodaiG:
Awesome!
/ok-to-test

prow/cmd/deck/job_history.go
```diff
@@ -363,6 +365,16 @@ func getBuildData(ctx context.Context, bucket storageBucket, dir string) (buildD
 		logrus.Debugf("failed to read finished.json (job might be unfinished): %v", err)
 	}
 
+	pj := prowapi.ProwJob{}
+	err = readJSON(ctx, bucket, path.Join(dir, "prowjob.json"), &pj)
```
petr-muller, line 369: I think `prowjob.json` should be turned into a constant that both this and the producer uses.

prow/cmd/deck/job_history.go
```diff
@@ -363,6 +365,16 @@ func getBuildData(ctx context.Context, bucket storageBucket, dir string) (buildD
...
+	if err != nil {
+		logrus.Debugf("failed to read prowjob.json: %v", err)
```
alvaroaleman, line 371:
```suggestion
		logrus.WithError(err).Debug("Failed to read prowjob.json")
```

alvaroaleman (approved):
This is pretty great, thanks!

PR #22337, opened 2021-05-25
kubernetes-csi: test capacity tracking on Kubernetes 1.21

pohly:
Kubernetes 1.21 is better than 1.20 because storage capacity tracking
is enabled for normal tests, not just for alpha.

### DIFF
--- a/config/jobs/kubernetes-csi/csi-driver-host-path/csi-driver-host-path-manual-job-config.yaml
+++ b/config/jobs/kubernetes-csi/csi-driver-host-path/csi-driver-host-path-manual-job-config.yaml
@@ -2,7 +2,7 @@
 
 presubmits:
   kubernetes-csi/csi-driver-host-path:
-  - name: pull-kubernetes-csi-csi-driver-host-distributed-on-kubernetes-1-19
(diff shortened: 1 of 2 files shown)


config/jobs/kubernetes-csi/csi-driver-host-path/csi-driver-host-path-manual-job-config.yaml
```diff
@@ -47,17 +47,21 @@ presubmits:
...
+            # this is mostly for building kubernetes
+            memory: "9000Mi"
+            # during the tests more like 3-20m is used
             cpu: 2000m
 periodics:
 - interval: 6h
-  # 1.20 is used because we currently have pre-built images for it.
-  # The Kubernetes version does not matter much here, we just care
-  # about exercising distributed provisioning on some Kubernetes version.
+  # 1.21 is used because we the distributed deployment then also
```
xing-yang, line 57: we the distributed deployment then also -> the distributed deployment also

pohly: Fixed and pushed.

PR #22339, opened 2021-05-25
remove deprecated linter check for vSphere CSI driver

divyenpatel:
make link target is being removed from the vSphere CSI driver with this PR - https://github.com/kubernetes-sigs/vsphere-csi-driver/pull/909/ 

https://github.com/golang/lint is deprecated.


divyenpatel:
> This is all covered by golanci-lint now, right?

static-check and go vet is suggested as per https://github.com/golang/go/issues/38968 
For vSpehre CSI driver we also have `golanci-lint` `static-check` and `go vet`, so we are covered.

PR #22358, opened 2021-05-27
Revert "Merge pull request #22341 from YuviGold/prow-job-title"

stevekuznetsov:
This reverts commit 6d5470c6523e609b584761121ebaadff92f42452, reversing
changes made to a121ad51251e955da8d48b46c15e322b16f4d059.

This caused unintended consequences downstream.


alvaroaleman:
The original change caused issues for us because we embed the clonerefs_options into an openshift build as env var which seems to mess up escaping when a json string field contains double quotes. We'll work around this downstream instead and get the env var handling fixed.
/close

PR #22362, opened 2021-05-27
prow: build images for arm64

LorbusChris:
With this commit, a minmal prow payload for arm64 can be built.
The payload only contains the following four images which are
the minimal requirements for scheduling Prow builds on an arm64 build
farm:
- clonerefs
- entrypoint
- initupload
- sidecar

The image will be tagged as before but with an additional `-arm64` suffix.

Blocked on: https://github.com/kubernetes/test-infra/pull/22450

/cc @stevekuznetsov


prow/Makefile
```diff
@@ -74,7 +74,7 @@ bazel-release-push:
 	@echo See https://bazel.build/ for install options.
 	@echo Be sure to setup authentication: https://github.com/bazelbuild/rules_docker#authentication
 	@echo Also run gcloud auth application-default login
-	bazel run //prow:release-push --platforms=@io_bazel_rules_go//go/toolchain:linux_amd64
+	bazel run //prow:release-push --platforms="@io_bazel_rules_go//go/toolchain:linux_amd64 @io_bazel_rules_go//go/toolchain:linux_arm64"
```
fejta, line 77: Isn't this going to push both linux and arm to the same tag, overriding one of them?

fejta:
Right now rules_docker does not have the ability to produce multi-platform images, so we'll either need to pull in the ability to do this, use different tags for other architectures, or something else.

https://github.com/bazelbuild/rules_docker/issues/1688
https://github.com/bazelbuild/rules_docker/issues/1599

LorbusChris:
the second commit currently contains a reference to a one-off rebuild of the git image. It should be replaced by a properly automated build from #22450
/hold

prow/def.bzl
```diff
@@ -48,6 +49,32 @@ def prow_image(
...
 
+# prow_image_arm64 is a macro for creating arm64 :app and :image targets
+def prow_image_arm64(
+        component,
+        name,  # use "image-arm64"
+        base = None,
+        stamp = True,  # stamp by default, but allow overrides
+        app_name = "app",
+        **kwargs):
+    go_image(
```
fejta, line 60: Is there some important reason we need both `prow_image` and `prow_image_arm64` macros?

It would be nice to just add these two `go_image` and `container_image` calls to the `prow_image` macro, potentially gated by an `declare_arm = False` macro which we can set to true for the handful of places you want to use it

prow/BUILD.bazel
```diff
@@ -59,6 +59,19 @@ prow_push(
     ),
 )
 
+prow_push(
+    name = "release-push-arm64",
+    bundle_name = "release-arm64",
+    images = tags_arm64(
```
fejta, line 65: How about just adding this to `release-push` images argument?

prow/def.bzl
LorbusChris: My reasoning was that having arm images might not be required for everyone, and this way one can make them separately.
I'm not married to doing it this way, but I'd like to hear @stevekuznetsov's thoughts on this, too.

LorbusChris: Gating via some bool on the macro would achieve the same, though. I'll go ahead and make the change.

LorbusChris:
@fejta I've updated the macro per your review. Please take another look

fejta (approved):
This looks really great! Thanks for all the work here.

LorbusChris:
great, thanks for the review!
/hold cancel

Issue #22259, opened 2021-05-19
prow: needs-rebase spam


BenTheElder:
**What happened**:

The PR needs rebase comment is being posted ~once an hour on https://github.com/kubernetes/kubernetes/pull/101685

**What you expected to happen**:

This comment should be posted once and not-reposted while the label is still there.

**How to reproduce it (as minimally and precisely as possible)**:

no idea! it's ongoing on prow.k8s.io though

**Please provide links to example occurrences, if any**:

https://github.com/kubernetes/kubernetes/pull/101685#issuecomment-844321333

**Anything else we need to know?**:

initially reported by @msau42 https://kubernetes.slack.com/archives/C09QZ4DQB/p1621448488023900

/area prow

cjwagner:
The logs are full of
```yaml
{
  client: "github"   
  component: "needs-rebase"   
  file: "prow/github/client.go:404"   
  func: "k8s.io/test-infra/prow/github.(*throttler).Wait.func1"   
  level: "warning"   
  msg: "Throttled clientside for more than 15 minutes"   
  throttle-duration: "1h11m3.41797216s"   
  throttled: true   
}
```
So this is likely poor behavior due to throttling. e.g. We keep trying to post comments because we haven't posted one yet, but they're all queued up due to throttling. We use the v4 Graphql API to periodically look for PRs that need to be updated and the v4 API has a separate rate limit from the v3 REST API used to post the comments.

I vote we get rid of the deployment for now since it is not very important and if someone cares about this functionality they can look into the throttling issues before redeploying.

BenTheElder:
> We keep trying to post comments because we haven't posted one yet, but they're all queued up due to throttling. 

But we _have_ posted one. So we are continuing to try without checking if we need to?

Even if it reduces how much we can post, I think every "post a comment that may be repeated" should be checking if the comment exists first, so you can't get this runaway comment behavior, even if it costs an extra API call and halves the overall throughput. I think people would much rather tolerate some extra latency on these than the excessive commenting.

alvaroaleman:
> Even if it reduces how much we can post, I think every "post a comment that may be repeated" should be checking if the comment exists first,

We do indirectly through the label:  https://github.com/kubernetes/test-infra/blob/97195cc60c7966428a0132e5b5d820db9772ada7/prow/external-plugins/needs-rebase/plugin/plugin.go#L202-L206

[...]

PR #22369, opened 2021-05-28
Needs rebase take action: Time out after five minutes

alvaroaleman:
Fixes https://github.com/kubernetes/test-infra/issues/22259

This change adds a five minute timeout to the `takeAction` method of the needs-rebase plugin. This is needed because needs-rebase is a heavy token consumer and thus often gets throttled, which can result in manipulating labels or comments based on data that is hours or more old. It also means that in case of throttling, it will internally end up with a pile of events that wait for throttler quota and that once they get it (often during off-hours) will spam prs with multiple comments.

The PR is divided into two commits, the first one adds a bunch of WithContext methods to the GitHub client and extends the throttler to respect it, the second makes needs-rebase use those.

/cc @cjwagner @chaodaiG

### DIFF
--- a/prow/github/client_test.go
+++ b/prow/github/client_test.go
@@ -2968,7 +2968,7 @@ func TestAllMethodsThatDoRequestSetOrgHeader(t *testing.T) {
 
 	clientMethods := getCallForAllClientMethodsThroughReflection(
 		ghClient,
-		func(methodName string) bool { return toSkip.Has(methodName) },
+		func(method reflect.Method) bool { return toSkip.Has(method.Name) },
 		func(typeName string) interface{} {
 			if typeName == "string" {
 				return "org"
@@ -3002,7 +3002,7 @@ func TestAllMethodsThatDoRequestSetOrgHeader(t *testing.T) {
 
 func getCallForAllClientMethodsThroughReflection(
 	c Client,
-	skip func(methodName string) bool,
+	skip func(reflect.Method) bool,
 	typeOverrides ...func(typeName string) (override interface{}),
 ) (getCalls []func() (methodName string, call func() error)) {
 
@@ -3011,7 +3011,7 @@ func getCallForAllClientMethodsThroughReflection(
 
 	for i := 0; i < clientType.NumMethod(); i++ {
 		i := i
-		if skip(clientType.Method(i).Name) {
+		if skip(clientType.Method(i)) {
 			continue
 		}
 		var args []reflect.Value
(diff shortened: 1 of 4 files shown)


prow/github/client_test.go
```diff
@@ -2980,45 +2983,83 @@ func TestAllMethodsThatDoRequestSetOrgHeader(t *testing.T) {
...
-				// We can not deal with interface types genererically, as there
-				// is no automatic way to figure out the concrete values they
-				// can or should be set to.
-				if arg.Type().String() == "context.Context" {
-					arg.Set(reflect.ValueOf(context.Background()))
-				}
-				if arg.Type().String() == "interface {}" {
-					arg.Set(reflect.ValueOf(map[string]interface{}{}))
-				}
+func getCallForAllClientMethodsThroughReflection(
```
alvaroaleman, line 2995: I factored this out from `TestAllMethodsThatDoRequestSetOrgHeader` so that I can test the context handling through reflect by checking on all methods whose name ends with `WithContext` rather than having to write a test by hand for each of them.

alvaroaleman:
> Pod got deleted unexpectedly 
/retest

prow/github/client_test.go
```diff
@@ -3237,3 +3278,42 @@ func TestCreatePullRequestReviewComment(t *testing.T) {
...
+		t.Fatalf("failed to use up the throttlers token: %v", err)
+	}
+
+	// Use a very low timeout so we don't have to wait
+	ctx, cancel := context.WithTimeout(context.Background(), time.Millisecond)
+	defer cancel()
+
+	clientMethods := getCallForAllClientMethodsThroughReflection(c,
+		// Skip all methods whose name doesn't end with 'WithContext'
+		func(methodName string) bool { return !strings.HasSuffix(methodName, "WithContext") },
```
stevekuznetsov, line 3301: While this is clever I do wonder if there's a more robust way to use the type system to confer this information rather than checking method name suffixes for strings and "type checking" with strings as well.

alvaroaleman: Hm. Maybe test if the method is exported and takes a context.Context as first arg?

stevekuznetsov: Or perhaps a second interface that has only context-using methods? I'm not really sure what's best

alvaroaleman: > Or perhaps a second interface that has only context-using methods? I'm not really sure what's best

That is difficult, because the github client is already divided in a bunch of sub interfaces. Additionally, relying on someone adding this to the right interface is no less error-prone than relying on them adhering to a namescheme.

I've changed this to just test all methods whose first argument is of type `context.Context`. IMHO this is as good as this will ever get.

stevekuznetsov (approved):

Issue #22072, opened 2021-05-04
Prow inrepoconfig: jobs definition in multiple files


chaodaiG:
**What would you like to be added**:

Prow inrepoconfig supports multiple files instead of only `prow.yaml` file. One option is a directory called `.prow` under the root level.

**Why is this needed**:

- Single file is sufficient for a small repo with small number of prow jobs that can be easily managed
- Single file is not very friendly with repo that has long history and lots of prow jobs.
- Also not possible to do fine grained OWNERS control on prow jobs.

The changes required to make this work, should be:
- Other than look for `.prow.yaml` file in the repo, also look for `.prow` directory, which happens in https://github.com/kubernetes/test-infra/blob/ad136bc20c7f33eedb546cd799ccdf04f47bddb9/prow/config/inrepoconfig.go#L34
- Also remember to update `checkconfig` if required at https://github.com/kubernetes/test-infra/blob/master/prow/cmd/checkconfig/main.go#L155
- Documentation would also need to be updated https://github.com/kubernetes/test-infra/blob/master/prow/inrepoconfig.md#L32

chaodaiG:
cc @cjwagner , @alvaroaleman : is there anything blocking us from doing so?

alvaroaleman:
I don't think so, this seems pretty straight forward

chaodaiG:
/good-first-issue
/help

Instructions will be added slightly later

chaodaiG:
Instructions added

fgimenez:
/assign
/remove-help

I'd like to give it a try, @chaodaiG will let you know how it goes :)

PR #22370, opened 2021-05-28
inrepoconfig: jobs definition in multiple files

fgimenez:
Closes: #22072

/cc @chaodaiG 

### DIFF
--- a/prow/config/inrepoconfig.go
+++ b/prow/config/inrepoconfig.go
@@ -22,6 +22,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path"
+	"path/filepath"
 
 	"github.com/sirupsen/logrus"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
@@ -32,6 +33,7 @@ import (
 
 const (
 	inRepoConfigFileName = ".prow.yaml"
+	inRepoConfigDirName  = ".prow"
 )
 
 // ProwYAML represents the content of a .prow.yaml file
@@ -57,7 +59,6 @@ func defaultProwYAMLGetter(
 	headSHAs ...string) (*ProwYAML, error) {
 
 	log := logrus.WithField("repo", identifier)
-	log.Debugf("Attempting to get %q.", inRepoConfigFileName)
 
 	if gc == nil {
 		log.Error("defaultProwYAMLGetter was called with a nil git client")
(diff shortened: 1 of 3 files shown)


fgimenez:
Added a couple more test cases for presets and non-ProwYAML files under `.prow`

prow/config/inrepoconfig.go
```diff
@@ -94,23 +95,61 @@ func defaultProwYAMLGetter(
...
-	if err := yaml.Unmarshal(bytes, prowYAML); err != nil {
-		return nil, fmt.Errorf("failed to unmarshal %q: %v", inRepoConfigFileName, err)
+			c.Postsubmits = append(a.Postsubmits, b.Postsubmits...)
+
+			return c
+		}
+
+		log.Debugf("Attempting to read config files under %q.", inRepoConfigDirName)
+		prowYAMLDirPath := path.Join(repo.Directory(), inRepoConfigDirName)
+		err := filepath.Walk(prowYAMLDirPath, func(p string, info os.FileInfo, err error) error {
```
alvaroaleman, line 128: Do we really want to make these two mutually exclusive?

prow/config/inrepoconfig.go
```diff
@@ -94,23 +95,61 @@ func defaultProwYAMLGetter(
...
+			if info == nil {
```
alvaroaleman, line 129:
```suggestion
			if err != nil {
```

prow/config/inrepoconfig.go
fgimenez: Talked with @chaodaiG about this and agreed to do it this way, fine with me if it is better to support both at the same time.

chaodaiG: Not able to observe a use case of when someone wants to use both `.prow.yaml` and `.prow` at the same time, so would rather enforce `.prow` for future to make our lives easier. I'm planning to further enforce this by a test in https://github.com/kubernetes/test-infra/tree/master/prow/cmd/checkconfig for any repo that enables inrepoconfig.

prow/config/inrepoconfig.go
```diff
@@ -94,23 +95,61 @@ func defaultProwYAMLGetter(
...
+				if err := yaml.Unmarshal(bytes, partialProwYAML); err != nil {
+					return fmt.Errorf("failed to unmarshal %q: %v", p, err)
+				}
+				prowYAML = mergeProwYAML(prowYAML, partialProwYAML)
+			}
+			return err
+		})
+		if err != nil {
+			if os.IsNotExist(err) {
+				log.Debugf("File %q does not exist.", inRepoConfigFileName)
```
chaodaiG, line 148: This should be `inRepoConfigDirName` right?

prow/inrepoconfig.md
```diff
@@ -87,4 +87,11 @@ postsubmits:
       - config/prow/cluster
 ```
 
+It is possible also to use multiple config files with this same format under a `.prow`
+directory in the root of your repo. All the YAML files under the `.prow` directory will
+be read and merged together recursively.This will make it easier to handle big repos with
+a large number of jobs and will allow to have fine grained OWNERS control on them.
+
+When both `.prow.yaml` file and `.prow` directory are present, `.prow.yaml` is preferred.
```
chaodaiG, line 95: Would suggest mentioning these two are mutual exclusive to make this clearer

chaodaiG (commented):
Thank you @fgimenez for the contribution!

Overall looks good me, left a few comments mostly styling nits

prow/config/inrepoconfig.go
```diff
@@ -94,23 +95,59 @@ func defaultProwYAMLGetter(
...
-	}
+	prowYAML := &ProwYAML{}
 
-	bytes, err := ioutil.ReadFile(prowYAMLFilePath)
-	if err != nil {
-		return nil, fmt.Errorf("failed to read %q: %v", inRepoConfigFileName, err)
-	}
+	log.Debugf("Attempting to read config files under %q.", inRepoConfigDirName)
+	prowYAMLDirPath := path.Join(repo.Directory(), inRepoConfigDirName)
+	if fileInfo, err := os.Stat(prowYAMLDirPath); !os.IsNotExist(err) && fileInfo.IsDir() {
```
chaodaiG, line 102: `fileInfo` could be nil when other error happens, please catch that before calling `fileInfo.IsDir()`

prow/config/inrepoconfig.go
```diff
@@ -94,23 +95,59 @@ func defaultProwYAMLGetter(
...
+		mergeProwYAML := func(a, b *ProwYAML) *ProwYAML {
+			c := &ProwYAML{}
+			c.Presets = append(a.Presets, b.Presets...)
```
chaodaiG, line 106: Do we need these empty lines here? Including a few lines below

prow/config/inrepoconfig.go
```diff
@@ -94,23 +95,59 @@ func defaultProwYAMLGetter(
...
-		return nil, fmt.Errorf("failed to check if file %q exists: %v", inRepoConfigFileName, err)
-	}
+	prowYAML := &ProwYAML{}
 
-	bytes, err := ioutil.ReadFile(prowYAMLFilePath)
-	if err != nil {
-		return nil, fmt.Errorf("failed to read %q: %v", inRepoConfigFileName, err)
-	}
+	log.Debugf("Attempting to read config files under %q.", inRepoConfigDirName)
+	prowYAMLDirPath := path.Join(repo.Directory(), inRepoConfigDirName)
```
chaodaiG, line 101: Nit: might be helpful swapping these two lines, and debug print `prowYAMLDirPath `

prow/config/inrepoconfig.go
```diff
@@ -94,23 +95,59 @@ func defaultProwYAMLGetter(
...
+		if err != nil {
+			return nil, fmt.Errorf("failed to read contents of directory %q: %w", inRepoConfigDirName, err)
+		}
+	} else {
+		log.WithField("file", inRepoConfigFileName).Debug("Attempting to get inreconfigfile")
+		prowYAMLFilePath := path.Join(repo.Directory(), inRepoConfigFileName)
+		if _, err := os.Stat(prowYAMLFilePath); err == nil {
+			bytes, err := ioutil.ReadFile(prowYAMLFilePath)
+			if err != nil {
+				return nil, fmt.Errorf("failed to read %q: %w", inRepoConfigFileName, err)
```
chaodaiG, line 141: Also nit: would it be helpful printing `prowYAMLFilePath ` instead? Same below

prow/config/inrepoconfig.go
```diff
@@ -94,23 +95,61 @@ func defaultProwYAMLGetter(
...
+			return c
+		}
+
+		log.Debugf("Attempting to read config files under %q.", inRepoConfigDirName)
+		prowYAMLDirPath := path.Join(repo.Directory(), inRepoConfigDirName)
+		err := filepath.Walk(prowYAMLDirPath, func(p string, info os.FileInfo, err error) error {
+			if err != nil {
+				return err
+			}
+			if !info.IsDir() && filepath.Ext(p) == ".yaml" {
```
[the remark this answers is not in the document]
alvaroaleman: We have it in other places so I think that would be good for consistency: https://github.com/kubernetes/test-infra/blob/df2711fc5cdc0caf6ec2f523feff3635f378a3e5/prow/config/config.go#L1160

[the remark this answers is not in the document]
fgimenez: Ok done and added a test case for it.

prow/config/inrepoconfig.go
fgimenez: Good catch! Fixed

prow/config/inrepoconfig.go
fgimenez: Nope, removed

prow/config/inrepoconfig.go
fgimenez: Absolutely, done

prow/config/inrepoconfig.go
fgimenez: Indeed, changed everywhere, PTAL :)

prow/config/inrepoconfig.go
chaodaiG: Sorry that I might have not made it super clear.  The case of `err != nil` is still not handled, and we probably don't want to go ahead and read `.prow.yaml` file if `os.Stat` failed with any other reasons. One suggestion here is replacing this line with:
```
if fileInfo, err := os.Stat(prowYAMLDirPath); err == nil && fileInfo.IsDir() {
```

And then after this, do:
```
} else if !os.IsNotExist(err) {
  return fmt.Sprintf("reading %q: %w", prowYAMLDirPath, err)
}
```

fgimenez: Sorry, I get it now, already pushed changes PTAL

chaodaiG (approved):
Thank you!

PR #12637, opened 2019-05-14
Enable /project for the website repo

zacharysarah:
This PR enables the `/project` command for the specified k/website team.

/assign @fejta
/assign @taragu 
/area prow
/sig docs

PR #22395, opened 2021-06-02
Configure additional plugins for the Kustomize repo

KnVerey:
@monopole to enable non-admins (like me) to add issues to projects on the Kustomize repo, we'll need to enable the "projects" Prow plugin here. Here's an example of it being done for another repo: https://github.com/kubernetes/test-infra/pull/12637/files. That said, if I'm understanding correctly, only maintainers will be able to manage projects even via Prow: "The project plugin allows members of a GitHub team to set the project and column on an issue or pull request."

I also browsed the [Plugin Catalog](https://prow.k8s.io/plugins) for others we might want to enable while we're here, and found a couple:
- `require-matching-label`: applies "needs X" labels automatically based on regex. I think this would be helpful (though not strictly necessary) for triage party, as well as for filtering through issues directly in Github. Kubectl has this enabled for "needs triage" only. We seem to use "kind" extensively on Kustomize, so this PR proposes enabling both "needs kind" and "needs triage".
- `merge-method-comment`: "adds a comment on how to request a different-from-default merge method to PRs with more than 1 commit" -- it took me a long time to realize this was an option, so I think it could be helpful to contributors!


monopole:
/lgtm

Thanks for figuring this out @KnVerey

config/prow/plugins.yaml
```diff
@@ -454,6 +454,11 @@ project_config:
               To do
             Doc Writing and Editing:
               To do
+    kubernetes-sigs:
+      org_maintainers_team_id: 3162587 # sig-testing-dummy-project-team
```
KnVerey, line 458: Is this right? I copied it from the `kubernetes` org above. Could I instead leave it off?

KnVerey (commented):
/assign @dims

config/prow/plugins.yaml
dims: let's use the `2842373` which is the id for the github admin team. ( see https://github.com/kubernetes/test-infra/blob/75e6b24b1b7afebfdddaba9104da55f5df66f288/config/prow/plugins.yaml#L417-L419 )

KnVerey: Come to think of it, shouldn't the team need to exist on the `kuberentes-sigs` org? I can't see a "Github Admin Team" on either org (presumably it is secret?) but I would assume the ID would be different even if an equivalent exists both places. No admin team for the SIGs org is mentioned in this file AFAICT. At a guess, maybe it is generated from [this part of the org file](https://github.com/kubernetes/org/blob/20eea7174db48d98eec676c8f8b15bd210c19ce4/config/kubernetes-sigs/org.yaml#L8-L17).  But how do I find the ID?

KnVerey:
Since I've been added to the Github team, we no longer have anyone who needs Projects access and doesn't have it. Furthermore, I think that the team referenced likely already gives project access to its members. So if we have folks in the future who are not maintainers and need Project access, we can revisit and potentially create a new team specifically for this kind of thing. But for now, I've removed the project plugin configuration from this PR. That means we don't need to figure out which team should own projects at the Kubernetes-sigs org level, which was blocking this PR.

I think the labels and merge method plugins will still be useful though, so I'd like to merge this if you agree @monopole

monopole:
IMO, the project needs more people who can apply labels.
/lgtm 
(still)

KnVerey:
@dims PTAL

spiffxp (approved):
/approve
/lgtm

PR #22400, opened 2021-06-03
imagebuilder/gcp: remove preset-service-account and reduce the job interval for debugging

cpanato:
- remove the `preset-service-account` because we already set the `serviceAccountName` to be used
- reduce the job interval for debugging purposes


cpanato:
/assign @CecileRobertMichon @codenrhoden 
need your review here to continue my debugging in this job to make it work, thanks!

PR #22414, opened 2021-06-03
add e2e tool that provisions warm pool

olemarkus:


rifelpet (commented):
I'm curious what we expect to see in this job. aws will provision the normal number of instances as well as one in the warm pool which will not join the cluster. we arent running cluster-autoscaler so the ASG will never be scaled up to include the node. does our `kops toolbox dump` and other artifact-fetching commands interact with the warm pool instance?

olemarkus:
I just want to see that hte resources are provisioned and deleted properly. I think running the e2e suite is a bit overkill here.

rifelpet (approved):

PR #22424, opened 2021-06-04
chore: Enable Docker-in-Docker for Azure Disk CSI Driver V2 integration & sanity tests

edreed:
Azure Disk CSI Driver V2 orchestrates volume creation and management through controllers using the Operator model. As a result, it has a dependency on the Kubernetes API server to do its work. To run the sanity and integration tests against the V2 driver, we now create a Minikube environment to host the driver. Docker-in-Docker support in the test jobs is necessary to build the driver container that is then deployed to the Minikube environment.


andyzhangx (approved):
/ok-to-test
/lgtm
/approve

PR #22438, opened 2021-06-05
Document Config Merger

chases2:
Initial documentation for the Config Merger and examples for its use.

xref https://github.com/GoogleCloudPlatform/oss-test-infra/issues/832


chaodaiG (approved):
hmm... couldn't even find a nit :)

Issue #22432, opened 2021-06-05
tide stopped merging Kubernetes PRs for 8 hours and counting


BenTheElder:
**What happened**:

Tide has not merged a PR to Kubernetes in 8+ hours even though there are green + lgtm + approved PRs.

https://prow.k8s.io/?type=batch

**What you expected to happen**:

PRs should keep merging unless they are failing tests or not lgtm + approved

**How to reproduce it (as minimally and precisely as possible)**:

No idea! currently affecting prow.k8s.io for kubernetes/kubernetes repo

**Please provide links to example occurrences, if any**:

https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/batch/pull-kubernetes-typecheck/1400920022850736128

[stuckjob.yml.txt](https://github.com/kubernetes/test-infra/files/6602011/stuckjob.yml.txt) (aside: github will not allow yaml files but will allow .txt so hence the double extension 🤷‍♂️ )

<img width="1345" alt="Screen Shot 2021-06-04 at 9 31 18 PM" src="https://user-images.githubusercontent.com/917931/120880042-a0d71980-c57c-11eb-9323-ea4ef91aa284.png">

**Anything else we need to know?**:

The job seems to be stuck pending since just before 2:00, 7.5 hours ago.

I don't see any evidence that it's continuing on with a serial merge https://github.com/kubernetes/test-infra/issues/13551, or giving up on the pending without artifacts for 7.5+ hours job. As far as I can tell we're just stuck now.

Previous bugs:
- `tide: do not stop merging when one PR is unmergable` https://github.com/kubernetes/test-infra/issues/10001 (no fix)
- `tide: serial merges should occur when batches fail` https://github.com/kubernetes/test-infra/issues/13551 (still open / frozen)

alvaroaleman:
So we correctly created a pod and transitioned the job to pending. Then nothing happened for about 15hours and after that the job was set to "Pod pending timeout". This should have happened after 15 mins though: https://github.com/kubernetes/test-infra/blob/df2711fc5cdc0caf6ec2f523feff3635f378a3e5/config/prow/config.yaml#L7

My ad hoc theory: We currently just do nothing if the pod is pending and we are below the `PodPendingTimeout`. Unfortunately this means that unless something modifies the pod or prowjob, we are never going to re-check on it again which means we are not going to correctly mark it as failed after 15 mins.

alvaroaleman:
/area prow/plank

[...]

PR #22439, opened 2021-06-05
Plank: Recheck on pending pods once they reached their timeout

alvaroaleman:
We have timeouts for pod scheduling and pod pending, however unless
something happens to change the pod or the prowjob when that timeout is
reached, nothing will trigger Plank to set the job to failed once the
timeout was reached.

This change makes plank re-check on pods that are not scheduled or not
started and below the respective timeout once the timeout was reached.

Much easier to review with whitespace changes disabled btw.

Fixes https://github.com/kubernetes/test-infra/issues/22432

### DIFF
--- a/prow/plank/controller_test.go
+++ b/prow/plank/controller_test.go
@@ -29,6 +29,7 @@ import (
 	"text/template"
 	"time"
 
+	"github.com/google/go-cmp/cmp"
 	"github.com/sirupsen/logrus"
 	v1 "k8s.io/api/core/v1"
 	kapierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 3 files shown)


prow/plank/controller_test.go
```diff
@@ -1335,7 +1337,7 @@ func TestSyncPendingJob(t *testing.T) {
 			ExpectedURL:      "homeless/error",
 		},
 		{
-			Name: "scheduled, pending started more than podUnscheduledTimeout ago",
```
alvaroaleman, line 1338: While the old testname correctly described the situatation, the podScheduledTimeout is completely irrelevant once the pod was scheduled so I have changed it to clarify

petr-muller (approved):

PR #22444, opened 2021-06-07
images/alpine: Build for arm64

LorbusChris:
This commit adds a Google Cloud Build definition for building alpine
arm64 images and pushing them as a manifest list together with their
amd64 counterparts.

This is a prereq for #16588 and #22362

~~Blocked on: https://github.com/kubernetes/test-infra/pull/22599~~

/cc @stevekuznetsov

### DIFF
--- a/images/alpine/cloudbuild.yaml
+++ b/images/alpine/cloudbuild.yaml
@@ -1,19 +1,21 @@
 steps:
-  - name: gcr.io/cloud-builders/docker
+  - name: gcr.io/k8s-testimages/gcb-docker-gcloud
+    entrypoint: /buildx-entrypoint
     args:
     - build
-    - -t
-    - gcr.io/$PROJECT_ID/alpine:$_GIT_TAG
+    - --tag=gcr.io/$PROJECT_ID/alpine:$_GIT_TAG
+    - --platform=linux/amd64,linux/arm64/v8
     - --build-arg=IMAGE_ARG=gcr.io/$PROJECT_ID/alpine:$_GIT_TAG
+    - --push
     - .
     dir: .
-  - name: gcr.io/cloud-builders/docker
+  - name: gcr.io/k8s-testimages/gcb-docker-gcloud
+    entrypoint: gcloud
     args:
-    - tag
+    - container
+    - images
+    - add-tag
     - gcr.io/$PROJECT_ID/alpine:$_GIT_TAG
     - gcr.io/$PROJECT_ID/alpine:latest
 substitutions:
   _GIT_TAG: '12345'
-images:
-  - 'gcr.io/$PROJECT_ID/alpine:$_GIT_TAG'
-  - 'gcr.io/$PROJECT_ID/alpine:latest'


LorbusChris:
looking at the file history, @fejta might be best suited for a review of these changes.
/cc @fejta 
/uncc @BenTheElder

spiffxp:
/cc
I'm going to guess this will fail in the same way as https://github.com/kubernetes/test-infra/pull/22450#issuecomment-858747549

images/alpine/cloudbuild.yaml
```diff
@@ -1,19 +1,13 @@
 steps:
-  - name: gcr.io/cloud-builders/docker
+  - name: gcr.io/k8s-staging-test-infra/docker-buildx:latest
```
LorbusChris, line 2: TODO: replace with `gcr.io/k8s-prow/docker-buildx` once https://github.com/kubernetes/test-infra/pull/22573 has merged

images/alpine/cloudbuild.yaml
```diff
@@ -1,19 +1,14 @@
 steps:
-  - name: gcr.io/cloud-builders/docker
+  - name: gcr.io/k8s-staging-test-infra/gcb-docker-gcloud
```
LorbusChris, line 2: TODO: replace with gcr.io/k8s-testimages/gcb-docker-gcloud once #22599 has merged

LorbusChris:
#22599 has landed. This PR should be ready now.

images/alpine/cloudbuild.yaml
```diff
@@ -1,19 +1,14 @@
...
     dir: .
-  - name: gcr.io/cloud-builders/docker
-    args:
-    - tag
-    - gcr.io/$PROJECT_ID/alpine:$_GIT_TAG
-    - gcr.io/$PROJECT_ID/alpine:latest
 substitutions:
   _GIT_TAG: '12345'
-images:
-  - 'gcr.io/$PROJECT_ID/alpine:$_GIT_TAG'
```
fejta, line 18: What is the reason for dropping these? (or rather moving pushing them into the build step)
https://cloud.google.com/build/docs/build-config#images

LorbusChris: This doesn't seem to work with manifests.
Using:
```
steps:
  - name: gcr.io/k8s-testimages/gcb-docker-gcloud
    entrypoint: /buildx-entrypoint
    args:
    - build
    - --tag=gcr.io/$PROJECT_ID/alpine:$_GIT_TAG
    - --tag=gcr.io/$PROJECT_ID/alpine:latest
    - --platform=linux/amd64,linux/arm64/v8
    - --build-arg=IMAGE_ARG=gcr.io/$PROJECT_ID/alpine:$_GIT_TAG
    - .
    dir: .
substitutions:
  _GIT_TAG: '12345'
images:
  - 'gcr.io/$PROJECT_ID/alpine:$_GIT_TAG'
  - 'gcr.io/$PROJECT_ID/alpine:latest'

```
yields:
```
ERROR: failed to find one or more images after execution of build steps: ["gcr.io/k8s-staging-test-infra/alpine:v20210618-1bc5e9d8bf" "gcr.io/k8s-staging-test-infra/alpine:latest"]
```

LorbusChris:
/hold
using the `--tag` flag twice in one command doesn't seem to work and will actually only tag the latter

images/alpine/cloudbuild.yaml
BenTheElder: GCR specific but cheap and simple https://cloud.google.com/sdk/gcloud/reference/container/images/add-tag

To do it generically you'd need to push the manifest again as a tag (and only the manifest, the underlying resources should already be pushed), [`crane`](https://github.com/google/go-containerregistry/tree/main/cmd/crane) can do this (and other super useful things!) see: [`crane tag`](https://github.com/google/go-containerregistry/blob/main/cmd/crane/doc/crane_tag.md)

LorbusChris:
Thanks @BenTheElder, let's go the easy way with `gcloud` as that's in the image already.
Should be fixed now.
/hold cancel

BenTheElder (approved):
lgtm

LorbusChris:
@BenTheElder it looks like this broke the postsubmit job https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-test-infra-push-alpine/1405964567472771072

LorbusChris:
https://github.com/kubernetes/test-infra/pull/22609

PR #22451, opened 2021-06-07
Split eviction tests out from node-serial job

ehashman:
~Note: this is only for the PR job so we can test this.~

~Seeing if excluding these tests from the serial job will avoid the 7h timeout...~

~I suggest if this succeeds that we split these into their own job, wdyt?~

Splits the node serial jobs in half so the eviction tests have their own job. Without them, per tests on https://github.com/kubernetes/kubernetes/pull/102691 and https://github.com/kubernetes/kubernetes/pull/102708 the serial tests take under 90m.

xref https://github.com/kubernetes/kubernetes/issues/102148#issuecomment-854777565

/cc @fromanirh @SergeyKanzhelev @dims

### DIFF
--- a/config/jobs/kubernetes/sig-node/node-kubelet.yaml
+++ b/config/jobs/kubernetes/sig-node/node-kubelet.yaml
@@ -203,6 +203,36 @@ periodics:
     description: "Contains all uncategorized tests, these are tests which are not marked with a [Node] tag. All node tests should be marked with [NodeFeature] or [NodeSpecialFeature] or [NodeAlphaFeature] or [NodeConformance] classification. Also skipped are [Flaky], [Benchmark], [Legacy]."
 
 - name: ci-kubernetes-node-kubelet-serial
+  interval: 4h
+  labels:
+    preset-service-account: "true"
+    preset-k8s-ssh: "true"
+  spec:
+    containers:
+    - image: gcr.io/k8s-testimages/kubekins-e2e:v20210601-ea6aa4e-master
+      args:
+      - --repo=k8s.io/kubernetes=master
+      - --timeout=240
+      - --root=/go/src
+      - --scenario=kubernetes_e2e
+      - --
+      - --deployment=node
+      - --gcp-project-type=node-e2e-project
+      - --gcp-zone=us-west1-b
+      - --node-args=--image-config-file=/workspace/test-infra/jobs/e2e_node/image-config-serial.yaml
+      - --node-test-args=--feature-gates=DynamicKubeletConfig=true,LocalStorageCapacityIsolation=true --kubelet-flags="--cgroups-per-qos=true --cgroup-root=/"
+      - --node-tests=true
+      - --provider=gce
+      - --test_args=--nodes=1 --focus="\[Serial\]" --skip="\[Flaky\]|\[Benchmark\]|\[NodeSpecialFeature:.+\]|\[NodeAlphaFeature:.+\]|\[NodeFeature:Eviction\]"
+      - --timeout=180m
+      env:
+      - name: GOPATH
+        value: /go
+  annotations:
+    testgrid-dashboards: sig-node-kubelet
+    testgrid-tab-name: node-kubelet-serial
+
+- name: ci-kubernetes-node-kubelet-eviction
   interval: 4h30m
   labels:
     preset-service-account: "true"
@@ -223,14 +253,14 @@ periodics:
       - --node-test-args=--feature-gates=DynamicKubeletConfig=true,LocalStorageCapacityIsolation=true --kubelet-flags="--cgroups-per-qos=true --cgroup-root=/"
       - --node-tests=true
       - --provider=gce
-      - --test_args=--nodes=1 --focus="\[Serial\]" --skip="\[Flaky\]|\[Benchmark\]|\[NodeSpecialFeature:.+\]|\[NodeAlphaFeature:.+\]"
+      - --test_args=--nodes=1 --focus="\[NodeFeature:Eviction\]"
       - --timeout=300m
       env:
       - name: GOPATH
         value: /go
   annotations:
     testgrid-dashboards: sig-node-kubelet
-    testgrid-tab-name: node-kubelet-serial
+    testgrid-tab-name: node-kubelet-eviction
 
 - name: ci-kubernetes-node-kubelet-serial-cpu-manager
   interval: 4h
--- a/config/jobs/kubernetes/sig-node/sig-node-presubmit.yaml
+++ b/config/jobs/kubernetes/sig-node/sig-node-presubmit.yaml
@@ -221,6 +221,40 @@ presubmits:
       - image: gcr.io/k8s-testimages/kubekins-e2e:v20210601-ea6aa4e-master
         args:
         - "--repo=k8s.io/$(REPO_NAME)=$(PULL_REFS)"
+        - --timeout=240
+        - --root=/go/src
+        - "--upload=gs://kubernetes-jenkins/pr-logs"
+        - --scenario=kubernetes_e2e
+        - --
+        - --deployment=node
+        - --gcp-project-type=node-e2e-project
+        - --gcp-zone=us-west1-b
+        - --node-args=--image-config-file=/workspace/test-infra/jobs/e2e_node/image-config-serial.yaml
+        - --node-test-args=--feature-gates=DynamicKubeletConfig=true,LocalStorageCapacityIsolation=true --kubelet-flags="--cgroups-per-qos=true --cgroup-root=/"
+        - --node-tests=true
+        - --provider=gce
+        - --test_args=--nodes=1 --focus="\[Serial\]" --skip="\[Flaky\]|\[Benchmark\]|\[NodeSpecialFeature:.+\]|\[NodeAlphaFeature:.+\]|\[NodeFeature:Eviction\]"
+        - --timeout=180m
+        env:
+        - name: GOPATH
+          value: /go
+  - name: pull-kubernetes-node-kubelet-eviction
+    always_run: false
+    optional: true
+    skip_report: false
+    skip_branches:
+    - release-\d+\.\d+  # per-release image
+    annotations:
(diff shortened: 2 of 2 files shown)


fromanirh:
/lgtm
because:
1. partial lane is better than no lane (- of course if it is indeed caused by eviction test we will need to dive deeper and fix them)
2. we are still monitoring the lane state so this is not the end of the story, but rather a step to move forward.

odinuge:
This change is ok, but I would prefer to keep all these jobs the same to avoid confusion and later a skew between then.

Isn't it easier to disable/skip the tests inside a `k8s/k8s` PR instead for testing? That is quicker, and allow us to run the pre submits with cri-o as well.

Agree that we can look at doing this in long term tho, to make the test suite more stable. :+1:

fromanirh:
> This change is ok, but I would prefer to keep all these jobs the same to avoid confusion and later a skew between then.
> 
> Isn't it easier to disable/skip the tests inside a `k8s/k8s` PR instead for testing? That is quicker, and allow us to run the pre submits with cri-o as well.

That's very interesting to verify if indeed skipping the eviction tests helps, but i can't think about a easy way atm (while I can think few pretty convoluted ways :\ ). Do you have any suggestion?

odinuge:
> That's very interesting to verify if indeed skipping the eviction tests helps, but i can't think about a easy way atm (while I can think few pretty convoluted ways :\ ). 

Not sure if we are talking about the same thing then.

> Do you have any suggestion?

By doing `git rm test/e2e_node/eviction_test.go` in k8s/k8s, and opening a PR with that change, we can verify using the pre submit tests, this test suite included. If that gives the satisfied results, we can update all the `serial` tests to temporary skip them while we work on fixing the issue. Does that make sense?

fromanirh:
> > That's very interesting to verify if indeed skipping the eviction tests helps, but i can't think about a easy way atm (while I can think few pretty convoluted ways :\ ).
> 
> Not sure if we are talking about the same thing then.

Nah, it's me, I was thinking about ginkgo flags/invokations

> 
> > Do you have any suggestion?
> 
> By doing `git rm test/e2e_node/eviction_test.go` in k8s/k8s, and opening a PR with that change, we can verify using the pre submit tests, this test suite included. If that gives the satisfied results, we can update all the `serial` tests to temporary skip them while we work on fixing the issue. Does that make sense?
> 
> We can also just add `[Flaky]` to those tests as well, and remove it when they work properly again.

Yes, this could be a good idea. Let me work on this later today. Let's put this on hold until we know a bit more.

fromanirh:
> > By doing `git rm test/e2e_node/eviction_test.go` in k8s/k8s, and opening a PR with that change, we can verify using the pre submit tests, this test suite included. If that gives the satisfied results, we can update all the `serial` tests to temporary skip them while we work on fixing the issue. Does that make sense?
> > We can also just add `[Flaky]` to those tests as well, and remove it when they work properly again.
> 
> Yes, this could be a good idea. Let me work on this later today. Let's put this on hold until we know a bit more.

Here: https://github.com/kubernetes/kubernetes/pull/102691

fromanirh:
> > > By doing `git rm test/e2e_node/eviction_test.go` in k8s/k8s, and opening a PR with that change, we can verify using the pre submit tests, this test suite included. If that gives the satisfied results, we can update all the `serial` tests to temporary skip them while we work on fixing the issue. Does that make sense?
> > > We can also just add `[Flaky]` to those tests as well, and remove it when they work properly again.
> > 
> > 
> > Yes, this could be a good idea. Let me work on this later today. Let's put this on hold until we know a bit more.
> 
> Here: [kubernetes/kubernetes#102691](https://github.com/kubernetes/kubernetes/pull/102691)

still no luck. Timed out after 420m. On my environment the average completion time is around 240m. 
Next step for me is understand if it is now still too slow or if there is another bad test.

ehashman:
Testing with https://github.com/kubernetes/kubernetes/pull/102708

fromanirh:
> @fromanirh:
> 
> > still no luck. Timed out after 420m. On my environment the average completion time is around 240m.
> > Next step for me is understand if it is now still too slow or if there is another bad test.
> 
> Strange... When you say locally, do you mean on your machine or on gce? The test suite runs on `n1-standard-2`, as context. So only 2 vCPUs.

I mean my environment. I have a machine I can use freely - but I can't make it available to the community. It's a old dual xeon (2x 6cores) with SSD disks. And thanks for the reminder, I somehow mixed up and thought we run tests on a much more powerful machine.

> What about taking all the `[Disruptive]` and "skipping" them to see what happen? I remember a few years ago when the "Critical pods" tests made everything unstable...

Yes, this is the next step. If even this fails, I will start from the opposite direction: clean sheet, adding test gradually.

odinuge:
Ahh. I think I have found our "issue"... It looks like the presubmit job runs on master, not the corresponding pr. The log upload is also wrong, so longs end up in the wrong folder. :sweat_smile: 

Could anyone take a look at https://github.com/kubernetes/test-infra/pull/22493? @ehashman @fromanirh

fromanirh:
> Ahh. I think I have found our "issue"... It looks like the presubmit job runs on master, not the corresponding pr. The log upload is also wrong, so longs end up in the wrong folder. sweat_smile
> 
> Could anyone take a look at #22493? @ehashman @fromanirh

good catch! lgtm'd, I think this is definitely a fix we want anyway

ehashman:
Based on the testing done in our two DNM PRs, I've split off the eviction tests only into their own job from the rest of the node-serial tests.

odinuge:
Thanks @ehashman!

However, can we let this soak for a few more days to actually see if we can find the root cause? The test failed after 3 hours in https://github.com/kubernetes/kubernetes/pull/102169, even with some Eviction tests passing. Still looks like there are some "too many files" for ginko, but it looks like a start.. Hopefully that is repeatable, and not just a one of.

https://github.com/kubernetes/kubernetes/pull/102209 should hopefully work as well, giving us output when the test times out, for easier debugging.

ehashman:
> Thanks @ehashman!
> 
> However, can we let this soak for a few more days to actually see if we can find the root cause? The test failed after 3 hours in [kubernetes/kubernetes#102169](https://github.com/kubernetes/kubernetes/pull/102169), even with some Eviction tests passing. Still looks like there are some "too many files" for ginko, but it looks like a start.. Hopefully that is repeatable, and not just a one of.

I'm not sure what we're waiting on to soak. The tests will still be broken if this merges, this just ensures the jobs will complete faster by splitting them into two more reasonably-sized jobs. It's going to be a lot easier to run (and hence fix!) the tests if we split them since they will run faster, and this also allows us to divide and conquer the CI fixes: someone can work on the main serial tests, someone else can fix the eviction tests in parallel.

fromanirh:
TL;DR: the kitchen sink approach in the serial lane was already something we needed to fix and this is a good chance to do improve. +1.

My take:
- just to make it obvious: the search is indeed not over yet and there are more fixes needed in quite some places to have stable and dependable lane. But I think we all pretty much agree here and splitting the lanes should give us better velocity.
- in general I'm all for more focused lanes with reasonnable amount of focused test, which take reasonnable time to complete, for all the usual well known reasons: we can run them more often (ideally every time) so we keep the tests healthy and working, we have better coverage and so fort.
- we have quite a lot of serial/disruptive tests, and that's ok. we need to continuously balance all the needs
- one thing we can lose in the process is to make sure tests don't leak/pollute the cluster state. This is hard to check during the test review stage, and we only incidentally checked with the giant lane with hard-to-debug scenarios. I wonder if there is a better way to tackle this issue.

fromanirh:
/lgtm
because reasons outlined in https://github.com/kubernetes/test-infra/pull/22451#issuecomment-858298134

fromanirh:
holding to give time to let others review and/or the discussion progress further (if needed)
feel free to remove anytime

config/jobs/kubernetes/sig-node/node-kubelet.yaml
```diff
@@ -203,6 +203,36 @@ periodics:
...
+      - --test_args=--nodes=1 --focus="\[Serial\]" --skip="\[Flaky\]|\[Benchmark\]|\[NodeSpecialFeature:.+\]|\[NodeAlphaFeature:.+\]|\[NodeFeature:Eviction\]"
+      - --timeout=180m
+      env:
+      - name: GOPATH
+        value: /go
+  annotations:
+    testgrid-dashboards: sig-node-kubelet
+    testgrid-tab-name: node-kubelet-serial
+
+- name: ci-kubernetes-node-kubelet-eviction
```
cynepco3hahue, line 235: Should we configure a new test grid tab for this job?

ehashman: We do so on line now 256: https://github.com/kubernetes/test-infra/pull/22451/files#diff-14f46258d64962b171c1662b99c4e01ea48674ca3ef00b1d9dec13acf2ad91f9R263

bobbypage:
/lgtm

+1 on splitting the serial tests, there were too many serial tests running in the same job. I think we should consider splitting serial tests further (e.g. disruptive) if needed since I think it's easier to debug/manage greater number of distinct jobs with fewer tests each.

odinuge:
Did some more testing, and kubernetes/kubernetes#102169 fix the most evident issues with the eviction tests (makes the current test suite take 3-4 hours). Still agree that we can split them and continue with working with them separately.

What is the plan with the eviction tests that work? Should we add them back one by one, or is the plan to keep them separate as long as one is failing? I think I like doing the latter; maybe even more with separating all the `Disruptive` tests into one test suite.

I would also prefer to do the same with `pull-kubernetes-node-kubelet-serial-crio-cgroupv{1,2}` to keep the consistency, so it would be nice if those could be updated as well.

ehashman:
@odinuge if you're okay with splitting can you please remove the hold?

ehashman:
Can we please propose that in a separate PR? I don't think it's relevant to unblocking the node-serial job.

odinuge:
Assuming you wanted to remove the hold and get this merged @ehashman?

/lgtm
/hold cancel

PR #22464, opened 2021-06-07
images/builder: update to cloud-sdk 343.0

spiffxp:
I think the version of `gcloud builds submit` at 258.0.0 doesn't understand `options.machineType` which is breaking the job introduced in https://github.com/kubernetes/test-infra/pull/22458 which is supposed to build https://github.com/kubernetes/k8s.io/pull/2134


BenTheElder (approved):
LGTM

PR #22472, opened 2021-06-08
chore: Remove sanity & integration test pipelines due to Minukube dependency

edreed:
Removes the sanity & integration test pipelines because the testing the Azure Disk CSI Driver V2 requires Minikube which is impractical to run in a container. We will use Github actions to run these tests instead.


andyzhangx (approved):
/lgtm
/approve
/ok-to-test

PR #22477, opened 2021-06-08
Update Kube dependencies to v1.21 level

alvaroaleman:
No specific reason other than hoping things will be somehow better :)


prow/plank/controller_test.go
```diff
@@ -1512,7 +1512,7 @@ func TestSyncPendingJob(t *testing.T) {
 			},
 			ExpectedState:    prowapi.PendingState,
 			ExpectedComplete: false,
-			ExpectedNumPods:  1,
+			ExpectedNumPods:  0,
```
alvaroaleman, line 1515: Needed because of https://github.com/kubernetes-sigs/controller-runtime/pull/1399, the fakeclient now handles finalizers by itself so updating an object with deletion timestamp set to remove a finalizer will result in deletion

BenTheElder:
What is prow.k8s.io running these days?

alvaroaleman:
This is just about the client libs, since we don't use any features that require a new apiserver version this should work just fine even with an old apiserver

cjwagner (approved):
> What is prow.k8s.io running these days?

Currently 1.17.17-gke.4900

PR #22478, opened 2021-06-08
Add debug message when crier skips reporting

chaodaiG:


fejta (approved):
Isn't this excessively noisy?
/hold

chaodaiG:
> Isn't this excessively noisy?
> /hold

Possibly, because of periodic jobs. Would `debug` be preferred?

fejta:
I just don't think this PR is necessary at all

chaodaiG:
I don't have a strong preference here, and since I had figured out why the report was skipped by looking at source code, so
/close

PR #22481, opened 2021-06-08
Generic autobumper skips running based on 'Active' field

chaodaiG:


prow/cmd/generic-autobumper/main.go
```diff
@@ -213,34 +213,54 @@ func validateOptions(o *options) error {
 	return nil
 }
 
+func isOncallActive(oncallAddress, oncallGroup string) bool {
+	_, oncallActive, _ := getOncallInfo(oncallAddress, oncallGroup)
```
mpherman2, line 217: nit: It looks like we are calling getOncallInfo twice (once here and once when generate the PR body) We could reduce it to 1 call and pass the assignment into the client instead of having PRTitleBody call getAssignment().

mpherman2:
/lgtm
/hold
Overall lgtm. Noticed one thing that could be changed, but don't think it is necessary.

prow/cmd/generic-autobumper/main.go
chaodaiG: It's a good point. Tried it but unfortunately not able to come up with clean implementation due to the complicated string formatted return value of `getAssignment`.  Will think about this more in the future

PR #22486, opened 2021-06-08
skip_report flag in prowjob only used by crier reporters that report back to CI system

chaodaiG:
`skip_report` initially was meant for skipping reporting on github as comment at https://github.com/kubernetes/test-infra/blob/3be10e82308d2c30a449e79bf6d626e087cd580a/prow/config/jobs.go#L231. This has later on been used for skipping all crier reporters as https://github.com/kubernetes/test-infra/blob/6c4a2b34246d3bf1a4d6f2c3fa7fc8e46520b9ce/prow/crier/controller.go#L181 and https://github.com/kubernetes/test-infra/blob/6c4a2b34246d3bf1a4d6f2c3fa7fc8e46520b9ce/prow/pjutil/pjutil.go#L126. 

gcsreporter is responsible for uploading `finished.json` when sidecar failed to do so, as well as uploading `prowjob.json`. These should happen independent of `skip_report`.

This PR classifies crier reporters to be either report back to or not to CI system, and `skip_report` is only respected by reporters that report back to CI system.

### DIFF
--- a/prow/crier/controller.go
+++ b/prow/crier/controller.go
@@ -178,7 +178,7 @@ func (r *reconciler) reconcile(ctx context.Context, log *logrus.Entry, req recon
 
 	log = log.WithField("jobName", pj.Spec.Job)
 
-	if !pj.Spec.Report || !r.reporter.ShouldReport(ctx, log, &pj) {
+	if !r.reporter.ShouldReport(ctx, log, &pj) {
 		return nil, nil
 	}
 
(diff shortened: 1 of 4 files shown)


fejta:
This PR is unnecessary

chaodaiG:
> This PR is unnecessary

Had synced offline with @fejta, and updated PR description with more context, also this PR is open to discussion

chaodaiG:
The other reporter could benefit from this is pubsub reporter

fejta (approved):

chaodaiG:
@alvaroaleman , any objection before I unhold this PR?

alvaroaleman:
no, sounds good to me. It makes the meaning of this field a bit confusing (maybe godoc that reporters themselves decide if they respect it?) but having something like k8s reporter always report makes a lot of sense.
/hold cancel

Issue #22456, opened 2021-06-07
Github throttler allows up to 2x hourly token budget in the first hour


alvaroaleman:
**What happened**:

When set up, the github throttler will both fill up the `burst` channel entirely and start a goroutine to refill it. This allows up to 2x `hourly-tokens` usage in the first hour:
https://github.com/kubernetes/test-infra/blob/c868e582b681ad469d3fb6b3b407be4c610ca40f/prow/github/client.go#L527-L540

@fejta you introduced the throttler in https://github.com/kubernetes/test-infra/pull/5620 and @cjwagner you reviewed that change. Is this behavior intended? I find this very surprising.

/area prow
/area prow/github

**What you expected to happen**:

The github throttler to only start a goroutine to refill the budget but not fill up the `burst` channel entirely

**How to reproduce it (as minimally and precisely as possible)**:

**Please provide links to example occurrences, if any**:

**Anything else we need to know?**:

cjwagner:
/assign
Good catch. I've thought about a few different ways to handle this and ended up with this: https://github.com/kubernetes/test-infra/pull/22489

PR #22489, opened 2021-06-09
Prevent GH client throttling from exceeding its cumulative token limit by the burst size.

cjwagner:
fixes https://github.com/kubernetes/test-infra/issues/22456

The idea behind pre-filling the channel is to prevent throttling immediately after the client is created. I don't think we considered that this allows the client to exceed the number of tokens it should have used in a given time by the burst size. If the burst size is the hourly limit then this makes it possible for the client to use double the hourly limit in the first hour.

I think we do need to at least partially pre-fill the channel so that clients are immediately usable. I considered adding another flag for controlling how much the channel is pre-filled, but I think there is a simpler and more correct answer:
We can just delay the goroutine that refills the channel for a while before it begins to refill. Specifically, delay it the amount of time it would take to refill the burst channel. This ensures the client never exceeds the hourly token limit while also providing the full burst at startup.

/assign @alvaroaleman

### DIFF
--- a/prow/github/client.go
+++ b/prow/github/client.go
@@ -539,13 +539,18 @@ func (c *client) Throttle(hourlyTokens, burst int, orgs ...string) error {
 		}
 		return nil
 	}
-	rate := time.Hour / time.Duration(hourlyTokens)
-	ticker := time.NewTicker(rate)
+	period := time.Hour / time.Duration(hourlyTokens) // Duration between token refills
+	ticker := time.NewTicker(period)
 	throttle := make(chan time.Time, burst)
 	for i := 0; i < burst; i++ { // Fill up the channel
 		throttle <- time.Now()
 	}
 	go func() {
+		// Before refilling, wait the amount of time it would have taken to refill the burst channel.
+		// This prevents granting too many tokens in the first hour due to the initial burst.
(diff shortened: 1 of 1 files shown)


prow/github/client.go
```diff
@@ -539,13 +539,18 @@ func (c *client) Throttle(hourlyTokens, burst int, orgs ...string) error {
...
+	period := time.Hour / time.Duration(hourlyTokens) // Duration between token refills
+	ticker := time.NewTicker(period)
 	throttle := make(chan time.Time, burst)
 	for i := 0; i < burst; i++ { // Fill up the channel
 		throttle <- time.Now()
 	}
 	go func() {
+		// Before refilling, wait the amount of time it would have taken to refill the burst channel.
+		// This prevents granting too many tokens in the first hour due to the initial burst.
+		for i := 0; i < burst; i++ {
```
alvaroaleman, line 551: That's elegant!

alvaroaleman (approved):

prow/github/client.go
cjwagner: Thanks, I had a `time.Sleep(period * time.Duration(burst))` at first, but this solution doesn't race with the ticker.

PR #22493, opened 2021-06-09
Fix log upload on pull-kubernetes-node-kubelet-serial

odinuge:
Will this fix the issue where no logs appear like in https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/102209/pull-kubernetes-node-kubelet-serial/1402355480164044800/?

I also believe this pre-submit run on master branch, not the PR it is started on, ref this log output: https://storage.googleapis.com/kubernetes-jenkins/logs/pull-kubernetes-node-kubelet-serial/1402355480164044800/build-log.txt

### DIFF
--- a/config/jobs/kubernetes/sig-node/sig-node-presubmit.yaml
+++ b/config/jobs/kubernetes/sig-node/sig-node-presubmit.yaml
@@ -220,9 +220,10 @@ presubmits:
       containers:
       - image: gcr.io/k8s-testimages/kubekins-e2e:v20210601-ea6aa4e-master
         args:
-        - --repo=k8s.io/kubernetes=master
+        - "--repo=k8s.io/$(REPO_NAME)=$(PULL_REFS)"
         - --timeout=440
         - --root=/go/src
+        - "--upload=gs://kubernetes-jenkins/pr-logs"
         - --scenario=kubernetes_e2e
         - --
         - --deployment=node


fromanirh:
/lgtm
I think this is a good fix to have anyway :)

config/jobs/kubernetes/sig-node/sig-node-presubmit.yaml
```diff
@@ -220,9 +220,10 @@ presubmits:
       containers:
       - image: gcr.io/k8s-testimages/kubekins-e2e:v20210601-ea6aa4e-master
         args:
-        - --repo=k8s.io/kubernetes=master
+        - "--repo=k8s.io/$(REPO_NAME)=$(PULL_REFS)"
```
fromanirh, line 223: just to doublecheck: this is a more generic case of the former code we had, right? meaning: after your change we still can handle the previous case IIUC, so we're just more flexible, right?

odinuge: I suspect that this is causing the pre-submit to always run on master, not on the PR iteslf (but I am not that familiar with this stuff). If you look here, it runs on master: https://storage.googleapis.com/kubernetes-jenkins/logs/pull-kubernetes-node-kubelet-serial/1402355480164044800/build-log.txt

All the other pre-submits have this as well..

ehashman: Oops, I missed that, good catch.

ehashman (commented):
/lgtm

PR #22502, opened 2021-06-09
Prow: skip_if_only_changed, the inverse of run_if_changed

2uasimojo:
Introduce a new prow config option, `skip_if_only_changed`, which is the opposite of `run_if_changed`.

`run_if_changed` positively looks for changed paths matching a regex and triggers the job (which must also be `always_run: false`). This is painful to use when the intent is to skip jobs that *only* touch a certain subset of paths -- particularly since golang regexes don't support lookarounds. For example, in order to say "Skip this job for changes that only touch paths under the `docs/` subdirectory," it was previously necessary to do something like:

`run_if_changed: ^([^d]|d[^o]|do[^c]|doc[^s]|docs[^/])`

With this change, the same thing can be accomplished via:

`skip_if_only_changed: ^docs/`

which is much more scrutable.

### DIFF
--- a/prow/config/config.go
+++ b/prow/config/config.go
@@ -2270,8 +2270,16 @@ func validatePodSpec(jobType prowapi.ProwJobType, spec *v1.PodSpec, decorationEn
 }
 
 func validateTriggering(job Presubmit) error {
-	if job.AlwaysRun && job.RunIfChanged != "" {
-		return fmt.Errorf("job %s is set to always run but also declares run_if_changed targets, which are mutually exclusive", job.Name)
+	if job.AlwaysRun {
+		if job.RunIfChanged != "" {
+			return fmt.Errorf("job %s is set to always run but also declares run_if_changed targets, which are mutually exclusive", job.Name)
+		}
+		if job.SkipIfOnlyChanged != "" {
+			return fmt.Errorf("job %s is set to always run but also declares skip_if_only_changed targets, which are mutually exclusive", job.Name)
+		}
+	}
+	if job.RunIfChanged != "" && job.SkipIfOnlyChanged != "" {
+		return fmt.Errorf("job %s declares run_if_changed and skip_if_only_changed, which are mutually exclusive", job.Name)
 	}
 
 	if (job.Trigger != "" && job.RerunCommand == "") || (job.Trigger == "" && job.RerunCommand != "") {
@@ -2451,10 +2459,16 @@ func setBrancherRegexes(br Brancher) (Brancher, error) {
 }
 
 func setChangeRegexes(cm RegexpChangeMatcher) (RegexpChangeMatcher, error) {
-	if cm.RunIfChanged != "" {
-		re, err := regexp.Compile(cm.RunIfChanged)
+	var reString, propName string
+	if reString = cm.RunIfChanged; reString != "" {
+		propName = "run_if_changed"
+	} else if reString = cm.SkipIfOnlyChanged; reString != "" {
+		propName = "skip_if_only_changed"
+	}
+	if reString != "" {
+		re, err := regexp.Compile(reString)
 		if err != nil {
-			return cm, fmt.Errorf("could not compile run_if_changed regex: %v", err)
+			return cm, fmt.Errorf("could not compile %s regex: %v", propName, err)
 		}
 		cm.reChanges = re
 	}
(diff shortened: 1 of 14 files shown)


2uasimojo:
TODO: Tests, docs.

Want to see if this idea has legs first.

BenTheElder:
last time I saw any new run_if_changed type fields discussed it was rejected IIRC, I'll leave it up to prow/OWNERS to debate that

BenTheElder:
aside: the bot is still quite bad at picking reviewers from the directories touched instead of the repo root :/

cblecker:
/uncc @michelle192837 
/cc @cblecker @alvaroaleman @stevekuznetsov @cjwagner @fejta 

Thoughts? There could be some value here, and the change is honestly smaller than I thought it would be when Eric was pitching me on it.

alvaroaleman:
Yeah, idea seems fine to me. Not 100% happy with the name but I can't come up with a better one either. Would like to hear what cole and @chaodaiG think though.

2uasimojo:
Thanks Alvaro.

Definitely happy to bikeshed the name. I considered several alternatives, but ended up settling here for (relative) brevity and syntactic proximity to the existing-and-known `run_if_changed`.

chaodaiG:
Not seeing any concern from my side either, it makes sense to be mutual exclusive with `run_if_changed`. Also for name, would it make sense to name it `skip_if_only_changed` or `skip_if_only_touched`?

alvaroaleman:
>  Also for name, would it make sense to name it skip_if_only_changed

Yeah, ended up with that one as well when thinking about it under the shower :)

2uasimojo:
> would it make sense to name it `skip_if_only_changed` or `skip_if_only_touched`?

I settled on `changed` to match up with `run_if_changed` (which might not be the greatest name itself, but that ship has sailed).

If I didn't care about brevity or sticking close to `run_if_changed`, I might choose something like `skip_if_all_changes_match`.

cjwagner:
+1 for `skip_if_only_changed`

2uasimojo:
- Renamed the option `skip_if_only_changed` per feedback.
- Fixed failing test case
- Added test cases

Re testing:
- I kinda went through and cloned/adapted _most of_ the tests for `run_if_changed`. It started to seem pretty redundant. But question to reviewers: Should I carry on and duplicate *all* of them, or do we feel covered as is?
- I would like to add a test case to validate mutual exclusivity with `always_run: true`, and between `run_if_changed` and `skip_if_only_changed`. So still `/hold`ing for that.

prow/config/config.go
```diff
@@ -2451,12 +2459,27 @@ func setBrancherRegexes(br Brancher) (Brancher, error) {
 }
 
 func setChangeRegexes(cm RegexpChangeMatcher) (RegexpChangeMatcher, error) {
-	if cm.RunIfChanged != "" {
-		re, err := regexp.Compile(cm.RunIfChanged)
-		if err != nil {
-			return cm, fmt.Errorf("could not compile run_if_changed regex: %v", err)
+	reprop := []struct {
```
alvaroaleman, line 2462: Nit: While I understand the desire of de-duplicating code, this is 21 LOC  and an additional indirection compared to 14 LOC it would take to have the check twice, once per field :)

2uasimojo:
> * add a test case to validate mutual exclusivity with `always_run: true`, and between `run_if_changed` and `skip_if_only_changed`.

Done.

Also added docs (in a separate commit, since I added content for `run_if_changed` as well).

Note that the actual code for this change is ~40LOC, including comments. It's the tests that blow it up. Would you like me to split the tests into one or more separate commits?

Otherwise I think this is ready to go.

prow/config/config.go
2uasimojo: I agree. LMK when done reviewing and I'll fix it up.

2uasimojo: I couldn't stand *not* de-duplicating, so I came up with this:

```go
	var reString, propName string
	if reString = cm.RunIfChanged; reString != "" {
		propName = "run_if_changed"
	} else if reString = cm.SkipIfOnlyChanged; reString != "" {
		propName = "skip_if_only_changed"
	}
	if reString != "" {
		re, err := regexp.Compile(reString)
		if err != nil {
			return cm, fmt.Errorf("could not compile %s regex: %v", propName, err)
		}
		cm.reChanges = re
	}
	return cm, nil
```

WDYT?

2uasimojo: Did this.

fejta:
/uncc 
/cc @chaodaiG 

Neat change, thanks!

alvaroaleman (approved):
/hold
In case others want to have a look

chaodaiG (approved):

PR #22503, opened 2021-06-09
Disable konnectivity on netpol jobs

knabben:
Netpol tests started to flake 2 days ago after enabling Konnectivity, this PR disables for Netpol

PR on k/k: https://github.com/kubernetes/kubernetes/pull/102661
https://testgrid.k8s.io/sig-network-gce#network-policies,%20google-gce

### DIFF
--- a/config/jobs/kubernetes/sig-network/sig-network-misc.yaml
+++ b/config/jobs/kubernetes/sig-network/sig-network-misc.yaml
@@ -109,6 +109,7 @@ presubmits:
         - --env=KUBE_NODE_OS_DISTRIBUTION=ubuntu
         - --env=KUBE_GCE_NODE_IMAGE=ubuntu-2004-focal-v20200423
         - --env=KUBE_GCE_NODE_PROJECT=ubuntu-os-cloud
+        - --env=KUBE_ENABLE_KONNECTIVITY_SERVICE=false
         - --ginkgo-parallel=15
         - --gcp-master-image=ubuntu
         - --gcp-node-image=ubuntu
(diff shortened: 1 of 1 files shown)


MrHohn:
Will defer to @Jefftree @cheftako for LGTM.
/approve

aojea:
@wojtek-t @liggitt @BenTheElder @spiffxp it is probably that the network policy job is more cpu intensive than other jobs, but if enabling the konnectivity server causes the job to start failing tests randomly, should we understand better why the konnectivity server causes this? It can be affecting other jobs too.

In addition, we recently had a lot of flakiness on some jobs due to some CSI components, i.e. one components was monitoring all pods, nodes, pvcs in all namespaces https://github.com/kubernetes/kubernetes/issues/102452#issuecomment-856854322 and disabling it solved the flakiness https://github.com/kubernetes/kubernetes/pull/102700

It is clear that CPU/IOPS/resource load affect CI health and cause flakiness, and it is very difficult to understand why jobs start to flake suddenly, since some of this component that causes the load are pulled from "external" repositories and there is no correlation with the git history, on the other side disabling components adhoc per job is not a good solution IMHO for the long term ...

I was thinking in that in short term we should:
- dump the list of kubernetes components used in the test and somehow report resource usage per component
- dump node resource usage (I don't know if we are doing this)

Thoughts?

aojea:
/lgtm
I was concerned about if the konnectivity server was going to be a fundamental component of kubernetes but Ben explained to me that is optional, so we can run without it

wojtek-t:
> dump the list of kubernetes components used in the test and somehow report resource usage per component

We already do that in scalability tests (maybe we're missing some components though...)

> dump node resource usage (I don't know if we are doing this)

We don't

The issue in both cases is that we report some statistics/percentiles. It's possible that particular percentiles look fine, but there are e.g. short spikes that will cause flakiness of your tests...

BenTheElder:
xref: https://github.com/kubernetes/kubernetes/issues/102904

PR #22509, opened 2021-06-10
Fix AWS CCM volume limit failure

olemarkus:
Use 1.22 so we don't have to explicitly enable EBS CSI and that we automatically filter the volume limit test

/cc @hakman @rifelpet


hakman (approved):
/retest

rifelpet (approved):

PR #22523, opened 2021-06-10
Switch WI bind script array expansion to a syntax compatible with bash versions < 4.4

cjwagner:
This syntax is a bit weird, but it is recommended [here](https://stackoverflow.com/questions/7577052/bash-empty-array-expansion-with-set-u) and seems to work with older and newer bash versions.
/assign @fejta

### DIFF
--- a/workload-identity/bind-service-accounts.sh
+++ b/workload-identity/bind-service-accounts.sh
@@ -66,7 +66,7 @@ members=($(
 
 want="serviceAccount:$project.svc.id.goog[$namespace/$name]"
 fix_policy=yes
-for member in "${members[@]}"; do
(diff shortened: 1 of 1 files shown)


cjwagner:
> I would suggest recommending upgrading to newer bash if possible, since they'll run into other problems somewhere else. If this is for mac it could be easily fixed by `brew install bash`

FWIW this was the only problem encountered on bash 3.2.57

fejta:
bash 4 came out in 2009

cjwagner:
Closing in favor of requiring bash >= 4.4: https://github.com/kubernetes/test-infra/pull/22544
/close

PR #22525, opened 2021-06-10
jobs: k8s-infra job changes

spiffxp:
Two changes here:
- drop the embedded bash for k8s-infra prow postsubmits now that we're using in-repo deploy scripts (added in https://github.com/kubernetes/k8s.io/pull/2153, fixed in https://github.com/kubernetes/k8s.io/pull/2189)
- make ci-k8sio-audit run twice as frequently (every 3h) to pick up changes sooner


BenTheElder (approved):
LGTM

PR #22527, opened 2021-06-10
chore: Increase Prow job timeout for Azure Disk CSI Driver E2E tests

edreed:
In addition to increasing the timeout on kubetest, we need to increase the Prow job timeout through the decoration_config.


andyzhangx (approved):
/lgtm
/approve

PR #22536, opened 2021-06-11
Revert "Adjust the timeout to 40m for scheduler-plugins integration test"

Huang-Wei:
This reverts PR #22212 as the timeout occurred in `go test` side, not prow.


Huang-Wei:
@BenTheElder Could you help lgtm? Thanks.

BenTheElder (approved):
lgtm

PR #22537, opened 2021-06-11
stop setting KUBE_TIMEOUT in kubernetes unit test jobs on the master …

BenTheElder:
…branch

https://github.com/kubernetes/kubernetes/pull/102768 enables this

xref: https://github.com/kubernetes/kubernetes/issues/102607

### DIFF
--- a/config/jobs/kubernetes/sig-testing/make-test.yaml
+++ b/config/jobs/kubernetes/sig-testing/make-test.yaml
@@ -17,12 +17,11 @@ presubmits:
       containers:
         - image: gcr.io/k8s-testimages/kubekins-e2e:v20210601-ea6aa4e-master
           command:
-            - runner.sh
-            - bash
-            - -c
-          args:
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes/sig-testing/make-test.yaml
```diff
@@ -17,12 +17,11 @@ presubmits:
       containers:
         - image: gcr.io/k8s-testimages/kubekins-e2e:v20210601-ea6aa4e-master
           command:
-            - runner.sh
-            - bash
-            - -c
-          args:
-            # TODO: stop depending on custom timeout
-            - 'make test KUBE_RACE=-race KUBE_TIMEOUT=--timeout=240s'
+            - make
```
BenTheElder, line 20: not strictly related, but normalizing how we specify the command vs the CI job and upcoming replacement rootless config

PR #22544, opened 2021-06-12
Make onboarding and WI scripts require bash >= 4.4.

cjwagner:
Some bash features like array expansion have changed since older versions of bash. We don't want to support such versions or have to debug issues that end up being caused by version skew.  Notably this will forbid the old version of bash that is installed on Macs by default.
/assign @fejta

### DIFF
--- a/prow/create-build-cluster.sh
+++ b/prow/create-build-cluster.sh
@@ -54,6 +54,16 @@ ADMIN_IAM_MEMBER="${ADMIN_IAM_MEMBER:-group:mdb.cloud-kubernetes-engprod-oncall@
 # Overriding output
 OUT_FILE="${OUT_FILE:-build-cluster-kubeconfig.yaml}"
 
+
+# Require bash version >= 4.4
+if ((${BASH_VERSINFO[0]}<4)) || ( ((${BASH_VERSINFO[0]}==4)) && ((${BASH_VERSINFO[1]}<4)) ); then
+  echo "ERROR: This script requires a minimum bash version of 4.4, but got version of ${BASH_VERSINFO[0]}.${BASH_VERSINFO[1]}"
+  if [ "$(uname)" = 'Darwin' ]; then
+    echo "On macOS with homebrew 'brew install bash' is sufficient."
+  fi
+  exit 1
+fi
+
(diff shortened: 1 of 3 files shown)


prow/create-build-cluster.sh
```diff
@@ -54,6 +54,19 @@ ADMIN_IAM_MEMBER="${ADMIN_IAM_MEMBER:-group:mdb.cloud-kubernetes-engprod-oncall@
 # Overriding output
 OUT_FILE="${OUT_FILE:-build-cluster-kubeconfig.yaml}"
 
+
+# Require bash version >= 4.4
+if [[ -z "${BASH_VERSINFO[@]}" ]]; then
```
BenTheElder, line 59: this should never happen in bash ™️

prow/create-build-cluster.sh
```diff
@@ -54,6 +54,19 @@ ADMIN_IAM_MEMBER="${ADMIN_IAM_MEMBER:-group:mdb.cloud-kubernetes-engprod-oncall@
...
+  echo "FAILURE: Couldn't determine bash version from \$BASH_VERSINFO. Please install bash version >= 4.4"
+  exit 1
+else
+  ver="${BASH_VERSINFO[0]}.${BASH_VERSINFO[1]}"
+  if (( $(echo "${ver} < 4.4" | bc -l) )); then
```
BenTheElder, line 64: we actually already have this implemented in kubernetes/kubernetes (though for 4.2)
https://github.com/kubernetes/kubernetes/blob/ea0764452222146c47ec826977f49d7001b0ea8c/hack/lib/util.sh#L725-L734

that implementation is pretty good, and I think the prompt to install with homebrew is helpful, FWIW.

prow/create-build-cluster.sh
cjwagner: I think I saw somewhere that it was added in bash 2.0, but that was ~25 years ago :upside_down_face:

prow/create-build-cluster.sh
cjwagner: Agreed, I used that instead.

fejta (approved):

PR #22545, opened 2021-06-13
Remove jeefy as a sig-ui lead

jeefy:
Stepping away from SIG-UI leadership :)


spiffxp (approved):
/approve
/lgtm
ref: https://github.com/kubernetes/community/pull/5828

PR #22547, opened 2021-06-14
testgrid: fix link for example widget

shamsher31:
Example widget points to https://testgrid.k8s.io/sig-testing-misc#ci-bazel


michelle192837:
Oh good catch, thanks! :D

/ok-to-test

michelle192837 (approved):

PR #22548, opened 2021-06-14
kops: Add e2e for cilium-etcd

olemarkus:
This got broken at some point. Adding to grid to understand what the situation is.


config/jobs/kubernetes/kops/build_jobs.py
```diff
@@ -968,11 +969,11 @@ def generate_presubmits_network_plugins():
     for plugin, run_if_changed in plugins.items():
         networking_arg = plugin
         skip_regex = skip_base
-        if plugin == 'cilium':
+        if plugin in ('cilium', 'cilium-etcd'):
             skip_regex += r'|should.set.TCP.CLOSE_WAIT'
         else:
             skip_regex += r'|Services.*functioning.*NodePort'
-        if plugin in ['calico', 'canal', 'weave', 'cilium']:
+        if plugin in ['calico', 'canal', 'weave', 'cilium', 'cilium-etcd']:
```
johngmyers, line 976: Neither of these two will have an effect as you didn't add `cilium-etcd` to the `plugins` hash.

olemarkus: I added this more for completeness. I dont think we nees the presubmit.

johngmyers: I think we should have a presubmit for (almost) each periodic, but that's a different change to the script.

johngmyers (approved):

PR #22553, opened 2021-06-14
Skip eviction tests on serial tests on systemd

odinuge:
Skip flaky eviction tests, as the other serial tests. The eviction tests
are moved to ci-kubernetes-node-kubelet-eviction and
pull-kubernetes-node-kubelet-eviction.

Follow up from https://github.com/kubernetes/test-infra/pull/22451


harche (approved):
/lgtm

ehashman (commented):
/lgtm

Am okay with this, but this means we won't have coverage for the eviction tests on cri-o. Are we also going to split them out for these jobs?

odinuge:
> Am okay with this, but this means we won't have coverage for the eviction tests on cri-o. Are we also going to split them out for these jobs?

Well yeah. But imo. it is better to start somewhere and add more test suites and configurations when they actually work. I don't think anyone actually benefit from permanently failing test suites. Also, the eviction tests are permanently failing with docker right now, so I think we should fix them first. (https://github.com/kubernetes/kubernetes/pull/102169 will at least make them work)

These are just pre-submits now, so when things stabilizes and things work, we can look at what to do next. Or?

ehashman:
> Well yeah. But imo. it is better to start somewhere and add more test suites and configurations when they actually work. I don't think anyone actually benefit from permanently failing test suites. Also, the eviction tests are permanently failing with docker right now, so I think we should fix them first. ([kubernetes/kubernetes#102169](https://github.com/kubernetes/kubernetes/pull/102169) will at least make them work)
> 
> These are just pre-submits now, so when things stabilizes and things work, we can look at what to do next. Or?

The reason I ask is because when I removed the selector from the node-serial jobs for these broken tests, I split the eviction tests into a separate job. Wasn't sure if we wanted to do the same here, but I'm okay not doing so.

mrunalp (approved):

PR #22555, opened 2021-06-14
Enable the configuration sharding for approve plugin

droslean:

### DIFF
--- a/prow/plugins/config.go
+++ b/prow/plugins/config.go
@@ -1771,7 +1771,7 @@ type Override struct {
 
 func (c *Configuration) mergeFrom(other *Configuration) error {
 	var errs []error
-	if diff := cmp.Diff(other, &Configuration{Plugins: other.Plugins, Bugzilla: other.Bugzilla}); diff != "" {
+	if diff := cmp.Diff(other, &Configuration{Plugins: other.Plugins, Bugzilla: other.Bugzilla, Approve: other.Approve}); diff != "" {
 		errs = append(errs, fmt.Errorf("supplemental plugin configuration has config that doesn't support merging: %s", diff))
 	}
 
@@ -1786,7 +1786,13 @@ func (c *Configuration) mergeFrom(other *Configuration) error {
 		errs = append(errs, fmt.Errorf("failed to merge .bugzilla from supplemental config: %w", err))
 	}
 
+	c.mergeApproveFrom(other.Approve)
+
 	return utilerrors.NewAggregate(errs)
+}
+
+func (c *Configuration) mergeApproveFrom(other []Approve) {
+	c.Approve = append(c.Approve, other...)
 }
 
 func (p *Plugins) mergeFrom(other *Plugins) error {
@@ -1856,7 +1862,7 @@ func (p *Bugzilla) mergeFrom(other *Bugzilla) error {
 }
 
 func (c *Configuration) HasConfigFor() (global bool, orgs sets.String, repos sets.String) {
-	if !reflect.DeepEqual(c, &Configuration{Plugins: c.Plugins, Bugzilla: c.Bugzilla}) || c.Bugzilla.Default != nil {
+	if !reflect.DeepEqual(c, &Configuration{Plugins: c.Plugins, Bugzilla: c.Bugzilla, Approve: c.Approve}) || c.Bugzilla.Default != nil {
 		global = true
 	}
 	orgs = sets.String{}
@@ -1875,6 +1881,16 @@ func (c *Configuration) HasConfigFor() (global bool, orgs sets.String, repos set
 		}
 		for repo := range orgConfig.Repos {
 			repos.Insert(org + "/" + repo)
+		}
+	}
+
+	for _, approveConfig := range c.Approve {
+		for _, orgOrRepo := range approveConfig.Repos {
+			if strings.Contains(orgOrRepo, "/") {
+				repos.Insert(orgOrRepo)
+			} else {
+				orgs.Insert(orgOrRepo)
+			}
 		}
 	}
 
--- a/prow/plugins/config_test.go
+++ b/prow/plugins/config_test.go
@@ -1995,6 +1995,7 @@ func TestHasConfigFor(t *testing.T) {
 			resultGenerator: func(fuzzedConfig *Configuration) (toCheck *Configuration, expectGlobal bool, expectOrgs sets.String, expectRepos sets.String) {
 				fuzzedConfig.Plugins = nil
 				fuzzedConfig.Bugzilla = Bugzilla{}
+				fuzzedConfig.Approve = nil
 				return fuzzedConfig, !reflect.DeepEqual(fuzzedConfig, &Configuration{}), nil, nil
 			},
 		},
@@ -2003,6 +2004,7 @@ func TestHasConfigFor(t *testing.T) {
 			resultGenerator: func(fuzzedConfig *Configuration) (toCheck *Configuration, expectGlobal bool, expectOrgs sets.String, expectRepos sets.String) {
 				// exclude non-plugins configs to test plugins specifically
 				fuzzedConfig.Bugzilla = Bugzilla{}
+				fuzzedConfig.Approve = nil
 				expectOrgs, expectRepos = sets.String{}, sets.String{}
 				for orgOrRepo := range fuzzedConfig.Plugins {
 					if strings.Contains(orgOrRepo, "/") {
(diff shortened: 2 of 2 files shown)


prow/plugins/config.go
```diff
@@ -1786,9 +1786,58 @@ func (c *Configuration) mergeFrom(other *Configuration) error {
...
+		return nil
+	}
+
+	// If the supplemental configuration points to a specific org/repo, we need to validate that there is
+	// only one approve configuration that holds that org/repo value in the `repos` field.
+	if org != "" && repo != "" {
+		if len(other) > 1 {
+			return fmt.Errorf("supplemental config for %s/%s must have only 1 approve plugin configuration", org, repo)
+		}
+		for _, approve := range other {
```
stevekuznetsov, line 1807: We just validated that the `other` slice has one entry, this loop was confusing to read - we should just validate the first member of it

prow/plugins/config.go
```diff
@@ -1786,9 +1786,58 @@ func (c *Configuration) mergeFrom(other *Configuration) error {
...
+			if len(approve.Repos) != 1 || (len(approve.Repos) == 1 && approve.Repos[0] != fmt.Sprintf("%s/%s", org, repo)) {
```
stevekuznetsov, line 1808: Couldn't this be

```go
if other[0].Repos != []string{fmt.Sprintf("%s/%s", org, repo)} {
    return fmt.Errorf("the repos field must hold the %s/%s value only", org, repo)
}
````

File: prow/plugins/config.go [hunk 3]
```diff
@@ -1786,9 +1786,58 @@ func (c *Configuration) mergeFrom(other *Configuration) error {
...
+				return fmt.Errorf("the repos field must hold the %s/%s value only", org, repo)
+			}
+		}
+	}
+
+	// If the supplemental configuration points to an organization, we need to validate that each of the configurations holds
```
stevekuznetsov, line 1814: Hm - values for `orgA/repoB` should not be in the `orgA` file - they should be in the repo-specific file. We should instead validate that an org-level file has *only* org level configuration

prow/plugins/config.go
```diff
@@ -1786,9 +1786,58 @@ func (c *Configuration) mergeFrom(other *Configuration) error {
...
+			return utilerrors.NewAggregate(errs)
+		}
+	}
+
+	if len(c.Approve) == 0 {
+		c.Approve = other
+		return nil
+	}
+
+	c.Approve = append(c.Approve, other...)
```
stevekuznetsov, line 1836: I don't think this is the correct implementation. Since the `[]Approve` slice is order-dependent, we must make sure that any org-level configuration comes *later* in this list than any repo-level configurations inside of that org, that way the repo-level configurations override the org-level ones.

prow/plugins/config_test.go
```diff
@@ -1547,7 +1547,7 @@ func TestConfigMergingProperties(t *testing.T) {
 			name: "Merging into empty config always succeeds and makes the empty config equal to the one that was merged in",
 			verification: func(t *testing.T, fuzzedMergeableConfig *Configuration) {
 				newConfig := &Configuration{}
-				if err := newConfig.mergeFrom(fuzzedMergeableConfig); err != nil {
```
stevekuznetsov, line 1550: This test needs new cases for org-only configs

prow/plugins/plugins.go
```diff
@@ -291,6 +291,25 @@ func (pa *ConfigAgent) Load(path string, supplementalPluginConfigDirs []string,
 				return nil
 			}
 
+			relPath, err := filepath.Rel(supplementalPluginConfigDir, path)
```
stevekuznetsov, line 294: This won't work in a configmap

prow/plugins/config.go
droslean: What about adding the ability to have an org-level config that affects only specific repos that belong in that org?

prow/plugins/config.go
droslean: well, we validate that for repo-level config this slice will have only 1 value. It can have multiple values only for org-level configs.

prow/plugins/plugins.go
droslean: The file will be a symbolic link to `..data/...`. We will still be able to read the path without search for the original link since we are not resolving it. Also, the above code ignores `..` prefixes for this reason.

prow/plugins/config.go
alvaroaleman: > What about adding the ability to have an org-level config that affects only specific repos that belong in that org?

That makes it repo specific and not org-specific anymore so has to be in repo config files

droslean: Thanks for the clarification. I'll make the changes.

prow/plugins/config.go
droslean: In this case, we will need to re-order the configuration after we merge all of them.

droslean: Well, it seems that we already do that. We search first for a configuration for the specific repo and if we won't find it we search again for the org only. See: https://github.com/kubernetes/test-infra/blob/master/prow/plugins/config.go#L751-L770

prow/plugins/plugins.go
```diff
@@ -291,6 +291,25 @@ func (pa *ConfigAgent) Load(path string, supplementalPluginConfigDirs []string,
...
+			if err != nil {
```
alvaroaleman, line 295: Please remove all the changes here. Both the supplemental file loading is (here) and the org repo directory structure validation (checkconfig) are already implemented.

prow/plugins/config.go
```diff
@@ -1769,9 +1769,9 @@ type Override struct {
 	AllowedGitHubTeams map[string][]string `json:"allowed_github_teams,omitempty"`
 }
 
-func (c *Configuration) mergeFrom(other *Configuration) error {
+func (c *Configuration) mergeFrom(other *Configuration, org, repo string) error {
```
alvaroaleman, line 1772: This is wrong, please remove it. You need to extend `HasConfigFor` instead.

prow/plugins/config.go
```diff
@@ -1786,9 +1786,50 @@ func (c *Configuration) mergeFrom(other *Configuration) error {
...
+		}
+		if !reflect.DeepEqual(other[0].Repos, []string{fmt.Sprintf("%s/%s", org, repo)}) {
+			return fmt.Errorf("the repos field must hold the %s/%s value only", org, repo)
+		}
+	}
+
+	// If the supplemental configuration points to an organization, we need to validate that each of the configurations holds
+	// values in the repo field that belong to that organization.
+	if org != "" && repo == "" {
+		if len(other) > 1 {
```
alvaroaleman, line 1815: You must not validate anything like that here. The only thing you must validate here is that `other` can be merged. I suppose that is always the case as long as there isn't already a config for the some org or repo.

The checking what orgs and repos are allowed in what file happens in checkonfig through HasConfigFor() which is distinct from the merge code.

prow/plugins/config_test.go
```diff
@@ -2031,6 +2034,27 @@ func TestHasConfigFor(t *testing.T) {
 				return fuzzedConfig, !reflect.DeepEqual(fuzzedConfig, &Configuration{Bugzilla: fuzzedConfig.Bugzilla}), expectOrgs, expectRepos
 			},
 		},
+		{
+			name: "Any config with approve is considered to be for the orgs and repos references there",
+			resultGenerator: func(fuzzedConfig *Configuration) (toCheck *Configuration, expectGlobal bool, expectOrgs sets.String, expectRepos sets.String) {
+				// exclude non-plugins configs to test bugzilla specifically
+				fuzzedConfig.Plugins = nil
+				fuzzedConfig.Bugzilla = Bugzilla{}
```
alvaroaleman, line 2042:
```suggestion
				fuzzedConfig = &Configuration{Approve: fuzzedConfig.Approve}
```

prow/plugins/config_test.go
```diff
@@ -2040,26 +2064,97 @@ func TestHasConfigFor(t *testing.T) {
 
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
-			for i := 0; i < 100; i++ {
-				fuzzedConfig := &Configuration{}
-				fuzzer.Fuzz(fuzzedConfig)
+			// for i := 0; i < 100; i++ {
```
alvaroaleman, line 2067: Please keep that :)

droslean: oops... debuging leftovers :)

prow/plugins/config.go
```diff
@@ -1786,9 +1786,27 @@ func (c *Configuration) mergeFrom(other *Configuration) error {
...
 	}
 
+	if err := c.mergeApproveFrom(other.Approve); err != nil {
+		errs = append(errs, fmt.Errorf("failed to merge .approve from supplemental config: %w", err))
+	}
+
 	return utilerrors.NewAggregate(errs)
 }
 
+func (c *Configuration) mergeApproveFrom(other []Approve) error {
```
alvaroaleman, line 1796: This never returns an error and just making it `append` without the two additional conditions might be simpler :)

alvaroaleman (approved):

PR #22566, opened 2021-06-15
Fix gerrit onboarder bugs

mpherman2:
Three bugs in the gerrit onboarder needed to be fixed.
1) If the repo name contained a slash, the workdir would not have the slash. So cloud-k8s-engprod/test-infra needed git clone ...cloud-k8s-engprod/test-infra, but the workdir would just be test-infra.

2) If the All-Projects repo is not accessible due to Host restriction settings we now just assume that the All-Projects config does not contain anything and continue instead of failing. (This is a best attempt)

3) If adding to the last section in the project.config there would be a random empty line in the middle of the file after addition. We now ignore whitespace when generating project.config map.

### DIFF
--- a/experiment/gerrit-onboarder/onboarder.go
+++ b/experiment/gerrit-onboarder/onboarder.go
@@ -193,7 +193,7 @@ func configToMap(configFile string) (map[string][]string, []string) {
 		if strings.HasPrefix(line, "[") {
 			curKey = line
 			orderedKeys = append(orderedKeys, line)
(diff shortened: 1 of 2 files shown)


chaodaiG (approved):

BenTheElder:
https://github.com/kubernetes/test-infra/tree/master/experiment/gerrit-onboarder should probably get an OWNERS with reviewers familiar with the project.

PR #22578, opened 2021-06-16
Fix empty dir handling in phaino

dhiller:
If there's an `emptyDir` present in the volumes section, the function
`resolveVolumeMounts` puts an empty string in the map. This restores
the handling of that case. Otherwise errors occur like these:

```
docker: Error response from daemon: failed to create shim: OCI runtime
create failed: invalid mount {Destination::/mount-name Type:bind
Source:/var/lib/docker/volumes/a187...adfe/_data Options:[rbind]}:
mount destination :/mount-name not absolute: unknown.
```

Can also be seen when phaino is run in the log:
```
...
"-v" \
 ":/mount-name" \
...
```


cjwagner (approved):

dhiller:
@fgimenez @rmohr FYI

PR #22570, opened 2021-06-15
Update snyk secretRef name per DNS-1123

navidshaikh:

PR #22583, opened 2021-06-16
Update snyk periodic job

navidshaikh:
- quiet snyk download
 - check if SNYK_TOKEN envvar is present before running snyk
 - check if 'snyk test' command passed before filtering results
- changed the interval to 1h to check if job is running correctly

/cc @dims @spiffxp

### DIFF
--- a/config/jobs/kubernetes/wg-k8s-infra/trusted/wg-k8s-infra-trusted.yaml
+++ b/config/jobs/kubernetes/wg-k8s-infra/trusted/wg-k8s-infra-trusted.yaml
@@ -89,9 +89,11 @@ periodics:
       - |
         set -euo pipefail; \
         apt update && apt -y install jq; \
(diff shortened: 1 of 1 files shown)


navidshaikh:
followup to https://github.com/kubernetes/test-infra/pull/22570

navidshaikh:
changed the interval to 1h to check if job is running correctly, I'll revert it back to 6h once its green

navidshaikh:
> changed the interval to 1h to check if job is running correctly, I'll revert it back to 6h once its green

realised I didnt push the commit to change it to 1h,  :-/ 
beside, the job failed again with 
```
Invalid character in header content ["authorization"]
```
seems like this is due to special characters in the snyk token, I've updated the secrets with quote and that should be picked up for next run

adding a follow up to change the interval to 1h 
cc: @dims

dims:
thanks @navidshaikh

PR #22584, opened 2021-06-16
Update exit code verification and change snyk job interval

navidshaikh:
- Update snyk job interval to 1h, this needs to be reverted after
   successful run to 6h

 - also update 'snyk test' command's exit code verification as
   snyk test returns exit code
     0: success, no vulns found
     1: action_needed, vulns found
     2: failure, try to re-run command
     3: failure, no supported projects detected
   now checking exit code -gt 1 for error case

PR #22595, opened 2021-06-17
Debug snyk continuous job

navidshaikh:
- A dummy token is set to be used which is not going to auth the snyk
   CLI
 - Now printing the SNYK_TOKEN in logs to verify if the dummy token set
   in gsm is being reflected in the job as is
 - Change the job interval to 3h


navidshaikh:
followup to https://github.com/kubernetes/test-infra/pull/22584

/cc @dims @spiffxp

spiffxp:
Oh I bet you have SNYK_TOKEN= in the secret value. You should get rid of that if it's there. The ExternalSecret specifies that

navidshaikh:
>Oh I bet you have SNYK_TOKEN= in the secret value. You should get rid of that if it's there. The ExternalSecret specifies that

@spiffxp true

I changed it as you suggested and the job is green now https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/ci-kubernetes-snyk-master/1405575732968558592

/close

on to next for reverting the job interval

PR #22597, opened 2021-06-17
Set the snyk job interval to 6h

navidshaikh:
also removes debug statements

/assign @dims 
/cc @spiffxp

### DIFF
--- a/config/jobs/kubernetes/wg-k8s-infra/trusted/wg-k8s-infra-trusted.yaml
+++ b/config/jobs/kubernetes/wg-k8s-infra/trusted/wg-k8s-infra-trusted.yaml
@@ -67,7 +67,7 @@ periodics:
 #   filtered from the snyk scan results and printed on stdout as well
 #   as "${ARTIFACTS}/snyk_results.json"
 - name: ci-kubernetes-snyk-master
-  interval: 1h
+  interval: 6h
   cluster: k8s-infra-prow-build-trusted
   decorate: true
   extra_refs:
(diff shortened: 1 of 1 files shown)


navidshaikh:
follow up to https://github.com/kubernetes/test-infra/pull/22595

config/jobs/kubernetes/wg-k8s-infra/trusted/wg-k8s-infra-trusted.yaml
```diff
@@ -92,10 +92,8 @@ periodics:
         wget -q -O /usr/local/bin/snyk https://github.com/snyk/snyk/releases/download/v1.605.0/snyk-linux && chmod +x /usr/local/bin/snyk; \
         mkdir -p "${ARTIFACTS}"; \
         [[ -z "${SNYK_TOKEN}" ]] && echo "SNYK_TOKEN env var is not set, required for snyk scan" && exit 1; \
-        echo "$PWD"; \
-        echo "SNYK_TOKEN is present"; \
-        snyk test --json-file-output=${ARTIFACTS}/result_all.json || [[ $? -gt 1 ]] && echo "failed to run snyk scan" && exit 1; \
-        cat ${ARTIFACTS}/result_all.json | jq '.vulnerabilities | .[] | select ((.type=="license") or (.version=="0.0.0") | not)' | tee "${ARTIFACTS}/snyk_results.json"
+        snyk test --json > ${ARTIFACTS}/result_unfiltered.json || [[ $? -gt 1 ]] && echo "failed to run snyk scan" && exit 1; \
+        cat ${ARTIFACTS}/result_unfiltered.json | jq '.vulnerabilities | .[] | select ((.type=="license") or (.version=="0.0.0") | not)' | tee "${ARTIFACTS}/result.json"
```
spiffxp, line 96: Style nits:
- since you're using `-|` you don't need to pretend like everything is a single line (you could do this for everything but I'm just editing the change)
- since we don't need to be multiline, you can use `if`
- I assumed you meant `$? -gt 0`, but if 1 is a valid exit code for `snyk` then this isn't quite right
- avoid `cat foo` when `<foo` will do

personal jq nits:
- I prefer to avoid unpacking arrays, because iterating one-json-per-line (which this isn't quite) is a bit of a pain.  As one example, I can't just `jq length` the output of your script to see how many vulnerabilities there are.  So, prefer `map` to `.[] |`
- some yaml parsers dislike bare lists (were this to ever find its way into yaml), so prefer to output with a root key (I arbitrarily chose to re-use `vulnerabilities`)
- to my eyes the select filter reads easier with `!= and !=` vs `== or == | not`, I'm still a bit unclear on how much of that expression the `not` applies to

```suggestion
        if ! snyk test --json > ${ARTIFACTS}/result_unfiltered.json; then
          echo "failed to run snyk scan"
          exit 1
        fi
        <${ARTIFACTS}/result_unfiltered.json jq \
         '{vulnerabilities: .vulnerabilities | map(select((.type != "license") and (.version != "0.0.0"))) }' \
        | tee "${ARTIFACTS}/result.json"
```

spiffxp (commented):
Post merge review since I took waaaaay too long to confirm my `jq` comments

Ideally this becomes `hack/verify-snyk.sh` or something in kubernetes/kubernetes for easier local testing, though I suppose in terms of time-to-merge, iterating here is faster

config/jobs/kubernetes/wg-k8s-infra/trusted/wg-k8s-infra-trusted.yaml
spiffxp: https://github.com/kubernetes/kubernetes/blob/master/hack/verify-e2e-test-ownership.sh for an example of the kind of json munging I've had to do with jq in the past that led me to these preferences

navidshaikh: >I assumed you meant $? -gt 0, but if 1 is a valid exit code for snyk then this isn't quite right

snyk exits with 1 if it found the vulnerabilities
```
       0: success, no vulns found
       1: action_needed, vulns found
       2: failure, try to re-run command
       3: failure, no supported projects detected
```

re: style / jq nits

thanks for insightful review! ❤️  that.
I'll work on the next iteration to incorporate the feedback.

PR #22599, opened 2021-06-17
images/gcb-docker-gcloud: enable docker-buildx builds

LorbusChris:
These changes make the gcb-docker-gcloud image usable as a docker-buildx builder, like so (example from #22444):

```yaml
steps:
  - name: gcr.io/k8s-staging-test-infra/gcb-docker-gcloud
    entrypoint: /buildx-entrypoint
    args:
    - build
    - --tag=gcr.io/$PROJECT_ID/alpine:$_GIT_TAG
    - --tag=gcr.io/$PROJECT_ID/alpine:latest
    - --platform=linux/amd64,linux/arm64/v8
    - --build-arg=IMAGE_ARG=gcr.io/$PROJECT_ID/alpine:$_GIT_TAG
    - --push
    - .
    dir: .
substitutions:
  _GIT_TAG: '12345'

```

/cc @spiffxp

### DIFF
--- a/images/gcb-docker-gcloud/cloudbuild.yaml
+++ b/images/gcb-docker-gcloud/cloudbuild.yaml
@@ -7,8 +7,14 @@ steps:
       - --build-arg=GO_VERSION=$_GO_VERSION
       - .
     dir: images/gcb-docker-gcloud/
+  - name: gcr.io/cloud-builders/docker
+    args:
+    - tag
+    - gcr.io/$PROJECT_ID/gcb-docker-gcloud:$_GIT_TAG
+    - gcr.io/$PROJECT_ID/gcb-docker-gcloud:latest
 substitutions:
   _GIT_TAG: '12345'
   _GO_VERSION: 1.16.2
 images:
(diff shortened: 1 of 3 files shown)


images/gcb-docker-gcloud/cloudbuild.yaml
```diff
@@ -7,8 +7,14 @@ steps:
       - --build-arg=GO_VERSION=$_GO_VERSION
       - .
     dir: images/gcb-docker-gcloud/
```
LorbusChris, line 9: Is this intentional?
I had do use `dir: .` for it work with
```
builder  --project=k8s-staging-test-infra --scratch-bucket=gs://k8s-staging-test-infra-gcb images/gcb-docker-gcloud/
```

spiffxp: I think it doesn't need to be this way but may require changes I don't have time to dig into right now. Better to follow the same convention as everything else

spiffxp (approved):
/approve
/lgtm

PR #22600, opened 2021-06-17
Crier slack reporter: allow no default slack token

chaodaiG:
We have a prow instance that itself doesn't need to report to slack, but tenants of prow need to do so, and potentially on different slacks. In this case it makes more sense to allow not setting up a default slack token, rather than faking a token

### DIFF
--- a/prow/cmd/crier/main.go
+++ b/prow/cmd/crier/main.go
@@ -111,8 +111,8 @@ func (o *options) validate() error {
 	}
 
 	if o.slackWorkers > 0 {
-		if o.slackTokenFile == "" {
-			return errors.New("--slack-token-file must be set")
+		if o.slackTokenFile == "" && len(o.additionalSlackTokenFiles) == 0 {
+			return errors.New("one of --slack-token-file or --additional-slack-token-files must be set")
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


prow/cmd/crier/main.go
```diff
@@ -237,18 +237,21 @@ func main() {
 		slackConfig := func(refs *prowapi.Refs) config.SlackReporter {
 			return cfg().SlackReporterConfigs.GetSlackReporter(refs)
 		}
-		if err := secretAgent.Add(o.slackTokenFile); err != nil {
-			logrus.WithError(err).Fatal("could not read slack token")
+		var tokensMap map[string]func() []byte
```
alvaroaleman, line 240: You need to initialize the map before you can write to it

cjwagner (approved):
LGTM
/hold

prow/cmd/crier/main.go
cjwagner: Nice catch.

chaodaiG: done

alvaroaleman (approved):

fejta (approved):

PR #22601, opened 2021-06-17
fix the image config path

Dragoncell:
/bug


bobbypage:
Image config is here: https://github.com/kubernetes/test-infra/blob/master/jobs/e2e_node/containerd/containerd-master/cgroupv2/image-config-cgroupv2.yaml

PR #22450, opened 2021-06-07
images/git: Build for arm64

LorbusChris:
This commit adds a Google Cloud Build definition for building git
arm64 images and pushing them as a manifest list together with their
amd64 counterparts.

This is a prereq for #16588 and #22362

~~Blocked on: https://github.com/kubernetes/test-infra/pull/22599  https://github.com/kubernetes/test-infra/pull/22444~~

/cc @stevekuznetsov

PR #22609, opened 2021-06-18
WIP: config/jobs: Use gcb-docker-gcloud image for building alpine and git images

LorbusChris:
#22444 has broken the alpine postsubmit job.
In #22450 the git image is supposed to get the same treatment, so I'm changing the base for both jobs here already.

I'm not sure fixing it is as easy as this. PTAL.

Edit: Nope, this won't work

/cc @BenTheElder 
/cc @fejta

### DIFF
--- a/config/jobs/kubernetes/test-infra/test-infra-trusted.yaml
+++ b/config/jobs/kubernetes/test-infra/test-infra-trusted.yaml
@@ -44,7 +44,7 @@ postsubmits:
     spec:
       serviceAccountName: pusher
       containers:
-      - image: gcr.io/k8s-testimages/gcloud-bazel:v20210128-v0.6-11-gf1c6399
+      - image: gcr.io/k8s-testimages/gcb-docker-gcloud:v20210617-6d302b2
         command:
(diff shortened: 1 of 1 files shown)


LorbusChris:
/hold
this doesn't work

fejta (approved):
Would expect this to get autobumped tomorrow (monday) but this will get things working now.

LorbusChris:
I don't think this here is the right approach.

Maybe the postsubmit job hit https://github.com/docker/buildx/issues/584, which might be solved by simply re-trying?

LorbusChris:
Also, for completeness, here's the prow postsubmit failure: https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-test-infra-push-alpine/1405964567472771072

fejta:
https://prow.k8s.io/?job=post-test-infra-push-alpine will have a rerun button (although maybe you can't click it so I did). Should be starting now

fejta:
Rerun seems to have passed?

LorbusChris:
excellent, thank you!
/close

PR #22613, opened 2021-06-19
cleanup check-dependency-stats script

dims:
I wanted to add this to the diff and ended up reformatting things a bit as well.
```
       -u  -U NUM  --unified[=NUM]
              Output NUM (default 3) lines of unified context.
```

### DIFF
--- a/config/jobs/kubernetes/sig-arch/kubernetes-depstat-periodical.yaml
+++ b/config/jobs/kubernetes/sig-arch/kubernetes-depstat-periodical.yaml
@@ -17,13 +17,20 @@ periodics:
         args:
         - -c
         - |
-          set -euo; \
-          export PATH=$PATH:$GOPATH/bin; \
-          mkdir -p "${ARTIFACTS}"; \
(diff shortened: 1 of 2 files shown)


config/jobs/kubernetes/sig-arch/kubernetes-depstat-periodical.yaml
```diff
@@ -17,13 +17,20 @@ periodics:
...
         - -c
         - |
-          set -euo; \
-          export PATH=$PATH:$GOPATH/bin; \
-          mkdir -p "${ARTIFACTS}"; \
-          pushd "$ARTIFACTS"; \
-          go install github.com/kubernetes-sigs/depstat@latest; \
-          popd; \
-          depstat stats --json | tee "${ARTIFACTS}/stats-base.json";
+          set -euo pipefail
```
BenTheElder, line 20: test-infra needs a linter for "size of inline prow bash" 😂 

I think this might not be over the limit yet but .. it's growing 👀

BenTheElder (approved):

PR #22614, opened 2021-06-19
test images: forward signal to test command and ensure it is backgrou…

BenTheElder:
…nded so signals are processed

note: after bootstrap is pushed, we will then need to bump it as the base in kubekins-e2e


amwat (commented):
/lgtm

PR #22615, opened 2021-06-19
drop KUBE_RACE from unit test CI definition

BenTheElder:
this is defaulted in the make target now

follow up to https://github.com/kubernetes/kubernetes/pull/102960


BenTheElder:
> +++ [0619 19:07:55] Running tests without code coverage and with -race

https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/102922/pull-kubernetes-unit/1406327127774597120

> Decorated Command: make test

PR #22620, opened 2021-06-20
fix broken depstat periodic job

dims:
pushed a wrong change in https://github.com/kubernetes/test-infra/commit/6435f39d6cf0885ad0e58909ddb9ef3e7b986c8c to the periodic job. we don't have/need a comparison, we just log the json.


BenTheElder (approved):

BenTheElder:
lgtm

PR #22631, opened 2021-06-21
GHProxy: Add metrics about inode usage

alvaroaleman:
This change adds metrics about the inode usage to ghproxy. The
background is that ghproxy recently silently stopped to work because
there were no inodes left on its disk, which resulted in a hard to debug
"Somehow ghproxy doesn't cache anymore" situation.

The metrics in here allow to alert on that.

### DIFF
--- a/greenhouse/diskutil/diskutil.go
+++ b/greenhouse/diskutil/diskutil.go
@@ -26,16 +26,19 @@ import (
 )
 
 // GetDiskUsage wraps syscall.Statfs for usage in GCing the disk
-func GetDiskUsage(path string) (percentBlocksFree float64, bytesFree, bytesUsed uint64, err error) {
+func GetDiskUsage(path string) (percentBlocksFree float64, bytesFree, bytesUsed uint64, percentInodeFree float64, inodeFree, inodeUsed uint64, err error) {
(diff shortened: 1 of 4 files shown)


greenhouse/diskutil/diskutil.go
```diff
@@ -26,16 +26,19 @@ import (
...
-		return 0, 0, 0, err
+		return 0, 0, 0, 0, 0, 0, err
 	}
 	percentBlocksFree = float64(stat.Bfree) / float64(stat.Blocks) * 100
 	bytesFree = stat.Bfree * uint64(stat.Bsize)
 	bytesUsed = (stat.Blocks - stat.Bfree) * uint64(stat.Bsize)
-	return percentBlocksFree, bytesFree, bytesUsed, nil
+
+	percentInodeFree = float64(stat.Ffree) / float64(stat.Files) * 100
+	inodeUsed = stat.Files - stat.Ffree
```
alvaroaleman, line 40: For reference, as the field names in statfs are IMHO not super self-explanatory: https://man7.org/linux/man-pages/man2/statfs.2.html

stevekuznetsov (approved):
LGTM - with the access pattern of this tool (many small files) this is a good idea

PR #22635, opened 2021-06-21
Generic-autobumper: avoid downloading cached oncall.json

chaodaiG:


chaodaiG:
This was fixed from the other end
/close

PR #22637, opened 2021-06-21
Lgtm config: Add sharding support

alvaroaleman:


prow/plugins/config_test.go
```diff
@@ -2086,61 +2104,46 @@ func TestHasConfigFor(t *testing.T) {
 	}
 }
 
-func TestApproveMergeFrom(t *testing.T) {
```
alvaroaleman, line 2089: The different testcaes didn't really test anything different (all of them: We merge one additional config into main config) so I have added a `TestMergeFrom` that tests both that and the lgtm merging

cjwagner (approved):

PR #22639, opened 2021-06-21
kOps - Focus on conformance for IPv6 test

hakman:
/cc @rifelpet @olemarkus


rifelpet:
Why do we need to focus on conformance for this? just curious

hakman:
There's a lot of noise from sig-storage due to external AWS EBS CSI driver
I think for now it makes sense to only focus on getting conformance passing.

olemarkus:
If there are actual failures with CSI Driver, I'd like to have a version with the full suite. Until recently, e2e failed due to a missing permission, but that should have been fixed today.

hakman:
@olemarkus https://testgrid.k8s.io/kops-misc#kops-grid-scenario-ipv6  if you want to take a look.
Other than that, this should be about networking and IPv6 stuff. Not sure why AWS EBS CSI is mandatory in general, but that is more philosophical.

olemarkus (approved):

PR #22640, opened 2021-06-21
cvelist-public: grab the validation script from the validate branch

tallclair:
I want to move the `validate-k8s.sh` script out of the `master` branch, so that the master branch tracks the upstream (https://github.com/CVEProject/cvelist), and we can open PRs against the upstream directly from our fork.

I check for absence  of the script as a hack to allow testing changes to the script against the `validate` branch itself. Open to suggestions for a better way to do this.

/committee product-security
/cc @dims @nikhita


dims:
@tallclair works for me!

PR #22643, opened 2021-06-21
Crier slack reporter: support no default slack settings

chaodaiG:
https://github.com/kubernetes/test-infra/pull/22600 allows no default slack token, which still fails crier deployment because default prow config missing. This PR allows no default slack settings

### DIFF
--- a/prow/crier/reporters/slack/reporter.go
+++ b/prow/crier/reporters/slack/reporter.go
@@ -130,6 +130,9 @@ func (sr *slackReporter) GetName() string {
 	return reporterName
 }
 
+// ShouldReport return true if:
+// - job state matches defined job types to be reported
+// - job type matches global config or slack channel defined in job config
 func (sr *slackReporter) ShouldReport(_ context.Context, logger *logrus.Entry, pj *v1.ProwJob) bool {
 	jobCfg := jobConfig(pj)
 	prowCfg := sr.getConfig(pj)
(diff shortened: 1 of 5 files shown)


prow/crier/reporters/slack/reporter.go
```diff
@@ -75,11 +80,16 @@ func channel(prowCfg config.SlackReporter, jobCfg *v1.SlackReporterConfig) (stri
 	return host, channel
```
cjwagner, line 80: I think we should either emit an error is `channel` is not specified or set it to a default value like `host` if there is a reasonable default.

prow/crier/reporters/slack/reporter.go
```diff
@@ -75,11 +80,16 @@ func channel(prowCfg config.SlackReporter, jobCfg *v1.SlackReporterConfig) (stri
...
+	if prowCfg != nil {
+		res = prowCfg.ReportTemplate
+	}
+	// Prefer config in job
 	if jobCfg != nil && jobCfg.ReportTemplate != "" {
-		return jobCfg.ReportTemplate
+		res = jobCfg.ReportTemplate
 	}
-	return prowCfg.ReportTemplate
+	return res
```
cjwagner, line 92: This also seems to hide errors due to missing config.

cjwagner (commented):
It looks like this might allow errors due to missing configuration to be silently ignored or lead to other errors with less useful error messages.

prow/crier/reporters/slack/reporter.go
chaodaiG: I don't think that would happen. Added comments on `ShouldReport` below, which returns true only if:
- global config is defined
- (OR) slack config in job is defined and slack channel is defined in job

The first case isn't covered by this PR, but not covered currently either
The second case wouldn't result in empty channel

prow/crier/reporters/slack/reporter.go
chaodaiG: ditto, not a bug introduced in this PR

chaodaiG (commented):
These don't seem like problems introduced in this PR, would be happy to address them separately in following up PR though

cjwagner:
> These don't seem like problems introduced in this PR, would be happy to address them separately in following up PR though

My understanding is that the problems were not introduced in this PR, but were previously only problematic if a global slack config existed, but was removed while the binary was running. Now the binary can be started without a global slack config and encounter problems immediately which is far more likely. Is this accurate?

If so I don't think we should try to use slack reporting without a global config until this is resolved. If you'd like to keep the PRs separate for organizational purposes that seems fine, but I don't think we should use this change without additional precautions. Ideally we would identify this configuration error at config load time so that `checkconfig` will reject job config changes that would lead to reporting problems.

cjwagner (approved):
/hold

chaodaiG:
> My understanding is that the problems were not introduced in this PR, but were previously only problematic if a global slack config existed, but was removed while the binary was running. Now the binary can be started without a global slack config and encounter problems immediately which is far more likely. Is this accurate?

This is true! I didn't think about this before but yeah you're right. tbh this is PR is not critical as I can always define global config with empty list. Dropping this PR for now, will revisit if there is a better idea

/close

PR #22645, opened 2021-06-21
Replace /bin/bash with /usr/bin/env bash

chaodaiG:
There is a mixed use of /bin/bash (38) and /usr/bin/env bash (62) in this repo, consolidate to be more mac friendly


BenTheElder (approved):

BenTheElder:
lgtm

PR #22656, opened 2021-06-22
Fix: Use kubekins image instead of minikube image

nearora-msft:


andyzhangx (commented):
/ok-to-test

andyzhangx (approved):
/lgtm
/approve

PR #22646, opened 2021-06-22
Update DNS presubmit to build and test images.

prameshj:
Config copied over from travis - https://github.com/kubernetes/dns/blob/master/.travis.yml, skipping ginkgo test for now.

PR #22659, opened 2021-06-22
Execute only `make test` as part of DNS presubmit.

prameshj:
If this works, I can make a script in k/dns repo with all the required steps and trigger that from the prow job. 
This is a follow-up to https://github.com/kubernetes/test-infra/pull/22646


MrHohn (approved):
/lgtm
/approve

PR #22664, opened 2021-06-23
Add the ability to shard the external plugins configuration

droslean:
/cc @alvaroaleman 

### DIFF
--- a/prow/plugins/config.go
+++ b/prow/plugins/config.go
@@ -1771,7 +1771,7 @@ type Override struct {
 
 func (c *Configuration) mergeFrom(other *Configuration) error {
 	var errs []error
-	if diff := cmp.Diff(other, &Configuration{Approve: other.Approve, Bugzilla: other.Bugzilla, Lgtm: other.Lgtm, Plugins: other.Plugins}); diff != "" {
+	if diff := cmp.Diff(other, &Configuration{Approve: other.Approve, Bugzilla: other.Bugzilla, ExternalPlugins: other.ExternalPlugins, Lgtm: other.Lgtm, Plugins: other.Plugins}); diff != "" {
 		errs = append(errs, fmt.Errorf("supplemental plugin configuration has config that doesn't support merging: %s", diff))
 	}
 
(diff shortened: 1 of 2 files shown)


prow/plugins/config.go
```diff
@@ -1789,6 +1789,23 @@ func (c *Configuration) mergeFrom(other *Configuration) error {
...
+}
+
+func (c *Configuration) mergeExternalPluginsFrom(other map[string][]ExternalPlugin) error {
+	var errs []error
+	for orgOrRepo, config := range other {
+		if _, ok := c.ExternalPlugins[orgOrRepo]; ok {
+			errs = append(errs, fmt.Errorf("found duplicate config for external-plugins.%s", orgOrRepo))
+			continue
+		}
+		c.ExternalPlugins[orgOrRepo] = config
```
alvaroaleman, line 1806: You need to make sure `c.ExternalPlugins` is non-nil before writing to it

prow/plugins/config_test.go
```diff
@@ -2147,3 +2163,57 @@ func TestMergeFrom(t *testing.T) {
 		}
 	}
 }
+
+func TestMergeExternalPluginsFrom(t *testing.T) {
```
alvaroaleman, line 2167: Could you extend `TestMergeFrom` instead of writing a new test so we don't duplicate so much testcode?

prow/plugins/config.go
```diff
@@ -1789,6 +1789,25 @@ func (c *Configuration) mergeFrom(other *Configuration) error {
...
+}
+
+func (c *Configuration) mergeExternalPluginsFrom(other map[string][]ExternalPlugin) error {
+	var errs []error
+	for orgOrRepo, config := range other {
+		if _, ok := c.ExternalPlugins[orgOrRepo]; ok {
+			errs = append(errs, fmt.Errorf("found duplicate config for external-plugins.%s", orgOrRepo))
+			continue
+		}
+		if c.ExternalPlugins != nil {
```
alvaroaleman, line 1806: We want to always merge the config, not only when the main config already has config. You need to initialize the map when its nil, not gate the writing to the map by a nilcheck. Please also add a testcase for this (main config has no `ExtternalPlugins` config, supplemental config has, it gets merged)

alvaroaleman (approved):

Issue #22380, opened 2021-06-01
Inconsistency between periodic jobs and presubmits in sig-scalability jobs


marseel:
Currently we have 1 CI  release blocking job that runs 100-nodes cluster scalability test:
https://github.com/kubernetes/test-infra/blob/612c22333fdc2229e87fedc3d4b628820225a464/config/jobs/kubernetes/sig-scalability/sig-scalability-release-blocking-jobs.yaml#L138

and 2 presubmits - one for kubernetes/kubernetes repository, one for kubernetes/perf-tests repository:
k/k:
https://github.com/kubernetes/test-infra/blob/612c22333fdc2229e87fedc3d4b628820225a464/config/jobs/kubernetes/sig-scalability/sig-scalability-presubmit-jobs.yaml#L2
k/perf-tests:
https://github.com/kubernetes/test-infra/blob/612c22333fdc2229e87fedc3d4b628820225a464/config/jobs/kubernetes/sig-scalability/sig-scalability-presubmit-jobs.yaml#L466

Unfortunately, there are some differences:
Config of CI job contains these flags:
```
 - --env=CL2_ENABLE_DNS_PROGRAMMING=true      - --env=CL2_SCHEDULER_THROUGHPUT_THRESHOLD=0      - --env=CL2_ENABLE_HUGE_SERVICES=true      - --env=CL2_ENABLE_API_AVAILABILITY_MEASUREMENT=true      - --env=CL2_API_AVAILABILITY_PERCENTAGE_THRESHOLD=99.5
```
But they are not part of presubmits to kubernetes/perf-tests and k/k, which might cause some regression.

I propose:
1. Enabling these additional flags (and veryfying if there are not other differences) for kubernetes/perf-tests presubmit
2. Verify that it's working correctly
3. Enable flags for kubernetes/kubernetes presubmit.

PR #22666, opened 2021-06-23
Add missing flags to perf-tests clusterloader2 presubmit

tosi3k:
/assign @marseel 

Part of https://github.com/kubernetes/test-infra/issues/22380.


marseel:
I've verified that both flags are present in CI jobs and CI jobs are consistently green.
/lgtm

PR #22667, opened 2021-06-23
allow branchprotector to protect branches matching a certain regex

### LINKED ISSUE #22657: allow branchprotector to protect branches matching a certain regex

There only seems to be an exclude option - https://github.com/kubernetes/test-infra/blob/1bd2a72cbe81fde4066a0e96a3ac3d75027bc624/prow/config/branch_protection.go#L53-L55

It would be ideal if we could supply a 'include' regex

**Why is this needed**:

As we introduce automated tooling or in the case someone pushes branch by accident to a repo they become protected and it becomes cumbersome to delete those.

ie. ideally we want to only protected `main` and `release-*` branches - so our inclusion regex would be `release-.*|main`

mcbenjemaa:
Add Include policy setting for branchprotector to allow protect certain branches.
Include can be emtpy and it has no effect. on current config.

Fixes #22657


prow/cmd/branchprotector/protect.go
```diff
@@ -337,7 +345,12 @@ func (p *protector) UpdateRepo(orgName string, repoName string, repo config.Repo
 		}
 		for _, b := range bs {
 			_, ok := repo.Branches[b.Name]
-			if !ok && branchExclusions != nil && branchExclusions.MatchString(b.Name) {
+			if !ok && branchInclusions != nil && branchInclusions.MatchString(b.Name) {
+				branches[b.Name] = b
+			} else if !ok && branchInclusions != nil && !branchInclusions.MatchString(b.Name) {
+				logrus.Infof("%s/%s=%s: excluded", orgName, repoName, b.Name)
+				continue
+			} else if !ok && branchInclusions == nil && branchExclusions != nil && branchExclusions.MatchString(b.Name) {
```
alvaroaleman, line 353: We know for sure at this point that `branchInclusions` is nil

```suggestion
			} else if !ok && branchExclusions != nil && branchExclusions.MatchString(b.Name) {
```

prow/cmd/branchprotector/protect.go
```diff
@@ -321,6 +321,14 @@ func (p *protector) UpdateRepo(orgName string, repoName string, repo config.Repo
 		return nil
 	}
 
+	var branchInclusions *regexp.Regexp
```
alvaroaleman, line 324: Please add tests for this new functionality

prow/config/branch_protection.go
```diff
@@ -53,6 +53,9 @@ type Policy struct {
 	// Exclude specifies a set of regular expressions which identify branches
 	// that should be excluded from the protection policy
 	Exclude []string `json:"exclude,omitempty"`
+	// Include specifies a set of regular expressions which identify branches
+	// that should be included from the protection policy
+	Include []string `json:"include,omitempty"`
```
alvaroaleman, line 58: Please document that this is mutually exclusive with `Exclude` and add a check into the config load that errors if both are set

prow/cmd/branchprotector/protect.go
```diff
@@ -337,7 +345,12 @@ func (p *protector) UpdateRepo(orgName string, repoName string, repo config.Repo
 		}
 		for _, b := range bs {
 			_, ok := repo.Branches[b.Name]
-			if !ok && branchExclusions != nil && branchExclusions.MatchString(b.Name) {
+			if !ok && branchInclusions != nil && branchInclusions.MatchString(b.Name) {
+				branches[b.Name] = b
+			} else if !ok && branchInclusions != nil && !branchInclusions.MatchString(b.Name) {
+				logrus.Infof("%s/%s=%s: excluded", orgName, repoName, b.Name)
```
alvaroaleman, line 351:
```suggestion
				logrus.Infof("%s/%s=%s: not included", orgName, repoName, b.Name)
```

prow/cmd/branchprotector/protect.go
```diff
@@ -337,7 +345,12 @@ func (p *protector) UpdateRepo(orgName string, repoName string, repo config.Repo
 		}
 		for _, b := range bs {
 			_, ok := repo.Branches[b.Name]
-			if !ok && branchExclusions != nil && branchExclusions.MatchString(b.Name) {
+			if !ok && branchInclusions != nil && branchInclusions.MatchString(b.Name) {
+				branches[b.Name] = b
+			} else if !ok && branchInclusions != nil && !branchInclusions.MatchString(b.Name) {
+				logrus.Infof("%s/%s=%s: not included", orgName, repoName, b.Name)
+				continue
+			} else if !ok && branchExclusions != nil && branchExclusions.MatchString(b.Name) {
```
alvaroaleman, line 353: Branchprotector respecting branchInclusions needs tests

prow/config/branch_protection.go
```diff
@@ -53,6 +53,9 @@ type Policy struct {
 	// Exclude specifies a set of regular expressions which identify branches
 	// that should be excluded from the protection policy
 	Exclude []string `json:"exclude,omitempty"`
```
alvaroaleman, line 55: Please also doc on `Exclude` that it is mutually exclusive with `Include`

prow/cmd/branchprotector/protect.go
mcbenjemaa: I have added a test case in the `TestApply`

prow/config/branch_protection_test.go
```diff
@@ -289,6 +289,18 @@ func TestApply(test *testing.T) {
 				Exclude: []string{"bar*", "foo*"},
 			},
 		},
+		{
+			name: "merge inclusion strings",
```
alvaroaleman, line 293: @mcbenjemaa this test seems to fail

mcbenjemaa:
Ready

alvaroaleman (approved):

PR #22668, opened 2021-06-23
Revert to a no-op presubmit for DNS

prameshj:
A followup will invoke a script instead of putting commands in the job spec directly.

Details in https://github.com/kubernetes/dns/issues/458


aojea:
/lgtm
/approve
these things take several attempts to get it right :)

MrHohn (approved):
/lgtm
/approve

PR #22670, opened 2021-06-23
Invoke presubmit script from prow job.

prameshj:
Invoking the newly added script. 

I think invoking directly should work. Alternative is to invoke via bash -c:

```
          - "bash"
          - "--"
          - "-c"
          - "presubmit.sh"
```

/assign @aojea @MrHohn


MrHohn (approved):
/lgtm
/approve

PR #22679, opened 2021-06-23
prow/plank/README: Fix default_decoration_configs -> default_decoration_config_entries

wking:
Catching up with 1acac76284 (#20656), which transitioned the content from a map to an entry.  Maybe there's some code I'm not finding that will automatically determine if `default_decoration_configs` is a map or slice and load it without these changes?  CC @cjwagner


cjwagner (approved):
Looks like an oversight, thanks for submitting a fix!

PR #22680, opened 2021-06-23
mark DNS presubmit as optional correctly.

prameshj:
Previous PR - https://github.com/kubernetes/test-infra/pull/22678 only made it run manually, but it was still required.

/assign @MrHohn


MrHohn (approved):
/lgtm
/approve

PR #22681, opened 2021-06-23
migrate DNS presubmit to podutils.

prameshj:
Used the guide in https://gist.github.com/dims/c1296f8ed42238baea0a5fcae45f4cf4.

bootstrap.py approach is deprecated. podutils also has the abillity to configure a job to be optional.

/assign @MrHohn @aojea


MrHohn (approved):
/lgtm
/approve

PR #22684, opened 2021-06-24
Reduce error level of first attempt to comment to Gerrit.

cjwagner:
Currently we get a lot of error spam from this log line even though the second attempt to comment (without the label) succeeds and indicates to the user that Prow doesn't haven't sufficient permissions to use the label. This problem is not actionable to Prow maintainers, but it is to the users or their repo admins so lets just reduce the log level here.
/assign @mpherman2 @chaodaiG

### DIFF
--- a/prow/crier/reporters/gerrit/reporter.go
+++ b/prow/crier/reporters/gerrit/reporter.go
@@ -291,7 +291,7 @@ func (c *Client) Report(ctx context.Context, logger *logrus.Entry, pj *v1.ProwJo
 
 	logger.Infof("Reporting to instance %s on id %s with message %s", gerritInstance, gerritID, message)
 	if err := c.gc.SetReview(gerritInstance, gerritID, gerritRevision, message, reviewLabels); err != nil {
-		logger.WithError(err).Errorf("fail to set review with label %q on change ID %s", reportLabel, gerritID)
+		logger.WithError(err).Infof("fail to set review with label %q on change ID %s", reportLabel, gerritID)
 
 		if reportLabel == "" {
 			return nil, nil, err


prow/crier/reporters/gerrit/reporter.go
```diff
@@ -291,7 +291,7 @@ func (c *Client) Report(ctx context.Context, logger *logrus.Entry, pj *v1.ProwJo
 
 	logger.Infof("Reporting to instance %s on id %s with message %s", gerritInstance, gerritID, message)
 	if err := c.gc.SetReview(gerritInstance, gerritID, gerritRevision, message, reviewLabels); err != nil {
-		logger.WithError(err).Errorf("fail to set review with label %q on change ID %s", reportLabel, gerritID)
+		logger.WithError(err).Infof("fail to set review with label %q on change ID %s", reportLabel, gerritID)
```
chaodaiG, line 294: there might be failures other than lack of permission, can we restrict this to be based on error message?

cjwagner: Are you talking about the case when `reportLabel` is not set? If there is a report label we try to comment again without the label and log any problems at the error level.
I generally prefer to avoid matching against an expected error message since we have no guarantees the match is complete or that the messages won't change format. If this is just about the case when `reportLabel` is unset I can change the behavior to make this an error in that case and avoid matching against an error message. WDYT?

chaodaiG: I care less about `reportLabel` not set, more the original purpose of setting this log level as `Error`, which imo is because we should be informed when this started happening a lot when it's not user error

cjwagner: > we should be informed when this started happening a lot when it's not user error

I totally agree, but doesn't this solution provide that?
The only difference between the first and second calls to the Gerrit API is the removal of the labels in the second call. If the problem is not related to setting the label the second attempt to comment will also fail and that is still logged at the error level. Since label settings are entirely out of our control I think it makes sense to avoid error logs whether they are caused by insufficient permissions or a non-existent label.

chaodaiG: good point!

chaodaiG (approved):

PR #22688, opened 2021-06-24
bump 30 minutes the gce job with docker runtime

aojea:
This job has exactly the same configuration that the one that uses containerd as runtime, however, it seems that is timing out continuously.

Bumping the timeout 30 mins in the meantime, sig-release should decide if they want to keep monitoring.
Another options is if someone wants to get deeper into these job issues to fully understand the causes of the timeouts, but since dockershim has already been deprecated I don't think this is going to happen :smile: 

Fixes: https://github.com/kubernetes/kubernetes/issues/103097


BenTheElder (approved):
lgtm

PR #22691, opened 2021-06-24
Update test image for e2e-framework

ShwethaKumbla:
#### What this PR does 
Update test image to kubekins which has docker and kind tool.

/cc @vladimirvivien @cpanato

### DIFF
--- a/config/jobs/kubernetes-sigs/e2e-framework/e2e-framework-presubmits.yaml
+++ b/config/jobs/kubernetes-sigs/e2e-framework/e2e-framework-presubmits.yaml
@@ -22,11 +22,15 @@ presubmits:
     path_alias: sigs.k8s.io/e2e-framework
     spec:
       containers:
-      - image: gcr.io/k8s-staging-releng/releng-ci:latest-go1.16
+      - image: gcr.io/k8s-testimages/kubekins-e2e:v20210615-c973edd-master
         imagePullPolicy: Always
         command:
+        - runner.sh
+        args:
         - make
         - test
+        securityContext:
+          privileged: true
     annotations:
       testgrid-dashboards: sig-testing-e2e-framework
       testgrid-tab-name: pull-e2e-framework-test


config/jobs/kubernetes-sigs/e2e-framework/e2e-framework-presubmits.yaml
```diff
@@ -6,7 +6,7 @@ presubmits:
     path_alias: sigs.k8s.io/e2e-framework
     spec:
       containers:
-      - image: gcr.io/k8s-staging-releng/releng-ci:latest-go1.16
+      - image: gcr.io/k8s-testimages/kubekins-e2e:v20210615-c973edd-master
```
cpanato, line 9: did you test if that will work?

vladimirvivien: @cpanato is there an easy way to test the image outside of prawl ?
cc @ShwethaKumbla

cpanato: here: https://github.com/kubernetes/test-infra/blob/master/prow/build_test_update.md#How-to-test-a-ProwJob

cpanato: we can merge this and try out as well :)

ShwethaKumbla: No. I am not sure how to test this changes in local. If you can suggest me how can I verify this change that would be great.

We need a kind cluster to test our code (Unit tests). I tried installing kind but it has dependency with docker and in the current images docker is not configured.  I observed most of the other projects use `kubekins` image so I updated the image with kubekins.

cpanato: hum, then maybe we need to setup the privileged to true for this job, can you please check other jobs that uses this image to see how they are setup?

ShwethaKumbla: Yes, I see other jobs which use kubekins configured with privileged to true and command as runner.sh

          securityContext:
            privileged: true

          command:
              # generic runner script, handles DIND, bazelrc for caching, etc.
              - runner.sh

config/jobs/kubernetes-sigs/e2e-framework/e2e-framework-presubmits.yaml
```diff
@@ -22,11 +25,14 @@ presubmits:
...
         imagePullPolicy: Always
         command:
+        - runner.sh
```
cpanato, line 31: I think should be something like
```
- command:
   - runner.sh
    args:
    - make
    - test
```

config/jobs/kubernetes-sigs/e2e-framework/e2e-framework-presubmits.yaml
```diff
@@ -6,11 +6,15 @@ presubmits:
...
```
cpanato, line 12: for this job we can keep the previous image, this is running lint and other checks, we don't need to have special things for this

cpanato (approved):
/lgtm

vladimirvivien (approved):
/approve

PR #22694, opened 2021-06-24
Include preset-dind-enabled label in DNS presubmit

prameshj:
/assign @aojea @MrHohn


aojea:
/lgtm
/approve
this is the good one

PR #22701, opened 2021-06-25
Phaino use grace and timeout from job

dhiller:
When executing a prow job using phaino, use the values for grace period and timeout from the prow job spec. If there is nothing configured it uses the option values as a fallback.

### DIFF
--- a/prow/cmd/phaino/local.go
+++ b/prow/cmd/phaino/local.go
@@ -554,7 +554,7 @@ func getTimeout(optionsTimeout time.Duration, prowJobTimeout time.Duration) time
 	if prowJobTimeout > 0 {
 		return prowJobTimeout
 	}
-	return optionsTimeout
+	return defaultTimeout
 }
 
 func getMinimumGracePeriod(minimum time.Duration, optionsGracePeriod time.Duration, prowJobGracePeriod time.Duration, log *logrus.Entry) (gracePeriod time.Duration) {
--- a/prow/cmd/phaino/local.go
+++ b/prow/cmd/phaino/local.go
@@ -499,10 +499,10 @@ func (opts *options) convertJob(ctx context.Context, log *logrus.Entry, pj prowa
 		return nil
 	}
 	log.Info("Starting job...")
-	// TODO(fejta): default grace and timeout to the job's decoration_config
-	if opts.timeout > 0 {
+	timeout := getTimeout(pj, opts)
+	if timeout > 0 {
 		var cancel func()
-		ctx, cancel = context.WithTimeout(ctx, opts.timeout)
+		ctx, cancel = context.WithTimeout(ctx, timeout)
 		defer cancel()
 	}
 	cmd, err := start(args)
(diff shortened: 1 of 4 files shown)


dhiller:
@alvaroaleman @michelle192837 can you please take a look? Thanks!

prow/cmd/phaino/local.go
```diff
@@ -550,3 +546,24 @@ func (opts *options) convertJob(ctx context.Context, log *logrus.Entry, pj prowa
...
+	if pj.Spec.DecorationConfig.Timeout.Duration > 0 {
+		return pj.Spec.DecorationConfig.Timeout.Duration
+	}
+	if opts.timeout > 0 {
+		return opts.timeout
+	}
+	return time.Duration(0)
+}
+
+func getMinimumGracePeriod(minimum time.Duration, pj prowapi.ProwJob, opts *options, log *logrus.Entry) time.Duration {
```
chaodaiG, line 560: This function only needs `pj.Spec.DecorationConfig.GracePeriod.Duration` and `opts.grace` instead of `pj` and `grace`, please limit to those so that it's easier to write unit test. And speaking of unit test, please add them as well

prow/cmd/phaino/local.go
```diff
@@ -550,3 +546,24 @@ func (opts *options) convertJob(ctx context.Context, log *logrus.Entry, pj prowa
...
+	if pj.Spec.DecorationConfig.GracePeriod.Duration >= minimum {
+		return pj.Spec.DecorationConfig.GracePeriod.Duration
+	}
+	if opts.grace >= minimum {
```
chaodaiG, line 564: my feeling is if the user passed `--grace` as argument, they want to override the spec from pj, so maybe do this check first?

prow/cmd/phaino/local.go
```diff
@@ -550,3 +546,24 @@ func (opts *options) convertJob(ctx context.Context, log *logrus.Entry, pj prowa
 	}
 	return fmt.Errorf("grace period expired, aborted: %v", ctx.Err())
 }
+
+func getTimeout(pj prowapi.ProwJob, opts *options) time.Duration {
```
chaodaiG, line 550: Same comment as for the next function

dhiller:
@chaodaiG thank you for your review! I've addressed the comments, PTAL!

prow/cmd/phaino/local.go
```diff
@@ -550,3 +546,25 @@ func (opts *options) convertJob(ctx context.Context, log *logrus.Entry, pj prowa
...
 }
+
+func getTimeout(optionsTimeout time.Duration, prowJobTimeout time.Duration) time.Duration {
+	if optionsTimeout != defaultTimeout {
+		return optionsTimeout
+	}
+	if prowJobTimeout > 0 {
+		return prowJobTimeout
+	}
+	return optionsTimeout
```
chaodaiG, line 557: nit: logically this is correct as optionsTimeout is equal to defaultTimeout at this point, but might not be as easy to understand than just returning defaultTimeout

chaodaiG (approved):
LGTM, thank you for the improvement!
/hold
in case you want to address my last nit

prow/cmd/phaino/local.go
dhiller: Thanks for the catch, I agree that returning `defaultTimeout` will make it more clear.

chaodaiG:
https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/test-infra/22701/pull-test-infra-bazel/1410599192383983616
Failed several linting tests, you can run:
```
./hack/update-bazel.sh
./hack/update-deps.sh
./hack/update-gofmt.sh
```

chaodaiG (approved):
/unhold

chaodaiG:
sounds like flake
/retest

PR #22707, opened 2021-06-25
Run Kubernetes master branch unit tests as non-root unprivileged user

BenTheElder:
Fixes https://github.com/kubernetes/kubernetes/issues/99881 _finally_!

Thanks @PushkarJ for fixing the last problematic unit test 🎉

### DIFF
--- a/config/jobs/kubernetes/sig-testing/make-test.yaml
+++ b/config/jobs/kubernetes/sig-testing/make-test.yaml
@@ -14,6 +14,10 @@ presubmits:
     always_run: true
     path_alias: k8s.io/kubernetes
     spec:
+      # unit tests have no business requiring root or doing privileged operations
+      securityContext:
+        runAsUser: 2000
+        allowPrivilegeEscalation: false
       containers:
         - image: gcr.io/k8s-testimages/kubekins-e2e:v20210615-c973edd-master
           command:
@@ -42,7 +46,7 @@ presubmits:
     always_run: false
     path_alias: k8s.io/kubernetes
     spec:
-      # let's see if we can run unit tests as a non-root user
+      # unit tests have no business requiring root or doing privileged operations
       securityContext:
         runAsUser: 2000
         allowPrivilegeEscalation: false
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes/sig-testing/make-test.yaml
```diff
@@ -14,6 +14,10 @@ presubmits:
     always_run: true
     path_alias: k8s.io/kubernetes
     spec:
+      # unit tests have no buisiness requiring root or doing privileged operations
```
BenTheElder, line 17: all of these comments misspelled business

BenTheElder: fixed.

config/jobs/kubernetes/sig-testing/make-test.yaml
```diff
@@ -14,6 +14,10 @@ presubmits:
     always_run: true
     path_alias: k8s.io/kubernetes
     spec:
+      # unit tests have no business requiring root or doing privileged operations
+      securityContext:
+        runAsUser: 2000
```
alvaroaleman, line 19: Does that user exist? Tests tend to do stuff in $HOME which if inexistent results in writing to / which wont work

BenTheElder: It doesn't, and they don't, unit tests should write to a tempdir and not the real $HOME.

We already had a test version of the job (`pull-kubernetes-unit-experimental`) that passes with this config.

config/jobs/kubernetes/sig-testing/make-test.yaml
```diff
@@ -14,6 +14,10 @@ presubmits:
...
+        allowPrivilegeEscalation: false
```
vinayakankugoyal, line 20: Should we drop all capabilities and also add `runAsGroup: 2000`?

BenTheElder: probably the right call, we should start in the -experimental job first though and make sure there isn't something depending on this (and if so fix it before rolling it out to everyone's PRs)

BenTheElder: closing the loop https://github.com/kubernetes/test-infra/pull/22780

PR #22708, opened 2021-06-25
Explicit empty http cookiefile config can override default

chaodaiG:
This is needed in cases where majority of org/repo/cluster requires http cookiefile set, instead of explicitly add them in all, providing a way for the exceptions to be able to override default with empty string


chaodaiG:
cc @krzyzacy

cjwagner (approved):

cjwagner (approved):

PR #22711, opened 2021-06-26
Add a rerun button to Spyglass

### LINKED ISSUE #14226: Add a rerun button to Spyglass.
## What would you like to be added:
We can now directly rerun ProwJobs from Prow's front page using GitHub oauth. This is feature is access through the rerun dialog that is exposed by clicking on one of the circular arrows on the left side of the page. See for yourself at https://prow.k8s.io

It would be super useful to expose this feature from the spyglass page since that is where job results are viewed and is almost certainly the most common way that devs use Deck.

## Why is this needed:

Currently there isn't an easy way to navigate from a spyglass page to the same job run on Decks front page. (This is a problem of its own, but it really only impacts the usability of the rerun button at this point.)  Devs need to use the filters on the front page to find matching jobs and even then there may be multiple runs of the job on the same commit that are not distinguishable from the UI.

@Katharine I think you are probably most familiar with this code. Do you think there are any 'gotchas' here regarding the oauth redirection?

cc @stevekuznetsov @fejta @alvaroaleman 
/help

YuviGold:
Resolves #14226

The `createRerunProwJobIcon` creating a rerun button is moved
from the deck module into the common package.

Setting the color of the text in `rerun-content` box as gray to
preserve consistency since it is currently handled on higher levels.

Add `handleRerunButton` function when loading spyglass page.
The function would call `createRerunProwJobIcon` same as the deck page.
The rerun button won't be shown in case the prowJob is unavailable.

Add `ReRunCreateJob` boolean option and `ProwJobName` to the
spyglass template to be used for the rerun command.

Rename `lensesTemplate` struct to `spyglassTemplate` since it includes more
than just lens data.

Screenshot from 2021-06-26 21-58-26

/cc @cjwagner @alvaroaleman

### DIFF
--- a/prow/cmd/deck/static/spyglass/spyglass.ts
+++ b/prow/cmd/deck/static/spyglass/spyglass.ts
@@ -1,9 +1,13 @@
+import {createRerunProwJobIcon} from "../common/common";
+import {getParameterByName} from "../common/urls";
 import {isTransitMessage, serialiseHashes} from "./common";
 
 declare const src: string;
 declare const lensArtifacts: {[index: string]: string[]};
 declare const lensIndexes: number[];
 declare const csrfToken: string;
+declare const rerunCreatesJob: boolean;
+declare const prowJobName: string;
 
 // Loads views for this job
 function loadLenses(): void {
@@ -160,4 +164,32 @@ window.addEventListener('hashchange', (e) => {
 // We can't use DOMContentLoaded here or we end up with a bunch of flickering. This appears to be MDL's fault.
 window.addEventListener('load', () => {
     loadLenses();
+    handleRerunButton();
 });
+
+function handleRerunButton() {
+  // In case prowJob is unavailable, the rerun button shouldn't be shown
+  if (!prowJobName) {
+    return;
+  }
+
+  const rerunStatus = getParameterByName("rerun");
+  const modal = document.getElementById('rerun')!;
+  const rerunCommand = document.getElementById('rerun-content')!;
+  window.onclick = (event: any) => {
+      if (event.target === modal) {
+          modal.style.display = "none";
+      }
+  };
+
+  const r = document.getElementById("header-title")!;
+  const c = document.createElement("div");
+  c.appendChild(createRerunProwJobIcon(modal, rerunCommand, prowJobName, rerunCreatesJob, csrfToken));
+  c.classList.add("icon-cell");
+  r.appendChild(c);
+
+  if (rerunStatus === "gh_redirect") {
+    modal.style.display = "block";
+    rerunCommand.innerHTML = "Rerunning that job requires GitHub login. Now that you're logged in, try again";
+  }
+}
--- a/prow/cmd/deck/template/spyglass.html
+++ b/prow/cmd/deck/template/spyglass.html
@@ -5,6 +5,8 @@
   var src = {{.Source}};
   var lensArtifacts = {{.LensArtifacts}};
   var lensIndexes = {{.LensIndexes}};
+  var rerunCreatesJob = {{.ReRunCreatesJob}};
+  var prowJobName = {{.ProwJobName}};
 </script>
 <script type="text/javascript" src="/static/spyglass_bundle.min.js"></script>
 <link rel="stylesheet" type="text/css" href="/static/spyglass/spyglass.css">
@@ -47,3 +49,7 @@
 {{end}}
 
 {{template "page" (settings mobileUnfriendly darkMode "spyglass" .)}}
+
+<div id="rerun">
+  <div id="rerun-content"></div>
+</div>
--- a/prow/cmd/deck/main.go
+++ b/prow/cmd/deck/main.go
@@ -1081,37 +1081,41 @@ lensesLoop:
 	}
 
 	var viewBuf bytes.Buffer
-	type lensesTemplate struct {
-		Lenses        map[int]spyglass.LensConfig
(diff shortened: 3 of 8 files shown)


prow/cmd/deck/template/spyglass.html
```diff
@@ -47,3 +49,7 @@
 {{end}}
 
 {{template "page" (settings mobileUnfriendly darkMode "spyglass" .)}}
+
+<div id="rerun">
+  <div id="rerun-content"></div>
```
cjwagner, line 54: Rather than always adding a rerun element at the bottom of the page, please conditionally add the rerun element as part of the link header defined on lines 23-35. We should make the rerun link more visible and only show up if there is still a ProwJob resource available to rerun. Notice that spyglass can return an empty string ProwJob name without erroring: https://github.com/kubernetes/test-infra/blob/f7e21a3c18f4f4bbc7ee170675ed53e4544a0632/prow/spyglass/spyglass.go#L298-L299
This is because spyglass can display job results from GCS even after the ProwJob k8s resource has been garbage collected. In this case we can't rerun since we don't have a ProwJob resource. (Usually a copy of the ProwJob resource is stored with the job results in GCS so we could actually use that to rerun, but if we want to add that lets do it in a separate PR.)

prow/cmd/deck/static/spyglass/spyglass.ts
```diff
@@ -160,4 +164,27 @@ window.addEventListener('hashchange', (e) => {
...
+  const rerunStatus = getParameterByName("rerun");
+  const modal = document.getElementById('rerun')!;
+  const rerunCommand = document.getElementById('rerun-content')!;
+  window.onclick = (event: any) => {
+      if (event.target === modal) {
+          modal.style.display = "none";
+      }
+  };
+
+  const r = document.getElementsByClassName("mdl-layout__header-row")[0];
```
cjwagner, line 180: Please make this more robust to unrelated changes by looking up the element in a more stable way, e.g. by ID.

cjwagner (changes requested):
Thanks this looks good, I just have a few requests.

cjwagner:
To help with review in the future please consider using separate commits for renaming/moving code and functional changes.

prow/cmd/deck/template/spyglass.html
YuviGold: Please let me know if I understood it right before fixing it:
1. Move the rerun button icon from the top of the page (next to the title) to be part of the links header.
I tried that at first with the icon, but it looked bad, so I thought it would look nicer at the top. Do you prefer only a "Rerun" title?
2. Make the rerun link appear conditionally in case the ProwJob resource is available.
I used the same implementation as the deck view. It shows the button and the popup with the command copy button unconditionally but there's an additional **Rerun** button that appears only in case _rerunCreatesJob_ boolean option is true.
So, do you want the button link not to be shown at all or did you refer only to the internal rerun button? Do you want the same behavior to be aligned with the deck view or does spyglass is expected to behave differently?

prow/cmd/deck/static/spyglass/spyglass.ts
YuviGold: I will add an id attribute to the header element. :+1:

prow/cmd/deck/template/spyglass.html
cjwagner: 1. I didn't realize the icon is currently near the title, that position is fine. I thought the button would appear at the bottom of the page because it is not defined in the `title` definition on line 1. Sounds like that is not the case for some reason?
2. Spyglass pages have an important distinction from the main deck page: The main page is entirely based off of ProwJob resource data. This data is garbage collected after a short amount of time (~24 hours on high traffic instances). So the main page only shows jobs that have run recently enough to still have ProwJob resources in the cluster. Spyglass pages are different in that they also consume job result data from GCS which persists for months or years. Since the ProwJob resource is needed to rerun the job, spyglass pages need to only show the rerun button when the ProwJob resource is still available in the cluster (or otherwise disable the button).  ProwJob resources are typically uploaded to GCS as json so we could make this smarter in a follow up PR, but we'll still need a way to disable or hide the button since we can't assume the ProwJob will be uploaded to GCS.
So with respect to whether the button link or internal rerun button should be hidden: both since neither can be provided if a ProwJob resource is not available.

YuviGold: 1. Yes. The icon is added as part of the load function.
https://github.com/kubernetes/test-infra/blob/953aa5b87533c2294f32d86159918eb7947cc767/prow/cmd/deck/static/spyglass/spyglass.ts#L185-L189
The div at the bottom of the page is the box that's going to pop up when the icon is clicked. Its style dimensions are defined on the CSS
https://github.com/kubernetes/test-infra/blob/779dbad39c4856489a1efe7b2485cfe9c42dd430/prow/cmd/deck/static/style.css#L308-L330
2. I made the button appear only in case prowJobName is avaiable.

YuviGold:
> To help with review in the future please consider using separate commits for renaming/moving code and functional changes.

@cjwagner I updated my code according to your review and I formatted my commits into 2 logical commits.

prow/cmd/deck/template/spyglass.html
cjwagner: :+1: LGTM

cjwagner (approved):
Thanks, Yuval!

PR #22712, opened 2021-06-26
Make spyglass URL links from deck relative to host

### LINKED ISSUE #22304: Original spyglass link is used from deck links on local run
**What happened**:
When deploying the deck locally with the [runlocal](https://github.com/kubernetes/test-infra/blob/master/prow/cmd/deck/runlocal) script, a local server is listening on http://localhost:8080.
The jobs links are pointing to the ProwJobs actual spyglass view instead of the local spyglass view.

for example
https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/ci-test-infra-update-slack-oncall/1395748929890422784
instead of 
http://localhost:8080/view/gs/kubernetes-jenkins/logs/ci-test-infra-update-slack-oncall/1395748929890422784

In case the build has a URL that have `/view` as a substring - it would use it for spyglass.
Like other hyperlinks, the link has to be related to the host location, then it would be easier to debug spyglass changes.
For example, the rerun button href link is built like this:
https://github.com/kubernetes/test-infra/blob/7ce0898a49812f933df76b1805aa1a02fa001f64/prow/cmd/deck/static/prow/prow.ts#L732-L733

1. Deploy local deck `./prow/cmd/deck/runlocal`
2. Open browser `http://localhost:8080`
3. Check the links on the `Job` column pointing to the original Prow host.

YuviGold:
Resolves #22304

### DIFF
--- a/prow/cmd/deck/static/prow/prow.ts
+++ b/prow/cmd/deck/static/prow/prow.ts
@@ -523,6 +523,11 @@ function redraw(fz: FuzzySearch, pushState: boolean = true): void {
             status: {startTime, completionTime = "", state = "", pod_name, build_id = "", url = ""},
         } = build;
 
+        let buildUrl = url;
+        if (url.includes('/view/')) {
+            buildUrl = `${window.location.origin}/${url.slice(url.indexOf('/view/') + 1)}`;
+        }
+
         let org = "";
         let repo = "";
         if (build.spec.refs !== undefined) {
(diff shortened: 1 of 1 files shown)


chaodaiG (approved):
LGTM
/hold
In case you want @alvaroaleman also take a look

YuviGold:
@alvaroaleman WDYT?

alvaroaleman (commented):
/hold cancel

Issue #22567, opened 2021-06-15
PR revision tooltip isn't shown on Prow Status page after redraw


YuviGold:
Continues https://github.com/kubernetes/test-infra/issues/22301

**What happened**:
When hovering above PR revisions on the main Prow Status page, the tooltip displaying the title of the PR is shown.
Afterward, in case filtering by any filter and causing the page to redraw itself - the tooltip won't work.

**What you expected to happen**:
As the tooltip is shown at first, it should work after filtering.

**How to reproduce it (as minimally and precisely as possible)**:
1. Goto Prow Status page filtered by presubmit jobs - https://prow.k8s.io/?type=presubmit
2. Try to hover above the revision links on the Revision column and see the tooltip is presented.
3. Goto the filter toolbar and try to change the filter and cause the page to redraw.
For example - choose type "all" and then go back to "presubmit"
4. Now once again try to hover above the revision links and see the tooltip isn't shown.

**Anything else we need to know?**:
The `redraw` function is called when the window reloads and each time a filter is chosen.
It seems that the current implementation just creates new rows with the same ProwJob info (title exists). Each new row gets a new ID. e.g. In case at the first time the row had attribute `pr-openshift/assisted-service-2014-tipID-1`  the next time after redraw it would be `pr-openshift/assisted-service-2014-tipID-2`
Additionally, the only difference between the rows DOM was `data-upgraded=",MaterialTooltip"` that's missing from the newly created row.

YuviGold:
@cjwagner Maybe you know? You added the tooltip element on https://github.com/kubernetes/test-infra/pull/8791

cjwagner:
I don't know sorry. This has been an issue for a while and manifests in a few places. These issues are likely the same root problem: 
- https://github.com/kubernetes/test-infra/issues/12081
- https://github.com/kubernetes/test-infra/issues/10605

xinydev:
/assign

It needs to call `upgradeElement` after creating new DOM elements dynamically. I will create a PR to fix this.   

https://getmdl.io/started/index.html#dynamic

Issue #12081, opened 2019-04-05
PR title tooltips on tide-history page disappear when filter is modified.


cjwagner:
The PR title tooltips that appear when you hover over a PR number on the tide-history page do not work after editing the filter selection. The page has to be refreshed after the filter is modified in order for the tooltips to work again.

Example: 
1) Go to the following link and notice the hover tooltips on the PR numbers: https://prow.k8s.io/tide-history?repo=kubernetes%2Fk8s.io
1) Change the branch filter to select `master` and notice that the tooltips no longer appear on hover.
1) Refresh the page and see that the tooltips work again.

I'm not sure why this is happening, but I suspect it is a simple fix. @Katharine do you know what is going on here?

/area prow/deck

stevekuznetsov:
xref #10605 ?

Katharine:
/reopen
/remove-lifecycle rotten

pretty sure we never fixed this

Issue #10605, opened 2019-01-04
Deck timestamp tooltips disappear when changing filter options


Katharine:
When changing the deck filter options the tooltips get lost. They shouldn't.

/kind bug
/area prow/deck
/assign

PR #22726, opened 2021-06-29
Fix PR revision tooltip isn't shown on Prow Status page after redraw

xinydev:
Fixes: https://github.com/kubernetes/test-infra/issues/22567 
Fixes: https://github.com/kubernetes/test-infra/issues/12081
FIxes: https://github.com/kubernetes/test-infra/issues/10605 

see https://getmdl.io/started/index.html#dynamic 

> Material Design Lite will automatically register and render all elements marked with MDL classes upon page load. However in the case where you are creating DOM elements dynamically you need to register new elements using the upgradeElement function

here we use [upgradeDom()](https://github.com/google/material-design-lite/blob/60f441a22ed98ed2c03f6179adf460d888bf459f/src/mdlComponentHandler.js#L38) to upgrade all DOM elements 

I test it locally, this fix works

### DIFF
--- a/prow/cmd/deck/static/vendor.d.ts
+++ b/prow/cmd/deck/static/vendor.d.ts
@@ -10,6 +10,8 @@ interface Window {
 // Needed for mysterious reasons, otherwise ts doesn't understand this is a module
 declare module 'dialog-polyfill';
 
+declare var componentHandler: any;
+
 // Enough typing for the Material Design library to be usable.
 interface MaterialSnackbarOptionsNoAction {
   message: string;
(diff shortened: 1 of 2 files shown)


alvaroaleman (commented):
/ok-to-test

prow/cmd/deck/static/vendor.d.ts
```diff
@@ -10,6 +10,8 @@ interface Window {
 // Needed for mysterious reasons, otherwise ts doesn't understand this is a module
 declare module 'dialog-polyfill';
 
+declare var componentHandler: any;
```
alvaroaleman, line 13: Can this not be of type `componentHandler` rather than `any`?

xinydev: To be honest, I don’t know the type of componentHandler, I refer to [stackoverflow](https://stackoverflow.com/questions/34800354/how-to-set-global-variable-to-material-design-lite-library-in-typescript) 

As the [JS Migration Guide](https://basarat.gitbook.io/typescript/type-system/migrating) say,  use `any`to suppressing errors?

alvaroaleman (approved):

Issue #22673, opened 2021-06-23
Allow GH teams to have override permissions for whole org


petr-muller:
**What would you like to be added**:
The override plugin already has `repo -> team` configuration, we'd like to have `org -> team`
https://github.com/kubernetes/test-infra/blob/88cd2798f36010e071a30c9827f90e647b59fc65/prow/plugins/config.go#L1767-L1769

**Why is this needed**:
We have a class of users who should be allowed this power in the whole (huge) org. Giving them admin for the whole org is too much, maintaining repo-level config is too much work.

I think it would be reasonable to simply allow org names in the above config list elements, not ivent a new field.

/cc @bparees

PR #22730, opened 2021-06-29
Allow GH teams to have overrides at the org level.

emilvberglind:
Resolves https://github.com/kubernetes/test-infra/issues/22673

Since the existing `AllowedGitHubTeams` field already contained `org/repo -> team` it seemed trivial enough to also allow it to contain `org -> team` rather than introducing a new field.


petr-muller:
@emilvberglind make sure you follow https://github.com/kubernetes/test-infra/pull/22730#issuecomment-870719170
/ok-to-test

petr-muller (approved):

fejta (approved):

PR #22734, opened 2021-06-29
Netpol skipping Feature:SCTP tests

knabben:
Skipping the regex from `SCTP` to `Feature:SCTP` under netpol tests


aojea:
/lgtm
/approve
let's see if we got it at the first shot :upside_down_face:

PR #22742, opened 2021-06-30
Add support for EasyCLA

### LINKED ISSUE #22721: CLA Plugin: Add support for EasyCLA

Add support to the [CLA Plugin](https://github.com/kubernetes/test-infra/blob/master/prow/plugins/cla/) for [EasyCLA](https://docs.linuxfoundation.org/lfx/v/v2/easycla).

ref: https://github.com/kubernetes/org/issues/2778

**Why is this needed**:
The LF plans to deprecate the current CLA service used by the project. They would like to migrate us off the old plugin by the 2nd week of August. They are handling the migration of the user databases, but if we want our labels to reflect the state of the cla, the plugin will have to support the `EasyCLA` context. 

For migration purposes it'd be great if it could support both the current CLA and the future one in case there is a need to back out of the migration on their side.

Example EasyCLA status: 
https://api.github.com/repos/CNCF-EasyCLAV2-Test/Repository1/statuses/31eb937fb47d258b060a762165f293de6f99d191

---
/priority important-soon
/sig contributor-experience testing

anusha94:
Added a new field `EasyCLAEnabled` in the plugins config and based on
this field, if set will assign the `claContextName` to `cla/easy-cla`, else
default to `cla/linuxfoundation`

Fixes #22721

### DIFF
--- a/prow/plugins/cla/cla.go
+++ b/prow/plugins/cla/cla.go
@@ -95,7 +95,7 @@ func handleStatusEvent(pc plugins.Agent, se github.StatusEvent) error {
 	return handle(pc.GitHubClient, pc.Logger, se, pc.PluginConfig.CLAConfig)
 }
 
-// 1. Check that the status event received from the webhook is for the CNCF-CLA. The valid values are cla/linuxfoundation and EasyCLA.
+// 1. Check that the status event received from the webhook is for the CNCF-CLA.
 // 2. Use the github search API to search for the PRs which match the commit hash corresponding to the status event.
 // 3. For each issue that matches, check that the PR's HEAD commit hash against the commit hash for which the status
 //    was received. This is because we only care about the status associated with the last (latest) commit in a PR.
@@ -239,8 +239,7 @@ func handleComment(gc gitHubClient, log *logrus.Entry, e *github.GenericCommentE
 	}
 
 	for _, status := range combined.Statuses {
-
-		// Valid status values are cla/linuxfoundation and EasyCLA
+		// Only consider "cla/linuxfoundation" or "EasyCLA" status
 		if contains(cc.CLAContextNames, status.Context) {
 
 			// Success state implies that the cla exists, so label should be cncf-cla:yes.
@@ -277,9 +276,6 @@ func handleComment(gc gitHubClient, log *logrus.Entry, e *github.GenericCommentE
 					}
 				}
 			}
-
-			// No need to consider other contexts once you find the one you need.
-			break
 		}
 	}
 	return nil
--- a/prow/plugins/cla/cla.go
+++ b/prow/plugins/cla/cla.go
@@ -95,7 +95,7 @@ func handleStatusEvent(pc plugins.Agent, se github.StatusEvent) error {
 	return handle(pc.GitHubClient, pc.Logger, se, pc.PluginConfig.CLAConfig)
 }
 
-// 1. Check that the status event received from the webhook is for the CNCF-CLA.
+// 1. Check that the status event received from the webhook is for the CNCF-CLA. The valid values are cla/linuxfoundation and EasyCLA.
 // 2. Use the github search API to search for the PRs which match the commit hash corresponding to the status event.
 // 3. For each issue that matches, check that the PR's HEAD commit hash against the commit hash for which the status
 //    was received. This is because we only care about the status associated with the last (latest) commit in a PR.
@@ -240,7 +240,7 @@ func handleComment(gc gitHubClient, log *logrus.Entry, e *github.GenericCommentE
 
 	for _, status := range combined.Statuses {
 
-		// Only consider "cla/linuxfoundation" status.
+		// Valid status values are cla/linuxfoundation and EasyCLA
 		if contains(cc.CLAContextNames, status.Context) {
 
 			// Success state implies that the cla exists, so label should be cncf-cla:yes.
--- a/prow/plugins/cla/cla.go
+++ b/prow/plugins/cla/cla.go
@@ -57,8 +57,7 @@ It may take a couple minutes for the CLA signature to be fully registered; after
 )
 
 var (
-	checkCLARe     = regexp.MustCompile(`(?mi)^/check-cla\s*$`)
-	claContextName string
+	checkCLARe = regexp.MustCompile(`(?mi)^/check-cla\s*$`)
 )
 
 func init() {
@@ -70,7 +69,7 @@ func helpProvider(config *plugins.Configuration, _ []config.OrgRepo) (*pluginhel
 	// The {WhoCanUse, Usage, Examples, Config} fields are omitted because this plugin cannot be
 	// manually triggered and is not configurable.
 	pluginHelp := &pluginhelp.PluginHelp{
-		Description: "The cla plugin manages the application and removal of the 'cncf-cla' prefixed labels on pull requests as a reaction to the " + claContextName + " github status context. It is also responsible for warning unauthorized PR authors that they need to sign the CNCF CLA before their PR will be merged.",
+		Description: "The cla plugin manages the application and removal of the 'cncf-cla' prefixed labels on pull requests as a reaction to the EasyCLA / cla-linuxfoundation github status context. It is also responsible for warning unauthorized PR authors that they need to sign the CNCF CLA before their PR will be merged.",
 	}
 	pluginHelp.AddCommand(pluginhelp.Command{
 		Usage:       "/check-cla",
@@ -93,7 +92,7 @@ type gitHubClient interface {
 }
 
 func handleStatusEvent(pc plugins.Agent, se github.StatusEvent) error {
-	return handle(pc.GitHubClient, pc.Logger, se, pc.PluginConfig.EasyCLAEnabled)
+	return handle(pc.GitHubClient, pc.Logger, se, pc.PluginConfig.CLAConfig)
 }
 
 // 1. Check that the status event received from the webhook is for the CNCF-CLA.
@@ -101,18 +100,12 @@ func handleStatusEvent(pc plugins.Agent, se github.StatusEvent) error {
 // 3. For each issue that matches, check that the PR's HEAD commit hash against the commit hash for which the status
 //    was received. This is because we only care about the status associated with the last (latest) commit in a PR.
 // 4. Set the corresponding CLA label if needed.
-func handle(gc gitHubClient, log *logrus.Entry, se github.StatusEvent, isEasyCLAEnabled bool) error {
-	if isEasyCLAEnabled {
-		claContextName = "cla/easy-cla"
-	} else {
-		claContextName = "cla/linuxfoundation"
-	}
-
+func handle(gc gitHubClient, log *logrus.Entry, se github.StatusEvent, cc plugins.CLAConfig) error {
 	if se.State == "" || se.Context == "" {
 		return fmt.Errorf("invalid status event delivered with empty state/context")
 	}
 
-	if se.Context != claContextName {
+	if !contains(cc.CLAContextNames, se.Context) {
 		// Not the CNCF CLA context, do not process this.
 		return nil
 	}
@@ -199,16 +192,10 @@ func handle(gc gitHubClient, log *logrus.Entry, se github.StatusEvent, isEasyCLA
 }
 
 func handleCommentEvent(pc plugins.Agent, ce github.GenericCommentEvent) error {
-	return handleComment(pc.GitHubClient, pc.Logger, &ce, pc.PluginConfig.EasyCLAEnabled)
+	return handleComment(pc.GitHubClient, pc.Logger, &ce, pc.PluginConfig.CLAConfig)
 }
 
-func handleComment(gc gitHubClient, log *logrus.Entry, e *github.GenericCommentEvent, isEasyCLAEnabled bool) error {
-	if isEasyCLAEnabled {
-		claContextName = "cla/easy-cla"
-	} else {
-		claContextName = "cla/linuxfoundation"
-	}
-
+func handleComment(gc gitHubClient, log *logrus.Entry, e *github.GenericCommentEvent, cc plugins.CLAConfig) error {
 	// Only consider open PRs and new comments.
 	if e.IssueState != "open" || e.Action != github.GenericCommentActionCreated {
 		return nil
@@ -254,7 +241,7 @@ func handleComment(gc gitHubClient, log *logrus.Entry, e *github.GenericCommentE
 	for _, status := range combined.Statuses {
 
 		// Only consider "cla/linuxfoundation" status.
-		if status.Context == claContextName {
+		if contains(cc.CLAContextNames, status.Context) {
 
 			// Success state implies that the cla exists, so label should be cncf-cla:yes.
 			if status.State == github.StatusSuccess {
@@ -296,4 +283,15 @@ func handleComment(gc gitHubClient, log *logrus.Entry, e *github.GenericCommentE
 		}
 	}
 	return nil
+}
+
+// contains checks if a string is present in a slice
+func contains(s []string, str string) bool {
+	for _, v := range s {
+		if v == str {
+			return true
+		}
+	}
+
+	return false
 }
--- a/prow/plugins/cla/cla.go
+++ b/prow/plugins/cla/cla.go
@@ -64,12 +64,6 @@ var (
 func init() {
 	plugins.RegisterStatusEventHandler(pluginName, handleStatusEvent, helpProvider)
 	plugins.RegisterGenericCommentHandler(pluginName, handleCommentEvent, helpProvider)
-	config := &plugins.Configuration{}
-	if config.EasyCLAEnabled {
-		claContextName = "cla/easy-cla"
-	} else {
-		claContextName = "cla/linuxfoundation"
-	}
 }
 
 func helpProvider(config *plugins.Configuration, _ []config.OrgRepo) (*pluginhelp.PluginHelp, error) {
@@ -99,7 +93,7 @@ type gitHubClient interface {
 }
 
 func handleStatusEvent(pc plugins.Agent, se github.StatusEvent) error {
-	return handle(pc.GitHubClient, pc.Logger, se)
+	return handle(pc.GitHubClient, pc.Logger, se, pc.PluginConfig.EasyCLAEnabled)
 }
 
 // 1. Check that the status event received from the webhook is for the CNCF-CLA.
@@ -107,7 +101,13 @@ func handleStatusEvent(pc plugins.Agent, se github.StatusEvent) error {
 // 3. For each issue that matches, check that the PR's HEAD commit hash against the commit hash for which the status
 //    was received. This is because we only care about the status associated with the last (latest) commit in a PR.
 // 4. Set the corresponding CLA label if needed.
-func handle(gc gitHubClient, log *logrus.Entry, se github.StatusEvent) error {
+func handle(gc gitHubClient, log *logrus.Entry, se github.StatusEvent, isEasyCLAEnabled bool) error {
+	if isEasyCLAEnabled {
+		claContextName = "cla/easy-cla"
+	} else {
+		claContextName = "cla/linuxfoundation"
+	}
+
 	if se.State == "" || se.Context == "" {
 		return fmt.Errorf("invalid status event delivered with empty state/context")
 	}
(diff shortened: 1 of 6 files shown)


anusha94:
Fixes #22742

prow/plugins/cla/cla.go
```diff
@@ -58,12 +57,19 @@ It may take a couple minutes for the CLA signature to be fully registered; after
...
 
 func init() {
 	plugins.RegisterStatusEventHandler(pluginName, handleStatusEvent, helpProvider)
 	plugins.RegisterGenericCommentHandler(pluginName, handleCommentEvent, helpProvider)
+	config := &plugins.Configuration{}
+	if config.EasyCLAEnabled {
+		claContextName = "cla/easy-cla"
+	} else {
+		claContextName = "cla/linuxfoundation"
+	}
```
MadhavJivrajani, line 72: If I'm not wrong, this will always set the `claContextName` to `cla/linuxfoundation` because `EasyCLAEnabled` will default to `false` in an empty `Configuration` struct?

anusha94: Ahhh...right. Thanks!

MadhavJivrajani:
Fixes #22721

prow/plugins/cla/cla_test.go
```diff
@@ -140,6 +141,21 @@ func TestCLALabels(t *testing.T) {
 			addedLabels:   []string{fmt.Sprintf("/#3:%s", labels.ClaNo)},
 			removedLabels: []string{fmt.Sprintf("/#3:%s", labels.ClaYes)},
 		},
+		{
+			name:           "cla/linuxfoundation status failure removes \"cncf-cla: yes\" label",
```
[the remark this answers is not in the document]
cjwagner: Please add test cases for successful status cases as well.

cjwagner (commented):
Thanks for the PR!

prow/plugins/config.go
```diff
@@ -90,6 +90,7 @@ type Configuration struct {
 	Welcome              []Welcome                    `json:"welcome,omitempty"`
 	Override             Override                     `json:"override,omitempty"`
 	Help                 Help                         `json:"help,omitempty"`
+	EasyCLAEnabled       bool                         `json:"easy_cla_enabled,omitempty"`
```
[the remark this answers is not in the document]
anusha94: Got it.

anusha94:
@cjwagner Thanks for the review :)

MadhavJivrajani:
@anusha94 seems like you have to run these commands to update the bazel rules and test fixtures:
```bash
# update hazel rules
bazel run @io_k8s_repo_infra//hack:update-bazel
# update dependencies
bazel run @io_k8s_repo_infra//hack:update-deps
# update test fixtures based on your changes (addition of `CLAConfig` struct)
UPDATE=true go test ./prow/plugins/... -run TestGenYamlDocs
```

anusha94:
Thanks @MadhavJivrajani  :)

prow/plugins/cla/cla.go
```diff
@@ -93,20 +92,20 @@ type gitHubClient interface {
...
 func handleStatusEvent(pc plugins.Agent, se github.StatusEvent) error {
-	return handle(pc.GitHubClient, pc.Logger, se)
+	return handle(pc.GitHubClient, pc.Logger, se, pc.PluginConfig.CLAConfig)
 }
 
 // 1. Check that the status event received from the webhook is for the CNCF-CLA.
 // 2. Use the github search API to search for the PRs which match the commit hash corresponding to the status event.
 // 3. For each issue that matches, check that the PR's HEAD commit hash against the commit hash for which the status
 //    was received. This is because we only care about the status associated with the last (latest) commit in a PR.
 // 4. Set the corresponding CLA label if needed.
```
MadhavJivrajani, line 102: nit: could you update this comment to reflect the new cla logic? I think only point `1` would need updating

prow/plugins/cla/cla.go
```diff
@@ -242,7 +241,7 @@ func handleComment(gc gitHubClient, log *logrus.Entry, e *github.GenericCommentE
 	for _, status := range combined.Statuses {
 
 		// Only consider "cla/linuxfoundation" status.
```
cjwagner, line 243: This comment is out of date.

prow/plugins/config.go
```diff
@@ -482,6 +483,11 @@ type ConfigUpdater struct {
 	GZIP bool `json:"gzip"`
 }
 
+// CLAConfig contains the list of available context names
+type CLAConfig struct {
+	CLAContextNames []string `json:"cla_context_names,omitempty"`
```
cjwagner, line 488: Without defaulting logic this is a breaking change for all existing users. We need to default `CLAContextNames` to `[]string{"cla/linuxfoundation"}` if it is not specified. (Since the empty slice would never be a reasonable config we can just check if the slice is empty to determine if we should default.)
Please add defaulting logic to this function: https://github.com/kubernetes/test-infra/blob/13e6fe9adce3cc1929cd96ad5816f20c79306b5b/prow/plugins/config.go#L947-L994

cjwagner (changes requested):
I see one breaking change that we need to avoid, but otherwise looks good to me.

prow/plugins/cla/cla.go
```diff
@@ -93,15 +93,21 @@ type gitHubClient interface {
...
 // 2. Use the github search API to search for the PRs which match the commit hash corresponding to the status event.
 // 3. For each issue that matches, check that the PR's HEAD commit hash against the commit hash for which the status
 //    was received. This is because we only care about the status associated with the last (latest) commit in a PR.
 // 4. Set the corresponding CLA label if needed.
-func handle(gc gitHubClient, log *logrus.Entry, se github.StatusEvent) error {
+func handle(gc gitHubClient, log *logrus.Entry, se github.StatusEvent, isEasyCLAEnabled bool) error {
+	if isEasyCLAEnabled {
+		claContextName = "cla/easy-cla"
+	} else {
+		claContextName = "cla/linuxfoundation"
```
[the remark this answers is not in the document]
mrbobbytables: > When migrating from the current CLA service to EasyCLA there will likely be some PRs with the old status context, some with the new, and maybe even some with both. A single boolean toggle to switch the context we respect doesn't allow us to handle this case gracefully. (@mrbobbytables appears to be more familiar with this and can hopefully confirm if my assumptions about the migration process are true.)

💯 

We want to be able to support both in the event there is an issue with easycla and we have to stay on the current version.

anusha94:
@cjwagner Could you please review again?

prow/plugins/cla/cla.go
```diff
@@ -241,8 +240,8 @@ func handleComment(gc gitHubClient, log *logrus.Entry, e *github.GenericCommentE
 
 	for _, status := range combined.Statuses {
 
-		// Only consider "cla/linuxfoundation" status.
-		if status.Context == claContextName {
+		// Valid status values are cla/linuxfoundation and EasyCLA
+		if contains(cc.CLAContextNames, status.Context) {
```
spiffxp, line 244: Can we rewrite this so that it iterates through statuses to determine CLA yes/no first, and then once outside the for loop, applies labels accordingly?  I'd like to avoid wasting GitHub tokens on label apply/remove calls while we decide which status wins if they start competing.

prow/plugins/plugin-config-documented.yaml
```diff
@@ -316,6 +316,9 @@ cherry_pick_unapproved:
     # Comment is the comment added by the plugin while adding the
     # `do-not-merge/cherry-pick-not-approved` label.
     comment: ' '
+cla_config:
+    cla_context_names:
```
spiffxp, line 320: Please regenerate docs so this field has a useful comment

prow/plugins/cla/cla_test.go
```diff
@@ -140,6 +141,34 @@ func TestCLALabels(t *testing.T) {
 			addedLabels:   []string{fmt.Sprintf("/#3:%s", labels.ClaNo)},
 			removedLabels: []string{fmt.Sprintf("/#3:%s", labels.ClaYes)},
 		},
+		{
+			name:      "EasyCLA status failure removes \"cncf-cla: yes\" label",
+			context:   "EasyCLA",
+			state:     "failure",
```
spiffxp, line 147: Since the plugin now supports multiple contexts, I think it's fair to ask the test cases be written to test multiple contexts (or that you add a new Test that supports this).

I would like to see test cases with multiple contexts having conflicting values, so we can confirm that it's last-context-wins in all cases

spiffxp (changes requested):
I have concerns over dueling contexts

prow/plugins/cla/cla_test.go
spiffxp: e.g. I would like to see something like
```go
contexts:
  "cla/linuxfoundation": "success" 
  "EasyCLA": "failure"
```

and vice-versa

prow/plugins/cla/cla.go
spiffxp: Is `combined.Statuses` sorted at this point? or could we be walking these in random order?

prow/plugins/cla/cla.go
```diff
@@ -93,20 +92,20 @@ type gitHubClient interface {
 }
 
 func handleStatusEvent(pc plugins.Agent, se github.StatusEvent) error {
-	return handle(pc.GitHubClient, pc.Logger, se)
+	return handle(pc.GitHubClient, pc.Logger, se, pc.PluginConfig.CLAConfig)
 }
 
-// 1. Check that the status event received from the webhook is for the CNCF-CLA.
+// 1. Check that the status event received from the webhook is for the CNCF-CLA. The valid values are cla/linuxfoundation and EasyCLA.
```
spiffxp, line 98: If these are the only valid values, I would add validation logic confirming as much.  Otherwise, this is a configuration-specific detail

prow/plugins/cla/cla_test.go
```diff
@@ -140,6 +141,34 @@ func TestCLALabels(t *testing.T) {
 			addedLabels:   []string{fmt.Sprintf("/#3:%s", labels.ClaNo)},
 			removedLabels: []string{fmt.Sprintf("/#3:%s", labels.ClaYes)},
 		},
+		{
+			name:      "EasyCLA status failure removes \"cncf-cla: yes\" label",
+			context:   "EasyCLA",
```
spiffxp, line 146: More of a nice-to-have cosmetic thing... is `EasyCLA` hardcoded on your end? It would be preferable to have something like `cla/EasyCLA`

mrbobbytables: On the LF side, its hard-coded =/
Here is an example:
https://api.github.com/repos/CNCF-EasyCLAV2-Test/Repository1/statuses/31eb937fb47d258b060a762165f293de6f99d191

anusha94: Just to confirm - so this remains as `EasyCLA` right? And not `cla/EasyCLA`

prow/plugins/cla/cla.go
anusha94: @mrbobbytables 
`The valid values are cla/linuxfoundation and EasyCLA` Is this a right statement to make?

prow/plugins/cla/cla.go
anusha94: @spiffxp Can you please elaborate on competing statuses? Going by the existing code `if status.Context == claContextName`, I assumed the `for loop` should not continue if it is not one of `cla/linuxfoundation` or `EasyCLA`

Or are you saying this entire `for loop `should be rewritten such that applying of `ClaNo` or `ClaYes` should be outside the for loop?
Thanks.

anusha94: > Is combined.Statuses sorted at this point? or could we be walking these in random order?

@mrbobbytables Do you know about the sort order of `combined.Statuses`? Or rather, how can I check / verify the order?

prow/plugins/cla/cla_test.go
spiffxp: Sure, if it's hardcoded I guess it has to be `EasyCLA`

prow/plugins/cla/cla.go
spiffxp: The comment "the valid values..." does not match the state of this code.  You're not validating that these are the only two values allowed.  _Any_ values are allowed as configuration data.  This plugin could just as easily be configured to apply the CLA label if the the `monkeys` status context was successful.  I think this is great.  Generally, more configurability for re-use is good.

If you agree with this, then you should drop the "the valid values" part of the comment.  But, If you disagree, and think this should be purpose-built for just those two status contexts, then you should validate that these are the only valid values in some kind of `validate` function.

prow/plugins/cla/cla.go
spiffxp: >Can you please elaborate on competing statuses?

I missed the `break` down below.  You're right, one of them will win.  But will it be the same one every time?  And how will I know which one that is?

prow/plugins/cla/cla.go
anusha94: Okay I get it now. I am going to remove the comment that those are the only two valid values.

prow/plugins/cla/cla.go
anusha94: I removed the `break` statement so that all contexts will be processed.

anusha94:
@spiffxp 

I removed the `break` statement from [here](https://github.com/kubernetes/test-infra/pull/22742/files#diff-45486175573234cb728c8f2bfc7f2da46d7431b8eb4f3f049bbbcc73c119af15L283) in order for all the contexts to be processed.

In `TestCheckCLA`, I changed the [context field from string to a map](https://github.com/kubernetes/test-infra/pull/22742/files#diff-1e758035e054060499e235849f012fda0bfbd157353c30ced428f790e0c38994R214). Does this seem okay? Or having a separate `Test` for multiple contexts make more sense?

prow/plugins/cla/cla.go
```diff
@@ -278,10 +276,18 @@ func handleComment(gc gitHubClient, log *logrus.Entry, e *github.GenericCommentE
 					}
 				}
 			}
-
-			// No need to consider other contexts once you find the one you need.
-			break
```
spiffxp, line 283: Since there's no break anymore, now we _are_ potentially adding/removing labels twice.

What I asked for in last review was to make the decision before acting, e.g.
```golang
claState := github.StatusPending
for _, status := range combined.Statuses {
  // code that sets claState
}
// code that sets labels based on claState
```

prow/plugins/cla/cla_test.go
```diff
@@ -348,11 +425,20 @@ func TestCheckCLA(t *testing.T) {
 				Number:     3,
 				IssueState: tc.issueState,
 			}
+			cc := plugins.CLAConfig{
+				CLAContextNames: []string{"cla/linuxfoundation", "EasyCLA"},
+			}
+			var statuses []github.Status
+			for context, state := range tc.contexts {
+				statuses = append(statuses, github.Status{
```
spiffxp, line 433: re: my "are the statuses sorted" comment above, your unit tests are relying on _this_ being the sort

what I lack is proof that you've verified this sort is what you actually see in production (show me code that presorts the statuses, or docs from github guaranteeing the order in which they deliver statuses)

prow/plugins/cla/cla_test.go
```diff
@@ -161,7 +190,10 @@ func TestCLALabels(t *testing.T) {
 			SHA:     tc.statusSHA,
 			State:   tc.state,
 		}
-		if err := handle(fc, logrus.WithField("plugin", pluginName), se); err != nil {
+		cc := plugins.CLAConfig{
+			CLAContextNames: []string{"cla/linuxfoundation", "EasyCLA"},
+		}
+		if err := handle(fc, logrus.WithField("plugin", pluginName), se, cc); err != nil {
```
spiffxp, line 196: OK, I think I assumed there was less duplication of logic before

So the way I read the StatusEvent handling code, is that it will respond to whatever the status is.  So if the CLA plugin is configured to listen to two CLA statuses, we will have a race between which of the two statuses reports in first.  Slowest status wins.

I think what would provide more consistency is to try to de-dupe the logic and do what `handleComment` does: get the PR's combined statuses, and ensure one of them consistently wins the race.  

e.g. let's say cla/linuxfoundation consistently wins:
```yaml
# happy path
state: 
  cla/linuxfoundation: pending
  EasyCLA: pending
  labels: []
receive:
  cla/linuxfoundation: success
result:
  add: "cncf-cla: yes"

# conflict: EasyCLA arrives last
state: 
  cla/linuxfoundation: success
  EasyCLA: pending
  labels: ["cncf-cla: yes"]
receive:
  EasyCLA: failure
result:
  # do nothing

# conflict: cla/linuxfoundation arrives last
state: 
  cla/linuxfoundation: pending
  EasyCLA: success
  labels: ["cncf-cla: yes"]
receive:
  cla/linuxfoundation: failure
result:
  add: "cncf-cla: no"
  remove: "cncf-cla: yes"
```

spiffxp (changes requested):
I'm not confident enough to deploy this plugin in a manner that listens to two status contexts simultaneously yet.  I wrote down some test cases I would want to see exercised to ensure consistency vs. a race between last-context-reporting-wins.

I can be convinced to allow the EasyCLA check to report to PRs as non-blocking, without this plugin configured to listen to it, to allow us to gather data on how often we encounter conflict between the old CLA check and the new CLA check.

prow/plugins/cla/cla_test.go
```diff
@@ -332,6 +373,42 @@ func TestCheckCLA(t *testing.T) {
 
 			removedLabel: fmt.Sprintf("/#3:%s", labels.ClaYes),
 		},
+		{
+			name: "cla/linuxfoundation status retains the cla-no label and removes cla-yes label when its state is \"failure\"",
+			contexts: map[string]string{
+				"cla/linuxfoundation": "failure",
+				"EasyCLA":             "success",
+			},
```
spiffxp, line 381: Thanks for the refactor down here.  I think I'm asking for a similar refactor to the status event handling test

anusha94: sure. will do :)

prow/plugins/cla/cla_test.go
anusha94: @spiffxp 

I hear you. But I could not find proof for sort order. Here is the docs for [getting combined status](https://docs.github.com/en/rest/reference/repos#get-the-combined-status-for-a-specific-reference)
and the [client code to get the same](https://github.com/google/go-github/blob/ac48d59eb13be558759e9dd1e6f8f75126a83a4e/github/repos_statuses.go#L114)

prow/plugins/cla/cla_test.go
anusha94: >  So if the CLA plugin is configured to listen to two CLA statuses

From the docs, `The most recent status for each context is returned`.
I think my question is - are `EasyCLA` and `cla/linuxfoundation` considered as two different contexts or different names for a single `CLA` context? How does getting combined status here work? Combined status is still going to return both the contexts - how do we know which one is the latest?

spiffxp: >I think my question is - are EasyCLA and cla/linuxfoundation considered as two different contexts or different names for a single CLA context

Well, I'm assuming I'm going to see two entries, one called `cla/linuxfoundation` and one called `EasyCLA` while the CNCF has both things running simultaneously on their end.  So they're different contexts.  You're updating the plugin to decide which one to treat as the authoritative source of truth from the CNCF.  This allows us to have the plugin pay attention to the new EasyCLA check, but defer to the old cla/linuxfoundation check as authoritative, without having to change anything.

Another possibly simpler option is to roll back some of what you've done, and update the plugin to listen to one-and-only-one context.  In this scenario we would:
- make the plugin listen to a configurable context
- start with cla/linuxfoundation as this plugin's source of truth
- setup EasyCLA on the CNCF's end to start reporting to PRs
- setup prow/branch-protector to treat EasyCLA contexts as non-blocking
- wait
- generate a report that shows over time, for all PRs within a given window, PRs where cla/linuxfoundation and EasyCLA differed
- figure out why and fix
- wait again
- generate report again
- repeat until there is no difference, and then we change branch-protector to make the EasyCLA context required, the cla/linuxfoundation context non-blocking, and the cla plugin to listen to EasyCLA to apply "cncf-cla: yes" labels

spiffxp: >How does getting combined status here work? Combined status is still going to return both the contexts - how do we know which one is the latest?

What I was saying above is that the `handle` function _doesn't call GetCombinedStatus_, it only consider the status from the StatusEvent it's handling.  So it will get called once whenever `cla/linuxfoundation` is update, and again whenever `EasyCLA` is updated, and there's no way to guarantee which of those is going to arrive first.

Calling GetCombinedStatus to get both contexts is the point.  We don't _care_ which one came in latest, we only care that it came in for the commit we're considering, and that the authoritative context (if set) always overrides the others.

Again, I'm thinking this may all be too confusing, and it might be simpler to support a single (configurable) context only.  But we need to see how EasyCLA behaves live (in terms of responsiveness, reliability, consistency, and agreement with cla/linuxfoundation) before I will ever consider changing our workflow to pay attention to that over cla/linuxfoundation.

prow/plugins/cla/cla_test.go
spiffxp: maybe this is a sign you should sort them before iterating over them?  Or, flip the loop so you iterate over the context names that you care about in the order you care about, and then see if they're contained in combined statuses

prow/plugins/cla/cla_test.go
mrbobbytables: EasyCLA has been enabled for the [k/org](https://github.com/kubernetes/org) repo in a non-blocking way as a test.
For some additional information, EasyCLA is currently in use by 3 other CNCF projects:
- Open telemetry (47 repos)
- grpc (26 repos)
- cloud custodian (6 repos)

For gRPC it's been some time (most of the kinks in the project were worked out with this one), I don't know when it was enabled for Otel or cloud custodian.

The LF has said if we do have problems we can roll back.

MadhavJivrajani:
FYI https://github.com/kubernetes/test-infra/pull/23648
cc @spiffxp @anusha94

MadhavJivrajani:
/close
closing this out in favor of a simpler switch over. Will send in a PR soon!

PR #22747, opened 2021-06-30
Add a title tooltip for builds on job history view

YuviGold:
Add a job-history.ts module that goes over the build rows and adds a
tooltip with the build title to them.

Screenshot from 2021-06-29 09-45-44

### DIFF
--- a/prow/cmd/deck/static/BUILD.bazel
+++ b/prow/cmd/deck/static/BUILD.bazel
@@ -201,6 +201,22 @@ rollup_bundle(
     ],
 )
 
+ts_library(
+    name = "job_history",
+    srcs = glob(["job-history/*.ts"]),
+    deps = [
+        ":common",
+    ],
+)
+
+rollup_bundle(
+    name = "job_history_bundle",
+    entry_point = ":job-history/job-history.ts",
+    deps = [
+        ":job_history",
+    ],
+)
+
 test_suite(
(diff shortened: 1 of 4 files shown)


chaodaiG:
Generally looks good to me, one quick question: why only a single row in your example contains title?

chaodaiG:
sorry forgot to ask: what will this look like for presubmit jobs run on multiple PRs?

YuviGold:
> Generally looks good to me, one quick question: why only a single row in your example contains title?

It is shown only when hovering over a row, next to it.

> sorry forgot to ask: what will this look like for presubmit jobs run on multiple PRs?

I'm not sure if the job-history page supports it.
There are other fields that are used from the build's singular pull. The only addition is the title is now in use.
Could you share with me an example of where it happens?

chaodaiG:
Screen Shot 2021-06-30 at 2 22 19 PM
It's this job https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/batch/pull-kubernetes-e2e-kind-ipv6/1410343710918447104
And interestingly job history page https://prow.k8s.io/job-history/gs/kubernetes-jenkins/pr-logs/directory/pull-kubernetes-e2e-kind-ipv6 doesn't have it, so 👍

chaodaiG (approved):

PR #22756, opened 2021-07-01
snyk-periodic: Fix style and jq nits

navidshaikh:
- export result with 'vulnerabilities' map key
 - multi-line commands
 - header-footer message before and after the scan

followup https://github.com/kubernetes/test-infra/pull/22597#pullrequestreview-686740984
/cc @spiffxp

### DIFF
--- a/config/jobs/kubernetes/wg-k8s-infra/trusted/wg-k8s-infra-trusted.yaml
+++ b/config/jobs/kubernetes/wg-k8s-infra/trusted/wg-k8s-infra-trusted.yaml
@@ -87,13 +87,23 @@ periodics:
       args:
       - -c
       - |
-        set -euo pipefail; \
-        apt update && apt -y install jq; \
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes/wg-k8s-infra/trusted/wg-k8s-infra-trusted.yaml
```diff
@@ -87,13 +87,23 @@ periodics:
       args:
       - -c
       - |
-        set -euo pipefail; \
```
navidshaikh, line 90: removed since `snyk test` might return 1 if vulns found

navidshaikh: however, this may silence any errors occurred earlier during `apt update` or while installing jq or snyk

navidshaikh: updated the PR
- keep `set -euo pipefail` and update the `snyk test` command with exit code check

PR #22760, opened 2021-07-01
Align job history with tide history view

YuviGold:
- Move formatDuration to common package

Remove `parseDuration` as it isn't used.
Align `cell.time` tooltip with `prRevision`.

- Move job history builds rendering to typescript module

Add a `Revision` column and align with tide history and deck PR revision view.
`Started` column time display is aligned with the tide history view.
Now it would show a tooltip when hovering over it, and use the same TZ
Support multiple pulls for batch jobs.

Screenshot from 2021-07-01 20-15-03

### DIFF
--- a/prow/cmd/deck/job_history.go
+++ b/prow/cmd/deck/job_history.go
@@ -67,7 +67,7 @@ type buildData struct {
 	Duration     time.Duration
 	Result       string
 	commitHash   string
-	Pull         *prowv1.Pull
+	Pulls        []prowv1.Pull
 }
 
 // storageBucket is an abstraction for unit testing
@@ -370,8 +370,8 @@ func getBuildData(ctx context.Context, bucket storageBucket, dir string) (buildD
 	if err != nil {
 		logrus.WithError(err).Debugf("failed to read %s", prowv1.ProwJobFile)
 	} else {
-		if pj.Spec.Refs != nil && len(pj.Spec.Refs.Pulls) > 0 {
-			b.Pull = &pj.Spec.Refs.Pulls[0]
+		if pj.Spec.Refs != nil {
+			b.Pulls = pj.Spec.Refs.Pulls
 		}
 	}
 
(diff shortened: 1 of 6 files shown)


chaodaiG:
Thank you!

a quick question before going over the code change: the time displayed on prow.k8s.io is PST timezone, while job history is UTC, can we make them the same to be consistent by the way?

prow/cmd/deck/job_history.go
```diff
@@ -371,7 +371,7 @@ func getBuildData(ctx context.Context, bucket storageBucket, dir string) (buildD
 		logrus.WithError(err).Debugf("failed to read %s", prowv1.ProwJobFile)
 	} else {
 		if pj.Spec.Refs != nil && len(pj.Spec.Refs.Pulls) > 0 {
```
chaodaiG, line 373: do we still need `len(pj.Spec.Refs.Pulls) > 0 `?

prow/cmd/deck/job_history.go
```diff
@@ -67,7 +67,7 @@ type buildData struct {
 	Duration     time.Duration
 	Result       string
 	commitHash   string
-	Pull         *prowv1.Pull
+	Pulls        []prowv1.Pull
```
chaodaiG, line 70: wouldn't this break `.Pull` from https://github.com/kubernetes/test-infra/pull/22747/files?

chaodaiG (changes requested):

prow/cmd/deck/job_history.go
YuviGold: The code was changed accordingly 
https://github.com/kubernetes/test-infra/pull/22760/files#diff-5f8d6554e048de4637c1ba0e5ba42031d49e9dff2021510486955ebe338054b0R27-R33

YuviGold:
> Thank you!
> 
> a quick question before going over the code change: the time displayed on prow.k8s.io is PST timezone, while job history is UTC, can we make them the same to be consistent by the way?

Actually, I didn't mention that as part of the commit message since I looked only for the tooltip preview, so I used the same function. But it was resolved as part of it! :)

prow/cmd/deck/job_history.go
YuviGold: I believe you are right. Will fix.

chaodaiG (approved):

Issue #22764, opened 2021-07-01
Prow PubSub triggering support for Presubmits and Postsubmits


cjwagner:
**What would you like to be added**:
Prow's `sub` component currently only supports a single message event type: requests to trigger periodic jobs. https://github.com/kubernetes/test-infra/blob/08e324b7f660d997201943752e400637fb245317/prow/pubsub/subscriber/subscriber.go#L144-L152
We should add support for `presubmit` and `postsubmit` events to trigger these types of jobs as well. The message payloads would be similar to that of a `periodic`, but would also include the [Refs](https://github.com/kubernetes/test-infra/blob/6d5470c6523e609b584761121ebaadff92f42452/prow/apis/prowjobs/v1/types.go#L853) for the ProwJob.

**Why is this needed**:
Users would like to be able to use pubsub to trigger presubmit and postsubmit jobs. Periodics are not associated with git refs and do not report back to code review.

/area prow prow/sub
/assign @cjwagner @chaodaiG

listx:
@chaodaiG I guess this issue can be closed?

chaodaiG:
yes
/close

PR #22765, opened 2021-07-01
Prow sub: interface supports presubmit and postsubmit jobs

chaodaiG:
This is just interface change, the real implementation will be added in follow up PRs

Part of #22764

### DIFF
--- a/prow/pubsub/subscriber/subscriber.go
+++ b/prow/pubsub/subscriber/subscriber.go
@@ -19,6 +19,7 @@ package subscriber
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 
 	"cloud.google.com/go/pubsub"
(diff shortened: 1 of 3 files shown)


prow/pubsub/subscriber/subscriber.go
```diff
@@ -141,62 +187,61 @@ func (s *Subscriber) handleMessage(msg messageInterface, subscription string) er
...
+		jh = &presubmitJobHandler{}
+	case postsubmitProwJobEvent:
+		jh = &postsubmitJobHandler{}
+	default:
+		l.WithField("type", eType).Error("Unsupported event type")
+		s.Metrics.ErrorCounter.With(prometheus.Labels{subscriptionLabel: subscription})
+		return fmt.Errorf("unsupported event type: %s", eType)
+	}
+	if err = s.handleProwJob(l, jh, msg, subscription); err != nil {
+		l.WithError(err).Error("failed to create Prow Periodic Job")
```
cjwagner, line 205: No longer just Periodic jobs.

prow/pubsub/subscriber/subscriber.go
```diff
@@ -141,62 +187,61 @@ func (s *Subscriber) handleMessage(msg messageInterface, subscription string) er
...
-	for _, job := range s.ConfigAgent.Config().AllPeriodics() {
-		if job.Name == pe.Name {
-			periodicJob = &job
-			break
-		}
-	}
-	if periodicJob == nil {
-		err := fmt.Errorf("failed to find associated periodic job %q", pe.Name)
+	prowJobSpec, labels, err := jh.getProwJobSpec(s.ConfigAgent.Config(), pe)
+	if err != nil || prowJobSpec == nil {
```
cjwagner, line 231: It only makes sense to check if `prowJobSpec` is nil if it is possible for both `err` and `prowJobSpec` to be nil. That isn't possible currently and probably shouldn't be allowed. If it is expected to be allowed then we should update this to populate `err` in that case, otherwise we can return nil in an error case.

cjwagner (approved):
Just some nits.

prow/pubsub/subscriber/subscriber.go
chaodaiG: fixed

prow/pubsub/subscriber/subscriber.go
chaodaiG: good catch

cjwagner (approved):

Issue #22744, opened 2021-06-30
kubernetes unit test blocking dashboard is not loading


BenTheElder:
It's debateable where this issue belongs, but see:
- https://github.com/kubernetes/test-infra/pull/22524#issuecomment-871629288
- https://github.com/kubernetes/k8s.io/issues/1900#issuecomment-871625877

We probably need to switch to package level results

BenTheElder:
So the junit does have `<testsuites>` at the package level, containing `<testcases>` of which we seem to have too many for testgrid. See e.g. https://storage.googleapis.com/kubernetes-jenkins/logs/ci-kubernetes-unit/1410664595521015808/artifacts/junit_20210701-182317.xml

Asking if we can filter these in testgrid, or perhaps produce a filtered version only for testgrid and not for deck, or else increase testgrid available memory.

BenTheElder:
after dropping to 7 days of results in https://github.com/kubernetes/test-infra/pull/22767 it still does not load. will try 1 day just to see if that approach is viable at all.

BenTheElder:
waiting for https://testgrid.k8s.io/q/config/sig-release-master-blocking/ci-kubernetes-unit to reflect https://github.com/kubernetes/test-infra/pull/22776 then checking https://testgrid.k8s.io/sig-release-master-blocking#ci-kubernetes-unit again

BenTheElder:
it loads now. we still have a lot https://testgrid.k8s.io/sig-release-master-blocking#ci-kubernetes-unit&width=20, we have 30,000 rows so understandably it is not quite as quick as some other pages :-)

PR #22776, opened 2021-07-02
drop days of results to 1 for unit job

BenTheElder:
it's still not loading, presumably due to too much data, if it doesn't load after this, altering this particular config is not going to be a solution period.
https://github.com/kubernetes/test-infra/issues/22744#issuecomment-873179127


ameukam:
/lgtm
/hold
Remove when you're ready.

PR #22777, opened 2021-07-02
Prow pubsub: supports presubmit jobs

chaodaiG:
Part of #22764

### DIFF
--- a/prow/cmd/sub/main.go
+++ b/prow/cmd/sub/main.go
@@ -33,6 +33,7 @@ import (
 	"k8s.io/test-infra/prow/crier/reporters/pubsub"
 	"k8s.io/test-infra/prow/flagutil"
 	configflagutil "k8s.io/test-infra/prow/flagutil/config"
+	"k8s.io/test-infra/prow/git/v2"
 	"k8s.io/test-infra/prow/interrupts"
 	"k8s.io/test-infra/prow/logrusutil"
 	"k8s.io/test-infra/prow/metrics"
@@ -45,6 +46,7 @@ var (
 
 type options struct {
 	client         flagutil.KubernetesOptions
+	github         flagutil.GitHubOptions
 	port           int
 	pushSecretFile string
 
@@ -80,6 +82,7 @@ func init() {
 
 	flagOptions.config.AddFlags(fs)
 	flagOptions.client.AddFlags(fs)
+	flagOptions.github.AddFlags(fs)
 	flagOptions.instrumentationOptions.AddFlags(fs)
 
 	fs.Parse(os.Args[1:])
@@ -93,16 +96,22 @@ func main() {
 		logrus.WithError(err).Fatal("Error starting config agent.")
 	}
 
-	var tokenGenerator func() []byte
+	var tokens []string
 	if flagOptions.pushSecretFile != "" {
-		var tokens []string
 		tokens = append(tokens, flagOptions.pushSecretFile)
+	}
+	if flagOptions.github.TokenPath != "" {
+		tokens = append(tokens, flagOptions.github.TokenPath)
+	}
+	secretAgent := &secret.Agent{}
+	if err := secretAgent.Start(tokens); err != nil {
+		logrus.WithError(err).Fatal("Error starting secrets agent.")
+	}
+	tokenGenerator := secretAgent.GetTokenGenerator(flagOptions.pushSecretFile)
 
-		secretAgent := &secret.Agent{}
-		if err := secretAgent.Start(tokens); err != nil {
-			logrus.WithError(err).Fatal("Error starting secrets agent.")
-		}
-		tokenGenerator = secretAgent.GetTokenGenerator(flagOptions.pushSecretFile)
+	gitClient, err := flagOptions.github.GitClient(secretAgent, flagOptions.dryRun)
+	if err != nil {
+		logrus.WithError(err).Fatal("Error getting Git client.")
 	}
 
 	prowjobClient, err := flagOptions.client.ProwJobClient(configAgent.Config().ProwJobNamespace, flagOptions.dryRun)
@@ -125,6 +134,7 @@ func main() {
 		ConfigAgent:   configAgent,
 		Metrics:       promMetrics,
 		ProwJobClient: kubeClient,
+		GitClient:     git.ClientFactoryFrom(gitClient),
 		Reporter:      pubsub.NewReporter(configAgent.Config), // reuse crier reporter
 	}
 
--- a/prow/pubsub/subscriber/subscriber.go
+++ b/prow/pubsub/subscriber/subscriber.go
@@ -33,6 +33,7 @@ import (
 	prowapi "k8s.io/test-infra/prow/apis/prowjobs/v1"
 	v1 "k8s.io/test-infra/prow/apis/prowjobs/v1"
 	"k8s.io/test-infra/prow/config"
+	"k8s.io/test-infra/prow/git/v2"
 	"k8s.io/test-infra/prow/pjutil"
 )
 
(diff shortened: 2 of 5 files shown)


prow/pubsub/subscriber/subscriber.go
```diff
@@ -43,6 +45,22 @@ const (
 	postsubmitProwJobEvent = "prow.k8s.io/pubsub.PostsubmitProwJobEvent"
 )
 
+// Ensure interface is intact
+var _ prowCfgClient = (*config.Config)(nil)
+
+type githubClient interface {
+	GetRef(org, repo, ref string) (string, error)
+	// Compatible with config.githubClient
+	GetPullRequestChanges(org, repo string, number int) ([]github.PullRequestChange, error)
```
cjwagner, line 54: We should try to avoid using a GitHub client, we don't want to assume that the job uses GitHub for source control.

This function is also unused.

prow/pubsub/subscriber/subscriber.go
```diff
@@ -154,16 +172,70 @@ func (peh *periodicJobHandler) getProwJobSpec(cfg *config.Config, pe ProwJobEven
...
+}
+
+func (prh *presubmitJobHandler) getProwJobSpec(cfg prowCfgClient, pe ProwJobEvent) (*v1.ProwJobSpec, map[string]string, error) {
+	// presubmit jobs require Refs and Refs.Pulls to be set
+	refs := pe.Refs
+	if refs == nil {
+		return nil, nil, errors.New("Refs must be supplied")
+	}
+	if len(refs.Pulls) != 1 {
+		return nil, nil, fmt.Errorf("expect 1 pull, got: %d", len(refs.Pulls))
```
cjwagner, line 187: I don't think there is any need to forbid batches.

prow/pubsub/subscriber/subscriber.go
```diff
@@ -154,16 +172,70 @@ func (peh *periodicJobHandler) getProwJobSpec(cfg *config.Config, pe ProwJobEven
...
+
+	var presubmitJob *config.Presubmit
+	org, repo, branch := refs.Org, refs.Repo, pr.Ref
+	orgRepo := org + "/" + repo
+	baseSHAGetter := func() (string, error) {
+		if refs.BaseSHA != "" {
+			return refs.BaseSHA, nil
+		}
+		// Fall back to fetch from GitHub
+		baseSHA, err := prh.GitHubClient.GetRef(refs.Org, refs.Repo, "heads/"+refs.BaseRef)
```
cjwagner, line 199: I'm not sure how I feel about this. On one hand it helps ensure that the presubmit spec is deterministic by ensuring the baseref is resolved once, rather than at runtime every time it is rerun. On the other hand this assumes the code review system is GitHub. This only plays a role if the user fails to supply the base SHA in their pubsub request so I don't feel too strongly, but I'd lean toward removing this assumption and just omit the BaseSHA if it is not specified in the request. WDYT?

prow/pubsub/subscriber/subscriber.go
```diff
@@ -154,16 +172,70 @@ func (peh *periodicJobHandler) getProwJobSpec(cfg *config.Config, pe ProwJobEven
...
+		logrus.WithError(err).Debug("Failed to get presubmits")
+		presubmits = cfg.GetPresubmitsStatic(orgRepo)
+	}
+
+	for _, job := range presubmits {
+		if !job.CouldRun(branch) { // filter out jobs that are not branch matching
+			continue
+		}
+		if job.Name == pe.Name {
+			presubmitJob = &job // running multiple jobs with the same name for a PR is not a good idea, so take the first one
```
cjwagner, line 223: If there are multiple jobs with the same name that match this could be very confusing to debug, so it would probably be a good idea to at least log a warning that multiple jobs matched, or maybe even fail to trigger the jobs all together and return an error..

prow/pubsub/subscriber/subscriber.go
chaodaiG: good point, deleted

prow/pubsub/subscriber/subscriber.go
chaodaiG: Agreed, allowed

prow/pubsub/subscriber/subscriber.go
chaodaiG: Totally makes sense, deferred to runtime

prow/pubsub/subscriber/subscriber.go
chaodaiG: returns an error instead

prow/pubsub/subscriber/subscriber.go
```diff
@@ -154,16 +165,65 @@ func (peh *periodicJobHandler) getProwJobSpec(cfg *config.Config, pe ProwJobEven
...
+	var presubmitJob *config.Presubmit
+	org, repo, branch := refs.Org, refs.Repo, refs.BaseRef
+	orgRepo := org + "/" + repo
+	baseSHAGetter := func() (string, error) {
+		return refs.BaseSHA, nil
+	}
+	var headSHAGetters []func() (string, error)
+	for _, pull := range refs.Pulls {
+		headSHAGetters = append(headSHAGetters, func() (string, error) {
+			return pull.SHA, nil
```
cjwagner, line 191: This captures the local variable `pull` which changes each iteration of the loop. We should create a copy of `pull` inside the `for` loop and reference that within the anonymous function so that each function correctly references a different pull.

prow/pubsub/subscriber/subscriber.go
```diff
@@ -154,16 +165,65 @@ func (peh *periodicJobHandler) getProwJobSpec(cfg *config.Config, pe ProwJobEven
 }
 
 // presubmitJobHandler implements jobHandler
-type presubmitJobHandler struct{}
+type presubmitJobHandler struct {
+	GitClient git.ClientFactory
```
cjwagner, line 169: This git client is never instantiated in the `handleMessage` function.

cjwagner (changes requested):
Please add some unit tests too.

prow/pubsub/subscriber/subscriber.go
chaodaiG: good catch, adde

prow/pubsub/subscriber/subscriber.go
chaodaiG: good point, done

prow/cmd/sub/main.go
```diff
@@ -93,16 +96,22 @@ func main() {
...
+		logrus.WithError(err).Fatal("Error starting secrets agent.")
+	}
+	tokenGenerator := secretAgent.GetTokenGenerator(flagOptions.pushSecretFile)
 
-		secretAgent := &secret.Agent{}
-		if err := secretAgent.Start(tokens); err != nil {
-			logrus.WithError(err).Fatal("Error starting secrets agent.")
-		}
-		tokenGenerator = secretAgent.GetTokenGenerator(flagOptions.pushSecretFile)
+	gitClient, err := flagOptions.github.GitClient(secretAgent, flagOptions.dryRun)
```
cjwagner, line 112: Does this couple us to GitHub or does it work with Gerrit as well? We just need a git client so we should probably avoid coupling to GH.

cjwagner (approved):

prow/cmd/sub/main.go
cjwagner: /hold

chaodaiG: This client is for inrepoconfig, which is not supported in gerrit yet. We can change here once inrepoconfig is supported in gerrit

cjwagner: That makes sense, but I'm more wondering how this behaves without GitHub credentials supplied. In subscriber.go@198 we try to use the git client independent of whether we are using GitHub or Gerrit. It seems strange to use a client for a different code review provider when know it won't work and that also seems like it will result in misleading error messages.

 --------
Looks like we never try to use the git client [until after we've checked if InRepoConfig is enabled for the repo](https://github.com/kubernetes/test-infra/blob/a739353cae1e6c87cf22645918e4a4954acfdaf7/prow/config/config.go#L328-L330) and flag validation should pass if no GitHub flags are supplied so I think this is ok.
/shrug
/hold cancel

cjwagner (approved):

PR #22780, opened 2021-07-02
drop caps and set runAsGroup in unit-experimental

BenTheElder:
ref: https://github.com/kubernetes/test-infra/pull/22707/files#r661893286

### DIFF
--- a/config/jobs/kubernetes/sig-testing/make-test.yaml
+++ b/config/jobs/kubernetes/sig-testing/make-test.yaml
@@ -48,8 +48,14 @@ presubmits:
     spec:
       # unit tests have no business requiring root or doing privileged operations
       securityContext:
-        runAsUser: 2000
+        # NOTE: these are arbitrary non-root values. They don't exist in the
+        # image and don't need to, the unit tests should only write to TMPDIR
+        runAsUser: 2001
+        runAsGroup: 2010
         allowPrivilegeEscalation: false
+        capabilities:
+          drop:
+          - ALL
       containers:
         - image: gcr.io/k8s-testimages/kubekins-e2e:latest-master
           imagePullPolicy: Always


BenTheElder:
cc @vinayakankugoyal

config/jobs/kubernetes/sig-testing/make-test.yaml
```diff
@@ -48,8 +48,14 @@ presubmits:
     spec:
       # unit tests have no business requiring root or doing privileged operations
       securityContext:
-        runAsUser: 2000
+        # NOTE: these are arbitrary non-root values. They don't exist in the
+        # image and don't need to, the unit tests should only write to TMPDIR
+        runAsUser: 2001
```
vinayakankugoyal, line 53: Just curious: why as this changed? we could just set the runAsGroup to 2000 as well.

BenTheElder: I thought maybe these are even less likely to be introduced into the images as "real values".

2001 & 2010 is a silly nod.

config/jobs/kubernetes/sig-testing/make-test.yaml
```diff
@@ -48,8 +48,14 @@ presubmits:
...
+        runAsGroup: 2010
         allowPrivilegeEscalation: false
+        capabilities:
```
BenTheElder, line 56: inexplicably this is not showing up on the real prowjob https://prow.k8s.io/prowjob?prowjob=05d2ab37-db78-11eb-a74b-4e89bee5b562, only the runAsGroup

I think these are the correct keys and indentation ...?

BenTheElder: https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/92316/pull-kubernetes-unit-experimental/1411066357079871488/podinfo.json

vinayakankugoyal: I think I know why,

this is the PodSecurityContext which does not have fields allowPrivilegeEscalation and capabilities. Those are in the containers SecurityContext

vinayakankugoyal: You will need to add a securityContext to each of the containers. so that in the end it looks like
```
Pod
 spec:
   securityContext: # all containers in this pod will run with uid: 2001 and gid: 2010
     runAsUser: 2001
     runAsGroup: 2010
  containers:
  - name: blah
     securityContext:
       allowPrivilegeEscalation: false
       capabilities:
         drop: ["ALL"]
 ```

BenTheElder: This is why we need https://github.com/kubernetes/test-infra/pull/21075

Thanks

BenTheElder: https://github.com/kubernetes/test-infra/pull/22784

PR #22784, opened 2021-07-02
make security context actually parse for pull-kubernetes-unit-experim…

BenTheElder:
…ental


vinayakankugoyal:
This looks good to me, do we want to move the other instances of `allowPrivilegeEscalation: false` for example on line 95, to the container securityContext? Right now they probably are not doing anything.

BenTheElder:
there's a risk that they do something and break the actual blocking jobs, so starting with this manually triggered one. once we resolve any issues or observe none we can roll the same changes including runAsGroup to the other jobs

PR #22788, opened 2021-07-03
Do not run automatic presubmit Prow jobs if the Gerrit change revision is draft

chizhg:
/cc @chaodaiG @cjwagner @fejta

### DIFF
--- a/prow/gerrit/adapter/adapter.go
+++ b/prow/gerrit/adapter/adapter.go
@@ -202,7 +202,7 @@ func failedJobs(account int, revision int, messages ...gerrit.ChangeMessageInfo)
 	return failures
 }
 
-// processChange creates new presubmit prowjobs base off the gerrit changes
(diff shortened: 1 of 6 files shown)


chizhg:
We also need to trigger "all" Prow jobs if the revision is changed from `draft` to `active`. Checking if it's possible..

chizhg:
> We also need to trigger "all" Prow jobs if the revision is changed from `draft` to `active`. Checking if it's possible..

When a Gerrit Change is changed from draft to active state, a message `Set Ready For Review` will be posted. It looks we can use it to decide if it's the presubmit Prow jobs need to be triggered. It doesn't look great, but I'm unable to find a better way to handle it...

fejta:
/uncc

LGTM but deferring to Cole and Chao for the review. Please add me back if I'm needed.

cjwagner (approved):

PR #22803, opened 2021-07-06
Override: Support test names

### LINKED ISSUE #21556: prow: override: use test names, not status contexts
Today, users interact with Prow jobs by using `/test <name>`. However, administrators must use `/override <context>` even when the context is not the test name. We should allow `/override` to use test names as that's what users are familiar with.

/cc @bparees 

WDYT @cjwagner @chaodaiG @alvaroaleman ?

YuviGold:
Resolves #21556

 `/override <token>`
Will search:

- job names
- contexts on the PR

### DIFF
--- a/prow/plugins/override/override.go
+++ b/prow/plugins/override/override.go
@@ -334,15 +334,32 @@ func handle(oc overrideClient, log *logrus.Entry, e *github.GenericCommentEvent,
 		return oc.CreateComment(org, repo, number, plugins.FormatResponseRaw(e.Body, e.HTMLURL, user, resp))
 	}
 
+	baseSHAGetter := shaGetterFactory(oc, org, repo, pr.Base.Ref)
+	presubmits, err := oc.presubmits(org, repo, baseSHAGetter, sha)
+	if err != nil {
+		msg := "Failed to get presubmits"
+		log.WithError(err).Error(msg)
+		return oc.CreateComment(org, repo, number, plugins.FormatResponseRaw(e.Body, e.HTMLURL, user, msg))
+	}
+
 	contexts := sets.NewString()
 	for _, status := range statuses {
 		if status.State == github.StatusSuccess {
 			continue
 		}
+
 		contexts.Insert(status.Context)
+
+		for _, job := range presubmits {
+			if job.Context == status.Context {
+				contexts.Insert(job.Name)
+				break
+			}
+		}
 	}
+
 	if unknown := overrides.Difference(contexts); unknown.Len() > 0 {
-		resp := fmt.Sprintf(`/override requires a failed status context to operate on.
+		resp := fmt.Sprintf(`/override requires a failed status context or a job name to operate on.
 The following unknown contexts were given:
 %s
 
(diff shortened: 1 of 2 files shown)


prow/plugins/override/override.go
```diff
@@ -334,15 +335,31 @@ func handle(oc overrideClient, log *logrus.Entry, e *github.GenericCommentEvent,
...
+
 	contexts := sets.NewString()
 	for _, status := range statuses {
 		if status.State == github.StatusSuccess {
 			continue
 		}
+
 		contexts.Insert(status.Context)
+
+		if result := funk.Find(presubmits, func(job config.Presubmit) bool {
```
alvaroaleman, line 354: Iterating over a slice to find a matching result is absolutely not worth an external dependency

prow/plugins/override/override.go
```diff
@@ -334,15 +335,31 @@ func handle(oc overrideClient, log *logrus.Entry, e *github.GenericCommentEvent,
...
+			return job.Context == status.Context
+		}); result != nil {
+			contexts.Insert(result.(config.Presubmit).Name)
+		}
 	}
+
 	if unknown := overrides.Difference(contexts); unknown.Len() > 0 {
-		resp := fmt.Sprintf(`/override requires a failed status context to operate on.
+		resp := fmt.Sprintf(`/override requires a failed status context or a Prow Job name to operate on.
```
alvaroaleman, line 362:
```suggestion
		resp := fmt.Sprintf(`/override requires a failed status context or a job name to operate on.
```

Prowjobs are the custom resource instances and they have a randomized name so this is slightly confusing right now

prow/plugins/override/override_test.go
```diff
@@ -587,6 +600,93 @@ func TestHandle(t *testing.T) {
...
+				"ci/prow/prow-job": {
+					Context:     "ci/prow/prow-job",
+					State:       github.StatusSuccess,
+					Description: description(adminUser),
+				},
+			},
+		},
+		{
+			name:    "override same context and prow job",
+			comment: "/override ci/prow/prow-job\n/override prow-job",
```
alvaroaleman, line 663: This test only differs in the order of the comment from the previous one?

prow/plugins/override/override_test.go
```diff
@@ -587,6 +600,93 @@ func TestHandle(t *testing.T) {
...
+					State:       github.StatusFailure,
+				},
+			},
+			presubmits: map[string]config.Presubmit{
+				"prow-job": {
+					JobBase: config.JobBase{
+						Name: "prow-job",
+					},
+					Reporter: config.Reporter{
+						Context: "ci/prow/prow-job",
```
alvaroaleman, line 648: This test would already have passed before your change. Lets do this with a presubmit whose context is not the same as the one in the override command

prow/plugins/override/override_test.go
YuviGold: Yes. I'll make the name clearer.

prow/plugins/override/override_test.go
YuviGold: The test would have failed before it wouldn't accept "prow-job" as a valid override.
It was created mostly to verify both prow-job and the context names can be used together. But I'll change it as you asked

fejta (approved):
Great idea, thanks!

/hold

Alvaro or Cole any more comments?

alvaroaleman (approved):
looks good, thanks!
/hold cancel

PR #22781, opened 2021-07-02
Export konnectivity-agent.log for tests.

jkh52:
We want agent logs for debugging https://github.com/kubernetes/kubernetes/issues/102904

PR #22811, opened 2021-07-07
Fix konnectivity agent test logs export.

jkh52:
I made a mistake in https://github.com/kubernetes/test-infra/pull/22781

Since this is a daemonset there is a name suffix (for example, at https://storage.googleapis.com/kubernetes-jenkins/logs/ci-gcp-compute-persistent-disk-csi-driver-stable-k8s-master-migration/1412724970550726656/artifacts/e2e-test-prow-minion-group-69q0/kubelet.log we see: "kube-system/konnectivity-agent-2mdrf").

### DIFF
--- a/logexporter/cluster/log-dump.sh
+++ b/logexporter/cluster/log-dump.sh
@@ -42,7 +42,7 @@ readonly node_ssh_supported_providers="gce gke aws"
 readonly gcloud_supported_providers="gce gke"
 
 readonly master_logfiles="kube-apiserver.log kube-apiserver-audit.log kube-scheduler.log kube-controller-manager.log etcd.log etcd-events.log glbc.log cluster-autoscaler.log kube-addon-manager.log konnectivity-server.log fluentd.log kubelet.cov"
-readonly node_logfiles="kube-proxy.log konnectivity-agent.log fluentd.log node-problem-detector.log kubelet.cov"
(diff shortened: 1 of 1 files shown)


jkh52:
Notes:

log-dump.sh seems to simply look for var/log/{log_name} on the node: https://github.com/kubernetes/test-infra/blob/master/logexporter/cluster/log-dump.sh#L278

In a kubetest --up cluster, I see kube-proxy.log, but not konnectivity-agent.log:

# ls /var/log/*log
/var/log/alternatives.log  /var/log/cloud-init-output.log  /var/log/cloud-init.log  /var/log/kube-proxy.log  /var/log/lastlog

Note that kube-proxy sets up this specific log (https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/kube-proxy/kube-proxy-ds.yaml#L44) but konnectivity doesn't (https://github.com/kubernetes/kubernetes/blob/master/cluster/gce/addons/konnectivity-agent/konnectivity-agent-ds.yaml#L32).

jkh52:
/unhold

I think I've got a version that should work, based on:

```
e2e-test-jkh-minion-group-jv6g /var/log # ls /var/log/containers/konnectivity-agent-*.log
/var/log/containers/konnectivity-agent-rnmds_kube-system_konnectivity-agent-e06d503b15115622af5abe4fb53af118d6477157704b048683240fa112684ac0.log
```

PR #22812, opened 2021-07-07
Fix override tests

YuviGold:
- Check expected comments exist between actual comments.
  `fakeClient`'s `CreateComment` function isn't under test, so ignore any
  failure in order to verify comments.
  Fix `comment for incorrect context` and `comment on get pr failure`
  tests.

- Sort `allTeams` on `whoCanUse` for consistent output.

### DIFF
--- a/prow/plugins/override/override_test.go
+++ b/prow/plugins/override/override_test.go
@@ -131,16 +131,6 @@ func (c *fakeClient) presubmits(_, _ string, _ config.RefGetter, _ string) ([]co
 }
 
 func (c *fakeClient) CreateComment(org, repo string, number int, comment string) error {
-	switch {
-	case org != fakeOrg:
-		return fmt.Errorf("bad org: %s", org)
-	case repo != fakeRepo:
-		return fmt.Errorf("bad repo: %s", repo)
-	case number != fakePR:
-		return fmt.Errorf("bad number: %d", number)
-	case strings.Contains(comment, "fail-comment"):
-		return errors.New("injected CreateComment failure")
-	}
 	c.comments = append(c.comments, comment)
 	return nil
 }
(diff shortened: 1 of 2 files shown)


prow/plugins/override/override_test.go
```diff
@@ -131,16 +131,6 @@ func (c *fakeClient) presubmits(_, _ string, _ config.RefGetter, _ string) ([]co
 }
 
 func (c *fakeClient) CreateComment(org, repo string, number int, comment string) error {
-	switch {
-	case org != fakeOrg:
```
fejta, line 135: Why are you regressing this?

fejta (commented):
/ok-to-test

prow/plugins/override/override_test.go
YuviGold: `comment on get pr failure` test is testing a failure for getting a pr with a wrong PR number.
In case having those validations on `CreateComment` function then it would fail as well for having a wrong PR number.

Since the override tests don't test creating comments by any test, but utilize them for testing others, I figured this shouldn't have any special treatment at all and the `fakeClient` would just create the comments as expected (for later being verified as this PR fixes)

prow/plugins/override/override_test.go
```diff
@@ -949,7 +956,7 @@ func TestWhoCanUse(t *testing.T) {
 
 	who := whoCanUse(override, "org1", "repo1")
 	if who != expectedWho {
-		t.Errorf("expected %s, got %s", expectedWho, who)
```
fejta, line 952: This more correctly follows golang style.

I'd consider %q if you're concerned about empty strings.

fejta (approved):
/hold

prow/plugins/override/override_test.go
YuviGold: I just added a colon to make it clearer. While the messages might get long
```
    override_test.go:959: expected: Repo administrators, and the following github teams:org1/repo1: team-foo team-bar, org1: team-foo-bar., got: Repo administrators, and the following github teams:org1/repo1: team-foo team-bar, org1: team-foo-bar.
```

But using `%q` works as well that adds quotes :) 
```
    override_test.go:959: expected "Repo administrators, and the following github teams:org1/repo1: team-foo team-bar, org1: team-foo-bar.", got "Repo administrators, and the following github teams:org1/repo1: team-foo team-bar, org1: team-foo-bar."
```

fejta (approved):
/hold cancel

PR #22815, opened 2021-07-07
Ensure cloud-provider-gcp E2E tests run against k8s v1.21.0

DangerOnTheRanger:
The `cloud-provider-gcp-e2e-full` job was incorrectly downloading the v1.22 test suite, which caused some tests that wouldn't have otherwise failed to fail. This PR corrects that problem.


jpbetz:
/lgtm

Thanks!

PR #22820, opened 2021-07-08
Pin latest kubekins image on sig-node job

knabben:
Pinning latest image on sig-node lock contention


ipochi (approved):
/lgtm

thanks for updating this @knabben

PR #22821, opened 2021-07-08
pull-security-profiles-operator-test-e2e: add `hostPID` and `/var/log volume`

saschagrunert:
We need to test a feature within the security-profiles-operator which
requires us to read `/var/log/syslog` or `/var/log/audit/audit.log` as
well as tracking the container process by its cgroup (via
`/proc/$PID/cgroup`). This means we require:

- `hostPID` to be able to read the process cgroup of the running
  container
- mounting `/var/log` to read `/var/log/syslog`

I assume that the prow machines run Debian 10, which does not ship
auditd per default but the syslog daemon.


saschagrunert:
@hasheddan @pjbgf @JAORMX @jhrozek PTAL

JAORMX (approved):
/lgtm

PR #22823, opened 2021-07-08
pull-security-profiles-operator-test-e2e: switch `/var/log` to `/var/log/audit`

saschagrunert:
We cannot read-only mount /var/log without breaking docker in docker.
Switch to `/var/log/audit` since auditd seems to be installed on the
nodes.

Follow-up of https://github.com/kubernetes/test-infra/pull/22821


saschagrunert:
@hasheddan @pjbgf @JAORMX PTAL

PR #22828, opened 2021-07-08
AlertManager: Redirect AWS account alerts to a different Slack channel

randomvariable:
Redirect AWS account alerts to a different Slack channel
Migrates configuration from deprecated `match_re` to `matchers` based on PromQL.

!!!!!!!
Note, I don't have access to AlertManager, so I'm not sure this is correct. Please be ready to revert if need be.
!!!!!!!

Signed-off-by: Naadir Jeewa <jeewan@vmware.com>

### DIFF
--- a/config/prow/cluster/monitoring/alertmanager-prow_secret.yaml
+++ b/config/prow/cluster/monitoring/alertmanager-prow_secret.yaml
@@ -19,11 +19,17 @@ stringData:
       repeat_interval: 4h
       receiver: 'slack-warnings'
       routes:
+      - receiver: 'cluster-api-aws-alerts'
+        group_interval: 5m
+        repeat_interval: 2h
+        matchers:
+          - boskos_type = aws-account
       - receiver: 'slack-alerts'
         group_interval: 5m
         repeat_interval: 2h
-        match_re:
-          severity: 'critical|high'
+        matchers:
+          - severity =~ "critical|high"
+
 
     receivers:
     - name: 'slack-warnings'
@@ -39,6 +45,13 @@ stringData:
         api_url: '{{ api_url }}'
         icon_url: https://avatars3.githubusercontent.com/u/3380462
         text: '@test-infra-oncall {{ template "custom_slack_text" . }}'
+        link_names: true
+    - name: 'cluster-api-aws-alerts'
+      slack_configs:
+      - channel: '#cluster-api-aws-alerts'
+        api_url: '{{ api_url }}'
+        icon_url: https://avatars3.githubusercontent.com/u/3380462
+        text: '@aws-account-users {{ template "custom_slack_text" . }}'
         link_names: true
 
     templates:


config/prow/cluster/monitoring/alertmanager-prow_secret.yaml
```diff
@@ -22,8 +22,14 @@ stringData:
...
-        match_re:
-          severity: 'critical|high'
+        matchers:
+          - boskos_type != aws-account
+          - severity =~ critical|high
+      - receiver: 'cluster-api-aws-alerts'
+        group_interval: 5m
+        repeat_interval: 2h
+        matchers:
+          - boskos_type = aws-account
```
cjwagner, line 32: Listing this route before the route for the `slack-alerts` receiver would simplify this and make it more clear. Only the first route that matches will be used unless `continue` is set to `true`. See the [docs for routes](https://prometheus.io/docs/alerting/latest/configuration/#route). This would also allow us to keep the main alerting route the same so we'll be less likely to break existing alerts.

randomvariable:
should be able to test with something like:
```
curl -XPOST <alertmanager_url> -d "[{ 
	\"status\": \"firing\",
	\"labels\": {
		\"boskos_type\": \"aws-account\",
		\"severity\":\"high\",
	},
	\"annotations\": {
		\"message\": \"This is a test boskos alert for the AWS folk\"
	},
	\"generatorURL\": \"http://amazonaws.com\"
}]"
```

cjwagner (commented):
I think this would work as is, but I think we can reduce the config churn and simplify as well.

config/prow/cluster/monitoring/alertmanager-prow_secret.yaml
randomvariable: Cool, i'll update.

randomvariable: Updated

randomvariable:
And related PR for creating the Slack group https://github.com/kubernetes/community/pull/5881
Order of merge shouldn't matter though.

config/prow/cluster/monitoring/alertmanager-prow_secret.yaml
```diff
@@ -19,11 +19,17 @@ stringData:
...
       - receiver: 'slack-alerts'
         group_interval: 5m
         repeat_interval: 2h
-        match_re:
-          severity: 'critical|high'
+        matchers:
+          - severity =~ "critical|high"
```
cjwagner, line 31: This didn't need to change, but it looks equivalent according to the docs.

cjwagner (approved):
I'll update the secret after this merges.

config/prow/cluster/monitoring/alertmanager-prow_secret.yaml
randomvariable: I did it for completeness since `match_re` is marked as deprecated.

PR #22832, opened 2021-07-08
Fix alertmanager config to not use 'matchers' field.

cjwagner:
Switching to `matchers` in https://github.com/kubernetes/test-infra/pull/22828 broke the config.
Also for debugging in the future, note that it appears that prometheus-operator manages an intermediate generated secret that is consumed by alertmanager. Invalid configs cause errors in prometheus-operator logs, not alertmanager's, and the generated config that is actually used won't be updated.

I've fixed the secret in the cluster already and the config seems to have loaded properly now.
/assign @randomvariable


randomvariable:
Ah, ok. Good to know there's some safety there.

/lgtm

PR #22842, opened 2021-07-09
Fix a bug to skip presubmit tests for WorkInProgress Gerrit changes correctly

chizhg:
The `revision.Draft` field has been removed from the latest Gerrit API, `change.WorkInProgress` should be used instead.

To use the new field, we need to update github.com/andygrunwald/go-gerrit dependency to the latest.

/cc @cjwagner @chaodaiG


chizhg:
This PR is now ready for review.

cjwagner (approved):

PR #22844, opened 2021-07-09
Enable OVA CI for image-builder

codenrhoden:
With the addition of https://github.com/kubernetes-sigs/image-builder/pull/425, we now have working CI for Azure and OVAs. This PR enables the OVA CI by default, and changes the logic of when to run the presubmits to the following conditions:

 - Makefile
 - Ansible config file or any Ansible roles
 - provider specific CI script
 - Default configurations
 - GOSS config
 - provider specific Packer builder
 - hack/ensure-* scripts that are called by provider's 'make deps'

This makes it to where something like changing the Azure packer config doesn't run the OVA CI and vice versa. Howevever, any changes to anything common between the two will run both.

/assign @CecileRobertMichon

### DIFF
--- a/config/jobs/kubernetes-sigs/image-builder/image-builder-ova-presubmits.yaml
+++ b/config/jobs/kubernetes-sigs/image-builder/image-builder-ova-presubmits.yaml
@@ -6,11 +6,12 @@ presubmits:
         preset-kind-volume-mounts: "true"
         preset-cluster-api-provider-vsphere-e2e-config: "true"
         preset-cloud-provider-vsphere-e2e-config: "true"
-      always_run: false
-      optional: true
       decorate: true
+      decoration_config:
+        timeout: 1h
+      run_if_changed: 'images/capi/(Makefile|ansible\.cfg)|images/capi/ansible/.*|images/capi/scripts/ci-ova\.sh|images/capi/packer/(config|goss|ova)/.*|images/capi/hack/ensure-(ansible|packer|goss).*'
       path_alias: sigs.k8s.io/image-builder
-      max_concurrency: 5
+      max_concurrency: 3
       spec:
         containers:
           - image: gcr.io/k8s-testimages/kubekins-e2e:v20210709-09116fb-master
--- a/config/jobs/kubernetes-sigs/image-builder/image-builder-presubmits.yaml
+++ b/config/jobs/kubernetes-sigs/image-builder/image-builder-presubmits.yaml
@@ -4,7 +4,7 @@ presubmits:
     labels:
       preset-azure-cred: "true"
     decorate: true
-    run_if_changed: 'images/capi/.*'
+    run_if_changed: 'images/capi/(Makefile|ansible\.cfg)|images/capi/ansible/.*|images/capi/scripts/ci-azure-e2e\.sh|images/capi/packer/(config|goss|azure)/.*|images/capi/hack/ensure-(ansible|packer|jq|azure-cli|goss).*'
     decoration_config:
       timeout: 1h
     max_concurrency: 5


config/jobs/kubernetes-sigs/image-builder/image-builder-ova-presubmits.yaml
```diff
@@ -6,11 +6,12 @@ presubmits:
         preset-kind-volume-mounts: "true"
         preset-cluster-api-provider-vsphere-e2e-config: "true"
         preset-cloud-provider-vsphere-e2e-config: "true"
-      always_run: false
-      optional: true
       decorate: true
+      decoration_config:
+        timeout: 1h
+      run_if_changed: 'images/capi/(Makefile|ansible\.cfg)|images/capi/ansible/.*|images/capi/scripts/ci-ova\.sh|images/capi/packer/(config|goss|ova)/.*'
```
kkeshavamurthy, line 12: Should we include `hack/*` in the list for both azure & ova?

codenrhoden: ooh, good catch. I meant to do that. I know it makes this line a bit unwieldy, but I think it would be better to list the exact files we care about for each builder. There's a lot of stuff under `hack/` that don't need to kick of a specific CI build. And some of them are provider specific. I'll just use the provider's `make deps` target as a template.

codenrhoden: Okay, I added the `hack/ensure*` scripts that each provider uses.

config/jobs/kubernetes-sigs/image-builder/image-builder-presubmits.yaml
```diff
@@ -4,7 +4,7 @@ presubmits:
     labels:
       preset-azure-cred: "true"
     decorate: true
-    run_if_changed: 'images/capi/.*'
+    run_if_changed: 'images/capi/(Makefile|ansible\.cfg)|images/capi/ansible/.*|images/capi/scripts/ci-azure-e2e\.sh|images/capi/packer/(config|goss|azure)/.*|images/capi/hack/ensure-(ansible|packer|goss).*'
```
kkeshavamurthy, line 7: Does `ensure-ansible-windows.sh` match in this regex?
Will prob need it in ova after we enable windows ova ci.

config/jobs/kubernetes-sigs/image-builder/image-builder-ova-presubmits.yaml
```diff
@@ -6,11 +6,12 @@ presubmits:
         preset-kind-volume-mounts: "true"
         preset-cluster-api-provider-vsphere-e2e-config: "true"
         preset-cloud-provider-vsphere-e2e-config: "true"
-      always_run: false
-      optional: true
       decorate: true
+      decoration_config:
+        timeout: 1h
+      run_if_changed: 'images/capi/(Makefile|ansible\.cfg)|images/capi/ansible/.*|images/capi/scripts/ci-ova\.sh|images/capi/packer/(config|goss|ova)/.*|images/capi/hack/ensure-(ansible|packer|jq|azure-cli|goss).*'
```
kkeshavamurthy, line 12: I think `azure-cli` goes in the other job.

config/jobs/kubernetes-sigs/image-builder/image-builder-presubmits.yaml
codenrhoden: It does, yes. it's essentially saying `images/capi/hack/ensure-ansible*`, which matches both `ensure-ansible.sh` and `ensure-ansible-windows.sh`.

config/jobs/kubernetes-sigs/image-builder/image-builder-ova-presubmits.yaml
codenrhoden: Thanks! fixed.

codenrhoden:
@CecileRobertMichon When you are ready to take a look at this one, I think it will really help our testing throughput by being more choosy about what CI jobs we run and when.

CecileRobertMichon (commented):
looks great, thanks for doing this

/lgtm

Issue #5183, opened 2017-10-26
batch merges should link to their batch run results


spiffxp:
This may be obviated by tide, but I'm documenting how I just ran into it with mungegithub

This PR: https://github.com/kubernetes/kubernetes/pull/53760
- passed all presubmits
- submit queue tested it individually (/test all [submit-queue is verifying that this PR is safe to merge])
- it failed pull-kubernetes-bazel-test
- k8s-ci-robot (prow) commented as such, and linked me to https://k8s-gubernator.appspot.com/pr/53760 which also showed the test failure

At this point, I wouldn't expect the PR to merge.  It did:

- k8s-merge-robot commented that it merged the PR with a batch test
- the individual commit's status still shows a red X, and its submit-queue status shows a refusal to merge the commit
<img width="804" alt="screen shot 2017-10-25 at 6 16 52 pm" src="https://user-images.githubusercontent.com/49258/32030589-c8e510ca-b9b0-11e7-8570-1531ce18d20c.png">

- Using my knowledge of the infra, I knew to go look at http://prow.k8s.io/?type=batch and search the UI for #53760.  
- There, I could click through to the pull-kubernetes-bazel-test run for the batch: https://k8s-gubernator.appspot.com/build/kubernetes-jenkins/pr-logs/pull/batch/pull-kubernetes-bazel-test/12337/

Problems with this:
- GitHub shows the commit as having merged despite it failing
- Audit trail for this PR isn't discoverable from the PR, I needed to know where to look
- If deck (or whichever components of prow) were restarted, I'm not sure I'd be able to find the right gubernator link for the batch run

What I would expect to happen:
- if the PR merges in batch, after having flaked/failed in single, the commit status contexts should update to point to the batch results
- ideally, there would be a more permanent landing page (maybe in gubernator?) for "all results for the batch run that included this PR" vs. me having to spelunk through prow.k8s.io
- alternatively, the results for the batch run that included this PR would show up in the PR's result history at https://k8s-gubernator.appspot.com/pr/53760

/area mungegithub
/area gubernator
since this is where I encountered it

/area prow
since this is where merge logic is heading

spiffxp:
/remove-lifecycle stale
this is still a usability bug, removing stale to see if this is a thing we need to revisit when tide rolls through

spiffxp:
/remove-lifecycle stale
/lifecycle frozen
[...]

PR #22857, opened 2021-07-11
Support reporting batch jobs

YuviGold:
In order to improve transparency and make users be more exposed to failures.
This would allow a PR owner to be able to watch failures related to his PR from its PR view.

Otherwise, in the case of repeated batch intervals, a green PR could be shown as "stuck" for hours. (unless going to tide dashboard to understand what's going on with it)

### DIFF
--- a/prow/config/config.go
+++ b/prow/config/config.go
@@ -1800,9 +1800,8 @@ func parseProwConfig(c *Config) error {
 	}
 
 	// validate entries are valid job types
-	// Currently only presubmit and postsubmit can be reported to github
 	for _, t := range c.GitHubReporter.JobTypesToReport {
-		if t != prowapi.PresubmitJob && t != prowapi.PostsubmitJob {
+		if t != prowapi.PresubmitJob && t != prowapi.PostsubmitJob && t != prowapi.BatchJob {
 			return fmt.Errorf("invalid job_types_to_report: %v", t)
 		}
 	}
--- a/prow/config/config_test.go
+++ b/prow/config/config_test.go
@@ -2827,7 +2827,7 @@ func TestValidGitHubReportType(t *testing.T) {
 github_reporter:
   job_types_to_report:
   - presubmit
-  - batch
+  - periodic
 `,
 			expectError: true,
 		},
@@ -2838,8 +2838,9 @@ github_reporter:
   job_types_to_report:
   - presubmit
   - postsubmit
+  - batch
 `,
-			expectTypes: []prowapi.ProwJobType{prowapi.PresubmitJob, prowapi.PostsubmitJob},
+			expectTypes: []prowapi.ProwJobType{prowapi.PresubmitJob, prowapi.PostsubmitJob, prowapi.BatchJob},
 		},
 	}
 
--- a/prow/github/report/report.go
+++ b/prow/github/report/report.go
@@ -74,17 +74,25 @@ func reportStatus(ghc GitHubClient, pj prowapi.ProwJob) error {
 		if err != nil {
 			return err
 		}
-		sha := refs.BaseSHA
-		if len(refs.Pulls) > 0 {
-			sha = refs.Pulls[0].SHA
+		shaRefs := make([]string, 0)
+
+		if len(refs.Pulls) == 0 {
+			shaRefs = append(shaRefs, refs.BaseSHA)
+		} else {
+			for _, pull := range refs.Pulls {
+				shaRefs = append(shaRefs, pull.SHA)
+			}
 		}
-		if err := ghc.CreateStatus(refs.Org, refs.Repo, sha, github.Status{
-			State:       contextState,
-			Description: config.ContextDescriptionWithBaseSha(pj.Status.Description, refs.BaseSHA),
-			Context:     pj.Spec.Context, // consider truncating this too
-			TargetURL:   pj.Status.URL,
-		}); err != nil {
-			return err
+
+		for _, sha := range shaRefs {
+			if err := ghc.CreateStatus(refs.Org, refs.Repo, sha, github.Status{
+				State:       contextState,
+				Description: config.ContextDescriptionWithBaseSha(pj.Status.Description, refs.BaseSHA),
+				Context:     pj.Spec.Context, // consider truncating this too
+				TargetURL:   pj.Status.URL,
+			}); err != nil {
+				return err
+			}
 		}
 	}
 	return nil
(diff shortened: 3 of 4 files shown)


alvaroaleman (commented):
I believe this is not wanted as it might result in seeing failures on a pull that was caused by a different pull.
/cc @cblecker @cjwagner

YuviGold:
> I believe this is not wanted as it might result in seeing failures on a pull that was caused by a different pull.
> /cc @cblecker @cjwagner 

That's why this option is configurable and it would just become supported.

Honestly I'd prefer to be exposed to failures that are related to my PR and be able to watch them on my PR view. Otherwise, in case of repeated batch intervals, a green PR could be shown as "stuck" for hours. (unless going to tide dashboard to understand what's going on with it)

spiffxp:
For full context, I think @YuviGold is going after some of the suggestions I made at the end of my friction log here: https://github.com/kubernetes/test-infra/issues/5183#issuecomment-782769506

spiffxp:
The confusing thing I _think_ this is trying to address is 

>2. crier? is updating status contexts for PR's that tide triggers as not part of a batch

>[fixing] 2 would reduce confusion (both kinds of tide triggers should update a PR's commit status, or not) but may be not worth it for a PR that is triggered for both reasons

a.k.a as someone trying to land their PR it seems weird/leaky that tide will update status contexts on my PR depending on whether it's part of a batch or not

but I get the counter-point, that it would be weird if a PR was left with failing contexts from a failed batch run, and then it wasn't picked up as part of the next batch, so it looks like it's sitting there failed when I know it passed all of its presubmits prior to being mergeable (and would this invalidate it from subsequent batch selection etc.)

alvaroaleman:
> a.k.a as someone trying to land their PR it seems weird/leaky that tide will update status contexts on my PR depending on whether it's part of a batch or not

The important difference is that batch jobs make a statement about all PRs in them combined and presubmit jobs make a statement about the single PR, regardless of if they got triggered by Trigger or Tide. Tide filters out PRs that have failed contexts for required jobs: https://github.com/kubernetes/test-infra/blob/f587c4ae1725cdf6be2d6e0f2c8286d3d85cb137/prow/tide/tide.go#L616-L619

This means that we both can not report batch jobs and should report presubmit jobs created by tide, so failing PRs get thrown out.

I do realize that tides retesting in general and even more so in case of batches is hard to reason about and an area that can be improved. If you have more ideas on how to do that, they are very welcome (but please post on an issue before writing code).

alvaroaleman:
(I didn't mention the Tide context filtering part earlier because I didn't know about it. I noticed this when working on an unrelated change. The original Tide design predates me so I can only speculate about the reasoning here but I assume we use the GitHub contexts because contrary to prowjobs in the cluster, they do not vanish after 24 hours).

YuviGold:
> The confusing thing I _think_ this is trying to address is
> 
> > 1. crier? is updating status contexts for PR's that tide triggers as not part of a batch
> 
> > [fixing] 2 would reduce confusion (both kinds of tide triggers should update a PR's commit status, or not) but may be not worth it for a PR that is triggered for both reasons
> 
> a.k.a as someone trying to land their PR it seems weird/leaky that tide will update status contexts on my PR depending on whether it's part of a batch or not
> 

@spiffxp That's right this part is confusing. But it's mostly used to improve user transparency.
When a repository merge strategy is managed by tide, then I am not concerned about it managing GitHub contexts since this is basically its field of expertise.
At the end of the day, a user looks most of the time at the PR checks view and it could help him realize what's going on under-the-hood without understanding how Tide works.


Although, 

> but I get the counter-point, that it would be weird if a PR was left with failing contexts from a failed batch run, and then it wasn't picked up as part of the next batch, so it looks like it's sitting there failed when I know it passed all of its presubmits prior to being mergeable (and would this invalidate it from subsequent batch selection etc.)

I didn't think about this part.
I believe that this is a game-changer because it means that this new behavior affects the PR status and would get it stuck.

> This means that we both can not report batch jobs and should report presubmit jobs created by tide, so failing PRs get thrown out.

@alvaroaleman  This part is a bit confusing. So reporting batch jobs, as being implemented by this PR, shouldn't happen.
But aren't presubmits already being reported?

> I do realize that tides retesting in general and even more so in case of batches is hard to reason about and an area that can be improved. If you have more ideas on how to do that, they are very welcome (but please post on an issue before writing code).

This is what I actually wanted to do. Maybe we can enhance tide context description to have a reason for the retest - if it's being done because the target branch got changed or if it's a batch job that fails, etc.
https://github.com/kubernetes/test-infra/blob/739ff5cdeeee7861113f611a03719085fb191839/prow/tide/status.go#L315

alvaroaleman:
> @alvaroaleman This part is a bit confusing. So reporting batch jobs, as being implemented by this PR, shouldn't happen.
But aren't presubmits already being reported?


What is confusing?

Yes, presubmits are already supported because presubmit reflect a result that is based on the changes in the PR. Batch jobs do not, they reflect a result based on all changes in all prs in that batch combined. 

> This is what I actually wanted to do. Maybe we can enhance tide context description to have a reason for the retest - if it's being done because the target branch got changed or if it's a batch job that fails, etc.

That already exists, it says `retesting $jobname[, $jobname...]`. The retesting always means that there was a successful job already but we need retesting because the base changed since that job ran. If that retesting happens in a batch or not is not included. Also very technically the statement actually means "Waiting for retesting, which might or might not be running right now".

The issue with including more info there is that there is a character limit in the status context. The code for this is here: https://github.com/kubernetes/test-infra/blob/913692b63c74ca65c70ed05b2c5da0f889ead390/prow/tide/status.go#L308

A better approach might be to include more information in the pr status page you get to when clicking on the tide status context link.

spiffxp:
@alvaroaleman 

>This means that we both can not report batch jobs and should report presubmit jobs created by tide, so failing PRs get thrown out.

So if I understand this correctly, the idea is that if there is a perma-failing PR in the pool, it would eventually get thrown out of the pool by:
- failing when it's run as the single-PR tide-triggered re-run
- having its status contexts set to failing
- no longer meeting the selection criteria for the pool on the next tide query

Is that correct? If that's the case, ok, yes, I agree the report-on-trigger behavior should remain the same.

(Anecdotally it feels like we haven't seen that benefit in kubernetes/kubernetes when things get backlogged, but maybe you have)

From my comment earlier...

>honestly only reporting on success to make this happen, combined with not reporting on the PR at all when triggering via tide (batch or not), would be just fine by me

I was imagining something where tide could set the status contexts just prior to merge, but on further reflection I suspect we don't retain enough information to do this.

>A better approach might be to include more information in the pr status page you get to when clicking on the tide status context link.

I think I agree.  The goal here is make it not impossible to get from GitHub to the batch jobs that caused my PR to merge (or are blocking it).  It can be confusing when the single-run jobs fail (and report to the PR) but the batch runs pass (and don't report to the PR), so it looks like the PR merged with failing tests.

YuviGold:
I'm closing this PR since this solution would affect the PR status and how tide treats it.
I am going to open an issue and search for another way how to enhance batch job integration with GitHub

PR #22840, opened 2021-07-09
use gs://k8s-release-dev for kubernetes CI builds

spiffxp:
Part of umbrella issue to migrate the project away from the google.com-owned gs://kubernetes-release-dev GCS bucket: https://github.com/kubernetes/k8s.io/issues/2318

I broke this up into multiple commits for ease of revert.  Each commit message includes a list of jobs potentially impacted

Since one of them involves a kubekins-e2e change, a followup bump PR will be required for all changes to take effect.

PR #22860, opened 2021-07-12
Update sig-release jobs to use new ci-images location

jackfrancis:
This PR updates the sig-release jobs that refer to `gcr.io/kubernetes-ci-images` to instead use the new `gcr.io/k8s-staging-ci-images` path.

### DIFF
--- a/config/jobs/kubernetes/sig-release/kubernetes-builds.yaml
+++ b/config/jobs/kubernetes/sig-release/kubernetes-builds.yaml
@@ -47,7 +47,7 @@ periodics:
       - --
       - --allow-dup
       - --extra-version-markers=k8s-master
-      - --registry=gcr.io/kubernetes-ci-images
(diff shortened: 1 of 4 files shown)


jackfrancis:
cc @chewong

jackfrancis:
@spiffxp In the mad rush to investigate broken tests I suspected this change may help. But as the affecting change (#22840) originated in test-infra, this is probably not actually a valid change.

Can you confirm? If this is not relevant, I'll close the PR. (or feel free to close it yourself)

spiffxp:
/close
Yeah these deprecated jobs are intentionally pushing to the deprecated locations, so I'm going to close this.  Thanks for the effort though.

It looks like you got it sorted out in https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1512, I sketched out my guess at root cause in https://github.com/kubernetes/test-infra/pull/22840#issuecomment-878607626

PR #22864, opened 2021-07-12
config/jobs: fix kubernetes deprecated build jobs

spiffxp:
The `--release` parameter was being picked up bootstrap.py instead of
the scenario it was invoking.

This should resolve https://github.com/kubernetes/kubernetes/issues/103647 but I'm not using the fixes keyword so we can close it after we've seen successful job runs

This fixes a bug introduced in https://github.com/kubernetes/test-infra/pull/22840, specifically in https://github.com/kubernetes/test-infra/pull/22840/commits/cbd67f42667b7b46db147fa5bac3cc37dfe28f91

### DIFF
--- a/config/jobs/kubernetes/sig-release/kubernetes-builds.yaml
+++ b/config/jobs/kubernetes/sig-release/kubernetes-builds.yaml
@@ -43,8 +43,8 @@ periodics:
       - --root=/go/src
       - --timeout=240
       - --scenario=kubernetes_build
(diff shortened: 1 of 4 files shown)


ameukam:
/lgtm
/hold
Removed the hold if needed.

spiffxp:
ref: https://github.com/kubernetes/k8s.io/issues/2318

justaugustus:
Retroactive LGTM!
Thanks Aaron!

PR #22867, opened 2021-07-13
Drop explicit repo-root from pull-kubernetes-e2e-gce-kubetest2

amwat:
Quoting this correctly seems to be having issues. Drop it altogether, kubetest2 will default to $PWD anyway.

follow up to: https://github.com/kubernetes/test-infra/pull/22863
ref: https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/92316/pull-kubernetes-e2e-gce-kubetest2/1414701760827625472

/cc @BenTheElder @spiffxp


spiffxp (approved):
/approve
/lgtm

PR #22869, opened 2021-07-13
Fix typo - update loop device count in the comment

mkumatag:
The count got updated via - https://github.com/kubernetes/test-infra/pull/18090 but missed updating the comment, hence this PR is for updating the count in the comment as well.

/cc @BenTheElder


BenTheElder (approved):
/lgtm
/approve
thanks!

PR #22875, opened 2021-07-13
Allow workload-identity scripts to use region

coryrc:
Just a documentation change.

I verified the following `gcloud container` commands don't care if you
pass a region in to --zone:

clusters describe
node-pools list
clusters update

Additionally the "context" name also matched in my kube config.

### DIFF
--- a/workload-identity/bind-service-accounts.sh
+++ b/workload-identity/bind-service-accounts.sh
@@ -19,7 +19,7 @@ set -o nounset
 set -o pipefail
 
 if [[ $# != 6 ]]; then
-  echo "Usage: $(basename "$0") <project> <zone> <cluster> <namespace> <name> <gcp-service-account>" >&2
+  echo "Usage: $(basename "$0") <project> <zone_or_region> <cluster> <namespace> <name> <gcp-service-account>" >&2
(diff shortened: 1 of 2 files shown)


workload-identity/bind-service-accounts.sh
```diff
@@ -19,7 +19,7 @@ set -o nounset
 set -o pipefail
 
 if [[ $# != 6 ]]; then
-  echo "Usage: $(basename "$0") <project> <zone> <cluster> <namespace> <name> <gcp-service-account>" >&2
+  echo "Usage: $(basename "$0") <project> <zone_or_region> <cluster> <namespace> <name> <gcp-service-account>" >&2
```
fejta, line 22: The doc changes are good. I don't think there's need to churn the varaible name

coryrc: so `s/location/zone/g` and you're good?

(the line this comment is on will not change, which is my only confusion)

fejta: sorry, yes <zone_or_region> in the comments sgtm!

fejta (approved):

PR #22876, opened 2021-07-13
Prow sub: create git client only when token is supplied

chaodaiG:
Fix the bug that crash looping sub when github token is not provided

### DIFF
--- a/prow/pubsub/subscriber/subscriber.go
+++ b/prow/pubsub/subscriber/subscriber.go
@@ -194,13 +194,14 @@ func (prh *presubmitJobHandler) getProwJobSpec(cfg prowCfgClient, pe ProwJobEven
 		})
 	}
 
-	// This will work with inrepoconfig
-	presubmits, err := cfg.GetPresubmits(prh.GitClient, orgRepo, baseSHAGetter, headSHAGetters...)
-	if err != nil {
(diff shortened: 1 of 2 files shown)


prow/pubsub/subscriber/subscriber.go
```diff
@@ -194,13 +194,13 @@ func (prh *presubmitJobHandler) getProwJobSpec(cfg prowCfgClient, pe ProwJobEven
...
-		// Fall back to static presubmits to avoid deadlocking when a presubmit is used to verify
-		// inrepoconfig
-		logrus.WithError(err).Debug("Failed to get presubmits")
-		presubmits = cfg.GetPresubmitsStatic(orgRepo)
+	presubmits := cfg.GetPresubmitsStatic(orgRepo)
+	if prh.GitClient != nil { // Get from inrepoconfig only when GitClient is provided
+		var err error
+		presubmits, err = cfg.GetPresubmits(prh.GitClient, orgRepo, baseSHAGetter, headSHAGetters...)
+		if err != nil {
+			logrus.WithError(err).Debug("Failed to get presubmits")
```
cjwagner, line 202: I think we need to either return an error or avoid overwriting `presubmits` to `nil` in the error case.

chaodaiG: done

cjwagner (approved):

PR #22886, opened 2021-07-14
Branchprotector: Update status contexts if there is at least one optional job

alvaroaleman:
Currently, the branchprotector will ignore repos that do not have
mandatory jobs, unless the repo has an explicit config.
This leads to orphaned required status contexts on GitHub if all jobs on
a repo are made optional.

This change will make the branchprotector always update the required
status contexts if there is at least one job, optional or not.

Note that this changes behavior that was deliberate (i.E. we tested for). I can not see any downside to this other than that it might cost us a bit more tokens. If there is any other downside I missed, please do tell :)

/cc @petr-muller @cjwagner

### DIFF
--- a/prow/cmd/branchprotector/request.go
+++ b/prow/cmd/branchprotector/request.go
@@ -58,7 +58,7 @@ func makeBool(val *bool) bool {
 // makeChecks renders a ContextPolicy into the corresponding GitHub api object.
 //
 // Returns nil when input policy is nil.
-// Otherwise returns non-nil Contexts (empty if unset) and Strict iff Strict is true
+// Otherwise returns non-nil Contexts (empty if unset) and Strict if Strict is true
 func makeChecks(cp *branchprotection.ContextPolicy) *github.RequiredStatusChecks {
 	if cp == nil {
 		return nil
--- a/prow/config/branch_protection.go
+++ b/prow/config/branch_protection.go
@@ -192,6 +192,9 @@ type BranchProtection struct {
 	// AllowDisabledJobPolicies allows a branch to choose to opt out of branch protection
 	// even if Prow has registered required jobs for that branch.
 	AllowDisabledJobPolicies *bool `json:"allow_disabled_job_policies,omitempty"`
+	// ProtectReposWithOptionalJobs will make the Branchprotector manage required status
+	// contexts on repositories that only have optional jobs (default: false)
+	ProtectReposWithOptionalJobs *bool `json:"protect_repos_with_optional_jobs,omitempty"`
 }
 
 func isPolicySet(p Policy) bool {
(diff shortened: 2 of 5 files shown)


cjwagner:
> Note that this changes behavior that was deliberate (i.E. we tested for). I can not see any downside to this other than that it might cost us a bit more tokens. If there is any other downside I missed, please do tell :)

It sounds like this might make branchprotector remove manually configured branch protection rules on repos that did not opt in to branch protection. Is that correct?
Erick wrote the test and may be able to clarify.
/assign @fejta

alvaroaleman:
> It sounds like this might make branchprotector remove manually configured branch protection rules on repos that did not opt in to branch protection. Is that correct?

Yeah but only for cases where ppl relied on the repo not being managed by branchprotector due to lack of mandatory jobs while having optional jobs. If someone explicitly disables branchprotection in any way, that will still get respected.

prow/cmd/branchprotector/request.go
```diff
@@ -58,7 +58,7 @@ func makeBool(val *bool) bool {
 // makeChecks renders a ContextPolicy into the corresponding GitHub api object.
 //
 // Returns nil when input policy is nil.
-// Otherwise returns non-nil Contexts (empty if unset) and Strict iff Strict is true
```
petr-muller, line 61: nit: `iff` is a [thing](https://en.wikipedia.org/wiki/If_and_only_if), looks like this property did not change?

petr-muller (approved):
Neat improvement and solves a pitfall. We now have an explicit config for "don't touch my BP". I'm not sure how to not break (unset their manual BP setting) existing users that rely on this behavior though.

/hold

stevekuznetsov:
@alvaroaleman can we add a global config option to the tool to opt into this behavior?

```yaml
protect_repos_with_optional_jobs: true
```

alvaroaleman:
> @alvaroaleman can we add a global config option to the tool to opt into this behavior?

Added an opt-in, so this is not going to break anyone by default anymore. Enabling it would IMHO be the better default, but I realize that that might cause issues.

/hold cancel

prow/config/branch_protection.go
```diff
@@ -377,6 +385,16 @@ func (c *Config) GetPolicy(org, repo, branch string, b Branch, presubmits []Pres
 	return &policy, nil
 }
 
+func (c *Config) shouldManageRequiredStatusCheck(requiredContetexts, requiredIfPresentContexts, optionalContexts []string) bool {
```
stevekuznetsov, line 388:
```suggestion
func (c *Config) shouldManageRequiredStatusCheck(requiredContexts, requiredIfPresentContexts, optionalContexts []string) bool {
```

stevekuznetsov (commented):
One typo, otherwise LGTM

PR #22890, opened 2021-07-14
config/tests: jobs should not use gs://kubernetes-release-pull

spiffxp:
Log policy violations rather than fail the test. Once migration away from
 this bucket is complete, the test can be switched to fail instead of log.

 Ideally though, this test would then be switched to a test that fails for
 _all_ job types if gs://kubernetes-release-pull shows up as a substring in
 _any_ argument.

 We could also remove it (possibly N months down the line).

Part of https://github.com/kubernetes/test-infra/issues/18789


amwat (commented):
/lgtm

PR #22893, opened 2021-07-14
Prow sub: add docs for presubmit and postsubmit jobs

chaodaiG:


chaodaiG:
/cc @cjwagner 
cc @2nd-half

2nd-half (approved):
Cool. Thanks

prow/cmd/sub/README.md
```diff
@@ -75,3 +76,53 @@ This will find and start the periodic job ```my-periodic-job```, and add / overw
...
+  "envs":{...},
+  "labels":{...},
+  "annotations":{...},
+
+  "refs":{
+    "base_ref": "main",
+    "base_sha": "abc123",
+    "pulls": [
+      {
+        "sha": "def456"
```
cjwagner, line 97: We should link to the full `Refs` struct since there are a lot of additional fields that may be needed to achieve expected behavior: https://github.com/kubernetes/test-infra/blob/62f15287bd0c21ea8018bd4e8b6958d0f05177b4/prow/apis/prowjobs/v1/types.go#L882-L923

e.g. `CloneURI`, `PathAlias`, Pull `Number` and `Author`.

There are also two required fields that we're missing (and for which we are also missing validation in Sub): `Org` and `Repo` :man_facepalming:

chaodaiG: 🤦

chaodaiG: Done

prow/cmd/sub/README.md
```diff
@@ -75,3 +76,67 @@ This will find and start the periodic job ```my-periodic-job```, and add / overw
...
+      }
+    ]
+  }
+}
+```
+
+This will start presubmit job ```my-presubmit-job```, clones source code like pull requests
+defined under ```pulls```, which merges to ```base_ref``` at ```base_sha```.
+
+(There are more fields can be supplied, see [full documentation](https://github.com/kubernetes/test-infra/blob/18678b3b8f4bc7c51475f41964927ff7e635f3b9/prow/apis/prowjobs/v1/types.go#L883). For example, if you want the job to be reported on the PR, add ```number``` field right next to ```sha```)
```
cjwagner, line 112: This seems a bit misleading. I'm pretty sure reporting will happen whether or not the PR number is supplied. IIRC the PR number is used for things like constructing links, aggregation in Tide, and front end job filtering, but ProwJob results are reported to GitHub based on the commit SHA, not the PR number.

cjwagner (approved):
I think we may find that fields like the PR number should also be required or more strongly encouraged in the docs, but it is probably easier to trying things out and then come back to update this doc once we've seen which Prow features depend on omitted fields.

Issue #22888, opened 2021-07-14
ApplyDefault doesn't allow setting arrays to empty when round tripping from golang -> yaml -> golang


howardjohn:
**What happened**:
Defined slack reporter config in golang:
```go
&prowjob.ReporterConfig{
	Slack: &prowjob.SlackReporterConfig{
		JobStatesToReport: []prowjob.ProwJobState{},
	},
}
```
Later marshaled it to yaml, which end up like:
```yaml
    reporter_config:
      slack: {}
```

When this is then read into prow, JobStatesToReport will be `nil` (NOT `[]prowjob.ProwJobState{}`). This causes https://github.com/kubernetes/test-infra/blob/master/prow/apis/prowjobs/v1/types.go#L315 to get the setting from the parent.

The thing really confusing me: JobStatesToReport field has not changed for 12mos. https://github.com/kubernetes/test-infra/blame/69a9f34cd9c08cb1725d89e0d12f1ec2773f75b5/prow/apis/prowjobs/v1/types.go#L291

**What you expected to happen**:
I should be able to unset JobStatesToReport (or arrays in general) in this manner.

cc @chaodaiG 

A side note: `gopkg.in/yaml.v3` was recommended as a solution, but I do not think it is

Testing using different libraries:
```
    generate_test.go:37: k8s slack: {}
    generate_test.go:39: ghodss slack: {}
    generate_test.go:41: gopkg slack:
            channel: ""
            jobstatestoreport: []
            reporttemplate: ""
    generate_test.go:43: json {"slack":{}}
```

gopkg.in is wrong, IMO. It does not respect json tags on the struct. Therefor we end up with invalid YAML output - see jobstatestoreport, it isn't a valid ProwJob field

howardjohn:
Even with giving the prow job yaml with explicit empty list, this still does not work

chaodaiG:
Direct proof of roundtrip bug https://github.com/kubernetes/test-infra/pull/22900

chaodaiG:
Briefly explaining what's happening in #22900:

```
reporter_config:
  slack:
    job_states_to_report: []
```
Unmarshal into `JobBase`, and get:
```
JobBase
{
	ReporterConfig: &prowapi.ReporterConfig
	{
		Slack: &prowapi.SlackReporterConfig
		{
			JobStatesToReport: []prowapi.ProwJobState{},
		},
},
```
Marshal into yaml then unmarshal back into `JobBase`, get:
```
JobBase
{
[...]

PR #22897, opened 2021-07-15
Revert "skip_report flag in prowjob only used by crier reporters that report back to CI system"

howardjohn:
Reverts kubernetes/test-infra#22486

This has broken our job reporting (it's impossible to disable reporting for our large number of tests that we don't want reports since it is spammy). There is no way to disable it on a per job basis anymore due to https://github.com/kubernetes/test-infra/issues/22888


howardjohn:
https://github.com/kubernetes/test-infra/pull/22901

PR #22693, opened 2021-06-24
Crier slack reporter: use pointers for merging configs

chaodaiG:
This makes slack configuration using the same defaulting logic as other prow configurations. Will allow negating global slack configuration in job config.

PR #22901, opened 2021-07-15
Revert "skip_report flag in prowjob only used by crier reporters that…

chaodaiG:
… report back to CI system"

This reverts commit 1bc06354b804392dd08b7c51e91be025b8a3f81a.

This had caused regression in istio prow since the old behavior was used by them, specifically slack notification relies on `skip_report` for filtering out a small set of jobs.

Have tried moving forward by allowing empty `report_config` in prow job overriding global defautl as in https://github.com/kubernetes/test-infra/pull/22693 and https://github.com/kubernetes/test-infra/pull/22897. However it didn't work due to a roundtrip bug as mentioned in https://github.com/kubernetes/test-infra/issues/22888. Reverting for now, will redo this PR once the bug was fixed

### DIFF
--- a/prow/crier/controller.go
+++ b/prow/crier/controller.go
@@ -178,7 +178,7 @@ func (r *reconciler) reconcile(ctx context.Context, log *logrus.Entry, req recon
 
 	log = log.WithField("jobName", pj.Spec.Job)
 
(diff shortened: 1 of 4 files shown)


chaodaiG:
cc @howardjohn

PR #22907, opened 2021-07-15
Require container-image-build CI job

codenrhoden:
Following the merge of https://github.com/kubernetes-sigs/image-builder/pull/651, we can now turn this job on and require it to run when any of the following are changed:

- Makefile
- Dockerfile (or .dockerignore)
- Any of the scripts that install a dependency (because we call `make deps` in the docker build)

That should make sure we validate that a new container image can be built any time what is going *in* the container is changing. I'm not running the job whenever the Ansible roles change because we can't validate them at this step - the docker build is merely copying those files.

/assign @CecileRobertMichon 
@kkeshavamurthy


cpanato (approved):
/lgtm

/hold for @CecileRobertMichon

config/jobs/kubernetes-sigs/image-builder/image-builder-presubmits.yaml
```diff
@@ -129,9 +129,8 @@ presubmits:
     labels:
       preset-dind-enabled: "true"
       preset-kind-volume-mounts: "true"
-    always_run: false
-    optional: true
     decorate: true
+    run_if_changed: 'images/capi/(Makefile|Dockerfile|\.dockerignore)|images/capi/hack/ensure-(ansible|packer|jq|azure-cli|goss).*'
```
CecileRobertMichon, line 133: @codenrhoden given that we only have two data points on this job so far (https://prow.k8s.io/job-history/gs/kubernetes-jenkins/pr-logs/directory/pull-container-image-build) can we start by adding run_if_changed and keeping it optional and then switching it to required after ~ 1 week of stable data? Just being extra conservative given our recent CI flakes.

codenrhoden:
@CecileRobertMichon totally fair.

I've updated the spec, and updated the PR title, description, and commit message to reflect reality.

PR #22912, opened 2021-07-15
Fix triggering presubmit Prow jobs for Gerrit change if it's changed from draft to active

chizhg:
The reason for having this is there are two ways to change the Gerrit change from `draft` to `active`, and the messages are different, so we need to parse both messages in order to trigger presubmit Prow jobs correctly.

/cc @chaodaiG

### DIFF
--- a/prow/gerrit/client/client.go
+++ b/prow/gerrit/client/client.go
@@ -51,9 +51,12 @@ const (
 	// New status indicates a Gerrit change is new (ie pending)
 	New = "NEW"
 
-	// ReadyForReviewMessage is the message for a Gerrit change if it's changed
-	// from Draft to Active
-	ReadyForReviewMessage = "Set Ready For Review"
+	// ReadyForReviewMessage are the messages for a Gerrit change if it's changed
+	// from Draft to Active.
+	// This message will be sent if users press the `MARK AS ACTIVE` button.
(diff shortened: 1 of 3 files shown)


prow/gerrit/client/client.go
```diff
@@ -51,9 +51,12 @@ const (
 	// New status indicates a Gerrit change is new (ie pending)
 	New = "NEW"
 
-	// ReadyForReviewMessage is the message for a Gerrit change if it's changed
-	// from Draft to Active
-	ReadyForReviewMessage = "Set Ready For Review"
+	// ReadyForReviewMessage are the messages for a Gerrit change if it's changed
+	// from Draft to Active.
+	// This message will be sent if users press the `MARK AS ACTIVE` button.
+	ReadyForReviewMessage1 = "Set Ready For Review"
```
chaodaiG, line 57: nit: not very descriptive var names, how about `ReadyForReviewMessageFixed` and `ReadyForReviewMessageCustomizable`?

chaodaiG (approved):
/hold
left a nit

prow/gerrit/client/client.go
chizhg: Done.

chaodaiG (approved):
/unhold

PR #22915, opened 2021-07-16
Clean up upgrade/downgrade test configs

liggitt:
xref https://github.com/kubernetes/kubernetes/issues/103697

* Removes solidly failing downgrade tests
* Changes stable1 / beta tests to use stable1 / latest, since the latest marker file is actually maintained
  * if there was already a stable1 / latest test, drop the stable1 / beta test
* Renames testgrid tabs to match reality of what is being tested (so we don't link to "1.13-1.14" tabs, which look obsolete)
* Adds human-readable job descriptions

/assign @spiffxp


config/jobs/kubernetes/sig-cloud-provider/gcp/gpu/gcp-gpu-upgrade-downgrade.yaml
```diff
@@ -178,65 +178,4 @@ periodics:
 
   annotations:
     testgrid-dashboards: google-gce-upgrade
-    testgrid-tab-name: gce-gpu-1.13-master-master-upgrade
-- cron: "0 10-23/12 * * *"
```
wojtek-t, line 182: Why we're removing these?

config/jobs/kubernetes/sig-cloud-provider/gcp/upgrade-gce.yaml
```diff
@@ -1,7 +1,7 @@
 periodics:
 # GCE upgrades
 - interval: 2h
-  name: ci-kubernetes-e2e-gce-stable1-beta-upgrade-cluster
+  name: ci-kubernetes-e2e-gce-stable1-master-upgrade-cluster
```
wojtek-t, line 4: you use latest everywhere else instead of master

wojtek-t: Same in couple other places

config/jobs/kubernetes/sig-cloud-provider/gcp/gpu/gcp-gpu-upgrade-downgrade.yaml
liggitt: because the upgrade tests are the priority, and the downgrade tests are solidly failing (so no one is paying attention to the signal from them)

config/jobs/kubernetes/sig-cloud-provider/gcp/upgrade-gce.yaml
liggitt: fixed

wojtek-t:
/approve

Seems that bazel is failing.

liggitt:
found/fixed a duplicate config and added job descriptions

spiffxp (approved):
/approve
/lgtm

spiffxp:
For anyone following along, a consequence of this is that it looks like we "lost" history from some of our infra's perspective (like testgrid or triage).  Our infra doesn't understand the concept of renames, so they instead look like old jobs going away, and new jobs starting fresh with no history.  The old job data necessary to regenerate history is still present in GCS, but it's at an old location not associated with the new job.

This is why testgrid shows a bunch of jobs with no results right now
Screen Shot 2021-07-16 at 2 14 46 PM

wojtek-t:
@liggitt @spiffxp - actually, it seems that something might have gone wrong - we still don't have any runs of those 3 days later...

jberkus:
+1, downgrade tests have been poorly maintained to completely unmaintained since 1.13 or so.  Ultimately, this is a signal that our community doesn't really care about downgrades, at least not for core canonical Kubernetes.  If someone does care, maybe they'll step forwards to maintain the jobs.

liggitt:
> @liggitt @spiffxp - actually, it seems that something might have gone wrong - we still don't have any runs of those 3 days later...

that's... unexpected... did I miss something here?

wojtek-t:
This is strange, because clearly I see some runs in prow:
https://prow.k8s.io/?job=ci-kubernetes-e2e-gce-stable1-latest-upgrade-cluster

It seems like testgrid-related issue...

wojtek-t:
OK - it seems it works now

PR #22917, opened 2021-07-16
Replace U+2011 NON-BREAKING HYPHEN with regular hyphen in gerrit commit

mpherman2:
Gerrit commits do not recognize "&#x2011;" as NON-BREAKING HYPHEN, so this change just replaces it with a regular hyphen.

### DIFF
--- a/prow/cmd/generic-autobumper/bumper/bumper.go
+++ b/prow/cmd/generic-autobumper/bumper/bumper.go
@@ -473,7 +473,8 @@ func MakeGitCommit(remote, remoteBranch, name, email string, stdout, stderr io.W
 }
 
 func makeGerritCommit(summary, commitTag, changeId string) string {
-	return fmt.Sprintf("%s\n\n[%s]\n\nChange-Id: %s", summary, commitTag, changeId)
+	//Gerrit commits do not recognize "&#x2011;" as NON-BREAKING HYPHEN, so just replace with a regular hyphen.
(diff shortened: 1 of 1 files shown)


prow/cmd/generic-autobumper/bumper/bumper.go
```diff
@@ -473,7 +473,8 @@ func MakeGitCommit(remote, remoteBranch, name, email string, stdout, stderr io.W
 }
 
 func makeGerritCommit(summary, commitTag, changeId string) string {
-	return fmt.Sprintf("%s\n\n[%s]\n\nChange-Id: %s", summary, commitTag, changeId)
+	//Gerrit commits do not recognize "&#x2011;" as NON-BREAKING HYPHEN, so just replace with a regular hyphen.
```
fejta, line 476: What is creating the non-breaking hyphens? Can we make this tool stop doing that?

fejta (approved):
/hold

prow/cmd/generic-autobumper/bumper/bumper.go
mpherman2: https://github.com/kubernetes/test-infra/blob/05d1f8276331b067480d524e3343edfe0f410bde/prow/cmd/generic-autobumper/helper.go#L46 date formating that we were using for github commits to prevent line-wrapping. We could also just remove this if we do not care about line wrapping in the github commits.

PR #22918, opened 2021-07-16
Remove omitempty from job_state_to_report

chaodaiG:
As discussed in https://github.com/kubernetes/test-infra/issues/22888, it's intended behavior that omitempty treats empty slice as nil, which makes it not possible to differentiate nil from empty slice. While we need this for allowing overriding global default with empty slice, removing omitempty from job_state_to_report

### DIFF
--- a/prow/apis/prowjobs/v1/types.go
+++ b/prow/apis/prowjobs/v1/types.go
@@ -288,7 +288,7 @@ type ReporterConfig struct {
 type SlackReporterConfig struct {
 	Host              string         `json:"host,omitempty"`
 	Channel           string         `json:"channel,omitempty"`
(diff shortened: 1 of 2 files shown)


chaodaiG:
```
--- FAIL: TestWhoCanUse (0.00s)
    override_test.go:952: expected Repo administrators, and the following github teams:org1/repo1: team-foo team-bar, org1: team-foo-bar., got Repo administrators, and the following github teams:org1: team-foo-bar, org1/repo1: team-foo team-bar.
```

have seen it more than once lately, feels like flaky

/retest

alvaroaleman (approved):
failing test is due to map ordering in an unrelated test, ill submit a follow-up to fix it
/retest

PR #22921, opened 2021-07-16
Add allow-same-origin to deck/spyglass sandbox

### LINKED ISSUE #22417: spyglass: missing allow-same-origin breaks viewer(s)
**What happened**:

The coverage result treemap viewer is broken due to:

> Uncaught SecurityError: Sandbox access violation: Blocked a frame at "null" from accessing a frame at "https://prow.k8s.io".  The frame requesting access is sandboxed and lacks the "allow-same-origin" flag.

Deck/spyglass should function without errors. This is breaking the ability to click through the treemap, probably other things.

Visit prow.k8s.io, inspect a job results page.

https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/ci-kubernetes-coverage-unit/1400476832872009728

/area prow
/area prow/deck
/area prow/spyglass

fgimenez:
Fixes #22417

Not sure what to do about https://github.com/kubernetes/test-infra/blob/master/prow/spyglass/write-a-lens.md#sandboxing, will this change affect the ability to access the parent window?

/cc @BenTheElder 

### DIFF
--- a/prow/cmd/deck/template/spyglass.html
+++ b/prow/cmd/deck/template/spyglass.html
@@ -41,7 +41,7 @@
     <div class="mdl-card__title lens-title"><h3 class="mdl-card__title-text">{{$config.Title}}</h3></div>
     <div id="{{$config.Name}}-view-container" class="lens-view-content mdl-card__supporting-text">
       <img src="/static/kubernetes-wheel.svg" alt="loading spinner" class="loading-spinner is-active lens-card-loading" id="{{$config.Name}}-loading">
-      <iframe class="lens-container" style="visibility: hidden;" id="iframe-{{$index}}" sandbox="allow-scripts allow-top-navigation allow-popups" data-lens-index="{{$index}}" data-lens-name="{{$config.Name}}"{{if $config.HideTitle}} data-hide-title="true"{{end}}></iframe>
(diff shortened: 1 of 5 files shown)


BenTheElder:
/ok-to-test

I'm not sure re: sandboxing, it's not my usual domain 😅 

I'm also not sure who currently working in the project would know more.

alvaroaleman (approved):
We control what we put into the iframe so this should be fine. Not a frontend specialist though.
/assign @cjwagner 
/hold

cjwagner:
/assign @e-blackwelder 
Could you please help review this PR, Ethan? I'm only vaguely familiar with how origin policies work and I don't feel comfortable approving without the eyes of a SME.

Katharine:
Since I'm looking at this and thinking about it, noting this for anyone who cares:

Having both `allow-scripts` and `allow-same-origin` removes any security properties provided by the sandbox: `allow-same-origin` allows the frame to modify its parent, which enables it to remove the sandbox properties. This is probably not important right now, but may matter if you ever enable remote lenses.

The other perhaps-unintentional effect `allow-same-origin` has is enabling access to APIs not explicitly provided to lenses, potentially resulting in surprise maintenance burdens. As designed, this was not supposed to be permitted — though I can't speak to whether you still care.

PR #22924, opened 2021-07-16
Prow sub: fix a runtime npe

chaodaiG:
Previously this was developed for jobs that all have labels, so it never happened. As this function being used by more job types, such as presubmit and postsubmit, there are cases where labels are nil


chaodaiG:
/cc @alvaroaleman 
cc @2nd-half

alvaroaleman (approved):

PR #22928, opened 2021-07-16
Lint for exported loop var pointers and fix occurences

alvaroaleman:
Finds issues like https://github.com/kubernetes/test-infra/pull/22926 (and did in fact find another issue in the same file).

Only drawback is that it complains when such a construct is followed by `break`, even though that is valid.

/assign @chaodaiG


chaodaiG (approved):
Awesome

PR #22933, opened 2021-07-19
Add presubmit Prow jobs for kubetest2 gke deployer

chizhg:
Fixes https://github.com/kubernetes-sigs/kubetest2/issues/152

### DIFF
--- a/config/jobs/kubernetes-sigs/kubetest2/kubetest2-gke-presubmits.yaml
+++ b/config/jobs/kubernetes-sigs/kubetest2/kubetest2-gke-presubmits.yaml
@@ -1,0 +1,32 @@
+presubmits:
+  kubernetes-sigs/kubetest2:
+  - name: gke-up-down-singlecluster
+    decorate: true
+    run_if_changed: "kubetest2-gke/.*"
+    labels:
+      preset-service-account: "true"
+      preset-k8s-ssh: "true"
+    spec:
+      containers:
+      - image: gcr.io/k8s-testimages/kubekins-e2e:v20210715-521d667-master
+        command:
+        - "runner.sh"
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes-sigs/kubetest2/kubetest2-gke-presubmits.yaml
```diff
@@ -0,0 +1,42 @@
...
+      containers:
+      - image: gcr.io/k8s-testimages/kubekins-e2e:v20210715-521d667-master
+        command:
+        - "runner.sh"
+        args:
+        - "./kubetest2-gke/ci-tests/buildupdown.sh"
+        - "--cluster-topology"
+        - "mc"
+        - "--build-strategy"
+        - "gke_make"
```
amwat, line 40: let's drop --build in upstream jobs since we won't have the sources.

chizhg: Got it, changed to `bazel`, if it is supported for OSS.

amwat: bazel is definitely not supported. I think it's fine to just drop build altogether and just have up down

chizhg: Thanks for the clarification. Removed.

amwat (commented):
/lgtm
/approve

/hold
do we want to start with a non blocking job and then make it blocking?

chizhg:
/unhold
I'll monitor the job runs and fix them as soon as I find any issues.

PR #22934, opened 2021-07-19
Use systemd cgroup driver with containerd cgroupv2 Tests

adisky:
runc recommends to use systemd cgroup driver with cgroupv2
https://github.com/opencontainers/runc/blob/master/docs/cgroup-v2.md#systemd


adisky:
Marking as WIP, since containerd config also needs to be updated to use systemd cgroup driver.

adisky:
cc @dims

adisky:
/unhold
containerd PR is merged

PR #22935, opened 2021-07-19
don't run test with ENABLE_POD_SECURITY_POLICY

aojea:
Quoting from Jordan Liggitt
" PodSecurityPolicy is optional and deprecated.
We shouldn't be enabling it in new jobs.
No e2e or addon should fail if it's not enabled."

Remove the ENABLE_POD_SECURITY_POLICY=true option
from all jobs that run against master branch

/assign @liggitt @dims @spiffxp @BenTheElder @tallclair

### DIFF
--- a/config/jobs/kubernetes/generated/generated.yaml
+++ b/config/jobs/kubernetes/generated/generated.yaml
@@ -95,7 +95,6 @@ periodics:
       - --timeout=120m
       - --test_args=--ginkgo.skip=\[Slow\]|\[Serial\]|\[Disruptive\]|\[Flaky\]|\[Feature:.+\] --minStartupPods=8
       - --ginkgo-parallel=30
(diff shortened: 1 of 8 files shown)


spiffxp:
/cc @mkorbi @savitharaghunathan 
FYI @kubernetes/ci-signal @kubernetes/release-team-leads 

I'm going to allow this to merge because I think it'd be in the interest of our users to catch sooner-vs-later that we have tests relying on something that has had its deprecation clock started.  If we find this causes a massive delta in CI Signal, we can revert.

In terms of jobs you watch, it's going to impact:
- ci-kubernetes-e2e-gci-gce 
- ci-kubernetes-e2e-ubuntu-gce
- ci-kubernetes-e2e-ubuntu-gce-containerd
- pull-kubernetes-e2e-gce-ubuntu

spiffxp (approved):
/approve
/lgtm

PR #22939, opened 2021-07-19
run peridoc sig-network tests every 6 hours

aojea:
I was going to change only the ingress jobs but I've found that we are running a lot of periodic jobs with very short intervals.

Running each job every 6 hours is more than enough, there are also jobs that take more than the interval specified


MrHohn (approved):
/lgtm
/approve

PR #22941, opened 2021-07-19
remove superfluous skip patterns

danwinship:
#22734 rewrote some skip rules but the new pattern doesn't match anything and so should just be removed.

(OK, fine, it _currently_ matches a few tests, but those tests were supposed to have had the `[Feature:SCTP]` removed from them after SCTP went GA, and I'm about to fix that, and _then_ the pattern here won't match anything.)

/cc @knabben @aojea

### DIFF
--- a/config/jobs/kubernetes/sig-network/sig-network-misc.yaml
+++ b/config/jobs/kubernetes/sig-network/sig-network-misc.yaml
@@ -121,7 +121,7 @@ presubmits:
         - --provider=gce
         # Skipping snat tests probably GCE related? https://github.com/kubernetes/test-infra/issues/20321
         # Skipping Cloud Provider specific tests: LoadBalancer, ESIPP (Source IP Preservation)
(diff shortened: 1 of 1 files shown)


danwinship:
https://github.com/kubernetes/kubernetes/pull/103778 removes the remaining `[Feature:SCTP]` tags

aojea:
> [kubernetes/kubernetes#103778](https://github.com/kubernetes/kubernetes/pull/103778) removes the remaining `[Feature:SCTP]` tags

should we wait for it , then?

danwinship:
No, we don't want to be skipping these tests, so it doesn't matter whether we stop skipping them because we stop skipping the tag or because the tag goes away. (But both of those should happen.)

aojea:
let's wait until 1.23 opens then

danwinship:
ok, 1.23 is open, and the `[Feature:SCTP]` tags also no longer exist

danwinship:
/remove-lifecycle stale
@aojea can you `/approve`?

aojea:
/approve
/lgtm
👀

PR #22948, opened 2021-07-20
chore: correct gcp-secrets-store-cred paths

tam7t:
From the docs I had these backwards: https://github.com/kubernetes/test-infra/blob/master/prow/prow_secrets.md#usage-prow-clients

```
  - key: <my-gsm-secret-name>     # name of the GCP secret
    name: <my-kubernetes-secret-name>   # key name in the k8s secret
```

Intent is to load `projects/secretmanager-csi-build/secrets/k8s-oss-prow/versions/latest` to K8s secret `test-pods/gcp-secrets-store-cred` at key `key.json`


aramase (commented):
/lgtm

PR #22968, opened 2021-07-21
Remove unsupported 1.18 jobs

xing-yang:


config/jobs/kubernetes-csi/external-provisioner/external-provisioner-config.yaml
```diff
@@ -2,102 +2,6 @@
 
 presubmits:
   kubernetes-csi/external-provisioner:
-  - name: pull-kubernetes-csi-external-provisioner-1-18-on-kubernetes-1-18
```
msau42, line 5: 1.18 should be removed from gen-jobs.sh, and then all the files should be updated.

xing-yang: ok

PR #22980, opened 2021-07-22
Update perf-tests branch for 1.21 and 1.22 release jobs

marseel:
/assign @wojtek-t

### DIFF
--- a/config/jobs/kubernetes/sig-release/release-branch-jobs/1.21.yaml
+++ b/config/jobs/kubernetes/sig-release/release-branch-jobs/1.21.yaml
@@ -288,7 +288,7 @@ periodics:
     org: kubernetes
     path_alias: k8s.io/kubernetes
     repo: kubernetes
-  - base_ref: master
+  - base_ref: release-1.21
     org: kubernetes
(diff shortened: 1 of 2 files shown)


config/jobs/kubernetes/sig-release/release-branch-jobs/1.21.yaml
```diff
@@ -288,7 +288,7 @@ periodics:
     org: kubernetes
     path_alias: k8s.io/kubernetes
     repo: kubernetes
-  - base_ref: master
+  - base_ref: release-1.21
```
wojtek-t, line 291: This problem is related to: https://github.com/kubernetes/test-infra/issues/17097

Can we not just fix the actual jobs, but address that bug too to avoid those issues in the future?

marseel: I believe @jupblb is looking into it currently.

marseel: I've opened PR for fixing forker: #22984

PR #22986, opened 2021-07-22
add postsubmit job to push kind images

aojea:


aojea:
/assign @BenTheElder 
/hold
we need to get the cloudbuild.yaml in kind

BenTheElder:
I think some questions remain about how we're going to actually publish and consume these. cc @tao12345666333 

I think we would have to use the image promotion process and figure out mirroring. Or we could maybe add a secret to GCB in the project for a dockerhub robot.

BenTheElder:
Image promotion might be the way to go, but it will create extra work we need to be aware of and it will reduce availability of these images for development + building node images.

aojea:
> I think some questions remain about how we're going to actually publish and consume these. cc @tao12345666333

that is a next step, this step is to avoid maintainers to have to create the images locally ... we don't have to change the whole process at once, at least we automate the more time consuming process

aojea:
/assign @BenTheElder 

so you can copy paste or redo

BenTheElder:
I think first thing we need to go add the cloudbuild.yaml files?

config/jobs/image-pushing/k8s-staging-kind.yaml
```diff
@@ -25,3 +25,81 @@ postsubmits:
       rerun_auth_config:
         github_team_ids:
           - 2921947 # kind-maintainers
+    - name: post-kind-push-base-image
+      cluster: k8s-infra-prow-build-trusted
+      run_if_changed: '^images\/base'
+      annotations:
+        testgrid-dashboards: sig-testing-kind, wg-k8s-infra-gcb
+        testgrid-alert-email: kubernetes-engprod+alerts@google.com,antonio.ojea.garcia@gmail.com
```
BenTheElder, line 33: we should stop sending it to this alias, kubernetes-engprod isn't staffing kind. we can just put bentheelder@google.com

config/jobs/image-pushing/k8s-staging-kind.yaml
```diff
@@ -25,3 +25,81 @@ postsubmits:
...
+        serviceAccountName: gcb-builder
+        containers:
+          - image: gcr.io/k8s-testimages/image-builder:v20210622-762366a
+            command:
+              - /run.sh
+            args:
+              - --project=k8s-staging-kind
+              - --scratch-bucket=gs://k8s-staging-kind-gcb
+              - --env-passthrough=PULL_BASE_SHA
+              - --build-dir=.
```
BenTheElder, line 46: we want --with-git-dir

config/jobs/image-pushing/k8s-staging-kind.yaml
```diff
@@ -25,3 +25,81 @@ postsubmits:
       rerun_auth_config:
         github_team_ids:
           - 2921947 # kind-maintainers
+    - name: post-kind-push-base-image
+      cluster: k8s-infra-prow-build-trusted
+      run_if_changed: '^images\/base'
```
BenTheElder, line 30: this is missing the common makefile, not sure what else

BenTheElder:
I'm going to fork this and update for https://github.com/kubernetes-sigs/kind/pull/3125

BenTheElder:
updated in https://github.com/kubernetes/test-infra/pull/29048 w/ co-authorship

PR #22988, opened 2021-07-22
Add ProwJobDefault config with TenantID

mpherman2:
This is the first part of the work towards Private Repo multi-tenancy. TenantID will be used to mark prowjobs with an ID that will be recognized by Deck. In order to make assigning Tennant IDs to every Prowjob easy, we mimc the behavior of DefaultDecorationConfig for Plank. This will allow users to mark All Prowjobs belonging to a given tenant by Cluster/OrgRepo in the ProwConfig.


prow/config/prow-config-documented.yaml
```diff
@@ -904,6 +904,31 @@ plank:
 pod_namespace: ' '
 
 
+# ProwJobDefaultEntries holds a list of defaults for specific values
+# Each entry in the slice specifies Repo and CLuster regexp filter fields to
+# match against the jobs and a corresponding ProwJobDefault . All entries that
+# match a job are used. Later matching entries override the fields of earlier
+# matching entires.
+prowjob_default_entries:
```
fejta, line 912: I would make it more obvious that cluster and repo are filter clauses that then apply a tenant_id

prow/config/prow-config-documented.yaml
```diff
@@ -904,6 +904,31 @@ plank:
...
+    # by sequentially merging with later entries overriding fields from earlier
+    # entries.
+    config:
+        tenant_id: ' '
+
+    # OrgRepo matches against the "org" or "org/repo" that the presubmit or postsubmit
+    # is associated with. If the job is a periodic, extra_refs[0] is used. If the
+    # job is a periodic without extra_refs, the empty string will be used.
+    # If this field is omitted all jobs will match.
+    repo: ' '
```
fejta, line 929: No interest in `repos` or `clusters`? It seems like this will be a common scenario.

prow/config/prow-config-documented.yaml
```diff
@@ -904,6 +904,31 @@ plank:
 pod_namespace: ' '
 
 
+# ProwJobDefaultEntries holds a list of defaults for specific values
+# Each entry in the slice specifies Repo and CLuster regexp filter fields to
+# match against the jobs and a corresponding ProwJobDefault . All entries that
+# match a job are used. Later matching entries override the fields of earlier
```
fejta, line 910: This seems prone to surprising behavior (aka today's job gets sent tomorrow to a new and wrong tenant because someone adds a new item to this list).

Did you consider ensuring that list items to not conflict with each other?

fejta (commented):
looking good generally

prow/config/prow-config-documented.yaml
mpherman2: I want to keep the naming consistent with `default_decoration_config_entries` as the behavior is the same.

prow/config/prow-config-documented.yaml
mpherman2: Same as above. I want to keep this consistent with `default_decoration_config_entries` There is already a fine way to do this as you can have multiple entries with the same tenant_id, but I don't believe this will be that common of a usecase.

prow/config/prow-config-documented.yaml
mpherman2: We actually want some list items to conflict with each other in some cases. For example myOrg has a default tenantID: foo but myOrg/publicRepo has tenantID: bar.

We will want to add presubmit validation to the config to ensure that the tenantID for each prowjob is what we are expecting for each tenant to increase confidence that we don't accidentally change a private prowjob to a public tenantID

prow/config/config.go
```diff
@@ -559,11 +586,37 @@ type DefaultDecorationConfigEntry struct {
 	Config *prowapi.DecorationConfig `json:"config,omitempty"`
 }
 
+func matches(OrgRepo, givenCluster, repo, cluster string) bool {
```
chaodaiG, line 589: rename `OrgRepo` with `orgRepo`

prow/config/config.go
```diff
@@ -535,6 +542,26 @@ type Plank struct {
...
+
+	// OrgRepo matches against the "org" or "org/repo" that the presubmit or postsubmit
+	// is associated with. If the job is a periodic, extra_refs[0] is used. If the
+	// job is a periodic without extra_refs, the empty string will be used.
+	// If this field is omitted all jobs will match.
+	OrgRepo string `json:"repo,omitempty"`
+	// Cluster matches against the cluster alias of the build cluster that the
+	// ProwJob is configured to run on. Recall that ProwJobs default to running on
+	// the "default" build cluster if they omit the "cluster" field in config.
+	Cluster string `json:"cluster,omitempty"`
```
cjwagner, line 556: nit: Since the matching fields are the same between `ProwJobDefaultEntry` and `DefaultDecorationConfigEntry` we could stick them into their own struct and make that an embedded struct of them both. That would allow the `matches(...)` function to become a method of the matcher struct and the two existing `matches(...)` methods could be omitted.

This is a minor change and there might be something I'm overlooking that makes you implementation better. Just something to consider to make this a bit more idiomatic.

prow/config/prow-config-documented.yaml
cjwagner: Configuration (like `tenant_id`) is distinguished from filter fields (`cluster` and `repo`) by being nested under `config`.

prow/config/prow-config-documented.yaml
cjwagner: +1 for consistency. Additional matching fields can be added as needed in the future with this design too.

prow/config/config.go
mpherman2: I think I might misunderstand how embedding works. My understanding is that it would make the current uses of DefaultDecorationConfig not backwards compatible because they would need to define `DefaultDecorationConfigEntry{ Matcher: { OrgRepo... } }` instead of `DefaultDecorationConfigEntry{ OrgRepo: ... }` Am I missing something?

prow/config/config.go
mpherman2: Done

prow/apis/prowjobs/v1/types.go
```diff
@@ -179,6 +179,10 @@ type ProwJobSpec struct {
 	// Presubmits and Postsubmits can also be set to hidden by
 	// adding their repository in Decks `hidden_repo` setting.
 	Hidden bool `json:"hidden,omitempty"`
+
+	// ProwJobDefault holds configuration options provided as defaults
+	// in the Prow config
+	ProwJobDefault *ProwJobDefault `json:"prowjob_defaults,omitempty"`
```
alvaroaleman, line 185: Nit: I find this name confusing, why not directly have a `tenant_id` field? We can use a string pointer to differentiate empty string from unset

prow/config/config.go
cjwagner: Here is an example in the go playground: https://play.golang.org/p/4J79nWGT21L
It will be compatible with existing config, but I think existing definitions in code would need to be updated. It doesn't make a difference to users to if that is a pain to switch then don't worry about it.

prow/config/config.go
```diff
@@ -180,6 +180,13 @@ type ProwConfig struct {
 	// ManagedWebhooks contains information about all github repositories and organizations which are using
 	// non-global Hmac token.
 	ManagedWebhooks ManagedWebhooks `json:"managed_webhooks,omitempty"`
+
+	// ProwJobDefaultEntries holds a list of defaults for specific values
+	// Each entry in the slice specifies Repo and CLuster regexp filter fields to
+	// match against the jobs and a corresponding ProwJobDefault . All entries that
+	// match a job are used. Later matching entries override the fields of earlier
+	// matching entires.
+	ProwJobDefaultEntries []*ProwJobDefaultEntry `json:"prowjob_default_entries,omitempty"`
```
alvaroaleman, line 189: Similiarly, why is this not simply called `TenantIDEntry`? We have a huge config and having an an option somewhere that is called `DefaultEntry` is pretty unclear about what it is.

Do we plan to extend this with other config options? If so, maybe re-use/extend the existing Preset mechanism?

prow/config/config.go
```diff
@@ -535,6 +542,26 @@ type Plank struct {
...
 }
 
+type ProwJobDefaultEntry struct {
+	// Matching/filtering fields. All filters must match for an entry to match.
+
+	// OrgRepo matches against the "org" or "org/repo" that the presubmit or postsubmit
+	// is associated with. If the job is a periodic, extra_refs[0] is used. If the
+	// job is a periodic without extra_refs, the empty string will be used.
+	// If this field is omitted all jobs will match.
+	OrgRepo string `json:"repo,omitempty"`
```
alvaroaleman, line 552: Nit: maybe `OrgOrRepo` and also make the json tag reflect that name?

prow/apis/prowjobs/v1/types.go
mpherman2: The point of not having a tenant_id field and instead using this default configuration is so that it will be easy to set tenantIDs based on cluster or org/repo the same way we do for defaultDecorationConfig. This design was based on default decoration config. Open to suggestions about variable naming.

prow/config/config.go
mpherman2: We want to leave it open to extend with other options as being able to set prowjob fields based on cluster or org/repo is useful. We can't use defaultDecorationConfig as it is not useable by all prowjobs, and we need tenantID (as well as other future fields) to be applied to all prowjobs.

prow/config/config.go
mpherman2: Keeping OrgRepo as the behavior is the same as defaultDecorationConfig and I want the naming to be consistent

prow/config/config.go
mpherman2: I think this is the correct thing to do, but this PR is blocking other work so I left a todo to clean it up later.

prow/apis/prowjobs/v1/types.go
cjwagner: It is also extensible for other ProwJob fields we may want to include in the future as we can do with the default decoration configs.

alvaroaleman: Can we then maybe embedd and inline it?

It also seems to be a bit of a premature optimization for me. It  will only be useful if we have other config options that we want to set based on the exact same matching logic as `tenant_id`.

cjwagner: Embedding was avoided to differentiate matching fields from configuration.

The majority of prow configuration defaulting logic is currently handled by org/repo indexing or cluster indexing so it seems likely to me that new fields would be satisfied by the existing matching fields. This is easily extensible to support more matching fields if needed.

mpherman2:
@alvaroaleman @cjwagner Anything blocking this?

cjwagner:
No blockers from me.  Lets hear from Alvaro before merging to ensure we've satisfied his review comments.

alvaroaleman (approved):

PR #22991, opened 2021-07-22
azure: remove 1.18 jobs and add 1.22 jobs

chewong:

/assign @CecileRobertMichon

### DIFF
--- a/config/jobs/kubernetes/sig-cloud-provider/azure/release-1.22.yaml
+++ b/config/jobs/kubernetes/sig-cloud-provider/azure/release-1.22.yaml
@@ -8,7 +8,7 @@ presubmits:
     run_if_changed: 'azure.*\.go'
     path_alias: k8s.io/kubernetes
     branches:
-      - release-1.18
+      - release-1.22
     labels:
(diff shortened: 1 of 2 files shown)


config/jobs/kubernetes/sig-cloud-provider/azure/release-1.22.yaml
```diff
@@ -8,7 +8,7 @@ presubmits:
     run_if_changed: 'azure.*\.go'
     path_alias: k8s.io/kubernetes
     branches:
-      - release-1.18
+      - release-1.22
```
CecileRobertMichon, line 11: we don't need to update generate.sh?

chewong: We don't need to because the script takes a list of versions as arguments and generate a YAML file for each version (i.e. `./config/jobs/kubernetes/sig-cloud-provider/azure/generate.sh 1.19 1.20 1.21 1.22 master`.

CecileRobertMichon (commented):
/lgtm

Issue #12296, opened 2019-04-21
Use k8s-triage-robot in lieu of fejta-bot for kubernetes issue/pr commenter jobs


nikhita:
Background: @fejta-bot automatically marks issues as stale/rotten and closes them if they don't have any activity for certain number of days. It is an unprivileged user and is not a member of any of the Kubernetes GitHub orgs.

While enabling fejta-bot automation across all Kubernetes GitHub orgs, it was suggested to rename it to something "official": https://groups.google.com/d/topic/kubernetes-dev/9ps8uNXJ10k/discussion

This issue is to track discussing renaming fejta-bot -  what to rename it as or if we should use another account entirely - and making sure that this is rolled out across orgs.

/sig contributor-experience
/sig testing
/cc @cblecker @fejta @spiffxp

fejta:
`serious-fejta-bot` with a grumpy cat partially converted into a robot for it's profile pic

nikhita:
/assign

Assigning it to myself to try to move this along in contribex

stevekuznetsov:
Any reason not to use one of the existing robots?

spiffxp:
We need prow to react to the comments posted by these jobs, so it can't be @k8s-ci-robot.  We could maybe use @k8s-github-robot, but I would prefer using a bot that has no privileges.

spiffxp:
You'll want to rename more than the account:
- https://testgrid.k8s.io/sig-testing-fejtabot
- https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/test-infra/fejta-bot-periodics.yaml

Things @fejta-bot does:
- /lifecycle stale | /lifecycle rotten | /close
- /retest
- /check-cla
- comment that API review may be needed
- create github issues

Naming is hard, WDYT of either of these?
- @k8s-triage-robot
- @k8s-commenter-robot

chris-short:
k8s-cuddle-bot
k8s-repo-sanity-bot
k8s-triage-bot

justaugustus:
+1 to k8s-triage-[ro]bot!
Can probably dovetail with some of the stuff mentioned here: https://github.com/kubernetes/community/issues/3456

dims:
+1 to k8s-triage-bot

InAnimaTe:
fajita-bot

nikhita:
+1 to k8s-triage-robot. Since we have `-robot` suffixes for our other bot accounts, using `-robot` feels consistent.

[...]

PR #22992, opened 2021-07-22
config/jobs: add ci-k8s-triage-robot-testcomment

spiffxp:
Part of https://github.com/kubernetes/test-infra/issues/12296

Followup to use token added in https://github.com/kubernetes/k8s.io/pull/2389


spiffxp:
The intent is to have this leave a single comment on https://github.com/kubernetes/test-infra/issues/11402

PR #22994, opened 2021-07-22
Adding duration timers to the Config initialization

bradmwilliams:
We have observed some very long processing time when calling `configAgent, err = o.prowconfig.ConfigAgent()`.  This PR adds a couple of trace messages, to see how many files are being processed and how long it's taking, to the 2 calls that are actually processing the config files.

### DIFF
--- a/prow/config/config.go
+++ b/prow/config/config.go
@@ -1201,15 +1201,17 @@ func ReadJobConfig(jobConfig string) (JobConfig, error) {
 			return err
 		}
 		jc, err = mergeJobConfigs(jc, subConfig)
-		logrus.Tracef("jobConfig: %s loaded in: %v", path, time.Now().Sub(begin))
-		jobConfigCount++
+		if err == nil {
+			logrus.WithField("jobConfig", path).WithField("duration", time.Now().Sub(begin)).Traceln("config loaded")
+			jobConfigCount++
+		}
 		return err
 	})
-	logrus.Tracef("Processed: %d jobConfigs in: %v\n", jobConfigCount, time.Now().Sub(start))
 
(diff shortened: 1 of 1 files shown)


alvaroaleman (approved):

prow/config/config.go
```diff
@@ -1193,17 +1195,23 @@ func ReadJobConfig(jobConfig string) (JobConfig, error) {
...
 		uniqueBasenames.Insert(base)
 
+		begin := time.Now()
 		var subConfig JobConfig
 		if err := yamlToConfig(path, &subConfig); err != nil {
 			return err
 		}
 		jc, err = mergeJobConfigs(jc, subConfig)
+		if err == nil {
+			logrus.WithField("jobConfig", path).WithField("duration", time.Now().Sub(begin)).Traceln("config loaded")
```
petr-muller, line 1205: `time.Now().Sub()` is `time.Since()`

prow/config/config.go
```diff
@@ -1193,17 +1195,23 @@ func ReadJobConfig(jobConfig string) (JobConfig, error) {
 		}
 		uniqueBasenames.Insert(base)
 
+		begin := time.Now()
```
petr-muller, line 1198: could we say something like `allStart` + `fileStart` instead of `start/begin` for reability?

prow/config/config.go
bradmwilliams: Updated all references!

prow/config/config.go
bradmwilliams: Done!

petr-muller (approved):
/ok-to-test

PR #22997, opened 2021-07-23
config/jobs: use k8s-triage-robot instead of fejta-bot

spiffxp:
Fixes: https://github.com/kubernetes/test-infra/issues/11402

/hold
For a notice to get sent out (which I can do tomorrow if nobody else wants to)


config/jobs/kubernetes/wg-k8s-infra/trusted/sig-contribex-k8s-triage-robot.yaml
```diff
@@ -1,30 +1,122 @@
...
+    description: Closes rotten issues after 30d of inactivity
+    testgrid-tab-name: close
+  spec:
+    containers:
+    - image: gcr.io/k8s-prow/commenter:v20210722-64d720b97e
+      command:
+      - /app/robots/commenter/app.binary
+      args:
+      - |-
+        --query=org:kubernetes
```
nikhita, line 103: I'm also good with this as-is but if we want to do a test run first, we could probably add a new job for triage-robot targeting only specific repos instead of the whole org.

spiffxp:
/hold cancel
Notification sent: https://groups.google.com/g/kubernetes-dev/c/oD_ijk7jpa8/m/f24rGVYyAwAJ

config/jobs/kubernetes/wg-k8s-infra/trusted/sig-contribex-k8s-triage-robot.yaml
spiffxp: I'm willing to tackle whatever noise comes from this not working as expected.  I'm reasonably confident that https://github.com/kubernetes/test-infra/pull/22992 caught all of the issues related to commenting

PR #23000, opened 2021-07-23
Conditionally use new conformance image directory

liggitt:
xref https://github.com/kubernetes/kubernetes/pull/103874 and https://github.com/kubernetes/kubernetes/issues/103867

/cc @BenTheElder @spiffxp @dims


spiffxp (approved):
/approve
/lgtm

BenTheElder (approved):

PR #23004, opened 2021-07-23
prow: make the secret agent a singleton

stevekuznetsov:
In parallel tests, we noticed a data race between the
`logrus.SetFormatter` call and calls to `logrus.Warn` in the secret
agent startup code. While that particular race is a logrus correctness
error, as the global formatter should be threadsafe, we noticed that
furthermore it is not safe to have more than one secret agent in a
process, as they race to set the formatter (for censoring output). There
is no correct way to have more than one agent, so it must be a
singleton. This patch makes it so.


maintenance/migratestatus/migratestatus.go
```diff
@@ -122,14 +121,7 @@ func main() {
 		logrus.WithError(err).Fatal("Invalid options")
 	}
 
-	secretAgent := &secret.Agent{}
-	if o.github.TokenPath != "" {
```
stevekuznetsov, line 126: The GitHub flag options does this for us, we never need to worry about it.

prow/cmd/branchprotector/protect.go
```diff
@@ -136,12 +136,11 @@ func main() {
 	cfg := ca.Config()
 	cfg.BranchProtectionWarnings(logrus.NewEntry(logrus.StandardLogger()), cfg.PresubmitsStatic)
 
-	secretAgent := &secret.Agent{}
-	if err := secretAgent.Start(nil); err != nil {
```
stevekuznetsov, line 140: Calls like this were just an initialization and have no effect with the singleton since it's initialized at program start

prow/config/secret/agent.go
```diff
@@ -39,37 +51,35 @@ type Agent struct {
...
 		return err
 	}
 
-	a.secretsMap = secretsMap
-	a.ReloadingCensorer = secretutil.NewCensorer()
-	a.refreshCensorer()
-
-	logrus.SetFormatter(logrusutil.NewFormatterWithCensor(logrus.StandardLogger().Formatter, a.ReloadingCensorer))
+	for key, value := range secretsMap {
+		agent.setSecret(key, value)
```
stevekuznetsov, line 61: Technically this ends up with us calling `refreshCensorer()` N times instead of once, as it was before, but I am not sure we can do better when multiple `Start()` calls may be happening at once and interleaved

prow/config/secret/agent.go
```diff
@@ -39,37 +51,35 @@ type Agent struct {
 // Start creates goroutines to monitor the files that contain the secret value.
 // Additionally, Start wraps the current standard logger formatter with a
 // censoring formatter that removes secret occurrences from the logs.
-func (a *Agent) Start(paths []string) error {
+func Start(paths []string) error {
```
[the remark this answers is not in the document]
stevekuznetsov: Yep, good point - I didn't remove it to keep things minimal, but I will go ahead and do this

stevekuznetsov:
OK, edited to make a v2 package

Issue #23003, opened 2021-07-23
config/jobs: missing e2e jobs for kubernetes release branches <= release-1.21


spiffxp:
**What happened**:

I can't find a "default gce e2e" tab on:
- https://testgrid.k8s.io/sig-release-1.21-blocking
- https://testgrid.k8s.io/sig-release-1.20-blocking
- https://testgrid.k8s.io/sig-release-1.19-blocking

**What you expected to happen**:

To find an equivalent to:
- https://testgrid.k8s.io/sig-release-1.22-blocking#gce-cos-k8sbeta-default
- https://testgrid.k8s.io/sig-release-master-blocking#gce-cos-master-default

Or maybe to:
- https://testgrid.k8s.io/sig-release-master-blocking#gce-ubuntu-master-containerd

I forget when the ubuntu job was moved to release-blocking

**How to reproduce it (as minimally and precisely as possible)**:
Identify which of the tabs listed on https://testgrid.k8s.io/sig-release-1.21-blocking#Summary provide equivalent coverage to https://testgrid.k8s.io/sig-release-1.22-blocking#gce-cos-k8sbeta-default

**Please provide links to example occurrences, if any**:
See above

**Anything else we need to know?**:
The jobs aren't on the release-informing dashboards either

/area release-eng
/sig release
/priority critical-urgent

spiffxp:
/assign @liggitt @puerco 
I don't know for sure that either of your recent PR's did this, but I'm giving y'all the heads up

spiffxp:
/remove-priority critical-urgent
/priority important-longterm

This is me rediscovering that the gce-default jobs were moved to -informing because they were too flaky back on 2019-10-07
- https://github.com/kubernetes/test-infra/issues/14531
- https://github.com/kubernetes/test-infra/pull/14572

And then I moved it back for what was "master" as of 2020-01-20 (which would be v1.18)
- https://github.com/kubernetes/test-infra/pull/15803

But because of the way jobs are generated, it kept getting dropped from blocking for release-branches

1.19 jobs are the earliest jobs that exist

Sooo... I'm moving all the release-branch jobs to blocking

spiffxp:
Opened https://github.com/kubernetes/test-infra/pull/23008 to fix

PR #22990, opened 2021-07-22
releng: Point 1.22 jobs to kubekins-e2e:v20210721-2b77449-1.22

puerco:
The generated jobs for the 1.22 picked the previous kubekins image
which do not have a 1.22 variant. This PR points them to
`gcr.io/k8s-testimages/kubekins-e2e:v20210721-2b77449-1.22`
to fix them before the general version bump.

/cc @kubernetes/release-engineering 
/priority critical-urgent

Signed-off-by: Adolfo García Veytia (Puerco) <adolfo.garcia@uservers.net>

PR #22959, opened 2021-07-21
releng: v1.22 branch jobs and testgrid conf

puerco:
This PR creates the jobs for the release-1.22 branch. It updates the testgrid dashboards and the gubernator configuration.

v1.22.0-rc.0 release issue:  https://github.com/kubernetes/sig-release/issues/1633

/sig release
/area release-eng
/priority critical-urgent

cc: @kubernetes/release-engineering

PR #23008, opened 2021-07-23
releng: make gce-default jobs release-blocking

spiffxp:
First commit regenerates release-branch jobs.  ~I think they fell out of sync somewhere between https://github.com/kubernetes/test-infra/pull/22959 and https://github.com/kubernetes/test-infra/pull/22990~.  No actually I think it's just that my system's python traversed things in a slightly different order? So the first commit is a regeneration, the second commit is the actual change here.

Second commit makes the "default" gce e2e jobs release-blocking again. The short version is: this job was blocking for 1.18, it should have been blocking for every version ever since. The fact that it was not is a bug. For full context see https://github.com/kubernetes/test-infra/issues/23003#issuecomment-885920171

Fixes: https://github.com/kubernetes/test-infra/issues/23003


spiffxp:
The crazy diff on regeneration should not be this crazy.  It looks like nothing changed at all other than ordering of the `testgrid-num-failures-to-alert` annotation.  I guess the script doesn't bother making its output consistently sorted.
```
$ git checkout master
$ <generated.yaml yq -y '.periodics | sort_by(.name)' > generated-old.yaml
$ git checkout make-gce-default-jobs-blocking
$ <generated.yaml yq -y '.periodics | sort_by(.name)' > generated-new.yaml
$ diff generated-{old,new}.yaml
```
```diff
199d198
<     testgrid-num-failures-to-alert: '6'
200a200
>     testgrid-num-failures-to-alert: '6'
245d244
<     testgrid-num-failures-to-alert: '6'
246a246
>     testgrid-num-failures-to-alert: '6'
334a335
>     testgrid-dashboards: sig-release-1.21-blocking
336d336
<     testgrid-dashboards: sig-release-1.21-informing
# ... more of the same
```

PR #23011, opened 2021-07-24
Remove `kind/design` from `default` section in favour of `kind/feature`

MadhavJivrajani:
- Remove `kind/design` from `default`
- Add `kind/design` explicitly to k/kubeadm, k-sigs/{controller-runtime, cluster-api} as per the discussion in
  https://github.com/kubernetes/community/issues/5641
- Repos that prefer having a `kind/design` label, would need to opt-in and add it explicitly to `labels.yaml`
- Subsequently generate docs for the corresponding changes

/hold

Fixes: https://github.com/kubernetes/community/issues/5641

Signed-off-by: Madhav Jivrajani <madhav.jiv@gmail.com>

### DIFF
--- a/label_sync/labels.yaml
+++ b/label_sync/labels.yaml
@@ -260,12 +260,6 @@ default:
       prowPlugin: label
       addedBy: anyone
     - color: c7def8
-      description: Categorizes issue or PR as related to design.
-      name: kind/design
-      target: both
-      prowPlugin: label
-      addedBy: anyone
-    - color: c7def8
       description: Categorizes issue or PR as related to documentation.
       name: kind/documentation
       previously:
@@ -1617,6 +1611,14 @@ repos:
         target: issues
         prowPlugin: label
         addedBy: anyone
+  kubernetes/kubeadm:
+    labels:
+      - color: c7def8
+        description: Categorizes issue or PR as related to design.
+        name: kind/design
+        target: both
+        prowPlugin: label
+        addedBy: anyone
   kubernetes-sigs/cluster-api:
     labels:
       - color: c7def8
@@ -1705,6 +1707,9 @@ repos:
       - color: 0052cc
         description: Issues or PRs related to cluster topology
         name: area/topology
+      - color: c7def8
+        description: Categorizes issue or PR as related to design.
+        name: kind/design
         target: both
         prowPlugin: label
         addedBy: anyone
@@ -1904,3 +1909,11 @@ repos:
         target: both
         addedBy: anyone
         prowPlugin: label
+  kubernetes-sigs/controller-runtime:
+    labels:
+      - color: c7def8
+        description: Categorizes issue or PR as related to design.
+        name: kind/design
+        target: both
+        prowPlugin: label
+        addedBy: anyone
--- a/label_sync/labels.md
+++ b/label_sync/labels.md
@@ -13,6 +13,7 @@
 - [Labels that apply to kubernetes-sigs/cluster-api-provider-aws, only for issues](#labels-that-apply-to-kubernetes-sigscluster-api-provider-aws-only-for-issues)
 - [Labels that apply to kubernetes-sigs/cluster-api-provider-azure, for both issues and PRs](#labels-that-apply-to-kubernetes-sigscluster-api-provider-azure-for-both-issues-and-prs)
 - [Labels that apply to kubernetes-sigs/contributor-tweets, for both issues and PRs](#labels-that-apply-to-kubernetes-sigscontributor-tweets-for-both-issues-and-prs)
+- [Labels that apply to kubernetes-sigs/controller-runtime, for both issues and PRs](#labels-that-apply-to-kubernetes-sigscontroller-runtime-for-both-issues-and-prs)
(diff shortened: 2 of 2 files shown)


fejta (approved):
Should we use `previously` in order to migrate these labels to `kind/feature`?

/hold

MadhavJivrajani:
> Should we use previously in order to migrate these labels to kind/feature?

That's a good point, I suspect this will end up affecting _all_ repos and not just `k/k`. I had already done a sweep of `k/k` and replaced `kind/design` with `kind/feature`, I can do that once again now, and once after the PR is merged just to be sure that everything is covered.

fejta (approved):

label_sync/labels.yaml
```diff
@@ -257,12 +257,6 @@ default:
       target: both
       prowPlugin: label
       addedBy: anyone
-    - color: c7def8
-      description: Categorizes issue or PR as related to design.
-      name: kind/design
-      target: both
-      prowPlugin: label
-      addedBy: anyone
```
spiffxp, line 265: This will prevent the label from being added to new repos, but will not actually delete the label from existing repos.  You need to use `deleteAfter` for that.  

But, doing so will delete the label from any previously closed issues that use it, and we'll lose that historical context if they have no other kind/ labels. e.g. the query [`is:closed repo:kubernetes/kubernetes label:kind/design -label:kind/feature`][1] yields 798 results

Without looking deeply at the code I'm not sure what the behavior of label syncer is when asked to simultaneously:
1. delete an org-wide label, add as repo-specific label
2. make an org-wide label a `previously` for another org-wide label, but preserve it for specific repos

If we care about preserving historical categorization, I foresee:
- testing what label_sync would do with case 2 (if there aren't already unit tests that prove this out)
- if it can't gracefully handle case 2, manually adding kind/feature to kind/design issues that don't already have it (closed or not)
- testing what label_sync would do with case 1
- if it can't gracefully handle case 1, manually delete the label from all repos that aren't specifically using kind/design

[1]: https://github.com/issues?q=is%3Aclosed+repo%3Akubernetes%2Fkubernetes+label%3Akind%2Fdesign+-label%3Akind%2Ffeature

spiffxp (changes requested):
tl;d;r I suspect there is a sequence of PRs and/or manual actions needed here if we care about preserving historical categorization.

If we're ok losing it, I'm still not sure this is going to do what you expect it to.

label_sync/labels.yaml
MadhavJivrajani: Thanks @spiffxp, this is really helpful!

A few clarifications:
> This will prevent the label from being added to new repos, but will not actually delete the label from existing repos.

Yes, that is the intention. From https://github.com/kubernetes/community/issues/5641, the general consensus was that `kind/design` be removed from `k/k` alone and not all repositories. If new repositories would like the `kind/design` label, then they would need to opt-in by adding their repositories to `labels.yaml`: https://github.com/kubernetes/community/issues/5641#issuecomment-882506726. This change of opting in will be communicated through the k-dev mailing list. Relevant slack discussion [here](https://kubernetes.slack.com/archives/C1TU9EB9S/p1626697872447500).

> make an org-wide label a previously for another org-wide label, but preserve it for specific repos

I had tried this, `label_sync` unfortunately does not allow this.

> if it can't gracefully handle case 2, manually adding kind/feature to kind/design issues that don't already have it (closed or not)

ack
I've done a sweep for open issues/PRs in `k/k` but not the closed ones. 

> testing what label_sync would do with case 1

From my initial understanding, this should work fine. But I will take closer look and try adding a few tests to be sure.

MadhavJivrajani: Hey @spiffxp! Quick update - I ran a few tests by running `label_sync` against a dummy repo:

> delete an org-wide label, add as repo-specific label

The behavior of doing this is as expected, that is:
- When a new repo is added, it will not get this deleted label, unless it is explicitly added in the `repos` section
- Repos that already have the deleted label, will continue to have it

> make an org-wide label a `previously` for another org-wide label, but preserve it for specific repos

If something like this is done, then it will sync at an org level, and all existing `kind/design` labels will be replaced with `kind/feature` (even in repos that actually use `kind/design`)
```yaml
---
default:
  labels:
    - color: c7def8
      name: kind/feature
      description: Feature
      previously:
      - color: c7def8
        name: kind/design
        description: Design
repos:
  org/repo:
    labels:
      - color: c7def8
        name: kind/design
        description: Design
```

MadhavJivrajani:
Hi, the next steps here after discussing with the GitHub admins would be to send out communications to the kdev mailing list regarding the changes. We're hoping to send out communications by end of the week (17th Sept), if folks any further feedback on this or the communication being sent, please let me know.

The planned communication to be sent out can be found [here](https://docs.google.com/document/d/1j0qID0FMyxB-oBQZa82P6LzrHnGIV0QGFgUYtj3IVt4/edit?usp=sharing).

cc @spiffxp @cblecker

MadhavJivrajani:
@spiffxp the lazy consensus period on this ends today (27th September), we should be good to go once its lgtm'd.

spiffxp (approved):
/approve
/lgtm

PR #15068, opened 2019-11-02
sig-release: Rearrange dash configs and prune some jobs from sig-release-job-config-errors

justaugustus:
This continues the pruning work I started in https://github.com/kubernetes/test-infra/pull/15039.

Here we're doing a few things:
- Add sig-release-generated dashboard to allow us to easily identify generated tests
- ~Allow kubernetes-bazel-build jobs to fork into release dashes by moving them to *-informing~
- Drop kops and Windows jobs from job-config-errors dash
- Drop GCE/GKE jobs from job-config-errors dash

The remaining jobs on the `sig-release-job-config-errors` are in [`config/jobs/kubernetes/sig-cli/sig-cli-config.yaml`](https://github.com/kubernetes/test-infra/blob/3c66bac45a151d8eb3a6908035da0b0d5fc1a374/config/jobs/kubernetes/sig-cli/sig-cli-config.yaml).

Several of them seem to be passing, but are incorrect named e.g., `1.12-1.13` when they're using something like stable1/stable2 markers
There are enough of them configured in confusing ways, that I'd like to tackle them in a separate PR.

PR #23022, opened 2021-07-26
releng: generate_tests hygiene fixes

spiffxp:
Followup to https://github.com/kubernetes/test-infra/pull/23008 where I noticed the diff from a few one-line changes was massive due to unsorted output

- sort `releng/generate_tests.py` input by job name prior to outputting (rely on dicts being ordered by insertion by default for output sorting)
- fix a ruamel.yaml deprecation warning I noticed when running `releng/generate_tests_test.py`
- fixed `releng/generate_tests_test.py`, which was broken by https://github.com/kubernetes/test-infra/pull/15068 ...
- ... which was easy to miss because it wasn't run by `bazel test //releng/...`, so add it as a `py_test` target

/kind cleanup


ameukam:
/lgtm
/hold
Remove hold when ready to deploy.

puerco:
Thank you for looking into this @spiffxp 
/lgtm

PR #23024, opened 2021-07-26
k/autoscaler - Remove branch protection from gh-pages branch

gjtempleton:
Also switch the required status check from Travis to GHActions runs

We want to follow the example of other repositories such as cluster-proportional-autoscaler and remove branch protection to allow automatic publishing of new versions of helm charts. (See e.g. #22996)

The change in required status checks is due to us now having moved from Travis to Github Actions for the repository's CI. See e.g. https://github.com/kubernetes/autoscaler/pull/4225

cc: @mwielgus

### DIFF
--- a/config/prow/config.yaml
+++ b/config/prow/config.yaml
@@ -173,7 +173,12 @@ branch-protection:
         autoscaler:
           required_status_checks:
             contexts:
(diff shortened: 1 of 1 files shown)


spiffxp (approved):
/approve
/lgtm

PR #23026, opened 2021-07-26
Revert "Revert "skip_report flag in prowjob only used by crier report…

chaodaiG:
…ers that report back to CI system""

This reverts commit 35e82f62e00372f56eab25f40b47aed433c42c33.

This is a revert of a revert, the previous revert was due to a regression in istio prow since the old behavior was used by them, specifically slack notification relies on skip_report for filtering out a small set of jobs.

This was fixed in https://github.com/kubernetes/test-infra/pull/22918, which fixed the yaml roundtrip error, and makes empty slices being respected


chaodaiG:
/cc @cjwagner 
cc @howardjohn

cjwagner (approved):

PR #23034, opened 2021-07-27
Do not comment help message for Gerrit repos that are opt-out

chizhg:
In https://github.com/kubernetes/test-infra/pull/22817 we added the feature to comment help message for running presubmit test, if it is requested from the Gerrit changes. However, for some repos that run their presubmit Prow jobs with multiple Prow instances, the comments could be confusing, so they want to opt-out from this feature.

This PR adds a flag to control the list of projects that want opt-out.

/cc @chaodaiG

### DIFF
--- a/prow/cmd/gerrit/main.go
+++ b/prow/cmd/gerrit/main.go
@@ -42,10 +42,11 @@ import (
 )
 
 type options struct {
-	cookiefilePath    string
-	tokenPathOverride string
-	config            configflagutil.ConfigOptions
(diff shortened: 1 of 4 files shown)


cjwagner (approved):
It'd be nice for this to be a config field rather than a CLI flag so that this could be changed at runtime without restarting. This is already a problem for some of the other Gerrit flags though so this probably shouldn't block merging this now. We can migrate flags to config all at once at a later time.

Issue #10423, opened 2018-12-13
Migrate Prow GitHub authentication from Personal Access Token to a GitHub App


ccojocar:
**What would you like to be added**:

Prow is currently using a Personal Access Token to authenticate with GitHub API. In the meantime, GitHub has revamped the API authentication mechanism and introduced [GitHub Apps](https://developer.github.com/apps/about-apps/), which will replace soon the Personal Access Token and OAuth tokens. 

Migrate prow from a Personal Access Token to a GitHub App. 

**Why is this needed**:

GitHub Apps provide some advantages over Personal Access Tokes which are listed here:
https://developer.github.com/apps/migrating-oauth-apps-to-github-apps/

cc @rawlingsj @wbrefvem

/kind feature
/area prow

stevekuznetsov:
Not sure it's required for V4 -- we're successfully using OAuth tokens for the V4 API access we do today. I think we've talked about this in the past but it was unclear what the effort involved would be and what the downsides might be for the project.

/cc @fejta @spiffxp

michaelneale:
certainly for access to checks api (perhaps not v4). Github are very keen on shifting to app vs PAT

ccojocar:
I updated the description of the issue with the official reasons mentioned by GitHub in order to avoid any controversy.

BenTheElder:
cc @cjwagner

jlewi:
I think an advantage of GitHub apps over using PAT is that GitHub apps are owned by the organization. I think this will make them much easier to administer than bot accounts which requires sharing access to a GitHub account.

I dug into this an it looks to me like the way it works is that GitHub apps use private certs to generate JWTs which are then used to authenticate to the GitHub APIs: e.g. see the examples
https://github.com/github-developer/using-the-github-api-in-your-app/blob/master/server.rb#L104

So I think the primary work would be updating your GitHub client to generate JWTs
https://github.com/kubernetes/test-infra/blob/e591cdfa2e2b442fd92b9e6aba4b7c78f95cbb0b/ghclient/core.go#L49

Some googling turned up the following library for generating JWTs in go:
https://github.com/dgrijalva/jwt-go

fejta:
Here is how we can update tackle to help make this easier: https://developer.github.com/apps/building-github-apps/creating-github-apps-from-a-manifest/

Note that this change would have some downsides. We could not, for example, react to a comment/issue

stevekuznetsov:
[...]

PR #23036, opened 2021-07-27
Tide: Finish apps auth support

alvaroaleman:
Ref https://github.com/kubernetes/test-infra/issues/10423

This change finishes the GitHub apps support in Tide by also adding it to the status controller and the merge blocker code. The sync controller already got support in https://github.com/kubernetes/test-infra/pull/20575.

It makes the two split their v4 queries by org if and only if apps auth is enabled. If not, behavior will remain unchanged.

### DIFF
--- a/prow/tide/blockers/blockers.go
+++ b/prow/tide/blockers/blockers.go
@@ -22,10 +22,13 @@ import (
 	"regexp"
 	"sort"
 	"strings"
+	"sync"
 	"time"
 
 	githubql "github.com/shurcooL/githubv4"
 	"github.com/sirupsen/logrus"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/apimachinery/pkg/util/sets"
 )
 
 var (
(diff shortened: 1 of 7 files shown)


prow/tide/blockers/blockers.go
```diff
@@ -70,16 +73,56 @@ func (b Blockers) GetApplicable(org, repo, branch string) []Blocker {
...
+
+		go func() {
+			defer wg.Done()
+			result, err := search(
+				ctx,
+				ghc,
+				org,
+				log,
+				query,
+			)
```
petr-muller, line 112: nit: liebreaking above looks a bit excessive

prow/tide/blockers/blockers.go
```diff
@@ -70,16 +73,56 @@ func (b Blockers) GetApplicable(org, repo, branch string) []Blocker {
...
+			lock.Lock()
+			defer lock.Unlock()
+			if err != nil {
+				errs = append(errs, err)
+			}
+			issues = append(issues, result...)
```
petr-muller, line 118: IMO we should not assume that `result` makes any sense when `err != nil`

prow/tide/blockers/blockers.go
```diff
@@ -70,16 +73,56 @@ func (b Blockers) GetApplicable(org, repo, branch string) []Blocker {
...
+func FindAll(ghc githubClient, log *logrus.Entry, label string, orgRepoTokensByOrg map[string]string, splitQueryByOrg bool) (Blockers, error) {
+	queries := make(map[string]string, len(orgRepoTokensByOrg))
+	for org, query := range orgRepoTokensByOrg {
+		if splitQueryByOrg {
+			queries[org] = blockerQuery(label, query)
+		} else {
+			if queries[""] != "" {
+				queries[""] += " "
+			}
+			queries[""] += blockerQuery(label, query)
```
petr-muller, line 85: Seems to me that we should simply use a set right away on `!splitQueryByOrg`, then you don't need to deal with first element, then split it the full string, then make a string from it again...

prow/tide/status.go
```diff
@@ -547,46 +559,93 @@ func (sc *statusController) sync(pool map[string]PullRequest, blocks blockers.Bl
...
+			result, err := search(sc.ghc.QueryWithGitHubAppsSupport, sc.logger, query, latestPR.Time, now, org)
+			log.WithField("duration", time.Since(now).String()).WithField("result_count", len(result)).Debug("Searched for open PRs.")
+
+			func() {
+				sc.storedStateLock.Lock()
+				defer sc.storedStateLock.Unlock()
+
+				log := log.WithField("latestPR", sc.storedState[org].LatestPR)
+				if len(result) == 0 {
+					log.Debug("no new results")
```
petr-muller, line 614: Looks like a good candidate for a `return` to allow shifting left the code below? (L618 too)

prow/tide/blockers/blockers.go
alvaroaleman: The code below doesn't know anything about github apps auth or not, it just executes all the queries in the map. This means we can not use different data structures in `qureries`, depending on with auth mechanism is used.

I've updated the code to always store the query in a `set.String`, let me know what you think

prow/tide/blockers/blockers.go
alvaroaleman: Which linebreaking exactly do you mean?

prow/tide/blockers/blockers.go
alvaroaleman: Updated

prow/tide/status.go
alvaroaleman: Good point, updated

prow/tide/blockers/blockers.go
petr-muller: `result, err := search(ctx, ghc, org, og, query)

prow/tide/blockers/blockers.go
petr-muller: Your update is good! Sorry for not being clear enough.

petr-muller (approved):

PR #23039, opened 2021-07-27
prow controller manager logs build cluster name when failed to schedu…

chaodaiG:
…le them

Without this the log just says failed creating pod in http://<IP>/<NS>..., makes it much harder to figure out which cluster it refers to

### DIFF
--- a/prow/plank/reconciler.go
+++ b/prow/plank/reconciler.go
@@ -18,6 +18,7 @@ package plank
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"sync"
 	"time"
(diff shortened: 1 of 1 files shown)


alvaroaleman (approved):

prow/plank/reconciler.go
```diff
@@ -845,6 +846,10 @@ func isRequestError(err error) bool {
 	code := 500 // This is what kerrors.ReasonForError() defaults to.
 	if statusErr, ok := err.(kerrors.APIStatus); ok {
 		code = int(statusErr.Status().Code)
+	} else if wrappedErr := errors.Unwrap(err); wrappedErr != nil {
```
chaodaiG, line 849: Also check wrapped error, it's checked at https://github.com/kubernetes/test-infra/pull/23039/files#diff-f2aec52c49557c430579f10ddb0e1f5408f1ce88ef68e2daf687725e1d6846e7L265

alvaroaleman: The whole passage can be simplified as:
```
	if status := APIStatus(nil); errors.As(err, &status) {
		code =  status.Status().Code
	}
	return metav1.StatusReasonUnknown
```

Which is in fact how the upstream matching supports wrapped errrors: https://github.com/kubernetes/kubernetes/blob/dadecb2c8932fd28de9dfb94edbc7bdac7d0d28f/staging/src/k8s.io/apimachinery/pkg/api/errors/errors.go#L649-L652

chaodaiG: good to know, done.

alvaroaleman (approved):

PR #23044, opened 2021-07-28
images/bootstrap: build multi-archs amd64 & arm64

morlay:
changes here could make people use `docker buildx build` to build multi-arch images

example images: https://github.com/orgs/querycap/packages?repo_name=k8s-prow-images

rel: https://github.com/kubernetes/test-infra/issues/16588

### DIFF
--- a/images/bootstrap/Dockerfile
+++ b/images/bootstrap/Dockerfile
@@ -72,9 +72,13 @@ RUN wget -q https://dl.google.com/dl/cloudsdk/channels/rapid/google-cloud-sdk.ta
         --bash-completion=false \
         --path-update=false \
         --usage-reporting=false && \
-    gcloud components install alpha beta kubectl && \
+    gcloud components install alpha beta && \
     gcloud info | tee /workspace/gcloud-info.txt
 
+# gcloud not provide kubectl arm64 binary
+ARG TARGETARCH
+RUN curl -LO "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/${TARGETARCH:-amd64}/kubectl" && \
+    chmod +x ./kubectl && mv ./kubectl /usr/local/bin/kubectl
 
 #
 # BEGIN: DOCKER IN DOCKER SETUP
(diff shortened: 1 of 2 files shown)


alvaroaleman:
Isn't this deprecated in favor of decoration utils?

deepsm007:
@morlay Can we also provide support for P/Z images along with arm?

morlay:
@deepsm007 what means `P/Z images` ?

deepsm007:
> @deepsm007 what means `P/Z images` ?

ppc64le and s390x

morlay:
@deepsm007 

I think it couldn't
cloud sdk not provide `ppc64le` and `s390x` binaries, https://cloud.google.com/sdk/docs/install
which seems not an opensource tool

mkumatag:
> @deepsm007
> 
> I think it couldn't
> cloud sdk not provide `ppc64le` and `s390x` binaries, https://cloud.google.com/sdk/docs/install
> which seems not an opensource tool

In docker file file I see this is the place used to he source and installing it, this may even work as is on ppc64le platform too,

https://github.com/ppc64le-cloud/test-infra/blob/f45196897c83015c049f417e5553e49bfd718b8e/images/all-in-one/Dockerfile#L128:L132

morlay:
@mkumatag 

if so, we could just add `--platform` to make build those images

mkumatag:
> @mkumatag
> 
> if so, we could just add `--platform` to make build those images

Lets wait and hear from @deepsm007 whether he can build image locally and things are fine. If yes - then we can modify this PR itself to include those options.

deepsm007:
> > @mkumatag
> > if so, we could just add `--platform` to make build those images
> 
> Lets wait and hear from @deepsm007 whether he can build image locally and things are fine. If yes - then we can modify this PR itself to include those options.

I tried building the images locally, hitting the docker rate limit issue so unable to build them.

chendave:
Looks like this image has been deprecated: https://github.com/kubernetes/test-infra/tree/master/images/bootstrap#readme, not sure whether the pursuing on this image is still acceptable. 

Might continue the work on krte image instead, pls see: https://github.com/kubernetes/test-infra/pull/24783

PR #23046, opened 2021-07-28
Fix yaml syntax in skip_if_only_changed documentation

zaneb:
If a yaml string is surrounded by single quotes, then it can contain literal backslashes. However if, as in this case, it is [surrounded by double quotes](https://yaml.org/spec/1.2/spec.html#id2787109) then backslash is an escape character and literal backslashes must themselves be escaped.

### DIFF
--- a/prow/jobs.md
+++ b/prow/jobs.md
@@ -194,7 +194,7 @@ presubmits:
   org/repo:
   - name: compile-job
     always_run: false
-    run_if_changed: "(\.[ch]|^Makefile)$"
+    run_if_changed: "(\\.[ch]|^Makefile)$"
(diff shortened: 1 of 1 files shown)


matthyx:
@zaneb but here do we really want a literal backslash? I think the goal here is to escape the dot to check the extension?

zaneb:
> @zaneb but here do we really want a literal backslash? I think the goal here is to escape the dot to check the extension?

We want to escape it in the regex, but not in the yaml. `\.` is not a legal escape in yaml, so the example doesn't parse as currently given.

alvaroaleman (approved):

PR #23048, opened 2021-07-28
config/jobs: use k8s-infra-ci-robot for k8s-infra PR-creating jobs

spiffxp:
Related:
- Part of: https://github.com/kubernetes/k8s.io/issues/1742
- Depends on: https://github.com/kubernetes/k8s.io/pull/2403

The main intent of this is to add an autobump job, but now that we have access to k8s-infra-ci-robot, let's switch the audit job to use it as well.

The jobs have short intervals for debugging and will be set back to longer intervals once verified as working.

Kept as separate commits in case we want to revert.

### DIFF
--- a/config/jobs/kubernetes/wg-k8s-infra/trusted/wg-k8s-infra-k8sio.yaml
+++ b/config/jobs/kubernetes/wg-k8s-infra/trusted/wg-k8s-infra-k8sio.yaml
@@ -1,6 +1,8 @@
 periodics:
 - name: ci-k8sio-audit
-  interval: 3h
+  # TODO(spiffxp): low interval for debugging / verification, remove and adjust cron when done
+  interval: 10m
+  # interval: 3h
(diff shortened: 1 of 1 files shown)


spiffxp:
Audit job currently running: https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/ci-k8sio-audit/1420446001876963328

Autobump job failing on secret... typo probably: https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/ci-k8sio-autobump-prow-build-clusters/1420460598528839680

>The pod could not start because it could not mount the volume "github": secret "oauth-token" not found

spiffxp:
PR up to fix the typo: https://github.com/kubernetes/test-infra/pull/23052

spiffxp:
PR to make the audit script use the github info that matches the github token (wonder if we could automatically get all that from the token...) https://github.com/kubernetes/k8s.io/pull/2410

PR #23050, opened 2021-07-28
config/jobs: shorten prow autobump slack msg

spiffxp:
Match the message format used for the build cluster deploy jobs

Current format:
Screen Shot 2021-07-28 at 12 15 47 PM

New format:
Screen Shot 2021-07-28 at 12 15 14 PM


chaodaiG (approved):
Fancy! Thank you!

PR #23056, opened 2021-07-28
config/jobs: slow down k8s-infra-ci-robot jobs

spiffxp:
Related:
- Part of: https://github.com/kubernetes/k8s.io/issues/1742
- Followup to: https://github.com/kubernetes/test-infra/pull/23048

The jobs were 10m intervals for debugging, but now that we've confirmed they work, let's use more realistic intervals:
- autobump prow build clusters 3h earlier than prow itself is autodeployed
- run audit every 2h (it was formerly 3h, but it takes about 90m, and more frequent updates could be useful)


rikatz:
/lgtm
Thanks Aaron :)

PR #23061, opened 2021-07-29
containerd version of pull-kubernetes-node-kubelet-serial job

dims:
Experimenting with a new CI job that runs the same tests as `pull-kubernetes-node-kubelet-serial` but using `containerd` as the CRI and not `docker`

### DIFF
--- a/config/jobs/kubernetes/sig-node/sig-node-presubmit.yaml
+++ b/config/jobs/kubernetes/sig-node/sig-node-presubmit.yaml
@@ -281,6 +281,44 @@ presubmits:
         env:
         - name: GOPATH
           value: /go
+  - name: pull-kubernetes-node-kubelet-serial-containerd
+    always_run: false
+    optional: true
+    skip_report: false
+    skip_branches:
+      - release-\d+\.\d+  # per-release image
+    annotations:
+      testgrid-dashboards: sig-node-kubelet
+      testgrid-tab-name: pr-node-kubelet-serial-containerd
+    labels:
+      preset-service-account: "true"
+      preset-k8s-ssh: "true"
+    spec:
+      containers:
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes/sig-node/sig-node-presubmit.yaml
```diff
@@ -281,6 +281,44 @@ presubmits:
         env:
         - name: GOPATH
           value: /go
+  - name: pull-kubernetes-node-kubelet-serial-containerd
```
ehashman, line 284: Do we also want to add a periodic? I don't think one exists right now

config/jobs/kubernetes/sig-node/sig-node-presubmit.yaml
```diff
@@ -281,6 +281,44 @@ presubmits:
...
+    always_run: false
+    optional: true
+    skip_report: false
+    skip_branches:
+      - release-\d+\.\d+  # per-release image
+    annotations:
+      testgrid-dashboards: sig-node-kubelet
+      testgrid-tab-name: pr-node-kubelet-serial-containerd
```
ehashman, line 292: Side note: at some point I wanna split all the presubmits into a separate tab

config/jobs/kubernetes/sig-node/sig-node-presubmit.yaml
dims: not yet, let's get the presubmit working first.

config/jobs/kubernetes/sig-node/sig-node-presubmit.yaml
dims: Agree for later.

PR #23067, opened 2021-07-29
Use kubekins-e2e image for gcp-compute-persistent-disk-csi-driver jobs

chrishenzie:
The previous image uses go1.13. Bumping to the image version used for other jobs which uses go1.16.

/sig storage
/assign @fejta 
/cc @saikat-royc


chrishenzie:
Assigning owners from file

/assign @msau42 
/assign @jingxu97

PR #23072, opened 2021-07-30
Make checkconfig's cluster validation verify cluster existence and liveness based on statuses published by Plank.

cjwagner:
Follow up to https://github.com/kubernetes/test-infra/pull/23060 (the first commit is from the previous PR).

/assign @chaodaiG


chaodaiG (approved):
/hold

chaodaiG (approved):

cjwagner:
I don't understand why I am getting this error on this PR:
```
level=error msg="Running error: deadcode: analysis skipped: errors in package: [/b/f/w/bazel-out/k8-fastbuild/bin/external/io_k8s_repo_infra/hack/verify-golangci-lint.runfiles/io_k8s_test_infra/prow/cmd/prow-controller-manager/main.go:43:2: could not import k8s.io/test-infra/prow/plank (prow/plank/reconciler.go:48:2: use of internal package k8s.io/test-infra/prow/crier/reporters/gcs/internal/util not allowed)]"
```

This PR doesn't modify either of those packages so I would have expected the error to be exposed on the previous PR that introduced it. Does this break the bazel presubmit for every open PR? I'm not sure how to debug how this happened, @fejta suggested that it could be an RBE caching problem.

In terms of the actual error message, I'd like to reuse this code rather than copy pasting it. I don't see any motivation for preventing reuse of this code so it is unclear to me why this package was marked as internal in the first place and I'm inclined to just remove that constraint. Does anyone have opinions?  @fejta @alvaroaleman

chaodaiG:
> In terms of the actual error message, I'd like to reuse this code rather than copy pasting it. I don't see any motivation for preventing reuse of this code so it is unclear to me why this package was marked as internal in the first place and I'm inclined to just remove that constraint. Does anyone have opinions?

Agree that this package should be shared

alvaroaleman:
> This PR doesn't modify either of those packages so I would have expected the error to be exposed on the previous PR that introduced it. Does this break the bazel presubmit for every open PR? I'm not sure how to debug how this happened, @fejta suggested that it could be an RBE caching problem.

I don't get this error when running the linter locally (where that line exists) and I have not seen it on any other PR. I can not tell you why are are seeing it here. I am and always have been in favor of not running sth like a linter in bazel, because it provides zero benefit but introduces a huge bottleneck because there is apparently only one person that can debug issues with it (@fejta ).

> In terms of the actual error message, I'd like to reuse this code rather than copy pasting it. I don't see any motivation for preventing reuse of this code so it is unclear to me why this package was marked as internal in the first place and I'm inclined to just remove that constraint. Does anyone have opinions? @fejta @alvaroaleman

Sure but any package in a given go module is allowed to import any other package in the same go module, regardless of it being `internal` or not. It only prevents other modules from importing it. So no idea why this error showed up here, but it is a false positive.

fejta:
IMO we should either commit to having more than me debug bazel issues or else use something other than bazel for releases.

fejta:
If we think this is a rules-go specific issue it seems like it might be useful to bring the issue up with them (either on their github repo, slack or google chat)

cjwagner:
> Sure but any package in a given go module is allowed to import any other package in the same go module, regardless of it being `internal` or not. It only prevents other modules from importing it. So no idea why this error showed up here, but it is a false positive.

[These golang docs](https://golang.org/doc/go1.4#internalpackages) suggest that `internal` packages limit access based on directory path, not module scope. That means that this error is legitimate, but it still seems wrong for it to be affecting this PR and not the PR that introduced the problem. I'm going to remove the `internal` directory level to resolve this.

alvaroaleman:
> These golang docs suggest that internal packages limit access based on directory path, not module scope.

If that was true then the repo (or at least the prow cm) wouldn't build anymore or not?

cjwagner:
> If that was true then the repo (or at least the prow cm) wouldn't build anymore or not?

Yes, I wouldn't expect it to build, but I also wouldn't expect errors on this PR when it doesn't touch the packages the errors are coming from :upside_down_face: . Maybe bazel's build rules for go don't respect the `internal` constraint?

In any case, removing the `internal` directory level fixed the presubmit error.
/retest

chaodaiG (approved):

PR #23084, opened 2021-07-30
labels: overhaul kubernetes/k8s.io labels

spiffxp:
Related:
- Implements https://github.com/kubernetes/k8s.io/issues/508#issuecomment-884608113

Trying to accomplish more consistency with folder names. Where before
area/infra was being used as some kind of pseudo namespace, move
everything into area/apps/{foo} that fits that pattern.

Things that are cross-cutting concerns or languages are left unnamespaced

Things that map directly to a folder name use that folder name

Set a deleteAfter date of the next wg-k8s-infra meeting

- add area/apps (formerly area/cluster-infra)
- add area/apps/{foo} for each app (formly area/infra/{app} for some)
- add area/bash
- add area/terraform (formerly area/infra/terraform)
- add area/infra
- add area/k8s.gcr.io
- add area/audit
- add area/policy
- rm area/infra/monitoring
- rm area/infra/reliability

### DIFF
--- a/label_sync/labels.yaml
+++ b/label_sync/labels.yaml
@@ -882,8 +882,80 @@ repos:
   kubernetes/k8s.io:
     labels:
       - color: 0052cc
-        description: Issues or PRs related to defining who has access to what (e.g. IAM, RBAC, google groups)
+        description: Define who has access to what via IAM bindings, role bindings, policy, etc.
         name: area/access
+        target: both
+        prowPlugin: label
+        addedBy: anyone
+      - color: 0052cc
+        description: Application management, code in apps/
+        name: area/apps
+        target: both
+        prowPlugin: label
+        addedBy: anyone
+        previously:
+          - name: area/cluster-infra
+      - color: 0052cc
+        description: cert-manager, code in apps/cert-manager/
+        name: area/apps/cert-manager
+        target: both
+        prowPlugin: label
+        addedBy: anyone
+        previously:
+          - name: area/infra/cert-manager
+      - color: 0052cc
+        description: gcsweb.k8s.io, code in apps/gcsweb/
+        name: area/apps/gcsweb
+        target: both
+        prowPlugin: label
+        addedBy: anyone
+      - color: 0052cc
+        description: k8s.io redirector, code in apps/k8s-io/
+        name: area/apps/k8s-io
+        target: both
+        prowPlugin: label
+        addedBy: anyone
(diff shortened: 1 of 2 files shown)


spiffxp:
/uncc @fejta @mpherman2 
/cc @dims @ameukam 
wg leads
/cc @rikatz @munnerz @justaugustus 
since you all commented on https://github.com/kubernetes/k8s.io/issues/508 at some point

label_sync/labels.yaml
```diff
@@ -882,26 +882,107 @@ repos:
...
+        description: Application management, code in apps/
+        name: area/apps
+        target: both
+        prowPlugin: label
+        addedBy: anyone
+        previously:
+          - name: area/cluster-infra
+      - color: 0052cc
+        description: cert-manager, code in apps/cert-manager/
+        name: area/apps/cert-manager
```
ameukam, line 901: Do we want to have this until cert-manager is decommissioned ?

spiffxp: I'm not sure if we want to remove this once cert manager is gone or not.  

Keeping it:
- pro: we keep historical data for e.g. "how much work did we save ourselves by moving to a hosted solution"
- con: we have additional label clutter

Dropping it:
- pro: we have fewer labels to choose from
- con: we lose historical data

I'm inclined to keep it for now.  We could also consider an `area/cert-management` or something to cover the area as a whole vs. one specific implementation

ameukam: Let's keep it for historical tracking.

label_sync/labels.yaml
```diff
@@ -882,26 +882,106 @@ repos:
...
+        prowPlugin: label
+        addedBy: anyone
+      - color: 0052cc
+        description: slack.k8s.io, slack-infra, code in apps/slack-infra
+        name: area/apps/slack-infra
+        target: both
+        prowPlugin: label
+        addedBy: anyone
+      - color: 0052cc
+        description: release.triage.k8s.io, slack-infra, code in apps/triageparty-release
```
ameukam, line 957:
```suggestion
        description: release.triage.k8s.io, triage-party, code in apps/triageparty-release
```

ameukam:
One little nit, otherwise LGTM.

spiffxp:
I would have accepted suggestion if I didn't have to re-run ./hack/update-labels.sh

PTAL

PR #23087, opened 2021-07-31
Set logexporter to capture konnectivity server and agent logs.

cheftako:
Capture konnectivity logs to help debug flaky scalability test where konnecivity-agent is sometimes restarted.


liggitt:
I'm +1 on capturing these, though I don't have context to know if this change accomplishes that :)

dims:
yes, this should do it.

/approve
/lgtm

dims:
@cheftako some of the CI jobs pick up from different locations (the list of files to dump), may be it would be good to update all of them?
https://cs.k8s.io/?q=startupscript&i=nope&files=&excludeFiles=cluster%2Faddons&repos=kubernetes/kubernetes,kubernetes/test-infra

logexporter/cmd/main.go
```diff
@@ -59,7 +59,7 @@ var (
 
 	// Cloud provider specific logfiles.
 	awsLogs      = []string{"cloud-init-output"}
-	gceLogs      = []string{"startupscript"}
+	gceLogs      = []string{"startupscript", "konnectivity-*"}
```
mborsz, line 62: Should it be "containers/konnectivity-*" to match https://github.com/kubernetes/test-infra/commit/8fbc6048fb3d844f8c013fb34695e307fc7124ef?

Also: https://github.com/kubernetes/test-infra/commit/8fbc6048fb3d844f8c013fb34695e307fc7124ef fetches this log only on nodes (for all cloud providers) with log-dump.sh and this change fetches only for gce and both on nodes and masters.

Is the discrepancy between log-dump.sh and logexporter intended? They both try to do the same thing so I would expect them to implement similar logic.

cheftako: Working on making it better match containers. log-dump and logexporter differences were part of dims response and working on that as well. The konnectivity system is only enabled on GCE (that I know of). I intended this to be on both nodes and masters. On nodes it grabs the agent logs and on masters it should grab the server log.

dims:
@cheftako looks like one of the CI jobs is still red. can you please assign it back to me when that is fixed? thanks!

/unassign

PR #23088, opened 2021-07-31
Enhance Prow integration test script and add a README file

YuviGold:
Split the prow integration test script file into 3 functions -
`setup`, `run`, `teardown`.
Support running the entire process as well as calling each function.

Support providing the `run` function with a `test_filter` parameter
to filter tests, and a `cache_test_results` parameter
to avoid caching results and being able to run the same tests.

Add a `README` file to explain how to use it.


chaodaiG (approved):
Thank you!

PR #23089, opened 2021-07-31
docs: use GitHub permalinks in "Life of a Prow Job"

listx:
Unfortunately, all links to pkg.go.dev are broken because they are not
permalinks.

We also sort all GitHub links based on the URL (2nd column) for
consistency.

/cc @fejta @spiffxp


cjwagner (approved):
Thank you for fixing these, Linus!

PR #23098, opened 2021-08-02
enable docker-in-docker for cloud-provider-gcp tasks.

jiahuif:
`k/cloud-provider-gcp` is switching to docker from Bazel to build container images, as a step of producing the final tarballs.

This PR enables docker-in-docker (`dind` for short) for all tasks that invokes building, so that docker can be used for building container image.

Note: this change will NOT break current configuration.

### DIFF
--- a/config/jobs/kubernetes/cloud-provider-gcp/cloud-provider-gcp-periodics.yaml
+++ b/config/jobs/kubernetes/cloud-provider-gcp/cloud-provider-gcp-periodics.yaml
@@ -12,6 +12,7 @@ periodics:
   labels:
     preset-service-account: "true"
     preset-k8s-ssh: "true"
+    preset-dind-enabled: "true"
   extra_refs:
   - org: kubernetes
     repo: cloud-provider-gcp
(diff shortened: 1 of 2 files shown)


jiahuif:
/assign @amwat 
Would you mind taking a look at this?

config/jobs/kubernetes/cloud-provider-gcp/cloud-provider-gcp-periodics.yaml
```diff
@@ -12,6 +12,7 @@ periodics:
   labels:
     preset-service-account: "true"
     preset-k8s-ssh: "true"
+    preset-dind-enabled: "true"
```
amwat, line 15: you will also want to add

```
securityContext:
    privileged: true
```

jiahuif: Thanks for that. Updating.

amwat (commented):
/lgtm
/approve

PR #23109, opened 2021-08-03
updating depstat jobs to include all staging modules

RinkiyaKeDad:
See https://github.com/kubernetes-sigs/depstat/issues/45 and [this](https://github.com/kubernetes-sigs/depstat/pull/48#issuecomment-891050444) comment. Basically this would allow us to use all staging modules as main modules thus providing more accurate stats for k/k.


liggitt:
@dims good to merge this now that 1.23 is open?

dims:
/hold cancel
/approve
/lgtm

sorry forgot about this @liggitt

PR #23115, opened 2021-08-03
config/jobs: promote metrics-bigquery-canary

spiffxp:
remove the old job which runs on google.com prow and runs using
bootstrap

rename the canary job to become the original job

the new job differs from the old job in that:
- it runs on k8s-infra prow
- it runs every 6h starting at 00:03 UTC instead of 24h interval
- it also shows up on wg-k8s-infra-prow testgrid

/hold
this is dependent on having successfully moved the bucket to the
kubernetes.io GCP org


spiffxp:
/hold cancel
`gs://k8s-metrics` now lives over in `kubernetes-public` in the `kubernetes.io` GCP org, ref: https://github.com/kubernetes/k8s.io/issues/1306#issuecomment-892031264

PR #23132, opened 2021-08-04
Trim slash from gcs path when publishing build clusters statuses

chaodaiG:
The resulted GCS path was prepended with a slash, causing it creating a directory named empty string


chaodaiG:
/hold
might need to fix more unit tests first

cjwagner (approved):

PR #23144, opened 2021-08-05
Add documentation to silence slack notifications at the ProwJob level

### LINKED ISSUE #22533: Can't silence job on slack reporter config
**What happened**:

We were thinking that by overriding slack reporter config on the job level we could silence jobs by applying a job config like this:
```
  reporter_config:
    slack:
      job_states_to_report: []
```

We expected the job to stop reporting. Unfortunately this doesn't work.

I've added a test case here: https://github.com/dhiller/test-infra/commit/9d96f46ffab22e0c800c87d3bf6ffbfb8ca4a981

Are we just using it wrong? Is there another way to silence a single job?

fgimenez:
Adds documentation to silence slack notifications by setting `reporter_config.slack.job_states_to_report` to `[]` at the ProwJob level.

### DIFF
--- a/prow/cmd/crier/README.md
+++ b/prow/cmd/crier/README.md
@@ -171,6 +171,21 @@ postsubmits:
             command:
               - echo
 ```
+To silence notifications at the ProwJob level you can pass an empty slice to `reporter_config.slack.job_states_to_report`:
+postsubmits:
+```yaml
+  some-org/some-repo:
+    - name: example-job
+      decorate: true
+      reporter_config:
+        slack:
+          job_states_to_report: []
+      spec:
+        containers:
+          - image: alpine
+            command:
+              - echo
+```
 
 ## Implementation details
 


prow/apis/prowjobs/v1/types.go
```diff
@@ -313,7 +313,13 @@ func (src *SlackReporterConfig) ApplyDefault(def *SlackReporterConfig) *SlackRep
 		merged.Host = def.Host
 	}
 	if merged.JobStatesToReport == nil {
-		merged.JobStatesToReport = def.JobStatesToReport
+		// check if src.JobStatesToReport is an empty slice, we want to keep it to
+		// silence alerts at the job level
+		if src.JobStatesToReport != nil {
```
chaodaiG, line 318: This code path is not covered by the new unit test added below

Also, is it possible that both `merged.JobStatesToReport == nil` and `src.JobStatesToReport != nil` are true? I think if the latter is true, then `merged` should be a deepcopy of `src`, and `JobStatesToReport` cannot be nil

fgimenez: Thanks a lot for your review @chaodaiG thanks to it I see now that one of the existing test cases should cover most of the failing functionality, need to dig deeper.

fgimenez: @chaodaiG I've verified that all is working as expected and #22533 is already fixed, sorry for the noise. Do you think it is worth adding a small note to crier's readme about how to silence slack notifications at the ProwJob level? I've updated this PR to do so.

chaodaiG: yeah, more doc is always better. Thank you!

chaodaiG (approved):

PR #19352, opened 2020-09-24
re-enable all node flaky test, move node performance test to new job

MHBauer:
Filtering of test runs in perf-image-config.yaml means only specific
tests matching "Node Performance Testing" run in flaky job.

 - Move flaky test job back to general image-config.yaml.
 - Create new job config specific to "Node Performance Testing".
 - Decrease ci interval of new "Node Performance Testing" job to 12h
   instead of 2h.

A separate PR in kubernetes/kubernetes needs to be made to take "Node
Performance Testing" tests out of the flaky job. PR #95040

PR #23147, opened 2021-08-05
re-enable all node flaky test, move node performance test to new job

manugupt1:
From: https://github.com/kubernetes/test-infra/pull/19352

Filtering of test runs in perf-image-config.yaml means only specific
tests matching "Node Performance Testing" run in flaky job.

Move flaky test job back to general image-config.yaml.
Create new job config specific to "Node Performance Testing".
Decrease ci interval of new "Node Performance Testing" job to 12h
instead of 2h.

Note: I was not able to check if the dashboard was created or not.

### DIFF
--- a/config/jobs/kubernetes/sig-node/node-kubelet.yaml
+++ b/config/jobs/kubernetes/sig-node/node-kubelet.yaml
@@ -126,7 +126,7 @@ periodics:
       - --deployment=node
       - --gcp-project-type=node-e2e-project
       - --gcp-zone=us-west1-b
-      - --node-args=--image-config-file=/workspace/test-infra/jobs/e2e_node/perf-image-config.yaml
+      - --node-args=--image-config-file=/workspace/test-infra/jobs/e2e_node/image-config.yaml
       - --node-test-args= --kubelet-flags="--cgroups-per-qos=true --cgroup-root=/" --server-start-timeout=420s
       - --node-tests=true
       - --provider=gce
@@ -531,3 +531,33 @@ periodics:
     testgrid-tab-name: kubelet-gce-e2e-swap-fedora
     testgrid-alert-email: ehashman@redhat.com, ikema@google.com
     description: Executes E2E suite with swap enabled on Fedora
+
+- name: ci-kubernetes-node-kubelet-performance-test
+  interval: 12h
+  labels:
+    preset-service-account: "true"
+    preset-k8s-ssh: "true"
+  spec:
+    containers:
+      - image: gcr.io/k8s-testimages/kubekins-e2e:v20210721-2b77449-master
+        args:
+          - --repo=k8s.io/kubernetes=master
+          - --timeout=90
+          - --root=/go/src
+          - --scenario=kubernetes_e2e
+          - --
+          - --deployment=node
+          - --gcp-project-type=node-e2e-project
+          - --gcp-zone=us-west1-b
+          - --node-args=--image-config-file=/workspace/test-infra/jobs/e2e_node/perf-image-config.yaml
+          - --node-test-args= --kubelet-flags="--cgroups-per-qos=true --cgroup-root=/" --server-start-timeout=420s
+          - --node-tests=true
+          - --provider=gce
+          - --test_args=--nodes=1
+          - --timeout=60m
+        env:
+          - name: GOPATH
+            value: /go
+  annotations:
+    testgrid-dashboards: sig-node-kubelet
+    testgrid-tab-name: node-performance-test


config/jobs/kubernetes/sig-node/node-kubelet.yaml
```diff
@@ -531,3 +531,33 @@ periodics:
...
     description: Executes E2E suite with swap enabled on Fedora
+
+- name: ci-kubernetes-node-kubelet-node-performance-test
+  interval: 12h
+  labels:
+    preset-service-account: "true"
+    preset-k8s-ssh: "true"
+  spec:
+    containers:
+      - image: gcr.io/k8s-testimages/kubekins-e2e:v20210721-2b77449-master
```
manugupt1, line 542: Updated the image; compared to previous PR

config/jobs/kubernetes/sig-node/node-kubelet.yaml
```diff
@@ -126,7 +126,7 @@ periodics:
       - --deployment=node
       - --gcp-project-type=node-e2e-project
       - --gcp-zone=us-west1-b
-      - --node-args=--image-config-file=/workspace/test-infra/jobs/e2e_node/perf-image-config.yaml
+      - --node-args=--image-config-file=/workspace/test-infra/jobs/e2e_node/image-config.yaml
```
SergeyKanzhelev, line 129: btw, this is why perf tests were executing: https://github.com/kubernetes/test-infra/blob/caeaacec0de5a867e5d420da85c8f69bf33fc1ca/jobs/e2e_node/perf-image-config.yaml#L8-L9

SergeyKanzhelev (approved):

config/jobs/kubernetes/sig-node/node-kubelet.yaml
manugupt1: Oh got it! thanks a lot.

SergeyKanzhelev: this is very illogical.

manugupt1: So, the internal config pointed to `Node Performance Testing`. Something that I should have gone down a level deeper.

config/jobs/kubernetes/sig-node/node-kubelet.yaml
```diff
@@ -531,3 +531,33 @@ periodics:
     testgrid-tab-name: kubelet-gce-e2e-swap-fedora
     testgrid-alert-email: ehashman@redhat.com, ikema@google.com
     description: Executes E2E suite with swap enabled on Fedora
+
+- name: ci-kubernetes-node-kubelet-node-performance-test
```
ehashman, line 535:
```suggestion
- name: ci-kubernetes-node-kubelet-performance-test
```

config/jobs/kubernetes/sig-node/node-kubelet.yaml
```diff
@@ -531,3 +531,33 @@ periodics:
...
+          - --node-tests=true
+          - --provider=gce
+          - --test_args=--nodes=1 --focus="Node Performance Testing"
+          - --timeout=60m
+        env:
+          - name: GOPATH
+            value: /go
+  annotations:
+    testgrid-dashboards: sig-node-kubelet
+    testgrid-tab-name: node-performance-testing
```
ehashman, line 563:
```suggestion
    testgrid-tab-name: node-performance-test
```

config/jobs/kubernetes/sig-node/node-kubelet.yaml
```diff
@@ -531,3 +531,33 @@ periodics:
...
+  interval: 12h
+  labels:
+    preset-service-account: "true"
+    preset-k8s-ssh: "true"
+  spec:
+    containers:
+      - image: gcr.io/k8s-testimages/kubekins-e2e:v20210721-2b77449-master
+        args:
+          - --repo=k8s.io/kubernetes=master
+          - --timeout=90
```
ehashman, line 545: Do these timeouts make sense? These tests are failing consistently but I don't know if that has anything to do with the runtime.

config/jobs/kubernetes/sig-node/node-kubelet.yaml
```diff
@@ -531,3 +531,33 @@ periodics:
...
+          - --root=/go/src
+          - --scenario=kubernetes_e2e
+          - --
+          - --deployment=node
+          - --gcp-project-type=node-e2e-project
+          - --gcp-zone=us-west1-b
+          - --node-args=--image-config-file=/workspace/test-infra/jobs/e2e_node/perf-image-config.yaml
```
ehashman, line 552: If we use the perf-image-config do we need to specify the focus below?

ehashman (changes requested):
I am okay with merging this and finding out for the perf test stuff but let's fix the names.

config/jobs/kubernetes/sig-node/node-kubelet.yaml
manugupt1: Should we not have timeouts when tests start passing to make sure that they do not take forever. I think it might be better to let them fail after a time regardless, I do not know what the correct value is though.

config/jobs/kubernetes/sig-node/node-kubelet.yaml
manugupt1: No we do not, thanks for catching that.

config/jobs/kubernetes/sig-node/node-kubelet.yaml
```diff
@@ -531,3 +531,33 @@ periodics:
...
+          - --node-test-args= --kubelet-flags="--cgroups-per-qos=true --cgroup-root=/" --server-start-timeout=420s
+          - --node-tests=true
+          - --provider=gce
+          - --test_args=--nodes=1"
```
ehashman, line 556: Looks like there's a stray "

```suggestion
          - --test_args=--nodes=1
```

ehashman (approved):
/lgtm

PR #23148, opened 2021-08-05
bump promoter jobs

listx:
This version includes
https://github.com/kubernetes-sigs/k8s-container-image-promoter/pull/385,
which has a fix for detecting tag moves (and exiting if an error if such
a "promotion" is attempted).

/cc @tylerferrara @spiffxp @dims


spiffxp (approved):
/approve
/lgtm

Issue #18117, opened 2020-06-30
Command to move issue to a new repository


sftim:
**What would you like to be added**:
GitHub lets you [transfer an issue to another repository](https://help.github.com/en/github/managing-your-work-on-github/transferring-an-issue-to-another-repository). I think this requires the GraphQL API.

Approvers with the appropriate access should be able to use a command to trigger moving an issue to a different repository.

**Why is this needed**:

Sometimes, people open great issues but in the wrong repository (eg documentation issues that ought to live in [k/website](https://github.com/kubernetes/website/) in [k/kubernetes](https://github.com/kubernetes/kubernetes), or issues about the built-in help for kubectl, which is mirrored onto the website, that should actually live in [k/kubectl](https://github.com/kubernetes/kubectl)).

Rather than tell those people that the great issue they've filed is in the wrong place, we can add tools to help the person doing triage move it there.

At the moment you need write access on both repositories to do the transfer, so there are few people with the privileges to implement moves. Involving test-infra automation can add richer access controls and broaden the number of people able to take part in this aspect of triage.

prasadkatti:
Another common use case would be [k/website](https://github.com/kubernetes/website) to [k/minikube](https://github.com/kubernetes/minikube).

sftim:
An example of an issue I'd nominate for move: https://github.com/kubernetes/website/issues/22465#issuecomment-657097160

spiffxp:
It looks like this isn't possible via the v3 api, would have to be done via the v4 graphql api, it's unclear to me whether different plugins could be written against different api's

sftim:
I'd like to see this implemented, but it wouldn't be used very often.

sftim:
Another example of an issue where Prow-facilitated moves could be useful: https://github.com/kubernetes/kubernetes/issues/98368

/remove-lifecycle rotten

spiffxp:
/sig contributor-experience
/assign @rikatz 
mentioned a proposal to write such a plugin in #sig-contribex https://kubernetes.slack.com/archives/C1TU9EB9S/p1611252110007400

proposal: https://docs.google.com/document/d/1uqKI5QEAh2fnq7LRRz-2Mhf-jInIpuiAUnPow36jCVU/edit?usp=sharing

rikatz:
Thanks Aaron

/assign @dougsland (Doug is going to work on this with me)

rikatz:
[...]

PR #23153, opened 2021-08-06
Add prow migrate issue

eddiezane:
Fixes: https://github.com/kubernetes/test-infra/issues/18117


eddiezane:
@spiffxp @BenTheElder this is ready for an early look while I write some tests and figure out how to run it 😅. Can confirm the GraphQL request works though.

Couple open questions that might be for someone familiar with Prow plugin UX.

prow/plugins/transfer-issue/transfer-issue.go
```diff
@@ -0,0 +1,96 @@
...
+
+type githubClient interface {
+	GetRepo(org, name string) (github.FullRepo, error)
+	CreateComment(org, repo string, number int, comment string) error
+	IsCollaborator(org, repo, user string) (bool, error)
+	TransferIssue(org, dstRepoNodeID string, issueNodeID string) (*github.TransferIssueMutation, error)
+}
+
+func handleGenericComment(pc plugins.Agent, e github.GenericCommentEvent) error {
+	// TODO: check if this is an actual comment on an issue?
```
petr-muller, line 47: Yeah, otherwise this will hit on pull requests

prow/plugins/transfer-issue/transfer-issue.go
```diff
@@ -0,0 +1,96 @@
...
+	dstRepo, err := gc.GetRepo(org, dstRepoName)
+	if err != nil {
+		log.WithError(err).Errorf("could not fetch destination repo: %s/%s", org, dstRepo)
+		// TODO: post comment. error here means something went wrong or that the repo doesn't exist (404)
+	}
+
+	isCollaboratorSrc, err := gc.IsCollaborator(org, srcRepo, user)
+	if err != nil {
+		log.WithError(err).Errorf("could not fetch if user: %s is collaborator of source repo: %s/%s", user, org, srcRepo)
+		// TODO: comment something went wrong?
```
petr-muller, line 76: AFAIK there isn't any consensus about whether plugins shall respond with comments on internal-ish errors like this call. Most plugins don't comment on API errors like that.

prow/plugins/transfer-issue/transfer-issue.go
```diff
@@ -0,0 +1,96 @@
...
+	if !(isCollaboratorSrc && isCollaboratorDst) {
+		// TODO: comment user is not able to transfer because they aren't an org member
+	}
+
+	issue, err := gc.TransferIssue(org, dstRepo.NodeID, e.NodeID)
+	if err != nil {
+		log.WithError(err).Errorf("could not transfer issue: from: to:")
+		// TODO: comment that issue could not be transferred
+	}
+	// TODO: is commenting at this point necessary or a race condition?
```
petr-muller, line 92: Personally I think it's not necessary, the comment itself serves as a record of who transferred the issue when to where. The only missing bit of information is *from* where it was transferred. 

Does GH itself show some kind of event about the transfer in the UI?

eddiezane: Thanks for the review!

I think you're right. GitHub does do a pretty good job of surfacing this https://github.com/eddiezane/test-repo-2/issues/3

eddiezane: It even preserves the journey if you ping pong it back and forth.

eddiezane:
This is ready for review. I've yet to test the plugin on an actual prow deployment but will start standing that up now (unless someone wants to give it a spin on theirs).

/cc @alvaroaleman

prow/github/client.go
```diff
@@ -261,6 +262,8 @@ type Client interface {
 	Throttle(hourlyTokens, burst int, org ...string) error
 	Query(ctx context.Context, q interface{}, vars map[string]interface{}) error
 	QueryWithGitHubAppsSupport(ctx context.Context, q interface{}, vars map[string]interface{}, org string) error
+	Mutate(ctx context.Context, m interface{}, input githubql.Input, vars map[string]interface{}) error
+	MutateWithGitHubAppsSupport(ctx context.Context, m interface{}, input githubql.Input, vars map[string]interface{}, org string) error
```
alvaroaleman, line 266: Lets please make this the only addition to the public interface

prow/github/client.go
```diff
@@ -119,6 +119,7 @@ type IssueClient interface {
 	ListOpenIssues(org, repo string) ([]Issue, error)
 	GetIssue(org, repo string, number int) (*Issue, error)
 	EditIssue(org, repo string, number int, issue *Issue) (*Issue, error)
+	TransferIssue(org, dstRepoNodeID string, issueNodeID string) (*TransferIssueMutation, error)
```
alvaroaleman, line 122: I don't think this should be part of the client

prow/plugins/transfer-issue/transfer-issue.go
```diff
@@ -0,0 +1,110 @@
...
+	if err != nil {
+		log.WithError(err).Errorf("could not fetch destination repo: %s", dstRepoPair)
+		// TODO: Might want to add another GetRepo type call that checks if a repo exists vs a bad request
+		return gc.CreateComment(
+			org, srcRepoName, e.Number,
+			plugins.FormatResponseRaw(e.Body, e.HTMLURL, user, fmt.Sprintf("Something went wrong or the destination repo %s does not exist.", dstRepoPair)),
+		)
+	}
+
+	isCollaboratorSrc, err := gc.IsCollaborator(org, srcRepoName, user)
```
alvaroaleman, line 84: We generally use this check to determine if we trust a user: https://github.com/kubernetes/test-infra/blob/0757032ed149f086f3eb69efe20142c75b371450/prow/plugins/welcome/welcome.go#L122

prow/plugins/transfer-issue/transfer-issue.go
```diff
@@ -0,0 +1,110 @@
...
+	if err != nil {
+		log.WithError(err).Errorf("could not fetch if user: %s is collaborator of source repo: %s", user, srcRepoPair)
```
alvaroaleman, line 86: Please do not use `Errorf`. Instead of logging the error here, we could wrap it and log it at the top level (which `hook` already does I believe)

prow/plugins/transfer-issue/transfer-issue.go
```diff
@@ -0,0 +1,110 @@
...
+		return err
+	}
+	isCollaboratorDst, err := gc.IsCollaborator(org, dstRepoName, user)
+	if err != nil {
+		log.WithError(err).Errorf("could not fetch if user: %s is collaborator of destination repo: %s", user, dstRepoPair)
+		return err
+	}
+	if !(isCollaboratorSrc && isCollaboratorDst) {
+		log.Warnf("user %s is not a collaborator of %s and/or %s", user, srcRepoPair, dstRepoPair)
```
alvaroaleman, line 95: This doesn't warrant a warning or a log at all, as there is nothing an administrator of Prow can or should do about this

prow/github/client.go
eddiezane: Should it be part of the top level `Client` instead of the `IssueClient`? If it's not part of the GitHub interface I'd need to also remove it from mine here https://github.com/kubernetes/test-infra/pull/23153/files#diff-57052e6c3c1188cdb7b7d007f9a17201f1bec3747e55ea28ecfedc0e9306793aR25

alvaroaleman: No, expose `MutateWithGitHubAppsSupport`, add the `TransferIssueMutation` to your plugin and call `MutateWithGitHubAppsSupport` there, similiar how we do it for `Query`

prow/plugins/transfer-issue/transfer-issue.go
eddiezane: I can switch to that but there's a mix of different usages. Is `trigger.TrustedUser` the way forward? If so I'll open an issue to track that refactor.

[lgtm: IsCollaborator](https://github.com/kubernetes/test-infra/blob/7af3c135a4cceb136a71516a86f2af94e030c86c/prow/plugins/lgtm/lgtm.go#L283)

[label: IsMember](https://github.com/kubernetes/test-infra/blob/7af3c135a4cceb136a71516a86f2af94e030c86c/prow/plugins/label/label.go#L190)

[lifecycle: IsCollaborator](https://github.com/kubernetes/test-infra/blob/7af3c135a4cceb136a71516a86f2af94e030c86c/prow/plugins/lifecycle/reopen.go#L54)

[releasenote: IsMember](https://github.com/kubernetes/test-infra/blob/7af3c135a4cceb136a71516a86f2af94e030c86c/prow/plugins/releasenote/releasenote.go#L138)

[sigmention: IsMember](https://github.com/kubernetes/test-infra/blob/7af3c135a4cceb136a71516a86f2af94e030c86c/prow/plugins/sigmention/sigmention.go#L150)

alvaroaleman: Hm good point, so `trigger.TrustedUser` is for running tests which is not necessarily the same trust level we want for this. So not sure. Note that `IsCollaborator` is the widest possible set, as it includes both org members and ppl that are  collaborators on the repo itself.

prow/plugins/transfer-issue/transfer-issue.go
```diff
@@ -0,0 +1,149 @@
...
+			plugins.FormatResponseRaw(e.Body, e.HTMLURL, user, "/transfer-issue must only be used once and with a single destination repo."),
+		)
+	}
+
+	dstRepoName := matches[0][1]
+	dstRepoPair := org + "/" + dstRepoName
+
+	dstRepo, err := gc.GetRepo(org, dstRepoName)
+	if err != nil {
+		log.WithError(err).Errorf("could not fetch destination repo: %s", dstRepoPair)
```
petr-muller, line 95: We generally prefer structured logs (`WithField`/`WithFields`) over `Errorf`. Also, this is processing user input, so IMO ti warrants a warning at best, otherwise logs would be spammed by typos etc.

eddiezane: I'll switch it to a warning since `GetRepo`'s failure doesn't differentiate be a not found and a failed request.

petr-muller (approved):
LGTM, thanks!

eddiezane:
I'll get this tested irl later today.

eddiezane:
It works!

https://github.com/kuberneddies/test-repo-2/issues/2



/assign @alvaroaleman

PR #23166, opened 2021-08-06
Protect kubernetes/enhancements

kikisdeliveryservice:
This should be sufficient to disallow manual merges of unapproved PRs in k/enhancments repo.

@cblecker @spiffxp I think this is what was suggested in our slack convo?

LMK if I need to do anything else?

### DIFF
--- a/config/prow/config.yaml
+++ b/config/prow/config.yaml
@@ -246,6 +246,10 @@ branch-protection:
             - continuous-integration/travis-ci
           required_pull_request_reviews:
             required_approving_review_count: 1
+        enhancements:
+          restrictions: # only allow admins
+            users: []
+            teams: []
         git-sync:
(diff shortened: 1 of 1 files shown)


config/prow/config.yaml
```diff
@@ -246,6 +246,10 @@ branch-protection:
             - continuous-integration/travis-ci
           required_pull_request_reviews:
             required_approving_review_count: 1
+        enhancements:
+            restrictions: # only allow admins
```
spiffxp, line 250:
```suggestion
          restrictions: # only allow admins
```

spiffxp (changes requested):
I suspect this is what's throwing checkconfig off

cblecker (approved):
/lgtm
/approve

PR #23154, opened 2021-08-06
config/jobs: misc fixes for k8s-infra triage jobs

spiffxp:
Related:
- Part of: https://github.com/kubernetes/k8s.io/issues/1305
- Followup to: https://github.com/kubernetes/test-infra/pull/23126
- Companion PR (no dependency though): https://github.com/kubernetes/k8s.io/pull/2479

PR #23168, opened 2021-08-06
triage: prepare for alternate projects, datasets, buckets

spiffxp:
Related:
- part of: https://github.com/kubernetes/k8s.io/issues/1305
- followup to: https://github.com/kubernetes/test-infra/pull/23154

Turns out `k8s-gubernator` was hardcoded in a bunch of places in the `update_summaries` script that kicks off the triage go binary

Add env vars with defaults that are still easily greppable, to allow the canary job to override:
- the project billed for bq usage
- the dataset:table to query for build info
- the dataset:table to write temp test results
- the gs://path/to/temp_test_results
- the gs://path/to/triage

Then setup the canary job to override these. It will still fail because it can't write to k8s-gubernator's datasets, but I'd like to see it get that far first.

### DIFF
--- a/triage/interactive.js
+++ b/triage/interactive.js
@@ -303,7 +303,7 @@ function getData() {
     setElementVisibility('btn-sig-group', false);
   }
 
-  var url = '/k8s-gubernator/triage/';
+  var url = '/triage/';
   if (document.location.host == 'storage.googleapis.com' && document.location.pathname.endsWith('index.html')) {
     // Use the bucket name where available
     var pathname = document.location.pathname;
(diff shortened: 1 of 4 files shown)


triage/interactive.js
```diff
@@ -303,7 +303,7 @@ function getData() {
     setElementVisibility('btn-sig-group', false);
   }
 
-  var url = '/k8s-gubernator/triage/';
+  var url = '/triage/';
```
ameukam, line 306: Not `/k8s-triage/triage` ?

ameukam:
Just one non-blocking question. Otherwise:
/lgtm
/hold
Remove hold when ready.

triage/interactive.js
spiffxp: I thought it was for the benefit of running locally, but it didn't seem to make a difference either way.  It could probably be `url=''` but I was mainly interested in getting the bucket name out of the source

spiffxp:
Postsubmits still failed off of the merge commit for this, e.g.

https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-test-infra-push-triage/1424767317648084992
>ERROR: (gcloud.builds.submit) INVALID_ARGUMENT: generic::invalid_argument: key "_PULL_BASE_REF" in the substitution data is not matched in the template

So I kept triggering cloudbuild with a locally-built version of image-builder until I got something that works: https://github.com/kubernetes/test-infra/pull/23180

PR #23180, opened 2021-08-09
triage: fix postsubmits

spiffxp:
Related:
- part of: https://github.com/kubernetes/k8s.io/issues/1305
- followup to: https://github.com/kubernetes/test-infra/pull/23168

Changes to fix the upload job:

- don't set ACLs when uploading static files, to support buckets that
  have UBLA enabled

Changes to fix the cloudbuild job:

- build using go1.16
- add dummy sub for _PULL_BASE_REF
- source all triage files from build container in second stage
- add go.mod to support building solely from triage/ dir
- run hack/update-deps.sh


ameukam:
late /lgtm

PR #23188, opened 2021-08-10
Run tests for Cluster API across multiple kubernetes versions

vincepri:

### DIFF
--- a/config/jobs/kubernetes-sigs/cluster-api/cluster-api-presubmits-main.yaml
+++ b/config/jobs/kubernetes-sigs/cluster-api/cluster-api-presubmits-main.yaml
@@ -110,6 +110,36 @@ presubmits:
     annotations:
       testgrid-dashboards: sig-cluster-lifecycle-cluster-api
       testgrid-tab-name: capi-pr-test-main
+  - name: pull-cluster-api-test-main-mink8s
+    decorate: true
+    path_alias: sigs.k8s.io/cluster-api
+    always_run: true
+    labels:
+      preset-service-account: "true"
+    branches:
+    # The script this job runs is not in all branches.
+    - ^master$
+    - ^operator-0.4$
+    spec:
+      containers:
+      - image: gcr.io/k8s-testimages/kubekins-e2e:v20210808-1eaeec7-1.21
+        args:
+        - runner.sh
+        - ./scripts/ci-test.sh
+        env:
+        # This value determines the minimum Kubernetes
+        # supported version for Cluster API management cluster.
+        #
+        # To check the available envtest in Kubebuilder, please
+        # refer to https://github.com/kubernetes-sigs/kubebuilder/tree/tools-releases.
+        - name: KUBEBUILDER_ENVTEST_KUBERNETES_VERSION
+          value: "1.19.2"
+        resources:
+          requests:
+            cpu: 7300m
+    annotations:
+      testgrid-dashboards: sig-cluster-lifecycle-cluster-api
+      testgrid-tab-name: capi-pr-test-main-mink8s
   - name: pull-cluster-api-e2e-main
     labels:
       preset-dind-enabled: "true"


config/jobs/kubernetes-sigs/cluster-api/cluster-api-presubmits-main.yaml
```diff
@@ -104,6 +104,84 @@ presubmits:
...
+    - ^operator-0.4$
+    spec:
+      containers:
+      - image: gcr.io/k8s-testimages/kubekins-e2e:v20210808-1eaeec7-1.21
+        args:
+        - runner.sh
+        - ./scripts/ci-test.sh
+        env:
+        - name: KUBEBUILDER_ENVTEST_KUBERNETES_VERSION
+          value: "1.21.2"
```
CecileRobertMichon, line 134: is there anyway this could pull the latest 1.21.x version instead? so we don't have to update this spec everytime a patch comes out (or risk staying behind, this is already not the latest)

fabriziopandini: We have somethings that can allow using version labels (e.g stable-1.21) in https://github.com/kubernetes-sigs/cluster-api/blob/b7aa6138aeeb8fd3094c3fe481f5af44576942e0/scripts/ci-e2e-lib.sh#L80

CecileRobertMichon: this is how we do it in CAPZ https://github.com/chewong/cluster-api-provider-azure/blob/master/test/e2e/helpers.go#L565

vincepri: These are coming from kubebuilder, which are built ad-hoc https://github.com/kubernetes-sigs/kubebuilder/tree/tools-releases/build

CecileRobertMichon: what would be the process to check if newer versions are available?

vincepri: I usually go in the repository to see if there is a newer version, I don't think there is any script to do so today

CecileRobertMichon: could we document this as a comment so others know where this version is coming from?

vincepri: Sure thing!

config/jobs/kubernetes-sigs/cluster-api/cluster-api-presubmits-main.yaml
```diff
@@ -104,6 +104,84 @@ presubmits:
...
         - ./scripts/ci-test.sh
+        env:
+        - name: KUBEBUILDER_ENVTEST_KUBERNETES_VERSION
+          value: "1.22.0"
+        resources:
+          requests:
+            cpu: 7300m
+    annotations:
+      testgrid-dashboards: sig-cluster-lifecycle-cluster-api
+      testgrid-tab-name: capi-pr-test-main
```
sbueringer, line 115: If my assumptions about how testgrid dashboards are correct, this would mean the job results are all in the same dashboard. Do we want to split that up to have one dashboard per job? (Imho this makes it easier to identify Kubernetes-version specific job failures (especially flakes))

vincepri:
@CecileRobertMichon @fabriziopandini Updated the PR to only add the min k8s version we support

config/jobs/kubernetes-sigs/cluster-api/cluster-api-presubmits-main.yaml
sbueringer: *ping* @vincepri  :)

sbueringer: Ah according to the failed test we *have* to change the tab name because duplicate names are not supported

PR #23192, opened 2021-08-10
test: adds job for e2e provider

nilekhc:
Signed-off-by: Nilekh Chaudhari <1626598+nilekhc@users.noreply.github.com>

### DIFF
--- a/config/jobs/kubernetes-sigs/secrets-store-csi-driver/secrets-store-csi-driver-config.yaml
+++ b/config/jobs/kubernetes-sigs/secrets-store-csi-driver/secrets-store-csi-driver-config.yaml
@@ -167,6 +167,39 @@ presubmits:
       testgrid-tab-name: pr-secrets-store-csi-driver-e2e-azure
       description: "Run e2e test with azure provider for Secrets Store CSI driver."
       testgrid-num-columns-recent: '30'
+  - name: pull-secrets-store-csi-driver-e2e-provider
+    decorate: true
+    decoration_config:
+      timeout: 25m
+    always_run: true
+    optional: true # change to false once the provider job is stable
+    path_alias: sigs.k8s.io/secrets-store-csi-driver
+    branches:
+    - master
+    labels:
+      # this is required because we want to run kind in docker
+      preset-dind-enabled: "true"
+      # this is required to make CNI installation to succeed for kind
+      preset-kind-volume-mounts: "true"
+      # sets up the azure keyvault parameters used for testing
+      preset-azure-secrets-store-creds: "true"
+    spec:
+      containers:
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes-sigs/secrets-store-csi-driver/secrets-store-csi-driver-config.yaml
```diff
@@ -167,6 +167,39 @@ presubmits:
...
+    path_alias: sigs.k8s.io/secrets-store-csi-driver
+    branches:
+    - master
+    labels:
+      # this is required because we want to run kind in docker
+      preset-dind-enabled: "true"
+      # this is required to make CNI installation to succeed for kind
+      preset-kind-volume-mounts: "true"
+      # sets up the azure keyvault parameters used for testing
+      preset-azure-secrets-store-creds: "true"
```
tam7t, line 185: should the fake provider have azure creds?

nilekhc: fixed

config/jobs/kubernetes-sigs/secrets-store-csi-driver/secrets-store-csi-driver-config.yaml
```diff
@@ -167,6 +167,37 @@ presubmits:
...
+    spec:
+      containers:
+      - image: gcr.io/k8s-testimages/kubekins-e2e:v20210808-1eaeec7-master
+        command:
+        - runner.sh
+        args:
+        - bash
+        - -c
+        - >-
+          make e2e-bootstrap e2e-fake-provider-container e2e-helm-deploy e2e-provider-deploy e2e-provider
```
aramase, line 193: Just `e2e-provider-container` instead of `e2e-fake-*` should be sufficient. We should follow a single convention across all: `e2e-provider` or `e2e-fake-provider`. WDYT?

aramase: Just a thought, we should consider replacing the `args` with just a script for the `e2e-provider`. The script can wrap all the logic required for testing the `e2e-provider`. Having individual targets here will make it harder for us to keep track and also require changes to test-infra config for any changes in the make target names in the driver.

nilekhc: @aramase  There is already another target with same name as `e2e-provider-container`. I preferred not to use `fake` in this implementation except this one target name. Happy to change it to another suitable name 😄 .

Also, good idea about single script. Let me take a look. Thanks!

PR #23195, opened 2021-08-10
Add a gcloud-terraform image and corresponding publishing job.

cjwagner:
/assign @chaodaiG 

I think this has all the utilities we'll need for validating terraform config and applying. I've confirmed that the Dockerfile builds locally.

### DIFF
--- a/images/gcloud-terraform/Dockerfile
+++ b/images/gcloud-terraform/Dockerfile
@@ -1,0 +1,51 @@
+# Copyright 2021 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


images/gcloud-terraform/Dockerfile
```diff
@@ -0,0 +1,51 @@
+# Copyright 2019 The Kubernetes Authors.
```
chaodaiG, line 1: 2021

images/gcloud-terraform/Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+    /google-cloud-sdk/install.sh \
+        --disable-installation-options \
+        --bash-completion=false \
+        --path-update=false \
+        --usage-reporting=false && \
+    gcloud components install alpha beta kubectl && \
+    gcloud info | tee /workspace/gcloud-info.txt
+
+# Install terraform
+RUN apk add --no-cache terraform
```
chaodaiG, line 49: can we add a version here?

cjwagner: Done (although I had to pick a newer version than the one specified in the README due to availability).

images/gcloud-terraform/Dockerfile
cjwagner: Fixed.

chaodaiG (approved):

PR #23214, opened 2021-08-11
Increase log verbosity in gce-master-scale-performance

jupblb:
Increase kube-controller-manager verbosity so that we can gather more information about DaemonSet failures.

This PR is part of debugging [the most recently failed test](https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/ci-kubernetes-e2e-gce-scale-performance/1418254982817255424).

/assign @mborsz


mborsz:
As discussed offline, let's try to debug that failed run without increasing log verbosity (the issue is hard to reproduce).
/close

PR #23228, opened 2021-08-12
Run e2e tests for Cluster API also with v1.19 management cluster

sbueringer:

Partially implements https://github.com/kubernetes-sigs/cluster-api/issues/5078

As long as we're already sure about the variable name, this PR can be merged independently of: https://github.com/kubernetes-sigs/cluster-api/pull/5081

### DIFF
--- a/config/jobs/kubernetes-sigs/cluster-api/cluster-api-periodics-main.yaml
+++ b/config/jobs/kubernetes-sigs/cluster-api/cluster-api-periodics-main.yaml
@@ -53,6 +53,41 @@ periodics:
     testgrid-tab-name: capi-e2e-main
     testgrid-alert-email: sig-cluster-lifecycle-cluster-api-alerts@kubernetes.io
     testgrid-num-failures-to-alert: "2"
+- name: periodic-cluster-api-e2e-main-mink8s
+  interval: 1h
+  decorate: true
+  labels:
+    preset-dind-enabled: "true"
+    preset-kind-volume-mounts: "true"
+  extra_refs:
+  - org: kubernetes-sigs
+    repo: cluster-api
+    base_ref: master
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes-sigs/cluster-api/cluster-api-periodics-main.yaml
```diff
@@ -53,6 +53,44 @@ periodics:
...
+      env:
+      - name: GINKGO_SKIP
+        value: "\\[Conformance\\] \\[K8s-Upgrade\\]"
+      # This value determines the minimum Kubernetes
+      # supported version for Cluster API management cluster.
+      #
+      # We should use the latest Kubernetes patch version with a published image, please
+      # refer to https://hub.docker.com/r/kindest/node/tags?page=1&ordering=last_updated
+      - name: KUBERNETES_VERSION_MANAGEMENT
+        value: "v1.19.11"
```
fabriziopandini, line 82: Q: should this be `stable-1.19` (if the kind image is not published, we are building it locally)

sbueringer: @fabriziopandini This would roughly double the test execution time imho. I guess that's not too critical in periodic tests.

Which variant do you prefer?

fabriziopandini: Lets use the release label, otherwise we will have to maintain the version label in sync.

If the duration is long of we have a problem in resources, we can always move periodic tests from running every hours to run every 2h (or 4h), that IMO it is still reasonable for the project needs

sbueringer: Sounds good to me, fixed

Issue #21330, opened 2021-03-12
Sidecar censoring tests do not pass locally.


cjwagner:
**What happened**:
```
INFO: From Testing //prow/sidecar:go_default_test:
==================== Test output for //prow/sidecar:go_default_test:
--- FAIL: TestArchiveMatchesTar (0.00s)
    censor_test.go:170: failed to archive input: could not walk source files to archive them: could not tar file: archive/tar: write too long
--- FAIL: TestRoundTrip (0.00s)
    censor_test.go:213: failed to archive input: could not walk source files to archive them: could not tar file: archive/tar: write too long
time="2021-03-11T23:45:27Z" level=error msg="cannot read /tmp/errors go into sidecar-errors854995690/metadata-2.txt" error="read /tmp/errors go into sidecar-errors854995690/metadata-2.txt: is a directory"
time="2021-03-11T23:45:27Z" level=error msg="Failed to unmarshal /tmp/errors go into sidecar-errors854995690/metadata-3.txt" error="invalid character 'j' looking for beginning of value"
time="2021-03-11T23:45:27Z" level=error msg="Failed to open /tmp/note when a part has a problem823745470/test2-log.txt" error="open /tmp/note when a part has a problem823745470/test2-log.txt: no such file or directory"
FAIL
````

**Anything else we need to know?**:
I'm not sure how to reproduce this or why it isn't failing in CI. I'm hoping that Steve might have some ideas about what could be causing this.

/assign @stevekuznetsov 
/area prow/sidecar

**Comment by bbguimaraes** [comment]
Any chance you have files of the special types listed here (hard/symbolic links, most likely) somehow in your source tree?

https://github.com/golang/go/blob/release-branch.go1.16/src/archive/tar/writer.go#L428-L430

```
$ go test --count 1 --run TestArchiveMatchesTar ./prow/sidecar/
ok      k8s.io/test-infra/prow/sidecar  0.018s
$ ln -s quote.txt prow/sidecar/testdata/archives/archive/link
$ go test --count 1 --run TestArchiveMatchesTar ./prow/sidecar/
--- FAIL: TestArchiveMatchesTar (0.00s)
    censor_test.go:170: failed to archive input: could not walk source files to archive them: could not tar file: archive/tar: write too long
FAIL
FAIL    k8s.io/test-infra/prow/sidecar  0.015s
FAIL
```

**Comment by stevekuznetsov** [comment]
A new clone of the repo seems to work as well:

```
[...]

PR #23237, opened 2021-08-13
censor: archive: handle symlinks

listx:
This change lets

    bazel test //prow/sidecar:go_default_test

pass locally.

According to https://stackoverflow.com/a/40003617, non-regular files can
break us if we don't call FileInfoHeader() with the right
argument (e.g., for symlinks). The change here merely lifts the change
suggested in the link above.

### DIFF
--- a/prow/sidecar/censor.go
+++ b/prow/sidecar/censor.go
@@ -323,7 +323,18 @@ func archive(srcDir, destArchive string) error {
 		if err != nil {
 			return err
 		}
-		header, err := tar.FileInfoHeader(info, info.Name())
+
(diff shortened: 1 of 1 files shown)


listx:
Fixes #21330

cjwagner:
Nice work, Linus! This fixes the issue for me locally too. I'd like Alvaro or Steve to confirm that this doesn't compromise censoring behavior, but it doesn't look like it would to me unless the symlink included a secret in the linked path. Am I understanding that right and is that something we're concerned with? 
/assign @stevekuznetsov @alvaroaleman

alvaroaleman (approved):

PR #23242, opened 2021-08-13
add `always_run` field to Postsubmit jobs

### LINKED ISSUE #23234: Postsubmit prow jobs respect `always_run: false`

Postsubmit prow jobs can use `always_run: false` so that it's not triggered automatically

**Why is this needed**:

In the past, there was not a use case of this, as `/test` doesn't work with postsubmit jobs. Now that postsubmit jobs can be triggered by pubsub, there are cases where certain postsubmit jobs are expected to only be triggered by pubsub, and `always_run: false` would be handy here.

A little detail: presubmit handles `always_run` at https://www.google.com/url?sa=D&q=https%3A%2F%2Fgithub.com%2Fkubernetes%2Ftest-infra%2Fblob%2F0e507b656399b836f38bfa355d1aa358dfb92661%2Fprow%2Fconfig%2Fjobs.go%23L359, which is not in postsubmit https://www.google.com/url?sa=D&q=https%3A%2F%2Fgithub.com%2Fkubernetes%2Ftest-infra%2Fblob%2F0e507b656399b836f38bfa355d1aa358dfb92661%2Fprow%2Fconfig%2Fjobs.go%23L333.

When implementing this, please make sure that the default behavior doesn't change. If `always_run` doesn't exist, it should be treated as true

listx:
If this field does not exist, we preserve the existing behavior of
defaulting to `true` in ShouldRun(). However if it is explicitly set,
the behavior of ShouldRun() depends on the value of `always_run`:

- true: return true _before_ considering RegexpChangeMatcher (this
  matches existing ShouldRun() behavior for Presubmit jobs)
- false: return false, but only as a default value

/cc @chaodaiG @cjwagner 
I'm not sure if this is the actual behavior we want (see `ShouldRun()`), but it's a start.
Fixes #23234


chaodaiG (commented):
/hold

This probably is the best solution I can think about at the moment too. There are several catches:
- This change implicitly sets `always_run` as true, while not exactly the same behavior, as it can still be overriden by `RegexpChangeMatcher`, this is different from how it's handled in presubmits
- Presubmit doesn't allow both `always_run` and `run_if_changed` to be supplied https://github.com/kubernetes/test-infra/blob/16f520d43f04ac1c480a9799bacc7fa6ee7c1396/prow/config/config.go#L2420, how to handle this in postsubmits is not easy (aka retro-fitting is hard), as it might break existing postsubmit jobs with `run_if_changed` set

chaodaiG:
@listx , as discussed in person, prow does static config checking to make sure that presubmit jobs don't have both `always_run` and `run_if_changed` configured. To make things consistent, please make configcheck also checks this for postsubmit jobs.

Presubmit is enforced at https://github.com/kubernetes/test-infra/blob/16f520d43f04ac1c480a9799bacc7fa6ee7c1396/prow/config/config.go#L2420

cjwagner (commented):
Looks great. LGTM after Chao's comment is addressed to prevent `run_if_changed` from being used with an explicitly set `always_run: true`.

listx:
I added some more tests. Sometimes it felt a bit kludgy (esp. the almost verbatim copied-over tests, e.g., `TestRunIfChangedPostsubmits`). Also, I stopped myself from refactoring some small naming changes to existing code, so that I could keep the PR smaller.

Any and all comments welcome; PTAL.

listx:
Thinking about this some more, should explicitly setting `always_run: ...` (true or false) paired with `run_if_changed`/`skip_if_only_changed` be an error? That would simplify some of these tests.

For `ShouldRun()` I'm a little bit worried about user expectations when `run_if_changed`/`skip_if_only_changed` is set, but `always_run` is (1) not mentioned at all or (2) explicitly set to `false`. In case (1) we could still default to "true" (line 380 of `ShouldRun()` and in case (2) we will return "false" (line 375 of `ShouldRun()`).

This theoretical divergence between case (1) and (2) is bothering me. But what really concerns me is that if the user sets `run_if_changed`/`skip_if_only_changed`, then just setting those fields alone makes it feel like `always_run` **should be automatically set to `false`**. But we currently don't do this --- should we though?

listx:
IOW, should the following 2 scenarios behave exactly the same way?

```
run_if_changed: "foo"
```
(`always_run` field missing)

vs

```
always_run: false
run_if_changed: "foo"
```

?

chaodaiG (commented):
LGTM, thank you!

imo this is a breaking change, can we add a line under https://github.com/kubernetes/test-infra/blob/master/prow/ANNOUNCEMENTS.md#breaking-changes?

chaodaiG:
sorry that I had overlooked your comments before.

> Thinking about this some more, should explicitly setting always_run: ... (true or false) paired with run_if_changed/skip_if_only_changed be an error? That would simplify some of these tests.

These should be errors, because their behaviors would contradict, for example when `always_run` is true but `run_if_changed` doesn't match, which decision should prow make?

listx:
> sorry that I had overlooked your comments before.
> 
> > Thinking about this some more, should explicitly setting always_run: ... (true or false) paired with run_if_changed/skip_if_only_changed be an error? That would simplify some of these tests.
> 
> These should be errors, because their behaviors would contradict, for example when `always_run` is true but `run_if_changed` doesn't match, which decision should prow make?

Ack, that will make this PR a lot easier to swallow for me. So I will make the above situation an error and also add a note in the Announcements per your earlier comment.

chaodaiG:
> Ack, that will make this PR a lot easier to swallow for me. So I will make the above situation an error and also add a note in the Announcements per your earlier comment.

I thought you already made this an error, did I misunderstand?

listx:
> > Ack, that will make this PR a lot easier to swallow for me. So I will make the above situation an error and also add a note in the Announcements per your earlier comment.
> 
> I thought you already made this an error, did I misunderstand?

I only made it an error if `always_run: true` is set and `run_if_changed` or `skip_if_only_changed` are set. I need to make it so that explicitly setting `always_run` (regardless whether it is true or false) and setting the run/skip ones will be an error.

chaodaiG:
> I only made it an error if `always_run: true` is set and `run_if_changed` or `skip_if_only_changed` are set. I need to make it so that explicitly setting `always_run` (regardless whether it is true or false) and setting the run/skip ones will be an error.

it should be good as long as the logic is the same as in presubmit jobs

listx:
> > I only made it an error if `always_run: true` is set and `run_if_changed` or `skip_if_only_changed` are set. I need to make it so that explicitly setting `always_run` (regardless whether it is true or false) and setting the run/skip ones will be an error.
> 
> it should be good as long as the logic is the same as in presubmit jobs

FTR the presubmit only performs a mutual exclusivity check when the `always_run` field is set to true explicitly (see https://github.com/kubernetes/test-infra/blob/d917e4a1e773ccf701ac58424a0b4ee3490e8dd1/prow/config/config.go#L2430-L2437). Because it's a regular boolean, when we unmarshal from the YAML there is no way to tell the difference between the user adding a `always_run: false` or not specifying it at all. That is, currently for Presubmits you could define

```
always_run: false
run_if_changed: "foo"
```

and this would be a valid configuration AFAICS. What I did so far is to match this behavior in the postsubmit. A user *could* define the same thing as above, and it will not be an error. However I am about to change it so that it will be an error (so technically we will diverge vs presubmit behavior). SG?

chaodaiG:
Would be nice not to diverge, so maybe both use pointers or not, sorry if I had mislead you. 

technically speaking `always_run: false` doesn't conflict with `run_if_changed`, so this might be fine? WDYT?

listx:
> Would be nice not to diverge, so maybe both use pointers or not, sorry if I had mislead you.

No worries!

> 
> technically speaking `always_run: false` doesn't conflict with `run_if_changed`, so this might be fine? WDYT?

This was my initial thinking. But it would simplify tests if we just say "if you specify `run_if_changed`, you cannot specify `always_run` at all". We have an opportunity to do this for Postsubmits because we're introducing the field for the first time.

But I'm not sure about pointer-izing the field for Presubmits to align it with the above approach. It looks like we already treat as normal (at least in our docs https://github.com/kubernetes/test-infra/blame/961a6a42f0c3a24c6578b836ea5fb8f87dafdd48/prow/jobs.md#L192-L211) setting both `run_if_changed` and `always_run: false`. So I assume there are many Presubmit configs out there that would be broken if we tried to make it match the possible simpler semantics in the Postsubmits.

So in summary it looks like we have 2 options:

(A) keep existing logic here and allow `always_run: false` (explicit) with `run_if_changed` or `skip_if_only_changed`, matching Presubmit logic. Makes tests a bit finicky but it's doable. This is the current state of this PR.

(B) forbid `always_run: ...` with `run_if_changed` or `skip_if_only_changed`, but diverge with Presubmit configs. Makes tests simpler because there are less edge cases to worry about. This is what I was tentatively going to do on my next update.

Maybe we should wait for others to chime in?

chaodaiG:
> So I assume there are many Presubmit configs out there that would be broken if we tried to make it match the possible simpler semantics in the Postsubmits.

agreed

my vote would be on (A), and totally agreed that can wait for others to chime in here

cjwagner:
I think either option is fine. (A) is slightly preferable for consistency and being more permissive.

listx:
PTAL

prow/config/jobs.go
```diff
@@ -334,12 +345,38 @@ func (ps Postsubmit) ShouldRun(baseRef string, changes ChangedFilesProvider) (bo
...
+	// At this point neither `run_if_changed` nor `skip_if_only_changed` were
+	// set. We're left with 2 cases: (1) `always_run: false` was provided, or
+	// (2) this field was not defined in the job at all. Only in the second case
+	// do we return "true".
+
+	// If the `always_run` field was explicitly set to false, override the
+	// default `true` response.
+	if ps.AlwaysRun != nil && !*ps.AlwaysRun {
+		return false, nil
+	}
```
cjwagner, line 376: nit: If we change this to 
```suggestion
	if ps.AlwaysRun != nil {
		return *ps.AlwaysRun, nil
	}
```
we don't need the other AlwaysRun check before the regexp change check. I think that will make this easier to understand.

prow/config/jobs_test.go
```diff
@@ -244,6 +244,39 @@ func TestRunIfChangedPresubmits(t *testing.T) {
 	}
 }
 
+// TestRunIfChangedPostsubmits is identical to TestRunIfChangedPostsubmits.
```
cjwagner, line 247: technically true, but a typo :smile:

cjwagner (commented):
Looks great. I spotted a typo and have a suggestion for a small simplification, but otherwise LGTM.

prow/config/jobs.go
listx: Agreed. FTR the only reason I did it the current way is because I wanted to align as closely as possible to the behavior of `ShouldRun` for Presubmits. There we perform an `if ps.AlwaysRun` check *before* we examine `ps.RegexpChangeMatcher` (see line 392). I assume your suggestion still stands?

PR #22758, opened 2021-07-01
Add e2e lock contention job

ipochi:
This PR adds a new job `ci-kubernetes-node-kubelet-lock-contention`
to the test suite.

This test is created due to the want of `--lock-file` and
`--exit-on-lock-contention` be moved as flags and into the Kubelet
configuration file rather than be dropped.

Adds a new tab in TestGrid named `kubelet-gce-e2e-lock-contention`
running tests focused on `NodeFeature:LockContention`.

This PR  is created to supercede #20106 

Requesting creator of the original PR #20106 @knabben to advise on further steps on his PR. Could #20106 be closed in favor of this ?

Requesting @ike-ma @SergeyKanzhelev , the reviewers of the original PR to please review this. 

@ike-ma I've addressed your https://github.com/kubernetes/test-infra/pull/20106#discussion_r617114498 and for this  https://github.com/kubernetes/test-infra/pull/20106#discussion_r617125341, I'd like to let you know that https://github.com/kubernetes/kubernetes/pull/97028 is now merged which was a pre-requisite for this PR to get merged.

cc: @rata 

Signed-off-by: Imran Pochi <imran@kinvolk.io>

PR #23243, opened 2021-08-13
Add E2E test for lock contention flags

ipochi:
Adds a new job ci-kubernetes-node-kubelet-lock-contention to the test suite.

This test is created due to the need of `--lock-file` and `--exit-on-lock-contention` 
be moved as flags and into the Kubelet configuration file rather than be dropped.

Adds a new tab in TestGrid named `kubelet-gce-e2e-lock-contention`
running tests focused on `NodeSpecialFeature:LockContention`.

This test was earlier added in the PR: https://github.com/kubernetes/test-infra/pull/22758

However it was later removed in this [commit](https://github.com/kubernetes/test-infra/commit/1b54c26a6ec7e483dcc4c07e81ae96c71a08b269#diff-14f46258d64962b171c1662b99c4e01ea48674ca3ef00b1d9dec13acf2ad91f9)

Corresponding e2e test is at https://github.com/kubernetes/kubernetes/pull/104334

Signed-off-by: Imran Pochi <imran@kinvolk.io>

### DIFF
--- a/config/jobs/kubernetes/sig-node/containerd.yaml
+++ b/config/jobs/kubernetes/sig-node/containerd.yaml
@@ -1274,3 +1274,32 @@ periodics:
     testgrid-tab-name: node-kubelet-containerd-performance-test
     testgrid-alert-email: kubernetes-sig-node-test-failures@googlegroups.com
     description: "Executes performance e2e tests"
+- name: ci-kubernetes-node-kubelet-lock-contention
+  interval: 4h
+  labels:
+    preset-service-account: "true"
+    preset-k8s-ssh: "true"
+  spec:
+    containers:
+      - image: gcr.io/k8s-staging-test-infra/kubekins-e2e:v20220211-6df2c53026-master
+        args:
+          - --repo=k8s.io/kubernetes=master
+          - --timeout=240
+          - --root=/go/src
+          - --scenario=kubernetes_e2e
+          - --
+          - --deployment=node
+          - --gcp-project-type=node-e2e-project
+          - --gcp-zone=us-west1-b
+          - --node-args=--image-config-file=/workspace/test-infra/jobs/e2e_node/image-config.yaml
+          - --node-test-args=--kubelet-flags="--exit-on-lock-contention --lock-file=/var/run/kubelet.lock"
+          - --node-tests=true
+          - --provider=gce
+          - --test_args=--nodes=1 --focus="\[NodeSpecialFeature:LockContention\]" --skip="\[Flaky\]|\[Serial\]"
+        env:
+          - name: GOPATH
+            value: /go
+  annotations:
+    testgrid-dashboards: sig-node-kubelet
+    testgrid-tab-name: kubelet-gce-e2e-lock-contention
+    description: "Contains disruptive tests for the Lock Contention feature."


elfinhe:
seems like all node serial tests will be running with this flag. Is that what we want? btw, pull-kubernetes-node-kubelet-serial and pull-kubernetes-node-kubelet-serial-containerd are not triggered. So we don't know if it will affect existing serial tests. 

maybe we can wait until Monday then do a quick verify, or it may be possible to update run_if_changed to trigger the serial tests to see the results.

ipochi:
> seems like all node serial tests will be running with this flag. Is that what we want? btw, pull-kubernetes-node-kubelet-serial and pull-kubernetes-node-kubelet-serial-containerd are not triggered. So we don't know if it will affect existing serial tests.
> 
> maybe we can wait until Monday then do a quick verify, or it may be possible to update run_if_changed to trigger the serial tests to see the results.

I don't there should be any impact on the existing tests, but it would be a good to trigger the serial tests based on the new change.

I had a question in the corresponding PR, is the question valid and possible ? https://github.com/kubernetes/kubernetes/pull/104334#issuecomment-898280645

ehashman:
/hold

I don't think we want to run all serial tests with this feature enabled... We discussed in SIG Node CI subproject a while back, not sure if there's been any progress there.

ipochi:
> /hold
> 
> I don't think we want to run all serial tests with this feature enabled... We discussed in SIG Node CI subproject a while back, not sure if there's been any progress there.

@ehashman Update the test job to reflect the correct string `NodeSpecialFeature` and added skip section to avoid the test being picked up by `Flaky` or `Serial` test suite.

config/jobs/kubernetes/sig-node/node-kubelet.yaml
```diff
@@ -456,7 +456,7 @@ periodics:
           - --node-test-args=--kubelet-flags="--exit-on-lock-contention --lock-file=/var/run/kubelet.lock"
           - --node-tests=true
           - --provider=gce
-          - --test_args=--nodes=1 --focus="\[NodeFeature:LockContention\]" --skip=""
+          - --test_args=--nodes=1 --focus="\[NodeSpecialFeature:LockContention\]" --skip="\[Flaky\]|\[Serial\]"
```
ehashman, line 459:
```suggestion
          - --test_args=--nodes=1 --focus="\[NodeSpecialFeature:LockContention\]" --skip="\[Flaky\]"
```

If these tests need to be run serially, we should include that on them.

ehashman (commented):
/hold cancel

We should first merge https://github.com/kubernetes/kubernetes/pull/104334 before merging this in case of future changes.

config/jobs/kubernetes/sig-node/node-kubelet.yaml
ipochi: This test need not be run serially, as the test is disruptive and restarts Kubelet.

ipochi:
@endocrimes Updated this PR (title and description as well) to bring back the E2E test as per (your comment)[https://github.com/kubernetes/kubernetes/pull/104334#issuecomment-1040551089].

/cc @ehashman @SergeyKanzhelev

ehashman (commented):
/approve
/lgtm

/hold 
for the test PR to merge first

SergeyKanzhelev (approved):
/unhold

PR #23254, opened 2021-08-16
job history: log errors structured instead of inline

petr-muller:
Use `logrus`'s `WithError` method to log errors in a structured, machine-readable way.

### DIFF
--- a/prow/cmd/deck/job_history.go
+++ b/prow/cmd/deck/job_history.go
@@ -361,7 +361,7 @@ func getBuildData(ctx context.Context, bucket storageBucket, dir string) (buildD
 				break
 			}
 		}
-		logrus.Debugf("failed to read finished.json (job might be unfinished): %v", err)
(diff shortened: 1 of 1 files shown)


prow/cmd/deck/job_history.go
```diff
@@ -495,13 +495,13 @@ func getJobHistory(ctx context.Context, url *url.URL, cfg config.Getter, opener
 			}
 			b, err := getBuildData(ctx, bucket, dir)
 			if err != nil {
-				logrus.Warningf("build %d information incomplete: %v", buildID, err)
+				logrus.WithError(err).Warningf("build %d information incomplete", buildID)
```
chaodaiG, line 498: nit: how about use `WithField` for `buildID`?

chaodaiG (approved):
/hold
feel free unhold by ignoring my nit

PR #23257, opened 2021-08-16
 e2e test images: Adds PULL_BASE_SHA to jobs

claudiubelu:
In the test image build jobs, the ``image-util.sh`` script is not being run in a git repository, which causes git log to fail.

We can use the ``PULL_BASE_SHA`` instead.

Needed By: https://github.com/kubernetes/kubernetes/pull/104393


spiffxp (approved):
/approve
/lgtm

PR #23259, opened 2021-08-16
Skip SCTP tests in gce-master-scale-correctness

jupblb:
Because of https://github.com/kubernetes/kubernetes/issues/104396 we need to disable SCTP tests until that issue is fixed

/assign @mborsz


aojea:
/hold
I think we can fix the tests,  let me try that first

jupblb:
Closing as https://github.com/kubernetes/kubernetes/pull/104401 is merged

PR #23264, opened 2021-08-16
Make Sub cache git clients used for inrepo config.

cjwagner:
This `InRepoConfigGitCache` will reuse `GitRepo` clients in order to avoid a local clone every time an inrepo config lookup is needed. Now the local clone is skipped and we can simply fetch and reuse the repo instead. The cache has no pruning or garbage collection at this time, but I don't think the `localgit` client does either so we shouldn't ever use more than twice the amount of disk space we used before and we may use less disk in high concurrency cases. This should also allow us to reduce the number of deployment replicas so even if each pod uses more disk we'll still likely use less over all.
We are just trying this out on Sub to start with.

/assign @chaodaiG 
/cc @listx @alvaroaleman

### DIFF
--- a/prow/config/inrepoconfig.go
+++ b/prow/config/inrepoconfig.go
@@ -23,6 +23,7 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"sync"
 
 	"github.com/sirupsen/logrus"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
@@ -186,4 +187,72 @@ func DefaultAndValidateProwYAML(c *Config, p *ProwYAML, identifier string) error
 	}
 
 	return utilerrors.NewAggregate(errs)
+}
+
+// InRepoConfigGitCache is a wrapper around a git.ClientFactory that allows for
+// threadsafe reuse of git.RepoClients when one already exists for the specified repo.
+type InRepoConfigGitCache struct {
+	git.ClientFactory
+	cache map[string]*skipCleanRepoClient
+	sync.RWMutex
+}
+
+func NewInRepoConfigGitCache(factory git.ClientFactory) git.ClientFactory {
+	if factory == nil {
+		// Don't wrap a nil git factory, keep it nil so that errors are handled properly.
+		return nil
+	}
+	return &InRepoConfigGitCache{
+		ClientFactory: factory,
+		cache:         map[string]*skipCleanRepoClient{},
+	}
+}
+
+func (c *InRepoConfigGitCache) ClientFor(org, repo string) (git.RepoClient, error) {
+	key := fmt.Sprintf("%s/%s", org, repo)
+	c.RLock()
+	cached, err := func() (git.RepoClient, error) {
+		defer c.RUnlock()
+
+		if client, ok := c.cache[key]; ok {
+			client.Lock()
+			// Don't unlock the client unless we get an error or the consumer indicates they are done by Clean()ing.
+			if err := client.Fetch(); err != nil {
+				client.Unlock()
+				return nil, err
+			}
+			return client, nil
+		}
+		return nil, nil
+	}()
+	if cached != nil || err != nil {
+		return cached, err
+	}
+
+	// The repo client was not cached, create a new one.
+	c.Lock()
+	defer c.Unlock()
+	coreClient, err := c.ClientFactory.ClientFor(org, repo)
+	if err != nil {
+		return nil, err
+	}
+	client := &skipCleanRepoClient{
+		RepoClient: coreClient,
+	}
+	client.Lock()
+	c.cache[key] = client
+	return client, nil
+}
+
+var _ git.RepoClient = &skipCleanRepoClient{}
+
+type skipCleanRepoClient struct {
+	git.RepoClient
+	sync.Mutex
+}
+
+func (rc *skipCleanRepoClient) Clean() error {
+	// Skip cleaning and unlock to allow reuse as a cached entry.
+	rc.Unlock()
+	return nil
 }
--- a/prow/config/inrepoconfig_test.go
+++ b/prow/config/inrepoconfig_test.go
@@ -20,8 +20,10 @@ import (
 	"errors"
 	"fmt"
 	"testing"
+	"time"
 
 	"k8s.io/test-infra/prow/git/localgit"
+	"k8s.io/test-infra/prow/git/v2"
 	"k8s.io/test-infra/prow/kube"
 )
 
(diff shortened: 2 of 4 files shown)


prow/config/inrepoconfig.go
```diff
@@ -187,3 +188,46 @@ func DefaultAndValidateProwYAML(c *Config, p *ProwYAML, identifier string) error
...
+}
+
+func NewInRepoConfigGitCache(factory git.ClientFactory) git.ClientFactory {
+	return &InRepoConfigGitCache{
+		ClientFactory: factory,
+		cache:         map[string]git.RepoClient{},
+	}
+}
+
+func (c *InRepoConfigGitCache) ClientFor(org, repo string) (git.RepoClient, error) {
```
listx, line 205: Nit: `GetClientFor` would read easier for me; but at the same time I notice that it is an existing naming convention (e.g., `c.ClientFactory.ClientFor()` on line 217). Nvm

prow/config/inrepoconfig.go
```diff
@@ -187,3 +188,46 @@ func DefaultAndValidateProwYAML(c *Config, p *ProwYAML, identifier string) error
...
+	c.cache[key] = client
+	return client, nil
+}
+
+type skipCleanRepoClient struct {
+	git.RepoClient
+}
+
+func (rc *skipCleanRepoClient) Clean() error {
+	// Skip cleaning to allow reuse.
```
listx, line 231: Nit: reword to

```
// Skip cleaning to allow reuse as a cached entry.
```

listx (commented):
Nits

prow/config/inrepoconfig.go
alvaroaleman, line 205: I am not really familiar with `sub`, but this will only work if it never modifies the clone. It is also not quite obvious how this is specific to `InRepoConfig`?

Lets please either move this into sub or have a big fat go doc warning that the clients must not be used to modify the clone and that the files could be modified while reading them (if someone else calls `ClientFor`)

cjwagner: > this will only work if it never modifies the clone

You mean editing files, right? I don't think checking out and merging various refs as `inrepoconfig` does should be a problem.

This may not be have to be specific to `InRepoConfig`, that is just the use case I have in mind and I chose to name it specifically because I am optimizing for that specific use case. If this ends up being generally useful I'm happy to choose a more generic name.

I'm realizing that what I have written doesn't provide sufficient synchronization though. I'm going to hold this and iterate. The next version may be more obviously coupled to `inrepo` config.
/hold

cjwagner: Yeah, it also has to have that exact name to match this interface: https://pkg.go.dev/k8s.io/test-infra/prow/git/v2#ClientFactory

alvaroaleman: Yeah I meant modifying files/leaving an unclean working directory.

Inrepoconfig doesn't do that but the resulting client is still not threadsafe, as the checkouts manipulate the same dir.

cjwagner: Yeah the threadsafety issue was what I mean by insufficient synchronization :slightly_smiling_face: 

I think I've resolved that issue and I've added a test to validate that the concurrent behavior is correct. PTAL.
/hold cancel

prow/config/inrepoconfig_test.go
```diff
@@ -631,3 +640,98 @@ func testDefaultProwYAMLGetter_RejectsNonGitHubRepo(clients localgit.Clients, t
...
+	<-signal1
+	// Now start the other two threads
+	go thread2()
+	go thread3()
+	// Ensure thread3 is not blocked by thread1 (repo2 client isn't blocked by concurrent repo1 client usage)
+	<-signal3
+	if cf.clientsCreated != 2 {
+		t.Errorf("Expected 2 clients to be created at this point, but got %d.", cf.clientsCreated)
+	}
+	// Try to confirm that thread2 is blocked by thread1 (disallow concurrent use of repo1 client)
```
alvaroaleman, line 721: We could avoid this by leveraging the race detector in what we purposely introduce a datarace in thread 1 and thread 2 (i.E. make them write to the same variable or sth like that).

This should make the racedector go off if the locking isn't working correctly and might simplify the testcase overall, as we only need to start the three threads, wait for them, verify two clients got created and the racedector didn't go off.

cjwagner: Sure, thats a good idea. I figured this kind of explicitly synchronized test would be more reliable/deterministic than depending on the race detector though. Do you know how reliable the race detector is?
Essentially I'm wondering if you think we should add what you are suggesting as a new test and keep the existing test, or replace the existing tests.

alvaroaleman: The racedetector fails whenever a data race is possible and doesn't require one to actually happen, so it is deterministic (will always fail/succeed for the same code). Interestlingly the golang docs do not explicitly say anything about that though:  https://golang.org/doc/articles/race_detector

It works by checking if there are unsynchronized variable accesses where at least one is a write. I couldn't quickly find a blogpost about this but I saw a talk some years ago. Something like https://www.youtube.com/watch?v=5erqWdlhQLA  likely has the details if you are interested.

cjwagner: Neat! Well I've updated the test to the simplified form you suggested, PTAL.

alvaroaleman (approved):

prow/config/inrepoconfig_test.go
```diff
@@ -631,3 +639,80 @@ func testDefaultProwYAMLGetter_RejectsNonGitHubRepo(clients localgit.Clients, t
...
+// - RepoClients are reused, only 1 client is created per repo.
+func TestInRepoConfigGitCacheConcurrency(t *testing.T) {
+	t.Parallel()
+
+	cf := &testClientFactory{}
+	cache := NewInRepoConfigGitCache(cf)
+	org, repo1, repo2 := "org", "repo1", "repo2"
+	// block channels are populated from the main thread, signal channels are read from the main thread.
+	signal := make(chan bool)
+	sharedRepo1State := 0
```
listx, line 674: Delete?

prow/config/inrepoconfig_test.go
```diff
@@ -631,3 +639,80 @@ func testDefaultProwYAMLGetter_RejectsNonGitHubRepo(clients localgit.Clients, t
...
+		client, err := cache.ClientFor(org, repo2)
+		if err != nil {
+			t.Errorf("Unexpected error getting repo client for thread 3: %v.", err)
+			return
+		}
+		client.Clean()
+		signal <- true
+	}()
+
+	for i := 0; i < 3; i++ {
```
listx, line 711: Question: I guess we're just testing that all 3 goroutines are able to finish? (And, if we don't get any of the signals that Golang will detect a deadlock?)

prow/config/inrepoconfig_test.go
alvaroaleman: Both the thread1 and thread2 routine write to it, which will make the race detector go off if `ClientFor` allows them to both have a client

prow/config/inrepoconfig_test.go
alvaroaleman: We are waiting, otherwise the test might finish when the routines still run

PR #23267, opened 2021-08-16
Add a default tenantID when none is provided

mpherman2:
Change config for ProwJobDefaults so if no TenantID is set, it is replaced with a global default.
Change Deck to look for this global default tenantID it no tenantID is set for Deck, so public instance of Deck will not need any changes. 

/assign @chaodaiG @cjwagner 
/hold

### DIFF
--- a/prow/config/config.go
+++ b/prow/config/config.go
@@ -70,6 +70,8 @@ const (
 	// it up if present. This allows components to include the config version
 	// in their logs, which can be useful for debugging.
 	ConfigVersionFileName = "VERSION"
+
+	DefaultTenantID = "GlobalDefaultID"
 )
 
 // Config is a read-only snapshot of the config.
@@ -621,6 +623,9 @@ func (pc *ProwConfig) mergeProwJobDefault(repo, cluster string, jobDefault *prow
 	merged = jobDefault.ApplyDefault(merged)
 	if merged == nil {
 		merged = &prowapi.ProwJobDefault{}
+	}
+	if merged.TenantID == "" {
+		merged.TenantID = DefaultTenantID
 	}
 	return merged
 }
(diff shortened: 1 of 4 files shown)


prow/config/config.go
```diff
@@ -622,6 +624,9 @@ func (pc *ProwConfig) mergeProwJobDefault(repo, cluster string, jobDefault *prow
 	if merged == nil {
 		merged = &prowapi.ProwJobDefault{}
 	}
+	if merged.TenantID == "" {
```
chaodaiG, line 627: I think this would work. but imo this should happen earlier, more specifically prior to calling `mergeProwJobDefault`. this is mimicking the behavior of inserting one entry on top of `ProwJobDefaultEntry` as:
```
- config:
    tenant_id: <DEFAULT_ID>
```

prow/config/config.go
```diff
@@ -70,6 +70,8 @@ const (
 	// it up if present. This allows components to include the config version
 	// in their logs, which can be useful for debugging.
 	ConfigVersionFileName = "VERSION"
+
+	DefaultTenantID = "Default"
```
chaodaiG, line 74: nit: can we naming this a bit longer so that there is no chance someone can accidentally come up the same name in their prow config?

chaodaiG (changes requested):

prow/config/config.go
mpherman2: If we make changes to the ProwJobDefault before calling `mergeProwJobDefault` then those changes will always override any other defaults. I am also not sure I understand the problem with the current implementation.

prow/config/config.go
mpherman2: Done

prow/config/config.go
chaodaiG: The idea I had in mind was adding 
```
- config:
    tenant_id: <DEFAULT_ID>
```
as the first entry in `pc.ProwJobDefaultEntries`, which won't override anything else, but has its own problem. I think your solution is actually better

chaodaiG (approved):

PR #23280, opened 2021-08-17
Filter Tide Pools, History and Queries based on TenantID

mpherman2:
Sorry for the long PR, most of the changes are unit tests.
Continuing work for Private repo multi-tenancy
/assign @chaodaiG @cjwagner 
/hold

### DIFF
--- a/prow/cmd/deck/tide.go
+++ b/prow/cmd/deck/tide.go
@@ -27,6 +27,7 @@ import (
 	"github.com/sirupsen/logrus"
 
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/test-infra/prow/config"
 	"k8s.io/test-infra/prow/tide"
 	"k8s.io/test-infra/prow/tide/history"
@@ -51,6 +52,9 @@ type tideAgent struct {
 	hiddenRepos func() []string
 	hiddenOnly  bool
 	showHidden  bool
+
+	tenantIDs []string
+	cfg       config.Config
 
 	sync.Mutex
 	pools   []tide.Pool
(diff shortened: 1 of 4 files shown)


prow/cmd/deck/tide.go
```diff
@@ -155,60 +159,103 @@ func (ta *tideAgent) updateHistory() error {
...
-		} else if needsHide == ta.hiddenOnly {
+		// curIDs are the IDs associated with all PJs in the Pool
+		// We want to add the ID associated with the OrgRepo for extra protection
+		curIDs := sets.String{}.Insert(pool.TenantIDs...)
+		orgRepoID := ta.cfg.GetProwJobDefault(pool.Org+"/"+pool.Repo, "*").TenantID
+		// If OrgRepo is associated with no TenantID, ignore it.
+		if orgRepoID != "" && orgRepoID != config.DefaultTenantID {
+			curIDs.Insert(orgRepoID)
+		}
+		if ta.matchingIDs(curIDs.List()) {
```
chaodaiG, line 177: would prefer `if len(ta.tenantIDs) > 0` instead of hiding this condition under `matchingIDs`

prow/cmd/deck/tide.go
```diff
@@ -155,60 +159,103 @@ func (ta *tideAgent) updateHistory() error {
...
 
-func (ta *tideAgent) filterHiddenPools(pools []tide.Pool) []tide.Pool {
-	if len(ta.hiddenRepos()) == 0 {
-		return pools
-	}
+func (ta *tideAgent) matchingIDs(ids []string) bool {
+	return len(ids) > 0 && len(ta.tenantIDs) > 0 && sets.String{}.Insert(ta.tenantIDs...).HasAll(ids...)
+}
 
+func (ta *tideAgent) filterHiddenPools(pools []tide.Pool) []tide.Pool {
```
chaodaiG, line 166: nit, the name of the function is not accurate any more, it also filters based on tenantIDs, not necessarily hidden

prow/cmd/deck/tide.go
```diff
@@ -155,60 +159,103 @@ func (ta *tideAgent) updateHistory() error {
...
-		if needsHide && ta.showHidden {
-			filtered[pool] = records
-		} else if needsHide == ta.hiddenOnly {
+		orgRepo := strings.Split(pool, ":")[0]
+		curIDs := recordIDs(records).Insert()
+		orgRepoID := ta.cfg.GetProwJobDefault(orgRepo, "*").TenantID
+		// If OrgRepo is associated with no TenantID, ignore it.
+		if orgRepoID != "" && orgRepoID != config.DefaultTenantID {
+			curIDs.Insert(orgRepoID)
+		}
```
chaodaiG, line 221: can this be a helper function?

prow/cmd/deck/tide.go
```diff
@@ -155,60 +159,103 @@ func (ta *tideAgent) updateHistory() error {
...
+		if ta.matchingIDs(curIDs.List()) {
```
chaodaiG, line 222: ditto `len(ta.tenantIDs)`

prow/cmd/deck/tide.go
```diff
@@ -155,60 +159,103 @@ func (ta *tideAgent) updateHistory() error {
...
+		curIDs := sets.String{}.Insert(pool.TenantIDs...)
+		orgRepoID := ta.cfg.GetProwJobDefault(pool.Org+"/"+pool.Repo, "*").TenantID
+		// If OrgRepo is associated with no TenantID, ignore it.
+		if orgRepoID != "" && orgRepoID != config.DefaultTenantID {
+			curIDs.Insert(orgRepoID)
+		}
+		if ta.matchingIDs(curIDs.List()) {
+			// Deck has tenantIDs and they match with the pool
 			filtered = append(filtered, pool)
+		} else if len(ta.tenantIDs) == 0 {
```
chaodaiG, line 180: can we save one layer of indentation by:
```
else if needsHide := matches(pool.Org+"/"+pool.Repo, ta.hiddenRepos()); needsHide {
```
contingent on the previous `if` condition uses `len(ta.tenantIDs)`

prow/cmd/deck/tide.go
```diff
@@ -155,60 +159,103 @@ func (ta *tideAgent) updateHistory() error {
...
-				includesHidden = true
-				break
-			}
-		}
-		if includesHidden && ta.showHidden {
-			filtered = append(filtered, qc)
-		} else if includesHidden == ta.hiddenOnly {
+		curIDs := qc.TenantIDs(ta.cfg)
+		needsHide := false
+		if ta.matchingIDs(curIDs) {
```
chaodaiG, line 242: ditto `if len(ta.tenantIDs)`

prow/cmd/deck/tide.go
mpherman2: We would still need to double check that the len(tenantIDs ==0) b/c the prev if statement would get skipped if len(tenantIDs) >0 but they did not match.

chaodaiG: that's true. I think it would work better by:
```
if len(ta.tenantIDs) > 0 {
        if  ta.matchingIDs(curIDs.List()) {
			// Deck has tenantIDs and they match with the pool
			filtered = append(filtered, pool)
        }
} else if needsHide := matches(pool.Org+"/"+pool.Repo, ta.hiddenRepos()); needsHide {
```

mpherman2: Couldn't exactly do this. Wasn't able to save an indenting butI think it should be easier to follow now.

prow/cmd/deck/tide.go
```diff
@@ -52,6 +53,9 @@ type tideAgent struct {
 	hiddenOnly  bool
 	showHidden  bool
 
+	tenantIDs []string
+	cfg       config.Config
```
cjwagner, line 57: This should be a `config.Getter` function rather than a `config.Config` struct so that we don't have to restart Deck in order for it to pick up config changes.

prow/cmd/deck/tide.go
```diff
@@ -147,68 +151,118 @@ func (ta *tideAgent) updateHistory() error {
...
 	ta.history = history
 	return nil
 }
 
-func (ta *tideAgent) filterHiddenPools(pools []tide.Pool) []tide.Pool {
-	if len(ta.hiddenRepos()) == 0 {
-		return pools
-	}
+func (ta *tideAgent) matchingIDs(ids []string) bool {
+	return len(ids) > 0 && sets.String{}.Insert(ta.tenantIDs...).HasAll(ids...)
```
cjwagner, line 163: nit: `sets.String{}.Insert(...)` can be simplified to `sets.NewString(...)`

It would be simpler and more efficient to just make `ta.tenantIDs` a `sets.String`.

prow/cmd/deck/tide.go
```diff
@@ -147,68 +151,118 @@ func (ta *tideAgent) updateHistory() error {
...
-		if needsHide && ta.showHidden {
-			filtered = append(filtered, pool)
-		} else if needsHide == ta.hiddenOnly {
+		// curIDs are the IDs associated with all PJs in the Pool
+		// We want to add the ID associated with the OrgRepo for extra protection
+		curIDs := sets.NewString(pool.TenantIDs...)
+		orgRepoID := ta.cfg.GetProwJobDefault(pool.Org+"/"+pool.Repo, "*").TenantID
+		// If the orgrepo is associated with no tenantID OR the default tenantID we ignore it here.
+		// This prevents already IDd pools from getting the default ID assigned to them when their orgrepo is not associated with an OrgRepo.
+		// Pools with no tenantID and with default tenantID behave the same, so adding the default ID just causes issues
```
cjwagner, line 175: > Pools with no tenantID and with default tenantID behave the same

I think this should be true, but it isn't currently. Currently the `matchingIDs` function always returns false if the list of ids is empty. If the list were to contain the default ID it could match and return true.

Also what issues are you referring to?

prow/cmd/deck/tide.go
```diff
@@ -147,68 +151,118 @@ func (ta *tideAgent) updateHistory() error {
...
+		if len(ta.tenantIDs) > 0 {
+			if ta.matchingIDs(curIDs.List()) {
+				// Deck has tenantIDs and they match with the History
+				filtered[pool] = records
+			}
+		} else if needsHide := matches(orgRepo, ta.hiddenRepos()); needsHide {
+			if ta.showHidden || ta.hiddenOnly {
+				filtered[pool] = records
+			}
+		} else if !ta.hiddenOnly && noTenantIDOrDefaultTenantID(curIDs.List()) {
```
cjwagner, line 237: Please refactor some of this common code into another shared function. It would be nice to have one function for the decision of whether we should keep an item based on the deck configuration and the hidden status + tenant IDs of the item.

mpherman2: Done. Tried thinking about how to do this before but my brain was fried and couldn't figure it out :) Looks much nicer now

PR #23287, opened 2021-08-18
config/prow: setup milestones for kubernetes-sigs/boskos

spiffxp:
Enable `milestoneapplier` for kubernetes-sigs so that milestones are auto-applied for repos that have it configured

Enable `milestone` for kubernetes-sigs/boskos so contributors can /milestone there

I'm opting not to configure a specific team for `milestone` mainly out of curiosity, I forget if this plugin supports cross-org teams or not, so I'd like to see if it defaults to `@kubernetes/milestone-maintainers`

### DIFF
--- a/config/prow/plugins.yaml
+++ b/config/prow/plugins.yaml
@@ -363,6 +363,8 @@ milestone_applier:
     main: v0.4
     release-0.4: v0.4
     release-0.3: v0.3
+  kubernetes-sigs/boskos:
+    master: v1.23
   kubernetes-sigs/controller-runtime:
     master: v0.10.x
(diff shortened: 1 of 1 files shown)


alvaroaleman:
This is not a strong opinion by any means, but does it really make sense to use Kubernetes releases as milestone in boskos?

spiffxp:
It lets us consistently keep track of what we've committed to during the release lifecycle across all sub projects. Which makes less sense for projects that are tagged and versioned independently like kind. But more sense for those that are not, and that may induce disruption during the release cycle. Like say if boskos stopped working mid v1.23

PR #23312, opened 2021-08-19
prow/github: expose more options for the client

stevekuznetsov:
Second commit fixes the issue we had last time, I think.

/assign @alvaroaleman


petr-muller (approved):
LGTM but maybe someone who actually looked into the previous regression should look
/hold

chaodaiG:
@stevekuznetsov , would you mind share a bit more why the previous attempt failed and how it's improved here?

PR #23314, opened 2021-08-19
Add presubmit job for scheduler-plugin release-1.20

Huang-Wei:
- add presubmit job for scheduler-plugin release-1.20
- bump go version to 1.16.7

/sig scheduling


Huang-Wei:
@damemi could you /lgtm this?

/cc @seanmalloy

damemi (approved):
/lgtm

PR #23323, opened 2021-08-20
kettle make_json: skip and log rows bigger than 100MB

### LINKED ISSUE #23135: kettle appears to be down
**What happened**:

- go.k8s.io/triage results only extend to july 28th
- triage job is passing https://testgrid.k8s.io/sig-testing-misc#triage&graph-metrics=test-duration-minutes&width=5
- kettle-metrics job is failing https://testgrid.k8s.io/sig-testing-misc#metrics-kettle&width=5
- kettle-staging deployment is crashloopbackoff due to out of disk
- kettle deployment still appears to be up

kettle should be up

....

see above

probably we should have alerting for the kettle-metrics job

matthyx:
fixes #23135 
/assign @BenTheElder

### DIFF
--- a/kettle/make_json.py
+++ b/kettle/make_json.py
@@ -32,6 +32,7 @@ except ImportError:
 
 import model
 
+MAX_ROW_SIZE = 104857600 # 100MB
 SECONDS_PER_DAY = 86400
 
 def buckets_yaml():
@@ -58,6 +59,16 @@ else:
     # This is safe because the only way we get here is by faling all attempts
     raise
 
+# https://stackoverflow.com/questions/39047624/is-there-an-easy-way-to-estimate-size-of-a-json-object
+class MeterFile:
+    """
+    File-like object, throws away everything you write to it but keeps track of the size.
+    """
+    def __init__(self, size=0):
+        self.size = size
+
+    def write(self, string):
+        self.size += len(string)
 
 class Build:
     """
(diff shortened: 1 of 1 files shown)


kettle/make_json.py
```diff
@@ -312,6 +322,11 @@ def main(db, opts, outfile):
 
     rows_emitted = set()
     for rowid, row in make_rows(db, builds):
+        meterfile = MeterFile()
```
BenTheElder, line 325: this is clever, but does it make more sense to wrap this up in a method that gets the json size and keep the class as an implementation detail?

kettle/make_json.py
```diff
@@ -312,6 +322,11 @@ def main(db, opts, outfile):
...
+        json.dump(row, meterfile, sort_keys=True)
+        if meterfile.size > 104857600:
```
BenTheElder, line 327: we should explain this constant.

kettle/make_json.py
```diff
@@ -312,6 +322,11 @@ def main(db, opts, outfile):
...
+            print('skipping row of size %d for path %s' % (meterfile.size, row['path']))
```
BenTheElder, line 328: we should mention that this is because it exceeds the limit

BenTheElder (commented):
thank you!

cc @spiffxp

matthyx:
@BenTheElder done

BenTheElder (approved):
/lgtm
/approve

I think we should try to get kettle unblocked with this, we can revisit later if we should say attempt to truncate the log instead.

PR #23324, opened 2021-08-20
remove CSIServiceAccountToken which will be removed in 1.23

pacoxu:
https://github.com/kubernetes/kubernetes/pull/104255 CI failed for removing the feature gate


pacoxu:
https://github.com/kubernetes/kubernetes/blob/7c83b5278d4a20fc1fa25cbf4168e2d1beec4ebc/pkg/features/kube_features.go#L585-L592

@zshihang
Would you confirm that we can remove this feature gate?

PR #23326, opened 2021-08-20
Fix ci-build-and-push-k8s-at-golang-tip-canary job

tosi3k:
The reason for consistent failures of the canary job was that it leveraged `kubekins-e2e` image in the config instead of `bootstrap`. `kubekins-e2e` has Go installation that conflicts with the [scripts](https://github.com/kubernetes/perf-tests/tree/master/golang) used in the job that install a custom Go version instead. This is why `bootstrap` was picked as the job's container image in the first place as it doesn't install Go at all.

I tested this change on my local Prow instance and it works correctly.

Fixes https://github.com/kubernetes/perf-tests/issues/1861.

/cc @ameukam 
/assign @spiffxp


spiffxp (approved):
/approve
/lgtm

PR #23327, opened 2021-08-20
test run_if_changed/skip_if_only_changed for postsubmits

listx:
These tests are lifted almost verbatim from their Test...Presubmit
counterparts. They are added here for completeness. We may want to
refactor them in the future because both Presubmits and Postsubmits rely
on the same embedded RexepChangeMatcher type. The tests added here
merely ensure that we continue to use the same RegexpChangeMatcher type
in both job types.

/cc @chaodaiG @cjwagner


listx:
Breaking this out of #23242 to keep that PR smaller.

PR #23335, opened 2021-08-23
Allow /approve to be misspelled as /approved

danwinship:
Tim did this [here](https://github.com/kubernetes/kubernetes/pull/95768#pullrequestreview-638066179) a while back and then Clayton just did it twice [here](https://github.com/openshift/origin/pull/26369#issuecomment-899554796) and [here](https://github.com/openshift/origin/pull/26351#issuecomment-899554605), and those are just ones I know about... (It doesn't seem to be possible to search PRs for the string "`/approved`"; github just ignores the slash.)

Anyway, a modest proposal.
(This intentionally does not change any of the documentation, though maybe it ought to be mentioned *somewhere*?)

### DIFF
--- a/prow/plugins/approve/approve.go
+++ b/prow/plugins/approve/approve.go
@@ -40,11 +40,12 @@ const (
 	// PluginName defines this plugin's registered name.
 	PluginName = "approve"
 
-	approveCommand       = "APPROVE"
-	cancelArgument       = "cancel"
-	lgtmCommand          = "LGTM"
-	noIssueArgument      = "no-issue"
-	removeApproveCommand = "REMOVE-APPROVE"
+	approveCommand          = "APPROVE"
+	alternateApproveCommand = "APPROVED"
+	cancelArgument          = "cancel"
+	lgtmCommand             = "LGTM"
+	noIssueArgument         = "no-issue"
+	removeApproveCommand    = "REMOVE-APPROVE"
 )
 
 var (
@@ -530,7 +531,7 @@ func isApprovalCommand(isBot func(string) bool, lgtmActsAsApprove bool, c *comme
 
 	for _, match := range commandRegex.FindAllStringSubmatch(c.Body, -1) {
 		cmd := strings.ToUpper(match[1])
-		if (cmd == lgtmCommand && lgtmActsAsApprove) || cmd == approveCommand || cmd == removeApproveCommand {
+		if (cmd == lgtmCommand && lgtmActsAsApprove) || cmd == approveCommand || cmd == alternateApproveCommand || cmd == removeApproveCommand {
 			return true
 		}
 	}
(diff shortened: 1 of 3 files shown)


ardaguclu:
@danwinship this one is another one https://github.com/openshift/origin/pull/26344

alvaroaleman (approved):
/hold
I like it, @cblecker @chaodaiG @stevekuznetsov WDYT?

prow/plugins/approve/approve.go
```diff
@@ -41,6 +41,7 @@ const (
 	PluginName = "approve"
 
 	approveCommand       = "APPROVE"
+	wrongApproveCommand  = "APPROVED"
```
chaodaiG, line 44: nit: `wrongApproveCommand` implies error to me, how about change it to `misspelledApproveCommand`? 

alternatively, it might also take a slice like `alternativeApproveCommands`, or just combine with `APPROVE` to be underl `approveCommands`

chaodaiG (commented):
LGTM as well, left a nit

There are two different places this could be mentioned:
https://github.com/kubernetes/test-infra/blob/ede436ecdabc933997307e1cca9b4d7b04fa3204/prow/plugins/approve/approve.go#L132
OR
https://github.com/kubernetes/test-infra/blob/master/prow/ANNOUNCEMENTS.md

prow/plugins/approve/approve.go
danwinship: ok, I went with `alternateApproveCommand`. Doing a slice would have required a bunch more rewriting...

chaodaiG:
LGTM, please fix GOFMT error and I'll approve

cblecker:
From a ContribEx standpoint, we have always shied away from aliases for core commands (especially undocumented ones). The reason for this, is it makes it difficult for new contributors to know and understand what the flow looks like and what they should be doing.

I know there's a desire to be smart here and try and "do the right thing" based on intent, but these have consistently shown to be problematic in the larger picture.

/hold

karuppiah7890:
> From a ContribEx standpoint, we have always shied away from aliases for core commands (especially undocumented ones). The reason for this, is it makes it difficult for new contributors to know and understand what the flow looks like and what they should be doing.
>
> I know there's a desire to be smart here and try and "do the right thing" based on intent, but these have consistently shown to be problematic in the larger picture.

Do we document such issues around command aliases / decisions about moving away from aliases? With data (examples etc) and reasoning.

karuppiah7890:
Maybe if aliases won't be supported, we can get more data around this later and if too many people use `/approved`, then we could maybe think about letting the bot comment and say something like `did you mean approve?` but that might be too much to implement and unnecessary too. Just an idea based on the comment here - https://github.com/kubernetes/kubernetes/pull/95768#issuecomment-822723560 and also based on experience in CLI commands which usually ask "Did you mean xyz?" when something unknown is typed

PR #23341, opened 2021-08-23
blunderbuss: support reviewerCount==0

dhellmann:
When the requested number of reviewers is explicitly 0, return no
reviewers.

This allows the plugin to effectively be disabled in one repository
when it is enabled for an entire organization.

Add unit test cases for zero reviewerCount and update existing unit
tests that relied in implicit behavior to set the reviewerCount
explicitly to a non-zero value.


petr-muller (approved):

alvaroaleman (approved):
Ideally we wouldn't need to add something like this into individual plugins, but the `excluded_repo` setting only allows to exclude a repo from all plugins defined on org level instead of per plugin, bummer.

PR #23342, opened 2021-08-23
Override branch protected required status checks

### LINKED ISSUE #23340: Override branch protected required status checks
Allow using the `/override` command to required status checks.

**Why is this needed**:
Currently, the override plugin looks only on ongoing commit statuses.
In case there is a required status check configured that doesn't appear under the commit statuses then it's impossible to override it.

e.g.
In the following example, Jenkins is configured as a required status check.
But Jenkins didn't return any result to the commit so it doesn't appear under statuses.
Screenshot from 2021-08-23 13-59-48

```bash
➜  test-infra git:(master) ✗ gh api "repos/openshift/assisted-service/statuses/8f2dc726627f0d2df0d32f4bdc27ebd33667f011" | jq '.[].context' | sort -u
"ci.ext.devshift.net PR build"
"ci/prow/ci-index"
"ci/prow/e2e-metal-assisted"
"ci/prow/e2e-metal-assisted-kube-api"
"ci/prow/e2e-metal-assisted-operator-ztp"
"ci/prow/images"
"ci/prow/lint"
"ci/prow/subsystem-aws"
"ci/prow/subsystem-kubeapi-aws"
"tide"
```
vs.
```bash
➜  test-infra git:(master) ✗ gh api "repos/openshift/assisted-service/branches/master" | jq '.protection.required_status_checks.contexts'
[
  "ci.ext.devshift.net PR build",
  "continuous-integration/jenkins/pr-merge",
  "ci/prow/lint",
  "ci/prow/ci-index",
  "ci/prow/images"
]
```

YuviGold:
Resolves #23340

Currently, the override plugin looks only on ongoing commit statuses.

In case there is a required status check configured that doesn't appear
under the commit statuses then it's impossible to override it.

### DIFF
--- a/prow/plugins/override/override.go
+++ b/prow/plugins/override/override.go
@@ -51,6 +51,7 @@ type githubClient interface {
 	GetRef(org, repo, ref string) (string, error)
 	HasPermission(org, repo, user string, role ...string) (bool, error)
 	ListStatuses(org, repo, ref string) ([]github.Status, error)
+	GetBranchProtection(org, repo, branch string) (*github.BranchProtection, error)
 	ListTeams(org string) ([]github.Team, error)
 	ListTeamMembers(org string, id int, role string) ([]github.TeamMember, error)
 }
@@ -94,6 +95,9 @@ func (c client) GetPullRequest(org, repo string, number int) (*github.PullReques
 }
 func (c client) ListStatuses(org, repo, ref string) ([]github.Status, error) {
 	return c.ghc.ListStatuses(org, repo, ref)
+}
+func (c client) GetBranchProtection(org, repo, branch string) (*github.BranchProtection, error) {
+	return c.ghc.GetBranchProtection(org, repo, branch)
 }
 func (c client) HasPermission(org, repo, user string, role ...string) (bool, error) {
 	return c.ghc.HasPermission(org, repo, user, role...)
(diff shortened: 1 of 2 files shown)


petr-muller:
Personally I don't think the plugin should have this functionality. There's IMO no reason why Prow of all things should be interacting with other CI systems' statuses. On the UX level this creates an inconsistency where some commands work with non-Prow statuses and some do not (and on top of it, using a fairly arbitrary criteria of "non-Prow statuses which happen to be configured in branch protection").

YuviGold:
> Personally I don't think the plugin should have this functionality. There's IMO no reason why Prow of all things should be interacting with other CI systems' statuses. On the UX level this creates an inconsistency where some commands work with non-Prow statuses and some do not (and on top of it, using a fairly arbitrary criteria of "non-Prow statuses which happen to be configured in branch protection").

@petr-muller But Prow already allows that.

1. You can override jobs that are not prowjobs. They just need to provide a status before. (it can still be a failure that gets overriden)
2. You can configure required status checks using branch protection on contexts that are not prowjobs using Prow.


So there's already this mix of Prow managing every merge configuration aspect.
This new functionality just makes this partial solution complete by giving full capabilities for GitHub.

fejta:
Perhaps we should create two different plugins or modes for /override

1)  By default it will only override prow-job contexts
2) Otherwise a flag can make it override any context.

Currently the behavior of any prow job as well as anything specified in the branch protection config seems a little unintuitive.

YuviGold:
> Perhaps we should create two different plugins or modes for /override
> 
> 1. By default it will only override prow-job contexts
> 2. Otherwise a flag can make it override any context.
> 
> Currently the behavior of any prow job as well as anything specified in the branch protection config seems a little unintuitive.

@fejta @petr-muller 
The linked issue examines the GH API. Its outputs are pure context regardless of prow jobs or not.
The override plugin **already** affects non-prow jobs context as well. It's about a GitHub wrapper integration that provides existing GitHub statuses.

The suggested behavior would overcome an edge case of a required context that its status wasn't provided for any reason, but it is still expected by GitHub (as defined in branch protection) and then it is blocked.
In case of a required Prow job not responding to GitHub with any status (e.g. a bug in the reporter component), this scenario would fit as well.

Yes, it would work for Prow jobs as well as non-prow jobs, but I think the question of the override plugin intention isn't related to this PR...
Changing the way the plugin **currently** works (which works for non-prow jobs) would mean a much bigger design change.

alvaroaleman (approved):
This seems reasonable to me
/hold

YuviGold:
@petr-muller can we unhold and merge this PR?

PR #23351, opened 2021-08-24
Implement git fetch

chaodaiG:
https://github.com/kubernetes/test-infra/pull/23264 uses this function

### DIFF
--- a/prow/git/git.go
+++ b/prow/git/git.go
@@ -507,11 +507,21 @@ func (r *Repo) MergeCommitsExistBetween(target, head string) (bool, error) {
 }
 
 // ShowRef returns the commit for a commitlike. Unlike rev-parse it does not require a checkout.
-func (i *Repo) ShowRef(commitlike string) (string, error) {
-	i.logger.Infof("Getting the commit sha for commitlike %s", commitlike)
-	out, err := i.gitCommand("show-ref", "-s", commitlike).CombinedOutput()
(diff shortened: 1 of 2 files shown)


prow/git/git.go
```diff
@@ -507,11 +507,21 @@ func (r *Repo) MergeCommitsExistBetween(target, head string) (bool, error) {
 }
 
 // ShowRef returns the commit for a commitlike. Unlike rev-parse it does not require a checkout.
-func (i *Repo) ShowRef(commitlike string) (string, error) {
-	i.logger.Infof("Getting the commit sha for commitlike %s", commitlike)
-	out, err := i.gitCommand("show-ref", "-s", commitlike).CombinedOutput()
+func (r *Repo) ShowRef(commitlike string) (string, error) {
+	r.logger.Infof("Getting the commit sha for commitlike %s", commitlike)
```
fejta, line 511: consider using a field for commitlike

fejta (approved):
/hold

fejta (approved):

cjwagner (approved):
Thanks, I'm surprised this is the first time we're using `fetch`.
/hold

prow/git/git.go
chaodaiG: this PR only renamed `i` to `r` for consistency reason, surprisingly github displays diff like this.

I can fix all the other logger in a separate PR

PR #23363, opened 2021-08-25
Update controller runtime master branch tests to run on Go 1.17

pigletfly:
/assign @alvaroaleman

since k/k has been updated to go1.17


alvaroaleman:
I don't think we should do this before there is a tag in the k8s.io/* dependencies that picked up the go 1.17 change and which we are using in controller-runtime.

PR #23379, opened 2021-08-26
Include the decorated volumes by the user in the config validation

droslean:
When a user specifies a `decoration_config.oauth_token_secret` of `decoration_config.ssh_key_secrets`, we eventually create the corresponding volumes plus the volumeMounts in the `clonerefs` container. 

With this change, we allow the user to reuse the volumes specified in the decoration config by mounting them into different containers.

### DIFF
--- a/prow/config/config.go
+++ b/prow/config/config.go
@@ -1732,7 +1732,7 @@ func validateJobBase(v JobBase, jobType prowapi.ProwJobType, podNamespace string
 	if err := validateAgent(v, podNamespace); err != nil {
 		return err
 	}
-	if err := validatePodSpec(jobType, v.Spec, v.DecorationConfig != nil); err != nil {
+	if err := validatePodSpec(jobType, v.Spec, v.DecorationConfig); err != nil {
 		return err
 	}
 	if err := ValidatePipelineRunSpec(jobType, v.ExtraRefs, v.PipelineRunSpec); err != nil {
@@ -2345,7 +2345,7 @@ func ValidatePipelineRunSpec(jobType prowapi.ProwJobType, extraRefs []prowapi.Re
 	return nil
 }
 
-func validatePodSpec(jobType prowapi.ProwJobType, spec *v1.PodSpec, decorationEnabled bool) error {
+func validatePodSpec(jobType prowapi.ProwJobType, spec *v1.PodSpec, decorationConfig *prowapi.DecorationConfig) error {
 	if spec == nil {
 		return nil
 	}
@@ -2361,7 +2361,7 @@ func validatePodSpec(jobType prowapi.ProwJobType, spec *v1.PodSpec, decorationEn
 		return utilerrors.NewAggregate(append(errs, fmt.Errorf("pod spec must specify at least 1 container, found: %d", n)))
 	}
 
-	if n := len(spec.Containers); n > 1 && !decorationEnabled {
+	if n := len(spec.Containers); n > 1 && decorationConfig == nil {
 		return utilerrors.NewAggregate(append(errs, fmt.Errorf("pod utility decoration must be enabled to use multiple containers: %d", n)))
 	}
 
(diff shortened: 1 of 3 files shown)


prow/apis/prowjobs/v1/types.go
```diff
@@ -646,6 +647,30 @@ func (d *DecorationConfig) Validate() error {
...
+	toolsMountName          = "tools"
+	gcsCredentialsMountName = "gcs-credentials"
+	s3CredentialsMountName  = "s3-credentials"
+)
+
+// Volumes returns a set of secret names that will eventually be created in the pod as volumes.
+func (d *DecorationConfig) Volumes() sets.String {
+	ret := sets.NewString(logMountName, codeMountName, toolsMountName, gcsCredentialsMountName, s3CredentialsMountName)
+	if d == nil {
+		return ret
```
alvaroaleman, line 662: This is wrong, if the decoration config is nil it has to return nil.

prow/apis/prowjobs/v1/types.go
```diff
@@ -646,6 +647,30 @@ func (d *DecorationConfig) Validate() error {
 	return nil
 }
 
+const (
+	logMountName            = "logs"
+	codeMountName           = "code"
+	toolsMountName          = "tools"
+	gcsCredentialsMountName = "gcs-credentials"
+	s3CredentialsMountName  = "s3-credentials"
```
alvaroaleman, line 655: Why do we redefine them instead of using `prow/pod-utils/decorate.VolumeMounts()`?

prow/apis/prowjobs/v1/types.go
droslean: updated

prow/apis/prowjobs/v1/types.go
droslean: updated

prow/config/config.go
```diff
@@ -2401,20 +2401,21 @@ func validatePodSpec(jobType prowapi.ProwJobType, spec *v1.PodSpec, decorationEn
 	}
 
 	volumeNames := sets.String{}
+	decoratedVolumeNames := decorationConfig.Volumes()
```
alvaroaleman, line 2404: I don't understand this, why do we not call decorationconfig.Volumes in `func (d *DecorationConfig) Volumes()` instead of forcing the caller to know that both need to be  included?

droslean: Because the user specifies the name of the volume. Example of
```yaml
    decoration_config:
      oauth_token_secret:
        key: oauth
        name: my-secret
```
we need to allow the user to be able to add in `volumeMounts` the volume named `my-secret` without adding it as a volume since that will happen in runtime.

alvaroaleman: Yeah I understand that, but right now the callsite has to check if decoration is enabled, if so call `decorationConfig.Volumes()`, then call `decorationConfig.VolumeMounts()` (the first being the package, the second the variable) and then assemble them together. Why not call the latter inside the former?

droslean: Because they are two different things. Do you mean that its better to just refactor [decorate.VolumeMounts()](https://github.com/kubernetes/test-infra/blob/master/prow/pod-utils/decorate/podspec.go#L70-L73) to get a decoration config and extract the volumes from there?

droslean: A union of those sets should be logical here.

alvaroaleman: the purpose of `decorationConfig.Volumes()` is to return all volume names that will be used by this decoration config or not? So including `decorationconfig.VolumeMounts()` in there seems to make sense.

It maybe be very logical for you working on this code right now, but it requires the caller to know a bunch of implementation details to get this right and that shouldn't be required and is likely to be misused.

droslean: The `decorate.VolumeMounts()` is not related with any decorationConfig configuration. So even now the user can misuse it. Do you think that its better to drop a different message for the user since he will be responsible for the input?

stevekuznetsov (approved):

PR #23388, opened 2021-08-27
Spyglass example match Gubernator docs

coryrc:
In [Gubernator's docs](https://github.com/kubernetes/test-infra/tree/master/gubernator#job-artifact-gcs-layout) it says:

> Any JUnit XML reports should be named junit_*.xml and placed under ./artifacts as well.

I highly doubt it was intended to read that as a regex and not a
glob (junit_____.xml only is not very useful compared to
junit_words-and-stuff.xml) so I believe the default Spyglass example
should match the rest of the system. Alternatively, I could update
Gubernator's page.

I also believe we endorse `junit_*.xml` being found in any subdirectory,
which neither of the statements support. I could update both to match
this if desired.

### DIFF
--- a/prow/spyglass/README.md
+++ b/prow/spyglass/README.md
@@ -100,7 +100,7 @@ deck:
     - lens:
         name: junit
       required_files:
-      - ^artifacts/junit.*\.xml$
+      - ^artifacts/junit_[^/]*\.xml$
     - lens:
         name: podinfo
       required_files:


matthyx:
Looks reasonable.
/lgtm

prow/spyglass/README.md
```diff
@@ -100,7 +100,7 @@ deck:
     - lens:
         name: junit
       required_files:
-      - ^artifacts/junit.*\.xml$
+      - ^artifacts/junit_[^/]*\.xml$
```
spiffxp, line 103: This would exclude a simple `junit.xml` file, no? Not sure I like that.

I'm not clear what the old regex picks up that it should not?

coryrc: That would be in accordance with Gubernator's docs to exclude `artifacts/junit.xml`. The page explicitly has an underscore in the filename.

The old regex matches `artifacts/junit/blah.xml` but TestGrid (and presumably Gubernator but untested) does not.

As I wrote in the beginning, Gubernator's page can also be changed. I just believe they should be consistent.

spiffxp: Gubernator is deprecated so I'm fine breaking with its convention. These days we treat the layout described by testgrid as authoritative. So yeah we could change gubernator's docs I guess? I don't know if the code needs changes

PR #23392, opened 2021-08-27
InRepoConfigGitCache: cleans up after merge conflict

chaodaiG:


fejta:
the `git add -A` is important, otherwise you won't delete new files.

chaodaiG:
> the `git add -A` is important, otherwise you won't delete new files.
> 
> Aborting non-normal modes also seems important. Did you test out these scenarios? AKA a merge conflict.

yeah I had tested locally, that `git reset --hard HEAD` cleans up after merge conflict

prow/git/git.go
```diff
@@ -263,6 +263,31 @@ func (r *Repo) Clean() error {
 	return os.RemoveAll(r.dir)
 }
 
+// ResetHard runs `git reset --hard`
+func (r *Repo) ResetHard(commitlike string) error {
+	// `git reset --hard` doesn't cleanup untracked file, adding them to be safe
```
petr-muller, line 268: You could also `git clean -fd` for more straightforward cleanup (just a suggestion)

chaodaiG: feels like better to understand than `add -A`, updated the PR to use `clean -df`

petr-muller (approved):

PR #23397, opened 2021-08-28
Fix command for controller-tools presubmit

alvaroaleman:
Right now it uses `make test-all` as the commant to exec:
```
could not start the process: exec: "make test-all": executable file not found in $PATH
```

Ref
https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_controller-tools/603/pull-controller-tools-test-master/1431416733045886976

/assign @vincepri

### DIFF
--- a/config/jobs/kubernetes-sigs/controller-tools/controller-tools-presubmits-master.yaml
+++ b/config/jobs/kubernetes-sigs/controller-tools/controller-tools-presubmits-master.yaml
@@ -10,7 +10,8 @@ presubmits:
       containers:
       - image: golang:1.16
         command:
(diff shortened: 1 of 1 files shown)


chases2 (approved):
I've seen this in some specs as 

```
command:
- make
args:
- test-all
```

although it's not clear to me which is preferable. Seems like they do the same thing.

/hold for assignee

alvaroaleman:
> although it's not clear to me which is preferable. Seems like they do the same thing.

Yeah, they end up doing the same so it comes down to stylistic preference

/hold cancel

PR #23406, opened 2021-08-30
prow: expose retest status in metrics

stevekuznetsov:
It is very useful to be able to measure the volume and frequency of
re-tests independently from the volume of total jobs. Adding a label
at ProwJob creation time makes this simple.

/assign @alvaroaleman 
/cc @deads2k

### DIFF
--- a/prow/plugins/trigger/generic-comment.go
+++ b/prow/plugins/trigger/generic-comment.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 
 	"github.com/sirupsen/logrus"
+	"k8s.io/test-infra/prow/kube"
 
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/test-infra/prow/config"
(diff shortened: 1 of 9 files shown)


prow/plugins/trigger/generic-comment.go
```diff
@@ -132,6 +133,12 @@ func handleGenericComment(c Client, trigger plugins.Trigger, gc github.GenericCo
 	if needsHelp, note := pjutil.ShouldRespondWithHelp(gc.Body, len(toTest)); needsHelp {
 		return addHelpComment(c.GitHubClient, gc.Body, org, repo, pr.Base.Ref, pr.Number, presubmits, gc.HTMLURL, commentAuthor, note, c.Logger)
 	}
+	// we want to be able to track re-tests separately from the general body of tests
+	if pjutil.RetestRe.MatchString(gc.Body) || pjutil.RetestRequiredRe.MatchString(gc.Body) {
+		for _, item := range toTest {
```
alvaroaleman, line 138: This has three issues:
* You are manipulating the iterator variable, which will not get propagated outside the loop
* Once you fix that, you are going to directly manipulate the loaded config instead of doing a change scoped to this function
* Labels might be nil in which case this will panic

alvaroaleman (approved):

PR #23422, opened 2021-08-31
Move autobump back 15 min so it does not interfere with auto-deploy autobump job

mpherman2:
This should hopefully fix occasional flake caused by this job and regular autobump job happening at the same time.
@chaodaiG  just double checking it is OK if this job happens 15 min later? 
/hold
/assign @chaodaiG

### DIFF
--- a/config/jobs/kubernetes/test-infra/test-infra-trusted.yaml
+++ b/config/jobs/kubernetes/test-infra/test-infra-trusted.yaml
@@ -874,7 +874,7 @@ periodics:
     testgrid-alert-email: k8s-infra-oncall@google.com
     testgrid-num-failures-to-alert: '2' # This could fail when it runs right in the middle of prow push, tolerate it once
     description: runs autobumper to create/update a PR that bumps prow to the latest RC with label 'skip-review'
-- cron: "30 * * * 1-5"  # Bump don't label `skip-review`. Run at :30 past every hour Mon-Fri
+- cron: "15 * * * 1-5"  # Bump don't label `skip-review`. Run at :30 past every hour Mon-Fri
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes/test-infra/test-infra-trusted.yaml
```diff
@@ -837,7 +837,7 @@ periodics:
     testgrid-dashboards: sig-testing-maintenance
     testgrid-tab-name: autoupdate-patch
     description: Weekly module update to latest patched versions
-- cron: "30 17-22/5 * * 1-5"  # Bump with label `skip-review`. Run at 10:30 and 15:30 PST (17:05 UTC) Mon-Fri
+- cron: "45 17-22/5 * * 1-5"  # Bump with label `skip-review`. Run at 10:45 and 15:45 PST (17:05 UTC) Mon-Fri
```
chases2, line 840: nit: `(17:45 and 22:45 UTC)`
Or just `(in UTC)`, the actual numbers are in the string itself

chases2 (approved):

config/jobs/kubernetes/test-infra/test-infra-trusted.yaml
chaodaiG: This job determines when prow is deployed, so the timing of this job is more important than the one without label, I would prefer moving the other one back

mpherman2: That makes sense. Changed it so we do the regular autobump job at 15 past the hour instead of 30

chaodaiG (approved):

PR #23427, opened 2021-08-31
Labels: Allow to set up restricted labels not everyone can set

alvaroaleman:
Currently, everyone who is trusted can add/remove labels through the
labels plugin. Sometimes only a restricted set of ppl should be able to
do that. This change extend the labels plugin to allow setting up
`restricted_labels` for this purpose.

/assign @stevekuznetsov

### DIFF
--- a/prow/github/client.go
+++ b/prow/github/client.go
@@ -201,6 +201,7 @@ type TeamClient interface {
 	RemoveTeamRepo(id int, org, repo string) error
 	ListTeamInvitations(org string, id int) ([]OrgInvitation, error)
 	TeamHasMember(org string, teamID int, memberLogin string) (bool, error)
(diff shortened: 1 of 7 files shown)


stevekuznetsov (commented):
The request is not worded well but I believe we need to support individual usernames as well as team names. Otherwise this code looks good.

alvaroaleman:
> The request is not worded well but I believe we need to support individual usernames as well as team names. Otherwise this code looks good.

I saw that, I just don't think it is a good idea as it encourages SPOF users too much whereas if you really want, you can also create a single-user group. Added it for now but yeah, I don't think giving individual users permissions is a ever a good idea.

stevekuznetsov (approved):

PR #23434, opened 2021-09-01
 Add cloud-provider-openstack cinder csi sanity tests

ramineni:
This PR adds csi sanity tests for cinder-csi-plugin
Issue ref: https://github.com/kubernetes/cloud-provider-openstack/issues/1613

### DIFF
--- a/config/jobs/kubernetes/cloud-provider-openstack/release-master-presubmits.yaml
+++ b/config/jobs/kubernetes/cloud-provider-openstack/release-master-presubmits.yaml
@@ -68,3 +68,22 @@ presubmits:
     annotations:
       testgrid-dashboards: provider-openstack-openstack-csi-cinder
       testgrid-tab-name: presubmit-e2e-test
+
+  - name: openstack-cloud-csi-cinder-sanity-test
+    always_run: false
+    run_if_changed: '^(pkg\/csi\/cinder\/|cmd\/cinder-csi-plugin\/|tests\/sanity\/cinder\/)'
+    decorate: true
+    branches:
+      - ^master$
+    path_alias: "k8s.io/cloud-provider-openstack"
+    spec:
+      containers:
+      - image: gcr.io/k8s-testimages/kubekins-e2e:v20210825-f1955d1-master
+        command:
+        - make
+        args:
+        - test-cinder-csi-sanity
+    annotations:
+      testgrid-dashboards: provider-openstack-openstack-csi-cinder
+      testgrid-tab-name: presubmit-sanity-test
+      description: Run cloud-provider-openstack csi sanity tests for cinder-csi-plugin


ramineni:
Issue ref: https://github.com/kubernetes/cloud-provider-openstack/issues/1613

config/jobs/kubernetes/cloud-provider-openstack/release-master-presubmits.yaml
```diff
@@ -68,3 +68,20 @@ presubmits:
     annotations:
       testgrid-dashboards: provider-openstack-openstack-csi-cinder
       testgrid-tab-name: presubmit-e2e-test
+
+  - name: openstack-cloud-csi-cinder-sanity-test
+    always_run: false
+    run_if_changed: '^(pkg\/csi\/cinder\/|cmd\/cinder-csi-plugin\/)'
```
lingxiankong, line 74: - and `tests/sanity/cinder`?
- I noticed you didn't specify any labels, is that intentional?
- `branches` field is missing
- Does this test need to create GCP VM?
- Are you sure the kubekins-e2e image can directly run `make xxx` instead of `runner.sh make xxx`

lingxiankong (changes requested):

config/jobs/kubernetes/cloud-provider-openstack/release-master-presubmits.yaml
ramineni: > Does this test need to create GCP VM?
@lingxiankong It doesnt need to interact with the cloud, so I assume not required?

ramineni: > Are you sure the kubekins-e2e image can directly run make xxx instead of runner.sh make xxx

@lingxiankong Yes I suppose. It is being used at some places directly. So I added it.
eg: https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/cloud-provider-vsphere/cloud-provider-vsphere-config.yaml#L38

ramineni: > and tests/sanity/cinder?

@lingxiankong yes, thanks. Will add it.

> I noticed you didn't specify any labels, is that intentional?

@lingxiankong  As I dont need here to interact with cloud, i removed them. Ill check again if required. Thanks for pointing out.

> branches field is missing

@lingxiankong  I would like to run this job on release branches as well, should I move this job to different file? A here we are adding jobs only run on master right?

lingxiankong: > I would like to run this job on release branches as well, should I move this job to different file? A here we are adding jobs only run on master right?

Yeah, I prefer to split CI jobs for different branches even they are the same at the moment in between (e.g. create another file `release-master-presubmits-release-1.22.yaml`), in case the job script may change in the future, we experienced the issue with OpenLab before. I see there are some repos are using the same approach.

Different branches share the same testgrid dashboards but in different tabs, e.g. `presubmit-sanity-test-release-1.22`

ramineni: @lingxiankong Ok,sounds good. Will update to master for this job and add different one for other release branches.

ramineni:
@lingxiankong PTAL ^^

PR #23437, opened 2021-09-01
Fix PJs not being labled with tenantIDs

mpherman2:
Added a bit to the horlogium integration test to make sure this change actually worked. I will add more robust integration testing around tenantIDs in a followup PR


chaodaiG (approved):
glad it's so simple :)

chaodaiG (approved):

PR #23438, opened 2021-09-01
Regenerate the CSI prow jobs to use kubernetes v1.20, v1.21 and v1.22

mauriciopoppe:
Steps:

- updated gen-jobs.sh following the first step in https://github.com/kubernetes-csi/csi-release-tools/blob/master/SIDECAR_RELEASE_PROCESS.md#adding-support-for-a-new-kubernetes-release
- run the gen-jobs.sh script

/cc @pohly


config/jobs/kubernetes-csi/gen-jobs.sh
```diff
@@ -24,23 +24,23 @@ base="$(dirname $0)"
...
+1.22
 "
 
 # The experimental version for which jobs are optional.
-experimental_k8s_version="1.21"
+experimental_k8s_version="1.22"
 
 # The latest stable Kubernetes version for testing alpha jobs
-latest_stable_k8s_version="1.20" # TODO: bump to 1.21 after testing a pull job
+latest_stable_k8s_version="1.21" # TODO: bump to 1.21 after testing a pull job
```
pohly, line 43: You can remove the "TODO" comment.

pohly: Or rather, turn that into "TODO: bump to 1.22 after testing a pull job". We should have done that for the "on-kubernetes-1.21" already much earlier. They were still marked as "optional" when they should have been "required".

That's a missing step 9 in https://github.com/kubernetes-csi/csi-release-tools/blob/master/SIDECAR_RELEASE_PROCESS.md#adding-support-for-a-new-kubernetes-release

pohly:
/hold

There's one problem with this: if `on-kubernetes-1.21` jobs are failing (and it looks like there are?), then this PR here will make the situation worse because those failing jobs are no longer optional, i.e. they must pass before a PR can be merged.

This means those jobs must be fixed first.

pohly:
> There's one problem with this: if on-kubernetes-1.21 jobs are failing (and it looks like there are?)

They seem to be more flaky, but not outright broken. Let's merge this.

/hold cancel
/lgtm
/approve

PR #23429, opened 2021-08-31
Clickable hyperlinks in Spyglass Build log view

YuviGold:
Resolves #23311

Solution taken from https://stackoverflow.com/questions/29055828/regex-to-make-links-clickable-in-only-a-href-and-not-img-src
Probably not the best solution, but it works for now until we will find a better one. :)

Screenshot from 2021-09-01 02-09-50

Issue #23311, opened 2021-08-19
Clickable hyperlinks in Spyglass Build log view


YuviGold:
**What would you like to be added**:
Clickable hyperlinks for Spyglass Build log view.

**Why is this needed**:
Whenever a developer looks at the build logs, sometimes there are links that can be used for further debugging.
It would be faster if the user could click on the link and it would open it instead of copying it to the browser URL input.

for example for the following links:
```
INFO[2021-08-19T10:48:21Z] Loading configuration from https://config.ci.openshift.org for openshift/assisted-service@master 
INFO[2021-08-19T10:48:21Z] Resolved source https://github.com/openshift/assisted-service to master@fd2dcd0a, merging: #2423 856171be @YuviGold 
INFO[2021-08-19T10:48:21Z] Resolved release candidate to quay.io/openshift-release-dev/ocp-release@sha256:e64c04c41ae7717fff4b341987ac37c313045d4c3aa7bb8c6bfe8bf8540a5025 
INFO[2021-08-19T10:48:22Z] Using namespace https://console.build01.ci.openshift.org/k8s/cluster/projects/ci-op-iff5f2pm 
```

YuviGold:
Might be useable
https://soapbox.github.io/linkifyjs/

PR #23442, opened 2021-09-01
Fix spyglass hyperlink href path

YuviGold:
Related to https://github.com/kubernetes/test-infra/issues/23311. Fixes bug introduced on https://github.com/kubernetes/test-infra/pull/23429 .

Need to remove the quotes, otherwise the path tried to be resolved is
relative to the host while we would like the href link to be absolute.

For example, the bug `href="'${match}'"` caused the link
`http://127.0.0.1:2379`
to redirect to
`<host>/spyglass/static/buildlog/'http://127.0.0.1:2379'`

Removing the quotes fix this issue.

Additionally, fixes the replacement code to change the `innerHTML` and not
the `outerHTML`. This way the `outerHTML` element isn't changed.

### DIFF
--- a/prow/cmd/deck/static/spyglass/lens.ts
+++ b/prow/cmd/deck/static/spyglass/lens.ts
@@ -195,7 +195,7 @@ class SpyglassImpl implements Spyglass {
     if (typeof attr !== 'undefined') {
         return match;
     }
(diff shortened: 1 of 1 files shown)


michelle192837 (approved):
/hold

In case you want Chao's LGTM first, feel free to cancel the hold!

chaodaiG (approved):
/unhold

PR #23444, opened 2021-09-01
Switch bazelbuild/rules-k8s from Service Account to Workload Identity

chases2:
Workload identity has been enabled and associated with this service account, so we should no longer need to explicitly set them.

/assign @cjwagner @fejta

### DIFF
--- a/config/jobs/bazelbuild/rules_k8s/rules_k8s_config.yaml
+++ b/config/jobs/bazelbuild/rules_k8s/rules_k8s_config.yaml
@@ -11,19 +11,11 @@ presubmits:
         - bash
         - -c
         - |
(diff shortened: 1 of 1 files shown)


fejta:
> Workload identity has been enabled and associated with this service account

Which service account does this use?

chases2:
> Which service account does this use?

`prowjob-default-sa` -> `prow-ci@rules-k8s.iam.gserviceaccount.com`

The GSA is the same as the one that was in this volume

fejta (approved):

chases2:
The permissions aren't quite right: 

https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/bazelbuild_rules_k8s/665/pull-rules-k8s-e2e/1433501744628764672

PR #23455, opened 2021-09-02
Image push job for aws-iam-authenticator

nckturner:
* Enable cloudbuild image pushes for aws-iam-authenticator.


nckturner:
ping @fejta

fejta (approved):

PR #23456, opened 2021-09-02
Enable prow transfer-issue plugin for kubernetes org

eddiezane:
This PR enables the Prow transfer-issue plugin for the kubernetes org.

Now that https://github.com/kubernetes/test-infra/pull/23153 is merged I'm unsure if there is more to get this enabled than this change.

Do we want to enable for individual repos instead? My use case is to be able to transfer kubectl issues from k/k to the kubectl repo.

### DIFF
--- a/config/prow/plugins.yaml
+++ b/config/prow/plugins.yaml
@@ -770,6 +770,7 @@ plugins:
     - size
     - skip
     - slackevents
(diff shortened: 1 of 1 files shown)


spiffxp (approved):
/approve
/lgtm

PR #23458, opened 2021-09-02
kubetest: update README.md

spiffxp:
Related:
- part of: https://github.com/kubernetes/k8s.io/issues/1523
- followup to: https://github.com/kubernetes/test-infra/pull/23457

This is mostly a change to trigger a new build of kubekins-e2e

/hold
Don't want this to merge until https://github.com/kubernetes/test-infra/pull/23457 has

### DIFF
--- a/kubetest/README.md
+++ b/kubetest/README.md
@@ -1,7 +1,7 @@
 # Deprecation Notice
 
 July 14, 2020:  
-kubetest will be deprecated in favor of [kubetest2](https://github.com/kubernetes-sigs/kubetest2), 
+kubetest is deprecated in favor of [kubetest2](https://github.com/kubernetes-sigs/kubetest2), 
 so kubetest is moving to a maintenance mode only and we are not taking PRs except for urgent bug fixes. 
 
 # Kubetest


orsenthil (approved):
LGTM.

cblecker (approved):

spiffxp:
One thing I am confused by here is why the kubekins images built by the canary job have a slightly longer git sha as part of their tag

https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-test-infra-push-kubekins-e2e/1433543709261041664
>2021/09/02 21:42:36 Successfully built image: gcr.io/k8s-testimages/kubekins-e2e:v20210902-e4567b8-1.20 gcr.io/k8s-testimages/kubekins-e2e:latest-1.20  

https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-test-infra-push-kubekins-e2e-canary/1433543709189738496
>2021/09/02 21:41:25 Successfully built image: gcr.io/k8s-staging-test-infra/kubekins-e2e:v20210902-e4567b8e9c-1.19 gcr.io/k8s-staging-test-infra/kubekins-e2e:latest-1.19  

The old job is `e4567b8` or 7 char longs, the new job is `e4567b8e9c` or 10 chars long

spiffxp:
Image builder uses the output of `git show --tags --always --dirty` which ends up being an abbreviated SHA if not on a tag.

My local copy of git (2.33.0), running the same command, gives a SHA of length 10 as well.

The jobs run in different images.  It's likely the old image has a version of `git` older than 2.11.0, which is when a change landed that caused the default length of abbreviated SHAs to scale based on the repo size (ref: https://github.com/git/git/blob/master/Documentation/RelNotes/2.11.0.txt#L22-L26)

PR #23466, opened 2021-09-03
kubetest: run post-test-cmd under the same conditions as pre-test-cmd

jupblb:
`post-test-cmd` runs despite `pre-test-cmd` being skipped when the test does not have the `test` or `test-cmd` set. We believe it'd be better if `post-test-cmd` ran in the same conditions as `pre-test-cmd` for the sake of consistency.

/assign @tosi3k

### DIFF
--- a/kubetest/e2e.go
+++ b/kubetest/e2e.go
@@ -319,7 +319,7 @@ func run(deploy deployer, o options) error {
 		}))
 	}
 
-	if o.postTestCmd != "" {
+	if o.postTestCmd != "" && (o.test || (kubemarkUpErr == nil && o.testCmd != "")) {
 		errs = util.AppendError(errs, control.XMLWrap(&suite, "post-test command", func() error {
 			cmdLineTokenized := strings.Fields(os.ExpandEnv(o.postTestCmd))
 			return control.FinishRunning(exec.Command(cmdLineTokenized[0], cmdLineTokenized[1:]...))


kubetest/e2e.go
```diff
@@ -319,7 +319,7 @@ func run(deploy deployer, o options) error {
 		}))
 	}
 
-	if o.postTestCmd != "" {
+	if o.preTestCmd != "" && o.postTestCmd != "" {
```
jupblb, line 322: Actually since `preTestCmd` is only ran if `o.test` or `o.testCmd` maybe we should add it here too?

tosi3k:
```suggestion
	if o.postTestCmd != "" && (o.test || (kubemarkUpErr == nil && o.testCmd != "")) {
```

tosi3k (changes requested):
Please add some description to the PR to explain the context of the change.

kubetest/e2e.go
```diff
@@ -319,7 +319,7 @@ func run(deploy deployer, o options) error {
 		}))
 	}
 
-	if o.postTestCmd != "" {
+	if o.postTestCmd != "" && (o.test || (kubemarkUpErr == nil && o.testCmd != "")) {
```
spiffxp, line 322: For true parity, preTestCmd looks like it's run twice?
- once if `--test` was passed: https://github.com/kubernetes/test-infra/blob/0bd5b532b2f3f1b98a2958a8bb059ab494471e5f/kubetest/e2e.go#L169-L176
- and again if  `--test-cmd` is passed (and kubemark didn't err if `--kubemark` was specified): https://github.com/kubernetes/test-infra/blob/0bd5b532b2f3f1b98a2958a8bb059ab494471e5f/kubetest/e2e.go#L226-L233

spiffxp (approved):
/approve
/lgtm
AFAICT `post-test-cmd` isn't used by any jobs in this repo

PR #23467, opened 2021-09-03
wg-k8s-infra: auto-deploy prowjob for k8s-infra-prow

ameukam:
Auto-deploy k8s-infra-prow if no-one is on-call.

### DIFF
--- a/config/jobs/kubernetes/wg-k8s-infra/trusted/wg-k8s-infra-prow.yaml
+++ b/config/jobs/kubernetes/wg-k8s-infra/trusted/wg-k8s-infra-prow.yaml
@@ -104,7 +104,8 @@ periodics:
       - /app/prow/cmd/generic-autobumper/app.binary
       args:
       - --config=hack/autobump-config.yaml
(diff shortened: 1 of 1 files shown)


ameukam:
cc @spiffxp

chaodaiG (approved):
/hold
Just to show my support :)

ameukam:
> this will work. But recommend creating two jobs:
> 
>     * one add `skip-review` label, runs twice a day
> 
>     * the other one doesn't add `skip-review` label, runs much more frequently, for example https://github.com/kubernetes/test-infra/blob/790a4398fcf59aaa77945dc568c3c1db8dc2453a/config/jobs/kubernetes/test-infra/test-infra-trusted.yaml#L877
> 
> 
> by adding the second one, you'll be able to easily pause bump by commenting `/hold` on bumping PR when something bad is going on

Thank you for your input! I plan to do this as a followup once I fixed : https://github.com/kubernetes/k8s.io/issues/2559.

Currently k8s-infra-prow is not operational so auto-bump is not really a big deal.

Issue #23469, opened 2021-09-03
Cherrypicker plugin: Find GitHub account and deploy


alvaroaleman:
We added manifests and config for the cherrypicker plugin in https://github.com/kubernetes/test-infra/pull/16329

However this is not quite enough because:
* It needs to use a different GitHub account, because Prow ignores events from itself
* It doesn't get deployed into the cluster yet

/cc @spiffxp @dims @vincepri

cjwagner:
cc @mpherman2 
/assign @droslean 
The invalid config is spamming the Prow logs: `"Post "http://cherrypicker": dial tcp: lookup cherrypicker on 10.111.240.10:53: no such host"`

spiffxp:
I'm gonna make an external cherry picker bot user for this

alvaroaleman:
We added the deployment in #23471 and from the manifests and config it looks to me like it should work, however when trying it out it does not: https://github.com/kubernetes-sigs/controller-runtime/pull/1640#issuecomment-915432338

Could someone with access to the prow cluster check if its indeed deployed and running?

cjwagner:
> Could someone with access to the prow cluster check if its indeed deployed and running?

The deployment exists, but there are no associated pods. The deployment status includes this:
```yaml
  - lastTransitionTime: "2021-09-08T15:05:51Z"
    lastUpdateTime: "2021-09-08T15:05:51Z"
    message: 'pods "cherrypicker-5d78d679f8-" is forbidden: error looking up service
      account default/cherrypicker: serviceaccount "cherrypicker" not found'
    reason: FailedCreate
    status: "True"
    type: ReplicaFailure
```

It doesn't look like we actually need a service account so I made a PR to remove the reference: https://github.com/kubernetes/test-infra/pull/23517

alvaroaleman:
> It doesn't look like we actually need a service account so I made a PR to remove the reference: #23517

It still doesn't appear to be working unfortunately: https://github.com/kubernetes-sigs/controller-runtime/pull/1640#issuecomment-916072276

spiffxp:
Still working on getting an account for you, it was flagged by abuse prevention on creation, still working on getting that addressed

spiffxp:
OK after the following PR's:
- https://github.com/kubernetes/k8s.io/pull/2690
- https://github.com/kubernetes/k8s.io/pull/2693
- https://github.com/kubernetes/test-infra/pull/23551
[...]

PR #23471, opened 2021-09-03
add cherrypicker deployment in bazel prow configuration

droslean:
In effort of solving https://github.com/kubernetes/test-infra/issues/23469

### DIFF
--- a/config/prow/cluster/BUILD.bazel
+++ b/config/prow/cluster/BUILD.bazel
@@ -91,6 +91,7 @@ release(
         MULTI_KIND,
         cluster = BUILD_CLUSTER,
     ),
+    component("cherrypicker", "deployment", "service"),
 )
 
 # TODO(fejta): do we want to auto-apply this?


droslean:
/hold 

I am not sure if this is enough to deploy cherrypicker.

chaodaiG (approved):

chaodaiG:
I think this is all what we need for deploying cherrypicker in prow.k8s.io, it doesn't need RBAC right?

droslean:
> I think this is all what we need for deploying cherrypicker in prow.k8s.io, it doesn't need RBAC right?

Regarding the way that external-plugins work, they shouldn't need any special RBACs if they don't directly access a resource in the cluster. I think in `cherrypicker` case it doesn't need any.

Feel free to hold cancel, once you confirm this.

alvaroaleman:
No it doesn't need rbac, it just talks to github
/hold cancel

PR #23472, opened 2021-09-03
Label config: Add sharding support

alvaroaleman:
This change adds sharding support for the label.restricted_labels config
option.


petr-muller (approved):
/hold

LGTM, holding for more reviews if needed

PR #23473, opened 2021-09-03
Make release notes editable with prow

### LINKED ISSUE #9098: prow: add command to add release notes
Currently, only a few members have write access to k/k (and rightly so). See https://github.com/kubernetes/community/issues/2465 for more details on this.

The only way to add release notes in k/k is to edit the PR body. If I create a cherry-pick today and the parent PR does not contain a release note, I will need to ping someone with write access or the author of the parent PR to add the release note to the PR body (release notes are generated using the release note on the parent PR body, not the cherry-pick PR body).

The number of people with write access is considerably low now, which means pinging people isn't feasible anymore. See https://github.com/kubernetes/kubernetes/pull/67393#issuecomment-413777550 for an example.

I had discussed about this on slack and @cblecker had mentioned:

> in general, we need to figure out a way to handle release notes that doesn't manually require editing PR bodies. I'm not sure what the best way to do that is.. but we do need to have a consistent, single source of truth way to know what release note we will apply. This is actually one of the reasons we haven't completely removed manual write access to k/k

I agree and honestly, I don't know the answer to the above problem. Happy to bikeshed over this though. :)

/cc @cblecker @cjwagner @BenTheElder @tpepper 
/sig contributor-experience
/sig release
/area prow

eddiezane:
This PR adds edit functionality to the Prow release-note plugin.

Fixes #9098 

The idea was to leverage the release-note code block to be able to handle multi-line and allow trailing comments.

https://github.com/kuberneddies/test-repo-1/pull/3

### DIFF
--- a/prow/plugins/releasenote/releasenote.go
+++ b/prow/plugins/releasenote/releasenote.go
@@ -61,6 +61,7 @@ var (
 	}
 
 	releaseNoteRe               = regexp.MustCompile(`(?mi)^/release-note\s*$`)
+	releaseNoteEditRe           = regexp.MustCompile(`(?mi)^/release-note-edit\s*$`)
 	releaseNoteNoneRe           = regexp.MustCompile(`(?mi)^/release-note-none\s*$`)
 	releaseNoteActionRequiredRe = regexp.MustCompile(`(?mi)^/release-note-action-required\s*$`)
 )
@@ -86,6 +87,12 @@ func helpProvider(_ *plugins.Configuration, _ []config.OrgRepo) (*pluginhelp.Plu
 		WhoCanUse:   "PR Authors and Org Members.",
 		Examples:    []string{"/release-note-none"},
 	})
+	pluginHelp.AddCommand(pluginhelp.Command{
+		Usage:       "/release-note-edit",
+		Description: "Replaces the release note block in the top level comment with the provided one.",
+		WhoCanUse:   "Org Members.",
+		Examples:    []string{"/release-note-edit\r\n```release-note\r\nThe new release note\r\n```"},
+	})
 	return pluginHelp, nil
 }
 
@@ -98,6 +105,7 @@ type githubClient interface {
 	ListIssueComments(org, repo string, number int) ([]github.IssueComment, error)
 	DeleteStaleComments(org, repo string, number int, comments []github.IssueComment, isStale func(github.IssueComment) bool) error
 	BotUserChecker() (func(candidate string) bool, error)
+	EditIssue(org, repo string, number int, issue *github.Issue) (*github.Issue, error)
 }
 
 func handleIssueComment(pc plugins.Agent, ic github.IssueCommentEvent) error {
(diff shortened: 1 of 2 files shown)


petr-muller (approved):
Code LGTM but I'm 100% not familiar with how this plugin is used so I think one more review by someone more knowledgeable would be useful.

/hold

prow/plugins/releasenote/releasenote.go
```diff
@@ -451,3 +463,60 @@ func labelsSet(labels []github.Label) sets.String {
...
+	}
+	if !isMember {
+		return gc.CreateComment(
+			org, repo, ic.Issue.Number,
+			plugins.FormatResponseRaw(ic.Comment.Body, ic.Issue.HTMLURL, user, "You must be an org member to edit the release note."),
+		)
+	}
+
+	newNote := getReleaseNote(ic.Comment.Body)
+	if len(newNote) == 0 {
```
alvaroaleman, line 487: Nit: When checking string for emptyness, it is IMHO nicer to compare against the empty string vs len, because then if someone reads the code its immediately obvious that what you are comparing (here: `newNote`) is of type string

prow/plugins/releasenote/releasenote_test.go
```diff
@@ -633,3 +634,121 @@ func TestShouldHandlePR(t *testing.T) {
...
+					if !strings.Contains(cm[0].Body, tc.comment) {
+						t.Fatalf("expected comment to contain: %s got: %s", tc.comment, cm[0].Body)
+					}
+				}
+			}
+			if len(tc.comment) == 0 && len(fc.IssueComments[issueNum]) != 0 {
+				t.Fatalf("unexpected comment: %v", fc.IssueComments[issueNum])
+			}
+			if tc.expectEdit {
+				if !strings.Contains(fc.Issues[issueNum].Body, "```release-note\r\nThe new note\r\n```") {
```
alvaroaleman, line 748: Can we compare the full body string here instead of a substring check? The current check would succeed even if we were to just append the new block to the existing body

prow/plugins/releasenote/releasenote_test.go
```diff
@@ -633,3 +634,121 @@ func TestShouldHandlePR(t *testing.T) {
...
+				if cm, ok := fc.IssueComments[tc.event.Issue.Number]; ok {
+					if !strings.Contains(cm[0].Body, tc.comment) {
+						t.Fatalf("expected comment to contain: %s got: %s", tc.comment, cm[0].Body)
+					}
+				}
+			}
+			if len(tc.comment) == 0 && len(fc.IssueComments[issueNum]) != 0 {
+				t.Fatalf("unexpected comment: %v", fc.IssueComments[issueNum])
+			}
+			if tc.expectEdit {
```
alvaroaleman, line 747: We wouldn't catch it at all if we incorrectly edit (`tc.expectEdit == false`), right?

eddiezane:
@alvaroaleman changes made.

prow/plugins/releasenote/releasenote_test.go
eddiezane: Should catch this and prevent panics now.

alvaroaleman (approved):

PR #23476, opened 2021-09-03
Revert "Switch from Service Account to Workload Identity"

chases2:
This reverts commit 5a9114f46bcc2131f698860171ab8088f325f398. (and PR #23444)

The test isn't working due to a permission issue: https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/bazelbuild_rules_k8s/665/pull-rules-k8s-e2e/1433501744628764672

### DIFF
--- a/config/jobs/bazelbuild/rules_k8s/rules_k8s_config.yaml
+++ b/config/jobs/bazelbuild/rules_k8s/rules_k8s_config.yaml
@@ -11,11 +11,19 @@ presubmits:
         - bash
         - -c
         - |
-          gcloud container clusters get-credentials testing --zone=us-central1-f --project=rules-k8s \
+          gcloud auth activate-service-account --key-file=/etc/gcp/service-account.json \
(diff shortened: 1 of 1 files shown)


fejta:
So do you expect `prow-ci@rules-k8s.iam.gserviceaccount.com` to have the `Required "container.clusters.get" permission(s) for "projects/rules-k8s/zones/us-central1-f/clusters/testing".`?

chases2:
I do. I think that account to already have that permission; it seemed that the account in the mounted secret (prow-ci) was using that permission to launch some test containers into the rules-k8s "testing" cluster as part of the e2e test.

My current theory is that the job isn't running as the user I expect it to (prow-ci).

fejta:
I believe you need a `serviceAccount: foo` in the pod spec in order for the test containers to have access to it.

https://github.com/kubernetes/test-infra/blob/eea057f2aa3b0ba041a3a606562f664550d127df/prow/pod-utils/decorate/podspec.go#L214-L221

chases2:
The fix-forward worked. Closing revert request

PR #23483, opened 2021-09-05
Commenter: Add comments-ceiling flag to avoid spamming

YuviGold:
In order to avoid spamming of the commenter robot, a new flag is added
named `comments-ceiling`.
Before creating a new comment, the robot would verify that it isn't
already exist <`comments-ceiling`> amount of times in sequence.

In order to avoid the robot from being blocked indefinitely,
the `comments-ceiling-margin` declares what period would represent spam and only in
that period the robot would avoid creating the same comment.
In case the period margin is passed, then the robot can comment once
again.

An example where the [openshift-bot](https://github.com/openshift-bot) was spamming a PR (its job is defined under openshift/release on [infra-periodics.yaml](https://github.com/openshift/release/blob/c121e55f68fb37af41d7cd16877eaa79eeb972f1/ci-operator/jobs/infra-periodics.yaml#L202-L241)):
https://github.com/openshift/assisted-service/pull/2188

Screenshot from 2021-09-05 15-55-21

### DIFF
--- a/robots/commenter/main.go
+++ b/robots/commenter/main.go
@@ -70,6 +70,7 @@ func flagOptions() options {
 	flag.StringVar(&o.comment, "comment", "", "Append the following comment to matching issues")
 	flag.BoolVar(&o.useTemplate, "template", false, templateHelp)
 	flag.IntVar(&o.ceiling, "ceiling", 3, "Maximum number of issues to modify, 0 for infinite")
+	flag.IntVar(&o.spamCeiling, "spam-ceiling", 3, "Maximum number of the same comment in sequence to post, 0 for infinite")
 	flag.Var(&o.endpoint, "endpoint", "GitHub's API endpoint")
 	flag.StringVar(&o.graphqlEndpoint, "graphql-endpoint", github.DefaultGraphQLEndpoint, "GitHub's GraphQL API Endpoint")
 	flag.StringVar(&o.token, "token", "", "Path to github token")
@@ -88,6 +89,7 @@ type meta struct {
 type options struct {
 	asc             bool
 	ceiling         int
+	spamCeiling     int
 	comment         string
 	includeArchived bool
 	includeClosed   bool
@@ -155,6 +157,7 @@ func makeQuery(query string, includeArchived, includeClosed, includeLocked bool,
 type client interface {
 	CreateComment(owner, repo string, number int, comment string) error
 	FindIssues(query, sort string, asc bool) ([]github.Issue, error)
+	ListIssueComments(org, repo string, number int) ([]github.IssueComment, error)
 }
 
 func main() {
@@ -201,7 +204,7 @@ func main() {
 		asc = true
 	}
 	commenter := makeCommenter(o.comment, o.useTemplate)
-	if err := run(c, query, sort, asc, o.random, commenter, o.ceiling); err != nil {
+	if err := run(c, query, sort, asc, o.random, commenter, o.ceiling, o.spamCeiling); err != nil {
 		log.Fatalf("Failed run: %v", err)
 	}
 }
@@ -220,7 +223,7 @@ func makeCommenter(comment string, useTemplate bool) func(meta) (string, error) 
 	}
 }
 
-func run(c client, query, sort string, asc, random bool, commenter func(meta) (string, error), ceiling int) error {
+func run(c client, query, sort string, asc, random bool, commenter func(meta) (string, error), ceiling, spamCeiling int) error {
 	log.Printf("Searching: %s", query)
 	issues, err := c.FindIssues(query, sort, asc)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


robots/commenter/main.go
```diff
@@ -242,29 +245,51 @@ func run(c client, query, sort string, asc, random bool, commenter func(meta) (s
...
+	}
+
+	sameCommentsSequenceCounter := 0
+	for _, issueComment := range issueComments {
+		if comment != issueComment.Body {
+			sameCommentsSequenceCounter = 0
+		} else {
+			sameCommentsSequenceCounter += 1
+
+			if spamCeiling > 0 && sameCommentsSequenceCounter == spamCeiling {
```
petr-muller, line 281: I don't think this logic is the right thing to do. Doing it like this would mean containing such sequence even long time ago would make the issue/PR ineligible to be included, ever?

Maybe the logic could be "x times in a recent time period"?

robots/commenter/main.go
```diff
@@ -242,29 +245,51 @@ func run(c client, query, sort string, asc, random bool, commenter func(meta) (s
...
-			log.Print(msg)
-			problems = append(problems, msg)
-		}
-		comment, err := commenter(meta{Number: number, Org: org, Repo: repo, Issue: i})
-		if err != nil {
-			msg := fmt.Sprintf("Failed to create comment for %s/%s#%d: %v", org, repo, number, err)
-			log.Print(msg)
-			problems = append(problems, msg)
-			continue
+		if err := createComment(c, commenter, spamCeiling, i); err != nil {
```
petr-muller, line 248: I think we'd like the issues/PRs skipped like this to not being counted into the `--ceiling` limit

robots/commenter/main.go
```diff
@@ -155,6 +157,7 @@ func makeQuery(query string, includeArchived, includeClosed, includeLocked bool,
 type client interface {
 	CreateComment(owner, repo string, number int, comment string) error
 	FindIssues(query, sort string, asc bool) ([]github.Issue, error)
+	ListIssueComments(org, repo string, number int) ([]github.IssueComment, error)
```
petr-muller, line 160: I'm worried about the token consumption. This is a paginated api with a max of 100 results, so on longer issues it may be multiple API calls per PR.

robots/commenter/main.go
YuviGold: Sounds fair. I'll add a new flag for the amount of time that this check would apply for.

robots/commenter/main.go
YuviGold: This isn't related to this PR as the ceiling was always kept the same (`n` was being incremented for every interval regardless of a success or a failure). Anyway, I'll fix that.

robots/commenter/main.go
YuviGold: Are there any alternatives for this API? Is this something that in your opinion is a blocker for this feature?

robots/commenter/main.go
petr-muller: I asked for this because  previously the loop did not do any decisions about whether it wants to comment on the PRs or not: all cases when it didn't were errors.

robots/commenter/main.go
petr-muller: It's not a blocker for having the feature IMO, but potential blocker to using the feature (I'm not sure if we could afford to use it in OpenShift, for example).

I don't think there's an alternative API for this specific check. Most of Prow avoids doing this kind of checks because they are API-expensive, and Search API is the preferred way how to obtain sets of PRs/issues that conform to some criteria. Of course the Search API capabilities are limited.

I think this PR is simply a bandaid for `retester` being a dump Search->Comment wrapper over GH API. What we *really* need is 1) to make retester only comment on PRs where it makes sense (I think we would need to integrate it with Tide config to work over actual merge criteria queries, so it could do its searches based on that) and 2) find some API-friendly way to implement some retester backoff (which is IMO more desirable than a simple ceiling).

robots/commenter/main.go
```diff
@@ -236,35 +242,65 @@ func run(c client, query, sort string, asc, random bool, commenter func(meta) (s
...
+
+	if spamCeiling > 0 {
+		issueComments, err := c.ListIssueComments(org, repo, number)
+		if err != nil {
+			return fmt.Errorf("Failed to get comments for %s/%s#%d: %v", org, repo, number, err)
+		}
+
+		sameCommentsInSequenceCounter := 0
+		for _, issueComment := range issueComments {
+			if issueComment.Body != comment || issueComment.CreatedAt.Before(time.Now().Add(-1*spamMargin)) {
```
petr-muller, line 285: Aren't the comments sorted timewise? Meaning, could we break when we hit a comment that's before the margin?

nit: `Sub` instead of `Add(-1)`

petr-muller (commented):
LGTM with a minor comment

robots/commenter/main.go
YuviGold: I agree. I took the current implementation and patched it in order to cover those gaps, but probably the `retester` job/bot should just be more sophisticated. I'll look at how it could be done.

petr-muller: I never looked into it too close, but I always thought retesting could be an (optional?) feature of Tide itself. It knows the merge criteria, so it could search for all PRs that would be mergeable if not for failing tests (so retest users would not need to do the guesswork [like this](https://github.com/openshift/release/blob/b36a66636ee955d500edd29ad8c86e670e149067/ci-operator/jobs/infra-periodics.yaml#L212-L216) to specify the retest criteria). It could also optionally store some limited state (IIRC it already does - for storing history) for the retest backoff.

robots/commenter/main.go
```diff
@@ -70,6 +70,8 @@ func flagOptions() options {
 	flag.StringVar(&o.comment, "comment", "", "Append the following comment to matching issues")
 	flag.BoolVar(&o.useTemplate, "template", false, templateHelp)
 	flag.IntVar(&o.ceiling, "ceiling", 3, "Maximum number of issues to modify, 0 for infinite")
+	flag.IntVar(&o.spamCeiling, "spam-ceiling", 3, "Maximum number of the same comment in sequence to post, 0 for infinite")
```
alvaroaleman, line 73: if anything, this needs to be a neutral and descriptive `comment-ceiling`. Considering it "spam" is a subjective view for your current situation and when I look at this setting, it is unclear what exactly "spam" means.

robots/commenter/main.go
YuviGold: Yes, I expect the comments to be sorted timewise.
I don't understand what do you mean by the 2nd question. The margin was according to your request on https://github.com/kubernetes/test-infra/pull/23483#discussion_r702483588 .
Regarding the nit - `Sub` has a different API and returns a `Duration` instead of `Time`. I saw this approach in other modules as well.
Additionally, I think that the conversions won't make it essentially cleaner. Let me know if you think differently.

robots/commenter/main.go
YuviGold: Could you open a ticket for that?

robots/commenter/main.go
petr-muller: > Meaning, could we break when we hit a comment that's before the margin?

I meant that if the comments are sorted, we could `break` from the spam detection loop, because we iterated outside of the margin period, we did not hit the spam threshold until now, and we will not hit it because we now read older comments.

YuviGold: Yes, you are right.
According to https://docs.github.com/en/rest/reference/issues#comments 
```
By default, Issue Comments are ordered by ascending ID.
```

Actually, I notice the `since` parameter that we should use instead of running this logic on our side.

I'll add on a new commit a new function that would make this functionality simpler under `github.ListIssueCommentsSince`. Let me know what do you think.

fejta:
Are we concerned about the additional token use here (and/or attempted to measure it)?

For a comment like https://github.com/openshift/assisted-service/pull/2188#issuecomment-877723939 I would expect it to ideally leave a comment that includes a `/label foo` that flags it as needing whatever attention, and then removing it once it has received this attention.

petr-muller:
> For a comment like [openshift/assisted-service#2188 (comment)](https://github.com/openshift/assisted-service/pull/2188#issuecomment-877723939) I would expect it to ideally leave a comment that includes a `/label foo` that flags it as needing whatever attention, and then removing it once it has received this attention.

That's actually the case, it adds a merge-blocking label (not via comment but directly). 

> AKA the core problem here seems to be that the search cannot not distinguish between "I found new issue in state X" and "I've moved the issue from state X to Y" and it seems like the best way to address this is by improving the search.

Kinda. Our problem is that we have differing merge criteria on different sets of branches (over different repos), so we would need to maintain a set of commenter invocations inferred from Tide config. "PRs that would be mergeable if tests passed" is a query that you can only build with Tide config.

And even with that, a retest backoff is probably a desirable feature.

PR #23484, opened 2021-09-06
Change retest help message to indicate optional jobs

### LINKED ISSUE #23062: Indicate optional job on retest comment
When the retest help message is shown, it should indicate what jobs are required and what are not.

**Why is this needed**:
On https://github.com/kubernetes/test-infra/pull/22850 the `/retest-required` command was introduced.
Now whenver the retest comment is presented, it suggests to use either the `/retest` command to run everything or the `/retest-required` to run only the mandatory jobs. But it isn't clear which jobs are required and which are optional.

e.g.
Current comment - https://github.com/openshift/assisted-service/pull/2263#issuecomment-888958497

New comment - 
```
Test name | Commit | Details | Required | Rerun command
-- | -- | -- | -- | --
ci/prow/e2e-metal-assisted-olm | c5bb910 | link | X | /test e2e-metal-assisted-olm
ci/prow/e2e-metal-assisted | c5bb910 | link | V | /test e2e-metal-assisted
ci/prow/e2e-metal-assisted-operator-ztp | c5bb910 | link | V | /test e2e-metal-assisted-operator-ztp

```

YuviGold:
Resolves #23062

### DIFF
--- a/prow/crier/reporters/github/reporter.go
+++ b/prow/crier/reporters/github/reporter.go
@@ -32,6 +32,7 @@ import (
 	v1 "k8s.io/test-infra/prow/apis/prowjobs/v1"
 	"k8s.io/test-infra/prow/config"
 	"k8s.io/test-infra/prow/gerrit/client"
+	"k8s.io/test-infra/prow/git/v2"
 	"k8s.io/test-infra/prow/github/report"
 )
 
@@ -42,7 +43,8 @@ const (
 
 // Client is a github reporter client
 type Client struct {
-	gc          report.GitHubClient
+	gitClient   git.ClientFactory
+	ghc         report.GitHubClient
 	config      config.Getter
 	reportAgent v1.ProwJobAgent
 	prLocks     *shardedLock
@@ -102,9 +104,10 @@ func (s *shardedLock) runCleanup() {
 }
 
 // NewReporter returns a reporter client
-func NewReporter(gc report.GitHubClient, cfg config.Getter, reportAgent v1.ProwJobAgent) *Client {
+func NewReporter(gitClient git.ClientFactory, ghc report.GitHubClient, cfg config.Getter, reportAgent v1.ProwJobAgent) *Client {
 	c := &Client{
-		gc:          gc,
+		gitClient:   gitClient,
+		ghc:         ghc,
 		config:      cfg,
 		reportAgent: reportAgent,
 		prLocks: &shardedLock{
@@ -159,11 +162,12 @@ func (c *Client) Report(ctx context.Context, log *logrus.Entry, pj *v1.ProwJob) 
 		if err := lock.Acquire(ctx, 1); err != nil {
 			return nil, nil, err
 		}
+
 		defer lock.Release(1)
 	}
 
 	// TODO(krzyzacy): ditch ReportTemplate, and we can drop reference to config.Getter
-	err := report.Report(ctx, c.gc, c.config().Plank.ReportTemplateForRepo(pj.Spec.Refs), *pj, c.config().GitHubReporter.JobTypesToReport)
+	err := report.Report(ctx, c.config(), c.gitClient, c.ghc, c.config().Plank.ReportTemplateForRepo(pj.Spec.Refs), *pj, c.config().GitHubReporter.JobTypesToReport)
 	if err != nil {
 		if strings.Contains(err.Error(), "This SHA and context has reached the maximum number of statuses") {
 			// This is completely unrecoverable, so just swallow the error to make sure we wont retry, even when crier gets restarted.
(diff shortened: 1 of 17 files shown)


prow/crier/reporters/github/reporter.go
```diff
@@ -159,11 +164,19 @@ func (c *Client) Report(ctx context.Context, log *logrus.Entry, pj *v1.ProwJob)
 		if err := lock.Acquire(ctx, 1); err != nil {
 			return nil, nil, err
 		}
+
+		prRefGetter := config.NewRefGetterForGitHubPullRequest(c.ghc, pj.Spec.Refs.Org, pj.Spec.Refs.Repo, pj.Spec.Refs.Pulls[0].Number)
+		presubmits, err := c.config().GetPresubmits(c.gitClient, pj.Spec.Refs.Org+"/"+pj.Spec.Refs.Repo, prRefGetter.BaseSHA, prRefGetter.HeadSHA)
+		if err != nil {
+			return nil, nil, fmt.Errorf("failed to get Presubmits for pull request %s/%s#%d: %v", pj.Spec.Refs.Org, pj.Spec.Refs.Repo, pj.Spec.Refs.Pulls[0].Number, err)
+		}
+		presubmit = &presubmits[0]
```
petr-muller, line 173: I don't understand this, how does the `GetPresubmits` get the one exact presubmit config that we need (the one matching the prowjob we report?

YuviGold: Fixed.

petr-muller (approved):

chaodaiG (approved):

chaodaiG:
@YuviGold , this had caused regression on google operated prow instance. The root cause is the introduction of inrepoconfig in crier make all github workers waiting on a single mutex, and technically become single threaded. And since the fetching of inrepoconfig takes several seconds for large repo, crier is literally processing one prow job ~ every 10 seconds, and thus is severely lagged behind on processing prow jobs. Since all what inrepoconfig does in crier is checking for the existence of `optional: ` key, can we alternatively get this from prowjob CR itself instead?

Overall, the feature added in this PR is `nice-to-have` imho, I'm not quite convinced that we want to sacrifice performance with this feature. WDYT?

YuviGold:
> @YuviGold , this had caused regression on google operated prow instance. The root cause is the introduction of inrepoconfig in crier make all github workers waiting on a single mutex, and technically become single threaded. And since the fetching of inrepoconfig takes several seconds for large repo, crier is literally processing one prow job ~ every 10 seconds, and thus is severely lagged behind on processing prow jobs. Since all what inrepoconfig does in crier is checking for the existence of `optional: ` key, can we alternatively get this from prowjob CR itself instead?
> 
> Overall, the feature added in this PR is `nice-to-have` imho, I'm not quite convinced that we want to sacrifice performance with this feature. WDYT?


@chaodaiG I find this feature really convinient for developers that aren't familiar with every job definition. Then when a bunch of tests fail, it is clear right away which jobs are mandatory.
I already commented dozens of times "don't worry job X is only optional".

I would love to use Prowjobs only, but the `optional` field is only defined for `Presubmit` struct and I couldn't find any other way to cast from a `Prowjob` to a `Presubmit`.

But I agree that in case it's that resources consuming than we can revert. It's not something I could test :)

chaodaiG:
> @chaodaiG I find this feature really convinient for developers that aren't familiar with every job definition. Then when a bunch of tests fail, it is clear right away which jobs are mandatory.
> I already commented dozens of times "don't worry job X is only optional".
> 
> I would love to use Prowjobs only, but the `optional` field is only defined for `Presubmit` struct and I couldn't find any other way to cast from a `Prowjob` to a `Presubmit`.
> 
> But I agree that in case it's that resources consuming than we can revert. It's not something I could test :)

@YuviGold, thank you for the explanation, totally agree that there is no way you could have tested the scenario we encountered, it's also obvious that I failed to observe this when I stamped this PR :)

It's unfortunate that there is not a good way of finding `optional` from prowjob, we had recently added this field in postsubmit job, so it might be ok to make this a field in base prow struct? @cjwagner and @alvaroaleman for opinions.

@YuviGold , if you don't mind please revert this for now, thank you!

cjwagner:
> It's unfortunate that there is not a good way of finding `optional` from prowjob, we had recently added this field in postsubmit job, so it might be ok to make this a field in base prow struct?

I don't think we did? Are you thinking of `always_run`?

Either way, if this feature is needed we should be indicating that the job is optional somewhere on the ProwJob resource, this doesn't seem worth making crier dependent on in-repo config and git clients. I'm not sure whether the `optional` field should be included in the ProwJob spec itself or via an annotation since it only is applicable to presubmits, but we can make a decision about that.

chaodaiG:
Created https://github.com/kubernetes/test-infra/pull/23599 btw

PR #23500, opened 2021-09-07
subscriber: use LRU cache

listx:
We add a basic LRU (Least Recently Used) cache to try to avoid expensive
lookups (config.GetPresubmits and config.GetPostsubmits). An LRU cache
is basically a smart map that has a set size, and evicts key/value pairs
based on when they were last accessed (the more a key/value pair is
unused, the more likely that it will get evicted).

The size of the cache is exposed as a flag in the "sub" command as
"-cache-size <INT>", which by default is set to 1000. Refer to the
cache.go file for an in-depth discussion about the caching
implementation. We try to be as robust as possible in the following
ways:

- we use an off-the shelf cache library (it uses the empty interface{}
  but it's a fair price to pay for robustness)
- if the cache cannot be initialized, we continue without one
- if a cache hit results in a bad type, we throw it away and heal the
  cache entry

/assign @cjwagner


prow/pubsub/subscriber/cache.go
```diff
@@ -0,0 +1,311 @@
...
+
+	// Cache miss. We have no choice but to construct the value (this call is
+	// expensive!) and add it to the cache.
+	valConstructed, err := valConstructor()
+	if err != nil {
+		return nil, false, false, err
+	}
+
+	// Add our constructed value to the cache.
+	evicted := cache.Add(key.String(), valConstructed)
```
cjwagner, line 308: I think the cache miss behavior is wrong for our use case. In particular, this cache will frequently encounter bursts of nearly simultaneous requests for new keys that we haven't cached yet. When that happens we want to calculate the value once and return that to all consumers, but the current behavior for a cache miss is to just calculate the value and then add it to the cache, even if multiple other threads are doing the same thing concurrently. Calculating the value once for concurrent cache misses is called request coalescing and `ghproxy` implements it if you are looking for an example.
I think this is something that needs to be supported at the cache locking layer so hopefully the LRU library supports this. If not, we could either add a redundant lock layer that we control, or we could make a very simple approximation of an LRU as we discussed before and use that instead.

prow/pubsub/subscriber/subscriber.go
```diff
@@ -212,7 +213,7 @@ func (prh *presubmitJobHandler) getProwJobSpec(cfg prowCfgClient, pe ProwJobEven
 
 	presubmits := cfg.GetPresubmitsStatic(orgRepo)
 	if prh.GitClient != nil { // Get from inrepoconfig only when GitClient is provided
-		presubmitsWithInrepoconfig, err := cfg.GetPresubmits(prh.GitClient, orgRepo, baseSHAGetter, headSHAGetters...)
+		presubmitsWithInrepoconfig, _, _, err := cache.GetPresubmitsFromCache(cfg.GetPresubmits, prh.GitClient, orgRepo, baseSHAGetter, headSHAGetters...)
```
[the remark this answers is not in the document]
listx: > So even if we just cache the ProwYAML, it will still be possible for the cache result to be stale if the merge method is changed.

Why? Even if the merge strategy is changed, if the baseSHA and headSHAs stay the same, the end result (what the git tree looks like post-merge) should be the same, right?

prow/pubsub/subscriber/cache.go
```diff
@@ -0,0 +1,311 @@
...
+	// Append "headSHAs" string information.
+	var headSHAs []string
+	for _, headSHAGetter := range headSHAGetters {
+		headSHA, err := headSHAGetter()
+		if err != nil {
+			return CacheKey{}, fmt.Errorf("failed to get headRef: %v", err)
+		}
+		headSHAs = append(headSHAs, headSHA)
+	}
+	key.headSHAs = headSHAs
```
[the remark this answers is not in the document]
chaodaiG: sorting conflicts with the comment at line #108:
```
// Users should take care to ensure that headSHAs remains stable (order
// matters).
```

tide already sorts PRs by their numbers: https://github.com/kubernetes/test-infra/blob/9d132105e57fd9690c1a8918dc54bea604806655/prow/tide/tide.go#L1094, so we can assume that the SHAs are already sorted, maybe not sorting makes a bit more sense?

prow/pubsub/subscriber/cache.go
```diff
@@ -0,0 +1,311 @@
...
+}
+
+// MakeCacheKey constructs a CacheKey type from uniquely-identifying
+// information. The only requirement is that we take all of the stringlike
+// parameters and concatenate them together to form a UUID. However to make
+// debugging easier, we
+func MakeCacheKey(
+	identifier string,
+	baseSHAGetter config.RefGetter,
+	headSHAGetters ...config.RefGetter) (CacheKey, error) {
```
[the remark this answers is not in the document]
listx: Instead of constructing a custom concatenated string I just marshaled it to JSON.

prow/pubsub/subscriber/subscriber.go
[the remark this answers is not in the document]
listx: I tried to move the defaulting stuff out of `defaultProwYAMLGetter` with a naive change but it broke a number of existing unit tests. Looks like the refactoring of `defaultProwYAMLGetter` to reuse the defaulting logic is tricker than I initially expected. I think this is the only blocker for this PR now.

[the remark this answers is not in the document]
listx: Done. PTAL

prow/config/inrepoconfig.go
```diff
@@ -186,6 +201,11 @@ func DefaultAndValidateProwYAML(c *Config, p *ProwYAML, identifier string) error
 		}
 	}
 
+	if len(errs) == 0 {
+		log := logrus.WithField("repo", identifier)
+		log.Debugf("Successfully got %d presubmits and %d postsubmits from %q.", len(p.Presubmits), len(p.Postsubmits), inRepoConfigFileName)
+	}
```
cjwagner, line 207: Did you meant to leave this? If so please remove the part about the file name since that is not necessarily accurate.

prow/config/cache.go
```diff
@@ -0,0 +1,388 @@
...
+	if ok {
+		return prowYAML, err
+	}
+
+	// Somehow, the value retrieved with GetFromCache has a malformed type. This
+	// can happen if some other function modified the cache. Ultimately, this is
+	// a price we pay for using a cache library that uses "interface{}" for the
+	// type of its items. In this case, we log a warning and return an error.
+	err = fmt.Errorf("cache value type error: expected value type '*config.ProwYAML', got '%T'", val)
+	logrus.Warn(err)
```
chaodaiG, line 267: Is this recoverable error? if not, this log should be on error level

prow/config/inrepoconfig.go
```diff
@@ -49,10 +49,11 @@ type ProwYAML struct {
 // their own implementation and set that on the Config.
 type ProwYAMLGetter func(c *Config, gc git.ClientFactory, identifier, baseSHA string, headSHAs ...string) (*ProwYAML, error)
 
-// Verify defaultProwYAMLGetter is a ProwYAMLGetter
+// Verify prowYAMLGetter is a ProwYAMLGetter
```
chaodaiG, line 52: update comment to mention both functions below

prow/config/cache.go
```diff
@@ -0,0 +1,388 @@
...
+		// delete the cached entry so that we may attempt to re-try again in the
+		// future (instead of waiting for the LRUCache to evict it on its own
+		// over time).
+		//
+		// TODO: If our cache implementation supports a TTL mechanism, then we
+		// could just set that instead and let the cached entry to expire on its
+		// own.
+		if promise.err != nil {
+			lruCache.Lock()
+			_ = lruCache.Remove(key)
```
chaodaiG, line 382: Not sure if it's the best strategy let all others fail when this thread failed, but this might be the best strategy so far, as we don't have any retry on inrepoconfig now. I would however expecting to see an error log here though, WDYT?

chaodaiG (commented):
Generally looks good to me, several nits

prow/config/cache.go
```diff
@@ -0,0 +1,388 @@
...
+//
+// This cache is resistant to cache stampedes because it uses a duplicate
+// suppression strategy. This is also called request coalescing.
+func (lruCache *LRUCache) GetOrAdd(
+	keyConstructor keyConstructor,
+	valConstructor valConstructor) (interface{}, error) {
+
+	// If the cache is unreachable, then fall back to cache-less behavior
+	// (construct the value from scratch).
+	if lruCache == nil {
```
cjwagner, line 286: nit: I'm not sure it makes sense to support using the `Get*FromCache` functions at all if the cache hasn't been initialized. Rather than optionally initializing a field and then checking if it initialized, it might be cleaner to handle and use the cache directly so that if the calls are made the cache must exist.
Not that important either way though, just something to consider.

prow/config/cache.go
```diff
@@ -0,0 +1,388 @@
...
+		valConstructed, err := valConstructor()
+		if err != nil {
+			return nil, err
+		}
+
+		return valConstructed, nil
```
cjwagner, line 292:
```suggestion
		return valConstructor()
```

prow/config/cache.go
```diff
@@ -0,0 +1,388 @@
...
+// (instead of needing to create an actual Git repo, etc.).
+func GetProwYAMLFromCache(
+	prowYAMLCache *ProwYAMLCache,
+	valConstructorHelper func(git.ClientFactory, string, RefGetter, ...RefGetter) (*ProwYAML, error),
+	gc git.ClientFactory,
+	identifier string,
+	baseSHAGetter RefGetter,
+	headSHAGetters ...RefGetter) (*ProwYAML, error) {
+
+	keyConstructor := func() (CacheKey, error) {
```
cjwagner, line 236: Do we this to be wrapped in a function? I think we can just calculate the cache key and pass it rather than a key constructor function.

prow/config/cache.go
```diff
@@ -0,0 +1,388 @@
...
+		//
+		// For now we just unlock the overall lruCache itself so that it can
+		// service other GetOrAdd() calls to it.
+		lruCache.Unlock()
+
+		// If the type is not a promise type, there's no need to wait and we can
+		// just return immediately with an error.
+		promise, ok = maybePromise.(*Promise)
+		if !ok {
+			return nil, fmt.Errorf("invalid cache entry type '%T', expected '*Promise'", maybePromise)
```
cjwagner, line 321: This would indicate a programmer error, right? If so please log it as such.

prow/config/cache.go
```diff
@@ -0,0 +1,388 @@
...
+
+// ProwYAMLCache is the user-facing cache.
+type ProwYAMLCache struct {
+	LRUCache
+}
+
+// LRUCache is the actual concurrent non-blocking cache.
+type LRUCache struct {
+	*sync.Mutex
+	*simplelru.LRU
```
cjwagner, line 57: Nice, I like how you cleaned up the locking.

prow/config/cache.go
```diff
@@ -0,0 +1,388 @@
...
+// single lock, and use an algorithm for a concurrent non-blocking cache to make
+// it both thread-safe and also resistant to so-called cache stampede, where
+// many concurrent threads all attempt to look up the same (missing) key/value
+// pair from the cache (see Alan Donovan and Brian Kernighan, "The Go
+// Programming Language" (Addison-Wesley, 2016), p. 277).
+
+// ProwYAMLCache is the user-facing cache.
+type ProwYAMLCache struct {
+	LRUCache
+}
```
cjwagner, line 52: nit: Its a bit unclear to me why we have both of these types. `ProwYAMLCache` doesn't seem to add any functionality or type assurances to `LRUCache`. I'd suggest either removing the extra layer or making it valuable by having it handle the type casting needed for the `LRUCache`.

Keeping these separate to keep the types general is appealing to me. This coalescing LRU cache seems generally useful and reusable. We may have additional applications for it in the future as well. :+1:

prow/config/cache.go
```diff
@@ -0,0 +1,388 @@
...
+		// If the value construction (expensive operation) failed, then we
+		// delete the cached entry so that we may attempt to re-try again in the
+		// future (instead of waiting for the LRUCache to evict it on its own
+		// over time).
+		//
+		// TODO: If our cache implementation supports a TTL mechanism, then we
+		// could just set that instead and let the cached entry to expire on its
+		// own.
+		if promise.err != nil {
+			lruCache.Lock()
```
cjwagner, line 381: There is technically a race here since we have to relock the cache. The entry we want to remove may already have been deleted and may even have been replaced with a new valid value.
That race is unlikely to occur and the outcome would at worst be unnecessarily invalidating a cache entry so this trick seems worth while and safe enough.

prow/config/config.go
```diff
@@ -362,6 +374,36 @@ func (c *Config) getProwYAML(gc git.ClientFactory, identifier string, baseSHAGet
...
 
+// getProwYAMLNoDefault is like getProwYAML, but it does __not__ use the
+// information found inside the retrieved *ProwYAML to modify the Config.
+func (c *Config) getProwYAMLNoDefault(gc git.ClientFactory, identifier string, baseSHAGetter RefGetter, headSHAGetters ...RefGetter) (*ProwYAML, error) {
+	if identifier == "" {
+		return nil, errors.New("no identifier for repo given")
+	}
+	if !c.InRepoConfigEnabled(identifier) {
+		return &ProwYAML{}, nil
+	}
```
cjwagner, line 385: As currently written, we'll cache these results as well. In other words we'll be caching the empty inrepo configs of repos that don't enable inrepo config which will waste cache space and negatively impact eviction behavior.
We should instead only cache inrepo config for repos that enable it. Checking if a repo enables inrepo config is efficient and does not need to be cached, we can check that every request.

cjwagner (commented):
Coalescing logic looks great! I have a hunch that this cache will be useful for other purposes down the road.

prow/config/cache.go
listx: I agree with your concern, because I don't think it's unreasonable to expect that users of `Get*FromCache` have already taken care of setting up the cache. I did some refactoring to move this nilcheck to downstream users (into `GetPresubmitsMaybeCached` and `GetPostsubmitsMaybeCached`).

prow/config/cache.go
listx: > Is this recoverable error?

To recover we would have to delete the bad key/value pair, but also perform some sort of retry. I think the engineering effort for handling this particular recovery is not worth it, given how it most likely boils down to programmer error.

> if not, this log should be on error level

done

> Maybe also indicate this is a programmer error.

done

prow/config/cache.go
listx: Done

prow/config/cache.go
listx: > nit: Its a bit unclear to me why we have both of these types. `ProwYAMLCache` doesn't seem to add any functionality or type assurances to `LRUCache`.

Agreed.

> I'd suggest either removing the extra layer or making it valuable by having it handle the type casting needed for the `LRUCache`.

I did the latter.

> Keeping these separate to keep the types general is appealing to me. This coalescing LRU cache seems generally useful and reusable. We may have additional applications for it in the future as well. 👍

I agree, and to keep things cleaner, moved out the lower-level `LRUCache` stuff to its own separate package.

prow/cmd/sub/main.go
```diff
@@ -136,6 +138,13 @@ func main() {
 	metrics.ExposeMetrics("sub", configAgent.Config().PushGateway, flagOptions.instrumentationOptions.MetricsPort)
 	pprof.Instrument(flagOptions.instrumentationOptions)
 
+	// Initialize cache for fetching Presubmit and Postsubmit information. If
+	// the cache cannot be initialized (e.g., cache size is too big), continue
+	// without one.
+	if err := configAgent.Config().InitProwYAMLCache(flagOptions.prowYAMLCacheSize); err != nil {
+		logrus.WithError(err).Warnf("unable to initialize ProwYAMLCache (prowYAMLCacheSize: %d); continuing without one", flagOptions.prowYAMLCacheSize)
```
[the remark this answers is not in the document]
listx: Done

prow/cmd/sub/main.go
```diff
@@ -81,6 +82,7 @@ func init() {
 
 	fs.BoolVar(&flagOptions.dryRun, "dry-run", true, "Dry run for testing. Uses API tokens but does not mutate.")
 	fs.DurationVar(&flagOptions.gracePeriod, "grace-period", 180*time.Second, "On shutdown, try to handle remaining events for the specified duration. ")
+	fs.IntVar(&flagOptions.prowYAMLCacheSize, "ProwYAMLCache-size", 1000, "Cache size for ProwYAMLCache")
```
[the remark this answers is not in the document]
listx: Done

listx (commented):
PTAL. @cjwagner In particular have a closer look at how the cache is used from `sub` to see if it makes sense.

prow/cache/cache.go
```diff
@@ -0,0 +1,203 @@
...
+		// For now we just unlock the overall lruCache itself so that it can
+		// service other GetOrAdd() calls to it.
+		lruCache.Unlock()
+
+		// If the type is not a promise type, there's no need to wait and we can
+		// just return immediately with an error.
+		promise, ok = maybePromise.(*Promise)
+		if !ok {
+			err := fmt.Errorf("Programmer error: expected cache entry type '*Promise', got '%T'", maybePromise)
+			logrus.Error(err)
```
chaodaiG, line 111: This log doesn't have context, suggest adding key to make this log more helpful
```suggestion
			logrus.WithField("key", key).Error(err)
```

chaodaiG (commented):
LGTM, left my last comment

prow/cache/cache.go
listx: done

prow/config/cache.go
```diff
@@ -0,0 +1,219 @@
...
+func MakeCacheKey(kp CacheKeyParts) (CacheKey, error) {
+	data, err := json.Marshal(kp)
+
+	return CacheKey(data), err
+}
+
+// GetPresubmitsCached is like GetPresubmits, but uses a cache lookup to get the
+// *ProwYAML value (cache hit), instead of computing it from scratch (cache
+// miss). It also stores the entry into the cache if there is a cache miss.
+func (c *Config) GetPresubmitsCached(pc *ProwYAMLCache, gc git.ClientFactory, identifier string, baseSHAGetter RefGetter, headSHAGetters ...RefGetter) ([]Presubmit, error) {
```
cjwagner, line 120: This isn't what I meant to describe. I don't think this really makes sense as a method of `Config`. It would be easier to use this if `ProwYAMLCache` was initialized with a reference to a config agent (or `config.Getter`) and a `git.ClientFactory`. Then we could just call `cache.GetPresubmits(identifer, baseSHAGetter, headSHAGetters...)` which is more concise.

This would also let us remove the need to pass a `git.ClientFactory` around separately. It would only be needed in the main function.
The `...MaybeCached` functions could also be dropped since we're either using the cache directly or not at all.

prow/config/config.go
```diff
@@ -106,6 +106,15 @@ type JobConfig struct {
...
 	ProwYAMLGetter ProwYAMLGetter `json:"-"`
 
+	// ProwYAMLGetterNoDefault is like ProwYAMLGetter, but without default
+	// values (it does not call DefaultAndValidateProwYAML()). Its sole purpose
+	// is to allow caching of ProwYAMLs that are retrieved purely from the
+	// inrepoconfig's repo, __without__ having the contents modified by the main
+	// Config's own settings (which happens mostly inside
+	// DefaultAndValidateProwYAML()). ProwYAMLGetterNoDefault is only used by
+	// GetProwYAMLCached().
+	ProwYAMLGetterNoDefault ProwYAMLGetter `json:"-"`
```
cjwagner, line 116: This function only needs to be a field of the config if you are mocking it for unit tests. Otherwise it can just be a normal method.

prow/cache/cache.go
```diff
@@ -0,0 +1,220 @@
...
+		if promise.err != nil {
+			logrus.WithField("key", key).Infof("promise was successfully resolved, but the call to resolve() returned an error; deleting key from cache...")
+			lruCache.Lock()
+			keyWasFoundBeforeRemoval := lruCache.Remove(key)
+			lruCache.Unlock()
+			if keyWasFoundBeforeRemoval {
+				logrus.WithField("key", key).Infof("successfully deleted")
+			} else {
+				err := fmt.Errorf("unexpected race: key deleted by the cache without our knowledge; our own deletion of this key was a NOP")
+				logrus.WithField("key", key).Error(err)
```
cjwagner, line 214: This could happen during normal operation (although very unlikely) and doesn't constitute a problem. Please lower this to at least info level and indicate that this is not problematic in the message.

prow/cache/cache.go
```diff
@@ -0,0 +1,220 @@
...
+// waitForResolution blocks the current thread until the first thread that
+// detected a cache miss has finished constructing the value (see resolve()).
+func (p *Promise) waitForResolution() {
+	<-p.valConstructionPending
+}
+
+// resolve resolves the Promise by constructing the value and closing the
+// valConstructionPending channel, thereby unblocking any other thread that has
+// been waiting for the value to be constructed.
+func (p *Promise) resolve() {
```
cjwagner, line 83: :+1:

prow/config/cache_test.go
```diff
@@ -0,0 +1,527 @@
...
+
+			maybeEnabled := make(map[string]*bool)
+			maybeEnabled[tc.identifier] = &tc.inRepoConfigEnabled
+			c := Config{
+				ProwConfig: ProwConfig{
+					InRepoConfig: InRepoConfig{
+						Enabled: maybeEnabled,
+					},
+				},
+			}
```
cjwagner, line 480: Please add a unit test to ensure that merging of the `Config` struct is working properly.
For example, add a test that ensures that the cache behaves properly with `DefaultDecorationConfigEntries`. Specifically, check that defaulting is occurring and that updates to `Config` are reflected even when retrieving cached values.

cjwagner (commented):
Looks good! My main remaining requests are for a change to how the cache is used, and unit testing of the config defaulting.

prow/config/cache_test.go
listx: done

prow/cache/cache.go
listx: done

prow/config/config.go
listx: done

prow/config/cache.go
listx: done

fejta:
/approve
/uncc

Add me back if you need my help on the reviews

prow/pipeline/listers/pipeline/v1alpha1/run.go
```diff
@@ -0,0 +1,99 @@
+/*
```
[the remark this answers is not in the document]
listx: When I ran `hack/update-codegen.sh`, even before I added the the ProwYAML deepcopy changes, these files got generated. So I added them here because ultimately they would get generated anyway.

prow/config/jobs.go
```diff
@@ -158,9 +167,30 @@ type Presubmit struct {
...
+// CopyableRegexp wraps around regexp.Regexp. It's sole purpose is to allow us to
+// create a manual DeepCopyInto() method for it, because the standard library's
+// regexp package does not define one for us (making it impossible to generate
+// DeepCopy() methods for any type that uses the regexp.Regexp type directly).
+type CopyableRegexp struct {
+	*regexp.Regexp
+}
+
+func (in *CopyableRegexp) DeepCopyInto(out *CopyableRegexp) {
+	*out = *in
```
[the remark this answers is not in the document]
listx: I saw that Regexp.Copy doesn't do anything special (https://cs.opensource.google/go/go/+/refs/tags/go1.17.1:src/regexp/regexp.go;l=117-120), and also it has been deprecated, which is why I did it here manually (the generated DeepCopy() function calls `new()` to allocate a separate variable.

Anyway, should we just make the DeepCopy and DeepCopyInto functions do a NOP instead?

prow/config/jobs.go
```diff
@@ -36,12 +37,16 @@ const (
...
+// +k8s:deepcopy-gen=true
+
 // Presets can be used to re-use settings across multiple jobs.
 type Preset struct {
-	Labels       map[string]string `json:"labels"`
-	Env          []v1.EnvVar       `json:"env"`
-	Volumes      []v1.Volume       `json:"volumes"`
-	VolumeMounts []v1.VolumeMount  `json:"volumeMounts"`
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
```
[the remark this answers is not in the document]
listx: Oops, forgot to delete these while I was experimenting with deepcopy-gen.

prow/config/config.go
```diff
@@ -513,25 +561,29 @@ type Plank struct {
 	PodUnscheduledTimeout *metav1.Duration `json:"pod_unscheduled_timeout,omitempty"`
 
 	// DefaultDecorationConfigs holds the default decoration config for specific values.
-	// Each entry in the slice specifies Repo and Cluster regexp filter fields to
-	// match against jobs and a corresponding DecorationConfig. All entries that
-	// match a job are used. Later matching entries override the fields of earlier
-	// matching entries.
```
cjwagner, line 519: nit: This documentation is pretty important because it explains the order of precedence of the defaults. Please do not remove it.

cjwagner (approved):
I have one remaining nit about some documentation, but the code looks great. Thanks for making all the changes!

prow/config/config.go
listx: These lines were duplicated in the original; if you look at the comment for `DefaultDecorationConfigEntries` it's still there. I just thought that we shouldn't duplicate comments so I cleaned this up but I can put it back if you want. WDYT?

listx: I guess it makes more sense to comment this block instead and remove the duplicated comment for `DefaultDecorationConfigEntries`, because ultimately the `DefaultDecorationConfigs` is what we're going to use downstream.

PR #23507, opened 2021-09-08
Update log key configuration name to reflect the change in kind.

pchan:
Update the environment flag in the pod template to reflect the change in [Kind](https://github.com/kubernetes-sigs/kind/pull/2387) . Resolves [10078](https://github.com/kubernetes/kubernetes/issues/100748).


serathius:
After periodic runs please confirm that components write json as expected.

PR #23509, opened 2021-09-08
Add serviceAccount to bazelbuild/rules-k8s job

chases2:
Attempt to fix-forward the permission issue at https://github.com/kubernetes/test-infra/pull/23476

Currently, the best way of testing seems to be to try and fix forward and test after merging.
Applying the prow job before merging leads to it simply hanging as "triggered"

/assign @fejta

### DIFF
--- a/config/jobs/bazelbuild/rules_k8s/rules_k8s_config.yaml
+++ b/config/jobs/bazelbuild/rules_k8s/rules_k8s_config.yaml
@@ -5,6 +5,7 @@ presubmits:
     always_run: true
     decorate: true
     spec:
+      serviceAccount: prowjob-default-sa
       containers:
       - image: gcr.io/rules-k8s/gcloud-bazel:v20210501-v0.6-28-gaf2e97d
         args:


fejta:
> the best way of testing

What I would recommend for you to test this sort of thing is:
* go to prow.k8s.io
* click the rerun button
* copy some link like https://prow.k8s.io/rerun?prowjob=8995c5a2-10cc-11ec-b73a-e65334ee5b42
  - obviously use a link for the correct job
* `curl -v -L https://prow.k8s.io/rerun?prowjob=8995c5a2-10cc-11ec-b73a-e65334ee5b42 -o /tmp/foo.yaml`
* `vim /tmp/foo.yaml` and patch in your change
* `kubectl --context=this-prow-cluster create -f /tmp/foo.yaml`
* wait for job to complete and pass

This of course depends on your ability to write prowjobs to the apiserver (which you can do but is not generally accessible to people).

fejta (approved):

PR #23517, opened 2021-09-08
Remove non-existent and unnecessary cherrypicker service account.

cjwagner:
This was preventing the deployment from creating pods. I don't think the service account is needed.
ref: https://github.com/kubernetes/test-infra/issues/23469
/assign @alvaroaleman


dims:
thanks @cjwagner 

/approve
/lgtm

PR #23465, opened 2021-09-03
config/jobs: use k8s-infra hosted kubekins-e2e in canary jobs

spiffxp:
Related:
- Part of: https://github.com/kubernetes/k8s.io/issues/1523#issuecomment-912031929
- Followup to: https://github.com/kubernetes/test-infra/pull/23458

Choosing frequent canary jobs that are passing
- ci-kubernetes-cached-make-test
- ci-kubernetes-e2e-gce-canary
- ci-kubernetes-e2e-prow-canary
- pull-kubernetes-unit-go-canary
- pull-kubernetes-integration-go-canary

PR #23532, opened 2021-09-09
config/prow: setup autobumping of community-hosted kubekins

spiffxp:
Related:
- Part of: https://github.com/kubernetes/k8s.io/issues/1523#issuecomment-912031929
- Followup to: https://github.com/kubernetes/test-infra/pull/23465

Calling out community-hosted kubekins to get bumped at the same time as everything in k8s-testimages

Then include a commit that will cause a new version of kubekins to get pushed, which should trigger a new autobump PR to confirm this worked


spiffxp:
/cc @chaodaiG 
test-infra-oncall

spiffxp:
Hrm, for some reason the kubetest change didn't trigger a postsubmit build for either job?

PR #23546, opened 2021-09-10
config/jobs: add pull-enhancements-test

spiffxp:
Related:
- Part of: https://github.com/kubernetes/enhancements/issues/2276
- Followup to: https://github.com/kubernetes/enhancements/pull/2927

Let's prevent `make test` and `make tools` from breaking again by running them as a blocking presubmit

### DIFF
--- a/config/jobs/kubernetes/enhancements/enhancements-presubmit.yaml
+++ b/config/jobs/kubernetes/enhancements/enhancements-presubmit.yaml
@@ -10,3 +10,13 @@ presubmits:
               - sh
               - "-c"
               - "export PATH=$PATH:$GOPATH/bin && ./hack/verify.sh"
+    - name: pull-enhancements-test
+      always_run: true
+      decorate: true
+      spec:
+        containers:
+          - image: golang:1.17
+            command:
+              - make
+              - test
+              - tools


config/jobs/kubernetes/enhancements/enhancements-presubmit.yaml
```diff
@@ -10,3 +10,13 @@ presubmits:
               - sh
               - "-c"
               - "export PATH=$PATH:$GOPATH/bin && ./hack/verify.sh"
+    - name: pull-enhancements-test
+      always_run: true
+      decorate: true
+      spec:
+        containers:
+          - image: golang:1.17
```
spiffxp, line 18: This ends up running one of the same tests that verify does.  Why not bump the verify job to the same golang image? To reduce blast radius.

Ideally the verify job would be `make verify` but there are duplicate targets and not all of them pass.  So since fixing that requires kubernetes/enhancement changes, hold off on bumping the verify image.

justaugustus: Heh. I was just about to ask this :)

justaugustus (approved):

PR #23547, opened 2021-09-10
Revert "Generate a schema for the ProwJob crd (#19522)"

chaodaiG:
This reverts commit 914737a267460d4d22e88d4e490efbb59fdbdbbb.

Seems like server side apply didn't work:
https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-test-infra-deploy-prow/1436369391678656512


alvaroaleman (approved):

config/prow/cluster/BUILD.bazel
```diff
@@ -73,11 +73,7 @@ release(
     component("prow-canary-k8s-io", "managedcertificate"),
     component("prow-k8s-io", "managedcertificate"),
     component("prow-kubernetes-io", "managedcertificate"),
-    component(
-        "prowjob",
-        "customresourcedefinition",
-        args = ["--server-side=true"],
```
chaodaiG, line 79: This was an attempt to pass `--server-side=true` to `kubectl apply` for the generated prowjob CRD, since it's too large to be applied on client side. I thought this would work as mentioned in https://github.com/kubernetes/test-infra/pull/19522#issuecomment-908697041, but somehow it didn't work. The postsubmit job https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-test-infra-deploy-prow/1436369391678656512 failed with:
```
$ /usr/bin/kubectl --kubeconfig= --cluster=gke_k8s-prow_us-central1-f_prow --context= --user= apply -f -
The CustomResourceDefinition "prowjobs.prow.k8s.io" is invalid: 
* metadata.annotations: Too long: must have at most 262144 bytes
* spec.preserveUnknownFields: Invalid value: true: must be false in order to use defaults in the schema
```

Note that `--server-side=true` was not included in the command.

+ @fejta here in case you know how to achieve this

PR #23549, opened 2021-09-10
Move prow image push job to run on tainted high cpu nodes

chaodaiG:
This job is currently running on default nodes with 8 CPUs in total, I have inspected a running job and found out that it uses 8 CPUs at peak, which might be bounded by the VM. This also affects other jobs running on the same node, including prow components. Make it running on tainted node should help.

I have already created the nodepool with 16 CPUs each and tainted with corresponding key:val

### DIFF
--- a/config/jobs/kubernetes/test-infra/test-infra-trusted.yaml
+++ b/config/jobs/kubernetes/test-infra/test-infra-trusted.yaml
@@ -203,6 +203,16 @@ postsubmits:
       - image: gcr.io/k8s-testimages/bazelbuild:v20210806-7bef894-test-infra
         command:
         - prow/push.sh
+      tolerations:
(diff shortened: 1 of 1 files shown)


cjwagner (approved):
I think this just allows the job to run on that node, but doesn't guarantee it. You may also want a `nodeSelector`. 
This would also be a good time to add a CPU resource request now that we have an estimate for what is needed.
/hold

chaodaiG:
> I think this just allows the job to run on that node, but doesn't guarantee it. You may also want a `nodeSelector`.
> This would also be a good time to add a CPU resource request now that we have an estimate for what is needed.
> /hold

good point, both added, and nodepool also labelled accordingly

/unhold

fejta (approved):
/hold

PR #23566, opened 2021-09-11
Delete .gitignore

Nancylsell:
Delete .gitignore


alvaroaleman:
This doesn't seem like a valid change. Please respect the maintainers time and do not create what appears to be a spam PR.
/close

PR #23568, opened 2021-09-12
Docs

listx:
Just some misc documentation updates. Also, renamed some variables in ghproxy coalescer for better readability.

The commits are split up to take care of each renaming operation, so have a look at those to double-check that this PR Is a NOP.

No logical change.

### DIFF
--- a/ghproxy/ghcache/coalesce.go
+++ b/ghproxy/ghcache/coalesce.go
@@ -45,7 +45,7 @@ type requestCoalescer struct {
 type firstRequest struct {
 	*sync.Cond
 
-	waiting bool
+	subscribers bool
 	resp    []byte
 	err     error
 }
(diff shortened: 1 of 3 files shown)


listx:
@cjwagner PTAL

cjwagner (approved):
Looks like a noop to me :+1: 
Personally I don't find all these renames to be better. For example `delegate` and `waiting` seemed more appropriate to me. Its just naming and personal preference though so if this is clearer to you this is fine with me!
/shrug

Also for the work in progress (WIP) plugin, it reacts to the PR title so include a `[WIP]` prefix or similar in the title to have the label apply automatically. You can find details about how to use plugins and how they are currently configured at the Prow instance's `/plugins` and `/command-help` endpoints. e.g. https://prow.k8s.io/plugins

listx:
> Personally I don't find all these renames to be better.

It would be great if we could either reuse the "promise/resolution" language in https://github.com/kubernetes/test-infra/blob/6d315feac15d352d4d5e80a5528014cac3ee8ead/prow/cache/cache.go#L60, or perhaps even better, reuse that data structure. But I think this PR is a step in the right direction for now.

> For example delegate and waiting seemed more appropriate to me. Its just naming and personal preference though so if this is clearer to you this is fine with me!

I didn't like `delegate` because it can act as both a verb and a noun, which may be confusing for people who are not native English language speakers.

PR #23573, opened 2021-09-13
ghproxy unit tests: avoid using time.Sleep for synchronization of goroutines

listx:
This builds on #23568.

I made the changes in this PR because I needed to closely understand how/why ghproxy's request coalescer worked, in order to implement it for #23500. I wanted to get to the bottom of why we were using time.Sleep() in the unit tests, and realized that it was unnecessary (at least the way they were being used). There are a large amount of test helper functions that allow us to get rid of time.Sleep() calls, but on the other hand we can now run the tests hundreds of times in less time than before, which gives us high confidence.

There are only a handful (3 to be exact) changes to the underlying business logic in coalesce.go, which seem to have been overlooked previously and should be NOP changes.

Anyway, using the knowledge I have learned here, I will implement similar unit tests for #23500, without relying on time.Sleep() for synchronization.

/assign @cjwagner

### DIFF
--- a/ghproxy/ghcache/coalesce.go
+++ b/ghproxy/ghcache/coalesce.go
@@ -44,10 +44,11 @@ type requestCoalescer struct {
 	sync.Mutex
 	cache map[string]*firstRequest
 
-	// requestExecutor is anything that can resolve a request by executing a single
-	// HTTP transaction, returning a Response for the provided Request. Using an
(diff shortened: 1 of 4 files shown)


listx:
/close

In retrospect I don't think it's worth the trouble to merge this PR, at least in its current form.

PR #23601, opened 2021-09-14
Defer pod deletion until after ProwJob status updates to prevent pods from being replaced after deletion.

cjwagner:
We are seeing pods schedule and pass after Prow deletes them and marks the job as error due to the unscheduled timeout. Here is an example flow:
> Create pod
Transition triggered -> pending
... 10 min delay ...
Deleted stale running pod
Pod is missing, starting a new pod
Transition pending -> error

It looks like at the time we delete the pod, the ProwJob resource has been marked complete in memory, but hasn't yet been patched to the API server. I think this causes a separate reconciliation thread to immediately react to the pod deletion before the ProwJob resource is patched to be marked complete. As a result a new pod is incorrectly created to replace the "missing" pod.

This race is resolved by deferring the pod deletion until after the ProwJob is patched.
/assign @alvaroaleman
/cc @chaodaiG @mpherman2

### DIFF
--- a/prow/plank/reconciler.go
+++ b/prow/plank/reconciler.go
@@ -521,6 +521,23 @@ func (r *reconciler) syncPendingJob(ctx context.Context, pj *prowv1.ProwJob) (*r
 		return nil, fmt.Errorf("patching prowjob: %w", err)
 	}
 
+	// If the ProwJob state has changed, we must ensure that the update reaches the cache before
+	// processing the key again. Without this we might accidentally replace intentionally deleted pods
+	// or otherwise incorrectly react to stale ProwJob state.
+	state := pj.Status.State
+	if prevPJ.Status.State == state {
+		return nil, nil
+	}
+	nn := types.NamespacedName{Namespace: pj.Namespace, Name: pj.Name}
+	if err := wait.Poll(100*time.Millisecond, 2*time.Second, func() (bool, error) {
+		if err := r.pjClient.Get(ctx, nn, pj); err != nil {
+			return false, fmt.Errorf("failed to get prowjob: %w", err)
+		}
+		return pj.Status.State == state, nil
+	}); err != nil {
+		return nil, fmt.Errorf("failed to wait for cached prowjob %s to get into state %s: %w", nn.String(), state, err)
+	}
+
 	return nil, nil
 }
 


alvaroaleman:
> I think this causes a separate reconciliation thread to immediately react to the pod deletion before the ProwJob resource is patched to be marked complete

Reconciliation is locked per key, there can only be one at a time. I would assume this happens because:
* We delete the timed out pod
* We patch and end reconciliation
* We get an event for the deleted pod, don't have the patched prowjob in our cache yet and re-create the pod

Which would mean that we need to block after the patch statement until we can retrieve the patched prowjob like we do e.G. when creating a pod: https://github.com/kubernetes/test-infra/blob/2b1842bf54cad06e7733bf7e82aba9b74daf2e89/prow/plank/reconciler.go#L702-L703

cjwagner:
@alvaroaleman Ah ok. It is unfortunate that we have to explicitly wait for cache freshness. It would be nice if the reconciler would automatically wait to reprocess the key until the cache is fresh.

Thanks for pointing this out, I've added a polling loop like the one used in `syncTriggeredJobs` to the end of `syncPendingJobs`. Presumably the 2 second timeout is appropriate here too.

alvaroaleman:
>  It would be nice if the reconciler would automatically wait to reprocess the key until the cache is fresh.

It can unfortunately not tell when the cash is fresh enough.

Can we remove the first commit? Right now if the pod deletion fails, we won't retry, just log an error.

cjwagner:
> Can we remove the first commit?

I don't think so? Don't we still need to wait to delete the pod until after we wait for the cache to update to include the correct ProwJob state?
I don't see any explicit retry logic. Are you saying this is changing the retry behavior due to logging the error rather than returning it? If so I can change the logic to avoid deferring the pod deletion and instead just do the deletion at the end before returning if needed.

alvaroaleman:
> I don't think so? Don't we still need to wait to delete the pod until after we wait for the cache to update to include the correct ProwJob state?

Not really, because that is the wrong controlflow, we shouldn't update the ProwJob because of a deleted pod when we didn't actually delete the pod. If the pod delete fails, we should return right there and not do anything with the prowjob.

> I don't see any explicit retry logic. Are you saying this is changing the retry behavior due to logging the error rather than returning it?

Yes, returning an error in the reconciler results in a retry after an exponential backoff.

cjwagner:
> Not really, because that is the wrong controlflow, we shouldn't update the ProwJob because of a deleted pod when we didn't actually delete the pod. If the pod delete fails, we should return right there and not do anything with the prowjob.

Makes sense, I got confused.
I removed the first commit.

alvaroaleman (approved):

PR #23595, opened 2021-09-14
prow/cmd: Add all components to README

spiffxp:
Related:
- Fixes: https://github.com/kubernetes/test-infra/issues/23594

I took my best guess based on the README.md inside of individual
component directories, or left under a "TODO: undocumented" heading for
those components that had none.

I took an educated guess and put two components under a "Deprecated"
heading, and suspect they could be removed in a later PR

PR #23613, opened 2021-09-15
prow/cmd: remove deprecated autobump cmd

spiffxp:
Related:
- Related to: https://github.com/kubernetes/k8s.io/issues/1523
- Followup to: https://github.com/kubernetes/test-infra/pull/23595

Rather than migrate the `gcloud-go` container to pull from `gcr.io/k8s-staging-test-infra` I would like to suggest we remove `autobump` as the sole thing that depends on it


alvaroaleman (approved):

spiffxp:
Missed that it was part of the prow release, removed.  Also updated prow docs to reference generic-autobumper

alvaroaleman (approved):

PR #23617, opened 2021-09-15
images: remove unused images

spiffxp:
Related:

Determined as part of a survey of gcr.io/k8s-testimages and cs.k8s.io that neither of these images are used


ameukam:
/lgtm
/hold
Possible impact on contributors outside of the k8s community ?

spiffxp:
I'll send an announcement to kubernetes-sig-testing and kubernetes-dev for these, stating that they'll no longer be updated

AFAICT everything showing up in github search is from forks of test-infra, or old forks of kubernetes/kubernetes that still have bazel files

spiffxp:
/hold cancel
Announcement: https://groups.google.com/g/kubernetes-dev/c/PhirMQ9KIo0

PR #23620, opened 2021-09-15
Prow/sub: report when prowjob is successfully triggered

chaodaiG:
Currently sub only reports back when it failed to trigger, a user wouldn't know how long they should poll the pubsub topic before deciding there is no error coming in, which is not a good experience


prow/pubsub/subscriber/subscriber.go
```diff
@@ -415,6 +425,10 @@ func (s *Subscriber) handleProwJob(l *logrus.Entry, jh jobHandler, msg messageIn
 		reportProwJobFailure(&prowJob, err)
 		return err
 	}
-	l.Infof("Job %q created as %q", pe.Name, prowJob.Name)
+	l.WithFields(logrus.Fields{
+		"event-name": pe.Name,
+		"job-name":   prowJob.Name,
```
cjwagner, line 430: IIRC these should be `job` and `name` respectively to be consistent with how we filter on these elsewhere.

prow/pubsub/subscriber/subscriber.go
```diff
@@ -351,16 +351,26 @@ func (s *Subscriber) handleProwJob(l *logrus.Entry, jh jobHandler, msg messageIn
 		return err
 	}
 
-	reportProwJobFailure := func(pj *prowapi.ProwJob, err error) {
-		pj.Status.State = prowapi.ErrorState
-		pj.Status.Description = err.Error()
+	reportProwJob := func(pj *prowapi.ProwJob, state v1.ProwJobState, err error) {
+		pj.Status.State = state
+		if err != nil {
+			pj.Status.Description = err.Error()
```
cjwagner, line 357: We should have a status description in non-error cases as well.

cjwagner (approved):
Just some nits.
/hold

prow/pubsub/subscriber/subscriber.go
chaodaiG: fixed

prow/pubsub/subscriber/subscriber.go
chaodaiG: good idea, done

prow/pubsub/subscriber/subscriber.go
```diff
@@ -415,6 +426,10 @@ func (s *Subscriber) handleProwJob(l *logrus.Entry, jh jobHandler, msg messageIn
 		reportProwJobFailure(&prowJob, err)
 		return err
 	}
-	l.Infof("Job %q created as %q", pe.Name, prowJob.Name)
+	l.WithFields(logrus.Fields{
+		"name": pe.Name,
+		"job":  prowJob.Name,
```
cjwagner, line 431: I think these are backwards.  Name should be the ProwJob resource name (a GUID) and job should be the job name defined in config.

chaodaiG: got it, this is confusing tbh

cjwagner: Yeah I agree. It would have been clearer if `pe.Name` was `pe.Job`, but that ship has sailed. :man_shrugging:

cjwagner (approved):

Issue #23407, opened 2021-08-30
GH cache should prune its storage to avoid hitting limits


petr-muller:
**What would you like to be added**:

`ghcache` should provide option(s) to delete old / unneeded files from its storage to avoid accumulating too many files / too large files. The exact mechanism about what to specify (storage limit / file count limit / file age limit) does not matter too much, IMO: it is design space.

**Why is this needed**:

In our `ghproxy` deployment, we need to routinely prune the cache with `find /cache/data /cache/temp -mtime +1 -type f -delete` to avoid exhausting inodes.

PR #23621, opened 2021-09-15
GHProxy: Cleanup old caches

alvaroaleman:
Currently, Ghproxy never cleans up caches. This can relatively quickly
lead to inode exhaustion when apps auth is used, as it results in many,
relatively shortlived caches (1h).

This change adds pruning for those which works as follows:
* The github client will add an expiry header if it sends a request with
  a token that expires
* Ghproxy will write the expiry time into a metadata file at the root of
  the cache partition
* A background routine in ghproxy will iterate over all cache partitions
  and delete them when they have expired

Fixes https://github.com/kubernetes/test-infra/issues/23407

### DIFF
--- a/ghproxy/ghcache/ghcache.go
+++ b/ghproxy/ghcache/ghcache.go
@@ -27,11 +27,19 @@ package ghcache
 
 import (
 	"context"
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
 	"net/http"
+	"os"
 	"path"
+	"path/filepath"
 	"strconv"
 	"strings"
 	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 
 	"github.com/gomodule/redigo/redis"
 	"github.com/gregjones/httpcache"
@@ -66,6 +74,10 @@ const (
 	// which metrics should be recorded if set. If unset, the sha256sum of
 	// the Authorization header will be used.
 	TokenBudgetIdentifierHeader = "X-PROW-GHCACHE-TOKEN-BUDGET-IDENTIFIER"
+
+	// TokenExpiryAtHeader includes a date at which the passed token expires and all associated caches
+	// can be cleaned up. It's value must be in RFC3339 format.
+	TokenExpiryAtHeader = "X-PROW-TOKEN-EXPIRES-AT"
 )
 
 func CacheModeIsFree(mode CacheResponseMode) bool {
@@ -214,7 +226,7 @@ const LogMessageWithDiskPartitionFields = "Not using a partitioned cache because
 // NewDiskCache creates a GitHub cache RoundTripper that is backed by a disk
 // cache.
 // It supports a partitioned cache.
-func NewDiskCache(delegate http.RoundTripper, cacheDir string, cacheSizeGB, maxConcurrency int, legacyDisablePartitioningByAuthHeader bool) http.RoundTripper {
+func NewDiskCache(delegate http.RoundTripper, cacheDir string, cacheSizeGB, maxConcurrency int, legacyDisablePartitioningByAuthHeader bool, cachePruneInterval time.Duration) http.RoundTripper {
 	if legacyDisablePartitioningByAuthHeader {
 		diskCache := diskcache.NewWithDiskv(
 			diskv.New(diskv.Options{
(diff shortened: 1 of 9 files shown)


ghproxy/ghcache/ghcache.go
```diff
@@ -232,37 +244,116 @@ func NewDiskCache(delegate http.RoundTripper, cacheDir string, cacheSizeGB, maxC
 			maxConcurrency,
 		)
 	}
+
+	go func() {
+		for range time.NewTicker(cachePruneInterval).C {
```
stevekuznetsov, line 249: iirc there's some gotcha with this construct that leaks tickers?

stevekuznetsov: if `ghproxy` uses `interrupts`, prefer `interrupts.Tick()`

ghproxy/ghcache/ghcache.go
```diff
@@ -232,37 +244,116 @@ func NewDiskCache(delegate http.RoundTripper, cacheDir string, cacheSizeGB, maxC
...
+			logrus.WithError(err).Warn("os.ReadDir failed")
+			// no continue, os.ReadDir returns partial results if it encounters an error
+		}
+		for _, cachePartitionCandidate := range cachePartitionCandidates {
+			if !cachePartitionCandidate.IsDir() {
+				continue
+			}
+			metadataPath := path.Join(base, cachePartitionCandidate.Name(), cachePartitionMetadataFileName)
+
+			// Read optimistically and return errors
```
stevekuznetsov, line 287: What do you mean by "return errors" ?

ghproxy/ghcache/ghcache.go
```diff
@@ -232,37 +244,116 @@ func NewDiskCache(delegate http.RoundTripper, cacheDir string, cacheSizeGB, maxC
...
+			if err := os.RemoveAll(paritionPath); err != nil {
+				logrus.WithError(err).WithField("path", paritionPath).Error("failed to delete expired cache parition")
+			}
+		}
+	}
+}
+
+func writecachePartitionMetadata(basePath, tempDir string, expiresAt time.Time) error {
+	if expiresAt.IsZero() {
+		return nil
```
stevekuznetsov, line 311: Won't this lead to leaks? Why not failsafe to writing metadata that expires at `time.Now().Add(time.Hour)` or something?

alvaroaleman: No, it won't and we can't do that. The whole reason the expiry information is passed on from the client is tokens validity varies and in the case of PAT, it never expires which will gets by an empty `expiresAt`. If we unconditionally added 1h here, we would always delete all caches after one hour, we cant do that.

ghproxy/ghcache/ghcache.go
alvaroaleman: Should have been "ignore errors", fixed.

ghproxy/ghcache/ghcache.go
stevekuznetsov: Ah, I see - this is to handle PAT. Perhaps a comment would be good to clarify this since it's implicit behavior?

stevekuznetsov: I guess I would have expected "no expiry header" -> "no call to writing metadata" rather than "no expiry header" -> "pass an invalid date" -> "do nothing" but maybe that's just me being confused by it all

stevekuznetsov: Also I guess it would not hurt to have a default cache TTL for PAT entries, too, since they could hit the same issues that apps auth hits, on a smaller machine or with fewer inodes free? Setting the TTL to a week or something should not cause adverse effects.

alvaroaleman: Added a comment and made it a pointer to further clarify this might not be set. The TTL is for the entire cache, not individual entries so we can never evict a PAT cache

spiffxp:
I just caught a flake over in https://github.com/kubernetes/test-infra/pull/23656#issuecomment-922063719 that I suspect is related to this, only because I've never seen ghproxy flake on me before and this just landed recently

alvaroaleman:
> I just caught a flake over in #23656 (comment) that I suspect is related to this, only because I've never seen ghproxy flake on me before and this just landed recently

Thanks. I suspect that is because the test relies on timings and if things to too slow, it will fail like this. I'll try to improve this through a fake clock.

PR #23629, opened 2021-09-16
[Bugzilla] Allow the bugzilla client to use a Bearer Authorization token

droslean:

Some of the Bugzilla instances support only Bearer Authorization headers. This change only adds the extra header to be used.

NOTE: I have tested the co-existence of the `X-BUGZILLA-API-KEY` and `Authorization header and it seems ok.

### DIFF
--- a/prow/bugzilla/client.go
+++ b/prow/bugzilla/client.go
@@ -677,8 +677,10 @@ func (c *client) GetComments(bugID int) ([]Comment, error) {
 func (c *client) request(req *http.Request, logger *logrus.Entry) ([]byte, error) {
 	if apiKey := c.getAPIKey(); len(apiKey) > 0 {
 		// some BugZilla servers are too old and can't handle the header.
-		// some don't want the query parameter. We can set both and keep
+		// some don't want the query parameter and some need Bearer Authorization instead.
+		// We can set all of them and keep
 		// everyone happy without negotiating on versions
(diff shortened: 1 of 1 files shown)


prow/bugzilla/client.go
```diff
@@ -676,13 +678,18 @@ func (c *client) GetComments(bugID int) ([]Comment, error) {
...
 func (c *client) request(req *http.Request, logger *logrus.Entry) ([]byte, error) {
 	if apiKey := c.getAPIKey(); len(apiKey) > 0 {
-		// some BugZilla servers are too old and can't handle the header.
-		// some don't want the query parameter. We can set both and keep
-		// everyone happy without negotiating on versions
-		req.Header.Set("X-BUGZILLA-API-KEY", string(apiKey))
-		values := req.URL.Query()
-		values.Add("api_key", string(apiKey))
-		req.URL.RawQuery = values.Encode()
+		if c.useBearerAuthorization {
```
alvaroaleman, line 681: Why not always set both and avoid the need for a flag?

droslean: I wasn't sure if that will break the call. If it's safe for those headers to co-exists, I am all in to remove the flag.

alvaroaleman: Well, try it out

droslean: updated.

alvaroaleman (approved):

PR #23630, opened 2021-09-16
Update getting started docs to suggest using an app instead of token

alvaroaleman:
This updates our getting started docs and examples to use a GitHub app,
as that simplifies management, scales with installations, makes it more
obvious that Prow is an automation and is the suggested way to build
automation.

Fixes https://github.com/kubernetes/test-infra/issues/10423

/assign @chaodaiG @BenTheElder @spiffxp

### DIFF
--- a/prow/getting_started_deploy.md
+++ b/prow/getting_started_deploy.md
@@ -6,26 +6,37 @@ Prow runs in any kubernetes cluster. Our `tackle` utility helps deploy it correc
 
 Both of these are focused on [Kubernetes Engine](https://cloud.google.com/kubernetes-engine/) but should work on any kubernetes distro with no/minimal changes.
 
-## GitHub bot account
+## GitHub App
 
-Before using `tackle` or deploying prow manually, ensure you have created a
-GitHub account for prow to use.  Prow will ignore most GitHub events generated
-by this account, so it is important this account be separate from any users or
-automation you wish to interact with prow. For example, you still need to do
-this even if you'd just setting up a prow instance to work against your own
-personal repos.
+First, you need to create a GitHub app. GitHub itself [documents this.](https://docs.github.com/en/developers/apps/building-github-apps/creating-a-github-app)
+Initially, it is sufficient to set a dummy url for the Webhook.
+The exact set of permissions needed varies based on what functionality you use. Below is a minimum
+set of permissions needed. Please keep in mind that any changes to the permissions your
+app requests (both added and removed) require everyone to re-install it.
 
-1. Ensure the bot user has the following permissions
-    - Write access to the repos you plan on handling
-    - Owner access (and org membership) for the orgs you plan on handling (note
-      it is possible to handle specific repos in an org without this)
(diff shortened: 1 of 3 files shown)


prow/getting_started_deploy.md
```diff
@@ -202,49 +215,20 @@ to start receiving GitHub events!
 
 ## Add the webhook to GitHub
```
chaodaiG, line 216: Suggest change this to something like "Setup webhook for prow user"

prow/getting_started_deploy.md
```diff
@@ -202,49 +215,20 @@ to start receiving GitHub events!
...
-- Click `Add webhook`.
-
-### Use `hmac` tool to manage webhooks and hmac tokens (recommended)
+To set up the webhook, you have to go the the GitHub UI and edit your app. Update
+the `Webhook URL` property to `https://prow.<<yourdomain.com>>/hook`. Use the URL
+shown above when getting the `Ingress`.
 
-If you need to configure webhooks for multiple orgs or repos, the manual process does not work that well as it
-can be error-prone, and it'll be painful when you want to replace the hmac token if it is accidentally leaked.
+## Install app
```
chaodaiG, line 222: Suggest change this to something like "Configure webhook for a repo"

chaodaiG (commented):
Looks good to me, left a few nits

prow/getting_started_deploy.md
alvaroaleman: But this section isn't needed when using a GH app

chaodaiG: The new line 218 you added does mention set up webhook though

alvaroaleman: Well, it is needed once for the entire app and will be used everywhere it is installed. I guess the larger question behind this is if we should keep existing PAT-related docs.

chaodaiG: Agreed on the larger question. My initial thought was actually we want to keep it, but not very sure. 

However, if we do want to keep it, we probably should make it very clear that gh app is the recommended approach

prow/getting_started_deploy.md
alvaroaleman: Updated, WDYT?

chaodaiG: lgtm

prow/getting_started_deploy.md
alvaroaleman: Hm I tend to say to not keep it because existing users already know this and new users should be setting up an app. Furthermore we'd then also have to have another set of `getting_started_{s3,gcs},yaml` manifests that use PAT. WDYT?

chaodaiG: Makes sense to me. I personally don't use this doc that much as operator of a couple of prow instances, so yeah I agree that existing user already know what to do. Also the set of started yaml is a great point.

chaodaiG (approved):

PR #23634, opened 2021-09-16
Tell tide to ignore EasyCLA on k8s.io/org

cblecker:
We are testing the new EasyCLA in k8s.io/org. This tells tide to ignore that context in mergeability calculations.

/assign @mrbobbytables


mrbobbytables:
thanks!
/lgtm

PR #23643, opened 2021-09-16
prow/repoowners: Track filtered (non-collaborator) owners

wking:
GitHub-side docs for "collaborators" semantics are linked from:

```console
$ grep -hB5 'func .*ListCollaborators' prow/github/client.go
// ListCollaborators gets a list of all users who have access to a repo (and
// can become assignees or requested reviewers).
//
// See 'IsCollaborator' for more details.
// See https://developer.github.com/v3/repos/collaborators/
func (c *client) ListCollaborators(org, repo string) ([]User, error) {
```

Filtering non-collaborators out is good, and we do that already, but exposing an interface for stale entries will make it easier for consumers (like the `approve` plugin) to inform folks about stale content.

### DIFF
--- a/prow/plugins/approve/approvers/owners.go
+++ b/prow/plugins/approve/approvers/owners.go
@@ -48,6 +48,7 @@ type Repo interface {
 	IsNoParentOwners(path string) bool
 	IsAutoApproveUnownedSubfolders(directory string) bool
 	Filenames() ownersconfig.Filenames
+	StaleRepositoryState() string
 }
 
 // Owners provides functionality related to owners of a specific code change.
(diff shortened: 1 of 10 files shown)


wking:
[Bazel][1]:

```
--- FAIL: TestDiskCachePruning (0.07s)
    ghproxy_test.go:57: got a reequest for path /app
    ghproxy_test.go:57: got a reequest for path /app/installations
    ghproxy_test.go:57: got a reequest for path /app/installations/0/access_tokens
    ghproxy_test.go:57: got a reequest for path /repos/org/repo/git/refs/dev
    ghproxy_test.go:97: expected two cache paritions, one for the app and one for the app installation, got 3
    ghproxy_test.go:107: expected one cache partition for the app as the one for the installation should be cleaned up, got  2
FAIL
```

seems unrelated.

/retest

[1]: https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/test-infra/23643/pull-test-infra-bazel/1440031063853240320

prow/plugins/approve/approvers/owners.go
```diff
@@ -670,6 +672,12 @@ Associated issue requirement bypassed by:{{range $index, $approval := .ap.ListNo
 
 The full list of commands accepted by this bot can be found [here]({{ .commandHelpLink }}?repo={{ .org }}%2F{{ .repo }}).
 
+{{ if len .staleRepositoryState -}}
```
alvaroaleman, line 675: I don't think this should be part of the approval message. This is shown to everyone who creates a PR but most of them don't have anything to do with managing OWNERS files.

I would suggest to instead build a CLI that allows to find stale entries.

wking: there is an outside chance that this will help a random PR-author by warning them that the nominal OWNER they had pinged for review was actually no longer an active collaborator.  And really, I would hope that shortly after this list got any entries, the active maintainers would prune `OWNERS` and future approval messages would be clean.  So I still like this. 
 But since I expect neither of those arguments to convince anyone:
 
 /close
 
 I'll happily re-open if I'm wrong and there is renewed interest.

PR #23645, opened 2021-09-16
Change retest help message to indicate optional jobs

YuviGold:
Resolves https://github.com/kubernetes/test-infra/issues/23062
After https://github.com/kubernetes/test-infra/pull/23484 got reverted on https://github.com/kubernetes/test-infra/pull/23599

Add an indication of a failed test github message if it's optional or
not.
Developers that aren't familiar with every job definition, can find it convenient
when a bunch of tests fail, it is clear right away which jobs are mandatory.

Add `prow.k8s.io/is-optional` as a new label for presubmit jobs.
This way a Prowjob struct can indicate whether the job is optional or
not.

Later on, this label is verified on the github reporter and it would
create an entry comment accordingly presenting if a failed job is
required or not.

Please let me know if you find this solution appliable.
/cc @chaodaiG @cjwagner @alvaroaleman

### DIFF
--- a/prow/github/report/report.go
+++ b/prow/github/report/report.go
@@ -22,12 +22,14 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"strconv"
 	"strings"
 	"text/template"
 
 	prowapi "k8s.io/test-infra/prow/apis/prowjobs/v1"
 	"k8s.io/test-infra/prow/config"
 	"k8s.io/test-infra/prow/github"
+	"k8s.io/test-infra/prow/kube"
 	"k8s.io/test-infra/prow/plugins"
 )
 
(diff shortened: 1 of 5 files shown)


prow/kube/prowjob.go
```diff
@@ -63,4 +63,7 @@ const (
 	PullLabel = "prow.k8s.io/refs.pull"
 	// RetestLabel exposes if the job was created by a re-test request.
 	RetestLabel = "prow.k8s.io/retest"
+	// OptionalLabel is added in resources created by prow and
+	// carries the Optional from a Presubmit job.
+	OptionalLabel = "prow.k8s.io/optional"
```
chaodaiG, line 68: nit, `optional` might be useful for something more general later, can we do `is-optional` instead?

prow/github/report/report.go
```diff
@@ -258,10 +260,21 @@ func parseIssueComments(pj prowapi.ProwJob, isBot func(string) bool, ics []githu
 }
 
 func createEntry(pj prowapi.ProwJob) string {
+	required := strconv.FormatBool(true)
```
chaodaiG, line 263: Any job triggered prior to this change will not have this label, and will be labeled required. Should we default to `unknown` instead of true to be more accurate?

prow/kube/prowjob.go
YuviGold: done

prow/github/report/report.go
YuviGold: done

chaodaiG (approved):
Thank you!

PR #23649, opened 2021-09-17
Increase deletion throughput in 5k test

mborsz:
With https://github.com/kubernetes/kubernetes/pull/104966 we should be able to achieve ~60 pods/s deletion throughput.
The expected test duration saving is ~40m (out of ~5h).

I have started a manual run using this value and it looks good (metrics are OK, there is a lot of spare CPU during the object deletion phase).

We should consider increasing pod creation throughput to match this (there is a lot of spare CPU in that part as well).

/assign @wojtek-t

### DIFF
--- a/config/jobs/kubernetes/sig-scalability/sig-scalability-release-blocking-jobs.yaml
+++ b/config/jobs/kubernetes/sig-scalability/sig-scalability-release-blocking-jobs.yaml
@@ -103,7 +103,7 @@ periodics:
       - --provider=gce
       - --metadata-sources=cl2-metadata.json
       - --env=CL2_LOAD_TEST_THROUGHPUT=50
-      - --env=CL2_DELETE_TEST_THROUGHPUT=30
+      - --env=CL2_DELETE_TEST_THROUGHPUT=50
       - --env=CL2_ENABLE_HUGE_SERVICES=true
       # Overrides CONTROLLER_MANAGER_TEST_ARGS from preset-e2e-scalability-periodics.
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes/sig-scalability/sig-scalability-release-blocking-jobs.yaml
```diff
@@ -103,7 +103,7 @@ periodics:
       - --provider=gce
       - --metadata-sources=cl2-metadata.json
       - --env=CL2_LOAD_TEST_THROUGHPUT=50
-      - --env=CL2_DELETE_TEST_THROUGHPUT=30
+      - --env=CL2_DELETE_TEST_THROUGHPUT=60
```
wojtek-t, line 106: How about aligning it for the moment with the creation throughput and setting to 50?

And potentially increasing both further in the subsequent change?

mborsz: I was thinking about this for a while. I don't have a strong opinion on this. I decided to use 60 as it's faster, we know that it will work (this is the value I used for manual test) and ensures stronger property we can use later internally.

wojtek-t: If you don't feel very strongly, I personally prefer consistency and setting it to 50.
Yes - it will be slightly slower, but the difference is not that significant. If we really want to speed it up, let's do that in a followup and bump both values.

mborsz: Agreed. Changed to 50.

Issue #23614, opened 2021-09-15
Make Crier's GitHub reporting optionally report only a single failure summary comment after all jobs complete. 


cjwagner:
**What would you like to be added**:
Prow should support reporting only a single presubmit failure summary comment to PRs once all jobs complete rather than continually updating the comment as jobs complete. This is the behavior of the Gerrit reporter so there may be some existing code there that can be leveraged.

**Why is this needed**:
It seems that maintaining this comment is expensive. In particular we are spending most of our GitHub API rate limit on one of our Prow instances for Crier to list the issue comments on PRs.

Some users have also indicated that it would be preferable to only receive a summary comment once all jobs have completed.

cc @chaodaiG @mpherman2 @listx

PR #23652, opened 2021-09-17
Crier supports a new comment mode: summary

chaodaiG:
Fixes https://github.com/kubernetes/test-infra/issues/23614


chaodaiG:
/cc @cjwagner @alvaroaleman 

Creating this PR with basic logic here first, please take a look. Will add unit tests if this direction looks good to you

2nd-half:
Any concern on this change? This would be very useful for us. Thanks!

chaodaiG:
@cjwagner or @alvaroaleman ^^

prow/crier/reporters/github/reporter.go
```diff
@@ -162,8 +165,61 @@ func (c *Client) Report(ctx context.Context, log *logrus.Entry, pj *v1.ProwJob)
 		defer lock.Release(1)
 	}
 
+	pjsToReport := func() []v1.ProwJob {
+		if len(pj.Spec.Refs.Pulls) != 1 {
+			return nil
+		}
+		// find all prowjobs from this PR
+		selector := map[string]string{}
+		for _, l := range []string{kube.OrgLabel, kube.RepoLabel, kube.PullLabel} {
```
alvaroaleman, line 174: We can already filter for presubmit jobs here

prow/crier/reporters/github/reporter.go
```diff
@@ -162,8 +165,61 @@ func (c *Client) Report(ctx context.Context, log *logrus.Entry, pj *v1.ProwJob)
...
+			// continue convince me that you are the last one of your kind
+			if existing, ok := latestBatch[pjob.Spec.Job]; !ok {
+				latestBatch[pjob.Spec.Job] = pjob
+			} else if pjob.CreationTimestamp.After(existing.CreationTimestamp.Time) {
+				latestBatch[pjob.Spec.Job] = pjob
+			}
+		}
+
+		// If current pj is not the latest, we should skip
+		if existing, ok := latestBatch[pj.Spec.Job]; !ok {
```
alvaroaleman, line 201: We need the latest job for the pull request, not the latest job for a given job or not?

prow/crier/reporters/github/reporter.go
```diff
@@ -162,8 +165,61 @@ func (c *Client) Report(ctx context.Context, log *logrus.Entry, pj *v1.ProwJob)
...
+		}
+
+		latestBatch := make(map[string]v1.ProwJob)
+		for _, pjob := range pjs.Items {
+			if len(pjob.Spec.Refs.Pulls) != 1 {
+				continue
+			}
+			if pj.Spec.Refs.Pulls[0].SHA != pjob.Spec.Refs.Pulls[0].SHA {
+				continue
+			}
```
alvaroaleman, line 190: I think we should also filter them through `c.ShouldReport()`

prow/crier/reporters/github/reporter.go
```diff
@@ -174,7 +230,26 @@ func (c *Client) Report(ctx context.Context, log *logrus.Entry, pj *v1.ProwJob)
...
 			err = nil
 		}
+		// Always return when there is any error reporting status context.
+		return []*v1.ProwJob{pj}, nil, err
+	}
+
+	// Check if this org or repo has opted out of failure report comments
+	refs := pj.Spec.Refs
+	fullRepo := fmt.Sprintf("%s/%s", refs.Org, refs.Repo)
+	for _, ident := range c.config().GitHubReporter.NoCommentRepos {
```
alvaroaleman, line 240: Please add a comment that mentions this check has to be here and not in `ShouldReport` because we always need to report the status context, just potentially not create a comment

prow/crier/reporters/github/reporter.go
```diff
@@ -162,8 +165,61 @@ func (c *Client) Report(ctx context.Context, log *logrus.Entry, pj *v1.ProwJob)
 		defer lock.Release(1)
 	}
 
+	pjsToReport := func() []v1.ProwJob {
```
alvaroaleman, line 168: This is a stylistic nit but I am neither a huge fan of inline functions nor of defining a function far away from its only callsite. There isn't really a need for this as the only inputn this needs is the prowjob or did I miss something?

prow/crier/reporters/github/reporter.go
```diff
@@ -162,8 +165,61 @@ func (c *Client) Report(ctx context.Context, log *logrus.Entry, pj *v1.ProwJob)
...
+			log.WithFields(logrus.Fields{"name": pj.Spec.Job, "job": pj.Name}).Error("Prowjob to be reported was not found")
+			return nil
+		} else if existing.Name != pj.Name {
+			return nil
+		}
+
+		// Check all jobs to make sure they are completed
+		var toReport []v1.ProwJob
+		for _, pjob := range latestBatch {
+			if !pjob.Complete() {
```
alvaroaleman, line 212: Small performance nit, but we can do this check and the return already above when iterating over the jobs we got from the cache and save ourselves a bunch of map entries and checking or not?

alvaroaleman:
On a conceptional level, both the github commenting and gerrit would be better served with a controller that enqueues PRs rather than prowjobs (CR in gerrits case). Not a big deal, but maybe something we could do in the future

prow/crier/reporters/github/reporter.go
chaodaiG: isn't `kube.PullLabel` sufficient for this check?

prow/crier/reporters/github/reporter.go
chaodaiG: this is checking for an edge condition that might only exists in my head :)

say that job `foo` with number N is about to finish, and the developer saw error log in prow UI ahead of time, and know it was a flake, commented `/test foo` and triggered a new prowjob numbered M, which was not fast enough to cancel N before it's finished. In this case crier saw both M and N, and we should not report. And actually while I'm writing this, noticed that this case is actually handled by the check later, so I think this can be deleted

alvaroaleman: Ah ok, yeah generally this can always race with job creation.

What I meant is that we should only report the oldest job to prevent the reporter from doing the comment reporting for every job once all jobs finished.

prow/crier/reporters/github/reporter.go
alvaroaleman: Oh you are right

prow/crier/reporters/github/reporter.go
chaodaiG: the idea here is to report when all running jobs are finished, did you mean latest instead of oldest?

alvaroaleman: yeah, but what I mean is that if e.G. we have seven ProwJobs, we will end up calling `report.ReportComment` seven times once all of them finished or did I miss something? The idea was to abort if we do not have the oldest pj here to only do it once.

prow/crier/reporters/github/reporter.go
```diff
@@ -162,8 +165,61 @@ func (c *Client) Report(ctx context.Context, log *logrus.Entry, pj *v1.ProwJob)
...
+		}
+		// find all prowjobs from this PR
+		selector := map[string]string{}
+		for _, l := range []string{kube.OrgLabel, kube.RepoLabel, kube.PullLabel} {
+			selector[l] = pj.ObjectMeta.Labels[l]
+		}
+		var pjs v1.ProwJobList
+		if err := c.lister.List(ctx, &pjs, ctrlruntimeclient.MatchingLabels(selector)); err != nil {
+			log.WithError(err).Errorf("Cannot list prowjob with selector %v", selector)
+			return nil
```
alvaroaleman, line 180: Why do we not return that error?

prow/crier/reporters/github/reporter.go
chaodaiG: Moved up

prow/crier/reporters/github/reporter.go
chaodaiG: good idea, moved it out, will add unit test later

prow/crier/reporters/github/reporter.go
chaodaiG: The first six finished jobs will be labeled as reported by github reporter, and won't be reconciled any more unless I missed something? So in this case `report.ReportComment` will only be called on the seventh job completion. It would only be called seven times if all seven jobs completed all at the same time, which should be rare?

prow/crier/reporters/github/reporter.go
chaodaiG: good call, should fail here, updated

prow/crier/reporters/github/reporter.go
```diff
@@ -174,11 +177,71 @@ func (c *Client) Report(ctx context.Context, log *logrus.Entry, pj *v1.ProwJob)
...
+	var pjs v1.ProwJobList
+	if err := lister.List(ctx, &pjs, ctrlruntimeclient.MatchingLabels(selector)); err != nil {
+		return nil, fmt.Errorf("Cannot list prowjob with selector %v", selector)
+	}
+
+	latestBatch := make(map[string]v1.ProwJob)
+	for _, pjob := range pjs.Items {
+		if !pjob.Complete() { // Any job still running should prevent from comments
+			return nil, nil
+		}
```
cjwagner, line 227: We need to be ignoring non-reporting jobs. Maybe filter them out immediately after listing? Please also add a test case for this.

prow/github/report/report.go
```diff
@@ -201,7 +224,7 @@ func parseIssueComments(pj prowapi.ProwJob, isBot func(string) bool, ics []githu
 		}
 		// Old report comments started with the context. Delete them.
 		// TODO(spxtr): Delete this check a few weeks after this merges.
```
cjwagner, line 226: We'd need to check if the comment contained any context, not just the one from the first PR.

That being said, this comment is almost 5 years old. We can just remove the check rather than continuing to update it. Any existing Prow instances that have not been updated in 5 years can deal with some stray comments if they ever decide to update :upside_down_face:

prow/github/report/report.go
```diff
@@ -282,19 +311,21 @@ func createEntry(pj prowapi.ProwJob) string {
...
 	if len(entries) > 1 {
 		plural = "s"
 	}
 	var b bytes.Buffer
-	if reportTemplate != nil {
-		if err := reportTemplate.Execute(&b, &pj); err != nil {
+	// Templating was default to report a single prowjob, which doesn't apply when
+	// multiple prowjobs are expected to be reported.
+	if reportTemplate != nil && len(pjs) == 1 {
+		if err := reportTemplate.Execute(&b, &pjs[0]); err != nil {
```
cjwagner, line 323: FWIW the report template is usually related to the PR not a specific PJ, even though it is using the PJ in the template. This is kind of unfortunate and doesn't really make sense given that we maintain one failure comment on PRs, not one per PJ.
So we might be better off using the first PJ and still executing the template even if there are multiple PJs. I'm not sure what is better here so I'll leave this to your discretion. Either way this is something we can change later if needed. 
Here is an example reportTemplate for reference:
> [Full PR test history](https://prow.k8s.io/pr-history?org=kubernetes&repo=test-infra&pr=23693). [Your PR dashboard](https://prow.k8s.io/pr?query=is%3Apr%20state%3Aopen%20author%3AchaodaiG). Please help us cut down on flakes by [linking to](https://git.k8s.io/community/contributors/devel/sig-testing/flaky-tests.md#filing-issues-for-flaky-tests) an [open issue](https://github.com/kubernetes/test-infra/issues?q=is:issue+is:open) when you hit one in your PR.

prow/crier/reporters/github/reporter.go
```diff
@@ -46,6 +48,7 @@ type Client struct {
 	config      config.Getter
 	reportAgent v1.ProwJobAgent
 	prLocks     *shardedLock
+	lister      ctrlruntimeclient.Reader
```
cjwagner, line 51: It doesn't look like this is ever initialized.

prow/github/report/report_test.go
```diff
@@ -205,7 +205,7 @@ func TestParseIssueComment(t *testing.T) {
 			isBot := func(candidate string) bool {
 				return candidate == "k8s-ci-robot"
 			}
-			deletes, entries, update := parseIssueComments(pj, isBot, tc.ics)
+			deletes, entries, update := parseIssueComments([]prowapi.ProwJob{pj}, isBot, tc.ics)
```
cjwagner, line 208: It would be good to test the new functionality for reporting multiple PJs at once. Please add unit tests for this at some level.

prow/github/report/report.go
chaodaiG: deleted

prow/github/report/report.go
chaodaiG: good point, shamelessly copy/pasted your comment as comment in the code, will go ahead and use the first PR as args for the template. Unit tests will be added

prow/github/report/report_test.go
chaodaiG: will do

chaodaiG (commented):
Pushing changes from review comments, this will break some tests, will fix tests and add more unit tests

chaodaiG:
unit test added, @cjwagner please take another look

cjwagner (approved):
LGTM

PR #23664, opened 2021-09-19
Add config/jobs/image-pushing/k8s-staging-public-pii.yaml

BobyMCbobs:
Builds the _public-log-vendor-matcher_ image used for generating data for reviewing Kubernetes Public artifact traffic

Depends on https://github.com/kubernetes/k8s.io/pull/2710.

PR #23665, opened 2021-09-20
Add public log vendor matcher periodic job

BobyMCbobs:
Adds a Prow periodic job for running the public-log-vendor-matcher data generation pipeline

Depends on https://github.com/kubernetes/test-infra/pull/23664 and https://github.com/kubernetes/k8s.io/pull/2710

### DIFF
--- a/config/jobs/kubernetes/k8s-infra-public-pii/k8s-infra-public-pii.yaml
+++ b/config/jobs/kubernetes/k8s-infra-public-pii/k8s-infra-public-pii.yaml
@@ -1,0 +1,29 @@
+periodics:
+  - interval: 7d
+    agent: kubernetes
+    name: public-log-vendor-matcher
+    decorate: true
+    spec:
+      containers:
+        - name: public-log-vendor-matcher
+          image: public-log-vendor-matcher #something
+          command:
+            - /usr/local/bin/docker-entrypoint.sh
+            - postgres
+          imagePullPolicy: Always
+          volumeMounts:
+            - name: gcp-app-creds
+              mountPath: /etc/asn-etl-pipeline
+          env:
+            - name: GCP_PROJECT
+              value: k8s-infra-ii-sandbox
+            - name: GCP_SERVICEACCOUNT
+              value: asn-etl@k8s-infra-ii-sandbox.iam.gserviceaccount.com
+            - name: GCP_BIGQUERY_DATASET
+              value: etl_script_generated_set
+            - name: GOOGLE_APPLICATION_CREDENTIALS
+              value: /etc/asn-etl-pipeline/asn-etl-pipeline-gcp-sa.json
+      volumes:
+        - name: gcp-app-creds
+          secret:
+            secretName: gcp-app-creds


BobyMCbobs:
The [_pull-test-infra-bazel_](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/test-infra/23665/pull-test-infra-bazel/1439743221289717760#1:build-log.txt%3A6605) job says that there needs to be a dashboard made, I think?

chases2:
> The [_pull-test-infra-bazel_](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/test-infra/23665/pull-test-infra-bazel/1439743221289717760#1:build-log.txt%3A6605) job says that there needs to be a dashboard made, I think?
> 
> > * configuration error for (TestGroup) public-log-vendor-matcher: Each Test Group must be referenced by at least 1 Dashboard Tab.

Yes! We have code that configures a [TestGrid](https://testgrid.k8s.io/) view for periodic jobs automatically, so you can see historical test results. New jobs should have [annotations](https://github.com/kubernetes/test-infra/blob/5226d6124d168d2f238cf5edad50ba57db681043/config/jobs/kubernetes/test-infra/test-infra-periodics.yaml#L25-L28) on them that say which dashboard the results should be on.

BobyMCbobs:
> > The [_pull-test-infra-bazel_](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/test-infra/23665/pull-test-infra-bazel/1439743221289717760#1:build-log.txt%3A6605) job says that there needs to be a dashboard made, I think?
> > > * configuration error for (TestGroup) public-log-vendor-matcher: Each Test Group must be referenced by at least 1 Dashboard Tab.
> 
> Yes! We have code that configures a [TestGrid](https://testgrid.k8s.io/) view for periodic jobs automatically, so you can see historical test results. New jobs should have [annotations](https://github.com/kubernetes/test-infra/blob/5226d6124d168d2f238cf5edad50ba57db681043/config/jobs/kubernetes/test-infra/test-infra-periodics.yaml#L25-L28) on them that say which dashboard the results should be on.

Thank you @chases2, that's helpful! I've taken a look at a few more existing jobs to add those annotations

chases2:
If you want to use a dashboard that already exists, add the annotations and delete `config/testgrids/kubernetes/k8s-infra-public-pii/config.yaml`. It's trying to build the dashboard twice.

If you're trying to build your own dashboard, `config/testgrids/kubernetes/k8s-infra-public-pii/config.yaml` is the right place to put that config, but it needs a unique name.

BobyMCbobs:
> If you want to use a dashboard that already exists, add the annotations and delete `config/testgrids/kubernetes/k8s-infra-public-pii/config.yaml`. It's trying to build the dashboard twice.
> 
> If you're trying to build your own dashboard, `config/testgrids/kubernetes/k8s-infra-public-pii/config.yaml` is the right place to put that config, but your dashboard and dashboard group need different names from each other.

I think i'll go for a new dashboard. Though unsure if I should be putting the dashboard with wg-k8s-infra or not.
Do you have any thoughts @spiffxp?

config/jobs/kubernetes/k8s-infra-public-pii/k8s-infra-public-pii.yaml
```diff
@@ -0,0 +1,34 @@
+periodics:
+  - interval: 168h
+    agent: kubernetes
+    name: public-log-vendor-matcher
```
spiffxp, line 4:
```suggestion
    name: public-log-vendor-matcher
    cluster: k8s-infra-prow-build-trusted
```

This shouldn't be running in the google.com build cluster

config/testgrids/kubernetes/k8s-infra-public-pii/config.yaml
```diff
@@ -0,0 +1,7 @@
+dashboard_groups:
+  - name: k8s-infra-public-pii
+    dashboard_names:
+      - k8s-infra-public-pii-all
+
+dashboards:
+  - name: k8s-infra-public-pii-all
```
spiffxp, line 7: This should go into the sig-k8s-infra dashboard group, with a dashboard named something like sig-k8s-infra-artifacts

config/jobs/kubernetes/k8s-infra-public-pii/k8s-infra-public-pii.yaml
```diff
@@ -0,0 +1,34 @@
...
+              mountPath: /etc/asn-etl-pipeline
+          env:
+            - name: GCP_PROJECT
+              value: k8s-infra-ii-sandbox
+            - name: GCP_SERVICEACCOUNT
+              value: asn-etl@k8s-infra-ii-sandbox.iam.gserviceaccount.com
+            - name: GCP_BIGQUERY_DATASET
+              value: etl_script_generated_set
+            - name: GOOGLE_APPLICATION_CREDENTIALS
+              value: /etc/asn-etl-pipeline/asn-etl-pipeline-gcp-sa.json
```
spiffxp, line 30: We are definitely not passing around service account keys as secrets, we should be using workload identity instead

spiffxp (changes requested):
We need to understand more about the intent of this job.  There is a possibility it needs to run in its own cluster, even apart from k8s-infra-prow-build-trusted.  Need to read more in the k8s.io PR that is introducing what is supposed to run.

BobyMCbobs:
There was discussion a few weeks back about where this should run.
From what I remember, the options were:
- cloudbuild
- cloudrun
- a completely isolate PII cluster

I would appreciate reiteration of clarity once more.

I'm unsure how to configure jobs in cloudbuild and cloudrun, especially how to tell it a ServiceAccount.

/hold

cc @spiffxp

thockin:
I want to help, but I don't really have context to review this - what input are you looking for from me?

BenTheElder:
I just did a bulk scan of spiffxp assigned PRs after nudging him about setting github account status to busy (so the robot won't assign or suggest for now) while still extended OOO -- I don't have context on this either, just assigned a different OWNER.

Riaankl:
The aim of this periodic job is to autmatically update the data used in BQ & Data Studio to show the costs of distrubuting K8s artifacts.
At the moment the updated are done manually by @BobyMCbobs , but he would like to automate it. 
He will pick this up again when he is back in office in 2 weeks.

BobyMCbobs:
> @thockin @BenTheElder, the purpose of this PR is to propose and make a solution of the automation of [public-log-asn-matcher](https://github.com/kubernetes/k8s.io/tree/main/images/public-log-asn-matcher) job. There were discussions with @spiffxp about how this should run. The job of course contains access to PII for the Kubernetes public bucket logs. I would like for the automation of this job to be pushed forward. It would likely need to run with a particular GCP service account so that it can access the buckets, BigQuery databases and produce DataStudio dashboards. The direction is likely one of the following
> 
> * Cloud Run
> * Cloud Build
> * Prow (somewhere); or
> * An alternate method
> 
> A goal is that it is well isolated and only few team on sig-k8s-infra have access to it, given PII NDA. The job should run perhaps once a week or so.
> 
> Something that I might've been stuck on when implementing this job is the service accounts with creation, access or configuring in the job.
> 
> Which way of running it might be considered the best? and also what are your thoughts?

@thockin, might you have any suggestion and/or would like to Pair on this?

dims:
it's been a year. do we want to talk about this in next SIG call?

Riaankl:
/close
From K8s-infra meeting 31 Aug having a periodic job does not make sense at the moment.
The Main purpose of the data was to find out use by vendors, which is accomplished.
If an update in the data is required the job can be run manually.

PR #23667, opened 2021-09-20
upgrade pause image to 3.6

kerthcet:
Signed-off-by: kerthcet <kerthcet@gmail.com>
upgradation of pause image to tag 3.6,  support for Windows Server 2022 was added to the k8s.gcr.io/pause:3.6 image.
refer to :https://github.com/kubernetes/kubernetes/issues/104890

### DIFF
--- a/config/jobs/kubernetes/sig-scalability/sig-scalability-presets.yaml
+++ b/config/jobs/kubernetes/sig-scalability/sig-scalability-presets.yaml
@@ -310,7 +310,7 @@ presets:
     preset-e2e-scalability-periodics: "true"
   env:
   - name: NODE_PRELOAD_IMAGES
-    value: k8s.gcr.io/pause:3.1
+    value: k8s.gcr.io/pause:3.6
 
 - labels:
     preset-e2e-scalability-periodics-master: "true"
--- a/config/jobs/kubernetes/sig-windows/windows-gce.yaml
+++ b/config/jobs/kubernetes/sig-windows/windows-gce.yaml
@@ -534,7 +534,7 @@ periodics:
       image: gcr.io/k8s-staging-test-infra/kubekins-e2e:v20210917-ee1e7c845b-master
       env:
       - name: CL2_CONTAINER_IMAGE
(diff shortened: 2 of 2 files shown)


mborsz:
The `NODE_PRELOAD_IMAGES` must be in sync with:
* https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/testing/load/deployment.yaml#L43
* https://github.com/kubernetes/perf-tests/blob/release-{$ver}/clusterloader2/testing/load/deployment.yaml#L43

Those values are used only for NODE_PRELOAD_IMAGES clusterloader tests and I'm not aware of any clusterloader test running on windows. We also still use pause:3.0 image in some places (https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/testing/load/config.yaml#L172) and we do need to use two separate images to properly validate rolling update of daemonset.

Do you really want to update this image?
I don't think it will give any value to clusterloader tests, but migration to pause:3.6 may require some nontrivial work.

kerthcet:
> The `NODE_PRELOAD_IMAGES` must be in sync with:
> 
> * https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/testing/load/deployment.yaml#L43
> * [https://github.com/kubernetes/perf-tests/blob/release-{$ver}/clusterloader2/testing/load/deployment.yaml#L43](https://github.com/kubernetes/perf-tests/blob/release-%7B$ver%7D/clusterloader2/testing/load/deployment.yaml#L43)
> 
> Those values are used only for NODE_PRELOAD_IMAGES clusterloader tests and I'm not aware of any clusterloader test running on windows. We also still use pause:3.0 image in some places (https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/testing/load/config.yaml#L172) and we do need to use two separate images to properly validate rolling update of daemonset.
> 
> Do you really want to update this image?
> I don't think it will give any value to clusterloader tests, but migration to pause:3.6 may require some nontrivial work.

thanks for advices, i thinks it's valueless since it's only related to the clusterloader tests, however we'll make a lot of extra works. close the pr then.

PR #23679, opened 2021-09-21
Add heapster node to periodic 100 nodes scalability test jobs

tosi3k:
/sig scalability
/assign @wojtek-t


marseel:
I'm personally against changing tests for older releases. Why do we want to make this change?

tosi3k:
> I'm personally against changing tests for older releases. Do we have some runs that confirm it doesn't break it?

Left out the release branch jobs configs, as suggested. I tested the master branch job and it works as expected.

PR #23683, opened 2021-09-21
kops grid: Make sure we get the RHEL 8.4 image

justinsb:
It looks like a patch revision of 8.2 was published, so this image was
being picked up in preference to the newer 8.4.


spiffxp (approved):
/approve
/lgtm

PR #23688, opened 2021-09-21
bugzilla: add cf_verified field

AlexNPavel:
Adds support for the custom verified field for redhat bugzilla.


matthyx:
@AlexNPavel are you supposed to use this field somewhere else than `types.go`?

stevekuznetsov (approved):

AlexNPavel:
> @AlexNPavel are you supposed to use this field somewhere else than `types.go`?

It won't be used elsewhere in this repo, but will be used by repos for some redhat projects that import this bugzilla client.

Issue #23692, opened 2021-09-21
Prow support Google Cloud Build as a runtime


chaodaiG:
**What would you like to be added**:

A new runtime for prow: Google Cloud Build

**Why is this needed**:

Provide a way for people to define a prowjob that results in creating a Google Cloud Build build instead of a pod.
This allows a team to schedule prow workloads that:

- run in their project without maintaining a GKE cluster
- securely create docker images

Designed as of: https://docs.google.com/document/d/13ceGB1XqOLtzUaf31ZCDHi5L66MbunWXwxqUhYSvc34/edit#

PR #23693, opened 2021-09-21
Create Google cloudbuild client package

chaodaiG:
This is a wrapper of cloudbuild package, creating ahead of time for easier code review later

Part of: https://github.com/kubernetes/test-infra/issues/23692

/cc @cjwagner @listx @mpherman2

### DIFF
--- a/prow/googlecloudbuild/client/client.go
+++ b/prow/googlecloudbuild/client/client.go
@@ -1,0 +1,192 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package client
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"time"
+
+	cloudbuild "cloud.google.com/go/cloudbuild/apiv1/v2"
+	"google.golang.org/api/iterator"
+	"google.golang.org/api/option"
+	cloudbuildpb "google.golang.org/genproto/googleapis/devtools/cloudbuild/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+)
+
+const (
+	// GCB automatically assigns random build ids to a new build, so it's not
+	// possible to associate a GCB build with a prow job by id without storing
+	// mapping somewhere else. Utilizing GCB tags for achieving this.
+
+	// ProwLabelSeparator formats labels key:val pairs into `{key} ::: {val}`
+	// The separator was arbitrarily selected.
+	ProwLabelSeparator = " ::: "
+)
+
+// Operator is the interface that's highly recommended to be used by any package
+// that imports current package. Unit test can be carried out with a mock
+// implemented under fake of current package.
+type Operator interface {
+	GetBuild(ctx context.Context, project, id string) (*cloudbuildpb.Build, error)
+	ListBuildsByTag(ctx context.Context, project string, tags []string) ([]*cloudbuildpb.Build, error)
+	CreateBuild(ctx context.Context, project string, bld *cloudbuildpb.Build) (*cloudbuildpb.Build, error)
+	CancelBuild(ctx context.Context, project, id string) (*cloudbuildpb.Build, error)
+}
+
+// ProwLabel formats labels key:val pairs into `{key} ::: {val}`.
+// These labels will be parsed by prow controller manager for mapping a GCB
+// build to a prow job.
+func ProwLabel(key, val string) string {
+	return key + ProwLabelSeparator + val
+}
+
+// KvPairFromProwLabel trims label into key:val pairs. returns empty string as value if
+// the label is not formatted as prow label format.
+func KvPairFromProwLabel(tag string) (key, val string) {
+	parts := strings.SplitN(tag, ProwLabelSeparator, 2)
+	if len(parts) != 2 {
+		return tag, ""
+	}
+	return parts[0], parts[1]
+}
+
+// GetProwLabels gets labels from cloud build struct, simulating k8s pods labels format
+// of map[string]string.
+func GetProwLabels(bld *cloudbuildpb.Build) map[string]string {
+	res := make(map[string]string)
+	for _, tag := range bld.Tags {
+		key, val := KvPairFromProwLabel(tag)
+		if val != "" {
+			res[key] = val
+		}
+	}
+	return res
+}
+
+var _ Operator = (*Client)(nil)
+
+// Client wraps native cloudbuild client.
+type Client struct {
+	interactor *cloudbuild.Client
+}
+
+// NewClient creates a new Client, with optional credentialFile.
+func NewClient(ctx context.Context, credentialFile string) (*Client, error) {
+	var opts []option.ClientOption
+	// Authenticating with key file if it's provided.
+	if len(credentialFile) > 0 {
+		opts = append(opts, option.WithCredentialsFile(credentialFile))
+	}
+
+	cbClient, err := cloudbuild.NewClient(ctx, opts...)
+	if err != nil {
+		return nil, err
+	}
+	return &Client{interactor: cbClient}, nil
+}
+
+// GetBuild gets build by GCB build id.
+func (c *Client) GetBuild(ctx context.Context, project, id string) (*cloudbuildpb.Build, error) {
+	return c.interactor.GetBuild(ctx, &cloudbuildpb.GetBuildRequest{
+		ProjectId: project,
+		Id:        id,
+	})
+}
+
+// ListBuildsByTag lists builds by GCB tags.
+//
+// This will be used by prow for listing builds triggered by prow, for example
+// `created-by-prow ::: true`.
+func (c *Client) ListBuildsByTag(ctx context.Context, project string, tags []string) ([]*cloudbuildpb.Build, error) {
+	// pageSize is used by ListBuildsByTag only, define here for locality
+	// reason. Can move up if pagination is needed by more functions.
+	const pageSize = 50
+	var res []*cloudbuildpb.Build
+	var tagsFilters []string
+	for _, tag := range tags {
+		tagsFilters = append(tagsFilters, "tags="+tag)
+	}
+
+	iter := c.interactor.ListBuilds(ctx, &cloudbuildpb.ListBuildsRequest{
+		ProjectId: project,
+		PageSize:  pageSize,
+		Filter:    strings.Join(tagsFilters, " AND "),
+	})
+	// ListBuilds already fetches all, just need to do pagination.
+	pager := iterator.NewPager(iter, pageSize, "")
+	for {
+		var buildsInPage []*cloudbuildpb.Build
+		nextPageToken, err := pager.NextPage(&buildsInPage)
+		if err != nil {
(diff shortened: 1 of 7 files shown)


chaodaiG:
test failure was due to bazel dependencies issue. Have tried:
1. ./hack/update-deps.sh
1. ./hack/update-deps.sh --minor ./prow/googlecloudbuild/client
1. manual `go get`, `go mod tidy`, `./hack/update-bazel.sh`, which is pretty similar to 2) above

didn't have much luck to make bazel happy
/uncc @cjwagner @listx @mpherman2 
Will add back once tests are happy

chaodaiG:
Error:
```
ERROR: /bazel-scratch/.cache/bazel/_bazel_root/05618a594cb3499a4b912817865867cd/external/com_google_cloud_go_cloudbuild/apiv1/v2/BUILD.bazel:3:11: GoCompilePkg external/com_google_cloud_go_cloudbuild/apiv1/v2/apiv1.a failed (Exit 1)
external/com_google_cloud_go_cloudbuild/apiv1/v2/cloud_build_client.go:190:33: undefined: cloudbuild.ApproveBuildRequest
external/com_google_cloud_go_cloudbuild/apiv1/v2/cloud_build_client.go:199:42: undefined: cloudbuild.ReceiveTriggerWebhookRequest
external/com_google_cloud_go_cloudbuild/apiv1/v2/cloud_build_client.go:199:106: undefined: cloudbuild.ReceiveTriggerWebhookResponse
...
```

The first error is from line https://github.com/googleapis/google-cloud-go/blob/5606b54b97bb675487c6c138a4081c827218f933/cloudbuild/apiv1/v2/cloud_build_client.go#L190, which references `cloudbuildpb.ApproveBuildRequest` from `cloudbuildpb "google.golang.org/genproto/googleapis/devtools/cloudbuild/v1"`, the go_repository rule of this repo is updated in this PR as:
```
    go_repository(
        name = "org_golang_google_genproto",
        build_file_generation = "on",
        build_file_proto_mode = "disable",
        importpath = "google.golang.org/genproto",
-      sum = "h1:wtujag7C+4D6KMoulW9YauvK2lgdvCMS260jsqqBXr0=",
-      version = "v0.0.0-20210602131652-f16073e35f0c",
+     sum = "h1:2ncG/LajxmrclaZH+ppVi02rQxz4eXYJzGHdFN4Y9UA=",
+     version = "v0.0.0-20210921142501-181ce0d877f6",
    )
```
And `ApproveBuildRequest` clearly exists at https://github.com/googleapis/go-genproto/blob/181ce0d877f633434cbea7d1fff60548fe280d94/googleapis/devtools/cloudbuild/v1/cloudbuild.pb.go#L3424.

Not sure how to debug further

chaodaiG:
Read a little more into this problem, and now understand it better. Have been puzzled by a question:
- why gazelle generated BUILD file references `"@go_googleapis//google/devtools/cloudbuild/v1:cloudbuild_go_proto"` for `"google.golang.org/genproto/googleapis/devtools/cloudbuild/v1"`, when it actually lives at https://github.com/googleapis/googleapis/blob/master/google/devtools/cloudbuild/v1 instead of where the generated proto go file at https://github.com/googleapis/go-genproto/blob/181ce0d877f633434cbea7d1fff60548fe280d94/googleapis/devtools/cloudbuild/v1/cloudbuild.pb.go

The answer turned out to be:
```
go_proto_library(
    name = "cloudbuild_go_proto",
    compilers = ["@io_bazel_rules_go//proto:go_grpc"],
    importpath = "google.golang.org/genproto/googleapis/devtools/cloudbuild/v1",
    protos = [":cloudbuild_proto"],
    deps = [
        "//google/api:annotations_go_proto",
        "//google/api:httpbody_go_proto",
        "//google/longrunning:longrunning_go_proto",
    ],
)
```

I think the fix for this is actually updating `go_googleapis` from this repo

chaodaiG:
cloudbuild golang client introduced breaking client interface change in v0.83.0 as https://github.com/googleapis/google-cloud-go/blob/v0.83.0/cloudbuild/apiv1/v2/cloud_build_client.go. So if possible would prefer not use an older version of client. 

Created https://github.com/bazelbuild/rules_go/pull/2959 for updating rules_go, which should fix the compatibility issue here

chaodaiG:
Updated `rules_go` repo, and got new error:
```
{bazel_dir}/external/io_bazel_rules_go/go/private/skylib/lib/versions.bzl:104:13:
...
Current Bazel version is 3.4.1, expected at least 4.0.0
```

will probably need to update bazel in this repo to >4.0.0

chaodaiG:
Rebased and reverted all bazel hacks, expecting bazel test to fail, which shouldn't matter after bazel is not blocking as mentioned in https://github.com/kubernetes/test-infra/issues/23796#issuecomment-1054702008

prow/googlecloudbuild/client/client.go
```diff
@@ -0,0 +1,182 @@
...
+
+// GetProwLabels gets labels from cloud build struct, simulating k8s pods labels format
+// of map[string]string
+func GetProwLabels(bld *cloudbuildpb.Build) map[string]string {
+	res := make(map[string]string)
+	for _, tag := range bld.Tags {
+		key, val := LabelFromTag(tag)
+		if val != "" {
+			res[key] = val
+		}
```
listx, line 75: Should we log cases where the tag was malformed (where we couldn't deserialize it)?

prow/googlecloudbuild/client/client.go
```diff
@@ -0,0 +1,182 @@
...
+	}
+	return res
+}
+
+// Client wraps native cloudbuild client
```
listx, line 80: Might be worth mentioning here that `Client` implements the Operator interface.

prow/googlecloudbuild/client/client.go
```diff
@@ -0,0 +1,182 @@
...
+type Client struct {
+	interactor *cloudbuild.Client
+}
+
+// NewClient creates a new Client, with optinal credentialFile
```
listx, line 85: s/optinal/optional

Nit: Also, I've noticed that your comment sentences lack a period at the end of the sentence. For cleanliness it would help to add those (at least for all top-level comments).

prow/googlecloudbuild/client/client.go
```diff
@@ -0,0 +1,182 @@
...
+func NewClient(ctx context.Context, credentialFile string) (*Client, error) {
+	var opts []option.ClientOption
+	// Authenticating with key file if it's provided
+	if len(credentialFile) > 0 {
+		if _, err := os.Stat(credentialFile); err != nil {
```
listx, line 90: Is this necessary? Wouldn't some other dependency error out if the credential file cannot be found?

prow/googlecloudbuild/client/client.go
```diff
@@ -0,0 +1,182 @@
...
+	}
+
+	cbClient, err := cloudbuild.NewClient(ctx, opts...)
+	if err != nil {
+		return nil, err
+	}
+	return &Client{interactor: cbClient}, nil
+}
+
+// GetBuild gets build by it's id
```
listx, line 103: s/it's/its

NIt: Mention what ID this is (GCB Build ID?)

prow/googlecloudbuild/client/client.go
```diff
@@ -0,0 +1,182 @@
...
+func (c *Client) GetBuild(ctx context.Context, project, id string) (*cloudbuildpb.Build, error) {
+	return c.interactor.GetBuild(ctx, &cloudbuildpb.GetBuildRequest{
+		ProjectId: project,
+		Id:        id,
+	})
+}
+
+// ListBuildsByTag lists builds by tags
```
listx, line 111: Nit: Explain a bit more what "tags" are. I assume we (this or a related component) will set these tags when we create the build.

prow/googlecloudbuild/client/client.go
```diff
@@ -0,0 +1,182 @@
...
+func (c *Client) ListBuildsByTag(ctx context.Context, project string, tags []string) ([]*cloudbuildpb.Build, error) {
+	const pageSize = 50
```
listx, line 113: Nit: move `pageSize` to the toplevel const declaration area (line 33)? Typically I like to see all consts together in one place, unless there's a reason otherwise (if so, could you comment here why)?

prow/googlecloudbuild/client/client.go
```diff
@@ -0,0 +1,182 @@
...
+	var res []*cloudbuildpb.Build
+	var tagsFilters []string
+	for _, tag := range tags {
+		tagsFilters = append(tagsFilters, "tags="+tag)
```
listx, line 117: So given tags ["a", "b", "c"], tagsFilters will look like ["tags=a", "tags=b", "tags=c"] ? Is this right? (Given the plural of `tags` I was expecting something like `tags=a,b,c`).

EDIT: I see that you do a ` AND ` strings.Join below, so it makes more sense now (although on first glance it still looks a bit awkward).

Should we be doing any empty string checks here? E.g., `tags=` by itself will probably break the Filter, right? E.g., `tags= AND tags=foo` will probably not work?

prow/googlecloudbuild/client/client.go
```diff
@@ -0,0 +1,182 @@
...
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	// CreateBuild returns CreateBuildOperation, wait until the operation results in build.
+	// CreateBuildOperation contains `Wait` method, which however polls every minute, which is
+	// too slow. So use `wait.Poll` instead.
+	var triggered *cloudbuildpb.Build
+	if err := wait.Poll(100*time.Millisecond, 30*time.Second, func() (bool, error) {
```
listx, line 157: I think 100*time.Millisecond should be a a named constant instead. But also, is this 0.10 second interval too short? I fear getting throttled by GCB if we're doing this across possibly hundreds of pending builds.

Also, what does the `30*time.Second` interval do? Can you name this constant (meaningfully) as well and use it here?

prow/googlecloudbuild/client/client.go
```diff
@@ -0,0 +1,182 @@
...
+		triggered, err = op.Poll(ctx)
+		if err != nil {
+			return false, fmt.Errorf("failed to create build in project %s: %w", project, err)
+		}
+		// Poll surprisingly waits until the build completes before returning build,
+		// Metadata somehow does
```
listx, line 163: Unfinished sentence?

prow/googlecloudbuild/client/client.go
```diff
@@ -0,0 +1,182 @@
...
+		}
+		triggered = meta.GetBuild()
+		return triggered != nil, nil
+	}); err != nil {
+		return nil, fmt.Errorf("failed waiting for build in project %s appear: %w", project, err)
+	}
+	return triggered, nil
+}
+
+// CancelBuild cancels build and wait for it
```
listx, line 176: Does it wait automatically or do we need to put in a `wait.Poll` call in here too like we did for CreateBuild?

listx (commented):
Haven't looked at tests yet. Mostly nits as the overall design looks correct to me.

prow/googlecloudbuild/client/client.go
chaodaiG: good point, yeah "tags=" would break. However I'm a bit split here, on the one hand we should guard against empty string, on the other hand, I also feel like this is caller's responsibility

prow/googlecloudbuild/client/client.go
chaodaiG: done

prow/googlecloudbuild/client/client.go
chaodaiG: kind of, finished

prow/googlecloudbuild/client/client.go
chaodaiG: it does wait by itself

listx:
Thanks!

/lgtm

chaodaiG (commented):
/unhold

PR #23695, opened 2021-09-21
Allow the bugzilla client to clone bugs with specific mutations

droslean:
/cc @alvaroaleman @AlexNPavel 

### DIFF
--- a/prow/bugzilla/client.go
+++ b/prow/bugzilla/client.go
@@ -64,7 +64,7 @@ type Client interface {
 	// https://bugzilla.redhat.com/docs/en/html/api/core/v1/comment.html#create-comments
 	CreateComment(bug *CommentCreate) (int, error)
 	// CloneBug clones a bug by creating a new bug with the same fields, copying the description, and updating the bug to depend on the original bug
-	CloneBug(bug *Bug) (int, error)
+	CloneBug(bug *Bug, mutations ...func(bug *BugCreate)) (int, error)
 	// UpdateBug updates the fields of a bug on the server
 	// https://bugzilla.readthedocs.io/en/latest/api/core/v1/bug.html#update-bug
 	UpdateBug(id int, update BugUpdate) error
@@ -619,7 +619,7 @@ func cloneBugStruct(bug *Bug, subcomponents map[string][]string, comments []Comm
 
 // clone handles the bz client calls for the bug cloning process and allows us to share the implementation
 // between the real and fake client to prevent bugs from accidental discrepencies between the two.
-func clone(c Client, bug *Bug) (int, error) {
+func clone(c Client, bug *Bug, mutations []func(bug *BugCreate)) (int, error) {
 	subcomponents, err := c.GetSubComponentsOnBug(bug.ID)
 	if err != nil {
 		return 0, fmt.Errorf("failed to check if bug has subcomponents: %w", err)
@@ -628,7 +628,13 @@ func clone(c Client, bug *Bug) (int, error) {
 	if err != nil {
 		return 0, fmt.Errorf("failed to get parent bug's comments: %w", err)
 	}
-	id, err := c.CreateBug(cloneBugStruct(bug, subcomponents, comments))
+
+	newBug := cloneBugStruct(bug, subcomponents, comments)
+	for _, mutation := range mutations {
+		mutation(newBug)
+	}
+
+	id, err := c.CreateBug(newBug)
 	if err != nil {
 		return id, err
 	}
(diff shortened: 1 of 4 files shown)


prow/bugzilla/client.go
```diff
@@ -65,6 +65,8 @@ type Client interface {
 	CreateComment(bug *CommentCreate) (int, error)
 	// CloneBug clones a bug by creating a new bug with the same fields, copying the description, and updating the bug to depend on the original bug
 	CloneBug(bug *Bug) (int, error)
+	// CloneBugToRelease is the same with CloneBug() but with the target_release included.
+	CloneBugToTargetRelease(bug *Bug, targetReleases []string) (int, error)
```
petr-muller, line 69: I'm honestly not sure if we need a specialized method for this... Why is it critical for the code to create the bug this way right away instead of creating the bug and subsequently updating its field? And if we *do* need something like this, maybe we need something more generic (are there other fields that *could* be added like this). Maybe we could add a `CloneBugWithMutate(bug *Bug, mutations []func(bug *Bug) error)` - make a local clone of the bug exactly matching `CloneBug` result, then run a bunch of mutation functions on the clone (in this case it would add a relesae), then actually Create this in BZ 

Also, why is the parmater a slice? The name of the method suggests "release" (singular) the param is a slice (plural). Can someone pass more releases there? If yes, what will happen?

droslean: Well, we can always use the existing `CloneBug()` function by mutating the object before passing it, but we want to avoid cloning a bug with the same `target_release` values, so if we do it this way we need to validate this. A specialized method just for the target release makes sense in a way of cloning a bug to specific releases. Also, the parameter is a slice because that's the value in Bugzilla as well. See https://github.com/kubernetes/test-infra/blob/master/prow/bugzilla/types.go#L101 as well. 

> make a local clone of the bug exactly matching CloneBug result, then run a bunch of mutation functions on the clone (in this case it would add a relesae), then actually Create this in BZ
This is already happening in [here](https://github.com/kubernetes/test-infra/blob/3e7e43871ab2fd9941bdb5fe31a7068dfef81301/prow/bugzilla/client.go#L565) bug we mutate specific fields. (That's why the `BugCreate` struct exists.

In general the target release is a special case when cloning the bug, and I believe that we should treat it in a different method. Your `CloneBugWithMutate` suggestion makes absolute sense but we will need to validate which fields are mutated and that will make things more complicated.

stevekuznetsov: >  we will need to validate which fields are mutated 

Why? This is a generic Bugzilla client, to be used by "anyone". The caller who decided to mutate is the owner of the source of truth for what should be mutated. The client library (this file) has no opinion on how a bug is mutated.

droslean: That will lead to clone the actual target_release without any mutation from the user side. In this case, the user will have to remove the value from target_release before cloning.

droslean: Currently, the `target_release` is not including in cloning, which means that we have to add it to the process. And, if we go with the mutation solution, as @petr-muller suggested, it will break the backwards compatibility, because we will force users that already using the client to use the mutate method to remove the target release instead.

petr-muller: I actually misread the API, I thought the clone gets a bug from BZ, does a client-side copy, then creates new bug in BZ via the copy. It does not do the first thing so IMO the need for a specialized method is even lower. I'm convinced we should not have a specialized method for this, it does not make sense in a generic client package.

I don't think we need to do any validation or break any compatibility. `CloneBug(bug)` would behave exactly like now, zero change (empty TargetRelease). Then you could do it like this:

```go
func addTarget(bug *BugCreate) { bug.TargetRelease = []string{"X" }
CloneBugWithMutate(bug, addTarget)
```

This way, the only chaange needed would be in `clone()`:
```diff
- id, err := c.CreateBug(cloneBugStruct(bug, subcomponents, comments, targetReleases))
+ bug := cloneBugStruct(bug, subcomponents, comments)
+ for _, f := range mutations {
+   f(bug)
+ }
+ id, err := c.CreateBug(bug)
```

You could do it even without a new method by adding variadic args to `CloneBug` (I think I'd actually prefer that). This way you allow the client lib to be super generic, and allow the client code to customize.

petr-muller: I don't think we need to do any validation - if the callsite wants to customize, it is not the packages' place to play police. If the customizations result in something bad, then BZ hopefully refuses it.

droslean: Thanks, @petr-muller. This approach makes sense. Updated.

prow/bugzilla/client.go
```diff
@@ -65,6 +65,8 @@ type Client interface {
 	CreateComment(bug *CommentCreate) (int, error)
 	// CloneBug clones a bug by creating a new bug with the same fields, copying the description, and updating the bug to depend on the original bug
 	CloneBug(bug *Bug) (int, error)
+	// CloneBugWithMutate does the same as CloneBug() but with extra given mutations
+	CloneBugWithMutate(bug *Bug, mutations []func(bug *BugCreate)) (int, error)
```
stevekuznetsov, line 69: You don't need a new method, make  `mutations` variadic and existing calls to `CloneBug` will just work.

petr-muller (approved):
Pretty clean stuff, thanks!

PR #23678, opened 2021-09-21
Switch kettle/triage service account keys to Workload Identity.

cjwagner:
Once this merge I'll do the following:
- For the `metrics-kettle` job I'll
    - bind the new KSA to the GSA
    - Switch the job to use the KSA rather than a GSA key.
- For the Kettle deployments I'll
    - Create the KSA
    - Bind the KSA to the GSA
    - Manually deploy staging
    - Manually deploy prod

/hold
/assign @MushuEE @fejta

PR #23698, opened 2021-09-21
Switch metrics-kettle job to use Workload Identity.

cjwagner:
Follow up to https://github.com/kubernetes/test-infra/pull/23678

/assign @MushuEE @fejta 
/cc @BenTheElder


fejta (approved):
/hold

PR #23711, opened 2021-09-22
DefaultDecorationConfigs: reinstate old comment

listx:
This comment was removed in 18edc565e619aca0d89e7b201bff1ed2ec1e50d4 (#23500), but we just reinstate it here.

### DIFF
--- a/prow/config/prow-config-documented.yaml
+++ b/prow/config/prow-config-documented.yaml
@@ -540,6 +540,7 @@ plank:
     build_cluster_status_file: ' '
 
     # DefaultDecorationConfigEntries is used to populate DefaultDecorationConfigs.
+
     # Each entry in the slice specifies Repo and Cluster regexp filter fields to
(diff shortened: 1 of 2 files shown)


prow/config/prow-config-documented.yaml
```diff
@@ -540,10 +540,6 @@ plank:
     build_cluster_status_file: ' '
 
     # DefaultDecorationConfigEntries is used to populate DefaultDecorationConfigs.
-    # Each entry in the slice specifies Repo and Cluster regexp filter fields to
-    # match against jobs and a corresponding DecorationConfig. All entries that
-    # match a job are used. Later matching entries override the fields of earlier
-    # matching entries.
```
cjwagner, line 546: I don't think we should remove this. If we do, this information is not included in the documented config file at all.
Since this is important info I don't think redundancy in the godocs hurts.

listx: done

cjwagner (approved):

PR #23724, opened 2021-09-23
Change the test back to always run and remove check-leaked-resources flag

lizhuqi:
This test shall be always run to add some test coverage for gce. I also remove check-leaked-resources flag for now because of https://github.com/kubernetes/kubernetes/issues/104131. I will add the check back once the issue is fixed.


andyzhangx (commented):
/ok-to-test
/hold
hold on until this test run pass on smb repo

andyzhangx (approved):
/lgtm
/approve
/hold cancel

PR #23726, opened 2021-09-23
add `prow/config` subdirectory to hack/verify-codegen.sh

listx:
This is required because #23500 introduced a new generated file under `config`, namely `prow/config/zz_generated.deepcopy.go`.

### DIFF
--- a/hack/verify-codegen.sh
+++ b/hack/verify-codegen.sh
@@ -36,7 +36,7 @@ DIFFROOT="${SCRIPT_ROOT}/prow"
 TMP_DIFFROOT="${TEST_TMPDIR}/prow"
 
 mkdir -p "${TMP_DIFFROOT}"
-cp -a "${DIFFROOT}"/{apis,client,spyglass} "${TMP_DIFFROOT}"
+cp -a "${DIFFROOT}"/{apis,client,config,spyglass} "${TMP_DIFFROOT}"
 
 clean=yes # bazel test files are read-only, must first delete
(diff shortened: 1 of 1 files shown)


listx:
@cjwagner Because all tests passed I assume the verify-codegen script is broken.

cjwagner:
> @cjwagner Because all tests passed I assume the verify-codegen script is broken.

Yep seems like it. Is it broken for the existing code gen too or just the codegen for the `config` package that you added?

listx:
> > @cjwagner Because all tests passed I assume the verify-codegen script is broken.
> 
> Yep seems like it. Is it broken for the existing code gen too or just the codegen for the `config` package that you added?

I've updated the PR to make changes in both codegen files, and it only caught the old one. So the existing codegen stuff was working fine after all.

Looking into this more, this is because the verify script does not check the entire repo but only select subdirs (and currently we do not include the `config` subdir). I'll follow up with a fix.

listx:
Great, it's now catching both places in https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/test-infra/23726/pull-test-infra-bazel/1441250545242214400. Let me just remove the purposely-offending changes to convert this test PR into a real one...

listx:
> Great, it's now catching both places in https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/test-infra/23726/pull-test-infra-bazel/1441250545242214400. Let me just remove the purposely-offending changes to convert this test PR into a real one...

done

cjwagner (approved):

PR #23721, opened 2021-09-23
Switch to cos-beta for cgroupv2 image config

bobbypage:
Pick up newer COS image under 5.10 kernel which includes cgroupv2 cpu stat fixes

See https://github.com/kubernetes/kubernetes/issues/103759#issuecomment-926024150 for background

```
$ gcloud compute images list --project="cos-cloud" --no-standard-images
NAME                       PROJECT    FAMILY      DEPRECATED  STATUS
cos-81-12871-1317-1        cos-cloud  cos-81-lts              READY
cos-85-13310-1308-19       cos-cloud  cos-85-lts              READY
cos-89-16108-534-2         cos-cloud  cos-89-lts              READY
cos-beta-93-16623-0-15     cos-cloud  cos-beta                READY
cos-dev-97-16687-0-0       cos-cloud  cos-dev                 READY
cos-stable-89-16108-534-2  cos-cloud  cos-stable              READY
```

cos-beta will use `cos-93` which is running under 5.10 kernel (https://cloud.google.com/container-optimized-os/docs/release-notes/m93) 

Fixes https://github.com/kubernetes/kubernetes/issues/103759

PR #23728, opened 2021-09-23
add cos-cgroupv2-containerd-e2e test-grid

Dragoncell:


config/jobs/kubernetes/sig-node/containerd.yaml
```diff
@@ -879,3 +879,33 @@ periodics:
...
+      - --
+      - --check-leaked-resources
+      - --env=ENABLE_POD_SECURITY_POLICY=true
+      - --env=KUBE_NODE_EXTRA_METADATA=user-data=/workspace/github.com/containerd/containerd/test/e2e/node.yaml,containerd-configure-sh=/workspace/github.com/containerd/containerd/cluster/gce/configure.sh,containerd-env=/workspace/test-infra/jobs/e2e_node/containerd/containerd-master/cgroupv2/env-cgroupv2
+      - --extract=ci/latest
+      - --gcp-node-image=gci
+      - --gcp-nodes=4
+      - --gcp-zone=us-west1-b
+      - --ginkgo-parallel=30
+      - --image-family=cos-beta
```
bobbypage, line 903: Maybe leave comment here why we using `cos-beta`, similar as https://github.com/kubernetes/test-infra/pull/23721/files

config/jobs/kubernetes/sig-node/containerd.yaml
```diff
@@ -879,3 +879,33 @@ periodics:
...
+      - --image-project=cos-cloud
+      - --provider=gce
+      - --test_args=--ginkgo.skip=\[Slow\]|\[Serial\]|\[Disruptive\]|\[Flaky\]|\[Feature:.+\] --minStartupPods=8
```
bobbypage, line 906: I think we need to also figure out how to enable systemd cgroup driver in k/k startup scripts.

Because for cgroupv2, we are using systemd cgroup driver -- https://github.com/kubernetes/test-infra/blob/master/jobs/e2e_node/containerd/containerd-master/cgroupv2/env-cgroupv2#L10

used here
https://github.com/containerd/containerd/blob/main/contrib/gce/configure.sh#L180
https://github.com/containerd/containerd/blob/main/contrib/gce/configure.sh#L209

But we also need to ensure kubelet will use systemd cgroup driver with `--cgroup-driver=systemd`.

bobbypage: you might be able to set `NODE_KUBELET_TEST_ARGS`
https://github.com/kubernetes/kubernetes/blob/master/cluster/gce/util.sh#L782

but it looks like it's already set here - https://github.com/kubernetes/test-infra/blob/68c0582bee9da1c995249839ff29cd1088952b03/config/jobs/kubernetes/sig-node/containerd.yaml#L22

so I'm not sure, but maybe there is way to override it for this specific job and set it to something like `--runtime-cgroups=/system.slice/containerd.service --cgroup-driver=systemd`

config/jobs/kubernetes/sig-node/containerd.yaml
Dragoncell: Thanks, added the comment for using `cos-beta`.

config/jobs/kubernetes/sig-node/containerd.yaml
Dragoncell: Thanks for the detail information. Yeah, `NODE_KUBELET_TEST_ARGS` should work by specifying the thing like ` - --env=NODE_KUBELET_TEST_ARGS=--cgroup-driver=systemd`. I didn't include the `--runtime-cgroups=/system.slice/containerd.service` as it is already specified in `KUBELET_TEST_ARGS` and be used to construct common kubelet args for both master and nodes: https://github.com/kubernetes/kubernetes/blob/master/cluster/gce/util.sh#L742.

bobbypage: hmm, I'm not 100% sure but I would assume that if you specify `--env=NODE_KUBELET_TEST_ARGS`, it will override the environment variable used above (i.e. [here](https://github.com/kubernetes/test-infra/blob/2ff402f8b64bb15323bec3b72c0182624c758e47/config/jobs/kubernetes/sig-node/containerd.yaml#L21-L22). 

So I think you'll need to specify both here to ensure `runtime-cgroups` is still passed as a flag, something like

```
--env=NODE_KUBELET_TEST_ARGS=runtime-cgroups=/system.slice/containerd.service --cgroup-driver=systemd
```

Dragoncell: The `NODE_KUBELET_TEST_ARGS` is not same as `KUBELET_TEST_ARGS` (missing`NODE`) used above actually([here](https://github.com/kubernetes/test-infra/pull/22130/commits/8a193964a1fd038ed257cf5599b9d98ab41ad112#R21))
so it won't get override. `NODE_KUBELET_TEST_ARGS` only appends to kubelet flags for node, `KUBELET_TEST_ARGS` is used for both master and nodes.

bobbypage: oh, I see, didn't notice the difference there :) 

Thinking about it some more though, I think we want this to be in `KUBELET_TEST_ARGS` actually.

With the current config, both the master and node will be running cgroupv2 and we want to ensure that the kubelet running on the master (running any static pods) is also using systemd cgroup driver, just like the nodes.

So I think we'll want 

```
--env=KUBELET_TEST_ARGS=runtime-cgroups=/system.slice/containerd.service --cgroup-driver=systemd
```

Dragoncell: I see. If we want master also running in cgroupv2, `KUBE_MASTER_EXTRA_METADATA` needs to be override as well. Will update it

bobbypage: Thanks, nice catch, I think it'll best if we test cgroupv2 on both master and node in these cluster E2E tests.

SergeyKanzhelev (approved):

PR #23733, opened 2021-09-24
jobs.md: explain always_run field for postsubmits

listx:
Also add a short paragraph about what the different types of jobs mean
in a practical sense for those who are new to Prow.

### DIFF
--- a/prow/jobs.md
+++ b/prow/jobs.md
@@ -31,14 +31,32 @@ that you can define pods in yaml.  Please see kubernetes documentation
 for help here, for example the [Pod overview] and [PodSpec api
 reference].
 
-Periodic config looks like so (see [GoDocs](https://pkg.go.dev/k8s.io/test-infra/prow/config#Periodic) for complete config):
+### Types of jobs
+
+There are 3 basic types of jobs:
+
+1. periodics
+2. presubmits
+3. postsubmits
+
+They are distinguished by **when** they run. **Periodics** run at a regular
+interval, just like [UNIX cronjobs](https://en.wikipedia.org/wiki/Cron).
+**Presubmits** run against a PR before it gets merged (for a particular repo).
+Some examples are unit tests, integration tests, and linters. Finally,
+**postsubmits** run *after* a PR gets merged (for a particular repo). A common
+use case is a job that builds and publishes artifacts for later consumption by
+downstream users (e.g., a Docker image or binary file).
+
+#### Periodics
+
+Here is a sample **periodic** config (see [GoDocs](https://pkg.go.dev/k8s.io/test-infra/prow/config#Periodic) for complete config):
 
 ```yaml
 periodics:
 - name: foo-job         # Names need not be unique, but must match the regex ^[A-Za-z0-9-._]+$
   decorate: true        # Enable Pod Utility decoration. (see below)
   interval: 1h          # Anything that can be parsed by time.ParseDuration.
-  # Alternatively use a cron instead of an interval, for example:
+  # Alternatively use a cron format instead of an interval, for example:
   # cron: "05 15 * * 1-5"  # Run at 7:05 PST (15:05 UTC) every M-F
   extra_refs:            # Periodic job doesn't clone any repo by default, needs to be added explicitly
   - org: org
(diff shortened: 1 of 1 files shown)


fejta:
LGTM
/uncc

prow/jobs.md
```diff
@@ -93,11 +94,119 @@ command that reruns all jobs. If unspecified, the default configuration makes
...
-and [requiring](#requiring-job-statuses) GitHub status contexts. A useful
-pattern when adding new jobs is to start with `always_run` set to false and
-`skip_report` set to true. Test it out a few times by manually triggering,
-then switch `always_run` to true. Watch for a couple days, then switch
-`skip_report` to false.
+and [requiring](#requiring-job-statuses) GitHub status contexts.
+
+A useful pattern when adding new jobs is to start with `always_run` set to
+`false` and `skip_report` set to `true`. Test it out a few times by manually
+triggering; this manual triggering phase allows you to iron out any bugs without
```
chaodaiG, line 101: would be good to mention `/test` for manual triggering

prow/jobs.md
```diff
@@ -93,11 +94,119 @@ command that reruns all jobs. If unspecified, the default configuration makes
...
+Postsubmits are configured per-repo, because most of the time they are
+configured to run when a branch updates on a repo (due to a push or PR merge).
+If no `branches` are specified, then they will run against every branch.
+
+Postsubmits do not always have to depend on branch updates. The trick is to set
+`always_run` to `false` (and also not specify `run_if_changed` or
+`skip_if_only_changed` fields). Then the postsubmit will not respond to any
+branch updates, and instead may be triggered by a Pub/Sub message when it is
+intercepted by
+[Sub](https://github.com/kubernetes/test-infra/tree/master/prow/cmd/sub).
```
chaodaiG, line 132: can use relative path instead

prow/jobs.md
```diff
@@ -31,14 +31,32 @@ that you can define pods in yaml.  Please see kubernetes documentation
...
-Periodic config looks like so (see [GoDocs](https://pkg.go.dev/k8s.io/test-infra/prow/config#Periodic) for complete config):
+### Types of jobs
+
+There are 3 basic types of jobs:
+
+1. periodics
+2. presubmits
+3. postsubmits
+
+They are distinguished by **when** they run. **Periodics** run at a regular
```
cjwagner, line 42: Another important part of this concept is that presubmits and postsubmits not only trigger **when** refs are pushed, but they are actually intrinsically associated with a specific git ref (either a specific ref of a specific PR or a specific ref that has been pushed to a branch). By this I mean that those git refs are dynamically specified based on the triggering event and that Prow knows to handle the job appropriately for the given refs (e.g. running presubmits will be aborted when new commits are pushed to the PR, Tide will retest presubmits if and only if needed, Crier will report presubmits back to the correct PRs based on the refs, etc.)

Periodics on the other hand are not intrinsically associated with any git refs. Prow can check out source for any ProwJob including periodics using `extra_refs`, but the dynamically supplied refs are only applicable to presubmits and postsubmits.

prow/jobs.md
```diff
@@ -93,11 +94,119 @@ command that reruns all jobs. If unspecified, the default configuration makes
...
+    - ^main$
+    skip_branches:        # Regexps, do not run against these branches.
+    - ^release-.*$
+```
+
+Postsubmits are configured per-repo, because most of the time they are
+configured to run when a branch updates on a repo (due to a push or PR merge).
+If no `branches` are specified, then they will run against every branch.
+
+Postsubmits do not always have to depend on branch updates. The trick is to set
```
cjwagner, line 127: Its confusing that we are only mentioning this specifically for Postsubmits. We should move this or clarify that this is true of Presubmits as well.

cjwagner: This is all pretty specific to an uncommonly used pattern so it might be better to just briefly mention this feature here and then link to the full explanation in the Sub docs to avoid overwhelming users with a bunch of details that aren't applicable.

cjwagner (commented):
Thanks, Linus!

prow/jobs.md
listx: I think the ref association concept is too much for users to handle at this point in this doc. It's interesting info, and if it's not recorded anywhere else we should record it maybe near the end of this doc for now. WDYT?

cjwagner: I agree that what I shared is too much info, definitely would be overwhelming for a reader. But I do think there is an important piece in there that we should highlight. Maybe just explaining that presubmits and postsubmits automatically check out the git ref that triggered them? If you think that is already clear from the doc (or obvious based on name/context) feel free to ignore this comment, I just want to make sure we get the point across that presubmits and postsubmits have a dynamic git ref and periodics don't.

stevekuznetsov (commented):
Might be good to link/dedupe with https://github.com/kubernetes/test-infra/blob/master/prow/jobs.md#standard-triggering-and-execution-behavior-for-jobs

listx:
In light of the new prow docs site that is under construction (see https://github.com/kubernetes/test-infra/issues/24821 for umbrella issue), I'm going to close this. The content proposed/discussed here will finds its way to the new docs site once it's deployed to Netlify (see https://github.com/kubernetes/org/issues/3177).

PR #23735, opened 2021-09-24
feat: add ProtectedBranches config for branchcleaner to protect branches

### LINKED ISSUE #23732: Prow: branchcleaner should support filter with source branch
As [#15224-comment](https://github.com/kubernetes/test-infra/issues/15224#issuecomment-925854223) said, I wish prow could support filter with source branch, if the source branch is listed in the filter list, prow would not delete the source branch even it's merged.

**Why is this needed**:
For some git workflow strategy, we don't want to cleanup the merged branches. By adding the filter feature, it would leave some rooms for our users.

Colstuwjx:
Fixed #23732.

### DIFF
--- a/prow/plugins/config.go
+++ b/prow/plugins/config.go
@@ -67,6 +67,7 @@ type Configuration struct {
 	Blockades            []Blockade                   `json:"blockades,omitempty"`
 	Blunderbuss          Blunderbuss                  `json:"blunderbuss,omitempty"`
 	Bugzilla             Bugzilla                     `json:"bugzilla,omitempty"`
+	BranchCleaner        BranchCleaner                `json:"branch_cleaner,omitempty"`
 	Cat                  Cat                          `json:"cat,omitempty"`
 	CherryPickUnapproved CherryPickUnapproved         `json:"cherry_pick_unapproved,omitempty"`
 	ConfigUpdater        ConfigUpdater                `json:"config_updater,omitempty"`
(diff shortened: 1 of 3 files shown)


prow/plugins/config.go
```diff
@@ -1791,6 +1792,23 @@ func (b *Bugzilla) OptionsForRepo(org, repo string) map[string]BugzillaBranchOpt
 	return options
 }
 
+// BranchCleaner contains the configuration for the branchcleaner plugin.
+type BranchCleaner struct {
+	// ProtectedBranches is a set of repo branches, branches in this white list
```
petr-muller, line 1797: s/white/allow/

prow/plugins/config.go
```diff
@@ -1791,6 +1792,23 @@ func (b *Bugzilla) OptionsForRepo(org, repo string) map[string]BugzillaBranchOpt
...
+	// are exempt from branch gc, even if the branches are already merged into the target branch
+	ProtectedBranches []string `json:"protected_branches,omitempty"`
```
petr-muller, line 1799: I wonder if we could come up with a different name? "Branch Protection" is a GH term (https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/about-protected-branches), there is even a Prow component that manages it, and it would be nice to avoid confusion if possible.

petr-muller (commented):
/ok-to-test

One concern is that the PR implements a global set of branches, but branch names are per-repo  entities, so I think this config should allow per-repo (and possibly per-org) configuration.

prow/plugins/config.go
Colstuwjx: OK, so how about "AllowedBranches" ?

Colstuwjx:
> One concern is that the PR implements a global set of branches, but branch names are per-repo entities, so I think this config should allow per-repo (and possibly per-org) configuration.

OK, I'll change it.

prow/plugins/config.go
petr-muller: Could work, although it's a little generic. Maybe `PreserveBranches`?

PR #23656, opened 2021-09-17
images: add/update READMEs to trigger image builds

spiffxp:
Related:
- Part of: https://github.com/kubernetes/k8s.io/issues/1523
- Part of: https://github.com/kubernetes/test-infra/issues/13129
- Followup to: https://github.com/kubernetes/test-infra/pull/23616

PR #23737, opened 2021-09-24
config/jobs: use k8s-staging-test-infra images for some jobs

spiffxp:
Related:
- Followup to: https://github.com/kubernetes/test-infra/pull/23656

### DIFF
--- a/config/jobs/kubernetes/sig-scalability/sig-scalability-periodic-jobs.yaml
+++ b/config/jobs/kubernetes/sig-scalability/sig-scalability-periodic-jobs.yaml
@@ -904,7 +904,7 @@ periodics:
     path_alias: k8s.io/kubernetes
   spec:
     containers:
-    - image: gcr.io/k8s-testimages/benchmarkjunit:latest
+    - image: gcr.io/k8s-staging-test-infra/benchmarkjunit:latest
       command:
       - /benchmarkjunit
       args:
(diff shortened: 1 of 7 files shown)


config/jobs/kubernetes/sig-scalability/sig-scalability-periodic-jobs.yaml
```diff
@@ -904,7 +904,7 @@ periodics:
     path_alias: k8s.io/kubernetes
   spec:
     containers:
-    - image: gcr.io/k8s-testimages/benchmarkjunit:latest
+    - image: gcr.io/k8s-staging-test-infra/benchmarkjunit:latest
```
ameukam, line 907: There is not tag for this image. Do we still want to do the flip ?

spiffxp: Good catch, I'm looking into why the image is missing...

spiffxp: Opened https://github.com/kubernetes/test-infra/pull/23771 to address

I'm going to use `gcrane` to manually sync something over in the meantime

spiffxp: `gcrane copy gcr.io/{k8s-testimages,k8s-staging-test-infra}/benchmarkjunit:latest`

PR #23739, opened 2021-09-24
kops-grid: wire through the cloud parameter into grid tests

justinsb:
Otherwise we weren't actually testing on GCE!

### DIFF
--- a/config/jobs/kubernetes/kops/build_jobs.py
+++ b/config/jobs/kubernetes/kops/build_jobs.py
@@ -124,6 +124,7 @@ def build_test(cloud='aws',
     tmpl = jinja2.Environment(loader=loader).get_template(tmpl_file)
     job = tmpl.render(
         job_name=job_name,
+        cloud=cloud,
         cron=cron,
         kops_ssh_user=kops_ssh_user,
         create_args=args,
--- a/config/jobs/kubernetes/kops/kops-periodics-grid.yaml
+++ b/config/jobs/kubernetes/kops/kops-periodics-grid.yaml
@@ -33838,9 +33838,7 @@ periodics:
 - name: e2e-kops-grid-gce-u2004-k22-docker
   cron: '38 23 * * *'
   labels:
-    preset-service-account: "true"
-    preset-aws-ssh: "true"
-    preset-aws-credential: "true"
+    preset-k8s-ssh: "true"
   decorate: true
   decoration_config:
     timeout: 90m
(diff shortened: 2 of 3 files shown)


justinsb:
Still figuring out which other settings we're going to need, but this one seems pretty clear :-)

/assign @rifelpet

justinsb:
Never mind - thanks for doing this for me @rifelpet  :-)

rifelpet:
I noticed two other issues:
* The original GCE jobs are using the wrong feature flag: `GoogleCloudBucketACL` -> `GoogleCloudBucketAcl`
* The grid GCE jobs are not setting this flag at all.
I'm not sure if you want it set for some or all of the jobs but we should at least fix the feature flag if we intend for it to be enabled.

justinsb:
Thanks @rifelpet  - I think we also need also to set up a GCP ServiceAccount of some sort (see https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/e2e-kops-grid-gce-calico-u2004-k22-containerd/1441095428933160960).... I am proposing k8s-kops-test in #23740 as that is what the periodics use, but we could also try `preset-service-account: "true"` which is what I think the periodics use on AWS.  I'm guessing that k8s-kops-test has permissions to GCS buckets for the state store.

Re GoogleCloudBucketACL I think we should probably leave it unset and narrow down why it is needed.  I'm hoping that if we get secure node bootstrap running on GCE, we can also just stop the nodes accessing GCS at all, and then we can avoid the need for subpath permissions entirely.  But good call on the flag, though I propose we iterate on the grid tests for a bit and clean them all up.

rifelpet:
I'm confused because the original periodic jobs are working just fine: 

https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/e2e-kops-gce-stable/1441116041529790464

and they use the same presets as the grid jobs, but the grid is failing due to credential issues.

The final prow job definitions look effectively identical:

https://prow.k8s.io/prowjob?prowjob=8199565d-1c95-11ec-883b-c61df66c07e7

https://prow.k8s.io/prowjob?prowjob=f2d35d2b-1ca0-11ec-87ca-52c7297b323c

justinsb:
The `e2e-kops-gce-stable` does have `serviceAccountName: k8s-kops-test` .. I think that Kubernetes ServiceAccount has workload-identity (the GCP name for what we call IRSA) to a GCP ServiceAccount that has permissions to the buckets etc.

PR #23740, opened 2021-09-24
kops-grid: use k8s-kops-test serviceAccount for periodics

justinsb:
This is what we use in the presubmit tests on GCE.


justinsb:
This is in relation to the discussion here: https://github.com/kubernetes/test-infra/pull/23739#issuecomment-926699301

Not sure if we reached consensus there @rifelpet, but we should probably continue the discussion here  :-)

/assign @rifelpet

rifelpet (approved):

Issue #23741, opened 2021-09-24
Test failures starting 22 September due to network timeouts


tam7t:
**What happened**:

On our around 22 September we started seeing CI failures in prow:

Example log from error:

```
Step 4/8 : RUN apk add --no-cache curl &&     curl -LO https://storage.googleapis.com/kubernetes-release/release/${KUBE_VERSION}/bin/linux/${ARCH}/kubectl &&     chmod +x kubectl
 ---> Running in d7e388707d87
fetch https://dl-cdn.alpinelinux.org/alpine/v3.14/main/x86_64/APKINDEX.tar.gz
{"component":"entrypoint","file":"prow/entrypoint/run.go:165","func":"k8s.io/test-infra/prow/entrypoint.Options.ExecuteProcess","level":"error","msg":"Process did not finish before 25m0s timeout","severity":"error","time":"2021-09-23T21:50:51Z"}
```

Ref run: https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_secrets-s[…]e-csi-driver-image-scan/1441072726373568512/build-log.txt

**What you expected to happen**:

Successful builds.

**How to reproduce it (as minimally and precisely as possible)**:

This happens consistently on all of our builds. We've tried reverting commits in PRs and cannot find anything related to the [test case](https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes-sigs/secrets-store-csi-driver/secrets-store-csi-driver-config.yaml) that would cause this. The same tests running on the [`k8s-infra-prow-build` cluster](https://github.com/kubernetes/test-infra/pull/23729) succeed.

We have also seen it fail on `apt`:

```
  Connection timed out [IP: 151.101.194.132 80]
[91mE: Failed to fetch http://deb.debian.org/debian/pool/main/u/util-linux/libblkid1_2.36.1-8_amd64.deb  Connection timed out [IP: 199.232.126.132 80]
E: Failed to fetch http://security.debian.org/debian-security/pool/updates/main/o/openssl/libssl1.1_1.1.1k-1%2bdeb11u1_amd64.deb  Connection timed out [IP: 151.101.2.132 80]
E: Failed to fetch http://security.debian.org/debian-security/pool/updates/main/o/openssl/openssl_1.1.1k-1%2bdeb11u1_amd64.deb  Connection timed out [IP: 151.101.194.132 80]
E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?
[...]

PR #23743, opened 2021-09-24
secret-store-csi-driver: prepend iptables fixes for #23741

tam7t:
Attempt fix for #23741 based on initial success of https://github.com/kubernetes-sigs/secrets-store-csi-driver/pull/749


aramase (commented):
/lgtm

PR #23745, opened 2021-09-24
config/jobs: use make for kubernetes/org

spiffxp:
Related:
- Followup to: https://github.com/kubernetes/org/pull/3007


spiffxp:
/hold
Can't merge until https://github.com/kubernetes/org/pull/3007 has merged

PR #23746, opened 2021-09-24
chore: add securityContext.privileged=true for secrets-store-csi-driver

aramase:

`privileged: true` required for setting iptables rule


tam7t:
So this is needed for the `iptables` command, but do those test cases hit the error we see?

aramase:
> So this is needed for the iptables command, but do those test cases hit the error we see?

Based on the issue yesterday, the `iptables` fix is probably only required for docker in docker. But keeping it here for all other jobs to make sure this doesn't hit any other issue in future. WDYT?

The long term fix would be to move the iptables to kubekins at which point this iptable rule will be in every run.

PR #23751, opened 2021-09-25
Add more logging to plank

eddiezane:
This PR adds more logging to plank that helped with debugging https://github.com/kubernetes/k8s.io/issues/2559.


prow/plank/reconciler.go
```diff
@@ -102,6 +102,10 @@ func add(
 
 	r := newReconciler(ctx, mgr.GetClient(), overwriteReconcile, cfg, opener, totURL)
 	for buildCluster, buildClusterMgr := range buildMgrs {
+		r.log.WithFields(logrus.Fields{
```
eddiezane, line 105: This one isn't necessary but would be helpful in debugging a similar issue with kubeconfigs.

alvaroaleman (approved):
thanks!

PR #23753, opened 2021-09-25
added pull-kubernetes-node-e2e-containerd-kubetest2 job

Namanl2001:
Signed-off-by: Namanl2001 <namanlakhwani@gmail.com>

This PR is part of the CI migration from kubetest to kubetest2
https://github.com/kubernetes/enhancements/tree/master/keps/sig-testing/2464-kubetest2-ci-migration

Added new pull-kubernetes-node-e2e-containerd-kubetest2 job

xref: kubernetes-sigs/kubetest2#164

cc: @amwat @dims @spiffxp

### DIFF
--- a/config/jobs/kubernetes/sig-node/sig-node-presubmit.yaml
+++ b/config/jobs/kubernetes/sig-node/sig-node-presubmit.yaml
@@ -220,8 +220,8 @@ presubmits:
             cpu: 4
             memory: 6Gi
   - name: pull-kubernetes-node-e2e-containerd-kubetest2
-    always_run: true
-    optional: false
+    always_run: false
+    optional: true
     branches:
     - master
     decorate: true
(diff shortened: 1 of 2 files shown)


config/jobs/kubernetes/sig-node/sig-node-presubmit.yaml
```diff
@@ -219,6 +219,53 @@ presubmits:
           requests:
             cpu: 4
             memory: 6Gi
+  - name: pull-kubernetes-node-e2e-containerd-kubetest2
+    always_run: true
+    optional: false
```
spiffxp, line 224: /hold
Hang on, do we have sufficient data to suggest this is ready to be always run and required?  The KEP doesn't call out replacing the existing node e2e job as part of beta.  I'm OK with doing this if we give sufficient notice (e.g. let kubernetes-dev@ know), and actually replace the existing job.  I just wasn't aware that was the plan, and would like to see evidence that we're ready for this step
```suggestion
    always_run: false
    optional: true
```

Namanl2001: I've done this by looking at the [existing job](https://github.com/kubernetes/test-infra/blob/8d536b6b41ddf6f02d54a36ce15937207ab29fe6/config/jobs/kubernetes/sig-node/sig-node-presubmit.yaml#L188).

> do we have sufficient data to suggest this is ready to be always run and required?

Where should I look for the sufficient data? Thanks

amwat: Let's stick to gradual rollout like the other canary jobs.
so +1 to the suggested changes.

PR #23758, opened 2021-09-27
Remove helm e2e test from prow

rikatz:
We already run this test via gh actions:

https://github.com/kubernetes/ingress-nginx/blob/main/.github/workflows/ci.yaml#L164

So probably we can remove this from prow

/cc @strongjz @tao12345666333 @cpanato


tao12345666333 (approved):
/lgtm

PR #23770, opened 2021-09-27
sig-k8s-infra: Add cleanup job for disk snapshots

ameukam:
Related:
  - Ref: https://github.com/kubernetes/k8s.io/issues/2241
  - Part of https://github.com/kubernetes/k8s.io/issues/1469

Add a cleanup job that will delete disk snapshots older than 30 days created by sig-scalability periodic
job ci-golang-tip-k8s-1-18-canary.

### DIFF
--- a/config/jobs/kubernetes/sig-scalability/sig-scalability-cleanup.yaml
+++ b/config/jobs/kubernetes/sig-scalability/sig-scalability-cleanup.yaml
@@ -31,3 +31,51 @@ periodics:
       - ci-kubernetes-e2e-gci-gce-scalability
       # Delete snapshots older than x days
       - "30"
+
+- name: ci-kubernetes-scalability-cleanup-golang-builds-canary
+  cluster: k8s-infra-prow-build
+  interval: 24h
+  labels:
+    preset-service-account: "true"
+  decorate: true
+  decoration_config:
+    timeout: 60m
+  extra_refs:
+  - org: kubernetes
+    repo: perf-tests
+    base_ref: master
+    path_alias: k8s.io/perf-tests
+  annotations:
+    # TODO(ameukam): update with email alert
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes/sig-scalability/sig-scalability-cleanup.yaml
```diff
@@ -31,3 +31,49 @@ periodics:
...
+    base_ref: master
+    path_alias: k8s.io/perf-tests
+  annotations:
+    # TODO(ameukam): update with email alert
+    testgrid-dashboards: sig-k8s-infra-canaries
+    testgrid-tab-name: golang-build-5k-project-snapshots-cleanup-canary
+    testgrid-alert-email: ameukam@gmail.com
+    testgrid-num-failures-to-alert: '3'
+  spec:
+    serviceAccountName: boskos-janitor
```
ameukam, line 55: @spiffxp Is it ok to use this job or do I need to create a separate SA with custom role ?

spiffxp: The default service account should also have rights to do this

https://github.com/kubernetes/k8s.io/blob/main/infra/gcp/bash/prow/ensure-e2e-projects.sh#L71-L83

spiffxp: If you add a TODO that we should revisit this that's good enough for now for me

spiffxp (changes requested):

spiffxp:
There's a part of me that asks why the jobs aren't doing this by default, but if this is how it's already done I'm fine proceeding this way

spiffxp (approved):
/approve
/lgtm

spiffxp:
ref: https://github.com/kubernetes/k8s.io/issues/2854

PR #23771, opened 2021-09-27
config/jobs: fix trigger regex for benchmarkjunit-canary image push job

spiffxp:
Related:
- Followup to: https://github.com/kubernetes/test-infra/pull/23616
- Addresses issue discovered in: https://github.com/kubernetes/test-infra/pull/23737#discussion_r715647329

Images haven't been getting pushed to gcr.io/k8s-staging-test-infra/benchmarkjunit because the canary image push job was never triggering because of an incorrect `run_if_changed` regex leftover from copy-pasting

### DIFF
--- a/config/jobs/image-pushing/k8s-staging-test-infra.yaml
+++ b/config/jobs/image-pushing/k8s-staging-test-infra.yaml
@@ -29,7 +29,7 @@ postsubmits:
           - images/bazelbuild/
     - name: post-test-infra-push-benchmarkjunit-canary
       cluster: k8s-infra-prow-build-trusted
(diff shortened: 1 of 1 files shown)


cjwagner (approved):
I don't think this explains why the periodic job is failing though? https://testgrid.k8s.io/sig-testing-canaries#benchmark-demo

cjwagner:
> I don't think this explains why the periodic job is failing though? https://testgrid.k8s.io/sig-testing-canaries#benchmark-demo

Nevermind, I was confused that the job was failing even though the image existed since that appeared to be the problem. Now I see the image exists because you just added it a few minutes ago :+1:

PR #23772, opened 2021-09-27
cloud-provider-gcp: non-Bazel testing support

jiahuif:
This PR enables job `cloud-provider-gcp-tests` to run tests without Bazel while retaining the ability to do so with Bazel to be compatible with existing release branches.

ref: https://github.com/kubernetes/cloud-provider-gcp/pull/259


jpbetz:
/lgtm

Looking forward to this!

PR #23791, opened 2021-09-28
Update config for graceful node shutdown job

bobbypage:
* Remove explicitly setting `GracefulNodeShutdown` feature gate, since
  it's now enabled by defaulted as a beta feature as of 1.21
* Enable `DynamicKubeletConfig` feature gate, since the test needs to
  update kubelet config at runtime

### DIFF
--- a/config/jobs/kubernetes/sig-node/node-kubelet.yaml
+++ b/config/jobs/kubernetes/sig-node/node-kubelet.yaml
@@ -420,7 +420,8 @@ periodics:
           - --gcp-project-type=node-e2e-project
           - --gcp-zone=us-west1-b
           - --node-args=--image-config-file=/workspace/test-infra/jobs/e2e_node/image-config-serial.yaml
-          - --node-test-args=--feature-gates=GracefulNodeShutdown=true --kubelet-flags="--cgroups-per-qos=true --cgroup-root=/"
+          # Remove DynamicKubeletConfig feature gate once https://github.com/kubernetes/kubernetes/issues/105047 is completed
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes/sig-node/node-kubelet.yaml
```diff
@@ -420,7 +420,8 @@ periodics:
           - --gcp-project-type=node-e2e-project
           - --gcp-zone=us-west1-b
           - --node-args=--image-config-file=/workspace/test-infra/jobs/e2e_node/image-config-serial.yaml
-          - --node-test-args=--feature-gates=GracefulNodeShutdown=true --kubelet-flags="--cgroups-per-qos=true --cgroup-root=/"
+          # Remove DynamicKubeletConfig feature gate once https://github.com/kubernetes/kubernetes/issues/105047 is completed
+          - --node-test-args=--feature-gates=DynamicKubeletConfig=true --kubelet-flags="--cgroups-per-qos=true --cgroup-root=/"
```
wzshiming, line 424: Maybe we can modify the test so that it no longer depends on XXXX?

wzshiming, line 424: Maybe we can modify the test so that it no longer depends on DynamicKubeletConfig?

bobbypage: I think there is work ongoing for that already in general for node e2e tests, see https://github.com/kubernetes/kubernetes/pull/104153

That PR changes how `tempSetCurrentKubeletConfig` works to avoid relying on  `DynamicKubeletConfig`. So once that PR is merged and completed, we should be able to remove this.

SergeyKanzhelev (approved):

PR #23800, opened 2021-09-28
Gerrit reads projects configuration from global

chaodaiG:
Gerrit repos are currently defined as command line arguments, and duplicated on all components that interact with gerrit. This change moves this to prow central config, just like all other prow components


prow/gerrit/adapter/adapter.go
```diff
@@ -65,19 +68,57 @@ type LastSyncTracker interface {
...
 		prowJobClient:      prowJobClient,
 		config:             cfg,
-		gc:                 gc,
+		gc:                 gerritClient,
 		tracker:            lastSyncTracker,
-		projectsOptOutHelp: projs,
+		projectsOptOutHelp: projectsOptOutHelpMap,
 	}
+	go func() {
+		for {
```
cjwagner, line 104: I think this loop needs some sort of delay. We don't need to check this more than once a second.

prow/gerrit/client/client.go
```diff
@@ -252,6 +254,37 @@ func (c *Client) Authenticate(cookiefilePath, tokenPath string) {
 	}
 }
 
+// UpdateClients update gerrit clients with new instances map
+func (c *Client) UpdateClients(instances map[string][]string) error {
+	c.lock.Lock()
```
cjwagner, line 259: What exactly does this lock protect? It seems like it protects the `authentication` field and some fields of the handlers?

The `handlers` map is used without holding the lock elsewhere, but that seems no longer safe.

prow/crier/reporters/gerrit/reporter.go
```diff
@@ -92,12 +93,21 @@ type JobReport struct {
...
 		return nil, err
 	}
 	gc.Authenticate(cookiefilePath, "")
+	go func() {
+		for {
+			orgReposConfig := cfg().Gerrit.OrgReposConfig
+			if orgReposConfig == nil {
+				continue
+			}
+			gc.UpdateClients(orgReposConfig.AllRepos())
```
cjwagner, line 108: I think doing this constantly in a separate goroutine is still technically racy. We could alternatively make the gerrit client create new handlers for new repos when we find that it is missing from the same goroutine.

prow/gerrit/client/client.go
chaodaiG: As I understand this lock protects `handlers` and `accounts` map, as well as `authentication` function.

Agree that this is no longer thread safe, updated all usage of handlers to be thread safe

prow/gerrit/adapter/adapter.go
chaodaiG: good point, added a sleep

prow/crier/reporters/gerrit/reporter.go
chaodaiG: both `UpdateClients` and `Authenticate` above are protected by the same lock, I'm not seeing any race condition here.

cjwagner: The race I'm imagining could occur if the gerrit config is added at the same time a job that uses the config is added. Since the gerrit clients are added in a separate goroutine, we may try to handle the new job before the appropriate client has been loaded. This race is really unlikely, but unless creating the clients is expensive, it might be simpler and slightly less racy to grab a write lock and create the needed client if it is not available at the time of use without a background goroutine.  I don't think this is a big deal so feel free to keep this as is, just food for thought.

PR #23802, opened 2021-09-29
spyglass: add nil-pointer check

listx:
Sometimes the job may not have a PodSpec (it will be nil) because it
will not be using a Kubernetes agent.

/assign @cjwagner


cjwagner (approved):
Thanks Linus.

PR #23808, opened 2021-09-29
[bugzilla] Control the authorization method via flags

droslean:
/cc @petr-muller 

### DIFF
--- a/prow/bugzilla/client.go
+++ b/prow/bugzilla/client.go
@@ -206,6 +206,7 @@ type delegate struct {
 	endpoint                string
 	githubExternalTrackerId uint
 	getAPIKey               func() []byte
+	authMethod              string
 }
 
 // the client is a Client impl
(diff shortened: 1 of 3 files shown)


prow/bugzilla/client.go
```diff
@@ -686,15 +687,25 @@ func (c *client) GetComments(bugID int) ([]Comment, error) {
...
+			req.Header.Set("X-BUGZILLA-API-KEY", string(apiKey))
+		default:
+			// some BugZilla servers are too old and can't handle the header.
+			// some don't want the query parameter and some need Bearer Authorization instead.
+			// We can set all of them and keep
+			// everyone happy without negotiating on versions
+			req.Header.Set("X-BUGZILLA-API-KEY", string(apiKey))
+			values := req.URL.Query()
+			values.Add("api_key", string(apiKey))
+			req.URL.RawQuery = values.Encode()
```
petr-muller, line 707: I think I'd like to see one of the following:

1. I think the current approach is reasonable. The default behavior (using two methods union) is not achievable via setting the param explicitly, so we should explain why that is (so change the comment to say something like "use the union of `query` and `x-bugzilla-api-key` to mimic the previous default behavior which attempted to satisfy different BugZilla server versions").

or
 
2. Change the code to allow multiple choice (=make the option a set of strings, not a string).

droslean: Option 1. should make sense and keep things simple.  Updated. Thanks!

PR #23817, opened 2021-09-29
Deflake TestPjsToReport

chaodaiG:
Wrong type of interface as sorting option is silently ignored, using correct interface and it works now, verified by go test --count=50


alvaroaleman (approved):
wow, that's a subtle one

Issue #21296, opened 2021-03-10
Modify flakes-daily query output to look like flakes query output


spiffxp:
**What would you like to be added**:
The metrics query that outputs to http://storage.googleapis.com/k8s-metrics/flakes-daily-latest.json should also output test names, as is done for http://storage.googleapis.com/k8s-metrics/flakes-latest.json

I think if we could run this query more frequently than daily, but have it draw from the previous 24h, it could be more relevant/current when there's so much deflaking going on that flakes from the previous 7d are gone (this is a good problem to have!)

I don't care a ton about preserving the existing output schema? But maybe it's useful.

We could also consider listing more than top 10 flakes as well, depends on how noisy it'd look.

**Why is this needed**:
http://storage.googleapis.com/k8s-metrics/flakes-latest.json has a high enough lag time that it is effectively "stale" if top flakes are stomped out quickly, eg: https://github.com/kubernetes/kubernetes/pull/99908#issuecomment-795429814

As a result, people end up referring to https://go.k8s.io/triage as source of truth for "what's flaking worst now", but:
- that doesn't necessarily show flakes, more... failures that have happened > N times,
- so hopefully many failures means maybe a flake?

Or we have jobs like https://testgrid.k8s.io/sig-testing-canaries#ci-kubernetes-generate-make-test-count10 pop up.  
- Not saying it's a bad approach, let's stomp down on flakes using all available tools! 
- but... do we then want something like this for other jobs? 
- if we could lean really hard on something like flakes-daily.json, could we save ourselves the compute? etc.

spiffxp:
/sig testing
/sig release
because ci-signal
/priority important-soon
/milestone v1.21

/assign @MushuEE 
not to voluntold you, but you might be able to bust this out quicker than me

otherwise... I will get to this if I have time later this week

spiffxp:
/remove-lifecycle stale
Kettle is back up and running so this should be pretty doable

spiffxp:
Opened https://github.com/kubernetes/test-infra/pull/23825 to address

spiffxp:
/close
http://storage.googleapis.com/k8s-metrics/flakes-daily-latest.json now matches the format of http://storage.googleapis.com/k8s-metrics/flakes-latest.json

PR #23825, opened 2021-09-29
metrics: make flakes-daily metrics match flakes metrics, run more frequently

spiffxp:
Related:
- Part of: https://github.com/kubernetes/test-infra/issues/21296

This will make http://storage.googleapis.com/k8s-metrics/flakes-daily-latest.json look like http://storage.googleapis.com/k8s-metrics/flakes-latest.json but using jobs in the last 24h instead of the last 7d

I was tempted to bump the job to run every 1h but I'd like to understand which of this and the triage job are responsible for our new bigquery spend over in k8s-infra land first...


spiffxp:
/cc @liggitt @ehashman 
Since you've both expressed interest in kettle being down, here is a thing that kettle being back up will allow us to improve / start using again

Issue #23778, opened 2021-09-27
Decommission gke-project-pool


spiffxp:
**What should be cleaned up or changed**:
We shouldn't be using gke-specific projects to run prowjobs for kubernetes.  We have mostly stopped running jobs that require these, it's time to decommission the pool and identify whether jobs that still use these projects can use gce-project or should be running elsewhere.

**Provide any links for context**:
Related to:
- https://github.com/kubernetes/test-infra/issues/12863
- https://github.com/kubernetes/k8s.io/issues/752

/sig testing
/priority important-longterm
/area boskos
/area jobs
/milestone v1.23

spiffxp:
Inspecting output of `ripgrep -l gke config/jobs` I believe the remaining offending jobs are:
- ci-kubernetes-e2e-gke-canary
- application-periodic-default-gke
- ci-kustomize-periodic-default-gke
- pull-sig-storage-local-static-provisioner-e2e-gke
- ci-sig-storage-local-static-provisioner-master-gke-default

Jobs I'm less certain about:
- pull-kubetest2-gke-up-down-singlecluster
- pull-kubetest2-gke-up-down-multicluster

spiffxp:
Another way of surveying which jobs are actually using the projects
```bash
gcloud logging read \
  --project=k8s-prow-builds \
  --format=json \
  --limit=100 \
  --freshness=1d \
  '
  jsonPayload.component="boskos"
  jsonPayload.msg=~"Request for a free gke-project from"
  ' \
| jq -r 'map(.jsonPayload.msg)[]' \
| sort \
| uniq -c
```
```
  12 Request for a free gke-project from application-periodic-default-gke, dest busy
  48 Request for a free gke-project from ci-kubernetes-e2e-gke-canary, dest busy
  12 Request for a free gke-project from ci-kustomize-periodic-default-gke, dest busy
   6 Request for a free gke-project from ci-sig-storage-local-static-provisioner-master-gke-default-kubetest2, dest busy
```

spiffxp:
Jobs to decomission or replace with alternatives:
- [ ] ci-kubernetes-e2e-gke-canary: suggest removing, there is a gce alternative
- [ ] application-periodic-default-gke: https://github.com/kubernetes-sigs/application/issues/219
- [ ] ci-kustomize-periodic-default-gke: https://github.com/kubernetes-sigs/kustomize/issues/4214
[...]

PR #23826, opened 2021-09-29
config/jobs: rm ci-kustomize-periodic-default-gke

spiffxp:
Related:
- Part of: https://github.com/kubernetes/test-infra/issues/23778
- Fixes: https://github.com/kubernetes-sigs/kustomize/issues/4214

If this job truly depends on GKE-specifics, it shouldn't be running on prow.k8s.io (or it should be updated to not depend on GKE specifics)

It sounds like we have better coverage elsewhere for this subproject per https://github.com/kubernetes-sigs/kustomize/issues/4214#issuecomment-930428714 so I'm pre-emptively opening this PR to remove the job

/hold
For comment from SIG CLI


KnVerey:
/lgtm

> /hold
> For comment from SIG CLI

/unhold
I thnk the agreement of the three most active Kustomize owners/approvers should suffice! 😄

PR #23844, opened 2021-09-30
sig-scalability: migrate ci-kubernetes-kubemark-gce-scale to k8s-infra

ameukam:
Migrate to community-owned infrastructure: 
- ci-kubernetes-kubemark-gce-scale
- ci-kubernetes-kubemark-gce-scale-scheduler
- ci-kubernetes-kubemark-high-density-100-gce


ameukam:
cc @marseel @mborsz

marseel:
FTR: ci-kubernetes-e2e-gce-scale-performance and these jobs are running on the same project k8s-infra-e2e-scale-5k-project and they are overlapping in terms of time. 
I've checked that they have been green so I don't worry to much (canaries version when both were running), but I would like to have second opinion from @mborsz 

@ameukam I also wonder if we could use `scalability-project` pool instead of this project? I've checked quota and it seems that it's 600 cores, which a little bit smaller than 84 * 8 = 672 plus some masters' sizes etc

ameukam:
> FTR: ci-kubernetes-e2e-gce-scale-performance and these jobs are running on the same project k8s-infra-e2e-scale-5k-project and they are overlapping in terms of time. I've checked that they have been green so I don't worry to much (canaries version when both were running), but I would like to have second opinion from @mborsz
> 
> @ameukam I also wonder if we could use `scalability-project` pool instead of this project? I've checked quota and it seems that it's 600 cores, which a little bit smaller than required 84 * 8 = 672 plus some masters' sizes for these tests.

@marseel You can definitively use the `scalability-project` pool. Are we talking about all the `kubemark-*gce-*` or a specific one ?


> I don't think we should be running two massive tests in a single project at the same time. They may interfere in many different ways (GCE API quota is just a first example).

We already raised the quotas to ensure massive jobs can run in the same project: https://github.com/kubernetes/k8s.io/issues/2225

> Can we use different projects or change schedule?

@mborsz We can change the schedule but I don't know what would be the good adjustment (Also I can't measure the impact on the on-call team). Feel free to comment the PR or update the prowjobs with new schedules.

marseel:
FTR from discussion on slack with @ameukam 
We will have only two jobs running on `k8s-infra-e2e-scale-5k-project`: `ci-kubernetes-e2e-gce-scale-correctness` and `ci-kubernetes-e2e-gce-scale-performance`. We will also bump a bit quota on `scalability-project` to fit 5k kubemark tests.

Issue #23642, opened 2021-09-16
Remove non-alpha features from job ci-kubernetes-e2e-gci-gce-alpha-features


mmiranda96:
**What should be cleaned up or changed**:
Jobs [`ci-kubernetes-e2e-gci-gce-alpha-features`](https://github.com/kubernetes/test-infra/blob/c22abc29319ca84054c05b326abf41a722c10d49/config/jobs/kubernetes/sig-cloud-provider/gcp/gcp-gce.yaml#L584) and [`pull-kubernetes-e2e-gce-alpha-features`](https://github.com/kubernetes/test-infra/blob/c22abc29319ca84054c05b326abf41a722c10d49/config/jobs/kubernetes/sig-cloud-provider/gcp/gcp-gce.yaml#L375) run for several features which are no longer in alpha status. Cleaning up the features requires:
1. Updating the job configurations to no longer focus non-alpha features.
2. (Optional, up to discussion) Adding new alpha features which should be tested by these jobs.
3. Ensuring the job is running properly. 

**Provide any links for context**:
Here are all the features tested by both jobs, together with its status and KEP link:

Feature | Status | KEP Link
----|----|----
ServiceAccountIssuerDiscovery | Stable | https://github.com/kubernetes/enhancements/issues/1393
StorageVersionAPI | **Alpha** | https://github.com/kubernetes/enhancements/tree/master/keps/sig-api-machinery/2330-migrating-api-objects-to-latest-storage-version 
Audit | Stable | https://github.com/kubernetes/enhancements/issues/22
PodPreset | **Alpha** | https://github.com/kubernetes/enhancements/issues/162
RunAsGroup | Stable | https://github.com/kubernetes/enhancements/issues/213
TTLAfterFinished | Stable | https://github.com/kubernetes/enhancements/issues/592
NodeLease | Stable | https://github.com/kubernetes/enhancements/issues/589
CSIStorageCapacity | Beta | https://github.com/kubernetes/enhancements/issues/1472
GenericEphemeralVolume | Stable | https://github.com/kubernetes/enhancements/issues/1698
DaemonSetUpdateSurge | Beta | https://github.com/kubernetes/enhancements/issues/1591
CrossNamespacePodAffinity | Beta | https://github.com/kubernetes/enhancements/issues/2249

mmiranda96:
Additionally, we could try to onboard jobs listed on https://github.com/kubernetes/kubernetes/issues/104556

mmiranda96:
[...]

PR #23845, opened 2021-09-30
Remove non-alpha features from jobs.

rayandas:
This PR remove non-alpha features from job ci-kubernetes-e2e-gci-gce-alpha-features and pull-kubernetes-e2e-gce-alpha-features.
Partially fixes https://github.com/kubernetes/test-infra/issues/23642

Signed-off-by: Rayan Das <rayandas91@gmail.com>

### DIFF
--- a/config/jobs/kubernetes/sig-cloud-provider/gcp/gcp-gce.yaml
+++ b/config/jobs/kubernetes/sig-cloud-provider/gcp/gcp-gce.yaml
@@ -405,7 +405,7 @@ presubmits:
         - --provider=gce
         - --runtime-config=api/all=true
         - --stage=gs://kubernetes-release-pull/ci/pull-kubernetes-e2e-gce-alpha-features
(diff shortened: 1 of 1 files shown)


rayandas:
cc: @mmiranda96

mmiranda96:
Thanks for your work @rayandas ! Let's run these jobs and see the result.
/test pull-kubernetes-e2e-gce-alpha-features
/test ci-kubernetes-e2e-gci-gce-alpha-features

mmiranda96:
I think I can't trigger the CI job. Triggering just the PR job...
/test pull-kubernetes-e2e-gce-alpha-features

rayandas:
@mmiranda96 I think we cant trigger pull-kubernetes-e2e-gce-alpha-features job also. It has to be one from the bot's message above.

mmiranda96:
Yeah, I'm not sure why the job doesn't trigger. I'll ask about it on Slack.

spiffxp (approved):
/approve
/lgtm
FWIW I'm inclined to suggest we just merge this and watch the CI job to see if it breaks.  If it does, we can revert and take the more cautious route of duplicating the presubmit as a manually triggered job under a new name like pull-kubernetes-e2e-gce-alpha-features-canary

/hold
Please `/hold cancel` if you agree with this approach, or update the PR to take the more cautious approach first

mmiranda96:
Considering that this PR only makes the test set smaller, I see very low risk.
/lgtm
/hold cancel

Thanks everyone!

PR #23848, opened 2021-09-30
fix sig-testing janitor jobs

spiffxp:
Related:
- Part of: https://github.com/kubernetes/sig-testing/issues/11


cjwagner (approved):
FYI: I'm not previously familiar with these projects or the CI janitor jobs so it might be good to get another review if you think one is needed.
/hold

PR #23854, opened 2021-09-30
Print a sorted and unique lists on test help message

### LINKED ISSUE #23767: Prow robot should print unique trigger commands for when trigger and rerun_command apply to multiple jobs

I would like the list of `/test` commands to be unique when people use the following parameters in a way such that one command can trigger multiple jobs. This is for the use-case whereby people use "sliced" jobs by some index because the tests are split into buckets for parallelism. 

Here is an example where Amazon Athena tests are slow, and thus we run them on 6 pods at once to make them faster. 

Prow Job Snippet:

```
name: helltool-pr-test-xdialect-athena-0
  trigger: "/test helltool-pr-test-xdialect-athena"
  rerun_command: "/test helltool-pr-test-xdialect-athena"
...
name: helltool-pr-test-xdialect-athena-1
  trigger: "/test helltool-pr-test-xdialect-athena"
  rerun_command: "/test helltool-pr-test-xdialect-athena"
...
name: helltool-pr-test-xdialect-athena-2
  trigger: "/test helltool-pr-test-xdialect-athena"
  rerun_command: "/test helltool-pr-test-xdialect-athena"
...
```

This allows me to run all Athena tests by a single `/test helltool-pr-test-xdialect-athena`

How it shows up:

Screen Shot 2021-09-27 at 8 13 02 AM

**Why is this needed**:

It lists duplicate job names and makes the page longer to scroll through.

Here's the messaging on Prow's docs:

>The trigger is a regexp that matches the rerun_command. Users will be told to input the rerun_command when they want to rerun the job. Actually, anything that matches trigger will suffice. This is useful if you want to make one command that reruns all jobs. If unspecified, the default configuration makes /test <job-name> trigger the job.

So its clear that the ability to use a single command to run multiple jobs is an intentional use-case! Probably a pretty easy fix too

YuviGold:
Resolves #23767

### DIFF
--- a/prow/pjutil/filter.go
+++ b/prow/pjutil/filter.go
@@ -40,10 +40,15 @@ var OkToTestRe = regexp.MustCompile(`(?m)^/ok-to-test\s*$`)
 // 1. presubmits that can be run with '/test all' command.
 // 2. optional presubmits commands that can be run with their trigger, e.g. '/test job'
 // 3. required presubmits commands that can be run with their trigger, e.g. '/test job'
-func AvailablePresubmits(changes config.ChangedFilesProvider, org, repo, branch string, presubmits []config.Presubmit, logger *logrus.Entry) ([]string, []string, []string, error) {
+func AvailablePresubmits(changes config.ChangedFilesProvider, org, repo, branch string,
+	presubmits []config.Presubmit, logger *logrus.Entry) (sets.String, sets.String, sets.String, error) {
+	runWithTestAllNames := sets.NewString()
+	optionalJobTriggerCommands := sets.NewString()
+	requiredJobsTriggerCommands := sets.NewString()
+
 	runWithTestAll, err := FilterPresubmits(TestAllFilter(), changes, branch, presubmits, logger)
 	if err != nil {
-		return nil, nil, nil, err
+		return runWithTestAllNames, optionalJobTriggerCommands, requiredJobsTriggerCommands, err
 	}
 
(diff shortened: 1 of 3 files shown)


prow/pjutil/filter.go
```diff
@@ -40,10 +41,15 @@ var OkToTestRe = regexp.MustCompile(`(?m)^/ok-to-test\s*$`)
 // 1. presubmits that can be run with '/test all' command.
 // 2. optional presubmits commands that can be run with their trigger, e.g. '/test job'
 // 3. required presubmits commands that can be run with their trigger, e.g. '/test job'
-func AvailablePresubmits(changes config.ChangedFilesProvider, org, repo, branch string, presubmits []config.Presubmit, logger *logrus.Entry) ([]string, []string, []string, error) {
+func AvailablePresubmits(changes config.ChangedFilesProvider, org, repo, branch string,
+	presubmits []config.Presubmit, logger *logrus.Entry) (prowflagutil.Strings, prowflagutil.Strings, prowflagutil.Strings, error) {
```
fejta, line 45: Why is this returning flags?

YuviGold: I searched for a strings set wrapper in the source code and found `Strings` under `flagutil`. A set to support unique values.
I will replace to use `StringsSet` directly instead.

petr-muller (approved):

fejta (approved):

PR #23866, opened 2021-10-01
skipping [Driver: gcepd] for ci jobs

leiyiz:
when the `CSIMigrationGCE` flag is on, we should skip gcepd jobs since they will fail without a pd driver installed

### DIFF
--- a/config/jobs/kubernetes/sig-cloud-provider/gcp/gcp-gce.yaml
+++ b/config/jobs/kubernetes/sig-cloud-provider/gcp/gcp-gce.yaml
@@ -48,7 +48,7 @@ presubmits:
         # Panic if anything mutates a shared informer cache
         - --env=ENABLE_CACHE_MUTATION_DETECTOR=true
         - --stage=gs://kubernetes-release-pull/ci/pull-kubernetes-e2e-gce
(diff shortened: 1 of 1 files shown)


leiyiz:
@msau42

msau42:
Do we have the equivalent jobs running in pd csi or cloud provider gcp repos? I want to make sure we don't lose coverage for k8s 1.23.

leiyiz:
> Do we have the equivalent jobs running in pd csi or cloud provider gcp repos? I want to make sure we don't lose coverage for k8s 1.23.

we run the in-tree stuff for managed driver through pdcsi test grid migration section. However, not for non-managed driver. As for Cloud-provider-gcp I don't think so.

msau42:
Do we have jobs running against kubernetes head? I want to make sure before we disable the jobs we have csi and in-tree (with migration) coverage already running elsewhere.

As for non-managed csi driver, I think we could delete those configs from k/k and just rely on the pdcsi jobs.

msau42:
In-tree migration coverage is at https://k8s-testgrid.appspot.com/provider-gcp-compute-persistent-disk-csi-driver#Migration%20Kubernetes%20Master%20Driver%20Stable
CSI coverage is at https://k8s-testgrid.appspot.com/provider-gcp-compute-persistent-disk-csi-driver#Kubernetes%20Master%20Driver%20Latest

PR #23867, opened 2021-10-01
include containerd-extra-script in  cos-cgroupv2-containerd-e2e

Dragoncell:
containerd-extra-script is needed for  cos-cgroupv2-containerd-e2e as it to point to https://github.com/containerd/containerd/blob/main/test/e2e_node/gci-init.sh, so it will have logic to enable cgroupv2 on the instance.


bobbypage:
Do you mind updating description of why we need to include `containerd-extra-script`? i.e. we need to include it to point to https://github.com/containerd/containerd/blob/main/test/e2e_node/gci-init.sh so it will have logic to enable cgroupv2 on the instance. Thanks!

/lgtm

SergeyKanzhelev (approved):
thank you

PR #23882, opened 2021-10-04
OWNERS_ALIASES: update sig-testing-leads

spiffxp:
Related:
- Followup to: https://github.com/kubernetes/community/pull/6069


fejta (approved):
/hold

PR #23891, opened 2021-10-05
clonerefs: allow to clone with GitHub App authentication

zoncoen:
Ref. #10423
This PR enables `clonerefs` to clone by [GitHub App authentication](https://docs.github.com/en/developers/apps/building-github-apps/authenticating-with-github-apps#http-based-git-access-by-an-installation) just like Personal Access Token.


prow/pod-utils/clone/clone.go
```diff
@@ -144,10 +170,18 @@ func gitCtxForRefs(refs prowapi.Refs, baseDir string, env []string, oauthToken s
...
+		if user != "" {
+			u, _ := url.Parse(g.repositoryURI)
+			u.User = url.UserPassword(user, token)
+			g.repositoryURI = u.String()
+		} else {
+			// GitHub requires that the personal access token is set as a username.
+			// e.g., https://<token>:x-oauth-basic@github.com/owner/repo.git
+			u, _ := url.Parse(g.repositoryURI)
+			u.User = url.UserPassword(token, "x-oauth-basic")
+			g.repositoryURI = u.String()
```
petr-muller, line 183: The `u, _ := url.Parse(g.repositoryURI)` line can be before the conditional, and the `g.repositoryURI = u.String()` can be after it. Only the `UserPassword` is conditional.

petr-muller (commented):
Just skimmed but looks fine

prow/pod-utils/clone/clone.go
zoncoen: Thank you for your comment. I fixed it in https://github.com/kubernetes/test-infra/pull/23891/commits/390b0c94df6fbba204e772fc4d50648cb8c510f6.

zoncoen:
Thanks for your reply 🙏 
/unhold

PR #23902, opened 2021-10-05
Make ghcache log metrics for non-GET requests.

cjwagner:
/assign @chaodaiG

### DIFF
--- a/ghproxy/ghcache/coalesce.go
+++ b/ghproxy/ghcache/coalesce.go
@@ -22,7 +22,6 @@ import (
 	"net/http"
 	"net/http/httputil"
 	"strconv"
-	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


ghproxy/ghcache/coalesce.go
```diff
@@ -83,15 +82,13 @@ func (coalescer *requestCoalescer) RoundTrip(req *http.Request) (*http.Response,
...
-				tokenBudgetName = coalescer.hasher.Hash(req)
-			}
-			collectMetrics(ModeNoStore, req, resp, tokenBudgetName)
+		var tokenBudgetName string
+		if val := req.Header.Get(TokenBudgetIdentifierHeader); val != "" {
+			tokenBudgetName = val
+		} else {
+			tokenBudgetName = coalescer.hasher.Hash(req)
 		}
+		collectMetrics(ModeSkip, req, resp, tokenBudgetName)
```
chaodaiG, line 91: can you explain why not using "ModeNoStore" here? Would it be better to distinguish between graphql and other requests?

cjwagner: `NO-STORE` is actually a possible `Cache-Control` header value https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control.
I thought it would be better to properly distinguish a response avoiding the cache due to a `NO-STORE` response header or due to the cache not being applicable for the request type and thus being bypassed.

chaodaiG: ack

chaodaiG (approved):
/hold
In case you also want @alvaroaleman to take a look

PR #23926, opened 2021-10-07
e2e: disable auto-updates on all Fedora CoreOS jobs

lucab:
This tweaks nodes configurations in order to disable auto-updates
on all jobs running on Fedora CoreOS (FCOS).
An auto-update agent (Zincati) is active by default on all FCOS nodes,
but we don't actually want those machines to perform updates in the
middle of a CI run.
This avoids any possible race between the test-runner and the
machine rebooting itself and switching to a different OS version
during a test run.

Signed-off-by: Luca BRUNO <luca.bruno@coreos.com>

/cc @harche @haircommander @mrunalp

### DIFF
--- a/jobs/e2e_node/crio/crio.ign
+++ b/jobs/e2e_node/crio/crio.ign
@@ -7,6 +7,17 @@
       "systemd.unified_cgroup_hierarchy=0"
     ]
   },
+  "storage": {
+    "files": [
+      {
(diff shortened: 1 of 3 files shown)


harche:
Both cgroups v1 and v2 crio jobs pass fine with this change in local testing. 

/hold cancel

PR #23928, opened 2021-10-07
Bump k8s to 1.20 1.21 and 1.22 for Kuberbuilder e2e tests

hickeyma:
Also remove older k8s versions

**Note:** e2e test will not pass for Kubernetes 1.22 until https://github.com/kubernetes-sigs/kubebuilder/pull/2371 is merged. That PR is waiting on this PR to be merged first.

Signed-off-by: Martin Hickey <martin.hickey@ie.ibm.com>

### DIFF
--- a/config/jobs/kubernetes-sigs/kubebuilder/kubebuilder-presubmits.yaml
+++ b/config/jobs/kubernetes-sigs/kubebuilder/kubebuilder-presubmits.yaml
@@ -19,7 +19,7 @@ presubmits:
     annotations:
       testgrid-dashboards: sig-api-machinery-kubebuilder
       testgrid-tab-name: kubebuilder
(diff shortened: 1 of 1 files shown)


camilamacedo86:
Hi @hickeyma, 

We mandatory need to ensure support for the latest 3 k8s versions. That does not mean that we cannot do tests with the old versions as well at all. See: https://kubernetes.io/releases/

We should be testing with the latest 3 ones but we do not need to remove the tests with the previous versions. Unless we check that do all these tests began to impact we contribute with the project because it is taking too long. IHMO that is not the case. So, we can only, for now, add the tests for what we are missing and after if/when required decide to remove the tests with the older versions. (IMHO) WDYT? Would that make sense?

c/c @estroz

PR #23935, opened 2021-10-07
Set the Kubebuilder k8s versions to kind k8s versions

hickeyma:
Follow up on #23928. The k8s versions need to be aligned to k8s version supported by Kind as Kubebuilder uses Kind for testing.

Signed-off-by: Martin Hickey <martin.hickey@ie.ibm.com>


hickeyma:
Thanks @dims

PR #23936, opened 2021-10-07
config/jobs: update 1.22 pull-kubernetes-integration to match ci

spiffxp:
Related:
- Part of: https://github.com/kubernetes/kubernetes/issues/105436
- Followup to: https://github.com/kubernetes/test-infra/pull/23932

Remove the pull-kubernetes-integration-1-22-canary job and migrate its config changes over to the pull-kubernetes-integration job for the release-1.22 branch. This should unblock 1.22 merges

Instead of explaining why with a comment, add an annotation with a link to the github issue/comment that led us to change the job config. This is because the job config gets stripped of comments by config-rotator, while the annotation will persist.

---

I'll summarize https://github.com/kubernetes/kubernetes/issues/105436#issuecomment-937275902 and https://github.com/kubernetes/kubernetes/issues/105436#issuecomment-937805327 here:

We think the recently improved I/O performance of the build cluster (ref: https://github.com/kubernetes/k8s.io/issues/1187#issuecomment-929466293) is causing integration tests to run faster, thus triggering etcd/apiserver connection issues that were never fully root caused during 1.22 (ref: https://github.com/kubernetes/kubernetes/issues/103512).

We think we're not seeing this in master because it's dropped a bunch of v1beta1 tests, thus below whatever threshold triggers these issues.  We think we're not seeing this in v1.22 CI because CI jobs run additional slow/expensive tests that may be spreading things out enough to avoid triggering the issues.

We're opting to make the presubmit match the CI job config for release-1.22 _only_ so as to unblock cherry picks destined for patch release.  Ultimately https://github.com/kubernetes/kubernetes/issues/103512 should be root caused.

### DIFF
--- a/config/jobs/kubernetes/sig-release/release-branch-jobs/1.22.yaml
+++ b/config/jobs/kubernetes/sig-release/release-branch-jobs/1.22.yaml
@@ -1738,20 +1738,18 @@ presubmits:
           requests:
             cpu: 1500m
             memory: 7Gi
-  # TODO(spiffxp): remove when finished investigating https://github.com/kubernetes/kubernetes/issues/105436
-  - always_run: false
+  - always_run: true
     annotations:
-      testgrid-create-test-group: "true"
-    optional: true
+      job-config-differs-because: "https://github.com/kubernetes/kubernetes/issues/105436#issuecomment-937805327"
     branches:
     - release-1.22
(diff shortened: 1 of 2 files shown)


aojea:
/hold cancel

is not likely we can find a better solution soon and the queue for cherry-picks starts to grow

PR #23944, opened 2021-10-07
Sub also supports authenticating with GH apps

chaodaiG:
Sub currently already supports authenticating with GH apps, but doesn't really allow co-existence of sub deployment with PAT token and GH app. This change allows using github-enabled-org, github-disabled-org etc for determining whether the pubsub message should be handled by PAT token or GH app.

On top of this, periodics sometimes doesn't have org/repo associated, so introduce another flag is-github-default-handler to make sure that this is only handled by the default deployment


chaodaiG:
appears that we might not need this, close for now
/close

PR #23978, opened 2021-10-12
Added entry in testgrid for v1.22 conformance results of s390x

guirish:
Added entry in testgrid for v1.22 conformance results of s390x


guirish:
@spiffxp Kindly add /lgtm label again as I updated the author field in commit as CLA was failing earlier.

PR #24022, opened 2021-10-15
Export GH Endpoint option

emilvberglind:
Need to be able to access the current GH endpoint externally, so exporting this option.


emilvberglind:
In hindsight, don't actually need this and will handle it differently.

PR #24064, opened 2021-10-20
WIP: Move eviction job to a swap environment.

mmiranda96:
Use a swap environment for eviction tests.

In order to run eviction tests with swap enabled, we need a swap environment. These are the main reasons to use the same job for existing eviction tests and future eviction-on-swap tests:
- They cover the same logic.
- Running non-swap test under such environment is ok, since access to swap is limited by a featureGate.
- We can reduce the number of jobs to be maintained and monitored.

Currently swap is only configured on Ubuntu, I still need to figure out if configuring it on COS is the same or if it requires some additional setup.

This work is related to https://github.com/kubernetes/kubernetes/issues/105023

/cc @ehashman @endocrimes
/kind feature
/priority important-soon
/sig node

### DIFF
--- a/config/jobs/kubernetes/sig-node/node-kubelet.yaml
+++ b/config/jobs/kubernetes/sig-node/node-kubelet.yaml
@@ -239,7 +239,7 @@ periodics:
       - --deployment=node
       - --gcp-project-type=node-e2e-project
       - --gcp-zone=us-west1-b
-      - --node-args=--image-config-file=/workspace/test-infra/jobs/e2e_node/image-config-serial.yaml
+      - --node-args=--image-config-file=/workspace/test-infra/jobs/e2e_node/image-config-serial-swap.yaml
       - --node-test-args=--feature-gates=LocalStorageCapacityIsolation=true --kubelet-flags="--cgroups-per-qos=true --cgroup-root=/"
       - --node-tests=true
       - --provider=gce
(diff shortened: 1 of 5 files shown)


config/jobs/kubernetes/sig-node/node-kubelet.yaml
```diff
@@ -264,7 +264,7 @@ periodics:
       - --deployment=node
       - --gcp-project-type=node-e2e-project
       - --gcp-zone=us-west1-b
-      - --node-args=--image-config-file=/workspace/test-infra/jobs/e2e_node/image-config-serial.yaml
+      - --node-args=--image-config-file=/workspace/test-infra/jobs/e2e_node/image-config-serial-swap.yaml
```
ehashman, line 267: Any reason to not reuse https://github.com/kubernetes/test-infra/blob/cc714da33d7ba85672aa4c7f58e0b3993155176d/jobs/e2e_node/swap/image-config-swap.yaml ?

ehashman (commented):
> Currently swap is only configured on Ubuntu, I still need to figure out if configuring it on COS is the same or if it requires some additional setup.

We also have it available on Fedora nodes.

Please rebase!

config/jobs/kubernetes/sig-node/node-kubelet.yaml
mmiranda96: It looks like I replaced the config on all jobs using it, so there's no point on having two separate ones. Merging them...

ehashman:
Closing this as a nudge to rebase... please reopen when ready for review!

/close

Issue #23915, opened 2021-10-06
Fix or delete containerd 1.4 canary jobs


mmiranda96:
**What should be cleaned up or changed**:
Some of the contained 1.4 canary jobs have been failing for some time now. Considering that we're working with containerd 1.5 (canaries will be provided by https://github.com/kubernetes/test-infra/pull/23215), it might make sense to either fix or delete containerd 1.4 canary jobs.

**Provide any links for context**:
This are the relevant links for the jobs:
Job name | Source code | TestGrid
----|----|----
`ci-containerd-build-1-4` | [sig-node-containerd.yaml#L33](https://github.com/kubernetes/test-infra/blob/3011f1d951e254c8484e3ab3b5c35480b9f03440/config/jobs/kubernetes/sig-k8s-infra/canaries/sig-node-containerd.yaml#L33) | [sig-node-containerd#containerd-build-1.4](https://testgrid.k8s.io/sig-node-containerd#containerd-build-1.4)
`ci-containerd-node-e2e-1-4` | [sig-node-containerd.yaml#L240](https://github.com/kubernetes/test-infra/blob/3011f1d951e254c8484e3ab3b5c35480b9f03440/config/jobs/kubernetes/sig-k8s-infra/canaries/sig-node-containerd.yaml#L240) | [sig-node-containerd#containerd-node-e2e-1.4](https://testgrid.k8s.io/sig-node-containerd#containerd-node-e2e-1.4)
`ci-containerd-node-e2e-features-1-4` | [sig-node-containerd.yaml#L316](https://github.com/kubernetes/test-infra/blob/3011f1d951e254c8484e3ab3b5c35480b9f03440/config/jobs/kubernetes/sig-k8s-infra/canaries/sig-node-containerd.yaml#L316) | [sig-node-containerd#containerd-node-e2e-features-1.4](https://testgrid.k8s.io/sig-node-containerd#containerd-node-e2e-features-1.4)

/sig node
/sig k8s-infra
/area config
/area jobs
/cc spiffxp
/assign @mmiranda96

mmiranda96:
Additionally, we might need to delete duplicate canaries under [config/jobs/kubernetes/sig-k8s-infra/canaries/sig-node-containerd.yaml](https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/sig-k8s-infra/canaries/sig-node-containerd.yaml).

dims:
+1 from me :) 

/cc @adisky

[...]

PR #24067, opened 2021-10-20
Fix containerd 1.4 node e2e

adisky:
- Update the k8s version for the job to 1.22 to avoid go error
- remove containerd 1.3 config files

May fix: https://github.com/kubernetes/test-infra/issues/23915
Signed-off-by: Aditi Sharma <adi.sky17@gmail.com>


endocrimes:
bumping to 1.22 seems reasonable to me, but I'm not sure how we've historically handled this

mmiranda96:
Based on the error logs from the 1.4 canaries, I don't think this is the cause. But we need the bump and removal of 1.3 anyway.

/approve

PR #24111, opened 2021-10-25
enabling blockade for sig docs blogs

RinkiyaKeDad:

Enabling the blockade plugin which will label PRs changing blogs that are more than 2 years old with `do-not-merge/blocked-paths`.

/cc @sftim @mrbobbytables


sftim:
I still don't think this is the right approach. Already, we are into a situation where posts older than 2 years are _not_ covered by this blockade configuration.

RinkiyaKeDad:
That's something we won't be able to fix by changing the regex, right @sftim? Should I close the PR then?

PR #24168, opened 2021-10-28
job_states_to_report should not be a required field

xrstf:
The new CRDs marked `job_states_to_report` to be a required field to ProwJobs. This however leads to jobs generated by mkpj not being applyable anymore, if the job definition does not include `job_states_to_report` (but only the Slack channel), for example a job with

```yaml
      reporter_config:
        slack:
          channel: dev-something
```

would lead to a ProwJob which reads

```yaml
  reporter_config:
    slack:
      channel: dev-something
      job_states_to_report: null
```

Causing 

> error: error validating "foo.yaml": error validating data: ValidationError(ProwJob.spec.reporter_config.slack): missing required field "job_states_to_report" in io.k8s.prow.v1.ProwJob.spec.reporter_config.slack; if you choose to ignore these errors, turn validation off with --validate=false

Prow merges the job-supplied reporter config later with the global configuration, so having no explicit `job_states_to_report` is absolutely fine.

### DIFF
--- a/config/prow/cluster/prowjob_customresourcedefinition.yaml
+++ b/config/prow/cluster/prowjob_customresourcedefinition.yaml
@@ -19955,8 +19955,6 @@ spec:
                         type: array
                       report_template:
                         type: string
-                    required:
-                    - job_states_to_report
                     type: object
                 type: object
               rerun_auth_config:
--- a/prow/apis/prowjobs/v1/types.go
+++ b/prow/apis/prowjobs/v1/types.go
@@ -305,7 +305,7 @@ type ReporterConfig struct {
 type SlackReporterConfig struct {
 	Host              string         `json:"host,omitempty"`
 	Channel           string         `json:"channel,omitempty"`
-	JobStatesToReport []ProwJobState `json:"job_states_to_report"`
+	JobStatesToReport []ProwJobState `json:"job_states_to_report,omitempty"`
 	ReportTemplate    string         `json:"report_template,omitempty"`
 }
 
--- a/prow/config/jobs_test.go
+++ b/prow/config/jobs_test.go
@@ -81,13 +81,27 @@ func TestReporterConfigRoundtrip(t *testing.T) {
 		{
 			content: `name: abc
 reporter_config:
-  slack:
-    job_states_to_report: []`,
+  slack: {}`,
 			expectedPJ: JobBase{
 				Name: "abc",
 				ReporterConfig: &prowapi.ReporterConfig{
 					Slack: &prowapi.SlackReporterConfig{
-						JobStatesToReport: []prowapi.ProwJobState{},
+						JobStatesToReport: nil,
+					},
+				},
+			},
+		},
+		{
+			content: `name: abc
+reporter_config:
+  slack:
+    job_states_to_report:
+    - test`,
+			expectedPJ: JobBase{
+				Name: "abc",
+				ReporterConfig: &prowapi.ReporterConfig{
+					Slack: &prowapi.SlackReporterConfig{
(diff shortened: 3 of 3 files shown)


fejta:
> a better way to do this

A gross but plausible way would be to treat `["REALLY_WEIRD_CONST"]` as equivalent to `[]`

fejta (approved):

chaodaiG:
> > a better way to do this
> 
> A gross but plausible way would be to treat `["REALLY_WEIRD_CONST"]` as equivalent to `[]`

While this works, I also agree that it's gross.

Since the entire idea of `job_states_to_report: []` is for opting out of reporting, an alternative idea is probably introducing a new field `skipReporting: <BOOL>`, which default to be false, and user who used to set `job_states_to_repot: []` can migrate over to this boolean

chaodaiG:
> Since the entire idea of `job_states_to_report: []` is for opting out of reporting, an alternative idea is probably introducing a new field `skipReporting: <BOOL>`, which default to be false, and user who used to set `job_states_to_report: []` can migrate over to this boolean.

@xrstf , @alvaroaleman , @fejta , what do you all think?

For either this idea or the idea of magic constant as Erick suggested in https://github.com/kubernetes/test-infra/pull/24168#issuecomment-961310937, it'll need to happen ahead of this PR to avoid regression. I can implement this idea of `skipReporting` and handle the migration if there is no objection from you.

alvaroaleman:
a `skipReporting` field seems a great idea but we will need to add defaulting to it based on if len($jobconfig.slack.states_to_report == 0), right? Maybe `report: true` is better as it avoids serializing anything into the kube api when the feature is unused?

chaodaiG:
I think defaulting this bool based on length is a great idea, which can prevent any regression. However, if we can already have `len(jobConfig.Slack.StatesToReport) == 0`, then we don't really need this bool. The problem here is that when hitting `Default`, the value of `StatesToReport` is always `nil` no matter it's defined as `nil` or `[]` when `omitempty` is specified. So this might not work?

Just to make sure that I understand correctly, if we do:
```
SkipReporting    bool         `json:"skip_reporting,omitempty"`
```
and the job config doesn't have `skip_reporting`, do you mean that `skip_reporting: false` is still passed through to kube api?

alvaroaleman:
> I think defaulting this bool based on length is a great idea, which can prevent any regression. However, if we can already have len(jobConfig.Slack.StatesToReport) == 0, then we don't really need this bool. The problem here is that when hitting Default, the value of StatesToReport is always nil no matter it's defined as nil or [] when omitempty is specified. So this might not work?

True, let me improve the wording: "Find a way to make this change compatible with existing config".

> do you mean that skip_reporting: false is still passed through to kube api?

No, it won't. What I mean is that there are presumably a lot more jobs that do not use the slack reporter and hence have `skip_reporting: true` which will get serialized. So by using a `report:` field instead, we will likely put less data in the kube apiserver.

chaodaiG:
I think I see your point now, and yeah it would be nice if we can find a way to go from input:
```
reporter_config:
    slack:
      job_states_to_report: []
```
to output:
```
reporter_config:
    slack:
      job_states_to_report: # nil or [], doesn't matter
      report: false
```

I think it's doable

chaodaiG:
Created a draft PR https://github.com/kubernetes/test-infra/pull/24280 , not super excited with where the logic is inserted. @alvaroaleman please take a look

chaodaiG:
https://github.com/kubernetes/test-infra/pull/24280 was merged, which accomplished the same purpose with this PR. I think we can close this PR now.

@xrstf , #24280 also updated prowjob CRD, could you try again?

chaodaiG:
this was fulfilled by #24280
/close

PR #24178, opened 2021-10-29
Gerrit client ensure authenticated at startup

chaodaiG:
Gerrit clients authentication is per host based, the authentication is setup once at startup, and updated only when the secret token value is updated. This change switches the order to ensure that hosts are authenticated at startup

### DIFF
--- a/prow/gerrit/adapter/adapter.go
+++ b/prow/gerrit/adapter/adapter.go
@@ -91,8 +91,6 @@ func NewController(ctx context.Context, prowJobClient prowv1.ProwJobInterface, o
 	if err != nil {
 		logrus.WithError(err).Fatal("Error creating gerrit client.")
 	}
-	gerritClient.Authenticate(cookiefilePath, tokenPathOverride)
-
 	c := &Controller{
 		prowJobClient:      prowJobClient,
 		config:             cfg,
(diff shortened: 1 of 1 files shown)


prow/gerrit/adapter/adapter.go
```diff
@@ -91,38 +91,52 @@ func NewController(ctx context.Context, prowJobClient prowv1.ProwJobInterface, o
...
 		tracker:            lastSyncTracker,
 		projectsOptOutHelp: projectsOptOutHelpMap,
 	}
+
+	// applyGlobalConfig must happen before authentication, otherwise initially
+	// the gerritClient doesn't contain any repos, and nothing is authenticated.
+	// The async goroutine for authentication only does something in real when httpcookiefile
+	// is updated.
+	c.applyGlobalConfig(cfg, gerritClient, lastSyncTracker)
+	gerritClient.Authenticate(cookiefilePath, tokenPathOverride)
```
cjwagner, line 107: Don't we need to call this after every call to `gerritClient.UpdateClients()` then?

chaodaiG: that's a good point, that would result in multiple goroutines doing async authentication though, will think about how to do this better

chaodaiG: actually I read it wrong, `Authenticate` is smart enough to only spin off goroutine once. Updated

chaodaiG:

prow/gerrit/adapter/adapter.go
```diff
@@ -91,38 +91,54 @@ func NewController(ctx context.Context, prowJobClient prowv1.ProwJobInterface, o
...
+}
+
+func (c *Controller) applyGlobalConfig(cfg config.Getter, gerritClient *client.Client, lastSyncTracker *syncTime, cookiefilePath, tokenPathOverride string) {
+	c.applyGlobalConfigOnce(cfg, gerritClient, lastSyncTracker, cookiefilePath, tokenPathOverride)
+
 	go func() {
 		for {
-			orgReposConfig := cfg().Gerrit.OrgReposConfig
-			if orgReposConfig == nil {
-				time.Sleep(time.Second)
```
cjwagner, line 107: Rather than checking every second and always processing if any config exists we can subscribe to config change notifications from the config agent: https://github.com/kubernetes/test-infra/blob/b5a1578a0c653b9d68a3c78ff7bd9b38f87a063f/prow/config/agent.go#L376-L384

prow/gerrit/adapter/adapter.go
```diff
@@ -91,38 +91,54 @@ func NewController(ctx context.Context, prowJobClient prowv1.ProwJobInterface, o
...
+		logrus.WithError(err).Error("Syncing states.")
+	}
+
+	c.lock.Lock()
+	// Updates a map, lock to make sure it's thread safe.
+	c.projectsOptOutHelp = orgReposConfig.OptOutHelpRepos()
+	// Authenticate creates a goroutine for rotating token secrets when called the first
+	// time, afterwards it only authenticate once.
+	gerritClient.Authenticate(cookiefilePath, tokenPathOverride)
+	c.lock.Unlock()
```
cjwagner, line 141: I think we can release this lock before authenticating.

chaodaiG: agreed, done

prow/gerrit/adapter/adapter.go
chaodaiG: I'll take a look, the usage of this function is really low though

cjwagner (approved):

cjwagner:
We failed to query changes across all instances after updating from `v20211027-fcb86429c4` to `v20211101-eeb8ee3f4c`. This PR seems most suspicious.

PR #24179, opened 2021-10-29
Create a way to dump parts of configmaps to metadata.json

WillMorrison:


amwat (commented):
/lgtm
/approve

amwat:
uhhh, we might need to revert it to be go 1.15 or (1.13 even) compatible
since the kubekins-e2e-1.18 and 1.19 variants are still using older go. 

https://github.com/kubernetes/test-infra/blob/master/images/kubekins-e2e/variants.yaml#L55

and seems like the current way kubetest binaries are being built into the image is at runtime (of the image building with whatever go version it has) as opposed to being copied into the image (after building it with the latest go version) (https://github.com/kubernetes/test-infra/blob/c5cdf1865f88d9bd2066d7e43f6c77ee36dd9fad/images/kubekins-e2e/cloudbuild.yaml#L4-L8)

https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-test-infra-push-kubekins-e2e/1456007921434365952#1:build-log.txt%3A119

`kubetest/gke.go:528:12: undefined: os.WriteFile`

WillMorrison:
> uhhh, we might need to revert it to be go 1.15 or (1.13 even) compatible since the kubekins-e2e-1.18 and 1.19 variants are still using older go.
> 
> `kubetest/gke.go:528:12: undefined: os.WriteFile`

It looks like the older go option is using `ioutil.WriteFile` instead.

PR #24194, opened 2021-11-01
Go deps updated by make rules

chaodaiG:
Note that after this change go deps will still need to be verified by bazel rules, as prow images pushing still requires bazel rules to be correct.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -50,6 +50,12 @@ clean:
 # gofmt
 #gofmt:
 #	hack/make-rules/update/gofmt.sh
+.PHONY: update-go-deps
+update-go-deps:
+	hack/make-rules/update/go-deps.sh
+.PHONY: verify-go-deps
+verify-go-deps:
+	hack/make-rules/verify/go-deps.sh
 ################################################################################
 # ================================== Linting ===================================
 # run linters, ensure generated code, etc.
--- a/hack/make-rules/verify/go-deps.sh
+++ b/hack/make-rules/verify/go-deps.sh
@@ -1,0 +1,87 @@
+#!/usr/bin/env bash
+# Copyright 2021 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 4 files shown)


chaodaiG:
hmm repo-infra uses go1.15.5 https://github.com/kubernetes/repo-infra/blob/5597dcdfb69ef4f4c7d994c2176034b9d6aa4721/repos.bzl#L33

chaodaiG:
this repo uses go1.16.2 https://github.com/kubernetes/test-infra/blob/98c428a5d7871ffdf7ed2e969f3904db0839efe3/WORKSPACE#L22

chaodaiG:
figured why test failed, go.sum was modified by gazelle https://github.com/kubernetes/repo-infra/blob/c1ad008c2a14f82268356aa493a58f541b9c2025/hack/update-deps.sh#L87

hack/make-rules/verify/go-deps.sh
```diff
@@ -0,0 +1,80 @@
...
+# exit trap cleanup for TMP_REPO
+cleanup() {
+  if [[ -n "${TMP_REPO}" ]]; then
+    rm -rf "${TMP_REPO}"
+  fi
+}
+
+# copies repo into a temp root saved to TMP_REPO
+make_temp_repo_copy() {
+  rm -f bazel-*
```
BenTheElder, line 42: seems a bit heavy-handed?

chaodaiG:
> figured why test failed, go.sum was modified by gazelle https://github.com/kubernetes/repo-infra/blob/c1ad008c2a14f82268356aa493a58f541b9c2025/hack/update-deps.sh#L87

Gazelle conflicts with go mod, this PR won't pass unless run gazelle as part of this PR. alternatively can do this as the last step of migration

Makefile
```diff
@@ -74,5 +74,9 @@ update-spelling:
 	hack/make-rules/update/misspell.sh
 verify-spelling:
 	hack/make-rules/verify/misspell.sh
+update-go-deps:
+	hack/make-rules/update/go-deps.sh
+verify-go-deps:
+	hack/make-rules/verify/go-deps.sh
 #################################################################################
-.PHONY: unit test verify go-lint update-gofmt verify-gofmt update-spelling verify-spelling
+.PHONY: unit test verify go-lint update-gofmt verify-gofmt update-file-perms verify-file-perms update-spelling verify-spelling update-go-deps verify-go-deps
```
BenTheElder, line 82: BTW: it's also possible to split these up to avoid rebase hell (stick .PHONY: foo alongside foo:) but under normal circumstances I wouldn't expect this to change as often anyhow (after we finish migrating)

chaodaiG: good point, will do that if I ever need to add more

BenTheElder (approved):
/lgtm
/approve

PR #24201, opened 2021-11-02
Made changes to export kubemark logs to GCS path

NikhilSharmaWe:
This PR is to fix https://github.com/kubernetes/kubernetes/issues/105927 issue of kubernetes/kubernetes repo, for exporting kubemark logs to GCS Path.

### DIFF
--- a/kubetest/e2e.go
+++ b/kubetest/e2e.go
@@ -246,7 +246,7 @@ func run(deploy deployer, o options) error {
 	var kubemarkDownErr error
 	if o.down && o.kubemark {
 		kubemarkWg.Add(1)
-		go kubemarkDown(&kubemarkDownErr, &kubemarkWg, o.provider, dump)
+		go kubemarkDown(&kubemarkDownErr, &kubemarkWg, o.provider, dump, o.logexporterGCSPath)
 	}
 
 	if o.charts {
@@ -762,18 +762,24 @@ func kubemarkGinkgoTest(testArgs []string, dump string) error {
 }
 
 // Brings down the kubemark cluster.
-func kubemarkDown(err *error, wg *sync.WaitGroup, provider, dump string) {
+func kubemarkDown(err *error, wg *sync.WaitGroup, provider, dump, logexporterGCSPath string) {
 	defer wg.Done()
 	control.XMLWrap(&suite, "Kubemark MasterLogDump", func() error {
 		logDumpPath := logDumpPath(provider)
-		cmd := exec.Command(logDumpPath, dump)
 		masterName := os.Getenv("MASTER_NAME")
+		var cmd *exec.Cmd
 		cmd.Env = append(
 			os.Environ(),
 			"KUBEMARK_MASTER_NAME="+masterName,
 			"DUMP_ONLY_MASTER_LOGS=true",
 		)
-		log.Printf("Dumping logs for kubemark master: %s", masterName)
+		if logexporterGCSPath != "" {
+			log.Printf("Dumping logs for kubemark master to GCS directly at path: %v", logexporterGCSPath)
+			cmd = exec.Command(logDumpPath, dump, logexporterGCSPath)
+		} else {
+			log.Printf("Dumping logs for kubemark master locally to: %v", dump)
+			cmd = exec.Command(logDumpPath, dump)
+		}
 		return control.FinishRunning(cmd)
 	})
 	*err = control.XMLWrap(&suite, "Kubemark TearDown", func() error {


NikhilSharmaWe:
@wojtek-t I am not sure that a condition like method should be used as it was in [here](https://github.com/kubernetes/test-infra/blob/ee1e7c845bb25177944bb5de37545f007f529059/kubetest/e2e.go#L561). Please inform for corrections.

kubetest/e2e.go
```diff
@@ -762,18 +762,18 @@ func kubemarkGinkgoTest(testArgs []string, dump string) error {
 }
 
 // Brings down the kubemark cluster.
-func kubemarkDown(err *error, wg *sync.WaitGroup, provider, dump string) {
+func kubemarkDown(err *error, logexporterGCSPath, wg *sync.WaitGroup, provider, dump string) {
 	defer wg.Done()
 	control.XMLWrap(&suite, "Kubemark MasterLogDump", func() error {
 		logDumpPath := logDumpPath(provider)
-		cmd := exec.Command(logDumpPath, dump)
+		cmd := exec.Command(logDumpPath, dump, logexporterGCSPath)
```
mborsz, line 769: logexporterGCSPath can be empty, in that case it shouldn't be added to exec.Command, like in: https://github.com/kubernetes/test-infra/blob/ee1e7c845bb25177944bb5de37545f007f529059/kubetest/e2e.go#L551

kubetest/e2e.go
```diff
@@ -762,18 +762,18 @@ func kubemarkGinkgoTest(testArgs []string, dump string) error {
 }
 
 // Brings down the kubemark cluster.
-func kubemarkDown(err *error, wg *sync.WaitGroup, provider, dump string) {
+func kubemarkDown(err *error, logexporterGCSPath, wg *sync.WaitGroup, provider, dump string) {
```
mborsz, line 765: You also need to change the code that calls this codepath, in like ~250. Please use o.logexporterGCSPath for as a value for the new parameter.

kubetest/e2e.go
NikhilSharmaWe: Just confirming. There should be an if else condition, one when logexporterGCSPath is present and second when it does not.

mborsz: Yes, this is correct. Thanks for help! :)

NikhilSharmaWe:
@mborsz could you please confirm the changes made.

kubetest/e2e.go
```diff
@@ -762,18 +762,24 @@ func kubemarkGinkgoTest(testArgs []string, dump string) error {
...
 			os.Environ(),
 			"KUBEMARK_MASTER_NAME="+masterName,
 			"DUMP_ONLY_MASTER_LOGS=true",
 		)
-		log.Printf("Dumping logs for kubemark master: %s", masterName)
+		if logexporterGCSPath != ""{
+			log.Printf("Dumping logs from kubemark master to GCS directly at path: %v", logexporterGCSPath)
+			cmd = exec.Command(logDumpPath, dump, logexporterGCSPath)
+		}else{
+			log.Printf("Dumping logs locally to: %v", dump)
```
mborsz, line 780: Can we include "kubemark master" in the log?

Like 
log.Printf("Dumping logs for kubemark master locally: %s", masterName)

kubetest/e2e.go
```diff
@@ -762,18 +762,24 @@ func kubemarkGinkgoTest(testArgs []string, dump string) error {
 }
 
 // Brings down the kubemark cluster.
-func kubemarkDown(err *error, wg *sync.WaitGroup, provider, dump string) {
+func kubemarkDown(err *error, o.logexporterGCSPath, wg *sync.WaitGroup, provider, dump string) {
```
mborsz, line 765: It should be like:

```
func kubemarkDown(err *error, wg *sync.WaitGroup, provider, dump, logexporterGCSPath string) {
```

So fix the syntax and move logexporterGCSPath to the last position.
And then change kubemarkDown call (around ~249 line in this file) to something like:

```
		go kubemarkDown(&kubemarkDownErr,  &kubemarkWg, o.provider, dump, o.logexporterGCSPath)

```

mborsz:
> Maciej Borsz could you please confirm the changes made.

Added two comments :)

kubetest/e2e.go
NikhilSharmaWe: I we will use log.Printf("Dumping logs for kubemark master locally: %s", masterName), the path to which the logs are dumped will not be printed. Can you please confirm me on that we need info of master name instead of the path in both the cases. By the way, mistakes you pointed out in the latter review are fixed.

NikhilSharmaWe: Do you want log.Printf("Dumping logs for kubemark master locally to: %v", dump) instead.

NikhilSharmaWe: Do you want log.Printf("Dumping logs for kubemark master locally to: %v", dump) instead.

NikhilSharmaWe:
@mborsz could you please confirm the changes made.

NikhilSharmaWe:
@mborsz @wojtek-t  could you please review the changes made.

kubetest/e2e.go
```diff
@@ -762,18 +762,24 @@ func kubemarkGinkgoTest(testArgs []string, dump string) error {
...
 		cmd.Env = append(
 			os.Environ(),
 			"KUBEMARK_MASTER_NAME="+masterName,
 			"DUMP_ONLY_MASTER_LOGS=true",
 		)
-		log.Printf("Dumping logs for kubemark master: %s", masterName)
+		if logexporterGCSPath != ""{
+			log.Printf("Dumping logs for kubemark master to GCS directly at path: %v", logexporterGCSPath)
+			cmd = exec.Command(logDumpPath, dump, logexporterGCSPath)
+		}else{
```
mborsz, line 779: Please add a spaces between "else" and "{"/"}" to make it "} else {" to follow style used in this file.

kubetest/e2e.go
```diff
@@ -762,18 +762,24 @@ func kubemarkGinkgoTest(testArgs []string, dump string) error {
 }
 
 // Brings down the kubemark cluster.
-func kubemarkDown(err *error, wg *sync.WaitGroup, provider, dump string) {
+func kubemarkDown(err *error, wg *sync.WaitGroup, provider, dump string, logexporterGCSPath string) {
```
mborsz, line 765: Nit: You can remove "string" in "dump string" to make it: "provider, dump, logexporterGCSPath string"

In golang, if there is no type next to the parameter name, it will use the next type (so "string" in logexporterGCSPath). This pattern is already used for "provider" here which takes "string" type.

kubetest/e2e.go
```diff
@@ -762,18 +762,24 @@ func kubemarkGinkgoTest(testArgs []string, dump string) error {
...
-		cmd := exec.Command(logDumpPath, dump)
 		masterName := os.Getenv("MASTER_NAME")
+		var cmd *exec.Cmd
 		cmd.Env = append(
 			os.Environ(),
 			"KUBEMARK_MASTER_NAME="+masterName,
 			"DUMP_ONLY_MASTER_LOGS=true",
 		)
-		log.Printf("Dumping logs for kubemark master: %s", masterName)
+		if logexporterGCSPath != ""{
```
mborsz, line 776: Please add a space before "{"

mborsz:
Looks good. Only styling comments.

NikhilSharmaWe:
@mborsz made changes according to the review, thanks for the help.

NikhilSharmaWe:
@mborsz made styling changes as you pointed out in your review, thanks for the help.

NikhilSharmaWe:
@mborsz Is this PR ready or is there something still to be worked on.

NikhilSharmaWe:
@mborsz Is this PR ready or is there something still to be worked on.

NikhilSharmaWe:
@mborsz Is this PR ready or is there something still to be worked on.

NikhilSharmaWe:
@mborsz Is this PR ready or is there something still to be worked on.

NikhilSharmaWe:
@mborsz Is this PR ready or is there something still to be worked on. If it is ready could you merge it.

mborsz:
It looks good!

/lgtm
/approve

PR #24204, opened 2021-11-02
- Revert "Revert "Gerrit client ensure authenticated at startup""

chaodaiG:
- Ensures that Authenticate is always initialized

This reverts commit d39781f626a1bde3f8ed4115b4c6808638a2ee14.

### DIFF
--- a/prow/gerrit/adapter/adapter.go
+++ b/prow/gerrit/adapter/adapter.go
@@ -91,8 +91,6 @@ func NewController(ctx context.Context, prowJobClient prowv1.ProwJobInterface, o
 	if err != nil {
 		logrus.WithError(err).Fatal("Error creating gerrit client.")
 	}
-	gerritClient.Authenticate(cookiefilePath, tokenPathOverride)
-
 	c := &Controller{
 		prowJobClient:      prowJobClient,
 		config:             cfg,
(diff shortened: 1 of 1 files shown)


prow/gerrit/adapter/adapter.go
```diff
@@ -91,38 +91,58 @@ func NewController(ctx context.Context, prowJobClient prowv1.ProwJobInterface, o
...
+	if orgReposConfig == nil {
+		return
+	}
+	// Use globally defined gerrit repos if present
+	if err := gerritClient.UpdateClients(orgReposConfig.AllRepos()); err != nil {
+		logrus.WithError(err).Error("Updating clients.")
+	}
+	if err := lastSyncTracker.update(orgReposConfig.AllRepos()); err != nil {
+		logrus.WithError(err).Error("Syncing states.")
+	}
```
cjwagner, line 140: I think we still need to include an `Authenticate` call some time after `UpdateClients` otherwise we only ever call `Authenticate` once.

cjwagner (changes requested):

prow/gerrit/adapter/adapter.go
chaodaiG: Authenticate spins off a goroutine updating when token refreshes, which is the only case where authenticate needs to be rerun when global config is not present

cjwagner: When global config is present and we add a new client to the `handler` map we need to call this line before using the handler, right? https://github.com/kubernetes/test-infra/blob/fcb86429c4b1f332290b2e000e58f4c6e7108b6c/prow/gerrit/client/client.go#L202

chaodaiG: ah, couldn't believe that I had deleted the `Authenticate` call below from original change! Awesome catch!

cjwagner (approved):
LGTM :crossed_fingers:

PR #24206, opened 2021-11-02
Run verify spelling with make rules

chaodaiG:


spiffxp (approved):
/approve
/lgtm

chaodaiG:
Merge conflict resolved, @spiffxp another lgtm please

Makefile
```diff
@@ -70,5 +70,9 @@ update-file-perms:
 	hack/make-rules/update/file-perms.sh
 verify-file-perms:
 	hack/make-rules/verify/file-perms.sh
+update-spelling:
+	hack/make-rules/update/misspell.sh
+verify-spelling:
+	hack/make-rules/verify/misspell.sh
```
BenTheElder, line 76: these should be under `.PHONY` (same comment as the other PR https://github.com/kubernetes/test-infra/pull/24213#discussion_r741517101)

chaodaiG: done

hack/make-rules/verify/misspell.sh
```diff
@@ -0,0 +1,84 @@
...
+echo "Ensuring go version."
+source ./hack/build/setup-go.sh
+
+# build misspell
+echo "Install misspell."
+cd "hack/tools"
+go build -o "$MISSPELL" github.com/client9/misspell/cmd/misspell
+cd "${REPO_ROOT}"
+
+trap 'echo ERROR: found unexpected instance of "Git"hub, use github or GitHub' ERR
```
BenTheElder, line 36: nit: why the quotes around Git here? we've actually found an instance of `Github` (no quotes) right?

hack/make-rules/verify/misspell.sh
```diff
@@ -0,0 +1,84 @@
...
+    -o -path '*/localdata/*' \
+    -o -path '*/gubernator/*' \
+    -o -path '*/prow/bugzilla/client_test.go' \
+    -o -path './.git/*' \
+    -o -path './_bin/*' \
+    -o -path './_output/*' \
+    -o -path './_artifacts/*' \
+    -o -path './bazel-*/*' \
+    \) -prune \
+    \) -exec grep -Hn 'Git'hub '{}' '+' ; then
```
BenTheElder, line 57: nit: why isn't this just
```suggestion
    \) -exec grep -Hn Github '{}' '+' ; then
```

hack/make-rules/verify/misspell.sh
chaodaiG: yeah the test itself would fail without quotes

hack/make-rules/verify/misspell.sh
chaodaiG: Same reason as above, this fail the test itself

hack/make-rules/verify/misspell.sh
BenTheElder: lol! in kubernetes we use `tac` to solve this 🙃

PR #24213, opened 2021-11-02
Run verify labels with make rules

chaodaiG:


Makefile
```diff
@@ -66,6 +66,10 @@ update-gofmt:
 	hack/make-rules/update/gofmt.sh
 verify-gofmt:
 	hack/make-rules/verify/gofmt.sh
+update-labels:
+	hack/make-rules/update/labels.sh
+verify-labels:
+	hack/make-rules/verify/labels.sh
```
BenTheElder, line 72: These should also be `.PHONY`.
https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html

hack/make-rules/verify/all.sh
```diff
@@ -35,6 +35,9 @@ if [[ "${VERIFY_GOFMT:-true}" == "true" ]]; then
   hack/make-rules/verify/gofmt.sh || res=1
   cd "${REPO_ROOT}"
 fi
+if [[ "${VERIFY_LABELS:-true}" == "true" ]]; then
+  echo "verifying labels ..."
+  hack/make-rules/verify/labels.sh || res=1
```
BenTheElder, line 40: missing a `fi` here.

Makefile
chaodaiG: added

hack/make-rules/verify/all.sh
chaodaiG: added

fejta (approved):

PR #24234, opened 2021-11-04
sig-node: skip DynamicKubeletConfig test under serial lanes

cynepco3hahue:
The DynamicKubeletConfig already deprecated and should be removed under
release 1.23, so it no taste to run it under serial lanes anymore.

Signed-off-by: Artyom Lukianov <alukiano@redhat.com>


SergeyKanzhelev (approved):
this was discussed on SIG Node CI meeting. We are decreasing code coverage, but the feature is deprecated and causing a lot of troubles for tests stability. So it is ok to skip.

PR #24236, opened 2021-11-04
Crier gerrit reporter authenticate right after orgs/repos update

chaodaiG:
Newly added orgs/repos are only authenticated by the goroutine when token secret rotated, which is up to 1 hour after config change. Explicitly call here to get them authenticated immediately.

### DIFF
--- a/prow/crier/reporters/gerrit/reporter.go
+++ b/prow/crier/reporters/gerrit/reporter.go
@@ -98,22 +98,49 @@ func NewReporter(cfg config.Getter, cookiefilePath string, projects map[string][
 	if err != nil {
 		return nil, err
 	}
+	// applyGlobalConfig reads gerrit configurations from global gerrit config,
+	// it will completely override previously configured gerrit hosts and projects.
+	// it will also by the way authenticate gerrit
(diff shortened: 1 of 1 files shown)


prow/crier/reporters/gerrit/reporter.go
```diff
@@ -106,7 +106,14 @@ func NewReporter(cfg config.Getter, cookiefilePath string, projects map[string][
 				time.Sleep(time.Second)
 				continue
 			}
+			// Updates clients based on global gerrit config.
 			gc.UpdateClients(orgReposConfig.AllRepos())
```
cjwagner, line 110: It doesn't look like we respect the global org repo config immediately when `NewReporter` is called, that happens asynchronously.  If we follow the same pattern used by the gerrit adapter (maybe even use the same code) we won't risk transient errors on start up.

chaodaiG: good point, updated the code.

> use the same code

Technically this could be done by adding this logic in gerrit Client, however it might be an overkill for gerrit Client to understand prow config, so duplicated code instead, which is not too much code anyways

cjwagner (approved):

PR #24238, opened 2021-11-04
config/jobs: run no-stage on k8s-infra, drop extract

spiffxp:
Related:
- Part of: https://github.com/kubernetes/test-infra/issues/18789

It's unclear to me why this job isn't working on the default build cluster due to a lack of docker.  That's the only difference I can see vs. the normal job, so let's try reconciling that.

Also drop the --extract=local flag, because my read of the bash deployer is it should work with what's been built locally

### DIFF
--- a/config/jobs/kubernetes/sig-cloud-provider/gcp/gcp-gce.yaml
+++ b/config/jobs/kubernetes/sig-cloud-provider/gcp/gcp-gce.yaml
@@ -62,6 +62,7 @@ presubmits:
           privileged: true
   - name: pull-kubernetes-e2e-gce-no-stage
     always_run: false
+    cluster: k8s-infra-prow-build
     annotations:
       testgrid-dashboards: presubmits-kubernetes-nonblocking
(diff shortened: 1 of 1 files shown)


amwat:
Docker errors during build mean mostly likely related to DIND. 

I think it needs 
```
securityContext:
  privileged: true
```
for DIND 

also yeah good point on removing extract altogether if it's not needed. 
That should fix extract relying on `--stage`

BenTheElder:
I have a hazy memory of extract being necessary but .. so long ago ...

let's try it
/lgtm
/approve
/hold

BenTheElder (approved):
/lgtm
/approve
let's try it

spiffxp:
@amwat 
> Docker errors during build mean mostly likely related to DIND.
> 
> I think it needs
> 
> ```
> securityContext:
>   privileged: true
> ```

Good catch, that absolutely sounds like the issue.  Updated the PR.  Will run it on k8s-infra anyway just for extra parity

spiffxp:
rebase fail, removed the dupe no-stage job definition

PR #24241, opened 2021-11-04
Prow config to run Local Volume Provisioner in Windows 2019

mauriciopoppe:
/sig-storage
/cc @jingxu97

### DIFF
--- a/config/jobs/kubernetes-sigs/sig-storage-local-static-provisioner/sig-storage-local-static-provisioner.yaml
+++ b/config/jobs/kubernetes-sigs/sig-storage-local-static-provisioner/sig-storage-local-static-provisioner.yaml
@@ -64,6 +64,44 @@ presubmits:
         # docker-in-docker needs privileged mode
         securityContext:
           privileged: true
+  - name: pull-sig-storage-local-static-provisioner-e2e-windows-2019
+    always_run: false
+    decorate: true
+    path_alias: sigs.k8s.io/sig-storage-local-static-provisioner
+    labels:
+      preset-k8s-ssh: "true"
+      preset-service-account: "true"
+      preset-common-gce-windows: "true"
+      preset-e2e-gce-windows: "true"
+      preset-bazel-scratch-dir: "true"
+      preset-bazel-remote-cache-enabled: "true"
+      preset-dind-enabled: "true"
+    spec:
+      containers:
+      - image: gcr.io/k8s-testimages/kubekins-e2e:v20210902-e4567b8-master
+        command:
+        - runner.sh
+        args:
+        - hack/e2e.sh
+        - --
+        - --test-cmd-args=--allowed-not-ready-nodes=10
+        - --test-cmd-args=--node-os-distro=windows
+        env:
+        # minimum size that supports Hyper-V
+        - name: NODE_SIZE
+          value: "e2-standard-2"
+        - name: NODE_LOCAL_SSDS
+          value: "2"
+        # added in https://github.com/kubernetes/kubernetes/pull/105999
+        - name: WINDOWS_ENABLE_HYPERV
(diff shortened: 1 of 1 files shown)


mauriciopoppe:
/assign @msau42 
This is my first iteration, I couldn't test it with pj-on-kind.sh because of the service account key limitation but hopefully I can get enough logs from the job to debug it later

config/jobs/kubernetes-sigs/sig-storage-local-static-provisioner/sig-storage-local-static-provisioner.yaml
```diff
@@ -64,6 +64,42 @@ presubmits:
...
+        # added in https://github.com/kubernetes/kubernetes/pull/105999
+        - name: WINDOWS_ENABLE_HYPERV
+          value: "true"
+        # if the value is changed, also change the hack/e2e.sh Windows image to build!
+        # this is done so that the e2e test runs faster
+        - name: WINDOWS_NODE_OS_DISTRIBUTION
+          value: "win2019"
+        # docker-in-docker needs privileged mode
+        securityContext:
+          privileged: true
```
jingxu97, line 102: is it supported already?

jingxu97: i see other windows test set it, so should be ok

config/jobs/kubernetes-sigs/sig-storage-local-static-provisioner/sig-storage-local-static-provisioner.yaml
```diff
@@ -64,6 +64,42 @@ presubmits:
...
+        command:
+        - runner.sh
+        args:
+        - hack/e2e.sh
+        - --
+        - --test-cmd-args=--allowed-not-ready-nodes=10
+        - --test-cmd-args=--node-os-distro=windows
+        env:
+        # minimum size that supports Hyper-V
+        - name: NODE_SIZE
```
msau42, line 91: Do you need to add local SSDs to the VM?

config/jobs/kubernetes-sigs/sig-storage-local-static-provisioner/sig-storage-local-static-provisioner.yaml
mauriciopoppe: I think this refers to the pod where the test runs

config/jobs/kubernetes-sigs/sig-storage-local-static-provisioner/sig-storage-local-static-provisioner.yaml
mauriciopoppe: The e2e tests create virtual hard disks (VHDs) so a physical disk is not needed, what I saw from this guide https://cloud.google.com/kubernetes-engine/docs/how-to/persistent-volumes/local-ssd#formatting_local_ssds_for_windows_server_clusters is that even if the VM has local SSDs attached they need to be formatted and mounted to a location using powershell commands (which we can do through SSH).

I could add this as another e2e test where instead of using VHDs the initialization scripts can format and mount the local SSDs instead.

I've modified the env var to create local SSDs (from https://github.com/kubernetes/kubernetes/blob/8ce440c45ce4b70a2d770b990cad2862dd4d9288/cluster/gce/config-default.sh?#L42) for the next iteration of the e2e test.

msau42: Thanks, I'm mainly wondering if there would be some difference between VHD and a physical disk, so it would be good to test both. For now, this is fine.

PR #24249, opened 2021-11-04
Remove graceful node shutdown e2e job

bobbypage:
The graceful shutdown feature is now beta and the test is running under
existing node-kubelet-serial job. As a result, the dedicated graceful
node shutdown e2e job is no longer necessary. Followup to
https://github.com/kubernetes/test-infra/pull/24154

### DIFF
--- a/config/jobs/kubernetes/sig-node/node-kubelet.yaml
+++ b/config/jobs/kubernetes/sig-node/node-kubelet.yaml
@@ -402,40 +402,6 @@ periodics:
     testgrid-alert-email: alukiano@redhat.com
     description: Executes memory manager e2e node tests
 
-- name: ci-kubernetes-node-kubelet-serial-graceful-node-shutdown
-  interval: 4h
-  labels:
-    preset-service-account: "true"
-    preset-k8s-ssh: "true"
-  spec:
-    containers:
-      - image: gcr.io/k8s-staging-test-infra/kubekins-e2e:v20211104-c3a88263fc-master
-        args:
-          - --repo=k8s.io/kubernetes=master
-          - --timeout=240
-          - --root=/go/src
-          - --scenario=kubernetes_e2e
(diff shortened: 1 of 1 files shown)


SergeyKanzhelev:
Do we know how disruptive for the kubelet the dbus notification? Is the expectation that the kubelet must keep working after cancelled shutdown? May it affect tests that run after it? Should we somehow cleanup after the test passed, like maybe restart the kubelet?

bobbypage:
> Do we know how disruptive for the kubelet the dbus notification?

The dbus notification will trigger to the kubelet to shutdown all of the pods. However since the the test is running in serial, it should be graceful shutdown test which receives the signal (and only affects that test).

> Is the expectation that the kubelet must keep working after cancelled shutdown? 

Yes, it is the expectation. After cancelled shutdown we expect kubelet to back to `NodeReady` and test waits for that condition to be true. 

https://github.com/kubernetes/kubernetes/blob/508e67937e3644ae7a844bf572cde5ef29543690/test/e2e_node/node_shutdown_linux_test.go#L173-L176 

> May it affect tests that run after it? Should we somehow cleanup after the test passed, like maybe restart the kubelet?

I don't expect that it should affect tests running after it assuming that tests passes and the node gets back into `ReadyCondition`. In fact it's running in serial right now and everything seems to be fine. https://testgrid.k8s.io/sig-node-kubelet#node-kubelet-serial

The one consideration is that if for some reason we sent dbus signal and the node went to not ready and did not recover as expected the kubelet would be in a strange state. So maybe to be safe, it may make sense to always restart kubelet at the end of the test. Perhaps it makes sense for every serial test to do that so it can start with "fresh" kubelet :)

SergeyKanzhelev (approved):
we may want to universally restart kubelet after serial test. Let's not block this PR on it.

PR #24252, opened 2021-11-04
Update/verify codegen by make rulew

chaodaiG:
To make code reviewers' lives easier, this PR contains two commits:
- Copy/paste blindly update/verify-codegen scripts using bazel
- Change to not use bazel


hack/make-rules/update/codegen.sh
```diff
@@ -17,50 +17,66 @@ set -o errexit
...
+cd $REPO_ROOT
+
+echo "Ensuring go version."
+source ./hack/build/setup-go.sh
+
+# build codegen tools
+echo "Install codegen tools."
+cd "hack/tools"
+clientgen=${REPO_ROOT}/_bin/client-gen
+go build -o "${REPO_ROOT}/_bin/client-gen" k8s.io/code-generator/cmd/client-gen
```
cjwagner, line 40: I think it would make more sense to define make targets for these and make them dependencies of the `update-codegen` target. That would also let us avoid rebuilding them when not needed.

hack/make-rules/verify/codegen.sh
```diff
@@ -17,39 +17,31 @@ set -o errexit
...
+cd $REPO_ROOT
 
-DIFFROOT="${SCRIPT_ROOT}/prow"
-TMP_DIFFROOT="${TEST_TMPDIR}/prow"
+# place to stick temp binaries
+BINDIR="${REPO_ROOT}/_bin"
 
+DIFFROOT="${REPO_ROOT}/prow"
+TMP_DIFFROOT="$(TMPDIR="${BINDIR}" mktemp -d "${BINDIR}/verify-codegen.XXXXX")"
 mkdir -p "${TMP_DIFFROOT}"
```
cjwagner, line 28: nit: I think this is handled my `mktemp -d` now.

We should also clean up `TMP_DIFFROOT` when we're done with it. As is, running this locally will continually grow the size of the BINDIR.

hack/make-rules/verify/codegen.sh
```diff
@@ -17,39 +17,31 @@ set -o errexit
...
-fi
-
-SCRIPT_ROOT=$PWD
+REPO_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/../../.." && pwd -P)"
+cd $REPO_ROOT
 
-DIFFROOT="${SCRIPT_ROOT}/prow"
-TMP_DIFFROOT="${TEST_TMPDIR}/prow"
+# place to stick temp binaries
+BINDIR="${REPO_ROOT}/_bin"
```
cjwagner, line 24: We should probably add this to the git ignore list.

hack/tools/tools.go
```diff
@@ -1,4 +1,3 @@
-//go:build tools
 // +build tools
```
cjwagner: I'm not sure what these directives do. Are they for bazel? If so don't we still need them?

hack/make-rules/update/codegen.sh
chaodaiG: Not quite sure that I'm getting the idea. If you mean create a new make target just for installing tools used by update-codegen, it would work but I'm not sure it's the best idea since this new target is used only by update-codegen, maybe not worth the effort of inflating root level Makefile?

For local development, this is pretty fast after the first time, mostly due to go cache I think

hack/make-rules/verify/codegen.sh
chaodaiG: good point, done

hack/make-rules/verify/codegen.sh
chaodaiG: it's already in .gitignore

hack/tools/tools.go
chaodaiG: These are not for bazel. gofmt enforces `//go:build tools` over `// +build tools` starting 1.17, and this change was an accident from the first commit, in current master there is only `// +build tools`

hack/make-rules/update/codegen.sh
cjwagner: Ok, sounds good.

cjwagner (approved):

PR #24254, opened 2021-11-04
run kind multizone tests in parallel

aojea:
it was running serial hence taking more than 1 hour.
Parallel runs in less than 30 mins and also checks that tests doesn't conflict between them


robscott:
Thanks!

/lgtm

PR #24267, opened 2021-11-05
Report problem when label added as both restricted and additional

jmguzik:
Validation to ensure that a restricted label is also configured as additional does not exist.
This PR addresses this issue and resolves a problem.


prow/cmd/checkconfig/main.go
```diff
@@ -1052,23 +1052,45 @@ func verifyOwnersPlugin(cfg *plugins.Configuration) error {
...
+func checkRestrictedLabelNotInAdditionalLables(restricted string, label plugins.Label) bool {
+	for _, additional := range label.AdditionalLabels {
+		if restricted == additional {
+			return true
+		}
+	}
+	return false
+}
+
+func verifyLabelPlugin(label plugins.Label) []error {
```
droslean, line 1064: Let's make this function to return an actual error from the error list, using `https://pkg.go.dev/k8s.io/apimachinery/pkg/util/errors#NewAggregate`

jmguzik: Done

prow/cmd/checkconfig/main.go
```diff
@@ -1052,23 +1052,45 @@ func verifyOwnersPlugin(cfg *plugins.Configuration) error {
 	return nil
 }
 
+func checkRestrictedLabelNotInAdditionalLables(restricted string, label plugins.Label) bool {
```
droslean, line 1055: Let's have this as a `Label{}` method in https://github.com/kubernetes/test-infra/blob/d0d3c6d7a8d0/prow/plugins/config.go#L381

Also the name can be something like `isRestrictedLabelInAdditionalLables()` ?

prow/cmd/checkconfig/main.go
```diff
@@ -1052,23 +1052,45 @@ func verifyOwnersPlugin(cfg *plugins.Configuration) error {
...
+	return false
+}
+
 func verifyLabelPlugin(label plugins.Label) error {
 	var orgRepos []string
+	var errs []error
+	misconfiguredLabels := make(map[string][]string)
 	for orgRepo, restrictedLabels := range label.RestrictedLabels {
 		for _, restrictedLabel := range restrictedLabels {
+			if found := checkRestrictedLabelNotInAdditionalLables(restrictedLabel.Label, label); found {
```
droslean, line 1070: assigning it to `found` var is not necessary.

prow/cmd/checkconfig/main.go
```diff
@@ -1052,23 +1052,45 @@ func verifyOwnersPlugin(cfg *plugins.Configuration) error {
...
+				misconfiguredLabels[restrictedLabel.Label] = append(misconfiguredLabels[restrictedLabel.Label], orgRepo)
+			}
 			if restrictedLabel.Label == "" {
 				orgRepos = append(orgRepos, orgRepo)
 			}
 		}
 	}
 
+	if len(misconfiguredLabels) > 0 {
```
droslean, line 1079: no need for this check here, since you `make` the map above, the map won't be nil, and if it's empty the interaction won't happen.

prow/cmd/checkconfig/main_test.go
```diff
@@ -2281,6 +2281,49 @@ func TestVerifyLabelPlugin(t *testing.T) {
...
+		},
+		{
+			name: "invalid when additional and restricted labels are the same in multiple orgRepos and empty string",
+			label: plugins.Label{
+				AdditionalLabels: []string{"cherry-pick-approved"},
+				RestrictedLabels: map[string][]plugins.RestrictedLabel{
+					"orgRepo1": {
+						{
+							Label:        "cherry-pick-approved",
+							AllowedTeams: []string{"some-other-team"},
```
droslean, line 2307: Is this tested for this case?

prow/cmd/checkconfig/main_test.go
```diff
@@ -2281,6 +2281,49 @@ func TestVerifyLabelPlugin(t *testing.T) {
...
+						},
+					},
+					"orgRepo2": {
+						{
+							Label:        "",
```
droslean, line 2312: No need for the empty field here.

prow/cmd/checkconfig/main_test.go
jmguzik: teams could be deleted

prow/cmd/checkconfig/main_test.go
jmguzik: empty field left here explicitly and mentioned in the test description to show the interaction between two types of errors. Do you want to delete it?

jmguzik: exactly the same is marked at line 2270 and 2277

droslean: It's fine, let's leave it to make the test more readable.

prow/plugins/config.go
```diff
@@ -401,6 +401,15 @@ func (l Label) RestrictedLabelsFor(org, repo string) map[string]RestrictedLabel
 	return result
 }
 
+func (l Label) IsRestrictedLabelInAdditionalLables(restricted string) bool {
```
droslean, line 404: Can you add godoc here?

droslean: Up to you. I see that the other method doesn't have either.

prow/cmd/checkconfig/main.go
```diff
@@ -1054,21 +1054,32 @@ func verifyOwnersPlugin(cfg *plugins.Configuration) error {
 
 func verifyLabelPlugin(label plugins.Label) error {
 	var orgRepos []string
+	var errs []error
+	misconfiguredLabels := make(map[string][]string)
```
alvaroaleman, line 1058: please use easier to understand variable names, e.G. s/misconfiguredLabels/restrictedAndAdditionalLabels and s/orgRepos/orgReposWithEmptyLabelConfig

jmguzik: Done

alvaroaleman (approved):

PR #24271, opened 2021-11-07
config: export `FinalizeJobConfig` method

petr-muller:
Enable experiments with dynamically generating Prowjobs and submitting them to the cluster. It is not hard to generate Prowjob configuration on the fly, but without exporting the finalization method (which is only called on `Load()` otherwise), they do not contain the defaults necessary to be successfuly submitted, such as default decoration configs.


petr-muller:
/hold

Let me try a little approach here, maybe export a new method that will explicitly just finalize and possibly validate an input job

PR #24274, opened 2021-11-08
Adjust CAPI job frequency on old branches

fabriziopandini:
Adjusting  job frequency on old branches
- periodic jobs from 1h to 4h (link check from 6h to 24h)
- upgrade jobs from 48h to 24h


fabriziopandini:
/cc @sbueringer @enxebre 
given they expressed opinions about this in slack

PR #24275, opened 2021-11-08
pull-containerd-node-e2e: skip dynamickubeletconfig tests

endocrimes:
DynamicKubeletConfig is not enabled by default (deprecated) and soon to be removed. Its tests are also incredibly flaky. 

As the various changes to the test suite and runner happened, it seems that containerd's presubmits were missed, Rather than enabling the feature gate here, lets skip the tests as they're about to be removed and are low signal/high noise in presubmits.

xref: https://github.com/containerd/containerd/issues/6215


adisky (commented):
/lgtm

adisky:
@endocrimes the periodic job `containerd-node-features` is also failing with the same reason https://testgrid.k8s.io/sig-node-containerd#containerd-node-features

PR #24222, opened 2021-11-03
Remove branch protection on `gh-pages` for external-dns & metrics-server Helm chart publish

stevehipwell:
The Helm chart releases for external-dns & metrics-server have been failing due to the `gh-pages` branch being protected, this PR aims to follow [cluster-autoscaler](https://github.com/kubernetes/test-infra/pull/23024) (amongst others) in fixing this.

PR #24278, opened 2021-11-08
Remove branch protection from gh-pages for Descheduler

damemi:
Similar to https://github.com/kubernetes/test-infra/pull/24222, this allows the Helm chart releaser action to run automatically (currently it gets blocked on the CLA check. Since we don't allow any PRs to this branch, and the action only runs after a successful merge to another branch that does have the CLA check, this shouldn't be necessary)

### DIFF
--- a/config/prow/config.yaml
+++ b/config/prow/config.yaml
@@ -467,6 +467,10 @@ branch-protection:
           branches:
             gh-pages:
               protect: false
+        descheduler:
+          branches:
+            gh-pages:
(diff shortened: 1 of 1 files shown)


damemi (commented):
ping @seanmalloy @ingvagabund FYI

damemi:
/cc @spiffxp 
could you please approve? Thanks!

spiffxp (approved):
/approve
/lgtm

PR #24280, opened 2021-11-08
Allow Slack states to report to be optional, while still respect empt…

chaodaiG:
…y slice

Resolving comment: https://github.com/kubernetes/test-infra/pull/24168#issuecomment-963347893.

### DIFF
--- a/prow/pjutil/pjutil.go
+++ b/prow/pjutil/pjutil.go
@@ -40,6 +40,8 @@ import (
 // NewProwJob initializes a ProwJob out of a ProwJobSpec.
 func NewProwJob(spec prowapi.ProwJobSpec, extraLabels, extraAnnotations map[string]string) prowapi.ProwJob {
 	labels, annotations := decorate.LabelsAndAnnotationsForSpec(spec, extraLabels, extraAnnotations)
+	specCopy := spec.DeepCopy()
+	setReportDefault(specCopy)
 
 	return prowapi.ProwJob{
 		TypeMeta: metav1.TypeMeta{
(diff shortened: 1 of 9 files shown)


chaodaiG:
Just an idea, this is probably the best place I could find to insert this logic. However, not feeling great injecting something almost "hacky" by working around the issue. Please let me know if there is a better solution to this

chaodaiG:
Follow up https://github.com/kubernetes/test-infra/pull/24168#issuecomment-963347893, @alvaroaleman , any thoughts?

prow/pjutil/pjutil.go
```diff
@@ -51,14 +53,33 @@ func NewProwJob(spec prowapi.ProwJobSpec, extraLabels, extraAnnotations map[stri
...
+//
+// `omitempty` is required for fields that are optional, otherwise strict prowjob CRD
+// validation will reject prowjob without the field.
+// For `ReporterConfig.Slack.JobStatesToReport`, it's certainly not a required field
+// for all prowjobs.
+// However, when omitempty is presented, `JobStatesToReport: []` roundtrip into
+// `JobStatesToReport: nil`, which prevents a prowjob from overriding global Slack
+// report configuration(ref: https://github.com/kubernetes/test-infra/issues/22888#issuecomment-881513368).
+// Use a boolean instead so that we can use `omitempty`.
+func setReportDefault(spec *prowapi.ProwJobSpec) {
```
alvaroaleman, line 74: This doesn't do anything, because a non-pointer bool field with `omitempty` will be omitted in the serializiation when the value is `false` as that is the default for an unititialized boolean.

I think we need to make it a pointer field to differentiate `uninitialized` from `false` and also have an `else` block in the second condition here that sets it to `true`?

chaodaiG: good point, updated the PR, as well as some unit tests

petr-muller (approved):
/hold

Holding for Alvaro/Cole

alvaroaleman (approved):

PR #24281, opened 2021-11-08
Bazel stops linting spelling and label by default

chaodaiG:
The have been migrated over to make targets


spiffxp (approved):
/approve
/lgtm

Issue #24190, opened 2021-11-01
k8s-testgrid is not displaying logs for ppc64le jobs


Rajalakshmi-Girish:
**What happened**:
https://k8s-testgrid.appspot.com/sig-node-ppc64le#conformance and
https://k8s-testgrid.appspot.com/sig-node-ppc64le#unit-tests are not displaying logs of prow job from gs bucket.
When tried to click on the `Job History` below error is seen:

```
failed to get job history: could not instantiate storage bucket: bucket not in allowed list; you may allow it by including it in `deck.additional_allowed_buckets`: bucket "ppc64le-kubernetes" not in allowed list ([containerd-integration k8s-ovn kubernetes-jenkins])
```

**What you expected to happen**:
The logs should get loaded from gs bucket and there should not be any error on viewing `Job History`.

**How to reproduce it (as minimally and precisely as possible)**:
Try viewing jobs under https://k8s-testgrid.appspot.com/sig-node-ppc64le#unit-tests and https://k8s-testgrid.appspot.com/sig-node-ppc64le#conformance

**Please provide links to example occurrences, if any**:
https://prow.k8s.io/view/gs/ppc64le-kubernetes/logs/periodic-kubernetes-conformance-test-ppc64le/1455178274941440000

**Anything else we need to know?**:
Should the Google Bucket `ppc64le-kubernetes` to be added to the list at https://github.com/kubernetes/test-infra/blob/master/config/prow/config.yaml#L97?
Was there any recent change that made this addition mandatory?
We were earlier able to view logs without adding the bucket name to that list.

PR #24282, opened 2021-11-08
Adds ppc64le-kubernetes bucket to the additional_allowed_buckets list

Rajalakshmi-Girish:
Fixes https://github.com/kubernetes/test-infra/issues/24190
Also there was an [Announcement](https://github.com/kubernetes/test-infra/blob/master/prow/ANNOUNCEMENTS.md) on `April 12th, 2021` that reads
```
End of grace period for storage bucket validation, additional buckets have to be allowed by adding them to the deck.additional_allowed_buckets list.
```

### DIFF
--- a/config/prow/config.yaml
+++ b/config/prow/config.yaml
@@ -97,6 +97,7 @@ deck:
   additional_allowed_buckets:
     - k8s-ovn
     - containerd-integration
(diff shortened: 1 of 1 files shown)


spiffxp (approved):
/approve
/lgtm

PR #24289, opened 2021-11-08
config: expose a new method to default a single input `Periodic`

petr-muller:
An alternative to https://github.com/kubernetes/test-infra/pull/24271

Enable experiments with dynamically generating Prowjobs and submitting them to the cluster. It is not hard to generate Prowjob configuration on the fly, but without exporting the finalization method (which is only called on Load() otherwise), they do not contain the defaults necessary to be successfuly submitted, such as default decoration configs.

### DIFF
--- a/prow/config/config.go
+++ b/prow/config/config.go
@@ -1766,16 +1766,19 @@ func defaultPostsubmits(postsubmits []Postsubmit, additionalPresets []Preset, c 
 	return utilerrors.NewAggregate(errs)
 }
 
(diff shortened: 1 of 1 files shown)


petr-muller:
/cc @alvaroaleman 

Do you have concerns about exposing this? Code change is not too invasive.
(pinging you because of your inrepoconfig history where you seemed to need something similar)

alvaroaleman (approved):

PR #24292, opened 2021-11-09
Use squash method for kubernetes-sigs/cluster-api

shamsher31:
While working on https://github.com/kubernetes-sigs/cluster-api/pull/5593, I realized that the `cluster-api` don't have the `squash` method enabled. After the PR was merged, it ended up with 3 commits instead of 1.

@fabriziopandini Is there any reason not to use `squash` for PR merge?

### DIFF
--- a/config/prow/config.yaml
+++ b/config/prow/config.yaml
@@ -628,6 +628,7 @@ tide:
   merge_method:
     kubernetes-client/csharp: squash
     kubernetes-client/gen: squash
+    kubernetes-sigs/cluster-api: squash
     kubernetes-sigs/cluster-api-provider-digitalocean: squash
     kubernetes-sigs/cluster-api-provider-ibmcloud: squash
     kubernetes-sigs/ibm-vpc-block-csi-driver: squash


fabriziopandini:
I remember discussing this option at some time and then discarding it given that we want to preserve it keep control over this, if I'm not wrong because in some cases having more commits makes sense and due to problems in preserving commits authors/co authors, but I might be wrong here

might be @vincepri or @CecileRobertMichon have some more context here

CecileRobertMichon:
wasn't there also a problem with the release notes generation when commits were merged using the tide squash?

vincepri:
Squash breaks our release note generation tooling which relies on merge commits from github that contain PR title information.

shamsher31:
> Squash breaks our release note generation tooling which relies on merge commits from github that contain PR title information.

Got it, but my only concern is if someone is working on a feature for a month, iterated with more than 10 commits, and PR got merged, There will be more than 10 commits for a single feature in the git history but ideally, it should be 1 commit. 

@vincepri Is there anything we can do to find parent commit for PR, use that in the release notes generation and enable Squash to maintain clean history?

vincepri:
@shamsher31 Usually approvers won't merge something if it's the PR's author hasn't squashed commits, we only ask to squash before merging, so the above hasn't really been an issue

PR #24295, opened 2021-11-09
Prow auto-deploy cron adjusts against daylight saving

chaodaiG:


config/jobs/kubernetes/test-infra/test-infra-trusted.yaml
```diff
@@ -600,7 +600,9 @@ periodics:
     testgrid-dashboards: sig-testing-maintenance
     testgrid-tab-name: autoupdate-patch
     description: Weekly module update to latest patched versions
-- cron: "30 17-22/5 * * 1-5"  # Bump with label `skip-review`. Run at 10:30 and 15:30 PST (17:05 UTC) Mon-Fri
+- cron: "30 18-23/5 * * 1-5"  # Bump with label `skip-review`. Run at 10:30 and 15:30 PST (18:05 UTC, fall) Mon-Fri
+  # Save for daylight saving:
+  # cron: "30 17-22/5 * * 1-5"  # Bump with label `skip-review`. Run at 10:30 and 15:30 PST (17:05 UTC, spring) Mon-Fri
```
BenTheElder, line 605: If we upgraded the cron library to v3 we could just set the desired timezone in the cron string 🙃 
https://pkg.go.dev/github.com/robfig/cron/v3 `CRON_TZ`

chaodaiG: good point! It's certainly a good idea. Appears that v3 is compatible with v1 and v2, so it's not a trivial task, I'll create an issue for now. Considering this is a once per 6 months trivial change, there might not be a huge rush to upgrade to v3.

PR #24297, opened 2021-11-09
Adding sig-net multizone periodic tests

robscott:
This is a follow up on #24242 that adds a periodic that is nearly identical to the optional presubmit that PR added. I'm hoping that a periodic task like this will help us detect failures and determine how reliable this config is.

/sig network
/priority important-soon


aojea:
/lgtm
/approve
:100:

PR #24300, opened 2021-11-10
There are no Feature:Performance tests any longer

SergeyKanzhelev:
/sig scalability
/cc @wojtek-t

Tests were removed: https://github.com/kubernetes/kubernetes/pull/83322/

Also see: https://github.com/kubernetes/kubernetes/pull/106283

Test grid links:
- https://testgrid.k8s.io/sig-testing-canaries#kubemark
- https://testgrid.k8s.io/sig-testing-canaries#scalability

both are empty

/kind cleanup
/priority backlog

### DIFF
--- a/config/jobs/kubernetes/sig-testing/kubetest-canaries.yaml
+++ b/config/jobs/kubernetes/sig-testing/kubetest-canaries.yaml
@@ -54,37 +54,6 @@ periodics:
     testgrid-dashboards: sig-testing-canaries
     testgrid-tab-name: gce
 
-- interval: 2h
-  name: ci-kubernetes-e2e-gce-scalability-canary
-  labels:
-    preset-service-account: "true"
-    preset-k8s-ssh: "true"
-    preset-e2e-scalability-common: "true"
-  decorate: true
-  decoration_config:
-    timeout: 140m
-  spec:
-    containers:
-    - command:
-      - runner.sh
-      - /workspace/scenarios/kubernetes_e2e.py
-      args:
-      - --cluster=e2e-big-canary
-      - --extract=ci/latest
-      - --gcp-nodes=100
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes/sig-testing/kubetest-canaries.yaml
```diff
@@ -54,37 +54,6 @@ periodics:
...
-      - runner.sh
-      - /workspace/scenarios/kubernetes_e2e.py
-      args:
-      - --cluster=e2e-big-canary
-      - --extract=ci/latest
-      - --gcp-nodes=100
-      - --gcp-project-type=scalability-project
-      - --gcp-zone=us-east1-b
-      - --provider=gce
-      - --test_args=--ginkgo.focus=\[Feature:Performance\] --gather-resource-usage=true --gather-metrics-at-teardown=true --gather-logs-sizes=true --minStartupPods=8
```
wojtek-t, line 78: We're now using tests from perf-tests repo instead - tests were redesigned and moved around - not removed.

Do we have (and need) canaries for them?
[i.e. should we replace the comamnds here instead of removing?

wojtek-t: @marseel @mborsz @ameukam

SergeyKanzhelev: I glanced over the per-tests repo and I do not see any tests with the "[Feature:Performance]", but scrapper is still looking for it. Please consider cleaning up as well like here: kubernetes/kubernetes#106283

SergeyKanzhelev:
/remove-lifecycle stale

any updates on this? Let's remove empty job as it consumes resources and track the re-adding it back as an issue?

PR #24304, opened 2021-11-10
Fix env config in periodic capi tests in CAPA release 0.7

Ankitasw:
The test cases for periodic-cluster-api-provider-aws-capi-e2e-release-0-7 is skipped in the runs because of env `E2E_UNMANAGED_FOCUS` set as `[unmanaged] [Cluster API Framework]`. This PR updates this env so that tests runs properly in testgrid


Ankitasw:
cc @sedefsavas

PR #24316, opened 2021-11-10
Prow sub can work with GitHub app

chaodaiG:


prow/cmd/sub/main.go
```diff
@@ -107,6 +107,9 @@ func main() {
 	if flagOptions.github.TokenPath != "" {
 		tokens = append(tokens, flagOptions.github.TokenPath)
 	}
+	if flagOptions.github.AppPrivateKeyPath != "" {
```
alvaroaleman, line 110: This shouldnt be needed, it happens in the flagutil when you call `GitHubClientWithLogFields`: https://github.com/kubernetes/test-infra/blob/ef5dc9a4f6b4dc988050e83d438ac0484439070f/prow/flagutil/github.go#L272-L274


Below is another check for `flagOptions.github.TokenPath `, that one needs to also allow apps auth

chaodaiG: ah. right. that's strange then. I'll take another looks

chaodaiG: found it, fixing in https://github.com/kubernetes/test-infra/pull/24317, and by the way removing duplicated logic in the PR

PR #19710, opened 2020-10-27
Enable new log dumping mechanism for all scalability periodic jobs

tosi3k:
Follow-up to https://github.com/kubernetes/test-infra/pull/19663 which enabled the new log dumping mechanism for scalability master branch periodic jobs. This would additionally turn them on for 1.16-1.19 scalability CI jobs.

The PR adds a `preset-e2e-scalability-periodics-master` on the occasion that would facilitate gradual rollout of env var-enabled features in the scalability jobs with a following flow:
* master periodics: add env var to `preset-e2e-scalability-periodics-master`,
* all periodics: after some soak time and having green Testgrid move that env var to `preset-e2e-scalability-periodics`,
* periodics + presubmits: after some additional soak time and having green Testgrid move that env var to `preset-e2e-scalability-common` and/or `preset-e2e-kubemark-common`.

/sig scalability
/assign @jkaniuk @wojtek-t

PR #24322, opened 2021-11-11
Promote CSI changes to all presubmit jobs

leiyiz:
Same logic with https://github.com/kubernetes/test-infra/pull/24357, promote to preset-e2e-scalability-presubmits to enable in all presubmit tests (but not 1.19, 1.20, 1.21, 1.22 where we don't want to enable this).

@mborsz

### DIFF
--- a/config/jobs/kubernetes/sig-release/release-branch-jobs/1.19.yaml
+++ b/config/jobs/kubernetes/sig-release/release-branch-jobs/1.19.yaml
@@ -1149,6 +1149,8 @@ presubmits:
         - --tear-down-previous
         - --env=CL2_ENABLE_CLUSTER_OOMS_TRACKER=true
         - --env=CL2_ENABLE_VIOLATIONS_FOR_SCHEDULING_THROUGHPUT=false
+        - --env=DEPLOY_GCI_DRIVER=false
+        - --env=PROMETHEUS_STORAGE_CLASS_PROVISIONER=kubernetes.io/gce-pd
         - --test=false
         - --test-cmd=$GOPATH/src/k8s.io/perf-tests/run-e2e.sh
         - --test-cmd-args=cluster-loader2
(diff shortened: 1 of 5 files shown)


mborsz:
We have to enable this in periodics first. Otherwise, once we submit k/k changes the presubmit will pass, but our periodics will start failing.

I created https://github.com/kubernetes/test-infra/pull/24351

mborsz:
In general, the order of enabling new features to our tests should follow: https://github.com/kubernetes/test-infra/pull/19710#issue-730566534

It has been enabled in https://github.com/kubernetes/test-infra/pull/24356 in master-periodics.

The next step is to add preset-e2e-scalability-periodics, but this will enable this feature also for pre-master versions which are using old versions of perf-tests where we don't want to make any change.

Please prepare a PR that:
* Moves DEPLOY_GCI_DRIVER and PROMETHEUS_STORAGE_CLASS_PROVISIONER from preset-e2e-scalability-periodics-master to preset-e2e-scalability-periodics preset (around https://github.com/mborsz/test-infra/blob/csi/config/jobs/kubernetes/sig-scalability/sig-scalability-presets.yaml#L311)
* In the same PR, find all tests (in 1.19, 1.20, 1.21 and 1.22 directories) that are using preset-e2e-scalability-periodics AND NOT using preset-e2e-scalability-periodics-master and disable those flags there (e.g. add --env=DEPLOY_GCI_DRIVER=false --env=PROMETHEUS_STORAGE_CLASS_PROVISIONER=kubernetes.io/gce-pd) to make this PR is no-op in non-master prowjobs.

**Why we do have to add to preset-e2e-scalability-periodics, instead of simply leaving this in preset-e2e-scalability-periodics-master:**
The problem is that once we copy "master" to a dedicated 1.23 job, it should use a new setup. If we update only -master and not -periodics (which affects all versions), the env variables won't be copied to the new jobs.

mborsz:
/lgtm

We still need to promote to preset-e2e-scalability-common and preset-e2e-kubemark-common. Personally I would do that in this PR, instead of preset-e2e-scalability-presubmits, but we can submit this now. I will prepare a follow up PR to migrate it even further.

PR #24324, opened 2021-11-11
config/jobs: run no-stage on k8s-infra, drop extract

aojea:


config/jobs/kubernetes/sig-network/sig-network-misc.yaml
```diff
@@ -142,6 +142,7 @@ presubmits:
     branches:
     - master
     always_run: false
+    cluster: k8s-infra-prow-build
```
BenTheElder, line 145: this cluster is for required presubmits and release blocking jobs currently.

aojea: what should I use then?

aojea: this is optional

BenTheElder: you can just drop this field, it will use the default cluster (`cluster: default`)

config/jobs/kubernetes/sig-network/sig-network-misc.yaml
```diff
@@ -168,7 +169,6 @@ presubmits:
         - /workspace/scenarios/kubernetes_e2e.py
         args:
         - --build=quick
-        - --extract=local
```
BenTheElder, line 171: we only dropped the extract, we should drop the `--stage` (this is the one that should fix the job)

aojea: `and` o `or`

BenTheElder: And

BenTheElder (approved):
/lgtm
/approve

PR #24328, opened 2021-11-11
OWNERS(justaugustus): Prune extraneous reviewer roles

justaugustus:
Taking an opportunity to prune myself from some aliases/review paths
to reduce my workload.

If:
- I am no longer reviewing in an area, prune
- I was already an approver for the area, prune from review path

OWNERS_ALIASES are also updated to match their analogs in k/community.

### DIFF
--- a/config/jobs/kubernetes/generated/OWNERS
+++ b/config/jobs/kubernetes/generated/OWNERS
@@ -2,11 +2,10 @@
 
 reviewers:
 - release-engineering-reviewers
-- justaugustus
 approvers:
 - release-engineering-approvers
-- justaugustus
 emeritus_approvers:
+- justaugustus    # 2021-11-15: https://github.com/kubernetes/test-infra/pull/24328
 - Katharine
 - krzyzacy
 - yguo0905
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -33,8 +33,7 @@ aliases:
   sig-apps-leads:
     - janetkuo
     - kow3ns
-    - mattfarina
-    - prydonius
+    - soltysh
   sig-architecture-leads:
     - derekwaynecarr
     - dims
@@ -44,11 +43,13 @@ aliases:
     - enj
     - liggitt
     - mikedanese
-    - tallclair
+    - ritazh
   sig-autoscaling-leads:
+    - gjtempleton
     - mwielgus
   sig-cli-leads:
-    - pwittrock
+    - KnVerey
+    - eddiezane
     - seans3
     - soltysh
   sig-cloud-provider-leads:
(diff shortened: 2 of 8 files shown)


justaugustus:
/assign @BenTheElder @spiffxp @fejta 

FYI to various groups
RelEng:
@kubernetes/release-engineering 

Cloud Provider Azure:
@kubernetes/provider-azure 

Cluster API Azure:
@alexeldeib @CecileRobertMichon @devigned @nader-ziada

images/kubekins-e2e/OWNERS
```diff
@@ -5,7 +5,6 @@ reviewers:
 - BenTheElder
 - cpanato
 - jeremyrickard
-- justaugustus
```
BenTheElder, line 8: hmm we generally prefer to have approvers be a subset of reviewers, to spread the review load 🥕

dims: +1

justaugustus: Fixed.

justaugustus: > hmm we generally prefer to have approvers be a subset of reviewers, to spread the review load 🥕

But... if we feel this way as a project, we should probably update https://go.k8s.io/owners#maintaining-owners-files:

> Good examples of OWNERS usage:
>
> - team aliases are used that correspond to sigs
> - there are more `reviewers` than `approvers`
> - the `approvers` are not in the `reviewers` section
> - OWNERS files that are regularly updated (at least once per release)

BenTheElder: Thanks for fixing.

At the very least this repo / sig testing operates this way. I would encourage all repos to do so, unclear if the org as a whole will do this and I'm a bit weary of trying to effect org-wide policy at the moment.

config/jobs/kubernetes/generated/OWNERS
```diff
@@ -2,7 +2,6 @@
 
 reviewers:
 - release-engineering-reviewers
-- justaugustus
 approvers:
 - release-engineering-approvers
 - justaugustus
```
BenTheElder, line 7: same thing here? https://github.com/kubernetes/test-infra/pull/24328#discussion_r747707785

BenTheElder (commented):
also feel free to make an entry in emeritus_approvers, which I believe is documented (opt-in) policy now?

config/jobs/kubernetes/generated/OWNERS
justaugustus: Fixed!

BenTheElder (approved):
thanks for cleaning up ALIASES and for your patience, and your reviews! :-)
/lgtm
/approve

PR #24330, opened 2021-11-11
skipping [Driver: gcepd] for presubmit jobs

leiyiz:
storage presubmit jobs need to skip in-tree gcepd tests to pass when `CSIMigrationGCE` is on


leiyiz:
@msau42

mattcary:
/lgtm

Note these tests will be picked up by the cloud-provider-gcp test suite (or the sig-storage tests should start using gce clusters launched from cloud-provider-gcp rather than in-tree).

saad-ali (approved):
/lgtm
/approve

Issue #23832, opened 2021-09-29
Building prow images without bazel


chaodaiG:
**What would you like to be added**:

Building prow images without bazel, the related bazel rules are:

- prow-image
- container-image
- go-image

Things to be considered:

- This will need to co-exist with current bazel rules for a short while, delete Prow images test targets from bazel once this confirmed working
- Building prow containers from [rules-docker](https://github.com/bazelbuild/rules_docker), under the hood using distroless language runtime
- Deck is a special case, images built with a set of rules including rollup_bundle, will likely benefit from #23829 
- `/images` don’t use bazel

**Why is this needed**:

This is part of #23796 

**Recommended testing approach**:
- Make test invokable by `make test`
- Create a PullRequest with change, manually run job by commenting `/test pull-test-infra-test-beta`
- [Optional]: if there is any tool that needs to exist in test image, update image at https://github.com/kubernetes/test-infra/blob/6e0d21ee50d690c8f78239257f588021d2ed57b5/config/jobs/kubernetes/test-infra/test-infra-presubmits.yaml#L145

/sig testing
/help

chaodaiG:
Most of prow binaries are purely written in Go, the only known exception is `Deck`, where it has quite a few typescripts at [prow/cmd/deck/static](https://github.com/kubernetes/test-infra/tree/master/prow/cmd/deck/static), so this work is probably better to be done after #23829 is landed.

chaodaiG:
rules_docker achieves images building and pushing by a mixture of bazel rules and Go packages, for example https://github.com/bazelbuild/rules_docker/blob/master/container/go/cmd/join_layers/join_layers.go for creating image tarballs. It's not going to be an easy task using the same stack that rules_docker uses.

Alternatively, this can be achieved by Dockerfile + Makefile. Production prow images can be published with Google Cloud Build, and local developers will be given the option of building/pushing using local docker, which will also be used by prow integration tests which require pushing images to local registry(GCB doesn't support).

For Google Cloud Build, technically could reuse https://github.com/kubernetes/test-infra/tree/master/config/jobs/image-pushing, however it only supports a single image per job, which means lots of prow jobs just for pushing prow images. Might prefer creating a single cloudbuild.yaml for pushing all prow images.

chaodaiG:
I'll start playing with images that are pure Golang

chaodaiG:
[...]

PR #24331, opened 2021-11-11
Dockerfile for building prow images

chaodaiG:


chaodaiG:
An idea of building prow images based on thoughts from https://github.com/kubernetes/test-infra/issues/23832#issuecomment-964579773.

/cc @cjwagner @BenTheElder

images/prow-base/Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+
+ARG IMAGE_NAME
+ENV IMAGE_NAME=${IMAGE_NAME}
+
+RUN echo "Image is: ${IMAGE_NAME}"
+
+COPY . .
+
+# https://github.com/moby/moby/issues/5509#issuecomment-633616514
+# CMD doesn't allow variable expansion, the best can be done for now is using a
```
BenTheElder, line 29: before docker supported --build-arg, kubernetes used a "template" + sed.

Maybe instead of "placeholder-binary" we can just name it compatibly with the bazel symlink? (something something app.binary right?)

BenTheElder:
I guess this doesn't handle deck though?
I feel like we might want to sort that out before proceeding

chaodaiG:
> I guess this doesn't handle deck though? I feel like we might want to sort that out before proceeding

yeah this does not. I have no intention to merge this PR without we know how to build deck. The work being done here so far is experimenting with ideas around how to ensure that the new images building strategy works with both GCB and local

chaodaiG:
Played with Cuelang, and generated Google Cloud Build yaml out of .cue template. Have tested on my toy project and it worked so far.

Building multi-arch images with `buildx` is blocked on s390x, so building individually just like how they are built currently with bazel rules.

chaodaiG:
/close
in favor of ko

PR #24340, opened 2021-11-12
Configure liveness and readiness probes for prow-controller-manager.

cjwagner:
Follow up to https://github.com/kubernetes/test-infra/pull/24042

Since the previous PR, `prow-controller-manager` can now have a slower startup due to retrying build manager creation. This PR should change the deployment strategy to allow two pods to coexist during deployment and it should make the rolling deployment strategy aware of when the build managers are ready for use so that the old pod is not removed until the new pod is ready.

I just ready all the k8s docs I could find on liveness/readiness probes and a couple articles about potential pitfalls. I think this configuration should give us the behavior we want without risk of crashlooping.
/assign @alvaroaleman @chaodaiG 
/hold

### DIFF
--- a/config/prow/cluster/prow_controller_manager_deployment.yaml
+++ b/config/prow/cluster/prow_controller_manager_deployment.yaml
@@ -22,6 +22,11 @@ metadata:
 spec:
   # Mutually exclusive with plank. Only one of them may have more than zero replicas.
   replicas: 1
+  strategy:
+    type: RollingUpdate
+    rollingUpdate:
+      maxSurge: 1
+      maxUnavailable: 1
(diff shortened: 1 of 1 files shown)


alvaroaleman (approved):

config/prow/cluster/prow_controller_manager_deployment.yaml
```diff
@@ -22,6 +22,11 @@ metadata:
 spec:
   # Mutually exclusive with plank. Only one of them may have more than zero replicas.
   replicas: 1
+  strategy:
```
chaodaiG, line 25: would be nice to also add comment here as being done below. However not worth blocking

PR #24341, opened 2021-11-13
Use KOPS_IRSA for all kOps IRSA tests

johngmyers:
/cc @rifelpet

### DIFF
--- a/config/jobs/kubernetes/kops/kops-periodics-misc2.yaml
+++ b/config/jobs/kubernetes/kops/kops-periodics-misc2.yaml
@@ -271,7 +271,7 @@ periodics:
     testgrid-days-of-results: '90'
     testgrid-tab-name: kops-grid-scenario-ipv6-cilium
 
-# {"cloud": "aws", "container_runtime": "containerd", "distro": "u2004", "extra_flags": "--api-loadbalancer-type=public --override=cluster.spec.serviceAccountIssuerDiscovery.discoveryStore=s3://k8s-kops-prow/e2e-dc69f71486-5831d.test-cncf-aws.k8s.io/discovery --override=cluster.spec.serviceAccountIssuerDiscovery.enableAWSOIDCProvider=true --override=cluster.spec.iam.useServiceAccountExternalPermissions=true", "k8s_version": "latest", "kops_channel": "alpha", "kops_version": null, "networking": "kubenet"}
+# {"cloud": "aws", "container_runtime": "containerd", "distro": "u2004", "extra_flags": "--api-loadbalancer-type=public --override=cluster.spec.serviceAccountIssuerDiscovery.discoveryStore=s3://k8s-kops-prow/e2e-kops-grid-scenario-service-account-iam/discovery --override=cluster.spec.serviceAccountIssuerDiscovery.enableAWSOIDCProvider=true --override=cluster.spec.iam.useServiceAccountExternalPermissions=true", "k8s_version": "latest", "kops_channel": "alpha", "kops_version": null, "networking": "kubenet"}
 - name: e2e-kops-grid-scenario-service-account-iam
   cron: '45 6-23/8 * * *'
   labels:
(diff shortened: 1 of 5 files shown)


rifelpet:
KOPS_IRSA is only used by the scenario scripts:

https://github.com/kubernetes/kops/blob/88623ef1862ae59782980755cfd6ce96bb659a13/tests/e2e/scenarios/lib/common.sh#L49-L53

so these jobs that invoke kubetest2-kops directly won't recognize it. We could add these --overrides in a similar way in build_jobs.py though

johngmyers:
Perhaps we should add a `irsa` argument or somesuch? Using `env` strikes me as non-obvious.

I do think we want to move to having the tests use IRSA by default, turning off only if k8s < ~1.22 or it's a special test for non-IRSA. That would be more likely to catch permissions issues, at the cost of failing to catch when the instance role gets too big.

olemarkus:
We have integration tests (many addons) that should catch most cases of policies being too big.

config/jobs/kubernetes/kops/kops-periodics-misc2.yaml
```diff
@@ -300,7 +300,7 @@ periodics:
           -v 2 \
           --up --down \
           --cloud-provider=aws \
-          --create-args="--image='099720109477/ubuntu/images/hvm-ssd/ubuntu-focal-20.04-amd64-server-20211021' --channel=alpha --networking=kubenet --container-runtime=containerd --api-loadbalancer-type=public --override=cluster.spec.serviceAccountIssuerDiscovery.discoveryStore=s3://k8s-kops-prow/e2e-dc69f71486-5831d.test-cncf-aws.k8s.io/discovery --override=cluster.spec.serviceAccountIssuerDiscovery.enableAWSOIDCProvider=true --override=cluster.spec.iam.useServiceAccountExternalPermissions=true" \
+          --create-args="--image='099720109477/ubuntu/images/hvm-ssd/ubuntu-focal-20.04-amd64-server-20211021' --channel=alpha --networking=kubenet --container-runtime=containerd --api-loadbalancer-type=public --override=cluster.spec.serviceAccountIssuerDiscovery.discoveryStore=s3://k8s-kops-prow/e2e-kops-grid-scenario-service-account-iam/discovery --override=cluster.spec.serviceAccountIssuerDiscovery.enableAWSOIDCProvider=true --override=cluster.spec.iam.useServiceAccountExternalPermissions=true" \
```
rifelpet, line 303: Is this change to move the discovery objects outside of the configBase intentional? This means `kops delete cluster` will no longer delete them which may lead to strange behavior if a prow job run inadvertently uses a previous run's objects.

config/jobs/kubernetes/kops/kops-periodics-misc2.yaml
```diff
@@ -1204,8 +1204,6 @@ periodics:
       args:
       - ./tests/e2e/scenarios/aws-ebs-csi/run-test.sh
```
rifelpet, line 1205: For the scenario script tests we either need to keep the KOPS_IRSA env var or pass an OVERRIDES env var containing the same --override flags. As-is, the `extra_flags` argument to `build_test()` and `presubmit_test()` isn't used when a scenario script is specified.

config/jobs/kubernetes/kops/kops-periodics-misc2.yaml
johngmyers: I'm trying to put them somewhere that is both specific to the job (so concurrent jobs don't step on each other) and calculatable by that function in build_jobs.py.

Normally the discovery objects would not be inside of configBase since they have significantly different access control. They normally shouldn't even be in the same bucket as the configBase.

I don't think there's much practical risk of a job using a previous run's discovery objects, but perhaps we should make sure delete cluster deletes the discovery objects.

rifelpet (approved):

PR #24343, opened 2021-11-13
Support application default credentials in gcsweb

barp:
This PR adds support for using default application credentials when using gcsweb.


cjwagner (approved):
/test all

PR #24347, opened 2021-11-14
Fix typo with cherrypicker in README

fmuyassarov:


petr-muller (commented):
The plugin's name is actually `cherrypick` so I think the README is correct:

https://github.com/kubernetes/test-infra/blob/3c697c581299d6b125a7c55df9460273bba631d3/prow/external-plugins/cherrypicker/server.go#L42

fmuyassarov:
@petr-muller then that's my bad. I thought it should be cherrypicker based on https://github.com/kubernetes/test-infra/blob/42e026a43088276c30fb1f1f4e0b0a6fdedfe2b5/config/prow/plugins.yaml#L1218, and we also named as cherrypicker. But perhaps names doesn't matter much ?

PR #24353, opened 2021-11-15
Migrate pull-kubernetes-kubemark-e2e-gce-scale to community-owned infrastructure

marseel:
/assign @tosi3k

### DIFF
--- a/config/jobs/kubernetes/sig-scalability/sig-scalability-presubmit-jobs.yaml
+++ b/config/jobs/kubernetes/sig-scalability/sig-scalability-presubmit-jobs.yaml
@@ -356,6 +356,7 @@ presubmits:
           privileged: true
 
   - name: pull-kubernetes-kubemark-e2e-gce-scale
+    cluster: k8s-infra-prow-build
     always_run: false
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes/sig-scalability/sig-scalability-presubmit-jobs.yaml
```diff
@@ -393,7 +394,7 @@ presubmits:
         - --flush-mem-after-build=true
         - --gcp-node-size=e2-standard-8
         - --gcp-nodes=84
-        - --gcp-project=k8s-presubmit-scale
+        - --gcp-project-type=scalability-project
```
wojtek-t, line 397: Can scalability-project handle 5k-node kubemark?

Don't we want a different class for those projects?

marseel: yes, scalability-project can handle 5k kubemarks (our kubemark 5k CI already uses this pool). 
We don't have different class right now.

PR #24356, opened 2021-11-15
Use CSI in GKE master periodics. Attempt #2

mborsz:
Reverts kubernetes/test-infra#24352

Let's try to enable it again. With https://github.com/kubernetes/perf-tests/pull/1938 it may work.

/assign @tosi3k 
/cc @leiyiz


mborsz:
First run with this: https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/ci-kubernetes-e2e-gci-gce-scalability/1460260346378850304

PR #24357, opened 2021-11-15
Promote CSI changes to all periodic jobs

mborsz:
Promotes https://github.com/kubernetes/test-infra/pull/24356 to preset-e2e-scalability-periodics to enable in all periodics tests (but not 1.19, 1.20, 1.21, 1.22 where we don't want to enable this).

**Why this approach vs simply keeping this in preset-e2e-scalability-periodics-master forever?**
Here we explicitly say which versions should use old logic. Once we start running tests for 1.23 (so the current master), we will use the new logic (which is desired), while the previous approach would require us to manually remember to migrate some env that time (and each time we add a new k8s version).

/assign @tosi3k 

/hold
Hold for now, until we soak https://github.com/kubernetes/test-infra/pull/24356.
Feel free to unhold once we see at least 1 successful run for both:
* https://k8s-testgrid.appspot.com/sig-scalability-kubemark#kubemark-100
* https://k8s-testgrid.appspot.com/sig-scalability-gce#gce-cos-master-scalability-100

Including https://github.com/kubernetes/test-infra/pull/24356 (you can check by going to a yaml file for a given test and checking env variables).

/cc @leiyiz

### DIFF
--- a/config/jobs/kubernetes/sig-release/release-branch-jobs/1.19.yaml
+++ b/config/jobs/kubernetes/sig-release/release-branch-jobs/1.19.yaml
@@ -274,6 +274,8 @@ periodics:
       - --kubemark-nodes=500
       - --provider=gce
       - --env=CL2_ENABLE_CLUSTER_OOMS_TRACKER=true
+      - --env=DEPLOY_GCI_DRIVER=false
+      - --env=PROMETHEUS_STORAGE_CLASS_PROVISIONER=kubernetes.io/gce-pd
       - --test=false
       - --test_args=--ginkgo.focus=xxxx
       - --test-cmd=$GOPATH/src/k8s.io/perf-tests/run-e2e.sh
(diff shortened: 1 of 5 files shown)


mborsz:
Jobs including this change:
* gce-cos-master-scalability-100 job: https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/ci-kubernetes-e2e-gci-gce-scalability/1460260346378850304 (about to pass)
* kubemark-100 job: https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/ci-kubernetes-kubemark-100-gce/1460271491932229632 (I just started it manually, waiting for result)

mborsz:
It fails for kubemark, as in https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/run-e2e.sh#L26 we should use KUBEMARK_ROOT_KUBECONFIG if we are in kubemark test. Otherwise we install csi driver in incorrect master.

mborsz:
https://github.com/kubernetes/perf-tests/pull/1939 tries to fix kubemark

leiyiz:
/hold cancel
reason: https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/ci-kubernetes-kubemark-500-gce/1460360949281067008 passed

PR #24367, opened 2021-11-16
Set k8s version to 1.21 for Azure Disk CSI Driver V2 tests

edreed:


andyzhangx (approved):
/lgtm
/approve

edreed:
@andyzhangx Could you please ok-to-test this PR? We need to upgrade the k8s version we test for the V2 driver.

PR #24374, opened 2021-11-16
Set K8s version to 1.21 for Azure Disk CSI Driver V2 multi-az test

edreed:


andyzhangx (approved):
/lgtm
/approve
/ok-to-test

PR #24377, opened 2021-11-16
Allow Override PR History link

chaodaiG:
PR history link /pr-history from deck is currently only supported for GitHub repos, which turned out to be internal error for gerrit based prow. Allow overriding the template could possibly let users pointing to `/repo=<REPO>&pull=<PULL>` etc

### DIFF
--- a/prow/cmd/deck/main.go
+++ b/prow/cmd/deck/main.go
@@ -32,7 +32,6 @@ import (
 	"net/url"
 	"os"
 	"path"
-	"strconv"
 	"strings"
 	"time"
 
@@ -101,6 +100,8 @@ const (
 	DecorationConfig string = "decoration_config"
 	// PodSpec maps to the serialized value of <ProwJob>.Spec.PodSpec.
 	PodSpec string = "pod_spec"
+
+	defaultPRHistLinkTemplate = "/pr-history?org={{.Org}}&repo={{.Repo}}&pr={{.Number}}"
 )
 
 type options struct {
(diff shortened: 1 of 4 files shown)


prow/cmd/deck/main.go
```diff
@@ -1022,7 +1023,14 @@ lensesLoop:
 	prHistLink := ""
 	org, repo, number, err := sg.RunToPR(src)
 	if err == nil && !cfg().Deck.Spyglass.HidePRHistLink {
-		prHistLink = "/pr-history?org=" + org + "&repo=" + repo + "&pr=" + strconv.Itoa(number)
+		PRHistLinkTemplate := cfg().Deck.Spyglass.PRHistLinkTemplate
```
e-blackwelder, line 1026: local variables should be lowerCamelCase, right? (Just like prHistLink)

prow/cmd/deck/main.go
```diff
@@ -1140,6 +1148,23 @@ lensesLoop:
 	return viewBuf.String(), nil
 }
 
+func PRHistLink(PRHistLinkTemplate, org, repo string, number int) (string, error) {
+	tmp, err := template.New("t").Parse(PRHistLinkTemplate)
+	if err != nil {
+		return "", fmt.Errorf("failed compiling template %q: %v", PRHistLinkTemplate, err)
+	}
+	tmpBuff := bytes.Buffer{}
+	if err = tmp.Execute(&tmpBuff, struct {
```
e-blackwelder, line 1157: Trying to think of any vulnerabilities... Looks vulnerable to [XSS](https://sucuri.net/guides/what-is-cross-site-scripting/). For example, what happens if I use a template like: `javascript:doSomethingEvil()`

... I think "sanitizing/validating" a template would be tricky because of HTML entities & URL encoding.

I forget if Spyglass configuration is considered  "safe" (requiring PR reviews, plus or minus in-repo config), but I'm assuming two bad actors could collaborate to merge it??

chaodaiG: Technically this is possible. But this field can only be changed in `config.yaml` file, which is supposed to be guarded by `OWNERS` file, to ensure that changes here can only be merged if a prow admin approve the change. So we might be fine?

prow/cmd/deck/main.go
chaodaiG: good point, updated

prow/cmd/deck/main.go
e-blackwelder: It's possible that a (new/distracted/etc) OWNER might not pay full attention to it.

I'll leave the final say to you/others.

chaodaiG: Prow had several other fields like https://github.com/kubernetes/test-infra/pull/24377/files#diff-88e6b42a331a7aa9260f9a830961e3ba3c77e0e6d781465378b510d7abc1517cR997 could also be exposed to this kind of problem. Generally this should be easy to identify during code review

alvaroaleman (approved):

PR #24379, opened 2021-11-16
Use current ingress-gce image from gcr.io/k8s-ingress-image-push

swetharepakula:
Update the ingress-gce glbc image to correct the issue seen in https://github.com/kubernetes/kubernetes/issues/106402

/assign @freehan @bowei 
/cc @aojea


aojea:
/lgtm
/approve
Thanks

PR #24381, opened 2021-11-16
Ignore EasyCLA Context on kubernetes-client,csi,sigs orgs

mrbobbytables:
Follow up to #23648
/assign @cblecker @BenTheElder

### DIFF
--- a/config/prow/config.yaml
+++ b/config/prow/config.yaml
@@ -657,6 +657,15 @@ tide:
         repos:
           dashboard:
             from-branch-protection: true
+      kubernetes-client:
+        optional-contexts:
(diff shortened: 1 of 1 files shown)


BenTheElder (approved):
/lgtm
/approve

brendandburns:
Should there be a unit test for this configuration to prevent similar mistakes in the future?

BenTheElder:
I don't think there's a test for this -- EasyCLA is not controlled by this repo at all (it's run by the CNCF, not Kubernetes), but since EasyCLA has been problematic we want to configure the merge robot to ignore it even though it's posting a status.

AIUI we intentionally don't currently force all of the organizations to have the same settings (e.g. to roll out a new setting incrementally), if we decided we wanted to do that for all settings we should test for it though.

PR #23777, opened 2021-09-27
config: shrink and begin migrating away from node-e2e boskos pool

spiffxp:
Related:
- Part of: https://github.com/kubernetes/test-infra/issues/23291
- Followup to: https://github.com/kubernetes/test-infra/pull/23292
- Followup to: https://github.com/kubernetes/test-infra/pull/17284

Apparently doing the project addition and removal in a single PR wasn't enough to cause boskos to grow the node-e2e pool, so undoing that change to reflect reality

Also, there is nothing special about the projects in the node-e2e pool, so let's stop using it, and begin to migrate jobs over to the main gce-project pool

I chose an arbitrary group of presubmits first, since I suspect presubmits are the main reason the pool gets exhausted occasionally

Issue #23291, opened 2021-08-18
boskos: node-e2e pool is getting exhausted


spiffxp:
**What happened**:

Alerts have been firing in #testing-ops more frequently over time:

>[FIRING:1] Resource exhausted (0% free) (node-e2e-project 104.197.27.114:9090 prow-monitoring/prow high node-e2e-project)
>@test-infra-oncall The Boskos resource "node-e2e-project" has been exhausted (currently 0% free). See the Boskos resource usage dashboard.

**What you expected to happen**:

For boskos not to run out of capacity

**How to reproduce it (as minimally and precisely as possible)**:

Wait for the project to receive enough traffic

**Please provide links to example occurrences, if any**:

**Anything else we need to know?**:

The node-e2e pool was an arbitrary pool made during a time when node-e2e jobs were pinned to a single GCP project that was shut down for security concerns.  There is nothing special about how the projects within it are configured.  We decided to use a pool rather than pin to a project, and opted for a new `node-e2e` pool rather than the main `gce` pool in case more projects got flagged for security concerns, so as to minimize the blast radius.

It now appears that there is more traffic or there are more jobs using this pool, and we're starting to run out of capacity more frequently. 

![Screen Shot 2021-08-18 at 6 43 13 AM](https://user-images.githubusercontent.com/49258/129908873-1b999870-55b4-4cd3-bc14-5601cea36c8b.png)
(from: https://monitoring.prow.k8s.io/d/wSrfvNxWz/boskos-resource-usage?panelId=9&fullscreen&orgId=1&from=now-30d&to=now)

Since the `node-e2e` pool was initially created by arbitrarily stealing projects from the `gke` pool, it makes sense to do so again to grow the capacity of the pool. Alternatively, we should get rid of the `node-e2e` pool and dump everything back into the larger `gce` pool.

/area boskos
/sig node
/sig testing
/priority critical-urgent

spiffxp:
Opting to grow the pool: https://github.com/kubernetes/test-infra/pull/23292

ameukam:
Related : https://github.com/kubernetes/test-infra/issues/23232.

ehashman:
Discussed previously at [SIG Node CI 08/12](https://docs.google.com/document/d/1fb-ugvgdSVIkkuJ388_nhp2pBTy_4HEVg5848Xy7n5U/edit#bookmark=id.f8kr3n4hgwdj).

/assign @SergeyKanzhelev

[...]

PR #24384, opened 2021-11-17
Remove kubelet-serial-containerd from node-e2e-project

adisky:
The job [ node-kubelet-serial-containerd ](https://testgrid.k8s.io/sig-node-release-blocking#node-kubelet-serial-containerd) is failing since log with the error 

```
Something went wrong: failed to prepare test environment: --provider=gce boskos failed to acquire project: resources not found
```

I have seen boskos releated constraint issues [here](https://github.com/kubernetes/test-infra/issues/23291)  and [here](https://github.com/kubernetes/test-infra/pull/23777) . I am not sure the exact reason why this job is failing, from the discussions above we want to migrate away from `node-e2e-project`, Since this job is anyways failing with boskos issue we can try to migrate it

/cc @spiffxp @endocrimes 
/sig node

### DIFF
--- a/config/jobs/kubernetes/sig-node/node-kubelet.yaml
+++ b/config/jobs/kubernetes/sig-node/node-kubelet.yaml
@@ -265,7 +265,6 @@ periodics:
       - --scenario=kubernetes_e2e
       - --
       - --deployment=node
-      - --gcp-project-type=node-e2e-project
       - --gcp-zone=us-west1-b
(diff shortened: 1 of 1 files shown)


endocrimes:
/lgtm
/assign @dims 

looks like some of the earlier failures are unrelated to these which we should take a look at when they run again

SergeyKanzhelev (approved):

PR #24388, opened 2021-11-17
jobs/sig-node: remove ci-kubernetes-node-kubelet-kubetest2 from release-blocking dashboards

palnabarun:
Ref: https://github.com/kubernetes/kubernetes/issues/106478#issuecomment-971517098

/cc @Namanl2001 
/assign @dims @SergeyKanzhelev

### DIFF
--- a/config/jobs/kubernetes/sig-node/node-kubelet.yaml
+++ b/config/jobs/kubernetes/sig-node/node-kubelet.yaml
@@ -54,7 +54,7 @@ periodics:
   annotations:
     fork-per-release: "true"
     fork-per-release-periodic-interval: 1h 2h 6h 24h
-    testgrid-dashboards: sig-release-master-blocking, sig-node-release-blocking
+    testgrid-dashboards: sig-node-release-blocking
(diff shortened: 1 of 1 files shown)


Namanl2001 (approved):
Thanks, @palnabarun

config/jobs/kubernetes/sig-node/node-kubelet.yaml
```diff
@@ -54,7 +54,7 @@ periodics:
   annotations:
     fork-per-release: "true"
     fork-per-release-periodic-interval: 1h 2h 6h 24h
-    testgrid-dashboards: sig-release-master-blocking, sig-node-release-blocking
+    testgrid-dashboards: sig-node-release-blocking
```
adisky, line 57: It could be added to `sig-node-critical` tab rather than release blocking one
https://testgrid.k8s.io/sig-node-critical

palnabarun: A possible follow up then? @Namanl2001

PR #24393, opened 2021-11-17
Add new CEL ValidationRules to alpha features we e2e test

jpbetz:
This appears to be the primary test we use to test alpha features. It's accessible via `/test pull-kubernetes-e2e-gce-alpha-features`.

Feature being tested: https://github.com/kubernetes/enhancements/tree/master/keps/sig-api-machinery/2876-crd-validation-expression-language


jpbetz:
@cheftako I got a request to add alpha e2e testing of the CEL alpha feature. I'm a bit unclear on why the only alpha feature test job is under sig-cloud-provider of it it's the right job to add tests to, but I couldn't find any else and it appears that this job runs some alpha feature tests that are not cloud provider specific, so maybe this is the right place.

PR #24399, opened 2021-11-17
Generic-autobumper can opt out assigning to oncall

chaodaiG:
By the way give it the capability of skipping assignment to anyone

/assign @mpherman2


mpherman2:
/lgtm
/hold
holding in case someone has objections, but I support this

fejta (approved):

chaodaiG:
/unhold
This PR only only touches https://github.com/kubernetes/test-infra/blob/master/prow/cmd/generic-autobumper/main.go, I believe it's not used by downstream, adding @alvaroaleman for awareness just in case

PR #24400, opened 2021-11-18
kubetest: Remove naming requirement for k/k directory

ehashman:
My kubernetes clone is at `~/src/k8s`, once I removed this check kubetest ran fine.

Is this useful?


dims:
/close

we can revisit when we get all things working in k/k repo without GOPATH.

PR #24403, opened 2021-11-18
attempt to revive the node alpha presubmits

SergeyKanzhelev:
- Removed "cluster"
- Changed quotes format in `'true'` -> `"true"`

I'm puzzled why it is not working. Compared with "pull-kubernetes-node-e2e" - they are identical except these two ^^^ (and focus). But test fails on initialization so I assume either of above is wrong.

### DIFF
--- a/config/jobs/kubernetes/sig-node/sig-node-presubmit.yaml
+++ b/config/jobs/kubernetes/sig-node/sig-node-presubmit.yaml
@@ -367,9 +367,10 @@ presubmits:
         - --image-config-file=/go/src/k8s.io/test-infra/jobs/e2e_node/containerd/containerd-master/image-config.yaml
         - --image-config-dir=/home/prow
   - name: pull-kubernetes-node-e2e-alpha
-    cluster: k8s-infra-prow-build
     branches:
     - master
+    annotations:
+      testgrid-create-test-group: "true"
     always_run: false
     max_concurrency: 12
     labels:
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes/sig-node/sig-node-presubmit.yaml
```diff
@@ -367,9 +367,10 @@ presubmits:
         - --image-config-file=/go/src/k8s.io/test-infra/jobs/e2e_node/containerd/containerd-master/image-config.yaml
         - --image-config-dir=/home/prow
   - name: pull-kubernetes-node-e2e-alpha
-    cluster: k8s-infra-prow-build
```
adisky, line 370: is community owned cluster also reason for suspected failure?

SergeyKanzhelev: replicating the job "pull-kubernetes-node-e2e". I'm super puzzled by it. Maybe try just quotes first

SergeyKanzhelev: wdyt?

adisky: agreed lets try just quotes first

adisky (commented):
passing this job is imp for grpc probe feature @SergeyKanzhelev 
/lgtm

PR #24410, opened 2021-11-18
sig-instrumentation: increase code coverage for JSON periodic job

pohly:
We need to trigger more log calls to catch potential mistakes and to get
representative log output for components under load.

Storage tests were chosen because a) they exercise quite a bit of additional
code in kubelet and b) because I know them.

/wg structured-logging
/assign @serathius

PR #24411, opened 2021-11-18
Test full conformance suite for JSON, match kind tests

ehashman:
Cribbing off of https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes-sigs/kind/kind.yaml#L123

/assign @pohly
/cc @BenTheElder 

/wg structured-logging

### DIFF
--- a/config/jobs/kubernetes/sig-instrumentation/sig-instrumentation-kind-periodics.yaml
+++ b/config/jobs/kubernetes/sig-instrumentation/sig-instrumentation-kind-periodics.yaml
@@ -8,8 +8,8 @@ periodics:
     testgrid-dashboards: sig-instrumentation-tests, sig-testing-kind
     testgrid-tab-name: kind-json-logging-master
     description: Conformance and storage tests with JSON logging enabled using sigs.k8s.io/kind
-    # TODO: anyone else?
-    testgrid-alert-email: bentheelder@google.com,siarkowicz@google.com,patrick.ohly@intel.com
+    # TODO: when job is stable, uncomment WG email for test failures
+    testgrid-alert-email: bentheelder@google.com,siarkowicz@google.com,patrick.ohly@intel.com #,kubernetes-wg-structured-logging+test-failures@googlegroups.com
     testgrid-num-columns-recent: '6'
     fork-per-release: "true"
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes/sig-instrumentation/sig-instrumentation-kind-periodics.yaml
```diff
@@ -43,13 +43,6 @@ periodics:
       # don't retry conformance tests
       - name: GINKGO_TOLERATE_FLAKES
         value: "n"
-      - name: FOCUS
-        value: \[Conformance\]|\[Driver:.csi-hostpath\]
```
pohly, line 47: This disables the storage tests that I just added in https://github.com/kubernetes/test-infra/pull/24410.

Is that intentional?

As I said in https://github.com/kubernetes/test-infra/pull/24410/files#r752422365, the config that you are changing here hasn't been used at all yet.

ehashman: It doesn't disable them; Focus is a filter/allow-list. If you don't specify it, you get all the tests.

BenTheElder: In this case FOCUS actually has to be . if you want all, because https://github.com/kubernetes-sigs/kind/blob/d1eecc46e30cac9d35cd32dc52677ef75ec22e18/hack/ci/e2e-k8s.sh#L225

pohly: https://github.com/kubernetes/test-infra/pull/24410 worked as intended, recent job runs executed conformance and storage tests.

pohly: > It doesn't disable them; Focus is a filter/allow-list. If you don't specify it, you get all the tests.

Removing the FOCUS variable causes the script to use the default, which is just `\[Conformance\]`. Therefore the change above removes the `\[Driver:.csi-hostpath\]` pattern and thus the storage tests that use that driver.

PR #24423, opened 2021-11-19
remove sig-storage tests from sig-network jobs

aojea:
The sig-network-gce is red for a long time since sig-storage did some changes 
Having jobs red can mask failures, and since nobody is going to look to the sig-storage failures in these jobs, is better to remove then and have better signal on the tests we are really interested on

https://testgrid.k8s.io/sig-network-gce#presubmit-network-ipvs,%20google-gce

### DIFF
--- a/config/jobs/kubernetes/sig-network/sig-network-misc.yaml
+++ b/config/jobs/kubernetes/sig-network/sig-network-misc.yaml
@@ -187,7 +187,7 @@ presubmits:
         - --gcp-zone=us-west1-b
         - --provider=gce
         - --ginkgo-parallel=30
-        - --test_args=--ginkgo.skip=\[Slow\]|\[Serial\]|\[Disruptive\]|\[Flaky\]|\[Feature:.+\] --minStartupPods=8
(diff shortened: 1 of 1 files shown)


aojea:
/assign @MrHohn 

unless you are tracking those jobs, I prefer to have the jobs healthy, because right now I had to stop the alerts because of the false positives

leiyiz:
Can you please revert this? we don't want to skip the entirety of `[sig-storage]`, just want to skip on `Driver:.gcepd` like https://github.com/kubernetes/test-infra/pull/24425

leiyiz:
I have changed sig-storage to gcepd in #24425

aojea:
> Can you please revert this? we don't want to skip the entirety of `[sig-storage]`, just want to skip on `Driver:.gcepd` like #24425

this are sig-network jobs only and sig-storage tests use to create a lot of noise and AFAIK there is no much interaction between storage and the network on these tests.

 is not possible to monitor these tests in another jobs under sig-storage https://testgrid.k8s.io/sig-storage?

PR #24425, opened 2021-11-20
skip gcepd test due to CSIMigrationGCE

leiyiz:
Fix: https://github.com/kubernetes/kubernetes/issues/106459


aojea:
do you like to monitor those tests under sig-network jobs?

PR #24433, opened 2021-11-22
fix: fetch error with depth 1

cuyl:
`clonerefs` failed when BaseRef is ahead of the BaseSHA. it is common when rerunning the post jobs.

job config:
```yaml
postsubmits:
  xxx/xxx:
  - name: post-job
    decorate: true
    clone_depth: 1
    skip_fetch_head: true
    skip_submodules: true
```
previous:
```
git fetch --depth 1 https://github.com/kubernetes/test-infra.git master
git checkout ed929b919c4b460353f867e055066e1ef8c522e2
fatal: reference is not a tree: ed929b919c4b460353f867e055066e1ef8c522e2
# Error: exit status 128
```
current:
```
git fetch --depth 1 https://github.com/kubernetes/test-infra.git ed929b919c4b460353f867e055066e1ef8c522e2
git checkout ed929b919c4b460353f867e055066e1ef8c522e2
```

### DIFF
--- a/prow/pod-utils/clone/clone.go
+++ b/prow/pod-utils/clone/clone.go
@@ -238,18 +238,22 @@ func (g *gitCtx) commandsForBaseRef(refs prowapi.Refs, gitUserName, gitUserEmail
 		commands = append(commands, g.gitFetch(fetchArgs...))
 	}
 
+	var fetchRef string
+	var target string
+	if refs.BaseSHA != "" {
+		fetchRef = refs.BaseSHA
+		target = refs.BaseSHA
+	} else {
+		fetchRef = refs.BaseRef
+		target = "FETCH_HEAD"
+	}
+
 	{
 		fetchArgs := append([]string{}, depthArgs...)
-		fetchArgs = append(fetchArgs, g.repositoryURI, refs.BaseRef)
+		fetchArgs = append(fetchArgs, g.repositoryURI, fetchRef)
 		commands = append(commands, g.gitFetch(fetchArgs...))
 	}
 
-	var target string
-	if refs.BaseSHA != "" {
-		target = refs.BaseSHA
-	} else {
-		target = "FETCH_HEAD"
-	}
 	// we need to be "on" the target branch after the sync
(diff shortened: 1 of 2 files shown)


cjwagner:
> `clonerefs` failed when BaseRef is ahead of the BaseSHA.

Do you mean the other way around? If the BaseSHA that is being checked out is older than the HEAD of the BaseRef then the SHA would be available for checkout. I'm assuming you are referring to the BaseSHA somehow being ahead of the BaseRef HEAD and therefore not available after fetching the BaseRef. 
Could you provide an example of when this could occur? I'm not sure what circumstances would allow for this.

> it is common when rerunning the post jobs.

Rerunning a job creates a new job with an identical ProwJob spec. This includes the git refs so its unclear to me how rerunning a job would lead to this. If you could share a concrete example that would be helpful.

Thanks for the contribution! The error you are describing seems a bit unexpected though so I want to make sure we're fixing this in the most appropriate way.
/test all

cjwagner:
Oh, I see. This is specifically a problem with postsubmits that set limited clone depth. In that case this makes sense, but I am a bit concerned that this could have unintended side effects.

@alvaroaleman @chaodaiG do you know if fetching the  BaseRef rather than the BaseSHA is intentional/significant? I can't think of anything off the top of my head.

cuyl:
I signed it

cuyl:
I signed it

alvaroaleman:
> @alvaroaleman @chaodaiG do you know if fetching the BaseRef rather than the BaseSHA is intentional/significant? I can't think of anything off the top of my head.

I don't think so. We did a similar change for presubmits (motivated by some GitHub issues) which to my knowledge didn't cause any issues: https://github.com/kubernetes/test-infra/pull/19928

/test all
/lgtm
/approve
/hold

chaodaiG:
Fetching only to the commit that triggered the postsubmit, imo is even better than fetching latest HEAD, as there is no reason for a postsubmit job to fetch source code that is not meant to be used. So
/lgtm
/approve
/unhold

PR #24434, opened 2021-11-22
capdo: add experimental job to use boskos to be able to push temporary container images

cpanato:
in order to test the cluster-api api upgrades, we need to push the capdo-controller to the registry and that need to be available to pull.

Looks like we can push temporary images when using boskos (https://kubernetes.slack.com/archives/C09QZ4DQB/p1637516130115600)  but not sure

this is needed by the WIP work here https://github.com/kubernetes-sigs/cluster-api-provider-digitalocean/pull/282

/assign @spiffxp


spiffxp (approved):
/approve
/lgtm

PR #24438, opened 2021-11-22
skip gcepd test for ci-kubernetes-e2e-gci-gce-containerd

leiyiz:
failing testgrid: https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/ci-kubernetes-e2e-gci-gce-containerd/1462622526948511744

Fix: the reason these test failed is due to `CSIMigrationGCE` turned on and no pdcsi driver installed in clusters brought up by k/k/kube-up. We can't install pdcsi driver by default in k/k due to ongoing effort to extract cloud-provider specific components and we don't want to maintain multiple sets of manifests all over places. So we need to skip them for now and eventually migrate these tests to use cloud-provider-gcp for cluster-up.

### DIFF
--- a/config/jobs/kubernetes/sig-node/containerd.yaml
+++ b/config/jobs/kubernetes/sig-node/containerd.yaml
@@ -1124,7 +1124,9 @@ periodics:
       - --gcp-zone=us-west1-b
       - --ginkgo-parallel=30
       - --provider=gce
-      - --test_args=--ginkgo.skip=\[Slow\]|\[Serial\]|\[Disruptive\]|\[Flaky\]|\[Feature:.+\]|\[NodeFeature:RuntimeHandler\] --minStartupPods=8
+      # [Driver: gcepd] tests are skipped because CSIMigrationGCE is on in 1.23 and can be un-skipped once the cluster-up process
+      # uses cloud-provider-gcp. see issue https://github.com/kubernetes/cloud-provider-gcp/issues/293
+      - --test_args=--ginkgo.skip=\[Slow\]|\[Serial\]|\[Disruptive\]|\[Driver:.gcepd\]|\[Flaky\]|\[Feature:.+\]|\[NodeFeature:RuntimeHandler\] --minStartupPods=8
       - --timeout=50m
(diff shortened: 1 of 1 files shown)


SergeyKanzhelev:
Can you please update the description with the explanation why they are failing and why skipping is the best approach? For example, if test requires special set up - it may be better to add a `[Feature:` tag to tests

leiyiz:
> Can you please update the description with the explanation why they are failing and why skipping is the best approach? 

done

> it may be better to add a `[Feature:` tag to tests

Although it would be more convenient to just add such tag where we have `[Driver: gcepd]` tag, we'd like to keep track of how many skips we did so that no coverage is lost silently

SergeyKanzhelev:
Is SIG storage OK with loosing this test coverage? Is there other tests that test it? Should we just disable feature gate here for now?

config/jobs/kubernetes/sig-node/sig-node-config.yaml
```diff
@@ -21,7 +21,7 @@ periodics:
       - --gcp-zone=us-west1-b
       - --ginkgo-parallel=30
       - --provider=gce
-      - --test_args=--ginkgo.skip=\[Slow\]|\[Serial\]|\[Disruptive\]|\[Flaky\]|\[Feature:.+\]|\[NodeFeature:RuntimeHandler\] --minStartupPods=8
+      - --test_args=--ginkgo.skip=\[Slow\]|\[Serial\]|\[Disruptive\]|\[Driver:.gcepd\]|\[Flaky\]|\[Feature:.+\]|\[NodeFeature:RuntimeHandler\] --minStartupPods=8
```
endocrimes, line 24: It could be worth adding a comment here explaining why gcepd tests are disabled, or a TODO with an issue link.

leiyiz: there are multiple places where we skip gcepd tests, I think it might be better to just track it where we need it to be resolved? https://github.com/kubernetes/cloud-provider-gcp/issues/293

SergeyKanzhelev (approved):
/lgtm

PR #24446, opened 2021-11-23
Removes printing of snyk scan results

PushkarJ:
https://github.com/kubernetes/test-infra/pull/24443 added scan result printing which is not needed for debugging purposes. Confirmed this by running locally.

- Reverting that change but keeping the debug flag
- Added a comment on why `exit 1` is commented for image scanning

/sig security testing
/cc @nehaLohia27 (fyi)
/assign @spiffxp


spiffxp (approved):
/approve
/lgtm

PR #24457, opened 2021-11-24
Do not add do not merge label when PR is merged

### LINKED ISSUE #23205: release-note check the merged PR and label 'do-not-merge'
**What happened**:

When the body of an already merged PR is modified, release-note checks that pr does not contain release-note block, and then adds a `do-not-merge/release-note-label-needed`, which looks a little strangeness.

Can we ignore the check on the PR that has been merged?

Related PR: https://github.com/pingcap/tidb/pull/23982#issuecomment-896748369

jmguzik:
This commit fixes the issue described in #23205. Release note plugin
should not label the PR when it is already merged.

/fixes #23205

### DIFF
--- a/prow/plugins/releasenote/releasenote.go
+++ b/prow/plugins/releasenote/releasenote.go
@@ -256,6 +256,10 @@ func handlePR(gc githubClient, log *logrus.Entry, pr *github.PullRequestEvent) e
 	labelToAdd := determineReleaseNoteLabel(pr.PullRequest.Body, prLabels)
 
 	if labelToAdd == labels.ReleaseNoteLabelNeeded {
+		//Do not add do not merge label when the PR is merged
+		if pr.PullRequest.Merged {
+			return nil
+		}
 		if !prMustFollowRelNoteProcess(gc, log, pr, prLabels, true) {
 			ensureNoRelNoteNeededLabel(gc, log, pr, prLabels)
 			return clearStaleComments(gc, log, pr, prLabels, nil)
--- a/prow/plugins/releasenote/releasenote_test.go
+++ b/prow/plugins/releasenote/releasenote_test.go
@@ -270,6 +270,7 @@ func TestReleaseNotePR(t *testing.T) {
 		issueComments      []string
 		IssueLabelsAdded   []string
 		IssueLabelsRemoved []string
+		merged             bool
 	}{
 		{
 			name:          "LGTM with release-note",
@@ -456,12 +457,26 @@ func TestReleaseNotePR(t *testing.T) {
 			body:             "",
 			IssueLabelsAdded: []string{labels.ReleaseNoteLabelNeeded},
 		},
+		{
+			name:             "Add do-not-merge/release-note-label-needed",
+			body:             "```release-note\n```",
+			initialLabels:    []string{},
+			IssueLabelsAdded: []string{labels.ReleaseNoteLabelNeeded},
+		},
+		{
+			name:             "Release note edited after merge, do not add do-not-merge/release-note-label-needed",
+			body:             "```release-note\n```",
+			merged:           true,
+			initialLabels:    []string{},
+			IssueLabelsAdded: []string{},
+		},
 	}
 	for _, test := range tests {
 		if test.branch == "" {
 			test.branch = "master"
 		}
 		fc, pr := newFakeClient(test.body, test.branch, test.initialLabels, test.issueComments, test.parentPRs)
+		pr.PullRequest.Merged = test.merged
 
 		err := handlePR(fc, logrus.WithField("plugin", PluginName), pr)
 		if err != nil {


prow/plugins/releasenote/releasenote.go
```diff
@@ -275,7 +275,7 @@ func handlePR(gc githubClient, log *logrus.Entry, pr *github.PullRequestEvent) e
 			}
 			if containsNoneCommand(comments) {
 				labelToAdd = labels.ReleaseNoteNone
-			} else if !prLabels.Has(labels.ReleaseNoteLabelNeeded) {
+			} else if !prLabels.Has(labels.ReleaseNoteLabelNeeded) && !pr.PullRequest.Merged {
```
petr-muller, line 278: So this prevents adding a comment but not the label (which the issue is about). Maybe we should discard events on merged PRs sooner? Is there even any valid action when then the PR is merged?

petr-muller (commented):
/ok-to-test

prow/plugins/releasenote/releasenote.go
jmguzik: About the first one, I noticed it and was about to modify the change.
Regarding the actions, I will look into it. Not sure if possible.

jmguzik: Ok, hopefully, moved this condition to a better place.

prow/plugins/releasenote/releasenote.go
```diff
@@ -226,10 +226,10 @@ func shouldHandlePR(pr *github.PullRequestEvent) bool {
 		return false
 	}
 
-	// Ignoring unrelated PR labels prevents duplicate release note messages
+	// Ignoring unrelated PR labels prevents duplicate release note messages, merged PRs also ignored
 	if pr.Action == github.PullRequestActionLabeled {
 		for _, rnLabel := range allRNLabels {
-			if pr.Label.Name == rnLabel {
+			if pr.Label.Name == rnLabel && !pr.PullRequest.Merged {
```
petr-muller, line 232: So we still fire on `Edited` though? And if *that* is desired, why bother with iterating?

Not sure about the actual logic TBH. I think it is just never right to add the actual `do-not-merge/...` label (and probably an associated comment) on a merged PR, but that's it. So personally, I'd focus on specifically not adding  that label when the PR is merged.

I don't know the expectations for the other labels, so I'd keep their behavior as it is.

prow/plugins/releasenote/releasenote.go
```diff
@@ -219,6 +219,11 @@ func handlePullRequest(pc plugins.Agent, pr github.PullRequestEvent) error {
 }
 
 func shouldHandlePR(pr *github.PullRequestEvent) bool {
+	//Do not handle the PR when it's merged
```
jmguzik, line 222: Actually, I was certain this is the case, but after a second look round, not so sure. Let me think about this before further discussion

prow/plugins/releasenote/releasenote.go
jmguzik: I misunderstood initially your comment but I think, probably the condition should be like:

` if pr.Action == github.PullRequestActionLabeled && !pr.PullRequest.Merged `

We still fire on edited in my opinion.

jmguzik: implemented

prow/plugins/releasenote/releasenote.go
```diff
@@ -219,6 +219,11 @@ func handlePullRequest(pc plugins.Agent, pr github.PullRequestEvent) error {
...
+	if pr.PullRequest.Merged {
```
jmguzik, line 223: I think it's safe to say do not handle the PR since it's merged.  HandlePR function is all about adding and removing labels (also posting comments related to labels), so it's not needed for a merged PR.

prow/plugins/releasenote/releasenote.go
```diff
@@ -255,6 +255,11 @@ func handlePR(gc githubClient, log *logrus.Entry, pr *github.PullRequestEvent) e
 	var comments []github.IssueComment
 	labelToAdd := determineReleaseNoteLabel(pr.PullRequest.Body, prLabels)
 
+	//Do not handle ReleaseNoteLabelNeeded do not merge label when the PR is merged
+	if labelToAdd == labels.ReleaseNoteLabelNeeded && pr.PullRequest.Merged {
```
jmguzik, line 259: I think now it should be ok. Analyzed all do-not-merge flags in this relm and it seems so this one is the only that could occur inside the function. So, if flag and merged we should stop.

petr-muller: nit: the `labelToAdd == labels.ReleaseNoteLabelNeeded` part is the same like in the `if` below so maybe it would be simpler to add a 
```
if pr.PullRequest.Merged {
  return nil
}
``` 

there but /shrug

petr-muller (approved):

prow/plugins/releasenote/releasenote.go
jmguzik: @petr-muller you are right. I am an if-else master. Corrected now.

petr-muller (approved):

fejta (approved):

Issue #24394, opened 2021-11-17
Auto Assigning reviewers when a specific label is added via the label plugin


smg247:
**What would you like to be added**:

The optional ability for the label plugin to request reviews from a subset of the members of `approvers` when a specific label is added.

**Why is this needed**:

For our specific usecase: We need to be able to notify users (via email) that the PR is ready for *their* review when a specific label is added. By assigning them the regular github notifications will handle the rest. I believe that this functionality will be useful for others as well, as it allows for a generic, multi-stage review process.

**Configuration**:

I believe that the configuration for this can mimic that of `RestrictedLabels` and look as follows:

```
label:
  additional_labels:
  - some-label
  auto_assign_labels:
    '*':
      label: some-label
      assigned_teams:
        - some-team
      assigned_users:
      	 - a-user
```

smg247:
@alvaroaleman can you take a look here and let me know what you think?

alvaroaleman:
I don't understand the indirection over labels. If you have $something that adds the label (some plugin? A human?) $something can instead just assign the respective ppl?

petr-muller:
I think the original purpose was lost in various conversations ;) The original ask was to notify people who can provide certain label (= logins+teams) in "restricted_labels" on a certain action - when something happens on a PR, notify the "labellers" that they can label. Via assignment or review request.

```yaml
restricted_labels:                                                                                                                                                                                                                                                                      
  org/repo:
  - allowed_users:
    - user
    label: final
    assign_on:
    - label: label-that-signals-that-pr-is-eligible-for-final-label
```

This uses the fact that the label configuration knows who is allowed to set the restricted label, and would allow to assign them on certain conditions.

alvaroaleman:
Ah ok, the example above makes sense (maybe extend it with allowing multiple labels that are ANDed or ORed)

smg247:
@petr-muller its very possible I am misunderstanding the initial request, but note the part from (DPTP-2612)[https://issues.redhat.com/browse/DPTP-2612] that states

> is it possible to send mail to the approve list after the "backport-risk-assessed" label is added?
It was my understanding that we needed to assign reviewers after the label was added so that they could receive the notifications?

petr-muller:
@smg247 yep, I don't think you misunderstand. They already have a configuration that says:

```yaml
restricted_labels:                                                                                                                                                                                                                                                                      
  org/repo:
  - allowed_users:
    - qe-person-1
    - qe-person-2
[...]

PR #24458, opened 2021-11-24
Add ability to auto-assign users when specific label is added to label plugin

smg247:
We need a way to automatically assign users to a PR when a specific label is added. We can leverage the configuration of `restricted labels`  in the label plugin to accomplish this. This is the implementation for the enhancement described in [Issue 24394](https://github.com/kubernetes/test-infra/issues/24394).

I also updated the plugin help to include information on `restricted labels`.


smg247:
verify CLA

smg247:
I signed it

prow/plugins/label/label.go
```diff
@@ -218,6 +227,17 @@ func handle(gc githubClient, log *logrus.Entry, config plugins.Label, e *github.
 		if err := gc.AddLabel(org, repo, e.Number, labelToAdd); err != nil {
 			log.WithError(err).Errorf("GitHub failed to add the following label: %s", labelToAdd)
 		}
+
+		// Assign configured users if present
+		for _, restrictedLabel := range restrictedLabels {
+			for _, assignOn := range restrictedLabel.AssignOn {
```
alvaroaleman, line 233: This code assumes that the `assign_on` label always gets added by the `label` plugin which might or might not be true. IMHO it should instead react to the label adding event.

smg247: I'm not sure I follow what you mean by "label adding event", can you elaborate?

alvaroaleman: https://docs.github.com/en/developers/webhooks-and-events/events/github-event-types#issuesevent

smg247: The intention, as stated in https://github.com/kubernetes/test-infra/issues/24394, was to use the `label` plugin itself. I do, however, agree that it would be more robust if _any_ addition of a label results in this functionality. Would this mean creating a separate plugin to handle this though?

alvaroaleman: The intention as stated in that issue is to use the label plugin to do the assignment, it doesn't really say anything about where the event comes from. Requiring that the label plugin adds the label to get the trigger seems unintuitive and prone to be an issue sooner or later

smg247: I think I get what you are saying now. So rather than having this new functionality exist in the `labelsToAdd` loop we would just check for the existence of the label and assign the users outside of that loop.

smg247:
This needs a little more. I need to be able to also run the label plugin on `IssueActionLabeled`. Working on that now.

prow/plugins/config.go
```diff
@@ -411,9 +411,16 @@ func (l Label) IsRestrictedLabelInAdditionalLables(restricted string) bool {
...
-	Label        string   `json:"label"`
-	AllowedTeams []string `json:"allowed_teams,omitempty"`
-	AllowedUsers []string `json:"allowed_users,omitempty"`
+	Label        string          `json:"label"`
+	AllowedTeams []string        `json:"allowed_teams,omitempty"`
+	AllowedUsers []string        `json:"allowed_users,omitempty"`
+	AssignOn     []AssignOnLabel `json:"assign_on,omitempty"`
+}
+
+// AssignOnLabel specifies the label(s) that would trigger the RestrictedLabel.AllowedUsers'
```
alvaroaleman, line 420:
```suggestion
// AssignOnLabel specifies the label that would trigger the RestrictedLabel.AllowedUsers'
```

There can be multiple `AssignOnLabel` but each of them can only have one label

prow/plugins/label/label.go
```diff
@@ -297,3 +311,26 @@ func canUserSetLabel(ghc githubClient, org string, user string, label string, re
...
+	repo := e.Repo.Name
+	number := e.Issue.Number
+	restrictedLabels := config.RestrictedLabelsFor(e.Repo.Owner.Login, e.Repo.Name)
+
+	for _, restrictedLabel := range restrictedLabels {
+		for _, assignOn := range restrictedLabel.AssignOn {
+			if strings.EqualFold(e.Label.Name, assignOn.Label) {
+				// It's okay to re-assign users so no need to check if they are assigned
+				if err := gc.AssignIssue(org, repo, number, restrictedLabel.AllowedUsers); err != nil {
+					log.WithError(err).Errorf("GitHub failed to assign reviewers for the following label: %s", restrictedLabel.Label)
```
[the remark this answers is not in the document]
smg247: ah, I had just copied that pattern from elsewhere in the file. I will fix all occurrences.

PR #24460, opened 2021-11-24
kOps - Disable IRSA for k8s 1.21 pre-submit

hakman:
Ref: https://github.com/kubernetes/kops/pull/12829#issuecomment-978052189

/cc @rifelpet @johngmyers @olemarkus


johngmyers:
Why just the presubmit? Is there not something more systematic?

hakman:
> Why just the presubmit? Is there not something more systematic?

I think it's the only pre-submit for 1.21, so not really worth the effort.
Periodic tests are handled differently.

johngmyers:
Okay, that's the only presubmit for k8s < 1.22 and everything that's not a presubmit was taken care of before.

/lgtm

PR #24464, opened 2021-11-24
Move GPU tests out of blocking serial lane

ehashman:
Addresses https://github.com/kubernetes/kubernetes/issues/106635 enough to get it out of the 1.23 milestone.

/sig node


mmiranda96:
Would it make sense to merge this TestGrid and node-kubelet-flaky?

ehashman:
I don't think so; the goal of that job is to run all tests explicitly marked `[Flaky]` which these are not. Also, I don't think that job runs tests serially.

mmiranda96:
Ah, makes sense.
/lgtm

Issue #23829, opened 2021-09-29
Build and test Typescript without bazel


chaodaiG:
**What would you like to be added**:

Build and test Typescript without bazel, the related bazel rules are:

- nodejs_binary
- nodejs_test
- ts_library
- jasmine_node_test
- rollup_bundle
- genrule

Things to be considered:

- This will need to co-exist with current bazel rules for a short while, delete Typescript build/test targets from bazel once this confirmed working
- Test driven by make rule and bash scripts
- Ensure hermetic testing, including fixed tools version, like node, yarn, jasmine ([source of tools](https://github.com/bazelbuild/rules_nodejs/blob/3.1.0/package.json))
- Artifacts generated for serving prow UI
- Test results are in junit results format

**Why is this needed**:

This is part of #23796 

**Recommended testing approach**:
- Make test invokable by `make test`
- Create a PullRequest with change, manually run job by commenting `/test pull-test-infra-test-beta`
- [Optional]: if there is any tool that needs to exist in test image, update image at https://github.com/kubernetes/test-infra/blob/6e0d21ee50d690c8f78239257f588021d2ed57b5/config/jobs/kubernetes/test-infra/test-infra-presubmits.yaml#L145

/sig testing
/help

chaodaiG:
Did some quick search on github, this repo testing using yarn and jasmine might help: https://github.com/radanalyticsio/equoid-ui#client-tests

BenTheElder:
I started on this, there is an official docker image we can use.

One issue is that our node stuff is generally pretty out of date, e.g. we can get tslint working but it turns out it's actually deprecated ... 🙃 

I am starting with tslint, when we get the mechanics in place for that we should have a path to expand it to the other rules, and it aligns with the progress migrating the other linters https://github.com/kubernetes/test-infra/issues/23897

BenTheElder:
Also today we rely on the default version of npm, yarn etc. in bazel so it took a bit to track those down, and we don't specify them anywhere. We will probably need an analogue of .go-version but for the node docker image.

BenTheElder:
We wound up just putting a fixed node image version in a run-in-node-container.sh script in https://github.com/kubernetes/test-infra/pull/24491, which seems to be fine.

chaodaiG:
This was done, bazel not fully deprecated here as it's still building prow images as part of it. Will close this once prow images are not built from bazel

chaodaiG:
After #25373 prow images are built without bazel
/close

Issue #23796, opened 2021-09-28
Identifying build system for replacing bazel in test-infra


chaodaiG:
**What would you like to be added**:

Identifying another build system for replacing bazel in test-infra.

**Why is this needed**:

Test-infra repo is a Go centric repo, as of today 77.8% of the source code in this repo is Go, followed by 7.1% of python. Bazel as current build tool, however isn't working really great with Golang, and can sometimes delay the development (for example https://github.com/kubernetes/test-infra/pull/23693). This was due to the fact:

- rules_go and other rules don't follow/update go.mod
- Sometimes don't use the same libraries as in go.mod

For Go build and test we can use native Go toolchains, which should be sufficient for our use case.

**What will be missed (Problems to be solved)**

- Bazel caching in CI, namely RBE for speeding up incremental tests in CI
- Tests results in junit format
- Rules for python and typescript
- Prow images building rules
- Cluster ops rules

chaodaiG:
@fejta @spiffxp @alvaroaleman @stevekuznetsov @cjwagner for awareness

alvaroaleman:
Love the idea of getting rid of bazel :) Is the idea here to just replace bazels go build and test or is it to replace bazel entirely?

spiffxp:
FWIW, some of the @kubernetes/release-engineering folks have been experimenting with `mage` as an alternative to `make`: https://magefile.org/

chaodaiG:
> Is the idea here to just replace bazels go build and test or is it to replace bazel entirely?

I'm inclined to replace entirely to avoid maintaining two systems

alvaroaleman:
> I'm inclined to replace entirely to avoid maintaining two systems

Ah ok, then this would probably mean setting something up in parallel for now and gradually migrating components over?

We have a bunch of binaries that are just go, those should be relatively easy. Then there is stuff like Deck, which uses typescript and go, that will be a bit more difficult. I don't know the python components, so not sure how difficult or not that is going to be.

> Bazel caching in CI, namely RBE for speeding up incremental tests in CI

There was at some point a proposal for a cache that would be using object storage and be able to be used for the gocache. This never actually happened though. Do you know what k/k ended up doing, was the standard `go build/test` quick enough to not bother with the problem at all?

> Tests results in junit format

Other projects I've worked on use [gotestsum](https://github.com/gotestyourself/gotestsum) for this

[...]

Issue #24200, opened 2021-11-01
TestDiskCachePruning is flaky


chaodaiG:
Found it in https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/test-infra/24199/pull-test-infra-unit-test/1455313585742811136

local reproed by `go test -count=10 ./ghproxy -run=TestDiskCachePruning`

BenTheElder:
another https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/test-infra/24491/pull-test-infra-unit-test/1468352617385037824

chaodaiG:
I'll give it a stab
/assign

chaodaiG:
I think I have figured out what happened.

This line https://github.com/kubernetes/test-infra/blob/c2a2c9389e885065507750ebf16eaeebb53472a1/prow/github/app_auth_roundtripper.go#L84 was invoked twice, which generates a random string based on current timestamp, the value is rounded to second, so if the test happens around half of a second then it'll generated a different token, thus fail the test

PR #24491, opened 2021-11-29
add dockerized tslint

BenTheElder:
https://github.com/kubernetes/test-infra/issues/23796, https://github.com/kubernetes/test-infra/issues/23829

### DIFF
--- a/hack/make-rules/verify/tslint.sh
+++ b/hack/make-rules/verify/tslint.sh
@@ -1,0 +1,34 @@
+#!/usr/bin/env bash
+# Copyright 2021 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# this script runs tslint against the repo using docker
+#
+# TODO: we should be using eslint instead now, but this repo is also
+# on a rather old version of node, first we get the build system migrated.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+REPO_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/../../.." && pwd -P)"
+cd "${REPO_ROOT}"
+
+# ensure deps are installed
+# TODO: move out into make dependency step?
+hack/build/ensure-node_modules.sh
+
+# TODO: make sure this matches bazel version
+hack/run-in-node-container.sh \
+    node_modules/tslint/bin/tslint -c tslint.json --project .
--- a/hack/run-in-node-container.sh
+++ b/hack/run-in-node-container.sh
@@ -1,0 +1,41 @@
+#!/usr/bin/env bash
+# Copyright 2021 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This script runs $@ in a node container
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+REPO_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd -P)"
+cd "${REPO_ROOT}"
+
+# NOTE: bazel is currently on node 12.x, but we should probably upgrade at some point
+NODE_IMAGE='node:12-bullseye-slim'
+
+# NOTE: yarn tries to read configs under $HOME and fails if it can't,
+# we don't need these configs but we need it to not fail.
+# We set HOME to somewhere read/write-able by any user, since our uid will not
+# exist in /etc/passwd in the node image and yarn will try to read from / and
+# fail instead if we don't.
+# We are running as the current host user/group so the files produced are
+# owned appropriately on the host.
+docker run \
+    --rm -i \
+    --user $(id -u):$(id -g) \
+    -e HOME=/tmp \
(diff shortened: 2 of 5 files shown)


BenTheElder:
/hold
I forgot one thing 🙃

hack/run-in-node-container.sh
```diff
@@ -0,0 +1,33 @@
...
+
+REPO_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd -P)"
+cd "${REPO_ROOT}"
+
+NODE_IMAGE="$(cat "${REPO_ROOT}"/.node-image | tr -d '\n')"
+
+docker run \
+    --rm -it \
+    --user $(id -u):$(id -g) \
+    -e HOME=/tmp \
```
BenTheElder, line 30: yarn tries to read $HOME/.yarnrc and put a cache under $HOME/.cache, when we change the user in the line above (so output is correctly owned on the host) the user doesn't exist and yarn fails on inability to open `/.yarnrc`. Setting it to a valid writeable location allows yarn to function and is pretty harmless, yarn will be writing a system cache copy of package on install under `/tmp/.cache`.

The rest of this is hopefully more self-explanatory.

hack/make-rules/verify/tslint.sh
```diff
@@ -0,0 +1,35 @@
...
+set -o nounset
+set -o pipefail
+
+REPO_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/../../.." && pwd -P)"
+cd "${REPO_ROOT}"
+
+# ensure deps are installed
+# TODO: factor out into common script / step?
+hack/run-in-node-container.sh \
+    yarn install
```
BenTheElder, line 31: this takes maybe 2s (some variability), the whole `tslint.sh` takes ~6s total. not great, but hardly the slowest thing we're doing. 

probably not worth the complexity trying to sometimes skip this?
we already have the yarn lock and `node_modules/` cache

BenTheElder:
If I introduce a trivial error into one of the typescript files this definitely catches it, but I'm not sure yet that it has all the same options as under bazel.

.node-image
```diff
@@ -0,0 +1 @@
+node:12-bullseye-slim
```
BenTheElder, line 1: I'm not sure adding all these one-off version files is great. In this case it might be fine to just stick it in `run-in-node-container.sh` since we should only need node at build-time and we can do that all with that script AFAICT.

Originally I thought we might need to do some caching, but gitignored `node_modules` seems to handle that fine.

regarding the choice of image:
- relatively small, but complete
- we are on node 12.X under bazel currently

BenTheElder:
This is currently failing because `pull-test-infra-verify-lint` doesn't have docker running.

We should also consider investigating something like https://github.com/nvm-sh/nvm instead, similar to how we're now using gimmee for go, but containerizing node is very straightforward and avoids interacting with host state (other than the repo dir) which I know we can do fairly easily with go binaries on-the-host-but-not-system-installed but less sure with the node ecosystem.

hack/make-rules/verify/tslint.sh
chaodaiG: 2s should be fine, we do have many other stuffs that take much longer

hack/run-in-node-container.sh
chaodaiG: should this explanation be a comment above `--user` line?

.node-image
chaodaiG: I like the idea of sticking in in `run-in-node-container.sh`, compilation for `deck` can also be done in a way similar to `yarn install`, we should be fine

chaodaiG (commented):
Looks good to me! Also add a target after https://github.com/kubernetes/test-infra/blob/724022ee191844c1a8ebc0a1f6505ddb3694844e/Makefile#L85?

chaodaiG:
> We should also consider investigating something like https://github.com/nvm-sh/nvm instead, similar to how we're now using gimmee for go, but containerizing node is very straightforward and avoids interacting with host state (other than the repo dir) which I know we can do fairly easily with go binaries on-the-host-but-not-system-installed but less sure with the node ecosystem.

At this point, I think it makes more sense to go with docker approach to ensure hermetic build/test. We can iterate on nvm once there is a domain expert become interested in this

.node-image
BenTheElder: I'll do this.

hack/make-rules/verify/tslint.sh
BenTheElder: it will be something like this *per* typescript target though, unless we split it out and use a make dependency or something.

right now it will avoid re-downloading everything, but 2s is a bit slower than I'd like to do at the beginning of each typescript related script/target.

hack/run-in-node-container.sh
BenTheElder: I can do it, but we'll have to abuse a bash trick ;-)

technically normal comments don't work between line continuation (`\`)

chaodaiG: Add it above `docker run`?

hack/make-rules/verify/tslint.sh
chaodaiG: I think it depends on whether anyone would do this locally or not, maybe `make verify-tslint`, doubt anyone would need to compile deck locally very frequently.

For prow jobs, I think `lint` job does it once, `integration-test` compiles deck once, and `push-prow-image` also compiles deck once, each with 2s overhead.

so we might be fine?

BenTheElder: it's also something we can improve on later, factoring it out into a seperate script will make that relatively easy to make a make target dependency later

.node-image
BenTheElder: done

hack/run-in-node-container.sh
BenTheElder: I wound up doing that :-)

You can do this for future reference:
```bash
foo \
 `#bar comment` \
  baz
```

This works with backticks (which some shells considered deprecated vs `$()` but realistically cannot ever remove) and does not actually create a subshell. But when I discovered it trying to find a good way to comment a more complex docker invocation in POSIX shell I couldn't find many examples of it in the wild.

In bash the decent answer is to place the args in an array and comment the array instead.

```bash
docker_options = (
    # bar comment
    'foo'
    # quux comment
    'baz'
)
docker run "${docker_options[@]}"
```

chaodaiG:
enabling dind in https://github.com/kubernetes/test-infra/pull/24540, should unblock this PR

chaodaiG:
```
hack/build/ensure-node_modules.sh: line 26: hack/run-in-node-container.sh: No such file or directory
```
chmod?

BenTheElder:
no, but one too many `/..` in the `REPO_ROOT` snippet

chaodaiG (approved):

BenTheElder:
https://github.com/kubernetes/test-infra/issues/24200
/retest

PR #24499, opened 2021-11-30
[kubetest2] adding node-kubelet periodic jobs

Namanl2001:
This PR is part of the CI migration from kubetest to kubetest2
https://github.com/kubernetes/enhancements/tree/master/keps/sig-testing/2464-kubetest2-ci-migration

added:
`ci-kubernetes-node-kubelet-flaky-kubetest2` 
`ci-kubernetes-node-kubelet-orphans-kubetest2`
`ci-kubernetes-node-kubelet-serial-kubetest2`

cc: @dims @amwat @SergeyKanzhelev

### DIFF
--- a/config/jobs/kubernetes/sig-node/node-kubelet.yaml
+++ b/config/jobs/kubernetes/sig-node/node-kubelet.yaml
@@ -285,6 +285,46 @@ periodics:
   annotations:
     testgrid-dashboards: sig-node-kubelet
     testgrid-tab-name: node-kubelet-flaky
+- name: ci-kubernetes-node-kubelet-flaky-kubetest2
+  interval: 2h
+  decorate: true
+  extra_refs:
+  - org: kubernetes
+    repo: kubernetes
+    base_ref: master
+    path_alias: k8s.io/kubernetes
+  - org: kubernetes
+    repo: test-infra
+    base_ref: master
+    path_alias: k8s.io/test-infra
+  decoration_config:
+    timeout: 60m
+  annotations:
+    testgrid-dashboards: sig-node-kubelet
+    testgrid-tab-name: node-kubelet-flaky-kubetest2
+  labels:
+    preset-service-account: "true"
+    preset-k8s-ssh: "true"
+  spec:
+    containers:
+    - image: gcr.io/k8s-staging-test-infra/kubekins-e2e:v20211124-2ed05120f3-experimental
+      env:
+      - name: GOPATH
+        value: /go
+      command:
+      - runner.sh
+      args:
+      - kubetest2
+      - noop
+      - --test=node
+      - --
+      - --repo-root=.
+      - --gcp-project-type=node-e2e-project
+      - --gcp-zone=us-west1-b
+      - --parallelism=1
+      - --focus-regex=\[Flaky\]
+      - --test-args=--kubelet-flags="--cgroups-per-qos=true --cgroup-root=/" --server-start-timeout=420s
+      - --image-config-file=/home/prow/go/src/k8s.io/test-infra/jobs/e2e_node/image-config.yaml
 
 - name: ci-kubernetes-node-kubelet-orphans
   interval: 12h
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes/sig-node/node-kubelet.yaml
```diff
@@ -285,6 +285,46 @@ periodics:
...
+        value: /go
+      command:
+      - runner.sh
+      args:
+      - kubetest2
+      - noop
+      - --test=node
+      - --
+      - --repo-root=.
+      - --gcp-project-type=node-e2e-project
```
ehashman, line 322: Can we avoid adding new things to this project?

config/jobs/kubernetes/sig-node/node-kubelet.yaml
```diff
@@ -316,6 +356,46 @@ periodics:
...
```
ehashman, line 393: same comment as above

config/jobs/kubernetes/sig-node/node-kubelet.yaml
```diff
@@ -355,6 +435,53 @@ periodics:
...
+        value: /go
+      command:
+      - runner.sh
+      args:
+      - kubetest2
+      - noop
+      - --test=node
+      - --
+      - --repo-root=.
+      - --gcp-zone=us-west1-b
```
ehashman, line 479: Here we didn't using the same project but the job was using it... checking if that was deliberate.

Namanl2001: Nice catch 👍 I missed it

config/jobs/kubernetes/sig-node/node-kubelet.yaml
Namanl2001: I've added to this project as the non-kubetest2 job was using it. Do we have any other options?

amwat: FWIW just to clarify this is not a single project, but a pool of projects
https://github.com/kubernetes/test-infra/blob/5dd9ad002f35335e89b1cbc613b7f5695d181af9/config/prow/cluster/build/boskos-resources/boskos-resources.yaml#L329-L349 
and one project from it is leased through [boskos](https://github.com/kubernetes-sigs/boskos), at runtime during each job.

The name of the pool is intentionally designed to easily identify the types of jobs using that pool. So I would say `node-e2e-project` is the right choice for e2e node tests. 
If we see load being an issue due to the new jobs being added, we might want to consider adding more projects to the pool.

amwat (commented):
/lgtm

Namanl2001:
can we have this PR merged in, now? @amwat @ehashman

config/jobs/kubernetes/sig-node/node-kubelet.yaml
```diff
@@ -355,6 +435,54 @@ periodics:
...
+      - kubetest2
+      - noop
+      - --test=node
+      - --
+      - --repo-root=.
+      - --gcp-project-type=node-e2e-project
+      - --gcp-zone=us-west1-b
+      - --parallelism=1
+      - --focus-regex=\[Serial\]
+      - --skip-regex=\[Flaky\]|\[Benchmark\]|\[NodeSpecialFeature:.+\]|\[NodeSpecialFeature\]|\[NodeAlphaFeature:.+\]|\[NodeAlphaFeature\]|\[NodeFeature:Eviction\]|\[Feature:DynamicKubeletConfig\]|\[Feature:Docker\]|\[NodeFeature:DevicePluginProbe\]
```
SergeyKanzhelev, line 483: please preserve the comment from above job about DynamicKubeletConfig

SergeyKanzhelev (approved):
/hold

config/jobs/kubernetes/sig-node/node-kubelet.yaml
Namanl2001: addressed. Thanks for the review

SergeyKanzhelev: please squash or clean up commits after the change

Namanl2001: squashed all commits into one

ehashman:
/close

these tests run on dockershim; we probably should have caught this before, but there's no point in migrating them at this point

PR #24505, opened 2021-11-30
experiment: fix print-job-image-summary.sh

spiffxp:
Related:

It was double-counting certain image names like "bazel" that appear
across multiple registry prefixes. Address by displaying the full path
to the image (minus gcr.io) for images where we're not sure which org
the project (e.g. gcr.io/project-foo/image) belongs to


spiffxp:
/cc @ameukam @dims 

Sample output here: https://github.com/kubernetes/k8s.io/issues/1523#issuecomment-982974028

PR #24511, opened 2021-11-30
Sub prometheus metrics differentiate between user and prow errors

chaodaiG:
/cc @cjwagner @listx

### DIFF
--- a/prow/pubsub/subscriber/metrics.go
+++ b/prow/pubsub/subscriber/metrics.go
@@ -23,6 +23,9 @@ import (
 const (
 	responseCodeLabel = "response_code"
 	subscriptionLabel = "subscription"
+	// The value of "failed-handle-prowjob" is the only case where prow operator
+	// should care
+	errorTypeLabel = "error_type"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


prow/pubsub/subscriber/metrics.go
```diff
@@ -36,7 +39,7 @@ var (
 	errorCounter = prometheus.NewCounterVec(prometheus.CounterOpts{
 		Name: "prow_pubsub_error_counter",
 		Help: "A counter of the webhooks made to prow.",
-	}, []string{subscriptionLabel})
+	}, []string{subscriptionLabel, errorTypeLabel})
```
cjwagner, line 42: Do you know what will happen to the existing metric data that doesn't have this label?

prow/pubsub/subscriber/subscriber.go
```diff
@@ -327,7 +331,10 @@ func (s *Subscriber) handleMessage(msg messageInterface, subscription string, al
 	eType, err := extractFromAttribute(msg.getAttributes(), prowEventType)
 	if err != nil {
 		l.WithError(err).Error("failed to read message")
-		s.Metrics.ErrorCounter.With(prometheus.Labels{subscriptionLabel: subscription})
+		s.Metrics.ErrorCounter.With(prometheus.Labels{
```
cjwagner, line 334: I don't think this actually increments the counter (and didn't previously either). We need to append a `.Inc()` call to actually use the counter that we've just indexed out of the vector. This applies to the other places as well.

prow/pubsub/subscriber/metrics.go
chaodaiG: I think I've seen it before, it's default to empty string when this field doesn't exist

prow/pubsub/subscriber/subscriber.go
chaodaiG: good catch, fixed

cjwagner (approved):

PR #24513, opened 2021-12-01
Add a `tidequeryresults` metric for alerting on Tide query errors.

cjwagner:
/assign @listx

### DIFF
--- a/prow/tide/tide.go
+++ b/prow/tide/tide.go
@@ -148,10 +148,11 @@ type Pool struct {
 var (
 	tideMetrics = struct {
 		// Per pool
-		pooledPRs  *prometheus.GaugeVec
-		updateTime *prometheus.GaugeVec
-		merges     *prometheus.HistogramVec
-		poolErrors *prometheus.CounterVec
+		pooledPRs    *prometheus.GaugeVec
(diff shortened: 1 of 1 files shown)


prow/tide/tide.go
```diff
@@ -457,9 +468,15 @@ func (c *Controller) query() (map[string]PullRequest, error) {
 			go func() {
 				defer wg.Done()
 				results, err := search(c.ghc.QueryWithGitHubAppsSupport, c.logger, q, time.Time{}, time.Now(), org)
+
+				resultString := "success"
+				if err != nil {
+					resultString = "error"
+				}
+				tideMetrics.queryResults.WithLabelValues(strconv.Itoa(i), org, resultString).Inc()
```
listx, line 476: Is the use of `i` here guilty of the same error described in this article https://dev.to/kkentzo/the-golang-for-loop-gotcha-1n35 ?

petr-muller:
/cc @jmguzik 
FYI

prow/tide/tide.go
cjwagner: Nice catch, there was an existing problem with this 5 lines below as well. Should be fixed now.

PR #24532, opened 2021-12-03
Optimize APIv4 rate limit usage in blunderbuss plugin

jmguzik:
This commit fixes random API v4 rate-limit per minute violation when
config option useStatusAvailability is set. In such cases, when a large
set of reviewers is found, there are situations that the client is sending
multiple queries in several seconds, blocking the possibility to query
the data and returning errors for all the left users (code 403). The solution
is to compose a dynamic graphql query that will ask for all the possible reviewers
in two calls (or more when change is huge), thus not firing one call per user
as it is currently implemented.

While dynamic queries are supported by graphql, they are a little tricky in qgrapql library that prow is using to compose a query. The possible solution using reflect is described in the issue https://github.com/shurcooL/githubv4/issues/17 though it is later noticed such queries are only good for asking for data and possibly not overcomplicated issues. 

The cost of asking for multiple reviewers via dynamic query would potentially be higher (although dry-run shows it's still 1), but one-minute API violation will go away.


jmguzik:
This is probably the first time graphql dynamic query is used in prow. I will leave some review hints.

prow/plugins/blunderbuss/blunderbuss.go
```diff
@@ -334,34 +360,74 @@ func findReviewer(ghc githubClient, log *logrus.Entry, useStatusAvailability boo
...
+		githubUserAvailabilityStructLimit = 300
+	)
+	currentIndex := 0
+	numberOfBatches := int(math.Ceil(float64(len(reviewers)) / githubUserAvailabilityStructLimit))
+	for batch := 0; batch < numberOfBatches; batch++ {
+		var fields []reflect.StructField
+		batchError := false
+		structsAdded := 0
+		for i := currentIndex; i < len(reviewers) && structsAdded < githubUserAvailabilityStructLimit; i++ {
+			fields = append(fields, reflect.StructField{
```
jmguzik, line 409: Instead of predefining a whole query as a struct,` reflect.StructField` is used to compose a dynamic query with the requested number of users. It is parsed by `qhc.query` function on almost the same basis as the predefined struct is parsed.

prow/plugins/blunderbuss/blunderbuss.go
```diff
@@ -334,34 +360,74 @@ func findReviewer(ghc githubClient, log *logrus.Entry, useStatusAvailability boo
...
+// graphqlTag helps in formating the graphql tag
+func graphqlTag(value string) reflect.StructTag {
+	return reflect.StructTag(strconv.AppendQuote([]byte("graphql:"), value))
+}
+
+// findBusyReviewers finds all the people that are busy and excluded from the review
+func findBusyReviewers(ghc githubClient, log *logrus.Entry, busyReviewers *sets.String, reviewers []string) {
+	const (
+		// will panic if too many fields with too much data in the reflect struct, this setting is very safe
+		githubUserAvailabilityStructLimit = 300
```
jmguzik, line 400: When huge reviewers set is queried, users are divided into batches. Batches have to have limits due to too big structure error in golang or too much data problem. This limit is safe. Allows asking for users where every user has more than 100 chars in its login.
This is introduced to prevent panic in heavy PRs:
- PR: https://github.com/openshift/cluster-authentication-operator/pull/517 asked for more then 1600 reviewers. In the current situation, it got answered around 30 times and the rest was unchecked due to rate limit error
- Another similar example is PR: https://github.com/openshift/operator-framework-olm/pull/219 more than 1400 single queries about the user.
Without division to batches, reflect would panic due to too large structure.

prow/plugins/blunderbuss/blunderbuss.go
```diff
@@ -334,34 +360,74 @@ func findReviewer(ghc githubClient, log *logrus.Entry, useStatusAvailability boo
...
+			fields = append(fields, reflect.StructField{
+				Tag:  graphqlTag(fmt.Sprintf("user%d :user(login: %q)", i, reviewers[i])),
+				Name: fmt.Sprintf("User%d", i),
+				Type: reflect.TypeOf(githubUserAvailability{}),
+			})
+			structsAdded++
+		}
+		query := reflect.New(reflect.StructOf(fields)).Elem()
+		err := ghc.Query(context.Background(), query.Addr().Interface(), nil)
+		if err != nil {
```
jmguzik, line 418: Previously, in case of a single error, we counted such reviewers as busy for the review. As the errors repeat over a limit, multiple reviewers were counted as busy ones. The same behavior is more or less recreated here. If error, the whole batch goes to the busy set. Let me know if that is ok.

petr-muller:
We had a conversation about this over Slack which will continue, but I will add some of my thoughts here:

1. My assumption is that free/busy state is not something that is too widely used
2. This PR adds quite a bit of complexity (mostly mandated by the limitation of the graphql library we use in how it does not directly support dynamic queries) in an effort to allow querying GitHub for free/busy state of many users at once.
3. I think the underlying problem is that blunderbuss somehow ends up doing these queries for too many people - the count of review requests is usually very low (it is configured to 2 in our Prow), so my expectation about the number of necessary calls would be: # of configured reviewers + # of hits where a configured reviewer is busy at the moment. 
4. For some reason, blunderbuss does not behave that way: like Jakub says in https://github.com/kubernetes/test-infra/pull/24532#discussion_r762847107, we see blunderbuss emit hundreds of these calls per PR.

So I think I need more evidence that optimizing the size of graphql query is really the right thing to do here. I think the added complexity is really only needed if we absolutely must do these large queries, but I think that blunderbuss should never really do large queries like that (no matter if in hundreds of small query request or in one big one). I think that blunderbuss should really only need this information for the small set of logins (the potential candidates for review requests).

One observation about the huge amount of calls in PRs in https://github.com/kubernetes/test-infra/pull/24532#discussion_r762847107 is that they both change hundreds of files. Maybe something in the code that handles per-file reviews does not scale that well wr.t. API usage? Not sure.

jmguzik:
PR https://github.com/kubernetes/test-infra/pull/24568 has been opened to determine how much of this implementation can be preserved/reused in the light of described events.

prow/plugins/blunderbuss/blunderbuss.go
jmguzik: This is the library's author's recommendation. Please refer to https://github.com/shurcooL/githubv4/issues/17.
Also, dynamic query example is pasted below:
```
{
  user1: user(login: "first-user") {
    id
    status {
      indicatesLimitedAvailability
    }
  }
  user2: user(login: "user") {
    id
    status {
      indicatesLimitedAvailability
    }
  }
  user3: user(login: "another-user") {
    id
    status {
      indicatesLimitedAvailability
    }
  }
}
```

prow/plugins/blunderbuss/blunderbuss.go
```diff
@@ -269,12 +271,20 @@ func getReviewers(rc reviewersClient, ghc githubClient, log *logrus.Entry, autho
...
+	fileReviewers := []layeredsets.String{}
 	busyReviewers := sets.NewString()
 	ownersSeen := sets.NewString()
 	if minReviewers == 0 {
 		return reviewers.List(), requiredReviewers.List(), nil
 	}
-	// first build 'reviewers' by taking a unique reviewer from each OWNERS file.
+
+	// find statuses of multiple reviewers in one query to respect API rate limits
+	if useStatusAvailability {
```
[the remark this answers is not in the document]
jmguzik: We are currently investigating why there are so many qraphql calls (maybe the error is actually in the logic) and how much busy states play a big role in it. Logging will be added to see more insights: https://github.com/kubernetes/test-infra/pull/24568

jmguzik:
I'll close this for now. The error is still there, however, we blocked this feature for now in our deployment.

PR #24533, opened 2021-12-03
Increase GKE podmonitor scrape timeout to 30s

chaodaiG:
Observed missing data points in metrics scraped by GKE workload metrics. Not sure what the default is, explicitly set it to make it better predictable.

The field name scrapeTimeout was grabbed by:
kubectl get crd podmonitors.monitoring.gke.io -oyaml


cjwagner (approved):
Thanks. This seems likely to be the problem. `prowjob_state_transitions` has high cardinality.

PR #24544, opened 2021-12-04
Remove kops tests for unsupported distros

johngmyers:
/cc @hakman


hakman:
@johngmyers Seems that we want to keep them around a little longer:
https://github.com/kubernetes/test-infra/pull/24542#discussion_r762378567

PR #24585, opened 2021-12-08
Display testsuite name if defined in junit spyglass lens.

dgoodwin:
We found cases in openshift where the same test was run in different suites, but we could not easily tell which we were seeing in the prow output.

This change will now display a prefix of "suitename." if the suite name is defined, otherwise nothing.

Also removed the leading : when a junit classname is not in use. (which we also hit in OpenShift)

### DIFF
--- a/prow/spyglass/lenses/junit/lens.go
+++ b/prow/spyglass/lenses/junit/lens.go
@@ -89,6 +89,7 @@ func (lens Lens) Callback(artifacts []api.Artifact, resourceDir string, data str
 
 type JunitResult struct {
 	junit.Result
+	SuiteName string
 }
 
 func (jr JunitResult) Duration() time.Duration {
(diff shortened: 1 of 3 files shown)


cjwagner:
/retest
This looks generally fine to me, but I wonder if existing users might already have suite names which will now just be repeated clutter, especially if they are verbose. I don't think we have a good way of evaluating that impact unfortunately.

Class names are usually used for namespacing tests cases. If OpenShift is currently omitting those could you just switch to supplying those rather than depending on the suite name for namespacing?

dgoodwin:
In our case we got into trouble with upgrade jobs where there were multiple test suite invocations (pre/post testing) resulting in multiple junit xml files. At the time we weren't using suite names, and doing so was our attempt to start differentiating between tests that happened to pop up in both suites, but we don't want treated as flakes. (as one was a hard fail) This worked in terms of testgrid and such which displays suite names in similar fashion to this PR, but then we found we couldn't tell which suite the test was failing in in prow easily.

Indeed someone might start seeing more info than they prefer with this change, though the use of suite names is optional if they had control over their junit xml generation. *Hopefully* this wouldn't break anything in particular, just additional visual info as far as I understand. (but I am new to this repo)

Suites seems to be the correct fit in our case, but if this change is not desirable then I will try to hack something up for our needs that uses different class names in each junit xml.

stevekuznetsov:
@dgoodwin adding a config option for this to `deck` and letting users opt into it should make this good to merge

Issue #24024, opened 2021-10-15
Prow error log: exceeded a secondary rate limit with GitHub


chaodaiG:
**What happened**:

Saw error log in prow.k8s.io:

```
jsonPayload: {
component: "tide"
error: "failed to query GitHub for prs: query 4, err: non-200 OK status code: 403 Forbidden body: "{\n  \"documentation_url\": \"https://docs.github.com/en/free-pro-team@latest/rest/overview/resources-in-the-rest-api#secondary-rate-limits\",\n  \"message\": \"You have exceeded a secondary rate limit. Please wait a few minutes before you try again.\"\n}\n""
file: "prow/cmd/tide/main.go:215"
func: "main.sync"
level: "error"
msg: "Error syncing."
}
```

This is happening on both tide and needs-rebase

listx:
What are the next steps for this bug (to prevent it from happening again in the future)?

listx:
FWIW this is still happening... for the `needs-rebase` component.

```
{
  "insertId": "hb60zh5t7agwi45o",
  "jsonPayload": {
    "error": "non-200 OK status code: 403 Forbidden body: \"{\\n  \\\"documentation_url\\\": \\\"https://docs.github.com/en/free-pro-team@latest/rest/overview/resources-in-the-rest-api#secondary-rate-limits\\\",\\n  \\\"message\\\": \\\"You have exceeded a secondary rate limit. Please wait a few minutes before you try again.\\\"\\n}\\n\"",
    "msg": "Encountered errors when querying GitHub but will process received results anyways",
    "level": "error",
    "func": "k8s.io/test-infra/prow/external-plugins/needs-rebase/plugin.HandleAll",
    "plugin": "needs-rebase",
    "component": "needs-rebase",
    "file": "prow/external-plugins/needs-rebase/plugin/plugin.go:157"
  },
  "resource": {
    "type": "k8s_container",
    "labels": {
      "location": "us-central1-f",
      "container_name": "needs-rebase",
      "project_id": "k8s-prow",
      "pod_name": "needs-rebase-5497778fbd-phzw9",
      "cluster_name": "prow",
      "namespace_name": "default"
    }
  },
  "timestamp": "2021-11-12T16:42:39Z",
  "severity": "ERROR",
  "labels": {
    "k8s-pod/pod-template-hash": "5497778fbd",
    "k8s-pod/app": "needs-rebase",
[...]

PR #24594, opened 2021-12-08
Lower possibility of hitting secondary API limit

jmguzik:
This commit changes the way graphql apiv4 calls are treated in tide.
Concurrent model is changed to sequential with an additional rate limit.


cjwagner:
Hi Jakub, could you explain the rate limit that is being encountered and under what conditions? I haven't seen any evidence of rate limiting problems with the Tide instances we manage and some are quite large.

I'm a little concerned that this could increase query time too much. If this is only necessary under certain conditions we may want to make it an opt-in feature to prevent slowing down Tide instances that don't need it.

jmguzik:
@cjwagner The change in tide is backed mostly by observation we have in our tide in Openshift org, Github documentation, and a similar change in another component.
Let me start with the documentation:  https://docs.github.com/en/rest/guides/best-practices-for-integrators#dealing-with-secondary-rate-limits This piece of the document states that we really should be careful with `POST` and we should even sleep between the queries. It is not explicitly said in the documentation that it is exactly the same situation as ours, but it's observable. The rules seem to apply to the situation, and GitHub is not sharing any more data besides the document.

The second thing is the statement that was made in `needs-rebase` component. Please see the line https://github.com/kubernetes/test-infra/blob/master/prow/external-plugins/needs-rebase/plugin/plugin.go#L145 we have a very similar situation, yet it does not prevent by hitting the secondary API abuse as it is seen in this issue: https://github.com/kubernetes/test-infra/issues/24024 (please see the comment about `needs-rebase`). The change is from April, and the error is from October.

Finally the evidence. Please find attached the log message that happens in our installation quite often:
```
Field | Value
-- | --
@ingestionTime | 1639008843822
@log | 320297955214:app-ci-pod-logs
@logStream | ip-10-0-143-5.ec2.internal
....
@timestamp | 1639008843136
component | tide
controller | sync
error | non-200 OK status code: 403 Forbidden body: "{\n \"documentation_url\": \"https://docs.github.com/en/free-pro-team@latest/rest/overview/resources-in-the-rest-api#secondary-rate-limits\",\n \"message\": \"You have exceeded a secondary rate limit. Please wait a few minutes before you try again.\"\n}\n"
file | prow/tide/tide.go:481
func | k8s.io/test-infra/prow/tide.(*Controller).query.func1
....
```
We have around 800 error messages from tide in the last 3 days and roughly 90% is from `prow/tide/tide.go:481` . Please see also the screenshot:
tide

What I believe we should do is to get rid of concurrency in this code. Ideally, a ticker should be added as you suggested as a config param (for now I got rid of it in my PR). It should help us find the right parameter not to hit these errors anymore. I also added right now a ticker to `needs-rebase` as it still causes problems even without the parallelization (https://github.com/kubernetes/test-infra/issues/24024), but that also can be a config param.

I hope this helps a little bit to see my rationale behind the change. Let me know what you think.

jmguzik:
I will leave like it is for the review, but to be honest I would like the idea of having a configurable ticker between query batches. I feel then the rules of not triggering the secondary API abuse would be fulfilled.

cjwagner:
Copying my response on slack here:
> I don't think we've had to do anything special with our tide config to mitigate this. https://github.com/kubernetes/test-infra/blob/master/config/prow/config.yaml#L518
I think I'd still like to see this as opt-in and configurable so that we don't negatively affect existing instances that are not encountering this problem.
What about making this only delay after receiving the rate limit abuse warning?
I wonder if you have something else using the v4 API with the same credentials which is increasing the concurrency conflicts.

jmguzik:
If we forget about a ticker right now, what do you think about getting rid of the concurrency? Would it be acceptable for you, or do you want to have it configurable also? I think the evidence against concurrency in this place is strong.

If you agree, let's merge this one and I will return to the topic after having new observations from our installation.

jmguzik:
We agreed on slack that the mechanism should be optional. I will try to rework my solution to allow that. 
Until that, solution is on hold.
/hold

jmguzik:
I reworked the code to add max_graphql_gororotines setting. If set all graphql executions in tide will be limited to x given goroutines at once. 
For the time being, I resign from adding timers, tickers, let's see how this one helps. I tested it also in my local development installation and it seems to be working. Let me know what you think.
/cc @cjwagner

petr-muller (approved):

prow/config/config.go
```diff
@@ -2234,6 +2234,10 @@ func parseProwConfig(c *Config) error {
 		return fmt.Errorf("tide has invalid max_goroutines (%d), it needs to be a positive number", c.Tide.MaxGoroutines)
 	}
 
+	if c.Tide.MaxGraphQLGoroutines < 0 {
```
stevekuznetsov, line 2237: You could have the type be an unsigned integer, and the YAML marshaller will error this for you

prow/tide/blockers/blockers.go
```diff
@@ -113,11 +118,14 @@ func FindAll(ghc githubClient, log *logrus.Entry, label string, orgRepoTokensByO
 				return
 			}
 			issues = append(issues, result...)
-
+			if useQueryPool {
```
stevekuznetsov, line 121: I don't think this does what you want it to do. Writing here will block the routine *after* it has touched the GH API, and without a reader you will never hit the `wg.Wait()` call. I suggest that you:

 - use a channel to put all the queries into
 - spawn N goroutines to read from the channel and process until it's done
 
```go
type query struct {
    org string
    query sets.String
}
toQuery := make(chan query, len(queries))
for org, querySet := range queries {
    toQuery <- query{org: org, query: querySet}
}

graphQLRoutines := maxGraphQLGoroutines
if maxGraphQLGoroutines == 0 {
    graphQLRoutines = len(queries)
}
var wg sync.WaitGroup
for i := 0; i < maxGraphQLGoroutines; i++ {
    wg.Add(1)
    go func() {
        defer wg.Done()
        for query := range toQuery {
        // process it
        }
    }
}

wg.Wait()
close(toQuery)
```

alvaroaleman:
FWIW, this is a very simplistic approach. I am 95% sure that GitHub does ratelimit per token and not per account, which means in the context of apps auth which openshift is using, you get one 1 qps per github organization (as all of them have different tokens). 

I am also very sure that if you end up setting Tide query parallelization to 1 globally, you are going to completely kill the performance of openshifts Tide. The parallelization was originally introduces due to performance issues due to the very high number of queries there: https://github.com/kubernetes/test-infra/pull/20550

Also, 800 errors over three days doesn't seem a lot to me:
* Openshift currently has 83 unique tide queries
* That get executed every two minutes
* Which makes for a total of 179280 queries over three days (30 * 24 * 3 * 83)
* Which makes 800 errors an error rate of about 0.44%

Do you see problematic behavior from Tide or is this change purely motivated by seeing errors in the log?

jmguzik:
It was more visible when tide was executed every 1m (we changed this to 2m), but the last time I checked it was indeed 800. It could be less it could be more, it varies.
We wanted to get rid of any API problems in general, small ones, big ones. It can grow over time. Probably the setting of 1 is not right after your remarks, but some higher could serve the purpose of limiting errors and not limiting the throughput, especially that we see sync of 1m->2m does not affect tide so much. 
Is you comment about not doing this change at all, or not setting it to 1? If it's the latter I would introduce the changes and continue with this.

alvaroaleman:
I am not convinced that this actually improves anything other than seeing less errors in the logs, which you can achieve much simpler by changing the loglevel if you encounter this particular message.

I'm not a fan of adding yet another setting without even a clear issue that is being solved. The proper solution seems to be to have a ratelimiter clientside that limits to 1 qps (in the case of apps auth scoped to org). In the absence of that, ignoring the fact that we are getting one error every two hundred requests because we are going to do that request again two minutes later seems also fine.

petr-muller:
The motivation for this is that we *do* have (many) problems with getting rate limited by GitHub: not just in Tide, but also in other components. Unfortunately we are now in the situation where parts of our Prow are using different auth methods. But we hit problems with both GH App authed components and those that still use PATs. So our goal was to start chipping away misbehaving (according to GH API best practices linked somewhere above) bits of Prow to reduce the noise that we are seeing.

Note that OpenShift Tide is still using the PAT (not app - we needed to revert that change back in August) so it seems that at least part of this PR may not be entirely relevant as they would not be parallel. 

Also note that the rate limiting (when you hit it) is not just a matter of waiting two minutes and retrying, because other parts of Prow are stuck (resulting in delays etc) while GH decides to rate limit us.

stevekuznetsov:
I think it's simple enough to add a configurable concurrency bound, we should ideally have such an option in all cases where a core component fans out so that users can tune their behavior. Whether or not that will improve this particular use-case or rate-limit is not clear to me but also not entirely important.

prow/tide/blockers/blockers.go
jmguzik: I am not an expert but it seems to be working at least for me. I engineered this method based on the article: https://go.dev/blog/pipelines
```
func main() {
        max := 3
	blocker := make(chan int, max)
	var wg sync.WaitGroup
	for i := 0; i < 20; i++ {
		blocker <- 1
		wg.Add(1)
		go func(i int) {
			defer wg.Done()
			fmt.Println("hello world ", i)
			time.Sleep(time.Second)
			<-blocker
		}(i)
	}
	wg.Wait()
	close(blocker)
}
```
This snippet which is a simplification also works fine. I have ran test version of tide on minikube to see any locks and at least for this small test there were none. Again, maybe I just don't see it (or you meant some specific case), but I think it will work.

stevekuznetsov: Sorry, I misread the channel manipulation. This is "clever" code and not IMHO a common Go-ism. I would prefer that we not use these patterns from the blog in 2014 - Go's concurrency has changed a lot over the years and that blog is old enough to entirely predate `context.Context` for managing cancellations.

Please queue data to be processed into a channel and use a bounded set of consumers to ensure concurrency <= N.

jmguzik: Ok, just wanted to ask because I did not understand what could possibly go wrong here.

jmguzik:
I found a moment today to do the suggested changes. I hope now it's better.

Let me also comment on the discussion. While I fully understand the need for performance upgrades, this change is not about the performance. It's more about correctness. In my comments before the deeper discussion, I have never pointed out the performance nature of the problem. I was rather focused on following the GitHub documentation.
It is not quite correct for a client to abuse the secondary API limit (repeatedly). Yes, I have executed the test and fortunately, the APIv4 abuse does not trigger APIv3 abuse by the way, but I still think users (so also us) should have the possibility to tune the system in a way, to not trigger such problems. I don't quite understand why adding another parameter that clearly tackles the problem: let's do an unlimited number of goroutines for each of our queries, and inside every goroutine let's fire up even more queries sequentially, is not needed because only a small % of errors go away. Other components also have these switches for concurrency. What is not ok from my perspective is not controlling this process.
In our installation tide has separate limits so it's probably (not fully confirmed, still looking into logs, metrics) it's not blocking much of the traffic in other deployments, but that is not a definitive argument as our goal here would be to stop being abusive.
Originally, I wanted to add also a ticker to limit sequential query paging inside a goroutine as it is observable in needs-rebase that even a sequential query (with paging) can trigger the abuse mechanism, but I think it's better to do these changes in steps and focus on observability first. I am not fully convinced the problem will go away entirely, but I still think it's not a bad practice from the client's perspective to have at least some level of control over the process of spawning routines (and errors).

alvaroaleman:
>  It's more about correctness. 

Yeah but the point is this change doesn't result in more correct behavior, it results in a switch that you can use to maybe achieve correct behavior. If it takes us half a second to do a query we are again behaving incorrectly, regardless of the value of this switch, as we end up doing more than 1 qps.

The proper solution for the issue is to have a throttler somewhere that limits us to 1 qps. That is obviously more work than the change here, which is why I said before its fine to not do that and just live with the incorrect usage as it isn't yet an actual problem.

I am not a huge fan though of introducing a new switch that attempts to solve a problem but can only approximate solving it, as a stopgap to the real solution.

jmguzik:
I have seen these limits, but your argument was that I will worsen tide performance by getting rid of concurrency entirely.
Now, I have to keep concurrency but have a limiter which slows everything down to 1 qps which is even slower than the serialization itself in that piece of code. How that won't cause problems in tide?
The point actually is, I think lots of things are unrevealed in the GH documentation though. I have seen that 1 qps is not that strict in my test env.
Still think that spawning unlimited number of routines is not a good idea. In other places this is limited.

prow/tide/status.go
```diff
@@ -610,34 +610,36 @@ func (sc *statusController) search() []PullRequest {
 	var prs []PullRequest
 	var errs []error
 	var lock sync.Mutex
-	var wg sync.WaitGroup
 
-	for org, query := range queries {
-		org, query := org, query
-		wg.Add(1)
+	toQuery := make([]queryData, 0)
```
stevekuznetsov, line 614: nit: the Go idiom is

```go
var toQuery []queryData
```

prow/tide/status_test.go
```diff
@@ -1303,7 +1317,13 @@ func TestStatusControllerSearch(t *testing.T) {
 		t.Run(tc.name, func(t *testing.T) {
 			ghc := &fgc{prs: tc.prs}
 			cfg := func() *config.Config {
-				return &config.Config{ProwConfig: config.ProwConfig{Tide: config.Tide{Queries: config.TideQueries{{Orgs: []string{"org-a", "org-b"}}}}}}
```
stevekuznetsov, line 1306: What's this diff about?

prow/tide/tide_test.go
```diff
@@ -4288,6 +4289,16 @@ func TestQueryShardsByOrgWhenAppsAuthIsEnabledOnly(t *testing.T) {
 			},
 			expectedNumberOfApiCalls: 2,
 		},
+		{
+			name:               "Apps auth is used, one call per org, goroutines limited",
```
stevekuznetsov, line 4293: In the PR comments, Petr noted that you were not using Apps auth - likely you want to add such a test to the matrix?

prow/tide/tide.go
```diff
@@ -534,6 +533,37 @@ func subpoolsInParallel(goroutines int, sps map[string]*subpool, process func(*s
 	wg.Wait()
 }
 
+type queryData struct {
+	org         string
+	query       string
+	metricIndex int
```
stevekuznetsov, line 539: If you added a metric for the search query (or made the handling optional) you would not need to repeat yourself in this file.

prow/tide/tide.go
```diff
@@ -461,38 +461,37 @@ func (c *Controller) query() (map[string]PullRequest, error) {
...
+				c.logger.WithField("query", q).WithError(err).Warn("Failed to execute query.")
+				errs = append(errs, fmt.Errorf("query %d, err: %w", q.metricIndex, err))
+				return
+			}
+			if err != nil {
+				c.logger.WithError(err).WithField("query", q).Warning("found partial results")
+			}
+
+			for _, pr := range results {
+				prs[prKey(&pr)] = pr
```
stevekuznetsov, line 492: I know this predates the current PR, but why is this not racy?

stevekuznetsov:
Ultimately, we do not know specifically what it is that triggers the GitHub secondary limits (by design). So, we must always implement some proxy for the "correct" behavior and hope that we come close to doing the right thing. Prow provides concurrency throttles in most other places I can think of, so I don't see a great reason not to do one here. 

The GitHub [documentation](https://docs.github.com/en/rest/guides/best-practices-for-integrators#dealing-with-secondary-rate-limits) on this does specifically call out making serial requests, as well as limiting to 1QPS, so if you do continue having issues @jmguzik after attempting to set low concurrency, a throughput-limiting round-tripper in GHProxy may be a good next step. I believe there's already prior art there for limiting the size of in-flight requests, but a sliding-window throughput limit would be useful.

alvaroaleman:
The reason I am not a big fan of this is that the concurrency does not cleanly map to a qps limit in a couple of ways:
* How long requests take to complete varies
* A query might actually result in multiple api calls when there is more than one page of results
* In the context of github apps auth, the secondary limit is likely per org, which this mechanism has no way to account for


So to do it properly, we will sooner or late get a throttler for the qps. Once we have that, we have suddenly have two knobs which will be super confusing to understand.

alvaroaleman:
This change has a cost in terms of code complexity and more importantly configuration complexity while frankly not being a good solution to the underlying problem.

stevekuznetsov:
I think it's fine if this change does not fix the QPS problem. It linearly relates to the compute used, which may in and of itself be a useful knob (and one that we support in pretty much every single other place where infinite fan-out could happen).

prow/tide/tide_test.go
jmguzik: For the concurrency bound it's indifferent. Actually, we won't benefit from changes in ``blockers.go`` and ``status.go`` as they will most likely be collapsed to 1 query. We will benefit from limiting concurrency in ``tide.go`` as every outer loop will have it's own iteration. As I wrote before, for the paged queries it may still be not enough though. Back to this change, if it's indifferent and I am just testing the mechanism, I just want to have more than 1 calls to API concurrently and it was a quicker way to invoke this in test IMHO.

prow/tide/tide.go
jmguzik: Actually this metric was added quite recently by google guys. They wanted to have a metric based on former outer loop index. So, it just about perservance of the things they wrote. Previously every goroutine started with this: ``org, q, i := org, q, i`` in the inner loop.

prow/tide/status_test.go
jmguzik: I added ``tc.maxGraphQLGoroutines`` and reformatted the code to have more readability. I find one liner to be quite difficult to read and edit.

prow/tide/tide.go
jmguzik: Probably the lock in 480 protects it? Maybe it's another Go thing, but I do not see the danger...

jmguzik:
> The GitHub [documentation](https://docs.github.com/en/rest/guides/best-practices-for-integrators#dealing-with-secondary-rate-limits) on this does specifically call out making serial requests, as well as limiting to 1QPS, so if you do continue having issues @jmguzik after attempting to set low concurrency, a throughput-limiting round-tripper in GHProxy may be a good next step. I believe there's already prior art there for limiting the size of in-flight requests, but a sliding-window throughput limit would be useful.

Yeah, I think we are slowly going to that direction with both APIs.

prow/tide/tide.go
stevekuznetsov: Right - I'm pointing out that you add two sets of code here which are identical but for the handling of the metric. You could use the same code if:

- there's a metric in both cases
- metric changes are conditional on `i>0`

prow/tide/status_test.go
stevekuznetsov: Easy to write, hard to review ;)

prow/tide/tide.go
jmguzik: Ok, so in package tide (which includes `status.go` and `tide.go`) I use one function ``queriesInParallel`` to handle case without metric and with metric, using one struct  `queryData`. Indeed, in  `status.go` index is useless, in `tide.go` I use it as you noticed.
Now you want me to:
1. Create another function for querying in pararell as using this struct in `status.go` is not quite right, since I do not use metricIndex there at all
2. Modify somehow the struct, do not create another function
3. Something else?

Sorry for being so descriptive, but I think I did not get the remark.

There is also similar function in `blockers.go` but it's another package and function itself is a little bit different. I just thaught, exporting it from package blockers to package tide is a little silly.

stevekuznetsov: I'm saying that this PR, in its current state, is checking in two large blocks of code that do an identical thing in an identical manner, aside from the metrics register. I would ask that the code be shared. I could see you sharing the code by either:

 - exposing errors in a metric for all types of queries (why should we see query errors in one place but not another?)
 - making the handling of the metric optional when none is specified in the `struct` you're operating over

jmguzik: Putting aside metrics I think you actually referring to 2 similar functions in `blockers.go` and `tide.go`. Note that `blockers` is a separate package. I can make one function probably, but where to put it? Could you advise on that? There is no something like `utils.go `in tide. It more belongs to tide package but including it in tide package and exporting means importing the whole tide in blockers. I don't think it's desired.

Regarding metrics you are right, I can add them to the other queries as well. Wouldn't hurt.

stevekuznetsov: `prow/tide/queryutil` or similar should suffice and make it so that there are no import cycles. You may need to move the metrics around

prow/tide/queryutil/queryutil.go
```diff
@@ -0,0 +1,54 @@
...
+)
+
+type QueryData struct {
+	Org         string
+	Query       string
+	MetricIndex int
+}
+
+// QueriesInParallel runs process function concurrently respecting maximum number of goroutines
+func QueriesInParallel(goroutines uint, queries []QueryData, process func(QueryData)) {
```
stevekuznetsov, line 32: Sorry, but now this construct is awkward. The reason that the previous code was not ideal was that it was doing the same thing - running GraphQL queries using a client against the same server, etc. This incarnation is a "generic" thread pool where the arguments are passed in a slice that's coupled to the execution functions being called. I don't think there's a huge amount of utility in "generic" solutions until the type system supports parametric polymorphism. To wit - this helper could be even more utilitarian if you close over the arguments to the function itself, but, again, that's not really that interesting of a module:

```go
func DoWhateverInParallel(concurrency uint, work []func()) {}
```

Notably, the actual sending of the graphQL query and the incrementation of the error counters is now (still) duplicated in both callers, and *that* was the common thread which could have used "DRY". Consider the case in the future when someone wants to change how `tide` interacts with GitHub v4 API. Given this current code, they'd need to go make the same change in a number of places.

prow/tide/tide.go
```diff
@@ -461,38 +484,37 @@ func (c *Controller) query() (map[string]PullRequest, error) {
...
-			}()
+			allQueries = append(allQueries, queryutil.QueryData{Query: q, Org: org, MetricIndex: i})
 		}
 	}
-	wg.Wait()
+	queryutil.QueriesInParallel(
+		c.config().Tide.MaxGraphQLGoroutines,
+		allQueries,
+		func(q queryutil.QueryData) {
+			results, err := search(c.ghc.QueryWithGitHubAppsSupport, c.logger, q.Query, time.Time{}, time.Now(), q.Org)
```
stevekuznetsov, line 495: For example - these lines (from `search` to the metrics update) must occur in *every* function that is passed to `QueriesInParallel` and are the bits that actually end up using the parameters in the query type.

jmguzik: `search` from `blockers.go` is different than search from `tide.go`. I am afraid it can't be put to the same bucket as you want.

jmguzik: So actually your finding here is not true. I found a time to look at this more closely. In `blockers.go` there is a need to query issues while in `tide.go` and `search.go` PRs (+ these are also dated queries). Maybe there can be found some kind of union of this, but may also result in Frankenstein function.
Metrics and error list is handled in both cases, but is this enough to create some kind of specific mechanism? Your comment^ may also apply to metrics and errors as well. Anyway, max that could be added to make above more specific is:
```
resultString := "success"
if err != nil {
    resultString = "error"
}
    tideMetrics.queryResults.WithLabelValues(strconv.Itoa(q.MetricIndex), q.Org, resultString).Inc()
```
Functions `search` are named the same, but the actual use cases of specific approaches are different (`tide.go`, `status.go` vs `blockers.go`, the body is different).
We had 2x queriesInPararell, one in `tide` package, second in `blockers`. We have now "too generic" one occurrence (sorry for that, this is how I interpreted your request). I get that you would like to unify more this code, but I don't see much more except the lines I pasted here.

prow/tide/queryutil/queryutil.go
jmguzik: Well, every query is a specific thing and you want a struct for its representation. Unless you decide to compose it dynamically (use reflect) you can't unify different cases. In `blockers` you have issues in `tide` you have PRs. Max you can do is this + maybe some kind of error handling and metrics (see another comment). Any way you will end up with the same quite generic mechanism.

jmguzik:
@alvaroaleman @stevekuznetsov I closed this PR due to lack of time, but recently due to the growth of openshift org and switching tide to GH Apps, rate limiting issues started to pop up again.
I have developed an algorithm for ghproxy which per org, per API is creating a queue of messages. It limited 403s related to secondary rate limiting, but after applying 1s delay between messages for API v4 (actually 1050 ms) still 10-15% of API v4 messages coming from openshift organization (big tide query) during daily peaks (12 CEST, 16-18 CEST) are landing with 403 error codes related to rate limiting.
So, @alvaroaleman I know you were against this PR, but at least it would help me debug what additional GH "rules" are playing a part when speaking about rate limiting at scale. If no objections I will try to rebase, refactor and reopen when time allows. Please treat this as a temporary countermeasure/observability opportunity as I plan to tackle the problem in general in tide in the future (more event based oriented research).

PR #24610, opened 2021-12-09
sig-instrumentation: increase code coverage for JSON periodic job

pohly:
We need to trigger more log calls to catch potential mistakes and to get
representative log output for components under load.

Previous runs at level 6 produced reasonably sized logs and there aren't that
many log calls with higher levels, so it should be save to also enable
those - unless those end up being called frequently...

### DIFF
--- a/config/jobs/kubernetes/sig-instrumentation/sig-instrumentation-kind-periodics.yaml
+++ b/config/jobs/kubernetes/sig-instrumentation/sig-instrumentation-kind-periodics.yaml
@@ -38,8 +38,7 @@ periodics:
         value: json
       - name: KIND_CLUSTER_LOG_LEVEL
         # Default is 4, but we want to exercise more log calls and get more output.
-        # TODO: go even higher, depending on log volume
-        value: "6"
+        value: "10"
       # don't retry conformance tests
       - name: GINKGO_TOLERATE_FLAKES
         value: "n"


pohly:
/wg structured-logging
/cc @serathius 

This is a follow-up to https://github.com/kubernetes/test-infra/pull/24410 and part of https://github.com/kubernetes/kubernetes/issues/102430.

ehashman (commented):
/lgtm
/approve

this job is only every 6h, please follow-up with artifact size so we don't burn too much storage on this, we may want to decrease frequency if it's too high

pohly:
/assign @serathius 

For approval.

> this job is only every 6h, please follow-up with artifact size so we don't burn too much storage on this, we may want to decrease frequency if it's too high

Will do. I already reduced the frequency from 1h to 6h in the previous PR. I think we can go for even less frequently because it's not like someone is making a change and then immediately waiting for a run with those changes. Well, except that now I do, but that's the exception... :grin:

pohly:
The size of all artifacts increased from 506M at log level 6 to 1.3G at log level 10. Is that still okay?

The size increase is primarily because client-go starts dumping the message content. Unfortunately it does that with `klog.Infof` after checking verbosity:
https://github.com/kubernetes/kubernetes/blob/cc6f12583f2b611e9469a6b2e0247f028aae246b/staging/src/k8s.io/client-go/rest/request.go#L1171-L1184

The resulting log messages are therefore recorded with "v=0", which makes the log files unsuitable for benchmarking. I'll fix it and also look for other cases of that pattern.

ehashman:
@pohly will it be a problem if we increase the interval time to about 2x (so 12h) to adjust for it? I'm not 100% sure if we're hitting any cost issues but may as well be sensitive. This is a benchmarking test so that's not an unreasonable interval

pohly:
Not at all, every 12h is fine. I don't know why it ran once per hour originally, that seemed way too often to me.

PR #24614, opened 2021-12-09
Fix SIG Instrumentation OWNERS

ehashman:
Cleanup - we're missing the leads alias, removing emeritus chair and any leads already on the alias.

/sig instrumentation


spiffxp (approved):
/approve
/lgtm

PR #24148, opened 2021-10-27
Prow deployment job uses make rule

chaodaiG:
/hold
This also changes from schemaless CRD to full CRD, kubectl apply would fail, will need to do it manually before merging this PR

PR #24615, opened 2021-12-09
Move prow CRD yaml to one level below

chaodaiG:
This will make it possible for make rules to do kubectl apply -f config/prow/cluster, without need to know which prow CRD is applied. For now we'll use schemaless CRD


chaodaiG:
Related https://github.com/kubernetes/test-infra/pull/24148#issuecomment-990121197

config/prow/Makefile
```diff
@@ -32,6 +32,7 @@ update-plugins: get-cluster-credentials
 	kubectl create configmap plugins --from-file=plugins.yaml=plugins.yaml --dry-run -o yaml | kubectl replace configmap plugins -f -
 
 deploy-prow: get-cluster-credentials
+	kubectl apply --server-side=true -f ./cluster/prowjob-crd/legacy/prowjob-schemaless_customresourcedefinition.yaml
```
cjwagner, line 35: Are there any issues with using server-side apply for the schemaless CRD?

cjwagner (commented):
Please also update [`deploy.sh`](https://github.com/kubernetes/test-infra/blob/07d01555bb5110f6b69f011a6579430c0a2a8170/prow/deploy.sh#L89)

config/prow/Makefile
chaodaiG: should not be

chaodaiG:
> Please also update [`deploy.sh`](https://github.com/kubernetes/test-infra/blob/07d01555bb5110f6b69f011a6579430c0a2a8170/prow/deploy.sh#L89)

nice catch, done!

cjwagner (approved):

PR #24624, opened 2021-12-09
Revert "Revert "Prow deployment job uses make rule""

chaodaiG:
This reverts commit 77dc9f8e78315d28bd3e61a6581564eac6c415e4.

depend on #24615 

Should be an easy migration since make rule deploys the same set of yamls from before

/cc @cjwagner @alvaroaleman

### DIFF
--- a/config/jobs/kubernetes/test-infra/test-infra-trusted.yaml
+++ b/config/jobs/kubernetes/test-infra/test-infra-trusted.yaml
@@ -132,10 +132,11 @@ postsubmits:
       containers:
       - image: gcr.io/k8s-testimages/gcloud-bazel:v20210128-v0.6-11-gf1c6399
         command:
-        - prow/deploy.sh
+        - make
         args:
-        - --confirm
-        - --config=trusted
+        - -C
+        - config/prow
+        - deploy-all
     annotations:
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes/test-infra/test-infra-trusted.yaml
```diff
@@ -132,10 +132,11 @@ postsubmits:
       containers:
       - image: gcr.io/k8s-testimages/gcloud-bazel:v20210128-v0.6-11-gf1c6399
```
cjwagner, line 133: Does this image have `make` and `kubectl`?

cjwagner (approved):
/hold

config/jobs/kubernetes/test-infra/test-infra-trusted.yaml
chaodaiG: yeah it's a ubuntu image with kubectl installed https://github.com/bazelbuild/rules_k8s/blob/d05cbea5c56738ef02c667c10951294928a1d64a/images/gcloud-bazel/Dockerfile#L58

I'll unhold tomorrow morning

chaodaiG:
Have tested with --dry-run=client and should work
/unhold

PR #24636, opened 2021-12-10
Deflake TestDiskCachePruning

chaodaiG:
As mentioned in https://github.com/kubernetes/test-infra/issues/24200#issuecomment-991312622, the test was flaky because of token generated based on timestamp rounded by second

tested with `go test -count=50` and it worked

/cc @alvaroaleman


alvaroaleman (approved):
thanks!

PR #24645, opened 2021-12-10
switch to containerd

lizhuqi:
Removal of dockershim is scheduled for Kubernetes v1.24 so switch to containerd


andyzhangx (approved):
/lgtm
/approve
/ok-to-test

Issue #24537, opened 2021-12-03
Allowances for non-Prow YAML in the ProwJob config directory tree.


cjwagner:
**Why is this needed**:

The newly added checkconfig validation of unknown fields in ProwJob config has helped us fix a number of problems, but we are unable to enable it everywhere we'd like due to conflicts with non-Prow YAML that exists in the ProwJob config directory tree. In particular this seems to come in two forms:
1. Non-Prow YAML files collocated with their related ProwJob config files.
e.g. Template files used for ProwJob config generation and K8s resource files associated with the jobs.
1. Non-Prow fields included in files with valid ProwJob config.
e.g. Extra fields to define YAML anchors and intentionally changing top level field names to disable parts of the config (`presubmits`->`presubmits_disabled`).

**What would you like to be added**:

I'd like to add some features to allow non-Prow YAML to be explicitly labelled as such so that it can be ignored by unknown field validation. I have an idea for each of the above cases:
1. Update directory walk in the config loading code for both JobConfig and in-repo config to ~check for a `# PROW IGNORE` comment on the first line of the file. Skip loading the file if it is found.~ respect `.prowignore` files that behave like `.gitignore` files.
1. Add a top level `prow-ignored` field to JobConfig and in-repo config that has type `json.RawMessage` to avoid parsing. Any YAML anchors can be defined here as could configuration that is intended to be disabled or for another tool.

Please let me know if you have any concerns or better ideas. I'll start with a quick PR for the second idea.
/assign @alvaroaleman @stevekuznetsov @chaodaiG

listx:
> 1. Update directory walk in the config loading code for both JobConfig and in-repo config to check for a # PROW IGNORE comment on the first line of the file. Skip loading the file if it is found.

Another idea is to create a `.prowignore` file in the style of `.gitignore` or `.rgignore`. Takes more work but from the user perspective I think it would feel cleaner.

> intentionally changing top level field names to disable parts of the config (presubmits->presubmits_disabled)

At first glance I am hesitant to try to cater to this practice. Isn't this what YAML comments are for? Why aren't users simply commenting out the whole field/block?

chaodaiG:
> At first glance I am hesitant to try to cater to this practice. Isn't this what YAML comments are for? Why aren't users simply commenting out the whole field/block?

yaml anchor will not work in comment, so this is not feasible. Same with .prowignore

cjwagner:
[...]

PR #24646, opened 2021-12-11
Make job config loading respect `.prowignore` files.

cjwagner:
Addresses part 1 of https://github.com/kubernetes/test-infra/issues/24537

/cc @alvaroaleman @chaodaiG

### DIFF
--- a/prow/config/config.go
+++ b/prow/config/config.go
@@ -37,6 +37,7 @@ import (
 	"text/template"
 	"time"
 
+	gitignore "github.com/denormal/go-gitignore"
 	"github.com/google/go-cmp/cmp"
 	"github.com/sirupsen/logrus"
 	pipelinev1alpha1 "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1alpha1"
@@ -72,6 +73,8 @@ const (
 	ConfigVersionFileName = "VERSION"
 
 	DefaultTenantID = "GlobalDefaultID"
+
+	ProwIgnoreFileName = ".prowignore"
 )
 
 // Config is a read-only snapshot of the config.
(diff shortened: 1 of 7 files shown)


prow/config/config.go
```diff
@@ -1357,6 +1358,10 @@ func ReadJobConfig(jobConfig string, yamlOpts ...yaml.JSONOpt) (JobConfig, error
 		return jc, nil
 	}
 
+	prowIgnore, err := gitignore.NewRepositoryWithFile(jobConfig, ".prowignore")
```
petr-muller, line 1361: nit: maybe `.prowignore` should be a constant, it's used from two places?

petr-muller (approved):
/hold

LGTM, holding for assigned reviewers

prow/config/config.go
```diff
@@ -1379,10 +1384,14 @@ func ReadJobConfig(jobConfig string, yamlOpts ...yaml.JSONOpt) (JobConfig, error
 			}
 			return nil
 		}
-
 		if filepath.Ext(path) != ".yaml" && filepath.Ext(path) != ".yml" {
```
chaodaiG, line 1387: understand that this is not from your PR, but I'm pretty sure that `.yml` files are not used by prow, do you know is this expected?

cjwagner: Looks like this changed back in 2018 https://github.com/kubernetes/test-infra/pull/9137. I'm pretty sure both are considered valid extensions so its definitely better to respect both.

prow/config/config.go
cjwagner: Switched to a constant throughout.

cjwagner:
Comments addressed, PTAL.

petr-muller (approved):

prow/config/config.go
chaodaiG: ahh.. taking a deeper look and the problem was `config/jobs/**/*.yaml` at https://github.com/kubernetes/test-infra/blob/91011caf8dca5b240f34bd04a4091567cbe44e73/config/prow/plugins.yaml#L516, which is the reason why `.yml` files are not respected

chaodaiG (approved):

prow/config/config.go
chaodaiG: btw https://github.com/kubernetes/test-infra/pull/24675

cjwagner:
Thanks!
/hold cancel

PR #24658, opened 2021-12-13
don't skip load balancer tests

aojea:
use the same regex as in the presubmit
```
  - name: pull-kubernetes-e2e-gci-gce-ingress
```


MrHohn (approved):
Thanks!
/lgtm

PR #24665, opened 2021-12-14
Make grafana deployment run as non-root and enable seccomp policy.

cjwagner:
/assign @mtaufen

### DIFF
--- a/config/prow/cluster/monitoring/grafana_deployment.yaml
+++ b/config/prow/cluster/monitoring/grafana_deployment.yaml
@@ -26,9 +26,9 @@ spec:
         readinessProbe:
           httpGet:
             path: /healthy
-            port: 80
+            port: 8080
         ports:
-          - containerPort: 80
+          - containerPort: 8080
             name: nginx
         securityContext:
           allowPrivilegeEscalation: false
@@ -101,8 +101,6 @@ spec:
           readOnly: false
       nodeSelector: {}
       securityContext:
-        capabilities:
-          add: ["NET_BIND_SERVICE"]
         seccompProfile:
           type: RuntimeDefault
         runAsUser: 1000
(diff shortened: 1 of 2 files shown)


config/prow/cluster/monitoring/grafana_deployment.yaml
```diff
@@ -17,6 +19,8 @@ spec:
       app: grafana
   template:
     metadata:
+      annotations:
+        seccomp.security.alpha.kubernetes.io/pod: 'runtime/default'
```
mtaufen, line 23: It looks like this is actually available in securityContext now and the annotations are getting removed in 1.25. Let's try the securityContext version: https://kubernetes.io/docs/tutorials/clusters/seccomp/#create-a-pod-with-a-seccomp-profile-for-syscall-auditing

config/prow/cluster/monitoring/grafana_deployment.yaml
```diff
@@ -30,6 +34,8 @@ spec:
         ports:
           - containerPort: 80
```
mtaufen, line 35: We might have to run with `CAP_NET_BIND_SERVICE` to bind low ports when running as non-root, depending on whether the Docker version is 1.19 or 1.20 (docker set `net.ipv4.ip_unprivileged_port_start=0` in 1.20, and I'm not sure what containerd does): https://github.com/moby/moby/pull/41030

cjwagner: I added a `capabilities` section to the pod's `securityContext` that I think should grant that capability.

cjwagner (commented):
Also replaced the annotation with the `seccompProfile` field.

config/prow/cluster/monitoring/grafana_deployment.yaml
stevekuznetsov: You could also just bind something higher and use the service to redirect it

cjwagner:
I've switched the nginx proxy to serve on 8080 and redirect to 80 via the service as Steve suggested. PTAL.

cjwagner:
The monitoring dashboard still seem to be working since this deployed. :+1:

Issue #24508, opened 2021-11-30
Draft PR converted to RfR, but stuck with do-not-merge/work-in-progress label


afirth:
**What happened**:
After converting https://github.com/kubernetes/ingress-nginx/pull/8000 from draft to ready, prow [removed the do-not-merge/work-in-progress label and then immediately re-added it](https://github.com/kubernetes/ingress-nginx/pull/8000#event-5694833013)

**What you expected to happen**:
The label should be removed

**How to reproduce it (as minimally and precisely as possible)**:
Not sure if a flake or really broken, nor what repo to test it in

**Please provide links to example occurrences, if any**:
https://github.com/kubernetes/ingress-nginx/pull/8000#event-5694833013

**Anything else we need to know?**:
I was able to work around the issue by prefixing `WIP` to the title and then removing it

https://github.com/kubernetes/test-infra/issues/11939 previous issue but long fixed

afirth:
thank you @shuheiktgw for attempting to reproduce. I guess from the lack of activity that this was a freaky flake and will
/close

PR #24668, opened 2021-12-14
Testing the wip plugin

shuheiktgw:
I'd like to look into https://github.com/kubernetes/test-infra/issues/24508 so let me create a PR to test the behavior


shuheiktgw:
Hmm, cannot reproduce the behavior...

PR #24683, opened 2021-12-15
Disable DisableCloudProviders feature gate for k8sstable1 tests

xmudrii:
The DisableCloudProviders feature gate is supposed to be disabled for 1.22+ tests because our tests are not using external CCMs.

/assign @saschagrunert 

### DIFF
--- a/releng/test_config.yaml
+++ b/releng/test_config.yaml
@@ -105,8 +105,7 @@ jobs:
     args:
     - --env=KUBE_PROXY_DAEMONSET=true
     - --env=ENABLE_POD_PRIORITY=true
-    # TODO(releng): add "InTreePluginGCEUnregister=false" to feature gates when this 1.21 reaches k8sstable1
-    - --env=KUBE_FEATURE_GATES=AllAlpha=true,CSIMigration=false
+    - --env=KUBE_FEATURE_GATES=AllAlpha=true,DisableCloudProviders=false,CSIMigration=false,InTreePluginGCEUnregister=false
     # Panic if anything mutates a shared informer cache
     - --env=ENABLE_CACHE_MUTATION_DETECTOR=true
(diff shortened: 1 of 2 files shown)


releng/test_config.yaml
```diff
@@ -105,8 +105,7 @@ jobs:
     args:
     - --env=KUBE_PROXY_DAEMONSET=true
     - --env=ENABLE_POD_PRIORITY=true
-    # TODO(releng): add "InTreePluginGCEUnregister=false" to feature gates when this 1.21 reaches k8sstable1
-    - --env=KUBE_FEATURE_GATES=AllAlpha=true,CSIMigration=false
+    - --env=KUBE_FEATURE_GATES=AllAlpha=true,DisableCloudProviders=false,CSIMigration=false,InTreePluginGCEUnregister=false
```
saschagrunert, line 108: Can we only change the `DisableCloudProviders` for now?

xmudrii: I copied this from the 1.23 job and I think we need at least `InTreePluginGCEUnregister=false` in addition to `DisableCloudProviders`.

* CSIMigration=false doesn't have an effect because even if it's enabled and there's no CSI driver, it will fallback to the in-tree cloud provider. We can eventually drop it.
* InTreePluginGCEUnregister=false is required because without it, GCE-backed PVs/PVCs would be disabled since we don't have a CSI driver. This might cause other tests to fail.

If you want, I can drop CSIMigration, but I think it's also fine to leave it since we already have it for 1.23 tests.

saschagrunert: Yes, that's fine thank you for the clarification!

saschagrunert (approved):

saschagrunert:
/override pull-test-infra-gubernator

To get the config change in and unblock the release.

PR #24686, opened 2021-12-15
[mkpj] add an option to faile mkpj if the triggered job fails

rmohr:
If `--trigger-job` and `--fail-with-job` is supplied, mkpj fails too if
the watched job fails. This allows easy ad-hoc integration of mkpj and
remote jobs.

One example is when one CI system has to trigger a job in prow.


petr-muller (approved):

rmohr:
@petr-muller thanks a lot for the review

/cc @cjwagner 

according to the prow notice.

cjwagner (approved):

Issue #24687, opened 2021-12-15
pull-test-infra-gubernator presubmit is consistently failing for all PRs


xmudrii:
**What happened**:

The `pull-test-infra-gubernator` job/presubmit is consistently failing for all PRs, therefore blocking them. It appears that there's some problem with Python dependencies. An example run can be found here: https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/test-infra/24683/pull-test-infra-gubernator/1471123708318846976

Here's a part of the log:

```
Collecting lazy-object-proxy (from astroid<1.5.0,>=1.4.5->pylint==1.6.4->-r test_requirements.txt (line 3))
  Downloading https://files.pythonhosted.org/packages/6f/3e/7c80e8536b9d5eb66e784fff4c359adf9e55b37460fd6928192256ba71f2/lazy-object-proxy-1.7.0.tar.gz (42kB)
...
      File "/tmp/easy_install-JsFNwm/setuptools_scm-6.3.2/setup.py", line 22
        f"{type(self).__name__} is forbidden, "
                                              ^
    SyntaxError: invalid syntax
    
    ----------------------------------------
```

**What you expected to happen**:

The `pull-test-infra-gubernator` job/presubmit is passing for all PRs.

**Please provide links to example occurrences, if any**:

#24683 but other PRs are affected as well.

**Anything else we need to know?**:

/priority critical-urgent
This is blocking Kubernetes 1.22.5 release scheduled for today.

cc: @kubernetes/release-engineering

michelle192837:
Looks like this started failing between these two runs:

```
1470940785301000192 | #24679Update Prow to v20211214-938f24ffd3 (44cef0a) by k8s-ci-robot | Dec 14 18:17:01Dec 14 2021, 18:17:01 UTC-0800 | 9s | FAILURE
-- | -- | -- | -- | --
1470910530590871552 | #24679Update Prow to v20211214-938f24ffd3 (f5a546f) by k8s-ci-robot | Dec 14 16:16:45Dec 14 2021, 16:16:45 UTC-0800 | 1m12s | SUCCESS
```

e.g. on Dec. 14 between 8:16 and 10:17 am PST

I don't think this lines up with recent Prow/image bumps:
- https://github.com/kubernetes/test-infra/pull/24667
- https://github.com/kubernetes/test-infra/pull/24677
- https://github.com/kubernetes/test-infra/pull/24674

[...]

PR #24688, opened 2021-12-15
Add Windows serial/slow capz jobs

jsturtevant:
serial/slow jobs are now supported in capz: https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1881

/sig windows
/assign @marosset 

fyi @nick5616 for adding ws2022 serial jobs.


jsturtevant:
failure is not related: https://github.com/kubernetes/test-infra/issues/24687

jsturtevant:
should be fix in https://github.com/kubernetes/test-infra/issues/24687

/retest

marosset:
I think that either the jobs shouldn't be `staging` jobs or we they should not be included in the `sig-windows-master-release` dashboards.

Should the job names be `ci-kubernetes-e2e-capz-master-*`?

jsturtevant:
this naming was from the now removed dockershim shim jobs which had this name where staging=master jobs.  I would like to rename all of them eventually away from master but was sort of waiting for k/k to move to main to avoid confusion

I will update this temporarily to make the renaming easier later.

PR #24691, opened 2021-12-15
Run python linter in docker

chaodaiG:
This is part of effort for build/test without bazel. Running in docker for ensuring toolchain consistency.


config/jobs/kubernetes/kops/build_jobs.py
```diff
@@ -16,8 +16,8 @@
 import math
 import json
 import re
-import jinja2 # pylint: disable=import-error
 import yaml
+import jinja2 # pylint: disable=import-error
```
BenTheElder, line 20: should we ensure the linter _can_ actually import? I think this just means we didn't install deps before linting?

hack/build/ensure-python_deps.sh
```diff
@@ -0,0 +1,34 @@
...
+set -o nounset
+set -o pipefail
+
+REPO_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd -P)"
+cd "${REPO_ROOT}"
+
+# Ensure virtual env
+# Trick from https://pythonspeed.com/articles/activate-virtualenv-dockerfile/
+echo "Ensure virtual env"
+export VIRTUAL_ENV=".python_virtual_env"
```
BenTheElder, line 28: do we need to export this? (seems like no?)

hack/run-in-python-container.sh
```diff
@@ -0,0 +1,47 @@
...
+#     afterwards.
+#   - Run `source venv_dir/bin/activate` as part of docker run argument. This
+#     cannot be done directly as `source` is a bash builtin, so have to do `bash
+#     -c `source.., python3..` stuff, and the quotes passing around almost
+#     killed me.
+docker run \
+    --rm -i \
+    -e HOME=/tmp \
+    -e PYLINTHOME=${PYLINTHOME} \
+    -e PATH="${REPO_ROOT:?}/.python_virtual_env/bin:/usr/local/bin:/bin" \
```
BenTheElder, line 42: I think maybe this should have a default and the path to the virtual env should be overridable with an env so we can ensure consistency between the scripts?

chaodaiG: ack and done

hack/build/ensure-python_deps.sh
chaodaiG: it's exported so that the default mentioned in `hack/run-in-python-container.sh` could work

config/jobs/kubernetes/kops/build_jobs.py
chaodaiG: doesn't have much context here, but seems like this package is imported in a weird way hack/run-in-python-container.sh

BenTheElder: It doesn't look to me like we install deps before linting?

hack/build/ensure-python_deps.sh
BenTheElder: at the time there wasn't one though?

hack/run-in-python-container.sh
```diff
@@ -0,0 +1,52 @@
...
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+REPO_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd -P)"
+cd "${REPO_ROOT}"
+
+PY_IMAGE='python:3.9-slim-buster'
+PYLINTHOME="$(mktemp -d)"
```
BenTheElder, line 26: why is pylint in a tempdir created every run?
does the tempdir need to be bind-mounted at all then?
why don't we delete the tempdir? or use something under /tmp inside the container and don't bind-mount?

hack/run-in-python-container.sh
```diff
@@ -0,0 +1,52 @@
...
+
+# Setting PATH is a trick learnt from
+# https://pythonspeed.com/articles/activate-virtualenv-dockerfile/, for
+# mimicking the behavior of `source venv_dir/bin/activate`, same as setting
+# PYTHONPATH to empty. Tried several other alternatives:
+#   - Set PATH in docker image. This doesn't work because the workspace is set
+#     afterwards.
+#   - Run `source venv_dir/bin/activate` as part of docker run argument. This
+#     cannot be done directly as `source` is a bash builtin, so have to do `bash
+#     -c `source.., python3..` stuff, and the quotes passing around almost
```
BenTheElder, line 40: I think just `docker run ... "${PY_IMAGE}" bash -c 'source venv_dir/bin/activate && exec "$@"' " $@"` should do what you describe here.

chaodaiG: I'll give it a shot

hack/run-in-python-container.sh
chaodaiG: this was almost copy/pasted from existing script, didn't look into that much detail what `PYLINTHOME` is used for, will take a deeper look

config/jobs/kubernetes/kops/build_jobs.py
chaodaiG: sorry for the confusion, I had a bad copy/paste in previous response, meant to link https://github.com/kubernetes/test-infra/blob/b5865074c4ed1cfb41f9e455e9860a6a4e60794b/gubernator/lint.sh#L17, package was linted with a separate script, and I'm not sure whether we want to incorporate this logic globally

hack/run-in-python-container.sh
chaodaiG: It's used by pylint according to https://pylint.pycqa.org/en/latest/faq.html#where-is-the-persistent-data-stored-to-compare-between-successive-runs, for storing pylint analysis results, which is not super critical

hack/run-in-python-container.sh
chaodaiG: good to learn this trick, seems like it should work, until another day of learning again "bash is hard"! I got stuck trying to understand this:

```
<script.sh>
echo "outside: $@"
bash -c 'echo "inside: $@"' "$@"
```

```
$ script.sh pip3 install -r requirements3.txt

outside: pip3 install -r requirements3.txt
inside: install -r requirements3.txt
```

somehow `pip3` disappeared

BenTheElder: Ah right, it will need to include `"$0 $@"` in the inner exec, because $0 is dropped from $@ each time.

The first time this is desired here. Repeat times it is not.

`argv[0]` should contain the pathname of the executable being invoked, so programs like bash treat it specially and it is not in $@ as a program argument (which is ~`argv[1:]`).

Reference for more: https://man7.org/linux/man-pages/man2/execve.2.html

chaodaiG: had an offline chat with Ben, though this works, the nested quotes don't look elegant. Tried to introduce `activate.sh` script that is depended by other python test scripts, which does the activation, seems like a good solution, only down side is more scripts

config/jobs/kubernetes/kops/build_jobs.py
BenTheElder: Er for gubernator I think this is because we're not handling importing it outside of GAE magic 
https://github.com/kubernetes/test-infra/blob/f7e21a3c18f4f4bbc7ee170675ed53e4544a0632/gubernator/app.yaml#L39

which feels bad, I'm not sure how kops was related though ..?

hack/build/ensure-python_deps.sh
```diff
@@ -0,0 +1,34 @@
...
+set -o nounset
+set -o pipefail
+
+REPO_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd -P)"
+cd "${REPO_ROOT}"
+
+# Ensure virtual env
+# Trick from https://pythonspeed.com/articles/activate-virtualenv-dockerfile/
+echo "Ensure virtual env"
+export VIRTUAL_ENV="${REPO_ROOT}/.python_virtual_env"
```
BenTheElder, line 28: seems like this should have the logic de-duped from activate-python_venv.sh below where we check if activate exists or not?

hack/run-in-python-container.sh
```diff
@@ -0,0 +1,33 @@
...
+set -o errexit
+set -o nounset
+set -o pipefail
+
+REPO_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd -P)"
+cd "${REPO_ROOT}"
+
+PY_IMAGE='python:3.9-slim-buster'
+
+docker run \
```
BenTheElder, line 27: I think maybe we should just always ensure and activate the venv in this script, it would de-dupe a lot of logic and ensure we don't accidentally try to do things outside of the venv?

chaodaiG: do you mean `docker run <IMAGE> "source activate.sh && $@"` here?

BenTheElder: that and or/ there's also some logic elsewhere to ensure the venv exists etc., is there a reason to be running in this environment without the venv?

chaodaiG: There is no reason to run this without venv, so I think we are safe to do it here.

chaodaiG: PR is updated

hack/make-rules/py-test/activate-python_venv.sh
```diff
@@ -0,0 +1,34 @@
...
+set -o errexit
+set -o nounset
+set -o pipefail
+
+REPO_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/../../.." && pwd -P)"
+cd "${REPO_ROOT}"
+
+# Ensure virtual env
+# Trick from https://pythonspeed.com/articles/activate-virtualenv-dockerfile/
+echo "Ensure virtual env"
```
BenTheElder, line 25: nit for follow-up: I think maybe printing this every time the parent script is invoked is just log-noise?

Also the echo below about activating the venv. I think we just need to debug if we created the env. The venv is a given now for python-in-docker.

config/jobs/kubernetes/kops/build_jobs.py
BenTheElder: I still think this should be fixed, but leaving to a follow-up. we should poke one of the kops owners at least?

hack/make-rules/py-test/activate-python_venv.sh
```diff
@@ -0,0 +1,34 @@
...
+export VIRTUAL_ENV="${REPO_ROOT}/.python_virtual_env"
```
BenTheElder, line 26: noting that it doesn't look like there's any accepted standard for where to place these, other than possibly when doing it manually under ~/.something

BenTheElder (approved):
/lgtm
/approve

PR #24694, opened 2021-12-16
don't run containerd only tests on docker for 1.23 tests

jsturtevant:
These 1.23 jobs are failing: https://testgrid.k8s.io/sig-windows-1.23-release#aks-engine-windows-dockershim-1.23

The set of tests are failing should not be run on dockershim as they have `[Excluded:WindowsDocker]` tags.

/sig windows
/assign @marosset


marosset:
/retitle don't run containerd only tests on docker for 1.23 tests

PR #8527, opened 2018-06-30
Trust repo collaborators unless disabled

fejta:
* Create `only_org_members: true` option in `triggers:` config.
  - Trust repo collaborators unless this options is set
* Set `only_org_members: true` for kubernetes repos.
* Fix some bugs where we used `trustedOrg` instead of `org`.
* Refactor a bunch of stuff in process

/assign @cblecker @cjwagner @krzyzacy 
/hold

PR #24708, opened 2021-12-17
prow/plugins/verify-owners: Default join-org URL to orgs/*/people

wking:
Bringing over the current version of the defaulting that the pull-request trigger has had since 2c80848bfd (#8527), to avoid an empty link like:

    ...add them as <a href="">members</a> of...

when `join_org_url` was not configured.

### DIFF
--- a/prow/plugins/verify-owners/verify-owners.go
+++ b/prow/plugins/verify-owners/verify-owners.go
@@ -334,6 +334,13 @@ func handle(ghc githubClient, gc git.ClientFactory, roc repoownersClient, log *l
 		}
 	}
 
+	var joinOrgURL string
+	if triggerConfig.JoinOrgURL != "" {
+		joinOrgURL = triggerConfig.JoinOrgURL
+	} else {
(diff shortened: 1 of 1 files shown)


prow/plugins/verify-owners/verify-owners.go
```diff
@@ -334,6 +334,13 @@ func handle(ghc githubClient, gc git.ClientFactory, roc repoownersClient, log *l
 		}
 	}
 
+	var joinOrgURL string
+	if triggerConfig.JoinOrgURL != "" {
+		joinOrgURL = triggerConfig.JoinOrgURL
+	} else {
+		joinOrgURL = fmt.Sprintf("https://github.com/orgs/%s/people", org)
```
wking, line 341: I didn't look into this too closely, but possibly it would be possible to centralize this defaulting in `prow/plugins/config.go` or some such, instead of relying on consumers to remember to default locally.

petr-muller (approved):

PR #24710, opened 2021-12-18
Remove typecheck for dockerless build tag

hakman:
As part of the removal of dockershim, we should also the `dockerless` build tag.

xRef: https://github.com/kubernetes/kubernetes/pull/107119


Namanl2001 (approved):
/lgtm

PR #24716, opened 2021-12-20
single node conformance local-up

aojea:
Conformance cluster requires 2 nodes, however, some Conformance tests work in single node clusters too.
The later doesn't mean the cluster is Conformance, that just passes these tests


adisky (commented):
/lgtm

aojea:
/assign @dims 
for approval

PR #24718, opened 2021-12-20
Adds --prepull-images=true to Windows jobs

claudiubelu:
A new option for E2E tests was introduced in Kubernetes v1.23 [1], which allows us to prepull some of the most common test images on the test nodes, like the busybox and agnhost images. Doing this will allow us to reduce the number of flakes we've been seeing that occur due to tests timing out (1 minute) while waiting for the images to be pulled.

[1] https://github.com/kubernetes/kubernetes/pull/105481

Related: https://github.com/kubernetes/test-infra/pull/24500

### DIFF
--- a/config/jobs/kubernetes-sigs/sig-windows/sig-windows-sac.yaml
+++ b/config/jobs/kubernetes-sigs/sig-windows/sig-windows-sac.yaml
@@ -46,7 +46,7 @@ periodics:
       - --aksengine-deploy-custom-k8s
       - --aksengine-agentpoolcount=2
       # Specific test args
-      - --test_args=--node-os-distro=windows --docker-config-file=$(DOCKER_CONFIG_FILE) --ginkgo.focus=\[Conformance\]|\[NodeConformance\]|\[sig-windows\]|\[sig-apps\].CronJob|\[sig-api-machinery\].ResourceQuota|\[sig-network\].EndpointSlice  --ginkgo.skip=\[LinuxOnly\]|\[Serial\]|\[Slow\]|\[alpha\]|GMSA|Guestbook.application.should.create.and.stop.a.working.application|device.plugin.for.Windows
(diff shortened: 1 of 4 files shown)


config/jobs/kubernetes-sigs/sig-windows/sig-windows-sac.yaml
```diff
@@ -101,7 +101,7 @@ periodics:
       - --aksengine-win-binaries
       - --aksengine-deploy-custom-k8s
       # Specific test args
-      - --test_args=--node-os-distro=windows --docker-config-file=$(DOCKER_CONFIG_FILE) --ginkgo.focus=(\[sig-windows\]|\[sig-scheduling\].SchedulerPreemption|\[sig-autoscaling\].\[Feature:HPA\]|\[sig-apps\].CronJob).*(\[Serial\]|\[Slow\])|(\[Serial\]|\[Slow\]).*(\[Conformance\]|\[NodeConformance\]) --ginkgo.skip=\[LinuxOnly\]|device.plugin.for.Windows
+      - --test_args=--node-os-distro=windows --docker-config-file=$(DOCKER_CONFIG_FILE) -prepull-images=true --ginkgo.focus=(\[sig-windows\]|\[sig-scheduling\].SchedulerPreemption|\[sig-autoscaling\].\[Feature:HPA\]|\[sig-apps\].CronJob).*(\[Serial\]|\[Slow\])|(\[Serial\]|\[Slow\]).*(\[Conformance\]|\[NodeConformance\]) --ginkgo.skip=\[LinuxOnly\]|device.plugin.for.Windows
```
aojea, line 104: the rest of flags has double `--` , does it matter?

claudiubelu: According to the golang flags documentation [1], "One or two minus signs may be used; they are equivalent.". But indeed, how many are used is a bit inconsistent. I've put only one, since that's how many I've put in the previous commit [2], and it worked fine [3]. It was consistent with how many minuses the docker-config-file config option had.

[1] https://pkg.go.dev/flag#hdr-Command_line_flag_syntax
[2] https://github.com/kubernetes/test-infra/pull/24500
[3] https://github.com/kubernetes/kubernetes/issues/102100#issuecomment-982659136

aojea: no idea honestly, just was doing the diff , these things always surprise me and doesn't seem to be consistency 
if you tested it then it has to be ok

aojea:
@claudiubelu the change seems pretty straightforward , do you want me to lgtm or is better to wait for some reviewer on this folder?

PR #24602, opened 2021-12-09
Switch windows gce jobs from dockershim to containerd

marosset:
Signed-off-by: Mark Rossetti <marosset@microsoft.com>

/sig windows
/assign @jsturtevant @pjh 

Also, 2004 is out of service so I removed those jobs.

PR #24730, opened 2021-12-22
Use containerd as the container runtime in the Windows storage jobs

mauriciopoppe:
Filed a bug in https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/issues/880, eventually found out that the master kubernetes repo removed dockershim support which is still the default value in kube-up.sh https://github.com/kubernetes/kubernetes/blob/c75d254beb662ea73736d2f123d0382b7f91dfbd/cluster/gce/config-default.sh#L102, sig-windows has already moved their jobs in https://github.com/kubernetes/test-infra/pull/24602

Removing the jobs for Windows distros that are no longer supported

### DIFF
--- a/config/jobs/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/gcp-compute-persistent-disk-csi-driver-windows.yaml
+++ b/config/jobs/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/gcp-compute-persistent-disk-csi-driver-windows.yaml
@@ -1,94 +1,4 @@
 periodics:
-- name: ci-gce-pd-csi-driver-latest-k8s-master-windows-containerd-2019
-  decorate: true
-  extra_refs:
-  - org: kubernetes-sigs
-    repo: gcp-compute-persistent-disk-csi-driver
-    base_ref: master
-    path_alias: sigs.k8s.io/gcp-compute-persistent-disk-csi-driver
-  interval: 4h
-  labels:
-    preset-k8s-ssh: "true"
-    preset-service-account: "true"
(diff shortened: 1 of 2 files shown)


config/jobs/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/gcp-compute-persistent-disk-csi-driver-windows.yaml
```diff
@@ -1,94 +1,4 @@
...
-      - name: PREPULL_YAML
-        value: "prepull-head.yaml"
-      image: gcr.io/k8s-staging-test-infra/kubekins-e2e:v20211217-ea95cec1d4-master
-      securityContext:
-          privileged: true
-  annotations:
-    testgrid-dashboards: provider-gcp-compute-persistent-disk-csi-driver
-    testgrid-tab-name: ci-win20h2-containerd-gcp-compute-persistent-disk-csi-driver
-    description: Kubernetes Integration tests for Kubernetes Master branch and Driver latest on Windows 20H2 with Containerd
 - name: ci-gce-pd-csi-driver-latest-k8s-master-windows-2004
```
jingxu97, line 2: we should remove 2004 job too?

mauriciopoppe: removed

mauriciopoppe:
hey @jingxu97, could you review this PR again please? I removed jobs for unused Windows distros

PR #24736, opened 2021-12-23
update tide docs and github client link

smg247:
While digging into `tide` I noticed a couple things that were outdated.


smg247:
@fejta would you mind taking a look at these quick documentation changes?

fejta (approved):

PR #24737, opened 2021-12-23
Replace deprecated kubectl `--serviceaccount` flag with an equivalent `--overrides` flag.

cjwagner:
> Flag --serviceaccount has been deprecated, has no effect and will be removed in 1.24.

/assign @listx
cc @2nd-half

### DIFF
--- a/workload-identity/bind-service-accounts.sh
+++ b/workload-identity/bind-service-accounts.sh
@@ -98,12 +98,14 @@ else
   sleep 2m
 fi
 
+
 pod-identity() {
   head -n 1 <(
     entropy=$(date +%S)
     set -o xtrace
     kubectl run --rm=true -i \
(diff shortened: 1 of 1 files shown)


workload-identity/bind-service-accounts.sh
```diff
@@ -98,12 +98,14 @@ else
...
 
+
 pod-identity() {
   head -n 1 <(
     entropy=$(date +%S)
     set -o xtrace
     kubectl run --rm=true -i \
-      "--context=$context" "--namespace=$namespace" "--serviceaccount=$name" \
+      "--context=$context" "--namespace=$namespace" \
+      --overrides="{\"spec\": {\"serviceAccount\": \"$name\"}}" \
```
listx, line 108: Nit: I would change `--overrides="` to `"--overrides=`.

cjwagner: Done.

PR #24751, opened 2021-12-28
kOps grid: pass extra_args even when not passing image

justinsb:
Fix a small logic bug in the generation logic.

Thanks to @jiahuif for finding the problem and suggesting the fix!


jiahuif:
/lgtm
Now the extra `--gce-service-account=default` added in #24749 applies to the job. Great!

PR #24755, opened 2021-12-29
kOps GCE: Pass --gce-service-account=default to hand-coded tests

justinsb:
Verified as working for the grid tests.


jiahuif (approved):
/lgtm
I was about to send a PR doing exactly the same.

PR #24673, opened 2021-12-14
Move hugepages tests to CRI

cynepco3hahue:
Fixes #24617

PR #24757, opened 2021-12-29
Use containerd in Ubuntu swap job

mmiranda96:
Migrate swap Ubuntu job to containerd. For this, I modified the following:
- Moved the job to the more proper `containerd.yaml`.
- Configured the job to use containerd as container runtime.
- Included required config files in the image metadata (and retrieve them using the same approach as in #24673).
- Add myself as an owner for the job.

/sig node
/kind cleanup
/area jobs
/priority important-soon


SergeyKanzhelev (approved):
/lgtm
/approve

thank you!

PR #24759, opened 2021-12-30
Refactor log_dump_custom_get_instances into get-node-names

pjh:
get-node-names() should include the custom get-instances logic to make it easier to call from anywhere in the log-dump script.

This change is intended to have no behavioral effect. It will make it easier to call get-node-names() from additional locations in subsequent changes, such as https://github.com/kubernetes/test-infra/pull/24760.

### DIFF
--- a/logexporter/cluster/log-dump.sh
+++ b/logexporter/cluster/log-dump.sh
@@ -148,31 +148,51 @@ function detect-project() {
   fi
 }
 
-# Detect Linux and Windows nodes created in the instance group.
+# Detect Linux and Windows nodes in the cluster.
 #
-# Vars set:
+# If a custom get-instances function has been set, this function will use it
+# to set the NODE_NAMES array.
+#
+# Otherwise this function will attempt to detect the nodes based on the GCP
+# instance group information. If Windows nodes are present they will be detected
+# separately. The following arrays will be set:
 #   NODE_NAMES
 #   INSTANCE_GROUPS
 #   WINDOWS_NODE_NAMES
 #   WINDOWS_INSTANCE_GROUPS
 function detect-node-names() {
+  NODE_NAMES=()
+  INSTANCE_GROUPS=()
+  WINDOWS_INSTANCE_GROUPS=()
+  WINDOWS_NODE_NAMES=()
+
+  if [[ -n "${use_custom_instance_list}" ]]; then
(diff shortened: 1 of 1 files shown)


pjh:
/hold until I can end-to-end test this.

Reviewers, I'd still appreciate your input anytime.

pjh:
cc @ibabou

pjh:
I've e2e tested this by following the [gce windows e2e test steps](https://github.com/kubernetes/kubernetes/blob/master/cluster/gce/windows/README-GCE-Windows-kube-up.md#running-e2e-tests-against-the-cluster) for a single test, after symlinking the modified test-infra `log-dump.sh` file into my `kubernetes/cluster/log-dump/` directory. To override the use of the upstream/vendored test-infra `log-dump.sh`, I manually removed a flag from the end of the `run-e2e.sh` script (reverting https://github.com/kubernetes-sigs/windows-testing/pull/271) and set an environment variable](https://github.com/kubernetes/test-infra/pull/22292/files).

Then I ran this command to invoke a single e2e test:

`rm -r e2e-artifacts/*; PROJECT=xyz ZONE=us-central1-b KUBERNETES_PROVIDER=gce USE_TEST_INFRA_LOG_DUMPING=false KUBE_GCE_INSTANCE_PREFIX=e2e-test-peterhornyack ./run-e2e.sh --node-os-distro=windows --ginkgo.focus="override\sSecurityContext"`

At the end of the run, the log-dump results and output with this PR were successful and essentially identical to the output without the PR.

Finally, to test log_dump_custom_get_instances I set this function before running the above command:

`function log_dump_custom_get_instances() {   if [[ $1 == "master" ]]; then return 0; fi;   gcloud compute instances list '--filter=(metadata.created-by ~ zones/us-central1-b/instanceGroupManagers/e2e-test-peterhornyack-minion-group)' '--format=get(name)'; }; export -f log_dump_custom_get_instances`

The log-dump results were also successful here, and only the logs for the custom instances specified in that function (only the Linux nodes, not the Windows nodes) were gathered.

pjh:
/unhold

This is ready for review, thanks!

pjh:
@tosi3k @mborsz @mm4tt @wojtek-t could someone please take a look? Thanks.

tosi3k:
I don't think I have sufficient ownership of this but the PR looks good to me.

/assign @mborsz

PR #24760, opened 2021-12-30
Dump more node info in log-dump.sh

pjh:
Causes log-dump to capture some node information even when not fetching full logs from every node. The information includes all node names and the nodes' kubernetes status.

For GCE/GKE we run some scalability tests that bring up a cluster with hundreds of nodes. Sometimes just a handful of those nodes fail to start, but the log-dump output that we get right now doesn't indicate which nodes failed because we don't gather logs individually for every single node. The summary nodes information that this PR adds should help for troubleshooting this sort of issue.

Depends on https://github.com/kubernetes/test-infra/pull/24759.

### DIFF
--- a/logexporter/cluster/log-dump.sh
+++ b/logexporter/cluster/log-dump.sh
@@ -809,6 +809,23 @@ function dump_nodes_with_logexporter() {
   fi
 }
 
+# Writes node information that's available through the gcloud and kubectl API
+# surfaces to a nodes/ subdirectory of $report_dir.
+function dump_node_info() {
+  nodes_dir="${report_dir}/nodes"
+  mkdir -p "${nodes_dir}"
+
+  detect-node-names
+  if [[ -n "${NODE_NAMES:-}" ]]; then
+    printf "%s\n" "${NODE_NAMES[@]}" > "${nodes_dir}/node_names.txt"
+  fi
+  if [[ -n "${WINDOWS_NODE_NAMES:-}" ]]; then
+    printf "%s\n" "${WINDOWS_NODE_NAMES[@]}" > "${nodes_dir}/windows_node_names.txt"
+  fi
+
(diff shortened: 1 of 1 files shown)


pjh:
cc @ibabou

pjh:
I'm trying to e2e test this by following the [gce windows e2e test steps](https://github.com/kubernetes/kubernetes/blob/master/cluster/gce/windows/README-GCE-Windows-kube-up.md#running-e2e-tests-against-the-cluster) for a single test, after copying the modified test-infra `log-dump.sh` file into my `kubernetes/cluster/log-dump/` directory. I believe that I can override the use of the test-infra `log-dump.sh` and use `cluster/log-dump/log-dump.sh` by [setting an environment variable](https://github.com/kubernetes/test-infra/pull/22292/files):

`USE_TEST_INFRA_LOG_DUMPING=false ./run-e2e.sh --node-os-distro=windows --ginkgo.focus="provide\sDNS\sfor\sthe\scluster"`

We'll see how it goes.

pjh:
Ok, see my [comment on the first PR](https://github.com/kubernetes/test-infra/pull/24759#issuecomment-1003169041) for full details on how I e2e tested this. The log-dump script now outputs these files in the artifacts directory:

`nodes/kubectl_get_nodes.yaml`
`nodes/node_names.txt`

pjh:
This is ready for review!

/hold still until https://github.com/kubernetes/test-infra/pull/24759 is merged, then I'll update this PR to be just the single commit.

pjh:
@tosi3k @mborsz @mm4tt @wojtek-t could someone please take a look? Thanks.

tosi3k:
This also looks good to me. Please remove the parent commit as it's already ingested in https://github.com/kubernetes/test-infra/pull/24759, however.

/assign @mborsz

pjh:
Rebased onto head now that #24759 is merged.
/unhold

Thanks @mborsz for taking a look!

PR #24770, opened 2022-01-04
Crier gerrit reporter: truncate when there are too many jobs

chaodaiG:
Gerrit by default limits to 5000 chars per comment, this could be reached when there are many prow jobs reporting.

PR #24778, opened 2022-01-05
Crier gerrit reporter: ensure reports all jobs

chaodaiG:
Gerrit adaptor depends on comments posted by crier gerrit reporter to figure out which prow jobs to be triggered with /retest. #24770 truncates gerrit reports when there are too many prow jobs, resulted in not all prow jobs are reported in the comments. So when a user comments /retest prow only triggers the jobs that were included in the report, instead of all prow jobs that failed.

### DIFF
--- a/prow/crier/reporters/gerrit/reporter.go
+++ b/prow/crier/reporters/gerrit/reporter.go
@@ -42,8 +42,10 @@ const (
 	hourglass  = "⏳"
 	prohibited = "🚫"
 
-	defaultProwHeader = "Prow Status:"
-	jobReportFormat   = "%s %s %s - %s\n"
+	defaultProwHeader            = "Prow Status:"
+	defaultProwjobsDetailsHeader = "Prow Jobs Details:"
+	jobReportFormat              = "%s %s %s - %s\n"
+	jobReportFormatWithoutURL    = "%s %s %s\n"
 
 	// lgtm means all presubmits passed, but need someone else to approve before merge (looks good to me).
 	lgtm = "+1"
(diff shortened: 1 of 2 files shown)


chaodaiG:
cc @2nd-half

chaodaiG:
This PR changes from a single gerrit comment to 2 comments:
- One vote on gerrit labels, comment doesn't contain URL, e.g. `jobA succeed`
- The other is pure comment with URL, e.g. `jobA succeed - https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/batch/pull-kubernetes-unit/1478569178959974400`

The first comment is short and helps ensure that all jobs are reported in the comment, which can then be used by gerrit adaptor to decide which jobs are failed. The second comment is helpful for helping debugging.

Theoretically these two comments could be combined into a single comment, which would work for most cases when there are not a lot of presubmit tests. However for repos that contain 10 - 20 presubmit tests, adding more contents(the lines without URLs) might throw their comments over the size limit. On the other hand it probably won't look good

chaodaiG:
/hold
This PR is based on top of #24770, the comment size limit was hard coded as 4000, which is actually configurable for each gerrit host. Will need to put more thoughts on it before proceeding

prow/crier/reporters/gerrit/reporter.go
```diff
@@ -336,19 +338,26 @@ func (c *Client) Report(ctx context.Context, logger *logrus.Entry, pj *v1.ProwJo
 
 	logger.Infof("Reporting to instance %s on id %s with message %s", gerritInstance, gerritID, message)
 	if err := c.gc.SetReview(gerritInstance, gerritID, gerritRevision, message, reviewLabels); err != nil {
-		logger.WithError(err).Infof("fail to set review with label %q on change ID %s", reportLabel, gerritID)
+		logger.WithError(err).WithField("gerrit_id", gerritID).WithField("label", reportLabel).Info("Fail to set review.")
```
fejta, line 341: "Failed to set review" sounds better than "Fail" (the failure already happened)

chaodaiG:
close in favor of https://github.com/kubernetes/test-infra/pull/24789

/close

PR #24783, opened 2022-01-05
Add arm64/v8 support for krte image

ruquanzhao:
How about we add multi-arch support of krte?

Signed-off-by: Ruquan Zhao ruquan.zhao@arm.com

### DIFF
--- a/images/krte/Dockerfile
+++ b/images/krte/Dockerfile
@@ -49,6 +49,7 @@ COPY wrapper.sh /usr/local/bin/
 # We're already inside docker though so we can be sure these are already mounted.
 # Trying to remount these makes for a very noisy error block in the beginning of
 # the pod logs, so we just comment out the call to it... :shrug:
+ARG TARGETARCH
 RUN echo "Installing Packages ..." \
         && apt-get update \
         && apt-get install -y --no-install-recommends \
@@ -72,7 +73,7 @@ RUN echo "Installing Packages ..." \
             unzip \
         && rm -rf /var/lib/apt/lists/* \
     && echo "Installing Go ..." \
-        && export GO_TARBALL="go${GO_VERSION}.linux-amd64.tar.gz"\
+        && export GO_TARBALL="go${GO_VERSION}.linux-${TARGETARCH:-amd64}.tar.gz"\
         && curl -fsSL "https://storage.googleapis.com/golang/${GO_TARBALL}" --output "${GO_TARBALL}" \
         && tar xzf "${GO_TARBALL}" -C /usr/local \
         && rm "${GO_TARBALL}"\
@@ -86,11 +87,12 @@ RUN echo "Installing Packages ..." \
             --bash-completion=false \
             --path-update=false \
             --usage-reporting=false \
-        && gcloud components install kubectl \
+        && curl -LO "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/${TARGETARCH:-amd64}/kubectl" \
+        && chmod +x ./kubectl && mv ./kubectl /usr/local/bin/kubectl \
     && echo "Installing Docker ..." \
         && curl -fsSL https://download.docker.com/linux/$(. /etc/os-release; echo "$ID")/gpg | apt-key add - \
         && add-apt-repository \
-            "deb [arch=amd64] https://download.docker.com/linux/$(. /etc/os-release; echo "$ID") \
+            "deb [arch=${TARGETARCH:-amd64}] https://download.docker.com/linux/$(. /etc/os-release; echo "$ID") \
             $(lsb_release -cs) stable" \
         && apt-get update \
         && apt-get install -y --no-install-recommends docker-ce \
--- a/images/krte/cloudbuild.yaml
+++ b/images/krte/cloudbuild.yaml
@@ -1,9 +1,11 @@
 timeout: 1800s
 steps:
-  - name: gcr.io/cloud-builders/docker
+  - name: gcr.io/k8s-testimages/gcb-docker-gcloud
+    entrypoint: /buildx-entrypoint
     args:
     - build
     - --tag=gcr.io/$PROJECT_ID/krte:$_GIT_TAG-$_CONFIG
+    - --platform=linux/amd64,linux/arm64/v8
     - --build-arg=GO_VERSION=$_GO_VERSION
     - --build-arg=K8S_RELEASE=$_K8S_RELEASE
     - --build-arg=BAZEL_VERSION_ARG=$_BAZEL_VERSION
@@ -11,11 +13,15 @@ steps:
     - --build-arg=CFSSL_VERSION=$_CFSSL_VERSION
     - --build-arg=UPGRADE_DOCKER_ARG=$_UPGRADE_DOCKER
     - --build-arg=IMAGE_ARG=gcr.io/$PROJECT_ID/krte:$_GIT_TAG-$_CONFIG
+    - --push
     - .
     dir: images/krte
-  - name: gcr.io/cloud-builders/docker
+  - name: gcr.io/k8s-testimages/gcb-docker-gcloud
+    entrypoint: gcloud
     args:
-    - tag
+    - container
+    - images
+    - add-tag
     - gcr.io/$PROJECT_ID/krte:$_GIT_TAG-$_CONFIG
     - gcr.io/$PROJECT_ID/krte:latest-$_CONFIG
     dir: images/krte
@@ -30,6 +36,3 @@ substitutions:
   _CFSSL_VERSION: R1.2
 options:
   substitution_option: ALLOW_LOOSE
-images:
-  - 'gcr.io/$PROJECT_ID/krte:$_GIT_TAG-$_CONFIG'
-  - 'gcr.io/$PROJECT_ID/krte:latest-$_CONFIG'


ruquanzhao:
cc @chendave

images/krte/cloudbuild.yaml
```diff
@@ -1,23 +1,29 @@
 timeout: 1800s
 steps:
-  - name: gcr.io/cloud-builders/docker
+  - name: gcr.io/k8s-testimages/gcb-docker-gcloud
```
chendave, line 3: Is this the new way to build the multi-arch image?

chendave (commented):
Have your verified with gcb?

chendave:
There is a serial of patches to make the test-infra/images support for multi-arch,  see the relevant change of this issue - https://github.com/kubernetes/test-infra/issues/16588.

And per our evaluation, this krte image is one of the bits that are not ignorable.

Might you take a look at this and let us know your thoughts? 

@aojea @BenTheElder @spiffxp

images/krte/variants.yaml
```diff
@@ -1 +0,0 @@
-../kubekins-e2e/variants.yaml
```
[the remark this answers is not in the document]
ruquanzhao: I noticed the bootstrap image is DEPRECATED, and kubekins-e2e based on bootstrap.
So, I thought it's a better way to maintain two variants.yaml.

ruquanzhao:
> Have your verified with gcb?

Yes. And here is the log of arm64. 

```
Step #0: #9 DONE 226.3s
Finished Step #0
Starting Step #1
Step #1: Already have image (with digest): gcr.io/k8s-testimages/gcb-docker-gcloud
Step #1: Created [asia-southeast1-docker.pkg.dev/arm-nwcs-ci-dev/arm64-k8s-ci/krte:latest-arm64].
Finished Step #1
PUSH
DONE
```

Full build log in https://pastebin.com/yvD1TGqL

images/krte/variants.yaml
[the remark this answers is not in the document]
BenTheElder: this is reverting a previous intentional change (which admittedly I didn't like), but is completely orthogonal to this PR. this change, if merged, should be a different PR.

I don't think arm64 should be a variant anyhow. we should just make the tags multi-arch if we're going to do this.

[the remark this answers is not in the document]
BenTheElder: using bootstrap directly is deprecated, but kubekins-e2e is the majority of kubernetes CI.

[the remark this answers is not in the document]
chendave: @BenTheElder thanks for the comments! let's see if we can make this change more clear to follow your suggestion!

images/krte/cloudbuild.yaml
```diff
@@ -1,23 +1,29 @@
...
-    - --tag=gcr.io/$PROJECT_ID/krte:$_GIT_TAG-$_CONFIG
-    - --build-arg=GO_VERSION=$_GO_VERSION
-    - --build-arg=K8S_RELEASE=$_K8S_RELEASE
-    - --build-arg=BAZEL_VERSION_ARG=$_BAZEL_VERSION
-    - --build-arg=OLD_BAZEL_VERSION=$_OLD_BAZEL_VERSION
-    - --build-arg=CFSSL_VERSION=$_CFSSL_VERSION
-    - --build-arg=UPGRADE_DOCKER_ARG=$_UPGRADE_DOCKER
-    - --build-arg=IMAGE_ARG=gcr.io/$PROJECT_ID/krte:$_GIT_TAG-$_CONFIG
-    - .
+      - build
```
chendave, line 6: Is the indent here and below necessary?

ruquanzhao: Thanks for your review!
It's not necessay, I have removed it.

images/krte/cloudbuild.yaml
```diff
@@ -30,6 +36,3 @@ substitutions:
   _CFSSL_VERSION: R1.2
 options:
   substitution_option: ALLOW_LOOSE
-images:
-  - 'gcr.io/$PROJECT_ID/krte:$_GIT_TAG-$_CONFIG'
-  - 'gcr.io/$PROJECT_ID/krte:latest-$_CONFIG'
```
chendave, line 35: why those lines are removed?

ruquanzhao: refer to #22444

ruquanzhao:
Dependence of `variants.yaml` is removed.
Would you mind taking a look?
/cc @BenTheElder

images/krte/Dockerfile
```diff
@@ -72,7 +73,7 @@ RUN echo "Installing Packages ..." \
             unzip \
         && rm -rf /var/lib/apt/lists/* \
     && echo "Installing Go ..." \
-        && export GO_TARBALL="go${GO_VERSION}.linux-amd64.tar.gz"\
+        && export GO_TARBALL="go${GO_VERSION}.linux-${TARGETARCH:-amd64}.tar.gz"\
```
BenTheElder, line 76: nit: can we just default this once with ARG TARGETARCH=amd64 and simplify here and below?

images/krte/cloudbuild.yaml
```diff
@@ -1,21 +1,27 @@
 timeout: 1800s
 steps:
-  - name: gcr.io/cloud-builders/docker
+  - name: gcr.io/k8s-testimages/gcb-docker-gcloud
+    entrypoint: /buildx-entrypoint
     args:
     - build
     - --tag=gcr.io/$PROJECT_ID/krte:$_GIT_TAG-$_CONFIG
+    - --platform=linux/amd64,linux/arm64/v8
```
BenTheElder, line 8: possibly something we should parameterize

BenTheElder:
> How about we add multi-arch support of krte?

I'm still missing the _why_, I mean we could build for arbitrarily many architectures but building them is not free, and this image is used to test kind on Kubernetes's CI, and isn't supported for other users. What's the use case?

I don't want to add build time and complexity and pay to store more image contents without a concrete use case, most of these CI images are amd64 because SIG k8s infra provides amd64 clusters to host CI and these images are just used as the project's CI environment.

See also the image's readme: https://github.com/kubernetes/test-infra/blob/master/images/krte/README.md#warning

> There is a serial of patches to make the test-infra/images support for multi-arch, see the relevant change of this issue - #16588.

OK, I see that these PRs exist, but in https://github.com/kubernetes/test-infra/issues/16588 the conversation settled that while redhat needed the core prow components (the "pod utilities") to be multi-arch, there was no decision regarding the rest.

My experience so far has been that making docker images multi-arch "just because" causes CI flakes and much slower builds due to the penalty of building under qemu, and costs far more energy than the initial effort put in to port it with essentially no reward for the upstream project.

In the main kubernetes repo we are not accepting new architectures without a KEP.

Is there a plan to run multi-arch CI within the Kubernetes project? Or some other reason SIG testing should continue to port everything?

chendave:
Since @ruquanzhao  is on vacation, so pls let me try to reply, just for something I heard from his early investigation, might not correct though. 

> I'm still missing the why, I mean we could build for arbitrarily many architectures but building them is not free, and this image is used to test kind on Kubernetes's CI, and isn't supported for other users. What's the use case?

Based on what @ruquanzhao told me, we need this to provide a containerized environment for test execution.  So, why we need this change?

1) As ARM server / instance is not rare on the market,  upstream project like kubevirt and kubeedge etc. have ARM CI ready beside the X86_64, the project like kubevirt are also using prow framework for testing. Currently, the ARM CI is running periodically on baremetal in the kubevirt due to the lack of support from the prow and the utility like *krte*. 

Whether we need ARM CI coverage for k8s is another topic and it's totally community's call but we can make this utility work-able for mulit-arch so that they are not struggle with containerized environment.

2) We are maintaining an internal CI infrastructure to track is there any issue in the upstream, so we can fix them timely. We hope those utilities are also supported on arm64 in the upstream, so that we are more aligned with upstream.

> In the main kubernetes repo we are not accepting new architectures without a KEP.

Sure, we can provide a KEP if this is needed, we might need to have a consensus on how far we could go.


> Is there a plan to run multi-arch CI within the Kubernetes project? 

Of course, we'd love to. 
Might start from minimal coverage (conformance?) and make it just as an experimental reference? This will benefit the cloud provider like AWS and the provider who has the plan to provide the ARM instance.
And if take it as an experiment, we can provide bare-metals.

Again, it's community's call to decide where we should go.

@ruquanzhao @BenTheElder  correct me?

BenTheElder:
We are plagued by timeouts in the image build today from a very lengthy build in these images as-is supporting all the necessary Kubernetes environments in CI, such as [this currently most recent build](https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-test-infra-push-kubekins-e2e/1485636516867739648).

> As ARM server / instance is not rare on the market, upstream project like kubevirt and kubeedge etc. have ARM CI ready beside the X86_64, the project like kubevirt are also using prow framework for testing. Currently, the ARM CI is running periodically on baremetal in the kubevirt due to the lack of support from the prow and the utility like krte.

I don't see how these are related, the KRTE image is the environment used for [prow.k8s.io](https://prow.k8s.io) to test KIND (again, please read the [README](https://github.com/kubernetes/test-infra/blob/ae487b7658048076ea6cedee6408421d6b642f30/images/krte/README.md) for KRTE).

Kubernetes can be tested on any architecture without this image. And there is no ARM CI in the Kubernetes project, certainly not in the KIND project, for which this image is supported only.

> Whether we need ARM CI coverage for k8s is another topic and it's totally community's call but we can make this utility work-able for mulit-arch so that they are not struggle with containerized environment.

Again, this image is only supported for the purposes of testing KIND in the project's CI, we do not have the maintainer bandwidth to support other things with this, the KIND project is already overloaded as-is. So it is not another topic. This image is an implementation detail of the KIND project's CI and is expressly not supported for other purposes.

> # krte image
> 
> krte - [KIND](https://sigs.k8s.io/kind) RunTime Environment
> 
> This image contains things we need to run kind in Kubernetes CI, and
> is maintained for the sole purpose of testing Kubernetes with KIND.
> 
> ## WARNING
> 
> This image is _not_ supported for other use cases. Use at your own risk.

https://github.com/kubernetes/test-infra/blob/ae487b7658048076ea6cedee6408421d6b642f30/images/krte/README.md

chendave:
@BenTheElder thanks for the detailed explanation!

I was thinking if we need to test kubernetes with KIND on ARM, we must need this, maybe I am wrong.

@ruquanzhao  could you pls take a look at @BenTheElder 's comments when you back? thanks!

ruquanzhao:
Just got back from a long vacation.

Thank you all for the comments and reviews! @BenTheElder @aojea @chendave 


> Again, this image is only supported for the purposes of testing KIND in the project's CI, we do not have the maintainer bandwidth to support other things with this, the KIND project is already overloaded as-is. So it is not another topic. This image is an implementation detail of the KIND project's CI and is expressly not supported for other purposes.

We(@chendave and I) are deploying k8s CI with Prow on arm locally and trying to align with upstream, preparing to make it public in the future if possible. We found that `prow.k8s.io` uses `kubekins-e2e` and `krte` for most k8s jobs. So I am wondering if we could add arm64 support to `kubekins-e2e` and `krte.` 



> Kubernetes can be tested on any architecture without this image.

Yes, indeed. Kubernetes can be tested on any architecture without `krte.` We need this because we want to test k8s with KIND, or else we must do it on bare-metal. Since we are using Prow, working with containers is a more elegant and maintainable way.

> I'm with Ben on this, if people wants to add a CI with kind with ARM I'm happy to support them, answer questions, .... , the image can be copied, and new jobs can be added in parallel, but people has to maintain their own CI,not at the cost of putting more load on us. This can also be reviewed in the future, and if there is more alignment, it is just a matter of merging things ...

We have dedicated resources to work or maintain such things if ARM CI is acceptable for KIND or K8S, so this is not a problem anymore. Thanks again for your comments! @aojia


Hope to hear your opinion! @BenTheElder

BenTheElder:
Please reach out to SIG k8s infra about a plan for community arm testing resources and staffing maintaining them if that approach is to be taken.

I don't think the community will need or want an additional prow deployment to accomplish testing, so effort to "align" is relatively wasted with respect to being part of community CI in the future. Prow can readily schedule to any kubernetes cluster from the same services cluster.

At this point in time the Kubernetes project and by extension the kind project do not have any non amd64 CI nor any plans to provide any given lack of funding.

This image is maintained for our own upstream testing purposes as documented and is not intended for reuse elsewhere. I do not think the community should pay the cost to maintain and host this for external CI, Antonio and I are stretched extremely thin as-is. It is not a blocker to run downstream kind tests using some other image, and we make no stability guarantees about it's reuse.

This image is an internal detail of the kind project, which is fully on amd64 provided by sig k8s infra.

PR #24785, opened 2022-01-05
Gerrit comment size limit corrected to 16k instead of 5000

chaodaiG:
It was misread initially, 5000 was the limit of counts of comments, 16k is the max size of comment


chaodaiG:
cc @2nd-half

alvaroaleman (approved):

PR #24791, opened 2022-01-06
Gerrit reporter: sort results also consider error state

chaodaiG:
It makes more sense to display jobs failed with error state over succeeded instead of mixed in


chaodaiG:
cc @2nd-half

cjwagner (approved):

PR #24803, opened 2022-01-06
Overrides value of server URL for CAPV prow jobs

srm09:
This patch overrides the server URL used in the CI E2E tests for CAPV with a private IP address. Since the IP address requires a special VPN connection, plaintext works as well.
Eventually, this will get replaced by the way of an ExternalSecret.


yastij (commented):
/approve

PR #24804, opened 2022-01-07
Switch cgroupv2 COS jobs to cos-dev image family

bobbypage:
Switch to use cos-dev image family to start testing against
cos-dev-97-16778-0-0. cos-dev-97-16778-0-0 is first COS release that
enables cgroupv2 by default.

image

Signed-off-by: David Porter <porterdavid@google.com>

### DIFF
--- a/config/jobs/kubernetes/sig-node/containerd.yaml
+++ b/config/jobs/kubernetes/sig-node/containerd.yaml
@@ -1131,9 +1131,8 @@ periodics:
       - --gcp-nodes=4
       - --gcp-zone=us-west1-b
       - --ginkgo-parallel=30
-      # Using cos-93-lts to pick up 5.10 kernel.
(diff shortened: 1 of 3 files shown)


Dragoncell:
Shall we change the image on this yaml file also ? https://github.com/kubernetes/test-infra/blob/master/jobs/e2e_node/containerd/containerd-master/cgroupv2/image-config-cgroupv2.yaml

bobbypage:
> Shall we change the image on this yaml file also ? https://github.com/kubernetes/test-infra/blob/master/jobs/e2e_node/containerd/containerd-master/cgroupv2/image-config-cgroupv2.yaml

Good catch, also updated that image config

Dragoncell:
Thanks looks good to me. /lgtm

SergeyKanzhelev (approved):
/lgtm
/approve

let's watch it closely and if it will start being flaky because of COS instability, revert back

PR #24806, opened 2022-01-07
Fix gerrit reporter log to print names instead of pointers.

cjwagner:
/assign @mpherman2 @listx

### DIFF
--- a/prow/crier/reporters/gerrit/reporter.go
+++ b/prow/crier/reporters/gerrit/reporter.go
@@ -349,8 +349,16 @@ func (c *Client) Report(ctx context.Context, logger *logrus.Entry, pj *v1.ProwJo
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


prow/crier/reporters/gerrit/reporter.go
```diff
@@ -349,10 +349,18 @@ func (c *Client) Report(ctx context.Context, logger *logrus.Entry, pj *v1.ProwJo
 		}
 	}
 
-	logger.Infof("Review Complete, reported jobs: %v", toReportJobs)
+	logger.Infof("Review Complete, reported jobs: %s", jobNames(toReportJobs))
 	return toReportJobs, nil, nil
 }
 
+func jobNames(jobs []*v1.ProwJob) []string {
```
chaodaiG, line 356: would be nice to add a unit test

PR #24811, opened 2022-01-07
Run kettle test with docker instead of bazel

chaodaiG:


chaodaiG:
/cc @BenTheElder  @MushuEE 

This is an example for how python tests will run after the bazel deprecation, will do this for other python packages once we agreed on the approach

kettle/make_db_test.py
```diff
@@ -159,9 +159,10 @@ def assert_main_output(self, threads, expected=None, db=None,
         self.assertEqual(result, expected)
         return db
 
-    def test_clean(self):
-        for threads in [1, 32]:
-            self.assert_main_output(threads)
+    # def test_clean(self):
```
chaodaiG, line 162: Somehow this test failed on my local even with `bazel test //kettle:make_db_test`, have talked with @MushuEE and they will work on a fix

requirements3.txt
```diff
@@ -1,6 +1,7 @@
 astroid==2.3.3
 backports.functools_lru_cache==1.6.1
 configparser==4.0.2
+chardet==4.0.0
```
chaodaiG, line 4: This was expressed as of https://github.com/kubernetes/test-infra/blob/8b71bc27720d9818c315c02e9abdc1d7b58bdcc9/kettle/BUILD.bazel#L14, but never defined anywhere in this repo, `pip3 install -r requirements3.txt` doesn't install it as a transitive dependency either. So explicitly define it here. The other dependencies under https://github.com/kubernetes/test-infra/blob/8b71bc27720d9818c315c02e9abdc1d7b58bdcc9/kettle/BUILD.bazel#L14 are all transitive dependencies so we won't need to worry about them

kettle/Makefile
```diff
@@ -44,5 +44,21 @@ update-staging: get-cluster-credentials
 rollback: get-cluster-credentials
 	kubectl rollout undo deployments/kettle
 
+ensure-requirements3:
+	../hack/run-in-python-container.sh pip3 install -r requirements3.txt
```
BenTheElder, line 48: is this cheap to do repeatedly or should we follow-up with making it depend on the requirements3.txt file / outputting a dummy marker so it can be skipped when requirements are unchanged?

kettle/make_db_test.py
BenTheElder: can we add a TODO comment so it's greppable?

BenTheElder (approved):
/lgtm
/approve
/hold

kettle/Makefile
chaodaiG: yeah it's pretty cheap, and for this case makefile is smart enough to only do it once

BenTheElder: well, once *per `make` invocation*, but every invocation will do this again?

chaodaiG: yeah that's true, and I would say it's very fast once all requirements were met

kettle/make_db_test.py
chaodaiG: done

PR #24815, opened 2022-01-07
CAPZ: use 1.23 image with go 1.17 for main and v1beta1 tests

CecileRobertMichon:
In order to move to go 1.17 and k8s client 1.23 in the next release, we need to move the main and v1beta1 jobs to the 1.23 image.

/assign @mboersma @jsturtevant @shysank @devigned


mboersma (approved):
lgtm

PR #24816, opened 2022-01-07
Use locked time to avoid any time.time race

MushuEE:
Seeing some possible failures likely due to time.Now() race.

This should lock everything and since time is referenced off the MockClient there is no need to use `freezegun`

/area kettle

### DIFF
--- a/kettle/make_db_test.py
+++ b/kettle/make_db_test.py
@@ -16,14 +16,13 @@
 
 """Tests for make_db."""
 
-import time
 import sys
 import unittest
 
 import make_db
 import model
 
-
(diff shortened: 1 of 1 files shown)


chaodaiG (approved):
Awesome! Thanks a lot!

chaodaiG:
while fixing test please also uncomment line 162 to re-enable the test

chaodaiG:
> while fixing test please also uncomment line 162 to re-enable the test

@MushuEE please also uncomment these

MushuEE:
@chaodaiG done

chaodaiG (approved):

PR #24817, opened 2022-01-07
Revert "CAPZ: use 1.23 image with go 1.17 for main and v1beta1 tests"

CecileRobertMichon:
This reverts commit 9657e71b2d884a928587407d8b129e9f7e61c394.

https://kubernetes.slack.com/archives/CEX9HENG7/p1641591933009800

/hold

this is plan B in case https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1956 doesn't fix the issue so we can unblock the PR tests.


CecileRobertMichon:
/close 

looks like we won't be needing this

PR #24820, opened 2022-01-07
Add fuzz test for parse config

mpherman2:
/assign @chaodaiG @cjwagner


prow/config/config_fuzz.go
```diff
@@ -0,0 +1,26 @@
...
+limitations under the License.
+*/
+
+package config
+
+import fuzz "github.com/google/gofuzz"
+
+func FuzzParseProwConfig(data []byte) int {
+	var c Config
+	fuzz.NewFromGoFuzz(data).Fuzz(&c)
```
chaodaiG, line 23: random value in prow config? have you tried it?

mpherman2: My understanding is that this generates a valid Prow config with random inputs for each field. I haven't tried it yet. test-infra fuzz job is not doing anything important rn so I thought it would be easiest to test in prod

stevekuznetsov:
@alvaroaleman knows a lot about properly fuzz testing - might want to take a look :)

prow/config/config_fuzz.go
```diff
@@ -0,0 +1,26 @@
...
+	parseProwConfig(&c)
+	return 0
```
alvaroaleman, line 25: No Cisms, please :)

prow/config/config_fuzz.go
```diff
@@ -0,0 +1,26 @@
...
+*/
+
+package config
+
+import fuzz "github.com/google/gofuzz"
+
+func FuzzParseProwConfig(data []byte) int {
+	var c Config
+	fuzz.NewFromGoFuzz(data).Fuzz(&c)
+	parseProwConfig(&c)
```
alvaroaleman, line 24: Why no error checking?

prow/config/config_fuzz.go
alvaroaleman: Why is this in the production code and unused and not a test? And what is it supposed to test, that we don't panic?

mpherman2: I'm experimenting with [ClusterfuzzLite](https://google.github.io/clusterfuzzlite/) to do some continuous fuzz testing. I am not very experienced with writing fuzz tests so this is just my first attempt. My understanding is that It should test that it does not panic, oom, index out of bounds etc.

alvaroaleman: Okay so the only way this might fail is if we panic? If so, maybe do a
```
FuzzParseProwConfig(data []byte) int} (err error) {
    defer func() {
         if r := recover(); r != nil {
            err = fmt.Errorf("found a panic: %v", r)
         }
     }
}
```

We btw have a bunch of fuzztests already that run as part of the normal `go test`: https://grep.app/search?q=gofuzz%22&filter[repo][0]=kubernetes/test-infra

mpherman2:
Closing as this fuzz test does not work the way I expected it to. I appreciate all of the input, it will be helpful next time I attempt to write a fuzz test :)

PR #24825, opened 2022-01-10
Add new metrics to prow plugins lgtm and approve

jmguzik:
This commit adds durations, successes, and failures metrics to measure transactions
in plugins: lgtm, approve. Some transactions take an exceptionally long time in
our installation. This change could confirm our findings and enable the possibility
to act acordingly.

### DIFF
--- a/prow/plugins/plugins.go
+++ b/prow/plugins/plugins.go
@@ -234,6 +234,24 @@ func (a *Agent) CommentPruner() (*commentpruner.EventClient, error) {
 	return a.commentPruner, nil
 }
 
+// GatherHandlerMetrics executes handler and gathers results & durations metrics.
+func (a *Agent) GatherHandlerMetrics(handler func() error, plugin, method string) error {
+	start := time.Now()
+	err := handler()
+	stop := time.Now()
+
+	promLabels := prometheus.Labels(map[string]string{"plugin": plugin, "method": method})
+	a.Metrics.HandlerDurations.With(promLabels).Observe(float64(stop.Sub(start).Seconds()))
+
+	resultString := "success"
+	if err != nil {
+		resultString = "error"
+	}
+	a.Metrics.HandlerResults.WithLabelValues(plugin, method, resultString).Inc()
+
+	return err
+}
+
 // ClientAgent contains the various clients that are attached to the Agent.
 type ClientAgent struct {
 	GitHubClient              github.Client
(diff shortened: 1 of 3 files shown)


prow/plugins/plugins.go
```diff
@@ -543,6 +561,24 @@ var configMapSizeGauges = prometheus.NewGaugeVec(prometheus.GaugeOpts{
 	Help: "Size of data fields in ConfigMaps updated automatically by Prow in bytes.",
 }, []string{"name", "namespace"})
 
+var handlerResults = prometheus.NewCounterVec(prometheus.CounterOpts{
+	Name: "plugin_handler_results",
+	Help: "Count of plugin handler results by handler method and result (success/error).",
+}, []string{
+	"plugin",
```
jmguzik, line 568: Probably `org` and `repo` would be welcomed here?

prow/plugins/plugins.go
```diff
@@ -543,6 +561,24 @@ var configMapSizeGauges = prometheus.NewGaugeVec(prometheus.GaugeOpts{
...
+	"method",
+	"result",
+})
+
+var handlerDurations = prometheus.NewHistogramVec(prometheus.HistogramOpts{
+	Name:    "plugin_handler_durations",
+	Help:    "Plugin handler durations",
```
petr-muller, line 575: Are these seconds? We should mention the units at least in `Help`

petr-muller (commented):
I like the simplicity and the reusability.

I am worried about the fact that the metrics are tied to implementation details (event handlers) rather than the semantic "transactions" that the plugins implement, which will IMO skew the metrics and make them less useful. I'll try to illustrate on the `lgtm` plugin:

The plugin has a handler for a generic comment event, so the metrics would record the result and duration of the handler, for *all* comments everywhere, including the comments that the plugin does not actually care about (comments on issues, closed PRs, or comments that do not contain the `/lgtm` command). From the operational perspective, the handler is only worth measuring/recording if it contains at least one important action or decision. If our metrics were as generic as implemented here, they would be heavilly skewed by the presence of very fast successful handlers for irrelevant comments, which would not help you that much with monitoring how much the plugin is impeded by e.g. GH API errors.

Could we find a way to only count events that are at least minimally useful (do at least some decision, reach to at least one external system... etc).

alvaroaleman:
There are already metrics for duration and error rate by event_type, action and plugin in hook: https://github.com/kubernetes/test-infra/blob/0595b59c63b16e0a713d83cb486144032fc41f05/prow/githubeventserver/metrics.go#L35-L44

jmguzik:
I will open a new PR in the future with an implementation that follows @alvaroaleman remarks, once #24832 gets sorted.

PR #24829, opened 2022-01-10
CSI Proxy presubmit integration tests

mauriciopoppe:
Run the CSI Proxy integration tests with a Prow job, after this is stable I'll run it with `always_run: true`.

/cc @jingxu97 @ddebroy 
/sig storage

### DIFF
--- a/config/jobs/kubernetes-csi/csi-proxy/csi-proxy-config.yaml
+++ b/config/jobs/kubernetes-csi/csi-proxy/csi-proxy-config.yaml
@@ -32,3 +32,51 @@ presubmits:
         resources:
           requests:
             cpu: 2000m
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes-csi/csi-proxy/csi-proxy-config.yaml
```diff
@@ -32,3 +32,51 @@ presubmits:
         resources:
           requests:
             cpu: 2000m
+  - name: pull-kubernetes-csi-csi-proxy-integration
```
pohly, line 35: Please remember that csi-proxy-config.yaml is generated by gen-jobs.sh. Changes like this will get overwritten by future updates of all jobs and need to be in a separate file.

The PR that run into this is fixing it.

Issue #23830, opened 2021-09-29
Build and test Python without bazel


chaodaiG:
**What would you like to be added**:

Build and test Python without bazel, the related bazel rules are:

- py_binary
- py_test

Things to be considered:

- This will need to co-exist with current bazel rules for a short while, delete Python test targets from bazel once this confirmed working
- Test driven by make rule and bash scripts
- Ensure hermetic testing, including fixed tools version, like python, python3, dependencies
- Test results are in junit results format

**Why is this needed**:

This is part of #23796 

**Recommended testing approach**:
- Make test invokable by `make test`
- Create a PullRequest with change, manually run job by commenting `/test pull-test-infra-test-beta`
- [Optional]: if there is any tool that needs to exist in test image, update image at https://github.com/kubernetes/test-infra/blob/6e0d21ee50d690c8f78239257f588021d2ed57b5/config/jobs/kubernetes/test-infra/test-infra-presubmits.yaml#L145

/sig testing
/help

BenTheElder:
IMHO we should make this _just_ py_test / py_lint, we don't need PAR or anything like that -- we will package things in docker images or not at all.

I think we already have py_test movement?

chaodaiG:
> IMHO we should make this _just_ py_test / py_lint, we don't need PAR or anything like that -- we will package things in docker images or not at all.
> 
> I think we already have py_test movement?

yeah we already have all of the py_test and py_lint working. it should be fine to mark this as done

PR #24836, opened 2022-01-11
Python: run metrics and releng with make rules

chaodaiG:
Part of #23830 

/cc @BenTheElder


chaodaiG:
Meant to do this together with `experiment` and `hack` directory, the PR turned out to be pretty large with just metrics and releng, so creating a PR for these two first

BenTheElder (approved):
/lgtm
/approve

PR #24841, opened 2022-01-11
Crier gerrit reporter: worker can be more than 1

chaodaiG:
Initially gerrit worker is restricted to be only 1, due to the fact that gerrit worker could potentially do duplicated work if they don't understand each other. This resulted in gerrit reporter to work single-threaded, which is a performance bottleneck when there are lots of workloads to be reported.

Similar problem had happened on github reporter too, that multiple jobs trying to access the same github comment. This problem was solved by introducing a ShardedLock. This PR refactors ShardedLock so that it can be used by both gerrit and github reporter.

### DIFF
--- a/prow/cmd/crier/main.go
+++ b/prow/cmd/crier/main.go
@@ -245,7 +245,7 @@ func main() {
 	}
 
 	if o.gerritWorkers > 0 {
-		gerritReporter, err := gerritreporter.NewReporter(cfg, o.cookiefilePath, o.gerritProjects, mgr.GetCache())
+		gerritReporter, err := gerritreporter.NewReporter(cfg, o.cookiefilePath, o.gerritProjects, mgr.GetClient())
 		if err != nil {
 			logrus.WithError(err).Fatal("Error starting gerrit reporter")
 		}
--- a/prow/cmd/crier/main.go
+++ b/prow/cmd/crier/main.go
@@ -78,14 +78,6 @@ type options struct {
 }
 
 func (o *options) validate() error {
-
-	// TODO(krzyzacy): gerrit && github report are actually stateful..
-	// Need a better design to re-enable parallel reporting
-	if o.gerritWorkers > 1 {
-		logrus.Warn("gerrit reporter only supports one worker")
-		o.gerritWorkers = 1
-	}
-
 	if o.gerritWorkers+o.pubsubWorkers+o.githubWorkers+o.slackWorkers+o.gcsWorkers+o.k8sGCSWorkers+o.blobStorageWorkers+o.k8sBlobStorageWorkers <= 0 {
 		return errors.New("crier need to have at least one report worker to start")
 	}
--- a/prow/crier/reporters/criercommonlib/shardedlock.go
+++ b/prow/crier/reporters/criercommonlib/shardedlock.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2018 The Kubernetes Authors.
+Copyright 2022 The Kubernetes Authors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -38,13 +38,18 @@ func NewSimplePull(org, repo string, number int) *SimplePull {
 
 // ShardedLock contains sharding information based on PRs
 type ShardedLock struct {
+	// semaphore is chosed over mutex, as Acquire from semaphore respects
+	// context timeout while mutex doesn't
 	mapLock *semaphore.Weighted
 	locks   map[SimplePull]*semaphore.Weighted
 }
 
 // NewShardedLock creates ShardedLock
-func NewShardedLock(mapLock *semaphore.Weighted, locks map[SimplePull]*semaphore.Weighted) *ShardedLock {
-	return &ShardedLock{mapLock: mapLock, locks: locks}
+func NewShardedLock() *ShardedLock {
+	return &ShardedLock{
+		mapLock: semaphore.NewWeighted(1),
+		locks:   map[SimplePull]*semaphore.Weighted{},
+	}
 }
 
 // GetLock aquires the lock for a PR
@@ -73,6 +78,11 @@ func (s *ShardedLock) Cleanup() {
 	defer s.mapLock.Release(1)
 
 	for key, lock := range s.locks {
+		// There is a very low chance of race condition, that two threads got
+		// different locks from the same PR, which would end up with duplicated
+		// report once. Since this is very complicated to fix and the impact is
+		// really low, would just keep it as is.
+		// For details see: https://github.com/kubernetes/test-infra/pull/20343
 		lock.Acquire(ctx, 1)
 		delete(s.locks, key)
 		lock.Release(1)
--- a/prow/crier/reporters/criercommonlib/shardedlock.go
+++ b/prow/crier/reporters/criercommonlib/shardedlock.go
@@ -1,0 +1,92 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package criercommonlib contains shared lib used by reporters
+package criercommonlib
+
+import (
+	"context"
+	"time"
+
+	"github.com/sirupsen/logrus"
+	"golang.org/x/sync/semaphore"
+)
+
+// SimplePull contains info for identifying a shard
+type SimplePull struct {
+	org, repo string
+	number    int
+}
+
+// NewSimplePull creates SimplePull
+func NewSimplePull(org, repo string, number int) *SimplePull {
+	return &SimplePull{org: org, repo: repo, number: number}
+}
+
+// ShardedLock contains sharding information based on PRs
+type ShardedLock struct {
+	mapLock *semaphore.Weighted
+	locks   map[SimplePull]*semaphore.Weighted
+}
+
+// NewShardedLock creates ShardedLock
+func NewShardedLock(mapLock *semaphore.Weighted, locks map[SimplePull]*semaphore.Weighted) *ShardedLock {
+	return &ShardedLock{mapLock: mapLock, locks: locks}
+}
+
+// GetLock aquires the lock for a PR
+func (s *ShardedLock) GetLock(ctx context.Context, key SimplePull) (*semaphore.Weighted, error) {
+	if err := s.mapLock.Acquire(ctx, 1); err != nil {
+		return nil, err
+	}
+	defer s.mapLock.Release(1)
+	if _, exists := s.locks[key]; !exists {
+		s.locks[key] = semaphore.NewWeighted(1)
+	}
+	return s.locks[key], nil
+}
+
+// Cleanup deletes all locks by acquiring first
+// the mapLock and then each individual lock before
+// deleting it. The individual lock must be acquired
+// because otherwise it may be held, we delete it from
+// the map, it gets recreated and acquired and two
+// routines report in parallel for the same job.
+// Note that while this function is running, no new
+// presubmit reporting can happen, as we hold the mapLock.
+func (s *ShardedLock) Cleanup() {
+	ctx := context.Background()
+	s.mapLock.Acquire(ctx, 1)
+	defer s.mapLock.Release(1)
+
+	for key, lock := range s.locks {
+		lock.Acquire(ctx, 1)
+		delete(s.locks, key)
+		lock.Release(1)
+	}
+}
+
+// RunCleanup asynchronously runs the cleanup once per hour.
+func (s *ShardedLock) RunCleanup() {
+	go func() {
+		for range time.Tick(time.Hour) {
+			logrus.Debug("Starting to clean up presubmit locks")
+			startTime := time.Now()
+			s.Cleanup()
+			logrus.WithField("duration", time.Since(startTime).String()).Debug("Finished cleaning up presubmit locks")
+		}
+	}()
+}
--- a/prow/crier/reporters/gerrit/reporter.go
+++ b/prow/crier/reporters/gerrit/reporter.go
@@ -19,19 +19,24 @@ package gerrit
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"sort"
 	"strconv"
 	"strings"
 	"time"
 
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+
 	"github.com/andygrunwald/go-gerrit"
 	"github.com/sirupsen/logrus"
+	"golang.org/x/sync/semaphore"
 	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	v1 "k8s.io/test-infra/prow/apis/prowjobs/v1"
 	"k8s.io/test-infra/prow/config"
+	"k8s.io/test-infra/prow/crier/reporters/criercommonlib"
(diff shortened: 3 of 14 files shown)


chaodaiG:
/hold
Will need unit test for it

prow/crier/reporters/gerrit/reporter.go
```diff
@@ -253,6 +265,45 @@ func (c *Client) ShouldReport(ctx context.Context, log *logrus.Entry, pj *v1.Pro
...
+		if err != nil {
+			return nil, nil, err
+		}
+		if err := lock.Acquire(ctx, 1); err != nil {
+			return nil, nil, err
+		}
+		defer lock.Release(1)
+
+		// In the case where several prow jobs from the same PR are finished one
+		// after anther, by the time the lock is acquired, this job might have
```
listx, line 289: s/anther/another

prow/crier/reporters/gerrit/reporter.go
```diff
@@ -253,6 +265,45 @@ func (c *Client) ShouldReport(ctx context.Context, log *logrus.Entry, pj *v1.Pro
...
+		// already been reported by another worker, refetch this pj to make sure
+		// that no duplicate report is produced
+		pjObjKey := ctrlruntimeclient.ObjectKeyFromObject(pj)
+		if err := c.lister.Get(ctx, pjObjKey, pj); err != nil {
+			if apierrors.IsNotFound(err) {
+				logger.Debug("object no longer exist")
+				return nil, nil, nil
+			}
+
+			return nil, nil, fmt.Errorf("failed to get prowjob %s: %w", pjObjKey.String(), err)
```
listx, line 299: Why do we return an error in the case where the listing of prowjob failed ("failed to get prowjob"), but we return OK if the prowjob no longer exists (IsNotFound)? I feel like in both cases we both "failed to get a prowjob", so they should return the same error.

chaodaiG: Job could be GC'ed or deleted for other reasons, this is not a prow error and should not be retried. I've added comments here

prow/crier/reporters/gerrit/reporter.go
chaodaiG: done

prow/crier/reporters/gerrit/reporter.go
```diff
@@ -559,3 +610,18 @@ func ParseReport(message string) *JobReport {
 func (r JobReport) String() string {
 	return fmt.Sprintf("%s\n%s", r.Header, r.Message)
 }
+
+func lockKeyForPJ(pj *v1.ProwJob) (*criercommonlib.SimplePull, error) {
```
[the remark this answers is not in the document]
chaodaiG: this was copy/pasted from github reporter, would keep it as is

prow/crier/reporters/gerrit/reporter.go
```diff
@@ -559,3 +610,18 @@ func ParseReport(message string) *JobReport {
...
+	// TODO(chaodaiG): remove postsubmit once
+	// https://github.com/kubernetes/test-infra/issues/22653 is fixed
+	if pj.Spec.Type != v1.PresubmitJob && pj.Spec.Type != v1.PostsubmitJob {
+		return nil, fmt.Errorf("can only get lock key for presubmit and postsubmit jobs, was %q", pj.Spec.Type)
+	}
+	if pj.Spec.Refs == nil {
+		return nil, errors.New("pj.Spec.Refs is nil")
+	}
+	if n := len(pj.Spec.Refs.Pulls); n != 1 {
+		return nil, fmt.Errorf("prowjob doesn't have one but %d pulls", n)
```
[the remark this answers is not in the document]
chaodaiG: this was copy/pasted from github reporter, would keep it as is

prow/crier/reporters/criercommonlib/shardedlock.go
```diff
@@ -0,0 +1,92 @@
...
+
+// NewSimplePull creates SimplePull
+func NewSimplePull(org, repo string, number int) *SimplePull {
+	return &SimplePull{org: org, repo: repo, number: number}
+}
+
+// ShardedLock contains sharding information based on PRs
+type ShardedLock struct {
+	mapLock *semaphore.Weighted
+	locks   map[SimplePull]*semaphore.Weighted
```
cjwagner, line 42: Is there a benefit to using semaphores rather than mutexes?

prow/crier/reporters/gerrit/reporter.go
```diff
@@ -118,10 +124,16 @@ func NewReporter(cfg config.Getter, cookiefilePath string, projects map[string][
...
+	c := &Client{
 		gc:     gc,
 		lister: lister,
-	}, nil
+		prLocks: criercommonlib.NewShardedLock(
+			semaphore.NewWeighted(1),
+			map[criercommonlib.SimplePull]*semaphore.Weighted{}),
+	}
+
+	c.prLocks.RunCleanup()
```
listx, line 135: Why is this line needed?

prow/crier/reporters/gerrit/reporter_test.go
```diff
@@ -28,13 +28,15 @@ import (
```
cjwagner, line 28: Please add a test to verify this is working and prevent regressions. I think it is sufficient to concurrently call `Report` for a few different PJs that target the same gerrit change and use a fake gerrit client that modifies some change-specific state without any additional synchronization. Go's race detector should then catch any races.

prow/crier/reporters/criercommonlib/shardedlock.go
```diff
@@ -0,0 +1,92 @@
...
+}
+
+// NewShardedLock creates ShardedLock
+func NewShardedLock(mapLock *semaphore.Weighted, locks map[SimplePull]*semaphore.Weighted) *ShardedLock {
```
cjwagner, line 46: Nit: This initialization doesn't need to be parameterized.

prow/crier/reporters/gerrit/reporter.go
```diff
@@ -559,3 +612,18 @@ func ParseReport(message string) *JobReport {
 func (r JobReport) String() string {
 	return fmt.Sprintf("%s\n%s", r.Header, r.Message)
 }
+
+func lockKeyForPJ(pj *v1.ProwJob) (*criercommonlib.SimplePull, error) {
+	// TODO(chaodaiG): remove postsubmit once
+	// https://github.com/kubernetes/test-infra/issues/22653 is fixed
```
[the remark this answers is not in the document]
chaodaiG: Technically postsubmit job doesn't know which PR it's from, gerrit knows it in a weird way which will be fixed by the issue. Once the issue is resolved then postsubmit job will not have `Refs.Pulls` any more. I have updated the comment to reflect this

prow/crier/reporters/criercommonlib/shardedlock.go
chaodaiG: this code is from github reporter, simply copy/pasted for consistency reason. I'm not quite sure why using semaphore here, maybe @alvaroaleman still remembers?

prow/crier/reporters/criercommonlib/shardedlock.go
chaodaiG: sgtm, I'll wait until Alvaro confirms the context of why using a semaphore first, it might be relevant

prow/crier/reporters/gerrit/reporter.go
chaodaiG: This is an async process keeps on cleaning up locks, it's useful to avoid accumulated memory allocation in a long running pod

listx: Sorry, I missed that this is done for the GitHub reporter already (we're just mimicking that here). LGTM

chaodaiG:
Unfortunately crier gerrit reporter generates reports from a batch of prow jobs, and need to update the previous report states before the other reporters start reporting, to avoid duplicated reports. In a multi-workers scenario, the previous report states need to be updated before locks releasing. Currently the implementation is refactoring crier controller to separate out the previous report states update functions, so that it can be reused by gerrit reporter.

I have thought about moving the locking mechanism to crier controller, which probably is a bad idea since the fact gerrit reporter reports multiple jobs at once is more like a corner case

chaodaiG:
> Unfortunately crier gerrit reporter generates reports from a batch of prow jobs, and need to update the previous report states before the other reporters start reporting, to avoid duplicated reports. In a multi-workers scenario, the previous report states need to be updated before locks releasing. Currently the implementation is refactoring crier controller to separate out the previous report states update functions, so that it can be reused by gerrit reporter.
> 
> I have thought about moving the locking mechanism to crier controller, which probably is a bad idea since the fact gerrit reporter reports multiple jobs at once is more like a corner case

@alvaroaleman and @cjwagner , does this sound reasonable to you?

prow/crier/reporters/criercommonlib/shardedlock.go
alvaroaleman: The semaphore usage was added in 3b97e7abebb406aeb487a3080ebebdf758b989da because its `Acquire` will respect the timeout of a context, there is nothing similiar available for the standard sync.Mutex, to my knowledge.

prow/crier/reporters/criercommonlib/shardedlock.go
```diff
@@ -0,0 +1,92 @@
...
+// the map, it gets recreated and acquired and two
+// routines report in parallel for the same job.
+// Note that while this function is running, no new
+// presubmit reporting can happen, as we hold the mapLock.
+func (s *ShardedLock) Cleanup() {
+	ctx := context.Background()
+	s.mapLock.Acquire(ctx, 1)
+	defer s.mapLock.Release(1)
+
+	for key, lock := range s.locks {
```
alvaroaleman, line 75: Note that there is a relatively unlikely case where two routines could end up getting a different lock for the same PR: https://github.com/kubernetes/test-infra/pull/20343

Completely eliminating that is pretty complicated (see the linked PR), so we decided that it is ok since it is unlikely to happen and even if it happens, the worst result is two identical comments in GitHub, which is tolerable. I am just mentioning this because I don't know the same applies to Gerrit.

chaodaiG: I think duplicate once is bearable imo

prow/crier/reporters/criercommonlib/shardedlock.go
chaodaiG: I've stole your comment and left it in the code for future reference

prow/crier/reporters/criercommonlib/shardedlock.go
chaodaiG: Stole your comment and add it as a comment here :)

prow/crier/reporters/gerrit/reporter_test.go
chaodaiG: I have added a test in gerrit/reporter_test.sh for this purpose, where 20 threads competing for the same lock, would that do?

prow/crier/reporters/criercommonlib/shardedlock.go
chaodaiG: done

prow/crier/reporters/gerrit/reporter.go
```diff
@@ -353,7 +406,25 @@ func (c *Client) Report(ctx context.Context, logger *logrus.Entry, pj *v1.ProwJo
...
+	// are from the same PR will still not understand that it's already
+	// reported, as the change of previous report state happens only after the
+	// returning of current function from the caller.
+	// Ideally the previous report state should be changed here.
+	logger.WithField("job-count", len(toReportJobs)).Info("Reported job(s), now will update pj(s).")
+	var err error
+	for _, pjob := range toReportJobs {
+		// if pjob.Status.PrevReportStates == nil {
+		// 	pj.Status.PrevReportStates = map[string]v1.ProwJobState{}
+		// }
```
cjwagner, line 420: Please remove.

prow/crier/reporters/gerrit/reporter.go
```diff
@@ -353,7 +406,25 @@ func (c *Client) Report(ctx context.Context, logger *logrus.Entry, pj *v1.ProwJo
...
+		if err = criercommonlib.UpdateReportStateWithRetries(ctx, pjob, logger, c.pjclientset, c.GetName()); err != nil {
+			logger.WithError(err).Error("Failed to update report state on prowjob")
+		}
+	}
+
+	// Let caller know that we are done with this job.
+	return []*v1.ProwJob{pj}, nil, err
```
cjwagner, line 427: `pj` needs to be reported before the lock is released and should already be included in `toReportJobs` so I don't think we should be returning it here.

prow/crier/reporters/gerrit/reporter_test.go
```diff
@@ -1504,26 +1843,159 @@ func TestReport(t *testing.T) {
...
+				},
+			},
+			Job:    "ci-bar",
+			Type:   v1.PresubmitJob,
+			Report: true,
+		},
+	}
+
+	// Running with 3 different batches to increase the chance of hitting races
+	for _, count := range []int{10, 20, 30} {
```
cjwagner, line 1915: FWIW Go's race detector is pretty magical and should pretty much guarantee to catch races if they're possible so this would probably work with just 2 goroutines as well. Scaling it up doesn't hurt unless we're debugging an error though so it doesn't really matter.
Thanks for adding this test.

prow/crier/reporters/gerrit/reporter_test.go
```diff
@@ -66,6 +69,7 @@ func (f *fgc) SetReview(instance, id, revision, message string, labels map[strin
 	if len(labels) > 0 {
 		f.reportLabel = labels
 	}
+	f.count++
```
cjwagner, line 72: Nit: This will be problematic in the future if we add any concurrency tests that operate on multiple PRs. This if fine for now though.

prow/crier/reporters/gerrit/reporter.go
chaodaiG: Agreed. I think it should be fine not to return anything here

prow/cmd/crier/main.go
```diff
@@ -253,7 +245,7 @@ func main() {
 	}
 
 	if o.gerritWorkers > 0 {
-		gerritReporter, err := gerritreporter.NewReporter(cfg, o.cookiefilePath, o.gerritProjects, mgr.GetCache())
+		gerritReporter, err := gerritreporter.NewReporter(cfg, o.cookiefilePath, o.gerritProjects, mgr.GetClient())
```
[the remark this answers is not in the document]
chaodaiG: With the changes from this PR, gerrit reporter need to update previous report states, so it cannot just be a reader

prow/crier/reporters/gerrit/reporter.go
chaodaiG: done

prow/cmd/crier/main.go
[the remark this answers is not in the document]
chaodaiG: Had an offline chat with @cjwagner , understand more about the concern here: is the client from `GetClient()` smart enough to use cache instead of calling api server? Based on GoDoc it seems like that the default Client uses cache https://github.com/kubernetes-sigs/controller-runtime/blob/f236f0345ad2933912ebf34bfcf0f93620769654/pkg/cluster/cluster.go#L266, if that's the case here then we might be fine. A question not very clear is: do we need to initialize the cache like https://github.com/kubernetes/test-infra/blob/66c01a3f217e7aac1afef2974a6f9b7ae71d75e0/prow/plank/reconciler.go#L95?

@alvaroaleman , did I understand it correctly?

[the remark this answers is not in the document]
alvaroaleman: Yeah, the client obtained through `GetClient` reads from the cache and writes onto the api by default. You don't need to initialize anything for that. It will automatically set up a cache for every object kind you request through it if it hasn't one yet.

The code you linked above sets up an index, which improves performance if you want to look up objects by something other than name or labels as you don't have to do a linear search over all objects.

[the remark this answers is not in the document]
chaodaiG: awesome! So this should just work!

cjwagner (approved):
LGTM (there is a merge conflict to resolve though)

chaodaiG:
conflict on a bazel file, rebased and regened the file

cjwagner (approved):

PR #24845, opened 2022-01-12
Allow to disable unique basename check for jobconfigs

xrstf:
#21870 introduced `use_full_path_as_key` to the ConfigUpdater plugin. However checkconfig still requires unique basenames. This PR changes the behaviour to make the basename check optional and allow users to disable it when appropriate.

`config-forker` and `config-rotator` always disable the check, because these tools are not there to validate the JobConfigs and will produce a stable output regardless of input basenames.

I don't really like dragging a new parameter through many function calls, but there isn't really a better option I could see. Open to suggestions though :-)

cc @xmudrii


alvaroaleman:
@xrstf sorry missed this before, can you rebase?

prow/config/config.go
```diff
@@ -1403,11 +1405,13 @@ func ReadJobConfig(jobConfig string, yamlOpts ...yaml.JSONOpt) (JobConfig, error
 			return nil
 		}
 
-		base := filepath.Base(path)
-		if uniqueBasenames.Has(base) {
-			return fmt.Errorf("duplicated basename is not allowed: %s", base)
+		if requireUniqueBasenames {
```
alvaroaleman, line 1408: It is a tad weird that this check is implemented in the config loading itself and not in checkconfig..?

PR #24847, opened 2022-01-12
Add additional, optional time throttling mechanism to ghproxy

jmguzik:
This commit adds a time throttling-based mechanism. The user needs to
specify a time-space between the arriving requests. This time-space is applied
per organization (GitHub app or alternatively hash generated from Auth
header field) to respect GH API limits applied per organization. Requests
are scheduled using only arrival time. There is no measurement of actual
request duration.

### DIFF
--- a/ghproxy/ghcache/ghcache.go
+++ b/ghproxy/ghcache/ghcache.go
@@ -36,6 +36,7 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -79,6 +80,9 @@ const (
 	// TokenExpiryAtHeader includes a date at which the passed token expires and all associated caches
 	// can be cleaned up. It's value must be in RFC3339 format.
 	TokenExpiryAtHeader = "X-PROW-TOKEN-EXPIRES-AT"
+
+	apiV3 = "v3"
+	apiV4 = "v4"
 )
 
 func CacheModeIsFree(mode CacheResponseMode) bool {
@@ -138,18 +142,123 @@ func cacheResponseMode(headers http.Header) CacheResponseMode {
 	return ModeMiss
 }
 
-func newThrottlingTransport(maxConcurrency int, roundTripper http.RoundTripper) http.RoundTripper {
-	return &throttlingTransport{sem: semaphore.NewWeighted(int64(maxConcurrency)), roundTripper: roundTripper}
+func newThrottlingTransport(maxConcurrency int, requestThrottlingTime, requestThrottlingTimeForGET uint, roundTripper http.RoundTripper, hasher ghmetrics.Hasher) http.RoundTripper {
+	return &throttlingTransport{
+		sem:                         semaphore.NewWeighted(int64(maxConcurrency)),
+		roundTripper:                roundTripper,
+		requestThrottlingTime:       requestThrottlingTime,
+		requestThrottlingTimeForGET: requestThrottlingTimeForGET,
+		hasher:                      hasher,
+		registryApiV3:               newTokensRegistry(requestThrottlingTime, requestThrottlingTimeForGET),
+		registryApiV4:               newTokensRegistry(requestThrottlingTime, requestThrottlingTimeForGET),
+	}
 }
 
-// throttlingTransport throttles outbound concurrency from the proxy
+func newTokensRegistry(requestThrottlingTime, requestThrottlingTimeForGET uint) tokensRegistry {
+	return tokensRegistry{
+		lock:                 sync.Mutex{},
+		tokens:               map[string]tokenInfo{},
+		throttlingTime:       time.Millisecond * time.Duration(requestThrottlingTime),
+		throttlingTimeForGET: time.Millisecond * time.Duration(requestThrottlingTimeForGET),
+	}
+}
+
+// tokenInfo keeps the last request timestamp and information whether it was GET request
+type tokenInfo struct {
+	getReq    bool
+	timestamp time.Time
+}
+
+// tokenRegistry keeps the timestamp of last handled request per token budget (appId or hash)
+type tokensRegistry struct {
+	lock                 sync.Mutex
+	tokens               map[string]tokenInfo
+	throttlingTime       time.Duration
+	throttlingTimeForGET time.Duration
+}
+
+func (tr *tokensRegistry) getRequestWaitDuration(tokenBudgetName string, getReq bool) time.Duration {
+	var duration time.Duration
+	tr.lock.Lock()
+	defer tr.lock.Unlock()
+	toQueue := time.Now()
+	if t, exists := tr.tokens[tokenBudgetName]; exists {
+		toQueue, duration = tr.calculateRequestWaitDuration(t, toQueue, getReq)
+	}
+	tr.tokens[tokenBudgetName] = tokenInfo{getReq: getReq, timestamp: toQueue}
+	return duration
+}
+
+func (tr *tokensRegistry) calculateRequestWaitDuration(lastRequest tokenInfo, toQueue time.Time, getReq bool) (time.Time, time.Duration) {
+	throttlingTime := tr.throttlingTime
+	// Previous request also was GET => use GET throttling time as a base
+	if lastRequest.getReq && getReq {
+		throttlingTime = tr.throttlingTimeForGET
+	}
+	duration := toQueue.Sub(lastRequest.timestamp)
+
+	if toQueue.Before(lastRequest.timestamp) || toQueue.Equal(lastRequest.timestamp) {
+		// There is already queued request, queue next afterwards.
+		difference := throttlingTime
+		if getReq {
+			difference = tr.throttlingTimeForGET
+		}
+		future := lastRequest.timestamp.Add(difference)
+		duration = future.Sub(toQueue)
+		toQueue = future
+	} else if duration >= throttlingTime || (getReq && duration >= tr.throttlingTimeForGET) {
+		// There was no request for some time, no need to wait.
+		duration = 0
+	} else {
+		// There is a queued request, wait until the next throttling tick.
+		difference := throttlingTime - duration
+		if getReq && !lastRequest.getReq {
+			difference = tr.throttlingTimeForGET - duration
+		}
+		duration = difference
+		toQueue = toQueue.Add(duration)
+	}
+	return toQueue, duration
+}
+
+// throttlingTransport throttles outbound concurrency from the proxy and adds QPS limit (1 request per given time) if enabled
 type throttlingTransport struct {
-	sem          *semaphore.Weighted
-	roundTripper http.RoundTripper
+	sem                         *semaphore.Weighted
+	roundTripper                http.RoundTripper
+	hasher                      ghmetrics.Hasher
+	requestThrottlingTime       uint
+	requestThrottlingTimeForGET uint
+	registryApiV3               tokensRegistry
+	registryApiV4               tokensRegistry
+}
+
+func (c *throttlingTransport) holdRequest(req *http.Request) {
+	var tokenBudgetName string
+	if val := req.Header.Get(TokenBudgetIdentifierHeader); val != "" {
+		tokenBudgetName = val
+	} else {
+		tokenBudgetName = c.hasher.Hash(req)
+	}
+	getReq := req.Method == http.MethodGet
+	var duration time.Duration
+	if strings.HasPrefix(req.URL.Path, "graphql") || strings.HasPrefix(req.URL.Path, "/graphql") {
+		duration = c.registryApiV4.getRequestWaitDuration(tokenBudgetName, getReq)
+		ghmetrics.CollectGitHubRequestWaitDurationMetrics(tokenBudgetName, req.Method, apiV4, duration)
+	} else {
+		duration = c.registryApiV3.getRequestWaitDuration(tokenBudgetName, getReq)
+		ghmetrics.CollectGitHubRequestWaitDurationMetrics(tokenBudgetName, req.Method, apiV3, duration)
+	}
+	if duration > 0 {
+		time.Sleep(duration)
+	}
 }
 
 func (c *throttlingTransport) RoundTrip(req *http.Request) (*http.Response, error) {
 	pendingOutboundConnectionsGauge.Inc()
+	if c.requestThrottlingTime > 0 && c.requestThrottlingTimeForGET > 0 {
+		c.holdRequest(req)
+	}
+
 	if err := c.sem.Acquire(context.Background(), 1); err != nil {
 		logrus.WithField("cache-key", req.URL.String()).WithError(err).Error("Internal error acquiring semaphore.")
 		return nil, err
@@ -210,10 +319,10 @@ func (u upstreamTransport) RoundTrip(req *http.Request) (*http.Response, error) 
 		resp.Header.Set("X-Conditional-Request", etag)
 	}
 
-	apiVersion := "v3"
+	apiVersion := apiV3
 	if strings.HasPrefix(req.URL.Path, "graphql") || strings.HasPrefix(req.URL.Path, "/graphql") {
 		resp.Header.Set("Cache-Control", "no-store")
-		apiVersion = "v4"
+		apiVersion = apiV4
 	}
 
 	ghmetrics.CollectGitHubTokenMetrics(tokenBudgetName, apiVersion, resp.Header, reqStartTime, responseTime)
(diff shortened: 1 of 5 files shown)


alvaroaleman:
It's great to see something like this in ghproxy :) I am not sure though if a fixed interval sleep between requests is the best approach for ratelimiting though, as it makes the throughput dependent on the request duration which is unlikely to be how it works on GitHubs end.


IIRC @petr-muller also mentioned that this ratelimit only applies to requests that do not use the GET verb?

Have you done some testing and an idea of how GitHub does the throttling? You mentioned the documented 1rps is incorrect, they allow more.

jmguzik:
> I am not sure though if a fixed interval sleep between requests is the best approach for ratelimiting though, as it makes the throughput dependent on the request duration which is unlikely to be how it works on GitHubs end.

In this code, I only make sure their spacing is preserved. 
The first thing is that it's not entirely fixed as there are cases (no request for a long time, recent request, and already queued). Please also take into consideration that limit is imposed per organization, so there could be more requests happening at the same time still. It perfectly works against the case where the org app is trying to fire something in the loop, it slows down the requests with close to none spacing (1-50 ms). 
The more important part is that the solution I presented is not dependent on request duration. It has a virtual "queue" and I care only about the spacing in regards to arrival time. I do not measure the _**actual**_ request duration. When I analyzed this I thought it would be overkill for the scenario where we have throttling per organization. 
So, the only reference point is staring time == the time request arrived in the `roundTripper`. If I see there was a prior request I "hold" the current one for some amount of time. If the execution of previous or next takes longer/shorter, fine, let it be. It won't probably trigger the abuse.
> 
> IIRC @petr-muller also mentioned that this rate limit only applies to requests that do not use the GET verb?

There was something about it, right. I will recheck this, although if we are throttling per org, it maybe does not matter so much.
> Have you done some testing and an idea of how GitHub does the throttling? You mentioned the documented 1rps is incorrect, they allow more.

Yes, usually to trigger the abuse you need quick shots of small messages or concurrent access in the scope of the same token (sorry, I can easily test only token). There is no visible rule on how quickly it will occur. Sometimes sooner, sometimes later. To be strict with the rules, we would need to set the spacing to 1000 ms = 1 s. I have chosen to accept input in milliseconds to possibly abuse a mechanism a little bit when needed. The correct setting would be 1s.

In the end, I just want to say, it's still a WIP, I am testing it locally, using tokens and firing requests to `ghproxy`, but also trying to keep it simple as I am already in a multithreaded scenario (roundTrippers are ofc routines) so I don't want to overshot with a complicated solution :)

jmguzik:
@alvaroaleman Answering one of your doubts with an experiment, I've hit this:
`GET https://api.github.com/users/jmguzik/repos: 403 You have exceeded a secondary rate limit. Please wait a few minutes before you try again`

Although it is stated in the docs that:
>If you're making a large number of `POST`, `PATCH`, `PUT`, or `DELETE` requests for a single user or client ID, wait at least one second between each request.

There is also a statement:
> We reserve the right to change these guidelines as needed to ensure availability.

Source: https://docs.github.com/en/rest/guides/best-practices-for-integrators#dealing-with-secondary-rate-limits

To sum up, I would keep `GET` in the scope of a mechanism. Maybe it hits the secondary API limit more rarely, as it's not listed above (maybe, I have not confirmed it), but definitely can trigger it.

alvaroaleman:
>  Answering one of your doubts with an experiment, I've hit this:

That might be solely a result and not a cause of the throttling though (i.E. it could be that the throttling only happens if you do too many mutating request but once you get throttled, it affects all requests)

jmguzik:
> That might be solely a result and not a cause of the throttling though (i.E. it could be that the throttling only happens if you do too many mutating request but once you get throttled, it affects all requests)

I wrote a small program that has 2 goroutines and is doing only `GET` calls in a loop.

petr-muller:
From https://docs.github.com/en/rest/guides/best-practices-for-integrators#dealing-with-secondary-rate-limits:

> Make requests for a single user or client ID serially. Do not make requests for a single user or client ID concurrently.
> If you're making a large number of POST, PATCH, PUT, or DELETE requests for a single user or client ID, wait at least one second between each request.

Would it be possible to implement the mechanism in a way to respect the above? Do not enforce QPS on `GET` requests, just serialize them. For non-GET requests, enforce the QPS?

jmguzik:
> Would it be possible to implement the mechanism in a way to respect the above? Do not enforce QPS on `GET` requests, just serialize them. For non-GET requests, enforce the QPS?

Yes, but as I summarized running the experiment above, it is not correct. `GET` is also affected and can trigger itself secondary API limit:
> I wrote a small program that has 2 goroutines and is doing only GET calls in a loop.

It hit the limit.

jmguzik:
@alvaroaleman @petr-muller 
I think I am finished, at least with the first iteration. I would like you to review and possibly comment on the mechanism.

A little more detailed explanation of what is actually happening here. The mechanism is based on the last timestamp of arrival requests per organization (`token`, `appId`). Only the last timestamp is stored, a _virtual_ queue is formed by imposing sleep times when needed per request. So, the request is actually "hold" from execution until calculated time ticks. There are 3 base scenarios:

1. There was no request for some time in the scope of the organization/token, so go ahead without sleep
2. There is a request already running, and the time-spacer has not ticked (so we are somewhere in the middle of the user-specified throttle time), then sleep a difference before further processing of the request.
3. We have some requests scheduled, so the current time is before or equal to the last processing time. It means there is more than 1 request scheduled so we need to take the last scheduled timestamp and add a wait duration time.

In the case of `GET` requests, the user input is replaced by 10 ms. Can't say it's enough, however in most cases that will impose some kind of serialization.

Now, there are ways to modify this:

1. Simplify. GET requests also could be scheduled using user time as lots of them is cached anyway. It will remove the need of special handling.
2. Complicate. There might be a case, especially if the `concurrency` semaphore is triggered that this ordering will stop working. If we have a low `concurrency` setting of  1 to 5 or so, messages are run in parallel and the queue is formed, it might be that ordering will go away on a semaphore waiting. This mechanism is an approximation since request durations are not measured, especially if the `concurrency` setting is a low number. We can give the option to disable `concurrency` setting or further complicate the mechanism by introducing an additional semaphore per queue (which will be double locking, as this semaphore should be probably stored in a map... you see why I resigned for now from implementing this). With the default setting of 25 it should be ok. I expect it will still work even with our current setting which is 8.

**Testing**: I tested the algorithm by writing UTs. I tested whole ghproxy running it inside minikube against tokenized (PAT) puppy apps sending requests and analyzing the logs. It seems to be working, but I understand it's a crucial place in the system so I ask for a careful review.

jmguzik:
@alvaroaleman @petr-muller please have a look at the current proposition once you find a moment.

cjwagner:
I chatted with Jakub on slack. He confirmed that the secondary rate limit is not triggered by sequential GET requests. If that's the case I think it would be better to enforce serialization rather than applying a 10ms request start spacing (10ms is only correct and efficient if the previous request happens to take exactly 10ms). Jakub understood [Alvaro's comment](https://github.com/kubernetes/test-infra/pull/24847#issuecomment-1011172774)  to suggest that we should not serialize in this way, but I thought that comment was specifically referring to the following rate limit:

> If you're making a large number of `POST`, `PATCH`, `PUT`, or `DELETE` requests for a single user or client ID, wait at least one second between each request.

not the serialization requirement for GET requests.
Could you please clarify, @alvaroaleman?

I am curious about the conditions that led you to encounter this rate limit. We manage some pretty large Prow instances and haven't had issues with this. Understanding the conditions might help us address the problem another way or anticipate encountering this problem on other Prow instances.

jmguzik:
> (10ms is only correct and efficient if the previous request happens to take exactly 10ms). Jakub understood [Alvaro's comment](https://github.com/kubernetes/test-infra/pull/24847#issuecomment-1011172774) to suggest that we should not serialize in this way, but I thought that comment was specifically referring to the following rate limit:

Just to clarify, I am more or less sure this algorithm plays nicely against `POST`, `PATCH`, `PUT`, or `DELETE` as it keeps spacing of 1s (or whatever value) nicely. I proposed 10ms for `GET`, because I am not convinced that the serialization in a sense: fire request, wait for a response, process response, signal (mutex, semaphore??), and fire next request... is needed here. I am not sure how GitHub API sees it. I am also _**not sure**_ about the minimal spacing of 10ms as it's kind of arbitrary. If you ask me, `GET` is most likely cached, so I would treat it as the rest not to complicate the algorithm. Anyway, the algorithm works per quota budget (`token`, `appId`) so I don't expect it would be a huge slowdown.

Answering a little bit of question about the scale, @alvaroaleman and @petr-muller will have more insights, but just to have a quick overview we have around 170 token budgets in our ghproxy metrics (`github_token_usage`). Our crier is configured to have 170 github workers, and imposing a limit of around 50 would probably result in slowdowns and queuing. Concurrency in ghproxy is limited to 8, and a setting of 5 was causing performance issues in the past. 
Anyway it's not only about the scale, having such an algorithm in ghproxy will result in the entire system following the GitHub API recommendations. `concurrency` guards only parallel access and does not check exactly which token budget is accessing API sequential manner and which in parallel.

alvaroaleman:
We can already serialize requests, ghproxy has a `concurrency` setting which defaults to 25. We don't do that, because last time we tried it killed the performance of Openshifts Prow, this was most visible in a Crier backlog only growing but not going down because it was unable to keep up. 

As a sidenote, this thing could be improved to enforce that concurrency on a per token + per api version (v3/v4) basis, rather than globally.

My comment was basically just suggesting to use a token bucket rate limiter rather than the fixed-interval sleep and to exclude GET requests from this, although I since learned that the latter would be wrong. Token bucket because I strongly suspect that is what their ratelimiter does and we have to mimic that.

> I am curious about the conditions that led you to encounter this rate limit

The real problem beneath all of this seems to be that the Openshift GitHub organization is too active and what GitHub documents, e.G. `If you're making a large number of POST, PATCH, PUT, or DELETE requests for a single user or client ID, wait at least one second between each request` is simply insufficient and doesn't scale.  Within the last 12 hours, the hourly increase for pre/postsubmit jobs in the Openshift organization peaked at 1.8k (`sum(increase(prowjob_state_transitions{org="openshift", type=~"presubmit|postsubmit"}[60m]))`). We do three api requests for each of them, one for triggered, one for pending, one when its done. 1800 * 3 = 5400 which is more than we have seconds in an hour and this is just for the crier reporting.

So at the end of the day what we are trying to do is to get the maximum number of api requests out which implies that we can not adhere to GitHubs documented limits as they are insufficient.

alvaroaleman:
Another thing that might help some is to (optionally?) not report "triggered", it just means "We created a job but there is nothing to see yet, that would save one api call per prowjob.

jmguzik:
> As a sidenote, this thing could be improved to enforce that concurrency on a per token + per api version (v3/v4) basis, rather than globally.

@alvaroaleman proposed solution is per token, where token is `appId` or hashed auth access token signature if `appId` not used. Divided as it is in metrics, measuring time per different budgets. You are probably right that further apiv3/4 division could be introduced here as well. Or you have other token definition that I somehow miss?

ghproxy/ghcache/ghcache.go
```diff
@@ -138,18 +139,122 @@ func cacheResponseMode(headers http.Header) CacheResponseMode {
...
+	if t, exists := tr.tokens[tokenBudgetName]; exists {
+		toQueue, duration = tr.calculateRequestWaitDuration(t, toQueue, getReq)
+	}
+	tr.tokens[tokenBudgetName] = tokenInfo{getReq: getReq, timestamp: toQueue}
+	return duration
+}
+
+func (tr *tokensRegistry) calculateRequestWaitDuration(t tokenInfo, toQueue time.Time, getReq bool) (time.Time, time.Duration) {
+	throttlingTime := tr.throttlingTime
+	followingGetRequests := t.getReq && getReq
```
petr-muller, line 191: The identifiers `t` and `tr` tend to mix with each other, I suggest to rename `t` to `lastRequest` (it is much more an info about the last request than anything related to tokens)

ghproxy/ghcache/ghcache.go
```diff
@@ -138,18 +139,122 @@ func cacheResponseMode(headers http.Header) CacheResponseMode {
...
+	}
+	getReq := req.Method == http.MethodGet
+	var duration time.Duration
+	if strings.HasPrefix(req.URL.Path, "graphql") || strings.HasPrefix(req.URL.Path, "/graphql") {
+		duration = c.apiv4.getRequestWaitDuration(tokenBudgetName, getReq)
+	} else {
+		duration = c.apiv3.getRequestWaitDuration(tokenBudgetName, getReq)
+	}
+
+	if duration > 0 {
```
petr-muller, line 247: I think `duration` would make a good metric to record somehow? If this grows too much then that is probably something the operator should be aware of.

petr-muller (commented):
I don't have concerns about whether to implement the throttling this way or another. One thing that comes to mind no matter what the implementation is whether we should also improve graceful shutdown. The requests "held" inside ghproxy when it gets e.g. bumped, evicted or whatever get lost, we should at least think whether we need something like that (or intentionally decide we don't)

ghproxy/ghcache/ghcache.go
jmguzik: Added metrics per org (token budget name), API and request type (`GET`, `POST`, etc)

jmguzik:
@petr-muller @alvaroaleman @cjwagner I did some changes to the algorithm. Now it's similar to the leaky bucket (except it's not leaky:) ). I tested it using `minikube` and load of requests and on such a scale, it works well. Unknown how it will work on a cluster as I can test it only after merge. All the algo cases are included in unit tests.

Throttling using this algorithm is optional. Two params need to be passed to get this algorithm enabled. The first is throttling time for all requests except `GET` and the second one is for `GET` requests only. I resigned from controversial 10ms in order to give users the full possibility to tune the system.

 The algorithm is per token budget (`appID`) and per API version (different queues for `APIv3` and `APIv4`). In our installation, `appID` is different for every organization, so it will have its own queue.

As we don't actually know the specifics of GH API interaction so I can't exactly tell how this could be better than other approaches, however, I believe it could be good enough to be used as a first try. I included the `duration` metrics so we could see how it scales up (metrics are per `token`, `request_type`, and `API`).

jmguzik:
> The requests "held" inside ghproxy when it gets e.g. bumped, evicted or whatever get lost, we should at least think whether we need something like that (or intentionally decide we don't).

To be honest I thought we could go with the full leaky bucket and just reject requests when sleeping time is higher than some value (eg 60 secs). However, I think that possibility of such a thing happening per org is not huge. I would leave such improvements for the next iteration if we detect the current mechanism is not good enough.

jmguzik:
@cjwagner @alvaroaleman @petr-muller could we push this topic forward?
If you are worried about performance, I just want to say this solution is completely optional, without effect on current installations. If it proves not successful I'll for sure will continue working on it, upgrade or change the approach if needed.

cjwagner (approved):
This is fine with me since it is opt-in and doesn't affect instances that aren't experiencing such rate limiting issues.
/hold

> > The scheduing-based algorithm with sleeps makes it impossible to at least try to empty the queue when needed.
> 
> Yes, but it's like the best what you can do if you already have roundtripper goroutines. At least I don't see a cleaner solution.

One option would be to use [`time.After`](https://pkg.go.dev/time#After) inside of a `select` statement instead of sleep. That would allow you to include another entry in the select statement for a channel that signals round trippers to abort early rather than continue blocking on the `time.After` channel.
Another option would be to block on the `Done()` channel of a context instead of sleeping and have a separate goroutine sleep and then cancel the context. When there is an interrupt that triggers a shutdown, all the active contexts can be cancelled early.

jmguzik:
> One option would be to use [`time.After`](https://pkg.go.dev/time#After) inside of a `select` statement instead of sleep. That would allow you to include another entry in the select statement for a channel that signals round trippers to abort early rather than continue blocking on the `time.After` channel. Another option would be to block on the `Done()` channel of a context instead of sleeping and have a separate goroutine sleep and then cancel the context. When there is an interrupt that triggers a shutdown, all the active contexts can be cancelled early.

@cjwagner many thanks for getting back to the PR.

`time.After` was even part of the early solution so that's definitely an option. My arguments for skipping back off or queue abandon mechanism for the first solution are:
1. `concurrency` setting, processed later on, does not have it either. In the end, `concurrency` is another method of throttling, and if set to a low value, a queue forms. Theoretically, the `concurrency` setting queue could be even larger as I make division per `appId` and API versions.
2. If that mechanism proves ineffective, the first approach should be "play with values". If that does not help I think we have a problem with the algorithm itself. I think we don't want to cancel requests after all.

Similar to a situation when `concurrency` is set to some small value, the above problem will be visible in metrics and the next solution or upgrade could be crafted based on actual data.

petr-muller:
Yeah, I think improving graceful shutdown can be done as a followup, incrementally. I think it *will* be necessary because with serializing the queue is implicit in some situations (you are right that the existing `concurrency` does not handle that case gracefully, too) and ditching the queue on shutdown is quite hostile to the clients.

As far as I'm concerned, unhold as needed.

petr-muller (approved):

PR #24849, opened 2022-01-12
Make windows capz job more frequent

jsturtevant:
With https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1932 merged we have better logging and some of the flakiness should be resolved. 

Would like to run this as frequently as our release informing job https://testgrid.k8s.io/sig-windows-signal#aks-engine-windows-containerd-master to help spot more flakes with the eventual goal of replacing the job with this one.

/sig windows
/assign @marosset


marosset (approved):
/lgtm

PR #24853, opened 2022-01-12
Add support to gitClientFactory for CloneURI Remote Resolver

mpherman2:
Starting the work to add gerrit support for InRepo Config. 
/assign @cjwagner @chaodaiG

### DIFF
--- a/prow/git/v2/client_factory.go
+++ b/prow/git/v2/client_factory.go
@@ -23,6 +23,8 @@ import (
 	"runtime"
 	"sync"
 
+	"net/url"
+
 	"github.com/sirupsen/logrus"
 	utilpointer "k8s.io/utils/pointer"
 )
@@ -55,6 +57,8 @@ type ClientFactoryOpts struct {
 	Host string
 	// UseSSH, defaults to false
 	UseSSH *bool
+	// CloneURI will use CloneURI Remote resolver if set.
+	CloneURI *url.URL
 	// The directory in which the cache should be
 	// created. Defaults to the "/var/tmp" on
 	// Linux and os.TempDir otherwise
@@ -93,6 +97,9 @@ func (cfo *ClientFactoryOpts) Apply(target *ClientFactoryOpts) {
 	if cfo.Username != nil {
 		target.Username = cfo.Username
 	}
+	if cfo.CloneURI != nil {
+		target.CloneURI = cfo.CloneURI
+	}
 }
 
 // ClientFactoryOpts allows to manipulate the options for a ClientFactory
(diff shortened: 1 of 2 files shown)


prow/git/v2/remote.go
```diff
@@ -144,3 +144,23 @@ func (f *pathResolverFactory) PublishRemote(org, repo string) RemoteResolver {
...
+	return func() (string, error) {
+		return f.cloneURI.String(), nil
+	}
+}
+
+// Gerrit does not use forks the same way that github does. This will be left Unimplemented.
+// Implementation of Gerrit Interactor will use Central Remote but commit to different Refs.
+func (f *cloneURIResolverFactory) PublishRemote(_, _ string) RemoteResolver {
+	return func() (string, error) {
+		return "", errors.New("Not Implemented")
```
stevekuznetsov, line 164: nit:
```suggestion
		return "", errors.New("not implemented")
```

stevekuznetsov: IMO could also be a panic

stevekuznetsov: Ideally, there would be some way to tell before merging a PR that a gerrit client is calling this and should not be

prow/git/v2/remote.go
```diff
@@ -144,3 +144,23 @@ func (f *pathResolverFactory) PublishRemote(org, repo string) RemoteResolver {
 		return path.Join(f.baseDir, org, repo), nil
 	}
 }
+
+// cloneURIResolverFactory is for clients where we want the Central Remote to return the CloneURI
+// and do not care about a seprate publisher Remote.
+type cloneURIResolverFactory struct {
+	cloneURI *url.URL
```
stevekuznetsov, line 151: Why hold the URL, not the string?

mpherman2: Good point. Changed

prow/git/v2/remote.go
mpherman2: When I implement the gerrit support I will keep this in mind as something to add. I added a TODO

prow/git/v2/client_factory.go
```diff
@@ -129,7 +136,11 @@ func NewClientFactory(opts ...ClientFactoryOpt) (ClientFactory, error) {
 		return nil, err
 	}
 	var remotes RemoteResolverFactory
-	if o.UseSSH != nil && *o.UseSSH {
+	if o.CloneURI != nil {
```
chaodaiG, line 139: Would expect something more strict than `CloneURI != nil -> Using cloneURIResolverFactory`, just like `o.UseSSH` below.

prow/git/v2/remote.go
mpherman2: Decided to actually have this return CloneURI as well. Github clients can use CloneURIRemoteResolverFactory when CentralRemote and PublisherRemote are both CloneURI. No need to leave this for gerrit only.

prow/git/v2/client_factory.go
mpherman2: Discussed offline. Added support for PublisherRemote for CloneURIResolverFactory so adding CloneURI won't cause surprising behavior

chaodaiG (approved):
/hold

PR #24855, opened 2022-01-12
skip pod cidrs test for azure dual-stack

aramase:

As part of https://github.com/kubernetes/kubernetes/issues/104544, `should have ipv4 and ipv6 node podCIDRs` test was removed in 1.23. We moved the azure dual-stack jobs to use `containerd` + `azurecni` with the dockershim deprecation which doesn't set both the pod CIDRs. Skipping the test in 1.21 and 1.22 as it's no longer valid for azure clusters.


MrHohn (approved):
/lgtm
/approve

PR #24621, opened 2021-12-09
capz: run presubmit exp job when managed* files change

jackfrancis:
This PR updates the regex match pattern for the capz `pull-cluster-api-provider-azure-e2e-exp` job to include "managed*"-named source file changes.

PR #24856, opened 2022-01-12
capz: fix AKS regex so that jobs run appropriately

jackfrancis:
PR #24621 introduced a "managed.go" regex match so that such-named files would trigger AKS runs. That match needs to not be in a group that includes a `^` (beginning of line) requirement. This PR fixes that, and adds an additional "aks" match to handle cluster template files such-named.


jackfrancis:
cc @zmalik @CecileRobertMichon

Issue #17719, opened 2020-05-27
maintenance-pull-janitor and maintenance-pull-scale-janitor still using bootstrap


ixdy:
**What should be cleaned up or changed**:
I'm attempting to update the various janitors that are using the `gcp_janitor.py` script inside the `boskos/` directory in preparation for removing boskos from test-infra (https://github.com/kubernetes/test-infra/issues/17614).

It looks like the `maintenance-pull-janitor` and `maintenance-pull-scale-janitor` jobs are still using bootstrap, which is apparently deprecated and no longer supported:

https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/maintenance-pull-janitor/1265432629549731842/
```
W0527 00:00:59.933] **************************************************************************
bootstrap.py is deprecated!
test-infra oncall does not support any job still using bootstrap.py.
Please migrate your job to podutils!
https://github.com/kubernetes/test-infra/blob/master/prow/pod-utilities.md
**************************************************************************
```

https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/maintenance-pull-scale-janitor/1265430112350769153/
```
W0526 23:50:37.804] **************************************************************************
bootstrap.py is deprecated!
test-infra oncall does not support any job still using bootstrap.py.
Please migrate your job to podutils!
https://github.com/kubernetes/test-infra/blob/master/prow/pod-utilities.md
**************************************************************************
```

These should probably either be fixed or deleted.

(Also, maintenance-pull-janitor has been failing continuously for weeks: https://testgrid.k8s.io/sig-testing-maintenance#pull-janitor&width=5)

BenTheElder:
What's the current status here?

ixdy:
It appears that nothing has changed.
I ended up not deleting `gcp_janitor.py` from test-infra, so as to not break these jobs further, but at some point they should probably be updated to use the `gcp_janitor.py` from kubernetes-sigs/boskos instead, since that one is maintained.

BenTheElder:
> ... but at some point they should probably be updated to use the gcp_janitor.py from kubernetes-sigs/boskos instead, since that one is maintained.

That seems important. Are you going to do this?

ixdy:
I think first we'd need to convert these jobs to using podutils, as that would let us more easily clone a second git repository (kubernetes-sigs/boskos). I think there may have been other issues when I was investigating this in May, but I don't remember anymore.

[...]

PR #24872, opened 2022-01-13
Document why /boskos directory exists, migrate to make rule

chaodaiG:
/cc @BenTheElder @spiffxp


boskos/README.md
```diff
@@ -3,3 +3,11 @@
 Boskos has been moved into its own repository under kubernetes-sigs:
 
 https://github.com/kubernetes-sigs/boskos
+
+# What's still here
+
+The only thing matters here is `./cmd/janitor/gcp_janitor.py` script, which is
+still used by
+[/scenarios/kubernetes_janitor.py#L66](https://github.com/kubernetes/test-infra/blob/d641897cea52d493ef883a4dfa6728ffdfa02dfa/scenarios/kubernetes_janitor.py#L66),
```
BenTheElder, line 11: I forget, is boskos using a different copy of this when deployed to janitor GCP or is this also the main janitor script?

chaodaiG: boskos has it's own `gcp_janitor.py` script https://github.com/kubernetes-sigs/boskos/blob/master/cmd/janitor/gcp_janitor.py, which is much more recent this this one, assuming the other one is the default?

BenTheElder: I don't know 🙈 , Jeff was probably the last person to know.

chaodaiG: Just in case @ixdy still receives GitHub notification ^^

ixdy: I *think* that the Boskos janitor uses the script from `kubernetes-sigs/boskos`, since I think it uses the copy baked into the container, which would come from that repo.

I opened #17719 regarding the copy of `gcp_janitor.py` in test-infra. There was a potentially hairy cleanup effort, and it also wasn't clear whether the jobs even needed to run anymore. With the resolution of #20760, this can probably finally be deleted.

BenTheElder: tempted to just disable scheduling those and see what breaks :+)

BenTheElder (approved):
/lgtm
/approve
/hold

PR #24882, opened 2022-01-14
Allow App Auth related Post Requests in Dry Run

nouseforaname:
When in dry run mode allow obtaining an app access token
to be able to perform authenticated get requests.

We we're trying to run a dry run// confirm=false with app auth and were not getting to the "skipping changes..." log lines that would've helped us evaluating which changes would be done. 

It would error out early within the check we removed here: 
https://github.com/kubernetes/test-infra/compare/master...nouseforaname:feature/dry-run-authenticated-get-requests-app-auth?expand=1#diff-43bdff30b56d33fef7db34c315dd186dab4fa8295ef72f8796fdf33984e959aeL4652-L4654

instead we added a regex check here:
https://github.com/kubernetes/test-infra/compare/master...nouseforaname:feature/dry-run-authenticated-get-requests-app-auth?expand=1#diff-43bdff30b56d33fef7db34c315dd186dab4fa8295ef72f8796fdf33984e959aeR991

to make sure that we only post to the token endpoint.
CC @rkoster

### DIFF
--- a/prow/github/app_auth_roundtripper_test.go
+++ b/prow/github/app_auth_roundtripper_test.go
@@ -61,67 +61,6 @@ func (frt *fakeRoundTripper) RoundTrip(r *http.Request) (*http.Response, error) 
 func TestAppsAuth(t *testing.T) {
 
 	const appID = "13"
-	testCasesDryRun := []struct {
-		name                string
-		githubBaseURL       string
-		cachedAppSlug       *string
-		cachedInstallations map[string]AppInstallation
-		cachedTokens        map[int64]*AppInstallationToken
-		doRequest           func(Client) error
-		responses           map[string]*http.Response
-		verifyRequests      func([]*http.Request) error
-	}{
-		{
-			name:          "App installation auth success, installations and token is requsted, No Writes executed",
-			cachedAppSlug: utilpointer.StringPtr("ci-app"),
-			doRequest: func(c Client) error {
-				_, err := c.GetOrg("org")
-				if err != nil {
-					return err
-				}
-				_, err = c.CreateRepo("org", false, RepoCreateRequest{RepoRequest: RepoRequest{Name: utilpointer.StringPtr("test-repo")}})
-				return err
-			},
-			responses: map[string]*http.Response{
-				"/app/installations":                 {StatusCode: 200, Body: serializeOrDie([]AppInstallation{{ID: 1, Account: User{Login: "org"}}})},
-				"/app/installations/1/access_tokens": {StatusCode: 201, Body: serializeOrDie(AppInstallationToken{Token: "the-token"})},
-				"/orgs/org":                          {StatusCode: 200, Body: serializeOrDie(Organization{})},
-			},
-			verifyRequests: func(r []*http.Request) error {
-				if n := len(r); n != 3 {
-					return fmt.Errorf("expected exactly three request, got %d", n)
-				}
-				if r[0].URL.Path != "/app/installations" {
-					return fmt.Errorf("expected first request to have path '/app/installations' but had %q", r[0].URL.Path)
-				}
-				if val := r[0].Header.Get("Authorization"); !strings.HasPrefix(val, "Bearer ") {
-					return fmt.Errorf("expected the Authorization header %q to start with 'Bearer '", val)
-				}
-				if val := r[0].Header.Get("X-PROW-GHCACHE-TOKEN-BUDGET-IDENTIFIER"); val != "ci-app" {
-					return fmt.Errorf("expected X-PROW-GHCACHE-TOKEN-BUDGET-IDENTIFIER header %q to have value ci-app", val)
-				}
-
-				if r[1].URL.Path != "/app/installations/1/access_tokens" {
-					return fmt.Errorf("expected second request to request a token, but had path %s", r[0].URL.Path)
-				}
-				if val := r[1].Header.Get("Authorization"); !strings.HasPrefix(val, "Bearer ") {
-					return fmt.Errorf("expected the Authorization header %q to start with 'Bearer '", val)
-				}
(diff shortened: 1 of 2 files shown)


matthyx:
Is it still a dry-run if you can get the token?

nouseforaname:
> Is it still a dry-run if you can get the token?

IMO: Authenticated dry-runs are quite common. In most cases creating a diff for the changes requires to be authenticated and authorized. Not quite sure how this can be achieved within this context without having access to some kind of authentication. 

Basically (at least from our perspective) it would be nice to have the same experience for peribolos whether you use 
`-github-token-path` vs `-github-app-id + -github-app-private-key-path` for auth

matthyx:
/cc @alvaroaleman 
WDYT Alavaro?

prow/github/client.go
```diff
@@ -986,7 +988,7 @@ func (c *client) requestRaw(r *request) (int, []byte, error) {
 }
 
 func (c *client) requestRawWithContext(ctx context.Context, r *request) (int, []byte, error) {
-	if c.fake || (c.dry && r.method != http.MethodGet) {
```
alvaroaleman, line 989: /cc @petr-muller @chaodaiG 

WDYT? I understand the thought here but it is also going to be surprising

petr-muller: I'm leaning towards allowing it. Authentication is special and is usually not thought as being a write operation, which is my bar for what I expect from dry runs. It has legit use cases. 

Technically it *is* a write operation though - are we sure there is no observable side effects of executing such dry-run other than possibly obtaining a new installation token that will expire anyway?

chaodaiG: On the exact same page with @petr-muller , imo it's fine to allow requesting this token as part of dry-run. Might be helpful to make it a bit more clear on the allowed request path, with one of the two approaches:

a).  Commenting the regex to make it clear why
b). Encapsulate the condition in a function called `isRequestingAppInstallationToken`

On the other hand, adding one more flag `--dryrun-allow-request-app-installation-token` to make this configurable could be helpful to avoid surprising side effect. We can add an announcement at https://github.com/kubernetes/test-infra/blob/master/prow/ANNOUNCEMENTS.md#new-features and stating that this feature will become default in 6 months etc.

alvaroaleman: sgtm, while it is a mutating request it doesn't result in any observable state change on GitHub

nouseforaname:
Any update on this? Not sure if this is stuck on me missing something

alvaroaleman:
/ok-to-test
The change seems fine, could you add a test for this please?

nouseforaname:
> /ok-to-test The change seems fine, could you add a test for this please?

Sorry for replying so late. Somehow overlooked the notification around the request.

I could add a test, but looking at the proposed change I'm not quite sure what to test for?

dims:
please rebase to latest to remove the broken `cla/linuxfoundation` status

alvaroaleman:
> I could add a test, but looking at the proposed change I'm not quite sure what to test for?

Well, that requests from the apps auth roundtripper get executed even when in dry run mode? I also think this is incomplete, the flow is to first get a token using the RSA key and app id and then getting an installation token. It doesn't seem like the former is allowlisted.

nouseforaname:
> RSA key and app id

But that isn't a request it's just creating a JWT using the PrivKey  and AppID (https://docs.github.com/en/developers/apps/building-github-apps/authenticating-with-github-apps) which you then use to get the actual Token.

nouseforaname:
> /ok-to-test The change seems fine, could you add a test for this please?

I tried adding a tests but I'm really bad at it. Let me know if that is what you expected and if not what needs to be changed

alvaroaleman:
> But that isn't a request it's just creating a JWT using the PrivKey and AppID (https://docs.github.com/en/developers/apps/building-github-apps/authenticating-with-github-apps) which you then use to get the actual Token or did I misunderstand that

Yes, but that actual token is needed to do anything useful. Effectively you will need to allowlist all the paths used in any of the methods of the apps auths roundtrippers github client here: https://github.com/kubernetes/test-infra/blob/5c3cd7ce2f5f442364c924863aafd6dd89da5efb/prow/github/app_auth_roundtripper.go#L40-L44

prow/github/app_auth_roundtripper_test.go
```diff
@@ -61,6 +61,67 @@ func (frt *fakeRoundTripper) RoundTrip(r *http.Request) (*http.Response, error)
 func TestAppsAuth(t *testing.T) {
 
 	const appID = "13"
+	testCasesDryRun := []struct {
+		name                string
```
alvaroaleman, line 65: Why did you redefine the struct used for the `testCases` and create a new slice rather than just appending your test to the `testCases` slice? Just add a `dryRunClient` parameter to the testCases and use that to configure if the client should be created with dry run mode or not. You can change the client construction from `NewAppsAuthClientWithFields` to `NewClientFromOptions`, that allows to configure the dry run mode

prow/github/client.go
```diff
@@ -1005,7 +1008,7 @@ func (c *client) requestRaw(r *request) (int, []byte, error) {
...
+	if c.fake || c.dry && (r.method != http.MethodGet && (!allowedDryRunPostPaths.MatchString(r.path) && r.method == http.MethodPost)) {
```
alvaroaleman, line 1011:
```suggestion
	if c.fake || c.dry && (r.method != http.MethodGet || !allowedDryRunPostPaths.MatchString(r.path) )) {
```

nouseforaname:
> 

Isn't that what the regex takes care of? I'm not sure which additional post paths to add there as the path where we post the JWT (that is created locally) is in there already.

nouseforaname:
bump

matthyx:
@alvaroaleman can you have a second look?

alvaroaleman:
@nouseforaname this still isn't addressed so this will not actually work: https://github.com/kubernetes/test-infra/pull/24882#issuecomment-1111337997

PR #24885, opened 2022-01-14
Crier: collects prometheus metrics for times spent reporting

chaodaiG:
And by the way fixing a minor issue:
- When a reporter reports on behalf of other jobs, it's also responsible for updating previous reported states on the other jobs, the current behavior is quit right away when one of them failed, this is probably not the best thing to do as when this job was reconciled again it'll pick up from where it's left at and there is a higher chance of error the next time around. Update the logic to continue updating the rest of the jobs and return with error. Note that this is only applicable to gerrit reporter at the moment

### DIFF
--- a/prow/crier/controller.go
+++ b/prow/crier/controller.go
@@ -205,14 +205,24 @@ func (r *reconciler) reconcile(ctx context.Context, log *logrus.Entry, req recon
 	}
 
 	log.WithField("job-count", len(pjs)).Info("Reported job(s), now will update pj(s).")
+	var lastErr error
 	for _, pjob := range pjs {
 		if err := r.updateReportStateWithRetries(ctx, pjob, log); err != nil {
 			log.WithError(err).Error("Failed to update report state on prowjob")
-			return nil, err
+			// The error above is alreay logged, so it would be duplicated
+			// effort to combine all errors to return, only capture the last
+			// error should be sufficient.
+			lastErr = err
 		}
 	}
 
-	return nil, nil
+	if pj.Status.CompletionTime != nil {
+		latency := time.Now().Unix() - pj.Status.CompletionTime.Unix()
+		crierMetrics.latency.WithLabelValues(r.reporter.GetName()).Observe(float64(latency))
+		log.WithField("latency", latency).Debug("Report latency.")
+	}
+
+	return nil, lastErr
 }
 
 func (r *reconciler) shouldHandle(pj *prowv1.ProwJob) bool {
--- a/prow/crier/metrics.go
+++ b/prow/crier/metrics.go
@@ -1,0 +1,39 @@
+/*
+Copyright 2022 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package crier reports finished prowjob status to git providers.
+package crier
+
+import "github.com/prometheus/client_golang/prometheus"
+
+// Prometheus Metrics
+var (
+	crierMetrics = struct {
+		latency *prometheus.HistogramVec
+	}{
+		latency: prometheus.NewHistogramVec(prometheus.HistogramOpts{
+			Name:    "crier_report_latency",
+			Help:    "Histogram of time spent reporting, calculated by the time difference between job completion and end of reporting.",
+			Buckets: []float64{1, 10, 20, 30, 60, 120, 300, 600, 1800, 3600},
+		}, []string{
+			"reporter",
+		}),
+	}
+)
+
+func init() {
+	prometheus.MustRegister(crierMetrics.latency)
+}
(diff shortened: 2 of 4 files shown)


prow/crier/metrics.go
```diff
@@ -0,0 +1,47 @@
...
+		reportTime: prometheus.NewHistogramVec(prometheus.HistogramOpts{
+			Name:    "report_time",
+			Help:    "Histogram of time spent reporting, calculated by the time difference between job completion and end of reporting.",
+			Buckets: reportTimeBuckets,
+		}, []string{
+			"org",
+			"repo",
+			"jobtype",
+			"jobname",
+			"reporter",
```
cjwagner, line 40: We're already seeing problems with excessive label cardinality with `prowjob_state_transitions`, this seems to use a similar label set, but also has 18 histogram counter buckets per label combination. I think this is going to cause even larger problems as a result.

prow/crier/metrics.go
```diff
@@ -0,0 +1,47 @@
...
+var (
+	// Use Fibonacci to get a good distribution. We care about report delayed
+	// for minutes, and even up to an hour.
+	reportTimeBuckets = []float64{1, 2, 3, 5, 8, 13, 21, 34, 55, 89, 144, 233, 377, 610, 987, 1597, 2584, 4181}
+
+	crierMetrics = struct {
+		reportTime *prometheus.HistogramVec
+	}{
+		reportTime: prometheus.NewHistogramVec(prometheus.HistogramOpts{
+			Name:    "report_time",
```
cjwagner, line 32: Can we call this reporting delay or latency rather than time? I think that is an important distinction that we may care about later.

cjwagner (changes requested):

prow/crier/metrics.go
chaodaiG: yeah, the same prowjob may report this multiple times, so this might be even worse than `prowjob_state_transitions`, I'll think about how to make this smaller. Please also update here if there is a better idea

cjwagner: I'd suggest getting rid of all the labels except `reporter` and/or using fewer histogram buckets.  I don't think we really care about the `jobname` for this metric. The `org`, `repo`, and `jobtype` could be a bit interesting, but are not critical. I think we mainly just need this metric scoped to the `reporter`.

chaodaiG: agreed, currently reporter should be sufficient. How about add a log with all these fields for easier debugging if ever needed?

prow/crier/metrics.go
chaodaiG: renamed `latency`

prow/crier/metrics.go
chaodaiG: done

prow/crier/controller.go
```diff
@@ -205,14 +205,22 @@ func (r *reconciler) reconcile(ctx context.Context, log *logrus.Entry, req recon
...
-			return nil, err
+			// The error above is alreay logged, so it would be duplicated
+			// effort to combine all errors to return, only capture the last
+			// error should be sufficient.
+			lastErr = err
 		}
 	}
 
-	return nil, nil
+	latency := time.Now().Unix() - pj.Status.CompletionTime.Unix()
```
cjwagner: This will segfault for jobs that have not completed yet, right? Using completion time to measure report latency only makes sense if we are reporting the job as completed. For triggered and pending jobs we should use a different time or avoid reporting latency in those cases and change the metric name accordingly.

prow/crier/controller.go
```diff
@@ -205,14 +205,22 @@ func (r *reconciler) reconcile(ctx context.Context, log *logrus.Entry, req recon
...
+	crierMetrics.latency.WithLabelValues(r.reporter.GetName()).Observe(float64(latency))
+	log.WithField("latency", latency).Debug("Time spent reporting.")
```
cjwagner: This is now latency, not the amount of time we spent reporting.

prow/crier/metrics.go
```diff
@@ -0,0 +1,39 @@
...
+
+// Prometheus Metrics
+var (
+	crierMetrics = struct {
+		latency *prometheus.HistogramVec
+	}{
+		latency: prometheus.NewHistogramVec(prometheus.HistogramOpts{
+			Name:    "report_time",
+			Help:    "Histogram of time spent reporting, calculated by the time difference between job completion and end of reporting.",
+			Buckets: []float64{1, 10, 60, 600, 1800, 3600},
```
cjwagner, line 30: We probably want some more buckets between 1 and 600 seconds, right? I'd expect most to fall into either the 10 or 60 bucket during normal operation which won't tell us very much. I think we'll want to be able to measure improvements more granularly than that? 
Since we've limited the label set to just `reporter` its safe to use more buckets again.

chaodaiG: good point, added

prow/crier/controller.go
chaodaiG: updated

prow/crier/controller.go
chaodaiG: added a condition

prow/crier/metrics.go
```diff
@@ -0,0 +1,39 @@
...
+
+import "github.com/prometheus/client_golang/prometheus"
+
+// Prometheus Metrics
+var (
+	crierMetrics = struct {
+		latency *prometheus.HistogramVec
+	}{
+		latency: prometheus.NewHistogramVec(prometheus.HistogramOpts{
+			Name:    "report_time",
```
cjwagner, line 28: Oops, I missed this on last review, this still needs to be renamed.

cjwagner (approved):

PR #24887, opened 2022-01-14
use correct set of images for windows 1.20 presubmit

jsturtevant:
The 1.20 presubmit for containerd is failing: https://testgrid.k8s.io/sig-windows-presubmit#pull-kubernetes-e2e-aks-engine-windows-containerd-1.20

should be using different repolist: https://github.com/kubernetes-sigs/windows-testing/tree/master/images#image-repository-list

The images are failing to pull:

```
concurrent-1642097940-btmdk
Jan 13 18:23:06.475: INFO: At 2022-01-13 18:19:06 +0000 UTC - event for concurrent-1642097820-lqnqr: {kubelet 2641k8s000} Failed: Error: ImagePullBackOff
Jan 13 18:23:06.475: INFO: At 2022-01-13 18:19:06 +0000 UTC - event for concurrent-1642097880-rfjnv: {kubelet 2641k8s000} Failed: Error: ImagePullBackOff
Jan 13 18:23:06.475: INFO: At 2022-01-13 18:19:08 +0000 UTC - event for concurrent-1642097940-btmdk: {kubelet 2641k8s001} Pulling: Pulling image "docker.io/library/busybox:1.29"
Jan 13 18:23:06.475: INFO: At 2022-01-13 18:19:11 +0000 UTC - event for concurrent-1642097940-btmdk: {kubelet 2641k8s001} Failed: Error: ImagePullBackOff
Jan 13 18:23:06.475: INFO: At 2022-01-13 18:19:11 +0000 UTC - event for concurrent-1642097940-btmdk: {kubelet 2641k8s001} Failed: Failed to pull image "docker.io/library/busybox:1.29": rpc error: code = NotFound desc = failed to pull and unpack image "docker.io/library/busybox:1.29": no match for platform in manifest: not found
Jan 13 18:23:06.475: INFO: At 2022-01-13 18:20:05 +0000 UTC - even
```

PR #24888, opened 2022-01-14
Don't use the secret for private images  in 1.20 jobs

jsturtevant:
Follow up to https://github.com/kubernetes/test-infra/pull/24887

Don't use the docker secret config file for <1.21

/sig windows
/assign @marosset


CecileRobertMichon (commented):
/lgtm

PR #24890, opened 2022-01-14
sig-cloud-provider: drop ci-kubernetes-e2e-ubuntu-gce

endocrimes:
Duplicates gce-ubuntu-master-containerd now that dockershim has been
removed, so can now be dropped from test-infra.

related to: https://github.com/kubernetes/kubernetes/issues/107469


SergeyKanzhelev (approved):
/lgtm

I don't know if there were any history why it is run by cloud-provider. Since configuraiton is the same as one we run in sig node, let's delete this duplicate

PR #24893, opened 2022-01-14
Allow Gerrit formatted Repos for InRepoConfig Config Validation

mpherman2:
Continued work on Gerrit support for InRepo-Config.
This work should allow InRepo-Config Configuration to work with Gerrit formatted repos.
/assign @chaodaiG @listx


petr-muller:
/uncc 
I know neither Gerrit nor InRepoConfig :(

prow/config/config.go
```diff
@@ -216,14 +216,28 @@ type InRepoConfig struct {
...
+	org, repo, err := SplitRepoName(identifier)
 	if c.InRepoConfig.Enabled[identifier] != nil {
-		return *c.InRepoConfig.Enabled[identifier]
+		return *c.InRepoConfig.Enabled[strings.Join([]string{org, repo}, "/")]
 	}
-	identifierSlashSplit := strings.Split(identifier, "/")
-	if len(identifierSlashSplit) == 2 && c.InRepoConfig.Enabled[identifierSlashSplit[0]] != nil {
-		return *c.InRepoConfig.Enabled[identifierSlashSplit[0]]
+	// If err then it cannot be split.
+	if c.InRepoConfig.Enabled[org] != nil && err == nil {
```
chaodaiG, line 239: If `err != nil` then both org and repo are empty string, then the check above would evaluate `/` instead, which doesn't seem safe. Should this error check happen earlier?

prow/config/config.go
```diff
@@ -233,15 +247,17 @@ func (c *Config) InRepoConfigEnabled(identifier string) bool {
...
 		if allowedCluster == clusterName {
 			return true
 		}
 	}
 
-	identifierSlashSplit := strings.Split(repoIdentifier, "/")
-	if len(identifierSlashSplit) == 2 {
-		for _, allowedCluster := range c.InRepoConfig.AllowedClusters[identifierSlashSplit[0]] {
+	// If err then it cannot be split.
+	if err == nil {
```
chaodaiG, line 259: same, error check should be earlier

chaodaiG (changes requested):

prow/config/config.go
```diff
@@ -216,14 +216,28 @@ type InRepoConfig struct {
...
+	return s[0], s[1], nil
+}
+
 // InRepoConfigEnabled returns whether InRepoConfig is enabled for a given repository.
 func (c *Config) InRepoConfigEnabled(identifier string) bool {
+
+	org, repo, err := SplitRepoName(identifier)
 	if c.InRepoConfig.Enabled[identifier] != nil {
-		return *c.InRepoConfig.Enabled[identifier]
+		return *c.InRepoConfig.Enabled[strings.Join([]string{org, repo}, "/")]
```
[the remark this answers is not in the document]
mpherman2: Fixed, and added more unit tests to capture. I do assume that https:// and http:// will not appear in the config. I will make sure to document this clearly

prow/config/config.go
mpherman2: fixed

prow/config/config.go
mpherman2: done

prow/config/config.go
```diff
@@ -216,14 +216,39 @@ type InRepoConfig struct {
...
+
+func SplitRepoName(repo string) (string, string, error) {
+	// Normalize repo name to remove http:// or https://, this is the case for some
+	// of the gerrit instances.
+	repo = trimRepoPrefix(repo)
+	s := strings.SplitN(repo, "/", 2)
+	if len(s) != 2 {
+		return "", "", fmt.Errorf("repo %s cannot be split into org/repo", repo)
+	}
+	return s[0], s[1], nil
```
listx, line 233: Maybe also check that s[0] and s[1] are valid-looking strings? "valid-looking" would be:

1. non-empty
2. begins/ends with alphanumeric chars

The alphanumeric check might be overkill, but an emptiness check should be straightforward enough.

prow/config/config.go
```diff
@@ -216,14 +216,39 @@ type InRepoConfig struct {
...
+	// SplitRepoName trims http:// and https:// from beginning of org name
+	split := false
+	org, _, err := SplitRepoName(identifier)
+	if err == nil {
+		split = true
 	}
-	identifierSlashSplit := strings.Split(identifier, "/")
-	if len(identifierSlashSplit) == 2 && c.InRepoConfig.Enabled[identifierSlashSplit[0]] != nil {
-		return *c.InRepoConfig.Enabled[identifierSlashSplit[0]]
+	normalizedIdentifier := trimRepoPrefix(identifier)
```
listx, line 245: Why not do this before the call to `SplitRepoName`, and then pass it along to `SplitRepoName`? That way you could avoid calling `trimRepoPrefix` twice.

prow/config/config.go
```diff
@@ -216,14 +216,39 @@ type InRepoConfig struct {
...
 // InRepoConfigEnabled returns whether InRepoConfig is enabled for a given repository.
+// Assumes that config will not include http:// or https://
 func (c *Config) InRepoConfigEnabled(identifier string) bool {
-	if c.InRepoConfig.Enabled[identifier] != nil {
-		return *c.InRepoConfig.Enabled[identifier]
+	// SplitRepoName trims http:// and https:// from beginning of org name
+	split := false
+	org, _, err := SplitRepoName(identifier)
+	if err == nil {
+		split = true
```
listx, line 243: I wonder if we can do away with this `split` boolean as it makes the code a bit harder to read, as we have to keep track of that state to reason about the rest of the code.

But also, isn't this a bug? If we return a non-nil error, then `org` is the empty string, yet we still proceed with the rest of this logic.

prow/config/config.go
mpherman2: trimRepoPrefix is still required in SplitRepoName as that function is used elsewhere, so we will need both calls anyway.

prow/config/config.go
mpherman2: Theoretically we do not need the split boolean and we can just use err==nil instead of split==true, but I thought that would be harder to understand.

listx: > Below we do not use org if split is false.

Shouldn't it be, "We do not use org if it is empty"?

In general I am suspicious of variables that have an exact 1:1 mapping (here, if split == true, then err == nil, if split == false then err == non-nil; and these mappings will always be the same).

Maybe I'm over-thinking this, curious to see what others think. @cjwagner @chaodaiG

mpherman2: Oh I actually totally agree. I was a little blinded by the error, but just checking if org != "" is much easier.

prow/config/config.go
```diff
@@ -216,14 +216,35 @@ type InRepoConfig struct {
...
+	return s[0], s[1], nil
+}
+
 // InRepoConfigEnabled returns whether InRepoConfig is enabled for a given repository.
+// Assumes that config will not include http:// or https://
 func (c *Config) InRepoConfigEnabled(identifier string) bool {
-	if c.InRepoConfig.Enabled[identifier] != nil {
-		return *c.InRepoConfig.Enabled[identifier]
+	// SplitRepoName trims http:// and https:// from beginning of org name
+	org, _, _ := SplitRepoName(identifier)
```
chaodaiG, line 240: should document if it is intended to ignore the error, also would be nice to move this closer to where it's used, for example at line 246:
```
if org, _, _ := SplitRepoName(identifier); org != "" && c.InRepoConfig.Enabled[org] != nil {
```

prow/config/config.go
```diff
@@ -233,15 +254,17 @@ func (c *Config) InRepoConfigEnabled(identifier string) bool {
 
 // InRepoConfigAllowsCluster determines if a given cluster may be used for a given repository
 func (c *Config) InRepoConfigAllowsCluster(clusterName, repoIdentifier string) bool {
-	for _, allowedCluster := range c.InRepoConfig.AllowedClusters[repoIdentifier] {
+	org, _, _ := SplitRepoName(repoIdentifier)
```
chaodaiG, line 257: same as above

prow/config/config.go
mpherman2: Fixed

prow/config/config.go
mpherman2: Fixed

prow/config/config.go
```diff
@@ -216,14 +216,36 @@ type InRepoConfig struct {
...
+	if c.InRepoConfig.Enabled[normalizedIdentifier] != nil {
+		return *c.InRepoConfig.Enabled[normalizedIdentifier]
 	}
-	identifierSlashSplit := strings.Split(identifier, "/")
-	if len(identifierSlashSplit) == 2 && c.InRepoConfig.Enabled[identifierSlashSplit[0]] != nil {
-		return *c.InRepoConfig.Enabled[identifierSlashSplit[0]]
+
+	// Errors if failed to split. We are ignoring this and just checking if org != "" instead
+	org, _, _ := SplitRepoName(identifier)
+	if org != "" && c.InRepoConfig.Enabled[org] != nil {
```
chaodaiG, line 247: org is not used outside of this if/else condition, should combine
```suggestion
	if org, _, _ := SplitRepoName(identifier); org != "" && c.InRepoConfig.Enabled[org] != nil {
```

mpherman2: Good catch. Fixed

prow/config/config.go
```diff
@@ -232,16 +254,19 @@ func (c *Config) InRepoConfigEnabled(identifier string) bool {
...
 			return true
 		}
 	}
 
-	identifierSlashSplit := strings.Split(repoIdentifier, "/")
-	if len(identifierSlashSplit) == 2 {
-		for _, allowedCluster := range c.InRepoConfig.AllowedClusters[identifierSlashSplit[0]] {
+	// Errors if failed to split. We are ignoring this and just checking if org != "" instead
+	org, _, _ := SplitRepoName(repoIdentifier)
+	if org != "" {
```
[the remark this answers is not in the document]
mpherman2: fixed

chaodaiG (approved):

PR #24911, opened 2022-01-18
Delete all dockershim presubmits

### LINKED ISSUE #24620: Move docker-based node presubmits to CRI

SIG Node has a number of presubmit jobs that need to be migrated off Docker.

I suggest we declare bankruptcy, remove all Docker-based jobs, and add jobs back under the CRI as needed.

/sig node

matthyx:
Relates to #24620 
/cc @ehashman


config/jobs/kubernetes/sig-release/release-branch-jobs/1.20.yaml
```diff
@@ -1067,43 +1067,6 @@ presubmits:
...
             memory: 6Gi
-  - always_run: true
-    branches:
-    - release-1.20
-    cluster: k8s-infra-prow-build
-    labels:
-      preset-k8s-ssh: "true"
-      preset-service-account: "true"
-    max_concurrency: 12
-    name: pull-kubernetes-node-e2e
```
matthyx, line 1078: Not sure it's good to remove it from 1.20 and 1.21 but if not config tests fail.

ehashman (changes requested):
Please don't delete any of the presubmits on previous release branches, this should only apply to master.

matthyx:
> Please don't delete any of the presubmits on previous release branches, this should only apply to master.

Yeah this is what I thought... I need to find a way to make tests pass if I keep them in the 1.2x branches.
I think deleting them from 1.23 is OK since we removed dockershim at that point?

ehashman:
No, 1.23 shipped with dockershim. Only 1.24+ is affected

matthyx:
> No, 1.23 shipped with dockershim. Only 1.24+ is affected

Ok, I will correct tomorrow and see how to make the tests pass. Thanks for your help Elana :-)

config/testgrids/config.yaml
```diff
@@ -77,6 +77,18 @@ test_groups:
 - name: test_pull_request_crio_e2e_rhel
   gcs_prefix: origin-federated-results/pr-logs/directory/test_pull_request_crio_e2e_rhel
 
+# dockershim
+- name: pull-kubernetes-node-e2e
+  gcs_prefix: kubernetes-jenkins/pr-logs/directory/pull-kubernetes-node-e2e
+- name: pull-kubernetes-node-e2e-alpha
+  gcs_prefix: skip
```
matthyx, line 84: need to set to non-existing prefix for config/tests/testgrids/config_test.go:TestKubernetesProwInstanceJobsMustHaveMatchingTestgridEntries

dims:
/approve

will leave `/lgtm` to either @ehashman or @SergeyKanzhelev

config/testgrids/config.yaml
ehashman: This is weird, we're removing them so why would we have to add these to testgrid configs?

matthyx: If you want I could take some time to document it in a comment...

BenTheElder: pretty sure the testgrid annotations on the prowjobs could be moved to the most recent branch that still has these jobs (most recent so this doesn't come back as old branches are removed from config)

this part:
```yaml
    annotations:
      testgrid-dashboards: sig-node-presubmits
      testgrid-tab-name: pr-node-kubelet-serial-kubetest2
```

matthyx: Sorry @BenTheElder I don't understand... test-infra needs to be always on the master branch, but the jobs corresponding to k/k 1.21 are defined in ‎config/jobs/kubernetes/sig-release/release-branch-jobs/1.21.yaml‎ for example, aren't they?
So I need to define the testgrid entries somewhere on test-infra master, no?

ehashman (commented):
/lgtm

Issue #24141, opened 2021-10-27
Merge type "merge" disallowed by repo settings


Jeffwan:
![image](https://user-images.githubusercontent.com/4739316/139007300-837bcd2d-6f99-494c-8898-51421fad9ce6.png)

**What happened**:
tide can not merge the PR because `Merge type "merge" disallowed by repo settings`. The weird thing is we does allow all merge option in the repo setting.

```
{"client":"github","component":"tide","controller":"status-update","file":"prow/github/client.go:889","func":"k8s.io/test-infra/prow/github.(*client).log","level":"info","msg":"CreateStatus(xxxxx, xxxxxx, 14b1d33fcaf7086cb68354a433c8ee7a0f55e57f, {error  Not mergeable. Merge type \"merge\" disallowed by repo settings tide})","severity":"info","time":"2021-10-26T09:25:12Z"}
```

![image](https://user-images.githubusercontent.com/4739316/139007234-580cae69-bc05-4d7d-863d-6142f186d9e0.png)

**What you expected to happen**:
tide should merge the PR and this should be allowed.

**How to reproduce it (as minimally and precisely as possible)**:
v20211022-effa07f2d2

**Please provide links to example occurrences, if any**:

**Anything else we need to know?**:

RinkiyaKeDad:
+1 - we've been facing the same issue. Here's the [slack thread](https://kubernetes.slack.com/archives/CDECRSC5U/p1635260608036300) where we were discussing this.

Jeffwan:
> +1 - we've been facing the same issue. Here's the [slack thread](https://kubernetes.slack.com/archives/CDECRSC5U/p1635260608036300) where we were discussing this.

Thanks for pointing to the discussion. Seems it's due to some recent change. Let me debug using same way as they suggest and report status back. I will also try earlier version to see what happens.

ludydoo:
I am facing the same issue using a github app. 
If i GET a repo from github API using a personal access token, i get the allows_merge, allows_squash, etc.
If i GET a repo from github API usign a github installation token, i do not get the fields.
There's a difference between the two APIs ...

Jeffwan:
I think I figure out the problem. 
When we enable github app, auth header is never set in the HTTP request. Based on the docs here, https://docs.github.com/en/rest/reference/repos#get-a-repository, we know that anonymous request won't give us fields like `allow_merge_commit`,`delete_branch_on_merge` etc. I think this is the major problem

[...]

PR #24917, opened 2022-01-19
Use Token from env even github app is used

Jeffwan:
This is not a solid PR to address https://github.com/kubernetes/test-infra/issues/24141. 

I am not that familiar with Github App Auth, I file this PR as a workaround to keep using Github Access Token for GithubApp setup. Otherwise, seem some clients miss Auth Header and won't get fields like `allow_merge_commit`, `delete_branch_on_merge` etc. 

The detail analysis is in  https://github.com/kubernetes/test-infra/issues/24141#issuecomment-1016042616

### DIFF
--- a/prow/github/client.go
+++ b/prow/github/client.go
@@ -29,6 +29,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"net/url"
+	"os"
 	"regexp"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


prow/github/client.go
```diff
@@ -687,8 +688,15 @@ func NewClientFromOptions(fields logrus.Fields, options ClientOptions) (TokenGen
 	options = options.Default()
 
 	// Will be nil if github app authentication is used
+	// Hack: We manually retrieve personal access token to support some operation missing GitHubApp jwtToken.
+	// This helps avoid situation that tide calls github api anonymously and leads to fields missing
 	if options.GetToken == nil {
-		options.GetToken = func() []byte { return nil }
+		personalAccessToken := os.Getenv("GITHUB_TOKEN")
```
alvaroaleman, line 694: /hold

We should not mix apps aush and PAT auth. There might be a bug in the area, but this is not a reasonable solution.

Jeffwan: I agree. Is there a better way that I can help with from your perspective?

Jeffwan: Technically, I feel app temp token should be generated and renewed, so we can return that credential in `GetToken`. I probably need more time to dig into the code and get familiar with app authentication

alvaroaleman: As mentioned in https://github.com/kubernetes/test-infra/issues/24141#issuecomment-1016624541, the issue is likely that your app has insufficient permissions, it needs `Contents: Read & Write`. Apps auth definitely does get an Authorization Header set, it just happens on a different layer in a http transport, as it needs to inspect the request in order to decide what token to set.

Jeffwan:
As we talked in https://github.com/kubernetes/test-infra/issues/24141#issuecomment-1016624541, this PR should be closed

PR #24927, opened 2022-01-19
Use prow/flagutil to replace old GitHub client

bear-redhat:


prow/flagutil/github.go
```diff
@@ -41,8 +41,8 @@ import (
 // Set AllowDirectAccess to be true if you want to suppress warnings on direct github access (without ghproxy).
 type GitHubOptions struct {
 	Host              string
-	endpoint          Strings
```
stevekuznetsov, line 44: These are not exposed on purpose. If you're not deleting the old client code for now, perhaps let's not do the translation/compat layer as you have and just use the old client when requested. Then, we can be certain that a user has no user-facing change until they change their flags, which makes the migration less surprising.

bear-redhat: We are trying to keep backward-compatibility for old command line arguments. In order to do so, we need to ensure the old flag (e.g. `--endpoint`) and their counterparts (e.g. `--github-endpoint`) are exclusive. We can check if old flag has been specified by checking if `o.endpoint` is same as default value, but we are unable to do so for new flag without exposing `github.Endpoint`. Maybe there are other ways we can detect if new flags are specified...?

stevekuznetsov: > Maybe there are other ways we can detect if new flags are specified...?

Yes, please see the `flag` package.

stevekuznetsov: You could also simply bind the flags to different varables.

bear-redhat: I have modified the implementation to check the presences of parameters through `flag.Visit()`, which allow us to keep `prow/flagutil/github.go` as-is.

stevekuznetsov (approved):

PR #24929, opened 2022-01-19
Removed more Windows flags in the PD CSI Driver presubmit config to make it like the linux job

mauriciopoppe:
It looks like the `decorate` or the `branches` params don't go well with the new job structure, made it similar to the linux one.

/cc @jingxu97


jingxu97:
/lgtm
/approve

but I am not what is the default timeout after removing it. It might not enough for windows job

PR #24932, opened 2022-01-20
Runs integration-test on default branch runs only

srm09:
This change edits the `integration-test` job to run only for PRs against the default branch.


yastij (commented):
/approve
/lgtm

PR #24933, opened 2022-01-20
Make Sub include a "https://" prefix when looking up presubmit/postsubmit jobs for gerrit.

cjwagner:
Currently users wishing to trigger presubmit or postsubmit jobs for gerrit via pubsub must incorrectly prepend `https://` to the `refs.org` field in order for Sub to be able to find the job's config. This allows the job to be triggered, but the incorrect `refs.org` value causes problems elsewhere (e.g. links on Prow's front end).

This PR updates Sub to prepend the `https://` during the config lookup if the job is associated with a gerrit revision.

/assign @chaodaiG
/cc @alvaroaleman

### DIFF
--- a/prow/pubsub/subscriber/subscriber.go
+++ b/prow/pubsub/subscriber/subscriber.go
@@ -34,6 +34,7 @@ import (
 	prowapi "k8s.io/test-infra/prow/apis/prowjobs/v1"
 	v1 "k8s.io/test-infra/prow/apis/prowjobs/v1"
 	"k8s.io/test-infra/prow/config"
+	"k8s.io/test-infra/prow/gerrit/client"
 	"k8s.io/test-infra/prow/pjutil"
 )
 
@@ -77,6 +78,11 @@ func (pe *ProwJobEvent) FromPayload(data []byte) error {
 
 // ToMessage generates a PubSub Message from a ProwJobEvent.
 func (pe *ProwJobEvent) ToMessage() (*pubsub.Message, error) {
+	return pe.ToMessageOfType(periodicProwJobEvent)
+}
+
+// ToMessage generates a PubSub Message from a ProwJobEvent.
+func (pe *ProwJobEvent) ToMessageOfType(t string) (*pubsub.Message, error) {
 	data, err := json.Marshal(pe)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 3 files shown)


prow/pubsub/subscriber/subscriber.go
```diff
@@ -199,6 +205,12 @@ func (prh *presubmitJobHandler) getProwJobSpec(cfg prowCfgClient, pc *config.InR
 	var presubmitJob *config.Presubmit
 	org, repo, branch := refs.Org, refs.Repo, refs.BaseRef
 	orgRepo := org + "/" + repo
+	// Add "https://" prefix to orgRepo if this is a gerrit job.
+	// (Unfortunately gerrit jobs use the full repo URL as the identifier.)
+	prefix := "https://"
+	if pe.Labels[client.GerritRevision] != "" && !strings.HasPrefix(orgRepo, prefix) {
```
chaodaiG, line 211: while I like the idea to explicitly check whether it's a gerrit job or not, are we sure that this label is reliable enough? In another word, if a user forgot to set this label in their job, the response they got would be something like `failed to locate prow job`, instead of `gerrit-revision label must be provided`.

How about try looking up prow jobs with `<orgRepo>` first, if not found then try `https://<orgRepo>`?

cjwagner: > are we sure that this label is reliable enough?

I think it is, it is not a label that users would configure on a job, this is set by the gerrit adapter or in the pubsub trigger message. Also reporting logic seems to use this as a differentiation already: https://github.com/kubernetes/test-infra/blob/e267164240be4d7b4d11cfe9258668687514a48c/prow/crier/reporters/gerrit/reporter.go#L189-L195

I was originally going to check both like you are suggesting, but I didn't want gerrit jobs to be able to be configured without the `https://` prefix or GitHub jobs to be configured with it. Its not great that gerrit jobs currently use the wrong format, but at least we have consistent formatting per platform right now. If we allow either format for either platform existing assumptions may be broken.

chaodaiG: I see. In this case I think we would make it clear in our documentations that if this is a gerrit job, it's required to set this label, for example at https://github.com/kubernetes/test-infra/tree/master/prow/cmd/sub#presubmit-prow-jobs

chaodaiG (approved):
/hold
feel free address my documentation suggestion in a separate PR

cjwagner:
/hold cancel
> feel free address my documentation suggestion in a separate PR

I'll do that now.

PR #24936, opened 2022-01-20
fix: containerd was installed under /home/containerd/usr/local/bin/containerd

pacoxu:
```--container-runtime-process-name=/bin/containerd```
will fit with bot `/usr/bin/containerd` and `/home/containerd/usr/local/bin/containerd`

Fixes https://github.com/kubernetes/kubernetes/issues/107062

According to https://github.com/kubernetes/kubernetes/issues/107062#issuecomment-1009446893.

### DIFF
--- a/config/jobs/kubernetes/sig-node/containerd.yaml
+++ b/config/jobs/kubernetes/sig-node/containerd.yaml
@@ -1160,7 +1160,7 @@ periodics:
       - --node-args=--image-config-file=/workspace/test-infra/jobs/e2e_node/containerd/containerd-master/cgroupv2/image-config-cgroupv2-serial.yaml
       - --deployment=node
       - --gcp-zone=us-west1-b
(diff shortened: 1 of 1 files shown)


bobbypage:
/lgtm

I tested locally with `/home/containerd/usr/local/bin/containerd` and it fixed kubernetes/kubernetes#107062 so I think this should be more generic approach since it should match both  `/usr/bin/containerd` and `/home/containerd/usr/local/bin/containerd` as you mentioned.

SergeyKanzhelev (approved):
/lgtm
/approve

this fix is good to go, you can unhold to make tests pass. 

/hold

Should we just use `containerd` and do it for all tests, not just this one? Please unhold if you want to send a separate PR for this

pacoxu:
> Should we just use `containerd` and do it for all tests, not just this one? Please unhold if you want to send a separate PR for this

I will send a separate PR if other tests have similar issues.

Issue #24832, opened 2022-01-11
Filtering meaningful handler executions from metrics in prow plugins


jmguzik:
**What would you like to be added:**
I would like to have the possibility to optionally eliminate from prow hook plugins metrics (eg for hook  LGTM) the obvious return cases that do not provide much valuable information about the transition made (eg to the GH API).

**Example:**
If the execution lands in the below lines and returns `nil`, success will be reported but no actual job was done by the plugin handler.
https://github.com/kubernetes/test-infra/blob/ff46c2790d6d0f795a5aacaa96ecd307da230605/prow/plugins/lgtm/lgtm.go#L191-L204

I tried first to approach the problem by introducing metrics that will wrap actual transaction work. In the discussed example that would be the function handle:
https://github.com/kubernetes/test-infra/blob/ff46c2790d6d0f795a5aacaa96ecd307da230605/prow/plugins/lgtm/lgtm.go#L207

As @alvaroaleman correctly pointed out in the PR comment https://github.com/kubernetes/test-infra/pull/24825#issuecomment-1008936324 there are existing metrics, but in a different place:
https://github.com/kubernetes/test-infra/blob/0595b59c63b16e0a713d83cb486144032fc41f05/prow/githubeventserver/metrics.go#L35-L44

I would like to use them, but a modification of the interface between hook and plugin handlers would be needed. The plugin handler should inform hook that there was no meaningful work done for the handled case. The easiest way to achieve this would be to filter using a special type of error, which will basically say `NoMeaningfullWork`, instead of `nil`. Probably there are other alternatives to the solution as well.

**Why is this needed**:
Some of our LGTM and/or Approve executions take more than 10 mins. Observing the real transactions (the ones that go to the API at least once) would be a nice way to start working towards the solution. I would like to reuse existing metrics, at least for the histogram and successes/failures, that is why I am proposing the modification.

_**I am assigning only to get a potential opinion.**_
/assign @petr-muller @alvaroaleman @jmguzik

petr-muller:
Adding a motivating example of the problem that we want to address. Let's say we are interested in `lgtm` plugin metrics. The plugin registers for comment events so our metrics suffer from the following:

1. 1000 comments across all PRs, issues etc
2. Only two of them say /lgtm or /lgtm cancel
3. 998 comments get "handled" by just running a  regex against the comment and deciding it is not an interesting comment. 2 comments with /lgtm both get rate limited and take 10 minutes, one of them fails
[...]

PR #24938, opened 2022-01-20
Modify hook/plugins interface to include status as a returned type 

jmguzik:
Sketch to push forward #24832 by following concrete example.

EDIT after the review:
This commit adds plugins.Status struct as the return for IssueComment,
PullRequest and Review actions. Currently, plugins.Status has only
one field: tookAction which is used to indicate whether there was
meaningful work done in the plugin (eg API call, jira call, etc).


jmguzik:
Ok, so it's just a proposition to get #24832 going. Do not worry about the naming etc. I just want to show the proposed direction which is: to add an additional interface and handle more extended status return to not change everything everywhere (as we have problems with only a limited number of plugins and actions).
@alvaroaleman 
/cc @petr-muller

prow/plugins/lgtm/lgtm.go
```diff
@@ -176,7 +176,7 @@ func handlePullRequestReviewEvent(pc plugins.Agent, e github.ReviewEvent) error
 	return handlePullRequestReview(pc.GitHubClient, pc.PluginConfig, pc.OwnersClient, pc.Logger, cp, e)
 }
 
-func handleGenericComment(gc githubClient, config *plugins.Configuration, ownersClient repoowners.Interface, log *logrus.Entry, cp commentPruner, e github.GenericCommentEvent) error {
+func handleGenericComment(gc githubClient, config *plugins.Configuration, ownersClient repoowners.Interface, log *logrus.Entry, cp commentPruner, e github.GenericCommentEvent) plugins.Status {
```
alvaroaleman, line 179: errors are always returned as distinct values and not as part of a struct. IMHO it would also make sense to use a `tookAction` boolean, because allowing an arbitrary string means that plugins will likely invent all kind of possible descriptions of what they did or did not do

alvaroaleman: Also not a fan of calling it `ExternalApiCall` because I assume there are plugins that make api calls to figure out if they need to do something. Maybe the definition of `tookAction` would be "Mutated any external state"?

jmguzik: Regarding errors you are right, I'm coming from the C++ world and there was no multiple returns... I have to get used to it, thanks.
`tookAction` is a fine name. I'll just rename it. I would still keep it in a struct, even if there is only one value to have it extendable in the future. What do you think?
I am not exactly sure what you mean by "allowing an arbitrary string". I think map params coded as labels for Prometheus have to be strings. We can have "true" value, "false" value but we should also indicate that param is unused or unknown (for the "classic" version of handlers). Returning the "false" string value would not be correct, as actually, we do not know whether `tookAction` happened.
Alternative to that, instead of extending the interface by another function version would actually be changing the interface, but that will affect the whole plugin infrastructure.

prow/plugins/plugins.go
```diff
@@ -160,6 +168,15 @@ func RegisterGenericCommentHandler(name string, fn GenericCommentHandler, help H
 	genericCommentHandlers[name] = fn
 }
 
+// GenericCommentHandler defines the function contract for a github.GenericCommentEvent handler.
```
petr-muller, line 171: nit: Fixup the names in the godoc

Also, I think there should be some place that explains how are "Extended" variants different from the normal ones.,

prow/plugins/plugins.go
```diff
@@ -160,6 +168,15 @@ func RegisterGenericCommentHandler(name string, fn GenericCommentHandler, help H
...
+type ExtendedGenericCommentHandler func(Agent, github.GenericCommentEvent) Status
+
+// RegisterExtendedGenericCommentHandler registers a plugin's github.GenericCommentEvent handler.
+func RegisterExtendedGenericCommentHandler(name string, fn ExtendedGenericCommentHandler, help HelpProvider) {
```
petr-muller, line 175: Maybe we would want to (somehow) prevent plugins from registering both a normal and extended handler?

prow/plugins/lgtm/lgtm.go
petr-muller: > tookAction is a fine name. I'll just rename it. I would still keep it in a struct, even if there is only one value to have it extendable in the future. What do you think?

Personally I think using a struct for extensibility is a good idea

petr-muller (commented):
Added some comments in addition to what Alvaro said

prow/plugins/lgtm/lgtm.go
alvaroaleman: Yes, the struct is fine (modulo the embedded error). What confused me is how this gets mapped to the prom label value. Could we maybe just set that to the empty string if we don't know and true/false otherwise?

prow/plugins/plugins.go
jmguzik: updated the comment and function signature

prow/plugins/plugins.go
jmguzik: Prevented by doing the check if the latter is added to extended/not-extended map. If yes, the error is logged and the handler is not added.

Any additional move would require an interface check. Please let me know if that's enough.

prow/plugins/lgtm/lgtm.go
jmguzik: Renamed to `tookAction` and moved `error` out of the struct. Also, reported empty string and true/false otherwise.

jmguzik:
`pull-test-infra-bazel ` is failing due to the inconsistent checksum error which is beyond my control and probably needs to be solved externally. Otherwise, this PR is ready after the review to be re-reviewed.

prow/hook/events.go
```diff
@@ -90,7 +90,7 @@ func (s *Server) handleReviewEvent(l *logrus.Entry, re github.ReviewEvent) {
 				re.PullRequest.Number,
 			)
 			start := time.Now()
-			labels := prometheus.Labels{"event_type": l.Data[eventTypeField].(string), "action": string(re.Action), "plugin": p}
+			labels := prometheus.Labels{"event_type": l.Data[eventTypeField].(string), "action": string(re.Action), "plugin": p, "external_call": ""}
```
alvaroaleman, line 93: Lets also name this `took_action`

prow/plugins/plugins.go
alvaroaleman: Do we plan to eventually move all plugins to be an `ExtendedGenericCommentHandler`? Maybe we can rename the existing registerer to `Legacy` to clarify that it shouldn't be used for new plugins?

prow/hook/events.go
jmguzik: Ups, missed thanks!

prow/plugins/plugins.go
jmguzik: I don't think so. It was introduced to add additional metrics for heavy plugins like `lgtm` and `approve`. I do not see other clear candidates to have the interface extended to have this information. I don't know though, it's my opinion.

prow/plugins/lgtm/lgtm.go
```diff
@@ -61,7 +61,7 @@ type commentPruner interface {
 }
 
 func init() {
-	plugins.RegisterGenericCommentHandler(PluginName, handleGenericCommentEvent, helpProvider)
```
stevekuznetsov, line 64: Plugins are a core Prow thing. I would be extremely surprised if anyone outside of the Prow core were using this interface, and this is not exposed as a reusable library. Why not just refactor the signature? If you want you could plumb down e.g. a `context` to decorate with values.

prow/plugins/lgtm/lgtm.go
```diff
@@ -201,11 +202,13 @@ func handleGenericComment(gc githubClient, config *plugins.Configuration, owners
...
 		wantLGTM = false
 	} else {
-		return nil
+		return status, nil
 	}
 
 	// use common handler to do the rest
-	return handle(wantLGTM, config, ownersClient, rc, gc, log, cp)
+	err := handle(wantLGTM, config, ownersClient, rc, gc, log, cp)
+	status.TookAction = true
```
stevekuznetsov, line 210: Ideally we wouldn't need the author to remember to do this correctly in every single case. Since we're creating a new plugin agent and sub-classing the clients etc inside of it, could we not detect that the GitHub client we gave to a specific plugin handler for a given event was used for something?

jmguzik: Does that mean return status from `handle` or else?

prow/plugins/lgtm/lgtm.go
jmguzik: I wanted to avoid full refactor as that will be used most probably in two plugins: `lgtm` and `approve`, but if you insist I can do that. I would keep the status object though, it seems to be more expressive.

jmguzik: Although keeping bool in a whole refactored environment will most likely mean losing one state: not used ("") in metrics, so I don't know, probably pointer could be the solution.

jmguzik:
@stevekuznetsov @alvaroaleman @petr-muller before I change everything everywhere following Steve's comment I have to ask if such signature:
```
type GenericCommentHandler func(Agent, github.GenericCommentEvent) (Status*, error)
```
would be accepted. `*` to have the `nil` available for not used `took_action` metrics property.

stevekuznetsov:
If we think there's value in returning a `struct` for future-proofing this call, make the field a pointer (`tookAction *bool`) instead of making the `struct` a pointer, or when a new field is added to the struct you're going to have to have complex logic to determine which of the fields is or is not intentionally set. If we're not explicitly using the structure to future-proof this, just add a `*bool` return.

jmguzik:
> If we think there's value in returning a `struct` for future-proofing this call, make the field a pointer (`tookAction *bool`) instead of making the `struct` a pointer,

`tookAction *bool` sounds good, thanks.

petr-muller:
Ugh, from the reviewer's perspective it would really help if this was split into at least two commits: one that changes the interfaces (because that one is mostly mechanical but is huge) and one that adds the new behavior (`TookAction()` and the processing of the status in the handlers' callsite) which is more interesting to review.

jmguzik:
> Ugh, from the reviewer's perspective it would really help if this was split into at least two commits

Yeah, I do not know how to do this easily right now. It was also not done this way.

prow/plugins/plugins.go
```diff
@@ -64,6 +64,25 @@ var (
...
+	s.tookAction = true
+}
+
+// GetAction returns a string indicating whether the action was taken
+func (s *Status) GetAction() string {
+	if s.tookAction {
+		return "true"
+	}
+	return "false"
+}
```
petr-muller, line 84: I'm not a Go guru myself, so take with a grain of salt: setters/getters are not that common in Go, but when they are present, the convention is that `TookAction` would be the getter and the setter would have a `Set`/`Record`/whatever name (https://go.dev/doc/effective_go#Getters)

petr-muller (commented):
I know this will sound insane after all the effort invested into this change (do not take this comment as my requirement to undo it, I'm just writing some feedback), but I wonder if trying to add the `TookAction` calls into all plugins is the right approach here. Of course we needed to change the interfaces and that made the change large, but I think that we could change the interface, implement the `TookAction` calls in one or two interesting plugins (also, we'd add tests for that behavior), and if the thing proved to be useful, we could extend the remaining plugins (and also get them tested, review whether the `TookAction` calls are placed correctly etc).

The sheer size here makes both reviewing (I think that the `TookAction` placements are actually review-worthy as there may be multiple good options, but also opportunities to miss some) and requiring the behavior to be tested almost inappropriate ;)

jmguzik:
`TookAction()` is mostly invoked after the first github client call in the handler (analyzed also nested examples). As I care mostly about actions in `lgtm` and `approve` for now, I have modified handlers that are used in these two plugins: `GenericCommentEvent`, `ReviewEvent` and `PullRequestEvent`. Comments are spread throughout the plugins so I changed handlers in all the plugins.

By the way, @stevekuznetsov I resigned from `tookAction *bool` and used normal `tookAction bool` since I there was a need to modify every handler manually anyway.

There are some exceptions from above "rule" where `TookAction()` was used. Sometimes, where there was a image getter (eg `cat` plugin), jira call or bugzilla call I also used there `tookAction` since it's meaningful work.

@petr-muller I am sorry, I don't know how to split this easily without repeating the work. I edited and tested plugins one by one and some code editions are valid for interface change and `tookAction()` call also. I leave this for now as it is.

prow/plugins/plugins.go
jmguzik: I know, but I wanted `tookAction` to be private, to have the state of status always false, and if the first occurrence of `action` then change to true and do not modify further. In the end, if action is taken why allow the user to modify it back? It's a more objective approach, please further comment if not needed here.
Regarding the naming, I can give whatever the names are needed but this particular getter is only for `Prometheus`, so maybe I can indicate that in a name?

jmguzik:
> `TookAction` calls into all plugins is the right approach here.

Can't comment if it's right but since I was there, and probably there won't be a second person interested to make this change so I did this "by the way".
Usually `TookAction()` is invoked after a first API call in the handler. Usually, this is GitHub API. Sometimes it's bugzilla call, jira call, or a random/different thing (getting images in `cat` :) ).

prow/plugins/bugzilla/bugzilla.go
```diff
@@ -597,22 +603,28 @@ func handle(e event, gc githubClient, bc bugzilla.Client, options plugins.Bugzil
...
+				return status, comment(response)
 			}
-			return nil
+			return status, nil
 		}
 	}
 	// merges follow a different pattern from the normal validation
 	if e.merged {
-		return handleMerge(e, gc, bc, options, log, allRepos)
+		err := handleMerge(e, gc, bc, options, log, allRepos)
```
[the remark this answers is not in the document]
jmguzik: It was a lot of work to decide where the action was taken, so before I revert all of this to leave the signature (which itself is NOT correct change as the status will always be false for now, so metrics will be corrupted), please confirm that is what you want.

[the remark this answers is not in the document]
jmguzik: Also, if you want to revert all of the plugins, please let me at least keep LGTM and approve using this mechanism for now, as my real purpose is to observe GH API rate limits and can't do this without this change.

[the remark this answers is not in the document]
stevekuznetsov: I think having an unused signature is fine for now, it doesn't corrupt anything - we just don't report data where it is not yet exposed. I do truly think we can be much more formal and abstract about taking action. I can see a small handful of ways where the GitHub client, the k8s client, the Bugzilla client, the Git client, etc - they can report whether or not they did something transparently to the plugin author having to track it and we do not impose brittle and labor-intensive tracking for status, unless the plugin wants to expose something that was *not* a delegate client action.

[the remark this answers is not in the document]
jmguzik: > I can see a small handful of ways where the GitHub client,

I am sorry, maybe it's because Golang it's not my native language, but I do not see a clear, harmless way to do this. Context is implicit and would have to be re-updated every handler execution since it's basically a new case. There could be a function wrapper using reflect which is obscure or interface re-implementation with the wrapper which is also not nice (probably per plugin which is even worse). Finally, other params could be added to the function but it also seems pointless. I maybe lack Golang experience but I miss a clean solution. Do you have a specific candidate in mind?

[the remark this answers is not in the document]
stevekuznetsov: @jmguzik when you inspect `plugins.NewAgent()` you will notice that we are actually creating a number of client objects with lifetimes local to individual invocations of handlers for individual GitHub webhook payloads. For instance, when we use `clientAgent.GitHubClient.WithFields(logger.Data).ForPlugin(plugin)` we decorate the individual GitHub client with logging data (which PR number, event GUID, etc) and with the plugin name, so that we may set the user agent header for calls to GitHub made while handling this particular webhook for this particular plugin, so that we may trace API requests per plugin.

Since we already have clients (for GitHub, Owners, and Bugzilla by the looks) that are individually created for a specific handler call, I think it should be as simple as adding a new member to those `struct`s, setting to to `true` if some API action is taken, and extending the interface they fulfill so that the member may be read after the handler is called in order to provide the correct label to the metric. I'm happy to link to an example commit if you'd like to see what I mean.

Of course such an approach would only handle the "simple" cases of client usage, if we wanted plugin authors to be able to mark "action taken" when no API interaction occurred, we would nevertheless need the function signature changes you made in this PR. However, being able to auto-detect what I think may be the most common cases would surely be nice.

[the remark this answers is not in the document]
stevekuznetsov: @jmguzik I'm sorry - I attempted to make this comment in my [first review](https://github.com/kubernetes/test-infra/pull/24938#discussion_r796889639) as well but that was _also_ nowhere near the first day that you had opened this PR, and nevertheless I was not effective at communicating. My apologies for this.

kofiowu:
May I know what conflict is a t about ?

PR #24939, opened 2022-01-20
Add more details about configuring tide.

Summer-Field:
[Configuring Tide](prow/cmd/tide/config.md) didn't describe some detailed configuration of tide such as `excludeRepos`.

So here I wrote some descriptive words to describe how to configure tide with `excludedRepos` attributes.

### DIFF
--- a/prow/cmd/tide/config.md
+++ b/prow/cmd/tide/config.md
@@ -45,6 +45,7 @@ It can consist of the following dictionary of fields:
 
 * `orgs`: List of queried organizations.
 * `repos`: List of queried repositories.
+* `excludedRepos`: List of ignored repositories.
 * `labels`: List of labels any given PR must posses.
 * `missingLabels`: List of labels any given PR must not posses.
 * `excludedBranches`: List of branches that get excluded when querying the `repos`.
@@ -147,6 +148,8 @@ and do not have `do-not-merge`, `do-not-merge/hold`, `do-not-merge/work-in-progr
 All PRs that conform to the criteria are processed and merged.
 The processing itself can include running jobs (e.g. tests) to verify the PRs are good to go.
 All commits in PRs from `github.com/kubeflow/community` repository are squashed before merging.
+
+For a full list of properties of queries, please refer to [https://github.com/kubernetes/test-infra/blob/27c9a7f2784088c2db5ff133e8a7a1e2eab9ab3f/prow/config/prow-config-documented.yaml#:~:text=meet%20merge%20requirements.-,queries%3A,-%2D%20author%3A%20%27%20%27](/prow/config/prow-config-documented.yaml).
 
 ### Persistent Storage of Action History
 


prow/cmd/tide/config.md
```diff
@@ -44,6 +44,7 @@ Each query corresponds to a set of **open** PRs as candidates for merging.
 It can consist of the following dictionary of fields:
 
 * `orgs`: List of queried organizations.
+  * `excludedRepos`: List of ignored repositories.
```
Jeffwan, line 47: Seems there's extra blanks here? Can you align with other lines

prow/cmd/tide/config.md
```diff
@@ -106,6 +107,17 @@ tide:
   target_url: https://prow.k8s.io/tide
 
   queries:
+  - orgs:
+    - kubeflow
+    excludedRepos:
+    - kubeflow/community
+    - kubeflow/examples
+    labels:
```
Jeffwan, line 115: there's already examples on labels and missingLabels. I think you just need to line 112 and 113. since `kubeflow/community` and `kubeflow/examples` have been used in `repos`, could you exclude `kubeflow/website` as an example?

Jeffwan (changes requested):

prow/cmd/tide/config.md
Summer-Field: Label `excludedRepos` is the sublist of `orgs`, so these extra 2 black is to show words like this:
* `orgs`: List of queried organizations.
  * `excludedRepos`: List of ignored repositories.

prow/cmd/tide/config.md
Summer-Field: You do have a point. I will exclude `kubeflow/website` to delineate which repos won't be queried by tide.
However, line 112 to 120 is to explain one function of `excludedRepos` that you could configure distinguish query condition of specific repositories in `repos` which is different from other repos in the orgs.

Summer-Field: Update, I've add `kubuflow/website` as an example of how tide ignore querying specific repository.

prow/cmd/tide/config.md
Summer-Field: My mistake. It seems that the `orgs` and `excludedRepos` are at the same level. I've correct the mistake...

prow/cmd/tide/config.md
```diff
@@ -141,8 +154,10 @@ tide:
                 - "optional_test"
 ```
 
-**Explanation**: The component starts periodically querying all PRs in `github.com/kubeflow/community` and
-`github.com/kubeflow/examples` repositories that have `lgtm` and `approved` labels set
+**Explanation**: The component starts periodically querying all PRs in `github.com/kubflow` organization excluding `github.com/kubeflow/community`, `github.com/kubeflow/examples` and `github.com/kubeflow/website`repositories that have `lgtm` and `approved` labels set and do not have `do-not-merge` and `do-not-merge/blocked-paths` labels set.
```
alvaroaleman, line 157: Not quite correct, it also keeps quering `kubeflow/community` and `kubeflow/examples` that have lgtm+approve but do not have `do-not-merge`, `do-not-merge/hold`, `do-not-merge/work-in-progress`, `needs-ok-to-test` and `needs-rebase`.

Rather than trying to document all possible attributes here, I would prefer if we just kept the simple example and additionally linked to https://github.com/kubernetes/test-infra/blob/27c9a7f2784088c2db5ff133e8a7a1e2eab9ab3f/prow/config/prow-config-documented.yaml#L1208
for a full list of properties. That document is generated from the code and we check that it stays in sync, so that is the better reference IMHO.

Summer-Field: Thanks a lot for your advice! I'll make some changes later.

alvaroaleman (approved):
thanks!

PR #24944, opened 2022-01-20
Increase context timeout for reporting gerrit prowjobs

mpherman2:
This should hopefully fix the Notification Spam we are seeing on gerrit.

/assign @chaodaiG 
/cc @listx

### DIFF
--- a/prow/crier/reporters/gerrit/reporter.go
+++ b/prow/crier/reporters/gerrit/reporter.go
@@ -412,10 +412,14 @@ func (c *Client) Report(ctx context.Context, logger *logrus.Entry, pj *v1.ProwJo
 	// reported, as the change of previous report state happens only after the
 	// returning of current function from the caller.
 	// Ideally the previous report state should be changed here.
+	// This operation takes a long time when there are a lot of jobs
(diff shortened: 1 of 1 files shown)


chaodaiG (approved):

prow/crier/reporters/gerrit/reporter.go
```diff
@@ -412,10 +412,14 @@ func (c *Client) Report(ctx context.Context, logger *logrus.Entry, pj *v1.ProwJo
 	// reported, as the change of previous report state happens only after the
 	// returning of current function from the caller.
 	// Ideally the previous report state should be changed here.
+	// This operation takes a long time when there are a lot of jobs
+	// in the batch, so we are creating a new context.
+	loopCtx, loopCancel := context.WithTimeout(ctx, 2*time.Minute)
```
listx, line 417: Is `extraTimeoutCtx` a better name for this? Or simply `newCtx`?

cjwagner, line 417: Doesn't this just add an additional time out? When `ctx` is canceled that will cause `loopCtx` to cancel as well.

Issue #24946, opened 2022-01-20
kettle is down again


BenTheElder:
**What happened**:

https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/metrics-kettle/1479915842807271424

> ERROR: table k8s-gubernator:build.all is 6.8 hours old. Max allowed: 6 hours.
> ERROR: table k8s-gubernator:build.week is 6.9 hours old. Max allowed: 6 hours.
> ERROR: table k8s-gubernator:build.day is 6.9 hours old. Max allowed: 6 hours.

**What you expected to happen**:

kettle should run without error and process current results

**How to reproduce it (as minimally and precisely as possible)**:

https://prow.k8s.io/job-history/gs/kubernetes-jenkins/logs/metrics-kettle ?

**Please provide links to example occurrences, if any**:

see above

**Anything else we need to know?**:

https://github.com/kubernetes/test-infra/issues/23135
https://kubernetes.slack.com/archives/C09QZ4DQB/p1642705919011300

BenTheElder:
This appears to be the most recent log for the kettle pod in google.com k8s-gubernator project g8r cluster:

> Info 2022-01-08 08:46:41.721 PST inserting build: gs://istio-circleci/test/439703

So it wedged inserting istio results again?

We should probably stop collecting results for external projects, this is what we hit last time as well https://github.com/kubernetes/test-infra/issues/23135#issuecomment-893015099

BenTheElder:
there is also a patch here that never merged because it fails tests https://github.com/kubernetes/test-infra/pull/23460

BenTheElder:
on the off chance this wedging is a fluke, I went ahead and deleted the pod to restart it (it's a deployment).

BenTheElder:
It has been inserting builds happily, but it will be a long time before it is back up to date, even if nothing goes wrong. I intended to work on this problem, but other things came up, and I can't spare much time for this.

I *suspect* this is unnecessarily serial.

BenTheElder:
... seems building kettle images is also broken https://github.com/kubernetes/test-infra/pull/24948#issuecomment-1018035026

BenTheElder:
> b'Error while reading data, error message: JSON parsing error in row starting at position 681087456: Parser terminated before end of string'

> b'Error while reading data, error message: JSON parsing error in row starting at position 1658689589: Parser terminated before end of string'

[...]

PR #24948, opened 2022-01-20
remove misc buckets from kettle

BenTheElder:
we only need to process big query data for kubernetes jobs. many of these buckets don't even exist anymore.

AFAICT nobody is consuming the big query data for jobs outside of the project, and if they were, they should be using their own deployment anyhow https://github.com/kubernetes/test-infra/issues/24946


BenTheElder:
I suspect we need to update some tests, but with the half-way bazel migration, I'm going to let CI run them for me first inbetween doing some other things 🙃

BenTheElder:
I don't think it's specific to this PR but it seems building kettle is broken. https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-test-infra-push-kettle-canary/1484315304040337408

FYI @chaodaiG since recent python migration 👀

BenTheElder:
see more on https://github.com/kubernetes/test-infra/commit/fce7c9d67efee467c9a358d168335625bd5f7592 statuses, not the only broken post-submit

PR #24951, opened 2022-01-21
upgrade pip before using pip

BenTheElder:
xref: https://github.com/kubernetes/test-infra/pull/24948#issuecomment-1018035026

pretty sure we're failing with `ModuleNotFoundError: No module named 'setuptools_rust'` because pip(3) is too old. we should keep pip up to date.

### DIFF
--- a/kettle/Dockerfile
+++ b/kettle/Dockerfile
@@ -30,7 +30,8 @@ RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime \
     python3-pip \
     && rm -rf /var/lib/apt/lists/*
 
-RUN pip3 install --no-cache-dir requests google-cloud-pubsub==2.3.0 google-cloud-bigquery==2.11.0 influxdb ruamel.yaml==0.16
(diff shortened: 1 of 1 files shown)


chaodaiG:
should we also get to a newer ubuntu version?

chaodaiG (approved):
/hold
in case you would also like to bump the ubuntu version

BenTheElder:
lets do one thing at a time, not sure how fragile this is currently 🙃 
/hold cancel

kettle/Dockerfile
```diff
@@ -30,7 +30,8 @@ RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime \
     python3-pip \
     && rm -rf /var/lib/apt/lists/*
 
-RUN pip3 install --no-cache-dir requests google-cloud-pubsub==2.3.0 google-cloud-bigquery==2.11.0 influxdb ruamel.yaml==0.16
+RUN pip3 install --no-cache-dir --upgrade pip \
```
BenTheElder, line 33: sigh, this should be: `... pip3 && \`

PR #24952, opened 2022-01-21
Remove pytest image

chaodaiG:
It was originally added to run python tests, the plan had changed to use native python image instead. Removing it to avoid confusion

/cc @BenTheElder


BenTheElder (approved):
/lgtm
/approve

Issue #23897, opened 2021-10-05
Run linter tests without bazel


chaodaiG:
**What would you like to be added**:

Run linter tests without bazel, the related bazel rules are:

- verify-pylint
- verify-tslint
- update-tslint
- verify-codegen
- verify-spelling
- update-spelling
- verify-label
- pylint_bin
- verify-gopherage

Things to be considered:

- This will need to co-exist with current bazel rules for a short while, delete linter test targets from bazel once this confirmed working
- Test driven by make rule and bash scripts

**Why is this needed**:

This is part of #23796 

**Recommended testing approach**:
- Make test invokable by `make test`
- Create a PullRequest with change, manually run job by commenting `/test pull-test-infra-test-beta`
- [Optional]: if there is any tool that needs to exist in test image, update image at https://github.com/kubernetes/test-infra/blob/6e0d21ee50d690c8f78239257f588021d2ed57b5/config/jobs/kubernetes/test-infra/test-infra-presubmits.yaml#L145

/sig testing
/help-wanted

chaodaiG:
As far as I can tell all linting targets are defined in https://github.com/kubernetes/test-infra/blob/master/hack/BUILD.bazel, went through the targets and migrated following targets to make rules:

- verify-codegen
- verify-spelling
- verify-labels
- verify-gofmt
- verify-golangci-lint

Remaining:

- verify-pylint. Better to be after https://github.com/kubernetes/test-infra/issues/23830
- verify-tslint. Better to be after https://github.com/kubernetes/test-infra/issues/23829
- verify-bazel. To be deprecated
- verify-boilerplate. Uses python, better to be after https://github.com/kubernetes/test-infra/issues/23830
- verify-deps. WIP https://github.com/kubernetes/test-infra/pull/24194, depends on gazelle deprecation
- verify-gopherage. Uses typescript, better after https://github.com/kubernetes/test-infra/issues/23829

BenTheElder:
typescript progress https://github.com/kubernetes/test-infra/pull/24491

BenTheElder:
updated checklist, tslint was https://github.com/kubernetes/test-infra/pull/24491 and boilerplate was https://github.com/kubernetes/test-infra/pull/24891.

chaodaiG:
After #25373 prow images are built without bazel
/close

PR #24966, opened 2022-01-21
Rollup gopherage typescript bundle with make rule

chaodaiG:
This is an example from reverse engineering rollup_bundle bazel target.

Gopherage typescript bundle is already source code controlled, verify-gopherage script is meant to ensure that the bundled js file is correct, however https://github.com/kubernetes/test-infra/blob/26e57a2169f9f24d8980bdf63c6bb268b3522ea1/hack/verify-gopherage.sh#L36 doesn't really catch error when the generated file doesn't exist, which doesn't really work.

Part of https://github.com/kubernetes/test-infra/issues/23897
/cc @BenTheElder


package.json
```diff
@@ -18,6 +18,7 @@
     "@rollup/plugin-commonjs": "14.0.0",
     "@rollup/plugin-json": "4.1.0",
     "@rollup/plugin-node-resolve": "8.4.0",
+    "rollup-plugin-terser": "7.0.2",
```
chaodaiG, line 21: Failed to discover the version of terser used by bazel, so picked the latest version of this plugin. Will take a further look trying to align as much as possible

gopherage/rollup.config.js
```diff
@@ -0,0 +1,20 @@
+import { nodeResolve } from '@rollup/plugin-node-resolve';
+import commonjs from '@rollup/plugin-commonjs';
+import { terser } from "rollup-plugin-terser";
+
+export default {
+  context: "window",
+  input: './gopherage/cmd/html/static/_output/browser.js',
+  output: [
+    {
```
chaodaiG, line 9: Double output pretty much from https://github.com/kubernetes/test-infra/blob/26e57a2169f9f24d8980bdf63c6bb268b3522ea1/def/ts.bzl#L35

package.json
```diff
@@ -18,6 +18,7 @@
     "@rollup/plugin-commonjs": "14.0.0",
     "@rollup/plugin-json": "4.1.0",
     "@rollup/plugin-node-resolve": "8.4.0",
+    "rollup-plugin-terser": "5.0.0",
```
chaodaiG, line 21: expect `terser 4.4.0` as defined below. Did a binary search https://openbase.com/js/rollup-plugin-terser/versions:

5.1.0 -> terser 4.1.0
5.1.2 -> terser 4.1.0
5.1.3 -> terser 4.1.0
5.2.0 -> terser 4.6.2

The closest one is 5.1.3

package.json
```diff
@@ -18,6 +18,7 @@
     "@rollup/plugin-commonjs": "14.0.0",
     "@rollup/plugin-json": "4.1.0",
     "@rollup/plugin-node-resolve": "8.4.0",
+    "rollup-plugin-terser": "5.1.3",
```
chaodaiG, line 21: expect terser 4.4.0 as defined below. Did a binary search https://openbase.com/js/rollup-plugin-terser/versions:

5.1.0 -> terser 4.1.0
5.1.2 -> terser 4.1.0
5.1.3 -> terser 4.1.0
5.2.0 -> terser 4.6.2

The closest one is 5.1.3

gopherage/Makefile
```diff
@@ -0,0 +1,22 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+include ../Makefile.base.mk
+# Makefile.base.mk requires REPO_ROOT to be set
+REPO_ROOT:=${CURDIR}/..
+
+rollup-js:
```
BenTheElder, line 19: this makefile is a good target for having non-phony "output file depends on rules & input files" (which also allows make to skip running when mtime of output is >= inputs), but we can refactor later.

BenTheElder (approved):
/lgtm
/approve

PR #24967, opened 2022-01-21
pip3 is a pip alias, upgrade pip package, not pip3 (which doesn't exist)

BenTheElder:
3rd time is the charm
/cc @MushuEE


chaodaiG (approved):

BenTheElder:
it built 🎉

Issue #24970, opened 2022-01-21
kubekins-e2e push postsubmit is broken


BenTheElder:
**What happened**:

https://prow.k8s.io/job-history/gs/kubernetes-jenkins/logs/post-test-infra-push-kubekins-e2e

1484549352604045312 |   | 07:32:35Jan 21 2022, 07:32:35 UTC-0800 | 10m14s | FAILURE
1484329400395632640 |   | Jan 20 16:58:48Jan 20 2022, 16:58:48 UTC-0800 | 10m2s | FAILURE
1484216057945133056 |   | Jan 20 09:28:11Jan 20 2022, 09:28:11 UTC-0800 | 8m40s | FAILURE
1483747084703109120 |   | Jan 19 02:24:47Jan 19 2022, 02:24:47 UTC-0800 | 8m42s | FAILURE
1483057404756103168 |   | Jan 17 04:44:09Jan 17 2022, 04:44:09 UTC-0800 | 10m15s | FAILURE
1471894589936766976 |   | Dec 17 09:27:16Dec 17 2021, 09:27:16 UTC-0800 | 9m32s | FAILURE
1470678992095809536 |   | Dec 14 00:56:46Dec 14 2021, 00:56:46 UTC-0800 | 2h15m3s | FAILURE
1469272505712644096 |   | Dec 10 03:47:57Dec 10 2021, 03:47:57 UTC-0800 | 9m18s | FAILURE
1468726168273293312 |   | Dec 08 15:37:01Dec 08 2021, 15:37:01 UTC-0800 | 2h15m2s | FAILURE

**What you expected to happen**:

no persistent failures

**How to reproduce it (as minimally and precisely as possible)**:

**Please provide links to example occurrences, if any**:

**Anything else we need to know?**:

Step #0: /go/pkg/mod/k8s.io/client-go@v0.22.2/plugin/pkg/client/auth/exec/metrics.go:21:2: package io/fs is not in GOROOT (/usr/local/go/src/io/fs)
Finished Step #0
ERROR
ERROR: build step 0 "golang:1.15.15" failed: step exited with non-zero status: 1

BenTheElder:
looks like it's the 1.20 variant that won't build, because we now have client-go that requires newer go. probably the kubetest build is failing? I am already working on #24946 at the moment.

BenTheElder:
[...]

PR #24971, opened 2022-01-21
[WIP] drop unused kubetest --deployment=conformance (and hopefully client-g…

BenTheElder:
…o as a kubetest dependency)

possible fix for #24970 

I don't see evidence of `--deployment=conformance` being used, and there are other, probably much better ways to run the conformance tests anyhow. Some analysis with `gomod` suggests that this is the only package in kubetest pulling in client-go.

We could attempt to rewrite this to shell out to `kubectl`, but I think this is package is basically code anyhow.

Users should look to kubetest2 for non-deprecated tooling.


BenTheElder:
simpler alternative https://github.com/kubernetes/test-infra/pull/24972

PR #24973, opened 2022-01-21
Crier gerrit reporter: don't report pjs that were replaced by newer ones

chaodaiG:
There are cases where these jobs were replaced but the prowjob CR somehow failed to be cleaned up, this job would not be reported as reporter knows to sort by creation time and always report the latest. However if somehow reconciler sees this prowjob CR again it'll say that this job can be reported and causing duplicated report.

Context: noticed that reports came in when an ancient prowjob CR was TTL'ed, which happened more than a single instance.

/cc @cjwagner @listx @mpherman2

### DIFF
--- a/prow/crier/reporters/gerrit/reporter.go
+++ b/prow/crier/reporters/gerrit/reporter.go
@@ -246,16 +246,20 @@ func (c *Client) ShouldReport(ctx context.Context, log *logrus.Entry, pj *v1.Pro
 	patchsetNum := patchsetNumFromPJ(pj)
 
 	// Check all other prowjobs to see whether they agree or not
-	return allPJsAgreeToReport([]string{client.GerritRevision, kube.ProwJobTypeLabel, client.GerritReportLabel}, func(pj *v1.ProwJob) bool {
-		if pj.Status.State == v1.TriggeredState || pj.Status.State == v1.PendingState {
+	return allPJsAgreeToReport([]string{client.GerritRevision, kube.ProwJobTypeLabel, client.GerritReportLabel}, func(otherPj *v1.ProwJob) bool {
+		if otherPj.Status.State == v1.TriggeredState || otherPj.Status.State == v1.PendingState {
 			// other jobs with same label are still running on this revision, skip report
 			log.Info("Other jobs with same label are still running on this revision")
 			return false
 		}
 		return true
-	}) && allPJsAgreeToReport([]string{kube.OrgLabel, kube.RepoLabel, kube.PullLabel}, func(pj *v1.ProwJob) bool {
+	}) && allPJsAgreeToReport([]string{kube.OrgLabel, kube.RepoLabel, kube.PullLabel}, func(otherPj *v1.ProwJob) bool {
+		// This job has duplicate(s) and there are newer one(s)
+		if otherPj.Spec.Job == pj.Spec.Job && otherPj.CreationTimestamp.After(pj.CreationTimestamp.Time) {
+			return false
+		}
 		// Newer patchset exists, skip report
-		return patchsetNumFromPJ(pj) <= patchsetNum
+		return patchsetNumFromPJ(otherPj) <= patchsetNum
 	})
 }
 
@@ -315,20 +319,19 @@ func (c *Client) Report(ctx context.Context, logger *logrus.Entry, pj *v1.ProwJo
 	pjTypeLabel := kube.ProwJobTypeLabel
 	gerritReportLabel := client.GerritReportLabel
 
+	var pjsOnRevisionWithSameLabel v1.ProwJobList
 	var toReportJobs []*v1.ProwJob
 	if pj.ObjectMeta.Labels[gerritReportLabel] == "" && pj.Status.State != v1.AbortedState {
 		toReportJobs = append(toReportJobs, pj)
 	} else { // generate an aggregated report
 
 		// list all prowjobs in the patchset matching pj's type (pre- or post-submit)
-
 		selector := map[string]string{
 			clientGerritRevision: pj.ObjectMeta.Labels[clientGerritRevision],
 			pjTypeLabel:          pj.ObjectMeta.Labels[pjTypeLabel],
 			gerritReportLabel:    pj.ObjectMeta.Labels[gerritReportLabel],
 		}
 
-		var pjsOnRevisionWithSameLabel v1.ProwJobList
 		if err := c.pjclientset.List(newCtx, &pjsOnRevisionWithSameLabel, ctrlruntimeclient.MatchingLabels(selector)); err != nil {
 			logger.WithError(err).WithField("selector", selector).Errorf("Cannot list prowjob with selector")
 			return nil, nil, err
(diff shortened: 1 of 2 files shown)


prow/crier/reporters/gerrit/reporter.go
```diff
@@ -246,16 +246,22 @@ func (c *Client) ShouldReport(ctx context.Context, log *logrus.Entry, pj *v1.Pro
...
-	}) && allPJsAgreeToReport([]string{kube.OrgLabel, kube.RepoLabel, kube.PullLabel}, func(pj *v1.ProwJob) bool {
+	}) && allPJsAgreeToReport([]string{kube.OrgLabel, kube.RepoLabel, kube.PullLabel}, func(otherPj *v1.ProwJob) bool {
 		// Newer patchset exists, skip report
-		return patchsetNumFromPJ(pj) <= patchsetNum
+		return patchsetNumFromPJ(otherPj) <= patchsetNum
+	}) && allPJsAgreeToReport([]string{kube.OrgLabel, kube.RepoLabel, kube.PullLabel}, func(otherPj *v1.ProwJob) bool {
+		// This job has duplicate(s) and there are newer one(s)
+		if otherPj.Spec.Job == pj.Spec.Job && otherPj.CreationTimestamp.After(pj.CreationTimestamp.Time) {
+			return false
+		}
```
cjwagner, line 263: nit: It would be more efficient to combine this with the previous call to `allPJsAgreeToReport` since they use the same selector.

prow/crier/reporters/gerrit/reporter.go
```diff
@@ -341,7 +347,11 @@ func (c *Client) Report(ctx context.Context, logger *logrus.Entry, pj *v1.ProwJo
 				mostRecentJob[pjOnRevisionWithSameLabel.Spec.Job] = &pjsOnRevisionWithSameLabel.Items[idx]
 			}
 		}
-		for _, pjOnRevisionWithSameLabel := range mostRecentJob {
+		for jobName, pjOnRevisionWithSameLabel := range mostRecentJob {
+			if jobName == pj.Spec.Job && pjOnRevisionWithSameLabel.CreationTimestamp.After(pj.CreationTimestamp.Time) {
+				logger.WithField("newer_job", pjOnRevisionWithSameLabel.Name).Debug("There is a newer job, skip reporting.")
+				return nil, nil, nil
+			}
```
cjwagner, line 354: Is this needed? I would have expected the `mostRecentJob` map to already include `pj` as the most recent job. If `pj` is not the most recent job we wouldn't be reporting.

cjwagner (commented):
I'm not sure that this completely addresses the problem. What if the newest, already reported job is garbage collected and that makes the older job seem like the newest? We can't permanently rely on the existence of a newer job in order to know that we shouldn't report the old one. I think we may want to mark the older job as reported once we know it shouldn't report due to a newer job running. That way we won't accidentally report the old job if the newer job is GCed first.

prow/crier/reporters/gerrit/reporter.go
chaodaiG: yeah this is the expected behavior and is what this piece of code does. Did you mean to do this check earlier?

cjwagner: No, I'm confused by the purpose of this. It looks like dead code to me. We should never have any job with the same name as `pj` in `mostRecentJob` except for `pj` itself, right?

chaodaiG:
> I'm not sure that this completely addresses the problem. What if the newest, already reported job is garbage collected and that makes the older job seem like the newest? We can't permanently rely on the existence of a newer job in order to know that we shouldn't report the old one. I think we may want to mark the older job as reported once we know it shouldn't report due to a newer job running. That way we won't accidentally report the old job if the newer job is GCed first.

I have thought about this approach of marking all jobs from the same PR that report on the same label once a report is determined to be sent. This, however would hurt performance because there might be several times more prowjobs to be processed and it would take quite some time(note that we have seen 10 seconds timeouts being hit multiple times on the same PR), and the reason for doing so is to prevent this corner case of a prowjob not cleaned up when re-queue happened.

On the other hand I think this check is good for majority of cases, because sinker GC prowjobs based on creation timestamp, so the newer prowjob must exist if an old prowjob still exists in prow. So maybe we should be fine here?

prow/crier/reporters/gerrit/reporter.go
chaodaiG: I see, you're right, `ShouldReport` already does this so yeah this is dead code, let me delete it

cjwagner:
> On the other hand I think this check is good for majority of cases, because sinker GC prowjobs based on creation timestamp, so the newer prowjob must exist if an old prowjob still exists in prow. So maybe we should be fine here?

This isn't great to rely on, especially if we have a large sinker loop period and/or large loop duration. Its probably true most of the time though.

> I have thought about this approach of marking all jobs from the same PR that report on the same label once a report is determined to be sent. This, however would hurt performance because there might be several times more prowjobs to be processed and it would take quite some time

Would there actually be many extra ProwJobs? This should only happen if we have multiple versions of the same job running which would require multiple `/test` commands or patchsets being pushed and for reporting to trigger before plank can abort the duplicates (this would also require the duplicates to complete and become available to report in this time window). This can certainly still happen, but it doesn't seem like that common of a case.

Even if it is initially more expensive, I wonder if we might actually get better performance in the long run by marking the jobs as reported so they can be quickly ignored when we process them rather than having to look up all jobs for the same revision to see if there is anything else in the reporting group or a newer run of the same job. On this note, either way we should probably have a check in `ShouldReport` to quit early if the previous report state is the current state.

chaodaiG:
> Would there actually be many extra ProwJobs? This should only happen if we have multiple versions of the same job running which would require multiple `/test` commands or patchsets being pushed and for reporting to trigger before plank can abort the duplicates (this would also require the duplicates to complete and become available to report in this time window). This can certainly still happen, but it doesn't seem like that common of a case.

I think repeatedly pushing new commits onto the same PR would make this operation expensive, which could be solved by skipping marking aborted jobs as reported, as I would imagine majority of old jobs were aborted, and `aborted` status is filtered out pretty early in the cycle https://github.com/kubernetes/test-infra/blob/30b07d16e76673365214ac3110080276b512b466/prow/crier/reporters/gerrit/reporter.go#L183 so I would consider it cheap.

So yeah we can ask gerrit reporter to mark all jobs for the same label as reported, as long as they completed and not aborted. WDYT @cjwagner ?

cjwagner:
> So yeah we can ask gerrit reporter to mark all jobs for the same label as reported, as long as they completed and not aborted.

Yeah I think that makes sense for this reporter since the gerrit adapter actually does the reporting for when a job starts (more tech debt).

chaodaiG:
> > So yeah we can ask gerrit reporter to mark all jobs for the same label as reported, as long as they completed and not aborted.
> 
> Yeah I think that makes sense for this reporter since the gerrit adapter actually does the reporting for when a job starts (more tech debt).

done

PR #24975, opened 2022-01-21
update python packages affected by CVEs

BenTheElder:
https://github.com/yaml/pyyaml/blob/master/CHANGES
https://jinja.palletsprojects.com/en/2.11.x/changelog/

https://github.com/kubernetes/test-infra/security/dependabot?q=is%3Aopen+sort%3Aseverity


gubernator/lint.sh
```diff
@@ -14,7 +14,7 @@
 # limitations under the License.
 
 set -o errexit -o pipefail -o nounset
```
BenTheElder, line 16: this script should probably go away as part of #23830 / #23897

BenTheElder:
jinja2 will require updating the pull-test-infra-gubernator image.

BenTheElder:
feels bad but we're probably OK for now and there's a lot of other things to do.

Issue #22731, opened 2021-06-29
Allow trusting a GH App bot-filed PRs enough to get them tested


petr-muller:
**What would you like to be added**:

GH apps can file PRs to repos, but because a GH app cannot be added as an org member or a collaborator (GH apps are not real users), it seems that we currently lack the way to "trust" Apps to get their PRs automatically tested, without needing a `/ok-to-test` by a kube member.

We could either have a config for this, or piggyback on some GH concept (trust an app if it's installed in an org?) Maybe we could do both.

**Why is this needed**:

Not having this feature makes automation scenarios harder

/area prow
/area prow/plugins

alvaroaleman:
@cjwagner @chaodaiG objections against making this a `trustedApps` map orgOrRepo ->appNames config option?

cjwagner:
> @cjwagner @chaodaiG objections against making this a `trustedApps` map orgOrRepo ->appNames config option?

No objection from me assuming the app name is a unique identifier.

mdbooth:
/remove-lifecycle stale

Our use case for this is still very much alive. We would like to enable automatic testing of PRs opened by our bot.

PR #24978, opened 2022-01-22
Define list of trusted users that can run tests

Ressetkk:
Looking at the code of the Prow I have noticed that it's not possible to flag specific users as trusted when they open a PR. The flagging pull requests can only be defined by `ok-to-test` flag, which can only be added by member of the organisation. 
With that it's not possible to implement some external repository integrations that raise PRs to have the tests run automatically.

This PR implements list of trusted users that are allowed to run tests in their PR. 
This gives the ability to trust external GitHub automations which use bot users, like dependabot, to have automated Jobs execution.

### DIFF
--- a/prow/plugins/config.go
+++ b/prow/plugins/config.go
@@ -429,6 +429,9 @@ type AssignOnLabel struct {
 type Trigger struct {
 	// Repos is either of the form org/repos or just org.
 	Repos []string `json:"repos,omitempty"`
+	// TrustedUsers is the explicit list of GitHub users whose PRs will be automatically
+	// considered as trusted. By default, trigger will ignore this list.
+	TrustedUsers []string `json:"trusted_users,omitempty"`
 	// TrustedOrg is the org whose members' PRs will be automatically built for
 	// PRs to the above repos. The default is the PR's org.
 	//
--- a/prow/plugins/trigger/trigger.go
+++ b/prow/plugins/trigger/trigger.go
@@ -222,7 +222,7 @@ type TrustedUserResponse struct {
 
 // TrustedUser returns true if user is trusted in repo.
 // Trusted users are either repo collaborators, org members or trusted org members.
-func TrustedUser(ghc trustedUserClient, onlyOrgMembers bool, trustedOrg, user, org, repo string) (TrustedUserResponse, error) {
+func TrustedUser(ghc trustedUserClient, onlyOrgMembers bool, trustedUsers []string, trustedOrg, user, org, repo string) (TrustedUserResponse, error) {
 	errorResponse := TrustedUserResponse{IsTrusted: false}
 	okResponse := TrustedUserResponse{IsTrusted: true}
 
@@ -250,6 +250,14 @@ func TrustedUser(ghc trustedUserClient, onlyOrgMembers bool, trustedOrg, user, o
 		if ok, err := ghc.IsCollaborator(org, repo, user); err != nil {
 			return errorResponse, fmt.Errorf("error in IsCollaborator: %w", err)
 		} else if ok {
+			return okResponse, nil
+		}
+	}
+
+	// Determine if user is on trusted_users list.
+	// This allows specific users, or GitHub automations that cannot be added as collaborators.
+	for _, trustedUser := range trustedUsers {
+		if user == trustedUser {
 			return okResponse, nil
 		}
 	}
--- a/prow/plugins/trigger/generic-comment.go
+++ b/prow/plugins/trigger/generic-comment.go
@@ -76,7 +76,7 @@ func handleGenericComment(c Client, trigger plugins.Trigger, gc github.GenericCo
 	}
 
 	// Skip untrusted users comments.
-	trustedResponse, err := TrustedUser(c.GitHubClient, trigger.OnlyOrgMembers, trigger.TrustedOrg, commentAuthor, org, repo)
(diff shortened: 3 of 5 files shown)


prow/plugins/config.go
```diff
@@ -429,6 +429,9 @@ type AssignOnLabel struct {
 type Trigger struct {
 	// Repos is either of the form org/repos or just org.
 	Repos []string `json:"repos,omitempty"`
+	// TrustedUsers is the explicit list of GitHub users whose PRs will be automatically
+	// considered as trusted. By default, trigger will ignore this list.
+	TrustedUsers []string `json:"trusted_users,omitempty"`
```
alvaroaleman, line 434: Trust happens implicitly by being a collaborator on the repo or member of the repos org. The only use case I am aware of where this doesn't work are GitHub apps, ref https://github.com/kubernetes/test-infra/issues/22731

Is that your usecase? If so, can we rename the field to clarify that it is for apps only and I believe we also need to append an `[app]` to the name then?

alvaroaleman: I'm not a fan of the idea of making this a `trused_users` setting rather than a `trusted_apps` one, because:
* It overlaps with trusting users by adding them to the org or repo and will make it difficult to clarify when to use which
* It is hard to discover that this is to be used when someone wants to trust an app

Ressetkk: @alvaroaleman You are right. I've edited the code to reflect this setting should address only GitHub apps that end with either `[app]` or `[bot]` suffix.

prow/plugins/trigger/trigger.go
```diff
@@ -254,6 +254,15 @@ func TrustedUser(ghc trustedUserClient, onlyOrgMembers bool, trustedOrg, user, o
 		}
 	}
 
+	// Determine if user is on trusted_apps list.
+	// This allows automatic tests execution for GitHub automations that cannot be added as collaborators.
+	for _, trustedApp := range trustedApps {
+		// username must explicitly end with either [bot] or [app]
+		if strings.HasSuffix(user, "[bot]") || strings.HasSuffix(user, "[app]") && user == trustedApp {
```
alvaroaleman, line 261: `[app]` can also be used? I've only seen `[bot]` so far and our code only handles that. Do you have an example for `[app]` being used?

Also, the conditional is wrong, right now it means: Has bot suffix or (has app suffix and user == trustedApp). You need to put the first two checks in braces to achieve (has bot suffix or has app suffix) and user == trustedApp)

Ressetkk: No, it was an assumption based on the previous conversation since all GitHub Apps I've experienced only had `[bot]` suffix. Couldn't find any docs about possible suffixes. I have removed this check. Thanks for pointing that out!

prow/plugins/trigger/trigger.go
```diff
@@ -254,6 +254,15 @@ func TrustedUser(ghc trustedUserClient, onlyOrgMembers bool, trustedOrg, user, o
 		}
 	}
 
+	// Determine if user is on trusted_apps list.
+	// This allows automatic tests execution for GitHub automations that cannot be added as collaborators.
+	for _, trustedApp := range trustedApps {
+		// username must explicitly end with [bot] suffix
+		if strings.HasSuffix(user, "[bot]") && user == trustedApp {
```
alvaroaleman, line 261: Could we instead ignore users that do not have the suffix and strip the suffix prior to comparing, so that ppl do not have to put it in their config? That would be a lot nicer to use.

alvaroaleman (commented):
One remark, other than that the code looks good. Could you also add a line in prow/ANNOUNCEMENTS.md as a couple of ppl have already asked for this in the past?

alvaroaleman:
Oh and there is already an issue for this, please add a fixes https://github.com/kubernetes/test-infra/issues/22731 to the body of your PR.

@petr-muller @cjwagner @chaodaiG in the referenced issue it was discussed to make this configurable global/per-org/per repo, do you think that is needed?

cjwagner:
> @petr-muller @cjwagner @chaodaiG in the referenced issue it was discussed to make this configurable global/per-org/per repo, do you think that is needed?

That should be possible with this PR since this is a field of the `Trigger` struct and our configuration is composed of a slice of these structs that match based on the `Repos` field. I guess global may not be supported by the current config, but this PR probably doesn't need to change that.
/approve

prow/plugins/trigger/trigger.go
Ressetkk: @alvaroaleman addressed your review comment. At first I wanted to have the usernames explicitly defined with [bot] suffix, to easily distinct GitHub apps from other users, but since the field will be unly used for bots it sure will be easier to use it that way.

chaodaiG:
> > @petr-muller @cjwagner @chaodaiG in the referenced issue it was discussed to make this configurable global/per-org/per repo, do you think that is needed?
> 
> That should be possible with this PR since this is a field of the `Trigger` struct and our configuration is composed of a slice of these structs that match based on the `Repos` field. I guess global may not be supported by the current config, but this PR probably doesn't need to change that. /approve

Agreed. In majority of cases being able to map a bot to an org is sufficient, and this is also what a prow operator would expect. So the current implementation works for me

/agree

petr-muller:
:+1: 
/approve

alvaroaleman (approved):

PR #24984, opened 2022-01-24
Add IgnoreDrafts option to blunderbuss plugin

Ressetkk:
Right now blunderbuss plugin assigns people to the PRs even when they are in Draft. This PR adds functionality to ignore requesting reviews when PR is Draft and assign reviews when PR changes status to ready for review.

This option works will work only when `ignore_drafts` is `true`. This option should not change the workflow of the plugin when not set. The change works only on PR events, so comments will operate with the plugin as usual.

The functionality is required because in [kyma-project org](https://github.com/kyma-project/test-infra/issues/4798) we do not want opened Draft PRs to be requested for review to avoid unnecessary notifications from GitHub.

### DIFF
--- a/prow/plugins/blunderbuss/blunderbuss_test.go
+++ b/prow/plugins/blunderbuss/blunderbuss_test.go
@@ -570,6 +570,8 @@ func TestHandlePullRequest(t *testing.T) {
 		filesChanged      []string
 		reviewerCount     int
 		expectedRequested []string
+		draft             bool
+		ignoreDrafts      bool
 	}{
 		{
 			name:              "PR opened",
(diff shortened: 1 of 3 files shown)


prow/plugins/blunderbuss/blunderbuss_test.go
```diff
@@ -593,21 +595,35 @@ func TestHandlePullRequest(t *testing.T) {
 			filesChanged:  []string{"a.go"},
 			reviewerCount: 1,
 		},
+		{
+			name:         "draft pr opened, ignore drafts",
+			action:       github.PullRequestActionOpened,
+			draft:        true,
+			ignoreDrafts: true,
+		},
```
cjwagner, line 603: If we set `draft: false` here does the test fail? If it doesn't please update it so that we ensure we're testing the desired behavior.

cjwagner (commented):
One question, otherwise this LGTM.

prow/plugins/blunderbuss/blunderbuss_test.go
Ressetkk: @cjwagner no, the test won't fail. I've updated it and added second test case when Opened PR is not a Draft. I hope this resolves the topic.

petr-muller (approved):
/ok-to-test

cjwagner (approved):

PR #24985, opened 2022-01-24
Switch scalability periodic tests to the new huge-service config

jupblb:
This is part of an effort to get rid of coupling between huge-service and scheduler-throughput tests. For reference see https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/testing/huge-service/config.yaml.

### DIFF
--- a/config/jobs/kubernetes/sig-scalability/sig-scalability-experimental-periodic-jobs.yaml
+++ b/config/jobs/kubernetes/sig-scalability/sig-scalability-experimental-periodic-jobs.yaml
@@ -100,7 +100,6 @@ periodics:
       - --gcp-project-type=scalability-project
       - --gcp-zone=us-east1-b
       - --provider=gce
-      - --env=CL2_ENABLE_HUGE_SERVICES=true
       - --env=CL2_SCHEDULER_THROUGHPUT_THRESHOLD=0
       - --test=false
       - --test-cmd=$GOPATH/src/k8s.io/perf-tests/run-e2e.sh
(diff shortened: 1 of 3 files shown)


config/jobs/kubernetes/sig-release/release-branch-jobs/1.20.yaml
```diff
@@ -361,6 +360,7 @@ periodics:
       - --test-cmd-args=--provider=gce
       - --test-cmd-args=--report-dir=$(ARTIFACTS)
       - --test-cmd-args=--testconfig=testing/load/config.yaml
+      - --test-cmd-args=--testconfig=testing/huge-service/config.yaml
```
mborsz, line 363: Have you backported this file to 1.20 branch in perf-tests repo?

Anyway, I don't think we want to change tests in OSS other than in master branch

jupblb: Thanks! This is indeed a good idea. Reverted all the non-master changes. Fixed.

jupblb: Good idea. Removed all the non-master changes. Fixed. :)

mborsz:
Please split this PR and enable this feature in periodics first, only then (after ensuring that the test still works) in presubmits

jupblb:
Done.

mborsz:
/lgtm
/approve
/hold

Please submit this tomorrow morning to ensure that if we break anything we will be able to react.

PR #24993, opened 2022-01-24
increase gce periodic prow job timeout

amacaskill:
We have recently added volume cloning to the GCP PD CSI Driver, and we have enabled all of the e2e tests for volume cloning (listed below). Each of these tests are run for 3 different storage classes and several different fsTypes(ext2, ext3, ext4, xfs).
- [multiVolume [Slow] should concurrently access the volume and its clone from pods on the same node [LinuxOnly]](https://github.com/kubernetes/kubernetes/blob/master/test/e2e/storage/testsuites/multivolume.go#L376)
- [provisioning should provision storage with pvc data source](https://github.com/kubernetes/kubernetes/blob/master/test/e2e/storage/testsuites/provisioning.go#L240)
- [provisioning should provision storage with pvc data source in parallel](https://github.com/kubernetes/kubernetes/blob/master/test/e2e/storage/testsuites/provisioning.go#L278) 

Before the volume cloning tests were enabled, the periodic prow job on average took about 133 minutes to run. After the volume cloning tests were enabled, the periodic prow job took on average about 143 minutes. This means that with the new volume cloning tests, the periodic prow job is now taking on average 10 minutes longer than it was previously. This is is why I decided to increase the timeout by 10 minutes.

### DIFF
--- a/config/jobs/kubernetes/sig-storage/sig-storage-gce-config.yaml
+++ b/config/jobs/kubernetes/sig-storage/sig-storage-gce-config.yaml
@@ -81,7 +81,7 @@ presubmits:
         - --repo=k8s.io/kubernetes=$(PULL_REFS)
         - --repo=k8s.io/release
         - --upload=gs://kubernetes-jenkins/pr-logs
-        - --timeout=150
+        - --timeout=160
         - --scenario=kubernetes_e2e
         - --
         - --build=quick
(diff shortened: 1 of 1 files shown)


saikat-royc:
> We have recently enabled volume cloning in the pdcsi driver, which added more tests. Sometimes the tests take a few minutes more than 90min because of these new tests.

mention about the cloning tests

mattcary:
Can you add how long the tests have been taking for you, so we can evaluate if the new threshold of 100m is reasonable or if we should bump it to 105m or 120m?

PR #24997, opened 2022-01-25
Make the csi-proxy-integration job required

mauriciopoppe:
/cc @ddebroy @jingxu97


mauriciopoppe:
/assign @msau42 

The job has been running for some time successfully as optional, this change makes it required

PR #24999, opened 2022-01-25
Use containerd directly as the PidOf can recognize

pacoxu:
Fixes kubernetes/kubernetes#107062

The last fix doesn't work.

`procfs.PidOf(xxx)` can only recognize the bin file name or absolute path

```
Pids of containerd:  [1812]
---
Pids of /usr/bin/containerd:  [1812]
---
Pids of /bin/containerd:  []
```

I run a simple test on my env. 

Refer to code here:
https://github.com/kubernetes/kubernetes/blob/c1e69551be1a72f0f8db6778f20658199d3a686d/pkg/util/procfs/procfs_linux.go#L99-L168


bart0sh (approved):
/lgtm

SergeyKanzhelev (approved):
/lgtm

PR #25001, opened 2022-01-25
Switch scalability presubmit tests to the new huge-service config

jupblb:
This is part of an effort to get rid of coupling between huge-service and scheduler-throughput tests. For reference see https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/testing/huge-service/config.yaml.

This is was by https://github.com/kubernetes/test-infra/pull/24985.

/assign @mborsz


jupblb:
https://github.com/kubernetes/test-infra/pull/24985 is merged but let's wait until next week to make sure nothing breaks.

PR #25002, opened 2022-01-25
run ingress conformance tests on CI

aojea:
we are not running the ingress conformance tests on the CI as seen in https://github.com/kubernetes/kubernetes/pull/107753

make the regex match on everything tagged as `Ingress`


robscott:
Thanks @aojea!

/lgtm

PR #25005, opened 2022-01-25
Updating KES from 6.4.0 to 8.5.1

chaodaiG:
In preparation for v1.22 migration. 6.4.0 wouldn't work with k8s v1.22


listx:
It would be nice if we had a job that diffed the YAML change using  https://github.com/homeport/dyff or something similar, as I find large YAML changes difficult to reason about.

chaodaiG:
/hold
RBAC change is needed

chaodaiG:
RBAC change is in
/unhold

mpherman2:
/lgtm
/hold
Updating KES sgtm. Not positive about all the specifics. Holding so you can merge when you're ready to keep an eye on deployment

cjwagner (approved):

PR #25006, opened 2022-01-25
Makefile: remove deprecated `test` target

listx:
This target should not be used. Instead, users should be using the
`test` target in the Makefile under the root folder, which has many
enhancements baked into it (such as using the top level `.go-version`
file to set the go version).

### DIFF
--- a/prow/Makefile
+++ b/prow/Makefile
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-all: build test
+all: build
 
 
 # ALPINE_VERSION is the version of the alpine image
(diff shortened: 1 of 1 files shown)


prow/Makefile
```diff
@@ -54,10 +54,7 @@ get-build-cluster-credentials:
 build:
```
chaodaiG, line 54: I would get rid of this as well, it should not be used either

listx: Is there some guidance I can write in the commit message about an alternate non-deprecated equivalent? Unlike for the `test` target, I don't see a `build` target in the top level Makefile.

listx: For now I just added a note about what this rule used to do, and how it's no longer encouraged.

chaodaiG: good point, build target is missing from top level, which should be added.

chaodaiG (approved):

prow/Makefile
listx: Should the top level build target do what the old one used to do? Or something else?

chaodaiG: I would first ask who would use it.

listx: > build target is missing from top level, which should be added
> I would first ask who would use it.

These statements sound contradictory to me. Can you clarify? Namely:

(1) Do we want a `build` target in the top-level Makefile? And (2) if so, should it be the same rule as before (locally installing all binaries under `prow/cmd/...`)?

PR #25007, opened 2022-01-25
crier: gerrit: fix wording for `/test all`

listx:
According to docs, `/test all` runs all tests, not just the failed ones.

See https://github.com/kubernetes/test-infra/blob/05c84566e96e2a6edfeea915d32b27ff81c43786/prow/plugins/trigger/trigger.go#L122.


listx:
/close

This is included with https://github.com/kubernetes/test-infra/pull/25009.

PR #25009, opened 2022-01-25
Parameterize go unit.sh

listx:


chaodaiG (approved):

hack/make-rules/go-test/unit.sh
```diff
@@ -17,6 +17,28 @@ set -o nounset
 set -o errexit
 set -o pipefail
 
+usage()
+{
+  >&2 cat <<EOF
+Usage: $0 [FOLDER(S)_TO_TEST]
```
cjwagner, line 23: nit: It was a bit confusing to me based on the wording that this only accepts a single `go test/build` style path rather than a list of directories.

cjwagner (approved):
/hold

hack/make-rules/go-test/unit.sh
listx: Fixed

cjwagner (approved):

PR #25010, opened 2022-01-25
Verify-codegen includes prowjob CRD

chaodaiG:
Prowjob CRD was generated by update-codegen, but was not included as part of verify-codegen from the very beginning.


chaodaiG:
/hold
@alvaroaleman , do we know whether it's intended not to verify generated prowjob CRD? It worries me a little bit that update-codegen locally generated so much new stuff in prowjob CRD

alvaroaleman (approved):
Defintely not intended. It seems most (all?) changes stem from the embedded podspec changing so not as bad as it could be

PR #25013, opened 2022-01-26
SIG-CLI: enable testing for krm-functions-registry repo

mengqiy:
Enable testing for https://github.com/kubernetes-sigs/krm-functions-registry.
Added OWNERS file that is the same as https://github.com/kubernetes-sigs/krm-functions-registry/blob/main/OWNERS


mengqiy:
@chases2 and @wojtek-t Please take a look

chases2 (approved):

PR #25014, opened 2022-01-26
ingress job don't run network policy tests

aojea:
follow up on previous PR
since the Ingress tag seems to match on network policy tests :shrug: 
https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/107753/pull-kubernetes-e2e-gci-gce-ingress/1486053322031894528


MrHohn (approved):
/lgtm

PR #25016, opened 2022-01-26
Do not add `needs-ok-to-test` if PR already has `ok-to-test`

saschagrunert:
Some automations, for example GitHub actions and dependabot can be
configured to automatically add labels to PRs. To increase CI and
testing feedback, we would now skip the `needs-ok-to-test` label if we
already provide the `ok-to-test` label on PR creation.

This avoids having double labels excluding themself, for example like there: 
https://github.com/kubernetes-sigs/downloadkubernetes/pull/180#issuecomment-1022054735

### DIFF
--- a/prow/plugins/trigger/pull-request.go
+++ b/prow/plugins/trigger/pull-request.go
@@ -286,8 +286,17 @@ I understand the commands that are listed [here](https://go.k8s.io/bot-commands?
 %s
 </details>
 `, author, org, org, more, joinOrgURL, labels.OkToTest, encodedRepoFullName, plugins.AboutThisBotWithoutCommands)
-		if err := ghc.AddLabel(org, repo, pr.Number, labels.NeedsOkToTest); err != nil {
+
+		l, err := ghc.GetIssueLabels(org, repo, pr.Number)
+		if err != nil {
 			errors = append(errors, err)
+		} else if !github.HasLabel(labels.OkToTest, l) {
+			// It is possible for bots and other automations to automatically
+			// add the ok-to-test label. If that's the case, then we will not
+			// add the needs-ok-to-test-label any more.
+			if err := ghc.AddLabel(org, repo, pr.Number, labels.NeedsOkToTest); err != nil {
+				errors = append(errors, err)
+			}
 		}
 	}
 
--- a/prow/plugins/trigger/pull-request_test.go
+++ b/prow/plugins/trigger/pull-request_test.go
@@ -22,6 +22,7 @@ import (
 	"strconv"
 	"testing"
 
+	"github.com/google/go-cmp/cmp"
 	"github.com/sirupsen/logrus"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	clienttesting "k8s.io/client-go/testing"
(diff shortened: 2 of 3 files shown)


matthyx:
\o/ indeed doesn't make much sense...
Maybe you could add a test case to prevent future regressions?

saschagrunert:
> Maybe you could add a test case to prevent future regressions?

Yes, added two unit test cases for it.

matthyx:
Thanks Sascha!
/lgtm

cpanato (approved):
cool, thanks for that

/lgtm

prow/plugins/trigger/pull-request.go
```diff
@@ -286,8 +286,19 @@ I understand the commands that are listed [here](https://go.k8s.io/bot-commands?
 %s
 </details>
 `, author, org, org, more, joinOrgURL, labels.OkToTest, encodedRepoFullName, plugins.AboutThisBotWithoutCommands)
-		if err := ghc.AddLabel(org, repo, pr.Number, labels.NeedsOkToTest); err != nil {
+
+		l, err := ghc.GetIssueLabels(org, repo, pr.Number)
+
+		fmt.Printf("::: %v\n", pr.Number)
```
alvaroaleman, line 292: Debugging leftover?

prow/plugins/trigger/pull-request_test.go
```diff
@@ -509,6 +529,12 @@ func TestHandlePullRequest(t *testing.T) {
 					t.Errorf("exptected job %s to not be set to complete.", tc.jobToAbort.Name)
 				}
 			}
+			if tc.noIssueLabelsAdded && len(tc.issueLabelsAdded) > 0 {
+				t.Errorf("exptected no issue labels added, but got %v", g.IssueLabelsAdded)
+			}
+			if len(tc.issueLabelsAdded) > 0 && !reflect.DeepEqual(tc.issueLabelsAdded, g.IssueLabelsAdded) {
```
alvaroaleman, line 535: The boolean check above seems redudundant if you remove the len check here. Also, please use `cmp` for human readable diffs, i.E. `if diff := cmp.Diff(tc.issueLabelsAdded, g.IssueLabelsAdded); diff != "" { err }`

saschagrunert: Alright, switched to the cmp.Diff method and changed dependent tests.

prow/plugins/trigger/pull-request_test.go
```diff
@@ -130,16 +131,18 @@ func TestHandlePullRequest(t *testing.T) {
...
+		ShouldComment      bool
+		HasOkToTest        bool
+		prLabel            string
+		prChanges          bool
+		prAction           github.PullRequestEventAction
+		prIsDraft          bool
+		eventSender        string
+		jobToAbort         *prowapi.ProwJob
+		issueLabelsAdded   []string
+		noIssueLabelsAdded bool
```
alvaroaleman, line 145: Please remove this, it is the same as `issueLabelsAdded: nil/empty`

prow/plugins/trigger/pull-request.go
```diff
@@ -286,8 +286,18 @@ I understand the commands that are listed [here](https://go.k8s.io/bot-commands?
 %s
 </details>
 `, author, org, org, more, joinOrgURL, labels.OkToTest, encodedRepoFullName, plugins.AboutThisBotWithoutCommands)
-		if err := ghc.AddLabel(org, repo, pr.Number, labels.NeedsOkToTest); err != nil {
+
+		l, err := ghc.GetIssueLabels(org, repo, pr.Number)
```
alvaroaleman, line 290: Nit: Could we remove the empty line between this and the err checl?

prow/plugins/trigger/pull-request_test.go
saschagrunert: Sure, removed that one

prow/plugins/trigger/pull-request.go
saschagrunert: Fixed as well :+1:

PR #25025, opened 2022-01-26
InRepoConfig supports gerrit format repo names

mpherman2:
Continue work to get InRepoConfig to support gerrit.
/assign @listx @chaodaiG

### DIFF
--- a/prow/config/config.go
+++ b/prow/config/config.go
@@ -222,6 +222,7 @@ func trimRepoPrefix(repo string) string {
 	return repo
 }
 
+// TODO(mpherman): Do not return error, just use org == ""
 func SplitRepoName(repo string) (string, string, error) {
 	// Normalize repo name to remove http:// or https://, this is the case for some
 	// of the gerrit instances.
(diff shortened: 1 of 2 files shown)


prow/config/config.go
```diff
@@ -2852,12 +2852,17 @@ func (repo OrgRepo) String() string {
...
+	// Trim off http:// or https:// to support gerrit
+	orgRepo = trimRepoPrefix(orgRepo)
 	parts := strings.Split(orgRepo, "/")
 	switch len(parts) {
 	case 1:
 		return &OrgRepo{Org: parts[0]}
 	case 2:
 		return &OrgRepo{Org: parts[0], Repo: parts[1]}
+	//Support Gerrit repos with name Host/Repo/Name
+	case 3:
```
chaodaiG, line 2864: I might be wrong but I think there could be more than 3. How about `parts := strings.SplitN(orgRepo, "/", 2)`? Also, I feel like this is done somewhere else, can that be shared?

mpherman2: Fixed

prow/config/config.go
```diff
@@ -2853,19 +2853,12 @@ func (repo OrgRepo) String() string {
...
-	case 2:
-		return &OrgRepo{Org: parts[0], Repo: parts[1]}
-	//Support Gerrit repos with name Host/Repo/Name
-	case 3:
-		return &OrgRepo{Org: parts[0], Repo: strings.Join(parts[1:3], "/")}
-	default:
-		return nil
+	org, repo, err := SplitRepoName(orgRepo)
+	// Unable to split to Org/Repo
+	if err != nil {
```
chaodaiG, line 2858: I understand that `SplitRepoName` only returns error when it can not be splitted, because I just reviewed the code yesterday. I bet I will be very confused when looking at this code some time later and ask: why do we return something when there is an error?

One alternative idea is to update `SplitRepoName` function to not return error at all, user should be able to decide by `repo == ""` instead of on error

mpherman2: I agree with this, but SplitRepoName is used in quite a few places. I will make this change in a follow up PR so that this PR isn't too large

chaodaiG: sure, please add a comment here explaining clearly why returning valid struct when there is an error

mpherman2: Done

chaodaiG (approved):

PR #25031, opened 2022-01-27
Add support for gerrit InRepoConfig with fallback

mpherman2:
PR enables InRepoConfig for Gerrit. If Getting InRepoConfig fails, this will log a warning and fall back to just getting static presubmits. Doing this as we do not have robust gerrit integration tests yet, and I do not want to break Gerrit component.
/assign @chaodaiG @cjwagner 
/hold

### DIFF
--- a/prow/gerrit/adapter/adapter.go
+++ b/prow/gerrit/adapter/adapter.go
@@ -37,6 +37,7 @@ import (
 	"k8s.io/test-infra/prow/config"
 	reporter "k8s.io/test-infra/prow/crier/reporters/gerrit"
 	"k8s.io/test-infra/prow/gerrit/client"
+	"k8s.io/test-infra/prow/git/v2"
 	"k8s.io/test-infra/prow/io"
 	"k8s.io/test-infra/prow/pjutil"
 )
@@ -60,6 +61,7 @@ type Controller struct {
 	tracker            LastSyncTracker
 	projectsOptOutHelp map[string]sets.String
 	lock               sync.RWMutex
+	cookieFilePath     string
 }
 
 type LastSyncTracker interface {
@@ -97,6 +99,7 @@ func NewController(ctx context.Context, prowJobClient prowv1.ProwJobInterface, o
 		gc:                 gerritClient,
 		tracker:            lastSyncTracker,
 		projectsOptOutHelp: projectsOptOutHelpMap,
+		cookieFilePath:     cookiefilePath,
 	}
 
 	// applyGlobalConfig reads gerrit configurations from global gerrit config,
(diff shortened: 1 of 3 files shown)


prow/gerrit/adapter/adapter.go
```diff
@@ -310,11 +313,30 @@ func (c *Controller) processChange(logger logrus.FieldLogger, instance string, c
...
-		// TODO: Do we want to add support for dynamic postsubmits?
-		postsubmits := c.config().PostsubmitsStatic[cloneURI.String()]
-		postsubmits = append(postsubmits, c.config().PostsubmitsStatic[cloneURI.Host+"/"+cloneURI.Path]...)
+		postsubmits := []config.Postsubmit{}
+		if gc != nil {
+			postsubmits, err = c.config().GetPostsubmits(gc, cloneURI.Host+"/"+cloneURI.Path, func() (string, error) { return baseSHA, nil })
+			if err != nil {
+				//TODO(mpherman): Return Error once we know this usually works
+				logger.Warn("Failed to get InRepoConfig for Postsubmits")
+				postsubmits = append(postsubmits, c.config().PostsubmitsStatic[cloneURI.Host+"/"+cloneURI.Path]...)
```
chaodaiG, line 336: moving this line to be under `gc != nil` statement could cause regression for prow instances that don't have `gc` set, this should also be handled as `else` condition

prow/gerrit/adapter/adapter.go
```diff
@@ -310,11 +313,30 @@ func (c *Controller) processChange(logger logrus.FieldLogger, instance string, c
 
 	changedFiles := listChangedFiles(change)
 
+	//Attempt to Get InRepoConfig
+	opts := git.ClientFactoryOpts{
+		CloneURI:       cloneURI.String(),
+		Host:           cloneURI.Host,
+		CookieFilePath: c.cookieFilePath,
+	}
+	gc, err := git.NewClientFactory(opts.Apply)
```
chaodaiG, line 322: while we don't have integration test, I think we'll need to set up unit test for this, you can use the same local git at https://github.com/kubernetes/test-infra/blob/e84d71dcb03994718ffe9e7d637acf0ae080b0b2/prow/config/inrepoconfig_test.go#L559 for unit testing this

prow/gerrit/adapter/adapter.go
mpherman2: Fixed

prow/gerrit/adapter/adapter.go
mpherman2: I thought I would not be able to at first because the code to get the prow yaml is too complicated to include in the Unit test for the adapter, but I forgot that the ProwYamlGetter can be mocked. Went back and added unit tests

prow/gerrit/adapter/adapter_test.go
```diff
@@ -1111,8 +1208,25 @@ func TestProcessChange(t *testing.T) {
 				gc:            &gc,
```
chaodaiG, line 1208: I think we should also test scenario where `gc = nil`

prow/gerrit/adapter/adapter_test.go
```diff
@@ -1111,8 +1208,25 @@ func TestProcessChange(t *testing.T) {
...
 				tracker:       &fakeSync{val: fakeLastSync},
 			}
+			cloneURI, err := makeCloneURI(tc.instance, tc.change.Project)
+			if err != nil {
+				t.Errorf("error making CloneURI %v", err)
+			}
+
+			lg, cf, err := localgit.NewV2()
```
chaodaiG, line 1216: since ProwYamlGetter is mocked, seems like this is unused?

prow/gerrit/adapter/adapter_test.go
mpherman2: Done

prow/gerrit/adapter/adapter_test.go
mpherman2: ProcessChange takes a clientFactory, and will only get to InRepoConfig if one is provided (if it is nil it will skip InRepoConfig). This is the best way to create one that is non-nil for the unit tests.

chaodaiG: would be nice to add a comment explaining it

mpherman2: Done

prow/gerrit/adapter/adapter_test.go
```diff
@@ -970,6 +1023,63 @@ func TestProcessChange(t *testing.T) {
...
+				Revisions: map[string]gerrit.RevisionInfo{
+					"1": {
+						Number: 1001,
+					},
+				},
+			},
+			instancesMap:     map[string]*gerrit.AccountInfo{testInstance: {AccountID: 42}},
+			instance:         testInstance,
+			shouldError:      false,
+			numPJ:            0,
```
chaodaiG, line 1080: there should also be a unit test fetching static jobs when gc is nil

mpherman2: Done

chaodaiG (approved):

mpherman2:
This was a good first pass, but I was reflecting on this and I am going to close this pull request and work on getting the cached version of InRepoConfig working for Gerrit instead of the uncached version. Closing this PR.

PR #25032, opened 2022-01-27
Add back --container-runtime for previous k8s releases 

adisky:
Containerd jobs using previous supported k8s release started failing with 'container-runtime' flag removal as default
runtime for older k8s releases is docker, so for older releases we are adding the flag again

ref https://github.com/kubernetes/kubernetes/issues/107800

Signed-off-by: Aditi Sharma <adi.sky17@gmail.com>

### DIFF
--- a/config/jobs/kubernetes/sig-node/containerd.yaml
+++ b/config/jobs/kubernetes/sig-node/containerd.yaml
@@ -230,7 +230,7 @@ periodics:
       - --node-args=--image-config-file=/workspace/test-infra/jobs/e2e_node/containerd/containerd-release-1.4/image-config.yaml
       - --deployment=node
       - --gcp-zone=us-central1-b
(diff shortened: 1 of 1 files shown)


adisky:
ref https://storage.googleapis.com/kubernetes-jenkins/logs/ci-containerd-node-e2e-1-5/1486531284850184192/artifacts/tmp-node-e2e-32b754a5-cos-89-16108-604-3/kubelet.log

image

`--container-runtime` is docker which was default before removal, `--container-runtime-endpoint` containerd.sock, so default being picked up by kubelet for older releases since we are not passing it here

Also cc @hakman

SergeyKanzhelev (approved):
/lgtm

hakman:
Thanks @adisky.
/lgtm

PR #25041, opened 2022-01-27
capz: skip in-tree tests instead of out-of-tree tests

jackfrancis:
The following PR updates its set of E2E test definitions so that out-of-tree (aka "external") cloud-provider-azure configurations are the default, and demotes in-tree cloud-provider to an optional test config:

https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1994

In response to the above change, we need to update the capz test-infra specs to invert their skip logic. Instead of skipping "external" cloud-provider-azure scenarios, we not want to skip "in-tree" cloud-provider azure scenarios. This PR does that.

This PR should merge only after capz PR 1994 merges.

### DIFF
--- a/config/jobs/kubernetes-sigs/cluster-api-provider-azure/cluster-api-provider-azure-presubmits-main.yaml
+++ b/config/jobs/kubernetes-sigs/cluster-api-provider-azure/cluster-api-provider-azure-presubmits-main.yaml
@@ -57,7 +57,7 @@ presubmits:
           - name: GINKGO_FOCUS
             value: "Workload cluster creation"
           - name: GINKGO_SKIP
-            value: "Creating a GPU-enabled cluster|.*Windows.*dockershim.*|.*AKS.*|Creating a cluster that uses the external cloud provider"
+            value: "Creating a GPU-enabled cluster|.*Windows.*dockershim.*|.*AKS.*|Creating a cluster that uses the in-tree cloud provider"
         # docker-in-docker needs privileged mode
         securityContext:
           privileged: true
--- a/config/jobs/kubernetes-sigs/cluster-api-provider-azure/cluster-api-provider-azure-presubmits-release-v1beta1.yaml
+++ b/config/jobs/kubernetes-sigs/cluster-api-provider-azure/cluster-api-provider-azure-presubmits-release-v1beta1.yaml
@@ -57,7 +57,7 @@ presubmits:
           - name: GINKGO_FOCUS
             value: "Workload cluster creation"
           - name: GINKGO_SKIP
-            value: "Creating a GPU-enabled cluster|.*Windows.*dockershim.*|.*AKS.*|Creating a cluster that uses the external cloud provider"
+            value: "Creating a GPU-enabled cluster|.*Windows.*dockershim.*|.*AKS.*|in-tree cloud provider|external cloud provider"
         # docker-in-docker needs privileged mode
         securityContext:
           privileged: true


CecileRobertMichon:
> This PR should merge only after capz PR 1994 merges.

/hold

config/jobs/kubernetes-sigs/cluster-api-provider-azure/cluster-api-provider-azure-presubmits-release-v1beta1.yaml
```diff
@@ -57,7 +57,7 @@ presubmits:
           - name: GINKGO_FOCUS
             value: "Workload cluster creation"
           - name: GINKGO_SKIP
-            value: "Creating a GPU-enabled cluster|.*Windows.*dockershim.*|.*AKS.*|Creating a cluster that uses the external cloud provider"
+            value: "Creating a GPU-enabled cluster|.*Windows.*dockershim.*|.*AKS.*|Creating a cluster that uses the in-tree cloud provider"
```
CecileRobertMichon, line 60: since we're not going to backport the change (I think?) we shouldn't change the jobs targetting release branches, only the main branch job + periodic jobs targeting main

jackfrancis: Makes sense.

Does a future 1.* release count as a "backport" in this way? (i.e., we will make it a point not to cherry-pick any template changes at any point in the future)

CecileRobertMichon: good question. for v1beta1 it does since the job is using a regex for the branch `^release-1.*` which would include any future `1.*` release branches. For the v1beta1 job we can maybe make the change additive and skip both "in tree" and "external" so that it keeps working for now and then skips the in-tree template once we cut a new release that includes the change (it won't skip external anymore since that won't be an explicit external test anymore), WDYT ?

jackfrancis: sgtm, done

CecileRobertMichon (commented):
/lgtm

config/jobs/kubernetes-sigs/cluster-api-provider-azure/cluster-api-provider-azure-presubmits-main.yaml
```diff
@@ -57,7 +57,7 @@ presubmits:
           - name: GINKGO_FOCUS
             value: "Workload cluster creation"
           - name: GINKGO_SKIP
-            value: "Creating a GPU-enabled cluster|.*Windows.*dockershim.*|.*AKS.*|Creating a cluster that uses the external cloud provider"
+            value: "Creating a GPU-enabled cluster|.*Windows.*dockershim.*|.*AKS.*|Creating a cluster that uses the in-tree cloud provider"
```
CecileRobertMichon, line 60: @jackfrancis is this PR still needed? #25307 should cover this no?

jackfrancis: Indeed I didn't realize at the time that we won't need to update test-infra anymore when we finally finish the inversion of "in-tree <---> out-of-tree" capz templates.

PR #25043, opened 2022-01-27
capz: add presubmit job for in-tree E2E test

jackfrancis:
This PR adds an in-tree specific job to run against PRs to main and v1beta1 release branches. This should merge before this following capz PR so that we can test those changes:

https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1994

### DIFF
--- a/config/jobs/kubernetes-sigs/cluster-api-provider-azure/cluster-api-provider-azure-presubmits-main.yaml
+++ b/config/jobs/kubernetes-sigs/cluster-api-provider-azure/cluster-api-provider-azure-presubmits-main.yaml
@@ -163,6 +163,38 @@ presubmits:
       testgrid-dashboards: sig-cluster-lifecycle-cluster-api-provider-azure
       testgrid-tab-name: capz-pr-e2e-exp-main
       testgrid-alert-email: k8s-infra-staging-cluster-api-azure@kubernetes.io
+  - name: pull-cluster-api-provider-azure-e2e-in-tree
+    path_alias: "sigs.k8s.io/cluster-api-provider-azure"
+    always_run: false
+    optional: true
+    decorate: true
+    max_concurrency: 5
+    labels:
+      preset-dind-enabled: "true"
+      preset-kind-volume-mounts: "true"
+      preset-azure-cred-only: "true"
+      preset-azure-anonymous-pull: "true"
+    branches:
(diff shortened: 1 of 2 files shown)


config/jobs/kubernetes-sigs/cluster-api-provider-azure/cluster-api-provider-azure-presubmits-main.yaml
```diff
@@ -163,6 +163,38 @@ presubmits:
       testgrid-dashboards: sig-cluster-lifecycle-cluster-api-provider-azure
       testgrid-tab-name: capz-pr-e2e-exp-main
       testgrid-alert-email: k8s-infra-staging-cluster-api-azure@kubernetes.io
+  - name: pull-cluster-api-provider-azure-e2e-in-tree
```
CecileRobertMichon, line 166: do we need a separate job for this specific flavor? Why not use the full e2e job like we currently do for external?

jackfrancis: So the "full" job just runs everything given that it has no skip config?

CecileRobertMichon: however we should update the GINGKO_SKIP for jobs that currently skip the external cloud provider to instead skip the in tree flavor https://github.com/kubernetes/test-infra/pull/25043/files#diff-de63480688e0c1f8b4117747cd58cbafd15ea017e2ad279d178828b0b5097cd5R60

jackfrancis: If so then I suppose we don't need it. Is there any benefit of having the option to test this discretely? (sounds like we're not doing that for other targeted flavor tests, so probably not?)

jackfrancis: Ya! That's what this PR is for:

https://github.com/kubernetes/test-infra/pull/25041

jackfrancis:
Will continue to use the "full" job instead

PR #25045, opened 2022-01-28
add kubernetes-sigs/oci-proxy presubmits

BenTheElder:
ref: https://github.com/kubernetes-sigs/oci-proxy/issues/7

### DIFF
--- a/config/jobs/kubernetes/sig-k8s-infra/oci-proxy/sig-k8s-infra-oci-proxy-presubmits.yaml
+++ b/config/jobs/kubernetes/sig-k8s-infra/oci-proxy/sig-k8s-infra-oci-proxy-presubmits.yaml
@@ -1,0 +1,46 @@
+presubmits:
+  kubernetes-sigs/oci-proxy:
+  - name: pull-oci-proxy-verify
+    decorate: true
+    always_run: true
+    optional: true
+    spec:
+      containers:
+      - image: golang
+        command:
+        - make
+        - verify
+        # shellcheck has its own job so we don't need docker
+        - VERIFY_SHELLCHECK=false
+  - name: pull-oci-proxy-shellcheck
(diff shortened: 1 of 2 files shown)


config/jobs/kubernetes/sig-k8s-infra/oci-proxy/sig-k8s-infra-oci-proxy-presubmits.yaml
```diff
@@ -0,0 +1,46 @@
...
+        - verify
+        # shellcheck has its own job so we don't need docker
+        - VERIFY_SHELLCHECK=false
+  - name: pull-oci-proxy-shellcheck
+    decorate: true
+    always_run: true
+    optional: true
+    spec:
+      containers:
+      # NOTE: We will have to keep this up to date with the repo ...
```
BenTheElder, line 21: TODO: we can maybe loosen the version check in the repo to a warning to make this easier.
I think it's worth not needing DinD

config/jobs/kubernetes/sig-k8s-infra/oci-proxy/sig-k8s-infra-oci-proxy-presubmits.yaml
```diff
@@ -0,0 +1,46 @@
+presubmits:
+  kubernetes-sigs/oci-proxy:
+  - name: pull-oci-proxy-verify
+    decorate: true
+    always_run: true
+    optional: true
+    spec:
+      containers:
+      - image: golang
```
BenTheElder, line 9: these just use golang as a known image with make / bash, if the golang version matches we will automatically use it but if it doesn't we will fetch it with gimme in the make system anyhow.

same for the other jobs except shellcheck

PR #25047, opened 2022-01-28
update oci-proxy shellcheck presubmit to use shellcheck installer

BenTheElder:
depends on https://github.com/kubernetes-sigs/oci-proxy/pull/11

the official shellcheck images lack bash, make etc. (even the alpine version), this is probably a better approach. should unblock making these jobs blocking


ameukam:
/lgtm

FWIW we already have `shellcheck`present in `gcr.io/k8s-staging-infra-tools/k8s-infra:latest`: https://github.com/kubernetes/k8s.io/blob/main/images/k8s-infra/Dockerfile#L38. (need to be updated)

BenTheElder:
ACK, I think I want to try this approach first because like everything else in the oci-proxy project tool versions are fully controlled in repo, in a single definition, so we don't have to chase around updating them in 20 places before CI passes :+)

PR #25050, opened 2022-01-28
add oci-proxy image pushing

BenTheElder:
see: https://github.com/kubernetes-sigs/oci-proxy/issues/8


BenTheElder:
fixed missing testgrid dashboard issue, passing relevant tests now (Still waiting on test-infra-integration, but that's for prow components, which we're not touching here ...)

PR #25051, opened 2022-01-28
label_sync: disable retries on 404 responses

bear-redhat:
There are no point to retry for requests with 404 (e.g. `repo does not exists`.) to save some tokens.

The situation described in github `client.go` ( https://github.com/kubernetes/test-infra/blob/1256a81e66d64ace1cf50520e82510a3316319d6/prow/github/client.go#L1038 ) seems not quite apply here since we are dealing with repos.

### DIFF
--- a/label_sync/main.go
+++ b/label_sync/main.go
@@ -705,6 +705,7 @@ type client interface {
 	FindIssues(query, order string, ascending bool) ([]github.Issue, error)
 	GetRepos(org string, isUser bool) ([]github.Repo, error)
 	GetRepoLabels(string, string) ([]github.Label, error)
+	SetMax404Retries(int)
 }
 
 func newClient(tokenPath string, tokens, tokenBurst int, dryRun bool, graphqlEndpoint string, hosts ...string) (client, error) {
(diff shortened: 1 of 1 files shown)


petr-muller:
/retitle label_sync: disable retries on 404 responses
/ok-to-test

petr-muller (approved):

fejta:
/assign
Will you please describe the motivation a bit in the description?

bear-redhat:
> /assign Will you please describe the motivation a bit in the description?

I guess there are no point to retry for requests with 404 (e.g. `repo does not exists`.) to save some tokens.

The situation described here ( https://github.com/kubernetes/test-infra/blob/1256a81e66d64ace1cf50520e82510a3316319d6/prow/github/client.go#L1038 ) seems not quite apply here since we are dealing with repos.

petr-muller:
Yeah, the motivation should usually go into the PR description so that the reviewer knows what is happening. I agree that `label_sync` is rarely run in situations where retrying 404 is useful. In our case we were affected when we ran `label_sync` while missing some permissions on the repo, got a lot of 404s as a result, and the execution took ages because all failing operations were retried several times with a back-off.

fejta:
My request is to have a non-empty PR description.

bear-redhat:
> My request is to have a non-empty PR description.

I have fixed the description of the issue.

fejta (approved):

PR #25053, opened 2022-01-28
Lower frequency of Windows release jobs

jsturtevant:
We typically run the release jobs daily, since the number of commits going in is much lower.

/sig windows
/assign @marosset


mboersma (approved):
lgtm

PR #25057, opened 2022-01-29
Add Gerrit adapter metrics for query and processing results.

cjwagner:
These will let us write alerts based for high gerrit adapter error rates. Including counters for success results lets us normalize against load.
/assign @chaodaiG


fejta (approved):
/hold

PR #25068, opened 2022-01-31
Additional CSI Proxy working directory setup for LVP

mauriciopoppe:
`/tmp` is defined in https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/blob/master/test/e2e/e2e_test.go#L55, the discovery directories created during the e2e are under `/tmp/<random_test_case>`

CSI_PROXY_FLAGS was added in https://github.com/kubernetes/kubernetes/pull/107806

/cc @jingxu97 
/assign @msau42

### DIFF
--- a/config/jobs/kubernetes-sigs/sig-storage-local-static-provisioner/sig-storage-local-static-provisioner.yaml
+++ b/config/jobs/kubernetes-sigs/sig-storage-local-static-provisioner/sig-storage-local-static-provisioner.yaml
@@ -99,6 +99,11 @@ presubmits:
         # this is done so that the e2e test runs faster
         - name: WINDOWS_NODE_OS_DISTRIBUTION
           value: "win2019"
+        # in the e2e tests the flag --working-dir set on CSI Proxy allows
+        # setting the locations where CSI Proxy is supposed to run
+        # the test location is defined in /test/e2e/e2e_test.go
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes-sigs/sig-storage-local-static-provisioner/sig-storage-local-static-provisioner.yaml
```diff
@@ -99,6 +99,11 @@ presubmits:
         # this is done so that the e2e test runs faster
         - name: WINDOWS_NODE_OS_DISTRIBUTION
           value: "win2019"
+        # in the e2e tests the flag --working-dir set on CSI Proxy allows
+        # setting the locations where CSI Proxy is supposed to run
+        # the test location is defined in /test/e2e/e2e_test.go
+        - name: CSI_PROXY_FLAGS
+          value: "--working-dir C:\\tmp"
```
msau42, line 106: I know on Linux, /tmp has special behavior regarding permissions that caused some tests to fail. In Windows, is tmp a special type of directory or normal?

mauriciopoppe: I believe the directory doesn't exist by default in Windows but it's created for these tests, it's just a regular directory as far as I know

PR #25081, opened 2022-02-01
Warn about planned change to default value of 'prow.k8s.io/gerrit-report-label'.

cjwagner:
/assign @chaodaiG


chaodaiG (commented):
When a job set not to report on any label by intention, this might not be the desired behavior. What about just allow jobs to not set label?

cjwagner:
> When a job set not to report on any label by intention, this might not be the desired behavior. What about just allow jobs to not set label?

We kind of support that via setting `skip_report: true`, but we don't currently offer a way to report a job without setting a label. So this wouldn't negatively affect any existing jobs, but I'll change this so that we can support jobs that report without setting a label going forward :+1:

chaodaiG (approved):

PR #25084, opened 2022-02-01
Add support for Cached InRepoConfig for Gerrit with fallback

mpherman2:
Enable InRepoConfig for Gerrit component. This will attempt to do inrepoconfig, but if it fails will just log a warning and continue without InRepoConfig 

/assign @chaodaiG @cjwagner

### DIFF
--- a/prow/cmd/gerrit/main.go
+++ b/prow/cmd/gerrit/main.go
@@ -51,6 +51,7 @@ type options struct {
 	kubernetes             prowflagutil.KubernetesOptions
 	storage                prowflagutil.StorageClientOptions
 	instrumentationOptions prowflagutil.InstrumentationOptions
+	inRepoConfigCacheSize  int
 }
 
 func (o *options) validate() error {
@@ -92,6 +93,7 @@ func gatherOptions(fs *flag.FlagSet, args ...string) options {
 	fs.StringVar(&o.lastSyncFallback, "last-sync-fallback", "", "The /local/path, gs://path/to/object or s3://path/to/object to sync the latest timestamp")
 	fs.BoolVar(&o.dryRun, "dry-run", false, "Run in dry-run mode, performing no modifying actions.")
 	fs.StringVar(&o.tokenPathOverride, "token-path", "", "Force the use of the token in this path, use with gcloud auth print-access-token")
+	fs.IntVar(&o.inRepoConfigCacheSize, "in-repo-config-cache-size", 1000, "Cache size for ProwYAMLs read from in-repo configs.")
 	for _, group := range []flagutil.OptionGroup{&o.kubernetes, &o.storage, &o.instrumentationOptions, &o.config} {
 		group.AddFlags(fs)
 	}
@@ -129,7 +131,7 @@ func main() {
 		logrus.WithError(err).Fatal("Error creating opener")
 	}
 
-	c := adapter.NewController(ctx, prowJobClient, op, cfg, o.projects, o.projectsOptOutHelp, o.cookiefilePath, o.tokenPathOverride, o.lastSyncFallback)
+	c := adapter.NewController(ctx, prowJobClient, op, ca, o.projects, o.projectsOptOutHelp, o.cookiefilePath, o.tokenPathOverride, o.lastSyncFallback, o.inRepoConfigCacheSize)
 
 	logrus.Infof("Starting gerrit fetcher")
 
--- a/prow/cmd/gerrit/main.go
+++ b/prow/cmd/gerrit/main.go
@@ -93,7 +93,7 @@ func gatherOptions(fs *flag.FlagSet, args ...string) options {
 	fs.StringVar(&o.lastSyncFallback, "last-sync-fallback", "", "The /local/path, gs://path/to/object or s3://path/to/object to sync the latest timestamp")
 	fs.BoolVar(&o.dryRun, "dry-run", false, "Run in dry-run mode, performing no modifying actions.")
 	fs.StringVar(&o.tokenPathOverride, "token-path", "", "Force the use of the token in this path, use with gcloud auth print-access-token")
-	fs.IntVar(&o.inRepoConfigCacheSize, "in-repo-config-cache-size", 1000, "Cache size for ProwYAMLs read from in-repo configs.")
+	fs.IntVar(&o.inRepoConfigCacheSize, "in-repo-config-cache-size", 100, "Cache size for ProwYAMLs read from in-repo configs. Each host receives its own cache.")
 	for _, group := range []flagutil.OptionGroup{&o.kubernetes, &o.storage, &o.instrumentationOptions, &o.config} {
 		group.AddFlags(fs)
 	}
--- a/prow/gerrit/adapter/adapter.go
+++ b/prow/gerrit/adapter/adapter.go
@@ -38,6 +38,7 @@ import (
 	"k8s.io/test-infra/prow/config"
 	reporter "k8s.io/test-infra/prow/crier/reporters/gerrit"
 	"k8s.io/test-infra/prow/gerrit/client"
+	"k8s.io/test-infra/prow/git/v2"
 	"k8s.io/test-infra/prow/io"
 	"k8s.io/test-infra/prow/pjutil"
 )
@@ -78,6 +79,7 @@ type Controller struct {
 	tracker            LastSyncTracker
 	projectsOptOutHelp map[string]sets.String
 	lock               sync.RWMutex
+	cookieFilePath     string
 }
 
 type LastSyncTracker interface {
@@ -115,6 +117,7 @@ func NewController(ctx context.Context, prowJobClient prowv1.ProwJobInterface, o
 		gc:                 gerritClient,
 		tracker:            lastSyncTracker,
 		projectsOptOutHelp: projectsOptOutHelpMap,
+		cookieFilePath:     cookiefilePath,
 	}
 
 	// applyGlobalConfig reads gerrit configurations from global gerrit config,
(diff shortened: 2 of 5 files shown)


prow/gerrit/adapter/adapter.go
```diff
@@ -166,12 +174,39 @@ func (c *Controller) applyGlobalConfigOnce(cfg config.Getter, gerritClient *clie
...
+	return cache, nil
+}
+
 // Sync looks for newly made gerrit changes
 // and creates prowjobs according to specs
 func (c *Controller) Sync() error {
 	syncTime := c.tracker.Current()
 	latest := syncTime.DeepCopy()
-
+	repoCacheMap := map[string]*config.InRepoConfigCache{}
```
listx, line 209: I guess each repo is assigned its own cache (even though the InRepoConfigCache is already keyed by the repo identifier string (so it can handle multiple repos)), because each repo needs its own cookeFilePath for auth? If so, I would mention this in the comments here.

Wait, I'm still confused why this is necessary because the cache itself doesn't care about auth (it's not an argument the cache takes when we use it). 🤔

prow/gerrit/adapter/adapter.go
```diff
@@ -332,9 +376,18 @@ func (c *Controller) processChange(logger logrus.FieldLogger, instance string, c
 
 	switch change.Status {
 	case client.Merged:
-		// TODO: Do we want to add support for dynamic postsubmits?
-		postsubmits := c.config().PostsubmitsStatic[cloneURI.String()]
-		postsubmits = append(postsubmits, c.config().PostsubmitsStatic[cloneURI.Host+"/"+cloneURI.Path]...)
+		postsubmits := []config.Postsubmit{}
+		if cache != nil {
```
listx, line 380: Might be easier to just always assume that the cache exists. That is, exit with a fatal error (which you seem to be already doing) if the cache cannot be initialized on startup. This is what the `sub` component does, and it makes the code usage around the cache easier because you don't have to do the "does the cache exist" if/else condition every time.

prow/gerrit/adapter/adapter.go
```diff
@@ -332,9 +376,18 @@ func (c *Controller) processChange(logger logrus.FieldLogger, instance string, c
...
+			postsubmits, err = cache.GetPostsubmits(cloneURI.Host+"/"+cloneURI.Path, func() (string, error) { return baseSHA, nil })
```
listx, line 381: Nit: you should do something like

```
hostPath := cloneURI.Host+"/"+cloneURI.Path
```

and use this variable here as an argument to `GetPostsubmits`. As it stands it's hard to talk about this argument as it is an unevaluated expression.

prow/gerrit/adapter/adapter.go
```diff
@@ -332,9 +376,18 @@ func (c *Controller) processChange(logger logrus.FieldLogger, instance string, c
...
+			if err != nil {
+				//TODO(mpherman): Return Error once we know this usually works
+				logger.Warn("Failed to get InRepoConfig for Postsubmits")
```
listx, line 384: Nit: This should be "Failed to get cached InRepoConfig for Postsubmits"

prow/gerrit/adapter/adapter.go
chaodaiG: I think it's because of `cloneURI` is different among each host, it's required for cloning repo

prow/gerrit/adapter/adapter.go
```diff
@@ -181,8 +216,23 @@ func (c *Controller) Sync() error {
...
 			})
+
+			cloneURI, err := makeCloneURI(instance, change.Project)
+			if err != nil {
+				return fmt.Errorf("makeCloneURI: %w", err)
+			}
+
+			var cache *config.InRepoConfigCache
+			var ok bool
+			if cache, ok = repoCacheMap[cloneURI.Host]; !ok {
```
chaodaiG, line 227:
```suggestion
			cache, ok = repoCacheMap[cloneURI.Host]
			if !ok {
```

prow/gerrit/adapter/adapter.go
chaodaiG: `Sync` is called every minute from upstream, which means this `repoCacheMap` is re-initialized every minute. Can we make this part of `Controller`, so that the cache persist among each run?

prow/gerrit/adapter/adapter.go
```diff
@@ -166,12 +174,39 @@ func (c *Controller) applyGlobalConfigOnce(cfg config.Getter, gerritClient *clie
...
+		return nil, nil
+	}
+	// Initialize cache for fetching Presubmit and Postsubmit information. If
+	// the cache cannot be initialized, exit with an error.
+	cache, err = config.NewInRepoConfigCache(
+		cacheSize,
+		configAgent,
+		config.NewInRepoConfigGitCache(gc))
+	// If we cannot initialize the cache, exit with an error.
+	if err != nil {
```
chaodaiG, line 197: Looking at the callstack of `config.NewInRepoConfigCache`, the only case where err is non nil is when `cacheSize < 0`, I think we should just error out here

prow/gerrit/adapter/adapter.go
```diff
@@ -166,12 +174,39 @@ func (c *Controller) applyGlobalConfigOnce(cfg config.Getter, gerritClient *clie
...
 }
 
+// Helper function to create the cache used for InRepoConfig. Currently only attempts to create cache and returns nil if failed.
+func createCache(cloneURI *url.URL, cookieFilePath string, cacheSize int, configAgent *config.Agent) (cache *config.InRepoConfigCache, err error) {
+	opts := git.ClientFactoryOpts{
+		CloneURI:       cloneURI.String(),
+		Host:           cloneURI.Host,
+		CookieFilePath: cookieFilePath,
+	}
+	gc, err := git.NewClientFactory(opts.Apply)
```
chaodaiG, line 184: err is non nil only when `ioutil.TempDir` failed, see https://github.com/kubernetes/test-infra/blob/5bc58aa2c9045da8549bf21a284f40d1160348c3/prow/git/v2/client_factory.go#L140. I think this is a legit error and it's totally fine to error out.

mpherman2: Done

prow/gerrit/adapter/adapter.go
mpherman2: Done

prow/gerrit/adapter/adapter.go
mpherman2: Done

prow/gerrit/adapter/adapter.go
mpherman2: Done

prow/gerrit/adapter/adapter.go
mpherman2: Done

prow/gerrit/adapter/adapter.go
mpherman2: Done

prow/gerrit/adapter/adapter.go
mpherman2: Done

chaodaiG (approved):
/hold
in case you want someone else to also take a look

prow/gerrit/adapter/adapter.go
```diff
@@ -332,9 +374,13 @@ func (c *Controller) processChange(logger logrus.FieldLogger, instance string, c
...
 	switch change.Status {
 	case client.Merged:
-		// TODO: Do we want to add support for dynamic postsubmits?
-		postsubmits := c.config().PostsubmitsStatic[cloneURI.String()]
-		postsubmits = append(postsubmits, c.config().PostsubmitsStatic[cloneURI.Host+"/"+cloneURI.Path]...)
+		postsubmits, err := cache.GetPostsubmits(trimmedHostPath, func() (string, error) { return baseSHA, nil })
+		if err != nil {
+			//TODO(mpherman): Return Error once we know this usually works
+			logger.Warn("Failed to get cached InRepoConfig for Postsubmits")
+			postsubmits = append(postsubmits, c.config().PostsubmitsStatic[trimmedHostPath]...)
```
cjwagner, line 381: I think we need to always append the jobs from `PostsubmitsStatic[trimmedHostPath]`, not only when in repo config fails. We must not have a unit test to confirm that jobs under the `trimmedHostPath` identifier are included, that would be nice to add.
This looks like it is here to find jobs that omit the `https://` prefix.

This also applies to presubmit lookup.

prow/cmd/gerrit/main.go
```diff
@@ -92,6 +93,7 @@ func gatherOptions(fs *flag.FlagSet, args ...string) options {
 	fs.StringVar(&o.lastSyncFallback, "last-sync-fallback", "", "The /local/path, gs://path/to/object or s3://path/to/object to sync the latest timestamp")
 	fs.BoolVar(&o.dryRun, "dry-run", false, "Run in dry-run mode, performing no modifying actions.")
 	fs.StringVar(&o.tokenPathOverride, "token-path", "", "Force the use of the token in this path, use with gcloud auth print-access-token")
+	fs.IntVar(&o.inRepoConfigCacheSize, "in-repo-config-cache-size", 1000, "Cache size for ProwYAMLs read from in-repo configs.")
```
cjwagner, line 96: Unlike with Sub, this cache size limit is per host rather than global. I think we should clarify this and reduce the default size significantly as well.

cjwagner (commented):
Looks good, I just have two small change requests.

cjwagner (approved):

PR #25086, opened 2022-02-02
Generic typescript rollup with make rule, source control generated ts bundle file

chaodaiG:
Part of: #23829 

Pretty much reverse engineered bazel rules for rollup_bundle, copy/pasted all related configurations/steps for rolling up bundles. Using job-history as an example, the generated bundle.min.js file was compared with https://prow.k8s.io/static/job_history_bundle.min.js and confirmed that they are identical

/cc @BenTheElder


chaodaiG:
/hold
this should trigger verify-codegen

chaodaiG:
/unhold
verify-codegen already ran as part of pull-test-infra-verify-lint

chaodaiG:
/hold
unfortunately modifying rollup.config.js file would break bazel rollup, will make a copy of the file for now

hack/make-rules/update/ts-rollup.sh
```diff
@@ -0,0 +1,30 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o nounset
+set -o errexit
+set -o pipefail
+
+TS_PACKAGES_DIR=(
```
BenTheElder, line 20: we should probably dry this out somewhere. e.g. we could source a file with this, or we could read in a file with package dirs one to a line.

hack/make-rules/verify/codegen.sh
```diff
@@ -22,12 +22,14 @@ cd $REPO_ROOT
 
 # place to stick temp binaries
 BINDIR="${REPO_ROOT}/_bin"
+mkdir -p "${BINDIR}"
```
BenTheElder, line 25: not sure if we want `-p`, since BINDIR doesn't appear to be overridable, and we should expect REPO_ROOT to exist? if this fails without -p then we have REPO_ROOT wrong and probably want to catch that.

BenTheElder (commented):
generally LGTM on first pass, some initial nits though

hack/make-rules/update/ts-rollup.sh
chaodaiG: done

hack/make-rules/verify/codegen.sh
chaodaiG: when _bin dir already exists, mkdir exits 0 with -p

BenTheElder: I mean, having `-p` means it will try to create the entire hierarchy if it `$REPO_ROOT/` exist.

chaodaiG: yeah I can see that, I think I can make it safer by:
```
if [[ ! -d "$BINDIR" ]]; then
  mkdir $BINDIR
fi
```

chaodaiG: done

BenTheElder (approved):
/lgtm
/approve

great stuff! :shipit:

PR #25087, opened 2022-02-02
sidecar: do not fail fast when censoring secrets

petr-muller:
`filepath.Walk` bails out when `WalkFunc` returns an error, so previously when we encountered a problematic artifact, we did not censor the remaining ones. This happened on multiple levels.

Make the censoring iterations never fail fast, so that all items are at least processed. Sprinkle some comments around warning future contributors to be careful here. There is already an `errors` channel set up and used from the nested `censor()` method, so I think it is safe to also use it from our iterations - all errors end up collected just before the method returns.

There is one more opportunity for the censoring to fail (if we fail to load just one secret we do not proceed with the censoring at all) but that is a separate problem so I just left a TODO.


petr-muller:
/hold

Let me try whip up a testcase...

prow/sidecar/censor.go
```diff
@@ -65,6 +65,11 @@ func (o Options) censor() error {
 
 	secrets, err := loadSecrets(o.CensoringOptions.SecretDirectories, o.CensoringOptions.IniFilenames)
 	if err != nil {
+		// TODO(petr-muller): This return makes the censoring mechanism fragile, single failure in `loadSecrets`
+		// will prevent us from censoring all other secrets that were successfully loaded. Alternatively,
+		// we could be more strict and just bail out at our callsite in run.go:preUpload() instead of just
```
stevekuznetsov, line 70: I can follow this up with an option at the censoring config level that makes the preUpload failure case the whole process to fail if set, something opposite of "BestEffort bool`?

stevekuznetsov (approved):

stevekuznetsov:
Offline chat: a corrupt archive would be a great test for this. Two other related fixes that may be useful:

- https://github.com/kubernetes/test-infra/pull/25091
- https://github.com/kubernetes/test-infra/pull/25091

fejta (approved):
/hold

petr-muller:
/hold cancel

Added a testcase from https://github.com/kubernetes/test-infra/pull/25146 (thanks @stevekuznetsov !), PTAL

stevekuznetsov (approved):
<3

PR #25092, opened 2022-02-02
sidecar/censoring: allow opting into fast failure

stevekuznetsov:
If a user has set up secret censoring on output and knows that it should
succeed in most cases, it may be useful for them to stop treating the
censoring process as best-effort. In such a case, failures during
censoring would cause the whole `sidecar` process to crash. This would
lead to terrible UX for the user of an affected job, but we can do no
better, as even uploading the job logs may be unsafe in such a
situation.

/assign @petr-muller

### DIFF
--- a/prow/sidecar/run.go
+++ b/prow/sidecar/run.go
@@ -203,7 +203,12 @@ func (o Options) preUpload() {
 
 	if o.CensoringOptions != nil {
 		if err := o.censor(); err != nil {
-			logrus.Warnf("Failed to censor data: %v", err)
+			log := logrus.WithError(err)
+			handler := log.Warnf
+			if o.CensoringOptions.Strict {
+				handler = log.Fatalf
+			}
+			handler("Failed to censor data: %v", err)
 		}
 	}
 }
--- a/prow/cmd/sidecar/main.go
+++ b/prow/cmd/sidecar/main.go
@@ -18,8 +18,12 @@ package main
 
 import (
 	"context"
+	"io"
+	"os"
+	"path/filepath"
 
 	"github.com/sirupsen/logrus"
+	"k8s.io/test-infra/prow/version"
 
 	"k8s.io/test-infra/prow/logrusutil"
 	"k8s.io/test-infra/prow/pod-utils/options"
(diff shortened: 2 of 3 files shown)


prow/sidecar/run.go
```diff
@@ -203,7 +203,11 @@ func (o Options) preUpload() {
 
 	if o.CensoringOptions != nil {
 		if err := o.censor(); err != nil {
-			logrus.Warnf("Failed to censor data: %v", err)
+			handler := logrus.Warnf
```
fejta, line 206: log := logrus.WithError(err)
handler := log.Warnf

fejta (approved):
/hold

stevekuznetsov:
Good catch, @fejta ! Updated

prow/sidecar/run.go
```diff
@@ -203,7 +203,12 @@ func (o Options) preUpload() {
 
 	if o.CensoringOptions != nil {
 		if err := o.censor(); err != nil {
-			logrus.Warnf("Failed to censor data: %v", err)
+			log := logrus.WithError(err)
+			handler := log.Warnf
+			if o.CensoringOptions.Strict {
+				handler = log.Fatalf
+			}
+			handler("Failed to censor data: %v", err)
```
petr-muller, line 211: Would it be possible to upload some kind of tombstone artifact so that this is at least communicated to the end user that comes searching their artifact? A simple .txt file with a simple message would probably save people time and effort to investigate.

Otherwise LGTM

petr-muller (approved):
I proposed a thing but I do not insist on it, feel free to unhold

fejta (approved):
/hold

stevekuznetsov:
/hold cancel

@petr-muller it would be a really large refactor to somehow upload one and only one file on failure here, since we'd need to muck with the GCS Options that otherwise are used. I added a commit to this PR that writes `sidecar` logs to a file that can be read (without the censoring fast failure flag set) for clarity for the meanwhile

fejta (approved):
/hold

stevekuznetsov:
I don't have the wherewithal to argue with `bazel` if I'm honest. @petr-muller feel free to take this over if you think it's a useful feature.

/close

PR #25097, opened 2022-02-03
Typescript rollup packages under prow/cmd/deck/static and gopherage/cmd/html/static/

chaodaiG:
Part of: https://github.com/kubernetes/test-infra/issues/23829

Following https://github.com/kubernetes/test-infra/pull/25086, roll up all other packages under prow/cmd/deck/static, the generated mini js files were diffed against prow.k8s.io and they were identical

There are more under prow/spyglass/lenses, have tried to roll up one of them, it was successful, but had trouble identifying the min js files from prow.k8s.io, will do them in a separate PR

/cc @BenTheElder


BenTheElder (approved):
/lgtm
/approve

PR #25105, opened 2022-02-03
hook: fixes event dispatch for external plugins

bartoszmajsak:
If an event type is not handled by built-in plugins (e.g. is of `repository` type), it will not be properly dispatched by `hook` to the external plugins. In such a case it lacks `srcRepo` value leading to no external plugins being selected.

This PR adds failing test showing the problem (dbd66d5a44f720e50205fe9d74dc57e2b608f447) and introduces the fix (363d771c7f39c685d6c5149f3832909907e05856).

### DIFF
--- a/prow/hook/server_test.go
+++ b/prow/hook/server_test.go
@@ -17,11 +17,16 @@ limitations under the License.
 package hook
 
 import (
+	"bytes"
+	"io/ioutil"
 	"net/http"
 	"net/http/httptest"
 	"reflect"
 	"strings"
 	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
 
 	"k8s.io/test-infra/prow/githubeventserver"
 	"k8s.io/test-infra/prow/plugins"
@@ -253,6 +258,55 @@ func TestNeedDemux(t *testing.T) {
 			},
 		},
 		{
+			name: "external plugins handling other events",
+
+			eventType: "repository",
+			srcRepo:   "kubernetes/test-infra",
+			plugins: map[string][]plugins.ExternalPlugin{
+				"kubernetes/test-infra": {
+					{
+						Name: "coffee",
+					},
+				},
+				"kubernetes/kubernetes": {
+					{
+						Name:   "gumbo",
+						Events: []string{"issue_comment"},
+					},
+				},
+				"kubernetes": {
+					{
+						Name:   "chicken",
+						Events: []string{"repository"},
+					},
+					{
+						Name: "water",
+					},
+					{
+						Name:   "chocolate",
+						Events: []string{"pull_request", "issue_comment", "repository"},
+					},
+				},
+			},
+
+			expected: []plugins.ExternalPlugin{
+				{
+					Name: "coffee",
+				},
+				{
+					Name: "water",
+				},
+				{
+					Name:   "chicken",
+					Events: []string{"repository"},
+				},
+				{
+					Name:   "chocolate",
+					Events: []string{"pull_request", "issue_comment", "repository"},
+				},
+			},
+		},
+		{
 			name: "we have variety but disabled that repo",
 
 			eventType: "issue_comment",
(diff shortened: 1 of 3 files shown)


matthyx:
@bartoszmajsak can you fix the CLA issue?

bartoszmajsak:
> @bartoszmajsak can you fix the CLA issue?

@matthyx I'm working on it. I need corporate one it seems.

bartoszmajsak:
> > @bartoszmajsak can you fix the CLA issue?
> 
> @matthyx I'm working on it. I need corporate one it seems.

@matthyx Done.

prow/hook/server_test.go
```diff
@@ -334,3 +386,180 @@ func TestNeedDemux(t *testing.T) {
...
+				"X-Hub-Signature":   hmac,
+				"content-type":      "application/json",
+			},
+			Body: body,
+
+			ExpectedDispatch: []string{"/coffee", "/water", "/chocolate"},
+		},
+	}
+
+	for _, tc := range testcases {
```
alvaroaleman, line 515: Please make these subtests so the go test output becomes legible if it errors
```suggestion
	for _, tc := range testcases {
	    t.Run(tc.name, func(t *testing.T){
	    ....
	    }
```

prow/hook/server_test.go
```diff
@@ -334,3 +386,180 @@ func TestNeedDemux(t *testing.T) {
...
+		name string
+
+		Method string
+		Header map[string]string
+		Body   string
+
+		ExpectedDispatch []string
+	}{
+		{
+			name: "Repository Event",
```
alvaroaleman, line 486: So this is a made-up event type? Can we maybe rename the test and event type to clarify that and avoid breaking the testcase should github actually add a `repository` event type, something like "unknown event type gets dispatched to external plugin"?

prow/hook/server_test.go
```diff
@@ -334,3 +386,180 @@ func TestNeedDemux(t *testing.T) {
...
+		if err != nil {
+			t.Fatal(err)
+		}
+		for k, v := range tc.Header {
+			r.Header.Set(k, v)
+		}
+		s.ServeHTTP(w, r)
+		s.wg.Wait()
+
+		if !containSameElements(tc.ExpectedDispatch, calledExternalPlugins) {
```
alvaroaleman, line 546: Again for legibility in the error case, please use `cmp`:
```suggestion
                if diff := cmp.Diff(tc.ExpectedDispatch, calledExternalPlugin, cmpopts.SortSlices(a, b string) bool { return a < b })); diff != "" {
		if !containSameElements(tc.ExpectedDispatch, calledExternalPlugins) {
```

alvaroaleman (commented):
Thanks for your work, two small comments for the new test. Please also squash your commits.

prow/hook/server_test.go
bartoszmajsak: It's an actual event type, in fact, the very reason for this PR, as I mentioned in the description :)

https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#repository

prow/hook/server_test.go
bartoszmajsak: Thank you for the suggestion. I didn't know about `cmp`. Done in 99a292f9e5

prow/hook/server_test.go
bartoszmajsak: @alvaroaleman Do you want me to add another test case with the made-up event as well? Please let me know what you prefer and I'm happy to squash all of the commits.

prow/hook/server_test.go
bartoszmajsak: Good catch, fixed in 66bae4b533.

prow/hook/server_test.go
```diff
@@ -17,6 +17,10 @@ limitations under the License.
 package hook
 
 import (
+	"bytes"
+	"github.com/google/go-cmp/cmp"
```
alvaroaleman, line 21: Please put them in the next import group, so that we have one that only has stdlib imports

bartoszmajsak: Done (d19843cfdf), sorry about it. I assumed my IDE did that and haven't looked.

prow/hook/server_test.go
bartoszmajsak: Added in 9454c43a44.

bartoszmajsak:
@alvaroaleman I squashed all the changes into one commit.

alvaroaleman (approved):
thanks!

prow/hook/server_test.go
```diff
@@ -334,3 +388,188 @@ func TestNeedDemux(t *testing.T) {
...
+}
+
+// newTestClient returns *http.Client with Transport replaced to avoid making real calls
+func newTestClient(fn roundTripFunc) *http.Client {
+	return &http.Client{
+		Transport: fn,
+	}
+}
+
+func TestDemuxEvent(t *testing.T) {
```
chaodaiG, line 406: @bartoszmajsak , this test is flaky, see https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/test-infra/25190/pull-test-infra-unit-test/1491504183361146880

bartoszmajsak: Oh, interesting. Thanks for sharing it @chaodaiG.
I ran that multiple times (both here through my subsequent commits as well as locally) and haven't noticed that problem yet. I will check what could be the cause and hopefully send a PR with the fix.

bartoszmajsak: I just ran it in a loop and indeed roughly one out of ten times it fails on my machine. I will work on the fix today.

bartoszmajsak: @chaodaiG https://github.com/kubernetes/test-infra/pull/25222 should fix it. Thanks again!

PR #25108, opened 2022-02-03
populate bool as true when omitempty

smg247:
In order for bool fields that are tagged as `json:xyz,omitempty` to appear on the generated yaml we need to populate them to the non-empty value of `true`. I attempted to solve this problem in a number of other ways (trying to change the tag, type, etc..), and nothing else that I came up with is possible. The downside to this solution is that the bools will show as `true` in the output yaml, but we currently have all strings showing as " ", for instance. This doesn't seem to deviate from that expectation.

### DIFF
--- a/pkg/genyaml/populate_struct.go
+++ b/pkg/genyaml/populate_struct.go
@@ -123,6 +123,10 @@ func PopulateStruct(in interface{}) interface{} {
 			concreteMap.SetMapIndex(key, value)
 
 			valueOf.Elem().Field(i).Set(concreteMap)
+		case reflect.Bool:
+			if strings.Contains(typeOf.Elem().Field(i).Tag.Get("json"), ",omitempty") {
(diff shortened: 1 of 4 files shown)


stevekuznetsov:
Why are other boolean fields already present? They are not `omitempty`?

smg247:
> Why are other boolean fields already present? They are not `omitempty`?

Yes, bools that are not omitempty, and those that are *bool are already present. This fixes the others.

alvaroaleman (approved):
Thanks! It is not ideal that we have to set the value like this as that means the value in the doc is not reflective of the default. Then again, they probably never are since we have defaulting code that isn't used, so this is probably not making anything worse.

alvaroaleman:
/assign @stevekuznetsov 
I can not approve for pkg/genyaml :upside_down_face:

PR #25110, opened 2022-02-03
kops-grid: create scenario for using registry-sandbox.k8s.io mirror

justinsb:
Still experimental, but we can start to gather results.


justinsb:
Depends on https://github.com/kubernetes/kops/pull/13195 and https://github.com/kubernetes/kops/pull/13194 (if running on GCE, which we're not here)

hakman (approved):

PR #25115, opened 2022-02-03
Refactored master-os-distro flag in runner

melvinj98:
Refactored master-os-distro flag to control-plane-os-distro

### DIFF
--- a/kubetest/e2e/runner.go
+++ b/kubetest/e2e/runner.go
@@ -184,7 +184,7 @@ func (t *GinkgoTester) Run(control *process.Control, extraArgs []string) error {
 	a.addIfNonEmpty("cluster-monitoring-mode", t.ClusterMonitoringMode)
 
 	a.addIfNonEmpty("container-runtime", t.KubeContainerRuntime)
-	a.addIfNonEmpty("master-os-distro", t.MasterOSDistribution)
+	a.addIfNonEmpty("control-plane-os-distro", t.MasterOSDistribution)
 	a.addIfNonEmpty("node-os-distro", t.NodeOSDistribution)
 	a.addInt("num-nodes", t.NumNodes)
 	a.addIfNonEmpty("report-dir", t.ReportDir)


kubetest/e2e/runner.go
```diff
@@ -184,7 +184,7 @@ func (t *GinkgoTester) Run(control *process.Control, extraArgs []string) error {
 	a.addIfNonEmpty("cluster-monitoring-mode", t.ClusterMonitoringMode)
 
 	a.addIfNonEmpty("container-runtime", t.KubeContainerRuntime)
-	a.addIfNonEmpty("master-os-distro", t.MasterOSDistribution)
+	a.addIfNonEmpty("control-plane-os-distro", t.MasterOSDistribution)
```
BenTheElder, line 187: This is a breaking change, the pattern kubeadm used is to support both names for some time and then switch to only control-plane.

The entirety of kubetest [is deprecated](https://github.com/kubernetes/test-infra/tree/master/kubetest#deprecation-notice) per the README though, so I think we should probably focus on doing this in [kubetest2](https://github.com/kubernetes-sigs/kubetest2) and removing usage of kubetest so we can delete kubetest.

BenTheElder: https://github.com/kubernetes/enhancements/issues/2464 (though amwat has moved on, I need to find someone else to take up the remainder of this ...)

BenTheElder: If we want to move on cleaning this up on the kubernetes/kubernetes test-script side more, we can change the environment variables set behind this flag (to both) without breakage though, I think that's a small enough change that we can add it despite ostensibly only accepting critical bug fixes since July 2020.

melvinj98: Yes if possible, I would need to make this change in order to make changes in this PR pass the remainder of testshttps://github.com/kubernetes/kubernetes/pull/107890

BenTheElder: ah I see the context now that you've linked this PR 😅 

we will need something slightly more complex than just renaming this here, as we also need kubetest to work for older kubernetes versions that won't have the new flag. kubetest2 will also require this change.

melvinj98: So where in kubetest2 should this change be made?

BenTheElder: I'm not sure, we can probably start by figuring out what would work here. https://github.com/kubernetes/kubernetes/pull/107890#discussion_r799090763

dims:
@melvinj98 please reopen after the changes @BenTheElder requested

dims:
@melvinj98 please reopen after the changes @BenTheElder requested

PR #25118, opened 2022-02-04
crier: GenerateReport: don't report blank URLs

listx:
If the URL is empty, we don't want to use `jobReportFormat`. So, just use `URL_NOT_FOUND` as a fallback instead.

### DIFF
--- a/prow/crier/reporters/gerrit/reporter.go
+++ b/prow/crier/reporters/gerrit/reporter.go
@@ -51,6 +51,11 @@ const (
 	jobReportFormatWithoutURL = "%s %s %s\n"
 	errorLinePrefix           = "NOTE FROM PROW"
 
+	// Stub for the URL field (when using jobReportFormat), if the URL is
+	// legitimately not set (e.g., because the job never got scheduled). This is
+	// what the user will see in place of an actual URL.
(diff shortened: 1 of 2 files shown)


prow/crier/reporters/gerrit/reporter.go
```diff
@@ -594,6 +599,13 @@ func GenerateReport(pjs []*v1.ProwJob, customCommentSizeLimit int) JobReport {
 	linesWithURLs := make([]string, len(report.Jobs))
 	linesWithoutURLs := make([]string, len(report.Jobs))
 	for i, job := range report.Jobs {
+		// If the URL cannot be found (for example, because the job has not been
+		// scheduled due to some other failure), then use the string
+		// URL_NOT_FOUND in its place instead. This way, we don't report jobs
+		// with an empty URL field.
+		if job.URL == "" {
```
chaodaiG, line 606: would also expect a unit test that produces comment in this style

listx: Isn't that covered in the test case I added?

chaodaiG: oh sorry I read it wrong, thought it's for parsing

chaodaiG (approved):

PR #25119, opened 2022-02-04
Building deck and sinker images with ko

chaodaiG:
Imaging pushing logic was primarily adapted from @BenTheElder 's work at https://github.com/kubernetes-sigs/oci-proxy/blob/main/hack/make-rules/images.sh

- POC building prow images with ko, so far so good with pure Go binaries
- One of the complexity is deck, which involves typescript compilation, static files embedding and discovery
-  There  are other complexities: pod utilities are multi-arch, non-Go images, such as mkpj and grandmatriarch, data files such as clonerefs. Will figure out how to do those in separate PR(s)

for code reviewers: I have tried the deck image on a local kind cluster, and the frontpage works fine

/cc @cjwagner


.prow-images
```diff
@@ -0,0 +1 @@
+prow/cmd/sinker
```
BenTheElder, line 1: something to consider later: we have a lot of "dot files" accumulating in the repo root, I think for things like "go version" it makes sense not to bury these, they will be edited somewhat often and are repo-wide.

for things like "prow images" I think we should consider placing them closer to the sources and not all in the repo root.

chaodaiG:
/hold
WIP

chaodaiG:
cool, at least prow integration test passed with the new deck image

chaodaiG:
the linter error is not related to this PR, fixing in https://github.com/kubernetes/test-infra/pull/25132

chaodaiG:
/unhold
ready for review now

prow/test/integration/prow/cluster/200_ingress.yaml
```diff
@@ -28,6 +28,13 @@ spec:
   rules:
   - http:
       paths:
+      - path: /
```
BenTheElder, line 31: is this change related?

chaodaiG: it's a bug introduced in the past, that `/` was not handled and none of the static files were served for this reason. in another word, without the ingress change the deck deployed with integration test doesn't have any icon, css, or js

prow/cmd/deck/gather-static.sh
```diff
@@ -0,0 +1,64 @@
...
+if [[ "${CLEAN:-}" == "true" ]]; then
+    echo "Running in cleanup mode"
+    CLEAN=true ./hack/make-rules/update/ts-rollup.sh "${REPO_ROOT}/prow/cmd/deck/.ts-packages"
+    rm -rf ${KO_DATA_PATH}
+    exit 0
+fi
+./hack/make-rules/update/ts-rollup.sh "${REPO_ROOT}/prow/cmd/deck/.ts-packages"
+
+STATIC_MAP=()
+while IFS= read -r map
```
fejta, line 41: I'd recommend putting this on a single line:

```bash
while foo; do
  whatever
done
```

fejta (approved):
/hold

fejta (approved):
/hold

hack/make-rules/prow-images.sh
```diff
@@ -0,0 +1,75 @@
+#!/usr/bin/env bash
```
cjwagner, line 1: I'm curious why this is under the `hack/make-rules` directory. Did you intend to add Makefile targets to use this script? Since this is specific to Prow not the whole repo, why not define it under `prow`?

prow/cmd/deck/main.go
```diff
@@ -195,6 +199,16 @@ func gatherOptions(fs *flag.FlagSet, args ...string) options {
 	o.storage.AddFlags(fs)
 	o.pluginsConfig.AddFlags(fs)
 	fs.Parse(args)
+
+	if o.spyglassFilesLocation == "" {
+		o.spyglassFilesLocation = fmt.Sprintf("%s%s", os.Getenv("KO_DATA_PATH"), defaultSpyglassFilesLocation)
```
cjwagner, line 204: Nit: These defaults can still be defined inline with the flag definitions.

prow/test/integration/integration-test.sh
```diff
@@ -33,6 +33,7 @@ function retry() {
 function setup() {
   "${bazel}" run //prow/test/integration:setup-local-registry "$@" || ( echo "FAILED: set up local registry">&2; return 1 )
 
+  PUSH=true KO_DOCKER_REPO="localhost:5001" ./hack/make-rules/prow-images.sh
```
cjwagner, line 36: So the integration test setup now publishes images with ko then overwrites them by publishing with bazel? Is that what is intended?

prow/test/integration/prow/cluster/200_ingress.yaml
cjwagner: Why is this part of the `hook` ingress? It would probably be clearer to move this to the existing `deck` ingress or create another ingress for deck if there is a conflict with the rewrite targets.

I'm also wondering if we should instead be allowing `deck` to optionally serve from a subpath prefix rather than the root path (e.g. serve paths under `/deck/` rather than `/`). That is what the `deck` ingress configuration is expecting.

prow/test/integration/integration-test.sh
chaodaiG: yeah at this stage it's intentional to run images building with both and let images built from bazel overriding the other, as they are what's used in production right now

hack/make-rules/prow-images.sh
chaodaiG: good point, moved over to be under prow

prow/cmd/deck/main.go
chaodaiG: good point, doen

chaodaiG:
Updated
/unhold

cjwagner (approved):

PR #25120, opened 2022-02-04
Make prow/config/config.loadConfig public

jupblb:
This is helpful when there is a need to load a config without validating it.


cjwagner (changes requested):
I don't think we should export this function, it could change in the future and exporting encourages external use. Config loading validation shouldn't be too restrictive. Any validation that may not be universally appropriate is included in `checkconfig` rather than the core config validation.

Could you please share why you have a need to load invalid Prow configuration? I suspect you may be able to leverage the `prow_ignored` field or `.prowignore` files instead.

jupblb:
That is both a good reason not to export it and a solution to my problem. Thank you!

PR #25126, opened 2022-02-04
cluster-api: fix branches config of postsubmit image push

sbueringer:

No strong opinions, but we should either push images on postsubmit for all release-branches or none.

@fabriziopandini @enxebre @vincepri @CecileRobertMichon 

Please let me know what you prefer.

I went with removing the config, because apparently nobody missed those images for >= v0.3


CecileRobertMichon:
/lgtm

since we're already pushing on tags I don't think we need to also push for release branches

sbueringer:
/assign @dims 
for approval

PR #25128, opened 2022-02-04
Typescript: roll up everything else with make rule

chaodaiG:
Part of #23829 

After this change all bazel rollup_bundle targets in this repo are replicated with make rules.

/cc @BenTheElder


BenTheElder (approved):
/lgtm
/approve

PR #25131, opened 2022-02-04
Save Prometheus metrics when collected

MikeSpreitzer:
This PR updates all the tests that collect Prometheus metrics to a specified directory also archive those metrics to the report directory so that they are preserved for general audiences.


aojea:
it lgtm but this belongs to scalability
/assign @wojtek-t

PR #25136, opened 2022-02-06
Fix a couple more instances of old CLA context

cblecker:
- Uses the right CLA context for branch protection
- Fixes a comment that refers to old context

/cc @mrbobbytables @MadhavJivrajan


mrbobbytables (commented):
/lgtm

PR #25146, opened 2022-02-07
sidecar: show resiliency to corrupt archive

stevekuznetsov:
If you try to read it, you get:

```
failed to unarchive input: could not read archive: gzip: invalid header
```

/hold
/assign @petr-muller 

### DIFF
--- a/prow/sidecar/censor_test.go
+++ b/prow/sidecar/censor_test.go
@@ -27,7 +27,6 @@ import (
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
-
 	"k8s.io/test-infra/prow/gcsupload"
 	"k8s.io/test-infra/prow/pod-utils/wrapper"
 	"k8s.io/test-infra/prow/secretutil"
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes/cloud-provider-vsphere/cloud-provider-vsphere-config.yaml
```diff
@@ -206,6 +206,46 @@ presubmits:
       testgrid-dashboards: vmware-presubmits-cloud-provider-vsphere, presubmits-cloud-provider-vsphere-blocking
       testgrid-alert-email: k8s-testing-cloud-provider-vsphere+alerts@groups.vmware.com
 
+  # Executes the test coverage.
+  - name: pull-cloud-provider-vsphere-coverage
```
fejta, line 210: This looks unrelated?

petr-muller:
I'm wondering how do the tests pass here without https://github.com/kubernetes/test-infra/pull/25087 . The corrupted archive should make us bail out and not censor the remaining artifacts :thinking:

stevekuznetsov:
@petr-muller do we know if the `filepath.Walk` is deterministic in order? Perhaps it will only fail sometimes, but not always? Let me try to make this fail ...

stevekuznetsov:
@petr-muller ah - it's not detected as an archive, so it's handled as a binary file:

```
time="2022-02-09T10:15:41-08:00" level=debug msg="Censoring file." path=/tmp/TestCensorIntegration1942421898/001/artifacts/corrupt.tar.gz
```

stevekuznetsov:
I think if I corrupt the data and not the header it should work like we expected

stevekuznetsov:
@petr-muller interesting - the Golang code to detect content type expected more from the header than technically necessary. I've updated the test and it now fails:

```
$ go test ./prow/sidecar/... -run TestCensorIntegration
--- FAIL: TestCensorIntegration (0.00s)
    censor_test.go:198: got an error from censoring: could not walk items to censor them: could not censor archive /tmp/TestCensorIntegration2228356612/001/artifacts/corrupt.tar.gz: could not unpack archive: could not read archive: unexpected EOF
FAIL
FAIL	k8s.io/test-infra/prow/sidecar	0.014s
FAIL
```

stevekuznetsov:
@petr-muller with your code on top we correctly censor everything but still return an `error` from `options.censor()` so you may need to invent some new test - perhaps another run of the integration test where the corruption is done and an error is ignored so we only assert on data? Up to you.

petr-muller:
Cool! I will have some time to polish this next week, got slightly distracted by other things. I think I will add a new test, exactly like you say.

fejta (approved):

PR #25151, opened 2022-02-07
Wrap grandmatriarch with Go program

chaodaiG:
In preparation for building all prow images with ko

/cc @BenTheElder @cjwagner

### DIFF
--- a/prow/cmd/grandmatriarch/BUILD.bazel
+++ b/prow/cmd/grandmatriarch/BUILD.bazel
@@ -1,13 +1,17 @@
-load("@io_bazel_rules_docker//container:image.bzl", "container_image")
+load("@io_bazel_rules_go//go:def.bzl", "go_binary", "go_library")
+load("//prow:def.bzl", "prow_image")
+
+NAME = "grandmatriarch"
 
 # Consider using :app :image pattern from
 # prow_image() in //prow:def.bzl.
-container_image(
+prow_image(
     name = "image",
     base = "@gcloud-base//image",
-    entrypoint = ["/bake.sh"],
+    component = NAME,
+    # Add files under into the root directory.
+    directory = "/",
     files = ["bake.sh"],
-    stamp = True,
     visibility = ["//visibility:public"],
 )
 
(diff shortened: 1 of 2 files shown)


BenTheElder:
perhaps a TODO related to https://github.com/kubernetes/test-infra/blob/a39970aa334aae29a91f2ce199ad920ce2a99937/prow/cmd/grandmatriarch/bake.sh#L16

I think it makes sense to port this to go anyhow. But if we're not planning to port it, it feels hacky to create a go binary just for `ko`.  There's probably another tool better suited to this 🤔

chaodaiG:
> perhaps a TODO related to
> 
> https://github.com/kubernetes/test-infra/blob/a39970aa334aae29a91f2ce199ad920ce2a99937/prow/cmd/grandmatriarch/bake.sh#L16

This TODO was already fulfilled by https://github.com/istio/test-infra/tree/master/authentikos, we've been directing all new users to use authentikos for quite a while, and hopefully can deprecate grandmatriarch soon. The problem is there are still untracked legacy users who might still be using grandmatriarch.
 
> I think it makes sense to port this to go anyhow. But if we're not planning to port it, it feels hacky to create a go binary just for `ko`. There's probably another tool better suited to this 🤔

As the aim is to deprecate grandmatriach, it might be an overkill to find an alternative solution just for this bash script?

cjwagner:
> The problem is there are still untracked legacy users who might still be using grandmatriarch.

FWIW I doubt anyone is using this besides us. We could probably just switch to using authentikos and get rid of this. If you'd prefer to stick with using a Go wrapper temporarily that is fine with me too, but please log a deprecation warning periodically.

chaodaiG:
> > The problem is there are still untracked legacy users who might still be using grandmatriarch.
> 
> FWIW I doubt anyone is using this besides us. We could probably just switch to using authentikos and get rid of this. If you'd prefer to stick with using a Go wrapper temporarily that is fine with me too, but please log a deprecation warning periodically.

added a warning

cjwagner (approved):

PR #25152, opened 2022-02-07
Add capz to windows signal dashboard

jsturtevant:
This adds capz to Windows signal dashboard. The intent to is to remove aks-engine now that capz tests are stable.  By monitoring along side aks-engine for a week we can confirm and then cut over to only cluster api for the signal dashboard in 1.24+.  

This also adds two other test runs that we monitor: the serial jobs and containerd nightly.

/sig windows
/assign @marosset  @ravisantoshgudimetla @claudiubelu


jsturtevant:
I've updated this to switch to CAPZ as the CI for release. See the notification in google groups for more information https://groups.google.com/g/kubernetes-sig-windows/c/OHWiVQ-BVJ4/m/hkZ1olvgAwAJ?utm_medium=email&utm_source=footer

PR #25154, opened 2022-02-07
Cloneref reads default GitHub fingerprint without reading files at ru…

chaodaiG:
…ntime

So that the image can work with bare minimal Go binary in the image

/cc @BenTheElder @cjwagner


cjwagner (commented):
Can we now remove the `github_known_hosts` and `ssh_config` files?

chaodaiG:
> Can we now remove the `github_known_hosts` and `ssh_config` files?

Deleted, this PR is scarier now 😨

cjwagner (approved):
/hold
> > Can we now remove the `github_known_hosts` and `ssh_config` files?
> 
> Deleted, this PR is scarier now :fearful:

Yeah, `clonerefs` is not exercised in the integration tests, right? If that's the case I think its worth publishing `clonerefs` with this change to a personal registry so that we can test this out on a single ProwJob (the pod utility images are defined via `decoration_config`).

chaodaiG:
> Details

done, it works fine afaict https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/test-infra/25154/pull-test-infra-unit-test/1490846398852108288:

```
Environment setup
$ golang: append "/tmp/known_hosts"
$ golang: lookup ssh path
# Cloning kubernetes/test-infra at master(e45010f2da5eba44bf6fc9d6ad3659632e9e0790)
```

chaodaiG:
/cc @stevekuznetsov 
this change is in preparing for building clonerefs image with `ko`, which would be a pain to insert `ssh_config` to be under `/etc/ssh` as of now. do you foresee any potential problem here?

petr-muller:
Looks safe to me, feel free to unhold

PR #25155, opened 2022-02-07
kops: test alternative registry for periodics-distros

ameukam:
Related:
  - https://github.com/kubernetes/k8s.io/issues/1834

Test different AWS AMI with registry-sandbox.k8s.io

### DIFF
--- a/config/jobs/kubernetes/kops/build_jobs.py
+++ b/config/jobs/kubernetes/kops/build_jobs.py
@@ -747,11 +747,12 @@ def generate_distros():
         distro_short = distro.replace('ubuntu', 'u').replace('debian', 'deb').replace('amazonlinux', 'amzn') # pylint: disable=line-too-long
         results.append(
             build_test(distro=distro_short,
-                       networking='calico',
+                       networking='kubenet',
                        k8s_version='stable',
                        kops_channel='alpha',
                        name_override=f"kops-aws-distro-image{distro}",
                        extra_dashboards=['kops-distros'],
+                       extra_flags=['--set=spec.assets.containerProxy=registry-sandbox.k8s.io'],
                        runs_per_day=3,
                        )
         )
--- a/config/jobs/kubernetes/kops/kops-periodics-distros.yaml
+++ b/config/jobs/kubernetes/kops/kops-periodics-distros.yaml
@@ -2,7 +2,7 @@
 # 8 jobs, total of 168 runs per week
 periodics:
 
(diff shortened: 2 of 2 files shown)


ameukam:
I also would like to extend this to other periodics for kops.

cc @justinsb

ameukam:
cc @hakman @johngmyers

config/jobs/kubernetes/kops/build_jobs.py
```diff
@@ -752,6 +752,7 @@ def generate_distros():
                        kops_channel='alpha',
                        name_override=f"kops-aws-distro-image{distro}",
                        extra_dashboards=['kops-distros'],
+                       extra_flags=['--set=spec.assets.containerProxy=registry-sandbox.k8s.io'],
```
hakman, line 755: This will require networking='kubenet' to work, as any Calico assets will b missing from `sandbox.k8s.io`, right?

hakman (changes requested):

config/jobs/kubernetes/kops/build_jobs.py
ameukam: The only assets concerned are the kubernetes container images from `k8s.gcr.io`. PR updated.

hakman: Yes, but this setting affects all assets, not just the `k8s.gcr.io ones`. Thanks for updating.

hakman (approved):

config/jobs/kubernetes/kops/build_jobs.py
ameukam: @hakman What's the best approach if I want to extend this change to other periodic jobs ? Is this change enough or do I need to make changes per use ?

hakman: Most jobs use CNIs (as kubenet style networking is not really a production use case). This means that setting this proxy will not work for that case, unless the proxy is made to proxy CNI repos too.
Most likely path is to decide on a final dns name for the CDN cached registry and switch all kOps manifests to use it during the kOps 1.24 development cycle. We can decide later during that cycle if we want to keep it or revert to `k8s.gcr.io`.

ameukam: Thanks for the help!

hakman: My pleasure. Thanks for trying to improve this. I think it will become a huge improvement for many k8s users in the future if successful.

PR #25156, opened 2022-02-07
Push prow images with make rules to staging registry

chaodaiG:
/hold
Depend on #25119 

Part of: #23832 

/cc @BenTheElder

PR #25157, opened 2022-02-08
Building prow images with ko other than cloneutils

chaodaiG:
Part of #25156

Base images were extracted from `BUILD.bazel` files under each component, for example `prow/cmd/branchprotector/BUILD.bazel`

/cc @BenTheElder @cjwagner


.ko.yaml
```diff
@@ -1 +1,31 @@
-defaultBaseImage: gcr.io/k8s-prow/git:v20211018-372b1c8
+# Distroless images:
+# - prow/cmd/admission
+# - prow/cmd/gerrit
+# - prow/cmd/mkpj
+# - prow/cmd/mkpod
+# - prow/cmd/pipeline
```
cjwagner, line 6: Will these still be capable of being execed into so that we can debug with a shell? I suspect it was an oversight that these are not using the alpine base.

.ko.yaml
```diff
@@ -1 +1,31 @@
...
+# defaultBaseImage: gcr.io/distroless/static:nonroot
```
cjwagner, line 7: Did you mean for this to be uncommented?

prow/test/integration/integration-test.sh
```diff
@@ -34,7 +34,7 @@ function setup() {
   "${bazel}" run //prow/test/integration:setup-local-registry "$@" || ( echo "FAILED: set up local registry">&2; return 1 )
 
   if [[ "${PUSH_IMAEG_WITH_KO:-}" == "true" ]]; then
-    PUSH=true KO_DOCKER_REPO="localhost:5001" ./prow/prow-images.sh
+    PUSH=true KO_DOCKER_REPO="localhost:5001" PROW_IMAGES_DEF_FILE="prow/test/integration/prow/.prow-images" ./prow/prow-images.sh
```
cjwagner, line 37: Do we test that all images can be built elsewhere? I'm pretty sure with bazel [all the images were built as part of presubmit testing](https://github.com/kubernetes/test-infra/blob/8b41e455b4f1cb387da060f597a77bddad0e1014/config/jobs/kubernetes/test-infra/test-infra-presubmits.yaml#L19) so it wasn't possible to submit a change that would break image building. If this is the only place we build the images in presubmit we should probably build them all rather than a subset.

.ko.yaml
chaodaiG: These images use distroless base, so they cannot be execed into. I'd prefer stick with what they are within current PR, as this is what's been used in prod today. I can change them to alpine later once confirmed that everything works

.ko.yaml
chaodaiG: doesn't have to, this is the implicit default, I'm adding a comment here to make it a bit more obvious to code reviewers

prow/test/integration/integration-test.sh
chaodaiG: good point. The image building process is slow right now, let me make them parallel to speed it up then we can push all

.ko.yaml
cjwagner: FWIW the explicit default is better in my opinion. Implicit defaults are invisible and could maybe change. Doesn't matter much though.

prow/test/integration/integration-test.sh
chaodaiG: actually `bazel test` does image building but not pushing, and pushing is where most of the time is spent as far as I can observe. I'll push the changes that improves ko performance but would like to keep integration test pushing only relevant images, I think we'll probably need a separate job to check image building if that's what we would like to do

cjwagner: > I think we'll probably need a separate job to check image building if that's what we would like to do

Until we have some job doing that it isn't safe to switch to using Makefile built containers in production. A separate job or including this in the unit testing job seems fine to me. I'm assuming that `ko` would benefit from Go's local build cache being warmed by building unit tests, making the binary compilation step faster at least.

chaodaiG:
note that the go test in integration test is cached, most likely from regular integration test:
```
//prow/test/integration/test:go_default_test                    (cached) PASSED in 0.1s
```

/hold

prow/test/integration/integration-test.sh
chaodaiG: added image building as part of unit test

hack/imagebuilder/main.go
```diff
@@ -0,0 +1,296 @@
...
+	if err := command.Start(); err != nil {
+		return "", err
+	}
+	scanner := bufio.NewScanner(stdOut)
+	var allOut string
+	for scanner.Scan() {
+		out := scanner.Text()
+		allOut = allOut + out
+		logrus.WithField("cmd", command.Args).Info(out)
+	}
```
cjwagner, line 111: Why log every line of output separately rather than using `io.ReadAll` like you do for stderr?

hack/imagebuilder/main.go
```diff
@@ -0,0 +1,296 @@
...
+
+func allTags(arch string) ([]string, error) {
+	baseTags, err := allBaseTags()
+	if err != nil {
+		return nil, err
+	}
+	if arch != allArch {
+		return baseTags, nil
+	}
+	var allTags = baseTags[:]
```
cjwagner, line 161: This is the same as `allTags := baseTags` right?

hack/imagebuilder/main.go
```diff
@@ -0,0 +1,296 @@
...
+		}
+		return nil
+	}
+	if _, err := runCmd(script); err != nil {
+		return err
+	}
+	return nil
+}
+
+func teardown(id *imageDef) error {
```
cjwagner, line 196: `setup` and `teardown` could be combined by adding a parameter.

prow/prow-images.sh
```diff
@@ -21,7 +21,36 @@ REPO_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd -P)"
 cd $REPO_ROOT
 source hack/build/setup-go.sh
```
cjwagner, line 23: We can delete this file now.

hack/imagebuilder/main.go
```diff
@@ -0,0 +1,296 @@
...
+func init() {
+	out, err := runCmd("git", "rev-parse", "--show-toplevel")
+	if err != nil {
+		logrus.WithError(err).Error("Failed getting git root dir")
+		os.Exit(1)
+	}
+	rootDir = out
+}
+
+func init() {
```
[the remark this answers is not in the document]
chaodaiG: good point, combined

hack/imagebuilder/main.go
chaodaiG: typescript rollup takes quite some time, I think it is better to stream logs so we know it's still ongoing

hack/imagebuilder/main.go
chaodaiG: yeah, this was deep clone in python :)

hack/imagebuilder/main.go
chaodaiG: good idea

prow/prow-images.sh
chaodaiG: done

prow/test/integration/integration-test.sh
```diff
@@ -34,7 +38,7 @@ function setup() {
   "${bazel}" run //prow/test/integration:setup-local-registry "$@" || ( echo "FAILED: set up local registry">&2; return 1 )
 
   if [[ "${PUSH_IMAEG_WITH_KO:-}" == "true" ]]; then
-    make -C prow REGISTRY="localhost:5001" push-images
+    go run ./hack/imagebuilder --ko-docker-repo="localhost:5001" --prow-images-file="prow/test/integration/prow/.prow-images" --push
```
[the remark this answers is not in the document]
chaodaiG: done

hack/build/setup-go.sh
```diff
@@ -14,12 +14,7 @@
 # limitations under the License.
 
 # script to setup go version with gimme as needed
-set -o errexit
-set -o nounset
-set -o pipefail
-SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
-REPO_ROOT="$(cd "${SCRIPT_DIR}/../.." && pwd )"
-cd "${REPO_ROOT}"
+# MUST BE RUN FROM THE REPO ROOT DIRECTORY
```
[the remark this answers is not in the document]
cjwagner: I don't follow. When this script is invoked by `prow/Makefile`, the `hack/third_party/gimme/gimme` path on line 36 won't be a proper relative path, right? Same with the `.go-version` file.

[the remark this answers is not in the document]
chaodaiG: good point. I totally forgot that it's also referenced by `prow/Makefile`, updated the script to populate root dir

Issue #25162, opened 2022-02-08
kettle out of disk for sqlite3 db


BenTheElder:
**What happened**:

Received email alert with:

`New error in k8s-gubernator gke_instances`:
```
sqlite3.OperationalError
database or disk is full
Traceback (most recent call last): File "make_db.py", line 376, in OPTIONS.buildlimit, File "make_db.py", line 336, in main download_junit(db, threads, client_class) File "make_db.py", line 318, in download_junit db.insert_build_junits(build_id, junits) File "/kettle/model.py", line 110, in insert_build_junits (path, memoryview(zlib.compress(data.encode('utf-8'), 9)))) sqlite3.OperationalError: database or disk is full
```

**What you expected to happen**:

not out of disk

**How to reproduce it (as minimally and precisely as possible)**:

run kettle for years

**Please provide links to example occurrences, if any**:

??

**Anything else we need to know?**:

thread in testing-ops:
https://kubernetes.slack.com/archives/C7J9RP96G/p1644346402826379

/area kettle
/assign

BenTheElder:
I think this is only in the staging instance, bumping disks anyhow as we are not that far from the limits.

PR #25164, opened 2022-02-08
~double kettle PV sizes

BenTheElder:
https://github.com/kubernetes/test-infra/issues/25162

I've manually deployed these with `kubectl apply` in the relevant cluster (see the makefile), as I didn't find any way to auto-deploy or anything like that :this-is-fine:

### DIFF
--- a/kettle/pv-staging.yaml
+++ b/kettle/pv-staging.yaml
@@ -6,7 +6,7 @@ metadata:
   name: kettle-data-staging
 spec:
   capacity:
-    storage: 100Gi
+    storage: 201Gi
   accessModes:
     - ReadWriteOnce
   persistentVolumeReclaimPolicy: Retain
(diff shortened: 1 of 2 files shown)


BenTheElder:
https://cloud.google.com/kubernetes-engine/docs/how-to/persistent-volumes/volume-expansion#using_volume_expansion

> Note: Don't edit the capacity of the PersistentVolume object directly. If you edit the capacity of a PersistentVolume, and then change the size of the corresponding PersistentVolumeClaim to the same value, volume expansion won't happen. Kubernetes will assume that the backing volume was manually resized, and that no expansion is required.


🙃 

I manually expanded the prod PV, and found that the staging PV is ... already 1TB ...

BenTheElder:
https://samdecrock.medium.com/how-to-resize-a-gcepersistentdisk-on-gke-60e60edaf89 by SSH to underlying vm to manually resize the filesystem (`sudo resize2fs /dev/sdc` in this case)

PR #25173, opened 2022-02-09
fix plugins' config propagation of trigger plugin

osherdp:
When trying to apply configurations in openshift/release/pull/26022, I saw it lacking the ability to preserve trigger plugin configurations.

This change should allow merging also trigger configuration from supplemental configuration to the bigger configuration.

### DIFF
--- a/prow/plugins/config.go
+++ b/prow/plugins/config.go
@@ -1875,7 +1875,7 @@ type Override struct {
 
 func (c *Configuration) mergeFrom(other *Configuration) error {
 	var errs []error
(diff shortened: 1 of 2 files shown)


osherdp:
/hold
(to wait for all reviews to come)

petr-muller:
Thanks for looking into this! I think we'll need tests for this (and also make the existing tests aware), essentially I think this should get the same testing treatment like e.g. `lgtm` gets - they are using the same per-repo configuration approach.

/ok-to-test

osherdp:
> Thanks for looking into this! I think we'll need tests for this (and also make the existing tests aware), essentially I think this should get the same testing treatment like e.g. `lgtm` gets - they are using the same per-repo configuration approach.
> 
> /ok-to-test

Done!

osherdp:
/unhold
as far as I know, it's ready to go
but please let me know if any other changes are needed here

osherdp:
/test pull-test-infra-prow-image-build-test
seems like a temporary registry unavailability

osherdp:
@petr-muller ptal :ok_hand:

petr-muller (approved):

petr-muller:
Thanks!

PR #25183, opened 2022-02-09
cluster-api: enable dind ipv6

sbueringer:

I have no clue why the test suddenly started to fail, but now that I realized that there is `DOCKER_IN_DOCKER_IPV6_ENABLED` env var to enable IPv6 in Docker I wonder how it could ever have worked before.

I would like to try this change on the main job to test if this fixes the issue and if it does roll it out in a follow-up PR to all our ipv6 jobs (on the other branches)

xref: https://github.com/kubernetes/test-infra/blob/882045fa58c930c112d82e9715a400a28d1a9de2/images/bootstrap/runner.sh#L44-L60


sbueringer:
To establish the link. This PR partially resolves: https://github.com/kubernetes-sigs/cluster-api/issues/6086 (only for one job, there's another PR to do it for all IPv6 jobs)

PR #25186, opened 2022-02-09
add unit tests around pr_status_base_urls

smg247:
During my attempt to get to the bottom of why our configuration for `pr_status_base_urls` doesn't behave properly I added some additional test cases. None of this was an issue, but I figured the extra coverage would be good to have.


matthyx (approved):
/lgtm

PR #25192, opened 2022-02-09
Always run filestore k8s integration test

saikat-royc:
The test is still kept optional, but will be always run


saikat-royc:
@leiyiz

PR #25193, opened 2022-02-09
azure: use correct capz release branches for cloud-provider-azure tests

jackfrancis:
This PR modifies the cloud-provider-azure test jobs in two ways:

1. Periodic release jobs now run against the capz "latest release" branch (instead of main). This is to get regular signal before a cloud-provider-azure release that there is something unhealthy in the cloud-provider-azure<-->capz dependency at the intersection that will block the next cloud-provider-azure release. By fixing these issues before a release, we will do a better job of preventing blocking release issues.
2. Adds a new set of cloud-provider-azure tests that use the "latest" capz release (the same release branch that the cloud-provider-azure release tests depend upon). The purpose of these jobs is to anticipate breakage in cloud-provider-azure + capz before the commits are cherry-picked into the cloud-provider-azure release branches. As above, this will give more advance signal of issues, and will help to keep cloud-provider-azure release branch commit histories cleaner.

### DIFF
--- a/config/jobs/kubernetes/sig-cloud-provider/azure/generate.sh
+++ b/config/jobs/kubernetes/sig-cloud-provider/azure/generate.sh
@@ -45,14 +45,17 @@ installCSIdrivers=""
 for release in "$@"; do
   output="${dir}/release-${release}.yaml"
   kubernetes_version="latest"
+  latest_capz_release="release-1.1"
 
   if [[ "${release}" == "master" ]]; then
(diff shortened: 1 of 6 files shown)


jsturtevant:
If possible, these changes should be incorporated into https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/sig-cloud-provider/azure/generate.sh

chewong (approved):

chewong (approved):

PR #25194, opened 2022-02-09
sig-k8s-infra: Add canary job for oci-proxy

ameukam:
Add a canary job of `ci-kubernetes-e2e-gce-scale-performance` that use
registry-sandbox.k8s.io


ameukam:
cc @BenTheElder @dims

PR #25201, opened 2022-02-10
azure: use capz 1.0 release branch for in-tree azure tests

jackfrancis:
Sorry for the thrashing today. This prior PR pinned in-tree cloud-provider-azure tests to the release-1.1 branch of capz:

https://github.com/kubernetes/test-infra/pull/25193

Initial test signal after change suggests that these tests are still not happy against capz @ `release-1.1`:

https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/107981/pull-kubernetes-e2e-capz-conformance/1491515486653911040

The original reason for all of these changes was a breakage in the capz commit @ `v1.0.1`. For example:

https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/107981/pull-kubernetes-e2e-capz-conformance/1491294976963776512

The root cause of the above failures is well-known (it's due to a cert-manager regression in versions of cluster-api prior to `v1.0.4`). Because (1) we have a prior history of success running these in-tree cloud-provider azure tests against a capz v1.0 version and (2) because the `release-1.0` branch of capz has the required, new version of cluster-api built into its CI and cluster definitions (see https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2038), lets pin these test-infra jobs to the `release-1.0` branch instead of `release-1.1`.

### DIFF
--- a/config/jobs/kubernetes/sig-cloud-provider/azure/release-master.yaml
+++ b/config/jobs/kubernetes/sig-cloud-provider/azure/release-master.yaml
@@ -18,7 +18,7 @@ presubmits:
     extra_refs:
       - org: kubernetes-sigs
         repo: cluster-api-provider-azure
-        base_ref: release-1.1
+        base_ref: release-1.0
         path_alias: sigs.k8s.io/cluster-api-provider-azure
         workdir: true
       - org: kubernetes-sigs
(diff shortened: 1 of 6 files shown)


config/jobs/kubernetes/sig-cloud-provider/azure/release-master.yaml
```diff
@@ -789,6 +789,6 @@ periodics:
           memory: "4Gi"
   annotations:
     testgrid-dashboards: provider-azure-master-signal
-    testgrid-tab-name: capz-latest-release-azure-disk-vmss
+    testgrid-tab-name: capz-release-azure-disk-vmss
```
jsturtevant, line 792: Changing the name here will lose the history but other wise will be fine

jsturtevant (approved):
/lgtm

andyzhangx (approved):
/lgtm
/approve

PR #25203, opened 2022-02-10
OpenShift testgrid - Add OCP 4.10 and 4.11 single-node testgrid

omertuc:
We already had 4.9 and 4.8, this will create one for 4.10 and 4.11 as well


petr-muller (approved):
/ok-to-test

PR #25214, opened 2022-02-10
Move httpCookiefile authentication to after mkdir

mpherman2:
inRepoConfig for gerrit does not currently work because it attempts to set httpCookiefile before the temp dir is made for the repo. This PR moves the httpcookiefile config to after the dir is created

/assign @chaodaiG @cjwagner

### DIFF
--- a/prow/git/v2/client_factory.go
+++ b/prow/git/v2/client_factory.go
@@ -236,12 +236,6 @@ func (c *clientFactory) bootstrapClients(org, repo, dir string) (cacher, cloner,
 			logger:   logger,
 		},
 	}
-	if len(c.cookieFilePath) > 0 {
-		err := client.Config("http.cookiefile", c.cookieFilePath)
-		if err != nil {
-			return nil, nil, nil, fmt.Errorf("unable to configure http.cookeifile: %w", err)
-		}
-	}
 	return client, client, client, nil
 }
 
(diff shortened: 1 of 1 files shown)


prow/git/v2/client_factory.go
```diff
@@ -306,6 +306,16 @@ func (c *clientFactory) ClientFor(org, repo string) (RepoClient, error) {
 	return repoClient, nil
 }
 
+func authenticateCookieFile(cookieFilePath string, client RepoClient) error {
+	if len(cookieFilePath) > 0 {
+		err := client.Config("http.cookiefile", cookieFilePath)
+		if err != nil {
+			return fmt.Errorf("unable to configure http.cookeifile: %w", err)
```
stevekuznetsov, line 313:
```suggestion
			return fmt.Errorf("unable to configure http.cookiefile: %w", err)
```

prow/git/v2/client_factory.go
```diff
@@ -285,13 +279,19 @@ func (c *clientFactory) ClientFor(org, repo string) (RepoClient, error) {
...
+			return nil, err
+		}
 		if err := cacheClientCacher.MirrorClone(); err != nil {
 			return nil, err
 		}
 	} else if err != nil {
 		// something unexpected happened
 		return nil, err
 	} else {
+		if err := authenticateCookieFile(c.cookieFilePath, repoClient); err != nil {
```
stevekuznetsov, line 292: Why do we need this here? Presumably the first call would have configured it the first time we cloned the repo?

prow/git/v2/client_factory.go
mpherman2: Done

prow/git/v2/client_factory.go
mpherman2: Done

prow/git/v2/client_factory.go
```diff
@@ -285,6 +279,9 @@ func (c *clientFactory) ClientFor(org, repo string) (RepoClient, error) {
 		if err := os.MkdirAll(cacheDir, os.ModePerm); err != nil && !os.IsExist(err) {
 			return nil, err
 		}
+		if err := authenticateCookieFile(c.cookieFilePath, repoClient); err != nil {
```
stevekuznetsov, line 282: Sorry, but if we have just one callsite let's not make a func for it - can just inline here

stevekuznetsov (approved):

chaodaiG (approved):

Issue #24209, opened 2021-11-02
prow.k8s.io: show skip reason


pohly:
**What would you like to be added**:

When an e2e test gets skipped with `e2eskipper.Skipf`, the reason passed to that function is visible in the JUnit file.  It would be nice to show that on prow.k8s.io.

**Why is this needed**:

It's not always obvious why certain tests didn't run in a job.

NikhilSharmaWe:
@pohly I would like to collaborate to add this feature in prow.k8s.io. Could you please elaborate on how should we proceed for the purpose.

pohly:
I can't provide guidance on this because I don't know how the code behind prow.k8s.io works. I think it can be found here: https://github.com/kubernetes/test-infra/tree/master/prow/spyglass

PR #25221, opened 2022-02-11
prow.k8s.io: add skip reason to Spyglass JUnit lens

Priyankasaggu11929:
**What does this PR do:**

At present, when an e2e test is skipped, the reason for the skip (if passed to the `e2eskipper.Skipf` function) is reflected in the artifacts/junit-*.xml files, but not on the Prow Spyglass JUnit lens: 

For example, in case of the following job:
- JUnit file: https://storage.googleapis.com/kubernetes-jenkins/logs/ci-kubernetes-csi-1-21-on-kubernetes-master/1491563973357604864/artifacts/junit_parallel.xml
- Spyglass lens: https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/ci-kubernetes-csi-1-21-on-kubernetes-master/1491563973357604864

This PR adds the skipped reason to the JUnit spyglass lens's [`template.html`](https://github.com/kubernetes/test-infra/blob/master/prow/spyglass/lenses/junit/template.html#L159) file (wherever passed). 

The outcome is as follows: 

Screenshot 2022-02-11 at 3 05 52 PM
Screenshot 2022-02-11 at 3 06 06 PM

---

**Addresses Issue:**

Fixes https://github.com/kubernetes/test-infra/issues/24209

### DIFF
--- a/prow/spyglass/lenses/junit/template.html
+++ b/prow/spyglass/lenses/junit/template.html
@@ -156,7 +156,11 @@
       {{range .Skipped}}
         {{$firstTest := index .Junit 0}}
         <tr>
-          <td class="mdl-data-table__cell--non-numeric test-name">{{$firstTest.ClassName}}: {{$firstTest.Name}}</td>
+	   {{if eq $firstTest.SkippedReason "" }}
+	   <td class="mdl-data-table__cell--non-numeric test-name">{{$firstTest.ClassName}}: {{$firstTest.Name}}</td>
+	   {{else}}
+	   <td class="mdl-data-table__cell--non-numeric test-name">{{$firstTest.ClassName}}: {{$firstTest.Name}}</br> </br><b>Reason:</b> {{$firstTest.SkippedReason}}</td>
+	   {{end}}
           <td class="mdl-data-table__cell--non-numeric">{{$firstTest.Duration}}</td>
         </tr>
       {{end}}
@@ -166,3 +170,4 @@
 </div>
 {{end}}
 {{end}}
+
--- a/prow/spyglass/lenses/junit/lens.go
+++ b/prow/spyglass/lenses/junit/lens.go
@@ -105,6 +105,14 @@ func (jr JunitResult) Status() testStatus {
 	return res
 }
 
+func (jr JunitResult) SkippedReason() string {
+	res := ""
+	if jr.Skipped != nil {
+		res = *jr.Skipped
+	}
+	return res
+}
+
 // TestResult holds data about a test extracted from junit output
 type TestResult struct {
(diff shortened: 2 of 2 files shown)


prow/spyglass/lenses/junit/template.html
```diff
@@ -156,7 +156,7 @@
       {{range .Skipped}}
         {{$firstTest := index .Junit 0}}
         <tr>
-          <td class="mdl-data-table__cell--non-numeric test-name">{{$firstTest.ClassName}}: {{$firstTest.Name}}</td>
+          <td class="mdl-data-table__cell--non-numeric test-name">{{$firstTest.ClassName}}: {{$firstTest.Name}}<br/>{{$firstTest.Skipped}}</td>
```
petr-muller, line 159: I wonder if it would be useful to include an explicit `Reason: ` prefix? Up to you.

```suggestion
          <td class="mdl-data-table__cell--non-numeric test-name">{{$firstTest.ClassName}}: {{$firstTest.Name}}<br/>Reason: {{$firstTest.Skipped}}</td>
```

/hold

petr-muller (approved):

prow/spyglass/lenses/junit/template.html
Priyankasaggu11929: @petr-muller, thank you for your feedback. I've revised the PR to include the suggestion for an explicit `Reason:` prefix.

Adding context to the new changes:

- There are cases (mainly) where skipped tests do not pass any skipped reason string to the `e2eskipper.Skipf` method. So I added an if/else conditional check in the go template to handle the scenarios of an empty skipped reason (which required the addition of a helper function `SkippedReason()` in the`lens.go` file)
- I created the new function because:
    - directly using  `{{ if eq $firstTest.Skipped "" }}` in the go template didn't work, throwing incompatible string & pointer type comparisons. [Here's](https://stackoverflow.com/questions/35609903/dereferencing-pointers-in-a-golang-text-template) a similar example.
    - Or in case of using`{{if $firstTest.Skipped}}`, it is never `False` (since it is always pointing to an empty skipped reason, therefore the conditional didn't move). 

The updated result is as follows: 

Screenshot 2022-02-14 at 11 07 36 AM

cjwagner (approved):

Priyankasaggu11929:
@cjwagner @petr-muller, thanks so much for the review!

The PR above has a '/hold' on it, which is stopping it from merging. I need help removing that. Thank you! 🙂

petr-muller:
/hold cancel

@Priyankasaggu11929 you can use commands like the one above too ;)

PR #25222, opened 2022-02-11
hook: fixes flaky test

bartoszmajsak:
Adds mutex to prevent concurrent overwrites of the slice used to capture external plugin calls for test verification.

Apologies for introducing it in the first place!


chaodaiG:
Thank you!
/lgtm

PR #25228, opened 2022-02-11
Prow sub component exits when there is a failure

chaodaiG:
Without this change, sub service idles waiting for interruption even when sub thread exited

/cc @cjwagner @mpherman2

### DIFF
--- a/prow/cmd/sub/main.go
+++ b/prow/cmd/sub/main.go
@@ -174,7 +174,7 @@ func main() {
 	pullServer := subscriber.NewPullServer(s)
 	interrupts.Run(func(ctx context.Context) {
 		if err := pullServer.Run(ctx); err != nil {
-			logrus.WithError(err).Error("Failed to run Pull Server")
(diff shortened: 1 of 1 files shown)


prow/cmd/sub/main.go
```diff
@@ -174,7 +174,7 @@ func main() {
 	pullServer := subscriber.NewPullServer(s)
 	interrupts.Run(func(ctx context.Context) {
 		if err := pullServer.Run(ctx); err != nil {
-			logrus.WithError(err).Error("Failed to run Pull Server")
+			logrus.WithError(err).Fatal("Failed to run Pull Server")
```
cjwagner, line 177: This probably isn't critical for sub, but we should technically be using `interrupts.Terminate()` to end the process so that other go routines using the `interrupts` package can exit gracefully.
https://github.com/kubernetes/test-infra/blob/4290b6519652a181e6dd32b387140f459274642f/prow/interrupts/interrupts.go#L78-L82

PR #25231, opened 2022-02-11
Prow sub don't fail the world when a topic failed auth

chaodaiG:

### DIFF
--- a/prow/pubsub/subscriber/server.go
+++ b/prow/pubsub/subscriber/server.go
@@ -24,6 +24,7 @@ import (
 	"net/http"
 	"reflect"
 	"strconv"
+	"strings"
 
 	"github.com/sirupsen/logrus"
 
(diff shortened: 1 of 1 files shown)


prow/pubsub/subscriber/server.go
```diff
@@ -194,8 +199,16 @@ func (s *PullServer) handlePulls(ctx context.Context, projectSubscriptions confi
...
-				if err != nil && !errors.Is(derivedCtx.Err(), context.Canceled) {
-					logrus.WithError(err).Errorf("failed to listen for subscription %s on project %s", sub.string(), project)
+				if err != nil {
+					if errors.Is(derivedCtx.Err(), context.Canceled) {
+						logger.WithError(err).Debug("Exiting as context cancelled")
+						return nil
+					}
+					if strings.Contains(err.Error(), "code = PermissionDenied") {
+						logger.WithError(err).Warn("Seems like missing permission.")
+						return nil
```
listx, line 209: I'm not familiar with how errGroup.Go works. When we return `nil`, does this mean that this topic will get re-tried again? Could you describe the before/after behavior more in your commit message or PR description?

PR #25233, opened 2022-02-11
Pass git dir into CAPZ image push job

Jont828:
This change passes the git directory into the CAPZ image push job so we can generate version info using git commands. Thanks to #25182


dims:
/ok-to-test
/approve

leaving lgtm for a CAPI reviewer

Jont828:
@sbueringer Could you take a look?

PR #25234, opened 2022-02-12
log grabbing: obtain timestamps on extra services

bobbypage:
Previously if any test jobs added additional log dumps, such as
`containerd-installation` for example, the logs would grabbed with `sudo
journalctl --output=cat`. `--output=cat` excludes timestamps which makes
debugging difficult. Let's use `short-precise` output similar to to the
other log grabbing commands which includes timestamps.


bobbypage:
This is same change as the one over in k/k since this script is duplicated across test-infra and k/k.

See https://github.com/kubernetes/kubernetes/pull/108057

mborsz:
/retest
/lgtm
/approve

Thanks!

PR #25240, opened 2022-02-14
feat: enable priviliged mode for CAPA verify jobs

richardcase:
The CAPA verify job needs to run a container to verify its documents so
we need to enable priviliged mode.

Signed-off-by: Richard Case <richard@weave.works>


richardcase:
Relates to: https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3194

PR #25204, opened 2022-02-10
Cloneref mounts default GitHub fingerprint at runtime

chaodaiG:
https://github.com/kubernetes/test-infra/pull/25154 attempted to read this fingerprint at runtime, and failed for cloning with `https` in prow, unfortunately I could not repro the error locally. Mounting these files at runtime is an alternative approach, I have built the image manually and tested against `https` in a prow job and it worked.

/cc @cjwagner @stevekuznetsov

PR #25243, opened 2022-02-14
Add default Github know host file in base git image

chaodaiG:
This base image will be used by clonerefs, so that clonerefs image can be built by `ko publish`.

Clonerefs requires ssh_config and github_known_host files, they were part of the clonerefs image built by bazel. In an attempt to build clonerefs image with pure Golang tool such as ko (#25204 and #25154 ), these two files were created when clonerefs binary starts, but unfortunately saw some strange behaviors and both PRs were reverted.

/cc @cjwagner @BenTheElder @stevekuznetsov


chaodaiG:
> I'd prefer if we could just make one `git` base image include these files rather than creating separate base images for GitHub and non-GitHub use cases. We haven't had any conflicts with non-GitHub use cases so continuing to use a single image for both cases should work fine.

ack, and done

cjwagner (approved):

PR #25244, opened 2022-02-14
Added ci-kubernetes-kubemark-100-gce-old-kubekins

pawbana:
Added new experimental test which is copy of ci-kubernetes-kubemark-100-gce-oldtest to verify if new kubekins image causes flakiness.


pawbana:
Wrong guess. No need to verify as on both images the gcloud version is the same.
```
# gcloud --version
Google Cloud SDK 369.0.0
alpha 2022.01.14
beta 2022.01.14
bq 2.0.72
core 2022.01.14
gsutil 5.6
kubectl 1.20.8
```

PR #25246, opened 2022-02-14
Update to latest ko

chaodaiG:
ko v.0.9.3 doesn't work with workload identity, updating to latest should fix

ref: https://knative.slack.com/archives/CAVRMBUMA/p1644860274915189


hack/make-rules/verify/all.sh
```diff
@@ -25,49 +25,59 @@ res=0
...
+  echo "verifying $name"
+  hack/make-rules/verify/pylint.sh || { res=1; FAILED+=($name); echo "ERROR: $name failed"; }
   cd "${REPO_ROOT}"
 fi
 if [[ "${VERIFY_BOILERPLATE:-true}" == "true" ]]; then
-  echo "verifying boilerplate ..."
-  hack/make-rules/verify/boilerplate.sh || res=1
+  name="boilerplate"
+  echo "verifying $name"
+  hack/make-rules/verify/boilerplate.sh || { res=1; FAILED+=($name); echo "ERROR: $name failed"; }
```
listx, line 80: Nit: you could drop the `res=1` and just check if `FAILED` array is not empty ("if res is not 0 or if FAILED is not empty, show errors...").

hack/make-rules/verify/file-perms.sh
```diff
@@ -33,7 +33,7 @@ echo "Checking .sh files permission..."
 # find all files named *.sh (approximate shell script detection ...)
 # - ignoring .git
 # - that are not executable by all
-files=$(find . -type f -name '*.sh' -not -perm "${desired_perm}" -not -path './.git/*')
+files=$(find . -type f -name '*.sh' -not -perm "${desired_perm}" -not -path './.git/*' -path 'node_modules' -path '.bin')
```
listx, line 36: Do you need to place a `-not` in front of each new `-path` argument?

Also, nit: for many-argument calls, I would prefer to have each of the logical arguments separated out onto its own line with `\` (I think it should work even inside `$(...)`).

chaodaiG:
Upgrading to go1.17.7 seems to fix the problem of https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/test-infra/25246/pull-test-infra-verify-lint/1493333734416453632

/hold
Updating Go version might worth a separate PR

hack/make-rules/verify/all.sh
chaodaiG: done

hack/make-rules/verify/file-perms.sh
chaodaiG: good catch, fixed

chaodaiG:
/unhold
It's now ready to be reviewed. @cjwagner please take a look

cjwagner (approved):

PR #25250, opened 2022-02-14
Support expanding up and down

fejta:
* Allows expanding a few skipped lines up or down
* Currently we only support expanding the entire block
* Add unit test coverage to the `Callback()` function
* Fixes to `runlocal`
* Can now click either the text or the icon (not just the text)

image

### DIFF
--- a/prow/spyglass/lenses/buildlog/buildlog.ts
+++ b/prow/spyglass/lenses/buildlog/buildlog.ts
@@ -46,16 +46,50 @@ function ansiToHTML(orig: string): string {
   });
 }
 
-async function replaceElementWithContent(element: HTMLDivElement) {
+interface ArtifactRequest {
+  artifact: string;
+  bottom: number;
+  length: number;
+  offset: number;
+  startLine: number;
+  top: number;
+}
+
+async function replaceElementWithContent(element: HTMLDivElement, top: number, bottom: number) {
+
+  // <div data-foo="1" data-bar="this"> will show up as element.dataset = {"foo": "1", "bar": "this"}
   const {artifact, offset, length, startLine} = element.dataset;
-  const content = await spyglass.request(JSON.stringify({
-    artifact, length: +length!, offset: +offset!, startLine: +startLine!}));
-  element.innerHTML = ansiToHTML(content);
-  fixLinks(element);
+
+  // length! => we know these values are non-null:
+  // - we know this because its tightly coupled with template.html
+  // TODO(fejta): consider more robust code, looser coupling.
+  const r: ArtifactRequest = {
+    artifact: artifact!,
+    bottom,
+    length: Number(length!),
+    offset: Number(offset!),
+    startLine: Number(startLine!),
+    top,
+  };
+  const content = await spyglass.request(JSON.stringify(r));
   showElem(element);
+  element.outerHTML = ansiToHTML(content);
+  fixLinks(document.documentElement);
+  for (const button of Array.from(document.querySelectorAll<HTMLDivElement>(".show-skipped"))) {
+    if (button.classList.contains("showable")) {
+      continue;
+    }
+    button.addEventListener('click', handleShowSkipped);
+    button.classList.add("showable");
+  }
+
+  for (const button of Array.from(document.querySelectorAll<HTMLDivElement>(".show-skipped"))) {
+    button.addEventListener('click', handleShowSkipped);
+  }
 
   // Remove the "show all" button if we no longer need it.
-  const log = document.getElementById(`${artifact}-content`)!;
+  // TODO(fejta): avoid id selectors: https://google.github.io/styleguide/htmlcssguide.html#ID_Selectors
+  const log = document.getElementById(`${r.artifact}-content`)!;
   const skipped = log.querySelectorAll<HTMLElement>(".show-skipped");
   if (skipped.length === 0) {
     const button = document.querySelector('button.show-all-button')!;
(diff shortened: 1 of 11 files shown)


prow/spyglass/lenses/buildlog/lens.go
```diff
@@ -208,12 +221,59 @@ func (lens Lens) Callback(artifacts []api.Artifact, resourceDir string, data str
...
 	}
 	if err != nil {
-		return fmt.Sprintf("failed to retrieve log lines: %v", err)
+		return fmt.Sprintf("Failed to retrieve log lines: %v", err)
 	}
 
+	var skipFirst bool
+	var skipLines []string
+	skipRequest := request
+	if t, n := request.Top, len(lines); t > 0 && t < n {
```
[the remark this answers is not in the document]
fejta: Done

prow/spyglass/lenses/buildlog/buildlog.ts
```diff
@@ -46,16 +46,45 @@ function ansiToHTML(orig: string): string {
...
+  bottom: number;
+  length: number;
+  offset: number;
+  startLine: number;
+  top: number;
+}
+
+async function replaceElementWithContent(element: HTMLDivElement, top: number, bottom: number) {
+  // +length => make it a number (as opposed to a string)
+  // length! => I know this value is non-null
```
michelle192837, line 60: Style: This isn't safe; should either be a check, or explain that it can't be null + why. (https://google.github.io/styleguide/tsguide.html#type-and-non-nullability-assertions)

prow/spyglass/lenses/buildlog/buildlog.ts
```diff
@@ -46,16 +46,45 @@ function ansiToHTML(orig: string): string {
...
+  artifact: string;
+  bottom: number;
+  length: number;
+  offset: number;
+  startLine: number;
+  top: number;
+}
+
+async function replaceElementWithContent(element: HTMLDivElement, top: number, bottom: number) {
+  // +length => make it a number (as opposed to a string)
```
michelle192837, line 59: Oh, cool! :D

However,
Style: Looks like this should be Number()? https://google.github.io/styleguide/tsguide.html#type-coercion

prow/spyglass/lenses/buildlog/buildlog.ts
```diff
@@ -46,16 +46,45 @@ function ansiToHTML(orig: string): string {
...
+  const content = await spyglass.request(JSON.stringify(r));
   showElem(element);
+  element.outerHTML = ansiToHTML(content);
+  fixLinks(document.documentElement);
+  for (const button of Array.from(document.querySelectorAll<HTMLDivElement>(".show-skipped"))) {
+    if (button.classList.contains("showable")) {
+      continue;
+    }
+    button.addEventListener('click', handleShowSkipped);
+    button.className = button.className + " showable";
```
michelle192837, line 79: I think you can do element.classList.add('class') instead?

prow/spyglass/lenses/buildlog/buildlog.ts
```diff
@@ -46,16 +46,45 @@ function ansiToHTML(orig: string): string {
...
+  }
+
+  for (const button of Array.from(document.querySelectorAll<HTMLDivElement>(".show-skipped"))) {
+    button.addEventListener('click', handleShowSkipped);
+  }
 
   // Remove the "show all" button if we no longer need it.
-  const log = document.getElementById(`${artifact}-content`)!;
+  const log = document.getElementById(`${r.artifact}-content`)!;
```
michelle192837, line 87: Style: Avoid ID selectors (https://google.github.io/styleguide/htmlcssguide.html#ID_Selectors)

(This one's more an FYI than a "please fix" since our web stuff is small and I know we break this in TestGrid all the time, but it's apparently discouraged).

prow/spyglass/lenses/buildlog/buildlog.ts
```diff
@@ -137,7 +183,9 @@ async function loadLine(artifact: string, line: number): Promise<boolean> {
   const showers = document.querySelectorAll<HTMLDivElement>(`.show-skipped[data-artifact="${artifact}"]`);
   for (const shower of Array.from(showers)) {
     if (line >= Number(shower.dataset.startLine) && line < Number(shower.dataset.endLine)) {
-      await replaceElementWithContent(shower);
+      // TODO(fejta): could maybe do something smarter here than the whole
+      // block.
+      await replaceElementWithContent(shower, 0, 0);
```
michelle192837, line 188: I think it's possible to use a details element, if you're okay loading the contents asynchronously or on load? (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details) But not sure that quite fits with what's happening, and also not sure where latency's coming from on the replace.

michelle192837 (commented):
Bunch of style nits, sorry 😞 Otherwise LGTM!

prow/spyglass/lenses/buildlog/buildlog.ts
fejta: Ack.

I intend to work in this area in subsequent PRs. The latency is coming from `replaceElementWithContent()` sending a callback request to deck to fetch the additional lines: `const content = await spyglass.request(JSON.stringify(r));`

prow/spyglass/lenses/buildlog/buildlog.ts
fejta: Added a TODO. Mind if I do that refactor in a subsequent PR? This PR doesn't add the `getElementById()` call.

prow/spyglass/lenses/buildlog/buildlog.ts
fejta: Done

prow/spyglass/lenses/buildlog/buildlog.ts
fejta: Done

prow/spyglass/lenses/buildlog/buildlog.ts
fejta: Done-ish...

I think the idea here (note that this was `+offset!` etc before, is that this data is tightly coupled with how its defined in template.html, like a proto might be. So added a comment and a TODO to make this more robust.

prow/spyglass/lenses/buildlog/buildlog.ts
michelle192837: Yup!

prow/spyglass/lenses/buildlog/buildlog.ts
michelle192837: Sounds good!

michelle192837 (approved):
/hold

Unhold at will!

PR #25253, opened 2022-02-14
Change cookiefile config to be global

mpherman2:
IIUC we need to use global config because we can't clone without the cookiefile, and we can't use regular config before we clone.

/assign @chaodaiG @cjwagner


cjwagner (approved):
/hold cancel

PR #25261, opened 2022-02-15
add plugins settings for kubernetes-sigs/prow repo

listx:
We try to follow kubernets/test-infra (k/t-i) settings as much as
possible because this is already what the Prow developers are using
today (i.e., k/t-i is where Prow development takes place).

/cc @cjwagner

### DIFF
--- a/config/prow/plugins.yaml
+++ b/config/prow/plugins.yaml
@@ -86,6 +86,7 @@ approve:
   - kubernetes-sigs/kind
   - kubernetes-sigs/kubetest2
   - kubernetes-sigs/mdtoc
+  - kubernetes-sigs/prow
   - kubernetes-sigs/release-notes
   - kubernetes-sigs/release-sdk
   - kubernetes-sigs/release-utils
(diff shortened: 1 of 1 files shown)


config/prow/plugins.yaml
```diff
@@ -936,6 +938,15 @@ plugins:
...
     - require-matching-label
 
+  # Use the same settings as kubernetes/test-infra.
+  kubernetes-sigs/prow:
+    plugins:
+    - config-updater
+    - milestone
+    - override
+    - project
+    - require-matching-label
```
cjwagner, line 948: This plugin won't do anything without configuration for the repo. We don't need such configuration at this time so please don't enable this plugin. https://github.com/kubernetes/test-infra/blob/21ad9e9c0b29d34d0520e6b94a2880eba9fe257e/config/prow/plugins.yaml#L550-L760

listx:
/hold

Need to address these comments but also have a look at the plugins enabled for k/t-i (`kubernetes` org) and bring in what `kubernetes-sigs` org does not have. Reference: https://prow.k8s.io/plugins?repo=kubernetes%2Ftest-infra

listx:
Obsolete due to https://github.com/kubernetes/test-infra/pull/26382.

/close

PR #25266, opened 2022-02-15
[CAPA] fix: add dind label

richardcase:
Adding a missed label for enabling dind support.

Signed-off-by: Richard Case <richard@weave.works>


richardcase:
@sedefsavas - could you approve when you get time?

PR #25268, opened 2022-02-15
Yaml lint from make rule, remove one job

chaodaiG:
Verify yamllint from make rule, it can be invoked as part of `make verify`.

/cc @cjwagner @alvaroaleman

### DIFF
--- a/config/jobs/kubernetes/test-infra/test-infra-presubmits.yaml
+++ b/config/jobs/kubernetes/test-infra/test-infra-presubmits.yaml
@@ -44,22 +44,6 @@ presubmits:
       testgrid-dashboards: presubmits-test-infra
       testgrid-tab-name: gubernator
 
-  - name: pull-test-infra-yamllint
-    always_run: true
-    decorate: true
-    spec:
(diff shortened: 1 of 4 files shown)


config/jobs/kubernetes/test-infra/test-infra-presubmits.yaml
```diff
@@ -44,22 +44,6 @@ presubmits:
       testgrid-dashboards: presubmits-test-infra
       testgrid-tab-name: gubernator
 
-  - name: pull-test-infra-yamllint
```
alvaroaleman, line 47: The reason this was an extra job is that this needs `yamllint` in the image that runs the job. Does the image that runs `verify/all.sh` have `yamllint` installed?

chaodaiG: verify/all.sh shells out to verify/yamllint.sh, which does a docker run, so basically this uses DinD

alvaroaleman (approved):

PR #25271, opened 2022-02-15
Prow image builder supports multiple destinations

chaodaiG:
Apparently currently prow images are pushed to two GCR registries in the same postsubmit job https://github.com/kubernetes/test-infra/blob/f713022e15dcb7ced39b28847867941175c27aad/prow/def.bzl#L192, make image builder support multiple registries.

Ko has it's own cache, which only caches Go binaries. So processing both registries in a single call, instead of  twice, to avoid static files being generated again.

/cc @cjwagner

### DIFF
--- a/hack/prowimagebuilder/main.go
+++ b/hack/prowimagebuilder/main.go
@@ -66,21 +66,21 @@ var (
 )
 
 func init() {
-	out, err := runCmd("git", "rev-parse", "--show-toplevel")
+	out, err := runCmd(nil, "git", "rev-parse", "--show-toplevel")
 	if err != nil {
 		logrus.WithError(err).Error("Failed getting git root dir")
 		os.Exit(1)
 	}
 	rootDir = out
 
-	if _, err := runCmdInDirFunc(path.Join(rootDir, "hack/tools"), "go", "build", "-o", path.Join(rootDir, "_bin/ko"), "github.com/google/ko"); err != nil {
+	if _, err := runCmdInDirFunc(path.Join(rootDir, "hack/tools"), nil, "go", "build", "-o", path.Join(rootDir, "_bin/ko"), "github.com/google/ko"); err != nil {
 		logrus.WithError(err).Error("Failed ensure ko")
 		os.Exit(1)
 	}
 }
 
 type options struct {
-	koDockerRepo      string
+	dockerRepo        string
 	prowImageListFile string
 	workers           int
 	push              bool
@@ -90,11 +90,12 @@ type options struct {
 // Mock for unit testing purpose
 var runCmdInDirFunc = runCmdInDir
 
-func runCmdInDir(dir, cmd string, args ...string) (string, error) {
+func runCmdInDir(dir string, additionalEnv []string, cmd string, args ...string) (string, error) {
 	command := exec.Command(cmd, args...)
 	if dir != "" {
 		command.Dir = dir
 	}
+	command.Env = append(os.Environ(), additionalEnv...)
 	stdOut, err := command.StdoutPipe()
 	if err != nil {
 		return "", err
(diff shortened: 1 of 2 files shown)


hack/prowimagebuilder/main.go
```diff
@@ -217,23 +217,26 @@ func buildAndPush(id *imageDef, koDockerRepo string, push bool) error {
 	if err := setup(id); err != nil {
 		return fmt.Errorf("setup: %w", err)
 	}
-	_, err = runCmd("_bin/ko", publishArgs...)
-	return err
+	for _, dockerRepo := range dockerRepos {
```
BenTheElder, line 220: maybe a comment mentioning that ko only supports on repo at a time?

alternatively does it make sense to copy over the images between repos instead of rebuilding? (or perhaps a future TODO)

chaodaiG: fat finger, accidentally resolved

chaodaiG: good point, added a comment.

ko has built in cache, which makes the subsequent `ko` calls relatively cheap, so we should be fine here

BenTheElder (approved):
/lgtm
/approve
/hold

hack/prowimagebuilder/main.go
```diff
@@ -223,26 +223,31 @@ func buildAndPush(id *imageDef, koDockerRepo string, push bool) error {
...
+		}
 	}
 	return nil
 }
 
 func main() {
 	var o options
 	flag.StringVar(&o.prowImageListFile, "prow-images-file", path.Join(rootDir, defaultProwImageListFile), "Yaml file contains list of prow images")
-	flag.StringVar(&o.koDockerRepo, "ko-docker-repo", os.Getenv("KO_DOCKER_REPO"), "KO_DOCKER_REPO override")
+	flag.StringVar(&o.dockerRepo, "docker-repo", "", "Dockers repos, separated by comma")
```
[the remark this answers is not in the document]
chaodaiG: makes sense, done.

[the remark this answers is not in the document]
chaodaiG: Please re-stamp

chaodaiG:
rebased, please re-stamp

hack/prowimagebuilder/main.go
```diff
@@ -223,26 +223,31 @@ func buildAndPush(id *imageDef, koDockerRepo string, push bool) error {
...
 		return fmt.Errorf("setup: %w", err)
 	}
-	if _, err = runCmd("_bin/ko", publishArgs...); err != nil {
-		return fmt.Errorf("running ko: %w", err)
+	// ko only supports a single docker repo at a time, running this repeatedly
+	// on different docker repos so that multiple docker repos can be supported.
+	// This process utilized the built in cache of ko, so that pushing to
+	// subsequent docker repo(s) is relatively cheap.
+	for _, dockerRepo := range dockerRepos {
+		if err = os.Setenv("KO_DOCKER_REPO", dockerRepo); err != nil {
```
cjwagner, line 231: This is not threadsafe, right? (`buildAndPush` is called concurrently.)

cjwagner (changes requested):

hack/prowimagebuilder/main.go
chaodaiG: good catch, indeed
/hold

chaodaiG: updated to limit the scope of env var on command execution level, should be good now
/unhold

cjwagner: This still needs to be removed.

cjwagner (changes requested):
LGTM otherwise.

hack/prowimagebuilder/main.go
chaodaiG: good catch, removed, thank you!

cjwagner (approved):

PR #25273, opened 2022-02-15
prow image builder supports retry

chaodaiG:
It's observed that the image building process is flaky, either pulling docker base image, or pushing images

/cc @cjwagner

### DIFF
--- a/hack/prowimagebuilder/main.go
+++ b/hack/prowimagebuilder/main.go
@@ -45,6 +45,8 @@ const (
 
 	defaultWorkersCount = 10
 
+	defaultRetry = 3
+
 	// noOpKoDocerRepo is used when images are not pushed
 	noOpKoDocerRepo = "do.not/matter/at/all"
 )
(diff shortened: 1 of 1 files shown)


hack/prowimagebuilder/main.go
```diff
@@ -253,7 +272,7 @@ func main() {
 			for {
 				select {
 				case id := <-imageChan:
-					if err := buildAndPush(&id, o.koDockerRepo, o.push); err != nil {
+					if err := buildAndPush(&id, o.koDockerRepo, o.push, imageChan, &wg); err != nil {
```
cjwagner, line 275: In my opinion it would simpler to implement the retry logic as a loop here rather than adding requeuing to every return statement and passing down the WaitGroup and input channel. This looks correct too though.

cjwagner (approved):
/hold

hack/prowimagebuilder/main.go
chaodaiG: excellent idea, let me update

chaodaiG: done

cjwagner (approved):

PR #25276, opened 2022-02-15
Migrate ci-golang-tip-k8s-1-23 to k8s-infra-prow-build cluster

jprzychodzen:
Job ci-golang-tip-k8s-1-23 fails on scalability cluster, it looks related to service account key.

Migrating this job to community owned cluster. It has enough resources to handle this job.

/cc @marseel

### DIFF
--- a/config/jobs/kubernetes/sig-scalability/sig-scalability-golang.yaml
+++ b/config/jobs/kubernetes/sig-scalability/sig-scalability-golang.yaml
@@ -55,7 +55,7 @@ periodics:
 
 - interval: 4h
   name: ci-golang-tip-k8s-1-23
-  cluster: scalability
+  cluster: k8s-infra-prow-build
   tags:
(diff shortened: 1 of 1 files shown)


config/jobs/kubernetes/sig-scalability/sig-scalability-golang.yaml
```diff
@@ -55,7 +55,7 @@ periodics:
 
 - interval: 4h
   name: ci-golang-tip-k8s-1-23
-  cluster: scalability
+  cluster: k8s-infra-prow-build
```
marseel, line 58: We can also change project to --gcp-project-type=scalability-project I think

jprzychodzen: Right, good catch, thanks!

PR #25277, opened 2022-02-15
Separate ts rollup from codegen

chaodaiG:
Typescript rollup is done as part of image building process, the generated code is not sourced controlled, so they should not be part of codegen

/cc @cjwagner

### DIFF
--- a/hack/make-rules/verify/all.sh
+++ b/hack/make-rules/verify/all.sh
@@ -86,6 +86,12 @@ if [[ "${VERIFY_YAMLLINT:-true}" == "true" ]]; then
   hack/make-rules/verify/yamllint.sh || { FAILED+=($name); echo "ERROR: $name failed"; }
   cd "${REPO_ROOT}"
 fi
+if [[ "${VERIFY_TS_ROLLUP:-true}" == "true" ]]; then
+  name="rollup typescripts"
+  echo "verifying $name"
+  hack/make-rules/verify/ts-rollup.sh || { FAILED+=($name); echo "ERROR: $name failed"; }
+  cd "${REPO_ROOT}"
+fi
 
 # exit based on verify scripts
 if [[ "${#FAILED[@]}" == 0 ]]; then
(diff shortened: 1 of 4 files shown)


hack/make-rules/verify/all.sh
```diff
@@ -86,6 +86,12 @@ if [[ "${VERIFY_YAMLLINT:-true}" == "true" ]]; then
   hack/make-rules/verify/yamllint.sh || { FAILED+=($name); echo "ERROR: $name failed"; }
   cd "${REPO_ROOT}"
 fi
+if [[ "${VERIFY_TS_ROLLUP:-true}" == "true" ]]; then
```
BenTheElder, line 89: > Typescript rollup is done as part of image building process, the generated code is not sourced controlled, so they should not be part of codegen

should we even have verification then ...? (versus just ensuring that the generator runs before building images?)

chaodaiG: gopherage is the corner case here, the generated bundle is source controlled ahead of the entire bazel deprecation work. gopherage image is not built the same way as other prow images either. Alternatively we can consider not validating gopherage rollup at all. wdyt?

BenTheElder: hmm, can we validate _only_ gopherage? or source control all of it? 🤔

BenTheElder: (or not source control any of it?)

chaodaiG: source control all of it means `verify/codegen.sh` will need to verify them every time, which is equal amount of work as of now, and not source control any of them is something we can think about in the future.

I like the idea of only validating gopherage, updated the PR and added a comment

BenTheElder: > source control all of it means verify/codegen.sh will need to verify them every time, which is equal amount of work as of now, and not source control any of them is something we can think about in the future.

that's true, but it brings consistency / reduces surprises around what needs to be generated.

> I like the idea of only validating gopherage, updated the PR and added a comment

SGTM

BenTheElder (approved):
/lgtm
/approve

PR #25281, opened 2022-02-15
Build podutils with ko

chaodaiG:
Now that the git base image contains default github fingerprint, all pod utils images can be built with ko

/cc @cjwagner

### DIFF
--- a/hack/prowimagebuilder/main.go
+++ b/hack/prowimagebuilder/main.go
@@ -48,7 +48,7 @@ const (
 	defaultRetry = 3
 
 	// noOpKoDocerRepo is used when images are not pushed
-	noOpKoDocerRepo = "do.not/matter/at/all"
+	noOpKoDocerRepo = "ko.local"
 )
 
 var (
--- a/hack/prowimagebuilder/main.go
+++ b/hack/prowimagebuilder/main.go
@@ -25,8 +25,10 @@ import (
 	"os"
 	"os/exec"
 	"path"
+	"strconv"
 	"strings"
 	"sync"
+	"time"
 
 	"context"
 
@@ -44,8 +46,7 @@ const (
 	defaultProwImageListFile = "prow/.prow-images.yaml"
 
 	defaultWorkersCount = 10
-
-	defaultRetry = 3
+	defaultRetry        = 3
 
 	// noOpKoDocerRepo is used when images are not pushed
 	noOpKoDocerRepo = "ko.local"
@@ -242,6 +243,12 @@ func main() {
 	}
 	if err := os.Setenv("KO_DOCKER_REPO", o.koDockerRepo); err != nil {
 		logrus.WithError(err).Error("Failed setting KO_DOCKER_REPO")
+		os.Exit(1)
+	}
+	// By default ensures timestamp of images, ref:
+	// https://github.com/google/ko#why-are-my-images-all-created-in-1970
+	if err := os.Setenv("SOURCE_DATE_EPOCH", strconv.Itoa(int(time.Now().Unix()))); err != nil {
+		logrus.WithError(err).Error("Failed setting SOURCE_DATE_EPOCH")
 		os.Exit(1)
 	}
 
--- a/.ko.yaml
+++ b/.ko.yaml
@@ -11,19 +11,23 @@
 baseImageOverrides:
   k8s.io/test-infra/prow/cmd/branchprotector: gcr.io/k8s-prow/alpine:v20200713-e9b3d9d
   k8s.io/test-infra/prow/cmd/checkconfig: gcr.io/k8s-prow/git:v20220215-ddc3ad9
+  k8s.io/test-infra/prow/cmd/clonerefs: gcr.io/k8s-prow/git:v20220215-ddc3ad9
   k8s.io/test-infra/prow/cmd/config-bootstrapper: gcr.io/k8s-prow/git:v20220215-ddc3ad9
   k8s.io/test-infra/prow/cmd/deck: gcr.io/k8s-prow/git:v20220215-ddc3ad9
   k8s.io/test-infra/prow/cmd/exporter: gcr.io/k8s-prow/alpine:v20200713-e9b3d9d
   k8s.io/test-infra/prow/cmd/crier: gcr.io/k8s-prow/git:v20220215-ddc3ad9
+  k8s.io/test-infra/prow/cmd/entrypoint: gcr.io/k8s-prow/git:v20220215-ddc3ad9
   k8s.io/test-infra/prow/cmd/generic-autobumper: gcr.io/k8s-prow/git:v20220215-ddc3ad9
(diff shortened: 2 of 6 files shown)


chaodaiG:
/retest
should be much stable after #25273 is merged

chaodaiG:
Did a local testing, realized that the entrypoints of pod utils images are all overridden, for example:

- entrypoint: `cp /entrypoint /tools/entrypoint`
- clonerefs: `/clonerefs`
- initupload: `/initupload`
- sidecar: `/sidecar`

These are not the place where `ko` bin place images, which is `/ko-app`, so it won't work as is. Will need to figure out a way to ensure backward compatibility
/hold

alvaroaleman:
I think a changed location will break some of the Openshift downstream tooling, cc @petr-muller 
What is the reason we excplicitly set the `command` rather than relying on `Entrypoint` being correct, that bazel doesn't set `Entrypoint`?

chaodaiG:
> I think a changed location will break some of the Openshift downstream tooling, cc @petr-muller What is the reason we excplicitly set the `command` rather than relying on `Entrypoint` being correct, that bazel doesn't set `Entrypoint`?

I don't see a particular reason for setting `command` for `clonerefs`, `initupload`, and `sidecar`, just tested with `docker run -it gcr.io/k8s-prow/initupload` and the Entrypoint seemed to be correctly set, same with `clonerefs` and `sidecar`, I think they can safely be dropped. 

`entrypoint` image is different, the image does `cp /entrypoint /tools/entrypoint` instead of executing the binary, would the location change here breaking Openshift downstream tooling?

alvaroaleman:
> entrypoint image is different, the image does cp /entrypoint /tools/entrypoint instead of executing the binary

A creative solution would be to add logic to `entrypoint` to copy itself to a location configured via a flag or env var?

I did some git archaeology and what broke last time the binary location changed is `clonerefs`, because we have code that copies it out of one image and copies it into another one. To avoid that in the future, there is now a downstream image that copies `clonerefs` to a well-known location, so that would need updating: https://github.com/openshift/release/blob/9a953c36206552da0d519981f95142b5e9447b07/clusters/app.ci/supplemental-ci-images/manage-clonerefs.yaml#L30

Other than that if we work around the problem by not setting the command, there is some downstream code that does decoration (because openshift has a testrunner that creates more pods that are also decorated), but _I think_ that should just work once it has an updated t-i dependency: https://github.com/openshift/ci-tools/blob/e5be54a1382f18b156ed3f41ae2808bb9e4770d7/pkg/steps/artifacts.go#L395

chaodaiG:
> > entrypoint image is different, the image does cp /entrypoint /tools/entrypoint instead of executing the binary
> 
> A creative solution would be to add logic to `entrypoint` to copy itself to a location configured via a flag or env var?

I have thought about this as well :). This would for sure work I believe, and the problem is that it'll be breaking change, the command of entrypoint image will be changed from:
```
- command: cp
  arg:
  - /entrypoint
  - /tools/entrypoint
```
to

```
- arg:
  - --copy-myself=/tools/entrypoint
```
Even if we make it smart enough to operate in these two modes, the old pod utils images will not work with new decorating logic. Any idea how to solve this


> I did some git archaeology and what broke last time the binary location changed is `clonerefs`, because we have code that copies it out of one image and copies it into another one. To avoid that in the future, there is now a downstream image that copies `clonerefs` to a well-known location, so that would need updating: https://github.com/openshift/release/blob/9a953c36206552da0d519981f95142b5e9447b07/clusters/app.ci/supplemental-ci-images/manage-clonerefs.yaml#L30
> 
> Other than that if we work around the problem by not setting the command, there is some downstream code that does decoration (because openshift has a testrunner that creates more pods that are also decorated), but _I think_ that should just work once it has an updated t-i dependency: https://github.com/openshift/ci-tools/blob/e5be54a1382f18b156ed3f41ae2808bb9e4770d7/pkg/steps/artifacts.go#L395

Thank you for the context

alvaroaleman:
> Even if we make it smart enough to operate in these two modes, the old pod utils images will not work with new decorating logic.

That is true, but do we assume there are Prow setups that update prow itself but do not update the decoration utils images at the same time? Because that is the only situation where this would be a problem or not?

chaodaiG:
> > Even if we make it smart enough to operate in these two modes, the old pod utils images will not work with new decorating logic.
> 
> That is true, but do we assume there are Prow setups that update prow itself but do not update the decoration utils images at the same time? Because that is the only situation where this would be a problem or not?

Yeah technically we don't necessarily need to support mismatched versions of prow component and decoration components. There is a slightly chance of race, that prow components and decoration configs are not guaranteed to be updated at the same time, so the jobs scheduled in between would fail, but on the other hand this window should be very narrow. @cjwagner , wdyt?

cjwagner:
> Yeah technically we don't necessarily need to support mismatched versions of prow component and decoration components.

We shouldn't worry about supporting this at all in my opinion. I think the short window of version mismatch is fine. Now that we use `Make` to deploy, the deployment job is much faster. We could even do a manual deploy right when the PR merges if we want to reduce downtime further.

chaodaiG:
Cool, see https://github.com/kubernetes/test-infra/pull/25305

chaodaiG:
The blockers were removed by #25305  and #25301 , this PR is now pretty simple
/unhold

alvaroaleman:
/hold
downstream consumers of podutils need to be updated before we merge this

hack/prowimagebuilder/main.go
```diff
@@ -44,11 +46,10 @@ const (
 	defaultProwImageListFile = "prow/.prow-images.yaml"
 
 	defaultWorkersCount = 10
-
-	defaultRetry = 3
+	defaultRetry        = 3
 
 	// noOpKoDocerRepo is used when images are not pushed
-	noOpKoDocerRepo = "do.not/matter/at/all"
+	noOpKoDocerRepo = "ko.local"
```
BenTheElder, line 52: FWIW it is useful even when not pushing to have a predictable and reasonable name so tarballs can be `docker load` ed for development testing (or `kind load image-archive`).

BenTheElder: I typically use the location I _would_ have pushed to.

chaodaiG: For some reason ko failed to build multi-arch images into tarballs, so I chose this as alternative, hopefully no one needs to `docker load` tarball for development

petr-muller:
Thanks for holding this until we can adapt. Would it be a problem if this waited for next week? Most of RH engineering will not be working tomorrow and this looks like can cause a pretty stressful outage if we overlook something.

chaodaiG:
> Thanks for holding this until we can adapt. Would it be a problem if this waited for next week? Most of RH engineering will not be working tomorrow and this looks like can cause a pretty stressful outage if we overlook something.

I think it's fine to wait until next week. The last thing I want to see is breaking downstream and had to revert upstream, repeatedly. Thank you @alvaroaleman and @petr-muller  for thinking ahead of time!

petr-muller:
OK, we're prepared on our side. Thanks for holding this!

chaodaiG:
> OK, we're prepared on our side. Thanks for holding this!

Awesome!
/unhold

chaodaiG:
@cjwagner @alvaroaleman , could you review?

alvaroaleman (approved):

PR #25283, opened 2022-02-16
Add HeadSHA to InRepoConfig for gerrit, and fix client map

mpherman2:
There were two issues for InRepoConfig in gerrit

1) the map of Host -> clientFactory was using the wrong key causing the remote to be wrong for the majority of repos
2) It was only sending the base SHA to getPresubmits and getPostsubmits instead of both Base and Head

this PR fixes both of these issues

/assign @chaodaiG @cjwagner

### DIFF
--- a/prow/gerrit/adapter/adapter.go
+++ b/prow/gerrit/adapter/adapter.go
@@ -220,12 +220,12 @@ func (c *Controller) Sync() error {
 				return fmt.Errorf("makeCloneURI: %w", err)
 			}
 
-			cache, ok := c.repoCacheMap[cloneURI.Host]
+			cache, ok := c.repoCacheMap[cloneURI.String()]
 			if !ok {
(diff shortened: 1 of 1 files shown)


prow/gerrit/adapter/adapter.go
```diff
@@ -220,12 +220,12 @@ func (c *Controller) Sync() error {
 				return fmt.Errorf("makeCloneURI: %w", err)
 			}
 
-			cache, ok := c.repoCacheMap[cloneURI.Host]
+			cache, ok := c.repoCacheMap[cloneURI.String()]
```
chaodaiG, line 223: lgtm, could you add a comment here explaining what's wrong with `Host`?

chaodaiG (approved):
Just for unblocking, consider add the comment in a separate PR

PR #25300, opened 2022-02-16
Add hint about test freeze for prow bot approve comment

saschagrunert:
We now determine if we're in test-freeze and give contributors a hint if their PR has to be cherry-picked into the new release branch.

This is part of https://github.com/kubernetes/release/issues/2386, while we will request cherry-picks as intermediate step until the auto fast-forward goes live (planned for the v1.25 cycle)

cc @kubernetes/release-engineering


saschagrunert:
@kubernetes/sig-testing-pr-reviews PTAL

prow/plugins/approve/testfreeze/testfreeze.go
```diff
@@ -0,0 +1,147 @@
...
+		latestSemver semver.Version
+		latestBranch string
+	)
+
+	for _, ref := range refs {
+		if ref.Name().IsBranch() {
+			branch := ref.Name().Short()
+
+			// Filter for release branches
+			if strings.HasPrefix(branch, releaseBranchPrefix) {
```
petr-muller, line 91: The cycle block could be more readable with guard clauses instead of the nested ifs, wdyt?

prow/plugins/approve/approve.go
```diff
@@ -462,7 +463,17 @@ func handle(log *logrus.Entry, ghc githubClient, repo approvers.Repo, githubConf
...
 	notifications := filterComments(commentsFromIssueComments, notificationMatcher(botUserChecker))
 	latestNotification := getLast(notifications)
-	newMessage := updateNotification(githubConfig.LinkURL, opts.CommandHelpLink, opts.PrProcessLink, pr.org, pr.repo, pr.branch, latestNotification, approversHandler)
+
+	var testFreezeResult *testfreeze.Result
+	// Enable test freeze only for the k/k repo and master branch
+	if pr.org == "kubernetes" && pr.repo == "kubernetes" && pr.branch == "master" {
+		testFreezeResult, err = testfreeze.NewChecker(log).InTestFreeze()
+		if err != nil {
+			return fetchErr("Test Freeze status", err)
```
petr-muller, line 472: Would it makes sense to "survive" this error and just log it and continue? The point of the change is to add additional information to the comment, it probably does not make sense to abort the rest of the handler on e.g. a random git flake.

prow/plugins/approve/approve.go
```diff
@@ -462,7 +463,17 @@ func handle(log *logrus.Entry, ghc githubClient, repo approvers.Repo, githubConf
 	start = time.Now()
 	notifications := filterComments(commentsFromIssueComments, notificationMatcher(botUserChecker))
 	latestNotification := getLast(notifications)
-	newMessage := updateNotification(githubConfig.LinkURL, opts.CommandHelpLink, opts.PrProcessLink, pr.org, pr.repo, pr.branch, latestNotification, approversHandler)
+
+	var testFreezeResult *testfreeze.Result
+	// Enable test freeze only for the k/k repo and master branch
+	if pr.org == "kubernetes" && pr.repo == "kubernetes" && pr.branch == "master" {
```
petr-muller, line 469: I'm not sure how much I like hardcoding k/k logic into an approve plugin that is supposed to be generic, tbh. It does not seem to be too invasive though.

prow/plugins/approve/testfreeze/testfreeze.go
saschagrunert: Agreed, fixed by checking for `!strings.HasPrefix(branch, releaseBranchPrefix)` and continuing if so.

prow/plugins/approve/approve.go
saschagrunert: Yes, we now only log the error if we're unable to fetch the status result.

petr-muller:
Code LGTM but I'd like to give other approvers a chance to voice an opinion on the special k/k@master behavior in a generic plugin, cc @alvaroaleman @BenTheElder @cjwagner 

/lgtm

alvaroaleman:
> Code LGTM but I'd like to give other approvers a chance to voice an opinion on the special k/k@master behavior in a generic plugin

I am not really a big fan of that either, is there a reason why this needs to be part of the `approve` plugin rather than making it its own plugin that comments?

saschagrunert:
> > Code LGTM but I'd like to give other approvers a chance to voice an opinion on the special k/k@master behavior in a generic plugin
> 
> I am not really a big fan of that either, is there a reason why this needs to be part of the `approve` plugin rather than making it its own plugin that comments?

I think we could add an own plugin, yeah. If that's the preferred way then I'm happy to go that route.

saschagrunert:
Closing in favor of https://github.com/kubernetes/test-infra/pull/25535

PR #25301, opened 2022-02-16
Clonerefs, initupload, sidecar images use default entrypoints

chaodaiG:
Currently they are used by explicitly set command as symlink of the real binaries. This is not really necessary, as these images already has entrypoints set to the true paths of the binaries.

I have manually inspected the images from today, as well as the images from one year ago, and confirmed that the entrypoints of these images are set correctly. For example:

```
$ docker inspect gcr.io/k8s-prow/sidecar:v20210215-f94bc1dc87 | grep Entrypoint -A 3
            "Entrypoint": [
                "/app/prow/cmd/sidecar/app.binary"
            ],
            "OnBuild": null,
```

This will partially help unblock the effort of building these images using `ko`, as the binaries are placed at different path in the image

/cc @cjwagner @alvaroaleman @fejta

### DIFF
--- a/prow/pod-utils/decorate/podspec.go
+++ b/prow/pod-utils/decorate/podspec.go
@@ -266,11 +266,10 @@ func CloneLogPath(logMount coreapi.VolumeMount) string {
 
 // Exposed for testing
 const (
-	entrypointName   = "place-entrypoint"
-	initUploadName   = "initupload"
-	sidecarName      = "sidecar"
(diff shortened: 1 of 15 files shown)


alvaroaleman (approved):

chaodaiG:
@alvaroaleman may I have another lgtm please?

alvaroaleman (approved):

PR #25303, opened 2022-02-16
Ensure head shas are fetched before attempting to checkout and merge

mpherman2:
/assign @chaodaiG @cjwagner

### DIFF
--- a/prow/config/inrepoconfig.go
+++ b/prow/config/inrepoconfig.go
@@ -108,11 +108,23 @@ func prowYAMLGetter(
 
 	mergeMethod := c.Tide.MergeMethod(orgRepo)
 	log.Debugf("Using merge strategy %q.", mergeMethod)
+	if err := ensureHeadCommits(repo, headSHAs...); err != nil {
+		return nil, fmt.Errorf("failed to fetch headSHAs: %v", err)
+	}
 	if err := repo.MergeAndCheckout(baseSHA, string(mergeMethod), headSHAs...); err != nil {
 		return nil, fmt.Errorf("failed to merge: %w", err)
 	}
 
 	return ReadProwYAML(log, repo.Directory(), false)
+}
+
+func ensureHeadCommits(repo git.RepoClient, headSHAs ...string) error {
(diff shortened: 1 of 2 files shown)


prow/config/inrepoconfig.go
```diff
@@ -108,13 +108,25 @@ func prowYAMLGetter(
...
 		return nil, fmt.Errorf("failed to merge: %w", err)
 	}
 
 	return ReadProwYAML(log, repo.Directory(), false)
 }
 
+func ensureHeadCommits(repo git.RepoClient, headSHAs ...string) error {
+	for _, sha := range headSHAs {
+		if err := repo.Fetch("origin", sha); err != nil {
+			return err
```
chaodaiG, line 124: I would do `fmt.Errorf("failed to fetch headSHA: %s", sha)` here and return err above so that it's easier to debug

chaodaiG (approved):
/hold
feel free address my comment in future PR

prow/config/inrepoconfig.go
mpherman2: Done

chaodaiG (approved):

prow/git/v2/adapter.go
```diff
@@ -91,7 +91,8 @@ func (a *repoClientAdapter) MirrorClone() error {
 }
 
 func (a *repoClientAdapter) Fetch(arg ...string) error {
-	return a.Repo.Fetch(arg...)
+	args := append([]string{"origin"}, arg...)
```
chaodaiG, line 94: why not make this part of the call?

mpherman2: This is a bug with the gitv1 -> v2 adapter. In the v2 client Fetch does `git fetch remote args...` but in v1 it does `git fetch args...` This change brings the v1 client in line with the v2 client I believe

mpherman2: Added a todo to figure out how to fix the differing functionality without hard-coding origin

chaodaiG (approved):

PR #25304, opened 2022-02-16
kops: use k8s.local instead of k8s.io

jiahuif:
because we want to use gossip DNS, and we do not have access to DNS API anyways.

Setting domain to end with `k8s.local` tells the kops-deployer we want such case.


jiahuif:
/assign @justinsb @hakman 
Could you take a look when available? The project *should* be almost there.

jiahuif:
Reference:

https://github.com/kubernetes/kops/blob/v1.24.0-alpha.2/pkg/dns/gossip.go#L24

justinsb:
Thanks @jiahuif 

/approve
/lgtm

PR #25305, opened 2022-02-16
Entrypoint image can run in copy only mode

chaodaiG:
Currently entrypoint container runs `cp /entrypoint /tools/entrypoint` in a prow job, which is the last prow image that cannot use binary as entrypoint. Adding a copy mode so that entrypoint container can use the binary as entrypoint.

Note that this is a workaround, but it should not be worse than the workaround of running cp from the container

Following @alvaroaleman 's creative idea from https://github.com/kubernetes/test-infra/pull/25281#issuecomment-1041918936

/cc @alvaroaleman @cjwagner

### DIFF
--- a/prow/cmd/entrypoint/main.go
+++ b/prow/cmd/entrypoint/main.go
@@ -17,7 +17,10 @@ limitations under the License.
 package main
 
 import (
+	"fmt"
+	"io/ioutil"
 	"os"
+	"path"
 
 	"github.com/sirupsen/logrus"
 	"k8s.io/test-infra/prow/entrypoint"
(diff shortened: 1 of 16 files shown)


chaodaiG:
I'll add unit test if this makes sense to you guys

prow/cmd/entrypoint/main.go
```diff
@@ -25,9 +28,61 @@ import (
 	"k8s.io/test-infra/prow/pod-utils/options"
 )
 
+const (
+	copyModeOnlyEnvVar = "COPY_BINARY_ONLY"
+
+	defaultCopyDst = "/tools/entrypoint"
+)
+
+var defaultCopySrc = []string{"/ko-app/entrypoint", "/entrypoint"}
```
alvaroaleman, line 37: Shouldn't we be able to use `os.Args[0]` for the source location (and derefernece it if needed, IIRC bazel gives you a softlink as entrypoint)?

prow/cmd/entrypoint/main.go
```diff
@@ -25,9 +28,61 @@ import (
...
+	if err := ioutil.WriteFile(dst, body, info.Mode()); err != nil {
+		return fmt.Errorf("write '%s': %w", dst, err)
+	}
+	return nil
+}
+
 func main() {
 	logrusutil.ComponentInit()
 
+	if os.Getenv(copyModeOnlyEnvVar) == "true" {
```
alvaroaleman, line 79: Is there a reason to not use a flag? The advantage of using a flag is that it is discoverable by calling `--help` which isn't the case for an env var

prow/cmd/entrypoint/main.go
```diff
@@ -25,9 +28,61 @@ import (
...
+		if err := copy(defaultCopySrc, defaultCopyDst); err != nil {
```
alvaroaleman, line 80: Maybe make the dest configurable?

prow/cmd/entrypoint/main.go
chaodaiG: good point, updated

prow/cmd/entrypoint/main.go
chaodaiG: done

prow/cmd/entrypoint/main.go
chaodaiG: actually I was wrong, os.Args[0] is the path of the symlink

chaodaiG: but the copy function is smart enough to copy the real binary

chaodaiG:
Unit test added
@cjwagner @alvaroaleman please take another look

cjwagner (approved):

prow/cmd/entrypoint/main_test.go
```diff
@@ -0,0 +1,75 @@
...
+			name:     "base",
+			fileMode: 0644,
+		},
+		{
+			name:     "another-mode",
+			fileMode: 0755,
+		},
+	}
+
+	srcDir, err := ioutil.TempDir("", "test-prow-entrypoint-main")
```
alvaroaleman, line 41: please use t.Tmpdir, otherwise this will be left behind by the test

